PR #22333, opened 2026-04-28
Column-level metadata loading and serialization

ywelsch:
When making changes to only a subset of columns, DuckDB currently requires loading and serializing the full column metadata for all other columns in the touched rowgroups too. For large tables, this can create a significant IO and memory overhead (as this memory is not tracked by the buffer pool).

This PR makes it so that operations such as `ALTER TABLE ... ADD COLUMN | REMOVE COLUMN | ALTER TYPE` as well as `UPDATE` operations only need to load and write out the metadata for the changed columns.

It further extends the `pragma_storage_info` table function with a new named parameter called `only_loaded_segments` which, when explicitly set to true, will only return the segment information for those segments that are actually loaded in memory. This helps test that the change has the desired effect (i.e. only relevant columns are loaded).


src/storage/table/row_group.cpp
```diff
@@ -453,20 +460,54 @@ unique_ptr<RowGroup> RowGroup::AlterType(RowGroupCollection &new_collection, con
 		column_data->Append(append_state, append_vector, scan_chunk.size());
 	}
 
+	auto supports_per_column_writes = SupportsPerColumnWrites(new_collection);
+	if (!supports_per_column_writes) {
+		// ensure all columns are loaded, as checkpointing will need to load them anyway, and it's better to fail-fast
+		// in case these don't fit in memory here
+		GetColumns();
```
Mytherin, line 484: I don't really see what the benefits of this are - can we remove this?

src/storage/table/row_group.cpp
```diff
@@ -493,25 +534,54 @@ unique_ptr<RowGroup> RowGroup::AddColumn(RowGroupCollection &new_collection, Col
 		}
 	}
 
+	if (!SupportsPerColumnWrites(new_collection)) {
+		// ensure all columns are loaded, as checkpointing will need to load them anyway, and it's better to fail-fast
+		// in case these don't fit in memory here
+		GetColumns();
```
Mytherin, line 545: Same here - this seems to just add complexity / reduce in-depth testing of these code paths for no benefit

src/storage/table/row_group.cpp
```diff
@@ -493,25 +534,54 @@ unique_ptr<RowGroup> RowGroup::AddColumn(RowGroupCollection &new_collection, Col
...
+	}
+	bool has_column_pointers = column_pointers.size() == columns.size();
+	bool has_per_column_blocks = per_column_metadata_blocks.size() == columns.size();
+	if (has_column_pointers) {
+		row_group->column_pointers.resize(new_column_count);
+	}
+	if (has_per_column_blocks) {
+		row_group->per_column_metadata_blocks.resize(new_column_count);
+	}
+	row_group->has_metadata_blocks = false;
```
Mytherin: Can we extract some of this code into a function? This is adding a lot of code that's duplicated between the alter methods.

src/include/duckdb/storage/table/row_group_collection.hpp
```diff
@@ -135,7 +135,7 @@ class RowGroupCollection {
 	void CommitDropTable();
 
 	vector<PartitionStatistics> GetPartitionStats() const;
-	vector<ColumnSegmentInfo> GetColumnSegmentInfo(const QueryContext &context);
+	vector<ColumnSegmentInfo> GetColumnSegmentInfo(const QueryContext &context, bool only_loaded_segments = false);
```
Mytherin: Can we make this an enum? Bools as parameters are a bit of an anti-pattern from a readability perspective and we've been trying to get rid of them as much as possible in new code.

src/storage/table/row_group.cpp
```diff
@@ -1503,6 +1733,9 @@ void RowGroup::Serialize(RowGroupPointer &pointer, Serializer &serializer) {
 		serializer.WriteProperty(104, "has_metadata_blocks", pointer.has_metadata_blocks);
 		serializer.WritePropertyWithDefault(105, "extra_metadata_blocks", pointer.extra_metadata_blocks);
 	}
+	if (serializer.ShouldSerialize(8)) {
+		serializer.WritePropertyWithDefault(106, "per_column_metadata_blocks", pointer.per_column_metadata_blocks);
```
Mytherin: Once we have this the `extra_metadata_blocks` are no longer really necessary, since we can compute them from `per_column_metadata_blocks` and `data_pointers`.

Maybe we can stop serializing `extra_metadata_blocks` when targeting newer versions (`>= 8`) and only serialize it when targeting older versions `>= 6 and < 8`)? We can then compute `extra_metadata_blocks` when we need them instead.

Or alternatively we could just stop (de)serializing `extra_metadata_blocks` entirely given it's optional. I think it's not critical to make it fast when checkpointing on an older DuckDB version after having checkpointed with a newer DuckDB version and vice versa. So that has my vote - but I think that change should definitely only go into `main` and not be part of a potential backport.

src/storage/table/row_group.cpp
```diff
@@ -1291,33 +1399,105 @@ bool RowGroup::CanReuseMetadata(RowGroupWriter &writer) const {
...
-			                        "group has %llu rows, column has %llu)",
-			                        column_idx, this->count.load(), column.count.load());
+
+		if (!column_has_changes) {
+			// reuse this column's metadata
+			result.reuse_column[column_idx] = true;
+			result.states.push_back(nullptr);
+			result.keep_column_loaded.push_back(true);
+			// carry forward existing column data and statistics
+			if (is_loaded && !is_loaded[column_idx]) {
```
Mytherin: `is_loaded && !is_loaded[column_idx]` is a pattern that is repeated a lot - I think I added a helper method for this.

src/storage/table/row_group.cpp
```diff
@@ -1230,6 +1324,40 @@ bool RowGroup::HasUnloadedDeletes() const {
 	return !deletes_is_loaded;
 }
 
+vector<vector<idx_t>> RowGroup::ComputePerColumnMetadataBlocks(bool contiguous_layout) const {
```
Mytherin: Maybe `LegacyFetchPerColumnMetadataBlocks` and remove the parameter, given it's always `true`?

src/storage/table/row_group.cpp
```diff
@@ -1370,62 +1552,110 @@ RowGroupPointer RowGroup::Checkpoint(RowGroupWriteData write_data, RowGroupWrite
...
+		if (is_reused) {
+			// reuse existing column pointer and per-column blocks
+			row_group_pointer.data_pointers.push_back(column_pointers[column_idx]);
+			row_group_pointer.per_column_metadata_blocks.push_back(per_column_metadata_blocks[column_idx]);
+
+			// collect all blocks for this reused column for ClearModifiedBlocks
+			reused_column_blocks.push_back(column_pointers[column_idx]);
+			for (auto &block_id : per_column_metadata_blocks[column_idx]) {
+				reused_column_blocks.emplace_back(block_id, 0);
+			}
```
Mytherin, line 1596: early-out: `continue`

src/storage/table/row_group.cpp
```diff
@@ -1513,6 +1746,8 @@ RowGroupPointer RowGroup::Deserialize(Deserializer &deserializer) {
 	result.deletes_pointers = deserializer.ReadProperty<vector<MetaBlockPointer>>(103, "delete_pointers");
 	result.has_metadata_blocks = deserializer.ReadPropertyWithExplicitDefault<bool>(104, "has_metadata_blocks", false);
 	result.extra_metadata_blocks = deserializer.ReadPropertyWithDefault<vector<idx_t>>(105, "extra_metadata_blocks");
+	result.per_column_metadata_blocks =
+	    deserializer.ReadPropertyWithDefault<vector<vector<idx_t>>>(106, "per_column_metadata_blocks");
```
Mytherin: Given that this is usually empty for most columns, at least for primitive column types, I wonder if a different representation than `vector<vector>>` makes sense? If we have 1000 columns and 1 column that needs this we will now write a vector with size 1000, followed by 999 empty vectors. 

I think a representation like this seems most efficient in most scenarios:

```
Column id #7
<blocks>
Column id #27
<blocks>
```

We could achieve this by e.g. serializing a struct like this (that gets written as a single `idx_t`).

```cpp
struct PerColumnMetadataBlock {
	bool is_column_index : 1;
	idx_t index : 63;
};
```

This is guaranteed to work because DuckDB already doesn't support block ids larger than `2^56` (also used in `MetadataPointer`). 

Even in the "worst case" where we have 1 block per column this is more efficient than this nested-vector representation, given that the overhead of a vector is more than storing an additional integer.

Mytherin (commented):
Thanks for the PR! Looks good - cool to get even more partial checkpointing. I think this is a positive overall although it's worth noting this will cause more fragmentation of metadata over time which can (1) cause metadata bloat, and (2) cause metadata for individual row groups to be more scattered across the file requiring more random access. Eventually we might need a "metadata vacuum" style operation to create densely packed metadata again.

I left some comments below.

src/storage/table/row_group.cpp
ywelsch: The situation this tries to address is when you are still on an old storage version, where we know that we will have to load all columns in order to checkpoint. Then it's preferable to load them all here as part of the actual ALTER TABLE command (as it used to be before this PR), as you otherwise risk ending up in a poisonous situation where the ALTER TABLE command is committed, DuckDB fails on a subsequent checkpoint, and any future checkpoint will fail too (because there's not enough memory available).

Do you still want me to remove this?

src/storage/table/row_group.cpp
ywelsch: I've reworked / deduplicated the code here

src/include/duckdb/storage/table/row_group_collection.hpp
ywelsch: I've made this an enum, both in the code, as well as when exposing the parameter to SQL. Let me know if this works for you, or if you only have the former but not the latter in mind.

src/storage/table/row_group.cpp
ywelsch: I've adapted it so that when targeting storage format 2.0 we stop serializing the legacy extra_metadata_blocks.

I do still want to keep the field though when targeting older storage formats, as we heavily rely on this optimization, and we will only switch the storage format some time (multiple weeks) after rolling out DuckDB 2.0.

src/storage/table/row_group.cpp
ywelsch: Thanks, I've used the convenience method `ColumnIsLoaded` now in all these places

src/storage/table/row_group.cpp
ywelsch: I've reworked this completely and removed the parameter

src/storage/table/row_group.cpp
ywelsch: done (as part of larger rework)

src/storage/table/row_group.cpp
ywelsch: I've adopted the above representation.

src/storage/table/row_group_collection.cpp
```diff
@@ -1681,72 +1689,102 @@ void RowGroupCollection::Checkpoint(TableDataWriter &writer, TableStatistics &gl
 		new_row_groups->AppendSegment(l, std::move(new_row_group));
 
 		if (debug_verify_blocks) {
-			if (!pointer_copy.has_metadata_blocks) {
+			if (!pointer_copy.has_metadata_blocks && !pointer_copy.has_per_column_metadata_blocks) {
```
ywelsch, line 1704: All the changes in this method below are additional verification code (when running with debug_verify_blocks enabled)

src/storage/table/row_group.cpp
```diff
@@ -1129,6 +1175,26 @@ CompressionType ColumnCheckpointInfo::GetCompressionType() {
 	return info.compression_types[column_idx];
 }
 
+shared_ptr<ColumnData> RowGroup::CheckpointColumn(const RowGroup &row_group, idx_t column_idx, RowGroupWriteInfo &info,
```
ywelsch, line 1211: this is just existing code factored out into a new helper method

src/storage/table/row_group.cpp
```diff
@@ -1340,26 +1475,34 @@ RowGroupPointer RowGroup::Checkpoint(RowGroupWriteData write_data, RowGroupWrite
...
+				    [&](idx_t block_id) { metadata_block_pointers_to_be_cleared.emplace_back(block_id, 0); });
+			} else {
+				metadata_block_pointers_to_be_cleared.reserve(write_data.existing_extra_metadata_blocks.size());
+				for (auto &block_pointer : write_data.existing_extra_metadata_blocks) {
+					metadata_block_pointers_to_be_cleared.emplace_back(block_pointer, 0);
+				}
 			}
 			metadata_manager->ClearModifiedBlocks(column_pointers);
-			metadata_manager->ClearModifiedBlocks(extra_metadata_block_pointers);
-			metadata_manager->ClearModifiedBlocks(deletes_pointers);
```
ywelsch, line 1391: This was not making much sense (especially as these were not even the deletes_pointers returned from CheckpointDeletes, but potentially old ones). The CheckpointDeletes method already takes care of calling `ClearModifiedBlocks`.

src/storage/table/row_group.cpp
```diff
@@ -1502,9 +1687,23 @@ void RowGroup::Serialize(RowGroupPointer &pointer, Serializer &serializer) {
...
+			pointer.per_column_metadata_blocks.ForEachBlock(
+			    [&](idx_t block_id) { extra_metadata_block_ids.push_back(block_id); });
+			serializer.WriteProperty(104, "has_metadata_blocks", true);
+			serializer.WritePropertyWithDefault(105, "extra_metadata_blocks", extra_metadata_block_ids);
+		} else {
+			serializer.WriteProperty(104, "has_metadata_blocks", pointer.has_metadata_blocks);
+			serializer.WritePropertyWithDefault(105, "extra_metadata_blocks", pointer.extra_metadata_blocks);
+		}
+	}
+	if (serializer.ShouldSerialize(8) || pointer.has_per_column_metadata_blocks) {
```
ywelsch: The second condition captures the case where the `force_column_metadata_reuse` flag was enabled.

It then allows double-writing both "extra_metadata_blocks" and "per_column_metadata_blocks" when targeting 1.4 or 1.5 storage format.

ywelsch:
Thanks for the review, Mark. I've adopted your review comments, but also made more substantial changes to the PR (better code structure / testing / ...).

In particular I was already anticipating what a 1.5 backport would look like (as that also has some implications on the 2.0 code, given that we need a clean transition between the two).

I've added a `force_column_metadata_reuse` setting (default is false) which forces re-use of row group metadata on a column-level when checkpointing on older storage versions 6 and 7. This breaks storage backward-compatibility with older DuckDB versions, so is something that a user would have to explicitly opt into.

We would need the same setting when migrating to DuckDB 2.0 as we bump our storage version only some weeks after migration (as we are also running for some time in a mixed environment where some DuckDBs are still on 1.5 while others are on 2.0, sharing the same databases). So it makes sense to have it here too.

The backport would be then be almost exactly the same code as here, the only difference being an adaptation of the SupportPerColumnWrites method (and small RowGroup::Serialize changes)

src/storage/table/row_group_collection.cpp
```diff
@@ -1681,72 +1689,102 @@ void RowGroupCollection::Checkpoint(TableDataWriter &writer, TableStatistics &gl
...
+			if (partial_reuse) {
+				for (idx_t col_idx = 0; col_idx < pointer_copy.data_pointers.size(); col_idx++) {
+					if (!reuse_column[col_idx]) {
+						continue;
+					}
+					// reused column: its start block and extra blocks should be cleared
+					if (!block_manager.GetMetadataManager().BlockHasBeenCleared(pointer_copy.data_pointers[col_idx])) {
+						throw InternalException("Partial reuse: column %llu start block was not cleared", col_idx);
+					}
+					for (auto &block_id : pointer_copy.per_column_metadata_blocks.GetBlocksForColumn(col_idx)) {
```
Mytherin: This is potentially `O(n^2)` as we're doing a full scan of `per_column_metadata_blocks` for each column. I would say ideally (1) `per_column_metadata_blocks` is sorted on column id, and (2) we would rework `GetBlocksForColumn` to have `GetBlocksForColumns` that would look more like this so we can do a single pass scan:

```cpp
vector<vector<idx_t>> GetBlocksForColumns(const vector<column_t> &columns);
```

I suppose it's unlikely to matter much but anything `O(n^2)` can blow up pretty fast.

src/include/duckdb/storage/data_pointer.hpp
```diff
@@ -78,7 +79,13 @@ struct RowGroupPointer {
 	bool has_metadata_blocks = false;
 	//! Metadata blocks of the columns that are not mentioned in "data_pointers"
 	//! This is often empty - but can be set for wide columns with a lot of metadata
+	//! When targeting 2.0 storage format, per_column_metadata_blocks is used instead
```
Mytherin, line 82: I think it would be great to get rid of `extra_metadata_blocks` and `has_metadata_blocks`. This will considerably simplify the code versus keeping both `extra_metadata_blocks` and `per_column_metadata_blocks` around.

src/storage/table/per_column_metadata_blocks.cpp
```diff
@@ -0,0 +1,81 @@
...
+namespace duckdb {
+
+static constexpr idx_t COLUMN_INDEX_BIT = idx_t(1) << 63;
+
+void PerColumnMetadataBlock::Serialize(Serializer &serializer) const {
+	idx_t packed = index;
+	if (is_column_index) {
+		packed |= COLUMN_INDEX_BIT;
+	}
+	serializer.WriteProperty<idx_t>(100, "packed", packed);
```
Mytherin: This means we're writing each entry as an object which introduces extra field ids in the serialization format for each value. I would say we should just write it as an `idx_t`. Check e.g. how we write `TableIndex` or `PhysicalIndex` in the serializer.

src/storage/table/row_group.cpp
```diff
@@ -1294,33 +1342,120 @@ bool RowGroup::CanReuseMetadata(RowGroupWriter &writer) const {
...
+	result_row_group->deletes_is_loaded = deletes_is_loaded.load();
+	result_row_group->owned_version_info = owned_version_info;
+	result_row_group->version_info = version_info.load();
+	// copy metadata pointers so checkpoint can access them for reused columns
+	result.existing_column_pointers = column_pointers;
+	result.has_per_column_metadata_blocks = has_per_column_metadata_blocks;
+	result.existing_per_column_metadata_blocks = per_column_metadata_blocks;
+	result_row_group->is_loaded = unique_ptr<atomic<bool>[]>(new atomic<bool>[GetColumnCount()]);
+	for (idx_t c = 0; c < GetColumnCount(); c++) {
+		result_row_group->is_loaded[c] = true;
```
Mytherin: If all columns are loaded we can leave this empty - maybe we can create and instantiate this lazily instead?

src/storage/table/row_group.cpp
```diff
@@ -1233,40 +1283,42 @@ bool RowGroup::HasUnloadedDeletes() const {
...
-		result_as_set.erase(ptr.block_pointer);
+	for (idx_t i = 0; i < column_pointers.size(); i++) {
+		unordered_set<idx_t> result_as_set;
+		auto &start = column_pointers[i];
+		if (i < last_idx) {
+			// follow the linked list to the next column's start
+			auto &next_start = column_pointers[i + 1];
+			MetadataReader reader(metadata_manager, start);
+			auto blocks = reader.GetRemainingBlocks(next_start);
+			for (auto &ptr : blocks) {
```
Mytherin: I suppose this is still able to use this method of getting the blocks because `ComputePerColumnMetadataBlocks()` should only be called for old files that do not have this information, right? That also means that for new files the `per_column_metadata_blocks` are not optional but mandatory - otherwise this code is incorrect.

Maybe we should just always use the full deserialization - I'm not sure how much time is really being saved here by not deserializing the columns. At least I think we should leave a comment here explaining this nuance.

src/storage/table/row_group.cpp
```diff
@@ -1294,33 +1342,120 @@ bool RowGroup::CanReuseMetadata(RowGroupWriter &writer) const {
...
-		result.existing_extra_metadata_blocks = GetOrComputeExtraMetadataBlocks();
-		return result;
+		result.fully_reuse_existing_metadata_blocks = true;
+		if (GetCollection().SupportsPerColumnWrites()) {
+			result.has_per_column_metadata_blocks = true;
+			if (has_per_column_metadata_blocks) {
+				result.existing_per_column_metadata_blocks = per_column_metadata_blocks;
+				return result;
+			}
+			result.existing_per_column_metadata_blocks = ComputePerColumnMetadataBlocks();
```
Mytherin: I think we can simplify the upstream code by just computing the `metadata_blocks` that we need to write here instead of propagating all of this information back up.

src/include/duckdb/storage/table/row_group.hpp
```diff
@@ -80,9 +82,15 @@ struct RowGroupWriteData {
 	vector<unique_ptr<ColumnCheckpointState>> states;
 	vector<BaseStatistics> statistics;
 	vector<bool> keep_column_loaded;
-	bool reuse_existing_metadata_blocks = false;
+	bool fully_reuse_existing_metadata_blocks = false;
```
Mytherin: I think this could be more readable if this was a single enum that signals what we're doing, e.g. something like this:

```cpp
enum class RowGroupWriteAction {
     REUSE_EXISTING_ROW_GROUP_METADATA,
     PARTIALLY_REUSE_COLUMN_METADATA,
     FULLY_CHECKPOINT_ROW_GROUP
};
```

src/include/duckdb/storage/table/row_group.hpp
```diff
@@ -80,9 +82,15 @@ struct RowGroupWriteData {
...
 	vector<idx_t> existing_extra_metadata_blocks;
+	bool has_per_column_metadata_blocks = false;
+	PerColumnMetadataBlocks existing_per_column_metadata_blocks;
+	vector<MetaBlockPointer> existing_column_pointers;
 	optional_idx write_count;
+	//! Per-column reuse flags for partial column checkpoint
+	//! When non-empty, states[i] is nullptr for columns where reuse_column[i] is true
+	vector<bool> reuse_column;
```
Mytherin: We can figure out the same information by checking `states[i]` then, no? I think we can remove this vector then, or is there another reason to keep it around?

Mytherin (commented):
Thanks for the updates! Looks good - left some comments.

src/storage/table/row_group_collection.cpp
ywelsch: The quadratic blow-up was in the assertions / block verification code. I agree it's not very nice though.
I've changed this in d2fc8b5c8f now (per_column_metadata_blocks is sorted on column id now, and the access is always linear).

src/include/duckdb/storage/data_pointer.hpp
ywelsch: I'm fine dropping support for extra_metadata_blocks and has_metadata_blocks when targeting this branch here (for DuckDB 2.0), but for a 1.5 backport I think this still needs to exist to provide a reasonable upgrade path for us (i.e. we first get the code in, then gradually switch users over to make use of `force_column_metadata_reuse` while providing a temporary phase where we can alternate between turning this on and off, and not lose out on the existing metadata_reuse optimization).

So I was thinking of first merging this into main here with the added complexity, and then follow up with a PR that drops it (once we have moved our users to `force_column_metadata_reuse`)?

src/storage/table/per_column_metadata_blocks.cpp
ywelsch: Thanks, I've changed this in 373cfa308f

src/storage/table/row_group.cpp
ywelsch: I've changed it in 46dd436d6d to carry at least the flag from the original rowgroup forward.

src/storage/table/row_group.cpp
ywelsch: > I suppose this is still able to use this method of getting the blocks because ComputePerColumnMetadataBlocks() should only be called for old files that do not have this information, right? That also means that for new files the per_column_metadata_blocks are not optional but mandatory - otherwise this code is incorrect.

Correct, those are the conditions / requirements.

> Maybe we should just always use the full deserialization - I'm not sure how much time is really being saved here by not deserializing the columns. At least I think we should leave a comment here explaining this nuance.

The deserialization should only come into play when targeting a storage format < 1.4 (which is still the default though in DuckDB).
There is a performance difference between just walking the linked list of blocks and fully deserializing the columns, but I can't quantify it. We are using storage format 1.4+ (with metadata reuse) in our service, so don't need to recompute this anymore. I'm fine therefore going back to the simpler full deserialization in dd8839d08a.

src/storage/table/row_group.cpp
ywelsch: I'm not sure what you mean - can you clarify?

src/include/duckdb/storage/table/row_group.hpp
ywelsch: Changed in 56831f684d and 6785b4e130

src/include/duckdb/storage/table/row_group.hpp
ywelsch: Makes sense, I've fixed this in eb16932ad5

src/storage/table/row_group.cpp
Mytherin: When `fully_reuse_existing_metadata_blocks` is set we are not writing any metadata - we are just calling `ClearModifiedBlocks` on the metadata blocks. Currently we are passing up a bunch of information that we can use to then compute the metadata blocks in that situation - but that seems to split up the complexity in two locations which is imo less understandable. It would be better if here we just say "we are re-using the existing row group metadata, here are the metadata blocks".

ywelsch: I've refactored this in [b5b1b3c](https://github.com/duckdb/duckdb/pull/22333/commits/b5b1b3ca63403b045acc74b95a03771dab1f777c) now and eliminated the need to pass this info explicitly. I hope this is what you meant.

I've still kept the calls of `ComputePerColumnMetadataBlocks` in the WriteToDisk method here as it is called in parallel, and allows this metadata loading to happen in parallel. If we move that to `RowGroup::Checkpoint`, then we need to do that work all in sequence. As this can amount to loading larger amounts of metadata for big databases (e.g. 20-30GB), I think it would be preferable to keep that in the parallel path. That's how it was previously too (for the full row-group level metadata reuse). Further we would have to duplicate the loading into the table-level reuse too (RowGroupCollection::Checkpoint) if we don't do it in the WriteToDisk path.

src/include/duckdb/storage/table/row_group.hpp
```diff
@@ -241,17 +249,24 @@ class RowGroup : public SegmentBase<RowGroup> {
...
 	void UnloadColumn(storage_t c);
+	bool HasUnchangedColumns() const;
+	static shared_ptr<ColumnData> CheckpointColumn(const RowGroup &row_group, idx_t column_idx, RowGroupWriteInfo &info,
+	                                               RowGroupWriteData &write_data);
 
 	bool HasUnloadedDeletes() const;
+	unique_ptr<RowGroup> CreateNewRowGroupCopy(RowGroupCollection &new_collection, idx_t new_column_count);
 
 private:
+	friend class RowGroupCollection;
```
Mytherin: Can we avoid this?

src/include/duckdb/storage/data_pointer.hpp
Mytherin: Sure, makes sense

Mytherin (commented):
Thanks - looks good now - one more minor comment

src/include/duckdb/storage/table/row_group.hpp
ywelsch: Fixed in [52a1313](https://github.com/duckdb/duckdb/pull/22333/commits/52a1313a762ada0223fcf205fa73479054cf0761)

Mytherin:
Thanks - changeset looks good to me, could you just merge with main?

ywelsch:
> Thanks - changeset looks good to me, could you just merge with main?

done, the remaining failure looks to be an infrastructure issue, unrelated to the changes in this PR (CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.)

Mytherin:
Thanks!

PR #22335, opened 2026-04-28
Add OPTIONS to HTTPUtil and httpfs

### LINKED ISSUE #21533: Add OPTIONS to HTTPUtil RequestType enum
Hi DuckDB team,

I noticed that the `RequestType` enum in `src/include/duckdb/common/http_util.hpp` currently supports `GET`, `PUT`, `HEAD`, `DELETE`, and `POST`, but not `OPTIONS`.

## Why OPTIONS?

`OPTIONS` is the standard HTTP method for capability discovery and content negotiation. It's commonly used by servers to advertise supported features via response headers — for example, CORS preflight already relies on it. Extensions that interact with HTTP APIs sometimes need to discover server capabilities before deciding how to structure requests (e.g., whether large payloads should be uploaded via a separate URL).

Right now the workaround is to use `HEAD` and parse headers from the response, which works but isn't semantically correct.

## What would the change look like?

Based on the existing pattern, it should be pretty minimal:

1. Add `OPTIONS_REQUEST` to the `RequestType` enum
2. Add an `OptionsRequestInfo` struct (similar to `HeadRequestInfo` — no body, just URL + headers)
3. Handle it in the HTTP client implementation (httplib dispatch)

Happy to submit a PR if that would be helpful. Just wanted to check if this seems reasonable first.

Thanks!

carlopi:
I have also a CORSExplorer demo that actually stresses these codepaths by testing whether various backends require CORS or not, but that can't land as is, and was mostly handy to verify the base exposed primitives where working.

Fixes https://github.com/duckdb/duckdb/issues/21533

### DIFF
--- a/.github/patches/extensions/httpfs/0003-add-options.patch
+++ b/.github/patches/extensions/httpfs/0003-add-options.patch
@@ -1,0 +1,55 @@
+diff --git a/src/httpfs_curl_client.cpp b/src/httpfs_curl_client.cpp
+index 8a641ce..83e0d1c 100644
+--- a/src/httpfs_curl_client.cpp
++++ b/src/httpfs_curl_client.cpp
+@@ -398,6 +398,34 @@ public:
+ 		return TransformResponseCurl(res);
+ 	}
+ 
++	unique_ptr<HTTPResponse> Options(OptionsRequestInfo &info) override {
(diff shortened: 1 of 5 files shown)


Mytherin (approved):
Thanks! LGTM

PR #22339, opened 2026-04-28
RowGroup Operator metrics: sequentially scanned row groups + total row groups + cumulative counterparts

artjomPlaunov:
Add operator metrics for number of row groups sequentially scanned and total RowGroups available to scan for the operator instance (includes local storage rowgroups that could potentially be scanned). 

We could, for example, accumulate all the json profiling output to see the information per table scan operator:

```sql
# create a view that let's us query table scan operator nodes as rows of the view. 
statement ok
CREATE VIEW operator_metrics AS
WITH RECURSIVE ops AS (
    SELECT unnest(children) AS op FROM '__TEST_DIR__/profiling_output.json'
    UNION ALL
    SELECT unnest(op.children) FROM ops WHERE len(op.children) > 0
)
SELECT op.operator_type,
       op.operator_row_groups_scanned AS scanned,
       op.operator_row_groups_total AS total
FROM ops;

query III
SELECT * FROM operator_metrics WHERE operator_type = 'TABLE_SCAN' ORDER BY total;
----
TABLE_SCAN	1	1
TABLE_SCAN	1	2
```

In this case, one table scan of the query was on a single rowgroup of persistent data, and the second table scan was on a single rowgroup out of two available rowgroups (one persistent, one local).

The above is just showing that we are collecting these metrics on a per operator basis. The cumulative version will just add up the operator metrics per sequential table scan: 

```sql
statement ok
CREATE VIEW cumulative_metrics AS
SELECT cumulative_row_groups_scanned::BIGINT AS scanned,
       cumulative_total_row_groups_to_scan::BIGINT AS total
FROM '__TEST_DIR__/profiling_output.json';

query II
SELECT * FROM cumulative_metrics;
----
2	3
```

So here we scanned two rowgroups out of a total of 3 rowgroups that were available to scan (even if the 3 is counting duplicates, I think this makes the most sense if we want to view this as some cumulative fraction, I guess this can be useful to view pruning effectiveness?)

There is also a limitation with LIMIT for both these new metrics and the existing rows_scanned metric, where it will report 0 rows scanned because a LIMIT prevents it from getting flushed. Will fix this in a subsequent PR just to see if the idea here for row group profiling makes sense first. I added a test case that showcases this scenario for rows_scanned, currently returns 0 but should report the rows scanned even with a LIMIT; same for row groups.


taniabogatsch (commented):
Let's also add 
- `CUMULATIVE_ROW_GROUPS_SCANNED`
- `CUMULATIVE_TOTAL_ROW_GROUPS_TO_SCAN` (how many could have been scanned - used for pruning observability).

src/include/duckdb/function/table_function.hpp
```diff
@@ -436,8 +439,12 @@ class TableFunction : public SimpleNamedParameterFunction { // NOLINT: work-arou
 	//! (Optional) cardinality function
 	//! Returns the expected cardinality of this scan
 	table_function_cardinality_t cardinality;
-	//! (Optional) returns the number of rows that have benn scanned
+	//! (Optional) returns the number of rows scanned by this table scan operator
 	table_function_rows_scanned_t rows_scanned;
+	//! (Optional) returns the number of row groups scanned by this table scan operator
+	table_function_row_groups_scanned_t row_groups_scanned;
+	//! (Optional) returns the total number of row groups available to this table scan operator
+	table_function_total_row_groups_to_scan_t total_row_groups_to_scan;
```
lnkuiper: Instead of adding more fields here would it make sense to add a single function that returns metrics? Then we don't have to modify `TableFunction` each time we want to add a metric.

Maybe it can return `unordered_map<MetricType, Value>`?

src/include/duckdb/main/query_profiler.hpp
```diff
@@ -50,6 +50,9 @@ struct OperatorInformation {
 	idx_t system_peak_buffer_manager_memory = 0;
 	idx_t system_peak_temp_directory_size = 0;
 	idx_t rows_scanned = 0;
+	idx_t row_groups_scanned = 0;
+	idx_t total_row_groups_to_scan = 0;
+	bool has_total_row_groups_to_scan = false;
```
lnkuiper: Instead of an `idx_t` and `bool` for `total_row_groups_to_scan`, could we do `optional_idx`?

src/storage/data_table.cpp
```diff
@@ -1847,6 +1847,15 @@ idx_t DataTable::GetTotalRows() const {
...
 
+idx_t DataTable::GetRowGroupCount() const {
+	return row_groups->GetRowGroupCount();
+}
+
+idx_t DataTable::GetRowGroupCountWithLocalStorage(ClientContext &context) {
+	auto &local_storage = LocalStorage::Get(context, db);
+	return GetRowGroupCount() + local_storage.AddedRowGroups(*this);
+}
+
```
lnkuiper, line 1862: Are both of these functions needed? I think only `GetRowGroupCountWithLocalStorage` is called from outside, maybe it's best to remove the other function?

lnkuiper (commented):
Thanks for the changes! I have two last nitpicks and then this is ready to go

artjomPlaunov:
I need to patch the lance extension

artjomPlaunov:
update: due to the upcoming patch release, it is not ideal to change any existing table_function callbacks. keeping the old rows_scanned, but still introducing a new get_metrics callback. PhysicalTableScan::GetMetrics checks for both (so extensions using rows_scanned still work), and we will deprecate rows_scanned on main.

lnkuiper:
Thanks!

PR #22340, opened 2026-04-28
Add small-build-side dictionary emission for hash joins

ahmetkarapinar:
<h1>Add small-build-side dictionary emission for hash joins</h1>
<h2>Motivation</h2>
<p>In a regular hash join, every matched probe row copies build-side payload values out of row-format storage into the output chunk via <code>GatherResult</code>. When the build side is small but the probe side is large, the <strong>same build-side payload values get copied thousands of times.</strong></p>
<p>Perfect hash join handles a narrow case of this by emitting dictionary vectors instead of gathered values, <strong>but only for integer keys in a small dense range with no duplicates, and only for inner joins.</strong></p>
<p>This PR materializes small build sides into dictionary arrays, embeds the dictionary index into each row's <code>NEXT_PTR</code> field, and emits matched rows as dictionary vectors during probing.</p>
<h2>Approach</h2>
<p>When the build side is small and the probe side is large enough that the same build-side values would otherwise be copied many times, we apply the same dictionary-vector mechanism perfect hash join uses. After <code>Finalize()</code>, pre-materialize each RHS output column into a reusable <code>VectorChildBuffer</code> (same <code>DictionaryVector::CreateReusableDictionary</code>, same stable <code>dictionary_id</code>). Overwrite each build row's <code>NEXT_PTR</code> field with a <code>uint32_t</code> index into those arrays. At probe time, <code>EmitDictVectors</code> loads the index from each matched row and emits one <code>Vector::Dictionary(dict_arrays[col], build_sel_vec, count)</code> per build-side column — zero data copies.</p>
<p>The implementation lives in a single function — <code>BuildDictionaryArrays</code> — invoked from <code>HashJoinFinalizeEvent::FinishEvent</code>. By the time <code>FinishEvent</code> fires, all parallel finalize tasks have completed, <code>chains_longer_than_one</code> is final, every <code>NEXT_PTR</code> value is in its final state, and the <code>TupleDataCollection</code> is fully pinned — so a single pass over the rows can gather the dict arrays, snapshot chain pointers (when needed), and overwrite <code>NEXT_PTR</code>.</p>
<p>If the build side has no duplicate keys, <code>NEXT_PTR</code> is never read after finalize, so overwriting it is safe with zero auxiliary cost. If duplicate keys exist (<code>chains_longer_than_one = true</code>), <code>BuildDictionaryArrays</code> first saves the original <code>NEXT_PTR</code> values into an <code>aux_next_ptrs</code> array, and chain-walking sites read through a templated <code>GetNextPointer&lt;USE_DICT_EMISSION&gt;</code> that resolves the indirection. The runtime flag is lifted to a compile-time <code>bool</code> template parameter at the two dispatch boundaries (<code>AdvancePointers</code>, <code>NextRightSemiOrAntiJoin</code>), so the <code>use_dict_emission == false</code> specialization compiles byte-for-byte to the existing two-instruction <code>LoadPointer(ptr + pointer_offset)</code> chain-walk loop on the main branch. Non-activated joins are unaffected.</p>
<h2>Activation conditions</h2>
<p>All five must hold in <code>JoinHashTable::CanUseDictionaryEmission</code> (called from <code>HashJoinFinalizeEvent::FinishEvent</code>); any failure falls back silently to <code>GatherResult</code>.</p>

Condition | Why
-- | --
Not external | Cached row pointers don't survive partition rebuild
build_payload_bytes <= 8 MiB (MAX_BUILD_PAYLOAD_BYTES) | Caps the actual dict_arrays allocation; 8 MiB keeps the arrays in L3
probe_cardinality >= 262,144 (MIN_PROBE_ROWS) | Probe must be large enough to amortize the one-time build cost
probe / build >= 100 (PROBE_BUILD_RATIO) | Fan-out must be high enough for the per-row savings to pay off
Not SINGLE join | NextSingleJoin writes NULLs via FlatVector::SetNull, needs a flat vector

<p>Plus a nested-type bailout — STRUCT, LIST, and ARRAY are rejected because <code>Vector::Dictionary()</code> asserts non-STRUCT and <code>GetTypeIdSize</code> doesn't meaningfully measure nested-type cost. The thresholds live as <code>static constexpr</code> members of a small <code>DictionaryEmissionActivation</code> struct in <code>join_hashtable.hpp</code>. Mutual exclusion with perfect hash join is structural: <code>ScheduleFinalize</code> is only called from the <code>if (!use_perfect_hash)</code> branch, so <code>FinishEvent</code> never runs on the perfect hash join path.</p>
<p>Together these gates cover a broader set of joins than perfect hash join: any key type (not just integer), duplicate keys, and any join type that isn't SINGLE.</p>
<h2>Downstream impact</h2>
<p>The per-join savings (avoiding the per-row payload copies that <code>GatherResult</code> performs) are real but limited. The larger win is that dictionary vectors <strong>propagate</strong>, and downstream operators that recognize them can avoid re-doing work for every row that shares a dictionary entry:</p>
<ul>
<li><strong>Hash aggregate's <code>TryAddDictionaryGroups</code></strong> looks up each distinct dictionary value in the aggregate hash table once, then caches the resulting group pointer keyed by code. For every subsequent row referencing the same code, it skips the hash and key-compare — one array lookup into the cached addresses.</li>
<li>The <strong>projection operator</strong> evaluates expressions over the dictionary's <code>VectorChildBuffer</code>, computing each expression once per distinct entry rather than once per output row. Compressed Materialization, which goes through projection, benefits from this directly.</li>
</ul>
<h2>Showcase</h2>
<pre><code class="language-sql">SELECT n.n_name, COUNT(*) AS cnt
FROM nation n
JOIN customer c ON hash(n.n_nationkey) = hash(c.c_nationkey)
GROUP BY n.n_name;
</code></pre>
<p>This query is used to showcase the benefits of the optimization. The <code>hash()</code> calls on both sides bypass perfect hash join activation (in order to represent a wide key range) — purely for testing. The same end-to-end win applies whenever perfect hash join cannot be used: a string key, a LEFT JOIN, a wide range, etc.</p>
<p>With the optimization enabled, the query runs <strong>~2× faster</strong> at TPC-H SF=100 (~50% less wall time), and the speedup is consistent across scale factors 1, 10, 20, 50, and 100. The benefit comes from three places along the pipeline:</p>
<ol>
<li><strong>Hash join itself is ~1.09× faster</strong> (8.4% less time) — <code>EmitDictVectors</code> skips the per-row payload copies that <code>GatherResult</code> performs.</li>
<li><strong>Compressed Materialization is ~6.25× faster</strong> (84% less time) — its projection runs over the 25-entry dictionary instead of every matched customer row.</li>
<li><strong>The single-column hash aggregate is ~4.5× faster</strong> (78% less time) — <code>TryAddDictionaryGroups</code> looks up each of the 25 <code>n_name</code> values in the aggregate hash table once, caches the resulting group pointers, and reuses them for every subsequent row sharing the same dictionary code. The per-row hash and key-compare are skipped entirely.</li>
</ol>
<p>The flame graph below shows profiling results of the optimization compared to the main branch.</p>
image
<h2>TPC-H Results</h2>
<p>Full suite run at SF=1, 10, 20, 50, 100, comparing this branch against <code>main</code>.</p>
<p>Hardware: Apple M4 Pro, 24 GB RAM.</p>
<p>Single-threaded setup, 10 runs per query, minimum reported.</p>
<p>The optimization activates on the following queries per scale factor:</p>
<ul>
<li><strong>SF=1</strong> → Q5, Q17, Q18, Q19, Q20</li>
<li><strong>SF=10</strong> → Q2, Q7, Q8, Q9, Q10, Q17, Q18, Q19, Q20</li>
<li><strong>SF=20</strong> → Q2, Q7, Q8, Q9, Q10, Q17, Q18, Q19, Q20</li>
<li><strong>SF=50</strong> → Q2, Q7, Q8, Q9, Q10, Q15, Q17, Q18, Q19, Q20</li>
<li><strong>SF=100</strong> → Q2, Q7, Q8, Q10, Q15, Q17, Q18, Q19, Q20</li>
</ul>
<p><strong>Zero regressions</strong> across all activated queries at any scale factor. There are no reproducible per-query wins on the activated TPC-H queries because operators that don't recognize dictionary vectors materialize them back to a flat vector before downstream dictionary-aware operators can exploit them — so the end-to-end pipeline win seen on the showcase query does not translate to TPC-H today.</p>
<p>This is itself a meaningful result: even when the emitted dictionaries are flattened almost immediately, enabling the optimization does not slow the query down. The activation gates are tight enough that the build-time cost is fully amortized, and we can ship the optimization safely.</p>
<h2>Future direction</h2>
<ul>
<li>Today only a handful of operators recognize dictionary vectors. As more dictionary-aware operators land in DuckDB over time, the value of emitting dictionary-encoded build-side columns from hash joins will compound.</li>
<li>A natural next step is to make these dictionaries pipeline-global: when a column arrives at a hash-join sink already dictionary-encoded, the build side stores only the 4-byte index in the row instead of re-flattening the value, so the encoding survives across a chain of joins. Our plan is to move directly to this idea after this PR, as the next part of my master thesis project.</li>
</ul>
<p>@lnkuiper — implementation of the project we discussed last week, ready for review.</p>
<p>Thank you for taking the time to review this! I'm looking forward to your feedback, and I'd be happy to discuss any part of the design.</p>

### DIFF
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -1072,23 +1072,31 @@ idx_t ScanStructure::ScanInnerJoin(DataChunk &keys, DataChunk &probe_data, Selec
 	}
 }
 
+template <bool USE_DICT_EMISSION>
+static void AdvancePointersLoop(JoinHashTable &ht, Vector &pointers, SelectionVector &out_sel,
+                                const SelectionVector &sel, const idx_t sel_count, idx_t &out_count) {
+	idx_t new_count = 0;
+	auto ptrs = FlatVector::GetDataMutable<data_ptr_t>(pointers);
+	for (idx_t i = 0; i < sel_count; i++) {
+		auto idx = sel.get_index(i);
+		ptrs[idx] = ht.GetNextPointer<USE_DICT_EMISSION>(ptrs[idx]);
+		if (ptrs[idx]) {
+			out_sel.set_index(new_count++, idx);
+		}
+	}
+	out_count = new_count;
+}
+
 void ScanStructure::AdvancePointers(const SelectionVector &sel, const idx_t sel_count) {
 	if (!ht.chains_longer_than_one) {
 		this->count = 0;
 		return;
 	}
-
-	// now for all the pointers, we move on to the next set of pointers
-	idx_t new_count = 0;
-	auto ptrs = FlatVector::GetDataMutable<data_ptr_t>(this->pointers);
-	for (idx_t i = 0; i < sel_count; i++) {
-		auto idx = sel.get_index(i);
-		ptrs[idx] = LoadPointer(ptrs[idx] + ht.pointer_offset);
-		if (ptrs[idx]) {
-			this->sel_vector.set_index(new_count++, idx);
-		}
+	if (ht.use_dict_emission) {
+		AdvancePointersLoop<true>(ht, pointers, sel_vector, sel, sel_count, count);
+	} else {
+		AdvancePointersLoop<false>(ht, pointers, sel_vector, sel, sel_count, count);
 	}
-	this->count = new_count;
 }
 
 void ScanStructure::AdvancePointers() {
@@ -1170,12 +1178,18 @@ void ScanStructure::NextInnerJoin(DataChunk &keys, DataChunk &probe_data, DataCh
 					result.SetCardinality(result_count);
 
 					// on the RHS, we need to fetch the data from the hash table
-					for (idx_t i = 0; i < ht.output_columns.size(); i++) {
-						auto &vector = result.data[ht.lhs_output_in_probe.size() + i];
-						const auto output_col_idx = ht.output_columns[i];
-						D_ASSERT(vector.GetType() == ht.layout_ptr->GetTypes()[output_col_idx]);
-						GatherResult(vector, chain_match_sel_vector, result_count, output_col_idx);
-						FlatVector::SetSize(vector, count_t(result_count));
+					if (ht.use_dict_emission) {
+						auto ptrs = FlatVector::GetData<data_ptr_t>(pointers);
+						ht.EmitDictVectors(ptrs, chain_match_sel_vector, result_count, result,
+						                   ht.lhs_output_in_probe.size());
+					} else {
+						for (idx_t i = 0; i < ht.output_columns.size(); i++) {
+							auto &vector = result.data[ht.lhs_output_in_probe.size() + i];
+							const auto output_col_idx = ht.output_columns[i];
+							D_ASSERT(vector.GetType() == ht.layout_ptr->GetTypes()[output_col_idx]);
+							GatherResult(vector, chain_match_sel_vector, result_count, output_col_idx);
+							FlatVector::SetSize(vector, count_t(result_count));
+						}
 					}
 
 					AdvancePointers();
@@ -1199,12 +1213,17 @@ void ScanStructure::NextInnerJoin(DataChunk &keys, DataChunk &probe_data, DataCh
 		result.SetCardinality(base_count);
 
 		// 2) gather RHS vectors
-		for (idx_t i = 0; i < ht.output_columns.size(); i++) {
-			auto &vector = result.data[ht.lhs_output_in_probe.size() + i];
-			const auto output_col_idx = ht.output_columns[i];
-			D_ASSERT(vector.GetType() == ht.layout_ptr->GetTypes()[output_col_idx]);
-			GatherResult(vector, base_count, output_col_idx);
-			FlatVector::SetSize(vector, count_t(base_count));
+		if (ht.use_dict_emission) {
+			auto rhs_ptrs = FlatVector::GetData<data_ptr_t>(rhs_pointers);
+			ht.EmitDictVectors(rhs_ptrs, base_count, result, ht.lhs_output_in_probe.size());
+		} else {
+			for (idx_t i = 0; i < ht.output_columns.size(); i++) {
+				auto &vector = result.data[ht.lhs_output_in_probe.size() + i];
+				const auto output_col_idx = ht.output_columns[i];
+				D_ASSERT(vector.GetType() == ht.layout_ptr->GetTypes()[output_col_idx]);
+				GatherResult(vector, base_count, output_col_idx);
+				FlatVector::SetSize(vector, count_t(base_count));
+			}
 		}
 	}
 }
@@ -1277,6 +1296,32 @@ void ScanStructure::NextAntiJoin(DataChunk &keys, DataChunk &probe_data, DataChu
 	finished = true;
 }
 
+template <bool USE_DICT_EMISSION>
+static void MarkChainsAsFoundLoop(JoinHashTable &ht, data_ptr_t ptrs[], const SelectionVector &chain_match_sel_vector,
+                                  const idx_t result_count, const data_ptr_t dead_end_ptr) {
+	for (idx_t i = 0; i < result_count; i++) {
+		const auto idx = chain_match_sel_vector.get_index(i);
+		auto &ptr = ptrs[idx];
+		if (Load<bool>(ptr + ht.tuple_size)) { // Early out: chain has been fully marked as found before
+			ptr = dead_end_ptr;
+			continue;
+		}
+
+		// Fully mark chain as found
+		while (true) {
+			// NOTE: threadsan reports this as a data race because this can be set concurrently by separate
+			// threads Technically it is, but it does not matter, since the only value that can be written is
+			// "true"
+			Store<bool>(true, ptr + ht.tuple_size);
+			auto next_ptr = ht.GetNextPointer<USE_DICT_EMISSION>(ptr);
+			if (!next_ptr) {
+				break;
+			}
+			ptr = next_ptr;
+		}
+	}
+}
+
 void ScanStructure::NextRightSemiOrAntiJoin(DataChunk &keys, DataChunk &probe_data) {
 	const auto ptrs = FlatVector::GetDataMutable<data_ptr_t>(pointers);
 	while (!PointersExhausted()) {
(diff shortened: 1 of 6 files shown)


src/include/duckdb/execution/join_hashtable.hpp
```diff
@@ -318,6 +350,18 @@ class JoinHashTable {
 	//! Mapping from lhs_output_columns positions to lhs_probe_data positions
 	vector<idx_t> lhs_output_in_probe;
 
+	//! True once BuildDictionaryArrays has embedded dictionary indices into NEXT_PTR
+	bool use_dict_emission = false;
+	//! Pre-materialized columnar data, one entry per RHS output column
+	vector<buffer_ptr<DictionaryEntry>> dict_arrays;
+	//! Saved NEXT_PTR values, indexed by dict index; only populated when chains_longer_than_one
+	vector<data_ptr_t> aux_next_ptrs;
```
lnkuiper: Could this be an `AllocatedData` that is allocated through the `BufferAllocator` instead? Then we're keeping track of the allocation and you can just access the array directly (instead of using this _safe_ vector)

src/include/duckdb/execution/join_hashtable.hpp
```diff
@@ -46,6 +47,16 @@ struct JoinHTScanState {
...
 
+//! Threshold constants gating small-build-side dictionary emission
+struct DictionaryEmissionActivation {
+	//! Caps the dict_arrays allocation; 8 MiB keeps them in L3 and bounds aux_next_ptrs
+	static constexpr idx_t MAX_BUILD_PAYLOAD_BYTES = 8ULL * 1024ULL * 1024ULL;
+	//! Below this, dictionary-emission setup is not amortized by probe-side savings
+	static constexpr idx_t MIN_PROBE_ROWS = 262144;
+	//! Below this ratio, per-row savings are unlikely to amortize the setup cost
+	static constexpr idx_t PROBE_BUILD_RATIO = 100;
+};
```
lnkuiper: This `struct` can be moved to the `.cpp` file.

src/execution/join_hashtable.cpp
```diff
@@ -1986,4 +2021,126 @@ void ProbeSpill::PrepareNextProbe() {
...
+	SelectionVector build_sel_vec(count);
+	for (idx_t i = 0; i < count; i++) {
+		const auto idx = ptr_sel.get_index(i);
+		build_sel_vec.set_index(i, Load<uint32_t>(row_ptrs[idx] + pointer_offset));
+	}
+	for (idx_t col_idx = 0; col_idx < output_columns.size(); col_idx++) {
+		auto &vector = result.data[rhs_col_offset + col_idx];
+		vector.Dictionary(dict_arrays[col_idx], build_sel_vec, count);
+	}
+}
```
lnkuiper, line 2091: I know one has a selection vector and the other hasn't, but can both `JoinHashTable::EmitDictVectors` be folded into one? Maybe using a `FlatVector::IncrementalSelectionVector` if there is no sel? You could template on `SelectionVector::IsSet` if necessary

src/execution/join_hashtable.cpp
```diff
@@ -1986,4 +2021,126 @@ void ProbeSpill::PrepareNextProbe() {
...
+		aux_next_ptrs.resize(build_count);
+		for (idx_t i = 0; i < build_count; i++) {
+			aux_next_ptrs[i] = LoadPointer(row_ptrs[i] + pointer_offset);
+		}
+	}
+
+	// embed the dict index into NEXT_PTR; GetNextPointer now resolves chains via aux_next_ptrs
+	for (idx_t i = 0; i < build_count; i++) {
+		Store<uint32_t>(static_cast<uint32_t>(i), row_ptrs[i] + pointer_offset);
+	}
```
lnkuiper, line 2133: Can you fold these last two loops into one? Saves an additional iteration over the tuples. I think that's worth more than a branch in the loop (the branch is const and can be predicted perfectly)

src/execution/join_hashtable.cpp
```diff
@@ -1986,4 +2021,126 @@ void ProbeSpill::PrepareNextProbe() {
 	consumer->InitializeScan();
 }
 
+//===--------------------------------------------------------------------===//
+// Small Build Side Dictionary Emission
+//===--------------------------------------------------------------------===//
+idx_t JoinHashTable::ComputeBuildPayloadBytes(const vector<LogicalType> &rhs_output_types) const {
+	idx_t payload_bytes_per_row = 0;
```
lnkuiper, line 2031: Can you add a comment here explaining why this doesn't count the size of variable-size strings? It's because they are already stored in the `TupleDataCollection`, but I think that should be explained here :)

src/execution/join_hashtable.cpp
```diff
@@ -1197,11 +1211,16 @@ void ScanStructure::NextInnerJoin(DataChunk &keys, DataChunk &probe_data, DataCh
...
+		if (ht.use_dict_emission) {
+			auto rhs_ptrs = FlatVector::GetData<data_ptr_t>(rhs_pointers);
+			ht.EmitDictVectors(rhs_ptrs, base_count, result, ht.lhs_output_in_probe.size());
+		} else {
+			for (idx_t i = 0; i < ht.output_columns.size(); i++) {
+				auto &vector = result.data[ht.lhs_output_in_probe.size() + i];
+				const auto output_col_idx = ht.output_columns[i];
+				D_ASSERT(vector.GetType() == ht.layout_ptr->GetTypes()[output_col_idx]);
+				GatherResult(vector, base_count, output_col_idx);
+			}
```
lnkuiper: I know the code here was already "kinda duplicated" but it's become a bit more obvious now. There are two "if use dict emission do X else do Y" in this function now. Do you think it's possible to separate this out into a helper function?

I can see very similar functionality again in `NextRightSemiOrAntiJoin`

src/execution/join_hashtable.cpp
```diff
@@ -1274,6 +1293,32 @@ void ScanStructure::NextAntiJoin(DataChunk &keys, DataChunk &probe_data, DataChu
...
+			// "true"
+			Store<bool>(true, ptr + ht.tuple_size);
+			auto next_ptr = ht.GetNextPointer<USE_DICT_EMISSION>(ptr);
+			if (!next_ptr) {
+				break;
+			}
+			ptr = next_ptr;
+		}
+	}
+}
```
lnkuiper, line 1318: Thanks for separating this logic out into a helper function, looks great :)

lnkuiper (commented):
Thanks for the PR! I think it looks great overall. I've left some comments below.

Can you add a benchmark file with your query?
```sql
SELECT n.n_name, COUNT(*) AS cnt
FROM nation n
JOIN customer c ON hash(n.n_nationkey) = hash(c.c_nationkey)
GROUP BY n.n_name;
```
That will allow us to keep track of performance over time and try not to regress.

Just pick the lowest SF that shows the performance benefit. Maybe it already shows some performance benefit at SF1 if you do `SET threads=1;`, then the benchmark file won't take too long to run due to the data generation. If not, you can use SF3/10

ahmetkarapinar:
@lnkuiper Thanks for the review! I've pushed the changes addressing all your comments — let me know if anything still needs attention.

Also rebased onto current main to resolve a conflict in join_hashtable.cpp.

src/execution/join_hashtable.cpp
```diff
@@ -1656,12 +1680,17 @@ void JoinHashTable::ScanFullOuter(JoinHTScanState &state, Vector &addresses, Dat
...
+		auto key_locs = FlatVector::GetData<data_ptr_t>(addresses);
+		EmitDictVectors(key_locs, sel_vector, found_entries, result, left_column_count);
+	} else {
+		for (idx_t i = 0; i < output_columns.size(); i++) {
+			auto &vector = result.data[left_column_count + i];
+			const auto output_col_idx = output_columns[i];
+			D_ASSERT(vector.GetType() == layout_ptr->GetTypes()[output_col_idx]);
+			data_collection->Gather(addresses, sel_vector, found_entries, output_col_idx, vector, sel_vector, nullptr);
+			FlatVector::SetSize(vector, count_t(found_entries));
+		}
```
lnkuiper: Is this another candidate for `GatherRHS`?

lnkuiper (commented):
Thanks for the changes! Looks great now. I have one last (small) comment and then this looks good to go from my side

ahmetkarapinar:
Done! Thanks for all the careful review @lnkuiper !

lnkuiper:
Thanks for the PR and all the changes! Let's run CI :)

lnkuiper:
Looks pretty good so far, but one CI run is failing. Not sure if related

PR #22342, opened 2026-04-28
Introduce `Bound(Scalar/Aggregate/Window)` functions

Maxxen:
This PR introduces a trio of new classes that represent _bound_ functions. 

These are as of now just empty subclasses of their corresponding function type, but still explicitly separates the mostly static lifetime that functions have as part of the catalog from their mutable copies created at bind-time, right after DuckDB has made an initial attempt at resolving their types and arguments. The goal is to eventually make the regular Scalar/Aggregate/Window function structs fully immutable/non-copyable and enriched with additional static metadata such as e.g. parameter names, documentation or type constraints.

A side effect of this is that we are now stricter (and will be even more stricter in the future) about when a function requires to be "bound". In particular we generally want a `BoundFunctionExpression`'s to ensure that the function object has had its arguments properly bound. To make this easier when we construct expressions internally (e.g. as part of the optimizer), I've changed the `::Bind` helper functions on the regular function structs to now return a `unique_ptr<Expression>` directly and require a client context. To keep this PR a bit smaller I've left some places left alone though and just constructed the `BoundScalarFunction` immediately from the `ScalarFunction` as the bound variants are implicitly constructible, but this will change in the future. We may need to add something like a `ScalarFunction::DefaultBind(vector<unique_ptr<Expression>> &args) -> BoundScalarFunction` for cases where we don't have a context but  would still like to make a best-effort at reordering or casting function arguments.


Maxxen:
I think this CI failure is also unrelated?

Mytherin:
Looks like indeed - thanks!

hawkfish (approved):
Nice!

PR #22344, opened 2026-04-28
Node Handle Scoping fix 

artjomPlaunov:
https://github.com/duckdblabs/duckdb-internal/issues/8907 (and probably any other bugs with the same stacktrace). 

We were grabbing a node handle for the leaf node that we were deleting the rowid in. If there was only one rowid remaining, then we inline the prefix chain, replacing the parent pointer of the prefix chain with the remaining rowid. To do this, we free the tree starting from the head of the prefix chain, and overwrite the rowid into the pointer slot which was the head of the prefix chain. 

We were freeing the tree while the leaf rowid node was still scoped by the handle, hence there was a reader on the buffer and we got an error when freeing. The fix is to just end the node handle scope before the prefix inlining check.


taniabogatsch (approved):

Mytherin:
Thanks!

PR #22346, opened 2026-04-28
Flat struct_map list to get the correct size

pdet:
Found this while fixing ducklake to work with duckdb `main`.


Mytherin:
Thanks!

PR #21569, opened 2026-03-23
feat: Add Suspended Query Contexts: Interleaved Streaming + DML

rustyconover:
Today, executing any statement on a DuckDB connection destroys an in-progress streaming scan on that connection — even if both are inside the same explicit transaction:

```sql
BEGIN;

-- Start streaming rows from a large table
SELECT rowid, * FROM target WHERE NOT processed;

-- Fetch the first batch of 1,000 rows...
-- Now mark those rows as processed, in the same transaction:
UPDATE target SET processed = true WHERE rowid IN (1, 2, 3, ...);
-- The SELECT's executor is destroyed. The scan position is lost.
-- There is no way to fetch the remaining rows.

COMMIT;
```

A streaming scan is stateful — the executor holds its position across `Fetch()` calls. When a new statement starts on the same connection, DuckDB's `InitialCleanup()` tears down the active executor. The scan doesn't pause. It's gone.

This forces a choice: materialize the entire result set up front, or finish the scan before doing any other work in that transaction. For large tables, neither is acceptable.

The simple example above could be rewritten as a single `UPDATE`, but the real motivation is multi-step workflows inside a long-running scan's transaction. Consider:

1. **Scan** a source table, streaming batches of rows.
2. **INSERT** each batch into a staging table.
3. **UPDATE** a target table by joining against the staged rows.
4. **DELETE** the processed rows from the source.
5. Repeat from step 1 with the next batch — the scan is still yielding rows.

Each step needs to see the writes from the previous one, which requires a single transaction. And the scan must survive across all of them, which is impossible today. This is the natural shape of any multi-step ETL, sync, or migration workflow.

The workarounds are all unsatisfying:

- **Materialize the full result set** - defeats the purpose of streaming; may not fit in memory.
- **LIMIT/OFFSET chunking** - slow, incompatible with `UPDATE ... RETURNING`, and offsets shift as rows are mutated underneath.
- **Separate connections** - separate transactions, so the INSERT in step 2 isn't visible to the UPDATE in step 3.
- **Implement cursors** - Implement something like cursors from Postgres.

### Solution

This PR enables interleaved query streaming and DML within a single explicit transaction by suspending and resuming query executor state on the same connection.

### How It Works

DuckDB enforces one active query per connection via `ActiveQueryContext`, which holds the executor and pipeline state. When a new query starts, `InitialCleanup()` destroys the active context.

[...]

PR #22307, opened 2026-04-27
Add SET TRANSACTION SNAPSHOT for sharing transactions across connections

rustyconover:
Hey @Mytherin — picking up the design you proposed in #21569: sharing an in-progress transaction across connections, with both the visibility snapshot **and** uncommitted writes shared. Posting as a draft for design review on the semantics rather than an immediate merge ask — happy to flex on anything you want different.

The motivating use case is the original RPC pattern from #21569: a server-side DuckDB that needs to apply an `UPDATE` whose predicate must run client-side. The client connection scans rowids on a worker connection and the matching rows are then updated by the coordinator connection — all inside one transaction.

## SQL surface

```sql
-- owner:
BEGIN;
SELECT duckdb_export_transaction();      -- '6f3a1a8c-…'

-- participant:
BEGIN;
SET TRANSACTION SNAPSHOT '6f3a1a8c-…';
SELECT count(*) FROM ...;                -- sees owner's uncommitted rows
INSERT INTO ...;                         -- writes into the shared transaction
COMMIT;                                  -- detaches; does not finalize

-- owner:
COMMIT;                                  -- waits for participants, then commits everyone
```

PEG grammar gets one new rule (`SetTransactionSnapshot <- 'SET' Transaction 'SNAPSHOT' StringLiteral`); `duckdb_export_transaction()` is a new scalar function.

## Owner-authoritative semantics

Once writes are shared, there's one logical transaction with N attached connections. Postgres's "every session commits independently" model doesn't translate (Postgres only shares the snapshot, not the transaction). Decisions:

| Action                                  | Effect                                                                                           |
|------------------------------------------|--------------------------------------------------------------------------------------------------|
| Participant `COMMIT` / `ROLLBACK`        | Detach only. Decrements participant count. Does not finalize anything.                            |
| Owner `COMMIT`                           | Marks the meta as committing (registry rejects new imports), then blocks for participants to detach, then commits. Interruptible via `Connection::Interrupt()` / Ctrl-C. |
| Owner `ROLLBACK`                         | Sets a force-detach signal (participants' next non-control statement errors once and the connection auto-detaches), blocks for participants to detach, then aborts. Symmetric with COMMIT — same wait pattern keeps per-database `DuckTransaction` objects alive while participants reference them. |
| Owner connection close                   | Implicit ROLLBACK.                                                                                |

Auto-detach on the participant side means a client doesn't need to issue a manual `ROLLBACK` after being force-detached — they see one error and the next query runs in fresh auto-commit context.

[...]

PR #22347, opened 2026-04-28
JOIN TRANSACTION: share an in-progress transaction across connections (v2 of #22307)

rustyconover:
Replaces #22307 with a cleaner design after Mytherin's review. Posting as a draft for design feedback before code review — the semantics changed from v1 in two ways worth discussing up front.

## Use case

Same as #22307 / #21569: a server-side embedding where one connection coordinates a transaction and other connections participate in it. The motivating pattern is an `UPDATE` whose predicate must run on a worker connection while the coordinator owns the writes — both inside one transaction.

## SQL surface

```sql
-- owner:
BEGIN;
INSERT INTO t VALUES (1);
SELECT duckdb_share_transaction();   -- '6/main'

-- joiner:
BEGIN;
JOIN TRANSACTION '6/main';
SELECT count(*) FROM t;              -- sees owner's uncommitted rows
INSERT INTO t VALUES (2);            -- writes into the shared transaction
COMMIT;                              -- detaches; the last detacher finalizes

-- owner:
COMMIT;                              -- if joiner already detached, this is the finalize
```

The id format is `<connection_id>/<database_name>`. No process-wide registry — `ConnectionManager::FindByConnectionId` does the lookup. The id is pinned on the first `duckdb_share_transaction()` call within a transaction, so repeated calls return the same value.

## Design changes from v1

**Renamed away from `SET TRANSACTION SNAPSHOT`.** PG's `SET TRANSACTION SNAPSHOT` shares only the visibility snapshot, not the write set. v1 reused PG syntax for non-PG semantics. `JOIN TRANSACTION '<id>'` describes what's actually happening — you join an existing transaction as a participant.

**Symmetric finalize, no force-detach, no cv-wait.** Every reference (owner or joiner) detaches the same way. The last detacher commits or rolls back at the storage layer. Any participant's rollback dooms the transaction; subsequent COMMITs from other participants fail eagerly with a clear error. v1's owner-blocks-on-participants pattern, the force-detach signal, and `transaction_snapshot_registry.{hpp,cpp}` are all gone — about 80 lines of state machinery removed.

**Per-`DuckTransaction` statement lock instead of per-`MetaTransaction`.** Queries on unrelated databases inside the same `MetaTransaction` no longer block each other.

## Decisions I want feedback on before review

1. **Owner COMMIT does not block on joiners.** v1 made `owner.COMMIT()` block until joiners detached — matching the user-intuition that "when COMMIT returns, the work is durable." v2 instead lets the *last detacher* finalize, which simplifies the implementation but means owner's COMMIT can return before durability if joiners are still attached. Happy to restore blocking semantics if you'd prefer.

2. **Doom on rollback is detected lazily.** A joiner whose owner has rolled back doesn't find out until they try to COMMIT (or any later operation that hits the doom flag). v1's force-detach made this immediate. v2 trades latency for one fewer synchronization edge.

3. **Single-write-DB invariant** still applies: a joiner that has joined two DBs can write to at most one of them per `MetaTransaction`. Test locks this in (`single-write-DB invariant holds across joined DBs`).

## Concurrency

Six concurrency issues were identified and fixed during review:

- **Statement lock acquired unconditionally** on every opened `DuckTransaction` in `BeginQuery`, not only when `count > 1`. The 1→shared transition in v1 left the owner unguarded mid-query.
- **Atomic lookup-and-bump** via `MetaTransaction::TryClaimParticipant` under `owner_meta.lock` closes a TOCTOU window where an owner-side DETACH could destroy the `DuckTransaction` between lookup and `TryAddParticipant`.
- **Stable lock acquisition order** — `BeginQuery` sorts the per-`DuckTransaction` locks by pointer to prevent AB/BA deadlock when joiners attached multiple DBs in different orders.
- **`active_query` write race** — the foreign `DuckTransaction`'s statement lock is acquired before `ImportTransaction` writes `active_query`, and the guard is registered with the active query so it stays held for the rest of the statement.
- **Cross-connection torn read of `current_transaction`** — added a per-`TransactionContext` `meta_lock` taken on writers (BeginTransaction/Commit/Rollback/ClearTransaction), with a single foreign entry point `TryForeignClaimParticipant` that performs the lookup-and-bump under the lock. Foreign callers see either an active transaction or a clean error, never a half-moved unique_ptr.
- **`CancelParticipation` last-detacher leak** — if the owner detached concurrently between `TryAddParticipant` and `CancelParticipation`, the cancel landed at count=0 but did nothing. Now drives storage-layer commit/rollback when it ends up being the last detacher.

## Restrictions (unchanged from v1)

- Joiner must be inside an explicit transaction (`BEGIN`) and have done no per-database work yet — matches PG's "must be the first command of the transaction" rule.
- `ATTACH` inside a shared transaction is rejected for both owner and joiners. The shared transaction's database set is fixed at join time.
- `DETACH` is rejected for the *shared* database; unrelated DBs can still be detached.
- Transaction id is process-local; it expires when the owner's transaction finalizes.

## Tests

- `test/api/test_join_transaction.cpp` — 26 cross-connection cases including:
  - Joiner sees owner's uncommitted writes; joiner writes are visible to owner.
  - Doom-on-rollback in both directions, with eager COMMIT failure.
  - Owner connection close while shared dooms the transaction.
  - DETACH unrelated DB allowed; DETACH shared DB rejected.
  - Cross-database read by joiner.
  - Late join after owner finalized is rejected.
  - Concurrent joiner writes serialize correctly (200×200 stress).
  - Rapid interleaved owner+joiner queries.
  - Multi-DB JOIN with stable lock ordering (no AB/BA deadlock).
  - Multi-DB end-to-end + single-write-DB invariant.
  - `TransactionInfo` serialization round-trip (covers the renamed `transaction_id` field, slash-in-DB-name id, empty id).
  - Input validation (empty / >1024 bytes / control characters).
- `test/sql/transactions/shared/parse_join_transaction.test` — parser/binder paths plus the pinning assertion `duckdb_share_transaction() = duckdb_share_transaction()`.

\`[join_txn]\`, \`[transactions]\`, and \`[api]\` suites pass. \`make format-check\` passes. \`make generate-files\` clean.

## Tracked for follow-up

- `MetaTransaction::context` raw-reference cleanup (pre-existing, noted in #22307).
- `DUCKDB_GUARDED_BY` annotations on `MetaTransaction` mutex.
- User-facing docs in `duckdb-web` after this lands.

## Test plan

- [x] `[join_txn]`, `[transactions]`, `[api]` suites pass
- [x] `make format-check` passes
- [x] `TREAT_WARNINGS_AS_ERRORS=1 make release` clean
- [x] Concurrent-writers stress test
- [ ] User-facing docs (separate `duckdb-web` PR)

Closes #21569 once this is in shape.


Mytherin (commented):
Thanks - I think this is still doing a lot of very difficult to reason about locking (that is also wrong given the failing thread sanitizer runs). Can we not make the transaction a `shared_ptr` and use shared pointer semantics to implement the "commit / rollback if we are the last transaction" behavior? See maybe [this PR](https://github.com/duckdb/duckdb/pull/20548) for some inspiration.

I also don't get why the `MetaTransaction` is modified this heavily - it should not be involved at all, really. The `MetaTransaction` is a holder of several other transactions that occur in different catalogs - it has nothing to do with the underlying `DuckTransaction` that is being shared.

I appreciate the effort but at this rate I'm not entirely convinced AI is going to create a solid, maintainable implementation of this versus just a hacky one with sharp corners and edge cases all over the place.

rustyconover:
@Mytherin You might be right it is struggling with the locking mechanism, in this latest update, I've really forced it to simplify.

rustyconover:
@Mytherin you're right this is complicated, my head hurts a bit, but thats always what it feels like when I'm learning something new.

I spent time today trying to use a `shared_ptr<DuckTransaction>` approach but that really caused problems, the RAII description changes the order transactions finalize. This caused a lot of craziness in the tests.  #20548 - seemed to have an easier time of it, but transaction commit seems far more complicated with destructor order sensitivity.

I've addressed the ThreadSanitizer errors.  And all of the tests are passing now.

I'm confident this PR isn't perfect, it is closer than the two previous ones, all of the tests pass but thats not a measure if it should be merged. This is an entirely new area of DuckDB to me. Transactions and locking is not one I feel especially comfortable in yet. I'll admit I'm a bit out of my depth on this too.  You've done a pretty great job, with all of this so far.

I knew about `MetaTransaction`s before when I was working on the `events` extension but `DuckTransaction`s are new to me.  

I think the main complexity is that I'm trying to put in the ability to keep references across transactions, and it's tricky trying to know when I'm the last one dealing with a transaction. I know you don't want the sharing complexity on the MetaTransaction, but I haven't yet figured out the right way to do it because of `active_query`/`content_lock` interacting with my newly added `statement_lock`

I've created a new `statement_lock` on a DuckTransaction to handle the case where two or more connections share the same transaction, it makes sure that only one of those transactions can write at a time. The complexity we both don't like is having the `statement_lock` interact with the context_lock.

Can you suggest another approach that I could take to make this locking headache not so bad?

src/include/duckdb/main/client_context.hpp
```diff
@@ -341,8 +344,19 @@ class ClientContextLock {
 	~ClientContextLock() {
 	}
 
+	//! Temporarily release the context_lock. Used by BeginQueryInternal to acquire
+	//! per-DuckTransaction statement locks outside the context_lock critical section,
+	//! avoiding the M0(context_lock) -> M1(statement_lock) ordering that would otherwise
+	//! cycle against StreamQueryResult::FetchInternal's M1 -> M0 acquisition.
+	void Unlock() {
+		client_guard.unlock();
```
Mytherin, line 352: This is very hacky and a bad idea - this means another thread can now pop in and **start a different query on this connection** which will almost certainly cause crashes and other weird behavior.

test/api/test_delete_query_node.cpp
```diff
@@ -0,0 +1,98 @@
...
+using namespace duckdb;
+
+static unique_ptr<DeleteStatement> ParseDelete(const string &sql) {
+	Parser p;
+	p.ParseQuery(sql);
+	REQUIRE(p.statements.size() == 1);
+	REQUIRE(p.statements[0]->type == StatementType::DELETE_STATEMENT);
+	return unique_ptr<DeleteStatement>(static_cast<DeleteStatement *>(p.statements[0].release()));
+}
+
```
Mytherin, line 18: nit: these files were deleted and are re-added here

src/parser/peg/grammar/statements/transaction.gram
```diff
@@ -1,8 +1,9 @@
-TransactionStatement <- BeginTransaction / RollbackTransaction / CommitTransaction
+TransactionStatement <- JoinTransaction / BeginTransaction / RollbackTransaction / CommitTransaction
 
 BeginTransaction <- StartOrBegin Transaction? ReadOrWrite?
 RollbackTransaction <- AbortOrRollback Transaction?
 CommitTransaction <- CommitOrEnd Transaction?
+JoinTransaction <- 'JOIN' Transaction StringLiteral
```
Mytherin, line 6: Let's go back to the Postgres syntax - the behavior is still close enough that I don't think this warrants different syntax.

Mytherin (commented):
Thanks - I think this is all still just happening in the wrong place. If I had to do this I would implement this almost entirely in the `DuckTransactionManager`. As mentioned the meta transaction, client context lock, etc are all entirely unrelated to this. These are higher level constructs that exist to manage an individual connection. What we're trying to achieve here is sharing *a transaction to a specific DuckDB database*, which is managed by a specific `DuckTransactionManager`. We're explicitly trying to *avoid* messing with individual connection locks - that's the whole point of this change.

Maybe as a starting point you could create a shared transaction container in `DuckTransactionManager` that is also guarded by the `transaction_lock`. 

* When someone exports their transaction it gets stored in that transaction container with a counter and the `idx_t share_count` in the `DuckTransaction` gets set to 1. 
* When someone imports a transaction they grab a reference to that transaction and put it inside their `MetaTransaction`, effectively replacing the current `DuckTransaction`, and increment the share count.
* When someone commits or rolls back a transaction that has been shared we look at the "share count" on the transaction. If it is higher than 1 we decrement it, and only commit / rollback if the share count is now 0. If the share count is 0 we also remove it from the shared container. If it's a rollback we also note in the transaction that a rollback has been performed and, even if the final participant of the transaction commits, we fall back to a rollback.

I can't give much more specific instructions than this without going in and effectively implementing it myself, so if you can't get it to work after this I think it's better to just park it at this point.

rustyconover:
@Mytherin thanks for the pointers, I'll work on giving this another try later on.

PR #22348, opened 2026-04-28
Add bounds checks for ALPRD scan sizes before memcpy

diegobmadero:
Add runtime validation for ALPRD scan metadata before using decoded sizes in memcpy.

Invalid dictionary sizes or exception counts now raise an IOException for corrupt storage data.


Mytherin:
Thanks!

PR #22358, opened 2026-04-29
Move `PEGTransformerFactory` into `ParserCache`

Dtenwolde:
Previously there was a static singleton for `PEGTransformerFactory::GetInstance()`, but there's no reason to have to manage the `PEGTransformerFactory` and `PEGMatcher` separately, when the goal is to always use them together. This PR has a couple of improvements: 

- Unify the `PEGTransformerFactory` and `PEGMatcher` into `PEGMatcherCache` since they are always used together. We can now invalidate both together, when we add support for extending the grammar.  
- Removed the `PEGParserExtension` from the `autocomplete` extension
- `DatabaseInstance` now owns the `unique_ptr<PEGMatcherCache>`, this is propagated to the `Parser` via the `ClientContext::GetParserOptions()`: `ParserOptions::peg_matcher_cache`. 

However, there is a problem I encountered, which is that the `Parser parser` is not always constructed throught the `ClientContext::GetParserOptions()`, and we thus don't always have access to the `DatabaseInstance`, see https://github.com/duckdb/duckdb/blob/59314c4899953b059e3d4552f342aef0306fcdb9/src/catalog/default/default_functions.cpp#L173, as an example (also in `appender.cpp`, `pragma_queries.cpp`)
These are internal queries that we parse, so I assume they will not rely on any extensible grammars and use the built-in SQL. For these cases I made a "default" fallback that is created when we first hit `ParseQuery()` and is local to the `Parser`. Otherwise, I would've needed to pull `DatabaseInstance` through there, which didn't feel like the way to go.


extension/autocomplete/autocomplete_extension.cpp
```diff
@@ -363,7 +363,7 @@ class ClientContextCatalogProvider : public AutoCompleteCatalogProvider {
 		return ::duckdb::SuggestSettingName(context);
 	}
 	shared_ptr<PEGMatcher> GetPEGMatcher() override {
-		return GetGlobalPEGMatcherCache().GetMatcher();
+		return DatabaseInstance::GetDatabase(context).GetPEGMatcherCache().GetMatcher();
```
Mytherin: Can we add `PEGMatcher::Get(context)` similar to how we do this for other APIs?

src/include/duckdb/main/database.hpp
```diff
@@ -97,6 +99,7 @@ class DatabaseInstance : public enable_shared_from_this<DatabaseInstance> {
 	unique_ptr<LogManager> log_manager;
 	unique_ptr<ExternalFileCache> external_file_cache;
 	unique_ptr<ResultSetManager> result_set_manager;
+	unique_ptr<PEGMatcherCache> peg_matcher_cache;
```
Mytherin: Maybe call this `ParserCache`? imo the fact that we use `PEG` is an implementation detail. We can start removing the PEG name from many things and start renaming them to Parser - I think that's more clear.

Mytherin (commented):
Thanks! Looks great - some minor comments

src/include/duckdb/main/database.hpp
Dtenwolde: Sure!

Mytherin:
Thanks!

PR #22359, opened 2026-04-29
Tread BufferedFileHandle for single 0-12288 range request on attach

carlopi:
Issue is as follow: adding logging to requests performed on the wire, and doing something like:
```sql
 ATTACH 'https://blobs.duckdb.org/data/tpch-sf1.db';
```
we (that is, duckdb `main` branch) currently performs:
```
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db       (start=0, len=1048576)
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    8	4
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    0	4096
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    4096	4096
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    8192	4096
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    259272704	262144
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    258748416	262144
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db.wal
```

After this PR, 4 range requests are removed, to:
```
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    (start=0, len= 1048576) <--this could be 12288
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    259272704	262144
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    258748416	262144
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db.wal
```

There might be a path to remove also some other requests, in particular the double HEAD at 1st and 3rd, but for now, 3 HEAD + 3 GET are better than current 3+7.

----

Implementation side, there are 2 infrastructure changes to be aware: a `BaseFileHandle`, this `BaseFileHandle` has only a very limited surface:
```c++
class BaseFileHandle {
public:
       DUCKDB_API virtual ~BaseFileHandle() = default;

       DUCKDB_API virtual void ReadIntoBuffer(QueryContext context, void *buffer, idx_t nr_bytes, idx_t location) = 0;
       DUCKDB_API virtual idx_t GetFileSize() = 0;
       DUCKDB_API virtual string GetPath() const = 0;
};
```
but that's enough to handle the `CheckMagicBytes` / `ResolveDatabaseType` / `ReadAndChecksum` codepaths.

Instantiations could be either a proper `FileHandle` or a `BufferedFileHandle`, that is a  new class that packages a FileHandle plus a single optional buffer to be prefetched.

In case of attaching a database, the buffer is sized to 12288 bytes, so that a single GET (with range) covers both type discovery AND reading the MainHeader + 2 DatabaseHeaders in a go.

Once the initialisation phase has completed, `BufferedFileHandle` decays to a simple `FileHandle` (that means, also Write path becomes possible) via a call to `unique_ptr<FileHandle> BufferedFileHandle::ExtractInnerFileHandle()`.

----

Note, there are like 4 relevant path worth testing:
```
duckdb https://blobs.duckdb.org/data/tpch-sf1.db
duckdb duckdb:https://blobs.duckdb.org/data/tpch-sf1.db
duckdb -c "ATTACH 'https://blobs.duckdb.org/data/tpch-sf1.db';"
duckdb -c "ATTACH 'https://blobs.duckdb.org/data/tpch-sf1.db' (TYPE duckdb);"
```
and logging is at the moment not tracking correctly all requests in these codepaths (that's for a different PR).

As of now, there are no added tests, main thing tested by CI at the moment is whether any difference is detected.

### DIFF
--- a/src/common/buffered_file_handle.cpp
+++ b/src/common/buffered_file_handle.cpp
@@ -1,0 +1,113 @@
+#include "duckdb/common/file_system/buffered_file_handle.hpp"
+
+#include "duckdb/common/helper.hpp"
+#include "duckdb/main/client_context.hpp"
+
+#include <cstring>
+
+namespace duckdb {
+
+BufferedFileHandle::BufferedFileHandle(unique_ptr<FileHandle> inner_p) : inner(std::move(inner_p)) {
+	D_ASSERT(inner);
+}
+
+BufferedFileHandle::~BufferedFileHandle() = default;
+
+void BufferedFileHandle::CopyBufferFrom(const BufferedFileHandle &other) {
+	if (other.state != BufferedFileHandleState::REALIZED) {
+		state = BufferedFileHandleState::EMPTY;
+		range_offset = 0;
+		range_size = 0;
+		buffer.reset();
+		return;
+	}
+	range_offset = other.range_offset;
+	range_size = other.range_size;
+	buffer = make_unsafe_uniq_array<char>(range_size);
+	std::memcpy(buffer.get(), other.buffer.get(), range_size);
+	state = BufferedFileHandleState::REALIZED;
+}
+
+void BufferedFileHandle::RegisterPrefetch(idx_t offset, idx_t size) {
+	D_ASSERT(inner);
+	// Idempotent: if we already realized this exact range, do nothing.
+	// This makes repeated CheckMagicBytes-style calls a no-op on the wire.
+	if (state == BufferedFileHandleState::REALIZED && range_offset == offset && range_size == size) {
+		return;
+	}
+	buffer.reset();
+	if (size == 0) {
+		state = BufferedFileHandleState::EMPTY;
+		range_offset = 0;
+		range_size = 0;
+		return;
+	}
+	state = BufferedFileHandleState::HINT;
+	range_offset = offset;
+	range_size = size;
+}
+
+void BufferedFileHandle::ReadIntoBuffer(QueryContext context, void *out, idx_t nr_bytes, idx_t location) {
+	D_ASSERT(inner);
+
+	if (state == BufferedFileHandleState::REALIZED) {
+		if (location >= range_offset && location + nr_bytes <= range_offset + range_size) {
+			std::memcpy(out, buffer.get() + (location - range_offset), nr_bytes);
+			return;
+		}
+		// miss → fall through to forward
+	} else if (state == BufferedFileHandleState::HINT) {
+		// Realize the hint if the request overlaps the hint range.
+		const idx_t hint_end = range_offset + range_size;
+		const bool overlaps = location < hint_end && location + nr_bytes > range_offset;
+		if (overlaps) {
+			// Clamp the prefetch to actual file size so we don't try to read past EOF.
+			const idx_t file_size = inner->GetFileSize();
+			if (range_offset >= file_size) {
+				// Hint is entirely past EOF; nothing to buffer. Forward verbatim.
+				state = BufferedFileHandleState::EMPTY;
+				range_offset = 0;
+				range_size = 0;
+			} else {
+				const idx_t available = file_size - range_offset;
+				const idx_t actual_size = MinValue<idx_t>(range_size, available);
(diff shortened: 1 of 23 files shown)


carlopi:
Had to bring back a pessimization, that is, first GET is 0-1048576, but all bytes past 12288 are thrown away. That's to be fixed in a later PR.

src/include/duckdb/main/database_manager.hpp
```diff
@@ -70,7 +72,8 @@ class DatabaseManager {
 	//! Returns the database type. This might require checking the header of the file, in which case the file handle is
 	//! necessary. We can only grab the file handle, if it is not yet held, even for uncommitted changes. Thus, we have
 	//! to lock for this operation.
-	void GetDatabaseType(ClientContext &context, AttachInfo &info, const DBConfig &config, AttachOptions &options);
+	void GetDatabaseType(ClientContext &context, AttachInfo &info, const DBConfig &config, AttachOptions &options,
+	                     unique_ptr<BufferedFileHandle> *out_handle = nullptr);
```
Mytherin, line 76: Maybe we can add this to `AttachInfo` instead? Then all of the plumbing can be removed.

src/common/buffered_file_handle.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+namespace duckdb {
+
+BufferedFileHandle::BufferedFileHandle(unique_ptr<FileHandle> inner_p) : inner(std::move(inner_p)) {
+	D_ASSERT(inner);
+}
+
+BufferedFileHandle::~BufferedFileHandle() = default;
+
+void BufferedFileHandle::CopyBufferFrom(const BufferedFileHandle &other) {
```
Mytherin, line 16: I don't really see the need to introduce this new class.

Mytherin (commented):
Thanks for the PR!

This is a great change but I think this is introducing a lot of systemic complexity that I think we can avoid.

How about this:

* Add `unique_ptr<FileHandle>` and `unique_ptr<data_t[]> prefetched_header` to `AttachInfo` (or `AttachOptions` - whichever is easier)
* Modify `CheckMagicBytes` to pass in an `AttachInfo`
* In `CheckMagicBytes`, open a file handle according to the `AttachInfo` (i.e. using the same file flags as `SingleFileBlockManager::GetFileFlags`)
* Fill in the file handle and prefetched header if the file exists
* In `SingleFileBlockManager::LoadExistingDatabase` - re-use the file handle and the prefetched header if available, otherwise open a new file and read into the prefetched_header struct

I think in this path we can add support for re-use of the file handle (which also avoids the extra head request) and we can avoid adding all of this extra (I think not necessary) infrastructure and plumbing.

src/include/duckdb/main/database_manager.hpp
carlopi: I'll try, and come back on this. I guess, it does work for replacing this PR.

One thing that is harder (I think), and makes having a FileHandle of sort required (I think) would be trying to do single GET to include both the headers AND a fixed limited number of blocks (say, 2, in any case 500KB or 1MB on remote are irrelevant vs multiple roundtrips). My idea for that would be: optimistically fetch 12288 + small fixed number of headers, plus having a way to make use of those prefetched bytes (I though, via Read or similar), combined with reserving metadata blocks for blocks with ID 0 or 1.

Mytherin: Isn't that already done effectively in the http file system? Do we need a separate layer for this that exists only here?

PR #22361, opened 2026-04-29
Deprecate `PRAGMA enable_verification` and friends, move `verify_serializer` to separate option and verification run

Mytherin:
Related PRs: https://github.com/duckdb/duckdb/pull/22334, https://github.com/duckdb/duckdb/pull/22345, https://github.com/duckdb/duckdb/pull/22349, https://github.com/duckdb/duckdb/pull/22350, https://github.com/duckdb/duckdb/pull/22354

This PR deprecates `PRAGMA enable_verification` and remove it from all of the tests. Using it now toggles a warning to be printed. In addition, several related pragmas are removed (`PRAGMA verify_serializer`, `PRAGMA enable_external_verification`) and moved to settings.

The crux of this change is that `PRAGMA enable_verification` enabled many different internal verifications to run:

* `Copy()` statement
* `ToString()` followed by reparsing a statement
* `Serialize` followed by `Deserialize` a statement
* Run `PREPARE <stmt>` followed by `EXECUTE <stmt>`
* Run without optimizations
* Run with `verify_fetch_row`, `enable_caching_operators = false`, `verify_external`
* Run with logical plan serialization
* Run `EXPLAIN <stmt>`

This has been around since the beginning and was useful for structurally finding issues in these operators. The functionality isn't going anywhere - it has just been split off into different test configs and settings:

* `test/configs/verify_statement_copy.json`
* `test/configs/verify_statement_to_string.json`
* `test/configs/verify_statement_explain.json`
* `test/configs/verify_statement_prepare.json`
* `test/configs/verify_serializer.json`
* `test/configs/verify_stats.json`
* `test/configs/verify_serialization.json`
* `test/configs/force_external.json`
* `test/configs/verify_fetch_row.json`
* `test/configs/disable_caching_operators.json`

This is preferable over the `PRAGMA enable_verification` approach. That had a number of drawbacks:

* `PRAGMA enable_verification` was slow because it ran a lot of different verification routines all at once with no way of running individual verifiers.
* Each verification routine has a set of tests which they need to skip. For example, we cannot run many queries without the optimizer because they would be too slow. As a result, `PRAGMA enable_verification` skipped a lot of tests. By splitting up into different verification routines we only need to skip the tests necessary for that one routine.
* It was often not immediately obvious why `PRAGMA enable_verification` would fail because it did so many things - by splitting this off into individual runs it becomes more obvious which part is failing.


Mytherin:
Casual 2288 files changed :)

PR #22365, opened 2026-04-29
Avoid `DefaultExpression` greedily consuming `NOT NULL` constraint

Dtenwolde:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9038

This PR fixes the following query, where the `DefaultExpression` also consumed the `ColumnConstraint`: 
```sql
CREATE TABLE uuids(u UUID DEFAULT gen_random_uuid() NOT NULL);
```

The PostgreSQL parser used for `DefaultExpression` the `b_expr`, which is almost equal to the `a_expr` but does not allow `AND`, `NOT`, `IS` and `IN`. I don't want to go the `b_expr` route, but for this case I made a `ColumnDefaultExpr` that is a subset, skipping the `IsExpression`, because that lead to the problematic greedy consumption of `NOT NULL`. 

Fun fact I found out, you can create a table with default value `NULL` with a `NOT NULL` constraint, but it'll fail whenever you try to insert a value (also works in PostgreSQL): 
```sql
memory D CREATE or replace TABLE uuids(u INTEGER DEFAULT null not null, a INTEGER); 
memory D insert into uuids(a) values(42); 
Constraint Error:
NOT NULL constraint failed: uuids.u
```


Mytherin:
Thanks! LGTM - can we now also remove the FIXME in `test/configs/verify_statement_to_string.json`?

```json
    {
      "reason": "FIXME: new parser precedence issue",
      "paths": [
      "test/sql/function/uuid/test_uuid.test"
      ]
    }
```

Dtenwolde:
> can we now also remove the FIXME in `test/configs/verify_statement_to_string.json`? 

Wasn't aware that test was being skipped, but seems to pass without problem now :)

Mytherin:
Thanks!

PR #22366, opened 2026-04-29
Reject running unittest without run_tests.py in CI workflows

smvv:
Using unittest directly in CI can cause [problems](https://github.com/duckdb/duckdb/actions/runs/25103836238/job/73560165546?pr=22224#step:23:3678):
- unittest does not retry any failed (flaky) tests.
- inefficient hardware utilization (increases CI job time by 4 min, on a 16 vCPU core machine).

This PR adds a simple check to verify that unittest uses `run_tests.py`.

Without the added invocation fixes, we would see this error in the CI job `prepare`:
```
AssertionError: workflow `run:` commands using `unittest` must use scripts/ci/run_tests.py:

- .github/workflows/CrossVersion.yml:68: ./build/release/test/unittest --force-storage --test-temp-dir my_local_folder || true rm -rf my_local_folder/hive

- .github/workflows/CrossVersion.yml:242: ./build/release/test/unittest --force-storage --test-temp-dir my_local_folder || true rm -rf my_local_folder/hive

- .github/workflows/Main.yml:1252: valgrind ./build/relassert/test/unittest test/sql/tpch/tpch_sf001.test_slow

- .github/workflows/Main.yml:1421: ./build/release/test/unittest --test-config test/configs/vacuum_rebuild_indexes_force_storage.json

- .github/workflows/NightlyTests.yml:684: build/relassert/test/unittest --test-config test/configs/hash_zero.json

- .github/workflows/Windows.yml:315: set -x echo "PWD=$PWD" echo "PATH=$PATH" where python python --version where unittest.exe || true if [ -f build/release/test/unittest.exe ]; then ls -l build/release/test/unitte...
```

### DIFF
--- a/.github/workflows/CrossVersion.yml
+++ b/.github/workflows/CrossVersion.yml
@@ -66,7 +66,7 @@ jobs:
       - name: Unit Test
         shell: bash
         run: |
-          ./build/release/test/unittest --force-storage --test-temp-dir my_local_folder || true
+          python3 scripts/ci/run_tests.py --test-flags="--force-storage --test-temp-dir my_local_folder" ./build/release/test/unittest || true
           rm -rf my_local_folder/hive
 
       - uses: actions/upload-artifact@v7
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks!

PR #22369, opened 2026-04-29
Use -mcpu=native on aarch64 when compiling with `NATIVE_ARCH=1`

krenzland:
On aarch64, -march=native does not fully tune for the current machine. Enabled features and applied optimizations might differ between march=native and mcpu=native.


Mytherin:
Thanks!

PR #22370, opened 2026-04-29
`GROUPING()`/`GROUPING_ID()` with no args uses all `GROUP BY` columns by default

yan-alex:
`GROUPING()`/`GROUPING_ID()` with no args uses all `GROUP BY` columns by default.


yan-alex:
CI Failure is unrelated

Mytherin:
Thanks - could you expand a bit on this change? Is this to match the behavior of another system?

yan-alex:
@Mytherin Yes, Spark and Databricks SQL do this. But besides that it also feels SQL friendly to me.

Mytherin:
Alright - sounds good to me. Thanks!

PR #22371, opened 2026-04-29
Use Clang++ and clangd version 20 in CI

smvv:
### Upgrade clang to v20

Detect more warnings, errors and use improved thread/address/ub sanitizers.

**We don't change compiler for most jobs.**
The compiler is only changed for the CI jobs `linux-relassert`, `threadsan` and `check-clangd-tidy`.

### Detected more tidy errors
```
src/execution/sample/reservoir_sample.cpp:707:8: Error: The variable 'col_type' is copy-constructed from a const reference but is only used as const reference; consider making it a const reference [performance-unnecessary-copy-initialization]
  705 |  
  706 |  	for (idx_t i = 0; i < reservoir_chunk->chunk.ColumnCount(); i++) {
  707 |  		auto col_type = types[i];
      |         ^~~~~~~~
  708 |  		if (ValidSampleType(col_type) || !stats_sample) {
  709 |  			D_ASSERT(this_.data[i].GetVectorType() == VectorType::FLAT_VECTOR);
src/function/aggregate/distributive/minmax.cpp:29:56: Error: The parameter 'type' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
   27 |  
   28 |  template <class OP>
   29 |  static AggregateFunction GetUnaryAggregate(LogicalType type) {
      |                                                         ^~~~
   30 |  	switch (type.InternalType()) {
   31 |  	case PhysicalType::BOOL:
src/function/aggregate/distributive/minmax.cpp:401:44: Error: The parameter 'name' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
  399 |  
  400 |  template <class OP, class OP_STRING, class OP_VECTOR>
  401 |  AggregateFunction GetMinMaxOperator(string name) {
      |                                             ^~~~
  402 |  	return AggregateFunction(std::move(name), {LogicalType::ANY}, LogicalType::ANY, nullptr, nullptr, nullptr, nullptr,
  403 |  	                         nullptr, nullptr, BindMinMax<OP, OP_STRING, OP_VECTOR>);
src/function/aggregate/distributive/minmax.cpp:436:10: Error: Returning a constant reference parameter may cause use-after-free when the parameter is constructed from a temporary [bugprone-return-const-ref-from-parameter]
  434 |  
  435 |  	static const T &GetValue(const T &val) {
  436 |  		return val;
      |           ^~~
  437 |  	}
  438 |  };
src/storage/table/row_group_reorderer.cpp:45:88: Error: The parameter 'end' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
   43 |  
   44 |  template <typename It, typename End>
   45 |  void AddRowGroups(multimap<Value, RowGroupSegmentNodeEntry> &row_group_map, It it, End end,
      |                                                                                         ^~~
   46 |                    vector<reference<SegmentNode<RowGroup>>> &ordered_row_groups, const idx_t row_limit,
   47 |                    const OrderByColumnType column_type, const OrderByStatistics stat_type) {
src/storage/table/row_group_reorderer.cpp:96:41: Error: The parameter 'end' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
   94 |  
   95 |  template <typename It, typename End>
   96 |  It SkipOffsetPrunedRowGroups(It it, End end, idx_t row_group_offset) {
      |                                          ^~~
   97 |  	while (row_group_offset > 0 && it != end) {
   98 |  		++it;
src/storage/table/row_group_reorderer.cpp:105:36: Error: The parameter 'end' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
  103 |  
  104 |  template <typename It, typename End>
  105 |  void InsertAllRowGroups(It it, End end, vector<reference<SegmentNode<RowGroup>>> &ordered_row_groups) {
      |                                     ^~~
  106 |  	for (; it != end; ++it) {
  107 |  		ordered_row_groups.push_back(it->second.row_group);
src/storage/table/row_group_reorderer.cpp:148:57: Error: The parameter 'end' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
  146 |  
  147 |  template <typename It, typename End>
  148 |  OffsetPruningResult FindOffsetPrunableChunks(It it, End end, const OrderByStatistics order_by,
      |                                                          ^~~
  149 |                                               const OrderByColumnType column_type, const idx_t row_offset) {
  150 |  	if (it == end) {
make: *** [Makefile:660: tidy-check-clangd] Error 1

```

### DIFF
--- a/.github/workflows/ExtendedTests.yml
+++ b/.github/workflows/ExtendedTests.yml
@@ -44,10 +44,7 @@ jobs:
         python-version: '3.12'
 
     - name: Install
-      shell: bash
-      run: |
-        python3 scripts/ci/retry.py -- make toolsci
-        python3 scripts/ci/retry.py -- sudo apt-get install -y -qq libcurl4-openssl-dev
+      run: python3 scripts/ci/retry.py -- make toolsci
 
     - uses: ./.github/actions/ccache-action
 
@@ -244,7 +241,6 @@ jobs:
          run: |
            python3 scripts/ci/retry.py -- make toolsci
            python3 scripts/ci/retry.py -- sudo apt-get install -y -qq llvm
-           python3 scripts/ci/retry.py -- pip install requests
 
        - uses: ./.github/actions/ccache-action
 
@@ -327,7 +323,6 @@ jobs:
          shell: bash
          run: |
            python3 scripts/ci/retry.py -- make toolsci
-           python3 scripts/ci/retry.py -- pip install requests
 
        - uses: ./.github/actions/ccache-action
 
--- a/scripts/ci/job_stages.py
+++ b/scripts/ci/job_stages.py
@@ -13,7 +13,6 @@ PULL_REQUEST_JOBS = [
     "linux-relassert-tests",
     "regression",
     "tidy-check",
-    "check-clangd-tidy",
     "extensions",
     "wasm-eh",
     "linux-release",
@@ -31,7 +30,7 @@ PULL_REQUEST_JOBS = [
 
 NIGHTLY_ONLY_JOBS = [
     "main_julia",
-    # "check-clangd-tidy",
+    "check-clangd-tidy",
     "valgrind",
     "static-libs-osx",
     "static-libs-windows-mingw",
--- a/scripts/ci/job_stages.py
+++ b/scripts/ci/job_stages.py
@@ -13,6 +13,7 @@ PULL_REQUEST_JOBS = [
     "linux-relassert-tests",
     "regression",
     "tidy-check",
+    "check-clangd-tidy",
     "extensions",
     "wasm-eh",
     "linux-release",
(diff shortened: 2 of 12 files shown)


scripts/ci/job_stages.py
```diff
@@ -30,7 +31,7 @@
 
 NIGHTLY_ONLY_JOBS = [
     "main_julia",
-    "check-clangd-tidy",
+    # "check-clangd-tidy",
```
smvv: - [x] Run `check-clangd-tidy` only on nightly jobs.

.github/workflows/ExtendedTests.yml
```diff
@@ -44,10 +44,7 @@ jobs:
         python-version: '3.12'
 
     - name: Install
-      shell: bash
-      run: |
-        python3 scripts/ci/retry.py -- make toolsci
-        python3 scripts/ci/retry.py -- sudo apt-get install -y -qq libcurl4-openssl-dev
```
smvv, line 50: I've moved `libcurl4-openssl-dev` and `python3-requests` into `make toolsci`, so we can get rid of these special cases.

Mytherin:
Thanks!

PR #22040, opened 2026-04-13
PEG grammar add type annotations

Dtenwolde:
This is part of a series of PRs meant to generate more of the transformer code than we currently do. 

This PR introduces type annotations for the PEG grammar files. The rulename now has the return type in between `[]`, so it becomes easier to generate the transformer code. 

There's also a couple of improvements to the scripts that generate the PEG grammar and transformer: 
- `inline_grammar.py` now also verifies the syntax of grammar files. Malformed grammar files would result in runtime errors when we tried to build the parser. This was annoying and it's easier to catch any syntax errors (heh) during the creation of the `inlined_grammar.gram` file. 
- `build_peg_grammar.sh` now also runs `generate_peg_transformer.py`, which provides a coverage check on `TransformRule`. In the future, this script will help generate the transformer code (the declaration, registration and a template for the definition (throwing a NotImplementedException). Right now it provides that as output of the script (So it's easy to copy/paste), but it doesn't modify any files. 
In a future iteration I will also generate the code that extracts the `ParseResult`s. Right now we still have to do that ourselves (`auto &list_pr = parse_result->Cast<ListParseResult>(0)`), but that is error prone. Any changes to a grammar rule can cause the indices to change, which may be not obvious (Same as with Postgres parser $1, $2, etc.).

PR #22372, opened 2026-04-29
Add type hints to the grammar

Dtenwolde:
Second attempt at adding type information for the grammar rules (previous attempt [here](https://github.com/duckdb/duckdb/pull/22040))

I added a `grammar_types.yml` file that groups grammar rules by their expected return type. This return type is used in the `generate_transformer.py` to provide code snippets with the correct return type (Previously it always used `unique_ptr<SQLStatement>`). This will later be used to auto-generate more parts of the transformer. If a grammar rule is missing from the `yml` file, the `generate_peg_transformer.py` script will error. 

I also moved the scripts related to the parser to a subfolder `scripts/parser`, I hope that's okay.


Mytherin:
Thanks!

PR #22382, opened 2026-04-30
ISSUE-22274: Fix LOGICAL_FILTER column lifetime projection mapping

### LINKED ISSUE #22274: INTERNAL Error: Vector::Reference used on vector of different type (source BOOLEAN referenced TINYINT)
### What happens?

Querying an Arrow table registered via `con.register()` that contains `string_view` columns crashes with `INTERNAL Error: Vector::Reference used on vector of different type (source BOOLEAN referenced TINYINT)` when the query uses an `IN (...)` clause on the `string_view` column.

DuckDB rewrites `IN (...)` into a MARK JOIN, and the mark vector type (BOOLEAN) conflicts with the source column type (TINYINT) during `ConstructMarkJoinResult`.

Replacing `pa.string_view()` with `pa.utf8()` (same data, same query) works fine.

### To Reproduce

```python
import duckdb
import pyarrow as pa

n = 50_000
makes = ['TOYOTA','HONDA','FORD','CHEVROLET','BMW','MERCEDES','NISSAN',
         'HYUNDAI','KIA','SUBARU','MAZDA','VPG','ISUZU','SMART','LUCID',
         'LOTUS','PLYMOUTH','DODGE_MITS','VINFAST','POLESTAR']

table = pa.table({
    'name': pa.array([makes[i % 20] for i in range(n)], type=pa.string_view()),
    'flag': pa.array([0 if i % 10 != 0 else 1 for i in range(n)], type=pa.int8()),
})

con = duckdb.connect()
con.register('data', table)

con.execute("""
    SELECT name, COUNT(*) FROM data
    WHERE flag < 1
      AND name IN ('VPG','ISUZU','SMART','LUCID','LOTUS','POLESTAR')
    GROUP BY 1
""")
```

Output:
```
INTERNAL Error: Vector::Reference used on vector of different type (source BOOLEAN referenced TINYINT)
```

Changing `pa.string_view()` to `pa.utf8()` makes the query succeed.

### OS:

Linux (Ubuntu 22.04), x86_64

### DuckDB Version:

1.5.2

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Colby Vickerson

### Affiliation:

Liberty Mutual

- [x] Yes, I have

- [x] Yes, I have

Yes

herbenderbler:
## Summary

Fixes incorrect `LogicalFilter` projection mapping during column lifetime analysis when a filter sits above plans that combine **predicate-only** payload columns with **`IN (...)` rewritten as a MARK join**. That mis-mapping could misalign physical chunks and surface as `INTERNAL Error: Vector::Reference used on vector of different type (source BOOLEAN referenced TINYINT)` (reported against Arrow `string_view` workloads in [#22274](https://github.com/duckdb/duckdb/issues/22274)).

The fix separates **bindings needed in the filter’s projected output** (referenced above the filter) from **bindings needed only to evaluate predicates**, visits predicates and children so child inputs stay live, then rebuilds `filter.projection_map` by remapping desired output bindings onto the child’s bindings **after** the child visit.

Fixes [#22274](https://github.com/duckdb/duckdb/issues/22274).

## Implementation Details

- **`src/optimizer/column_lifetime_analyzer.cpp`**
  - Specialized handling for `LogicalOperatorType::LOGICAL_FILTER`:
    - Collect `desired_output_bindings` from the child using only bindings still referenced **above** the filter (via `column_references` before predicate visitation mutates reference sets).
    - Call `VisitOperatorExpressions(op)` and `VisitOperatorChildren(op)` so predicate columns remain live for traversal without becoming part of output liveness.
    - Rebuild `filter.projection_map` with `GenerateProjectionMapByBindings(current_child_bindings, desired_output_bindings, ...)`, which maps desired bindings to post-visit child column indices and clears the map to identity when mapping would be invalid or full-width.
  - Add `GenerateProjectionMapByBindings` helper for binding-based projection map construction.

## Tests

- **`test/optimizer/issue_22274_column_lifetime_mark_join.test`**
  - SQL regression: `flag < 1` plus `IN (...)` with enough scalar RHS values to force MARK join; validates row counts, grouped aggregates, and an `EXPLAIN` guard for `FILTER` + `Join Type: MARK`.

- **`test/arrow/arrow_filter_pushdown.cpp`**
  - C++ regression aligned with review feedback: deterministic `name` / `flag` fixture; with `produce_arrow_string_view` + Arrow 1.5 output, asserts standalone `FILTER`, no filter pushdown into `ARROW_SCAN`, `MARK` present, `SEMI` absent, and the same predicate pattern’s grouped results on the backing table for semantic checks.

## Checklist

- [x] Regression tests cover the reported failure pattern and a plan-shape guard where applicable.
- [x] Changes are limited to column lifetime analysis for `LOGICAL_FILTER` plus targeted tests.
- [x] Ran relevant fast tests locally (`unittest` for the new Arrow case / filter group and the new optimizer `.test`).

### DIFF
--- a/scripts/reproduce_github_issue_22274_arrow.py
+++ b/scripts/reproduce_github_issue_22274_arrow.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+"""
+Manual reproducer for https://github.com/duckdb/duckdb/issues/22274
+
+Uses pyarrow + DuckDB register() with produce_arrow_string_view and Arrow 1.5 output,
+matching the failure mode discussed on the issue and in test/arrow/arrow_filter_pushdown.cpp.
+
+Requires: pip install duckdb pyarrow
+
+Run (from repo root, with a duckdb build on PYTHONPATH or default pip duckdb):
+  python3 scripts/reproduce_github_issue_22274_arrow.py
+"""
+
+from __future__ import annotations
+
+import sys
+
+try:
+	import duckdb
+except ImportError as e:
+	print("Install duckdb: pip install duckdb", file=sys.stderr)
+	raise SystemExit(1) from e
+
+try:
+	import pyarrow  # noqa: F401 — required for Relation.arrow() / register()
+except ImportError as e:
+	print("Install pyarrow: pip install pyarrow", file=sys.stderr)
+	raise SystemExit(1) from e
+
+
+def main() -> None:
+    con = duckdb.connect()
+    con.execute("PRAGMA threads=1")
+    con.execute("SET produce_arrow_string_view = true")
+    con.execute("SET arrow_output_version = '1.5'")
+
+    con.execute(
+        """
+		CREATE TABLE src AS SELECT
+			['TOYOTA','HONDA','FORD','CHEVROLET','BMW','MERCEDES','NISSAN','HYUNDAI','KIA','SUBARU',
+			 'MAZDA','VPG','ISUZU','SMART','LUCID','LOTUS','PLYMOUTH','DODGE_MITS','VINFAST',
+			 'POLESTAR'][((i % 20) + 1)::INTEGER] AS name,
+			CASE WHEN i % 10 != 0 THEN 0::TINYINT ELSE 1::TINYINT END AS flag
+		FROM range(50000) tbl(i)
+	"""
+    )
+
+    rel = con.sql("SELECT name, flag FROM src")
+    arrow_tbl = rel.arrow()
+    # register as in-memory Arrow table (same integration surface as many pyarrow workflows)
+    con.register("data", arrow_tbl)
+
+    sql = """
+	SELECT name, COUNT(*) AS cnt FROM data
+	WHERE flag < 1 AND name IN ('VPG','ISUZU','SMART','LUCID','LOTUS','POLESTAR')
+	GROUP BY 1 ORDER BY 1
+	"""
+    try:
+        print(con.execute(sql).fetchall())
+        print("OK: query completed without error.")
+    except duckdb.Error as e:
+        print(f"Query failed: {e}", file=sys.stderr)
+        raise SystemExit(2)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/execution/operator/join/physical_hash_join.cpp
+++ b/src/execution/operator/join/physical_hash_join.cpp
@@ -74,6 +74,32 @@ PhysicalHashJoin::PhysicalHashJoin(PhysicalPlan &physical_plan, LogicalOperator 
 		lhs_output_columns.col_types.push_back(lhs_input_types[lhs_col]);
 	}
 
+	// MARK join output types must be lhs_payload + BOOLEAN (ConstructMarkJoinResult). Refresh only when the logical
+	// plan disagrees with the materialized LHS layout (#22274) — usually they already match after optimizer fixes.
+	if (join_type == JoinType::MARK) {
+		const idx_t n = lhs_output_columns.col_types.size();
+		const idx_t expected_cols = n + 1;
+		bool mismatch = types.size() != expected_cols;
+		if (!mismatch) {
+			for (idx_t i = 0; i < n; i++) {
+				if (types[i] != lhs_output_columns.col_types[i]) {
+					mismatch = true;
+					break;
+				}
+			}
+			if (!mismatch && types[n].id() != LogicalTypeId::BOOLEAN) {
+				mismatch = true;
+			}
+		}
+		if (mismatch) {
+			types.resize(expected_cols);
+			for (idx_t i = 0; i < n; i++) {
+				types[i] = lhs_output_columns.col_types[i];
+			}
+			types[n] = LogicalType::BOOLEAN;
+		}
+	}
+
 	// initialize residual predicate structures if present
 	if (residual_info) {
 		InitializeResidualPredicate(lhs_input_types, probe_cols);
@@ -130,20 +156,29 @@ void PhysicalHashJoin::ExtractResidualPredicateColumns(unique_ptr<Expression> &p
 void PhysicalHashJoin::InitializeResidualPredicate(const vector<LogicalType> &lhs_input_types,
                                                    const vector<idx_t> &probe_cols) {
 	D_ASSERT(residual_info);
-	// build lhs_probe_columns (output + predicate columns)
-	unordered_set<idx_t> required_probe_cols;
+	// Build lhs_probe layout with lhs OUTPUT indices first — they must remain prefix-aligned with
+	// PhysicalHashJoin::lhs_output_columns so ConstructMarkJoinResult can pair result vector types with probe vectors.
+	// Sorting merged {output ∪ predicate} indices lexicographically scrambled that order (#22274).
+	unordered_set<idx_t> seen;
+	lhs_probe_columns.col_idxs.reserve(lhs_output_columns.col_idxs.size() + probe_cols.size());
+	lhs_probe_columns.col_types.reserve(lhs_probe_columns.col_idxs.capacity());
 	for (auto col : lhs_output_columns.col_idxs) {
-		required_probe_cols.insert(col);
+		if (!seen.insert(col).second) {
+			continue;
+		}
+		lhs_probe_columns.col_idxs.push_back(col);
+		lhs_probe_columns.col_types.push_back(lhs_input_types[col]);
 	}
+	vector<idx_t> predicate_only_extra;
 	for (auto col : probe_cols) {
-		required_probe_cols.insert(col);
+		if (seen.insert(col).second) {
+			predicate_only_extra.push_back(col);
+		}
 	}
-
-	lhs_probe_columns.col_idxs.assign(required_probe_cols.begin(), required_probe_cols.end());
-	std::sort(lhs_probe_columns.col_idxs.begin(), lhs_probe_columns.col_idxs.end());
-
-	for (auto col_idx : lhs_probe_columns.col_idxs) {
-		lhs_probe_columns.col_types.push_back(lhs_input_types[col_idx]);
+	std::sort(predicate_only_extra.begin(), predicate_only_extra.end());
+	for (auto col : predicate_only_extra) {
+		lhs_probe_columns.col_idxs.push_back(col);
+		lhs_probe_columns.col_types.push_back(lhs_input_types[col]);
 	}
 
 	// build mapping for predicate probe columns
@@ -156,15 +191,11 @@ void PhysicalHashJoin::InitializeResidualPredicate(const vector<LogicalType> &lh
 		}
 	}
 
-	// build lhs_output_in_probe mapping
+	lhs_output_in_probe.clear();
 	lhs_output_in_probe.reserve(lhs_output_columns.col_idxs.size());
-	for (auto output_col_idx : lhs_output_columns.col_idxs) {
-		for (idx_t i = 0; i < lhs_probe_columns.col_idxs.size(); i++) {
-			if (lhs_probe_columns.col_idxs[i] == output_col_idx) {
-				lhs_output_in_probe.push_back(i);
-				break;
-			}
-		}
+	for (idx_t i = 0; i < lhs_output_columns.col_idxs.size(); i++) {
+		lhs_output_in_probe.push_back(i);
+		D_ASSERT(lhs_probe_columns.col_idxs[i] == lhs_output_columns.col_idxs[i]);
 	}
 }
 
--- a/src/execution/physical_plan/plan_comparison_join.cpp
+++ b/src/execution/physical_plan/plan_comparison_join.cpp
@@ -19,7 +19,12 @@ static void RewriteJoinCondition(unique_ptr<Expression> &root_expr, idx_t offset
 }
 
 PhysicalOperator &PhysicalPlanGenerator::PlanComparisonJoin(LogicalComparisonJoin &op) {
-	// now visit the children
+	// Refresh join output types from children before planning; keeps MARK plans consistent when logical types lag
+	// behind child layouts (see ColumnBindingResolver after optimizer rewrites). Skip for delim joins: resolving
+	// before child plans are built breaks LogicalRecursiveCTE::ResolveTypes (recursive CTE key aggregation).
+	if (op.type != LogicalOperatorType::LOGICAL_DELIM_JOIN) {
+		op.ResolveOperatorTypes();
+	}
 	D_ASSERT(op.children.size() == 2);
 	idx_t lhs_cardinality = op.children[0]->EstimateCardinality(context);
 	idx_t rhs_cardinality = op.children[1]->EstimateCardinality(context);
--- a/src/execution/physical_plan/plan_filter.cpp
+++ b/src/execution/physical_plan/plan_filter.cpp
@@ -1,12 +1,8 @@
 #include "duckdb/execution/operator/filter/physical_filter.hpp"
 #include "duckdb/execution/operator/projection/physical_projection.hpp"
 #include "duckdb/execution/physical_plan_generator.hpp"
-#include "duckdb/optimizer/matcher/expression_matcher.hpp"
-#include "duckdb/planner/expression/bound_comparison_expression.hpp"
-#include "duckdb/planner/expression/bound_constant_expression.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "duckdb/planner/operator/logical_filter.hpp"
-#include "duckdb/planner/operator/logical_get.hpp"
 
 namespace duckdb {
 
--- a/src/optimizer/column_lifetime_analyzer.cpp
+++ b/src/optimizer/column_lifetime_analyzer.cpp
@@ -17,6 +17,58 @@
 
 namespace duckdb {
 
+//! Peek through FILTER/PROJECTION/ORDER unary chains to detect the IN (...) MARK join InClauseRewriter produces.
+//! (Broad DFS is unsafe — deeper MARK joins unrelated to this FILTER must keep projection maps intact.)
+static const LogicalComparisonJoin *PeelUnaryChainToMarkJoin(const LogicalOperator *op) {
+	while (op && op->children.size() == 1) {
+		switch (op->type) {
+		case LogicalOperatorType::LOGICAL_FILTER:
+		case LogicalOperatorType::LOGICAL_PROJECTION:
+		case LogicalOperatorType::LOGICAL_ORDER_BY:
+			op = op->children[0].get();
+			continue;
+		default:
+			op = nullptr;
+			break;
+		}
+	}
+	if (!op || op->type != LogicalOperatorType::LOGICAL_COMPARISON_JOIN) {
+		return nullptr;
+	}
+	auto &comp_join = op->Cast<LogicalComparisonJoin>();
+	return comp_join.join_type == JoinType::MARK ? &comp_join : nullptr;
+}
+
+static void GenerateProjectionMapByBindings(const vector<ColumnBinding> &current_bindings,
+                                            const vector<ColumnBinding> &desired_output_bindings,
+                                            vector<ProjectionIndex> &projection_map) {
+	projection_map.clear();
+	if (desired_output_bindings.empty()) {
+		return;
+	}
+	projection_map.reserve(desired_output_bindings.size());
+	for (const auto &desired_binding : desired_output_bindings) {
+		bool found = false;
+		for (idx_t i = 0; i < current_bindings.size(); i++) {
+			if (current_bindings[i] != desired_binding) {
+				continue;
+			}
+			projection_map.emplace_back(i);
+			found = true;
+			break;
+		}
+		if (!found) {
+			// Child rewrites can remove or replace bindings. Fallback to identity map if we cannot express this
+			// mapping.
+			projection_map.clear();
+			return;
+		}
+	}
+	if (projection_map.size() == current_bindings.size()) {
+		projection_map.clear();
+	}
+}
+
 void ColumnLifetimeAnalyzer::ExtractUnusedColumnBindings(const vector<ColumnBinding> &bindings,
                                                          column_binding_set_t &unused_bindings) {
 	for (idx_t i = 0; i < bindings.size(); i++) {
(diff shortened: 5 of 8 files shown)


lnkuiper:
The test you added already completes successfully on v1.5.2, and the query in the original issue:
```sql
create table data as select getvariable('makes')[range % 20 + 1] as name, case when range % 10 != 0 then 0 else 1 end as flag from range(50_000);
SELECT name, COUNT(*) FROM data
WHERE flag < 1
  AND name IN ('VPG','ISUZU','SMART','LUCID','LOTUS','POLESTAR')
GROUP BY 1;
```
Also completes successfully. What exactly is your diff fixing?

The original issue states that changing from `string_view` to `utf8` fixes the problem, so perhaps this is a Python issue, not an issue in the optimizer?

herbenderbler:
> The test you added already completes successfully on v1.5.2, and the query in the original issue:
> 
> ```sql
> create table data as select getvariable('makes')[range % 20 + 1] as name, case when range % 10 != 0 then 0 else 1 end as flag from range(50_000);
> SELECT name, COUNT(*) FROM data
> WHERE flag < 1
>   AND name IN ('VPG','ISUZU','SMART','LUCID','LOTUS','POLESTAR')
> GROUP BY 1;
> ```
> 
> Also completes successfully. What exactly is your diff fixing?
> 
> The original issue states that changing from `string_view` to `utf8` fixes the problem, so perhaps this is a Python issue, not an issue in the optimizer?

Thanks for your eagle 👀 😄.  You’re right that many plain-SQL variants, including the issue’s query, can succeed on released builds. The failure is not _Arrow vs SQL_ in the sense of a separate Python bug.  It’s an optimizer bug in `ColumnLifetimeAnalyzer` for a specific physical/logical shape: a `LogicalFilter` directly above a MARK join produced from IN (...), together with other predicates on real payload columns (e.g. flag < 1).

In that configuration, we could generate a bad `LogicalFilter.projection_map`.  We treated “unused” child columns before fully accounting for predicate column refs.  We could keep the internal MARK result column in the filter’s projection map, even though it should not flow through like a normal payload column. That misaligned the filter’s projected chunk with what downstream code expected and surfaced as `Vector::Reference` … `BOOLEAN` vs `TININT` at execution.

My fix does two things:

1. visit filter predicates before deciding unused bindings
2. drop the MARK-index binding from the filter projection map when the child is a MARK join.  I believe this mirrors how IN lowering treats that column.

The utf8 vs string_view / Python angle in the report is plausibly what nudges plans or types so that bad shape shows up more often, but the crash is fixed in core for that shape.  The regression test now includes an `EXPLAIN` guard so we don’t silently “pass” if the planner stops emitting FILTER-over-MARK for that query.

Hopefully this clarifies my PR  better than I initially did.  If not, please feel free to lob any and all questions I negelcted to answer 😅

lnkuiper:
Thanks for explaining! I think we need to be careful though: I don’t think this PR is ready as-is.

## Problem with the current fix

The underlying issue is not that the `MARK` column needs a special-case after recomputing the filter projection.
The issue is that `LOGICAL_FILTER` has two distinct "liveness" columns:

1. Columns needed by operators above the filter: these determine the filter’s output projection_map.
2. Columns needed to evaluate the filter predicates: these need to be be preserved in the filter input, but might need to be projected out by the filter's projection map.

The current PR mixes these two cases up by visiting filter predicates before deciding the filter output projection.
This makes predicate-only columns, e.g., flag in `WHERE flag < 1`, appear live above the filter.
Then it tries to repair the `IN (...) MARK` case by pruning the `MARK` binding afterwards, which is fragile because "empty projection_map" = "identity projection" in DuckDB.

So after pruning `MARK`, if the projection map becomes empty, the filter may accidentally carry all child columns again, including the `MARK` column. That makes the fix in this PR unreliable.

## Proposed fix

The filter case should preserve the current separation between filter output and input. Roughly:

```C++
case LogicalOperatorType::LOGICAL_FILTER: {
	auto &filter = op.Cast<LogicalFilter>();
	if (everything_referenced) {
		break;
	}

	// First decide what the FILTER should output, based only on references
	// from operators above the filter.
	column_binding_set_t unused_bindings;
	ExtractUnusedColumnBindings(op.children[0]->GetColumnBindings(), unused_bindings);

	// Now visit filter predicates so columns needed to evaluate the filter are
	// live for the child traversal, but not necessarily emitted by the filter.
	VisitOperatorExpressions(op);
	VisitOperatorChildren(op);

	GenerateProjectionMap(op.children[0]->GetColumnBindings(), unused_bindings, filter.projection_map);
	return;
}
```

The exact implementation needs care because `VisitOperatorChildren(op)` can rewrite child binding/projection maps.
If child bindings can change, the code should compute the desired output bindings before visiting the child, then remap them against the child bindings after the visit, similar to
`LogicalOperatorVisitor::VisitChildOfOperatorWithProjectionMap`.

## Testing

I tried to reproduce with SQL-only, also by disabling filter pushdown, but could not manage to reproduce the issue.
This makes the currently added test not very useful, but I also don't really know how to reproduce from just SQL.
Maybe this needs a C++ test? We do have some Arrow tests. I would like to have a reproducible example within this PR.

herbenderbler:
> Thanks for explaining! I think we need to be careful though: I don’t think this PR is ready as-is.
> ## Problem with the current fix
> 
> The underlying issue is not that the `MARK` column needs a special-case after recomputing the filter projection. The issue is that `LOGICAL_FILTER` has two distinct "liveness" columns:
> 
>     1. Columns needed by operators above the filter: these determine the filter’s output projection_map.
> 
>     2. Columns needed to evaluate the filter predicates: these need to be be preserved in the filter input, but might need to be projected out by the filter's projection map.
> 
> 
> The current PR mixes these two cases up by visiting filter predicates before deciding the filter output projection. This makes predicate-only columns, e.g., flag in `WHERE flag < 1`, appear live above the filter. Then it tries to repair the `IN (...) MARK` case by pruning the `MARK` binding afterwards, which is fragile because "empty projection_map" = "identity projection" in DuckDB.
> 
> So after pruning `MARK`, if the projection map becomes empty, the filter may accidentally carry all child columns again, including the `MARK` column. That makes the fix in this PR unreliable.
> ## Proposed fix
> 
> The filter case should preserve the current separation between filter output and input. Roughly:
> 
> ```c++
> case LogicalOperatorType::LOGICAL_FILTER: {
> 	auto &filter = op.Cast<LogicalFilter>();
> 	if (everything_referenced) {
> 		break;
> 	}
> 
> 	// First decide what the FILTER should output, based only on references
> 	// from operators above the filter.
> 	column_binding_set_t unused_bindings;
> 	ExtractUnusedColumnBindings(op.children[0]->GetColumnBindings(), unused_bindings);
> 
> 	// Now visit filter predicates so columns needed to evaluate the filter are
> 	// live for the child traversal, but not necessarily emitted by the filter.
> 	VisitOperatorExpressions(op);
> 	VisitOperatorChildren(op);
> 
> 	GenerateProjectionMap(op.children[0]->GetColumnBindings(), unused_bindings, filter.projection_map);
> 	return;
> }
> ```
> 
> The exact implementation needs care because `VisitOperatorChildren(op)` can rewrite child binding/projection maps. If child bindings can change, the code should compute the desired output bindings before visiting the child, then remap them against the child bindings after the visit, similar to `LogicalOperatorVisitor::VisitChildOfOperatorWithProjectionMap`.
> ## Testing
> 
> I tried to reproduce with SQL-only, also by disabling filter pushdown, but could not manage to reproduce the issue. This makes the currently added test not very useful, but I also don't really know how to reproduce from just SQL. Maybe this needs a C++ test? We do have some Arrow tests. I would like to have a reproducible example within this PR.

Thank you so much for those details.  It helped clarify a number of pieces where my 🧠 fell through 😅 

I changed my PR to follow the separation you described:

- `LOGICAL_FILTER` now computes **output liveness first** (only bindings referenced above the filter).
- Filter predicates are then visited only to preserve **input liveness** for child traversal.
- After visiting children, the filter output is rebuilt by remapping desired bindings against the child's current bindings (binding-based remap), so child binding rewrites are handled safely.
- The MARK-specific post-prune logic was removed.

I _believe_ this will help avoid the prior fragility where predicate-only columns leaked into output liveness and where an empty `projection_map` could accidentally re-enable identity projection.

Validation done locally:
- `test/optimizer/issue_22274_column_lifetime_mark_join.test` passes.
- Related optimizer regressions also pass (`remove_unused_filter_column`, `projection_pullup`, `pullup_filters`, `pushdown_set_op`).

As always, please let me know if I completely went sideways with this PR 🙇

lnkuiper:
Thanks, this looks a lot better. I would really like to see an added test, however. I think it should be possible to recreate the original arrow issue in a C++ test. We should have some examples, look for `test/arrow/*.cpp`

herbenderbler:
> Thanks, this looks a lot better. I would really like to see an added test, however. I think it should be possible to recreate the original arrow issue in a C++ test. We should have some examples, look for `test/arrow/*.cpp`

I’ve added a C++ regression under test/arrow/arrow_filter_pushdown.cpp: Arrow filter projection over IN mark join. It builds deterministic name/flag data (matching the flag < 1 + multi-value IN (...) → MARK join recipe), drives it through arrow_scan with produce_arrow_string_view / Arrow 1.5, and asserts on EXPLAIN that we still see a standalone FILTER.  There's no filter pushdown into ARROW_SCAN, MARK in the plan, and no SEMI. There’s also a grouped-result check on the same predicate pattern against the backing table so we’re not only matching plan strings.

The existing sqllogictest test/optimizer/issue_22274_column_lifetime_mark_join.test remains (counts, aggregates, and a FILTER.*Join Type: MARK guard).

lnkuiper:
The added C++ test does not fail without your diff, so it isn't a true test of the problem. I've played around a bit until I got it to break.

Updated `arrow_filter_pushdown.cpp` file:
```C++
#include "catch.hpp"

#include "arrow/arrow_test_helper.hpp"

using namespace duckdb;

// Helper: create an ArrowTestFactory from a query result (all rows, no filter)
static unique_ptr<ArrowTestFactory> MakeArrowFactory(Connection &con, const string &query, bool use_string_view) {
	if (use_string_view) {
		REQUIRE(!con.Query("SET produce_arrow_string_view = true")->HasError());
		REQUIRE(!con.Query("SET arrow_output_version = '1.5'")->HasError());
	} else {
		REQUIRE(!con.Query("SET produce_arrow_string_view = false")->HasError());
		REQUIRE(!con.Query("SET arrow_output_version = '1.0'")->HasError());
	}
	auto client_properties = con.context->GetClientProperties();
	auto result = con.context->Query(query, false);
	REQUIRE(!result->HasError());
	auto types = result->types;
	auto names = result->names;
	return make_uniq<ArrowTestFactory>(std::move(types), std::move(names), std::move(result), false, client_properties,
	                                   *con.context);
}

// Helper: get the EXPLAIN output for an arrow_scan with a filter
static string GetExplainForFilter(Connection &con, ArrowTestFactory &factory, const string &filter_expr) {
	const auto params = ArrowTestHelper::ConstructArrowScan(factory);
	const auto rel = con.TableFunction("arrow_scan", params)->Filter(filter_expr);
	const auto explain_result = rel->Explain();
	REQUIRE(!explain_result->HasError());
	auto &mat = explain_result->Cast<MaterializedQueryResult>();
	return mat.GetValue(1, 0).ToString();
}

// Helper: check for a standalone FILTER operator node in the explain output
static bool StandaloneFilter(const std::string &explain_str) {
	// Look for FILTER as a standalone operator node name (not "Filters:" which is a scan attribute)
	std::string::size_type pos = 0;
	while ((pos = explain_str.find("FILTER", pos)) != std::string::npos) {
		// Check this is the standalone node name, not part of "Filters:"
		if (pos + 6 >= explain_str.size() || explain_str[pos + 6] != 's') {
			return true;
		}
		pos += 6;
	}
	return false;
}

// Helper: check for a scan node with filter pushdown in the explain output
static bool FilterInScan(const std::string &explain_str) {
	auto arrow_pos = explain_str.find("ARROW_SCAN");
	if (arrow_pos == std::string::npos) {
		return false;
	}
	return explain_str.find("Filters:", arrow_pos) != std::string::npos;
}

TEST_CASE("Arrow filter pushdown - view types disable pushdown", "[arrow]") {
	DuckDB db;
	Connection con(db);

	// Create a test table with an INT id column and a VARCHAR name column
	REQUIRE(!con.Query("CREATE TABLE src AS SELECT i AS id, i::VARCHAR AS name FROM range(10) tbl(i)")->HasError());

	SECTION("String view column: filter above scan (not pushed)") {
		auto factory = MakeArrowFactory(con, "SELECT * FROM src", true);
		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
		REQUIRE(StandaloneFilter(explain_str));
		REQUIRE(!FilterInScan(explain_str));
	}

	SECTION("Regular string column: filter pushed into scan") {
		auto factory = MakeArrowFactory(con, "SELECT * FROM src", false);
		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
		REQUIRE(!StandaloneFilter(explain_str));
		REQUIRE(FilterInScan(explain_str));
	}

	SECTION("Integer-only table: filter pushed into scan") {
		auto factory = MakeArrowFactory(con, "SELECT id FROM src", false);
		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
		REQUIRE(!StandaloneFilter(explain_str));
		REQUIRE(FilterInScan(explain_str));
	}
}

TEST_CASE("Arrow filter pushdown - nested view types disable pushdown", "[arrow]") {
	DuckDB db;
	Connection con(db);

	REQUIRE(!con.Query("CREATE TABLE src AS SELECT i AS id, 'val_' || i::VARCHAR AS name FROM range(10) tbl(i)")
	             ->HasError());

	SECTION("Struct containing string_view") {
		auto factory = MakeArrowFactory(con, "SELECT id, {'s': name} AS nested FROM src", true);
		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
		REQUIRE(StandaloneFilter(explain_str));
		REQUIRE(!FilterInScan(explain_str));
	}

	SECTION("List containing string_view") {
		auto factory = MakeArrowFactory(con, "SELECT id, [name] AS names FROM src", true);
		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
		REQUIRE(StandaloneFilter(explain_str));
		REQUIRE(!FilterInScan(explain_str));
	}
}

TEST_CASE("Arrow filter projection over IN mark join (issue 22274)", "[arrow]") {
	DuckDB db;
	Connection con(db);
	REQUIRE(!con.Query("PRAGMA threads=1")->HasError());

	REQUIRE(!con.Query(
	             "CREATE TABLE src AS SELECT "
	             "['TOYOTA','HONDA','FORD','CHEVROLET','BMW','MERCEDES','NISSAN','HYUNDAI','KIA','SUBARU',"
	             "'MAZDA','VPG','ISUZU','SMART','LUCID','LOTUS','PLYMOUTH','DODGE_MITS','VINFAST',"
	             "'POLESTAR'][((i % 20) + 1)::INTEGER] AS name, "
	             "CASE WHEN i % 10 != 0 THEN 0::TINYINT ELSE 1::TINYINT END AS flag "
	             "FROM range(50000) tbl(i)")
	             ->HasError());

	auto explain_factory = MakeArrowFactory(con, "SELECT name, flag FROM src", true);
	auto explain_str = GetExplainForFilter(con, *explain_factory,
	                                       "flag < 1 AND name IN ('VPG','ISUZU','SMART','LUCID','LOTUS','POLESTAR')");

	REQUIRE(StandaloneFilter(explain_str));
	REQUIRE(!FilterInScan(explain_str));
	REQUIRE(explain_str.find("MARK") != string::npos);
	REQUIRE(explain_str.find("SEMI") == string::npos);

	auto query_factory = MakeArrowFactory(con, "SELECT name, flag FROM src", true);
	ArrowStreamParameters parameters;
	auto stream = ArrowTestFactory::CreateStream(reinterpret_cast<uintptr_t>(query_factory.get()), parameters);
	REQUIRE(duckdb_arrow_scan(reinterpret_cast<duckdb_connection>(&con), "data",
	                          reinterpret_cast<duckdb_arrow_stream>(&stream->arrow_array_stream)) == DuckDBSuccess);
	auto result = con.Query("SELECT name, COUNT(*) AS cnt FROM data "
	                        "WHERE flag < 1 AND name IN ('VPG','ISUZU','SMART','LUCID','LOTUS','POLESTAR') "
	                        "GROUP BY 1 ORDER BY 1");
	REQUIRE(!result->HasError());
	auto &mat = result->Cast<MaterializedQueryResult>();
	REQUIRE(mat.RowCount() == 6);

	const vector<string> expected_names {"ISUZU", "LOTUS", "LUCID", "POLESTAR", "SMART", "VPG"};
	for (idx_t row_idx = 0; row_idx < expected_names.size(); row_idx++) {
		REQUIRE(mat.GetValue(0, row_idx).ToString() == expected_names[row_idx]);
		REQUIRE(mat.GetValue(1, row_idx).ToString() == "2500");
	}
}
```
The interesting thing is that this test still fails even with the fix in this PR, meaning that the problem has not yet been resolved. I was only able to reproduce in a `reldebug`/`release` build, not a `debug` build. Please have another look.

herbenderbler:
> The added C++ test does not fail without your diff, so it isn't a true test of the problem. I've played around a bit until I got it to break.
> 
> Updated `arrow_filter_pushdown.cpp` file:
> 
> ```c++
> #include "catch.hpp"
> 
> #include "arrow/arrow_test_helper.hpp"
> 
> using namespace duckdb;
> 
> // Helper: create an ArrowTestFactory from a query result (all rows, no filter)
> static unique_ptr<ArrowTestFactory> MakeArrowFactory(Connection &con, const string &query, bool use_string_view) {
> 	if (use_string_view) {
> 		REQUIRE(!con.Query("SET produce_arrow_string_view = true")->HasError());
> 		REQUIRE(!con.Query("SET arrow_output_version = '1.5'")->HasError());
> 	} else {
> 		REQUIRE(!con.Query("SET produce_arrow_string_view = false")->HasError());
> 		REQUIRE(!con.Query("SET arrow_output_version = '1.0'")->HasError());
> 	}
> 	auto client_properties = con.context->GetClientProperties();
> 	auto result = con.context->Query(query, false);
> 	REQUIRE(!result->HasError());
> 	auto types = result->types;
> 	auto names = result->names;
> 	return make_uniq<ArrowTestFactory>(std::move(types), std::move(names), std::move(result), false, client_properties,
> 	                                   *con.context);
> }
> 
> // Helper: get the EXPLAIN output for an arrow_scan with a filter
> static string GetExplainForFilter(Connection &con, ArrowTestFactory &factory, const string &filter_expr) {
> 	const auto params = ArrowTestHelper::ConstructArrowScan(factory);
> 	const auto rel = con.TableFunction("arrow_scan", params)->Filter(filter_expr);
> 	const auto explain_result = rel->Explain();
> 	REQUIRE(!explain_result->HasError());
> 	auto &mat = explain_result->Cast<MaterializedQueryResult>();
> 	return mat.GetValue(1, 0).ToString();
> }
> 
> // Helper: check for a standalone FILTER operator node in the explain output
> static bool StandaloneFilter(const std::string &explain_str) {
> 	// Look for FILTER as a standalone operator node name (not "Filters:" which is a scan attribute)
> 	std::string::size_type pos = 0;
> 	while ((pos = explain_str.find("FILTER", pos)) != std::string::npos) {
> 		// Check this is the standalone node name, not part of "Filters:"
> 		if (pos + 6 >= explain_str.size() || explain_str[pos + 6] != 's') {
> 			return true;
> 		}
> 		pos += 6;
> 	}
> 	return false;
> }
> 
> // Helper: check for a scan node with filter pushdown in the explain output
> static bool FilterInScan(const std::string &explain_str) {
> 	auto arrow_pos = explain_str.find("ARROW_SCAN");
> 	if (arrow_pos == std::string::npos) {
> 		return false;
> 	}
> 	return explain_str.find("Filters:", arrow_pos) != std::string::npos;
> }
> 
> TEST_CASE("Arrow filter pushdown - view types disable pushdown", "[arrow]") {
> 	DuckDB db;
> 	Connection con(db);
> 
> 	// Create a test table with an INT id column and a VARCHAR name column
> 	REQUIRE(!con.Query("CREATE TABLE src AS SELECT i AS id, i::VARCHAR AS name FROM range(10) tbl(i)")->HasError());
> 
> 	SECTION("String view column: filter above scan (not pushed)") {
> 		auto factory = MakeArrowFactory(con, "SELECT * FROM src", true);
> 		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
> 		REQUIRE(StandaloneFilter(explain_str));
> 		REQUIRE(!FilterInScan(explain_str));
> 	}
> 
> 	SECTION("Regular string column: filter pushed into scan") {
> 		auto factory = MakeArrowFactory(con, "SELECT * FROM src", false);
> 		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
> 		REQUIRE(!StandaloneFilter(explain_str));
> 		REQUIRE(FilterInScan(explain_str));
> 	}
> 
> 	SECTION("Integer-only table: filter pushed into scan") {
> 		auto factory = MakeArrowFactory(con, "SELECT id FROM src", false);
> 		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
> 		REQUIRE(!StandaloneFilter(explain_str));
> 		REQUIRE(FilterInScan(explain_str));
> 	}
> }
> 
> TEST_CASE("Arrow filter pushdown - nested view types disable pushdown", "[arrow]") {
> 	DuckDB db;
> 	Connection con(db);
> 
> 	REQUIRE(!con.Query("CREATE TABLE src AS SELECT i AS id, 'val_' || i::VARCHAR AS name FROM range(10) tbl(i)")
> 	             ->HasError());
> 
> 	SECTION("Struct containing string_view") {
> 		auto factory = MakeArrowFactory(con, "SELECT id, {'s': name} AS nested FROM src", true);
> 		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
> 		REQUIRE(StandaloneFilter(explain_str));
> 		REQUIRE(!FilterInScan(explain_str));
> 	}
> 
> 	SECTION("List containing string_view") {
> 		auto factory = MakeArrowFactory(con, "SELECT id, [name] AS names FROM src", true);
> 		auto explain_str = GetExplainForFilter(con, *factory, "id > 5");
> 		REQUIRE(StandaloneFilter(explain_str));
> 		REQUIRE(!FilterInScan(explain_str));
> 	}
> }
> 
> TEST_CASE("Arrow filter projection over IN mark join (issue 22274)", "[arrow]") {
> 	DuckDB db;
> 	Connection con(db);
> 	REQUIRE(!con.Query("PRAGMA threads=1")->HasError());
> 
> 	REQUIRE(!con.Query(
> 	             "CREATE TABLE src AS SELECT "
> 	             "['TOYOTA','HONDA','FORD','CHEVROLET','BMW','MERCEDES','NISSAN','HYUNDAI','KIA','SUBARU',"
> 	             "'MAZDA','VPG','ISUZU','SMART','LUCID','LOTUS','PLYMOUTH','DODGE_MITS','VINFAST',"
> 	             "'POLESTAR'][((i % 20) + 1)::INTEGER] AS name, "
> 	             "CASE WHEN i % 10 != 0 THEN 0::TINYINT ELSE 1::TINYINT END AS flag "
> 	             "FROM range(50000) tbl(i)")
> 	             ->HasError());
> 
> 	auto explain_factory = MakeArrowFactory(con, "SELECT name, flag FROM src", true);
> 	auto explain_str = GetExplainForFilter(con, *explain_factory,
> 	                                       "flag < 1 AND name IN ('VPG','ISUZU','SMART','LUCID','LOTUS','POLESTAR')");
> 
> 	REQUIRE(StandaloneFilter(explain_str));
> 	REQUIRE(!FilterInScan(explain_str));
> 	REQUIRE(explain_str.find("MARK") != string::npos);
> 	REQUIRE(explain_str.find("SEMI") == string::npos);
> 
> 	auto query_factory = MakeArrowFactory(con, "SELECT name, flag FROM src", true);
> 	ArrowStreamParameters parameters;
> 	auto stream = ArrowTestFactory::CreateStream(reinterpret_cast<uintptr_t>(query_factory.get()), parameters);
> 	REQUIRE(duckdb_arrow_scan(reinterpret_cast<duckdb_connection>(&con), "data",
> 	                          reinterpret_cast<duckdb_arrow_stream>(&stream->arrow_array_stream)) == DuckDBSuccess);
> 	auto result = con.Query("SELECT name, COUNT(*) AS cnt FROM data "
> 	                        "WHERE flag < 1 AND name IN ('VPG','ISUZU','SMART','LUCID','LOTUS','POLESTAR') "
> 	                        "GROUP BY 1 ORDER BY 1");
> 	REQUIRE(!result->HasError());
> 	auto &mat = result->Cast<MaterializedQueryResult>();
> 	REQUIRE(mat.RowCount() == 6);
> 
> 	const vector<string> expected_names {"ISUZU", "LOTUS", "LUCID", "POLESTAR", "SMART", "VPG"};
> 	for (idx_t row_idx = 0; row_idx < expected_names.size(); row_idx++) {
> 		REQUIRE(mat.GetValue(0, row_idx).ToString() == expected_names[row_idx]);
> 		REQUIRE(mat.GetValue(1, row_idx).ToString() == "2500");
> 	}
> }
> ```
> 
> The interesting thing is that this test still fails even with the fix in this PR, meaning that the problem has not yet been resolved. I was only able to reproduce in a `reldebug`/`release` build, not a `debug` build. Please have another look.

Thank you for your 🦅 👀 and for your improved arrow_filter_pushdown.cpp setup.  That made the gap obvious.  My earlier C++ case wasn’t a real regression, even though it passed without the fix and your version is obviously the right direction.

I’ll take another pass with your test as the baseline and follow up once that’s pinned down.

PR #22384, opened 2026-04-30
Add `ArgProperties` for stats forward-bounds

J-Meyers:
Adds per-argument `ArgProperties` metadata on `ScalarFunction` and uses it in `StatisticsPropagator::PropagateExpression(BoundFunctionExpression&)` to derive output min/max via corner evaluation — `f(lo_args)` and `f(hi_args)` with decreasing args swapped — for any annotated function that lacks a custom statistics callback. Drops `NegateBindStatistics` since the generic path subsumes it.

Going forward more properties can be built on and added to `ArgProperties`, broken off from #22286 for easier reviewability

## What was unhandled before

Most scalar functions outside a hand-written allowlist had no statistics callback, so `STATS(f(col))` returned `NULL,NULL` and downstream optimizations (impossible-filter elimination, NULL-set inference, range pushdown) saw the function as a black box.

```sql
CREATE TABLE m(x DOUBLE);
INSERT INTO m VALUES (1.5), (9.7);

-- baseline:
SELECT s.min, s.max FROM (SELECT STATS(floor(x)) s FROM m LIMIT 1);
-- Binder Error:
-- Cannot extract field 'min' from expression "s" because it is not a struct, union, map, or json


-- with this PR:
-- min=1.0, max=9.0
```

Stats implications:

```sql
EXPLAIN SELECT count(*) FROM m WHERE floor(x) > 100;
-- baseline: SEQ_SCAN -> UNGROUPED_AGGREGATE
-- this PR : EMPTY_RESULT -> UNGROUPED_AGGREGATE
```

Same change unlocks `negate`, `ceil`, `round`, `trunc`, `abs`, `epoch_*`, `date_trunc`, `time_bucket`, `make_date`, plus arithmetic `+`/`-`/unary-`-` against bounded inputs.

## Design

```cpp
struct ArgProperties {
	Monotonicity monotonicity = Monotonicity::UNKNOWN;

	ArgProperties &StrictlyIncreasing();
	ArgProperties &NonDecreasing();
	ArgProperties &StrictlyDecreasing();
	ArgProperties &NonIncreasing();
	ArgProperties &Constant();
};

// Annotation site:
ScalarFunctionSet set("floor");
set.AddFunction(...);
set.SetUnaryArgProperties(ArgProperties().NonDecreasing());  // floor is non-strict — many-to-one
```

## Custom stats callbacks take precedence

A function with a registered statistics callback keeps it; the generic forward-bounds path only fires as a fallback. This matters for integer `+`/`-` and decimal arithmetic, which use `PropagateNumericStats` to do something the generic path **cannot**: when the input ranges prove no overflow is possible in the output type, the callback **rewrites the bound function** at plan time to the non-overflow-checking variant (`AddOperator` instead of `AddOperatorOverflowCheck`), removing per-row overflow detection from the hot loop. The forward-bounds path computes output stats only — it has no analogous bind-time rewrite — so we keep both paths and let the existing callback win.

## What remains unannotated

`*`, `/`, `%`, `pow`, etc. are deliberately not annotated:

Annotating these incorrectly would propagate wrong bounds to downstream optimizations. Leaving them unannotated keeps today's behavior exactly. A future PR can add richer per-arg metadata (e.g. preimage callbacks for non-monotonic-but-piecewise-monotonic functions) when there's a use case worth the complexity.

```sql
SELECT era('infinity'::TIMESTAMPTZ);  -- NULL
-- propagator sees out_hi=NULL and returns nullptr
```

- `TIME +/- INTERVAL` and `TIMETZ +/- INTERVAL` are unannotated (24h wrap breaks monotonicity).
- `DATE +/- INTERVAL`, `TIMESTAMP +/- INTERVAL` annotate the date/timestamp arg only — the INTERVAL arg is unannotated because `INTERVAL '29 days' < INTERVAL '1 month'` (interval-normalize) does not match calendar arithmetic (`DATE '2025-02-01' + 29d = 2025-03-02` but `+ 1mo = 2025-03-01`).

## Out of scope

- Per-arg `ArgProperties` on `BoundCastInfo` (cast-side analogue) — separate PR, mechanical mirror of this work.
- The peel optimizer that consumes `injective` and `requires_finite_input` — separate PR.
- ICU year-of-era refiner — lands with the peel PR where it has a consumer.

### DIFF
--- a/src/include/duckdb/function/arg_properties.hpp
+++ b/src/include/duckdb/function/arg_properties.hpp
@@ -1,0 +1,58 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// duckdb/function/arg_properties.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/common.hpp"
+
+namespace duckdb {
+
+//! Monotonicity of a function in one argument (other args held constant).
+enum class Monotonicity : uint8_t {
+	UNKNOWN = 0,
+	CONSTANT,
+	NON_DECREASING,
+	STRICTLY_INCREASING,
+	NON_INCREASING,
+	STRICTLY_DECREASING,
+};
+
+//! Per-argument metadata for a scalar function.
+struct ArgProperties {
+	Monotonicity monotonicity = Monotonicity::UNKNOWN;
+
+	//! Default is non-strict (NON_DECREASING / NON_INCREASING) — the conservative claim.
+	//! Pass `true` only when distinct inputs are guaranteed to map to distinct outputs.
+	ArgProperties &Increasing(bool strict = false) {
+		monotonicity = strict ? Monotonicity::STRICTLY_INCREASING : Monotonicity::NON_DECREASING;
+		return *this;
+	}
+	ArgProperties &Decreasing(bool strict = false) {
+		monotonicity = strict ? Monotonicity::STRICTLY_DECREASING : Monotonicity::NON_INCREASING;
+		return *this;
+	}
+	ArgProperties &Constant() {
+		monotonicity = Monotonicity::CONSTANT;
+		return *this;
+	}
+};
+
+constexpr bool IsMonotonicIncreasing(Monotonicity m) {
+	return m == Monotonicity::NON_DECREASING || m == Monotonicity::STRICTLY_INCREASING;
+}
+constexpr bool IsMonotonicDecreasing(Monotonicity m) {
+	return m == Monotonicity::NON_INCREASING || m == Monotonicity::STRICTLY_DECREASING;
+}
+constexpr bool IsKnownMonotonic(Monotonicity m) {
+	return m != Monotonicity::UNKNOWN;
+}
+constexpr bool IsStrict(Monotonicity m) {
+	return m == Monotonicity::STRICTLY_INCREASING || m == Monotonicity::STRICTLY_DECREASING;
+}
+
+} // namespace duckdb
--- a/src/include/duckdb/function/arg_properties.hpp
+++ b/src/include/duckdb/function/arg_properties.hpp
@@ -26,14 +26,20 @@ enum class Monotonicity : uint8_t {
 struct ArgProperties {
 	Monotonicity monotonicity = Monotonicity::UNKNOWN;
 
-	//! Default is non-strict (NON_DECREASING / NON_INCREASING) — the conservative claim.
-	//! Pass `true` only when distinct inputs are guaranteed to map to distinct outputs.
-	ArgProperties &Increasing(bool strict = false) {
-		monotonicity = strict ? Monotonicity::STRICTLY_INCREASING : Monotonicity::NON_DECREASING;
+	ArgProperties &StrictlyIncreasing() {
+		monotonicity = Monotonicity::STRICTLY_INCREASING;
 		return *this;
 	}
-	ArgProperties &Decreasing(bool strict = false) {
-		monotonicity = strict ? Monotonicity::STRICTLY_DECREASING : Monotonicity::NON_INCREASING;
+	ArgProperties &NonDecreasing() {
+		monotonicity = Monotonicity::NON_DECREASING;
+		return *this;
+	}
+	ArgProperties &StrictlyDecreasing() {
+		monotonicity = Monotonicity::STRICTLY_DECREASING;
+		return *this;
+	}
+	ArgProperties &NonIncreasing() {
+		monotonicity = Monotonicity::NON_INCREASING;
 		return *this;
 	}
 	ArgProperties &Constant() {
--- a/src/optimizer/statistics/expression/propagate_function.cpp
+++ b/src/optimizer/statistics/expression/propagate_function.cpp
@@ -1,7 +1,133 @@
 #include "duckdb/optimizer/statistics_propagator.hpp"
+
+#include "duckdb/execution/expression_executor.hpp"
+#include "duckdb/function/arg_properties.hpp"
+#include "duckdb/planner/expression/bound_constant_expression.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/storage/statistics/numeric_stats.hpp"
 
 namespace duckdb {
+
+static bool TryEvaluateAtConstants(ClientContext &context, const BoundFunctionExpression &func,
+                                   const vector<Value> &arg_values, Value &result) {
+	vector<unique_ptr<Expression>> children;
+	children.reserve(arg_values.size());
+	for (auto &v : arg_values) {
+		children.push_back(make_uniq<BoundConstantExpression>(v));
+	}
+	auto bind_info_clone = func.bind_info ? func.bind_info->Copy() : nullptr;
+	BoundFunctionExpression clone(func.GetReturnType(), func.function, std::move(children), std::move(bind_info_clone),
+	                              func.is_operator);
+	return ExpressionExecutor::TryEvaluateScalar(context, clone, result);
+}
+
+//! Evaluate `func` at the lo/hi corner of each child's value range to derive output min/max.
+//! Decreasing args are swapped so f(lo_args) and f(hi_args) bracket the output range.
+static unique_ptr<BaseStatistics> TryPropagateMonotoneBounds(ClientContext &context, BoundFunctionExpression &func,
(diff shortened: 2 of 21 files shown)


src/include/duckdb/function/arg_properties.hpp
```diff
@@ -0,0 +1,58 @@
...
+	STRICTLY_DECREASING,
+};
+
+//! Per-argument metadata for a scalar function.
+struct ArgProperties {
+	Monotonicity monotonicity = Monotonicity::UNKNOWN;
+
+	//! Default is non-strict (NON_DECREASING / NON_INCREASING) — the conservative claim.
+	//! Pass `true` only when distinct inputs are guaranteed to map to distinct outputs.
+	ArgProperties &Increasing(bool strict = false) {
```
Mytherin: `bool` arguments are an anti-pattern from a readability perspective - can we maybe make this separate functions? 

`StrictlyIncreasing()`
`NonDecreasing()`
`StrictlyDecreasing()`
`NonIncreasing()`

src/optimizer/statistics/expression/propagate_function.cpp
```diff
@@ -1,8 +1,134 @@
...
+			return false;
+		}
+	};
+	if (is_unusable(out_lo) || is_unusable(out_hi)) {
+		return nullptr;
+	}
+	// Ordering violation indicates the annotation is wrong (function isn't actually monotonic in
+	// the direction claimed). Assert in debug builds so this surfaces during testing; bail in
+	// release builds so a misannotation can't propagate unsound stats.
+	D_ASSERT(!(out_hi < out_lo));
```
Mytherin: Can we turn this into an `InternalException`?

src/optimizer/statistics/expression/propagate_function.cpp
```diff
@@ -1,8 +1,134 @@
...
+static unique_ptr<BaseStatistics> TryPropagateMonotoneBounds(ClientContext &context, BoundFunctionExpression &func,
+                                                             const vector<BaseStatistics> &child_stats) {
+	if (!func.function.HasArgProperties() || func.children.empty()) {
+		return nullptr;
+	}
+	if (func.function.GetStability() != FunctionStability::CONSISTENT) {
+		return nullptr;
+	}
+	if (func.function.GetNullHandling() != FunctionNullHandling::DEFAULT_NULL_HANDLING) {
+		return nullptr;
```
Mytherin: Should we bail here? If a function has monotonic arguments but introduces `NULL` for some of them I don't think that's actually a problem, no? Outside of needing to always set `StatsInfo::CAN_HAVE_NULL_VALUES` in this scenario.

src/optimizer/statistics/expression/propagate_function.cpp
```diff
@@ -1,8 +1,134 @@
...
+	}
+	if (BaseStatistics::GetStatsType(func.GetReturnType()) != StatisticsType::NUMERIC_STATS ||
```
Mytherin, line 34: Should this be limited to numeric stats? Although I'm kind of struggling to think of string functions that would be monotonic - maybe `upper / lower`? But that's probably not the case with unicode handling either :)

src/optimizer/statistics/expression/propagate_function.cpp
```diff
@@ -1,8 +1,134 @@
...
+	}
+	// NaN/NULL at a corner means the input was NaN/NULL (column contains NaN, or year(±infinity));
+	// the function may still be monotonic on its finite domain, but no bounds derivable here.
+	const auto is_unusable = [](const Value &v) {
+		if (v.IsNull()) {
+			return true;
+		}
+		switch (v.type().id()) {
+		case LogicalTypeId::DOUBLE:
+			return Value::IsNan(v.GetValue<double>());
```
Mytherin, line 102: Worth noting that this is a choice that's made to make more functions able to have more floating point functions have these monotonically increasing. It would be valid to consider `NAN` here as well since they have a well defined order in DuckDB (larger than all other values) but that would make functions like `Negate` no longer monotonically decreasing (since `-nan` is still `nan`, and therefore still the largest value in the domain).

Mytherin (commented):
Thanks! Looks good - this is a cool idea. @Maxxen could you also have a look. Left some comments

Maxxen:
Functions are currently undergoing a large refactor to split into separate "static" and "runtime (bound)" representations in order to facilitate more flexible type resolution and argument binding. As part of that change we will be able to "annotate" function parameters with additional metadata. I've only really thought of adding "names" and const-ness/default values, but I can see how stats semantics like this could also make sense. However, I'm not convinced of this approach. Do we really need to declare these properties statically so that they are known before binding? 

I agree that the current stats propagating callback is a bit unergonomic, but I think it would make more sense to add helper functions either there or in the `scalar_function_bind_t` callback once the argument types are actually resolved, e.g something like.
```c++
static unique_ptr<FunctionData> IntegerNegateBind(BindScalarFunctionInput &input) { 
  // ...
  input.GetArgument(0).SetProperty(Monotonic::Increasing());
  input.GetArgument(1).SetProperty(Monotonic::Decreasing();
}
```

Or maybe we can just make the stats propagation callback more ergonomic so that it is easier to propagate stats in general without needing a per-argument abstraction. 

I also wonder about the type of properties/what other properties could be useful, or if this is only really useful for arithmetic functions. Increasing/Decreasing makes sense for numeric types - but what about other types like `VARCHAR`/`GEOMETRY`/`ARRAY`/extension types? Or does increasing/decreasing in this sense follow our existing sort-order (as defined by ORDER BY)?

Mytherin:
> Or maybe we can just make the stats propagation callback more ergonomic so that it is easier to propagate stats in general without needing a per-argument abstraction.

We already have two uses for this property: stats propagation and [min/max pushdown on stats](https://github.com/duckdb/duckdb/pull/22286). I can also easily see more purposes for it (e.g. partitioned execution over `ORDER BY [function]`, row group min/max pruning over `[function] = x`, etc)

I think these are just useful properties to know about functions during optimization besides stats propagation.

> I agree that the current stats propagating callback is a bit unergonomic, but I think it would make more sense to add helper functions either there or in the scalar_function_bind_t callback once the argument types are actually resolved, e.g something like.

What's the benefit of defining this in a bind callback versus defining it up-front in the function signature? I think one of our goals should be to have *fewer* callbacks per function, not *more* callbacks per function. This would necessitate every function that wants to define these properties to *also* define a bind callback.

> I also wonder about the type of properties/what other properties could be useful, or if this is only really useful for arithmetic functions. Increasing/Decreasing makes sense for numeric types - but what about other types like VARCHAR/GEOMETRY/ARRAY/extension types? Or does increasing/decreasing in this sense follow our existing sort-order (as defined by ORDER BY)?

It should follow the existing sort order yes - although I struggle to think of functions that maintain this sort order outside of numeric functions.

Maxxen:
Ok, I can see that too. But I think we should allow modification of these args in the bind callback too (just like we allow modifications/specializations/narrowing of arguments in general).

Like I said this should be able to snap neatly into the function parameter rework/abstraction I'm working on anyway, but this can also go in first and I will adapt to it later.

J-Meyers:
- `bool strict` → replaced `Increasing(bool)`/`Decreasing(bool)` with explicit `StrictlyIncreasing()`, `NonDecreasing()`, `StrictlyDecreasing()`, `NonIncreasing()`
- `D_ASSERT` + silent bail → `InternalException`
- Null-handling bail removed; `output_can_have_null` now initialises to `true` for `SPECIAL_HANDLING` so bounds still propagate for those functions

I wasn't sure of monotonic non numeric stats and how they might be handled so felt best to limit it for now, then possibly expand in the future once I had concrete examples to test with.

I updated the comment around NaN handling explaining it as a deliberate choice. I find reasoning around NaN rather annoying and easy to get wrong potentially introducing bugs, I also expect it to be rare in real data and not give particularly meaningful ranges. As well as immediately limit the number of functions we can use right now.

In the future I planned on adding a `refine_monotonicity` with type `Monotonicity (*)(const BoundFunctionExpression &expr, idx_t arg_idx, const Value &lo, const Value &hi) ` that would allow more fine grained handling, given a very specific input range for domain limited functions like `abs`, `*`, `/` but that adds a bit of complexity and wanted to just get the base functionality in that can be agreed upon first.

I think these args can be modified in the bind callback via `GetBoundFunction()` combined with `SetArgProperties` although there may be tricks with this related to functions with custom serialization

Mytherin:
Thanks!

PR #22385, opened 2026-04-30
Limit choco output, and show rclone command

smvv:
### Note: `pip install requests` error is not related

The installation error is fixed in this [PR](https://github.com/duckdb/duckdb/pull/22371).

This is the error in CI:
```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
[retry] attempt 3/3 failed (exit code: 1) for: pip install requests
```

### DIFF
--- a/scripts/extension-upload-single.sh
+++ b/scripts/extension-upload-single.sh
@@ -79,7 +79,7 @@ fi
 if ! command -v rclone >/dev/null 2>&1; then
     case "$(uname -s)" in
       MINGW*|MSYS*|CYGWIN*)
-        choco install rclone -y
+        choco install rclone -y --limit-output --no-progress
         ;;
       *)
         install_runner=(bash)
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #22386, opened 2026-04-30
Downcasting decimal fix incorrect out of range error

Dtenwolde:
Fixes https://github.com/duckdblabs/appian/issues/87

There was a bug in the downcasting of `DECIMAL` which caused an incorrect `out of range` error to be thrown. 
For example the following would throw an error: 
```sql
SELECT '9.90'::DECIMAL(4,3)::DECIMAL(3,2);
```
While `9.90` can be represented in a `DECIMAL(3,2)`.
The bug was caused by looking at the source's scale rather than the difference in scale, calculated here:
https://github.com/duckdb/duckdb/blob/bbe027da7d10d542e831df9ad5cb291059860ecd/src/function/cast/decimal_cast.cpp#L171

In the query above (9.90 is represented as 9900 internally), the source scale is 3 (incorrect), while the scale difference is 1
This is the logic used to check if we can round (omitting negatation for a second here): 
```cpp
auto value = input % divisor;
auto rounded_input = input;
...
if (value >= divisor / 2) {
    rounded_input += divisor;
}
return rounded_input < data.limit && rounded_input > -data.limit;
```
**Incorrect case** (source scale is 3): 
```
divisor = 10^3 = 1000
value = 9900 % 1000 = 900
value >= divisor / 2: 900 >= 500 -> True
  rounded_input += divisor = 9900 + 1000 = 10900
10900 < limit (10000) -> return False
Throws out of range exception
```
**Correct case** (scale difference / factor = 1): 
```
divisor = 10^1 = 10
value = 9900 % 10 = 0
value >= divisor / 2: 0 >= 5 -> False
9900 < limit (10000) -> return True
Passes correctly
```

### DIFF
--- a/src/function/cast/decimal_cast.cpp
+++ b/src/function/cast/decimal_cast.cpp
@@ -117,7 +117,7 @@ struct DecimalScaleDownOperator {
 // This function detects if we can scale a decimal down to another.
 template <class INPUT_TYPE>
 static bool CanScaleDownDecimal(INPUT_TYPE input, DecimalScaleInput<INPUT_TYPE> &data) {
-	int64_t divisor = UnsafeNumericCast<int64_t>(NumericHelper::POWERS_OF_TEN[data.source_scale]);
+	int64_t divisor = UnsafeNumericCast<int64_t>(data.factor);
 	auto value = input % divisor;
 	auto rounded_input = input;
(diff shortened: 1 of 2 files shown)


Mytherin (approved):
Thanks! LGTM after CI passes

PR #21917, opened 2026-04-07
Fix double evicted data decrement

dentiny:
Hi team, when I was reading the BufferManager code I found evicted data is decremented twice for some time.

Right above the PR code change, `DeleteTemporaryFile` is called; if we take a look at the function evicted data is already decremented
- https://github.com/duckdb/duckdb/blob/48163c9b4cf4ace0c2ae93ea3f7017f8168d7f46/src/storage/standard_buffer_manager.cpp#L595
- https://github.com/duckdb/duckdb/blob/48163c9b4cf4ace0c2ae93ea3f7017f8168d7f46/src/storage/standard_buffer_manager.cpp#L603

I ran CI on my own fork: https://github.com/dentiny/duckdb/pull/85, failures are unrelated.

PR #22394, opened 2026-04-30
Fix double decrement of evicted_data_per_tag in .block read-back

krleonid:
## Summary

`StandardBufferManager::ReadTemporaryBuffer` (variable-sized `.block` path) called `DeleteTemporaryFile`, which itself decrements `evicted_data_per_tag[tag]`, and then decremented the **same** counter a second time. Each `.block` read-back therefore subtracted `2 × eviction_size` from the per-tag uint64 counter, underflowing it; `ClampReportedMemory` then pinned the per-tag `temporary_storage_bytes` reported by `duckdb_memory()` to `0` even while `.block` files remained on disk.

The fix removes the duplicate decrement in `ReadTemporaryBuffer` so `DeleteTemporaryFile` (called on the line above) is the single source of truth for the per-tag bookkeeping, matching how it works for the destructor path (`~BlockMemory`).

## Why this matters

- `duckdb_memory().temporary_storage_bytes` for `OVERFLOW_STRINGS` (and any other tag whose buffers go through the `.block` path) misreports `0` once the bug fires — making the view useless for capacity tracking and potentially misleading any logic that consumes it.
- The corrupted counter does not cause data loss on its own (file deletion happens correctly via `DeleteTemporaryFile`), but it breaks any feedback loop that relies on per-tag temporary-storage accounting.

## Tests

Added two slow regression tests under `test/sql/outofcore/`:

- **`block_file_evicted_bytes_accounting.test_slow`** — minimal invariant: with a 50 MB budget and ~57 MB of incompressible OVERFLOW_STRINGS data, asserts that whenever `.block` files exist on disk, `OVERFLOW_STRINGS.temporary_storage_bytes > 0`.
- **`block_file_double_decrement.test_slow`** — explicit write → full-scan → write cycle, asserts the per-tag counter recovers after a fresh wave of evictions following a full read-back scan.

Both tests **fail before the fix** (counter clamped to 0 while files exist) and **pass after the fix** (13/13 and 20/20 assertions).

## Test plan
- [x] `build/release/test/unittest test/sql/outofcore/block_file_evicted_bytes_accounting.test_slow` — passes
- [x] `build/release/test/unittest test/sql/outofcore/block_file_double_decrement.test_slow` — passes
- [x] `build/release/test/unittest test/sql/outofcore/evicted_memory_bytes_updates.test_slow` — passes (existing HASH_TABLE / `.tmp` path test, unaffected by the change)
- [x] Verified the new tests deterministically fail without the fix (10/11 and 11/12 assertions before the regression assertion fires)

### DIFF
--- a/src/storage/standard_buffer_manager.cpp
+++ b/src/storage/standard_buffer_manager.cpp
@@ -563,10 +563,9 @@ unique_ptr<FileBuffer> StandardBufferManager::ReadTemporaryBuffer(QueryContext c
 	handle.reset();
 
 	// Delete the file and return the buffer.
+	// DeleteTemporaryFile already decrements evicted_data_per_tag for the .block path; do not
+	// decrement again here or the counter underflows on every read-back.
 	DeleteTemporaryFile(block.GetMemory());
-
-	// Decrement evicted size.
-	evicted_data_per_tag[uint8_t(tag)] -= buffer->AllocSize();
 
 	return buffer;
 }
--- a/test/sql/outofcore/block_file_double_decrement.test_slow
+++ b/test/sql/outofcore/block_file_double_decrement.test_slow
@@ -1,0 +1,85 @@
+# name: test/sql/outofcore/block_file_double_decrement.test_slow
+# description: Stricter regression test for the .block read-back double-decrement bug.
+#              Forces an explicit write→full-scan→write cycle and asserts that the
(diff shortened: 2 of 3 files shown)


dentiny:
Thanks for catching and sorry for the confusion, I should've targeted my fix (https://github.com/duckdb/duckdb/pull/21917) on v1.5

Mytherin:
Thanks!

PR #22398, opened 2026-04-30
Allow colon aliasing in `RENAME` and `REPLACE`

Dtenwolde:
Inspired by https://github.com/duckdb/duckdb/discussions/21677

In the `REPLACE` and `RENAME` clauses, we only permitted aliasing with an explicit `AS`, while in many other places we allow omitting the `AS` or have the colon-style aliasing (`alias : expr`). This PR also adds these two alias stylings to the `REPLACE` and `RENAME` clauses: 

```sql
memory D create table integers(i bigint, j bigint); 
memory D SELECT * RENAME i: renamed_col FROM integers;
┌─────────────┬───────┐
│ renamed_col │   j   │
│    int64    │ int64 │
└─────────────┴───────┘
memory D INSERT INTO integers values (1,2); 
memory D SELECT * REPLACE i: i*2 FROM INTEGERS; 
┌───────┬───────┐
│   i   │   j   │
│ int64 │ int64 │
├───────┼───────┤
│     2 │     2 │
└───────┴───────┘
memory D SELECT * REPLACE i*2 i FROM INTEGERS; 
┌───────┬───────┐
│   i   │   j   │
│ int64 │ int64 │
├───────┼───────┤
│     2 │     2 │
└───────┴───────┘
memory D SELECT * REPLACE (i: i*2, j: j*100) FROM INTEGERS; 
┌───────┬───────┐
│   i   │   j   │
│ int64 │ int64 │
├───────┼───────┤
│     2 │   200 │
└───────┴───────┘
```
An alias is still required for `REPLACE`: 
```sql
memory D SELECT * REPLACE i FROM integers; 
Parser Error:
Expected an alias for REPLACE expression
memory D SELECT * REPLACE (i) FROM integers; 
Parser Error:
Expected an alias for REPLACE expression
```
Of course using `AS` remains possible: 
```sql
memory D SELECT * RENAME i AS k FROM integers; 
┌───────┬───────┐
│   k   │   j   │
│ int64 │ int64 │
├───────┼───────┤
│     1 │     2 │
└───────┴───────┘
memory D SELECT * REPLACE i * 3 AS i FROM integers; 
┌───────┬───────┐
│   i   │   j   │
│ int64 │ int64 │
├───────┼───────┤
│     3 │     2 │
└───────┴───────┘

```

Happy to hear feedback :)


Mytherin:
Thanks for the PR!

I think not having the brackets for replace make it a bit confusing, e.g. this:

```sql
SELECT * REPLACE i: i*2, j: j + 10 FROM INTEGERS;
```

Is this parsed as:

```sql
SELECT * REPLACE (i: i*2), j: j + 10
```

Or as:

```sql
SELECT * REPLACE (i: i*2, j: j + 10)
```

I'm guessing the former - because if it's the latter we can never have other expressions after the `* REPLACE` - but that's not intuitively readable imo - so I would leave the bracket requirement.

Adding support for `x: ` aliases is great of course.

Dtenwolde:
Ah right, that's a good catch. I got a bit confused, but there also seems to be unexpected behaviour with the PostgreSQL parser (v1.5.2)
```sql
DuckDB v1.5.2 (Variegata)
memory D create table integers(i bigint, j bigint);
memory D INSERT INTO integers values (1,2);
memory D call disable_peg_parser(); -- to be sure
┌─────────┐
│ success │
│ boolean │
└─────────┘
  0 rows
memory D SELECT * REPLACE i+1 AS i, j+1 AS j FROM integers;
┌───────┬───────┬───────┐
│   i   │   j   │   j   │
│ int64 │ int64 │ int64 │
├───────┼───────┼───────┤
│     2 │     2 │     3 │
└───────┴───────┴───────┘
```
Where this query returns an extra `j` it seems (`SELECT i+1, j, j+1 FROM integers`)

So yes, it gets parsed as: `SELECT * REPLACE (i: i*2), j: j + 10`

I think removing the single, no-bracket case is best since it can easily lead to confusion imo.

Mytherin:
Ah - I missed that this used to work in the old parser. Given this used to work I think it should keep working also in the new parser. We could consider deprecating it but we should do that gradually (i.e. not just break the behavior in the new parser).

Mytherin:
At least the case without colon aliases, we can require brackets for the case with colon aliases since that never worked in the old parser.

Dtenwolde:
Okay, that makes more sense haha

I'll revert back the old behaviour!

PR #22402, opened 2026-04-30
AFTER trigger firing for INSERT (FOR EACH ROW)

EtgarDev:
## Summary
Implements `AFTER INSERT FOR EACH ROW` trigger firing using a bulk body transformation approach.

Rather than executing the trigger body once per row, the body is transformed at bind time to operate over `__duckdb_trigger_base` CTE, containing all rows inserted by the triggering statement.
This preserves per-row semantics while executing in a single bulk pass.

## Transformation
`trigger_body_transformer` rewrites the trigger body at two points:
1. `CREATE TRIGGER` - validates the body is transformable (by injecting a dummy base)
2. Firing - transforms the body to reference `__duckdb_trigger_base` DML CTE

For `INSERT` bodies, we:
1. Rewrite `NEW.col` -> `__duckdb_trigger_base.col`
2. Replace `FROM new` with `FROM __duckdb_trigger_base`
3. Convert `VALUES (...)` bodies to `SELECT ... FROM __duckdb_trigger_base`

## Blocked at `CREATE TRIGGER` (future work)
1. AFTER UPDATE/DELETE FOR EACH ROW (requires `OLD` row mechanism)
2. `OLD` references in trigger bodies
3. Non-`INSERT` trigger bodies
4. Set operations (`UNION/INTERSECT/EXPECT`) in `INSERT` bodies
5. Multi row `VALUES`

## Demo
image

```
memory D EXPLAIN ANALYZE INSERT INTO my_table VALUES (1, 'hello'), (2, 'world');
┌─────────────────────────────────────┐
│┌───────────────────────────────────┐│
││    Query Profiling Information    ││
│└───────────────────────────────────┘│
└─────────────────────────────────────┘
EXPLAIN ANALYZE INSERT INTO my_table VALUES (1, 'hello'), (2, 'world');
│┌──────────────────────────────────────────────┐│
││              Total Time: 0.0778s             ││
│└──────────────────────────────────────────────┘│
┌───────────────────────────┐
│           QUERY           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      EXPLAIN_ANALYZE      │
│    ────────────────────   │
│                           │
│           0 rows          │
│           0.00s           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│            CTE            │
│    ────────────────────   │
│         CTE Name:         │
│   __duckdb_trigger_base   │
│                           │
│       Table Index: 0      ├──────────────┐
│           0 rows          │              │
│           0.00s           │              │
│         PROJECTION        ││            CTE            │
│             id            ││         CTE Name:         │
│                           ││     __duckdb_trigger_1    │
│                           ││       Table Index: 1      ├──────────────┐
│                           ││                           │              │
│           0.00s           ││                           │              │
│                           ││                           │              │
│           2 rows          ││           0 rows          │              │
│          (0.00s)          ││           0.00s           │              │
└─────────────┬─────────────┘└─────────────┬─────────────┘              │
┌─────────────┴─────────────┐┌─────────────┴─────────────┐┌─────────────┴─────────────┐
│           INSERT          ││           INSERT          ││    UNGROUPED_AGGREGATE    │
│    ────────────────────   ││    ────────────────────   ││    ────────────────────   │
│                           ││                           ││        Aggregates:        │
│                           ││                           ││        count_star()       │
│           0.00s           ││           0.00s           ││                           │
│                           ││                           ││                           │
│                           ││                           ││                           │
│           2 rows          ││           1 row           ││           1 row           │
│          (0.00s)          ││          (0.00s)          ││           0.00s           │
└─────────────┬─────────────┘└─────────────┬─────────────┘└─────────────┬─────────────┘
┌─────────────┴─────────────┐┌─────────────┴─────────────┐┌─────────────┴─────────────┐
│      COLUMN_DATA_SCAN     ││          CTE_SCAN         ││          CTE_SCAN         │
│                           ││    ────────────────────   ││    ────────────────────   │
│                           ││        CTE Index: 0       ││        CTE Index: 0       │
│           0.00s           ││                           ││                           │
│                           ││                           ││                           │
│                           ││                           ││                           │
│           2 rows          ││           2 rows          ││           2 rows          │
│          (0.00s)          ││           0.00s           ││           0.00s           │
└───────────────────────────┘└───────────────────────────┘└───────────────────────────┘
```


lnkuiper:
Exciting stuff! I wanted to try this out but got this compile error:
```
/Users/laurens/git/duckdb/main-duckdb/src/planner/binder/statement/trigger_body_transformer.cpp:21:13: error: 'type' is a protected member of 'duckdb::BaseExpression'
   21 |                 if (expr->type != ExpressionType::COLUMN_REF) {
      |                           ^
/Users/laurens/git/duckdb/main-duckdb/src/include/duckdb/parser/base_expression.hpp:86:17: note: declared protected here
   86 |         ExpressionType type;
      |                        ^
In file included from /Users/laurens/git/duckdb/main-duckdb/build/debug/src/planner/binder/statement/ub_duckdb_bind_statement.cpp:8:
/Users/laurens/git/duckdb/main-duckdb/src/planner/binder/statement/trigger_body_transformer.cpp:70:22: error: 'type' is a protected member of 'duckdb::BaseExpression'
   70 |                 if (found || expr->type != ExpressionType::COLUMN_REF) {
      |                                    ^
/Users/laurens/git/duckdb/main-duckdb/src/include/duckdb/parser/base_expression.hpp:86:17: note: declared protected here
   86 |         ExpressionType type;
      |                        ^
2 errors generated.
```

EtgarDev:
> Exciting stuff! I wanted to try this out but got this compile error:
> 
> ```
> /Users/laurens/git/duckdb/main-duckdb/src/planner/binder/statement/trigger_body_transformer.cpp:21:13: error: 'type' is a protected member of 'duckdb::BaseExpression'
>    21 |                 if (expr->type != ExpressionType::COLUMN_REF) {
>       |                           ^
> /Users/laurens/git/duckdb/main-duckdb/src/include/duckdb/parser/base_expression.hpp:86:17: note: declared protected here
>    86 |         ExpressionType type;
>       |                        ^
> In file included from /Users/laurens/git/duckdb/main-duckdb/build/debug/src/planner/binder/statement/ub_duckdb_bind_statement.cpp:8:
> /Users/laurens/git/duckdb/main-duckdb/src/planner/binder/statement/trigger_body_transformer.cpp:70:22: error: 'type' is a protected member of 'duckdb::BaseExpression'
>    70 |                 if (found || expr->type != ExpressionType::COLUMN_REF) {
>       |                                    ^
> /Users/laurens/git/duckdb/main-duckdb/src/include/duckdb/parser/base_expression.hpp:86:17: note: declared protected here
>    86 |         ExpressionType type;
>       |                        ^
> 2 errors generated.
> ```

@lnkuiper oh I see that it became protected very recently (https://github.com/duckdb/duckdb/commit/6d69fb1bb18763d34f1a666eb0e16bc719f9cbaa) so my rebase onto main broke things.
Anyway I'm working on the final touch-ups today and this will also be fixed :)

lnkuiper (commented):
Thanks for the PR! Really cool stuff :)

I have played around with this a bit and I think this is broken in a few ways because the behaviour is not true `FOR EACH ROW` behaviour but somewhere between row variables (e.g., `NEW.id`) and transition tables (`REFERENCING .. AS`).

## Things that break

1. `NEW.col` fails when used in `SELECT ... FROM VALUES`

```sql
CREATE TABLE t(id INTEGER, val VARCHAR);
CREATE TABLE audit(tid INTEGER, msg VARCHAR);

CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit SELECT * FROM VALUES (NEW.id, 'inserted');

INSERT INTO t VALUES (1, 'a'), (2, 'b');
```
Gives:
```
Binder Error: Referenced table `"__duckdb_trigger_base"` not found!
```

The transformer rewrites `NEW.id` to `__duckdb_trigger_base.id`, but because the body already has a FROM clause, it does not add `__duckdb_trigger_base` to the query's FROM.

2. `NEW.col` fails in `SELECT`

```sql
CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit SELECT NEW.id, 'inserted';
CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit SELECT NEW.id, 'inserted' FROM range(10);
```

Gives:
```
Binder Error: Referenced table "__duckdb_trigger_base" not found!
```

3. `FROM new` works, but probably shouldn't (also, what if we do `CREATE TABLE main.new ...`? This will conflict!)

```sql
CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit SELECT count(*), 'cnt' FROM new;

INSERT INTO t VALUES (1, 'a'), (2, 'b'), (3, 'c');
SELECT * FROM audit;
```

This is transition-table behavior, not scalar row-trigger behavior. PostgreSQL does not expose new as an implicit table. It requires `REFERENCING NEW TABLE AS ...`
This PR should probably reject `FROM new` for now, or pivot to implementing explicit transition tables.

4. `LIMIT`/window/aggregate clauses expose set-wise semantics through `FROM new` (not row-wise semantics!)

```sql
CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit SELECT NEW.id, 'x' FROM new LIMIT 1;

INSERT INTO t VALUES (1, 'a'), (2, 'b'), (3, 'c');
```

This only inserts one row into `audit`, even though we inserted 3 rows into `t`

5. `ON CONFLICT DO UPDATE` fires the insert trigger for updated rows

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY, val VARCHAR);
CREATE TABLE audit(tid INTEGER, msg VARCHAR);

CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit VALUES (NEW.id, 'inserted');

INSERT INTO t VALUES (1, 'a');

INSERT INTO t VALUES (1, 'replace'), (2, 'b')
ON CONFLICT DO UPDATE SET val = excluded.val;

SELECT * FROM audit ORDER BY tid;
```

Current result:
```
1 inserted
1 inserted
2 inserted
```

Expected PostgreSQL/SQLite-compatible behavior:
```
1 inserted
2 inserted
```

The conflict-updated row should fire update triggers, not insert triggers.

## Potential fixes

I think @Mytherin should probably also chime in here, as he's knowledgeable about this part of the code.

A big issue is in `src/planner/binder/statement/trigger_body_transformer.cpp:19`. `RewriteNewColumnRefs` rewrites `NEW.id` into a qualified column reference against `__duckdb_trigger_base`, but the transformer only sometimes adds that CTE to the body's `FROM`.

A better direction would be:

- Treat scalar `NEW.col` support separately from transition-table support.
- Do not rewrite `FROM new` to `__duckdb_trigger_base` unless explicit transition-table syntax exists.
- For now, reject `FROM new` with a clear “transition tables are not supported yet” error.
- Make `NEW.col` work by injecting the base CTE in a way that preserves per-row semantics.

For the current bulk strategy, that likely means transforming supported trigger bodies into a query that always ranges over the base CTE exactly once per inserted row.

These should both
```sql
INSERT INTO audit VALUES (NEW.id, 'inserted')
INSERT INTO audit SELECT NEW.id, 'inserted'
```
Become this
```sql
INSERT INTO audit
SELECT base.id, 'inserted'
FROM __duckdb_trigger_base base
```

For trigger bodies that already have a `FROM`, the implementation needs to decide whether to support lateral-style correlation. If supported:

```sql
INSERT INTO audit SELECT NEW.id, r.range::VARCHAR FROM range(10) r
```

would need to become something like:

```sql
INSERT INTO audit
SELECT base.id, r.range::VARCHAR
FROM __duckdb_trigger_base base, range(10) r
```

That is semantically “execute the body for each base row”. That is probably too much for this PR, so you can reject bodies with `NEW` references plus `FROM` clauses until a correlated rewrite exists.

For name resolution, `ReplaceNewTableRef` should not rewrite arbitrary `BaseTableRef` whose `table_name == "new"` if it has a catalog/schema qualifier. But I think the better fix is to remove this rewrite entirely for row triggers, because implicit `FROM new` should not be supported.

For ON CONFLICT, the base CTE in `src/planner/binder.cpp:619` should include only rows that actually fired the event type. `INSERT ... RETURNING *` currently returns both inserted and conflict-updated rows for `ON CONFLICT DO UPDATE`. That means the trigger expansion needs either event-specific returned row sets from the insert binder/executor, or it should reject `ON CONFLICT DO UPDATE` on tables with row insert triggers until that distinction exists.

I think we should either take a conservatie path for this PR and reject a lot of the things that are causing issues, or we should think about a lateral-style correlated rewrite. @kryonix might have some thoughts on that.

Mytherin:
One thing to consider here is that these rewrites will be very hard to do at the parse-tree level for anything except for the simplest cases. For example, whether or not a function is an aggregate is not really known in the parse tree. The semantics of this query changes depending on whether or not `random_function` is an aggregate:

```sql
CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit SELECT random_function(NEW.id)
```

e.g. these two are completely different in behavior and how we need to handle the pushdown:

```sql
CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit SELECT cos(NEW.id)

CREATE TRIGGER trg AFTER INSERT ON t FOR EACH ROW
  INSERT INTO audit SELECT sum(NEW.id)
```

I think it might make sense to rethink how we do the planning so that we plan the logical operator tree first, and then do the pushdown / rewrite on the logical operator tree, similar to how this is done for correlated subqueries. Then we can handle the rewrite operator-by-operator. Otherwise you would have to essentially do a "sort of planning" on a parse tree which will end up getting very ugly / error-prone very quickly.

Mytherin:
For reference, what we do with correlated subqueries is we generate a plan node that we don't know how to execute - `LogicalDependentJoin`. We could generate a similar plan node here - e.g. `LogicalTrigger`. The idea is then that we either (1) get rid of the logical trigger during planning, allowing successful execution of the plan, or (2) throw an error if this is not possible (due to e.g. unsupported operators).

An example of what this might look like could be something like:

```sql
CREATE TABLE t (id INTEGER, val VARCHAR);
CREATE TABLE audit (tid INTEGER, msg VARCHAR);
CREATE TRIGGER trg_new_col AFTER INSERT ON t FOR EACH ROW
    INSERT INTO audit VALUES (NEW.id, 'inserted');

INSERT INTO t VALUES (1, 'a');
```

```

┌───────────────────────────┐
│     LOGICAL_TRIGGER       ├──────────────┐
│    ────────────────────   │              │
│      AFTER EACH ROW       │              │
│           INSERT          ││           INSERT          │
│             t             ││            audit          │
└─────────────┬─────────────┘└─────────────┬─────────────┘
│      COLUMN_DATA_SCAN     ││         PROJECTION        │
│           ~1 row          ││     NEW.id, 'inserted'    │
└───────────────────────────┘└─────────────┬─────────────┘
                             ┌─────────────┴─────────────┐
                             │         DUMMY_SCAN        │
                             └───────────────────────────┘

```

Basically we plan the trigger query completely separately from the main query. We then try to do the pushdown at the logical plan level, and end up with this in this scenario:

```sql
┌───────────────────────────┐
│            CTE            │
│    ────────────────────   │
│       CTE Name: cte       │
│       Table Index: 0      ├──────────────┐
│           ~1 row          │              │
│           INSERT          ││           INSERT          │
│             t             ││            audit          │
│                           │└─────────────┬─────────────┘
└─────────────┬─────────────┘┌─────────────┴─────────────┐
┌─────────────┴─────────────┐│         PROJECTION        │
│      COLUMN_DATA_SCAN     ││    ────────────────────   │
│    ────────────────────   ││             id            │
│                           ││         'inserted'        │
│           ~1 row          ││           ~1 row          │
└───────────────────────────┘└─────────────┬─────────────┘
                             ┌─────────────┴─────────────┐
                             │          CTE_SCAN         │
                             │    ────────────────────   │
                             │        CTE Index: 0       │
                             │                           │
                             │           ~1 row          │
                             └───────────────────────────┘
```

Mytherin:
For subqueries this happens in `FlattenDependentJoins::DecorrelateIndependent` in `planner.cpp`:

```cpp
	if (this->plan) {
		auto max_tree_depth = Settings::Get<MaxExpressionDepthSetting>(context);
		CheckTreeDepth(*plan, max_tree_depth);

		this->plan = FlattenDependentJoins::DecorrelateIndependent(*this->binder, std::move(this->plan));
	}
```

We could add another line there for triggers.

kryonix:
Thanks for pinging me @lnkuiper. We can think about a lateral rewrite, for sure. But I think @Mytherin got it spot in with his review. Doing such a rewrite that early in the query execution chain will very likely result in issues later on. It reminds me a bit of how CTE inlining/materialization used to work—also way too early. We can do much more powerful transformations if we postpone the heavy-lifting to the logical plan layer. Because if we have a `LogicalTrigger` operator, we can rewrite that in any way we need to, if necessary, it could even become a `DependentJoin/DelimJoin` etc. Also, we may be able to omit adding too many magic names to the system. For me, that's always kind of risky.

EtgarDev:
Thanks for the detailed feedback @Mytherin @lnkuiper , that is very helpful.
@Mytherin it's a good point that parse-tree rewrites are fragile here. The aggregate case is a good example, `sum(NEW.id)` rewrites to valid SQL that produces wrong semantics, and it seems like `LogicalTrigger` with your suggestion is the right long-term direction.

The question is whether it makes sense to adopt it now, or whether we prefer to keep the current approach for the simple cases, by resolving @lnkuiper 's cases (blocking `FROM new`, tightening the `NEW.col` injection), and also this case by blocking aggregate/window bodies post-binding, and defer `LogicalTrigger` to follow-up once the simpler cases are stable.
One disadvantage I can see in deferring is that all the transformer code written for upcoming `AFTER UPDATE/DELETE` support would be thrown away once we adopt the `LogicalTrigger` approach (rewriting turns into decorrelation).

Mytherin:
Given that we will almost certainly want to move towards the logical plan-level rewrite I would say it makes most sense to make the switch early on - otherwise we'd be doing a bunch of work on code that we know will be thrown away again.

lnkuiper:
Agree with Mark, I think taking it slow and implementing the decorrelation approach will be the winning strategy in the long run. This means we can do `FOR EACH ROW` semantics without actually executing a subquery for each row, similar to how we execute correlated subqueries, which avoids quadratic complexities.

I think that what you've implemented here pretty is close to `FOR EACH STATEMENT` with `REFERENCING NEW TABLE AS` semantics, which I do think will have to be implemented in the transform/bind layer, which we also want to implement.

EtgarDev:
@Mytherin I agree.
I'll explore the references you pointed to and get to understand how I can make something similar here.

EtgarDev:
@lnkuiper As for your last point - do you think we can reuse the current approach (transformer) for the `FOR EACH STATEMENT` mechanics? Or we should maybe use the `LogicalTrigger` approach as a baseline for all kind of triggers execution?

lnkuiper:
@EtgarDev the current trigger execution doesn't need `LogicalTrigger`, and I think that `FOR EACH STATEMENT` with `REFERENCING NEW TABLE AS` doesn't need to either. There is no correlation. I think we will only need the `LogicalTrigger` decorrelation approach with `FOR EACH ROW`. I would consider adapting the current PR to implement `FOR EACH STATEMENT` + `REFERENCING NEW TABLE AS` instead, as you have built some of the machinery for that already, then it doesn't go to waste. In the meantime, we can let the idea of `LogicalTrigger` + decorrelation marinate in our heads a bit :)

EtgarDev:
@lnkuiper That makes sense :) I'll pivot the PR to implement `FOR EACH STATEMENT + REFERENCING NEW TABLE AS` and keep `FOR EACH ROW` blocked until we digest the `LogicalTrigger` approach

PR #22408, opened 2026-04-30
Expose RETURNING-referenced columns to catalog extensions for write-side projection pushdown

rustyconover:
## Summary

Adds a `vector<column_t> returning_referenced_columns` field to `LogicalInsert`, `LogicalUpdate`, `LogicalDelete`, and `LogicalMergeInto`, populated in `Binder::BindReturning` from the bound RETURNING projection.

This lets catalog extensions (Iceberg connectors, remote table formats, etc.) implementing `Catalog::PlanInsert` / `PlanUpdate` / `PlanDelete` / `PlanMergeInto` see which target-table columns the user's RETURNING list references, so they can ship only those columns over the wire instead of the full row.

Discussion: #22406

## Implementation

- **Field**: added on the four logical write ops with a doc comment that calls out empty-vector semantics ("no projection info available; treat as a request for all columns") to avoid the foot-gun of an extension treating empty as "send nothing."
- **Binder**: `BindReturning` walks the bound `projection_expressions` for `BoundColumnRefExpression` nodes, filters to the target table's `update_table_index` (so refs from correlated subqueries don't leak in), and stores deduplicated logical column indices on the child write operator. Insertion-ordered.
- **Serialization**: new field in `logical_operator.json` with `default: vector<column_t>()` so older serialized plans deserialize cleanly. `serialize_logical_operator.cpp` regenerated via `make generate-files`.

## Backwards compatibility

The change is purely additive metadata.

- Built-in INSERT/UPDATE/DELETE/MERGE physical operators don't read the new field. Existing dataflow (write op emits full rows; the projection above computes RETURNING) is unchanged.
- Existing catalog extensions that ignore the field continue to receive full rows. They recompile cleanly.
- Extensions that opt in are responsible for coordinating with their backend.

## Test plan

- [x] New sqllogictest `test/sql/returning/returning_referenced_columns.test` (13 assertions) covering INSERT/UPDATE/DELETE/MERGE INTO with single-column, multi-column, `*`, expression, constant-only, and no-RETURNING cases — verified via `json_serialize_plan`.
- [x] Existing `[returning]` suite passes (567 assertions across 8 test cases).
- [x] `make format-fix` clean.

### DIFF
--- a/src/include/duckdb/planner/operator/logical_delete.hpp
+++ b/src/include/duckdb/planner/operator/logical_delete.hpp
@@ -26,6 +26,12 @@ public:
 	bool return_chunk;
 	vector<idx_t> return_columns;
 	vector<unique_ptr<BoundConstraint>> bound_constraints;
+	//! The set of target-table column indices referenced by the RETURNING list.
+	//! Populated by the binder when RETURNING is present so catalog extensions can
+	//! perform write-side projection pushdown for affected-row payloads.
+	//! An empty vector means no projection information is available; treat as a request
+	//! for all columns. Built-in DELETE ignores this field.
+	vector<column_t> returning_referenced_columns;
 
 public:
 	void Serialize(Serializer &serializer) const override;
--- a/src/include/duckdb/planner/operator/logical_insert.hpp
+++ b/src/include/duckdb/planner/operator/logical_insert.hpp
@@ -67,6 +67,12 @@ public:
 	vector<unique_ptr<Expression>> bound_defaults;
 	//! The constraints used by the table
 	vector<unique_ptr<BoundConstraint>> bound_constraints;
+	//! The set of target-table column indices referenced by the RETURNING list.
+	//! Populated by the binder when RETURNING is present so catalog extensions can
+	//! perform write-side projection pushdown for affected-row payloads.
+	//! An empty vector means no projection information is available; treat as a request
+	//! for all columns. Built-in INSERT ignores this field.
+	vector<column_t> returning_referenced_columns;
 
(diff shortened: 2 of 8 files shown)


Mytherin:
Isn't this already present in `return_columns`?

rustyconover:
Hi @Mytherin,

From my looking at the code:

LogicalInsert doesn't have return_columns.
LogicalUpdate doesn't have return_columns.
LogicalMergeInto has delete_return_columns but only for the delete branch.
LogicalDelete has return_columns but it's for a different use, mostly for scan routing from the physical slots to the returning slots in the DataChunk.

Still in all of these operations the engine assumes the resulting DataChunk has every column in the table.  Which is really cheap to do if you're using the local on disk storage, but can be expensive if using storage elsewhere.

Rusty

Mytherin:
Ah yes, you're right, IIRC work on the projection pushdown was started but hasn't been finished yet. 

> Still in all of these operations the engine assumes the resulting DataChunk is every column in the table. Which is really cheap to do if you're using the local on disk storage, but can be expensive if using storage elsewhere.

Agreed, that's the real issue (regardless of if you're using DuckDB with a remote storage or local on-disk this is unnecessarily expensive). It's just a limitation of the current implementation. The correct way forward is addressing this limitation by supporting projection pushdown / elimination in the `RETURNING` clause. This PR is more of work-around for that core issue.

If you want to tackle projection pushdown into `RETURNING` that would be a valuable contribution we'd be happy to merge, I would do it operator-by-operator, however. The work-around in this PR is not desirable.

rustyconover:
@Mytherin great!  Putting it on the queue.

PR #22409, opened 2026-05-01
Add unit tests for Dictionary Vectors

txbm:
[This commit](https://github.com/duckdb/duckdb/commit/a11267dcd155dde53cca73293552c49b2cc6eda1) replaced the optional dictionary sizing with the option to check the child size, creating the possibility for `CanCacheHashes` to return `true` in cases where hashing would cause unsafe memory access.

This was tested by running the included regression test on the [prior commit]( https://github.com/duckdb/duckdb/commit/a22e598fe802647354c60ff5b234591e828b6ca3) and observing it succeed after having run it on and past the commit above and observing it fail.

It is now [fixed here](https://github.com/duckdb/duckdb/commit/82ee5951f23a14e44e497926b9f23ae782d7aade), and this test passes on the current HEAD

For context, this was discovered by way of failing DuckLake tests, but that is no longer relevant. This PR now offers a regression test for dictionary vectors and the associated hashing behavior to prevent future regressions.

### DIFF
--- a/test/common/CMakeLists.txt
+++ b/test/common/CMakeLists.txt
@@ -16,6 +16,7 @@ add_library_unity(
   test_parse_expression.cpp
   test_parse_logical_type.cpp
   test_utf.cpp
+  test_vectors.cpp
   test_allocator_debug_info.cpp
   test_buffer_pool_reservation.cpp
   test_storage_fuzz.cpp
--- a/test/common/test_vectors.cpp
+++ b/test/common/test_vectors.cpp
@@ -1,0 +1,65 @@
+#include "catch.hpp"
+
+#include "duckdb/common/string.hpp"
+#include "duckdb/common/types.hpp"
+#include "duckdb/common/types/vector.hpp"
+#include "duckdb/common/vector/flat_vector.hpp"
+#include "duckdb/common/vector/string_vector.hpp"
+#include "duckdb/common/vector_operations/vector_operations.hpp"
+
+#include <string>
+#include <vector>
+
+using namespace duckdb; // NOLINT
+
+static void FillVarcharVector(Vector &target, const duckdb::vector<string> &values) {
+	auto writer = FlatVector::Writer<string_t>(target, values.size());
+	for (const auto &value : values) {
+		writer.WriteValue(StringVector::AddString(target, value));
+	}
+	FlatVector::SetSize(target, count_t(values.size()));
+}
+
+static duckdb::vector<hash_t> HashVector(Vector &input, idx_t count) {
+	Vector hashes(LogicalType::HASH, count);
+	VectorOperations::Hash(input, hashes, count);
+	auto data = FlatVector::GetData<hash_t>(hashes);
+	return duckdb::vector<hash_t>(data, data + count);
+}
+
+TEST_CASE("Vector hash matches dictionary slice values", "[vector][dictionary][hash]") {
+	constexpr idx_t dictionary_size = 4;
+	constexpr idx_t selection_size = 2;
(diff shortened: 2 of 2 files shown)


Mytherin:
Thanks - but your included test does not seem to trigger any issues (neither on the released version nor on main). Could you perhaps submit a test that triggers this issue? Then we would be happy to look at fixing it. This kind of AI-based static analysis is very often wrong so it's hard to take at face value without a concrete example showcasing the issue.

txbm:
> Thanks - but your included test does not seem to trigger any issues (neither on the released version nor on main). Could you perhaps submit a test that triggers this issue? Then we would be happy to look at fixing it. This kind of AI-based static analysis is very often wrong so it's hard to take at face value without a concrete example showcasing the issue.

Oh neither this nor the related issue were uncovered via AI static analysis - I observed the crashes intermittently in both CI and locally while running the DuckLake filter stress tests - but because they are intermittent I would not expect those tests to deterministically fail on the main/release branches but I am happy to continue trying to find a more deterministic repro, sure.

txbm:
@Mytherin I have updated the test to reproduce the segfault more reliably and scoped the fix to the most narrow conditions I can reliably reproduce.

For convenience, here is a repro-only job I pushed to [demonstrate the segfault](https://github.com/txbm/duckdb/actions/runs/25236075580/job/74002458084).

During the initial attempt of this fix, including the other PR, I no longer observed DuckLake filter stress test failures, and falsely concluded they were both necessary.

I should have performed an ablation test on DuckDB directly prior to the patches instead of relying on the DuckLake test as a proxy. DuckLake forces a more specific code path which improves the odds of the `PlainSelect` path via the `DuckLakeMultiFileReader`. This led me to wrongly assume direct DuckDB tests would not fail reliably anyway.

Having narrowed the reproduction criteria via a more reliable test, I no longer observe either failure against DuckDB core or DuckLake with just this minimal fix. I am closing the other issue, concluding it was a transient artifact of my original repro attempts.

In any-case, this was all reviewed and reproduced with human eyeballs and fingers so if I am still mistaken somewhere I will happily stand corrected. No need to blame it on AI 😄

Mytherin:
Thanks - I can reproduce the issue now, apologies for jumping the gun :) we've been getting a lot of AI-generated "issue reports" that end up not being issues. I'll have a closer look tomorrow.

txbm:
No worries I am happy to help anyway I can so even if my proposed patch isn't the direction you want to see I'd love to learn how to approach it better!

Mytherin:
Thanks - I had another look and it seems this regression was already fixed by https://github.com/duckdb/duckdb/pull/22426 - the test now passes on `main`. Could you verify if that's the case?

txbm:
Hey @Mytherin , thank you for looking into it. I think I have a better understanding now that this was a regression in the dictionary vector data structure itself rather than a mis-use of it by the Parquet column reader as I had previously misunderstood. I confirmed the original segfault is fixed by the referenced commits, and also traced the regression in the dictionary vector, mostly for own learning, but also to provide a properly focused test for future regression protection if that's at all helpful. I've just repurposed this PR to provide that test.

If that's not helpful for some reason then no worries at all, feel free to close, but I can confirm everything works as expected now. Thanks again.

Mytherin:
Thanks - in general we prefer not to add C++ tests unless we have a very good reason to do so, so let's leave it at that.

txbm:
> Thanks - in general we prefer not to add C++ tests unless we have a very good reason to do so, so let's leave it at that.

Ok I didn't know but I'll keep that in mind for the future. Thanks anyway!

PR #8974, opened 2023-09-18
Issue #8917: Timestamp Variant Casts

hawkfish:
Add missing S => MS, NS and MS => NS casts.
Fix/replace previously failing tests.

fixes #8917

PR #22412, opened 2026-05-01
Internal #8974: Nanosecond Timestamp TimeZone

hawkfish:
* Checkpoint commit of code changes to add support for TIMESTAMP_TZ_NS type.
* Fix most test_all_types() tests.
* Fix TIMESTAMP_NS => VARCHAR casts
* Fix generated file correctness.
* Update variant TSTZNS expected results
* Update Python TSTZNS expected results


Mytherin:
Thanks for the PR!

Codewise this looks fine to me, but looking at the amount of code required this does stress to me how much nicer it would be to make the timestamp types more generic :)

hawkfish:
The only remaining test failure says:

> This might fail spuriously if changes to content of test database / generation script happened

which is exactly what happened (`test_all_types() has a new column).

Mytherin:
We shouldn't be generating a new version of the `storage_version.db` anymore since we're now backwards compatible - there's no need to keep on generating new versions of this database. We should now always be able to read this database. I've updated the comments in the code and fixed the test to be robust against newly added types in `test_all_types`.

Mytherin:
Thanks!

alippai:
This is great, thank you!

PR #22415, opened 2026-05-01
Skip filter eval and decode for filter-only columns on always-true row groups

J-Meyers:
For a parquet scan with a filter on a column that is *not* projected (filter-only), if a row group's per-column statistics prove the filter is always-true on that row group, neither the filter eval nor the column's decode contributes anything to the result. This changes the parquet reader to skip both.

`count(*)` needs the same treatment: after `RemoveUnusedColumns` runs, `projection_ids` is empty, but the reader can't tell that apart from the" no projection info, treat every column as projected" case.

I still rely on the `projection_pushdown_done` flag because the assumption `projection_ids[i] < column_ids.size()` is relied upon in a massive number of areas, including effectively every reader and supporting functions like `LogicalGet::ResolveTypes` and `LogicalGet::GetColumnBindings`, so changing it forces a lot of changes.

## Benchmark tests

100M-row / 997-row-group parquet (486 MB on disk). Row group 0 contains a mix of `'a*'` and `'b*'` values; the remaining 996 row groups contain only `'b*'` values.

```sql
COPY (
  SELECT
    -- 20 distinct values heavily repeated → dictionary encoding, near-zero decode cost
    (i % 20)::INTEGER AS id,
    CASE WHEN i < 100000 AND (i % 2 = 0)
         THEN 'a' || lpad(i::VARCHAR, 33, '0')
         ELSE 'b' || lpad(i::VARCHAR, 33, '0')
    END AS s
  FROM range(100000000) t(i)
) TO '/tmp/bench_pr.parquet' (
  ROW_GROUP_SIZE 100000, COMPRESSION 'snappy', OVERWRITE_OR_IGNORE
);

-- 'a' + '8' + 32 zeros (34 chars). Lex above every 'a0…' in row group 0,
-- lex below every 'b…' in row groups 1+, so it lands inside row group 0's
-- stats range and below every later row group's min.
SET VARIABLE k = 'a800000000000000000000000000000000';
```

Hot cache, `parquet_metadata_cache=true`, default thread pool, median of 4 runs after one warm-up.

| Query | main | this PR + #22429 + #22432 | speedup |
|---|---|---|---|
| `SELECT sum(id) FROM file WHERE s != getvariable('k')` | 188 ms | 10 ms | **19×** |
| `SELECT count(*) FROM file WHERE s != getvariable('k')` | 174 ms | 3 ms | **58×** |
| `SELECT count(*) FROM file WHERE s >= 'k'` | 175 ms | <1 ms | **~250×** |

The first two rows are the runtime path firing: 996 of 997 row groups have the predicate always-true, so for those groups we skip filter eval and (because `s` isn't projected) never decode it. The third row needs the string zone-map recovery from #22429 to classify `s >= 'a'` as always-true table-wide; combined with #22432's column-prune, the query collapses to a metadata-only row count.


src/include/duckdb/common/multi_file/multi_file_states.hpp
```diff
@@ -164,6 +164,8 @@ struct MultiFileGlobalState : public GlobalTableFunctionState {
 
 	idx_t max_threads = 1;
 	vector<idx_t> projection_ids;
+	//! True iff projection_ids is authoritative — see BaseFileReader::projection_ids.
+	bool projection_pushdown_done = false;
```
Mytherin: Maybe we can have a `COLUMN_IDENTIFIER_EMPTY` in the `projection_ids` instead to cover this scenario?

Mytherin (commented):
Thanks - cool optimizations, but this is doing a bunch of different things - could we split these into smaller PRs?

PR #22416, opened 2026-05-01
Fix tidy error performance-move-const-arg

smvv:
Merge queue and main are red because of this clang tidy error:
```
Using clangd: Ubuntu clangd version 20.1.2 (0ubuntu1~24.04.2)
src/optimizer/rule/like_optimizations.cpp:146:79: Error: Std::move of the const variable 'function' has no effect; remove std::move() or make the variable non-const [performance-move-const-arg]
  144 |  	// replace LIKE by an optimized function
  145 |  	unique_ptr<Expression> result;
  146 |  	auto new_function = make_uniq<BoundFunctionExpression>(expr.GetReturnType(), std::move(function),
      |                                                                                ^~~~~~~~~~
  147 |  	                                                       std::move(expr.children), nullptr);
  148 |
make: *** [Makefile:669: tidy-check-clangd] Error 1
```

### Add timeout to all install steps

The install step of `Tidy check` took at least 55 min:

Screenshot 2026-05-01 at 09 22 59

so I've added a timeout to fail earlier and avoid wasting more resources.

### DIFF
--- a/.github/workflows/Main.yml
+++ b/.github/workflows/Main.yml
@@ -125,6 +125,7 @@ jobs:
               - tools/juliapkg/**
 
       - name: Check CI
+        timeout-minutes: 10
         if: steps.changed.outputs.github_any_changed == 'true'
         run: |
           echo "::add-matcher::.github/actionlint-matcher.json"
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #22423, opened 2026-05-01
Improved `FILE_SIZE_BYTES` / `ROW_GROUPS_PER_FILE`

lnkuiper:
This PR improves the accuracy of `FILE_SIZE_BYTES` and `ROW_GROUPS_PER_FILE` when many threads write to the same file. This is currently perfectly accurate when only one thread writes to a file (`SET threads=1;` / `PER_THREAD_OUTPUT`), but when many threads write to the same file we can over- or undershoot the target value.

## Current Implementation

When writing Parquet files, most of the time is spent converting DuckDB data to an encoded/compressed Parquet row group. This step is called the `prepare` step. Comparatively, much less time is spent in the `flush` step, which is to append the row group to the file. In the current implementation, if you have to `prepare` a row group for a specific file, and then also flush it to that file. This creates a problem when trying to accurately achieve the target `FILE_SIZE_BYTES` / `ROW_GROUPS_PER_FILE`, as we have to commit to writing a row group to a file before the costly `prepare` step.

Because we have to commit so early, we tend to overshoot the target value. Take this query, for example:
```sql
COPY tpch_sf10.lineitem TO 'my' (FORMAT PARQUET, ROW_GROUPS_PER_FILE 8, OVERWRITE);
```
In the current implementation, when I run this on my laptop with 10 threads, I get 33 files that have up to 13 row groups. This is a pretty big overshoot already, and it get worse the more threads are active.

Then, consider this query:
```sql
COPY tpch_sf10.lineitem TO 'my' (FORMAT PARQUET, FILE_SIZE_BYTES '32MiB', OVERWRITE);
```
The current implementation creates 122 files of 17MB. It undershoots here because we have a "band-aid" kind of solution that tries to estimate the size of an added row group before adding it to the file. However, because this estimate is off by quite a bit (it overestimates the size of the row group), it ends up moving on to the next file before it actually reached the file size.

## This PR

This PR decouples the `prepare` and `flush` steps, meaning that we do `prepare` on any file, and `flush` it to any file. The implementation will actually do all `prepare` against the same file, and then `flush` to arbitrary files. By decoupling we can delay the decision which file to flush a row group to until the very last moment.

Using the same queries as above in this PR yields a very different result. The `ROW_GROUPS_PER_FILE 8` query yields 62 files with 8 row groups, and the `FILE_SIZE_BYTES '32MiB` query yields 62 files of 33MB. This is a lot more accurate, while utilizing threads much more efficiently.

##### Challenge: `VARIANT`

One challenge that I had to overcome in this PR is `VARIANT`. The Parquet specification describes how to "shred" the `VARIANT` type, but requires all row groups to shred `VARIANT` in the same way, as the shredding needs to be defined in the schema. DuckDB handles this by lazily initializing the schema during `prepare`. However, this PR makes it so some files are never `prepare`d against, only `flush`d. I've had to implement schema initialization on `flush` specifically for the `VARIANT` type because of this.

## Benchmark

The implementation in this PR writes files much more accurately in parallel, but that should not sacrifice the performance! We can fully serialize writes and get perfectly accurately files, but that would be very slow. I've ran a little benchmark to show how the execution times of this PR vs. the v1.5.2. The benchmark copies the `lineitem` table from TPC-H at SF 10 to one or more Parquet files.

| Which | Execution Time [s] |
|:-|-:|
| v1.5.2 single file baseline | 2.3 |
| v1.5.2 `PER_THREAD_OUTPUT` baseline | 2.2 |
| v1.5.2 `FILE_SIZE_BYTES '32MiB'` | 2.8 |
| v1.5.2 `ROW_GROUPS_PER_FILE 8` | 2.4 |
| This PR `FILE_SIZE_BYTES '32MiB'` | 2.3 |
| This PR `ROW_GROUPS_PER_FILE 8` | 2.3 |

Using `PER_THREAD_OUTPUT` still yields the best performance, because virtually no locks are involved. However, the single file baseline is only slightly behind, which writes all data to a single file, which of course, involved locking. This PR brings the performance of writing to multiple files with `FILE_SIZE_BYTES` and `ROW_GROUPS_PER_FILE` in line with the performance of writing to a single file, so it is not only more accurate, but also faster than the current implementation.

## Future Work

1. I want to do another pass over the `PhysicalCopyToFile` operator and look for any I/O that happens under a big lock, including things like opening a file. These operations should be done under more fine-grained locks (e.g., a file lock instead of the global state lock). This should improve the performance on many-core machines especially when writing to a remote location.
2. Allowing `FILE_SIZE_BYTES` / `ROW_GROUPS_PER_FILE` for partitioned writes.
2. Sorted writes.

## Fixes

Fixes https://github.com/duckdblabs/duckdb-internal/issues/9076

Spurious failure in partitioned writes was fixed by holding the `active_writes_lock` for the entire duration of finding/creating a file.


extension/parquet/include/parquet_writer.hpp
```diff
@@ -239,6 +245,13 @@ class ParquetWriter {
 
 private:
 	void GatherWrittenStatistics();
+	void InitializeColumnOrders(idx_t unique_columns);
+	void InitializeStatsUnifiers();
+	void InitializeSchemaFromPreparedRowGroup(const PreparedRowGroup &prepared);
+	void InitializeColumnWriters();
+	idx_t InitializeColumnWriterSchemaIndices();
+	PreparedParquetLayout ExportPreparedLayout() const;
+	idx_t CountLeafColumnWriters() const;
```
Tishj: nit: could put these in an `#ifdef DEBUG` block, since they're only used for verification

extension/parquet/parquet_writer.cpp
```diff
@@ -1160,19 +1198,50 @@ void ParquetWriter::InitializeSchemaElements() {
...
+	auto &layout = prepared.layout;
+	if (file_meta_data.schema.empty()) {
+		if (layout.schema.empty()) {
+			throw InternalException("Prepared Parquet row group is missing schema");
+		}
+		if (HasPreparedShreddingTypes(layout.shredding_types)) {
+			shredding_types = layout.shredding_types.Copy();
+			InitializeColumnWriters();
+		}
+		InitializeColumnWriterSchemaIndices();
```
Tishj: nit: could use the same verification as you added elsewhere
`auto unique_columns = ...`
`D_ASSERT(unique_columns == prepared.row_group.columns.size())`

Tishj (approved):
Thanks, I think the schema initialization logic from the PreparedRowGroup looks sound 👍

lnkuiper:
Thanks for the review! Hope CI is green when I check after the weekend

lkzcgfvf:
@lnkuiper thanx for working on this~

Could you also check that this PR fix issue with changing data ordering  ( how rows are sorted in Parquet file). 
Rows in output data files should be sorted the same way as they were in original file.
I assume that in multi thread mode it could do.

lnkuiper:
Hi @lkzcgfvf DuckDB supports preserving order when writing in parallel, but not in all cases. If you do `COPY (FROM 'input.parquet') TO 'output.parquet';` the order is preserved (single output file). This PR doesn't change that in any way. If you copy from an input file to more than one output file (e.g., when enabling `FILE_SIZE_BYTES`), order is not preserved.

PR #22425, opened 2026-05-01
CUMULATIVE_VACUUM_TIME metric

artjomPlaunov:
Add a vacuum timing metric that tracks elapsed time of individual vacuum tasks as well as index rebuilding if the vacuum_rebuild_indexes setting is enabled (The latter is just a current implementation consideration, when vacuuming is supported with indexes by default, the implementation of this metric will also change, but the user facing metric remains the same). 

As a basic first pass, I didn't include vacuum planning/scheduling itself, as I think vacuum tasks and index rebuilding are the most important parts to profile, and to avoid cluttering the checkpointing code with more timers until it becomes necessary. 

With multiple threads and vacuum tasks, the vacuum time could exceed reported checkpointing latency, since we are summing the individual elapsed times for each vacuum task. Whereas if you run it with one thread, you will see closer to the "true" CPU time for the vacuum tasks.


src/storage/checkpoint/table_data_writer.cpp
```diff
@@ -54,6 +57,13 @@ unique_ptr<TaskExecutor> TableDataWriter::CreateTaskExecutor() {
 	return make_uniq<TaskExecutor>(TaskScheduler::GetScheduler(GetDatabase()));
 }
 
+void TableDataWriter::AddCumulativeVacuumTime(idx_t elapsed_nanos) {
```
lnkuiper: Does this need to be a member function of `TableDataWriter`? Can't we just get the profiler from the context at the call site instead? Then we don't pollute `TableDataWriter` with profiling-related stuff

lnkuiper (changes requested):
Thanks! I have one comment then this is ready to go:

src/storage/checkpoint/table_data_writer.cpp
artjomPlaunov: Perhaps we could move it out to a helper in Query Profiler. the main reason I did it here was because we're using the client context store in TableDataWriter, so we would have to expose that

artjomPlaunov: We would have to expose the context from table data writer, and also find a suitable place to place the function (unless we want to duplicate the logic twice at each call site).

lnkuiper: I prefer it moved elsewhere, maybe add a `TryGetClientContext` to `TableDataWriter` that returns the optional ptr and implement the method elsewhere

artjomPlaunov: sounds good!

artjomPlaunov:
as well as the review considerations above, I may have looked over how this affects performance if the metric is not enabled, so still WIP.

artjomPlaunov:
@lnkuiper thanks for the suggestions, I think this is better: I added a helper to get the client context from TableDataWriter. 

I also modified the code to use the existing ActiveTimer RAII wrapper for the clocks, since before I was unconditionally starting/stopping the timer. The wrapper stops the timer if it goes out of scope (or we can explicitly end it ourselves), and does an early return if profiling is not active.

lnkuiper (approved):

artjomPlaunov:
@lnkuiper I removed a flaky test, it was the sanity check that if we have multiple threads, the cumulative vacuum time could exceed checkpoint wall clock time (they could be running in parallel, but we are summing their individual component times up). I removed that since it was failing on linux

PR #22428, opened 2026-05-01
Split `(Bound)Function` inheritance and introduce `Function(Signature/Parameter)`

Maxxen:
- Sever the inheritance between `*Function and `*BoundFunction` classes. You can no longer implicitly create a bound function from a non-bound function
- Move FunctionCallbacks/FunctionProperties into CRTP mixin class to reduce duplication
- Modify aggregate function callbacks to actually take `BoundAggregateFunction` in more places
- Return functions from catalog sets by const reference instead of copying
- Encapsulate additional properties (such as function name)
- Add `FunctionSignature` and `FunctionParameter` classes to the unbound functions.
- Don't pass return type in `BoundFunctionExpression` constructor, derive it from the function itself instead.


Mytherin:
Thanks!

Maxxen:
@Mytherin Got dropped from the merge queue due to conflicts, but green again!

PR #22429, opened 2026-05-01
String stats zone map: FILTER_ALWAYS_TRUE recovery and exact-bound logic

J-Meyers:
Extend StringStats::CheckZonemap to recognize when a filter must be entirely satisfied by every value in the [min, max] domain, returning FILTER_ALWAYS_TRUE so callers can skip the per-row evaluation. 

Previously the directional comparisons (>, >=, <, <=) only returned NO_PRUNING_POSSIBLE or FILTER_ALWAYS_FALSE, leaving the filter in place even when zone-map analysis could prove it always true.

For >= and <=, also use a length-aware exact-bound comparison when parquet stats mark the stored bound as not truncated. This handles cases where the constant shares a prefix with the bound but exceeds it in length (e.g. 's <= xyza' against max='xyz' is always-true).

Split form #22415

### DIFF
--- a/extension/parquet/parquet_reader.cpp
+++ b/extension/parquet/parquet_reader.cpp
@@ -1201,9 +1201,12 @@ static FilterPropagateResult CheckParquetStringFilter(BaseStatistics &stats, con
 				if (constant.value.type().id() == LogicalTypeId::VARCHAR) {
 					auto &min_value = pq_col_stats.min_value;
 					auto &max_value = pq_col_stats.max_value;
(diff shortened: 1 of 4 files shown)


Mytherin:
Thanks - I think this is a great change, but I've pushed some changes to the string stats that should make this easier. See https://github.com/duckdb/duckdb/pull/22649. Could you perhaps merge with that and adjust the changes so that they use the new stats types for this pruning?

PR #22430, opened 2026-05-01
Internal #8947: Builtin Window Batching

hawkfish:
* Convert all the custom windowed aggregates to use the new vectoraised "batch" API;
* Add some missing aggregator capability checks;


Mytherin:
Thanks!

PR #22432, opened 2026-05-01
Prune dead filter-only columns from scans

J-Meyers:
When the optimizer eliminates a filter as always-true via table-level statistics, the column it referenced is left in the scan's `column_ids` and decoded at scan time for no reason. This adds a small pass at physical-plan time that drops columns from `column_ids` that are neither projected nor referenced by any remaining table filter.

Most visible when the dead column is expensive to decode — combined with #22429 skipping the string decode entirely.

There isn't a way to test for this change in behavior right now explicitly, because the EXPLAIN only shows the projected ones, not the decoded ones, benchmarks can clearly show the difference.
  
Split from #22415.

## Benchmark (Reused in other PRs)

```sql
COPY (
  SELECT
    -- 20 distinct values heavily repeated → dictionary encoding, near-zero decode cost
    (i % 20)::INTEGER AS id,
    CASE WHEN i < 100000 AND (i % 2 = 0)
         THEN 'a' || lpad(i::VARCHAR, 33, '0')
         ELSE 'b' || lpad(i::VARCHAR, 33, '0')
    END AS s
  FROM range(100000000) t(i)
) TO '/tmp/bench_pr.parquet' (
  ROW_GROUP_SIZE 100000, COMPRESSION 'snappy', OVERWRITE_OR_IGNORE
);
```

| Query | main | this branch + #22429 | speedup |
|---|---|---|---|
| `SELECT sum(k) FROM file WHERE s >= '0'` | 188 ms | 10 ms | **19×** |

### DIFF
--- a/src/optimizer/optimizer.cpp
+++ b/src/optimizer/optimizer.cpp
@@ -281,6 +281,7 @@ void Optimizer::RunBuiltInOptimizers() {
 	RunOptimizer(OptimizerType::COLUMN_LIFETIME, [&]() {
 		ColumnLifetimeAnalyzer column_lifetime(*this, *plan, true);
 		column_lifetime.VisitOperator(*plan);
+		column_lifetime.FinalizeDeadColumnPrunes();
 	});
 
 	// Once we know the column lifetime, we have more information regarding
@@ -359,6 +360,7 @@ void Optimizer::RunBuiltInOptimizers() {
 	RunOptimizer(OptimizerType::COLUMN_LIFETIME, [&]() {
 		ColumnLifetimeAnalyzer column_lifetime(*this, *plan, true);
 		column_lifetime.VisitOperator(*plan);
+		column_lifetime.FinalizeDeadColumnPrunes();
 	});
 
 	// apply simple expression heuristics to get an initial reordering
--- a/benchmark/micro/parquet/dead_filter_column_prune.benchmark
+++ b/benchmark/micro/parquet/dead_filter_column_prune.benchmark
@@ -1,0 +1,30 @@
+# name: benchmark/micro/parquet/dead_filter_column_prune.benchmark
+# description: Filter is always-true via table stats; the column it referenced should not be decoded.
+# group: [parquet]
+
+name Dead Filter Column Prune
+group micro
+subgroup parquet
+
+require parquet
+
+storage persistent
+
+load
+COPY (
+  SELECT
+    (i % 20)::INTEGER AS id,
+    CASE WHEN i < 100000 AND (i % 2 = 0)
+         THEN 'a' || lpad(i::VARCHAR, 33, '0')
+         ELSE 'b' || lpad(i::VARCHAR, 33, '0')
+    END AS s
+  FROM range(100000000) t(i)
+) TO '${BENCHMARK_DIR}/dead_filter_column_prune.parquet' (
+  ROW_GROUP_SIZE 100000, COMPRESSION 'snappy', OVERWRITE_OR_IGNORE
+);
+
+run
+SELECT sum(id) FROM '${BENCHMARK_DIR}/dead_filter_column_prune.parquet' WHERE s >= '0'
+
+result I
+950000000
--- a/benchmark/micro/parquet/row_group_skip_count.benchmark
+++ b/benchmark/micro/parquet/row_group_skip_count.benchmark
@@ -1,0 +1,30 @@
+# name: benchmark/micro/parquet/row_group_skip_count.benchmark
+# description: count(*) with a filter-only column whose per-row-group stats prove the predicate always-true on most groups.
+# group: [parquet]
+
+name Row Group Skip Filter-Only Count
+group micro
+subgroup parquet
+
+require parquet
+
+storage persistent
+
+load
+COPY (
+  SELECT
+    (i % 20)::INTEGER AS id,
+    CASE WHEN i < 100000 AND (i % 2 = 0)
+         THEN 'a' || lpad(i::VARCHAR, 33, '0')
+         ELSE 'b' || lpad(i::VARCHAR, 33, '0')
+    END AS s
+  FROM range(100000000) t(i)
+) TO '${BENCHMARK_DIR}/row_group_skip_count.parquet' (
+  ROW_GROUP_SIZE 100000, COMPRESSION 'snappy', OVERWRITE_OR_IGNORE
+);
+
+run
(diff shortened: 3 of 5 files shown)


src/execution/physical_plan/plan_get.cpp
```diff
@@ -192,9 +192,50 @@ PhysicalOperator &PhysicalPlanGenerator::CreatePlan(LogicalGet &op) {
 		proj.children.push_back(table_scan);
 		return proj;
 	}
+	// Prune columns that are neither projected nor referenced by any remaining table filter — eliminates
+	// unnecessary decoding when the optimizer removed a filter as always-true via table-level statistics.
+	vector<idx_t> old_to_new;
+	if (!op.projection_ids.empty()) {
```
Mytherin: This feels like the wrong place to do this - this is an optimizer pass that is now effectively run during scanning. Shouldn't this be handled in `RemoveUnusedColumns`?

J-Meyers: Hmm, I've now integrated it into the StatisticsPropagation since that is what actually causes the filters to be removed, it didn't work in RemoveUnusedColumns because the filters weren't yet removed at that point

Mytherin: Maybe the `ColumnLifetimeAnalyzer` is the place to do this? That generates the projection maps

lnkuiper:
Is the benchmark you mention in the description added in a `.benchmark` file in any of the other PRs? I think it would be nice to have it, then our weekly regression test runner will catch regressions on this optimization

J-Meyers:
I've moved it to `ColumnLifetimeAnalyzer` and added the two benchmark tests.

The first benchmark `benchmark/micro/parquet/dead_filter_column_prune.benchmark` will only work once the PR for string stats FILTER_ALWAYS_TRUE is in and this PR is in, on string columns is where there is the biggest benefit from this group of PRs because decoding them costs the most.

The second benchmark `benchmark/micro/parquet/row_group_skip_count.benchmark` will see a benefit once the main PR is in that allows individual row group skipping (and also requires the string stats pr)

src/optimizer/optimizer.cpp
```diff
@@ -281,6 +281,7 @@ void Optimizer::RunBuiltInOptimizers() {
 	RunOptimizer(OptimizerType::COLUMN_LIFETIME, [&]() {
 		ColumnLifetimeAnalyzer column_lifetime(*this, *plan, true);
 		column_lifetime.VisitOperator(*plan);
+		column_lifetime.FinalizeDeadColumnPrunes();
```
Mytherin, line 284: Is this separate pass necessary? Can this not be integrated with the rest of the column lifetime analyzer?

J-Meyers: I tried that at first and couldn't get it to work cleanly without forcing larger changes. I was running into the issue that if we ended up pruning any columns and then remapping the projection ids mid recursion, then later the `lhs_unused` wont match up with the later `op.children[0]->GetColumnBindings()`.

I can make it not appear within optimizer.cpp by switching to the following which basically checks if it is top level, then only executes it then, but I haven't yet come up with a good way to actually do the removal in the middle of the main recursion:
```cpp
if (owned_dead_gets) {
    FinalizeDeadColumnPrunes();
}
```

Part of this not matching up with the existing infrastructure is because the filter only columns only ever exist in the `column_ids` not in the `projection_ids` which means the `GenerateProjectionMap` doesn't have an explicit set to map against, and the way the earlier snapshotting of children's bindings prior to recursing leads to issues.

It is possible to simplify it significantly and integrate it into the recursion if instead of actually removing unused `column_ids` instead overwriting them with `GetAnyColumn`, but then there is still some (albeit minor) cost being paid on table functions that support virtual columns but not `COLUMN_IDENTIFIER_EMPTY` , but some might not have any virtual columns that are cheap.

Mytherin: Looking at the code again I actually wonder - wouldn't we achieve the same by running `RemoveUnusedColumns` again post-statistics propagation? It feels like the code is more or less just re-implementing that optimizer partially. Maybe we can just call that optimizer a second time?

J-Meyers: I tried this, but it doesn't compose with the rest of the pipeline. The targeted ColumnLifetimeAnalyzer change in this PR mutates only `LogicalGet.column_ids`  (plus projection_ids/table_filters) and fixes upstream bindings via a single `ColumnBindingReplacer.VisitOperator(root)` — it operates on a plan in a known state and touches one kind of state in one direction.                                                                                                                                                                                                                                                                                                                                  

A second `RemoveUnusedColumns` pass instead re-runs the full column-pruning algorithm over a plan whose state has already been mutated by the intermediate passes (`COLUMN_LIFETIME` wrote integer projection_maps; `COMMON_SUBPLAN` introduced `LOGICAL_MATERIALIZED_CTE`; equality substitutions already happened). The structural problem is that the integer `projection_map` indexes carry no record of which column they point to, so renumbering scan `column_ids` silently invalidates them, and there's no easy local way to detect or repair it. 

In particular:                                                                                                                                                                                                                                                                                                                                                                                         
  * Clearing projection_maps before the second pass fixes asof / relation_api binding-resolver failures, but `TOP_N_WINDOW_ELIMINATION` reads them as semantically meaningful (e.g. `right_projection_map` non-empty as the way to find the row_number column on a comparison join).
  * Re-running `COLUMN_LIFETIME` to regenerate them fixes qualify, but produces different maps than were originally there, breaking IEJoin layouts with `prefer_range_joins=true`.
  *  Validate-and-clear (drop maps whose indexes are out of range) catches qualify but not asof: a 2-column GET pruned to 1 column leaves the map's surviving index in range, just pointing at a different column.
  * `RemoveUnusedColumns`'s `LOGICAL_MATERIALIZED_CTE` handler is also order-dependent — the first pass runs before `COMMON_SUBPLAN`, so it never saw a CTE; running it post-COMMON_SUBPLAN exposes a latent bug where positional propagation through set-ops makes it think fewer CTE columns are referenced than actually are.

Mytherin: Thanks for the investigation. Optimizers should ideally be re-entrant and should be able to fire at any stage of the plan. The fact that that doesn't work is a bug that should be fixed imo. I think it would be better to make sure `RemoveUnusedColumns` works also post-column-lifetime than to introduce what is effectively a "sort-of-remove-unused-columns" optimizer.

PR #4430, opened 2022-08-17
Linux ARM binaries

samansmink:
This PR is unfortunately still a draft. 

It attempts to add binaries to the CI for the main duckdb executable and the JDBC driver.

The compilation is done through cross compiling whereas QEMU is used for running the tests. The basics seems to work as i have had the tests succeed on a aws box, i'm still running into some CI issues which i haven't figured out yet.

PR #22435, opened 2026-05-02
Fix duck fuzz #4430

dentiny:
Closes https://github.com/duckdb/duckdb-fuzzer/issues/4430

The issue is clear from the error message "Database name db1 was already used by a different database for this meta transaction", which means [`used_databases`](https://github.com/duckdb/duckdb/blob/89f51782ee12353646ff98d525d95de8ed498a8e/src/include/duckdb/transaction/meta_transaction.hpp#L100-L101) are not cleared on database detach

### DIFF
--- a/src/main/database_manager.cpp
+++ b/src/main/database_manager.cpp
@@ -2,6 +2,7 @@
 
 #include "duckdb/catalog/catalog.hpp"
 #include "duckdb/catalog/catalog_search_path.hpp"
+#include "duckdb/common/exception/transaction_exception.hpp"
 #include "duckdb/main/attached_database.hpp"
 #include "duckdb/main/client_data.hpp"
 #include "duckdb/main/database.hpp"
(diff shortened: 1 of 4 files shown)


Mytherin:
Thanks - but I'm not so sure this is the right solution here. This means we have outstanding changes for a given database within the transaction, but are not keeping the database alive. I would have to think about it more but this definitely has the potential of introducing life-time issues.

Mytherin:
It might just be that the right solution is to not allow detaching a database in a transaction that the current transaction has changes to (i.e. the detach would fail in this scenario).

src/main/database_manager.cpp
```diff
@@ -250,14 +251,34 @@ void DatabaseManager::DetachDatabase(ClientContext &context, const string &name,
...
+	}
 	if (!attached_db) {
 		if (if_not_found == OnEntryNotFound::THROW_EXCEPTION) {
 			throw BinderException("Failed to detach database with name \"%s\": database not found", name);
 		}
 		return;
 	}
 
+	// Reject detach if current transaction already has outstanding work on the database.
+	if (attached_db->GetVisibility() != AttachVisibility::HIDDEN) {
```
Mytherin: `HIDDEN` is a property that can also be set by the user - so this doesn't seem like the right solution to me. `HIDDEN` shouldn't change mechanics but just visibility of the database.

dentiny: Sorry it's my bad -- I misunderstood.

The biggest issue to reject database detach properly is `read_duckdb` also attach/detach legally.
In the latest commit I move the validation logic to `PhysicalDetach`.

dentiny:
Hi @Mytherin , I'm wondering if you're willing to take another look when you have some time? Thank you!

Mytherin:
Thanks!

PR #22259, opened 2026-04-24
ISSUE-22235: Fix timetz casting offset

herbenderbler:
## Description

Fixes https://github.com/duckdb/duckdb/issues/22235.

`TIME -> TIMETZ` ignored the session `TimeZone` and always produced a `+00`
offset, even though `VARCHAR -> TIMETZ` and `TIMESTAMPTZ -> TIMETZ` correctly
honor it. The result was that `'00:00:00'::TIMETZ` and
`'00:00:00'::TIME::TIMETZ` disagreed under any non-UTC session TZ, e.g. with
`SET TimeZone='Asia/Kolkata'`:

```text
SELECT '00:00:00'::TIMETZ;            -- 00:00:00+05:30
SELECT '00:00:00'::TIME::TIMETZ;      -- 00:00:00+00   (bug)
```

In addition to the cast itself, the optimizer's statistics propagation was
treating the same cast as offset-preserving, so propagated min/max for
`TIME -> TIMETZ` (and `TIMESTAMPTZ -> TIMETZ`) would diverge from the runtime
value once this fix is in place. Both are corrected here.

## Implementation Details

Primary fix (issue #22235):

- `extension/icu/include/icu-casts.hpp`: declare `ICUToTimeTZ::CastFromTime`
  and `ICUToTimeTZ::BindCastFromTime`, mirroring the existing `CastToTimeTZ`
  / `BindCastToTimeTZ` pair.
- `extension/icu/icu-timezone.cpp`: implement the `TIME -> TIME_TZ` ICU cast.
  It reads the session UTC offset (zone + DST) from the bound calendar and
  produces `dtime_tz_t(input, offset)`. The logic intentionally matches the
  no-offset branch of `ICUStrptime::VarcharToTimeTZ` so the VARCHAR and TIME
  paths agree. Registered in `ICUToTimeTZ::AddCasts` alongside the existing
  `TIMESTAMP_TZ -> TIME_TZ` override, with the implicit-cost computed via
  `CastRules::ImplicitCast(TIME, TIME_TZ)`.
- The default static operator in `src/common/operator/cast_operators.cpp` is
  intentionally left unchanged: without ICU there is no TZ database to
  consult, and `+00` remains the correct fallback for the no-ICU build.
- `test/sql/function/timetz/test_icu_cast.test`: new regression cases under
  `America/Phoenix`, `Asia/Kolkata`, and `UTC`, including a direct equality
[...]

PR #22439, opened 2026-05-02
Use ArgProperties for cast stats forward-bounds

J-Meyers:
Cast-side mirror of #22384. `BoundCastInfo` gains an `arg_properties` field, the `*CastSwitch` factories opt monotonic casts in via `SetArgProperties(...)`, and `StatisticsPropagator::PropagateExpression(BoundCastExpression&)` derives output min/max by mapping the child's min/max through the bound cast.

Broken off from #22286 for review; the peel optimizer that also consumes this metadata is a separate follow-up.

## Why this PR is worth landing on its own

The set of *new* type-pair propagations visible from this PR alone is modest — the timestamp-precision FIXMEs in `CanPropagateCast` (TIMESTAMP[_TZ] → TIMESTAMP_NS/MS/S and inverses); unsigned-to-wider numeric widening (UTINYINT → USMALLINT/UINTEGER/UBIGINT/HUGEINT/FLOAT/DOUBLE and analogues — main's physical-type switch only accepted signed integer + float targets, so no unsigned source propagated through it); `TIME → TIMETZ` under the no-extension build; plus the correctness fix for ICU TIMESTAMP_TZ propagation via the active registry. The MIN/MAX peel PR is what consumes this metadata for the user-visible 25–63× speedups. Reasons to land plumbing first:

- **Extension casts can now participate in (or override) stats propagation.** The static `CanPropagateCast` allowlist in core hardcodes which type pairs propagate; extensions that register custom casts had no hook. With per-`BoundCastInfo` metadata, an extension's `BindCast` returns `BoundCastInfo(...).SetArgProperties(...)` and immediately participates — same opt-in mechanism recently introduced for extension functions.
- **Removes the allowlist from the runtime expression path.** `BoundCastExpression::PropagateExpression` no longer consults `CanPropagateCast`. The static API is retained only for storage-layer callers (struct/variant stats, multi-file column mapper) that operate on bare `LogicalType` pairs without a `BoundCastInfo` or `ClientContext`.
- **Active-registry correctness for ICU.** The previous `Value::DefaultTryCastAs` propagation constructed a fresh `CastFunctionSet` that didn't see ICU's TIMESTAMP_TZ overrides. The new path uses `Value::TryCastAs(context, ...)` so propagated bounds match runtime values when extensions are loaded.
- **Plumbing for follow-ups.** The MIN, MAX peel optimizer and more will be able to share code between functions and casts.

## What was unhandled before

`BoundCastExpression` stats went through a static type-pair allowlist (`CanPropagateCast` → `StatisticsOperationsNumericNumericCast` via `Value::DefaultTryCastAs`). That path:

- Couldn't see extension overrides — `DefaultTryCastAs` constructs a fresh `CastFunctionSet` containing only the default casts, so for ICU's `TIMESTAMP_TZ` overrides the propagated bounds didn't match runtime values.
- Bailed on the timestamp-precision conversions (`TIMESTAMP[_TZ] → TIMESTAMP_NS/MS/SEC` and the inverse) — explicitly marked `// FIXME: perform actual stats propagation for these casts`.
- Bailed on `TIME → TIMETZ` and `TIMESTAMP_TZ → TIMETZ` per #22259, because the static `DefaultTryCastAs` produces UTC-only bounds while runtime (with ICU) is session-zone-aware. The new path uses the active registry, so the bound matches runtime.
- Allowed only signed integer + float/double physical types in the source/target switch, so any cast involving an unsigned source or unsigned target bailed at the type-filter step. The new path's `IsOrderPreservingNumericCast<SRC, DST>()` annotates: same-signedness pairs; any integer→float pair (unsigned→float is just non-negative-into-mantissa); and unsigned→strictly-wider-signed (UTINYINT→SMALLINT/INTEGER/BIGINT/HUGEINT, USMALLINT→INTEGER/BIGINT/HUGEINT, etc., where every source value lands in the target's non-negative half). Same-width cross-signedness (e.g. UTINYINT→TINYINT) and signed→unsigned are deliberately left UNKNOWN — they overflow on the boundary and need range-aware refinement to be safe.

## Annotations

Numeric same-signedness widening (strict; non-decreasing for float-precision-loss cases), `INT → DECIMAL` (strict), `DECIMAL → DECIMAL` widening, `DECIMAL → FLOAT/DOUBLE` (non-decreasing), `DECIMAL → INT` truncation (non-decreasing), `DATE` → temporal widenings (strict), temporal → `VARCHAR` (strict), `TIMESTAMP_*` precision conversions (strict for widening, non-decreasing for truncation), `TIMESTAMP → DATE` (non-decreasing), and `TIME → TIMETZ` (strict; default cast packs offset 0 — ICU's session-zone override returns its own `BoundCastInfo` without `SetArgProperties` so propagation correctly bails when ICU is loaded).

### Deliberately left UNKNOWN

- **Signed→unsigned numeric casts** — not order-preserving across zero.
- **`TIMESTAMP → TIME`, `TIMESTAMP → TIMETZ`, `TIMESTAMP_TZ → TIMETZ`** — drop the date; distinct timestamps on different days collapse to the same time.
- **ICU `TIMESTAMP_TZ` casts** — DST transitions break monotonicity (the fall-back hour renders distinct UTC instants to the same naive timestamp). Since the `BoundCastInfo` defaults to `UNKNOWN` and ICU's bind path doesn't call `SetArgProperties`, no change needed in the ICU extension.

`try_cast = true` casts retain their existing `CAN_HAVE_NULL_VALUES` handling — the propagator marks the result nullable to reflect that overflow becomes NULL.

### DIFF
--- a/src/execution/expression_executor/CMakeLists.txt
+++ b/src/execution/expression_executor/CMakeLists.txt
@@ -10,7 +10,8 @@ add_library_unity(
   execute_function.cpp
   execute_operator.cpp
   execute_parameter.cpp
-  execute_reference.cpp)
+  execute_reference.cpp
+  verify_arg_properties.cpp)
 set(ALL_OBJECT_FILES
     ${ALL_OBJECT_FILES} $<TARGET_OBJECTS:duckdb_expression_executor>
     PARENT_SCOPE)
--- a/src/execution/expression_executor/execute_cast.cpp
+++ b/src/execution/expression_executor/execute_cast.cpp
@@ -1,6 +1,7 @@
 #include "duckdb/common/vector/flat_vector.hpp"
 #include "duckdb/common/vector_operations/vector_operations.hpp"
 #include "duckdb/execution/expression_executor.hpp"
+#include "duckdb/execution/verify_arg_properties.hpp"
 #include "duckdb/function/scalar_function.hpp"
 #include "duckdb/planner/expression/bound_cast_expression.hpp"
 
@@ -67,6 +68,7 @@ void ExpressionExecutor::Execute(const BoundCastExpression &expr, ExpressionStat
 		result.SetVectorType(VectorType::CONSTANT_VECTOR);
 	}
 	FlatVector::SetSize(result, count_t(count));
+	VerifyCastArgProperties(expr, child, result, count);
 }
 
 } // namespace duckdb
--- a/src/execution/expression_executor/execute_function.cpp
+++ b/src/execution/expression_executor/execute_function.cpp
@@ -1,11 +1,9 @@
 #include "duckdb/common/type_visitor.hpp"
 #include "duckdb/common/vector/flat_vector.hpp"
 #include "duckdb/execution/expression_executor.hpp"
+#include "duckdb/execution/verify_arg_properties.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
 #include "duckdb/common/types/uuid.hpp"
-#include "duckdb/common/value_operations/value_operations.hpp"
-
-#include <numeric>
 
 namespace duckdb {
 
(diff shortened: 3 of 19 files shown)


lnkuiper (commented):
Thanks for the PR! Really useful stuff.

I think there is a very small correctness issue here:
```sql
PRAGMA explain_output='optimized_only';

CREATE TABLE ts AS
SELECT (TIMESTAMP '2024-01-01' + INTERVAL (i) MICROSECOND) AS t
FROM range(100) _(i);

SELECT s.min, s.max, s.approx_unique
FROM (SELECT STATS(t::TIMESTAMP_MS) s FROM ts LIMIT 1);
-- currently reports approx_unique = 100

SELECT count(DISTINCT t::TIMESTAMP_MS), approx_count_distinct(t::TIMESTAMP_MS)
FROM ts;
-- actual cast result has 1 distinct value
```
Not sure how much we care about that / how easy it is to fix. Maybe it can be fixed by adding an `enum class Injectivity` field to `ArgProperties` that essentially says whether unique count is preserved.

The rest looks good, but I am slightly worried about maintainability, having to annotate all the functions, and also about potential incorrect annotations in the future. Maybe we can add a `VerifyArgProperties` function that is guarded by `#ifdef DEBUG` that verifies that the annotated arg properties are actually valid during query execution? Similar to our other internal verification.

J-Meyers:
I've fixed the approx_unique case for stats and added handling for functions to propagate their distinct counts as well.

I'm was trying to think about VerifyArgProperties, I think there are a few different options for what could be checked:
- Do the min and max actually contain all the actual outputs
- Much more expensive checks:
  - Actually check monotonicity + injectivity by doing a sort of inputs and outputs and comparing
  - Checking approx_unique seems always difficult because even if it propagates correctly if the baseline approx was off then we can't actually check that the propagated values make sense, but I guess maybe we could effectively compute the actual values before and after, then check if it was maintained


One other thing with this DistinctCount is it seems there are other opportunities possible to track strict upper bounds of distinctness (max - min + 2) - when there are integer columns at any point in the propagation, and row count both serve as upper bounds and can be further propagated (I list both because sometimes one may be much bigger than the other and by using both we should be able to get a lower bound, although in some cases y'all are already using row count), they are a bit fundamentally different than estimates from sketches, but could give more information to join order optimizations.

lnkuiper:
Thanks for the changes! Checking monotonicity should only require comparing the input and outputs, which is not cheap, but if guarded by `#ifdef DEBUG` that's fine. We already do something very similar with `VerifyNullHandling`. I think this is quite important because incorrect property annotations can cause the optimizer to create a query plan that produces incorrect results due to incorrect statistics. I'm much less concerned about injectivity.

Would you be open to adding that verification to this PR? The distinct estimation is a cool idea, but definitely for a separate PR.

J-Meyers:
I wasn't quite sure what you wanted, but these should provide checks for both monotonicity and injectivity across every row possible, though it does require sorts.

The alternative that would be much more efficient is basically just computing the limits of each column, then doing the corner analysis and checking that values are within those bounds, that isn't as extensive and thorough as the true row by row checks (and can't check injectivity to any degree), but should be faster if that is a significant concern.

lnkuiper:
Thanks! Looks good, and the verification seems to trigger an issue (or the verification is incorrect, could be that, too, but this seems to have been fruitful already!).

I also worry that using the `Value`-based API for this will be too slow. It's OK for debug builds to be slow, but this might be too slow. Perhaps this would be better with some templating. Since there are two vectors this would require templating on two types, which kind of blows up, but maybe that's not required? What happens if we cast back and then do the verification templated on one type? Then we can use more efficient `VectorOperations`. Or would that nullify the usefulness of the verification?

PR #22445, opened 2026-05-02
Fix Arrow extension propagation in nested appenders (UNION/STRUCT/LIST/MAP/ARRAY)

### LINKED ISSUE #22444: BOOLEAN child of a UNION written bit-packed but exported as arrow.bool8 (byte-packed) under arrow_lossless_conversion
### What happens?

When `arrow_lossless_conversion = true`, a `BOOLEAN` child of a `UNION` is corrupted on the way out to Arrow. The exported schema declares the bool child as the `arrow.bool8` extension type (byte-packed, 1 byte per value), but the appender writes the underlying buffer **bit-packed** (1 bit per value). Any consumer reading the buffer according to the schema interprets each byte as one row, so an input of N `TRUE` rows comes out as `[True, False, False, …]` (just the first row is correct, by accident).

This is observable directly off `con.execute(...).arrow()` — no IPC, no extension, no other tooling required.

### Cause (from a quick read of the source)

Two code paths disagree on the layout for a `BOOLEAN` that is a child of a `UNION`:

* **Schema export** — `src/common/arrow/arrow_converter.cpp` `SetArrowFormat`, the `LogicalTypeId::BOOLEAN` case:
  ```cpp
  case LogicalTypeId::BOOLEAN:
      if (options.arrow_lossless_conversion) {
          SetArrowExtension(root_holder, child, type, context);   // declares arrow.bool8 (format "c", byte-packed)
      } else {
          child.format = "b";                                     // declares plain bool (bit-packed)
      }
      break;
  ```
* **Data appender for union children** — `src/common/arrow/appender/union_data.cpp` `ArrowUnionData::Initialize`:
  ```cpp
  for (auto &child : UnionType::CopyMemberTypes(type)) {
      auto child_buffer = ArrowAppender::InitializeChild(child.second, capacity, result.options);
      // ^^ no `extension_type_cast` arg, so BOOLEAN dispatches to ArrowBoolData (bit-packed)
      result.child_data.push_back(std::move(child_buffer));
  }
  ```

The top-level `ArrowAppender` constructor (`src/common/arrow/arrow_appender.cpp`, lines 21–30) plumbs `extension_type_cast` per-column into `InitializeChild`, so a top-level BOOLEAN routes through `ArrowBool8::DuckToArrow` and the schema/data agree (both byte-packed). The union appender does not propagate that, so the union's BOOLEAN child uses bit-packed `ArrowBoolData` while the schema still claims `arrow.bool8`.

The single-row case happens to look right by accident: the bit-packed first byte has bit 0 set, and the byte-wise reader interprets that same byte as `0x01` = `True`. Anything beyond row 0 reads zero bytes and comes out `False`.

### To Reproduce

```python
"""Minimal reproduction: BOOLEAN child of a UNION corrupts in Arrow output
when arrow_lossless_conversion = true.

Run with:
    uv run --no-project --isolated --with duckdb --with pyarrow \
        python duckdb_union_bool_bug.py
"""
import duckdb
import pyarrow as pa

print(f"duckdb {duckdb.__version__}, pyarrow {pa.__version__}")

con = duckdb.connect()
con.execute("SET arrow_lossless_conversion = true")

# 4 rows, all with the BOOLEAN variant of a UNION(int, bool) set to TRUE.
batch = con.execute("""
    SELECT * FROM (VALUES
        (union_value(b := true)::UNION(i INT, b BOOLEAN)),
        (union_value(b := true)),
        (union_value(b := true)),
        (union_value(b := true))
    ) t(u)
""").arrow().read_next_batch()

bool_child = batch.column(0).field(1)

print(f"union type:        {batch.column(0).type}")
print(f"bool child type:   {bool_child.type}")
print(f"bool child buffer: {bytes(bool_child.buffers()[1])!r}  ({bool_child.buffers()[1].size} bytes)")
print(f"bool child values: {bool_child.to_pylist()}")
print(f"expected values:   {[True, True, True, True]}")

assert bool_child.to_pylist() == [True, True, True, True]
```

Output:

```
duckdb 1.5.2, pyarrow 24.0.0
union type:        sparse_union<i: int32=0, b: extension<arrow.bool8>=1>
bool child type:   extension<arrow.bool8>
bool child buffer: b'\xff\x00\x00\x00'  (4 bytes)
bool child values: [True, False, False, False]
expected values:   [True, True, True, True]
  File "duckdb_union_bool_bug.py", line 32, in <module>
    assert bool_child.to_pylist() == [True, True, True, True]
AssertionError
```

The buffer is `b'\xff\x00\x00\x00'`: the appender wrote the four `True` values bit-packed into the low 4 bits of byte 0 (`0xff`), but the schema says each byte is one row, so PyArrow reads `[True, False, False, False]`.

Reading the same buffer bit-packed (which is what the appender intended) gives `[True, True, True, True]`.

### Notes

* The bug needs `arrow_lossless_conversion = true` (the default for several integrations); without it, `SetArrowFormat` writes `format = "b"` (bit-packed bool) and schema/data agree.
* It surfaces for any number of rows ≥ 2; single-row coincidentally appears correct only when the input row is `True`.
* It also surfaces inside any container whose Arrow output goes through `SetArrowFormat`/`ArrowUnionData` — e.g., `UNION inside STRUCT`, `UNION inside MAP value`, etc.
* Two reasonable fixes:
  * (preferred) make `ArrowUnionData::Initialize`/`Append` look up the `arrow.bool8` extension for `BOOLEAN` children and route through it, mirroring the top-level appender's logic in `arrow_appender.cpp`;
  * (narrower) in `arrow_converter.cpp` `SetArrowFormat`, force `child.format = "b"` for `BOOLEAN` when emitted as a union member, accepting the loss of round-trip distinction between DuckDB BOOLEAN and Arrow plain bool inside unions.

rustyconover:
Fixes #22444. Targeting `v1.5-variegata` since this is a correctness regression.

### Summary

The original issue described a `BOOLEAN` child of a `UNION` corrupting under `arrow_lossless_conversion = true`. While preparing the fix I noticed the same root cause affects every container appender, not just `UNION`. This PR fixes the class of bug.

### Root cause

When `arrow_lossless_conversion = true`, `SetArrowFormat` declares a column's schema through an Arrow type extension when one applies (e.g. `arrow.bool8` for `BOOLEAN`). The top-level `ArrowAppender` constructor handled this for root columns by routing values through the extension's `duckdb_to_arrow` callback. But every nested container appender (struct/list/list_view/fixed-size list/map/union) called `ArrowAppender::InitializeChild` without any extension info and used the plain logical-type appender — so for `BOOLEAN` children the schema said `arrow.bool8` (1 byte / value, byte-packed) while the buffer was bit-packed (1 bit / value).

The most user-visible case is `BOOLEAN`. A consumer that reads the buffer per the schema sees row 0 correctly only by accident (the bit-packed byte `0xff` is also `True` byte-wise) and reads `False` for every subsequent row. For `BOOLEAN` map keys this even crashes ingest with a duplicate-key error because all keys collapse to the same byte value.

Reproduces in stock pip-installed `duckdb 1.5.2` + `pyarrow 24.0.0`:

```python
import duckdb
con = duckdb.connect()
con.execute("SET arrow_lossless_conversion = true")
batch = next(iter(con.execute("""
    SELECT [(i % 2 = 0), (i % 3 = 0)]::BOOLEAN[] AS l FROM range(4) tbl(i)
""").arrow()))
print(batch.column(0).flatten().to_pylist())
# Buggy: [True, False, False, False, False, False, False, False]
```

### Fix

Centralise extension handling so it propagates to every nesting level.

* `ArrowAppender::InitializeChild` auto-resolves the Arrow extension for the given type when no explicit override is passed (via the same `DBConfig` lookup the top-level appender already performed). The `bitshift_boolean` gate (top-level `BOOLEAN` without lossless stays bit-packed) moves into `InitializeChild` so it applies uniformly.
* New `ArrowAppendData::AppendChild(input, from, to, input_size)` member: when `extension_data` is set, flattens the source, runs `duckdb_to_arrow` into a correctly-sized internal vector, then forwards to the child's `append_vector`. The flatten covers sliced/dictionary-shaped child vectors that several extension callbacks (e.g. `ArrowBool8::DuckToArrow`) don't handle. The internal vector is sized to `MaxValue(input_size, STANDARD_VECTOR_SIZE)` so list children whose total element count exceeds the default vector size don't write OOB.
* Each container appender (struct/list/list_view/fixed-size list/map/union) now calls `child_data.AppendChild(...)` instead of `child_data.append_vector(...)` directly.
* Top-level `ArrowAppender::ArrowAppender` and `ArrowAppender::Append` are simplified to share the same paths.

### Tests

* New: `Test Arrow nested BOOLEAN with arrow.bool8 roundtrip` — `BOOLEAN` inside `STRUCT`, `LIST`, `FIXED_SIZE_LIST`, `MAP`-value, `MAP`-key, `LIST`-of-`STRUCT`, `STRUCT`-of-`LIST`, and large `LIST(BOOLEAN)` whose child element count exceeds `STANDARD_VECTOR_SIZE` (regression for the OOB write the centralised fix exposed).
* Kept: `Test Arrow UNION with BOOLEAN member roundtrip` — original repro from #22444 covering single- and multi-row, mixed variants, three-member unions, NULL rows, large batches crossing `STANDARD_VECTOR_SIZE`, and bool-in-union nested inside a struct.

### Verification

macOS arm64. With this commit on top of `v1.5-variegata`:

* `[arrow]` bucket: 32 cases / 31,625 assertions, all pass.
* `[capi]` bucket: 154 cases / 704,679 assertions, all pass.
* Both new test cases fail before this change and pass after.

### DIFF
--- a/src/common/arrow/appender/append_data.cpp
+++ b/src/common/arrow/appender/append_data.cpp
@@ -26,4 +26,33 @@ void ArrowAppendData::AppendValidity(UnifiedVectorFormat &format, idx_t from, id
 	}
 }
 
+void ArrowAppendData::AppendChild(Vector &input, idx_t from, idx_t to, idx_t input_size) {
+	if (extension_data && extension_data->duckdb_to_arrow) {
+		// Mirror the top-level ArrowAppender::Append: convert the DuckDB-typed
+		// vector into the extension's internal Arrow-typed vector before
+		// handing it to the (internal-typed) child appender. The conversion
+		// runs over the full `input_size` to match `append_vector`'s
+		// expectations of a fully-resolved input vector.
+		//
+		// Flatten first because some duckdb_to_arrow callbacks (e.g.
+		// ArrowBool8::DuckToArrow) read `format.data[i]` without honoring
+		// `format.sel`, so feeding them a sliced/dictionary vector reads out
+		// of bounds. Container appenders routinely produce sliced child
+		// vectors, so the conversion path can't rely on the source being
+		// flat the way the top-level append path does.
+		input.Flatten(input_size);
+		// Allocate the internal vector with at least input_size capacity:
+		// container appenders can produce child vectors larger than
+		// STANDARD_VECTOR_SIZE (e.g. a 2048-row LIST whose elements average
+		// 2 entries → 4096-element child), and several duckdb_to_arrow
+		// callbacks write input_size values into the result without
+		// bounds-checking.
+		Vector internal(extension_data->GetInternalType(), MaxValue<idx_t>(input_size, STANDARD_VECTOR_SIZE));
+		extension_data->duckdb_to_arrow(*options.client_context, input, internal, input_size);
+		append_vector(*this, internal, from, to, input_size);
+	} else {
+		append_vector(*this, input, from, to, input_size);
+	}
+}
+
 } // namespace duckdb
--- a/src/common/arrow/appender/fixed_size_list_data.cpp
+++ b/src/common/arrow/appender/fixed_size_list_data.cpp
@@ -23,7 +23,10 @@ void ArrowFixedSizeListData::Append(ArrowAppendData &append_data, Vector &input,
 	auto array_size = ArrayType::GetSize(input.GetType());
 	auto &child_vector = ArrayVector::GetEntry(input);
 	auto &child_data = *append_data.child_data[0];
-	child_data.append_vector(child_data, child_vector, from * array_size, to * array_size, size * array_size);
+	// AppendChild routes through the child's Arrow extension (if any) so
+	// e.g. arrow.bool8 BOOLEAN array-elements get the duckdb_to_arrow
+	// conversion before being fed into the byte-packed appender.
+	child_data.AppendChild(child_vector, from * array_size, to * array_size, size * array_size);
 	append_data.row_count += size;
 }
 
--- a/src/common/arrow/appender/struct_data.cpp
+++ b/src/common/arrow/appender/struct_data.cpp
@@ -24,7 +24,10 @@ void ArrowStructData::Append(ArrowAppendData &append_data, Vector &input, idx_t 
 	for (idx_t child_idx = 0; child_idx < children.size(); child_idx++) {
 		auto &child = children[child_idx];
 		auto &child_data = *append_data.child_data[child_idx];
-		child_data.append_vector(child_data, *child, from, to, size);
+		// AppendChild routes through the child's Arrow extension (if any) so
+		// e.g. arrow.bool8 BOOLEAN children get the duckdb_to_arrow conversion
+		// before being fed into the byte-packed appender.
+		child_data.AppendChild(*child, from, to, input_size);
 	}
 	append_data.row_count += size;
 }
--- a/src/common/arrow/appender/union_data.cpp
+++ b/src/common/arrow/appender/union_data.cpp
@@ -49,7 +49,7 @@ void ArrowUnionData::Append(ArrowAppendData &append_data, Vector &input, idx_t f
 	for (idx_t child_idx = 0; child_idx < child_vectors.size(); child_idx++) {
 		auto &child_buffer = append_data.child_data[child_idx];
 		auto &child = child_vectors[child_idx];
-		child_buffer->append_vector(*child_buffer, child, 0, size, size);
+		child_buffer->AppendChild(child, 0, size, size);
 	}
 	append_data.row_count += size;
 }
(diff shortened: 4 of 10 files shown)


rustyconover:
@evertlammerts @pdet this is an Arrow bug we could make progress on if we had the ability to write tests against the reference implementation.

evertlammerts:
1. I think that it's cleaner to fix the arrow extension to use the selection vector, instead of flattening before passing in the conversion. That leaves it to the extension to decide what to do. I have a PR ready for this, putting it up in a bit.
2. Testing is a bit of a nuisance, though in this case the unit test might be enough. I'm testing with duckdb-python and will put up a link to the results.

evertlammerts:
Closing in favor of #23190

Note on testing against the reference implementation: it's a nuisance at the moment but not impossible. duckdb-python includes duckdb as a git submodule. If you have a fork of duckdb-python and duckdb, then you can point the submodule in duckdb-python to your pushed fork + branch of duckdb and run the tests locally or in CI, also with new tests in duckdb-python. We could introduce complexity in CI for more automated support, but I think it would be better to test in core directly, e.g. against a non-python reference implementation (maybe arro3?).

PR #21279, opened 2026-03-10
Extend Join Pushdown with Prefix Range Filter

d-justen:
This PR introduces the **Prefix Range Filter** as an alternative for bloom filter pushdown in joins. As its name implies, the prefix range filter is range-based instead of hash-based - essentially it is an equi-width bitmap histogram based on the join key bit prefixes. It allows probabilistic filtering just like bloom filters (i.e., only false-positives, no false negatives) and is even quite a bit faster to ingest/probe.

## Why?

Range- and value-based summaries allow for row group pruning with min/max statistics in join filter pushdown, whereas hash-based summaries can't. Especially with remote Parquet files, this can make a huge performance difference. I have included an illustrative micro-benchmark on a local duckdb database file, which shows a 10x improvement on my MacBook. This is to be taken with a grain of salt of course, because the speedup simply comes from "more pruning = less computing" and less computing is of course always faster.

DuckDB already has this kind of pruning with the Perfect Hash Join (= exact version of the prefix range filter) pushdown, but that is limited by the join key value range, build size, and does not trigger for many-to-many joins. Finally, in real-world data, join columns are far more often clustered and/or correlated than in academic benchmarks. This makes range-based summaries more likely to be effective. This also shows in the benchmarks, which show some improvements in JOB but not so much in TPC-H.

## Limitations

At the moment, the prefix range filter is only built conservatively in one of these two cases:
- The join key range is smaller than the max filter size (`2^20` bits). In this case, the prefix range filter is exact.
- The build side is smaller than half the max filter size. In that case, at least half of the bits are unset, even if there is no join key clustering on the build sides.

## TODOs

For the future, there is also definitely still quite a lot of work, e.g.,
- Some form of build side cluster detection and, in general, a more informed decision making when to build min/max, PRF, bloom, ...
- Joins with multiple keys
- Data type support (hugeint, string, ...)
- Filter Sizing (currently fixed to `2^20 bits` if key range is larger than `2^20`)

PR #22447, opened 2026-05-02
Fix crash on truncated UTF-8 string statistics

J-Meyers:
## Root cause

`StringStatsData` stores min/max as 8-byte arrays (`MAX_STRING_MINMAX_SIZE = 8`). Strings longer
than 8 bytes are truncated. For multibyte UTF-8 sequences whose codepoint boundary falls inside that
window, the stored bytes are not valid UTF-8.

---

## Bug 1 — `PartitionedExecutionAddStatsNodes` (`src/optimizer/partitioned_execution.cpp`, PR #22336)

```cpp
case StatisticsType::STRING_STATS:
    min = StringStats::Min(stats);   // std::string with raw truncated bytes
    max = StringStats::Max(stats);
    break;
```

Assigning a raw `std::string` to `Value` via the implicit `Value::Value(string)` constructor
validates UTF-8 and throws `"Invalid unicode (byte sequence mismatch)"` when the 8-byte window
splits a multibyte codepoint. Fires on ORDER BY / GROUP BY / WINDOW over a VARCHAR column whose
per-row-group stats contain a truncated multibyte sequence.

**Fix**: guard in `CanUseStats` before the `Value(string)` call. After confirming the stats type is
VARCHAR and `HasMinMax`, extract the raw min/max strings and return `false` if either fails
`Value::StringIsValid`. Row groups with invalid stats are dropped from the optimization — conservative
but never incorrect. `AddStatsNodes` is only reached when the stats are valid, so `Value(string)` is
safe there.

Note: `PartitionedExecutionMinMaxEqual` in the same file already uses `Value::BLOB_RAW` (as does the
PRF fix), but `AddStatsNodes` is different — its values flow eventually into `BoundColumnRefExpression` nodes typed by `range.min.type()`. Using `BLOB_RAW` there would produce BLOB-typed column references for VARCHAR columns.

---

## Bug 2 — `PrefixRangeTableFilter::CheckStatistics` (`src/planner/filter/prefix_range_filter.cpp`, PR #21279)

```cpp
min = Value(StringStats::Min(stats));
max = Value(StringStats::Max(stats));
```

Same UTF-8 validation throw. Fires during join pushdown when the probe-side table has per-row-group
stats with a truncated multibyte sequence.

**Fix**: use `Value::BLOB_RAW` instead of `Value(string)`. The values are consumed by `LookupRange`
via raw `string_t` bytes and never become typed plan expressions, so BLOB_RAW is safe here — no
UTF-8 validation needed and no type mismatch risk. A `HasMinMax` guard is also added before the
construction, consistent with the numeric branch above it.

### DIFF
--- a/test/sql/join/inner/test_prefix_range_filter_non_utf8.test
+++ b/test/sql/join/inner/test_prefix_range_filter_non_utf8.test
@@ -1,0 +1,32 @@
+# name: test/sql/join/inner/test_prefix_range_filter_non_utf8.test
+# description: PrefixRangeFilter does not crash on parquet stats with non-UTF-8 bytes
+# group: [inner]
+
+require parquet
+
+require strinline
+
+statement ok
+SET disabled_optimizers = 'join_order,build_side_probe_side';
+
+statement ok
+COPY (
+    SELECT CASE WHEN i < 5 THEN 'ascii' || lpad(i::VARCHAR, 3, '0')
+                ELSE            chr(26085) || chr(26412) || chr(35486) || lpad(i::VARCHAR, 4, '0')
+           END AS k,
+           (i % 3)::INTEGER AS g
+    FROM range(10) t(i)
+) TO '{TEST_DIR}/prf_non_utf8_probe.parquet' (ROW_GROUP_SIZE 5);
+
+statement ok
+CREATE OR REPLACE TABLE prf_build_non_utf8 (k VARCHAR, g INTEGER, payload INTEGER);
+
+statement ok
+INSERT INTO prf_build_non_utf8 VALUES ('a', 0, 1), ('z', 0, 2), ('m', 1, 3);
+
+query I
+SELECT count(*)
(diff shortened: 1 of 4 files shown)


test/sql/join/inner/test_prefix_range_filter_non_utf8.test
```diff
@@ -0,0 +1,32 @@
+# name: test/sql/join/inner/test_prefix_range_filter_non_utf8.test
+# description: PrefixRangeFilter does not crash on parquet stats with non-UTF-8 bytes
+# group: [inner]
+
+require parquet
+
+require strinline
+
+statement ok
+SET disabled_optimizers = 'join_order,build_side_probe_side';
```
Mytherin: Are these extra qualifiers in the test necessary?

lnkuiper:
Thanks, fix looks good to me :)

PR #22449, opened 2026-05-03
date_trunc avoid decomposition

J-Meyers:
For units whose length is a constant number of microseconds (`hour`, `minute`, `second`, `millisecond`), replace the calendar-decomposition path with a 2-op floor division. Variable-length units (`day`, `week`, `month`, `quarter`, `year`, `decade`, `century`, `millennium`) keep the existing path because their boundaries depend on the calendar.

Existing tests already cover negative timestamps so there were no interesting tests to add.

## Benchmark — 2.48×

```sql
CREATE TABLE t AS (
      -- Make a big table of not dictionary encoded times over a wide range relatively evenly spread out
      SELECT epoch_ms(((range * 2654435761) % 315360000000 + 1000000000000)::BIGINT) AS ts
      FROM range(50000000)
);
```
```sql
SELECT sum(extract(epoch from date_trunc('minute', ts))) FROM t;
```

50 M shuffled timestamps, single-threaded. Alternating runs to rule out branch-predictor / ordering effects.

| main | new | speedup |
|--------:|:-:|:-:|
| **531ms** | 214ms | **2.48x** |

### DIFF
--- a/benchmark/micro/date/date_trunc.benchmark
+++ b/benchmark/micro/date/date_trunc.benchmark
@@ -1,0 +1,17 @@
+# name: benchmark/micro/date/date_trunc.benchmark
+# description: date_trunc on a fixed-width unit
+# group: [date]
+
+name Date Trunc
+group date
+
+load
+CREATE TABLE timestamps AS
+SELECT epoch_ms(((range * 2654435761) % 315360000000 + 1000000000000)::BIGINT) AS ts
+FROM range(0, 50000000);
+
+run
+SELECT sum(epoch_ms(date_trunc('minute', ts)))::HUGEINT FROM timestamps
+
+result I
+57883998804880080000
--- a/extension/core_functions/scalar/date/date_trunc.cpp
+++ b/extension/core_functions/scalar/date/date_trunc.cpp
@@ -96,52 +96,42 @@ struct DateTrunc {
 		}
 	};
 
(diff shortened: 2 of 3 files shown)


hawkfish (changes requested):
Thanks for the PR! This looks great - just can you also add `day` (we don't model leap seconds, and DST is handled in the ICU extension.)

Also, can you add a benchmark test to show the improvement? (It doesn't need to be single-threaded because we control that in the regression framework.)

J-Meyers:
Multithreaded timings (note there is some noise)
| Unit          |   main |    PR | Wall speedup | Net trunc cost (main → PR) | Net trunc speedup |
| :------------ | -----: | ----: | :----------: | :------------------------: | :---------------: |
| `NONE` (sum)  |  10 ms | 10 ms |      —       |    — (baseline floor)      |         —         |
| `hour`        |  37 ms | 16 ms |    2.31×     |        27 → 6 ms           |       4.5×        |
| `minute`      |  36 ms | 18 ms |    2.00×     |        26 → 8 ms           |       3.25×       |
| `second`      |  36 ms | 17 ms |    2.12×     |        26 → 7 ms           |       3.7×        |
| `millisecond` |  39 ms | 16 ms |    2.44×     |        29 → 6 ms           |       4.83×       |
| `day`         |  25 ms | 17 ms |    1.47×     |        15 → 7 ms           |       2.14×       |

Mytherin:
Thanks!

PR #22450, opened 2026-05-03
Fix TRY_CAST to BIGNUM for invalid strings

### LINKED ISSUE #22324: TRY_CAST fails for BIGNUM
### What happens?

Based on documentation:
> TRY_CAST will never throw an error, and will instead return NULL if a cast is not possible.

```sql
SELECT TRY_CAST('hello' AS BIGNUM) AS i;
```

> Conversion Error:
> Could not convert string 'hello' to Bignum

```sql
SELECT TRY_CAST('1e+24' AS BIGNUM) AS i;
```

> Conversion Error:
> Could not convert string '1e+24' to Bignum

### To Reproduce

SELECT TRY_CAST('hello' AS BIGNUM) AS i;
SELECT TRY_CAST('1e+24' AS BIGNUM) AS i;

### OS:

arm64

### DuckDB Version:

v1.5.2 (Variegata) 8a5851971f

### DuckDB Client:

cli, python

### Hardware:

_No response_

### Full Name:

Mārtiņš Kalvāns

### Affiliation:

NYM Technology

- [x] Yes, I have

- [x] Yes, I have

Yes

Schwarf:
This PR fixes `TRY_CAST(... AS BIGNUM)` for invalid string inputs.

Previously, inputs such as `'hello'` and `'1e+24'` raised a conversion error even when using `TRY_CAST`. With this change, invalid string inputs now return `NULL`.
A regression test was added for the fixed behavior.
Fixes #22324.
## Testing
- Added `test/sql/types/bignum/test_bignum_try_cast.test`
- Verified that the same branch passed all CI checks in [my fork PR](https://github.com/Schwarf/duckdb/pull/1)

### DIFF
--- a/src/function/cast/bignum_casts.cpp
+++ b/src/function/cast/bignum_casts.cpp
@@ -167,6 +167,12 @@ bignum_t HugeintCastToBignum::Operation(hugeint_t int_value, StringHeap &heap) {
 template <>
 bool TryCastToBignum::Operation(string_t input_value, bignum_t &result_value, Vector &result,
                                 CastParameters &parameters) {
+	idx_t start_pos, end_pos;
+	bool is_negative, is_zero;
+	if (!Bignum::VarcharFormatting(input_value, start_pos, end_pos, is_negative, is_zero)) {
+		return false;
+	}
+
 	auto blob_string = Bignum::VarcharToBignum(input_value);
 
 	uint32_t blob_size = static_cast<uint32_t>(blob_string.size());
--- a/src/function/cast/bignum_casts.cpp
+++ b/src/function/cast/bignum_casts.cpp
@@ -173,7 +173,7 @@ bool TryCastToBignum::Operation(string_t input_value, bignum_t &result_value, Ve
 		return false;
 	}
 
-	auto blob_string = Bignum::VarcharToBignum(input_value);
+	auto blob_string = Bignum::EncodeBignum(input_value, start_pos, end_pos, is_negative, is_zero);
 
 	uint32_t blob_size = static_cast<uint32_t>(blob_string.size());
 	result_value = bignum_t(StringVector::EmptyString(result, blob_size));
--- a/src/common/types/bignum.cpp
+++ b/src/common/types/bignum.cpp
@@ -163,6 +163,65 @@ bool Bignum::VarcharFormatting(const string_t &value, idx_t &start_pos, idx_t &e
 	return true;
 }
 
+string Bignum::EncodeBignum(const string_t &value, idx_t start_pos, idx_t end_pos, bool is_negative, bool is_zero) {
+	if (is_zero) {
+		// Return Value 0
+		return InitializeBignumZero();
+	}
+	auto int_value_char = value.GetData();
+	idx_t actual_size = end_pos - start_pos;
+
+	// we initialize result with space for our header
+	string result(BIGNUM_HEADER_SIZE, '0');
+	unsafe_vector<uint64_t> digits;
+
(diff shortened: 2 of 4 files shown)


Schwarf:
Friendly ping ... is this PR still relevant? If not, I’m happy to close it.

src/function/cast/bignum_casts.cpp
```diff
@@ -167,6 +167,12 @@ bignum_t HugeintCastToBignum::Operation(hugeint_t int_value, StringHeap &heap) {
 template <>
 bool TryCastToBignum::Operation(string_t input_value, bignum_t &result_value, Vector &result,
                                 CastParameters &parameters) {
+	idx_t start_pos, end_pos;
+	bool is_negative, is_zero;
+	if (!Bignum::VarcharFormatting(input_value, start_pos, end_pos, is_negative, is_zero)) {
+		return false;
+	}
+
 	auto blob_string = Bignum::VarcharToBignum(input_value);
```
pdet: Bignum::VarcharToBignum runs the `VarcharFormatting` again, but in this case we are sure that the Varchar is a valid bignum, can we refactor the code to avoid that?

pdet (commented):
Hi, thanks for the PR, sorry for the belated review, I only had a small comment!

src/function/cast/bignum_casts.cpp
Schwarf: Hi @pdet,

I currently see two possible ways forward to avoid running `VarcharFormatting` twice.

1. Add an overload of `VarcharToBignum` that takes the already computed formatting information (`start_pos`, `end_pos`, `is_negative`, `is_zero`) and reuses it for the conversion.

2. Do a slightly larger refactor around `VarcharFormatting`. At the moment it seems to mix several responsibilities e.g. syntax validation, parsing the sign, trimming/handling leading zeros, and computing `start_pos`/`end_pos` for the later conversion. I would also be willing to put effort into this cleaner split, but only if you think that is desirable for this PR.

What do you think?

pdet: Maybe we can do something like, e.g.,

```cpp
  static string EncodeBignum(const string_t &value, idx_t start_pos, idx_t end_pos, bool is_negative, bool is_zero);

  string Bignum::VarcharToBignum(const string_t &value) {
      idx_t start_pos, end_pos;
      bool is_negative, is_zero;
      if (!VarcharFormatting(value, start_pos, end_pos, is_negative, is_zero)) {
          throw ConversionException("Could not convert string \'%s\' to Bignum", value.GetString());
      }
      return EncodeBignum(value, start_pos, end_pos, is_negative, is_zero);
  }
```

Schwarf:
Hi @pdet,
I hope I grasped your suggestion correctly. I rebased the branch onto the latest `upstream/main` and reran the fork CI, but the fork `Linux Release` build still fails with what looks like an unrelated extension build error:

```cpp
test_utils_extension.hpp:11:14: error:
‘void duckdb::TestUtilsExtension::Load(duckdb::DuckDB&)’ marked ‘override’, but does not override

**Comment by Schwarf** [comment]
Hi @pdet,

The `TidyCheck` failure is in an untouched example file:

```cpp
examples/embedded-c++-windows/cppintegration.cpp:11:1: error:
do not use namespace using-directives; use using-declarations instead
```
Could you please retrigger this job?

Schwarf:
Hi @pdet, 
could you take another look when you have a moment? I implemented the EncodeBignum split you suggested.

pdet:
Thanks, LGTM!

PR #22452, opened 2026-05-03
Fix eviction size metrics report

dentiny:
The issue is:
- When we increment `evicted_data_per_tag`, we increment by compressed size, [if applicable](https://github.com/duckdb/duckdb/blob/6be1980b1e1b82f15553802bcc76aad270008ec1/src/storage/standard_buffer_manager.cpp#L491)
- When we decrement, it's reduced by decompressed size

I found this issue by [asserting](https://github.com/dentiny/duckdb/pull/95) the `evicted_data_per_tag` value cannot underflow.


Mytherin:
Thanks!

PR #22454, opened 2026-05-04
Adding extension aliasing

ccfelius:
This PR introduces aliasing for extensions when loading extensions, e.g.

`LOAD loadable_extension_demo as demo`

For now, we can load extensions with the same name at the same time, if the user provides a different alias. For example 

```
LOAD loadable_extension_demo as demo
LOAD loadable_extension_demo as demo2
```

will work, _if the function signatures in these extensions differ_. Otherwise it will throw an error since the function is already registered in the function catalog.

```
LOAD loadable_extension_demo as demo
LOAD loadable_extension_demo 
```

Will throw an error.

Note that calling functions from extensions like

```
LOAD loadable_extension_demo as demo
demo.test_alias_hello()
```

will not work yet.

We will introduce explicit loading of extensions in separate extension schema's gradually in follow up PRs


ccfelius:
@Maxxen

Maxxen:
LGTM implementation-wise, but I have two questions:

What is the actual difference between an extension "name", "alias" and "internal alias" and is there a reason why we use different terminology for each?

Why can't we load an extension twice, with and without an alias? E.g. why does:
```
LOAD loadable_extension_demo as demo
LOAD loadable_extension_demo 
``` 
need to fail? The extension has different names in both cases right? (`demo` and `loadable_extension_demo`)

ccfelius:
> What is the actual difference between an extension "name", "alias" and "internal alias" and is there a reason why we use different terminology for each?

Internal alias is static and resolved at build-time. Extension name is the original extension name until the function is registered in the catalog. When this happens, the extension name will be the alias used, if any is given. My thought was that it would be more 'safe' to have the original extension name around for a while, and that the original name can always be retrieved by the alias.

> Why can't we load an extension twice, with and without an alias? E.g. why does:
> 
> ```
> LOAD loadable_extension_demo as demo
> LOAD loadable_extension_demo 
> ```
> 
> need to fail? The extension has different names in both cases right? (`demo` and `loadable_extension_demo`)

It does have two different names and it does not _need_ to fail. I just though it would be nicer to be more explicit about this, since e.g. there _might_ be cases where autoloading an already loaded aliased extension can run into (more) vague errors otherwise.

Mytherin:
> It does have two different names and it does not need to fail. I just though it would be nicer to be more explicit about this, since e.g. there might be cases where autoloading an already loaded aliased extension can run into (more) vague errors otherwise.

In e.g. Python this is allowed but what happens is that we only load the extension once. Setting up the alias only affects how the user can access the functions in the extension,  e.g.:

```py
>>> import pandas
>>> import pandas as pd
>>> print(pandas.DataFrame == pd.DataFrame)
True
```

I think that could be explored in a follow-up PR though.

Maxxen (approved):

Mytherin:
Thanks!

PR #22456, opened 2026-05-04
`OVERLAY()` string function

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/9105
Support OVERLAY function in parser
The PostgreSQL parser had support for the OVERLAY function, but the new parser does not (yet)


src/parser/peg/grammar/statements/expression.gram
```diff
@@ -250,6 +250,10 @@ TrimBoth <- 'BOTH'
 TrimLeading <- 'LEADING'
 TrimTrailing <- 'TRAILING'
 TrimSource <- Expression? 'FROM'
+OverlayExpression <- 'OVERLAY' Parens(OverlayArguments)
+OverlayArguments <- OverlayParameters / OverlayExpressionList
+OverlayParameters <- Expression 'PLACING' Expression 'FROM' Expression ('FOR' Expression)?
```
Dtenwolde: Looks good. What'd make this cleaner is moving the `'FROM' Expression` and `'FOR' Expression` into two separate rules. The latter being optional here of course. 
These can then be reused by the `SubstringParameters` grammar rule: 
`SubstringParameters <- Expression 'FROM' Expression 'FOR' Expression`

test/sql/function/string/overlay.test
```diff
@@ -0,0 +1,68 @@
+# name: test/overlay.test
+# group: [test]
+
+query I
```
Dtenwolde, line 4: What happens with negative `FROM` or `FOR`?
```sql
SELECT overlay('w3333333rce' placing 'resou' from 0);
```

Dtenwolde:
Oh and please also add the return types `unique_ptr<ParsedExpression>` to `scripts/parser/grammar_types.yml` (Currently not enforced, but will be used in the future to help transformer code is the idea)

src/include/duckdb/parser/peg/inlined_grammar.gram
```diff
@@ -790,7 +790,7 @@ MapExpression <- 'MAP' MapStructExpression
 MapStructExpression <- '{' List(MapStructField)? '}'
 MapStructField <- Expression ':' Expression
 
-GroupingExpression <- GroupingOrGroupingId Parens(List(Expression))
+GroupingExpression <- GroupingOrGroupingId Parens(List(Expression)?)
```
Dtenwolde, line 793: Did this sneak in there?

Dtenwolde: Maybe the `inlined_grammar.gram` file wasn't generated in the previous PR?

scripts/parser/grammar_types.yml
```diff
@@ -358,6 +358,7 @@ Expression:
     - SubstringExpression
     - TrimExpression
     - TrimSource
+    - OverlayExpression
```
Dtenwolde, line 361: All the new grammar rules should be listed here :)

test/sql/function/string/overlay.test
```diff
@@ -0,0 +1,100 @@
...
+----
+resou3rce
+
+query I
+SELECT overlay('33333rce' placing 'resou' from -1);
+----
+resou33rce
+
+query I
+SELECT overlay('33333rce' placing 'resou' from -100);
```
Dtenwolde, line 72: Are these results in line with PostgreSQL or another system? :)

test/sql/function/string/overlay.test
yan-alex: With negative FROM it replaces less characters of the original string.
With negative FOR, it's the same as not using FOR. (I followed Spark's functionality here)

Dtenwolde (changes requested):

src/include/duckdb/parser/peg/inlined_grammar.gram
Dtenwolde: Should be fine to leave in! Just make sure to use `./scripts/parser/build_grammar.sh` to generate the grammar files rather than calling `inline_grammar.py`

yan-alex: Yep I missed it on the last PR.

test/sql/function/string/overlay.test
yan-alex: In line with Spark

scripts/parser/grammar_types.yml
yan-alex: Done!

Dtenwolde (approved):
Thanks!

src/function/scalar/string/overlay.cpp
```diff
@@ -0,0 +1,126 @@
...
+		auto inputs = UnifiedVectorFormat::GetData<string_t>(input_data);
+		auto repls = UnifiedVectorFormat::GetData<string_t>(repl_data);
+		auto starts = UnifiedVectorFormat::GetData<int64_t>(start_data);
+		auto counts = UnifiedVectorFormat::GetData<int64_t>(count_data);
+
+		result.SetVectorType(VectorType::FLAT_VECTOR);
+		auto result_entries = FlatVector::GetDataMutable<string_t>(result);
+		auto &result_validity = FlatVector::ValidityMutable(result);
+
+		for (idx_t i = 0; i < args.size(); i++) {
```
Mytherin: Let's use the new vector iteration APIs (`input_data.Values<string_t>()`, etc) and the `FlatVector::Writer` instead of using the unified vector format here.

src/function/scalar/string/overlay.cpp
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	idx_t byte_start = CharPosToByte(input_data, input_size, start);
+	idx_t byte_end = CharPosToByte(input_data, input_size, start + count);
+
+	idx_t before_size = byte_start;
+	idx_t after_size = input_size - byte_end;
+	idx_t total_size = before_size + repl_size + after_size;
+
+	auto result_string = StringVector::EmptyString(result, total_size);
```
Mytherin: `auto &blob = result_data.WriteEmptyString(blob_size);` (once we use the writer).

src/function/scalar/string/overlay.cpp
```diff
@@ -0,0 +1,126 @@
...
+			    !start_data.validity.RowIsValid(start_idx) || !count_data.validity.RowIsValid(count_idx)) {
+				result_validity.SetInvalid(i);
+				continue;
+			}
+			int64_t count_value = counts[count_idx] < 0 ? (int64_t)repls->GetSize() : counts[count_idx];
+			result_entries[i] =
+			    OverlayScalar(result, inputs[input_idx], repls[repl_idx], starts[start_idx], count_value);
+		}
+	} else {
+		TernaryExecutor::Execute<string_t, string_t, int64_t, string_t>(
```
Mytherin: Maybe we can just check if the count parameter is present to avoid having to also use the `TernaryExecutor`? e.g. have an `optional_ptr<VectorIterator<int64_t>> counts`

Mytherin (commented):
Thanks! Looks good - left some comments

Mytherin:
Thanks!

PR #22393, opened 2026-04-30
Fix validity map bisect underflow

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22392

In the provided sql test, `mid` becomes 0 and `right` underflows.

PR #21894, opened 2026-04-06
Fix parquet metadata type casting

dentiny:
Closes https://github.com/duckdb/duckdb-fuzzer/issues/4381

The issue is not related to external file cache (as the stacktrace shows), but the parquet metadata is invalid from the start.
See analysis at https://github.com/duckdb/duckdb-fuzzer/issues/4381#issuecomment-4190542498

So prevent silent overflow/underflow problem, in this PR I explicitly use `NumericCast` to replace implicit cast, and also check whether certain fields are valid.

After this change, the error message becomes the following instead of INTERNAL exception
```sh
memory D from read_parquet('/tmp/20260406_718d2a.parquet', filename='my_cool_filename');
Invalid Input Error:
Failed to read file "/tmp/20260406_718d2a.parquet": metadata is corrupt. Column has invalid data page offset (-256)
```

PR #22457, opened 2026-05-04
Introduce checked integer numeric util class

dentiny:
Recently I noticed and fixed a few bugs related to integer operation underflow and overflow (ex: https://github.com/duckdb/duckdb/pull/21894, https://github.com/duckdb/duckdb/pull/21917, https://github.com/duckdb/duckdb/pull/22452, https://github.com/duckdb/duckdb/pull/22393), so I think it'd be helpful to introduce an util class to automatically detect bad operations and prevent program going worse.
Briefly discuss with Laurens here and offline: https://github.com/duckdb/duckdb/pull/21917#discussion_r3044523529

A few design goals and trade-offs:
- Only support integer type operations, without floating point support
- Works as a drop-in replacement as plain numeric types as possible
- Able to detect underflow and overflow during operation and assignment
- Exception could be customized on different usage
  + Metrics collection underflow should be InternalException since it's DuckDB bug, invalid parquet metadata could be InvalidInputException since it's coming from corrupted and invalid parquet files

To prove its usefulness, in this PR I integrated with StandardBufferManager, and it found out a new bug: https://github.com/duckdb/duckdb/pull/22452 :)

Followup items:
- This class adds little perf overhead, on performance critical path we could enable it only under DEBUG/ASSERTION mode, if necessary
  + StandardBufferManager, which is updated in this PR, is not sensitive to this (IO dominates)
- We could identify more use cases to adopt this util class
- Modulo operation is not supported -- it's much tricker than I thought

I've ran CI in my forked branch (for most of the commits): https://github.com/dentiny/duckdb/pull/95 (CI failure seems unrelated)


src/include/duckdb/common/checked_integer.hpp
```diff
@@ -0,0 +1,666 @@
...
+//!
+//! A few key features:
+//! - On underflow and overflow, it throws a templated exception type (by default InternalException).
+//! - It's by default initialized to 0, which avoids accessing uninitialized memory.
+//! - It serves as drop-in replacement for the underlying type in most cases.
+//!
+//! Known and intentional incompatibilities:
+//! - For all the operations, only integral template arguments are accepted.
+//! - It doesn't support bit operations.
+//! - Modulo operation is not supported.
```
dentiny, line 34: Mod operation is trickier than I thought
```cpp
std::cerr << int64_t(-17) % uint64_t(5) << std::endl; // prints 4
std::cerr << (-17 % 5) << std::endl; // prints -2
```

Mytherin:
Thanks - I think this is a good idea but I've already added something very similar in https://github.com/duckdb/duckdb/pull/21659 (see [ubigint.hpp](https://github.com/duckdb/duckdb/blob/main/src/include/duckdb/common/ubigint.hpp)). 

Maybe we could merge those efforts and expand `ubigint` to also have other similar classes (`uinteger_t`, `integer_t`, etc) which could potentially all be derived from `CheckedInteger`?

dentiny:
Ah I didn't notice that, and yes they serve for the same purpose.

For migration, I think we provide type alias based on the templated `CheckedInteger`?
For example, 
```cpp
using ubigint_t = CheckedInteger<uint64_t>;
using bigint_t = CheckedInteger<int64_t>;
// type alias for other types
```
If you're ok with it, migration should be fairly smooth. :)

Have to call out there're two functionality gaps for `CheckedInteger` compared with `ubigint_t`: 
- modulo operation is not implemented in this PR, which I plan to followup in next PR.
  + The main blocker is I haven't thought out a good way to reason  https://github.com/duckdb/duckdb/pull/22457#discussion_r3180309468
- I use INTERNAL exception since I'd love to keep exception consistency with [numeric cast failure](https://github.com/duckdb/duckdb/blob/bca135e67f8a803b29ae97a53a72702009c28d0d/src/include/duckdb/common/numeric_utils.hpp#L50-L51)

Mytherin:
The way I see `ubigint` and friends is that you would use them in locations where there *might* be an overflow, in particular in e.g. user-provided data (such as in `list_resize` which is the case I implemented it for). In that case this is not tracking an internal property and an internal exception is inappropriate. Maybe this can be configurable in the `CheckedInteger` class?

dentiny:
> Maybe this can be configurable in the CheckedInteger class?

Yes the exception type is templatized and configurable for checked integer.

dentiny:
A few changes since last review:
- Create alias for all integer types in STL
- Use OutOfRange exception as the default per https://github.com/duckdb/duckdb/pull/22457#issuecomment-4371392295
- Delete current `ubigint` implementation to avoid duplication

Mytherin:
Great stuff, thanks!

PR #22164, opened 2026-04-20
Fix window self join optimizer

kryonix:
The`WindowSelfJoinOptimizer` used a home grown re-implementation of a `LogicalOperator` copy + table index remapping. However, we already have that functionality as `LogicalOperatorDeepCopy` for CTE inlining. This PR therefore removes all the duplication and changes the optimizer to use the existing DeepCopy framework instead. This fixes a bug where a table index could still be duplicated, but more importantly, significantly simplifies the code.

fix: https://github.com/duckdb/duckdb/issues/22133

PR #22460, opened 2026-05-04
Re-apply Window Self Join Optimizer Fix

kryonix:
Re-apply https://github.com/duckdb/duckdb/pull/22164 to main.

fix: https://github.com/duckdblabs/duckdb-internal/issues/9010


Mytherin:
Thanks!

PR #22461, opened 2026-05-04
test: per-test-file parallelism for unittest (--jobs / --shard-*)

rustyconover:
## Summary

Adds `--jobs N`, `--shard-index K`, and `--shard-count N` to the unittest binary so test files can run concurrently as separate child processes.

When `--jobs > 1` (or any `--shard-*` flag is set), the binary runs as a coordinator: it enumerates the tests matching the user's filter via `Catch::filterTests`, then spawns up to N children via `posix_spawn`. Each child runs exactly one `.test` file (= one Catch2 TEST_CASE) in its own `--test-temp-dir`. The coordinator captures each child's output via pipe and prints atomically as each child exits, using the same prefix format as Catch2's `renderTestProgress`:

```
$ ./build/release/test/unittest --jobs 8 "test/sql/aggregate/aggregates/*"
Running 47 tests with 8 parallel workers
[1/47] (2%): test/sql/aggregate/aggregates/test_count.test
[2/47] (4%): test/sql/aggregate/aggregates/test_count_star.test
...
[47/47] (100%): test/sql/aggregate/aggregates/test_sum.test

====================================================
Results: 47/47 passed, 0 failed
```

When `--jobs` is 1 (default), behavior is identical to before — no risk to existing CI.

## Details

* `-j N` is the short form of `--jobs N`.
* `--shard-index K --shard-count N` round-robins matched tests across shards (`i % count`). Useful for CI matrices.
* Catch's `-a` / `-x N` are observed by the coordinator: it stops launching new children on the threshold and SIGTERMs in-flight ones. The flag is also forwarded to children so each child honors it within its single-test run.
* `-h` / `-l` / `-t` / `--list-*` defer to `session.run()` so metadata queries don't spawn children.
* The new flags are registered with Catch's Clara parser so they appear in `unittest -h` output.
* The coordinator's auto-generated base temp dir uses `mkdtemp` under `TMPDIR` (falling back to `/tmp`). User-supplied `--test-temp-dir` mirrors the serial path's *must-not-pre-exist* contract.
* Children inherit HOME isolation via the existing serial-path `setenv`.
* `SIGINT` / `SIGTERM` propagate to all children; coordinator returns 130.

## Platform support

POSIX-only: on Windows, `--jobs` / `--shard-*` fall back to serial mode with a one-line note. A `CreateProcess` + `WaitForMultipleObjects` port is straightforward but not in this PR — happy to do it as a follow-up if maintainers prefer to land both at once.

## Known limitations (suitable for follow-up issues)

* **Reporter aggregation:** with `-r junit/xml`, each child writes its own output and they overwrite each other. A coordinator-level merging reporter would close this gap.
* **Per-child startup amortisation:** every child re-loads DuckDB. A worker-pool variant where each child runs a queue of tests would win significantly for short tests. Trade-off chosen here favors implementation simplicity and hard isolation between tests.

## Test plan

- [x] `--jobs 1` (default) — identical to existing behavior; full DuckDB unittest suite passes.
- [x] `--jobs N` for N > 1 — multiple `.test` files run in parallel; output matches Catch2's progress format; final summary correct.
- [x] `--shard-index K --shard-count N` — non-overlapping partitions; together they cover all matched tests.
- [x] `-a` / `-x N` — coordinator stops launching after threshold; in-flight children receive SIGTERM.
- [x] `-h` / `-l` / `-t` — defer to `session.run()`; no children spawned.
- [x] `--test-temp-dir <path>` — `must-not-pre-exist` contract honored in both serial and coordinator modes.
- [x] `TMPDIR=/some/path` — coordinator base dir is created under it (verified via `ls`).
- [x] Failed test under `--jobs N` — child's full stdout/stderr is printed atomically; aggregate failure list at end.
- [x] No CI matrix entry yet — happy to wire `--jobs` into one of the existing workflows; would appreciate a pointer to the preferred entry point.

## Stats

`test/unittest.cpp`: +674 / −43 lines (single file change).

### DIFF
--- a/test/unittest.cpp
+++ b/test/unittest.cpp
@@ -8,7 +8,451 @@
 #include "test_helpers.hpp"
 #include "test_config.hpp"
 
+#include <algorithm>
+#include <chrono>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <deque>
+#include <set>
+#include <string>
+#include <vector>
+
+#ifndef _WIN32
+#include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <signal.h>
+#include <spawn.h>
+#include <sys/wait.h>
+#include <unistd.h>
+extern char **environ;
+#endif
+
 using namespace duckdb;
+
+// =============================================================================
+// Per-test-file parallelism for unittest.
+//
+// When the user passes --jobs N (or --shard-index/--shard-count), this binary
+// runs as a coordinator: it enumerates the tests that match the user's filter,
+// then spawns up to N child processes via posix_spawn, each running exactly one
+// .test file (= one Catch2 TEST_CASE) in its own --test-temp-dir. Children's
+// output is captured per-pipe and printed atomically as each child exits.
+//
+// When --jobs is 1 (default), behavior is identical to the prior single-process
+// implementation.
+//
+// POSIX-only for now: on Windows --jobs / --shard-* fall back to serial mode
+// with a warning. A CreateProcess + WaitForMultipleObjects port is straight-
+// forward future work but not in this PR.
+//
+// Known limitations (worth a follow-up issue, not in this PR):
+//   * Reporter aggregation: with -r junit/xml, each child writes its own
+//     output file and they overwrite each other. A coordinator-level
+//     reporter that gives each child a unique --out and merges results
+//     would close this gap.
+//   * Per-child startup amortisation: every child re-loads DuckDB. A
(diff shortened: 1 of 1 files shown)


Mytherin:
We already have this in the form of various Python scripts (which also work on Windows). These are used in our CI. I would prefer not to duplicate this functionality in the unittest runner. We can perhaps move these scripts to separate location so that they can also be used in extensions.

rustyconover:
I'd like that, because right now without parallelism it takes a long time to run tests.

PR #22464, opened 2026-05-04
[Parquet][Variant] Fix ability to disable shredding of VARIANT columns

Tishj:
We previously used `ANY` to disable shredding, with the `SHREDDING` copy option.
But with the change to string->type parsing, `ANY` is no longer an accepted option, so I've switched it to `NULL`, since we can't shred on NULL anyways.

I've added tests that use this now, verifying that we can correctly read written variants that aren't shredded.


Mytherin:
Thanks!

PR #22469, opened 2026-05-04
Use parser cache for Appender flushes

Mike-Dax:
I was benchmarking the write performance of the Appender API and noticed flush-heavy workloads were slow.

This PR:

1. adds an appender flush-heavy benchmark
2. uses the parser cache for appender flushes
3. caches appender statement templates across flushes

I have a [second PR that builds on this](https://github.com/duckdb/duckdb/pull/22470) that adds an autovectorisable fast path for zonemap creation when appending contiguous buffers (which is why the benchmark is so heavily templated).

`./test/appender/test_appender.cpp` has an extra test that flushes between changing column configurations.

# Benchmarks

These results are the `AppendDataChunkI32AllValid1ColFlushEvery128Rows` benchmark but with `1024*1024` rows (instead of 100k) of an i32 column which flushes every 128 rows.

- CPU: `Apple M1 Max`, 10 physical, 10 logical cores
- OS: `Darwin 24.6.0`, `arm64`
- RAM: `64.0 GiB`

| Commit | Samples |    Mean ms |              95% CI ms |    Rows/s | MB/s | Speedup vs previous |
| ------ | ------: | ---------: | ---------------------: | --------: | ---: | ------------------: |
| 1      |      15 | 13,687.567 | [13,620.014, 13,755.121] |    76,608 |  0.3 |               1.00x |
| 2      |      15 |  1,359.114 |   [1,345.502, 1,372.726] |   771,514 |  3.1 |              10.07x |
| 3      |      15 |    732.004 |     [716.819, 747.190] | 1,432,472 |  5.7 |               1.86x |

- CPU: `AMD Ryzen AI 7 PRO 350 w/ Radeon 860M`, 8 physical, 16 logical cores
- OS: `Linux 7.0.2-arch1-1`, `x86_64`
- RAM: `30.5 GiB`

| Commit | Samples |   Mean ms |            95% CI ms |    Rows/s | MB/s | Speedup vs previous |
| ------ | ------: | --------: | -------------------: | --------: | ---: | ------------------: |
| 1      |      15 | 8,024.012 | [8,001.742, 8,046.281] |   130,680 |  0.5 |               1.00x |
| 2      |      15 |   704.309 |   [700.126, 708.492] | 1,488,801 |  6.0 |              11.39x |
| 3      |      15 |   364.923 |   [361.539, 368.307] | 2,873,417 | 11.5 |               1.93x |

### DIFF
--- a/benchmark/micro/append.cpp
+++ b/benchmark/micro/append.cpp
@@ -1,5 +1,7 @@
 #include "benchmark_runner.hpp"
 #include "duckdb_benchmark_macro.hpp"
+#include "duckdb/common/types/data_chunk.hpp"
+#include "duckdb/common/vector/flat_vector.hpp"
 #include "duckdb/main/appender.hpp"
 
 using namespace duckdb;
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks for the PR!

The parser on `main` is still undergoing very active work in order to speed it up - it's likely that once we are done with it the speed-up from this parser cache will be much less significant than it is right now. Before we do any work on caching parsed statements I would wait until we are done with the work on improving parser performance. 

For the appender - it would make sense to use prepared statements and improve on the prepared statement cache. That would allow the appender to keep more state cached than only the parsed statement and would improve performance (potentially) much more. However, that's a much bigger rework / project than this PR.

PR #22470, opened 2026-05-04
Add contiguous numeric zonemap fast paths for appends

Mike-Dax:
~This builds on https://github.com/duckdb/duckdb/pull/22469~ This no longer builds on that PR.

I was benchmarking the write performance of the Appender API and noticed `appender.AppendDataChunk` wasn't as fast as I would have expected.

For my contiguous append f32 buffer use case, 30%+ of the time was spent in `duckdb::GreaterThan::Operation<float>(float const&, float const&)`. Despite the buffers being contiguous, the zonemap stats creation was not being autovectorised, and the comparison function was not being inlined.

I happened to read https://blobs.duckdb.org/slides/DiDi-04.pdf recently and figured the key normalisation could be applied to the floating point zonemap creation.

This PR adds contiguous buffer fast paths that are autovectorised.

- ~I've extended the benchmarks to cover i32 and f32 with and without nulls (I can add all the primitive types easily if you like?)~ I've added SQL benchmarks for all numeric types.
- The primitive integer fast path is a simple for loop that now triggers the autovectoriser.
- The floating point fast path is inspired by `./src/include/duckdb/common/radix.hpp` and is bit-manipulation heavy. It is ~30% faster than Radix::EncodeFloat in a microbenchmark (both end up being branchless when compiled).
- Nulls are handled via batching within valid runs.
- `./test/sql/storage/append_contiguous_numeric_stats.test` contains SQL tests (it wasn't obvious to me that this exercised the right path - I can write appender based tests if that's more clear?)

The floats are encoded into an order-preserving word, the stats are calculated in the loop, then the min and max are decoded back into floats. I've discarded the +0.0 and -0.0 distinction, as well as the exact bit payload of NaNs since I believe they're essentially unobservable and the NaN-payload-to-NaN-payload sort semantics were undefined.

Enabling LTO did inline the `GreaterThan` calls but didn't replicate these performance gains. Enabling LTO _and_ this PR increases the performance gains.

# Benchmarks

I tested on an M1 Macbook and a Ryzen system, the timings show the baseline per benchmark then with the changes. The benchmarks create `1024*1024` rows of a numeric column but do not flush within the loop.

The benchmarking environment wasn't _pristine_, so the few regressions are hopefully noise.

I've added just the 100k row versions of i32 and f32, all valid and with nulls to the benchmark file. These results are expanded:

<summary>

### M1

<details>

- OS: `Darwin 24.6.0`, `arm64`
- CPU: `Apple M1 Max`, 10 physical, 10 logical cores
- RAM: `64.0 GiB`

## Timings

| Commit   |  Type | Validity   | Samples | Mean ms |        95% CI ms |      Rows/s |    MB/s | Speedup vs previous |
| -------- |  ---- | ---------- | ------: | ------: | ---------------: | ----------: | ------: | ------------------: |
| baseline |  u32  | all valid  |      15 |   4.363 |   [4.215, 4.511] | 240,319,027 |   961.3 |               1.00x |
| baseline |  u32  | with nulls |      15 |   6.820 |   [6.609, 7.031] | 153,748,644 |   615.0 |               1.00x |
| baseline |  u16  | all valid  |      15 |   6.140 |   [6.048, 6.232] | 170,787,122 |   341.6 |               1.00x |
| baseline |  u16  | with nulls |      15 |   7.813 |   [7.581, 8.045] | 134,209,139 |   268.4 |               1.00x |
| baseline |  u8   | all valid  |      15 |   6.387 |   [6.280, 6.493] | 164,178,618 |   164.2 |               1.00x |
| baseline |  u8   | with nulls |      15 |   8.010 |   [7.740, 8.279] | 130,915,992 |   130.9 |               1.00x |
| baseline |  i32  | all valid  |      15 |   4.311 |   [4.170, 4.452] | 243,232,661 |   972.9 |               1.00x |
| baseline |  i32  | with nulls |      15 |   6.394 |   [6.260, 6.528] | 163,995,454 |   656.0 |               1.00x |
| baseline |  i16  | all valid  |      15 |   6.233 |   [6.131, 6.334] | 168,240,542 |   336.5 |               1.00x |
| baseline |  i16  | with nulls |      15 |   7.810 |   [7.694, 7.926] | 134,262,984 |   268.5 |               1.00x |
| baseline |  i8   | all valid  |      15 |   6.479 |   [6.357, 6.601] | 161,840,594 |   161.8 |               1.00x |
| baseline |  i8   | with nulls |      15 |   7.839 |   [7.594, 8.084] | 133,767,413 |   133.8 |               1.00x |
| baseline |  f32  | all valid  |      15 |  17.457 | [15.325, 19.590] |  60,065,761 |   240.3 |               1.00x |
| baseline |  f32  | with nulls |      15 |  18.343 | [17.938, 18.748] |  57,164,082 |   228.7 |               1.00x |
| baseline |  f64  | all valid  |      15 |  18.131 | [17.119, 19.144] |  57,832,474 |   462.7 |               1.00x |
| baseline |  f64  | with nulls |      15 |  18.825 | [18.489, 19.162] |  55,700,065 |   445.6 |               1.00x |
| changes  |  u32  | all valid  |      15 |   3.757 |   [3.629, 3.886] | 279,069,569 | 1,116.3 |               1.16x |
| changes  |  u32  | with nulls |      15 |   5.768 |   [5.649, 5.888] | 181,777,249 |   727.1 |               1.18x |
| changes  |  u16  | all valid  |      15 |   3.773 |   [3.687, 3.860] | 277,886,256 |   555.8 |               1.63x |
| changes  |  u16  | with nulls |      15 |   5.846 |   [5.735, 5.956] | 179,380,724 |   358.8 |               1.34x |
| changes  |  u8   | all valid  |      15 |   3.936 |   [3.897, 3.975] | 266,424,555 |   266.4 |               1.62x |
| changes  |  u8   | with nulls |      15 |   5.430 |   [5.355, 5.505] | 193,105,548 |   193.1 |               1.48x |
| changes  |  i32  | all valid  |      15 |   3.774 |   [3.639, 3.909] | 277,851,894 | 1,111.4 |               1.14x |
| changes  |  i32  | with nulls |      15 |   5.912 |   [5.754, 6.069] | 177,378,007 |   709.5 |               1.08x |
| changes  |  i16  | all valid  |      15 |   5.274 |   [3.694, 6.854] | 198,829,924 |   397.7 |               1.18x |
| changes  |  i16  | with nulls |      15 |   5.839 |   [5.710, 5.967] | 179,593,738 |   359.2 |               1.34x |
| changes  |  i8   | all valid  |      15 |   3.949 |   [3.862, 4.036] | 265,547,433 |   265.5 |               1.64x |
| changes  |  i8   | with nulls |      15 |   5.640 |   [5.484, 5.796] | 185,915,533 |   185.9 |               1.39x |
| changes  |  f32  | all valid  |      15 |   4.041 |   [3.866, 4.216] | 259,467,164 | 1,037.9 |               4.32x |
| changes  |  f32  | with nulls |      15 |   6.671 |   [5.520, 7.822] | 157,184,230 |   628.7 |               2.75x |
| changes  |  f64  | all valid  |      15 |   4.413 |   [4.255, 4.570] | 237,635,825 | 1,901.1 |               4.11x |
| changes  |  f64  | with nulls |      15 |   6.346 |   [6.142, 6.550] | 165,228,956 | 1,321.8 |               2.97x |

`Speedup vs previous` is computed per benchmark. The first commit for each benchmark is `1.00x`.

## Comparisons

| Type | Validity   | Old MB/s | New MB/s | Improvement |
| ---- | ---------- | -------: | -------: | ----------: |
| u32  | all valid  |    961.3 |  1,116.3 |       1.16x |
| u32  | with nulls |    615.0 |    727.1 |       1.18x |
| u16  | all valid  |    341.6 |    555.8 |       1.63x |
| u16  | with nulls |    268.4 |    358.8 |       1.34x |
| u8   | all valid  |    164.2 |    266.4 |       1.62x |
| u8   | with nulls |    130.9 |    193.1 |       1.48x |
| i32  | all valid  |    972.9 |  1,111.4 |       1.14x |
| i32  | with nulls |    656.0 |    709.5 |       1.08x |
| i16  | all valid  |    336.5 |    397.7 |       1.18x |
| i16  | with nulls |    268.5 |    359.2 |       1.34x |
| i8   | all valid  |    161.8 |    265.5 |       1.64x |
| i8   | with nulls |    133.8 |    185.9 |       1.39x |
| f32  | all valid  |    240.3 |  1,037.9 |       4.32x |
| f32  | with nulls |    228.7 |    628.7 |       2.75x |
| f64  | all valid  |    462.7 |  1,901.1 |       4.11x |
| f64  | with nulls |    445.6 |  1,321.8 |       2.97x |

</details>
</summary>

<summary>

### Ryzen

<details>

- OS: `Linux 7.0.2-arch1-1`, `x86_64`
- CPU: `AMD Ryzen AI 7 PRO 350 w/ Radeon 860M`, 8 physical, 16 logical cores
- RAM: `30.5 GiB`

## Timings

| Commit   | Type | Validity   | Samples | Mean ms |        95% CI ms |      Rows/s |    MB/s | Speedup vs previous |
| -------- | ---- | ---------- | ------: | ------: | ---------------: | ----------: | ------: | ------------------: |
| baseline | u32  | all valid  |      15 |   3.630 |   [3.505, 3.756] | 288,858,607 | 1,155.4 |               1.00x |
| baseline | u32  | with nulls |      15 |   4.801 |   [4.742, 4.861] | 218,392,669 |   873.6 |               1.00x |
| baseline | u16  | all valid  |      15 |   5.531 |   [5.145, 5.917] | 189,588,486 |   379.2 |               1.00x |
| baseline | u16  | with nulls |      15 |   6.154 |   [6.078, 6.231] | 170,376,420 |   340.8 |               1.00x |
| baseline | u8   | all valid  |      15 |   5.813 |   [5.434, 6.193] | 180,378,449 |   180.4 |               1.00x |
| baseline | u8   | with nulls |      15 |   6.791 |   [6.511, 7.072] | 154,403,683 |   154.4 |               1.00x |
| baseline | i32  | all valid  |      15 |   3.753 |   [3.502, 4.004] | 279,411,639 | 1,117.6 |               1.00x |
| baseline | i32  | with nulls |      15 |   5.478 |   [5.129, 5.826] | 191,427,493 |   765.7 |               1.00x |
| baseline | i16  | all valid  |      15 |   5.074 |   [4.975, 5.173] | 206,643,106 |   413.3 |               1.00x |
| baseline | i16  | with nulls |      15 |   6.544 |   [6.261, 6.827] | 160,236,351 |   320.5 |               1.00x |
| baseline | i8   | all valid  |      15 |   5.297 |   [5.241, 5.353] | 197,944,123 |   197.9 |               1.00x |
| baseline | i8   | with nulls |      15 |   6.393 |   [6.049, 6.738] | 164,014,265 |   164.0 |               1.00x |
| baseline | f32  | all valid  |      15 |  10.204 |  [9.368, 11.040] | 102,761,270 |   411.0 |               1.00x |
| baseline | f32  | with nulls |      15 |  11.013 | [10.839, 11.188] |  95,209,685 |   380.8 |               1.00x |
| baseline | f64  | all valid  |      15 |   9.850 |  [9.628, 10.072] | 106,453,696 |   851.6 |               1.00x |
| baseline | f64  | with nulls |      15 |  12.216 | [11.682, 12.750] |  85,835,812 |   686.7 |               1.00x |
| changes  | u32  | all valid  |      15 |   3.805 |   [3.556, 4.053] | 275,602,593 | 1,102.4 |               0.95x |
| changes  | u32  | with nulls |      15 |   5.426 |   [5.104, 5.747] | 193,264,524 |   773.1 |               0.88x |
| changes  | u16  | all valid  |      15 |   3.500 |   [3.324, 3.675] | 299,598,850 |   599.2 |               1.58x |
| changes  | u16  | with nulls |      15 |   4.814 |   [4.590, 5.038] | 217,827,080 |   435.7 |               1.28x |
| changes  | u8   | all valid  |      15 |   3.690 |   [3.273, 4.107] | 284,187,475 |   284.2 |               1.58x |
| changes  | u8   | with nulls |      15 |   4.464 |   [4.318, 4.611] | 234,878,519 |   234.9 |               1.52x |
| changes  | i32  | all valid  |      15 |   3.971 |   [3.691, 4.252] | 264,031,828 | 1,056.1 |               0.94x |
| changes  | i32  | with nulls |      15 |   4.802 |   [4.746, 4.858] | 218,353,255 |   873.4 |               1.14x |
| changes  | i16  | all valid  |      15 |   3.328 |   [3.275, 3.381] | 315,114,797 |   630.2 |               1.52x |
| changes  | i16  | with nulls |      15 |   5.273 |   [4.763, 5.784] | 198,839,979 |   397.7 |               1.24x |
| changes  | i8   | all valid  |      15 |   3.557 |   [3.454, 3.660] | 294,803,291 |   294.8 |               1.49x |
| changes  | i8   | with nulls |      15 |   4.835 |   [4.423, 5.247] | 216,874,966 |   216.9 |               1.32x |
| changes  | f32  | all valid  |      15 |   2.859 |   [2.764, 2.954] | 366,771,756 | 1,467.1 |               3.57x |
| changes  | f32  | with nulls |      15 |   4.371 |   [4.334, 4.408] | 239,901,164 |   959.6 |               2.52x |
| changes  | f64  | all valid  |      15 |   3.785 |   [3.714, 3.857] | 277,000,458 | 2,216.0 |               2.60x |
| changes  | f64  | with nulls |      15 |   5.838 |   [5.645, 6.030] | 179,624,503 | 1,437.0 |               2.09x |

`Speedup vs previous` is computed per benchmark. The first commit for each benchmark is `1.00x`.

## Comparisons

 | Type | Validity   | Old MB/s | New MB/s | Improvement |
 | ---- | ---------- | -------: | -------: | ----------: |
 | u32  | all valid  |  1,155.4 |  1,102.4 |       0.95x |
 | u32  | with nulls |    873.6 |    773.1 |       0.88x |
 | u16  | all valid  |    379.2 |    599.2 |       1.58x |
 | u16  | with nulls |    340.8 |    435.7 |       1.28x |
 | u8   | all valid  |    180.4 |    284.2 |       1.58x |
 | u8   | with nulls |    154.4 |    234.9 |       1.52x |
 | i32  | all valid  |  1,117.6 |  1,056.1 |       0.94x |
 | i32  | with nulls |    765.7 |    873.4 |       1.14x |
 | i16  | all valid  |    413.3 |    630.2 |       1.52x |
 | i16  | with nulls |    320.5 |    397.7 |       1.24x |
 | i8   | all valid  |    197.9 |    294.8 |       1.49x |
 | i8   | with nulls |    164.0 |    216.9 |       1.32x |
 | f32  | all valid  |    411.0 |  1,467.1 |       3.57x |
 | f32  | with nulls |    380.8 |    959.6 |       2.52x |
 | f64  | all valid  |    851.6 |  2,216.0 |       2.60x |
 | f64  | with nulls |    686.7 |  1,437.0 |       2.09x |

</details>
</summary>

My application-end benchmark (using [@duckdb/node-neo](https://github.com/duckdb/duckdb-node-neo), baseline v1.5.2-r.1) improved end-to-end with these two PRs (tested on the M1 Macbook):

| type |    size | baseline (MiB/s) | candidate (MiB/s) | candidate vs baseline improvement | candidate outcome |
| ---: | ------: | ---------------: | -------------------: | ---------------------------: | :------------------- |
|  f32 | 1048576 |           230.62 |              801.234 |                      3.47x | improved             |
|  i32 | 1048576 |          382.572 |              716.391 |                      1.87x | improved             |
|  u16 | 1048576 |          310.239 |              658.197 |                      2.12x | improved             |
|   u8 | 1048576 |           304.93 |              627.586 |                      2.05x | improved             |

### DIFF
--- a/src/storage/compression/fixed_size_uncompressed.cpp
+++ b/src/storage/compression/fixed_size_uncompressed.cpp
@@ -1,13 +1,21 @@
+#include "duckdb/common/helper.hpp"
+#include "duckdb/common/limits.hpp"
 #include "duckdb/common/types/null_value.hpp"
 #include "duckdb/common/types/vector.hpp"
+#include "duckdb/common/vector_operations/validity_executor.hpp"
 #include "duckdb/function/compression_function.hpp"
 #include "duckdb/storage/buffer_manager.hpp"
 #include "duckdb/storage/checkpoint/write_overflow_strings_to_disk.hpp"
 #include "duckdb/storage/segment/uncompressed.hpp"
+#include "duckdb/storage/statistics/base_statistics.hpp"
+#include "duckdb/storage/statistics/numeric_stats_traits.hpp"
+#include "duckdb/storage/statistics/stats_writer.hpp"
 #include "duckdb/storage/table/append_state.hpp"
 #include "duckdb/storage/table/column_data_checkpointer.hpp"
 #include "duckdb/storage/table/column_segment.hpp"
 #include "duckdb/storage/table/scan_state.hpp"
+
+#include <type_traits>
 
 namespace duckdb {
 
@@ -203,36 +211,240 @@ static unique_ptr<CompressionAppendState> FixedSizeInitAppend(ColumnSegment &seg
 	return make_uniq<CompressionAppendState>(std::move(handle));
 }
 
+template <class T>
+struct FixedSizeStatsWriterTraits {
+	using TYPE = typename std::remove_cv<T>::type;
+	static constexpr bool SUPPORTS_NUMERIC_STATS_WRITER =
+	    std::is_integral<TYPE>::value || std::is_floating_point<TYPE>::value || std::is_same<TYPE, hugeint_t>::value ||
+	    std::is_same<TYPE, uhugeint_t>::value;
+};
+
+template <class T>
+using FixedSizeStatsWriter = typename std::conditional<FixedSizeStatsWriterTraits<T>::SUPPORTS_NUMERIC_STATS_WRITER,
+                                                       StatsWriter<T>, StatsWriter<void>>::type;
+
+static constexpr idx_t FIXED_SIZE_UNCOMPRESSED_VALIDITY_EXTRACT_RUN_LENGTH = 16;
+
+template <class T>
+static inline void AppendContiguousValidFixedSizeValues(BaseStatistics &stats, FixedSizeStatsWriter<T> &writer,
+                                                        T *__restrict target, const T *__restrict source, idx_t count) {
+	D_ASSERT(count > 0);
+	if constexpr (FixedSizeStatsWriterTraits<T>::SUPPORTS_NUMERIC_STATS_WRITER) {
+		using OPERATIONS = NumericStatsTraits<T>;
+		FixedSizeStatsWriter<T> local_writer;
+		local_writer.SetHasValid();
+		for (idx_t i = 0; i < count; i++) {
+			const auto input = OPERATIONS::LoadInput(source + i);
+			OPERATIONS::StoreInput(target + i, input);
+			local_writer.UpdateMinMaxFromInput(input);
+		}
+		local_writer.Merge(writer);
+	} else {
+		writer.SetHasValid();
+		for (idx_t i = 0; i < count; i++) {
+			const auto value = source[i];
+			target[i] = value;
+			stats.UpdateNumericStats<T>(value);
+		}
+	}
+}
+
+template <class T>
+static inline void AppendContiguousFixedSizeValues(BaseStatistics &stats, T *__restrict target,
+                                                   const T *__restrict source, idx_t count) {
+	FixedSizeStatsWriter<T> writer;
+	AppendContiguousValidFixedSizeValues<T>(stats, writer, target, source, count);
+	writer.Merge(stats);
+}
+
+template <class T>
+static inline void AppendContiguousInvalidFixedSizeValues(FixedSizeStatsWriter<T> &writer, T *__restrict target,
+                                                          idx_t count) {
+	D_ASSERT(count > 0);
+	writer.SetHasNull();
+	if constexpr (FixedSizeStatsWriterTraits<T>::SUPPORTS_NUMERIC_STATS_WRITER) {
+		using OPERATIONS = NumericStatsTraits<T>;
+		const auto null_input = OPERATIONS::NullInput();
+		for (idx_t i = 0; i < count; i++) {
+			OPERATIONS::StoreInput(target + i, null_input);
+		}
+	} else {
+		for (idx_t i = 0; i < count; i++) {
+			target[i] = NullValue<T>();
+		}
+	}
+}
+
+// Validity slices are word-local fragments from ValidityExecutor. Numeric fixed-size types use NumericStatsTraits
+// for local min/max reduction; other fixed-size types update validity through StatsWriter<void>.
+template <class T>
+static inline void AppendContiguousValiditySlice(BaseStatistics &stats, FixedSizeStatsWriter<T> &writer,
+                                                 T *__restrict target, const T *__restrict source,
+                                                 const ValidityWordSlice &word) {
+	const auto slice_count = word.Count();
+	const auto has_valid = word.HasValid();
+	const auto has_invalid = word.HasInvalid();
+	if constexpr (FixedSizeStatsWriterTraits<T>::SUPPORTS_NUMERIC_STATS_WRITER) {
+		using OPERATIONS = NumericStatsTraits<T>;
+		FixedSizeStatsWriter<T> local_writer;
+		if (has_valid) {
+			local_writer.SetHasValid();
+		}
+		if (has_invalid) {
+			local_writer.SetHasNull();
+		}
+		for (idx_t i = 0; i < slice_count; i++) {
+			if (word.RowIsValid(i)) {
+				const auto input = OPERATIONS::LoadInput(source + i);
+				OPERATIONS::StoreInput(target + i, input);
+				local_writer.UpdateMinMaxFromInput(input);
+			} else {
+				OPERATIONS::StoreInput(target + i, OPERATIONS::NullInput());
+			}
+		}
+		local_writer.Merge(writer);
+	} else {
+		if (has_valid) {
+			writer.SetHasValid();
+		}
+		if (has_invalid) {
+			writer.SetHasNull();
+		}
+		for (idx_t i = 0; i < slice_count; i++) {
+			if (word.RowIsValid(i)) {
+				const auto value = source[i];
+				target[i] = value;
+				stats.UpdateNumericStats<T>(value);
+			} else {
+				target[i] = NullValue<T>();
+			}
+		}
+	}
+}
+
+template <class T>
+static inline void AppendContiguousNullableFixedSizeValues(BaseStatistics &stats, T *__restrict target,
+                                                           const T *__restrict source, const ValidityMask &validity,
+                                                           idx_t source_offset, idx_t count) {
+	FixedSizeStatsWriter<T> writer;
+	auto valid_func = [&](idx_t local_offset, idx_t append_count) {
+		AppendContiguousValidFixedSizeValues<T>(stats, writer, target + local_offset, source + local_offset,
+		                                        append_count);
+	};
+	auto invalid_func = [&](idx_t local_offset, idx_t append_count) {
+		AppendContiguousInvalidFixedSizeValues<T>(writer, target + local_offset, append_count);
+	};
+	auto validity_slice_func = [&](idx_t local_offset, ValidityWordSlice word) {
+		AppendContiguousValiditySlice<T>(stats, writer, target + local_offset, source + local_offset, word);
+	};
+
+	ValidityExecutor::Execute<FIXED_SIZE_UNCOMPRESSED_VALIDITY_EXTRACT_RUN_LENGTH>(
+	    validity, source_offset, count, valid_func, invalid_func, validity_slice_func);
+	writer.Merge(stats);
+}
+
+template <class T>
+static inline void AppendSelectedValidFixedSizeValues(BaseStatistics &stats, T *__restrict target,
+                                                      const T *__restrict source, const SelectionVector &sel,
+                                                      idx_t source_offset, idx_t count) {
+	D_ASSERT(count > 0);
+	if constexpr (FixedSizeStatsWriterTraits<T>::SUPPORTS_NUMERIC_STATS_WRITER) {
+		using OPERATIONS = NumericStatsTraits<T>;
+		StatsWriter<T> writer;
+		writer.SetHasValid();
+		for (idx_t i = 0; i < count; i++) {
+			const auto source_idx = sel.get_index(source_offset + i);
+			const auto input = OPERATIONS::LoadInput(source + source_idx);
+			OPERATIONS::StoreInput(target + i, input);
+			writer.UpdateMinMaxFromInput(input);
+		}
+		writer.Merge(stats);
+	} else {
+		stats.SetHasNoNullFast();
+		for (idx_t i = 0; i < count; i++) {
+			const auto source_idx = sel.get_index(source_offset + i);
+			const auto value = source[source_idx];
+			target[i] = value;
+			stats.UpdateNumericStats<T>(value);
+		}
+	}
+}
+
+template <class T>
+static inline void AppendSelectedNullableFixedSizeValues(BaseStatistics &stats, T *__restrict target,
+                                                         const T *__restrict source, const SelectionVector &sel,
+                                                         const ValidityMask &validity, idx_t source_offset,
+                                                         idx_t count) {
+	D_ASSERT(count > 0);
+	if constexpr (FixedSizeStatsWriterTraits<T>::SUPPORTS_NUMERIC_STATS_WRITER) {
+		using OPERATIONS = NumericStatsTraits<T>;
+		StatsWriter<T> writer;
+		bool has_valid = false;
+		bool has_null = false;
+		for (idx_t i = 0; i < count; i++) {
+			const auto source_idx = sel.get_index(source_offset + i);
+			if (validity.RowIsValid(source_idx)) {
+				has_valid = true;
+				const auto input = OPERATIONS::LoadInput(source + source_idx);
+				OPERATIONS::StoreInput(target + i, input);
+				writer.UpdateMinMaxFromInput(input);
+			} else {
+				has_null = true;
+				OPERATIONS::StoreInput(target + i, OPERATIONS::NullInput());
+			}
+		}
+		if (has_valid) {
+			writer.SetHasValid();
+		}
+		if (has_null) {
+			writer.SetHasNull();
+		}
+		writer.Merge(stats);
+	} else {
+		bool has_valid = false;
+		bool has_null = false;
+		for (idx_t i = 0; i < count; i++) {
+			const auto source_idx = sel.get_index(source_offset + i);
+			if (validity.RowIsValid(source_idx)) {
+				has_valid = true;
+				const auto value = source[source_idx];
+				target[i] = value;
+				stats.UpdateNumericStats<T>(value);
+			} else {
+				has_null = true;
+				target[i] = NullValue<T>();
+			}
+		}
+		if (has_valid) {
+			stats.SetHasNoNullFast();
+		}
+		if (has_null) {
+			stats.SetHasNullFast();
+		}
+	}
+}
+
 struct StandardFixedSizeAppend {
 	template <class T>
 	static void Append(BaseStatistics &stats, data_ptr_t target, idx_t target_offset, UnifiedVectorFormat &adata,
 	                   idx_t offset, idx_t count) {
 		auto sdata = UnifiedVectorFormat::GetData<T>(adata);
 		auto tdata = reinterpret_cast<T *>(target);
-		if (adata.validity.CanHaveNull()) {
-			for (idx_t i = 0; i < count; i++) {
-				auto source_idx = adata.sel->get_index(offset + i);
-				auto target_idx = target_offset + i;
-				bool is_null = !adata.validity.RowIsValid(source_idx);
-				if (!is_null) {
-					stats.SetHasNoNullFast();
-					stats.UpdateNumericStats<T>(sdata[source_idx]);
-					tdata[target_idx] = sdata[source_idx];
-				} else {
-					stats.SetHasNullFast();
-					// we insert a NullValue<T> in the null gap for debuggability
-					// this value should never be used or read anywhere
-					tdata[target_idx] = NullValue<T>();
-				}
+		if (!adata.sel->IsSet()) {
+			// Contiguous buffer fast path
+			auto source = sdata + offset;
+			auto target_data = tdata + target_offset;
+			if (adata.validity.CannotHaveNull()) {
+				AppendContiguousFixedSizeValues<T>(stats, target_data, source, count);
+			} else {
+				AppendContiguousNullableFixedSizeValues<T>(stats, target_data, source, adata.validity, offset, count);
 			}
+			return;
+		} else if (adata.validity.CanHaveNull()) {
+			AppendSelectedNullableFixedSizeValues<T>(stats, tdata + target_offset, sdata, *adata.sel, adata.validity,
+			                                         offset, count);
 		} else {
-			stats.SetHasNoNullFast();
-			for (idx_t i = 0; i < count; i++) {
-				auto source_idx = adata.sel->get_index(offset + i);
-				auto target_idx = target_offset + i;
-				stats.UpdateNumericStats<T>(sdata[source_idx]);
-				tdata[target_idx] = sdata[source_idx];
-			}
+			AppendSelectedValidFixedSizeValues<T>(stats, tdata + target_offset, sdata, *adata.sel, offset, count);
 		}
 	}
 };
@@ -259,6 +471,13 @@ idx_t FixedSizeAppend(CompressionAppendState &append_state, ColumnSegment &segme
 	auto target_ptr = append_state.handle.GetDataMutable();
 	idx_t max_tuple_count = segment.SegmentSize() / sizeof(T);
 	idx_t copy_count = MinValue<idx_t>(count, max_tuple_count - segment.count);
+
+	// Compress() flushes when Append() returns _fewer_ rows than requested. If the previous call filled this segment
+	// exactly, this call starts with no remaining capacity, so copy_count is zero.
+	// StandardFixedSizeAppend assumes at least one row, so if copy_count is zero, exit early here.
+	if (copy_count == 0) {
+		return 0;
+	}
 
 	OP::template Append<T>(stats, target_ptr, segment.count, data, offset, copy_count);
 	segment.count += copy_count;
--- a/benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+++ b/benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
@@ -1,0 +1,25 @@
+# name: ${FILE_PATH}
+# description: ${DESCRIPTION}
+# group: [zonemaps]
+#
+# VALUE_MODULUS is the repeat period for the row values. It needs to fit in the physical type, but doesn't need to match it.
+
+argument rows 50_000_000
+
+name Append ${TYPE_NAME} Zonemap Stats
+group zonemaps
+subgroup append_numeric_stats
+storage transient
+
+load
+SET threads=1;
+PRAGMA force_compression='uncompressed';
+CREATE TABLE source AS SELECT ((i % ${VALUE_MODULUS})::${TYPE}) AS value FROM range(${rows}) tbl(i);
+CREATE TABLE target(value ${TYPE});
+
+run
+INSERT INTO target SELECT value FROM source;
+
+cleanup
+DROP TABLE target;
+CREATE TABLE target(value ${TYPE});
--- a/benchmark/micro/zonemaps/append_numeric_stats_bigint.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_bigint.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_bigint.benchmark
+# description: Benchmark SQL append while computing BIGINT zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_bigint.benchmark
+DESCRIPTION=Benchmark SQL append while computing BIGINT zonemap statistics
+TYPE_NAME=BIGINT
+TYPE=BIGINT
+VALUE_MODULUS=10000000000
--- a/benchmark/micro/zonemaps/append_numeric_stats_decimal_18_6.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_decimal_18_6.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_decimal_18_6.benchmark
+# description: Benchmark SQL append while computing DECIMAL(18,6) zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_decimal_18_6.benchmark
+DESCRIPTION=Benchmark SQL append while computing DECIMAL(18,6) zonemap statistics
+TYPE_NAME=DECIMAL(18,6)
+TYPE=DECIMAL(18,6)
+VALUE_MODULUS=10000000000
--- a/benchmark/micro/zonemaps/append_numeric_stats_decimal_38_10.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_decimal_38_10.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_decimal_38_10.benchmark
+# description: Benchmark SQL append while computing DECIMAL(38,10) zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_decimal_38_10.benchmark
+DESCRIPTION=Benchmark SQL append while computing DECIMAL(38,10) zonemap statistics
+TYPE_NAME=DECIMAL(38,10)
+TYPE=DECIMAL(38,10)
+VALUE_MODULUS=10000000000
--- a/benchmark/micro/zonemaps/append_numeric_stats_decimal_4_1.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_decimal_4_1.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_decimal_4_1.benchmark
+# description: Benchmark SQL append while computing DECIMAL(4,1) zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_decimal_4_1.benchmark
+DESCRIPTION=Benchmark SQL append while computing DECIMAL(4,1) zonemap statistics
+TYPE_NAME=DECIMAL(4,1)
+TYPE=DECIMAL(4,1)
+VALUE_MODULUS=1000
--- a/benchmark/micro/zonemaps/append_numeric_stats_decimal_9_4.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_decimal_9_4.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_decimal_9_4.benchmark
+# description: Benchmark SQL append while computing DECIMAL(9,4) zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_decimal_9_4.benchmark
+DESCRIPTION=Benchmark SQL append while computing DECIMAL(9,4) zonemap statistics
+TYPE_NAME=DECIMAL(9,4)
+TYPE=DECIMAL(9,4)
+VALUE_MODULUS=100000
--- a/benchmark/micro/zonemaps/append_numeric_stats_double.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_double.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_double.benchmark
+# description: Benchmark SQL append while computing DOUBLE zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_double.benchmark
+DESCRIPTION=Benchmark SQL append while computing DOUBLE zonemap statistics
+TYPE_NAME=DOUBLE
+TYPE=DOUBLE
+VALUE_MODULUS=10000000000
--- a/benchmark/micro/zonemaps/append_numeric_stats_float.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_float.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_float.benchmark
+# description: Benchmark SQL append while computing FLOAT zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_float.benchmark
+DESCRIPTION=Benchmark SQL append while computing FLOAT zonemap statistics
+TYPE_NAME=FLOAT
+TYPE=FLOAT
+VALUE_MODULUS=10000000
--- a/benchmark/micro/zonemaps/append_numeric_stats_float_special_values.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_float_special_values.benchmark
@@ -1,0 +1,33 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_float_special_values.benchmark
+# description: Benchmark FLOAT append stats with sparse special floating-point values
+# group: [zonemaps]
+
+argument rows 50_000_000
+
+name Append FLOAT Special Values Zonemap Stats
+group zonemaps
+subgroup append_numeric_stats
+storage transient
+
+load
+SET threads=1;
+PRAGMA force_compression='uncompressed';
+CREATE TABLE source AS
+	SELECT
+		CASE
+			WHEN i % 4096 = 0 THEN 'nan'::FLOAT
+			WHEN i % 4096 = 1 THEN 'inf'::FLOAT
+			WHEN i % 4096 = 2 THEN '-inf'::FLOAT
+			WHEN i % 4096 = 3 THEN '-0.0'::FLOAT
+			WHEN i % 4096 = 4 THEN '0.0'::FLOAT
+			ELSE (((i % 1000000) - 500000)::FLOAT)
+		END AS value
+	FROM range(${rows}) tbl(i);
+CREATE TABLE target(value FLOAT);
+
+run
+INSERT INTO target SELECT value FROM source;
+
+cleanup
+DROP TABLE target;
+CREATE TABLE target(value FLOAT);
--- a/benchmark/micro/zonemaps/append_numeric_stats_hugeint.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_hugeint.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_hugeint.benchmark
+# description: Benchmark SQL append while computing HUGEINT zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_hugeint.benchmark
+DESCRIPTION=Benchmark SQL append while computing HUGEINT zonemap statistics
+TYPE_NAME=HUGEINT
+TYPE=HUGEINT
+VALUE_MODULUS=10000000000
--- a/benchmark/micro/zonemaps/append_numeric_stats_integer.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_integer.benchmark
@@ -1,0 +1,10 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_integer.benchmark
+# description: Benchmark SQL append while computing INTEGER zonemap statistics
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_integer.benchmark
+DESCRIPTION=Benchmark SQL append while computing INTEGER zonemap statistics
+TYPE_NAME=INTEGER
+TYPE=INTEGER
+VALUE_MODULUS=1000000000
--- a/benchmark/micro/zonemaps/append_numeric_stats_nullable.benchmark.in
+++ b/benchmark/micro/zonemaps/append_numeric_stats_nullable.benchmark.in
@@ -1,0 +1,27 @@
+# name: ${FILE_PATH}
+# description: ${DESCRIPTION}
+# group: [zonemaps]
+#
+# NULL_EXPR is evaluated over the generated row id i. Rows where it is true are NULL.
+
+argument rows 50_000_000
+
+name Append ${TYPE_NAME} Nullable ${SHAPE_NAME} Zonemap Stats
+group zonemaps
+subgroup append_numeric_stats_nullable
+storage transient
+
+load
+SET threads=1;
+PRAGMA force_compression='uncompressed';
+CREATE TABLE source AS
+	SELECT CASE WHEN ${NULL_EXPR} THEN NULL ELSE ((i % ${VALUE_MODULUS})::${TYPE}) END AS value
+	FROM range(${rows}) tbl(i);
+CREATE TABLE target(value ${TYPE});
+
+run
+INSERT INTO target SELECT value FROM source;
+
+cleanup
+DROP TABLE target;
+CREATE TABLE target(value ${TYPE});
--- a/benchmark/micro/zonemaps/append_numeric_stats_nullable_double_all_null.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_nullable_double_all_null.benchmark
@@ -1,0 +1,12 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_nullable_double_all_null.benchmark
+# description: Benchmark SQL append while computing DOUBLE zonemap statistics with all NULL values
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats_nullable.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_nullable_double_all_null.benchmark
+DESCRIPTION=Benchmark SQL append while computing DOUBLE zonemap statistics with all NULL values
+TYPE_NAME=DOUBLE
+TYPE=DOUBLE
+VALUE_MODULUS=1000000000
+SHAPE_NAME=all null
+NULL_EXPR=true
--- a/benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_1_per_64.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_1_per_64.benchmark
@@ -1,0 +1,12 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_1_per_64.benchmark
+# description: Benchmark SQL append while computing DOUBLE zonemap statistics with one NULL per validity word
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats_nullable.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_1_per_64.benchmark
+DESCRIPTION=Benchmark SQL append while computing DOUBLE zonemap statistics with one NULL per validity word
+TYPE_NAME=DOUBLE
+TYPE=DOUBLE
+VALUE_MODULUS=1000000000
+SHAPE_NAME=1 null per 64
+NULL_EXPR=(i % 64) IN (0)
--- a/benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_32_clustered_per_64.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_32_clustered_per_64.benchmark
@@ -1,0 +1,12 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_32_clustered_per_64.benchmark
+# description: Benchmark SQL append while computing DOUBLE zonemap statistics with half clustered NULLs per validity word
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats_nullable.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_32_clustered_per_64.benchmark
+DESCRIPTION=Benchmark SQL append while computing DOUBLE zonemap statistics with half clustered NULLs per validity word
+TYPE_NAME=DOUBLE
+TYPE=DOUBLE
+VALUE_MODULUS=1000000000
+SHAPE_NAME=32 clustered nulls per 64
+NULL_EXPR=(i % 64) > 31
--- a/benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_4_clustered_per_64.benchmark
+++ b/benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_4_clustered_per_64.benchmark
@@ -1,0 +1,12 @@
+# name: benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_4_clustered_per_64.benchmark
+# description: Benchmark SQL append while computing DOUBLE zonemap statistics with four clustered NULLs per validity word
+# group: [zonemaps]
+
+template benchmark/micro/zonemaps/append_numeric_stats_nullable.benchmark.in
+FILE_PATH=benchmark/micro/zonemaps/append_numeric_stats_nullable_double_nulls_4_clustered_per_64.benchmark
+DESCRIPTION=Benchmark SQL append while computing DOUBLE zonemap statistics with four clustered NULLs per validity word
+TYPE_NAME=DOUBLE
+TYPE=DOUBLE
+VALUE_MODULUS=1000000000
+SHAPE_NAME=4 clustered nulls per 64
+NULL_EXPR=(i % 64) < 4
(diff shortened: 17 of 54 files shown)


Mytherin:
Thanks for the PR!

Improving performance of floating point stats computation sounds like it could be useful - could you split this off from the other PR into a separate PR? Could you also add a SQL-only benchmark to showcase this performance improvement?

Mike-Dax:
I've rebased this onto main, no longer on top of the parser cache PR. I've replaced the C++ appender benchmarks with templated SQL benchmarks that exercise every numeric type (this PR improves performance for almost all numeric types).

As a non floating point example (on my M1 macbook), the old path did this for i32s:
 
```asm
; old, contiguous all-valid i32 path
; x8  = source pointer
; x9  = target pointer
; x19 = remaining count
; x20 = SegmentStatistics/BaseStatistics
; [x20,#0x30] = current min
; [x20,#0x40] = current max

ldr     w10, [x8]             ; load source[i]

ldr     w11, [x20, #0x30]     ; load stats min
cmp     w11, w10              ; compare min vs value
csel    w11, w11, w10, lt     ; min = old_min < value ? old_min : value
str     w11, [x20, #0x30]     ; store stats min every row

ldr     w11, [x20, #0x40]     ; load stats max
cmp     w11, w10              ; compare max vs value
csel    w10, w11, w10, gt     ; max = old_max > value ? old_max : value
str     w10, [x20, #0x40]     ; store stats max every row

ldr     w10, [x8], #0x4       ; reload source[i], advance source
str     w10, [x9], #0x4       ; copy to target, advance target

subs    x19, x19, #0x1        ; count--
b.ne    loop
```

And the new path:

```asm
; new, contiguous all-valid i32 path
; x10 = source pointer + 32
; x11 = target pointer + 32
; x12 = vectorized element count, multiple of 16
; v0/v2/v3/v4 = min accumulators
; v1/v5/v6/v7 = max accumulators

ldp     q16, q17, [x10, #-0x20] ; load i32 values 0..7
ldp     q18, q19, [x10], #0x40  ; load i32 values 8..15, advance source by 64B

stp     q16, q17, [x11, #-0x20] ; store i32 values 0..7
stp     q18, q19, [x11], #0x40  ; store i32 values 8..15, advance target by 64B

smin.4s v0, v16, v0             ; reduce min for values 0..3
smin.4s v2, v17, v2             ; reduce min for values 4..7
smin.4s v3, v18, v3             ; reduce min for values 8..11
smin.4s v4, v19, v4             ; reduce min for values 12..15

smax.4s v1, v16, v1             ; reduce max for values 0..3
smax.4s v5, v17, v5             ; reduce max for values 4..7
smax.4s v6, v18, v6             ; reduce max for values 8..11
smax.4s v7, v19, v7             ; reduce max for values 12..15

subs    x12, x12, #0x10         ; advance by 16 i32 values
b.ne    loop

; then after the loop, it does a horizontal reduction:

smax.4s v1, v1, v5
smax.4s v1, v1, v6
smax.4s v1, v1, v7
smaxv.4s s1, v1
fmov    w10, s1                 ; local max

smin.4s v0, v0, v2
smin.4s v0, v0, v3
smin.4s v0, v0, v4
sminv.4s s0, v0
fmov    w11, s0                 ; local min
```

So these changes are letting the compiler autovectorise the loop and do a 4x interleave to reduce the overall dependency chains.

The SQL benchmarks insert 50 million rows each. Here are the new benchmark results (I don't have access to the Ryzen at this second so I've tested on an older Intel machine)

- OS: Darwin 24.6.0, arm64
- CPU: Apple M1 Max, 10 physical, 10 logical cores
- RAM: 64.0 GiB

| Type           |       Old time ms (95% CI) |       New time ms (95% CI) |      Old MiB/s (95% CI) |      New MiB/s (95% CI) | Improvement |
| -------------- | -------------------------: | -------------------------: | ----------------------: | ----------------------: | ----------: |
| BIGINT         | 184.930 [176.673, 193.187] | 150.323 [144.799, 155.847] | 2062.8 [1974.6, 2159.2] | 2537.7 [2447.7, 2634.5] |       1.23x |
| DECIMAL(18,6)  | 156.979 [155.062, 158.896] | 131.477 [129.569, 133.385] | 2430.1 [2400.7, 2460.1] | 2901.4 [2859.9, 2944.1] |       1.19x |
| DECIMAL(38,10) | 604.156 [593.127, 615.185] | 609.924 [603.179, 616.669] | 1262.8 [1240.2, 1286.3] | 1250.9 [1237.2, 1264.9] |       0.99x |
| DECIMAL(4,1)   | 212.597 [211.584, 213.610] |    93.838 [91.944, 95.732] |    448.6 [446.5, 450.7] |  1016.3 [996.2, 1037.2] |       2.27x |
| DECIMAL(9,4)   | 141.435 [135.266, 147.604] | 108.061 [106.731, 109.391] | 1348.6 [1292.2, 1410.1] | 1765.1 [1743.6, 1787.1] |       1.31x |
| DOUBLE         | 764.034 [756.552, 771.516] | 142.842 [140.769, 144.915] |    499.3 [494.4, 504.2] | 2670.6 [2632.4, 2709.9] |       5.35x |
| FLOAT          | 781.979 [721.520, 842.438] | 112.199 [110.940, 113.458] |    243.9 [226.4, 264.4] | 1700.0 [1681.1, 1719.3] |       6.97x |
| HUGEINT        | 622.793 [616.810, 628.776] | 630.457 [627.109, 633.805] | 1225.0 [1213.4, 1236.9] | 1210.1 [1203.7, 1216.6] |       0.99x |
| INTEGER        | 152.995 [144.999, 160.991] | 122.722 [117.489, 127.955] | 1246.7 [1184.8, 1315.4] | 1554.2 [1490.6, 1623.4] |       1.25x |
| SMALLINT       | 228.790 [223.643, 233.937] | 107.875 [104.647, 111.103] |    416.8 [407.7, 426.4] |    884.1 [858.4, 911.3] |       2.12x |
| TINYINT        | 253.032 [227.408, 278.656] | 106.208 [102.500, 109.916] |    188.4 [171.1, 209.7] |    449.0 [433.8, 465.2] |       2.38x |
| UBIGINT        | 178.667 [170.107, 187.227] | 148.819 [146.232, 151.406] | 2135.1 [2037.5, 2242.5] | 2563.3 [2519.5, 2608.7] |       1.20x |
| UHUGEINT       | 617.282 [614.215, 620.349] | 632.957 [627.609, 638.305] | 1236.0 [1229.9, 1242.1] | 1205.4 [1195.3, 1215.6] |       0.98x |
| UINTEGER       | 153.407 [141.440, 165.374] | 120.259 [119.304, 121.214] | 1243.3 [1153.4, 1348.5] | 1586.0 [1573.5, 1598.7] |       1.28x |
| USMALLINT      | 230.206 [223.683, 236.729] | 108.949 [103.719, 114.179] |    414.3 [402.9, 426.4] |    875.3 [835.2, 919.5] |       2.11x |
| UTINYINT       | 236.498 [224.456, 248.540] | 106.245 [101.926, 110.564] |    201.6 [191.9, 212.4] |    448.8 [431.3, 467.8] |       2.23x |

- OS: Linux 6.17.0-20-generic, x86_64
- CPU: 13th Gen Intel(R) Core(TM) i7-13700KF, 16 physical, 24 logical cores
- RAM: 62.6 GiB

| Type           |       Old time ms (95% CI) |       New time ms (95% CI) |      Old MiB/s (95% CI) |      New MiB/s (95% CI) | Improvement |
| -------------- | -------------------------: | -------------------------: | ----------------------: | ----------------------: | ----------: |
| BIGINT         | 255.311 [251.210, 259.412] | 250.531 [249.556, 251.506] | 1494.1 [1470.5, 1518.5] | 1522.6 [1516.7, 1528.6] |       1.02x |
| DECIMAL(18,6)  | 153.977 [152.857, 155.097] | 146.810 [146.182, 147.438] | 2477.4 [2459.6, 2495.6] | 2598.4 [2587.3, 2609.6] |       1.05x |
| DECIMAL(38,10) | 328.058 [326.166, 329.950] | 326.871 [325.380, 328.362] | 2325.6 [2312.3, 2339.1] | 2334.1 [2323.5, 2344.8] |       1.00x |
| DECIMAL(4,1)   | 171.869 [170.314, 173.424] | 111.674 [111.307, 112.041] |    554.9 [549.9, 559.9] |    854.0 [851.2, 856.8] |       1.54x |
| DECIMAL(9,4)   | 134.311 [133.033, 135.589] | 122.806 [122.439, 123.173] | 1420.1 [1406.7, 1433.7] | 1553.1 [1548.5, 1557.8] |       1.09x |
| DOUBLE         | 260.681 [258.571, 262.791] | 166.013 [164.249, 167.777] | 1463.4 [1451.6, 1475.3] | 2297.8 [2273.7, 2322.5] |       1.57x |
| FLOAT          | 277.379 [276.029, 278.729] | 128.221 [127.949, 128.493] |    687.6 [684.3, 691.0] | 1487.6 [1484.4, 1490.7] |       2.16x |
| HUGEINT        | 437.727 [433.181, 442.273] | 439.755 [437.706, 441.804] | 1743.0 [1725.0, 1761.2] | 1734.9 [1726.9, 1743.0] |       1.00x |
| INTEGER        | 235.224 [231.635, 238.813] | 227.117 [226.602, 227.632] |    810.9 [798.7, 823.4] |    839.8 [837.9, 841.7] |       1.04x |
| SMALLINT       | 272.716 [268.768, 276.664] | 214.317 [213.682, 214.952] |    349.7 [344.7, 354.8] |    445.0 [443.7, 446.3] |       1.27x |
| TINYINT        | 232.499 [228.392, 236.606] | 210.925 [210.134, 211.716] |    205.1 [201.5, 208.8] |    226.1 [225.2, 226.9] |       1.10x |
| UBIGINT        | 252.576 [248.675, 256.477] | 251.605 [251.040, 252.170] | 1510.3 [1487.3, 1534.0] | 1516.1 [1512.7, 1519.6] |       1.00x |
| UHUGEINT       | 411.698 [407.462, 415.934] | 417.943 [416.150, 419.736] | 1853.2 [1834.3, 1872.4] | 1825.5 [1817.7, 1833.3] |       0.99x |
| UINTEGER       | 236.444 [232.633, 240.255] | 228.222 [227.619, 228.825] |    806.7 [793.9, 819.9] |    835.7 [833.5, 838.0] |       1.04x |
| USMALLINT      | 279.854 [275.849, 283.859] | 214.957 [214.443, 215.471] |    340.8 [336.0, 345.7] |    443.7 [442.6, 444.7] |       1.30x |
| UTINYINT       | 235.101 [231.022, 239.180] | 210.239 [209.656, 210.822] |    202.8 [199.4, 206.4] |    226.8 [226.2, 227.4] |       1.12x |

src/storage/compression/fixed_size_uncompressed.cpp
```diff
@@ -202,13 +206,273 @@ static unique_ptr<CompressionAppendState> FixedSizeInitAppend(ColumnSegment &seg
...
+	using type = FixedSizeGenericContiguousAppendTag;
+};
+
+#define DUCKDB_FIXED_SIZE_PRIMITIVE_INTEGER_APPEND_TYPE(T)                                                             \
+	template <>                                                                                                        \
+	struct FixedSizeContiguousAppendTag<T> {                                                                           \
+		using type = FixedSizePrimitiveIntegerContiguousAppendTag;                                                     \
+	};
+
+DUCKDB_FIXED_SIZE_PRIMITIVE_INTEGER_APPEND_TYPE(int8_t)
```
Mytherin: In general we prefer not to use defines - could you remove these?

src/storage/compression/fixed_size_uncompressed.cpp
```diff
@@ -202,13 +206,273 @@ static unique_ptr<CompressionAppendState> FixedSizeInitAppend(ColumnSegment &seg
...
+		T result;
+		Store<WORD>(bits, data_ptr_cast(&result));
+		return result;
+	}
+};
+
+template <class T>
+static inline void AppendContiguousFixedSizeValues(BaseStatistics &stats, T *__restrict target,
+                                                   const T *__restrict source, idx_t count,
+                                                   FixedSizeGenericContiguousAppendTag) {
```
Mytherin: Maybe I'm missing something but what's the point of this tag?

src/storage/compression/fixed_size_uncompressed.cpp
```diff
@@ -202,13 +206,273 @@ static unique_ptr<CompressionAppendState> FixedSizeInitAppend(ColumnSegment &seg
...
+		// of unsigned integer space with reversed order, so higher magnitude negative
+		// values sort before lower magnitude negative values.
+		auto key = bits ^ (SIGN_BIT | sign_mask);
+
+		// Clear the sign bit, leaving the exponent and fractional parts
+		const auto abs = bits & ABS_MASK;
+
+		// +0.0 and -0.0 are treated the same and will be reconstructed as +0.0
+		const auto zero_mask = Mask(abs == 0);
+		key = (key & ~zero_mask) | (SIGN_BIT & zero_mask);
```
Mytherin: This means this information is lost - previously this was kept around. I don't feel strongly about it - we could make this decision - but it's a change in behavior. For example currently this returns `-0.0`:

```sql
create table t(d double);
insert into t values ('-0.0'::double);
select min(d) from t;
┌────────┐
│ min(d) │
│ double │
├────────┤
│   -0.0 │
└────────┘
```

With this change we will now return `0.0`.

src/storage/compression/fixed_size_uncompressed.cpp
```diff
@@ -202,13 +206,273 @@ static unique_ptr<CompressionAppendState> FixedSizeInitAppend(ColumnSegment &seg
...
+	T min = NumericLimits<T>::Maximum();
+	T max = NumericLimits<T>::Minimum();
+	for (idx_t i = 0; i < count; i++) {
+		auto value = source[i];
+		target[i] = value;
+		min = value < min ? value : min;
+		max = value > max ? value : max;
+	}
+
+	stats.UpdateNumericStats<T>(min);
```
Mytherin: This should unfold to the same operation as what is going on in the loop - is that not happening? It would be nice if we could somehow fix that so that `UpdateNumericStats` has the same performance cost as the above code. It runs effectively the same code internally:

```cpp
min = LessThan::Operation(new_value, min) ? new_value : min;
max = GreaterThan::Operation(new_value, max) ? new_value : max;
```

src/storage/compression/fixed_size_uncompressed.cpp
```diff
@@ -202,13 +206,273 @@ static unique_ptr<CompressionAppendState> FixedSizeInitAppend(ColumnSegment &seg
...
+
+		if (valid_bits == active_mask) {
+			stats.SetHasNoNullFast();
+			AppendContiguousFixedSizeValues<T>(stats, target + local_offset, source + local_offset, bits_in_entry, tag);
+		} else if (valid_bits == 0) {
+			stats.SetHasNullFast();
+			FillNullValues<T>(target + local_offset, bits_in_entry);
+		} else {
+			stats.SetHasNullFast();
+			stats.SetHasNoNullFast();
```
Mytherin: I don't think this is necessarily true if `count` does not span the entire bit width (i.e. if `bits_in_entry` is < `ValidityMask::BITS_PER_VALUE`).

src/storage/compression/fixed_size_uncompressed.cpp
```diff
@@ -202,13 +206,273 @@ static unique_ptr<CompressionAppendState> FixedSizeInitAppend(ColumnSegment &seg
...
+template <class T, class TAG>
+static inline void AppendContiguousNullableFixedSizeValues(BaseStatistics &stats, T *__restrict target,
+                                                           const T *__restrict source, const ValidityMask &validity,
+                                                           idx_t source_offset, idx_t count, TAG tag) {
+	idx_t local_offset = 0;
+	// Process a validity-mask word at a time.
+	while (local_offset < count) {
+		idx_t entry_idx;
+		idx_t idx_in_entry;
+		ValidityMask::GetEntryIndex(source_offset + local_offset, entry_idx, idx_in_entry);
```
Mytherin: `idx_in_entry` should always be 0 if we are doing a batch operation like this, right?

src/storage/compression/fixed_size_uncompressed.cpp
Mytherin: Maybe this optimized loop is not necessary for the last entry, i.e. we run this only as long as the entries are full and have 64 values each? That might also speed up the loops as the compiler knows there's exactly 64 values to process.

src/storage/compression/fixed_size_uncompressed.cpp
```diff
@@ -202,13 +206,273 @@ static unique_ptr<CompressionAppendState> FixedSizeInitAppend(ColumnSegment &seg
...
+			// null_bits: 0 0 1 0 0 0 1 0
+			// iteration: [ 1 ] [  2  ] ^ leftover final valid run
+
+			auto null_bits = (~valid_bits) & active_mask;
+			idx_t cursor = 0;
+			while (null_bits) {
+				// Find the lowest set bit
+				const auto null_pos = CountZeros<validity_t>::Trailing(null_bits);
+				if (null_pos > cursor) {
+					AppendContiguousFixedSizeValues<T>(stats, target + local_offset + cursor,
```
Mytherin: This is interesting but seems somewhat optimized for one scenario (few NULL values) - it's likely this will cause a regression in performance if there are many `NULL` values as we're doing more stuff per NULL value than before. 

I wonder if we could make this "more adaptive" by e.g. looking at the amount of bits that are set.

Mytherin (commented):
Thanks for the update - nice results. I've left some comments

src/storage/compression/fixed_size_uncompressed.cpp
Mytherin: This type of optimized loop also already exists in `UnaryExecutor::ExecuteFlat`. It would be great if this could be unified so that we can directly use that instead of reimplementing that loop.

src/storage/compression/fixed_size_uncompressed.cpp
Mike-Dax: In `./src/include/duckdb/common/radix.hpp` the zero sign bit is destroyed during canonicalisation of floats, both `-0.0` and `+0.0` become `+0.0`. I was matching that and a bit unsure of how this was handled in the rest of DuckDB. I've poked around a bit more, here's what I'm thinking.

Currently:

```sql
CREATE TABLE t(d DOUBLE UNIQUE);
INSERT INTO t VALUES ('-0.0'::DOUBLE);
INSERT INTO t VALUES ('0.0'::DOUBLE);
```

The second insert fails as a duplicate key due to the ART index canonicalisation via radix.hpp.

On main the stats update becomes roughly the following for `-0.0` and `+0.0`:

```cpp
min = (min > new_value) ? new_value : min;
max = (new_value > max) ? new_value : max;
```

When multiple zeros of different signs are processed, the 'first processed wins', and it becomes sticky.

```cpp
// given
// -0.0 > +0.0 == false
// -0.0 < +0.0 == false
// (-0.0 == +0.0) == true

// `-0.0` then `+0.0`, on second iteration:
min = (-0.0 > +0.0) ? +0.0 : -0.0; // min = -0.0
max = (+0.0 > -0.0) ? +0.0 : -0.0; // max = -0.0

// `+0.0` then `-0.0`, on second iteration:
min = (+0.0 > -0.0) ? -0.0 : +0.0; // min = +0.0
max = (-0.0 > +0.0) ? -0.0 : +0.0; // max = +0.0
```

With the canonicalisation in the zonemap float encode:

```cpp
const auto zero_mask = Mask(abs == 0);
key = (key & ~zero_mask) | (SIGN_BIT & zero_mask);
```

We destroy that bit and canonicalise to `+0.0`, matching `radix.hpp`. But then this behaviour is weird:

```
create table t(d double);
insert into t values ('-0.0'::double);
select min(d) from t;
┌────────┐
│ min(d) │
│ double │
├────────┤
│   0.0  │
└────────┘
```

Removing the canonicalisation gives the following behaviour _within an append_, but cross-transaction appends become the same 'sticky' behaviour 

| Case           | Main                     | Potential behaviour      |
| -------------- | ------------------------ | ------------------------ |
| `+0.0`         | `+0.0`                   | `+0.0`                   |
| `-0.0`         | `-0.0`                   | `-0.0`                   |
| `-0.0`, `+0.0` | `min = -0.0, max = -0.0` | `min = -0.0, max = +0.0` |
| `+0.0`, `-0.0` | `min = +0.0, max = +0.0` | `min = -0.0, max = +0.0` |

But across transactions, we still have that sticky 'first set wins' behaviour.

I think `-0.0` should just be canonicalised to `+0.0` globally, so zonemaps and ART indexes are consistent?

src/storage/compression/fixed_size_uncompressed.cpp
Mike-Dax: For floats that's what's happening, with other types without specialisations, the `GreaterThan::Operation` etc calls are inlined. (If LTO is used the float specialisation is also inlined.)

So we're getting to this in the best case:

```cpp
for (...) {
  value = source[i];
  target[i] = value;

  stats_min = value < stats_min ? value : stats_min;
  stats_max = value > stats_max ? value : stats_max;
}
```

but that's still a per loop load and store. We want to get to:

```cpp
// hoisted local min / max declarations, loading once
T local_min = stats_min;
T local_max = stats_max;
for (...) {
  // local accumulator comparisons
  local_min = ...;
  local_max = ...;
}
// single store
stats_min = local_min;
stats_max = local_max;
```


so that the autovectoriser can transform it, but LLVM can't do the alias analysis / reduction recognition with what it's given.

https://llvm.org/docs/Vectorizers.html#diagnostics has some helpful diagnostic commands for this kind of thing.

Here's a [godbolt snippet](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEIADM0gAOqAqETgwe3r7%2BpClpjgIhYZEsMXGJtpj2hQxCBEzEBFk%2BfgF2mA4Z9Y0ExRHRsQnSCg1NLTntY32hA2VDiQCUtqhexMjsHCYaAILm8aHI3lgA1Cbxbsij6FhU59jbe2YHDEdep%2BeX16EE94%2BPo2IXgcJx6BAUZwA7FZdic4ScfvEzAB9AgnFihc4wnbwhGCJGo9FMVRY/6QgAipN2jwAbqg8OgTlEmOhkWwWCQAJ7I0ZMcEQNAMUZ4ggEtEAKgUq3WmFIIrFJ3FY2AmAIcuuIBAaQAXphCQw5WCIZKGuDFlDsbj%2BMQThANVq8LrCXgzvFyScNFiEa63CcGF68NZrOaTNDHrjcYiUWiaWIvJhXe6pWsNiYAKwWQNpynxS0RuHK1XpzPp93nd2x7yYKl7WH5k688EAWnuGIYiZOlfjPobpoULfi2DbroAYp24wmQL2%2Bf3W5jc%2BH643Z4OWMSO12E/dp83W%2BvzmPNycp8uB0PiTXcaHKdSKf9dnSGSdgKgMMjaKhRLQeX2BQJhVGhKStKGxyoBEqFmqvboJqOp6miBqgn2irLiGYZ1nC4EnB%2BX5sqEHannO/oLhh8rRthn5iGy%2B5ujuK7niSJG1ji8LWra9pwc6Haerm3qfH6AZBpYaF5vmWFHuWDYgdWGZZjmokRpBxZyRuE41ou%2BY4VRw6SRJFwUbhOnxIeE7HgZ2nzgpuJad%2Ba6qKpVautg5m2TRJmOVONnUYxCnXveLFwoRq74ZJXltpe8JBQxHZhReTHXhwyy0Jwaa8H4HBaKQqCcG4QkWFJKZbs8PCkAQmiJcsADWCTxAAdPEDWNU1TUAGz6JwkhpeVWWcLwCggBopXlcscCwEgaAsEkdCxOQlATVN9BxEchjAGYZhcINNC0AQsT9RAUTdVEoSNJynAlUdzDEJyADyUTaJ0ZXcLwE1sII10MLQp0ZbwWBRF4wBuGItD9U9pBYGuRjiN9YN4MQD14DSmAg5lmCqJ0Xg7WdvA/NU3W0HgUTECdHhYN1BDEHgLBY6QiPEFEqSYOSmAQ8A%2BNGMNfAGMACgAGp4JgADu11JIw1P8IIIhiOwUgyIIigqOo0O6Fw%2BgrSgeX6AT/WQMsqBJLUINNtd8QnE2ABKSRMAoCjlh%2BetNojDgkI6Cbm5b1tNhi1uYOgtuvkkDsPc7uqmxbVv9oYYicmkNtunbAeO0QFO6rwqC0xTWDaxAywdF0zgQK4ExtNIwSzKU5QJGmeSpOkAhF/4yQ17U/TlwsVe57UPTjJ4rQN1UNTdNMLeDBU7fTPXZgjEPZcj5XOfSRISUpV10PZRwJyqAAHC1TYtZIJzLUYJxrbVXC1Rotq4IQJBnMViy8I9WiLMsCCYCyQzZ6Q1UNfVzV/w1bVkocE6qQdKmU159QGkNb6I0YCIBAKsAgSQMazT/JNaaxBwisE2FvHee9nzIGQCcLgaZapmF4D7a%2BGc9Di2EF%2BaW0haHyzUN1ZWpABZEySFjJeHBUqgO6mva6GNkFolQFQDe29d770PsAY%2BZhT7n1tB4dBi1b7xC4PfaBT8qo1V/v/Zq7VgEr3Ab1WwUDH4VUMeQ/hq9TEWOfjTXaGQQCSCAA%3D%3D%3D).

We could do an `acc = stats.GetNumericStatsAccumulator<T>()` then `acc.Update(value)` and `stats.StoreNumericStatsFromAccumulator(acc)` type thing, where `acc` is some local object?

src/storage/compression/fixed_size_uncompressed.cpp
Mike-Dax: I'll have a crack at an adaptive one that can be reused.

src/storage/compression/fixed_size_uncompressed.cpp
Mike-Dax: Sure

src/storage/compression/fixed_size_uncompressed.cpp
Mike-Dax: The tag is for compile-time 'categorical' overloads, picking between the default 'load compare store each row' `UpdateNumericStats` overload (`FixedSizeGenericContiguousAppendTag`), the 'load, local accumulator, store once' overload (`FixedSizePrimitiveIntegerContiguousAppendTag`) and the floating point overload (`FixedSizeFloatingPointContiguousAppendTag`).

src/storage/compression/fixed_size_uncompressed.cpp
Mike-Dax: https://github.com/duckdb/duckdb/blob/30b7c8ca2c707a4a172dd5e41a2831059c7ec5de/src/include/duckdb/storage/storage_info.hpp#L28-L31

https://github.com/duckdb/duckdb/blob/30b7c8ca2c707a4a172dd5e41a2831059c7ec5de/src/include/duckdb/storage/block_manager.hpp#L143-L146

https://github.com/duckdb/duckdb/blob/30b7c8ca2c707a4a172dd5e41a2831059c7ec5de/src/include/duckdb/common/types/validity_mask.hpp#L172-L175

With a `262,144` byte block size, the header takes up `8` bytes, which gives `262,136` bytes of usable space. Assuming 4 bytes per row:

```
262,136 bytes / 4 bytes per row = 65,534 rows
standard vector 2048 rows * 4 bytes = 8192 bytes

65,534 rows % 2048 rows = 2046
so 31 full 2048 + 1 partial 2046
```

Appending the 32nd 2048 sized vector spills over that boundary

```
entry_idx    = 2046 / 64 = 31
idx_in_entry = 2046 % 64 = 62
```

It will have just 2 active bits in the validty entry for the 2 spilled over rows.

src/storage/compression/fixed_size_uncompressed.cpp
Mike-Dax: The branch is predicated on the active slice:

```cpp
// Limit the validity bitmask slice to the bits left in the current validity word
// and the rows left in the append.
const auto bits_in_entry = MinValue<idx_t>(ValidityMask::BITS_PER_VALUE - idx_in_entry, count - local_offset);
const auto active_mask = ValidityMask::EntryWithValidBits(bits_in_entry);

// Shift the current validity bitmask slice down to bit 0 and mask off rows outside the slice.
const auto valid_bits = (entry >> idx_in_entry) & active_mask;
```

So if `bits_in_entry` is 5:

```
active_mask = 0b11111
// some entry could be loaded, lets say its 0b01010111
entry = validity.GetValidityEntryUnsafe(entry_idx)

// validity mask looks like
bit:       7 6 5 4 3 2 1 0
entry:     0 1 0 1 0 1 1 1
active:          ^ ^ ^ ^ ^
                 4 3 2 1 0

// the shift looks like thi:
valid_bits = entry & active_mask
            = 0b01010111
            & 0b00011111
            = 0b00010111

// and the branches
valid_bits != active_mask   // 0b10111 != 0b11111
valid_bits != 0             // 0b10111 != 0b00000

// so we fall through
```

Regarding the last entry - yeah good idea, splitting it up like:

```
partial head
64-bit sections
partial tail
```

Will probably help, I'll add some benchmarks.

Mytherin:
I've recently reworked how we collect stats - maybe you can rework this into the new `StatsWriter`?

Mike-Dax:
Sure thing, I'm taking a look now.

Mike-Dax:
I didn't have much time as I thought over the past couple ~days~ weeks, here's where I got up to:

I moved the zonemap float / double key encoder and decoder to radix.hpp, they now share the vast majority of their code. Zonemap float / double keys are the same as ART keys now, except for the +inf and -inf representations.
`-0.0` and `+0.0` are canonicalised to `+0.0` now to match ART keys. The behaviour was previously based on the order of insertion within an append. I also found an issue with segments containing only NaNs, I think the behaviours are now consistent:

| Case           | Main                     | PR      |
| -------------- | ------------------------ | ------------------------ |
| `+0.0`         | `+0.0`                   | `+0.0`                   |
| `-0.0`         | `-0.0`                   | `+0.0`                   |
| `-0.0`, `+0.0` | `min = -0.0, max = -0.0` | `min = +0.0, max = +0.0` |
| `+0.0`, `-0.0` | `min = +0.0, max = +0.0` | `min = +0.0, max = +0.0` |
| `NaN` | `min = +inf, max = NaN` | `min = NaN, max = NaN` |
| `NaN`, `NaN` | `min = +inf, max = NaN` | `min = NaN, max = NaN` |

I needed to add 'merge' methods to the StatsWriter, passing a shared one through produced poor assembly and took away some of the performance gains.

> This is interesting but seems somewhat optimized for one scenario (few NULL values) - it's likely this will cause a regression in performance if there are many NULL values as we're doing more stuff per NULL value than before.

> I wonder if we could make this "more adaptive" by e.g. looking at the amount of bits that are set.

> This type of optimized loop also already exists in UnaryExecutor::ExecuteFlat. It would be great if this could be unified so that we can directly use that instead of reimplementing that loop.

> Maybe this optimized loop is not necessary for the last entry, i.e. we run this only as long as the entries are full and have 64 values each? That might also speed up the loops as the compiler knows there's exactly 64 values to process.

I moved the loop into `/src/include/duckdb/common/vector_operations/validity_executor.hpp`.

The `ValidityExecutor` takes a compile-time 'minimum row run length' parameter, and `valid_run`, `invalid_run`, `unknown_validity` functions. It processes the first partial validity word, then full word chunks, then the tail. The 'run' functions are only called with runs of at least the minimum row run length.

Within each validity word slice, it has fast paths for all or no valid rows, otherwise it does run classification. 

The classification has a constant time run detection part (SetBitRunStartMask is a fun function!), then it iterates only over the run starts themselves, dispatching them to `valid_run` or `invalid_run`, otherwise dispatching the interleaving work to the `unknown_validity` function.

I did a benchmark sweep on my M1 and found a min length of 16 to be optimal. I tried joining runs across words, adaptively changing the min run length, but this 'simple' method turned out to be the fastest overall (at least on my laptop).

This shape of API lets the ValidityExecutor handle the run detection, but the caller determine what an optimal run size is, and how to handle mixed runs.

Zonemap stats use it to dispatch in `AppendContiguousNullableFixedSizeValues`.

I couldn't get good codegen without the `always_inline` unfortunately. 

Overall with these changes we get good code generation, with autovectorisation for all of these loops.

I did some initial experimentation with the unary executor and ungrouped aggregate operators using the ValidityExecutor but unfortunately ran out of time.

Here are the benchmark results from my M1, I believe this is an overall improvement. 

| Type                                    | Old time ms | New time ms | Old MiB/s | New MiB/s | Improvement |
| --------------------------------------- | ----------: | ----------: | --------: | --------: | ----------: |
| WIDE_FLOAT_8COLS                        |        4828 |         719 |     331.4 |    2226.5 |       6.72x |
| FLOAT                                   |         661 |         113 |     302.4 |    1764.0 |       5.83x |
| DOUBLE                                  |         698 |         146 |     572.9 |    2741.4 |       4.79x |
| FLOAT_SPECIAL_VALUES                    |         560 |         121 |     356.9 |    1654.8 |       4.64x |
| WIDE_NUMERIC_4COLS                      |        1579 |         382 |     506.5 |    2093.9 |       4.13x |
| FLOAT_NULLS_4_CLUSTERED_PER_64          |         672 |         175 |     297.8 |    1144.3 |       3.84x |
| WIDE_NUMERIC_8COLS                      |        2958 |         785 |     540.9 |    2038.7 |       3.77x |
| FLOAT_NULLS_1_PER_64                    |         672 |         185 |     297.6 |    1078.2 |       3.62x |
| DOUBLE_NULLS_4_CLUSTERED_PER_64         |         677 |         196 |     590.5 |    2045.6 |       3.46x |
| DOUBLE_NULLS_1_PER_64                   |         688 |         204 |     581.8 |    1961.3 |       3.37x |
| FLOAT_NULLS_RANDOM_10                   |         690 |         282 |     289.7 |     710.3 |       2.45x |
| DOUBLE_NULLS_RANDOM_10                  |         677 |         283 |     591.1 |    1412.8 |       2.39x |
| FLOAT_NULLS_32_CLUSTERED_PER_64         |         439 |         187 |     455.7 |    1068.7 |       2.35x |
| UTINYINT                                |         235 |         103 |     212.4 |     486.0 |       2.29x |
| DOUBLE_NULLS_32_CLUSTERED_PER_64        |         469 |         208 |     853.3 |    1927.6 |       2.26x |
| TINYINT                                 |         239 |         106 |     208.9 |     469.5 |       2.25x |
| DECIMAL_4_1                             |         226 |         103 |     883.4 |    1932.9 |       2.19x |
| SMALLINT                                |         234 |         110 |     427.7 |     907.2 |       2.12x |
| USMALLINT                               |         231 |         109 |     432.5 |     915.6 |       2.12x |
| DOUBLE_NULLS_ALTERNATING_50             |         457 |         225 |     875.4 |    1781.5 |       2.04x |
| TINYINT_NULLS_1_PER_64                  |         282 |         162 |     177.6 |     309.3 |       1.74x |
| TINYINT_NULLS_4_CLUSTERED_PER_64        |         277 |         165 |     180.6 |     303.4 |       1.68x |
| UTINYINT_NULLS_4_CLUSTERED_PER_64       |         269 |         162 |     185.7 |     309.2 |       1.67x |
| UTINYINT_NULLS_1_PER_64                 |         271 |         165 |     184.8 |     303.6 |       1.64x |
| DECIMAL_4_1_NULLS_1_PER_64              |         269 |         172 |     742.5 |    1160.6 |       1.56x |
| DECIMAL_4_1_NULLS_4_CLUSTERED_PER_64    |         267 |         173 |     748.6 |    1158.8 |       1.55x |
| USMALLINT_NULLS_4_CLUSTERED_PER_64      |         283 |         186 |     352.9 |     538.5 |       1.53x |
| SMALLINT_NULLS_4_CLUSTERED_PER_64       |         277 |         184 |     360.6 |     544.5 |       1.51x |
| USMALLINT_NULLS_1_PER_64                |         284 |         189 |     352.2 |     528.3 |       1.50x |
| UTINYINT_NULLS_32_CLUSTERED_PER_64      |         252 |         174 |     198.5 |     286.9 |       1.45x |
| SMALLINT_NULLS_1_PER_64                 |         278 |         193 |     359.3 |     517.0 |       1.44x |
| SMALLINT_NULLS_32_CLUSTERED_PER_64      |         240 |         177 |     417.3 |     565.4 |       1.35x |
| FLOAT_NULLS_RANDOM_50                   |         741 |         548 |     270.0 |     364.8 |       1.35x |
| DOUBLE_NULLS_RANDOM_50                  |         740 |         548 |     540.1 |     729.7 |       1.35x |
| DECIMAL_4_1_NULLS_32_CLUSTERED_PER_64   |         229 |         170 |     873.8 |    1177.9 |       1.35x |
| USMALLINT_NULLS_32_CLUSTERED_PER_64     |         243 |         183 |     412.1 |     547.7 |       1.33x |
| TINYINT_NULLS_32_CLUSTERED_PER_64       |         238 |         181 |     210.0 |     276.8 |       1.32x |
| WIDE_INTEGER_8COLS                      |        1148 |         890 |    1394.1 |    1798.4 |       1.29x |
| UBIGINT_NULLS_1_PER_64                  |         259 |         201 |    1544.5 |    1990.9 |       1.29x |
| HUGEINT_VALIDS_1_PER_64                 |         367 |         301 |    2180.8 |    2654.0 |       1.22x |
| UBIGINT                                 |         183 |         151 |    2183.4 |    2650.6 |       1.21x |
| DECIMAL_9_4                             |         149 |         123 |    1341.3 |    1621.6 |       1.21x |
| TINYINT_VALIDS_1_PER_64                 |         251 |         208 |     199.4 |     240.1 |       1.20x |
| DECIMAL_38_10_NULLS_4_CLUSTERED_PER_64  |         643 |         535 |    1244.1 |    1496.6 |       1.20x |
| TINYINT_NULLS_RANDOM_10                 |         327 |         272 |     152.9 |     183.6 |       1.20x |
| FLOAT_NULLS_RANDOM_90                   |         323 |         269 |     620.1 |     744.1 |       1.20x |
| HUGEINT_NULLS_4_CLUSTERED_PER_64        |         674 |         562 |    1186.8 |    1424.1 |       1.20x |
| HUGEINT_NULLS_1_PER_64                  |         687 |         573 |    1164.4 |    1396.1 |       1.20x |
| UINTEGER                                |         151 |         128 |    1322.0 |    1567.9 |       1.19x |
| INTEGER                                 |         150 |         127 |    1331.7 |    1574.1 |       1.18x |
| DECIMAL_38_10_NULLS_1_PER_64            |         655 |         555 |    1222.6 |    1442.7 |       1.18x |
| FLOAT_VALIDS_1_PER_64                   |         250 |         212 |     799.3 |     943.2 |       1.18x |
| UBIGINT_NULLS_32_CLUSTERED_PER_64       |         253 |         215 |    1581.3 |    1862.8 |       1.18x |
| UTINYINT_VALIDS_1_PER_64                |         242 |         205 |     206.8 |     243.4 |       1.18x |
| UHUGEINT                                |         638 |         544 |    1253.6 |    1470.5 |       1.17x |
| DOUBLE_VALIDS_1_PER_1024                |         304 |         260 |    1314.2 |    1540.2 |       1.17x |
| DECIMAL_38_10_VALIDS_1_PER_64           |         351 |         299 |    2280.0 |    2672.2 |       1.17x |
| DECIMAL_38_10_NULLS_RANDOM_10           |         660 |         564 |    1212.1 |    1419.4 |       1.17x |
| INTEGER_NULLS_4_CLUSTERED_PER_64        |         212 |         181 |     943.8 |    1104.2 |       1.17x |
| INTEGER_NULLS_32_CLUSTERED_PER_64       |         219 |         188 |     914.0 |    1064.8 |       1.17x |
| SMALLINT_NULLS_RANDOM_10                |         326 |         281 |     307.0 |     355.5 |       1.16x |
| BIGINT_NULLS_4_CLUSTERED_PER_64         |         239 |         207 |    1672.5 |    1936.7 |       1.16x |
| DECIMAL_18_6_NULLS_1_PER_64             |         221 |         191 |    1807.0 |    2090.7 |       1.16x |
| UINTEGER_NULLS_4_CLUSTERED_PER_64       |         211 |         183 |     948.7 |    1095.8 |       1.16x |
| DECIMAL_38_10                           |         624 |         541 |    1282.5 |    1480.0 |       1.15x |
| HUGEINT                                 |         637 |         553 |    1255.6 |    1446.4 |       1.15x |
| DECIMAL_18_6_NULLS_4_CLUSTERED_PER_64   |         224 |         194 |    1788.4 |    2060.2 |       1.15x |
| HUGEINT_NULLS_RANDOM_10                 |         698 |         606 |    1147.0 |    1320.2 |       1.15x |
| DECIMAL_9_4_NULLS_32_CLUSTERED_PER_64   |         214 |         186 |     932.7 |    1073.5 |       1.15x |
| UINTEGER_NULLS_32_CLUSTERED_PER_64      |         223 |         194 |     897.5 |    1031.2 |       1.15x |
| UTINYINT_NULLS_RANDOM_10                |         314 |         273 |     159.2 |     182.8 |       1.15x |
| DECIMAL_4_1_NULLS_RANDOM_10             |         297 |         259 |     672.7 |     770.9 |       1.15x |
| SMALLINT_VALIDS_1_PER_64                |         237 |         207 |     422.4 |     483.2 |       1.14x |
| INTEGER_NULLS_1_PER_64                  |         210 |         184 |     950.7 |    1087.6 |       1.14x |
| UBIGINT_NULLS_4_CLUSTERED_PER_64        |         235 |         205 |    1704.5 |    1950.0 |       1.14x |
| UHUGEINT_NULLS_1_PER_64                 |         684 |         599 |    1169.6 |    1334.5 |       1.14x |
| BIGINT_NULLS_32_CLUSTERED_PER_64        |         246 |         216 |    1627.5 |    1855.4 |       1.14x |
| UHUGEINT_VALIDS_1_PER_64                |         355 |         312 |    2254.7 |    2568.1 |       1.14x |
| DOUBLE_ALL_NULL                         |         295 |         259 |    1357.5 |    1546.2 |       1.14x |
| UINTEGER_NULLS_1_PER_64                 |         207 |         182 |     965.8 |    1097.1 |       1.14x |
| BIGINT_NULLS_1_PER_64                   |         235 |         207 |    1701.3 |    1931.0 |       1.14x |
| BIGINT_VALIDS_1_PER_64                  |         282 |         249 |    1420.3 |    1609.2 |       1.13x |
| DECIMAL_18_6                            |         168 |         149 |    2378.8 |    2685.7 |       1.13x |
| HUGEINT_NULLS_32_CLUSTERED_PER_64       |         494 |         438 |    1620.3 |    1824.5 |       1.13x |
| TINYINT_NULLS_RANDOM_50                 |         664 |         592 |      75.3 |      84.4 |       1.12x |
| UHUGEINT_NULLS_RANDOM_10                |         683 |         611 |    1171.4 |    1309.6 |       1.12x |
| DECIMAL_9_4_NULLS_1_PER_64              |         199 |         180 |    1003.7 |    1114.1 |       1.11x |
| USMALLINT_NULLS_RANDOM_10               |         328 |         295 |     305.4 |     338.7 |       1.11x |
| DECIMAL_9_4_NULLS_4_CLUSTERED_PER_64    |         197 |         178 |    1015.7 |    1124.4 |       1.11x |
| DECIMAL_4_1_VALIDS_1_PER_64             |         229 |         207 |     873.4 |     966.8 |       1.11x |
| BIGINT_NULLS_RANDOM_50                  |         692 |         627 |     578.6 |     637.6 |       1.10x |
| BIGINT                                  |         175 |         159 |    2281.0 |    2513.7 |       1.10x |
| TINYINT_NULLS_RANDOM_90                 |         309 |         282 |     161.8 |     177.3 |       1.10x |
| INTEGER_LONG_RUNS                       |         207 |         189 |     964.1 |    1056.6 |       1.10x |
| UTINYINT_NULLS_RANDOM_50                |         654 |         597 |      76.5 |      83.8 |       1.09x |
| INTEGER_NULLS_RANDOM_50                 |         660 |         603 |     303.2 |     331.7 |       1.09x |
| UHUGEINT_NULLS_4_CLUSTERED_PER_64       |         662 |         605 |    1208.5 |    1322.1 |       1.09x |
| DECIMAL_38_10_NULLS_32_CLUSTERED_PER_64 |         470 |         431 |    1701.8 |    1856.7 |       1.09x |
| DOUBLE_NULLS_RANDOM_90                  |         330 |         303 |    1211.5 |    1321.8 |       1.09x |
| DECIMAL_9_4_NULLS_RANDOM_50             |         584 |         537 |     342.1 |     372.2 |       1.09x |
| DECIMAL_18_6_NULLS_RANDOM_50            |         593 |         546 |     674.1 |     732.7 |       1.09x |
| SMALLINT_NULLS_RANDOM_50                |         640 |         590 |     156.2 |     169.6 |       1.09x |
| DECIMAL_4_1_NULLS_RANDOM_50             |         562 |         521 |     355.8 |     383.5 |       1.08x |
| INTEGER_VALIDS_1_PER_64                 |         246 |         229 |     812.5 |     872.6 |       1.07x |
| UINTEGER_NULLS_RANDOM_50                |         658 |         613 |     303.7 |     326.2 |       1.07x |
| USMALLINT_VALIDS_1_PER_64               |         227 |         211 |     441.4 |     473.6 |       1.07x |
| DECIMAL_18_6_NULLS_32_CLUSTERED_PER_64  |         228 |         212 |    1756.8 |    1885.0 |       1.07x |
| UBIGINT_NULLS_RANDOM_50                 |         667 |         623 |     599.7 |     642.3 |       1.07x |
| UINTEGER_VALIDS_1_PER_64                |         242 |         226 |     825.9 |     884.5 |       1.07x |
| INTEGER_VALIDS_1_PER_1024               |         274 |         256 |     731.0 |     782.2 |       1.07x |
| INTEGER_NULLS_ALTERNATING_50            |         231 |         218 |     865.0 |     916.9 |       1.06x |
| DECIMAL_18_6_VALIDS_1_PER_64            |         264 |         250 |    1515.9 |    1600.8 |       1.06x |
| INTEGER_ALL_NULL                        |         259 |         246 |     770.9 |     811.8 |       1.05x |
| DECIMAL_38_10_NULLS_RANDOM_50           |         698 |         663 |    1146.1 |    1206.8 |       1.05x |
| UBIGINT_VALIDS_1_PER_64                 |         266 |         253 |    1503.7 |    1583.4 |       1.05x |
| SMALLINT_NULLS_RANDOM_90                |         300 |         286 |     332.8 |     349.8 |       1.05x |
| USMALLINT_NULLS_RANDOM_50               |         643 |         613 |     155.5 |     163.1 |       1.05x |
| BIGINT_NULLS_RANDOM_90                  |         333 |         319 |    1200.1 |    1255.3 |       1.05x |
| DOUBLE_VALIDS_1_PER_64                  |         265 |         254 |    1507.7 |    1577.1 |       1.05x |
| UHUGEINT_NULLS_RANDOM_50                |         761 |         727 |    1051.6 |    1100.0 |       1.05x |
| UTINYINT_NULLS_RANDOM_90                |         306 |         293 |     163.3 |     170.6 |       1.04x |
| DECIMAL_18_6_NULLS_RANDOM_90            |         311 |         298 |    1285.8 |    1341.1 |       1.04x |
| DECIMAL_9_4_VALIDS_1_PER_64             |         236 |         228 |     847.9 |     876.7 |       1.03x |
| UHUGEINT_NULLS_32_CLUSTERED_PER_64      |         466 |         451 |    1715.8 |    1772.4 |       1.03x |
| DECIMAL_38_10_NULLS_RANDOM_90           |         368 |         356 |    2172.9 |    2244.6 |       1.03x |
| HUGEINT_NULLS_RANDOM_90                 |         386 |         374 |    2072.6 |    2136.8 |       1.03x |
| INTEGER_NULLS_RANDOM_90                 |         300 |         292 |     665.5 |     686.1 |       1.03x |
| HUGEINT_NULLS_RANDOM_50                 |         772 |         749 |    1037.0 |    1068.1 |       1.03x |
| UINTEGER_NULLS_RANDOM_90                |         301 |         294 |     663.3 |     680.5 |       1.03x |
| DECIMAL_18_6_NULLS_RANDOM_10            |         293 |         287 |    1367.0 |    1395.7 |       1.02x |
| BIGINT_NULLS_RANDOM_10                  |         313 |         311 |    1277.7 |    1287.9 |       1.01x |
| INTEGER_NULLS_RANDOM_10                 |         293 |         290 |     683.7 |     688.5 |       1.01x |
| DECIMAL_4_1_NULLS_RANDOM_90             |         273 |         272 |     732.3 |     736.0 |       1.00x |
| DECIMAL_9_4_NULLS_RANDOM_90             |         278 |         277 |     720.9 |     723.1 |       1.00x |
| UBIGINT_NULLS_RANDOM_10                 |         313 |         313 |    1275.8 |    1277.1 |       1.00x |
| UBIGINT_NULLS_RANDOM_90                 |         316 |         316 |    1265.4 |    1265.4 |       1.00x |
| UHUGEINT_NULLS_RANDOM_90                |         386 |         387 |    2070.1 |    2066.0 |       1.00x |
| UINTEGER_NULLS_RANDOM_10                |         288 |         289 |     695.1 |     690.9 |       0.99x |
| USMALLINT_NULLS_RANDOM_90               |         283 |         286 |     352.8 |     350.0 |       0.99x |
| DECIMAL_9_4_NULLS_RANDOM_10             |         264 |         269 |     756.8 |     742.4 |       0.98x |

src/storage/compression/fixed_size_uncompressed.cpp
Mike-Dax: I've removed the tags, using constexprs instead to make it clearer.

Mike-Dax:
I've rebased this on main. Is the scope of this acceptable? I'm wary of extending this into the unary / other executors all in one PR.

Mytherin:
Thanks - the scope is indeed too large, and we really don't need 30 benchmark files. Let's try 1 benchmark file and a smaller scope to start out with.

PR #22472, opened 2026-05-05
Remove unused ccache and fix julia CI

smvv:
- Remove unused `ccache` action from job `prepare`.
- Add [missing](https://github.com/duckdb/duckdb/actions/runs/25352299682/job/74335946735) `DUCKDB_TYPE_TIMESTAMP_TZ_NS` to julia.
- Run julia CI jobs for PR when C types can change (to prevent future breakage).

### DIFF
--- a/.github/workflows/Main.yml
+++ b/.github/workflows/Main.yml
@@ -261,10 +261,6 @@ jobs:
             echo "static_libs_linux_matrix=$static_libs_linux_matrix"
           } >> "$GITHUB_OUTPUT"
 
-      - uses: ./.github/actions/ccache-action
-        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/Main.yml
```diff
@@ -122,6 +122,11 @@ jobs:
             tests_slow:
               - test/**/*.cpp
               - test/**/*.test_slow
+            capi:
```
smvv, line 125: note: perhaps there are more files to list here, but this seems to be "good enough" to test julia breakage for now.

lnkuiper:
Thanks! Good idea to rerun julia on capi changes :)

PR #22475, opened 2026-05-05
Fix enable_logging() silently resetting logging_storage

guillesd:
When doing:

```sql
SET logging_storage = stdout;
CALL enable_logging('QueryLog');
```

The setting was silently overriden by the `enable_logging` function, even if the args for `storage_path` and/or `storage`  were not set.

### DIFF
--- a/src/function/table/system/logging_utils.cpp
+++ b/src/function/table/system/logging_utils.cpp
@@ -87,6 +87,12 @@ static unique_ptr<FunctionData> BindEnableLogging(ClientContext &context, TableF
 		result->config.storage = LogConfig::FILE_STORAGE_NAME;
 	}
 
+	// If the user did not specify a storage, preserve the currently configured one. Otherwise the
+	// default LogConfig would silently reset logging_storage back to 'memory'.
+	if (!storage_isset && !storage_path_isset) {
+		result->config.storage = context.db->GetLogManager().GetConfig().storage;
+	}
+
 	// Process positional params
 	if (!input.inputs.empty()) {
 		if (input.inputs[0].type() == LogicalType::VARCHAR) {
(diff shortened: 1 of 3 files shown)


carlopi:
The current code is somewhat twisted (I did found a prexisting but), but I think I found also a corner case in this PR:
```sql
CALL enable_logging('QueryLog', storage_config:={path:'file.csv'});
```
that before the PR fails to validated, while now it does enable `QueryLog`, but it silently skips storage_config (either validating it or passing it down).

Same for:
```
CALL enable_logging('QueryLog',  storage_config:={aaaargh:'blaargh'})
```
that now just passes.

guillesd:
@carlopi checking

guillesd:
btw this thing @carlopi :

```
CALL enable_logging('QueryLog',  storage_config:={aaaargh:'blaargh'})
```
Happened because this goes through the `ShellLogStorage` which doesn't have any config so it basically inherits `UpdateConfig` from the base class. We could allso fix this in `UpdateConfig` in the `ShellLogStorage` implementation but I think both are fine.

lnkuiper:
Looks good! Just have one question: With this PR, `storage_config` without `storage=` is applied to the currently configured storage, except when it contains path, which forces file storage. Is that the intended behavior? For example, should `CALL enable_logging(storage_config:={normalize:false})` configure the current
storage, or should `storage_config` without an explicit `storage=` be rejected except for the path shorthand case?

guillesd:
@lnkuiper I've thought about this quite a bit and I think this is a better alternative. If you have a setting specified, let's say `SET logging_storage = 'file'` and then some time later in the same process/session I decide that I want to configure a different logger via `CALL enable_logging`, then I should be explicit about it.

In general, without breaking the API, there is no perfect solution in my opinion. What I tried to do with this PR is always keep this idea that `logging is set > enable_logging without storage does not reset`; `logging is set, enable_logging with storage resets`. Users just need to be explicit about what they want to do, which I think is fine.

lnkuiper:
Thanks for clarifying! Then I think this is good to go. Should be documented so I'll add the label

PR #22266, opened 2026-04-24
[MultiFileReader][Parquet] Support pushdown of `struct_extract` expression to the scan

Tishj:
In #14839 we implemented struct extract pushdown as still emitting a `STRUCT(a, b, c)` but leaving `a` and `c` as NULL if only `b` was referenced.

Back in https://github.com/duckdb/duckdb/pull/20063 we improved on that strategy, for our own internal table scan method (details are in the PR)

With this PR, we implement that same improved strategy for the `parquet_scan` method, and laying the ground work for `MultiFileReader` support for this in general.

### Summary of changes:
- add `supports_pushdown_extract` and `statistics_extended` implementations
- propagate the `ColumnIndex` further down through the `MultiFileReader` instead of bastardizing it down to a `column_t`
- detect `IsPushdownExtract` in the `StructColumnReader` and inject an `ExpressionColumnReader` if a cast is pushed down on top of the `struct_extract`.

PR #22478, opened 2026-05-05
[MultiFileReader][Parquet] Add `variant_extract` pushdown

Tishj:
This PR is a follow up to #22266 

### Summary of Changes
- `ColumnReader::Stats` is now virtual
- `VariantColumnReader::Stats` applies the `ColumnIndex index` (pushdown) to the returned stats.
- `ExpressionColumnReader::Select` makes sure to not scan child rows that are filtered out
- `BaseFileReader::expression_map` is no longer indexed by 'column_id', instead using the index into 'column_indexes' to key it.

When creating the ColumnReader for a VARIANT, we've added a new path that gets taken if `column_index.IsPushdownExtract()` is true.

If `IsFullyShredded` (the stats have proven that all rows are null or non-null and of the shredded type), then 
we'll perform a pushed down `struct_extract` into the internals of the VARIANT structure, directly emitting the leaf `typed_value`, optionally wrapping it in an `ExpressionColumnReader` if a CAST was also pushed down.

Otherwise we'll create the VARIANT ColumnReader as normal, adding the `column_index` to the reader, to be used in `Stats` and `Read` calls.


extension/parquet/include/reader/variant_column_reader.hpp
```diff
@@ -21,8 +21,11 @@ class VariantColumnReader : public ColumnReader {
 public:
 	VariantColumnReader(ClientContext &context, const ParquetReader &reader, const ParquetColumnSchema &schema,
 	                    vector<unique_ptr<ColumnReader>> child_readers_p);
+	VariantColumnReader(ClientContext &context, const ParquetReader &reader, const ParquetColumnSchema &schema,
+	                    vector<unique_ptr<ColumnReader>> child_readers_p, const struct ColumnIndex &index);
 
 	ClientContext &context;
+	struct ColumnIndex index;
```
Tishj: For extracts we couldn't push down (i.e the VARIANT isn't fully shredded), we save the index so we can perform the extract in `Read`

extension/parquet/reader/variant_column_reader.cpp
```diff
@@ -80,6 +88,32 @@ void VariantColumnReader::InitializeRead(idx_t row_group_idx_p, const vector<Col
...
+unique_ptr<BaseStatistics> VariantColumnReader::Stats(idx_t row_group_idx_p, const vector<ColumnChunk> &columns) {
+	auto result = ColumnReader::Stats(row_group_idx_p, columns);
+	if (result && index.IsPushdownExtract()) {
+		auto storage_index = StorageIndex::FromColumnIndex(index);
+		return result->PushdownExtract(storage_index.GetChildIndexes()[0]);
+	}
+	return result;
+}
+
+static vector<VariantPathComponent> GetVariantExtractPath(const ColumnIndex &index) {
```
Tishj: nit: we could cache this, compute it once instead of every Read call

extension/parquet/parquet_reader.cpp
```diff
@@ -1156,66 +1321,6 @@ const FileMetaData *ParquetReader::GetFileMetadata() const {
 	return metadata->metadata.get();
 }
 
-static unique_ptr<BaseStatistics> ReadStatisticsInternal(const FileMetaData &file_meta_data,
```
Tishj, line 1170: This is just moved, as `ReadColumnStatistics` is now used by `CreateReaderRecursive`

src/execution/expression_executor.cpp
```diff
@@ -99,6 +99,15 @@ void ExpressionExecutor::ExecuteExpression(DataChunk &input, Vector &result) {
 	ExecuteExpression(result);
 }
 
+void ExpressionExecutor::ExecuteExpression(DataChunk &input, Vector &result, const SelectionVector &sel, idx_t count) {
```
Tishj, line 102: This is related but orthogonal, maybe we should try to move this out to a different PR
(together with the ExpressionColumnReader changes)

extension/parquet/parquet_reader.cpp
```diff
@@ -592,6 +592,138 @@ ParquetColumnSchema ParquetReader::ParseColumnSchema(const SchemaElement &s_ele,
...
+		if (!chunk_stats) {
+			return nullptr;
+		}
+		if (!column_stats) {
+			column_stats = std::move(chunk_stats);
+		} else {
+			column_stats->Merge(*chunk_stats);
+		}
+	}
+	return column_stats;
```
pdet: This looks very similar to the code from `ReadColumnStatistics`. It seems the only difference is it first indexes `root_schema.children[file_col_idx]`.  Could it just be `return ReadColumnStatistics(file_meta_data, root_schema.children[file_col_idx], parquet_options);` ?

extension/parquet/parquet_reader.cpp
```diff
@@ -592,6 +592,138 @@ ParquetColumnSchema ParquetReader::ParseColumnSchema(const SchemaElement &s_ele,
...
+		auto &child = path_iter.get().GetChildIndexes()[0];
+		auto &field_name = child.GetFieldName();
+		auto child_column_index = typed_value.get().GetChildIndexByName(field_name);
+		if (!child_column_index.IsValid()) {
+			throw InternalException("Can't locate the child by name '%s' in the VARIANT column", field_name);
+		}
+		auto &child_column = typed_value.get().GetChildByIndex(child_column_index.GetIndex());
+		if (child_column.type.id() != LogicalTypeId::STRUCT) {
+			throw InternalException("Extracted field for '%s' from 'typed_value', is not a struct (received: %s)",
+			                        child_column.type.ToString());
```
pdet: Are we missing an argument for a placeholder?

extension/parquet/parquet_reader.cpp
```diff
@@ -662,11 +794,44 @@ unique_ptr<ColumnReader> ParquetReader::CreateReaderRecursive(ClientContext &con
...
 		children.resize(schema.children.size());
-		for (idx_t child_index = 0; child_index < schema.children.size(); child_index++) {
+		if (schema.children.size() != 3 || !column_id.IsPushdownExtract()) {
+			for (idx_t child_index = 0; child_index < schema.children.size(); child_index++) {
+				children[child_index] =
+				    CreateReaderRecursive(context, ColumnIndex(child_index), schema.children[child_index]);
+			}
+			return make_uniq<VariantColumnReader>(context, *this, schema, std::move(children));
+		}
+		//! VARIANT is shredded -  it has a 'typed_value' column
```
pdet, line 813: nit `//!` in cpp

extension/parquet/parquet_reader.cpp
```diff
@@ -662,11 +794,44 @@ unique_ptr<ColumnReader> ParquetReader::CreateReaderRecursive(ClientContext &con
...
+		//! And the extract is pushed down into the scan
+		auto &typed_value_schema = schema.children[2];
+		D_ASSERT(typed_value_schema.name == "typed_value");
+		auto variant_stats = ReadColumnStatistics(*GetFileMetadata(), schema, parquet_options);
```
pdet: Is this recomputing it for each use? Does it make sense to check for reuse for the same shredded VARIANT column?

src/execution/expression_executor.cpp
```diff
@@ -99,6 +99,24 @@ void ExpressionExecutor::ExecuteExpression(DataChunk &input, Vector &result) {
...
+	if (count == 0) {
+		return;
+	}
+	if (count == input.size()) {
+		Execute(*expression, state->root_state.get(), nullptr, count, result);
+		return;
+	}
+	Vector intermediate_result(expression->GetReturnType());
+	Execute(*expression, state->root_state.get(), &sel, count, intermediate_result);
+	FillSwitch(intermediate_result, result, sel, NumericCast<sel_t>(count));
```
pdet: Can we test:
```sql  
COPY (FROM (VALUES
({'arr': [1::INTEGER,2,3], 'tag': 'a'}::VARIANT),
({'arr': [4::INTEGER,5,6], 'tag': 'b'}::VARIANT),
({'arr': [7::INTEGER,8,9], 'tag': 'c'}::VARIANT)) t(v))
TO 'f.parquet' (SHREDDING {'v': 'STRUCT(arr INTEGER[], tag VARCHAR)'});

SELECT v.arr::INTEGER[3] FROM 'f.parquet' WHERE v.tag = 'b';
```

extension/parquet/parquet_column_schema.cpp
```diff
@@ -12,6 +12,24 @@
 
 namespace duckdb {
 
+optional_idx ParquetColumnSchema::GetChildIndexByName(const string &name) const {
+	for (idx_t i = 0; i < children.size(); i++) {
+		auto &child = children[i];
+		if (StringUtil::CIEquals(name, child.name)) {
+			return i;
```
pdet:
```sql
COPY (SELECT {'Foo': i::INTEGER}::VARIANT AS v FROM range(10) t(i)) TO 'foo.parquet' (SHREDDING {'v': 'STRUCT(Foo INTEGER)'});
  SELECT count(*) FROM 'foo.parquet' WHERE v.foo = 3;                       
  SELECT count(*) FROM (VALUES ({'Foo':3::INTEGER}::VARIANT)) t(v) WHERE v.foo = 3;
```

It seems that the String checks differ from pushdown or not?

pdet (commented):
Thanks for the PR. Just had a few comments.

extension/parquet/parquet_reader.cpp
Tishj: Wasn't aware we had this rule, I thought all comments use `//!` ?
But sure, I'll remove it

extension/parquet/parquet_reader.cpp
Tishj: Maybe we could cache this on the ParquetReader, as a `mutable` (since the method is const)
Might need a lock as well then

src/execution/expression_executor.cpp
Tishj: Hmmmm, I'll try reverting to the old method I had, I hoped this would be cleaner

Tishj: Yea, it works correctly with `ReverseSelectionVector`
It's just uglier imo
But to fix that uglyness requires a much bigger structural change to parquet reading in the presence of filters

extension/parquet/parquet_column_schema.cpp
Tishj: Good call, ~~hard to make a test for, I'm trying with JSON now~~ (EDIT: didn't need JSON)

Tishj: Hmm, I think there might be a bug in `ChildShreddingTypes`, because that should not be an `identifier_map_t`
We *could* still leave it as `identifier_map_t`, just not use `find` on it, but that communicates the wrong idea I think

Mytherin:
Thanks!

extension/parquet/include/parquet_reader.hpp
```diff
@@ -464,6 +467,8 @@ class ParquetReader : public BaseFileReader {
 	mutable unique_ptr<CachingFileHandle> prewarmed_scan_handle;
 	//! Scan handle shared by all scan states of this reader
 	mutable shared_ptr<CachingFileHandle> shared_scan_handle;
+	mutable annotated_mutex variant_stats_lock;
+	mutable unordered_map<idx_t, unique_ptr<BaseStatistics>> variant_stats_cache;
```
Mytherin, line 472: After reading through the code a bit more I'm a bit worried about the life-time of the variant stats cache. `GetVariantStats` returns an `optional_ptr` to a `unique_ptr` that's obtained under the lock, but `variant_stats_cache` is not strictly additive (there's a `.clear()` in `ParquetReader::InitializeScan`).

Maybe this should be a "per row-group, additive" map instead? Or this should not live in the `ParquetReader` but be moved into the scan states?

src/include/duckdb/common/multi_file/base_file_reader.hpp
```diff
@@ -59,7 +59,7 @@ class BaseFileReader : public enable_shared_from_this<BaseFileReader> {
 	vector<ColumnIndex> column_indexes;
 	//! The set of table filters (adjusted to local indexes)
 	unique_ptr<TableFilterSet> filters;
-	//! Expression to execute for a given column (BEFORE executing the filter)
+	//! Expression to execute for a local scan-column position (BEFORE executing the filter)
 	//! NOTE: this is only set when we have filters - it can be ignored for readers that don't have filter pushdown
 	unordered_map<column_t, BaseFileReaderExpression> expression_map;
```
Mytherin: This is a pretty subtle change that might be hard to detect / understand / fix for out-of-tree extensions. Can we turn this into `unordered_map<ProjectionIndex, ...>` instead so the change is more explicit (and causes compilation errors if not addressed)? I think that's also what's intended here.

Mytherin (commented):
One more comment

extension/parquet/parquet_reader.cpp
```diff
@@ -1486,6 +1600,10 @@ void ParquetReader::InitializeScan(ClientContext &context, ParquetReaderScanStat
 	state.filter_count = 0;
 	state.group_index = group_to_read;
 	state.sel.Initialize(STANDARD_VECTOR_SIZE);
+	{
+		const annotated_lock_guard<annotated_mutex> guard(variant_stats_lock);
+		variant_stats_cache.clear();
```
Mytherin: This probably shouldn't be getting cleared since we will be re-computing per-file stats unnecessarily

Mytherin:
Thanks!

PR #22480, opened 2026-05-05
Explicitly set an extension schema

ccfelius:
This PR allows to create, set and add an explicit schema where extension functions are registered. The extension itself has explicit control over this.

An explicit schema can be (1) created, (2) set and (3) added to the search path by using the following functions from the `ExtensionLoader` class:

```
CreateSchema(const string &extension_schema_name)
UseDefaultSchema(const string &name = DEFAULT_SCHEMA) 
AddSchemaToSearchPath(const string &schema_name)
```

Or you can do all in one by

```
UseDedicatedSchemaForExtension(const string &extension_schema_name);
```

Or, you can use a duckdb suggested name (the extension name or alias) by using the an overload

```
UseDedicatedSchemaForExtension();
```

Then, if the user loads the extension using an alias all functions will automatically be registered in a schema that corresponds to this alias. For example:

```
LOAD extension as ext
ext.func()
```

The explicit schema is tracked by the`ExtensionCallbackManager`.  When an extension is finished loading (in `FinalizeLoad`) the catalog search path is updated to include the extension schema. But I've also added a seperate function to do this: `RefreshSearchPath(ClientContext &context)`.

This PR allows to call functions by both

```
SELECT explicit_schema.dedicated_schema_function();
SELECT dedicated_schema_function();
```

part of https://github.com/duckdblabs/duckdb-internal/issues/4838


ccfelius:
@Maxxen

src/main/extension/extension_loader.cpp
```diff
@@ -119,13 +146,15 @@ void ExtensionLoader::RegisterFunction(TableFunction function) {
 void ExtensionLoader::RegisterFunction(TableFunctionSet function) {
 	D_ASSERT(!function.name.empty());
 	CreateTableFunctionInfo info(std::move(function));
+	info.schema = extension_schema;
```
Maxxen: I guess we only want to override the schema if it's not set explicitly by the user right?

src/main/database_manager.cpp
```diff
@@ -460,4 +460,14 @@ Catalog &DatabaseManager::GetSystemCatalog() {
 	return system->GetCatalog();
 }
 
+void DatabaseManager::AddExtensionSchema(const string &schema) {
+	lock_guard<mutex> lock(schemas_lock);
```
Maxxen: So im not sure about the locking here... maybe it's unavoidable, but perhaps the schemas could be set as entries in the `ExtensionCallbackManager` instead? Then all locking/versioning would be taken care of automatically I would think. Not sure how that interfaces with the current search path code though.

src/catalog/catalog_search_path.cpp
```diff
@@ -139,6 +139,10 @@ void CatalogSearchPath::Reset() {
 	SetPathsInternal(empty);
 }
 
+void CatalogSearchPath::RefreshSetPaths() {
```
Maxxen, line 143: Is this ever called?

src/main/database_manager.cpp
ccfelius: I'll look at doing this different

src/catalog/catalog_search_path.cpp
ccfelius: yes

```
ClientData::Get(context.client).catalog_search_path->RefreshSetPaths();
```

src/main/extension/extension_loader.cpp
ccfelius: `extension_schema` is initialized with `DEFAULT_SCHEMA`, so if it's not explicitly set it should be an idempotent operation, at least I thought so? Or am I misunderstanding something?

src/main/extension/extension_loader.cpp
```diff
@@ -35,13 +38,36 @@ void ExtensionLoader::SetDescription(const string &description) {
...
 }
 
+void ExtensionLoader::CreateExtensionSchema() const {
+	auto &system_catalog = Catalog::GetSystemCatalog(db);
+	auto data = CatalogTransaction::GetSystemTransaction(db);
+
+	CreateSchemaInfo info;
+	info.schema = extension_schema;
+	info.internal = true;
+	info.on_conflict = OnCreateConflict::IGNORE_ON_CONFLICT;
```
Maxxen: What happens here if we ignore the creation on conflict, and then continue to register e.g. functions in the non-existing schema?

Maxxen:
Thinking about this some more... maybe its better if we don't create and register the items in the schema automatically, but instead just provide a way for the extension to:

- Get its "assigned name" from the loader (e.g. default name, or alias), 
- Create a schema
- Add a (arbitrary) schema to the search path

A well-behaved extension would then do all three and also make sure to explicitly set the schema of all of the items it registers to the one provided by duckdb. But it also gives the extension the option to disregard all of that and/or register items in multiple schemas and/or the default schema.

ccfelius:
> Thinking about this some more... maybe its better if we don't create and register the items in the schema automatically, but instead just provide a way for the extension to:
> 
> * Get its "assigned name" from the loader (e.g. default name, or alias),
> * Create a schema
> * Add a (arbitrary) schema to the search path
> 
> A well-behaved extension would then do all three and also make sure to explicitly set the schema of all of the items it registers to the one provided by duckdb. But it also gives the extension the option to disregard all of that and/or register items in multiple schemas and/or the default schema.

So basically, just split the current PR in three separate functions that can be called through the ExtensionLoader?

src/main/extension/extension_loader.cpp
ccfelius: I'm not sure what you mean by this, can you elaborate? If you would ignore the creation on conflict, functions will just error out if you try to register the same function within the same extension (`ERROR_ON_CONFLIT` will then be the default iirc). This should never happen though?

Maxxen:
Looks good! Just some nits about the naming (but could also be a separate PR).

I think maybe we should drop the "Extension" part of the names of these functions, essentially they have nothing to do with extensions, but just schemas in general right?

Instead of:
```
CreateExtensionSchema(const string &extension_schema_name)
SetExtensionSchema(const string &name)
AddExtensionSchemaToSearchPath(const string &schema_name)

RegisterExtensionInSeparateSchema(const string &extension_schema_name);
```

How about:
```
CreateSchema(const string &extension_schema_name)
UseDefaultSchema(const string &name = DEFAULT_SCHEMA) // reset to default if not supplied
AddSchemaToSearchPath(const string &schema_name)

UseDedicatedSchemaForExtension(const string &extension_schema_name, bool add_to_search_path = true));
```

In the future we can then make an overload for `UseDedicatedSchemaForExtension()` that doesn't take a name if we want to use the automatic DuckDB/user-supplied name instead.

ccfelius:
```
Looks good! Just some nits about the naming (but could also be a separate PR).
```

You're right, I'll change the naming now, it's a small effort. Thanks for having a look!

Maxxen:
Nice!

Maxxen (approved):

Mytherin:
Thanks!

PR #22483, opened 2026-05-05
Temporary disable Windows ARM64 on CI

staticlibs:
There was a report (in Discord) about build failures on `windows_arm64` due to the roll out of the Visual Studio 2026 on the `windows-11-arm` image.

This PR temporary removes the `windows_arm64` from `opt_in_archs` to allow builds to pass while the problem is being investigated.


staticlibs:
Seems to be a false alarm, closing for now.

PR #22063, opened 2026-04-14
Revamp extension build / patching system for better local development, enabled when `DUCKDB_NEW_EXTENSION_BUILD=1` is defined

Mytherin:
This PR creates an initial version of a better extension build / patching system for local development, as the current system is a bit of a mess.

* We need to have `USE_MERGED_VCPKG_MANIFEST=1 BUILD_ALL_EXT=1 make extension_configuration`  to generate a merged vcpkg manifest list. This clones all extension repositories in `build/extension_configuration/`
* We need to then run `USE_MERGED_VCPKG_MANIFEST=1 BUILD_ALL_EXT=1 make reldebug`, which clones the extension repositories again in `build/reldebug/_deps/` and applies the patches. Patches are left as uncommitted / unstaged in the repositories.
* If we then want to actually change these repositories in order to fix builds or generate new patches, we need to jump through a bunch of hoops to do so - the dev experience is very poor. In addition, this system gravitates towards generating a single giant patch which leads to terrible merge conflicts. 

### New Build System

This PR introduces a new build system (that ideally eventually replaces the old one) that addresses these issues:

* Extensions are cloned once in `extension/external/`, so if we build delta this will end up in `extension/external/delta`
* `USE_MERGED_VCPKG_MANIFEST=1` is obsolete. If we are building any extensions that require vcpkg we will automatically generate a `vcpkg.json` file in `build/vcpkg.json` containing the merged vcpkg manifests. 
* Patches are now applied as commits in the repository with as commit message the patch file name, e.g. `0001_fix1.patch`, `0002_fix2.patch`. This allows subsequent development / fixes to be build on top of previous patches.

In order to use the new build system you have to use `DUCKDB_NEW_EXTENSION_BUILD=1`. I've placed it in `~/.zshrc`:

```
export DUCKDB_NEW_EXTENSION_BUILD=1
```

### Development

If we perform development in the repo (e.g. make changes to `extension/external/....`) - we have a few new workflows:

If we have uncommitted changes this error will be thrown:

```
ERROR: Extension 'delta' has uncommitted changes:
 M src/delta_extension.cpp
To resolve:
* FORCE_APPLY_PATCHES=1           — discard local changes and re-apply patches
* DUCKDB_SKIP_APPLYING_PATCHES=1  — skip patch checks and compile as-is
make: *** [sync_out_of_tree_extensions] Error 1
```

[...]

PR #22489, opened 2026-05-06
Internal #9096: Strong Timestamp Typing 

hawkfish:
* Redefine timestamp_t and friends as independent templated types based on timebase_t;
* Add IsFinite member to the template;
* Replace arithmetic members with appropriate Operator calls;
* Fix inconsistent downcast rounding;


hawkfish:
Not sure how to fix the failing extension builds?

Mytherin:
You need to generate patches - see https://github.com/duckdb/duckdb/pull/22063 where I detail how I'm currently doing this

hawkfish:
> You need to generate patches - see #22063 where I detail how I'm currently doing this

Got it - sounds like what I had to do for the fuzzer.

hawkfish:
The only remaining failure appears to be Windows being unable to find Ninja... I believe this has passed in some of the other runs and the only thing that failed on the last one was a test output changed slightly for ducklake.

Mytherin:
Thanks!

PR #20477, opened 2026-01-10
Add `invoke` function

Maxxen:
This PR just strips out the `lambda_invoke` function from PR https://github.com/duckdb/duckdb/pull/20379 and renames it to just `invoke`. Even if we don't decide on special LET binding syntax I think it makes sense to try to get the base functionality in anyway as it can still be useful for users and opens up a lot of other interesting work.

PR #22490, opened 2026-05-06
Deprecate `BoundBetweenExpression` and implement between as a `BoundFunctionExpression`

Mytherin:
This PR deprecates the `BoundBetweenExpression` and moves it to the `LegacyBoundBetweenExpression` class. Instead, between is now implemented as a function. This allows between to use the generic function execution infrastructure instead of having its own execution pipeline.

`BoundBetweenExpression` remains as a helper struct:

```cpp
struct BoundBetweenExpression {
	static unique_ptr<Expression> Create(unique_ptr<Expression> input, unique_ptr<Expression> lower,
	                                     unique_ptr<Expression> upper, bool lower_inclusive, bool upper_inclusive);

	static bool LowerInclusive(const BoundFunctionExpression &between_expr);
	static bool UpperInclusive(const BoundFunctionExpression &between_expr);

	static const Expression &Input(const BoundFunctionExpression &between_expr);
	static const Expression &LowerBound(const BoundFunctionExpression &between_expr);
	static const Expression &UpperBound(const BoundFunctionExpression &between_expr);

	static unique_ptr<Expression> &InputMutable(BoundFunctionExpression &between_expr);
	static unique_ptr<Expression> &LowerBoundMutable(BoundFunctionExpression &between_expr);
	static unique_ptr<Expression> &UpperBoundMutable(BoundFunctionExpression &between_expr);

	static ExpressionType LowerComparisonType(const BoundFunctionExpression &between_expr) {
		return LowerInclusive(between_expr) ? ExpressionType::COMPARE_GREATERTHANOREQUALTO
		                                    : ExpressionType::COMPARE_GREATERTHAN;
	}
	static ExpressionType UpperComparisonType(const BoundFunctionExpression &between_expr) {
		return UpperInclusive(between_expr) ? ExpressionType::COMPARE_LESSTHANOREQUALTO
		                                    : ExpressionType::COMPARE_LESSTHAN;
	}
};
```

In order to make this possible, three new callbacks are added to the scalar function:

```cpp
//! Convert a scalar function to string
typedef string (*function_to_string_t)(FunctionToStringInput &input);

//! Get the expression type of a function
typedef ExpressionType (*function_get_expression_type_t)(FunctionToStringInput &input);

//! Legacy serialize function for expressions that were converted into functions
typedef unique_ptr<Expression> (*function_legacy_serialize_t)(FunctionToStringInput &input);
```

* `function_to_string_t` converts between to the SQL standard syntax: `x BETWEEN y AND z`
* `function_get_expression_type_t` returns a different expression type for the function (`ExpressionType::COMPARE_BETWEEN`). The expression class remains `ExpressionClass::BOUND_FUNCTION`. This allows the optimizer to keep on using the old expression type to detect `BETWEEN` constructs.
* `function_legacy_serialize_t` constructs an alternative expression to serialize if we are targeting serialization versions `< 2.0` - in this case the `LegacyBoundBetweenExpression`


src/include/duckdb/function/scalar/comparison_functions.hpp
```diff
@@ -0,0 +1,28 @@
...
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/function/function_set.hpp"
+
+namespace duckdb {
+
+struct BetweenFun {
+	static constexpr const char *Name = "__between";
```
Maxxen, line 19: Does this really need to have a secret name instead of just "between"?

src/function/scalar/comparison/between.cpp
```diff
@@ -0,0 +1,247 @@
...
+		return make_uniq<BetweenFunctionData>(lower_inclusive, upper_inclusive);
+	};
+
+	bool Equals(const FunctionData &other_p) const override {
+		auto &other = other_p.Cast<BetweenFunctionData>();
+		return lower_inclusive == other.lower_inclusive && upper_inclusive == other.upper_inclusive;
+	}
+};
+
+void BetweenFunction(DataChunk &args, ExpressionState &state, Vector &result) {
```
Maxxen, line 31: So why is this a function at all? Can't we turn this into a pair of comparison expressions instead using the `bind_expression_t` callback? That would make `BETWEEN` work even for extension types as long as they implement `>`, `>=`, `<` and `<=` instead of just the types defined by `VectorOperations`

src/function/scalar/comparison/between.cpp
```diff
@@ -143,9 +144,14 @@ unique_ptr<FunctionData> BindBetweenFun(BindScalarFunctionInput &input) {
 	throw InvalidInputException("Between function cannot be called directly");
 }
 
+string BetweenToString(FunctionToStringInput &input) {
+	return BetweenExpression::ToString(*input.children[0], *input.children[1], *input.children[2]);
```
Maxxen, line 186: Is this important for round-tripping somehow? Since this is a `BoundExpression` anyway I don't see why we can't just print it like a normal function `between(a, b, c)` - it might even make it easier to read the plan IMO.

src/function/scalar/comparison/between.cpp
```diff
@@ -143,9 +144,14 @@ unique_ptr<FunctionData> BindBetweenFun(BindScalarFunctionInput &input) {
...
+}
+
 ScalarFunction BetweenFun::GetFunction() {
 	ScalarFunction between_fun("__between", {LogicalType::ANY, LogicalType::ANY, LogicalType::ANY},
```
Maxxen, line 214: Maybe we want to use templates here?

src/include/duckdb/function/scalar/comparison_functions.hpp
Mytherin: We don't support binding it at this point (that's still handled in `ExpressionBinder::BindExpression`).

src/function/scalar/comparison/between.cpp
Mytherin: There's two reasons:

* Optimization - `BETWEEN` is faster than running two comparisons
* Volatile functions - `random() BETWEEN x AND y` cannot be rewritten to `random() >= x AND random() <= y`

src/function/scalar/comparison/between.cpp
Mytherin: We're not supporting binding here at all yet - I would leave that for a potential follow-up

src/function/scalar/comparison/between.cpp
Maxxen: - I can see that, but we could still expand `ParsedBetweenExpressions` into comparisons in the binder, and the do an optimizer pass to fold them back into a `between(a, b, c)` function call if the types are supported. 

- Sure, but if the argument is volatile we could bind it as a lambda parameter and invoke the comparison in the body, using https://github.com/duckdb/duckdb/pull/20477 (although it got cut from v1.5)

But okay, future work.

src/function/scalar/comparison/between.cpp
Mytherin: I suppose that's true - but for backwards compatibility I'll keep it as a `BETWEEN` expression for now.

PR #22005, opened 2026-04-10
Unify constant filter with expression filter

kryonix:
This PR extracts the `ExpressionFilter` unification from https://github.com/duckdb/duckdb/pull/21762. Instead of doing everything at once, this PR only sets up some common infrastructure required for all TableFilter refactoring (eliminations), and removes `ConstantFilter` from all code paths except for legacy de/serialization.

There will be a few follow-up PRs regarding:
- [ ] `ConjunctionFilter`
- [ ] `DynamicFilter`
- [ ] `InFilter`
- [ ] `Is[Not]NullFilter`
- [ ] `OptionalFilter`
- [ ] `BloomFilter`
- [ ] `PerfectHashJoinFilter`
- [ ] `PrefixRangeFilter`
- [ ] `SelectivityOptionalFilter`
- [ ] `StructFilter`

PR #22496, opened 2026-05-06
Add comparison execution fast path

kryonix:
TableFilters used to have specialized `FilterSelection` logic before refactoring them (https://github.com/duckdb/duckdb/pull/22005). Removing this has led to a regression. This PR adds a similar fast path back to the system. But instead of being limited to TableFilters, it's included in the general `execute_comparison` logic.


Mytherin:
Thanks!

PR #22497, opened 2026-05-06
Parquet Prefetch Strategy

pdet:
This PR reworks Parquet prefetching to pick one of three strategies per row group, and adds a logger so we can inspect these decisions.                                                                                                                                                                               

 Strategies:                                                                                                                                                                                                                                                                                                             
 1. `whole_group` : issue one prefetch covering the entire row group. We pick it when filters look unselective (i.e., do not filtere all data out) and we're scanning `>=95%` of the row group.
2. `column_wise_eager` : We group projected/filtered columns based on their physical proximity, hitchhicker columns (i.e., gaps between our requested columns that are smaller than 16kb) can also be grouped together to diminish the number of requests.                                                                                                   
3. `prefetch_filters`: Similar to `column_wise_eager`. The main difference is that here we detected that the filters are most likely fully selective, hence we group and prefetch them before doing unselective filters and projected columns.                                                                         

To track selectivity, we count how many row groups had any surviving row after filtering, if most row groups have fully selective filters we then use the  `prefetch_filters` strategy. We also attempt to record the subset of filters that exclude all rows, and group these first, in the case with many filters, but only a subset contributes to full selectivity. 

For the logger, we introduce `ParquetPrefetch`. It basically stores, the strategy used for a row group, how it grouped the columns, and what is the `minimal_filters` subset. Hitchhiking columns are also presented betwen `*`

```sql
statement ok
SELECT sum(j)
FROM read_parquet('{TEST_DIR}/rgo_proj_gap.parquet')
WHERE a >= 0 AND c >= 0

query ITTT
SELECT row_group_id, fully_filtered, strategy, prefetch_groups
FROM duckdb_logs_parsed('ParquetPrefetch')
WHERE file_path LIKE '%rgo_proj_gap.parquet'
ORDER BY row_group_id
----
0	false	column_wise_eager	[[a, *b*, c], [j]]
1	false	column_wise_eager	[[a, *b*, c], [j]]
2	false	column_wise_eager	[[a, *b*, c], [j]]
3	false	column_wise_eager	[[a, *b*, c], [j]]

statement ok
SELECT sum(a + b + c + d)
FROM read_parquet('{TEST_DIR}/rgo_ab.parquet')
WHERE a > 50 AND b > 50 AND c = 0

query ITTTT
SELECT row_group_id, fully_filtered, strategy, prefetch_groups, minimal_filters
FROM duckdb_logs_parsed('ParquetPrefetch')
WHERE file_path LIKE '%rgo_ab.parquet'
ORDER BY row_group_id
----
0	true	whole_group	[[a, b, c, d]]	[a, b]
1	true	prefetch_filters	[[a, b], [c], [d]]	[a, b]
2	true	prefetch_filters	[[a, b], [c], [d]]	[a, b]
3	true	prefetch_filters	[[a, b], [c], [d]]	[a, b]

```

For benchmarking, we executed with a partitioned TPCH SF-100 and parquet files on S3, running Q6, with an EC2 machine on the same availability zone:

```
  ┌────────┬───────────────────┬───────────────────────────────────┬───────────────┐                                                                                                                                                                                                                                                                                   
  │  stat  │ main (63d32ec47d) │ parquet_survival-row (b15ab9b8b0) │       Δ       │                                                                                                                                                                                                                                                                                   
  ├────────┼───────────────────┼───────────────────────────────────┼───────────────┤                                                                                                                                                                                                                                                                                   
  │ min    │ 56.053s           │ 51.582s                           │ -4.5s (-8%)   │                                                                                                                                                                                                                                                                                   
  ├────────┼───────────────────┼───────────────────────────────────┼───────────────┤                                                                                                                                                                                                                                                                                   
  │ median │ 78.406s           │ 63.283s                           │ -15.1s (-19%) │                                                                                                                                                                                                                                                                                   
  ├────────┼───────────────────┼───────────────────────────────────┼───────────────┤
  │ mean   │ 72.930s           │ 62.105s                           │ -10.8s (-15%) │                                                                                                                                                                                                                                                                                   
  ├────────┼───────────────────┼───────────────────────────────────┼───────────────┤                                                                                                                                                                                                                                                                                   
  │ max    │ 85.603s           │ 73.053s                           │ -12.6s (-15%) │
  ├────────┼───────────────────┼───────────────────────────────────┼───────────────┤                                                                                                                                                                                                                                                                                   
  │ stdev  │ 12.027s           │ 8.170s                            │ -3.9s         │
  └────────┴───────────────────┴───────────────────────────────────┴───────────────┘    
```

In general the benchmarks here are difficult as they can be very noisy, when using my laptop I managed to see 30% difference in `stdev` and `mean`.


Lineitem has the following column order:
`  [l_orderkey, l_partkey, l_suppkey, l_linenumber, l_quantity, l_extendedprice, l_discount, l_tax, l_returnflag, l_linestatus, l_shipdate, l_commitdate, l_receiptdate, l_shipinstruct, l_shipmode, l_comment]`

Q6 filters on `l_shipdate`, `l_discount` and `l_quantity.` and projects `l_extendedprice`. Since the filters are never fully selective, as seen below:
```
┌──────────────────┬────────────────┬────────────────────┬──────────────────┬───────────────────┬─────────────┐
│ total_row_groups │ fully_filtered │ not_fully_filtered │ prefetch_filters │ column_wise_eager │ whole_group │
│      int64       │     int64      │       int64        │      int64       │       int64       │    int64    │
├──────────────────┼────────────────┼────────────────────┼──────────────────┼───────────────────┼─────────────┤
│             4844 │              0 │               4844 │                0 │              4844 │           0 │
└──────────────────┴────────────────┴────────────────────┴──────────────────┴───────────────────┴─────────────┘
```

all row groups use the `column_wise_eager` and group columns in two seperate requests as follows `[[l_quantity, l_extendedprice, l_discount], [l_shipdate]]`.


extension/parquet/parquet_reader.cpp
```diff
@@ -1373,6 +1481,9 @@ void ParquetReader::InitializeScan(ClientContext &context, ParquetReaderScanStat
 			state.scan_filters.emplace_back(context, entry.GetIndex(), entry.Filter());
 		}
 	}
+	if (state.filter_eliminated_all_rows.size() != state.scan_filters.size()) {
+		state.filter_eliminated_all_rows.assign(state.scan_filters.size(), false);
+	}
```
lnkuiper, line 1451: This flag is set when a filter brings `filter_count` to 0, and it's never cleared. In `ColumnWisePrefetch` we split after every filter that has the flag set. This has a few consequences:

 - The order in which filters are evaluated (which is adaptive through `AdaptiveFilter`) matters. Also, if filter a does not eliminate all rows, and filter b does not eliminate all rows, but a and b together eliminate all rows, then the flag is set on whichever filter ran last in that row group. Over time multiple filters can end up flagged, and it's never cleared.
 - Once multiple filters are flagged, row groups that don't fully eliminate pay one roundtrip per filter group instead of one combined fetch. Since this PR's main goal is reducing fetches/roundtrips on remote storage, this could be a real regression (although the impact might be limited).
 - There is eventual correction through `PREFETCH_FILTER_MINIMUM_MATCH_RATIO` to flip back to fetching the whole row group, so we don't stay in column-wise feching mode forever, but it could cost a few roundtrips before this is corrected.

Not a blocker, would just like to hear your thoughts on this.

extension/parquet/parquet_reader.cpp
```diff
@@ -1524,50 +1733,22 @@ AsyncResult ParquetReader::Scan(ClientContext &context, ParquetReaderScanState &
...
-							has_non_optional_filter = true;
-						}
-					}
-					if (has_non_optional_filter) {
-						lazy_fetch = true;
-					}
+			bool filters_look_unselective = false;
+			if (filters) {
+				if (state.prefetch_metrics.row_groups_executed == 0) {
+					filters_look_unselective = true;
```
lnkuiper: This defaults to `true` at the start, but I was wondering if we could maybe use cardinality estimates to initialize better? We could use cardinality estimates, or even better, we might know from the min/max values how many row groups survive, and use that as a guide to initialize this to `true`/`false` based on that?

Also, did you try running the benchmark with this initialized to `false` to compare the performance? Just curious.

Also not a blocker.

extension/parquet/parquet_reader.cpp
```diff
@@ -1626,10 +1807,22 @@ AsyncResult ParquetReader::Scan(ClientContext &context, ParquetReaderScanState &
 				ColumnReaderInput reader_input(scan_count, define_ptr, repeat_ptr);
 				child_reader.Filter(reader_input, result_vector, scan_filter.filter, *scan_filter.filter_state,
 				                    state.sel, filter_count, is_first_filter);
+				auto &filters_used = state.prefetch_metrics.logger.filters_used;
```
lnkuiper: Can we gate this behind `log_prefetch`? I don't think this is necessary if that's `false`.

extension/parquet/include/parquet_reader.hpp
```diff
@@ -124,10 +192,14 @@ struct ParquetReaderScanState {
 	bool prefetch_mode = false;
 	bool current_group_prefetched = false;
 
+	ParquetPrefetchMetrics prefetch_metrics;
```
lnkuiper, line 209: These metrics are per-thread and reset per file. I wonder how common it is for us to read, for example, a Parquet file containing 20 row groups on a box that has 32 threads. Then, each thread will read one row group and none of the metrics are reused for subsequent reads. I wonder if there is a way for us to get more mileage out of these metrics.

lnkuiper (commented):
Looks great! No blockers from my side, left a few comments though:

extension/parquet/include/parquet_reader.hpp
pdet: You are right that they are per-thread, that's by design so we can avoid any kind of locking. I think there is also not much we could realistically do in a scenario where threads > row_groups.

Wrt being reset per file, they are not, these metrics definetly "leak" to subsequent files in the same thread.

(I've added a new test to document this, will push it soon)

extension/parquet/parquet_reader.cpp
pdet: That's right, I guess using the cardinalities is a good idea, will have a go at that, previously this would default on the other way around.

extension/parquet/parquet_reader.cpp
pdet: Your analysis is absolutely right, and this was the design I came up with Mark. We could potentially clean it up after X row groups, I'm open to suggestions.

alippai:
Can this be configurable per query or globally? Some storage solutions prefer large sequential reads and have relatively high ramp-up time (but they achieve hundreds of gbps).

PR #22499, opened 2026-05-06
Add adaptive transient segment allocation

EtgarDev:
## Background
There are cases when we write single values / small batches, and the `IN_MEMORY_TABLE` buffer allocations are big regardless of the rows that are actually written.
@taniabogatsch brought up the idea of making those segments buffer allocations adaptive.
This also makes sense because we anyway compact the segments again when we `CHECKPOINT`.

## Solution
When `adaptive_segment_allocation` is enabled, transient column segments start at `initial_segment_rows * type_size` and double on each overflow, capped at `block_size`.

A `STANDARD_MASK_SIZE` floor is applied for validity (BIT) columns to avoid a dead-segment chain in `ValidityAppend` (`max_tuples` in https://github.com/duckdb/duckdb/blob/5826ebc40750a1af07a54fa0cefb7897fc9ffcdd/src/storage/compression/validity_uncompressed.cpp#L548 would be set to 0 without this floor)

New settings: `adaptive_segment_allocation` (bool, default `false`), `initial_segment_rows` (UBIGINT, default `2048`).

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -16,6 +16,13 @@
         ]
     },
     {
+        "name": "adaptive_segment_allocation",
+        "description": "When enabled, transient column segments start small and double in size with each overflow, up to the block size",
+        "type": "BOOLEAN",
+        "default_scope": "global",
+        "default_value": "false"
+    },
+    {
         "name": "allocator_background_threads",
         "description": "Whether to enable the allocator background thread.",
         "type": "BOOLEAN",
@@ -811,6 +818,13 @@
         "on_callbacks": [
             "set"
         ]
+    },
+    {
+        "name": "initial_segment_rows",
+        "description": "Number of rows allocated for the first transient column segment when adaptive_segment_allocation is enabled. Subsequent segments double this until reaching the block size.",
+        "type": "UBIGINT",
+        "default_scope": "global",
+        "default_value": "2048"
     },
     {
         "name": "integer_division",
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -16,13 +16,6 @@
         ]
     },
     {
-        "name": "adaptive_segment_allocation",
-        "description": "When enabled, transient column segments start small and double in size with each overflow, up to the block size",
-        "type": "BOOLEAN",
-        "default_scope": "global",
-        "default_value": "false"
-    },
-    {
         "name": "allocator_background_threads",
         "description": "Whether to enable the allocator background thread.",
         "type": "BOOLEAN",
@@ -820,8 +813,8 @@
         ]
     },
     {
-        "name": "initial_segment_rows",
-        "description": "Number of rows allocated for the first transient column segment when adaptive_segment_allocation is enabled. Subsequent segments double this until reaching the block size.",
+        "name": "initial_column_segment_size",
+        "description": "The initial memory (in bytes) reserved for the first transient column segment. Subsequent segments double in size until reaching the block size.",
         "type": "UBIGINT",
         "default_scope": "global",
         "default_value": "2048"
--- a/src/include/duckdb/storage/table/column_data.hpp
+++ b/src/include/duckdb/storage/table/column_data.hpp
@@ -254,6 +254,8 @@ protected:
 	mutable mutex stats_lock;
 	//! Total transient allocation size
 	atomic<idx_t> allocation_size;
+	//! Size of the last transient segment - drives the adaptive doubling in AppendTransientSegment
+	idx_t last_transient_segment_size = 0;
 	//! The stats of the root segment
 	unique_ptr<SegmentStatistics> stats;
 
--- a/src/storage/table/column_data.cpp
+++ b/src/storage/table/column_data.cpp
@@ -638,39 +638,24 @@ void ColumnData::UpdateColumn(TransactionData transaction, DuckTableEntry &table
 
 void ColumnData::AppendTransientSegment(SegmentLock &l, idx_t start_row) {
 	const auto block_size = block_manager.GetBlockSize();
-	const auto type_size = GetTypeIdSize(type.InternalType());
 
 	auto &db = GetDatabase();
 	auto &config = DBConfig::GetConfig(db);
 
 	idx_t segment_size;
-	if (Settings::Get<AdaptiveSegmentAllocationSetting>(config)) {
-		if (last_transient_segment_size == 0) {
-			const auto initial_rows = Settings::Get<InitialSegmentRowsSetting>(config);
-			segment_size = MinValue<idx_t>(block_size, MaxValue<idx_t>(type_size, initial_rows * type_size));
-		} else {
-			segment_size = MinValue<idx_t>(block_size, last_transient_segment_size * 2);
-		}
-		// BIT (validity) segments can only hold rows in multiples of STANDARD_VECTOR_SIZE;
-		// any segment below STANDARD_MASK_SIZE triggering a dead-segment overflow chain
-		if (type.InternalType() == PhysicalType::BIT) {
-			segment_size = MaxValue(segment_size, ValidityMask::STANDARD_MASK_SIZE);
-		}
-		last_transient_segment_size = segment_size;
+	if (last_transient_segment_size == 0) {
+		// Starting with the `initial_bytes` but making sure we have enough space for at least one row
+		const auto initial_bytes = Settings::Get<InitialColumnSegmentSizeSetting>(config);
+		segment_size = MinValue<idx_t>(block_size, MaxValue<idx_t>(GetTypeIdSize(type.InternalType()), initial_bytes));
 	} else {
-		auto vector_segment_size = block_size;
-
-		if (data_type == ColumnDataType::INITIAL_TRANSACTION_LOCAL && start_row == 0) {
-#if STANDARD_VECTOR_SIZE < 1024
-			vector_segment_size = 1024 * type_size;
-#else
-			vector_segment_size = STANDARD_VECTOR_SIZE * type_size;
-#endif
-		}
-
-		// The segment size is bound by the block size, but can be smaller.
-		segment_size = block_size < vector_segment_size ? block_size : vector_segment_size;
+		segment_size = MinValue<idx_t>(block_size, last_transient_segment_size * 2);
 	}
+	// BIT (validity) segments can only hold rows in multiples of STANDARD_VECTOR_SIZE;
+	// any segment below STANDARD_MASK_SIZE triggers a dead-segment overflow chain
+	if (type.InternalType() == PhysicalType::BIT) {
+		segment_size = MaxValue(segment_size, ValidityMask::STANDARD_MASK_SIZE);
+	}
+	last_transient_segment_size = segment_size;
 	allocation_size += segment_size;
 
 	auto function = config.GetCompressionFunction(CompressionType::COMPRESSION_UNCOMPRESSED, type.InternalType());
--- a/src/storage/table/column_data.cpp
+++ b/src/storage/table/column_data.cpp
@@ -1,6 +1,8 @@
 #include "duckdb/storage/table/column_data.hpp"
 
 #include "duckdb/common/exception/transaction_exception.hpp"
+#include "duckdb/common/types/validity_mask.hpp"
+#include "duckdb/main/settings.hpp"
 #include "duckdb/common/serializer/binary_deserializer.hpp"
 #include "duckdb/common/vector/flat_vector.hpp"
 #include "duckdb/common/serializer/serializer.hpp"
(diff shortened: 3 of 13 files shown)


src/include/duckdb/storage/table/column_data.hpp
```diff
@@ -254,6 +254,8 @@ class ColumnData : public enable_shared_from_this<ColumnData> {
 	mutable mutex stats_lock;
 	//! Total transient allocation size
 	atomic<idx_t> allocation_size;
+	//! Size of the last transient segment - drives the adaptive doubling in AppendTransientSegment
+	idx_t last_transient_segment_size = 0;
```
Mytherin, line 258: Instead of adding this variable (which will be reset to 0 on every database restart) could we not pass in the last segment into `AppendTransientSegment` and get the last segment size from there?

src/common/settings.json
```diff
@@ -15,6 +15,13 @@
             "set"
         ]
     },
+    {
+        "name": "adaptive_segment_allocation",
+        "description": "When enabled, transient column segments start small and double in size with each overflow, up to the block size",
```
taniabogatsch: > transient column segments start small and double in size with each overflow, up to the block size

I think we might want the adaptive segment allocation to be the default behavior (cc @Mytherin), i.e., we want this to always be enabled (no setting). If we double in size with each overflow and if we allow the initial segment size to be configurable, then I don't think that people need to configure anything here. If we have a sensible default value for the `initial_column_segment_size`.

src/common/settings.json
```diff
@@ -812,6 +819,13 @@
             "set"
         ]
     },
+    {
+        "name": "initial_segment_rows",
```
taniabogatsch: Should we rename this to something like `initial_column_segment_size` or similar?

Maybe we can also rewrite the description a bit: `The initial segment memory reserved by DuckDB for appends to a column (insert, not-in-place update, ...). Small values indicate few rows being appended, whereas bigger values reserve space for bulk allocations. Subsequent segments double in size until reaching the block size.`

src/include/duckdb/storage/table/column_data.hpp
taniabogatsch, line 258: nit: Let's use `optional_idx` here? That way it's slightly more explicit whether there is a previous segment or not. Maybe we can also rename to `prev_transient_segment_size`?

src/storage/table/column_data.cpp
```diff
@@ -637,24 +639,41 @@ void ColumnData::UpdateColumn(TransactionData transaction, DuckTableEntry &table
...
+		if (data_type == ColumnDataType::INITIAL_TRANSACTION_LOCAL && start_row == 0) {
 #if STANDARD_VECTOR_SIZE < 1024
-		vector_segment_size = 1024 * type_size;
+			vector_segment_size = 1024 * type_size;
 #else
-		vector_segment_size = STANDARD_VECTOR_SIZE * type_size;
+			vector_segment_size = STANDARD_VECTOR_SIZE * type_size;
 #endif
-	}
+		}
```
taniabogatsch: Let's get rid of this branch and only use the adaptive strategy.

src/storage/table/column_data.cpp
```diff
@@ -637,24 +639,41 @@ void ColumnData::UpdateColumn(TransactionData transaction, DuckTableEntry &table
...
+	auto &db = GetDatabase();
+	auto &config = DBConfig::GetConfig(db);
+
+	idx_t segment_size;
+	if (Settings::Get<AdaptiveSegmentAllocationSetting>(config)) {
+		if (last_transient_segment_size == 0) {
+			const auto initial_rows = Settings::Get<InitialSegmentRowsSetting>(config);
+			segment_size = MinValue<idx_t>(block_size, MaxValue<idx_t>(type_size, initial_rows * type_size));
+		} else {
+			segment_size = MinValue<idx_t>(block_size, last_transient_segment_size * 2);
```
taniabogatsch: Maybe to be a bit more memory efficient we can use `DUCKDB_API uint64_t NextPowerOfTwo(uint64_t v);` here to align this with allocations that are the power of two? WDYT? 

Or it might make sense to align the value here maybe if we don't just go to the next power of two?
```cpp
//! Align value (ceiling) (not for pointer types)
template<class T, T val=8, typename = typename std::enable_if<!std::is_pointer<T>::value>::type>
static inline T AlignValue(T n) {
	return ((n + (val - 1)) / val) * val;
}
```

src/storage/table/column_data.cpp
```diff
@@ -637,24 +639,41 @@ void ColumnData::UpdateColumn(TransactionData transaction, DuckTableEntry &table
...
 
-	if (data_type == ColumnDataType::INITIAL_TRANSACTION_LOCAL && start_row == 0) {
+	auto &db = GetDatabase();
+	auto &config = DBConfig::GetConfig(db);
+
+	idx_t segment_size;
+	if (Settings::Get<AdaptiveSegmentAllocationSetting>(config)) {
+		if (last_transient_segment_size == 0) {
+			const auto initial_rows = Settings::Get<InitialSegmentRowsSetting>(config);
+			segment_size = MinValue<idx_t>(block_size, MaxValue<idx_t>(type_size, initial_rows * type_size));
```
taniabogatsch: We can enforce that `InitialSegmentRowsSetting` is never `0` and then remove the `MaxValue` here.

src/storage/table/column_data.cpp
```diff
@@ -637,24 +639,41 @@ void ColumnData::UpdateColumn(TransactionData transaction, DuckTableEntry &table
...
-		vector_segment_size = STANDARD_VECTOR_SIZE * type_size;
+			vector_segment_size = STANDARD_VECTOR_SIZE * type_size;
 #endif
-	}
+		}
 
-	// The segment size is bound by the block size, but can be smaller.
-	idx_t segment_size = block_size < vector_segment_size ? block_size : vector_segment_size;
+		// The segment size is bound by the block size, but can be smaller.
+		segment_size = block_size < vector_segment_size ? block_size : vector_segment_size;
```
taniabogatsch: I think we can turn this into an assertion now, since (if we get rid of the `else` branch) we should never have assigned a value that exceeds `block_size`.

test/sql/storage/adaptive_segment_allocation.test
```diff
@@ -0,0 +1,219 @@
...
+CHECKPOINT
+
+statement ok
+CREATE TABLE mem_adaptive (a INT NOT NULL, b INT NOT NULL, c INT NOT NULL, d INT NOT NULL, e INT NOT NULL)
+
+statement ok
+INSERT INTO mem_adaptive VALUES (1, 2, 3, 4, 5)
+
+# adaptive: 5 cols * (8 bytes data + 256 bytes validity) = ~1320 bytes
+# non-adaptive allocates ~51200 bytes for the same insert
```
taniabogatsch: Really nice gains here! 🚀

test/sql/storage/adaptive_segment_allocation.test
taniabogatsch, line 1: Looks great - should we add a small error test to ensure that passing zero to the initial size raises an exception?

taniabogatsch (commented):
Hi, left some comments but looking great from my side already. 🚀

src/storage/table/column_data.cpp
```diff
@@ -637,24 +639,41 @@ void ColumnData::UpdateColumn(TransactionData transaction, DuckTableEntry &table
...
-	auto vector_segment_size = block_size;
 
-	if (data_type == ColumnDataType::INITIAL_TRANSACTION_LOCAL && start_row == 0) {
+	auto &db = GetDatabase();
+	auto &config = DBConfig::GetConfig(db);
+
+	idx_t segment_size;
+	if (Settings::Get<AdaptiveSegmentAllocationSetting>(config)) {
+		if (last_transient_segment_size == 0) {
+			const auto initial_rows = Settings::Get<InitialSegmentRowsSetting>(config);
```
Mytherin: I think using "rows" is a misnomer here - this only really models the number of rows for fixed-width types. For variable length types (strings, nested types) this is just a "bag of bytes" that will be filled. As such I don't think it makes much sense to model this as "# of rows". Let's instead change this to just "bytes" and then start with e.g. 1-2KB or so regardless of column type.

Mytherin (commented):
Thanks - looks good, agreed with @taniabogatsch we should just always switch to this strategy. Then the test config can also go. Left some more comments

EtgarDev:
@Mytherin @taniabogatsch 
Thank you for the reviews! :)
I'll work on it and hope to ship the fixes soon.

Regarding the default behaviour + size instead of amount of rows - I agree, (1) size is way more descriptive and it makes things more expected, and (2) the adaptive allocation is something that would be relevant to tons of workloads.

However, I wanted to flag something making this new adaptive behaviour to be the default - setting this as the default (and only) behaviour introduces a tradeoff worth a discussion, regardless of what number we set as the default.

The old behaviour was:
```python
first segment = STANDARD_VECTOR_SIZE * type_size # e.g. 16KB for BIGINT
```
and then immediately `block_size` (256KB) for every overflow.

With doubling from 2KB, the ramp-up produces segments of 2KB -> 4KB -> 8KB -> ... -> 128KB -> 256KB, and this ramp up footprint converges to `2 * block_size`  ~512KB per column, compared to ~16KB in the old path.

So the 2KB default (or actually any starting point we will choose) saves memory for small writes, but increases peak `IN_MEMORY_TABLE` usage for bulk inserts compared to the old behaviour.

Two options as I see it:
1. Keep the doubling default and accept the bulk-insert tradeoff
2. Default to `0` meaning auto `STANDARD_VECTOR_SIZE * type_size` as a first segment, and then `block_size` for the next segments, so small-write workloads opt in.

WDYT?

Mytherin:
We should never exceed `block_size` in this doubling, it should always be capped at that amount imo. All that should change here is the ramp up (`2K -> 4K -> ... -> [BLOCK_SIZE]` instead of `8K -> [BLOCK_SIZE]`).

EtgarDev:
> We should never exceed `block_size` in this doubling, it should always be capped at that amount imo. All that should change here is the ramp up (`2K -> 4K -> ... -> [BLOCK_SIZE]` instead of `8K -> [BLOCK_SIZE]`).

@Mytherin 
Yes that's true, I have a cap of `block_size` of course.
After recalculating everything my concern is not real.

However I found another issue in regards to a test we have: `test/sql/storage/buffer_manager_temp_dir.test` where we verify that a large insert under a tight memory limit succeeds by evicting completed blocks to disk.
However, with the doubling ramp-up it fails - because the ramp-up segments (2KB -> 128KB) are allocated via `RegisterSmallMemory` which makes them non-evictable. In the current behaviour where the first segment is 16KB and the rest are 256KB, only the first is unevictable so we will manage to spill more memory to the block files without reaching to OOM.

I thought that we could maybe somehow enable `skip_register_small_memory` parameter to https://github.com/duckdb/duckdb/blob/8f3e918202324db47c05c75fee7eebd9c6a39c12/src/storage/standard_buffer_manager.cpp#L133 but then I understood that `Allocates` always allocates a full `block_size` buffer and it's not just about making the allocation evictable or not.

WDYT? Should we maybe change expectation for tight-memory behaviour but keep the adaptive behaviour as the default?

EDIT - I'm actually not sure about the fact that `RegisterSmallMemory` is unevictable, it's an hypothesis. I'm still trying to figure our what makes the test to fail under those circumstances.

A second edit - OK I think my hypothesis is correct:
https://github.com/duckdb/duckdb/blob/8f3e918202324db47c05c75fee7eebd9c6a39c12/src/storage/standard_buffer_manager.cpp#L402-L404
It seems like `TINY_BUFFER` allocations (from `RegisterSmallMemory`) results in no-op under `Unpin`. Is a change over there trivial? I guess we will also have to add support for this kind of buffers in other places.

src/common/settings.json
EtgarDev: fixed :)

src/storage/table/column_data.cpp
EtgarDev: fixed

Mytherin:
It should be possible to make these allocations evictable but we need to make sure we are not writing them to individual files but grouping them together in larger files. We pool same-sized blocks into temporary files in batches of these available sizes:

```cpp
enum class TemporaryBufferSize : uint64_t {
	INVALID = 0,
	S32K = 32768,
	S64K = 65536,
	S96K = 98304,
	S128K = 131072,
	S160K = 163840,
	S192K = 196608,
	S224K = 229376,
	DEFAULT = DEFAULT_BLOCK_ALLOC_SIZE,
};
```

So it would make sense to allow spilling starting from 32KB (although these are post-compression, so we will likely still be wasting space).

We could also consider expanding this list to include a few even smaller sizes, e.g. 2KB, 4KB, 8KB, 16KB. 

I would say that's a separate PR though. For this PR I wouldn't mind slightly increasing the memory limit for that specific test.

src/storage/table/column_data.cpp
EtgarDev: fixed :)

EtgarDev:
@Mytherin Nice, thanks, I'll try to look into it

taniabogatsch:
Superseded by https://github.com/duckdb/duckdb/pull/22542

src/include/duckdb/storage/table/column_data.hpp
taniabogatsch: fixed 👍

PR #22500, opened 2026-05-06
Overload ART::Scan for batch equality lookups

tanner-lai:
Overload ART::Scan for batch equality lookups

Adds a Scan overload that takes a DataChunk of keys and returns all
matching row_ids. Acquires the lock once for the entire batch.
Follows the same pattern as Insert/Append/TryDelete.

Extensions resolving batches of PK values to row_ids currently call
Scan in a loop, paying per-call mutex + key generation overhead. In
my testing scanning in batch improved performance at 100K lookups on
10M rows from 20 to 4 ms (on m1 pro).

Tests cover INTEGER, BIGINT, VARCHAR, composite keys, unique indexes,
and non-unique indexes with hit/miss/empty/full-batch cases.

### DIFF
--- a/src/execution/index/art/art.cpp
+++ b/src/execution/index/art/art.cpp
@@ -801,6 +801,24 @@ bool ART::Scan(IndexScanState &state, const idx_t max_count, set<row_t> &row_ids
 	return SearchCloseRange(key, upper_bound, left_equal, right_equal, max_count, row_ids);
 }
 
+void ART::Scan(DataChunk &input, vector<row_t> &row_ids) {
+	ArenaAllocator arena(Allocator::Get(db));
+	unsafe_vector<ARTKey> keys(input.size());
+	GenerateKeys(arena, input, keys);
+
+	lock_guard<mutex> l(lock);
+	for (idx_t i = 0; i < input.size(); i++) {
+		if (keys[i].Empty()) {
+			continue;
+		}
+		set<row_t> result;
+		SearchEqual(keys[i], NumericLimits<idx_t>::Maximum(), result);
+		for (auto &rid : result) {
+			row_ids.push_back(rid);
+		}
+	}
+}
+
 //===--------------------------------------------------------------------===//
 // More Constraint Checking
 //===--------------------------------------------------------------------===//
--- a/src/include/duckdb/execution/index/art/art.hpp
+++ b/src/include/duckdb/execution/index/art/art.hpp
@@ -79,6 +79,10 @@ public:
 	//! If all row IDs were fetched, it return true, else false.
 	bool Scan(IndexScanState &state, idx_t max_count, set<row_t> &row_ids);
 
+	//! Batch equality scan: looks up all keys in the DataChunk and appends matching row_ids to the output.
+	//! Acquires the lock once for the entire batch. Supports non-unique indexes (multiple row_ids per key).
+	void Scan(DataChunk &input, vector<row_t> &row_ids);
+
 	//! Simple merge: scan source ART and delete each (key, rowid) from this ART.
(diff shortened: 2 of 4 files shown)


tanner-lai:
had to fix the test to work with the vector size ci job

Mytherin:
Thanks - is there any place in the code we could benefit from this? Generally we strongly prefer to avoid adding code that is only used in C++ tests (and strongly prefer adding C++ tests in general).

taniabogatsch:
We could benchmark/consider replacing the `TryScanIndex` code with this batched approach in the presence of many filter expressions to scan (e.g., `IN` filter). 

```cpp
	for (const auto &filter_expr : expressions) {
		for (auto &art_ref : arts_to_scan) {
			auto &art_to_scan = art_ref.get();
			auto scan_state = art_to_scan.TryInitializeScan(*index_expr, *filter_expr);
			if (!scan_state) {
				return false;
			}

			// Check if we can use an index scan, and already retrieve the matching row ids.
			if (!art_to_scan.Scan(*scan_state, max_count, row_ids)) {
				row_ids.clear();
				return false;
			}
		}
	}
```

I think maybe the `ARTIndexScanState` could contain a data chunk of the key columns (would currently contain only one vector) of ALL keys that we want to scan, instead of the current per-iteration (per-filter-expression) values. 

Of course we cap the `IN` filter count at which we do index scans at 100 IIRC so I am not sure if we would see gains already at that size.

taniabogatsch:
Additional note (on Mark's comment): `and strongly prefer adding SQL tests in general`

tanner-lai:
> We could benchmark/consider replacing the `TryScanIndex` code with this batched approach in the presence of many filter expressions to scan (e.g., `IN` filter).
> 
> ```c++
> 	for (const auto &filter_expr : expressions) {
> 		for (auto &art_ref : arts_to_scan) {
> 			auto &art_to_scan = art_ref.get();
> 			auto scan_state = art_to_scan.TryInitializeScan(*index_expr, *filter_expr);
> 			if (!scan_state) {
> 				return false;
> 			}
> 
> 			// Check if we can use an index scan, and already retrieve the matching row ids.
> 			if (!art_to_scan.Scan(*scan_state, max_count, row_ids)) {
> 				row_ids.clear();
> 				return false;
> 			}
> 		}
> 	}
> ```
> 
> I think maybe the `ARTIndexScanState` could contain a data chunk of the key columns (would currently contain only one vector) of ALL keys that we want to scan, instead of the current per-iteration (per-filter-expression) values.
> 
> Of course we cap the `IN` filter count at which we do index scans at 100 IIRC so I am not sure if we would see gains already at that size.

Let me look into this, I had tried a more minimal change to have IN filter execution use the datachunk parameterized scan but it wasn't much of a win. Thanks for the suggestion!

taniabogatsch:
> but it wasn't much of a win

Yes, my hunch is that we spend significantly more time fetching the rows from storage (probe) than in the ART scan, so the gains will be marginal here (but maybe using a bulk-scan anyways could be nice for future use, too)... I also left some more ideas in the linked issue, I do see this function being useful in the future for different ART scenarios.

tanner-lai:
> > but it wasn't much of a win
> 
> Yes, my hunch is that we spend significantly more time fetching the rows from storage (probe) than in the ART scan, so the gains will be marginal here (but maybe using a bulk-scan anyways could be nice for future use, too)... I also left some more ideas in the linked issue, I do see this function being useful in the future for different ART scenarios.

Just wanted to follow up, did you end up identifying any use cases for this API?

taniabogatsch:
@artjomPlaunov is currently looking into the second part of work for vacuuming of indexes, and that's where I think it most likely to have a use case.

PR #22501, opened 2026-05-07
Fix bare numeric interval parsing at end of string

### LINKED ISSUE #22484: `TRY_CAST('42' AS INTERVAL)` returns NULL but `TRY_CAST('42 ' AS INTERVAL)` returns 42 seconds
## Describe the bug

Casting a bare integer string to `INTERVAL` returns `NULL`, but adding any trailing
whitespace to the same string makes it parse correctly as seconds. Leading whitespace
does not help — only a trailing character after the digits triggers the correct behavior.

```sql
SELECT TRY_CAST('42 seconds' AS INTERVAL);  -- 00:00:42  ✓
SELECT TRY_CAST('42 '        AS INTERVAL);  -- 00:00:42  ✓  trailing space works
SELECT TRY_CAST('42'         AS INTERVAL);  -- NULL      ✗  no trailing char → NULL
SELECT TRY_CAST(' 42'        AS INTERVAL);  -- NULL      ✗  leading space doesn't help
```

## Expected behavior

All four statements should return `00:00:42`. A bare integer with no unit should be
interpreted as seconds, consistent with how `'42 seconds'` and `'  42  '` are handled.

## Actual behavior

`'42'` and `' 42'` return `NULL`. Any input where the digits reach end-of-string
without a trailing character returns `NULL`.

## Root cause (identified)

`Interval::FromCString` in `src/common/types/interval.cpp` uses a goto-based state
machine. The number-parsing loop (`interval_parse_number`, ~line 96) has two exit paths:

- A **non-digit character** terminates the number → `goto interval_parse_identifier` (line 127)
- **End-of-string** terminates the number → `goto end_of_string` (line 130)

The "bare number implies seconds" special case lives in `interval_parse_identifier` (~line 169):

```cpp
// Special case SS[.FFFFFF] - implied SECONDS/MICROSECONDS
if (specifier_str.empty() && !found_any) {
    IntervalTryAddition<int64_t>(result.micros, number, MICROS_PER_SEC);
    found_any = true;
    // consume trailing whitespace...
    goto end_of_string;
}
```

When the input is `'42'`, the number loop exits via `goto end_of_string` at line 130,
bypassing `interval_parse_identifier` entirely. `found_any` remains `false` and the
function returns `false` → `NULL`.

When the input is `'42 '`, the trailing space is a non-digit character, so the loop
exits via `goto interval_parse_identifier` at line 127. The special case fires and 42
seconds is returned.

## Suggested fix

The `goto end_of_string` at line 130 should also apply the implied-seconds logic when
`!found_any`, or fall through to `interval_parse_identifier` unconditionally.

## Environment

- Verified on DuckDB 

### To Reproduce

```sql
SELECT TRY_CAST('42 seconds' AS INTERVAL);  -- 00:00:42  ✓
SELECT TRY_CAST('42 '        AS INTERVAL);  -- 00:00:42  ✓  trailing space works
SELECT TRY_CAST('42'         AS INTERVAL);  -- NULL      ✗  no trailing char → NULL
SELECT TRY_CAST(' 42'        AS INTERVAL);  -- NULL      ✗  leading space doesn't help
```

### OS:

OSX 26.3.1 arm64

### DuckDB Version:

**v1.1.3** and **v1.5.2**

### DuckDB Client:

duckdb cli

### Hardware:

Macbook Pro

### Full Name:

Dave York

### Affiliation:

Govini

- [x] Yes, I have

- [x] Yes, I have

M1llerF:
Fixes #22484 

The interval parser handled bare numeric strings inconsistently depending on whether parsing terminated at a non-digit character or at end-of-string.

Inputs like `42 ` properly triggered implied-seconds handling while inputs like `42` and ` 42` bypassed that path and failed to parse with the implied-seconds handling. This change routes end-of-string numeric termination through the same implied-seconds handling path. 

Validation: 

- Added regression test for bare integer interval parsing (see `test/issues/general/test_22484.test`)
- Ran `test/issues/general/test_22484.test`
- Ran `test/sql/types/interval/test_interval.test`
- Ran `make format-fix`


hawkfish:
Hi @M1llerF - thanks for the PR!

I am seeing a test failure on your branch: 

```sql
1. test/sql/types/interval/test_interval.test:392
Query unexpectedly succeeded! (test/sql/types/interval/test_interval.test:392)!
SELECT '100'::INTERVAL;
CAST('100' AS INTERVAL)	
INTERVAL	
[ Rows: 1]
00:01:40
```

I use a shell script to generate a list of test files and then run them all with the `-f` flag like so:

```sh
$ cat smoke.sh
# !/bin/sh
git grep -E -l -w -i $1 -- '*.test' > smoke.test
git grep -E -l -w -i $1 -- '*.test_*' > smoke.test_slow
cat smoke.test smoke.test_slow | wc -l 
```

Then to get all the tests that use `INTERVAL`:

```sh
$ ./smoke.sh interval
$ ./build/release/test/unittest -f smoke.test
$ ./build/release/test/unittest -f smoke.test_slow
$ ./build/debug/test/unittest -f smoke.test
```

test/issues/general/test_22484.test
```diff
@@ -0,0 +1,18 @@
+# name: test/issues/general/test_22484.test
```
hawkfish: I think this can just go in the existing `test/sql/types/interval/interval_try_cast.test`

src/common/types/interval.cpp
```diff
@@ -129,7 +129,19 @@ bool Interval::FromCString(const char *str, idx_t len, interval_t &result, strin
...
 		}
 	}
-	goto end_of_string;
+	{
+		if (pos == start_pos) {
+			return false;
+		}
+		string_t nr_string(str + start_pos, UnsafeNumericCast<uint32_t>(pos - start_pos));
+		number = Cast::Operation<string_t, int64_t>(nr_string);
+		fraction = 0;
```
hawkfish, line 114: This will drop fractions.  I think the code needs to `break` at the last `else` in the loop and move the rest of the else clause into your bracket (instead of the copy-and-paste of part of it).

hawkfish (changes requested):
Also, fix the test that now succeeds.

src/common/types/interval.cpp
```diff
@@ -102,34 +102,35 @@ bool Interval::FromCString(const char *str, idx_t len, interval_t &result, strin
 			// colon: we are parsing a time
 			goto interval_parse_time;
 		} else {
-			if (pos == start_pos) {
```
hawkfish, line 105: I think this test needs to be moved too?

hawkfish (changes requested):
Getting there - and good work on adding the fractional tests.

src/common/types/interval.cpp
M1llerF: Yep it's at line 109 in the block after the loop

hawkfish: Sorry - GitHub is doing something strange when I try to look at the diff. And I can't resolve comments...

hawkfish (approved):

src/common/types/interval.cpp
M1llerF: No worries - appreciate the review!

hawkfish:
Hi @M1llerF - I have one more request: This is a correctness issue, so I'd like it to get into 1.5.3 next week.  Can you resubmit it as a merge to that branch?

M1llerF:
@hawkfish Opened a port PR targeting v1.5-variegata: #22534

PR #22503, opened 2026-05-07
Fix BlockAllocator invalid memory access

dentiny:
In the current implementation, we suffer invalid memory access with the unit test.
ASAN reports see below
```sh
hjiang@hjiangs-MacBook-Pro$ ./build/debug/test/unittest "[block_allocator]"
Filters: [block_allocator]
[0/1] (0%): BlockAllocator usage and de-allocation on different threads                                                              =================================================================
==23765==ERROR: AddressSanitizer: stack-use-after-scope on address 0x00016f4ff6f8 at pc 0x00010d437368 bp 0x00016f58abf0 sp 0x00016f58abe8
READ of size 8 at 0x00016f4ff6f8 thread T1
    #0 0x00010d437364 in std::__1::unique_ptr<duckdb::BlockQueue, std::__1::default_delete<duckdb::BlockQueue>>::get[abi:ne200100]() const unique_ptr.h:281
    #1 0x00010d2cddd4 in duckdb::unique_ptr<duckdb::BlockQueue, std::__1::default_delete<duckdb::BlockQueue>, false>::operator->() const unique_ptr.hpp:41
    #2 0x00010d2cd000 in duckdb::BlockAllocatorThreadLocalState::Clear() block_allocator.cpp:141
    #3 0x00010d434348 in duckdb::BlockAllocatorThreadLocalState::~BlockAllocatorThreadLocalState() block_allocator.cpp:97
    #4 0x00010d2cb87c in duckdb::BlockAllocatorThreadLocalState::~BlockAllocatorThreadLocalState() block_allocator.cpp:96
    #5 0x00019527fdc8 in invocation function for block in dyld::ThreadLocalVariables::finalizeList(void*)+0x34 (libdyld.dylib:arm64e+0x3dc8)
```

The root cause is (which is the plain english explanation for the unit test)
- Assume we have two threads: A and B
- BlockAllocator is created in thread-A, and allocates memory in thread-B, which means there's one [thread-local state](https://github.com/duckdb/duckdb/blob/67a0a733f12119424dddee68aed6ad271792ec42/src/storage/block_allocator.cpp#L204) created in thread-B and records free blocks
- When BlockAllocator is destructed in thread-A, a new thread-local state will be created and cleared => thread-A's thread-local state is unchanged
- Now when thread-A destructs, it [attempts to access BlockAllocator](https://github.com/duckdb/duckdb/blob/67a0a733f12119424dddee68aed6ad271792ec42/src/storage/block_allocator.cpp#L139-L149), which leads to invalid memory access

In one word, there's no guarantee on lifecycle for BlockAllocator and threa-local state, this PR simply adds a guard to block possible invalid access.

BlockAllocator is a [per-db instance](https://github.com/duckdb/duckdb/blob/67a0a733f12119424dddee68aed6ad271792ec42/src/include/duckdb/main/config.hpp#L175-L176), so the issue could happen in cases where user creates a DB in one worker thread and performs query.

### DIFF
--- a/src/include/duckdb/storage/block_allocator.hpp
+++ b/src/include/duckdb/storage/block_allocator.hpp
@@ -12,6 +12,7 @@
 #include "duckdb/common/unique_ptr.hpp"
 #include "duckdb/common/mutex.hpp"
 #include "duckdb/common/optional_idx.hpp"
+#include "duckdb/common/shared_ptr.hpp"
 
 namespace duckdb {
 
@@ -86,6 +87,9 @@ private:
 	unsafe_unique_ptr<BlockQueue> untouched;
 	//! Touched by block IDs
 	unsafe_unique_ptr<BlockQueue> touched;
+
+	//! Token used to indicate whether current BlockAllocator is alive.
+	shared_ptr<atomic<bool>> alive_token;
 };
 
 } // namespace duckdb
--- a/src/storage/block_allocator.cpp
+++ b/src/storage/block_allocator.cpp
@@ -136,21 +136,24 @@ public:
 	}
 
 	void Clear() {
-		// Return all local blocks back to global
-		if (!touched.empty()) {
-			block_allocator->touched->q.enqueue_bulk(touched.begin(), touched.size());
-			touched.clear();
+		if (alive_token && alive_token->load()) {
+			// Allocator is still alive — return local blocks to global queues
+			if (!touched.empty()) {
+				block_allocator->touched->q.enqueue_bulk(touched.begin(), touched.size());
+			}
+			if (!untouched.empty()) {
+				block_allocator->untouched->q.enqueue_bulk(untouched.begin(), untouched.size());
+			}
 		}
-		if (!untouched.empty()) {
-			block_allocator->untouched->q.enqueue_bulk(untouched.begin(), untouched.size());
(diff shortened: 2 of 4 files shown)


dentiny:
```sh
error: building azure-storage-blobs-cpp:x64-linux-release failed with: BUILD_FAILED
See https://learn.microsoft.com/vcpkg/troubleshoot/build-failures?WT.mc_id=vcpkg_inproduct_cli for more information.
Elapsed time to handle azure-storage-blobs-cpp:x64-linux-release: 7.5 s
Please ensure you're using the latest port files with `git pull` and `vcpkg update`.
Then check for known issues at:
  https://github.com/microsoft/vcpkg/issues?q=is%3Aissue+is%3Aopen+in%3Atitle+azure-storage-blobs-cpp
You can submit a new issue at:
  https://github.com/microsoft/vcpkg/issues/new?title=%5Bazure-storage-blobs-cpp%5D%20build%20error%20on%20x64-linux-release&body=Copy%20issue%20body%20from%20%2Fduckdb_build_dir%2Fbuild%2Frelease%2Fvcpkg_installed%2Fvcpkg%2Fissue_body.md
```
Hmm unrelated failure

lnkuiper:
Great find, thanks! LGTM once CI passes.

PR #22225, opened 2026-04-22
Hive-partitioned Write Rework

lnkuiper:
This PR fully reimplements `COPY ... (PARTITION_BY ...)`, aka hive-partitioned writes.

### Implementation

This PR introduces `PartitionedCopy`, which implements hive-partitioned writes, and moves all logic for this out of the `PhysicalCopyToFile` class and corresponding local/global states. This creates a better separation between partitioned and non-partitioned writes.
 
`PartitionedCopy` uses `SortStrategy`, which is capable of partitioning/sorting as it was originally implemented for the `PhysicalWindow`. Threads sink data into the `SortStrategy`, and once `partitioned_write_flush_threshold` is reached by at least one thread, threads exchange data and flush it out. This solves one of the problems in the current hive-partitioned write implementation, which is that threads do not exchange data with each other, which can lead to many tiny row groups being written.

This exchange happens during parallel query execution, for which I've adapted the parallel state machine that is implemented in `PhysicalWindow`. The partitioning/sorting/masking phases are the same as in the `PhysicalWindow`. After that, we get two phases that are specific to file writing, which are `BATCH` and `FLUSH`. These respectively create and flush a batch, _i.e._, row group.

### Benchmark

| Benchmark | Threads | Current [s] | This PR [s] | Relative Time |
|:-|-:|-:|-:|:-|
| `to_parquet_partition_by_few.benchmark` (1 column, 2 partitions) | 2 | **0.13** | 0.19 | ~1.46x slower |
| `to_parquet_partition_by_few.benchmark` (1 column, 2 partitions) | 10 | **0.47** | 0.62 | ~1.32x slower |
| `to_parquet_partition_by_many.benchmark` (1 column, 10 partitions) | 2 | 2.21 - 3.25 | **1.54 - 1.62** | ~0.58x faster |
| `to_parquet_partition_by_many.benchmark` (1 column, 10 partitions) | 10 | 1.15 - 1.25 | **0.53 - 0.61** | ~0.48x faster |

The current implementation is better when there are fewer partitions. This PR scales much better with the number of partitions (and number of threads!). The difference here is also explained by the benchmark file only writing one column to the output file. If we have more columns, the new implementation starts winning out. Consider these queries:
```sql
 -- 4 partitions
COPY tpch_sf10.lineitem TO 'my' (FORMAT PARQUET, PARTITION_BY (l_returnflag, l_linestatus), OVERWRITE);
 -- 50 partitions
[...]

PR #22505, opened 2026-05-07
Adapt Hive-partitioned write flushing to number of partitions

lnkuiper:
Addresses the issue raised in https://github.com/duckdb/duckdb/pull/22225#issuecomment-4299804259.

This is the simplest fix for the issue that I could think of, which should cover most use cases, but if needed, the approach could be improved in a future PR.

## HyperLogLog!

I've added a `HyperLogLog` to the `Sink` phase of `PartitionedCopyState`.

The resulting estimated unique count is used to estimate the number of rows per partition. Only if the number of rows per partition exceeds the `partitioned_write_flush_threshold` setting, we initiate a flush. If there are very few partitions, the behaviour is very similar to before: we flush after at least one thread accumulates `partitioned_write_flush_threshold` rows. If there are many partitions, the flush is delayed according to the rule above. The performance impact of computing `HyperLogLog` is negligible.

This is somewhat coarse-grained and not skew-resistant, but despite being such a simple approach it is already a significantly improvement over the current behaviour.

For the large partition count query in https://github.com/duckdb/duckdb/pull/22225#issuecomment-4299804259 (partitioning `lineitem` by `l_shipdate`) the number of files/row groups is minimized: there are 2526 files, each with one row group.

For the small partition count query (partitioning `lineitem` by `l_returnflag, l_linestatus`), the number of files/row groups is unchanged.

### DIFF
--- a/src/common/sort/hashed_sort.cpp
+++ b/src/common/sort/hashed_sort.cpp
@@ -1,6 +1,7 @@
 #include "duckdb/common/sorting/hashed_sort.hpp"
 #include "duckdb/common/sorting/sorted_run.hpp"
 #include "duckdb/common/radix_partitioning.hpp"
+#include "duckdb/common/types/hyperloglog.hpp"
 #include "duckdb/parallel/thread_context.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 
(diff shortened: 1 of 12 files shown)


lnkuiper:
Apparently the cost of computing HLL is noticeable for the single-column write 2-partition benchmark:
```
  benchmark/micro/copy/to_parquet_partition_by_few.benchmark
  Old timing: 0.862488
  New timing: 1.032004
  Delta: +0.170s (19.65%)
```

But the single-column write 1000-partition benchmark improves by more:
```
  benchmark/micro/copy/to_parquet_partition_by_many.benchmark
  Old timing: 2.170336
  New timing: 1.569719
```

I think we should accept this trade-off

PR #21927, opened 2026-04-07
Run ArrowConverter::ToArrowSchema in a transaction

evertlammerts:
Fixes https://github.com/duckdb/duckdb-spatial/issues/788

When populating the arrow schema we might need access to the catalog. In this case the GeoArrow extension needs to figure out the CRS type in `ArrowGeometry::PopulateSchema`. This works fine in the context of an active transaction, but not when consuming the results as a stream (as `RecordBatch` or the Arrow C Stream API). E.g. through the C API or the python client.

This PR wraps `ArrowConverter::ToArrowSchema` in `RunFunctionInTransaction` to maek sure we always have a transaction.

It also fixes a bug in `duckdb_query_arrow_schema` which did not set an error if it occurred, resulting in an error if the error was fetched.

PR #22508, opened 2026-05-07
Add shared pointer to client context to QueryResult

evertlammerts:
This PR replaces the `optional_ptr<ClientContext>` in `ClientProperties` with a proper `shared_ptr<ClientContext>` on `QueryResult`. The real change here is that the lifetime of a client context object is now tied to that of result objects. I.e. a connection stays open as long a a result object exists. This was already the case for `StreamQueryResult`, which held a shared pointer to the client context.

This is triggered by #21927 / #22007, which started with duckdb/duckdb-spatial#788. In short: the GeoArrow extension's `PopulateSchema` callback needs `ClientContext::ParseLogicalType` to bind the CRS type, which requires a valid client context. When Arrow results are consumed lazily, e.g. via the Arrow C Stream API, ADBC, the Python client, or any callback-driven consumer, then the original context may have been closed, and `PopulateSchema` fails with: 

> INTERNAL Error: TransactionContext::ActiveTransaction called without active transaction

~The underlying issue is that Arrow extensions are registered in the catalog, and the catalog is queried whenever an Arrow schema is created or record batches are created.~

Edit: The underlying issue is that Arrow type extensions (which may be implemented by duckdb extension authors) implement callbacks that in their signature promise to provide a valid client context.

As a first fix I've patched duckdb-python in duckdb/duckdb-python#423 by caching the Arrow schema eagerly so it doesn't need to be regenerated after the connection has closed. But that's a workaround that only works in Python and for the specific arrow C stream issue (i.e. the `get_schema` callback through Python).

### Summary of changes
- `ClientProperties::client_context` is removed.
- The property `shared_ptr<ClientContext> client_context` is added to `QueryResult` and initialized in the constructors.
- The `MaterializedQueryResult`, `StreamQueryResult`, and `ArrowQueryResult` constructors are updated.
- The private `shared_ptr<ClientContext> context` member of `StreamQueryResult` is dropped in favor of the inherited context.
- `ArrowConverter::ToArrowSchema/ToArrowArray`, `ArrowUtil::TryFetchChunk/FetchChunk` take `ClientContext &`, and `ArrowAppender` takes a `shared_ptr<ClientContext>`.
- All collectors (`PhysicalMaterializedCollector`, `PhysicalBatchCollector`, `PhysicalBufferedCollector`, `PhysicalBufferedBatchCollector`, `PhysicalArrowCollector`, `PhysicalArrowBatchCollector`) pass context.shared_from_this() when constructing results.
- C-API: `CClientArrowOptionsWrapper` now wraps the context. The consuming functions use this instead of the options.
  - Tests updated to match the new signatures.

### DIFF
--- a/src/common/arrow/arrow_appender.cpp
+++ b/src/common/arrow/arrow_appender.cpp
@@ -8,6 +8,7 @@
 #include "duckdb/common/arrow/appender/append_data.hpp"
 #include "duckdb/common/arrow/appender/list.hpp"
 #include "duckdb/function/table/arrow/arrow_duck_schema.hpp"
+#include "duckdb/main/client_context.hpp"
 
 namespace duckdb {
 
@@ -15,9 +16,11 @@ namespace duckdb {
 // ArrowAppender
 //===--------------------------------------------------------------------===//
 
-ArrowAppender::ArrowAppender(vector<LogicalType> types_p, const idx_t initial_capacity, ClientProperties options,
+ArrowAppender::ArrowAppender(vector<LogicalType> types_p, const idx_t initial_capacity,
+                             const shared_ptr<ClientContext> &client_context,
                              unordered_map<idx_t, const shared_ptr<ArrowTypeExtensionData>> extension_type_cast)
-    : types(std::move(types_p)), options(options) {
+    : types(std::move(types_p)), client_context(client_context) {
+	auto options = client_context->GetClientProperties();
 	for (idx_t i = 0; i < types.size(); i++) {
 		unique_ptr<ArrowAppendData> entry;
 		bool bitshift_boolean = types[i].id() == LogicalTypeId::BOOLEAN && !options.arrow_lossless_conversion;
@@ -40,8 +43,7 @@ void ArrowAppender::Append(DataChunk &input, const idx_t from, const idx_t to, c
 	for (idx_t i = 0; i < input.ColumnCount(); i++) {
 		if (root_data[i]->extension_data && root_data[i]->extension_data->duckdb_to_arrow) {
 			Vector input_data(root_data[i]->extension_data->GetInternalType());
-			root_data[i]->extension_data->duckdb_to_arrow(*options.client_context, input.data[i], input_data,
-			                                              input_size);
+			root_data[i]->extension_data->duckdb_to_arrow(*client_context, input.data[i], input_data, input_size);
 			root_data[i]->append_vector(*root_data[i], input_data, from, to, input_size);
 		} else {
 			root_data[i]->append_vector(*root_data[i], input.data[i], from, to, input_size);
@@ -104,6 +106,7 @@ ArrowArray *ArrowAppender::FinalizeChild(const LogicalType &type, unique_ptr<Arr
 //! Returns the underlying arrow array
 ArrowArray ArrowAppender::Finalize() {
 	D_ASSERT(root_data.size() == types.size());
+	auto options = client_context->GetClientProperties();
 	auto root_holder = make_uniq<ArrowAppendData>(options);
 
 	ArrowArray result;
--- a/src/common/arrow/arrow_converter.cpp
+++ b/src/common/arrow/arrow_converter.cpp
@@ -4,10 +4,6 @@
 #include "duckdb/common/arrow/arrow_converter.hpp"
 #include "duckdb/common/exception.hpp"
 #include "duckdb/common/helper.hpp"
-#include "duckdb/common/types/interval.hpp"
-#include "duckdb/common/types/sel_cache.hpp"
-#include "duckdb/common/types/vector_cache.hpp"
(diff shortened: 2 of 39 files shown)


Mytherin:
Can't we do a `weak_ptr`?

`shared_ptr` has given us issues in the past - namely with connections being kept open unexpectedly. I think this change can cause problems for various clients. Especially for a bug-fix release targeting change that seems like not a great idea. 

We could consider it for `main` but I would still prefer `weak_ptr` in user-facing  objects. We've been trying to move away from `shared_ptr` in general for user-facing objects for those reasons (hard to reason about object lifetimes, confused users when `.Close()` does not actually close a database / connection, etc).

evertlammerts:
> Can't we do a `weak_ptr`?
> 
> `shared_ptr` has given us issues in the past - namely with connections being kept open unexpectedly. I think this change can cause problems for various clients. Especially for a bug-fix release targeting change that seems like not a great idea.
> 
> We could consider it for `main` but I would still prefer `weak_ptr` in user-facing objects. We've been trying to move away from `shared_ptr` in general for user-facing objects for those reasons (hard to reason about object lifetimes, confused users when `.Close()` does not actually close a database / connection, etc).

Fair enough, I don't have a strong preference on the semantics. For the record: `StreamQueryResult` already carries a shared pointer to a client context (understandably). I suspect that lazy generation (`StreamQueryResult` but also arrow streams / arrow record batches) makes it harder to enforce a convention for decoupling connection lifetimes from results. An alternative could be to give a different meaning to calling `Close()` on a connection / db vs letting a pointer to a connection / db go out of scope. If you have a function where you open a connection (and you get a `shared_ptr<ClientContext>`), run a query and return only the result object - meaning you let the client context go out of scope without having called `Close()` anywhere - then it's not unreasonable that the connection is still alive because it's referenced in the result, I think. Maybe `Close()` could propagate to all depending resources / objects. But that's too much for this PR. (Maybe we can discuss this offline in the context of the C API as well, cc @Maxxen.)

### For now...

The concrete problem with a `weak_ptr` right now is lazy result consumption with the Arrow C Stream API. In our implementation, an `ArrowArrayStream` (which implements the Arrow C Stream API) wraps a result object of any type (`MaterializedQueryResult`, `StreamQueryResult`, `ArrowQueryResult`). Two functions on these stream objects need a client context:
* `get_schema` (implemented in `ResultArrowArrayStreamWrapper::MyStreamGetSchema`)
* `get_next` (implemented in `ResultArrowArrayStreamWrapper::MyStreamGetNext`)

In both cases the functions might have to call functions on registered arrow type extensions (which can be registered by duckdb extensions as well). These might or might not provide one of the following callbacks:
* `typedef void (*cast_arrow_duck_t)(ClientContext &context, Vector &source, Vector &result, idx_t count);`
* `typedef void (*cast_duck_arrow_t)(ClientContext &context, Vector &source, Vector &result, idx_t count);`
* `typedef void (*populate_arrow_schema_t)(DuckDBArrowSchemaHolder &root_holder, ArrowSchema &child, const LogicalType &type, ClientContext &context, const ArrowTypeExtension &extension);`
* `typedef unique_ptr<ArrowType> (*get_type_t)(ClientContext &context, const ArrowSchema &schema, const ArrowSchemaMetadata &schema_metadata);`

A few [internal arrow extension](https://github.com/duckdb/duckdb/blob/b6e2d6d27cd1234215dc91bef130c65437d4e0a2/src/common/arrow/arrow_type_extension.cpp#L549) implement one or more of these callbacks, and the `geoarrow.wkb` extension is the first (of the internal extensions) to actually use the client context for a catalog lookup.

### Options

For v1.5 I suggest we leave things as they are. Apart from the spatial issue I haven't seen any issues pop up, and the workaround for the spatial issue works.

For v2.0 we should give some attention to the arrow type extension hooks and the arrow c stream api. We could also consider to formalize semantics around connection closing for lazy result production paths.

Mytherin:
I would prefer to also remove the `shared_ptr` in the streaming result and turn it into a `weak_ptr`. Shared pointers between these structures have just been a headache in the past and I would prefer to remove them rather than add more of them.

> Maybe Close() could propagate to all depending resources / objects. 

We also tried this before but it's also a lifetime nightmare compared to just using `weak_ptr` where methods that need the underlying object to be alive just fail if it has since been destroyed.

> The concrete problem with a weak_ptr right now is lazy result consumption with the Arrow C Stream API. In our implementation, an ArrowArrayStream (which implements the Arrow C Stream API) wraps a result object of any type (MaterializedQueryResult, StreamQueryResult, ArrowQueryResult). Two functions on these stream objects need a client context:


If that's all then maybe `ResultArrowArrayStreamWrapper` could have the `shared_ptr<ClientContext>` then? The scope of that change is much smaller than `QueryResult`. I would still prefer to have that be a `weak_ptr` eventually but we can do that in the new C API.

PR #22516, opened 2026-05-07
Patch httplib by making ThreadPool constructor more solid on pthread_create failures

carlopi:
Behaviour should be unchanged in the happy case, but somewhat more solid in the bad one.


carlopi:
I pinged the httplib's maintainers at https://github.com/yhirose/cpp-httplib/issues/2444, since it might be potentially also worth upstreaming there (if this makes sense at all).

Mytherin:
Thanks!

PR #22517, opened 2026-05-07
Add RESULT_COLLECTOR profiling metric

krleonid:
## Summary
- Adds a new `RESULT_COLLECTOR` phase timing metric that measures time spent in `executor.GetResult()` during result materialization
- Helps diagnose latency gaps between DuckDB's reported query execution time and actual wall-clock time (e.g., SELECT * with many columns showing ~3s wall time but only ~3ms CPU time)
- Includes dedicated test covering metric presence, root-only scoping, and disable behavior

## Test plan
- [x] `test_custom_profiling_result_collector.test` passes (30 assertions)
- [x] Generated enum files via `scripts/generate_metric_enums.py` (72 metrics, 10 phase_timing)
- [x] Full build passes without errors


Tishj:
Can we check how this behaves with a `StreamQueryResult` ?
This will likely need a c++ test, because I don't think we can do that from pure SQL.

The StreamQueryResult is a bit of a special case, because it creates the `QueryResult` before the final pipeline has finished.

See:
```c++
	//! Issues a query to the database and returns a QueryResult. This result can be either a StreamQueryResult or a
	//! MaterializedQueryResult. The result can be stepped through with calls to Fetch(). Note that there can only be
	//! one active StreamQueryResult per Connection object. Calling SendQuery() will invalidate any previously existing
	//! StreamQueryResult.
	DUCKDB_API unique_ptr<QueryResult>
	SendQuery(const string &query, QueryParameters query_parameters = QueryResultOutputType::ALLOW_STREAMING);
```

src/main/client_context.cpp
```diff
@@ -397,9 +397,14 @@ unique_ptr<QueryResult> ClientContext::FetchResultInternal(ClientContextLock &lo
 	unique_ptr<QueryResult> result;
 	D_ASSERT(executor.HasResultCollector());
 	// we have a result collector - fetch the result directly from the result collector
+	auto &profiler = QueryProfiler::Get(*this);
+	profiler.StartPhase(MetricType::RESULT_COLLECTOR);
 	result = executor.GetResult();
+	profiler.EndPhase();
```
lnkuiper: I don't think this is measuring all result collector activity. It might work somewhat for streaming query results, but definitely not for materialized query results.

I've tried this:
```sql
PRAGMA enable_profiling = 'json';
PRAGMA profiling_output = 'profiling_output.json';
PRAGMA custom_profiling_settings='{"RESULT_COLLECTOR": "true"}';
SET threads=1;
FROM tpch_sf10.lineitem;
```
The query takes 5.43s, of which 1.66s is `TABLE_SCAN`, but result collector takes `1.25e-7` seconds. This does not explain the almost 4 second gap.

Our result collectors like `PhysicalMaterializedCollector` are just operators that support the `Sink` interface. I think this should already be measured in `PipelineExecutor`, but they aren't showing up in profiling because `OperatorRequiresProfiling()` does not include `PhysicalOperatorType::RESULT_COLLECTOR`.

I think we could consider just adding result collector to the white list.

lnkuiper (changes requested):
Thanks for the PR! I don't think this is correctly measuring:

PR #22518, opened 2026-05-07
Avoid 3 instances of idx_t - idx_t > 0, and avoid unnecessary check on zLine

carlopi:
Cleanups some warnings raised by tooling. I tested the checksum, has no effect on compatibility, since previous code was already correct (due to the presence of iteration). Nevertheless, the more formally correct form is better, clearer and robust to code changes.


lnkuiper:
Thanks! Meant to add to merge queue but just noticed this was v1.5 and it just went in immediately 😅 . Should be fine, looked good.

PR #22525, opened 2026-05-08
Optimizer: Rewrite contains [large list] to a hash join

DinosL:
`col IN [1, 2, ..., N]` is transformed to `contains(list_value(1,...,N), col)`, which bypassed the `InClauseRewriter` and was pushed into the scan as a filter that runs in O(N).

This PR adds an optimizer rule that rewrites `contains` to `COMPARE_IN` before filter pushdown runs. Adds a guard to prevent `COMPARE_IN` with large lists from being pushed into the scan, so InClauseRewriter can convert them to a hash join.

I think the same can be done for `NOT(COMPARE_IN)`, but I will do this at a later time.

## Benchmarks

Running a `FROM bench_table WHERE val in [large_list]` query on a table with 100 million rows

Before
```
benchmark/micro/optimizer/contains_to_in_rewriter.benchmark     1       3.286797
benchmark/micro/optimizer/contains_to_in_rewriter.benchmark     2       3.152438
benchmark/micro/optimizer/contains_to_in_rewriter.benchmark     3       3.466974
benchmark/micro/optimizer/contains_to_in_rewriter.benchmark     4       3.273688
benchmark/micro/optimizer/contains_to_in_rewriter.benchmark     5       3.240413
```
After
```
benchmark/micro/optimizer/contains_to_in_rewrite.benchmark      1       0.071765
benchmark/micro/optimizer/contains_to_in_rewrite.benchmark      2       0.069603
benchmark/micro/optimizer/contains_to_in_rewrite.benchmark      3       0.069411
benchmark/micro/optimizer/contains_to_in_rewrite.benchmark      4       0.075491
benchmark/micro/optimizer/contains_to_in_rewrite.benchmark      5       0.070934
```


src/optimizer/rule/contains_to_in_clause.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "duckdb/optimizer/expression_rewriter.hpp"
+#include "duckdb/planner/expression/bound_constant_expression.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/planner/expression/bound_operator_expression.hpp"
+
+namespace duckdb {
+
+// ContainsToInClauseRule::ContainsToInClauseRule(ExpressionRewriter &rewriter) : Rule(rewriter) {
+// 	root = make_uniq<ExpressionMatcher>(ExpressionClass::BOUND_FUNCTION);
+// }
```
lnkuiper: Can this comment be removed?

src/optimizer/filter_combiner.cpp
```diff
@@ -537,6 +537,12 @@ FilterPushdownResult FilterCombiner::TryPushdownInFilter(TableFilterSet &table_f
 		return FilterPushdownResult::PUSHED_DOWN_FULLY;
 	}
 
+	// Lists large enough to benefit from a hash join should not be pushed into the
+	// scan. Use same threshold as InClauseRewriter
+	if (func.children.size() - 1 >= 6) {
```
lnkuiper: Maybe we can add a `constexpr` in the `InclauseRewriter` header and refer to that here and in `in_clause_rewriter.cpp`? Then we don't hardcode 6 in multiple places. I also think this is off-by-one vs. what `InClauseRewriter` does.

src/optimizer/filter_combiner.cpp
```diff
@@ -537,6 +537,12 @@ FilterPushdownResult FilterCombiner::TryPushdownInFilter(TableFilterSet &table_f
...
+		return FilterPushdownResult::NO_PUSHDOWN;
+	}
+
 	//! Check if values are consecutive, if yes transform them to >= <= (only for integers)
 	// e.g. if we have x IN (1, 2, 3, 4, 5) we transform this into x >= 1 AND x <= 5
```
lnkuiper, line 518: The early-out above now happens before we check `IsDenseRange` below, so I think this means we are losing the optimization where we rewrite to `x >= min AND x <= max` and push down fully. You might need to change your test which uses a dense range once you fix this.

lnkuiper (changes requested):
Thanks for the PR! Looks good, but I think we need to do a few fixes before it's ready to merge:

src/optimizer/filter_combiner.cpp
DinosL: Maybe I am missing something here, but for a dense range with a large enough list, transforming the `IN` to a hash join is faster than pushing `x >= min AND x <= max` into the scan. There could be side effects I'm not aware of (e.g. other optimizers relying on the push down etc.).

lnkuiper: That seems unlikely. I would expect that, for example, doing `min >= 1 AND max <= 1000` is faster than joining with 1000 values.

lnkuiper (approved):

lnkuiper:
`unittest --test-config test/configs/verify_statement_prepare.json test/sql/function/generic/test_in.test` seems to be failing

DinosL:
Added a check for empty list. Should be ok now!

PR #22537, opened 2026-05-09
Keep stats at unsigned integer cast

dentiny:
Hi team, I found for unsigned integer cast, stats are not applied as expected, which leads to unnecessary table scan.
```sql
memory D CREATE TABLE t_signed AS SELECT i::INT AS id FROM range(0, 100) t(i);
memory D EXPLAIN SELECT * FROM t_signed WHERE id::BIGINT > 100;

┌───────────────────────────┐
│        EMPTY_RESULT       │
└───────────────────────────┘
memory D CREATE TABLE t_unsigned AS SELECT i::UINTEGER AS id FROM range(0, 100) t(i);
memory D EXPLAIN SELECT * FROM t_unsigned WHERE id::BIGINT > 100;

┌───────────────────────────┐
│          SEQ_SCAN         │
│    ────────────────────   │
│           Table:          │
│   memory.main.t_unsigned  │
│                           │
│   Type: Sequential Scan   │
│      Projections: id      │
│                           │
│          Filters:         │
│ (CAST(id AS BIGINT) > 100)│
│                           │
│          ~20 rows         │
└───────────────────────────┘
```
In the above example, table `t_unsigned`'s min/max stats should be kept IMO, which could avoid the scan.

### DIFF
--- a/src/optimizer/statistics/expression/propagate_cast.cpp
+++ b/src/optimizer/statistics/expression/propagate_cast.cpp
@@ -38,6 +38,11 @@ bool StatisticsPropagator::CanPropagateCast(const LogicalType &source, const Log
 	case PhysicalType::INT32:
 	case PhysicalType::INT64:
 	case PhysicalType::INT128:
+	case PhysicalType::UINT8:
+	case PhysicalType::UINT16:
+	case PhysicalType::UINT32:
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #22538, opened 2026-05-09
optimizer: don't return truncated VARCHAR MIN/MAX from statistics

### LINKED ISSUE #22515: MIN(varchar) and MAX(varchar) over read_parquet() returns string truncated to 8 bytes (regressed in 1.5.0, still present in 1.5.3.dev7)
### What happens?

When aggregating with `MIN()` or `MAX()` on a `VARCHAR` column read directly via `read_parquet()`, the result is silently truncated to its first 8 bytes for any input string longer than 8 characters. The parquet column statistics themselves are correct; the truncation happens in the optimizer's statistics-propagation path. Disabling that optimizer (`SET disabled_optimizers = 'statistics_propagation'`) restores correct behavior, which confirms the path.

This is a silent correctness bug: the query succeeds and returns a wrong value. We hit it in query where a downstream `::DATE` cast on the truncated value happened to fail and surfaced the truncation; otherwise we would have produced wrong values without noticing.

### To Reproduce

```
import duckdb
con = duckdb.connect(":memory:")
con.execute("""
    COPY (SELECT 'A123456789' AS v) TO '/tmp/bug.parquet'
    (FORMAT PARQUET, USE_TMP_FILE FALSE)
""")

# BUG: returns 'A1234567' (8 bytes) on 1.5.x; should return 'A123456789'
print(con.execute("SELECT MIN(v) FROM read_parquet('/tmp/bug.parquet')").fetchone())

# Workaround: disabling the optimizer restores correctness
con.execute("SET disabled_optimizers = 'statistics_propagation'")
print(con.execute("SELECT MIN(v) FROM read_parquet('/tmp/bug.parquet')").fetchone())
```

Output on 1.5.2 / 1.5.3.dev7:

```
('A1234567',)        # ← truncated to 8 bytes (BUG)
('A123456789',)      # ← correct, after disabling optimizer
```

Output on 1.4.1 / 1.4.4:

```
('A123456789',)      # correct
('A123456789',)      # correct
```

## Scope (probed empirically on 1.5.2)

| Configuration | Result |
|---|---|
| String length ≤ 8 | Correct |
| String length ≥ 9 | Truncated to 8 bytes |
| Row count (1, 10, 100, 1000, 5000) | All affected — even 1 row triggers it |
| `MIN(v)` on `read_parquet(...)` | **BUG** |
| `MAX(v)` on `read_parquet(...)` | **BUG** |
| `FIRST(v)` on `read_parquet(...)` | OK |
| `ARG_MAX(v, v)` on `read_parquet(...)` | OK |
| `SELECT v FROM read_parquet(...)` (no aggregate) | OK |
| `MIN(v)` on a regular DuckDB table | OK |
| `MIN(v)` on `VALUES` | OK |
| `MIN(v)` on a CTAS-loaded copy of the parquet | OK |
| With `SET disabled_optimizers = 'statistics_propagation'` | OK |

`EXPLAIN` for the buggy query shows `COLUMN_DATA_SCAN ~1 row` with no aggregate — the optimizer fully eliminated the query and is returning a literal value derived from the parquet column statistics. The root cause appears to be in that derivation path, where it's reading the inline-string prefix portion of the `VARCHAR` representation (the 8-byte fast-comparison prefix) rather than materializing the full string.

## Versions

| Version | `MIN('A123456789')` on parquet |
|---|---|
| 1.4.4 (latest LTS) | `'A123456789'` ✅ |
| 1.5.0 | `'A1234567'` ❌ regression introduced |
| 1.5.2 | `'A1234567'` ❌ |
| 1.5.3.dev7 | `'A1234567'` ❌ still broken |

### OS:

Reproduced on macOS arm64 (Darwin 24.5.0) with Python 3.12.4, but the bug is in the optimizer's statistics-propagation path, not platform-specific

### DuckDB Version:

1.5.0 introduced; latest 1.5.3.dev7 still affected

### DuckDB Client:

Python (duckdb PyPI package), but the bug is in the engine, should reproduce identically via the CLI

### Hardware:

_No response_

### Full Name:

Misael Ferreira

### Affiliation:

Civitas Learning

- [x] Yes, I have

- [x] Yes, I have

nahomar:
Fixes #22515.

### Problem

```python
con.execute("COPY (SELECT 'A123456789' AS v) TO '/tmp/bug.parquet' (FORMAT PARQUET)")
con.execute("SELECT MIN(v) FROM read_parquet('/tmp/bug.parquet')").fetchone()
# -> ('A1234567',)   # truncated to 8 bytes
```

The optimizer takes a shortcut: when partition statistics report `MinMaxIsExact`, `StatisticsPropagator::TryGetValueFromStats` returns the stats' min/max directly as the result of `MIN()` / `MAX()`. The Parquet reader's `MinMaxIsExact` (`extension/parquet/parquet_reader.cpp:1434`) trusts the file-level `is_min_value_exact` / `is_max_value_exact` flags introduced in PARQUET-2352.

But DuckDB's `StringStatsData` stores only the first 8 bytes of each min/max:

```cpp
constexpr static uint32_t MAX_STRING_MINMAX_SIZE = 8;
data_t min[MAX_STRING_MINMAX_SIZE];
data_t max[MAX_STRING_MINMAX_SIZE];
```

So even when Parquet says "the bounds are exact", once the bytes flow through our stats they're truncated, and the shortcut returns the truncated prefix as the user-visible result.

This is a regression in 1.5.x — the workaround documented in the issue is `SET disabled_optimizers = 'statistics_propagation'`.

### Fix

In `TryGetValueFromStats`, after `MinMaxIsExact` returns true, additionally require that the column's known max string length does **not** exceed the prefix size. If it does (or if the max length is unknown), the stats may be a truncated prefix and we can't use them as the aggregate result; the optimizer falls back to scanning the column, which produces the correct value.

```cpp
if (!StringStats::HasMaxStringLength(*column_stats) ||
    StringStats::MaxStringLength(*column_stats) > StringStatsData::MAX_STRING_MINMAX_SIZE) {
    return false;
}
```

The row-group implementation of `MinMaxIsExact` (in `src/storage/table/row_group.cpp`) already encodes the same invariant (`max_length == StringStats::Max(stats).length()`), so this is effectively making the consumer side enforce it regardless of which `PartitionStatistics` source provided the stats.

### Tests

Added `test/parquet/min_max_long_varchar_22515.test` covering:
- `MIN`/`MAX` over Parquet of strings >8 bytes (the bug)
- `MIN`/`MAX` over Parquet of strings ≤8 bytes (still optimized via stats)
- Mixed short + long values

All `[parquet]` tests pass locally (23080 assertions, 162 cases).

### DIFF
--- a/src/optimizer/statistics/operator/propagate_aggregate.cpp
+++ b/src/optimizer/statistics/operator/propagate_aggregate.cpp
@@ -91,14 +91,8 @@ bool TryGetValueFromStats(const PartitionStatistics &stats, const StorageIndex &
 			// No min/max statistics availabe
 			return false;
 		}
-		// String statistics store at most an 8-byte prefix of the min/max
-		// values (see StringStatsData::MAX_STRING_MINMAX_SIZE). If the
-		// actual maximum string length exceeds that prefix, the stored
-		// prefix is not the true min/max even when the underlying source
-		// (e.g. Parquet `is_min_value_exact` / `is_max_value_exact`)
-		// reports the bounds as exact. Returning the prefix in that case
-		// silently truncates MIN()/MAX() results for VARCHAR aggregates
-		// over Parquet (issue #22515).
+		// String statistics store at most an 8-byte prefix of the min/max values.
+		// If the actual maximum string length exceeds that prefix, the stored prefix is not the true min/max
 		if (!StringStats::HasMaxStringLength(*column_stats) ||
 		    StringStats::MaxStringLength(*column_stats) > StringStatsData::MAX_STRING_MINMAX_SIZE) {
 			return false;
--- a/src/optimizer/statistics/operator/propagate_aggregate.cpp
+++ b/src/optimizer/statistics/operator/propagate_aggregate.cpp
@@ -91,6 +91,18 @@ bool TryGetValueFromStats(const PartitionStatistics &stats, const StorageIndex &
 			// No min/max statistics availabe
 			return false;
 		}
+		// String statistics store at most an 8-byte prefix of the min/max
+		// values (see StringStatsData::MAX_STRING_MINMAX_SIZE). If the
+		// actual maximum string length exceeds that prefix, the stored
+		// prefix is not the true min/max even when the underlying source
+		// (e.g. Parquet `is_min_value_exact` / `is_max_value_exact`)
+		// reports the bounds as exact. Returning the prefix in that case
+		// silently truncates MIN()/MAX() results for VARCHAR aggregates
+		// over Parquet (issue #22515).
+		if (!StringStats::HasMaxStringLength(*column_stats) ||
+		    StringStats::MaxStringLength(*column_stats) > StringStatsData::MAX_STRING_MINMAX_SIZE) {
+			return false;
+		}
 	}
 	result = comparator.GetVal(*column_stats);
 	return true;
--- a/test/parquet/min_max_long_varchar_22515.test
+++ b/test/parquet/min_max_long_varchar_22515.test
@@ -1,0 +1,47 @@
+# name: test/parquet/min_max_long_varchar_22515.test
+# description: Issue #22515 - MIN/MAX(VARCHAR) over read_parquet() must not be truncated to 8 bytes by statistics propagation.
+# group: [parquet]
+
+require parquet
+
+# Strings longer than the 8-byte StringStats prefix.
+statement ok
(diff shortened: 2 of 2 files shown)


d-justen (approved):
Thanks for picking this up. This looks good to me once CI has run through.

d-justen:
Could you please retarget this PR to `v1.5-variegata`?

pdet:
Thanks!

Issue #22487, opened 2026-05-05
INSTALL extension ignores OS `http_proxy` / `HTTP_PROXY` environment variables


DaiZack:
### What happens?

# `INSTALL` extension ignores OS `http_proxy` / `HTTP_PROXY` environment variables

## What happens?

DuckDB's `INSTALL` command does not respect the standard OS-level `http_proxy` / `HTTP_PROXY` environment variables. Behind a corporate proxy, `INSTALL <extension>` hangs indefinitely unless `SET http_proxy` is explicitly called beforehand.

```

DuckDB v1.5.2 (Variegata)
Enter ".help" for usage hints.
memory D install httpfs;
IO Error:
Failed to download extension "httpfs" at URL "http://extensions.duckdb.org/v1.5.2/linux_amd64/httpfs.duckdb_extension.gz"
Extension "httpfs" is an existing extension.

For more info, visit https://duckdb.org/docs/stable/extensions/troubleshooting?version=v1.5.2&platform=linux_amd64&extension=httpfs (ERROR Connection timed out)
```

## To Reproduce

```bash
export http_proxy="http://my-proxy.example.com:8080"
export https_proxy="http://my-proxy.example.com:8080"

# CLI - hangs forever
duckdb -c "INSTALL httpfs;"

# Python - also hangs
python3 -c "import duckdb; duckdb.connect().execute('INSTALL httpfs')"
```

## Expected behavior

DuckDB should automatically read `http_proxy` / `HTTP_PROXY` (and `https_proxy` / `HTTPS_PROXY`) environment variables when performing HTTP requests for extension downloads, similar to how curl, wget, pip, and most other tools respect these standard variables.

## Workaround

Explicitly set the proxy before installing:

```sql
SET http_proxy = 'http://my-proxy.example.com:8080';
INSTALL httpfs;
```

For the CLI, this can be persisted in `~/.duckdbrc`:
```
SET http_proxy='http://my-proxy.example.com:8080';
```

For Python, a `sitecustomize.py` monkey-patch can auto-set it:
```python
import builtins, os

_proxy = os.environ.get("http_proxy") or os.environ.get("HTTP_PROXY", "")
if _proxy:
    _orig_import = builtins.__import__
    _patched = False

    def _import_hook(name, *args, **kwargs):
        global _patched
        mod = _orig_import(name, *args, **kwargs)
        if name == "duckdb" and not _patched:
            _patched = True
            _orig_connect = mod.connect
[...]

PR #22541, opened 2026-05-10
Move http_proxy setting to global setting, and use GetEnvVariable('HTTP_PROXY') as default

carlopi:
This makes so that:
`HTTP_PROXY=abc duckdb -c "SELECT current_setting('http_proxy');"` gives `abc`, and more importantly that extensions installs will also pass through the proxy. Note, that since the introduction of curl, curl automatically uses HTTP_PROXY and HTTPS_PROXY, so now implicitly also DuckDB handles those parameters (if `httpfs` is loaded AND curl, the default, enabled).

I think it makes sense to adapt also the INSTALL path, let's enforce the commonly used convention of `HTTP_PROXY` being respected.

Fixes https://github.com/duckdb/duckdb/issues/22487.

Only issue is with this being a patch release. This is inconsistent now, I am not sure if inconsistency or change is worse. I would probably opt for fixing inconsistency.

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -672,11 +672,14 @@
     },
     {
         "name": "http_proxy",
-        "description": "HTTP proxy host",
+        "description": "HTTP proxy host (defaults to the HTTP_PROXY environment variable when unset)",
         "type": "VARCHAR",
         "scope": "global",
-        "default_value": "",
-        "struct": "HTTPProxySetting"
+        "struct": "HTTPProxySetting",
+        "custom_implementation": [
+            "set",
+            "reset"
+        ]
     },
     {
         "name": "http_proxy_password",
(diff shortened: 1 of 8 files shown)


krlmlr:
Thanks, nice! I've seen two independent reports for this issue. Do you think this can be extended further:

- Also query `$HTTPS_PROXY`, let it take precedence
- Parse username and password from the URL and set the corresponding DuckDB options

Happy to contribute.

carlopi:
Yes, I think it would be interesting to expand further, `HTTPS_PROXY` is sort of handled automagically through curl, although that's not visible to duckdb, so could be better integrated. Scope of this PR was limited to `HTTP_PROXY` since that's the one that control extensions installs.

On parsing username/password from URL, that would also be cool, but looks independent.

Both to me seems current gaps, that could indeed be fixed, contribution is very welcome!

krlmlr:
I have an implementation in https://github.com/krlmlr/duckdb/pull/3 that looks solid to me but does not satisfy the "no gen AI" policy. Would that be acceptable as squashed commit if I go over the diff and hand-tweak where appropriate?

Issue #22539, opened 2026-05-10
Cannot write small decimal variants


dentiny:
### What happens?

DuckDB fails to write variant value with small decimal value; see my repro sql, should be straightforward.

### To Reproduce

```sql
memory D CREATE TABLE t (data VARIANT);
memory D INSERT INTO t VALUES (1.23::DECIMAL(3, 2)::VARIANT);
memory D INSERT INTO t VALUES (12.345::DECIMAL(5, 3)::VARIANT);
memory D COPY t TO '/tmp/small_variant.parquet';
Invalid Input Error:
Can't convert VARIANT DECIMAL(3, 2) to Parquet VARIANT
```

### OS:

macos

### DuckDB Version:

ae0aec232abb91f18667cc0a6af82a34dfc20bcf

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

PR #22544, opened 2026-05-10
Fix variant write small decimal

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22539

From the error message, it's easy to trace to https://github.com/duckdb/duckdb/blob/01eda16d6e3a6648accda27230b9e30edd8b9703/extension/parquet/writer/variant/convert_variant.cpp#L568-L570

I think it doesn't make too much sense to disallow small decimal; considering parquet only allows decimal4/8/16, so I promote small decimals to big one and store.
Ref: https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#encoding-types

I have ran CI in my fork: https://github.com/dentiny/duckdb/pull/97 (failure unrelated)


Tishj:
Yep sounds correct 👍

Mytherin:
Thanks!

PR #22547, opened 2026-05-10
Fix parquet metadata cache validation

dentiny:
I wrote a silly bug :(


Mytherin:
Thanks! No worries, happens to everyone :)

PR #22552, opened 2026-05-11
Fix MVCC visibility during unique index creation after DELETE

### LINKED ISSUE #22420: CREATE UNIQUE INDEX Ignores Prior DELETE Within the Same Transaction
### What happens?

`CREATE UNIQUE INDEX` within a transaction raises `ConstraintException: Data contains duplicates on indexed column(s)` even though a prior `DELETE` in the same transaction has already removed the duplicate rows. `SELECT` queries within the transaction correctly show zero duplicates, but the unique index build sees the pre-delete state.

This regressed in v1.5.0. It works correctly in v1.4.1.

### To Reproduce

```python
import duckdb
import tempfile
from pathlib import Path

db = Path(tempfile.mkdtemp()) / "bug.duckdb"
conn = duckdb.connect(str(db))

conn.execute("CREATE TABLE t (id INT, x INT)")
conn.execute("INSERT INTO t VALUES (1, 10), (2, 10)")

conn.execute("BEGIN TRANSACTION")
conn.execute("DELETE FROM t WHERE id = 1")

# Verify: only 1 row remains, no duplicate groups
assert conn.execute("SELECT COUNT(*) FROM t").fetchone() == (1,)
assert conn.execute(
    "SELECT x, COUNT(*) FROM t GROUP BY x HAVING COUNT(*) > 1"
).fetchall() == []

# Should succeed (only 1 row, no duplicates) — but fails in 1.5.x
conn.execute("CREATE UNIQUE INDEX u ON t(x)")
conn.execute("COMMIT")
```

```text
# DuckDB 1.4.1 — correct behavior
Traceback (most recent call last):  # no error, COMMIT succeeds

# DuckDB 1.5.2 — regression
  File "repro.py", line 19, in <module>
    conn.execute("CREATE UNIQUE INDEX u ON t(x)")
_duckdb.ConstraintException: Constraint Error: Data contains duplicates
on indexed column(s)
```

**Also reproducible via pure SQL:**

```sql
CREATE TABLE t (id INT, x INT);
INSERT INTO t VALUES (1, 10), (2, 10);
BEGIN TRANSACTION;
DELETE FROM t WHERE id = 1;
-- SELECT COUNT(*) FROM t;  → 1
-- SELECT x, COUNT(*) FROM t GROUP BY x HAVING COUNT(*) > 1;  → empty
CREATE UNIQUE INDEX u ON t(x);
-- ERROR: Constraint Error: Data contains duplicates on indexed column(s)
```

**Expected behavior:** `CREATE UNIQUE INDEX` should succeed. The DELETE has removed the duplicate; only one row with `x=10` remains.

**Additional findings (DuckDB 1.5.2):**

| Operation in tx after DELETE | Result |
|-----------------------------|--------|
| `CREATE INDEX` (non-unique) | succeeds |
| `CREATE UNIQUE INDEX` | **fails** (ConstraintError) |
| `ALTER TABLE ADD PRIMARY KEY` | **fails** (same error) |
| DELETE outside tx, then `CREATE UNIQUE INDEX` in a separate tx | succeeds |

The bug is specific to **unique constraint validation** during index creation inside the same transaction that performed the DELETE. Non-unique indexes are unaffected.

**Suspected cause:** The v1.5.0 "Buffer Managed Indexes Part 4" (PR #19303 — segment handles for ART base leaf) combined with the MVCC `old_transactions` removal (PR #19381) broke the contract from the v0.x fix in PR #4416 ("no longer disallows updates when committed data is being removed from indexes").

### OS:

macOS 15.4 (aarch64) — also reproduces on Linux x86_64 via CI

### DuckDB Version:

1.5.2 (also present in 1.5.0, 1.5.1; works correctly in 1.4.1)

### DuckDB Client:

Python 3.11 (also reproducible via DuckDB CLI)

### Hardware:

_No response_

### Full Name:

Ofri Wolfus

### Affiliation:

https://github.com/chunkhound/chunkhound

- [x] Yes, I have

- [x] Yes, I have

shivampr:
This change makes `CREATE UNIQUE INDEX` and `ALTER TABLE ADD PRIMARY KEY` respect rows deleted earlier in the same transaction during uniqueness validation.

### Testing

  - Added regression coverage for `CREATE UNIQUE INDEX` after same-transaction `DELETE`.
  - Added regression coverage for `ALTER TABLE ADD PRIMARY KEY` after same-transaction `DELETE`.
  - Verified both focused sqllogictests pass locally.

Fixes #22420

### DIFF
--- a/src/execution/index/art/art_index.cpp
+++ b/src/execution/index/art/art_index.cpp
@@ -12,6 +12,7 @@
 #include "duckdb/storage/data_table.hpp"
 #include "duckdb/storage/storage_manager.hpp"
 #include "duckdb/storage/table/append_state.hpp"
+#include "duckdb/transaction/duck_transaction.hpp"
 
 namespace duckdb {
 namespace {
@@ -22,6 +23,7 @@ namespace {
 class ARTBuildBindData : public IndexBuildBindData {
 public:
 	bool sorted = false;
+	bool has_deletes = false;
 };
 
 unique_ptr<IndexBuildBindData> ARTBuildBind(IndexBuildBindInput &input) {
@@ -32,6 +34,11 @@ unique_ptr<IndexBuildBindData> ARTBuildBind(IndexBuildBindInput &input) {
 	// We used to not sort for VARCHAR and multi-column indexes with the old sort implementation
 	// The new sorting implementation handles these cases much better and sorting improves performance now
 	bind_data->sorted = true;
+	if (input.info.constraint_type == IndexConstraintType::UNIQUE ||
+	    input.info.constraint_type == IndexConstraintType::PRIMARY) {
+		auto &transaction = DuckTransaction::Get(input.context, input.table.catalog);
+		bind_data->has_deletes = transaction.GetUndoProperties().has_deletes;
+	}
 
 	return std::move(bind_data);
 }
@@ -47,6 +54,8 @@ bool ARTBuildSort(IndexBuildSortInput &input) {
 class ARTBuildGlobalState : public IndexBuildGlobalState {
 public:
 	unique_ptr<BoundIndex> global_index;
+	unique_ptr<BoundIndex> delete_index;
+	bool has_delete_index = false;
 };
 
 unique_ptr<IndexBuildGlobalState> ARTBuildGlobalInit(IndexBuildInitGlobalStateInput &input) {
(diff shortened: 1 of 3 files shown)


artjomPlaunov:
Thanks again for your effort in this PR :-) as mentioned in the issue, however, we will be taking over this issue internally.

PR #22555, opened 2026-05-11
Auto-generate part of transformer (Part 1/N)

Dtenwolde:
Part of https://github.com/duckdblabs/duckdb-internal/issues/6380

The goal is to generate the parts of the `Transformer` that can be generated. Specifically, extracting the types from `ParseResult` at the correct indices. The script `gen_transformer_v2` (Will delete the other script and rename this one in the future), takes the parsed grammar files, adds the type information in `grammar_types.yml` and generated `TransformRuleInternal` methods. This will automate the extraction of manual retrieval of `ParseResult` with indices and make it easier for developers to write transformer functions. The script does not register generated functions yet (to be done). 

It currently generates only for the following grammar files: 
- `transaction.gram`
- `use.gram`
- `export.gram`
- `detach.gram`

Take `use.gram` as an example: 
```
UseStatement <- 'USE' UseTarget
UseTarget <- UseTargetCatalogSchema / SchemaName / CatalogName
UseTargetCatalogSchema <- CatalogName '.' ReservedSchemaName DotIdentifier*
DotIdentifier <- '.' Identifier
```

The script generates the following file: 
```cpp
// AUTO-GENERATED by scripts/parser/gen_transformer_v2.py -- DO NOT EDIT
#include "duckdb/parser/peg/transformer/peg_transformer.hpp"

namespace duckdb {

unique_ptr<SQLStatement> PEGTransformerFactory::TransformUseStatementInternal(PEGTransformer &transformer,
                                                                              ParseResult &parse_result) {
	auto &list_pr = parse_result.Cast<ListParseResult>();
	auto use_target = transformer.Transform<QualifiedName>(list_pr, 1);
	return TransformUseStatement(use_target);
}

QualifiedName PEGTransformerFactory::TransformUseTargetInternal(PEGTransformer &transformer,
                                                                ParseResult &parse_result) {
	auto &list_pr = parse_result.Cast<ListParseResult>();
	auto &choice_pr = list_pr.Child<ChoiceParseResult>(0);
	return TransformUseTarget(transformer, choice_pr.GetResult());
}

QualifiedName PEGTransformerFactory::TransformUseTargetCatalogSchemaInternal(PEGTransformer &transformer,
                                                                             ParseResult &parse_result) {
	auto &list_pr = parse_result.Cast<ListParseResult>();
	auto catalog_name = list_pr.Child<IdentifierParseResult>(0).identifier;
	auto reserved_schema_name = list_pr.Child<IdentifierParseResult>(2).identifier;
	auto &dot_identifier_opt = list_pr.Child<OptionalParseResult>(3);
	vector<string> dot_identifier;
	if (dot_identifier_opt.HasResult()) {
		auto &dot_identifier_repeat = dot_identifier_opt.GetResult().Cast<RepeatParseResult>();
		for (auto &dot_identifier_item : dot_identifier_repeat.GetChildren()) {
			dot_identifier.push_back(transformer.Transform<string>(dot_identifier_item));
		}
	}
	return TransformUseTargetCatalogSchema(catalog_name, reserved_schema_name, dot_identifier);
}

string PEGTransformerFactory::TransformDotIdentifierInternal(PEGTransformer &transformer, ParseResult &parse_result) {
	auto &list_pr = parse_result.Cast<ListParseResult>();
	auto identifier = list_pr.Child<IdentifierParseResult>(1).identifier;
	return TransformDotIdentifier(identifier);
}

} // namespace duckdb
```
 
The script can currently handle simple rule references, optional results, repeat, choices and simple macros. Nested macro (`Parens(List(Rule))`) is not yet supported. 

Some grammar rules provide no semantic value, for example since they only match a choice of keywords. Those grammar rules can be excluded from generation by adding them to the `excluded` section in `grammar_types.yml`

The script also generates declarations for the `Internal` and `user-implemented` stubs with the correct types: 
```cpp
static unique_ptr<SQLStatement> TransformUseStatementInternal(PEGTransformer &transformer, ParseResult &parse_result);
static unique_ptr<SQLStatement> TransformUseStatement(QualifiedName use_target);
```
This is included in the `PEGTransformerFactory`: https://github.com/Dtenwolde/duckdb/blob/0761edb240deb637cf747fbdeb746157c54d95a0/src/include/duckdb/parser/peg/transformer/peg_transformer.hpp#L1213

I added the generate script to `build_grammar.sh` together with a `make format-fix` since the generated code does not always pass the format checker, but perhaps this is not the cleanest. The generator script can be improved in this aspect

For future work: 
- Make the script work with nested macros
- Generate more complex grammar files / rewrite the user-implemented stubs
- Make script emit properly formatted code


src/parser/peg/transformer/generated/transform_export_generated.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+unique_ptr<SQLStatement> PEGTransformerFactory::TransformExportStatementInternal(PEGTransformer &transformer,
+                                                                                 ParseResult &parse_result) {
+	auto &list_pr = parse_result.Cast<ListParseResult>();
+	string export_source {};
+	transformer.TransformOptional(list_pr, 2, export_source);
+	auto string_literal = transformer.Transform<string>(list_pr, 3);
+	vector<GenericCopyOption> generic_copy_option_list {};
+	transformer.TransformOptional(list_pr, 4, generic_copy_option_list);
+	return TransformExportStatement(export_source, string_literal, generic_copy_option_list);
```
Mytherin: I think we can throw all of these into a single file - otherwise we will end up with many (small) files - and nobody really needs to read or edit this usually anyway.

src/parser/peg/transformer/peg_transformer_factory.cpp
```diff
@@ -434,7 +434,7 @@ void PEGTransformerFactory::RegisterDescribe() {
 
 void PEGTransformerFactory::RegisterDetach() {
 	// detach.gram
-	REGISTER_TRANSFORM(TransformDetachStatement);
+	Register("DetachStatement", &TransformDetachStatementInternal);
```
Mytherin: Instead of having these explicit `Register` calls for every function maybe we can have a list of methods instead, similar to how e.g. `src/catalog/default/default_functions.cpp` does it? We would then also need to generate the wrapper instead of calling `Register` with a template, but given that it's all generated that doesn't seem like a particularly large issue.

This would look something like this:

```cpp
typedef unique_ptr<TransformResultValue> (*transform_function_t) PEGTransformer &transformer, ParseResult &parse_result);

struct TransformRule {
    const char *name;
    transform_function_t transform;    
};

static const TransformRule builtin_transform_rules[] = {
	{"DetachStatement", TransformDetachStatementInternal},
	{nullptr, nullptr}
};
```

Mytherin: As a side-note, I think it would be great if we can get rid of the string look-ups we are doing by looking up the rule names in `transform_functions` all the time, but that's a separate topic

Mytherin (commented):
Thanks for the PR! Looks great - some comments

src/parser/peg/transformer/peg_transformer_factory.cpp
Dtenwolde: > Instead of having these explicit Register calls for every function maybe we can have a list of methods instead, similar to how e.g. src/catalog/default/default_functions.cpp does it?

Good point! I'll make that change

Dtenwolde: > As a side-note, I think it would be great if we can get rid of the string look-ups we are doing by looking up the rule names in transform_functions all the time, but that's a separate topic

This remains future work, but can be more easily done once more functions are auto generated I think. The other feedback has been implemented

src/parser/peg/transformer/transform_generated.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+unique_ptr<TransformResultValue> PEGTransformerFactory::TransformImportStatementInternal(PEGTransformer &transformer,
+                                                                                         ParseResult &parse_result) {
+	auto &list_pr = parse_result.Cast<ListParseResult>();
+	auto string_literal = transformer.Transform<string>(list_pr, 2);
+	auto result = TransformImportStatement(string_literal);
+	return make_uniq<TypedTransformResult<unique_ptr<SQLStatement>>>(std::move(result));
+}
+
+void PEGTransformerFactory::RegisterGenerated() {
```
Dtenwolde, line 122: See here for auto registering rules

hannes:
somewhat related - have you considered replacing the indices with strings that match the parser rule entries?

src/include/duckdb/parser/peg/transformer/peg_transformer.hpp
```diff
@@ -1228,6 +1212,10 @@ class PEGTransformerFactory {
 	static string TransformVacuumOption(PEGTransformer &transformer, ParseResult &parse_result);
 	static vector<string> TransformNameList(PEGTransformer &transformer, ParseResult &parse_result);
 
+#define DUCKDB_INSIDE_PEG_TRANSFORMER_HPP
+#include "duckdb/parser/peg/transformer/peg_transformer_generated.hpp"
```
Mytherin, line 1216: What we do elsewhere in code is have comments or other signals that signal where to start / end the generated code, then we don't need this include. e.g. something like:

```
// START GENERATED RULES

// END GENERATED RULES
```

Maybe you can adopt that in a follow-up PR

Mytherin:
Thanks! Looks great

Dtenwolde:
> somewhat related - have you considered replacing the indices with strings that match the parser rule entries?

I did, but this would introduce another map with strings for the rulename and the goal it to minimize string comparisons. With the current approach, the developer would never need to use indices again as he only has to write a function that has as arguments the child rules with the correct types. For instance `UseTarget` results in `QualifiedName`, so the user has to implement only the following: 
```cpp
// UseStatement <- 'USE' UseTarget
unique_ptr<SQLStatement> PEGTransformerFactory::TransformUseStatement(const QualifiedName &use_target) {
	string value_str;
	if (IsInvalidSchema(use_target.schema)) {
		value_str = SQLIdentifier::ToString(use_target.name);
	} else {
		value_str = SQLIdentifier(use_target.schema) + "." + SQLIdentifier(use_target.name);
	}

	auto value_expr = make_uniq<ConstantExpression>(Value(value_str));
	return make_uniq<SetVariableStatement>("schema", std::move(value_expr), SetScope::AUTOMATIC);
}
```

Therefore, we don't need to worry about indices as the script handles that.

PR #22558, opened 2026-05-11
Move Jemalloc into core

Maxxen:
Supersedes https://github.com/duckdb/duckdb/pull/22479

This PR removes `jemalloc` as an extension and instead includes it as a static `third_party` library which is included a linked by default on linux. It also adds a new top-level cmake option to explicitly disable/enable jemalloc on supported platforms.


Makefile
```diff
@@ -147,7 +147,10 @@ ifeq (${BUILD_JSON}, 1)
 	BUILD_EXTENSIONS:=${BUILD_EXTENSIONS};json
 endif
 ifeq (${BUILD_JEMALLOC}, 1)
-	BUILD_EXTENSIONS:=${BUILD_EXTENSIONS};jemalloc
+	CMAKE_VARS:=${CMAKE_VARS} -DENABLE_JEMALLOC=1
+endif
+ifeq (${BUILD_JEMALLOC}, 0)
+	CMAKE_VARS:=${CMAKE_VARS} -DENABLE_JEMALLOC=0
```
smvv:
```makefile
ifdef BUILD_JEMALLOC
	CMAKE_VARS:=${CMAKE_VARS} -DENABLE_JEMALLOC=${BUILD_JEMALLOC}
endif
```
note: this might be simpler?

src/common/allocator.cpp
```diff
@@ -18,22 +18,132 @@
...
+}
+}
 
-#ifdef __GLIBC__
-#include <malloc.h>
 #endif
 
 namespace duckdb {
 
+#ifdef USE_JEMALLOC
```
lnkuiper: Can we move the jemalloc-related calls into a separate file and include it here instead of inlining the jemalloc stuff in `allocator.cpp`?

Would be really clean to have this is a separate header/implementation file. Don't worry about the allocation calls not being inlined then, they weren't before either

lnkuiper (commented):
Thanks! Looks great, I just have one comment, then this is ready to go from my side:

src/common/allocator.cpp
Maxxen: Done! LMK what you think

Maxxen: Actually, can we maybe do this later? there's a bunch of external builds that glob up all the source files (amalgamation, swift) and making them conditionally include the right file is going to be messy.

Maxxen: Nvm think I fixed it.

lnkuiper:
Looks great! Just missing an include now for the tidy check - can be merged once fixed

Mytherin:
Thanks! This is great but I worry this might cause upstream breakages due to the `CMake` variables changing. Several upstream client distributions call into our CMake directly, and `-DSKIP_EXTENSIONS=jemalloc` or `-DBUILD_EXTENSIONS=jemalloc` now no longer work. That could cause some upstream annoyances which would be best avoided for a bug-fix release imo. Maybe we can parse these for `jemalloc` in the CMake and toggle `ENABLE_JEMALLOC` based on those, at least for `v1.5-variegata`? We could potentially add in a warning and disable this again in `v2.0`.

Mytherin:
Can also be fixed in a follow-up PR, doesn't necessarily have to be picked up in this one

Maxxen:
Pushed a followup that enables controlling jemalloc linking through BUILD/SKIP_EXTENSIONS here https://github.com/duckdb/duckdb/pull/22594

Issue #22103, opened 2026-04-16
read_json_auto regression in v1.5.2: timestamp format mismatch when JSON has mixed ISO 8601 formats across fields


guizmaii:
### What happens?

After upgrading from DuckDB v1.5.1 to v1.5.2, `read_json_auto` fails when reading a JSON file that contains timestamp-like strings with different ISO 8601 variants across different fields.

The auto-detection locks onto format `%Y-%m-%dT%H:%M:%SZ` (from a field with `Z`-suffixed timestamps) and then fails when parsing another field that uses `+00:00` offset with milliseconds.

This worked in v1.5.1 because those fields were inferred as `VARCHAR`. The regression was probably introduced by #21660 which added new timestamp format patterns.

### To Reproduce

```sql
-- Create a JSON file with two fields using different ISO 8601 timestamp variants
COPY (
  SELECT '{"date_z":"2026-08-12T00:00:00Z","date_offset":"2026-08-12T00:00:00.000+00:00"}'
) TO '/tmp/mixed_timestamps.json' (FORMAT CSV, QUOTE '', HEADER 0);

-- This fails in v1.5.2 but worked in v1.5.1 (fields were VARCHAR)
SELECT * FROM read_json_auto('/tmp/mixed_timestamps.json');
```

**Error:**
```
Invalid Input Error: JSON transform error in file "/tmp/mixed_timestamps.json", in record/value 1:
Could not parse string "2026-08-12T00:00:00.000+00:00" according to format specifier "%Y-%m-%dT%H:%M:%SZ"
```

### Expected behavior

Either:
1. Each column should independently select the best matching format (rather than sharing one format across all columns), or
2. The auto-detection should fall back to `VARCHAR` when a single format can't cover all values, or
3. The format list should be tried per-value with fallback (e.g., try `%Y-%m-%dT%H:%M:%SZ` first, then `%Y-%m-%dT%H:%M:%S.%f%z`)

### Root cause

In `extension/json/json_scan.cpp`, the comment says _"assume this is consistent across columns"_. The format detected from one column is applied to all timestamp-candidate columns. Before #21660, the `+00:00` variant didn't match any format so those fields stayed `VARCHAR`. After #21660, the new patterns cause DuckDB to recognize them as timestamp candidates, but the shared format (locked from a different column) doesn't match.

### Workaround

[...]

PR #22559, opened 2026-05-11
Allow json ts format variation across columns - issue 22103

benfleis:
Fix issue #22103 -- timestamp format inputs bound per-scan (!per-column)

Currently the JSON parser chooses a format when handling a TIMESTAMP,
and keeps that format throughout the scan; instead move to a per-column
model where each column is probed individually, and the matching format
kept.

Also

- add explicit (previously breaking) test to confirm.
- add missing ISO format without TZ
- add safety test to confirm not unintentionally swallowing trailing TZ info

### DIFF
--- a/extension/json/json_functions/json_structure.cpp
+++ b/extension/json/json_functions/json_structure.cpp
@@ -336,7 +336,6 @@ bool JSONStructureNode::EliminateCandidateFormats(const idx_t vec_count, Vector 
 		}
 
 		if (success) {
-			date_format_map.ShrinkFormatsToSize(type, i);
 			return true;
 		}
 	}
--- a/extension/json/json_functions/json_transform.cpp
+++ b/extension/json/json_functions/json_transform.cpp
@@ -293,28 +293,22 @@ static bool TransformFromString(yyjson_val *vals[], Vector &result, const idx_t 
 	return success;
 }
 
-template <class OP, class T>
-static bool TransformStringWithFormat(Vector &string_vector, const StrpTimeFormat &format, const idx_t count,
-                                      Vector &result, JSONTransformOptions &options) {
-	const auto source_strings = FlatVector::GetData<string_t>(string_vector);
-	const auto &source_validity = FlatVector::Validity(string_vector);
-
-	auto target_vals = FlatVector::GetData<T>(result);
-	auto &target_validity = FlatVector::Validity(result);
-
-	bool success = true;
-	for (idx_t i = 0; i < count; i++) {
-		if (!source_validity.RowIsValid(i)) {
-			target_validity.SetInvalid(i);
-		} else if (!OP::template Operation<T>(format, source_strings[i], target_vals[i], options.error_message)) {
-			target_validity.SetInvalid(i);
-			if (success && options.strict_cast) {
-				options.object_index = i;
-				success = false;
-			}
+static bool TryOneFormat(const StrpTimeFormat &fmt, LogicalTypeId type_id, const string_t &input, Vector &result,
+                         idx_t i, string &error_message) {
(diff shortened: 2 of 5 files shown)


extension/json/json_functions/json_transform.cpp
```diff
@@ -857,6 +914,7 @@ bool JSONTransform::Transform(yyjson_val *vals[], yyjson_alc *alc, Vector &resul
 	if ((result_type == LogicalTypeId::TIMESTAMP || result_type == LogicalTypeId::DATE) && options.date_format_map &&
 	    options.date_format_map->HasFormats(result_type.id())) {
 		// Auto-detected date/timestamp format during sampling
+		options.current_column_id = column_index ? column_index->GetPrimaryIndex() : DConstants::INVALID_INDEX;
 		return TransformFromStringWithFormat(vals, result, count, options);
```
lnkuiper, line 879: I don't like adding this to the options, these should ideally not be changed other than for error handling during transformation. We could pass the column id down as an `optional_idx` instead (also instead of using `DConstants::INVALID_INDEX`.

The column ID strategy also doesn't work for nested JSON. Example:
```json
{"duck": 42, "nested": {"goose": 43}}
```
Here, both `"duck"` and `"goose"` have the same column index.

extension/json/json_functions/json_structure.cpp
```diff
@@ -336,7 +336,6 @@ bool JSONStructureNode::EliminateCandidateFormats(const idx_t vec_count, Vector
 		}
 
 		if (success) {
-			date_format_map.ShrinkFormatsToSize(type, i);
 			return true;
```
lnkuiper, line 339: It should still work when people pass an explicit "timestamp_format" parameter. Is that still the case? Then that should be the only one that is accepted.

extension/json/json_functions/json_transform.cpp
```diff
@@ -325,22 +346,58 @@ static bool TransformFromStringWithFormat(yyjson_val *vals[], Vector &result, co
...
-	auto &format = options.date_format_map->GetFormat(result_type);
+	const auto result_type_id = result.GetType().id();
+	D_ASSERT(result_type_id == LogicalTypeId::DATE || result_type_id == LogicalTypeId::TIMESTAMP);
+	const auto &formats = options.date_format_map->GetFormats(result_type_id);
+	const auto source_strings = FlatVector::GetData<string_t>(string_vector);
+	const auto &source_validity = FlatVector::Validity(string_vector);
+	auto &target_validity = FlatVector::Validity(result);
+
+	// Look up or detect the preferred format index for this column
+	idx_t preferred_idx = DConstants::INVALID_INDEX;
```
lnkuiper: I think this is doing more then necessary. If we're doing this on a vector-by-vector basis, why do we need to save the preferred idx? Can't we just do:
1. Detect preferred format for this vector using the first value
2. Apply this format to the vector
3. Start over for the next vector

If we do this we don't need to keep any state, no mutexes, just per-vector detection.

lnkuiper (changes requested):
Thanks for the PR! I think this can be simplified a bit:

extension/json/json_functions/json_transform.cpp
benfleis: Simplified to keep matched_idx local per vector.

extension/json/json_functions/json_structure.cpp
benfleis: Doing so yields a vector singleton, added a test to confirm that when a format is specified, other list-defaults do not get used.

lnkuiper (approved):

lnkuiper:
Thanks for the changes, LGTM now

Mytherin:
Thanks!

PR #22561, opened 2026-05-11
Fix incorrect profiling results when using `LIMIT`

sebastiaan-dev:
This PR resolves https://github.com/duckdblabs/duckdb-internal/issues/9085.

The bug originated from the fact that the placement of `FinishSource` assumed that a pipeline would only finish when the source was exhausted. However `LIMIT` (among others) is an intermediate operator which can trigger an early finish of the pipeline, in these cases `FinishSource` would never trigger.

As a fix the `FinishSource` has been refactored to take any operator, irrespective of if it is active or not, and `FinishSource` has been placed in the finalize phase that catches all finish reasons. Because the meaning of the function has changed slightly it has been renamed to `FinalizeSourceProfiling`.

Because the above changed the context of `FinalizeSourceProfiling` from operator scoped to pipeline scoped a change was made in how `extra_info` metadata is managed. Where we now update the global property instead of overwriting with the local object to account for multiple flushes.

Finally, we made row counting more granular (before it was either 0 or row group size). Such that it in case of for example `CUMULATIVE_ROWS_SCANNED`, the amount of rows positions processed by the table scan is returned, before filters are applied.

### Limitation

The current patch will result in a zero count in profiling for the following query:

```sql
SELECT * FROM range(10) LIMIT 10;
```

This happens because `LIMIT` stops the pipeline before the source can emit a chunk with zero rows. This leads to a situation where there are no metrics, and the source is not exhausted, so we return `0`. After discussion with @lnkuiper this limitation is left as-is, documented through a test.

### DIFF
--- a/test/sql/pragma/profiling/test_custom_profiling_rows_scanned.test
+++ b/test/sql/pragma/profiling/test_custom_profiling_rows_scanned.test
@@ -69,3 +69,121 @@ SELECT
 FROM metrics_output;
 ----
 true
+
+# Show nonzero count when using LIMIT below late-materialization threshold.
+
+statement ok
+CREATE TABLE t AS SELECT range i FROM range(400000);
+
+statement ok
+PRAGMA enable_profiling = 'json';
+
+statement ok
+PRAGMA profiling_output = '__TEST_DIR__/profiling_output.json';
+
+statement ok
+PRAGMA custom_profiling_settings='{"CUMULATIVE_CARDINALITY": "true", "CUMULATIVE_ROWS_SCANNED": "true", "BLOCKED_THREAD_TIME": "true"}';
+
+statement ok
+SELECT * FROM t LIMIT 1;
+
+statement ok
+PRAGMA disable_profiling;
+
+statement ok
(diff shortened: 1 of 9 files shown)


test/sql/pragma/profiling/test_custom_profiling_rows_scanned.test
```diff
@@ -68,4 +68,122 @@ SELECT
...
+statement ok
+PRAGMA disable_profiling;
+
+statement ok
+CREATE OR REPLACE TABLE metrics_output AS SELECT * FROM '__TEST_DIR__/profiling_output.json'
+
+query I
+SELECT cumulative_rows_scanned FROM metrics_output;
+----
+0
```
artjomPlaunov: extra edge case to consider: SELECT * FROM range(10) LIMIT 10

artjomPlaunov:
(Once the edge case I proposed is handled, it would be good to get an extra reviewer on this)

artjomPlaunov (approved):
Great, thank you! :-) And thanks for the extra review @lnkuiper

lame-dev:
LGTM

PR #22563, opened 2026-05-11
Add workflow to mirror open issues labeled PR submitted for more than 10 days

DinosL:
Issues in the public repository may be overlooked after being labeled "PR submitted". This PR adds a workflow that periodically checks for such issues and mirrors them to the internal tracker if they remain open for more than ~~14~~ 10 days.


szarnyasg (approved):

lnkuiper:
Hey, not sure if I'm reading this right but this will also run on forks, right? To avoid this we put our cron jobs in https://github.com/duckdb/duckdb-workflow-trigger

DinosL:
I see! I removed the cron and added a trigger in workflow-trigger https://github.com/duckdb/duckdb-workflow-trigger/pull/14

.github/workflows/MirrorStalePRSubmittedIssues.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  workflow_dispatch: {}  # allow running manually from the github ui
+
+env:
+  GH_TOKEN: ${{ secrets.DUCKDBLABS_BOT_TOKEN }}
+  DAYS_BEFORE_MIRROR: 10
+
+jobs:
+  create_or_label_mirror_issue:
+    runs-on: ubuntu-latest
```
smvv, line 12: Do we need something like `if: ${{ github.repository_owner == 'duckdb' }}` here to restrict these jobs to the duckdb org?

edit: oh i see the cron schedule is removed. I guess this is OK to merge.

Mytherin:
Thanks!

PR #22566, opened 2026-05-11
Fix fuzzer issue on empty subscript expression

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/9063

The following query failed with an internal exception, while this was previously a parser error: 
```sql
n''''+ST�&6>ttacrotablesampimportleT�&6>tta�rordeT�Niqeger[], e�tfilter%
t_not_currorderent;
;
```

It looks like an odd query, but it seems to be caused by the `[]` which matches a `SliceExpression`. Every element of this subrule is optional: 
`SliceBound <- Expression? EndSliceBound? StepSliceBound?`
But a completely empty `SliceExpression` should not be accepted (Previously led to a generic `syntax error`). So now we check if the result is empty and throw a descriptive error if this happens. 

I am surprised this wasn't previously caught in other tests


Mytherin:
Thanks!

PR #22568, opened 2026-05-11
Fix BOOLEAN cast to accept ASCII-padded inputs

### LINKED ISSUE #22565: TRY_CAST('true' AS BOOLEAN) returns true but TRY_CAST('  true  ' AS BOOLEAN) returns NULL — analogous to #22484
## Describe the bug

The string-to-boolean cast uses a length-switched parser with no whitespace handling. Any leading or trailing ASCII whitespace (space, tab, newline) on the input causes `TRY_CAST(... AS BOOLEAN)` to return `NULL` and `CAST(... AS BOOLEAN)` to raise a conversion error. The fix from #22484 covered numeric parsers; the boolean parser is a separate code path that was not touched.

## Expected behavior

Symmetric with the numeric-cast behaviour (per #22484): leading and trailing ASCII whitespace is trimmed before the parse, so `CAST('  true  ' AS BOOLEAN)` returns `true`.

## Actual behavior

All eight accepted boolean spellings (`true`, `false`, `t`, `f`, `yes`, `no`, `1`, `0`) fail with any surrounding whitespace.

## Root cause (identified)

- `src/include/duckdb/common/operator/cast_operators.hpp:483-541` — `TryCastStringBool` is a `switch (input_size)` with explicit cases for sizes 1, 2, 3, 4, 5. The default branch returns `false`. **No `trim`, `isspace`, or `StringUtil::CharacterIsSpace` call.** Input of size 8 (`'  true  '`) hits the default and fails.
- `src/common/operator/cast_operators.cpp:945-950` — `TryCast::Operation<string_t, bool, ...>` calls `TryCastStringBool` directly with `input.GetData()` / `input.GetSize()`; no preprocessing.
- `src/function/cast/string_cast.cpp:94-95` — registers VARCHAR→BOOLEAN through `VectorCastHelpers::TryCastStrictLoop<string_t, bool, duckdb::TryCast>`; no trimming wrapper.

Contrast with the integer parser:

- `src/include/duckdb/common/operator/integer_cast_operator.hpp:400-405` — `TryIntegerCast` opens with `while (len > 0 && StringUtil::CharacterIsSpace(*buf)) { buf++; len--; }` (the #22484 fix family).

The two cast paths use different operations and the boolean one was never updated.

## Suggested fix

Add a leading/trailing whitespace skip step to `TryCastStringBool` before the length switch — or factor the integer parser's whitespace skip into a shared helper and call it from both paths. The fix is mechanical and self-contained.

I'd like to work on a regression test and a minimal fix for this.

### To Reproduce

```sql
SELECT TRY_CAST('true'      AS BOOLEAN);    -- true
SELECT TRY_CAST('  true  '  AS BOOLEAN);    -- NULL  ← bug
SELECT TRY_CAST(' true'     AS BOOLEAN);    -- NULL  ← bug
SELECT TRY_CAST('true '     AS BOOLEAN);    -- NULL  ← bug
SELECT TRY_CAST(E'true\n'   AS BOOLEAN);    -- NULL  ← bug
SELECT TRY_CAST(E'\ttrue'   AS BOOLEAN);    -- NULL  ← bug

-- Reference: integers accept the same padding (per #22484)
SELECT TRY_CAST('  1  '     AS INTEGER);    -- 1
SELECT TRY_CAST('  3.14  '  AS DOUBLE);     -- 3.14

-- Strict CAST raises an error:
SELECT CAST('  true  ' AS BOOLEAN);
-- Conversion Error
```

### OS:

OSX 26.3.1 arm64

### DuckDB Version:

**v1.5.2** and **main**

### DuckDB Client:

duckdb cli

### Hardware:

Macbook Pro

### Full Name:

Dave York

### Affiliation:

Govini

- [x] Yes, I have

- [x] Yes, I have

d4v3y0rk:
Fixes #22565

The string-to-boolean parser (`TryCastStringBool` in `src/include/duckdb/common/operator/cast_operators.hpp`) dispatched on `input_size` alone, so padded inputs like `'  true  '` (size 8) fell through the length switch's default branch and failed. `CAST('  true  ' AS BOOLEAN)` raised a conversion error and `TRY_CAST(...)` returned NULL, while `CAST('  1  ' AS INTEGER)` succeeded via `TryIntegerCast`'s whitespace skip.

This change adds a leading- and trailing-whitespace skip at the top of `TryCastStringBool`, mirroring `TryIntegerCast`'s pattern in `integer_cast_operator.hpp:400-405`. All eight accepted boolean spellings (`true`, `false`, `t`, `f`, `yes`, `no`, `1`, `0`) now parse identically with or without surrounding whitespace.

## Validation

- Added regression test cases to `test/sql/cast/test_boolean_cast.test` covering padded `TRY_CAST` of all eight boolean spellings, strict `CAST` with surrounding whitespace, and rejection of whitespace-padded garbage
- Ran `test/sql/cast/test_boolean_cast.test` — 55 assertions pass
- Ran broader cast and bool suites (`test/sql/cast/*`, `bool_casts.test`, `test_boolean_not.test`, `test_bool.test`, `boolean_autocast.test`) — all pass
- Ran `make format-fix`


d4v3y0rk:
Heads up: the only failing check is `windows / Windows (64 Bit)`, and the failure is in `src/common/complex_json.cpp` — a file this PR doesn't touch.

The error is the classic `windows.h` `GetObject` macro collision rewriting `ComplexJSON::GetObject` to `ComplexJSON::GetObjectA`:

```
src/common/complex_json.cpp(20): error C2039: 'GetObjectA' is not a member of 'duckdb::ComplexJSON'
src/common/complex_json.cpp(21): error C2065: 'type' is undeclared (cascade)
src/common/complex_json.cpp(22): error C2065: 'obj_value' is undeclared (cascade)
```

This diff only touches `src/include/duckdb/common/operator/cast_operators.hpp` and `test/sql/cast/test_boolean_cast.test`, neither of which is reachable from `complex_json.cpp`. Every other check passes — Linux Release / Relassert / TSan / Tidy, regression suite, all extensions.

#22566 merged a few hours later on a newer `main` with `Windows (64 Bit)` passing, so it looks like a transient include-ordering hit on my CI run's base (`a697a1c976`). Happy to merge `main` and push for a fresh run if that helps, otherwise feel free to re-run the failing job whenever convenient.

src/include/duckdb/common/operator/cast_operators.hpp
```diff
@@ -481,6 +482,17 @@ DUCKDB_API bool TryCast::Operation(double input, double &result, bool strict);
 // String -> Numeric Casts
 //===--------------------------------------------------------------------===//
 static inline bool TryCastStringBool(const char *input_data, idx_t input_size, bool &result, bool strict) {
+	// Skip leading and trailing ASCII whitespace before dispatching on
+	// length. Without this, padded inputs like '  true  ' (size 8) fall
+	// through to the default case and fail, while CAST('  1  ' AS INTEGER)
+	// succeeds via TryIntegerCast's whitespace skip. Fixes #22565.
+	while (input_size > 0 && StringUtil::CharacterIsSpace(*input_data)) {
```
Mytherin: Maybe we can only do this if the code below fails to parse the bool? The below code seems to be very optimized and this code is likely going to impact performance negatively when in most cases it will be unnecessary.

d4v3y0rk: I will make the requested changes. FYI this and the other bug I found are among about 6 I have found so far using a metamorphic fuzzer with different invariants. I am submitting them one at a time so I can follow along with each one and so as not to overload your team. Happy to discuss the fuzzing technique I am using if you or the team are interested in this kind of testing.

d4v3y0rk: Also the reason I did this the way I did was because I thought it was appropriate considering DuckDB does the same thing in other places. 
- TryIntegerCast (integer_cast_operator.hpp:400-405): strips leading whitespace unconditionally at the top, then handles trailing inside the parse loop
 - TryDoubleCast (double_cast_operator.hpp:17-52): strips leading whitespace unconditionally at the top, skips trailing after the parse result

PR #22571, opened 2026-05-11
Fix for SIGABRT in setting size on zero-capacity vector

ccfelius:
We should return early after setting size on a zero capacity vector instead of falling through. 

fixes https://github.com/duckdblabs/duckdb-internal/issues/9071


pdet:
Thanks, LGTM!

PR #22572, opened 2026-05-11
Push partial aggregates below joins

zurferr:
## Summary

Pre-aggregates `SUM`/`COUNT`/`MIN`/`MAX` (and AVG-state via decomposition) below a comparison join when the aggregate-side dominates cardinality. Catches the TPC-DS Q22 shape where the fact table is grouped by dimension columns — the partial aggregate collapses the fact stream before the join, the upper aggregate finalises from state.

This is the cleaned-up successor to my own #22572 (draft). All of Mark's review comments on that PR are addressed here:

- ✅ The 200-line `TryPushdownAggregate` is split into <50-line helpers (`AnalyzePushdown`, `BuildLowerGroupMap`, `BindPushdownAggregates`, `CreateLowerAggregate`, `CreateJoin`, `CreateUpperAggregate`).
- ✅ Supported-aggregate gate uses `BoundAggregateExpression::HasStateCombineCallback()` instead of a hard-coded name list — any aggregate exposing the `EXPORT_STATE`/`COMBINE`/`FINALIZE` callbacks is eligible.
- ✅ `EXPLAIN`/`REGEX` plan-shape tests in `test/optimizer/partial_aggregate_pushdown.test`: positive match for `finalize_combine_aggr` on baseline + ROLLUP, negative match when join is highly selective.

## How it works

For an aggregate over a comparison join where the GROUP BY only references one side's columns ("dimension side"), the pass:

1. Splits each pushable aggregate into an `EXPORT_STATE` partial below the join and a `finalize_combine_aggr` finaliser above it.
2. Inserts a lower `LogicalAggregate` on the aggregate-side child whose groups are the (a) original GROUP BY keys reachable from that side and (b) all join-key columns on that side.
3. Rewrites the join to reference the lower aggregate's output (column-binding remapping via `ColumnBindingReplacer`).
4. The upper aggregate consumes the partial states; ROLLUP/CUBE/GROUPING SETS are preserved because `grouping_sets` indices remain valid (the upper groups are built in the same order as the original groups).

`finalize_combine_aggr` is a new aggregate function in `aggregate_export.cpp` that mirrors `combine_aggr` but binds to the underlying aggregate's actual return type — same `AGGREGATE_STATE → combine → finalize` pipeline already used by `EXPORT_STATE` / `COMBINE_AGGR`.

### Heuristics (`PartialAggregatePushdownHeuristics`)

- `MIN_DIMENSION_GROUPS = 4` — skip if the dimension produces fewer distinct groups than this; pushdown buys nothing.
- `MIN_AGGREGATE_TO_DIMENSION_RATIO = 4` — skip unless `aggregate_side.cardinality ≥ 4 × dimension_side.cardinality`.
- `MAX_JOIN_SELECTIVITY_INV = 8` — skip if `join.cardinality × 8 < aggregate_side.cardinality`. (Without this, a fact ⋈ filtered date_dim shape pre-aggregates 700K rows when post-aggregating 1K rows is cheaper — caught from a Q92-shape regression.)

### AVG under ROLLUP

`AvgRewriteRule::ShouldSkip` previously bailed for any `multi_grouping_set` aggregate. `SUM(x)/COUNT(x)` is mathematically identical to `AVG(x)` under any grouping (both ignore NULLs identically), so the guard was over-conservative. Relaxing it (to keep only the `grouping_functions` guard) is the prerequisite for PAP to fire on Q22 — once AVG decomposes into SUM + COUNT, PAP can push both exact aggregates down.

PAP itself still has a defensive belt against AVG under multi-grouping-set because the AVG `AGGREGATE_STATE` blob serialises a fixed-width SUM that overflows to `inf` when many partial states combine at shallow ROLLUP levels. The rewriter path means PAP sees `SUM`/`COUNT`, not `AVG`, so the belt rarely engages.

## Performance — TPC-DS SF10

Measured against vanilla `main`, single-threaded, 99-query suite:

| | vanilla | this PR | speedup |
|---|---:|---:|---:|
| **Q22** (ROLLUP on inventory) | 8.65 s | **0.78 s** | **11×** |
| Q11 (gross-profit YoY) | 2.43 s | 1.73 s | 1.4× |
| Q3 (single-fact aggregation) | 0.24 s | 0.27 s | — |
| Q7 (item-promotion aggregation) | 0.52 s | 0.54 s | — |
| Q4 (year-over-year customer total) | 4.02 s | 2.85 s | 1.4× |
| **Total engine** (sum of 99) | **94.33 s** | **83.46 s** | **1.13× (11.5% faster)** |
| Passing | 99/99 | 99/99 | — |

Numbers from same-binary A/B (same duckdb build, only `SET disabled_optimizers='partial_aggregate_pushdown'` toggled between runs).

Q22 plan before:

```
HASH_GROUP_BY (ROLLUP, 5 sets)               ← 8s
  Aggregates: avg(inv_quantity_on_hand)
└── HASH_JOIN inv ⋈ item
    ├── HASH_JOIN inv ⋈ date_dim
    │   └── SEQ_SCAN inventory               ← 133M rows at SF10
```

After:

```
HASH_GROUP_BY (ROLLUP, 5 sets)
  Aggregates: finalize_combine_aggr(sum_state), finalize_combine_aggr(count_state)
└── HASH_JOIN inv_pre_agg ⋈ item             ← join input now ~3M rows
    └── HASH_GROUP_BY (...keys + dim cols...)
        Aggregates: export_aggregate(sum), export_aggregate(count)
        └── HASH_JOIN inv ⋈ date_dim
```

## Files

- `src/optimizer/partial_aggregate_pushdown.{cpp,hpp}` — the pass.
- `src/optimizer/aggregate_function_rewriter.cpp` — relax `AvgRewriteRule::ShouldSkip` for multi-grouping-set.
- `src/function/scalar/system/aggregate_export.cpp` + `generic_common.hpp` — new `FinalizeCombineAggregateFunction`.
- `src/optimizer/optimizer.cpp` + `CMakeLists.txt` — registration.
- Bookkeeping (enum_util, metric_type, optimizer_type, profiling_utils) via `make generate-files`.
- `test/optimizer/partial_aggregate_pushdown.test` — correctness + EXPLAIN/REGEX plan-shape checks.

## Test plan

- [x] `ninja duckdb` (release) clean build on this branch.
- [x] `test/optimizer/partial_aggregate_pushdown.test` — correctness vs disabled-optimizer baseline, ROLLUP correctness, AVG single-grouping-set, COUNT_STAR must not crash, selectivity-guard negative EXPLAIN check.
- [x] TPC-DS SF10 99-query suite — 99/99 pass, results numerically identical to baseline.
- [ ] CI.

## Open questions

- The `MAX_JOIN_SELECTIVITY_INV = 8` threshold was calibrated by hand against Q92. Happy to swap for a stats-driven check if you have a preferred form.
- AVG-state precision under deep ROLLUP is not addressed here — the rewriter rerouting is the practical fix. If you'd prefer a widened state serialisation as the proper fix, I can take that on as a follow-up.


src/optimizer/partial_aggregate_pushdown.cpp
```diff
@@ -0,0 +1,326 @@
...
+}
+
+void PartialAggregatePushdown::VisitOperator(unique_ptr<LogicalOperator> &op) {
+	LogicalOperatorVisitor::VisitOperator(op);
+	if (TryPushdownAggregate(op)) {
+		LogicalOperatorVisitor::VisitOperator(op);
+	}
+}
+
+bool PartialAggregatePushdown::TryPushdownAggregate(unique_ptr<LogicalOperator> &op) {
```
lnkuiper: This function is ~200 lines, could you split it up into multiple functions that each have a distinct use?

src/optimizer/partial_aggregate_pushdown.cpp
```diff
@@ -0,0 +1,326 @@
...
+	binding = expr.Cast<BoundColumnRefExpression>().binding;
+	return true;
+}
+
+static bool IsSumAggregate(const BoundAggregateExpression &expr) {
+	if (expr.IsDistinct() || expr.filter || expr.order_bys || expr.children.size() != 1) {
+		return false;
+	}
+	return StringUtil::Lower(expr.function.GetName()) == "sum";
+}
```
lnkuiper: Should be possible to get this to work with many more aggregate functions, maybe in a follow-up PR. I understand that you're using `SUM` here because you can re-aggregate, but we have `EXPORT_STATE`/`COMBINE`/`FINALIZE`/`COMBINE_AGGR` that can do this for a wide range of aggregate functions.

Using just `sum` in this initial PR is good, but it should use the functionality we have for exporting aggregate states rather than relying on the semantics of `SUM`.

test/optimizer/partial_aggregate_pushdown.test
```diff
@@ -0,0 +1,93 @@
...
+FROM (
+	SELECT * FROM partial_aggregate_disabled
+	EXCEPT ALL
+	SELECT * FROM partial_aggregate_enabled
+);
+----
+0
+
+statement ok
+SET disabled_optimizers='';
```
lnkuiper: This needs a few `EXPLAIN`/`REGEX` tests to verify that the optimization is actually being triggered.

lnkuiper (changes requested):
Thanks for the PR! I've left some comments:

src/function/scalar/system/aggregate_export.cpp
```diff
@@ -968,11 +984,18 @@ ScalarFunctionSet CombineFun::GetFunctions() {
...
-	                      FunctionNullHandling::SPECIAL_HANDLING, nullptr, CombineAggrBind, nullptr, nullptr, nullptr);
-	function.SetNullHandling(FunctionNullHandling::SPECIAL_HANDLING);
+	auto function = AggregateFunction("combine_aggr", {LogicalTypeId::AGGREGATE_STATE}, LogicalTypeId::AGGREGATE_STATE,
+	                                  nullptr, nullptr, CombineAggrUpdate, nullptr, CombineAggrFinalize, nullptr,
+	                                  CombineAggrBind, nullptr, nullptr, nullptr);
+	function.SetNullHandling(FunctionNullHandling::DEFAULT_NULL_HANDLING);
+	return function;
+}
+
+AggregateFunction FinalizeCombineAggregateFunction::GetFunction() {
```
lnkuiper: Instead of implementing this function that does `combine_aggr` and `finalize`, can we just push a projection in the optimizer on top of the aggregate that does the `finalize`? That reuses existing code and seems cleaner overall.

src/optimizer/partial_aggregate_pushdown.cpp
```diff
@@ -0,0 +1,453 @@
...
+	// here. Bail before FindAggregateSide tries to access expr.children[0].
+	if (expr.children.size() != 1) {
+		return false;
+	}
+	// Gate on the EXPORT_STATE / FINALIZE_COMBINE infrastructure rather than on
+	// a hard-coded name list (per Mark's review on duckdb#22572). Any aggregate
+	// that exposes a state-combine callback can be split into a lower partial
+	// aggregate (yielding AGGREGATE_STATE) and an upper finalize_combine_aggr
+	// (consuming that state) — that's exactly the pipeline this pass builds.
+	if (!expr.function.HasStateCombineCallback()) {
```
lnkuiper: I think this should check `!expr.function.HasGetStateTypeCallback()` instead. Also, please drop the excessive comments that link back to the PR feedback, keep it brief.

src/optimizer/partial_aggregate_pushdown.cpp
```diff
@@ -0,0 +1,453 @@
...
+		return false;
+	}
+	// AVG's exported state serializes to a fixed blob that overflows to `inf`
```
lnkuiper: I don't think any `avg` function serializes to a blob anymore. I also don't understand why we need this defensive check here. The rewrite should be semantically equivalent to the original query so why should we get any `inf`?

lnkuiper (changes requested):
Thanks for the changes! I've left a few more comments:

lnkuiper:
In second thought, the new aggr is OK, maybe call it finalize_aggr

zurferr:
Thanks again for the review. I folded the cleanup into the PR and the CI is looking good so far.

The rewrite now uses the existing `combine_aggr` + `finalize` path instead of a custom helper, with a small guard for wide lower groups to avoid the earlier Q67-style regression while keeping the Q04/Q11/Q22 wins.

lnkuiper (approved):
Thanks for the changes! I've added some more tests myself, and they all passed. Can be merged once CI passes.

Issue #22546, opened 2026-05-10
variant_extract returns incorrect result


dentiny:
### What happens?

```sql
memory D SELECT variant_extract((SELECT list(i) FROM range(500) t(i))::VARIANT, 300::UINTEGER)::BIGINT;
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ CAST(variant_extract(CAST((SELECT list(i) FROM "range"(500) AS t(i)) AS "VARIANT"), CAST(300 AS "UINTEGER")) AS BIGINT) │
│                                                          int64                                                          │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│                                                                                                                      43 │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```
which I expect 299 as return; I suspect we truncate / mis-cast the index, since 300-43 is roughly 256

### To Reproduce

Repro sql attached above.

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

PR #22573, opened 2026-05-11
Fix variant selection vector index

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22546

Both types are of `uint32_t`, we shouldn't downcast here which leads to incorrect result.


pdet:
Thanks! LGTM!

PR #22588, opened 2026-05-12
Limit parallel linker jobs to avoid out-of-memory errors

smvv:
This also exists in the branch `main` and avoids OOM errors on high CPU core machines.


Tishj (approved):
Thanks, can confirm this fixes compilation failure on linux machines 👍

Mytherin:
Thanks!

PR #22591, opened 2026-05-12
ARTOperator::Delete return false if rowid not found in nested ART leaf

artjomPlaunov:
Delete was returning true regardless if the rowid is found in the leaf of a nested ART, which was causing a false positive for the partially deleted removal target exception in RowGroupCollection::RemoveFromIndexes in a concurrent checkpointing scenario. 

https://github.com/duckdblabs/duckdb-internal/issues/9132


taniabogatsch (approved):

Mytherin:
Thanks!

PR #22600, opened 2026-05-12
Bump AWS extension

staticlibs:
This PR updates `duckdb-aws` extension to include
duckdb/duckdb-aws#144 change. This change introduces new `rds` secret type.

The same way like with other secret types (`s3`, `r2` etc) in `duckdb-aws`, the type itself is registered on consumer side in `duckdb-postgres` extension (added to `extension_entries.hpp` in #22579) and the provider for it is registered in `duckdb-aws`.

Unlike existing secret types (`s3`, `r2` etc) there is no `config` provider registered for `rds` in `duckdb-postgres` - only the `credential_chain` provider (in `duckdb-aws`) is used for this type.


carlopi:
FYI: @Tishj, I think you were looking into bumping aws as well, so you know.

Mytherin:
Thanks!

PR #22601, opened 2026-05-12
Add support for variables in sqllogictest loops, rewrite parallel CSV test to be a sqllogictest, and add `rejects_line_size_limit` option to CSV reader

Mytherin:
The "test parallel CSV" C++ tests parse CSV files with buffer sizes based on the maximum line length to stress test the CSV reader. In order to get the maximum line length, we currently use an internal debug flag (`debug_set_max_line_length`) which keeps track of the maximum line length while parsing. This is not very pretty.

This PR instead uses the rejects tables to obtain the maximum line length and removes the `debug_set_max_line_length` flag. Essentially we try to parse the CSV file with the correct columns, but incorrect types (converting every column type to `STRUCT(zxcvb INT)`). This will then succeed in parsing the CSV file and getting the correct line, but fail during result conversion. This then adds the line to the rejects table as the CSV reader has read it - which we can then use to obtain the maximum line length.

This does require the addition of the `rejects_line_size_limit` option. Currently rejects lines are silently truncated to 10K bytes. This PR makes this configurable instead (with `rejects_line_size_limit=0` is unlimited) - and also fixes a bug where this truncation could introduce invalid UTF8 into the system.

### SQLLogicTest Loops

To facilitate this, this PR adds support for using variables in sqllogic test loops using the `<variable:{VAR}>` syntax, e.g.:

```sql
statement ok
set variable csv_files = (select list(file) from (select file from glob('...'));

foreach file <variable:csv_files>

// do something with the file

endloop
```

The variable should be a list. 

This allows some pretty flexible / powerful loops to be written as we can now loop over files, etc, dynamically instead of having to hard-code these lists.


pdet:
It is just crazy how this is possible today 😂

PR #9197, opened 2023-10-03
Bug report template: ask for imports to be included

szarnyasg:

PR #22604, opened 2026-05-12
Internal #9197: Timestamp Cleanup

hawkfish:
* Use templating to unify infinity formatting.
* Use templating to unify casting timestamps to strings.
* Use epoch instead of zero.
* Replace ICUCast with Cast.


Mytherin:
Thanks!

PR #22298, opened 2026-04-27
Prefer filtered build sides for join filter pushdown

zurferr:
This keeps a selective join input on the build side when that input can drive join filter pushdown into a much larger scan.

The build/probe side optimizer normally picks the cheapest hash-table build. For queries such as TPC-DS Q13 and Q85, that can move a filtered `date_dim` side to the probe side, so the later join filter can no longer be pushed into the large sales scan.

The change is narrow:

- reuse the join-filter pushdown expression and join-type support checks
- only consider equality conditions with an existing pushdown target
- require a filtered candidate build side and a large target scan relative to that build side
- apply the dynamic-filter benefit as a multiplier on the existing build-size cost, not as a separate override

TPC-DS SF100, sorted Parquet, 4 threads, 12GB memory, cold cache spot-check after cleanup:

| Query | main | this PR | Speedup |
|---|---:|---:|---:|
| Q13 | 57.239s | 11.618s | 4.93x |
| Q85 | 13.276s | 4.504s | 2.95x |
| Q23 | 58.881s | 53.664s | 1.10x |
| Q67 | 57.799s | 57.787s | 1.00x |

All result hashes matched main.

Validation:

- `python3 scripts/ci/run_tests.py --workers=50% build/release/test/unittest`
- `./build/release/test/unittest '[joins]'`
- `./build/release/test/unittest '[pushdown]'`
- `./build/release/test/unittest test/optimizer/joins/join_filter_build_side.test`

PR #22607, opened 2026-05-12
Prefer filtered build sides for runtime filters

zurferr:
This keeps the runtime-filter change local to build/probe-side optimization. When a filtered join side can produce a join filter for a large scan on the other side, the optimizer gives that build side a guarded bonus on top of the existing build-side cost.

The bonus uses the estimated cardinality of the scan that would receive the dynamic filter, and is only applied when the build side is small relative to that target. The expression and join-type checks are shared with `JoinFilterPushdownUtil`, so the cost signal stays aligned with the runtime filters the engine can actually produce.

I removed the broader join-order cost-model change after CI showed IMDB benchmark regressions. This PR no longer changes `cost_model.cpp`.

Impact:

SF100 TPC-DS on sorted Parquet, Azure `Standard_D4ads_v5` (4 vCPU), `threads=4`, `memory_limit='12GB'`, Linux page cache dropped before every query.

Base `3dd09a1d53` -> this PR `ca08677d10`:

- 99-query total: 1172.13s -> 1135.69s (1.03x, -3.1%)
- geomean query-time ratio: 0.985 (1.016x)
- Q13: 56.91s -> 11.41s (4.99x)
- Q85: 13.37s -> 4.45s (3.00x)
- Q72 is neutral in this reduced PR: 31.72s -> 31.82s

Raw output hashes match for 96/99 queries. The remaining Q18/Q65/Q71 results differ only by row order; sorted outputs match.

I also repeated Q13/Q23/Q44/Q46/Q48/Q85 in the opposite order (`PR` first, `base` second). Q44/Q46/Q48 were neutral there, and Q23 was neutral as well (58.47s vs 58.37s). Q13/Q85 retained the same large wins.

Tests:

- `ninja -C build/release duckdb unittest`
- `python3 scripts/ci/run_tests.py build/release/test/unittest test/optimizer/joins/join_filter_build_side.test`
- `python3 scripts/ci/run_tests.py build/release/test/unittest test/optimizer/*`
- `build/release/test/unittest`
- `PATH=/tmp/duckdb-format-venv/bin:$PATH python3 scripts/format.py origin/main --check --silent`
- `git diff --check`

### DIFF
--- a/.github/regression/micro.csv
+++ b/.github/regression/micro.csv
@@ -8,6 +8,7 @@ benchmark/micro/copy/to_parquet_partition_by_few.benchmark
 benchmark/micro/copy/to_parquet_partition_by_many.benchmark
 benchmark/micro/limit/parallel_limit.benchmark
 benchmark/micro/filter/parallel_complex_filter.benchmark
+benchmark/micro/join/join_filter_build_side.benchmark
 benchmark/micro/catalog/add_column_empty.benchmark
 benchmark/micro/groupby-parallel/many_groups_large_values_small.benchmark
 benchmark/tpch/csv/lineitem_csv_auto_detect.benchmark
--- a/benchmark/micro/join/join_filter_build_side.benchmark
+++ b/benchmark/micro/join/join_filter_build_side.benchmark
@@ -1,0 +1,28 @@
+# name: benchmark/micro/join/join_filter_build_side.benchmark
+# description: Prefer a filtered build side when it can produce a join filter for a large probe scan
+# group: [join]
+
+name Join Filter Build Side
+group micro
+subgroup join
+
+require parquet
+
+cache join_filter_build_side.duckdb
+
+load
+CREATE TABLE build_payload AS PIVOT (
+	SELECT i, 'c' || j::VARCHAR AS name, NULL::VARCHAR AS val
+	FROM range(200000) _(i), range(400) __(j)
+) ON name USING first(val) GROUP BY i;
+CREATE TABLE build AS SELECT (i * 50)::INTEGER AS id, i < 100 AS keep, * EXCLUDE(i) FROM build_payload;
+DROP TABLE build_payload;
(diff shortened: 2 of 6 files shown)


Tmonster:
Hmm, I'm a bit confused as to what scenario this is solving, and how it is solving it. 
From my understanding on the PR and the description, if a build side of a join produces a join_filter for the probe side, the build_side cost goes down? That seems a bit counter-intuitive, because the actual cost of building the hash table will remain the same right?

I looked at the IMDB benchmarks, and according to the results, some queries do look a bit faster, but the logic is not making a lot of sense to me. One thing that confuses me is the use of "bonus", which usually means some kind of value-add, but the line `right_side_build_cost /= DynamicFilterBuildBonus(join, 0, 1, rhs_cardinality);` contradicts that.

One scenario I could see is that you want to prefer a scan for a build side because it can produce a selective join filter for the probe side, so you artificially keep the build side cost low. That would make sense, but you do not know the filter conditions until you build the build side.

Could you add a micro regression to show the performance improvements?

zurferr:
Thanks, added a micro benchmark in `benchmark/micro/join/join_filter_build_side.benchmark`.

It uses a filtered wide build input joined against a large sorted Parquet probe scan, so it exercises the intended case: keeping the filtered side as the hash-build side lets the runtime filter prune the probe scan. Locally, through the regression runner, this goes from 0.255s to 0.124s.

And agreed, the hash table build itself is not cheaper here. The adjustment is a discount for avoided probe-side scan work once the runtime filter is available, not a reduction in hash-build cost. I can also rename the “bonus” terminology to make that clearer.

lnkuiper:
Superseded by https://github.com/duckdb/duckdb/pull/22298 which has been merged.

I can see this PR adds a benchmark file but I think our benchmark suite will already pick up regressions, since the PR that has been merged improves a few queries already in TPC-DS/IMDB.

Issue #22584, opened 2026-05-12
Temporary table doesn't seem to shrink after deletion


dentiny:
### What happens?

```sql
-- Allocate certain bytes for query.
memory D SET memory_limit = '15MB';
memory D SET temp_directory = '/tmp/duckdb_temp_1234';
memory D CREATE TEMP TABLE big1 AS SELECT hash(range) AS c FROM range(135000000);
memory D SELECT count(*) AS file_count, sum(size) AS total_bytes FROM duckdb_temporary_files();
┌────────────┬─────────────┐
│ file_count │ total_bytes │
│   int64    │   int128    │
├────────────┼─────────────┤
│          3 │  1224704000 │
└────────────┴─────────────┘
memory D CREATE TEMP TABLE big2 AS SELECT hash(range) AS c FROM range(135000000);
memory D SELECT count(*) AS file_count, sum(size) AS total_bytes FROM duckdb_temporary_files();
┌────────────┬─────────────┐
│ file_count │ total_bytes │
│   int64    │   int128    │
├────────────┼─────────────┤
│          3 │  2462547968 │
└────────────┴─────────────┘
memory D DROP TABLE big1;
memory D SELECT count(*) AS file_count, sum(size) AS total_bytes FROM duckdb_temporary_files();
┌────────────┬─────────────┐
│ file_count │ total_bytes │
│   int64    │   int128    │
├────────────┼─────────────┤
│          3 │  2462580736 │
└────────────┴─────────────┘
memory D CREATE TEMP TABLE big3 AS SELECT hash(range) AS c FROM range(135000000);
memory D SELECT count(*) AS file_count, sum(size) AS total_bytes FROM duckdb_temporary_files();
┌────────────┬─────────────┐
│ file_count │ total_bytes │
│   int64    │   int128    │
├────────────┼─────────────┤
│          4 │  3493494784 │
└────────────┴─────────────┘
```
Hi team, randomly found this when I was working on extension [table_inspector](https://duckdb.org/community_extensions/extensions/table_inspector)
One thing weird is: after dropping table `big1`, the final size should be similar to `2462547968` since either temp files are deleted, or free blocks are reused, but the total bytes still seem to be 3GiB (the total size of three full tables).

### To Reproduce

Attach the repro sql above.

### OS:

macos

### DuckDB Version:

v1.5-variegata

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

[...]

PR #22616, opened 2026-05-13
Fix free block for temporary file manageer

dentiny:
Related to https://github.com/duckdb/duckdb/issues/22584

Hi team, current code reads to me: even there're free blocks in any of the temporary files, we still allocate new file
But I don't think the current implementation (first-fit) is perfect, see [here](https://github.com/duckdb/duckdb/issues/22584#issuecomment-4428682699); I'm thinking a better way might be iterating through all existing temporary files and find the best fit one, might be a followup PR in the main branch.


Mytherin:
Thanks! Makes sense.

PR #21762, opened 2026-04-01
Unify TableFilters with Expressions

kryonix:
This PR tries to move away from having `TableFilter` as a second, parallel
expression-like structure inside DuckDB.

Historically, pushed-down scan predicates lived in their own `TableFilter`
hierarchy, while the rest of the engine used expressions. That worked, but
it also meant we kept modeling the same ideas twice at some places. Simple
comparisons, `IN` predicates, `IS NULL`, conjunctions, struct-child filters,
and various runtime pruning wrappers all had a `TableFilter` shape even
though most of the engine already knew how to reason about them as expressions.
Over time that created a lot of duplication.

The main point of this PR is to stop doing that. `ExpressionFilter` is now the
single active runtime representation for pushed filters.

For the simple cases, this means scan filters are just ordinary bound
expressions now. For the filter kinds that really do have special scan-time
semantics, the runtime represents them as function calls to internal
`__internal_tablefilter_*` scalar functions. That includes things like
dynamic filters, optional/selectivity-optional filters, bloom filters,
perfect-hash-join filters, and prefix-range filters. These functions are
not meant as user-facing SQL surface area, they are just the mechanism
that lets us preserve the old table-filter behavior while keeping the
engine on one representation. Similar to other internal functions, they
will just throw if you try to call them directly from SQL.

The old `TableFilter` subclasses still exist, but only as a compatibility layer.
They remain useful for reading and writing older serialized plans/filters and
for translating legacy payloads into the new representation. They are no longer
intended to be part of the normal runtime path, and their operational entry
points now fail fast so we do not slowly drift back into maintaining two live
filter systems.

## Why this refactor is useful

At a high level, this change is about removing a long-standing split brain in the filter pipeline.

Before this PR, there was always a boundary where "normal expressions" became
"table filters", and a lot of code existed purely to cross that boundary. That
was especially awkward once filters became more sophisticated: pushdown wanted
to reuse expression logic, optimizer code wanted to derive statistics from the
same predicates, and execution wanted to preserve specialized scan-time behavior.
Keeping all of that in sync across two representations was fragile.

With this refactor, pushed filters behave much more like the rest of DuckDB's
expression system. There is now one main representation flowing through planning,
optimization, pushdown, and execution. The special cases still exist where they
need to exist, but they are expressed explicitly and locally instead of forcing
the whole system to keep a second generic predicate hierarchy alive.

That gives us a cleaner internal model, removes a lot of `TableFilterType`
switching from active code, and makes future work on filter pushdown and
[...]

PR #22617, opened 2026-05-13
Unify remaining TableFilters with ExpressionFilter

kryonix:
This PR extracts the remaining `ExpressionFilter` unification from https://github.com/duckdb/duckdb/pull/21762. It also adds patches for many extensions using `TableFilters`. All `TableFilters` except `ExpressionFilter` are renamed to `Legacy...`, to explicitly mark them as deprecated, so that the compiler will complain. They are only used at de/serialization time now.

Unfortunately, this slice is much bigger than I had hoped for. This is because the remaining cases are much more intertwined than the cases before. So I had the choice of either constructing temporary workarounds (that then would have to be removed in later PRs), or doing more work in one PR.

Done previously:
- [x] `ConstantFilter`
- [x] `InFilter`
- [x] `Is[Not]NullFilter`
- [x] `StructFilter`

Done in this PR:
- [x] `ConjunctionFilter`
- [x] `DynamicFilter`
- [x] `OptionalFilter`
- [x] `BloomFilter`
- [x] `PerfectHashJoinFilter`
- [x] `PrefixRangeFilter`
- [x] `SelectivityOptionalFilter`


Mytherin:
Thanks!

lnkuiper:
This test has some cases that are regressions/correctness issues that should be fixed in a follow-up PR.
```sql
# name: test/temp.test
# description: Temporary reproducers for filter statistics pruning issues
# group: [temp]

statement ok
PRAGMA threads=1

statement ok
SET disabled_optimizers='join_order,build_side_probe_side'

statement ok
SET dynamic_or_filter_threshold=0

# `IS DISTINCT FROM` with a non-null constant must keep NULL rows. The unified
# expression/table-filter stats pruning currently treats all-null stats as
# always false for every comparison, which incorrectly removes this table.
statement ok
CREATE TABLE distinct_from_nulls AS
SELECT NULL::INTEGER AS a
FROM range(10000)

query I
SELECT count(*)
FROM distinct_from_nulls
WHERE a IS DISTINCT FROM 5
----
10000

# The same predicate must not prune a segment whose non-null min/max are exactly
# the constant if the segment can also contain NULL values.
statement ok
CREATE TABLE distinct_from_mixed_nulls AS
SELECT CASE WHEN i % 2 = 0 THEN NULL ELSE 5 END::INTEGER AS a
FROM range(10000) t(i)

query I
SELECT count(*)
FROM distinct_from_mixed_nulls
WHERE a IS DISTINCT FROM 5
----
5000

# Numeric prefix-range filters still use row-group statistics. With two sparse
# build keys, the probe scan should only produce the two matching keys.
statement ok
CREATE TABLE probe_num AS
SELECT i::BIGINT AS k, (i % 7)::INTEGER AS g
FROM range(0, 1000000) t(i)

statement ok
CREATE TABLE build_num AS
SELECT i::BIGINT AS k, (i % 7)::INTEGER AS g
FROM (VALUES (100000), (900000)) v(i)

statement ok
CHECKPOINT

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM probe_num p
JOIN build_num b
  ON p.k = b.k
 AND p.g >= b.g
----
analyzed_plan	<REGEX>:.*IN PRF\(k\).*

query I
SELECT count(*)
FROM probe_num p
JOIN build_num b
  ON p.k = b.k
 AND p.g >= b.g
----
2

# The varchar equivalent should also use string row-group statistics in the PRF
# pruning callback. The current PR regresses this: the plan still contains the
# PRF dynamic filter, but the probe scan produces hundreds of thousands of rows.
statement ok
CREATE TABLE probe_str AS
SELECT lpad(i::VARCHAR, 8, '0') AS k, (i % 7)::INTEGER AS g
FROM range(0, 1000000) t(i)

statement ok
CREATE TABLE build_str AS
SELECT lpad(i::VARCHAR, 8, '0') AS k, (i % 7)::INTEGER AS g
FROM (VALUES (100000), (900000)) v(i)

statement ok
CHECKPOINT

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM probe_str p
JOIN build_str b
  ON p.k = b.k
 AND p.g >= b.g
----
analyzed_plan	<!REGEX>:.*[1-9][0-9]{2},[0-9]{3} rows.*

query I
SELECT count(*)
FROM probe_str p
JOIN build_str b
  ON p.k = b.k
 AND p.g >= b.g
----
2

# Optional runtime filters should still evaluate their child filter when pruning
# constant columns such as hive partition values. The expression-backed optional
# scalar returns TRUE for every value, so all partition files are opened here.
statement ok
CREATE TABLE partition_src AS
SELECT i::INTEGER AS id, (i % 100)::INTEGER AS part
FROM range(0, 100000) t(i)

statement ok
COPY partition_src
TO '{TEMP_DIR}/optional_partition_prune_22617'
(FORMAT PARQUET, PARTITION_BY part, OVERWRITE true)

statement ok
CREATE TABLE build_part AS
SELECT v::INTEGER AS part
FROM (VALUES (42), (43)) t(v)

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM read_parquet('{TEMP_DIR}/optional_partition_prune_22617/*/*.parquet',
                  hive_partitioning=1,
                  hive_types={'part': INTEGER}) p
JOIN build_part b
  ON p.part = b.part
----
analyzed_plan	<!REGEX>:.*Total Files Read: 100.*

query I
SELECT count(*)
FROM read_parquet('{TEMP_DIR}/optional_partition_prune_22617/*/*.parquet',
                  hive_partitioning=1,
                  hive_types={'part': INTEGER}) p
JOIN build_part b
  ON p.part = b.part
----
2000
```

Tishj:
In the future perhaps we should create a window for our in-house maintained extensions to bump their `main` branch and bump the hash in duckdb, so a proper patch can be made as part of the PR introducing the change.

As it stands, the extensions that suffer from this haven't updated their `main` in ages, and now have to merge the `v1.5-variegata` branch into `main` + apply these patches, which will conflict.
Making the work done on these patches almost redundant/wasted in the process.
On top of that, all the work done in `v1.5-variegata` that will conflict with these changes have to be entirely rewritten to work with this new logic.

PR #22620, opened 2026-05-13
Partitioned/Sorted writes

lnkuiper:
This PR implements sorted writes in `PhysicalCopyToFile`. A lot of the infrastructure was already there since https://github.com/duckdb/duckdb/pull/22225.

## Parser/Transformer

Most of the diff is in the `Parser`/`Transformer`. We already have the functionality to pass a parenthesised expression list to a `COPY` option, as this is required for `PARTITION_BY (col1, col2)`. For sorted writes we also need to be able to do `ORDER_BY (col1 ASC NULLS LAST, col2 DESC NULLS FIRST)`. This required adding some new rules to the parser.

The next step is the transformer. I pulled some code out of `transform_attach.cpp` and put it into a separate file `transform_generic_copy_option.cpp` just to clean things up a bit. The parser now generates a `vector<OrderByNode>`, which is just stripped down to expressions again if it the `COPY` option is not `ORDER_BY`. This is the part I'm a bit unsure about @Dtenwolde, not sure how exactly to disambiguate better, because `ORDER_BY (col1, col2)` should use the default order specification.

Once we've transformed the order specification, we `ToString` it again because `COPY` options are required to be `Value`, not `OrderByNode`.

## Binding

I've implemented `ParseOrderByColumns` in `bind_helpers.cpp` to transform the `Value` into a `BoundOrderNode` with `BoundReferenceExpression`s that can be used directly during execution. This is done by adding a generic table binding on a temporarily created `Binder`, binding the columns, which creates `BoundColumnRefExpression`s, which are then converted to `BoundReferenceExpression`s using the column names.

## Execution

The smallest diff is here, since most of the required functionality was already there. The main difference is that we can't keep appending to open partition files because we need to make sure that the file is locally sorted. I've implemented `EnsureFreshPartitionFileForSortedRun` which is only used for sorted writes, which creates a new file once we start appending the next sorted run.

## Testing

I've added a few tests with a low `partitioned_write_flush_threshold` that use `file_name` and `file_row_number` along with the `lag` window function to assert that files are locally sorted.

## Limitations

The `ORDER_BY` option can only be used if `PARTITION_BY` is also used, but that is a temporary restriction that can be lifted in the future.

## Performance

Just a very basic benchmark/comparison:
```sql
-- Partitioned but not sorted: 4 output files, ~4.2s on M1 Max
copy tpch_sf10.lineitem to 'lineitem' (format parquet, partition_by (l_returnflag, l_linestatus));
-- Partitioned AND sorted: 64 output files, ~5.9s on M1 Max
copy tpch_sf10.lineitem to 'lineitem' (format parquet, partition_by (l_returnflag, l_linestatus), order_by (l_shipdate));
```

The performance hit of sorting seems reasonable, but the significant increase in the number of output files caused by having to start writing to a fresh file on every flush is undesirable. I think there will always be some increase in the number of output files, but I think we can do better. I have plans to improve this in a future PR.

CC: @arjenpdevries @gijshendriksen


src/parser/peg/grammar/statements/copy.gram
```diff
@@ -28,7 +28,9 @@ ForceNullOption <- 'FORCE' 'NOT'? 'NULL' ColumnList
 StarSymbol <- '*'
```
Dtenwolde, line 28: `ORDER BY` should also be added as a `SpecializedOption`

Dtenwolde: `OrderByOption <- OrderByClause` This might be enough? Though the transformer rule needs to do some rewriting similar to the `GenericOption` path

lnkuiper:
Daniel and found some parser issues that will be addressed in a future PR

Dtenwolde:
Related: https://github.com/duckdblabs/duckdb-internal/issues/9279 (to be addressed in a follow up PR)

src/parser/peg/grammar/statements/copy.gram
Dtenwolde: To be done in a future PR

Dtenwolde (approved):

PR #22622, opened 2026-05-13
Bump sqlsmith, remove patch

carlopi:
Note: this is to a commit from March 17. There are further developments vs `main`, but that seems to work out of the box on `v1.5-variegata`.

FYI: @c-herrewijn


Mytherin:
Thanks!

PR #22327, opened 2026-04-28
Fuse repeated scalar aggregate scans

zurferr:
This adds a narrow optimizer pass for scalar aggregate subqueries that repeatedly scan the same serializable Parquet input.

The rewrite only fires for uncorrelated, ungrouped scalar aggregates over the same scan/projection shape, with deterministic filters and arguments. The fused plan computes the aggregates in one pass using filtered aggregate expressions.

The pass is intentionally conservative:
- native table scans are skipped before the recursive rewrite walk
- distinct aggregates are not fused
- scan identity is checked by serializing the GET with filters/projections temporarily normalized
- common predicates are kept as regular scan filters so Parquet pruning is preserved

It also includes the small planner simplification for scalar subqueries that are already known to return exactly one row. That exposes aggregate scalar subqueries as plain cross-product branches, which this optimizer can then reason about.

Azure VM validation against current main:
- SF10 ordered Parquet q09: 2.120s -> 1.468s, same output hash
- SF10 ordered Parquet q28: 1.820s -> 1.820s, same output hash; the distinct-count pattern is no longer fused
- SF100 native q09 hot, alternated order: 11.872s main vs 11.909s PR avg, same output hash

Tests:
- `test/optimizer/scalar_aggregate_fusion.test`
- `test/optimizer/scalar_aggregate_fusion_parquet.test`
- `[optimizer]`
- full local release unittest: 4253 test cases, 968058 assertions, 361 skipped

PR #22627, opened 2026-05-13
Two optimizers for TPC-DS Q22/Q88: scalar-aggregate fusion + partial-aggregate pushdown

zurferr:
## Summary

Two new optimizer passes — `SCALAR_AGGREGATE_FUSION` and `PARTIAL_AGGREGATE_PUSHDOWN` — plus a small relaxation of `AggregateFunctionRewriter` to let AVG decompose under `ROLLUP`/`CUBE`/`GROUPING SETS`. The combination closes the planner gap on TPC-DS Q22, Q88 and (mildly) Q92, and is a net **~10 % faster on the full SF10 99-query suite** with no regressions outside noise.

This is the follow-up to two earlier attempts:

- #22327 (closed): SAF on its own. Mark closed it citing a hot-cache regression on Q9 over native tables. Addressed here by a cost-model floor (`NATIVE_TABLE_MIN_ROWS = 100000`) — Parquet/external sources are always eligible, native tables only when they're wide *and* large.
- #22572 (draft): PAP on its own. Mark asked for (a) the 200-line function to be split, (b) the supported-aggregate list to be replaced by the `EXPORT_STATE`/`COMBINE`/`FINALIZE` infrastructure, and (c) `EXPLAIN`/`REGEX` plan-shape tests. All three are done in this branch.

### `SCALAR_AGGREGATE_FUSION`
Fuses N sibling scalar aggregates that read from the same source into a single aggregate node with N conditional aggregates. Catches the Q88 shape:

```sql
SELECT a, b, c, ... FROM
  (SELECT count(*) a FROM store_sales JOIN time_dim ... WHERE t_hour=8 AND ...),
  (SELECT count(*) b FROM store_sales JOIN time_dim ... WHERE t_hour=9 AND ...),
  ... 6 more ...
```

→ one shared scan + join chain, one aggregate with 8 `count_star() FILTER (...)` expressions.

Source-equality is determined by serialising the `LogicalGet` and comparing the binary blob, as `CommonSubplanOptimizer` does (Mark's suggestion on #22327).

### `PARTIAL_AGGREGATE_PUSHDOWN`
Pushes `SUM`/`COUNT`/`MIN`/`MAX`/AVG-state below a comparison join when the aggregate-side is the dominant cardinality and the join doesn't already cut it more than 8× (named `MAX_JOIN_SELECTIVITY_INV`). The upper aggregate consumes the partial state via a new `finalize_combine_aggr` function — same `EXPORT_STATE`/`COMBINE`/`FINALIZE` pipeline already used by `combine_aggr`, so any aggregate that exposes a state-combine callback can be split (gated by `BoundAggregateExpression::HasStateCombineCallback()`).

Q22 (TPC-DS, SF10) plan before:

```
HASH_GROUP_BY (ROLLUP, 5 sets)        ← 8s
  Aggregates: avg(inv_quantity_on_hand)
└── HASH_JOIN inv ⋈ item
    ├── HASH_JOIN inv ⋈ date_dim
    │   └── SEQ_SCAN inventory  (133M rows)
```

After:

```
HASH_GROUP_BY (ROLLUP, 5 sets)
  Aggregates: finalize_combine_aggr(sum_state), finalize_combine_aggr(count_state)
└── HASH_JOIN inv_pre_agg ⋈ item      ← input now ~3M rows
    └── HASH_GROUP_BY (...keys + dim cols...)
        Aggregates: export_aggregate(sum), export_aggregate(count)
        └── HASH_JOIN inv ⋈ date_dim
```

The pre-aggregation collapses the inventory stream from ~80M rows to ~3M rows before the `item` join — the same trick Snowflake's optimizer applies on this query.

### `AggregateFunctionRewriter::AvgRewriteRule`
Previously bailed out for any `multi_grouping_set` aggregate. `SUM(x)/COUNT(x)` is mathematically identical to `AVG(x)` under any grouping (both ignore NULLs identically), so the guard was over-conservative. Decomposing AVG → SUM/COUNT here is the prerequisite for PAP to fire on Q22 (PAP rejects AVG under ROLLUP because the AGGREGATE_STATE blob serialization for AVG loses precision when many partial states combine — so the rewriter sees AVG, splits it into SUM/COUNT, then PAP pushes those exact aggregates down). The remaining guard is for `grouping_functions`, whose per-row grouping-set bitmasks the rewrite has no way to translate.

## Performance — TPC-DS SF10, full 99 queries, single-threaded duckdb shell

| | vanilla `main` | this branch | speedup |
|---|---:|---:|---:|
| Total engine time (sum of 99) | 94.51 s | **85.66 s** | **1.10× (~10 % faster)** |
| Q22 (ROLLUP on inventory) | 7.97 s | **0.84 s** | **9.46×** |
| Q88 (8-way cross-product) | 0.58 s | **0.25 s** | **2.31×** |
| Q92 (correlated subquery) | 0.11 s | 0.10 s | 1.15× (noise) |
| Passing | 99 / 99 | 99 / 99 | — |

Other notable gainers: Q11 (1.40× faster, gross-profit YoY), Q7 (2.23×, item-promotion), Q3 (4.13×). Regressions are all under 0.3 s absolute and within run-to-run noise on a loaded box (largest is +0.30 s on Q14).

Raw per-query CSVs and a longer write-up are in `PARITY_REPORT.md` at the repo root — happy to drop those from the tree before merge if preferred.

## Reviewer feedback addressed

From #22327:
- ✅ New `OptimizerType` enum values + `make generate-files` (bookkeeping in `enum_util.cpp`, `metric_type.cpp`, `profiling_utils.cpp`).
- ✅ Source-equality via `BinarySerializer` instead of pointer / name comparison.
- ✅ Cost-model floor against the Q9 hot-cache regression (`NATIVE_TABLE_MIN_ROWS`).

From #22572:
- ✅ The 200-line `TryPushdownAggregate` is split into `AnalyzePushdown`, `BuildLowerGroupMap`, `BindPushdownAggregates`, `CreateLowerAggregate`, `CreateJoin`, `CreateUpperAggregate` — each under 50 lines.
- ✅ Supported-aggregate gate switched from a hard-coded name list to `BoundAggregateExpression::HasStateCombineCallback()`.
- ✅ `EXPLAIN`/`REGEX` plan-shape tests in `test/optimizer/partial_aggregate_pushdown.test` (positive matches for `finalize_combine_aggr` on baseline + ROLLUP, negative match under high join selectivity) and `test/optimizer/scalar_aggregate_fusion.test` (positive/negative `CROSS_PRODUCT` presence to assert fusion fired or didn't).

## Test plan

- [x] `make` (release, this branch HEAD) — clean build.
- [x] `test/optimizer/scalar_aggregate_fusion.test` — correctness against the no-optimizer baseline, cost-model floor on small tables, optimizer is queryable via `duckdb_optimizers()`.
- [x] `test/optimizer/partial_aggregate_pushdown.test` — correctness against baseline (with vs without PAP), ROLLUP correctness, AVG single-grouping-set, COUNT_STAR (non-pushable, must not crash), selectivity-guard negative check.
- [x] TPC-DS 99-query SF10 — 99/99 pass, numerically identical to baseline (row counts and aggregate sums verified).
- [ ] Full `make unittest` on CI.

## Known limitations

- Q92 is essentially unchanged at this scale. DuckDB's existing `DELIM_JOIN` + `JOIN_FILTER_PUSHDOWN` already does sideways info-passing; what's missing relative to Snowflake is *scan-time* bitmap filtering. That's a separate, larger change to the runtime-filter machinery — deferred.
- `PARITY_REPORT.md`, `bench_q22_q88_q92.sql`, and the two `bench_sf10_*.csv` files at the repo root are perf-validation artifacts. Happy to drop them from the tree pre-merge if upstream prefers PR-body-only.


zurferr:
Closing to reorganize — splitting into smaller, more reviewable PRs.

PR #22594, opened 2026-05-12
Enable/disable jemalloc linking through BUILD/SKIP_EXTENSIONS

Maxxen:
Followup to https://github.com/duckdb/duckdb/pull/22558

This PR also enables controlling the new `ENABLE_JEMALLOC` flag through `BUILD_EXTENSIONS=jemalloc` and `SKIP_EXTENSIONS=jemalloc` for backwards compatibility, but also throws a warning.

PR #22628, opened 2026-05-13
Fix new jemalloc plumbing

mlafeldt:
This fixes two bugs (CMake ordering, missing `JemallocCTL`) I ran into while trying to get jemalloc profiling to work in duckdb-go (https://github.com/duckdblabs/duckdb-internal/issues/9148)

I guess this also means we're missing important CI steps?

Related:

- https://github.com/duckdb/duckdb/pull/22558
- https://github.com/duckdb/duckdb/pull/22594

### DIFF
--- a/.github/workflows/Regression.yml
+++ b/.github/workflows/Regression.yml
@@ -69,7 +69,10 @@ jobs:
       CXX: g++-10
       GEN: ninja
       BUILD_BENCHMARK: 1
-      BUILD_JEMALLOC: 1
+      # Regression compares against the base branch binary. Keep this on the
+      # standard allocator so allocator changes do not look like query
+      # performance regressions.
+      BUILD_JEMALLOC: 0
       CORE_EXTENSIONS: "json;tpch;tpcds;httpfs;inet;icu"
 
     steps:
@@ -360,5 +363,4 @@ jobs:
         shell: bash
         run: |
           python scripts/plan_cost_runner.py --old duckdb/build/release/duckdb --new build/release/duckdb --dir=benchmark/tpch_plan_cost
-
 
--- a/.sanitizer-thread-suppressions.txt
+++ b/.sanitizer-thread-suppressions.txt
@@ -1,4 +1,5 @@
 deadlock:InitializeIndexes
+deadlock:TemplatedSortKeySetPayload
 race:InsertMatchesAndIncrementMisses
 race:NextInnerJoin
 race:NextRightSemiOrAntiJoin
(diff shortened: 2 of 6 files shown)


Maxxen:
This is really strange - I got hit with a bunch of failing CI jobs when drafting the initial PR due to the jemalloc files not linking properly so it does get enabled for sure - but theres no way it should have passed without the missing function...

mlafeldt:
Re CI failure: `duckdb/common/operator/numeric_cast.hpp` says `NOTE: When included directly, this file produces 'unused static method' warnings/errors.`. Yet, that's exactly what `allocator_jemalloc.cpp` does.

Maxxen:
Ah, its the `set(JEMALLOC_ENABLED TRUE)` that needs to be moved up (like you've done in this PR) to actually compile the `allocator_jemalloc.cpp` 🤦 

CC @carlopi This PR actually fixes and links in jemalloc.

carlopi:
Thank you both!!

mlafeldt:
I _think_ this is ready, but the last commit only has the check "Draft on Synchronize / Mark as draft (pull_request)". How to run all checks again?

carlopi:
I think it would be handy to check the failures, or maybe re-run to gather more info.

mlafeldt:
> I _think_ this is ready, but the last commit only has the check "Draft on Synchronize / Mark as draft (pull_request)". How to run all checks again?

I pushed a potential CI fix, but I'm running into this again. Would love to know what to do.

.github/workflows/Regression.yml
```diff
@@ -69,7 +69,10 @@ jobs:
       CXX: g++-10
       GEN: ninja
       BUILD_BENCHMARK: 1
-      BUILD_JEMALLOC: 1
+      # Regression compares against the base branch binary. Keep this on the
+      # standard allocator so allocator changes do not look like query
+      # performance regressions.
+      BUILD_JEMALLOC: 0
```
carlopi, line 75: This I am not sure, in the sense, I would prefer to test vs default allocator (that on Linux is jemalloc).

If that means CI fail on the PR, that's reasonable as well.

carlopi:
> > I _think_ this is ready, but the last commit only has the check "Draft on Synchronize / Mark as draft (pull_request)". How to run all checks again?
> 
> I pushed a potential CI fix, but I'm running into this again. Would love to know what to do.

Could you move to draft and then wait a minute, then mark as ready? That should retrigger CI (I am not sure what's happening, but that should work).

mlafeldt:
> /usr/bin/ld: final link failed: No space left on device

Rebased since I can't re-run failed jobs. Now doing the draft / undraft dance again.

Mytherin:
Thanks!

PR #22632, opened 2026-05-13
Add dummy cmake target for jemalloc

evertlammerts:
Follow up on #22558 and #22594

In python we call `target_link_libraries(${target_name} PRIVATE ${ext}_extension)` for every extension we link in. For jemalloc that target does not exist anymore, so builds fail. By adding a dummy target we make sure that this won't break other consumers.


carlopi:
Failure is irrelevant, it's due to networking randomness.

Mytherin:
Thanks!

PR #22639, opened 2026-05-13
Backport windows-32bit test fix to 1.4

staticlibs:
Backporting a test fix #21432 to `v1.4-andium` just in case.

Ref: duckdblabs/duckdb-internal#8404


Mytherin:
Thanks!

PR #21112, opened 2026-02-27
Properly update nested validity stats

lnkuiper:
Fixes https://github.com/duckdb/duckdb/issues/21100

PR #22645, opened 2026-05-13
Fix tablefilter unification issues

kryonix:
Follow up of https://github.com/duckdb/duckdb/pull/22617 fixing all issues reported in https://github.com/duckdb/duckdb/pull/22617#issuecomment-4441306489.

### DIFF
--- a/src/execution/operator/join/physical_hash_join.cpp
+++ b/src/execution/operator/join/physical_hash_join.cpp
@@ -36,6 +36,7 @@
 #include "duckdb/storage/buffer_manager.hpp"
 #include "duckdb/storage/temporary_memory_manager.hpp"
 #include "duckdb/planner/expression_iterator.hpp"
+#include "duckdb/common/types/row/tuple_data_iterator.hpp"
 
 namespace duckdb {
 
@@ -1101,8 +1102,12 @@ bool JoinFilterPushdownInfo::CanUseBloomFilter(const ClientContext &context, con
 	// building the bloom filter is costly on the build to make probing faster,
 	// so only use it if there are less build tuples than probing tuples
 	static constexpr double BUILD_TO_PROBE_RATIO_THRESHOLD = 1.0;
+	auto build_count = ht->Count();
+	if (build_count == 0) {
+		build_count = ht->GetSinkCollection().Count();
+	}
 	const double build_to_probe_ratio =
-	    static_cast<double>(ht->Count()) / static_cast<double>(op.children[0].get().estimated_cardinality);
+	    static_cast<double>(build_count) / static_cast<double>(op.children[0].get().estimated_cardinality);
 	if (build_to_probe_ratio > BUILD_TO_PROBE_RATIO_THRESHOLD) {
 		return false;
 	}
@@ -1111,7 +1116,7 @@ bool JoinFilterPushdownInfo::CanUseBloomFilter(const ClientContext &context, con
 	static constexpr double NON_FILTERING_RATIO_THRESHOLD = 0.1;
 	static constexpr idx_t NON_FILTERING_BUILD_SIDE_THRESHOLD = 4194304;
 	if (!build_side_has_filter && build_to_probe_ratio > NON_FILTERING_RATIO_THRESHOLD &&
-	    ht->Count() > NON_FILTERING_BUILD_SIDE_THRESHOLD) {
+	    build_count > NON_FILTERING_BUILD_SIDE_THRESHOLD) {
 		return false;
 	}
 
@@ -1122,6 +1127,9 @@ bool JoinFilterPushdownInfo::CanUsePrefixRangeFilter(ClientContext &context, opt
                                                      const PhysicalComparisonJoin &op, const ExpressionType &cmp,
                                                      const Value &min, const Value &max) const {
 	if (!CanUseBloomFilter(context, op, cmp, ht)) {
+		return false;
+	}
+	if (ht->Count() == 0) {
 		return false;
 	}
 
@@ -1174,6 +1182,34 @@ static unique_ptr<Expression> CreateRuntimeFilterInputExpression(ClientContext &
 	return input;
 }
 
+static void BuildBloomFilterFromPartitionedData(JoinHashTable &ht) {
+	auto &bloom_filter = ht.GetBloomFilter();
+	if (bloom_filter.IsInitialized()) {
+		return;
+	}
+	auto &sink_collection = ht.GetSinkCollection();
+	if (sink_collection.Count() == 0) {
+		return;
+	}
+	bloom_filter.Initialize(ht.context, sink_collection.Count());
+	Vector hashes(LogicalType::HASH);
+	auto hash_data = FlatVector::GetDataMutable<hash_t>(hashes);
+	for (auto &partition : sink_collection.GetPartitions()) {
+		if (!partition || partition->Count() == 0) {
+			continue;
+		}
+		TupleDataChunkIterator iterator(*partition, TupleDataPinProperties::KEEP_EVERYTHING_PINNED, false);
+		do {
+			auto count = iterator.GetCurrentChunkCount();
+			auto row_locations = iterator.GetRowLocations();
+			for (idx_t i = 0; i < count; i++) {
+				hash_data[i] = Load<hash_t>(row_locations[i] + ht.pointer_offset);
+			}
+			bloom_filter.InsertHashes(hashes, count);
+		} while (iterator.Next());
+	}
+}
+
 void JoinFilterPushdownInfo::PushBloomFilter(ClientContext &context, const PhysicalOperator &op, JoinHashTable &ht,
                                              const JoinFilterPushdownFilter &info, idx_t filter_idx,
                                              ProjectionIndex filter_col_idx) const {
@@ -1183,6 +1219,9 @@ void JoinFilterPushdownInfo::PushBloomFilter(ClientContext &context, const Physi
 	const auto key_type = ht.conditions[0].GetLHS().GetReturnType();
 	auto filter_input_type = GetRuntimeFilterInputType(info.columns[filter_idx], key_type);
 	ht.SetBuildBloomFilter(true);
+	if (ht.Count() == 0) {
+		BuildBloomFilterFromPartitionedData(ht);
+	}
 	float selectivity_threshold;
 	idx_t n_vectors_to_check;
 	GetThresholdAndVectorsToCheck(SelectivityOptionalFilterType::BF, selectivity_threshold, n_vectors_to_check);
@@ -1468,6 +1507,10 @@ SinkFinalizeType PhysicalHashJoin::Finalize(Pipeline &pipeline, Event &event, Cl
 			}
 			sink.local_hash_tables.clear();
 			sink.owned_local_hash_tables.clear();
+			if (filter_pushdown && !sink.skip_filter_pushdown && ht.GetSinkCollection().Count() > 0) {
+				auto filter_min_max = filter_pushdown->FinalizeMinMax(*sink.global_filter_state);
+				filter_pushdown->FinalizeFilters(context, *this, std::move(filter_min_max), &ht, nullptr);
+			}
 			D_ASSERT(sink.temporary_memory_state->GetReservation() >= sink.probe_side_requirement);
 			sink.hash_table->PrepareExternalFinalize(sink.temporary_memory_state->GetReservation() -
 			                                         sink.probe_side_requirement);
--- a/src/execution/operator/join/physical_hash_join.cpp
+++ b/src/execution/operator/join/physical_hash_join.cpp
@@ -24,6 +24,7 @@
 #include "duckdb/parallel/pipeline.hpp"
 #include "duckdb/parallel/thread_context.hpp"
 #include "duckdb/planner/expression/bound_aggregate_expression.hpp"
+#include "duckdb/planner/expression/bound_cast_expression.hpp"
 #include "duckdb/planner/expression/bound_comparison_expression.hpp"
 #include "duckdb/planner/expression/bound_constant_expression.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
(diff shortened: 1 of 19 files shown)


lnkuiper:
Looks great! Can merge like this, but I think I need another follow up (or add to this PR) a fix for:
```sql
# name: test/temp.test
# description: Scratch regressions for join filter pushdown through casts
# group: [optimizer]

statement ok
PRAGMA threads=1

statement ok
SET disabled_optimizers='join_order,build_side_probe_side'

statement ok
SET dynamic_or_filter_threshold=0

# PR #21791 made non-min/max runtime join filters work when the probe-side
# storage type needs to be cast to the hash-table key type. PR #22645 currently
# keeps the min/max filters but drops the PRF in this shape.
statement ok
CREATE TABLE probe_cast_up AS
SELECT i::INTEGER AS k, (i % 7)::INTEGER AS g
FROM range(0, 1000000) t(i)

statement ok
CREATE TABLE build_cast_up AS
SELECT i::BIGINT AS k, (i % 7)::INTEGER AS g
FROM (VALUES (100000), (900000)) v(i)

statement ok
CHECKPOINT

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM probe_cast_up p
JOIN build_cast_up b
  ON p.k = b.k
 AND p.g >= b.g
----
analyzed_plan	<REGEX>:.*IN PRF\(k\).*

query I
SELECT count(*)
FROM probe_cast_up p
JOIN build_cast_up b
  ON p.k = b.k
 AND p.g >= b.g
----
2

# The reverse cast direction still has the PRF. This keeps the test from being
# only a generic "PRF disabled" check.
statement ok
CREATE TABLE probe_cast_down AS
SELECT i::BIGINT AS k, (i % 7)::INTEGER AS g
FROM range(0, 1000000) t(i)

statement ok
CREATE TABLE build_cast_down AS
SELECT i::INTEGER AS k, (i % 7)::INTEGER AS g
FROM (VALUES (100000), (900000)) v(i)

statement ok
CHECKPOINT

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM probe_cast_down p
JOIN build_cast_down b
  ON p.k = b.k
 AND p.g >= b.g
----
analyzed_plan	<REGEX>:.*IN PRF\(CAST\(k AS BIGINT\)\).*

query I
SELECT count(*)
FROM probe_cast_down p
JOIN build_cast_down b
  ON p.k = b.k
 AND p.g >= b.g
----
2

# Perfect hash join filters have the same cast-up gap. This query has only a
# single equality condition and a small build range so PHJ is preferred over PRF
# or BF when the filter is allowed through.
statement ok
CREATE TABLE phj_probe_cast_up AS
SELECT i::INTEGER AS k
FROM range(0, 1000000) t(i)

statement ok
CREATE TABLE phj_build_cast_up AS
SELECT i::BIGINT AS k
FROM range(100000, 100010) t(i)

statement ok
CHECKPOINT

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM phj_probe_cast_up p
JOIN phj_build_cast_up b
  ON p.k = b.k
----
analyzed_plan	<REGEX>:.*IN PHJ\(k\).*

query I
SELECT count(*)
FROM phj_probe_cast_up p
JOIN phj_build_cast_up b
  ON p.k = b.k
----
10

statement ok
CREATE TABLE phj_probe_cast_down AS
SELECT i::BIGINT AS k
FROM range(0, 1000000) t(i)

statement ok
CREATE TABLE phj_build_cast_down AS
SELECT i::INTEGER AS k
FROM range(100000, 100010) t(i)

statement ok
CHECKPOINT

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM phj_probe_cast_down p
JOIN phj_build_cast_down b
  ON p.k = b.k
----
analyzed_plan	<REGEX>:.*IN PHJ\(CAST\(k AS BIGINT\)\).*

query I
SELECT count(*)
FROM phj_probe_cast_down p
JOIN phj_build_cast_down b
  ON p.k = b.k
----
10

# Bloom filters need a larger sparse build side so PRF is not selected first.
# The residual inequality also prevents using the perfect hash join path.
statement ok
CREATE TABLE bf_probe_cast_up AS
SELECT i::INTEGER AS k, (i % 7)::INTEGER AS g
FROM range(0, 2000000) t(i)

statement ok
CREATE TABLE bf_build_cast_up AS
SELECT (100000 + i * 3)::BIGINT AS k, ((100000 + i * 3) % 7)::INTEGER AS g
FROM range(0, 600000) t(i)

statement ok
CHECKPOINT

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM bf_probe_cast_up p
JOIN bf_build_cast_up b
  ON p.k = b.k
 AND p.g >= b.g
----
analyzed_plan	<REGEX>:.*IN BF\(k\).*

query I
SELECT count(*)
FROM bf_probe_cast_up p
JOIN bf_build_cast_up b
  ON p.k = b.k
 AND p.g >= b.g
----
600000

statement ok
CREATE TABLE bf_probe_cast_down AS
SELECT i::BIGINT AS k, (i % 7)::INTEGER AS g
FROM range(0, 2000000) t(i)

statement ok
CREATE TABLE bf_build_cast_down AS
SELECT (100000 + i * 3)::INTEGER AS k, ((100000 + i * 3) % 7)::INTEGER AS g
FROM range(0, 600000) t(i)

statement ok
CHECKPOINT

query II
EXPLAIN ANALYZE
SELECT count(*)
FROM bf_probe_cast_down p
JOIN bf_build_cast_down b
  ON p.k = b.k
 AND p.g >= b.g
----
analyzed_plan	<REGEX>:.*IN BF\(CAST\(k AS BIGINT\)\).*

query I
SELECT count(*)
FROM bf_probe_cast_down p
JOIN bf_build_cast_down b
  ON p.k = b.k
 AND p.g >= b.g
----
600000
```

src/execution/operator/join/physical_hash_join.cpp
```diff
@@ -1157,53 +1166,105 @@ static unique_ptr<ExpressionFilter> CreateSelectivityOptionalExpressionFilter(un
...
+	if (bloom_filter.IsInitialized()) {
+		return;
+	}
+	auto &sink_collection = ht.GetSinkCollection();
+	if (sink_collection.Count() == 0) {
+		return;
+	}
+	bloom_filter.Initialize(ht.context, sink_collection.Count());
+	Vector hashes(LogicalType::HASH);
+	auto hash_data = FlatVector::GetDataMutable<hash_t>(hashes);
```
Mytherin: Let's use `FlatVector::Writer<hash_t>` here

src/storage/table/standard_column_data.cpp
```diff
@@ -23,6 +25,42 @@ void StandardColumnData::SetDataType(ColumnDataType data_type) {
 	validity->SetDataType(data_type);
 }
 
+FilterPropagateResult StandardColumnData::CheckZonemap(ColumnScanState &state, TableFilter &filter) {
```
Mytherin: This looks like a complete copy of `ColumnData::CheckZonemap` - is this necessary?

src/execution/operator/join/physical_hash_join.cpp
```diff
@@ -1157,53 +1166,105 @@ static unique_ptr<ExpressionFilter> CreateSelectivityOptionalExpressionFilter(un
...
+		do {
+			auto count = iterator.GetCurrentChunkCount();
+			auto row_locations = iterator.GetRowLocations();
+			for (idx_t i = 0; i < count; i++) {
+				hash_data[i] = Load<hash_t>(row_locations[i] + ht.pointer_offset);
+			}
+			bloom_filter.InsertHashes(hashes, count);
+		} while (iterator.Next());
+	}
+}
```
lnkuiper: I think this should still be done during the build in `JoinHashTable`, that saves us from iterating over the collection hashes a second time. I think this change might be the reason that a few tests are OOMing in CI.

kryonix: Good point, I changed that part accordingly.

src/execution/operator/join/physical_hash_join.cpp
kryonix: I removed the function and implemented it another way, so the code is gone entirely now.

src/storage/table/standard_column_data.cpp
kryonix: Let me revisit that. Maybe there is another solution. I did that, because of a segment-level pruning bug that I encountered. After checkpoint, the base segment stats for a column with a lot of NULLs looked like constant 5, no nulls, while the actual NULL information lived in the separate validity child. That made a IS DISTINCT FROM 5 look like FILTER_ALWAYS_FALSE, so RowGroup::CheckZonemapSegments() skipped the first four vectors and only scanned the tail 904 rows. I fixed this by overriding StandardColumnData::CheckZonemap() to evaluate pruning against merged base + validity-child segment stats (and the merged update stats path), which restores the correct semantics.

kryonix: I think it's much better now.

src/storage/table/column_data.cpp
```diff
@@ -46,6 +46,14 @@ static bool IsDirectNullCheckFilter(const TableFilter &filter) {
 	return op.children[0]->GetExpressionClass() == ExpressionClass::BOUND_REF;
 }
 
+ColumnData::ZonemapCheckStats ColumnData::GetSegmentStatsForZonemap(ColumnScanState &state, TableFilter &filter) {
+	auto &segment_stats =
+	    IsDirectNullCheckFilter(filter) && !state.child_states.empty() && state.child_states[0].current
```
Mytherin: I'm a bit confused as to why this is required - nullness should also be set in the stats for the non-validity segment, no? Why do we need to check the validity segment for these types of filters?

It also just seems kind of hacky - a different function may also require null-ness information, this is just hardcoding a solution for `x IS NULL`.

kryonix: Let me just rip this out and check again.

kryonix: CI is still running, but this will fail now:

```
./build/release/test/unittest --test-config test/configs/force_storage.json "test/sql/optimizer/*"                                                              
Filters: test/sql/optimizer/*
[23/45] (51%): test/sql/optimizer/table_filters.test
1. test/sql/optimizer/table_filters.test:36
Wrong result in query! (test/sql/optimizer/table_filters.test:36)!
SELECT count(*)
FROM distinct_from_mixed_nulls
WHERE a IS DISTINCT FROM 5;
Mismatch on row 1, column count_star()(index 1)
904 <> 5000
Expected result:
5000

Actual result:
904
```

Mytherin: I think that's related to this fix that Peter pushed here - https://github.com/duckdb/duckdb/pull/22678#pullrequestreview-4306197222

kryonix: I debugged this further. The Callstack is:
```
TryGetFilterStats
CheckComparisonStatistics
CheckFunctionStatistics
ExpressionFilter::CheckExpressionStatistics
ExpressionFilter::CheckStatistics
ColumnData::CheckZonemap                     
```

All of these functions are called with the stats from `ColumnData::CheckZonemap`. The expression going into `TryGetFilterStats` is a `BOUND_REF`, for which the stats are returned unchanged. However, in these stats, `has_null` is false. Hence, `FILTER_ALWAYS_FALSE` is returned, which is wrong. This however only happens after checkpoint. So somewhere in this stack, we are doing something wrong.

[aeb2de4](https://github.com/duckdb/duckdb/pull/22645/commits/aeb2de4760544e3c890ecdb63edcb3145321a34a) Patches this locally in `CheckComparisonStatistics` — but I'm not sure if that's really a proper solution, or just a patch.

lnkuiper: This sounds like a problem in the correctness of NULL statistics. The statistics should be correct - no band-aid in the pruning logic. I remember fixing a NULL statistics issue in https://github.com/duckdb/duckdb/pull/21112 for nested types, maybe that's related and can help with figuring out the problem?

Alternatively, is this fix for a test that was also added in this PR? We could consider taking the test/fix out of this PR and creating an internal issue for it if that's the case so this PR becomes unblocked.

Mytherin: The stats should be complete at this point - if they aren't we're calling this with the wrong stats somehow. You can check per-segment stats using `PRAGMA storage_info('...')`.

I do think the logic in `CheckComparisonStatistics` is just incorrect for `DISTINCT FROM`, however.

```cpp
	if (constant.IsNull()) {
		return FilterPropagateResult::NO_PRUNING_POSSIBLE;
	}
	if (!filter_stats->CanHaveNoNull()) {
		return FilterPropagateResult::FILTER_ALWAYS_FALSE;
	}
```

In this branch the constant is a non-NULL value (e.g. `5`), and the stats specify the column only has `NULL` values. `NULL IS DISTINCT FROM 5` is `true`, not `false`, hence the fix that was in that patch.

kryonix: `PRAGMA storage_info` shows this:

```
memory D CREATE TABLE distinct_from_mixed_nulls AS
         SELECT CASE WHEN i % 2 = 0 THEN NULL ELSE 5 END::INTEGER AS a
         FROM range(10000) t(i);
memory D PRAGMA storage_info('distinct_from_mixed_nulls');
┌──────────────┬─────────────┬───────────┬─────────────┬────────────┬──────────────┬───────┬───────┬──────────────┬─────────────────────────────────────────────────────────────┬─────────────┬────────────┬──────────┬──────────────┬──────────────┬──────────────────────┐
│ row_group_id │ column_name │ column_id │ column_path │ segment_id │ segment_type │ start │ count │ compression  │                            stats                            │ has_updates │ persistent │ block_id │ block_offset │ segment_info │ additional_block_ids │
│    int64     │   varchar   │   int64   │   varchar   │   int64    │   varchar    │ int64 │ int64 │   varchar    │                           variant                           │   boolean   │  boolean   │  int64   │    int64     │   varchar    │       int64[]        │
├──────────────┼─────────────┼───────────┼─────────────┼────────────┼──────────────┼───────┼───────┼──────────────┼─────────────────────────────────────────────────────────────┼─────────────┼────────────┼──────────┼──────────────┼──────────────┼──────────────────────┤
│            0 │ a           │         0 │ [0]         │          0 │ INTEGER      │     0 │   510 │ Uncompressed │ {'has_no_null': true, 'has_null': true, 'max': 5, 'min': 5} │ false       │ false      │     NULL │         NULL │              │ NULL                 │
│            0 │ a           │         0 │ [0]         │          1 │ INTEGER      │   510 │  1022 │ Uncompressed │ {'has_no_null': true, 'has_null': true, 'max': 5, 'min': 5} │ false       │ false      │     NULL │         NULL │              │ NULL                 │
│            0 │ a           │         0 │ [0]         │          2 │ INTEGER      │  1532 │  2046 │ Uncompressed │ {'has_no_null': true, 'has_null': true, 'max': 5, 'min': 5} │ false       │ false      │     NULL │         NULL │              │ NULL                 │
│            0 │ a           │         0 │ [0]         │          3 │ INTEGER      │  3578 │  4094 │ Uncompressed │ {'has_no_null': true, 'has_null': true, 'max': 5, 'min': 5} │ false       │ false      │     NULL │         NULL │              │ NULL                 │
│            0 │ a           │         0 │ [0]         │          4 │ INTEGER      │  7672 │  2328 │ Uncompressed │ {'has_no_null': true, 'has_null': true, 'max': 5, 'min': 5} │ false       │ false      │     NULL │         NULL │              │ NULL                 │
│            0 │ a           │         0 │ [0, 0]      │          0 │ VALIDITY     │     0 │ 10000 │ Uncompressed │ {'has_no_null': true, 'has_null': true}                     │ false       │ false      │     NULL │         NULL │              │ NULL                 │
└──────────────┴─────────────┴───────────┴─────────────┴────────────┴──────────────┴───────┴───────┴──────────────┴─────────────────────────────────────────────────────────────┴─────────────┴────────────┴──────────┴──────────────┴──────────────┴──────────────────────┘
memory D checkpoint;
memory D PRAGMA storage_info('distinct_from_mixed_nulls');
┌──────────────┬─────────────┬───────────┬─────────────┬────────────┬──────────────┬───────┬───────┬──────────────┬──────────────────────────────────────────────────────────────┬─────────────┬────────────┬──────────┬──────────────┬──────────────┬──────────────────────┐
│ row_group_id │ column_name │ column_id │ column_path │ segment_id │ segment_type │ start │ count │ compression  │                            stats                             │ has_updates │ persistent │ block_id │ block_offset │ segment_info │ additional_block_ids │
│    int64     │   varchar   │   int64   │   varchar   │   int64    │   varchar    │ int64 │ int64 │   varchar    │                           variant                            │   boolean   │  boolean   │  int64   │    int64     │   varchar    │       int64[]        │
├──────────────┼─────────────┼───────────┼─────────────┼────────────┼──────────────┼───────┼───────┼──────────────┼──────────────────────────────────────────────────────────────┼─────────────┼────────────┼──────────┼──────────────┼──────────────┼──────────────────────┤
│            0 │ a           │         0 │ [0]         │          0 │ INTEGER      │     0 │ 10000 │ Constant     │ {'has_no_null': true, 'has_null': false, 'max': 5, 'min': 5} │ false       │ true       │       -1 │            0 │              │ []                   │
│            0 │ a           │         0 │ [0, 0]      │          0 │ VALIDITY     │     0 │ 10000 │ Uncompressed │ {'has_no_null': true, 'has_null': true}                      │ false       │ true       │       -1 │            0 │              │ []                   │
└──────────────┴─────────────┴───────────┴─────────────┴────────────┴──────────────┴───────┴───────┴──────────────┴──────────────────────────────────────────────────────────────┴─────────────┴────────────┴──────────┴──────────────┴──────────────┴──────────────────────┘
```

This matches the debugging results and explains why `filter_stats->CanHaveNull()` did not trigger, i.e., why we returned `auto result = CheckZonemapAgainstConstants(*filter_stats, comparison_type, array_ptr<const Value>(&constant, 1));`, instead of `FilterPropagateResult::NO_PRUNING_POSSIBLE`.

Mytherin: It seems the stats are then gathered incorrectly for constant vectors - we should probably fix that

kryonix: I tried to narrow this down a bit more. Seems like the statistics are only wrong when RLE is used. I found a minimal fix: [d93b2df](https://github.com/duckdb/duckdb/pull/22645/commits/d93b2df08310bbcd5272365f0342c2791524a1aa) but CI also passes without this commit. So we could also split this into a separate PR instead.

src/planner/filter/expression_filter.cpp
```diff
@@ -257,13 +257,11 @@ static FilterPropagateResult CheckComparisonStatistics(const BoundFunctionExpres
...
-	if (comparison_type == ExpressionType::COMPARE_DISTINCT_FROM &&
-	    result == FilterPropagateResult::FILTER_ALWAYS_FALSE) {
-		// Mixed-null checkpointed segments can still reach this path with value stats that only capture the
-		// non-null payload. Avoid claiming the filter is always false unless another layer proves away NULLs.
-		return FilterPropagateResult::NO_PRUNING_POSSIBLE;
-	}
 	if (filter_stats->CanHaveNull()) {
+		if (comparison_type == ExpressionType::COMPARE_DISTINCT_FROM &&
+		    result == FilterPropagateResult::FILTER_ALWAYS_FALSE) {
+			return FilterPropagateResult::NO_PRUNING_POSSIBLE;
```
Mytherin: This seems like a work-around for a bug still

src/storage/table/column_data.cpp
Mytherin: Cool - that's indeed the correct fix

Mytherin (commented):
Thanks - looks one more comment

src/planner/filter/expression_filter.cpp
kryonix: Ah right. That's too much of a revert.

Mytherin:
Thanks!

lnkuiper:
`test/sql/join/external/external_join_many_duplicates.test_slow` failed in the merge queue, smells like some kind of merge conflict with https://github.com/duckdb/duckdb/pull/22755 which was merged in the queue earlier?

kryonix:
I'll rebase to main and try to reproduce and fix the error.

PR #22647, opened 2026-05-13
Arithmetic filter pushdown

Damon07:
Filters like `col * 0.3 > 5.0` are pushed down but cannot leverage row group min/max statistics. This PR adds `TryPushdownArithmeticFilter` which injects an additional `OptionalFilter` with a relaxed column bound (e.g., col > 16.66), enabling zonemap to skip irrelevant row groups.


Mytherin:
Thanks - but I think this implementation is in the wrong place. This should be part of the functions themselves. Can this be implemented using the `filter_prune` callback in the specific functions?

Damon07:
Thanks for review, that makes sense.

Damon07:
Hi @Mytherin, after looking through the code, I'm a bit confused about the suggestion of using the filter_prune callback. For filter `a * 3 < 10` will be bound and planned like this `<( *(a, 3), 10 )`, the outer is the comparison. `*` is not filter function so I think I need to implement the `function_statistics_t` callback for function `*` instead, e.g. for a given zonemap `[min: 5, max: 10]`, `*(a, 3)` will be in range `[min: 15, max: 30]`. The calculated output stats will be used in the comparison expression. So pruning this filter would require recursively deriving statistics for the comparison's child expressions, including deriving output statistics for `*(a, 3)`. While the `function_statistics_t` callback needs `ClientContext` to be passed in which is not accessible in `ExpressionFilter::CheckExpressionStatistics` call chain. So it seems like there is no way to call `function_statistics_t` in row group pruning if I wasn't missing anything.

Mytherin:
Yes that sounds like the right way to do it. I think that would just require some plumbing to get the `ClientContext` through - there is no particular reason that's not passed in right now other than the fact that it wasn't necessary in the past.

PR #22648, opened 2026-05-13
Fix null sentinel for keys in dict vector (Arrow)

evertlammerts:
`ColumnArrowToDuckDBDictionary` in src/function/table/arrow_conversion.cpp started throwing `Vector::SetSize out of range - trying to set size to [N] for vector with capacity [N-1]` for dictionary-encoded Arrow data. The root cause is that `FlatVector::SetData` swaps the buffer for one whose capacity equals the count (= dict_length). This was causing tests in duckdb-python to fail (in test_dictionary_arrow.py and test_arrow_extensions.py).


Mytherin:
Thanks!

PR #22654, opened 2026-05-13
fix: list_zip SEGFAULT with empty / NULL argument

fivetran-kwoodbeck:
When `list_zip` is called with an empty or `NULL` argument whose element type is nested (`STRUCT, VARIANT, ARRAY`), the function can **segfault** and crash the DuckDB process. The crash is because the selection vector referenced `index 0` in a zero-length list vector, i.e. there was **no backing memory**.

Fix: When a list has no data, set the corresponding result struct child to a constant-null vector before calling `Flatten`.
`SetVectorType(CONSTANT_VECTOR) + ConstantVector::SetNull` recursively propagates the `NULL` through all nested struct children.

Example of the problem:
```
% duckdb
DuckDB v1.5.2 (Variegata)
memory D SELECT MAP_FROM_ENTRIES(LIST_ZIP(CAST(['123'] AS VARIANT[]), CAST([] AS VARIANT[])));
zsh: segmentation fault  duckdb
```

### DIFF
--- a/src/function/scalar/list/list_zip.cpp
+++ b/src/function/scalar/list/list_zip.cpp
@@ -112,7 +112,11 @@ static void ListZipFunction(DataChunk &args, ExpressionState &state, Vector &res
 		offset += len;
 	}
 	for (idx_t child_idx = 0; child_idx < args_size; child_idx++) {
-		if (args.data[child_idx].GetType() != LogicalType::SQLNULL) {
+		if (args.data[child_idx].GetType() == LogicalType::SQLNULL ||
+			ListVector::GetListSize(args.data[child_idx]) == 0) {
+			struct_entries[child_idx]->SetVectorType(VectorType::CONSTANT_VECTOR);
+			ConstantVector::SetNull(*struct_entries[child_idx], true);
+		} else {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - could you retarget this fix to `v1.5-variegata` so this can be shipped in the bug-fix release?

fivetran-kwoodbeck:
> Thanks - could you retarget this fix to `v1.5-variegata` so this can be shipped in the bug-fix release?

@Mytherin No problem, I retargeted it.  Let me know if anything else is needed!

carlopi:
This fails CI in [Main / Tests a release build with different configurations (pull_request)](https://github.com/duckdb/duckdb/actions/runs/25918999135/job/76477610744?pr=22654).

I have https://github.com/duckdb/duckdb/pull/22726 up, that it's the same commits but removing the `VARIANT` part of the test. It could be added again as a separate PR, while removing it from relevant configs.

lnkuiper:
Superseded by https://github.com/duckdb/duckdb/pull/22726

Thanks again for the fix! We're keeping your commits. We needed a small change but can't wait for review/implementation because we want to build the release binaries soon!

PR #22658, opened 2026-05-14
Fix parquet stats for enum types

### LINKED ISSUE #22548: Parquet min/max stats incorrect for enum
### What happens?

```sql
memory D CREATE TYPE myenum AS ENUM('a', 'b', 'c');
memory D CREATE TABLE t AS SELECT 'a'::myenum AS m FROM range(10) t(i);
memory D COPY t TO '/tmp/enum.parquet';
memory D SELECT stats_min, stats_max FROM parquet_metadata('/tmp/enum.parquet');
┌───────────┬───────────┐
│ stats_min │ stats_max │
│  varchar  │  varchar  │
├───────────┼───────────┤
│ a         │ c         │
└───────────┴───────────┘
-- Both min and max should be 'a'
```
Enum columns don't populate stats correctly.

### To Reproduce

Attached repro SQL statements above.

### OS:

macos

### DuckDB Version:

01eda16d6e3a6648accda27230b9e30edd8b9703

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22548

In the current implementation we updates stats with all possible enum values, fixed by only updating min/max after all enum rows written. ~~To avoid frequently string comparison, lex ordering are pre-computed in advance.~~ Boolean array is used to indicate whether an enum type exists or not.

### DIFF
--- a/extension/parquet/include/writer/enum_column_writer.hpp
+++ b/extension/parquet/include/writer/enum_column_writer.hpp
@@ -34,6 +34,10 @@ public:
 	~EnumColumnWriter() override = default;
 
 	uint32_t bit_width;
+	// Maps enum index to rank in lexicographic order over the enum's string values.
+	// Precomputed once so the on enum value write hot-path, we could compare integer instead of string comparison when
+	// finding the chunk's lex min/max.
+	vector<uint32_t> lex_rank;
 
 public:
 	unique_ptr<ColumnWriterStatistics> InitializeStatsState() override;
--- a/extension/parquet/include/writer/enum_column_writer.hpp
+++ b/extension/parquet/include/writer/enum_column_writer.hpp
@@ -33,13 +33,6 @@ public:
 	EnumColumnWriter(ParquetWriter &writer, ParquetColumnSchema &&column_schema, vector<string> schema_path_p);
 	~EnumColumnWriter() override = default;
 
-	uint32_t bit_width;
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks for the PR!

So the stats are not necessarily incorrect - just too wide. This is allowed by the Parquet spec provided we are setting `is_min_value_exact = false / is_max_value_exact = false`. I think for `v1.5-variegata` we can provide a more minimal fix by just setting those values to false always for enums.

As for writing exact stats - I think a simpler solution would be to just keep a `vector<bool> seen_enum` that is initialized to false for each enum, setting `seen_enum[enum_id] = true` for every value we write. We can then just call `stats.Update()` once for each enum value we have seen, instead of for all values in the enum. The current solution seems much more involved (and likely also slower). Maybe we can push that fix to `main`?

Mytherin:
I'm not sure on the exact order of operations - but we don't need to emit unseen enums in the dictionary either. Perhaps for unseen enums we can instead flush a `NULL` value or an empty string in `FlushDictionary`?'

Edit: I think an empty string would be easiest / best.

Mytherin:
Thanks!

PR #22659, opened 2026-05-14
Bump quack

carlopi:
On top of https://github.com/duckdb/duckdb/pull/22631, but given tests, expecially on Windows, might be unstable, better skip them while investigating.


Mytherin:
Thanks!

Issue #22661, opened 2026-05-14
Query MIN(file_row_number) crashes


dentiny:
### What happens?

See my repro SQL for details, should be straightforward.

### To Reproduce

```sql
memory D CREATE TABLE t AS SELECT i as val FROM range(0, 200000) t(i);
memory D COPY t TO '/tmp/crash.parquet' (FORMAT PARQUET);
memory D SELECT MIN(file_row_number) FROM read_parquet('/tmp/crash.parquet', file_row_number=true);
INTERNAL Error:
Attempted to access index 1 within vector of size 1
```

### OS:

macos

### DuckDB Version:

834c31be727279dcad62fa508869d6131f220f73

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

PR #22662, opened 2026-05-14
Fix invalid access for file row number

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22661

The out-of-bound access happens [here](https://github.com/duckdb/duckdb/blob/2957fd1e3fd608a6b03fd63795aea3cb7abad40f/extension/parquet/parquet_reader.cpp#L1430-L1441), when optimizer tries to decide whether the stats value for `file_row_number` is exact or not.

Here's the root cause:
- When `file_row_number` requested in the query, DuckDB adds it as a generated column to the root schema [here](https://github.com/duckdb/duckdb/blob/2957fd1e3fd608a6b03fd63795aea3cb7abad40f/extension/parquet/parquet_reader.cpp#L809-L819), but row group columns only contain physical columns
- So in this PR, I treat file row number as a special column and return early at exact stats check

A related thing:
- In the unit test, we already have test cases to get min/max/avg/count for the file row number [here](https://github.com/duckdb/duckdb/blob/2957fd1e3fd608a6b03fd63795aea3cb7abad40f/test/sql/copy/parquet/parquet_row_number.test#L8-L11), but it doesn't trigger the bug
- The reason is, we [return early](https://github.com/duckdb/duckdb/blob/2957fd1e3fd608a6b03fd63795aea3cb7abad40f/src/optimizer/statistics/operator/propagate_aggregate.cpp#L160-L161) if the queried items include ones that are not min/max/count

### DIFF
--- a/extension/parquet/parquet_reader.cpp
+++ b/extension/parquet/parquet_reader.cpp
@@ -1348,6 +1348,12 @@ struct ParquetPartitionRowGroup : public PartitionRowGroup {
 		D_ASSERT(metadata.row_groups.size() > row_group_idx);
 		D_ASSERT(root_schema->children.size() > primary_index);
 
+		// Special handle generated columns.
+		const auto &column_schema = root_schema->children[primary_index];
+		if (column_schema.schema_type == ParquetColumnSchemaType::FILE_ROW_NUMBER) {
+			return true;
+		}
+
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #22665, opened 2026-05-14
Follow-up fixes to StringStatsRework

Mytherin:
* Fix up string stats handling in Parquet reader
* Emit non-truncated stats for string literals 
* Extend `Utf8Proc::FindNextLegalUTF8`, and add `TryGetValidMin / TryGetValidMax` that get a valid UTF8 bound given a set of truncated string stats


test/sqlite/sqllogic_test_runner.cpp
```diff
@@ -863,7 +863,6 @@ void SQLLogicTestRunner::ExecuteFile(string script) {
 						reason += " " + token.parameters[i];
 					}
 				}
-				Printer::PrintF("mode skip %s", reason);
```
Mytherin, line 866: CC @smvv I removed this print because it's popping up when running the unittester normally and cluttering the terminal - not sure if this is a load-bearing print or if this was left in unintentionally

PR #22230, opened 2026-04-23
Clustered Aggregation

peterboncz:
Clustered Aggregation: speed up aggregates when there are few GROUP BY keys, by clustering the selection vector

This PR speeds up grouped aggregation when a chunk contains only a small number of groups (<= 256 currently).

The main idea is to build, once per chunk, a clustered selection vector that groups tuple positions by aggregate state. Aggregates can then process one contiguous run per group instead of repeatedly jumping between states tuple-by-tuple.

Introduces a per-chunk ClusteredAggr structure in AggrInputData that reorders tuple positions by group while preserving order within each group, threads this clustered representation through the aggregate update path as an optional optimizationm and adds a generic clustered execution path in the aggregate executor:

1) For simple aggregates (sum, min/max, bit_and/or/xor, bool_and/or, product) the STRONG path reads the aggregate state once per run, aggregates locally ("in register"), and writes back once at the end of the run.

2) Keeps a WEAKER clustered fallback for all other in-tree duckdb aggregates so they still benefit from run grouping by ccessing their state repeately (memory locality)
without in-register aggregation.

3) It also adds OPTIMIZED handling for count(*) and count(arg) that take advantage of the already known cluster sizes (zero effort) as well as in first/last

The optimization targets queries like TPC-H Q1, where multple aggregates are computed over just 4 groups. On sf=10, mentioned Q1 gets ~2x faster.

PR #22671, opened 2026-05-14
pb/few-groups-aggr-opt-followup

peterboncz:
A followup of https://github.com/duckdb/duckdb/pull/22230 which improved om my MacBook on sf10 (set threads=1;) TPC-H q01 from 1.15 to 0.68s -- but on parquet files only improved from 1.97s to 1.75s. This follow-up gets to 1.55s. This follow-up accelerates hugeint sums specifically, by performing local aggregation on int64.

What happens? When querying parquet files, there are no minmax stats propagated (yet), hence there is no 
- compressed materialization
- perfect aggregation
- sum_no_overflow

in such environments, sums tend to return hugeint, which is slowish to manipulate. It means that any optimization has relatively less impact. Further, because the sum computation takes longer, the CPU has more breathing pause between aggregate updates that hammer the same state, such that the hot-state-latency gap that clusterered aggregation targets, actually shrinks. This means the optimization benefits shrink both relatively and absolutely wrt DuckDB native storage.

But, ClusteredAggregation can help here too: whereas we now optimize the logic by "in register" aggregation (calculating a local aggregate value, and only adding it once per run to the state in the table) using the **original** type, for **hugeint_t** we can try to do the local aggregation on **int64_t**, which is faster:
- for flat integer vectors, typically we can sum at least a vector(2048) values without overflowing int64_t: we call such values "vector-overflow-safe" (rather than global/table granularity safe, which enables sum_no_overflow). We check for this, and only once per run add a local int64_t to huge_int in the state.
- for dictionary vectors with id (they are used many times in a row), we can analyze their values and if they are vector-overflow-safe, convert them to a int64_t array where we  also set NULL values to 0. We can then blindly add such dictionary values to  a local int64_t total without overflow or null checking. And add that once per run to the hugeint state.

Note that in a parquet scan, columns like l_quantity (50 values) and l_discount (11 values) become DICT vectors so the above applies in q01. Also, parquet always puts in a null value (for lolz?) in the dictionary, so validity checks get triggered. This follow-up sidesteps that.


peterboncz:
@lnkuiper adding sum():hugeint optimizations to ClusteredAggregation

src/include/duckdb/common/vector_operations/aggregate_executor.hpp
```diff
@@ -566,6 +572,95 @@ class AggregateExecutor {
...
+					}
+				}
+			}
+		}
+	}
+
+	template <class STATE_TYPE, class INPUT_TYPE>
+	static void ExecuteUnaryClusteredFlatI64HugeintSum(const INPUT_TYPE *vals, const ClusteredAggr &clustered,
+	                                                   const SelectionVector *isel, const sel_t *cluster_iter) {
+		idx_t pos = 0;
```
Mytherin: I think the aggregate executor is the wrong place for such a highly specific function - this should really be part of the `sum` implementation in some manner.

extension/core_functions/aggregate/distributive/sum.cpp
```diff
@@ -52,6 +52,8 @@ struct ClusteredAddOp {
 
 template <class BASE, class LOCAL_OP>
 struct ClusteredSumOperation : public ClusteredSumStateCopy<BASE> {
+	static constexpr bool kClusteredI64HugeintSum = std::is_same<LOCAL_OP, ClusteredAddOp<AddToHugeint>>::value;
```
Mytherin: Can we remove the `k` prefix - we generally don't use that

src/common/clustered_aggregate.cpp
```diff
@@ -312,4 +314,58 @@ bool ClusteredAggrState::TryBuild(ClusteredAggr &clustered, const uint64_t *grou
 	retry_backoff = MinValue<idx_t>(NumericLimits<idx_t>::Maximum() / 2, retry_backoff) * 2;
 	return false;
 }
+
+const DictProps *ClusteredAggrState::GetDictProps(const Vector &input) const {
```
Mytherin: nit: `optional_ptr<const DictProps>`

Mytherin (commented):
Thanks - left some comments

src/include/duckdb/common/vector_operations/aggregate_executor.hpp
peterboncz: right, moved this to sum.cpp

extension/core_functions/aggregate/distributive/sum.cpp
peterboncz: done

src/common/clustered_aggregate.cpp
peterboncz: done

lnkuiper:
Looks good from my side, but I wonder if, moving forward, we might want to implement more clustered aggregation fast paths (this implements `SUM` but we could consider the same optimisation for `AVG`, and potentially more in the future?), so we might want to implement a generic hook rather than the special case here. Wonder what Mark's take on this is?

Issue #22669, opened 2026-05-14
DELIM_LEFT_JOIN -> COMPARISON_JOIN optimization leads to regressions


myrrc:
### What happens?

Hello Duckdb team!

In Vortex extension we have filter pushdown. I'm currently working on https://github.com/vortex-data/vortex/pull/7895
and I believe I've found an incorrectly applied optimisation.

On TPC-H scale factor 10 query 17 here is the query plan if we report at least
one filter as not pushed:

<details>
<summary>At least one filter not pushed</summary>

```
┌───────────────────────────┐
│         PROJECTION        │
│         avg_yearly        │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│    UNGROUPED_AGGREGATE    │
│    Aggregates: sum(#0)    │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│      l_extendedprice      │
│      ~11,991,813 rows     │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│             #1            │
│      ~11,991,813 rows     │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│           FILTER          │
│    ────────────────────   │
│ (CAST(l_quantity AS DOUBLE│
│       ) < SUBQUERY)       │
│                           │
│      ~11,991,813 rows     │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      LEFT_DELIM_JOIN      │
│    ────────────────────   │
│      Join Type: LEFT      │
│                           │
│        Conditions:        │
│ p_partkey IS NOT DISTINCT ├───────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────┐
│       FROM p_partkey      │                                           │                                                                                      │
│                           │                                           │                                                                                      │
│                           │                                           │                                                                                      │
│                           │                                           │                                                                                      │
│       Delim Index: 1      │                                           │                                                                                      │
└─────────────┬─────────────┘                                           │                                                                                      │
┌─────────────┴─────────────┐                             ┌─────────────┴─────────────┐                                                          ┌─────────────┴─────────────┐
│         HASH_JOIN         │                             │         HASH_JOIN         │                                                          │       HASH_GROUP_BY       │
│    ────────────────────   │                             │    ────────────────────   │                                                          │    ────────────────────   │
[...]

PR #22672, opened 2026-05-14
Don't optimize join in Deliminator if any filter has been pushed to table function

myrrc:
This is a fix for https://github.com/duckdb/duckdb/issues/22669
as an PoC. Optimizing the join to a hash join may lead to regressions on some benchmarks. This doesn't reproduce on Parquet becase Parquet reader doesn't do complex filter pushdown outside hive columns / filename column reduction.


myrrc:
We've discussed it privately with Duckdb team and they adviced this won't be an issue in 2.0

PR #22674, opened 2026-05-14
Invalidate transactions when any error is encountered by default

Mytherin:
Currently transactions are invalidated only by certain types of errors. "User" errors early on in the planning phase do not invalidate transactions, in particular `Parser` and `Binder` exceptions are ignored. For example:

```sql
create table t(i int);
begin;
insert into t values (42, 84);
-- Binder Error: table t has 1 columns but 2 values were supplied
insert into t values (42);
commit; - success!
```

This behavior was intentionally chosen for user-friendliness - for interactive use if you would make a typo you would have to re-type a transaction. However, it creates a footgun for non-interactive use. For example, if you had this SQL file:

#### transaction.sql

```sql
create or replace table audit(i int);
create or replace table audit_backup(i int);
insert into audit values (42);

begin;
inser into audit_backup from audit; -- oops, missing "t"
delete from audit;
commit;

from audit;
from audit_backup;
```

And you would run that:

```
duckdb < transaction.sql
```

You will see both `audit` and `audit_backup` are empty after the transaction:

```
┌────────┐
│   i    │
│ int32  │
└────────┘
  0 rows
┌────────┐
│   i    │
│ int32  │
└────────┘
  0 rows
```

The system has "helpfully" ignored the erroneous statement and continued running the transaction, nuking your `audit` table without saving to your backup as was originally intended.

This PR changes this behavior to instead have all errors invalidate transactions, which also mimics how e.g. Postgres behaves, preventing these kinds of footguns.

The behavior is configurable with the `default_transaction_invalidation_policy` setting. This restores the old behavior:

```sql
SET default_transaction_invalidation_policy='SYNTACTIC_ERRORS_DO_NOT_INVALIDATE';
```


thomas-0816:
I'm testing v2.0.0-alpha39998:

```
// insert are successful
create table audit(i int);
begin;
insert into audit values (21);
inser into audit_backup from audit; -- parser error
insert into audit values (22);
commit;
select count(*) from audit; -- 2
```

```
// inserts are lost
CREATE TABLE t1 (id INTEGER);
begin;
INSERT INTO t1 VALUES (1);
SELECT INVALID; -- binder error
INSERT INTO t1 VALUES (2);
COMMIT;
SELECT count(*) FROM t1; -- 0
```

results:
- v2: 2, 0
- v1.5.5: 2, 2

expected for "Invalidate transactions when **any error** is encountered":
- 0, 0

To me, if a transaction is invalidated, the following "commit" should not be successful.
If we are strict, a parser error should also invalidate the transaction.

Mytherin:
The `COMMIT` automatically turns into a `ROLLBACK` if a transaction is invalidated - this is standard in databases. We could push a warning here perhaps.

thomas-0816:
e.g. postgres18 invalidates on parser error and rejects following inserts:

```
psql (18.6 (Ubuntu 18.6-1.pgdg24.04+2))
Type "help" for help.

postgres=# begin;
BEGIN
postgres=*# insert into audit values (21);
INSERT 0 1
postgres=*# inser into audit_backup from audit;
ERROR:  syntax error at or near "inser"
LINE 1: inser into audit_backup from audit;
        ^
postgres=!# insert into audit values (22);
ERROR:  current transaction is aborted, commands ignored until end of transaction block
postgres=!# commit;
ROLLBACK
```

v2 does not invalidate on parser error and commits following inserts:

```
memory D create table audit(i int);
memory D begin;
memory D insert into audit values (21);
memory D inser into audit_backup from audit; -- parser error
Parser Error:
syntax error at or near "into"

LINE 1: inser into audit_backup from audit; -- parser error
              ^^^^
memory D insert into audit values (22);
memory D commit;
memory D select count(*) from audit; -- 2
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            2 │
└──────────────┘
```

(mysql is more the v1.5.5 behavior)

Mytherin:
That indeed seems like a bug - looks like parser errors no longer trigger transaction invalidation.

Issue #22676, opened 2026-05-14
Query enum type from parquet leads to incorrect result


dentiny:
### What happens?

```sql
memory D CREATE TYPE mood AS ENUM ('joy', 'ok', 'happy');
memory D CREATE TABLE t AS SELECT 'joy'::mood AS m FROM range(10) t(i);
memory D COPY t TO '/tmp/enum_stats_single.parquet' (FORMAT PARQUET);
memory D SELECT MIN(m) FROM read_parquet('/tmp/enum_stats_single.parquet');
┌─────────┐
│ min(m)  │
│ varchar │
├─────────┤
│ happy   │
└─────────┘
memory D SELECT MAX(m) FROM read_parquet('/tmp/enum_stats_single.parquet');
┌─────────┐
│ max(m)  │
│ varchar │
├─────────┤
│ ok      │
└─────────┘
```
Both of the select query should return `joy`.

This issue is related to stats population https://github.com/duckdb/duckdb/issues/22548, but for different problem.

### To Reproduce

Repro SQL attached above.

### OS:

macos

### DuckDB Version:

v1.5.2

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

PR #22677, opened 2026-05-14
Fix enum type write to parquet

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22676

This PR followup on Mark's comment here https://github.com/duckdb/duckdb/pull/22658#issuecomment-4450133718, to split the enum type stats population into two parts:
- This PR fixes an existing bug that, min/max considers all possible enum values, but don't mark them as inexact, which leads to incorrect query result. (v1.5 fix branch)
- I will followup with another PR to improve the stats population (main branch)


Mytherin:
Thanks!

PR #22678, opened 2026-05-14
Be careful at midnight

peterboncz:
Eliminate per-row DATE→TIMESTAMP cast when comparing date column to midnight timestamp consts

In predicates like TPC-H Q1's l_shipdate <= date '1998-12-01' - interval '90' day, the RHS evaluates to timestamp '1998-09-02 00:00:00', causing the optimizer to insert CAST(l_shipdate AS TIMESTAMP) on every row.

The comparison simplification rule already tries to move casts from the column side to the constant side, but it bails out for TIMESTAMP→DATE because that cast is lossy in general (many timestamps map to the same date). This patch adds a targeted exception: when the column is DATE and the constant is TIMESTAMP, we check if the specific constant value round-trips through DATE losslessly. A midnight timestamp like '1998-09-02 00:00:00' round-trips perfectly, so the comparison is safely rewritten to l_shipdate <= '1998-09-02'::DATE.

The round-trip is restricted to the DATE→TIMESTAMP pair because it is provably lossless in the forward direction: every date maps to exactly one midnight timestamp, with no possibility of overflow. Non-midnight timestamps fail the round-trip check and are left unchanged.

Before: CAST(l_shipdate AS TIMESTAMP) <= '1998-09-02 00:00:00'::TIMESTAMP
After: l_shipdate <= '1998-09-02'::DATE


src/optimizer/rule/comparison_simplification.cpp
```diff
@@ -58,8 +58,19 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op,
 		// Is the constant cast invertible?
 		if (!cast_constant.IsNull() &&
 		    !BoundCastExpression::CastIsInvertible(cast_expression.GetReturnType(), target_type)) {
-			// Cast is not invertible, so we do not rewrite this expression to ensure that the cast is executed
-			return nullptr;
+			// DATE→TIMESTAMP is lossless (every date is a valid midnight timestamp),
```
hawkfish: It occurs to me that this also holds for any discrete => continuous constant (e.g, integer => decimal).

src/optimizer/rule/comparison_simplification.cpp
```diff
@@ -58,8 +58,19 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op,
...
+			// so we can check if this specific timestamp value round-trips through DATE.
+			if (target_type.id() != LogicalTypeId::DATE ||
```
Mytherin: I wonder if we can generalize this to work for all timestamps. We're looking at specifically a comparison here, and we know dates always map to 00:00:00. It seems to me that depending on the constant timestamp value we just need to adjust the comparison:

```sql
-- truncate changes comparison
SELECT d < TIMESTAMP '2020-01-01 02:30:00';
-- equivalent to
SELECT d <= DATE '2020-01-01';

-- midnight edge case
SELECT d < TIMESTAMP '2020-01-01 00:00:00';
-- equivalent to
SELECT d < DATE '2020-01-01';

-- for greater than equals we can turn this into greater than on the truncated value
SELECT d >= TIMESTAMP '2020-01-01 02:30:00';
-- equivalent to
SELECT d > DATE '2020-01-01';

-- but not at midnight!
SELECT d >= TIMESTAMP '2020-01-01 00:00:00';
-- equivalent to
SELECT d >= DATE '2020-01-01';

-- equality is never equal
SELECT d = TIMESTAMP '2020-01-01 02:30:00';
-- equivalent to
SELECT constant_or_null(d, false);

-- except for midnight!
SELECT d = TIMESTAMP '2020-01-01 00:00:00';
-- equivalent to
SELECT d = DATE '2020-01-01';
```

peterboncz: right. added non-midnight timestamp const handling for =, <>, <=, <, >=, > with a date column.

Note: a previous iteration of this PR generalized the midnight idea with irreversible casts between other types (e.g. signed/unsigned/decimal/floating numbers): just try to cast back and if identical convert to the column type. Was a bit scared for the ultimate correctness of the current reversibility logic so stayed away. But we might take it there as well. Avoiding full column casts that the user doesn't ask for and likely doesn't expect is a usability feature in a way.

src/optimizer/rule/comparison_simplification.cpp
```diff
@@ -58,8 +59,34 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op,
...
-			return nullptr;
+			// DATE→TIMESTAMP is lossless: every DATE maps to a unique midnight TIMESTAMP.
+			// For comparisons with a TIMESTAMP constant, we can rewrite using the floor (cast to DATE),
+			// adjusting by +1 day for operators where the column-as-midnight value is strictly excluded
+			// by a non-midnight constant.
+			if (target_type.id() != LogicalTypeId::DATE ||
+			    cast_expression.GetReturnType().id() != LogicalTypeId::TIMESTAMP) {
+				return nullptr;
+			}
+			if (Timestamp::GetTime(constant_value.GetValue<timestamp_t>()) != dtime_t(0)) {
```
Mytherin: Can we move this to a separate function? Maybe like `DateTimestampComparisonIsInvertible(..)`?

src/optimizer/rule/comparison_simplification.cpp
```diff
@@ -58,8 +59,34 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op,
...
+					return ExpressionRewriter::ConstantOrNull(std::move(cast_expression.child),
+					                                          Value::BOOLEAN(op == ExpressionType::COMPARE_NOTEQUAL));
+				}
+				bool needs_plus_one_day = column_ref_left ? (op == ExpressionType::COMPARE_LESSTHAN ||
+				                                             op == ExpressionType::COMPARE_GREATERTHANOREQUALTO)
+				                                          : (op == ExpressionType::COMPARE_LESSTHANOREQUALTO ||
+				                                             op == ExpressionType::COMPARE_GREATERTHAN);
+				if (op != ExpressionType::COMPARE_LESSTHAN && op != ExpressionType::COMPARE_LESSTHANOREQUALTO &&
+				    op != ExpressionType::COMPARE_GREATERTHAN && op != ExpressionType::COMPARE_GREATERTHANOREQUALTO) {
+					return nullptr;
```
Mytherin: All we're missing here is `COMPARE_DISTINCT_FROM / COMPARE_NOT_DISTINCT_FROM` - can we handle those as well? I think they should just be returning a constant `false / true`.

Mytherin (commented):
Thanks - looks good - some more minor comments

src/optimizer/rule/comparison_simplification.cpp
peterboncz: done

src/optimizer/rule/comparison_simplification.cpp
peterboncz: done

src/optimizer/rule/comparison_simplification.cpp
```diff
@@ -58,8 +103,15 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op,
 		// Is the constant cast invertible?
 		if (!cast_constant.IsNull() &&
 		    !BoundCastExpression::CastIsInvertible(cast_expression.GetReturnType(), target_type)) {
-			// Cast is not invertible, so we do not rewrite this expression to ensure that the cast is executed
-			return nullptr;
+			if (target_type.id() != LogicalTypeId::DATE ||
```
Mytherin: Maybe we can also move this whole branch into a function `ConstantCastIsInvertible`? The branching here is not super readable

src/optimizer/rule/comparison_simplification.cpp
```diff
@@ -1,13 +1,58 @@
...
+	if (Timestamp::GetTime(constant_value.GetValue<timestamp_t>()) == dtime_t(0)) {
+		return nullptr; // it's midnight: no replacement needed
+	}
+	auto op = expr.GetExpressionType();
+	bool if_left_plus = false, pred_neq = false;
+	switch (op) {
+	case ExpressionType::COMPARE_NOTEQUAL:
+	case ExpressionType::COMPARE_DISTINCT_FROM:
+		pred_neq = true;
+		DUCKDB_EXPLICIT_FALLTHROUGH;
```
Mytherin: I think this simpler single-line version is more readable than the switch statement shenanigans:

```cpp
bool pred_neq = op == ExpressionType::COMPARE_NOTEQUAL || op == ExpressionType::COMPARE_DISTINCT_FROM;


bool if_left_plus = op == ExpressionType::COMPARE_LESSTHAN || op == ExpressionType::COMPARE_GREATERTHANOREQUALTO;
```

Mytherin (commented):
Thanks - some more minor readability notes - and can you also look at the failing test?

src/optimizer/rule/comparison_simplification.cpp
peterboncz: sure!

src/optimizer/rule/comparison_simplification.cpp
peterboncz: ok

peterboncz:
eventually I went to using switch (no fallthrough though) in an attempt to make the code easier to read using also examples in comments.

src/planner/filter/expression_filter.cpp
```diff
@@ -212,12 +212,19 @@ static FilterPropagateResult CheckComparisonStatistics(const BoundFunctionExpres
...
 	if (!filter_stats->CanHaveNoNull()) {
-		return FilterPropagateResult::FILTER_ALWAYS_FALSE;
+		return distinct_from ? FilterPropagateResult::FILTER_ALWAYS_TRUE : FilterPropagateResult::FILTER_ALWAYS_FALSE;
 	}
 	auto result = CheckZonemapAgainstConstants(*filter_stats, comparison_type, array_ptr<const Value>(&constant, 1));
-	if (result == FilterPropagateResult::FILTER_ALWAYS_TRUE && filter_stats->CanHaveNull()) {
-		return FilterPropagateResult::NO_PRUNING_POSSIBLE;
+	if (filter_stats->CanHaveNull()) {
+		if (distinct_from) {
+			return result == FilterPropagateResult::FILTER_ALWAYS_FALSE ? FilterPropagateResult::NO_PRUNING_POSSIBLE
```
Mytherin: I'm not really following this change - `col IS DISTINCT FROM <non_null_const>` should be exactly the same as `x <> <non_null_const>`, no? Why the special code path here?

src/planner/filter/expression_filter.cpp
```diff
@@ -212,12 +212,19 @@ static FilterPropagateResult CheckComparisonStatistics(const BoundFunctionExpres
...
+			                                                            : result;
+		}
+		if (result == FilterPropagateResult::FILTER_ALWAYS_TRUE) {
+			return FilterPropagateResult::NO_PRUNING_POSSIBLE;
```
Mytherin: This kind of seems like it's working around a bug either upstream or downstream - in either case it doesn't look like the correct solution. Maybe this is meant to now return `FILTER_TRUE_OR_NULL`? But it seems like `NULL` values are not being consistently handled here.

Mytherin: These fixes should ideally also be moved to a separate PR.

Mytherin (commented):
Thanks - the timestamp code looks good to me - some comments about the statistics propagation changes

src/planner/filter/expression_filter.cpp
peterboncz: OK, I think this distinct-zonemap stuff is indeed something different and I removed it all from the PR

Mytherin:
Thanks!

PR #22680, opened 2026-05-15
Fix/eviction queue purge correctness

krleonid:
## Summary

This PR fixes three related bugs in the buffer pool's eviction queue purge path. Together they caused dead nodes to silently accumulate in the eviction queue, alive-but-pinned nodes to be dropped (disturbing LRU and corrupting the dead-node counter), and the multi-iteration purge loop to make essentially no progress under contention.

The fix is small (3 files, +37 / −16) but the underlying issues interact and are worth fixing together so the purge accounting becomes self-consistent again.

## Background

`EvictionQueue` uses a `duckdb_moodycamel::ConcurrentQueue<BufferEvictionNode>` where each node holds:

- `weak_ptr<BlockMemory>` — the block this entry refers to.
- `handle_sequence_number` — incremented every time the block is re-added to the queue.

A node is logically *dead* if either:

1. The `weak_ptr` is expired (the `BlockMemory` was destroyed), or
2. The block's current `eviction_seq_num` is newer than the node's recorded sequence number (a newer entry superseded it).

Dead nodes are never proactively removed — they are dequeued lazily by `EvictionQueue::Purge()`, which runs on one thread at a time and tries to keep the queue's dead-node ratio under control.

### 1. `IncrementDeadNodes` was skipped when the buffer was already null

`BlockMemory::~BlockMemory` only called `IncrementDeadNodes` when `GetBuffer()` returned non-null:

```cpp
// before
if (GetBuffer() && GetBufferType() != FileBufferType::TINY_BUFFER) {
    GetBufferManager().GetBufferPool().IncrementDeadNodes(*this);
}
```

But the queue entry's lifetime is independent of whether the buffer is currently loaded — only the dequeue side ever removes it. If the block had previously been evicted (buffer = nullptr) and then the `BlockMemory` itself was destroyed, the queue entry that pointed at it became dead but was never accounted for. This systematically under-counted `total_dead_nodes`, which made the purge ratio early-out (`approx_alive_nodes * (ALIVE_NODE_MULTIPLIER - 1) > approx_dead_nodes`) fire too often and let dead entries accumulate.

**Fix:** drop the `GetBuffer()` guard.

### 2. Pinned-but-current nodes were classified as dead during purge

`PurgeIteration` used `TryGetBlockMemory()` to decide whether a dequeued node was alive:

```cpp
shared_ptr<BlockMemory> BufferEvictionNode::TryGetBlockMemory() {
    auto shared_memory_p = memory_p.lock();
    if (!shared_memory_p) {
        return nullptr;
    }
    if (!CanUnload(*shared_memory_p)) {  // also checks readers / pinning
        return nullptr;
    }
    return shared_memory_p;
}
```

`CanUnload` is **not** a deadness predicate — it also returns `false` when the block is currently pinned (`readers != 0`). So a perfectly valid latest-version node that just happened to be pinned at the moment of the purge was:

- Dropped from the queue (LRU position lost).
- Counted toward `total_dead_nodes -= actually_dequeued - alive_nodes`, decrementing the dead counter for a node that was never dead.

This actively desyncs `total_dead_nodes` from reality and combines badly with bug.

**Fix:** introduce a dedicated `BufferEvictionNode::IsDeadNode()` that checks only the two real deadness conditions (expired weak_ptr or stale sequence number) and use it for purge classification:

```cpp
bool BufferEvictionNode::IsDeadNode() {
    auto shared_memory_p = memory_p.lock();
    if (!shared_memory_p) {
        return true;
    }
    if (handle_sequence_number != shared_memory_p->GetEvictionSequenceNumber()) {
        return true;
    }
    return false;
}
```

`PurgeIteration` now decrements `total_dead_nodes` by an explicit `dead_count`, which can never include pinned-but-current entries.

### 3. Multi-iteration purge churned the purge thread's own moodycamel sub-queue

This is the most impactful bug.

`moodycamel::ConcurrentQueue` is **not** a single FIFO. Internally it maintains one sub-queue per producer thread (implicit producers are created lazily on first enqueue from a given thread). `try_dequeue_bulk` without a token walks the producer list and tends to drain the most recently used sub-queues first.

The previous purge loop re-enqueued alive nodes between iterations using an implicit producer:

```cpp
// PurgeIteration (old)
q.enqueue_bulk(purge_nodes.begin(), alive_nodes);
```

That enqueue happens on the purge thread, which in nearly all interesting scenarios was not previously a producer for this queue (producers are worker threads). It creates an implicit-producer sub-queue owned by the purge thread, and on the **next** `try_dequeue_bulk` call moodycamel preferentially re-drains it — pulling back the very alive nodes the purge just inserted, instead of reaching dead entries sitting in worker producer sub-queues. Under sustained pressure the loop made little to no real progress: it kept "purging" the same alive nodes round and round while real dead entries accumulated elsewhere.

**Fix:** use explicit `ConsumerToken` and `ProducerToken` to separate the dequeue and re-enqueue paths into distinct sub-queues. The `ConsumerToken` remembers which sub-queue it last consumed from and progresses sequentially through the producer list, ensuring forward progress. The `ProducerToken` routes re-enqueued alive nodes into a dedicated sub-queue that the consumer has already passed, preventing the purge thread from immediately re-draining its own output:

```cpp
// EvictionQueue members
duckdb_moodycamel::ConsumerToken purge_consumer_token;
duckdb_moodycamel::ProducerToken purge_producer_token;

void EvictionQueue::PurgeIteration(const idx_t purge_size) {
    // Dequeue using consumer token — progresses through sub-queues sequentially
    const idx_t actually_dequeued = q.try_dequeue_bulk(purge_consumer_token, purge_nodes.begin(), purge_size);
    // ... classify alive vs dead ...
    // Re-enqueue alive nodes via producer token — goes into a dedicated sub-queue
    // that the consumer token has already passed
    if (alive_count > 0) {
        q.enqueue_bulk(purge_producer_token, purge_nodes.begin(), alive_count);
    }
}
```

This guarantees that each purge iteration sees fresh entries from worker sub-queues instead of bouncing the purge thread's own buffer, while keeping alive nodes in the queue at all times (unlike the alternative of holding them aside externally, which would starve the eviction process).

## How they interact

| Bug | Effect on `total_dead_nodes` | Effect on the queue |
|---|---|---|
| 1 — buffer-null skip | Under-counts dead nodes | Dead entries leak in |
| 2 — pinned treated as dead | Decremented for non-dead nodes | Pinned entries are dropped (LRU disturbed) |
| 3 — sub-queue churn | — | Multi-iteration purge does no useful work |

Bugs 1 and 2 push the counter in opposite directions, masking each other intermittently and making the symptom hard to trace. Bug 3 means even when the ratio check correctly demands more aggressive purging, the loop fails to actually purge.

## Files changed

- `src/include/duckdb/storage/buffer/buffer_pool.hpp` — declare `BufferEvictionNode::IsDeadNode`.
- `src/storage/buffer/buffer_pool.cpp` — implement `IsDeadNode`, restructure `Purge` / `PurgeIteration` with consumer/producer tokens.
- `src/storage/buffer/block_handle.cpp` — drop `GetBuffer()` guard before `IncrementDeadNodes`.

## Test plan

- [ ] `make reldebug` builds cleanly.
- [ ] `build/reldebug/test/unittest` passes (full unit suite).
- [ ] Targeted buffer-manager / eviction tests pass (`build/reldebug/test/unittest "[buffermanager]"`).
- [ ] Local repro of dead-node accumulation under heavy churn no longer grows `total_dead_nodes` unboundedly between purges.
- [ ] LRU-sensitive workloads do not show pinned blocks being repeatedly evicted/reloaded after purge.
- [ ] Multi-iteration purge actually shrinks the queue under contention (verified by observing `q.size_approx()` between iterations in a stress harness).

## Notes for reviewers

- `BufferEvictionNode::IsDeadNode()` is intentionally *not* `const` because `weak_ptr::lock()` is a mutating operation only in spirit — happy to mark it `const` if preferred (it would still compile since `lock()` is `const` on the weak_ptr).
- The `ConsumerToken` / `ProducerToken` approach keeps alive nodes in the queue at all times during purge (unlike the earlier "accumulate externally" approach), so concurrent `TryDequeueWithLock` on the same queue can still find evictable entries even while a purge is in progress.
- The producer token's sub-queue will be visited by the consumer token on a subsequent purge cycle, ensuring these re-enqueued nodes are eventually revisited (and purged if they've since become dead).

### DIFF
--- a/src/include/duckdb/parallel/concurrentqueue.hpp
+++ b/src/include/duckdb/parallel/concurrentqueue.hpp
@@ -25,12 +25,14 @@ class BlockingConcurrentQueue;
 
 struct ProducerToken {
 	//! Constructor
-	template <typename T, typename Traits>
-	explicit ProducerToken(ConcurrentQueue<T> &);
+	template <typename T>
+	explicit ProducerToken(ConcurrentQueue<T> &) {
+	}
 	//! Constructor
-	template <typename T, typename Traits>
-	explicit ProducerToken(BlockingConcurrentQueue<T> &);
-	//! Constructor
+	template <typename T>
+	explicit ProducerToken(BlockingConcurrentQueue<T> &) {
+	}
+	//! Move constructor
 	ProducerToken(ProducerToken &&) {
 	}
 	//! Is valid token?
--- a/src/include/duckdb/parallel/concurrentqueue.hpp
+++ b/src/include/duckdb/parallel/concurrentqueue.hpp
@@ -39,6 +39,20 @@ struct ProducerToken {
 	}
 };
 
+struct ConsumerToken {
+	//! Constructor
+	template <typename T>
+	explicit ConsumerToken(ConcurrentQueue<T> &) {
+	}
+	//! Constructor
+	template <typename T>
+	explicit ConsumerToken(BlockingConcurrentQueue<T> &) {
+	}
+	//! Move constructor
+	ConsumerToken(ConsumerToken &&) {
+	}
+};
+
 template <typename T>
 class ConcurrentQueue {
 private:
@@ -84,6 +98,11 @@ public:
 		}
 		return max;
 	}
+	//! Dequeues several elements from the queue using a consumer token.
+	template <typename It>
+	size_t try_dequeue_bulk(ConsumerToken &, It itemFirst, size_t max) {
+		return try_dequeue_bulk(itemFirst, max);
+	}
 
 	template <typename It>
 	bool enqueue_bulk(It itemFirst, size_t count) {
@@ -91,6 +110,11 @@ public:
 			q.push(std::move(*itemFirst++));
 		}
 		return true;
+	}
+	//! Enqueues several elements using a producer token.
+	template <typename It>
+	bool enqueue_bulk(ProducerToken const &, It itemFirst, size_t count) {
+		return enqueue_bulk(itemFirst, count);
 	}
 };
 
--- a/src/include/duckdb/storage/buffer/buffer_pool.hpp
+++ b/src/include/duckdb/storage/buffer/buffer_pool.hpp
@@ -34,7 +34,7 @@ struct BufferEvictionNode {
 
 	bool CanUnload(BlockMemory &memory);
 	shared_ptr<BlockMemory> TryGetBlockMemory();
-	bool IsDeadNode();
+	bool IsDeadNode(idx_t debug_sleep_micros = 0);
 };
 
 //! The BufferPool is in charge of handling memory management for one or more databases. It defines memory limits
--- a/src/include/duckdb/storage/buffer/buffer_pool.hpp
+++ b/src/include/duckdb/storage/buffer/buffer_pool.hpp
@@ -34,7 +34,7 @@ struct BufferEvictionNode {
 
 	bool CanUnload(BlockMemory &memory);
 	shared_ptr<BlockMemory> TryGetBlockMemory();
-	bool IsDeadNode(idx_t debug_sleep_micros = 0);
+	bool IsDeadNode(optional_idx debug_sleep_micros = optional_idx());
 };
 
 //! The BufferPool is in charge of handling memory management for one or more databases. It defines memory limits
--- a/src/include/duckdb/storage/buffer/buffer_pool.hpp
+++ b/src/include/duckdb/storage/buffer/buffer_pool.hpp
@@ -34,6 +34,7 @@ struct BufferEvictionNode {
 
 	bool CanUnload(BlockMemory &memory);
 	shared_ptr<BlockMemory> TryGetBlockMemory();
+	bool IsDeadNode();
 };
 
 //! The BufferPool is in charge of handling memory management for one or more databases. It defines memory limits
--- a/src/storage/buffer/buffer_pool.cpp
+++ b/src/storage/buffer/buffer_pool.cpp
@@ -293,6 +293,7 @@ bool BufferPool::AddToEvictionQueue(shared_ptr<BlockHandle> &handle) {
 	// The block handle is locked during this operation (Unpin),
 	// or the block handle is still a local variable (ConvertToPersistent)
 	D_ASSERT(memory.GetReaders() == 0);
+	memory.MarkAddedToEvictionQueue();
 	auto ts = memory.NextEvictionSequenceNumber();
 	if (track_eviction_timestamps) {
 		memory.SetLRUTimestamp(std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now())
--- a/src/storage/buffer/buffer_pool.cpp
+++ b/src/storage/buffer/buffer_pool.cpp
@@ -298,7 +298,6 @@ bool BufferPool::AddToEvictionQueue(shared_ptr<BlockHandle> &handle) {
 	// The block handle is locked during this operation (Unpin),
 	// or the block handle is still a local variable (ConvertToPersistent)
 	D_ASSERT(memory.GetReaders() == 0);
-	memory.MarkAddedToEvictionQueue();
 	auto ts = memory.NextEvictionSequenceNumber();
 	if (track_eviction_timestamps) {
 		memory.SetLRUTimestamp(std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now())
--- a/src/storage/buffer/buffer_pool.cpp
+++ b/src/storage/buffer/buffer_pool.cpp
@@ -66,6 +66,17 @@ shared_ptr<BlockMemory> BufferEvictionNode::TryGetBlockMemory() {
 	return shared_memory_p;
 }
 
+bool BufferEvictionNode::IsDeadNode() {
+	auto shared_memory_p = memory_p.lock();
+	if (!shared_memory_p) {
+		return true;
+	}
+	if (handle_sequence_number != shared_memory_p->GetEvictionSequenceNumber()) {
+		return true;
+	}
+	return false;
+}
+
 typedef duckdb_moodycamel::ConcurrentQueue<BufferEvictionNode> eviction_queue_t;
 
 struct EvictionQueue {
@@ -109,7 +120,7 @@ public:
 
 private:
 	//! Bulk purge dead nodes from the eviction queue. Then, enqueue those that are still alive.
-	void PurgeIteration(const idx_t purge_size);
+	void PurgeIteration(const idx_t purge_size, vector<BufferEvictionNode> &alive_nodes_out);
 
 public:
 	//! The type of the buffers in this queue and helper function (both for verification only)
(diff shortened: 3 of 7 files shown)


src/storage/buffer/buffer_pool.cpp
```diff
@@ -227,25 +251,21 @@ void EvictionQueue::PurgeIteration(const idx_t purge_size) {
...
-	idx_t alive_nodes = 0;
+	idx_t dead_count = 0;
 	auto debug_sleep_micros = debug_eviction_queue_sleep.load(std::memory_order_relaxed);
 	for (idx_t i = 0; i < actually_dequeued; i++) {
 		auto &node = purge_nodes[i];
-		auto handle = node.TryGetBlockMemory();
 		if (debug_sleep_micros > 0) {
-			// Debug race conditions regarding the ownership of the BlockMemory.
 			ThreadUtil::SleepMicroSeconds(debug_sleep_micros);
 		}
```
taniabogatsch: We still take ownership of the `weak_ptr` in `IsDeadNode` below, so let's move the debug code inside that function to ensure our tests for race conditions on block memory ownership stay functional.

```cpp
node.IsDeadNode(debug_sleep_micros)
// and then have the conditional in that function to sleep after acquiring the shared pointer
```

test/api/test_buffer_pool_eviction.cpp
```diff
@@ -262,3 +262,139 @@ TEST_CASE("Test buffer pool eviction: failed to allocate space if every page and
...
+	Connection con(db);
+	auto &context = *con.context;
+	auto &buffer_manager = BufferManager::GetBufferManager(context);
+	auto &buffer_pool = DatabaseInstance::GetDatabase(context).GetBufferPool();
+	const idx_t initial_memory = buffer_pool.GetUsedMemory();
+
+	constexpr idx_t page_size = 1024 * 1024; // 1 MiB
+	constexpr idx_t total_pages = 6;
+	constexpr idx_t held_pages = 2;
+	const idx_t actual_alloc_size = BufferManager::GetAllocSize(page_size + Storage::DEFAULT_BLOCK_HEADER_SIZE);
```
taniabogatsch: Can we rename these to be more DuckDB idiomatic? We don't use `pages`. Are you referring to buffers here? Same for the test below.

test/api/test_buffer_pool_eviction.cpp
taniabogatsch, line 1: I think C++ tests make sense here to trigger specific queue behavior. Could we additionally add a concurrent/high load test (maybe `test_slow`) using the `duckdb_eviction_queues` that ensures that no underflow has happened by checking that the queue sizes are sane?

taniabogatsch (changes requested):
Great find! Left a small review.

src/include/duckdb/parallel/concurrentqueue.hpp
taniabogatsch, line 1: Why do we need these changes instead of the previous vector to store the alive nodes in?

src/include/duckdb/storage/buffer/buffer_pool.hpp
```diff
@@ -34,6 +34,7 @@ struct BufferEvictionNode {
 
 	bool CanUnload(BlockMemory &memory);
 	shared_ptr<BlockMemory> TryGetBlockMemory();
+	bool IsDeadNode(idx_t debug_sleep_micros = 0);
```
taniabogatsch: `optional_idx` to indicate "not set"

src/storage/buffer/buffer_pool.cpp
```diff
@@ -66,13 +66,27 @@ shared_ptr<BlockMemory> BufferEvictionNode::TryGetBlockMemory() {
 	return shared_memory_p;
 }
 
+bool BufferEvictionNode::IsDeadNode(idx_t debug_sleep_micros) {
+	auto shared_memory_p = memory_p.lock();
+	if (debug_sleep_micros > 0) {
```
taniabogatsch: `debug_sleep_micros.IsValid()`

src/storage/buffer/buffer_pool.cpp
```diff
@@ -191,10 +209,10 @@ void EvictionQueue::Purge() {
 	// guaranteeing that we always exit the loop.
 
 	idx_t max_purges = approx_q_size / purge_size;
+
 	while (max_purges != 0) {
 		PurgeIteration(purge_size);
 
-		// update relevant sizes and potentially early-out
```
taniabogatsch, line 197: Keep comment?

src/include/duckdb/parallel/concurrentqueue.hpp
krleonid: This is a good point: The vector solution removes nodes from the queue, reducing the number of nodes (perhaps the alive ones). Therefore, this might lead to latency and even OOM during the eviction process. The eviction process uses the same queue to find the alive node(s).

src/storage/buffer/buffer_pool.cpp
```diff
@@ -224,28 +243,36 @@ void EvictionQueue::PurgeIteration(const idx_t purge_size) {
...
-			ThreadUtil::SleepMicroSeconds(debug_sleep_micros);
-		}
-		if (handle) {
-			purge_nodes[alive_nodes++] = std::move(node);
+		if (node.IsDeadNode(debug_sleep_micros)) {
+			dead_count++;
+		} else {
+			// Move alive nodes to the front for bulk re-enqueue
+			if (alive_count != i) {
+				purge_nodes[alive_count] = std::move(node);
```
lnkuiper: Until we encounter the first dead node, `alive_count` will be equal to `i`, which will cause us to skip `purge_nodes[alive_count] = std::move(node);`, right? This means that the node goes missing from the eviction queue, and can never be evicted.

Why isn't this done unconditionally with `purge_nodes[alive_nodes++] = std::move(node);` like before?

lnkuiper (commented):
Thanks for the PR! Looks good, I have one comment:

src/storage/buffer/buffer_pool.cpp
krleonid: @lnkuiper 

 When alive_count == i, the node is already in the correct position (purge_nodes[i]), so there's no need to move it. The alive_count++ on line 265 still runs unconditionally (it's outside the if), so the node isn't lost.
  The optimisation skips the self-assignment purge_nodes[i] = std::move(purge_nodes[i]), which would be a no-op.

krleonid: I've changed it

lnkuiper: Thanks for the explanation. Was OK before then, but this should be a bit easier to read now. Let's run CI

src/include/duckdb/parallel/concurrentqueue.hpp
taniabogatsch: Ah, that makes sense, thanks!

taniabogatsch (approved):

PR #22681, opened 2026-05-15
Parse mode skip/require reasons once and summarize

smvv:
Print summarized `mode skip ...` only at the end of test run:
```shell
$ build/debug/test/unittest '[afl]' 
Filters: [afl]
[30/30] (100%): test/sql/copy/csv/afl/fuzz_20250226.test                                                                                                        
====================================================================
All tests passed (5 skipped tests, 347 assertions in 25 test cases)


Skipped tests for the following reasons:
mode skip flaky: 1
require json: 5
```

(It was printed after each test that had one or more mode skip statements.)

### DIFF
--- a/scripts/ci/run_tests.py
+++ b/scripts/ci/run_tests.py
@@ -333,54 +333,41 @@ def parse_skipped_tests_count(output: str):
     return int(match.group(1))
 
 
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks!

PR #22683, opened 2026-05-15
Fix Mark Join Null Handling

d-justen:
fixes https://github.com/duckdb/duckdb/issues/19182
fixes https://github.com/duckdblabs/duckdb-internal/issues/6065

This PR fixes SQL three-valued logic for uncorrelated row-valued IN, NOT IN, and ANY comparisons implemented through MARK joins.

  The central bug was that DuckDB treated NULL too coarsely for multi-column comparisons:

  - A NULL anywhere on the RHS could turn every unmatched probe into NULL, even when another column made the comparison definitively false.
  - A NULL anywhere on the probe side could immediately turn the result into NULL, even when every RHS row differed decisively in another column.
  - Nested values such as LIST, ARRAY, MAP, STRUCT, and UNION did not consistently propagate internal NULLs.
  - Row-valued <> ANY was incorrectly decomposed into independent conditions.
  - A row represented as one tuple-valued physical condition could bypass the new handling.

This PR implements a Postgres-style mark join null handling for those queries. For that, the join operator stores build rows in a separate collection and then compares each dangling tuple from the left-hand side with that collection individually.

This post-processing step has quadratic complexity, which is expected ([1](https://vldb.org/cidrdb/papers/2026/p3-birler.pdf)), but there are a few shortcuts that we could look into in the future. For now, I would favor result correctness over more complicated changes in this PR.

There are cases, in which we may treat a probe result of NULL and FALSE the same, specifically if we only select TRUE marked tuples afterward. In that case, we do not create the additional tuple collection.


Mytherin (commented):
Thanks - I realize the change is not easy but this seems to be adding a lot of complexity that could maybe be reduced. It does not seem to integrate at all with the existing (already complex) mark join logic (`correlated_mark_join_info`). Maybe it would be possible to instead generalize that to solve this issue?

d-justen:
Okay, so what I tried to do with this refactoring now is to process as much of the mark join post-processing logic in `MarkJoinPostProcessor`, which has been distributed over `PhysicalComparisonJoin`, `PhysicalHashJoin`, `JoinHashTable`, and so on. This whole thing is honestly a bit difficult to generalize because there are many parameters that may change the desirable strategy:

- simple single-condition MARK joins keep the cheap fast path
- correlated single-column ANY still uses the existing count-based logic, but now behind the same abstraction
- row-valued equality MARK joins use the new remainder-based refinement
- nested loop MARK joins can also reuse the same post-processing flow, with a condition-scan refinement when the coarse has_null rule is not sufficient
- queries in which the differentiation between null/false does not matter skip the post-processing completely

Correlated multi-column mark joins are still not supported, and I would have to check in future PR(s) if that case is at least in parts generalizable with the correlated single-condition case.

lnkuiper:
There is another PR up https://github.com/duckdb/duckdb/pull/22549 But I think this PR is actually the right direction.

I think this is pretty close to ready, but Codex has found some correctness issues:
```sql
# name: test/temp.test
# description: Temporary review repros for PR #22683 row-valued <> ANY null handling
# group: [temp]

query I
SELECT (1, 2) <> ANY (SELECT * FROM (VALUES (1, NULL::INT)) t(x, y));
----
NULL

query I
SELECT (1, 2) <> ANY (SELECT * FROM (VALUES (NULL::INT, 2)) t(x, y));
----
NULL

query I
SELECT (1, 2) <> ANY (SELECT * FROM (VALUES (1, NULL::INT), (1, 2)) t(x, y));
----
NULL
```
Feel free to cross-reference vs. PostgreSQL.

Also, the newly defined `enum` in `mark_join_post_processor.hpp` makes it so everywhere that we include `enum_util.hpp`, we also include `mark_join_post_processor.hpp`. I think it might be better to place the enum in `duckdb/common/enums/<enum_name>.hpp`.

Once this has been resolved it's ready to go :)

lnkuiper:
Looks structurally sound, Codex found some issues still:
```sql
# name: test/temp.test
# description: Temporary review repros for PR #22683
# group: [temp]

# The null-remainder path must also run when every RHS row is filtered from the hash table.
query II
SELECT *
FROM (VALUES (0, 0), (1, 0)) outer_tab(f1, f2)
WHERE (f1, f2) NOT IN (
	SELECT * FROM (VALUES (0, NULL::INT)) inner_tab(c1, c2)
)
ORDER BY ALL;
----
1	0

# The preserved row comparison must coerce both sides to a common tuple type.
query I
SELECT (1::SMALLINT, 2::SMALLINT) <> ANY (
	SELECT * FROM (VALUES (1::BIGINT, 3::BIGINT)) rhs(a, b)
);
----
true

# A scalar UNION payload inside an unnamed row must not be treated as another struct.
query TT
SELECT i, (union_value(num := 1)::UNION(num INT, str VARCHAR), 2) <> ANY (
	SELECT u, v
	FROM (VALUES (1, union_value(num := 1)::UNION(num INT, str VARCHAR), 3)) rhs(k, u, v)
	WHERE rhs.k = outer_t.i
)
FROM (VALUES (1)) outer_t(i);
----
1	true
```

The `UNION` thing is a bit annoying.
> Equal UNION tags recurse into the member value, but that value may be scalar; it then reaches StructValue::GetChildren and crashes. Could we add a scalar base case here? This three-valued row-equality logic is also duplicated in nested_loop_join_mark.cpp and mark_join_post_processor.cpp. It would be good to
share one helper so all MARK-join paths use identical NULL semantics.

d-justen:
Thanks for looking into it again @lnkuiper. I am doing a larger refactoring of this at the moment as it has gotten a bit out of hand. Will let you know once this is ready for review again :)

d-justen:
Hey @lnkuiper, could you please take another look now? :) Regarding your last test case: this PR only focusses on uncorrelated, multi-column mark joins. The correlated ones are not supported yet and would need some extra attention in a follow-up.

lnkuiper:
Thanks, this looks much simpler now! I found two more correctness issues:

1. `<> ANY` still returns `NULL` if a NULL field comes before a later mismatch. For example,
   `(NULL, 3) <> ANY (VALUES (1, 2))` should be `TRUE`. The same issue appears in the hash path when the RHS returns the row as a single tuple column. I think both paths need to share a row comparison helper that keeps looking for a decisive mismatch after seeing a NULL.

2. A correlated multi-column comparison can bypass the new `not yet supported` check when the row is returned as one tuple-valued column. The guard only checks the number of expressions/types, which are both one in that case, and the query returns `FALSE` instead of `NULL`. Since correlated multi-column MARK joins are out of scope here, I think this should be rejected based on the tuple arity as well.

Both issues also exist on the base commit, but are relevant to this PR. Also happy to review a follow-up PR that fixes these, let me know what you prefer.

Repros:

```sql
query I
SELECT (NULL::INT, 3) <> ANY (
	SELECT * FROM (VALUES (1, 2)) rhs(a, b)
);
----
true

query I
SELECT row(NULL::INT, 3) = ANY (
	SELECT s FROM (VALUES (row(1, 2))) rhs(s)
);
----
false

query I
SELECT row(a, b) = ANY (
	SELECT s
	FROM (VALUES (1, row(1, NULL::INT))) rhs(k, s)
	WHERE k = lhs.a
)
FROM (VALUES (1, 1)) lhs(a, b);
----
NULL
```

lnkuiper:
The previous comments are fixed, looking really good now.

I found one edge-case performance issue:

Nested keys now always run the refinement scan, even when neither side contains a nested `NULL`. Every unmatched LHS row scans all of `uncorrelated_condition_rows`, which makes this quadratic. With one thread and 10K disjoint rows I get `0.01/0.01/0.01s` on base `a387bda31d` versus `0.47/0.48/0.49s` on `e8af0a28d` after warm-up. Could we track
nested NULLs and skip the refinement when neither side can produce `UNKNOWN`? Or perhaps this can be sped up using bulk validity operations?

```sql
.timer on
SET threads=1;

CREATE TABLE rhs AS
SELECT list_value((i + 100000)::BIGINT) AS a, 0::INT AS b
FROM range(10000) t(i);

SELECT count(*)
FROM range(10000) lhs(i)
WHERE row(list_value(i), 0::INT) NOT IN (SELECT a, b FROM rhs);
```

As a side note: `GetTotalSize()` includes `uncorrelated_condition_rows`, while `GetRemainingSize()` still
does not, even though the collection is kept across external hash-join rounds.

lnkuiper:
Thanks! LGTM now

PR #22684, opened 2026-05-15
Fix numeric stats serde

dentiny:
Min/max stats are swapped on serialization, it doesn't show up as a bug since deserialization also swaps as well.
I hope it won't compatibility issue for `quack`.


Mytherin:
Thanks - but I think this fix will break all of our storage format backwards compatibility if we fix it in this manner. I think a better fix would be to just change the name of the property (i.e. swap `"min"` with `"max"` and vice versa`). In practice that name has no effect on serialization to binary, which is effectively always how we serialize numeric stats.

dentiny:
> but I think this fix will break all of our storage format backwards compatibility if we fix it in this manner.

Yeah that's true.. swapped the wording, also retarget to main branch, not a bug fix

Mytherin:
Thanks!

dentiny:
```sh
select count(*) from read_csv_auto('duckdb_unittest_tempdir/166333/t1.csv',buffer_size = 262144, sample_size=-1);
Actual result:
Out of Memory Error: could not allocate block of size 260.0 KiB (2.6 MiB/2.8 MiB used)
```
Weird, it shouldn't have anything to do with this PR

Mytherin:
Thanks!

PR #22688, opened 2026-05-15
Fix max file row number

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22686

The issue is:
- Currently, max = min + row number, which should be max = min + row number - 1
- Actually existing unit tests already surface the issue when returning min/max stats directly (without scanning data)


Mytherin:
Thanks!

PR #22690, opened 2026-05-15
Add vacuum_rebuild_indexes as an (experimental) ATTACH option

evertlammerts:
This adds `vacuum_rebuild_indexes` as an ATTACH option in addition to the global setting.

This allows users to enable vacuuming of indexed tables per-table rather than as a global setting.

Note that the setting (both as attach and as global option) is experimental. We should revisit it before v2.0 and see in what form we want to keep it, if at all.

### DIFF
--- a/src/include/duckdb/main/attached_database.hpp
+++ b/src/include/duckdb/main/attached_database.hpp
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include "duckdb/main/settings.hpp"
 #include "duckdb/main/config.hpp"
 #include "duckdb/catalog/catalog_entry.hpp"
 
@@ -57,7 +58,7 @@ struct StoredDatabasePath {
 //! they have to apply to any database file type (duckdb, sqlite, etc.).
 struct AttachOptions {
 	//! Constructor for databases we attach outside of the ATTACH DATABASE statement.
-	explicit AttachOptions(const DBConfigOptions &options);
+	explicit AttachOptions(const DBConfig &config);
 	//! Constructor for databases we attach when using ATTACH DATABASE.
 	AttachOptions(const unordered_map<string, Value> &options, const AccessMode default_access_mode);
 
@@ -77,6 +78,8 @@ struct AttachOptions {
 	AttachVisibility visibility = AttachVisibility::SHOWN;
 	//! The stored database path (in the path manager)
 	unique_ptr<StoredDatabasePath> stored_database_path;
+	//! Per-database override of vacuum_rebuild_indexes. If not set, the global setting value is used.
+	optional_idx vacuum_rebuild_indexes_threshold;
 };
 
 //! The AttachedDatabase represents an attached database instance.
@@ -131,6 +134,14 @@ public:
 	AttachVisibility GetVisibility() const {
 		return visibility;
 	}
+	//! vacuum_rebuild_indexes threshold for this attached database.
+	//! Falls back to the global VacuumRebuildIndexesSetting if not overridden.
+	idx_t GetVacuumRebuildIndexThreshold() const {
+		if (vacuum_rebuild_threshold.IsValid()) {
+			return vacuum_rebuild_threshold.GetIndex();
+		}
+		return Settings::Get<VacuumRebuildIndexesSetting>(db);
+	}
 	const unordered_map<string, Value> &GetAttachOptions() const {
 		return attach_options;
 	}
@@ -156,6 +167,7 @@ private:
 	bool is_initial_database = false;
 	bool is_closed = false;
 	shared_ptr<mutex> close_lock;
+	optional_idx vacuum_rebuild_threshold;
 	unordered_map<string, Value> attach_options;
 
 private:
--- a/src/include/duckdb/main/attached_database.hpp
+++ b/src/include/duckdb/main/attached_database.hpp
@@ -136,7 +136,6 @@ public:
 	//! vacuum_rebuild_indexes threshold for this attached database.
 	//! Falls back to the global VacuumRebuildIndexesSetting if not overridden.
 	idx_t GetVacuumRebuildIndexThreshold() const;
-	void SetVacuumRebuildIndexThreshold(idx_t threshold);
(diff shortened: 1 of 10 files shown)


artjomPlaunov:
Thanks for the PR Evert! Looks good to me, just one comment :-)

edit: more than one comment

src/include/duckdb/main/attached_database.hpp
```diff
@@ -77,6 +78,8 @@ struct AttachOptions {
 	AttachVisibility visibility = AttachVisibility::SHOWN;
 	//! The stored database path (in the path manager)
 	unique_ptr<StoredDatabasePath> stored_database_path;
+	//! Per-database override of vacuum_rebuild_indexes. If not set, the global setting value is used.
+	optional_idx vacuum_rebuild_indexes_threshold;
```
artjomPlaunov, line 81: This is a bit of an extreme edge case, but if the user tries to set the max UBIGINT threshold value, 

```sql
statement ok
ATTACH '__TEST_DIR__/vacuum_rebuild_indexes_max.db' AS db1 (vacuum_rebuild_indexes 18446744073709551615)
```

then there will be an internal error/crash, since this is actually the invalid_idx for optional_idx. 

Kind of niche edge case, but also I could imagine someone trying to set the max 64 bit int value as a sentinel for "always rebuild", either that, or we document and say don't use this as the max value for always rebuilding (just pick a large number just not this one). I think it doesn't matter too much since this is an experimental feature that will be deprecated once we have vacuuming by default, but just pointing it out in case.

test/sql/storage/vacuum/vacuum_rebuild_indexes_attach_overrides_global.test_slow
```diff
@@ -0,0 +1,54 @@
...
+----
+122880	122880
+
+query II
+SELECT i, rowid FROM db1.t WHERE i = 245759
+----
+245759	245759
+
+statement ok
+DETACH db1
```
artjomPlaunov, line 54: does the setting work for ATTACH or REPLACE as well, i.e. what happens if we ATTACH with a setting, then ATTACH or REPLACE with a new threshold?

src/include/duckdb/main/attached_database.hpp
```diff
@@ -131,6 +134,14 @@ class AttachedDatabase : public CatalogEntry, public enable_shared_from_this<Att
 	AttachVisibility GetVisibility() const {
 		return visibility;
 	}
+	//! vacuum_rebuild_indexes threshold for this attached database.
+	//! Falls back to the global VacuumRebuildIndexesSetting if not overridden.
+	idx_t GetVacuumRebuildIndexThreshold() const {
```
artjomPlaunov: nit: do we perhaps want to move this into the implementation file?

src/main/attached_database.cpp
```diff
@@ -31,8 +32,12 @@ void StoredDatabasePath::OnDetach() {
 //===--------------------------------------------------------------------===//
 // Attach Options
 //===--------------------------------------------------------------------===//
-AttachOptions::AttachOptions(const DBConfigOptions &options)
-    : access_mode(options.access_mode), db_type(options.database_type) {
+AttachOptions::AttachOptions(const DBConfig &config)
+    : access_mode(config.options.access_mode), db_type(config.options.database_type) {
+	auto threshold = Settings::Get<VacuumRebuildIndexesSetting>(config);
```
artjomPlaunov: Is it necessary to change the constructor API here/have this logic in this constructor? Would the code work as is without this, since we already set it on the explicit attach path, and it falls back to the global value otherwise?

artjomPlaunov: i.e. if I'm not mistaken I think we can leave this overload of the constructor as it was before

carlopi:
One comment, it seems very specific the current opt-in (`vacuum_rebuild_indexes`), maybe something more generic like a `checkpoint_options = {vacuum_rebuild_indexes = X}`? So that a more generic place to bundle checkpoint/vacuum related options? Then changing names / semantic, or even supporting both old and new syntax becomes somewhat easier.

taniabogatsch:
> One comment, it seems very specific the current opt-in

This is a setting that will get deprecated again/change once we're able to do vacuuming efficiently in the background. So I don't think we need to introduce a new "checkpoint settings" setting when we don't even know what other checkpoint settings would be. Especially because existing ones that could maybe be classified as such are already in their own settings.

src/include/duckdb/main/attached_database.hpp
taniabogatsch: > or we document and say don't use this as the max value for always rebuilding

I would go with this and catch it/disallow setting the max 64 bit value.

artjomPlaunov (approved):
LGTM, thank you!

src/include/duckdb/main/attached_database.hpp
evertlammerts: Added a check!

src/include/duckdb/main/attached_database.hpp
evertlammerts: Done!

src/main/attached_database.cpp
evertlammerts: Good point, fixed!

test/sql/storage/vacuum/vacuum_rebuild_indexes_attach_overrides_global.test_slow
evertlammerts: It didn't, now it does. ATTACH or REPLACE is a bit of a tricky path I think, but it should be fine to re-apply this.

evertlammerts:
Pulled in latest v1.5, re-running tests, then (if all well) ready to land. Thanks for the feedback @artjomPlaunov !

evertlammerts:
I can't seem to get CI to run...

Edit: https://www.githubstatus.com/

evertlammerts:
Failures should be unrelated but re-running CI to make sure

lnkuiper:
LGTM, thanks!

PR #22693, opened 2026-05-15
Fix Jemalloc build

NiclasHaderer:
Two regressions introduced by #22558 masking each other, leaving jemalloc silently disabled: the build compiles, but `allocator_jemalloc.cpp` is never picked up and `duckdb_jemalloc` is never linked.

**Bug:**
The block that sets `JEMALLOC_ENABLED` lives at the bottom of `CMakeLists.txt`, but `src/CMakeLists.txt` and `src/common/allocator/CMakeLists.txt` both read it during `add_subdirectory(src)`, before the gate runs. As a result `allocator_standard.cpp` is compiled, but `duckdb_jemalloc` is dropped from `DUCKDB_LINK_LIBS`. This hides the missing `JemallocCTL` helper which was lost during the refactoring.


Maxxen:
Hello! I believe we have this in progress [here](https://github.com/duckdb/duckdb/pull/22628)

NiclasHaderer:
happy to close mine if theirs gets merged

Mytherin:
Fixed now in https://github.com/duckdb/duckdb/pull/22628

PR #22694, opened 2026-05-15
Postgres Compatibility For SETOF Returning Functions

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/8675

This PR aligns aliasing behavior with Postgres for `generate_series`, and `unnest`. With that, both `generate_series(...) t` and `generate_series(...) t(i)` are interpreted with `i` being the column name.

Previously:
```sql
select i from generate_series(1, 2) i;
┌────────────────────────────────┐
│               i                │
│ struct(generate_series bigint) │
├────────────────────────────────┤
│ {'generate_series': 1}         │
│ {'generate_series': 2}         │
└────────────────────────────────┘
```

This PR/Postgres:
```sql
select i from generate_series(1, 2) i;
 
┌───────┐
│   i   │
│ int64 │
├───────┤
│     1 │
│     2 │
│     3 │
└───────┘
```

### DIFF
--- a/extension/icu/icu-table-range.cpp
+++ b/extension/icu/icu-table-range.cpp
@@ -247,6 +247,7 @@ struct ICUTableRange {
 		    RangeDateTimeLocalInit);
 		generate_series_function.in_out_function = ICUTableRangeFunction<true>;
 		generate_series_function.cardinality = Cardinality;
+		generate_series_function.postgres_setof_compat = true;
 		generate_series.AddFunction(generate_series_function);
 
 		loader.RegisterFunction(generate_series);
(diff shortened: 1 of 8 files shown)


extension/icu/icu-table-range.cpp
```diff
@@ -247,6 +247,7 @@ struct ICUTableRange {
 		    RangeDateTimeLocalInit);
 		generate_series_function.in_out_function = ICUTableRangeFunction<true>;
 		generate_series_function.cardinality = Cardinality;
+		generate_series_function.postgres_setof_compat = true;
```
Mytherin: Can we turn this into an enum and name this as:

```cpp
enum class TableFunctionReturnType {
    TABLE_RETURNING_FUNCTION,
     SET_RETURNING_FUNCTION
};
```

Mytherin (commented):
Thanks! Looks good - one comment about naming

Mytherin:
Thanks!

PR #22696, opened 2026-05-15
Add make targets to run test configs easily

smvv:
- Add separated make targets for test configs.
- Use log groups and keep them opened if a test config run failed.

Screenshot 2026-05-15 at 16 55 15

The log groups stay open when one test config failed so it's easy to spot failures in the CI logs.


Mytherin:
Thanks!

PR #22697, opened 2026-05-15
Fix timer lifetime/timing issues

taniabogatsch:
Mainly, this disentangles the timers for `commit_local_storage_latency` and `write_to_wal_latency`, of which `commit_local_storage_latency` is currently also timing the WAL writes.


Mytherin:
Thanks!

PR #22699, opened 2026-05-15
Bump ducklake extension and remove merged patches

smvv:
Fix off-by-one error for ducklake in retry transaction logic, so CI on main becomes green again.

This error only affected `linux_(amd64|arm64)_musl` CI [jobs](https://github.com/duckdb/duckdb/actions/runs/25894779395/job/76106553482) because musl is much slower than glibc, and the retry logic error is time-sensitive.

See PR: https://github.com/duckdb/ducklake/pull/1150/.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/9184.


Mytherin:
Thanks!

PR #22700, opened 2026-05-15
Optimize key-existence self joins

zurferr:
This adds a conservative CTE rewrite for key-existence self joins.

If a materialized CTE builds pairs from the same input with matching key columns and a differing value column, and all CTE consumers only use the key columns in duplicate-insensitive contexts, the pair relation can be reduced to a grouped key-variation check:

- group by the key columns
- compute `min(value)` and `max(value)`
- keep keys where `min(value) <> max(value)`

The rewrite only fires for the same logical source on both sides, equality predicates on matching key columns, a single same-column inequality predicate, and key-only CTE consumers. It also filters null keys before grouping to preserve the original `key = key` join semantics.

Validation:
- `cmake --build build/release --target unittest -j4`
- `build/release/test/unittest test/optimizer/cte_inlining.test`
- TPC-DS SF100 sorted Parquet Q95 on current `main`: `17.52s -> 2.22s`, same output hash

### DIFF
--- a/src/include/duckdb/optimizer/materialized_cte_optimizer.hpp
+++ b/src/include/duckdb/optimizer/materialized_cte_optimizer.hpp
@@ -1,0 +1,28 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// duckdb/optimizer/materialized_cte_optimizer.hpp
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/planner/logical_operator.hpp"
+
+namespace duckdb {
+
+class Optimizer;
+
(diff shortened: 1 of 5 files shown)


kryonix:
I have not really looked at the generated code in detail yet. But I don't understand why this should be part of the `cte_inlining` optimizer at all. I think this breaks separation of concern, because the only thing this PR uses is the ref count.

zurferr:
Thanks, agreed. I moved this out of `CTEInlining` into a separate materialized CTE optimizer pass using the existing `materialized_cte` optimizer slot. `CTEInlining` is back to only deciding whether to inline CTEs, and the key-existence rewrite now runs before inlining on materialized CTE definitions.

PR #22702, opened 2026-05-15
Compact UpdateInfo allocation to reduce transaction memory 

krleonid:
## Summary
- Root UpdateInfo now allocates for actual entry count (min 8, power-of-two) instead of always STANDARD_VECTOR_SIZE
- Re-allocates with doubled capacity when more rows in the same vector are updated
- Reduces transaction memory from ~98KB to ~168 bytes per updated vector (16K vector size)

## Problem
With `STANDARD_VECTOR_SIZE=16384`, each `UpdateInfo::GetAllocSize` allocates `(sizeof(sel_t) + type_size) * 16384` bytes regardless of how many rows are actually updated. For single-row updates scattered across many vectors (common in multi-tenant/multi-DB workloads), this wastes 586x more memory than needed per allocation.

## Changes
- `src/include/duckdb/transaction/update_info.hpp` — added `GetAllocSize(type_size, capacity)`, `GetCompactCapacity(count)`, and `Initialize(..., capacity)` overloads
- `src/storage/table/update_segment.cpp` — root UpdateInfo uses compact capacity; re-allocates on merge overflow
- `test/sql/update/compact_update_allocation.test` — 95 assertions covering memory, correctness, re-allocation, checkpoint, rollback, NULL handling, multiple types

## Test plan
- [x] All existing update tests pass (37 tests, 25168 assertions)
- [x] All delete tests pass (14 tests)
- [x] All transaction tests pass (72 tests)
- [x] All storage tests pass (537 tests)
- [x] All `[update]` tagged tests pass (46 tests including `larger_than_memory_update`)
- [x] New `compact_update_allocation.test` passes (95 assertions)
- [x] Memory benchmark confirms 98.65% reduction for sparse updates

### DIFF
--- a/src/storage/table/update_segment.cpp
+++ b/src/storage/table/update_segment.cpp
@@ -103,6 +103,33 @@ idx_t UpdateInfo::GetAllocSize(idx_t type_size) {
 	return AlignValue<idx_t>(sizeof(UpdateInfo) + (sizeof(sel_t) + type_size) * STANDARD_VECTOR_SIZE);
 }
 
+idx_t UpdateInfo::GetAllocSize(idx_t type_size, idx_t capacity) {
+	return AlignValue<idx_t>(sizeof(UpdateInfo) + (sizeof(sel_t) + type_size) * capacity);
+}
+
+idx_t UpdateInfo::GetCompactCapacity(idx_t count) {
+	// Round up to next power of two, with a minimum of 8, capped at STANDARD_VECTOR_SIZE
+	if (count <= 8) {
+		return 8;
+	}
+	idx_t capacity = NextPowerOfTwo(count);
+	return MinValue<idx_t>(capacity, STANDARD_VECTOR_SIZE);
+}
+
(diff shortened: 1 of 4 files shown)


src/storage/table/update_segment.cpp
```diff
@@ -1345,12 +1372,54 @@ void UpdateSegment::Update(TransactionData transaction, DuckTableEntry &table_en
...
+				auto next_pin = new_info.next.Pin();
+				auto &next_info = UpdateInfo::Get(next_pin);
+				next_info.prev = new_handle.GetBufferPointer();
+			}
+
+			// Update root pointer
+			root->info[vector_index] = new_handle.GetBufferPointer();
+			root_pointer = root->info[vector_index];
+			new_root_pin = std::move(new_handle);
+			base_info_ptr = &UpdateInfo::Get(new_root_pin);
```
lnkuiper: `UpdateSegment::Update` is already a ~150 line function. Could the re-allocation logic be moved to a separate helper function?

lnkuiper (commented):
Thanks for the PR! Good idea. I've left one comment:

lnkuiper:
LGTM now, let's run CI :)

PR #22703, opened 2026-05-15
Improve join costing for imbalanced equality joins

zurferr:
The join order cost model currently prices a join by estimated output cardinality. For highly imbalanced leaf equality joins, this can miss useful plans where attaching a small relation to a very large relation early enables better downstream runtime filtering.

This patch adds a narrow cost adjustment for leaf/leaf equality joins where one side is tiny/small and the other side is large. The adjustment is guarded by absolute size and ratio thresholds. For non-tiny small inputs, both sides must have additional equality join context so ordinary one-spoke dimension joins keep using the existing cost logic.

The patch also keeps cheap selectivity-optional min/max filters active during native row-group scans. Heavy optional runtime filters remain statistics-only at this point. This fixes the native Q72 regression without changing the sorted-Parquet plan improvement.

TPC-DS SF100 sorted Parquet, hot comparison:

| Query | Before | After |
|---|---:|---:|
| Q64 | 25.72s | 11.31s |
| Q72 | 16.68s | 2.81s |

TPC-DS SF100 native, hot comparison:

| Query | Before | After |
|---|---:|---:|
| Q64 | 57.73s | 57.49s |
| Q72 | 2.82s | 2.43s |

Validation:
- TPC-DS SF100 sorted Parquet Q64/Q72: same result hashes.
- TPC-DS SF100 native Q64/Q72: same result hashes.
- TPC-DS SF10 native all 99 queries executed; warmed reruns of apparent regressions were neutral or faster.
- `python3 scripts/ci/run_tests.py --workers 4 build/release/test/unittest`
- `make format-check-silent enum-integrity-check -j4`

### DIFF
--- a/src/optimizer/join_order/cost_model.cpp
+++ b/src/optimizer/join_order/cost_model.cpp
@@ -1,8 +1,48 @@
 #include "duckdb/optimizer/join_order/join_node.hpp"
 #include "duckdb/optimizer/join_order/join_order_optimizer.hpp"
 #include "duckdb/optimizer/join_order/cost_model.hpp"
+#include "duckdb/planner/expression/bound_comparison_expression.hpp"
 
 namespace duckdb {
+
+struct ImbalancedJoinHeuristics {
+	static constexpr idx_t LARGE_INPUT_THRESHOLD = 1000000;
+	static constexpr idx_t TINY_INPUT_LOWER_BOUND = 10;
+	static constexpr idx_t SMALL_INPUT_CONTEXT_THRESHOLD = 100;
+	static constexpr idx_t SMALL_INPUT_THRESHOLD = 10000;
+	static constexpr double INPUT_RATIO_THRESHOLD = 100;
+};
+
+static bool HasEqualityFilter(NeighborInfo &connection) {
+	for (auto &filter : connection.filters) {
+		if (filter->filter && filter->filter->GetExpressionType() == ExpressionType::COMPARE_EQUAL &&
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks for the PR! Great performance improvements on Parquet. However, this seems to regress TPC-DS SF100 Q72 by around ~10% when using a regular DuckDB database file (not sorted, not Parquet). I wonder if this optimisation is a bit too specific? Or if there is a better way of tackling the problem that also leads to improvements on regular DuckDB database files?

zurferr:
Thanks for flagging this. I pushed an update to address the native-storage regression.

The issue was that cheap selectivity-optional min/max filters were being disabled after row-group statistics checks in native scans. The update keeps those filters active for tuple-level scan filtering, while leaving the heavier runtime filters statistics-only at that point.

Validation looks good now:

- SF100 sorted Parquet:
  - Q64: 25.72s -> 11.31s
  - Q72: 16.68s -> 2.81s
- SF100 native:
  - Q64: 57.73s -> 57.49s
  - Q72: 2.82s -> 2.43s
- Result hashes match for all measured queries.
- SF10 native all 99 queries executed; warmed reruns of apparent regressions were neutral or faster.
- `python3 scripts/ci/run_tests.py --workers 4 build/release/test/unittest`
- `make format-check-silent enum-integrity-check -j4`

lnkuiper:
I've made some big sweeping changes in the join order code. I'm interested to see if this PR still improves performance. Can you the current `main` vs. `main` with these changes updated?

PR #22705, opened 2026-05-16
Improve IsValid decision for parquet metadata cache

dentiny:
In the current implementation, whether a cache entry needs to be validated is decided on its construction, even if later users flip [`validate_external_file_cache`](https://github.com/duckdb/duckdb/blob/c7a9c8fe3bcd10a923037b8e03ab7936515e47c4/src/include/duckdb/main/settings.hpp#L1757) it won't take affect. This PR decides whether to validation based on the [caching file handle](https://github.com/duckdb/duckdb/blob/c7a9c8fe3bcd10a923037b8e03ab7936515e47c4/src/storage/external_file_cache/caching_file_system.cpp#L148-L149) (decides validation on handle constructor).

I target this PR against main instead v1.5 since it's a half fix, half improvement -- we didn't check the latest setting on each `IsValid` call anyway, the purpose for this PR  is to prevent situations where user flips from "no check" -> "check" but cache always be kept in memory.

I have already ran CI on my fork: https://github.com/dentiny/duckdb/pull/100, failed CI unrelated

### DIFF
--- a/extension/parquet/include/parquet_file_metadata_cache.hpp
+++ b/extension/parquet/include/parquet_file_metadata_cache.hpp
@@ -57,7 +57,6 @@ public:
 	ParquetCacheValidity IsValid(const OpenFileInfo &info, ClientContext &context) const;
 
 private:
-	bool validate;
 	timestamp_t last_modified;
 	string version_tag;
 };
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #22710, opened 2026-05-16
Reuse identical aggregate filters

zurferr:
Filtered aggregates currently build a separate filtered payload for every aggregate. When several aggregates use the same `FILTER` expression, the input projection can contain repeated references to the same boolean value and execution applies the same selection repeatedly.

This reuses identical non-volatile aggregate filter expressions while extracting aggregate inputs, and caches the filtered payload for the current input chunk in `AggregateFilterDataSet`. Volatile filters are left separate.

On a 50M-row generated query with five filters shared across `count`/`avg`/`avg` aggregates:

- main: 2.067s median
- this PR: 0.920s median

A same-size query with unique filters was neutral: 2.283s vs 2.299s median.

Validation:

- `python3 scripts/ci/run_tests.py --workers 4 build/release/test/unittest test/sql/filter/test_filter_clause.test_slow`
- `python3 scripts/ci/run_tests.py --workers 1 build/release/test/unittest test/sql/copy/partitioned/partitioned_group_by.test`
- `python3 scripts/ci/run_tests.py --workers 8 --test-list /tmp/duckdb_non_extension_tests.txt build/release/test/unittest`

### DIFF
--- a/src/execution/operator/aggregate/aggregate_object.cpp
+++ b/src/execution/operator/aggregate/aggregate_object.cpp
@@ -72,18 +72,48 @@ void AggregateFilterDataSet::Initialize(ClientContext &context, const vector<Agg
 		// no filters: nothing to do
 		return;
 	}
-	filter_data.resize(aggregates.size());
+	filter_map.resize(aggregates.size());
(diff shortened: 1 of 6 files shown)


zurferr:
Closing this to fold it into a broader PR that includes the optimizer rewrite using this execution improvement.

PR #22712, opened 2026-05-16
Fuse repeated scalar aggregate scans

zurferr:
This adds a scalar aggregate fusion pass for repeated scalar aggregate branches over the same serializable Parquet scan. The rewrite turns the cross product of scalar aggregate branches into a single aggregate with filtered aggregates, keeps common predicates as scan predicates, and adds an OR guard for branch-local predicates so Parquet pruning can still skip unrelated row groups.

The pass is deliberately scoped to serializable Parquet table functions for now. Repeated native table scans can still be faster once data is hot, so those are left alone.

This also reuses identical aggregate filter expressions within a chunk. That addresses the filtered-aggregate overhead that showed up in the earlier version of this change.

TPC-DS SF100 sorted Parquet, Azure 4c/16GB VM:
- Q09 hot alternating: 19.38s -> 7.29s (2.66x), max absolute result difference 5.7e-14
- Full 99-query pass: 641.94s -> 621.72s; second main pass was 636.15s
- Targeted checks for Q07/Q24/Q28/Q44/Q67 were neutral or faster, with matching hashes

Validation:
- `python3 scripts/ci/run_tests.py --workers 8 build/release/test/unittest`
- IMDB plan-cost regression: no differences detected


src/planner/binder/query_node/plan_subquery.cpp
```diff
@@ -79,6 +92,11 @@ static unique_ptr<Expression> PlanUncorrelatedSubquery(Binder &binder, BoundSubq
 		// figure out the table index of the bound table of the entry which we want to return
 		auto bindings = plan->GetColumnBindings();
 		D_ASSERT(bindings.size() == 1);
+		if (PlanReturnsExactlyOneRow(*plan)) {
+			auto result = make_uniq<BoundColumnRefExpression>(expr.GetName(), expr.GetReturnType(), bindings[0]);
+			root = LogicalCrossProduct::Create(std::move(root), std::move(plan));
```
Mytherin, line 97: This could be a separate PR - it seems unrelated aside from likely being required for the optimizer to function

src/common/enums/metric_type.cpp
```diff
@@ -57,6 +57,7 @@ profiler_settings_t MetricsUtils::GetAllMetrics() {
 		MetricType::OPTIMIZER_ROW_GROUP_PRUNER,
 		MetricType::OPTIMIZER_ROW_NUMBER_REWRITER,
 		MetricType::OPTIMIZER_SAMPLING_PUSHDOWN,
+		MetricType::OPTIMIZER_SCALAR_AGGREGATE_FUSION,
```
Mytherin, line 60: "Scalar aggregate fusion" just seems like a very specific optimization - what this really is is an extension to the common subplan optimizer to allow fusing scans like this:

```sql
SELECT <aggr1>
FROM <source>
WHERE <filter1>


SELECT <aggr2>
FROM <source>
WHERE <filter2>
```

Into a single plan like this:

```sql
SELECT <aggr1> FILTER <filter1>, <aggr2> FILTER <filter2>
FROM <source>
```

In fact - the aggregate itself being scalar has nothing to do with this, this optimization should also be able to work with a `GROUP BY` clause (why not) - as long as the `GROUP BY` clause matches. 

Perhaps you can try to simplify this PR and integrate it into the common subplan optimizer, while also generalizing the optimization further.

Mytherin (commented):
Thanks for the PR! Nice performance results - left some comments

PR #22715, opened 2026-05-17
Add Arrow export support for VARIANT

geoHeil:
## Summary
- handle `LogicalTypeId::VARIANT` in Arrow schema generation by exposing DuckDB's existing struct-backed VARIANT layout
- use the struct Arrow appender for VARIANT vectors so Arrow/ADBC result batches can be materialized
- add direct Arrow export coverage plus an ADBC regression using a stored VARIANT column

Fixes #21471.

## Tests
- `make debug`
- `build/debug/test/unittest "Test Arrow VARIANT export"`
- `DUCKDB_INSTALL_LIB=/Users/geoheil/development/forked-oss/duckdb/build/debug/src/libduckdb.dylib build/debug/test/unittest "ADBC - Query VARIANT"`

### DIFF
--- a/src/common/arrow/arrow_appender.cpp
+++ b/src/common/arrow/arrow_appender.cpp
@@ -281,6 +281,7 @@ static void InitializeFunctionPointers(ArrowAppendData &append_data, const Logic
 		InitializeAppenderForType<ArrowUnionData>(append_data);
 		break;
 	case LogicalTypeId::STRUCT:
+	case LogicalTypeId::VARIANT:
 		InitializeAppenderForType<ArrowStructData>(append_data);
 		break;
 	case LogicalTypeId::ARRAY:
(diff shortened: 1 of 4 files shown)


geoHeil:
keep active

rustyconover:
@geoHeil I took a look at this PR today, because I'm starting to desire VARIANT support but it seems like this will be a bit more complicated than just serializing the struct.  From my reading of the docs we need the binary Parquet VARIANT writer.

But before I really get into any of that, it seems like PyArrow doesn't have support for the Variant extension type yet.  Until that lands, I'm reluctant to really try this out.  I'd like to see the final Arrow C++ implementation to check my work.

rustyconover:
It is not stale.

rustyconover:
Still not stale

PR #22717, opened 2026-05-17
Propagate parent query text to MultiStatement sub-statements

bleskes:
We rely on the equivalent of `current_query()` in our infra and we noticed that statements that are re-written to a multi-statement transaction (example `ALTER TABLE ADD COLUMN <c> DEFAULT <non-literal>`, see test)  don't have the query information populated (and thus `current_query()` fails - in some of the paths, the statement [query is used](https://github.com/duckdb/duckdb/blob/6ff3b4f26cc9a916896f89167922704296ef9b6a/src/main/client_context.cpp#L1169) ) . 

This PR propagates the parent MultiStatement's query text to each child during the same wrap-up loop. It also adds a regression test that uses current_query() as the non-literal default; the test fails (empty result) without the fix and passes with it.

It is debatable if one should propagate the original query when that happens, or the underlying query that led to this (relying on statement ToString). I chose the former because it's closer to what the user types, but would like advice.

### DIFF
--- a/src/parser/parser.cpp
+++ b/src/parser/parser.cpp
@@ -7,6 +7,7 @@
 #include "duckdb/parser/query_node/select_node.hpp"
 #include "duckdb/parser/statement/create_statement.hpp"
 #include "duckdb/parser/statement/extension_statement.hpp"
+#include "duckdb/parser/statement/multi_statement.hpp"
 #include "duckdb/parser/statement/select_statement.hpp"
 #include "duckdb/parser/statement/update_statement.hpp"
 #include "duckdb/parser/tableref/expressionlistref.hpp"
@@ -366,6 +367,19 @@ void Parser::ParseQuery(const string &query) {
 			if (statement->type == StatementType::CREATE_STATEMENT) {
 				auto &create = statement->Cast<CreateStatement>();
 				create.info->sql = statement->query;
+			}
+			// Synthetic sub-statements inside a MultiStatement (e.g. CREATE TYPE generated for
+			// a PIVOT, or the ALTER+UPDATE+ALTER rewrite for ADD COLUMN ... DEFAULT
+			// <non-literal>) inherit no source offsets. Propagate the parent query text so
+			// downstream readers (current_query(), active_query->query, query history) see the
+			// SQL the user actually typed during each child's execution.
+			if (statement->type == StatementType::MULTI_STATEMENT) {
+				auto &multi = statement->Cast<MultiStatement>();
+				for (auto &child : multi.statements) {
+					child->query = statement->query;
+					child->stmt_location = 0;
+					child->stmt_length = statement->query.size();
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


bleskes:
After initially opening the PR, I found out about the yet to be released https://github.com/duckdb/duckdb/pull/22057 

That PR fixes the original problem I bumped into (the ALTER table issue), but it does so in a targeted fashion.  

The approach in this PR takes a more structural solution, but at the expense of current_query() exposing the top level user entered query. 

I added a test that uses the PIVOT statement, to show other statement still suffer from this issue. 

Please advise.

Mytherin:
Thanks for the PR! I think showing the original user query makes sense - the only issue is that statement locations now become bogus (since they refer to the actual query that is executed). We can go with your fix.

bleskes:
thanks @Mytherin - I'm not sure what you mean with locations can become bogus - [did you mean this change](https://github.com/duckdb/duckdb/pull/22717/changes#diff-6c24e708c314318559c88eb1ef461db2a9ac2b4d5bfd0d54a2444e17e947a600R380) ? note how it is being fixed to the same 0 + query.size() (which sounds off) above:

```
			statement->query = query.substr(statement->stmt_location, statement->stmt_length);
			statement->stmt_location = 0;
			statement->stmt_length = statement->query.size();
```

LMK what is needed and I'll try do it asap.

Mytherin:
When doing a multi-statement the query is reparsed, so any locations of expressions point into the new queries, not into the old one. That means any reported error locations are bogus (i.e. the `^` that points to where in a query an error occurs).

bleskes:
I see. Yeah, that's the challenge with query re-writes. It basically assumes all validation of syntactic correctness happens first, and that the newly produced queries are always correct. I can't think of a way around it. 

Do I need to change anything in the PR?

Mytherin:
I'm alright with showing the original user query - but CI is not happy with the changes right now

bleskes:
@Mytherin thanks. I will try to get CI to be happy.

bleskes:
@Mytherin  , I thought about it a bit more, after fixing the test, and I think that we are trying to have two conflicting needs (it becomes ugly where a Begin transaction statement has an empty query, or a long pivot one). The first, is represent to users what statements their query would result in / tell them where errors happened. The other is to have an internal ToString representation of the queries being run (the later is why I stumbled on this).  I think these are different enough to merit two fields, but that's a bigger change, and I'm not sure if there's enough need.

I think I prefer to just close this PR. 

I can also just fix the pivot statement in the same style as #22057 . Change is [here](https://github.com/duckdb/duckdb/compare/main...bleskes:duckdb:pivot_substatement_query?expand=1#diff-edb7ec28c10fb55b6847dd84bba0c5083cf50f6814288450436d5b9fb8d2c45cR121) 

LMK if you want me to open a new PR with the above.

Mytherin:
Sure, if that passes CI feel free to open the other one. That is lower risk for sure.

bleskes:
Closed in favor of #22769

PR #22719, opened 2026-05-18
Fix iterator invalidation in ConnectionManager::GetConnectionList

bleskes:
The previous range-based for loop called connections.erase(it.first) on the current element and then continued iterating - undefined behavior, because erase invalidates the iterator the range-based for is holding. Switch to a manual iterator loop that uses erase's return value to advance past the removed element. Update connection_count once at the end instead of on every erase.


pdet:
Thanks!

PR #22724, opened 2026-05-18
Initial implementation of `variant_keys` processing function

sebastiaan-dev:
This PR resolves https://github.com/duckdblabs/duckdb-internal/issues/9130#issue-4376008267, introducing an initial implementation for `variant_keys`. This function takes a `VARIANT`-type object as input and an optional path, returning the keys present at either the root or the supplied path in the object.

### Examples

```bash
variant_keys({'a': {'c': 1, 'd': 2}}::VARIANT)
> ['a']

variant_keys({'a': {'c': 1, 'd': 2}}::VARIANT, 'a')
> ['c', 'd']

variant_keys({'a': {'c': 1, 'd': 2}}::VARIANT, ['a', 'a'])
> [['c', 'd'], ['c', 'd']]
```

### Interface

`variant_keys(variant[, path])`

#### Parameters

- `path` is optional and of type `VARCHAR` or `VARCHAR[]`

### Follow-up work
The below points are known limitations and deferred to later PRs:

1. When supplying the variant parameter explicit casting is required (i.e. `::VARIANT`).
2. There is no support for JSONPath or JSON Pointer syntax, and one can only index a maximum of 1 level deep into an object (e.g. `variant_keys({'a': {'b': {'c': 1}}}, 'a.b')` will not work as 'a.b' is interpreted as a key in its entirety).
3. In contrast to `variant_extract` there is no support for using a numeric index to access part of a variant (e.g. `variant_keys([{'a': 1}]::VARIANT, 0)` will not work). I chose not to implement this as the equivalent `json_` function also does not support numeric indexing, instead focusing on supporting the more extensive JSONPath or JSON Pointer syntax.
4. There is no fast path for shredded variants, the function will always access the unshredded variant.

For the above points the respective issues can be found at https://github.com/duckdblabs/duckdb-internal/issues/9125#issue-4375370873


src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,377 @@
...
+			paths.push_back({VariantPathComponent(path)});
+		}
+	}
+}
+
+unique_ptr<FunctionData> VariantKeysBindData::Copy() const {
+	return make_uniq<VariantKeysBindData>(*this);
+}
+
+static bool VariantPathComponentEquals(const VariantPathComponent &a, const VariantPathComponent &b) {
```
Tishj: Maybe implement on `VariantPathComponent` instead?
```c++
bool operator == (const VariantPathComponent &other) const;
```

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,377 @@
...
+		}
+	}
+	return true;
+}
+
+struct VariantKeysResult {
+	//! By row found keys at the requested path in the VARIANT
+	vector<vector<idx_t>> key_ids_by_row;
+	//! By row flag if the requested path exists in the VARIANT
+	vector<bool> path_exists_by_row;
```
Tishj: `vector<bool>` is a bit of a pitfall to be avoided (see https://en.cppreference.com/cpp/container/vector_bool)
I don't know if it really matters here, but I think a `ValidityMask` would be a better fit, or `vector<uint8_t>` if you prefer keeping it simpler

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,377 @@
...
+	throw BinderException("'variant_keys' received an unexpected input type for the first argument: %s",
+	                      input_type.ToString());
+}
+
+static void VariantKeysFunction(DataChunk &input, ExpressionState &state, Vector &result) {
+	D_ASSERT(input.ColumnCount() == 1 || input.ColumnCount() == 2);
+	const auto count = input.size();
+
+	Vector variant_vec(LogicalType::VARIANT());
+	CastParameterToVariant(input.data[0], variant_vec, count);
```
Mytherin: This shouldn't be done here - we should push a Cast into the plan by setting the parameter type to `VARIANT` but adding an implicit cast cost that allows `JSON -> Variant` casts to be added automatically.

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+	if (arguments.size() == 1) {
+		// No path supplied, execute function over the root of the variant
+		return make_uniq<VariantKeysBindData>();
+	}
+
+	const auto &path_expr = *arguments[1];
+	const auto &return_type = path_expr.GetReturnType();
+	if (return_type.id() != LogicalTypeId::VARCHAR && return_type.id() != LogicalTypeId::LIST) {
+		throw BinderException("'variant_keys' expects the second argument to be of type VARCHAR or VARCHAR[], not %s",
```
Mytherin, line 286: Why should `path` be a constant?

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,377 @@
...
+				return false;
+			}
+		}
+	}
+	return true;
+}
+
+struct VariantKeysResult {
+	//! By row found keys at the requested path in the VARIANT
+	vector<vector<idx_t>> key_ids_by_row;
```
Mytherin: Can this be a `Vector` instead of a `vector<vector>`?

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,377 @@
...
+	variant_keys.GetSignature().GetParameter(1).SetType(LogicalType::LIST(LogicalType::VARCHAR));
+	variant_keys.SetReturnType(LogicalType::LIST(LogicalType::LIST(LogicalType::VARCHAR)));
+	fun_set.AddFunction(variant_keys);
+}
+
+ScalarFunctionSet VariantKeysFun::GetFunctions() {
+	ScalarFunctionSet fun_set;
+
+	AddFunctionsWithParameterType(fun_set, LogicalType::VARIANT());
+	AddFunctionsWithParameterType(fun_set, LogicalType::VARCHAR);
```
Mytherin: We shouldn't have this function overload - the function should only take a `VARIANT`.

Mytherin (commented):

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: I did this because the fact that casting `VARCHAR` to `VARIANT` is a little ambiguous. Do we want to create a `VARIANT` with a primitive as the root, or do we want to interpret the `VARCHAR` as an object? I thought that in the case of this function there is not really a case where you want the former, but you would expect the latter instead. Let me know what you think on how this should be handled (catch it globally where we try to parse first, and if it fails put the string as a primitive type?). I did this foremost to have similar behaviour to the `json_keys` function.

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,377 @@
...
+	for (idx_t i = 0; i < components.size(); i++) {
+		auto &component = components[i];
+		auto &input_indices = i % 2 == 0 ? value_index_sel : new_value_index_sel;
+		auto &output_indices = i % 2 == 0 ? new_value_index_sel : value_index_sel;
+
+		if (component.lookup_mode == VariantChildLookupMode::BY_INDEX) {
+			throw InternalException("'variant_keys' does not support path indexes");
+		}
+
+		(void)VariantUtils::CollectNestedData(variant, VariantLogicalType::OBJECT, input_indices, count, optional_idx(),
```
Tishj, line 102: Don't we need to check this return value?

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+	ValidityMask object_validity(count);
+	for (idx_t row_idx = 0; row_idx < count; row_idx++) {
+		if (!result.path_exists_by_row[row_idx]) {
+			object_validity.SetInvalid(row_idx);
+		}
+	}
+
+	const auto &input_indices = components.size() % 2 == 0 ? value_index_sel : new_value_index_sel;
+	(void)VariantUtils::CollectNestedData(variant, VariantLogicalType::OBJECT, input_indices, count, optional_idx(), 0,
```
Tishj: This one we definitely have to check, now you're assuming the path always ends in a valid OBJECT value, no?

Tishj (changes requested):
Thanks, I think the PR is off to a strong start, just noticed some potential problems
Also I wonder if we potentially abstract away some of logic that's shared with `variant_extract` into a function that both use?

src/function/scalar/variant/variant_keys.cpp
Mytherin: The reason this works for `json_keys` is because the `JSON` type has an implicit cast defined (`VARCHAR -> JSON`) that involves parsing the JSON. Variant does not have this implicit cast defined. That's a type-level property that has nothing to do with this specific function. I would for now just remove the `VARCHAR` path entirely and just make this operate on `VARIANT`. If we want to add the behavior that this works with a `VARCHAR` that is then parsed as a `JSON` we will add the implicit cast at a later stage.

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: Changed

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: Sounds good, removed it.

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: I like this, changed it

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: Removed

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: I used `variant_extract` as a precedent here, but I don't think it has to be. Would you want me to change this in the current PR or make an issue for it?

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: Changed it

src/function/scalar/variant/variant_keys.cpp
Mytherin: Could be an issue as well - in any case the behavior should mirror the json functions which I believe don't have this restriction, so perhaps `variant_extract` also needs to be extended to support non-const paths.

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: I don't think we need to. How I understand it `CollectNestedData` walks all the variants row-by-row, updating references inside the variant of each row along the way. Those references, along with a validity mask are the main output. The actual return value only informs us of the first row where the shape is unexpected with regards to the path that we have supplied. I think this is not an error as variants can differ in shape, instead it just means that the supplied path cannot be traced in the object at that row (which is then handled by setting the validity).

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: I use `object_validity` in `CollectNestedData` as a final pass, in combination with the earlier defined `path_validity` to distinguish between: (1) the path could not be resolved inside the variant and (2) the path could be resolved but does not point to an object (we do this because the former should produce `NULL` and the latter `[]`).

Then with the below statement (:169-:170) we only try to produce keys for rows that do not have one of the above cases:
```
if (!result.path_validity.RowIsValid(row_idx) || !object_validity.RowIsValid(row_idx)) {
  continue;
}
```

sebastiaan-dev: Let me see if I can improve the clarity of the code however.

src/function/scalar/variant/variant_keys.cpp
sebastiaan-dev: I agree, issue here: https://github.com/duckdblabs/duckdb-internal/issues/9353

src/function/scalar/variant/variant_keys.cpp
Tishj: Yea I think that's correct, nvm 👍

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -0,0 +1,396 @@
...
+	auto &key_id_child = ListVector::GetChildMutable(result.key_ids);
+	const auto key_ids_data = FlatVector::GetDataMutable<idx_t>(key_id_child);
+
+	idx_t current_offset = 0;
+	for (idx_t row_idx = 0; row_idx < count; row_idx++) {
+		auto &entry = list_entries[row_idx];
+		entry.offset = current_offset;
+		entry.length = 0;
+
+		if (!result.path_validity.RowIsValid(row_idx) || !object_validity.RowIsValid(row_idx)) {
```
Tishj: Ah nvm, I see that this is handled later
so it's fine, just a little weird to not communicate this in the intermediate result

sebastiaan-dev: You mean it's better to have another property specifically tracking this in `VariantKeysResult`?

Tishj: No, you're already tracking it
But in the intermediate keys_id_list you're creating you're not making these rows NULL, leaving them as [] instead

sebastiaan-dev: I see what you mean, so it would be better to make use of the `ValidityMask` of `key_ids` so the intermediate is more representative of what the output will be?

sebastiaan-dev: Changed it! Let me know if this is better or needs some additional work.

Tishj (approved):
Thanks, looks good to me now 👍

Mytherin:
Thanks!

PR #22725, opened 2026-05-18
Fix adaptive segment allocation resetting after checkpoint

EtgarDev:
## Summary
After a checkpoint, `InitializeAppend` was passing the last persistent segment as `prev_segment` to `AppendTransientSegment`.
Since segments are resized to `block_size` during checkpoint, the first new transient segment started at `block_size` instead of the configured `initial_column_segment_size`.

- Fix: pass `nullptr` when the last segment is persistent, so the initial size is respected.
- Adds a regression test that verifies memory stays small after a CHECKPOINT + re-insert cycle.

### DIFF
--- a/src/storage/table/column_data.cpp
+++ b/src/storage/table/column_data.cpp
@@ -505,9 +505,7 @@ void ColumnData::InitializeAppend(ColumnAppendState &state) {
 	    !last_segment.GetCompressionFunction().init_append) {
 		// we cannot append to this segment - append a new segment
 		auto total_rows = segment->GetRowStart() + last_segment.count;
-		// Persistent segments are resized to block_size during checkpoint, wo we pass nullptr to start fresh
-		const bool is_persistent = last_segment.segment_type == ColumnSegmentType::PERSISTENT;
-		AppendTransientSegment(l, total_rows, is_persistent ? nullptr : &last_segment);
+		AppendTransientSegment(l, total_rows, last_segment);
 		state.current = data.GetLastSegment(l);
 	} else {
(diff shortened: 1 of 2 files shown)


src/storage/table/column_data.cpp
```diff
@@ -574,37 +505,32 @@ void ColumnData::InitializeAppend(ColumnAppendState &state) {
 	    !last_segment.GetCompressionFunction().init_append) {
 		// we cannot append to this segment - append a new segment
 		auto total_rows = segment->GetRowStart() + last_segment.count;
-		AppendTransientSegment(l, total_rows, last_segment);
+		// Persistent segments are resized to block_size during checkpoint, wo we pass nullptr to start fresh
+		const bool is_persistent = last_segment.segment_type == ColumnSegmentType::PERSISTENT;
+		AppendTransientSegment(l, total_rows, is_persistent ? nullptr : &last_segment);
```
lnkuiper: I wonder if it's better to just past `last_segment` like before and change the check in `AppendTransientSegment` to:
```sql
if (!prev_segment || prev_segment->segment_type == ColumnSegmentType::PERSISTENT) {
```
Then the behaviour is proper for all callers.

EtgarDev: @lnkuiper I'm not sure I have a preference over her, but I do think it will be a bit cleaner to scope this constraint to this business logic.
I'll make this change

taniabogatsch (approved):

taniabogatsch:
Thanks! Great catch, let's un-draft/run CI. :)

PR #22726, opened 2026-05-18
fix: list_zip SEGFAULT with empty / NULL argument

carlopi:
Based on PR https://github.com/duckdb/duckdb/pull/22654 from @fivetran-kwoodbeck, removing the lines:
```
statement ok
CREATE TABLE variant_zip_test AS SELECT CAST(['123'] AS VARIANT[]) a, CAST([] AS VARIANT[]) b

query I
SELECT list_zip(a, b) FROM variant_zip_test
----
[(123, NULL)]

statement ok
DROP TABLE variant_zip_test
```
that make it unsuitable vs older storage formats.


lnkuiper:
Symbol leakage has been fixed in a different PR and threadsan issue is unrelated

Issue #22720, opened 2026-05-18
factorial returns 1 for negative integers instead of raising an error


kumarUjjawal:
### What happens?

factorial(n) returns 1 when n is negative, instead of raising an error or returning NULL. Factorial is mathematically undefined for negative integers, so any non-error result is  incorrect.

### To Reproduce

```  
D SELECT factorial(-1);

  ┌───────────────┐
  │ factorial(-1) │
  │    int128     │
  ├───────────────┤
  │             1 │
  └───────────────┘
```

```
  D SELECT factorial(-5);

  ┌───────────────┐
  │ factorial(-5) │
  │    int128     │
  ├───────────────┤
  │             1 │
  └───────────────┘
```

  Expected behavior

  One of:
  1. Raise an error: Invalid Input Error: factorial is not defined for negative numbers
  2. Return NULL

PostgreSQL raises ERROR: factorial of a negative number is undefined. This matches standard mathematical semantics.

### OS:

macOS (Darwin 25.4.0, arm64)

### DuckDB Version:

v1.5.2 (Variegata)

### DuckDB Client:

CLI

### Full Name:

Kumar Ujjawal

### Affiliation:

The Apache Foundation

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

DinosL:
Thanks for reporting! I have a PR up to fix this https://github.com/duckdb/duckdb/pull/22731

PR #22731, opened 2026-05-18
Return an error if a negative number is passed to the factorial function

DinosL:
Fixes: https://github.com/duckdb/duckdb/issues/22720


Mytherin:
Thanks!

PR #22732, opened 2026-05-18
Add `CONNECT` / `DISCONNECT` statement scaffolding

carlopi:
Grammar (src/parser/peg/grammar/statements/connect.gram):
```
  ConnectStatement    <- 'CONNECT' SessionTarget?
  DisconnectStatement <- 'DISCONNECT' 
  SessionTarget       <- 'LOCAL' / StringLiteral / CatalogName
```

Once combined, there will be 5 forms:
```sql
CONNECT;
CONNECT LOCAL;
CONNECT <identifier>;
CONNECT '<string>';
DISCONNECT;
```

As per this PR, those are parsed and then all variants throw some version of NotImplementedException.
Plan is for follow ups PRs to enable some subsets and explicit forbid others.

Goal would be moving towards allowing some version of:
```
ATTACH 'quack:1.2.3.4' AS my_remote (TOKEN '@#$%');
CONNECT my_remote;
PRAGMA platform;    --- executed on 1.2.3.4
DISCONNECT;
PRAGMA platform;    --- executed on 'local' duckdb

ATTACH 'postgres:<connection_string>' AS my_other_remote;
CONNECT my_other_remote EXECUTE <SOME SQL>;
```


src/include/duckdb/parser/parsed_data/disconnect_info.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+struct DisconnectInfo : public ParseInfo {
+public:
+	static constexpr const ParseInfoType TYPE = ParseInfoType::DISCONNECT_INFO;
+
+public:
+	DisconnectInfo() : ParseInfo(TYPE) {
+	}
+
+	//! Optional target — empty for bare `DISCONNECT;` and for `DISCONNECT LOCAL;` (in the latter
```
Mytherin: What does `DISCONNECT LOCAL` mean?

src/include/duckdb/parser/parsed_data/connect_info.hpp
```diff
@@ -0,0 +1,43 @@
...
+	ConnectInfo() : ParseInfo(TYPE) {
+	}
+
+	//! Target — empty for bare `CONNECT;` and for `CONNECT LOCAL;` (in the latter case,
+	//! `target_is_local` is true). Otherwise either an identifier (attached-db name) or the
+	//! contents of a string literal (connection-string form).
+	string name;
+	//! True iff the target was parsed as the LOCAL keyword (`CONNECT LOCAL;`). When true, `name`
+	//! is empty and `name_is_string_literal` is false.
+	bool target_is_local = false;
```
Mytherin, line 29: What does `CONNECT LOCAL` mean?

src/execution/operator/helper/physical_disconnect.cpp
```diff
@@ -0,0 +1,20 @@
+#include "duckdb/execution/operator/helper/physical_disconnect.hpp"
+#include "duckdb/common/exception.hpp"
+
+namespace duckdb {
+
+SourceResultType PhysicalDisconnect::GetDataInternal(ExecutionContext &context, DataChunk &chunk,
```
Mytherin, line 6: I don't think it really makes sense for `DISCONNECT` to be a logical / physical operator given how we envisioned on it being implemented, no? `DISCONNECT` will not go through the regular parsing / binding / planning phase. We can maybe add support for `DISCONNECT` to the main grammar which will just throw an error (`No client is currently connected` or so).

Mytherin (commented):
Thanks for the PR! Some questions from my side

src/include/duckdb/parser/parsed_data/disconnect_info.hpp
Mytherin: Actually what does a qualified `DISCONNECT` mean at all?

carlopi: This we sorted, no qualified `DISCONNECT`

src/include/duckdb/parser/parsed_data/connect_info.hpp
carlopi: `CONNECT LOCAL` is meant to be useful like in situations like:

```
CONNECT my_remote;
SELECT 'I am in the remote';
CONNECT LOCAL EXECUTE 'I am local';
```
where one has a way (mid-connection) to send a query to the `LOCAL` connect target (that is, just plain duckdb).

src/execution/operator/helper/physical_disconnect.cpp
carlopi: Discussed live, I think it might be OK like this. Idea is that `DISCONNECT` is still a statement, executed locally only.

Mytherin:
Thanks!

PR #22733, opened 2026-05-18
Optimize ordered TopN prefixes

zurferr:
This improves two related TopN cases.

For ordered rollup/rank queries, the optimizer can now split the ordered NULL-prefix rows from the remaining rollup rows when that prefix is guaranteed to sort first. The remaining branch is guarded by the prefix count, so the full rollup input is only materialized if the prefix does not satisfy the limit.

For multi-key TopN scans, the existing dynamic filter is extended with a stats-gated secondary filter. When row-group stats prove that the first order key has reached the global boundary, and multiple row groups share that boundary, the scan can also receive an optional dynamic filter for the second order key.

Impact measured on the Azure SF100 sorted-Parquet setup:

- TPC-DS Q67: ~49s on main to ~7s with this branch, same result hash
- Full TPC-DS SF100 sorted Parquet: 99/99, Q67 6.99s in the final run
- Existing TPC-H TopN benchmark shape (`lineitem ORDER BY l_shipdate, l_orderkey LIMIT 5`): ~2.5-3.7x faster on SF10
- Standard TPC-H queries tested were effectively neutral

Validation:

- `python3 scripts/ci/run_tests.py --workers 4 build/release/test/unittest`
- Focused TopN optimizer tests
- Full TPC-DS SF100 sorted-Parquet run: 99/99

### DIFF
--- a/extension/core_functions/aggregate/distributive/arg_min_max.cpp
+++ b/extension/core_functions/aggregate/distributive/arg_min_max.cpp
@@ -713,7 +713,7 @@ void ArgMinMaxNUpdate(Vector inputs[], AggregateInputData &aggr_input, idx_t inp
 		auto arg_val = STATE::ARG_TYPE::Create(arg_format, arg_idx);
 		auto val_val = STATE::VAL_TYPE::Create(val_format, val_idx);
 
-		state.heap.Insert(aggr_input.allocator, arg_val, val_val);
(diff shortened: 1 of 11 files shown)


Mytherin (commented):
Thanks for the PR!

These are two orthogonal changes - can you split them into two separate PRs?

Having "second_XXX" seems pretty hard-coded - why limit this to exactly two? Can we make this a `vector<...>` instead and generalize this?

Issue #22728, opened 2026-05-18
power(0.0, negative) returns inf instead of error


kumarUjjawal:
### What happens?

power(0.0, n) returns inf for any negative exponent instead of raising an error. The same applies to the ^ operator (which is an alias for power). 

DuckDB silently returns IEEE-754 inf, which can propagate through aggregations and joins and corrupt downstream results without any indication that anything went wrong.

DuckDB already raises domain errors for analogous undefined-input cases (ln(-1), log(0), sqrt(-1) all error), so this case is inconsistent with the rest of the math-function family.

### To Reproduce

 -- Returns inf instead of erroring:
```
  SELECT power(0.0::float8, -1.0::float8);
  -- ┌────────┐
  -- │ power… │
  -- │ double │
  -- ├────────┤
  -- │    inf │
  -- └────────┘
```
```
  SELECT power(0.0::float8, -2.0::float8);     -- inf
  SELECT power(0.0::float8, -0.5::float8);     -- inf
  SELECT 0.0::float8 ^ -1.0::float8;            -- inf  (operator form, same result)
```

```
  -- For comparison, other undefined-input math functions correctly error:
  SELECT ln(-1.0::float8);    -- Out of Range Error: cannot take logarithm of a negative number
  SELECT log(0.0::float8);    -- Out of Range Error: cannot take logarithm of zero
  SELECT sqrt(-1.0::float8);  -- Out of Range Error: cannot take square root of a negative number
```

```
  -- For comparison, PostgreSQL:
  --   SELECT power(0.0::float8, -1.0::float8);
  --   ERROR:  zero raised to a negative power is undefined
```

### OS:

macOS (Darwin 25.4.0, arm64)

### DuckDB Version:

 v1.5.2 (Variegata) 8a5851971f

### DuckDB Client:

CLI

### Full Name:

Kumar Ujjawal

### Affiliation:

The Apache Foundation

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

DinosL:
Thanks for reporting! Opened a PR here ~~https://github.com/duckdb/duckdb/pull/22734~~

EDIT: Superseded by https://github.com/duckdb/duckdb/pull/22753

PR #22734, opened 2026-05-18
Fix: Raise error on pow(0, negative) instead of returning inf

DinosL:
fixes: https://github.com/duckdb/duckdb/issues/22728

### DIFF
--- a/extension/core_functions/scalar/math/numeric.cpp
+++ b/extension/core_functions/scalar/math/numeric.cpp
@@ -944,6 +944,9 @@ namespace {
 struct PowOperator {
 	template <class TA, class TB, class TR>
 	static inline TR Operation(TA base, TB exponent) {
+		if (base == 0.0 && exponent < 0.0) {
+			throw OutOfRangeException("zero raised to a negative power is undefined");
+		}
 		return std::pow(base, exponent);
 	}
 };
--- a/test/sql/function/numeric/test_pow.test
+++ b/test/sql/function/numeric/test_pow.test
@@ -33,3 +33,12 @@ select power(b, a) from powerme
 ----
 10.045109
 
+statement error
+SELECT power(0.0::float8, -2.0::float8);
+----
+zero raised to a negative power is undefined
+
+statement error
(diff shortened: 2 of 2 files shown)


Mytherin:
Thanks for the PR! This should likely be dependent on `IeeeFloatingPointOpsSetting` similar to other methods. I think `pow` behavior might be defined in that standard.

DinosL:
Ahh I see! After digging into this, you're right that `pow(0, y)` for `y < 0` as defined by IEEE 754 signals `divideByZero` and returns `inf` as result. That said, unless I'm missing something, it looks like `IeeeFloatingPointOpsSetting` isn't respected by any of the math functions. For example, `log(0)` should also return `inf` and signal `divideByZero`, but it unconditionally throws regardless of the setting. Would you prefer we fix all the math functions in follow up PRs, or should we keep `pow` throwing unconditionally for now to stay consistent with the current behavior?

carlopi:
For what it's worth, failures are known and already fixed in `v1.5-variegata` tip of the branch.

Mytherin:
> Ahh I see! After digging into this, you're right that pow(0, y) for y < 0 as defined by IEEE 754 signals divideByZero and returns inf as result. That said, unless I'm missing something, it looks like IeeeFloatingPointOpsSetting isn't respected by any of the math functions. For example, log(0) should also return inf and signal divideByZero, but it unconditionally throws regardless of the setting. Would you prefer we fix all the math functions in follow up PRs, or should we keep pow throwing unconditionally for now to stay consistent with the current behavior?

There have been recent PRs on this in `main` that address this, e.g.:

```sql
select log(0);
┌────────┐
│ log(0) │
│ double │
├────────┤
│   -inf │
└────────┘
```

Mytherin:
In any case - this is a behavioral change, so it shouldn't be targeting v1.5-variegata at all imo.

DinosL:
Closing this for now. I’ll reopen it against `main` later after making the changes to make this depend on IeeeFloatingPointOpsSetting.

PR #22739, opened 2026-05-18
Bump delta

smvv:
Bump delta to have https://github.com/duckdb/duckdb-delta/pull/311 log behavior on main.

Context: i'm building all extensions using prebuilt docker containers. This rust extension failed in CI and wrote all build output into a log file, and that made understanding why CI failed difficult to understand. I've changed that log behavior so it writes to the CI console on failure.


Mytherin:
Thanks!

PR #22744, opened 2026-05-18
RowGroupPruner: treat UNSET LIMIT as unbounded

### LINKED ISSUE #22656: RowGroupPruner produces wrong results for ORDER BY ... OFFSET n without LIMIT (v1.5.x)
### What happens?

When a query has `ORDER BY <col> OFFSET n` against a persisted table with more than one row group, and there is **no `LIMIT` clause**, the new `RowGroupPruner` optimizer (introduced in v1.5.0) truncates the scan to a row-group boundary, so rows past that boundary are silently dropped. The query returns far fewer rows than expected — and `OFFSET 0` returns **zero** rows.

Root cause appears to be in `src/optimizer/row_group_pruner.cpp` (still present on current `main`, commit `b424b75208`):

```cpp
void RowGroupPruner::GetLimitAndOffset(const LogicalLimit &logical_limit,
                                       optional_idx &row_limit,
                                       optional_idx &row_offset) const {
    if (logical_limit.limit_val.Type() == LimitNodeType::CONSTANT_VALUE) {
        row_limit = logical_limit.limit_val.GetConstantValue();
    } else if (logical_limit.limit_val.Type() == LimitNodeType::UNSET) {
        row_limit = 0;     // <-- treats "no LIMIT" as LIMIT 0
    }
    ...
}
// ...
optional_idx combined_limit = row_limit.IsValid()
    ? row_limit.GetIndex() + (row_offset.IsValid() ? row_offset.GetIndex() : 0)
    : optional_idx();
```

When the user supplies only `OFFSET n` and no LIMIT, `row_limit` becomes 0, so `combined_limit = 0 + n = n`. The pruner then tells the scan to stop after enough row groups to cover only `n` rows, and the user-level OFFSET skips past everything that was scanned. Suggested fix: when `limit_val.Type() == UNSET`, leave `row_limit` invalid (`optional_idx()`) so `combined_limit` stays unset.

We also see a related but more severe failure for `LIMIT n OFFSET m` against MotherDuck-backed tables (which provide `get_partition_stats`, enabling the offset-pruning branch in `CreateRowGroupReordererOptions`). For example `SELECT * FROM test_table ORDER BY id LIMIT 500 OFFSET 8000` against an 8521-row MotherDuck table returns **0** rows. That branch does not fire for vanilla DuckDB native tables, so the reproduction below covers only the simpler `UNSET → 0` case;

### To Reproduce

Run against a **persisted** database (the optimizer is a no-op for an in-memory single-segment table):

```sh
rm -f /tmp/rgp.db
duckdb /tmp/rgp.db <<'SQL'
-- 300,000 rows spans 3 default-size row groups (122,880 / 122,880 / 54,240).
CREATE TABLE t AS SELECT range AS id FROM range(300000);
CHECKPOINT;

-- Sanity check row-group layout
SELECT row_group_id, count
FROM pragma_storage_info('t')
WHERE segment_type = 'VALIDITY' AND column_id = 0
ORDER BY row_group_id;
-- 0 | 122880
-- 1 | 122880
-- 2 |  54240

-- Bug 1: OFFSET 0 with no LIMIT returns 0 rows
SELECT COUNT(*) FROM (SELECT id FROM t ORDER BY id OFFSET 0);
-- expected: 300000
-- actual:        0

-- Bug 2: OFFSET past a row-group boundary loses trailing groups
SELECT COUNT(*) FROM (SELECT id FROM t ORDER BY id OFFSET 200000);
-- expected: 100000
-- actual:    45760     -- scan stops at boundary 245,760

-- Disconfirmation: adding a sufficiently large LIMIT restores correctness
SELECT COUNT(*) FROM (SELECT id FROM t ORDER BY id LIMIT 100000000 OFFSET 200000);
-- 100000   -- correct

-- Disconfirmation: without ORDER BY the optimizer doesn't fire
SELECT COUNT(*) FROM (SELECT id FROM t OFFSET 200000);
-- 100000   -- correct
SQL
```

### OS

macOS arm64

### DuckDB Version

v1.5.2 (Variegata) 8a5851971f. Also reproduced against current `main` (`b424b75208`).

### DuckDB Client

CLI

### Hardware

_No response_

### Full Name

John Law

### Affiliation

Dexibit

Yes

- [X] Yes, I have

- [X] Yes, I have

ColtenOuO:
Fixes #22656.

`GetLimitAndOffset` treated `LimitNodeType::UNSET` as `row_limit = 0`.
Combined with `combined_limit = row_limit + row_offset`, this told the scan to stop after ~`row_offset` rows; the user-level OFFSET then skipped almost everything that was actually scanned — silently dropping the tail row groups, or returning zero rows for `OFFSET 0`.

Repro from the issue, on a persisted table with 3 row groups
(122880 / 122880 / 54240):

| Query                                | Expected | Before | After  |
|--------------------------------------|---------:|-------:|-------:|
| `... ORDER BY id OFFSET 0`           |   300000 |      0 | 300000 |
| `... ORDER BY id OFFSET 200000`      |   100000 |  45760 | 100000 |

Fix: drop the UNSET branch so `row_limit` stays invalid and `combined_limit` stays unbounded. The `offset_val` UNSET → 0 is correct and untouched (no OFFSET == OFFSET 0).

Out of scope: the related MotherDuck-only `get_partition_stats` offset-pruning bug noted in the issue.

### DIFF
--- a/src/optimizer/row_group_pruner.cpp
+++ b/src/optimizer/row_group_pruner.cpp
@@ -90,10 +90,9 @@ bool RowGroupPruner::TryOptimize(LogicalOperator &op) const {
 
 void RowGroupPruner::GetLimitAndOffset(const LogicalLimit &logical_limit, optional_idx &row_limit,
                                        optional_idx &row_offset) const {
+	// UNSET = no LIMIT = unbounded; leave row_limit invalid.
 	if (logical_limit.limit_val.Type() == LimitNodeType::CONSTANT_VALUE) {
 		row_limit = logical_limit.limit_val.GetConstantValue();
-	} else if (logical_limit.limit_val.Type() == LimitNodeType::UNSET) {
-		row_limit = 0;
 	}
 
 	if (logical_limit.offset_val.Type() == LimitNodeType::CONSTANT_VALUE) {
--- a/test/sql/topn/test_top_n_offset_no_limit_pruner.test
+++ b/test/sql/topn/test_top_n_offset_no_limit_pruner.test
@@ -1,0 +1,35 @@
+# name: test/sql/topn/test_top_n_offset_no_limit_pruner.test
+# description: RowGroupPruner must not truncate scan when LIMIT is unset
+# group: [topn]
+
+load {TEST_DIR}/top_n_offset_no_limit_pruner.duckdb
+
+# 300,000 rows -> 3 default-size row groups (122880 / 122880 / 54240)
+statement ok
+CREATE TABLE t AS SELECT range AS id FROM range(300000)
+
+restart
+
+# OFFSET 0 with no LIMIT must return all rows (previously returned 0)
+query I
+SELECT COUNT(*) FROM (SELECT id FROM t ORDER BY id OFFSET 0)
+----
+300000
+
+# OFFSET past a row-group boundary with no LIMIT must keep trailing groups
+query I
+SELECT COUNT(*) FROM (SELECT id FROM t ORDER BY id OFFSET 200000)
+----
+100000
+
+# Trailing values must be present (regression: group 2 was being dropped)
+query I
+SELECT MAX(id) FROM (SELECT id FROM t ORDER BY id OFFSET 200000)
+----
+299999
+
+# Sanity: large LIMIT + OFFSET still correct
+query I
+SELECT COUNT(*) FROM (SELECT id FROM t ORDER BY id LIMIT 100000000 OFFSET 200000)
+----
+100000


Mytherin:
Thanks - could you just retarget this to `v1.5-variegata` so this can make the bug-fix release?

ColtenOuO:
> Thanks - could you just retarget this to `v1.5-variegata` so this can make the bug-fix release?

Done!

carlopi:
@ColtenOuO: I think you would need to cherry-pick (on top of `v1.5-variegata`) only the 2 commits of yours, not all the other ones. And then force push.

ColtenOuO:
> @ColtenOuO: I think you would need to cherry-pick (on top of `v1.5-variegata`) only the 2 commits of yours, not all the other ones. And then force push.

done! sorry, I forget this.

Mytherin:
Thanks!

PR #22746, opened 2026-05-18
Break early when getting cgroupv1 CPU entry

dentiny:
Found this when I was checking a default thread number issue.


Mytherin:
Thanks!

PR #22747, opened 2026-05-19
Close FD on fcntl error

dentiny:
I was checking on filesystem options today, and found this issue by accident, otherwise we leak FD.


Mytherin:
Thanks!

PR #22751, opened 2026-05-19
Merge `v1.5-variegata` into main

Mytherin:


Mytherin:
I think this regression is down to jemalloc - I will leave it for now in the interest of getting this in.

PR #22664, opened 2026-05-14
Parquet ApplyPendingSkips really skip decode

flashmouse:
Before I implement https://github.com/duckdb/duckdb/pull/22188#issuecomment-4288312713, I'd like to raise this PR first, because without this optimization, the benchmark for https://github.com/duckdb/duckdb/pull/22188#issuecomment-4288312713 may not show an obvious improvement.

This PR can skip page I/O and decoding completely when executing ``ApplyPendingSkips``.

PR #22188, opened 2026-04-21
Parquet skip page scan

flashmouse:
I found that DuckDB Parquet doesn't seem to use page statistics to skip whole page, and I think we can optimize this.

On my macbook, I generated a Parquet file with one row group and one integer column that has 1,000,000 rows. In a benchmark query where most pages are filtered by page statistics, the time decreased obviously. Additionally, I didn't find any performance drop in `benchmark/tpch/sf1-parquet/.*`.

main:
```
build/release/benchmark/benchmark_runner "benchmark/micro/parquet/page_stats_skip_v1.benchmark"
name    run     timing
benchmark/micro/parquet/page_stats_skip_v1.benchmark    1       0.007275
benchmark/micro/parquet/page_stats_skip_v1.benchmark    2       0.007097
benchmark/micro/parquet/page_stats_skip_v1.benchmark    3       0.007314
benchmark/micro/parquet/page_stats_skip_v1.benchmark    4       0.006794
benchmark/micro/parquet/page_stats_skip_v1.benchmark    5       0.007098
```

this:
```
build/release/benchmark/benchmark_runner "benchmark/micro/parquet/page_stats_skip_v1.benchmark" 
name    run     timing
benchmark/micro/parquet/page_stats_skip_v1.benchmark    1       0.004088
benchmark/micro/parquet/page_stats_skip_v1.benchmark    2       0.004239
benchmark/micro/parquet/page_stats_skip_v1.benchmark    3       0.004325
benchmark/micro/parquet/page_stats_skip_v1.benchmark    4       0.004190
benchmark/micro/parquet/page_stats_skip_v1.benchmark    5       0.004266
```

PR #22752, opened 2026-05-19
Parquet skip page decode by previous column filter result

flashmouse:
This PR is a follow-up to #22664 and #22188. It allows the Parquet reader to skip page decoding for a column based on the filter result of the previous columns.

Looks like I added a bit too many parameters to ReadPageHeaders, though...

I tried to create a benchmark for this. In the Parquet file I used, column A has 24 pages, and column B has 48 pages. Decoding column B takes a lot more CPU time, so skipping its decode gives an obvious improvement. Here's the result:

main:
```
build/reldebug/benchmark/benchmark_runner "benchmark/micro/parquet/page_skip_select.benchmark"
name    run     timing
benchmark/micro/parquet/page_skip_select.benchmark      1       0.002723
benchmark/micro/parquet/page_skip_select.benchmark      2       0.002927
benchmark/micro/parquet/page_skip_select.benchmark      3       0.002558
benchmark/micro/parquet/page_skip_select.benchmark      4       0.002555
benchmark/micro/parquet/page_skip_select.benchmark      5       0.002473
```

this branch:
```
build/reldebug/benchmark/benchmark_runner "benchmark/micro/parquet/page_skip_select.benchmark"

name    run     timing
benchmark/micro/parquet/page_skip_select.benchmark      1       0.001906
benchmark/micro/parquet/page_skip_select.benchmark      2       0.001806
benchmark/micro/parquet/page_skip_select.benchmark      3       0.001847
benchmark/micro/parquet/page_skip_select.benchmark      4       0.001672
benchmark/micro/parquet/page_skip_select.benchmark      5       0.001746
```


Mytherin (commented):
Thanks - but this seems quite complex, do we need to push the `SelectionVector` into the page read? Wouldn't it be sufficient to integrate this into the `Skip`?

flashmouse:
> Thanks - but this seems quite complex, do we need to push the `SelectionVector` into the page read? Wouldn't it be sufficient to integrate this into the `Skip`?

Sorry, could you explain your idea in more detail?

Also, I'm a bit doubtful whether this PR is valuable. The goal of this PR is to completely avoid page data read and decode if the `SelectionVector` says a page has no data we need to use. But when I looked into DuckDB's own storage engine filter code, I noticed that `ColumnData::Select` and `ColumnData::Filter` seem to loop over multiple segments and apply the filter, meaning they still read all segment data even if some segments are completely filtered out by the `SelectionVector`, right? If we follow that same behavior, then maybe we don't need to merge this PR.

Mytherin:
The selection vector doesn't really say whether or not you need a page, it specifies which exact rows you need within a vector (2048 rows). Pages are large chunks of data, usually at least on the order of hundreds of thousands of rows, so saying you need rows `[1, 3, 7, 2022]` doesn't really help with page filtering. This would only be relevant if pages had a very small amount of rows, but I think that's very unlikely and would only help with very poorly written Parquet files, so I don't think that's worth optimizing for.

What helps with page filtering primarily is when vectors are *fully* skipped, i.e. there are no rows we need from a given vector. That scenario should be covered mostly by `Skip`. I'm not sure if we implement page-level skipping in that scenario, but that's where I would see page-level skipping being valuable.

flashmouse:
> What helps with page filtering primarily is when vectors are _fully_ skipped, i.e. there are no rows we need from a given vector. That scenario should be covered mostly by `Skip`. I'm not sure if we implement page-level skipping in that scenario, but that's where I would see page-level skipping being valuable.

I think the optimization you mentioned is already implemented in #22664, so we can close this PR. Thanks for your  comment!

PR #22753, opened 2026-05-19
Make pow IEEE 754 compliant

### LINKED ISSUE #22728: power(0.0, negative) returns inf instead of error

DinosL:
fixes: https://github.com/duckdb/duckdb/issues/22728

### DIFF
--- a/extension/core_functions/scalar/math/numeric.cpp
+++ b/extension/core_functions/scalar/math/numeric.cpp
@@ -978,14 +978,26 @@ namespace {
 struct PowOperator {
 	template <class TA, class TB, class TR>
 	static inline TR Operation(TA base, TB exponent) {
+		if (base == 0.0 && exponent < 0.0) {
+			throw OutOfRangeException("zero raised to a negative power is undefined");
+		}
+		return std::pow(base, exponent);
+	}
+};
+
+struct IEEEPowOperator {
+	template <class TA, class TB, class TR>
+	static inline TR Operation(TA base, TB exponent) {
 		return std::pow(base, exponent);
 	}
 };
 
 } // namespace
 ScalarFunction PowOperatorFun::GetFunction() {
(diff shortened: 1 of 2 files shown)


DinosL:
CI error does not seem relevant to my changes

PR #20288, opened 2025-12-22
Unify StorageVersion and SerializationVersion

ccfelius:
This PR unifies the `StorageVersion` and `SerializationVersion` since they have merged in recent DuckDB versions. It also changes hardcoded constants into semantically meaningful constants.

Starting with DuckDB `v1.5.0`, the legacy `serialization_version` (which utilized constants ranging from 1 to 7) will be phased out. This means, that where we formerly wrote the `serialization_version`, we align it with a `storage_version` and we now write a `storage_version` >= 68 (`v1.5.0)`.

We do keep the `serialization_version` around for backwards compatibility, which is specifically necessary for serialization version 1, 2 and 3, which all fall under storage version 64 (the default, < `v1.2.0`).

To  keep the storage version backwards compatible, we changed the `optional_idx storage_version` into a `StorageVersionMapping` struct:

```
struct StorageVersionMapping {
    optional_idx version;
    string version_string;
};
```

When a database version is read, we now determine the _semantically_ corresponding storage version by (i) for versions `<v1.5.0` converting the legacy `serialization_version` into its associated version string and (ii) for versions `>v1.5.0` we convert the serialized `storage_version` into its associated string.

In code paths where explicit version checks are required for any serialization version prior to version 4  / 64 (`<v1.2.0`), we look at the version string, which is input for the `GetSerializationVersionDeprecated()` method. The returned deprecated serialization constant is then used to facilitate comparisons against these old serialization versions.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/5766.

Feedback very welcome @carlopi & @Tishj

PR #22754, opened 2026-05-19
Unify storage version and serialization version

ccfelius:
Second attempt for unifying the storage and serialization version (first attempt: https://github.com/duckdb/duckdb/pull/20288).

This PR unifies the StorageVersion and SerializationVersion since they have merged in recent DuckDB versions. It also changes hardcoded constants into semantically meaningful constants.

Starting with DuckDB v2.0, the legacy `serialization_version` (which utilized constants ranging from 1 to 8) will be phased out. This means, that where we formerly wrote the `serialization_version`, we align it with a `storage_version` and we now write a storage_version >= 69 (`StorageVersion::V2_0_0`).

We do keep the `serialization_version` around for backwards compatibility, to be able to read DuckDB files written prior to v2.0.

**Important:**
1) Note that prior to this PR, we kept track of a storage version in the main header _and_ a serialization version in the two database headers. These versions often did not align. As of now, the storage version in the main header is set to 999 and will essentially not be used anymore in future duckdb versions.

2) Another issue is that `v1.0.0` translates to storage version 64, which is the storage version also used for `v1.1.0` and `v0.10.2`. So writing `v1.0.0` or `v1.1.0` effectively means writing `v0.10.2`. A concrete example for why this is problematic, is that e.g. view dependencies are persisted when written to disk in `v1.0.0` and above. But we cannot write `StorageVersion::V1_0_0` without breaking compatibility. `StorageVersion::V1_0_0` maps to 64, and in that case view dependencies will also be serialized in `StorageVersion::V0_10_2`. So with this PR, _we break compatibility of some `V0.10.2` (and any files below `V1.0.0`) and some `V1.1.0` files._

Fixes https://github.com/duckdblabs/duckdb-internal/issues/5766


lnkuiper:
Hi @Mytherin just want your input on this. The PR looks good overall but there's an issue with versions < v1.2.0, where the storage and serialization version were not yet unified. There are 3 places where we call `ShouldSerialize` with a version < 4 (before unification). These 3 places are `hyperloglog.cpp`, `parquet_multi_file_info.cpp`, and `serialize_create_info.cpp`. This can lead to issues like this:
```sql
-- file my.sql
SET storage_compatibility_version='v1.0.0';
CREATE TABLE t AS SELECT range % 10 AS i, range % 30 AS j FROM range(100);
SELECT stats(i), stats(j) FROM t LIMIT 1;
CHECKPOINT;
```
And then:
```sh
/path/to/pr/build/duckdb my.duckdb -f my.sql
~/.duckdb/cli/1.0.0/duckdb my.duckdb
Error: unable to open database "my.duckdb": Serialization Error: Unknown HyperLogLog storage type!
```
To fix this issue, we need to implement specialized pre-unification logic, for these 3 places specifically. Is this worth the effort, or should we consider letting DuckDB v2.0.0 only write database files with storage compatibility >= v1.2.0?

Mytherin:
Nice catch! I would say we should only write `v1.0.0` and `>v1.2.0` - we shouldn't break capability of writing v1.0 files.

src/planner/binder/statement/bind_create.cpp
```diff
@@ -487,7 +487,7 @@ SchemaCatalogEntry &Binder::BindCreateTriggerInfo(CreateTriggerInfo &create_trig
 	auto &attached = catalog.GetAttached();
 	if (attached.HasStorageManager()) {
 		auto &storage_manager = attached.GetStorageManager();
-		const auto since = SerializationCompatibility::FromString("v2.0.0").serialization_version;
+		const auto since = StorageCompatibility::FromString("v2.0.0").storage_version;
```
carlopi:
```suggestion
		const auto since = StorageVersion::V2_0_0;
```
? I think this is equivalent right?

carlopi (commented):
Minor comment, plus testing found an issue:
```sql
ATTACH 'abc.db' as db (STORAGE_VERSION 'v1.5.0');
CREATE TABLE db.tbl AS SELECT 52 answer;
.exit
```
```sql
ATTACH 'abc.db' as db (STORAGE_VERSION 'v2.0.0');
CREATE TABLE db.tbl2 AS SELECT 52 answer;
.exit
```
```sql
ATTACH 'abc.db' as db;
```
```
Invalid Input Error:
Deprecated Serialization Version is not found!
```

I think the reattach-bumping-storage is a very much under tested path in general.

src/planner/binder/statement/bind_create.cpp
ccfelius: yes

PR #22756, opened 2026-05-19
Auto generate transformer (Part 2/N)

Dtenwolde:
Follow-up of: https://github.com/duckdb/duckdb/pull/22555
Part of: https://github.com/duckdblabs/duckdb-internal/issues/9265

Further improving and auto generating more grammar files. Apologies for the large diff.
This PR covers most of the easily auto-generated grammar files.  
The script can now also generate nested macros such as: `Parens(List())` kind of rules (found in `vacuum.gram`) which allows for the generation of more grammar files. 

Now auto generating also these: 
- `analyze.gram`
- `attach.gram`
- `call.gram`
- `checkpoint.gram`
- `create_schema.gram`
- `create_secret.gram`
- `create_view.gram`
- `deallocate.gram`
- `execute.gram`
- `vacuum.gram`

```bash
echo "Running format-fix..."
make format-fix > /dev/null 2>&1
```
I added this to the `build_peg_grammar.sh` script to silence the output of `make format-fix` because `gen_transformer_v2.py` provides some useful output which is otherwise not visible: 1. What rules were skipped. 2. What files were generated and functions to update and 3. What functions the user should implement.

Previously, we had some `TransformEnum` rules such as: 
```cpp
	RegisterEnum<string>("OptVerbose", "verbose");
	RegisterEnum<string>("OptFreeze", "freeze");
	RegisterEnum<string>("OptAnalyze", "analyze");
	RegisterEnum<string>("OptFull", "full");
```
But this was a bit arbitrarily defined and their standard transformer rule equivalent are just simple one liners that can easily be written as well: 
```cpp
string PEGTransformerFactory::TransformOptVerbose(PEGTransformer &transformer) {
	return "verbose";
}
```
This will allow their parent rules to be more easily auto-generated. So I'm slowly phasing out the special `TransformEnum` cases.


Mytherin:
Thanks!

PR #22757, opened 2026-05-19
Populate is_generated and generation_expression in information_schema.columns

aedrianescultura:
### Summary
information_schema.columns has always exposed is_generated and generation_expression columns (as required by the SQL standard), but both were hardcoded as NULL::VARCHAR which makes them unusable for discovering generated columns.

This PR fixes that by:
- Adding is_generated (BOOLEAN) and generation_expression (VARCHAR) to the `duckdb_columns` table function, sourced from the existing ColumnDefinition::Generated() and ColumnDefinition::GeneratedExpression() internals
- Wiring information_schema.columns to use these real values: is_generated returns 'ALWAYS' or 'NEVER' (consistent with PostgreSQL), and generation_expression returns the expression string for generated columns and NULL otherwise
### Before
```
CREATE TABLE t (
             i INT,
             j INT GENERATED ALWAYS AS (i + 1),
             k VARCHAR DEFAULT 'hello'
         );

SELECT column_name, is_generated, generation_expression
FROM information_schema.columns WHERE table_name = 't';
-- is_generated and generation_expression are always NULL
```
### After
```
SELECT column_name, is_generated, generation_expression
FROM information_schema.columns WHERE table_name = 't';
-- i  | NEVER  | NULL
-- j  | ALWAYS | CAST((i + 1) AS INTEGER)
-- k  | NEVER  | NULL
```
### Testing
Extended test/sql/generated_columns/virtual/gcol_duckdb_columns.test with assertions covering both duckdb_columns and information_schema.columns.


Mytherin:
Thanks - looks good to me, but this is a new feature, can you target `main`?

aedrianescultura:
> Thanks - looks good to me, but this is a new feature, can you target `main`?

Great! Now it targets `main`

Mytherin:
Thanks!

PR #22759, opened 2026-05-19
Add AFTER INSERT trigger with REFERENCING NEW TABLE AS support

EtgarDev:
## Summary

- Adds `REFERENCING NEW TABLE AS <alias>` clause to `CREATE TRIGGER`, exposing the inserted rows as a named table in the trigger body
- `FOR EACH ROW` remains fully blocked with `NotImplementedException`
- `REFERENCING OLD TABLE AS` is blocked
- ON CONFLICT DO UPDATE is blocked when triggers use `REFERENCING NEW TABLE AS`

## How it works

At INSERT time, `ExpandAfterTriggers` builds:

```sql
WITH
    __duckdb_trigger_base AS MATERIALIZED (INSERT INTO t ... RETURNING *),
    inserted AS (SELECT * FROM __duckdb_trigger_base),
    __duckdb_trigger_1 AS (INSERT INTO audit SELECT id FROM inserted)
SELECT count_star() FROM __duckdb_trigger_base
```

The alias CTE is a view over `__duckdb_trigger_base`, letting the trigger body reference the table by the user-declared name.


lnkuiper (changes requested):
Thanks for the PR! Looks great, really exciting stuff.

I have a few things:
 - I would prefer if we make the the internal names collision-proof by construction instead of by reservation. We could consider iterating over the available CTEs and making sure that the name of the CTE we create doesn't collide with any existing CTEs. Alternatively, we could use a UUID (easy, but not as pretty).
- `DELETE` trigger + `REFERENCING NEW TABLE AS` is not allowed by PostgreSQL, but this PR silently accepts - we should also disallow this too by throwing. This should be tested (as well as `UPDATE` trigger bodies - the tests only have `INSERT` bodies).

EtgarDev:
> Thanks for the PR! Looks great, really exciting stuff.
> 
> I have a few things:
> 
> * I would prefer if we make the the internal names collision-proof by construction instead of by reservation. We could consider iterating over the available CTEs and making sure that the name of the CTE we create doesn't collide with any existing CTEs. Alternatively, we could use a UUID (easy, but not as pretty).
> * `DELETE` trigger + `REFERENCING NEW TABLE AS` is not allowed by PostgreSQL, but this PR silently accepts - we should also disallow this too by throwing. This should be tested (as well as `UPDATE` trigger bodies - the tests only have `INSERT` bodies).

Thanks for the review @lnkuiper :)

1. Great idea, I'll just concatenate a UUID as a suffix and release the restrictions.
2. I disallowed it in the next PR when I take care of `AFTER DELETE` https://github.com/EtgarDev/duckdb/pull/39 . However, I'll backport this restriction to the current PR :)

src/planner/binder/statement/bind_create.cpp
```diff
@@ -519,12 +541,16 @@ SchemaCatalogEntry &Binder::BindCreateTriggerInfo(CreateTriggerInfo &create_trig
...
-	// Set up and trigger_creation_table to detect recursive triggers during the validation.
+	// Set up trigger_creation_table to detect recursive triggers during the validation.
 	auto validation_binder = Binder::CreateBinder(context);
 	validation_binder->global_binder_state->trigger_expanded_tables.insert(table);
 	validation_binder->global_binder_state->trigger_creation_table = &table;
 	validation_binder->global_binder_state->trigger_creation_name = create_trigger_info.trigger_name;
 	auto body_copy = create_trigger_info.trigger_action->Copy();
+
+	if (!create_trigger_info.referencing_new_table.empty()) {
+		body_copy->cte_map.map[create_trigger_info.referencing_new_table] = MakeTriggerValidationCTE(table);
```
lnkuiper, line 561: This can replace a user-created CTE with the same name. Postgres lets this "shadow" the user-created CTE with the same name.

lnkuiper (commented):
Thanks for the changes! I have some more comments:
1. `UPDATE OF ... REFERENCING NEW TABLE` should be rejected (PG says `transition tables cannot be specified for triggers with column lists` but we accept it).
2. `UPDATE OF` are ignored at firing time (maybe we should throw at creation time?)

Here's a test file that tests the two points above (and the comment below):
```sql
# name: test/temp.test
# description: Temporary review repros for PR 22759

statement ok
CREATE TABLE t_update_of (i INTEGER, j INTEGER);

statement ok
CREATE TABLE log_update_of (i INTEGER);

statement error
CREATE TRIGGER tr_update_of AFTER UPDATE OF i ON t_update_of
REFERENCING NEW TABLE AS updated_rows
FOR EACH STATEMENT
    INSERT INTO log_update_of SELECT i FROM updated_rows;
----
Binder Error:

statement ok
CREATE TABLE t_before_ref (i INTEGER);

statement ok
CREATE TABLE log_before_ref (i INTEGER);

statement error
CREATE TRIGGER tr_before_ref BEFORE INSERT ON t_before_ref
REFERENCING NEW TABLE AS inserted_rows
FOR EACH STATEMENT
    INSERT INTO log_before_ref SELECT i FROM inserted_rows;
----
Binder Error:

statement ok
CREATE TABLE t_alias_shadow (i INTEGER);

statement ok
CREATE TABLE log_alias_shadow (s VARCHAR);

statement ok
CREATE TRIGGER tr_alias_shadow AFTER INSERT ON t_alias_shadow
REFERENCING NEW TABLE AS n
FOR EACH STATEMENT
    WITH n AS (SELECT 'local' AS s)
    INSERT INTO log_alias_shadow SELECT s FROM n;

statement ok
INSERT INTO t_alias_shadow VALUES (1);

query I
SELECT s FROM log_alias_shadow;
```

src/planner/binder/statement/bind_create.cpp
EtgarDev: Interesting, I didn't think of this case.
Fixed by skipping the injection.

EtgarDev:
great points, thank you!
fixed

lnkuiper:
Great! Thanks for the changes, this is now ready to go once CI passes :)

lnkuiper (approved):

PR #22765, opened 2026-05-19
Refactor / rework how metrics are gathered and emitted in the `QueryProfiler`

Mytherin:
Currently we have a static system for registering metrics: they need to be added to a JSON (`src/common/enums/metric_type.json`) after which they are added to the `MetricType` enum at compile time. This enum is then used everywhere in the metrics stack. As a result, metrics are not extensible through extensions and are essentially fixed in place for a given DuckDB version. This enum is also large and complex. There are several layers where the *value* of the enum has semantic meaning (e.g. from the values `26..99` are supposed to match the `OptimizerType` enum in layout to allow for casting to the optimizer type).

The code for *gathering* metrics, however, tries to be generic. We have a bunch of helper methods for gathering metrics in a generic manner - adding them to a hash-map (`unordered_map<MetricType, Value>`) and a bunch of methods for doing metric aggregation and similar.

The issue with our current implementation is two-fold:

#### Which metrics we gather need to be generic, but are not

Metrics should be general-purpose and extensible. Extensions should be able to easily register new metrics for the metrics system to be usable long-term. We cannot rely on everything being defined up-front as metrics are so specific to specific tasks - for example, DuckLake might want to add a metric to track `retries`, Quack might want to add socket timeouts, etc.

Currently metrics are not used for this but we instead use logging in extensions as logging is generic and extensible. However, logging is more verbose than metrics and harder to access - we need to look into and parse logs versus just having this information available in `EXPLAIN ANALYZE`. Metrics are also intended to be more lightweight than logging and not necessarily persisted like logs usually are.

Given that we have a system for setting up metrics that is becoming more important, it should be used, also in extensions. 

#### Gathering of metrics does not need to be generic, but tries very hard to be, creating unnecessarily complex code

The gathering of metrics is highly specialized and cannot really be generic. How we gather data for a metric is fully dependent on what type of metric we are gathering. For example, if we are gathering `average time` we need to do an `avg` over a bunch of measurements. If we are doing `peak memory usage` we need to do `max` over a bunch of measurements. Metrics might also just be random strings that are gathered (and potentially overwritten). They can be anything. Gathering of specific metrics is inherently a specialized task.

Currently this tries to use generic infrastructure - but that infrastructure (by virtue of being generic) is slow and verbose. Instead of having an `idx_t peak_memory_usage` we have an `unordered_map<MetricType, Value>` that we store this in. If we were to generalize metrics this would be even worse - because we would need to replace this enum with a `string` and do string look-ups / comparisons.

As such, our current implementation is essentially the wrong way around. The metrics themselves, which need to be generic, are not generic. The gathering of metrics, which is highly specialized, tries to be generic (but still needs loads of specialized code per metric because inherently metrics are specialized). 

This PR makes some steps towards resolving this - we start removing the `MetricType` as much as possible from the `QueryProfiler`, and now use specialized code to gather metrics (in `OperatorInformation`) which is more efficient than the generic code used. Only when we flush and display the metrics do we go to the actual `MetricType` (to be replaced with a generic string in a follow-up PR).

### Other Changes

* `custom_profiling_settings` is an odd-ball setting, it can set as either a *struct*, *JSON* or *list*. Whichever is used to set it then leads to a list of displayed / gathered metrics - so this all ends up being the same in the back-end. However, we store the type of the Value that was used to set up the setting and then convert the setting back to that value when printing the value. This is odd - and in my view unnecessary - so I removed this and just made it emit a `LIST`.

I am also tempted to simplify this behavior entirely and just have `custom_profiling_settings` take a `LIST` but for now I have left the input side alone for backwards compatibility.

* Removed excessive / pointless metrics. Many metrics were added by accident to `EXPLAIN ANALYZE` (likely when converting to the very generic metric structure). This resulted in some weird metrics, e.g. here's the current profiling output for Q06:

You will see that every operator in the tree has global metrics repeated, like `system_peak_temp_dir_size`, `result_set_size`, `system_peak_buffer_memory`, `cumulative_rows_scanned`, `cumulative_cardinality` - in addition to metrics that don't apply to a given operator (e.g. `operator_rows_scanned` for non-scan operators).

The metrics are also laid out unintuitively - the top-level object has global metrics, but then has `children` to recurse into operator info. I've renamed this now to `operator_info` which is more descriptive as to what is actually part of the metrics here.

```json
{
    "total_memory_allocated": 0,
    "total_bytes_written": 0,
    "total_bytes_read": 54001664,
    "system_peak_temp_dir_size": 0,
    "system_peak_buffer_memory": 90177536,
    "rows_returned": 1,
    "result_set_size": 16,
    "latency": 0.013280666,
    "wal_replay_entry_count": 0,
    "extra_info": {},
    "commit_local_storage_latency": 0.0,
    "attach_load_storage_latency": 0.0,
    "query_name": "SELECT\n    sum(l_extendedprice * l_discount) AS revenue\nFROM\n    lineitem\nWHERE\n    l_shipdate >= CAST('1994-01-01' AS date)\n    AND l_shipdate < CAST('1995-01-01' AS date)\n    AND l_discount BETWEEN 0.05\n    AND 0.07\n    AND l_quantity < 24;\n",
    "cpu_time": 0.14901529599999996,
    "checkpoint_latency": 0.0,
    "cumulative_cardinality": 228321,
    "waiting_to_attach_latency": 0.0,
    "write_to_wal_latency": 0.0,
    "attach_replay_wal_latency": 0.0,
    "blocked_thread_time": 0.0,
    "cumulative_rows_scanned": 6001215,
    "children": [
        {
            "system_peak_temp_dir_size": 0,
            "result_set_size": 16,
            "operator_type": "UNGROUPED_AGGREGATE",
            "operator_timing": 0.0005011179999999999,
            "operator_rows_scanned": 0,
            "extra_info": {
                "Aggregates": "sum_no_overflow(#0)"
            },
            "system_peak_buffer_memory": 0,
            "cumulative_rows_scanned": 6001215,
            "cpu_time": 0.14900579599999997,
            "operator_name": "UNGROUPED_AGGREGATE",
            "cumulative_cardinality": 228321,
            "operator_cardinality": 1,
            "children": [
                {
                    "cumulative_cardinality": 228320,
                    "operator_cardinality": 114160,
                    "cumulative_rows_scanned": 6001215,
                    "operator_name": "PROJECTION",
                    "cpu_time": 0.14850467799999997,
                    "operator_rows_scanned": 0,
                    "system_peak_buffer_memory": 0,
                    "extra_info": {
                        "Projections": "(l_extendedprice * l_discount)",
                        "Estimated Cardinality": "1200243"
                    },
                    "operator_timing": 0.001838812000000001,
                    "system_peak_temp_dir_size": 0,
                    "result_set_size": 913280,
                    "operator_type": "PROJECTION",
                    "children": [
                        {
                            "system_peak_temp_dir_size": 0,
                            "result_set_size": 1826560,
                            "operator_type": "TABLE_SCAN",
                            "operator_timing": 0.14666586599999998,
                            "operator_rows_scanned": 6001215,
                            "extra_info": {
                                "Table": "tpchsf1.main.lineitem",
                                "Type": "Sequential Scan",
                                "Projections": [
                                    "l_discount",
                                    "l_extendedprice"
                                ],
                                "Filters": [
                                    "l_shipdate>='1994-01-01'::DATE AND l_shipdate<'1995-01-01'::DATE",
                                    "l_discount>=0.05 AND l_discount<=0.07",
                                    "l_quantity<24.00"
                                ],
                                "Estimated Cardinality": "1200243"
                            },
                            "system_peak_buffer_memory": 0,
                            "cumulative_rows_scanned": 6001215,
                            "cpu_time": 0.14666586599999998,
                            "operator_name": "SEQ_SCAN",
                            "cumulative_cardinality": 114160,
                            "operator_cardinality": 114160,
                            "children": []
                        }
                    ]
                }
            ]
        }
    ]
}

```

If we enable detailed profiling mode all extra detailed statistics are just added to the top-level, since we effectively don't support nested metrics (aside from the operator info) - this also creates a long list of effectively unstructured / hard to read metrics. There is structure but it is not preserved when displaying the result, and everything is just kind of thrown onto a big pile. I plan to work on this in a follow-up.

```json
{
    "all_optimizers": 0.00021125099999999997,
    "physical_planner_column_binding": 0.000002416,
    "cumulative_optimizer_timing": 0.00021087599999999998,
    "physical_planner": 0.000018791,
    "physical_planner_create_plan": 0.000014583,
    "physical_planner_resolve_types": 8.4e-8,
    "planner": 0.00014883399999999998,
    "planner_binding": 0.000144834,
    "optimizer_expression_rewriter": 0.000071167,
    "optimizer_filter_pullup": 0.00000225,
    "optimizer_filter_pushdown": 0.000037375,
    "optimizer_empty_result_pullup": 0.000001625,
    "optimizer_cte_filter_pusher": 7.09e-7,
    "optimizer_regex_range": 7.09e-7,
    "optimizer_in_clause": 7.91e-7,
    "optimizer_join_order": 0.000028791,
    "optimizer_deliminator": 4.16e-7,
    "optimizer_unnest_rewriter": 0.000001791,
    "optimizer_unused_columns": 0.000014708,
    "optimizer_statistics_propagation": 0.000020542,
    "optimizer_common_subexpressions": 0.000003792,
    "optimizer_common_aggregate": 4.59e-7,
    "optimizer_column_lifetime": 0.000002125,
    "optimizer_build_side_probe_side": 0.000001375,
    "optimizer_limit_pushdown": 7.08e-7,
    "optimizer_top_n": 4.58e-7,
    "total_memory_allocated": 0,
    "optimizer_compressed_materialization": 0.0,
    "total_bytes_written": 0,
    "total_bytes_read": 54001664,
    "system_peak_temp_dir_size": 0,
    "system_peak_buffer_memory": 108462080,
    "rows_returned": 1,
    "result_set_size": 16,
    "optimizer_join_elimination": 0.000002,
    "optimizer_late_materialization": 0.000001792,
    "optimizer_common_subplan": 0.000009167,
    "optimizer_top_n_window_elimination": 0.00000175,
    "latency": 2.187764626,
    "optimizer_window_self_join": 6.67e-7,
    "optimizer_sum_rewriter": 0.000002666,
    "cumulative_cardinality": 228321,
    "blocked_thread_time": 0.0,
    "write_to_wal_latency": 0.0,
    "cpu_time": 0.055568304000000006,
    "waiting_to_attach_latency": 0.0,
    "optimizer_extension": 0.0,
    "query_name": "SELECT\n    sum(l_extendedprice * l_discount) AS revenue\nFROM\n    lineitem\nWHERE\n    l_shipdate >= CAST('1994-01-01' AS date)\n    AND l_shipdate < CAST('1995-01-01' AS date)\n    AND l_discount BETWEEN 0.05\n    AND 0.07\n    AND l_quantity < 24;\n",
    "wal_replay_entry_count": 0,
    "optimizer_materialized_cte": 0.0,
    "extra_info": {},
    "optimizer_join_filter_pushdown": 3.33e-7,
    "commit_local_storage_latency": 0.0,
    "optimizer_row_group_pruner": 3.75e-7,
    "cumulative_rows_scanned": 6001215,
    "optimizer_cte_inlining": 3.34e-7,
    "optimizer_duplicate_groups": 0.000001209,
    "attach_load_storage_latency": 0.0,
    "attach_replay_wal_latency": 0.0,
    "optimizer_reorder_filter": 3.34e-7,
    "checkpoint_latency": 0.0,
    "optimizer_sampling_pushdown": 4.58e-7,
    ...
}
```


Maxxen:
> Gathering of metrics does not need to be generic, but tries very hard to be, creating unnecessarily complex code

100%! In the larger monitoring/telemetry ecosystem, there's a pretty strong consensus around the terminology of "counters", "gauges" and "histograms" as the fundamental "instruments" used by applications to collect metrics - and metrics are all numbers. They also have "labels" which make things a bit complex, but I think  it would be great if we could start with  something similar to just generalize aggregations. E.g. a short sketch:

```c++

template<MetricType = MetricType::SUM> // (or MAX, MIN, AVG, HISTOGRAM)
struct Metric {
public:
  void Record(idx_t value);
  void Clear();
  // ...
private: 
  string name;
  shared_ptr<atomic<idx_t>> value; // or specialized based n the MetricType
}

struct MyOperator : PhysicalOperator { 

  Metric<MetricType::SUM> my_metric;
  
  MyOperator(ClientContext &ctx) { 
    my_metric = Metric<MetricType::SUM>("total_rows_scanned");
   
    auto &registry = MetricsRegistry::Get(ctx);
    
    registry.Register(my_metric);
  }

  void Scan(ClientContext &ctx, DataChunk &chunk) { 
    // ...
    my_metric.Record(chunk.GetSize());
    
    // Or with a label (essentially grouping key for the aggregation)
    my_metric.Record(chunk.GetSize(), "from_query:" + ctx.GetQueryId())
  }
}



```

szarnyasg:
I am sorely missing a section titled "Test plan"!

Mytherin:
> > Gathering of metrics does not need to be generic, but tries very hard to be, creating unnecessarily complex code
> 
> 100%! In the larger monitoring/telemetry ecosystem, there's a pretty strong consensus around the terminology of "counters", "gauges" and "histograms" as the fundamental "instruments" used by applications to collect metrics - and metrics are all numbers. They also have "labels" which make things a bit complex, but I think it would be great if we could start with something similar to just generalize aggregations. E.g. a short sketch:
> 
> ```c++
> template<MetricType = MetricType::SUM> // (or MAX, MIN, AVG, HISTOGRAM)
> struct Metric {
> public:
>   void Record(idx_t value);
>   void Clear();
>   // ...
> private: 
>   string name;
>   shared_ptr<atomic<idx_t>> value; // or specialized based n the MetricType
> }
> 
> struct MyOperator : PhysicalOperator { 
> 
>   Metric<MetricType::SUM> my_metric;
>   
>   MyOperator(ClientContext &ctx) { 
>     my_metric = Metric<MetricType::SUM>("total_rows_scanned");
>    
>     auto &registry = MetricsRegistry::Get(ctx);
>     
>     registry.Register(my_metric);
>   }
> 
>   void Scan(ClientContext &ctx, DataChunk &chunk) { 
>     // ...
>     my_metric.Record(chunk.GetSize());
>     
>     // Or with a label (essentially grouping key for the aggregation)
>     my_metric.Record(chunk.GetSize(), "from_query:" + ctx.GetQueryId())
>   }
> }
> ```

We can add that if useful, but I think many metrics are too generic for there to be that much here. We don't want atomics in many metrics because we are doing local pre-aggregations and atomics get expensive on many-core machines.

We can of course create a bunch of templated metric helpers and I can see that being useful for some generic cases - but I will leave that as future work. This is more on redesigning individual metrics than on the metric system as a whole.


> I am sorely missing a section titled "Test plan"!

Can you expand on that?

There's already testing - but it is indeed way too little to cover many of the metrics and many of the metrics are (known) to be wrong. However, due to how generic metrics are, "testing metrics" is not really something that can be done generically. We can test *specific* metrics (e.g. compare I/O metrics versus actually measured I/O) - but each of those has to be specifically tailored towards specific metrics. 

That's a good idea - but orthogonal / unrelated to the work I'm doing here which is on (partially) redesigning the metrics framework itself.

szarnyasg:
> Can you expand on that?

Was just kidding. Every other Claude-generated issue and PR comes with a section titled "Test plan" :)

Mytherin:
Sadly I wrote this myself :)

PR #22767, opened 2026-05-19
Fix `TemporaryFileManager` reported size to reflect live blocks

DinosL:
Currently, `duckdb_temporary_files()` reports the allocated file space using `max_index`, which reflects the highest allocated block index rather than the actual live data in the file. When a table is dropped, its blocks are freed, but `max_index` remains unchanged if other live blocks still exist in the file. As a result, `GetTemporaryFile()` can over-report the file size. This PR changes that to report sizes based on the current number of active blocks.

Fixes: https://github.com/duckdblabs/duckdb-internal/issues/9296#event-25701818741
Fixes: https://github.com/duckdb/duckdb/issues/22584


Mytherin:
Thanks!

PR #22768, opened 2026-05-19
[Backport] Column-level metadata loading and serialization

ywelsch:
This is a backport of #22333 to the v1.5 branch. The branch-specific adaptations are in https://github.com/duckdb/duckdb/commit/85e1d5fb6af4b0256572e137b87bc42444b912cb and the effective change is gated by an experimental setting (force_column_metadata_reuse).


src/storage/table/row_group.cpp
```diff
@@ -464,16 +528,33 @@ unique_ptr<RowGroup> RowGroup::RemoveColumn(RowGroupCollection &new_collection,
...
+		}
+		if (!row_group->column_pointers.empty()) {
+			row_group->column_pointers[target_idx] = column_pointers[i];
+		}
+		target_idx++;
 	}
-
+	if (has_per_column_metadata_blocks) {
+		row_group->per_column_metadata_blocks = per_column_metadata_blocks;
+		row_group->per_column_metadata_blocks.RemoveColumn(target_idx);
```
Mytherin: Should this not be `removed_column`? Isn't `target_idx` here equal to `columns.size() - 1`?

src/storage/table/row_group_collection.cpp
```diff
@@ -1649,17 +1645,31 @@ void RowGroupCollection::Checkpoint(TableDataWriter &writer, TableStatistics &gl
...
 		idx_t row_start = new_total_rows;
-		bool metadata_reuse = row_group_write_data.reuse_existing_metadata_blocks;
+		auto write_action = row_group_write_data.write_action;
+		auto debug_verify_blocks = Settings::Get<DebugVerifyBlocksSetting>(GetAttached().GetDatabase()) &&
+		                           dynamic_cast<SingleFileTableDataWriter *>(&checkpoint_state.writer) != nullptr;
+		std::vector<bool> reuse_column;
+		if (debug_verify_blocks) {
+			if (write_action == RowGroupWriteAction::REUSE_EXISTING_ROW_GROUP_METADATA) {
+				auto existing_column_count = entry->ReferenceNode()->GetColumnCount();
+				reuse_column.reserve(existing_column_count);
```
Mytherin: Either this needs to be `resize` or we need to use `push_back` to add elements to this vector

src/storage/table/row_group_collection.cpp
```diff
@@ -1649,17 +1645,31 @@ void RowGroupCollection::Checkpoint(TableDataWriter &writer, TableStatistics &gl
 		}
 		auto &row_group_write_data = checkpoint_state.write_data[segment_idx];
 		idx_t row_start = new_total_rows;
-		bool metadata_reuse = row_group_write_data.reuse_existing_metadata_blocks;
+		auto write_action = row_group_write_data.write_action;
+		auto debug_verify_blocks = Settings::Get<DebugVerifyBlocksSetting>(GetAttached().GetDatabase()) &&
+		                           dynamic_cast<SingleFileTableDataWriter *>(&checkpoint_state.writer) != nullptr;
+		std::vector<bool> reuse_column;
```
Mytherin: nit: we prefer to just use `vector`, not `std::vector`

src/include/duckdb/storage/table/row_group_collection.hpp
```diff
@@ -127,7 +127,10 @@ class RowGroupCollection {
 	void CommitDropTable();
 
 	vector<PartitionStatistics> GetPartitionStats() const;
-	vector<ColumnSegmentInfo> GetColumnSegmentInfo(const QueryContext &context);
+	vector<ColumnSegmentInfo>
+	GetColumnSegmentInfo(const QueryContext &context,
+	                     ColumnSegmentInfoScanType scan_type = ColumnSegmentInfoScanType::ALL) const;
```
Mytherin: This is a feature / API change that should be left out on this branch

Mytherin (commented):
Thanks - I still think this is very large for a bug-fix patch that also touches on some critical sections of the code (storage / checkpointing). Even though most of it is gated behind a flag many of the refactors are not. I'm OK with merging it but it would help if you could do extra testing on your end to ensure this works correctly and there's no other hidden issues that pop up.

src/include/duckdb/storage/table/row_group_collection.hpp
ywelsch: addressed in [fb82839](https://github.com/duckdb/duckdb/pull/22768/commits/fb82839065c6cde38d2b9f56659e069a69edbf68) (reduces testing coverage though).

src/storage/table/row_group_collection.cpp
ywelsch: muscle memory, sorry (fixed in [394bb32](https://github.com/duckdb/duckdb/pull/22768/commits/394bb32f5acf4650d7fd016813bcc81a82e5182b))

src/storage/table/row_group.cpp
ywelsch: Good catch. This was effectively trying to remove the extra metadata at target_idx = columns.size(), which does not exist, i.e. keeping the metadata as is. It's not a big deal though, as the metadata was fixed during the following checkpoint.

I've addressed this in [394bb32](https://github.com/duckdb/duckdb/pull/22768/commits/394bb32f5acf4650d7fd016813bcc81a82e5182b)

src/storage/table/row_group_collection.cpp
ywelsch: ++, fixed in [394bb32](https://github.com/duckdb/duckdb/pull/22768/commits/394bb32f5acf4650d7fd016813bcc81a82e5182b)

ywelsch:
> Thanks - I still think this is very large for a bug-fix patch that also touches on some critical sections of the code (storage / checkpointing). Even though most of it is gated behind a flag many of the refactors are not. I'm OK with merging it but it would help if you could do extra testing on your end to ensure this works correctly and there's no other hidden issues that pop up.

Agreed. If you're fine with the change (i.e. LGTM) I will merge it into our staging environments today to get some extra testing cycles (first with the flag off) to make sure there are no regressions.

Note that the fixes in [394bb32](https://github.com/duckdb/duckdb/pull/22768/commits/394bb32f5acf4650d7fd016813bcc81a82e5182b) should also make it back to the main branch (would you like me to create a PR, or can you fix this while merging 1.5 changes back to main?)

Mytherin:
Thanks!

PR #22769, opened 2026-05-19
Set query text on PIVOT MultiStatement sub-statements at construction

bleskes:
PIVOT is rewritten into a MultiStatement whose children (synthetic CREATE TYPE enum(s) plus the rewritten SELECT) inherit no query text. Any read of current_query() / active_query->query from inside a child saw an empty string.

Populate each child's query field at construction in CreatePivotStatement using the child's own ToString()


bleskes:
Thanks @Mytherin - I was just typing the following:  I fixed the test (worked around another problem with `current_query()` failing because SelectStatement::Serialize/Deserialize do not preserve the query). Not sure what the process is now. For some reason the failure doesn't run on my fork GHA but locally it passes

Mytherin:
Thanks!

PR #22771, opened 2026-05-19
MultiFileReader allow_empty

BrynjarM:
I have run into this a few times when reading partitioned datasets where some partitions simply do not have files. 
Today that usually means adding error handling outside of DuckDB, which gets a bit awkward.

~This PR adds a small read_parquet_fallback helper that behaves like read_parquet, but runs a fallback query if no files are found. For myuse case this makes it much easier to return an empty/default result for missing partitions.~

~This is a rough implementation of the idea. If it does not fit in core I can keep it as an extension, but~ I have seen similar needs discussed in #11505 and #19072, so I thought it was worth proposing.

Changed to allow_empty that returns single column "empty BOOL" with zero rows. Can be used with union all by name to get the schema you want


Mytherin:
Thanks - but I think this variant is too complex and hard to set up for users. I think the idea of allowing empty globs is fine - maybe a parameter `allow_empty = true` for the `MultiFileReader` so this works for all `read_XXX` functions - and then just always emitting a single column "empty BOOL" with zero rows?

This also shouldn't be implemented by looking at exception text.

BrynjarM:
> Thanks - but I think this variant is too complex and hard to set up for users. I think the idea of allowing empty globs is fine - maybe a parameter `allow_empty = true` for the `MultiFileReader` so this works for all `read_XXX` functions - and then just always emitting a single column "empty BOOL" with zero rows?
> 
> This also shouldn't be implemented by looking at exception text.

The exception text check was just to keep it simple for proof of concept.

Changed the approach to `allow_empty`. I tried a few different iterations, and this one ended up being the best compromise/least bad option. It’s pretty tricky to integrate this cleanly with the existing FileGlobOptions (ALLOW_EMPTY and FALLBACK_GLOB) without making the logic confusing or needing a much bigger refactor.

Mytherin:
Thanks!

PR #15623, opened 2025-01-08
Bitpacking mode info

arjenpdevries:
Version of #15622 without (I think) merge conflicts, which was a continuation of #15544 

This implements the suggested serialize/deserialize, in line with the suggestions by @Tishj (and building on his initials steps).

More detailed info in #15544

PR #22778, opened 2026-05-20
Parallelize  pragma_storage_info, and move block scanning to the physical execution

bleskes:
`pragma_storage_info` is a nice debugging tool, but when the database gets large it can interfere with ongoing queries. Specifically, it does a full metadata scan while holding the table's RowGroup lock (see `RowGroupCollection::GetColumnSegmentInfo`). That prevents any commit who wants to check point from proceeding, but without visibility to the commit's decision to checkpoint (with all the consequences of that).

This PR moves the metadata block scanning to the physical execution phase, and adds parallelization while at it. 

I followed the same pattern as the normal table scans, but introduced new container structures as the one used in table scans have a lot more going in them (filters, projections etc). 

All the unittests pass and I did some manual tests at scale but if there's any way to beef up testing, I'll be curious to hear.


Mytherin:
Thanks - I think reworking this is a great idea (and this has been on my laundry list for a while).

I've looked at this a bit and the main reason `pragma storage_info` is slow is because of `segment_info`. In particular the segment info added in https://github.com/duckdb/duckdb/pull/15623 means we are not just reading metadata, but reading all of the data of bitpacked segments (which are often many) causing this method to be many times slower than necessary as we're now reading thousands of blocks instead of the few metadata blocks the data set contains. 

Given we now have `ColumnSegmentInfoScanType` I would propose adding a `ColumnSegmentInfoScanType::EXTENDED`, and defaulting to `ColumnSegmentInfoScanType::STANDARD` which does not emit `segment_info` (or leaves it as `NULL`). I think this will also make the method much more usable with larger tables than it is now.

bleskes:
Thanks @Mytherin .. I will add that option to filter out the culumn segments. FWIW - we need it for our research of what columns go in what block (but don't need the actually storage format for the segments), but I'm fine with this taking a while, it's just for research purposes.

bleskes:
@Mytherin I added the new scan type, but I have two jives with it - none major. One - it's really an orthogonal question - include the segment info, or just look at loaded segments or so an flag based enum seems better but I couldn't find any SQL construct that would fit the bill. Second - this also feels like projection push down - if someone doesn't need segment_info they can not specify it and then we won't load it. The down side is that the default (`call table_ino(xxx)`) will not have this optimization. I'm inclined to do the projection push down, but please advise.

Mytherin:
> One - it's really an orthogonal question - include the segment info, or just look at loaded segments or so an flag based enum seems better but I couldn't find any SQL construct that would fit the bill. 

Agreed that makes sense - they are orthogonal. Maybe just a second parameter?

> Second - this also feels like projection push down - if someone doesn't need segment_info they can not specify it and then we won't load it. The down side is that the default (call table_ino(xxx)) will not have this optimization. I'm inclined to do the projection push down, but please advise.

Yeah I think this is the issue - having the default be the slow path means people will mostly use the slow path. We could internally implement it using projection pushdown instead of a flag. But however it works internally (through projection pushdown or a flag) is not super relevant - the default call to this function just needs to not include `segment_info` imo.

bleskes:
@Mytherin I think this is ready for another review. I went with two booleans because two enums felt like quite an over kill (and requires naming 6 different things instead of 2 :) ). 

LMK WYT

Mytherin:
Thanks - I think this is good to go once CI passes

bleskes:
@Mytherin I think we're there (not sure if this ping is required)

PR #22779, opened 2026-05-20
Add COPY TO sink result callback

JimChengLin:
## Summary

This PR adds an optional `copy_to_sink_with_result` callback for `COPY TO` functions that use the legacy sink/combine API.

The existing legacy `copy_to_sink` callback returns `void`, so `PhysicalCopyToFile::Sink` always reports `SinkResultType::NEED_MORE_INPUT` after flushing a legacy COPY batch. That makes it impossible for a COPY implementation to propagate `FINISHED` or `BLOCKED` through the physical sink contract. The new callback keeps the existing API intact and gives copy functions an opt-in path when they need to return a sink result.

## Behavior

- Adds `copy_to_sink_with_result_t` to `CopyFunction`.
- Initializes the new callback to `nullptr`, so existing copy functions keep their current behavior.
- Uses `copy_to_sink_with_result` in the legacy COPY batch path when it is set.
- Falls back to the existing `copy_to_sink` callback and returns `NEED_MORE_INPUT` when the new callback is not set.
- Propagates the returned `SinkResultType` back through `PhysicalCopyToFile::Sink`.
- Still runs the legacy combine callback for the local copy state after flushing the batch.

## Testing

- Adds a C++ API test that registers a custom `COPY TO` function using `copy_to_sink_with_result` instead of `copy_to_sink`. The callback returns `FINISHED` on the first chunk, and the test verifies that COPY stops before consuming the full input.
- Runs the existing C API COPY function test to verify the old `copy_to_sink` path remains unchanged.

Commands run:

```text
cmake --build build/debug --target unittest -j 8
build/debug/test/unittest "COPY TO supports sink result callbacks"
build/debug/test/unittest "Test Copy Functions in C API"
```

### DIFF
--- a/src/execution/operator/persistent/physical_copy_to_file.cpp
+++ b/src/execution/operator/persistent/physical_copy_to_file.cpp
@@ -1736,8 +1736,8 @@ SinkResultType PhysicalCopyToFile::Sink(ExecutionContext &context, DataChunk &ch
 	if (batch_analyzer.MeetsFlushCriteria()) {
 		lstate.batch_append_state.current_chunk_state.handles.clear();
 		auto &file_state_ptr = per_thread_output ? lstate.global_file_state : gstate.global_state;
-		PrepareAndFlushBatch(context.client, gstate, file_state_ptr, gstate.create_file_state_fun,
-		                     std::move(lstate.batch));
+		return PrepareAndFlushBatch(context.client, gstate, file_state_ptr, gstate.create_file_state_fun,
+		                            std::move(lstate.batch));
 	}
 
 	return SinkResultType::NEED_MORE_INPUT;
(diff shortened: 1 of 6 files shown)


JimChengLin:
It has been merged in our internal fork. The original copy func cannot do early termination if the sink have already received enough data, i.e. reach size limit etc.

Mytherin:
Thanks - but given that `copy_to_sink` is now marked as a legacy function I don't think we should be extending it, we should ideally be moving away from it. @lnkuiper might have some more thoughts there.

JimChengLin:
> Thanks - but given that `copy_to_sink` is now marked as a legacy function I don't think we should be extending it, we should ideally be moving away from it. @lnkuiper might have some more thoughts there.

I can change this PR to new flush API only since copy_to_sink is deprecated now(not in our internal fork though).

lnkuiper:
We should only use prepare/flush from now on, so let's not extend the legacy API. Can you explain a bit why the file writer needs to return `BLOCKED`/`FINISHED`?

JimChengLin:
> We should only use prepare/flush from now on, so let's not extend the legacy API. Can you explain a bit why the file writer needs to return `BLOCKED`/`FINISHED`?

Sure. The reason is that it is hard to estimate the final output size before ETL operator(copy function). Suppose the client just want 100MB data after some operators. One way to do is inserting a dedicated limit operator but that is just for row count, not the final binary size. Even there was a binary-size-based limit operator, which is still not capable to know the final size because of compression done by file format(e.g. parquet).

lnkuiper:
I wonder if this is the right direction. I don't think a `BLOCKED`/`FINISHED` call will be handled properly in all cases, as we also have partitioned writes that follow a slightly different path. We also have the `PhysicalBatchCopyToFile` operator that also does not handle `BLOCKED`/`FINISHED`.

I think the underlying problem needs to be solved differently. We have `file_size_bytes`, maybe that could be an interesting way of solving this problem.

JimChengLin:
Thanks, that makes sense.

 I agree that having the COPY function return `FINISHED`/`BLOCKED` directly is not the right abstraction for upstream. The COPY operator should own pipeline interruption, since it is the layer that can handle the different execution paths consistently, including partitioned writes and `PhysicalBatchCopyToFile`.

 Using `CopyFunction::file_size_bytes` as the writer-provided signal also sounds like a better direction. The writer can expose the current output size, and the COPY operator can decide whether to rotate or stop consuming input based on an operator-level policy. That is less flexible than allowing arbitrary sink callbacks to return a state directly, but it gives DuckDB a single place to implement the behavior correctly across COPY paths.

For our internal use case, I will keep using the current callback-based approach in our internal fork for now. I will close this PR rather than push this API upstream in its current form.

PR #22782, opened 2026-05-20
Fuzzer issues related to parser

Dtenwolde:
Fixes: https://github.com/duckdb/duckdb-fuzzer/issues/4536, https://github.com/duckdb/duckdb-fuzzer/issues/4537, https://github.com/duckdb/duckdb-fuzzer/issues/4490

Supersedes: https://github.com/duckdb/duckdb/pull/22567 (I made the same grammar changes to `TypeLiteral` to transform the cast properly)

These queries caused internal exceptions instead of giving a descriptive error message
```sql
explain (FORMAT$X) select 1;

export database '' (FORMAT X!);
```

For `PIVOT` there was some logic to check if we can do a cast, but since types are bound during the `Binder` we cannot do that (We have `UNBOUND` types at transformer time)


Dtenwolde:
Interestingly the following benchmark regresses after these changes, I had the same issue in https://github.com/duckdb/duckdb/pull/22567
I don't immediately see how that is related to the changes to `TypeLiteral`, but perhaps I am missing something? 
```sql
run
SELECT NULLIF(ADD(c4,
    NULLIF(ADD(c8,NULLIF(ADD(c1,NULLIF(ADD(c0,
    NULLIF(ADD(c2,NULLIF(ADD(c5,NULLIF(ADD(c7,
    NULLIF(ADD(c9,NULLIF(ADD(c3,NULLIF(ADD(c10,
    NULLIF(ADD(c6, c11),0)),0)),0)),0)),0)),0)),0)),0)),0)),0)),0)
FROM foo;
```

PR #22786, opened 2026-05-20
Support qualified drop schema and drop index with reserved keywords

Dtenwolde:
cc @Tishj 

Previously, dropping a schema with the name of a reserved keyword was not possible when a catalog was specified, while this should've been possible: 
```
DROP SCHEMA my_datalake.default CASCADE;
Parser Error: syntax error at or near "default"
```
The same was true for `DROP INDEX` so I improved the grammar for that case as well. 

```
QualifiedIndexName <- CatalogReservedSchemaIndex / SchemaReservedIndex / IndexName
SchemaReservedIndex <- SchemaQualification ReservedIndexName
CatalogReservedSchemaIndex <- CatalogQualification ReservedSchemaQualification ReservedIndexName
```
This is the same strategy we use for other possibly qualified entries, such as `BaseTableName`. Essentially the first entry of the qualifier cannot be a reserved keyword (`default.foo` is not allowed), but any subsequent entries can be anything (`foo.default`) is allowed.  

I suspect, now also with `Quack` that there will be more cases around nested catalog/schemas where we need to improve the grammar. For instance, the following should be a valid query but is not currently (To be addressed in future PRs): 
```
from remote.catalog.namespace.table;
```


Mytherin:
Thanks!

PR #22787, opened 2026-05-20
[Dev] Fix behavior for the `force_full_download` open-file-info option, with the new workings of the ExternalFileCache

Tishj:
This PR fixes the intended behavior of this PR to duckdb-httpfs: https://github.com/duckdb/duckdb-httpfs/pull/67

The `force_full_download` performs a GET request to cache the entire files contents and get the file size, without the need to first make a HEAD request.

`force_full_download` is acted on inside of `FileHandle::Initialize`, which is now called multiple times.
To make sure we don't perform a GET request every time, we strip the option and add a cache entry covering the entire file instead.

Also add back `ExpressionFilter::DebugToString`, for debugging.

### DIFF
--- a/src/include/duckdb/planner/filter/expression_filter.hpp
+++ b/src/include/duckdb/planner/filter/expression_filter.hpp
@@ -64,6 +64,7 @@ public:
 
 	FilterPropagateResult CheckStatistics(BaseStatistics &stats) const;
 	string ToString(const string &column_name) const;
+	string DebugToString() const;
 	bool Equals(const ExpressionFilter &other) const;
 	unique_ptr<ExpressionFilter> Copy() const;
 	unique_ptr<Expression> ToExpression(const Expression &column) const override;
--- a/src/include/duckdb/storage/external_file_cache/caching_file_system.hpp
+++ b/src/include/duckdb/storage/external_file_cache/caching_file_system.hpp
@@ -60,6 +60,10 @@ public:
 	DUCKDB_API void Seek(idx_t location);
 
 private:
+	//! Remove the 'force_full_download' option from the file handle if present, and return whether it was present
+	bool StripForceFullDownloadIfPresent();
+
+private:
 	QueryContext context;
 
 	//! The client caching file system that was used to create this CachingFileHandle
--- a/src/planner/filter/expression_filter.cpp
+++ b/src/planner/filter/expression_filter.cpp
@@ -510,6 +510,10 @@ string ExpressionFilter::ToString(const string &column_name) const {
 	return ExpressionToFriendlyString(*expr, column_name);
 }
 
+string ExpressionFilter::DebugToString() const {
+	return ExpressionToFriendlyString(*expr, "c1");
+}
+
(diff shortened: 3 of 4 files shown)


carlopi:
Aren't we here then performing a `HEAD` to get the file size?
Like, I am not sure, but isn't this moving the case like:
```
SET force_full_download = true;
FROM read_blob('https://duckdb.org/');
```
To require 1 HEAD + 1 GET?

I am not sure how many it is now, but maybe worth checking.

I get this solves the issue where we could do N GET, but maybe it's possible to keep both.

Obviously, I am not really sure.

Tishj:
`SET force_full_download = ...` is not a thing, it's only set by extensions (in this case duckdb-iceberg)

Happy to share the test we have for this on duckdb-iceberg, which I've verified this patch with:
```sql
query III nosort expected_res
select * from my_datalake.default.table_unpartitioned order by all;
----
2023-03-01	1	a
2023-03-02	2	b
2023-03-03	3	c
2023-03-04	4	d
2023-03-05	5	e
2023-03-06	6	f
2023-03-07	7	g
2023-03-08	8	h
2023-03-09	9	i
2023-03-10	10	j
2023-03-11	11	k
2023-03-12	12	l

# 0 head requests to scan all files
query I
select count(*) from duckdb_logs_parsed('HTTP') where request.type = 'HEAD';
----
0

query I
select count(*) from duckdb_logs_parsed('HTTP')
where request.type = 'GET'
and (request.url like '%.avro'
or request.url like '%.parquet')
----
14

statement ok
CALL truncate_duckdb_logs();

query III nosort expected_res
select * from my_datalake.default.table_unpartitioned order by all;
----

# No additional HEAD requests should be made
query I
select request.url from duckdb_logs_parsed('HTTP') where request.type = 'HEAD';
----


# No additional GET requests should be made
query I
SELECT
	count(*)
FROM duckdb_logs_parsed('HTTP')
WHERE request.type = 'GET' AND (request.url LIKE '%.avro' OR request.url LIKE '%.parquet')
----
0
```

carlopi:
Ok, so size is already known (provided by metadata).

Tishj:
> Ok, so size is already known (provided by metadata).

No it's not, the size is retrieved together with the GET request

I'll draw out the first and second access logic, as I understand it:

first access:
\- 0 blocks are cached, so it needs `GetFileHandle()`
\- `GetFileHandle()` ends in `HTTPFileHandle::Initialize` which acts on `force_full_download`, performing a GET request for the full content of the file, retrieving the file_size in the process.
\- We strip the `force_full_download` and add the `file_size` that was already saved.
\- We create a cached block out of the entire files contents.

second access:
\- 1 block is cached, so `GetFileHandle()` is not called
\- The `force_full_download` gets stripped, and the `file_size` gets added, from the `CachedFile` instance.
\- When `GetFileHandle()` is called, it now sees the `file_size` so it doesn't need to perform a HEAD request to get it.

Mytherin:
Thanks!

PR #22788, opened 2026-05-20
Push down type casts to file readers

### LINKED ISSUE #8310: Epic: Duckdb projection expression pushdown

Subtasks:

Upstreaming:

Additional work:
  BoundColumnRef.

### LINKED ISSUE #22738: Type pushdown isn't called for casts.
### What happens?

On a query with type pushdown and casts, casts are not propagated as pushdown_type calls and instead treated as a separate projection in the query plan.

### To Reproduce

Simplest example uses CSV because CSV reader is the only one implementing type_pushdown table function callback.

```sql
create table test(d ubigint);
insert into test (d) select * from generate_series(10);
copy (select * from test) to 'test.csv';
explain select cast(d as uhugeint) from 'test.csv';
```

```
┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│    CAST(d AS UHUGEINT)    │
│                           │
│          ~5 rows          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│       READ_CSV_AUTO       │
│    ────────────────────   │
│         Function:         │
│       READ_CSV_AUTO       │
│                           │
│       Projections: d      │
│                           │
│          ~5 rows          │
└───────────────────────────┘
```

A real world example is a `SELECT list_sum(GT) FROM ...` query where GT is `LIST(UBIGINT)`. Duckdb upcasts list type to avoid overflow, so this query produces the following plan:

```
┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│   CAST(GT AS HUGEINT[])   │
│                           │
│       ~100,000 rows       │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│        READ_VORTEX        │
│    ────────────────────   │
│   Function: Vortex Scan   │
│      Projections: GT      │
│                           │
│       ~100,000 rows       │
```

Even though our table function may support type pushdown, duckdb doesn't call it, which results in Vector-by-Vector upcasts which are slow. Our benchmarks show that for the following query roughly 70% of the CPU time is spent in `VectorCastHelpers::TryCastLoop<unsigned long, uhugeint>`.

### OS:

Debian 13 Trixie

### DuckDB Version:

1.5.2

### DuckDB Client:

N/A

### Hardware:

_No response_

### Full Name:

Mikhail Kot

### Affiliation:

Vortex

- [x] Yes, I have

- [x] Yes, I have

myrrc:
As reported in https://github.com/duckdb/duckdb/issues/22738, type casts are usually not propagated to file readers. For Vortex, this means list aggregation routines like `list_sum(GT)` (executed as `list_sum(CAST(GT AS UHUGEINT[]))`) do upcasting per-Vector instead of doing it once in Vortex.

This PR solves this by transforming `CAST(col, T) -> GET` nodes into `GET (col as T)` if there are no conflicting casts.
This PR replaces type_pushdown function with projection_expression_pushdown which can choose to push down specific projection expressions. This only uses cast pushdown as for now.

This PR also adds MultiFileInfo projection expression (casts only) pushdown. support for Parquet.

Resolves: #22738


myrrc:
You can verify the PR this way:

```sh
~/duckdb duckdb -c "explain select list_sum(GT) from '/home/myrrc/vortex/vortex-bench/data/statpopgen/100000/parquet/gnomad.genomes.v3.1.2.hgdp_tgp.chr21.parquet'"
```

Before:

```sh

┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│    list_aggr(CAST(GT AS   │
│     HUGEINT[]), 'sum')    │
│                           │
│       ~100,000 rows       │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│        PARQUET_SCAN       │
│    ────────────────────   │
│         Function:         │
│        PARQUET_SCAN       │
│                           │
│      Projections: GT      │
│                           │
│       ~100,000 rows       │
└───────────────────────────┘
```

After:

```
┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│    list_aggr(#0, 'sum')   │
│                           │
│       ~100,000 rows       │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│        PARQUET_SCAN       │
│    ────────────────────   │
│         Function:         │
│        PARQUET_SCAN       │
│                           │
│      Projections: GT      │
│                           │
│       ~100,000 rows       │
└───────────────────────────┘
```

Mytherin (commented):
Thanks - extending the usage of `type_pushdown` sounds like a good idea to me - but I think this is perhaps better done as an optimizer pass rather than as part of the binder. It would be great if this also replaces the existing implementation in `CastLogicalOperatorToTypes` which does a more limited version of this. 

As for the actual implementation - we need to find **all** casts that occur for a given column, not just all casts that occur in the operator above the `GET`. The current implementation leaves any casts that occur prior to the current operator alone which causes issues, e.g. this now crashes because we don't detect the conflict in casts as we only find the cast in the `LogicalFilter`, missing the cast in `LogicalProjection`.

```sql
COPY (SELECT '2020-01-01 02:20:30' ts) TO 'file.csv';
SELECT ts::TIMESTAMP FROM 'file.csv' WHERE ts::DATE = '2020-01-01';
```

It's also not sufficient just to find conflicts in *casts*, if the column is at any point used in a place where it is not **directly** cast to a given type there is a conflict. For example:

```sql
COPY (SELECT '2020-01-01 02:20:30' ts) TO 'file.csv';
SELECT ts, ts::DATE FROM 'file.csv';
```

In general these kinds of changes are complex and it is easy to introduce bugs so it would be great if you could add more tests as well.

As a side-note, I think the MultiFile type pushdown for Parquet is not having any effect as these queries work with Parquet somehow. Likely the type pushdown is just being silently ignored by the Parquet reader and the original type is emitted.

myrrc:
Thanks, I'll definitely add more tests today.
As for your specific example, I think it depends whether the file reader supports cast expression pushdowns, i.e. if Parquet did some type casting, this would be fine:

```
┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│   CAST(ts AS TIMESTAMP)   │
│           ~1 row          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│        PARQUET_SCAN       │
│    ────────────────────   │
│         Function:         │
│        PARQUET_SCAN       │
│                           │
│      Projections: ts      │
│                           │
│          Filters:         │
│ (CAST(ts AS DATE) = '2020 │
│       -01-01'::DATE)      │
│           ~1 row          │
└───────────────────────────┘
```

In the plan filter and projection are decoupled, so reader function can return casted timestamp.

Mytherin:
I think you might be conflating filter pushdown and type pushdown. What `type_pushdown` specifies is the result type of the given column, i.e. in that example we tell the scan `the ts column is of type DATE`. This was originally implemented for the CSV reader since types are not known there and we do auto-detection, so this basically assists it by (1) allowing it to skip auto-detection, and (2) prevent it from incorrectly detecting the wrong types. No other readers implement this at the moment.

In the given example if we are telling the scan `ts is of type DATE` then the query plan is wrong, because later on `ts` is used as a `TIMESTAMP`. Imagine if `ts` was a string inside the source file - if we cast it to `DATE` in the scan first and emit it as a `DATE` we lose information (namely the hour/minute/second component of the value) and the query produces wrong results.

myrrc:
> In the given example if we are telling the scan ts is of type DATE then the query plan is wrong

No, I'm saying if a file reader supports expression pushdown (cast pushdown specifically), we can request "ts AS TIMESTAMP" as projection, because "ts AS DATE" will be pushed as a filter, and there are no casted or direct usages of ts.

And I also believe the reason Parquet/Arrow readers doesn't support type pushdown is because of UHUGEINT/IHUGEINT. This is the case for Vortex as well. So if you're open to breaking changes for 2.0, I'd suggest type_pushdown should either behave how pushdown_complex_filter does (get a list of output types, push down those supported, let duckdb handle the rest), or how expression pushdown does (separate function can_pushdown_type).
With this API i.e. Vortex would push everything but UHUGEINT/IHUGEINT.

myrrc:
I've pushed an updated version with tests.
This still won't push down _all_ types, i.e. on `SELECT list_sum(cast(list as USMALLINT[]))` only the inner cast to small int will be propagated, but to fix this you need to make the traversal and replacement run multiple times, and I'm not sure it's the best for performance.

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+static void ReplaceInOp(LogicalOperator &op, const Replace &replacements) {
+	LogicalOperatorVisitor::EnumerateExpressions(op,
+	                                             [&](auto *expr_ptr) { ReplaceCastTypes(*expr_ptr, replacements); });
+	for (auto &child : op.children) {
+		ReplaceInOp(*child, replacements);
+	}
+}
+
+static void FindGetsWithTypePushdown(LogicalOperator &op, vector<LogicalGet *> &gets) {
```
Mytherin: nit: `vector<reference<LogicalGet>>`

src/optimizer/optimizer.cpp
```diff
@@ -201,6 +203,16 @@ void Optimizer::RunBuiltInOptimizers() {
 		plan = filter_pullup.Rewrite(std::move(plan));
 	});
 
+	/* Push down type casts in SELECT e.g. SELECT num::UHUGEINT to file readers.
+	 * This pass must run before FILTER_PUSHDOWN. After filter pushdown
+	 * get.table_filters are populated because WHERE clauses may have been
+	 * pushed. This makes type pushdown much more complex.
```
Mytherin, line 215: Optimizers should ideally always be re-entrant - so if this case is not handled in the optimizer we should at least check for this case (and bail if we encounter it).

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+TypePushdown::TypePushdown(ClientContext &context) : context(context) {
+}
+
+using GetCastMap = unordered_map<column_t, LogicalType>;
+using GetConflicts = unordered_set<column_t>;
+using GetReplace = unordered_map<column_t, LogicalType>;
+
+struct GetAnalysis {
+	LogicalGet *get;
```
Mytherin: nit: `reference<LogicalGet>`

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+ * visible as LOGICAL_FILTER nodes. That makes uncasted column usage detectable
+ * by CollectFromOp without need to inspect table_filters.
+ */
+
+namespace duckdb {
+
+TypePushdown::TypePushdown(ClientContext &context) : context(context) {
+}
+
+using GetCastMap = unordered_map<column_t, LogicalType>;
```
Mytherin: Are all of these top-level `using` clauses necessary? Looking at the code I think almost all of them can be replaced with `auto &` instead?

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+	if (get_replace_it == get_replace.end() || get_replace_it->second != bound_cast.GetReturnType()) {
+		return replace_children();
+	}
+
+	expr = make_uniq<BoundColumnRefExpression>(get_replace_it->second, bound_column.binding);
+}
+
+static void CollectFromOp(LogicalOperator &op, Analyses &analyses) {
+	LogicalOperatorVisitor::EnumerateExpressions(op, [&](auto *expr_ptr) { CollectCastTypes(**expr_ptr, analyses); });
```
Mytherin: Instead of manually doing the iteration here can we implement a `LogicalOperatorVisitor`?

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+	}
+	const auto &bound_cast = expr.Cast<BoundCastExpression>();
+
+	if (bound_cast.child->GetExpressionType() != ExpressionType::BOUND_COLUMN_REF) {
+		return collect_children();
+	}
+	const auto &bound_column = bound_cast.child->Cast<BoundColumnRefExpression>();
+
+	if (bound_column.depth > 0) {
+		return collect_children();
```
Mytherin: This should never happen at this stage - I think we can throw an internal exception

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	const auto it = analyses.find(bound_column.binding.table_index);
+	if (it == analyses.end()) {
+		return collect_children();
```
Mytherin: Should we recurse here again?

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+	// We're in a leaf
+
+	const column_t projection_id = bound_column.binding.column_index;
+	if (IsVirtualColumn(projection_id)) {
+		return;
+	}
+
+	GetAnalysis &analysis = it->second;
+	GetCastMap &cast_map = analysis.cast_map;
+	const column_t column_id = analysis.get->GetColumnIds()[projection_id].GetPrimaryIndex();
```
Mytherin: This won't work for `IsPushdownExtract()` - but it works right now because that is only generated as part of an optimizer that runs after this optimizer. To keep optimizers re-entrant / composable should at least check for `if (col_id.IsPushdownExtract())` and bail if this is the case.

Mytherin: We might also want to add a test just in case:

```sql
COPY (SELECT {'i': 42, 'j': 84} s) TO '{TEST_DIR}/struct_projection_pushdown.parquet';
SELECT s.i::UINTEGER, s.j::UBIGINT FROM "{TEST_DIR}/struct_projection_pushdown.parquet";
```

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+static void CollectCastTypes(const Expression &expr, Analyses &analyses) {
+	auto collect_children = [&] {
+		ExpressionIterator::EnumerateChildren(expr,
+		                                      [&](const Expression &child) { CollectCastTypes(child, analyses); });
+	};
+
+	if (expr.GetExpressionClass() == ExpressionClass::BOUND_COLUMN_REF) {
+		const auto &colref = expr.Cast<BoundColumnRefExpression>();
+		if (colref.depth != 0) {
+			return;
```
Mytherin: Same as before - this should never happen at this stage

test/optimizer/pushdown/csv_type_pushdown.test
```diff
@@ -0,0 +1,51 @@
+# name: test/optimizer/pushdown/csv_type_pushdown.test
+# description: Test CSV With Pushing Down Type Casts
+# group: [pushdown]
+
+# Column is used uncasted
+query II
+EXPLAIN SELECT column00 FROM read_csv('{DATA_DIR}/csv/real/lineitem_sample.csv', delim='|');
```
Mytherin, line 7: We shouldn't run only the `EXPLAIN` in these tests but also run the query to ensure this works end-to-end - can you add those tests?

test/optimizer/pushdown/parquet_type_pushdown.test
```diff
@@ -0,0 +1,183 @@
+# name: test/optimizer/pushdown/parquet_type_pushdown.test
+# description: Test Parquet With Pushing Down Type Casts
+# group: [pushdown]
+
+require parquet
+
+# Column is only used casted
+query II
+EXPLAIN SELECT smallint_col::UINTEGER FROM "{DATA_DIR}/parquet-testing/arrow/alltypes_plain.parquet";
```
Mytherin: Same here - can we also run the test itself?

Mytherin (commented):
Thanks! Looks good - I've left some comments.

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+			// Column is used uncasted
+			analysis.conflicts.insert(column_id);
+		}
+		return;
+	}
+
+	if (expr.GetExpressionClass() != ExpressionClass::BOUND_CAST) {
+		return collect_children();
+	}
+	const auto &bound_cast = expr.Cast<BoundCastExpression>();
```
Mytherin: We should check `try_cast` here as well - can you do this + add a test with `try_cast` and a mix of `cast + try_cast` on the same column?

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+ * by CollectFromOp without need to inspect table_filters.
+ */
+
+namespace duckdb {
+
+TypePushdown::TypePushdown(ClientContext &context) : context(context) {
+}
+
+using GetCastMap = unordered_map<column_t, LogicalType>;
+using GetConflicts = unordered_set<column_t>;
```
Mytherin: Let's avoid using raw `column_t` and instead use `ProjectionIndex` here - the intent then is more clear.

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+unique_ptr<LogicalOperator> TypePushdown::Optimize(unique_ptr<LogicalOperator> op) {
+	vector<LogicalGet *> gets;
+	FindGetsWithTypePushdown(*op, gets);
+	if (gets.empty()) {
+		return op;
+	}
+
+	Analyses analyses(gets.size());
+	for (size_t i = 0; i < gets.size(); ++i) {
```
Mytherin: nit: `idx_t`

src/optimizer/type_pushdown.cpp
```diff
@@ -0,0 +1,215 @@
...
+		analyses.emplace(gets[i]->table_index.index, GetAnalysis {gets[i]});
```
Mytherin: nit: no need for `table_index.index`

src/optimizer/type_pushdown.cpp
myrrc: Just a matter of taste, I prefer AAAA (always almost avoid auto), but I can remove them, not a big deal)

Mytherin: For sure, it's a nit, but in general we prefer to have a consistent code-style as it simplifies reading / understanding code in the code base without having different dialects / styles in different modules.

src/optimizer/type_pushdown.cpp
myrrc: But we're operating on column_t after all, since most methods return ColumnIndex (which isn't hashable), so constructing a ProjectionIndex is extra work.

myrrc: Also, CastMap must operate on column_t due to TableFunction API function signature.

src/optimizer/type_pushdown.cpp
myrrc: I'm not sure what should this visitor do, since the only meaningful work is calling CollectCastTypes which is already specific to the optimizer?

Mytherin: `CollectFromOp` effectively re-implements a visitor over a logical operator (iterating expressions and operator children) - I'm suggesting instead of re-implementing the visitor we use the visitor directly (i.e. sub-class from it and implement the required methods).

src/optimizer/type_pushdown.cpp
Mytherin: Ah yeah, you're right, we're operating on `column_t`. I think it would probably be better to do this on the `ProjectionIndex` level instead of resolving to the base column ids but let's save that for future work.

src/optimizer/type_pushdown.cpp
myrrc: Does try_cast need any special handling? If a cast fails, it's on the extension to propagate this correctly since cast pushdown is already an extension point. +1 for tests

Mytherin: The extension currently does not know a `TRY_CAST` was requested, so this will likely turn from working and returning `NULL` into an error:

```sql
COPY (SELECT 'hello world' varchar_col) TO 'varchar.parquet';
SELECT TRY_CAST(varchar_col AS DATE) FROM read_parquet('varchar.parquet');
```

myrrc:
@Mytherin I've pushed an updated version. The question about type casts which are not supported in extension i.e. HUGEINT for Vortex still remains - we won't be able to use pushdown if there's no way to decline the cast.

Maybe this PR should also use expression_pushdown TableFunction callback to check whether this is supported?

Mytherin:
`pushdown_expression` is about table filters so I think that's not the right abstraction here.

We could redesign the `type_pushdown` to maybe pushing a generic expression instead? e.g.:

```
bool pushdown_projection(ClientContext &context, const LogicalGet &get, const Expression &expr, idx_t proj_index);
```

That would be more flexible going forward - then we just wouldn't accept anything other than casts for now.

Mytherin:
If we redesign the function we should also move towards using a struct for input parameters (similar to `TableFunctionGetPartitionInput` and such).

myrrc:
`bool pushdown_projection(ClientContext &context, const LogicalGet &get, const Expression &expr, idx_t proj_index);`

What if multiple projections for a column were requested e.g. `SELECT len(str), substr(str, 2)`? 

I've thought about this for a while in https://github.com/vortex-data/vortex/issues/8190. If we go this route, Duckdb can request multiple same columns in projections i.e. for the query above
column_ids=[0], projection_ids=[0, 0], and pushdown_projection may be called twice:

```
pushdown_projection(.., Expression{len}, 0) -- first col in SELECT
pushdown_projection(.., Expression{len}, 1) -- second col in SELECT
```

However, if we want to further expand this to aggregate functions, we should also be notified of the output types. This shouldn't be hard, though

Mytherin:
I would say we can skip projection pushdown in that case similar to how we currently skip cast pushdown if the same column is scanned with different types - but it's an interesting angle to consider perhaps going forward.

myrrc:
Ok, I've changed the function to handle the exception, and also updated Parquet/CSV tests. Let's leave the optimizer just doing the casts pushdown for now?

Mytherin:
Yes, I think that's great- I'll kick off CI.

myrrc:
I've pushed the updated version only for CSV which passes local tests.

myrrc:
One thing I'm unsure about is the signature of expression pushdown. We have an existing api of pushdown_expression which evaluates whether we can do some thing, so on the one hand the bool return type seems excessive -- we can check whether we can push, and then just push as it's done for complex filter pushdown.

On the other hand, I can imagine a situation where a projection index is required to determine whether you can push some expression or not: imagine you have a very fast strlen() function but not for binary data: pushdown_expression can't decide with the expression alone whether to push or not.

@Mytherin WDYT?

Mytherin:
I think the current API is fine - the existing pushdown expression API is more meant for filter pushdown.

Could you resolve the merge conflicts? Then we can kick off CI again.

myrrc:
Rebased, thanks

Mytherin:
Thanks!

PR #22789, opened 2026-05-20
Add compressed probing hash join

ahmetkarapinar:
# Add Compressed Probing to Hash Joins

## Motivation

A regular hash join probe pays a fixed per-row cost: hash the key, look it up in the pointer table, salt-compare, re-check key bytes. On a 2048-row chunk that’s 2048 hashes, 2048 cache-missing lookups, 2048 key compares.

When the probe key arrives dictionary-compressed, this is wasted work. If we probe only the distinct dictionary values instead of every row, we eliminate redundant hashing, redundant pointer-table lookups, and redundant key-byte comparisons — reducing the per-chunk cost from N down to D, where D is the dictionary size.

## Approach

We mirror the optimization that `GroupedAggregateHashTable::TryAddDictionaryGroups` already applies on the aggregate side, this time for single-column hash joins. When the LHS join-key column arrives as a `DICTIONARY_VECTOR` with a stable `dictionary_id`, we route into a new `JoinHashTable::TryProbeDictionary` that does the work once per distinct dictionary slot, not once per probe row, and caches the results across chunks so subsequent chunks under the same id pay almost nothing.

The cache lives on the per-thread `ProbeState` as a `ProbeDictionaryState` with three fields:

- `dictionary_pointers` — one slot per dictionary entry, holding the head-of-chain pointer for hits and `nullptr` for misses.
- `found_entry` — one bool per dictionary entry, marking whether we’ve already tried to resolve that slot.
- `dictionary_id` — the id this cache is currently bound to.

On each eligible chunk, `TryProbeDictionary` does six things:

1. **Eligibility check** — confirm the dictionary size and id pass the gates.
2. **Rebind** — if the `dictionary_id` rotated, reset the cache; otherwise reuse it as-is.
3. **Collect unseen slots** — walk the chunk’s selection vector once and gather the distinct dictionary slots we haven’t resolved yet. We improved this step compared to the hash aggregate version by adding a `resolved_count` variable that tracks how many distinct slots have already been resolved under the current id; once it reaches `dict_size`, the walk is skipped entirely on warm chunks, eliminating an O(N) operation. This improvement will be ported back to the hash aggregate side in a separate PR.
4. **Probe the unseen slots** — slice those D keys out of the child vector, hash them, and call `GetRowPointers` on D rows instead of N.
5. **Populate the cache** — write the resolved head pointers into `dictionary_pointers` at their slot positions.
6. **Fan out to N rows** — for each probe row, look up its dictionary slot and copy the cached head pointer into the `ScanStructure`. Then the existing chain walker takes over, unchanged.

The first chunk under a new id pays the full D-row probe cost. Every chunk after that, for as long as the id holds, skips steps 3–5 entirely and only pays for the O(N) fan-out in step 6 — no hashing, no hash-table touch, no key comparison.

## Activation conditions

Three layers; any failure falls through to `Probe()` unchanged.

**Layer 1 — once per operator state (`CanUseCompressedProbe`):** not external, not perfect-hash, `conditions.size() == 1`, comparison is `COMPARE_EQUAL` or `COMPARE_NOT_DISTINCT_FROM`, LHS is `BOUND_REF`, LHS type is not `STRUCT`.

**Layer 2 — once per chunk (`LHSChunkIsDictionaryEligible`):** LHS vector is `DICTIONARY_VECTOR` with a valid `DictionarySize`.

**Layer 3 — inside `TryProbeDictionary` (source of truth):** `dict_size != 0`; if `dictionary_id` is non-empty then `dict_size < 20000`; if empty then `dict_size * 2 < keys.size()` (no cross-chunk reuse, only worth it when D ≪ N).
<h2>Showcase</h2>
<pre><code class="language-sql">SET threads = 1;
CREATE TABLE dim  AS SELECT format('key_{:05d}', i)       AS k, i AS v      FROM range(100)       t(i);
CREATE TABLE fact AS SELECT format('key_{:05d}', i % 100) AS k, i AS amount FROM range(50000000)  t(i);

SELECT COUNT(*) FROM fact f JOIN dim d ON f.k = d.k;
</code></pre>
<h3>Workload shape</h3>

Property | Value
-- | --
D (dim table size) | 100
N (probe rows) | 50,000,000
Global N/D | 500,000
Per-row-group D | 101 (100 + 1 sentinel slot)
Per-row-group N | 122,880
Per-row-group N/D | ~1,217
Number of vectors (chunks) | 24,414

<h3>Cache behaviour — expected vs observed</h3>

Mechanism | Expected | Observed
-- | -- | --
Activations per chunk | 24,414 (50M ÷ 2048) | 24,414 ✓
Rebinds (first bind + one per row group rotation) | 407 | 407 ✓
Walk executed (cold chunks) | 407 | 407 ✓
Walk SKIPPED (FULL CACHE HIT) | 24,007 (= 24,414 − 407) | 24,007 ✓
GetRowPointers probes D, not N | D=100, not N=2048 | D=100 every time ✓

<p><code>GetRowPointers</code> runs on <strong>D=100 keys per row group instead of N=2048 per chunk</strong> — one cold walk per row group (407 row groups total) gives ~41K HT lookups versus ~50M on <code>main</code>.</p>
<h3>main vs dict-probing (D=100, N=50M)</h3>

Metric | Main | Dict-probing | Wall-time reduction | Speedup
-- | -- | -- | -- | --
End-to-end, min | 0.286658 s | 0.119606 s | −58.28% | 2.397×
End-to-end, geomean | 0.291225 s | 0.122105 s | −58.07% | 2.385×
Probe path (PhysicalHashJoin::ExecuteInternal) | — | — | −62.8% | 2.688×

<h2>Benchmarks</h2>
<h3>TPC-H</h3>
<p><strong>No activation at SF=1, 10, 20, 50, 100.</strong></p>
<h3>TPC-DS</h3>
<p>One query activates:</p>

Query | SF1 | SF10 | SF100
-- | -- | -- | --
Q95 | yes | yes | yes
all others | no | no | no

<p><strong>SF100 Q95:</strong></p>

Metric | Main | Dict-probing | Wall-time reduction | Speedup
-- | -- | -- | -- | --
End-to-end, min | 11.428301 s | 11.256842 s | −1.50% | 1.015×
End-to-end, geomean | 11.449038 s | 11.314315 s | −1.18% | 1.012×

<h2>Constant-vector extension</h2>
<p><code>TryProbeDictionary</code> resolves D distinct keys per chunk and fans them out to N rows. A <strong>constant vector</strong> is the degenerate case D = 1 — the whole chunk is one value, produced by a literal projected into the key column or by storage’s constant-compressed numeric segments. This PR also adds <code>JoinHashTable::TryProbeConstant</code>, a smaller sibling: one <code>GetRowPointers</code> call with <code>count = 1</code>, fan the single resolved chain-head pointer out into every kept row’s slot. No cache, no rebind, no <code>dictionary_id</code> lifecycle — D = 1 collapses the per-id machinery to a single lookup per chunk. NULL handling falls out of the existing <code>RowMatcher</code> and <code>PrepareKeys</code> behaviour, so no special-casing is needed. The constant branch is dispatched before the dictionary branch in <code>ExecuteInternal</code>; the two are mutually exclusive by vector type and share the structural gate <code>CanUseCompressedProbe</code>.</p>
<hr>
<p>@lnkuiper — implementation of Compressed probing, constant vector probing is also included, ready for review.</p>
<p>Thank you for taking the time to review! I’m looking forward to your feedback.</p>

### DIFF
--- a/src/execution/operator/join/physical_hash_join.cpp
+++ b/src/execution/operator/join/physical_hash_join.cpp
@@ -1645,6 +1645,10 @@ public:
 		} else {
 			spill_chunk.Reset();
 		}
+		// discard cached build-side pointers; the HT may have been reset between iterations (e.g. recursive CTE)
+		if (probe_state.dict_state) {
+			probe_state.dict_state = make_uniq<JoinHashTable::ProbeDictionaryState>();
+		}
 		// perfect_hash_join_state will be lazily initialized on first Execute when we have a real ExecutionContext
 	}
 };
--- a/src/execution/operator/join/physical_hash_join.cpp
+++ b/src/execution/operator/join/physical_hash_join.cpp
@@ -3,6 +3,7 @@
 #include "duckdb/common/assert.hpp"
 #include "duckdb/common/projection_index.hpp"
 #include "duckdb/common/radix_partitioning.hpp"
+#include "duckdb/common/vector/dictionary_vector.hpp"
 #include "duckdb/common/types.hpp"
 #include "duckdb/common/types/value_map.hpp"
 #include "duckdb/common/uhugeint.hpp"
@@ -1575,6 +1576,44 @@ SinkFinalizeType PhysicalHashJoin::Finalize(Pipeline &pipeline, Event &event, Cl
 //===--------------------------------------------------------------------===//
 // Operator
 //===--------------------------------------------------------------------===//
+//! Structural gate for the compressed-vector probe paths; evaluated once per operator state
+static bool CanUseCompressedProbe(const HashJoinGlobalSinkState &sink, const vector<JoinCondition> &conditions) {
+	if (sink.external) {
+		// external joins re-finalize the HT mid-probe, invalidating the per-id pointer cache
+		return false;
+	}
+	if (sink.perfect_join_executor) {
+		return false;
+	}
+	if (conditions.size() != 1) {
+		return false;
+	}
+	const auto cmp = conditions[0].GetComparisonType();
+	if (cmp != ExpressionType::COMPARE_EQUAL && cmp != ExpressionType::COMPARE_NOT_DISTINCT_FROM) {
+		return false;
+	}
+	if (conditions[0].GetLHS().GetExpressionClass() != ExpressionClass::BOUND_REF) {
+		return false;
+	}
+	if (conditions[0].GetLHS().GetReturnType().InternalType() == PhysicalType::STRUCT) {
+		return false;
+	}
+	return true;
+}
+
+//! Per-chunk fast-reject so non-dictionary chunks skip the TryProbeDictionary call entirely
+static bool LHSChunkIsDictionaryEligible(const Vector &lhs_key) {
+	if (lhs_key.GetVectorType() != VectorType::DICTIONARY_VECTOR) {
+		return false;
+	}
+	return DictionaryVector::DictionarySize(lhs_key).IsValid();
+}
+
+//! Per-chunk fast-reject so non-constant chunks skip the TryProbeConstant call entirely
+static bool LHSChunkIsConstant(const Vector &lhs_key) {
+	return lhs_key.GetVectorType() == VectorType::CONSTANT_VECTOR;
+}
+
 class HashJoinOperatorState : public CachingOperatorState {
 public:
 	HashJoinOperatorState(ClientContext &context, const PhysicalHashJoin &op_p, HashJoinGlobalSinkState &sink)
@@ -1594,6 +1633,8 @@ public:
 	JoinHashTable::ProbeState probe_state;
 	//! Chunk to sink data into for external join
 	DataChunk spill_chunk;
+	//! Cached result of CanUseCompressedProbe for this operator state
+	bool compressed_probe_enabled = false;
 
 public:
 	void Finalize(const PhysicalOperator &op, ExecutionContext &context) override {
@@ -1647,6 +1688,8 @@ unique_ptr<OperatorState> PhysicalHashJoin::GetOperatorState(ExecutionContext &c
 		sink.InitializeProbeSpill();
 	}
 
+	state->compressed_probe_enabled = CanUseCompressedProbe(sink, conditions);
+
 	return std::move(state);
 }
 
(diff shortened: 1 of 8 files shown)


src/execution/operator/join/physical_hash_join.cpp
```diff
@@ -1594,6 +1633,8 @@ class HashJoinOperatorState : public CachingOperatorState {
 	JoinHashTable::ProbeState probe_state;
 	//! Chunk to sink data into for external join
 	DataChunk spill_chunk;
+	//! Cached result of CanUseCompressedProbe for this operator state
+	bool compressed_probe_enabled = false;
```
lnkuiper: Instead of having this boolean here, could you instead make the `ProbeDictionaryState` in `ProbeState` a `unique_ptr` which is only initialized when `CanUseCompressedProbe` is true? Then we can move all of the dictionary probing logic like `LHSChunkIsDictionaryEligible` and `LHSChunkIsConstant` into `join_hashtable.cpp`, and just call `Probe` in `PhysicalHashJoin::ExecuteInternal` like before. Then `Probe` can handle all the compressed probing logic.

lnkuiper (changes requested):
Thanks for the PR! I'd like to move some code around:

src/execution/operator/join/physical_hash_join.cpp
ahmetkarapinar: Done!

src/execution/operator/join/physical_hash_join.cpp
```diff
@@ -1647,6 +1674,11 @@ unique_ptr<OperatorState> PhysicalHashJoin::GetOperatorState(ExecutionContext &c
 		sink.InitializeProbeSpill();
 	}
 
+	if (CanUseCompressedProbe(sink, conditions)) {
+		// non-null dict_state is the enabling signal for the compressed-vector probe paths in Probe()
+		state->probe_state.dict_state = make_uniq<JoinHashTable::ProbeDictionaryState>();
```
lnkuiper, line 1683: I think we also need to "refresh" this state, i.e., do another `make_uniq` of this thing in `HashJoinOperatorState::Reset()`, otherwise we will keep using the old vectors after the pipeline has been reset (in recursive CTEs).

lnkuiper (commented):
Thanks for the changes! I think I have just one more comment then this is ready to go:

src/execution/operator/join/physical_hash_join.cpp
ahmetkarapinar: Done! Thanks for the catch 🙂

lnkuiper (approved):
Thanks!

PR #22793, opened 2026-05-20
[Dev] Create a `patched` branch in the cloned repo if `APPLY_PATCHES` is set

Tishj:
Every patch becomes a commit to the `patched` branch, making it easier to write a new patch, since the working tree will be clean.
It's also easier to develop a patch, all you need is a `git reset HEAD^` to turn the last applied patch back into uncommitted changes in the working tree.

This also makes the script much simpler, as previously we jumped through some weird hoops to determine if the repository had uncommitted local changes not created by patches.

Now that we always leave the working tree clean after applying the patches, this check is much simpler.

### DIFF
--- a/scripts/apply_extension_patches.py
+++ b/scripts/apply_extension_patches.py
@@ -9,6 +9,8 @@ import tempfile
 directory = sys.argv[1]
 patch_pattern = f"{directory}*.patch"
 
+PATCHED_BRANCH = "patched"
+
 if os.environ.get("DUCKDB_SKIP_APPLYING_PATCHES") == "1":
     exit(0)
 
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks for the PR!

I'm fine with merging this but note that I've done this + more in https://github.com/duckdb/duckdb/pull/22063 to simplify local development w.r.t. patches - you can use `export DUCKDB_NEW_EXTENSION_BUILD=1` to use the new build system. Ideally this would be the only build system but the CI still runs on the old one so I haven't changed the default (CC @smvv).

Tishj:
Ah 😅 
That looks like it's exactly what I'm after
Funny that we separately made very similar changes to the patch-applying mechanism

I'll just add `export DUCKDB_NEW_EXTENSION_BUILD=1` to my zshrc 👍

PR #9307, opened 2023-10-11
[Dev] Fix Wasm.yml, hardcoding extension config to latest duckdb-wasm

carlopi:

PR #22794, opened 2026-05-20
Internal #9307: Window Batching

hawkfish:
* Refactor CanUsePartitionedAggregate to expose HasSingleValuePartitions;
* Add partition batch detection to LogicalWindow planning;
* Add batch support to PhysicalWindow;
* Use OperatorPartitionInfo to avoid naming confusion;
* Guard strategy_sinks access with a mutex;
* Handle mask construction with unused partition expressions;
* Add validation tests;
* Add benchmark (100M rows, 2 columns, 77 partitions);
* Performance improvement over baseline 1.90s from 2.24s (iMac Pro);


lnkuiper:
I think there's an issue where if we partition by `p1, p2` but the Window only specifies partitioning by `p1`, we still compute the window over `p1, p2`:
```sql
# name: test/temp.test
# description: Review repros for PR #22794 window batching correctness
# group: [temp]

require parquet

statement ok
CREATE TABLE partitioned_tbl AS
SELECT
	i % 2 AS p1,
	i % 3 AS p2,
	i AS x
FROM range(60) tbl(i)

statement ok
COPY partitioned_tbl TO '{TEST_DIR}/window_batch_subset_partition'
	(FORMAT parquet, PARTITION_BY (p1, p2))

statement ok
DROP TABLE partitioned_tbl

statement ok
CREATE VIEW partitioned_tbl AS FROM '{TEST_DIR}/window_batch_subset_partition/**/*.parquet'

# The source files have single values for p1 and p2, but a PARTITION BY p1
# window partition spans multiple files. Batching by source partition would split
# the logical window partition and reset lag at every p2 file.
query I
SELECT count(*)
FROM (
	SELECT p1, p2, x, lag(x) OVER (PARTITION BY p1 ORDER BY x) AS prev
	FROM partitioned_tbl
	QUALIFY prev IS NOT NULL AND x - prev <> 2
)
----
0

# Same correctness issue through a running aggregate window: the frame must see
# all previous rows in the full p1 partition, not just the current p1/p2 file.
query I
SELECT count(*)
FROM (
	SELECT
		p1,
		x,
		sum(x) OVER (
			PARTITION BY p1
			ORDER BY x
		) AS running
	FROM partitioned_tbl
	WHERE p1 = 0
	QUALIFY running <> ((x // 2) * ((x // 2) + 1))
)
----
0
```

lnkuiper:
There still seems to be a correctness error around different partition specifications:
```sql
require parquet

statement ok
CREATE TABLE partitioned_window_repro AS
SELECT
	i % 2 AS p1,
	i % 3 AS p2,
	i AS x
FROM range(60) tbl(i)

statement ok
COPY partitioned_window_repro TO '{TEST_DIR}/window_batch_stale_columns'
	(FORMAT parquet, PARTITION_BY (p1))

statement ok
DROP TABLE partitioned_window_repro

statement ok
CREATE VIEW partitioned_window_repro AS
FROM '{TEST_DIR}/window_batch_stale_columns/**/*.parquet'

# The first window is not batchable from a source partitioned only by p1, but
# HasSingleValuePartitions leaves its p2 column in partitioned_columns[0]. The
# later p1 window is batchable and is planned first, but physical construction
# reads partitioned_columns by physical window index instead of expression index,
# so it asks the source for p2 partition data and aborts.
query I
SELECT count(*)
FROM (
	SELECT
		lag(x) OVER (PARTITION BY p2 ORDER BY x) AS prev_p2,
		lag(x) OVER (PARTITION BY p1 ORDER BY x) AS prev_p1,
		x
	FROM partitioned_window_repro
) t
WHERE prev_p1 IS NOT NULL AND x - prev_p1 <> 2
----
0
```

hawkfish:
> There still seems to be a correctness error around different partition specifications:

Doh - all integers look the same...

lnkuiper (approved):
Thanks! LGTM now :)

PR #22795, opened 2026-05-20
Fixup ctrl + L, should clean screen and jump to top

carlopi:
I have an hard time reproducing this either way, but on v1.5.2 I have saw instances of ctrl + L cleaning a single row. Claude pointed to this fix, that makes sort of sense (single write() instead of two in particular), but very hard to tell them apart

### DIFF
--- a/tools/shell/linenoise/linenoise.cpp
+++ b/tools/shell/linenoise/linenoise.cpp
@@ -1759,7 +1759,8 @@ int Linenoise::Edit() {
 			EditMoveEndOfLine();
 			break;
 		case CTRL_L: /* ctrl+l, clear screen */
-			linenoiseClearScreen();
+			/* the screen clear is emitted by RefreshLine() in the same write
+			 * as the redraw, so the prompt is guaranteed to land at row 1 */
 			clear_screen = true;
 			RefreshLine();
 			break;
--- a/tools/shell/linenoise/rendering.cpp
+++ b/tools/shell/linenoise/rendering.cpp
@@ -130,7 +130,7 @@ static void renderText(size_t &render_pos, char *&buf, size_t &len, size_t pos, 
  *
  * Rewrite the currently edited line accordingly to the buffer content,
  * cursor position, and number of columns of the terminal. */
-void Linenoise::RefreshSingleLine() const {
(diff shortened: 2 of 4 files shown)


tools/shell/linenoise/linenoise.cpp
```diff
@@ -1759,7 +1759,8 @@ int Linenoise::Edit() {
 			EditMoveEndOfLine();
 			break;
 		case CTRL_L: /* ctrl+l, clear screen */
-			linenoiseClearScreen();
```
Mytherin, line 1762: Why does this function still exist if it is removed here? Don't we now have two separate paths of clearing the screen that are going to be subtly different in behavior?

tools/shell/linenoise/rendering.cpp
```diff
@@ -142,6 +142,10 @@ void Linenoise::RefreshSingleLine() const {
 	renderText(render_pos, render_buf, render_len, pos, ws.ws_col, plen, highlight_buffer, Highlighting::IsEnabled());
 
 	AppendBuffer append_buffer;
+	if (clear_screen) {
```
Mytherin, line 145: `RefreshSingleLine` is legacy and is basically never used - no need to change it

tools/shell/linenoise/rendering.cpp
```diff
@@ -908,6 +913,11 @@ void Linenoise::RefreshMultiLine() {
 	/* First step: clear all the lines used before. To do so start by
 	 * going to the last row. */
 	AppendBuffer append_buffer;
+	if (do_clear_screen) {
+		/* erase screen + home cursor as part of the same write as the
+		 * redraw, so the prompt is guaranteed to land at row 1 */
+		append_buffer.Append("\x1b[2J\x1b[H");
```
Mytherin, line 919: This looks like we're repeating the clear - at that point I would suggest just removing `Terminal::ClearScreen` entirely.

Mytherin (commented):
Thanks - left some comments. There's two fixes here (changing `2J` to `H` in the escape code, and doing a clear followed by a single write with a clear). Can you verify which is the correct fix and then fix that fix so that it is the only clear path used.

PR #22796, opened 2026-05-20
append to new row group if vacuum_rebuild_indexes is enabled

artjomPlaunov:
In RowGroupCollection::InitializeAppend, if there were no indexes present we would honor SUGGEST_NEW, i.e. try to append to a new row group, or if there are indexes present and space in the last row group, we would append to an existing row group (to mitigate the current limitation of no vacuuming with indexes). 

Modify this so that if the vacuum_rebuild_indexes setting is enabled, we append to a new row group (since vacuuming will happen assuming the row count is below the threshold). 

Temporary workaround while I am working on vacuuming in the presence of indexes by default, at which point we can remove the vacuum_rebuild_indexes and related logic around it.

### DIFF
--- a/src/storage/table/row_group_collection.cpp
+++ b/src/storage/table/row_group_collection.cpp
@@ -26,6 +26,19 @@
 
 namespace duckdb {
 
+static bool CanRebuildExistingIndexesAfterVacuum(DataTableInfo &info, DatabaseInstance &db, idx_t total_rows) {
+	auto &indexes = info.GetIndexes();
+	if (indexes.Empty() || indexes.HasUnbound()) {
+		return false;
+	}
+	auto vacuum_rebuild_threshold = Settings::Get<VacuumRebuildIndexesSetting>(db);
+	if (vacuum_rebuild_threshold == 0 || total_rows > vacuum_rebuild_threshold) {
+		return false;
+	}
+	auto index_types = indexes.DistinctIndexTypes();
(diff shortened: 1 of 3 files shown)


src/storage/table/row_group_collection.cpp
```diff
@@ -24,11 +22,23 @@
 #include "duckdb/storage/table/scan_state.hpp"
 #include "duckdb/storage/table_storage_info.hpp"
 #include "duckdb/transaction/duck_transaction.hpp"
-#include "duckdb/common/serialization_compatibility.hpp"
 #include "duckdb/common/type_visitor.hpp"
 
 namespace duckdb {
 
+static bool CanRebuildExistingIndexesAfterVacuum(DataTableInfo &info, DatabaseInstance &db, idx_t total_rows) {
```
evertlammerts, line 29: much nicer than the oneliner that it was 👍

artjomPlaunov: @evertlammerts agreed :-) I will wait for your ATTACH PR to go in then edit this one accordingly

evertlammerts (approved):
LGTM!

evertlammerts:
Maybe you want to retarget v1.5...?

artjomPlaunov:
@evertlammerts ah yes, planning to do that soon. Once your ATTACH PR is in, I will retarget this to 1.5 as well as make the one line edit (I think) to make sure it works along with that PR.

PR #22797, opened 2026-05-20
Add Disjunctive Join Rewriter Optimizer

EslamAhmed171:
### Problem

DuckDB currently falls back to `BLOCKWISE_NL_JOIN` for disjunctive join conditions, joins where the `ON` clause is an `OR` of equality predicates.

For example, a query like:

```sql
SELECT *
FROM left_table l
LEFT JOIN right_table r
  ON l.a = r.x OR l.b = r.y;
```

previously could not use a hash join and would execute as a nested loop join.

### Solution

This PR introduces a new optimizer pass, `DisjunctiveJoinRewriter`, that rewrites disjunctive `ANY JOIN`s into a union of `LogicalComparisonJoin`s (hash joins). The rewrite is triggered when the join condition is a disjunction (`OR`) of equality comparisons.

#### Supported join types

`INNER`, `LEFT`, `RIGHT`, `FULL OUTER`, `SEMI`, `ANTI`.

#### Query transformation

The rewrite works as follows:

1. **Materialize left and right inputs as CTEs**  
   Each input is wrapped in a CTE and injected with a synthetic `row_number()` column (aliased as `left_rowid` / `right_rowid`) so we can identify matching rows later.

   ```sql
   WITH left_cte AS (
       SELECT *, row_number() OVER () AS left_rowid
       FROM left_table
   ),
   right_cte AS (
       SELECT *, row_number() OVER () AS right_rowid
       FROM right_table
   )
   ```

2. **Build a `match_cte` via `UNION` of per-branch joins**  
   Each disjunctive branch becomes its own inner hash join between the two CTEs, filtered to exclude nulls.  
   The result of every branch is projected to `(left_rowid, right_rowid)` and unioned together:

   ```sql
   match_cte AS (
       SELECT l.left_rowid, r.right_rowid
       FROM left_cte l
       INNER JOIN right_cte r ON l.a = r.x
       WHERE l.a IS NOT NULL AND r.x IS NOT NULL

       UNION

       SELECT l.left_rowid, r.right_rowid
       FROM left_cte l
       INNER JOIN right_cte r ON l.b = r.y
       WHERE l.b IS NOT NULL AND r.y IS NOT NULL
   )
   ```

3. **Reconstruct the original join semantics using the match set**  
   The match CTE is then joined back against the original CTEs to produce the correct output:

   - **INNER**
     ```sql
     SELECT l.*, r.*
     FROM left_cte l
     INNER JOIN match_cte m ON l.left_rowid = m.left_rowid
     INNER JOIN right_cte r ON m.right_rowid = r.right_rowid
     ```

   - **LEFT**
     ```sql
     SELECT l.*, r.*
     FROM left_cte l
     LEFT JOIN match_cte m ON l.left_rowid = m.left_rowid
     LEFT JOIN right_cte r ON m.right_rowid = r.right_rowid
     ```

   - **RIGHT**
     ```sql
     SELECT l.*, r.*
     FROM right_cte r
     LEFT JOIN match_cte m ON r.right_rowid = m.right_rowid
     LEFT JOIN left_cte l ON m.left_rowid = l.left_rowid
     ```

   - **FULL OUTER**
     ```sql
     SELECT l.*, r.*
     FROM left_cte l
     LEFT JOIN match_cte m ON l.left_rowid = m.left_rowid
     FULL OUTER JOIN right_cte r ON m.right_rowid = r.right_rowid
     ```

   - **SEMI**
     ```sql
     SELECT l.*
     FROM left_cte l
     SEMI JOIN match_cte m ON l.left_rowid = m.left_rowid
     ```

   - **ANTI**
     ```sql
     SELECT l.*
     FROM left_cte l
     ANTI JOIN match_cte m ON l.left_rowid = m.left_rowid
     ```

cc: @lnkuiper


kryonix:
I do think the idea is interesting—and for your small examples it works. However, I think this is pretty dangerous territory we are entering here, because using (materialized) CTEs like this will cause issues. Take this query from your tests, for example:

```sql
SELECT o.o_orderkey, o.o_custkey, l.l_orderkey, l.l_partkey
FROM orders o JOIN lineitem l ON o.o_orderkey = l.l_orderkey OR o.o_custkey = l.l_partkey;
```

Your optimization asks the system to fully materialize `lineitem`, `orders`, and the `match_cte` in memory. I tried this with TPC-H sf 100 and memory consumption blew up at lot. This eventually led to spilling and the query failed. Without this optimizer, memory consumption is much smaller and the query can be computed successfully.

lnkuiper:
Really cool, I'll try to review this next tuesday when I have the time.

I think Denis' concern is valid - we might want to guard against excessive materialization, as well as do some benchmarks to see when this improves performance and when it doesn't. That can help in deciding when this optimisation should be activated. Nested loop join complexity is terrible, of course, but if the build side is tiny it might not be so bad. Similarly, if the probe side is massive, materializing it as a CTE might not be the right call. If we have some heuristics in place the optimisation will become safer.

src/optimizer/disjunctive_join_rewriter.cpp
```diff
@@ -0,0 +1,533 @@
...
+		D_ASSERT(false);
+		return op;
+	}
+
+	epilogue = NormaliseOutput(std::move(epilogue), orig_bindings, orig_types, left_cte, right_cte, join.join_type);
+
+	// wrap in CTEs
+	epilogue = make_uniq<LogicalMaterializedCTE>("match_cte", match_cte.table_index, match_cte.output_types.size(),
+	                                             std::move(match_result.plan), std::move(epilogue),
+	                                             CTEMaterialize::CTE_MATERIALIZE_NEVER);
```
kryonix: `CTEMaterialize::CTE_MATERIALIZE_NEVER` has no effect here, because this optimizer pass is placed after CTE inlining is already done.

EslamAhmed171: Nice catch!

EslamAhmed171:
> I think Denis' concern is valid - we might want to guard against excessive materialization, as well as do some benchmarks to see when this improves performance and when it doesn't. That can help in deciding when this optimisation should be activated. Nested loop join complexity is terrible, of course, but if the build side is tiny it might not be so bad. Similarly, if the probe side is massive, materializing it as a CTE might not be the right call. If we have some heuristics in place the optimisation will become safer.

Agreed we need some heuristics to make the optimization safe, I'll add heuristics to check build/probe cardinality before applying the rewrite, and run TPC-H benchmarks (sf1–sf100) this weekend to find the safe crossover points. I'll share the results here once I have them.

EslamAhmed171:
@lnkuiper Sorry for the wait

I've updated the PR with memory-aware heuristics to guard against OOM:
1. Skip tiny tables
2. Compute materialization costs for:
Left CTE (if no physical rowid available)
Right CTE (if no physical rowid available)
Match CTE (UNION result with matched row pairs)
If total estimated cost exceeds 50% of configured memory limit, the optimization is skipped and the original plan is used. This is a conservative threshold, in most cases queries will run fully in-memory without spilling. However,in some workloads I tested, spilling to disk can still be faster than falling back to Blockwise NL join. We're keeping the threshold strict for safety.
3. When the underlying table scan exposes a physical ROWID column , we reuse it directly instead of injecting a ROW_NUMBER() window function.
4. The optimizer currently only triggers when both join sides are simple table scans. This ensures cardinality estimates from `RelationStatisticsHelper` are reliable.

lnkuiper:
Great, thanks! I'll review when I find the time :)

kryonix:
Maybe with the PipelineExchange (i.e. streaming CTEs), my initial concern is not valid anymore? I'm not entirely sure. I can't remember the plan shape produced here. But this may be worth checking out.

lnkuiper:
I think we should be able to avoid materialisation with the CTE fan-out, but I think the CTE fan-out has to be improved for it to work. The CTE fan-out should:
- Start the build-side fan-out first, feeding every hash-join build pipeline.
- Wait until those builds are finalized.
- Start the probe producer and stream each probe chunk to all match and anti/mark pipelines.
- Apply backpressure based on the slowest consumer so probe storage stays bounded.
- Fall back to materialization if consumers have cyclic dependencies or cannot run concurrently.

Probably best if @kryonix looks into this.

Then this optimiser can rewrite this without materializing the probe side or building the full `(lhs_id, rhs_id)` match set. The important bit is that `MARK` joins are not used to deduplicate match pairs: later match branches recheck the earlier predicates on the specific `(lhs, rhs)` pair.

| Join type | Rewrite |
|---|---|
| `INNER` | Disjoint hash-join branches combined with `UNION ALL` |
| `LEFT` | Inner matches plus an LHS chain of anti joins |
| `RIGHT` | Same as `LEFT`, with the sides swapped |
| `FULL` | Inner matches, probe-side anti chain, and unmatched build rows using a bounded set of matched build IDs |
| `SEMI` | Chained MARK joins, filtering once at the end if any marker is true |
| `ANTI` | Chained anti joins |

For example:

```sql
CREATE TABLE lhs (id INTEGER, a INTEGER, b INTEGER);
CREATE TABLE rhs (id INTEGER, a INTEGER, b INTEGER);

INSERT INTO lhs VALUES (1, 1, 2), (2, 7, 7);
INSERT INTO rhs VALUES (10, 1, 9), (11, 8, 2), (12, 1, 2);
```

### Inner join

```sql
-- Original
SELECT lhs.id, rhs.id
FROM lhs JOIN rhs ON lhs.a = rhs.a OR lhs.b = rhs.b;

-- Rewritten
WITH
lhs_fanout AS MATERIALIZED (FROM lhs),
rhs_fanout AS MATERIALIZED (FROM rhs)
SELECT lhs.id, rhs.id
FROM lhs_fanout lhs JOIN rhs_fanout rhs ON lhs.a = rhs.a
UNION ALL
SELECT lhs.id, rhs.id
FROM lhs_fanout lhs JOIN rhs_fanout rhs ON lhs.b = rhs.b
WHERE (lhs.a = rhs.a) IS NOT TRUE;
```

Both queries return `(1, 10)`, `(1, 11)`, and `(1, 12)`. The pair `(1, 12)` matches both predicates, but the second branch excludes only that specific duplicate pair.

### Left join

```sql
-- Original
SELECT lhs.id, rhs.id
FROM lhs LEFT JOIN rhs ON lhs.a = rhs.a OR lhs.b = rhs.b;

-- Rewritten
WITH
lhs_fanout AS MATERIALIZED (FROM lhs),
rhs_fanout AS MATERIALIZED (FROM rhs)
SELECT lhs.id, rhs.id
FROM lhs_fanout lhs JOIN rhs_fanout rhs ON lhs.a = rhs.a
UNION ALL
SELECT lhs.id, rhs.id
FROM lhs_fanout lhs JOIN rhs_fanout rhs ON lhs.b = rhs.b
WHERE (lhs.a = rhs.a) IS NOT TRUE
UNION ALL
SELECT lhs.id, NULL::INTEGER
FROM lhs_fanout lhs
ANTI JOIN rhs_fanout rhs_a ON lhs.a = rhs_a.a
ANTI JOIN rhs_fanout rhs_b ON lhs.b = rhs_b.b;
```

Both queries return `(1, 10)`, `(1, 11)`, `(1, 12)`, and `(2, NULL)`. With streaming CTE fan-out, the LHS can feed the match and anti-join pipelines without being fully materialized.

EDIT: I think this would be much preferred over the current shape, as it can be quite risky.

EslamAhmed171:
> Then this optimiser can rewrite this without materializing the probe side or building the full `(lhs_id, rhs_id)` match set. The important bit is that `MARK` joins are not used to deduplicate match pairs: later match branches recheck the earlier predicates on the specific `(lhs, rhs)` pair.
> 
> Join type	Rewrite
> `INNER`	Disjoint hash-join branches combined with `UNION ALL`
> `LEFT`	Inner matches plus an LHS chain of anti joins
> `RIGHT`	Same as `LEFT`, with the sides swapped
> `FULL`	Inner matches, probe-side anti chain, and unmatched build rows using a bounded set of matched build IDs
> `SEMI`	Chained MARK joins, filtering once at the end if any marker is true
> `ANTI`	Chained anti joins

That's an elegant solution. I have implemented each join type as mentioned, but for `FULL JOIN`, I did not get the idea of matched build IDs, I used a reverse chained anti join to get the unmatched RHS rows.

For CTE-fanout support for `PhysicalHashJoin`, I have tried over the past two days to add support for it but failed. I enabled external support for the hash join using:

```cpp
PipelineExternalInputSupport GetExternalInputSupport() const override {
    return PipelineExternalInputSupport::SUPPORTED;
}
```
And I hinted for the CTE that if it could not be `DIRECT`, it should be `BUFFERED`.

The results are correct for `INNER`, `ANTI`, and `SEMI`. However, for `LEFT`, `RIGHT`, and `FULL`, there are some extra rows produced by the chained anti join in each of those three.

I suspect that making both `rhs_cte` and `lhs_cte` act as both build and probe sides in the query is causing these wrong results. When I disabled the `BuildProbeSide` optimizer, I got the correct results.

I am not sure where the wrong results are coming from. It would be nice if @kryonix could help check if there is something missing in `PipelineBroadcastExchange` that should be implemented to support `PhysicalHashJoin` properly, or if there is a race condition when a CTE has multiple `DIRECT` consumers alongside a `BUFFERED` or `MATERIALIZED` consumer.

kryonix:
I looked into this. This is not primarily a race in the exchange's buffering logic. What is missing is dependency-aware external input for hash joins.

Simply returning `PipelineExternalInputSupport::SUPPORTED` from `PhysicalHashJoin` is unsafe. A normal hash-join probe pipeline starts only after its build pipeline has finalized. A directly fed pipeline is instead driven from the CTE producer, and currently that path does not transfer the build-completion dependency to the producer. I reproduced this with the earlier experimental commit: probe state could be constructed against a null or unfinalized hash table. In release builds, probing an incomplete table also explains the additional rows from the anti joins.

`BuildProbeSide` exposes the issue because it can independently swap the generated joins. A CTE may consequently feed build pipelines in some branches and probe pipelines in others, creating a dependency cycle. Disabling `BuildProbeSide` makes the orientations consistent, but that would be a workaround rather than a valid solution. Forcing the other consumers to `BUFFERED` is also insufficient because buffering is not a hash-build finalization barrier.

The exchange needs a general readiness-aware extension: build-side fanout must complete and finalize before probe-side fanout starts, and direct fanout must fall back to materialization when the resulting pipeline dependencies are cyclic. This should be derived from the final physical pipeline graph, without hints from this optimizer.

For now, I would keep the hash-join consumers materialized and not enable `GetExternalInputSupport()` for `PhysicalHashJoin`. I will look into handling the required exchange extension separately. Once that lands, uniformly oriented branches can stream fully. Mixed orientations will safely materialize where necessary. The reverse anti-join strategy for `FULL JOIN` inherently uses both CTEs as build and probe inputs, so fully streaming that case would additionally require the matched-build-ID approach, otherwise one side must remain materialized.

EslamAhmed171:
@kryonix Thanks for your help!

Now I have a clear picture of the pipeline dependencies.
I have a question: would using the `ROW_NUMBER()` function to generate synthetic IDs for the RHS be safe for streaming? For now, I am using the RHS columns as an ID.

EslamAhmed171:
@lnkuiper Is there anything else I should change in the rewriter before the review?

kryonix:
While testing this PR stacked on top of the pipeline external-input changes, I found a `FULL JOIN` correctness issue that is independent of the exchange implementation.

**Matched `FULL JOIN` rows containing `NULL` can be emitted again as unmatched**

```sql
SELECT *
FROM (VALUES (NULL::INTEGER, 40)) l(a, b)
FULL JOIN (VALUES (NULL::INTEGER, 40)) r(a, b)
  ON l.a = r.a OR l.b = r.b
ORDER BY ALL NULLS FIRST;
```

Expected:

```text
NULL  40  NULL  40
```

Result with `disjunctive_join_rewriter`:

```text
NULL  NULL  NULL  40
NULL  40    NULL  40
```

Disabling `disjunctive_join_rewriter` returns the expected single row. I also verified that buffered/materialized and direct exchange execution produce the same incorrect result, so this originates in the rewrite rather than the pipeline exchange.

`CreateUnmatchedBuildSide` projects the matched right rows and then anti-joins the original right side against them using ordinary `COMPARE_EQUAL` for every column. If any right-side column is `NULL`, the row cannot compare equal to its matched projection and is incorrectly emitted again as unmatched.

The rewrite needs NULL-safe row identity, such as `NOT DISTINCT FROM` comparisons or a stable row identifier. A regression test should also cover NULLs in unrelated columns and duplicate right-side rows.

EslamAhmed171:
> The rewrite needs NULL-safe row identity, such as `NOT DISTINCT FROM` comparisons or a stable row identifier. A regression test should also cover NULLs in unrelated columns and duplicate right-side rows.

Nice catch! I've fixed it and updated the test cases to cover duplicate right-side rows and NULLs in unrelated columns. I also expanded the rewrite to support `IS NOT DISTINCT FROM` and added test cases for both standalone `IS NOT DISTINCT FROM` and mixed `EQUAL`/`IS NOT DISTINCT FROM` conditions.

lnkuiper:
Looking a lot better now. Some more comments:

1. This succeeds with the optimizer disabled but throws when enabled:
```sql
  CREATE TABLE l(a INTEGER, s VARCHAR);
  INSERT INTO l VALUES (1, 'bad');
  CREATE TABLE r(a INTEGER, b INTEGER);
  INSERT INTO r VALUES (1, 0);

  SELECT *
  FROM l JOIN r
    ON l.a = r.a OR CAST(l.s AS INTEGER) = r.b;
```
We should guard for throwing/volatile expressions.

2. The RHS needs a stable row identifier produced before fan-out and carried through the matched-build path.
```sql
  CREATE TABLE l(k BOOLEAN, z BOOLEAN);
  INSERT INTO l VALUES (true, false);

  CREATE TABLE r(p DOUBLE, q BOOLEAN);
  INSERT INTO r VALUES
      (0.0, true),
      (-1.0::DOUBLE / 'Infinity'::DOUBLE, true);

  SELECT l.k, l.z, signbit(r.p), r.q
  FROM l FULL JOIN r
    ON l.k = signbit(r.p) OR l.z = r.q
  ORDER BY ALL NULLS FIRST;
```
  +0.0 and -0.0 compare `IS NOT DISTINCT FROM`, but signbit distinguishes them. Only -0.0 matches. Expected:
```
  NULL   NULL    false  true
  true   false   true   true
```
With the rewrite, the unmatched +0.0 row disappears because it is considered identical to the matched -0.0 row.

3. There is duplicated code between `CreateMatchedBranches` and `CreateUnmatchedBuildSide` which could be moved to a shared helper.

4. Not sure how important but this benchmark with a `LIMIT` got a lot slower
```sql
name DisjunctiveJoinLimit
group review
subgroup join

argument n 1000000

run
SELECT l.i, r.j
FROM range(${n}) l(i)
JOIN range(${n}) r(j)
  ON l.i = r.j OR l.i + ${n} = r.j
LIMIT 1;
```

lnkuiper:
Just wanted to add that only point 2 and 3 are blocking for me, the other points are minor and can be fixed later. Really exciting optimisation!

EslamAhmed171:
> Just wanted to add that only point 2 and 3 are blocking for me, the other points are minor and can be fixed later. Really exciting optimisation!

Sorry for the wait, it was out of my hands! I have addressed points 1, 2, and 3.

Regarding point 4: The first join branch (l.i = r.j) quickly satisfies the LIMIT, but the second branch (l.i + ${n} = r.j) computes all 1M rows and does not find a match. There is no backward cancellation for child pipelines when the limit is satisfied, it does not signal the sibling pipelines that they can safely abort.

I am not entirely sure if this cancellation mechanism is completely missing or if it just isn't triggered in this specific scenario.

lnkuiper:
Really exciting stuff, thanks for working on this! We're about to create the v2.0 branch, and I think this should go towards `main` so we can do some hardening, and it'll go into the v2.1 release. I'll be sure to review this soon!

PR #22800, opened 2026-05-21
Fix `WindowSelfJoinOptimizer` ignore exception

qsliu2017:
Revert the narrowed `catch (NotImplementedException &)` introduced in https://github.com/duckdb/duckdb/pull/22164. This causes errors leak out of the optimizer instead of "skiping this optimization".

Specifically, pg_duckdb uses a `PostgresScanTableFunction` without registered in the system catalog. `DeepCopy` this operator throws a `CatalogException` instead.

### DIFF
--- a/src/optimizer/window_self_join.cpp
+++ b/src/optimizer/window_self_join.cpp
@@ -128,7 +128,7 @@ unique_ptr<LogicalOperator> WindowSelfJoinOptimizer::OptimizeInternal(unique_ptr
 		LogicalOperatorDeepCopy deep_copy(optimizer.binder, nullptr);
 		try {
 			copy_child = deep_copy.DeepCopy(window.children[0]);
-		} catch (NotImplementedException &ex) {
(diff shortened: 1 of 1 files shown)


qsliu2017:
[windows / Windows (ARM64)](https://github.com/duckdb/duckdb/actions/runs/26213237580/job/77131382270?pr=22800#logs) fails at `Run ./.github/actions/ccache-action`, but I don't really see what's wrong.

Mytherin:
Thanks - should this maybe be rebased to `v1.5-variegata` so that this will land in v1.5.4?

qsliu2017:
> Thanks - should this maybe be rebased to `v1.5-variegata` so that this will land in v1.5.4?

Rebased. But it says "You're not authorized to push to this branch".

Mytherin:
That probably means you tried to push to `duckdb/duckdb` by accident - I'll re-run CI.

PR #22801, opened 2026-05-21
Remove redundant part lock guard for sort_key_payload_state

smvv:
Fix this threadsan warning:
```
WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock) (pid=77147)
  Cycle in lock order graph: M0 (0x7f693220b108) => M1 (0x7f693220d108) => M0
```
instead of suppressing the warning.

See threadsan full error in CI log: https://github.com/duckdb/duckdb/actions/runs/25797353015/job/75778737814?pr=22498#step:7:876

### DIFF
--- a/src/common/types/row/tuple_data_allocator.cpp
+++ b/src/common/types/row/tuple_data_allocator.cpp
@@ -423,7 +423,8 @@ void TupleDataAllocator::InitializeChunkStateInternal(TupleDataPinState &pin_sta
 
 		if (sort_key_payload_state) {
 			D_ASSERT(!layout.IsSortKeyLayout()); // This must be the payload collection
-			lock_guard<mutex> guard(part.lock);
+			// SortKeySetPayload() guards sort-key payload mutations with the sort-key chunk lock.
+			// Avoid nesting part.lock with that lock, which can introduce lock-order inversions.
 			SortKeySetPayload(row_locations, offset, next, *sort_key_payload_state);
 		}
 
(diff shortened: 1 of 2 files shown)


src/common/types/row/tuple_data_allocator.cpp
```diff
@@ -423,7 +423,8 @@ void TupleDataAllocator::InitializeChunkStateInternal(TupleDataPinState &pin_sta
 
 		if (sort_key_payload_state) {
 			D_ASSERT(!layout.IsSortKeyLayout()); // This must be the payload collection
-			lock_guard<mutex> guard(part.lock);
+			// SortKeySetPayload() guards sort-key payload mutations with the sort-key chunk lock.
+			// Avoid nesting part.lock with that lock, which can introduce lock-order inversions.
 			SortKeySetPayload(row_locations, offset, next, *sort_key_payload_state);
```
smvv, line 428: Locking `part.lock` here caused a threadsan warning:
```
WARNING: ThreadSanitizer: lock-order-inversion (potential deadlock) (pid=77147)
  Cycle in lock order graph: M0 (0x7f693220b108) => M1 (0x7f693220d108) => M0
```

Locking `part.lock` isn’t needed there because that code path does **not** mutate `TupleDataChunkPart` state.

In that block we only:
- prepare `row_locations` (scan-local state), and
- call `SortKeySetPayload(...)`.

`SortKeySetPayload(...)` already takes the correct lock (`sort_key_chunk_state.chunk_lock`) around payload-pointer updates.

`part.lock` is still needed in the heap-recompute path where `part` fields (like `base_heap_ptr`) can change.
So keeping it around `SortKeySetPayload(...)` was unnecessary and created the TSan lock-order cycle.

See threadsan full error in CI log: https://github.com/duckdb/duckdb/actions/runs/25797353015/job/75778737814?pr=22498#step:7:876

lnkuiper:
Thanks, I think this makes sense - it already grabs the correct lock further down the call stack.

PR #22802, opened 2026-05-21
Add support for large integer values for CSV

### LINKED ISSUE #22750: csv with long ID columns get inferred as DOUBLE instead of HUGEINT / VARCHAR causing "correctness problems"
### What happens?

duckdb reads csv columns with huge integers as DOUBLE instead of HUGEINT or VARCHAR which causes "data corruption" / rounding.
This is especially bad for data with long numeric IDs getting inferred/casted to DOUBLE leading to wrong results down the road.

IMO duckdb's friendly and accessible approach should not modify these raw values.

I know I can explicitly provide data types, but the default should already cover this "problem" IMO 🤓 

### To Reproduce

data.csv
```
value,name
12345678901234567890,abc
99999999999999999999,abc
```

sql
```sql
from data.csv
```

result
```
┌────────────────────────┬─────────┐
│         value          │  name   │
│         double         │ varchar │
├────────────────────────┼─────────┤
│ 1.2345678901234567e+19 │ abc     │ --> value changed / rounded 
│                  1e+20 │ abc     │ --> value changed / rounded 
└────────────────────────┴─────────┘
```

problem:
- actual values changed

expectation:
- infer as HUGEINT
- infer as VARCHAR

### OS:

MacOs M3

### DuckDB Version:

DuckDB v1.5.2 (Variegata) 8a5851971f

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Julian Steden

### Affiliation:

Data Enthusiast

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22750

I think in the [current CSV implementation](https://github.com/duckdb/duckdb/blob/eb90869f574e7ab88099bb79a5b61a0460d83c2a/src/execution/operator/csv_scanner/sniffer/type_detection.cpp#L121264) we don't support (U)HUGEINT and BIGNUM, which means for all the values larger than BIGINT we interpret it as floating point value which loses precision on interpretation.

This PR adds supported for all integer types.
The behavior for CSV is tricky, which I tried to keep compatibility with the current one.
```sql
-- integer with positive sign is treated as VARCHAR
memory D CREATE TABLE large_integer_detection AS SELECT * FROM (VALUES('+3')) AS t(x);
memory D COPY large_integer_detection TO '/tmp/int.csv' (FORMAT CSV, HEADER);
memory D SELECT typeof(x), x FROM read_csv_auto('/tmp/int.csv');
┌───────────┬─────────┐
│ typeof(x) │    x    │
│  varchar  │ varchar │
├───────────┼─────────┤
│ VARCHAR   │ +3      │
└───────────┴─────────┘

-- space prefix and suffix is ignored and could be correctly interpreted as integer
memory D CREATE TABLE large_integer_detection AS SELECT * FROM (VALUES(' -3 ')) AS t(x);
memory D COPY large_integer_detection TO '/tmp/int.csv' (FORMAT CSV, HEADER);
memory D SELECT typeof(x), x FROM read_csv_auto('/tmp/int.csv');
┌───────────┬───────┐
│ typeof(x) │   x   │
│  varchar  │ int64 │
├───────────┼───────┤
│ BIGINT    │    -3 │
└───────────┴───────┘

-- integer with leading zeros are considered as VARCHAR
memory D CREATE TABLE large_integer_detection AS SELECT * FROM (VALUES(' 003 ')) AS t(x);
memory D COPY large_integer_detection TO '/tmp/int.csv' (FORMAT CSV, HEADER);
memory D SELECT typeof(x), x FROM read_csv_auto('/tmp/int.csv');
┌───────────┬─────────┐
│ typeof(x) │    x    │
│  varchar  │ varchar │
├───────────┼─────────┤
│ VARCHAR   │  003   │
└───────────┴─────────┘
```

I ran CI in my fork: https://github.com/dentiny/duckdb/pull/103 (failure seems unrelated)


src/include/duckdb/execution/operator/csv_scanner/csv_reader_options.hpp
```diff
@@ -83,9 +83,9 @@ struct CSVReaderOptions {
 	bool columns_set = false;
 	//! Types considered as candidates for auto-detection ordered by ascending specificity (~ from low to high)
 	vector<LogicalType> auto_type_candidates = {
-	    LogicalType::VARCHAR,      LogicalType::DOUBLE,    LogicalType::BIGINT,
-	    LogicalType::TIMESTAMP_TZ, LogicalType::TIMESTAMP, LogicalType::DATE,
-	    LogicalType::TIME,         LogicalType::BOOLEAN,   LogicalType::SQLNULL};
+	    LogicalType::VARCHAR, LogicalType::DOUBLE, LogicalType::BIGNUM,       LogicalType::UHUGEINT,
+	    LogicalType::HUGEINT, LogicalType::BIGINT, LogicalType::TIMESTAMP_TZ, LogicalType::TIMESTAMP,
```
pdet: So, the reason we only did BIGINT was because this was our biggest numeric type at the time. The idea was that we did not want to accidentally go to a smaller data type during sniffing.

With that said, I would probably only keep `LogicalType::HUGEINT`, and maybe `LogicalType::BIGNUM`. The main reason I'm not convinced about BIGNUM is that most arithmetic operations will still cast to double anyway.

dentiny: > The main reason I'm not convinced about BIGNUM is that most arithmetic operations will still cast to double anyway.

Thanks for the suggestion! Let me double check.

BIGNUM is the largest integer value we support; the issue with falling back to DOUBLE is we loss precision in some case.

dentiny: >  I would probably only keep LogicalType::HUGEINT, and maybe LogicalType::BIGNUM

I removed the UHUGEINT auto-detection, which is compatible that we auto detect BIGINT but not UBIGINT

pdet:
Thanks, LGTM!

PR #22805, opened 2026-05-21
Throw error if last token is unterminated string or quoted identifier

Dtenwolde:
Thanks @carlopi 

Reference issue: https://github.com/duckdblabs/duckdb-internal/issues/9376

If a query was being executed directly, using `-c` or `-cmd` we would not catch unterminated string literals or quoted identifiers properly during tokenization.


Mytherin:
Thanks!

PR #22811, opened 2026-05-21
Fix jemalloc packaging for JDBC/ODBC

staticlibs:
This PR adds a fix to a packaging script that is used when importing sources into JDBC and ODBC repos.

It also adds a missing header for `StringUtil` usage.

### DIFF
--- a/scripts/package_build.py
+++ b/scripts/package_build.py
@@ -39,6 +39,7 @@ def third_party_includes():
     includes += [os.path.join('third_party', 'vergesort')]
     includes += [os.path.join('third_party', 'yyjson', 'include')]
     includes += [os.path.join('third_party', 'zstd', 'include')]
+    includes += [os.path.join('third_party', 'jemalloc', 'include')]
     return includes
 
 
@@ -55,6 +56,7 @@ def third_party_sources():
     sources += [os.path.join('third_party', 'mbedtls')]
     sources += [os.path.join('third_party', 'yyjson')]
     sources += [os.path.join('third_party', 'zstd')]
+    sources += [os.path.join('third_party', 'jemalloc')]
     return sources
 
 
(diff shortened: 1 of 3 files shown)


carlopi:
Thanks!
Shouldn't this go into `v1.5-variegata` ?

staticlibs:
> Shouldn't this go into `v1.5-variegata`

No, I think it is not necessary there, we have a working setup (with this change applied as a patch) in `v1.5-variegata` branch of JDBC/ODBC. The idea is to use the existing setup with minimal changes for possible additional v1.5.x updates.

carlopi:
But isn't this more general? Like used also by possibly other clients?

staticlibs:
> But isn't this more general? Like used also by possibly other clients?

M, I believe it is not used by Python/Rust/Go. Not sure about other clients.

There should be no harm in adding it to v1.5-variegata too, just I thought that it is not necessary.

Mytherin:
Thanks. The CI failure looks related - the Swift build also uses that script and doesn't seem to like this somehow.

staticlibs:
> the Swift build also uses that script and doesn't seem to like this somehow

This appeared to be the case. I tried enabling jemalloc for Swift explicitly, but there is some kind of runtime incompatibility (seen recommendation to use shared jemalloc lib instead), so I filtered out jemalloc sources for Swift instead.

Mytherin:
Thanks!

krlmlr:
The R client chokes on that too, I'm working around locally.

Do you think that jemalloc be enabled in R, or are there good reasons to keep it off in this environment?

https://github.com/duckdb/duckdb-r/issues/2365

staticlibs:
@krlmlr 

Hi! In general in clients we try to keep jemalloc enabled where possible. There were a number of performance/memory issues reported when jemalloc was inadvertently disabled in earlier versions of 1.5. So in JDBC/ODBC we have it enabled (with [this packaging script](https://github.com/duckdb/duckdb-java/blob/main/vendor.py) and `-DDUCKDB_ENABLE_JEMALLOC` flag) on Linux and macOS.

PR #22815, opened 2026-05-21
Render MAP values as valid SQL in Value::ToSQLString()

carlopi:
Add a MAP case that emits the `MAP {k: v, ...}::MAP(...)` constructor literal, recursing through ToSQLString() on each key and value so nested MAPs/STRUCTs/LISTs are handled. The explicit `::type` cast is always appended (mirroring the DATE/BLOB cases) because a bare `MAP {...}` literal infers its element types from the entries -- and `MAP {}` infers MAP(INTEGER, INTEGER) -- so it does not faithfully round-trip on its own.

Bumped on this working on `quack`, have a workaround, but this seems generally a nice to have.

### DIFF
--- a/src/common/types/value.cpp
+++ b/src/common/types/value.cpp
@@ -1743,6 +1743,24 @@ string Value::ToSQLString() const {
 		ret += "]";
 		return ret;
 	}
+	case LogicalTypeId::MAP: {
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #22821, opened 2026-05-21
Auto generate transformer (Part 3)

Dtenwolde:
Part of https://github.com/duckdblabs/duckdb-internal/issues/9265

Now auto generating a bunch more `.gram` files
`alter.gram` is the biggest so far and it caused me to slightly refactor how the transformer rule for `ColumnConstraint` was done. 
```
ColumnConstraint <- NotNullConstraint / UniqueConstraint / PrimaryKeyConstraint / DefaultValue / CheckConstraint / ForeignKeyConstraint / ColumnCollation / ColumnCompression
```
This rule can return: `unique_ptr<ParsedExpression>`, `CompressionType`, `ConstraintType`, `unique_ptr<Constraint>`. 
This is slightly annoying and couldn't be auto-generated since the generator requires that all options in a choice return the same type. So now I wrapped these in the struct `ColumnConstraintEntry` together with the `ConstraintName`. This allows the transformer rules `ColumnDefinition` and `AddColumnEntry` to transform  the constraints more easily and we can auto generate these rules better.

Files auto-generated in this PR:
- `alter.gram`
- `comment.gram`
- `create_trigger.gram`
- `delete.gram`
- `explain.gram`
- `load.gram`
- `pragma.gram`
- `prepare.gram`
- `update.gram`


Mytherin:
Looks great, thanks!

PR #22822, opened 2026-05-21
Allow gaps in rowids

artjomPlaunov:
Allow gaps in rowids. This is being done to support vacuuming in the presence of indexes, since any dropped row groups were previously shifting up the entire logical rowid range, which requires rebuilding the entire index to keep rowid's consistent. If we allow for gaps, then we can perform vacuum task local rowid remappings within indexes instead of rebuilding the entire index. 

The vacuum task remapping will be a subsequent PR, however in this PR we already add support for allowing row groups to be dropped in the presence of indexes: this follows from allowing gaps since we can just drop a row group and subsequent nodes in the segment tree retain their row start values. 

Previously, total_rows served a dual purpose as the total number of rows, and the next logical row_id offset relative to the collection base. This PR splits off total_rows into total_rows (only used for tracking the former now), and next_row_id for the latter. 

This is backward compatible since reading an old file where rowid's are densely packed is the same as total_rows == next_rowid, i.e. all the gap aware logic still works, so we can assign next_row_id a default value of total_rows when reading an older file. 

As mentioned in TODO 3 - have not considered any further storage compatibility issues, need to look into the storage compatibility config failures. 

-------------------------

Extra notes:

- changed the binary search to be aware of gaps, as well as search in a [lower, upper) half open range, to get rid of the previous possible underflow/exception in the search. 
- Even though we are allowing gaps in rowid's, there are many row group collection/row group related code places that can still assume old invariants of densely packed rowid's, for example in appends, so in those locations we can still maintain old invariants where next_row_id == total_rows and there are no gaps. I also split up verify to check for either densely packed ranges or for gaps + ordering, depending where it is being called from.


src/storage/checkpoint/table_data_writer.cpp
```diff
@@ -207,6 +213,7 @@ void SingleFileTableDataWriter::FinalizeTable(const TableStatistics &global_stat
 	serializer.WriteList(
 	    104, "index_storage_infos", index_storage_infos.ordered_infos.size(),
 	    [&](Serializer::List &list, idx_t i) { list.WriteElement(index_storage_infos.ordered_infos[i].get()); });
+	serializer.WriteProperty(105, "next_row_id", next_row_id);
```
Mytherin: We can only write this when writing to the new serialization version (`serializer.ShouldSerialize(v2`)) - otherwise we break backwards compatibility on write.

test/tmp2.test
```diff
@@ -0,0 +1,51 @@
+# name: test/tmp2.test
```
Mytherin: These should be moved to `test/sql/storage' and renamed

src/storage/table/row_group_collection.cpp
```diff
@@ -1389,56 +1427,29 @@ void RowGroupCollection::InitializeVacuumState(CollectionCheckpointState &checkp
 		state.can_vacuum_deletes = false;
 		return;
 	}
-	bool dropped_any_rowgroups = false;
 	for (auto &entry : checkpoint_state.row_groups.SegmentNodes()) {
 		auto &row_group = entry.GetNode();
 		auto row_group_count = row_group.GetCommittedRowCount();
+		if (row_group_count == 0) {
+			// empty row group - we can drop it entirely.
+			// Dropping the whole row group does not remap row IDs since we preserve row_starts, i.e we allow for gaps
```
Mytherin: We can only do this when targeting the latest storage version - we need to verify this. Maybe we can add another bool to the commit state (`bool can_leave_gaps_in_row_ids`).

Mytherin (commented):
Thanks! Looks great - left some comments

artjomPlaunov:
@Mytherin thanks for the initial review! 

I handled the storage compatibility issues. In the TableDataWriter constructor, I now check:

```cpp
can_leave_gaps_in_row_ids = storage_compatibility.storage_version >= StorageVersion::V2_0_0;
```

Also in table_data_writer we check if we should write the new next_row_id field:

```cpp
void SingleFileTableDataWriter::FinalizeTable ....
	if (serializer.ShouldSerialize(StorageVersion::V2_0_0)) {
		serializer.WriteProperty(105, "next_row_id", next_row_id);
	}
	// ¬serializer.ShouldSerialize(StorageVersion::V2_0_0) ==> (next_row_id == total_rows)
	D_ASSERT(serializer.ShouldSerialize(StorageVersion::V2_0_0) || (next_row_id == total_rows));
}
```

so we have two separate flags: can_leave_gaps_in_row_ids, as well as can_change_row_ids (for example, we might be allowed to make gaps, but we do not support local vacuum remapping yet and so we are not allowed to change row_ids). 

In InitializeVacuumState, when we have an empty row group we check if we are not allowed to leave gaps in the rowid space (because a dropped row group might shift rowids). If we are not allowed to change rowid's, then we skip vacuuming this row group. 

We have to keep the trailing delete logic for targeting older storage versions; trailing deletes are implicitly handled by allowing row groups to be dropped + allowing gaps, but these are skipped if we are not allowed to leave gaps, so I had to reintroduce logic for trailing deletes when we are targeting older storage.

During checkpointing we also have to check if we are allowed to leave gaps or not; during execution we could create gaps, but if we are targeting an older storage version we need to repack the rowids (it shouldn't be possible to get to a state where we need to repack rowid's but are not allowed to shift rowid's, e.g. due to indexes, since that is caught earlier during initialize vacuum state. That is, it's possible to get to checkpoint with gaps in the rowid's, and we can repack them for older storage.

src/storage/table/row_group_collection.cpp
```diff
@@ -1566,8 +1645,12 @@ void RowGroupCollection::Checkpoint(TableDataWriter &writer, TableStatistics &gl
 			if (vacuum_tasks) {
 				// vacuum tasks were scheduled - don't schedule a checkpoint task yet
 				total_vacuum_tasks++;
-				vacuum_state.row_ids_changed = true;
-				writer.SetRowIdsChanged();
+				if (vacuum_state.can_change_row_ids) {
+					// With stable rowids, ScheduleVacuumTasks only permits merges that
```
Mytherin: I think this is still triggering `row_ids_changed` too often - we should probably always check if vacuums are "consecutive" to inform this setting

artjomPlaunov:
@Mytherin 

I changed the logic for rebuilding indexes under the vacuum_rebuild_indexes setting, i.e. by determining when we should set row_ids_changed. This is not exact, but this is the best over-approximation we can get for cheap. Also, somewhat obvious point, but all these conditions depend on can_change_rowids being set to true in the first place. 

1) In InitializeVacuumState, we can determine if we will do dense rowid repacking for legacy storage, by seeing if we have a gap followed by a row group with live rows in it. (This is exact). 

2/3 are in ScheduleVacuumTasks, and look at the shape of the candidate merge tasks to see if they *may* repack rowids. 

2) If a vacuum task is scheduled across a rowid gap, we know rowids will be changed (this part is exact). 

3) if a vacuum task is scheduled that has partial deletes, we know rowids *may* be changed (this is the over-approximation, but the only case we will spuriosuly rebuild now is if we have a compaction with trailing deletes in the final row group, but otherwise the rowids are contiguous across the row groups being selected for vacuuming).

Mytherin:
Thanks!

PR #22824, opened 2026-05-22
common/interval: guard micros against INT64_MIN negation in AGO parsing

Dev-X25874:
The AGO interval parser negates all three fields of `interval_t` to flip
the sign. Before doing so it checks that `months` and `days` are not at
their `int32_t` minimum (since negating INT_MIN is UB and wraps on
two's-complement targets), but the identical guard was missing for
`micros`, which is `int64_t`.

Add the missing check:

    result.micros == NumericLimits<int64_t>::Minimum()

to the existing condition so that all three fields are protected
uniformly before negation. Throws the same `OutOfRangeException` already
used for the other two fields.

Reproducer (before fix, silently returns wrong sign):
    SELECT INTERVAL '9223372036854775807 microseconds' AGO;

### DIFF
--- a/src/common/types/interval.cpp
+++ b/src/common/types/interval.cpp
@@ -280,7 +280,8 @@ interval_parse_ago:
 		}
 	}
 	// invert all the values
-	if (result.months == NumericLimits<int32_t>::Minimum() || result.days == NumericLimits<int32_t>::Minimum()) {
+	if (result.months == NumericLimits<int32_t>::Minimum() || result.days == NumericLimits<int32_t>::Minimum() ||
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - can we also add a test?

Dev-X25874:
@Mytherin test added in 90b173d — ready for another look.

Dev-X25874:
@Mytherin — test is in place, all three fields are now uniformly guarded. Would appreciate another look when you get a chance!

Mytherin:
Could you have a look at the failing CI?

PR #22825, opened 2026-05-22
Fix case-insensitive column match in INSERT ... SELECT ON CONFLICT

### LINKED ISSUE #22641: Regression in 1.4.0 when case insensitive column doesn't match declared case in INSERT on NOT NULL column
### What happens?

An `INSERT` query that used to work in 1.3.2 no longer works in 1.4.0 and later

### To Reproduce

```sql
create table t (i int not null primary key);
insert into t (I) values (1) on conflict do nothing;
```

This raises:

> SQL Error: Constraint Error: NOT NULL constraint failed: t.i

It should work. The workaround is to use a lower case `i` in the column list of the `INSERT` statement:

```sql
insert into t (i) values (1) on conflict do nothing;
```

Or, to omit the `ON CONFLICT` clause:

```sql
insert into t (I) values (1);
```

### OS:

Microsoft Windows [Version 10.0.26200.8246]

### DuckDB Version:

1.4.0 and later

### DuckDB Client:

JDBC

### Hardware:

_No response_

### Full Name:

Lukas Eder

### Affiliation:

Data Geekery

- [x] Yes, I have

- [x] Yes, I have

Yes

remsky:
Fixes #22641.

Cause: The `INSERT ... ON CONFLICT` rewrite compared column names case-sensitively, so `INSERT INTO t (I) SELECT ...` didn't match column `i` and filled it with NULL. 1.4.4 fixed this for the VALUES form but missed the `INSERT ... SELECT` form.

Fix: compare case-insensitively. 

Tests: added to `upsert_default.test`, ran the upsert/merge/insert suites and all passed

### DIFF
--- a/src/planner/binder/statement/bind_insert.cpp
+++ b/src/planner/binder/statement/bind_insert.cpp
@@ -432,7 +432,7 @@ unique_ptr<MergeIntoStatement> Binder::GenerateMergeInto(InsertStatement &stmt, 
 			// now push another subquery that adds the default columns
 			auto select_stmt = make_uniq<SelectStatement>();
 			auto select_node = make_uniq<SelectNode>();
-			unordered_set<string> set_columns;
+			case_insensitive_set_t set_columns;
 			for (auto &set_col : stmt.columns) {
 				set_columns.insert(set_col);
 			}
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - can we rebase this to `v1.5-variegata` so this can make it to the next bug-fix release?

remsky:
Hey @Mytherin, retargeted + rebased onto v1.5-variegata; should be good to go.

Mytherin:
Thanks!

PR #22830, opened 2026-05-22
Add format_bytes overload for HUGEINT

adityasingh2400:
## Summary

Fixes #17413. `format_bytes` (and its aliases `formatReadableSize`, `formatReadableDecimalSize`) previously only accepted `BIGINT`, so users hit a binder error when passing `HUGEINT` even though buffer sizes for very large datasets routinely exceed `INT64_MAX`. szarnyasg confirmed on the issue that a `HUGEINT` overload is wanted; this PR adds one.

## Implementation

- New `FormatBytesHugeintFunction` template, registered alongside the existing `BIGINT` variant in a `ScalarFunctionSet`.
- For absolute values that fit in `idx_t`, the new overload defers to the existing `StringUtil::BytesToHumanReadableString`, so behavior for typical inputs is byte-for-byte identical to the `BIGINT` path.
- Beyond `UINT64_MAX`, it walks the unit table in `HUGEINT` space using `Hugeint::DivModPositive` and extends the scale up to EiB/ZiB/YiB. This mirrors the pattern @Tishj pointed at in `GetAllocationSize`.
- `MIN` HUGEINT negation is handled via `TryNegate` with a one-byte clamp to `MAX` for the unrepresentable `|MIN|`.
- Updated `functions.json` to use `scalar_function_set` so the generator-managed header and function list stay in sync (verified by re-running `scripts/generate_functions.py`).

## Tests

Added cases to `test/sql/function/string/format_bytes.test`:
- Basic HUGEINT inputs (`0`, `1`, `1024`, `1099511627776`, `-1`, `NULL`)
- `INT64_MAX::HUGEINT` (verifies the small-value path produces the same `8191.9 PiB` as the `BIGINT` path)
- `1 ZiB` (`2^70`) and `1 YiB` (`2^80`) — past `UINT64_MAX`, exercising the HUGEINT-only path
- Negative HUGEINT past `UINT64_MAX`
- `formatReadableSize` / `formatReadableDecimalSize` HUGEINT inputs, including a decimal value past `UINT64_MAX` (`1 ZB`)


Mytherin:
Thanks - given that `bigint -> hugeint` auto-casts, can we instead only have the hugeint function and remove the `bigint` function?

test/sql/function/string/format_bytes.test
```diff
@@ -126,3 +126,74 @@ query I
...
+
+# At INT64_MAX the HUGEINT path defers to the BIGINT formatter, so output matches.
+query I
+SELECT format_bytes(9223372036854775807::HUGEINT);
+----
+8191.9 PiB
+
+# 1 ZiB = 2^70 = 1180591620717411303424 (past UINT64_MAX, exercises the HUGEINT-only path)
+query I
+SELECT format_bytes(1180591620717411303424::HUGEINT);
```
Mytherin, line 169: Can we try `SELECT format_bytes(hugeint) from test_all_types()` to try the bounds?

adityasingh2400:
Good point on the auto-cast. I will drop the bigint overload and keep only the hugeint one, and post the format_bytes(hugeint) from test_all_types() output to verify the bounds.

adityasingh2400:
Pushed 0e842c7 which drops the BIGINT registrations from both format_bytes and formatReadableDecimalSize so they rely on the BIGINT to HUGEINT auto-cast.

On the test_all_types() check, I could not finish a local reldebug build to capture the live output (the link-time static library is about 1.2GB and my disk only has around 1.6GB free), but tracing through the HUGEINT path by hand for the three rows that test_all_types() produces, the expected output for SELECT format_bytes(hugeint) FROM test_all_types() is:

row 0 (HUGEINT min, -2^127): -140737488355327.9 YiB
row 1 (HUGEINT max, 2^127-1): 140737488355327.9 YiB
row 2 (NULL): NULL

The min row exercises the TryNegate clamp (since -(-2^127) is not representable) which falls back to HUGEINT_MAX, so the magnitude printed matches the max row. Both rows go through the HUGEINT walk (since 2^127-1 does not fit in idx_t). Happy to follow up with the actual captured output as soon as I can free up disk space, or let CI confirm.

PR #22834, opened 2026-05-22
Allow collation for VARCHAR alias column types (e.g. CHAR, BPCHAR, STRING, TEXT)

yan-alex:
Previously this would throw: Only VARCHAR columns can have collations!


Mytherin:
Thanks!

PR #22835, opened 2026-05-22
Fix undefined behavior in Parquet prefetch registration

leviathan0992:
Description:
=======================
RegisterPrefetch stored ReadHead pointers in a std::set whose comparator depends on the pointed-to range location and size. When a new range was merged into an existing ReadHead, the ReadHead was mutated while it was still in the set. The result can be undefined because mutating key fields of an element that participates in std::set ordering violates the set ordering invariant.

Solution:
========================
Replace the set-based merging with an explicit prefetch finalization step. The prefetch flow is now:

1. Register prefetch ranges and validate that each requested range is inside the file.
2. Keep mergeable ranges in a pending list, while non-mergeable ranges are added directly to the finalized read-head list.
3. Finalize registration by sorting pending ranges by file offset and coalescing overlapping, adjacent, and READ_HEAD_MERGE_GAP-close ranges.
4. Prefetch finalized read heads, skipping ranges that were already materialized by the read path to avoid duplicate IO.

This is my first PR to DuckDB. Please let me know if I missed anything.

### DIFF
--- a/extension/parquet/include/thrift_tools.hpp
+++ b/extension/parquet/include/thrift_tools.hpp
@@ -50,85 +50,54 @@ struct ReadHead {
 	}
 };
 
-struct ReadHeadComparator {
-	static constexpr uint64_t DEFAULT_ACCEPTED_COLUMN_GAP = 1 << 14; // 16 KiB
+// Maximum gap between prefetch ranges that can still be coalesced.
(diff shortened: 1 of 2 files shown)


leviathan0992:
@pdet Hi, I updated this patch to address the failing case.

Mytherin:
Seems like there's some merge conflicts - could you have another look?

Issue #22716, opened 2026-05-17
FPE crash when progress bar enabled


jjnsn:
### What happens?

When using the DuckLake and SQLite extensions, DuckDB crashes when the progress bar is enabled and runs OK when it is not

### To Reproduce

```
.timer on

PRAGMA disable_progress_bar; --enable this to cause crash
PRAGMA explain_output='all';

INSTALL ducklake;
LOAD ducklake;
INSTALL sqlite;
LOAD sqlite;

SET memory_limit='1GB';
SET threads=1;
SET temp_directory='./tempfiles';

ATTACH 'ducklake:sqlite:metadata.sqlite' AS dl (DATA_PATH 'data_files/');

CREATE TABLE dl.metars AS SELECT * FROM read_parquet('../raw_metars') LIMIT 0;

ALTER TABLE dl.metars SET PARTITIONED BY (YEAR(timeObs), MONTH(timeObs));
ALTER TABLE dl.metars SET SORTED BY (stationName ASC, timeObs ASC);

CALL dl.set_option('parquet_compression', 'zstd');
CALL dl.set_option('parquet_compression_level', 12);

INSERT INTO dl.metars SELECT * FROM read_parquet('../raw_metars/') ;
```

When I `cat` this into duckdb, eg `cat create.sql | duckdb`

Produces this crash when I run it with gdb:

```Thread 1 "duckdb" received signal SIGFPE, Arithmetic exception.
0x0000000000894cc7 in duckdb_shell::ShellProgressBarDisplay::PrintProgressInternal(int, double, bool) ()
(gdb) bt
#0  0x0000000000894cc7 in duckdb_shell::ShellProgressBarDisplay::PrintProgressInternal(int, double, bool) ()
#1  0x000000000095e108 in duckdb::TerminalProgressBarDisplay::Update(double)
    ()
#2  0x000000000095a605 in duckdb::ProgressBar::Update(bool) ()
#3  0x0000000000d6e8c5 in duckdb::ClientContext::ExecuteTaskInternal(duckdb::ClientContextLock&, duckdb::BaseQueryResult&, bool) ()
#4  0x0000000000d6eaac in duckdb::PendingQueryResult::ExecuteInternal(duckdb::ClientContextLock&) ()
#5  0x0000000000d6ec56 in duckdb::PendingQueryResult::Execute() ()
[...]

PR #22836, opened 2026-05-22
fix progress bar output and crash when piping SQL

DinosL:
When running `duckdb < query.sql`, the progress bar produced a lot of trailing spaces per update line and on some occasions crashed with a floating point exception. When stdin is a pipe or file, `ioctl` fails and leaves the width uninitialized, returning a garbage value (or zero). The progress bar then attempted to render a line of that width, causing an arithmetic overflow and SIGFPE.

Using a default value of 120 feels a bit hacky, but we already follow the same approach in `BoxRendererImplementation::Initialize()`, so I applied it here as well for consistency.

Fixes: https://github.com/duckdb/duckdb/issues/22716

### DIFF
--- a/src/common/printer.cpp
+++ b/src/common/printer.cpp
@@ -73,8 +73,11 @@ idx_t Printer::TerminalWidth() {
 	rows = csbi.srWindow.Right - csbi.srWindow.Left + 1;
 	return rows;
 #else
-	struct winsize w;
+	struct winsize w = {};
 	ioctl(0, TIOCGWINSZ, &w);
+	if (w.ws_col == 0) {
+		return 120;
+	}
 	return w.ws_col;
 #endif
 #else
--- a/tools/shell/shell.cpp
+++ b/tools/shell/shell.cpp
@@ -1192,7 +1192,9 @@ void ShellState::OpenDB(ShellOpenFlags flags) {
 			}
 		}
 		auto &client_config = duckdb::ClientConfig::GetConfig(*conn->context);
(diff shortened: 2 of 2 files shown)


Tishj:
Is it maybe possible to disable the progress bar if we detect that output is being redirected?

DinosL:
That's the other solution I considered.

```diff
git diff
diff --git a/tools/shell/shell.cpp b/tools/shell/shell.cpp
index 2568026670..c644c788f6 100644
--- a/tools/shell/shell.cpp
+++ b/tools/shell/shell.cpp
@@ -1192,7 +1192,9 @@ void ShellState::OpenDB(ShellOpenFlags flags) {
                        }
                }
                auto &client_config = duckdb::ClientConfig::GetConfig(*conn->context);
-               client_config.display_create_func = CreateProgressBar;
+               if (stdin_is_interactive) {
+                       client_config.display_create_func = CreateProgressBar;
+               }
 #ifdef SHELL_INLINE_AUTOCOMPLETE
                db->LoadStaticExtension<duckdb::AutocompleteExtension>();
 #endif
```

I went with fixing `TerminalWidth()` instead because it is a bug regardless of the progress bar context. That said, I am happy to go either way.

Mytherin:
Thanks for the PR! Looks good. Can we do both fixes? Only it shouldn't be `stdin_is_interactive` but rather `stdout_is_console`.

DinosL:
Done!

Mytherin:
Thanks!

Issue #22810, opened 2026-05-21
CLI 1.5.3: terminal background-color probe response echoed into REPL on Ghostty (inverse of #20029)


cs2018ncsa:
### What happens?

DuckDB CLI 1.5.3 on Ghostty enters an infinite loop on startup. The terminal's OSC 11 background-color response is read back into the REPL as input, which triggers a redraw, which re-queries, which echoes again — forever.

Output (excerpt, from a fresh `duckdb -init /dev/null`):

```
DuckDB v1.5.3 (Variegata)
rgb:1a1a/1b1b/2626^G^[]11;rgb:1a1a/1b1b/2626^G^[]11;rgb:1a1a/1b1b/2626^G^[]11;...
```

This is the inverse failure of #20029. There, VSCode's terminal never responded and DuckDB hung in `read()`. Here, Ghostty responds per spec, but `Terminal::TryGetBackgroundColor` (terminal.cpp:431, called from `linenoiseGetTerminalColorMode` at linenoise-c.cpp:160) appears not to drain the full response, so the leftover bytes are consumed by the line editor as user input.

### To Reproduce

1. Install DuckDB v1.5.3 on macOS
2. Open Ghostty 1.3.1
3. Run: `duckdb -init /dev/null`

### Workarounds tried (none worked)

- `NO_COLOR=1 duckdb`
- `duckdb -init /dev/null`

### OS

macOS 26.4.1 (Darwin 25.4.0)

### DuckDB Version

v1.5.3 (Variegata) 14eca11bd9

### DuckDB Client

CLI

### Terminal

Ghostty 1.3.1 stable (Zig 0.15.2, ReleaseFast, Metal renderer)

Related: #20029, #15930

cs2018ncsa:
Additional terminals tested with `duckdb -init /dev/null` on macOS 26.4.1:

| Terminal | Symptom | Usable? |
|---|---|---|
| Ghostty 1.3.1 | Infinite loop, response continually re-echoed | No |
| Zed embedded terminal (alacritty TUI) | Single character (`/f`) bleeds at startup, REPL functional after | Yes, cosmetic only |

Both terminals are strictly OSC-conformant. The difference appears to be in response timing/buffering — when the response arrives fast enough to land before the line editor activates fully, the entire sequence loops; when a tail character arrives slightly later, only that tail leaks.

This suggests the fix needs to drain stdin to a terminator (BEL `\a` or ST `\e\\`) rather than relying on a fixed read window.

[...]

PR #22838, opened 2026-05-22
Remove time-out waiting for terminal background color

Mytherin:
Hopefully fixes https://github.com/duckdb/duckdb/issues/22810

This time-out was added as a work-around for the color detection being broken in some terminals - but it can itself break terminals if we exceed this time-out. We should just remove it and wait until a full response is returned.


cs2018ncsa:
Holy quick fix! 

Confirmed — built the PR branch locally and the infinite loop is gone.

Tested on:
- Ghostty 1.3.1 inside tmux (TERM=tmux-256color)
- Ghostty 1.3.1 standalone

Stock 1.5.3 still spews `^[]11;rgb:...` forever in both; the patched build starts cleanly. Thanks!

PR #22839, opened 2026-05-22
Reject MERGE clauses that follow an unconditional clause of the same condition

### LINKED ISSUE #22624: MERGE unexpectedly reordered WHEN MATCHED clauses, if the catch all clause is not at the end
### What happens?

The `MERGE` statement can be made to delete a row that should have been already updated by a previous `MATCHED` clause

### To Reproduce

```sql
create table t (id int primary key, val text);
insert into t (id) values (1),(2),(3),(4);

merge into t
using (select 1)
on t.id > 2
when matched then update set val = 'x'
when matched and id = 4 then delete;

select * from t;
```

This produces:

```
|id |val|
|---|---|
|1  |   |
|2  |   |
|3  |x  |
```

It should produce

```
|id |val|
|---|---|
|1  |   |
|2  |   |
|3  |x  |
|4  |x  |
```

Alternatively, an error could be raised because the "catch all" `MATCHED` clause (without `AND` condition) should not be followed by an additional `MATCHED` clause, which should never apply.

A workaround is to add `AND TRUE` to the "catch all" `MATCHED` clause (which should produce equivalent behaviour):

```sql
merge into t
using (select 1)
on t.id > 2
when matched and true then update set val = 'x'
when matched and id = 4 then delete;
```

When looking at this example:

```sql
merge into t
using (select 1)
on t.id > 2
when matched then update set val = coalesce(val, '') || 'x'
when matched and id = 4 then update set val = coalesce(val, '') || 'y';
```

The result is:

```
|id |val|
|---|---|
|1  |   |
|2  |   |
|3  |x  |
|4  |y  |
```

Perhaps, it appears that the `WHEN MATCHED` clauses are re-ordered, unexpectedly.

### OS:

Microsoft Windows [Version 10.0.26200.8246]

### DuckDB Version:

1.5.2.1

### DuckDB Client:

JDBC

### Hardware:

_No response_

### Full Name:

Lukas Eder

### Affiliation:

Data Geekery

- [x] Yes, I have

- [x] Yes, I have

Yes

adityasingh2400:
A conditional WHEN [NOT] MATCHED clause that follows an unconditional clause of the same condition type is unreachable, but the parser was silently reordering all unconditional clauses to the end of their condition group. In the example from #22624, an unconditional UPDATE got reordered behind a conditional DELETE, so a row that should have been updated was deleted instead.

This change rejects the query at parse time so the unreachable clause cannot go unnoticed, and removes the reorder pass so the remaining clauses execute in declaration order. The error message also points at the unreachable clause rather than only mentioning duplicate unconditional clauses, since the same situation now covers both cases. Existing tests in test/sql/merge already place catch-all clauses last and continue to pass; the existing duplicate-unconditional test in test/sql/merge/merge_into.test was updated to match the new wording.

The new regression test in test/issues/general/test_22624.test covers both the rejection of the original reproducer and that the equivalent declaration-order form still runs and produces the expected rows.

Closes #22624

### DIFF
--- a/src/parser/peg/transformer/transform_merge_into.cpp
+++ b/src/parser/peg/transformer/transform_merge_into.cpp
@@ -1,3 +1,4 @@
+#include "duckdb/common/set.hpp"
 #include "duckdb/parser/peg/transformer/peg_transformer.hpp"
 #include "duckdb/parser/statement/merge_into_statement.hpp"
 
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #22840, opened 2026-05-22
Make `GetSQLValueFunction` pluggable through a planner extension

Mytherin:
In DuckDB we have the concept of "sql value functions" - which are functions that column references get resolved to if none is found, e.g.:

```sql
SELECT current_timestamp;
```

We have a list of functions that DuckDB supports, but it might be desirable for extensions to modify / change this list. This makes that possible.

### DIFF
--- a/src/planner/column_qualifier.cpp
+++ b/src/planner/column_qualifier.cpp
@@ -7,6 +7,7 @@
 #include "duckdb/parser/expression/operator_expression.hpp"
 #include "duckdb/parser/expression/positional_reference_expression.hpp"
 #include "duckdb/planner/expression_binder/having_binder.hpp"
+#include "duckdb/planner/planner_extension.hpp"
 
 namespace duckdb {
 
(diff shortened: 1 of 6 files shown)


src/planner/column_qualifier.cpp
```diff
@@ -45,7 +46,14 @@ string GetSQLValueFunctionName(const string &column_name) {
 	return string();
 }
 
-unique_ptr<ParsedExpression> ColumnQualifier::GetSQLValueFunction(const string &column_name) {
+unique_ptr<ParsedExpression> Binder::GetSQLValueFunction(const string &column_name) {
+	for (auto &ext : PlannerExtension::Iterate(context)) {
+		if (ext.get_sql_value_function) {
+			PlannerExtensionInput input {context, *this, ext.planner_info.get()};
+			return ext.get_sql_value_function(input, column_name);
```
Tishj: Don't you want this to first check if the result isn't `nullptr`, and let other extensions try until something succeeds?
Or is the idea to also let the extensions override existing sql value functions to disable them?
Perhaps we should differentiate in the return value to make that clear, differentiating between "I don't substitute this" and "I don't want this to be substituted at all"

Mytherin: Yeah the idea is they can also disable the built-in ones. I agree a different return value would be cleaner.

PR #22843, opened 2026-05-22
Fix small-build dictionary emission activation gate

ahmetkarapinar:
# Fix small-build dictionary emission activation gate

Small-build dictionary emission ([#22340](https://github.com/duckdb/duckdb/pull/22340)) is silently inactive on `main`. The optimization targets *small* build sides (≤ 100K rows after the 8 MiB payload cap), and any build under `PARALLEL_CONSTRUCT_THRESHOLD = 1,048,576` rows finalizes single-threaded — so every workload it was designed for hits a finalize path that bypasses it.

The gates in `CanUseDictionaryEmission` still work; they just aren't consulted. A later finalize refactor (`08c2bb766c`, "Remove code duplication of fast-path") added a `build_dictionary_arrays` parameter to `FinishTasks` — `false` on the single-threaded `ExecuteDirectly()` path, `true` on the parallel `FinishEvent()` path — which short-circuits the activation call before the gates run on small builds. Confirmed with the original author as an oversight.

**Safety.** `BuildDictionaryArrays` depends only on invariants both finalize paths satisfy at the call site — all finalize work complete, `chains_longer_than_one` finalized, every `NEXT_PTR` written, data collection pinned (`VerifyEverythingPinned()` runs inside `FinishTasks` on both paths). External joins remain excluded (`CanUseDictionaryEmission` returns `false` when `external`); perfect hash join never reaches `ScheduleFinalize` / `FinishEvent`; the `aux_next_ptrs` allocation path inside `BuildDictionaryArrays` is identical regardless of caller. The forced-parallel run already exercises the same activation on the same data shape and produces correct results; this fix routes the single-threaded path through the same gate and builder, on hash tables that have already passed the same pinning check.

## Update: recursive-CTE regression fix

While verifying this branch under recursive CTEs, I found a separate latent bug that surfaced once the gate fix re-enabled dictionary emission on the single-threaded finalize path. Commit b50745ef83 (recursive-CTE state-reuse) added `ResetForNewIterationSinglePartition` but didn't clear the build-side dict-emission state (`dict_arrays`, `aux_next_ptrs`, `aux_next_ptrs_data`, `use_dict_emission`). As a result, `dict_arrays` accumulated across iterations, which blew up runtime (`aoc24.test_slow` went from single-digit seconds to >25 minutes) and, in release builds, produced wrong values as later iterations indexed into stale dictionaries from earlier iterations.

The fix is a four-line teardown at the end of `ResetForNewIterationSinglePartition`, kept in lock-step with `BuildDictionaryArrays`. Added two recursive-CTE regression cases to `hash_join_dictionary_emission.test_slow` that fail without the fix and pass with it.

---

@kryonix — as discussed on Discord, this restores activation of [#22340](https://github.com/duckdb/duckdb/pull/22340). @lnkuiper for visibility.

### DIFF
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -2165,6 +2165,12 @@ void JoinHashTable::ResetForNewIterationSinglePartition() {
 	prefix_range_filter.reset();
 	should_build_prefix_range_filter = false;
 	ResetCorrelatedMarkJoinInfo(*this);
+	// The next iteration may rebuild a different small build side, so this iteration's dictionary
+	// state is stale. Keep in lock-step with BuildDictionaryArrays, which sets these four fields.
+	dict_arrays.clear();
+	aux_next_ptrs.Reset();
+	aux_next_ptrs_data = nullptr;
+	use_dict_emission = false;
 }
 
 bool JoinHashTable::PrepareExternalFinalize(const idx_t max_ht_size) {
--- a/src/execution/operator/join/physical_hash_join.cpp
+++ b/src/execution/operator/join/physical_hash_join.cpp
@@ -898,7 +898,7 @@ public:
 		auto &ht = *sink.hash_table;
 		auto prefix_range_state = ht.ShouldBuildPrefixRangeFilter() ? RegisterPrefixRangeState(ht) : nullptr;
 		ExecuteHashJoinFinalizeTask(sink, optional_idx(), prefix_range_state);
-		FinishTasks(false);
+		FinishTasks();
 	}
 
 	void Schedule() override {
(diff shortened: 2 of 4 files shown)


ahmetkarapinar:
Hi Laurens! Sorry, I rebased onto latest main to refresh CI, which auto-disabled the auto-merge. Should be safe to re-enable or merge directly once CI clears.

Thanks!

lnkuiper:
Thanks! I think the thread sanitizer issue is unrelated, but I can't merge unless everything is green. I'll rerun that one once I can :)

ahmetkarapinar:
Hey @lnkuiper! Just checking in, anything you need from my side? 🙂

lnkuiper:
Could you update to latest `main` and try to run CI again? I think this is still hanging since a GitHub actions outage a few days ago

ahmetkarapinar:
> Could you update to latest `main` and try to run CI again? I think this is still hanging since a GitHub actions outage a few days ago

Rebased onto latest main. The failing test (csv_disk_reload.test_slow) looks unrelated to this PR.

ahmetkarapinar:
Hey @lnkuiper ! I identified the issue with the recursive-CTE regression. It turns out commit b50745ef83 (recursive-CTE state-reuse) added ResetForNewIterationSinglePartition but didn't clear the build-side dict-emission state (dict_arrays, aux_next_ptrs, use_dict_emission). Once the gate fix on this branch re-enabled emission on the single-threaded finalize path, this latent bug became reachable, dict_arrays just accumulated across iterations, blowing up runtime and (in release builds) producing wrong values from stale dictionaries.

The fix is small: clear the four dict-emission fields at the end of ResetForNewIterationSinglePartition, in lock-step with BuildDictionaryArrays. I also added recursive-CTE regression cases to hash_join_dictionary_emission.test_slow that fail without the fix and pass with it.

Thanks!

lnkuiper:
Great that you found this! Looks good :)

PR #22848, opened 2026-05-22
Fix #22791: nested window self-join leaves stale bindings

### LINKED ISSUE #22791: 1.5.3 regression: nested partitioned window query fails with `unordered_map::at` unless `window_self_join` optimizer is disabled
### What happens?

DuckDB 1.5.3 fails on a valid nested window query with:

```text
Invalid Error: unordered_map::at: key not found
```

The same SQL succeeds on DuckDB 1.5.2. It also succeeds on DuckDB 1.5.3 if the
`window_self_join` optimizer is disabled:

```sql
SET disabled_optimizers='window_self_join';
```

### To Reproduce

```sql
SELECT version();

CREATE TABLE t(id INTEGER, x DOUBLE);
INSERT INTO t VALUES (1, 10.0);

SELECT id, SUM(x) OVER (PARTITION BY id) AS outer_sum
FROM (
  SELECT id, x, LAG(total, 1, 0) OVER (PARTITION BY id) AS prev_total
  FROM (
    SELECT id, x, COUNT(*) OVER (PARTITION BY id) AS total
    FROM t
  ) AS a
) AS b;
```

### OS:

macOS 26.4.1 arm64

### DuckDB Version:

1.5.3

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Lysle Boller

### Affiliation:

University of Pennsylvania

- [x] Yes, I have

- [x] Yes, I have

kevintruong:
Closes #22791

cc @DinosL — picking this up as discussed in the issue thread.

The `WindowSelfJoinOptimizer` recurses with a single shared `ColumnBindingReplacer` that is only flushed at the end of `Optimize`. When an inner aggregate window gets rewritten into a join+aggregate, a non-aggregate window above it (e.g. `LAG`) is left with stale `BoundColumnRef` bindings to the removed window operator. `DeepCopy` of that subtree for the outer-window rewrite then carries those stale references into the clone, triggering `unordered_map::at: key not found` on subsequent type resolution.

Fix: use a local replacer per child and flush it into the current operator before any further transformation, mirroring `TopNWindowElimination::OptimizeInternal`.

Regression test added to `test/sql/optimizer/test_window_self_join.test`.

### DIFF
--- a/src/optimizer/window_self_join.cpp
+++ b/src/optimizer/window_self_join.cpp
@@ -122,8 +122,18 @@ unique_ptr<LogicalOperator> WindowSelfJoinOptimizer::OptimizeInternal(unique_ptr
 	if (op->type == LogicalOperatorType::LOGICAL_WINDOW) {
 		auto &window = op->Cast<LogicalWindow>();
 
-		// Check recursively
-		window.children[0] = OptimizeInternal(std::move(window.children[0]), replacer);
+		// Check recursively using a local replacer so that any deferred binding replacements
+		// produced by an inner-window rewrite are flushed into this window before we proceed.
+		// Without this, a non-optimizable window sitting between two optimizable windows
+		// would carry stale BoundColumnRef bindings into DeepCopy, leaving references that
+		// point at operators outside the cloned subtree (issue #22791).
+		ColumnBindingReplacer local;
+		window.children[0] = OptimizeInternal(std::move(window.children[0]), local);
+		if (!local.replacement_bindings.empty()) {
+			local.VisitOperator(*op);
+			replacer.replacement_bindings.insert(replacer.replacement_bindings.end(),
+			                                     local.replacement_bindings.begin(), local.replacement_bindings.end());
+		}
 
 		auto &w_expr0 = window.expressions[0]->Cast<BoundWindowExpression>();
 		for (auto &expr : window.expressions) {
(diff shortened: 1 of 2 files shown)


hawkfish:
Hi @kevintruong -

Thanks for the PR!  But I'm not convinced this rewrite should even apply if there are stacked window operators as joining them could be very resource intensive. I have a different fix in test that simply restores rejecting the upper level optimisations.

kevintruong:
@hawkfish : I don't aware about it , thanks for noticing.  no problem . Could you give me your branch . I want to learn about your way

hawkfish:
> @hawkfish : I don't aware about it , thanks for noticing. no problem . Could you give me your branch . I want to learn about your way

Actually your analysis was much more thorough than mine! I just looked at the plan and went "Why"?

I just restored a check that was removed - the PR is [here](https://github.com/duckdb/duckdb/pull/22844).

kevintruong:
Hi @hawkfish  
Confirmed by inspection — #22844  restores the pre-refactor CanRebind whitelist that I should have seen as the right shape from history. Closing this in favor of it. My regression test variant (3-row, multi-partition) is in the diff at test/sql/optimizer/test_window_self_join.test; happy to PR it on top of your PR once merged if useful.

hawkfish:
> My regression test variant (3-row, multi-partition) is in the diff at test/sql/optimizer/test_window_self_join.test; happy to PR it on top of your PR once merged if useful.

I think I already grabbed the test from the report?

hawkfish:
Actually, I think this would be a useful addition to the code, so feel free to add it after my PR is in.

Issue #22826, opened 2026-05-22
INTERNAL Error: Attempted to access index 0 within vector of size 0


DerZc:
### What happens?

Hi,

The following test case triggers an internal error.

First, create a JSON file /tmp/timestamp.json with:

```
{
    "ts": ["2022-06-01T06:41:58Z", "2021-08-21T08:26:55.5Z", "2009-11-15T21:58:54.636Z"]
}
```
Then execute the following SQL:

```
CREATE TABLE t(ts TIMESTAMP[]);
COPY t FROM '/tmp/timestamp.json'
    (FORMAT JSON, timestampformat '%Q');
```

This is the output:

```
INTERNAL Error:
Attempted to access index 0 within vector of size 0

Stack Trace:

./build/release/duckdb(+0x92a63d) [0x6382b3a8b63d]
./build/release/duckdb(+0x92a6b8) [0x6382b3a8b6b8]
./build/release/duckdb(+0x92e695) [0x6382b3a8f695]
./build/release/duckdb(+0x64dcbb) [0x6382b37aecbb]
./build/release/duckdb(+0x6726b7) [0x6382b37d36b7]
./build/release/duckdb(+0x15b90ed) [0x6382b471a0ed]
./build/release/duckdb(+0x15ba38a) [0x6382b471b38a]
./build/release/duckdb(+0x15ba885) [0x6382b471b885]
./build/release/duckdb(+0x1e16524) [0x6382b4f77524]
./build/release/duckdb(+0x1e18c68) [0x6382b4f79c68]
./build/release/duckdb(+0x1e1d145) [0x6382b4f7e145]
./build/release/duckdb(+0x1e19432) [0x6382b4f7a432]
./build/release/duckdb(+0x1e1bb97) [0x6382b4f7cb97]
./build/release/duckdb(+0x1e60e1b) [0x6382b4fc1e1b]
./build/release/duckdb(+0x1e60fbe) [0x6382b4fc1fbe]
./build/release/duckdb(+0x1e4c7ee) [0x6382b4fad7ee]
./build/release/duckdb(+0x1b96bed) [0x6382b4cf7bed]
./build/release/duckdb(+0xb47713) [0x6382b3ca8713]
./build/release/duckdb(+0xb5447c) [0x6382b3cb547c]
./build/release/duckdb(+0xb54872) [0x6382b3cb5872]
./build/release/duckdb(+0xb49884) [0x6382b3caa884]
[...]

PR #22850, opened 2026-05-22
Fix: check if `ParseFormatSpecifier` returns unrecognized format

DinosL:
The return value of `ParseFormatSpecifier` was silently disregarded in `DateFormatMap::AddFormat`, which is a problem when an unrecognized format is passed to `timestampformat`. 

fixes: https://github.com/duckdb/duckdb/issues/22826


Mytherin:
Thanks!

PR #22852, opened 2026-05-23
Fix alias propagation when replacement scan is wrapped in SubqueryRef

### LINKED ISSUE #22841: CLI: Cannot alias underscore syntax (used to reference the last result)
### What happens?

CLI: Underscore syntax (to reference previous table) throws an message when aliased and accessed in SELECT.

### To Reproduce

```sql
select 'a' as x;
select * from _ as d1, _ as d2;
```

This works fine:

```text
┌─────────┬─────────┐
│    x    │    x    │
│ varchar │ varchar │
├─────────┼─────────┤
│ a       │ a       │
└─────────┴─────────┘
```

```sql
select 'a' as x;
select d1.x, d2.x from _ as d1, _ as d2;
```

This fails:

```console
Binder Error:
Referenced table "d1" not found!
Candidate tables: "unnamed_subquery", "unnamed_subquery2"

LINE 1: select d1.x, d2.x from _ as d1, _ as d2;
               ^
```

### OS:

any

### DuckDB Version:

1.5.3 and 1.5.4-nightly

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Gabor Szarnyas

### Affiliation:

DuckDB Labs

- [x] Yes, I have

- [x] Yes, I have

Yes

truffle-dev:
Closes #22841.

The `_` replacement scan in the shell extension returns a `ColumnDataRef`, which falls through the `TABLE_FUNCTION` / `SUBQUERY` branches in `BindWithReplacementScan` and gets wrapped in a fresh `SubqueryRef`. The alias was applied to the inner `replacement_function` before the wrap, but never carried onto the outer `SubqueryRef`, so the wrap ended up unaliased — hence the `unnamed_subquery` fallback when running `SELECT d.x FROM _ AS d`.

In the else-branch wrap, save the inner ref's alias into a local before the move and assign it onto the wrapping `SubqueryRef` afterward. The inner ref's alias is left intact (the binder needs it set for scope resolution) and the outer wrap now carries the same alias, so qualified references resolve as expected. `TABLE_FUNCTION` and `SUBQUERY` are unaffected.

Added two shell tests covering qualified column references and self-join over `_`.

### DIFF
--- a/src/planner/binder/tableref/bind_basetableref.cpp
+++ b/src/planner/binder/tableref/bind_basetableref.cpp
@@ -55,6 +55,13 @@ BoundStatement Binder::BindWithReplacementScan(ClientContext &context, BaseTable
 		if (!replacement_function) {
 			continue;
 		}
+		if (!ref.alias.empty()) {
+			// user-provided alias overrides the default alias
+			replacement_function->alias = ref.alias;
+		} else if (replacement_function->alias.empty()) {
+			// if the replacement scan itself did not provide an alias we use the table name
+			replacement_function->alias = ref.table_name;
+		}
 		if (replacement_function->type == TableReferenceType::TABLE_FUNCTION) {
 			auto &table_function = replacement_function->Cast<TableFunctionRef>();
 			table_function.column_name_alias = ref.column_name_alias;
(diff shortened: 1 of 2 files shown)


truffle-dev:
The first commit moved the alias propagation block to run after the type-dispatch instead of before. That left the inner `ColumnDataRef` with an empty alias, and the binder calls `BindingAlias::GetAlias` on it during scope resolution, so every replacement-scan query crashed with `Calling BindingAlias::GetAlias on a non-set alias`.

50df5951 keeps the pre-dispatch propagation intact (inner ref's alias stays set like upstream) and carries the alias to the wrapping `SubqueryRef` from inside the else-branch via a local save before the move. Net diff vs upstream is +5 lines, all in the else branch. Updated the PR body to match.

Mytherin:
Thanks! Makes sense to me - could you just rebase to `v1.5-variegata`?

truffle-dev:
Rebased onto v1.5-variegata (78ff7cf).

Mytherin:
Thanks!

truffle-dev:
Thanks for picking it up.

PR #21211, opened 2026-03-06
Improve test runner

smvv:
### Context

This is a follow-up of https://github.com/duckdb/duckdb/pull/21181.

### Important changes

- Filter `skip_tests` entries from `--list-tests` with `--test-config`.
  - For safety, I've added a runtime error for a `--test-config` without `skip_compiled=true`. We might need additional filtering to support that use case.
- Add flag `--test-config=...` to forward `--test-config=...` easily to the unittest binary.
- Use new test runner for plain unittest binary invocations.
- Run threadsan tests with default config in parallel instead of starting the test runner sequentially.
  - See catch2 [CLI examples](https://github.com/catchorg/Catch2/blob/devel/docs/command-line.md#examples) for the `[group1],[group2],some.test` syntax logic.
- Override `$HOME` in unittest to isolate parallel test runs.
   - Fixes an extension locking issue [caused](https://github.com/duckdb/duckdb/pull/21212#issuecomment-4010777023) by parallel tests. @Mytherin FYI.
- Use a small namespace runner (only for `duckdb/duckdb`, not forks) for the `prepare` job.
  - I have another PR where I'm integrating `format-check` into `prepare`.
  - For now, this change means we can start `prepare` even under CI load caused by many parallel v1.5 CI runs.

### Minor changes

- Simplify some CI job unittest invocations with existing make targets.
- Print elapsed test time: `all tests passed in ...s` at the end.
- Print the test config dynamically instead of hardcoding keys/values.

PR #22419, opened 2026-05-01
Fix typo in env var for linux-relassert CI job

smvv:
### Relassert files in artifact

```
relassert artifact includes:
  duckdb
  repository/51c97af758/linux_amd64/core_functions.duckdb_extension
  repository/51c97af758/linux_amd64/parquet.duckdb_extension
  src/libduckdb.so
  test/extension/loadable_extension_demo.duckdb_extension
  test/extension/loadable_extension_optimizer_demo.duckdb_extension
  test/unittest
```

### Before (1.1 GiB, compressed)

```
relassert artifact size:
605M	build/relassert-artifact/relassert/src
895M	build/relassert-artifact/relassert/test/extension
970M	build/relassert-artifact/relassert/test
1.1G	build/relassert-artifact/relassert/repository
1.1G	build/relassert-artifact/relassert/repository/51c97af758
1.1G	build/relassert-artifact/relassert/repository/51c97af758/linux_amd64
3.2G	build/relassert-artifact/relassert
+ tar -C build/relassert-artifact -cf - relassert | pigz -4
+ ls -lh build/relassert-artifact.tar.gz
-rw-r--r-- 1 runner runner 1.1G May  1 08:08 build/relassert-artifact.tar.gz
```

### After (471 MiB, compressed)

```
relassert artifact size:
4.1M	build/relassert-artifact/relassert/test/extension
80M	build/relassert-artifact/relassert/test
148M	build/relassert-artifact/relassert/repository
148M	build/relassert-artifact/relassert/repository/92c40caf5a
148M	build/relassert-artifact/relassert/repository/92c40caf5a/linux_amd64
605M	build/relassert-artifact/relassert/src
1.4G	build/relassert-artifact/relassert
+ tar -C build/relassert-artifact -cf - relassert | pigz -4
+ ls -lh build/relassert-artifact.tar.gz
-rw-r--r-- 1 runner runner 471M May  1 08:32 build/relassert-artifact.tar.gz
```

PR #22854, opened 2026-05-24
Test runner: allow to run without changing HOME

staticlibs:
One of the improvements added in #21211 is changing the `HOME` directory for the `unittest` process before tests are run. This improvement has an undesired side effect with extensions:

 - extension uses `require <extension_name>` in SQLlogic file
 - for known extensions this `require` is a no-op, so they are auto-installed into `duckdb_unittest_tempdir/<pid>/.duckdb/extensions/<version>/<platform>/<extension_name>.duckdb_extension'` and loaded from there
 - when running under debugger, the extension shared library is being loaded from a relative path with `HOME` rewritten - this confuses GDB that cannot find the symbols (despite them residing inside the file), like this:

```
Loaded 'duckdb_unittest_tempdir/93433/.duckdb/extensions/80554e9433/linux_amd64/mysql_scanner.duckdb_extension'. Cannot find or open the symbol file.
```

Note, with the error above I double-checked that the extension library installed under `duckdb_unittest_tempdir/<pid>/.duckdb/extensions/` is actually the correct version (with debuginfo included inside) taken from the specified `LOCAL_EXTENSION_REPO`.

The `HOME` is changed unconditionally, so it makes it not possible to debug unmodified SQLlogic test files.

This PR adds new switch `--keep-home` (off by default), that, when activated, skips the `HOME` switching logic.

This flag allows straightforward debugging setup:

 - add symlinks to required local-built extensions to `~/.duckdb/extensions/<version>/<platform>/`
 - use debugger config like this (VSCode example):

```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "(gdb) Launch",
      "type": "cppdbg",
      "request": "launch",
      "program": "${workspaceFolder}/build/debug/test/unittest",
      "args": [
        "--keep-home",
        "${workspaceFolder}/test/sql/attach_many_nulls.test"
      ],
      "cwd": "${fileDirname}",
      "externalConsole": false,
      "MIMode": "gdb",
      "setupCommands": [
        {
          "description": "Enable pretty-printing for gdb",
          "text": "-enable-pretty-printing",
          "ignoreFailures": true
        }
      ]
    }
  ]
}
```
 - 1.5GB extension lib is loaded through symlink without copying it into temporary `HOME` on every `unittest` run

Ref: duckdblabs/duckdb-internal#7731

### DIFF
--- a/test/unittest.cpp
+++ b/test/unittest.cpp
@@ -17,7 +17,7 @@ int main(int argc_in, char *argv[]) {
 
 	auto &test_config = TestConfiguration::Get();
 	test_config.Initialize();
-	bool isolate_tests = false;
+	bool keep_home = false;
 
 	idx_t argc = NumericCast<idx_t>(argc_in);
 	int new_argc = 0;
@@ -37,8 +37,8 @@ int main(int argc_in, char *argv[]) {
 			SetTestDirectory(test_dir);
 		} else if (argument == "--require") {
 			AddRequire(string(argv[++i]));
-		} else if (argument == "--isolate-tests") {
-			isolate_tests = true;
+		} else if (argument == "--keep-home") {
+			keep_home = true;
 		} else if (!test_config.ParseArgument(argument, argc, argv, i)) {
 			new_argv[new_argc] = argv[i];
 			new_argc++;
@@ -58,7 +58,7 @@ int main(int argc_in, char *argv[]) {
 	}
 
 	// Override the home dir so the .duckdb dir is isolated per test process.
-	if (isolate_tests) {
+	if (!keep_home) {
 #ifdef DUCKDB_WINDOWS
 		if (_putenv_s("USERPROFILE", dir.c_str()) != 0) {
 			fprintf(stderr, "Failed to set USERPROFILE environment variable\n");
--- a/test/unittest.cpp
+++ b/test/unittest.cpp
@@ -17,6 +17,7 @@ int main(int argc_in, char *argv[]) {
 
 	auto &test_config = TestConfiguration::Get();
 	test_config.Initialize();
+	bool keep_home = false;
 
 	idx_t argc = NumericCast<idx_t>(argc_in);
 	int new_argc = 0;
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks - given that this isolation is mostly required for parallel testing I wonder if this should be opt-out instead of opt-in (i.e. `--keep-home` would be the default) - then the parallel test runner can just override the home directory. @smvv what do you think?

smvv:
> Thanks - given that this isolation is mostly required for parallel testing I wonder if this should be opt-out instead of opt-in (i.e. `--keep-home` would be the default) - then the parallel test runner can just override the home directory. @smvv what do you think?

Indeed, I would let the test runner script enable `--isolate-tests` to let the unittest binary change the home dir to the test dir (and later can also enable more code needed for parallel test runs).

For a single unittest binary process, most "isolation/parallel testing logic" can be turned off (if that helps the debugger). So, the test runner `scripts/ci/run_tests.py` can set `--isolate-tests` and the unittest binary enables those flags.

This means we don't need additional configuration for debugging a test case. However, it could mean that if a file exists, the second test case or run could behave differently (hitting different code paths).

Just to be sure: is the binary extension in the test dir wiped before gdb can read the file causing the error? See:
```c++
	if (DeleteTestPath()) {
		TestDeleteDirectory(dir);
	}
```

Related to binary file size; a relassert build (in CI) uses dynamic linking with `EXTENSION_STATIC_BUILD: 0` because the static extension files are [huge](https://github.com/duckdb/duckdb/pull/22419) when sanitizers are enabled.

For context; I'm working on moving most environment variables from the yaml file to the makefile, so local and CI build commands are identical, and we can avoid copy-pasting from the yaml file to reproduce issues locally.

staticlibs:
@smvv 

Continuing from Discord, I've renamed the flag to `--isolate-tests` and made it disabled by default. Please let me know WDYT!

smvv:
I misunderstood the problem. I think that `--keep-home` is useful to add (so let's merge the original idea!), but we should not have a toggle to enable test isolation because test cases should be isolated by default.

Otherwise, we increase the problem of "works on my machine, but not in CI" when running the unittest binary directly because it re-uses previous duckdb/test state.

I need more time to understand/context on why the debugger fails to locate the extension binary file with symbols. That's not clear for me from the problem and by only adding `--keep-home` we keep the problem when attaching a debugger to a test failure/crash when using the multi-process test runner.

smvv (approved):
For now, the original idea of adding `--keep-home` is good enough and helps with Alex' use case.

staticlibs:
Reverted it back to `--keep-home`.

PR #22857, opened 2026-05-24
Evict dead ExternalFileCache opportunisticaly

### LINKED ISSUE #22730: Unbounded memory growth in ExternalFileCache::cached_files
### What happens?

While investigating RSS leak/growth for [ducklake#1136](https://github.com/duckdb/ducklake/issues/1136), I think I traced it to `ExternalFileCache::cached_files`, but I'd appreciate a sanity check from someone who knows this code better.

### What I'm seeing

`cached_files` (`external_file_cache.hpp:96`) is an `unordered_map<string, unique_ptr<CachedFile>>`. As far as I can tell, `GetOrCreateCachedFile` inserts on every `OpenFile` call, and the only path that removes entries is `SetEnabled(false) → cached_files.clear()`.
`OpenFile` calls `GetOrCreateCachedFile` before any `IsEnabled()` check, so `SET enable_external_file_cache=false` doesn't seem to prevent insertion, it only gates range reuse later inside `CachingFileHandle`. 

So it feels like a memory leak to me. 

### To Reproduce

[ducklake#1136](https://github.com/duckdb/ducklake/issues/1136) has a reproducer. But logically the memory leak seems pretty legit.

### OS:

macOS and Linux 

### DuckDB Version:

1.5.1

### DuckDB Client:

Java, Python or CLI

### Hardware:

_No response_

### Full Name:

Zhiming Guo

### Affiliation:

O11yLite

- [x] Yes, I have

- [x] Yes, I have

Yes

zhming0:
## Summary

Fix #22730.

Fixes unbounded growth of `ExternalFileCache::cached_files`. The map only ever shrinks on `SET enable_external_file_cache = false`, so long-running processes that open many distinct paths accumulate `CachedFile` skeletons indefinitely, even after the `BufferManager` has evicted all their cached content.

## Change

Opportunistically prune stale entries in `GetOrCreateCachedFile` on the miss path. An entry is stale when:
1. `ref_count == 0` (no live `CachingFileHandle`), **and**
2. All its `CacheBlock`s report `block_handle->GetMemory().IsUnloaded()` (BufferManager already released the buffers).
The cost is amortized over inserts. 

## Alternatives I've considered

I looked at a few other places to put the eviction and settled on this one:

- **Prune in `~CachingFileHandle()`** — kills cross-query cache reuse: the moment the last handle closes, the entry is gone, so the next open of the same path re-reads from source. Defeats the cache.
- **BufferManager → EFC callback on block eviction** — semantically the right place, but `BufferManager` is intentionally agnostic of `ExternalFileCache` and adding a notification hook is a pretty big cross-cutting change, not sure if worth it.
- **Bounded LRU + new `external_file_cache_max_entries` setting** — the new knob's semantic is quite awkward, does not feel right. 
- **Background sweeper / scheduled task** — new lifecycle infrastructure, it might be the right option but bigger change, happy to switch to this one if people feel it's worth it.

The chosen approach has the smallest blast radius: ~60 lines across 3 files, no public API change. 

## Notes

- `ref_count` is incremented inside `GetOrCreateCachedFile` (under the cache lock) rather than in the `CachingFileHandle` constructor, to close a TOCTOU window where a concurrent prune could evict the entry between lookup and pin. Decrement remains in the destructor.
- Soft guarantee: entries with live handles are never pruned, even if their content is unloaded. This is correct — the handle reader may re-populate blocks at any moment.

### DIFF
--- a/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp
+++ b/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp
@@ -58,6 +58,10 @@ public:
 		string version_tag DUCKDB_GUARDED_BY(meta_lock);
 		bool can_seek DUCKDB_GUARDED_BY(meta_lock) = false;
 		bool on_disk_file DUCKDB_GUARDED_BY(meta_lock) = false;
+
+		//! Number of live CachingFileHandles referencing this CachedFile.
+		//! Used by ExternalFileCache to skip pruning of in-use entries.
+		atomic<idx_t> ref_count {0};
 	};
 
 public:
@@ -88,14 +92,21 @@ private:
 	void ReindexCachedFileCore(CachedFile &cached_file, idx_t file_size, idx_t old_block_size, idx_t new_block_size)
 	    DUCKDB_REQUIRES(cached_file.map_lock);
 
+	//! Erase entries that have no live CachingFileHandle and whose cached content
+	//! has already been released by the BufferManager.
+	void PruneStaleEntries() DUCKDB_REQUIRES(lock);
+
+	//! Return true if `file` has no live handle and no resident cached content.
+	static bool IsStale(const CachedFile &file);
+
 	//! The BufferManager used to cache files
 	BufferManager &buffer_manager;
 	//! Whether or not file caching is enabled
 	atomic<bool> enable;
+	//! Lock for accessing cached_files.
+	mutable annotated_mutex lock;
 	//! Mapping from file path to cached file with cached blocks
-	unordered_map<string, unique_ptr<CachedFile>> cached_files;
-	//! Lock for accessing the cached files
-	mutable mutex lock;
+	unordered_map<string, unique_ptr<CachedFile>> cached_files DUCKDB_GUARDED_BY(lock);
 };
 
 } // namespace duckdb
--- a/src/storage/external_file_cache/external_file_cache.cpp
+++ b/src/storage/external_file_cache/external_file_cache.cpp
@@ -200,15 +200,49 @@ bool ExternalFileCache::IsEnabled() const {
 }
 
 void ExternalFileCache::SetEnabled(bool enable_p) {
-	lock_guard<mutex> guard(lock);
+	annotated_lock_guard<annotated_mutex> guard(lock);
 	enable = enable_p;
 	if (!enable) {
 		cached_files.clear();
 	}
 }
 
+bool ExternalFileCache::IsStale(const CachedFile &file) {
+	if (file.ref_count.load() != 0) {
+		return false;
+	}
+	// The BufferManager owns CacheBlock buffers and is not aware of the file cache.
+	// An entry becomes stale when all its blocks have been evicted from the buffer
+	// pool: the skeleton CachedFile remains but holds no resident content.
+	annotated_lock_guard<annotated_mutex> map_guard(file.map_lock);
+	if (file.blocks.empty()) {
+		return true;
+	}
+	for (const auto &block_entry : file.blocks) {
+		const auto &block = *block_entry.second;
+		annotated_lock_guard<annotated_mutex> block_guard(block.mtx);
+		if (block.state != CacheBlockState::LOADED || !block.block_handle) {
+			continue;
+		}
+		if (!block.block_handle->GetMemory().IsUnloaded()) {
+			return false; // still holds resident content
+		}
+	}
+	return true;
+}
+
+void ExternalFileCache::PruneStaleEntries() {
+	for (auto it = cached_files.begin(); it != cached_files.end();) {
+		if (IsStale(*it->second)) {
+			it = cached_files.erase(it);
+		} else {
+			++it;
+		}
+	}
+}
+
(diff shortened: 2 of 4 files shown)


dentiny (commented):
I like you solution better, but we should add unit test to verify entry counts in the map?

dentiny (commented):
nvm

src/include/duckdb/storage/external_file_cache/external_file_cache.hpp
```diff
@@ -88,13 +92,20 @@ class ExternalFileCache {
 	void ReindexCachedFileCore(CachedFile &cached_file, idx_t file_size, idx_t old_block_size, idx_t new_block_size)
 	    DUCKDB_REQUIRES(cached_file.map_lock);
 
+	//! Erase entries that have no live CachingFileHandle and whose cached content
+	//! has already been released by the BufferManager. Called with `lock` held.
```
dentiny: Could you please do me a favor to use annotated mutex here? Think it should be an easy change and more clear then the comment. The following works on main.
```diff
diff --git a/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp b/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp
index 59354525f5..d214bced1c 100644
--- a/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp
+++ b/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp
@@ -92,10 +92,10 @@ private:
        BufferManager &buffer_manager;
        //! Whether or not file caching is enabled
        atomic<bool> enable;
-       //! Mapping from file path to cached file with cached blocks
-       unordered_map<string, unique_ptr<CachedFile>> cached_files;
        //! Lock for accessing the cached files
-       mutable mutex lock;
+       mutable annotated_mutex lock;
+       //! Mapping from file path to cached file with cached blocks
+       unordered_map<string, unique_ptr<CachedFile>> cached_files DUCKDB_GUARDED_BY(lock);
 };
 
 } // namespace duckdb
diff --git a/src/storage/external_file_cache/external_file_cache.cpp b/src/storage/external_file_cache/external_file_cache.cpp
index f93a09ba97..fb55e3045f 100644
--- a/src/storage/external_file_cache/external_file_cache.cpp
+++ b/src/storage/external_file_cache/external_file_cache.cpp
@@ -200,7 +200,7 @@ bool ExternalFileCache::IsEnabled() const {
 }
 
 void ExternalFileCache::SetEnabled(bool enable_p) {
-       lock_guard<mutex> guard(lock);
+       annotated_lock_guard<annotated_mutex> guard(lock);
        enable = enable_p;
        if (!enable) {
                cached_files.clear();
@@ -208,7 +208,7 @@ void ExternalFileCache::SetEnabled(bool enable_p) {
 }
 
 vector<CachedFileInformation> ExternalFileCache::GetCachedFileInformation() const {
-       unique_lock<mutex> files_guard(lock);
+       annotated_unique_lock<annotated_mutex> files_guard(lock);
        vector<CachedFileInformation> result;
        for (const auto &file : cached_files) {
                annotated_lock_guard<annotated_mutex> map_guard(file.second->map_lock);
@@ -243,7 +243,7 @@ BufferManager &ExternalFileCache::GetBufferManager() const {
 }
 
 ExternalFileCache::CachedFile &ExternalFileCache::GetOrCreateCachedFile(const string &path) {
-       lock_guard<mutex> guard(lock);
+       annotated_lock_guard<annotated_mutex> guard(lock);
        auto &entry = cached_files[path];
        if (!entry) {
                entry = make_uniq<CachedFile>(path);
(END)
```

zhming0: Fancy stuff! Done ✅

zhming0:
Added a test case ✅ .

dentiny (approved):

Mytherin:
Thanks - looks like there's some merge conflicts, could you have a look?

src/storage/external_file_cache/external_file_cache.cpp
```diff
@@ -243,12 +277,20 @@ BufferManager &ExternalFileCache::GetBufferManager() const {
...
+	annotated_lock_guard<annotated_mutex> guard(lock);
+	auto existing = cached_files.find(path);
+	if (existing != cached_files.end()) {
+		// Pin the entry before returning so a concurrent prune cannot evict it
+		// between this call and CachingFileHandle's constructor.
+		++existing->second->ref_count;
+		return *existing->second;
 	}
-	return *entry;
+	// Opportunistic cleanup: drop entries whose content has been released by the
```
dentiny, line 288: I think the downside of current the pruning logic is, every cache entry miss leads to full map scan, which we don't have an explicit upper bound (it could be the case where all cache blocks are referenced so we don't evict).
Checked with @zhming0 offline, I will do some small modification to this PR, and plan to put a PR these days.

zhming0: 👍 I will close this PR for now. Looking forward to your PR 🙏

dentiny:
Hi @zhming0 I cannot be more grateful if you could provide any feedback on this PR: https://github.com/duckdb/duckdb/pull/23052, thank you!

Issue #22827, opened 2026-05-22
INTERNAL Error: Attempting to dereference an optional pointer that is not set


DerZc:
### What happens?

Hi,

The following test case triggers an internal error.

First, create a file /tmp/scalar_inference.son with 

```
{"a":{"k1":[1]},"b":{"u":1}}
```

Then execute the following sql:

```
DESCRIBE FROM read_json(
    '/tmp/scalar_inference.json',
    map_inference_threshold=1
);
```

This is the output:

```
INTERNAL Error:
Attempting to dereference an optional pointer that is not set

LINE 1: DESCRIBE FROM read_json(
                      ^

Stack Trace:

./build/release/duckdb(+0x92a63d) [0x6408f42d963d]
./build/release/duckdb(+0x92a6b8) [0x6408f42d96b8]
./build/release/duckdb(+0x92e695) [0x6408f42dd695]
./build/release/duckdb(+0x184022) [0x6408f3b33022]
./build/release/duckdb(+0x1e090d3) [0x6408f57b80d3]
./build/release/duckdb(+0x1e09825) [0x6408f57b8825]
./build/release/duckdb(+0x1e0b9f8) [0x6408f57ba9f8]
./build/release/duckdb(+0x1e0bfd4) [0x6408f57bafd4]
./build/release/duckdb(+0x1e4f27f) [0x6408f57fe27f]
./build/release/duckdb(+0x1e5e3e5) [0x6408f580d3e5]
./build/release/duckdb(+0x1dc39db) [0x6408f57729db]
./build/release/duckdb(+0x1e4d3d3) [0x6408f57fc3d3]
./build/release/duckdb(+0x1073e95) [0x6408f4a22e95]
./build/release/duckdb(+0x1076363) [0x6408f4a25363]
./build/release/duckdb(+0x10e7000) [0x6408f4a96000]
./build/release/duckdb(+0x1013daa) [0x6408f49c2daa]
./build/release/duckdb(+0x1016d67) [0x6408f49c5d67]
./build/release/duckdb(+0x10c6461) [0x6408f4a75461]
./build/release/duckdb(+0x106c8e3) [0x6408f4a1b8e3]
./build/release/duckdb(+0x106d604) [0x6408f4a1c604]
./build/release/duckdb(+0x10e7060) [0x6408f4a96060]
./build/release/duckdb(+0x1013daa) [0x6408f49c2daa]
./build/release/duckdb(+0x1016d67) [0x6408f49c5d67]
[...]

PR #22862, opened 2026-05-25
Fix: JSON add list type check

DinosL:
List type check in `CalculateTypeSimilarity` to prevent internal exception with `map_inference_threshold=1`

fixes:https://github.com/duckdb/duckdb/issues/22827


test/sql/json/issues/map_inference_threshold_internal_exception.test
```diff
@@ -1,8 +1,11 @@
-# name: test/sql/json/issues/internal_issue3813.test
+# name: test/sql/json/issues/map_inference_threshold_internal_exception.test
```
protomn, line 1: Would suggest following the naming convention?? issue22862.test

protomn (approved):

Mytherin:
Thanks!

PR #22865, opened 2026-05-25
Add CSV zstd compression level support and regression test

### LINKED ISSUE #22820: Cannot set compression level when exporting to .csv.zst.
### What happens?

There is no way to set compression level for .csv.zst files.

### To Reproduce

For example:
```sql
COPY(SELECT 1)
TO 'result.csv.zst' WITH (compression zstd, compression_level 19)
```

Expected: file created with high compression level
Got: `Error: Not implemented Error: Unrecognized option "compression_level" for csv`

### OS:

Windows

### DuckDB Version:

1.5.3

### DuckDB Client:

JDBC

### Hardware:

_No response_

### Full Name:

Andrey Sotnikov

### Affiliation:

Rostelecom jsc

- [x] Yes, I have

- [x] Yes, I have

Yes

M1llerF:
Fixes #22820 

This PR adds support for `COMPRESSION_LEVEL` when writing CSV files with `COMPRESSION zstd`, e.g.:
```sql
COPY (SELECT 1)
TO 'result.csv.zst'
WITH (
    COMPRESSION zstd,
    COMPRESSION_LEVEL 19
);
```

Previously, CSV COPY rejected this option during binding with:
`Unrecognized option "compression_level" for csv`

## Changes
- Add a `FileCompressionOptions` carrier so file open flags can pass both the compression type and optional compression level through the virtual file system layer.
- Thread compression options through `VirtualFileSystem::OpenFileExtended`, compressed file-system wrappers, and `BufferedFileWriter`.
- Teach the zstd file-system wrapper to apply the requested compression level via `ZSTD_CCtx_setParameter(..., ZSTD_c_compressionLevel, ...)`.
- Register and parse CSV `COMPRESSION_LEVEL` as a write-only COPY option.
- Validate CSV compression levels using zstd min/max bounds.
- Reject `COMPRESSION_LEVEL` for non-zstd CSV compression modes.
- Pass the parsed CSV compression level from `CSVReaderOptions` into `CSVWriter` and file open flags.
- Add a regression test covering:
  - zstd CSV export with `COMPRESSION_LEVEL`
  - rejection for gzip + compression level
  - rejection for out-of-range zstd levels
  - actual size differences between zstd level 1 and 19 output

## Testing
```bash
build/debug/test/unittest
test/sql/copy/csv/csv_write_compression_level.test
```

### DIFF
--- a/extension/parquet/include/zstd_file_system.hpp
+++ b/extension/parquet/include/zstd_file_system.hpp
@@ -22,6 +22,9 @@ namespace duckdb {
 
 class ZStdFileSystem : public CompressedFileSystem {
 public:
+	ZStdFileSystem();
+	explicit ZStdFileSystem(int64_t compression_level);
+
 	unique_ptr<FileHandle> OpenCompressedFile(QueryContext context, unique_ptr<FileHandle> handle, bool write,
 	                                          const FileCompressionOptions &compression_options) override;
 
@@ -36,6 +39,9 @@ public:
 	static int64_t DefaultCompressionLevel();
 	static int64_t MinimumCompressionLevel();
 	static int64_t MaximumCompressionLevel();
+
+private:
+	int64_t compression_level;
 };
 
 } // namespace duckdb
--- a/extension/parquet/include/zstd_file_system.hpp
+++ b/extension/parquet/include/zstd_file_system.hpp
@@ -22,9 +22,6 @@ namespace duckdb {
 
 class ZStdFileSystem : public CompressedFileSystem {
 public:
-	ZStdFileSystem();
-	explicit ZStdFileSystem(int64_t compression_level);
-
 	unique_ptr<FileHandle> OpenCompressedFile(QueryContext context, unique_ptr<FileHandle> handle, bool write,
 	                                          const FileCompressionOptions &compression_options) override;
 
@@ -32,16 +29,13 @@ public:
 		return "ZStdFileSystem";
 	}
 
-	unique_ptr<StreamWrapper> CreateStream() override;
+	unique_ptr<StreamWrapper> CreateStream(const FileCompressionOptions &compression_options) override;
 	idx_t InBufferSize() override;
 	idx_t OutBufferSize() override;
 
 	static int64_t DefaultCompressionLevel();
 	static int64_t MinimumCompressionLevel();
 	static int64_t MaximumCompressionLevel();
-
-private:
-	int64_t compression_level;
 };
 
 } // namespace duckdb
--- a/extension/parquet/include/zstd_file_system.hpp
+++ b/extension/parquet/include/zstd_file_system.hpp
@@ -22,7 +22,7 @@ namespace duckdb {
 
 class ZStdFileSystem : public CompressedFileSystem {
 public:
-	unique_ptr<FileHandle> OpenCompressedFile(QueryContext context, unique_ptr<FileHandle> handle, bool write,
+	unique_ptr<FileHandle> OpenCompressedFile(QueryContext context, unique_ptr<FileHandle> handle,
 	                                          const FileCompressionOptions &compression_options) override;
 
 	std::string GetName() const override {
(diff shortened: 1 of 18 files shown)


extension/parquet/include/zstd_file_system.hpp
```diff
@@ -35,6 +39,9 @@ class ZStdFileSystem : public CompressedFileSystem {
 	static int64_t DefaultCompressionLevel();
 	static int64_t MinimumCompressionLevel();
 	static int64_t MaximumCompressionLevel();
+
+private:
+	int64_t compression_level;
```
Mytherin: The FileSystem is the wrong place to attach this - that's a global database-wide setting, when we need to specify this per-file.

src/include/duckdb/common/file_open_flags.hpp
```diff
@@ -16,6 +16,20 @@ namespace duckdb {
...
+	constexpr FileCompressionOptions() = default;
+	constexpr FileCompressionOptions(FileCompressionType type) // NOLINT: allow implicit conversion
+	    : type(type) {
+	}
+	constexpr FileCompressionOptions(FileCompressionType type, int64_t compression_level)
+	    : type(type), compression_level(compression_level), has_compression_level(true) {
+	}
+
+	FileCompressionType type = FileCompressionType::UNCOMPRESSED;
+	int64_t compression_level = 0;
```
Mytherin: `optional<int64_t>`

extension/parquet/include/zstd_file_system.hpp
```diff
@@ -22,7 +22,11 @@ namespace duckdb {
 
 class ZStdFileSystem : public CompressedFileSystem {
 public:
-	unique_ptr<FileHandle> OpenCompressedFile(QueryContext context, unique_ptr<FileHandle> handle, bool write) override;
+	ZStdFileSystem();
+	explicit ZStdFileSystem(int64_t compression_level);
+
+	unique_ptr<FileHandle> OpenCompressedFile(QueryContext context, unique_ptr<FileHandle> handle, bool write,
```
Mytherin: If we are going to change this constructor - can we also move the `bool write` to `FileCompressionOptions`?

src/execution/operator/csv_scanner/util/csv_reader_options.cpp
```diff
@@ -407,6 +408,15 @@ void CSVReaderOptions::SetWriteOption(const string &loption, const Value &value)
 		prefix = ParseString(value, loption);
 	} else if (loption == "suffix") {
 		suffix = ParseString(value, loption);
+	} else if (loption == "compression_level") {
+		auto val = ParseInteger(value, loption);
+		auto min_level = static_cast<int64_t>(duckdb_zstd::ZSTD_minCLevel());
```
Mytherin, line 413: This should use `ZStdFileSystem::MinimumCompressionLevel` etc instead of using the zstd internals directly

Mytherin (commented):
Thanks - looks good, I've left some comments

src/execution/operator/csv_scanner/util/csv_reader_options.cpp
M1llerF: I’ve routed this through `ZStdFileSystem` in the latest commit. I noticed that `csv_reader_options.cpp` is core code, while `ZStdFileSystem` lives in the parquet extension.
I'm wondering if you'd prefer me to move the compression-level validation helpers into a shared file-compression utility to avoid introducing a dependency from core into the parquet extension?

Mytherin: Ah yeah. I think the zstd file system should actually move into `core`. The reason it lives in the parquet extension is legacy (we didn't used to have `zstd` in core). However, for this PR maybe revert back to using `duckdb_zstd` directly here.

M1llerF: Got it, that makes sense. I’ve reverted this back to using `duckdb_zstd` for this PR

M1llerF:
Rebased onto current `main` and addressed the prior review feedback. I verified locally with the CSV compression-level test and the existing Parquet compression-level test.
Would appreciate another look when you have a chance

M1llerF:
Closing this to free up #22820 for someone else to take over, as I don’t currently have the bandwidth to continue maintaining the PR. The current branch incorporates the review feedback above and may still be useful as a starting point. Thanks for the initial review

PR #22871, opened 2026-05-26
Fetch co-located rows in batch

dentiny:
Closes https://github.com/duckdblabs/duckdb-internal/issues/6963

In the current implementation, `RowGroupCollection::Fetch` check row visibility and fetch in the row granularity, which could be fetched in batch when requested rows co-locate.

In the added benchmark, (1) all rows deleted are co-located together; (2) rows are fetched in batch.

Before (on main branch)
```sh
hjiang@hjiangs-MacBook-Pro$ build/release/benchmark/benchmark_runner benchmark/micro/index/delete/delete_colocated_fetch.benchmark
name run timing
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 1 0.842937
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 2 0.867070
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 3 0.861112
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 4 0.866807
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 5 0.844375
```
After (current branch)
```sh
hjiang@hjiangs-MacBook-Pro$ build/release/benchmark/benchmark_runner benchmark/micro/index/delete/delete_colocated_fetch.benchmark
name run timing
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 1 0.788078
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 2 0.794573
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 3 0.785105
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 4 0.813504
benchmark/micro/index/delete/delete_colocated_fetch.benchmark 5 0.794658
```


src/include/duckdb/storage/table/standard_column_data.hpp
```diff
@@ -43,6 +43,9 @@ class StandardColumnData : public ColumnData {
 	idx_t Fetch(ColumnScanState &state, row_t row_id, Vector &result) override;
 	void FetchRow(TransactionData transaction, ColumnFetchState &state, const StorageIndex &storage_index, row_t row_id,
 	              Vector &result, idx_t result_idx) override;
+	void FetchRows(TransactionData transaction, ColumnFetchState &state, const StorageIndex &storage_index,
```
Mytherin, line 44: Can we remove `FetchRow`? I would prefer not to have two variants of the same method around.

src/include/duckdb/storage/table/row_group.hpp
```diff
@@ -160,9 +160,17 @@ class RowGroup : public SegmentBase<RowGroup> {
 
 	//! For a specific row, returns true if it should be used for the transaction and false otherwise.
 	bool Fetch(TransactionData transaction, idx_t row);
+	//! Bulk visibility check. For each offset in [0, count), writes the input index into `visible_sel` if that row is
+	//! visible to the transaction. Returns the number of visible rows.
+	idx_t Fetch(TransactionData transaction, const idx_t *offsets, idx_t count, SelectionVector &visible_sel);
```
Mytherin, line 163: Same here - let's try to remove `Fetch` and only have this method

Mytherin (commented):

src/include/duckdb/storage/table/standard_column_data.hpp
dentiny: for sure, that makes sense, updated!

dentiny:
Hi @Mytherin , do you mind taking another look when you have some time? Thanks!

src/storage/table/column_data.cpp
```diff
@@ -689,6 +689,52 @@ void ColumnData::FetchRow(TransactionData transaction, ColumnFetchState &state,
 	FetchUpdateRow(transaction, row_id, result, result_idx);
 }
 
+void ColumnData::FetchRows(TransactionData transaction, ColumnFetchState &state, const StorageIndex &storage_index,
+                           const idx_t *offsets, const SelectionVector &sel, idx_t fetch_count, Vector &result,
+                           idx_t result_offset) {
+	// Default safe fallback: per-row FetchRow.
+	for (idx_t i = 0; i < fetch_count; i++) {
+		const idx_t offset = offsets[sel.get_index(i)];
+		FetchRow(transaction, state, storage_index, NumericCast<row_t>(offset), result, result_offset + i);
```
Mytherin: I don't quite follow what the benefit is in keeping the old `FetchRow` around - can we not unify this further?

src/storage/table/column_data.cpp
```diff
@@ -689,6 +689,52 @@ void ColumnData::FetchRow(TransactionData transaction, ColumnFetchState &state,
...
+	if (fetch_count == 0) {
+		return;
+	}
+
+	const sel_t *sel_data = sel.data();
+	optional_ptr<SegmentNode<ColumnSegment>> current_segment;
+	idx_t segment_start = 0;
+	idx_t segment_end = 0;
+	for (idx_t idx = 0; idx < fetch_count; idx++) {
+		const idx_t offset = offsets[sel_data ? sel_data[idx] : idx];
```
Mytherin: Can we just use `sel.get_index(idx)` here?

src/storage/table/column_data.cpp
```diff
@@ -689,6 +689,52 @@ void ColumnData::FetchRow(TransactionData transaction, ColumnFetchState &state,
...
+		}
+		const idx_t index_in_segment = offset - segment_start;
+		current_segment->GetNode().FetchRow(state, NumericCast<row_t>(index_in_segment), result, result_offset + idx);
+	}
+	{
+		const lock_guard<mutex> update_guard(update_lock);
+		if (updates) {
+			for (idx_t idx = 0; idx < fetch_count; idx++) {
+				const idx_t offset = offsets[sel_data ? sel_data[idx] : idx];
+				updates->FetchRow(transaction, NumericCast<idx_t>(offset), result, result_offset + idx);
```
Mytherin: Can we also do this in a batch?

src/storage/table/geo_column_data.cpp
```diff
@@ -142,16 +142,20 @@ idx_t GeoColumnData::Fetch(ColumnScanState &state, row_t row_id, Vector &result)
 
 void GeoColumnData::FetchRow(TransactionData transaction, ColumnFetchState &state, const StorageIndex &storage_index,
                              row_t row_id, Vector &result, idx_t result_idx) {
+	const idx_t offset = NumericCast<idx_t>(row_id);
+	const SelectionVector identity_sel;
+
 	// Not a shredded column, so just emit the binary format immediately
 	if (storage_type == GeometryStorageType::WKB) {
-		return base_column->FetchRow(transaction, state, storage_index, row_id, result, result_idx);
+		return base_column->FetchRows(transaction, state, storage_index, &offset, identity_sel, /*count=*/1, result,
```
Mytherin: We don't need `identity_sel` - we can pass in `FlatVector::IncrementalSelectionVector()` instead

src/storage/table/row_group_collection.cpp
```diff
@@ -432,32 +432,79 @@ RowGroupIterationHelper RowGroupCollection::Chunks(DuckTransaction &transaction,
...
+				break;
+			}
+			offsets[pos - run_start] = rid - row_start;
+			pos++;
+		}
+		const idx_t run_count = pos - run_start;
+
+		// 3. bulk visibility check for the whole run.
+		idx_t visible_count = 0;
+		const SelectionVector *sel_for_fetch = nullptr;
```
Mytherin: nit: `const_reference<SelectionVector> sel_for_fetch`

Mytherin (commented):
Thanks for the updates - looks better, but I think it would be nicer if we could fully remove `FetchRow` instead of the current implementation where this is still half-present. I left some more comments:

src/storage/table/column_data.cpp
dentiny: Yes, updated in batch style.

src/storage/table/column_data.cpp
dentiny: Updated to batch mode `FetchRows`.

I didn't do that at beginning because column segment and compression function is still [per-row based](https://github.com/duckdb/duckdb/blob/c37a4f5b27480f0236bd3333218b713ca16ad6e5/src/storage/table/column_segment.cpp#L156-L161), which doesn't enjoy the benefit for like fewer locking, fewer transaction visibility checks, etc.

Mytherin:
Thanks!

PR #22875, opened 2026-05-26
Bump default storage to `v2.0.0`

lnkuiper:
Picked up where @taniabogatsch left off, and with @ccfelius storage/serialization version unification, it should now be possible to bump to a newer default. I've been CI debugging quite a bit, and it's looking pretty good now, hopefully everything passes in one go, but otherwise I'll be able to use the faster CI here to get this in!


src/storage/storage_info.cpp
```diff
@@ -11,7 +11,7 @@ constexpr idx_t Storage::MIN_BLOCK_ALLOC_SIZE;
 constexpr idx_t Storage::DEFAULT_BLOCK_HEADER_SIZE;
 constexpr uint64_t MainHeader::DEFAULT_ENCRYPTION_KEY_LENGTH;
 
-const uint64_t VERSION_NUMBER = 64;
+const uint64_t VERSION_NUMBER = 69;
```
ccfelius, line 14: small NIT and I should have done this in my previous PR, so it's more a note to self - but I think it would be more clean to also convert these into meaningful constants, if possible.

src/storage/storage_info.cpp
```diff
@@ -11,7 +11,7 @@ constexpr idx_t Storage::MIN_BLOCK_ALLOC_SIZE;
...
 const uint64_t VERSION_NUMBER_LOWER = 64;
 const uint64_t VERSION_NUMBER_UPPER = 69;
```
ccfelius, line 16: Both `VERSION_NUMBER_LOWER` and `VERSION_NUMBER_UPPER` are as far as I can see only used [here](https://github.com/duckdb/duckdb/blob/1910c8d68fd6bdd2f16d952fa96d45cf87b5823b/src/storage/single_file_block_manager.cpp#L224-L244). 

Since the default storage version is bumped, we will never touch this code block anymore which means that we should be able to remove these constants.

lnkuiper:
Thanks for the feedback! Since CI passed here and I don't want any new tests that potentially break to sneak in, I'll merge this now and open a new PR to address your feedback

PR #15607, opened 2025-01-08
Build/test/distribute linux_amd64_musl core extensions

carlopi:
Connected to https://github.com/duckdb/duckdb/pull/15429, adding the `_musl` targets to duckdb/duckdb and https://github.com/duckdb/extension-ci-tools/pull/122, adding the dockerfile for building them.

After this PR extensions are built, tested and distributed for an extra platform: `linux_amd64_musl`.

Core extensions that are not currently blocked by build-time issues are: `arrow`, `iceberg` and `mysql-scanner`. `jemalloc` had also some problem building, so I removed it for now, but that should not be visible to users.

There are at the moment 4 tests that fail on the musl build, I have not investigated yet why, possibly we are relying on some undefined behaviour in the libc (`musl` might do different choices from `glibc` in some areas).
Given the vast majory of other tests and extensions tests looks to be working, I would consider the platform as viable BUT experimental.
This also comes with the fact that this is basically untested in the wild (= by actual users).

Unconnected, but likely to generate conflicts, by I changed the names of the repositories to align to the new `duckdb-*` scheme.

PR #22877, opened 2026-05-26
Re-enable all explicitly disabled tests on musl

smvv:
There are no more tests left with `require notmusl`.

Tests were disabled in https://github.com/duckdb/duckdb/pull/15607.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/3827.


Mytherin:
Thanks!

PR #22878, opened 2026-05-26
Implement `CONNECT <attached_db_name>` and `DISCONNECT`, and opt-in in `quack`

carlopi:
Allows the following:
```sql
ATTACH 'quack:1.2.3.4:443' AS remote_db; --- with some secret, or via inline `TOKEN '1234'`
CONNECT remote_db;
--- now, all other queries will be sent AND executed on remote_db
DISCONNECT;
--- now back to plain in-process duckdb
```

Note that `CONNECT` and `DISCONNECT` statement are special, and executed only top-level (e.g. in the "local"/"in-process" duckdb).

Current rules are somewhat stricter that they should be, like:
```sql
ATTACH 'quack:1.2.3.4' AS remote_db_1; --- with some secret, or via inline `TOKEN '1234'`
ATTACH 'quack:9.8.7.6' AS remote_db_2; --- with some secret, or via inline `TOKEN '1234'`
CONNECT remote_db_1;
CONNECT remote_db_2; ---- forbidden, must do explicit `DISCONNECT`
DISCONNECT;
CONNECT remote_db_2;  ---- works!
```

Also, if a different connection where to issuing a `DETACH remote_resource`, then CONNECT would stop being functional, and that we detect, but still explicit `DISCONNECT` in the current connection is required, this to avoid implicit changes of state influenced by other connections.

Currently planned are the following improvements (for follow ups):
* allow to connect to `postgres`
* make CLI prompt aware, so that it shows something like: `memory @ quack:1.2.3.4 D` (to be finalized, for another PR)
* allow single query escape hatch: `CONNECT <remote_name> EXECUTE <SQL>` (here valid both if NOT connected, if connnected already to `<remote_name>` or if mid-connection to some other remote)
* make current state discoverable via SQL (like `SELECT current_connected_to()` to be called like `CONNECT LOCAL EXECUTE SELECT current_connected_to()`). This depends on previous point.
* implicit `ATTACH`, like `CONNECT 'quack:1.2.3.4';` to be both unnamed `ATTACH` and `CONNECT` to it.

Preparatory work was at https://github.com/duckdb/duckdb/pull/22732.

Now implements also comment from https://github.com/duckdb/duckdb/pull/22914#pullrequestreview-4379344865, where it uses infra from https://github.com/duckdb/duckdb/pull/22914 (`RemotePushdown`) but renaming as `RemoteExecute`.


src/include/duckdb/main/client_context.hpp
```diff
@@ -97,7 +98,23 @@ class ClientContext : public enable_shared_from_this<ClientContext> {
...
+	//! Routing target for pass-through SQL execution (CONNECT/DISCONNECT).
+	//! When is_bound is true and bound_database can be locked, non-control SQL is rewritten at
+	//! the chokepoint as `SELECT * FROM <fn>('cat', '<sql>')` and runs through the normal pipeline.
+	weak_ptr<AttachedDatabase> bound_database;
+	bool is_bound = false;
+
 public:
+	//! Bind this client to a remote-style AttachedDatabase. Subsequent non-control SQL routes via
+	//! Catalog::GetConnectFunctionName. Use UnbindCatalog() to revert to LOCAL.
+	DUCKDB_API void BindToCatalog(const shared_ptr<AttachedDatabase> &target);
```
Mytherin: `Bind / Unbind` is different wording than `Connect / Disconnect` which is imo confusing. These methods should mirror `Connect / Disconnect`.

src/include/duckdb/storage/storage_extension.hpp
```diff
@@ -39,6 +39,12 @@ class StorageExtension {
 	virtual ~StorageExtension() {
 	}
 
+	//! Whether catalogs from this storage extension accept `CONNECT name`. Default false; backends opt in
+	//! by overriding this AND Catalog::GetConnectFunctionName. Gates CONNECT before the first query runs.
+	virtual bool SupportsPassthrough() const {
+		return false;
```
Mytherin: Why do we need this in the storage extension vs just being in the catalog?

src/execution/operator/helper/physical_connect.cpp
```diff
@@ -14,7 +21,31 @@ SourceResultType PhysicalConnect::GetDataInternal(ExecutionContext &context, Dat
...
+
+	auto target = DatabaseManager::Get(client).GetDatabase(info->name);
+	if (!target) {
+		throw InvalidInputException("Database \"%s\" is not attached", info->name);
+	}
+	auto ext = target->GetStorageExtension();
+	if (!ext || !ext->SupportsPassthrough()) {
+		throw InvalidInputException(
+		    "Database \"%s\" does not support pass-through SQL execution (CONNECT). The storage extension must "
+		    "override StorageExtension::SupportsPassthrough() and the catalog must implement "
```
Mytherin: This is leaking internal implementation details in the error message which I think is unnecessary - the first sentence is enough.

src/main/client_context.cpp
```diff
@@ -944,6 +984,33 @@ unique_ptr<PendingQueryResult> ClientContext::PendingStatementOrPreparedStatemen
 unique_ptr<PendingQueryResult> ClientContext::PendingStatementOrPreparedStatement(
     ClientContextLock &lock, const string &query, unique_ptr<SQLStatement> statement,
     shared_ptr<PreparedStatementData> &prepared, const PendingQueryParameters &parameters) {
+	// CONNECT chokepoint: when bound, non-control SQL is rewritten in place and falls through to
+	// the normal pipeline. No recursion — the rewrite goes through PendingStatementInternal, not back here.
+	if (is_bound && statement) {
+		bool is_control = statement->type == StatementType::CONNECT_STATEMENT ||
```
Mytherin: A few notes here: 

* Relying on a statement being here means we must also have already parsed it. This seems wrong to me - since that effectively means we can't pass-through SQL unless DuckDB itself understands the SQL. However there are other challenges with that up-stream anyway, e.g. statement splitting and such, so maybe the answer to that is that remote catalogs "just" need to override the parser. 
* There's no guarantee for a statement to be here - this function takes **either** a query **or** a statement. See this downstream switch:

```cpp
		if (statement) {
			pending = PendingStatementInternal(lock, query, std::move(statement), parameters);
		} else {
			pending = PendingPreparedStatement(lock, query, prepared, parameters);
		}
```

This means right now in the former case `CONNECT` is silently ignored / by-passed. In this case we would need to parse the SQL string and extract CONNECT / DISCONNECT.

src/include/duckdb/main/client_context.hpp
```diff
@@ -97,7 +98,23 @@ class ClientContext : public enable_shared_from_this<ClientContext> {
 	//! Data for the currently running transaction
 	TransactionContext transaction;
 
+	//! Routing target for pass-through SQL execution (CONNECT/DISCONNECT).
+	//! When is_bound is true and bound_database can be locked, non-control SQL is rewritten at
+	//! the chokepoint as `SELECT * FROM <fn>('cat', '<sql>')` and runs through the normal pipeline.
+	weak_ptr<AttachedDatabase> bound_database;
```
Mytherin: Can these be `private`?

Mytherin: Also naming - `connected_to_database`

Mytherin (commented):
Thanks - looks good. I've left some comments

src/include/duckdb/storage/storage_extension.hpp
carlopi: I think the advantage of StorageExtension is, in my mind, that it allows for fail-fast in situations where there is not yet a Catalog. Like the future `CONNECT 'file.db' (TYPE sqlite)`, here we are able to fail before actually instantiating the Catalog, we know `duckdb` (at least, at this moment) don't support `CONNECT` semantic.

Slightly more stretched, but this also means we can query for capabilities of storage extensions.

Note that returning "" or throwing from `CatalogX::GetConnectFunctionName(ClientContext& context);` are also valid choices (like, a specific DB option might prevent `connect`-ability).

src/include/duckdb/main/client_context.hpp
carlopi: Thanks, handled renaming / private / other comments.

src/include/duckdb/storage/storage_extension.hpp
Mytherin: I would just keep it in the catalog - duplicating this just to throw an error message a bit earlier seems like not a great trade-off.

carlopi: Discussed also live, indeed, let's remove.

src/main/client_context.cpp
carlopi: Point 1: right, fixed.

Point 2: `PREPARED` staments are messy, expecially parametrized ones. I'll ship a first version that accept only non-parametrized prepared, and `throw NotImplementedException` otherwise, then iterate on this.

carlopi:
Thanks @Mytherin for the review, iterated a bit, now main hole to be aware it's likely requiring more thinking AND untested are prepared statements. I think current default (throw in non-obvious cases) should be OK, but I need to iterate on this, I would propose in the next PR (or the one after).

carlopi:
I don't think the failure in wasm can be related, for me this is ready to go.

carlopi:
Test failure on `Linux Relassert Tests` tracked internally, not really relevant.

src/main/client_context.cpp
```diff
@@ -942,6 +986,46 @@ unique_ptr<PendingQueryResult> ClientContext::PendingStatementOrPreparedStatemen
...
     ClientContextLock &lock, const string &query, unique_ptr<SQLStatement> statement,
     shared_ptr<PreparedStatementData> &prepared, const PendingQueryParameters &parameters) {
+	// CONNECT chokepoint: when connected, non-control SQL is rewritten in place and falls through to
+	// the normal pipeline. No recursion — the rewrite goes through PendingStatementInternal, not back here.
+	if (is_connected) {
+		bool is_control = false;
+		if (statement && IsConnectControlStatement(statement->type)) {
+			is_control = true;
+		}
+		if (!is_control) {
```
Mytherin, line 996: This would be nicer to move to a separate function

src/main/client_context.cpp
```diff
@@ -942,6 +986,46 @@ unique_ptr<PendingQueryResult> ClientContext::PendingStatementOrPreparedStatemen
 unique_ptr<PendingQueryResult> ClientContext::PendingStatementOrPreparedStatement(
     ClientContextLock &lock, const string &query, unique_ptr<SQLStatement> statement,
     shared_ptr<PreparedStatementData> &prepared, const PendingQueryParameters &parameters) {
+	// CONNECT chokepoint: when connected, non-control SQL is rewritten in place and falls through to
+	// the normal pipeline. No recursion — the rewrite goes through PendingStatementInternal, not back here.
+	if (is_connected) {
+		bool is_control = false;
+		if (statement && IsConnectControlStatement(statement->type)) {
```
Mytherin, line 993: `statement` is not guaranteed to be set so this is not right. This will lead to not being able to disconnect ever on clients that don't have `statement` defined here.

Mytherin (commented):
Thanks - looks good. Left some more minor comments. Some more future work here:

* Result types are lost, so also queries like `CREATE TABLE` that aren't supposed to return query results now return results, leading to them being printed in the CLI:

```sql
memory D create table t(i int);
┌────────┐
│ Count  │
│ int64  │
└────────┘
  0 rows
```

* The currently connected database is not shown in the CLI which makes it rather confusing to use at this moment. 

* This doesn't work yet with multi-statement queries like `PIVOT`, e.g.:
```sql
CREATE OR REPLACE TABLE monthly_sales(empid INT, amount INT, month TEXT);

INSERT INTO monthly_sales VALUES
    (1, 10000, '1-JAN'),
    (1, 400, '1-JAN'),
    (2, 4500, '1-JAN'),
    (2, 35000, '1-JAN'),
    (1, 5000, '2-FEB'),
    (1, 3000, '2-FEB'),
    (2, 200, '2-FEB'),
    (2, 90500, '2-FEB'),
    (1, 6000, '3-MAR'),
    (1, 5000, '3-MAR'),
    (2, 2500, '3-MAR'),
    (2, 9500, '3-MAR'),
    (1, 8000, '4-APR'),
    (1, 10000, '4-APR'),
    (2, 800, '4-APR'),
    (2, 4500, '4-APR');

PIVOT monthly_sales ON MONTH USING SUM(AMOUNT);

-- Query did not return any columns
```

I think this can all be picked up as future work, though.

Mytherin:
Thanks!

carlopi:
Thanks for the review!

I noted down the improvements, CLI prompt in particular I have in the follow up that's ready (https://github.com/duckdb/duckdb/pull/22950), I'll work also on the other items.

PR #22879, opened 2026-05-26
Initial implementation of `variant_exists` processing function

sebastiaan-dev:
This PR resolves https://github.com/duckdblabs/duckdb-internal/issues/9394, introducing an initial implementation for `variant_exists`. This function takes a `VARIANT`-type object as input and a path, returning true if the supplied path exists in the object and false otherwise.

### Examples

```bash
variant_exists({'a': {'c': 1, 'd': 2}}::VARIANT, '')
> true

variant_exists({'a': {'c': 1, 'd': 2}}::VARIANT, 'a')
> true

variant_exists({'a': {'c': 1, 'd': 2}}::VARIANT, ['a', 'z'])
> [true, false]
```

### Interface

`variant_exists(variant, path)`

#### Parameters

- `path` is of type `VARCHAR` or `VARCHAR[]`

### Follow-up work
The below points are known limitations and deferred to later PRs:

1. When supplying the variant parameter explicit casting is required (i.e. `::VARIANT`).
2. There is no support for JSONPath or JSON Pointer syntax, and one can only index a maximum of 1 level deep into an object (e.g. `variant_keys({'a': {'b': {'c': 1}}}, 'a.b')` will not work as 'a.b' is interpreted as a key in its entirety).
3. In contrast to `variant_extract` there is no support for using a numeric index to access part of a variant (e.g. `variant_keys([{'a': 1}]::VARIANT, 0)` will not work). I chose not to implement this as the equivalent `json_` function also does not support numeric indexing, instead focusing on supporting the more extensive JSONPath or JSON Pointer syntax.
4. There is no fast path for shredded variants, the function will always access the unshredded variant.
5. The supplied `path` has to be a constant.

For the above points the respective issues can be found at https://github.com/duckdblabs/duckdb-internal/issues/9125#issue-4375370873


src/include/duckdb/function/scalar/variant_utils.hpp
```diff
@@ -93,6 +139,11 @@ struct VariantUtils {
 	CollectNestedData(const UnifiedVariantVectorData &variant, VariantLogicalType expected_type,
 	                  const SelectionVector &value_index_sel, idx_t count, optional_idx row, idx_t offset,
 	                  VariantNestedData *child_data, ValidityMask &validity);
+	//! Generic unshredded traversal over a variant using a provided path
+	DUCKDB_API static void TraversePath(const UnifiedVariantVectorData &variant,
```
Mytherin, line 143: I wonder if we could make this take a `Vector &` instead?

Ideally eventually we would be able to also traverse shredded variants (and we would maybe get rid of the `UnifiedVariantVectorData` entirely). We can perhaps for now move the usage of `UnifiedVariantVectorData` into this function.

src/include/duckdb/function/scalar/variant_utils.hpp
```diff
@@ -93,6 +139,11 @@ struct VariantUtils {
...
+	                                    const vector<VariantPathComponent> &components, idx_t count,
+	                                    VariantNestedData *nested_data, ValidityMask &validity,
```
Mytherin: nit: we want to avoid raw pointers as much as possible - can this be a reference or an `optional_ptr` instead?

Mytherin (commented):
Thanks - looks good, some minor comments

src/include/duckdb/function/scalar/variant_utils.hpp
Tishj: I'm to blame for this because I used the same pattern elsewhere, I didn't use `optional_ptr` because it's an array, rather than a single instance.
Not sure what should be used for this, I agree that the raw pointer looks bad

Tishj (commented):
Thanks, I haven't looked too deeply into the logic, but from a glance it looks quite similar to the logic in place for `variant_keys`. I was wondering if we could somehow unify these into a shared abstraction.

Taking paths, traversing the structure, using a predicate to locate the items of interest and a callback to handle these items ? (or creating an iterator, as Mark prefers that over callback-based APIs ;))
Also fine to wait for `variant_array_length` to maybe get a clearer picture of how this method should be structured.

src/include/duckdb/function/scalar/variant_utils.hpp
Mytherin: `array_ptr`?

sebastiaan-dev: Replaced raw pointers

src/include/duckdb/function/scalar/variant_utils.hpp
sebastiaan-dev: Currently I don't think we can easily/cleanly do this. For example `variant_keys` requires `UnifiedVariantVectorData` for writing the key values into the result vector. If we only passed `Vector &` we would have `RecursiveToUnifiedFormat` inside and outside of `TraversePath`.

sebastiaan-dev:
@Tishj I've made some shared helpers such as `VariantUtils::TraversePath`/`VariantUtils::ExecutePathFunction` and shared bind data `VariantPathBindData` for processing functions that work with a `path`. Would you want me to abstract this further? I do see there is similar/duplicated skeleton structure for some of the functions, so I could see if that can be generalized for `variant_array_length` (such as writing outer-inner list)?

src/function/cast/variant/from_variant.cpp
```diff
@@ -210,14 +208,8 @@ static bool CastVariant(FromVariantConversionData &conversion_data, Vector &resu
...
                                  idx_t offset, idx_t count, optional_idx row) {
-	auto &allocator = Allocator::DefaultAllocator();
-
-	AllocatedData owned_child_data;
-	VariantNestedData *child_data = nullptr;
-	if (count) {
-		owned_child_data = allocator.Allocate(sizeof(VariantNestedData) * count);
-		child_data = reinterpret_cast<VariantNestedData *>(owned_child_data.get());
-	}
+	const auto owned_child_data = make_unsafe_uniq_array_uninitialized<VariantNestedData>(count);
```
Tishj, line 211: I'm not sure why this is changed, can we at least add an assertion that count isn't 0 ? Because I know from experience that this was added *because* count==0 causes a crash here

src/function/cast/variant/from_variant.cpp
```diff
@@ -286,14 +278,8 @@ static bool ConvertVariantToList(FromVariantConversionData &conversion_data, Vec
...
                                   const SelectionVector &sel, idx_t offset, idx_t count, optional_idx row) {
-	auto &allocator = Allocator::DefaultAllocator();
-
-	AllocatedData owned_child_data;
-	VariantNestedData *child_data = nullptr;
-	if (count) {
-		owned_child_data = allocator.Allocate(sizeof(VariantNestedData) * count);
-		child_data = reinterpret_cast<VariantNestedData *>(owned_child_data.get());
-	}
+	const auto owned_child_data = make_unsafe_uniq_array_uninitialized<VariantNestedData>(count);
```
Tishj, line 281: Same as above

src/function/cast/variant/from_variant.cpp
```diff
@@ -353,14 +339,8 @@ static bool ConvertVariantToArray(FromVariantConversionData &conversion_data, Ve
...
 	auto &target_type = result.GetType();
-	auto &allocator = Allocator::DefaultAllocator();
-
-	AllocatedData owned_child_data;
-	VariantNestedData *child_data = nullptr;
-	if (count) {
-		owned_child_data = allocator.Allocate(sizeof(VariantNestedData) * count);
-		child_data = reinterpret_cast<VariantNestedData *>(owned_child_data.get());
-	}
+	const auto owned_child_data = make_unsafe_uniq_array_uninitialized<VariantNestedData>(count);
```
Tishj, line 342: Same as above

src/include/duckdb/function/scalar/variant_utils.hpp
```diff
@@ -16,6 +16,21 @@
...
+	explicit VariantPathBindData(const string &input_path);
+	explicit VariantPathBindData(const vector<string> &input_paths);
+	VariantPathBindData(const VariantPathBindData &other) = default;
+
+public:
+	unique_ptr<FunctionData> Copy() const override;
+	bool Equals(const FunctionData &other) const override;
+
+public:
+	vector<vector<VariantPathComponent>> paths;
```
Tishj, line 31: From the constructors, it seems to me that the nested `vector` is always of size 1 or empty. If that's true, I'm not sure what we gain from this nested vector ?

src/function/scalar/variant/variant_bind_utils.cpp
```diff
@@ -14,4 +69,50 @@ bool VariantBindUtils::GetConstantArgument(ClientContext &context, const Express
...
+	}
+
+	if (arguments.size() == 1) {
+		// No path supplied, execute function over the root of the variant
+		return make_uniq<VariantPathBindData>();
+	}
+
+	const auto &path_expr = *arguments[1];
+	const auto &return_type = path_expr.GetReturnType();
+	if (return_type.id() != LogicalTypeId::VARCHAR && return_type.id() != LogicalTypeId::LIST) {
```
Tishj, line 90: If have a hunch that this needs to account for `SQLNULL` or `UNKNOWN`, or both(?)
I wonder what happens when we provide `NULL` as the path, without any casting.
As for `UNKNOWN`, I believe this gets created when we use `prepare` to create a prepared statement, for example `variant_keys(col, $my_paths_arg)`

See for example `union_tag.cpp`
Though I might be running on outdated knowledge and this gets handled upstream generically these days

src/function/scalar/variant/variant_exists.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+namespace duckdb {
+
+// TODO: Currently collection will always happen on the unshredded variant, introduce a fast path for shredded variants.
+static ValidityMask CollectVariantExistence(const UnifiedVariantVectorData &variant,
+                                            const vector<VariantPathComponent> &components, const idx_t count) {
+	ValidityMask path_validity(count);
+	VariantPathSelection path_selection(count);
+
+	const auto owned_nested_data = make_unsafe_uniq_array_uninitialized<VariantNestedData>(count);
```
Tishj, line 14: Same comment on `count`

test/sql/function/variant/variant_exists.test
```diff
@@ -0,0 +1,177 @@
...
+query I
+select variant_exists({'a': 1, 'b': {'c': {'d': 12}}}::VARIANT, '$.b.c');
+----
+false
+
+# Incorrect parameters
+
+statement error
+select variant_exists({'a': 1}::VARIANT, [1]);
+----
```
Tishj, line 144: Please supply an expected error (partial or full both work)

test/sql/function/variant/variant_exists.test
```diff
@@ -0,0 +1,177 @@
...
+
+statement error
+select variant_exists({'a': 1}::VARIANT, [NULL]);
```
Tishj, line 148: Can we also add a test for an empty list `[]` as `paths` argument?
And a test with both a valid path and a NULL: `['c', NULL]`

Perhaps `[]::VARCHAR[]` to make sure it gets past the logic in place to check for type mismatches

Tishj (changes requested):
Thanks for the changes! I like the sharing of bind logic and generalizing the path traversal logic, removes a lot of duplicated code already 👍 

I just have a few remarks about the code and a couple requests for additional tests, but other than that I think it looks great

src/function/cast/variant/from_variant.cpp
sebastiaan-dev: I replaced the raw pointer occurrences with `array_ptr<VariantNestedData>` in `VariantUtils::TraversePath`. This function uses `CollectNestedData` and `FindChildValues` which also used the raw pointers so I modified the interfaces for these function as well, then updated the calling sites to use `array_ptr`.

sebastiaan-dev: When you talk about a crash, I think you mean allocating with a size of 0 in `Allocator::AllocateData`? What `make_unsafe_uniq_array_uninitialized` seems to map to is `new VariantNestedData[n]` which seems fine for the 0 case.

src/include/duckdb/function/scalar/variant_utils.hpp
sebastiaan-dev: I made this nested with the idea of later supporting multi-level paths (i.e. `$.a.b`) through JSONPath and JSON Pointer syntax. Currently we don't do any parsing of the path argument and that is why, for now, I just put the entirety of the incoming path as the first component of a path. Then when we do have parsing it should produce multiple components.

sebastiaan-dev: So the nested vectors represent multiple multi-component paths (i.e. `['$.a.b', '$.a.c']`)

test/sql/function/variant/variant_exists.test
sebastiaan-dev: Empty list is already covered by `select variant_exists(a, []) from struct_cast_data();`. Added the case with `['c', NULL]`.

For the explicit cast, you mean apply this as:

```sql
select json_exists({'c': 1}, []::VARCHAR[]);
# output
> []
```

Or as:

```sql
select json_exists({'c': 1}, ['c', NULL]::VARCHAR[]);
# output
> Binder Error: JSON path cannot be NULL
```

(Both cases don't change behavior with/without casting)

src/function/scalar/variant/variant_bind_utils.cpp
sebastiaan-dev: When we do `select variant_exists({'a': 1}::VARIANT, NULL);` this returns `NULL`, which is the same as `json_exists`.

For the `UNKNOWN` case I can add the same exception (`ParameterNotResolvedException`)? But I was not able to trigger this exception with:

```sql
PREPARE variant_keys_path AS
SELECT variant_keys({'a': 1}::VARIANT, $1);

EXECUTE ...
```

So I am not sure how I should test for this?

PR #22880, opened 2026-05-26
Auto generate util functions (`Copy`, `Equals`, `Hash`, `Iterate`) for ParsedExpressions

Mytherin:
Currently all of these functions are hand-rolled. This PR instead moves towards auto-generating them into a single file (`util_parsed_expression.cpp`). This mirrors how the serialization routines are now implemented - and the same JSON with properties for each class is used to generate this code.


src/include/duckdb/parser/parsed_expression.hpp
```diff
@@ -14,8 +14,151 @@
 
 namespace duckdb {
 class Deserializer;
+class ParsedExpression;
 class Serializer;
 
+//! Lightweight const view of a ParsedExpression's direct children.
+//! Stores up to INLINE_CAPACITY children on the stack; spills to heap beyond that.
+//! Iterating yields const ParsedExpression& — read-only access, works on const expressions.
+class ConstChildrenView {
```
Maxxen, line 23: Is this really necessary? Would it not be simpler to just have a protected `vector<ParsedExpression*> children` in the base `ParsedExpression` class that remain empty for all "leaf" expressions? If a `vector<>` is too heavy, we could alternatively do a linked list (only adds two pointers).

Mytherin: Sure, but that’s a much bigger redesign since that means changing many public members. We can do that but step 1 is making those all private.

Mytherin: Or do you mean lazily filling up a separate  `vector<reference<ParsedExpression>> `in addition to the actual children that are stored in a variety of places?

That could be done as well I suppose, but I think it would be nicer to move to just having every expression having a `vector<unique_ptr<ParsedExpression>>`

Mytherin: Given the current mutable nature actually keeping a cache like that seems like a bad idea. I think the current implementation is better - we can then later remove it again when the child accessors are unified.

PR #22881, opened 2026-05-26
`COMMENT ON` aliased view column

yan-alex:
Previously this would fail:
```sql
memory D CREATE TEMPORARY VIEW v(c1) AS SELECT 1;
memory D COMMENT ON column v.c1 is 'hello'; 
Binder Error:
View "v" does not have a column with name "c1"                                                                                                                                     
memory D from v;
┌───────┐
│  c1   │                                                                                                                                                                          
│ int32 │                                                                                                                                                                          
├───────┤                                                                                                                                                                          
│     1 │                                                                                                                                                                          
└───────┘
```
Now this is fixed


Mytherin:
Thanks!

PR #22882, opened 2026-05-26
Fix geometry stats checkpointing when no changes are detected

Maxxen:
Even if no changes have been made to the geometry column data row group, we need to "lift up" the stats from the inner column on checkpoint, or they will be empty on deserialization until the next time the stats gets merged.

Issue originally reported in https://github.com/duckdb/duckdb-spatial/issues/817


Mytherin (approved):
Thanks! LGTM - will just retrigger CI

robjhornby (commented):
The test cases make sense to me, thanks. I'm not familiar with the rest of the codebase

PR #22885, opened 2026-05-26
Fix out-of-bounds write in decode ignore mode

jmestwa-coder:
decode(blob, 'ignore') sizes the cleaned output with EmptyString(new_str.size()) and then writes a terminating '\0' at output[new_str.size()], one byte past the allocation. string_t values are length-prefixed and never NUL-terminated, so a result of exactly 12 bytes overruns the inlined string_t. The write was introduced in 62f9c995b; removing it restores the original behavior.


Mytherin:
Thanks!

PR #22887, opened 2026-05-26
Pass INPUT_TYPE and STATE_TYPE to fix windows compilation

smvv:
In CI, Windows 64-bit amd64 MinGW [fails](https://github.com/duckdb/duckdb/actions/runs/26396998900/job/77708828366#step:28:1032):
```c++
// FAILED: [code=1] src/function/aggregate/distributive/CMakeFiles/duckdb_aggr_distr.dir/ub_duckdb_aggr_distr.cpp.obj
```

because

`HasClusteredOperation<OP, void_t_helper<decltype(&OP::template ClusteredOp<int32_t, int32_t, OP>)>>` checks whether `OP` has a `ClusteredOp` specialization for exactly `(INPUT_TYPE=int32_t, STATE_TYPE=int32_t)`.

For ops like `FirstFunctionString`, `ClusteredOp` is a template that is only valid when:

- `STATE_TYPE `looks like `FirstState<string_t>` (has `is_set`, `is_null`, `value`)
- `INPUT_TYPE` is compatible with string handling (`string_t`)

So probing with `(int32_t, int32_t)` forces an invalid specialization on MinGW:

- `state.is_set` on `int32_t` is invalid
- `SetValue(..., vals[idx], ...)` with `vals[idx]` as `int32_t` is invalid for string-only `SetValue`.

Thus, pass `INPUT_TYPE` and `STATE_TYPE` to fix the windows MinGW compilation error.

### DIFF
--- a/src/include/duckdb/common/vector_operations/aggregate_executor.hpp
+++ b/src/include/duckdb/common/vector_operations/aggregate_executor.hpp
@@ -560,10 +560,11 @@ public:
 	}
 
 	// true if OP defines ClusteredOp (early-exit optimization for first/last/any_value).
-	template <class OP, class = void>
+	template <class OP, class INPUT_TYPE, class STATE_TYPE, class = void>
 	struct HasClusteredOperation : std::false_type {};
-	template <class OP>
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #22888, opened 2026-05-26
Support empty string casts to BIT

### LINKED ISSUE #22863: DuckDB does not support casting an empty string to BIT
### What happens?

 DuckDB does not handle empty string to BIT conversion gracefully, it throws an error instead of returning 0 as PostgreSQL does.

### To Reproduce

```
-- DuckDB
memory D SELECT ''::bit;
Conversion Error:
Cannot cast empty string to BIT
```

```
-- PG
SELECT ''::bit;
bit 
-----
 0
(1 row)
```

### OS:

wasm_eh

### DuckDB Version:

v1.5.2

### DuckDB Client:

DuckDB Wasm (https://shell.duckdb.org)

### Hardware:

_No response_

### Full Name:

Yashneet Vinayak

### Affiliation:

AWS

- [x] Yes, I have

- [x] Yes, I have

Yes

feichai0017:
Closes #22863.

This treats an empty string cast to BIT as a one-bit zero, matching PostgreSQL behavior, while keeping empty BLOB casts and empty hex BIT strings rejected. The fix handles the empty input in the BIT conversion helpers so the output size and encoded bitstring stay consistent.

### DIFF
--- a/src/common/types/bit.cpp
+++ b/src/common/types/bit.cpp
@@ -130,8 +130,13 @@ string Bit::ToString(bitstring_t str) {
 }
 
 bool Bit::TryGetBitStringSize(string_t str, idx_t &str_len, string *error_message) {
-	auto data = const_data_ptr_cast(str.GetData());
 	auto len = str.GetSize();
+	if (len == 0) {
+		str_len = ComputeBitstringLen(1);
+		return true;
+	}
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #22890, opened 2026-05-26
Fix RevertAppend with empty trailing column segments

### LINKED ISSUE #22861: SegmentTree::TryGetSegmentIndex throws InternalException when RevertAppend runs after OOM leaves an empty-shell single-node segment tree
## What

`SegmentTree::TryGetSegmentIndex` in `src/include/duckdb/storage/table/segment_tree.hpp` (lines 207-242) does an unguarded `upper = index - 1` on `idx_t` (= `uint64_t`). When `index == 0` and `row_number < entry.row_start`, `upper` underflows to `UINT64_MAX`, the next loop iteration produces a huge `index`, and the bounds check on line 222 throws `InternalException("Segment tree index not found for row number ...")`.

PR #18585 ([link](https://github.com/duckdb/duckdb/pull/18585)) proposed exactly this defensive bounds fix and was closed in favor of fixing one specific upstream trigger via PR #18876. **However, the defensive code in `TryGetSegmentIndex` is still unchanged, and we just hit a new upstream trigger in production that #18876 does not cover.**

## Where (current main HEAD)

`src/include/duckdb/storage/table/segment_tree.hpp:207-242`:

```cpp
bool TryGetSegmentIndex(SegmentLock &l, idx_t row_number, idx_t &result) {
    ...
    idx_t lower = 0;
    idx_t upper = nodes.size() - 1;
    while (lower <= upper) {
        idx_t index = (lower + upper) / 2;
        if (index >= nodes.size()) {
            string segments;
            for (auto &entry : nodes) {
                segments += StringUtil::Format("Start %d Count %d",
                                               entry.row_start, entry.node->count.load());
            }
            throw InternalException("Segment tree index not found for row number %d\nSegments:%s",
                                    row_number, segments);
        }
        auto &entry = nodes[index];
        if (row_number < entry.row_start) {
            upper = index - 1;     // ← idx_t underflow when index == 0
        } else if (row_number >= entry.row_start + entry.node->count) {
            lower = index + 1;
        } else {
            result = index;
            return true;
        }
    }
    return false;
}
```

## New upstream trigger we observed (not covered by #18876)

We hit this twice in production (v1.4.4) within 6 hours on different pods. The exact error:

```
InternalException: Segment tree index not found for row number 367626
Segments:Start 368640 Count 0
```

Note the single segment with `Count=0` — i.e. the segment tree has exactly one node and that node is an **empty shell**.

### Reproduction conditions (high level)

1. DuckDB running in **in-memory mode**, **no `temp_directory` set** → no spill path
2. Several concurrent transactions doing `UPDATE ... FROM temp_table_N` against a high-row-count table
3. Memory pressure rises to the `memory_limit` (we set 18.6 GiB)
4. One transaction reaches `DuckTransaction::Commit()`. Inside:
   - `storage->Commit()` / `undo_buffer.Commit()` attempts to allocate a 256 KiB block → throws `OutOfMemoryException`
   - Catch block calls `undo_buffer.RevertCommit(...)`
5. RevertCommit walks down the chain:
   ```
   UndoBuffer::RevertCommit
     → DataTable::RevertAppend
       → RowGroupCollection::RevertAppendInternal
         → RowGroup::RevertAppend
           → StandardColumnData::RevertAppend
             → ColumnData::RevertAppend
               → SegmentTree::TryGetSegmentIndex(row_number=367626)
                 → idx_t underflow → throw InternalException
   ```
6. In `ColumnData::RevertAppend`, the early-return check `start_row >= last_segment->start + last_segment->count` is `367626 >= 368640 + 0 = 368640` → **false** (off by one — the segment is an empty shell at 368640, but `start_row` is below it). So we proceed into the binary search.
7. Binary search: `lower=0, upper=0, index=0`. `entry.row_start = 368640`. `row_number(367626) < 368640` → `upper = 0 - 1` → underflow → throw.

### Why this is a different trigger than #18876

PR #18876 patched `RevertAppend` to fully delete row groups instead of truncating to 0 rows, plus other fuzzer-found fixes. Our case is different:

- The empty-shell segment at `Start=368640, Count=0` was not produced by `RevertAppend` truncation. We believe it was left behind by the **OOM-aborted append in the first phase of commit** (a segment node was allocated, but the data write failed mid-flight and the node was left in the tree with count=0).
- `RevertAppend` is then called with `start_row=367626` — the row this transaction *intended* to start appending at, which is **before** the empty shell's `start=368640`.

This `start_row < smallest_segment.start` case is what triggers the binary search underflow. We have not seen this combination called out in #18585 / #18876 / #16331 / #4978.

## Combined impact with #22860

Because `DuckTransaction::Commit()` is declared `noexcept` and **does not** wrap the `RevertCommit` call in a try/catch (that's filed separately at #22860), this InternalException escapes the noexcept boundary → `__cxa_call_terminate` → `abort()` → entire process is killed.

So the practical effect is: any OOM during a commit on a memory-constrained in-memory database can kill the whole DuckDB process, taking down all attached databases. We hit this twice in 6 hours in prod.

## Asks

1. **(Primary)** Decide whether `TryGetSegmentIndex` warrants the defensive bounds fix proposed in PR #18585 given that recurring upstream triggers keep being discovered (this one, #16331, #4978, #3789, etc.). I am willing to send a new PR if defensive guarding is now acceptable.

2. **(Or)** Fix this specific upstream trigger in `ColumnData::RevertAppend` / `RowGroup::RevertAppend` so that an OOM-left empty-shell segment does not corner the binary search. One option: in `ColumnData::RevertAppend`, treat `start_row < last_segment->start` symmetrically to the existing "nothing was ever appended" early-return.

3. Either fix should also help the related closed reports: #16331, #4978, #3789, #7710, #622.

## Affected versions

- v1.4.4 (confirmed prod reproduction, two crashes within 6 hours)
- main HEAD as of 2026-05 (verified — bug still present)
- PR #18876 (merged Sep 2025) does NOT cover this trigger

Happy to share more details (stack pointers decoded via addr2line, full schema / formula context, etc.) if useful.

feichai0017:
Fixes #22861.

This fixes a rollback path in `ColumnData::RevertAppend` where an append can leave a zero-count trailing transient segment behind before any rows are written to it. If rollback then asks to revert to a count before that empty segment's start, `GetSegmentIndex` looks for a row in the gap and can throw an internal segment-tree error.

The fix drops zero-count trailing column segments before looking up the segment that contains the revert point. This keeps the fix in the column append/revert path instead of changing `SegmentTree::TryGetSegmentIndex` into a defensive guard for invalid states.

A C++ regression test is included because the SQL-level trigger depends on OOM timing during commit rollback and was not stable enough for a deterministic sqllogictest. The test constructs the minimal storage state directly: a valid column segment followed by an empty trailing segment, then calls `ColumnData::RevertAppend` back to the valid end.

### DIFF
--- a/src/storage/table/column_data.cpp
+++ b/src/storage/table/column_data.cpp
@@ -645,10 +645,9 @@ void ColumnData::RevertAppend(row_t new_count_p) {
 			return;
 		}
 	}
-	auto &last_segment = last_segment_node->GetNode();
-	if (new_count >= last_segment_node->GetRowStart() + last_segment.count) {
+	if (new_count >= last_segment_node->GetRowEnd()) {
 		// the start row is equal to the final portion of the column data: nothing was ever appended here
-		D_ASSERT(new_count == last_segment_node->GetRowStart() + last_segment.count);
+		D_ASSERT(new_count == last_segment_node->GetRowEnd());
 		return;
 	}
 	// find the segment index that the current row belongs to
--- a/src/storage/table/column_data.cpp
+++ b/src/storage/table/column_data.cpp
@@ -631,6 +631,20 @@ void ColumnData::RevertAppend(row_t new_count_p) {
 	if (!last_segment_node) {
 		return;
 	}
+	while (last_segment_node->GetCount() == 0 && new_count <= last_segment_node->GetRowStart()) {
+		// Appends can create the next transient segment before any rows are written to it. If the append is
+		// reverted after that point, drop the empty tail before looking up the row to truncate at.
+		auto last_segment_index = last_segment_node->GetIndex();
+		data.EraseSegments(l, last_segment_index);
+		if (last_segment_index > 0) {
+			auto previous_segment = data.GetSegmentByIndex(l, UnsafeNumericCast<int64_t>(last_segment_index - 1));
+			previous_segment->SetNext(nullptr);
+		}
+		last_segment_node = data.GetLastSegment(l);
+		if (!last_segment_node) {
+			return;
+		}
+	}
 	auto &last_segment = last_segment_node->GetNode();
 	if (new_count >= last_segment_node->GetRowStart() + last_segment.count) {
 		// the start row is equal to the final portion of the column data: nothing was ever appended here
--- a/test/sql/storage/test_storage.cpp
+++ b/test/sql/storage/test_storage.cpp
@@ -124,4 +124,5 @@ TEST_CASE("RevertAppend removes trailing empty column segments", "[storage]") {
 		REQUIRE(last_segment->GetRowEnd() == revert_count);
 		REQUIRE(column.count == revert_count);
 	}
+	REQUIRE_NOTHROW(con.Rollback());
 }
--- a/test/sql/storage/test_storage.cpp
+++ b/test/sql/storage/test_storage.cpp
@@ -1,7 +1,14 @@
 #include "catch.hpp"
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/catalog/catalog_entry/duck_table_entry.hpp"
 #include "duckdb/common/file_system.hpp"
-#include "test_helpers.hpp"
+#include "duckdb/main/config.hpp"
+#include "duckdb/main/database.hpp"
 #include "duckdb/main/appender.hpp"
+#include "duckdb/storage/data_table.hpp"
+#include "duckdb/storage/table/column_segment.hpp"
+#include "duckdb/storage/table/row_group.hpp"
+#include "test_helpers.hpp"
 
 using namespace duckdb;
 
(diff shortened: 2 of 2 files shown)


artjomPlaunov:
Just a heads up, this PR may need to be modified once this PR is merged: https://github.com/duckdb/duckdb/pull/22822/

since we will be allowing for gaps in rowids, and also removing the exception condition from the binary search (although it seems like your fixes don't rely on the exception, so that's good I think/hopefully your fix here is mostly orthogonal to the changes in the linked PR).

feichai0017:
> Just a heads up, this PR may need to be modified once this PR is merged: #22822
> 
> since we will be allowing for gaps in rowids, and also removing the exception condition from the binary search (although it seems like your fixes don't rely on the exception, so that's good I think/hopefully your fix here is mostly orthogonal to the changes in the linked PR).

Thanks for the heads up!  I’ll keep an eye on #22822 and adjust this PR if needed once that lands.

feichai0017:
I also checked this against current main after #22822 landed. The PR still applies cleanly, and the fix remains orthogonal to the rowid gap changes: it removes trailing empty column segments before calling GetSegmentIndex, so it does not rely on the old binary-search exception behavior.

Mytherin:
Thanks - could you try to create a SQL test instead of a C++ test that triggers this behavior?

feichai0017:
> Thanks - could you try to create a SQL test instead of a C++ test that triggers this behavior?

I tried to construct a SQL-only regression using low memory limits, small initial_column_segment_size, VARCHAR appends, commit conflicts, and a small ROW_GROUP_SIZE. The SQL cases either fail before the empty trailing segment is linked, or leave a non-empty segment covering the rollback row, so they do not reproduce the SegmentTree underflow on the old code.

The current C++ test is intentionally constructing the minimal storage state from the production failure: a trailing empty column segment after the append boundary. I do not think an OOM-based sqllogictest would be stable enough or would fail pre-fix for the right reason.

PR #22891, opened 2026-05-26
Allow substring with only `for` or `from`

Dtenwolde:

Small regression fix, the `SUBSTRING` used to work with only `FOR` or only `FROM`, but didn't work with the new parser:
```sql
SELECT SUBSTRING('hello' FROM 2); 
SELECT SUBSTRING('hello' FOR 3); 
```


Mytherin:
Thanks!

Issue #22304, opened 2026-04-27
Internal error with "ANY" operator in some case


tanscorpio7:
### What happens?

A simple select query with ANY operator runs into INTERNAL Error: Expected unified vector format of type INT32, but found type INT64.

While looking at a different issue https://github.com/duckdb/duckdb/issues/19241, ran into this issue.

```
memory D CREATE TABLE tenk1 AS SELECT i AS unique1, i*2 AS unique2 FROM range(10) t(i);
memory D SELECT unique1, unique2 FROM tenk1 WHERE ((unique1, unique2) < ANY (SELECT 2, 4));
INTERNAL Error:
Expected unified vector format of type INT32, but found type INT64

Stack Trace:

0        duckdb::Exception::ToJSON(duckdb::ExceptionType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 48
1        duckdb::Exception::Exception(duckdb::ExceptionType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 36
2        duckdb::InternalException::InternalException(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 20
3        duckdb::InternalException::InternalException<duckdb::PhysicalType, duckdb::PhysicalType const&>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, duckdb::PhysicalType&&, duckdb::PhysicalType const&) + 52
4        void duckdb::UnifiedVectorFormat::VerifyVectorType<int>() const + 116
5        unsigned long long duckdb::(anonymous namespace)::TemplatedDistinctSelectOperation<duckdb::DistinctGreaterThan>(duckdb::Vector&, duckdb::Vector&, duckdb::optional_ptr<duckdb::SelectionVector const, true>, unsigned long long, duckdb::optional_ptr<duckdb::SelectionVector, true>, duckdb::optional_ptr<duckdb::SelectionVector, true>, duckdb::optional_ptr<duckdb::ValidityMask, true>) + 9232
[...]

PR #22894, opened 2026-05-26
Add regression tests for row comparison ANY/ALL with mismatched types

adityasingh2400:
Follow-up to #22304, now test-only.

When I opened this it also carried a binder fix. That is no longer needed: the `LogicalType::TUPLE` work on main reconciles the LHS row and the subquery columns through `TryGetMaxLogicalType`, so the internal error in the struct comparator is gone and #22304 is closed. I dropped my version of the fix and rebased onto main so the diff is purely the regression coverage that main does not have yet.

What the added cases pin down:

- the original repro from #22304 (`(unique1, unique2) < ANY (SELECT 2, 4)` against an INTEGER/BIGINT mix) across `<`, `<=`, `>`, `>=` and `ALL`
- mixed integer widths in both directions, so neither a wider LHS nor a wider RHS regresses
- a three-element row mixing TINYINT, INTEGER and BIGINT
- DOUBLE and DECIMAL targets, where the common type is not an integer at all
- genuinely incompatible element types still producing a clean binder error rather than an internal one
- row and column count mismatches in both directions still producing the width error

I verified every expectation against a current nightly build. The incompatible-types case now reports the full tuple types rather than the element pair, so that expectation is written as a regex against the message main actually produces.

### DIFF
--- a/test/sql/subquery/any_all/test_row_comparison_any_all.test
+++ b/test/sql/subquery/any_all/test_row_comparison_any_all.test
@@ -246,3 +246,80 @@ query I
 SELECT (0, 0) < ANY(SELECT 1, 0);
 ----
 true
+
+# Issue #22304: row comparison where the LHS struct and the subquery columns
+# have different but castable types must reconcile both sides to a common type.
+# Previously this triggered an internal error in the struct comparator
+# (e.g. "Expected unified vector format of type INT32, but found type INT64").
+statement ok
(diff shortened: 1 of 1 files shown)


adityasingh2400:
Rebased onto main and reduced this to tests only.

The binder fix I had here is redundant now that `LogicalType::TUPLE` handles the type reconciliation, so I dropped it and kept just the regression cases for #22304, which main does not cover yet. I re-checked every expected value against a current nightly. One expectation changed as a result: the incompatible-element-type error now names the full tuple types instead of the element pair, so I match on that message instead.

PR #21990, opened 2026-04-09
de-dupe tests, avoid abs/rel path dups in extension runs

benfleis:
In extension builds, local tests (can) get inserted twice -- once due to ./test path loading (relative path) and once due to CMake configuring test paths (absolute path). To avoid dup test runs and save CI, build a map and de-dupe all the tests, to canonicalized absolute paths. In e.g. delta, this saves O(40) duplicate tests; YMMV.

PR #22896, opened 2026-05-26
Test runner: prevent double-registration of tests

staticlibs:
There is a problem with `unittest` SQLlogic runner that, when tests are run from the extension repo directory, extension tests from under `test/sql` can be registered twice. First they are registered from under the relative `test` directory and the second time from [ExtensionHelper::LoadedExtensionTestPaths()](https://github.com/duckdb/duckdb/blob/a93109abaed85521e4199770f6ff205aa6995a87/extension/generated_extension_loader.cpp.in#L25) that is generated at configure-time and contains `TEST_DIR` argument passed to `duckdb_extension_load` CMake function.

There is no tests deduplication on registration:

 - if `TEST_DIR` is specified as `test` or `test/sql` then the runner fails with:

```
error: TEST_CASE( "test/sql/storage/attach_simple.test" ) already defined.
    First seen at /home/alex/projects/duckdb-sqlite/duckdb/test/sqlite/test_sqllogictest.cpp:41
    Redefined at /home/alex/projects/duckdb-sqlite/duckdb/test/sqlite/test_sqllogictest.cpp:41
```

 - if `TEST_DIR` is not specified, the absolute path `/path/to/extension/repo/test/sql` is used for it; that results into the test to be registered twice - once on relative path and once on absolute path.

The problem is not severe, when the test runner is invoked with a pattern argument like this (used in most cases):

```
./build/release/unittest 'test/*'
```

Then only relative path-registered tests are run.

But the problem stil remains with `.test_slow`. Such tests are registed with `[.]` group label. The double registration problem makes it impossible (at least I didn't find the way without additional `require-env`) to run all tests once excluding `.test_slow` (that is desired for ASan/TSan test runs):

 - `./build/release/unittest` - runs tests (excluding `.test_slow`) twice
 - `./build/release/unittest 'test/*'` - runs test once, but does not exclude `.test_slow`
 - `./build/release/unittest '~[.]'` (Catch2 group negation) - runs tests (excluding `.test_slow`) twice

This PR attempts to detect that relative `test/sql` path and `ExtensionHelper::LoadedExtensionTestPaths()` path are resolved to the same path and skip the second registration.

### DIFF
--- a/test/sqlite/test_sqllogictest.cpp
+++ b/test/sqlite/test_sqllogictest.cpp
@@ -1,5 +1,6 @@
 #include "catch.hpp"
 #include "duckdb.hpp"
+#include "duckdb/common/path.hpp"
 #include "duckdb/common/string_util.hpp"
 #include "duckdb/main/extension/generated_extension_loader.hpp"
 #include "duckdb/parser/parser.hpp"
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks - this was already proposed here as well - https://github.com/duckdb/duckdb/pull/21990 - see my comments there. 

I'm fine with merging this but it's really a work-around slapped on another work-around. The main issue is that the way we run tests in the main repo (through `LOAD_TESTS`) is hacky. This used to be in the `extension-template` repo as well (in `extension_config.cmake`) which causes extension repositories that still have `LOAD_TESTS` to double-register tests. This can be fixed per extension by just removing `LOAD_TESTS` from the extension config.

Ideally, however, we would deprecate `LOAD_TESTS` entirely and just use `--test-dir <test_dir>` to run tests in the main repo. By deprecating that setting all extensions that still have it will just get a deprecation message and the issue is solved.

staticlibs:
Thanks for details! Closing in favour of #21990.

PR #22900, opened 2026-05-27
Mark json_deserialize_sql as fallible

### LINKED ISSUE #22761: Optimizer triggers parser error on JSON
### What happens?

Hi,

I found that the following test case triggers a parser error:

```
CREATE TABLE t(j JSON);
INSERT INTO t VALUES
  ('{"error":false,"statements":[]}'::JSON),
  (json_serialize_sql('SELECT 1'));

SELECT count(*)
FROM t
WHERE json_exists(j, '$.statements[0]')
  AND json_deserialize_sql(j) IS NOT NULL; -- Error parsing json: no statements
```

However, if I disable the optimizer, then the query can execute successfully:

```
PRAGMA disable_optimizer;

CREATE TABLE t(j JSON);
INSERT INTO t VALUES
  ('{"error":false,"statements":[]}'::JSON),
  (json_serialize_sql('SELECT 1'));

SELECT count(*)
FROM t
WHERE json_exists(j, '$.statements[0]')
  AND json_deserialize_sql(j) IS NOT NULL;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
```

### To Reproduce

```
CREATE TABLE t(j JSON);
INSERT INTO t VALUES
  ('{"error":false,"statements":[]}'::JSON),
  (json_serialize_sql('SELECT 1'));

SELECT count(*)
FROM t
WHERE json_exists(j, '$.statements[0]')
  AND json_deserialize_sql(j) IS NOT NULL; -- Error parsing json: no statements
```

### OS:

Ubuntu 24.04

### DuckDB Version:

527a98456f954fc970e6a070faa9dea4c034341

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Chi Zhang

### Affiliation:

Tsinghua university

- [x] Yes, I have

- [x] Yes, I have

Yes

feichai0017:
Fixes #22761.

`json_deserialize_sql` can throw parser errors while deserializing invalid serialized SQL JSON, but the scalar function was registered with the default non-fallible error mode. That allowed filter pushdown to move `json_deserialize_sql(j) IS NOT NULL` below a guarding `json_exists` predicate and evaluate it for rows that should have been filtered first.

This marks `json_deserialize_sql` as fallible so existing `CanThrow()` checks keep the expression above the scan when needed.

### DIFF
--- a/extension/json/json_functions/json_serialize_sql.cpp
+++ b/extension/json/json_functions/json_serialize_sql.cpp
@@ -246,8 +246,10 @@ static void JsonDeserializeFunction(DataChunk &args, ExpressionState &state, Vec
 
 ScalarFunctionSet JSONFunctions::GetDeserializeSqlFunction() {
 	ScalarFunctionSet set("json_deserialize_sql");
-	set.AddFunction(ScalarFunction({LogicalType::JSON()}, LogicalType::VARCHAR, JsonDeserializeFunction, nullptr,
-	                               nullptr, JSONFunctionLocalState::Init));
+	auto function = ScalarFunction({LogicalType::JSON()}, LogicalType::VARCHAR, JsonDeserializeFunction, nullptr,
+	                               nullptr, JSONFunctionLocalState::Init);
+	function.SetFallible();
+	set.AddFunction(std::move(function));
 	return set;
 }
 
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #22904, opened 2026-05-27
Add test stabilization reruns to run_tests

smvv:
### Rerun changed tests
- Add `--stabilize-tests` to rerun selected tests multiple times (mainly used in extension repo CI).
- Rerun newly added or changed tests when `--changed-tests` is enabled.
- Limit the total test run count to avoid excessive test time in CI: 
  - fast tests run 10 times, slow tests 3 times.
  - if >500 tests changed, rerun fast tests 3 times.
- Fails the config if any stabilization rerun fails.

Note: `--changed-tests` is only used in CI jobs `linux release` and `linux relassert`.
Other CI jobs are unaffected.

### Misc
- Add prefix `warn: ` to runtime threshold message to indicate its a warning instead of an error.


Mytherin:
Thanks!

PR #22905, opened 2026-05-27
Validate width/scale in duckdb_create_decimal_type and duckdb_create_decimal

mlafeldt:
This adds missing validation to the C API to reject invalid decimals as the CLI does.

Also fixes a misleading error message in the binder.


mlafeldt:
`duckdb_create_decimal` has the same problem. Fixing.

Mytherin:
Thanks!

PR #22911, opened 2026-05-27
Use non-deleted row count in `RowGroupReorderer`

lnkuiper:
Instead of the committed count. This is essentially the most conservative row count that we can do.

Follow-up of https://github.com/duckdb/duckdb/pull/22884
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9415 (hopefully for real this time)

### DIFF
--- a/src/storage/table/row_group_reorderer.cpp
+++ b/src/storage/table/row_group_reorderer.cpp
@@ -22,9 +22,10 @@ bool CompareValues(const Value &v1, const Value &v2, const OrderByStatistics ord
 	return (order == OrderByStatistics::MAX && v1 < v2) || (order == OrderByStatistics::MIN && v1 > v2);
 }
 
-idx_t GetQualifyingTupleCount(RowGroup &row_group, BaseStatistics &stats, const OrderByColumnType type) {
+idx_t GetMinimumQualifyingTupleCount(RowGroup &row_group, BaseStatistics &stats, const OrderByColumnType type,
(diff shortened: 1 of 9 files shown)


Mytherin:
Can't we push the transaction data for the actual scan we are doing to get the correct row counts for that transaction? Then we could use the existing `GetVisibleRowCount` method.

src/storage/table/row_group_reorderer.cpp
```diff
@@ -24,7 +24,7 @@ bool CompareValues(const Value &v1, const Value &v2, const OrderByStatistics ord
 
 idx_t GetQualifyingTupleCount(RowGroup &row_group, BaseStatistics &stats, const OrderByColumnType type) {
```
Mytherin: Minor side note but maybe this method should be renamed to `GetMinimumQualifyingTupleCount` because the current naming is confusing as it's not returning an exact count but rather a lower bound.

Mytherin:
> Can't we push the transaction data for the actual scan we are doing to get the correct row counts for that transaction? Then we could use the existing `GetVisibleRowCount` method.

Looking at the code we have a `ClientContext` everywhere the `RowGroupReorderer` is created so this shouldn't be too much of a plumbing effort. You can call `TransactionData(DuckTransaction::Get(context, storage.GetAttached()))` to get a `TransactionData`.

Mytherin:
Thanks!

PR #22914, opened 2026-05-27
Add the `RemotePushdownOptimizer` - a pre-bind optimizer that analyzes a parse tree and pushes down SQL directly into remote catalogs

Mytherin:
This PR adds the `RemotePushdownOptimizer`. This is a new kind of optimizer that runs *before binding*, i.e. it operates on the parse-tree directly. The core idea behind this optimizer is that, when interacting with a remote catalog (as exposed through e.g. `quack`), we might not want to do binding of (parts of) the query locally. Binding requires information such as which tables are available, which columns those tables have, etc. Doing binding locally brings along certain complications:

* We can do a round-trip to remote to figure out which tables exist and what those tables look like, but this is potentially expensive. In particular for small short-running queries this round-trip can easily double the execution time of those queries.
* The natural next step is to do caching - we can cache which tables are present and their schemas, but this brings along cache invalidation issues. We might now be binding and planning queries against stale information (e.g. we might not know certain tables exist, or the schema of tables might have changed). We can solve this by doing re-binding in those scenarios, but this is not very pretty either.

In a lot of cases queries are either "all or nothing" - i.e. they either run fully locally, or fully remotely. For the full remote case there's no benefit to doing any binding locally. We might as well just ship the query to be bound and executed remotely. This is particularly the case for simple queries which benefit the most from the reduced round-trips, e.g. for a query like `INSERT INTO tbl VALUES (42)` - the quickest way to execute this query is to just directly ship the SQL to the server instead of first trying to bind the type of `42` to `tbl` locally.

### RemotePushdownOptimizer

This PR allows exactly this to happen. We introduce a new property to catalogs: `IsRemoteCatalog()`, together with a new callback `RemotePushdown`:

```cpp
virtual unique_ptr<TableRef> RemotePushdown(ClientContext &context, unique_ptr<QueryNode> node);
```

If `IsRemoteCatalog()` is set then the catalog will qualify for the `RemotePushdownOptimizer`. The `RemotePushdown` function pushes a query node into the catalog, and returns a new table ref. This function will for `quack` for example return `quack_query_by_name('{catalog_name}', '{sql_query}')`.

### Pushdown Considerations

The optimizer effectively does a "partial bind" where we bind exclusively to local catalogs (usually the `temp` and `system` catalogs) and avoid ever binding to remote catalogs. We try to figure out which parts of the query can be fully pushed to a single remote catalog, and then call `RemotePushdown` on the remote catalog to generate the pushdown query.

Note that we're not doing full binding / resolution here - we just make assumptions based on catalog / schema information that is present in the query. For example, a direct catalog reference is easy:

```sql
-- "remote" is a remote catalog, we can ship this query to remote
SELECT COUNT(*) FROM remote.tbl;
```

Search path resolution is more difficult, but we can use the process-of-elimination to figure out if a table has to be remote or not. Essentially if we have only a single remote catalog in the search path, we can just check all of the local catalogs in the search path for the table. If none of the local catalogs have the table we know it must either not exist or be in the remote. In this case we just ship it to the remote. For example:

```sql
-- search path is now "temp, remote, system"
USE remote;
-- if "tbl" is no in "temp" or in "system", we know it must be in remote OR not exist at all, allow shipping to remote
SELECT COUNT(*) FROM tbl;
```

The optimizer traverses the parse tree and "peels off" layers that it knows are either (1) not associated with any catalog (e.g. `VALUES`, `SELECT` without a table, functions like `range`, etc), or (2) belong to the same remote catalog, and tries to push as much as possible into the remote catalog. 

### Partial Pushdown & Correlated Subqueries

This doesn't necessarily have to be an all-or-nothing on the entire query level, e.g. we can also pushdown subqueries even if outer queries refer to local components, for example:

```sql
-- we can pushdown the subquery into the remote
SELECT * FROM (SELECT COUNT(*)) FROM remote.tbl) subq, local_tbl;
```

However, we need to be conservative here as we are not actually doing any binding. Correlated subqueries complicate this as columns can cross what appear to be boundaries at which we can do this pushdown. For example:

```sql
-- does "a" belong to "local_tbl" or to "remote_tbl"?
SELECT (SELECT SUM(a) FROM remote.tbl) FROM local_tbl;
```

In these types of mixed scenarios we are conservative and just assume that columns belong to local tables to ensure we don't break any queries that otherwise would work. In the future we could potentially extend this by actually doing a look-up into `local_tbl` and figuring out if `a` belongs to any local tbl or not, effectively using the same process-of-elimination as we are doing at the table level. For now we just assume if there is a local table in the query that all columns that *can* reference it belong to it as well and prevent the pushdown.

### Local Functions, Macros & Types

When dealing with functions, users can define functions in a local catalog, e.g.:

```sql
CREATE TEMPORARY FUNCTION my_func(a) AS a + 42;
```

When pushing to remote, we would either have to push these functions there as well, or execute them locally. For now we opt to have these functions stop pushdown and skip pushing down subtrees that make references to these local functions.

Similarly they can define types:

```sql
CREATE TEMPORARY TYPE my_type AS STRUCT(u UBIGINT);
```

And we similarly avoid pushing down subtrees that contain local types.

### Separate Test Runner Databases

This PR also introduces a new feature to the test runner allowing different connections to be launched in different `DuckDB` databases. This uses the `db_name:connection_name` syntax. This is mostly useful for `quack` as we can still communicate across the database boundaries. For example:

```sql
statement ok db1:con1
call quack_serve('quack:localhost', token='asdf');

statement ok db1:con1
create table t1(i int);

statement ok db1:con1
insert into t1 from range(10);

statement ok
attach 'quack:localhost' as rpc (token 'asdf')

query I
select sum(i) from rpc.t1;
----
45
```

### Future Work

Currently for remote catalogs we assume DuckDB SQL can be pushed as-is into the remote catalog. This is of course not true for other types of catalogs, e.g. for Postgres or MySQL. In a follow-up PR I plan to extend the catalog interface to allow more selectivity in what gets pushed down to allow catalogs to state e.g. if they support certain functions or expressions.


carlopi (commented):
One question is on the signature of:
```c++
unique_ptr<TableRef> Catalog::RemotePushdown(ClientContext &context, unique_ptr<QueryNode> node);
```
Second parameter, should it be a `unique_ptr<QueryNode>`, or could it be directly a `string` ? Or maybe could it take both?

In https://github.com/duckdb/duckdb/pull/22878 I am basically reusing a similar primitive, but there I have only a string (since query might not be fully parsed), and also here as far as I can see only a string is actually used.
I am trying to figure out what's the shape of the generalization.

Maybe should we have both:
```c++
// on Catalog:
bool IsRemote(); 
unique_ptr<TableRef> RemotePushdown(ClientContext &context, unique_ptr<QueryNode> node);
unique_ptr<TableRef> RemotePushdown(ClientContext &context, string sql);
```
?

And, for example for quack, the implementation of the first can become just: `return RemotePushdown(context, node->ToString())` ?

Naming wise: why not `RemoteExecute` (instead of `RemotePushdown`) Like, the fact that we use for pushdown is the caller business?

Mytherin:
The idea behind it being a `QueryNode` is that we can modify it easily if required, instead of having to (again) parse it. This is only used as a string right now but that will not always be the case in the future.

We can have a second callback for just pushing a string for `CONNECT`. I think having two different callbacks there makes sense.

I'm fine with renaming to `RemoteExecute`.

carlopi:
I'll move `CONNECT` PR to be similar to this, and yes, I see `QueryNode` is more powerful, even just to allow `Catalog` specific rewrites rules.

PR #22912, opened 2026-05-27
Add AFTER DELETE trigger with REFERENCING OLD TABLE AS support

EtgarDev:
## Summary

- Adds `REFERENCING OLD TABLE AS <alias>` support for `AFTER DELETE` triggers - the alias exposes deleted rows via `DELETE ... RETURNING *` captured in the base CTE
- Extends the `ReferencingClause` grammar to support both `NEW TABLE` and `OLD TABLE` in a single clause, with a new `TriggerTableReferencingInfo` struct
- Validates alias/event combinations at `CREATE TRIGGER` time: 
    * `OLD TABLE` on INSERT rejected as `BinderException`
    * `NEW TABLE` on DELETE rejected as `BinderException`
    * `OLD TABLE` on UPDATE is rejected as `NotImplementedException`
- Splits generic referencing validation tests into `trigger_referencing_validation.test`

Based on https://github.com/duckdb/duckdb/pull/22759

PR #22916, opened 2026-05-27
Change paths-ignore to paths in CI workflows

smvv:
@lnkuiper noticed that `NightlyTests.yml` and `ExtendedTests.yml` ran for a [PR](https://github.com/duckdb/duckdb/pull/22912) that should not run those workflows (no related workflow files changes).

Codex noticed that `paths-ignore` is the wrong filter for only running the workflow file when specific files are changing, because the `!` is not supported.

In the CI workflows, we have:
```yaml
- paths-ignore: ['**', '!.github/workflows/...yml']
```
But GitHub’s documented pattern is:

- use `paths-ignore` only for pure exclusions
- use `paths` (with `!` negation) when mixing include + exclude

So this config is effectively not a reliable “only run when this workflow file changes” gate on `pull_request` events, and the PR opened event still triggered both runs.

Recommended fix:
```yaml
on:
  pull_request:
    types: [opened, reopened, ready_for_review]
    paths:
      - '.github/workflows/ExtendedTests.yml'
```
(and similarly for NightlyTests.yml).

From the github [docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore):
> Use the `paths` filter when you want to include file path patterns or when you want to both include and exclude file path patterns.
> Use the `paths-ignore` filter when you only want to exclude file path patterns.
>
> You cannot use both the `paths` and `paths-ignore` filters for the same event in a workflow.
>
> If you want to both include and exclude path patterns for a single event, use the `paths` filter prefixed with the `!` character to indicate which paths should be excluded.


Mytherin:
Thanks!

PR #8953, opened 2023-09-16
[PythonDev] Protect `test_replacement_scan.py` from side effects

Tishj:
Using `duckdb.sql` makes it prone to outside influence, so they now all use their own connection, hopefully fixes CI failing randomly in a non-reproducable manner

PR #22918, opened 2026-05-27
Internal #8953: Uncorrelated STRUCT Sizes

hawkfish:
* Properly test that the compared struct size matches the subquery row count.


Mytherin:
Thanks!

PR #22922, opened 2026-05-28
Parse short timestamp times with UTC suffixes

### LINKED ISSUE #22920: DuckDB fails to parse `1970-01-01 00:00Z` as `TIMESTAMP` but accepts similar variants
### What happens?

DuckDB rejects a `TIMESTAMP` literal when the time is written as `HH:MM` followed by `Z`, even though the same value is accepted without `Z`, and is also accepted when seconds are included before `Z`.

This works:

```sql
SELECT TIMESTAMP '1970-01-01 00:00';
```

This also works:

```sql
SELECT TIMESTAMP '1970-01-01 00:00:00Z';
```

But this fails:

```sql
SELECT TIMESTAMP '1970-01-01 00:00Z';
```

with:

```text
Conversion Error:
invalid timestamp field format: "1970-01-01 00:00Z", expected format is (YYYY-MM-DD HH:MM:SS[.US][±HH[:MM[:SS]]| ZONE])
```

This seems inconsistent. If `HH:MM` is accepted, and `Z` is accepted after `HH:MM:SS`, I would expect `HH:MMZ` to either be accepted as well, or for the accepted formats to be stricter across the board.

All three cases work in Postgres.

### To Reproduce

Run
```sql
SELECT TIMESTAMP '1970-01-01 00:00Z';
SELECT TIMESTAMP '1970-01-01 00:00';
SELECT TIMESTAMP '1970-01-01 00:00:00Z';
```
in DuckDB and Postgres (via https://pglite.dev/repl/).

### OS:

macOS 26.5

### DuckDB Version:

DuckDB v1.5.2 (Variegata)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Zimo Li

### Affiliation:

Garner

- [x] Yes, I have

- [x] Yes, I have

feichai0017:
Fixes #22920.

`Timestamp::TryConvertTimestampTZ` parses the date first, then parses the remaining time portion before handling timestamp suffixes such as `Z`, UTC offsets, and timezone names. The time parser accepts short `HH:MM` input only when the parsed time reaches the end of the provided string, so `TIMESTAMP '1970-01-01 00:00Z'` failed before the timestamp suffix handling could consume the `Z`.

This keeps the fix in the timestamp parser: it first tries the existing full time parse, and only falls back to parsing a time prefix when the following characters can start a timestamp suffix that is handled by the existing code below. This avoids changing the broader `TIME`/`INTERVAL` parsing behavior.

### DIFF
--- a/src/common/types/timestamp.cpp
+++ b/src/common/types/timestamp.cpp
@@ -30,12 +30,7 @@ static bool CanStartTimestampSuffix(char c) {
 	return c == 'Z' || c == '+' || c == '-' || StringUtil::CharacterIsSpace(c);
 }
 
-static bool TryConvertTimestampTimePrefix(const char *str, idx_t len, idx_t &time_pos, dtime_t &time,
-                                          optional_ptr<int32_t> nanos) {
-	if (Time::TryConvertInterval(str, len, time_pos, time, false, nanos)) {
-		return true;
-	}
-
+static idx_t TimestampTimeLength(const char *str, idx_t len) {
 	idx_t suffix_pos = 0;
 	while (suffix_pos < len && StringUtil::CharacterIsSpace(str[suffix_pos])) {
 		suffix_pos++;
@@ -43,16 +38,7 @@ static bool TryConvertTimestampTimePrefix(const char *str, idx_t len, idx_t &tim
 	while (suffix_pos < len && !CanStartTimestampSuffix(str[suffix_pos])) {
 		suffix_pos++;
 	}
-	if (suffix_pos == 0 || suffix_pos == len) {
-		return false;
-	}
-
-	// HH:MM is valid when it is followed by a timestamp suffix that is parsed below.
-	time_pos = 0;
-	if (!Time::TryConvertInterval(str, suffix_pos, time_pos, time, false, nanos)) {
-		return false;
-	}
-	return time_pos == suffix_pos;
+	return suffix_pos;
 }
 
 // timestamp/datetime uses 64 bits, high 32 bits for date and low 32 bits for time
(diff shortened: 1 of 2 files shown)


src/common/types/timestamp.cpp
lzm0: I wonder if the expected format syntax should also be updated, since the second part is optional.

feichai0017: Thanks for pointing this out. I updated the expected format hint to show that seconds are optional now: `YYYY-MM-DD HH:MM[:SS[.US]][±HH[:MM[:SS]]| ZONE]`.

src/common/types/timestamp.cpp
```diff
@@ -66,7 +95,7 @@ TimestampCastResult Timestamp::TryConvertTimestampTZ(const char *str, idx_t len,
 	// TryConvertTime may recursively call us, so we opt for a stricter
 	// operation. Note that we can't pass strict== true here because we
 	// want to process any suffix.
```
jsoref: This comment should _probably_ be updated as it was previously referencing the function on the next line and now _indirectly_ calls it via `TryConvertTimestampTimePrefix`

feichai0017: sry for that, alreay updated

src/common/types/timestamp.cpp
```diff
@@ -26,6 +26,35 @@ static inline T TemporalRound(T value, T scale) {
...
+	while (suffix_pos < len && !CanStartTimestampSuffix(str[suffix_pos])) {
+		suffix_pos++;
+	}
+	if (suffix_pos == 0 || suffix_pos == len) {
+		return false;
+	}
+
+	// HH:MM is valid when it is followed by a timestamp suffix that is parsed below.
+	time_pos = 0;
+	if (!Time::TryConvertInterval(str, suffix_pos, time_pos, time, false, nanos)) {
```
Mytherin: Do we need to do this double-parse? It feels like this could be implemented in a cleaner manner

Mytherin (commented):
Thanks - left a comment

src/common/types/timestamp.cpp
feichai0017: Thanks, reworked this to avoid the double-parse. The timestamp parser now first scans for the suffix boundary and then calls Time::TryConvertInterval once on the time prefix, leaving Z/offset/zone handling in the timestamp parser.

feichai0017: Thanks, reworked this to avoid the double-parse. The timestamp parser now first scans for the suffix boundary and then calls Time::TryConvertInterval once on the time prefix, leaving Z/offset/zone handling in the timestamp parser.

Mytherin:
Thanks!

PR #22923, opened 2026-05-28
Fix BIT comparison ordering for unequal lengths

### LINKED ISSUE #22527: BIT comparison with unequal lengths produces incorrect results compared to PostgreSQL
### What happens?

When comparing two BIT values of unequal lengths, DuckDB produces results that differ from PostgreSQL.

### To Reproduce

-- DuckDB
```

memory D select '0'::bit < '10101010'::bit;
┌────────────────────────────────────────────────────┐
│ (CAST('0' AS BIT(1)) < CAST('10101010' AS BIT(1))) │
│                      boolean                       │
├────────────────────────────────────────────────────┤
│ false                                              │
└────────────────────────────────────────────────────┘
```

-- PostgreSQL
```
postgres=# select '0'::bit < '10101010'::bit;
 ?column? 
----------
 t
(1 row)
```

### OS:

wasm_eh

### DuckDB Version:

v1.5.2

### DuckDB Client:

DuckDB Wasm (https://shell.duckdb.org)

### Hardware:

_No response_

### Full Name:

Yashneet Vinayak

### Affiliation:

AWS

- [x] Yes, I have

- [x] Yes, I have

Yes

feichai0017:
Fixes #22527.

## Summary

BIT values use the VARCHAR physical type internally. The first byte stores padding information and padded bits are finalized in the stored representation, so generic string comparison compares internal bytes rather than the logical bit sequence.

This PR adds a logical BIT comparator and routes BIT comparisons through it for:

- scalar/vector comparison execution
- join row matching and nested-loop/mark joins
- sort key encoding for ORDER BY
- min/max aggregates
- string statistics/zonemap pruning for BIT columns

For BIT string statistics, exact short values use logical BIT ordering. If BIT min/max stats would be truncated, min/max pruning is disabled to keep zonemap pruning conservative.

### DIFF
--- a/extension/core_functions/function_list.cpp
+++ b/extension/core_functions/function_list.cpp
@@ -120,6 +120,7 @@ static const StaticFunctionDefinition core_functions[] = {
 	DUCKDB_AGGREGATE_FUNCTION_SET(BitXorFun),
 	DUCKDB_SCALAR_FUNCTION_SET(BitStringFun),
 	DUCKDB_AGGREGATE_FUNCTION_SET(BitstringAggFun),
+	DUCKDB_SCALAR_FUNCTION(BitStringSortKeyFun),
 	DUCKDB_AGGREGATE_FUNCTION(BoolAndFun),
 	DUCKDB_AGGREGATE_FUNCTION(BoolOrFun),
 	DUCKDB_SCALAR_FUNCTION(CanCastImplicitlyFun),
--- a/extension/core_functions/include/core_functions/scalar/bit_functions.hpp
+++ b/extension/core_functions/include/core_functions/scalar/bit_functions.hpp
@@ -55,4 +55,14 @@ struct BitStringFun {
 	static ScalarFunctionSet GetFunctions();
 };
 
+struct BitStringSortKeyFun {
+	static constexpr const char *Name = "bitstring_byte_comparable";
+	static constexpr const char *Parameters = "bitstring";
+	static constexpr const char *Description = "Converts a BIT to a binary comparable sort key";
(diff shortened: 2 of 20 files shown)


Mytherin:
Thanks for the PR! This change touches many places in the code. We've solved this in the past by adding a collation to a binary comparable type - see e.g. how this is done for `TIME_TZ`. Could you try that approach instead?

feichai0017:
> Thanks for the PR! This change touches many places in the code. We've solved this in the past by adding a collation to a binary comparable type - see e.g. how this is done for `TIME_TZ`. Could you try that approach instead?

Thanks for the pointer. That makes sense - the current patch wires BIT ordering through too many execution paths. I’ll rework this to follow the TIME_TZ-style collation approach with a byte-comparable BIT sort key instead.

feichai0017:
Thanks for the pointer. I reworked the PR to follow the TIME_TZ-style collation approach.

The new version adds a `bitstring_byte_comparable(BIT) -> BLOB` helper that encodes logical bits into a binary-comparable key, and registers a BIT collation push in the binder. This lets comparisons, ORDER BY, IN/BETWEEN, joins, etc. use the existing collation machinery instead of adding BIT-specific branches across the execution paths.

I also adjusted `min/max(BIT)` to use the existing collation-based `arg_min`/`arg_max` rewrite, so aggregates follow the same logical BIT ordering while returning the original BIT value.

The previous broad BIT-specific wiring in comparison/join/sort-key/statistics paths has been removed in the follow-up commit. I kept the regression coverage and added a direct check for the byte-comparable key.

Mytherin:
Thanks!

PR #22927, opened 2026-05-28
Add support for includes and variables in sqllogic test files

Mytherin:
This is useful for cleaning up tests / avoiding a lot of duplicate set-up code, especially in extensions.

Syntax:

#### set variable
```sql
set variable my_var my_value

query I
select '{my_var}'
----
my_value
```

#### include
```sql
include test/sql/tpch/tpch_setup.test_template
```


Tishj:
I feel like I was called crazy for suggesting `include` syntax at some point :joy:

PR #22928, opened 2026-05-28
[Dev] Refactor `MultiFileReader::GetVirtualColumnExpression` and `BaseFileReader::expression_map`

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9393 and supersedes https://github.com/duckdb/duckdb-iceberg/pull/856

### Context

In `MultiFileColumnMapper::CreateFilters` we aim to push a filter into the underlying reader for a file.
To do so, we look at the Expression constructed to transform a column from the file's schema to the result schema and push this expression into the reader.
This is necessary because the filter expects the data to have the shape of the result schema, so we need to perform this transformation before the filter executes.

### Problem

The `parquet_scan`'s `ExpressionColumnReader` could only hold a single child reader, so an expression can reference at most 1 column from the local file's schema.

For a virtual column (such as iceberg's `_row_id`), this expression can reference multiple columns, which we weren't able to push down into the reader before.

### Solution

We now push these column indexes into the `expression_map`, and adjust `ExpressionColumnReader` to have an arbitrary amount of children, so we correctly gather the inputs for the expression before executing (in `ExpressionColumnReader::Read`)


Mytherin:
Thanks!

PR #22929, opened 2026-05-28
`LEFT` Join Reordering

lnkuiper:
This was supposed to be a small change: let `LEFT` joins into the join graph following "Dynamic Programming Strikes Back" and follow-up correction "On the Correct and Complete Enumeration of the Core Search Space" papers by Moerkotte et al. However, this exposed a few assumptions in the join order optimizer and cardinality estimator from @Tmonster's [master thesis](https://blobs.duckdb.org/papers/tom-ebergen-msc-thesis-join-order-optimization-with-almost-no-statistics.pdf) that were only mostly OK as long as we only reordered inner joins. This PR fixes those issues as part of enabling `LEFT` join reordering.

## Current Implementation

The current join order optimizer is built around the assumption that reorderable joins are mostly inner joins. This works well for many queries, but has a few downsides:

  1. `LEFT` joins are treated as barriers: we optimize the parents and children of the join separately.
  2. Join predicates are used for several different purposes at once: graph connectivity, equality closure, cardinality estimation domains, and plan reconstruction. This makes it really hard to separate the logic for equality and non-equality predicates.
  3. The cardinality estimator's denominator assembly is fragile for redundant transitive equality edges (`a = b AND b = c` implies `a = c`), and composite-key joins.
  4. `LEFT` joins preserve the cardinality of their left side, which confuses the basic cost model we have, because chained `LEFT` joins are all costed the same (and thus ordering doesn't seem to matter).

## New Implementation

The new implementation allows `LEFT` joins to be reordered, while keeping the right side of the `LEFT` join pinned. This means the optimizer can move independent or preserved-side joins around a `LEFT` join, but it cannot pull relations out of the nullable side.

The main changes are:

  1. A new `JoinPredicateModel` classifies predicates once and then reuses that classification throughout join ordering. This centralizes the logic around predicates.
  2. Cardinality estimation now uses the predicate model instead of rebuilding equality sets from raw filters. The denominator code handles disconnected predicate subgraphs, redundant transitive equality edges, composite same-pair equality joins, and `LEFT`/`SEMI`/`ANTI` joins.
  3. The cost model adds a small `LEFT`-specific input-work correction. Since a `LEFT` join's output cardinality preserves the left side, the cost also accounts for the optional RHS input and for inner joins that are being delayed by performing the `LEFT` join first.
  4. The plan enumerator caches query graph connection lookups and neighbor subset relation sets. This avoids doing the same graph/set work repeatedly during enumeration, which became more noticeable after adding the predicate model.

I've also done a bunch of restructuring and tidying under the `join_order` directories.

## Benchmarks

I ran TPC-DS at SF 10 with 4 threads on my laptop. These are all queries containing the word `LEFT`:
```
benchmark/tpcds/sf1/q05.benchmark
Old timing: 0.150975
New timing: 0.156879

benchmark/tpcds/sf1/q40.benchmark
Old timing: 0.022014
New timing: 0.00901

benchmark/tpcds/sf1/q49.benchmark
Old timing: 0.116231
New timing: 0.1156

benchmark/tpcds/sf1/q72.benchmark
Old timing: 0.168353
New timing: 0.174776

benchmark/tpcds/sf1/q75.benchmark
Old timing: 0.311437
New timing: 0.303004

benchmark/tpcds/sf1/q77.benchmark
Old timing: 0.10853
New timing: 0.10451

benchmark/tpcds/sf1/q78.benchmark
Old timing: 0.714756
New timing: 0.670759

benchmark/tpcds/sf1/q80.benchmark
Old timing: 0.16512
New timing: 0.135912

benchmark/tpcds/sf1/q93.benchmark
Old timing: 0.15492
New timing: 0.148666
```

With a ~2% geomean improvement on TPC-DS overall.

I also ran IMDB and it reported a 1% improvement overall. However, `scripts/plan_cost_runner.py` detects one regressed (11d) and one improved (11c) query on IMDB, so this will fail CI. We should accept this tradeoff.

## Next Steps

There are still things we can improve in follow-up PRs:

  1. Better modeling of dynamic join filters in the cost model. I experimented with this, but it did not really work yet and increased the complexity of this PR.
  2. More cleanup of the cardinality estimator. The denominator path is less brittle now, but the estimator could use some more cleanup.
  3. Try to get a small fix for correlated columns in using basic heuristics, which should help a lot with `date_dim` joins in TPC-DS.


Mytherin:
Thanks! Looks great

abramk:
@lnkuiper found a significant regression that I describe in https://github.com/duckdblabs/appian/issues/92. Would appreciate a look when you have a chance.

lnkuiper:
Thanks for testing this out! I'll have a look. We can communicate through the internal issue

PR #22933, opened 2026-05-28
Support `EXCEPT` syntax

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/9090

We now support `EXCEPT` as well as `EXCLUDE`


Mytherin:
Thanks!

PR #22934, opened 2026-05-28
Split __comparison function into dedicated functions

kryonix:
Comparisons were unified with regular functions in https://github.com/duckdb/duckdb/pull/22514. However, instead of using dedicated comparison scalar functions (for `<`, `>`, `=` etc.), one convenience function `__compare` was used. This PR keeps the unification, but breaks up the `__comparison` function into individual cases.


Mytherin:
Nice! Can we remove `__comparison` entirely now?

kryonix:
I think so, just wanted a CI run to verify before doing the full removal.

Mytherin:
Thanks!

PR #22935, opened 2026-05-28
Verify CLI major version is non-zero on main

smvv:
This PR can help investigate https://github.com/duckdblabs/duckdb-internal/issues/9461.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/9461.

### DIFF
--- a/.github/workflows/Main.yml
+++ b/.github/workflows/Main.yml
@@ -627,7 +627,7 @@ jobs:
       shell: bash
       run: |
         ./build/release/duckdb -c "PRAGMA platform; PRAGMA version;"
-        if [ "${{ github.ref_name }}" == "main" ]; then
+        if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
           # Fail on non-zero major version.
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

.github/workflows/Main.yml
```diff
@@ -623,8 +625,14 @@ jobs:
           make -C \"$PWD\"
         "
 
-    - name: Print platform
-      run: ./build/release/duckdb -c "PRAGMA platform;"
+    - name: CLI check
+      shell: bash
+      run: |
+        ./build/release/duckdb -c "PRAGMA platform; PRAGMA version;"
+        if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
```
carlopi, line 632: Curiosity: why only on workflow_dispatch?

smvv: i used workflow_dispatch (= release + nightly + lunch CI runs) so it guards on those CI runs.

Perhaps we can expand it to other CI runs, but was not sure if they would all have a valid version (in fork, merge queue jobs, pr jobs, etc).

Mytherin:
Thanks!

Issue #16436, opened 2025-02-27
Self-referential Foreign Key Causes Duplicate Primary Key Error on UPDATE


PcdPanda:
### What happens?

When using a self-referential foreign key in DuckDB, updating the foreign key column to reference the same row causes a Duplicate key "sid: 1" violates primary key constraint error. This behavior is unexpected since the row already exists and should be able to reference itself without violating the primary key constraint. Seems DuckDB is trying to insert a new row.

Expected Behavior:
The UPDATE statement should succeed, allowing the row to reference itself via the foreign key.

Actual Behavior:
DuckDB throws the following error:

```
Constraint Error: Duplicate key "sid: 1" violates primary key constraint.
If this is an unexpected constraint violation please double check with the known index limitations section in our documentation (https://duckdb.org/docs/sql/indexes).
[SQL: UPDATE sid SET currency_sid = 1 WHERE sid = 1;]
```

### To Reproduce

Steps to Reproduce:
```
CREATE TABLE sid (
    sid INTEGER PRIMARY KEY,
    sid_name VARCHAR NOT NULL UNIQUE,
    currency_sid INTEGER,
    FOREIGN KEY (currency_sid) REFERENCES sid(sid)
);

INSERT INTO sid (sid, sid_name) VALUES (1, 'TEST_SID');
```

```
UPDATE sid SET currency_sid = 1 WHERE sid = 1;  # Error Here
```

### OS:

Ubuntu 22.04.4 LTS (GNU/Linux 5.15.167.4-microsoft-standard-WSL2 x86_64)

### DuckDB Version:

1.2.0

### DuckDB Client:

Python

### Full Name:

Chongdan Pan

### Affiliation:

University of Michigan

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have not tested with any build

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

szarnyasg:
Hi @PcdPanda, thanks for reporting this issue. This is a known limitation: https://duckdb.org/docs/sql/statements/create_table.html#limitations

> Inserting into tables with self-referencing foreign keys is currently not supported and will result in the following error:
[...]

PR #22937, opened 2026-05-28
Fix spurious foreign key violation when updating non-PK columns of a referenced table

### LINKED ISSUE #16409: Insert or update doesn't work with arrays
### What happens?

When doing an `update` or `insert or replace` on a primary table, if there is an array column is in the updated fields, it fails with a foreign key error.

Using v1.2.0 5f5512b827

### To Reproduce

```
create table test (id int primary key, arr string[]); 
insert into test values (1, '[asdf, 2e3f]');
create table test_fk (test_id int references test (id), val string);
insert into test_fk values (1, 'blkaj');

insert or replace into test values (1, '[lkjsdf]');
update test set arr = [] where id = 1;
insert into test values (1, '[lkjsdf]') on conflict (id) do update set arr = excluded.arr;
insert into test values (1, '[lkjsdf]') on conflict do update set arr = excluded.arr;

```

Output: 

```
8:03:59 ~ $ duckdb
v1.2.0 5f5512b827
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
D 
D create table test (id int primary key, arr string[]); 
D insert into test values (1, '[asdf, 2e3f]');
D create table test_fk (test_id int references test (id), val string);
D insert into test_fk values (1, 'blkaj');
D 
D insert or replace into test values (1, '[lkjsdf]');
Constraint Error:
Violates foreign key constraint because key "test_id: 1" is still referenced by a foreign key in a different table
D update test set arr = [] where id = 1;
Constraint Error:
Violates foreign key constraint because key "test_id: 1" is still referenced by a foreign key in a different table
D insert into test values (1, '[lkjsdf]') on conflict (id) do update set arr = excluded.arr;
Constraint Error:
Violates foreign key constraint because key "test_id: 1" is still referenced by a foreign key in a different table
D insert into test values (1, '[lkjsdf]') on conflict do update set arr = excluded.arr;
Constraint Error:
Violates foreign key constraint because key "test_id: 1" is still referenced by a foreign key in a different table
D 
D select * from test  join test_fk on id = test_id;
┌───────┬──────────────┬─────────┬─────────┐
│  id   │     arr      │ test_id │   val   │
│ int32 │  varchar[]   │  int32  │ varchar │
├───────┼──────────────┼─────────┼─────────┤
│   1   │ [asdf, 2e3f] │    1    │ blkaj   │
└───────┴──────────────┴─────────┴─────────┘
D 
```

### OS:

MacOS 15.2 (24C101) (arm64, M4)

### DuckDB Version:

v1.2.0 5f5512b827

### DuckDB Client:

CLI, Python

### Hardware:

_No response_

### Full Name:

Jeffrey Melloy

### Affiliation:

None

Yes

- [x] Yes, I have

- [x] Yes, I have

berges99:
Fixes #16409.

Updating a LIST/ARRAY/STRUCT/MAP column on a table referenced by a foreign key can raise a foreign-key violation even when the referenced key is unchanged. The update is rewritten as DELETE + INSERT, and the transient delete incorrectly checks whether the unchanged key is still referenced.

This change skips the delete-side check only when all referenced key columns are disjoint from the actual update assignments and none of the foreign keys is self-referential. Insert-side checks and checks for real deletes or referenced-key changes remain in place. The same handling applies to UPDATE, MERGE, INSERT OR REPLACE, and ON CONFLICT DO UPDATE.

The assignment set is captured before CHECK constraints and RETURNING add unchanged columns to the projection. Self-referential foreign keys (#16436) remain outside this change.

Updated against current main, preserving its UPDATE/MERGE behavior and assigning the LogicalUpdate flag serialization ID 210 to avoid the new upstream fields.

Validation:
- Built the merged branch and test extension fixtures locally.
- Expanded regression test: 45 assertions passed, including RETURNING, CHECK constraints, explicit MERGE, and rejection of genuine key changes.
- Constraint/UPDATE tests: 99 test cases passed.
- MERGE/upsert tests: 74 test cases passed.
- Foreign-key tests passed with logical-plan serialization verification and with persistent storage (28 test cases per configuration).
- Changed-file formatting and whitespace checks passed.

- Full local unit suite: 6,933 passed, 451 skipped by test directives or unavailable extension/platform requirements (535 seconds).

[Fork CI](https://github.com/berges99/duckdb/actions/runs/35353131950) is running; its preparation and formatting gate passed.

### DIFF
--- a/src/catalog/catalog_entry/table_catalog_entry.cpp
+++ b/src/catalog/catalog_entry/table_catalog_entry.cpp
@@ -11,6 +11,7 @@
 #include "duckdb/parser/parsed_data/create_table_info.hpp"
 #include "duckdb/planner/binder.hpp"
 #include "duckdb/planner/constraints/bound_check_constraint.hpp"
+#include "duckdb/planner/constraints/bound_foreign_key_constraint.hpp"
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/operator/logical_get.hpp"
 #include "duckdb/planner/operator/logical_projection.hpp"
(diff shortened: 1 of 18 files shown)


dentiny:
Guess this is still a problem? should we re-open it?

berges99:
Yes, this still reproduces on 1.5.5 and 2.0.0.dev2609121639. All four variants from #16409 raise the same foreign-key violation when updating the list column without changing the primary key. Both the issue and PR were auto-closed for inactivity. Please reopen this PR and #16409.

PR #22940, opened 2026-05-28
Rowgroup index append

artjomPlaunov:
@evertlammerts this supersedes https://github.com/duckdb/duckdb/pull/22796, just rebased to 1.5 + uses attached database instead, so should be basically the same review wise :-)


evertlammerts (approved):
LGTM! Same as before :)

PR #22941, opened 2026-05-28
Unify named arguments + default arguments for `(Scalar/Aggregate/Window)Functions`

Maxxen:
This PR unifies the behavior of passing function arguments by name for `(Scalar/Aggregate/Window)Functions`, giving them similar capabilities as user-defined macro functions.

This is done by extending the `FunctionSignature` to not just support per-parameter names, but also "default values". This avoids the very cumbersome situation where we previously had to define multiple, almost identical overloads for functions that want to provide "optional" arguments, and/or define complex `Bind` callbacks that tried to inspect the aliases of trailing vararg expressions to figure out how to resolve these optional parameters.

As an example, this PR updates the `json_serialize_sql` function in the `json` extension to no longer require 5 different overloads, but instead be defined by a single `FunctionSignature` (we still need to keep the bind callback for backwards compatibility though).

Having this be part of the signature also opens the door for providing better auto-generated documentation. E.g. instead of looking at:
```
select function_name, parameter_types from duckdb_functions() where function_name = 'json_serialize_sql';
┌────────────────────┬───────────────────────────────────────────────┐
│   function_name    │                parameter_types                │
│      varchar       │                   varchar[]                   │
├────────────────────┼───────────────────────────────────────────────┤
│ json_serialize_sql │ [VARCHAR]                                     │
│ json_serialize_sql │ [VARCHAR, BOOLEAN]                            │
│ json_serialize_sql │ [VARCHAR, BOOLEAN, BOOLEAN]                   │
│ json_serialize_sql │ [VARCHAR, BOOLEAN, BOOLEAN, BOOLEAN]          │
│ json_serialize_sql │ [VARCHAR, BOOLEAN, BOOLEAN, BOOLEAN, BOOLEAN] │
└────────────────────┴───────────────────────────────────────────────┘
```

And trying to figure out how the function works, we could (in the future) instead give you a much more informative signature:
```sql
json_serialize_sql(
	sql VARCHAR, 
	skip_null BOOLEAN := false, 
	skip_empty BOOLEAN := false,
	skip_default BOOLEAN := false,
	format BOOLEAN := false
) -> VARCHAR
```

### Argument Resolution

Im not going to elaborate too much on all the details of how function arguments are resolved because I think most of us who have programmed in any language that supports optional arguments have a pretty good intuition for how it works, but its basically the same as for user-defined macros.

The only difference is handling of "varargs". Since we don't have a way to define a "parameter pack", or python-style `*args` or `**kwargs`, any named arguments that do not match a parameter name will be appended in order as a trailing argument, for functions that support varargs, just like normal positional arguments work.

### Backwards compatability

Before this PR we didn't _really_ support named _arguments_ (even for macros), we just relied on expression aliases to kind of fake a named parameter assignment. In my initial iteration of this PR I thought this would continue to work well, but unfortunately it turns out that we create a lot of aliased expressions "by accident" through various rewrite rules or optimizations. This is essentially a bug that breaks macros that has been around since macros was first added, but didn't really become apparent until the initial version of this PR made _every_ function became alias-sensitive.

To fix this I've introduced a new `FunctionArgument` struct as part of the parsed `FunctionExpression`, and the parser now emits this struct when encountering the `:=` or `=>` syntax. This allows us to differentiate between ambient expression aliases, and _intentional_ argument assignment by the user. 

In order to not break backwards compatibility:

- When serializing to an old format, extract all expressions from the `FunctionArgument`s and alias them by the argument name (if passed by name).

This ensures that old functions that were "alias sensitive", e.g. `struct_pack`, `json_serialize_sql` or old macros get any named arguments with aliases, as they expect. Other functions simply ignore aliases.

- When deserializing an old `FunctionExpression`, _dont_ interpret the aliases as argument names, but keep them around, and mark this function expression as a "legacy function call". 

Once we reach the binder we simply pass all arguments positionally. "alias" sensitive functions can then keep their old behavior of inspecting the aliases to resolve named args, but new functions don't get any accidental named args.

### Future work

- Window functions currently error out in the parser when passed a named argument. This should be relatively easy to support since they now also use `FunctionSignature`s and should be able to go through the same path in the `FunctionBinder`, but I've left it as a TODO for now as this PR is getting big enough already and I dont really know how to write a good test for these as we only support a limited number of window functions.

- Unify with `(Scalar/Table)Macros`. These have their own binding logic and named parameter, but it's basically identical to how `FunctionSignature`s work and so should ideally go through the same path too. Could probably remove a lot of code.

- `FunctionSignature` for table/pragma functions. These have their own branch in the inheritance hierarchy just cause they used to have special handling for named parameters, but now that named/default parameters are part of the function signature these could use the same infrastructure as well.

- "Parameter Packs" / `*args` and `**kwargs`. This could now be specified in the `FunctionSignature` and handled generically by the binder as well and would make it much easier to write variadic functions IMO.

- Argument spread syntax (e.g. `select foo(bar...)` to automatically unpack structs/lists as function arguments. This, and other sort of per-argument modifiers is easier to do now that we have a `FunctionArgument` struct.


src/common/util/util_parsed_expression.cpp
```diff
@@ -638,15 +643,15 @@ unique_ptr<ParsedExpression> FunctionExpression::Copy() const {
...
-	for (auto &child : children) {
-		copy->children.push_back(child->Copy());
-	}
 	copy->filter = filter ? filter->Copy() : nullptr;
 	copy->order_bys = order_bys ? unique_ptr_cast<ResultModifier, OrderModifier>(order_bys->Copy()) : nullptr;
 	copy->distinct = distinct;
 	copy->is_operator = is_operator;
 	copy->export_state = export_state;
 	copy->catalog = catalog;
+	for (auto &arg : arguments) {
```
Mytherin, line 637: Should we also copy `is_legacy_function_call`?

src/include/duckdb/function/function.hpp
```diff
@@ -178,24 +197,64 @@ class FunctionSignature {
...
 
-	auto AddParameter(LogicalType type) -> void {
-		auto name = parameters.empty() ? string("col") : string("col") + to_string(parameters.size() + 1);
+	auto AddParameter(string name, LogicalType type) -> FunctionSignature & {
+		parameters.emplace_back(std::move(name), std::move(type));
+		return *this;
+	}
 
+	auto AddParameter(LogicalType type) -> FunctionSignature & {
+		auto name = parameters.empty() ? string("col") : string("col") + to_string(parameters.size());
```
Mytherin: This went from `parameters.size() + 1` to `parameters.size()` - maybe we should keep these consistent with the old code?

Mytherin (commented):
Thanks! Looks great - some minor comments

hawkfish:
@Maxxen I haven't looked at this too closely, but I think it would be really nice if `LEAD` and `LAG` supported this. Right now the most common second argument (offset) is `1` but the third argument (default) is often what I want to modify e.g., when converting an event table to a state table:

```sql
create view states as
select
    instrument,
    event, 
    when as begin, 
    lead(when, 1, 'infinity') over (partition by instrument order by when) as end,
from events
```

Maxxen:
@hawkfish I gotta fix some merge conflicts anyway so I'll see if I can add support for windows too, if I can test the implementation with lead/lag then it should not be too hard!

rustyconover:
Named arguments for scalar functions with defaults is awesome!

Maxxen:
@Mytherin This should be good to go now - I ended up leaving out the window functions after all but will get to it asap in a follow up.

Mytherin:
Thanks!

PR #22944, opened 2026-05-28
Fix hang in PartitionedCopyState with regression test

smvv:
### Test timeouts on main

Multiple timeouts happened for the [same](https://github.com/duckdb/duckdb/actions/runs/26582773380/job/78321912631#step:8:321) [test](https://github.com/duckdb/duckdb/actions/runs/26582773380/job/78321912631#step:8:321) [in](https://github.com/duckdb/duckdb/actions/runs/26580988638/job/78318138359#step:8:263) different CI runs when running tests with `--test-config test/configs/force_external.json`:
```shell
  batch timed out after 600 seconds
  
  === stdout ===
  
  [0/10] (0%): test/sql/copy/partitioned/hive_partitioned_write.test
  [1/10] (10%): test/sql/copy/partitioned/hive_partition_case_insensitive_column.test
  [2/10] (20%): test/sql/copy/partitioned/hive_partition_append.test
  [3/10] (30%): test/sql/copy/partitioned/partitioned_window.test
  [4/10] (40%): test/sql/copy/partitioned/hive_partition_join_pushdown.test
  [5/10] (50%): test/sql/copy/partitioned/partition_issue_6304.test
  [6/10] (60%): test/sql/copy/partitioned/partitioned_order_by.test  # <-- hangs
```

I noticed the timeout because the CI job `linux-release-configs` on branch `main` started to take much longer in namespace's job analytics:

Screenshot 2026-05-28 at 21 36 21

### Stacktrace when hanging

I reproduced the hang locally (it happened a few times per ~10 test runs) for a release build on my laptop:
```shell
GEN=ninja DUCKDB_EXTENSIONS='icu;tpch;tpcds;fts;json;inet;parquet;autocomplete;httpfs' \
DISABLE_SANITIZER=1 BUILD_BENCHMARK=1 BUILD_JEMALLOC=1 \
make clean release VCPKG_TOOLCHAIN_PATH="$PWD/vcpkg/scripts/buildsystems/vcpkg.cmake";
```

I asked Codex to create a reproducing test `test/sql/copy/partitioned/partitioned_order_by_flush_race.test` that hanged without any code changes.

Next, I asked Codex to use `sample` on macOS on the hung `unittest` process. It showed threads spending most time here:

- `PartitionedCopyFinalizeTask::ExecuteTask`
- `PartitionedCopy::Flush`
- heavy `pthread_mutex_lock` / `__psynch_mutexwait`

### Timeout cause

The timeout was caused by a race during **partitioned COPY flush/finalize**.

In some runs, `Flush()` was called while `flushing_state->global_source_state` was still null.
At that moment, the code just returned early with “nothing to do”.

The problem: if combine had already finished (`combined == locals`), that state actually *was* ready to be finalized, but we skipped finalization and kept looping. Under `force_external`, this could repeat and eventually hit the timeout (`600s`!).

### Fix

In `PartitionedCopy::Flush`, when `global_source_state` is null:

1. If combine is **not** done yet, return as before.
2. If combine **is** done (`combined == locals`), call `FinalizeState(...)` immediately.
3. Then continue normal flush work.

So instead of waiting for some other path to finalize, `Flush()` now finalizes lazily as soon as it has enough information, preventing the stuck loop.

### DIFF
--- a/src/execution/operator/persistent/physical_copy_to_file.cpp
+++ b/src/execution/operator/persistent/physical_copy_to_file.cpp
@@ -360,6 +360,7 @@ public:
 
 public:
 	bool ShouldInitiateFlush(const idx_t &local_append_count);
+	bool IsCombineComplete() const DUCKDB_REQUIRES(lock);
 	void CreateTaskList() DUCKDB_REQUIRES(lock);
 	bool HasCompleted() const;
 	optional<PartitionedCopyTask> TryAssignTask();
@@ -843,6 +844,10 @@ bool PartitionedCopyState::ShouldInitiateFlush(const idx_t &local_append_count) 
 	return total_count / unique_count >= Settings::Get<PartitionedWriteFlushThresholdSetting>(partitioned_copy.context);
 }
 
+bool PartitionedCopyState::IsCombineComplete() const {
+	return combined == locals;
+}
+
 void PartitionedCopyState::CreateTaskList() {
 	global_source_state =
 	    partitioned_copy.sort_strategy->GetGlobalSourceState(partitioned_copy.context, *global_sink_state);
@@ -913,7 +918,6 @@ optional<PartitionedCopyTask> PartitionedCopyState::TryAssignTask() {
 	while (next_group < partition_blocks.size()) {
 		const auto group_idx = partition_blocks[next_group++].second;
 		active_groups.emplace_back(group_idx);
-
 		auto &hash_group = hash_groups[group_idx];
 		annotated_lock_guard<annotated_mutex> guard(hash_group->lock);
 		hash_group->TryPrepareNextStage();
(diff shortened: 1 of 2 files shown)


lnkuiper:
Looks great! Thanks. Only thing I think might need changing is that the `.test` file is renamed to `.test_slow` if it proves to be too slow in the future, but this is good to merge for now.

smvv:
> Looks great! Thanks. Only thing I think might need changing is that the `.test` file is renamed to `.test_slow` if it proves to be too slow in the future, but this is good to merge for now.

@lnkuiper i looked it up and the test takes:
```
$ time ./build/relassert/test/unittest --test-config test/configs/force_external.json test/sql/copy/partitioned/partitioned_order_by_flush_race.test 
23.27s user 1.85s system 272% cpu 9.212 total

$ time ./build/release/test/unittest --test-config test/configs/force_external.json test/sql/copy/partitioned/partitioned_order_by_flush_race.test 
2.82s user 1.11s system 260% cpu 1.505 total
```
so I think that seems good enough.

PR #22949, opened 2026-05-29
Ensure ROW_GROUP_PRUNER considers deleted columns when processing limit

tianjq16:
When processing the LIMIT clause, `GetQualifyingTupleCount` is invoked to determine the number of rows within a RowGroup. However, within the `GetQualifyingTupleCount` function, deleted rows are not taken into account; consequently, certain groups may be incorrectly pruned.


Mytherin:
Thanks - but this has been solved in a better manner in https://github.com/duckdb/duckdb/pull/22911

PR #22950, opened 2026-05-29
`CONNECT`: CLI prompt, implement for `postgres` and add `Supports(RemoteCapability::<X>)`

carlopi:
Follow up on https://github.com/duckdb/duckdb/pull/22878

#### CLI
* Modify default prompt to encode connected database, connection string defaults to attached db name, but overridable per catalog
* fix a but on display length calculations
* First Ctrl-D while connected implies `DISCONNECT`

CLI: prompt generation code currently assumes each segment is computed as a separate query. To be considered: constructing a single query, that would allow 1-shot remote query even for arbitrary promtps.
Currently, it's still 1-shot (to determine current_database + current_schema), but for arbitrary prompts it could be N components.
To opt-out, provide custom prompt would be the path.

#### Infrastructure
* Change from `IsRemoteCatalog` to `Supports(RemoteCapability::<X>)`, to allow distinguish specific capabilities 

#### Postgres
* enable connect-to `postgres`, but do not allow (yet) RemotePushdown

Postgres allows to show more clearly that syntax is sent as literals

#### Quack
* Bump version, now extra function available

----

Examples:
image
image


carlopi:
Failure is a timeout. Can be restarted, but should be independent.

carlopi:
Test failure is not on this PR, I think.

tools/shell/shell_prompt.cpp
```diff
@@ -249,6 +251,30 @@ string Prompt::HandleSetting(ShellState &state, const PromptComponent &component
...
+			return ExecuteSQL(state, "SELECT current_schema()");
+		}
+		if (component.literal == "current_database_and_schema") {
+			return ExecuteSQL(state, "SELECT CASE WHEN current_schema() = 'main' THEN current_database() "
+			                         "ELSE current_database() || '.' || current_schema() END");
+		}
+		if (component.literal == "connect_suffix") {
+			auto connected = context.TryGetConnectedCatalog();
+			if (connected) {
+				return " @ " + connected->GetCatalog().GetConnectDisplay();
```
Mytherin: I think this should be a prefix - not a suffix (i.e. `quack:localhost`).

Mytherin: `@` is usually used to indicate username so I think that is confusing here. I would vote for using a different color, maybe green?

Image

We can get designer input as well here.

.github/patches/extensions/postgres_scanner/0004-connect.patch
```diff
@@ -0,0 +1,70 @@
...
++unique_ptr<TableRef> PostgresCatalog::RemoteExecute(ClientContext &context, const string &sql) {
++	vector<unique_ptr<ParsedExpression>> args;
++	args.push_back(make_uniq<ConstantExpression>(Value(GetName())));
++	args.push_back(make_uniq<ConstantExpression>(Value(sql)));
++	auto func_ref = make_uniq<TableFunctionRef>();
++	func_ref->function = make_uniq<FunctionExpression>("postgres_query", std::move(args));
++	return func_ref;
++}
++
++string PostgresCatalog::GetConnectDisplay() {
```
Mytherin, line 41: Maybe this should just be the attached database name, instead of the full query path / URI?

Mytherin: We can split this into `connected_database_name` and `connected_database_uri`, rename this function to "GetURI", then by default display the (shorter) `connected_database_name`.

I think we should avoid the default prompt becoming very large.

Mytherin (commented):
Thanks! Looks good - left some comments

tools/shell/shell_prompt.cpp
carlopi: I changed to come first, and it has now it's own `{highlight_element:prompt_connect}{setting:connect_name_prefix}` highlighting schema. I am not sure on green, I went for same orange but not bold, unsure, but now there are easy places where to plug changes (with limited C++ knowledge).

.github/patches/extensions/postgres_scanner/0004-connect.patch
carlopi: I made so there are now three possible prompt blocks:
* `{highlight_element:prompt_connect}` controls the highlighting for connect
* `{setting:connect_name_prefix}` renders the attached database name (so user controlled)
* `{setting:connect_display_prefix}` renders the name as reported by the catalog, so overridable (or the database name, if not provided)

Default is now:
```c++
      default_prompt = "{max_length:50}{highlight_element:prompt_connect}{setting:connect_name_prefix}{highlight_element:"
                       "prompt}{setting:current_database_and_schema}"
                       "{color:reset} D ";
```
that is via `connect_name_prefix`, but it's pluggable by providing a different prompt.

carlopi:
Updated thanks to review comment. PR message now with screenshots of what it would look like light/dark mode.

Unsure on rendering style, but that can be fixed up later.

Mytherin:
Thanks - looks good. Agreed we can fix up rendering later on.

PR #22953, opened 2026-05-29
Make url a value, not a const ref

NiclasHaderer:
BaseRequest has a const ref string as member making it very easy to create lifecycle issues


MBkkt:
btw such issue already exists in httpfs: https://github.com/duckdb/duckdb-httpfs/pull/321

Mytherin:
Thanks!

PR #22954, opened 2026-05-29
Fix SetVectorType for VectorStructBuffer

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/9258

`VectorStructBuffer` is initialized as a `FLAT_VECTOR` but its child vectors may be dictionary vectors or others. However, there are some places where we want to set a vector type to `CONSTANT_VECTOR` and assume it is allowed because the current vector type is flat. That is not necessarily true for nested data types, specifically `STRUCT`, where this results in an error if any of the children is not flat and not constant.


Mytherin:
Thanks!

PR #22955, opened 2026-05-29
Make division and average functions return `DECIMAL` instead of `DOUBLE`

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/8441
Fix: https://github.com/duckdblabs/duckdb-internal/issues/8442

This PR changes the behaviour for division (`/`) and the `avg` function to return `DECIMAL` rather `DOUBLE` when the input types are also `DECIMAL`. 

For division, the resulting `scale` and `width` are calculated by the following formula (also used by SQL Server): 
For `a / b` with input types `DECIMAL(p1, s1)` and `DECIMAL(p2, s2)`:
- Result scale: `max(6, s1 + p2 + 1)`
- Result width: `p1 - s1 + s2 + result_scale`

For `avg()`: 
- Result scale: same as the input scale
- Result width: same as the input width

## Compatibility setting
I've added a setting `decimal_arithmetic`, that can be set to either `DOUBLE` or `DECIMAL` (default), to control the return types for `/` and `avg`. Setting it to `DOUBLE` ensures the old behaviour, while `DECIMAL` allows for the new behaviour: 
```sql
SET decimal_arithmetic = 'double';
```

In tests that relied on the old behaviour I have added the setting set to `double` to ensure the old behaviour remains. 

I think the name of the setting can be improved, but I honestly wasn't sure what's best. 

cc @taniabogatsch


lnkuiper:
Hi @Mytherin just want to get your take on this change, since it is a breaking change. This makes decimal division/avg return decimal instead of double.

PG returns double for division (but we're not following them anyway since they also do `integer / integer -> integer`), and the python decimal package returns decimal (but they can do a different scale per division, while we have to decide per column).

I think you can argue both sides, we just have to pick a side on what the default setting will be.

Dtenwolde:
I think a gradual transition would make sense. First we make it opt-in (add the setting, but default would be `DOUBLE`), then the next release we switch the default to `DECIMAL`.

src/function/scalar/operator/arithmetic.cpp
```diff
@@ -1109,13 +1110,180 @@ unique_ptr<FunctionData> BindBinaryFloatingPoint(BindScalarFunctionInput &input)
...
+	arguments[0]->GetReturnType().GetDecimalProperties(p1, s1);
+	arguments[1]->GetReturnType().GetDecimalProperties(p2, s2);
+
+	uint8_t result_scale = MaxValue<uint8_t>(6, s1 + p2 + 1);
+	uint8_t result_width = p1 - s1 + s2 + result_scale;
+
+	// Cap to the maximum representable decimal width (38). When the
+	// formula yields a width above 38 we reduce the scale by the same
+	// amount so the integer part is preserved, then floor scale at 6.
+	if (result_width > Decimal::MAX_WIDTH_DECIMAL) {
```
Mytherin, line 1211: For multiplication we throw an error in this case (stating we can't accurately represent the result) - maybe we should do that here as well?

Mytherin: See [here](https://github.com/duckdb/duckdb/blob/main/src/function/scalar/operator/arithmetic.cpp#L888)

src/function/scalar/operator/arithmetic.cpp
```diff
@@ -1109,13 +1110,180 @@ unique_ptr<FunctionData> BindBinaryFloatingPoint(BindScalarFunctionInput &input)
...
+		    hugeint_t numerator = hugeint_t(a) * scale_factor;
+		    hugeint_t divisor = hugeint_t(b);
+
+		    // Work with absolute values so the rounding logic is sign-agnostic.
+		    bool negative = (numerator.upper < 0) != (divisor.upper < 0);
+		    hugeint_t abs_num = numerator.upper < 0 ? -numerator : numerator;
+		    hugeint_t abs_div = divisor.upper < 0 ? -divisor : divisor;
+
+		    hugeint_t q, r;
+		    if (abs_div.upper == 0) {
```
Mytherin, line 1153: No matter the types we're now always doing (slow) hugeint ops - that seems sub-optimal especially when decimals will in many cases not be hugeints. Can we rework this?

src/function/scalar/operator/arithmetic.cpp
```diff
@@ -1109,13 +1110,180 @@ unique_ptr<FunctionData> BindBinaryFloatingPoint(BindScalarFunctionInput &input)
...
+
+		// Guard against uint8_t underflow: excess can be larger than
+		// result_scale (e.g. p1=38,s1=0,s2=5 gives excess=26, scale=21).
+		// In that case clamp to 0 before applying the minimum-scale floor.
+		if (excess <= result_scale) {
+			result_scale -= excess;
+		} else {
+			result_scale = 0;
+		}
+		result_scale = MaxValue<uint8_t>(6, result_scale);
```
Mytherin, line 1222: The `6` here as minimum scale seems kind of arbitrary - this should probably be configurable

Mytherin (commented):
Thanks for the PR!

The main issue with decimal division, and why this is challenging, is that the result scale is not dependent on the input scale. Similarly, the result *might not be accurately representable at all* in a decimal number. This is different from multiplication / addition in which we can always accurately represent the result in a given output scale given the input scales.

This new division method can have better or worse accuracy than the current method. It is not necessarily more accurate than floating point division. For example:

```sql
memory D select 10.0 / 3.0;
┌──────────────┐
│ (10.0 / 3.0) │
│ decimal(9,6) │
├──────────────┤
│     3.333333 │
└──────────────┘
```

Versus:

```sql
memory D select 10.0 / 3.0;
┌────────────────────┐
│    (10.0 / 3.0)    │
│       double       │
├────────────────────┤
│ 3.3333333333333335 │
└────────────────────┘
```

`3.3333333333333335` is closer to the "correct" value than `3.3333330` (the correct value being `3.33333333333333333..`).

The current implementation is more accurate here than this new one. Because this is not always better than the current implementation I don't think it's worthwhile to replace the default division operator and going through the steps of breaking backwards compatibility. I would propose not to have this replace the existing division method, but to instead have a separate division method (`decimal_division`). That method can then also takes extra parameters (like e.g. the minimum scale, which is now hard-coded to 6). 

If we want to do true accurate division - what is done in more mathematically oriented languages is to have the division operator return a `numerator / denominator` pair. This can then correctly handle any divisions between decimal numbers. That's a much bigger rework than just implementing this new decimal division method, though, and likely belongs in an extension more than in the core system.

Dtenwolde:
Superseded by https://github.com/duckdb/duckdb/pull/23097 and https://github.com/duckdb/duckdb/pull/23069

PR #22958, opened 2026-05-29
testing: add new fuzz harness to be consumed by OSS-Fuzz

DavidKorczynski:
Adds a harness targeting the parquet stack. The main goal is to start adding more fuzzers that will have impact on the  code coverage, for which there is a reason report here: https://storage.googleapis.com/oss-fuzz-coverage/duckdb/reports/20260528/linux/src/report.html


Mytherin:
Thanks!

PR #22959, opened 2026-05-29
Internal #9434: ENUM IN STRING_LITERAL

hawkfish:
* Instead of casting the ENUM to a VARCHAR, cast the literals to ENUMs.
* Test case gets about 10x faster with this change.


src/planner/binder/expression/bind_operator_expression.cpp
```diff
@@ -28,9 +25,48 @@ LogicalType ExpressionBinder::ResolveCoalesceType(OperatorExpression &op, vector
 	LogicalType max_type = ExpressionBinder::GetExpressionReturnType(*children[0]);
 	bool is_in_operator = (op.GetExpressionType() == ExpressionType::COMPARE_IN ||
 	                       op.GetExpressionType() == ExpressionType::COMPARE_NOT_IN);
+
+	// Look up the domain for ENUM (NOT) IN STRING_LITERAL
+	vector<unique_ptr<Expression>> in_children;
+	string_map_t<uint64_t> enum_domain;
```
Mytherin: The planner / binder is the wrong place for this imo, maybe we can write an optimizer rule to detect this scenario instead?

Mytherin: I think the pattern would be something like `CAST(enum_col AS VARCHAR) IN (<varchar list>)`

Mytherin (commented):
Thanks! Good idea - I just think this should be implemented as an optimizer rule. Could you also add the test case and/or benchmark you mentioned?

lnkuiper:
Thanks! Looks great. Really nice that we remove invalid enum values in the optimizer already. I think there is an edge case where `x IN ('invalid-enum-value-1', 'invalid-enum-value-2')` will still fall back to the slow string-based path even though we can just rewrite to `false`, but that's not too important, and could be fixed in a future PR :)

hawkfish:
> Thanks! Looks great. Really nice that we remove invalid enum values in the optimizer already. I think there is an edge case where `x IN ('invalid-enum-value-1', 'invalid-enum-value-2')` will still fall back to the slow string-based path even though we can just rewrite to `false`, but that's not too important, and could be fixed in a future PR :)

Yes - or `true` for `NOT IN`. Should be simple enough - I'll make another issue.

PR #22961, opened 2026-05-30
Fix external file cache disablement

dentiny:
Hi team, I found even when external file cache is disabled, we still populate the filepath in the `cached_files` map [here](https://github.com/duckdb/duckdb/blob/bbd990d55458620be783f90da2b3d92ace15c834/src/storage/external_file_cache/external_file_cache.cpp#L245-L252), which definitely doesn't match user expectation.

In this PR, no entries will be inserted into the map when cache disabled. To keep change set little, I decided to store `shared_ptr<CachedFile>` in the map, instead of a unique pointer so it could be returned to caching filesystem as well.

Another reason I think it a proper fix: I'm reworking https://github.com/duckdb/duckdb/pull/22857, a big challenge is to reason about the `CachedFile` ownership when external file cache disabled, thus clearing the whole map. Shared pointer makes implementation 10x easy without hurting clarity on `CachedFile` ownership.

v1.5 backport plan: since I rewrote the external file cache in latest main, it's quite different from what it is in v1.5, so if needed, I will backport two fix PRs together to v1.5, after they've been merged into main.

### DIFF
--- a/src/include/duckdb/storage/external_file_cache/caching_file_system.hpp
+++ b/src/include/duckdb/storage/external_file_cache/caching_file_system.hpp
@@ -36,7 +36,7 @@ public:
 
 public:
 	DUCKDB_API CachingFileHandle(QueryContext context, CachingFileSystem &caching_file_system, const OpenFileInfo &path,
-	                             FileOpenFlags flags, optional_ptr<FileOpener> opener, CachedFile &cached_file);
+	                             FileOpenFlags flags, optional_ptr<FileOpener> opener);
 	DUCKDB_API ~CachingFileHandle();
 
 public:
@@ -78,8 +78,8 @@ private:
 	optional_ptr<FileOpener> opener;
 	//! Cache validation mode for this file
 	CacheValidationMode validate;
-	//! The associated CachedFile with cached blocks
-	CachedFile &cached_file;
+	//! Associated cached file.
+	shared_ptr<CachedFile> cached_file;
 
 	//! Used to ensure file handle and cached file metadata is only initialized once.
 	annotated_mutex file_handle_mutex;
--- a/src/include/duckdb/storage/external_file_cache/caching_file_system.hpp
+++ b/src/include/duckdb/storage/external_file_cache/caching_file_system.hpp
@@ -62,6 +62,8 @@ public:
 private:
 	//! Remove the 'force_full_download' option from the file handle if present, and return whether it was present
 	bool StripForceFullDownloadIfPresent();
+	//! Refresh the cached file if the global cache state has changed.
+	shared_ptr<CachedFile> EnsureCachedFileCurrent();
 
 private:
 	QueryContext context;
--- a/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp
+++ b/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp
@@ -70,6 +70,8 @@ public:
 	bool IsEnabled() const;
 	void SetEnabled(bool enable);
 	vector<CachedFileInformation> GetCachedFileInformation() const;
+	//! Number of files tracked in the cache map, expose for testing purpose.
+	idx_t GetCachedFileCount() const;
 
 	//! Re-index to `current_block_size` if it differs from the cache block size.
 	//! Return the blocks cached for the given range.
@@ -77,8 +79,9 @@ public:
 	                                                       idx_t first_block, idx_t num_blocks);
 
 	BufferManager &GetBufferManager() const;
-	//! Gets the cached file, or creates it if is not yet present
-	CachedFile &GetOrCreateCachedFile(const string &path);
+	//! Gets the shared cached file for the given path, creating it if not yet present.
+	//! When caching is disabled, returns a transient CachedFile that is not tracked in the cached file map.
+	shared_ptr<CachedFile> GetOrCreateCachedFile(const string &path);
 
 	DUCKDB_API static bool IsValid(bool validate, const string &cached_version_tag, timestamp_t cached_last_modified,
 	                               const string &current_version_tag, timestamp_t current_last_modified);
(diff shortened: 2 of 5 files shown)


lnkuiper:
Thanks! Codex seems to think there is an issue from using the `shared_ptr` now:
```C++
#include "catch.hpp"
#include "caching_test_utils.hpp"
#include "duckdb/common/file_system.hpp"
#include "duckdb/common/types/value.hpp"
#include "duckdb/main/database.hpp"
#include "duckdb/storage/external_file_cache/caching_file_system.hpp"

namespace duckdb {
namespace {

OpenFileInfo MakeNoValidationOpenFileInfo(const string &path) {
	OpenFileInfo info(path);
	info.extended_info = make_shared_ptr<ExtendedOpenFileInfo>();
	info.extended_info->options["validate_external_file_cache"] = Value::BOOLEAN(false);
	return info;
}

string ReviewReadFull(CachingFileHandle &handle, idx_t size) {
	auto group = handle.Read(size, 0);
	string result(size, '\0');
	group.CopyTo(reinterpret_cast<data_ptr_t>(&result[0]), size);
	return result;
}

void ReviewWriteTestContent(const string &path, const string &content) {
	auto local_fs = FileSystem::CreateLocal();
	auto handle = local_fs->OpenFile(path, FileFlags::FILE_FLAGS_WRITE | FileFlags::FILE_FLAGS_FILE_CREATE);
	handle->Write(QueryContext(), const_cast<char *>(content.data()), content.size(), 0);
	handle->Sync();
}

} // namespace

TEST_CASE("TEMP PR 22961: re-enabled cache must not use stale live handle blocks", "[external_file_cache][review]") {
	DuckDB db(":memory:");
	auto &db_instance = *db.instance;
	auto &cache = db_instance.GetExternalFileCache();

	const idx_t file_size = 64;
	const string content_a(file_size, 'A');
	const string content_b(file_size, 'B');
	CachingTestFileGuard test_file("temp_pr_22961_live_handle.bin", content_a);

	auto tracking_fs = make_uniq<SimpleTrackingFileSystem>();
	CachingFileSystem cfs(*tracking_fs, db_instance);

	auto handle = cfs.OpenFile(MakeNoValidationOpenFileInfo(test_file.GetPath()), FileFlags::FILE_FLAGS_READ);
	REQUIRE(ReviewReadFull(*handle, file_size) == content_a);
	REQUIRE(cache.GetCachedFileCount() == 1);

	cache.SetEnabled(false);
	REQUIRE(cache.GetCachedFileCount() == 0);
	ReviewWriteTestContent(test_file.GetPath(), content_b);
	REQUIRE(ReviewReadFull(*handle, file_size) == content_b);

	cache.SetEnabled(true);
	REQUIRE(ReviewReadFull(*handle, file_size) == content_b);
}

TEST_CASE("TEMP PR 22961: handle opened while disabled should not cache into an untracked object",
          "[external_file_cache][review]") {
	DuckDB db(":memory:");
	auto &db_instance = *db.instance;
	auto &cache = db_instance.GetExternalFileCache();

	const idx_t file_size = 64;
	const string content(file_size, 'C');
	CachingTestFileGuard test_file("temp_pr_22961_disabled_open.bin", content);

	auto tracking_fs = make_uniq<SimpleTrackingFileSystem>();
	CachingFileSystem cfs(*tracking_fs, db_instance);

	cache.SetEnabled(false);
	auto handle = cfs.OpenFile(MakeNoValidationOpenFileInfo(test_file.GetPath()), FileFlags::FILE_FLAGS_READ);
	REQUIRE(ReviewReadFull(*handle, file_size) == content);
	REQUIRE(cache.GetCachedFileCount() == 0);

	cache.SetEnabled(true);
	REQUIRE(ReviewReadFull(*handle, file_size) == content);
	REQUIRE(cache.GetCachedFileCount() == 1);
}

} // namespace duckdb
```

Suggested fix:
```
• Yes. Suggested fix: make CachingFileHandle refresh its cached_file whenever the global cache state changes.

  A clean approach:

  - Add an atomic generation/epoch to ExternalFileCache.
  - Increment it in SetEnabled() whenever the enabled state changes, especially before/after clearing cached_files.
  - Return the current generation alongside GetOrCreateCachedFile, or expose GetGeneration().
  - Store cached_file_generation in CachingFileHandle.
  - Before any cached read path, call something like EnsureCachedFileCurrent():
      - If cache is disabled: do nothing, read uncached.
      - If generation differs: reacquire cached_file = external_file_cache.GetOrCreateCachedFile(path.path).
      - Reinitialize metadata for the new CachedFile from the already-open file_handle, or reopen/validate through GetFileHandle() in a way that updates the new cached_file.

  Important detail: simply assigning cached_file = GetOrCreateCachedFile() in Read() is not enough, because GetFileHandle() currently returns early when file_handle already exists and would skip populating metadata on the new CachedFile.

  I’d suggest splitting metadata refresh into a helper, e.g.:

  void CachingFileHandle::InitializeCachedFileMetadata();

  Then GetFileHandle() can call it after opening, and EnsureCachedFileCurrent() can call it after reacquiring a new cached file when file_handle is already open.

  This should fix both temp repros: stale orphaned cached blocks after re-enable, and handles opened while disabled caching into an untracked transient object after enable.
```
Take it with a grain of salt :) you probably know best!

dentiny:
Thanks for checking! The issue seems to be disable + re-enable leads to stale access, but current implementation doesn't handle it as well (which I want to address the followup PR).

Current implementation: when disabled, cache handles are [clearly directly](https://github.com/duckdb/duckdb/blob/e88f952d470a35ffd8264987e7ee470c02fea9c1/src/storage/external_file_cache/external_file_cache.cpp#L206), which leads to [dangling pointer](https://github.com/duckdb/duckdb/blob/e88f952d470a35ffd8264987e7ee470c02fea9c1/src/include/duckdb/storage/external_file_cache/caching_file_system.hpp#L82) in caching filesystem.

But let me fix it in the following commit.

dentiny:
@lnkuiper updated :)

lnkuiper:
This looks great! Thanks

PR #22962, opened 2026-05-30
Improve testing docs

### LINKED ISSUE #6787: Documentation about running unit tests unclear (or hard to find)
After submitting a PR, I wanted to run unittests locally, but could not find any clear information on how to do run tests, what tests I should be running (or skipping if they take too long?), how to run just single failing tests, etc.

In the documentation, I've found https://duckdb.org/docs/current/dev/sqllogictest/overview, which I think talks about how to *write* tests, but does not have any practical info on how to actually build the test runner and how to run it.

I've also been linked to https://github.com/duckdb/duckdb/blob/main/CONTRIBUTING.md#ci-for-pull-requests, which says you should run "local CI", without any reference as to how (I am not sure if "local" here means "github actions in your own fork", or "commands on your own PC", but some pointers about both would be helpful).

https://github.com/duckdb/duckdb/blob/main/CONTRIBUTING.md#testing and https://github.com/duckdb/duckdb/tree/main#development suggest there is a `make unit` and `make all-unit` commands for running tests. The former is actually good in terms of what to expect, but still does not tell how to run a single test or a group of tests. Also, I've run `make unit` locally and it has been burning CPU time for almost an hour now, instead of the 1 minute the documentation promises (throwing out "batch timed out after 600 seconds" errors every now and then, which is probably also not expected). It is also unclear if you need to build duckdb first or can just run these commands right away.

That page also talks about "sqllogictests" vs "C++ tests", but this distinction is not otherwise explained.

I've seen https://github.com/duckdb/duckdb/blob/main/test/README.md which is minimal and is only geared to writing tests.

Is there maybe some additional documentation somewhere that I missed and could be linked on some of these places?

matthijskooijman:
This clarifies some things that were not immediately obvious from reading the existing docs. This partially resolves https://github.com/duckdb/duckdb-web/issues/6787.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,8 @@ This project and everyone participating in it is governed by a [Code of Conduct]
 ## CI for pull requests
 
 * Pull requests will need to pass all continuous integration checks before merging.
-* When all checks look good on your local CI, you can open a PR to run the CI on the main DuckDB repo. Submitting changes to an open pull request will move it to 'Draft' state. In that case you can mark it as "Ready for Review" once you've applied all fixes and it passes in your fork to run the public CI again ('ready for review', via the Web UI button on the bottom right).
+* The CI worklow running inside the main repository must be approved on every push by new contributors (until a commit or PR by them is merged). It is recommended to run the same workflow in your own fork (without needing approval) to ensure everything works before opening your PR. See below under "Testing" how to enable this.
+* When all CI checks look good on your fork, you can open a PR to run the CI on the main DuckDB repo. Submitting changes to an open pull request will move it to 'Draft' state. In that case you can mark it as "Ready for Review" once you've applied all fixes and it passes in your fork to run the public CI again ('ready for review', via the Web UI button on the bottom right).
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -36,7 +36,8 @@ This project and everyone participating in it is governed by a [Code of Conduct]
 ## CI for pull requests
 
 * Pull requests will need to pass all continuous integration checks before merging.
-* When all checks look good on your local CI, you can open a PR to run the CI on the main DuckDB repo. Submitting changes to an open pull request will move it to 'Draft' state. In that case you can mark it as "Ready for Review" once you've applied all fixes and it passes in your fork to run the public CI again ('ready for review', via the Web UI button on the bottom right).
+* The CI worklow running inside the main repository must be approved on every push. It is recommended to run the same workflow in your own fork (without needing approval) to ensure everything works before opening your PR. See below under "Testing" how to enable this.
```
Tishj: This is only true for new contributors

matthijskooijman: Thanks for the feedback. Do you know when you stop being a "new" contributor? Is that once a PR is merged? Then I'll clarify.

matthijskooijman: I've found https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories that explains you are no longer new when a commit or PR by you is merged into the repo. I've updated this PR to clarify this.

I'll leave the comment open for you to resolve.

matthijskooijman:
Adressed one comment and rebased on top of main.

matthijskooijman:
I think this PR wis ready to merge from my side. It's a fairly trivial documentation update, anyone wants to review and approve it? :-)

PR #22975, opened 2026-05-31
Report <eof> in check_peg_parser parse errors

fallintoplace:
## Summary
- report `<eof>` when `check_peg_parser` fails after consuming every token
- route the error-token selection through a tiny helper used by the binder error path
- add regression coverage for the EOF token selection branch

## Why
`CheckPEGParserBind` formatted parse failures with `root_tokens[state.token_index].text`.
When the matcher had already advanced to the end of the token stream, that indexed one past the end instead of reporting EOF.

## Impact
Invalid PEG parser inputs that fail at end-of-input now report `<eof>` instead of reading past the token vector.

## Testing
- `python3 scripts/ci/run_tests.py build/debug/test/unittest "[autocomplete]"`
- `python3 scripts/ci/run_tests.py build/debug/test/unittest test/sql/peg_parser/parser/prepare_statement.test`

### DIFF
--- a/extension/autocomplete/autocomplete_extension.cpp
+++ b/extension/autocomplete/autocomplete_extension.cpp
@@ -558,6 +558,11 @@ static duckdb::unique_ptr<FunctionData> CheckPEGParserBind(ClientContext &contex
 	// `+ 1` accounts for the EOI sentinel — the autocomplete walk may report SUCCESS without
 	// consuming it.
 	if (match_result != MatchResultType::SUCCESS || state.token_index + 1 < root_tokens.size()) {
+		auto error_token = string("<eof>");
+		if (state.token_index < root_tokens.size() && root_tokens[state.token_index].type != TokenType::END_OF_INPUT &&
+		    root_tokens[state.token_index].type != TokenType::END_OF_INPUT_AUTOCOMPLETE) {
+			error_token = root_tokens[state.token_index].text;
+		}
 		string token_list;
 		for (idx_t i = 0; i < root_tokens.size(); i++) {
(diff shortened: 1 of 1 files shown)


Dtenwolde:
Hi there, thanks for the PR! I think this is a good addition. 

I wonder if we should make `check_peg_parser`'s error handling the same as the real parser, so you get better informed error messages rather than a list of the tokens. This can be tackled in a future PR of course :)

src/include/duckdb/parser/peg/autocomplete_error.hpp
```diff
@@ -0,0 +1,19 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/parser/peg/matcher.hpp"
+
+namespace duckdb {
+
+inline string GetCheckPEGParserErrorToken(const vector<MatcherToken> &tokens, idx_t token_index) {
```
Dtenwolde: IMO  this doesn't really need a separate function and new file

Dtenwolde (changes requested):

Dtenwolde (approved):

lnkuiper:
Looks like CI got stalled indefinitely, and now there's a merge conflict :(

fallintoplace:
@lnkuiper I can fix the merge conflict later the day.

Dtenwolde:
My bad, I was late on pinging to start the CI workflow. From my side the changes look good, though things might have changed here slightly after https://github.com/duckdb/duckdb/pull/23027

Dtenwolde (approved):

PR #22978, opened 2026-06-01
Add setting to allows to get compile time STANDARD_VECTOR_SIZE from SQL

carlopi:
Minor, but handy to double check what's the compiled in `STANDARD_VECTOR_SIZE`

```sql
SELECT current_setting('standard_vector_size');
--- 2048
```

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -1176,6 +1176,17 @@
         "custom_implementation": true
     },
     {
+        "name": "standard_vector_size",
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks - but let's not add any new `PRAGMA` functions - maybe we can just expose this in `duckdb_settings` as a setting that can be read but not set?

carlopi:
Sure, I was on the fence whether non-settable setting vs `PRAGMA`, I'll change direction.

PR #22981, opened 2026-06-01
Allow `OPERATOR` to be qualified

Dtenwolde:
Fix https://github.com/duckdblabs/duckdb-internal/issues/8702

The following query is now allowed: 
```sql
select 1 OPERATOR(main.+) 2;
```

Referencing a catalog as well as a schema does not work:
```sql
memory D select 1 operator(memory.main.*) 2;
Parser Error:
Too many identifiers found, expected schema.operator or operator
```


Mytherin:
Thanks!

PR #22989, opened 2026-06-01
change sqllogictest branch

ccfelius:
In V1.5 the WHERE clause evaluates a chain of INT multiplications that overflow, which are on their turn causing an Out of Range Error.

Currently the multiplication is never executed since all input columns are non-null. So the test should pass since the error is optimized away.

Fixes https://github.com/orgs/duckdblabs/projects/43/views/3?filterQuery=-status%3ANew%2CBacklog%2C%22Ready+to+Pick+Up%22%2CDone+assignee%3A%40me&pane=issue&itemId=182928659&issue=duckdblabs%7Cduckdb-internal%7C9115

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -792,7 +792,7 @@ format-configs:
 
 
 third_party/sqllogictest:
-	git clone --depth=1 --branch hawkfish-statistical-rounding https://github.com/duckdb/sqllogictest.git third_party/sqllogictest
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks - I think the test just has to be added to a test config ignore (since it relies on the optimizer to work)

PR #22994, opened 2026-06-01
Fix partial column metadata reuse bug

ywelsch:
Fixes an (unreleased) bug from https://github.com/duckdb/duckdb/pull/22333 where the column metadata would not be carried correctly forward between `RowGroup::WriteToDisk` and `RowGroup::Checkpoint` when forcing column metadata reuse was enabled (disabled by default in 1.5).

Found this while fuzzing the existing test suite with a combination of force_column_metadata_reuse / debug_verify_blocks / "force_restart": "true" and  "block_size": "16384" and added a test to specifically catch this.


Mytherin:
Thanks!

PR #22998, opened 2026-06-02
Support multiple triggers per table event

EtgarDev:
## Summary

Multiple triggers on the same table and event are now supported. When a DML statement fires, all matching triggers execute in alphabetical order by name.

**Note:** PostgreSQL uses byte (ASCII) ordering for trigger names, where uppercase sorts before lowercase.
We use the `CatalogEntryMap` which its comparator is not case-sensitive: https://github.com/duckdb/duckdb/blob/06efe8d484311db189cca15426445664b77d6201/src/catalog/catalog_set.cpp#L48-L50

### DIFF
--- a/src/catalog/catalog_entry/duck_table_entry.cpp
+++ b/src/catalog/catalog_entry/duck_table_entry.cpp
@@ -1385,6 +1385,11 @@ optional_ptr<CatalogEntry> DuckTableEntry::CreateTrigger(CatalogTransaction tran
 		if (old_entry) {
 			return nullptr;
 		}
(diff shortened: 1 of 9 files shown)


lnkuiper:
Thanks! Looks good from my side, the issues that I raised in the original PR have been fixed :)

EtgarDev:
> Thanks! Looks good from my side, the issues that I raised in the original PR have been fixed :)

@lnkuiper Thank you!
I've added now the multiple triggers test to the `encryption` configuration so the auto-merge was disabled

PR #23000, opened 2026-06-02
Peeling statements in the parser

carlopi:
Idea is moving away from whole `Program` parsing, and into peeling one `TopLevelStatement` at a time.

This is relevant since it allows to interleave parsing and execution, for example:
```sql
LOAD <my_extension>; quack quack quack;
```
would parse correctly IF `<my_extension>` where to provide a parser_function that recognize `quack quack quack` as a valid statement.

In particular, working on `CONNECT`-statement, I found that it was restrictive to force all parsing to happen ahead of time, and we might want to allow interleaving of parse and execution (with side effects on parser itself).

Currently as per this PR, only infrastructure is added to introduce an StatementIterator of sort, but all code still does first tokenization (in full), then parsing (one at a time, until completion), then execution.
A next PR will interleave parsing and execution in certain code-paths.

A nice consequence of this, I think, is allowing parser extensions to be more proper first class citizens, where the interface changes to allow for only partially recognizing a statement.

Say the `demo_loadable_extension`, that now recognize the most relevant string `quack quack quack`, can be interleaved like:
```sql
SELECT 42; quack quack quack; SELECT 42;
```
where before, this would fail, or this would force to duplicate the parser, now this is peeled once at a time like:
1. `SELECT 42;` is the first statement peeled, recognized by PEG parser, (rest is ` quack quack quack; SELECT 42;`)
2. `quack quack quack; SELECT 42;` is NOT recognized by PEG parser
3. are there parser extensions? Yes, let's iterate them
4. `demo_loadable_extension` recognize `quack quack quack; SELECT 42;` as valid, until 17th char. Remaining `; SELECT 42;`
5. PEG resume peeling, removes `;` (empty statement, not added)
6. PEG peels `SELECT 42;`

Result, there are 3 statements: `SELECT 42;` + `quack quack quack` + `SELECT 42;`.

The extension DOES rework the `parse_function`, so this will require extensions changes.

### DIFF
--- a/extension/autocomplete/autocomplete_extension.cpp
+++ b/extension/autocomplete/autocomplete_extension.cpp
@@ -528,7 +528,14 @@ static duckdb::unique_ptr<FunctionData> CheckPEGParserBind(ClientContext &contex
 	names.emplace_back("success");
 	return_types.emplace_back(LogicalType::BOOLEAN);
 
-	const auto sql = StringValue::Get(input.inputs[0]);
+	// CheckPEGParserBind feeds the input to `peg_matcher->Root().Match(state)`, which goes
+	// through `ListMatcher::Match`'s out-of-tokens autocomplete-suggestion path. That path
+	// returns FAIL whenever the next child is mandatory — and after the peeling refactor the
+	// last child is `(';'+ / EndOfInput)`, where `;'+` is mandatory. Without a `;` to consume,
+	// every well-formed query gets reported as a parse failure here. Append a `;` so the
+	// matcher reaches a clean state. The principled fix (two distinct sentinels for "real
+	// EOI" vs "autocomplete cursor") is tracked separately.
+	auto sql = StringValue::Get(input.inputs[0]) + ";";
 
 	vector<MatcherToken> root_tokens;
 	string clean_sql;
--- a/src/include/duckdb/common/string.hpp
+++ b/src/include/duckdb/common/string.hpp
@@ -10,10 +10,12 @@
 
 #include "duckdb/original/std/sstream.hpp"
 #include <string>
+#include <string_view>
 #include <locale>
 
 namespace duckdb {
 using std::string;
+using std::string_view;
 } // namespace duckdb
 
 namespace duckdb {
--- a/src/include/duckdb/main/statement_iterator.hpp
+++ b/src/include/duckdb/main/statement_iterator.hpp
@@ -1,0 +1,72 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// duckdb/main/statement_iterator.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/string.hpp"
+#include "duckdb/common/unique_ptr.hpp"
+#include "duckdb/common/vector.hpp"
+
+namespace duckdb {
+class ClientContext;
+class Parser;
+class SQLStatement;
+struct MatcherToken;
+
+//! Iterator over the SQL statements contained in a multi-statement query.
+//!
+//! Usage:
+//!   auto it = context.ExtractStatements(sql);
+//!   while (it.Peek(context)) {
+//!       auto stmt = it.GetStatement();
+//!       // dispatch stmt
+//!   }
+//!
+//! Peek does the work — it advances the byte cursor through `sql` and parses one
+//! TopLevelStatement per call via `Parser::ParseStatement`. The iterator buffers at most one
+//! statement at a time; GetStatement yields it and clears the buffer for the next Peek.
+//!
+//! Separator-only stretches (e.g. ";;;") are transparently skipped inside Peek — the caller
+//! always sees either a real statement or a clean exhaustion.
+class StatementIterator {
+public:
+	DUCKDB_API explicit StatementIterator(string sql);
+	DUCKDB_API ~StatementIterator();
+
+	StatementIterator(const StatementIterator &) = delete;
+	StatementIterator &operator=(const StatementIterator &) = delete;
+	DUCKDB_API StatementIterator(StatementIterator &&) noexcept;
+	DUCKDB_API StatementIterator &operator=(StatementIterator &&) noexcept;
+
+	//! Returns true if a statement is currently available (after parsing as needed). Returns
+	//! false when the iterator is exhausted. Non-const: parses on demand and buffers the result.
+	DUCKDB_API bool Peek(ClientContext &context);
+
+	//! Returns the next buffered statement and clears the buffer. Returns nullptr if no
+	//! statement is buffered — the caller should call Peek(context) first.
+	DUCKDB_API unique_ptr<SQLStatement> GetStatement();
+
+private:
+	string sql;
+	//! Parser instance kept alive across Peek calls so its PEG matcher / transformer caches
+	//! stay warm. Constructed lazily on the first Peek.
+	unique_ptr<Parser> parser;
+	//! Tokenized view of `sql`. Populated once on the first Peek and walked thereafter via
+	//! `token_cursor`, avoiding O(N²) re-tokenization across N statements.
+	unique_ptr<vector<MatcherToken>> tokens;
+	//! Index into `tokens` at which the next match starts.
+	idx_t token_cursor = 0;
+	//! Single-statement buffer holding the result of the most recent Peek. Cleared by
+	//! GetStatement.
+	unique_ptr<SQLStatement> current_statement;
+	//! Once Peek determines there are no more statements (cursor past end of tokens), we stay
+	//! exhausted; subsequent Peek calls return false without re-invoking the parser.
+	bool exhausted = false;
+};
+
+} // namespace duckdb
--- a/src/include/duckdb/parser/parser.hpp
+++ b/src/include/duckdb/parser/parser.hpp
@@ -20,6 +20,7 @@
 namespace duckdb {
 
 struct ParserCache;
+struct MatcherToken;
 class GroupByNode;
 struct UnicodeSpace {
 	UnicodeSpace(idx_t pos, idx_t bytes) : pos(pos), bytes(bytes) {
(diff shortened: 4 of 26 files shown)


extension/autocomplete/autocomplete_extension.cpp
```diff
@@ -528,7 +528,14 @@ static duckdb::unique_ptr<FunctionData> CheckPEGParserBind(ClientContext &contex
...
 
-	const auto sql = StringValue::Get(input.inputs[0]);
+	// CheckPEGParserBind feeds the input to `peg_matcher->Root().Match(state)`, which goes
+	// through `ListMatcher::Match`'s out-of-tokens autocomplete-suggestion path. That path
+	// returns FAIL whenever the next child is mandatory — and after the peeling refactor the
+	// last child is `(';'+ / EndOfInput)`, where `;'+` is mandatory. Without a `;` to consume,
+	// every well-formed query gets reported as a parse failure here. Append a `;` so the
+	// matcher reaches a clean state. The principled fix (two distinct sentinels for "real
+	// EOI" vs "autocomplete cursor") is tracked separately.
+	auto sql = StringValue::Get(input.inputs[0]) + ";";
```
Dtenwolde, line 538: I don't think we should have to append `;` to get valid SQL, it is only necessary for splitting statements. Seeing end of input should be enough for a single statement. 

So I guess the ';'+ should really be ';'*

src/include/duckdb/common/string.hpp
```diff
@@ -10,10 +10,12 @@
 
 #include "duckdb/original/std/sstream.hpp"
 #include <string>
+#include <string_view>
 #include <locale>
 
 namespace duckdb {
 using std::string;
+using std::string_view;
```
Dtenwolde, line 18: I don't know the difference precisely between `string` and `string_view`, but do we need this?

src/include/duckdb/parser/peg/transformer/parse_result.hpp
```diff
@@ -347,6 +362,10 @@ class ChoiceParseResult : public ParseResult {
 		return result;
 	}
 
+	idx_t GetSelectedIdx() const {
```
Dtenwolde, line 365: When do we use this? The index of the `ChoiceResult` shouldn't ever really matter I think

src/include/duckdb/parser/peg/transformer/parse_result.hpp
```diff
@@ -196,6 +199,18 @@ struct KeywordParseResult : ParseResult {
 	}
 };
 
+struct EndOfInputParseResult : ParseResult {
```
Dtenwolde, line 202: Why do we need an explicit `EndOfInput`? That is not entirely clear to me

src/include/duckdb/parser/peg/inlined_grammar.hpp
```diff
@@ -4,7 +4,8 @@
 namespace duckdb {
 
 const char INLINED_PEG_GRAMMAR[] = {
-	"Program <- Statement? (';'+ Statement)* ';'*\n"
+	"Program <- TopLevelStatement*\n"
+	"TopLevelStatement <- ';'+ / (Statement (';'+ / EndOfInput))\n"
```
Dtenwolde, line 8: If we don't have an explicit `EndofInput` I think this rule can stay largely the same as the old. 

Though I also now wonder if it is correct that we accept an empty query, since `TopLevelStatement*` means 0 or more... Perhaps only a technicality

src/include/duckdb/parser/parser.hpp
```diff
@@ -47,6 +48,15 @@ class Parser {
 	//! variable.
 	void ParseQuery(const string &query);
 
+	//! Parse a single TopLevelStatement from an already-tokenized stream starting at
+	//! `token_cursor`. On success advances `token_cursor` past the consumed tokens and returns
+	//! the SQLStatement. Returns nullptr at end-of-input or when the matched TLS was a
```
Dtenwolde, line 53: What is `TLS`?

src/include/duckdb/main/statement_iterator.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+	//! Returns the next buffered statement and clears the buffer. Returns nullptr if no
+	//! statement is buffered — the caller should call Peek(context) first.
+	DUCKDB_API unique_ptr<SQLStatement> GetStatement();
+
+private:
+	string sql;
+	//! Parser instance kept alive across Peek calls so its PEG matcher / transformer caches
+	//! stay warm. Constructed lazily on the first Peek.
+	unique_ptr<Parser> parser;
```
Dtenwolde, line 58: I think this should be `ParserCache` such that we can invalidate it/update it in case an extension loads new grammar/transformer rules

Dtenwolde: Oh nvm, `Parser` has the `ParserCache` as a private member

src/include/duckdb/parser/parser.hpp
Dtenwolde: `TopLevelStatement` I suppose

src/parser/peg/transformer/peg_transformer_factory.cpp
```diff
@@ -53,95 +53,64 @@ static unique_ptr<SQLStatement> ExtractAndTransformStatement(PEGTransformer &tra
...
-	// Program[2] = optional repeat of trailing ';'
-	auto &prog = match_result->Cast<ListParseResult>();
+	// Advance the caller's cursor past the consumed tokens.
+	token_cursor = state.token_index;
+
+	// TopLevelStatement <- ';'+ / (Statement (';'+ / EndOfInput))
+	// Same tree shape walked inside Transform(); see comments there.
+	auto &tls = match_result->Cast<ListParseResult>();
+	auto &outer_choice = tls.Child<ChoiceParseResult>(0);
+	if (outer_choice.GetSelectedIdx() == 0) {
```
Dtenwolde, line 90: I want to avoid having to rely on the order of the rules in the grammar, which is why I avoided exposing the `select_idx` for `ChoiceParseResult`.

Dtenwolde: Perhaps we can create just a "syntax-only" rule for `;` 

Then something along these lines: 
```
TopLevelStatement <- Semicolon+ / Statements
Statements <- Statement EndOfQuery
EndOfQuery <- Semicolon / EndofInput
```

src/parser/peg/transformer/peg_transformer_factory.cpp
```diff
@@ -53,95 +53,64 @@ static unique_ptr<SQLStatement> ExtractAndTransformStatement(PEGTransformer &tra
...
+		// separator-only TopLevelStatement — no statement to yield
+		return nullptr;
+	}
+	auto &inner_list = outer_choice.GetResult().Cast<ListParseResult>();
+	auto &stmt_pr = inner_list.GetChild(0);
+	auto &term_wrapper = inner_list.Child<ListParseResult>(1);
+	auto &term_choice = term_wrapper.Child<ChoiceParseResult>(0);
+
+	optional_idx terminator_offset;
+	if (term_choice.GetSelectedIdx() == 0) {
```
Dtenwolde, line 100: Again, relying on the order of the choices

src/parser/peg/matcher.cpp
```diff
@@ -1412,6 +1455,8 @@ shared_ptr<PEGMatcher> ParserCache::GetMatcher() {
 #else
 	new_matcher->root = factory.CreateMatcher(const_char_ptr_cast(INLINED_PEG_GRAMMAR), "Program");
 #endif
+	// TopLevelStatement is referenced by Program, so it has already been built and cached.
+	new_matcher->top_level_root = factory.GetMatcher("TopLevelStatement");
```
Dtenwolde, line 1459: What is different from the `root` and `top_level_root`?

src/parser/parser.cpp
```diff
@@ -258,34 +258,59 @@ void Parser::ParseQuery(const string &query) {
...
-			for (auto &stmt : peg_statements) {
+			auto stmt = ParseTopLevelStatement(tokens, token_cursor);
+			if (stmt) {
 				statements.push_back(std::move(stmt));
 			}
 		} catch (ParserException &e) {
-			// fall back to parse_function extensions for unknown statement types
 			bool parsed = false;
 			if (options.extensions && options.extensions->HasParserExtensions()) {
+				idx_t failure_byte = token_cursor < tokens.size() ? tokens[token_cursor].offset : query.size();
```
Dtenwolde, line 277: I don't understand why we are doing all of this. What is `allowed_boundaries` and why do we need to sort and deduplicate tokens?

test/api/test_statement_iterator.cpp
```diff
@@ -0,0 +1,147 @@
...
+	REQUIRE(s0);
+
+	// Move-construct a new iterator from `it`. The remaining statements should be reachable
+	// through the moved-to iterator.
+	StatementIterator moved(std::move(it));
+	REQUIRE(moved.Peek(ctx));
+	auto s1 = moved.GetStatement();
+	REQUIRE(s1);
+	REQUIRE_FALSE(moved.Peek(ctx));
+}
```
Dtenwolde, line 147: Would also like to see some tests with `;;;;;;;SELECT1;;;;; SELECT 2;;;;;`

Dtenwolde (commented):
Mostly questions :) I think the idea of the PR is nice. Just wondering about whether we can make this implementation a bit cleaner. Also around `root` and `top_level_root` wondering if we can improve on that.

src/include/duckdb/parser/peg/inlined_grammar.hpp
carlopi: I think empty queries where accepted also before, so they should likely stay.

src/parser/parser.cpp
carlopi: I'll need to expand, we need `allowed_boundaries` to be able to fail on non-aligned splits. Basically, we are asking the callback to handle some number of bytes, but informing that mid-token it's forbidden.

I need to think how to make this more clear.

carlopi:
Discussed with @Dtenwolde, let's first land `END_OF_INPUT`, that should clean up part of the complexity here.

carlopi:
Thanks @Dtenwolde for the feedback, incorporated as https://github.com/duckdb/duckdb/pull/23032

PR #23001, opened 2026-06-02
Let AsyncExecutionTask always notify blocked pipeline when a task throws

smvv:
In the NightlyTests workflow, the job `Forcing async Sinks/Sources` [fails](https://github.com/duckdb/duckdb/actions/runs/26699912877/job/78690840068#logs) with two test timeouts:

```
batch timed out after 600.0 seconds

=== stdout ===
[0/10] (0%): test/sql/copy/parquet/test_parquet_scan.test
[1/10] (10%): test/sql/copy/parquet/parquet_prefetch_projection.test
[2/10] (20%): test/sql/copy/parquet/parquet2.test
[3/10] (30%): test/sql/copy/parquet/parquet_6044.test
[4/10] (40%): test/sql/copy/parquet/recursive_parquet_union_by_name.test

...

batch timed out after 600.0 seconds

=== stdout ===
[0/10] (0%): test/sql/copy/partitioned/hive_partitioned_write.test
[1/10] (10%): test/sql/copy/partitioned/hive_filter_pushdown.test
[2/10] (20%): test/sql/copy/partitioned/hive_partition_case_insensitive_column.test
[3/10] (30%): test/sql/copy/partitioned/hive_partitioning_overwrite.test
[4/10] (40%): test/sql/copy/partitioned/partition_issue_6304.test
[5/10] (50%): test/sql/copy/partitioned/partitioned_order_by.test
[6/10] (60%): test/sql/copy/partitioned/hive_partition_recursive_cte.test
```

The hangs happen because `FORCE_ASYNC_SINK_SOURCE=1` makes parquet scans create fake async tasks for testing.

Some of those fake async tasks throw `ThrowAsyncTask`. Before this change, `AsyncExecutionTask` only called the interrupt callback after the async task finished successfully. If the task threw, the callback was skipped.

That left recursive CTE execution waiting forever here:

`ExecuteRecursivePipelineInline -> InterruptDoneSignalState::Await`

The fix makes `AsyncExecutionTask` always wake the waiting pipeline when an async task fails. On normal success, behavior stays the same: only the last async task in the batch wakes the waiter. On failure, we wake immediately so the query can see the executor error and stop instead of hanging.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/9550.
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9527.
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9524.
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9496.
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9474.
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9450.
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9442.
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9380.
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9314.

### DIFF
--- a/src/parallel/async_result.cpp
+++ b/src/parallel/async_result.cpp
@@ -25,6 +25,8 @@ private:
 };
 
 class AsyncExecutionTask : public ExecutorTask {
+	enum class CompletionSignal { BATCH_FINISHED, BATCH_ERRORED };
+
 public:
 	AsyncExecutionTask(Executor &executor, unique_ptr<AsyncTask> &&async_task, InterruptState &interrupt_state,
 	                   shared_ptr<Counter> counter)
@@ -35,10 +37,10 @@ public:
 		try {
 			async_task->Execute();
 		} catch (...) {
-			SignalCompletion(true);
+			SignalCompletion(CompletionSignal::BATCH_ERRORED);
 			throw;
 		}
-		SignalCompletion(false);
+		SignalCompletion(CompletionSignal::BATCH_FINISHED);
 		return TaskExecutionResult::TASK_FINISHED;
 	}
 
@@ -47,9 +49,9 @@ public:
 	}
 
 private:
-	void SignalCompletion(bool force_callback) {
+	void SignalCompletion(CompletionSignal signal) {
 		auto finished = counter->IterateAndCheckCounter();
-		if (force_callback || finished) {
+		if (signal == CompletionSignal::BATCH_ERRORED || finished) {
 			interrupt_state.Callback();
 		}
 	}
--- a/src/parallel/async_result.cpp
+++ b/src/parallel/async_result.cpp
@@ -32,10 +32,13 @@ public:
 	      counter(std::move(counter)) {
 	}
 	TaskExecutionResult ExecuteTask(TaskExecutionMode mode) override {
-		async_task->Execute();
-		if (counter->IterateAndCheckCounter()) {
-			interrupt_state.Callback();
+		try {
+			async_task->Execute();
+		} catch (...) {
+			SignalCompletion(true);
+			throw;
 		}
+		SignalCompletion(false);
 		return TaskExecutionResult::TASK_FINISHED;
 	}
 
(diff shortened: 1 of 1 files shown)


kryonix (approved):
This looks sane to me. I was considering if we would want `SignalCompletion` to be a `nothrow`, but `Callback` would throw an `InternalException`. Therefore, I think it's fine as it is.

carlopi:
Would this means that the same Callback can be called more than once on the same `interrupt_state`? Should that not be guarded? Like putting `counter` in an invalid state?

carlopi:
I am not super sure if the invariant here is OK, or if it's the consumer that needs to adapt, or if expanding the API is needed.

smvv:
> Would this means that the same Callback can be called more than once on the same `interrupt_state`? Should that not be guarded? Like putting `counter` in an invalid state?

I think that's what the method checks:
```c++
	bool IterateAndCheckCounter() {
		D_ASSERT(counter.load() > 0);
		idx_t post_decreast = --counter;
		return (post_decreast == 0);
	}
```
because it only calls the callback if true is returned, right?

carlopi:
No, the callback will now be run (as per the PR) in 2 cases: `signal == CompletionSignal::BATCH_ERRORED || finished`

`finished` is, by design, happening at most once.
But there might be 0+ BATCH_ERRORED, possibly combined with `finished`.
That I think might lead to `Callback` being called more than once.

smvv:
> No, the callback will now be run (as per the PR) in 2 cases: `signal == CompletionSignal::BATCH_ERRORED || finished`
> 
> `finished` is, by design, happening at most once. But there might be 0+ BATCH_ERRORED, possibly combined with `finished`. That I think might lead to `Callback` being called more than once.

i see what you mean! i've changed the code to send the callback once.

carlopi:
I am not sure even with the latest change, it should be the Counter to own whether it has been sent or not. The invariant is that the `Counter` should fire at most once, and given multiple different `AsyncTasks` might use the same `Counter`, I think changing API there is likely the fix, not caller side.

carlopi:
My last comment is outdated: looks good on `45ab61cb0f7802ffa3a0a722e7a97596b5527824`

Mytherin:
Thanks!

PR #23002, opened 2026-06-02
Async Thread Pool / Task Queue

lnkuiper:
This PR separates the `TaskScheduler`'s thread pool and task queue into separate classes, `TaskSchedulerPool` and `TaskSchedulerQueue`. This is done to create two pools and queues, one for `TaskSchedulerType::REGULAR` and one for `TaskSchedulerType::ASYNC`. The `async_threads` setting is added to tune the number of threads in the async pool, which defaults to the hardware concurrency just like the `threads` setting. The new pool and queue are mostly unused except for the tasks that are scheduled in the `CachingFileSystem`, but this will enable async IO development in future PRs.


src/include/duckdb/parallel/task_scheduler.hpp
```diff
@@ -13,103 +13,110 @@
...
-	void RelaunchThreadsInternal(int32_t n, bool destroy);
+	TaskSchedulerPool &GetPool(TaskSchedulerType pool_type);
+	TaskSchedulerQueue &GetQueue(TaskSchedulerType pool_type) const;
+
+	//! Fetches a task, returns true if successful or false if no tasks were available
+	bool GetTaskInternal(shared_ptr<Task> &task);
+	bool GetTaskInternal(shared_ptr<Task> &task, TaskSchedulerType pool_type);
+	bool TryDequeueAndProcessTask(const DBConfig &config, TaskSchedulerQueue &queue, shared_ptr<Task> &task);
+
+	void SetThreadsInternal(TaskSchedulerType pool_type, idx_t n);
```
dentiny, line 107: Hmm is the function unused?

src/parallel/task_scheduler_pool.cpp
```diff
@@ -0,0 +1,199 @@
...
+	if (threads.size() < new_thread_count) {
+		// we are increasing the number of threads: launch them and run tasks on them
+		idx_t create_new_threads = new_thread_count - threads.size();
+
+		// Whether to pin threads to cores
+		static constexpr idx_t THREAD_PIN_THRESHOLD = 64;
+		const auto pin_threads =
+		    pool_type == TaskSchedulerType::REGULAR && // Only pin regular threads!
+		    (pin_thread_mode == ThreadPinMode::ON ||
+		     (pin_thread_mode == ThreadPinMode::AUTO && std::thread::hardware_concurrency() > THREAD_PIN_THRESHOLD));
```
dentiny, line 169: I bad some memory on `hardware_concurrency` not respecting cgroup, wondering if we should use cgroup-aware core number everywhere in the codebase?

lnkuiper: This uses cgroup-aware core number through the function implemented in `DBConfig`.

src/include/duckdb/parallel/task_scheduler.hpp
lnkuiper: You are right, the other functions call the function on the pool directly, but they should call this function instead

pdet:
LGTM, just one question, should the Async Result hold the pool type, so ScheduleTasks can route it?

lnkuiper:
> LGTM, just one question, should the Async Result hold the pool type, so ScheduleTasks can route it?

@pdet I would assume that Async Result just implicitly goes to the async queue, but if we need that flexibility we can add it. I haven't really thought about how async reads should use this, just separated out the pool/queue for now.

PR #23003, opened 2026-06-02
Parquet Prefetch - CostModel for defining column gaps

pdet:
This PR introduces a cost model to adaptively calculate the accepted column gap size for requests.

Prior to this PR, the gap was hardcoded to be 16Kib. 

For example, say we have a table in a remote parquet file with columns `C1|C2|C3`, but only `C1` and `C3` where being queried, we could have either a request with all columns, or two requests, one for `C1` and another for `C3`, this decision was made by checking if `C2` was `<= 16kib`.

The hardcoded value was problematic for remote files, as the decision of the gap size is highly dependent on the latency (i.e., costs of doing a request) and bandwidth (speed at which you can transfer bytes), hence we now calculate the gap by multiplying latency and bandwidth.

The latency and bandwidth are captured directly in the httpfs extension, with time to first byte representing latency and transfer speed representing bandwidth. The on-disk behavior is unchanged, and the first network read still falls into the `16kb` default.

The network capture system is lock-free, as each thread keeps its own estimates, and it also passes to the next files, avoiding file warm-up.

## Benchmark
To benchmark this, ive created several parquet files with many row groups where we have needed columns (k) and gap columns (f),
with the schema looking like: `k0|f0|k1|f1|k2|f2|k3|f3|k4|f4|k5|f5|k6|f6|k7`, the gap columns are ~ 64Kb. The files sit in S3, and I'm querying it from my laptop.

The query looks like: 
```sql
SELECT sum(k0)+sum(k1)+sum(k2)+sum(k3)+sum(k4)+sum(k5)+sum(k6)+sum(k7)
  FROM read_parquet('s3://.../prefetch_bench_wide/part_*.parquet');
```

In this benchmark, the baseline will always do 8 requests, while our algorithm will quickly adapt to do one request by also bringing the filler columns, as the bandwidth is fast and the latency is high. With 8 requests, this query runs in 28 seconds, while with our new algorithm, it runs in 5 seconds, hence, it brings a 5-6x better performance in this scenario.

### DIFF
--- a/src/storage/external_file_cache/caching_file_system.cpp
+++ b/src/storage/external_file_cache/caching_file_system.cpp
@@ -376,6 +376,10 @@ bool CachingFileHandle::OnDiskFile() {
 	return GetFileHandle().OnDiskFile();
 }
 
+bool CachingFileHandle::TryGetNetworkThroughput(NetworkThroughputEstimate &result) {
+	return GetFileHandle().TryGetNetworkThroughput(result);
+}
+
 idx_t CachingFileHandle::SeekPosition() {
 	return position;
 }
--- a/.github/patches/extensions/httpfs/combined.patch
+++ b/.github/patches/extensions/httpfs/combined.patch
@@ -1,8 +1,92 @@
+diff --git a/src/httpfs.cpp b/src/httpfs.cpp
+index 1a9c466..348feb0 100644
+--- a/src/httpfs.cpp
++++ b/src/httpfs.cpp
+@@ -365,6 +365,18 @@ unique_ptr<HTTPResponse> HTTPFileSystem::GetRangeRequest(FileHandle &handle, str
+ 	auto response = http_util.Request(get_request, http_client);
+ 
+ 	hfh.StoreClient(std::move(http_client));
++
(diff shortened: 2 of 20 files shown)


Mytherin:
Thanks!

src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -386,26 +358,28 @@ bool CachingFileHandle::Validate() const {
...
-		annotated_lock_guard<annotated_mutex> guard(current_cached_file->meta_lock);
-		return current_cached_file->on_disk_file;
+		annotated_lock_guard<annotated_mutex> guard(cached_file.meta_lock);
+		return cached_file.on_disk_file;
 	}
 	return GetFileHandle().OnDiskFile();
 }
 
+bool CachingFileHandle::TryGetNetworkThroughput(NetworkThroughputEstimate &result) {
+	return GetFileHandle().TryGetNetworkThroughput(result);
```
dentiny: I think if the file handle has been cached, we should directly use it; we tried hard to avoid creating new handles which involves multiple IO requests. https://github.com/duckdb/duckdb/pull/25855

PR #23004, opened 2026-06-02
Improve test runner

smvv:
- Use test timeout on slow runners (having a few test workers)
  - use `300s` for fast runners (10+ workers).
  - otherwise, use `600s`.
- Make test reproduce command easier to read/paste.
- Parse `All tests were skipped` message correctly (was ignored previously and counted as - Extract test time from a test batch.
  - Removes the implicitly forced `--batch-size=1` when `--track-runtime` is used.


Mytherin:
Thanks!

PR #23007, opened 2026-06-02
Make casts extensible for CASE expression

yan-alex:
This is an attempt to improve extensibility of casts for the CASE expression.
# The problem
When trying to implement the spark compatibility extension, I ran into the following issue:

This works in spark
```sql
CREATE TEMPORARY VIEW t AS SELECT 1;
SELECT IF(true, cast(1 as tinyint), cast(2 as string)) FROM t;
```
But in duckdb:
```sql
memory D CREATE TEMPORARY VIEW t AS SELECT 1
memory D SELECT IF(true, cast(1 as tinyint), cast(2 as string)) FROM t;
```
```
Binder Error:
Cannot mix values of type VARCHAR and TINYINT in CASE expression - an explicit cast is required    
```

Inside extensions we can register new implicit cast rules just fine. Example:

```c++
void myfunction(ExtensionLoader &loader) {

loader.RegisterCastFunction(LogicalType::VARCHAR, LogicalType::DOUBLE, BindSparkCastWrapper, 150)
```

But that doesn't get picked up in our query, because `CombineUnequalTypes` calls `CastRules::ImplicitCast()`, which is just a hardcoded set of rules, not pluggable by extensions.

# The fix
This PR lets `CombineUnequalTypes` call `CastFunctionSet::ImplicitCastCost(*context, right, left)` instead, which does look at the casts registered by extensions.

### DIFF
--- a/src/common/types.cpp
+++ b/src/common/types.cpp
@@ -1183,8 +1183,8 @@ struct TryGetTypeOperation {
 struct ForceGetTypeOperation {
 	static bool Operation(optional_ptr<ClientContext> context, const LogicalType &left, const LogicalType &right,
 	                      LogicalType &result) {
-		(void)context;
-		result = LogicalType::ForceMaxLogicalType(left, right);
+		result = context ? LogicalType::ForceMaxLogicalType(*context, left, right)
+		                 : LogicalType::DefaultForceMaxLogicalType(left, right);
 		return true;
 	}
 };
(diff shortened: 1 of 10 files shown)


src/common/types.cpp
```diff
@@ -1294,7 +1307,7 @@ static idx_t GetLogicalTypeScore(const LogicalType &type) {
 
 LogicalType LogicalType::ForceMaxLogicalType(const LogicalType &left, const LogicalType &right) {
```
Mytherin: Instead of this by default not having the context - let's maybe flip it around and have this method propagate the context, and add a new method (`DefaultForceMaxLogicalType`) that doesn't take a client context.

src/common/types.cpp
```diff
@@ -1182,12 +1195,12 @@ bool LogicalType::TryGetMaxLogicalType(ClientContext &context, const LogicalType
...
 		return true;
 	}
-	return TryGetMaxLogicalTypeUnchecked(left, right, result);
+	return TryGetMaxLogicalTypeInternal<TryGetTypeOperation>(&context, left, right, result);
 }
 
 bool LogicalType::TryGetMaxLogicalTypeUnchecked(const LogicalType &left, const LogicalType &right,
                                                 LogicalType &result) {
-	return TryGetMaxLogicalTypeInternal<TryGetTypeOperation>(left, right, result);
+	return TryGetMaxLogicalTypeInternal<TryGetTypeOperation>(nullptr, left, right, result);
```
Mytherin, line 1207: Same here

Mytherin (commented):
Thanks - looks good, but let's try to "by default" take the `ClientContext` so that we don't accidentally forget it, and have a separate method that explicitly doesn't take the client context

src/common/types.cpp
yan-alex: Ill add one more commit to make a `DefaultTryGetMaxLogicalTypeInternal()` without a `context` parameter.
Even though this function is not used outside this file, we rather not have this nullptr passing at the callsite, right?

yan-alex: Hmm actually nvm, I think it's fine like this.

PR #23008, opened 2026-06-02
Tokenizer add 2 EOF tokens: `END_OF_INPUT` and `END_NOW_AUTOCOMPLETE`

carlopi:
This is https://github.com/duckdb/duckdb/pull/23000, with a commit on top where two explicit EOF tokens are added, one to mark actual `END_OF_INPUT` (end of actual input/program) and `END_NOW_AUTOCOMPLETE` to mark where to start autocomplete (so, end of current input, to be extended).

Currently those are conflated by checking if we got to `token.size()`, but are two different and slightly different items. This allows to simplify and hopefully future-proof the autocomplete logic.

https://github.com/duckdb/duckdb/pull/23000 makes first to be reviewed on its own (de)merits, then this on top.


carlopi:
Outdated, split as https://github.com/duckdb/duckdb/pull/23027

PR #23009, opened 2026-06-02
Join filter mirroring

### LINKED ISSUE #9165: Missed scan pushdown in self-join
### What happens?

When doing a self join like

```sql
select *
from t as t1
left join t as t2  -- or inner join
  on t1.f = t2.f
where right(t1.f, 1) = '2'
```

DuckDB pushes down the `right(...)` filter to one of the `t`s but not both, missing the opportunity to multiple order of magnitude speedup. In my real world example, adding an extra

```sql
  and right(t2.f, 1) = '2'
```

speeds up the query ~ 100x.

### To Reproduce

```sql
create table t (f string);

insert into t (f) values ('42'), ('43');

explain
select *
from t as t1
left join t as t2
  on t1.f = t2.f
where right(t1.f, 1) = '2';
```

```
┌───────────────────────────┐
│         HASH_JOIN         │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│            LEFT           │
│           f = f           ├──────────────┐
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │              │
│           EC: 0           │              │
│          Cost: 0          │              │
│           FILTER          ││         SEQ_SCAN          │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│    (right(f, 1) = '2')    ││             t             │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│           EC: 0           ││             f             │
│                           ││   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│                           ││           EC: 0           │
└─────────────┬─────────────┘└───────────────────────────┘
┌─────────────┴─────────────┐
│         SEQ_SCAN          │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             t             │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│             f             │
│   ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─   │
│           EC: 0           │
└───────────────────────────┘
```

### OS:

macOS

### DuckDB Version:

0.9

### DuckDB Client:

cli

### Full Name:

Jonas Haag

### Affiliation:

QuantCo

### Have you tried this on the latest `main` branch?

I have tested with a release build (and could not test with a main build)

- [X] Yes, I have

DinosL:
`FilterCombiner` already propagates constant filters across an equi-join when the filter is on a plain column. This PR extends that to (non volatile) expressions on the joined column.

fixes: https://github.com/duckdb/duckdb/issues/9165
fixes: https://github.com/duckdblabs/duckdb-internal/issues/436

### DIFF
--- a/src/include/duckdb/optimizer/filter_combiner.hpp
+++ b/src/include/duckdb/optimizer/filter_combiner.hpp
@@ -51,6 +51,8 @@ public:
 
 	void GenerateFilters(const std::function<void(unique_ptr<Expression> filter)> &callback);
 	bool HasFilters();
+	void GenerateEquivalentFilters(const Expression &filter,
+	                               const std::function<void(unique_ptr<Expression> filter)> &callback);
 	TableFilterSet GenerateTableScanFilters(const vector<ColumnIndex> &column_ids,
 	                                        vector<FilterPushdownResult> &pushdown_results);
 
--- a/src/optimizer/filter_combiner.cpp
+++ b/src/optimizer/filter_combiner.cpp
@@ -111,6 +111,42 @@ FilterResult FilterCombiner::AddFilter(unique_ptr<Expression> expr) {
 	return result;
 }
 
+void FilterCombiner::GenerateEquivalentFilters(const Expression &filter,
+                                               const std::function<void(unique_ptr<Expression> filter)> &callback) {
+	if (filter.IsVolatile()) {
+		return;
+	}
+	// collect every column reference that belongs to an equivalence set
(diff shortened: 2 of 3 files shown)


Mytherin:
Thanks!

PR #23010, opened 2026-06-02
Adding global timeout setting

ccfelius:
This PR adds a `global_timeout` setting. Unlike `max_execution_time`, `global_timeout` applies to the entire process, causing any query to be cancelled once the timeout is reached. When a query is cancelled due to this timeout, a timeout exception is raised.

This is useful e.g. if you want to limit the execution time of all queries on a quack server.

If both `global_timeout` and `max_execution_time` are set, the lower value wins.

Intended for https://github.com/duckdblabs/duckdb-internal/issues/9424


Mytherin:
Thanks - but maybe we can instead make `max_execution_time` [a generic setting](https://github.com/duckdb/duckdb/pull/18447)? Generic settings can be set both locally and globally.

src/main/client_context.cpp
```diff
@@ -1276,7 +1292,7 @@ void ClientContext::InterruptCheck() const {
 	if (query_deadline.IsValid() && ++timeout_check_counter % TIMEOUT_CHECK_INTERVAL == 0) {
 		auto now = NumericCast<idx_t>(duration_cast<milliseconds>(steady_clock::now().time_since_epoch()).count());
 		if (now >= query_deadline.GetIndex()) {
-			throw InterruptException();
+			throw QueryTimeoutException("Query exceeded maximum execution time");
```
Mytherin, line 1295: Might be easier to add support for attaching a message to the `InterruptException` - then we don't need to add a new exception type

ccfelius:
> Thanks - but maybe we can instead make max_execution_time https://github.com/duckdb/duckdb/pull/18447? Generic settings can be set both locally and globally. 

ah, sounds better, will open another pr to do that

PR #23018, opened 2026-06-02
Change scope of max execution time

ccfelius:
This PR changes the scope of `max_execution_time` and throws an appropriate exception when this set time is exceeded. This is useful e.g. if you want to limit the execution time of _all_ queries on a quack server.

Intended for https://github.com/duckdblabs/duckdb-internal/issues/9424


rustyconover:
If we're revisiting max_execution_time could we make it take an interval rather than knowing "time" equals seconds.

Mytherin:
Thanks! Looks good - CI failure is unrelated

Mytherin:
Agreed making this an interval would be nicer, that could perhaps be done in a separate PR.

PR #23021, opened 2026-06-03
loop in sleep_ms

ccfelius:
Instead of performing a single long sleep, we now sleep in a loop so that `sleep_ms()` can be e.g. cancelled while it's waiting


src/common/thread_util.cpp
```diff
@@ -1,16 +1,20 @@
...
+#include "duckdb/common/helper.hpp"
 
 namespace duckdb {
 
 #ifndef DUCKDB_NO_THREADS
 void ThreadUtil::SleepMs(idx_t sleep_ms) {
-	std::this_thread::sleep_for(std::chrono::milliseconds(sleep_ms));
+	static constexpr idx_t SLEEP_INTERVAL_MS = 100;
+	for (idx_t remaining = sleep_ms; remaining > 0; remaining -= MinValue(remaining, SLEEP_INTERVAL_MS)) {
+		std::this_thread::sleep_for(milliseconds(MinValue(remaining, SLEEP_INTERVAL_MS)));
```
Mytherin: We should also check the interrupted flag in between calls if we want this to be cancellable

Mytherin: We might also want to sleep with a "target time" - i.e. use `Timestamp::GetCurrentTimestamp() ` to get the current time, then add `sleep_ms` milliseconds to get the "target time", and re-check the current time in-between sleeps. That's because sleep is not exactly accurate but depends on CPU cycles, i.e. you might sleep for 6ms instead of 5ms which can add up in inaccuracy over a bunch of different sleeps.

Mytherin (commented):

src/common/thread_util.cpp
```diff
@@ -1,16 +1,31 @@
...
+	target_time.value += static_cast<int64_t>(sleep_ms) * Interval::MICROS_PER_MSEC;
+	static constexpr idx_t DEFAULT_SLEEP_INTERVAL_MS = 100;
+
+	auto sleep_interval = MinValue(DEFAULT_SLEEP_INTERVAL_MS, sleep_ms);
+	while (Timestamp::GetCurrentTimestamp() < target_time) {
+		// check interrupt flag
+		if (context && context->IsInterrupted()) {
+			throw InterruptException();
+		}
+		std::this_thread::sleep_for(milliseconds(sleep_interval));
```
Mytherin: Maybe `MinValue(target_time - <current_time>, sleep_interval)` to prevent us from entering a `100ms` sleep when there's only 10ms remaining?

Mytherin:
Thanks!

PR #23023, opened 2026-06-03
Bump MySQL

staticlibs:
This PR updates MySQL extension:

 - all patches are applied
 - a few bits of logic around optimizer/filter pushdows were moved to the common [database-connector](https://github.com/duckdb/database-connector) repo; this should not create problems for new patches, as the `database-connector` is used as a plain submodule (without `vcpkg` or `CMake` packaging involved), so new patches should be applicable to it after the checkout in MySQL repo
 - it seems that the extension config was overwritten recently by some merge (perhaps 1.4 branch?), the majority of numerous additions in `extension_entries.hpp` were already present in `1.5.2` and `1.5.3`, actual change there is minor


Mytherin:
Thanks!

PR #23024, opened 2026-06-03
Preserve simple CASE subject during execution

### LINKED ISSUE #22576: Simple CASE rewrites subject into searched form, losing the distinction and causing wrong results with non-deterministic subjects
### What happens?

`TransformCase` in `transform_case.cpp` rewrites simple CASE (`CASE x WHEN v1 THEN r1 ...`) into searched form by wrapping each WHEN value in `ComparisonExpression(COMPARE_EQUAL, root_arg->Copy(), val)` and discarding the original subject. The resulting `CaseExpression` AST has no way to distinguish simple vs searched CASE — there is no subject field, only `case_checks` and `else_expr`.

This causes incorrect results when the subject is non-deterministic, because each WHEN branch independently re-evaluates its own copy of the subject:

```sql
SELECT CASE cast(floor(random() * 3) AS integer)
  WHEN 0 THEN 'zero'
  WHEN 1 THEN 'one'
  WHEN 2 THEN 'two'
  ELSE 'other'
END
FROM generate_series(1, 10000);
```

`cast(floor(random() * 3) AS integer)` only produces 0, 1, or 2 — so `'other'` should never appear. In practice it appears ~30% of the time because each `WHEN` branch rolls `random()` independently.

**Expected behavior**

The subject expression should be evaluated exactly once per row. The ELSE branch should never be reached.

**Root cause**

In `src/parser/transform/expression/transform_case.cpp`, when `root_arg` is non-null (simple CASE), each WHEN value is wrapped as:
```cpp
comp_expr = make_uniq<ComparisonExpression>(
    ExpressionType::COMPARE_EQUAL, root_arg->Copy(), std::move(test_raw));
```

The subject is `.Copy()`'d per branch, and `root_arg` itself is not stored anywhere on the resulting `CaseExpression`. At execution time each branch evaluates its copy independently.

**Suggested fix**

Preserve the simple CASE subject on `CaseExpression` — e.g. add an optional `unique_ptr<ParsedExpression> subject` field. When `subject` is non-null, the executor evaluates it once and compares the cached result against each WHEN value.

### To Reproduce

```
pip3 install duckdb
python3 -c "import duckdb; duckdb.sql(\"<SQL>\").show()"
```
SQL used:
```sql
-- 'other' should never appear but it does
SELECT result, count(*) FROM (
  SELECT CASE cast(floor(random() * 3) AS integer)
    WHEN 0 THEN 'zero'
    WHEN 1 THEN 'one'
    WHEN 2 THEN 'two'
    ELSE 'other'
  END AS result
  FROM generate_series(1, 100000)
) GROUP BY result ORDER BY result;
```

### OS:

Redhat

### DuckDB Version:

v1.2.2

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Bikramjeet Vig

### Affiliation:

Meta Platforms Inc.

- [x] Yes, I have

- [x] Yes, I have

feichai0017:
Fixes #22576.

This restores the generic `invoke(lambda, args...)` scalar function and rewrites simple CASE expressions into `invoke(lambda subject: CASE WHEN subject = ... THEN ... END, original_subject)` during parsing. This makes the simple CASE subject a regular invoke argument, so volatile subjects are evaluated once per row while the bound/executed CASE expression remains the existing searched CASE form.

If a simple CASE branch contains a subquery, the parser keeps DuckDB's existing searched CASE rewrite because lambda expressions currently reject subqueries. Subject subqueries still go through `invoke` because they are outside the lambda body.

The PR also keeps lambda argument slots in natural parameter order and updates the list-comprehension rewrite to match that ordering.


Mytherin:
Thanks but this is a complex rewrite that is introducing a lot of complexity specifically for this one scenario. A better solution is to pick up adding `invoke` again - https://github.com/duckdb/duckdb/pull/20477 - which then allows this to just be rewritten to `invoke` which prevents having to re-execute the same parameter in each branch, e.g:

```sql
select invoke(lambda x: CASE 
  WHEN x=0 THEN 'zero'
  WHEN x=1 THEN 'one'
  WHEN x=2 THEN 'two'
  ELSE error('should never happen', cast(floor(random() * 3) AS integer))
FROM generate_series(1, 10000);
```

feichai0017:
Thanks for the guidance. I agree this PR is adding too much CASE-specific machinery. I’ll stop pursuing this approach and look into reviving `invoke` from #20477, then rewrite simple CASE through `invoke` so the subject is evaluated once without adding a separate CASE-specific execution/serialization path.

Maxxen:
I was actually looking into adding invoke back just now, you can pick it up if you want but just note that it was removed due to breaking backwards compatability with serialized plans, so we would have to add a test for that and verify that it can be executed correctly when serialized across versions. Most likely we will need to flip the lambda binding order when serializing to older format as the initial invoke PR attempted to "fix" it - but I don't remember all the details since it was a while ago.

Maxxen:
Some context: https://github.com/duckdb/duckdb/pull/20995

feichai0017:
@Mytherin  @Maxxen 
Thanks both. I have reworked the PR in the `invoke` direction now: `invoke` is restored, and simple CASE is rewritten to `invoke(lambda __duckdb_simple_case_subject: CASE WHEN ... END, subject)`. This lets us evaluate the simple CASE subject once without adding CASE-specific execution or serialization state.

I also looked at #20995. I agree the serialized-plan/backwards-compatibility risk needs explicit coverage, especially around lambda reference ordering. I will add targeted serialization/BWC tests before requesting another review.

One caveat I found is that DuckDB currently does not support subqueries inside lambda expressions. This is not needed for the issue reported here, but it means a fully uniform rewrite of every simple CASE through `invoke` would require adding lambda subquery support first. For now I kept the existing searched-CASE fallback for CASE bodies that contain subqueries, because rejecting those queries would be a behavior regression. If you would prefer not to keep that fallback, I think the cleaner path would be to first handle full lambda subquery support in a separate PR, then make the simple CASE rewrite uniform afterwards.

Maxxen:
I think deferring lambda subqueries is fine for now, I want to rework lambda bindings in general soon anyway.

Mytherin:
Let's try to split this off into separate PRs and make a PR introducing `invoke` first. We can then separately revisit using `invoke` to solve the original `CASE` issue.

feichai0017:
> Let's try to split this off into separate PRs and make a PR introducing `invoke` first. We can then separately revisit using `invoke` to solve the original `CASE` issue.

Agreed. I will split this into two PRs.

The first PR will only reintroduce the generic invoke(lambda, args...) scalar function, including the lambda binding/foldability plumbing and invoke-specific tests. I will keep the simple CASE rewrite, the __duckdb_simple_case_subject serialization fallback, and the simple CASE regression tests out of that PR.

After the invoke PR lands, I will rebase this PR and reduce it to only the simple CASE -> invoke rewrite for #22576.

feichai0017:
Hi @Mytherin 
Now that the invoke PR has been merged, I’ve updated this PR on top of the latest main and resolved the resulting conflicts. The diff is now focused on the simple CASE rewrite and its regression/compatibility coverage. This should be ready for another review.

Mytherin:
Thanks - I think this implementation is cleaner, but the serialization logic is not great. I will have to think about how to improve this though.

PR #23026, opened 2026-06-03
Use buffer allocator to track memory management - Parquet Prefetch

pdet:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/9233


Mytherin:
Thanks!

PR #23027, opened 2026-06-03
PEG/1: Make EndOfInput explicit, by having Tokenizer add `END_OF_INPUT` or `END_OF_INPUT_AUTOCOMPLETE`

carlopi:
Two new TokenType values are appended as a sentinel at the end of every tokenized vector:
* `END_OF_INPUT`, input genuinely ended; consumed by the new EndOfInput PEG rule via EndOfInputMatcher.
* `END_OF_INPUT_AUTOCOMPLETE`, cursor is here; ListMatcher and RepeatMatcher key off this token instead of running off the end of the vector.

BaseTokenizer::TokenizeInput appends the sentinel from the virtual GetEndOfInputType() on clean exit (autocomplete tokenizer overrides to `END_OF_INPUT_AUTOCOMPLETE`), and unconditionally `END_OF_INPUT` on the comment / dollar-quoted-string error exits.

Program now terminates with `EndOfInput`. EndOfInputMatcher is wired in via AddRuleOverride outside the generated block; IMPLICIT_RULES in the grammar generator skips the undefined-reference check for it. The transformer's error reporter walks back past the trailing sentinel so syntax errors name a real token. SQLTokenizeFunction, SQLFormatter and the linenoise highlighter ignore the sentinel; the autocomplete extension's check accepts state.token_index landing one short of tokens.size().

This adds some code, but do make explicit the terminator.
This would allow (in a follow up), to rewrite from:
```
Program <- Statement? (';'+ Statement)* ';'* EndOfInput
```
to
```
Program <- TopLevelStatement*
```

### DIFF
--- a/src/include/duckdb/parser/peg/transformer/parse_result.hpp
+++ b/src/include/duckdb/parser/peg/transformer/parse_result.hpp
@@ -32,7 +32,11 @@ enum class TokenType {
 	TABLE_FUNCTION,
 	PRAGMA_FUNCTION,
 	SETTING_NAME,
-	ERROR
+	ERROR,
+	//! Sentinel for real end of input — consumed by EndOfInputMatcher.
+	END_OF_INPUT,
+	//! Sentinel for cursor position in autocomplete mode — List/Repeat fire suggestion walk.
+	END_NOW_AUTOCOMPLETE
 };
 
 inline string TokenTypeToString(TokenType type) {
@@ -71,6 +75,10 @@ inline string TokenTypeToString(TokenType type) {
 		return "PRAGMA_FUNCTION";
 	case TokenType::SETTING_NAME:
 		return "SETTING_NAME";
+	case TokenType::END_OF_INPUT:
+		return "END_OF_INPUT";
+	case TokenType::END_NOW_AUTOCOMPLETE:
+		return "END_NOW_AUTOCOMPLETE";
 	default:
 		return "UNKNOWN";
 	}
@@ -91,6 +99,7 @@ enum class ParseResultType : uint8_t {
 	EXTENSION,
 	NUMBER,
 	STRING,
+	END_OF_INPUT,
 	INVALID
 };
 
@@ -120,6 +129,8 @@ inline const char *ParseResultToString(ParseResultType type) {
 		return "NUMBER";
 	case ParseResultType::STRING:
 		return "STRING";
+	case ParseResultType::END_OF_INPUT:
+		return "END_OF_INPUT";
 	case ParseResultType::INVALID:
 		return "INVALID";
 	}
@@ -178,6 +189,19 @@ struct IdentifierParseResult : ParseResult {
 	                      const std::string &indent, bool is_last) const override {
 		ParseResult::ToStringInternal(ss, visited, indent, is_last);
 		ss << ": " << identifier << "\n";
+	}
+};
+
+struct EndOfInputParseResult : ParseResult {
+	static constexpr ParseResultType TYPE = ParseResultType::END_OF_INPUT;
+
+	EndOfInputParseResult() : ParseResult(TYPE, optional_idx()) {
+	}
+
+	void ToStringInternal(std::stringstream &ss, std::unordered_set<const ParseResult *> &visited,
+	                      const std::string &indent, bool is_last) const override {
+		ParseResult::ToStringInternal(ss, visited, indent, is_last);
+		ss << "\n";
 	}
 };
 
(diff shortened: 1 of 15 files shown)


carlopi:
Regression I think not on this PR.

src/parser/peg/tokenizer/base_tokenizer.cpp
```diff
@@ -507,21 +507,30 @@ bool BaseTokenizer::TokenizeInput() {
...
+	};
+	auto append_end_of_input = [&]() {
+		tokens.emplace_back("", sql.size(), TokenType::END_OF_INPUT);
+	};
 	switch (state) {
 	case TokenizeState::SINGLE_LINE_COMMENT:
 	case TokenizeState::MULTI_LINE_COMMENT:
 		PushToken(last_pos, sql.size(), TokenType::COMMENT);
-		// no suggestions in comments or dollar-quoted strings
+		append_end_of_input();
```
Dtenwolde: Move this outside the function depending on the return false/true of this function

src/parser/peg/tokenizer/base_tokenizer.cpp
```diff
@@ -507,21 +507,30 @@ bool BaseTokenizer::TokenizeInput() {
...
 	case TokenizeState::DOLLAR_QUOTED_STRING:
 		PushToken(last_pos, sql.size(), TokenType::STRING_LITERAL, true);
+		append_end_of_input();
 		return false;
 	default:
 		break;
 	}
 	string last_word = sql.substr(last_pos, sql.size() - last_pos);
 	OnLastToken(state, std::move(last_word), last_pos);
+	append_sentinel();
```
Dtenwolde: Same here

src/parser/peg/transformer/peg_transformer_factory.cpp
```diff
@@ -73,6 +73,11 @@ vector<unique_ptr<SQLStatement>> PEGTransformerFactory::Transform(vector<Matcher
 		if (error_token_idx >= tokens.size()) {
 			error_token_idx = tokens.size() - 1;
 		}
+		// Walk back past the EOI sentinel so the error message names a real token.
+		while (error_token_idx > 0 && (tokens[error_token_idx].type == TokenType::END_OF_INPUT ||
```
Dtenwolde: There's only 1 end of input token

src/include/duckdb/parser/peg/transformer/parse_result.hpp
```diff
@@ -71,6 +75,10 @@ inline string TokenTypeToString(TokenType type) {
 		return "PRAGMA_FUNCTION";
 	case TokenType::SETTING_NAME:
 		return "SETTING_NAME";
+	case TokenType::END_OF_INPUT:
+		return "END_OF_INPUT";
+	case TokenType::END_NOW_AUTOCOMPLETE:
+		return "END_NOW_AUTOCOMPLETE";
```
Dtenwolde: Maybe rename to `EOI` & `EOI_AUTOCOMPLETE`

src/parser/peg/matcher.cpp
```diff
@@ -794,8 +842,11 @@ class NumberLiteralMatcher : public Matcher {
 
 private:
 	static bool MatchNumberLiteral(MatchState &state) {
+		if (state.token_index >= state.tokens.size()) {
+			return false;
+		}
 		auto &token_text = state.tokens[state.token_index].text;
-		if (!BaseTokenizer::CharacterIsInitialNumber(token_text[0])) {
+		if (token_text.empty() || !BaseTokenizer::CharacterIsInitialNumber(token_text[0])) {
```
Dtenwolde, line 849: `token_text.empty()` I'm not sure if we need this

Dtenwolde (commented):
Couple of comments

src/parser/peg/matcher.cpp
carlopi: I am not sure either, I think it should be guaranteed to be always not-empty, but I think the extra-guard is sort of nice to have.

src/include/duckdb/parser/peg/transformer/parse_result.hpp
carlopi: Moved to `END_OF_INPUT` / `END_OF_INPUT_AUTOCOMPLETE`, slightly more verbose.

src/parser/peg/tokenizer/base_tokenizer.cpp
carlopi: Yes, implemented

carlopi:
Thanks for the comments!! Adapted and rebased onto `main`.

Dtenwolde (approved):
Thanks, looks good to me!

PR #23028, opened 2026-06-03
[INSERT] Resolve nested defaults at a more granular level

Tishj:
This PR changes the `DEFAULT` resolution for columns of type `STRUCT`

type: `STRUCT(a VARCHAR, b BOOLEAN)`
default: `{'a': NULL, 'b': true}`
in: `{'a': 'test'}`

### Current behavior

result: `{'a': 'test', 'b': NULL}`

### New behavior

result: `{'a': 'test', 'b': true}`

### Details

At every level of the struct, we evaluate whether a value for a field was provided.
It's always used if provided, but now it's substituted by the DEFAULT if present, before falling back to NULL if it's not.

### DIFF
--- a/src/execution/physical_plan/plan_insert.cpp
+++ b/src/execution/physical_plan/plan_insert.cpp
@@ -151,19 +151,7 @@ PhysicalOperator &PhysicalPlanGenerator::ResolveDefaultsProjection(LogicalInsert
 					children.push_back(make_uniq<BoundConstantExpression>(CreateStructMapping(original_type, "")));
 
 					//! defaults
-					child_list_t<Value> default_children;
-					auto &struct_default_children = StructValue::GetChildren(bound_default.value);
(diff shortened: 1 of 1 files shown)


Tishj:
TODO:
We probably need to make this behavior pluggable somehow, so an extension (such as duckdb-iceberg) can override this, instead of changing the default behavior in core.

Also, this makes use of:
```c++
auto &original_type = child.children[0].get().types[mapped_index];
```
This is hacky at best, and relies on the parent to be the `COLUMN_DATA_SCAN` that contains the unmodified input types.
This is only true for `INSERT BY NAME` currently.
I think we should change the `INSERT BY POSITION` behavior to also delay the transformation from input types -> expected types until the Projection.
Furthermore I think we shouldn't be injecting this new projection for the DEFAULT values in the *planning* phase, but move this up to the *binding* phase.

PR #23029, opened 2026-06-03
PEG/2: Restructure top level `Program` definition

carlopi:
PR stacked on top of https://github.com/duckdb/duckdb/pull/23027, with an extra commit.

Now changing top level grammar from:
```
Program <- Statement? (';'+ Statement)* ';'* EndOfInput
```
to
```
Program <- TopLevelStatement*

TopLevelStatement <- Statement? (';'+ / EndOfInput)
```

Note there are less special cases (why the first `Statement` was special), and this allows to move to a `TopLevelStatement` level grammar where as long as we identify a valid `TopLevelStatement`, we can split there, allowing for interleaved parsing and execution (not yet as part of this PR).

PR #23032, opened 2026-06-03
PEG/3: Peeling statements stacked

carlopi:
Stacked on https://github.com/duckdb/duckdb/pull/23029 (that's stacked on https://github.com/duckdb/duckdb/pull/23027). Now also rebased on origin/main.

Idea is moving away from whole `Program` parsing, and into peeling one `TopLevelStatement` at a time.

This is relevant since it allows to interleave parsing and execution, for example:
```sql
LOAD <my_extension>; quack quack quack QUACK;
```
would parse correctly IF `<my_extension>` where to provide a parser_function that recognize `quack quack quack QUACK` as a valid statement.

In particular, working on `CONNECT`-statement, I found that it was restrictive to force all parsing to happen ahead of time, and we might want to allow interleaving of parse and execution (with side effects on parser itself).

Currently as per this PR, only infrastructure is added to introduce an StatementIterator of sort, but all code still does first tokenization (in full), then parsing (one at a time, until completion), then execution.
A next PR will interleave parsing and execution in certain code-paths.

A nice consequence of this, I think, is allowing parser extensions to be more proper first class citizens, where the logic changes from:
```
1. Does PEG recognise a query string? (possibly multiple statements)
2. If not 1, do any extension recognise a query string?
3. If not 2, errors.
```
to
```
while (true):
     1. Does PEG can peel out a statement from a query string? (possibly multiple statements)
     2. If not 1, do any extension recognise a statement from the start of the query string?
     3. if not 2, errors
     4. now, are there tokens left not consumed? if so, continue, else break
```

Say we have an extension that recognise:
```
quack;
quack QUACK;
quack quack QUACK QUack;
... basically, any amounts of quack (case insensitive)
```

```sql
SELECT 42; quack quack quack; SELECT 42;
```
Before this would not parse, given extension recognise only second part, and  PEG recognise the first and third, but no available parser is able to recognize the whole query.

Now, after the PR, statements are peeled once at a time:
```sql
SELECT 42;  ##--- default parser handled
           quack quack quack; ##--- extension handled
                              SELECT 42;  ##--- default parser handled
```
This basically means the PEG parser IS extensible at the `TopLevelStatement` level, that becomes sort of:
```
TopLevelStatement <- Statement? (';'+ / EndOfInput)
TopLevelStatement <- Statement? (';'+ / EndOfInput) / TopLevelExtensionStatement
```

that also means, if the extensions where to check for `(';'+ / EndOfInput)` as well, an extension can register a fully functional `Statement` equivalent.

This PR DOES rework the `parse_function`, so this will require extensions changes.

Alternative to https://github.com/duckdb/duckdb/pull/23000.

### DIFF
--- a/src/include/duckdb/parser/parser_extension.hpp
+++ b/src/include/duckdb/parser/parser_extension.hpp
@@ -13,9 +13,22 @@
 #include "duckdb/common/enums/statement_type.hpp"
 #include "duckdb/function/table_function.hpp"
 #include "duckdb/parser/sql_statement.hpp"
+#include "duckdb/parser/peg/token_type.hpp"
 
 namespace duckdb {
 struct DBConfig;
+
+//! A minimal token view handed to parser extensions: the token text together with its classified
+//! TokenType.
+struct SimpleToken {
+	SimpleToken(string text_p, TokenType type_p) : text(std::move(text_p)), type(type_p) {
+	}
+
+	//! The raw token text as it appears in the query
+	string text;
+	//! The classified type of the token
+	TokenType type;
+};
 
 //! The ParserExtensionInfo holds static information relevant to the parser extension
 //! It is made available in the parse_function, and will be kept alive as long as the database system is kept alive
@@ -74,9 +87,24 @@ struct ParserExtensionParseResult {
 	string error;
 	//! The error location (if unsuccessful)
 	optional_idx error_location;
+	//! How many leading tokens (from the start of the `tokens` view) the extension claimed:
+	//!    > 0 : SUCCESS — the extension accepted that many tokens; the peeler resumes parsing
+	//!          after them.
+	//!   == 0 : the extension ran fine but did not claim any tokens (it's not taking this input) —
+	//!          the peeler lets the next extension try.
+	//!    < 0 : the extension wants to surface an error — the peeler throws (using `error` /
+	//!          `error_location` if set).
+	//! A positive count must not exceed the number of tokens in the view, or the peeler throws.
+	int64_t consumed_tokens = 0;
 };
 
-typedef ParserExtensionParseResult (*parse_function_t)(ParserExtensionInfo *info, const string &query);
+//! Called when the PEG parser fails to parse a statement.
+//!
+//! `tokens` is the tokenized view of the source tail from the PEG failure point onward — one
+//! SimpleToken (text + TokenType) per token, in source order. The extension dispatches on this
+//! token stream and reports, via `ParserExtensionParseResult::consumed_tokens`, how many leading
+//! tokens it claimed (> 0 success, 0 = ran but claimed nothing, < 0 = throw).
+typedef ParserExtensionParseResult (*parse_function_t)(ParserExtensionInfo *info, const vector<SimpleToken> &tokens);
 //===--------------------------------------------------------------------===//
 // Plan
 //===--------------------------------------------------------------------===//
--- a/src/include/duckdb/parser/peg/matcher.hpp
+++ b/src/include/duckdb/parser/peg/matcher.hpp
@@ -116,8 +116,8 @@ struct MatcherSuggestion {
 
 struct MatchState {
 	MatchState(vector<MatcherToken> &tokens, vector<MatcherSuggestion> &suggestions, ParseResultAllocator &allocator,
-	           idx_t &max_token_index, bool preserve_identifier_case_p = true)
-	    : tokens(tokens), suggestions(suggestions), token_index(0), allocator(allocator),
+	           idx_t &max_token_index, bool preserve_identifier_case_p = true, idx_t starting_token_index = 0)
+	    : tokens(tokens), suggestions(suggestions), token_index(starting_token_index), allocator(allocator),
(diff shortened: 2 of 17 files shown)


carlopi:
`parse_function` change is a breaking change, I think improvement but worth double checking properly.

src/include/duckdb/parser/peg/matcher.hpp
```diff
@@ -226,13 +226,17 @@ struct PEGMatcher {
...
+		return *top_level_root;
+	}
 
 	static shared_ptr<PEGMatcher> Get(ClientContext &context);
 	static shared_ptr<PEGMatcher> Get(DatabaseInstance &db);
 
 private:
 	friend struct ParserCache;
 	optional_ptr<Matcher> root;
+	optional_ptr<Matcher> top_level_root;
```
Dtenwolde: Name is unclear

Dtenwolde: `ProgramMatcher` and `TopLevelStatementMatcher`

Dtenwolde: `StatementsMatcher`

src/include/duckdb/parser/parser_extension.hpp
```diff
@@ -74,9 +74,26 @@ struct ParserExtensionParseResult {
...
-typedef ParserExtensionParseResult (*parse_function_t)(ParserExtensionInfo *info, const string &query);
+//! Called when the PEG parser fails to parse a statement.
+//!
+//! `query` is the tail of the source from the PEG failure point onward.
+//! `allowed_boundaries` is the set of valid stopping points inside `query` — the start AND end
+//! offsets of every token, plus a final entry equal to `query.size()`. Sorted ascending,
+//! deduplicated (adjacent tokens with no whitespace between them contribute a single shared
+//! boundary). On success, the extension must report `consumed_chars` equal to one of these
+//! values. Reporting 0 declines and lets the next extension or the original PEG error surface.
+typedef ParserExtensionParseResult (*parse_function_t)(ParserExtensionInfo *info, const string &query,
```
Dtenwolde: I think this should pass only this information:
```
	TokenType type;
	string text;
```
From the `MatcherToken` for each token

Return the number of tokens consumed. Bigger than 0 means accepted. 0 did not parse/accept.

Dtenwolde: One question is what should happen if an extension accepts part of a query and returns another part of the query. 

`SELECT 42;QUACK QUACK QUACK;SELECT 42;`

PEG accepts `SELECT 42;` and passes `QUACK QUACK QUACK;select 42;` to the extension parser. The extension only accepts `QUACK QUACK`, returns a valid `unique_ptr<SQLStatement>` and `QUACK;SELECT 42;`. What should happen in this case? I think this might just be valid, but it's a bit unclear to me. Might also be a very made up example

src/parser/peg/matcher.cpp
```diff
@@ -1432,6 +1443,7 @@ Matcher &MatcherFactory::CreateMatcher(const char *grammar, const char *root_rul
 	//===--------------------------------------------------------------------===//
 	// END GENERATED RULE OVERRIDES
 	//===--------------------------------------------------------------------===//
+	AddRuleOverride("EndOfInput", allocator.Allocate(make_uniq<EndOfInputMatcher>()));
```
Dtenwolde: This shouldn't be here. See line below

src/parser/peg/matcher.cpp
```diff
@@ -1472,6 +1484,8 @@ shared_ptr<PEGMatcher> ParserCache::GetMatcher() {
 #else
 	new_matcher->root = factory.CreateMatcher(const_char_ptr_cast(INLINED_PEG_GRAMMAR), "Program");
 #endif
+	// TopLevelStatement is referenced by Program, so it has already been built and cached.
+	new_matcher->top_level_root = factory.GetMatcher("TopLevelStatement");
```
Dtenwolde: Should be `TopLevelStatement`, see comment above.

test/api/test_statement_iterator.cpp
```diff
@@ -0,0 +1,147 @@
...
+	auto it = ctx.ExtractStatements("SELECT FROM;");
+	REQUIRE_THROWS(it.Peek(ctx));
+}
+
+TEST_CASE("StatementIterator: movable", "[api][statement_iterator]") {
+	DuckDB db(nullptr);
+	Connection con(db);
+	auto &ctx = *con.context;
+
+	auto it = ctx.ExtractStatements("SELECT 1; SELECT 2;");
```
Dtenwolde: Please add a test for multiple semicolons, before, after, in the middle.

Dtenwolde (commented):
Looks good! Couple of comments especially around `ParserExtensionParseResult`.

src/include/duckdb/parser/parser_extension.hpp
carlopi: Thanks! I think now handled

carlopi:
I think this is ready to go.

Mytherin:
Thanks!

PR #23033, opened 2026-06-03
[Dev] Resolve `DEFAULT`s for INSERT/UPDATE/MERGE_INTO at bind-time rather than plan-time

Tishj:
This PR is part of #23028 

It deprecates `PhysicalPlanGenerator::ResolveDefaultsProjection` and replaces it with `Binder::ResolveDefaultsProjection`

It introduces 2 virtual methods on the `TableCatalogEntry`:
- 1 to produce a type which to bind an ExpressionList with, which can be `INVALID` to signal that the expressions should be bound as-is and the type for the column in the input relation should be inferred from combining the return_type of the expressions.
- 1 to produce an expression for the situation where an input value is present (the column isn't omitted) as well as a DEFAULT expression. By default this ignores the DEFAULT.

### Planning -> Binding

`DEFAULT` expressions used to be resolved with a pushed `PhysicalProjection` during Planning.
This has been moved up and now happens during Binding.
As a result `column_index_map` has been deprecated, as its only used for the creation of this extra `PhysicalProjection`.

### (De)Serialization

We introduce a `default` for the `column_index_map` in both `LogicalInsert` and `BoundMergeIntoAction` so it can be safely deprecated now and removed in a later version.


src/planner/binder/statement/bind_insert.cpp
```diff
@@ -93,12 +98,39 @@ void Binder::ExpandDefaultInValuesList(InsertQueryNode &node, TableCatalogEntry
...
+		auto mapped_index = insert.column_index_map[col.Physical()];
+		if (mapped_index == DConstants::INVALID_INDEX) {
+			// Push default value
+			select_list.push_back(std::move(insert.bound_defaults[storage_idx]));
+			continue;
+		}
+
+		auto &original_type = source_types[mapped_index];
+		auto source_binding = source_bindings[mapped_index];
+		select_list.push_back(make_uniq<BoundColumnRefExpression>(original_type, source_binding));
```
Mytherin: Let's move this to a callback in the `TableCatalogEntry` that does this

src/planner/binder/statement/bind_insert.cpp
```diff
@@ -82,7 +85,9 @@ void Binder::ExpandDefaultInValuesList(InsertQueryNode &node, TableCatalogEntry
 
 			// set the expected types as the types for the INSERT statement
 			auto &column = table.GetColumn(table_col_idx);
-			expr_list.expected_types[col_idx] = column.Type();
+			expr_list.expected_types[col_idx] = preserve_struct_types && column.Type().id() == LogicalTypeId::STRUCT
```
Mytherin: This should just always be `LogicalType::INVALID`

src/planner/binder/statement/bind_insert.cpp
Mytherin: This should then also add a cast

src/planner/binder/statement/bind_insert.cpp
```diff
@@ -613,7 +647,20 @@ BoundStatement Binder::BindNode(InsertQueryNode &node) {
 		// inserting from a select - check if the column count matches
 		CheckInsertColumnCountMismatch(expected_columns, root_select.types.size(), !node.columns.empty(), table.name);
 
-		root = CastLogicalOperatorToTypes(root_select.types, insert->expected_types, std::move(root_select.plan));
+		auto source_types = root_select.types;
+		auto target_types = insert->expected_types;
+		if (!insert->column_index_map.empty()) {
+			root_select.plan = ResolveDefaultsProjection(*insert, std::move(root_select.plan), source_types);
```
Mytherin: This method name should then also probably be changed to `ResolveInputProjection`

Mytherin: And this should supersede `CastLogicalOperatorToTypes`

src/planner/binder/statement/bind_insert.cpp
```diff
@@ -93,12 +98,39 @@ void Binder::ExpandDefaultInValuesList(InsertQueryNode &node, TableCatalogEntry
...
+			select_list.push_back(std::move(insert.bound_defaults[storage_idx]));
+			continue;
+		}
+
+		auto &original_type = source_types[mapped_index];
+		auto source_binding = source_bindings[mapped_index];
+		select_list.push_back(make_uniq<BoundColumnRefExpression>(original_type, source_binding));
+	}
+
+	auto projection = make_uniq<LogicalProjection>(GenerateTableIndex(), std::move(select_list));
```
Mytherin, line 149: Maybe an optimization that if this just references all columns we don't create the projection

Mytherin: Or maybe we have an optimizer

src/planner/binder/statement/bind_insert.cpp
Mytherin: We should not do this only for struct types that's too opinionated - let's just always move the casts to the table catalog entry callback

src/planner/binder/tableref/bind_expressionlistref.cpp
```diff
@@ -27,25 +27,44 @@ BoundStatement Binder::Bind(ExpressionListRef &expr) {
...
 
-		SetCanContainNulls(true);
+		this->can_contain_nulls = true;
 		vector<unique_ptr<Expression>> list;
 		for (idx_t val_idx = 0; val_idx < expression_list.size(); val_idx++) {
 			if (!result.types.empty()) {
 				D_ASSERT(result.types.size() == expression_list.size());
 				binder.target_type = result.types[val_idx];
+			} else {
+				binder.target_type = LogicalType(LogicalTypeId::INVALID);
```
Mytherin, line 37: This file probably shouldn't be changed - we should maybe move `ExpandDefaultInValuesList` to the `TableCatalogEntry`

Mytherin (commented):

Tishj:
I am deprecating `column_index_map`, as it is no longer necessary to save, since we now handle this at bind time.

I'll still need to keep the deserialize logic and code that acts on it in the planning phase, to have backwards-compatibility for plans written by older duckdb versions.
But we won't be populating them in new duckdb versions going forward.

src/storage/serialization/serialize_logical_operator.cpp
```diff
@@ -209,7 +209,7 @@ void BoundMergeIntoAction::Serialize(Serializer &serializer) const {
 	serializer.WritePropertyWithDefault<unique_ptr<Expression>>(201, "condition", condition);
 	serializer.WritePropertyWithDefault<vector<PhysicalIndex>>(202, "columns", columns);
 	serializer.WritePropertyWithDefault<vector<unique_ptr<Expression>>>(203, "expressions", expressions);
-	serializer.WriteProperty<IndexVector<idx_t, PhysicalIndex>>(204, "column_index_map", column_index_map);
+	serializer.WritePropertyWithDefault<IndexVector<idx_t, PhysicalIndex>>(204, "column_index_map", column_index_map, IndexVector<idx_t, PhysicalIndex>());
```
Mytherin: This will break compat with older versions because older versions will always expect this field to be there.

I ran into the same issue in https://github.com/duckdb/duckdb/pull/23161 and fixed it by adding a new field to the serialization properties there:

```cpp
"required_until": "v2.0.0",
```

This will basically use `WriteProperty` for older storage versions, and `WritePropertyWithDefault` for newer ones, allowing migration away from the required properties.

Mytherin:
Thanks!

PR #23036, opened 2026-06-03
Make HTTP 502 responses retryable

### LINKED ISSUE #22673: HTTP 502 responses are failed without retry
### What happens?

An HTTP request that receives an HTTP 502 response is failed without retry. HTTP 502 errors should be retried because they are transient.

This can be a big problem in ducklake scenarios that ingest large amounts of data into object storage (e.g. Cloudflare R2). A single 502 can break the whole ingestion pipeline, because the request is not retried.

```
_duckdb.HTTPException: HTTP Error: Unable to connect to URL r2://XXX/datalake/tables/srd/daily/ducklake-019e26ad-1eef-728e-88ae-46b1899b3368.parquet: Bad Gateway (HTTP code 502)
```

### To Reproduce

I do not have specific reproduction steps, but I can point to the code that I believe is responsible: [`HTTPResponse::ShouldRetry()`](https://github.com/duckdb/duckdb/blob/v1.5-variegata/src/main/http/http_util.cpp#L110-L126)

Adding the following `case` should fix this issue:

```C
case HTTPStatusCode::BadGateway_502:
```

### OS:

macOS Tahoe 26.0.1

### DuckDB Version:

1.5.2

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Bill Zissimopoulos

### Affiliation:

N/A

- [x] Yes, I have

- [ ] Yes, I have

billziss-gh:
Fixes #22673

DuckDB already retries 500 and 504 responses. Retrying 502 is consistent with that behavior, since it is another transient server-side/gateway failure. Practically a lot of object storage services return 502 intermittently and this makes HTTP accesses resilient.

### DIFF
--- a/src/main/http/http_util.cpp
+++ b/src/main/http/http_util.cpp
@@ -117,6 +117,7 @@ bool HTTPResponse::ShouldRetry() const {
 	case HTTPStatusCode::ImATeapot_418:
 	case HTTPStatusCode::TooManyRequests_429:
 	case HTTPStatusCode::InternalServerError_500:
+	case HTTPStatusCode::BadGateway_502:
 	case HTTPStatusCode::ServiceUnavailable_503:
 	case HTTPStatusCode::GatewayTimeout_504:
 		return true;


Mytherin:
Thanks!

PR #23041, opened 2026-06-03
PEG/4: StatementIterator interface and use to interleave parsing and execution

carlopi:
Stacked on top of https://github.com/duckdb/duckdb/pull/23032, only last commit is relevant.

This allows a given `query` string to be unpacked into multiple statements, extracted and executed one at a time, that means that for example the state of the parser might change.

Example:
```
carlo@MacBook-Pro-2 duckdb-stable % ./build/release/duckdb -unsigned
DuckDB v1.6.0-dev8256 (Development Version, 4523194c2f)
memory D QUACK quack quack;
Parser Error:
syntax error at or near "quack"

LINE 1: QUACK quack quack;
              ^
memory D LOAD './build/release/test/extension/loadable_extension_demo.duckdb_extension'; QUACK quack quack;
┌─────────┐
│  quack  │
│ varchar │
├─────────┤
│ QUACK   │
│ QUACK   │
│ QUACK   │
└─────────┘
memory D 
memory D 
```

Where `LOAD loadable_extension_demo.duckdb_extension` changes to the parser are directly visible in the same statement.

### DIFF
--- a/src/main/statement_iterator.cpp
+++ b/src/main/statement_iterator.cpp
@@ -1,0 +1,180 @@
+#include "duckdb/main/statement_iterator.hpp"
+
+#include "duckdb/main/client_context.hpp"
+#include "duckdb/main/extension_callback_manager.hpp"
+#include "duckdb/main/query_profiler.hpp"
+#include "duckdb/parser/parser.hpp"
+#include "duckdb/parser/parser_extension.hpp"
+#include "duckdb/parser/peg/matcher.hpp"
+#include "duckdb/parser/peg/tokenizer/parser_tokenizer.hpp"
+#include "duckdb/parser/peg/transformer/parse_result.hpp"
+#include "duckdb/parser/sql_statement.hpp"
+#include "duckdb/parser/statement/create_statement.hpp"
+#include "duckdb/parser/parsed_data/create_info.hpp"
+
+namespace duckdb {
+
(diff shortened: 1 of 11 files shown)


carlopi:
Need to fix the tests for `.bail off`, that I am not really sure if can be fixed or we should just accept failure there.

carlopi:
Found the issue, needed to add error location, back on track.

src/main/statement_iterator.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include "duckdb/parser/parsed_data/create_info.hpp"
+
+namespace duckdb {
+
+StatementIterator::StatementIterator(string sql_p, bool preprocess)
+    : sql(std::move(sql_p)), preprocess_on_peek(preprocess) {
+	// Mirror Parser::ParseQuery's up-front normalization: reject invalid UTF-8 (otherwise the
+	// tokenizer can recurse on bad bytes — see ossfuzz clusterfuzz-test-24) and strip non-ASCII
+	// Unicode spaces.
+	Parser::ValidateUTF8Query(sql);
```
Mytherin, line 22: Both of these should be in the parser not here

Mytherin (commented):
Thanks - looks good, one minor comment

carlopi:
I think I like more the version at https://github.com/duckdb/duckdb/pull/23291, but let's see what CI says, then I might close this one, since the other is more polished.

PR #23048, opened 2026-06-03
use StringUtil::CharacterToLower in strptime parser

jmestwa-coder:
the strptime parser passes a raw signed char to std::tolower when matching month/day names in TryParseCollection (%B/%A/%a/%b) and the AM/PM branch (%p). data is the untrusted first argument to strptime/try_strptime, so any non-ascii byte (e.g. ä) reaches tolower as a negative value, which is undefined behavior per STR37-C and locale-dependent. switch to StringUtil::CharacterToLower, the ascii range check already used across src/common/types.


Mytherin:
Thanks!

PR #23052, opened 2026-06-04
Rework pr22857: fix external file cache memleak

### LINKED ISSUE #22730: Unbounded memory growth in ExternalFileCache::cached_files

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22730
This PR reworks https://github.com/duckdb/duckdb/pull/22857

Currently there's memleak for external file cache: it stores all [`cache_files`](https://github.com/duckdb/duckdb/blob/11ee25cd67a260bd629671fa8176a8781e623bc5/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp#L102-L103), whose entries don't get evicted even there're no valid data blocks.

There're a few ways I could think of, along with community contributions. Overall, we need to balance a few tradeoffs: (1) eviction overhead; (2) memory pressure; (3) implementation complexity; (4) bespoke solution vs general solution

Thought about a few possible solutions and their implementation / tradeoff analysis:
- Hook callback on block load and unload, on unload if the map entry has no data blocks and no longer referenced by any file handle, it's evicted right away
  + Implementation: https://github.com/duckdb/duckdb/pull/22856
  + pro: (1) no performance hurt; (2) little memory pressure, all entries are evicted right away; (3) concept easy to understand
  + con: implementation complexity, mostly because we don't have a callback mechanism for block (un)load
- Buffer pool manager checks unnecessary map entries on block eviction
  + Implementation: https://github.com/dentiny/duckdb/pull/106
  + pro: (1) no performance hurt; (2) little memory pressure, compatible with other buffer manager owned memory; (3) easy concept and implementation
  + con: bespoke implementation; I spent quite some time to justify and debate whether it's a good solution, the main concern is, map structure is so common in DuckDB, and we already have a general solution to manager KV-structure which is object cache, better to follow the same pattern
- object cache based implementation (this PR)
  + pro: (1) no performance hurt; (2) little memory pressure; (3) easy concept
  + con: medium implementation complexity, need to keep certain invariants to ensure correctness
- clear unnecessary under circumstance
  + Implementation: https://github.com/duckdb/duckdb/pull/22857
  + pro: easy concept and implementation
  + con: (1) no centralized memory management (for DuckDB, it's buffer pool manager for now); (2) unpredictable latency, largely depending on the number of cache file entries in the map, and active file handles -- I'm worried if file handles is cached we will suffer high latency every cache miss (I did it in my extension cache httpfs https://duckdb.org/community_extensions/extensions/cache_httpfs)

Overall, I lean towards (2) and (3), I slightly prefer (3) because it's a general purpose solution with acceptable implementation complexity and easy concept. (1) could be the long term plan, but require certain code reference beforehand Happy to discuss.

I have ran CI in my own fork: https://github.com/dentiny/duckdb/pull/105, failure unrelated


extension/parquet/parquet_file_metadata_cache.cpp
```diff
@@ -30,6 +30,10 @@ string ParquetFileMetadataCache::ObjectType() {
 	return "parquet_metadata";
 }
 
+string ParquetFileMetadataCache::ObjectCacheKey(const string &path) {
+	return StringUtil::Format("parquet-metadata-%s", path);
```
lnkuiper: Maybe this can be a `static` member function of `ObjectCacheEntry` that derived classes implement? Then we can do all the logic to handle the keys in the `ObjectCache` itself, instead of putting the responsibility on the caller, which is more error-prone. Perhaps we can reuse the `ObjectType` function for this?

src/storage/external_file_cache/external_file_cache.cpp
```diff
@@ -10,9 +10,45 @@
...
+		return "external_file_cache";
+	}
+
+	string GetObjectType() override {
+		return ObjectType();
+	}
+
+	optional_idx GetEstimatedCacheMemory() const override {
+		return object_cache_key.size() * 2 + cached_file->path.size();
+	}
```
lnkuiper, line 38: I think this underestimates the size of the entries. It just uses the sizes of the strings, but it could be much larger since the cached file holds a map of byte ranges. For this to evict properly we should allow resizing of object cache entries to reflect their size as it grow (I've attached a Codex-generated diff that does this, but I'm not sure if it's good; it might help), or we should implement the "opportunistic stale entry removal" that the original PR implemented, which would be to try and evict one stale cache entry at the back of the LRU queue each time the external file cache is touched. I'm not sure which approach is better, I'll leave that up to you.

[my.txt](https://github.com/user-attachments/files/28590514/my.txt)

lnkuiper (changes requested):
Thanks for the PR! Looks great. I have left two comments below that should be addressed before we can merge this.

src/storage/external_file_cache/external_file_cache.cpp
dentiny: > I think this underestimates the size of the entries. It just uses the sizes of the strings, but it could be much larger since the cached file holds a map of byte ranges.

Yes this is something I considered
- Currently object cache and LRU cache only support static size account, dynamic resizing, if necessary, should be another followup PR IMO
- The overall size of the object cache entry = static assignment (the memory estimation above) + N * data blocks (including metadata + data block); considering the data block size is significantly smaller than data block (and data block is already under buffer manager's track), I think it might be fine to ignore

In a word:
- It's a known issue we could improve in the followup PR, via cache entry resize. I foresee it could be a useful feature
- The target for this PR is to make cache file map entries accountable under buffer manager

extension/parquet/parquet_file_metadata_cache.cpp
dentiny: sure, updated

src/storage/external_file_cache/external_file_cache.cpp
dentiny: Or another way we could do is to introduce metadata bookkeeping memory account for data blocks.

lnkuiper (approved):
Thanks!

src/storage/external_file_cache/external_file_cache.cpp
lnkuiper: I was thinking that maybe it's good enough if we assume we have the whole file cached for the size estimate? It's an overestimate but the metadata shouldn't be too large anyway. We could do size estimate = file size / current block size * cached range metadata size. That would have a small code footprint and work pretty well as it would never underestimate the size.

dentiny: > That would have a small code footprint and work pretty well as it would never underestimate the size.

Yeah that works as well

PR #23054, opened 2026-06-04
Add `CATALOG_MODE` and `CONNECT_MODE` attach options

carlopi:
Allow to explicitly opt-in (or opt-out) to make an Attached Database queriable OR connetable.

This could make sense both for assert a given backend has capacities OR for restrict a given AttachedDatabase capabilities.
This is a building block for `CONNECT '<connection_string>';`, since in that case we would like to set the underlying attach to: `CONNECT_MODE ENABLED, CATALOG_MODE NONE`.

Defaults for both is `AUTO`, that means, defer to the capabilities of a given Catalog. But you can decide to say do:
```sql
ATTACH 'file1.db' (CATALOG_MODE NONE);
ATTACH 'file2.db' (CATALOG_MODE HIDDEN);
ATTACH 'file3.db' (CATALOG_MODE AUTO);
ATTACH 'file4.db' (CATALOG_MODE ENABLE);
```
Now, file1 tables / views are not visible, file2's are visible but not auto-discovered (same as `HIDDEN`), file3 and file4 are regular.

Now:
```sql
FROM duckdb_databases();
```
gives:
```
┌───────────────┬──────────────┬──────────────────────────────────┬─────────┬───┬─────────┬──────────┬───────────┬─────────┬──────────────────────┐
│ database_name │ database_oid │               path               │ comment │ … │  type   │ readonly │ encrypted │ cipher  │       options        │
│    varchar    │    int64     │             varchar              │ varchar │ … │ varchar │ boolean  │  boolean  │ varchar │ map(varchar, varcha… │
├───────────────┼──────────────┼──────────────────────────────────┼─────────┼───┼─────────┼──────────┼───────────┼─────────┼──────────────────────┤
│ file4         │        22184 │ /Users/carlo/duckdb-stable/file… │ NULL    │ … │ duckdb  │ false    │ false     │ NULL    │ {}                   │
│ file3         │        22180 │ /Users/carlo/duckdb-stable/file… │ NULL    │ … │ duckdb  │ false    │ false     │ NULL    │ {}                   │
│ memory        │        20680 │ NULL                             │ NULL    │ … │ duckdb  │ false    │ false     │ NULL    │ {}                   │
│ system        │            0 │ NULL                             │ NULL    │ … │ duckdb  │ false    │ false     │ NULL    │ {}                   │
│ temp          │        22162 │ NULL                             │ NULL    │ … │ duckdb  │ false    │ false     │ NULL    │ {}                   │
└───────────────┴──────────────┴──────────────────────────────────┴─────────┴───┴─────────┴──────────┴───────────┴─────────┴──────────────────────┘
```
with file2 still reachable but `HIDDEN`, and file1 not reachable. Example:
```
memory D FROM file1.T;
Binder Error:
Catalog "file1" does not exist!
memory D FROM file2.T;
Catalog Error:
Table with name T does not exist!

LINE 1: FROM file2.T;
             ^
memory D FROM file3.T;
Catalog Error:
Table with name T does not exist!

LINE 1: FROM file3.T;
             ^
```
Where `file1` is not a catalog, `file2` is an hidden catalog, and `file3` and `file4` regular catalog.

Similarly for `CONNECT_MODE`:
```
memory D ATTACH 'quack:localhost' as a (TOKEN '1234');
memory D CONNECT a;
a memory D DISCONNECT;
memory D DETACH a;
memory D
memory D ATTACH 'quack:localhost' as b (TOKEN '1234', CONNECT_MODE NONE);
memory D CONNECT b;
Invalid Input Error:
Database "b" does not support CONNECT
memory D DETACH b;
```
here the first can be `CONNECT`-ed regularly, but `CONNECT_MODE NONE` make CONNECT not viable. Note, quack will not (at this moment) enforce this on `quack_query_by_name`, so arbitrary queries can still be sent to be backend, but I think a `quack` change that would make sense is reducing surface depending on `connect_mode`.

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -1192,7 +1192,7 @@ vector<reference<SchemaCatalogEntry>> Catalog::GetAllSchemas(ClientContext &cont
 	auto &db_manager = DatabaseManager::Get(context);
 	auto databases = db_manager.GetDatabases(context);
 	for (auto &database : databases) {
-		if (database->GetVisibility() == AttachVisibility::HIDDEN) {
+		if (database->IsHidden()) {
 			continue;
 		}
 		auto &catalog = database->GetCatalog();
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -93,6 +93,11 @@ optional_ptr<Catalog> Catalog::GetCatalogEntry(CatalogEntryRetriever &retriever,
 	if (!entry) {
 		return nullptr;
 	}
+	if (entry->GetCatalogMode() == CatalogMode::NONE) {
+		// CATALOG_MODE NONE: catalog provides no tables and is unreachable from binding.
+		// Treat as if the catalog doesn't exist for lookup purposes (reads + writes both fail).
+		return nullptr;
+	}
 	return &entry->GetCatalog();
 }
 
--- a/src/common/enum_util.cpp
+++ b/src/common/enum_util.cpp
@@ -1376,6 +1376,25 @@ ConflictManagerMode EnumUtil::FromString<ConflictManagerMode>(const char *value)
 	return static_cast<ConflictManagerMode>(StringUtil::StringToEnum(GetConflictManagerModeValues(), 2, "ConflictManagerMode", value));
 }
 
+const StringUtil::EnumStringLiteral *GetConnectModeValues() {
+	static constexpr StringUtil::EnumStringLiteral values[] {
+		{ static_cast<uint32_t>(ConnectMode::NONE), "NONE" },
+		{ static_cast<uint32_t>(ConnectMode::AUTO), "AUTO" },
+		{ static_cast<uint32_t>(ConnectMode::ENABLE), "ENABLE" }
+	};
+	return values;
+}
+
+template<>
+const char* EnumUtil::ToChars<ConnectMode>(ConnectMode value) {
+	return StringUtil::EnumToString(GetConnectModeValues(), 3, "ConnectMode", static_cast<uint32_t>(value));
+}
+
+template<>
+ConnectMode EnumUtil::FromString<ConnectMode>(const char *value) {
+	return static_cast<ConnectMode>(StringUtil::StringToEnum(GetConnectModeValues(), 3, "ConnectMode", value));
+}
+
 const StringUtil::EnumStringLiteral *GetConstraintTypeValues() {
 	static constexpr StringUtil::EnumStringLiteral values[] {
 		{ static_cast<uint32_t>(ConstraintType::INVALID), "INVALID" },
--- a/src/common/enum_util.cpp
+++ b/src/common/enum_util.cpp
@@ -1052,6 +1052,26 @@ CatalogLookupBehavior EnumUtil::FromString<CatalogLookupBehavior>(const char *va
 	return static_cast<CatalogLookupBehavior>(StringUtil::StringToEnum(GetCatalogLookupBehaviorValues(), 3, "CatalogLookupBehavior", value));
 }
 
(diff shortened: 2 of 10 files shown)


carlopi:
We could consider to retire `HIDDEN` after this land. Currently `HIDDEN` is also valid option, and resolution will make so that only consistent options are OK.

Mytherin:
Thanks for the PR!

I'm not quite sure why `HIDDEN` is part of this? To me it seems quite orthogonal to remote capabilities.

`CATALOG_MODE` is also maybe not a right fit for a name here. From reading just the option name it's completely unclear to me what `CATALOG_MODE ENABLE ` or `CATALOG_MODE NONE` would do. Maybe this should be `REMOTE_CAPABILITIES` or so instead? Or be extremely explicit in the exact semantics of what the option is trying to achieve - e.g. `DISABLE_CONNECT`, `DISABLE_REMOTE_PUSHDOWN`, etc?

Mytherin:
For remote pushdown I think we are most interested in the ability to disable hybrid execution, since that can silently lead to pulling lots of data from remote to local, e.g. for what Im working on now for MySQL:

```sql
-- runs remotely
SELECT SUM(i) FROM remote_tbl;

-- runs locally, pulling remote_tbl to local, because MySQL doesn't support `median`
SELECT MEDIAN(i) FROM remote_tbl;
```

I think we want some way of flagging the latter as an error, so maybe the right flag here is something like `DISABLE_HYBRID` or `REMOTE_EXECUTION_ONLY`?

carlopi:
I don't this this is, strictly speaking, about remote capabilities, or at least in my mind this is about what the interface looks like SQL wise. Like "Can I `CONNECT` to this" or "Can I create/insert/query tables from this catalog?".

Minimal functionality I had in mind would be allowing:
```
CONNECT '<some string>';
```
to be translated to:
```
ATTACH '<some string>' as ___hidden ( <some params> );
CONNECT ___hidden;
```
but avoid `__hidden` influencing "local" state or state of different connections, like:
```
CONNECT LOCAL EXECUTE FROM duckdb_tables();
--- hidden not reported
CONNECT LOCAL EXECUTE FROM ___hidden.tbl;
--- this must also error
```

`HIDDEN` param currently does only the first part of this part of this, given tables are still reachable.

What I though was that, to simplify, we could keep to 2 axis: `CATALOG`-capabilities, and `CONNECT`-capabilities. `HIDDEN` in my mind folded with `CATALOG`-capabilities, but that might not be desirable.

Mytherin:
So your idea is to add a "super-duper hidden" flag, effectively, that hides the catalog in more situations?

I would say we can also just expand `HIDDEN` so that it hides tables from `duckdb_tables` and such. That seems well within the scope of `HIDDEN`. The fact that that's not happening now is more of an oversight and less of an intentional choice.

Mytherin:
Ah nevermind, you're saying that `HIDDEN` tables are still queryable in some capacity, just not shown, and you want to prevent them from being queried entirely, i.e. this is not about *hiding* tables this is about making them unreachable.

I would say this is not really the same thing then - it's more of "can I use this attached database as an attached database" or "can I connect to this database". `HIDDEN` is orthogonal to that imo.

I think this idea needs a bit more fleshing out, I would say we should do some more thinking / iterations on this and what's actually desirable behavior / user-intuitive here. I can see that from a user perspective having attach and connect being two different but related things could definitely be confusing and we should have clear semantics there, but I think we need to think a bit more about how to make that more clear and take away footguns from the user.

Mytherin:
I think also this is kind of orthogonal to `CONNECT <string>` - it's more of a "nice to have" there and not really a pre-requisite

carlopi:
Yes, I did realize the same, that `HIDDEN` would do for now, and this can move on a side but it's not a blocker.

PR #23057, opened 2026-06-04
Expose bytes to parquet variant function

Tmonster:
This PR slightly modifies the VariantMetadata Structure to include the length of the metadata and exposes a function `variant_bytes_to_variant` to read a variant value encoded as `metadata + value` binary values. 

This function is needed so that DuckDB-Iceberg can data files based on variant upper and lower bounds. The Iceberg spec specifies that these values are encoded as `metadata + value`. DuckDB has a lot of this logic in the parquet extension, and I would rather not duplicate that logic in DuckDB-Iceberg, and I would rather expose it as a scalar function.

### DIFF
--- a/extension/parquet/reader/variant/variant_shredded_conversion.cpp
+++ b/extension/parquet/reader/variant/variant_shredded_conversion.cpp
@@ -339,19 +339,15 @@ static vector<VariantValue> ConvertBinaryEncoding(Vector &metadata, Vector &valu
 		if (!validity.RowIsValid(index)) {
 			continue;
 		}
+		//! 'metadata' and 'value' are the same vector: each row holds the full binary Variant value (metadata
+		//! followed by the value blob). Decode the metadata, then read the value right after it.
+		auto &metadata_value = metadata_data[metadata_format.sel->get_index(i)];
+		VariantMetadata variant_metadata(metadata_value);
+		auto binary_value = value_data[index].GetData();
 		if (add_metadata_offset) {
-			//! 'metadata' and 'value' are the same vector: each row holds the full binary Variant value (metadata
-			//! followed by the value blob). Decode the metadata, then read the value right after it.
-			auto &binary_blob = value_data[index];
-			VariantMetadata variant_metadata(binary_blob);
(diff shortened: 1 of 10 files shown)


extension/parquet/reader/variant/variant_shredded_conversion.cpp
```diff
@@ -337,7 +336,17 @@ static vector<VariantValue> ConvertBinaryEncoding(Vector &metadata, Vector &valu
...
-		if (validity.RowIsValid(index)) {
+		if (!validity.RowIsValid(index)) {
+			continue;
+		}
+		if (add_metadata_offset) {
+			//! 'metadata' and 'value' are the same vector: each row holds the full binary Variant value (metadata
+			//! followed by the value blob). Decode the metadata, then read the value right after it.
+			auto &binary_blob = value_data[index];
+			VariantMetadata variant_metadata(binary_blob);
+			auto binary_value = const_data_ptr_cast(binary_blob.GetData()) + variant_metadata.total_size;
```
Tishj: I feel like this duplication can be avoided, can't we just add an `if (add_metadata_offset) { binary_value += variant_metadata.total_size; }`?

extension/parquet/reader/variant_column_reader.cpp
```diff
@@ -204,4 +204,18 @@ bool VariantColumnReader::TypedValueLayoutToType(const LogicalType &typed_value,
...
 }
 
+static void FromParquetVariant(DataChunk &input, ExpressionState &state, Vector &result) {
+	auto num_values = input.size();
+	auto &metadata_value = input.data[0];
+
+	VariantShreddedConversion::ConvertBinaryToVariant(metadata_value, 0, num_values, num_values, result);
+}
+
+ScalarFunction VariantColumnReader::GetBytesToVariantFunction() {
```
Tishj: I think this shouldn't be part of VariantColumnReader
It seems confusing because the VariantColumnReader doesn't use this method at all

Tishj (commented):
Thanks, glad we found a clean way of creating this method with a small footprint, just have a few small comments

lnkuiper:
Windows failure seems related

Tmonster:
I think the failing CI just needs to be restarted

PR #23059, opened 2026-06-04
Support named/default arguments for window functions too

Maxxen:
Followup to #22941 that adds support for window functions too.

To test, this PR only changes the `lead/lag` window functions to now be a single overload with signature: 

```sql
lead(col T, "offset" BIGINT := 1, "default" ANY := NULL) -> T
```
- Since both `offset` and `default` are reserved keywords in SQL these have to be passed in quotes.
- `default` can't be `T` but must be `ANY` as the function currently force-default casts it to whatever the return type is. So it should not be used together with the input value to infer the result type `T`. In theory it probably _should_ be `T` too and not force-cast, but I didn't want to change any semantics in this PR.

In the future we should change more functions as well.


src/parser/expression/window_expression.cpp
```diff
@@ -132,4 +147,88 @@ bool WindowExpression::HasBoundedParts() {
...
+
+	if (!serializer.ShouldSerialize(StorageVersion::V2_0_0)) {
+		// Legacy serialization.
+		vector<unique_ptr<ParsedExpression>> children;
+		for (auto &arg : arguments) {
+			auto copy = arg.GetExpression().Copy();
+			copy->SetAlias(arg.GetName());
+			children.push_back(std::move(copy));
+		}
+		serializer.WritePropertyWithDefault<vector<unique_ptr<ParsedExpression>>>(203, "children", children);
```
Mytherin, line 164: Can't we keep this as being generated using the serialization script? We should be able to add functions for these using the properties, no?

I just realized this is also now manually written for `FunctionExpression` - I think we can keep that within the generation script as well.

Mytherin (commented):

src/parser/expression/window_expression.cpp
Maxxen: I dont think we can, I prodded around with it a bit but was unsuccessful. 

One alternative would be to make FunctionArgument an expression too in the future (AliasExpression ;)) at which point there is no serialization changes needed!

Maxxen: But ok, if we can merge this now I'll have a look at the generator again and see if we can make any changes to support this case too.

Mytherin: That just moves the problem since we need to then turn `AliasExpression` into setting the alias on the child for BWC :)

Sure, we can look into this in a follow-up

Mytherin: I think we should be able to do this by setting `serialize_property` to a function that generates `children` (`LegacySerializeChildren()`) - and on deserialization using a `constructor` that takes both `children` and `arguments` and then uses whichever is provided

Mytherin: An alternative path would be to keep `children` as a `vector<unique_ptr<ParsedExpression>>` and move arguments to a separate structure instead of unifying it

PR #23060, opened 2026-06-04
[Bug] Fix string-heap copy for updates mixing NULL and non-NULL values

pdet:
Found while trying to get DuckLake up to date with DuckDB main.


Mytherin:
Thanks!

PR #23061, opened 2026-06-04
Refactor of Parquet Scan for ASYNC I/O

pdet:
In this PR, we split the parquet scan function into two phases: 1. scheduling and 2. processing as a preparation step for async I/O. This is just a refactor, and the behavior is unchanged. The `Schedule()` function executes the I/O tasks directly and returns `HAVE_MORE_OUTPUT`, so we do the same as the old sync prefetch for now.


Mytherin:
Nice, thanks!

PR #23069, opened 2026-06-04
Add `decimal_division` scalar function

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/8441
Previous PR here: https://github.com/duckdb/duckdb/pull/22955

Second attempt at adding `decimal_division`, but now as an explicit scalar function, rather than overriding `/`:
`decimal_division(x, y)` 
`decimal_division(x, y, scale)`

If the `scale` is provided by the user this is the exact `scale` used for the output. 
If the `scale` is not given by the user explicitly we still use the SQL server formulas:
```
  result_scale = max(6, s1 + p2 + 1)
  result_width = p1 - s1 + s2 + result_scale
```

We now throw an `OutOfRangeException` if `scale` or `width` is greater than 38, rather than capping the precision.

We operate on `int64_t` whenever it is safe from overflow and fallback to `hugeint_t` when needed. 

I'll work on `decimal_avg` in a follow-up PR.


Mytherin:
Thanks!

PR #23070, opened 2026-06-04
Generate transformer part 6

Dtenwolde:
Part of: https://github.com/duckdblabs/duckdb-internal/issues/9265

Now generating also: 
- `common.gram`
- `create_table.gram`

I changed the `generate_transformer_v2.py` (name change still pending) to be more recursive/composable, similar to the `Matcher`, to make certain grammar rules easier to support. 

Previously I ran a full `make format-fix` after the generator script, but I have added a `make format-parser-grammar` that formats only the relevant files, I hope that's okay. 

The two left are `select.gram` and `expression.gram` for which I will make a separate PR each I think, those are quite large (130 and 250 rules each respectively)


Mytherin:
Thanks!

Issue #15826, opened 2025-01-21
DuckDB ignores per-column key-value metadata of Parquet files


elveshoern32:
### What happens?

DuckDB ignores per-column key-value metadata of Parquet files

### To Reproduce

According to e.g. https://github.com/duckdb/duckdb/pull/9798 DuckDB shows in the result of parquet_metadata() the per-column key-value metadata.
Indeed there is a field for this in the output, but the value is always {}; the true metadata never show up.

Example:
This Python-3.11.4/PyArrow-14.0.1 script creates a tiny Parquet file with some per-column metadata:
```
import pyarrow as pa
import pyarrow.parquet as pq

schema = pa.schema([
    pa.field("col1", pa.int8(),    metadata={'keycol1':'valuecol1'}),
    pa.field("col2", pa.string(),  metadata={'keycol2':'valuecol2','foo':'bar'}),
    pa.field("col3", pa.float64(), metadata={'keycol3':'valuecol3','thud':'blat','yin':'yang'})
])

table = pa.table([
    pa.array([1, 2, 3, 4, 5]),
    pa.array(["a", "b", "c", "d", "e"]),
    pa.array([1.0, 2.0, 3.0, 4.0, 5.0])
  ], schema=schema
)

pq.write_table(table, "example.parquet")
```
This Python/PyArrow script proves that the metadata are really written to the file:
```
import pyarrow.parquet as pq

table = pq.read_table("example.parquet")

for f in ('col1','col2','col3'):
  print(table.schema.field(f).metadata)
```
The output is:
```
{b'keycol1': b'valuecol1'}
{b'keycol2': b'valuecol2', b'foo': b'bar'}
{b'keycol3': b'valuecol3', b'thud': b'blat', b'yin': b'yang'}
```

When examining the file with DuckDB/parquet_metadata(), however, ...
```
duckdb -c 'SELECT * FROM parquet_metadata('example.parquet')'
```
... the key-value metadata are not shown (please scroll to the right):
```
┌─────────────────┬──────────────┬────────────────────┬──────────────────────┬───┬──────────────────────┬──────────────────────┬────────────────────┐
│    file_name    │ row_group_id │ row_group_num_rows │ row_group_num_colu…  │ … │ total_compressed_s…  │ total_uncompressed…  │ key_value_metadata │
│     varchar     │    int64     │       int64        │        int64         │   │        int64         │        int64         │  map(blob, blob)   │
[...]

PR #23071, opened 2026-06-04
Write column-level key/value metadata to Parquet

bendrucker:
DuckDB can already write file-level Parquet key/value metadata and read column-level metadata back, but it has no way to *write* column-level metadata. This adds a `COLUMN_KV_METADATA` `COPY` option that attaches key/value pairs to individual columns. The pairs land on each leaf column chunk's `ColumnMetaData.key_value_metadata` (Thrift field 8), so they round-trip through the existing `parquet_metadata()` function with no reader changes.

## Usage

A `STRUCT` keyed by column name carries the metadata. A `MAP` value is the leaf payload. A `STRUCT` value nests into a complex column. Values may be `VARCHAR` or `BLOB`.

```sql
-- Flat column
COPY tbl TO 'f.parquet' (FORMAT parquet, COLUMN_KV_METADATA {
    id: MAP {'pii': 'false', 'source': 'crm'}
});

-- Complex columns: STRUCT nests by name, MAP is the leaf payload
COPY tbl TO 'f.parquet' (FORMAT parquet, COLUMN_KV_METADATA {
    addr:  {zip: MAP {'pii': 'true'}},                    -- struct field
    tags:  {element: MAP {'source': 'ingest'}},           -- list/array element
    attrs: {key:   MAP {'encoding': 'utf8'},              -- map key
            value: MAP {'encoding': 'utf8'}}              -- map value
});
```

Read it back with `parquet_metadata()`, addressing leaves by `path_in_schema`:

```sql
SELECT path_in_schema, key_value_metadata
FROM parquet_metadata('f.parquet')
WHERE cardinality(key_value_metadata) > 0;
-- addr, zip                  {pii=true}
-- tags, list, element        {source=ingest}
-- attrs, key_value, key      {encoding=utf8}
-- attrs, key_value, value    {encoding=utf8}
```

Column names match case-insensitively. The metadata is written onto every row group, so a leaf carries the same pairs in each of its column chunks.

## Design

I modeled the feature on the existing `field_ids` machinery, which already solves the same problem: a column-name-keyed, recursively nested `COPY` option threaded from bind through serialization. The leaf payload is a `vector<pair<string, string>>` instead of a single id, and it attaches to `ColumnMetaData` (per chunk, leaves only) rather than onto the column's `SchemaElement`.

## Validation

The bind step rejects specs that can't be honored, rather than writing surprising output:

- A leaf `MAP` on a complex column, or a nesting `STRUCT` on a primitive column.
- Values that aren't `VARCHAR` or `BLOB` (e.g. an integer or a nested value), which would otherwise be coerced to an internal debug string.
- `NULL` specifications, and `NULL` keys or values inside a leaf `MAP`.
- `VARIANT`, `UNION`, and `AGGREGATE_STATE` columns. They are nested (`PhysicalType::STRUCT`) but expose no addressable leaf, so the nested-type guards above reject them rather than silently dropping the metadata.

## Testing

I included tests for the supported leaf shapes (flat `VARCHAR`/`BLOB`, struct fields, list and array elements, map key/value), replication across row groups, case-insensitive matching, and the empty-`MAP` case.

### DIFF
--- a/extension/parquet/CMakeLists.txt
+++ b/extension/parquet/CMakeLists.txt
@@ -23,6 +23,7 @@ set(PARQUET_EXTENSION_FILES
     parquet_prefetch_cost_model.cpp
     parquet_reader.cpp
     parquet_field_id.cpp
+    parquet_column_kv.cpp
     parquet_statistics.cpp
     parquet_timestamp.cpp
     parquet_writer.cpp
--- a/extension/parquet/column_writer.cpp
+++ b/extension/parquet/column_writer.cpp
@@ -35,6 +35,7 @@
 #include "duckdb/common/uhugeint.hpp"
 #include "miniz.hpp"
 #include "parquet_field_id.hpp"
+#include "parquet_column_kv.hpp"
 #include "parquet_shredding.hpp"
 #include "parquet_timestamp.hpp"
 #include "parquet_types.h"
@@ -275,6 +276,7 @@ unique_ptr<ColumnWriter> ColumnWriter::CreateWriterRecursive(ClientContext &cont
                                                              const string &name, bool allow_geometry,
                                                              optional_ptr<const ChildFieldIDs> field_ids,
                                                              optional_ptr<const ShreddingType> shredding_types,
+                                                             optional_ptr<const ChildColumnKV> column_kv,
                                                              idx_t max_repeat, idx_t max_define, bool can_have_nulls) {
 	const bool parquet_write_timestamp_as_int96 = writer.WriteTimestampAsInt96();
 	path_in_schema.push_back(name);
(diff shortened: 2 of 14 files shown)


bendrucker:
I read https://github.com/duckdb/duckdb/issues/15826 before working on this and targeted spec compliance here in keeping with that discussion.

It seems like Arrow may not support writing this metadata?

https://github.com/apache/arrow/issues/31018

But that doesn't/shouldn't preclude DuckDB from implementing it.

I thought about some similar mechanic for putting metadata on the columns in DuckDB and then outputting that. But that would require arbitrary key/value tagging of columns which Duck doesn't expose, just `COMMENT`.

bendrucker:
Happy to come back to this anytime if there's interest!

PR #23073, opened 2026-06-04
fix(adbc): implement ADBC 1.1.0 Rich Error Metadata API

eitsupi:
This implements Rich Error Metadata API added in the ADBC ​​1.1.0 spec.

cc @lidavidm


Mytherin:
Thanks!

PR #9579, opened 2023-11-06
pin run-vcpkg action version

samansmink:
Solves issue where the run-vcpkg action would fail due to a glibc error. in v11.2 of https://github.com/lukka/run-vcpkg, Node was bumped to node20. 

between now and june 2024 when the manylinux_2014 image goes EOL, we should probably bump manylinux and update this action to the most recent version again

PR #23074, opened 2026-06-04
Internal #9579: ENUM IN empty

hawkfish:
* Optimize the edge case where none of the strings are in the enum.


src/optimizer/rule/in_clause_simplification_rule.cpp
```diff
@@ -125,6 +125,13 @@ unique_ptr<Expression> InEnumSimplificationRule::Apply(LogicalOperator &op, vect
 	//	then swap out the children for the valid ENUM values
 	if (in_children.size() > 1) {
 		children.swap(in_children);
+	} else if (expr.GetExpressionType() == ExpressionType::COMPARE_IN) {
+		//	IN () => false
+		return make_uniq<BoundConstantExpression>(Value::BOOLEAN(false));
```
Mytherin: Should this maybe be `CONSTANT_OR_NULL`? Can we add a test where the enum has `NULL` values and this is used in a `SELECT` list (so that the distinction between false / `NULL` matters)

Mytherin (commented):

src/optimizer/rule/in_clause_simplification_rule.cpp
hawkfish: I think we need to know if the probe has `NULL`s. So a more complex replacement is needed. Adding a row of all `NULL`s shows this.

hawkfish: `ISNULL(expr, NULL, false)`

Mytherin: That's what `CONSTANT_OR_NULL` does, no?

hawkfish: Not sure what that is?

```sh
(duckdb) hawkfish@Richards-MacBook-Pro-9 duckdb % git grep CONSTANT_OR_NULL
(duckdb) hawkfish@Richards-MacBook-Pro-9 duckdb % 
```

hawkfish: `FilterPropagateResult::FILTER_FALSE_OR_NULL`?

Mytherin: It's a function (`constant_or_null`) that returns either a constant or a NULL value - see `ExpressionRewriter::ConstantOrNull`. It does effectively what you're proposing (and the optimizer knows how to handle it in subsequent steps).

Mytherin:
Thanks!

PR #23075, opened 2026-06-04
regression test for issue 21931

josephbuckingham:
single unit test to assert expected behaviour in relation to #21931. Fails on 1.5.3, but passes currently.


Mytherin:
Thanks - but I think what we want to do with this issue is fix it in `v1.5-variegata` as well so that we can fix the regression in a bug-fix release.

PR #23079, opened 2026-06-05
Improve test output and highlight functions in stack trace

smvv:
### Misc

- Increase batch size from 10 to 30 to speed up test runs.
  - Fewer test processes needed, so reducing startup overhead.
- Omit redundant `FAILURES SUMMARY` block shown in a failed test output.

### Highlight function names and trim runner frames

Remove Catch2 function frames from the stack trace (less noise).
Highlight function names with teal so they are easier to spot/scan.

Screenshot 2026-06-05 at 09 24 59


Mytherin:
Thanks!

PR #23082, opened 2026-06-05
Merge v1.4-andium into v1.5-variegata, and add storage versions v1.4.5 and v1.5.4

carlopi:


carlopi:
@artjomPlaunov: could you possibly double check the diff at https://github.com/duckdb/duckdb/pull/23082/changes, and decide if `Cleanup()` should stay `noexcept` or not?
I tried to follow the history of the code, but unsure if I got it right.

Thanks!

Mytherin:
Thanks! I think it's fine to remove the `noexcept` there.

PR #23086, opened 2026-06-05
Fix strict-prototypes warnings

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/8943

This PR generalizes the C-API generation script with `create_parameter_list` and `create_argument_list` as a companion so that empty parameter lists now declare functions with `void` if they have no parameters.


Mytherin:
Thanks!

PR #23095, opened 2026-06-05
Fix Parquet thrift byte order on windows + relax geometry stats pruning

Maxxen:
This PR fixes an issue where the byte-order check used in the parquet thrift coded was incorrect on windows, leading to `double` values being encoded wrong. I think this was caused by pulling out all the boost library stuff (which I presume originally included the windows check) when the thrift code was originally vendored.

Since the only place where we serialize `double`s in thrift is in the parquet `GeometryStatistics` field, as far as I can tell the consequences of this bug has been basically non-existent until we added support for native parquet geometry around the end of last year. And even then, it's not a problem if you read files on windows that were produced by windows as the byte order is at least consistent, even if it's wrong. I only managed to find this while using my Mac to investigate stats-pruning with a parquet file produced by a user on windows. 

It's difficult to write a test for this, but I have been verifying manually by building duckdb on a windows machine and comparing the stats output of files written both on windows and on osx with and without this patch. 

Additionally this PR also relaxes the geometry stats pruning check slightly to allow intersection checks even if the bounds of only one of the x/y axes are defined.

### DIFF
--- a/src/include/duckdb/common/types/geometry.hpp
+++ b/src/include/duckdb/common/types/geometry.hpp
@@ -113,10 +113,27 @@ public:
 		return GeometryExtent {EMPTY_MIN, EMPTY_MIN, EMPTY_MIN, EMPTY_MIN, EMPTY_MAX, EMPTY_MAX, EMPTY_MAX, EMPTY_MAX};
 	}
 
-	// Does this extent have any X/Y values set?
-	// In other words, is the range of the x/y axes not empty and not unknown?
+	// Does this extent have the X axis set?
+	// In other words, is the range of the x-axis not empty and not unknown?
+	bool HasX() const {
+		return std::isfinite(x_min) && std::isfinite(x_max);
+	}
+	// Does this extent have the Y axis set?
+	// In other words, is the range of the y-axis not empty and not unknown?
+	bool HasY() const {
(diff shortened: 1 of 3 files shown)


Maxxen:
To reproduce the original issue is easy, just:

```sql
 copy (select 'POINT(0 1)'::GEOMETRY) to 'geo_file.parquet' (GEOPARQUET_VERSION V2)l
 ```
 
on windows/osx, and then on the other
 
 ```sql
 select geo_bbox from parquet_metadata('geo_file.parquet');
 ----
┌──────────────────────────────────────────────────────────────────────────────┐
│                                   geo_bbox                                   │
│ struct(xmin double, xmax double, ymin double, ymax double, zmin double, zmax │
│                       double, mmin double, mmax double)                      │
├──────────────────────────────────────────────────────────────────────────────┤
│ {                                                                            │
│   'xmin': 0.0,                                                               │
│   'xmax': 0.0,                                                               │
│   'ymin': 3.03865e-319,                                                      │
│   'ymax': 3.03865e-319,                                                      │
│   'zmin': null,                                                              │
│   'zmax': null,                                                              │
│   'mmin': null,                                                              │
│   'mmax': null                                                               │
│ }                                                                            │
└──────────────────────────────────────────────────────────────────────────────┘
 ```

Mytherin:
Thanks!

PR #23097, opened 2026-06-05
Add `decimal_average` aggregate function

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/8442

Adds `decimal_average(x)` as an aggregate function that computes the average of a `DECIMAL` column. Returns a `DECIMAL` with the same `width` and `scale`. 

Uses banker's rounding. 
Throws an `OutOfRangeException` if the intermediate result overflows.


lnkuiper:

Codex found this UB:
```sql
# The accumulated sum can be the minimum HUGEINT value; negating it in finalize overflows.
query I
SELECT decimal_average(x)
FROM (
	VALUES
		(-85070591730234615865843651857942052864::DECIMAL(38,0)),
		(-85070591730234615865843651857942052864::DECIMAL(38,0))
) t(x);
----
-85070591730234615865843651857942052864
```
Which should be fixed before we merge.

It also found this row-order issue, which is not a blocker, just interesting:
```sql
# decimal_average should not depend on row order for a multiset whose exact average fits DECIMAL(38,0).
query I
SELECT decimal_average(x)
FROM (
	VALUES
		(99999999999999999999999999999999999999::DECIMAL(38,0)),
		(99999999999999999999999999999999999999::DECIMAL(38,0)),
		(-99999999999999999999999999999999999999::DECIMAL(38,0))
) t(x);
----
33333333333333333333333333333333333333

query I
SELECT decimal_average(x)
FROM (
	VALUES
		(99999999999999999999999999999999999999::DECIMAL(38,0)),
		(-99999999999999999999999999999999999999::DECIMAL(38,0)),
		(99999999999999999999999999999999999999::DECIMAL(38,0))
) t(x);
----
33333333333333333333333333333333333333
```

Dtenwolde:
> It also found this row-order issue, which is not a blocker, just interesting:

Interesting find, but I am not sure this is easy to fix...Or workaround. 

I fixed the other issue, that was a good find.

lnkuiper:
No worries, that was not a blocker. Some `DOUBLE` operations are also order-dependent I think. LGTM

PR #23098, opened 2026-06-05
Step 1: Remove duplicate-eliminated joins and replace them with materialized CTEs

kryonix:
This is the first step of removing duplicate-eliminated joins (`DELIM_JOIN`) from the system by replacing them with materialized CTEs. `DELIM_JOIN`s are used in the system when we are dealing with correlation. The `DELIM_JOIN` operator computes two magic sets based on the correlation variables, once with duplicates and once without duplicates. Then, these sets are used to express the semantics of the correlated subquery. Details about this can be found in the papers "Unnesting Arbitrary Queries" and "Improving Unnesting of Complex Queries" by Thomas Neumann et. al.

However, there is nothing inherently special about `DELIM_JOIN`s, and we can express the same semantics using materialized CTEs. This is what this PR does. It replaces `DELIM_JOIN`s with materialized CTEs that compute the same magic sets. For now, this rewriting is behind a feature flag `delim_join_as_cte`, and I can not recommend enabling it, because of missing optimizations. I will address these in follow-up PRs. The main goal of this PR is to show that we can express the semantics of `DELIM_JOIN`s using materialized CTEs, and to lay the groundwork.

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -476,6 +476,13 @@
         "default_value": "ALL_ERRORS_INVALIDATE_TRANSACTION"
     },
     {
+        "name": "delim_join_as_cte",
+        "description": "Rewrite delim joins to materialized CTEs during dependent join flattening",
+        "type": "BOOLEAN",
+        "default_scope": "local",
+        "default_value": "false"
+    },
+    {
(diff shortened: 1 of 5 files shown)


Mytherin:
Awesome, thanks!

PR #23100, opened 2026-06-05
Add hardening to many DuckDB/Parquet decompression/deserializing paths

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9584

Summary of changes:

- `ALP` + `ALPRD` decompression had their `Load<...>` values checked for logical correctness, and certain `D_ASSERT(...)` upgraded to `InternalException` to also include release builds in the checks.
- Parquet's `VARIANT` binary decoding now uses offsets and checks for out-of-bounds reads.
- Parquet's RLE bit-unpack logic now checks for malicious/broken bitwidths.
- Patas decompression checks `Load<...>` values and verifies offsets are within bounds of the block.
- RLE decompression checks `Load<...>` values and verifies the `index_pointer` and `data_pointer` accesses are within bounds.


carlopi:
Could you possibly draft / undraft? Then CI should restart, thanks.

Mytherin (commented):
Thanks! I just think these shouldn't be InternalExceptions but IOExceptions - they come from invalid files, which should not invalidate the database.

Maybe we can also check in some of the broken DuckDB files (gzip compressed in `data/storage/...`)?

Mytherin:
Thanks!

src/storage/compression/rle.cpp
```diff
@@ -258,52 +258,71 @@ void RLEFinalizeCompress(CompressionState &state_p) {
...
+			//! This would make the index_pointer point into a region reserved for the header data
+			throw IOException("Corrupted RLE segment: rle_count_offset is corrupted");
+		}
+		if (segment.GetBlockOffset() + rle_count_offset > segment.GetBlockSize()) {
+			//! This would make the index_pointer start outside of the segment
+			throw IOException("Corrupted RLE segment: rle_count_offset is corrupted");
+		}
+		if ((rle_count_offset - RLEConstants::RLE_HEADER_SIZE) / sizeof(T) > max_entry_pos) {
+			//! This would make the indexing of the index_pointer[entry_pos] reach outside of the segment
+			throw IOException("Corrupted RLE segment: rle_count_offset is corrupted");
```
gishor, line 279: @Tishj We are debugging why we are getting such exceptions. Is it possible to add more details to the following three exception messages? Such that they aren't the same and it is clear from the message in which branch we are.

PR #23101, opened 2026-06-05
Step2: Push DELIM filters into CTE rewrite

kryonix:
Follow up of https://github.com/duckdb/duckdb/pull/23098.

This PR actually pushes DELIM filters to the right location, so that all rewritten `DELIM_SCAN`s (that are now `CTE_REF`s) read the already filtered data.

This PR also makes join bloom filters robust against underestimated build cardinality, by:
- Tracking the row count used when JoinHashTable::PrepareBuildBloomFilter initializes the bloom filter.
- Before publishing the bloom filter in hash-join finalize, compare actual ht.Count() with that initialization count.
- If actual count is more than 2x the initialization count, reset/reinitialize the bloom filter for ht.Count() and rebuild it from the finalized hash-table key data.

Finally, I have implemented `GetSinkProgress` for CTEs, so that the progress bar works for them.

### DIFF
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -821,13 +821,15 @@ void JoinHashTable::PrepareBuildBloomFilter(idx_t estimated_row_count) {
 	}
 }
 
-void JoinHashTable::PrepareBloomFilterForPushdown() {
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks!

PR #23102, opened 2026-06-05
validate variant value byte offsets when decoding

jmestwa-coder:
the variant decoder trusts the per-value byte offset from the variant data without bounds checking:
- a corrupt variant, for example read from an untrusted database file, with an out-of-range byte offset reads past the data blob
- DecodeStringData / DecodeNestedData / DecodeDecimalData and the VariantVisitor fixed-width path all index `data + byte_offset` directly
- the invariant was only a `D_ASSERT` in `VariantUtils::Verify`, so release builds had no check (a crafted file makes `SELECT v::VARCHAR` segfault or return hundreds of MB of out-of-bounds heap as a string)

enforce `byte_offset <= data size` at decode time and bound the decoded string length within the blob.


Tishj:
#23100 

Ah nvm, this is DuckDB Variant, not Parquet

jmestwa-coder:
right, this is the native variant decode path (`VariantUtils::Decode*` / `VariantVisitor`), separate from the parquet variant binary decoding in #23100, so they don't overlap. this just covers the `byte_offset` trust on the duckdb-native side. marked it ready for review.

jmestwa-coder:
pushed a clang-format fix for the format check failure, should be green now.

PR #23105, opened 2026-06-05
Fix ORDER BY true/false bypassing non-integer literal validation

### LINKED ISSUE #23076: ORDER BY true/false bypasses non-integer literal validation
### What happens?

`ORDER BY true` and `ORDER BY false` are silently accepted and generate a full sort pipeline (`TOP_N`/`ORDER_BY`) instead of raising a binder error when `order_by_non_integer_literal=false` (the default), inconsistent with how other non-integer constants like `'a'`, `NULL`, and `3.14` are handled.

### To Reproduce

**Execute these queries on DuckDB CLI**
```
create table products(id int, name varchar(10));
select * from products order by true;

--aggregate expression example
SELECT array_agg(x ORDER BY true) FROM (VALUES (1), (2)) t(x);
```

**Current Output**
```
┌───────┬─────────┐
│  id   │  name   │
│ int32 │ varchar │
└───────┴─────────┘
      0 rows

┌────────────────────────────────────────────┐
│ array_agg(x ORDER BY CAST('t' AS BOOLEAN)) │
│                  int32[]                   │
├────────────────────────────────────────────┤
│ [1, 2]                                     │
└────────────────────────────────────────────┘
```

**Expected Output**
Should raise a binder error consistent with other non-integer constants:
```
Binder Error: ORDER BY non-integer literal has no effect.
* SET order_by_non_integer_literal=true to allow this behavior.
```

**Examples where the error is correctly invoked for other non-integer literals**
```
select * from products order by 'a';
select * from products order by NULL;
select * from products order by 3.5;

--aggregate expression example
SELECT array_agg(x ORDER BY 'a') FROM (VALUES (1), (2)) t(x);
```

**Root Cause**
Boolean literals are parsed as `CastExpression(BOOLEAN, Constant('t'|'f'))` in `TransformConstantLiteral` ([`src/parser/peg/transformer/transform_expression.cpp`](https://github.com/duckdb/duckdb/blob/0ce48ae355b2358414808c95216d8b79f8ee111d/src/parser/peg/transformer/transform_expression.cpp#L1545)), while all other non-integer constants (`'a'`, `NULL`, `3.14`) produce `ConstantExpression`.

The validation in `OrderBinder::TryGetProjectionReference` ([`src/planner/expression_binder/order_binder.cpp`](https://github.com/duckdb/duckdb/blob/main/src/planner/expression_binder/order_binder.cpp#L56)) only happens for `ExpressionClass::CONSTANT`, so boolean literals bypass it entirely and fall through to the general binding path. The same gap exists in bind_aggregate_expression.cpp ([`src/planner/binder/expression/bind_aggregate_expression.cpp`](https://github.com/duckdb/duckdb/blob/main/src/planner/binder/expression/bind_aggregate_expression.cpp#L171)) for aggregate ORDER BY clauses.

**Note**
PostgreSQL, which DuckDB's ORDER BY constant handling is based on, rejects `ORDER BY true` with `ERROR: non-integer constant in ORDER BY`. DuckDB is inconsistent with that behavior.

### OS:

macOS

### DuckDB Version:

v1.5.3 (Variegata)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Ajay Panthagani

### Affiliation:

ninja-van

- [x] Yes, I have

- [x] Yes, I have

ajaypanthagani:
Fixes [#23076](https://github.com/duckdb/duckdb/issues/23076)

`ORDER BY true` and `ORDER BY false` were silently accepted and bound as real sort keys. Other non-integer literals (`'a'`, `NULL`, `3.14`) already followed the expected path, they either raised an error (default behavior) or were ignored when `order_by_non_integer_literal=true`.

### Root cause

`TransformConstantLiteral` emitted `TRUE`/`FALSE` as `CastExpression(BOOLEAN, ConstantExpression('t'|'f'))` instead of plain `ConstantExpression` nodes.

`OrderBinder::TryGetProjectionReference()` and `bind_aggregate_expression.cpp` both only recognize plain constant expressions (`ExpressionClass::CONSTANT` / `ExpressionType::VALUE_CONSTANT`). Because boolean literals were wrapped in `CastExpression`, they bypassed both checks and fell through to the general binding path, where `CreateExtraReference()` appended them to the SELECT list as real sort keys.

### Solution

Fixed the root cause in the parser.

`TransformConstantLiteral` now produces a plain `ConstantExpression(val)` for booleans also.

No binder changes are needed, the existing non-integral constant checks in `TryGetProjectionReference()` and `bind_aggregate_expression.cpp` already handle `ConstantExpression(Value(true/false))` correctly.

### Tests

```bash
build/reldebug/test/unittest test/sql/order/test_order_by_non_ordinal_literal.test
build/reldebug/test/unittest test/sql/order/test_order_by_exceptions.test
build/reldebug/test/unittest test/sql/binder/order_by_view.test
build/reldebug/test/unittest test/sql/pivot/pivot_star.test

### DIFF
--- a/src/parser/peg/transformer/transform_expression.cpp
+++ b/src/parser/peg/transformer/transform_expression.cpp
@@ -1538,12 +1538,7 @@ unique_ptr<ParsedExpression> PEGTransformerFactory::TransformConstantLiteral(PEG
                                                                              ParseResult &parse_result) {
 	auto &list_pr = parse_result.Cast<ListParseResult>();
 	auto val = transformer.Transform<Value>(list_pr.Child<ChoiceParseResult>(0).GetResult());
-	if (val.IsNull()) {
-		return make_uniq<ConstantExpression>(val);
-	} else {
-		auto varchar_val = make_uniq<ConstantExpression>(val.GetValue<string>().substr(0, 1));
-		return make_uniq<CastExpression>(LogicalType::BOOLEAN, std::move(varchar_val));
-	}
+	return make_uniq<ConstantExpression>(val);
 }
 
 Value PEGTransformerFactory::TransformFalseLiteral(PEGTransformer &transformer, ParseResult &parse_result) {
--- a/test/sql/binder/order_by_view.test
+++ b/test/sql/binder/order_by_view.test
@@ -29,12 +29,14 @@ CreatedAt >= '2023-12-01' AND
 CreatedAt < '2023-12-13' 
 GROUP BY HasCustomAddress ;
 
-statement ok
-SELECT HasCustomAddress, count(*) AS total_records 
-FROM model 
-WHERE  1 = 1 AND 
+statement error
+SELECT HasCustomAddress, count(*) AS total_records
+FROM model
+WHERE  1 = 1 AND
 CreatedAt >= '2023-12-01' AND
-CreatedAt < '2023-12-13' 
+CreatedAt < '2023-12-13'
 GROUP BY HasCustomAddress
-ORDER BY true, total_records DESC NULLS LAST 
+ORDER BY true, total_records DESC NULLS LAST
 LIMIT 250 OFFSET 0;
+----
+ORDER BY non-integer literal has no effect
--- a/test/sql/order/test_order_by_non_ordinal_literal.test
+++ b/test/sql/order/test_order_by_non_ordinal_literal.test
@@ -1,0 +1,125 @@
+# name: test/sql/order/test_order_by_non_ordinal_literal.test
+# description: Non-ordinal literal validation (ORDER BY, DISTINCT ON, UNION)
+# group: [order]
+
+statement ok
+CREATE TABLE t (a INTEGER, b INTEGER);
+
+statement ok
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by Mytherin** [comment]
Thanks - but this is introducing a lot of special case code that is introducing more complexity than this fix is worth. Maybe a better approach is to change the parser so that `true / false` emit just a single constant instead of emitting a cast?

**Comment by ajaypanthagani** [comment]
Sure! Will update the PR.

**Comment by ajaypanthagani** [comment]
@Mytherin I pushed the parser side changes

**Review by Dtenwolde — APPROVED** [review]
Thanks! I think this looks good. I did this because the old parser would emit `CAST('t' as BOOLEAN` for `true`, which I always found odd, but perhaps there's a good reason PostgreSQL does it this way: 
```sql
memory D select true;
┌──────────────────────┐
│ CAST('t' AS BOOLEAN) │
│       boolean        │
├──────────────────────┤
│ true                 │
└──────────────────────┘
```

Anyhow, I think it's good to emit just a `ConstantExpression` here :)

Mytherin:
Thanks!

PR #23110, opened 2026-06-06
fix(adbc): fill metadata of GetObjects

eitsupi:
Currently, much of the metadata is empty.
It can be added by referring to the JDBC driver.

cc @lidavidm @kentkwu


lidavidm (approved):

Mytherin:
Thanks!

Issue #23114, opened 2026-06-08
json_object and json_group_object silently drop the NULL-key entry


DerZc:
### What happens?

Hi,

I found that the json_object and json_group_object silently drop the NULL-key entry without any warning or message; this might cause data loss silently.

This is an example:

```
SELECT json_object(CAST(NULL AS VARCHAR), 'two');
┌───────────────────────────────────────────┐
│ json_object(CAST(NULL AS VARCHAR), 'two') │
│                   json                    │
├───────────────────────────────────────────┤
│ {}                                        │
└───────────────────────────────────────────┘

SELECT json_group_object(k, v) FROM (VALUES ('ok', 'one'), (NULL, 'two')) t(k, v);
┌─────────────────────────┐
│ json_group_object(k, v) │
│          json           │
├─────────────────────────┤
│ {"ok":"one"}            │
└─────────────────────────┘
```

### To Reproduce

```
SELECT json_object(CAST(NULL AS VARCHAR), 'two');
SELECT json_group_object(k, v) FROM (VALUES ('ok', 'one'), (NULL, 'two')) t(k, v);
```

### OS:

Ubuntu 24.04

### DuckDB Version:

4dac4343d654d967afb9bc057c3b3638158cd00b

### DuckDB Client:

CLI

### Full Name:

Chi Zhang

### Affiliation:

Tsinghua University

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

PR #23116, opened 2026-06-08
Reject NULL json key

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23114

In the current implementation, when NULL used as json object key, it's ignored directly without error.
I think a better approach is directly reject, same as postgres behavior: https://onecompiler.com/postgresql/44rnf6uxd


Mytherin:
Thanks!

PR #23119, opened 2026-06-08
Leonid/before triggers

krleonid:
## Summary

- Extends the existing AFTER trigger DML-CTE expansion to also fire BEFORE INSERT/UPDATE/DELETE triggers (FOR EACH STATEMENT). `TryExpandAfterTriggers` is renamed to `TryExpandTriggers` and unified to handle both timings in a single pass — it fetches the BEFORE and AFTER lists, builds a single combined CTE plan, and reuses the same recursion guard.
- Ordering is enforced via `LogicalMaterializedCTE` execution order: BEFORE body CTEs are registered first in the `cte_map`, then the base (DML) CTE, then AFTER body CTEs. Because `cte_map.map` is an `InsertionOrderPreservingMap` and `bind_cte_node.cpp` wraps CTEs into nested `LogicalMaterializedCTE`s in registration order, this gives BEFORE → DML → AFTER at runtime.
- Scope matches the existing AFTER feature: FOR EACH STATEMENT only. FOR EACH ROW and REFERENCING (transition tables) for BEFORE remain rejected at `CREATE TRIGGER` time. `RETURNING` is rejected on tables with any trigger; `ON CONFLICT DO UPDATE` with `REFERENCING NEW TABLE` triggers remain rejected (only AFTER can carry REFERENCING, so the check is unchanged in spirit).

## Implementation notes

- `src/planner/binder.cpp` — `TryExpandTriggers` / `ExpandTriggers` (replace `*AfterTriggers`). BEFORE bodies use a distinct CTE name prefix (`__duckdb_trigger_before_body_`) so they're easy to spot in `EXPLAIN`. BEFORE bodies are not given transition-table aliases (REFERENCING is rejected for BEFORE at CREATE time).
- `src/planner/binder/statement/bind_{insert, update, delete}.cpp` — call-site rename (3 lines).
- `src/planner/binder/statement/bind_create.cpp` — **untouched** by this PR. CREATE TRIGGER validation was already timing-agnostic.
- `test/sql/trigger/trigger_after_insert.test` — one error message updated (`"AFTER triggers"` → `"triggers"`) since `RETURNING` is now rejected for tables with any trigger, not just AFTER.

## Test plan

- [x] New: `test/sql/trigger/trigger_before_insert.test` — basic firing, multi-row INSERT, FOR EACH ROW rejection, REFERENCING rejection, error rollback, multiple BEFORE triggers, DML-body variants, RETURNING rejection, recursion guard, DROP TRIGGER, **ordering proof** (BEFORE body observes pre-INSERT row count).
- [x] New: `test/sql/trigger/trigger_before_update.test` — same coverage for UPDATE, including UPDATE OF columns and ordering proof (BEFORE body observes old `val`).
- [x] New: `test/sql/trigger/trigger_before_delete.test` — same coverage for DELETE, ordering proof (BEFORE body sees row still present).
- [x] New: `test/sql/trigger/trigger_before_after_mixed.test` — strongest ordering test: same table has both BEFORE and AFTER triggers; BEFORE records pre-DML state, AFTER records post-DML state. Also: failure in BEFORE aborts both the DML and the AFTER trigger.
- [x] All 17 trigger test cases pass (871 assertions). Adjacent CTE (2915 assertions) and INSERT (433 assertions) suites are unaffected.

### DIFF
--- a/src/planner/binder.cpp
+++ b/src/planner/binder.cpp
@@ -600,6 +600,29 @@ unique_ptr<BoundStatement> Binder::TryExpandTriggers(QueryNode &node,
 	auto txn = table.ParentCatalog().GetCatalogTransaction(context);
 	auto before_triggers = table.GetTriggersForEvent(txn, TriggerTiming::BEFORE, event_type);
 	auto after_triggers = table.GetTriggersForEvent(txn, TriggerTiming::AFTER, event_type);
+
+	// UPDATE OF <cols>: drop triggers whose OF list is disjoint from the SET list.
+	// Triggers without an OF list are unrestricted and always fire.
+	if (event_type == TriggerEventType::UPDATE_EVENT && node.type == QueryNodeType::UPDATE_QUERY_NODE) {
+		auto &update_node = node.Cast<UpdateQueryNode>();
+		case_insensitive_set_t updated_columns;
+		if (update_node.set_info) {
+			updated_columns.insert(update_node.set_info->columns.begin(),
+			                       update_node.set_info->columns.end());
+		}
+		auto trigger_does_not_fire = [&](const_reference<TriggerCatalogEntry> trig) {
+			const auto &of_cols = trig.get().columns;
+			return !of_cols.empty() &&
+			       std::none_of(of_cols.begin(), of_cols.end(),
+			                    [&](const string &c) { return updated_columns.count(c) > 0; });
+		};
+		auto drop_non_firing = [&](vector<const_reference<TriggerCatalogEntry>> &triggers) {
+			triggers.erase(std::remove_if(triggers.begin(), triggers.end(), trigger_does_not_fire), triggers.end());
+		};
+		drop_non_firing(before_triggers);
+		drop_non_firing(after_triggers);
+	}
+
 	if (before_triggers.empty() && after_triggers.empty()) {
 		return nullptr;
 	}
@@ -661,16 +684,9 @@ BoundStatement Binder::ExpandTriggers(QueryNode &node, vector<unique_ptr<ParsedE
 	from_ref->table_name = base_cte_name;
 	outer->from_table = std::move(from_ref);
 
-	// CTE registration order = materialized-CTE execution order (first registered is outermost
-	// LogicalMaterializedCTE, runs first). So:
-	//   1. BEFORE body CTEs first  -> run before the DML
-	//   2. Base CTE (the DML)      -> runs in the middle
-	//   3. AFTER  body CTEs last   -> run after the DML
-	//
-	// Within each timing the catalog returns triggers alphabetically (case-insensitive) — see
-	// GetTriggersForEvent.
+	// CTE definition order == execution order: BEFORE bodies, then base (DML), then AFTER bodies.
 
-	// 1. BEFORE bodies (no transition tables — REFERENCING is rejected at CREATE TRIGGER time)
+	// BEFORE bodies (no transition tables — REFERENCING is rejected at CREATE TRIGGER time)
 	for (idx_t i = 0; i < before_triggers.size(); i++) {
 		auto &trigger = before_triggers[i].get();
 		auto body_cte_name = string(TRIGGER_BEFORE_BODY_CTE_PREFIX) + to_string(i + 1) + "_" + uuid_suffix;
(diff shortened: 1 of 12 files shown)


src/planner/binder.cpp
```diff
@@ -640,33 +643,56 @@ static unique_ptr<CommonTableExpressionInfo> MakeTransitionTableAliasCTE(const s
...
-	base_cte->is_trigger_generated = true;
-
 	// count(*) over the base CTE gives CHANGED_ROWS ("N rows affected") to the client
 	auto outer = make_uniq<SelectNode>();
 	outer->select_list.push_back(make_uniq<FunctionExpression>("count_star", vector<unique_ptr<ParsedExpression>>()));
 	auto from_ref = make_uniq<BaseTableRef>();
 	from_ref->table_name = base_cte_name;
 	outer->from_table = std::move(from_ref);
+
+	// CTE registration order = materialized-CTE execution order (first registered is outermost
```
EtgarDev: I would tune down a bit the verbosity of the comments
Something like "CTE definition order == execution order" would be pretty enough IMO

lnkuiper:
Thanks for the PR! Looks great. Making really solid progress on triggers before the next release!

Codex has found some cases where a trigger fires which it thinks it should not (cross-referenced with SQLite and PostgreSQL):
```sql
# name: test/temp.test
# description: Temporary breaking trigger tests for PR review
# group: [trigger]

require noforcestorage

# BEFORE UPDATE OF should only fire when one of the listed columns is updated.
statement ok
CREATE TABLE before_update_of_target (id INTEGER, val INTEGER, other INTEGER);

statement ok
INSERT INTO before_update_of_target VALUES (1, 10, 100);

statement ok
CREATE TABLE before_update_of_log (msg VARCHAR);

statement ok
CREATE TRIGGER before_update_of_trg
BEFORE UPDATE OF val ON before_update_of_target
FOR EACH STATEMENT
    INSERT INTO before_update_of_log VALUES ('before val updated');

statement ok
UPDATE before_update_of_target SET other = 200 WHERE id = 1;

query I
SELECT COUNT(*) FROM before_update_of_log;
----
0

statement ok
UPDATE before_update_of_target SET val = 20 WHERE id = 1;

query I
SELECT COUNT(*) FROM before_update_of_log;
----
1

# AFTER UPDATE OF should use the same column filtering.
statement ok
CREATE TABLE after_update_of_target (id INTEGER, val INTEGER, other INTEGER);

statement ok
INSERT INTO after_update_of_target VALUES (1, 10, 100);

statement ok
CREATE TABLE after_update_of_log (msg VARCHAR);

statement ok
CREATE TRIGGER after_update_of_trg
AFTER UPDATE OF val ON after_update_of_target
FOR EACH STATEMENT
    INSERT INTO after_update_of_log VALUES ('after val updated');

statement ok
UPDATE after_update_of_target SET other = 200 WHERE id = 1;

query I
SELECT COUNT(*) FROM after_update_of_log;
----
0

statement ok
UPDATE after_update_of_target SET val = 20 WHERE id = 1;

query I
SELECT COUNT(*) FROM after_update_of_log;
----
1
```
These cases should be fixed before we can merge.

test/sql/trigger/trigger_before_after_mixed.test
```diff
@@ -0,0 +1,157 @@
+# name: test/sql/trigger/trigger_before_after_mixed.test
```
EtgarDev, line 1: I think it would be nice to add a test case for when we have cross-timing cycle:

```sql
# Cross-timing cycle: BEFORE on A -> inserts B, AFTER on B -> inserts A
# The second CREATE TRIGGER should be blocked because it would form a cycle
statement ok
CREATE TABLE cyc_ba_a (id INTEGER);

statement ok
CREATE TABLE cyc_ba_b (id INTEGER);

statement ok
CREATE TRIGGER cyc_ba_before BEFORE INSERT ON cyc_ba_a FOR EACH STATEMENT
    INSERT INTO cyc_ba_b VALUES (1);

statement error
CREATE TRIGGER cyc_ba_after AFTER INSERT ON cyc_ba_b FOR EACH STATEMENT
    INSERT INTO cyc_ba_a VALUES (1);
----
<REGEX>:.*[Rr]ecurs.*

# Cross-timing cycle: AFTER on A -> inserts B, BEFORE on B -> inserts A
# The second CREATE TRIGGER should be blocked
statement ok
CREATE TABLE cyc_ab_a (id INTEGER);

statement ok
CREATE TABLE cyc_ab_b (id INTEGER);

statement ok
CREATE TRIGGER cyc_ab_after AFTER INSERT ON cyc_ab_a FOR EACH STATEMENT
    INSERT INTO cyc_ab_b VALUES (1);

statement error
CREATE TRIGGER cyc_ab_before BEFORE INSERT ON cyc_ab_b FOR EACH STATEMENT
    INSERT INTO cyc_ab_a VALUES (1);
----
<REGEX>:.*[Rr]ecurs.*
```

EtgarDev (commented):
I suggested a test case :)

lnkuiper:
Thanks for the changes, feel free to run CI when you think the PR is ready :)

PR #23127, opened 2026-06-08
Clarify BIGNUM C API data is big endian

mlafeldt:
And actually verify byte order in tests, which only round-tripped raw bytes and so never caught that the max/min double fixtures were themselves wrong.


Mytherin:
Thanks!

PR #23128, opened 2026-06-08
Fix correlated joinside regression

kryonix:
Some changes in the binder introduced a regression where some queries would fail with `Non-inner join on correlated columns not supported`, while in reality, we could execute those queries safely. This PR fixes that.


Mytherin:
Thanks!

PR #23133, opened 2026-06-08
Use min/max as a proxy for distinct count if estimates are absent

lnkuiper:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9573

Unlike DuckDB base tables, which have HyperLogLog estimates, we don't have distinct counts for Parquet/DuckLake/Iceberg/Delta. We already try to [infer distinct counts from the join graph using base table cardinalities](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://blobs.duckdb.org/papers/tom-ebergen-msc-thesis-join-order-optimization-with-almost-no-statistics.pdf&ved=2ahUKEwj059WkuPeUAxX8-gIHHXsBKoQQFnoECA0QAQ&usg=AOvVaw2EBGOgFh_PKtmiIEEeAk7B), which works quite well. However, we don't use this method for filter selectivity estimation, which defaults to 20%. This PR adds a distinct count estimate using min/max (integral columns only) that is used when a more exact distinct count is not available. This can greatly improve filter selectivity estimates, which in turn can improve join ordering.

CC @guillesd


src/optimizer/join_order/relation_statistics_helper.cpp
```diff
@@ -28,6 +32,123 @@ bool ExpressionBinding::FoundColumnRef() const {
...
+
+static idx_t GetMinMaxDistinctCount(const BaseStatistics &base_stats, idx_t base_table_cardinality) {
+	if (base_table_cardinality == 0 || base_stats.GetStatsType() != StatisticsType::NUMERIC_STATS ||
+	    !NumericStats::HasMinMax(base_stats)) {
+		return 0;
+	}
+
+	switch (base_stats.GetType().id()) {
+	case LogicalTypeId::BOOLEAN:
+		return GetBooleanMinMaxDistinctCount(base_stats, base_table_cardinality);
```
Mytherin, line 103: Can we do this on physical type instead? This seems to unnecessarily exclude timestamps.

lnkuiper: Great idea, done

PR #23134, opened 2026-06-08
Rework aggregate states - make aggregate states use extension type info, and fully deprecate legacy aggregate states

Mytherin:
Currently we have two aggregate states:

* Legacy aggregate state - a blob type that holds an unstructured aggregate state (directly exported bytes from the aggregate hash table)
* Aggregate state - a struct type that holds a structured aggregate state

This setup requires a lot of code in various code paths to handle this as a top-level type. It's also limiting because aggregate states now **must** be struct types, when we might not want them / need them to be struct types in all situations. For example, for `sum`, a single number would be sufficient as aggregate state. No need to wrap it in a struct in theory.

This PR reworks aggregate states to instead be only an annotation to an existing type. We set the type alias to `AGGREGATE_STATE`, and set the necessary information for using the aggregate state in the `ExtensionTypeInfo`. This is  `function_name VARCHAR, parameters TYPE[]` (effectively the aggregate function signature).

We also rework the way that the export state function is called. Instead of creating a separate function (`aggregate_state_export_[aggregate_name]`) which has a separate bind data, we only overload the `Finalize` method and the return type of the aggregate function. That is because aggregate state export is effectively "only" a different finalize method where we export the state instead of computing the aggregate result. Doing it in this way means we can preserve any original properties of the function, including e.g. any necessary `FunctionData`, etc, which allows this to work much more cleanly with all aggregate functions (which is the ultimate goal).

```sql
select sum(42) export_state;
│     sum(42) EXPORT_STATE     │
│       aggregate_state        │
│ {'isset': true, 'value': 42} │
```

### `to_aggregate_state`

The `aggregate_state` is just a decorator over any existing type - any type can be an aggregate state (depending on what the aggregate function defines as its internal aggregate state). This also means we can write aggregate states to other storage destinations and not just to DuckDB files. For example, we can write them to Parquet:

```sql
copy (select sum(42) export_state as state) to 'file.parquet';
from file.parquet;
┌───────────────────────────────────────┐
│                 state                 │
│ struct(isset boolean, "value" double) │
├───────────────────────────────────────┤
│ {'isset': true, 'value': 42.0}        │
└───────────────────────────────────────┘
```

However, this round-trip loses the aggregate state type as we cannot write that to Parquet. In order to recover that type, we can use the `to_aggregate_state` function. This function takes the aggregate state read from the file and the function signature of the aggregate function and converts it back into an aggregate state object:

```sql
select to_aggregate_state(state, 'sum', ['INTEGER']) as state from file.parquet;
┌────────────────────────────────┐
│             state              │
│        aggregate_state         │
├────────────────────────────────┤
│  {'isset': true, 'value': 42}  │
└────────────────────────────────┘
```

We can then operate on the aggregate state with `combine` and `finalize` again, e.g.:

```sql
select finalize(to_aggregate_state(state, 'sum', ['INTEGER'])) as result from file.parquet; 
┌────────┐
│ result │
│ int128 │
├────────┤
│     42 │
└────────┘
```

### Legacy Aggregate States Deprecation

Legacy aggregate states are fully deprecated. However, if we encounter them in a file, we can still read them - we just read them as blobs. The original legacy aggregate state information is discarded. The reasoning for deprecating this type fully and removing it immediately is that the legacy aggregate states were very much not fully implemented and undocumented. They also tie directly to aggregate state internals, so trying to maintain them is painful and error-prone. Given that (almost) nobody likely used them and the feature was undocumented just removing them seems like the best way forward.


Maxxen:
Not too happy about reviving ExtensionTypeInfo as that was planned for removal anyway - but we can revisit it later. 

>  It's also limiting because aggregate states now must be struct types, when we might not want them / need them to be struct types in all situations. For example, for sum, a single number would be sufficient as aggregate state. No need to wrap it in a struct in theory.

So why not `VARIANT`? I can maybe guess why, but I've brought it up before that I think we should have a well-defined (schema-stored-separately) row-wise binary serialization that could be used both in the hash-table/tuple data format, sort-key optimizations, non-shredded variant, as well as the (now reintroduced) aggregate states.

Mytherin:
Because they aren't `VARIANT`, they are well-defined structured types. The internal structure just depends on the function signature itself. Making them `VARIANT` just adds complication. It doesn't remove any of the complication. The complication behind aggregate states isn't that they are arbitrary types, it's that we need to tie them to a specific function signature.

The alternative approach I considered is to not use the type system at all, but then you need to pass in the function signature into the `combine` and `finalize` functions always, which makes the SQL more ugly.

rustyconover:
As somebody who's likely been one of the few who has used this, this seems pretty cool!  It could really simplify my data sketches extension.

I know we used `INTEGER` in the examples here, do we have a support for parameterized types like DECIMAL(18,3) or TIMESTAMP[ns] what about structs and other nested types?

Mytherin:
Yes, that's a `TYPE` type which can represent any type.

PR #23135, opened 2026-06-08
Extend `RemotePushdownOptimizer` to be more flexible w.r.t. unsupported remote catalog operations, and integrate with MySQL

Mytherin:
Follow-up to https://github.com/duckdb/duckdb/pull/22914

In the initial implementation the `RemotePushdownOptimizer` only considered the locality of (sub)-queries, and would always push execution to the remote catalog if they were done on tables that were part of that remote catalog. While this works fine if the local and remote are the same, it does not work if they have different semantics, as is the case when we are dealing with e.g. a different database system as the remote catalog.

This PR extends the `RemotePushdownOptimizer` so that it instead can query the catalog for whether or not certain constructs are supported. This is done through the following overloads in the `Catalog`:

```cpp
virtual bool SupportsPushdown(const ParsedExpression &expression);
virtual bool SupportsPushdown(const TableRef &ref);
virtual bool SupportsPushdown(const QueryNode &node);
```

The catalog can then opt to support pushdown more granularly, and e.g. reject pushdown of functions, aggregates or query constructs that it does not support.

As a proof-of-concept, we also integrate this into the MySQL catalog. In the patch to the MySQL catalog we check the constructs that are present in the query and only push those that match DuckDB's semantics. We also do a rewrite - instead of blindly calling `QueryNode::ToString` we implement a separate ToString routine that takes MySQL's query dialect into account (and uses e.g. MySQL's quoting rules and remaps functions that have different names).

The `SupportsPushdown` methods are implemented in MySQL to deal with what MySQL supports, and what it does not. Here's an example of the `SupportsPushdown` implementation for `QueryNode`:

```cpp
bool MySQLCatalog::SupportsPushdown(const QueryNode &node) {
	if (!node.cte_map.map.empty() && !version.SupportsCTEs()) {
		// common table expressions require MySQL 8.0 / MariaDB 10.2
		return false;
	}
	if (!MySQLSupportsCTEMap(node.cte_map)) {
		return false;
	}
	switch (node.type) {
	case QueryNodeType::SELECT_NODE: {
		auto &select = node.Cast<SelectNode>();
		if (select.qualify) {
			// MySQL cannot execute QUALIFY
			return false;
		}
		if (select.aggregate_handling == AggregateHandling::FORCE_AGGREGATES) {
			// GROUP BY ALL is DuckDB-specific and would be serialized verbatim
			return false;
		}
		if (select.groups.grouping_sets.size() > 1) {
			// ROLLUP / CUBE / GROUPING SETS are serialized as GROUP BY GROUPING SETS,
			// which MySQL only supports with a secondary engine (HeatWave)
			return false;
		}
		if (select.sample) {
			// MySQL has no USING SAMPLE
			return false;
		}
		if (select.having && select.groups.grouping_sets.empty()) {
			// HAVING without GROUP BY: MySQL resolves bare column references against the select
			// list and filters rows, while DuckDB turns the query into an aggregation and raises
			// a binder error for non-aggregated columns
			return false;
		}
		if (!MySQLSupportsSelectAliasUsage(select)) {
			// the query references select-list aliases at positions MySQL cannot resolve
			return false;
		}
		if (select.from_table && MySQLJoinTreeHasCoalescingJoin(*select.from_table)) {
			// NATURAL / USING joins coalesce the join columns when expanding an unqualified *,
			// and MySQL orders the result columns differently than DuckDB
			for (auto &expr : select.select_list) {
				if (expr->GetExpressionClass() == ExpressionClass::STAR &&
				    expr->Cast<StarExpression>().RelationName().empty()) {
					return false;
				}
			}
		}
		return MySQLSupportsResultModifiers(node, &select.select_list);
	}
	case QueryNodeType::SET_OPERATION_NODE: {
		auto &setop = node.Cast<SetOperationNode>();
		if (setop.setop_type == SetOperationType::UNION_BY_NAME) {
			// UNION BY NAME is DuckDB-specific
			return false;
		}
		if (setop.setop_type == SetOperationType::EXCEPT || setop.setop_type == SetOperationType::INTERSECT) {
			// EXCEPT / INTERSECT require MySQL 8.0.31 / MariaDB 10.3 (ALL variants: MariaDB 10.5)
			if (!version.SupportsExceptIntersect(setop.setop_all)) {
				return false;
			}
		}
		// there is no select list to resolve positional ORDER BY references against
		return MySQLSupportsResultModifiers(node, nullptr);
	}
	case QueryNodeType::RECURSIVE_CTE_NODE: {
		auto &cte = node.Cast<RecursiveCTENode>();
		if (!version.SupportsCTEs()) {
			// recursive CTEs require MySQL 8.0 / MariaDB 10.2
			return false;
		}
		if (!cte.key_targets.empty()) {
			// USING KEY is DuckDB-specific
			return false;
		}
		return MySQLSupportsResultModifiers(node, nullptr);
	}
	case QueryNodeType::INSERT_QUERY_NODE:
	case QueryNodeType::DELETE_QUERY_NODE:
	case QueryNodeType::UPDATE_QUERY_NODE:
		// DML cannot be executed through mysql_query - these statements are handled by the
		// regular catalog insert/update/delete paths instead (the SELECT part of an INSERT
		// can still be pushed down separately)
		return false;
	default:
		// unknown query node type
		return false;
	}
}
```

CC @staticlibs


staticlibs:
This change makes the query generation in MySQL way more straightforward! One side effect of it is that it seems that it is not leaving much common logic to share between MySQL and Postgres. As all query transformation logic is already common in the core and checks in the catalog are DB-specific. Perhaps `MySQLSQLWriter` can be generalized.

Going to add the following to MySQL/Postgres:

 - `UDPDATE`/`DELETE` support in query writer and `*_query()` functions
 - I do not think the change in `mysql_query()` streaming logic is correct, but the current logic is not good either. It was added with the assumption that `mysql_query()` is unlikely to be used as a subquery. But with with `RemoteExecute` it is the likely usage now. Needs to be reworked to be on par with Postgres
 - transactions need to be started by both `mysql_scan` and `mysql_query`
`FunctionExpression`s, currently only started by `mysql_scan`

Mytherin:
Sounds great! I'm planning to do a bit more work here w.r.t. allowing arbitrary statements to be pushed down, which can then translate into pushing down various DDL statements as well at this level (`DropStatement`, `CreateXStatement`, `AlterStatement`). I think `MergeIntoStatement` is also currently missing.

I'll push all of that into `quack` first but this can then be integrated into Postgres / MySQL later.

The main advantage to doing it all at this level is that we don't need to do binding at all and we can just ship SQL directly to the remote catalog. This also means that we are doing much less local binding which likely means we can get rid of the metadata cache in the catalog extensions.

It would even be possible to build up a remote catalog without implementing most of the catalog abstractions. If everything us pushed as SQL to the catalog it’s not necessary to create TableCatalogEntry, etc anymore. We can just always either push down or fail. Even queries like “FROM remote.tbl” can be “pushed down” into FROM mysql_query(“SELECT * FROM tbl”).

The main missing link there are hybrid DML statements, e.g. INSERT INTO <remote_tbl> FROM <local_tbl>. This can’t be represented using the existing Query infrastructure outside of pushing e.g. a large VALUES clause, but we will also want to stream these. For these we will still need to create an operator. We could perhaps generalize that as well in some capacity but that would require some more thought.

Issue #22992, opened 2026-06-01
File modification times are wrongly reported on Windows


billziss-gh:
### What happens?

On Windows file modification times are erroneously reported. For example a newly created file can be reported to have a modification time in the future. This problem can happen in SQL that uses `read_blob` or in extensions that use `LocalFileSystem::GetLastModifiedTime` (e.g. `cache_httpfs` which is how I discovered this problem).

### To Reproduce

The following SQL:

```sql
COPY (SELECT 42 AS i) TO 'duckdb_mtime_probe.csv';
SELECT
    filename,
    last_modified,
    now() AS duckdb_now,
FROM
    read_blob('duckdb_mtime_probe.csv');
```

Produces the following output on Windows. Notice how the last_modifed timestamp is ahead of `now()`.

```
┌────────────────────────┬──────────────────────────┬───────────────────────────────┐
│        filename        │      last_modified       │          duckdb_now           │
│        varchar         │ timestamp with time zone │   timestamp with time zone    │
├────────────────────────┼──────────────────────────┼───────────────────────────────┤
│ duckdb_mtime_probe.csv │ 2026-06-02 02:18:57+03   │ 2026-06-01 23:18:57.926621+03 │
└────────────────────────┴──────────────────────────┴───────────────────────────────┘
```

For comparison here is (correct) output on macOS:

```
┌────────────────────────┬──────────────────────────┬───────────────────────────────┐
│        filename        │      last_modified       │          duckdb_now           │
│        varchar         │ timestamp with time zone │   timestamp with time zone    │
├────────────────────────┼──────────────────────────┼───────────────────────────────┤
│ duckdb_mtime_probe.csv │ 2026-06-01 23:23:05+03   │ 2026-06-01 23:23:05.530427+03 │
└────────────────────────┴──────────────────────────┴───────────────────────────────┘
```

### OS:

Windows 11

### DuckDB Version:

v1.5.2

### DuckDB Client:

CLI

### Full Name:

Bill Zissimopoulos

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

PR #23136, opened 2026-06-08
Fix windows last modification timestamp

dentiny:
Closes https://github.com/duckdb/duckdb/issues/22992

`FileMetadata`'s modification timestamp expects UTC timezone.
- For windows implementation, we convert it into local timezone [here](https://github.com/duckdb/duckdb/blob/3bafe7a84c9b467ef65a23c278961332c6f5f597/src/common/local_file_system.cpp#L1077)
- For unix one, it's [kept](https://github.com/duckdb/duckdb/blob/3bafe7a84c9b467ef65a23c278961332c6f5f597/src/common/local_file_system.cpp#L238) in UTC timezone

### DIFF
--- a/src/common/local_file_system.cpp
+++ b/src/common/local_file_system.cpp
@@ -880,7 +880,7 @@ static timestamp_t FiletimeToTimeStamp(FILETIME file_time) {
 	// Adapted from: https://stackoverflow.com/questions/6161776/convert-windows-filetime-to-second-in-unix-linux
 	const auto WINDOWS_TICK = 10000000;
 	const auto SEC_TO_UNIX_EPOCH = 11644473600LL;
-	return Timestamp::FromTimeT(fileTime64 / WINDOWS_TICK - SEC_TO_UNIX_EPOCH);
+	return Timestamp::FromEpochSeconds(fileTime64 / WINDOWS_TICK - SEC_TO_UNIX_EPOCH);
 }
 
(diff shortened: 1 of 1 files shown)


src/common/local_file_system.cpp
```diff
@@ -880,7 +880,7 @@ static timestamp_t FiletimeToTimeStamp(FILETIME file_time) {
 	// Adapted from: https://stackoverflow.com/questions/6161776/convert-windows-filetime-to-second-in-unix-linux
 	const auto WINDOWS_TICK = 10000000;
 	const auto SEC_TO_UNIX_EPOCH = 11644473600LL;
-	return Timestamp::FromTimeT(fileTime64 / WINDOWS_TICK - SEC_TO_UNIX_EPOCH);
+	return Timestamp::FromEpochSeconds(fileTime64 / WINDOWS_TICK - SEC_TO_UNIX_EPOCH);
```
billziss-gh, line 883: @dentiny I agree with your fix. Windows [`FILETIME`](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime) is already in UTC and therefore it makes sense not to change timezones.

I note that this is inline with the UNIX implementation which also uses `Timestamp::FromEpochSeconds`.

Mytherin:
Thanks!

PR #23138, opened 2026-06-08
[Parquet][Dev] Fix prefetching error on skipped column

Tishj:
See commit for more details
Encountered while working on duckdb-iceberg

We were reaching `GetFileOffset` while populating `pending_registrations` for a column that had been skipped (`InitializeRead` was never called), so the `std::runtime_error` in `GetFileOffset` was getting hit.


Mytherin:
Thanks!

PR #23140, opened 2026-06-08
Fix native geometry parquet stats pruning and add `OPERATOR_ROW_GROUPS_SCANNED` to parquet reader

Maxxen:
This PR makes two minor fixes to the native parquet geometry filter pushdown:
1. Dont expose the fact that the parquet geometry schema type has a child column schema for the underlying BLOB data - that makes the `MultiFileReader` think that the geometry column within a file is different from the global schema and thus won't push down filters (as it thinks it needs to be cast first)
2. Look past `GEOMETRY(crs?) -> GEOMETRY(other_crs?)` casts when looking for constant bboxes in the `GeometryStatistics::CheckZoneMap` code. These casts are inserted automatically in lots of places, but only change the type-metadata (not the actual geometry values) and should not prevent filter pushdown.

Additionally, in order to write a regression test, this PR builds on top of https://github.com/duckdb/duckdb/pull/22339 and adds support for emitting `OPERATOR_ROW_GROUPS_SCANNED` and `OPERATOR_TOTAL_ROW_GROUPS_TO_SCAN` metrics from the parquet reader. This makes it much easier for users to determine how effective the stats in their parquet files are, and for us to write tests to verify that row group pruning actually takes place (without just relying on httpfs number of requests/bytes read)


Maxxen:
I think the CI failure is unrelated, it succeeded on my fork https://github.com/Maxxen/duckdb/actions/runs/27201083642/job/80325278487

Mytherin:
Thanks!

PR #23142, opened 2026-06-08
Parquet Reader with Async I/O

pdet:
This PR continues the work on implementing async I/O for the parquet reader.

The first thing we did was a further refactor to also split the filter-processing path (i.e. the `ParquetPrefetchStrategy::PREFETCH_FILTERS` strategy) into a schedule and a process step. This is necessary because this prefetch strategy initially only prefetches the filter columns, checks if any rows survive, and only then prefetches the remaining columns. As a result, scheduling also happens in the process step, not just the initial schedule. This is unlike `WHOLE_GROUP` and `COLUMN_WISE_EAGER`, which schedule all of their I/O up front.

Second, we wired these functions to return AsyncResult tasks scheduled on the `TaskSchedulerType::ASYNC` pool, so the scan returns `BLOCKED` and its I/O runs on the dedicated async thread pool instead of blocking the worker thread.

## Benchmark
To benchmark this we created a dataset of 16 files, about 8.2 GB total, with 17 columns (16 random doubles plus one filter column), 64M rows across 640 row groups. We stored it on S3 and read it from a c6in.4xlarge instance in the same region. Across the workloads we measured, async I/O gives roughly a 1.5x to 1.7x speedup over main. All numbers are the mean of 5 timed runs after 1 warmup

| Workload | main | async (16T) | async (32T) | Improvement (32T vs main) |
|---|---|---|---|---|
| All columns (whole_group) | 10.19 s | 8.15 s | 6.08 s | 1.68× |
| Column-wise eager (12 of 17 cols) | 9.34 s | 6.12 s | 6.34 s | 1.47× |
| Filtered (`WHERE g=500`) | 8.92 s | 6.83 s | 5.62 s | 1.59× |

One thing we learnt from this benchmark is that defaulting our async threads to 2x the system threads seems sensible, so this PR also changes the default accordingly.

## Next steps
  1. We should have our parquet reads from local disk also default to the prefetch strategy, so they will do async too. We probably have to revisit the network cost model to make sure it is sensible for local speeds.
  2. On httpfs, each thread has a single file handle, and if we want to do parallel reads to httpfs we might need more connections. This could be the next performance booster here.

### DIFF
--- a/extension/parquet/include/parquet_reader.hpp
+++ b/extension/parquet/include/parquet_reader.hpp
@@ -197,7 +197,7 @@ public:
 	int64_t current_group;
 	idx_t offset_in_group;
 	idx_t group_offset;
-	unique_ptr<CachingFileHandle> file_handle;
+	shared_ptr<CachingFileHandle> file_handle;
 	vector<unique_ptr<ColumnReader>> column_readers;
 	duckdb_base_std::unique_ptr<duckdb_apache::thrift::protocol::TProtocol> thrift_file_proto;
 
--- a/extension/parquet/include/parquet_reader.hpp
+++ b/extension/parquet/include/parquet_reader.hpp
@@ -394,6 +394,12 @@ private:
 	AsyncResult Schedule(ClientContext &context, ParquetReaderScanState &state, DataChunk &result, bool log_prefetch);
 	//! Process up to STANDARD_VECTOR_SIZE rows of the current row group into result.
 	SourceResultType Process(ParquetReaderScanState &state, DataChunk &result, bool log_prefetch);
+	//! Run the filters
+	idx_t EvaluateFilters(ParquetReaderScanState &state, DataChunk &result, vector<bool> &need_to_read,
+	                      idx_t scan_count, uint8_t *define_ptr, uint8_t *repeat_ptr, bool log_prefetch);
+	//! Read the remaining (non-filter) columns into result.
+	void DecodeRemainingColumns(ParquetReaderScanState &state, DataChunk &result, const vector<bool> &need_to_read,
+	                            idx_t filter_count, uint8_t *define_ptr, uint8_t *repeat_ptr);
 	ParquetColumnSchema ParseColumnSchema(const SchemaElement &s_ele, idx_t max_define, idx_t max_repeat,
 	                                      idx_t schema_index, idx_t column_index,
 	                                      ParquetColumnSchemaType type = ParquetColumnSchemaType::COLUMN);
--- a/extension/parquet/include/parquet_reader.hpp
+++ b/extension/parquet/include/parquet_reader.hpp
@@ -394,12 +394,12 @@ private:
 	AsyncResult Schedule(ClientContext &context, ParquetReaderScanState &state, DataChunk &result, bool log_prefetch);
 	//! Process up to STANDARD_VECTOR_SIZE rows of the current row group into result.
 	SourceResultType Process(ParquetReaderScanState &state, DataChunk &result, bool log_prefetch);
-	//! Run the filters
-	idx_t EvaluateFilters(ParquetReaderScanState &state, DataChunk &result, vector<bool> &need_to_read,
-	                      idx_t scan_count, uint8_t *define_ptr, uint8_t *repeat_ptr, bool log_prefetch);
+	//! Run the filters into state.sel; returns the surviving row count. Advances every filter column.
+	idx_t EvaluateFilters(ParquetReaderScanState &state, DataChunk &result, idx_t scan_count, uint8_t *define_ptr,
+	                      uint8_t *repeat_ptr, bool log_prefetch);
 	//! Read the remaining (non-filter) columns into result.
-	void DecodeRemainingColumns(ParquetReaderScanState &state, DataChunk &result, const vector<bool> &need_to_read,
-	                            idx_t filter_count, uint8_t *define_ptr, uint8_t *repeat_ptr);
+	void DecodeRemainingColumns(ParquetReaderScanState &state, DataChunk &result, idx_t filter_count,
+	                            uint8_t *define_ptr, uint8_t *repeat_ptr);
 	ParquetColumnSchema ParseColumnSchema(const SchemaElement &s_ele, idx_t max_define, idx_t max_repeat,
 	                                      idx_t schema_index, idx_t column_index,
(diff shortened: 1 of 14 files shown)


src/main/config.cpp
```diff
@@ -631,7 +629,8 @@ idx_t DBConfig::GetSystemMaxAsyncThreads(FileSystem &fs) {
 #ifdef DUCKDB_NO_THREADS
 	return 0;
 #else
-	return GetSystemMaxThreads(fs);
+	// via benchmark, it seems that ~2x system threads is the I/O-concurrency sweet spot for remote (S3) scans
+	return 2 * GetSystemMaxThreads(fs);
 #endif
 }
```
carlopi, line 635: I guess one question is: what happens when `DUCKDB_NO_THREADS` is set?

src/execution/operator/scan/physical_table_scan.cpp
```diff
@@ -186,12 +187,15 @@ SourceResultType PhysicalTableScan::GetDataInternal(ExecutionContext &context, D
...
+			{
+				annotated_lock_guard<annotated_mutex> guard(g_state.lock);
+				if (g_state.CanBlock()) {
+					data.async_result.ScheduleTasks(input.interrupt_state, context.pipeline->executor);
+					return SourceResultType::BLOCKED;
+				}
 			}
-			return SourceResultType::FINISHED;
+			data.async_result.ExecuteTasksSynchronously();
+			return SourceResultType::HAVE_MORE_OUTPUT;
```
carlopi, line 198: I am not seeing why the code was `return FINISHED` before, and I am not sure here we should skip `REGULAR` tasks we are waiting on? Like, could this be ever leading to re-entrancy issues? Or some other weirdness?

src/main/config.cpp
pdet: Worker threads can also do async work, but async threads can only do async work, so it should work fine, but ofc serialized.

src/execution/operator/scan/physical_table_scan.cpp
pdet: Good question, the reason I hit this was due to an order by + limit + filter, that could create a race condition and drop the wrong prefetching task, issuing a wrong result, so `return FINISHED` was clearly buggy.

Maybe @lnkuiper has opinions?

extension/parquet/include/parquet_reader.hpp
```diff
@@ -197,7 +197,7 @@ struct ParquetReaderScanState {
 	int64_t current_group;
 	idx_t offset_in_group;
 	idx_t group_offset;
-	unique_ptr<CachingFileHandle> file_handle;
+	shared_ptr<CachingFileHandle> file_handle;
```
Mytherin, line 200: Given that we're now doing parallel reads on this file handle I think we need to ensure we pass in `FileOpenFlags::FILE_FLAGS_PARALLEL_ACCESS` otherwise we might run into race conditions in the httpfs caches.

extension/parquet/parquet_reader.cpp
```diff
@@ -1773,21 +1787,180 @@ AsyncResult ParquetReader::Schedule(ClientContext &context, ParquetReaderScanSta
...
+		}
+		ColumnReaderInput reader_input(result.size(), define_ptr, repeat_ptr);
+		child_reader.Select(reader_input, result_vector, state.sel, filter_count);
+	}
+}
+
+// When doing ParquetPrefetchStrategy::PREFETCH_FILTERS, we gotta block mid processing to get the other columns.
+// on the re-entry the chunk is reset in the multifilescan which borkes our datachunk, hence we need to copy for
+// ownership.
+// FIXME: maybe we can change this to be able to move? or not have the multifile reset? bigger refactor though.
```
Mytherin, line 1905: This is indeed a bit ugly - the multi-file reader should be able to handle the `BLOCKED` call and call without resetting. We can do that in a follow-up.

extension/parquet/parquet_reader.cpp
```diff
@@ -1563,24 +1564,34 @@ void ParquetReader::GetPartitionStats(const duckdb_parquet::FileMetaData &metada
 // An I/O task that fetches the bytes of a ReadHead.
 class ParquetIOAsyncTask : public AsyncTask {
 public:
-	ParquetIOAsyncTask(ReadHead &read_head, CachingFileHandle &file_handle)
-	    : read_head(read_head), file_handle(file_handle) {
+	ParquetIOAsyncTask(ReadHead &read_head, shared_ptr<CachingFileHandle> file_handle,
+	                   std::shared_ptr<duckdb_apache::thrift::transport::TTransport> transport)
```
Mytherin: The lifetimes here are not super clear to me - we don't need `transport`, but as far as I understand we keep it here because it keeps the `ReadHead` alive.

Maybe we can instead turn `read_heads` into a `vector<shared_ptr<ReadHead>>` and pass in a `shared_ptr<ReadHead>` here? Then we don't need to pass in the transport (or make it a shared ptr at all). Then it's also a lot more obvious what's happening.

Mytherin (commented):
Thanks for the PR! Looks good - I left some comments. Maybe you can also locally do a threadsan run with a test config that has `SET prefetch_all_parquet_files=true` just to ensure there's no multithreading issues?

src/main/config.cpp
lnkuiper: Regular workers will schedule a tasks in the async queue, and then block in the pipeline task, then pick up the task themselves from the async queue, so the behaviour should be essentially synchronous

src/execution/operator/scan/physical_table_scan.cpp
lnkuiper: I had Codex dig into this, and it reproduced the issue with `CanBlock() = false`, which is indeed caused by a `LIMIT`. So, I think we should indeed be returning `HAVE_MORE_OUTPUT` here. Alternatively we can add a loop here to avoid re-entering through the pipeline executor but I think this is good already

pdet:
I've also executed :
``` 
TSAN_OPTIONS="suppressions=./.sanitizer-thread-suppressions.txt" build/reldebug/test/unittest --test-config test/configs/prefetch_all_parquet_files.json
```

All tests passed.

lnkuiper:
Looks great!

Mytherin:
Thanks!

PR #23145, opened 2026-06-08
Add support for test config extending

NiclasHaderer:
Sometimes we need to run tests with a slightly different config without duplicating the entire file. This adds an "extends" key to test config JSON that loads one or more base configs, then lets the current file overwrite specific keys.


test/helpers/test_config.cpp
```diff
@@ -378,7 +378,24 @@ void TestConfiguration::LoadConfig(const string &config_path) {
...
+				auto path = child.GetValue<string>();
+				if (!config_dir.empty() && !path.empty() && path[0] != '/') {
+					path = config_dir + "/" + path;
+				}
+				LoadConfig(path);
+			}
+		}
+
 		for (auto &entry : json_values) {
+			if (StringUtil::CIEquals(entry.first, "extends")) {
```
Mytherin: Can this be added to the option list instead of ignoring it in `ParseOption`?

Mytherin (commented):

Mytherin:
Thanks!

PR #23153, opened 2026-06-09
RETURNING support for tables with triggers

krleonid:
## Summary

Allow `RETURNING` on `INSERT`/`UPDATE`/`DELETE` when the target table has triggers — previously this threw `NotImplementedException`.

The DML and its trigger bodies are expanded into a materialized CTE chain (BEFORE bodies → base DML with `RETURNING *` → AFTER bodies), and the user's `RETURNING` list is then bound through **`ReturningBinder` over a base-table binding of the real target table** — the same binder and rules as the no-trigger path — rather than as a plain `SELECT` projection. Binding it as a regular SELECT (an earlier iteration) diverged from `RETURNING` semantics; routing it through `ReturningBinder` gives full parity:

- aggregates / subqueries / window functions are rejected, exactly as on non-triggered tables
- schema/catalog-qualified target columns resolve (`RETURNING schema.tbl.col`)
- a real column named `rowid` is bound as that real column, not the virtual `rowid`
- generated columns return correct values regardless of their position in the schema

Virtual columns (e.g. `rowid`) referenced in `RETURNING` on a `DELETE` are materialized by the base CTE and excluded from `REFERENCING NEW/OLD TABLE` transition-table aliases so trigger bodies don't see them; their sentinel bindings are remapped to the materialized positions.

The trigger-expansion code is extracted out of `binder.cpp` into `src/planner/binder/query_node/bind_trigger_expansion.cpp`.

## Test plan

- [x] `trigger_returning.test` — generated columns, `rowid` (bare/qualified/aliased/in-expressions), `RETURNING *` / `EXCLUDE` / `REPLACE`, `ON CONFLICT DO NOTHING`, `WITH` wrapping, 0-row DML, `UPDATE OF` non-firing, DEFAULT/NULL values, LIST/MAP nested types, case-variant real `rowid` column, negative-parity error cases
- [x] `trigger_returning_binder_parity.test` — aggregates/subqueries/windows rejected, schema-qualified columns, physical `rowid` column shadowing (incl. `REFERENCING OLD TABLE`)
- [x] full `[trigger]` group passes (19 test cases, 1170 assertions)
- [x] `returning` / `upsert` / `insert` / `update` / `delete` / `cte` suites pass; full fast `unittest` suite green
- [x] `make reldebug` builds cleanly


lnkuiper (changes requested):
Thanks for the PR! Looks good overall, but I think there is an issue with `InjectVirtualColumns`. This function is complex and error-prone. Here is a test case that fails:
```sql
# name: test/temp.test
# description: Temporary breaking trigger tests for PR review
# group: [trigger]

require noforcestorage

# Trigger RETURNING should preserve alias binding rules for virtual columns.
# Without triggers this errors because alias x hides the base table name; the
# trigger rewrite must not make t.rowid bindable through the hidden table name.
statement ok
CREATE TABLE alias_rowid_target (id INTEGER);

statement ok
INSERT INTO alias_rowid_target VALUES (1);

statement ok
CREATE TABLE alias_rowid_log (msg VARCHAR);

statement ok
CREATE TRIGGER alias_rowid_trg
AFTER DELETE ON alias_rowid_target
FOR EACH STATEMENT
    INSERT INTO alias_rowid_log VALUES ('deleted');

statement error
DELETE FROM alias_rowid_target AS x RETURNING alias_rowid_target.rowid;
----
Referenced table "alias_rowid_target" not found!

# The explicit target alias should still bind to the virtual rowid.
statement ok
CREATE TABLE alias_rowid_ok_target (id INTEGER);

statement ok
INSERT INTO alias_rowid_ok_target VALUES (1);

statement ok
CREATE TABLE alias_rowid_ok_log (msg VARCHAR);

statement ok
CREATE TRIGGER alias_rowid_ok_trg
AFTER DELETE ON alias_rowid_ok_target
FOR EACH STATEMENT
    INSERT INTO alias_rowid_ok_log VALUES ('deleted');

query I
WITH d AS (DELETE FROM alias_rowid_ok_target AS x RETURNING x.rowid)
SELECT typeof(rowid) FROM d;
----
BIGINT

# Without a target alias, the base table name should still qualify rowid.
statement ok
CREATE TABLE table_rowid_qual_target (id INTEGER);

statement ok
INSERT INTO table_rowid_qual_target VALUES (1);

statement ok
CREATE TABLE table_rowid_qual_log (msg VARCHAR);

statement ok
CREATE TRIGGER table_rowid_qual_trg
AFTER DELETE ON table_rowid_qual_target
FOR EACH STATEMENT
    INSERT INTO table_rowid_qual_log VALUES ('deleted');

query I
WITH d AS (DELETE FROM table_rowid_qual_target RETURNING table_rowid_qual_target.rowid)
SELECT typeof(rowid) FROM d;
----
BIGINT

# Unaliased expressions over virtual columns should keep the same output name as
# the no-trigger path. The trigger rewrite currently exposes the synthetic
# internal virtual-column name, breaking references from an outer query.
statement ok
CREATE TABLE rowid_expr_name_target (id INTEGER);

statement ok
INSERT INTO rowid_expr_name_target VALUES (1);

statement ok
CREATE TABLE rowid_expr_name_log (msg VARCHAR);

statement ok
CREATE TRIGGER rowid_expr_name_trg
AFTER DELETE ON rowid_expr_name_target
FOR EACH STATEMENT
    INSERT INTO rowid_expr_name_log VALUES ('deleted');

query I
WITH d AS (DELETE FROM rowid_expr_name_target RETURNING rowid + 1)
SELECT "(rowid + CAST(1 AS BIGINT))" FROM d;
----
1

# Explicit aliases should be preserved for expressions over virtual columns.
statement ok
CREATE TABLE rowid_expr_alias_target (id INTEGER);

statement ok
INSERT INTO rowid_expr_alias_target VALUES (1);

statement ok
CREATE TABLE rowid_expr_alias_log (msg VARCHAR);

statement ok
CREATE TRIGGER rowid_expr_alias_trg
AFTER DELETE ON rowid_expr_alias_target
FOR EACH STATEMENT
    INSERT INTO rowid_expr_alias_log VALUES ('deleted');

query I
WITH d AS (DELETE FROM rowid_expr_alias_target RETURNING rowid + 1 AS rid_plus_one)
SELECT rid_plus_one FROM d;
----
1

```
Finally, while comments are welcome (I especially like the pass 1/2/3 which make the code more readable), the comments seem very verbose here. Could you minimize the comments to explain only what's necessary?

lnkuiper:
I've looked at this together with Codex in a bit more detail, and I think the main problem is that the `RETURNING` list becomes a regular select list.

There are a few problems:
1. aggregates/subqueries/etc. are now legal if a trigger exists, which would not be legal in a regular `RETURNING`.
2. There are also some issues regarding qualified columns.
3. A real column named `rowid` is incorrectly treated as a virtual column if there are triggers + returning

Here is a test that includes things that break:
```sql
# name: test/temp.test
# description: Temporary breaking trigger RETURNING tests for PR review
# group: [trigger]

require noforcestorage

# Trigger-expanded RETURNING should use the RETURNING binder rules, not SELECT
# projection rules: aggregates are rejected on the no-trigger path.
statement ok
CREATE TABLE returning_agg_target (a INTEGER);

statement ok
CREATE TABLE returning_agg_log (x INTEGER);

statement ok
CREATE TRIGGER returning_agg_trg
AFTER INSERT ON returning_agg_target
FOR EACH STATEMENT
    INSERT INTO returning_agg_log VALUES (1);

statement error
INSERT INTO returning_agg_target VALUES (1), (2) RETURNING sum(a);
----
<REGEX>:Binder Error.*Aggregate functions are not supported.*

# Subqueries are also rejected by ReturningBinder, but the trigger expansion
# currently binds the expression as a normal SELECT expression.
statement ok
CREATE TABLE returning_subquery_target (a INTEGER);

statement ok
CREATE TABLE returning_subquery_log (x INTEGER);

statement ok
CREATE TRIGGER returning_subquery_trg
AFTER INSERT ON returning_subquery_target
FOR EACH STATEMENT
    INSERT INTO returning_subquery_log VALUES (1);

statement error
INSERT INTO returning_subquery_target VALUES (1), (2)
RETURNING a IN (SELECT a FROM returning_subquery_target);
----
<REGEX>:Binder Error.*SUBQUERY is not supported.*

# Window expressions must not become legal only because a trigger is present.
statement ok
CREATE TABLE returning_window_target (a INTEGER);

statement ok
CREATE TABLE returning_window_log (x INTEGER);

statement ok
CREATE TRIGGER returning_window_trg
AFTER INSERT ON returning_window_target
FOR EACH STATEMENT
    INSERT INTO returning_window_log VALUES (1);

statement error
INSERT INTO returning_window_target VALUES (1), (2)
RETURNING row_number() OVER (ORDER BY a), a;
----
<REGEX>:.*WINDOW.*

# Schema-qualified target columns should bind the same way as the no-trigger
# path. The trigger expansion currently aliases the base CTE only as "target".
statement ok
CREATE SCHEMA returning_schema_qual;

statement ok
CREATE TABLE returning_schema_qual.target (a INTEGER);

statement ok
INSERT INTO returning_schema_qual.target VALUES (7);

statement ok
CREATE TABLE returning_schema_qual.log (x INTEGER);

statement ok
CREATE TRIGGER returning_schema_qual_trg
AFTER DELETE ON returning_schema_qual.target
FOR EACH STATEMENT
    INSERT INTO returning_schema_qual.log VALUES (1);

query I
DELETE FROM returning_schema_qual.target
RETURNING returning_schema_qual.target.a;
----
7

# For virtual columns, the qualifier must still match the actual binding. A
# wrong schema should not be stripped and accepted as rowid.
statement ok
CREATE SCHEMA returning_wrong_schema;

statement ok
CREATE TABLE returning_wrong_schema.target (a INTEGER);

statement ok
INSERT INTO returning_wrong_schema.target VALUES (7);

statement ok
CREATE TABLE returning_wrong_schema.log (x INTEGER);

statement ok
CREATE TRIGGER returning_wrong_schema_trg
AFTER DELETE ON returning_wrong_schema.target
FOR EACH STATEMENT
    INSERT INTO returning_wrong_schema.log VALUES (1);

statement error
DELETE FROM returning_wrong_schema.target
RETURNING not_returning_wrong_schema.target.rowid;
----
Referenced table "not_returning_wrong_schema.target" not found!

# A physical column named rowid shadows the virtual rowid. RETURNING *, rowid
# should preserve the no-trigger column order and types.
statement ok
CREATE TABLE returning_physical_rowid_target (rowid INTEGER, id VARCHAR);

statement ok
INSERT INTO returning_physical_rowid_target VALUES (42, 'x');

statement ok
CREATE TABLE returning_physical_rowid_log (x INTEGER);

statement ok
CREATE TRIGGER returning_physical_rowid_trg
AFTER DELETE ON returning_physical_rowid_target
FOR EACH STATEMENT
    INSERT INTO returning_physical_rowid_log VALUES (1);

query ITI
DELETE FROM returning_physical_rowid_target RETURNING *, rowid;
----
42	x	42

# The same shadowing bug can affect trigger bodies: the physical rowid column
# must remain visible in REFERENCING OLD TABLE aliases.
statement ok
CREATE TABLE returning_physical_rowid_ref_target (rowid INTEGER, id VARCHAR);

statement ok
INSERT INTO returning_physical_rowid_ref_target VALUES (42, 'x');

statement ok
CREATE TABLE returning_physical_rowid_ref_cap (rowid INTEGER, id VARCHAR);

statement ok
CREATE TRIGGER returning_physical_rowid_ref_trg
AFTER DELETE ON returning_physical_rowid_ref_target
REFERENCING OLD TABLE AS old_rows
FOR EACH STATEMENT
    INSERT INTO returning_physical_rowid_ref_cap SELECT * FROM old_rows;

query I
DELETE FROM returning_physical_rowid_ref_target RETURNING rowid;
----
42

query IT
SELECT * FROM returning_physical_rowid_ref_cap;
----
42	x

```

I think we should be using `ReturningBinder` still even if the table has triggers. Something like this:
```C++
BoundStatement Binder::ExpandTriggers(...) {
	auto user_returning = std::move(GetDMLReturningList(node));
	bool has_returning = !user_returning.empty();

	auto base_node = node.Copy();
	auto &base_returning = GetDMLReturningList(*base_node);

	// The base CTE should materialize the rows needed by triggers and by the
	// final RETURNING projection. It can project a stable implementation-level
	// row shape, but the user expressions should not be rebound as SELECT exprs.
	base_returning.push_back(make_uniq<StarExpression>());
	auto injected_virtuals = AddOnlyActuallyBoundVirtualColumns(user_returning, base_returning, table, node);

	auto base_cte = BuildMaterializedBaseCTE(std::move(base_node));
	auto trigger_ctes = BuildBeforeAndAfterTriggerCTEs(base_cte_name, injected_virtuals);

	if (!has_returning) {
		auto outer = BuildChangedRowsSelect(base_cte_name, trigger_ctes);
		return Bind(*outer);
	}

	// Bind the user's RETURNING list with the same binder used by the no-trigger
	// path. The child operator is a scan of the materialized base CTE with a
	// table binding that mirrors the target table binding closely enough for
	// RETURNING semantics, including alias/schema/catalog and virtual columns.
	auto returning_child = BuildCTEScanAsReturningTarget(base_cte_name, table, node);
	auto result = BindReturningLikeDML(std::move(user_returning),
	                                  table,
	                                  GetTargetBindingAlias(node, table),
	                                  returning_table_index,
	                                  std::move(returning_child),
	                                  GetVirtualColumnsActuallyProjected(base_returning));

	AttachTriggerCTEsToPlan(result.plan, std::move(trigger_ctes), std::move(base_cte));
	result.properties.output_type = QueryResultOutputType::FORCE_MATERIALIZED;
	result.properties.return_type = StatementReturnType::QUERY_RESULT;
	return result;
}
```

lnkuiper:
Thanks for the changes! Looks good now :)

lnkuiper (approved):

lnkuiper:
static libs failure is unrelated, I'll rerun once the rest is done :)

lnkuiper:
We ran into a conflict at the last second, could you fix and push again?

PR #23157, opened 2026-06-09
Add file_row_group_number virtual column to the Parquet reader

hannes:
This adds a virtual column `file_row_group_number` to the parquet reader that allows (somewhat) efficient reading of individual row groups from parquet files, e.g.
```SQL
FROM 'my_file.parquet' WHERE file_row_group_number=42;
```
or 
```SQL
FROM 'my_file.parquet' WHERE file_row_group_number IN (42, 43);
```

### DIFF
--- a/extension/parquet/include/parquet_reader.hpp
+++ b/extension/parquet/include/parquet_reader.hpp
@@ -222,6 +222,11 @@ public:
 	//! (optional) pointer to the PhysicalOperator for logging
 	optional_ptr<const PhysicalOperator> op;
 
+	//! (optional) counters (owned by the scan's global state) for row groups whose data was read / skipped,
+	//! incremented as row groups are processed and surfaced as profiling metrics
+	optional_ptr<atomic<idx_t>> row_groups_read;
+	optional_ptr<atomic<idx_t>> row_groups_skipped;
+
 	//! Prefetch cost model
 	PrefetchCostModelState cost_model_state;
 };
@@ -287,6 +292,10 @@ struct ParquetUnionData : public BaseUnionData {
 };
 
 class ParquetReader : public BaseFileReader {
+public:
+	//! Virtual column identifier for the "file_row_group_number" column (the file-relative row group index of each row)
+	static constexpr column_t COLUMN_IDENTIFIER_FILE_ROW_GROUP_NUMBER = UINT64_C(9223372036854775820);
+
 public:
 	ParquetReader(ClientContext &context, OpenFileInfo file, ParquetOptions parquet_options,
 	              shared_ptr<ParquetFileMetadataCache> metadata = nullptr);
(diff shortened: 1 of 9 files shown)


extension/parquet/include/parquet_reader.hpp
```diff
@@ -222,6 +222,11 @@ struct ParquetReaderScanState {
 	//! (optional) pointer to the PhysicalOperator for logging
 	optional_ptr<const PhysicalOperator> op;
 
+	//! (optional) counters (owned by the scan's global state) for row groups whose data was read / skipped,
+	//! incremented as row groups are processed and surfaced as profiling metrics
+	optional_ptr<atomic<idx_t>> row_groups_read;
```
Mytherin: Maybe this can just be `idx_t row_groups_read` that we track locally, and then flush to the global state in `MultiFileReaderInterface::FinishReading`? Making this an atomic can cause contention, but more importantly this being a pointer to the global state also introduces a lifetime dependency between the two which can always end up causing hard to track down issues.

extension/parquet/reader/row_number_column_reader.cpp
```diff
@@ -64,4 +64,28 @@ idx_t RowNumberColumnReader::Read(ColumnReaderInput &input, Vector &result) {
...
+                                          TProtocol &protocol_p) {
+	row_group_idx = row_group_idx_p;
+}
+
+idx_t RowGroupColumnReader::Read(ColumnReaderInput &input, Vector &result) {
+	auto &num_values = input.num_values;
+
+	// the row group number is constant for all rows within a row group
+	auto value = UnsafeNumericCast<int64_t>(row_group_idx);
+	auto data_ptr = FlatVector::Writer<int64_t>(result, num_values);
```
Mytherin: Can we emit a `ConstantVector` here?

extension/parquet/parquet_column_schema.cpp
```diff
@@ -95,12 +95,32 @@ ParquetColumnSchema ParquetColumnSchema::FileRowNumber() {
...
 
+ParquetColumnSchema ParquetColumnSchema::FileRowGroupNumber() {
+	ParquetColumnSchema res;
+	res.name = "file_row_group_number";
+	res.max_define = 0;
+	res.max_repeat = 0;
+	res.schema_index = 0;
+	res.column_index = 0;
+	res.schema_type = ParquetColumnSchemaType::FILE_ROW_GROUP_NUMBER;
+	res.type = LogicalType::BIGINT, res.repetition_type = duckdb_parquet::FieldRepetitionType::type::OPTIONAL;
```
Mytherin: Can we keep this on separate lines?

Also maybe `UBIGINT` as return type?

Mytherin (commented):

extension/parquet/include/parquet_reader.hpp
hannes: Indeed, but this might be nicer just to be a metric then?

Mytherin: Yes, that makes sense

Mytherin:
Thanks!

Issue #23143, opened 2026-06-08
JSON keys are case-sensitive at the top level but case-insensitive at nested levels in read_json


DerZc:
### What happens?

Hi,

I found the function read_json, JSON keys are case-sensitive at the top level but case-insensitive at nested levels, this is an example:

First, create two JSON files:
The first one is `/tmp/test1.json` with `{"x":1,"X":2}`, the second is `/tmp/test2.json` with `{"outer":{"x":1,"X":2}}`.

Then execute the following queries:

```
SELECT * FROM read_json('/tmp/test1.json');
┌───────┬───────┐
│   x   │  X_1  │
│ int64 │ int64 │
├───────┼───────┤
│     1 │     2 │
└───────┴───────┘

SELECT * FROM read_json('/tmp/test2.json');
Not implemented Error:
Duplicate name "X" in struct auto-detected in JSON, try ignore_errors=true

LINE 1: SELECT * FROM read_json('/tmp/test2.json');
```

### To Reproduce

First, create two JSON files:
The first one is `/tmp/test1.json` with `{"x":1,"X":2}`, the second is `/tmp/test2.json` with `{"outer":{"x":1,"X":2}}`.

Then execute the following queries:

```
SELECT * FROM read_json('/tmp/test1.json');

SELECT * FROM read_json('/tmp/test2.json');
```

### OS:

Ubuntu 24.04

### DuckDB Version:

d7b51fbc7d2cc1ba81847d6ecfc25a524e2247ab

### DuckDB Client:

CLI

### Full Name:

Chi Zhang

### Affiliation:

Tsinghua University

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

dentiny:
I'm at main branch d5b9549e39a3db7bd2fb8cdf3fd9b0a989423202, seems to work on my end
```sql
memory D COPY (SELECT '{"x":1,"X":2}') TO '/tmp/test1.json';
memory D COPY (SELECT '{"outer":{"x":1,"X":2}}') TO '/tmp/test2.json';
memory D SELECT * FROM read_json('/tmp/test1.json');
┌─────────────────┐
│ '{"x":1,"X":2}' │
│     varchar     │
├─────────────────┤
│ {"x":1,"X":2}   │
└─────────────────┘
[...]

Issue #10751, opened 2024-02-19
`NotImplementedException` - Querying JSON file with duplicate column names crashes CLI 


hello-world-bfree:
### What happens?

When querying a newline-delimited JSON file with `read_json_auto` DuckDB crashes with a `NotImplementedException`:

```
libc++abi: terminating due to uncaught exception of type duckdb::NotImplementedException: {"exception_type":"Not implemented","exception_message":"Error while casting - duplicate name \"platform\" in struct"}
```

The column name `platform` is just the first of a number of the same kind of duplicates. They're duplicates in the sense that their lower cased representation matches - "Platform" == "platform". This jives with DuckDB elsewhere; i.e.,

```
create or replace table test as select 1 as id, struct_pack(platform := 2, "Platform" := 3);
Error: Binder Error: Duplicate struct entry name "Platform"
```

So rejecting these is consistent. The bug is the lack of error handling or validation.

If the file is altered so that the field names are truly identical - `platform` == `platform` instead of `Platform` and `platform`, we get the expected behavior and no crash:

```
select * from read_json_auto('bug_report.json', format = 'newline_delimited');
Error: Invalid Input Error: Duplicate key "platform" in object {"time":1707878164,"distinct_id":"20110941","$app_version_string":"9.10.1","$city":"Cookstown","$insert_id":"244b811e-378a-5444-b373-d1a5d6e9bba0","$os":"Android","platform":"google","platform":"google","event_id":1037203}
```

### To Reproduce

Create a table with the [attached file](https://github.com/duckdb/duckdb/files/14334220/bug_report.json):

```
create table json_test as select * from read_json_auto('bug_report.json', format = 'newline_delimited');
```

Query the newly created table:

```
select * from json_test;
```

### OS:

macOS

### DuckDB Version:

0.10.0

### DuckDB Client:

CLI

### Full Name:

Brandon Freeman

### Affiliation:

Hallow

### Have you tried this on the latest [nightly build](https://duckdb.org/docs/installation/?version=main)?

I have not tested with any build

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [X] Yes, I have

szarnyasg:
Thanks! Reproduced, we'll take a look.

Mytherin:
[...]

PR #23158, opened 2026-06-09
Fix json nested field case-sensitivity

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23143

As the issue reporter mentioned, currently outer and inner json keys are treated differently, with the former being case insensitive while the later throw exception.

In the PR, I mimic how we handle outer fields with `RenameCaseInsensitiveDuplicate`, namely to record the existent keys in a map and suffix with appearance counter.

Related to https://github.com/duckdb/duckdb/issues/10751, which we require explicit `ignore_errors` to suppress.
I have ran CI in my fork https://github.com/dentiny/duckdb/pull/109, failures seem unrelated.


lnkuiper (changes requested):
I don't think this works across different levels of nestedness:
```sql
# name: test/temp.test
# description: Temporary JSON duplicate nested field repro
# group: [json]

require json

statement ok
COPY (SELECT '{"left":{"a":1,"A":2},"right":{"a":3,"A_1":4}}') TO '__TEST_DIR__/nested_duplicate_key_repro.json' (FORMAT csv, QUOTE '', HEADER 0, DELIMITER '\t');

query II
SELECT "left", "right" FROM read_json('__TEST_DIR__/nested_duplicate_key_repro.json');
----
{'a': 1, 'A_1': 2}	{'a': 3, 'A_1': 4}
```

Whatever structure we have deal with nested duplicate (case sensitive) keys, it has to have a nested structure as well.

I would really like to have this fixed, but I think it requires something like this:
```C++
struct JSONTransformHelper {
    vector<JSONTransformHelper> children;
};
```
This could also hold stuff like `projected_indices` and allocated `Vector`s which are currently done on-the-fly, but could be cached across iterations.

This might be a bit too involved for a bugfix release, but the current state of this PR is not mergeable yet.

dentiny:
Discussed offline: the nested json is of low-priority and involved, I will close the PR for now.

PR #23160, opened 2026-06-09
Parquet DBP fast skip decode

flashmouse:
This PR actually copy idea from [arrow-rs](https://github.com/apache/arrow-rs/pull/9769).  fast decode/skip DBP data when bitwidth=0


Mytherin:
Thanks - can we add a test as well? Could you also share what the benefits are of this optimization?

flashmouse:
> Thanks - can we add a test as well? Could you also share what the benefits are of this optimization?

benchmark in this PR shows:

this branch:
```
build/release/benchmark/benchmark_runner "benchmark/micro/parquet/dbp_bw0_skip.benchmark"       

name    run     timing
benchmark/micro/parquet/dbp_bw0_skip.benchmark  1       0.019439
benchmark/micro/parquet/dbp_bw0_skip.benchmark  2       0.018726
benchmark/micro/parquet/dbp_bw0_skip.benchmark  3       0.018765
benchmark/micro/parquet/dbp_bw0_skip.benchmark  4       0.018660
benchmark/micro/parquet/dbp_bw0_skip.benchmark  5       0.018758
```

main branch:
```
build/release/benchmark/benchmark_runner "benchmark/micro/parquet/dbp_bw0_skip.benchmark"   

name	run	timing
benchmark/micro/parquet/dbp_bw0_skip.benchmark	1	0.034843
benchmark/micro/parquet/dbp_bw0_skip.benchmark	2	0.034945
benchmark/micro/parquet/dbp_bw0_skip.benchmark	3	0.035426
benchmark/micro/parquet/dbp_bw0_skip.benchmark	4	0.034851
benchmark/micro/parquet/dbp_bw0_skip.benchmark	5	0.034761
```

If bw = 0, that means the data in the column is very regular. For example, a column has values [1, 3, 5, 7, 9]. In that case, first_value = 1, delta = 2, bitwidth = 0, and the bitpacked encode is an empty byte. When SKIP_READ is set, we can skip creating target_values.

I just realized that it seems if bw = 0, we may could also skip ``ParquetDecodeUtils::BitUnpackAligned``, but currently I think the result of this pr is enough.

PR #23161, opened 2026-06-09
Make `Identifier` a class instead of using string everywhere

Mytherin:
Identifiers are a central concept in DuckDB, we use them in many locations:

* Names of tables, views, ...
* Attached database names, catalog names, schema names, ...
* Names of columns
* Expression aliases, table aliases
* Names of functions, aggregates, window functions
* Function parameter names
* Struct field names
* Names of types
* Secret types, option names, ....

Currently identifiers are modeled as regular strings - but they have semantics that are different from strings. In particular, in DuckDB identifiers are currently case-insensitive. This means that everywhere where we are operating on identifiers we need to *remember* that they are case-insensitive, and use appropriate tools (`StringUtil::CIHash`, `StringUtil::CIEquals`, `case_insensitive_set_t`, `case_insensitive_map_t`, etc). This ends up being quite a lot of places in the codebase, and naturally this goes wrong in many cases.

This PR makes it more explicit which strings are identifiers by adding the `Identifier` class and using that instead of `string` everywhere where identifiers are used. This class automatically follows the correct rules for comparisons (comparisons are case-insensitive). For maps / sets, the `identifier_map_t` and `identifier_set_t` can be used. The advantage to this approach are two-fold:

* Type-safety means we can't forget about semantics anymore. Identifiers are always going to compare in a case-insensitive - no special code required.
* If we want to ever change this behavior or make it configurable, there is now one central place to do so, instead of the semantics of identifiers being scattered throughout the entire codebase.

Because identifiers are such a pervasive type, and effectively account for the majority of the `string` types in the system, this PR ended up being quite massive. The PR is not perfect and there's more follow-up improvements that I would like to make but I think it's good to get this initial refactor in and we can do further iterations / clean-up afterwards. In particular, two follow-up changes I want to make:

* Turn `QueryResult::names` into `vector<Identifier>`
* Refactor the table function `Bind` callback (or maybe add a `BindV2`) that emits `Identifier` instead of `vector<string> names`.


src/include/duckdb/common/identifier.hpp
```diff
@@ -0,0 +1,200 @@
...
+	Identifier() = default;
+	//! Construction from a string literal is implicit: literals in the source are identifiers by intent.
+	Identifier(const char *str) : value(str) { // NOLINT: implicit conversion from literals is intentional
+	}
+	//! Construction from a runtime string is explicit: an Identifier carries case-insensitive semantics, so
+	//! promoting a runtime string must be a deliberate choice at the call site.
+	explicit Identifier(const string &str) : value(str) {
+	}
+	explicit Identifier(string &&str) : value(std::move(str)) {
+	}
```
Maxxen, line 37: Do we want a debug-mode UTF8 validation check here?

Mytherin: I think that would definitely make sense but I'll save it for a follow-up - for now I mostly just want to get this in to avoid having to deal with more merge conflicts :)

Mytherin:
CC @smvv I've disabled the tidy-check diff in this PR, it's taking forever (>6 hours) to run because of the large changeset, and we're already running tidy-check with clangd on all files so I think it's not providing much value anymore at this point. Maybe there's a reason to keep it around, I'll let you pick this up in a follow-up.

CC @staticlibs I've disabled mysql / postgres in this PR. The patches are there and the patches work but the patching mechanism doesn't work correctly with the submodules (`database-connector`) that are present in those repositories. In particular the "double-apply" doesn't work - see [this CI failure where the first build succeeds but the second one fails](https://github.com/duckdb/duckdb/actions/runs/27260187211/job/80511353591).

staticlibs:
Thanks for heads up! Submodules patching should be fixed in #23191.

Mytherin:
Thanks - can you also re-enable the mysql / postgres extensions there?

PR #23165, opened 2026-06-09
Fold complement operations when NO_NULLs

carlopi:
Idea is that Equal(left, right) and NotEqual(left, rigth), when there is guarantee there are no NULLs can be folded in a single instantiation. Same for LessThan / GreaterThan / LessEqualThan / GreaterEqualThan, that can be folded in a single instantiation.

This reduces binary sizes by about 0.5%, for no performance penalty and limited additional code complexity.

### DIFF
--- a/src/include/duckdb/common/vector_operations/binary_executor.hpp
+++ b/src/include/duckdb/common/vector_operations/binary_executor.hpp
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "duckdb/common/exception.hpp"
+#include "duckdb/common/operator/comparison_operators.hpp"
 #include "duckdb/common/optional.hpp"
 #include "duckdb/common/types/vector.hpp"
 #include "duckdb/common/vector/constant_vector.hpp"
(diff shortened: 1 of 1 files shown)


src/include/duckdb/common/vector_operations/binary_executor.hpp
```diff
@@ -19,6 +20,28 @@
...
+	static constexpr bool FOLD = false;
+};
+template <>
+struct ComparisonSelectComplement<NotEquals> {
+	static constexpr bool FOLD = true;
+	using COMPLEMENT = Equals;
+	static constexpr bool SWAP_OPERANDS = false;
+};
+template <>
+struct ComparisonSelectComplement<GreaterThanEquals> {
```
Mytherin, line 38: Can the same not be applied to `LessThanEquals`?

carlopi: I think that's currently should be corrently done elsewhere, but indeed, let me check check.

carlopi: In particular, in `src/common/vector_operations/comparison_operators.cpp`: https://github.com/duckdb/duckdb/blob/main/src/common/vector_operations/comparison_operators.cpp#L240-L252.

Mytherin: Alright - can we just shorten the comments a bit?

carlopi: Yes, and referenced in a commit that those are handled elsewhere.

Mytherin:
Thanks!

PR #23166, opened 2026-06-09
BinaryExecutor: unrolled path, also there perform blind store on selection vector

carlopi:
Idea, already applied in other paths, is that location is always valid to write to, and will be (if it is to be used) rewritten.

Minor local optimization to align the code to what already done elsewhere in the same file.


Mytherin:
Thanks!

PR #23169, opened 2026-06-09
Remove deprecated settings 'enable_http_logging' and 'http_logging_output'

carlopi:
Reducing some complexity, both as unused code and reducing/cleaning up SQL surface.


Mytherin:
Thanks!

PR #23173, opened 2026-06-09
Pluggable combine types rules

yan-alex:
[Step 3](https://github.com/duckdblabs/duckdb-internal/issues/9598) out of 3 to [Make type combine functions extensible](https://github.com/duckdblabs/duckdb-internal/issues/9595)

Tried to apply these changes on top of my duckdb fork to try it with spark-compat extension but got conflicts. It seems like too much has changed since my duckdb fork's base to patch these changes cleanly on top. A full rebase to main might be in order.


src/include/duckdb/function/cast/cast_function_set.hpp
```diff
@@ -64,11 +84,25 @@ class CastFunctionSet {
...
+	                          LogicalType &result, bool &success);
+	bool TryCombineUnequalTypes(LogicalTypeResolver &resolver, const LogicalType &left, const LogicalType &right,
+	                            LogicalType &result, bool &success);
+
 private:
 	optional_ptr<DBConfig> config;
 	vector<BindCastFunction> bind_functions;
 	//! If any custom cast functions have been defined using RegisterCastFunction, this holds the map
 	optional_ptr<MapCastInfo> map_info;
+	vector<CombineTypesRule> combine_equal_rules;
```
Mytherin: I don't think these need to be two different sets - we should be able to just have one set of combine rules. Could you look into merging the equal / unequal rules into one set of rules?

Mytherin (commented):

src/common/types.cpp
```diff
@@ -1260,6 +1241,13 @@ static bool TryGetMaxLogicalTypeInternal(LogicalTypeResolver &logical_type_resol
...
 		return true;
 	}
+	if (logical_type_resolver.context) {
+		bool success;
+		if (CastFunctionSet::Get(*logical_type_resolver.context)
+		        .TryCombineTypes(logical_type_resolver, left, right, result, success)) {
+			return success;
+		}
+	}
 	if (left.id() != right.id()) {
```
Mytherin: Can we flatten this even more, and also turn the alias rules up there into a rule?

The `UnequalTypeRules` would just start with:

```cpp
if (left.id() == right.id()) {
     // rule only applies to unequal types
    return false;
}
```

And the equal types would start with:

```cpp
if (left.id() != right.id()) {
     // rule only applies to equal types
    return false;
}
```

That way everything is just one list of rules.

src/common/types.cpp
```diff
@@ -1260,6 +1241,13 @@ static bool TryGetMaxLogicalTypeInternal(LogicalTypeResolver &logical_type_resol
 		result = right;
 		return true;
 	}
+	if (logical_type_resolver.context) {
+		bool success;
+		if (CastFunctionSet::Get(*logical_type_resolver.context)
+		        .TryCombineTypes(logical_type_resolver, left, right, result, success)) {
```
Mytherin: This only allows adding new rules and only in a specific location - we should instead have the rules all be stored in the CastFunctionSet, with a default set of rules that is used if the context is not present, similar to what we do for casts. The `CombineTypesRule` should move into the `combine_rules` in the `CastFunctionSet`. That way we could potentially in the future also allow removing rules, re-ordering rules, etc.

src/common/types.cpp
```diff
@@ -1131,7 +1112,7 @@ static bool CombineStructTypes(LogicalTypeResolver &logical_type_resolver, const
 
 struct CombineEqualTypesRule {
```
Mytherin: Is this struct used? Can we just only have the functions?

Mytherin (commented):
Thanks - looking good, left some more comments

Mytherin:
Thanks!

PR #23174, opened 2026-06-09
Fix crash when storage path is not set

guillesd:
I think this is a bit of a design problem again, where an issue is only really caught when calling `logger->Flush()`. Also more in general `logger->Flush()` should not interrupt an exit process, which it now does because the exception is not properly handled.. but that could be the scope of another PR.

This PR simply focuses on:
- Adding a guard against `set logging_storage=file` if there is no storage_path (which by default there cannot be since there is not setting for storage_path!!! We will refactor this in main).
- Adding a check when `call enable_logging` to ensure there is a `storage_path` provided if the `storage=file`.


guillesd:
cc: @carlopi

PR #23177, opened 2026-06-09
Cache Arrow schema in the ArrowQueryResult

evertlammerts:
Fixes the duckdb core part of duckdb/duckdb-python#475 / duckdb/duckdb-spatial#788

This adds caching of the arrow schema to `ArrowQueryResult` so that consumers do not need to rely on an open transaction when consuming an arrow stream or table. Right now there are a number of scenarios where consumers can run into `InternalException`s, e.g. when a schema is constructed that needs catalog access outside of a transaction.

### DIFF
--- a/src/common/arrow/physical_arrow_batch_collector.cpp
+++ b/src/common/arrow/physical_arrow_batch_collector.cpp
@@ -18,16 +18,21 @@ SinkFinalizeType PhysicalArrowBatchCollector::Finalize(Pipeline &pipeline, Event
 	auto total_tuple_count = gstate.data.Count();
 	if (total_tuple_count == 0) {
 		// Create the result containing a single empty result conversion
-		gstate.result = make_uniq<ArrowQueryResult>(statement_type, properties, names, types,
-		                                            context.GetClientProperties(), record_batch_size);
+		auto result = make_uniq<ArrowQueryResult>(statement_type, properties, names, types,
+		                                          context.GetClientProperties(), record_batch_size);
+		result->BuildCachedSchema();
+		gstate.result = std::move(result);
 		return SinkFinalizeType::READY;
 	}
 
 	// Already create the final query result
-	gstate.result = make_uniq<ArrowQueryResult>(statement_type, properties, names, types, context.GetClientProperties(),
-	                                            record_batch_size);
+	auto result = make_uniq<ArrowQueryResult>(statement_type, properties, names, types, context.GetClientProperties(),
+	                                          record_batch_size);
+	// Cache the schema while the producing transaction is still active (see duckdb/duckdb-python#475).
+	result->BuildCachedSchema();
+	auto &arrow_result = *result;
+	gstate.result = std::move(result);
 	// Spawn an event that will populate the conversion result
-	auto &arrow_result = gstate.result->Cast<ArrowQueryResult>();
 	auto new_event = make_shared_ptr<ArrowMergeEvent>(arrow_result, gstate.data, pipeline);
 	event.InsertEvent(std::move(new_event));
 
--- a/src/common/adbc/nanoarrow/schema.cpp
+++ b/src/common/adbc/nanoarrow/schema.cpp
@@ -441,6 +441,13 @@ int ArrowSchemaDeepCopy(struct ArrowSchema *schema, struct ArrowSchema *schema_o
 		return result;
 	}
 
+	// ArrowSchemaInit hard-codes flags to ARROW_FLAG_NULLABLE; preserve the source
+	// flags instead so a deep copy is semantically identical to its source. DuckDB
+	// sets non-default flags (root struct non-nullable, map keys/entries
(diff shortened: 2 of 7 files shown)


lnkuiper:
I don't have much knowledge about Arrow, but I've done a quick pass over this with Codex. It found two issues, just want to run them by you to see if they are worth blocking the PR over:
>   1. `src/common/arrow/arrow_query_result.cpp:64` uses `ArrowSchemaDeepCopy`, but that helper does not preserve `ArrowSchema::flags`. In `src/common/adbc/nanoarrow/schema.cpp:421` it initializes every copied schema with `ARROW_FLAG_NULLABLE` and then copies format/name/metadata/children, never `schema_out->flags = schema->flags`. DuckDB schemas explicitly set root flags to 0 and map entries / map keys to non-nullable in `src/common/arrow/arrow_converter.cpp:102`. So `ArrowQueryResult::GetSchema()` can return a semantically different schema, especially for `MAP`. I’d fix the recursive deep copy to copy flags, and add a small cached-schema test for a map column.
>
>  2. Open question: `src/common/arrow/arrow_wrapper.cpp:67` still rebuilds schemas with `ArrowConverter::ToArrowSchema`. If any downstream path wraps an `ArrowQueryResult` in this stream wrapper after auto-commit, the GEOMETRY/CRS failure remains. I didn’t find an in-repo call site, so this may be intentionally handled in duckdb-python, but it’s worth confirming before merge.

src/common/arrow/physical_arrow_batch_collector.cpp
```diff
@@ -18,16 +18,21 @@ SinkFinalizeType PhysicalArrowBatchCollector::Finalize(Pipeline &pipeline, Event
 	auto total_tuple_count = gstate.data.Count();
 	if (total_tuple_count == 0) {
 		// Create the result containing a single empty result conversion
-		gstate.result = make_uniq<ArrowQueryResult>(statement_type, properties, names, types,
-		                                            context.GetClientProperties(), record_batch_size);
+		auto result = make_uniq<ArrowQueryResult>(statement_type, properties, names, types,
+		                                          context.GetClientProperties(), record_batch_size);
+		result->BuildCachedSchema();
```
carlopi, line 23: Should `BuildCachedSchema` gated by a ClientContext option? So that then there is a path to enable and disable this behaviour? Mostly thinking if there are clients for which the previous behaviour would make more sense (I am not sure how).

If this makes sense, a setting needs to be added and all callsites could be gated.

evertlammerts:
> > 1. `src/common/arrow/arrow_query_result.cpp:64` uses `ArrowSchemaDeepCopy`, but that helper does not preserve `ArrowSchema::flags`. In `src/common/adbc/nanoarrow/schema.cpp:421` it initializes every copied schema with `ARROW_FLAG_NULLABLE` and then copies format/name/metadata/children, never `schema_out->flags = schema->flags`. DuckDB schemas explicitly set root flags to 0 and map entries / map keys to non-nullable in `src/common/arrow/arrow_converter.cpp:102`. So `ArrowQueryResult::GetSchema()` can return a semantically different schema, especially for `MAP`. I’d fix the recursive deep copy to copy flags, and add a small cached-schema test for a map column.

That's a nice catch. `ArrowSchemaDeepCopy` was already used by single_batch_array_stream.cpp, which is used by our adbc driver. This PR would have also introduced it on the query result. I've added the flags now as well.

> > 2. Open question: `src/common/arrow/arrow_wrapper.cpp:67` still rebuilds schemas with `ArrowConverter::ToArrowSchema`. If any downstream path wraps an `ArrowQueryResult` in this stream wrapper after auto-commit, the GEOMETRY/CRS failure remains. I didn’t find an in-repo call site, so this may be intentionally handled in duckdb-python, but it’s worth confirming before merge.

`ArrowQueryResult` can not be wrapped in a `ResultArrowArrayStreamWrapper`, but it's true that the bug can still happen through there. In Python we have a custom wrapper that calls (will call) `HasCachedSchema`/`GetSchema`. I think the only place that consumes that path is the C API right now. We should probably follow up at some point to fix that as well.

src/common/arrow/physical_arrow_batch_collector.cpp
evertlammerts: I don't think we need to put this behind a flag. `BuildCachedSchema` doesn't do much real work and the code paths are already used. The deepcopy by the adbc driver and the schema construction by `ToArrowSchema`. The real difference is that it does it during `Finalize`, i.e. upfront. I don't really think it's a very risky change.

evertlammerts:
@Mytherin I think we can put this in v1.5.4 without much risk, and it does address a bug we've had a few reports about. Wdyt?

carlopi:
I think I understood (from @evertlammerts) that this might not be required, and a different solution might be on its way.

evertlammerts:
Yes, let's skip this for now. I'm taking a different approach in python.

PR #23179, opened 2026-06-09
Allow `MERGE INTO` in triggers

Dtenwolde:
This PR allows the `MERGE INTO` to be used within `CREATE TRIGGER`, which previously only allowed `UPDATE`, `DELETE`, or `INSERT`.


Mytherin:
Thanks! Looks great - can we add some tests as well?

Mytherin:
Thanks!

PR #23183, opened 2026-06-09
aggregate_export: reject importing states with custom destructors

bendrucker:
Rejects importing aggregate states for functions whose state owns heap memory, closing a heap-corruption hole in the aggregate-state import path.

## Issue

A single query aborts the process with an invalid free (SIGABRT, exit 134):

```sql
SELECT finalize(combine(
  to_aggregate_state({'value':'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa','isset':true}, 'min', ['VARCHAR']),
  to_aggregate_state({'value':'zzzzzzzzzzzzz','isset':true}, 'min', ['VARCHAR'])
));
```

Some aggregate states own a heap string freed with `delete[] value.GetData()` in a custom destructor. That destructor assumes `value` was allocated with `new char[]`. `min`/`max` over `VARCHAR` use `MinMaxStringState`. `last` over `VARCHAR` uses `FirstFunctionStringBase` with `UnaryAggregateDestructor`.

The aggregate-state import functions (`to_aggregate_state`, `finalize`, `combine`, `combine_aggr`) deserialize an input vector's `string_t` directly into the state buffer in `DeserializeState`. That pointer is owned by the input vector's arena-backed `StringHeap`. The custom destructor assumes a `new char[]` allocation. `combine` then reallocates and calls the destructor, which runs `delete[]` on the arena-owned pointer. The abort surfaces later in `StringHeap::~StringHeap` / `ArenaAllocator`.

`EXPORT_STATE` already refuses such functions. `ExportAggregateFunction::Bind` throws a `BinderException` when `bound_function.HasStateDestructorCallback()`. The import binders did not apply that restriction.

## Affected Functions

Any aggregate whose state has a custom destructor was reachable through the import path and could corrupt the heap. This includes:

- `min`/`max` over `VARCHAR`/`BLOB`
- `last` over `VARCHAR`/`BLOB`/nested types
- `arg_min`/`arg_max` over strings and other self-owning state aggregates

`first` and `any_value` were not affected. Their state has no custom destructor and their combine copies the `string_t` without freeing. They remain fully functional.

## Changes

Adds the same guard to `BindExportedAggregate` in `src/function/scalar/system/aggregate_export.cpp`. This is the single chokepoint that re-binds the aggregate for every import entry point (`ToAggregateStateBind`, `BindAggregateState`, `CombineAggrBind` all route through it). After the aggregate is re-bound, the guard rejects any function with a custom state destructor:

```
Binder Error: Cannot import/finalize/combine aggregate state for function min with a custom destructor
```

Legitimate exportable aggregates (`sum`, `count`, `avg`, numeric `min`/`max`, `first`, `any_value`, etc.) have no custom destructor and are unaffected.

## Testing

Adds `test/sql/aggregate/aggregate_state_export/string_state_import_no_destructor.test`:

- Asserts string `min`, `max`, and `last` imports raise the new binder error, in both the `to_aggregate_state(...)` and `finalize(combine(...))` forms.
- Asserts the original crashing `min` roundtrip now errors instead of aborting.
- Confirms `first` and `any_value` over `VARCHAR` still import and finalize correctly (single state and combined), proving the guard is not over-broad.
- Confirms numeric `min`/`max` state still roundtrips through `EXPORT_STATE` + `finalize`/`combine`.

The existing `test/sql/aggregate/aggregate_state_export/*` suite continues to cover the unaffected exportable aggregates.


Mytherin:
Thanks, but this code is under such active construction right now that these kinds of patches are not very helpful. This will need to work anyway so this contribution will be removed again within a few days.

bendrucker:
No worries, appreciate the look!

PR #23185, opened 2026-06-10
Add invoke scalar function

feichai0017:
Split out from #23024

This PR reintroduces the generic `invoke(lambda, args...)` scalar function as a standalone change. The simple CASE rewrite is intentionally left out and can be revisited separately after `invoke` lands.

Changes:
- add `invoke` registration and implementation
- support lambda-first scalar binding for `invoke` while keeping existing list lambda functions in their current `(list, lambda)` shape
- generalize lambda bind data access used by foldability checks
- add sqllogictest coverage for nested invokes, multiple arguments, captures, NULL handling, non-SELECT use, restart persistence, and binder errors


Maxxen:
Okay, sorry for the delay, this looks good to me! Good approach to move the argument reordering into the invoke function internals to avoid breaking BWC with the other lambda functions.

Good to go from my side CC @Mytherin

Maxxen (approved):

Mytherin:
Thanks!

PR #23187, opened 2026-06-10
Step 3: Make duplicate eliminator optimizer logic work for CTE rewrite

kryonix:
This PR implements a duplicate eliminator optimizer (deliminator) equivalent for the CTE-based rewrite of `DELIM_JOIN`/`DELIM_SCAN`. This runs immediately after flattening, but before the optimizer. However, I have hooked into the same optimizer hooks to provide both, profiling information, and the possibility to disable the pass.

Also, now with the duplicate eliminator optimizer logic, the CTE rewrite is quite a bit of code. So I moved everything into it's own dedicated file.


Tishj:
Not deliminator, duplicate eliminator
Can you change the title + description for future reference?

kryonix:
I changed it, but left a reference in. The optimizer pass I've reimplemented is called deliminator.

Tishj:
Thanks for making the changes, though I am of the opinion that `Deliminator` is a misnomer that we shouldn't perpetuate if we have the opportunity

lnkuiper:
We will be removing the deliminator/delim join for v2.0.0 if we get through all the steps for this so we should hopefully not have any confusion about it in the future. Going to give this a review today

src/planner/subquery/delim_join_cte_rewriter.cpp
```diff
@@ -0,0 +1,1907 @@
...
+                                                                  idx_t depth) const {
+	if (op->type == LogicalOperatorType::LOGICAL_DELIM_JOIN) {
+		if (!op->children.empty()) {
+			FindJoinsWithGeneratedDedupRefs(op->children[0], joins, depth + 1);
+		}
+		return;
+	}
+
+	for (auto &child : op->children) {
+		FindJoinsWithGeneratedDedupRefs(child, joins, depth);
```
lnkuiper: This doesn't increment depth as it goes through the tree, while the deliminator does Is that on purpose?

lnkuiper (approved):
Thanks for the PR! Just one comment below. Not blocking, can be fixed in a follow-up if necessary.

lnkuiper:
Ah, just saw the merge conflict. Maybe we can fix the comment and the conflict in one go :)

src/planner/subquery/delim_join_cte_rewriter.cpp
kryonix: Good catch. I put it in now with the rebase.

Mytherin:
Thanks!

PR #23188, opened 2026-06-10
Add support for RECORDS option for COPY FROM

### LINKED ISSUE #23118: Unexpected error `Invalid Input Error` for COPY FROM
### What happens?

Hi,

In the following test case, read_json executes successfully from a valid JSON file, but COPY FROM triggers an error.

First, write a JSON file `/tmp/test.json` with `[1,2,3]`. Then execute the following query:

```
SELECT *
         FROM read_json(
             '/tmp/test.json',
             format='array',
             records='false',
             columns={j: 'JSON'}
         );
┌──────┐
│  j   │
│ json │
├──────┤
│ 1    │
│ 2    │
│ 3    │
└──────┘


CREATE TABLE t(j JSON);
COPY t FROM '/tmp/test.json' (FORMAT JSON, ARRAY TRUE);
Invalid Input Error:
JSON transform error in file "/tmp/test.json", in record/value 1: Expected OBJECT, but got UBIGINT: 1
Try setting 'auto_detect' to true, specifying 'format' or 'records' manually, or setting 'ignore_errors' to true.
```

### To Reproduce

First, write a JSON file `/tmp/test.json` with `[1,2,3]`. Then execute the following query:

```
SELECT *
         FROM read_json(
             '/tmp/test.json',
             format='array',
             records='false',
             columns={j: 'JSON'}
         );

CREATE TABLE t(j JSON);
COPY t FROM '/tmp/test.json' (FORMAT JSON, ARRAY TRUE);
```

### OS:

Ubuntu 24.04

### DuckDB Version:

4dac4343d654d967afb9bc057c3b3638158cd00b

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Chi Zhang

### Affiliation:

Tsinghua University

- [x] Yes, I have

- [x] Yes, I have

Yes

ajaypanthagani:
Fixes #23118

The error message thrown when using `COPY FROM` without `RECORDS FALSE` suggests manually specifying the `format` or `records` option.

The expected query is:

```sql id="q7m4ra"
COPY t FROM '/tmp/test.json' (FORMAT JSON, ARRAY TRUE, RECORDS FALSE);
```

However, this currently fails with:

```text id="3p9vke"
Not implemented Error:
Unsupported option for COPY FROM: records
```

### Root cause

`ParseCopyOption` (the `COPY FROM` path) does not handle the `records` option. Additionally, `InitializeOptions` hard-codes `record_type = RECORDS` (JSON object mode) with no way to override it from `COPY`.

The JSON execution path already supports scalar (`VALUES`) mode correctly via `record_type`; the option simply was not exposed through the `COPY` path.

### Solution

1. Add a `records` case in `ParseCopyOption` so `COPY FROM` can configure `record_type`, similar to `read_json`.
2. Add type handling for both:

   * `RECORDS FALSE` (BOOLEAN literal)
   * `RECORDS 'false'` (VARCHAR)

   since `COPY` option values are not automatically cast like table function parameters.
3. Add early validation in `FinalizeCopyBind` for the multi-column + `VALUES` combination, matching the existing validation already enforced by `BindReader` on the `read_json` path.

### Tests

```bash id="1ywx8h"
build/reldebug/test/unittest test/sql/json/copy_json_records_option.test
```

### DIFF
--- a/extension/json/json_multi_file_info.cpp
+++ b/extension/json/json_multi_file_info.cpp
@@ -256,6 +256,25 @@ bool JSONMultiFileInfo::ParseCopyOption(ClientContext &context, const string &ke
 		}
 		return true;
 	}
+	if (loption == "records") {
+		JSONCheckSingleParameter(key, values);
+		auto &val = values.back();
+		if (val.type().id() == LogicalTypeId::BOOLEAN) {
+			options.record_type = BooleanValue::Get(val) ? JSONRecordType::RECORDS : JSONRecordType::VALUES;
+		} else {
+			auto arg = StringUtil::Lower(StringValue::Get(val));
+			if (arg == "auto") {
+				options.record_type = JSONRecordType::AUTO_DETECT;
+			} else if (arg == "true") {
+				options.record_type = JSONRecordType::RECORDS;
+			} else if (arg == "false") {
+				options.record_type = JSONRecordType::VALUES;
+			} else {
+				throw BinderException("COPY (FORMAT JSON) \"records\" parameter must be one of ['auto', 'true', 'false'].");
+			}
+		}
+		return true;
+	}
 	return false;
 }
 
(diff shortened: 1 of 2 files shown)


ajaypanthagani:
@Mytherin happy to make any adjustments if needed.

ajaypanthagani:
Gentle bump on this, anything I can do to make this PR easier to review? @Mytherin

Issue #23184, opened 2026-06-10
Internal Error with CAST in WHERE clause


egan8888:
### What happens?

Query using CAST in a WHERE clause leads to an internal error.

LLM-generated RCA:
```
Crash site: RowGroupReorderer::GetRootSegment
  auto stats = row_group.GetNode().GetStatistics(options.column_idx);
  if (IsNullOnly(*stats))   // ← stats is nullptr → crash

Root cause:
  When ORDER BY contains CAST(struct_col.field AS DOUBLE) and the source
  field type cannot propagate statistics to the target type,
  StructStats::PushdownExtract returns nullptr instead of valid stats.
  The RowGroupReorderer dereferences it unconditionally.

  VARCHAR's PhysicalType is absent from the signed-int/float switch in
  CanPropagateCast, so TryPropagateCast returns nullptr → crash.

Trigger condition:
  The RowGroupPruner only fires when the ORDER BY expression is a
  BOUND_COLUMN_REF.  Selecting a column other than the struct being cast
  forces the binder to project CAST(acv.amount AS DOUBLE) as a hidden
  column, which the pruner traces back to the struct scan.
```

### To Reproduce

```python
import duckdb

con = duckdb.connect()
con.execute("CREATE TABLE t (id INTEGER, acv STRUCT(amount VARCHAR, currency VARCHAR))")
con.execute("INSERT INTO t VALUES (1, {'amount': '42', 'currency': 'USD'})")
con.execute("SELECT id FROM t ORDER BY CAST(acv.amount AS DOUBLE) DESC LIMIT 10").fetchall()
```

Result:

```
_duckdb.InternalException: INTERNAL Error: Attempted to dereference unique_ptr that is NULL!

Stack Trace:

0        _ZN6duckdb9Exception6ToJSONENS_13ExceptionTypeERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEE + 48
1        _ZN6duckdb17InternalExceptionC1ERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEE + 32
2        PyInit__duckdb + 301636
3        _ZN6duckdb17RowGroupReorderer14GetRootSegmentERNS_19RowGroupSegmentTreeE + 192
4        _ZN6duckdb18RowGroupCollection22InitializeParallelScanERNS_27ParallelCollectionScanStateE + 140
[...]

PR #23189, opened 2026-06-10
Fix: guard againt null row group reorder stats

DinosL:
This PR adds a guard against missing statistics in row group reorderer

Fixes:https://github.com/duckdb/duckdb/issues/23184


Mytherin:
Thanks!

PR #23190, opened 2026-06-10
Fix selection vector use in Arrow extension callbacks

### LINKED ISSUE #22444: BOOLEAN child of a UNION written bit-packed but exported as arrow.bool8 (byte-packed) under arrow_lossless_conversion

evertlammerts:
Replaces #22445 / fixes #22444

If arrow extensions are used and nested types contain values that should use the extension, we currently do not correctly resolve the extension for those children. E.g. nested BOOLEAN children were written bit-packed even if the Arrow schema declared them as arrow.bool8 (byte-packed) when using `arrow_lossless_conversion`.

In this PR, `InitializeChild` auto-resolves the Arrow extension for child types. `ArrowAppendData::AppendChild` runs the `duckdb_to_arrow` conversion, and all container appenders route through it. The difference with #22445 is that instead of flattening input vectors before the callback, we fixed `ArrowBool8::DuckToArrow` to use the selection vector. This makes sure that e.g. constant vectors are not needlessly expanded. On `ArrowToDuck` I've added a `D_ASSERT` as documentation for the input vector always being flat (which it always should be). The converted vector is also sized to the actual child count to avoid an OOB write on large lists.

### DIFF
--- a/src/common/arrow/appender/append_data.cpp
+++ b/src/common/arrow/appender/append_data.cpp
@@ -26,4 +26,18 @@ void ArrowAppendData::AppendValidity(UnifiedVectorFormat &format, idx_t from, id
 	}
 }
 
+void ArrowAppendData::AppendChild(Vector &input, idx_t from, idx_t to, idx_t input_size) {
+	if (extension_data && extension_data->duckdb_to_arrow) {
+		// Convert the DuckDB-typed input into the extension's internal Arrow type before
+		// handing it to the (internal-typed) child appender. Size the internal vector to the
+		// actual input_size: container children can exceed STANDARD_VECTOR_SIZE (e.g. a 2048-row
+		// LIST whose elements average two entries), and duckdb_to_arrow writes input_size values.
+		Vector internal(extension_data->GetInternalType(), MaxValue<idx_t>(input_size, STANDARD_VECTOR_SIZE));
+		extension_data->duckdb_to_arrow(*options.client_context, input, internal, input_size);
+		append_vector(*this, internal, from, to, input_size);
+	} else {
+		append_vector(*this, input, from, to, input_size);
+	}
+}
+
 } // namespace duckdb
--- a/src/common/arrow/appender/fixed_size_list_data.cpp
+++ b/src/common/arrow/appender/fixed_size_list_data.cpp
@@ -23,7 +23,7 @@ void ArrowFixedSizeListData::Append(ArrowAppendData &append_data, Vector &input,
 	auto array_size = ArrayType::GetSize(input.GetType());
 	auto &child_vector = ArrayVector::GetEntry(input);
 	auto &child_data = *append_data.child_data[0];
-	child_data.append_vector(child_data, child_vector, from * array_size, to * array_size, size * array_size);
+	child_data.AppendChild(child_vector, from * array_size, to * array_size, size * array_size);
 	append_data.row_count += size;
 }
 
--- a/src/common/arrow/appender/struct_data.cpp
+++ b/src/common/arrow/appender/struct_data.cpp
@@ -24,7 +24,7 @@ void ArrowStructData::Append(ArrowAppendData &append_data, Vector &input, idx_t 
 	for (idx_t child_idx = 0; child_idx < children.size(); child_idx++) {
 		auto &child = children[child_idx];
 		auto &child_data = *append_data.child_data[child_idx];
-		child_data.append_vector(child_data, *child, from, to, size);
+		child_data.AppendChild(*child, from, to, input_size);
 	}
 	append_data.row_count += size;
 }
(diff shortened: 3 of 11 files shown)


Mytherin:
Thanks!

PR #23191, opened 2026-06-10
Support submodules in extensions patching process

staticlibs:
When a build is re-run in the same build tree, the out-of-tree extensions patching process may be invoked multiple times.

This PR enhances the cleanup step of the patching script to also clean the submodules before applying the pathes. This should fix the problem with patching encountered in #23161.


carlopi (approved):

Mytherin:
Thanks!

PR #23193, opened 2026-06-10
Replace ARTConflictType::TRANSACTION with fatal exception

artjomPlaunov:
There is a possible scenario where we can have duplicates in unique ART indexes: if we do a delete + insert on the same key, this turns into the reverse during commit: insert into the main index (creates temporary duplicate state) -> then delete the old rowid (restores back to non-duplicate state). 

However, in ARTOperator::Insert It shouldn't be possible to reach this temporary gated duplicate state due to transaction + wal locks, i.e. this should now be unreachable, unless the ART is corrupted and hence we now throw a fatal exception at this point. 

Note that VerifyLeaf could still hit the duplicate leaves unique index scenario, so the logic there remains.


lnkuiper:
Hey, looks good - just checking: is `ARTConflictType::TRANSACTION` unused now?

artjomPlaunov:
@lnkuiper ah yea, I thought this was the only location, but there is still a D_ASSERT, as well as a test I forgot to modify that checks for it (although AFAIK, it wasn't actually hitting that case anymore). Let me double check these and will ping you again for review.

artjomPlaunov:
@lnkuiper but yes I think ARTConflictType::TRANSACTION should be unused now, so I also have to update the enum, etc., let me handle that now :-)

lnkuiper:
Great, feel free to mark as ready to merge once you've done that :)

Mytherin:
Thanks!

PR #23194, opened 2026-06-10
[Dev] Fix variant shredding analysis logic discrepancy with shredded writing

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9069

During analysis fields were being treated as case-insensitive, but during writing we correctly differentiate the two.
This discrepancy could cause us to hit an internal exception


Mytherin:
Thanks!

Issue #23149, opened 2026-06-09
VARIANT changes the json_type of value


DerZc:
### What happens?

Hi,

I found that the VARIANT sometimes changes the json_type of value, this is an example:

```
WITH src(j) AS (
    VALUES ('{"n":123456789012345678901234567890}'::JSON)
)
SELECT
    j AS original_json,
    json_type(j, '$.n') AS original_type,
    (j::VARIANT)::JSON AS roundtrip_json,
    json_type((j::VARIANT)::JSON, '$.n') AS roundtrip_type
FROM src;

┌──────────────────────────────────────┬───────────────┬────────────────────────────────────────┬────────────────┐
│            original_json             │ original_type │             roundtrip_json             │ roundtrip_type │
│                 json                 │    varchar    │                  json                  │    varchar     │
├──────────────────────────────────────┼───────────────┼────────────────────────────────────────┼────────────────┤
│ {"n":123456789012345678901234567890} │ DOUBLE        │ {"n":"123456789012345678901234567890"} │ VARCHAR        │
└──────────────────────────────────────┴───────────────┴────────────────────────────────────────┴────────────────┘
```

### To Reproduce

```
WITH src(j) AS (
    VALUES ('{"n":123456789012345678901234567890}'::JSON)
)
SELECT
    j AS original_json,
    json_type(j, '$.n') AS original_type,
    (j::VARIANT)::JSON AS roundtrip_json,
    json_type((j::VARIANT)::JSON, '$.n') AS roundtrip_type
FROM src;
```

### OS:

Ubuntu 24.04

### DuckDB Version:

d7b51fbc7d2cc1ba81847d6ecfc25a524e2247ab

### DuckDB Client:

CLI

### Full Name:

Chi Zhang

### Affiliation:

Tsinghua University

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

DinosL:
Fixed in https://github.com/duckdb/duckdb/pull/23195

Issue #23122, opened 2026-06-08
Data type lost in PARQUET file


DerZc:
### What happens?

Hi,

I found that I saved a BLOB value to a PARQUET file, but read the VARCHAR data from the original value.

```
COPY (
    SELECT '\x00\x01ABC'::BLOB::VARIANT AS v
) TO '/tmp/test.parquet' (
    FORMAT PARQUET
);

SELECT
    variant_typeof('\x00\x01ABC'::BLOB::VARIANT) AS original_type,
    hex('\x00\x01ABC'::BLOB) AS original_blob_hex,
    variant_typeof(v) AS reread_type,
    v::VARCHAR AS reread_text,
    hex(v::BLOB) AS reread_blob_hex
FROM read_parquet('/tmp/test.parquet');
┌───────────────┬───────────────────┬─────────────┬─────────────┬──────────────────┐
│ original_type │ original_blob_hex │ reread_type │ reread_text │ reread_blob_hex  │
│    varchar    │      varchar      │   varchar   │   varchar   │     varchar      │
├───────────────┼───────────────────┼─────────────┼─────────────┼──────────────────┤
│ BLOB          │ 0001414243        │ VARCHAR     │ AAFBQkM=    │ 41414642516B4D3D │
└───────────────┴───────────────────┴─────────────┴─────────────┴──────────────────┘
```

### To Reproduce

```
COPY (
    SELECT '\x00\x01ABC'::BLOB::VARIANT AS v
) TO '/tmp/test.parquet' (
    FORMAT PARQUET
);

SELECT
    variant_typeof('\x00\x01ABC'::BLOB::VARIANT) AS original_type,
    hex('\x00\x01ABC'::BLOB) AS original_blob_hex,
    variant_typeof(v) AS reread_type,
    v::VARCHAR AS reread_text,
    hex(v::BLOB) AS reread_blob_hex
FROM read_parquet('/tmp/test.parquet');
```

### OS:

Ubuntu 24.04

### DuckDB Version:

4dac4343d654d967afb9bc057c3b3638158cd00b

### DuckDB Client:

CLI

### Full Name:

Chi Zhang

### Affiliation:

Tsinghua University

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

DinosL:
[...]

PR #23195, opened 2026-06-10
Variant fixes

DinosL:
Fixes: https://github.com/duckdb/duckdb/issues/23149 : Large number as json was written as `RAW` by `yyjson`, which we store as `VARCHAR`, so converting from `JSON` to `VARIANT` and back to `JSON` turned the number into a string.

patch by @Tishj in https://github.com/duckdblabs/duckdb-internal/issues/9628#issuecomment-4660002550

Fixes: https://github.com/duckdb/duckdb/issues/23122 :  A `BLOB` stored inside a `VARIANT` was decoded early to `base64`, turning it into a `VARCHAR`


extension/parquet/reader/variant/variant_binary_decoder.cpp
```diff
@@ -220,12 +217,9 @@ VariantValue VariantBinaryDecoder::PrimitiveTypeDecode(const VariantValueMetadat
...
 	}
 	case VariantPrimitiveType::BINARY: {
-		//! Follow the JSON serialization guide by converting BINARY to Base64:
-		//! For example: `"dmFyaWFudAo="`
 		auto size = Load<uint32_t>(data);
-		auto string_data = reinterpret_cast<const char *>(data + sizeof(uint32_t));
-		auto base64_string = Blob::ToBase64(string_t(string_data, size));
-		return VariantValue(Value(base64_string));
+		auto string_data = data + sizeof(uint32_t);
+		return VariantValue(Value::BLOB(string_data, size));
```
Tishj: This depends on what they serialize, but this might need to be `BLOB_RAW(...)` instead

extension/parquet/reader/variant/variant_shredded_conversion.cpp
```diff
@@ -107,7 +107,9 @@ VariantValue ConvertShreddedValue<timestamp_ns_t>::Convert(timestamp_ns_t val) {
 //! binary
 template <>
 VariantValue ConvertShreddedValue<string_t>::ConvertBlob(string_t val) {
-	return VariantValue(Value(Blob::ToBase64(val)));
+	//! Keep the raw bytes as a BLOB so the type is preserved when reconstructing a VARIANT. The conversion to Base64
+	//! happens now in VariantValue::ToJSON.
+	return VariantValue(Value::BLOB(const_data_ptr_cast(val.GetData()), val.GetSize()));
```
Tishj, line 111: Same as above, maybe `BLOB_RAW(...)`

test/parquet/variant/variant_basic.test
```diff
@@ -53,7 +53,7 @@ from '{DATA_DIR}/parquet-testing/variant_int32.parquet';
 query II
 from '{DATA_DIR}/parquet-testing/variant_binary.parquet';
 ----
-1	CgsMDQ==
+1	\x0A\x0B\x0C\x0D
```
Tishj, line 56: Can we add another test where we select `octet_length(var)` and verify that it's 4 ?
If that passes, you can disregard all the other comments
If not, then those other comments might help solve this

Tishj: Actually nvm, I made a build and checked:
```sql
query III
select *, octet_length(var::BLOB) from '{DATA_DIR}/parquet-testing/variant_binary.parquet';
----
1	\x0A\x0B\x0C\x0D	4
```

This already passes 👍

DinosL: I see there is a merge conflict, so I will add this test as well

lnkuiper:
@DinosL Can you check if the failing CI is related?

lnkuiper:
Actually, `test/sql/join/semianti/right_anti.test` seems to time out. Probably unrelated. I'll have a look

DinosL:
I think the issue is `test/sql/join/full_outer/test_full_outer_join_complex.test`. I am trying to figure out why it times out

DinosL:
It's not reproducing locally, so I think we should be ok to rerun.

lnkuiper:
I think it will time out again. I've opened a PR here: https://github.com/duckdb/duckdb/pull/23224

PR #23050, opened 2026-06-03
Support scalar-wrapped aggregate macros in window

mach-kernel:
##### Problem

Resolves #23049: a macro that has an aggregate function in its body mixed with other scalar ops is rejected, but would otherwise be able to run if the window spec was pushed down. `json_group_object` is an example of this:

```sql
CREATE TABLE fruit(person VARCHAR, fruit VARCHAR, qty INT);
INSERT INTO fruit VALUES
    ('alice', 'apple',  3),
    ('alice', 'banana', 2),
    ('bob',   'cherry', 5);
SELECT person, json_group_object(fruit, qty) OVER (PARTITION BY person) FROM fruit;
Binder Error:
Window function macros must be functions

----------------------------------------------------------------------------------------------------------------------------------------------------------------

D select person, CAST('{' || string_agg(to_json(fruit::VARCHAR) || ':' || CASE WHEN qty IS NULL THEN 'null'::JSON ELSE to_json(qty) END, ',') OVER (PARTITION BY person) || '}' AS JSON) from fruit;
┌─────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ person  │ CAST((('{' || string_agg(((to_json(CAST(fruit AS VARCHAR)) || ':') || CASE  WHEN ((qty IS NULL)) THEN (CAST('null' AS "JSON")) ELSE to_json(qty) END), ',') OVER (PARTITION BY person)) || '}') AS "JSON") │
│ varchar │                                                                                                    json                                                                                                    │
├─────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ bob     │ {"cherry":5}                                                                                                                                                                                               │
│ alice   │ {"apple":3,"banana":2}                                                                                                                                                                                     │
│ alice   │ {"apple":3,"banana":2}                                                                                                                                                                                     │
└─────────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

This is desirable, e.g. Snowflake [`OBJECT_AGG`](https://docs.snowflake.com/en/sql-reference/functions/object_agg) works in window. 

##### Changes
- Walk macro expr and collect all aggregate fns (where we validate they are agg fns via catalog)
  - Error unless there is exactly one aggregate function in the macro body
[...]

PR #21963, opened 2026-04-08
Issue #21908: Window Macro Expansion 

hawkfish:
* Don't drop FILTER clauses
* Don't drop DISTINCT tags
* Combine multiple FILTER clauses
* Don't expand LIST(...ORDER BY) when parsing a macro (so it can be used in windowing)
* Transfer argument ORDER BYs to the window's arg_order_bys.
* Update window macro error messages.

PR #23198, opened 2026-06-10
[v1.5] Backport scalar-wrapped agg macros in window

mach-kernel:
Backports #23050 to v1.5-variegata. 

Does not include bigger changes from #21963.

### DIFF
--- a/src/planner/binder/expression/bind_macro_expression.cpp
+++ b/src/planner/binder/expression/bind_macro_expression.cpp
@@ -1,5 +1,8 @@
 #include "duckdb/catalog/catalog_entry/scalar_macro_catalog_entry.hpp"
+#include "duckdb/catalog/entry_lookup_info.hpp"
 #include "duckdb/common/enums/expression_type.hpp"
+#include "duckdb/common/enums/on_entry_not_found.hpp"
+#include "duckdb/common/exception/binder_exception.hpp"
 #include "duckdb/function/scalar_macro_function.hpp"
 #include "duckdb/parser/expression/function_expression.hpp"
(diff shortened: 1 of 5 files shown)


src/planner/binder/expression/bind_macro_expression.cpp
```diff
@@ -91,6 +94,55 @@ void ExpressionBinder::ReplaceMacroParameters(unique_ptr<ParsedExpression> &expr
...
+
+	// Transfer the macro function attributes
+	auto &window_expr = expr->Cast<WindowExpression>();
+	window_expr.catalog = agg_fn_expr.catalog;
+	window_expr.schema = agg_fn_expr.schema;
+	window_expr.function_name = agg_fn_expr.function_name;
+	window_expr.children = std::move(agg_fn_expr.children);
+	window_expr.distinct = agg_fn_expr.distinct;
+	window_expr.filter_expr = std::move(agg_fn_expr.filter);
+	// TODO: transfer order_bys when window functions support them
```
hawkfish: That is the `arg_orders` member.

hawkfish (changes requested):
Window functions already support ordered arguments in 1.5 (and earlier), so can we port that transfer as well?

src/planner/binder/expression/bind_macro_expression.cpp
mach-kernel: I copied the order bys and added an SLT

hawkfish (approved):
LGTM!

Mytherin:
Thanks but after some discussion we would consider this a feature, not a bug fix, so it should not be backported to a bug-fix release

PR #23201, opened 2026-06-10
Zero the inlined buffer in string_t's length-only constructor

krlmlr:
This is an error that surfaced when checking the R package with valgrind. The fix looks consistent with the more specific constructor just below the change, I'm not sure about the implications.

Claude writes:

StringVector::EmptyString hands out a string_t whose inlined buffer stays indeterminate until the caller writes len bytes and Finalize() zeroes the tail. The 16-byte value is bitwise-copied before that happens, so valgrind tracks the indeterminate bytes into StringValueComparison. Zero the buffer up front, matching what the (const char *, uint32_t) constructor and Finalize() already do.


Mytherin:
Thanks - can we just remove the comment?

PR #23202, opened 2026-06-10
Initialize all `BaseStatistics` members and zero `stats_union`

krlmlr:
I've seen Valgrind complaining when the move constructor touches fields which are not initialized by default constructors. I can't tell whether this problem is "harmless by design" or whether we should really initialize and rely on the compiler optimizing it out where it matters.

Claude writes:

The default constructor and `Construct()` left `has_null`, `has_no_null` and `stats_union` (including its padding bytes) indeterminate; copies made during statistics propagation in the optimizer can read those bytes before `CreateEmpty`/`CreateUnknown` overwrite them, which valgrind flags in `StringValueComparison`. The new values match `InitializeEmpty()`, so behaviour is unchanged.

### DIFF
--- a/src/storage/statistics/base_statistics.cpp
+++ b/src/storage/statistics/base_statistics.cpp
@@ -12,7 +12,8 @@
 
 namespace duckdb {
 
-BaseStatistics::BaseStatistics() : type(LogicalType::INVALID) {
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks!

PR #23204, opened 2026-06-10
Out-of-line `SelectionData` destructor to silence g++-16 `-Warray-bounds`

krlmlr:
From what I remember, the underlying issue is fairly old, and we shouldn't hold our breath. I understand very little about what's happening here, but can confirm that this change silences the warning (that is checked by CRAN).

Claude writes:

GCC 14+ emits a spurious array-bounds warning when IPA-ICF folds the `shared_ptr` dispose specialization of `SelectionData` with the structurally similar one of `TemplatedValidityData`. A user-declared out-of-line (defaulted) destructor keeps the disposal sequence non-inline at the call site and avoids the false positive; behaviour is unchanged.

### DIFF
--- a/src/common/types/selection_vector.cpp
+++ b/src/common/types/selection_vector.cpp
@@ -16,6 +16,8 @@ SelectionData::SelectionData(idx_t count) {
 #endif
 }
 
+SelectionData::~SelectionData() = default;
+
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #23205, opened 2026-06-10
parquet: initialize `ParquetReader::rows_read`

krlmlr:
Another Valgrind warning that went away after this change.

Question similar to #23203: do we support/use this new syntax, or should all initialization happen (perhaps redundantly) in the specific constructors?

Claude writes:

The `OpenFileInfo` constructor of `ParquetReader` does not list `rows_read` in its initializer list, leaving the atomic counter with uninitialized storage. Valgrind flags the read in `GetProgressInFile`, from where the taint propagates into the pipeline progress machinery.


Mytherin:
Thanks!

PR #23213, opened 2026-06-11
Fix VARIANT shredding, avoid including empty object keys

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9647

Because we don't support unnamed structs in the storage layer, we have to avoid empty keys in the shredding.
This also exposed another problem with our roundtrip verification.
If structs are partially shredded (both unshredded and shredded value), these have to be merged and the result is unpredictable.
To fix this, we add normalize to it.

As mentioned in a comment below, due to DuckDB STRUCTs being either named (no empty field names) or unnamed (unnamed tuple), partially named structs (1 or more fields are "") aren't possible and the other named fields will have their names stripped.

### DIFF
--- a/test/sql/storage/types/variant/variant_shredding_empty_keys.test
+++ b/test/sql/storage/types/variant/variant_shredding_empty_keys.test
@@ -1,0 +1,38 @@
+# name: test/sql/storage/types/variant/variant_shredding_empty_keys.test
+# group: [variant]
+
+require json
+
+load {TEST_DIR}/repro_eka.db readwrite v1.5.0
+
+statement ok
+SET variant_minimum_shredding_size=0;
+
+statement ok
+create table succeeds as SELECT '{"": 1, "x": {"y": "t"}}'::JSON::VARIANT;
+
+statement ok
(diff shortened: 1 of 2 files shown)


test/sql/storage/types/variant/variant_shredding_empty_keys.test
```diff
@@ -0,0 +1,38 @@
...
+
+restart
+
+statement ok
+CHECKPOINT
+
+query I
+select * from fails;
+----
+(world, hello)
```
Tishj, line 33: This is technically wrong, but has to do with unnamed structs being all-or-nothing in DuckDB. Partially unnamed structs aren't possible, so the named keys are disappearing.

lnkuiper:
I think this has the same timeout as https://github.com/duckdb/duckdb/pull/23195
Probably unrelated

lnkuiper:
I've opened a PR here https://github.com/duckdb/duckdb/pull/23224 to fix the timeout, should be unrelated

PR #23214, opened 2026-06-11
sqllogictest: Add `statement skip_test_on_error`, and macro `assert_true`

carlopi:
See https://github.com/duckdb/duckdb/pull/23214/changes/f9190f63beb82806412101cabc8064320ebcbdb3 and https://github.com/duckdb/duckdb/pull/23214/changes/defd34743faeaf088001c660d475279f31546c0b for the actual machinery, the rest is search-and-replace.

Idea is allowing a more generic form of `require`, that doesn't need to be registered ahead of time. Currently we have for example:
```
require notwindows
```
and they can all be expressed in a more powerful way as just SQL statements, like in this case:

```sql
statement skip_test_on_error
SELECT assert_true(platform NOT LIKE 'windows_%', 'requires a non-windows platform') FROM pragma_platform()
```

Idea is that statement is executed (possibly with replacement of variables), and if it throws an error, then the rest of the test is skipped.

Also, adding `assert_true` SQL macro, there is precedence in [Spark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.assert_true.html) and [Databricks](https://docs.databricks.com/aws/en/sql/language-manual/functions/assert_true)

Currently replaced platform related ones and `vector_size` related ones.
Core idea is less magic, more SQL.

Old require form are not retired from the code, to avoid churn extension side. But that might happen in a next iteration, let's see what others think of this.


Tishj:
I like it!

With one addition:

```sql
# If result doesn't match, also skip the test
statement require
SELECT 42;
----
42
```

carlopi:
> I like it!
> 
> With one addition:
> 
> ```sql
> # If result doesn't match, also skip the test
> statement require
> SELECT 42;
> ----
> 42
> ```

I was thinking of this, but then it would need to become `require query I`, that is a bit cumbersome I thought, but I can see that makes for simpler writing, unsure.

Mytherin:
Thanks - I think something like this is a good idea - maybe this format is more composable with what we already have in the test runner:

```sql
statement ok
set variable is_not_mingw = (SELECT platform NOT LIKE '%_mingw' FROM pragma_platform());

require <variable:is_not_mingw>
```

carlopi:
One feature explicitly here is using the fact that any statement might work, like:
```sql
statement require
LOAD httpfs;
```
is also valid, and that can't be composed as a select.

I have to think about what's the path forward.

carlopi:
One path forward, what about:

1. adding `assert_true` (there is precedence in [Spark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.assert_true.html) and [Databricks](https://docs.databricks.com/aws/en/sql/language-manual/functions/assert_true))
```diff
diff --git a/src/catalog/default/default_functions.cpp b/src/catalog/default/default_functions.cpp
index afe503e16f..d5324deb3c 100644
--- a/src/catalog/default/default_functions.cpp
+++ b/src/catalog/default/default_functions.cpp
@@ -126,6 +126,9 @@ static const DefaultMacro internal_macros[] = {
      "(x, n) AS CASE ((abs(x) * power(10, n+1)) % 10) WHEN 5 THEN round(x/2, n) * 2 ELSE round(x, n) END"},
     {DEFAULT_SCHEMA, "roundbankers", "(x, n) AS round_even(x, n)"},
     {DEFAULT_SCHEMA, "nullif", "(a, b) AS CASE WHEN a=b THEN NULL ELSE a END"},
+    {DEFAULT_SCHEMA, "assert_true",
+     "(condition) AS CASE WHEN condition THEN NULL ELSE error('Assertion failed') END, "
+     "(condition, message) AS CASE WHEN condition THEN NULL ELSE error('Assertion: ' || message) END"},
     {DEFAULT_SCHEMA, "list_append", "(l, e) AS list_concat(l, list_value(e))"},
     {DEFAULT_SCHEMA, "array_append", "(arr, el) AS list_append(arr, el)"},
     {DEFAULT_SCHEMA, "list_prepend", "(e, l) AS list_concat(list_value(e), l)"},
```
2. adding the new form `statement require` (or `require statement` ?) (as per first commit in the PR)
3. have the conditions like:
```sql
statement require
SELECT assert_true(current_setting('standard_vector_size')::INT >= 2048, 'requires vector_size>=2048');
```

Would this fly?

carlopi:
From @Mytherin:
> Maybe instead of `statement require` we can go `statement skip_test_on_error` or something more descriptive

And yes, it's better.

lnkuiper:
Looks good, but these silently return NULL instead of failing the assertion:
```sql
# name: test/temp.test
# description: Repro for assert_true with a NULL custom message
# group: [generic]

statement error
SELECT assert_true(false, NULL)
----
Assertion failed

statement error
SELECT assert_true(NULL, NULL)
----
Assertion failed
```
Maybe we should disallow `NULL` comments?

carlopi:
Thanks for spotting this, addressed!

Mytherin (commented):
I'm still not super happy with this - how about instead of replacing the existing `require` with these verbose flags (while we still need to keep the old `require` code around anyway) we add a new `require`, e.g.:

```
require statement
SELECT assert_true(platform NOT LIKE '%_mingw', 'requires a non-mingw platform') FROM pragma_platform()
```

Or if that's hard to add, this could be `require_statement`:

```
require_statement
SELECT assert_true(platform NOT LIKE '%_mingw', 'requires a non-mingw platform') FROM pragma_platform()
```

We can then leave the tests as-is (instead of adding a bunch of duplicated complex conditions in many tests) while still having the added functionality of this PR.

Mytherin:
If you want to keep the SQL-level checks, we can internally replace the current conditions (based on `#ifdef` etc) with a short-hand in the SQLLogic runner, e.g.:

```sql
unordered_map<string, string> require_map = {
   "notmingw", "SELECT assert_true(platform NOT LIKE '%_mingw', 'requires a non-mingw platform') FROM pragma_platform()",
    // ...
};
```

carlopi:
Thanks for the feedback.

I think maybe this has exploded in scope over the commit, I'll need a fresh look.

I think I would like to:
* add `assert_true` as a macro
* add some form of `require statement` or `require_statement` as escape hatch
* (in a separate PR) consider some form of SQL level `ASSERT 42;` that rewrites to `SELECT assert_true(42);`, that in my mind can be handy in making preconditions clear (like, this table needs to be empty, or what not)

Indeed touching all the test is excessive, I'll close the PR (also, currently in the merge queue) and start from scratch.

Thanks!

PR #23216, opened 2026-06-11
Fix in remote pushdown optimizer for query macro 

ccfelius:
In quack there exists a `query()` macro, which should be used like `<quack_catalog>.query()`. This works fine for single statement inputs, but fails for multiple statements. 

Currently, the `RemotePushdownOptimizer` intercepts `<quack_catalog>.query()` and serializes it back as a SQL string _before_ the binder gets a chance to expand the query macro into `quack_query_by_name()`. This leads to a CI failure in quack, because of an `Invalid Input Error: Expected a single SELECT statement` error. 

This PR should fix this by special casing the `query()` function in the remote pushdown optimizer. 

fixes https://github.com/duckdblabs/duckdb-internal/issues/9657


ccfelius:
@kryonix

Mytherin:
Thanks - but we should fix this in `quack`, not in core DuckDB. `quack` can use the overload `QuackCatalog::SupportsPushdown(TableRef &ref)` to indicate that it can push down a table function or not.

PR #23217, opened 2026-06-11
Re-fix fsync implementation

dentiny:
Closes https://github.com/duckdblabs/duckdb-internal/issues/9664

The issue description is correct, use leveldb's implementation as reference
- `HAVE_FULLFSYNC` is defined [here](https://github.com/google/leveldb/blob/7ee830d02b623e8ffe0b95d59a74db1e58da04c5/CMakeLists.txt#L53) and used [here](https://github.com/google/leveldb/blob/7ee830d02b623e8ffe0b95d59a74db1e58da04c5/util/env_posix.cc#L398)
- `HAVE_FDATASYNC` is defined [here](https://github.com/google/leveldb/blob/7ee830d02b623e8ffe0b95d59a74db1e58da04c5/CMakeLists.txt#L52) and used [here](https://github.com/google/leveldb/blob/7ee830d02b623e8ffe0b95d59a74db1e58da04c5/util/env_posix.cc#L408)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,12 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "SunOS")
   set(SUN TRUE)
 endif()
 
+include(CheckCXXSymbolExists)
+check_cxx_symbol_exists(fdatasync "unistd.h" DUCKDB_HAVE_FDATASYNC)
+if(DUCKDB_HAVE_FDATASYNC)
+  add_definitions(-DDUCKDB_HAVE_FDATASYNC=1)
+endif()
+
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - but given this was not enabled before I think this has the potential of causing unexpected performance regressions, especially for a bug-fix release that seems like it could cause upstream issues for users. Maybe we can instead move this behind a setting and disable it by default similar to what Postgres does?

dentiny:
> Maybe we can instead move this behind a setting and disable it by default similar to what Postgres does?

Let me investigate on that

dentiny:
I will revisit the PR after https://github.com/duckdblabs/duckdb-internal/issues/6517

PR #23218, opened 2026-06-11
Add row group count to parquet RETURN_STATS

pdet:


src/execution/operator/persistent/physical_copy_to_file.cpp
```diff
@@ -2566,6 +2566,9 @@ void PhysicalCopyToFile::ReturnStatistics(DataChunk &chunk, CopyToFileInfo &info
 
 	// partition_keys map(varchar, varchar)
 	chunk.data[5].Append(info.partition_keys);
+
+	// row group count BIGINT
+	chunk.data[6].Append(file_stats.row_group_count);
```
Mytherin: Maybe we can make this more extensible by making this `extra_info` - `MAP(VARCHAR, VARIANT)`?

Mytherin:
Thanks!

PR #23220, opened 2026-06-11
Don't serialize types in type-values

Maxxen:
This PR just makes a minor change to how the type in `Value::TYPE`s is stored. It used to be de/serialized from/to a blob, but there's not really any reason for it. Its way more common that we pass types in `Value`s than `Vector`s , and so storing the type in its non-serialized form instead saves a bunch of unnecessary de/serialization round-trips. Particularly when e.g. parsing type expressions and binding types, which resolves types as values through `ConstantExpression`s.

Type-values already has a special-case for de/serialization into storage, so this has no impact on BWC.


Maxxen:
Doesnt work cause type does not have its own physical type...

PR #8417, opened 2023-07-29
[ADBC] Finish ADBC spec compliance

Tishj:
This PR finishes our compliance with the ADBC spec as described here #7141 

Namely this PR adds:

### ADBC Methods

- ConnectionGetInfo
This method allows you to interrogate the connection to request specific types of information about it
- StatementBind
Through this method it's possible to bind an ArrowArray to a parameter in a prepared statement.
- StatementGetParameterSchema
Get the schema of the parameters present in a prepared statement.

### Notable changes

As part of the above changes, we have added portions of `nanoarrow` to the project.
This replaces our old `arrow.hpp` header that we previously used to declare these structures and defines.

The C-API has been updated to add `duckdb_bind_value` as a generic way to bind parameters, it takes a `duckdb_value` which already existed but was almost never used before.
Also the `duckdb_prepared_arrow_schema` method is added, this is used internally in StatementGetParameterSchema.

We have added the `adbc_driver_duckdb` module as part of the Python package.
This makes it possible to import and use the ADBC driver in python, with a similar approach to the SQLite version of this driver.

A lot of the ADBC methods have had their memory security improved, every pointer is checked before being accessed now.

Bug fix: The names of the ArrowSchema produced by DuckDB are now heap allocated, previously they were stack allocated which could cause issues.

PR #23230, opened 2026-06-11
fix(adbc): support the ADBC Statistics API

eitsupi:
Support the ADBC 1.1.0 Statistics API.

- `ConnectionGetStatistics`: returns the table-level `ROW_COUNT` statistic from `duckdb_tables().estimated_size`, as `float64` with `statistic_is_approximate = true`.
- `ConnectionGetStatisticNames`: returns an empty result (DuckDB has no driver-specific statistics).
- Also fixes the bundled nanoarrow's `ArrowSchemaDeepCopy` dropping `schema->flags` (nullability), aligning it with upstream.

The spec requires `statistic_value` to be a dense union, which DuckDB's Arrow export does not support, so the result batch is built manually with malloc'd buffers instead of via a SQL query.
This is needed to keep the bundled nanoarrow minimal per the existing policy (it lacks the array-builder APIs), but I'm happy to rework this on top of an expanded nanoarrow bundle if that is preferred.

Since `arrow_scan` cannot consume dense unions either, the C++ tests inspect the result buffers directly, and I additionally verified the built driver end-to-end with the Python ADBC driver manager + PyArrow (imports cleanly as `dense_union<int64=0, uint64=1, float64=2, binary=3>` with correct values).

cc @lidavidm


lidavidm (commented):
How bad would it be to update the bundled nanoarrow? AFAIK this should still be a single C file and header?

eitsupi:
Context: #8417

At least that point (3 years ago), it seems that there is a deliberate attempt to avoid adding the entire nanoarrow.
(In thi PR, Incidentally, I considered expanding nanoarrow module to support dense union, but it seemed like it would require hundreds of additional lines, so I've decided against it for now.)

1. Should that policy be continued at this point? Or is it no longer a problem to include nanoarrow?
2. Is it possible to separate the ADBC ​​driver into a separate repository, similar to ODBC and JDBC?
    If so, would it bundling nanoarrow justify?
    (The timing of the DuckDB 2.0 release may be optimal to remove ADBC from libduckdb.)

lidavidm:
I see. I think we need a DuckDB maintainer to chime in, then

pdet:
Thanks! LGTM!

PR #8309, opened 2023-07-19
Finish specialization of the LIST aggregate function

taniabogatsch:
This PR finishes the planned improvements of the `LIST` aggregate function (close #3788). During aggregation (`Update`), we build linked lists with increasing-size links for the different aggregate groups. Then, we combine these linked lists during `Combine`. In `Finalize`, we build a list vector from the resulting linked list of each group.

Improvements in this PR are as follows.

### Zero-copy combine and separate window function

Before this PR, when combining two linked lists, we copied all values of one list into the other instead of appending one to the other (appending includes moving memory to the new state/thread). This was necessary because window functions might require the same linked list multiple times, and after moving the memory out of one state, it is no longer available (see #5554, #5523).

Now, we use the provided window function template for windowing and append lists during `Combine`. We also changed some hashtable implementations to combine their memory (arena-allocated) into the memory of the receiving hashtable during `Combine` (`GroupedAggregateHashTable` and `PerfectAggregateHashTable`). This achieves zero-copy behavior during `Combine` calls and a much improved windowing performance. I also added a new windowing benchmark for the `LIST` aggregate function, and below are some benchmark numbers of benchmarks using `LIST`.

| Benchmark      | master branch | PR |
| ----------- | ----------- | ----------- |
| benchmark/micro/window/window_list_aggr.benchmark | 0.94 | 0.74 |
| benchmark/micro/list/column_data_collection_copy/large_lists.benchmark | 2.29 | 2.09 |
| benchmark/micro/list/column_data_collection_copy/small_lists.benchmark | 0.03 | 0.03 |
| benchmark/micro/list/list_order_by.benchmark | 0.37 | 0.34 |
| benchmark/micro/list/unnest_rewrite.benchmark | 0.19 | 0.17 |

### Others

- removed `RecursiveFlatten`, instead use `RecursiveUnifiedVectorFormat` on each input
- removed custom state-destroy
- unified `FrameBounds` into a struct with two fields

```cpp
struct FrameBounds {
	FrameBounds() : start(0), end(0) {};
	FrameBounds(idx_t start, idx_t end) : start(start), end(end) {};
	idx_t start = 0;
	idx_t end = 0;
};
```

PR #23231, opened 2026-06-11
Rework ordered aggregates to use the `list` infrastructure

Mytherin:
Similar to https://github.com/duckdb/duckdb/pull/23228

Ordered aggregates effectively just gather all inputs, then in the finalize step do a sort and a re-aggregation. Gathering all inputs has been implemented [efficiently by the `LIST` aggregate](https://github.com/duckdb/duckdb/pull/8309).  In addition, we now support that structure (`LinkedList`) explicitly for [aggregate state export](https://github.com/duckdb/duckdb/pull/23199).

As such, both for performance as well as code re-usability reasons, it would be nice if ordered aggregates could just re-use the existing infrastructure of the `LIST` aggregate. That also means that aggregate state export for these aggregates becomes significantly more straightforward (in a follow-up PR), as well as any potential disk spilling as that can then be shared between the sorted aggregates and the `LIST` aggregate.

This PR does exactly that - and refactors the sorted aggregates to effectively just run the same code as `LIST`. All required data is gathered in a single struct vector which is then pushed into the `list` aggregate, i.e. if we have an aggregate like so:

```sql
select string_agg(l_comment order by l_linenumber) from lineitem group by l_orderkey;
```

We would ~more or less run:

```sql
SELECT LIST({'l_comment': l_comment, 'l_linenumber': l_linenumber}) FROM lineitem GROUP BY l_orderkey;
```

But in the finalize step, instead of emitting the list, we scan the list and perform the ordered aggregation.

In particular with many small groups this has substantial performance benefits, primarily because `LIST` has been optimized for exactly this case:

```sql
-- Many Groups
-- 1.5M groups, ~4 rows each group
select string_agg(l_comment order by l_linenumber) from lineitem group by l_orderkey;
```

| Version | Timing (s) |
|---------|------------|
| v1.5    | 2.5s       |
| New     | 0.75s      |

For enormous large aggregates there is a slight performance penalty for this approach, likely because the `ColumnDataCollection`-per-value is better optimized for large scans / appends than the linked list.

```sql
-- Few Groups
-- 4 groups, 1.5M rows per group
select max(strlen(strings)) from (select string_agg(l_comment order by l_orderkey) strings from lineitem group by l_returnflag);
```

| Version | Timing (s) |
|---------|------------|
| v1.5    | 0.28s      |
| New     | 0.33s      |

However, despite the small performance penalty in this scenario, the benefits / added simplicity of unifying this code and the performance boost in the many small groups seems like it is still very worthwhile to make this change. 

CC @hawkfish

### DIFF
--- a/benchmark/micro/aggregate/sorted_string_agg.benchmark
+++ b/benchmark/micro/aggregate/sorted_string_agg.benchmark
@@ -1,0 +1,19 @@
+# name: benchmark/micro/aggregate/sorted_string_agg.benchmark
+# description: STRING_AGG with ORDER BY over many groups
+# group: [aggregate]
+
+name Sorted String Agg (Grouped)
+group aggregate
+
+load
+CREATE TABLE t AS SELECT range % 100000 AS g, range AS r, 'val' || (range % 97)::VARCHAR AS v FROM range(20000000);
+
+run
+SELECT count(*), max(strlen(s)) FROM (
+	SELECT g, STRING_AGG(v, ',' ORDER BY r DESC) AS s
+	FROM t
+	GROUP BY g
+)
+
+result II
+100000	1179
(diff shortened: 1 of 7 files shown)


hawkfish:
@Mytherin It's not just ordered aggregates - I would think this is better for any holistic aggregate!

Mytherin:
For some other holistic aggregates (e.g. `mode`) there's a "reducing" effect to their current implementation (hash map). Although one could argue `mode` should just be implemented using the actual aggregate hash table similar to how `DISTINCT` aggregation works :)

PR #23232, opened 2026-06-11
Fix gzip compression write overflow

dentiny:
Closes https://github.com/duckdblabs/duckdb-internal/issues/9649

The issue is
- miniz's available in bytes uses [unsigned int](https://github.com/duckdb/duckdb/blob/d8f343577c04ab84c32c19f201e1082ea6fa8ade/third_party/miniz/miniz.hpp#L262), which is a type alias for uint32_t in multiple platforms
- overflow happens when ingestion data stream exceeds UINT32_MAX

The fix is straightforward: for each compression iteration, we cap the available bytes with an upper bound.


dentiny:
I'll be surprised if the windows build error is caused by my change
```sh
D:\a\duckdb\duckdb\third_party\fmt\include\fmt\format.h(326,43): error C2653: 'stdext': is not a class or namespace name [D:\a\duckdb\duckdb\src\catalog\catalog_entry\duckdb_catalog_entries.vcxproj]
  (compiling source file '../../catalog/catalog_entry/sequence_catalog_entry.cpp')
```

PR #23236, opened 2026-06-12
Auto generate `select.gram`

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/9667

This PR attempts to auto generate as much of `select.gram` as possible. There's some logic in there that prevent all rules from auto generating, such as `WindowDefinition`, where we set a boolean to prevent nested window functions, but this cannot be cleanly generated I think: 
```cpp
unique_ptr<ParsedExpression> PEGTransformerFactory::TransformWindowDefinition(PEGTransformer &transformer,
                                                                              ParseResult &parse_result) {
	auto &list_pr = parse_result.Cast<ListParseResult>();
	transformer.in_window_definition = true;
	auto window_function = transformer.Transform<unique_ptr<WindowExpression>>(list_pr.Child<ListParseResult>(2));
	transformer.in_window_definition = false;
	window_function->SetAlias(list_pr.Child<IdentifierParseResult>(0).identifier);
	return std::move(window_function);
}
```

Similar cases are `SelectStatementInternal`, `SimpleSelect`, `TableRef`, `WithClause`. 

One thing I wasn't quite happy with, and would like some feedback on, is the representation of `OptionalParseResult`, for example take: 
`TablePivotClause <- 'PIVOT' Parens(TablePivotClauseBody) TableAlias?`
`TableAlias` is optional here, but in previous iterations of the transformer generator, this was just initialized as an empty string. Similarly, `unique_ptr<ParsedExpression>` would become a `nullptr` 
A change I made for the generator in this PR is that `OptionalParseResult` types become `optional<T>`, so they maintain their "optionality semantics" (I don't know a better word). So `TablePivotClause` now generates the following:
```cpp
TransformTablePivotClause(PEGTransformer &transformer, unique_ptr<TableRef> table_pivot_clause_body, const optional<TableAlias> &table_alias)
```
`const optional<TableAlias> &table_alias`

Now my question is, is `optional<T>` the right shape?

Next up, `expression.gram`


Dtenwolde:
My local `make format-fix` didn't find the same issues....

lnkuiper:
LGTM from my side, I think using `optional` is a good idea, very clear

lnkuiper (approved):

PR #23237, opened 2026-06-12
Transform type literals into unbound logical types

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/9538

Fix typed string literals to always bind their target type through the normal unbound type path instead of eagerly materializing partial LogicalTypes in the PEG parser. This avoids internal assertions for incomplete parameterized or special types such as MAP, DECIMAL, and ENUM, and restores the correct binder/conversion errors or successful casts.


Maxxen:
👍 Ideally the parser should not emit any "bound" logical types.

Dtenwolde:
Could you try to change the grammar rule to `TypeLiteral <- Type StringLiteral` while we are at it? Superseding https://github.com/duckdb/duckdb/pull/22567 essentially. Right now, this rule doesn't work with lists, or qualified types, though I'd argue it should :)

`Type` already returns an unbound logical type: `return LogicalType::UNBOUND(std::move(type));`

Mytherin:
Thanks!

Issue #22704, opened 2026-05-15
DuckDB fails to compile with the latest MSVC


TheServer201:
### What happens?

As per the [changelog](https://github.com/microsoft/STL/wiki/Changelog#msvc-build-tools-1451) of MSVC Build Tools 14.51 which ship by default with VS 2026 18.6 the previously deprecated stdext::checked_array_iterator has been removed. The mainline fmtlib/fmt has already fixed this in [issues 3540](https://github.com/fmtlib/fmt/issues/3540). Could you please update the local version of third_party/fmt ?

### To Reproduce

Compile DuckDB with VS 2026 18.6

### OS:

Windows 11 x86_64

### DuckDB Version:

v1.5.2

### DuckDB Client:

C++

### Full Name:

Florent Cabret

### Affiliation:

PhD student

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

PR #22706, opened 2026-05-16
Remove checked_array_iterator from third_party/fmt

staticlibs:
This PR removes [stdext::checked_array_iterator](https://learn.microsoft.com/en-us/cpp/standard-library/checked-array-iterator-class?view=msvc-180) usage applying the changes from [this commit](https://github.com/fmtlib/fmt/commit/9bea6ec04a79bb0a342ed654025c4a15d2016226) to the version of the `fmt` library included in the `third_party` dir.

It is checked on Visual Studio 18.6.0 released on May 12.

Fixes: #22704

PR #23239, opened 2026-06-12
Remove checked_array_iterator from fmt dep (1.4)

staticlibs:
This is a backport of the PR #22706 (#23238 in 1.5) to `v1.4-andium` stable branch.

This PR removes [stdext::checked_array_iterator](https://learn.microsoft.com/en-us/cpp/standard-library/checked-array-iterator-class?view=msvc-180) usage applying the changes from [this commit](https://github.com/fmtlib/fmt/commit/9bea6ec04a79bb0a342ed654025c4a15d2016226) to the version of the `fmt` library included in the `third_party` dir.

It is checked on Visual Studio 18.6.0 released on May 12.

Fixes: #22704

### DIFF
--- a/third_party/fmt/include/fmt/format.h
+++ b/third_party/fmt/include/fmt/format.h
@@ -321,24 +321,13 @@ inline typename Container::value_type* get_data(Container& c) {
   return c.data();
 }
 
-#ifdef _SECURE_SCL
-// Make a checked iterator to avoid MSVC warnings.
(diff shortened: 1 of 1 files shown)


carlopi:
I just merged https://github.com/duckdb/extension-ci-tools/pull/369 on extension-ci-tools side. There is a chance that might fix it also for this CI job.

Could we possibly restart the whole job once again? (I am not sure if restarting only the failures will cut it, since by then it might be already referenced, I am not sure)

PR #23240, opened 2026-06-12
Step 4: Make unnest rewriter recognize CTE-based plans

kryonix:
This PR extends the existing unnest rewrite optimizer to recognize and optimize CTE-based plans. Where necessary, the tests have been adapted to either accept `DELIM_JOIN`-based plans, or CTE-based plans.

### DIFF
--- a/src/include/duckdb/optimizer/unnest_rewriter.hpp
+++ b/src/include/duckdb/optimizer/unnest_rewriter.hpp
@@ -63,6 +63,15 @@ private:
 	//! Find delim joins that contain an UNNEST
 	void FindCandidates(unique_ptr<LogicalOperator> &root, unique_ptr<LogicalOperator> &op,
 	                    vector<reference<unique_ptr<LogicalOperator>>> &candidates);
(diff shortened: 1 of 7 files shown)


lnkuiper (approved):

Codex found one breakage, but this is actually already broken (breaks with and without `delim_join_as_cte`, so it's not a blocker for this PR:
```sql
# name: test/temp.test
# description: Temporary regression tests for PR review
# group: [temp]

statement ok
CREATE TABLE t(a INT, b INT, l INT[]);

statement ok
INSERT INTO t VALUES (1, 10, [5, 6]), (2, 20, [7]);

query IIIII
SELECT a, b, x, s, sa
FROM t, (SELECT UNNEST(l) AS x, a + b AS s, a AS sa) u
ORDER BY ALL;
----
1	10	5	11	1
1	10	6	11	1
2	20	7	22	2

```
Issue is caused by `UpdateRHSBindings`, which can treat user-visible RHS payload columns as removable columns.

PR #23246, opened 2026-06-12
Add explicit `-dark-mode` and `-light-mode` options to the CLI, and improve terminal background color detection

Mytherin:
Follow-up fix to https://github.com/duckdb/duckdb/pull/22838

The 1s time-out silently appeared to make the terminal background detection not interfere with certain shells (that would silently ignore the request - followed by us aborting). Removing that timeout now means those shells get stuck permanently.

This PR fixes up the terminal background detection by sending two requests: the OSC 11 query to get the background color, followed by a DA1 request to get the primary device attributes. The DA1 request is much more broadly supported by many more terminals. If a terminal silently ignores the OSC 11 query, it will likely still respond to the DA1 request. As a result, if we see the DA1 response only, we know that the terminal doesn't support the OSC11 response and we can immediately return without having to wait on a time-out.

Because this is all pretty fragile, we also add a fail-safe where we have a 5 second timeout on terminal background detection now. If that time-out expires, we print the following message:

```
Timeout trying to read terminal background color (> 5s elapsed).
Disable terminal background color detection by using duckdb -dark-mode or duckdb -light-mode.
This likely means duckdb does not correctly support your CLI.
Please file an issue.
```

We now also allow users to explicitly choose the mode using `duckdb -dark-mode` or `duckdb -light-mode` - which fully side-steps the terminal background color detection and can be used as a (temporary) work-around for any further unsupported terminals / CLIs.

### DIFF
--- a/tools/shell/linenoise/terminal.cpp
+++ b/tools/shell/linenoise/terminal.cpp
@@ -472,27 +472,47 @@ bool Terminal::TryGetBackgroundColor(TerminalColor &color) {
 	}
 
 	bool success = false;
-	if (write(ofd, "\x1b]11;?\007", 7) == 7) {
-		// Read the response: until \a or until we fill up our buffer
+	// We send the OSC 11 query to request the background color, followed by a DA1 ("Primary Device Attributes")
+	// Terminals answer queries in order, and virtually every terminal responds to DA1 - but not every
+	// terminal responds to OSC 11. When we receive the DA1 response without a preceding OSC 11 response, we know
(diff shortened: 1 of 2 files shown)


Mytherin:
@carlopi should I still merge this or just close this given you've cherry picked it already?

carlopi:
> @carlopi should I still merge this or just close this given you've cherry picked it already?

I think I cherry-picked the commits from the PRs (instead of the PRs). I think they can as well just be merged (so "needs documentation" or so work), having the same commit in the history should not be a problem.

Code wise, it's irrelevant whether this is merged or closed.

PR #23249, opened 2026-06-12
bounds check the parquet variant binary decoder

jmestwa-coder:
the parquet variant binary decoder trusts sizes and ids read from the file without bounds checking:
- VariantMetadata reads dictionary_size and the string offsets and builds string_t views (bytes + offset) without checking them against the metadata blob size
- ObjectDecode uses the field_id from the value blob as an index into metadata.strings with no range check
reading a parquet file whose variant metadata or value is corrupt hits a heap overread (std::length_error / out-of-range vector read); bound both against the blob size and the dictionary length.


Tishj:
You're welcome to add any additional hardening not already present in https://github.com/duckdb/duckdb/pull/23100 in a PR to v1.5-variegata that will eventually get upstreamed to main :+1:

PR #23250, opened 2026-06-12
Remove redundancies in join filter creation

d-justen:
This PR aims to eliminate redundant dynamic join filters, which can be harmful for performance. For example, a hash join could produce an optional `IN` filter, a selectivity-optional `Min/Max` filter, and a selectivity-optional, exact prefix range filter. Thus, we would for a given row group evaluate all three filters on the statistics and both `Min/Max` and `PRF` per tuple, even though `PRF` is enough to pre-filter all tuples that are not in the build side.

With that, this PR also removes the perfect hash join filter as that is essentially the same as an exact `PRF`, minus the build overhead, which is negligible though, as the build side for perfect hash joins is small by definition. Moreover, only the `PRF` can be used for effective row group pruning.

For this to work, I also had to make a minor change in the row group code, as it would previously disable selectivity-optional  AND optional filters for per-tuple filtering if they are not part of a conjunctive expression.

Generally, the implemented strategy is the following: with a build side of up to 50 keys, we create an optional `IN` filter. If the build range is below `2^26`, we build an exact prefix range filter. If the range is larger, we would build either a `PRF` or a Bloom filter, depending on which would have the smaller size. If we choose the Bloom filter, we also install the Min/Max Filter as an optional filter.


lnkuiper:
This test failure seems related: https://github.com/duckdb/duckdb/actions/runs/27421003312/job/81048130769?pr=23250#step:8:478

Probably just needs to change the regex or skip the test under that config, the expected filter is there but the regex isn't matching under that config. Going to do a review now

d-justen:
Yes, these regex tests are a bit fragile when it comes to whitespace at the moment. But I think the `force_external` failure might hint that prefix range filters currently cause trouble with out-of-core execution. I'm looking into it.

src/include/duckdb/execution/operator/join/join_filter_pushdown.hpp
lnkuiper: Nit: this still says PHJ

lnkuiper (commented):
Thanks for the PR! Looks great. I have just one nit (below), and the failing CI should be fixed, of course, but then this is ready to merge. If out-of-core tests are failing somehow, it's because we can't do BF/PRF for external joins. We should still be able to do min/max, however.

PR #23252, opened 2026-06-12
Add `vertex_extract` function to get coordinates out of point-geometries

Maxxen:
This PR adds a `vertex_extract` function, with shorthand `.x / .y / .z / .m` syntax sugar to extract the x/y/z/m coordinate out of `POINT` geometries. Any other geometry sub-kind will return null.

While this doesn't sound very complicated on its own, this function is heavily optimized to perform stats-propagation, meaning that relatively common range expressions like `select ... where geom.x > 2 AND geom.x < 10` can now push down stats pruning into storage. We also propagate null-flags depending on what geometry sub-types are present in the stats, and can even short-circuit and return a constant null if we know at bind time that there are no non-empty point geometries with the requested coordinate axis present.

In the future I also want to add a `part_extract` function to extract nested geometry parts for non-point geometries by index, e.g. `('LINESTRING(0 0, 1 1, 2 2)'::GEOMETRY)[2] -> 'POINT(1 1)'`. Maybe even support slicing too (e.g. `[1...n]` returns a sub-line string).

### DIFF
--- a/src/function/function_list.cpp
+++ b/src/function/function_list.cpp
@@ -268,6 +268,7 @@ static const StaticFunctionDefinition function[] = {
 	DUCKDB_SCALAR_FUNCTION_SET(VariantKeysFun),
 	DUCKDB_SCALAR_FUNCTION(VariantNormalizeFun),
 	DUCKDB_SCALAR_FUNCTION(VariantTypeofFun),
+	DUCKDB_SCALAR_FUNCTION(VertexExtractFun),
 	DUCKDB_SCALAR_FUNCTION_SET(WriteLogFun),
 	DUCKDB_SCALAR_FUNCTION(ConcatOperatorFun),
 	DUCKDB_SCALAR_FUNCTION(LikeFun),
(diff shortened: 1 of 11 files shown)


Mytherin:
Thanks!

I wonder if we should just have a generic `element_extract` function that extensions can add overloads for? The current implementation is not extensible (which is likely why you're pushing this into core instead of spatial?). It would be cool if we could make this fully extensible going forward.

Maxxen:
Yeah we definitely should have a small set of well-defined built-in functions for all the syntax-sugar operators we got, that can be overloaded on a type-by-type basis. I made a proposal for python style "dunder functions" like `__in__`, `__getattr__` and `__getitem__` a while back, but the exact naming isn't important I guess. There's also the current issue where a lot of our syntax gets bound to functions that are part of `core_functions` and not built-in, e.g. `list_value`, so our base SQL language is dependent on an extensions - which should probably be fixed too.

That said, since `GEOMETRY` is in core now I think it makes sense to keep this here as well, there is not really a equivalent `ST_` function for this functionality, but if this lands I will probably change the dimension-specific `st_x`/`st_y`/`st_z`/`st_m` functions in spatial to just be macros calling this instead.

Mytherin:
Fair enough - thanks!

Issue #23243, opened 2026-06-12
Date functions return `NULL` for infinite dates, but `IS NULL` evaluates to `false`


asd25213:
### What happens?

Date functions (`WEEKOFYEAR`, `WEEKDAY`, `DAYOFMONTH`) return `NULL` when given an infinite date value (`'-Inf'` or `'Inf'`), but the `IS NULL` predicate on the function result evaluates to `false`.

The function result and the `IS NULL` check should agree. If the result is `NULL`, expr `IS NULL` should evaluate to `true`.

### To Reproduce

```sql
CREATE TABLE t0(c0 DATE);
INSERT INTO t0(c0) VALUES ('-Inf');

SELECT (t0.c0), WEEKOFYEAR(t0.c0), WEEKOFYEAR(t0.c0) IS NULL FROM t0;
SELECT (t0.c0), WEEKDAY(t0.c0), WEEKDAY(t0.c0) IS NULL from t0;
SELECT (t0.c0), DAYOFMONTH(t0.c0), DAYOFMONTH(t0.c0) IS NULL from t0;
```

```
┌───────────┬───────────────────┬─────────────────────────────┐
│    c0     │ weekofyear(t0.c0) │ (weekofyear(t0.c0) IS NULL) │
│   date    │       int64       │           boolean           │
├───────────┼───────────────────┼─────────────────────────────┤
│ -infinity │              NULL │ false                       │
└───────────┴───────────────────┴─────────────────────────────┘
┌───────────┬────────────────┬──────────────────────────┐
│    c0     │ weekday(t0.c0) │ (weekday(t0.c0) IS NULL) │
│   date    │     int64      │         boolean          │
├───────────┼────────────────┼──────────────────────────┤
│ -infinity │           NULL │ false                    │
└───────────┴────────────────┴──────────────────────────┘
┌───────────┬───────────────────┬─────────────────────────────┐
│    c0     │ dayofmonth(t0.c0) │ (dayofmonth(t0.c0) IS NULL) │
│   date    │       int64       │           boolean           │
├───────────┼───────────────────┼─────────────────────────────┤
│ -infinity │              NULL │ false                       │
└───────────┴───────────────────┴─────────────────────────────┘
```

### OS:

x86_64

### DuckDB Version:

v1.5.3 (Variegata) 14eca11bd9

### DuckDB Client:

CLI

### Full Name:

Wang Weipeng

### Affiliation:

Nanjing University of Aeronautics and Astronautics

[...]

PR #23254, opened 2026-06-12
Fix NULL propagation for date parts of infinite dates

DinosL:
`PropagateSimpleDatePartStatistics` copied child validity unconditionally, so date parts of infinite dates were marked non null. (`PropagateDatePartStatistics` is doing it right)

Fixes: https://github.com/duckdb/duckdb/issues/23243


test/sql/function/date/date_part_stats.test
```diff
@@ -199,3 +199,23 @@ SELECT stats(WEEKOFYEAR(d)) FROM dates LIMIT 1;
 ----
 <REGEX>:.*1.*54.*
 
+statement ok
```
hawkfish, line 204: Can we add tests for the other infinite temporal types? (`TIMESTAMP`, `TIMESTAMP_XX`)?

hawkfish (changes requested):
I think this also needs to be added to the ICU date part code in `icu-datepart.cpp`? Along with tests for `TIMESTAMP_TZ`?

test/sql/function/date/date_part_stats.test
DinosL: I added tests for the other infinite temporal types. I don't think the ICU date part code needs any changes though. It already nulls out infinite values via the `Timestamp::IsFinite` guard. I added a  test to confirm it behaves correctly.

hawkfish (approved):
LGTM!

hawkfish:
There is also `TIMESTAMP_TZ_NS` on `main` these days but we can wait on that.

PR #23259, opened 2026-06-12
[v1.5] Lock append operation when ongoing index creation

dentiny:
Closes https://github.com/duckdblabs/duckdb-internal/issues/9642
Closes https://github.com/duckdblabs/duckdb-internal/issues/9673

`test/sql/parallelism/interquery/concurrent_writes_during_index_creation.test_slow` SQL test fails intermittently because rows committed during the index build window can be present in the table but absent from the finalized index, since in the current implementation, the new index is not registered in table storage until finalization.

There're a few possible ways to handle:
- Detect conflict at finalization and throw retriable exception
- Use a staged append to buffer rows during index creation
- (this PR) Acquire a lock on index creation, so no rows will be appended at the same time

(3) is selected because
- It's the simplest and most straightforward solution to implement
- We don't it to be the common case that append and index creation to happen at the same time
- (hopefully not) if we do have perf issue reported, it's easy to revert/change/fix

We already have test case to repro the issue, it's just not stable and deterministic. I ran the tests for a couple of times and it didn't repro any more. 

I have ran CI on my own fork: https://github.com/dentiny/duckdb/pull/108, CI failure unrelated

### DIFF
--- a/src/execution/operator/schema/physical_create_index.cpp
+++ b/src/execution/operator/schema/physical_create_index.cpp
@@ -49,9 +49,7 @@ public:
 
 unique_ptr<GlobalSinkState> PhysicalCreateIndex::GetGlobalSinkState(ClientContext &context) const {
 	auto gstate = make_uniq<CreateIndexGlobalSinkState>();
-	if (!alter_table_info) {
-		gstate->append_guard = table.GetStorage().LockAppendsForCreateIndex();
-	}
+	gstate->append_guard = table.GetStorage().LockAppendsForCreateIndex();
 
 	IndexBuildInitGlobalStateInput global_state_input {bind_data.get(),     context,    table, *info,
 	                                                   unbound_expressions, storage_ids};
--- a/src/execution/operator/schema/physical_create_index.cpp
+++ b/src/execution/operator/schema/physical_create_index.cpp
@@ -43,14 +43,14 @@ PhysicalCreateIndex::PhysicalCreateIndex(PhysicalPlan &physical_plan, LogicalOpe
 //---------------------------------------------------------------------------------------------------------------------
 class CreateIndexGlobalSinkState : public GlobalSinkState {
 public:
-	unique_lock<mutex> append_lock;
+	DataTable::IndexBuildAppendGuard append_guard;
 	unique_ptr<IndexBuildGlobalState> gstate;
 };
 
 unique_ptr<GlobalSinkState> PhysicalCreateIndex::GetGlobalSinkState(ClientContext &context) const {
 	auto gstate = make_uniq<CreateIndexGlobalSinkState>();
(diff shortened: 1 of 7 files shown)


Mytherin (commented):
Thanks for the PR!

Do we need the condition variable? Can't we just hold the existing mutex? I feel this is now adding extra complexity which may well introduce new bugs.

dentiny:
> Do we need the condition variable? Can't we just hold the existing mutex? I feel this is now adding extra complexity which may well introduce new bugs.

That was my initial implementation, which fails TSAN; I think the reason is lock is acquired in task scheduler worker thread, but released in client query thread, while unique lock doesn't allow that happening on different threads.

I met the issue on my own fork CI
- TSAN CI: https://github.com/dentiny/duckdb/actions/runs/27397464317/job/80968384793

dentiny:
To repro, I changed back to lock only in https://github.com/dentiny/duckdb/pull/110, and CI reproduces the TSAN failure

dentiny:
Hi @Mytherin , I'm wondering if you're willing to take another look when you have some time? Thank you!

Mytherin:
Thanks for the updates!

The locking code still seems quite complex to me with the condition variable, the index build count and the mutex all being grabbed. Given that we chose locking as it's the easiest path I wonder if we should reconsider given it doesn't appear to be that simple to do correctly. The other approaches are also more attractive from a concurrency perspective.

I've given this some thought and this approach actually seems relatively straightforward:

> Use a staged append to buffer rows during index creation

The reason being we already have code to buffer changes made to the index - namely the `UnboundIndex` already buffers changes with `BufferChunk` (both insertions and deletions).

Maybe what we could do is register a dummy `UnboundIndex` in the table in `PhysicalCreateIndex::GetGlobalSinkState` - which will then buffer inserts and deletes that should have been made to the index. Then in `PhysicalCreateIndex::Finalize` - instead of just adding the index to the list, we "swap" the `UnboundIndex` with our new index and apply the buffered replays (if any) to the index.
 
I think that should be safe and means we don't need to hold the lock during index creation - offering much better concurrency during the index creation process.

dentiny:
Hi @Mytherin , since concurrent write for index creation and data ingestion is taking me some time, I'm wondering if you think it makes sense to check it into **v1.5.5 bugfix release only**, while I working on the more proper fix on the main at the same time? Hopefully it addresses reported crash at least. I've rebased the branch to v1.5, let me know if you think it's ok, thank you!

Mytherin:
I think this is a somewhat high risk fix for what is an issue that has already existed in the system for a long time. I would prefer to push the preferred (more involved) fix to main and not merge this into a bug-fix release late in the cycle.

dentiny:
cool, I will close this PR then, and open a new PR for proper fix.

PR #22819, opened 2026-05-21
Add expression stats pruning for arithmetic filters

Damon07:
This PR supersedes #22647. It extends expression filter statistics pruning so pushed filters can derive statistics for scalar expressions, including arithmetic and struct extraction, and use those derived stats for zonemap pruning.

  Main changes:

  - Pass ClientContext through relevant zonemap pruning paths when available.
  - Add `ExpressionFilter::CheckStatistics(ClientContext &, ...)` / `CheckExpressionStatistics(optional_ptr<ClientContext>, ...)` paths while keeping the existing no-context overloads.
  - Recursively derive filter-side statistics for constants, column refs, scalar functions, struct extraction, comparisons, IN, NULL checks, and conjunctions.
  - Add statistics propagation for arithmetic +, -, and *, including floating-point bound widening.
  - Register struct extract statistics propagation for `struct_extract` and `struct_extract_at`.

PR #23264, opened 2026-06-13
Parquet: add expression stats pruning for arithmetic filters

Damon07:
Follow-up to #22819, this PR supports arithmetic filters pruning for parquet files in row-group scan. Since it is a little complicated to pass context through page skip path and  currently there is no observability at the page level(no per-page skip metric/counter) so it is not included in this PR.


Mytherin:
Thanks!

PR #23266, opened 2026-06-13
fix out-of-bounds read in strict time parsing

jmestwa-coder:
strict time parsing reads one byte past the end of the input:
- TryConvertInternal takes the HH:MM branch when pos == len after a single-digit minute, then reads buf[pos] as the seconds separator
- with pos == len that read is buf[len], one past the buffer; asan flags a heap-buffer-overflow read of size 1 in TryConvertInternal
- reachable from strict TIME casts on untrusted strings (csv type sniffing/refinement, comparison constant folding) for input like '1:2'

the date parser already guards each separator read with pos >= len; the time parser used pos > len and let pos == len through. handled the pos == len case so the separator is only read while pos < len.


test/common/test_time_parsing.cpp
```diff
@@ -0,0 +1,30 @@
+#include "catch.hpp"
```
hawkfish: Can we just write a SQL test for this? (We try to avoid unit tests like this.) With your fix we should be able to just test that we get an error on an invalid string.

hawkfish (changes requested):
Thanks for the PR! Just move the validation into a SQL test. Maybe add them to `test_incorrect_timestamp.test`?

test/common/test_time_parsing.cpp
jmestwa-coder: done. dropped the unit test and added the cases to time_parsing.test, since that's where the rest of the time parse tests live (can move them to test_incorrect_timestamp.test if you'd prefer).

note that a plain `'1:2'::TIME` isn't strict, so it parses to 01:02:00 without erroring. the strict path that does the over-read is reached through the variant cast, so the test runs `'1:2'::VARIANT::TIME` and checks it errors.

hawkfish (approved):

hawkfish:
LGTM!  Can you mark the PR as ready for review?

jmestwa-coder:
@hawkfish 
Done. I mark the PR as ready for review

Mytherin:
Thanks!

PR #23269, opened 2026-06-14
Row variables of NULL-padded join sides evaluate to NULL

carlopi:
This is arguably a postgres compatibility issue / and generally it allows POSITIONAL JOINs to carry more informations that they would otherwise, since:

```sql
CREATE TABLE A (i int);
INSERT INTO A VALUES (1), (2), (3), (NULL), (NULL);
CREATE TABLE B (i int);
INSERT INTO B VALUES (1), (2), (3), (NULL);

SELECT a, b FROM a POSITIONAL JOIN b;
```
would currently return:
```
┌───────────────────┬───────────────────┐
│         a         │         b         │
│ struct(i integer) │ struct(i integer) │
├───────────────────┼───────────────────┤
│ {'i': 1}          │ {'i': 1}          │
│ {'i': 2}          │ {'i': 2}          │
│ {'i': 3}          │ {'i': 3}          │
│ {'i': NULL}       │ {'i': NULL}       │
│ {'i': NULL}       │ {'i': NULL}       │
└───────────────────┴───────────────────┘
```
while, after the PR:
```
┌───────────────────┬───────────────────┐
│         a         │         b         │
│ struct(i integer) │ struct(i integer) │
├───────────────────┼───────────────────┤
│ {'i': 1}          │ {'i': 1}          │
│ {'i': 2}          │ {'i': 2}          │
│ {'i': 3}          │ {'i': 3}          │
│ {'i': NULL}       │ {'i': NULL}       │
│ {'i': NULL}       │ NULL              │
└───────────────────┴───────────────────┘
```

Basic is: there is more information available, at a cost of extra machinery.

Mechanism: a new LogicalRowPresence operator is placed below the NULL-paddable side(s) of LEFT/RIGHT/FULL/POSITIONAL joins. It passes its child through unchanged and adds one BOOLEAN column (true) under its own table index; the join's existing NULL-padding covers it, so no join executor changes are needed. Row variables expand to CASE WHEN <presence> THEN struct_pack(...) ELSE NULL END. The presence binding is hidden from star expansion and unqualified column resolution. Physically the operator is planned as a projection.

The rule is uniform across source kinds: base tables, subqueries, views, CTEs, table functions and set operations.

Currently no test (I think) assert on this behaviour, so I think there is still wiggle room, might not be intentional. A simpler change could also add a test that documents the current behaviour is intended.


Mytherin:
Thanks - but this seems like way too much added complexity / extra machinery for what seems like a minor issue. Adding extra logical operators in particular seems like way too overkill here.

Mytherin:
I'm not opposed to the behavioral change but this needs to be fixed in a much less structurally impactful manner.

Mytherin:
I think structurally the correct fix here is to have the `LogicalGet` emit the struct, instead of what we currently do where the `struct_pack` is constructed in the projection (which is *after* joins have been resolved, instead of before).

Mytherin:
I also think this is not really related to positional joins, left joins have the same issue (and are actually supported in Postgres, showing the discrepency):

```sql
create table t1(i int);
insert into t1 values (1), (2), (3);
create table t2(i int, v varchar);
insert into t2 values (1, 'hello'), (3, 'world');
select t2 from t1 left join t2 using (i);
│              t2              │
│ struct(i integer, v varchar) │
│ {'i': 1, 'v': hello}         │
│ {'i': 3, 'v': world}         │
│ {'i': NULL, 'v': NULL}       │
```

carlopi:
> I think structurally the correct fix here is to have the `LogicalGet` emit the struct, instead of what we currently do where the `struct_pack` is constructed in the projection (which is _after_ joins have been resolved, instead of before).

I am not sure I see this completely, since not all tables are physical tables (that is a `LogicalGet`). There the only paths I am seeing it's either something like this new operator (probably it would be better to emit `struct_pack` directly there, moving it up before the join, instead of a bool), that to me guarantees uniformly to wrap any child. An alternative would be a generic virtual column such as `rowid` is guaranteed to be populated with no NULL values.

carlopi:
Anyhow, this is not super relevant, if one needs it can be worked around, for example packing an explicit extra bool like:
```sql
SELECT a, b
    FROM (SELECT a, true as a_is_present FROM a)
    POSITIONAL JOIN b WHERE a_is_present;
```

Mytherin:
You can just push a projection that does the `struct_pack` before the join:

```sql
select t2 from t1 left join (select t2, i from t2) using (i);
│              t2              │
│ struct(i integer, v varchar) │
│ {'i': 1, 'v': hello}         │
│ {'i': 3, 'v': world}         │
│ NULL                         │
explain select t2 from t1 left join (select t2, i from t2) using (i);

┌───────────────────────────┐
│         PROJECTION        │
│    ────────────────────   │
│             t2            │
│                           │
│          ~3 rows          │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         HASH_JOIN         │
│    ────────────────────   │
│      Join Type: RIGHT     │
│     Conditions: i = i     ├──────────────┐
│          ~3 rows          │              │
│         PROJECTION        ││          SEQ_SCAN         │
│     struct_pack(i, v)     ││   Table: memory.main.t1   │
│             i             ││   Type: Sequential Scan   │
│                           ││       Projections: i      │
│          ~2 rows          ││          ~3 rows          │
└─────────────┬─────────────┘└───────────────────────────┘
┌─────────────┴─────────────┐
│          SEQ_SCAN         │
│    ────────────────────   │
│   Table: memory.main.t2   │
│   Type: Sequential Scan   │
│                           │
│        Projections:       │
│             i             │
│             v             │
│                           │
│          ~2 rows          │
└───────────────────────────┘

```

carlopi:
Improved version, with no operator, at https://github.com/duckdb/duckdb/pull/23305.

I wanted to see if I could see the end of this, but it's indeed complex for limited value, and might be as well filed with expected behaviour or first doing others cleanups, and then approach this.

Issue #22325, opened 2026-04-28
Parquet DATA_PAGE_V2 truncated payload leaves unread heap tail reachable by the delta decoder


fuyu0425:
## Summary

Current DuckDB head `0da6e1e84c60c0d5aebd35d7758fcdb0f8bb6eb7`
(`v1.6.0-dev4696`) still lets a malformed Parquet `DATA_PAGE_V2` header reach
`DELTA_BINARY_PACKED` value decoding and trigger an uninitialised read before
the page is rejected.

In this specific page header, the original compact-Thrift bytes encode:

- `PageHeader.type = DATA_PAGE_V2`,
- `uncompressed_page_size = 10`,
- `compressed_page_size = 12`,
- `num_values = 5`,
- `num_nulls = 0`,
- `num_rows = 5`,
- `encoding = DELTA_BINARY_PACKED`,
- `definition_levels_byte_length = 0`, and
- `repetition_levels_byte_length = 0`.

The one-byte mutation changes only the compact-encoded
`compressed_page_size` value from `12` to `0`, leaving the rest of that page
header intact.

The reproduced current-head behavior is:

- a checked-in DuckDB Parquet fixture works normally under memcheck,
- a one-byte mutation changes only the page header byte at file offset `72`
  from `0x18` to `0x00`,
- the control query still returns `1, 2, 3, 4, 5`,
- the mutated query exits with memcheck `error-exitcode=88`, and
- memcheck reports uninitialised-value use in
  `ParquetDecodeUtils::VarintDecode()` while the allocation origin points back
  to `ResizeableBuffer::resize()`.

The same run then surfaces:

```text
Invalid Input Error: Parquet file has invalid number of miniblocks per block for DELTA_BINARY_PACKED
```

That error happens only after the unread heap tail has already been consumed by
the decoder.

## Reproduction

Prerequisites:

- `git`
- `make`
- `cmake`
- a C++17 compiler
- `valgrind`

Build:

```bash
git clone https://github.com/duckdb/duckdb.git
cd duckdb
git checkout 0da6e1e84c60c0d5aebd35d7758fcdb0f8bb6eb7
DISABLE_SANITIZER=1 make debug -j4
```

Create the minimal mutant from DuckDB's checked-in fixture:

```bash
cp data/parquet-testing/arrow/datapage_v2.snappy.parquet datapage_v2-truncated.parquet
[...]

PR #23279, opened 2026-06-15
parquet: reject DATA_PAGE_V2 pages with inconsistent compressed_page_size

benfleis:
Reject parquet file and avoid invalid memory reads when:

1. compressed_page_size < rep+def level bytes: would produce a negative compressed_bytes, silently skipping decompression and leaving the allocated buffer uninitialized before value decoding.
2. compressed_bytes == 0 with a non-empty value region (uncompressed_page_size > uncompressed level bytes): zero-sized compressed payload for non-zero data, same effect — decoder reads uninitialized heap.
3. assert that signed vals which may only be non-negative are in fact non-negative

Tests exercising these conditions included.

Addresses #22325


benfleis:
Leaving draft until CI infra less in-demand

PR #23283, opened 2026-06-15
Async Parquet Writes

lnkuiper:
This PR implements asynchronous file writes and integrates them into the Parquet writer. This removes the row group flushing from global file lock, which should make row group flushing cheaper: only registering and scheduling async writes now happens under lock.

## Current Implementation

Parquet currently writes using `BufferedFileWriter`, so row group flushing performs the actual file writes while holding the Parquet writer lock. This is not great for remote writes, where writes can involve roundtrips. We have some optimizations in our remote file systems for this, but in general we are doing too much work under lock.

## New Implementation

This PR adds a reusable async write layer and uses it from `ParquetWriter`.

The main pieces are:

1. `AsyncWriteQueue` is the low-level positional write scheduler. It owns async task scheduling for independent physical write requests.
2. `ManagedAsyncWriteQueue` adds memory accounting, and backpressure (integration with `TemporaryMemoryManager`!).
3. `AsyncFileWriter`, an easy replacement for `BufferedFileWriter`, with buffering behavior for small writes and synchronous fallback when `async_threads=0`.

For local files, the async writer uses smaller coalescing similar to `BufferedFileWriter`. For remote files, it waits for larger coalesced writes before scheduling the writes, so tiny writes do not immediately turn into remote roundtrips.

The file system now exposes `SupportsPositionalWrites()`. Local files report support, and file systems that do not support independent positional writes fall back to sequential draining. This keeps the async writer usable without requiring every file system to support parallel positional writes immediately. I'm not 100% sure about this addition, but I needed something to activate the new implementation safely.

## Parquet Integration

The Parquet writer now uses `AsyncFileWriter` instead of `BufferedFileWriter`.

The row group preparation path now also prepares the bytes that will be written later. For primitive columns, final page flushing, dictionary page materialization, compression, page header serialization, and footer-key encryption of page payloads are done before the row group is flushed. `FlushRowGroup` then enters an `AsyncFileWriter` batch, registers the prepared buffers in order, updates the row group metadata, and finishes the batch so async draining can start.

This keeps the flushing path (under lock) cheap while preserving the existing logical write order and file offsets.

## Backpressure And Close Semantics

Queued async write memory is tracked through `TemporaryMemoryManager`, with an upper bound derived from the number of regular execution threads. Backpressure makes the registering thread help drain work when pending bytes exceed the current budget.

## Benchmarks

I would have loved to do an S3 Parquet writing benchmark, but that is not yet possible because our S3 file system (in `duckdb-httpfs`) does its own internal buffering which makes it so the writes aren't really asynchronous after all. This is future work. For now, however, I have done a little local writing benchmark on my laptop.

The query I chose for this is:
```sql
COPY lineitem
TO '/tmp/lineitem_async' (
	FORMAT parquet,
	FILE_SIZE_BYTES '64mib',
	PER_THREAD_OUTPUT false,
	OVERWRITE
);
```
I've compared `main` vs. this PR, with `PER_THREAD_OUTPUT true` as well.
Annoyingly, the difference of all four configurations is just a few milliseconds, well within expected noise.
One the one hand, this is great news, because that means that `PER_THREAD_OUTPUT` should hopefully not be needed anymore in the future.
On the other hand, this means that my benchmark won't reveal meaningful performance differences.

If I do `threads=8` (my laptop has 10 cores) and measure the difference of this query with `async_threads=0` and the default (20), I can see a ~4% improvement with async threads on.


Mytherin:
Looks great, thanks!

hannes:
Maybe there is a way to effectively disable buffering in httpfs?

lnkuiper:
I'd like to disabled buffering by default there but that would require a lot more changes in DuckDB and extensions. Maybe we can implement opt-in disabled buffering for the s3 file system first?

PR #23284, opened 2026-06-15
Support AFTER INSERT FOR EACH ROW triggers

EtgarDev:
## Support `AFTER INSERT FOR EACH ROW` triggers
Implements `FOR EACH ROW` execution for `AFTER INSERT` triggers.
The trigger body is planned as a `LogicalTrigger` and fired once per affected row by reusing DuckDB's subquery decorrelation, with `NEW.col` references resolved as correlated columns.

### Semantics
1. The body fires once per row actually written by the `INSERT` (driven by the DML's returning output).
2. All firings run as a single `set-based` batch against one transaction snapshot.
3. A constraint failure on the base table rolls back the statement and the trigger doesn't fire.

### Rejected at bind time
- `BEFORE` and `UPDATE` / `DELETE` row triggers
- Bodies that don't reference any `NEW` column (the whole machinery is driven by the correlation (an attempt to inject a "phantom" `NEW` correlation to force per-row failed but I'm planning to give it another try in the future)
- Self-referential bodies (writing to the trigger table)
- Cascades - a body that writes to a table which itself has a `FOR EACH ROW` trigger
- Mixing `FOR EACH ROW` and `FOR EACH STATEMENT` on the same table
- `RETURNING` and `ON CONFLICT` on a table with row triggers

### Known limitations
Because every row fires as one batch against a single snapshot rather than iterating rows sequentially:
- Visibility: a firing cannot see rows written by an earlier firing in the same statement
- Order: rows fire in an unspecified order, and with multiple triggers each trigger fires for the whole batch before the next, not per-row interleave.

Correct semantics for these probably require per-row iterative execution, which is not yet implemented.


src/planner/subquery/flatten_dependent_join.cpp
```diff
@@ -1012,6 +1016,29 @@ vector<ColumnBinding> FlattenDependentJoins::PushDownCorrelatedNode(unique_ptr<L
...
+				child.ResolveOperatorTypes();
+			}
+		}
+		// DML output does not carry the child columns, so re-expose the correlation keys in a separate DELIM_GET that
+		// the parent DelimJoin can reference.
+		auto expose_idx = binder.GenerateTableIndex();
+		unique_ptr<LogicalOperator> expose_delim = make_uniq<LogicalDelimGet>(expose_idx, delim_types);
+		plan = LogicalCrossProduct::Create(std::move(plan), std::move(expose_delim));
+		return CreateContiguousState(ColumnBinding(expose_idx, ProjectionIndex(0)));
+	}
```
kryonix: I think this deserves it's own function, instead of inlining it here (Before I refactored this module, the `PushDownCorrelatedNode` function was huge and monolithic, I'd prefer not going into that direction again.)

EtgarDev: @kryonix thanks :)
I made a new function - `PushDownDML`

lnkuiper (changes requested):
Thanks for the PR! Agree with Denis' comment, and I have some more:

## Breakages

I've found two breakages with Codex:
```sql
# name: test/temp.test
# description: Temporary reproducer for PR 23284 review
# group: [trigger]

require noforcestorage

# 1. Direct correlated DELETE trigger bodies currently produce a wrong result
statement ok
CREATE TABLE src(id INTEGER);

statement ok
CREATE TABLE other(id INTEGER);

statement ok
INSERT INTO other VALUES (1), (2), (3);

statement ok
CREATE TRIGGER src_delete_trg AFTER INSERT ON src FOR EACH ROW
    DELETE FROM other WHERE id = NEW.id;

statement ok
INSERT INTO src VALUES (1), (3);

query I
SELECT id FROM other ORDER BY id;
----
2

# 2. Row triggers on tables with generated columns can hit an internal assertion
statement ok
CREATE TABLE g(a INTEGER, b INTEGER GENERATED ALWAYS AS (a + 1));

statement ok
CREATE TABLE g_log(a INTEGER);

statement ok
CREATE TRIGGER g_trg AFTER INSERT ON g FOR EACH ROW
    INSERT INTO g_log VALUES (NEW.a);

statement ok
INSERT INTO g(a) VALUES (1), (2);

query I
SELECT a FROM g_log ORDER BY a;
----
1
2

```

### Relevant code for (1.)

- `FlattenDependentJoins::PushDownCorrelatedNode` has a new DML branch for `LOGICAL_INSERT`, `LOGICAL_UPDATE`, and `LOGICAL_DELETE` in `src/planner/subquery/flatten_dependent_join.cpp:1019`.
- That branch skips projection trimming for `LOGICAL_DELETE` because `PhysicalDelete` reads row IDs by fixed index, see `src/planner/subquery/flatten_dependent_join.cpp:1023`.
- It then re-exposes correlation keys using a `LogicalDelimGet` cross product in `src/planner/subquery/flatten_dependent_join.cpp:1035`.

The plan shape for direct `DELETE FROM other WHERE id = NEW.id` appears to decorrelate differently from `DELETE FROM other WHERE id IN (SELECT NEW.id)`: the subquery form deletes correctly in my probe, while the direct correlated predicate does not.

### Relevant code for (2.):

- `Binder::ExpandRowTriggers` injects `RETURNING *`: `src/planner/binder/query_node/bind_trigger_expansion.cpp:473`.
- The `NEW` binding uses `table.GetColumns().Physical()`: `src/planner/binder/query_node/bind_trigger_expansion.cpp:502`.
- Trigger creation validation also builds row-scope `NEW` from physical columns only: `src/planner/binder/statement/bind_create.cpp:586`.

PostgreSQL and SQLite both expose generated columns through `NEW` in the comparable cases. If DuckDB does not want to support that yet, this should be rejected cleanly rather than asserting.

### Cascade

The goal is to reject triggers we can't fire during `CREATE`, but this is still rejected at execution time instead:

```sql
  CREATE TABLE a(id INTEGER);
  CREATE TABLE b(id INTEGER);
  CREATE TABLE c(id INTEGER);

  CREATE TRIGGER b_trg AFTER INSERT ON b FOR EACH ROW
      INSERT INTO c VALUES (NEW.id);

  CREATE TRIGGER a_trg AFTER INSERT ON a FOR EACH ROW
      INSERT INTO b VALUES (NEW.id);

  INSERT INTO a VALUES (1);
```

Is that intentional?

## Discussion: ORDER BY?

Set-based (instead of row-based) triggers will have great performance. However, because of the set-based logic we will have non-deterministic triggers. We could consider adding an `ORDER BY` to make it deterministic. Just putting this out here so maybe more people can chime in

### Relevant code:

Relevant code:

- `RewriteTriggersToDependent` injects a synthetic `delim_index` in `src/planner/planner.cpp:46`.
- `CreateRowNumberWindow` accepts optional `orders` but the trigger path currently calls it without an order in `src/planner/subquery/flatten_dependent_join.cpp:296`.
- `PrepareDependentJoinLeft` creates the synthetic row-number window when `perform_delim` is false in `src/planner/subquery/flatten_dependent_join.cpp:332`.

lnkuiper:
Hey @kryonix @Mytherin just wanted to get your opinion on adding the ORDER BY.

Consider you have a trigger that logs inserts. The subquery decorrelation would make the order in which the log is inserted non-deterministic, which could be very confusing. Maybe we should add the ORDER BY? Or not and then document the behaviour.

Mytherin:
I think it's better to not order by default (and have it be fast) and allow users to add the order themselves if they want to maintain order.

lnkuiper:
Thanks for chiming in! We can add the option to make the trigger firing deterministic in a future PR then

EtgarDev:
@lnkuiper @kryonix Thank you very much for your review :)
Great findings, I tackled the `DELETE` issue, it seems like it failed because of the join-reorder optimizer, which for the first time encountered DMLs as the operator which is being "peeled", and the reconstruction's output it makes results without the `DELETE`. It "luckily" didn't break `UPDATE` and `INSERT` because `INSERT`'s child is always an `EXPRESSION_GET` because of the `VALUES(...)`  or `PROJECTION` when we insert from a `SELECT` clause, and those children have cases in the long switch-case that call `AddRelation` with the original `INSERT` operator regardless of how far the [loop](https://github.com/duckdb/duckdb/blob/86040b84069c1d3064173f7e16cbc4b2f50d748e/src/optimizer/join_order/relation_manager.cpp#L275) peeled down.
`UPDATE` is also preserved because its child is always a `PROJECTION` because `PhysicalUpdate` requires it to project new column values + row_id.


I also covered this with more test cases.

Before the fix, we can see that the `DELETE` is not on the optimized logical plan:
```sql
CREATE TABLE src(id INTEGER);
CREATE TABLE other(id INTEGER);
INSERT INTO other VALUES (1), (2), (3);
CREATE TRIGGER src_delete_trg AFTER INSERT ON src FOR EACH ROW
    DELETE FROM other WHERE id = NEW.id;
EXPLAIN INSERT INTO src VALUES (1), (3);

││  Optimized Logical Plan   ││
┌───────────────────────────┐
│            CTE            │
│    ────────────────────   │
│         CTE Name:         │
│__duckdb_trigger_base_a163b│
│     b53-18e4-4004-a5ac    │
│       -68843c63c279       ├──────────────┐
│       Table Index: 0      │              │
│           ~1 row          │              │
│         PROJECTION        ││         PROJECTION        │
│      Expressions: id      ││      Expressions: #0      │
│          ~2 rows          ││           ~1 row          │
└─────────────┬─────────────┘└─────────────┬─────────────┘
│           INSERT          ││            CTE            │
│                           ││         CTE Name:         │
│                           ││__duckdb_trigger_body_row_a│
│                           ││   163bb53-18e4-4004-a5ac  │
│                           ││       -68843c63c279       ├─────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                           ││                           │                                                                                                     │
│                           ││      Table Index: 24      │                                                                                                     │
│                           ││                           │                                                                                                     │
│          ~2 rows          ││           ~1 row          │                                                                                                     │
└─────────────┬─────────────┘└─────────────┬─────────────┘                                                                                                     │
┌─────────────┴─────────────┐┌─────────────┴─────────────┐                                                                                       ┌─────────────┴─────────────┐
│         PROJECTION        ││         PROJECTION        │                                                                                       │         PROJECTION        │
│    ────────────────────   ││    ────────────────────   │                                                                                       │    ────────────────────   │
│      Expressions: id      ││      Expressions: #0      │                                                                                       │        Expressions:       │
│                           ││                           │                                                                                       │        count_star()       │
│                           ││                           │                                                                                       │                           │
│          ~2 rows          ││          ~2 rows          │                                                                                       │           ~1 row          │
└─────────────┬─────────────┘└─────────────┬─────────────┘                                                                                       └─────────────┬─────────────┘
┌─────────────┴─────────────┐┌─────────────┴─────────────┐                                                                                       ┌─────────────┴─────────────┐
│       EXPRESSION_GET      ││         PROJECTION        │                                                                                       │         AGGREGATE         │
│    ────────────────────   ││    ────────────────────   │                                                                                       │    ────────────────────   │
│                           ││      Expressions: #0      │                                                                                       │        Expressions:       │
│                           ││                           │                                                                                       │        count_star()       │
│                           ││                           │                                                                                       │                           │
│          ~2 rows          ││          ~2 rows          │                                                                                       │           ~1 row          │
└─────────────┬─────────────┘└─────────────┬─────────────┘                                                                                       └─────────────┬─────────────┘
┌─────────────┴─────────────┐┌─────────────┴─────────────┐                                                                                       ┌─────────────┴─────────────┐
│         DUMMY_SCAN        ││         DELIM_JOIN        │                                                                                       │          CTE_SCAN         │
│    ────────────────────   ││    ────────────────────   │                                                                                       │    ────────────────────   │
│                           ││      Join Type: INNER     │                                                                                       │        CTE Index: 0       │
│                           ││                           │                                                                                       │                           │
│                           ││        Conditions:        ├──────────────┐                                                                        │                           │
│                           ││    (delim_index IS NOT    │              │                                                                        │                           │
│                           ││ DISTINCT FROM delim_index)│              │                                                                        │                           │
│                           ││                           │              │                                                                        │                           │
│                           ││          ~2 rows          │              │                                                                        │          ~2 rows          │
└───────────────────────────┘└─────────────┬─────────────┘              │                                                                        └───────────────────────────┘
                             │           WINDOW          ││      COMPARISON_JOIN      │
                             │        Expressions:       ││      Join Type: INNER     │
                             │        limit_rownum       ││   Conditions: (id = id)   ├──────────────┐
                             │                           ││                           │              │
                             │          ~2 rows          ││           ~1 row          │              │
                             └─────────────┬─────────────┘└─────────────┬─────────────┘              │
                             ┌─────────────┴─────────────┐┌─────────────┴─────────────┐┌─────────────┴─────────────┐
                             │         PROJECTION        ││          SEQ_SCAN         ││       CROSS_PRODUCT       │
                             │    ────────────────────   ││    ────────────────────   ││    ────────────────────   │
                             │      Expressions: id      ││           Table:          ││                           │
                             │                           ││     memory.main.other     ││                           │
                             │                           ││                           ││                           ├──────────────┐
                             │                           ││   Type: Sequential Scan   ││                           │              │
                             │                           ││                           ││                           │              │
                             │          ~2 rows          ││          ~3 rows          ││           ~1 row          │              │
                             └─────────────┬─────────────┘└───────────────────────────┘└─────────────┬─────────────┘              │
                             ┌─────────────┴─────────────┐                             ┌─────────────┴─────────────┐┌─────────────┴─────────────┐
                             │          CTE_SCAN         │                             │         DELIM_GET         ││         DELIM_GET         │
                             │    ────────────────────   │                             │    ────────────────────   ││    ────────────────────   │
                             │        CTE Index: 0       │                             │                           ││                           │
                             │                           │                             │                           ││                           │
                             │          ~2 rows          │                             │           ~1 row          ││           ~1 row          │
                             └───────────────────────────┘                             └───────────────────────────┘└───────────────────────────┘
```

And after the fix:
```sql
││  Optimized Logical Plan   ││
┌───────────────────────────┐
│            CTE            │
│    ────────────────────   │
│         CTE Name:         │
│__duckdb_trigger_base_a6dc2│
│     9ca-955c-4a50-ace9    │
│       -7ed12ef6f5f0       ├──────────────┐
│       Table Index: 0      │              │
│           ~1 row          │              │
│         PROJECTION        ││         PROJECTION        │
│      Expressions: id      ││      Expressions: #0      │
│          ~2 rows          ││           ~1 row          │
└─────────────┬─────────────┘└─────────────┬─────────────┘
│           INSERT          ││            CTE            │
│                           ││         CTE Name:         │
│                           ││__duckdb_trigger_body_row_a│
│                           ││   6dc29ca-955c-4a50-ace9  │
│                           ││       -7ed12ef6f5f0       ├─────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                           ││                           │                                                                                                     │
│                           ││      Table Index: 24      │                                                                                                     │
│                           ││                           │                                                                                                     │
│          ~2 rows          ││           ~1 row          │                                                                                                     │
└─────────────┬─────────────┘└─────────────┬─────────────┘                                                                                                     │
┌─────────────┴─────────────┐┌─────────────┴─────────────┐                                                                                       ┌─────────────┴─────────────┐
│         PROJECTION        ││         PROJECTION        │                                                                                       │         PROJECTION        │
│    ────────────────────   ││    ────────────────────   │                                                                                       │    ────────────────────   │
│      Expressions: id      ││      Expressions: #0      │                                                                                       │        Expressions:       │
│                           ││                           │                                                                                       │        count_star()       │
│                           ││                           │                                                                                       │                           │
│          ~2 rows          ││          ~2 rows          │                                                                                       │           ~1 row          │
└─────────────┬─────────────┘└─────────────┬─────────────┘                                                                                       └─────────────┬─────────────┘
┌─────────────┴─────────────┐┌─────────────┴─────────────┐                                                                                       ┌─────────────┴─────────────┐
│       EXPRESSION_GET      ││         PROJECTION        │                                                                                       │         AGGREGATE         │
│    ────────────────────   ││    ────────────────────   │                                                                                       │    ────────────────────   │
│                           ││      Expressions: #0      │                                                                                       │        Expressions:       │
│                           ││                           │                                                                                       │        count_star()       │
│                           ││                           │                                                                                       │                           │
│          ~2 rows          ││          ~2 rows          │                                                                                       │           ~1 row          │
└─────────────┬─────────────┘└─────────────┬─────────────┘                                                                                       └─────────────┬─────────────┘
┌─────────────┴─────────────┐┌─────────────┴─────────────┐                                                                                       ┌─────────────┴─────────────┐
│         DUMMY_SCAN        ││         DELIM_JOIN        │                                                                                       │          CTE_SCAN         │
│    ────────────────────   ││    ────────────────────   │                                                                                       │    ────────────────────   │
│                           ││      Join Type: INNER     │                                                                                       │        CTE Index: 0       │
│                           ││                           │                                                                                       │                           │
│                           ││        Conditions:        ├──────────────┐                                                                        │                           │
│                           ││    (delim_index IS NOT    │              │                                                                        │                           │
│                           ││ DISTINCT FROM delim_index)│              │                                                                        │                           │
│                           ││                           │              │                                                                        │                           │
│                           ││          ~2 rows          │              │                                                                        │          ~2 rows          │
└───────────────────────────┘└─────────────┬─────────────┘              │                                                                        └───────────────────────────┘
                             │           WINDOW          ││       CROSS_PRODUCT       │
                             │        Expressions:       ││                           │
                             │        limit_rownum       ││                           ├──────────────┐
                             │                           ││                           │              │
                             │          ~2 rows          ││           ~1 row          │              │
                             └─────────────┬─────────────┘└─────────────┬─────────────┘              │
                             ┌─────────────┴─────────────┐┌─────────────┴─────────────┐┌─────────────┴─────────────┐
                             │         PROJECTION        ││         DELIM_GET         ││           DELETE          │
                             │    ────────────────────   ││    ────────────────────   ││    ────────────────────   │
                             │      Expressions: id      ││                           ││     Expressions: rowid    │
                             │                           ││                           ││                           │
                             │          ~2 rows          ││           ~1 row          ││           ~1 row          │
                             └─────────────┬─────────────┘└───────────────────────────┘└─────────────┬─────────────┘
                             ┌─────────────┴─────────────┐                             ┌─────────────┴─────────────┐
                             │          CTE_SCAN         │                             │      COMPARISON_JOIN      │
                             │    ────────────────────   │                             │    ────────────────────   │
                             │        CTE Index: 0       │                             │      Join Type: INNER     │
                             │                           │                             │   Conditions: (id = id)   ├──────────────┐
                             │                           │                             │                           │              │
                             │          ~2 rows          │                             │           ~1 row          │              │
                             └───────────────────────────┘                             └─────────────┬─────────────┘              │
                                                                                       │          SEQ_SCAN         ││         DELIM_GET         │
                                                                                       │           Table:          ││                           │
                                                                                       │     memory.main.other     ││                           │
                                                                                       │   Type: Sequential Scan   ││                           │
                                                                                       │          ~3 rows          ││           ~1 row          │
                                                                                       └───────────────────────────┘└───────────────────────────┘
                                                                                       ```

**Comment by EtgarDev** [comment]
@lnkuiper 
And blocking the cascade during runtime is indeed not enough but I thought it will be a bit odd to block it only from a single angle.
It's easy to block it from the angle of:
```
"check if the created trigger writes to a table that has its own `FOR EACH ROW` trigger"
```
but it will be harder to:
```
"check if the created trigger is set on a table which there are triggers on other tables that write into this table"
```
we could make some mechanism of storing some metadata on the target tables about triggers that are being created but it would be a bit out of scope for this PR.
So I chose to leave it to fail during runtime (it's still caught, just not at `CREATE` time).

WDYT? Should we block it from only one angle?

**Comment by lnkuiper** [comment]
Thanks for the changes!

> WDYT? Should we block it from only one angle?

I think blocking from one angle is cleaner, but I'm also OK with fixing that in a follow-up PR. Not a blocker for this PR.

This PR is getting close to ready to go from my side, I have two things that I think need to be fixed. At `trigger_for_each_row.test:216` there is a FIXME along with a test skip. However, this returns incorrect results (your comment correctly says we need per-row iterative trigger execution for that). I think we should reject `UPDATE` statements in row-trigger bodies in the binder for now rather than returning incorrect results.

There is also a subtle error here:
```sql
# Replacing a statement trigger with a same-name row trigger should not count as
# mixing, because the final catalog contains only the replacement trigger.
statement ok
CREATE TABLE replace_t(id INTEGER);

statement ok
CREATE TABLE replace_log(id INTEGER);

statement ok
CREATE TRIGGER replace_trg AFTER INSERT ON replace_t FOR EACH STATEMENT
    INSERT INTO replace_log VALUES (0);

statement ok
CREATE OR REPLACE TRIGGER replace_trg AFTER INSERT ON replace_t FOR EACH ROW
    INSERT INTO replace_log VALUES (NEW.id);

statement ok
INSERT INTO replace_t VALUES (1);

query I
SELECT id FROM replace_log ORDER BY id;
----
1
```
This throws an error, but it replaces the trigger that causes the error, so it should succeed.

EtgarDev:
@lnkuiper thanks :)
so I made some changes:
1. I wrote the single-angle defensive approach, and also added a skipped test case for the reversible angle.
2. good catch! I fixed the case by also comparing the names and checking whether we are in a REPLACE mode

EDIT - Ohh and now I'm also blocking UPDATE triggers.
However, there are still UPDATE paths that work, so I'm wondering if we should block the whole `UPDATE` path because of this.
I'll check how easy will it be just to block this accumulative case.

EtgarDev:
@lnkuiper done :)
Chose to block `UPDATE` triggers for now

lnkuiper (approved):
Excellent! Thanks for all the changes

PR #23286, opened 2026-06-15
Initial implementation of `variant_type` processing function

sebastiaan-dev:
This PR resolves https://github.com/duckdblabs/duckdb-internal/issues/9679, introducing an initial implementation for `variant_type`. This function takes a `VARIANT`-type object as input and an optional path. It returns the type of the supplied `variant`. If `path` is specified, return the type at the given `path`. If `path` is a list, the result will be a list of types. The return type is formatted as `VARCHAR`.

This implementation differs from `json_type` in that a `VARIANT`-type has higher type resolution, which we propagate in the result.

### Examples

```sql
variant_type({'a': [1, 2]}::VARIANT)
> OBJECT

variant_type({'a': [1, 2]}::VARIANT, 'a')
> ARRAY

variant_type({'a': 1}::VARIANT, ['a', 'a'])
> [INT32, INT32]
```

### Interface

`variant_type(variant[, path])`

#### Parameters

- `path` is optional and of type `VARCHAR` or `VARCHAR[]`

### Refactor

Added a function set constructor to reduce duplication across path based variant processing functions. In addition, we have moved the existing variant processing functions over to use an executor which internally uses a variant iterator. This move also means we no longer have an intermediate collection step.

### Follow-up work


Tishj (approved):
Thanks, looks good to me
Small nit that you could use `test_all_types()` instead of defining the list of different types manually.

Also perhaps we can replace `variant_typeof` with a macro. Only difference is that typeof puts the keys in an object and the decimal width+scale in the decimal.
I wonder if that can be expressed with a CASE expr

^ that will need a function to get the decimal width+scale of a variant though

Tishj (approved):
Looks great, the variant iterator cleanly integrates into the existing code, nice!

src/function/scalar/variant/variant_keys.cpp
```diff
@@ -1,119 +1,40 @@
...
 		return;
 	}
 
-	const auto list_entries = FlatVector::GetData<const list_entry_t>(key_ids);
-	const auto &child = ListVector::GetChild(key_ids);
-	const auto key_ids_data = FlatVector::GetData<const idx_t>(child);
-
-	auto &entry = list_entries[row_idx];
-	auto row_writer = result_writer.WriteList(entry.length);
+	vector<string_t> keys;
```
lnkuiper: Could we avoid materializing the keys a second time here? `GetObjectChildren(LEXICOGRAPHIC)` already materializes and sorts the entries, so this adds another `vector<string_t>` allocation and another pass for every object row.

I think we can write into the result directly:

```cpp
auto keys_writer = list_writer.WriteDynamicList();
for (const auto &entry : node->GetObjectChildren(VariantIterationOrder::LEXICOGRAPHIC)) {
	keys_writer.WriteElement().WriteValue(entry.key);
}
```

I ran a quick single-threaded benchmark over 200K unshredded rows containing objects with 32 keys:

```sql
-- Run using the DuckDB CLI
.timer on

PRAGMA threads = 1;

CREATE TEMP TABLE t AS
SELECT {
	'k31': i,
	'k30': i,
	'k29': i,
	'k28': i,
	'k27': i,
	'k26': i,
	'k25': i,
	'k24': i,
	'k23': i,
	'k22': i,
	'k21': i,
	'k20': i,
	'k19': i,
	'k18': i,
	'k17': i,
	'k16': i,
	'k15': i,
	'k14': i,
	'k13': i,
	'k12': i,
	'k11': i,
	'k10': i,
	'k09': i,
	'k08': i,
	'k07': i,
	'k06': i,
	'k05': i,
	'k04': i,
	'k03': i,
	'k02': i,
	'k01': i,
	'k00': i
}::VARIANT AS v
FROM range(200000) r(i);

SELECT sum(length(variant_keys(v))) FROM t;
SELECT sum(length(variant_keys(v))) FROM t;
SELECT sum(length(variant_keys(v))) FROM t;
```

Results:

- Base `71f0c7a0bf`: 46 / 44 / 44 ms
- PR `74b68348c2`: 123 / 113 / 114 ms

The benchmark covers the whole new path, including the lexicographic sorting, so the slowdown isn't all caused by this temporary vector. Still, this allocation and extra pass look easy to avoid.

sebastiaan-dev: I was not aware of the dynamic list, good catch, changed.

lnkuiper:
Thanks for the changes! Looks cleaner now. Performance is still worse than before (96ms vs 44ms) but good enough for me, looks clean

PR #23287, opened 2026-06-15
Optimize rollup / cube / grouping sets queries by folding them into CTEs to allow for cascading aggregation

Mytherin:
ROLLUP / CUBE / Grouping sets are currently implemented by just using multiple hash tables. However, they can be often implemented using a cascading aggregation. Instead of using one hash table per grouping set, we can compute the "widest" hash table first, and then use that hash table as input for the second aggregation, etc.

For example, if we have a query like so:

```sql
SELECT l_returnflag, l_linestatus, SUM(l_quantity)
FROM lineitem
GROUP BY CUBE(l_returnflag, l_linestatus)
ORDER BY ALL;
```

We are really computing four grouping sets:

```
l_returnflag, l_linestatus
l_returnflag
l_linestatus
(empty) // i.e. ungrouped
```

By doing a cascading grouping we can save a lot of time - the first hash table `l_returnflag, l_linestatus` only has four rows. So instead of pushing every single row of `lineitem` into four different hash tables, we can push it into a single hash table, and re-aggregate over that hash table.

Using materialized CTEs and aggregate states, we have all of the parts necessary to enable this. The above query can effectively be rewritten into:

```sql
with 
	group_by_returnflag_linestatus as (
		select l_returnflag, l_linestatus, sum(l_quantity) export_state l_quantity_state from lineitem group by l_returnflag, l_linestatus
	),
	group_by_returnflag as (
		select l_returnflag, NULL AS l_linestatus, combine_aggr(l_quantity_state) l_quantity_state from group_by_returnflag_linestatus group by l_returnflag		
	),
	group_by_linestatus as (
		select NULL AS l_returnflag, l_linestatus, combine_aggr(l_quantity_state) l_quantity_state from group_by_returnflag_linestatus group by l_linestatus		
	),
	ungrouped as (
		select NULL AS l_returnflag, NULL AS l_linestatus, combine_aggr(l_quantity_state) l_quantity_state from group_by_returnflag
	)
SELECT l_returnflag, l_linestatus, finalize(l_quantity_state) AS l_quantity_sum FROM group_by_returnflag_linestatus
UNION ALL
SELECT l_returnflag, l_linestatus, finalize(l_quantity_state) AS l_quantity_sum FROM group_by_returnflag
UNION ALL
SELECT l_returnflag, l_linestatus, finalize(l_quantity_state) AS l_quantity_sum FROM group_by_linestatus
UNION ALL
SELECT l_returnflag, l_linestatus, finalize(l_quantity_state) AS l_quantity_sum FROM ungrouped;
```

Effectively we compute the first aggregation and export the aggregate state - and then "cascade" over the result of that aggregation. The result can then be obtained by finalizing the aggregate states that have been collected. 

In the above example, we get roughly a ~2x speed-up with `CUBE` (TPC-H SF100):

| Version | Timing (s) |
|---------|------------|
| v1.5    | 0.8s       |
| New     | 0.4s       |

CC @kryonix

### DIFF
--- a/src/optimizer/optimizer.cpp
+++ b/src/optimizer/optimizer.cpp
@@ -237,6 +237,12 @@ void Optimizer::RunBuiltInOptimizers() {
 		plan = deliminator.Optimize(std::move(plan));
 	});
 
+	// rewrite aggregates over multiple grouping sets (ROLLUP/CUBE/GROUPING SETS) into a cascade of aggregations
+	RunOptimizer(OptimizerType::GROUPING_SETS, [&]() {
+		GroupingSetsOptimizer grouping_sets_optimizer(*this);
+		grouping_sets_optimizer.VisitOperator(plan);
+	});
+
 	// try to inline CTEs instead of materialization
 	RunOptimizer(OptimizerType::CTE_INLINING, [&]() {
 		CTEInlining cte_inlining(*this);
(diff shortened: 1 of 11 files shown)


src/optimizer/optimizer.cpp
```diff
@@ -303,6 +304,12 @@ void Optimizer::RunBuiltInOptimizers() {
 		remove.VisitOperator(*plan);
 	});
 
+	// rewrite aggregates over multiple grouping sets (ROLLUP/CUBE/GROUPING SETS) into a cascade of aggregations
```
kryonix: The optimizer runs after CTE inlining is already done (twice). We may be able to inline some of the generated CTEs, omitting some materialization overhead.

kryonix: Had a look at the generated plan. This seems fine.

PR #23288, opened 2026-06-15
Pass `ParserOptions` in default macro generation

nicholasjng:
This avoids rebuilding a fresh `ParserCache` on every default macro instantiation, which happens on database instantiation with the JSON extension preloaded.

Adds a `ParserOptions` overload for both default macro generators, in addition to the existing ones.

--------------------

Hey! I'm currently building https://github.com/nicholasjng/ducky to explore how much performance I can get for ML workloads in Python from duckDB's C API paired with nanobind's zero-copy-via-DLPack array construction. 

By benchmarking against duckDB's own Python bindings, I noticed I'm significantly slower on connection creation compared to the last released `duckdb-python`. Profiling then lead me to repeated parser cache creation on JSON extension import.

This change reduces the time spent in a `ducky.connect()` call from ~11ms to ~6ms  as measured on my Apple M1 Pro.

The reason I'm hitting this is because I statically link `json` into my Python bindings module, since I haven't been able to figure out yet how to lazy-load duckDB extensions via the C API. From a quick survey of the other builtin extensions, it seems like JSON is the only one building the macros on load.

### DIFF
--- a/src/catalog/default/default_functions.cpp
+++ b/src/catalog/default/default_functions.cpp
@@ -4,6 +4,8 @@
 #include "duckdb/parser/statement/create_statement.hpp"
 #include "duckdb/catalog/catalog_entry/scalar_macro_catalog_entry.hpp"
 #include "duckdb/common/string_util.hpp"
+#include "duckdb/main/client_context.hpp"
+#include "duckdb/main/database.hpp"
 
 namespace duckdb {
 
@@ -222,11 +224,16 @@ static const DefaultMacro internal_macros[] = {
     {nullptr, nullptr, nullptr}};
 
 unique_ptr<CreateMacroInfo> DefaultFunctionGenerator::CreateInternalMacroInfo(const DefaultMacro &default_macro) {
+	return CreateInternalMacroInfo(default_macro, ParserOptions());
+}
+
+unique_ptr<CreateMacroInfo> DefaultFunctionGenerator::CreateInternalMacroInfo(const DefaultMacro &default_macro,
+                                                                              ParserOptions options) {
 	auto bind_info = make_uniq<CreateMacroInfo>(CatalogType::MACRO_ENTRY);
 	// Build a full CREATE MACRO statement and let the parser handle parameters, types, and defaults.
 	// macro_definition may contain multiple comma-separated overloads, e.g. "(x) AS x, (x, y) AS x+y".
 	auto sql = StringUtil::Format("CREATE MACRO __dummy__%s", default_macro.macro_definition);
-	Parser parser;
+	Parser parser(options);
 	parser.ParseQuery(sql);
 	D_ASSERT(parser.statements.size() == 1);
 	D_ASSERT(parser.statements[0]->type == StatementType::CREATE_STATEMENT);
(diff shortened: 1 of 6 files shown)


src/catalog/default/default_functions.cpp
```diff
@@ -271,7 +278,7 @@ DefaultFunctionGenerator::DefaultFunctionGenerator(Catalog &catalog, SchemaCatal
 
 unique_ptr<CatalogEntry> DefaultFunctionGenerator::CreateDefaultEntry(ClientContext &context,
                                                                       const Identifier &entry_name) {
-	auto info = GetDefaultFunction(schema.name, entry_name);
+	auto info = GetDefaultFunction(schema.name, entry_name, context.GetParserOptions());
```
Dtenwolde: This is risky when the session has altered parser settings such as `integer_division`. For default macros we only want to reuse the `ParserCache` from the `ClientContext`. The macro definitions themselves should still be parsed with default `ParserOptions`, which is what happened before this change I think.

```cpp
  ParserOptions options;
  options.parser_cache = &context.db->GetParserCache();
  auto info = GetDefaultFunction(schema.name, entry_name, options);
```

An example where we change the output (with current changes): 
```sql 
memory D   SET integer_division = true;
memory D   SELECT weighted_avg(v, w)
           FROM (VALUES (1, 1), (2, 2)) t(v, w);
┌────────────────────┐
│ weighted_avg(v, w) │
│       int128       │
├────────────────────┤
│                  1 │
└────────────────────┘
```
This result happens because the `/` inside the `weighted_avg` macro is parsed using the `integer_division` setting we just set. I think for built-in macros, we shouldn't be doing that and stick to the default settings.

The expected result for reference: 
```sql
memory D set integer_division = true;
memory D SELECT weighted_avg(v, w)
                    FROM (VALUES (1, 1), (2, 2)) t(v, w);
┌────────────────────┐
│ weighted_avg(v, w) │
│       double       │
├────────────────────┤
│ 1.6666666666666667 │
└────────────────────┘
```

src/catalog/default/default_table_functions.cpp
```diff
@@ -130,18 +136,18 @@ DefaultTableFunctionGenerator::CreateTableMacroInfo(const DefaultTableMacro &def
...
+			return DefaultTableFunctionGenerator::CreateTableMacroInfo(internal_table_macros[index], options);
 		}
 	}
 	return nullptr;
 }
 
 unique_ptr<CatalogEntry> DefaultTableFunctionGenerator::CreateDefaultEntry(ClientContext &context,
                                                                            const Identifier &entry_name) {
-	auto info = GetDefaultTableFunction(schema.name, entry_name);
+	auto info = GetDefaultTableFunction(schema.name, entry_name, context.GetParserOptions());
```
Dtenwolde: Same goes for this line, same fix.

Dtenwolde (changes requested):
Thanks for the PR! One comment regarding how we should retrieve the `ParserCache`, but other than that looks good :)

src/catalog/default/default_functions.cpp
nicholasjng: Done. Would you like this example as a regression test?

src/catalog/default/default_table_functions.cpp
nicholasjng: Done.

src/catalog/default/default_functions.cpp
Dtenwolde: > Would you like this example as a regression test?

Yes please! :)

nicholasjng: Added in the second commit.

Dtenwolde:
Thanks! Please undraft your PR to run CI :)

nicholasjng:
Okay, why did the bot move it to draft again? I missed that.

Dtenwolde:
It does that automatically for non-DuckLabs members whenever a new commit is added after the PR is opened

lnkuiper:
I've restarted CI. Could you have a look at the failing tidy check?

nicholasjng:
Forgot to include the database header, fixed.

Mytherin:
Thanks!

PR #23289, opened 2026-06-15
Dead node counting fixes

artjomPlaunov:
TL;DR: Fix the remaining (hopefully) dead node counting issues. Fix an existing bug that confused unloadability with dead node count, and also create a new boolean flag to replace using eviction_seq_number as a "has a live node on the queue" check (although eviction seq num is still used for staleness). This also restores eviction sequence number to be always increasing, instead of reset, since I think having a reset fixed another bug locally while possibly introducing more counting issues, but since we have an explicit "live" boolean, that shouldn't be an issue anymore. 

Trying to get rid of the TSAN suppression, need to see if CI still fails + maybe run this on linux to be sure. 

--------------------------------------------------

More notes (Sorry a bit long, just trying to write down all the logic considered here). The basic goal is to only increment a dead node exactly at the point where that node transitions from live -> dead, and to have a matching decrement when that same node is dequeued. 

One localized bug present in the previous version of the code had to do with BufferEvictionNode::TryGetBlockMemory: there were three checks in there wrapped into one return value that would signal if we should decrement the dead node count: 1) the block memory is still available to lock, 2) the sequence number is not stale and 3) the block is unloadable. 

However, if we define a dead node (i.e. the IsDeadNode() function), then we should only be checking the first two conditions. I tried to fix this locally, but then this lead to some over-counting issues instead. I think the real fix here is to separate out the eviction queue sequence number (used to check for staleness on a per node basis, only one condition of being "dead"), and a separate boolean, "has_queue_entry" that tells us, for a given BlockMemory, if it has a corresponding "live", i.e. "not dead" entry on the queue. Previously, sequence number was used as a proxy for this. 

### Increment sites 

for dead nodes (under the two conditions): this is defining the places where a previous node transitions from live -> dead. 

- AddToEvictionQueue : if there is an existing live queue entry (new boolean flag), we increment the dead node count (i.e. previous live node becomes stale, so it transitions from live -> dead). 
- BlockMemory destructor: again, check the flag + the previous TINY_BUFFER condition, and increment if we have a previous live node on the queue, again, this is a transition from live -> dead). 

### Decrement Sites

- IterateUnloadableBlocks: we see a stale sequence number (so it is dead by definition) - decrement
- IterateUnloadableBlocks: memory is expired (again, dead by definition, so we decrement)
- PurgeIteration applies the above conditions, but in bulk. 

### NOP scenario

Previously (in the bug mentioned above for case (3)), we mixed up "unloadability" with "dead", so if it is in fact alive, but unloadable, this is a NOP for the dead node count. The count should only change when we are either introducing a dead node, or removing a dead node. 

I needed to add a previous boolean flag because cases kept coming up where checking the sequence number was not correct as a proxy for "presence of a not-dead node" i.e. a node that "has a fresh seq num + memory is not expired", so I can't go wrong by just tracking that with a boolean. Although I am threading a lock through the AddToEvictionQueue API now, I think it is not adding any overhead and just making it explicit this is in a scenario where a) we have a lock to thread through or b) there is no contention to grab it (i.e., in ConvertToPersistent). Out of curiosity, I got rid of the TSAN suppression, but I have to try running this on linux to see if it works. 

### has_queue_entry tracking

Again, this is separate from "can this block be unloaded". For example, we can have a live node on the queue that cannot be unloaded, but it's fine to remove it from the queue. This is just maintaining a boolean to tell if we have a node that "has a fresh sequence number ^ memory is not expired"

- set to false on construction of BlockMemory
- Set to true when a new live queue node is created
- Set to false when a live node is dequeued

Some extra edge cases:

- Unfortunately, I think there is still a possible transient mis-ordering between an increment and a decrement, i.e. where a decrement can come before a matching increment; however, this should be a very small window, and the end result should still be correct. If a consumer of the concurrent queue fails on the weak_ptr::lock(), this could happen before ~BlockMemory runs its decrement, so again, if this happens it should be a very tiny window where the count is inconsistent. I think this can be a follow up, but is likely fine to keep it as it is. 
- In AddToEvictionQueue, in the scenario where a previous live node becomes stale, we bump the dead node count before generating a new eviction sequence number. This is if purging can run (without a lock), and observe a stale sequence number, thus leading to another mis-ordering between decrement and increment.


test/sql/storage/test_eviction_queue_dead_nodes_underflow.test
```diff
@@ -0,0 +1,102 @@
+# name: test/sql/storage/test_eviction_queue_dead_nodes_underflow.test
```
lnkuiper, line 1: This should probably be a `.test_slow` since it's doing somewhat heavy stuff like sorting 10m rows. This might time out in the nightlies

lnkuiper (approved):
Thanks! Looks great :) I have one comment below but that's non-blocking, can be fixed later

PR #23290, opened 2026-06-15
Add Parquet ASYNC I/O for local files

pdet:
This PR adds ASYNC I/O for parquet when reading local files, by using the same prefetch code locally. Since both remote and local files go through the same code path we removed the `prefetch_all_parquet_files` test configuration.

The main thing to adapt the prefetch code-base to deal with local files was to make sure the cost model that calculates gap sizes was also reasonable for local files. We do that by changing `CachingFileHandle` to time reads and record latency and bandwidth with the same strategy as httpfs, and we reuse the `TryGetNetworkThroughput` interface. We calculate latency and bandwidth with a least-squares fit of the line `total_time = intercept + slope * bytes`over reads of different sizes. The slope is the extra seconds per byte,  and bandwidth is 1/slope. We also clamp the gap between GAP_MIN (4 KiB) and GAP_MAX (32 MiB) to keep a bad fit from producing an extreme gap.
  
Since the gap size is now flexible for local files, we also added an option `parquet_prefetch_column_gap` where the column gap can be set, to ensure existing tests would still test specific cases and also to allow users to set these manually if they so wish.
 

## Benchmark
For the benchmark, we generated 16 files of ~512Mb each, with 16 columns, all doubles. Each file has 40 row groups (every 100,000 rows), and each file has 4,000,000 rows.

| Cache | Main   | Branch | Improvement |
|--------|--------|--------|-------------|
| Warm   | 0.44 s | 0.42 s | 4.5% |
| Cold   | 6.9 s  | 4.2 s  | 39.1% |

For warm we do 2 warmup and capture the median of 5 runs after. For cold we do 5 runs with cache eviction (`hdiutil detach /Volumes/prefetchbench -quiet && hdiutil attach /tmp/prefetchbench.sparseimage -quiet`)


extension/parquet/parquet_extension.cpp
```diff
@@ -1001,8 +1001,12 @@ static void LoadInternal(ExtensionLoader &loader) {
...
 	                          LogicalType::BOOLEAN, Value(false));
 	config.AddExtensionOption("prefetch_all_parquet_files",
-	                          "Use the prefetching mechanism for all types of parquet files", LogicalType::BOOLEAN,
-	                          Value(false));
+	                          "(deprecated) Parquet files are now always prefetched, this setting has no effect",
+	                          LogicalType::BOOLEAN, Value(false));
+	config.AddExtensionOption(
+	    "parquet_prefetch_column_gap",
+	    "Byte gap under which Parquet prefetch I/O ranges are coalesced (0 lets the cost model adapt it)",
+	    LogicalType::UBIGINT, Value::UBIGINT(0));
```
Mytherin: `0` is imo the wrong sentinel number to use here since that's a valid number that you might want to set (in order to disable prefetch gaps entirely). Maybe this can be `NULL` to disable instead?

src/include/duckdb/main/settings.hpp
```diff
@@ -1148,6 +1148,16 @@ struct ForceCompressionSetting {
 	static void OnSet(SettingCallbackInfo &info, Value &input);
 };
 
+struct ForceFileCacheSetting {
+	using RETURN_TYPE = bool;
+	static constexpr const char *Name = "force_file_cache";
```
Mytherin: Maybe `cache_local_files`?

extension/parquet/parquet_reader.cpp
```diff
@@ -1693,7 +1706,11 @@ ParquetPrefetchStrategy ParquetReader::ColumnWisePrefetch(ParquetReaderScanState
 }
 
 AsyncResult ParquetReader::Scan(ClientContext &context, ParquetReaderScanState &state, DataChunk &result) {
-	result.Reset();
+	if (!state.filter_done) {
```
Mytherin: It seems that there's a bunch of booleans here that together encode a state of the reader - maybe this can just be a single enum that we switch on?

Mytherin (commented):
Thanks - looks good, left some comments

src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -256,9 +257,14 @@ FileBufferHandleGroup CachingFileHandle::Read(const idx_t nr_bytes, const idx_t
...
 	}
 
-	if (!external_file_cache.IsEnabled()) {
+	// Uncached files are read directly into one contiguous buffer, skipping the per block syscalls and copies
+	if (!external_file_cache.IsEnabled() || !external_file_cache.ShouldCacheFile(path.path)) {
 		auto buf = external_file_cache.GetBufferManager().Allocate(MemoryTag::EXTERNAL_FILE_CACHE, nr_bytes);
-		GetFileHandle().Read(context, buf.GetDataMutable(), nr_bytes, location);
+		auto &handle = GetFileHandle();
+		const auto read_start = steady_clock::now();
+		handle.Read(context, buf.GetDataMutable(), nr_bytes, location);
```
Mytherin: Maybe this `Read` call should be wrapped in a helper method that adds the timing?

I also wonder - why are we only recording the read throughput here, and not for all `Read` calls made on the file?

Mytherin:
Thanks!

PR #23293, opened 2026-06-15
Fix time duration calculation

dentiny:
Followup fix for https://github.com/duckdb/duckdb-httpfs/pull/339

During the fix, I found quite a few places in our places are using system clock difference to calculate interval, which could lead to INTERNAL exception of cast failure.

Key changes summary:
- Use monotonic clock to replace system clock when interval calculation involved
- System clock util function is kept since it's still used
- Also replace high resolution clock usage with monotonic clock, when interval needed
  + [monotonic clock](https://en.cppreference.com/cpp/chrono/high_resolution_clock) doesn't guarantee implementation or monotonic increasing property, so it could throw exception
  + In certain platforms, high resolution clock is just an alias for monotonic clock


Mytherin:
Thanks - given this seems to affect many places, maybe it makes more sense to just change `Timestamp::GetCurrentTimestamp` to use a monotonic clock over a system clock?

Mytherin:
Ah maybe that's not possible. I wonder if it makes sense for `Timestamp::GetMonotonicTimestamp` to return a `timestamp_t` at all given it's not really a timestamp, at least according to Claude:

>   - system_clock's epoch is defined (since C++20) as the Unix epoch, 1970-01-01 UTC.
>   - steady_clock's epoch is unspecified — typically system boot time, but the standard makes no guarantee. There's no to_time_t for it, by design.

>   So you can't ask "what date was this steady_clock reading?" The value steady_clock::now() might be "3.2 billion nanoseconds since boot," which says nothing about UTC.

Maybe this should be separate from `Timestamp`, and be e.g. `Profiler::Tick()` or so?

dentiny:
> given this seems to affect many places, maybe it makes more sense to just change `Timestamp::GetCurrentTimestamp` to use a monotonic clock over a system clock?

Yeah as I mentioned in the PR description, we have a few places still relying on system clock.

dentiny:
> I wonder if it makes sense for Timestamp::GetMonotonicTimestamp to return a timestamp_t at all given it's not really a timestamp

This is something I considered as well, two options on top of my head
- Reuse the existing timestamp type for both system clock and monotonic clock
- Add a new class to represent monotonic clock like `MonotonicTimestamp`

The reason I choice (1) is because 
- (1) and (2) have nothing in difference -- both are just a wrapper around an integer, and they support same set of operations, don't feel strong need to have a new class
- Less code change

The only concern for this approach is whether people could mix usage for two timestamp types, but I think it's very low risk, because usually when people calculate timestamp difference, they do
```cpp
auto start = GetMonotonicTime();
do_something();
auto end = GetMonotonicTime(); // end timestamp is calculated right away, only 2LOC after
```
, which is hard to get it wrong.

There're a few possible followups we could do, if we find it necessary
- Assign timestamp type (monotonic vs system), so operator- is disallowed across different types
- Better timestamp difference util (i.e., have a separate timer class, similar to what we have for profiler)

Mytherin:
I prefer (2) - yes both options are just wrappers around an integer, but they represent different concepts, so imo they should be different classes. You can already see places in the code where this is currently confusing as the two are mixed:

```cpp
	// System clock start timestamp
	timestamp_t request_system_start;
	// Monotonic clock start and end timestamp
	timestamp_t request_monotonic_start;
	timestamp_t request_monotonic_end;
```

It would be easy to write code to take `request_monotonic_start` and use it like a timestamp right now.

I think the names "monotonic" and "system" aren't descriptive names either - in order to understand them the reader needs to understand the intricacies of C++ timers because the names come from those (`system` comes from `system_clock` and `monotonic` comes from a property of `steady_clock`). 

I would vote for adding a new class - `TimePoint` (taken from the C++ standard library `time_point`) - that holds a `time_point<steady_clock>`. It should then have clear methods - e.g.:

```cpp
auto start = TimePoint::Tick();
auto end = TimePoint::Tick();

TimePoint::ElapsedMillis(start, end);
TimePoint::ElapsedMicros(start, end);
TimePoint::ElapsedNanos(start, end);
```

We can also integrate that into the `Profiler` class which currently has more or less these properties and methods as well.

dentiny:
> I would vote for adding a new class - TimePoint (taken from the C++ standard library time_point)

Sure! In the latest commit, I made the new time class and update all monotonic clock usages.

test/sql/function/generic/test_sleep.test
```diff
@@ -58,12 +58,6 @@ SELECT sleep_ms(-10)
 ----
 NULL
 
-# Test sleep with overflowing duration (fuzzer duckdb-fuzzer#4570)
```
dentiny, line 61: Before this change, we use `timestamp_t` to represent system clock which stores timestamp in microseconds, naturally INT64_MAX overflows. In the updated impl everything is done in milliseconds and no multiplication / increment operation involved.

a27981731-rgb:
One shoe in 


Sent from Yahoo Mail for iPhone


On Saturday, June 27, 2026, 7:44 PM, dentiny ***@***.***> wrote:




@dentiny commented on this pull request.

In test/sql/function/generic/test_sleep.test:
> @@ -58,12 +58,6 @@ SELECT sleep_ms(-10)
 ----
 NULL
 
-# Test sleep with overflowing duration (fuzzer duckdb-fuzzer#4570)

Before this change, we use timestamp_t to represent system clock which stores timestamp in microseconds, naturally INT64_MAX overflows. In the updated impl everything is done in milliseconds and no multiplication / increment operation involved.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

a27981731-rgb:
Wtf is this a job yalll are fuckjng weird codes and shit good god do u know how to brush your hair by yourself


Sent from Yahoo Mail for iPhone


On Saturday, June 27, 2026, 10:56 PM, ***@***.*** wrote:

One shoe in 


Sent from Yahoo Mail for iPhone


On Saturday, June 27, 2026, 7:44 PM, dentiny ***@***.***> wrote:




@dentiny commented on this pull request.

In test/sql/function/generic/test_sleep.test:
> @@ -58,12 +58,6 @@ SELECT sleep_ms(-10)
 ----
 NULL
 
-# Test sleep with overflowing duration (fuzzer duckdb-fuzzer#4570)

Before this change, we use timestamp_t to represent system clock which stores timestamp in microseconds, naturally INT64_MAX overflows. In the updated impl everything is done in milliseconds and no multiplication / increment operation involved.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

test/sql/function/generic/test_sleep.test
kghzjk4hdj-arch: Speak in normal talk not a robot plz

a27981731-rgb:
I hope ur ok


Sent from Yahoo Mail for iPhone


On Saturday, June 27, 2026, 11:00 PM, kghzjk4hdj-arch ***@***.***> wrote:




@kghzjk4hdj-arch commented on this pull request.

In test/sql/function/generic/test_sleep.test:
> @@ -58,12 +58,6 @@ SELECT sleep_ms(-10)
 ----
 NULL
 
-# Test sleep with overflowing duration (fuzzer duckdb-fuzzer#4570)

Speak in normal talk not a robot plz

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

Mytherin:
Thanks! Looks great

dentiny:
```sh
python3 scripts/verify_enum_integrity.py src/include/duckdb.h
Successfully verified the integrity of enum DUCKDB_TYPE (43 entries)
Successfully verified the integrity of enum duckdb_state (2 entries)
Successfully verified the integrity of enum duckdb_pending_state (4 entries)
Successfully verified the integrity of enum duckdb_result_type (4 entries)
Successfully verified the integrity of enum duckdb_statement_type (28 entries)
Successfully verified the integrity of enum duckdb_error_type (43 entries)
Successfully verified the integrity of enum duckdb_cast_mode (2 entries)
Successfully verified the integrity of enum duckdb_file_flag (6 entries)
Successfully verified the integrity of enum duckdb_config_option_scope (4 entries)
Successfully verified the integrity of enum duckdb_catalog_entry_type (10 entries)
ERROR: Could not find a version that satisfies the requirement black==24.* (from versions: none)
ERROR: No matching distribution found for black==24.*
make: *** [Makefile:713: format_venv] Error 1
```
The PR gets removed from merge queue, while the error message doesn't looks like my PR's issue. 
Maybe @Mytherin could help me re-merge it? Thank you!

kghzjk4hdj-arch:
Grow the fuck up pussys

PR #23241, opened 2026-06-12
initialize shredded geometry base stats correctly

Maxxen:
Fixes an issue found internally

PR #23295, opened 2026-06-15
Fix more geom stats

Maxxen:
Followup to https://github.com/duckdb/duckdb/pull/23241/changes


carlopi:
Also `test/configs/verify_fetch_row.json` needs to skip `test/sql/types/geo/geometry_shred_more.test`

carlopi (commented):
Logically this already passed CI (but a skip was missing).
Unsure if we can consider just merging this?

PR #23296, opened 2026-06-16
Step 5: Make Mark to Semi join conversion more stable

kryonix:
In this PR, I try to make the full decorrelation cases more stable and robust, while allowing more cases. There were a few areas in the optimizer that prevented the system from finding good join types and/or join orders which can regress performance:

- For correlated mark joins, `convert_mark_to_semi` was statically set to false. However, this was more of a guardrail to prevent the system from establishing semi-optimal join orders, because we were missing some heuristics in the `build_probe_side` optimizer for semi joins. This has been added now.
- Decorrelated plans contain many `... IS NOT DISTINCT ...` predicates. Problem is, this can prevent adding dynamic filters. `pushdown_mark_join` now tries to reduce them to regular equality predicates, if the join has been rewritten to a semi join.
- The `delim_cte_rewriter` technically did the right thing for full decorrelation cases. However, it is not always desireable to fully decorrelate plans, instead of stopping early. This is a bit hard to decide without cost/cardinality information (maybe we should change that at some point). Therefore, it is a bit conservative when full decorrelation is used. Still, we allow it for more cases now (e. g. TPC-H Q21).

For TPC-H Q21, this PR improves performance by about 30-40%, for Q4 by about 25%, while avoiding to regress Q22 and Q17.

### DIFF
--- a/src/include/duckdb/planner/subquery/delim_join_cte_rewriter.hpp
+++ b/src/include/duckdb/planner/subquery/delim_join_cte_rewriter.hpp
@@ -23,9 +23,9 @@ private:
 
 	void Rewrite(unique_ptr<LogicalOperator> &plan);
 	void RewriteDelimJoinsToCTEs(unique_ptr<LogicalOperator> &plan, LogicalOperator &rewrite_root,
-	                             bool null_rejecting_filter_above = false);
+	                             bool null_rejecting_filter_above = false, bool preserve_evidence_side = false);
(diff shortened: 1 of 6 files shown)


lnkuiper:
Failure is unrelated, I'll look into that

lnkuiper:
Looks great, thanks!

EDIT: I can't add to the merge queue until CI is fixed, maybe @Mytherin can do that. I'll work on fixing CI :)

lnkuiper (approved):

Mytherin:
Thanks!

PR #23298, opened 2026-06-16
Handle exceptions thrown while recovering from a failed commit

Huey97:
Hi Mark @Mytherin , could you help take a look at this when you get a chance? Thanks!

## Problem

When `temp_directory` is configured with a size limit, undo buffer blocks
managed by BufferManager can be evicted to disk. In the noexcept functions
`DuckTransaction::Commit()`, `DuckTransaction::WriteToWAL()`, and
`DuckTransaction::Cleanup()`, the `UndoBuffer::IterateEntries()` method
calls `Pin(block)` to reload evicted blocks. Under memory pressure, `Pin()`
can throw an OOM exception. Since these functions are `noexcept`, the
exception triggers `std::terminate`, crashing the process.

The crash path:
1. Transaction commits → undo buffer blocks created
2. Memory pressure evicts undo blocks to temp directory
3. Another thread triggers Cleanup via cleanup_queue (or Commit/RevertCommit)
4. `IterateEntries()` → `Pin(block)` → OOM exception
5. Exception escapes noexcept → `std::terminate` → SIGABRT

## Fix

Pre-pin all undo buffer blocks into memory **before** entering the critical
section. Once pinned, blocks cannot be evicted, so subsequent `Pin()` calls
inside `IterateEntries()` are guaranteed to succeed.

If pre-pinning itself fails (OOM), no state has been modified yet:
- `WriteToWAL()` / `Commit()`: return `ErrorData` safely
- `Cleanup()`: silently skip (caller is noexcept, cleanup is GC — skipping
  means old versions linger but system stays consistent)

## Changes

- **`undo_buffer.hpp/cpp`**: Add `PinAllBlocks()` — iterates all undo buffer
  entries and pins their blocks, returning RAII handles that keep them pinned
- **`duck_transaction.cpp`**: Call `PinAllBlocks()` before the critical section
  in `WriteToWAL()`, `Commit()`, and `Cleanup()`

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -314,6 +314,20 @@
         "default_value": "0"
     },
     {
+        "name": "debug_force_commit_failure",
+        "description": "DEBUG SETTING: force transaction commit to fail after the undo buffer has been committed, used for testing commit error recovery",
+        "type": "BOOLEAN",
+        "default_scope": "global",
+        "default_value": "false"
+    },
+    {
+        "name": "debug_force_commit_revert_failure",
+        "description": "DEBUG SETTING: force RevertCommit to fail while recovering from a commit failure, used for testing",
+        "type": "BOOLEAN",
+        "default_scope": "global",
+        "default_value": "false"
+    },
+    {
         "name": "debug_force_external",
         "description": "DEBUG SETTING: force out-of-core computation for operators that support it, used for testing",
         "type": "BOOLEAN",
--- a/src/include/duckdb/main/settings.hpp
+++ b/src/include/duckdb/main/settings.hpp
@@ -531,6 +531,28 @@ struct DebugEvictionQueueSleepMicroSecondsSetting {
 	static constexpr idx_t SettingIndex = NEXT_SETTING_INDEX();
 };
 
+struct DebugForceCommitFailureSetting {
+	using RETURN_TYPE = bool;
+	static constexpr const char *Name = "debug_force_commit_failure";
+	static constexpr const char *Description = "DEBUG SETTING: force transaction commit to fail after the undo buffer "
+	                                           "has been committed, used for testing commit error recovery";
+	static constexpr const char *InputType = "BOOLEAN";
+	static constexpr const char *DefaultValue = "false";
+	static constexpr SettingScopeTarget Scope = SettingScopeTarget::GLOBAL_DEFAULT;
+	static constexpr idx_t SettingIndex = NEXT_SETTING_INDEX();
+};
+
+struct DebugForceCommitRevertFailureSetting {
+	using RETURN_TYPE = bool;
+	static constexpr const char *Name = "debug_force_commit_revert_failure";
+	static constexpr const char *Description =
+	    "DEBUG SETTING: force RevertCommit to fail while recovering from a commit failure, used for testing";
+	static constexpr const char *InputType = "BOOLEAN";
+	static constexpr const char *DefaultValue = "false";
+	static constexpr SettingScopeTarget Scope = SettingScopeTarget::GLOBAL_DEFAULT;
+	static constexpr idx_t SettingIndex = NEXT_SETTING_INDEX();
+};
+
 struct DebugForceExternalSetting {
 	using RETURN_TYPE = bool;
 	static constexpr const char *Name = "debug_force_external";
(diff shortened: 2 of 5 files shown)


Mytherin:
Thanks - but I don't think pre-pinning is the correct solution here. We should recover from the exception instead.

Huey97:
Thanks Mark — agreed, I'll drop the pre-pinning approach and recover from the exception instead.

While looking at how to do that, I noticed `WriteToWAL()` already handles this exact concern: the main body is in one `try/catch` that only records the error, and `RevertCommit()` is then called in a *separate* guarded block, with the comment:

```cpp
	ErrorData error_data;
	try {
		...
		storage->Commit(commit_state.get());
		...
	} catch (std::exception &ex) {
		// Call RevertCommit() outside this try-catch as it itself may throw
		error_data = ErrorData(ex);
	}

	if (commit_state && error_data.HasError()) {
		try {
			commit_state->RevertCommit();
			commit_state.reset();
		} catch (std::exception &) {
			// Ignore this error. If we fail to RevertCommit(), just return the original exception
		}
	}

```

But `DuckTransaction::Commit()` is declared `noexcept` and yet calls the revert logic *inside* its catch block:

```cpp
} catch (std::exception &ex) {
    undo_buffer.RevertCommit(iterator_state, this->transaction_id);
    if (commit_state) {
        // if we have written to the WAL - truncate the WAL on failure
        commit_state->RevertCommit();
    }
    return ErrorData(ex);
}
```

Since `RevertCommit()` iterates the undo buffer and can call `Pin()`, under memory pressure it may throw again — and because we're already in the catch block of a `noexcept` function, that escapes and triggers `std::terminate`. This is exactly the crash path in this PR.

Is there a specific reason `Commit()` doesn't follow the same structure as `WriteToWAL()` here? If not, I'd propose restructuring `Commit()` the same way — record the error in the inner `try/catch`, then run `RevertCommit()` in a separate guarded block so a failing revert can't break the `noexcept` guarantee. 

Happy to send that change if it sounds right to you.

Mytherin:
I think it makes perfect sense to catch the `RevertCommit` failure - however, we can't just swallow the exception there. We need to invalidate the database if this happens, because the database will be left in an undefined state if this happens.

Mytherin:
The reason we can swallow the exception in the WAL case is that we will just not finish attaching the database (i.e. the database will not be left in an undefined state because it will not be attached at all).

Huey97:
Thanks for the clear explanation, Mark — that makes sense.

I've updated the PR accordingly:

- `Commit()` now follows the same structure as `WriteToWAL()`: the main body records the error in the inner `try/catch`, and `RevertCommit()` runs in a separate guarded block.
- In that guarded block, instead of swallowing the exception, I now **invalidate the database** if `RevertCommit()` fails — since at this point the database is live and would otherwise be left in an undefined state.

I also added two parameters to make this path testable (to deterministically force the failure during commit/revert so the invalidation behavior can be verified).

Could you take another look when you get a chance and let me know if this approach looks reasonable? Thanks a lot for the guidance!

Huey97:
Hi Mark @Mytherin , could you help take another look when you get a chance and let me know whether this approach looks reasonable? If it does, could you also approve/rerun the workflows so the tests can run? Thanks!

Mytherin:
Thanks for the updates - I think this approach looks good, can you just look at the CI failures?

Huey97:
Thanks Mark, sure. I’ll investigate the CI failures and update the PR accordingly.

Huey97:
Thanks Mark @Mytherin, I’ve pushed an update addressing your review:

1. I reworked `Commit()` so exceptions during commit recovery are handled safely. `Commit()` is `noexcept`, but the recovery path can also throw while reverting the undo buffer. The revert logic now runs in a separate guarded `try`/`catch`, mirroring the structure used by `WriteToWAL()`, so a failing `RevertCommit()` cannot escape the `noexcept` boundary. If the revert itself fails, the database is left in an undefined state and is invalidated through `ValidChecker`.
2. I fixed the CI failures from the new test by guarding it with `require skip_reload` and `require no_alternative_verify`.

Could you approve the pending workflow runs when you get a chance? Thanks!

Huey97:
Thanks Mark @Mytherin, all checks are passing now. Please let me know if you have any other comments or suggestions.

Mytherin:
Thanks!

PR #23301, opened 2026-06-16
Remove stale PR workflow

DinosL:
With the introduction of pr-tracker, the stale PR mirroring workflow is no longer necessary.

Removing the trigger as well here: https://github.com/duckdb/duckdb-workflow-trigger/pull/15


Mytherin:
Thanks!

PR #23302, opened 2026-06-16
Add ICU date_diff support for TIMESTAMPTZ_NS

jingdaws:
## Background

  This PR is a prerequisite for supporting Iceberg v3 `timestamptz_ns` in `duckdb-iceberg`, tracked in https://github.com/duckdb/duckdb-iceberg/issues/1019.

  `duckdb-iceberg` implements temporal partition transforms by building regular DuckDB scalar expressions, rather than doing custom per-row timestamp logic inside the extension. For example, the current path uses `GetDateDiffFunction(...)` to
  bind expressions such as:

  ```sql
  date_diff('year', ..., source_column)
  date_diff('month', ..., source_column)
  date_diff('day', ..., source_column)
  date_diff('hour', ..., source_column)
  ```

  This works for the timestamp types that DuckDB already supports in date_diff, but it does not work for TIMESTAMPTZ_NS. DuckDB already has the TIMESTAMPTZ_NS logical type, but the ICU extension currently only registers date_diff(VARCHAR,
  TIMESTAMPTZ, TIMESTAMPTZ), not date_diff(VARCHAR, TIMESTAMPTZ_NS, TIMESTAMPTZ_NS).

  As a result, duckdb-iceberg cannot bind Iceberg v3 timestamptz_ns partition transforms through the existing generic DuckDB expression path.

  ## Design

  The goal is to add the missing DuckDB capability first, and then let duckdb-iceberg reuse it unchanged.

  Instead of adding Iceberg-specific helper functions such as iceberg_year, iceberg_month, iceberg_day, or iceberg_hour for TIMESTAMPTZ_NS, this PR extends the existing ICU date_diff implementation to support the nanosecond-precision timezone
  timestamp type directly.

  That keeps the responsibility in the right layer:

  - DuckDB provides the generic date_diff(..., TIMESTAMPTZ_NS, TIMESTAMPTZ_NS) SQL function.
  - duckdb-iceberg can keep using its existing GetDateDiffFunction(...) partition transform path.
  - Iceberg support does not need private temporal helper functions or a lossy cast from TIMESTAMPTZ_NS to TIMESTAMPTZ.

  The implementation reuses the existing ICU calendar and timezone semantics for TIMESTAMPTZ, adds TIMESTAMPTZ_NS overloads for both date_diff and datediff, and updates the extension metadata so binding and autoloading can discover the new
  overloads.

  ## Testing

  The added SQL tests cover:

  - date_diff / datediff with TIMESTAMPTZ_NS
  - year, month, day, hour, and other supported date parts
  - dynamic date part binding
  - 9-digit nanosecond literals
  - pre-epoch timestamps
  - infinity handling
  - DST behavior consistent with the existing TIMESTAMPTZ path


test/sql/function/timestamp/test_icu_datediff.test
```diff
@@ -118,6 +200,11 @@ SELECT date_diff('week', '2015-10-06 04:22:11'::timestamptz, '2016-11-25 23:19:3
 ----
 59
 
+query I
+SELECT date_diff('week', '2015-10-06 04:22:11.000000001'::TIMESTAMPTZ_NS, '2016-11-25 23:19:37.000000001'::TIMESTAMPTZ_NS);
```
hawkfish, line 222: Can we add some tests where the ns put us on either side of a boundary?

test/sql/function/timestamp/test_icu_datediff.test
```diff
@@ -138,6 +225,22 @@ FROM issue9673;
 2022-10-30 02:17:00+02	2022-10-30 02:00:21+01	43
 2021-10-31 02:39:00+02	2021-10-31 02:38:20+01	59
 
+statement ok
+CREATE TABLE issue9673_ns(starttime TIMESTAMPTZ_NS, recordtime TIMESTAMPTZ_NS);
+
+statement ok
+INSERT INTO issue9673_ns VALUES ('2022-10-30 02:17:00.000000001+02', '2022-10-30 02:00:21.000000001+01');
```
hawkfish, line 258: Can we add boundary crossing here too?

extension/icu/icu-datefunc.cpp
```diff
@@ -176,7 +176,7 @@ uint64_t ICUDateFunc::SetTimeNS(icu::Calendar *calendar, timestamp_tz_ns_t date)
 	int64_t nanos = date.value % Interval::NANOS_PER_MSEC;
 	if (nanos < 0) {
 		--millis;
-		nanos += Interval::MICROS_PER_MSEC;
+		nanos += Interval::NANOS_PER_MSEC;
```
hawkfish, line 179: Oops!

hawkfish (changes requested):
Thanks for the PR! 

The code looks fine to me (and apologies for my typo...). Some tests where the ns values affect the result would be good (.999999999 and .000000001).  Also, there appear to be conflicts that need resolving.

test/sql/function/timestamp/test_icu_datediff.test
jingdaws: Good point, thanks. I added TIMESTAMPTZ_NS tests where nanosecond values land on different sides of microsecond, millisecond, second, and week boundaries.

test/sql/function/timestamp/test_icu_datediff.test
jingdaws: Added this as well. There is now a TIMESTAMPTZ_NS boundary-crossing test through the DST fold case.

jingdaws:
hi @hawkfish 
Thanks again for the review. I’ve rebased the branch, resolved the conflict, and added the requested TIMESTAMPTZ_NS boundary-crossing tests in a separate follow-up commit.

This should be ready for another look. Please let me know if you’d like me to adjust anything further.

hawkfish (approved):
Thanks! LGTM.

lnkuiper:
Could you have a look at the failing format check?

jingdaws:
> Could you have a look at the failing format check?

Thanks for the heads-up. I checked it locally, fixed the formatting in `extension/icu/icu-datesub.cpp`, and the format check is clean now.

jingdaws:
Hi, sorry to bother you again.

  I noticed that the PR is still waiting on `Summary Expected`, while the visible `Summary` checks seem to have passed. Would you mind taking a look when you have a chance? I’m not sure whether this is a CI/merge automation issue or if there is anything I should do on my side.

  Thanks a lot!

lnkuiper:
Thanks for the reminder, I got it to run now :)

jingdaws:
Hi
  I pushed a follow-up fix for the CI issues: the generated `extension_entries.hpp` is now aligned, and the new `TIMESTAMPTZ_NS` test tables are temporary to avoid storage compatibility failures.
  I also verified the targeted test and storage compatibility check locally. Could you please help re-trigger the CI checks? Thanks!

PR #23303, opened 2026-06-16
[Variant] Fix crash for variant_extract on non-object shredded column

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9335

#### Misc

Also fixes dormant issue where root NULL values weren't counting towards VariantColumnStatsData::total_count in variant shredding analysis.
This change affects all VariantVisitor implementations that define VisitMetadata, as it's now also called for root NULL values.


Mytherin:
Thanks!

PR #23305, opened 2026-06-16
Whole-row variables NULL on join padding: is_present virtual column

carlopi:
Alternative to https://github.com/duckdb/duckdb/pull/23269

A whole-row variable (e.g. SELECT t FROM a LEFT JOIN t) over a NULL-padded join side must evaluate to NULL, not a struct of NULLs. Emit row_is_present at the source, below the joins, so the join might null it, and the reconstruct the struct_pack.

Example:
```sql
create table t1(i int);
insert into t1 values (1), (2), (3);
create table t2(i int, v varchar);
insert into t2 values (1, 'hello'), (3, 'world');
select t2 from t1 left join t2 using (i);
```

Before:
```
│              t2              │
│ struct(i integer, v varchar) │
│ {'i': 1, 'v': hello}         │
│ {'i': 3, 'v': world}         │
│ {'i': NULL, 'v': NULL}       │
```

With this PR:
```
│              t2              │
│ struct(i integer, v varchar) │
│ {'i': 1, 'v': hello}         │
│ {'i': 3, 'v': world}         │
│ NULL                         │
```

Implementation, there are two sides:
1. for sources that have virtual_column_map (that is LogicalGet or TableFunctions), we add a `row_is_present` virtual column (that is Value(true) for each row)
2. for other sources, we add a last hidden projection that is also Value(true), (and save the mapping)
3. once we reconstruct struct_pack, we do so conditionally on `row_is_present` being valid (that is, not NULL-ed by left/positional joins)

That does add complexity, but it needs to do so to work around these implementation issues:
* generated columns in structs, materializing a single Value(true) is simpler and doable independently of other columns
* virtual_columns being there only for LogicalGet (no CTE, for example)

While looking at this, I think the fact that:
```sql
select t2 from t1 left join t2 using (i);
select t2 from t1 left join (SELECT t2, i FROM t2) using (i);
```
give different results means that struct_pack (or struct extract) can't be safely moved around (hoisted / pushed-down), so that sounds like a weird behaviour that might inhibit possible optimizations, but I am not sure if that warrants the proposed diff and increased complexity. Or if the correct path is first generalizing virtual columns to be there in all bindings, and then this comes sort of easier.

Only cleanup provided, then DefaultVirtualColumnData, that moves into a base class the always-throw default behaviour, to be then implemented by whoever inherits.


Mytherin:
Thanks but I think this is still the wrong approach - this should not be `LogicalGet` specific because the problem is not `LogicalGet` specific.

Given these queries:

```sql
create table t1(i int);
insert into t1 values (1), (2), (3);
create table t2(i int, v varchar);
insert into t2 values (1, 'hello'), (3, 'world');
```

These should both return the same result:

```sql
select t2 from t1 left join t2 using (i);
select t2 from t1 left join (select 1 i, 'hello' v union all select 3, 'world') t2 using (i);
```

I think this PR brings the current state into an even weirder / worse state than `main` where these now *differ*, so instead of semantics being unclear we now have inconsistent semantics inside our own system.

I think the correct solution is to just push the struct pack into a `LogicalProjection` that sits underneath the operator where it comes from.

Getting the correct semantics is already possible in SQL just by adding a projection as well:

```sql
select t2 from t1 left join (select t2, i from t2) using (i);
│              t2              │
│ struct(i integer, v varchar) │
│ {'i': 1, 'v': hello}         │
│ {'i': 3, 'v': world}         │
│ NULL                         │
```

carlopi:
They do return the same values, I just checked, `{'i': 1, 'v': hello}  │ {'i': 3, 'v': world}  │ NULL` for both. But then, I am likely not the right person for this.

PR #23306, opened 2026-06-16
Simplify subquery planning for provably single-row subqueries

lnkuiper:
Bypasses the more complicated logic (which has side-effects through the `error` function).

Cherry-picked from https://github.com/duckdb/duckdb/pull/22712

### DIFF
--- a/src/planner/binder/query_node/plan_subquery.cpp
+++ b/src/planner/binder/query_node/plan_subquery.cpp
@@ -24,6 +24,19 @@
 
 namespace duckdb {
 
+static bool PlanReturnsExactlyOneRow(const LogicalOperator &op) {
+	switch (op.type) {
+	case LogicalOperatorType::LOGICAL_AGGREGATE_AND_GROUP_BY: {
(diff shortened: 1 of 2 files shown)


src/planner/binder/query_node/plan_subquery.cpp
```diff
@@ -24,6 +24,19 @@
 
 namespace duckdb {
 
+static bool PlanReturnsExactlyOneRow(const LogicalOperator &op) {
+	switch (op.type) {
+	case LogicalOperatorType::LOGICAL_AGGREGATE_AND_GROUP_BY: {
+		auto &aggr = op.Cast<LogicalAggregate>();
+		return aggr.groups.empty() && aggr.grouping_sets.empty() && aggr.grouping_functions.empty();
+	}
+	case LogicalOperatorType::LOGICAL_PROJECTION:
```
Mytherin, line 46: Can we add `LOGICAL_DUMMY_SCAN` here as well? Then we should be able to trigger the simple path on queries like `SELECT (SELECT 42)`.

PR #23307, opened 2026-06-16
Add `variant_comparator` which ensures variants are ordered / compared in the same manner as native DuckDB values

Mytherin:
Currently when we sort by a `variant` the sort order is rather arbitrary - we sort by internal byte representation, which gives kind of arbitrary sort orders, e.g.:

```sql
from range(-2, 2) t(i) order by i::variant;
┌───────┐
│   i   │
│ int64 │
├───────┤
│     0 │
│     1 │
│    -2 │
│    -1 │
└───────┘
```

```sql
from range(-2, 2) t(i) order by i::float::variant;
┌───────┐
│   i   │
│ int64 │
├───────┤
│     0 │
│    -2 │
│     1 │
│    -1 │
└───────┘
```

This PR adds the `variant_comparator` function which turns a variant into a byte-comparable blob. As a result, variants now sort in a similar manner to native DuckDB types.

### Variant Sorting / Comparison Rules

Variant values are now no longer just sorted / compared by type and value - but by **bucket**. Different types get put into the same bucket for sorting and comparison. The buckets are as follows:

| Rank |    Bucket    |                                           Types                                            |
|-----:|--------------|--------------------------------------------------------------------------------------------|
| 1    | BOOLEAN      | BOOL_TRUE, BOOL_FALSE                                                                      |
| 2    | NUMBER       | INT8, INT16, INT32, INT64, INT128, UINT8, UINT16, UINT32, UINT64, UINT128, DECIMAL, BIGNUM |
| 3    | REAL         | FLOAT, DOUBLE                                                                              |
| 4    | VARCHAR      | VARCHAR                                                                                    |
| 5    | BLOB         | BLOB                                                                                       |
| 6    | UUID         | UUID                                                                                       |
| 7    | TIMESTAMP    | DATE, TIMESTAMP_SEC, TIMESTAMP_MILIS, TIMESTAMP_MICROS, TIMESTAMP_NANOS                    |
| 8    | TIMESTAMP_TZ | TIMESTAMP_MICROS_TZ, TIMESTAMP_NANOS_TZ                                                    |
| 9    | TIME         | TIME_MICROS, TIME_NANOS                                                                    |
| 10   | TIME_TZ      | TIME_MICROS_TZ                                                                             |
| 11   | INTERVAL     | INTERVAL                                                                                   |
| 12   | GEOMETRY     | GEOMETRY                                                                                   |
| 13   | BITSTRING    | BITSTRING                                                                                  |
| 14   | ARRAY        | ARRAY                                                                                      |
| 15   | OBJECT       | OBJECT                                                                                     |
| 16   | NULL_VALUE   | VARIANT_NULL                                                                               |

Types within the same bucket are sorted interleaved - i.e. `2::TINYINT` sorts above `1::INTEGER`, which sorts above `0::TINYINT`:

```sql
SELECT v, variant_typeof(v) type FROM (SELECT UNNEST([0::TINYINT::VARIANT, 1::INTEGER::VARIANT, 2::TINYINT::VARIANT]) v) ORDER BY v;
┌─────────┬─────────┐
│    v    │  type   │
│ variant │ varchar │
├─────────┼─────────┤
│ 0       │ INT8    │
│ 1       │ INT32   │
│ 2       │ INT8    │
└─────────┴─────────┘
```

Types that fall in different buckets sort separately, e.g. `0::DOUBLE` sorts above `10::INTEGER` because the real bucket (`3 - REAL`) is above the integer bucket (`2 - NUMBER`):

```sql
SELECT v, variant_typeof(v) type FROM (SELECT UNNEST([0::DOUBLE::VARIANT, 10::INTEGER::VARIANT]) v) ORDER BY v;
┌─────────┬─────────┐
│    v    │  type   │
│ variant │ varchar │
├─────────┼─────────┤
│ 10      │ INT32   │
│ 0.0     │ DOUBLE  │
└─────────┴─────────┘
```

As decimals fall into the `NUMBER` bucket, they also behave "nicely" when mixed with integer types:

```sql
SELECT v, variant_typeof(v) type FROM (SELECT UNNEST([0::TINYINT::VARIANT, (-1.5)::VARIANT, 2::TINYINT::VARIANT, 1.5::VARIANT]) v) ORDER BY v;
┌─────────┬───────────────┐
│    v    │     type      │
│ variant │    varchar    │
├─────────┼───────────────┤
│ -1.5    │ DECIMAL(2, 1) │
│ 0       │ INT8          │
│ 1.5     │ DECIMAL(2, 1) │
│ 2       │ INT8          │
└─────────┴───────────────┘
```

Dates and timestamps also mix:

```sql
SELECT v, variant_typeof(v) type FROM (SELECT UNNEST([DATE '2000-01-01'::VARIANT, TIMESTAMP '1999-01-01'::VARIANT, TIMESTAMP '2000-01-01 12:30:30'::VARIANT]) v) ORDER BY v;
┌─────────────────────┬──────────────────┐
│          v          │       type       │
│       variant       │     varchar      │
├─────────────────────┼──────────────────┤
│ 1999-01-01 00:00:00 │ TIMESTAMP_MICROS │
│ 2000-01-01          │ DATE             │
│ 2000-01-01 12:30:30 │ TIMESTAMP_MICROS │
└─────────────────────┴──────────────────┘
```

Struct fields are order-insensitive, and compared in lexicographic order. For example:

```sql
SELECT v, variant_typeof(v) type FROM (SELECT {'x': 1, 'y': 1}::VARIANT v UNION ALL SELECT {'z': 0, 'x': 10}::VARIANT) ORDER BY v;
┌───────────────────┬──────────────┐
│         v         │     type     │
│      variant      │   varchar    │
├───────────────────┼──────────────┤
│ {'x': 1, 'y': 1}  │ OBJECT(x, y) │
│ {'x': 10, 'z': 0} │ OBJECT(x, z) │
└───────────────────┴──────────────┘
```

Just like DuckDB's own nested types, `NULL` is always sorted as `NULLS LAST` for children of nested types.

### Sort Key

Initially I planned to integrate this into sort keys - but that turned out to not work because the `variant_comparator` function does not round-trip due to the bucketing (i.e. post-bucketing we don't know the original type anymore). Because of that, sort keys for variants are untouched in this PR - and this is instead implemented as a collation over variants (which already existed with `variant_normalize`).

### Verification

To verify that variant sorting matches DuckDB's native sorting, this PR also adds a debug flag to help test this:

```sql
SET debug_order_verification='variant';
```

This pushes a cast to variant in every `ORDER BY`, e.g. this:

```sql
SELECT * FROM tbl ORDER BY col1, col2
```

Is turned into:

```sql
SELECT * FROM tbl ORDER BY col1::VARIANT, col2::VARIANT
```

A config is added as well which marks the remaining gaps:

* Varchar collations are lost when casting to `VARIANT`
* Union / enum ordering is subtly different
* Unnamed structs aren't supported in variant

CC @Tishj @sebastiaan-dev


Mytherin:
Failing CI is caused by https://github.com/duckdb/duckdb/pull/23309

PR #23310, opened 2026-06-16
Auto generate transform `expression.gram` (Part 8)

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/9668

This is the final* part of the auto generation of the transformer rules, covering most of the rules for `expression.gram`. There's a couple excluded from auto generation:
- `TransformExpression` has a manual `transformer.StackCheck()`, but after moving it to transformer-style we might also be able to get rid of this. 
- `TransformPrefixExpression` uses the raw parse tree to handle `- NumberLiteral` cases. 
- `TransformOverClause` has a guard to avoid nested window functions around the transform calls. 

\* I say final because I added the `optional<T>` in the last PR (https://github.com/duckdb/duckdb/pull/23236) only for `select.gram` and `expression.gram`. So I will have to go over the others once more, but should be a significantly smaller pass :)


Mytherin:
Thanks!

PR #23313, opened 2026-06-16
autovec-friendly logic

peterboncz:
## Summary

This PR adds auto-vectorization-friendly execution paths for integer and floating-point  data. 

Yes, it is about SIMD -- but autovec-style, purely relying on compilers. SIMD an also accelerate decompression, specifically, bit-unpacking. It can do comparisons >=,>,==,!=,<,<= and operators +,-,* real fast; especially if one has thin columns (8, 16, 32, and, well ok, 64)-bits; or double or float. This PR molds the respective code in DuckDB such that auto-vectorizing compilers can generate these instructions, without introducing any architecture-specific instructions.

The main pieces are:

- replaced FastPFor bit[un]packing with an own generic C++ implementation (single template). This new generic path which is always used for packing and is a non-SIMD fallback for unpacking, is supplemented with a novel autovec-friendly shuffle-based kernel, using a compiler builtin in clang/gcc. Shuffle-unpack is always used on ARM, on x86 almost always (a runtime check). Code is shorter and thanks to this change, **this PR deletes more LoC than it introduces**.
- bitmap-backed selection predicate evaluation for flat integer vectors up to 64 bits. Because autovec-comparisons return a boolean, or a bitmap -- not a selection vector. It supports  `=`, `!=`, `<`, `<=`, `>`, `>=`, including constant-on-left predicates. Conjunctions and disjunctions are supported without leaving the bitmap path. The new `SelectionResult` hence can hold either a sel_t[] or a bitmap; `SelectionVector` is its parent class that when instantiated converts bitmap to sel_t[]. This happens using a high-perf adaptive bitmap->sel_t[] conversion method that first samples and then decides for a sparse or dense strategy. The sparse strategy, i.e. for bitmaps with many 0's, uses the compiler builtin that finds the next 1-bit (skipping). The dense strategy advances 8 bits-at-a-time in 256-lookup table with all possible selection vector sequences.
- dense projection execution for safe thin-integer `+`, `-`, `*` number-crunching. For flat vectors, but also for sliced 'compatible' (=same SelectionVector) dictionary vectors. For these, we now also support mixed-typed input that require no casts.

## ARM vs x86

The PR was first developed on ARM, and later improved for x86. In x86 there are two complications: 
- there, DuckDB relies on a gcc13 build mostly, rather than clang (better at auto-vectorizing)
- both clang and gcc still do not emit AVX2 (2013) code by default. Intel still sells some ultra-low-end CPUs that have that disabled. But even low-end CPUs and all above have had this for at least 10 years (>95% of x86 machines now).

To tackle the second issue, the autovec predicate-pushdown code, and the shuffle-based bitunpacking *only* get compiled for the AVX2 target, but is only used when the CPU supports that (autovec predicate pushdown is slower without it, so not using it is preferable then). In the shuffle-unpacking, for the 8/16 bit lanes, x86 does not have a variable-shift instruction, so I added a workaround that uses multiplication.

## TPC-H SF30

The performance gains are all across TPC-H (set threads=1; best of 6 subsequent runs), and other benchmarks (see e.g. the regression tests):

ARM Graviton3 (c7g.8xlarge)
| |q01|q03|q04|q05|q06|q07|q08|q09|q10|q12|q14|q15|q17|q19|q20|q21|
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| **main **|3.58|3.68|3.12|3.89|1.04|2.55|2.50|10.65|5.13|2.55|1.89|1.28|2.79|2.14|3.27|5.83|
| **autovec** |3.38|3.33|2.84|3.61|0.78|2.31|2.31|10.36|4.64|2.40|1.66|1.07|2.45|1.81|3.02|5.43|
| **for** |3.03|3.23|2.66|3.46|0.69|2.23|1.99|10.09|4.58|2.34|1.50|1.00|2.08|1.57|2.97|5.29|

x86 (c5a.8xlarge - Zen 2)
| |q01|q03|q04|q05|q06|q07|q08|q09|q10|q12|q14|q15|q17|q19|q20|q21|
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| **main** |3.74|3.67|2.88|3.81|0.98|2.52|2.26|12.37|5.39|2.54|2.01|1.27|2.61|2.03|3.19|6.67|
| **autovec** |3.22|3.23|2.57|3.45|0.74|2.18|2.05|11.68|4.67|2.22|1.73|1.05|2.21|1.74|2.65|6.12|
| **for** |2.96|3.20|2.44|3.35|0.61|2.18|1.88|11.31|4.71|2.14|1.53|0.99|1.98|1.54|2.87|5.75|

MacBook M2 Max
| |q01|q03|q04|q05|q06|q07|q08|q09|q10|q12|q14|q15|q17|q19|q20|q21|
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| **main** |1.88|1.43|1.40|1.69|0.48|1.09|1.05|4.60|2.35|1.19|0.89|0.60|1.11|0.87|1.58|2.91|
| **autovec** |1.55|1.25|1.22|1.55|0.32|0.93|0.91|4.49|1.87|0.98|0.71|0.44|0.85|0.68|1.35|2.56|
| **for** |1.37|1.20|1.08|1.39|0.27|0.86|0.76|4.43|1.87|0.96|0.65|0.40|0.76|0.62|1.33|2.39|

Note that FOR vectors (https://github.com/duckdb/duckdb/pull/23314) build on this and make DuckDB use thin types more often. The overall performance gain compound further with that.


Mytherin:
Thanks for the PR!

The speed-ups are very promising, but implementation-wise this is very bolted-on and hard-coded only for filters. In general we've been trying to unify filters with expression execution, because there is no reason for filters to be anything but expression execution, and by moving optimizations into expression execution everything benefits and not just filters that are pushed down into table scans. This moves back in the other direction again by special casing implementation code for filters only. 

Two different implementation approaches I could see:

* Allow `SelectionVector` to be *either* a bitmap *or* a set of indices, where if you access it using certain APIs (e.g. `.data()` or `.get_index`) a transformation to the "standard" selection vector will be forced similar to what is going on in this API. Code can then take advantage of bitmap selection vectors where performance is desired.
* Add a new vector type to model "bitmap selection over a vector", separate from `DICTIONARY_VECTOR`, then take advantage of that vector type in certain code paths.

I think integrating this into the `SelectionVector` is the most promising. We can then proceed with integrating this into the expression execution layer (instead of the filter layer directly).

peterboncz:
Thanks Mark — I integrated the bitmap representation into the SelectionVector.
- I added a subclass SelectionResult of SelectionVector (private inheritance): the subclass backs its selection with either an index array or a bitmap (SelectionData gained the bitmap buffer + row span); the base class is always a sel_t array. Existing code calls selvec.get_index(i) in performance-critical loops and must never bump into a bitmap — private inheritance makes that a compile-time guarantee: a SelectionResult can only cross into SelectionVector land through an explicit Flattened(), a once-per-vector materialization.
-  The branchless comparison kernels live in the expression Select layer and emit a bitmap whenever the caller's output selection is a SelectionResult (bitmap-aware consumers then intersect directly, e.g. conjunction AND). So the operator opts in, by declaring its result a SelectionResult — this is helpful because  producing bitmap output always regresses in operators that immediately need indices.

Net effect is less filter-specific code, with the fast path living where all selection happens. I did try first with all in SelectionVector, but  would have had to call Flatten() on it in many code locations, which is error-prone. Existing extension code that uses SelectionVector can safely continue to use it without danger of bumping into a bitmap.

Happy to iterate on the SelectionVector API shape if you'd like it drawn differently.

I also decided to **delete fastpfor** (from Daniel Lemire) introducing an own implementation because it is faster:
- a completely **generic** C++ templated naive [un]pack is faster because fastpfor compiled into 32 C methods that yielded 8,16,32 tuples (depending on width). This meant that the C function had to be called many times for e.g. 1024 tuples in a segment: call overhead and lost compiler optimization opportunities. Auto-vectorization and templating support in compiler got better in the years since fastpfor was created, favoring plain C++.
- a **shuffle** variant for certain bitwidths with clang/gcc. SIMD can help a lot (alas DuckDB does not use _interleaved_ bitpacking, then this would fly) but even with the sequential bit-layout used in DuckDB, one can take advantage of byte-shuffle instructions to assemble a destination lane, by putting input bytes there, and then doing shift and mask.

The new autovec codepaths are disabled in DUCKDB_SMALL_BINARY to keep code size down there.
They are taken by clang/gcc compiled ARM targets and x86 if they support AVX2 (a   runtime condition there), but are avoided on other platforms (those presumably without any good SIMD support).

The below table shows unpack ns/tuple (1024 segment) on my M2 Max laptop:

| W | out | clang fastpfor | clang generic | clang shuffle | gcc fastpfor | gcc generic | gcc shuffle |
|---|-----|----------------|---------------|---------------|--------------|-------------|-------------|
| 1 | u8 | 0.162 | 0.155 | 0.023 | 0.172 | 0.071 | 0.036 |
| 2 | u8 | 0.163 | 0.152 | 0.023 | 0.172 | 0.105 | 0.024 |
| 3 | u8 | 0.201 | 0.170 | 0.039 | 0.208 | 0.170 | 0.042 |
| 4 | u8 | 0.162 | 0.153 | 0.020 | 0.171 | 0.152 | 0.020 |
| 5 | u8 | 0.239 | 0.193 | 0.044 | 0.243 | 0.193 | 0.043 |
| 6 | u8 | 0.241 | 0.189 | 0.036 | 0.246 | 0.193 | 0.038 |
| 7 | u8 | 0.283 | 0.256 | 0.038 | 0.282 | 0.207 | 0.039 |
| 9 | u16 | 0.607 | 0.187 | 0.036 | 0.239 | 0.187 | 0.036 |
| 10 | u16 | 0.226 | 0.188 | 0.036 | 0.231 | 0.188 | 0.036 |
| 11 | u16 | 0.631 | 0.221 | 0.060 | 0.251 | 0.202 | 0.061 |
| 12 | u16 | 0.607 | 0.191 | 0.036 | 0.234 | 0.188 | 0.036 |
| 13 | u16 | 0.625 | 0.255 | 0.061 | 0.270 | 0.213 | 0.061 |
| 14 | u16 | 0.266 | 0.257 | 0.060 | 0.272 | 0.214 | 0.061 |
| 15 | u16 | 0.589 | 0.294 | 0.061 | 0.290 | 0.224 | 0.061 |
| 17 | u32 | 0.310 | 0.167 | 0.094 | 0.257 | 0.179 | 0.066 |
| 18 | u32 | 0.320 | 0.160 | 0.119 | 0.259 | 0.178 | 0.065 |
| 19 | u32 | 0.327 | 0.177 | 0.067 | 0.262 | 0.187 | 0.068 |
| 20 | u32 | 0.311 | 0.156 | 0.106 | 0.258 | 0.173 | 0.067 |
| 21 | u32 | 0.333 | 0.184 | 0.088 | 0.268 | 0.193 | 0.069 |
| 22 | u32 | 0.331 | 0.174 | 0.101 | 0.283 | 0.191 | 0.069 |
| 23 | u32 | 0.338 | 0.201 | 0.085 | 0.286 | 0.208 | 0.070 |
| 24 | u32 | 0.317 | 0.155 | 0.081 | 0.257 | 0.192 | 0.067 |
| 25 | u32 | 0.357 | 0.208 | 0.082 | 0.284 | 0.230 | 0.070 |
| 26 | u32 | 0.376 | 0.184 | 0.109 | 0.301 | 0.213 | 0.071 |
| 27 | u32 | 0.366 | 0.227 | 0.123 | 0.302 | 0.246 | 0.116 |
| 28 | u32 | 0.348 | 0.163 | 0.075 | 0.296 | 0.202 | 0.071 |
| 29 | u32 | 0.383 | 0.247 | 0.117 | 0.309 | 0.263 | 0.116 |
| 30 | u32 | 0.396 | 0.218 | 0.115 | 0.320 | 0.243 | 0.115 |
| 31 | u32 | 0.399 | 0.266 | N.A. | 0.329 | 0.278 | N.A. |

conclusion: shuffle >> generic >> fastpfor. For x86 we also us shuffle when the CPU supports AVX2, and fall back to generic if not.

src/execution/expression_executor/execute_function.cpp
```diff
@@ -1,13 +1,29 @@
...
 #include "duckdb/execution/expression_executor.hpp"
+#include "duckdb/execution/expression_executor/bitmap_comparison.hpp"
 #include "duckdb/main/config.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
 
 namespace duckdb {
 
+namespace {
+bool IsSafeAutoVecArithmetic(const BoundFunctionExpression &expr) {
+	auto autovec_arith = DUCKDB_AUTOVEC && CpuBenefitsFromAutoVec();
```
Mytherin, line 12: Let's not hardcode function-specific things here - instead we could add a flag or different callback to the function itself

src/execution/expression_executor/execute_function.cpp
```diff
@@ -56,22 +76,54 @@ bool ExecuteFunctionState::TryExecuteDictionaryExpression(const BoundFunctionExp
...
+		return false;
 	}
-
 	const auto input_dictionary_size = input_dictionary_size_opt.GetIndex();
-	if (input_dictionary_size >= MAX_DICTIONARY_SIZE_THRESHOLD) {
-		return false; // Dictionary is too large, bail
+	auto &input_sel = DictionaryVector::SelVector(unary_input);
+	if (safe_autovec_arith && AutoVecCountPaysOff(args.size()) && // dense dictionary arithmetic
+	    input_dictionary_size <= STANDARD_VECTOR_SIZE &&
+	    DenseAutoVecPaysOff(args.size(), input_dictionary_size, GetTypeIdSize(result.GetType().InternalType()))) {
```
Mytherin, line 92: Doing this in the dictionary code seems like an odd place, shouldn't this be part of the function definition somewhere? Maybe this can be in the scalar executors hidden behind a different flag (e.g. `ScalarExecutor::Execute<AutoVectorization::ENABLED>`

src/execution/expression_executor/execute_function.cpp
```diff
@@ -259,10 +306,21 @@ void ExpressionExecutor::Execute(const BoundFunctionExpression &expr, Expression
 
 idx_t ExpressionExecutor::Select(const BoundFunctionExpression &expr, ExpressionState *state,
                                  const SelectionVector *sel, idx_t count, SelectionVector *true_sel,
-                                 SelectionVector *false_sel) {
+                                 SelectionVector *false_sel, SelectionResult *bitmap_sel) {
```
Mytherin, line 309: I don't quite follow what `bitmap_sel` is doing here

src/include/duckdb/common/types/selection_result.hpp
```diff
@@ -0,0 +1,122 @@
...
+#include "duckdb/common/autovec.hpp"
+#include "duckdb/common/bit_utils.hpp"
+#include "duckdb/common/types/selection_vector.hpp"
+#include "duckdb/common/types/validity_mask.hpp"
+
+#include <cstring>
+
+namespace duckdb {
+//! Bitmap-capable selection result; Flattened() materializes an index selection.
+struct SelectionResult : private SelectionVector {
```
Mytherin, line 19: Can we integrate this directly into `SelectionVector` instead (i.e. make every selection vector bitmap-capable)?

Mytherin (commented):
Thanks for the fixes! I think it would be good to split this off into two PRs (one for the bit packing changes, and the other for the bitmap selection vectors).

For the bitmap selection vectors - the main issue here is that this is touching code that it shouldn't be touching. Function-specific code shouldn't live in the generic `ExpressionExecutor` - it should live in the specific functions that it affects.

I would say for the vectorization we should move towards `SelectionVector` being representable as either a set of indexes, or as a bitmap index, where trying to access a bitmap-based selection vector using the traditional interfaces (`.get_index`, etc) "flattens" the bitmap selection vector into a traditional selection vector. That should then allow these bitmap indexes to be passed down various functions in the current infrastructure. 

Maybe this can be a `SelectionVectorType` that has three values:

```cpp
INCREMENTAL
INDEX_BASED
BITMAP_BASED
```

Where `INCREMENTAL` is what is currently represented as `sel_vector == nullptr`, and `INDEX_BASED` is what is currently represented as `sel_vector` is present.

Perhaps this is also something that is also better split off into separate PRs.

src/execution/expression_executor/execute_function.cpp
peterboncz: I can replace this with a Function property `bool autovectorized = false; `

src/execution/expression_executor/execute_function.cpp
peterboncz: this dictionary code recognizes flat columns sliced with the same selvec, and diverts them into flat code, ignoring the selvec. I could split off a `ExecuteFunctionState::TryExecuteDenseFunction(expr, args, state, result)` - could that help address your concern? We do need to take action in dictionary code, I think.

src/execution/expression_executor/execute_function.cpp
peterboncz: bitmap_sel is an alternative output form: instead of returning a selvec in `true_sel`, the callee may return the surviving rows still as a bitmap, so `ExpressionExecutor::Select(BoundConjunctionExpression)` can combine children. 

In my design, I could have consolidated (`true_sel, false_sel, bitmap_sel`) into a `struct SelectionTarget`

src/include/duckdb/common/types/selection_result.hpp
peterboncz: Well, the bitmap already lives in `SelectionVector` — `SelectionData` contains bitmap_data/row_span/is_bitmap, and data(), operator[] and Sort() already flatten on access via Flatten(). The only accessor that doesn't is `get_index(),` which asserts instead.

So do you want get_index() to flatten automatically? That's a lot of call sites, in the hottest loops (join_hashtable, update_segment, comparison_operators, tuple_data_scatter_gather, plus the executors).

Assuming the answer is no, a stray bitmap poses a silent regression (not a compile error). This is why I applied inheritance with SelectionResult such that a filter-oriented operator can opt in safely.

Note that some operators use selvecs as mappings, not filters: DictionaryVector::SelVector, the build-row-id vector in EmitDictVectors (join_hashtable.cpp:2838), sort permutations — duplicates, arbitrary order, domain unrelated to count. These cannot be bitmaps.

Maybe another approach is to explicitly extract the selvec in some struct, that one can call get_index() or simply [] on.  I stayed away from that because I already have so many LoC changed and did not want more.
```
enum class SelectionVectorType : uint8_t { INCREMENTAL, INDEX_BASED, BITMAP_BASED };

//! Index-based view of a selection — the form hot loops consume. Borrows; owns nothing.
struct FlatSelection {
      inline idx_t operator[](idx_t i) const { 
           D_ASSERT(!sel || i < capacity); 
           return sel ? sel[i] : i; 
      }
      const sel_t *data() const { return sel; }
      sel_t *data() { return sel; }
      bool IsIncremental() const { return sel == nullptr; }
private:
      friend class SelectionVector;
      sel_t *sel = nullptr;
      idx_t capacity = 0;
};

class SelectionVector {
public:
      FlatSelection Flatten();  //! Materialize BITMAP_BASED -> INDEX_BASED. 
      bool IsSet() const { return sel_type != SelectionVectorType::INCREMENTAL; }
private:
      SelectionVectorType sel_type = SelectionVectorType::INCREMENTAL;
      FlatSelection flat;                      // primary state; may point outside selection_data
      buffer_ptr<SelectionData> selection_data; // optional owner of flat.sel and/or the bitmap
};
```

peterboncz:
Thanks for the feedback!  I can split and try to further reduce the code impact as follows:
- PR1: autovec.hpp and the bitunpack replacements
- PR2: autovec-friendly flat arithmetic and the dense evaluation over a shared selvec dictionary inputs
- PR3: SelectionVector getting  autovec-friendly comparisons into bitmaps (and AND/OR on these)
- PR4: FOR vectors

src/include/duckdb/common/types/selection_result.hpp
Mytherin: That's a good point - I would say we should then aim to change `SelectionVector` to `SelectionResult` in the locations where we're selecting tuples. That shouldn't inherit from `SelectionVector`, though, it should be a separate class that can internally be represented as either a `SelectionVector` or a bitmap (with maybe a helper method to turn it into a `SelectionVector`).

We probably then also want to introduce a new vector type (`VectorType::BITMAP`) which is similar to a dictionary vector but with a bitmap selection over the underlying child vector.

I think this can be constructed in a relatively incremental manner in a series of PRs and seems like a good direction to go into.

PR #23316, opened 2026-06-16
Add request body length to http logs

DinosL:
While debugging S3 upload PUT timeouts, I found the HTTP log gives us each request's start/end time and duration, but not how many bytes the request body carried. Without that I can't tell how far an upload got before it stalled, or compute its throughput (bytes / duration), which would be useful to tell a dead connection apart from a slow but progressing one.


Mytherin:
Thanks!

PR #23323, opened 2026-06-17
Disable arrow_scan dup in window_self_join opt

### LINKED ISSUE #713: Optimizer incorrectly rewrites window plans involving arrow scan inputs to self-joins with multiple refs to the scan
This regressed in 1.5.0.

This affects queries that use window functions but _do not_ reference the arrow stream table more than once, e.g. `SELECT g, v, sum(v) OVER (PARTITION BY g) AS s FROM %s`.

1.4.4 always left this as one ARROW_SCAN_DUMB in the plan. 1.5.0 rewrites this to a self-join with multiple ARROW_SCAN_DUMB nodes, which doesn't work. Manifests as:

`java.sql.SQLException: Invalid Input Error: This stream has been released`

Related to: https://github.com/duckdb/duckdb-python/issues/70

I'm filing a separate issue because that preexisting issue describes a query that references the arrow stream multiple times. If that were the limitation, it's much clearer boundary: There are still situations where you can use arrow stream inputs and it's easy to understand when you can and when you can't (multiple references no, one reference yes). This new regression breaks that contract.

(PR: https://github.com/duckdb/duckdb/pull/23323)

randalloveson:
This fixes an issue that began with 1.5.0 where very simple queries that reference an arrow_scan table only once but which involve window functions start to fail due to the window_self_join opt.

Fixes: [duckdb-java issue #713](https://github.com/duckdb/duckdb-java/issues/713)

Somewhat related: [duckdb-python issue #70](https://github.com/duckdb/duckdb-python/issues/70)

New here so I apologize if the new field on table_function is presumptuous. There may well be a better solution.

Conspicuously missing is a similar fix for the cte_inlining optimization, which also duplicates upstream-subtrees that could contain non-duplicable table functions like arrow_scan.

Alternative solution: `read_csv` throws `NotImplementedException` when it's serialized, which just so happens to disable these same sorts of optimizations. Could do the same thing for arrow_scan. Feels hacky but it sidesteps the need for a new prop that everybody needs to respect explicitly.

### DIFF
--- a/src/function/table/arrow.cpp
+++ b/src/function/table/arrow.cpp
@@ -341,6 +341,7 @@ void ArrowTableFunction::RegisterFunction(BuiltinFunctions &set) {
 	arrow.filter_prune = true;
 	arrow.supports_pushdown_type = ArrowPushdownType;
 	arrow.parallelism = TableFunctionParallelism::SEQUENTIAL;
+	arrow.supports_multiple_scans = false;
 	set.AddFunction(arrow);
 
 	TableFunction arrow_dumb("arrow_scan_dumb", {LogicalType::POINTER, LogicalType::POINTER, LogicalType::POINTER},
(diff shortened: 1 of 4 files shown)


Tishj:
I believe I remember hearing that we aim to detect multiple scans generically and optimize those into a materialized CTE, preventing the multiple scan issue

But I'm not 100% on that, and can't give you a timeline

randalloveson:
Sounds like this fix is superseded by something else on the roadmap, so maybe the changes to TableFunction itself are undesirable.

Perhaps the hackier-but-narrower fix of simply throwing `NotImplementedException` during serialization like `read_csv` does would be acceptable? That would fix the regression as well, and it would confine the changes to `arrow_scan`.

PR #23327, opened 2026-06-17
Show transport errors in HTTP log

DinosL:
When an HTTP request fails at the transport level (connection timeout, connection refused etc.) the request is logged as `'response': {'status': INVALID, 'reason': '', 'headers': {}}` which is not very informative. This PR adds the error message in the `reason`.


Mytherin:
Thanks - can you just fix the merge conflict?

Mytherin:
Thanks!

PR #23328, opened 2026-06-17
CMake: Set the extension base directory to the duckdb module dir

nicholasjng:
DuckDB's extension-patch step hardcodes `${CMAKE_SOURCE_DIR}`, which only equals the submodule root in a standalone build. When DuckDB is pulled in via `add_subdirectory`, it resolves to the parent project's root, which causes the patch script and patch-dir lookups to happen in the wrong place.

I ran into this because I'm vendoring duckDB as a submodule for a Python bindings project, and statically link extensions from the submodule checkout in the build process.

----------------

Hello again! Right now I'm symlinking `apply_extension_patches.py` into my project's `scripts/`, which also contains other scripts, so it's not a huge issue. But I thought I might as well raise this to get your opinion.


Mytherin:
Thanks!

PR #23329, opened 2026-06-17
Fix JSON COPY date/time formatting for nested values

### LINKED ISSUE #23152: dateformat does not work for the top layer date value
### What happens?

Hi,

I found that the dateformat option does not work for the whole JSON value. This is an example:

```
COPY (
    SELECT
        DATE '2026-01-02' AS d,
        {'d': DATE '2026-01-02'} AS payload
) TO '/tmp/test.json'
  (FORMAT JSON, dateformat '%d/%m/%Y');

SELECT
    json_extract_string(json, '$.d') AS top_level,
    json_extract_string(json, '$.payload.d') AS nested
FROM read_json_objects('/tmp/test.json');
┌────────────┬────────────┐
│ top_level  │   nested   │
│  varchar   │  varchar   │
├────────────┼────────────┤
│ 02/01/2026 │ 2026-01-02 │
└────────────┴────────────┘
```

### To Reproduce

```
COPY (
    SELECT
        DATE '2026-01-02' AS d,
        {'d': DATE '2026-01-02'} AS payload
) TO '/tmp/test.json'
  (FORMAT JSON, dateformat '%d/%m/%Y');

SELECT
    json_extract_string(json, '$.d') AS top_level,
    json_extract_string(json, '$.payload.d') AS nested
FROM read_json_objects('/tmp/test.json');
```

### OS:

Ubuntu 24.04

### DuckDB Version:

a7b16c6f323b405d9cfce9b1d6867df21687b19f

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Chi Zhang

### Affiliation:

Tsinghua University

- [x] Yes, I have

- [x] Yes, I have

Yes

OlexG:
Fixes #23152

This updates JSON COPY formatting so `dateformat` and `timestampformat` are applied by the recursive JSON serializer, including nested `STRUCT` and `LIST` values.

Partition path formatting remains handled by the existing COPY projection macros, so partition directory values keep the current behavior.

Validation:
- `DUCKDB_EXTENSIONS='json;icu' make reldebug`
- `build/reldebug/test/unittest test/sql/json/table/read_json_dates.test`

### DIFF
--- a/extension/json/json_functions/json_create.cpp
+++ b/extension/json/json_functions/json_create.cpp
@@ -9,7 +9,9 @@
 #include "duckdb/function/cast/default_casts.hpp"
 #include "duckdb/function/function_binder.hpp"
 #include "duckdb/function/scalar/strftime_format.hpp"
+#include "duckdb/planner/expression/bound_cast_expression.hpp"
 #include "duckdb/planner/expression/bound_constant_expression.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "duckdb/planner/expression/bound_parameter_expression.hpp"
 #include "duckdb/common/serializer/deserializer.hpp"
@@ -18,6 +20,8 @@
 #include "json_functions.hpp"
 
 namespace duckdb {
+
+static constexpr const char *JSON_COPY_TO_JSON_INTERNAL_NAME = "__internal_json_copy_to_json";
 
 struct StructNames {
 	void Insert(const string &name) {
@@ -315,6 +319,13 @@ static bool BindJSONCopyFormat(ClientContext &context, Expression &argument, con
 	return true;
 }
 
+static void ParseJSONCopyFormatString(const string &format_string, const string &name, StrfTimeFormat &format) {
+	auto error = StrTimeFormat::ParseFormatSpecifier(format_string, format);
+	if (!error.empty()) {
+		throw InvalidInputException("Failed to parse %s format specifier %s: %s", name, format_string, error);
+	}
+}
+
 static unique_ptr<Expression> BindJSONCopyTimestampTZFormatter(ClientContext &context, const string &format_string,
                                                                const LogicalType &type) {
 	vector<unique_ptr<Expression>> children;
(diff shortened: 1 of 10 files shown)


extension/json/json_functions/json_create.cpp
```diff
@@ -54,6 +71,62 @@ struct JSONCreateFunctionData : public FunctionData {
...
 };
 
+struct JSONCopyFormatOptions {
+	optional_ptr<StrfTimeFormat> date_format;
+	optional_ptr<StrfTimeFormat> timestamp_format;
+	optional_ptr<ClientContext> context;
+	optional_ptr<const Expression> timestamptz_format_expression;
+};
+
+struct JSONCopyToJSONFunctionData : public FunctionData {
```
OlexG, line 88: This adds COPY-specific bind data instead of extending `JSONCreateFunctionData` because the date/time formats are only valid for `COPY ... TO ... (FORMAT JSON, ...)`. Public `to_json`, `json_quote`, `row_to_json`, `array_to_json`, and JSON casts should keep their existing behavior.

extension/json/json_functions/json_create.cpp
```diff
@@ -208,6 +281,82 @@ static unique_ptr<FunctionData> RowToJSONBind(BindScalarFunctionInput &input) {
 	return JSONCreateBindParams(bound_function, arguments, false);
 }
 
+static bool BindJSONCopyFormat(ClientContext &context, Expression &argument, const string &name, string &format_string,
```
OlexG, line 299: This is the binder for the internal `json_copy_to_json(value, date_format, timestamp_format)` helper.

Also, a separate TIMESTAMPTZ formatter is on purpose, so timezones are handled correctly.

lnkuiper (changes requested):
Thanks for the PR. The recursive formatting looks pretty good already!

I have found some issues that need to be addressed (and tested) before this can be merged.

1. `TIMESTAMP_NS` partition formatting loses nanosecond precision. The JSON payload uses the new `CreateValuesTimestampNS` path and preserves `%n`, but partition columns still go through `json_copy_strftime_if_timestamp`, which has no `TIMESTAMP_NS` overload. That makes the partition path round to microseconds while the JSON body keeps nanoseconds.

Repro:

```sql
# name: test/temp.test
# description: Repro for JSON COPY TIMESTAMP_NS partition precision
# group: [json]

require json

statement ok
COPY (
    SELECT TIMESTAMP_NS '2026-01-02 03:04:05.123456789' AS ts, 42 AS v
) TO '{TEMP_DIR}/json_copy_tsns_partition_precision'
  (FORMAT JSON, PARTITION_BY (ts), WRITE_PARTITION_COLUMNS, timestampformat '%Y/%m/%d %H:%M:%S.%n');

query II
SELECT filename, trim(content, chr(10))
FROM read_text('{TEMP_DIR}/json_copy_tsns_partition_precision/*/*.json') ORDER BY filename;
----
{TEMP_DIR}/json_copy_tsns_partition_precision/ts=2026%2F01%2F02%2003%3A04%3A05.123456789/data_0.json	{"ts":"2026/01/02 03:04:05.123456789","v":42}
```

On this branch the actual filename contains `123457000`, while the JSON body contains `123456789`.

2. `json_copy_to_json` is now directly callable from SQL. It is marked `internal=true` in `duckdb_functions()`, but that does not prevent normal SQL calls, so users can call COPY-only formatting behavior directly. Scalar functions do have default-argument support now through `FunctionSignature`, so this helper does not seem necessary just because defaults are unavailable. Could this be bound directly into the generated COPY plan instead of registered in the catalog, or otherwise kept out of the SQL surface?

3. Could we make the `json_create.cpp` diff smaller? The date/timestamp/timestamp_ns/timestamptz paths repeat a lot of “default cast or format, then emit strings” logic. A shared helper, or one timestamp formatting function with type-specific conversion, would make the changes easier to read/review/maintain.

4. MAP keys do not get the COPY formatting options. MAP values recurse with `options`, but keys are default-cast to `VARCHAR` before formatting is considered. For example, `json_copy_to_json(map([DATE '2026-01-02'], [DATE '2026-01-03']), '%d/%m/%Y', NULL)` formats the value but leaves the key as `2026-01-02`.

OlexG:
Got it, thanks for the review @lnkuiper. Will look at these

OlexG:
1. Great catch, added TIMESTAMP_NS as a macro so partition paths don't lose formatting

2. I could have added default args to `to_json` as you proposed, but does it make sense to change the API surface for this? In my opinion, no (but I could be wrong, haha)
Instead, I opted for your second suggestion of binding json_copy_to_json directly. This keeps it out of the SQL surface. LMK if you would like something else, and I'd be happy to fix it :)!

3. Done, did a refactor with some helpers

4. Great catch, didn't consider this. Implemented.


cc @lnkuiper

lnkuiper:
Thanks for the changes! Looks great now.

One small thing: I think throwing the error in the (de-)serialization is causing an issue:
```sql
# name: test/temp.test
# description: Repro for JSON COPY format plan serialization
# group: [json]

require json

query I
SELECT json_extract(json_serialize_plan($$COPY (SELECT DATE '2026-01-02' AS d) TO '{TEMP_DIR}/json_plan_format_repro.json' (FORMAT JSON, dateformat '%d/%m/%Y')$$), '$.error')::BOOLEAN;
----
false
```
Any reason the error should be thrown there?

OlexG:
I added it to stop debug plan verification from deserializing an unregistered private function; we'd need to add it back to the catalog to fix this. 
I looked around the codebase a bit, and I see we have this pattern of `__internal` functions that are registered but not callable, which fits this really well. Honestly should have looked for this before... 
I will follow how these are implemented.

OlexG:
@lnkuiper fixed the serialization issue, but this means registering `__internal_json_copy_to_json` in the catalog. 

Still, this function isn't callable by the user. The registered catalog function has a bind callback that always throws `BinderException("__internal_json_copy_to_json is for internal use only")`, which follows existing patterns this codebase has. 
I added tests for all this as well. 

LMK if anything else looks concerning to you! And thanks for your time reviewing this.

lnkuiper:
Great! Thanks for implementing the feedback. Let's run CI :)

lnkuiper (approved):

OlexG:
seems like the tests need a local file system, fixing @lnkuiper

OlexG:
@lnkuiper could you kick off CI again. 

Also unrelated, but this PR has been sitting for a while - https://github.com/duckdb/duckdb-httpfs/pull/344
not sure if you work on this ^, but pretty small change

PR #23332, opened 2026-06-17
Use optional semantic values for generated transformer rules

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/9728

Now using optional semantics across all transformer files consistently.


Mytherin:
Thanks!

PR #23333, opened 2026-06-17
Step 6: Flip the switch and have delim join as CTEs by default

kryonix:
This was supposed to be a "one line" PR. Of course, it was not that easy 😉 CI showed a few remaining issues with regressions and a few underspecified test cases. So this PR:

- Flips the flag to now have DELIM joins expressed as CTEs by default.
- Fixes the CSE optimizer to recognize more identical plan shapes even with different column projections.
- Fixes CTE column pruning (again..) to fix regressions with parquet-based TPC-H.

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -487,7 +487,7 @@
         "description": "Rewrite delim joins to materialized CTEs during dependent join flattening",
         "type": "BOOLEAN",
         "default_scope": "local",
-        "default_value": "false"
+        "default_value": "true"
     },
     {
         "name": "deprecated_using_key_syntax",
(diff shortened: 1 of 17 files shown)


hannes:
u7347911374_an_enthusiastic_marching_crowd_that_is_waving_sig_c31fcfab-1d1d-4534-b092-f3826b8a9caa_2

lnkuiper:
Looks great, ready to go from my side. I think there might be a very minor issue in the `CommonSubplanOptimizer`: `LOGICAL_ASOF_JOIN` is included in the new expanded-binding and projection-map clearing paths, but `ConvertProjectionMaps` still only strips projection maps for `LOGICAL_ANY_JOIN` and `LOGICAL_COMPARISON_JOIN`. I don't think this can cause any issues, but just want to comment that here. Adding this to the queue!

PR #23337, opened 2026-06-17
Optimize StringUtil::Replace

jprafael:
Optimize StringUtil::Replace to avoid repeated in-place string replacement. The previous implementation called std::string::replace once per match, which can repeatedly move the remaining tail of large strings. This can become very expensive when replacing many occurrences in a large generated string. This was observed at the final step of ducklake commit, where the metadata query has many instances of `{SNAPSHOT_ID}`.

This change first counts matches, reserves the final result size, and then builds the output with append operations. The behavior is preserved, including cases where the replacement text contains the search text. A small regression test was added for the relevant edge cases.

### DIFF
--- a/src/common/string_util.cpp
+++ b/src/common/string_util.cpp
@@ -548,13 +548,36 @@ string StringUtil::Replace(string source, const string &from, const string &to) 
 	if (from.empty()) {
 		throw InternalException("Invalid argument to StringUtil::Replace - empty FROM");
 	}
+	if (source.empty() || from == to) {
+		return source;
+	}
+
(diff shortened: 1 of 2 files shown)


jprafael:
An alternative implementation would be similar to boost::replace_all, where the final size is not precomputed. That has the benefit of not requiring a double pass on the string but may cause reallocations of the string due to resizing.

lnkuiper:
Thanks for the PR! This seems to fail CI (`build/relassert/test/unittest 'Test replace strings'`), could you have a look?

PR #23339, opened 2026-06-17
Allow triggering internal join types

hannes:
This PR adds syntax to trigger DuckDB's internal join operators like `MARK_JOIN`, `RIGHT_SEMI` etc. 

The syntax looks like this to avoid adding reserved keywords for important words like 'Mark':
```SQL
FROM x JOIN BY (TYPE MARK_JOIN) y 
```


Mytherin:
Thanks for the PR!

Maybe we can use brackets with parameters like so:

```sql
FROM x JOIN BY (TYPE MARK) y;
```

That would allow us to slot in extra parameters (e.g. `MARKER_NAME`) more easily.

Mytherin:
Thanks!

PR #23340, opened 2026-06-17
fix(adbc): support the ADBC Statistics API (with full nanoarrow)

eitsupi:
Support the ADBC 1.1.0 Statistics API.

- `ConnectionGetStatistics`: returns the table-level `ROW_COUNT` statistic from `duckdb_tables().estimated_size`, as `float64` with `statistic_is_approximate = true`.
- `ConnectionGetStatisticNames`: returns an empty result (DuckDB has no driver-specific statistics).

This is the same fix of #23230, but the implementation has been simplified by bundling full nanoarrow 0.8.0.

cc @lidavidm


pdet:
Closing this for now in favor of https://github.com/duckdb/duckdb/pull/23230

PR #23341, opened 2026-06-17
Add additional guards to `DICT_FSST` to prevent exception during compression with small block size

sebastiaan-dev:
Resolves https://github.com/duckdblabs/duckdb-internal/issues/9685 by introducing additional guards during the analyze phase for `DICT_FSST`. Previously we could encounter a situation where during compression we realize the produced intermediate result can never fit in the block size of the current database. In turn leading to a fatal exception.

To resolve this we added guards for two cases:
- `len(string) + metadata` does not fit inside the available space in a block, we then disable `DICT_FSST` entirely for the given row group.
- `len(string)*2 + metadata` does not fit inside the available space in a block, we then disable `FSST` encoding paths within `DICT_FSST`. This guards the worst case scenario for `FSST` encoding where every byte gets an exception marker, while still allowing for plain dictionary encoding.

In addition we have introduced `DictFSSTGetSegmentInfo` which provides the specific mode of `DICT_FSST` that has been applied to a given segment. This information is accessible through `pragma_storage_info` and can be viewed in the `segment_info` column. This function is part of the current PR as we need it to test the newly introduced code paths by the guards.

```sql
> select compression, segment_info FROM pragma_storage_info('table');
┌─────────────┬──────────────────┐
│ compression │   segment_info   │
│   varchar   │     varchar      │
├─────────────┼──────────────────┤
│ DICT_FSST   │ DICTIONARY: 2000 │
└─────────────┴──────────────────┘
```

Last, we have modified `vector_format.validity.RowIsValid(i)` to use the selection vector index instead of the iteration count.


src/storage/compression/dict_fsst/analyze.cpp
```diff
@@ -6,26 +6,75 @@ namespace dict_fsst {
...
+	size = AlignValue<idx_t>(size);
+	size += dict_indices_space;
+
+	return size;
+}
+
+//! A segment cannot be spread out over multiple blocks, so if a string cannot fit in an empty segment
+//! the encoding will fail.
+static bool StringFitsBlock(const idx_t block_size, const idx_t str_len) {
+	return RequiredSpace(str_len, false) <= block_size - DictFSSTCompression::FSST_SYMBOL_TABLE_SIZE;
```
Mytherin: If we're using dictionary encoding we shouldn't be using the FSST symbol table, shouldn't this subtraction be moved to `FSSTFitsBlock`?

src/storage/compression/dict_fsst/analyze.cpp
```diff
@@ -6,26 +6,75 @@ namespace dict_fsst {
...
+
+	// Dictionary stores only one valid string.
+	const idx_t dict_indices_width = BitpackingPrimitives::MinimumBitWidth(1);
+	const idx_t dict_indices_space = BitpackingPrimitives::GetRequiredSize(1, dict_indices_width);
+
+	idx_t size = 0;
+	size += sizeof(dict_fsst_compression_header_t);
+	size = AlignValue<idx_t>(size);
+	size += str_len;
+	if (fsst_encoded) {
```
Mytherin, line 25: We're doing all of these operations for every single string - but it seems these are almost all constant operations. Looking at this it looks like the result of all of this are just two string lengths (`max_str_length` and `max_str_length_fsst`). Can we maybe just compute those two lengths and then use those in the loop below?

Mytherin: Ideally we would fold `max_str_length` into `DictFSSTCompression::STRING_SIZE_LIMIT` as well since that's also a similar length check

src/storage/compression/dict_fsst/compression.cpp
```diff
@@ -849,7 +851,7 @@ void DictFSSTCompressionState::Compress(Vector &scan_vector, idx_t count) {
 				break;
 			}
 
-			if (append_state == DictionaryAppendState::REGULAR) {
+			if (append_state == DictionaryAppendState::REGULAR && !analyze->disable_fsst) {
```
Mytherin: Maybe we should set `append_state` to  `DictionaryAppendState::NOT_ENCODED` instead of also checking `disable_fsst` here? `REGULAR` means "we haven't figured out if we want to use FSST or not" - `NOT_ENCODED` means "we know we are not going to use FSST".

Mytherin (commented):

src/storage/compression/dict_fsst/analyze.cpp
sebastiaan-dev: We have this subtraction because we do the following when in `REGULAR` mode, just before checking the available space:

```c++
if (APPEND_STATE == DictionaryAppendState::REGULAR) {
    available_space -= DictFSSTCompression::FSST_SYMBOL_TABLE_SIZE;
}
```

So while we could theoretically fit a string that fills the segment, the writer will reject the value due to bookkeeping for a potential `FSST` encoding. As this PR is mainly a bugfix I didn't want to change too much on the writer side, but the newly introduced guard is a little too conservative wrt the `NOT_ENCODED` path.

Mytherin: Maybe if we determine we can't use FSST we set the append state to `DictionaryAppendState::NOT_ENCODED` instead?

src/storage/compression/dict_fsst/compression.cpp
sebastiaan-dev: Ah yes, that makes sense. This indeed also resolves the weird case where we had to subtract by `DictFSSTCompression::FSST_SYMBOL_TABLE_SIZE` for the plain encoding case. Great!

src/storage/compression/dict_fsst/analyze.cpp
sebastiaan-dev: I've moved the calculation to the constructor of the analyzer, where we now do a binary search for the string limit based on the current block size. This search is then upper-bounded with `DictFSSTCompression::STRING_SIZE_LIMIT`.

src/storage/compression/dict_fsst/analyze.cpp
sebastiaan-dev: Fixed!

src/storage/compression/dict_fsst/analyze.cpp
Mytherin: Why do we need a binary search, isn't this just a single static computation?

Mytherin: I think we can just fill in `string_lengths_width = 32` given that's our max string size, which should then set `string_lengths_space = 8`.

sebastiaan-dev: I used binary search so we know the exact amount of bytes that we can use for a string, whereas using the max string length in DuckDB would slightly overestimate the bytes required for the metadata. So I traded runtime for accuracy, but maybe the loss in supported string length is not large enough to warrant a binary search. I do realize we are recomputing the same value for every row group and `VARCHAR` column, which is not ideal.

Your suggestion did make me think of the fact that instead of the global max string length we can use `DictFSSTCompression::STRING_SIZE_LIMIT` instead, which brings the width to `14` (or `15` when taking into account FSST).

Mytherin: There should also be a closed-form mathematical formula to get to the correct amount of bytes, but being off by 2 bytes here is likely not that important :) In any case I think the binary search shouldn't be required

sebastiaan-dev: Yeah makes sense, changed it!

Mytherin:
Thanks!

PR #23344, opened 2026-06-17
Fix leak on pending query close

evertlammerts:


src/main/pending_query_result.cpp
```diff
@@ -71,6 +75,15 @@ PendingExecutionResult PendingQueryResult::ExecuteTaskInternal(ClientContextLock
...
 	CheckExecutableInternal(lock);
 
+	// Drop the context reference on every exit. Don't call Close() here: it would re-acquire the
+	// already-held context lock, and ~PendingQueryResult may run under that lock on this path.
+	struct ContextRefGuard {
+		shared_ptr<ClientContext> &context;
+		~ContextRefGuard() {
+			context.reset();
+		}
+	} context_guard {context};
```
Mytherin: I'm not really following this - why are we messing with the `context` in the `PendingQueryResult`?

Mytherin (commented):
Thanks - but in my experience this kind of destructor magic that happens outside of the system can cause not-very-fun issues esp when locks are involved. We can keep the addition for `Close` but maybe we can skip the destructor addition?

src/main/pending_query_result.cpp
evertlammerts: This was meant to make sure the context was released even if an exception was thrown, but yes, it's overkill, I removed it.

evertlammerts:
> Thanks - but in my experience this kind of destructor magic that happens outside of the system can cause not-very-fun issues esp when locks are involved. We can keep the addition for `Close` but maybe we can skip the destructor addition?

Yes, you're right of course, this shouldn't have been in the destructor. It's now just in Close(). I've also added it to the StreamQueryResult.

Mytherin:
Thanks!

evertlammerts:
For austerity: this PR makes `PendingQueryResult::Close()` and `StreamingQueryResult::Close()` break a cyclic reference that causes a leak.

The cyclic ref works as follows. A call the `ClientContext::PendingQuery` does the following:

* On the main thread, through `PendingStatementOrPreparedStatement`:
   * set `ClientContext::active_query` (a `unique_ptr<ActiveQueryContext>`) in `BeginQueryInternal`
   * then through `PendingStatementInternal` -> `CreatePreparedStatement` -> `PendingPreparedStatementInternal`:
      * set `active_query->executor = make_uniq<Executor>(*this)`
      * build pipelines with `executor.Initialize(std::move(collector));`
      * then through `Executor::Initialize` -> `InitializeInternal` -> `ScheduleEvents`:
         * create a `make_shared_ptr<PipelineInitializeEvent>` and schedule it
* Then when a worker runs the scheduled `PipelineInitializeEvent`:
   * through `PipelineInitializeTask::ExecuteTask` -> `ResetSink`:
      * gets the sink state through `PhysicalMaterializedCollector::GetGlobalSinkState`, which:
         * creates a `unique_ptr<MaterializedCollectorGlobalState> state`
         * sets a `shared_ptr<ClientContext>` on the state: `state->context = context.shared_from_this();`
         * returns the state
      * then saves the state on `Pipeline`'s `optional_ptr<PhysicalOperator> sink;`

This creates the `shared_ptr` loop, which is essentially:
```
ClientContext → active_query → executor → pipeline ─┐
^───────────────────────────────────────────────────┘
```

Some notes:
* This PR doesn't **fix** the cycle. It just breaks it when `PendingQueryResult::Close()` is called. I'll try to change `state->context` to a `weak_ptr<ClientContext>`, and if that works I'll put up a PR for it.
* `PhysicalMaterializedCollector` has a shared pointer to the context, and that collector is never used for streaming results, so the StreamQueryResult change here is more about hygiene than about the leak.
* I noticed that the cycle _only_ gets created on Linux (maybe because of thread scheduling) and only when a threadpool for workers already exists.

PR #23350, opened 2026-06-18
Deprecate Implicit Table-Function Identifier String Conversion

kryonix:
This PR starts deprecation of automatic column identifier conversion in table function parameters:
```sql
select * from read_csv(lineitem.csv) limit 10;
```

There are three settings now for `table_function_identifier_conversion`:
- `DEFAULT`: warn and keep implicit conversion
- `ENABLE_IMPLICIT_STRING`: keep implicit conversion silently
- `DISABLE_IMPLICIT_STRING`: reject implicit conversion with a binder error

We will keep it on `DEFAULT` for 2.0. Switch it to `DISABLE_IMPLICIT_STRING` with 2.1. And potentially remove the configuration option with 2.2.

fix: https://github.com/duckdblabs/duckdb-internal/issues/9302

### DIFF
--- a/src/common/enum_util.cpp
+++ b/src/common/enum_util.cpp
@@ -71,6 +71,7 @@
 #include "duckdb/common/enums/storage_block_prefetch.hpp"
 #include "duckdb/common/enums/stream_execution_result.hpp"
 #include "duckdb/common/enums/subquery_type.hpp"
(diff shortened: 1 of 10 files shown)


Mytherin:
Thanks!

PR #23352, opened 2026-06-18
Dispatch Rust nightly build

mlafeldt:
For periodic integration testing.

Rust job: https://github.com/duckdb/duckdb-rs/pull/791

To do: figure out if the PAT can trigger duckdb-rs workflows.

refs duckdblabs/duckdb-internal#7516


carlopi (approved):
Thanks!

I think doing as python and passing down also the branch name might be informative, but that can also be reconstructed simply duckdb-rs side via git describe.

mlafeldt:
> I think doing as python and passing down also the branch name might be informative, but that can also be reconstructed simply duckdb-rs side via git describe.

My understanding is that this job is specific to `v1.5-variegata`, which corresponds to `main` in duckdb-rs/duckdb-go right now. There's no branch for v1.5 in either, which is why I set `ref` to `main`.

mlafeldt:
(Double dispatches when merged into main are something we can deal with later, either here or on the Rust side.)

Mytherin:
Thanks!

PR #23353, opened 2026-06-18
New `debug_local_file_system_delay_ms` setting and asynchronous file initialization/finalization

lnkuiper:
This PR adds a local-file debug delay that can be used to test async IO on a local machine. I've used this setting to identify and fix bottlenecks in the `COPY TO` implementation.

## Debug Local File System Delay

This new setting, `debug_local_file_system_delay_ms`, has a default value of `0`. When enabled, the Unix local file system sleeps before open/read/write operations, which can be used as a proxy for a remote file system.

When `debug_local_file_system_delay_ms > 0`, `AsyncFileWriter::IsLocalFile()` returns `false`. This makes delayed local writes use the same coalescing behavior as remote writes, which is important for the proxy: otherwise the local writer path still behaves too differently from remote files even though each local operation sleeps. I considered making `FileSystem::IsRemoteFile()` return `true` when this setting is activated but I thought that might be too invasive.

## Async COPY File Lifecycle

This PR also changes `COPY TO` so file initialization and finalization can run on the async task scheduler instead of blocking regular threads.

The main change is that file state creation is split into two phases:

1. A lock-held reservation phase records the output path, created-file metadata, partition values, written-file info, and stats state.
2. An async lifecycle phase runs `copy_to_initialize_global`, initializes the writer state, and later runs `copy_to_finalize`.

`FileStateHandle` is the shared object between those phases. A handle can reference either a pending open job or a ready `GlobalFileState`, and call sites use `EnsureFileStateReady` when they actually need the writer state. Finalization uses the same lifecycle executor and waits for a pending open before finalizing the file.

The lifecycle executor uses `TaskExecutor(context, TaskSchedulerType::ASYNC)`, bounds the number of queued lifecycle tasks, records the first exception, and lets waiters help drain lifecycle work (if things aren't ready or if work has built up). Query finalization and result production call `WaitForLifecycleTasks()` so returned file/stats information and temporary-file renames only happen after all async open/finalize work has completed.

The life cycle implementation is shared by all write paths, regardless of the files being remote, partitioned, etc. Partitioned file-state publication happens under the partition/global locks, while scheduling and finalization are done after releasing those locks.

There is something to be said about these tasks being scheduled on async threads even though they do more than just IO. I think this is necessary because we're calling opaque callbacks that can be implemented in extensions. This PR achieves asynchronous behaviour cleanly without modifying extensions/callbacks.

## Results / Benchmarks

These are runs on my laptop with TPC-H SF 10 with `debug_local_file_system_delay_ms=1` and `SET threads=8;`.

| Query | Execution Time [s] |
| --- | ---: |
| `PARTITION_BY (l_quantity)` (~50 partitions) | 6.51 |
| `PARTITION_BY (l_shipdate)` (~2.5k partitions) | 8.65 |
| `PARTITION_BY (l_suppkey)` (~100k partitions) | 8.63 |
| Single parquet file | 3.87 |
| `FILE_SIZE_BYTES '32MiB'` (62 files) | 3.70 |

Without the life cycle changes or with `async_threads=0`, all of these queries could take 20 to 30 seconds.


dentiny:
Hi Laurens, I'm wondering if you think possible to apply this debug setting to all filesystems, instead of just local filesystem? It's useful to (1) test out possible race condition, (2) simulate network latency for local benchmark purpose.
I actually wrote an extension for httpfs: https://duckdb.org/community_extensions/extensions/latency_injection_fs

lnkuiper:
Hi @dentiny I think that should be possible by wrapping the FS like you did in the extension, but then it's harder to turn on/off for `FileSystem` instances with a longer lifetime. But since it's for debugging purposes (and named as such) users can't really expect much from this setting, so maybe that's OK. Happy to discuss if you'd like to pick this up

dentiny:
What I have in mind is a `DebugFileSystem` (maybe a better name) which behaves similar to [`OpenerFileSystem`](https://github.com/duckdb/duckdb/blob/main/src/include/duckdb/common/opener_file_system.hpp) which does everything necessary before delegating to virtual filesystem then actual filesystem.

A few things in mind for debugging/troubleshooting purpose: injected latency, controlled throughput, error injection (to see whether DuckDB throws INTERNAL exception on IO failure), etc.

lnkuiper:
Sounds great!

PR #23354, opened 2026-06-18
[MultiFileReader] Prevent NULL MAP keys because of (likely missing) default value

Tishj:
This PR is related to https://github.com/duckdb/duckdb-iceberg/issues/903

The parquet file in question is missing field-ids for the `key` and `value` parts of the MAP, which result in getting substituted with NULL in the MultiFileReader, which then isn't caught when producing the final chunk.

We should likely also check the correctness of the produced chunk in `MultiFileReader::FinalizeChunk`
Though I imagine that's up to the reader implementation to verify, no?
That a MAP key is non-null


Mytherin:
Thanks!

PR #23355, opened 2026-06-18
Remove `TransformEnum`

Dtenwolde:
In the last PR I forgot to clean up some stuff. Also, now we don't use `TransformEnum` anymore, we just have simple one-line `Transform` rules instead to keep it consistent, so that part can also be removed


Mytherin:
Thanks!

PR #23356, opened 2026-06-18
`CompressedMaterialization` and `JoinFilterPushdown` for `VARIANT`

lnkuiper:
This PR implements `CompressedMaterialization` for fully-shredded `VARIANT` columns. The stats will tell us if it's fully shredded, which allows us to do join/agg/sort on the shredded type by pushing a `CAST`. I had to change my PR a bit after https://github.com/duckdb/duckdb/pull/23307 to strip off the `variant_comparator` function when possible, but in general the `CompressedMaterialization` infrastructure was able to deal with `VARIANT` quite well with only a few modifications.

I also added `JoinFilterPushdown` for `VARIANT` through `VARIANT`/integral casts. Similar to how we already allow for `JoinFilterPushdown` through integral casts, we can allow this through casts (and `variant_comparator`) for joins to do dynamic join filtering on the native integer type.


src/function/scalar/variant/variant_normalize.cpp
```diff
@@ -260,9 +260,15 @@ static void VariantNormalizeFunction(DataChunk &input, ExpressionState &state, V
...
 
+static unique_ptr<BaseStatistics> VariantNormalizeStats(ClientContext &context, FunctionStatisticsInput &input) {
+	// variant_normalize re-encodes the VARIANT in a canonical binary form, but does not change any values
+	return input.child_stats[0].ToUnique();
+}
+
 ScalarFunction VariantNormalizeFun::GetFunction() {
 	auto variant_type = LogicalType::VARIANT();
-	return ScalarFunction("variant_normalize", {variant_type}, variant_type, VariantNormalizeFunction);
+	return ScalarFunction("variant_normalize", {variant_type}, variant_type, VariantNormalizeFunction, nullptr,
```
Mytherin, line 270: Unrelated to this PR but I think we might be able to remove `variant_normalize` now that we have the comparator. CC @Tishj is that accurate or is this still used elsewhere?

Tishj: We have a vector verification that uses it, but only for the same reason of comparing values regardless of object order

There was a worry that variant_extract could cause the bulk of the variant data to stay allocated, due to the way we construct a dictionary vector to make extract a zero(ish) copy operation

Normalize acts as a rewrite of the data, so that could be used to counteract this
But whether that's still relevant, I don't know

Tishj: More unrelated, but I found a call to `VariantUtils::ConvertVariantToValue` in a value comparison path that *should* now be dead code
I was thinking of removing it, but I first want to test if iceberg doesn't hit this still somehow

Which requires 1.5 to get merged into main first (for variant_to_parquet_variant_bytes)

PR #23360, opened 2026-06-18
Add dictionary surviving hash join

ahmetkarapinar:
# Add Dictionary Surviving Hash Join

## Motivation

DuckDB's hash-join build side is **not dictionary-aware**. As each chunk arrives at the sink, `Scatter` materializes every payload column into the row store at its native physical width. When the column already arrived dictionary-encoded over a small set of distinct values, this throws the encoding away: the same few values are copied into the row store once per build row, and their string heap is re-materialized on top of the upstream child that already holds it. That wastes both **memory** and **CPU**.

## Global Dictionaries

We introduce a new concept: a **global dictionary**. A dictionary is global when an operator builds it once and guarantees that *every* chunk it emits for that column carries the exact same dictionary for the operator's entire lifetime. No chunk is ever emitted in a flat form or with a different dictionary. That guarantee is what makes it safe for a downstream sink to commit to the dictionary after seeing only the **first** chunk, knowing every later chunk will match it.

Two operators already produce such dictionaries:

- **Perfect Hash Join**: `perfect_hash_table[i]` is wrapped by every probe-output chunk for the operator's lifetime.
- **Small-build-side dictionary emission** (#22340): `dict_arrays[col_idx]` is wrapped by every matched-row emission once activated.

Storage-emitted dictionaries do **not** qualify (they rotate per row group and fall back to flat), so they are excluded for now.

## Approach

When a build-side payload column arrives at a hash-join sink as a global dictionary, the sink stores a compact **index** into the dictionary instead of flattening the value:

- On the **first** build chunk, the sink inspects the live vectors, narrows each eligible payload column's row-store slot to the smallest index width that can address its dictionary, and pins a self-owned copy of the dictionary child (the producer's string heap is recycled when its pipeline finishes, so the copy must outlive it).
- During the build, `Scatter` writes the 1/2/4-byte index per row instead of the full value.
- At probe time, `GatherRHS` reads the index and re-emits a `DICTIONARY_VECTOR` over the pinned child, preserving the same `dictionary_id`, so the dictionary survives the join.

**Index width** is chosen per column from the dictionary size `D`: `D ≤ 256 → 1 B` (`UTINYINT`), `D ≤ 65536 → 2 B` (`USMALLINT`), else `4 B` (`UINTEGER`). The eligibility gate is `index_width(D) < native_width`, so a row can only ever shrink, never grow. This also unlocks narrow non-string payloads (a `D ≤ 256` `INTEGER` narrows 4 B → 1 B).

Because the dictionary survives intact, the optimization **chains**: the re-emitted output of a dictionary-surviving join is itself a first-class global producer, so a chain of consecutive joins can carry one dictionary end-to-end under a single id.

## Two supporting changes

Two existing parts of DuckDB had to change for a global dictionary to reach the sink intact.

### 1. Delayed `TupleDataLayout` construction

To store an index instead of a full value, the slot width must be fixed **before the first row is written**, because everything afterward reads from fixed offsets. But the layout is currently built in the `JoinHashTable` **constructor**, from logical types alone, before any chunk exists, so it cannot know a column will arrive dictionary-encoded or how large its dictionary is.

The fix splits `JoinHashTable` initialization into two phases. The constructor now does only the work that does *not* depend on the `TupleDataLayout`; everything that does is moved into a new `FinishInitWithLayout`, which runs once the first chunk has arrived:

1. The first thread to reach `Sink` becomes the **publisher**: it inspects the arriving vectors, picks each payload column's slot width, and builds one canonical `TupleDataLayout`.
2. It stores the layout in a `LayoutGate` (the layout plus an `atomic<bool> published`, double-checked under a mutex). Every other build thread reads the *same* layout rather than building its own.
3. Because every per-thread `JoinHashTable` shares the one published layout, all threads write rows in an identical byte format, so the parallel `Merge` can splice row blocks together without re-encoding.

### 2. Global-Dictionary-aware caching operator

Every `PhysicalHashJoin` is a `CachingPhysicalOperator` (via `PhysicalJoin`), so every hash join's probe output passes through a caching layer that buffers and combines small chunks before reaching the next sink. The cache is built type-blind and flat, so it flattens the dictionary on append, dropping the `buffer_ptr`, the `id`, and the `pipeline_global` flag.

This **breaks the global-dictionary guarantee**: the cache emits a fresh, unrelated dictionary on flush, while the sink has already pinned the original entry from the first chunk, so the sink would scatter indices that point into a *different* child than the one it gathers from.

The fix teaches the caching operator to recognize a global dictionary and preserve it. Each cached column is handled one of two ways:

- **Flat column (common case):** cached exactly as before, behind a single `dict_cache_active` flag, so non-dictionary workloads are byte-for-byte the old path.
- **Global dict column:** the cache keeps a reference to the upstream dictionary and appends each chunk's indices into a running selection vector; on flush it re-emits a `DICTIONARY_VECTOR` over that same dictionary, so `id` and `pipeline_global` are preserved automatically.

This is also cheaper per row than flattening.

## Activation conditions

Checked in two layers; any failure falls through silently to the unchanged path.

**Join-level:** the join must not be external, its type must not be `SINGLE` / `LEFT` / `OUTER`, its build side must not be fed by multiple sources, it must not use Perfect Hash Join at finalize, and its RHS output must be non-empty.

**Per-column:** the column must arrive as a **Global** `DICTIONARY_VECTOR`, be of a non-nested type, not be read by a residual predicate, and have an index width strictly smaller than its native width.

## Benchmarks

Single-threaded, 10 runs per query, comparing this branch against `main`. Peak memory is the buffer-pool high-water mark over the whole query (an `atomic<int64_t>` added to the buffer pool, reset at query start, read at query end); we report the minimum across runs.

### IMDB / Join Order Benchmark

Top queries by peak-memory reduction, with runtime:

| Query | main peak (MB) | branch peak (MB) | peak reduction | main (s) | branch (s) | speedup |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 07c | 4753.88 | 827.53 | 82.6% | 0.435 | 0.185 | **2.35×** |
| 13c | 155.09 | 102.36 | 34.0% | 0.103 | 0.094 | 1.09× |
| 13b | 157.97 | 105.76 | 33.1% | 0.116 | 0.107 | 1.08× |
| 13d | 97.70 | 70.65 | 27.7% | 0.098 | 0.093 | 1.06× |
| 16c | 300.39 | 284.99 | 5.1% | 0.169 | 0.162 | 1.04× |
| 16d | 290.53 | 276.56 | 4.8% | 0.159 | 0.154 | 1.03× |
| 20a | 336.24 | 326.53 | 2.9% | 0.311 | 0.303 | 1.03× |

07c is the flagship: a 12.6k-name dictionary referenced by 1.6M build rows, peak dropping from 4.75 GB to 0.83 GB and the query running 2.35× faster.

### TPC-H

The clearest TPC-H result is **Q21 at SF100**, where `s_name` survives three consecutive joins as a global dictionary.

Peak memory (10 runs, fully deterministic so min = median = mean):

| Metric | main (MB) | branch (MB) | reduction (MB) | reduction % |
| Min | 8248.10 | 7573.60 | 674.50 | 8.18% |

Runtime (10 timed runs, real seconds):

| Metric | main | branch | reduction | speedup |
| Geomean | 11.877 s | 11.578 s | 2.52% | **1.026×** |

### Showcase: avoiding a spill

A large enough reduction flips a spilling join into an in-memory one, and the disk-spill penalty far outweighs any per-row saving.

```sql
SET threads = 1;
SET disabled_optimizers = 'join_order';
SET memory_limit = '600MB';

CREATE TABLE nation AS
  SELECT i AS n_nationkey,
         'NATION_NAME_…_LONG_COUNTRY_NAME_…_'      || (i % 5) AS n_name,
         'NATION_COMMENT_…_SUBSTANTIAL_PADDING_…_' || (i % 7) AS n_comment
  FROM range(25) t(i);
CREATE TABLE customer AS
  SELECT i * 100 AS c_custkey, i % 25 AS c_nationkey FROM range(2000000) t(i);
CREATE TABLE orders AS
  SELECT i AS o_orderkey, (i % 2000000) * 100 AS o_custkey FROM range(6000000) t(i);

SELECT n.n_name, n.n_comment
FROM nation n
JOIN customer c ON n.n_nationkey = c.c_nationkey   -- Perfect Hash Join (dict producer)
JOIN orders   o ON o.o_custkey   = c.c_custkey;     -- dictionary-surviving join fires here
```

The build-side row store goes from **641 MB on `main`** (> the ~600 MB reservation → spills to disk) **to 68 MB on this branch** (fits, stays in memory), a 9.5× reduction that flips the spill decision:

| Metric | main | this branch | reduction | speedup |
| Min | 0.362 s | 0.246 s | 32.1% | **1.47×** |
| Geomean | 0.412 s | 0.266 s | 35.3% | **1.55×** |

---

@lnkuiper, PR is ready to review, Thanks!

### DIFF
--- a/src/common/types/vector.cpp
+++ b/src/common/types/vector.cpp
@@ -188,6 +188,8 @@ void Vector::Reinterpret(const Vector &other) {
 		new_vector.Reinterpret(DictionaryVector::Child(other));
 		auto &old_dict = buffer->Cast<DictionaryBuffer>();
 		auto new_entry = make_shared_ptr<DictionaryEntry>(std::move(new_vector));
+		// reinterpret re-mints the entry, so carry the flag across by hand
+		new_entry->pipeline_global = old_dict.GetEntry().pipeline_global;
 		buffer = make_buffer<DictionaryBuffer>(old_dict.GetSelVector(), old_dict.Capacity(), std::move(new_entry));
 	}
 }
--- a/src/common/types/vector.cpp
+++ b/src/common/types/vector.cpp
@@ -188,7 +188,8 @@ void Vector::Reinterpret(const Vector &other) {
 		new_vector.Reinterpret(DictionaryVector::Child(other));
 		auto &old_dict = buffer->Cast<DictionaryBuffer>();
 		auto new_entry = make_shared_ptr<DictionaryEntry>(std::move(new_vector));
-		// reinterpret re-mints the entry, so carry the flag across by hand
+		// reinterpret re-mints the entry; the id and global flag are one contract and must survive together
+		new_entry->id = old_dict.GetEntry().id;
 		new_entry->global_dictionary = old_dict.GetEntry().global_dictionary;
 		buffer = make_buffer<DictionaryBuffer>(old_dict.GetSelVector(), old_dict.Capacity(), std::move(new_entry));
 	}
--- a/src/common/types/vector.cpp
+++ b/src/common/types/vector.cpp
@@ -189,7 +189,7 @@ void Vector::Reinterpret(const Vector &other) {
 		auto &old_dict = buffer->Cast<DictionaryBuffer>();
 		auto new_entry = make_shared_ptr<DictionaryEntry>(std::move(new_vector));
 		// reinterpret re-mints the entry, so carry the flag across by hand
-		new_entry->pipeline_global = old_dict.GetEntry().pipeline_global;
+		new_entry->global_dictionary = old_dict.GetEntry().global_dictionary;
 		buffer = make_buffer<DictionaryBuffer>(old_dict.GetSelVector(), old_dict.Capacity(), std::move(new_entry));
 	}
 }
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -165,7 +165,7 @@ void JoinHashTable::Merge(JoinHashTable &other) {
 
 	sink_collection->Combine(*other.sink_collection);
 
-	// Reconcile per-column pinned dictionary entries. For pipeline-global producers every thread pins the
+	// Reconcile per-column pinned dictionary entries. For global dictionary producers every thread pins the
 	// same buffer_ptr, so this is normally a no-op or a "take theirs" adoption.
 	if (dict_registry.size() < other.dict_registry.size()) {
 		dict_registry.resize(other.dict_registry.size(), nullptr);
@@ -479,7 +479,7 @@ uint8_t JoinHashTable::GetDictSurvivingIndexWidth(idx_t build_col_idx, const Vec
 	if (incoming.GetVectorType() != VectorType::DICTIONARY_VECTOR) {
 		return 0;
 	}
-	if (!DictionaryVector::IsPipelineGlobal(incoming)) {
+	if (!DictionaryVector::IsGlobalDictionary(incoming)) {
 		return 0;
 	}
 	// an empty/invalid dictionary keeps native width
@@ -543,9 +543,9 @@ void JoinHashTable::Build(PartitionedTupleDataAppendState &append_state, DataChu
 			// Slot was narrowed on the first chunk; no fallback for a non-dict later chunk. Throw, not D_ASSERT:
 			// the Cast<DictionaryBuffer> below is UB in release on a non-dict vector, scattering foreign bytes.
 			if (incoming.GetVectorType() != VectorType::DICTIONARY_VECTOR ||
-			    DictionaryVector::DictionaryId(incoming).empty() || !DictionaryVector::IsPipelineGlobal(incoming)) {
+			    DictionaryVector::DictionaryId(incoming).empty() || !DictionaryVector::IsGlobalDictionary(incoming)) {
 				throw InternalException("dict-surviving join: narrowed column %llu received a "
-				                        "non-pipeline-global chunk; build pipeline is not single-source",
+				                        "non-global-dictionary chunk; build pipeline is not single-source",
 				                        static_cast<uint64_t>(i));
 			}
 			// Pin the dictionary on the first chunk; enforce id continuity on subsequent chunks
@@ -559,7 +559,7 @@ void JoinHashTable::Build(PartitionedTupleDataAppendState &append_state, DataChu
 				// producer that grows the child past it fails loudly instead of truncating in the UnsafeNumericCast.
 				D_ASSERT(child_count <= (idx_t(1) << (8 * index_width)));
 				auto owned_entry =
-				    DictionaryVector::CreateReusablePipelineGlobalDictionary(upstream_child.GetType(), child_count);
+				    DictionaryVector::CreateReusableGlobalDictionary(upstream_child.GetType(), child_count);
 				if (child_count > 0) {
 					VectorOperations::Copy(upstream_child, owned_entry->data, child_count, 0, 0);
 				}
@@ -2679,16 +2679,15 @@ void JoinHashTable::BuildDictionaryArrays(const PhysicalHashJoin &op) {
 	const auto row_ptrs = FlatVector::GetData<data_ptr_t>(row_pointer_vector);
 
 	// LEFT / OUTER joins fill unmatched probe rows with a constant-NULL vector (NextLeftJoin), so they do not
-	// wrap this entry on every chunk; it is only pipeline-global when every chunk goes through EmitDictVectors.
+	// wrap this entry on every chunk; it is only a global dictionary when every chunk goes through EmitDictVectors.
 	const bool dict_on_every_chunk = join_type != JoinType::LEFT && join_type != JoinType::OUTER;
 
 	// gather RHS output columns into columnar dictionary arrays
 	const auto &sel = *FlatVector::IncrementalSelectionVector();
 	for (idx_t col_idx = 0; col_idx < op.rhs_output_columns.col_types.size(); col_idx++) {
 		const auto &type = op.rhs_output_columns.col_types[col_idx];
-		auto dict_entry = dict_on_every_chunk
-		                      ? DictionaryVector::CreateReusablePipelineGlobalDictionary(type, build_count)
-		                      : DictionaryVector::CreateReusableDictionary(type, build_count);
+		auto dict_entry = dict_on_every_chunk ? DictionaryVector::CreateReusableGlobalDictionary(type, build_count)
+		                                      : DictionaryVector::CreateReusableDictionary(type, build_count);
 		const auto output_col_idx = output_columns[col_idx];
 		collection.Gather(row_pointer_vector, sel, build_count, output_col_idx, dict_entry->data, sel, nullptr);
 		dict_arrays.emplace_back(std::move(dict_entry));
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -83,18 +83,29 @@ JoinHashTable::JoinHashTable(ClientContext &context_p, const PhysicalOperator &o
 	// at least one equality is necessary
 	D_ASSERT(!equality_types.empty());
 
-	// Types for the layout
-	auto layout = make_shared_ptr<TupleDataLayout>();
-	vector<LogicalType> layout_types(condition_types);
-	layout_types.insert(layout_types.end(), build_types.begin(), build_types.end());
-	if (PropagatesBuildSide(join_type)) {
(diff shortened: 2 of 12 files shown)


src/include/duckdb/common/vector/dictionary_vector.hpp
```diff
@@ -151,6 +154,16 @@ struct DictionaryVector {
 		return DictionarySize(vector).IsValid() && !DictionaryId(vector).empty() && CanCacheHashes(vector.GetType());
 	}
 	static buffer_ptr<DictionaryEntry> CreateReusableDictionary(const LogicalType &type, const idx_t &size);
+	//! Mint a reusable dictionary entry whose lifetime spans the entire producing operator instance
+	static buffer_ptr<DictionaryEntry> CreateReusablePipelineGlobalDictionary(const LogicalType &type,
+	                                                                          const idx_t &size);
+	//! True iff vector is a DICTIONARY_VECTOR whose entry is pipeline-global
+	static inline bool IsPipelineGlobal(const Vector &vector) {
```
Mytherin: This `PipelineGlobal` sounds a bit specific to me - can this be just a `GlobalDictionary` instead?

src/execution/physical_operator.cpp
```diff
@@ -422,6 +416,119 @@ static CachingPhysicalOperatorExecuteMode SelectExecutionMode(const DataChunk &c
...
+		// no dict columns: plain flat append
+		cache.Append(source);
+		return;
+	}
+	const idx_t base = cache.size();
+	const idx_t added = source.size();
+	// accumulated_sel is sized STANDARD_VECTOR_SIZE and the caching state machine guarantees base + added stays
+	// within it. Index accumulation has no overrun guard of its own (unlike the flat Append), so assert it.
+	D_ASSERT(base + added <= STANDARD_VECTOR_SIZE);
+	for (idx_t col_idx = 0; col_idx < cache.ColumnCount(); col_idx++) {
```
Mytherin, line 466: I wonder to what extend this cache code relies on (pipeline) global dictionaries, versus just being useful with repeated dictionaries in general.

Could we not do this type of dictionary-specialized accumulation here for all dictionaries that have unique identifiers where we just flatten if the dictionary uuid changes out from under us (where this will just not happen for these pipeline global dictionaries)?

That could then (1) be a separate PR, and (2) would benefit all repeated dictionaries (including Parquet dictionaries, dictionaries from storage, etc).

Mytherin (commented):
Thanks for the PR! Cool work - left some initial comments.

src/execution/physical_operator.cpp
ahmetkarapinar: Thanks for the review @Mytherin ! That's a great idea. We can generalize this approach to any dictionary with a stable id, if the id changes, we just flatten what's been accumulated so far and continue. Since (pipeline) global dictionaries use a single id throughout, they're a natural fit and the mechanism works for them without modification. Other dictionary types would also benefit. Happy to do this as a follow-up PR after this one lands.

src/include/duckdb/common/vector/dictionary_vector.hpp
ahmetkarapinar: Done! Changed every PipelineGlobal reference to GlobalDictionary.

src/execution/join_hashtable.cpp
```diff
@@ -427,8 +532,63 @@ void JoinHashTable::Build(PartitionedTupleDataAppendState &append_state, DataChu
 	}
 	idx_t col_offset = keys.ColumnCount();
 	D_ASSERT(build_types.size() == payload.ColumnCount());
+	// Scratch index vectors for narrowed columns, allocated via the buffer manager so the bytes are accounted.
+	// buffers own the bytes, vectors are flat views; both must outlive the ToUnifiedFormat / AppendUnified below.
+	vector<AllocatedData> dict_idx_buffers;
+	vector<Vector> dict_idx_vectors;
 	for (idx_t i = 0; i < payload.ColumnCount(); i++) {
```
lnkuiper, line 601: Before this loop used to be
```C++
	for (idx_t i = 0; i < payload.ColumnCount(); i++) {
		source_chunk.data[col_offset + i].Reference(payload.data[i]);
    }
```
but now it's gotten quite long. Could you look into separating some of this logic into helper functions?

src/execution/join_hashtable.cpp
```diff
@@ -1438,10 +1598,45 @@ void JoinHashTable::GatherRHS(Vector &row_ptrs, const SelectionVector &ptr_sel,
...
+					break;
+				case sizeof(uint16_t):
+					LoadDictIndices<uint16_t>(ptrs, ptr_sel, count, col_offset, build_sel_vec);
+					break;
+				default:
+					LoadDictIndices<uint32_t>(ptrs, ptr_sel, count, col_offset, build_sel_vec);
+					break;
+				}
+				vector.Dictionary(dict_registry[payload_idx], build_sel_vec, count);
+				continue;
```
lnkuiper, line 1641: Could we do the `switch` in a helper function `LoadDictIndices` which then calls the templated function?

lnkuiper (commented):
Thanks for the PR! I've two minor code organization comments below. I also saw that these tests are failing in CI (configs/vector sizes):
```
test/sql/cte/recursive_cte_key_hll_aggregation.test
test/optimizer/joins/delim_left_join_pushdown.test
```
Could you check if these are related?

I also saw that this one is failing:
```
build/release/test/unittest --test-config test/configs/internal_vector_verification.json test/sql/copy/partitioned/partitioned_order_by.test
```
But that's probably my fault! I'll have a look at that myself

src/execution/operator/join/physical_hash_join.cpp
```diff
@@ -479,6 +538,124 @@ class HashJoinLocalSinkState : public LocalSinkState {
...
+	layout->Initialize(layout_types, TupleDataValidityType::CAN_HAVE_NULL_VALUES);
+	return layout;
+}
+
+//! Join-level gate: shape-only eligibility checks, mirroring the dict-emission path plus a PHJ exclusion
+static bool CanUseDictSurvivingJoin(const PhysicalHashJoin &op, const JoinHashTable &ht, bool external,
+                                    bool can_use_perfect_hash, bool build_side_multi_source) {
+	// external joins rebuild partitions; pinned upstream entries cannot survive the rebuild
+	if (external) {
+		return false;
```
lnkuiper: This only gates against `SET debug_force_external=true;`, the actual decision to do an external join happens later.

I think it should be possible to use this optimisation even when doing external joins. Maybe this is already tested. Could you have a look and maybe write a test?

src/common/types/vector.cpp
```diff
@@ -188,6 +188,8 @@ void Vector::Reinterpret(const Vector &other) {
 		new_vector.Reinterpret(DictionaryVector::Child(other));
 		auto &old_dict = buffer->Cast<DictionaryBuffer>();
 		auto new_entry = make_shared_ptr<DictionaryEntry>(std::move(new_vector));
+		// reinterpret re-mints the entry, so carry the flag across by hand
+		new_entry->global_dictionary = old_dict.GetEntry().global_dictionary;
```
lnkuiper, line 193: This keeps `global_dictionary` but drops the dictionary ID. Maybe it should keep the dictionary ID too?

ahmetkarapinar:
Thanks for the review @lnkuiper! I addressed the comments and rebased onto the latest main.

About `test/sql/cte/recursive_cte_key_hll_aggregation.test`, it turned out this one was actually related to this PR. Under `internal_vector_verification`, the caching operator's flush/append path was leaving a stale DICTIONARY_VECTOR in a chunk whose cardinality had been reset to 0.

Fix: in `CachingPhysicalOperator::Execute`, when about to return an empty chunk, flatten any stale non-flat / non-constant columns. The fix only runs when the result is empty, so real dictionary emissions (which have rows) are not affected.

lnkuiper:
Thanks for the changes! LGTM from my side now

lnkuiper:
Looks like there is a formatting issue now. Could you have a look?

ahmetkarapinar:
Found one more bug related to the new caching operator. After a cache flush, a stale dictionary column was kept in the output chunk, causing the probe's termination check to see a non-zero row count and loop indefinitely (only reproducible with STANDARD_VECTOR_SIZE=2). Fix: build the dict cache slots with real, resettable vectors instead of zero-width placeholders. DataChunk::Reset() can then clear the column to size 0 and the probe terminates correctly.

CI is green now 🚀

lnkuiper:
Thanks for the changes! Queued up to merge now :)

PR #23368, opened 2026-06-18
Fix a fitering bug.

keithjjones:
I ran into this while working on the DuckDB Zeek extension.  I used AI to try to make a reproducer that didn't depend on the Zeek extension, and it kept telling me a C++ example like this was best.  I also used AI on the fix.  I hope this MR is helpful:

# DuckDB Bug Report: IN on unsupported filter types corrupts projection_ids

## Symptom

When a table function sets `projection_pushdown = true` AND has
`supports_pushdown_type` that returns `false` for some column type,
queries with `IN` filters on that column using **multi-column projections**
produce wrong results or crash.

Single-column projections and `=` filters on the same column work fine.

## SQL Reproducer

No built-in DuckDB function sets `supports_pushdown_type` (only Arrow and
custom extensions do). A self-contained C++ reproducer is provided below,
or the `inet` community extension can be used with any extension that
provides a table function with `supports_pushdown_type`, like the `zeek` 
extension.

## C++ Reproducer

Compile and run against any DuckDB:

```cpp
#include "duckdb.hpp"
#include "duckdb/parser/parsed_data/create_table_function_info.hpp"
#include <iostream>
using namespace duckdb;

struct Fn {
	struct D : public TableFunctionData { bool done = false; };
	static unique_ptr<FunctionData> Bind(ClientContext &, TableFunctionBindInput &,
	                                     vector<LogicalType> &ret, vector<string> &n) {
		ret.push_back(LogicalType::INTEGER); n.push_back("id");
		ret.push_back(LogicalType::VARCHAR); n.push_back("val");
		return make_uniq<D>();
	}
	static void F(ClientContext &, TableFunctionInput &d, DataChunk &o) {
		auto &data = (D &)*d.bind_data;
		if (!data.done) {
			o.SetValue(0, 0, Value(42));
			o.SetValue(1, 0, Value("hello"));
			o.SetCardinality(1);
			data.done = true;
		}
	}
	static bool S(const FunctionData &, idx_t c) { return c != 1; }
};

int main() {
	DuckDB db(nullptr);
	Connection con(db);
	con.BeginTransaction();
	auto &ctx = *con.context;
	auto &cat = Catalog::GetSystemCatalog(ctx);
	TableFunction fn("pd_test", {}, Fn::F, Fn::Bind);
	fn.supports_pushdown_type = Fn::S;
	fn.projection_pushdown = true;
	auto info = CreateTableFunctionInfo(fn);
	cat.CreateTableFunction(ctx, info);
	con.Commit();
	auto r = con.Query("SELECT id, val FROM pd_test() WHERE val IN ('hello');");
	if (r->HasError()) { std::cerr << "CRASH: " << r->GetError() << "\n"; return 1; }
	r->Print();
	return 0;
}
```

```
# Unpatched:
$ c++ -std=c++17 repro.cpp -L/path/to/duckdb/lib -lduckdb -o repro && ./repro
CRASH: Invalid Input Error: Failed to cast value: Could not convert string 'hello' to INT32

# Patched:
$ ./repro
42	hello
```

## Root Cause

**File:** `src/execution/physical_plan/plan_get.cpp`, lines 129-132

In DuckDB, an empty `projection_ids` vector is the identity mapping convention:
all columns in `column_ids` are output columns (see `LogicalGet::ResolveTypes()`,
line 168). When `plan_get.cpp` extracts unsupported table filters and converts
them to `PhysicalFilter` nodes, the code unconditionally pushes to
`projection_ids`:

```cpp
if (!found_projection) {
    projection_ids.push_back(entry.first);     // BUG: breaks identity mapping
    column_id_filter = projection_ids.size() - 1;
}
```

When `projection_ids` is empty (identity), this changes the semantics from
"all columns" to "only the filter column". `op.ResolveOperatorTypes()` then
recalculates `op.types` from the corrupted `projection_ids`, silently dropping
all non-filter columns from the scan output. The downstream operators disagree
on column count, causing out-of-bounds access.

### Contributing factor: `filter_combiner.cpp`

The `=` operator's pushdown path checks `TypeSupportsConstantFilter(type)` and
returns `NO_PUSHDOWN` for unsupported types. But `TryPushdownInFilter` only
checks this for the single-element shortcut; the multi-element path
unconditionally creates an `InFilter`, even for unsupported types. This
inconsistency pushes `IN` filters as table filters, triggering the bug. `=`
filters remain as expressions and avoid it.

### Fix 1: `plan_get.cpp`

```cpp
if (!found_projection) {
    if (projection_ids.empty()) {
        column_id_filter = entry.first;
    } else {
        projection_ids.push_back(entry.first);
        column_id_filter = projection_ids.size() - 1;
    }
}
```

Also fix `filter_types` construction for `PhysicalFilter`:

```cpp
if (projection_ids.empty()) {
    for (auto &col_id : column_ids) {
        // all columns in identity mapping
        ...
    }
} else {
    for (auto &c : projection_ids) {
        // existing logic
    }
}
```

### Fix 2: `filter_combiner.cpp`

```cpp
if (!TypeSupportsConstantFilter(type)) {
    return FilterPushdownResult::NO_PUSHDOWN;
}
```

## Affected Files

| File | Change |
|------|--------|
| `src/execution/physical_plan/plan_get.cpp` | Fix `projection_ids` identity mapping and `filter_types` construction |
| `src/optimizer/filter_combiner.cpp` | Add `TypeSupportsConstantFilter` guard to multi-element `IN` path |
| `test/sql/function/table/table_filter_pushdown_projection.cpp` | C++ test registering a table function with `supports_pushdown_type`, verifying multi-column `IN` filter works |
| `test/sql/function/table/CMakeLists.txt` | Added new test file to build |
| `test/sql/filter/inet_filter_projection.test` | SQL regression test using `require inet` |

### DIFF
--- a/src/execution/physical_plan/plan_get.cpp
+++ b/src/execution/physical_plan/plan_get.cpp
@@ -127,8 +127,13 @@ PhysicalOperator &PhysicalPlanGenerator::CreatePlan(LogicalGet &op) {
 					}
 				}
 				if (!found_projection) {
-					projection_ids.push_back(entry.first);
-					column_id_filter = projection_ids.size() - 1;
+					if (projection_ids.empty()) {
+						// Identity mapping: column is already in output at position entry.first
+						column_id_filter = entry.first;
+					} else {
+						projection_ids.push_back(entry.first);
+						column_id_filter = projection_ids.size() - 1;
+					}
 				}
 				auto column = make_uniq<BoundReferenceExpression>(column_type, column_id_filter);
 				select_list.push_back(entry.second->ToExpression(*column));
@@ -141,13 +146,26 @@ PhysicalOperator &PhysicalPlanGenerator::CreatePlan(LogicalGet &op) {
 
 		if (!select_list.empty()) {
 			vector<LogicalType> filter_types;
-			for (auto &c : projection_ids) {
-				auto column_id = column_ids[c].GetPrimaryIndex();
-				if (IsVirtualColumn(column_id)) {
-					auto &column = virtual_columns.at(column_id);
-					filter_types.push_back(column.type);
-				} else {
-					filter_types.push_back(op.returned_types[column_id]);
+			if (projection_ids.empty()) {
+				// Identity mapping: scan outputs all column_ids, filter passes them all through
+				for (auto &col_id : column_ids) {
+					auto primary_id = col_id.GetPrimaryIndex();
+					if (IsVirtualColumn(primary_id)) {
+						auto &column = virtual_columns.at(primary_id);
+						filter_types.push_back(column.type);
+					} else {
+						filter_types.push_back(op.returned_types[primary_id]);
+					}
+				}
+			} else {
+				for (auto &c : projection_ids) {
+					auto column_id = column_ids[c].GetPrimaryIndex();
+					if (IsVirtualColumn(column_id)) {
+						auto &column = virtual_columns.at(column_id);
+						filter_types.push_back(column.type);
+					} else {
+						filter_types.push_back(op.returned_types[column_id]);
+					}
 				}
 			}
 			filter = Make<PhysicalFilter>(filter_types, std::move(select_list), op.estimated_cardinality);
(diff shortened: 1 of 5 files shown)


keithjjones:
I didn't have the latest code.  Sorry about that.  Closing.

PR #23371, opened 2026-06-19
fix out-of-bounds read in json path ReadKey lookahead

jmestwa-coder:
Two lookahead reads in the JSON path parser (ReadKey) run one byte past the path while only one byte is guaranteed left:
- the `**` wildcard check reads `*(ptr + 1)`, hitting `*end` when `*` ends the path
- the recursive `.` branch reads the same byte, then walks ptr past end

Reachable from a non-constant path arg whose string_t is non-inlined (> 12 bytes, so not null-terminated), e.g. `json_extract(j, '$.aaaaaaaaaaaa.*')`. ASan flags a heap-buffer-overflow read of size 1 in ReadKey. Constant paths come from a null-terminated copy and are unaffected. Bounded both lookaheads against end and reject a trailing recursive `.`.

### DIFF
--- a/extension/json/json_common.cpp
+++ b/extension/json/json_common.cpp
@@ -134,14 +134,14 @@ static inline idx_t ReadInteger(const char *ptr, const char *const end, idx_t &i
 static inline JSONKeyReadResult ReadKey(const char *ptr, const char *const end) {
 	D_ASSERT(ptr != end);
 	if (*ptr == '*') { // Wildcard
-		if (*(ptr + 1) == '*') {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks! Can you rebase/retarget this to `v1.5-variegata`?

jmestwa-coder:
rebased onto v1.5-variegata and retargeted the PR. the fix and both test cases carried over cleanly.

Mytherin:
Thanks!

PR #23372, opened 2026-06-19
Fix client ctx ref cycle

evertlammerts:
See context in #23344 

The cyclic ref works as follows. A call the `ClientContext::PendingQuery` does the following:

* On the main thread, through `PendingStatementOrPreparedStatement`:
   * set `ClientContext::active_query` (a `unique_ptr<ActiveQueryContext>`) in `BeginQueryInternal`
   * then through `PendingStatementInternal` -> `CreatePreparedStatement` -> `PendingPreparedStatementInternal`:
      * set `active_query->executor = make_uniq<Executor>(*this)`
      * build pipelines with `executor.Initialize(std::move(collector));`
      * then through `Executor::Initialize` -> `InitializeInternal` -> `ScheduleEvents`:
         * create a `make_shared_ptr<PipelineInitializeEvent>` and schedule it
* Then when a worker runs the scheduled `PipelineInitializeEvent`:
   * through `PipelineInitializeTask::ExecuteTask` -> `ResetSink`:
      * gets the sink state through `PhysicalMaterializedCollector::GetGlobalSinkState`, which:
         * creates a `unique_ptr<MaterializedCollectorGlobalState> state`
         * sets a `shared_ptr<ClientContext>` on the state: `state->context = context.shared_from_this();`
         * returns the state
      * then saves the state on `Pipeline`'s `optional_ptr<PhysicalOperator> sink;`

This creates the `shared_ptr` loop, which is essentially:
```
ClientContext → active_query → executor → pipeline ─┐
^───────────────────────────────────────────────────┘
```

### DIFF
--- a/src/execution/operator/helper/physical_materialized_collector.cpp
+++ b/src/execution/operator/helper/physical_materialized_collector.cpp
@@ -15,7 +15,8 @@ class MaterializedCollectorGlobalState : public GlobalSinkState {
 public:
 	mutex glock;
 	unique_ptr<ColumnDataCollection> collection;
-	shared_ptr<ClientContext> context;
+	//! This is weak to avoid creating a cyclical reference
+	weak_ptr<ClientContext> context;
 };
 
 class MaterializedCollectorLocalState : public LocalSinkState {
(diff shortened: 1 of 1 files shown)


src/execution/operator/helper/physical_materialized_collector.cpp
```diff
@@ -64,12 +65,12 @@ unique_ptr<LocalSinkState> PhysicalMaterializedCollector::GetLocalSinkState(Exec
 
 unique_ptr<QueryResult> PhysicalMaterializedCollector::GetResult(GlobalSinkState &state) const {
 	auto &gstate = state.Cast<MaterializedCollectorGlobalState>();
+	auto cc = gstate.context.lock();
```
Mytherin, line 68: Can we check if this returns nullptr and then throw an exception?

evertlammerts: We can, but I don't think we need to? The `GlobalSinkState` can't outlive the context, it's owned by it through the `pipeline` -> `executor` -> `ClientContext::active_query` and `PhysicalMaterializedCollector::GetResult` is not (and should not ever) be called through ClientContext's destructor.

I could add as `D_ASSERT`? But `PhysicalBufferedBatchCollector::GetResult` and `PhysicalBufferedCollector::GetResult` don't assert or check either.

Mytherin: Alright - makes sense

PR #23378, opened 2026-06-19
Clean up examples folder

d-justen:
- removes python example
- aligns C and C++ example so that they do the same thing

fixes https://github.com/duckdblabs/duckdb-internal/issues/7306


Mytherin:
Thanks!

PR #23379, opened 2026-06-19
check for interrupt state in ExecuteTaskInternal

evertlammerts:
`PendingQueryResult::ExecuteTask` would only see that it was interrupted if there was a task available for execution. Which wouldn't happen if e.g. a streaming result has a full buffer. ExecuteTask would just keep returning RESULT_READY as long as nothing consumes the result.

the fix here is to check for the interrupt state top level, in ExecuteTask directly (same as in `SimpleBufferedData::ExecuteTaskInternal`)

### DIFF
--- a/src/main/client_context.cpp
+++ b/src/main/client_context.cpp
@@ -738,7 +738,8 @@ PendingExecutionResult ClientContext::ExecuteTaskInternal(ClientContextLock &loc
 	bool invalidate_transaction = true;
 	try {
 		// Surface a pending interrupt even when this thread runs no task that reaches InterruptCheck.
-		if (!dry_run && interrupt_state.load(std::memory_order_relaxed) == ClientInterruptState::INTERRUPTED) {
+		// IsInterrupted() rather than InterruptCheck(): we must not enforce query_deadline here.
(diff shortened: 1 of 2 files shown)


src/main/client_context.cpp
```diff
@@ -737,6 +737,10 @@ PendingExecutionResult ClientContext::ExecuteTaskInternal(ClientContextLock &loc
 	D_ASSERT(active_query->IsOpenResult(result));
 	bool invalidate_transaction = true;
 	try {
+		// Surface a pending interrupt even when this thread runs no task that reaches InterruptCheck.
+		if (!dry_run && interrupt_state.load(std::memory_order_relaxed) == ClientInterruptState::INTERRUPTED) {
+			throw InterruptException();
```
lnkuiper, line 743: Shouldn't this be calling `ClientContext::InterruptCheck`?

evertlammerts: We could, but that function also checks a `query_deadline` check, which I don't think we necessarily want, plus it has the side effect of increasing the `timeout_check_counter`.

I missed `IsInterrupted()` before, that's now updated.

Mytherin:
Thanks!

PR #23381, opened 2026-06-19
[SqllogicTester] Add `--stdin` option to read statements from stdin

Tishj:
In the new `--stdin` mode, the unittester reads statements from stdin, instead of from files on disk, and it performs eager execution for every statement it encounters, rather than gather all statements first.

As can be seen in `tools/sqllogic/tests/test_unittest_stdin.py` this allows us to use the tester with more granularity, opening the possibility to perform other work (such as modifying shared tables in an Iceberg REST Catalog) between statements.

### DIFF
--- a/.github/workflows/OSX.yml
+++ b/.github/workflows/OSX.yml
@@ -200,6 +200,12 @@ jobs:
         run: |
           python -m pytest tools/shell/tests --shell-binary build/release/duckdb
 
+      - name: Unittest Stdin Tests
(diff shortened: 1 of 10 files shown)


benfleis:
I _think_ this now allows reading any arbitrary file-like thing (think /dev/fd, unix sockets), right? That opens a nice door by making the whole thing stream friendly!

Mytherin:
Thanks!

PR #23384, opened 2026-06-20
fix out-of-bounds read in string to struct cast

jmestwa-coder:
**Out-of-bounds read in SplitStruct**
After consuming the opening `{`/`(` and skipping whitespace, the empty-struct check reads `buf[pos]` without testing `pos` against `len`, so a non-inlined string that is just the opening character followed by spaces reads one byte past the buffer. The list and map splitters already guard this spot (`pos < len` / `pos == len`); matched the struct path to them.

ASan on `CAST('{               ' AS STRUCT(a INTEGER))` before the fix:
```
ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 1
    #0 ... duckdb::VectorStringToStruct::SplitStruct(...) vector_cast_helpers.cpp:462
```

### DIFF
--- a/src/function/cast/vector_cast_helpers.cpp
+++ b/src/function/cast/vector_cast_helpers.cpp
@@ -460,6 +460,9 @@ bool VectorStringToStruct::SplitStruct(const string_t &input, vector<unique_ptr<
 	auto end_char = buf[pos] == '{' ? '}' : ')';
 	pos++;
 	SkipWhitespace(input_state);
+	if (pos == len) {
+		return false;
+	}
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks! Can you rebase/retarget this to `v1.5-variegata`?

jmestwa-coder:
done, rebased onto `v1.5-variegata` and retargeted the base. diff is just the SplitStruct guard plus the two regression cases.

Mytherin:
Thanks!

PR #23390, opened 2026-06-20
fix: allow DETACH inside a transaction without invalidating it

### LINKED ISSUE #23374: DETACH inside a transaction throws an error and rolls back the transaction on commit in >=1.5.3.0
### What happens?

After upgrading DuckDB from `1.5.2` to `>=1.5.3`, detaching a database (in this case Postgres) inside an active transaction fails with a `TransactionContext Error`.

Additionally, after the `DETACH` throws an error, the subsequent `COMMIT` seems to roll back the transaction context. When we try to read from the database again, we see that the table created and data inserted during the transaction were lost. 

This check becomes problematic when using application frameworks (like Spring Boot) that automatically manage the transaction lifecycles.

### To Reproduce

```sql
START TRANSACTION;
ATTACH 'host=localhost port=5432 dbname=postgres user=postgres password=postgres' AS postgres_db (TYPE postgres);
CREATE TABLE postgres_db.test
         (
             ts timestamp
         );


INSERT INTO postgres_db.test(ts) VALUES (now());
SELECT * FROM postgres_db.test;
DETACH postgres_db;
COMMIT;
```

This example works in `v1.5.2`, but does not in `>=v1.5.3'.

### OS:

macOS Version 26.5.1 

### DuckDB Version:

>= 1.5.3.0

### DuckDB Client:

CLI

### Hardware:

Apple M5 (ARM64)

### Full Name:

Wout Bosteels

### Affiliation:

DataLabs

- [x] Yes, I have

- [x] Yes, I have

sandy-sachin7:
Fixes #23374

Previously DETACH threw a TransactionException when the current transaction had outstanding work on the database (e.g. after ATTACH + INSERT). This error invalidated the entire transaction, causing a subsequent COMMIT to silently convert to ROLLBACK.

The outstanding-work check was added in commit 7146d836a1 which also removed the MetaTransaction::DetachDatabase() cleanup call from DatabaseManager::DetachDatabase.

**Changes:**
1. Removes the outstanding-work check from PhysicalDetach
2. Restores MetaTransaction::DetachDatabase() in DatabaseManager::DetachDatabase to clean up used_databases
3. Updates tests to expect DETACH to succeed within transactions
4. Adds a new test case for ATTACH -> work -> DETACH -> COMMIT

### DIFF
--- a/src/execution/operator/schema/physical_detach.cpp
+++ b/src/execution/operator/schema/physical_detach.cpp
@@ -1,12 +1,6 @@
 #include "duckdb/execution/operator/schema/physical_detach.hpp"
 #include "duckdb/parser/parsed_data/detach_info.hpp"
-#include "duckdb/catalog/catalog.hpp"
-#include "duckdb/common/exception/transaction_exception.hpp"
 #include "duckdb/main/database_manager.hpp"
-#include "duckdb/main/attached_database.hpp"
-#include "duckdb/main/database.hpp"
-#include "duckdb/storage/storage_extension.hpp"
-#include "duckdb/transaction/meta_transaction.hpp"
 
 namespace duckdb {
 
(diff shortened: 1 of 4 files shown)


Mytherin:
Thanks - but I don't think this is the right direction for this fix.

Maybe what we want to do is instead not allow *re-attaching* of a database we detached with changes in the same transaction, instead of preventing the detach. See also https://github.com/duckdb/duckdb/pull/22435 for more context on why this fix is not the right one.

PR #23392, opened 2026-06-21
Fix metadata reader offset advancement

wufengwind:
In MetadataReader::ReadData, the offset was being advanced by read_size instead of to_read.

After read_size -= to_read, read_size is the remaining bytes to read, not the bytes just copied. offset += read_size could skip past the actual data position.

Fixed by advancing offset by to_read instead.


Mytherin:
Thanks!

PR #23393, opened 2026-06-21
Add support for compressed materialization for geometry points

Maxxen:
If we know we only have POINT XY geometries, we can just stash their X/Y unto a UHUGEINT.

This can avoid the need for the variable-length part of the tuple data format. It gives a pretty good speedup for ORDER BY's (which admittedly is not something you do very often with GEOMETRY, the binary sort order is not very useful, unless you ORDER BY ALL), but Im optimistic that the group-by/payload use case could benefit spatial/distance joins in the future.

Some micro benchmarks (10_000_000 generated points)

  ```
  ┌──────────────────────────────────────────┬─────────────────┬────────────────┬─────────┐
  │                 Workload                 │ CM off (median) │ CM on (median) │ Speedup │
  ├──────────────────────────────────────────┼─────────────────┼────────────────┼─────────┤
  │ ORDER BY g — geometry as sort key        │ 0.880 s         │ 0.196 s        │ ~4.5×   │
  ├──────────────────────────────────────────┼─────────────────┼────────────────┼─────────┤
  │ ORDER BY id, geometry carried as payload │ 0.153 s         │ 0.120 s        │ ~1.3×   │
  └──────────────────────────────────────────┴─────────────────┴────────────────┴─────────┘
```


lnkuiper:
Looks excellent, thanks!

PR #23394, opened 2026-06-22
Infer geometry type based on WKB string stats

Maxxen:
When calling `ST_GeomFromWKB()` we can determine if a column contains only a single geometry type by looking at the 5 byte prefix of min/max string statistics. This is pretty common, and detecting it may allow us to perform further optimizations downstream.

### DIFF
--- a/src/function/scalar/geometry/geometry_functions.cpp
+++ b/src/function/scalar/geometry/geometry_functions.cpp
@@ -54,8 +54,28 @@ static auto FromWKBStats(ClientContext &context, FunctionStatisticsInput &input)
 		// Unsupported or invalid geometry type, we can't say anything about the types
 		return result.ToUnique();
 	}
+
+	// Z/M may also be signalled through the Extended-WKB high bits (matching Geometry::FromBinary)
+	const auto has_z = ((flag_id & 0x01) != 0) || ((meta & 0x80000000) != 0);
+	const auto has_m = ((flag_id & 0x02) != 0) || ((meta & 0x40000000) != 0);
+
 	const auto geom_type = static_cast<GeometryType>(type_id);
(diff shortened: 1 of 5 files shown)


lnkuiper:
Looks good to me, but Codex thinks it found two breakages:
```sql
# name: test/temp.test
# description: temporary reproducer for PR 23394 review
# group: [temp]

statement ok
CREATE TABLE bad_wkb(wkb BLOB);

statement ok
INSERT INTO bad_wkb VALUES ('\x01\x02\x00\x00\x00'::BLOB);

statement error
SELECT vertex_extract(ST_GeomFromWKB(wkb), 'x') FROM bad_wkb;
----
<REGEX>:Invalid Input Error: Unexpected end of binary data at position 5

statement ok
CREATE TABLE ewkb_point_z(wkb BLOB);

statement ok
INSERT INTO ewkb_point_z VALUES
  ('\x01\x01\x00\x00\x80\x00\x00\x00\x00\x00\x00\xF0\x3F\x00\x00\x00\x00\x00\x00\x00\x40\x00\x00\x00\x00\x00\x00\x08\x40'::BLOB);

query I
SELECT ST_AsText(ST_GeomFromWKB(wkb)) FROM ewkb_point_z;
----
POINT Z (1 2 3)

query I
SELECT vertex_extract(ST_GeomFromWKB(wkb), 'z') FROM ewkb_point_z;
----
3.0

```

Can you have a look and see if they're valid?

Maxxen:
Thanks!

I've fixed the EWKB case, and hardened the inference a bit by verifying the geometry type _can be valid_ given the min length of the string. However, we can't detect if the actual geometry bodies are semantically valid too without parsing each row. But I think we are generally okay with optimizing assuming valid input, so I think this is fine.

PR #23396, opened 2026-06-22
Bound peak scan memory with a sub-vector byte budget

baolinhuang:
for discuss  https://github.com/duckdb/duckdb/discussions/23275/

---
  Problem:
  ========
  Reads of a row group materialize a full STANDARD_VECTOR_SIZE (2048) row vector at
  a time. For wide columns that is unbounded in bytes: 2048 rows of 300KB blobs is
  ~600MB of pinned overflow strings in a single vector, which cannot spill and so
  aborts the query, the commit, or the checkpoint under any modest memory_limit.
  
  Solution:
  ========
  Add scan_target_size_bytes, a byte budget that splits each vector into
  sub-batches. Budget 0 (the default) keeps the original fixed 2048-row stepping
  byte-for-byte.
  
  ScanSizePredictor sizes each batch as target_bytes / worst_case_row_width, with
  the width summed over the scanned columns from the row group's own statistics:
  the exact width for fixed-size types, the string_t slot plus MaxStringLength for
  VARCHAR, and a 256-byte fallback for LIST/MAP and for columns without statistics.

Effectiveness verification
===================
## 1. Reproduction

**Resource configuration**

| Item | Value |
|---|---|
| Build | `make reldebug` |
| `A` = current `main` | `fd94c824f2` (setting does not exist) |
| `B`/`C` = this patch | `main` + 1 commit |
| `threads` | `1` |
| `memory_limit` | `15MB` |
| `max_temp_directory_size` | `0B` (buffer manager must fail, not spill) |

**Table initialization** — 2048 rows × 50 KB string ≈ 100 MB per full vector

```sql
CREATE TABLE t AS
SELECT i, repeat(chr((65 + (i % 26))::INTEGER), 50000) AS b FROM range(0, 2048) tbl(i);
```

**Query under test** — identical in all three runs

```sql
PRAGMA threads=1;
SET memory_limit='15MB';
SET max_temp_directory_size='0B';
-- <the setting line for B / C goes here>
SELECT SUM(LENGTH(b)) FROM t;
```

**Three-way comparison (measured output, verbatim)**

| | Setting line | Rows / scan step | Bytes / scan step | Result |
|---|---|---|---|---|
| **A** current `main` | — | 2048 | ~100 MB | `Out of Memory Error: failed to allocate data of size 128.0 MiB (320.0 KiB/14.3 MiB used)` |
| **B** patch, off | `SET scan_target_size_bytes=0;` | 2048 | ~100 MB | `Out of Memory Error: failed to allocate data of size 128.0 MiB (320.0 KiB/14.3 MiB used)` — **byte-identical to A** |
| **C** patch, 1 MB | `SET scan_target_size_bytes=1048576;` | **20** | **~1 MB** | `102400000` |

A ≡ B confirms the patch is a no-op when unset (default `0`). C's step size, TRACE log verbatim:

```
ScanSizePredictor batch: target_bytes=1048576 max_rows=2048 worst_bpr=50016.0 safe=20 scan_count=20
```

## 2. Scenario matrix

| # | Scenario | Scope | Data / `memory_limit` | `=0` (≡ current `main`) | `=1M` / `=16M` |
|---|---|---|---|---|---|
| 1 | Table scan, no filter | `SET` | 2048 × 50 KB / 15 MB | `Out of Memory` | `102400000` |
| 2 | Table scan, 4 threads | `SET` | 4096 × 50 KB / 30 MB | `Out of Memory` | `204800000` |
| 3 | Scan + pushed-down filter | `SET` | 2048 × 300 KB / 30 MB | `Out of Memory` | `614100000` |
| 4 | Scan over deletes | `SET` | 2048 × 50 KB, 21 deleted / 15 MB | `Out of Memory` | `101350000` |
| 5 | Scan over pending updates | `SET` | 2048 × 50 KB, 10 updated / 15 MB | `Out of Memory` | `102400000` |
| 6 | `CHECKPOINT` | **`GLOBAL`** | 2500 × 300 KB / 300 MB | `Failed to create checkpoint … failed to pin block` | `CHECKPOINT` ok, `750000000` |
| 7 | Commit-time WAL write | **`GLOBAL`** | 2048 × 300 KB in 1 txn / 300 MB | `Failed to commit … failed to pin block` | `COMMIT` ok, `614400000` |
| 8 | `ALTER … SET NOT NULL` | **`GLOBAL`** | 2048 × 300 KB / 300 MB | `Out of Memory`, constraint not added | ALTER ok, constraint enforced |
| 9 | `ALTER … SET DATA TYPE` | **`GLOBAL`** | 2048 × 300 KB / 300 MB | `Out of Memory`, stays `VARCHAR` | becomes `BLOB`, `614400000` |
| 10 | Vacuum merge-read | **`GLOBAL`** | 16384 × 300 KB, 2 sparse groups / 300 MB | `Checkpoint failed … failed to pin block`, 2 groups | ok, merged to **1** group |

`SET` = session-local works. **`GLOBAL`** = the path runs without a `ClientContext`, so only `SET GLOBAL` is visible to it.

## 3. Code path per scenario

| # | Bounded call site | Covering test |
|---|---|---|
| 1–5 | `RowGroup::Scan` (plain / filter / delete / pending-update variants) | `storage/scan_target_size_bytes_memory.test_slow` |
| 6 | `ColumnDataCheckpointer::ScanSegments` → `ScanSizePredictor::PredictBatchSizeSingle` | `storage/checkpoint_blob_memory.test_slow` |
| 7 | `DataTable::WriteToLog` → `DataTable::ScanTableSegment` | `storage/scan_table_segment_wal_memory.test_slow` |
| 8 | `RowGroupCollection::VerifyNewConstraint` | `storage/alter_scan_target_size_bytes_memory.test_slow` TC1 |
| 9 | `RowGroupCollection::AlterType` | same file TC2 |
| 10 | `VacuumTask` row-group merge scan | `storage/vacuum/vacuum_scan_target_size_bytes.test_slow` |

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -1229,6 +1229,13 @@
         "default_value": "true"
     },
     {
+        "name": "scan_target_size_bytes",
+        "description": "Target maximum size in bytes for each scan result chunk. When set, the scan dynamically reduces the number of rows per chunk to stay under this limit. 0 = disabled (scan full vectors of 2048 rows).",
+        "type": "UBIGINT",
+        "scope": "local",
+        "custom_implementation": true
+    },
+    {
         "name": "scheduler_process_partial",
         "description": "Partially process tasks before rescheduling - allows for more scheduler fairness between separate queries",
         "type": "BOOLEAN",
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -1268,6 +1268,13 @@
         "default_value": "true"
     },
     {
+        "name": "scan_target_size_bytes",
+        "description": "Target maximum size in bytes for each large-column read sub-batch, applied to query scans as well as checkpoint, commit (WAL write), and rollback reads. When set, these reads use byte-budgeted sub-batches instead of materializing a full 2048-row vector at once, bounding peak memory. GLOBAL_LOCAL: the global value covers background/shutdown/commit paths that run without a session; a local value overrides it for the current session's query scans. 0 = disabled (read full vectors).",
+        "type": "UBIGINT",
+        "scope": "global_local",
+        "custom_implementation": true
+    },
+    {
         "name": "scheduler_process_partial",
         "description": "Partially process tasks before rescheduling - allows for more scheduler fairness between separate queries",
         "type": "BOOLEAN",
--- a/src/function/table/table_scan.cpp
+++ b/src/function/table/table_scan.cpp
@@ -4,6 +4,7 @@
 #include "duckdb/catalog/dependency_list.hpp"
 #include "duckdb/common/enums/expression_type.hpp"
 #include "duckdb/common/mutex.hpp"
+#include "duckdb/logging/logger.hpp"
 #include "duckdb/common/serializer/deserializer.hpp"
 #include "duckdb/common/serializer/serializer.hpp"
 #include "duckdb/common/typedefs.hpp"
@@ -316,12 +317,15 @@ public:
 		}
 
 		l_state->scan_state.options.force_fetch_row = Settings::Get<DebugForceFetchRowSetting>(context.client);
+		l_state->scan_state.options.scan_target_size_bytes =
+		    ClientConfig::GetConfig(context.client).scan_target_size_bytes;
 		return std::move(l_state);
 	}
 
 	void TableScanFunc(ClientContext &context, TableFunctionInput &data_p, DataChunk &output) override {
 		auto &l_state = data_p.local_state->Cast<TableScanLocalState>();
 		l_state.scan_state.options.force_fetch_row = Settings::Get<DebugForceFetchRowSetting>(context);
+		l_state.scan_state.options.scan_target_size_bytes = ClientConfig::GetConfig(context).scan_target_size_bytes;
 
 		do {
 			if (bind_data.is_create_index) {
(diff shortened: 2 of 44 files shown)


Mytherin:
Thanks for the PR!

I think this is a real problem that needs tackling but I don't really like the idea of having a separate global config option for it. I think having a separate code path for doing scans also seems error prone to me to maintain long term.

I would say that ideally we should have a "target vector size in bytes" that we set to a sensible default (e.g. 100MB, or so), and the scan then dynamically emits smaller vectors when large values are encountered that would cause this limit to be exceeded. Maybe we can look at the maximum string length / total string length in the stats to guide this, e.g. if the max string length is 1MB we would emit vectors of 100 rows.

baolinhuang:
Thank you for your reply. I agree with the above points:

1. Adding global configuration is indeed not a good solution. As mentioned in the TODO, there is a plan to adaptively adjust the vector size using bytes as the unit of measurement. I will expedite the implementation of this module.

2. The addition of RowGroup::ScanSubBatch is intended to reduce conflicts with upstream components. I will change it to a direct approach within RowGroup::Scan.
---
I will submit a new patch as soon as possible.

baolinhuang:
Submitted a patch containing two modifications:
1. The change is located within `RowGroup::Scan`.
2. introduced setting `scan_target_size_bytes`, and a `ScanSizePredictor` to estimate target rows.

baolinhuang:
Hi @Mytherin 

The remaining failures (Main Extensions) doesn't seem related to my changes (similar issues have been observed in other PRs).

I’ve finished the changes; please review them. I’d appreciate it if you could verify its feasibility.

baolinhuang:
Some background
I am from the Alibaba Cloud MySQL team (https://github.com/alibaba/AliSQL). We have integrated DuckDB into MySQL to provide online database services, so DuckDB's unstable memory consumption poses a significant issue.

Maxxen:
I think this would be a great change to get in eventually - not stale

baolinhuang:
This patch has already been deployed to our production environment, and its effectiveness has been verified.
Looking forward to the review.

PR #23399, opened 2026-06-22
Rework parquet variant reading to use `ParquetVariantIterator` and remove `VariantValue` class

Mytherin:
This PR reworks the parquet variant reading code to use an iterator model similar to https://github.com/duckdb/duckdb/pull/23359. The same "unshredding" code is shared between the Parquet variant reading code and the unshredding code - where we create an iterator and use the `VariantBuilder` to construct the new (unshredded) variant.

This improves performance by around ~20% and allows us to get rid of a bunch of code (removing `VariantValue` and `VariantValue::ToVARIANT`). Below are timings for reading JSONBench (for which currently this path is the primary bottleneck due to us not yet doing pushdown into variants in Parquet).

| Version | Timing (s) |
|---------|------------|
| Old     | 1.05s      |
| New     | 0.8s       |


Tishj:
Please keep in mind that v1.5 has https://github.com/duckdb/duckdb/pull/23057 unmerged with main still, not sure if that can continue working if you remove `VariantBinaryDecoder` in its entirety

Tishj:
As well as https://github.com/duckdb/duckdb/pull/23100 that reworks the reading to use an offset and a size, instead of a raw pointer, to add hardening to the deserialization

Tishj:
And https://github.com/duckdb/duckdb/pull/23195 is also lost / will likely result in hidden or hard to properly resolve merge conflicts

Mytherin:
Hm yeah, fair enough, let me merge v1.5 into main first then

Mytherin:
This also includes the `v1.5-variegata -> main` merge

PR #23404, opened 2026-06-22
Add managed async task queue 

guillesd:
The existing infrastructure around `AsyncWriteQueue` / `ManagedAsyncWriteQueue` was designed with the parquet writer in mind (writing bytes to a file at an offset). This PR pulls the reusable components out of it and reuses it in a generic, memory-managed task queue for arbitrary `AsyncTask`s. This is particularly interesting for quack: regular execution threads can serialize a batch and hand it off as a task, and dedicated async-pool threads do the blocking work (an HTTP POST, say) while the regular threads keep crunching.

## Parts

- `AsyncTaskQueue`: the bare scheduling/draining layer. One unit per drain task, so up to K tasks run concurrently and independent tasks actually overlap (no coalescing, that only makes sense for contiguous writes). Multi-producer, with the same error capture/rethrow behavior as the write path.
- `ManagedAsyncTaskQueue`: the public wrapper. Register(task, byte_size), ApplyBackpressure(), WaitAll(), Close(), and memory-bounded draining via a TemporaryMemoryManager reservation.
- `ManagedAsyncMemoryGovernor`: the one shared bit. The coarse-growth reservation heuristic plus the backpressure budget, now lives in one place and is used by both the write queue and the task queue.

## Alternative ideas

I did not follow the template approach on purpose because I thought it would add complexity. However, there is an argument to be made that, even if this approach is more involved, it can create a single reference for a `ManagedAsyncQueueBase`. This is initially what I had in mind but turned down due to complexity. Happy to try it out if you think is best.

cc: @lnkuiper

### DIFF
--- a/src/common/serializer/CMakeLists.txt
+++ b/src/common/serializer/CMakeLists.txt
@@ -2,6 +2,7 @@ add_library_unity(
   duckdb_common_serializer
   OBJECT
   async_file_writer.cpp
+  async_task_queue.cpp
   async_write_queue.cpp
   binary_serializer.cpp
   binary_deserializer.cpp
--- a/src/common/serializer/CMakeLists.txt
+++ b/src/common/serializer/CMakeLists.txt
@@ -2,6 +2,7 @@ add_library_unity(
   duckdb_common_serializer
   OBJECT
   async_file_writer.cpp
+  async_memory_governor.cpp
   async_task_queue.cpp
   async_write_queue.cpp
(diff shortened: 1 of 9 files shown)


lnkuiper:
Thanks! Looks great now, let's run CI

lnkuiper:
Looks like the tidy check failed - could you have a look?

Mytherin:
Thanks!

PR #23407, opened 2026-06-22
Add missing statement types to capi

hrl20:
Adds the remaining statement types to the C API for completion. right now if one tries to get the statement type of an `MERGE INTO` statement through e.g. duckdb-go binding which uses the C API, they would see Invalid statemen type.


Mytherin:
Thanks!

PR #23410, opened 2026-06-23
Clean async tasks after schedule

dentiny:
Implementation doesn't confirm its documentation https://github.com/duckdb/duckdb/blob/ababb622f71f2ed49ff891fe3332bcd2679980a7/src/include/duckdb/parallel/async_result.hpp#L46-L49

Current implementation leaves vector there but all tasks inside are default.

I have ran CI on my fork: https://github.com/dentiny/duckdb/pull/112, failure unrelated


lnkuiper:
Thanks!

PR #9375, opened 2023-10-17
[Export Database] Produce up-to-date query for ViewCatalogEntry

Tishj:
This PR fixes #9367 

We regenerate the `sql` when ToSQL() is called, this is different from before in that *all* names are provided as explicit aliases.
This is done because name changes of the referenced table is not reflected in the view anyways, so it's fine to just record the names at the time of creation.

This behavior (not letting changes to the referenced query impact the names of the view) is copied from PostgreSQL

PR #23411, opened 2026-06-23
Internal #9375: TIME_BUCKET For TIME

hawkfish:
* Generalise the time_bucket code to support TIME.
* Generate test suite from timestamp using 1970-01-01
* Don't mark TIME versions as monotonic.


lnkuiper:
Looks good! This needs a change to `src/include/duckdb/main/extension_entries.hpp`. Once that's fixed feel free to mark `Ready to Merge`.

PR #23412, opened 2026-06-23
fix(optimizer): deduplicate identical window expressions in window_se…

### LINKED ISSUE #23383: INTERNAL Error: "Failed to bind column reference" — window-over-aggregate used both standalone and inside arithmetic in a GROUP BY projection
### What happens?

In the projection of a `GROUP BY` query, using the **same window-over-aggregate expression** (`MAX(SUM(n)) OVER (PARTITION BY name)`) **both standalone and as a sub-expression of arithmetic** crashes the binder with an internal assertion instead of executing:

```
INTERNAL Error: Failed to bind column reference "" [19.1] (bindings: {#[22.0], #[22.1], #[20.0], #[19.0]})
... in duckdb::ColumnBindingResolver::VisitReplace ...
```

The same query with the window used only once, or used twice but never inside arithmetic, runs correctly — so it is specifically the *duplicated* window expression (one bare, one nested inside arithmetic) that fails to rebind.

### To Reproduce

Failing query (CLI, no extensions, inline data):

```sql
SELECT
  name,
  MAX(SUM(n)) OVER (PARTITION BY name)          AS w1,
  SUM(n) / MAX(SUM(n)) OVER (PARTITION BY name) AS w2
FROM (VALUES ('a',1),('a',2),('b',3)) t(name, n)
GROUP BY name;
```

Output:

```
INTERNAL Error: Failed to bind column reference "" [19.1] (bindings: {#[22.0], #[22.1], #[20.0], #[19.0]})

For more information, see https://duckdb.org/docs/stable/dev/internal_errors
```

(Stack trace top frames: `ColumnBindingResolver::VisitReplace` -> `LogicalOperatorVisitor::VisitExpression` -> ... -> `PhysicalPlanGenerator::ResolveAndPlan`.)

Negative controls — both succeed:

```sql
-- (1) window used once, inside arithmetic only:
SELECT name, SUM(n) / MAX(SUM(n)) OVER (PARTITION BY name) AS w2
FROM (VALUES ('a',1),('a',2),('b',3)) t(name,n)
GROUP BY name;

-- (2) window used twice but neither occurrence inside arithmetic:
SELECT name,
  MAX(SUM(n)) OVER (PARTITION BY name) AS w1,
  MAX(SUM(n)) OVER (PARTITION BY name) AS w2
FROM (VALUES ('a',1),('a',2),('b',3)) t(name,n)
GROUP BY name;
```

The bug reproduces across versions: CLI `v1.5.2 (Variegata) 8a5851971f`, `@duckdb/node-bindings 1.5.3-r.2`, and `@duckdb/node-api 1.5.4-r.1` (latest stable at time of filing). It is therefore not a recent regression.

(Found while reducing SQL generated by Malloy for an "ungrouped aggregate" / level-of-detail query, but the reproduction above is plain SQL with no Malloy or external dependency.)

### OS:

Debian GNU/Linux 12 (bookworm), aarch64 (Linux 6.6.99)

### DuckDB Version:

1.5.2 (also reproduced on 1.5.3 and 1.5.4)

### DuckDB Client:

CLI (also reproduced via Node.js @duckdb/node-api)

### Hardware:

_No response_

### Full Name:

Lloyd Tabb

### Affiliation:

Malloy

- [X] Yes, I have

- [X] Yes, I have

Youssef-joe:
…lf_join

Fixes #23383 — an INTERNAL Error: Failed to bind column reference crash triggered when the same window-over-aggregate expression appears more than once in a query's SELECT list (e.g. standalone and inside arithmetic).

Root cause:

The binder pushes one entry into node.windows per occurrence of a window expression, without deduplication. A query like:

SELECT
  MAX(SUM(n)) OVER (PARTITION BY name) AS w1,
  SUM(n) / MAX(SUM(n)) OVER (PARTITION BY name) AS w2
FROM ... GROUP BY name;
produces a LogicalWindow with two identical expressions at slots (window_index, 0) and (window_index, 1).

The window_self_join optimizer previously created one aggregate expression per window expression (including duplicates), yielding two identical aggregate outputs (aggregate_index, 0) and (aggregate_index, 1), and registered a replacement binding for each. A subsequent optimizer (column_lifetime) correctly identified (aggregate_index, 1) as a redundant duplicate and pruned it via the join's right_projection_map. The expressions that had been rewritten to reference (aggregate_index, 1) then had no matching column at ColumnBindingResolver time, producing the internal error.

Fix:

In OptimizeInternal (src/optimizer/window_self_join.cpp), before building the new LogicalAggregate's expression list, deduplicate window expressions using BoundWindowExpression::Equals. Duplicate expressions reuse the earlier aggregate's output index rather than creating a new one. The replacement bindings are updated accordingly, so all references to the same logical window result are collapsed to a single aggregate output column — nothing spurious is created, nothing gets incorrectly pruned.

Test:

Added a regression test to test/sql/optimizer/test_window_self_join.test covering the exact pattern from the issue report.

### DIFF
--- a/src/optimizer/window_self_join.cpp
+++ b/src/optimizer/window_self_join.cpp
@@ -193,9 +193,22 @@ unique_ptr<LogicalOperator> WindowSelfJoinOptimizer::OptimizeInternal(unique_ptr
 			groups.push_back(std::move(part_copy));
 		}
 
-		for (auto &expr : window.expressions) {
-			auto &w_expr = expr->Cast<BoundWindowExpression>();
-			aggregates.emplace_back(TranslateAggregate(w_expr));
+		// Build deduplicated aggregates: multiple identical window expressions map to the same aggregate output.
+		vector<idx_t> agg_index_for_window(window.expressions.size());
+		for (idx_t i = 0; i < window.expressions.size(); ++i) {
+			auto &w_expr = window.expressions[i]->Cast<BoundWindowExpression>();
+			bool found = false;
+			for (idx_t j = 0; j < i; ++j) {
+				if (w_expr.Equals(*window.expressions[j])) {
+					agg_index_for_window[i] = agg_index_for_window[j];
+					found = true;
+					break;
+				}
+			}
+			if (!found) {
+				agg_index_for_window[i] = aggregates.size();
+				aggregates.emplace_back(TranslateAggregate(w_expr));
+			}
 		}
 
 		// args: group_index, aggregate_index, ...
@@ -224,12 +237,11 @@ unique_ptr<LogicalOperator> WindowSelfJoinOptimizer::OptimizeInternal(unique_ptr
 		join->children.push_back(std::move(agg_op));
 		join->ResolveOperatorTypes();
 
-		// Replace aggregate bindings
-		// Old window column: (window.window_index, x)
-		// New constant column: (aggregate_index, x)
+		// Replace window column bindings with the deduplicated aggregate output bindings.
+		// Old: (window.window_index, i)  →  New: (aggregate_index, agg_index_for_window[i])
 		for (idx_t column_index = 0; column_index < window.expressions.size(); ++column_index) {
 			ColumnBinding old_binding(window.window_index, ProjectionIndex(column_index));
-			ColumnBinding new_binding(aggregate_index, ProjectionIndex(column_index));
+			ColumnBinding new_binding(aggregate_index, ProjectionIndex(agg_index_for_window[column_index]));
 			replacer.replacement_bindings.emplace_back(old_binding, new_binding);
 		}
 
--- a/test/sql/optimizer/test_window_self_join.test
+++ b/test/sql/optimizer/test_window_self_join.test
@@ -265,3 +265,16 @@ FROM (
 ) AS b;
 ----
 1	10.0
+
+# Regression test: duplicate window-over-aggregate expressions (issue #23383)
+# Using the same window expression standalone AND inside arithmetic must not crash.
+query III rowsort
+SELECT
+    name,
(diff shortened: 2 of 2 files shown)


taniabogatsch:
Hi @Youssef-joe - thanks for the contribution! This should already be [fixed here](https://github.com/duckdb/duckdb/pull/23409), let me know if I missed something. :)

Youssef-joe:
> Hi @Youssef-joe - thanks for the contribution! This should already be [fixed here](https://github.com/duckdb/duckdb/pull/23409), let me know if I missed something. :)

Hi @taniabogatsch, thanks for the pointer!

I see that the linked PR disables the CommonAggregateOptimizer for MATERIALIZED_CTE, while my approach attempted to make the optimizer handle this case correctly instead of skipping it.

From your perspective, is the intended behavior to avoid this optimization inside MATERIALIZED_CTE altogether? If so, then the linked fix makes perfect sense. I just wanted to make sure I wasn't addressing a valid missing optimization case..

hawkfish:
Hi @Youssef-joe - thanks for taking the time to look at the problem. Unfortunately, this has nothing to do with window functions per se, but with aggregate deduplication.  You can probably generate the problem without window functions by just using a query that looks like the window self join optimisation.

Youssef-joe:
> Hi @Youssef-joe - thanks for taking the time to look at the problem. Unfortunately, this has nothing to do with window functions per se, but with aggregate deduplication.  You can probably generate the problem without window functions by just using a query that looks like the window self join optimisation.

Thanks for the explanation!

That makes sense - I focused too much on the window function aspect because that's how I first encountered the issue.

I'll try to come up with a minimal reproduction that triggers the problem through aggregate deduplication directly, without window functions, to better understand where the incorrect rewrite happens.

PR #23414, opened 2026-06-23
Transformer trampoline skeleton

Dtenwolde:

This PR introduces the skeleton for the heap-based trampoline-style transformer. The goal is to replace the current recursive transformer that uses the C++ call stack, because we were running into situations where we can run out of stack frames, for example with deeply nested expressions or `FROM duckdb_functions()` in `debug` mode.

The current transformer receives a `ParseResult` for each grammar rule, which can be thought of as a frame, and
recursively calls `transformer.Transform<T>(child)` for child rules. This becomes problematic for deeply nested
expressions. Expression precedence is encoded through multiple layered grammar rules (`AND` has higher precedence than `OR`), so even a single expression can pass through many intermediate transform frames before accounting for the expression's own nesting.

To solve this, we designed a trampoline-style transformer that stores transformation frames explicitly using a heap-allocated frame stack rather than recursive C++ call-stack frames. 

Each grammar rule's transform logic is split into two steps (and therefore two functions): `Initialize` and `Finalize`. Both of these phases can largely be generated thanks to work done in https://github.com/duckdb/duckdb/pull/23310 and the related PRs. 
 
- During `Initialize`, the initialization function will push child frames (`0..N`) onto the stack. Every frame has a `frame index`, `ParseResult`, `parent index` on the stack and the `slot index`, where the stack driver should store the child frame's finalized result in the parent. Pushing the child frames has to be done in LIFO order onto the stack to ensure we process rules in the correct order (e.g. anonymous parameters). After `Initialize` is finished, the status of the frame is set to `Waiting`. The next time the transformer reaches this frame, we will execute `Finalize`.
- During `Finalize` all of the child frames will have finished transforming, and their results have been stored in the parent slots. The current frame can now finish the transformation, and place its result into its parent. 

This cycle continues until the explicit frame stack is empty (i.e. no more frames) at which point transformation is complete.

For this PR I added the experimental setting `transformer_trampoline_style=<true|false>` to enable the trampoline-style transformer. In this skeleton PR, the trampoline path only supports the generated `use.gram` rules. As a result, `SET transformer_trampoline_style=false` is not yet supported once trampoline mode is enabled, because `SetStatement` has not been ported yet.

This PR only ports the `use.gram` rules to prove that the generated trampoline structure works.

**This PR does not port the full grammar; that is left for follow-up PRs.**


src/parser/peg/transformer/peg_transformer.cpp
```diff
@@ -1,12 +1,151 @@
...
+			frame.ops.initialize(transformer, *this, frame);
+			break;
+		case TransformFrameState::WAITING: {
+			auto result = frame.ops.finalize(transformer, *this, frame);
+			if (!result) {
+				throw InternalException("Trampoline transformer finalize for rule '%s' returned nullptr",
+				                        frame.ops.name);
+			}
+			frame_stack.pop_back();
+			if (frame.parent == DConstants::INVALID_INDEX) {
```
kryonix: Maybe use an `optional_idx` instead here. That would remove a few `DConstants::INVALID_INDEX` uses from the trampoline.

src/include/duckdb/parser/peg/transformer/peg_transformer.hpp
```diff
@@ -85,6 +85,90 @@ struct PEGTransformerState {
...
+		if (!typed_result) {
+			throw InternalException("Unexpected trampoline transformer result type for slot %llu in rule '%s'", slot,
+			                        ops.name);
+		}
+		auto result = std::move(typed_result->value);
+		child_results[slot].reset();
+		return result;
+	}
+
+	transform_frame_index_t frame_index;
```
kryonix: The `frame_index` can never change, right? Should it be const?

src/include/duckdb/parser/peg/transformer/peg_transformer.hpp
```diff
@@ -85,6 +85,90 @@ struct PEGTransformerState {
...
+	ParseResult &parse_result;
+	const TransformFrameOps &ops;
+	transform_frame_index_t parent;
```
kryonix: The `parent` should also maybe be const.

src/parser/peg/transformer/peg_transformer.cpp
```diff
@@ -1,12 +1,151 @@
...
+		result << " parent=#" << frame.parent << " slot=" << frame.parent_slot;
+	}
+	return result.str();
+}
+
+string TransformStack::FormatParentChain(transform_frame_index_t frame_index) const {
+	stringstream result;
+	auto current = frame_index;
+	bool first = true;
+	while (current != DConstants::INVALID_INDEX) {
```
kryonix: Can we accidentally introduce (unintended) loops in the frames vector? If so, this may loop forever. That'd be of course an error elsewhere.

scripts/parser/generate_transformer_trampoline.py
kryonix, line 1: For the CAPI v2 generator, we transitioned away from raw string construction, and used a templating engine instead. Maybe we want to do the same here eventually?

kryonix (commented):
Really excited for this 🙌🏻 Hope we can finally get rid of the stack depth limitations.

src/include/duckdb/parser/peg/transformer/peg_transformer.hpp
Dtenwolde: Indeed, good point :)

src/parser/peg/transformer/peg_transformer.cpp
Dtenwolde: This was indeed a bit ugly. I wrapped the `frame_idx` and `slot` (both `const`)  into `TransformFrameResultTarget` (The things related to the parent) and made that `optional<>`. It's only optional because the root doesn't have a parent result, which I am not sure is the cleanest, but I guess other methods are more messy.

src/parser/peg/transformer/peg_transformer.cpp
Dtenwolde: Hmm it shouldn't because there should never be cycles (that should already never pass the parser I don't think). Anyhow, I don't think it's bad to add a check in a debug function :)

kryonix: True that. But we are handling stack frame indexes kind of manually. So, theoretically, we could introduce loops there. (Which would be a bug.)

scripts/parser/generate_transformer_trampoline.py
Dtenwolde: Going to leave that for a future PR :)

src/main/client_context.cpp
```diff
@@ -1535,6 +1535,7 @@ ParserOptions ClientContext::GetParserOptions() const {
 	ParserOptions options;
 	options.preserve_identifier_case = Settings::Get<PreserveIdentifierCaseSetting>(*this);
 	options.integer_division = Settings::Get<IntegerDivisionSetting>(*this);
+	options.transformer_trampoline_style = Settings::Get<TransformerTrampolineStyleSetting>(*this);
```
Mytherin: I don't quite follow why we need / want this to be a setting. Can we remove this?

Mytherin (commented):
Thanks! Great that you got this to work - one comment

src/main/client_context.cpp
Dtenwolde: I need this because there's no `ClientContext` in the transformer (https://github.com/duckdblabs/duckdb-internal/issues/9703) and this is the only way to pass this option. (Unless I am missing something)

My plan is to have it for 1) this PR, 2) the follow-up PR with the generated code, where I add a test config and run all the tests with the trampoline transformer enabled. And finally, in PR 3 I plan to make the switch to make trampoline default and remove the setting.

Mytherin: Alright - if the goal is to remove it then that's fine by me. Maybe prefix the setting with "debug_" if we intend to remove it

Mytherin:
Thanks!

PR #23415, opened 2026-06-23
Use std::shared_mutex in StorageLock

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/8928

With this PR, we use `std::shared_mutex` in `StorageLock` while preserving the existing checkpoint locking semantics. We still need a bit of scaffolding around `std::shared_mutex` to preserve the existing writer-prioritization, i.e., no new shared locks can be acquired when an exclusive lock was requested. Moreover, this introduces `CheckpointLock`, which encapsulates the logic to upgrade a shared lock to an exclusive lock if it is the only existing shared lock.

### DIFF
--- a/src/function/table/table_scan.cpp
+++ b/src/function/table/table_scan.cpp
@@ -30,6 +30,7 @@
 #include "duckdb/planner/filter/table_filter_functions.hpp"
 #include "duckdb/transaction/duck_transaction.hpp"
 #include "duckdb/transaction/local_storage.hpp"
+#include "duckdb/transaction/vacuum_lock.hpp"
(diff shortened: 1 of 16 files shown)


lnkuiper:
Good idea! Thread sanitizer doesn't seem happy yet, but that should be fixable.

Maybe we can also add `shared_mutex` to the namespace in `mutex.hpp` so we can drop the `std::` prefix.

dentiny:
> Maybe we can also add shared_mutex to the namespace in mutex.hpp so we can drop the std:: prefix.

Notice this PR -- I think I should port all the mutex we use to annotated version as well.

PR #23416, opened 2026-06-23
Fix #23405: Pass args.size() to UnaryExecutor in JSON extraction to respect selection vectors

### LINKED ISSUE #23405: Conversion Error when a JSON arrow-operator condition follows another condition in WHERE
### What happens?

The JSON arrow operator causes a Conversion Error if it is not the first condition.

### To Reproduce

```sql
from (values ('{"k":"v"}')) t(d)
where d is null or d->>'k' = 'v';
```
errors with
```
Conversion Error:
Could not convert string '{"k":"v"}' to BOOL when casting from source column d

LINE 2: where d is null or d->>'k' = 'v';
                           ^
```

The error can be avoided by flipping the conditions:
```sql
from (values ('{"k":"v"}')) t(d)
where d->>'k' = 'v' or d is null
```

### OS:

macOS 14.8.7 (arm64)

### DuckDB Version:

1.5.4

### DuckDB Client:

wasm, cli

### Hardware:

_No response_

### Full Name:

Fabian Iwand

### Affiliation:

-

- [x] Yes, I have

- [x] Yes, I have

Yes

arham0254A:
->PR Description
Problem: 
When evaluating `WHERE` clauses, the JSON extraction operators (like `->>`) were ignoring the active selection vectors generated by preceding filter conditions. Because `args.size()` was omitted, `UnaryExecutor` defaulted to eagerly parsing the entire chunk. This caused a `Conversion Error` to be thrown when the chunk contained invalid JSON strings that were supposed to be filtered out by earlier conditions.

Solution: 
Passed `args.size()` into the three `UnaryExecutor::Execute` calls inside `json_executors.hpp` (for both standard and wildcard binary execution paths). This ensures the executor correctly references the selection vector and only parses valid, unfiltered rows.

Closes #23405

### DIFF
--- a/extension/json/include/json_executors.hpp
+++ b/extension/json/include/json_executors.hpp
@@ -29,7 +29,7 @@ public:
 		auto alc = lstate.json_allocator->GetYYAlc();
 
 		const auto &inputs = args.data[0];
-		UnaryExecutor::Execute<string_t, T>(inputs, result, [&](string_t input) -> optional<T> {
+		UnaryExecutor::Execute<string_t, T>(inputs, result, args.size(), [&](string_t input) -> optional<T> {
 			auto doc = JSONCommon::ReadDocument(input, JSONCommon::READ_FLAG, alc);
 			return fun(doc->root, alc, result);
 		});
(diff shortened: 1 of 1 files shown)


Mytherin:
This is invoking a deprecated method that only exists in `main`. The error is reported in `v1.5.4`. I can't see how this fixes the issue.

PR #23417, opened 2026-06-23
Use github script to add member label and move PR to draft

smvv:
This PR uses `pull_request_target`, but i carefully read GitHub Security Lab's [Preventing pwn requests](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/):

> 
> The article says `pull_request_target` is intended for cases like labeling or commenting that “do not require dangerous processing” of PR contents, and calls it a “logical shortcut” for that kind of metadata-only automation.
>
> Your workflow does exactly that:
> - it reads only GitHub event metadata in the `Prepare` step
> - it uses `actions/github-script` to call GitHub APIs
> - it does **not** use `actions/checkout`
> - it does **not** run PR scripts, builds, tests, or package managers
> - it does **not** read artifacts from an untrusted workflow
>
> It avoids the insecure pattern: `pull_request_target` should not check out or execute untrusted PR code.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/9288.


Mytherin:
Thanks!

PR #23419, opened 2026-06-23
Refactor Multi-File Scan for ASYNC I/O

pdet:
This PR refactors the multi-file scan to split scheduling from decoding, it also introduces one new virtual function: `BaseFileReader::ScheduleIO`, that readers can override if they do support ASYNC I/O scheduling.

### DIFF
--- a/src/include/duckdb/common/multi_file/multi_file_function.hpp
+++ b/src/include/duckdb/common/multi_file/multi_file_function.hpp
@@ -480,11 +480,8 @@ public:
 					if (old_file_index != scan_data.file_index) {
 						InitializeFileScanState(context, current_reader_data, scan_data, gstate.projection_ids);
 					}
-					// schedule I/O while holding the parallel lock, so each unit is scheduled once (This is important
-					// for read-ahead)
-					parallel_lock.lock();
-					scan_data.scheduled_io =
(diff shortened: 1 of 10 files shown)


src/include/duckdb/common/multi_file/multi_file_function.hpp
```diff
@@ -662,6 +664,86 @@ class MultiFileFunction : public TableFunction {
...
+		if (scheduled.GetResultType() == AsyncResultType::BLOCKED) {
+			return HandleBlocked(data_p, scheduled);
+		}
+		return false;
+	}
+
+	static bool DecodePhase(ClientContext &context, TableFunctionInput &data_p, MultiFileLocalState &data,
+	                        MultiFileGlobalState &gstate, MultiFileBindData &bind_data, DataChunk &output) {
+		auto &scan_chunk = data.scan_chunk;
+		if (!data.reader->OwnsChunkReset()) {
```
Mytherin: I don't quite follow the `OwnsChunkReset` requirement here - as far as I understand it we just shouldn't be resetting the chunk if a previous iteration has returned `BLOCKED`. Isn't that the correct modeling here that would also solve the Parquet use case?

Mytherin (commented):

Mytherin:
Thanks!

PR #23426, opened 2026-06-23
Bind syntax sugar to generic `field`, `element` and `slice`

Maxxen:
This PR changes the `a.x` / `x[n]` and `x[n:m]` syntax forms to be bound to a set of generic `field`, `element` and `slice` functions. 

This makes it easy to support these syntax forms for custom types by simply overloading these functions, similar to how we the `IN` syntax is extensible by overloading `contains`, instead of having the binder hard-code the supported types and special-case to e.g. struct_extract, array_extract etc etc.

Im not 100% on the naming, but I figured since these are pretty fundamental operations it might be nice to keep them short, instead of prefixing with e.g. a `get_field`/`get_element`. 

The old `struct_extract`/`array_extract` and friends remain for backwards compatability - but could probably be changed to macros or something now.

Speaking of which, a follow-up task would be to also attempt binding macros when resolving these functions, which would enable end-users to opt into these syntax forms for arbitrary types through SQL.

Unfortunately I've had to keep some ugliness and type-special casing in the binder as both the variant and JSON type are doing a bunch of implicit-cast-rule-circumvention crimes I don't have patience to deal with now.


Mytherin:
Thanks for the PR!

I think this is a great idea. My only concern here is backwards compat of plans - but I think since we're doing this in the binder only and are keeping the old functions around that this should work as-is. Turning the old functions from scalar functions into macros will run into plan deserialization issues, though, so we probably need to (at least for now) keep those as functions.

As for naming - maybe using `extract_` makes intent more clear which matches our previous naming conventions here (`extract_field`, `extract_element`, `extract_slice`). I suppose naming is not really important anyway since users will not be typing these function names.

PR #23427, opened 2026-06-23
Skip caching when no version metadata available

### LINKED ISSUE #23255: Malformed JSON error fetching ND-JSON twice from same URL if data changes
### What happens?

I have a script which updates data on a server and fetches an ND-JSON report before and after the update. After the update DuckDB **1.5.3** fails with an error like:
```
Malformed JSON in file "http://127.0.0.1:8000/", at byte 1 in line 4: unexpected character.
```
The line number is always two more than the number of lines in the report.
DuckDB **1.4.4** is fine.

### To Reproduce

Run this Python HTTP server:
[test_server.py](https://github.com/user-attachments/files/28886895/test_server.py)
```python
#!/usr/bin/env python3

import json
from http.server import BaseHTTPRequestHandler, HTTPServer
from io import StringIO
from itertools import count


class JsonServer(BaseHTTPRequestHandler):
    _incr = count(0)

    def do_headers(self):
        self.send_response(200)
        self.send_header("Content-Type", "application/x-ndjson")
        self.end_headers()

    def do_HEAD(self):
        self.do_headers()
        self.wfile.write(b"")

    def do_GET(self):
        self.do_headers()

        n = next(self._incr)
        rspns = StringIO()
        for i in range(2):
            rspns.write(
                json.dumps(
                    {
                        "a": i,
                        "b": "x" * n,
                    }
                )
            )
            rspns.write("\n")
        self.wfile.write(rspns.getvalue().encode("utf-8"))


if __name__ == "__main__":
    httpd = HTTPServer(("127.0.0.1", 8000), JsonServer)
    httpd.serve_forever()
```
(It changes the contents of the lines served with each response, but not the number of lines.)

and then do:
```sh
duckdb -c "FROM read_json('http://127.0.0.1:8000/'); FROM read_json('http://127.0.0.1:8000/')"
```

### OS:

MacOS

### DuckDB Version:

1.5.3

### DuckDB Client:

CLI, Python

### Hardware:

Apple M1

### Full Name:

James Gilbert

### Affiliation:

Wellcome Sanger Institute

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23255

The current problem is: when there's no version information available for the requested file (neither last modification timestamp nor etag), file blocks are still cached in external file cache, but we have no way to validate its freshness so it's prune to stale read.

There're two ways I could think of
- Do not cache files without version information (implemented in this PR)
- Cache all files, whether version info available, but requires extra setting on read (for example, [disable external file cache validation](https://github.com/duckdb/duckdb/blob/e884eef8ddaa9434dc33715b1aaa7fec29f8eb4b/src/include/duckdb/main/settings.hpp#L1964)) to explicitly skip validation

I prefer (1) because 
- etag should be generally available for remote storage like S3/GCS, so practically speaking it shouldn't affect perf
- it doesn't require users to do extra setup and prioritize correctness

### DIFF
--- a/src/storage/external_file_cache/caching_file_system.cpp
+++ b/src/storage/external_file_cache/caching_file_system.cpp
@@ -257,8 +257,11 @@ FileBufferHandleGroup CachingFileHandle::Read(const idx_t nr_bytes, const idx_t 
 		return FileBufferHandleGroup();
 	}
 
-	// Uncached files are read directly into one contiguous buffer, skipping the per block syscalls and copies
-	if (!external_file_cache.IsEnabled() || !external_file_cache.ShouldCacheFile(path.path)) {
+	// Only cache when file metadata is available.
+	const bool no_validation_metadata =
+	    Validate() && version_tag.empty() && (!last_modified.IsFinite() || last_modified == timestamp_t(0));
+
+	if (!external_file_cache.IsEnabled() || !external_file_cache.ShouldCacheFile(path.path) || no_validation_metadata) {
 		auto buf = external_file_cache.GetBufferManager().Allocate(MemoryTag::EXTERNAL_FILE_CACHE, nr_bytes);
 		ReadAndRecord(context, buf.GetDataMutable(), nr_bytes, location);
 		vector<FileBufferHandleGroup::MemoryHandle> mem_handles;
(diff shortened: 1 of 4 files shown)


josephbuckingham:
I found the last modified field to initialize to random values which would probably avoid the check you implemented.

Mytherin:
Thanks - should this target `v1.5-variegata`?

dentiny:
> Thanks - should this target `v1.5-variegata`?

External file cache is largely refactored, which I'm afraid hard to cherry-pick, so I usually file two PR recently, one against main, another against v1.5

lnkuiper:
Thanks!

PR #23429, opened 2026-06-24
fix out-of-bounds read in substring with negative offset

jmestwa-coder:
**out-of-bounds read in substring with a negative offset**

on the backward scan, when the input has no utf-8 character boundary (for example bytes that are all continuation bytes), `start_pos` never gets set and the loop that skips continuation bytes walks past the end of the buffer. the `end_pos` loop directly below already bounds itself with `< input_size`, so applied the same bound to `start_pos`.

### DIFF
--- a/src/function/scalar/string/substring.cpp
+++ b/src/function/scalar/string/substring.cpp
@@ -139,7 +139,7 @@ string_t SubstringUnicode(Vector &result, string_t input, int64_t offset, int64_
 				}
 			}
 		}
-		while (!IsCharacter(input_data[start_pos])) {
+		while (start_pos < input_size && !IsCharacter(input_data[start_pos])) {
 			start_pos++;
 		}
 		while (end_pos < input_size && !IsCharacter(input_data[end_pos])) {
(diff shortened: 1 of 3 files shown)


lnkuiper:
Thanks for the PR. Looks good, but I was wondering if this is possible to trigger through SQL? I can't seem to trigger an error using:
```sql
select substring(repeat(chr(128), 20), -1, 5);
```

jmestwa-coder:
right, `chr(128)` encodes U+0080 as valid utf-8 (`0xC2 0x80`), so `repeat(chr(128), 20)` has a character boundary every 2 bytes. the backward scan finds one, sets `start_pos`, and the continuation-byte loop stops well inside the buffer. that's why the SQL doesn't trip it.

to hit the read you need a VARCHAR of raw continuation bytes with no boundary (all `0x80`), which is invalid utf-8. SQL ingestion validates utf-8 (literals, csv, parquet, read_text, decode), so you can't produce that string from pure SQL. the unvalidated paths are the C API / appender / extensions that build a VARCHAR directly: `StringVector::AddString` doesn't validate and `string_t::VerifyUTF8` is DEBUG-only. the `test/common/test_substring_oob.cpp` case in this PR constructs that input and fails under ASan at substring.cpp:142 before the patch.

lnkuiper:
Thanks for the clarification!

PR #23432, opened 2026-06-24
Make regex operators match PostgreSQL behaviour

lnkuiper:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/8627

DuckDB's `~` and `!~` have meant full match for a long time, which does not match PostgreSQL behaviour ([documented here](https://duckdb.org/docs/current/sql/dialect/postgresql_compatibility#behavior-of-regular-expression-match-operators)). This PR changes the operator semantics to a partial match to match. I've also implemented `~*` and `!~*`. The previous behaviour can be restored using `SET regex_match_operator_semantics = 'FULL';` (defaults to `'PARTIAL'`). I had to be careful to make this work for the `COLUMNS` clause too but I think it worked.

Note that this also applies to `SIMILAR`. I've ran the test queries in PostgreSQL to make sure that the behaviour matches.


Mytherin:
Thanks - can we keep `SIMILAR TO` the same as before? We have different semantics there than Postgres anyway (they don't use regex - they have their own text parser algorithm there).

lnkuiper:
Actually, this already separated the logic for `SIMILAR` from the operators, I just forgot to update my test. I think this should be OK now, `SIMILAR` behaviour is the same as it always was, only the operators have changed

Mytherin:
Could we maybe verify that if we create a view with these operators in v1.5, and then read with these changes, that we would get the same behavior as in v1.5? i.e. if we do this in v1.5:

```sql
create view v1 as select * from tbl where col ~ '<cond>'
```

And then read that view in the new version, we get the same result regardless of the setting.

I think because this is a transformer-only transformation that should work given we seem to be converting to the named functions in the transformer.

lnkuiper:
Excellent idea, done!

PR #23437, opened 2026-06-24
Allow non-"list" LIST child name in remap_struct

regadas:
`remap_struct` keys a LIST's child as `"list"`, but the multi-file reader names it after the Parquet repeated group — which is `array` (parquet-avro/Hive) or `bag` (Spark legacy), not always `list`. So a `LIST<STRUCT>` that needs remapping throws `Source value array not found`.

A LIST only has one child, so the name is unambiguous — fall back to the single child instead of throwing. STRUCT/MAP left as-is.

Ran into this through the Iceberg name-mapping path, but it hits any reader doing a nested LIST remap.


lnkuiper:
Thanks!

PR #23438, opened 2026-06-24
Fix type inconsistency for thread count variables

c-herrewijn:
Contains fix for: https://github.com/duckdblabs/duckdb-internal/issues/9785

Data types used for thread counts are currently a mix of `idx_t` and `int32_t`. Which requires a lot of casting and opens the door for bugs like the one above.

This PR homogenizes all to `idx_t`
This is in line with:
  - `DBConfigOptions.maximum_threads`  (already of type `idx_t`)
  - `ParquetMetadataGlobalState.max_threads` (already of type `idx_t`)
  - Note `Select current_setting('threads')` returns type `int64`.

### changes
- change type of `requested_thread_count` and `current_thread_count` variables to type `idx_t` 
- removes `NumericCast<idx_t>()` where they are no longer needed

### tests
I didn't add a test, because the test scenario would require something like: `set threads=2147483649;` which is probably not wise in resource constrained test settings.


Mytherin:
Thanks!

PR #23440, opened 2026-06-24
Httplib backport security issues

ProjectMutilation:
Provided security patches from upstream of yhirose/cpp-httplib

### DIFF
--- a/third_party/httplib/httplib.hpp
+++ b/third_party/httplib/httplib.hpp
@@ -9207,18 +9207,10 @@ inline bool ClientImpl::create_redirect_client(
     // Setup basic client configuration first
     setup_redirect_client(redirect_client);
 
-    // SSL-specific configuration for proxy environments
-    if (!proxy_host_.empty() && proxy_port_ != -1) {
-      // Critical: Disable SSL verification for proxy environments
-      redirect_client.enable_server_certificate_verification(false);
-      redirect_client.enable_server_hostname_verification(false);
-    } else {
-      // For direct SSL connections, copy SSL verification settings
-      redirect_client.enable_server_certificate_verification(
-          server_certificate_verification_);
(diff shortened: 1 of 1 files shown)


ProjectMutilation:
@Mytherin @lnkuiper hi, failed job is CI problem 
```
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
```
Can you help with it?

third_party/httplib/httplib.hpp
```diff
@@ -9425,7 +9471,11 @@ inline bool ClientImpl::write_request(Stream &strm, Request &req,
     }
 
     // Write request line and headers
-    detail::write_request_line(bstrm, req.method, path_with_query);
+    if (!detail::check_and_write_headers(bstrm, req.headers, header_writer_,
```
lnkuiper, line 9476: Upstream httplib has:
```C++
detail::write_request_line(bstrm, req.method, path_with_query);
```
Before these lines, see: https://github.com/yhirose/cpp-httplib/blob/98048a033a532ff22320ce1d11789f8d5710dfcd/httplib.h#L10975-L10978

lnkuiper (commented):
Thanks for the PR! I've compared with upstream httplib and I think there's a small difference. Or maybe I'm checking the wrong hash. Can you have a look?

third_party/httplib/httplib.hpp
ProjectMutilation: My bad, I'll fix it, thanks

ProjectMutilation: fixed

PR #23442, opened 2026-06-24
fix out-of-bounds read in UTF8ToCodepoint on truncated input

jmestwa-coder:
Utf8Proc::UTF8ToCodepoint reads the continuation bytes implied by the lead byte without checking how many are left:
- a heap-backed string_t (> 12 bytes, not null-terminated) ending in a truncated 2/3/4-byte sequence reads past the buffer end
- such non-utf8 values reach string functions through the C API / appender, which skip utf8 validation
- hit by upper/lower, translate, and reverse/length via the grapheme iterator

The decoder now takes the bytes available and rejects an incomplete trailing sequence before reading it; callers with null-terminated buffers keep the default. test/common/test_utf8_codepoint_oob.cpp fails under ASan before the change.


jmestwa-coder:
Folding this into #23459 so both can be reviewed together, as requested. That PR now covers the same call sites plus ascii(), the shell render path, and FindNextLegalUTF8.

PR #23444, opened 2026-06-25
Fix eviction node memleak when external file cache disabled

dentiny:
Related to https://github.com/duckdb/ducklake/issues/1136

@zhming0's using DuckLake for continuous metrics data ingestion. During the process he found memory leak with the following setup and phenomenon. Credit to Zhiming for the repro script!
- Workload: long-running ingestion + merge
- External file cache is disabled
- While container RSS keeps growing until OOM, `duckdb_memory` doesn't report it

I'm able to reproduce the issue on my end, and I found a lot of dead nodes accumulated in eviction queue with `duckdb_eviction_queues`.

The root cause is
- In the current implementation, when external file cache disabled, we always allocate a buffer, read and return it right away
- The default [`DestroyBufferUpon`](https://github.com/duckdb/duckdb/blob/main/src/include/duckdb/common/enums/destroy_buffer_upon.hpp) option is `EVICTION`, which means on block destruction it's placed into eviction queue
- Eviction queue is [only visited and evicted](https://github.com/duckdb/duckdb/blob/0361de441a122f11fe07c40ce4531cd58f5eaf7a/src/storage/buffer/buffer_pool.cpp#L396-L401) when buffer manager detects insufficient memory **under its management**
- The situation for us is, the overall memory consumption under buffer manager's track is low, but quite a lot is wasted upon useless eviction nodes

This PR switches the DestroyBufferUpon mode from EVICTION to UNPIN, which destructs memory directly within going through the eviction queue. I've verified eviction queue size shows empty after my change.

### DIFF
--- a/src/storage/external_file_cache/caching_file_system.cpp
+++ b/src/storage/external_file_cache/caching_file_system.cpp
@@ -3,11 +3,13 @@
 #include "duckdb/common/checksum.hpp"
 #include "duckdb/common/chrono.hpp"
 #include "duckdb/common/enums/cache_validation_mode.hpp"
+#include "duckdb/common/enums/destroy_buffer_upon.hpp"
 #include "duckdb/common/enums/memory_tag.hpp"
 #include "duckdb/common/file_system.hpp"
 #include "duckdb/main/client_context.hpp"
 #include "duckdb/parallel/task_executor.hpp"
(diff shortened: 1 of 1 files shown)


lnkuiper:
Looks great, thanks!

PR #23447, opened 2026-06-25
Add dictionary-aware cache operator

ahmetkarapinar:
# Add dictionary-aware cache operator

Follow-up to #23360, implementing [[Mark's suggestion](https://github.com/duckdb/duckdb/pull/23360#discussion_r3436616932)](https://github.com/duckdb/duckdb/pull/23360#discussion_r3436616932): instead of keying the cache's dictionary accumulation on the `global_dictionary` flag, accumulate **any** dictionary with a stable id, and flatten only if the id changes (which never happens for global dictionaries).

## Motivation

#23360 only carried a dictionary through the cache when it was **global**. But many runtime dictionaries already carry a stable, non-empty `DictionaryEntry::id`; the only thing special about a global one is its lifetime contract, not its id. So storage (`DictionaryCompressionStorage`, `DictFSST`), Parquet, and expression-emitted dictionaries all have stable ids and were simply flattened on the way into the cache. This PR accumulates them too.

## Benefits

The cache no longer materializes these dictionary columns into flat values. Instead of copying the full value for every row of every small chunk, it just appends the selection indices and re-wraps the dictionary on flush, so the distinct values stay in the shared child and are never re-materialized. This saves both CPU and memory.

## Approach

The cache stops reading the `global_dictionary` flag for admission. It now admits any `DICTIONARY_VECTOR` with a non-empty id, and keys per-chunk identity on **entry-pointer equality**. Per seeded column:

- **same dictionary** → concatenate selection indices into `accumulated_sel` (steady state, identical to the prior global path).
- **dictionary changed, non-global** → **flatten that one column in place** into its existing flat cache slot and ride flat for the rest of the fill; it re-seeds on the next fill.
- **dictionary changed, global** → **throw** (a global dict rotating is a producer bug, and the sink depends on it not happening).

---

@Mytherin @lnkuiper, this is the follow-up to your suggestion on #23360. Ready for review, thanks!

### DIFF
--- a/src/execution/physical_operator.cpp
+++ b/src/execution/physical_operator.cpp
@@ -416,23 +416,28 @@ SelectExecutionMode(const DataChunk &chunk, const OperatorResultType child_resul
 	return CachingPhysicalOperatorExecuteMode::RETURN_CHUNK;
 }
 
-static bool ChunkHasGlobalDictionary(const DataChunk &chunk) {
+//! Empty-id dicts (e.g. slice-of-flat) mint a fresh entry per slice, so their sels must not be accumulated.
+static inline bool IsAccumulableDictionary(const Vector &vector) {
+	return vector.GetVectorType() == VectorType::DICTIONARY_VECTOR && !DictionaryVector::DictionaryId(vector).empty();
+}
+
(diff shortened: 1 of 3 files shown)


lnkuiper:
Looks great! Thanks

PR #23453, opened 2026-06-25
Copy table_filters to LHS scan in late materialization

baolinhuang:
Problems：
========
  `LateMaterializationHelper::CreateLHSGet` does not copy `table_filters` from the RHS scan to the LHS scan. This causes the LHS  (which reads all payload columns) to rely solely on bloom filter + rowid range dynamic filters pushed down from the HASH_JOIN.  When the target rows are a small fraction of the table and their rowids are interleaved with other groups, the LHS candidate set is far larger than necessary, leading to:

  1. Massive bloom filter false positives (candidate rows × FPR)
  2. Expensive reads of wide payload columns for rows that will be discarded by the SEMI JOIN
  3. Late materialization becoming **slower** than no late materialization (negative performance)

Solution：
=======
  Copy `table_filters` to the LHS `LogicalGet` in `CreateLHSGet`:

Test：
====
### LIMIT 50 (median of 5 runs)

| Scenario | Time (ms) | vs Baseline |
|----------|-----------|-------------|
| No late materialization (baseline) | **78** | — |
| Late mat WITHOUT table_filters copy (before fix) | **95** | **1.22× slower (negative!)** |
| Late mat WITH table_filters copy (after fix) | **48** | **1.6× faster** |

### EXPLAIN ANALYZE diff

**Before (LHS TABLE_SCAN — no filters):**

```
TABLE_SCAN wide_table
  Type: Sequential Scan
  Projections: operator_id, ts, id, col1..col20
  (no Filters)                          ← only dynamic rowid/BF filters
  50 rows, 0.05s
```

**After (LHS TABLE_SCAN — has filters):**

```
TABLE_SCAN wide_table
  Type: Sequential Scan
  Projections: operator_id, ts, id, col1..col20
  Filters: (operator_id = 1)            ← table_filters copied from RHS
           (ts < '2024-04-01')
           optional: Dynamic Filter (ts)
  50 rows, 0.01s
```

### DIFF
--- a/src/optimizer/late_materialization_helper.cpp
+++ b/src/optimizer/late_materialization_helper.cpp
@@ -1,9 +1,10 @@
 #include "duckdb/optimizer/late_materialization_helper.hpp"
+#include "duckdb/planner/filter/expression_filter.hpp"
+#include "duckdb/planner/filter/table_filter_functions.hpp"
 
 namespace duckdb {
 
 unique_ptr<LogicalGet> LateMaterializationHelper::CreateLHSGet(const LogicalGet &rhs, Binder &binder) {
-	// we need to construct a new scan of the same table
 	auto table_index = binder.GenerateTableIndex();
 	auto new_get = make_uniq<LogicalGet>(table_index, rhs.function, rhs.bind_data->Copy(), rhs.returned_types,
 	                                     rhs.names, rhs.virtual_columns);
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks for the PR! It looks like this is causing CI failures, however. Could you have a look?

Since the rows are already filtered out - I think we also want to wrap the filters in an `optional_filter` so that these are only used for zone-map pruning, and not actually executed. Otherwise this might cause regressions as we're doing additional (unnecessary) computation.

baolinhuang:
Have pushed a new commit include 2 modifications.
1.  Skip filters containing dynamic expressions. If not the dynamic filter references shared mutable DynamicFilterData whose boundary is updated by the RHS pipeline  and would be stale on the LHS.
2. Wrap copied filters in OptionalFilter.

Mytherin:
Thanks!

PR #23456, opened 2026-06-25
Clear stale projection maps after CTE rewrite pass

kryonix:
Common subplan materialization can replace child subplans with CTE refs after column_lifetime has installed positional projection maps. Those maps then describe the pre-replacement child layout and can hide bindings still referenced above the join.

The reproducer is unfortunately big, but the bug only triggers in some very specific scenarios.

fix: https://github.com/duckdblabs/duckdb-internal/issues/9778


lnkuiper:
Looks good, thanks!

PR #23458, opened 2026-06-25
Fix false RLE corruption error

DinosL:
More info: https://github.com/duckdblabs/duckdb-internal/issues/9791#issuecomment-4798551979


Mytherin:
Thanks!

PR #23459, opened 2026-06-25
bound multi-byte reads in UTF8ToCodepoint

jmestwa-coder:
UTF8ToCodepoint reads up to 4 bytes off the leading byte but took no length, so a string ending in a truncated multi-byte sequence reads past the buffer:
- reachable on untrusted varchar via upper/lower, translate, ascii, and the grapheme helpers behind length_grapheme/substring
- a heap-backed (> 12 byte, non-terminated) string ending in a lead byte with missing continuation bytes is a heap-buffer-overflow read (ASan, size 1)
- added a length arg that rejects truncated sequences, passing the remaining byte count at every call site
regression test under test/common reads out of bounds on the unpatched tree under ASan

### DIFF
--- a/extension/core_functions/scalar/string/ascii.cpp
+++ b/extension/core_functions/scalar/string/ascii.cpp
@@ -12,7 +12,7 @@ struct AsciiOperator {
 			return str[0];
 		}
 		int utf8_bytes = 4;
-		return Utf8Proc::UTF8ToCodepoint(str, utf8_bytes);
+		return Utf8Proc::UTF8ToCodepoint(str, utf8_bytes, input.GetSize());
 	}
 };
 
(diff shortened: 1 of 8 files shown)


lnkuiper:
This has failed CI, and the failure seems related. Could you have a look?

lnkuiper:
This also seems to be touching a lot of the same files as https://github.com/duckdb/duckdb/pull/23442 please add those changes to this PR and close the other one so it can be reviewed together

jmestwa-coder:
both addressed:

- ci: the truncated-sequence path threw `InternalException`, which aborts under `CRASH_ON_ASSERT` in the relassert configs. since this is reachable from untrusted varchar it should be a user-facing error, so it now throws `InvalidInputException` (same for the existing invalid lead byte / surrogate cases). the regression test passes under relassert now.
- closed #23442 and pulled its grapheme-iterator test coverage in here. this PR already covered the same call sites plus ascii(), the shell render path, and FindNextLegalUTF8, so it's the superset.

verified `[utf8]` and the upper/lower/translate/ascii/reverse tests pass, and upper/lower/translate/ascii/length_grapheme/reverse still behave correctly on valid input.

PR #23462, opened 2026-06-25
Implement and use a shadow index while checkpointing

sebastiaan-dev:
This PR resolves https://github.com/duckdblabs/duckdb-internal/issues/9661 and is a prerequisite for the colocation of index buffers across tables. 

The main goal of this PR is to change the checkpointing for ART indexes. Previously, ART checkpointing updated the block references of the live buffers directly. This was safe because the index lock was held until the newly produced blocks were flushed to disk.

However, this model breaks down when we want to colocate the buffers of multiple indexes. We would need to retain every index lock we encounter until we have iterated over all catalog entries and any newly produced blocks have been flushed. As a result indexes would be unavailable for most of the duration of the checkpoint.

This PR introduces a new `TableIndexWriter` and a corresponding `SingleFileTableIndexWriter` to manage indexes during a checkpoint. ART checkpointing now constructs a shadow index from the live index while holding the index lock. Unmodified buffers are referenced from the existing index, while dirty buffers are copied into new partial blocks. Once the shadow index has been constructed, the live index lock can be released. After all indexes for the table have been checkpointed, the writer flushes the new blocks and the table index list installs the shadow indexes in place of the live ones.

As changes to index code can have downstream effects, we introduce a checkpointing mode for indexes. By default this is set to `IMMEDIATE` which flushes blocks before releasing the lock of the respective index. When colocation across distinct indexes is desired, the mode can be overridden with the `DEFERRED` mode. The latter is supported only when the index also supports delta indexes.

Of note is that this PR also enables colocation for different indexes in the same table, but it does not colocate indexes across different tables.

### Noteworthy Changes

- Removed the `RemoveEmptyBuffers()` call from `PrepareSerialize()` in favor of inlining this behavior in the construction of the shadow, which changes the model from pruning buffers to constructing for serialization.
- Removed `IndexSerializationResult`, which maintained an ordering invariant across two vectors.
- Changed `UnboundIndex` to store its `IndexStorageInfo` in a `shared_ptr<const IndexStorageInfo>`, this solves a potential future case where a reference could be invalidated by transforming an `Unbound` index into a `Bound` index during the checkpoint.


artjomPlaunov:
@sebastiaan-dev Thanks :-) I will come back with a more thorough review soon, but until then it would also be good to add some testing around this. Some things I can think of that would be worthwhile to test: 

1) (definitely add this one) colocation of different indexes on the same table (i.e. writing out a bunch of small indexes on the same table), i.e. check that there isn't a block being written for each index. 

also some ideas for other tests, although only add these if they seem to meaningfully test the new code in some way/aren't already encoded in existing tests: 

2)  trigger a failure during shadow art construction in the checkpoint (checkpoint_abort.hpp, could add a new enum there if needed), before it is swapped out
3) concurrent index operations while checkpointing

sebastiaan-dev:
The current implementation has an outstanding issue where a checkpoint invalidates all `Index&` references. This happens for example in the spatial extension, which stores the `Index&` in the bind data and later uses it during scan execution.

Storing `Index&` references was safe previously, as the `Index` itself was stable across its lifetime. This PR changes this model: During checkpointing a swap makes `entry->index` (from the `TableIndexList`) point to the shadow index. Thus making any existing references to the old `Index` stale.

artjomPlaunov:
@sebastiaan-dev as discussed with @taniabogatsch I think we will handle https://github.com/duckdblabs/duckdb-internal/issues/9840 first, then get back to this PR once that is implemented :-)

artjomPlaunov (commented):
Thanks for the PR! I'm still spending some time digging into all the details, will take some time to get back with a more thorough review, but in the meantime two general points/questions: 

1) What are the performance tradeoffs here? From a glance it looks like we might have to do a lot more work for clean ART buffers than before, and similarly for VSS indexes. 

2) Let's add more tests, more generally to verify the benefits of packing multiple indexes from the same table, but also to cover more edge cases in the new code :-)

src/storage/data_table.cpp
```diff
@@ -1595,13 +1595,16 @@ void DataTable::Checkpoint(TableDataWriter &writer, Serializer &serializer) {
...
-	//   table pointer
-	//   index data
-	writer.FinalizeTable(global_stats, *info, *row_groups, serializer);
+
+	const auto storage_version = serializer.GetOptions().storage_compatibility.storage_version;
+	const auto index_writer = writer.GetTableIndexWriter(storage_version);
+	vector<shared_ptr<const IndexStorageInfo>> index_infos;
+	if (index_writer) {
+		// Only checkpoint indexes when we write to disk
+		index_infos = info->GetIndexes().CheckPoint(*index_writer);
```
artjomPlaunov, line 1604: nit: Checkpoint instead of CheckPoint :-)

src/storage/table_index_list.cpp
```diff
@@ -487,6 +495,29 @@ unordered_set<column_t> TableIndexList::GetIndexedColumns() const {
...
+	annotated_lock_guard lock(index_entries_lock);
+
+	vector<unique_ptr<BoundIndex>> shadows;
+	vector<shared_ptr<const IndexStorageInfo>> infos;
+	shadows.reserve(index_entries.size());
+	infos.reserve(index_entries.size());
+
+	for (const auto &entry : index_entries) {
+		auto checkpoint = entry->Checkpoint(writer);
+		shadows.push_back(std::move(checkpoint.shadow_index));
```
artjomPlaunov, line 508: Maybe we can store each pending shadow with its target IndexEntry, rather than matching by position? This is currently fine with the list lock, but I think once we extend this to cross table packing we wont have this guarantee.

.github/patches/extensions/vss/0002-checkpoint-api.patch
```diff
@@ -0,0 +1,146 @@
...
+ 	info.name = name;
+ 	info.root = root_block_ptr.Get();
+ 
+-	// Use the partial block manager to serialize allocator data.
+-	auto &block_manager = table_io_manager.GetIndexBlockManager();
+-	PartialBlockManager partial_block_manager(context, block_manager, PartialBlockType::FULL_CHECKPOINT);
+-	linked_block_allocator->SerializeBuffers(partial_block_manager);
+-	partial_block_manager.FlushPartialBlocks();
+-	info.allocator_infos.push_back(linked_block_allocator->GetInfo());
++	// Recreate the usearch index from the serialized representation so the shadow owns an independent index.
```
artjomPlaunov, line 76: I might be misunderstanding here, but HNSW uses immediate checkpointing mode, right? And its entry therefore remains locked until flushing completes, so why do we need to do a reconstruction + swap, can't we retain the original "immediate" path?

src/include/duckdb/storage/checkpoint/table_index_writer.hpp
```diff
@@ -0,0 +1,61 @@
...
+public:
+	explicit TableIndexWriter(PartialBlockManager &partial_block_manager, StorageVersion version);
+	virtual ~TableIndexWriter();
+
+public:
+	PartialBlockManager &GetPartialBlockManager() const {
+		return partial_block_manager;
+	}
+
+	virtual PartialBlockManager CreateIsolatedPartialBlockManager() = 0;
```
artjomPlaunov, line 37: Nit: not sure about the name, perhaps something more like: CreatePerIndexPartialBlockManager, and the shared accessor GetSharedPartialBlockManager? Basically trying to make the distinction between packing scopes clearer, not sure what's a good name here

src/storage/index_entry.cpp
```diff
@@ -468,19 +464,44 @@ idx_t IndexEntry::GetInMemorySize() const {
...
+
+	auto partial_block_manager = index_writer.CreateIsolatedPartialBlockManager();
+	auto checkpoint = owned_index->Checkpoint(partial_block_manager, storage_version);
+	partial_block_manager.FlushPartialBlocks();
+
+	SwapInternal(std::move(checkpoint.shadow_index));
+	return checkpoint;
+}
+
+void IndexEntry::Swap(unique_ptr<BoundIndex> shadow_index) {
```
artjomPlaunov, line 487: Maybe we can call it InstallShadow()? since we aren't swapping per se, i.e. we don't get the "swapped" old object back, we are just installing the shadow.

src/include/duckdb/storage/index.hpp
```diff
@@ -57,6 +69,14 @@ class Index {
 	//! The index constraint type
 	virtual IndexConstraintType GetConstraintType() const = 0;
 
+	//! Returns how this index participates in a checkpoint.
+	virtual IndexCheckpointMode GetCheckpointMode() const {
+		return IndexCheckpointMode::IMMEDIATE;
+	}
+
+	//! Checkpoint an index, based on the mode supported by the index flushing may be deferred.
+	virtual CheckpointedIndex Checkpoint(PartialBlockManager &partial_block_manager, const StorageVersion version);
```
artjomPlaunov, line 78: Nit: Maybe rename this to something like PrepareCheckpoint()? It can return a shadow index whose backing has not been flushed yet, so we can distinguish between preparation and the caller's flush + installation.

src/storage/index_entry.cpp
```diff
@@ -468,19 +464,44 @@ idx_t IndexEntry::GetInMemorySize() const {
...
+	auto entry_lock = lock.GetExclusiveLock();
+	SwapInternal(std::move(shadow_index));
+}
+
+void IndexEntry::SwapInternal(unique_ptr<BoundIndex> shadow_index) {
```
artjomPlaunov, line 492: also maybe InstallShadowInternal or something like that

src/include/duckdb/storage/checkpoint/table_index_writer.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "duckdb/storage/index_storage_info.hpp"
+
+namespace duckdb {
+class PartialBlockManager;
+class SingleFileCheckpointWriter;
+class BoundIndex;
+
+//! Checkpointing result for an index. The shadow index may be empty either when
+//! the index is unbound, or the index does not support deferred checkpointing.
+struct CheckpointedIndex {
```
artjomPlaunov, line 22: nit: maybe IndexCheckpointState, or IndexCheckpointResult?

src/execution/index/fixed_size_buffer.cpp
```diff
@@ -239,6 +221,22 @@ void FixedSizeBuffer::SetAllocationSize(const idx_t available_segments, const id
 	allocation_size = max_offset * segment_size + bitmask_offset;
 }
 
+idx_t FixedSizeBuffer::GetNewAllocationSize(const idx_t available_segments, const idx_t segment_size,
```
artjomPlaunov, line 224: nit, mechanical refactor: Could SetAllocationSize call this helper after its existing dirty check? The bitmap scan and size calculation are duplicated now, and both serialization paths should use the same calculation.

src/include/duckdb/execution/index/bound_index.hpp
```diff
@@ -196,12 +196,7 @@ class BoundIndex : public Index {
 	//! Returns true if the index is affected by updates on the specified column IDs, and false otherwise
 	bool IndexIsUpdated(const vector<PhysicalIndex> &column_ids) const;
 
-	//! Serializes index memory to disk and returns the index storage information.
-	virtual IndexStorageInfo SerializeToDisk(QueryContext context, const case_insensitive_map_t<Value> &options)
-	    DUCKDB_EXCLUDES(lock);
-	//! Serializes index memory to the WAL and returns the index storage information.
-	//! The caller must have exclusive access to the index.
-	virtual IndexStorageInfo SerializeToWAL(const case_insensitive_map_t<Value> &options);
+	virtual IndexStorageInfo SerializeToWAL(StorageVersion storage_version);
```
artjomPlaunov, line 199: can we retain/rewrite some of the comments that we removed here to make them up to date?

.github/patches/extensions/vss/0002-checkpoint-api.patch
```diff
@@ -0,0 +1,146 @@
...
++			return size == reader.ReadData(static_cast<data_ptr_t>(data), size);
++		});
++		if (!load_result) {
++			throw InternalException("Failed to load checkpointed HNSW index: %s", load_result.error.what());
++		}
++	}
+ 
+-	return info;
++	auto new_linked_block_allocator = linked_block_allocator->Persist(partial_block_manager);
++	new_linked_block_allocator->RemoveEmptyBuffers();
```
artjomPlaunov, line 90: is this redundant?  (also in spatial patch)

src/include/duckdb/storage/checkpoint/table_index_writer.hpp
```diff
@@ -0,0 +1,61 @@
...
+	virtual ~TableIndexWriter();
+
+public:
+	PartialBlockManager &GetPartialBlockManager() const {
+		return partial_block_manager;
+	}
+
+	virtual PartialBlockManager CreateIsolatedPartialBlockManager() = 0;
+	//! Writes the index buffers to disk
+	virtual void Flush() = 0;
```
artjomPlaunov: Could Flush be implemented directly on TableIndexWriter? The base already holds the shared partial-block manager, and the override just calls FlushPartialBlocks() on it, without any single file specific behavior.

src/include/duckdb/storage/index.hpp
```diff
@@ -14,11 +14,23 @@
 #include "duckdb/common/typedefs.hpp"
 #include "duckdb/common/vector.hpp"
 #include "duckdb/common/unordered_set.hpp"
+#include "duckdb/storage/checkpoint/table_index_writer.hpp"
```
artjomPlaunov, line 17: nit, small mechanical/dependency cleanup: Could we use forward declarations for the checkpoint types here instead of including table_index_writer.hpp? The base Index interface doesn’t use the writer itself, and only needs these types for the method declaration.

src/storage/table_index_list.cpp
```diff
@@ -549,6 +563,14 @@ void TableIndexList::MergeCheckpointDeltas(const optional_idx checkpoint_id) con
 	}
 }
 
+void TableIndexList::Serialize(const vector<shared_ptr<const IndexStorageInfo>> &infos, Serializer &serializer) {
+	// write empty block pointers for forwards compatibility
+	const vector<BlockPointer> compat_block_pointers;
+	serializer.WriteProperty(103, "index_pointers", compat_block_pointers);
```
artjomPlaunov, line 569: Could we keep this in SingleFileTableDataWriter::FinalizeTable? These property IDs and the compatibility field are part of the table's serialized format, and FinalizeTable already writes the surrounding fields. Keeping this there would keep the format more logically together and avoid adding a public helper to TableIndexList for it.

src/execution/index/fixed_size_buffer.cpp
```diff
@@ -80,70 +80,52 @@ FixedSizeBuffer::~FixedSizeBuffer() {
...
-	// Adjust the allocation size.
-	D_ASSERT(segment_count != 0);
-	SetAllocationSize(available_segments, segment_size, bitmask_offset);
-
-	// The buffer is in memory.
-	D_ASSERT(InMemory());
-	if (OnDisk()) {
-		// We copied it onto a new buffer when loading from disk.
-		block_manager.MarkBlockAsModified(block_pointer.block_id);
+		block_manager.IncreaseBlockReferenceCount(block_pointer.block_id);
```
artjomPlaunov, line 99: nit, mechanical refactor: Can we combine this with the clean-buffer branch above? 

Both now increment the reference count and construct the same replacement, regardless of whether the source was already resident. We could handle !dirty && OnDisk() once, then reject !InMemory() before the serialization path.

src/execution/index/art/art.cpp
```diff
@@ -167,6 +172,22 @@ ART::ART(const Identifier &name, const IndexConstraintType index_constraint_type
 	}
 }
 
+ART::ART(const ART &src, shared_ptr<AllocatorArray> allocators_ptr)
```
artjomPlaunov, line 175: let's make this constructor private

src/execution/index/unbound_index.cpp
```diff
@@ -54,13 +55,17 @@ void UnboundIndex::ResetStorage() {
 	}
 }
 
+CheckpointedIndex UnboundIndex::Checkpoint(PartialBlockManager &, StorageVersion) {
+	return {GetStorageInfoPtr(), nullptr};
+}
+
 IndexStorageInfo UnboundIndex::CopyStorageInfo() const {
```
artjomPlaunov, line 62: is this method still used?

src/include/duckdb/execution/index/unbound_index.hpp
```diff
@@ -103,6 +104,9 @@ class UnboundIndex final : public Index {
 		return create_info->Cast<CreateIndexInfo>();
 	}
 	const IndexStorageInfo &GetStorageInfo() const {
+		return *storage_info;
+	}
+	const shared_ptr<const IndexStorageInfo> &GetStorageInfoPtr() const {
```
artjomPlaunov, line 109: maybe we can return {storage_info, nullptr} directly from Checkpoint() and remove this accessor? It's only used within UnboundIndex, so we don't need to expose another public method for accessing the shared pointer.

artjomPlaunov:
Sorry for spamming everyone I was clicking "comment" instead of "start review" 😭 lesson learned

src/execution/index/art/art.cpp
```diff
@@ -167,6 +172,22 @@ ART::ART(const Identifier &name, const IndexConstraintType index_constraint_type
...
+    : BoundIndex(src.name, ART::TYPE_NAME, src.index_constraint_type, src.column_ids, src.table_io_manager,
+                 src.unbound_expressions, src.db),
+      tree(src.tree), allocators(std::move(allocators_ptr)), owns_data(true), storage_version(src.storage_version),
```
artjomPlaunov: maybe target_version to be consistent with serializeToWal?

src/include/duckdb/storage/checkpoint/table_index_writer.hpp
artjomPlaunov, line 37: I think we can also make it private?

src/execution/index/art/art.cpp
artjomPlaunov, line 96: Let's use the new AllocatorArray :-) also in a couple other places I think we can replace it too to make the code easier to read.

src/execution/index/fixed_size_buffer.cpp
```diff
@@ -80,70 +80,52 @@ FixedSizeBuffer::~FixedSizeBuffer() {
...
 		if (!OnDisk() || dirty) {
 			throw InternalException("invalid or missing buffer in FixedSizeAllocator");
 		}
-		return;
+		block_manager.IncreaseBlockReferenceCount(block_pointer.block_id);
+		return make_uniq<FixedSizeBuffer>(block_manager, segment_count, allocation_size, block_pointer);
 	}
 
 	// Early-out, if the buffer is already on disk and not dirty.
 	if (!dirty && OnDisk()) {
```
artjomPlaunov, line 98: More substantial question concerning performance, this is where I would be curious to maybe run some benchmarks and make sure we don't run into any serious slowdowns: 

Previously we had this as a scenario:

1) ART is checkpointed already 
2) queries touch the ART, load some saved buffers into working memory. Buffers remain clean (just doing reads)
3) A later checkpoint processes the index, allocations stay in memory
4) Queries read from the allocations that didn't get evicted 

Now we’re creating replacements and discarding those working allocations when we install the shadow. Of course, this doesn't mean we have to fetch from disk necessarily, since they might stay in memory still, but this is where I'd be curious to see if we run into any regressions + if so, if we can do something here to have the old optimization. Or, if there are regressions but they aren't too bad, we just accept the trade-off and can come back to this as a follow up.

artjomPlaunov (changes requested):
Apologies again for the borked review up above, here are a few more comments. I'm going to do a separate pass over the extension patches again tomorrow, I am still a tad bit unsure about what is going on the extension indexes. 

I think the general approach in core looks good, just code nits mainly. I have one question about performance compared to previous behavior, but it might be an acceptable trade-off for the feature/something we can optimize later.

PR #23273, opened 2026-06-14
Allow empty STRUCT types

Maxxen:
This PR enables STRUCT types to be empty, e.g. `struct_pack()` now returns `{}`.

Syntax-wise, this means that `NULL:STRUCT`, and `CREATE TABLE t1(s STRUCT)` is now valid - `STRUCT` represents an empty struct, but `STRUCT()` is still a syntax-error, as type parameters lists can't be empty, but maybe thats something we want to support in the future.

The existing struct-functions should all work with empty structs, except `UNNEST` (and `struct_extract`) which will throw appropriate errors.

Attempting to persist an empty struct in an older database format will throw an error as well.

PR #23465, opened 2026-06-25
Add `LogicalType::TUPLE`

Maxxen:
This PR adds a new `LogicalType::TUPLE` which replaces the previous "unnamed" special case of struct-types with a dedicated type.

### Syntax Changes

- `{}` now returns a valid empty `STRUCT` (follow up to #23273)
- `()` now returns a valid empty `TUPLE`
- `(x,)` (with a trailing comma) is parsed as a single-element tuple (python style)
- `(a, b, c,)` - trailing commas are ok for tuple literals in general.

This also round-trips in string conversions, e.g. `(1,)::VARCHAR` is printed with a trailing comma.

### JSON/Parquet/Arrow/`VARIANT` conversion

This PR also adds support for json/parquet/arrow conversion of tuples.

- For JSON and `VARIANT`, we follow the python convention and turn tuples into arrays.
- For parquet and arrow, we turn the tuples into structs with dummy names, `element1, element2, ... elementN`, following the same convention as when `UNNEST`ing a tuple.

### Storage Compatibility

- We preserves backwards (and forwards) compatibility by transparently serializing tuples as empty structs when targeting old database versions, and deserializing empty structs as tuples when reading them.
- This also means that we can now store and persist `TUPLE`s in columns, which was not supported for unnamed structs.
- Additionally, tuples and structs are trivially implicitly castable to each other, so any (future) function signature changes should not affect user code.

### C-API Compatibility

Similarly to how we deal with tuples in storage, we do not add a new `DUCKDB_LOGICAL_TYPE_TUPLE`, but transparently represent tuples as unnamed structs, and vice-versa. So any existing C-API code should just work and not have to make any changes.

### Code Changes

I've made most of the internal struct type helper (e.g. `StructType::GetChildren`) also work on tuples so that most of the execution engine and extension code will remain oblivious to this change. However, I've had to keep the `StructType::IsUnnamed` as we somehow create (partially?) empty/unnamed structs in various places when dealing with `VARIANT`s. This should be investigated and resolved in a follow-up PR - ideally we should enforce the invariant that structs cannot be unnamed all over the system and make the need for this function redundant. Or maybe we decide that  an empty string is a valid struct member, given that structs can't have duplicate members anyway. Either way we need to work it out.

There are also a bunch of `struct_` functions that could be simplified and/or overloaded for TUPLEs. I've tried to keep the changes minimal in this PR, but we should probably make another pass over these in the future as well.


lnkuiper:
Codex found an issue in the CAPI:
```C++
  TEST_CASE("C API unnamed struct cannot bypass tuple storage version gate", "[capi]") {
  	CAPITester tester;
  	REQUIRE(tester.OpenDatabase(nullptr));

  	auto path = TestCreatePath("capi_unnamed_struct_v15.db");
  	std::remove(path.c_str());
  	std::remove((path + ".wal").c_str());

  	auto attach = "ATTACH '" + path + "' AS v15 (STORAGE_VERSION 'v1.5.0')";
  	auto attach_result = tester.Query(attach);
  	REQUIRE_NO_FAIL(*attach_result);

  	auto child_type = duckdb_create_logical_type(DUCKDB_TYPE_INTEGER);
  	duckdb_logical_type child_types[] = {child_type};
  	const char *child_names[] = {""};

  	auto struct_type = duckdb_create_struct_type(child_types, child_names, 1);
  	REQUIRE(struct_type);
  	REQUIRE(duckdb_get_type_id(struct_type) == DUCKDB_TYPE_STRUCT);

  	auto child_value = duckdb_create_int32(42);
  	duckdb_value child_values[] = {child_value};

  	auto struct_value = duckdb_create_struct_value(struct_type, child_values);
  	REQUIRE(struct_value);
  	REQUIRE(duckdb_get_type_id(duckdb_get_value_type(struct_value)) == DUCKDB_TYPE_STRUCT);

  	duckdb_prepared_statement stmt;
  	REQUIRE(duckdb_prepare(tester.connection, "CREATE TABLE v15.t AS SELECT ? AS x", &stmt) == DuckDBSuccess);
  	REQUIRE(duckdb_bind_value(stmt, 1, struct_value) == DuckDBSuccess);

  	duckdb_result result;
  	REQUIRE(duckdb_execute_prepared(stmt, &result) == DuckDBError);
  	REQUIRE(StringUtil::Contains(string(duckdb_result_error(&result)),
  	                             "TUPLE columns are not supported in storage versions prior to v2.0.0"));

  	duckdb_destroy_result(&result);
  	duckdb_destroy_prepare(&stmt);
  	duckdb_destroy_value(&struct_value);
  	duckdb_destroy_value(&child_value);
  	duckdb_destroy_logical_type(&struct_type);
  	duckdb_destroy_logical_type(&child_type);
  }
```
Other than that, looks good. I think `row_to_json` might need changing after this, but that doesn't have to happen in this PR.

Maxxen:
@lnkuiper Thanks for the feedback! I've added the test case, made `duckdb_create_struct_type` actually create TUPLE types internally when given a child list that would qualify as an empty struct, but also added a check for actual empty structs in the storage just-in-case.

lnkuiper:
Failing CI is unrelated. I've reproduced locally and it's a spurious issue related to one of my PRs from last week. Working on a fix

lnkuiper:
The PR got pulled from the merge queue last second because a merge conflict

lnkuiper:
Just a formatting issue now

PR #23466, opened 2026-06-25
Add a JSON API and use that instead of calling `yyjson` directly

Mytherin:
This PR adds the `JSONDocument` C++ API. This is a wrapper over various `yyjson` methods that is more ergonomic and also means the system is less explicitly tied to `yyjson` as a JSON library (in case we ever want to make this pluggable in the future). This also replaces the previous `ComplexJSON` API which, ironically enough given the name, a limited C++ wrapper that supported a subset of JSON functionality.

API comparison, reading/writing a simple JSON:

```json
{"key1": "value", "key2": 42}
```
#### yyjson

```cpp
auto doc = yyjson_mut_doc_new(nullptr);
auto root = yyjson_mut_obj(doc);
yyjson_mut_doc_set_root(doc, root);
yyjson_mut_obj_add_strcpy(doc, root, "key1", "value");
yyjson_mut_obj_add_int(doc, root, "key2", 42);

size_t len;
char *data = yyjson_mut_write(doc, 0, &len);
string result(data, len);
free(data);
yyjson_mut_doc_free(doc);                    
```

#### JSONDocument

```cpp
JSONWriter writer;
auto root = writer.CreateObject();
root.AddString("key1", "value");
root.Add("key2", writer.CreateSignedInteger(42));
writer.SetRoot(root);

string result = writer.ToString();
```

#### yyjson

```cpp
auto doc = yyjson_read(result.c_str(), result.size(), 0);
auto root = yyjson_doc_get_root(doc);
string key1 = yyjson_get_str(yyjson_obj_get(root, "key1"));
int64_t key2 = yyjson_get_int(yyjson_obj_get(root, "key2"));
yyjson_doc_free(doc);               
```

#### JSONDocument

```cpp
// throws if an error is encountered, there's also TryParse which returns an error object instead
auto doc = JSONDocument::Parse(result.c_str(), result.size());
auto root = doc->GetRoot();
string key1 = root.GetMember("key1").GetString();
int64_t key2 = root.GetMember("key2").GetSignedInteger();
```


Tishj:
RAII around doc lifetime and also stringify result is great to put into core, I've seen every lakehouse extension write their own version of these by now 😅

src/common/json_document.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+JSONMutableValue::JSONMutableValue(yyjson_mut_doc *doc_p, yyjson_mut_val *val_p) : doc(doc_p), val(val_p) {
+}
+
+bool JSONMutableValue::IsValid() const {
+	return val != nullptr;
+}
+
+void JSONMutableValue::Add(const string &key, JSONMutableValue value) {
+	auto key_val = yyjson_mut_strncpy(doc, key.c_str(), key.size());
```
Tishj, line 242: Yes, 1000x times yes 👍 
It's such a foot gun to work with `yyjson_mut_obj_add`, because the `key` string doesn't get copied

PR #23468, opened 2026-06-25
Attached database invalidation scope 

artjomPlaunov:
Add invalidation granularity for attached databases. This is currently only used for checkpointing. Previously, non-fatal exceptions during checkpointing were swallowed. This may have worked "fine" in a lot of cases, but this could corrupt in memory state (for example, the wal path being set to the concurrent checkpoint wal, and never being reset). The more desirable solution is probably to make checkpoints themselves recoverable, but for now we add invalidation granularity for attached databases, so if any exception occurs during checkpoint we can now just detach and re-attach that database (added a non-fatal checkpoint abort).

This part I wasn't entirely sure about: I currently have it so that if this happens on the initial database, we still do instance wide invalidation. 

A note on modifications of existing tests:

I split the existing ```concurrent_checkpoint_abort``` test into two tests. The reason is that the test was depending on instance wide invalidation. The first part of the test (checking that the data is correct), relies on instance wide invalidation so we test it on the initial database. The second part, triggering failures during wal replay, relies on attaching a database so I split that off to the scoped version and changed the invariant being tested at the end.

Also, I got rid of the ```block_file_double_decrement``` test, however this is just a stronger version of another test that I kept by refactoring to use an ATTACH, from #22394. The reason I got rid of it is because it implicitly relies on previous behavior where the auto checkpoint was OOMing and was not being reported. I tried to refactor it to use an attach, as well as to disable autocheckpointing, but then it led to an OOM in another scenario. This test is an example of behavior that was depending on a checkpoint exception being fine in practice, but we don't want to rely on this behavior. 

It's worth noting it is possible for a transaction to successfully commit but the auto-checkpoint to fail. With this PR, in this scenario the attached database gets invalidated (unless it is the initial database, in which case the entire instance is invalidated). On the next operation, the error that causes the failure will be printed as part of the invalidation message. We may want to extend this and have the error propagate directly upwards, however this can be semantically weird because the commit is successful. Decided not to make it part of this PR to separate out the design considerations, to focus on attached database invalidation here.


src/transaction/meta_transaction.cpp
```diff
@@ -54,6 +54,9 @@ optional_ptr<Transaction> MetaTransaction::TryGetTransaction(AttachedDatabase &d
 }
 
 Transaction &MetaTransaction::GetTransaction(AttachedDatabase &db) {
+	if (ValidChecker::IsInvalidated(db)) {
+		throw IOException("%s", ValidChecker::InvalidatedMessage(db));
```
Mytherin, line 58: Maybe we should also check this in Commit / Rollback?

Mytherin (commented):
Thanks! Looks great - one comment

src/transaction/meta_transaction.cpp
artjomPlaunov: @Mytherin let me know if this looks correct, added checks in commit and rollback for MetaTransaction, also added two tests for this (so the entire commit will fail, although as we discussed it could still be possible individual commits succeed before reaching the invalidated db)

Mytherin:
Thanks!

a27981731-rgb:
Don’t drop that dun dun dun


Sent from Yahoo Mail for iPhone


On Saturday, June 27, 2026, 1:18 AM, Mark ***@***.***> wrote:



Merged #23468 into main.
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

PR #23469, opened 2026-06-25
Improve hamming function test coverage

hsusul:
## Description

This pr tightens the sqllogictest coverage for `mismatches`/`hamming` error cases.

The existing test cases checked that invalid inputs errored, but left the expected error output blank. This updates those cases to assert the current `Invalid Input Error` messages thrown by the implementation:
- unequal input lengths
- empty string inputs
It also fixes a small typo in the `hamming` function description: “between to strings” -> “between two strings”.

No behavior changes.

## Tests

- [x] `git diff --check`
- [ ] `build/debug/test/run test/sql/function/string/test_mismatches.test`

I could not run the focused sqllogictest locally because this checkout is missing `build/debug/test/unittest`; the checked-in test runner delegates to that binary.

### DIFF
--- a/extension/core_functions/scalar/string/functions.json
+++ b/extension/core_functions/scalar/string/functions.json
@@ -97,7 +97,7 @@
     {
         "name": "hamming",
         "parameters": "s1,s2",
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - but this is not really useful. Please have a look at the contributing guidelines for how to build so you can test changes locally as well.

PR #23472, opened 2026-06-25
[ASYNC I/O] Refactor Multifile Scan for Read Ahead

pdet:
This PR refactors the multi-file scan further, so we will be able to support the upcoming async-I/O read-ahead. 
The main idea is that the scan now has jobs that can be decoupled from the decoding thread.


Mytherin:
Thanks!

PR #23473, opened 2026-06-25
Introduce debug filesystem

dentiny:
Laurens [recently added](https://github.com/duckdb/duckdb/pull/23353) injected latency to local filesystem for debugging and troubleshooting purpose.
I want to extend the scope for the feature a bit, so the latency could be applied to all possible filesystem.
I actually developed and maintained a [community extension](https://duckdb.org/community_extensions/extensions/latency_injection_fs), which I found useful to (1) surface data race issues; (2) for local performance benchmark.

Design goals:
- The latency is by default off, with happy path (production use case) suffer no performance impact under good branch prediction; perf is not a big concern for debug use case (we inject latency anyway), so lock is used for code clarity
- Different with the initial PR, which implements a fixed latency, I use the log-normal distribution to simulate IO latency
- Certain debugging feature should be applied to other filesystems with no extra code needed (in detail, we shouldn't add another setting for httpfs), so `DebugFileSystem` is written in the same way as we do for `OpenerFileSystem`
  + Discussion thread here: https://github.com/duckdb/duckdb/pull/23353#issuecomment-4788030841

Future work:
- Only latency is introduced in this PR, I'm thinking there're a few useful features we could introduce if necessary, for example throughput-based throttling, error injection, etc

I have ran CI on my own fork: https://github.com/dentiny/duckdb/pull/114


Mytherin:
Good idea to abstract this away - maybe this is something we can integrate into `unittest` instead of shipping in the main system, though? e.g. we could have a statically linked `unittest` extension similar to how the shell has one (`tools/shell/shell_extension.cpp`). I think there might be more stuff we can expose there.

dentiny:
> e.g. we could have a statically linked unittest extension similar to how the shell has one

Usually the integration point for DuckDB filesystems are virtual filesystem, where devs could wrap already registered filesystems with their own fs wrapper, but for `DebugFileSystem` (or whatever you call it) it's supposed to wrap virtual filesystem -- maybe we could wrap the filesystem instance in `DBConfig` (which is for now `OpenerFileSystem`).

dentiny:
>  maybe this is something we can integrate into unittest instead of shipping in the main system, though

Yeah that makes sense. A few changes in the latest commit 
- Move the whole debug filesystem into a `debug_fs` in-tree extension, which by default doesn't build, link and load
- Allow users to set random seed for the random engine, so they could reproduce the pseudo-random behavior if any

dentiny:
Hi @lnkuiper , I'm wondering if you might have some interest to take a look when you have some time? Thank you!

lnkuiper:
Looks good! I thought about it for a while and decided to implement Mark's feedback of putting in under the test dir, we can always manually load the unsigned extension for benchmarking purposes

lnkuiper:
Thanks! I've added it to the queue :)

PR #23476, opened 2026-06-26
Make LogicalGet SupportSerialization check more robust

kryonix:
For custom (C/C++) table functions, the common subplan eliminator could flag a subtree as valid for the optimization, when in reality it was not. Reason for that: `LogicalGet::SupportSerialization` did not check if the function actually has serialization callbacks. However, just checking that is also not enough, since some internal functions like `range` also do not have serialization callbacks, but are still safe, because of rebinding support.


kryonix:
The CI failure is unrelated.

PR #23482, opened 2026-06-26
Improve cache memory estimation on filepath cache object

dentiny:
Followup PR for https://github.com/duckdb/duckdb/pull/23052#discussion_r3355054926

Currently for `ExternalFileCacheObjectCacheEntry`, we only consider filepath but not the `CacheBlock` entries, which vary a lot based on the file size and block size.
In the PR, for simplicity I do an over-estimation on overall memory consumption, which assumes the whole file is cached.

Alternatives considered:
- Implement the object cache size adjustment, but that requires much larger code change from LRU cache to object cache

### DIFF
--- a/src/storage/external_file_cache/external_file_cache.cpp
+++ b/src/storage/external_file_cache/external_file_cache.cpp
@@ -34,7 +34,18 @@ public:
 	}
 
 	optional_idx GetEstimatedCacheMemory() const override {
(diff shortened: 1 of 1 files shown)


lnkuiper:
Great! Overestimating is fine and safe, much easier than implementing cache size adjustment. Thanks!

PR #23483, opened 2026-06-26
Enabled `ALP` and `ALP_RD` for storage version v1.5.0 and up when using smaller block sizes

sebastiaan-dev:
Resolves https://github.com/duckdblabs/duckdb-internal/issues/9727. 

In addition we moved the eligibility check to the `InitAnalyze` phase since the result is based on the database configuration and not on the incoming data.


Mytherin:
Thanks - LGTM. Is this something we want for v1.5, or should we retarget this to main?

PR #23484, opened 2026-06-26
Fix partial join filter pushdown through projections

ted-wq-x:
## Summary
Fix join filter pushdown so projection and set operation paths can push down the join keys that remain valid, instead of abandoning the whole target when another join key cannot be rewritten.

This allows queries such as joins against `UNION ALL` branches with constant label columns plus real key columns to still push dynamic filters on the real keys.

## Changes
- Allow partial join filter pushdown through projections and set operations.
- Track each pushed column’s original join condition index with `join_filter_idx`.
- Update hash join filter finalization to use the tracked join condition index instead of assuming pushed columns are positionally aligned with `join_condition`.
- Add regression coverage for:
  - one pushable key plus one unpushable projected key;
  - one pushable key plus multiple unpushable projected keys;
  - multiple pushable keys plus multiple unpushable projected keys.

## Testing
- `cmake --build /Users/mac/Documents/CLionProjects/duckdb/build/codex_release --target unittest -j10`
- `./build/codex_release/test/unittest test/optimizer/pushdown/join_filter_pushdown.test`
- `./build/codex_release/test/unittest`

### DIFF
--- a/src/execution/operator/join/physical_hash_join.cpp
+++ b/src/execution/operator/join/physical_hash_join.cpp
@@ -1279,8 +1279,8 @@ bool JoinFilterPushdownInfo::CanUseInFilter(const ClientContext &context, option
 	return ht && ht->Count() > 1 && ht->Count() <= dynamic_or_filter_threshold && cmp == ExpressionType::COMPARE_EQUAL;
 }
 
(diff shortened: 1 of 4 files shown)


Mytherin:
Thanks!

PR #23490, opened 2026-06-27
Make `QualifiedName` immutable and instead override the entire qualified name when modified

Mytherin:
Currently we do a lot of in-place modification of the `QualifiedName` - this PR moves away from that and instead just generates a new `QualifiedName` when doing these modifications, e.g. by calling `QualifiedName::WithName` or `QualifiedName::WithQualification`. 

In general we should move away as much as possible from setting individual components of the `QualifiedName`, but this at least cleans this up a bit and makes it less error-prone / messy.


Tishj:
Euhm, is it intended that these patches start with `zzzzzzz` making it (near) impossible to add more patches on top ?

PR #23501, opened 2026-06-28
Optimize aggregate fetch group lookup

yashwantraju:
## Description

This optimizes aggregate result fetching by separating read-only group lookup from the insert-oriented `FindOrCreateGroups` path.

`FetchAggregates` only needs to resolve addresses for groups that already exist in the aggregate hash table before finalizing aggregate states. Previously it reused `FindOrCreateGroups`, which also contains logic for creating missing groups and tracking newly inserted entries. This patch adds a private `FindGroupsInternal` helper for the fetch path, avoiding that insertion-specific work during result materialization.

## Changes

- Added `GroupedAggregateHashTable::FindGroupsInternal`.
- Updated `FetchAggregates` to use the read-only lookup path.
- Kept `FindOrCreateGroups` unchanged for sink/update paths that still need insertion behavior.

## Expected Benefit

This reduces unnecessary work while scanning/finalizing aggregate hash table results, especially for queries with many groups and cheap aggregate functions.

### DIFF
--- a/src/execution/aggregate_hashtable.cpp
+++ b/src/execution/aggregate_hashtable.cpp
@@ -691,9 +691,8 @@ void GroupedAggregateHashTable::FetchAggregates(DataChunk &groups, DataChunk &re
 		return;
 	}
 
-	// find the groups associated with the addresses
-	// FIXME: this should not use the FindOrCreateGroups, creating them is unnecessary
-	FindOrCreateGroups(groups, state.addresses);
+	groups.Hash(state.hashes);
+	FindGroupsInternal(groups, state.hashes, state.addresses);
 	// now fetch the aggregates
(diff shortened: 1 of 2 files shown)


Tishj:
I'm very far from the authority on this, but still, some thoughts:
It feels like there's too much duplication between the two paths. I suspect they can be unified under a shared internal method with a template parameter?

Without a benchmark that proofs the gains of this PR (and a comparison of the results of said benchmark(s) between `main` and this PR), I doubt we'd even consider accepting the risk added by the PR (extra complexity, duplicated code that could silently get stale / diverge / break, new and exciting possibilities for bugs, added maintenance burden)

lnkuiper:
Agreed. Implementing this with minimal code duplication, perhaps by templating, would be preferrable.

PR #23506, opened 2026-06-29
Add eager aggregation to PartialAggregatePushdown

ywelsch:
This PR extends the `PartialAggregatePushdown` optimizer to perform eager aggregation — pushing a partial `GROUP BY` below an inner join so the join processes and replicates far fewer rows, with a compensating aggregate above the join that recovers the exact result. It follows the ideas in Yan & Larson ("Eager Aggregation and Lazy Aggregation", VLDB 1995).

It adds two strategies:
- a one-sided pushdown that exports an aggregate's intermediate state down one join input and merges it back above the join via `combine`/`finalize` (so it works for any state-exportable aggregate), and
- a two-sided "double-eager" pushdown for fact×fact joins that pre-aggregates both inputs by the join key and reconstructs each measure from the per-key partials and row counts (for the distributive aggregates `sum`/`count`/`min`/`max`, with `avg` handled as `sum`/`count`).

The rewrite is cost-gated by cardinality heuristics so it only fires when it pays off — fan-out joins and joins where the aggregate side collapses — and stays off for selective joins, leaving standard star-schema workloads unaffected. It is correctness-preserving: floating-point sums are reassociated only where grouped aggregation is already order-dependent (the deterministic `kahan_sum` is never pushed), and exact-typed decompositions that could overflow are excluded. On a self-amplifying fact×fact join this turns a quadratic cross-product into roughly linear work (orders-of-magnitude speedups), while on TPC-DS at SF100 it produces several wins with no regressions and identical results.

# Benchmarks:

## Self-amplifying fact×fact join

`fact_a` (150k rows) ⋈ `dim` (unique-key dimension, 2000 rows) ⋈ `fact_b` (150k rows),
Zipfian keys concentrated on the same hot keys, grouped by a dimension attribute,
`count(*)` + `avg`. The naive plan's join cross-product is **443,270,469** rows.

| metric | before | after | ratio |
|:-------|-------:|------:|------:|
| query time (min of 5 warm) | 1.866 s | 0.003 s | **~622× faster** |

- **Plan:** before, the join builds the full 443M-row cross-product; after, each fact
  is pre-aggregated by key and the result reconstructed above the join.
- **Correctness:** `count(*)` is bit-exact; `avg` (DOUBLE) matches to ~9 significant
  digits (~3e-9 relative) — the expected, benign FP reassociation of a precision-only
  sum below the join.
- The gap widens with fact size (the naive cross-product grows ~quadratically in
  rows-per-key while the eager-aggregated plan stays ~linear); 150k rows/fact keeps the
  baseline runnable. At 3 ms the optimized side is near the timer floor, so the exact
  multiple is noisy, but the order of magnitude (hundreds of ×) is robust.

## TPCH

no changes

## TPC-DS SF100 (19 plan-changed queries)

| Query | base (s) | head (s) | ratio | result | note |
|------:|---------:|---------:|------:|:------:|:-----|
| 4  | 2.2200 | 1.1460 | **1.94×** | same | win |
| 74 | 1.1750 | 0.7350 | **1.60×** | same | win |
| 36 | 0.3410 | 0.2420 | **1.41×** | same | win |
| 70 | 0.3140 | 0.2780 | 1.13× | same | marginal |
| 21 | 0.0220 | 0.0210 | 1.05× | same | noise (<50 ms) |
| 23 | 5.0020 | 4.7880 | 1.04× | same | neutral |
| 26 | 0.0500 | 0.0480 | 1.04× | same | noise (<50 ms) |
| 14 | 2.5690 | 2.5230 | 1.02× | same | neutral |
| 93 | 0.4120 | 0.4090 | 1.01× | same | neutral |
| 85 | 0.1120 | 0.1120 | 1.00× | same | neutral |
| 62 | 0.1200 | 0.1220 | 0.98× | same | neutral |
| 7  | 0.2400 | 0.2470 | 0.97× | same | neutral |
| 5  | 0.4210 | 0.4400 | 0.96× | same | neutral |
| 98 | 0.2420 | 0.2550 | 0.95× | same | neutral |
| 99 | 0.1610 | 0.1690 | 0.95× | same | neutral |
| 83 | 0.0330 | 0.0350 | 0.94× | same | noise (<50 ms) |
| 77 | 0.2380 | 0.2600 | 0.92× | same | neutral |
| 12 | 0.0340 | 0.0420 | 0.81× | same | noise (<50 ms) |
| 20 | 0.0200 | 0.0250 | 0.80× | same | noise (<50 ms) |

**3 clear wins** (q4 1.94×, q74 1.60×, q36 1.41×), one marginal (q70), the rest
neutral, **zero regressions, zero result differences**. The sub-0.05 s queries
(q12/20/21/83) show ratios to 0.80× but those are 4–8 ms absolute deltas below the
timing noise floor — not real.


lnkuiper:
Thanks for the PR! I think this is really interesting, but needs some more work. I'll add some commits on top

lnkuiper:
Besides doing some cleanup/nitpicks, I've implemented the `combine_aggr(state, n)` aggregate function that combines an aggregate state with itself `n` times. It has a fast path implementation for `min`/`max`/`sum`/etc. (the functions that were white-listed in the optimiser before) to do this in constant time. For aggregate functions that don't have a fast path, we can do this in logarithmic time.

I've lifted two restrictions:

1. It should now work for all aggregate states that can be exported using the function described above. This includes the `kahan` aggregates.
2. The one-sided optimisation can be pushed down the non-nullable side of `LEFT`/`RIGHT` joins.

This should hopefully make the optimiser more useful. I think this is looking pretty close to ready now, but CI might take some convincing. We'll see.

PR #20224, opened 2025-12-16
Correctly handle default expressions in WAL when doing `ALTER TABLE ... ADD COLUMN ... DEFAULT ... `

yan-alex:
[5055](https://github.com/duckdblabs/duckdb-internal/issues/5055)
# The issue
When running a statement such as 
```sql
statement ok
ALTER TABLE t ADD COLUMN d TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP;
```

DuckDB would write an instruction to the WAL to execute `CURRENT_TIMESTAMP` again, so on replay the resulting column's values would be inconsistent with the original data.

# The fix
During TransformAlter, we now check if the column's DefaultValue ExpressionClass is CONSTANT. If not, we do a workaround that consists of the following statements:

1. ALTER TABLE t ADD COLUMN u <type> DEFAULT NULL;
2. UPDATE t SET u = <expression>;
3. ALTER TABLE t ALTER u SET DEFAULT <expression>;

By now doing an UPDATE, we force materialization of the values, which makes WAL replays consistent.

## Considerations
The functions requiring this workaround are those **not marked** `FunctionStability::CONSISTENT`, a subset of the functions affected by these proposed changes. This limitation arises because our logic runs during the transform step,  at which point we do not yet have a bound expression to evaluate stability. (binding happens after transform)

PR #23507, opened 2026-06-29
[v1.5] Quickfix `ALTER TABLE ADD COLUMN IF NOT EXISTS ... DEFAULT` regression

yan-alex:
https://github.com/duckdblabs/duckdb-internal/issues/9721
## The problem
When ADD COLUMN IF NOT EXISTS targets a column that already exists and the clause carries a DEFAULT, DuckDB 1.5.x does not treat the statement as a no-op: it re-applies the default expression to every existing row, silently destroying the column's current data.
## The fix
Don't do the 3 multi-statement rewrite workaround from https://github.com/duckdb/duckdb/pull/20224 when there is a `IF NOT EXISTS`
This is just a v1.5-variegata quickfix. We can tackle this properly later on for v2.0.

### DIFF
--- a/src/parser/transform/statement/transform_alter_table.cpp
+++ b/src/parser/transform/statement/transform_alter_table.cpp
@@ -127,19 +127,26 @@ unique_ptr<SQLStatement> Transformer::TransformAlter(duckdb_libpgquery::PGAlterT
 			}
 			column_entry.SetName(column_names.back());
 			if (column_names.size() == 1) {
-				// ADD COLUMN
-				if (!column_entry.HasDefaultValue() ||
-				    column_entry.DefaultValue().GetExpressionClass() == ExpressionClass::CONSTANT) {
+				if (command->missing_ok) {
+					// SQL query has `IF NOT EXISTS`, so we fall back to the old path to avoid a regression due to the
+					// multistatement rewrite.
+					// This is just a v1.5-variegata quickfix. We can tackle this properly later on for v2.0.
 					result->info =
 					    make_uniq<AddColumnInfo>(std::move(data), std::move(column_entry), command->missing_ok);
-					break;
+				} else {
+					// ADD COLUMN
+					if (!column_entry.HasDefaultValue() ||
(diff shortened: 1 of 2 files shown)


lnkuiper (changes requested):
The materializing multi-statement rewrite is now skipped, which fixes the linked issue, but regresses a WAL replay consistency. Repro:
```sql
# name: test/temp.test
# group: [temp]

statement ok
ATTACH '__TEST_DIR__/wal_replay_if_not_exists.db';

statement ok
USE wal_replay_if_not_exists;

statement ok

statement ok
SET wal_autocheckpoint='1TB'

statement ok
CREATE TABLE t AS SELECT range::INT AS id FROM range(10);

statement ok
ALTER TABLE t ADD COLUMN IF NOT EXISTS r DOUBLE DEFAULT RANDOM();

statement ok
CREATE TABLE original AS SELECT id, r FROM t;

statement ok
USE memory;

statement ok
DETACH wal_replay_if_not_exists;

statement ok
ATTACH '__TEST_DIR__/wal_replay_if_not_exists.db';

statement ok
USE wal_replay_if_not_exists;

query I
SELECT COUNT(r) != 0 FROM t;
----
true

query I sort r1
SELECT r FROM t;
----

query I sort r1
SELECT r FROM original;
----
```

lnkuiper:
Is this PR https://github.com/duckdb/duckdb/pull/23522 related?

yan-alex:
> Is this PR #23522 related?

Yes it addresses the same problem. That one targets main though, which we want to tackle differently I believe.

yan-alex:
@Mytherin I implemented [your suggestion on the internal issue ](https://github.com/duckdblabs/duckdb-internal/issues/9721#issuecomment-4727868667) which Laurens wans't aware of

lnkuiper (approved):
I think the issue that this fixes is more important than the issue it creates so we should merge it as-is, and come up with a proper solution for v2.0.0

PR #23510, opened 2026-06-29
autocheckpoint error propagation

artjomPlaunov:
Follow up to https://github.com/duckdb/duckdb/pull/23468 

Add error propagation for the specific "successful commit, autocheckpoint failure" scenario; before this was only surfacing if trying to use the invalidated database. 

I think there are several alternatives here, I tried this way first to avoid having to change any signatures/logic for CommitTransaction. Alternatives could be creating a new exception type, changing the return of CommitTransaction (both duck_transaction_manager, as well as the virtual interface in transaction_manager.hpp), or modifying ErrorData in some way. Those seemed more involved, and this is a very specific scenario so I tried out just setting it in the transaction context, which keeps existing commit/rollback logic untouched, and allows us to check if an error is present in TransactionContext::Commit(). Happy to try out any other approaches though :-)


src/transaction/duck_transaction_manager.cpp
```diff
@@ -18,9 +18,22 @@
...
 namespace duckdb {
 
+static ErrorData BuildAutocheckpointError(AttachedDatabase &db, const std::exception &ex) {
+	ErrorData original(ex);
+	string recovery = db.IsInitialDatabase()
+	                      ? "Reopen the database instance to recover (committed data is restored from the WAL)."
+	                      : "Detach and reattach the database to recover (committed data is restored from the WAL).";
+	string msg = StringUtil::Format("Transaction COMMIT succeeded and is durable, but the autocheckpoint failed. "
+	                                "%s %s\n\nStack Trace:\n%s",
+	                                recovery, original.RawMessage(), StackTrace::GetStackTrace());
```
artjomPlaunov: actually, I don't think this stack trace makes sense going to remove it.

lnkuiper:
I asked Codex to try and break this PR and it found this test:
```sql
# PR 23510 repro: NO_WAL_WRITES commits are not WAL-durable.
statement ok
PRAGMA disable_checkpoint_on_shutdown

statement ok
SET auto_checkpoint_skip_wal_threshold = 9999999999

statement ok
SET wal_autocheckpoint = '1TB'

statement ok
ATTACH '{TEST_DIR}/pr23510_no_wal_autockpt.db' AS nw (RECOVERY_MODE NO_WAL_WRITES)

statement ok
CREATE TABLE nw.tbl(i INTEGER)

statement ok
CHECKPOINT nw

statement ok
SET wal_autocheckpoint = '1KB'

statement ok
PRAGMA debug_checkpoint_abort = 'before_header_non_fatal'

statement error
INSERT INTO nw.tbl SELECT range::INTEGER FROM range(10000)
----
<REGEX>:Transaction Error: Failed to commit: Checkpoint failed for database "nw".*
```

With this suggestion:
> The fix is likely to distinguish `should_write_to_wal` in the catch: only
  use the “durable/restored from WAL” error when WAL was actually written; otherwise return/merge the checkpoint error as a normal commit failure.

artjomPlaunov:
> With this suggestion:
> 
> > The fix is likely to distinguish `should_write_to_wal` in the catch: only
> > use the “durable/restored from WAL” error when WAL was actually written; otherwise return/merge the checkpoint error as a normal commit failure.

ah nice, I think this may have actually been a latent bug (i.e. it was silent before, but putting in the incorrect "commit succeeded" message seems to have helped codex catch it :-) )

artjomPlaunov:
@lnkuiper thanks! added an explicit wal_written boolean (as well as the test)

artjomPlaunov:
also handled in memory databases correctly, and added a test for that case

lnkuiper:
Thanks!

Issue #5065, opened 2022-10-22
[Python][Arrow] Pushdown for offset/limit


changhiskhan:
Is there's a way to push down offset/limit to the data access layer?
We have a special pyarrow scanner subclass that knows how to read the exact byte range for offset/limit. So if there are no aggregates, it would be really efficient to push the offset/limit down as extra ScanOptions for arrow (esp if the projection columns are large blobs like image bytes or something).

PR #23520, opened 2026-06-30
Expose limit hints to C++ table functions

mrdrivingduck:
## Background

DuckDB table functions already receive several scan-level properties during initialization, such as projected columns and filters. This lets table functions avoid unnecessary work when the optimizer can prove that only part of the source needs to be read.

There is currently no equivalent way for a C++ table function to observe a query-level `LIMIT` at scan initialization time. This matters for extensions that read from external systems, remote APIs, lakehouse metadata, or paginated sources, where knowing a conservative upper bound can avoid opening extra files, planning extra splits, or issuing extra remote requests.

I noticed that similar needs have come up in a few DuckDB repositories. In the main DuckDB repository, [duckdb/duckdb#5065](https://github.com/duckdb/duckdb/issues/5065) asks about pushing offset/limit down to the data access layer for an Arrow scanner. In the Postgres extension, [duckdb/duckdb-postgres#233](https://github.com/duckdb/duckdb-postgres/issues/233) reported that `LIMIT` was not pushed down and caused the entire table to be copied, which was later addressed by [duckdb/duckdb-postgres#313](https://github.com/duckdb/duckdb-postgres/pull/313). In the Iceberg extension, [duckdb/duckdb-iceberg#55](https://github.com/duckdb/duckdb-iceberg/issues/55) also contains user reports around `LIMIT` not being pushed into `iceberg_scan`.

Those implementations are necessarily extension-specific today. This PR adds a small core-level hint so C++ table functions can observe a conservative limit bound during scan initialization, while DuckDB still keeps the real limit operator in the plan.

## Changes

- Add an advisory `limit` field to the C++ `TableFunctionInitInput`; the C table function API is unchanged.
- Propagate constant `LIMIT` and `LIMIT/OFFSET` bounds into simple table scans, including projection-above-scan cases.
- Keep the existing limit operators in the plan; the hint is for scan planning only and does not change correctness semantics.
- Keep propagation conservative: no filters, order/top-n, joins, aggregates, windows, table-in-out input, overflowed bounds, or other row-changing operators.
- Keep this separate from the existing small-limit projection rewrite, so larger limits can still be exposed as scan hints.

## Testing

Added a C++ table-function test that records the observed limit during global and local initialization, covering limited and unlimited scans plus conservative non-propagation cases.

### DIFF
--- a/src/execution/operator/scan/physical_table_scan.cpp
+++ b/src/execution/operator/scan/physical_table_scan.cpp
@@ -19,13 +19,14 @@ PhysicalTableScan::PhysicalTableScan(PhysicalPlan &physical_plan, vector<Logical
                                      vector<ColumnIndex> column_ids_p, vector<idx_t> projection_ids_p,
                                      vector<string> names_p, unique_ptr<TableFilterSet> table_filters_p,
                                      idx_t estimated_cardinality, ExtraOperatorInfo extra_info,
-                                     vector<Value> parameters_p, virtual_column_map_t virtual_columns_p)
+                                     vector<Value> parameters_p, virtual_column_map_t virtual_columns_p,
+                                     optional_idx limit_p)
     : PhysicalOperator(physical_plan, PhysicalOperatorType::TABLE_SCAN, std::move(types), estimated_cardinality),
 
       function(std::move(function_p)), bind_data(std::move(bind_data_p)), returned_types(std::move(returned_types_p)),
       column_ids(std::move(column_ids_p)), projection_ids(std::move(projection_ids_p)), names(std::move(names_p)),
       table_filters(std::move(table_filters_p)), extra_info(std::move(extra_info)), parameters(std::move(parameters_p)),
-      virtual_columns(std::move(virtual_columns_p)) {
+      virtual_columns(std::move(virtual_columns_p)), limit(limit_p) {
 }
 
 class TableScanGlobalSourceState : public GlobalSourceState {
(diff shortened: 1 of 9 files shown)


lnkuiper:
DuckDB stops the pipeline as soon as the `LIMIT` has been reached, which also works when the query plan is more complex, e.g., joins going into the `LIMIT`. The issues with `LIMIT` not being pushed down into other catalogs has been resolved generically in https://github.com/duckdb/duckdb/pull/22914 . I feel like that should cover the relevant cases. Is this needed?

mrdrivingduck:
@lnkuiper Thanks for pointing this out. I agree that the pipeline stop handles the execution-side early termination, and that #22914 covers full-query pushdown into remote catalogs.

The gap I am trying to address here is slightly different: C++ table functions that still execute as local DuckDB scans currently cannot observe a query-level limit during `init_global`/`init_local`. For some table functions, a significant part of the cost can happen before the first chunk is produced, e.g. planning splits, opening files, listing metadata, or preparing paginated remote reads. Pipeline termination can stop consumption once enough rows have flowed, but it does not give the scan initialization code a conservative bound to plan against.

#22914 helps when the query can be pushed into a remote catalog as SQL. This PR is for arbitrary C++ table functions, including table-function based scanners that are not remote catalogs and cannot use `RemotePushdown`.

The hint is intentionally advisory: the normal DuckDB limit operators stay in the plan and still own correctness. The table function can use the bound for scan planning if useful, but it is not required to enforce it.

That said, if you think this should be exposed through a different interface, or be scoped even more narrowly, I am happy to adjust.

PR #23522, opened 2026-06-30
Fix data overwrite on ALTER TABLE ADD COLUMN IF NOT EXISTS with DEFAULT

### LINKED ISSUE #23209: ALTER TABLE ADD COLUMN IF NOT EXISTS ... DEFAULT on an existing column silently overwrites all row values with the default (1.5.0 regression)
### What happens?

When `ADD COLUMN IF NOT EXISTS` targets a column that **already exists** and the clause carries a `DEFAULT`, DuckDB 1.5.x does not treat the statement as a no-op: it re-applies the default expression to **every existing row**, silently destroying the column's current data. No error, no warning.

This breaks the core contract of `IF NOT EXISTS` (idempotent re-run safety) and is a silent data-loss bug for any migration runner that re-applies idempotent `.sql` files on startup -- the second run wipes whatever the first run computed.

Behavior by version (same script, CPython 3.12.10, Windows 11 x64; also reproduced on an on-disk database with reconnect between runs):

| duckdb | result |
|--------|--------|
| 1.1.3, 1.2.2, 1.3.2, 1.4.2, 1.4.4 | correct -- statement is a no-op |
| 1.5.0, 1.5.1, 1.5.2, 1.5.3 | **bug -- default re-applied, data lost** |

Also reproduced on Ubuntu 24.04 x86_64 with CPython 3.12.3 and an isolated `duckdb==1.5.3` pip install.

Notes:
- Without a `DEFAULT` clause, `ADD COLUMN IF NOT EXISTS` on an existing column is still a correct no-op on 1.5.x -- the bug is specific to the DEFAULT path.
- Volatile defaults (`DEFAULT now()`) are also re-evaluated and overwrite the column, confirming the default-application step runs despite the column existing.
- Possibly related to the internal rewrite of `ADD COLUMN ... DEFAULT` introduced in 1.5.0 (cf. "Set query field for statements in ALTER TABLE ... ADD COLUMN ... DEFAULT ... workaround" in the release notes): the rewrite's apply-default-to-existing-rows step appears to run unconditionally, before/without the IF-NOT-EXISTS existence check short-circuiting it.

### To Reproduce

```python
import duckdb

con = duckdb.connect()
con.execute("CREATE TABLE t (i INTEGER)")
con.execute("INSERT INTO t VALUES (1), (2)")
con.execute("ALTER TABLE t ADD COLUMN IF NOT EXISTS b BOOLEAN DEFAULT false")
con.execute("UPDATE t SET b = true")
print(con.execute("SELECT list(b) FROM t").fetchone()[0])   # [True, True]

# column b already exists -> statement should be a no-op
con.execute("ALTER TABLE t ADD COLUMN IF NOT EXISTS b BOOLEAN DEFAULT false")
print(con.execute("SELECT list(b) FROM t").fetchone()[0])
# expected: [True, True]
# actual (1.5.0-1.5.3): [False, False]
```

Output on DuckDB 1.5.3:

```text
[True, True]
[False, False]
```

### OS:

Windows 11 x64; Ubuntu 24.04 x86_64

### DuckDB Version:

1.5.3 (regression introduced in 1.5.0; 1.4.4 and earlier correct)

### DuckDB Client:

Python (duckdb pip package)

### Hardware:

x86_64

### Full Name:

Luiz Piccini

### Affiliation:

Independent

- [x] Yes, I have

- [x] Yes, I have

preksha-kamalesh:
This fixes issue #23209 where ADD COLUMN IF NOT EXISTS would silently overwrite existing values in the column if it already existed.

The issue was caused by the parser rewriting the ALTER TABLE statement into a 3-statement MultiStatement (ADD COLUMN, UPDATE, ALTER DEFAULT). While the first statement respected the if_not_exists flag, the UPDATE and ALTER DEFAULT statements ran unconditionally.

This fix prevents the MultiStatement rewrite if the if_not_exists flag is present, allowing the single AlterStatement to handle the no-op correctly at runtime when the column exists.

Fixes #23209

### DIFF
--- a/src/parser/peg/transformer/transform_alter.cpp
+++ b/src/parser/peg/transformer/transform_alter.cpp
@@ -25,7 +25,8 @@ unique_ptr<SQLStatement> PEGTransformerFactory::TransformAlterStatement(PEGTrans
 	}
 	auto &add_column = alter_table.Cast<AddColumnInfo>();
 	if (!add_column.new_column.HasDefaultValue() ||
-	    add_column.new_column.DefaultValue().GetExpressionClass() == ExpressionClass::CONSTANT) {
+	    add_column.new_column.DefaultValue().GetExpressionClass() == ExpressionClass::CONSTANT ||
+	    add_column.if_column_not_exists) {
 		return std::move(result);
 	}
 	auto &column_entry = add_column.new_column;
--- a/test/sql/alter/add_col/test_add_col_if_not_exists_default.test
+++ b/test/sql/alter/add_col/test_add_col_if_not_exists_default.test
@@ -1,0 +1,69 @@
+# name: test/sql/alter/add_col/test_add_col_if_not_exists_default.test
+# description: Test that ADD COLUMN IF NOT EXISTS with a non-constant DEFAULT does not overwrite existing values
+# group: [add_col]
+
+# Create a table and add a column with a non-constant default
+statement ok
+CREATE TABLE test(id INTEGER);
+
+statement ok
+INSERT INTO test VALUES (1), (2), (3);
+
+statement ok
+ALTER TABLE test ADD COLUMN c INTEGER DEFAULT 1 + 1;
+
+# Verify the default was applied
+query II
+SELECT id, c FROM test ORDER BY id;
+----
+1	2
+2	2
+3	2
+
+# Update the column to a known value
+statement ok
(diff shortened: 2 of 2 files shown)


lnkuiper:
Thanks for the PR! https://github.com/duckdb/duckdb/pull/23507 fixes the same issue, so closing this now

PR #23526, opened 2026-06-30
Add ClientContext::BindStatement

evertlammerts:
Introduce `ClientContext::BindStatement(unique_ptr<SQLStatement>) -> StatementSignature` to enable inspecting a statement's result/parameter schema.

`BindStatement` runs `Planner::CreatePlan` and returns the signature of the statement, i.e. the input (parameter names and types) and the output (column names and types), and the `StatementProperties`. It's basically the same path as `ClientContext::Prepare` on a statement that has parameters and is not cacheable, with some divergences:
- It stops after binding, and skips the optimizer and physical plan.
- It doesn't cache its bound state, so it must always rebind. It also doesn't hold a pointer to the client context like a PreparedStatement does.
- It doesn't touch an active result, so it's safe to call while a client is streaming a result. It does run under the context lock, inside a transaction.


test/api/test_bind_statement.cpp
```diff
@@ -0,0 +1,64 @@
...
+using namespace duckdb;
+
+static StatementSignature BindOne(Connection &con, const string &sql) {
+	Parser parser(con.context->GetParserOptions());
+	parser.ParseQuery(sql);
+	REQUIRE(parser.statements.size() == 1);
+	return con.context->BindStatement(std::move(parser.statements[0]));
+}
+
+TEST_CASE("BindStatement reports the result and parameter schema", "[api][bind_statement]") {
```
pdet, line 21: Can we add a test to verify if the binding is transactionally safe?

e.g.,
```sql
CREATE TABLE t(a INTEGER)
BEGIN TRANSACTION
INSERT INTO t VALUES (1)
REQUIRE_THROWS(BindOne(con, "SELECT * FROM blaa"));
SELECT * FROM t
COMMIT
SELECT count(*) FROM t
```

src/main/client_context.cpp
```diff
@@ -881,6 +882,33 @@ unique_ptr<PreparedStatement> ClientContext::Prepare(unique_ptr<SQLStatement> st
...
+		    signature.types = planner.types;
+		    signature.properties = std::move(planner.properties);
+		    // Parameter types from the bound parameter map (as in PreparedStatementData::TryGetType).
+		    for (auto &entry : named_param_map) {
+			    LogicalType type(LogicalTypeId::UNKNOWN);
+			    auto it = planner.value_map.find(entry.first);
+			    if (it != planner.value_map.end()) {
+				    type = it->second->return_type.id() != LogicalTypeId::INVALID ? it->second->return_type
+				                                                                  : it->second->GetValue().type();
+			    }
```
pdet: Is it possible to have an entry not present in the value_map?

src/include/duckdb/main/client_context.hpp
```diff
@@ -64,6 +64,22 @@ struct PendingQueryParameters {
 	QueryParameters query_parameters;
 };
 
+//! A statement parameter: identifier ($1 -> "1"), binding index, and inferred type (UNKNOWN if not inferred).
```
pdet, line 67: Can we test the non inferred type? e.g.,  `SELECT $1 case asserting parameters[0].type == UNKNOWN`

src/include/duckdb/main/client_context.hpp
```diff
@@ -64,6 +64,22 @@ struct PendingQueryParameters {
...
+struct StatementParameter {
+	Identifier identifier;
+	idx_t index;
```
pdet, line 70: Can we add a test with multi parameters? `_$1, ...$3_

test/api/test_bind_statement.cpp
```diff
@@ -0,0 +1,64 @@
...
+	// A bind error throws.
+	REQUIRE_THROWS(BindOne(con, "SELECT * FROM no_such_table"));
+}
+
+TEST_CASE("BindStatement does not disturb a live streaming result", "[api][bind_statement]") {
+	DuckDB db(nullptr);
+	Connection con(db);
+
+	// Open a stream and consume one chunk, leaving it live mid-flight.
+	auto stream = con.SendQuery("SELECT i FROM range(5000) t(i)");
```
pdet, line 52: Do we need to decrease `streaming_buffer_size`?

pdet (commented):
Thanks for the PR, I had a couple of comments.

test/api/test_bind_statement.cpp
evertlammerts: Good find, it wasn't safe for an existing transaction. Fixed in `BindStatement` with a catch-all try catch, and then either 1. invalidating the transaction by re-throwing if `Exception::InvalidatesDatabase(error.Type())` or 2. re-throwing the exception _outside_ of the transaction to make sure it continues.

Tested [below](https://github.com/duckdb/duckdb/pull/23526/changes#diff-68349786eccb74bf4cb261164d3458beff03007d5136847fc0d507d80d88dc74R89).

test/api/test_bind_statement.cpp
evertlammerts: Now also tested below (but without `streaming_buffer_size`)

src/include/duckdb/main/client_context.hpp
evertlammerts: Added a bunch of [tests](https://github.com/duckdb/duckdb/pull/23526/changes#diff-68349786eccb74bf4cb261164d3458beff03007d5136847fc0d507d80d88dc74R172) for parameter type inference

src/include/duckdb/main/client_context.hpp
evertlammerts: Added a bunch of [tests](https://github.com/duckdb/duckdb/pull/23526/changes#diff-68349786eccb74bf4cb261164d3458beff03007d5136847fc0d507d80d88dc74R172) for parameter type inference

pdet:
Thanks!

PR #23527, opened 2026-06-30
Add 'connection' secret storage, for connection scoped contexts

carlopi:
This is a by-product of https://github.com/duckdb/duckdb-quack/pull/187, where I tried to figure out how secrets could be passed onto quack, and this is the duckdb-side change that would make possible to store client-scoped secrets in a quack server (since 1 client = 1 Connection). I think this could be of some general usage, together with current default secret storage that are instance-wide.

Allows
```sql
CREATE SECRET <name> IN connection (...);
```
to be stored in a `Connection`-level secret storage, and be used only by query on the same Connection. This allows some degree of isolation between connections, given secrets that normally would clash and resolve all together can be scoped down to single connection.
`Connection` secret storage stores only `TEMPORARY` secrets.

Initial use case was connected to `quack`, and this can be implemented purely extension side (give secret manager is extensible), but this is general it's (in my view) worth landing in `duckdb/duckdb`.

### DIFF
--- a/src/main/secret/secret_storage.cpp
+++ b/src/main/secret/secret_storage.cpp
@@ -336,9 +336,13 @@ constexpr const char *CONNECTION_SECRET_STATE_KEY = "connection_secret_storage";
 
 //! Per-connection secret container. Lives on the ClientContext's RegisteredStateManager, so it is destroyed exactly
 //! when the connection (its ClientContext) goes away - giving automatic, crash-robust cleanup with no cooperation.
+//! The secrets are held in a CatalogSet so that they follow the transaction they were created in: an aborted
+//! transaction undoes the CREATE/DROP, exactly like the global storages. No mutex is needed, CatalogSet locks itself.
 struct ConnectionSecretState : public ClientContextState {
-	mutex lock;
-	identifier_map_t<unique_ptr<SecretEntry>> secrets;
+	explicit ConnectionSecretState(Catalog &catalog) : secrets(make_uniq<CatalogSet>(catalog)) {
+	}
+
+	unique_ptr<CatalogSet> secrets;
 };
 
 //! Fetch the calling connection's secret container. With create=false returns nullptr when there is no context or no
@@ -349,7 +353,8 @@ optional_ptr<ConnectionSecretState> GetConnectionState(optional_ptr<CatalogTrans
 	}
 	auto &context = transaction->GetContext();
 	if (create) {
-		return context.registered_state->GetOrCreate<ConnectionSecretState>(CONNECTION_SECRET_STATE_KEY).get();
+		auto &catalog = Catalog::GetSystemCatalog(*context.db);
+		return context.registered_state->GetOrCreate<ConnectionSecretState>(CONNECTION_SECRET_STATE_KEY, catalog).get();
 	}
 	return context.registered_state->Get<ConnectionSecretState>(CONNECTION_SECRET_STATE_KEY).get();
 }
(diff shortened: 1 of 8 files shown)


lnkuiper:
I think creating connections is a great idea, maybe that should even be the default? I don't like the default of each connection polluting the db instance secret manager. What do you think @samansmink ?

carlopi:
I think one blocker to making this the default is that there are places where we rely on spawning a `Connection` for internal usage, say to execute queries mid table-function, those places would need to then inherit connection secret storage from the `Connection`.

This is also a weak point of this PR, that will not work for such cases.

I still think adding this (non-default) and then ironing out, and then consider the default would be the correct way forward.

src/main/secret/secret_storage.cpp
```diff
@@ -246,4 +252,122 @@ void LocalFileSecretStorage::RemoveSecret(const string &secret, OnEntryNotFound
...
+		default: // REPLACE_ON_CONFLICT
+			break;
+		}
+	}
+
+	auto entry = make_uniq<SecretEntry>(std::move(secret));
+	entry->persist_type = SecretPersistType::TEMPORARY;
+	entry->storage_mode = storage_name;
+	auto result = make_uniq<SecretEntry>(*entry);
+	state->secrets[name] = std::move(entry);
```
lnkuiper, line 433: This bypasses the catalog transaction machinery used by `TemporarySecretStorage`, so connection-scoped secrets are not rolled back. A secret created inside a rolled-back transaction remains visible afterward.

Minimal repro:

```sql
statement ok
BEGIN;

statement ok
CREATE SECRET txn_conn IN connection (TYPE http, SCOPE 'https://txn.example.com');

statement ok
ROLLBACK;

query I
SELECT count(*) FROM duckdb_secrets() WHERE name = 'txn_conn' AND storage = 'connection'
----
0
```
This currently returns `1`.

src/main/secret/secret_storage.cpp
```diff
@@ -246,4 +252,122 @@ void LocalFileSecretStorage::RemoveSecret(const string &secret, OnEntryNotFound
...
+	return make_uniq<SecretEntry>(*entry->second);
+}
+
+void ConnectionSecretStorage::DropSecretByName(const Identifier &name, OnEntryNotFound on_entry_not_found,
+                                               optional_ptr<CatalogTransaction> transaction) {
+	auto state = GetConnectionState(transaction, false);
+	idx_t erased = 0;
+	if (state) {
+		lock_guard<mutex> guard(state->lock);
+		erased = state->secrets.erase(name);
```
lnkuiper: Same transaction issue for drops: this erase is immediate, so rolling back a transaction that drops a connection-scoped secret does not restore it. `memory` secrets currently preserve the secret after `ROLLBACK` because they go through `CatalogSet`.

lnkuiper (changes requested):
Looks good, just needs to respect transactions:

Tishj:
I think this should perhaps even be a breaking change, as I mentioned in #23882 , `TEMPORARY` in the system currently means "connection-scoped" (though arguably, our settings that are connection-scoped are called LOCAL, so it's not very streamlined already), but the `TEMPORARY` secrets are actually **session** scoped.
Session scoped temporaries are what Spark refers to as `GLOBAL TEMPORARY`, maybe it makes sense to adopt that term?

carlopi:
Thanks for the review, I've updated to use `CatalogSet`, now properly transactional.

src/main/secret/secret_storage.cpp
carlopi: Thanks, handled in commit on top, and added test.

lnkuiper:
Could we avoid using a database-global `CatalogSet` with connection lifetime here? The set either needs to remain alive through the transaction-manager, or the connection state needs its own transaction commit/rollback mechanism.

Failing repro:
```C++
DuckDB database(nullptr);
Connection older(database);
older.Query("BEGIN");
older.Query("SELECT count(*) FROM duckdb_secrets()");
{
	Connection creator(database);
	creator.Query("CREATE SECRET s IN connection (TYPE http)");
} // destroys the connection-owned CatalogSet
```

Transaction manager might be the best option here?

carlopi:
I am not sure what's best here, I see some possible paths:
* Revert to a plain map on the connection state, no rollback/MVCC for connection secrets, but simple lifetimes
* Moving from `CatalogSet` to a simpler structure, using the fact that 1 connection = 1 transaction (that is MVCC is too much, this is a simpler problem)
* pin this on `DuckTransaction`, but that gets complex fast, and I am not sure I can pull it off (this is your suggestion, as I understood it)

I might go for 2nd option, but looking for feedback.

carlopi:
I went for option 2, abusing the fact that 1 connection  = 1 transaction, so MVCC is simple enough.

carlopi:
@lnkuiper: I think this now works properly, feel free to have a look.

I experimented with turning the default from `TEMPORARY` to `CONNECTION`, and all but tests explicitly testing cross-connection semantics works, so my proposal would be for merging this, and then considering swapping (this was also @Tishj suggestion, thanks).

@samansmink: do you happen to have any thoughts here? Should a connection by default connection scoped or not?

samansmink:
@carlopi I think default should remain `TEMPORARY`, its the semantics people expect now and I don't really see a clear argument why switching the default now is *that* valuable

src/main/secret/secret_storage.cpp
```diff
@@ -321,4 +328,170 @@ void LocalFileSecretStorage::RemoveSecret(const string &secret, OnEntryNotFound
...
+	// inserted under a garbage key (breaking later name-keyed lookups like DROP / GetSecretByName).
+	auto name = secret->GetName();
+
+	auto existing = state->secrets.find(name);
+	if (existing != state->secrets.end()) {
+		switch (on_conflict) {
+		case OnCreateConflict::ERROR_ON_CONFLICT:
+			throw InvalidInputException("Connection secret with name '%s' already exists", name.GetIdentifierName());
+		case OnCreateConflict::IGNORE_ON_CONFLICT:
+			return make_uniq<SecretEntry>(*existing->second);
```
lnkuiper: This returns the existing secret on `IGNORE_ON_CONFLICT`, while the other secret storage implementations return `nullptr`. We should follow the existing behaviour.

lnkuiper (commented):
One tiny comment then good to go:

lnkuiper (approved):
Thanks!

PR #23528, opened 2026-06-30
Push down scalar functions to file readers

### LINKED ISSUE #8310: Epic: Duckdb projection expression pushdown

myrrc:
This PR supports pushing down single-argument scalar functions in SELECT expression of form SELECT fn(T) into file readers.
Changes:

- New optimizer pass reusing most code from type pushdown
- Parquet support for pushdown: pushes octet_length() and strlen() to file reader. Both are not pushed for pushdown-extract columns, for multiple files, and for custom schema readers like Ducklake.
- projection_expression_pushdown now takes a column index in LOGICAL_GET instead of projection index. This allows getting a full ColumnIndex which in turn would allow supporting pushdown-extract columns in the future.

### DIFF
--- a/extension/parquet/parquet_multi_file_info.cpp
+++ b/extension/parquet/parquet_multi_file_info.cpp
@@ -336,20 +336,27 @@ static unique_ptr<FunctionData> ParquetScanDeserialize(Deserializer &deserialize
 	parquet_bind_data.projection_expressions = std::move(projection_expressions);
 
 	for (const auto &[idx, expr] : parquet_bind_data.projection_expressions) {
-		if (idx < inner_bind_data.columns.size()) {
-			inner_bind_data.columns[idx].type = expr.return_type;
-		}
-		if (idx < inner_bind_data.types.size()) {
-			inner_bind_data.types[idx] = expr.return_type;
-		}
-		if (auto &schema = inner_bind_data.reader_bind.schema; !schema.empty() && idx < schema.size()) {
-			schema[idx].type = expr.return_type;
+		const bool is_nested =
+		    idx < inner_bind_data.columns.size() && inner_bind_data.columns[idx].type.id() == LogicalTypeId::STRUCT;
+		if (!is_nested) {
+			if (idx < inner_bind_data.columns.size()) {
+				inner_bind_data.columns[idx].type = expr.return_type;
+			}
+			if (idx < inner_bind_data.types.size()) {
+				inner_bind_data.types[idx] = expr.return_type;
+			}
+			if (auto &schema = inner_bind_data.reader_bind.schema; !schema.empty() && idx < schema.size()) {
+				schema[idx].type = expr.return_type;
+			}
 		}
 		if (!inner_bind_data.initial_reader) {
 			continue;
 		}
 		auto &reader = inner_bind_data.initial_reader->Cast<ParquetReader>();
 		reader.projection_expressions[idx] = expr;
+		if (is_nested) {
+			continue;
+		}
 		if (idx < reader.columns.size()) {
 			reader.columns[idx].type = expr.return_type;
 		}
--- a/extension/parquet/parquet_multi_file_info.cpp
+++ b/extension/parquet/parquet_multi_file_info.cpp
@@ -299,6 +299,8 @@ static void ParquetScanSerialize(Serializer &serializer, const optional_ptr<Func
 	}
 	if (serializer.ShouldSerialize(StorageVersion::V2_0_0)) {
 		serializer.WriteProperty(105, "projection_expressions", parquet_data.projection_expressions);
+	} else if (!parquet_data.projection_expressions.empty()) {
+		throw SerializationException("Version too old to serialize projection_expressions but they are non-empty");
 	}
 }
 
@@ -441,11 +443,12 @@ static bool ParquetProjectionExpressionPushdown(ClientContext &context,
 
 	auto &reader = bind_data.initial_reader->Cast<ParquetReader>();
 	auto &children = reader.root_schema->children;
+	if (idx >= children.size()) {
+		return false;
+	}
 	for (const auto &group : reader.GetFileMetadata()->row_groups) {
-		if (idx >= children.size()) {
-			continue;
-		}
 		const idx_t column_flat_idx = children[idx].column_index;
+		D_ASSERT(column_flat_idx < group.columns.size());
 		for (const Encoding::type type : group.columns[column_flat_idx].meta_data.encodings) {
 			if (type == duckdb_parquet::Encoding::DELTA_LENGTH_BYTE_ARRAY) {
 				return false;
@@ -454,9 +457,6 @@ static bool ParquetProjectionExpressionPushdown(ClientContext &context,
 	}
 
 	const LogicalType type = LogicalType::BIGINT;
-	if (auto &schema = bind_data.reader_bind.schema; !schema.empty()) {
-		schema[idx].type = type;
-	}
 	bind_data.types[idx] = type;
 	bind_data.columns[idx].type = type;
 
--- a/extension/parquet/parquet_multi_file_info.cpp
+++ b/extension/parquet/parquet_multi_file_info.cpp
@@ -2,13 +2,13 @@
 #include "duckdb/main/client_context.hpp"
 
 #include <stdint.h>
-#include <atomic>
 #include <unordered_map>
-#include <vector>
 
 #include "duckdb/common/multi_file/multi_file_function.hpp"
 #include "duckdb/common/serializer/serializer.hpp"
 #include "duckdb/common/serializer/deserializer.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/planner/table_filter.hpp"
 #include "parquet_crypto.hpp"
 #include "duckdb/function/table_function.hpp"
 #include "duckdb/common/assert.hpp"
@@ -30,6 +30,7 @@
 #include "duckdb/parser/parsed_expression.hpp"
 #include "parquet_column_schema.hpp"
 #include "parquet_file_metadata_cache.hpp"
+#include "parquet_reader.hpp"
 #include "parquet_types.h"
 
 namespace duckdb {
@@ -58,6 +59,7 @@ struct ParquetReadBindData : public TableFunctionData {
 	idx_t initial_file_data_size = 0;
 	idx_t explicit_cardinality = 0; // can be set to inject exterior cardinality knowledge (e.g. from a data lake)
 	unique_ptr<ParquetFileReaderOptions> options;
+	unordered_map<idx_t, ParquetReaderProjectionExpression> projection_expressions;
 
 	ParquetOptions &GetParquetOptions() {
 		return options->options;
@@ -74,6 +76,7 @@ struct ParquetReadBindData : public TableFunctionData {
 		result->initial_file_data_size = initial_file_data_size;
 		result->explicit_cardinality = explicit_cardinality;
 		result->options = make_uniq<ParquetFileReaderOptions>(options->options);
+		result->projection_expressions = projection_expressions;
 		return std::move(result);
 	}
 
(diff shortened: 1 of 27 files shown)


myrrc:
Testing manually:

```
COPY (SELECT repeat(md5(i::VARCHAR), 200) AS s FROM range(1500000) t(i))
TO '/tmp/lentest.parquet' (COMPRESSION 'uncompressed', ROW_GROUP_SIZE 100000);

-- cache warmup
.timer on
SELECT sum(strlen(s)) FROM '/tmp/lentest.parquet';

SELECT sum(strlen(s)) FROM '/tmp/lentest.parquet';

SET disabled_optimizers='scalar_fn_pushdown';
SELECT sum(strlen(s)) FROM '/tmp/lentest.parquet';
```

With pushdown user time mean: 0.46s.
Without pushdown user time mean: 0.92s

extension/parquet/parquet_multi_file_info.cpp
```diff
@@ -342,6 +348,54 @@ static void ParquetScanGetMetrics(TableFunctionGetMetricsInput &input) {
...
+	for (idx_t proj_idx = 0; proj_idx < col_ids.size(); proj_idx++) {
+		if (col_ids[proj_idx].GetPrimaryIndex() != idx) {
+			continue;
+		}
+		if (input.get.table_filters.HasFilter(ProjectionIndex(proj_idx))) {
+			return false;
+		}
+		break;
+	}
+	bind_data.types[idx] = LogicalType::BIGINT;
```
lnkuiper: This updates `bind_data.types`/`bind_data.columns`, but those are not the only schemas used to initialize readers. `MultiFileFunction` uses `reader_bind.schema` when it is present, and DuckLake supplies an explicit schema through its custom multi-file reader. That leaves DuckLake reading this column as `VARCHAR` while the rewritten scan output expects `BIGINT`, matching the CI failure:
```
ExpressionExecutor::Execute called with a result vector of type BIGINT that does not match expression type VARCHAR.
```

Please either update every schema source used by the reader, including `reader_bind.schema`, or decline this pushdown when an explicit/custom reader schema is present.

extension/parquet/parquet_reader.cpp
```diff
@@ -1490,6 +1502,11 @@ void ParquetReader::InitializeScan(ClientContext &context, ParquetReaderScanStat
 	for (idx_t i = 0; i < column_indexes.size(); i++) {
 		auto &index = column_indexes[i];
 		auto column_id = index.GetPrimaryIndex();
+		if (length_pushdown_columns.count(column_id)) {
+			auto &schema = root_schema->children[column_id];
+			state.column_readers[i] = make_uniq<ByteArrayLengthColumnReader>(*this, schema);
```
lnkuiper, line 1526: This path also needs to account for Parquet encodings that still expect a string column reader. In particular, `DELTA_LENGTH_BYTE_ARRAY` currently requires `PhysicalType::VARCHAR` and casts the reader to `StringColumnReader`, so a valid Parquet file with that encoding can fail once the schema is rewritten to`BIGINT`.

Please either add length-reader support for the affected encodings or reject length pushdown for unsupported encodings.

extension/parquet/parquet_multi_file_info.cpp
```diff
@@ -57,6 +59,9 @@ struct ParquetReadBindData : public TableFunctionData {
 	idx_t initial_file_data_size = 0;
 	idx_t explicit_cardinality = 0; // can be set to inject exterior cardinality knowledge (e.g. from a data lake)
 	unique_ptr<ParquetFileReaderOptions> options;
+	// storage column indices where projection_expression_pushdown pushed
+	// strlen/octet_length
+	unordered_set<idx_t> length_pushdown_columns;
```
lnkuiper: If we add things to the bind data, they must also be serialized, otherwise logical plans are not round-trippable. We need to be careful when adding new serialization, because we need these plans to stay roundtrippable (we can't just go back on our decisions after a release and invalidate old plans, but we could ignore fields if we want). So, I'd say we need to be careful with adding things here. It also seems very specific to length pushdown, which is not generic/extensible in the future.

Maybe something like:
```C++
enum class ParquetReaderProjectionExpressionType : uint8_t {
    BYTE_LENGTH,
};

struct ParquetReaderProjectionExpression {
    ParquetReaderProjectionExpressionType type;
    LogicalType return_type;
};

unordered_map<idx_t, ParquetReaderProjectionExpression> projection_expressions;
```

test/configs/verification_projection.json
```diff
@@ -4,9 +4,10 @@
   "skip_compiled": "true",
   "skip_tests": [
     {
-      "reason": "Type pushdown optimizer collapses projections and checks for that",
+      "reason": "Type and function pushdown optimizers collapses projections and check for that",
       "paths": [
-        "test/optimizer/pushdown/csv_type_pushdown.test"
+        "test/optimizer/pushdown/csv_type_pushdown.test",
+        "test/optimizer/pushdown/scalar_function_pushdown.test",
```
lnkuiper: CI is failing due to a trailing comma here

lnkuiper (changes requested):
Thanks for the PR, I like the idea! I've left some comments:

extension/parquet/parquet_reader.cpp
myrrc: I've initially written support for delta_length but removed it to avoid making the PR even bigger. I can either make changes here or reject delta here and make a folloup pr with its implementation, what do you prefer?

lnkuiper: Either is fine, as long as we add a test in this PR so we make sure the behaviour doesn't break

test/configs/verification_projection.json
myrrc: +1

extension/parquet/parquet_multi_file_info.cpp
myrrc: I'm not sure how we can do it for all readers except bind_data.initial_reader since they're not populated eagerly. I've updated the initial reader check

extension/parquet/parquet_reader.cpp
myrrc: Afaik delta length is the only special encoding so far, I've added a check for it.

extension/parquet/parquet_multi_file_info.cpp
myrrc: +1

lnkuiper:
Looks good! Almost ready, there's still one issue: `DELTA_LENGTH_BYTE_ARRAY` is only checked on the initial Parquet reader. Repro:
```sql
# name: test/temp.test
# description: Temporary repro for scalar function pushdown review
# group: [temp]

require parquet

statement ok
COPY (SELECT 'a'::VARCHAR AS s)
TO '{TEST_DIR}/00_scalar_fn_plain.parquet' (FORMAT PARQUET, PARQUET_VERSION v1);

statement ok
COPY (SELECT range::VARCHAR AS s FROM range(30))
TO '{TEST_DIR}/99_scalar_fn_delta.parquet' (FORMAT PARQUET, PARQUET_VERSION v2, DICTIONARY_SIZE_LIMIT 1);

query I
SELECT strlen(s)
FROM read_parquet(['{TEST_DIR}/00_scalar_fn_plain.parquet', '{TEST_DIR}/99_scalar_fn_delta.parquet'])
ORDER BY ALL;
----
1
1
1
1
1
1
1
1
1
1
1
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
```

myrrc:
The issue with other readers is that we don't know metadata for them beforehand because they're parsed lazily unless it's a UNION ALL. I've done a solution with TryLoadCaches and didn't like its complexity (there are a couple other hidden issues with it) so let's do single-file pushdown as for now.

I've also fixed a couple of other correctness bugs, feel free to re-review.

lnkuiper:
This breaks for nested stuff:
```sql
# name: test/temp.test
# description: Temporary repro for scalar function pushdown review
# group: [temp]

require parquet

statement ok
COPY (
    SELECT {'x': 'a', 'y': 'bbbb'} AS s, 'zzzzzzzz' AS trailing
) TO '{TEST_DIR}/scalar_fn_nested_field_wrong_leaf.parquet' (FORMAT PARQUET, PARQUET_VERSION v1);

query I
SELECT strlen(s.x)
FROM '{TEST_DIR}/scalar_fn_nested_field_wrong_leaf.parquet';
----
1
```

myrrc:
Thanks, I forgot about PushdownExtract. Fixed

lnkuiper:
The failing CI is not related to your changes I think, we'll look into fixing that

lnkuiper:
I think there is still an issue:
```sql
# name: test/temp.test
# description: Temporary repro for scalar function pushdown review
# group: [temp]

require parquet

statement ok
COPY (
    SELECT {'x': 'a', 'y': 'bbbb'} AS s
) TO '{TEST_DIR}/scalar_fn_nested_mixed.parquet' (FORMAT PARQUET, PARQUET_VERSION v1);

query IT
SELECT strlen(s.x), s.y
FROM '{TEST_DIR}/scalar_fn_nested_mixed.parquet';
----
1	bbbb
```

We can either choose to reject scalar function pushdown for these cases, or key the pushed expression by the full `ColumnIndex` path.

myrrc:
Let's do both. I want to allow changing this in the future without breaking compatibility and also make the diff a bit smaller.

extension/parquet/parquet_multi_file_info.cpp
```diff
@@ -290,6 +293,11 @@ static void ParquetScanSerialize(Serializer &serializer, const optional_ptr<Func
 	if (serializer.ShouldSerialize(StorageVersion::V1_2_0)) {
 		serializer.WriteProperty(104, "table_columns", bind_data.table_columns);
 	}
+	if (serializer.ShouldSerialize(StorageVersion::V2_0_0)) {
+		serializer.WriteProperty(105, "projection_expressions", parquet_data.projection_expressions);
+	} else if (!parquet_data.projection_expressions.empty()) {
+		throw SerializationException("Version too old to serialize projection_expressions but they are non-empty");
```
lnkuiper: This version gate seems too strict for query-plan serialization. The deserializer already has an explicit empty default for older payloads below, so old serialized plans can still be read. But on the write side, `json_serialize_plan(..., optimize := true)` uses the default serializer compatibility, and after scalar function pushdown `projection_expressions` is non-empty, so serialization returns an error JSON instead of a plan:

`Version too old to serialize projection_expressions but they are non-empty`

Since this state is required to deserialize the optimized scan with the correct output types, I think we either need to serialize it for plan serialization whenever it is present, or make the plan serializer use latest compatibility. Otherwise optimized Parquet scans using this pushdown become unserializable.

Repro:
```sql
require json

require parquet

statement ok
COPY (SELECT 'abc' AS s, 42 AS i)
TO '{TEMP_DIR}/scalar_fn_serialized.parquet' (FORMAT PARQUET);

query I
SELECT json_extract(
    json_serialize_plan(
        'SELECT strlen(s), i FROM ''{TEMP_DIR}/scalar_fn_serialized.parquet'' ORDER BY i',
        optimize := true
    ),
    '$.error'
)::BOOLEAN;
----
0
```

Maybe it's best to serialize it whenever it is present.

lnkuiper (commented):
Looking good! There's a small tricky issue with serialization, then this is good to go

extension/parquet/parquet_multi_file_info.cpp
myrrc: I'm worried about the situation where the version is too old to serialize pushdown expressions, but they're non-empty. In this case we have already changed types, so we need to serialize these as well, but we can't.

lnkuiper: Yeah this is tricky indeed. I think if we have pushed down expressions, we just need to accept that older versions can't deserialise the plan. Throwing on serialisation makes it so we can't even use the plan in newer versions

myrrc: Ok, I'll add a comment to explain this situation

myrrc:
I've checked both failures and they seem unrelated

lnkuiper (approved):
Thanks!

PR #20997, opened 2026-02-18
Better filter cardinality estimation using automatically maintained table samples

sourceterm:

PR #23530, opened 2026-06-30
Configurable approximate join order threshold

jonaskr8:
### Hello
I am a computer science student of the university of Tübingen and have just started working on my masters thesis supervised by @Teggy  and @timfi. Over the course of my masters thesis, I will take a closer look at the optimizer and try to find and implement some improvements. This PR is the first (tiny) contribution.

## Changes
Introduces a new setting `approximate_join_order_threshold`. This setting lets users configure the maximum number of tables for which the optimizer computes the join order using the exact algorithm, instead of it being fixed to 12 join partners.

## Benchmark
From the [join-order-benchmark](https://github.com/gregrahn/join-order-benchmark), some queries profit from changing the threshold. For example, query `7c` which has 8 join partners speeds up a lot with the approximate algorithm. The measurements presented were made on my laptop, the trend is the same on the universities server though.

Using `.timer on` :
|threshold | query execution time|
|-----------|------------------------|
|8 (approximate)|Run Time (s): real 0.208 user 0.643378 sys 0.470962|
|9 (exact)| Run Time (s): real 1.772 user 2.194499 sys 4.251899|

Using detailed profiling with `PRAGMA enable_profiling;`
|threshold| join_order_optimizer time| sum of all join `operator_timing`s|
|----------|-----------------------------|------------------------------------------|
|8|0.00023|0.4561|
|9|0.00088|20.6082|

The fact that actually doing the joins takes that much longer with the exact algorithm is not what I expected. 

## Testing
Extended an existing unit test to verify that the newly introduced setting doesn't break anything.

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -115,6 +115,13 @@
         "custom_implementation": true
     },
     {
+        "name": "approximate_join_order_threshold",
+        "description": "The minimum number of tables in a join to determine the optimal join order approximately instead of exactly.",
+        "type": "UBIGINT",
+        "default_scope": "local",
+        "default_value": "12"
+    },
+    {
         "name": "arrow_large_buffer_size",
         "description": "Whether Arrow buffers for strings, blobs, uuids and bits should be exported using large buffers",
         "type": "BOOLEAN",
--- a/src/include/duckdb/main/settings.hpp
+++ b/src/include/duckdb/main/settings.hpp
@@ -258,6 +258,17 @@ struct AllowedPathsSetting {
 	static Value GetSetting(const ClientContext &context);
 };
 
+struct ApproximateJoinOrderThresholdSetting {
+	using RETURN_TYPE = idx_t;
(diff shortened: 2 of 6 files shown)


lnkuiper:
Thanks for the PR! Cool that you want to look into this :) I recently made some changes to add `LEFT` join ordering and improve cardinality estimates a bit, but there is probably still a lot of work to be done to get better join orders on some queries.

There is https://github.com/duckdb/duckdb/pull/20997 which adds sample-based estimations, and is a big change (with merge conflicts probably now) that we haven't had the time to review thoroughly.

I think a large pain point is our filter selectivity estimation, which we basically don't have because we don't have the samples yet. Additionally, I think we might want to improve the way we do estimations for filters on columns where values correlate with each other (like the `date_dim` table in TPC-DS).

We also recently merged https://github.com/duckdb/duckdb/pull/22298 which does build side swapping based on join filters. We might be able to do something similar in the join order optimiser cost model.

PR #23534, opened 2026-06-30
Fix arrow type extension bugs

evertlammerts:
Fixes duckdb/duckdb-python#517

This is a patch for the reported issue. The underlying issue is that we don't have a central, authoritative source of truth for the encoding we need to produce, and if the ArrowAppender and type extensions disagree then we can get a schema that does not fit the output.

We should fix this properly in main.

Python regression tests are up in https://github.com/duckdb/duckdb-python/pull/520, I'll land that together with a submodule bump after this lands.


pdet:
Thanks, LGTM!

PR #23535, opened 2026-06-30
Don't push down CAST if there's a TRY_CAST

### LINKED ISSUE #8310: Epic: Duckdb projection expression pushdown

myrrc:
This a fix for https://github.com/duckdb/duckdb/pull/22788.
Previously, we allowed to push CAST and TRY_CAST at once if type for column was the same.

It doesn't work for readers like Vortex which have cast() but not try_cast(). After this PR, CAST+TRY_CAST is registered as a conflict

### DIFF
--- a/src/optimizer/type_pushdown.cpp
+++ b/src/optimizer/type_pushdown.cpp
@@ -202,10 +202,11 @@ unique_ptr<Expression> CastCollect::VisitReplace(BoundCastExpression &expr, uniq
 		if (top_level_casts.count(&expr)) {
 			col_to_cast.emplace(binding->column_index, &expr);
 		}
-	} else if (it->second == nullptr || it->second->GetReturnType() != expr.GetReturnType()) {
+	} else if (it->second == nullptr || it->second->GetReturnType() != expr.GetReturnType() ||
+	           it->second->Cast<BoundCastExpression>().IsTryCast() != expr.IsTryCast()) {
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks for the PR! I think this still pushes down `TRY_CAST` in some cases:
```sql
# name: test/temp.test
# group: [temp]

statement ok
COPY (SELECT 'hello' AS column0) TO '{TEST_DIR}/try_cast_pushdown.csv' (HEADER false);

query I
SELECT TRY_CAST(column0 AS INTEGER)
FROM read_csv('{TEST_DIR}/try_cast_pushdown.csv', header=false, auto_detect=false, columns={'column0':'VARCHAR'});
----
NULL
```

myrrc:
> I think this still pushes down TRY_CAST in some cases:

The goal of this PR is not to avoid pusing down TRY_CAST but to avoid pushing down a pair of CAST,TRY_CAST as one of them.

Your example has only a TRY_CAST which should be pushed. In our case, Vortex will reject the push because we don't support TRY_CAST, but other reader may as well push it.

lnkuiper:
Ah, I see. The test I shared succeeds on v1.5.4, but not on `main` (and therefore not on this branch either). The problem is caused by #22788. So this PR seems good to go then, but that issue should also be fixed.

PR #23536, opened 2026-06-30
add --emit-on-skip flag -- emits SKIP_TEST to stderr, skip

benfleis:
Useful for external tooling (eg pytest) to get per-test attribution of skips. (Previously this is only aggregated by Catch/unittest.)

### DIFF
--- a/test/helpers/test_helpers.cpp
+++ b/test/helpers/test_helpers.cpp
@@ -23,6 +23,7 @@ namespace duckdb {
 static string custom_test_directory;
 static case_insensitive_set_t required_requires;
 static bool delete_test_path = true;
+static bool emit_on_skip = false; // --emit-on-skip opt-in: emit a [SKIP_TEST] marker per skipped test
 
 bool NO_FAIL(QueryResult &result) {
 	if (result.HasError()) {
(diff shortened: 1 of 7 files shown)


smvv:
> Useful for external tooling (eg pytest) to get per-test attribution of skips. (Previously this is only aggregated by Catch/unittest.)

fyi we're collecting mode and skip reasons in the test runner [here](https://github.com/duckdb/duckdb/blob/3cb65aa794b4c985a98493f188fb58db79a370f8/scripts/ci/run_tests.py#L1183-L1205).

Does this PR also emit mode skip statements?

benfleis:
> > Useful for external tooling (eg pytest) to get per-test attribution of skips. (Previously this is only aggregated by Catch/unittest.)
> 
> fyi we're collecting mode and skip reasons in the test runner [here](https://github.com/duckdb/duckdb/blob/3cb65aa794b4c985a98493f188fb58db79a370f8/scripts/ci/run_tests.py#L1183-L1205).
> 
> Does this PR also emit mode skip statements?

No, it doesn't! Let me add that.

Afterward marking as Draft, since i want to consider a simple but far more useful emitter.

benfleis:
obviated by the more advanced stats emitting variant.

PR #23539, opened 2026-06-30
Issue #18039: Nested Comparison Semantics 

hawkfish:
* Change nested comparisons to use SQL NULL semantics at all levels instead of using DISTINCT semantics internally.
* Also use DISTINCT semantics when checking nested types in aggregate hash tables.
* Change UNION equality comparisons to only compare single columns.


hawkfish:
I changed some test queries to match what I thought the original intent was; the others it seemed that the results were the important change.

lnkuiper:
Thanks, looks great! Codex found one issue with `UNION` but other than that it's all good:
```sql
# name: test/temp.test
# group: [temp]

query ITT
WITH t(id, l, r) AS (
	VALUES
		(0,
			[union_value(num := 1)::UNION(num INT, str VARCHAR), union_value(num := 2)::UNION(num INT, str VARCHAR)]::UNION(num INT, str VARCHAR)[],
			[union_value(num := 1)::UNION(num INT, str VARCHAR), union_value(num := 2)::UNION(num INT, str VARCHAR)]::UNION(num INT, str VARCHAR)[]),
		(1,
			[union_value(str := 'x')::UNION(num INT, str VARCHAR)]::UNION(num INT, str VARCHAR)[],
			[union_value(str := 'x')::UNION(num INT, str VARCHAR)]::UNION(num INT, str VARCHAR)[])
)
SELECT id, l = r, l IS NOT DISTINCT FROM r FROM t ORDER BY id;
----
0	true	true
1	true	true

```

Feel free to mark `Ready To Merge` once you've resolved this :)

hawkfish:
> Thanks, looks great! Codex found one issue with `UNION` but other than that it's all good:

"You are in a maze of twisty little selection vectors all alike..."

PR #23540, opened 2026-06-30
Dont special-case geometry stats pruning, pass child stats to filter callback

Maxxen:
This PR turns the geometry "&&" stats pruning into a "filter_prune" scalar function callback, instead of having a special case in the expression filter evaluation.

Geometry stats are now also propagated across geometry CRS casts automatically.

Additionally the function filter_prune input now contains the bound function expression, as well as the stats of all child arguments, not just the first child. I've updated the existing join-filter-pushdown functions to adapt to this change.

### DIFF
--- a/src/optimizer/statistics/expression/propagate_cast.cpp
+++ b/src/optimizer/statistics/expression/propagate_cast.cpp
@@ -228,6 +228,11 @@ unique_ptr<BaseStatistics> StatisticsPropagator::TryPropagateCast(const BaseStat
 		// the cast shreds every value into a single bucket - mirror the (possibly nested) source as typed stats
 		return VariantStats::StatisticsPropagateToVariant(source, stats);
 	}
+	if (source.id() == LogicalTypeId::GEOMETRY && target.id() == LogicalTypeId::GEOMETRY) {
+		// A geometry -> geometry cast only changes CRS metadata, not coordinates, so the bounding box,
+		// type set and null-ness are unchanged: propagate the statistics as-is.
+		return stats.Copy().ToUnique();
+	}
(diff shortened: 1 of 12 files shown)


src/optimizer/statistics/expression/propagate_cast.cpp
```diff
@@ -228,6 +228,11 @@ unique_ptr<BaseStatistics> StatisticsPropagator::TryPropagateCast(const BaseStat
 		// the cast shreds every value into a single bucket - mirror the (possibly nested) source as typed stats
 		return VariantStats::StatisticsPropagateToVariant(source, stats);
 	}
+	if (source.id() == LogicalTypeId::GEOMETRY && target.id() == LogicalTypeId::GEOMETRY) {
+		// A geometry -> geometry cast only changes CRS metadata, not coordinates, so the bounding box,
+		// type set and null-ness are unchanged: propagate the statistics as-is.
+		return stats.Copy().ToUnique();
```
pdet, line 234: Is this a problem?

```sql
SET debug_verify_stats=true;
CREATE TABLE t(g GEOMETRY);
INSERT INTO t VALUES ('POINT(0 1)');
SELECT g::GEOMETRY('OGC:CRS84') FROM t;  
```

pdet (commented):
Thanks for the PR, just one comment.

src/optimizer/statistics/expression/propagate_cast.cpp
Maxxen: No this is a legal cast, you can always cast a non-crs geom to a crs-geom and vice-versa, the cast only fails if both sides have a CRS and they are different.

PR #23543, opened 2026-07-01
Fold empty list contains filters to EMPTY_RESULT

### LINKED ISSUE #23542: List contain does sequential iteration for empty list
### What happens?

Hi team, I found sequential scan is performed when we do `list_contains` against an empty list, see repro example, should be straightforward.

### To Reproduce

```sql
-- t is a table, doesn't matter
memory D EXPLAIN SELECT * FROM t WHERE list_contains([], a);
╭─ Seq Scan ────────────────────╮
│ Table: memory.main.t          │
│ Projections: a                │
│ Filters: list_contains([], a) │
│ ~1 row                        │
```

### OS:

macos

### DuckDB Version:

3cb65aa794b4c985a98493f188fb58db79a370f8

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23542

This PR handles empty list in the contains optimizer to emit empty result directly, so we don't do sequential scan, as it's described in the issue.

### DIFF
--- a/src/optimizer/rule/contains_to_in_clause.cpp
+++ b/src/optimizer/rule/contains_to_in_clause.cpp
@@ -9,7 +9,8 @@ namespace duckdb {
 
 ContainsToInClauseRule::ContainsToInClauseRule(ExpressionRewriter &rewriter) : Rule(rewriter) {
 	auto func = make_uniq<FunctionExpressionMatcher>();
-	func->function = make_uniq<SpecificFunctionMatcher>("contains");
+	identifier_set_t functions = {"contains", "list_contains", "list_has", "array_contains", "array_has"};
+	func->function = make_uniq<ManyFunctionMatcher>(functions);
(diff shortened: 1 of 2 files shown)


lnkuiper (changes requested):
Looks good, just needs a little change to make sure we don't change this behaviour:
```sql
# name: test/temp.test
# group: [temp]

query I
SELECT list_contains([NULL], a) FROM (VALUES (1)) t(a);
----
false
```

dentiny:
> just needs a little change to make sure we don't change this behaviour:

Cool, updated -- checked with quite a few list operations, should be compatible now.

lnkuiper (approved):
Thanks for the changes!

PR #23546, opened 2026-07-01
Address test warning

dentiny:
I get a few warning messages on tests, so I asked agents to update them all.
```sh
[65/193] (33%): ...l/copy/parquet/tpch_parquet_copy_cast.test_slow took 0.009sReplacing deprecated string __TEST_DIR__ in path "COPY (SELECT range AS a FROM range(102400)) TO '__TEST_DIR__/expr_prune_int.parquet' (ROW_GROUP_SIZE 10240)" - please replace with {TEST_DIR}
Replacing deprecated string __TEST_DIR__ in path "COPY (SELECT range::DOUBLE AS d FROM range(102400)) TO '__TEST_DIR__/expr_prune_double.parquet' (ROW_GROUP_SIZE 10240)" - please replace with {TEST_DIR}
```

### DIFF
--- a/test/sql/copy/csv/test_copy_gzip_large_block.test_slow
+++ b/test/sql/copy/csv/test_copy_gzip_large_block.test_slow
@@ -3,12 +3,12 @@
 # group: [csv]
 
 statement ok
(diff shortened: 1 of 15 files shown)


pdet:
Thanks, LGTM

PR #23548, opened 2026-07-01
Rename columns to original names before passing to reader if they were aliased

### LINKED ISSUE #8632: Duckdb bug: aliased column in where is folded with aliased name to vortex
```
EXPLAIN
WITH cte AS (SELECT col AS other FROM '$__TEST_DIR__/cte-alias.vortex')
SELECT * FROM cte WHERE other IS NOT NULL ORDER BY other;

   physical_plan ┌───────────────────────────┐
   │          ORDER_BY         │
   │    ────────────────────   │
   │       cte.other ASC       │
   └─────────────┬─────────────┘
   ┌─────────────┴─────────────┐
   │        READ_VORTEX        │
   │    ────────────────────   │
   │   Function: Vortex Scan   │
   │          Filters:         │
   │    is_not_null($.other)   │
   │      Projections: col     │
   └───────────────────────────┘
```

This is a duckdb bug since it should request an original
column's name when pushing CTE through Vortex.

#8327

myrrc:
In the following query

WITH cte AS (SELECT col AS other FROM '1.vortex') SELECT * WHERE other > 0 FROM cte,

Vortex reader fails because a complex filter pushed ("WHERE other > 0") contains a column "other" not present in the file. This happens because BoundColumnRefExpression contains the aliased name, not the original one.

When CTE inline optimizer collapses the CTE, the aliased column name is not changed.
This PR renames columns back to their unaliased names before passing to the reader.

### DIFF
--- a/src/optimizer/pushdown/pushdown_get.cpp
+++ b/src/optimizer/pushdown/pushdown_get.cpp
@@ -4,14 +4,46 @@
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/expression/bound_operator_expression.hpp"
 #include "duckdb/planner/expression/bound_parameter_expression.hpp"
-#include "duckdb/planner/operator/logical_filter.hpp"
+#include "duckdb/planner/expression_iterator.hpp"
 #include "duckdb/planner/operator/logical_get.hpp"
 #include "duckdb/planner/operator/logical_empty_result.hpp"
 
 namespace duckdb {
+
+/**
+ * When a BoundColumnRefExpression that's part of expr (a filter) arrives here, its
+ * name will be set to the projection name i.e. "other" for SELECT col as other.
+ * If CTE inlining optimizer collapses the CTE in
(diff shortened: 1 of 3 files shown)


lnkuiper:
Is this working around a problem in Vortex specifically? DuckDB is always able to correctly identify the column through its `ColumnIndex`, not by name. I feel like Vortex should be able to do the same? Aliases are never used for indexing columns in DuckDB logical/physical plans, the binder has already done that work

myrrc:
> Is this working around a problem in Vortex specifically?

Any reader which reads column by name (which is a valid option since it's exposed through API) should be able to read the correct name. This is also a case for Vortex.

Reason number two is that we, in fact, can't use the column binding. When we're pushing down a complex filter, we can get a storage index from the scan index provided by ColumnBinding, but at this point we don't yet have the fields (they will be supplied in the function() call).
Vortex doesn't have a way to extract a positional column from a file since the file is a Struct with named fields.

lnkuiper:
Right, I don't see any harm in this, so I think this is OK to merge

myrrc:
@lnkuiper Thank you!

PR #23557, opened 2026-07-01
Collect exact rowgroup stats when possible

### LINKED ISSUE #23556: Count on tables with no deletion does sequential scan
### What happens?

Hi team, I found when I create a table with no rows deleted, `count` operation does sequential scan, which I expect to directly return the stats. See example SQL for repro, should be straightforward.

### To Reproduce

```sql
memory D CREATE TABLE big AS
SELECT i AS id, (i % 100) AS g, random() AS v
FROM range(0, 5_000_000) t(i);
memory D EXPLAIN SELECT count(*) FROM big;
╭─ Ungrouped Aggregate ────╮
│ Aggregates: count_star() │
│ ~1 row │
╰─────────────┬────────────╯
╭─ Seq Scan ──┴────────────╮
│ Table: memory.main.big │
│ Type: Sequential Scan │
│ ~5,000,000 rows │
╰──────────────────────────╯
```

### OS:

macos

### DuckDB Version:

10d6a343137e396db3adf38a6dffa4ec56e3a73e

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23556

As described in the issue, when there's no deletion current `count` does sequential scan somehow.
Root cause is: when finalizing a row group, version info is always [appended](https://github.com/duckdb/duckdb/blob/d9a775e4c03b23ecb3784f879196aa81adf0ac1c/src/storage/table/row_group_collection.cpp#L684), so fetching partitions stats always falls into [approximate branch](https://github.com/duckdb/duckdb/blob/d9a775e4c03b23ecb3784f879196aa81adf0ac1c/src/storage/table/row_group.cpp#L1917)

In this PR, I considered deletions in the version info as well.

I have ran CI on my fork: https://github.com/dentiny/duckdb/pull/117


src/include/duckdb/function/partition_stats.hpp
```diff
@@ -41,6 +41,8 @@ struct PartitionStatistics {
 	idx_t count;
 	//! Whether or not the count is exact or approximate
 	CountType count_type;
+	//! Whether the partition's min/max column statistics are exact.
+	bool min_max_exact = false;
```
lnkuiper: This creates a second way of judging whether min/max is exact, next to `PartitionRowGroup::MinMaxIsExact()`, and these could in theory disagree. I think there should be a single source of truth, which should be the function

lnkuiper (changes requested):
Thanks! I've left a comment:

src/include/duckdb/function/partition_stats.hpp
dentiny: Yes that's valid point; I separate `HasPendingWrite` out in the latest commit.

lnkuiper:
Thanks for implementing the feedback! Looks good.

Codex found this repro failing:
```sql
# name: test/temp.test
# description: Reproducer for folded count(*) prepared across rollback
# group: [temp]

statement ok
CREATE TABLE t AS SELECT range::INTEGER AS i FROM range(11);

statement ok
BEGIN;

statement ok
DELETE FROM t WHERE i = 0;

statement ok
PREPARE s AS SELECT count(*) FROM t;

query I
EXECUTE s
----
10

statement ok
ROLLBACK;

query I
EXECUTE s
----
11
```

dentiny:
image
CI summary failure doesn't seem to be related

lnkuiper:
I'll rerun the failing checks, and then do one last review :)

lnkuiper (approved):
Thanks for the changes! LGTM

PR #23559, opened 2026-07-02
Optimizer: use UTF-8 successor for LIKE prefix pushdown

wufengwind:
LIKE table filter pushdown derives a range filter from the constant prefix of a LIKE pattern.

The upper bound was previously built by incrementing the last byte of the prefix directly. Prefix filter pushdown already uses Utf8Proc::FindNextLegalUTF8 for this, which avoids constructing invalid UTF-8 boundaries and handles boundary prefixes consistently.

This changes LIKE prefix range pushdown to use the same UTF-8 successor helper. If no successor exists, the pushdown keeps the safe lower bound and leaves the original LIKE predicate in place.

A sqllogictest covers boundary UTF-8 prefixes through the LIKE table filter pushdown path.

### DIFF
--- a/test/optimizer/filter_pushdown/like_prefix_boundary.test
+++ b/test/optimizer/filter_pushdown/like_prefix_boundary.test
@@ -1,0 +1,25 @@
+# name: test/optimizer/filter_pushdown/like_prefix_boundary.test
+# description: Test LIKE filter pushdown with boundary UTF-8 prefixes
+# group: [filter_pushdown]
+
+statement ok
+CREATE TABLE strings(s VARCHAR);
+
+statement ok
+INSERT INTO strings VALUES (chr(127)), (chr(127) || 'abc'), (chr(128)), ('abc');
+
+query I
+SELECT encode(s) FROM strings WHERE s LIKE chr(127) || '%' || 'c' ORDER BY 1
+----
+\x7Fabc
+
+statement ok
+DELETE FROM strings;
+
+statement ok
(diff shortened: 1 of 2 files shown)


test/optimizer/filter_pushdown/like_prefix_boundary.test
```diff
@@ -0,0 +1,27 @@
...
+# group: [filter_pushdown]
+
+statement ok
+CREATE TABLE strings(s VARCHAR);
+
+statement ok
+INSERT INTO strings VALUES (chr(127)), (chr(127) || 'abc'), (chr(128)), ('abc');
+
+query I
+SELECT encode(s) FROM strings WHERE s LIKE chr(127) || '%' ORDER BY 1
```
lnkuiper: This pattern is constant-folded and then rewritten by `LikeOptimizationRule` into `prefix(s, chr(127))`, so it exercises `TryPushdownPrefixFilter` rather than the `TryPushdownLikeFilter` code changed in this PR.

Can we add a non-prefix-optimizable LIKE pattern that still has the same fixed prefix, e.g. `s LIKE chr(127) || '%' || 'c'` expecting only `\x7Fabc`? The same applies to the `chr(2047) || '%'` case below. That should keep the residual LIKE and cover the UTF-8 successor logic in the modified branch.

wufengwind: Thanks for the catch! Updated the tests to use non-prefix-optimizable LIKE patterns with the same fixed prefixes, so the residual LIKE is kept while covering the UTF-8 successor range pushdown path.

PR #23561, opened 2026-07-02
Fold unsatisfiable NOT(comparison) filters to empty result

### LINKED ISSUE #23545: Filter with NOT(comparison) is not pruned in optimization
### What happens?

Hi team, I found NOT is sometimes not optimized away for unsatisfied predicate. See example, should be straightforward.

### To Reproduce

```sql
memory D CREATE TABLE t AS SELECT * FROM (VALUES (1),(2),(3)) v(a);

-- The predicate is unsatisfied, but we do sequential scan for the table.
memory D EXPLAIN SELECT * FROM t WHERE a > 1 AND NOT (a > 1);
╭─ Seq Scan ─────────────────────────╮
│ Table: memory.main.t │
│ Type: Sequential Scan │
│ Projections: a │
│ Filters: (a > 1) AND (NOT (a > 1)) │
│ ~1 row │

-- Another predicate without NOT is pruned correctly.
memory D EXPLAIN SELECT * FROM t WHERE a = 2 AND a != 2;
╭─ Empty Result ─╮
╰────────────────╯
```

### OS:

macos

### DuckDB Version:

latest main commit: 3cb65aa794b4c985a98493f188fb58db79a370f8

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

EshwarCVS:
Fixes #23545

## Problem

Predicates like `a > 1 AND NOT(a > 1)` are always false, but the optimizer
did not fold them to `EMPTY_RESULT`. A similar contradiction without `NOT`
(e.g. `a = 2 AND a != 2`) was already pruned correctly.

`FilterCombiner` only understands direct comparisons, so `NOT(comparison)`
was left unprocessed and the scan was kept.

## Solution

- Add `NotComparisonSimplificationRule` to rewrite `NOT(comparison)` into
  the negated comparison (e.g. `NOT(a > 1)` → `a <= 1`) before filter pushdown.
- Fix a boundary case in `CompareValueInformation` so `a <= C AND a >= C`
  is treated as satisfiable (`a = C`), while `a > C AND a <= C` remains
  unsatisfiable.

## Before / After

Screenshot 2026-07-02 at 12 39 55 AM

Before: `EXPLAIN SELECT * FROM t WHERE a > 1 AND NOT (a > 1)` → Seq Scan  
After:  `EXPLAIN SELECT * FROM t WHERE a > 1 AND NOT (a > 1)` → Empty Result

## Test plan

- [x] Added `test/sql/optimizer/test_not_comparison_simplification.test`
- [x] Covers contradictions for `>`, `>=`, `<`, `<=`, `=`
- [x] Covers non-contradictory `NOT(comparison)` still returning correct rows
- [x] `make format-fix`
- [x] `build/reldebug/test/unittest test/sql/optimizer/test_not_comparison_simplification.test`
- [ ] `make unit` (or full CI on fork)

### DIFF
--- a/src/optimizer/rule/not_comparison_simplification.cpp
+++ b/src/optimizer/rule/not_comparison_simplification.cpp
@@ -1,0 +1,41 @@
+#include "duckdb/optimizer/rule/not_comparison_simplification.hpp"
+#include "duckdb/planner/expression/bound_operator_expression.hpp"
+#include "duckdb/planner/expression/bound_comparison_expression.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/common/enums/expression_type.hpp"
+
+namespace duckdb {
+
+NotComparisonSimplificationRule::NotComparisonSimplificationRule(ExpressionRewriter &rewriter) : Rule(rewriter) {
+	// match BOUND_OPERATOR with type OPERATOR_NOT
+	auto op = make_uniq<ExpressionMatcher>(ExpressionClass::BOUND_OPERATOR);
+	op->expr_type = make_uniq<SpecificExpressionTypeMatcher>(ExpressionType::OPERATOR_NOT);
+	root = std::move(op);
+}
+
+unique_ptr<Expression> NotComparisonSimplificationRule::Apply(LogicalOperator &op,
+                                                              vector<reference<Expression>> &bindings,
+                                                              bool &changes_made, bool is_root) {
+	auto &not_expr = bindings[0].get().Cast<BoundOperatorExpression>();
+	D_ASSERT(not_expr.GetExpressionType() == ExpressionType::OPERATOR_NOT);
+	D_ASSERT(not_expr.GetChildren().size() == 1);
+
+	auto &child = not_expr.GetChildrenMutable()[0];
+
+	// check if the child is a comparison expression
+	if (!BoundComparisonExpression::IsComparison(*child)) {
+		return nullptr;
+	}
+
+	auto &comparison = child->Cast<BoundFunctionExpression>();
+	auto negated_type = NegateComparisonExpression(comparison.GetExpressionType());
+
+	// set the negated comparison type (updates both expression type and bind data)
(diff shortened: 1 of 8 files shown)


test/sql/optimizer/test_not_comparison_simplification.test
```diff
@@ -0,0 +1,41 @@
+# name: test/sql/optimizer/test_not_comparison_simplification.test
+# description: Test NOT(comparison) simplification and pruning
+# group: [optimizer]
+
+statement ok
```
dentiny, line 5: I think we should at least check plan with `EXPLAIN SELECT` in the unit test?

src/optimizer/rule/not_comparison_simplification.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "duckdb/common/enums/expression_type.hpp"
+
+namespace duckdb {
+
+NotComparisonSimplificationRule::NotComparisonSimplificationRule(ExpressionRewriter &rewriter) : Rule(rewriter) {
+	// match BOUND_OPERATOR with type OPERATOR_NOT
+	auto op = make_uniq<ExpressionMatcher>(ExpressionClass::BOUND_OPERATOR);
+	op->expr_type = make_uniq<SpecificExpressionTypeMatcher>(ExpressionType::OPERATOR_NOT);
+	root = std::move(op);
+}
```
lnkuiper, line 14: We could also simplify `NOT(x IS DISTINCT FROM y)` to `x IS NOT DISTINCT FROM y`

lnkuiper:
Thanks for the PR! Agree with @dentiny 's comment, and I think we can make this even better by also matching `DISTINCT`

EshwarCVS:
Will write, update and push code.

EshwarCVS:
Looks like the issue PR is trying to resolve has been already resolved in the PR https://github.com/duckdb/duckdb/pull/23563

PR #23562, opened 2026-07-02
[variegata] Bump extensions

carlopi:


carlopi:
Added "Bump extension_entries.hpp" on top.

PR #23564, opened 2026-07-02
Add a direct appender

dentiny:
Hi team, I want to write some rows into DuckDB table in my extension, but I found current appender APIs, whether `Appender` or `QueryAppender`, both converts row into SQL statement first.
In this PR, I want to add a new direct appender, as its name suggests, it appends row directly into the table without SQL statement generation and parsing roundtrip.

### DIFF
--- a/src/include/duckdb/main/appender.hpp
+++ b/src/include/duckdb/main/appender.hpp
@@ -201,6 +201,26 @@ protected:
 	void FlushInternal(ColumnDataCollection &collection) override;
 };
 
+class DirectAppender : public BaseAppender {
+public:
+	DUCKDB_API DirectAppender(Connection &con, const Identifier &database_name, const Identifier &schema_name,
+	                          const Identifier &table_name,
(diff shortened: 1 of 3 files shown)


src/include/duckdb/main/appender.hpp
```diff
@@ -201,6 +201,26 @@ class QueryAppender : public BaseAppender {
 	void FlushInternal(ColumnDataCollection &collection) override;
 };
 
+class DirectAppender : public BaseAppender {
+public:
+	DUCKDB_API DirectAppender(Connection &con, const Identifier &database_name, const Identifier &schema_name,
```
Maxxen, line 206: These overloads should probably take a QualifiedName now instead of separate database/schema/table identifiers.

Maxxen:
LGTM, pretty clean, but lets wait until @Mytherin is back to figure out how to proceed.

Mytherin:
I don't really follow the goal of this addition - why do we want to add this / what's the benefit of this versus the `QueryAppender`?

PR #23566, opened 2026-07-02
Fix/capi scalar bind subquery crash

krleonid:
## Summary

- `BoundSubqueryExpression::Copy()` throws `SerializationException` when a scalar UDF with a `ScalarBinder` receives a correlated subquery as an argument
- The exception escaped the C API boundary, passed through CGo/C frames with no C++ unwind tables, and called `std::terminate()` → `abort()` — uncatchable from Go
- Fix wraps the `Copy()` call in a `try/catch` inside `duckdb_scalar_function_bind_get_argument` (`src/main/capi/scalar_function-c.cpp`), sets the error via `duckdb_scalar_function_bind_set_error`, and returns `nullptr` — surfacing it as a normal query error
- Regression test added in `test/api/capi/capi_scalar_functions.cpp` verifies the query fails gracefully instead of aborting the process

## Test plan

- [ ] `test/api/capi/capi_scalar_functions.cpp` — new case in `Test Scalar Function with Bind Info`: passes a correlated subquery to `get_connection_id` (which uses a `ScalarBinder`) and asserts `REQUIRE_FAIL` instead of process abort
- [ ] All existing scalar function bind tests continue to pass

### DIFF
--- a/src/main/capi/scalar_function-c.cpp
+++ b/src/main/capi/scalar_function-c.cpp
@@ -354,11 +354,13 @@ duckdb_expression duckdb_scalar_function_bind_get_argument(duckdb_bind_info info
 		return nullptr;
 	}
 	auto &bind_info = GetCScalarFunctionBindInfo(info);
+	ExpressionWrapper *wrapper = nullptr;
 	try {
-		auto wrapper = new ExpressionWrapper();
+		wrapper = new ExpressionWrapper();
 		wrapper->expr = bind_info.arguments[index]->Copy();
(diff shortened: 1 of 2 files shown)


Tishj (approved):
Thanks, indeed looks like a missing try/catch in the c++ wrapper

krleonid:
@Tishj @lnkuiper The tests have passed. I've made changes after your review. Please review the code again. I added a smart pointer to ExpressionWrapper to avoid a memory leak in case of an exception.

PR #23570, opened 2026-07-02
`Index` lifetime refactor

sebastiaan-dev:
Resolves https://github.com/duckdblabs/duckdb-internal/issues/9840.

This PR introduces a way to work with `Index` objects which do not have a stable object identity across their logical lifetime. Currently whenever an `Index` object is created external code can reference this index using `Index&`, which is safe since `Index` itself does not change. However with the future introduction of a shadow index a checkpoint will swap in a new `Index` object in place of the previous `Index` object. This causes all outstanding `Index&` references to become stale.

To resolve this lifetime issue we encapsulate index logic where possible, and introduce a handle for cases where encapsulation is not possible.

#### Encapsulation

Currently `Index` specific code is scattered across different domains (`RowGroup`, `DataTable`, etc), exposing the internals of how indexes operate and cluttering functions which should instead delegate these operations. We thus move most of the index iterations into `TableIndexList` functions, and introduce an extended API over `IndexEntry`. These new functions are internally able to work with the `TableIndexList::index_entries_lock` and `IndexEntry::lock` and raw `Index` references, which then exposes these operations through a safe API.

#### Handles

Unfortunately encapsulation is not always possible, as is the case with extensions. We do not know how an extension wants to use their specific `Index` implementation up front, so we cannot define functions statically for them in core. To resolve this, but keep the scope limited in this PR, we introduce handles. There are both `ReadHandle` and `WriteHandle` objects which can be retrieved from an `IndexEntry`. The benefit being that a `GetReadHandle<YourIndex>` requires a cast, thus exposing your custom index operations. As these handles internally acquire a read/write lock, a user can be guaranteed the index will not go stale while holding a live handle.

However, using handles has an important rule while using them: "When holding a live handle to an index, you may not execute operations over the `IndexEntry` of that same index". So you cannot do this:

```c++
{
auto handle = index_entry->GetReadHandle<SomeIndex>(); // <- acquires read lock
... // some code
auto name = index_entry->GetName(); // <- acquires read lock
}
```

The above may deadlock when another thread requests a write lock on the same index in between the read lock acquisitions. Because we use a phase-fair lock no read lock acquisition requests will be satisfied after a write lock request. This means the thread holding the handle cannot continue, while the thread waiting for the write lock is also stuck, thus resulting in a deadlock.

What is possible is doing:

```c++
auto name = index_entry->GetName();
auto handle = index_entry->GetReadHandle<SomeIndex>();
```

Or doing:

```c++
{
    {
        auto handle = index_entry->GetReadHandle<SomeIndex>();
        ... // some code
    } // <- lock released
    auto name = index_entry->GetName();
}
```

In addition, handles cannot completely prevent abuse of the API. A user may define a method on their `Index` which returns a reference to that `Index`, thereby possibly bypassing the lifetime guarantee offered by the handle.

Example:

```c++
BoundIndex &ref;
{
auto handle = index_entry->GetReadHandle<SomeIndex>();
ref = handle->GetReferenceToMyIndex();
} // <- handle goes out of scope, Index lifetime no longer guaranteed
// do something with ref
```

#### Considerations

Handles are imperfect, but they are introduced to prevent scope screep on an already significant PR. I think it might be good to think about if we may actually want to develop a different, more managed interface, for indexes. Something which allows users to 'register' their custom index implementations, similar to how we offer callback hooks into the query lifecycle with for example table functions.


sebastiaan-dev:
Patches will be added after review confirms this is the approach we want to take

artjomPlaunov (commented):
Will be leaving some more comments, just leaving this one about the ResetStorage for now

src/storage/table_index_list.cpp
```diff
@@ -69,41 +69,41 @@ IndexEntry &TableIndexIterationHelper<IndexEntry>::TableIndexIterator::operator*
...
-		auto &index = *index_entries[i]->index;
-		if (index.GetIndexName() == name) {
-			if (!index.IsBound()) {
+		const auto index = index_entries[i]->PinIndex();
+		if (index->GetIndexName() == name) {
+			if (!index->IsBound()) {
 				unbound_count--;
 			}
-			index.ResetStorage();
+			index->ResetStorage();
```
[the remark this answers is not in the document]
artjomPlaunov: (disregard this shouldn't be a problem actually)

src/storage/data_table.cpp
```diff
@@ -846,23 +848,24 @@ void DataTable::VerifyUniqueIndexes(TableIndexList &indexes, optional_ptr<LocalT
...
+		D_ASSERT(index->IsBound());
+		auto &art = index->Cast<ART>();
 		if (storage) {
 			auto delete_index = storage->delete_indexes.Find(art.GetIndexName());
 			D_ASSERT(!delete_index || delete_index->IsBound());
-			manager->AddIndex(art, delete_index);
+			manager->AddIndex(shared_ptr<BoundIndex>(index, &art), delete_index);
 		} else {
-			manager->AddIndex(art, nullptr);
+			manager->AddIndex(shared_ptr<BoundIndex>(index, &art), nullptr);
```
artjomPlaunov: are we able to use the PinIndexCast in these places?

src/common/types/conflict_info.cpp
```diff
@@ -3,15 +3,15 @@
 
 namespace duckdb {
 
-bool ConflictInfo::ConflictTargetMatches(Index &index) const {
-	if (only_check_unique && !index.IsUnique()) {
+bool ConflictInfo::ConflictTargetMatches(const shared_ptr<Index> &index) const {
```
artjomPlaunov: Pass a const reference here instead:

```cpp
 bool ConflictTargetMatches(const Index &index) const;
```

src/storage/table_index_list.cpp
```diff
@@ -229,20 +232,21 @@ void TableIndexList::Bind(ClientContext &context, DataTableInfo &table_info, con
...
 		index_entry->bind_state = IndexBindState::BOUND;
-		index_entry->index = std::move(bound_idx);
+		index_entry->ReplaceIndex(std::move(bound_idx));
 		unbound_count--;
 	}
 }
 
-bool IsForeignKeyIndex(const vector<PhysicalIndex> &fk_keys, Index &index, ForeignKeyType fk_type) {
-	if (fk_type == ForeignKeyType::FK_TYPE_PRIMARY_KEY_TABLE ? !index.IsUnique() : !index.IsForeign()) {
+bool IsForeignKeyIndex(const vector<PhysicalIndex> &fk_keys, const shared_ptr<Index> &index,
```
artjomPlaunov: we can pass a const reference here as well

src/include/duckdb/storage/table/table_index_list.hpp
```diff
@@ -25,22 +25,49 @@ class TableIndexIterationHelper;
...
+
+	//! Replace the current snapshot with another
+	void ReplaceIndex(unique_ptr<Index> index) {
+		lock_guard<mutex> lock(index_pointer_lock);
+		owned_index = std::move(index);
+	}
+
 	atomic<IndexBindState> bind_state;
-	//! lock that should be used if access to "index" and "deleted_rows_in_use" at the same time is necessary
+	//! lock that should be used if access to "owned_index" and "deleted_rows_in_use" at the same time is necessary
```
artjomPlaunov: perhaps this comment is slightly misleading, since the owned_index itself is protected under index_pointer_lock; maybe we can expand a bit more here.

src/execution/operator/persistent/physical_insert.cpp
```diff
@@ -520,30 +520,34 @@ idx_t PhysicalInsert::OnConflictHandling(DuckTableEntry &table, ExecutionContext
...
-	reference_set_t<Index> matching_indexes;
+	vector<shared_ptr<BoundIndex>> matching_indexes;
 
 	if (conflict_info.column_ids.empty()) {
-		auto &global_indexes = data_table.GetDataTableInfo()->GetIndexes();
+		const auto &global_indexes = data_table.GetDataTableInfo()->GetIndexes();
 		// We care about every index that applies to the table if no ON CONFLICT (...) target is given
-		for (auto &index : global_indexes.Indexes()) {
-			if (!index.IsUnique()) {
+		for (auto &index : global_indexes.PinIndexes()) {
```
artjomPlaunov: nit: I think technically here and in some other places we can have const auto& to signify the shared_ptr itself cannot be reassigned (that's independent of modifying the Index I think)

src/include/duckdb/storage/table/table_index_list.hpp
```diff
@@ -25,22 +25,49 @@ class TableIndexIterationHelper;
 enum class IndexBindState : uint8_t { UNBOUND, BINDING, BOUND };
 
 //! IndexEntry contains an atomic in addition to the index to ensure correct binding.
+//! The IndexEntry provides a stable logical identity which refers to an interchangeable snapshot of an index.
 struct IndexEntry {
```
artjomPlaunov: should we consider moving this out into a separate class/following the standard class layout now, maybe, since it has pinning capabilities/some methods defined on it now

artjomPlaunov (commented):
Thanks, looks great to me :-) 

Just some nits/style comments, will ask Tania to take a pass at it as well.

src/storage/write_ahead_log.cpp
```diff
@@ -393,11 +393,11 @@ void SerializeIndex(AttachedDatabase &db, WriteAheadLogSerializer &serializer, T
 		options["v1_0_0_storage"] = v1_0_0_storage;
 	}
 
-	for (auto &index : list.Indexes()) {
-		if (name == index.GetIndexName()) {
+	for (const auto &index : list.PinIndexes()) {
```
artjomPlaunov: I think this might introduce a potential problem here, as PinIndexes returns the indexes, but doesn't hold onto the index_entrieslock, whereas before we were holding onto the lock for the entire loop

src/storage/data_table.cpp
```diff
@@ -1485,8 +1488,8 @@ void DataTable::RevertIndexAppend(TableAppendState &state, DataChunk &chunk, row
 
 void DataTable::RevertIndexAppend(TableAppendState &state, DataChunk &chunk, Vector &row_identifiers) {
 	D_ASSERT(IsMainTable());
-	for (auto &index : info->indexes.Indexes()) {
-		auto &main_index = index.Cast<BoundIndex>();
+	for (const auto &index : info->indexes.PinIndexes()) {
+		auto &main_index = index->Cast<BoundIndex>();
```
artjomPlaunov: Thinking a little ahead to the shadow ART here, but will this function still work correctly if we have to revert an index append, but the ART was swapped? i.e. we pin an old version, ART gets swapped, then we revert from the old instead of the new/current ART?

artjomPlaunov: In general we should be careful about calling a pin + doing some mutation, it probably isn't an issue right now but could become an issue with the shadow ART (so perhaps it just has to be handled in that PR)

src/include/duckdb/storage/table/table_index_list.hpp
```diff
@@ -25,22 +25,49 @@ class TableIndexIterationHelper;
...
 	//! Data that was removed from the index during the last checkpoint
 	unique_ptr<BoundIndex> removed_data_during_checkpoint;
 	//! The last checkpoint index that was written with this index
 	optional_idx last_written_checkpoint;
+
+private:
+	//! The owning pointer of the index
+	shared_ptr<Index> owned_index;
+	//! Lock held when accessing or modifying the owned_index pointer
+	mutable mutex index_pointer_lock;
```
artjomPlaunov: follow up to the comment above: let's say we swap out an owned_index, the index_pointer_lock guards the owned_index, whereas the lock field is protecting other fields (this may be safe depending where you plan on doing the merge for the delta indexes, as we discussed today, so maybe this isn't a problem, but we should be careful here).

artjomPlaunov (commented):
took a longer look, some stuff worth being careful about/making sure nothing breaks here, but also some of it is in anticipation of the shadow ART changes and likely isn't a problem now.

src/storage/write_ahead_log.cpp
artjomPlaunov: Edit: Sorry for notification spam, placed this in the wrong box

Here's a reproducer that causes a SIGSEGV related to this code path (thanks to Claude, we can also add this as a test case):

```sql
# name: test/tmp.test
# description: ART::SerializeToWAL races concurrent index scan
# group: [tmp]
statement ok
SET checkpoint_threshold = '1TB';

statement ok
ATTACH '{TEST_DIR}/h1_index_wal_scan_race.db' AS db (STORAGE_VERSION 'v1.0.0');

statement ok

statement ok
CREATE TABLE db.t AS SELECT range AS i, range % 100 AS j FROM range(200000);

concurrentloop threadid 0 8

loop i 0 60

onlyif threadid=0
statement ok
CREATE INDEX idx ON db.t(j);

onlyif threadid=0
statement ok
DROP INDEX db.idx;

loop k 0 100

skipif threadid=0
query I
SELECT count(*) FROM db.t WHERE j = 42;
----
2000

endloop

endloop

endloop

statement ok
DETACH db
```
Previously the lock was held for all the SerializeToWAL calls as mentioned, maybe we just need to move a lock into SerializeToWAL?

src/include/duckdb/storage/table/table_index_list.hpp
artjomPlaunov: edit: this might not be an issue since in a lot of contexts where I thought it could be (merge checkpoint deltas + scans, we still grab the index_entries_lock)

extension/core_functions/scalar/debug/index_key.cpp
```diff
@@ -99,17 +99,17 @@ static string GetStringArgument(ClientContext &context, const Expression &expr,
...
 	if (found) {
-		return *found;
+		return found;
 	}
 
 	auto qualified_table = path.qualified_name.ToString(QualifiedNameToStringMode::HIDE_DEFAULT_SCHEMA);
 	vector<Identifier> available;
-	for (auto &idx : index_list.Indexes()) {
-		available.push_back(idx.GetIndexName());
+	for (const auto &idx : index_list.PinIndexes()) {
```
taniabogatsch: Two things here: `PinIndexes` is a bit of a misnomer, because what the function actually does is to return a shared copy of all indexes. So, this function should probably be called something like `index_list.MakeShared()` or `index_list.SharedCopy)`

That being said, if we just want to iterate that index list to extract all the names, then I don't think we need to do a full ownership copy, only to iterate over that copy, and then throw it away again. Why can we not stick with the existing `Indexes` iterator, which acquires a lock through `index_entries_lock` while the iterator is alive?

If we ensure that the same lock needs to be held for making any changes to that index list.

```cpp
TableIndexIterationHelper<Index> TableIndexList::Indexes() const {
	return TableIndexIterationHelper<Index>(index_entries_lock, index_entries);
}
```

extension/core_functions/scalar/debug/index_key.cpp
```diff
@@ -131,10 +132,10 @@ struct IndexKeyBindData : public FunctionData {
 
 	bool Equals(const FunctionData &other_p) const override {
 		auto &other = other_p.Cast<IndexKeyBindData>();
-		return &art == &other.art && key_types == other.key_types;
+		return art.get() == other.art.get() && key_types == other.key_types;
 	}
 
-	ART &art;
+	shared_ptr<ART> art;
```
taniabogatsch: Maybe we can add a note here that the index key scalar function needs to keep the index alive to, e.g., scan its keys, so it has to be able to be the last owner. And could we add a test here? I think this is a cool example, have one connection run this function (very long), so its reading, while another throws away the table (including the index)?

extension/core_functions/scalar/debug/index_key.cpp
```diff
@@ -192,7 +192,8 @@ static unique_ptr<FunctionData> IndexKeyBind(BindScalarFunctionInput &input) {
 		bound_function.GetArguments().push_back(key_type);
 	}
 
-	return make_uniq<IndexKeyBindData>(art, std::move(key_types));
+	auto art = PinIndexCast<ART>(bound_index);
```
taniabogatsch: Here, I'd also say we need to rename `PinIndexCast` to something like `MakeShared`  - casting is implied because we pass the type in the template, I think. Also, this function should not be global. Can it be a member function of the `bound_index`?

src/execution/operator/persistent/physical_insert.cpp
```diff
@@ -520,30 +520,34 @@ idx_t PhysicalInsert::OnConflictHandling(DuckTableEntry &table, ExecutionContext
 	}
 
 	ConflictInfo conflict_info(conflict_target);
-	reference_set_t<Index> matching_indexes;
+	vector<shared_ptr<BoundIndex>> matching_indexes;
```
taniabogatsch: While looking at the code here I was wondering:

The idea is that we keep the view on the "old" index alive with the shadow ART, right? While the indexes in the table list move forward, right? Because if we don't always point to the latest committed changes in the table index list (i.e., if the shadow ART were the **new** index), then how would we know what to point constraint checking like we do here against?

src/execution/operator/persistent/physical_insert.cpp
```diff
@@ -520,30 +520,34 @@ idx_t PhysicalInsert::OnConflictHandling(DuckTableEntry &table, ExecutionContext
...
 			}
-			D_ASSERT(index.IsBound());
-			if (conflict_info.ConflictTargetMatches(index)) {
-				matching_indexes.insert(index);
+			if (!conflict_info.ConflictTargetMatches(*index)) {
+				continue;
 			}
+			D_ASSERT(index->IsBound());
+
+			matching_indexes.push_back(PinIndexCast<BoundIndex>(index));
```
taniabogatsch: If we make the shared copy only here, then we can just use the iterator above (instead of `PinIndexes`) instead of copying twice. Same for the changes below.

src/execution/operator/schema/physical_create_index.cpp
```diff
@@ -160,9 +160,9 @@ SinkFinalizeType PhysicalCreateIndex::Finalize(Pipeline &pipeline, Event &event,
 
 	} else {
 		// Ensure that there are no other indexes with that name on this table.
-		auto &indexes = storage.GetDataTableInfo()->GetIndexes();
-		for (auto &index : indexes.Indexes()) {
-			if (index.GetIndexName() == info->GetIndexName()) {
+		const auto &indexes = storage.GetDataTableInfo()->GetIndexes();
+		for (const auto &index : indexes.PinIndexes()) {
```
taniabogatsch: No need to make a shared copy, just iterate behind the lock.

src/storage/write_ahead_log.cpp
sebastiaan-dev: I think a lock in `SerializeToWAL` is good, since for storage version `v1.0.0` it will transform the in-memory representation to the deprecated layout. But this might not be sufficient, taking another look.

src/execution/index/art/art.cpp
```diff
@@ -1054,6 +1054,8 @@ IndexStorageInfo ART::SerializeToDisk(QueryContext context, const case_insensiti
 }
 
 IndexStorageInfo ART::SerializeToWAL(const case_insensitive_map_t<Value> &options) {
+	lock_guard<mutex> guard(lock);
```
taniabogatsch: Do we have a test/comment for why this was added?

taniabogatsch (commented):
Thanks for the changes! I think introducing explicit lifetime ownership to the indexes is a great idea. 

I've left some comments, maybe we can get them implemented and then I'll take another look. Some of them might align with what Artjom pointed out regarding when/how we would use/swap the shadow ART.

taniabogatsch:
I think what currently is a misconception in some sense is that we're creating multiple owners. We're not actually pinning anything (and I don't think we want to). Rather, we DO allow the indexes to be moved out of the table index list and **only** be kept alive by another owner.

src/execution/index/art/art.cpp
artjomPlaunov: @sebastiaan-dev maybe Tania's comment above about using the iteration helper + index_entries_lock can be applied here as well to get rid of the regression/not need to introduce this fix?

src/storage/data_table.cpp
sebastiaan-dev: I am looking into this, but I realized there is another bug (unrelated to the changes in this PR).

`AppendToIndexes` mutates either the main index or a delta-index based on if a checkpoint is active, but if something fails then the rollback will only be performed on the main index in `RevertIndexAppend`. This means that anything added to the delta index will still get merged into the main index at `MergeCheckpointDeltas`. This will then result in stale indexes being present in the main index, and you can get constraint violations which are invalid.

src/include/duckdb/storage/table/table_index_list.hpp
artjomPlaunov: edit: actually the comment was fine, but can we double check if we want to expand the comment to state that it is synchronizing the owned_index as well as the all delta indexes, not just deleted_rows_in_use, or is it only used for this scenario?

src/storage/data_table.cpp
sebastiaan-dev: Issue here: https://github.com/duckdblabs/duckdb-internal/issues/9893

artjomPlaunov: good find!

src/include/duckdb/storage/table/table_index_list.hpp
sebastiaan-dev: Updated the comment

sebastiaan-dev:
I think that another point of confusion here is that the `index_entries_lock` actually takes on the role of the `IndexEntry::lock` in some places. I would think the clean model is the following:

- `index_entries_lock`: Responsible for keeping the table index list stable, so no conflicting additions/removals/moves can happen when this lock is held.
- `IndexEntry::lock`: Protects mutations to the main index and the delta indexes owned by the entry.
- `IndexLock`: Protects the internal representation of a given index, for example the allocators and buffers of the `ART`

I think as future work it would be good to separate the usage of `index_entries_lock` and `IndexEntry::lock` into their own domains. This would potentially also allow interleaving of index accesses, given we model `index_entries_lock` with read and write usage.

src/execution/operator/persistent/physical_insert.cpp
sebastiaan-dev: The index stored in the `TableIndexList` is always authoritative. When we construct the shadow index, we base it on the index currently stored in the `IndexEntry`, so their content is the same.

If changes happen during a checkpoint, they target the logical index identity represented by the `IndexEntry`. Those changes are written to the delta indexes owned by the entry, not to the shadow index. These delta indexes remain untouched by the swap operation.

After the shadow index is swapped into the `IndexEntry`, the delta indexes are merged into that current index, which is now the former shadow index. At that point the checkpoint can finish.

src/include/duckdb/storage/table/table_index_list.hpp
sebastiaan-dev: As discussed not an issue currently due to `index_entries_lock`

src/storage/data_table.cpp
sebastiaan-dev: To conclude, the shadow ART will not introduce incorrect behavior because it should be managed via the delta indexes during a checkpoint. But for this to work the previous issue should be resolved first.

src/execution/index/art/art.cpp
sebastiaan-dev: No longer necessary

extension/core_functions/scalar/debug/index_key.cpp
sebastiaan-dev: Renamed

src/storage/write_ahead_log.cpp
sebastiaan-dev: Restored iteration with `index_entries_lock` and removed the lock in `SerializeToWAL`

src/execution/operator/persistent/physical_insert.cpp
sebastiaan-dev: Changed

extension/core_functions/scalar/debug/index_key.cpp
sebastiaan-dev: I've tried making a test for this function, but could not get it to fail in the before case where this was a reference. I think this is due to the fact that currently the `index_key` function does not really use any internal state of the `ART` object, it only executes `GenerateKeys` but this does not touch internal state of the `ART`.

taniabogatsch: Ah yes, then that makes sense, if there is no internal state. 👍

taniabogatsch:
> `index_entries_lock`: Responsible for keeping the table index list stable, so no conflicting additions/removals/moves can happen when this lock is held.
`IndexEntry::lock`: Protects mutations to the main index and the delta indexes owned by the entry.
`IndexLock`: Protects the internal representation of a given index, for example the allocators and buffers of the ART

Yes, this sounds very sensible. So in this PR we might still be a bit over-eager with grabbing the higher level `index_entries_lock`, but we could make that more fine-grained in follow-up work?

src/execution/operator/persistent/physical_insert.cpp
```diff
@@ -520,30 +520,40 @@ idx_t PhysicalInsert::OnConflictHandling(DuckTableEntry &table, ExecutionContext
...
+			D_ASSERT(index.IsBound());
+
+			matching_indexes.push_back(entry.GetSharedIndex<BoundIndex>());
 		}
-		auto &local_indexes = local_storage.GetIndexes(context.client, data_table);
-		for (auto &index : local_indexes.Indexes()) {
+
+		const auto &local_indexes = local_storage.GetIndexes(context.client, data_table);
+		for (auto &entry : local_indexes.IndexEntries()) {
+			lock_guard<mutex> lock(entry.lock);
```
taniabogatsch: Nice, so to confirm: We lock the table index list against modifications in the iterator (`for ...`), and then lock the index itself against modifications with `entry.lock`?

src/storage/data_table.cpp
```diff
@@ -848,23 +850,24 @@ void DataTable::VerifyUniqueIndexes(TableIndexList &indexes, optional_ptr<LocalT
 
 	// The conflict manager is only provided for statements containing ON CONFLICT.
 	auto &conflict_info = manager->GetConflictInfo();
+	const auto pinned_indexes = indexes.MakeShared();
```
taniabogatsch: Sorry for being so nitpick-y about the "pin" word. Can we use `snapshot_indexes` or similar here (and below)?

taniabogatsch (approved):

src/storage/data_table.cpp
```diff
@@ -649,9 +650,10 @@ idx_t LocateErrorIndex(ConflictManager &manager, const bool is_append, const idx
 	return FirstMissingMatch(manager, count);
 }
 
-static string ConstructForeignKeyError(optional_idx conflict, bool is_append, Index &index, DataChunk &input) {
-	D_ASSERT(index.IsBound());
-	auto &bound_index = index.Cast<BoundIndex>();
+static string ConstructForeignKeyError(optional_idx conflict, bool is_append, const shared_ptr<Index> &index,
```
Mytherin: Do we really need this to get a `shared_ptr<Index>`? maybe we can move this method into the IndexEntry instead?

Mytherin (commented):
Thanks for the PR!

I think moving to making the `IndexEntry` the stable index is a great idea, also to ensure everything goes through the proper locks. However, I would propose trying to change the code so that we never leak the internal `Index` and instead have everything go through the IndexEntry.

I think right now we have a bit of an odd situation where we have locks on both the `IndexEntry` and internally in the `Index`. There's also some locking ugliness because the lock in `IndexEntry` and all members there are public - so none of the lifetimes are nicely guarded by the type system and instead the code must just do the right thing (which is in practice usually not what happens everywhere...).

Maybe we can make `shared_ptr<IndexEntry>` the "share-able" state, keeping the index a `unique_ptr<Index>` and only allowing access to the index through members on the `IndexEntry`? i.e. we wouldn't introduce the `GetSharedIndex()` and `MakeShared()` methods that are proposed here, and instead replace `shared_ptr<Index>` with `shared_ptr<IndexEntry>` where that makes sense.

Mytherin:
Thanks for the fixes! I think this is looking better - but it seems there's still a lot of machinery to allow code to live outside of the index structures - i.e. we're still leaking internals of the index structure instead of containing the complexity of indexes into their relevant classes. For example:

```cpp
for (auto index : indexes.MutableIndexHandles()) {
	if (!index->IsBound()) {
		auto unbound_index = std::move(index).Into<UnboundIndex>();
		unbound_index->BufferChunk(chunk, row_id_vector, BufferedIndexReplay::INSERT_ENTRY);
		continue;
	}
	auto &bound_index = index.Cast<BoundIndex>();
	bound_index.Append(chunk, row_id_vector);
	auto bound_index = std::move(index).Into<BoundIndex>();
	bound_index->Append(chunk, row_id_vector);
}
```

This code lives in `WriteAheadLogDeserializer::ReplayRowGroupData` - but it's basically all index-specific code.

Couldn't we move this to a method like `TableIndexList::Append(DataChunk &, Vector &)`?

If we could do this to as many places as possible then that would mean moving index specific code into the actual index / index lists which would keep the code contained (and subsequently easier to change / debug / reason about / etc). We should then be able to remove all of the various "leaky internals" from anywhere but the index code itself (i.e. remove `MutableIndexHandles`, `std::move(...).Into<BoundIndex>`, etc).

Mytherin:
Some related concerns from the clanker about the new code leaking locks now and holding onto them for too long / potentially introducing deadlocks:

```


2. row_group_collection.cpp:1335 — vacuum holds exclusive index locks for the whole task. VacuumIndexRemapper's constructor now takes                              
GetMutableHandle<BoundIndex>() on every index and holds them across the entire scan/merge loop plus checkpoint_task->ExecuteTask(). Previously the lock was taken  
only inside Apply(). Parallel VacuumTasks on a table now fully serialize, and MutableIndexHandles() blocks on entries while holding index_entries_lock. Take the   
handles inside Apply().

3. Extension patches — long-lived shared handles can deadlock. vss/0011-index-lifetime.patch:228,236 (+ hnsw_optimize_join.cpp) and
spatial/0004-index-lifetime.patch:86,95 keep an IndexHandle alive for the whole operator/query. StorageLock's shared acquisition is phase-fair, so once a writer
registers a ticket, later readers queue behind it. Concrete case: INSERT INTO t SELECT ... ORDER BY array_distance(...) LIMIT k on a table with both an HNSW index
and a PK — the scan's global state holds the shared handle, the insert's VerifyUniqueIndexes registers an exclusive waiter on that same entry, and neither side can
drain.
```

src/storage/table/row_group_collection.cpp
```diff
@@ -1347,10 +1171,15 @@ class CheckpointTask : public BaseCheckpointTask {
...
-	VacuumIndexRemapper(const vector<reference<BoundIndex>> &indexes, RowGroupCollection &collection)
-	    : indexes(indexes), table_types(collection.GetTypes()) {
+	VacuumIndexRemapper(const vector<shared_ptr<IndexEntry>> &entries, RowGroupCollection &collection)
+	    : table_types(collection.GetTypes()) {
 		DataChunk index_chunk;
 		TableIndexList::InitializeIndexChunk(index_chunk, table_types, mapped_column_ids, collection.GetTableInfo());
+
+		index_handles.reserve(entries.size());
+		for (const auto &entry : entries) {
+			index_handles.push_back(entry->GetMutableHandle<BoundIndex>());
```
artjomPlaunov: For the vacuum path, I think there are two overlapping things going on here that could point to a solution:

The checkpoint + vacuum path is already establishing a coarse scope with the checkpoint, vacuum, and table checkpoint locks. These exclude external operations that can add/remove indexes or modify the physical indexes. CanVacuumRemap() only selects already bound indexes, and the proposed shadow swap will occur only after the vacuum is complete and all checkpoint tasks are complete. Thus the selected entries + physical indexes are guaranteed to remain stable during this phase. 

The idea with the MutableIndexHandle is to provide a narrower per-entry scope/guarantee about the index lifetime. It prevents swapping out the physical index, and serializes mutable accesses to the entry. Acquiring handles for every entry in the remapper is therefore "duplicating" some of the guarantees we already have along this code path, while also introducing an exclusive entry lock on every single index, thus serializing all the remap tasks (the issue Mark pointed out above). 

Maybe we can just treat this code path as already being "safe" against external changes? A simple fix could then be to just store shared_ptr\<IndexEntry\> in the remapper, and acquire a MutableIndexHandle once at a time inside of Apply(), which would still make sure we serialize concurrent mutable operations on a given index entry, and avoid holding the entry locks for the entire vacuum task.

artjomPlaunov (changes requested):
I still need to look through the remaining changes when you mark it as ready to review, but I like the overall direction, especially moving more involved index logic out of so many code paths :-) Just leaving an idea about the vacuuming issue.

extension/core_functions/scalar/debug/index_key.cpp
```diff
@@ -98,19 +98,20 @@ static BoundIndex &FindBoundIndex(TableIndexList &index_list, const Identifier &
...
 	}
 
 	bool Equals(const FunctionData &other_p) const override {
 		auto &other = other_p.Cast<IndexKeyBindData>();
-		return &art == &other.art && key_types == other.key_types;
+		return index_entry.get() == other.index_entry.get() && key_types == other.key_types;
 	}
 
-	ART &art;
+	shared_ptr<IndexEntry> index_entry;
```
artjomPlaunov, line 114:
```sql
# name: test/tmp.test
# description: Cached index_key bind data must not outlive detached database storage
# group: [tmp]

statement ok
ATTACH '{TEST_DIR}/pr23570_index_key_detach.db' AS aux;

statement ok
CREATE TABLE aux.t(a INTEGER);

statement ok
CREATE INDEX idx ON aux.t(a);

statement ok
INSERT INTO aux.t VALUES (1), (2), (3);

statement ok
CHECKPOINT aux;

statement ok
PREPARE s AS
SELECT index_key({'catalog': 'aux', 'table': 't'}, 'idx',
                 CAST(random() * 0 AS INTEGER) + 42);
statement ok
DETACH aux;

statement ok
DEALLOCATE s;
```

This causes a SIGSEGV, looks like there is a mismatch between the physical index and storage lifetimes. here, i.e. DETACH will destroy database storage, but the prepared statement still owns IndexEntry + ART. This is kind of an edgy edge case found by codex, but this might pertain to extensions as well. 

One possibility for index_key could possibly be that it doesn't need to store the index, but rather maybe just the index key generation logic (in which case we could just do a follow up PR for this and not worry about this right now). 

But if this brings up a pattern that might also come up in extensions, maybe we need some extra explicit state added to IndexEntries to make sure we don't run into similar problems (again might just be a follow up PR we can do though). 

```cpp
void IndexEntry::Retire() {
    auto lock = GetExclusiveLock();

    deltas.Reset();
    owned_index.reset(); // BlockManager must still be alive here
    state = IndexEntryState::RETIRED;
}
```

artjomPlaunov (changes requested):
Looks good! Nice to see a lot of index specific code pulled out to clean up some critical code paths :-) 

There was one crash that codex found, it's possible this issue might come up with extensions since holding onto an IndexEntry reference may extend the index lifetime without accounting for underlying storage lifetime.

artjomPlaunov:
I am thinking that the index API doesn't neatly support the type of lifetime management for extension indexes we are trying to design here at the moment. After talking to @sebastiaan-dev, one possible idea here is to just defer extension index lifetime concerns by limiting index swaps to ART indexes (for the shadow ART + colocation feature), and still get a lot of the benefits of the core index lifetime refactor from this PR vis a vis IndexEntry and read/write locks, which I think is still a very worthwhile addition.

If we choose to go with this approach, we could also probably remove the handle abstractions and just rely on the read/write locks for IndexEntry, further simplifying the code a bit.

extension/core_functions/scalar/debug/index_key.cpp
sebastiaan-dev: This has been added!

src/storage/table/row_group_collection.cpp
sebastiaan-dev: Currently a method which is encapsulated and that can be called from `Apply()`

artjomPlaunov:
To expand on my last point a bit, it seems to me that the handles are solving a very specific problem: safely swapping the physical index contained in a logical IndexEntry. In core, this is already handled by encapsulating index operations behind the IndexEntry read/write lock. For extensions, because we cannot encapsulate their arbitrary index-specific operations in core, this PR introduces new handle types as an escape hatch.

I am wondering whether supporting this specific lifetime requirement is worth introducing an API with a deadlock footgun. As I understand it, the main tradeoff is between making extension indexes compatible with the physical index replacement required by the future colocation feature, and exposing an API whose correct use depends on a fairly subtle locking rule. 

It seems to me like we could instead keep extension indexes opted out for now and add a more compatible lifetime Index API (or something of the sort) in a subsequent PR if extensions need to handle this lifetime support/there are a lot more scenarios where we are doing these sort of physical index swaps.

artjomPlaunov (approved):
I will say this is good to go, albeit with the caveat that it would be nice if we could revisit the Index API with @Maxxen and see how we want to integrate these lifetime considerations more neatly within the Index API so that we can support index vacuuming + shadow swapping for colocation while also ideally getting rid of this deadlock concern in the introduced handle API. 

In general the cleanest solution in core has been to just kick the can down the road with extension indexes, which is also why I think I'm okay with introducing this maybe imperfect solution for extension index lifetimes, so we can start improving on how we integrate extension indexes with core optimizations.

Mytherin:
Thanks!

PR #23572, opened 2026-07-02
unittest: emit event w/ stats for harnesses (pytest)

benfleis:
- emits (stderr) "[TEST_EVENT] {...}" tiny json event with name counts and message
- opt-in via --emit-test-stats
- allows pytest harness to recognize and count tests/fails/skips correctly

This could eventually be superseded by a structured reporter, e.g. custom json via Catch.

### DIFF
--- a/test/helpers/test_helpers.cpp
+++ b/test/helpers/test_helpers.cpp
@@ -127,6 +127,16 @@ bool DeleteTestPath() {
 	return delete_test_path;
 }
 
+static bool emit_test_events = false;
+
+void SetEmitTestEvents(bool emit) {
+	emit_test_events = emit;
+}
+
+bool EmitTestEventsEnabled() {
+	return emit_test_events;
+}
+
 void ClearTestDirectory() {
 	if (!DeleteTestPath()) {
 		return;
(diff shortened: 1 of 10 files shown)


lnkuiper:
I had Codex look into this and it found two issues, let me know if you think they are serious enough:

1. Concurrent Loop Miscounts Failure As Pass
```
build/reldebug/test/unittest --emit-test-events --stdin <<'EOF'
# name: <stdin>
# description: concurrent event count repro
# group: [temp]

concurrentloop i 0 1

query I
SELECT 1
----
2

endloop
EOF
```

  2. Cleanup Failure Drops Terminal Event
```
build/reldebug/test/unittest --emit-test-events --stdin --on-cleanup "SELECT error('cleanup')" <<'EOF'
# name: <stdin>
# description: cleanup terminal event repro
# group: [temp]

query I
SELECT 1
----
1
EOF
```

benfleis:
@lnkuiper the first one indeed, the second one ... unclear. I fixed and employed Claude for a more intense review as well, found a couple more cleanups, noting them here:

- add {only/skip}if support
- add consistent fail locator
- emit events lines atomically

PR #23574, opened 2026-07-02
unittest: support harnessed (unit)test: add controls for TEMP_DIR (and related)

benfleis:
unittest: support harnessed (unit)test: add controls for TEMP_DIR (and related)

Give the unittest binary an explicit, composable temp-dir contract so an external
harness (the pytest driver) can own the temp-dir lifecycle and place tests
deterministically, replacing the bundled side-effects of --test-temp-dir.
This specifically support a pytest based driver, as well as extended testing
of e.g. Azure storage.

Previous --test-temp-dir has 1 specific policy; avoid adding another specific-
use-case knob, and instead expose the policy knobs needed to be used by a harness.

Structure — the temp dir is (up to) three nested levels, $BASE/[$RUN_ID]/[$TEST_ID]:

```
--temp-dir-base <path>       $BASE (env TEMP_DIR_BASE); local or remote (s3://, ...)
--run-id {auto|<id>}         run identity -> RUN_ID env (always set); auto = generate
--temp-dir-run-id {on|off}   is RUN_ID a path level? (default on)
--temp-dir-test-id {on|off}  is TEST_ID (the test name) a path level? (default on)
```

--run-id (identity) is split from --temp-dir-run-id (path inclusion): the RUN_ID value
is always resolved and exported, whether or not it is a path segment. A standalone run
generates one (timestamp + mnemonic, UTC); a batched harness passes one shared id so a
run's invocations co-locate.

Lifecycle dispositions, inherited down every level:

```
--temp-dir-create  {never|on-absent|always}   (default on-absent: mkdir -p, no clobber)
--temp-dir-destroy {never|on-success|always}  (default on-success: keep on failure)
```

Destroy is recursive, bottom-up, and removes only what this invocation created, stopping at
the first pre-existing/non-empty level -- so $BASE is reclaimed only when the run owns it and
it's empty. A remote $BASE clamps create/destroy to never (the test owns materialization).

Default `unittest` (no flags) is now duckdb_unittest_tempdir/$RUN_ID/$TEST_ID
(on-absent/on-success): each test gets its own dir, a failed run's dirs are retained,
replacing the shared per-pid dir wiped between tests.

Env vars exported to tests (UpdateEnvironment() now runs after the temp dir is prepared, so
all reflect the final context):
- RUN_ID           run identity; always set, a path level iff --temp-dir-run-id on
- TEST_ID          test name -> one path component ([^A-Za-z0-9_-] -> _, incl '.'); suffix
                    kept, so foo.test / foo.test_slow don't collide
- TEMP_DIR         per-test path $BASE/[RUN_ID]/[TEST_ID]; == {TEST_DIR}
- TEMP_DIR_BASE    the base $BASE
- TEMP_DIR_ABSOLUTE  absolute form of TEMP_DIR
- DATA_DIR         fixture root (see below)
- CATALOG_DIR      alias of TEST_DIR, kept for back-compat until callers move off it

Fixes/tweaks alongside the core knobs:
- HOME: one isolated home sandbox per invocation ("<run-root>-home", a sibling of the run
  root, set once), so ~/.duckdb (extensions, secrets) is isolated yet never lands inside a
  {TEST_DIR} that tests whitelist via allowed_directories -- which would defeat permission
  tests (e.g. INSTALL-is-denied). Materialized in PrepareTempDir, reclaimed in DestroyTempDir.
- DATA_DIR is now first-class: --data-dir / DUCKDB_TEST_DATA_DIR / `data_dir` config (default
  {WORKING_DIR}/data; relative, absolute, or remote e.g. az://). Resolved fresh each time so an
  explicit value survives cwd changes. Also fixes a latent bug where any test_env override of
  a standard var was reverted to its default by the first cwd-changing test in a run.
- Out-of-tree extension tests chdir mid-run; for those only, {TEST_DIR}/TEMP_DIR are pinned to
  the absolute path (captured pre-chdir) so they don't resolve against the extension dir.
  Everything else stays relative on purpose (no machine paths in compared output).

Removed --test-temp-dir: its must-not-pre-exist guard has no clean disposition equivalent, and
--temp-dir-base + --temp-dir-destroy never covers its use. In-tree callers
(scripts/test_zero_initialize.py, scripts/test_storage_compatibility.py) migrated to the family
flags; the dead custom_test_directory / SetTestDirectory excised.


test/helpers/test_helpers.cpp
```diff
@@ -92,29 +110,365 @@ bool IsRequired(string require) {
...
+	}
+	fs.CreateDirectoriesRecursive(path);
+	std::reverse(to_create.begin(), to_create.end()); // outermost..leaf
+	created = to_create;
+}
+
+// Recursive bottom-up reclaim: remove `leaf` (and its contents), then walk up `created`
+// removing each ancestor iff it is empty and this step created it; stop at the first
+// non-empty / not-this-step-created level. Shared by both destroy lifecycle steps.
+static void ReclaimLevels(FileSystem &fs, const string &leaf, const vector<string> &created) {
```
lnkuiper, line 367: This removes `leaf` recursively before checking whether the current invocation created it. With `--temp-dir-create never`, an existing harness-owned base records no created levels, but `DestroyTempDir()` still deletes the whole root on success/default destroy. I confirmed this by pre-creating a temp base with a sentinel file and running with `--temp-dir-create never --temp-dir-destroy always`; the pre-existing directory was removed. This should only remove levels recorded in `created`.

test/sqlite/sqllogic_test_runner.cpp
lnkuiper: This still deletes loaded databases under `TestDirectoryPath()` even when the caller requested `--temp-dir-destroy never`. The old `--test-temp-dir` path avoided this through `DeleteDatabase()` being a no-op for custom temp dirs, but that guard is gone. This breaks migrated callers like `test_zero_initialize.py`, which retain the temp dir specifically to compare the generated DB files after the unittest process exits. A minimal `load {TEST_DIR}/retain_me.db` test passes but leaves the retained temp directory empty.

lnkuiper (commented):
Thanks for the PR! Codex found two issues related to the cleanup:

test/helpers/test_helpers.cpp
benfleis: This is by design -- if you set policy always, you get always. The harness in some cases will want this behavior for guaranteed cleanup.

I considered a mechanism like you describe: tracking the created subtrees, but that would be a different policy, not "always". Is there a material value in a `--temp-dir-delete=created`? It also suffers from the additional challenge of being orthogonal to `on-success`, because you might want `on-success | created`. I aimed for KISS policy here.

test/sqlite/sqllogic_test_runner.cpp
benfleis: That's a miss, I'll have a look!

test/helpers/test_helpers.cpp
lnkuiper: Makes sense!

test/sqlite/sqllogic_test_runner.cpp
benfleis: Agreed offline to add an explicit flag for this case, as DB removal is not same as dir removal. Update coming shortly.

benfleis: Manual test w/ update:

```
0:43:51 @puck d/d/local/manage-temp-dirs !!> python3 scripts/test_zero_initialize.py --unittest build/relassert/test/unittest
Running test test/sql/index/art/storage/test_art_checkpoint.test
Running test in one-initialize mode
Running test in zero-initialize mode
Success!
Running test test/sql/storage/compression/simple_compression.test
Running test in one-initialize mode
Running test in zero-initialize mode
Success!
Running test test/sql/storage/delete/test_store_deletes.test
Running test in one-initialize mode
Running test in zero-initialize mode
Success!
Running test test/sql/storage/mix/test_update_delete_string.test
Running test in one-initialize mode
Running test in zero-initialize mode
Success!
Running test test/sql/storage/nested/struct_of_lists_unaligned.test
Running test in one-initialize mode
Running test in zero-initialize mode
Success!
Running test test/sql/storage/test_store_integers.test
Running test in one-initialize mode
Running test in zero-initialize mode
Success!
Running test test/sql/storage/test_store_nulls_strings.test
Running test in one-initialize mode
Running test in zero-initialize mode
Success!
Running test test/sql/storage/update/test_store_null_updates.test
Running test in one-initialize mode
Running test in zero-initialize mode
Success!
```

lnkuiper:
Thanks!

PR #23578, opened 2026-07-02
Heap-based transformer (Trampoline) script

Dtenwolde:

The large diff is almost entirely thanks to the script used to generate the heap-based transformer rules. The logic for the heap-based implementation can be found in https://github.com/duckdb/duckdb/pull/23414, so you won't find it in this diff. Here we have a couple of grammar changes to accomodate the script. The script itself can probably still be further improved to include more and more complex grammar-syntax, but I'll leave that for future work. The script currently supports all the grammar.   

This is the second step for the trampoline transformer after https://github.com/duckdb/duckdb/pull/23414. This PR features a script `generate_transformer_trampoline.py` that generates the trampoline style `Initialize/Finalize` methods used by this transformer. 

I added an extended test variation `transformer_trampoline_style` that runs all the test with `SET transformer_trampoline_style=true`. 

If this passes, the follow-up PR will be switching the default transformer to be the trampoline style (and thus finally fixing `duckdb_functions()` in debug mode).

Another follow-up PR is to add a CI check that verifies that the grammar is properly updated using a diff-checker, similar to `make generate-files` or `make format-fix`. 

There's a couple of transformer rules that require to be manual still: 
  - LiteralExpression: specialized literal-to-expression construction.
  - PrefixExpression: operator-specific expression construction.
  - OverClause: mixed named/inline window specification handling.
  - PivotStatement: statement-level pivot state while transforming children.
  - UnpivotStatement: statement-level unpivot state while transforming children.
  - SelectStatementInternal: query assembly across WITH, set ops, and select/value forms.
  - SimpleSelect: clause ordering and select-core assembly.
  - WithClause: CTE metadata and body transformation.
  - TableRef: table-ref dispatch plus join/context handling.
  - WindowDefinition: inherited window, partition/order, and frame assembly.

Some of these can still be generated if we move some logic to the `Binder`, such as resolving parameters and checking nested window clauses. But this is not in the scope of this PR.


lnkuiper:
Thanks!

PR #23582, opened 2026-07-03
Propagate client context for stats fetch

### LINKED ISSUE #23581: Timestamp predicate not always pruned
### What happens?

Hi team, I found for some semantically equivalent queries show different row group access and performance, see repro example for details, should be straightforward.

### To Reproduce

```sql
memory D ATTACH '/tmp/big3.db' AS big3;
memory D USE big3;
big3 D CREATE TABLE events3 AS
SELECT (TIMESTAMP '2020-01-01' + to_milliseconds(range*200)) AS ts, range::INT AS v
FROM range(500000000);
big3 D CHECKPOINT;
┌─────────┐
│ Success │
│ boolean │
└─────────┘
0 rows

-- Close the connection, reattach DuckDB file and perform queries.
memory D ATTACH '/tmp/big3.db' AS big3;
memory D USE big3;

-- Timestamp range query shows correct row group pruning -- only two row groups accessed.
big3 D EXPLAIN ANALYZE SELECT count(v) FROM events3 WHERE ts >= '2022-01-01'AND ts < '2023-01-01';
╭─ Summary ───────────╮
│ Total Time: 1.23s │
│ Data Read: 524.2 kB │
╰─────────────────────╯
╭─ Projection ─────────────────────────────────────╮
│ Projections: "+"(157655040, #0) │
│ 1 row 0µs │
╭─ Ungrouped Aggregate ───┴────────────────────────╮
│ Aggregates: count_star() │
│ 1 row 0µs │
│ Table: big3.main.events3 │
│ Type: Sequential Scan │
│ Projections: v │
│ Filters: │
│ (ts >= '2022-01-01 00:00:00'::TIMESTAMP) AND (ts │
│ < '2023-01-01 00:00:00'::TIMESTAMP) │
│ Row Groups Scanned: 2 / 4070 │
│ 24,960 rows 0µs │

type .web to open the query profile in a browser

-- The equivalent query using year function accesses far more row groups.
big3 D EXPLAIN ANALYZE SELECT count(v) FROM events3 WHERE year(ts) = 2022;
╭─ Summary ───────────╮
│ Total Time: 0.0143s │
│ Data Read: 2.8 MB │
╰─────────────────────╯
│ 1 row 0µs │
│ Table: big3.main.events3 │
│ Type: Sequential Scan │
│ Projections: v │
│ Filters: "year"(ts) = 2022 │
│ Row Groups Scanned: 1285 / 4070 │
│ 157,680,000 rows 160.0ms │
```
Both queries produce the correct result, which means the second query doesn't leverage stats and prune row groups properly; for this query, only two row groups, where data in year 2022 is stored together with year 2021 and year 2023 (in total two of them) should be accessed.

### OS:

macos

### DuckDB Version:

a81547b43c56c671a7ad1a636f83886713efdd8a

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23581

When getting filter stats, if client context is not passed down, the code path is as follows, only struct stats is attempted to fetch, otherwise NULL is directly returned.
https://github.com/duckdb/duckdb/blob/a81547b43c56c671a7ad1a636f83886713efdd8a/src/planner/filter/expression_filter.cpp#L232-L247

When client context is passed down, children function stats are fetched one by one.
https://github.com/duckdb/duckdb/blob/a81547b43c56c671a7ad1a636f83886713efdd8a/src/planner/filter/expression_filter.cpp#L249-L257

No new test is added, existing tests already prove the optimization.


test/optimizer/aggregate_precompute_function_filter.test
```diff
@@ -0,0 +1,25 @@
...
+
+restart
+
+statement ok
+ATTACH '{TEST_DIR}/func_filter_agg.duckdb' AS db (ROW_GROUP_SIZE 2048)
+
+query II
+EXPLAIN ANALYZE SELECT count(*) FROM db.t WHERE year(ts) = 2022
+----
+analyzed_plan	<REGEX>:.*"\+"\(2048, #0\).*Row Groups Scanned: 1 / 4.*
```
dentiny: Before this change, plan looks like
```sql
-- row group scanned is 2 out of 4
╭─ Summary ───────────╮
│ Total Time: 0.0007s │
│ Data Read: 262.1 kB │
╰─────────────────────╯
│ 1 row                  0µs │
╰──────────────┒┎────────────╯
╭─ Table Scan ─┚┖────────────╮
│ Table: db.main.t           │
│ Filters: "year"(ts) = 2022 │
│ Row Groups Scanned: 2 / 4  │
│ 2,190 rows             0µs │
```

PR #23590, opened 2026-07-03
Install bundled fmt headers

### LINKED ISSUE #23256: HEAD: duckdb/function.hpp includes non-installed dependency fmt/core.h
### What happens?

After installing DuckDB built from HEAD, a minimal C++ application that only includes duckdb.hpp fails to compile.

Compilation fails with:
```bash
/path/to/include/duckdb/function/function.hpp:19:10: error: 'fmt/core.h' file not found
#include "fmt/core.h"
         ^~~~~~~~~~~~
```

The error originates from the installed DuckDB header itself (`duckdb/function/function.hpp`), not from application code.

`duckdb/function/function.hpp` in HEAD contains:

```
#include "fmt/core.h"
```

However, `fmt/core.h` is not installed alongside the public headers.

### To Reproduce

I built DuckDB from the current HEAD following the official build instructions:

```bash
GEN=ninja make debug
```
https://duckdb.org/docs/current/dev/building/overview

The build documentation does not describe an installation step. To install the generated artifacts, I used the standard CMake install workflow shown below.

```bash
cd build/debug

cmake ../../ \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_INSTALL_PREFIX=/path/to/install

cmake --build . -j
cmake --install .
```

After installation, I attempted to build a minimal C++ application using the installed SDK:

```cpp
#include "duckdb.hpp" 

auto main() -> int { 
    return 0; 
}
```

Compilation fails with:
```
/path/to/include/duckdb/function/function.hpp:19:10: error: 'fmt/core.h' file not found
#include "fmt/core.h"
         ^~~~~~~~~~~~
```

Compile the source file using only the installed DuckDB SDK include and library directories.

Observe that compilation fails because `duckdb/function/function.hpp` includes `fmt/core.h`, which is not installed.

### OS:

MacOS 15.7.7（24G720）

### DuckDB Version:

HEAD

### DuckDB Client:

C++

### Hardware:

1.2 GHz Quad core Intel Core i7

### Full Name:

Kazuhiko TAMURA

### Affiliation:

out of work

- [x] Yes, I have

- [x] Yes, I have

zupengwang:
Fixes #23256.

DuckDB public headers include bundled fmt headers, but the CMake install rules only installed DuckDB's own public headers. This meant consumers using an installed DuckDB SDK could fail to compile with missing fmt headers.

This change installs the bundled fmt headers alongside the DuckDB headers and includes the fmt include directory in the build-tree CMake package config.

Validation:
- git diff --check
- Configured DuckDB with CMake/Ninja
- Compiled a minimal #include "duckdb.hpp" smoke test against a simulated installed include tree

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1026,7 +1026,7 @@ set(CMAKE_CONFIG_VERSION_TEMPLATE
 if(EXISTS ${CMAKE_CONFIG_TEMPLATE} AND EXISTS ${CMAKE_CONFIG_VERSION_TEMPLATE})
 
   # Configure cmake package config for the build tree
-  set(CONF_INCLUDE_DIRS "${PROJECT_SOURCE_DIR}/src/include")
+  set(CONF_INCLUDE_DIRS "${PROJECT_SOURCE_DIR}/src/include;${PROJECT_SOURCE_DIR}/third_party/fmt/include")
   configure_file(${CMAKE_CONFIG_TEMPLATE}
                  "${PROJECT_BINARY_DIR}/DuckDBConfig.cmake" @ONLY)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -141,6 +141,11 @@ install(
   FILES_MATCHING
   PATTERN "*.hpp"
   PATTERN "*.ipp")
+install(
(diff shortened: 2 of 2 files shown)


Tishj:
Thanks but I don't think this is the correct fix.
The policy is that anything in `third_party` shouldn't be included in public-facing headers, so those includes should be removed, rather than the build adjusted.

PR #23591, opened 2026-07-03
[Parquet][Variant] Don't produce VARIANT metadata with `sorted_strings` set

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9814

Not all VARIANT production logic produces sorted strings


Mytherin:
Thanks - should we push this into v1.5?

Mytherin:
Thanks!

PR #23599, opened 2026-07-03
Fix missing traversal for (correlated) subqueries in non-inner join conditions

kryonix:
Non-inner join planning could leave BoundSubqueryExpression nodes in ON conditions, which later failed during plan copying. Plan these subqueries through RecursiveDependentJoinPlanner using a narrow join-condition entry point after join condition extraction.

Classify subqueries by the side of the current join they reference, while treating outer-query-only correlations as not belonging to either current side. Keep pushed child filters covered, and reject subqueries that depend on both current join sides with a clearer error.

Add regressions for LEFT JOIN ON EXISTS/IN conditions, LHS/RHS/outer-only correlations, and a nested correlated comparison case that is now supported.

fix: https://github.com/duckdblabs/duckdb-internal/issues/9857

### DIFF
--- a/src/include/duckdb/planner/subquery/recursive_dependent_join_planner.hpp
+++ b/src/include/duckdb/planner/subquery/recursive_dependent_join_planner.hpp
@@ -14,6 +14,7 @@
 namespace duckdb {
 
 class Binder;
+class LogicalJoin;
 
 /*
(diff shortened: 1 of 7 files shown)


lnkuiper:
Looks good, thanks!

PR #23606, opened 2026-07-04
guard negative utf8proc_iterate return in string functions

jmestwa-coder:
utf8proc_iterate returns a negative error code on invalid or truncated utf8, but several callers add it straight to a byte index:
- rtrim/trim advance the copy `end` past the input, and lpad/rpad size the output from the same underflowed count
- strpos rewinds the read pointer before the buffer (asan: heap-buffer-overflow read of size 1 in utf8proc_iterate)
- the csv header trimmer walks column names with the same loop

Invalid utf8 reaches these through the appender / C API, which don't validate it. Added Utf8Proc::DecodeCharacter, which always advances at least one byte, and routed the callers through it.


jmestwa-coder:
any update?

PR #23611, opened 2026-07-04
Allow partial aggregate precompute with inexact row group stats

Damon07:
Similar to  what we did in `StatisticsPropagator::TryExecuteAggregates`, we can also specify the specific partitions with  inexact stats for scanning instead of bailing out the optimization entirely. This PR also tries to make the code more readable since the `StatisticsPropagator::TryExecuteAggregates` is very long and not easy to be maintained.


The-Alchemist:
can't wait to use this feature in https://github.com/The-Alchemist/duckdb-iceberg/pull/5 to speed up `MIN/MAX` queries against Iceberg tables!

Damon07:
not stale

PR #23613, opened 2026-07-05
Prefetch DuckDB header on attach to cut redundant reads

carlopi:
This reimplements #22359.

When attaching a DuckDB database, the file was opened once to detect its type (magic bytes) and then re-opened to read the header, issuing several separate reads. For remote files each read is a network request, so this adds avoidable round-trips.

While detecting the file type, read the whole header region in one go (MainHeader + both DatabaseHeaders, 3 x 4 KiB) and carry those bytes on AttachOptions. When the storage layer later opens the database, the initial header reads are served from the prefetched bytes instead of re-reading.

Only the bytes are kept, not the open handle: the actual database handle is opened later, so we never hold a handle to a file during type detection. 

```sql
 ATTACH 'http://127.0.0.1:8080/empty.db';
```
we (that is, duckdb `main` branch) currently performs:
```
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db       (start=0, len=12288)
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    8	4
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    0	4096
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    4096	4096
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    8192	4096
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db.wal
```

After this PR, 4 range requests are removed, to:
```
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db
GetRangeRequest 	https://blobs.duckdb.org/data/tpch-sf1.db	    (start=0, len= 12288)
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db
HeadRequest         https://blobs.duckdb.org/data/tpch-sf1.db.wal
```
There is a way to cut also 1st and 3rd request, so moving 8 -> 4 -> 2 requests.

### DIFF
--- a/src/common/file_buffer.cpp
+++ b/src/common/file_buffer.cpp
@@ -116,6 +116,11 @@ void FileBuffer::Read(QueryContext context, FileHandle &handle, uint64_t locatio
 	handle.Read(context, internal_buffer, internal_size, location);
 }
 
+void FileBuffer::ReadFromMemory(const_data_ptr_t source, uint64_t location) {
+	D_ASSERT(type != FileBufferType::TINY_BUFFER);
+	memcpy(internal_buffer, source + location, internal_size);
+}
+
 void FileBuffer::Read(QueryContext context, MemoryMappedFile &handle, uint64_t location) {
 	D_ASSERT(type != FileBufferType::TINY_BUFFER);
 	// Zero-copy: adopt a pointer into the mapping.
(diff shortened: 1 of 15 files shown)


pdet:
Looks fine to me, one minor nit, the comments look a bit excessive, can we keep them to one line, max 2 if truly necessary?

PR #21458, opened 2026-03-18
Buffer managed indexes, Pt. 1

artjomPlaunov:
(edit: rewriting to make the changes up to date after all reviews)

This is an intermediate step towards buffer managed indexes, this is one of the bigger refactors that was required; if all goes according to plan then the subsequent PRs will not touch as much core ART code at once. 

### Node Handle/ConstNodeHandles

NodeHandle and ConstNodeHandle are no longer templated classes. The concrete node type is now supplied on the getter. Also as a refresher, the idea with the handles is to keep the underlying buffer pinned RAII style when a handle is in scope. NodeHandle marks the underlying buffer as dirty, whereas ConstNodeHandle does not and enforces constness on its methods. 

```cpp
NodeHandle handle(art, node);
auto &node4 = handle.Get<Node4>();
```

Each handle contains a SegmentHandle and the node type; one thing that was lost pushing the template into the getter is some static type safety. To sort of make up for this I at least added some asserts to check that T::TYPE on the getter matches the stored run-time type of the handle. 

### Handle-scoped scanners

The original scanner stored references to slots within a node, assuming that node should always be pinned in memory.  This PR adds a scanner that keeps copied Node values (in the next PR, Node will be renamed to NodePtr to make it clear this is just a 64 bit pointer value inherited from IndexPointer), e.g. when we're talking about a "Node Value" it's just a 64 bit pointer. 

ARTScanChildren will pin a node using a node handle, and passes each child slot reference to a child_handler. This allows us to pass the reference to the pointer slot while keeping the parent pinned, so we can e.g. update the parent in place. 

-------------------------------
old PR description (will delete after finished writing new above): 
### NodeHandle/ConstNodehndles

Decided to push templating into the methods of the (Const)NodeHandles, instead of on the class. I actually don't think this is strictly necessary, but I ended up liking how the code looks/makes the NodeHandle slightly more generic except for when we do method invocations and need to specify the node type. 

### ARTScanPreOrder

Replace the ARTScanner with buffer managed scanners. Previously, we are assumed that ART nodes always remain pinned in memory so we were able to do in place updates with any references we obtained into the tree. For any stack based traversals, I don't think that we want to push pinned nodes onto the stack. Since we got rid of the max key length restriction on the ART, we could hit a degenerate case with a really long key that creates a stack of pinned nodes, and again, exceed memory without being able to evict anything. 

[...]

PR #23619, opened 2026-07-05
Buffer managed indexes 3 - NodePtrHandle + owning PrefixHandle + InsertIntoNode and Prefix::New rewrite

artjomPlaunov:
https://github.com/duckdblabs/duckdb-internal/issues/10403

Introduce a "SlotHandle" abstraction which is the abstraction for providing a reference to a NodePtr reference or "slot" that can be updated in place. This can be with or without a pinned node encapsulating, as demonstrated by thw two constructors:

```cpp
	//! Create a slot handle for memory that is valid without this object owning a pin.
	explicit SlotHandle(NodePtr &slot_p) : slot(slot_p) {
	}

	//! Create a slot handle that owns the pin keeping slot_p valid.
	explicit SlotHandle(NodePtr &slot_p, NodeHandle &&pin_p) : slot(slot_p), pin(std::move(pin_p)) {
	}
```

For example, in a stack local context, or for storing the root pointer of the ART, we have a guarantee those references are valid where they are being used. In case the slot is within a node, the SlotHandle stored a pinned NodeHandle for the NodePtr slot. 

To test this out, we rewrite InsertIntoNode and add PrefixHandle::New (a re-write of Prefix::New). In particular for PrefixHandle::New, we use the SlotHandle as a cursor in constructing the new prefix chain, i.e. the flow is: pin the pointer we want to update -> allocate a new prefix node and have the slot point to the new allocation -> traverse to the new prefix node and rebind the slot cursor to be referencing the child pointer in the new node -> repeat. 

Note that we also have PrefixHandle be a wrapper over a NodeHandle; previously we had PrefixHandle as a collection of static helpers on NodeHandles that were meant to be interpreted as prefix nodes, now we make this more explicit with PrefixHandle just wrapping the NodeHandle (also allows us to provide some D_ASSERT safety checks).


baolinhuang:
This is highly valuable work; the inability to unpin from memory severely impacts the use of ART.

artjomPlaunov:
@baolinhuang great to hear! :-)

artjomPlaunov:
Copying this discussion point from https://github.com/duckdb/duckdb/pull/21458, as I think this PR will be the right place to make sure we have this distinction: 

I am wondering about whether it makes more sense to strictly copy out (and operate on copies of) Node, or to treat them consistently as references (I saw that change in a few places)? Although maybe that does not matter as much, as eventually they will (almost all) be handles, as we cannot copy out/etc. anymore without accidentally evicting the memory? And we'll reason about the places where we copy out by stating why we do not need to pin for these?

Say we look at the below pattern: We should have a wrapper object or so, which always keeps the previous handle alive? So that we can acquire the handle on the child in the next iteration with the parent still alive. But I guess that's future PRs.

Hmmm, or maybe we don't even need to keep the parent pinned, as we only need the information in that Node (that we copied out). Sorry, just rubber ducking a bit I guess. But in that case I would like to make that more explicit maybe? Can we alias the Node or so? Or maybe it'll become more obvious when we rename it to NodePtr 😆

Node current = node;
while ...
   NodeHandle handle(art, my_node);  
   current = child;
Maybe we can defer this discussion to PR 3. I think there will be two distinct semantics: either we only need a copy of the Node (NodePtr), or we need a reference to the actual NodePtr slot, in which case we must keep the containing parent pinned. PR 3 will introduce something like a SlotHandle abstraction for the latter case. PR 2 will first rename Node to NodePtr, which should make this distinction clearer. We can then revisit during PR 3 how best to express and enforce these two cases.

taniabogatsch:
> Although maybe that does not matter as much, as eventually they will (almost all) be handles, as we cannot copy out/etc. anymore without accidentally evicting the memory? And we'll reason about the places where we copy out by stating why we do not need to pin for these?

Indeed, I think copying out is an unsafe way of keeping a snapshot to memory that can change/get evicted. So, in my "vision", that does not happen anymore, and anything we hold is always guaranteed to be pinned.

taniabogatsch:
> Or maybe it'll become more obvious when we rename it to NodePtr

I thought that was what that whole renaming idea was about :D

src/include/duckdb/execution/index/art/node_handle.hpp
```diff
@@ -50,6 +53,48 @@ class NodeHandle {
...
 
+//! SlotHandle wraps a mutable reference to a NodePtr slot.
+//! Ref() is valid for the lifetime of this object.
+//! If constructed with a pin, SlotHandle owns the NodeHandle keeping the slot valid.
+//! If constructed without a pin, the caller must guarantee the slot remains valid
+class SlotHandle {
+public:
+	//! Create a slot handle for memory that is valid without this object owning a pin.
+	explicit SlotHandle(NodePtr &slot_p) : slot(slot_p) {
+	}
```
taniabogatsch: I feel like this is a bit duplicated and just adds another layer to an unsafe abstraction?

src/execution/index/art/prefix.cpp
taniabogatsch, line 64: Diving more into this PR, I find the name "slot" a bit counter intuitive, as it's not really something that we've been using in our code base. How about `*prefix.child_ptr`? Maybe even `*prefix.child_node_ptr`? 

Similarly, I am not fully on board with the usage of `SlotHandle`. Now that we've renamed `Node` to `NodePtr`, couldn't we just go with `NodePtrHandle`? Since that's effectively what it is? I.e., we no longer pass around `Node &node`, but now we pass `NodePtrHandle &handle`?

test/sql/index/art/insert_update_delete/test_art_slot_handle_prefix_new.test
taniabogatsch, line 1: Our existing tests aren't already covering this code path?

artjomPlaunov:
> > Although maybe that does not matter as much, as eventually they will (almost all) be handles, as we cannot copy out/etc. anymore without accidentally evicting the memory? And we'll reason about the places where we copy out by stating why we do not need to pin for these?
> 
> Indeed, I think copying out is an unsafe way of keeping a snapshot to memory that can change/get evicted. So, in my "vision", that does not happen anymore, and anything we hold is always guaranteed to be pinned.

Just to confirm and clarify the distinction, however, I want to confirm that is is ok to copy out a NodePtr "value" in a lot of scenarios (for example lookup, stack based traversal) whereas what was unsafe before was copying out the _references_ to NodePtrs, not the NodePtrs themselves. 

E.g. a reference to a NodePtr presupposes some stable memory for where the NodePtr is resident, which is the unsafe part. Whereas the NodePtr is just 64 bits referring to a stable node entity in the ART that we can always re-pin if we need to get the node into memory.

src/include/duckdb/execution/index/art/node_handle.hpp
artjomPlaunov: In some scenarios, more specifically 1) the ART root NodePtr and 2) local function stack NodePtr variables, the caller can guarantee the storage for NodePtr is there, which is different from assuming it is within a pinned node in memory. 

I can see if we can change this, but, it makes the code a lot easier to be able to have e.g. cursors where we initially start with a stack variable binding, and traverse down a chain of nodes.

artjomPlaunov: I guess this was more of an ease of coding thing to be able to handle common scenarios where we initiate a cursor using a stack local variable, and descend into a chain of nodes. I will play around with this, maybe it isn't so bad to just pass in the stack local or ART root NodePtr& explicitly in these cases.

artjomPlaunov: Although I'm wondering if really this is more of a cursor abstraction I introduced, rather than a NodePtrHandle (esp with the rebind functions).

taniabogatsch: Yea, the usage makes sense to me. I am just a bit worried that it blurs the lines between the now safe `NodePtrHandle`, and then suddenly that handle can also be unsafe. But that being said, if you've played around with it and it makes the stack code specifically more readable/generic, then let's use it like this!

artjomPlaunov:
@taniabogatsch I tried out a redesign to make the NodePtrHandle safer and avoid allowing an arbitrary reference as part of the constructor. I also made the constructor private here and a method on PrefixHandle (friend class) is responsible for creating the NodePtrHandle, e.g. we can only create it from an active pin.

taniabogatsch (approved):

Mytherin:
Thanks!

PR #23630, opened 2026-07-06
Storage: validate dictionary string indexes

wufengwind:
Dictionary-compressed string scans read dictionary offsets and selection indexes from the storage block.

Those values were only constrained by writer-side invariants. In release builds, a checksum-valid corrupted dictionary-compressed string segment could use an out-of-range dictionary offset or dictionary index and decode incorrect strings instead of reporting storage corruption.

This validates dictionary metadata during scan initialization and checks dictionary indexes and offsets before using them. Invalid metadata is reported as an IOException.

A sqllogictest covers checksum-valid corrupted dictionary-compressed database files with an invalid dictionary offset and an invalid dictionary index.


src/storage/compression/dictionary/decompression.cpp
```diff
@@ -43,6 +77,10 @@ void CompressedStringScanState::Initialize(ColumnSegment &segment, bool initiali
 	block_size = segment.GetBlockSize();
 
 	dict = DictionaryCompression::GetDictionary(segment, *handle);
+	if (dict.size > block_size || dict.end > block_size || dict.size > dict.end || index_buffer_count == 0) {
```
lnkuiper: This still leaves one checksum-valid corruption path unchecked: dict.size <= dict.end <= block_size does not guarantee that the dictionary region is actually after the packed selection data and index buffer. If dict.end - dict.size overlaps the index buffer, FetchStringFromDict will read index words as string
bytes and silently decode garbage instead of throwing corruption. I reproduced that by patching only dict_end in a valid dictionary-compressed block; the scan returned "\4\0\0\0", "\b\0\0\0", etc. instead of the original strings. I think this init check also needs to validate dict.end - dict.size >=
index_buffer_offset + sizeof(uint32_t) * index_buffer_count (or equivalent layout invariants for the full segment).

lnkuiper (changes requested):
Thanks for the PR! I think there's still an issue:
```sql
statement ok
ATTACH '/tmp/review23630_bad_dict_end.db' AS repro_db
statement ok
USE repro_db
query III
SELECT
    count(*) FILTER (WHERE b = 'aaaa') AS aaaa_count,
    count(*) FILTER (WHERE b IS NULL) AS null_count,
    sum(length(b)) AS total_len
FROM test
----
0    2000    32000

```

[review23630_bad_dict_end.zip](https://github.com/user-attachments/files/29797844/review23630_bad_dict_end.zip)

src/storage/compression/dictionary/decompression.cpp
wufengwind: Thanks, good catch. I added the dictionary layout invariant so the dictionary region must start after the packed selection data and index buffer. I also added a checksum-valid corrupted fixture that patches only dict_end to make the dictionary overlap the index buffer, and the scan now reports corruption.

lnkuiper:
Thanks for the changes! Looks good to me now, just need to get CI green. The failure is not related to your changes, seems like some extension broke. We'll look into it.

lnkuiper (approved):

PR #23631, opened 2026-07-06
WAL group commit

ywelsch:
DuckDB's commit path is currently "one query at a time", which includes making the changes persistent. Concurrent committers queue behind each other's fsyncs and throughput is pinned near `1 / fsync_latency` regardless of how many connections are committing. That is most painful on high-latency durable storage (e.g. networked disks), where a single fsync can take milliseconds to potentially hundreds of milliseconds.

This PR introduces group commits, decoupling the fsync on the commit path and sharing it amongst concurrent committers. The visibility and durability guarantees stay the same.

## What this builds on, and how it works

The PR here builds on #25187 and #25389 in order to keep the diff small. It caps the visibility bound based on durability, which is now tracked separately. While commits are pending durability, `StartTransaction` caps a new snapshot's visibility bound below the first commit that is not yet durable, so no transaction can see a commit a crash could still lose. Acknowledging a commit advances the bound, which means read-your-writes still holds: once `COMMIT` returns, that commit is durable and every subsequently started transaction sees it.

## Code changes

- The WAL sync splits in two: `FlushMarker` writes the `WAL_FLUSH` marker and pushes the buffer to the OS, still under the WAL lock; `SyncUpTo(offset)` runs after the locks are released and blocks until the WAL is durable to that offset, with one caller syncing on behalf of every offset requested so far.
- A committed transaction remains in `active_transactions` until its WAL sync completes. The durable bound advances by whichever thread's sync covered a commit, and `WaitForDurability` waits until no active transaction still awaits its sync; checkpoints use it before swapping or deleting the WAL.
- A commit with no other transaction syncs under the transaction lock, as on main. It has nothing to share the fsync with, and since no snapshot can start below it, the `deleted_rows_in_use` fast path for index deletes stays available. Only commits that overlap other transactions defer their sync.
- A pending commit holds the lowest visibility bound like any active transaction, so nothing a bounded snapshot might still read is vacuumed, and `HasOtherTransactions` counts it like any other transaction.

## Benchmark

`benchmark/micro/concurrent_commits.cpp` (`[wal]`): concurrent single-row-commit throughput on a persistent database with no fsync, and with 1 ms and 10 ms fsync latency injected through a `LocalFileSystem` subclass on the DB config. The no-fsync column skips the sync altogether and measures the commit path itself. Fixed total work per column (16000, 800 and 160 commits), median of 5 runs, reported in commits per second; this PR, with `main` in parentheses; higher is better. Measured on an Apple M4 Pro (14 cores: 10 performance, 4 efficiency), 48 GB, local SSD, macOS 15.7.

| threads         | no fsync      | 1 ms        | 10 ms      |
| --------------- | ------------: | ----------: | ---------: |
| 1 (no batching) | 7383 (7476)   | 480 (456)   | 68 (71)    |
| 4               | 19969 (19948) | 1278 (584)  | 148 (74)   |
| 8               | 24370 (24899) | 2334 (588)  | 287 (74)   |
| 16              | 23775 (24342) | 4717 (598)  | 544 (75)   |
| 32              | 24465 (23828) | 9920 (579)  | 1195 (74)  |

`main` is invariant to concurrency: it sits at `1 / fsync_latency` whether 1 or 32 threads are committing, about 580 commits per second at 1 ms and 74 at 10 ms. This PR scales with committer count: at 1 ms, 4 threads commit 2.2x faster than on `main`, 8 threads 4.0x, 16 threads 7.9x and 32 threads 17x, and similarly at 10 ms. One sync runs at a time, so a commit arriving during a sync waits for it and joins the next; per-commit latency stays between one and two fsyncs, and the throughput comes from every sync covering all commits flushed since the previous one.

### DIFF
--- a/src/transaction/duck_transaction_manager.cpp
+++ b/src/transaction/duck_transaction_manager.cpp
@@ -1,4 +1,6 @@
 #include "duckdb/transaction/duck_transaction_manager.hpp"
+
+#include <chrono>
 #include "duckdb/logging/log_manager.hpp"
 
 #include "duckdb/main/client_data.hpp"
@@ -12,6 +14,7 @@
 #include "duckdb/catalog/dependency_manager.hpp"
 #include "duckdb/storage/storage_manager.hpp"
 #include "duckdb/transaction/duck_transaction.hpp"
+#include "duckdb/transaction/transaction_data.hpp"
 #include "duckdb/main/client_context.hpp"
 #include "duckdb/main/connection_manager.hpp"
 #include "duckdb/main/attached_database.hpp"
@@ -19,6 +22,7 @@
 #include "duckdb/transaction/meta_transaction.hpp"
 #include "duckdb/main/settings.hpp"
 #include "duckdb/storage/checkpoint/checkpoint_options.hpp"
+#include "duckdb/storage/write_ahead_log.hpp"
 #include "duckdb/common/string_util.hpp"
 
 namespace duckdb {
@@ -45,8 +49,8 @@ bool DuckCleanupInfo::ScheduleCleanup() noexcept {
 }
 
 DuckTransactionManager::DuckTransactionManager(AttachedDatabase &db) : TransactionManager(db) {
-	// start timestamp starts at two
-	current_start_timestamp = 2;
+	// real transactions start above the system transaction's timestamp
+	current_start_timestamp = SYSTEM_TRANSACTION_TIMESTAMP + 1;
 	// transaction ID starts very high:
 	// it should be much higher than the current start timestamp
 	// if transaction_id < start_timestamp for any set of active transactions
@@ -86,15 +90,20 @@ Transaction &DuckTransactionManager::StartTransaction(ClientContext &context) {
 	} // LCOV_EXCL_STOP
 
 	// obtain the start time and transaction ID of this transaction
-	transaction_t start_time = current_start_timestamp++;
+	transaction_t unique_start_time = current_start_timestamp++;
 	transaction_t transaction_id = current_transaction_id++;
+	// snapshots must not observe commits that are not yet durable, nor a newer catalog version
+	auto snapshot_bound = GetSnapshotBound();
+	auto start_time = MinValue<transaction_t>(unique_start_time, snapshot_bound.start_time);
+	auto catalog_version = MinValue<idx_t>(last_committed_version, snapshot_bound.catalog_version);
 	if (active_transactions.empty()) {
 		lowest_active_start = start_time;
 		lowest_active_id = transaction_id;
 	}
 
 	// create the actual transaction
-	auto transaction = make_uniq<DuckTransaction>(*this, context, start_time, transaction_id, last_committed_version);
+	auto transaction =
+	    make_uniq<DuckTransaction>(*this, context, start_time, unique_start_time, transaction_id, catalog_version);
 	auto &transaction_ref = *transaction;
 
 	// store it in the set of active transactions
@@ -121,6 +130,10 @@ DuckTransactionManager::CheckpointDecision::~CheckpointDecision() {
 }
 
 bool DuckTransactionManager::HasOtherTransactions(DuckTransaction &transaction) {
+	if (HasUnsyncedCommits()) {
+		// commits pending durability behave like active transactions (see UnsyncedCommit)
+		return true;
+	}
 	for (auto &active_transaction : active_transactions) {
 		if (!RefersToSameObject(*active_transaction, transaction)) {
 			return true;
@@ -164,8 +177,7 @@ DuckTransactionManager::GetCheckpointType(DuckTransaction &transaction, const Un
 	bool has_other_transactions = HasOtherTransactions(transaction);
 	if (has_other_transactions) {
 		if (undo_properties.has_updates || undo_properties.has_dropped_entries) {
-			// if we have made updates/catalog changes in this transaction we cannot checkpoint
-			// in the presence of other transactions
+			// other transactions - or snapshots bounded below pending commits - may need older data
 			string other_transactions;
 			for (auto &active_transaction : active_transactions) {
 				if (!RefersToSameObject(*active_transaction, transaction)) {
@@ -175,22 +187,17 @@ DuckTransactionManager::GetCheckpointType(DuckTransaction &transaction, const Un
 					other_transactions += "[" + to_string(active_transaction->transaction_id) + "]";
 				}
 			}
-			if (!other_transactions.empty()) {
-				// there are other transactions!
-				// these active transactions might need data from BEFORE this transaction
-				// we might need to change our strategy here based on what changes THIS transaction has made
-				if (undo_properties.has_dropped_entries) {
-					// this transaction has changed the catalog - we cannot checkpoint
-					return CheckpointDecision(
-					    "Transaction has dropped catalog entries and there are other transactions "
-					    "active\nActive transactions: " +
-					    other_transactions);
-				}
-				// this transaction has performed updates - we cannot checkpoint
-				return CheckpointDecision(
-				    "Transaction has performed updates and there are other transactions active\nActive transactions: " +
-				    other_transactions);
+			if (other_transactions.empty()) {
+				other_transactions = "[commits pending durability]";
 			}
+			if (undo_properties.has_dropped_entries) {
+				return CheckpointDecision("Transaction has dropped catalog entries and there are other transactions "
+				                          "active\nActive transactions: " +
+				                          other_transactions);
+			}
+			return CheckpointDecision(
+			    "Transaction has performed updates and there are other transactions active\nActive transactions: " +
+			    other_transactions);
 		}
 		// otherwise - we need to do a concurrent checkpoint
 		checkpoint_type = CheckpointType::CONCURRENT_CHECKPOINT;
@@ -207,6 +214,11 @@ DuckTransactionManager::GetCheckpointType(DuckTransaction &transaction, const Un
 void DuckTransactionManager::Checkpoint(ClientContext &context, bool force) {
 	if (ValidChecker::IsInvalidated(db)) {
 		throw IOException("%s", ValidChecker::InvalidatedMessage(db));
+	}
+	// drain pending commits here, where the query is still cancellable: inside the checkpoint an
+	// exception would invalidate the database. The authoritative drain under the WAL lock follows
+	if (!db.IsSystem() && db.HasStorageManager() && !db.GetStorageManager().InMemory()) {
+		WaitForDurability(context);
 	}
 	auto &storage_manager = db.GetStorageManager();
 	auto current = Transaction::TryGet(context, db);
@@ -244,7 +256,7 @@ void DuckTransactionManager::Checkpoint(ClientContext &context, bool force) {
 		}
 	}
 	CheckpointOptions options;
-	if (GetLastCommit() > LowestActiveStart()) {
+	if (!VisibleToSnapshot(GetLastCommit(), LowestActiveStart())) {
 		// we cannot do a full checkpoint if any transaction needs to read old data
 		options.type = CheckpointType::CONCURRENT_CHECKPOINT;
 	}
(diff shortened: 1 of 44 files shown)


MBkkt:
@ywelsch we had similar patch, it gives similar performance numbers but was quite simpler.
It will be nice if you can check it: https://github.com/serenedb/duckdb/commit/24fefe85405f2f8882effe399388519aab39098d

It also time independent, so it doesn't need configure option and it doesn't slowdown single thread

ywelsch:
> @ywelsch we had similar patch, it gives similar performance numbers but was quite simpler. It will be nice if you can check it: [serenedb@24fefe8](https://github.com/serenedb/duckdb/commit/24fefe85405f2f8882effe399388519aab39098d)
> 
> It also time independent, so it doesn't need configure option and it doesn't slowdown single thread

Thanks for the link. I had a quick look - it seems unsafe as it fsyncs only after making the changes visible to other transactions (i.e. visible-before-durable). Also, only the fsync coalesces, no shared-append concurrency. So in its current form it is unsound and not as powerful.

MBkkt:
You were right about the visible-before-durable part — a transaction starting while the fsync was still in flight could observe a commit that a crash would lose. I reworked that, and since the design ended up quite different from this PR, opened it separately rather than dumping it into this thread: #23655.

The gist: commit timestamps stay at commit time, but new snapshots are bounded at the durable horizon — a transaction that starts while a commit's fsync is pending simply doesn't see that commit yet. Same guarantee as assigning timestamps at publish time, but nothing blocks and publish order never changes, which is what keeps the patch small.

On "not as powerful": I ran your group_commit benchmark (extended with 2/32 threads and a 100ms latency tier) against both branches on a 32-core machine, numbers are in the PR. The main structural difference is that committers issue overlapping fsyncs instead of parking behind a single sync leader, which hides the fsync round trip on networked storage without needing a batching delay — whether fsyncs overlap or serialize is taken from the file system type rather than measured at runtime, and on local disks a single late-snapping fsync stream batches as well as a delay does. Appends do still serialize under the WAL lock, deliberately, since a page-cache append is orders of magnitude cheaper than the fsync it precedes.

One thing taken from this PR while reworking: the fsync-failure poisoning. The retry in the commit you looked at was indeed unsafe after a failed fsync, your handling is the correct one.

Mytherin (commented):
Thanks for the PR!

Very cool feature!

At a high level I'm wondering if we can split this up into separate PRs. This feature is rather complex and there's a lot of changes that are introduced that seem to be non-critical to the feature itself, e.g. I'm a bit worried about all of these locking changes - and they seem non-critical to getting the core feature to work. These seem like orthogonal optimizations and not related to the core feature.

I would prefer to split this up into individual, easy to review PRs - perhaps either starting with a restructuring of the WAL / undo buffer code to allow for group commits to be implemented on top of, or starting with the core group commit code (delaying fsync) but without all of the changes to the locking.

Maybe I'm missing something and they are all necessary for this change, though, happy to hear your perspective.

src/include/duckdb/storage/table/data_table_info.hpp
```diff
@@ -72,6 +78,10 @@ struct DataTableInfo {
 	vector<IndexStorageInfo> index_storage_infos;
 	//! Lock held while checkpointing
 	StorageLock checkpoint_lock;
+	//! Gate excluding DDL on this table from concurrent group-commit publishes.
+	//! Group commits touching this table hold it SHARED through publish; DDL on this table takes it
+	//! EXCLUSIVE. Per-table, so DDL on one table doesn't block others; separate from checkpoint_lock.
+	StorageLock publish_gate;
```
Mytherin: I don't really follow why this is necessary - if all we're doing is delaying the `fsync` on the WAL the commit has already happened while we are waiting for (concurrent) commits to come in, why do we need extra locking on tables here?

ywelsch: This PR defers more than the fsync, it defers the **publish**. A commit's changes only become visible to other transactions after its flush marker is durable, otherwise a transaction could observe (and act on) a commit that a crash would lose. So while we wait for the fsync, the commit has deliberately *not* happened yet from the rest of the system's perspective.

The publish_gate then follows from what `transaction_lock` currently provides on main. There, a commit writes its WAL entries first (unvalidated, with the transaction lock released), and then, under `transaction_lock`, runs the conflict check (the `HasParent()` check in `CommitState::CommitEntry`: "has another transaction altered a table I modified?") and applies the consequences (version flips, index removal, becoming visible), entry by entry. A failed check reverts everything: the exclusively-held WAL is truncated and the applied entries are rolled back. DDL never takes `transaction_lock`; it doesn't need to, because a commit remains fully abortable until it completes. From DDL's perspective a commit is effectively a point event: `AlterEntry`'s new table version lands either early enough that a check catches it (clean abort) or after the commit is fully visible.

Deferring publish breaks the abortability that made this work: once a commit's flush marker is durable in a WAL that other commits keep appending to, it can neither be truncated nor rolled back (without bigger surgery and storage-related changes, e.g. a compensating abort record), so replay will apply it. The check must therefore move *before* the flush marker (out of the `transaction_lock` section, since the WAL write happens with `t_lock` released), while the consequences run *after* the fsync (back under `t_lock`, at publish). `t_lock` can't span the gap, as that would stall transaction start and all other commits for the duration of an fsync, which is the thing the feature exists to avoid. So there is now a window in which a commit has passed its checks, can no longer abort, and hasn't acted. An `ALTER TABLE t` landing inside it means the per-entry check in `CommitEntry` fires *past the point of no return*: the only remaining option is a fatal error, for what should be an ordinary transaction conflict (and I believe that failing the ALTER TABLE is not a viable option, DDL should not be sniped by DML).

The check also can't simply be skipped for deferred commits: the ALTER's rebuild and a `CREATE INDEX`'s attach would then proceed against a table state that excludes the pending commit. Concretely, a committed `UPDATE` on a retyped column gets lost (the rewrite bakes in the old value and the update lands on the discarded column), and a freshly attached index never receives the removals a publishing `DELETE` applies to every *attached* index, leaving a permanently untracked stale entry. The conflict is real; it just has to be resolved while aborting is still possible.

That's all the publish gate does: it pins the table state the validation saw until publish consumes it. It is the per-table replacement for the `transaction_lock` check/act piece. It can be seen as a **commit-scoped schema-stability lock**: commits hold it shared (they never block each other, so batching is unaffected; statements and reads never touch it), DDL on that table takes it exclusive for the duration of the catalog change. In the steady state it's uncontended; the only pairing that ever waits is a DDL vs. an in-flight commit on the same table, for at most about one fsync. It is the same pairing that on main resolves by aborting one side.

(Longer term I think the exclusive section can shrink to nearly nothing with a delta-catch-up ALTER — rebuild against a snapshot, apply the changes that committed meanwhile, swap under the gate for a final millisecond delta — which would also fix a family of existing rebuild races on main, e.g. rows committed between `CREATE INDEX`'s build scan and its attach are missing from the index today. That's a separate project though.)

**I think this is one of the key pieces of this work to discuss. If you agree with the direction, I can factor out this piece here as a first step. If you disagree, happy to discuss alternatives.**

I've went through the following options:

- **Compensating WAL abort records** (make post-marker abort possible): a WAL-format and replay change of its own magnitude, the abort record needs its own fsync before the failure can be reported, and reverting a commit a concurrent ALTER already rebuilt against reintroduces the structural races.
- **Fail or delay the DDL instead**: Implementing it requires atomically testing for in-flight commits and blocking new validations, which *is* the gate; waiting then preferred over failing (a busy table always has a commit in flight, and a fail-fast ALTER would hence starve).
- **Coarser locks**: holding `transaction_lock` across the fsync stalls transaction start and all commits; a global gate or having DDL take the WAL lock stalls *every* table's commits for the duration of an ALTER's table rewrite; reusing `append_lock` (a plain mutex) serializes same-table commits across the fsync, and reusing the per-table checkpoint lock creates a genuine deadlock cycle with the WAL lock ordering.
- **Delta-catch-up ALTER** (online-DDL style: rebuild against a snapshot, apply changes that committed meanwhile, swap under the gate for a final millisecond delta): the one genuine successor — it shrinks the exclusive publish gate section to nearly nothing *and* fixes a family of existing rebuild races on main (e.g. rows committed between `CREATE INDEX`'s build scan and its attach are missing from the index today). It's a separate, larger project.

So the publish gate (schema stability lock) felt like the best option.

Mytherin: Could you expand a bit on deferring the publish? What's the benefit of deferring the publish versus only deferring the fsync? This seems to introduce a lot of complexity and I don't quite see the benefits.

ywelsch: If we publish before fsyncing, it means that a transaction (e.g. a read-only transaction) might already see the changes before they are durable.

Trying to address this would mean trying to keep new snapshots / transactions capped below the durable horizon (which is the path that https://github.com/duckdb/duckdb/pull/23655 went down, while working through test failures I assume), which means revisiting the transaction versioning system. I had briefly tried looking into this, but from what I remember, it came with its own set of challenges, while being a more foundational change to the system than the publish gate (which is just a table lock), interacting in subtle ways with clean-up, catalog visibility, checkpointing and vacuum.

I can try break this apart further (and look into it again more deeply) if you want.

MBkkt: @Mytherin Hi, after some iterations (I initially wrote buggy version of groupped fsync after commit lock, then tried multiple iterations with it, then tried this PR approach and second branch of @ywelsch), in the end I choose to defer fsync, instead of defer publish, I think defer fsync is also quite complicated, but with defer-publush approach main issue that I faced, our custom indexes can fails to apply (index delete can throw), but with this PR approach it's not possible to make apply non-fatal fail.

Here is my patches:
https://github.com/serenedb/duckdb/commit/1226356a0d279b1dad00a99d5b80b303391cfc6e (first, after multiple iterations)
https://github.com/serenedb/duckdb/commit/204666268bd39bcf579e026ac6ccb3340b938d84 (second, fix for first)

This seems to works for us correctly and scales with threads.
I'm not insist, I even closed mine PR, but in case of someone interested in alternative

In general I don't think I like either of approaches, both feels meh, although some way to make fsync groupped is really required. I think at some point I tried to put groupped fsync between something like prepare commit and actualy commit, and it looked clean, but the issue that it's kind of impossible to do this correctly with current optimistic concurrency control (at least I didn't found solution).

MBkkt: Maybe interesting point that is actually one thing is really easy, if we can accept that non-yet-durable changes can be read (breaks current guarantees), in theory it can be setting. In such case we can just put groupped fsync outside of commit lock very easy. From practice side I don't think it's bad (postgres/other OLTP systems commonly doesn't fsync with transaction commit), but I really dislike that it became worse than current main especially when it feels like some solution exists (probably complicated).

Mytherin: I agree that it would be better if data that is not durable could not be read by other transactions - but I think the locking approach seems to introduce a lot of complexity.

Given the fact that we're MVCC anyway, perhaps an easier path is to have new transactions start at timestamps *prior* to any "committed-but-not-published" commits, e.g.:


```
COMMIT T1;     -- commits ts "2", not yet published
START T2;      -- starts at ts "1" (avoids reading committed but unpublished data)
PUBLISH T1;    -- now publish ts "2"
START T3;      -- starts at ts "2" since this transaction has been published
```

I feel that with this path we can avoid making any locking changes - all we need to do is start new transactions so that they can only see snapshots that have been both committed **and** published.

MBkkt: @Mytherin this exactly the shape that I tried to make. I think the result is working, without locking change (you can check commits above, few lines in them are unrelated, they needed for our eventual consistent index, sorry), but it's still quite complicated, I didn't find easy way of doing this, most complicated part is checkpoint.

Mytherin: Ah yeah, I think that design makes the most sense / seems the easiest from where I'm sitting, but I haven't dove deep into implementing this myself. I fully understand that some changes need to be made to accommodate this but I don't quite follow what's extremely complicated here - checkpoints already need to deal with transactions that are active and need access to read older data (which is now done through `LowestActiveStart`).

ywelsch: I've worked through the details of the "bounded snapshots" alternative here: https://github.com/duckdb/duckdb/commit/60dd969e5a2dc1b8f2e59e3e2840da4f00bef97c

If it directionally looks good to you, I can update the PR here accordingly.

Key points:
- it tracks the *pending* "written-to-the-wal" but not yet fsynced commits
- visibility control takes these into account (i.e. bounds snapshotted views	below that)
- adapts the various mechanisms (checkpoints, gc, catalog versions, ...)
- A key difference to main is that a bounded snapshot can now start AT a commit id, while on main start times and commit ids were distinct draws from one shared counter. This requires adapting some of the visibility/conflict detection predicates, which I've started refactor into methods so that these (now adapted) bound checks are not spread throughout the codebase.

More detail in the commit message above.

Mytherin: That looks great. I think a big question here is how to add thorough testing for this as it's obviously quite complex. Maybe we could add some debug settings that e.g. add artificial delays or artificial failures in places to allow for more thorough testing?

ywelsch: ok, I will bring that back here to this PR and ping you when it's ready.

> Maybe we could add some debug settings that e.g. add artificial delays or artificial failures in places to allow for more thorough testing?

Definitely!

ywelsch: @Mytherin this is ready for a new look (PR description updated too). Please also feel free to directly push code changes to the PR branch here if you find that useful.

ywelsch:
@Mytherin could you give this another look? Thank you :)

src/transaction/duck_transaction_manager.cpp
```diff
@@ -439,8 +495,54 @@ ErrorData DuckTransactionManager::CommitTransaction(ClientContext &context, Tran
...
+		t_lock.unlock();
+		if (notify_others) {
+			durability_cv.notify_all();
+		}
+	}
+
 	CleanupTransactions();
 
+	if (checkpoint_decision.can_checkpoint && (undo_properties.has_updates || undo_properties.has_dropped_entries) &&
+	    GetLastCommit() >= LowestVisibilityBound()) {
```
Mytherin, line 537: `GetLastCommit` is currently also moved by read-only transactions which aren't blocked in this path (as the transaction lock is released here) - I think this might introduce a race in the current state of the code. I think this is separately fixed by https://github.com/duckdb/duckdb/pull/24764 which makes `last_commit` only move when a write transaction commits - so I think this is fine given that PR goes in first.

src/transaction/duck_transaction_manager.cpp
```diff
@@ -427,8 +480,11 @@ ErrorData DuckTransactionManager::CommitTransaction(ClientContext &context, Tran
 	bool store_transaction = undo_properties.has_updates || undo_properties.has_index_deletes ||
 	                         undo_properties.has_catalog_changes || error.HasError();
 
-	// Remove the transaction from the list of active transactions and gather cleanup information.
-	QueueCleanup(RemoveTransaction(transaction, store_transaction, CreateCleanupInfo()));
+	if (!commit_wal) {
+		// Remove the transaction from the list of active transactions and gather cleanup information.
+		// A commit that needs a WAL sync stays active until the sync below has completed.
+		QueueCleanup(RemoveTransaction(transaction, store_transaction, CreateCleanupInfo()));
```
Mytherin, line 486: Given we're now no longer immediately queueing for cleanup in some cases this might leave `recently_committed_transactions` out-of-commit-order which is assumed elsewhere (`SweepCommittedTransactions`) to be in this order.

I wonder if we could e.g. always enqueue and then pause the sweeping, or make `QueueCleanup` no longer insert blindly at the end / change the data structure used for `recently_committed_transactions` to keep the transactions in order?

Mytherin (commented):
Thanks for the changes! Looks much better - some comments

src/transaction/duck_transaction_manager.cpp
ywelsch: Good catch. This was potentially delaying cleanups. I've pushed [2523a27](https://github.com/duckdb/duckdb/pull/23631/commits/2523a27377efdb8f63bbae55f7ac35ec28471f9e) to make sure recently_committed_transactions keeps them in order. I chose not to change the data structure as I'm expecting this list/set to be usually quite small, and keeping it a vector also keeps the code introduced in https://github.com/ywelsch/duckdb/commit/aa4ce8ee406ec1953e08cb7ddf252961945f52df to avoid allocation failures intact.

I've further added a test that will catch any regression here in FORCE_DEBUG mode.

src/transaction/duck_transaction_manager.cpp
ywelsch: `last_commit` moves under read-only commits here, but that should be harmless: It can make the check fire when it didn't need to (the checkpoint is skipped and happens on a later commit), but it can't make the check miss a transaction that still needs the pre-commit state. We should address this once https://github.com/duckdb/duckdb/pull/24764 lands.

Mytherin:
Thanks!

PR #23643, opened 2026-07-06
Use all hardware threads during benchmarks

smvv:
Remove `--threads` to use all CPU cores on namespace CI runners (16 vCPU, 32 GB RAM).

Increase timed runs from 5 to 20 for tpc-h/tpc-ds queries to help stabilize the benchmark results.

### DIFF
--- a/scripts/regression/benchmark.py
+++ b/scripts/regression/benchmark.py
@@ -37,6 +37,7 @@ class BenchmarkRunnerConfig:
     max_timeout: int = MAX_TIMEOUT
     root_dir: str = ""
     no_summary: bool = False
+    timed_runs: Optional[int] = None
 
     @classmethod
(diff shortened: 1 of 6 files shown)


scripts/regression/benchmark.py
```diff
@@ -105,6 +110,23 @@ def __init__(self, config: BenchmarkRunnerConfig):
         self.benchmark_list: List[str] = []
         with open(self.config.benchmark_file, 'r') as f:
             self.benchmark_list = [x.strip() for x in f.read().split('\n') if len(x) > 0]
+        self.supports_timed_runs = self.detect_timed_runs_support()
+
+    def detect_timed_runs_support(self):
```
smvv, line 115: note: this is needed for the "regression base build" that does not (yet) support the CLI flag `--timed-runs` added in this PR.

PR #23646, opened 2026-07-06
Packrat parsing

Dtenwolde:
Fix: https://github.com/duckdblabs/duckdb-internal/issues/9259
Fix: https://github.com/duckdblabs/duckdb-internal/issues/6965
Fix: https://github.com/duckdb/duckdb-fuzzer/issues/4625

Thanks @c-herrewijn for finding these cases :)

This PR introduces packrat-style memoization to the PEG parser to avoid exponential backtracking in some cases. 

The parser now memoizes selected grammar rules (listed in `grammar_types.yml` and generated) by `matcher_id, token_idx`. So if a same matcher is evaluated at the same token position, it will instead reuse the previous result instead of recursively parsing the same subtree again. 

Consider the following as the most extreme case: 
```sql 
SELECT (((((((((((((((((;
```
Before this PR, this would cause parse time to grow exponentially because the parser repeatedly tried the same expression rules. With memoization used only for high-impact rules, this is no longer causes exponential runtime. I did measurements before starting this PR to gather high-impact rules and got inspiration from the author of `cpp_peglib`: https://github.com/duckdb/duckdb/discussions/21297. 

Currently the two most important rules are `Expression` and its subtree and `ColumnReference` & `Identifier`.
 
Each memoized matcher gets a stable `matcher_id` when the grammar matchers are built. The
packrat cache key is `(matcher_id, token_idx)`. Cached successful results store an `optional_ptr<ParseResult>` into the per-parse `ParseResultAllocator`, so cache hits can safely reuse the parse result by pointer within a single parse run. No  copying is needed because the allocator owns all parse results for the lifetime of that parse.

Some rough measurements using wall clock timing: 
| Workload | Packrat on | Packrat off | Delta | Speedup |
| bracket syntax error | 0.010s | 108.510s | -108.500s | ~10851x |
| `duckdb_functions()`, 200 queries | 1.100s | 1.242s | -0.142s | 1.13x |
| TPC-H SF0.01, q1-q22 | 0.104s | 0.128s | -0.024s | 1.23x |
| TPC-DS SF0.01, q1-q99 | 0.808s | 1.076s | -0.268s | 1.33x |

From https://github.com/duckdb/duckdb-fuzzer/issues/4625: 
On `main`: Run Time (s): real 10.152 user 9.563358 sys 0.561430
On `packrat-parsing`: Run Time (s): real 0.011 user 0.010642 sys 0.000172


Dtenwolde:
I don't think the `MinGW` failure is related

Dtenwolde:
I will probably have to update with main after https://github.com/duckdb/duckdb/pull/23578 is merged

PR #23647, opened 2026-07-06
Fix check in clustered aggregation

lnkuiper:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9878


hannes:
thanks that was fast!

hannes:
CC @peterboncz

PR #23648, opened 2026-07-06
Bump HTTPFS

lnkuiper:


carlopi:
I think draft/undraft is needed to start CI (that I would assume to not pass since there is an extra setting)

PR #23649, opened 2026-07-06
Create alpha releases on main CI run

smvv:
A daily, alpha release workflow could be:

- For each "nightly run": run CI for duckdb main with a version (like `v2.0.0-alpha42`).
- When duckdb CI succeeded, notify clients (Python, Java, etc) and extensions to build for that version.

Future work:
- At the end of the day, collect what clients and extensions are built and tested.
- If 80%(?) passed automated testing, tag the version and notify clients/extensions so they publish.


carlopi:
I reverted the auto-enrolment via https://github.com/duckdb/duckdb/pull/23755

Issue #23641, opened 2026-07-06
CUME_DIST with argument ORDER BY can underflow on future ROWS frames


Yibo-Dong:
### What happens?

CUME_DIST with an argument ORDER BY can underflow when the explicit ROWS frame starts after the current peer group.

This happens on current main (3d164e405e) and also on v1.5.4.

### To Reproduce

To reproduce:

```sql
WITH t(id) AS (
  VALUES (1), (2), (3)
)
SELECT
  id,
  CUME_DIST(ORDER BY id) OVER (
    ORDER BY id
    ROWS BETWEEN 2 FOLLOWING AND 2 FOLLOWING
  ) AS actual,
  0.0 AS expected
FROM t
ORDER BY id;
```

Run:

```sh
./build/release/duckdb -f bug.sql
```

Actual:

stdout:
| id | actual | expected |
| --- | --- | --- |
| 1 | 1.8446744073709552e+19 | 0.0 |
| 2 | 0.0 | 0.0 |
| 3 | 0.0 | 0.0 |

Expected:

| id | actual | expected |
| --- | --- | --- |
| 1 | 0.0 | 0.0 |
| 2 | 0.0 | 0.0 |
| 3 | 0.0 | 0.0 |

For row `id = 1`, the explicit frame contains only row `id = 3`. Under the
argument ordering `ORDER BY id`, no row in the frame has `id <= 1`, so the
`CUME_DIST` numerator is zero and the result should be `0.0`, not a huge value
from unsigned arithmetic.

Notes From Minimization:

- `2 PRECEDING` is correct.
- `ORDER BY` is required.
---

Cause Analysis (generated by coding agents):

The failure is in `WindowCumeDistExecutor::GetData` in `src/function/window/window_rank_function.cpp`: with argument-local ordering, DuckDB computes `peer_end` from the argument-order token tree and then calls `CumeDist(frame_begin, frame_end, peer_end)`, whose numerator is `peer_end - begin` in unsigned `idx_t`; when a future-only frame contains no peer at or below the current row, `peer_end < frame_begin`, so the subtraction underflows and becomes the huge floating-point result shown by the repro.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (3d164e405e)

### DuckDB Client:

CLI

### Full Name:

Yibo Dong

### Affiliation:

National University of Singapore

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

[...]

PR #23651, opened 2026-07-06
Issue #23641: CUME_DIST Underflow

hawkfish:
* Clamp numerator.


DinosL (approved):
LGTM! I made similar changes in https://github.com/duckdb/duckdb/pull/23637. Maybe we should check all window functions for similar problems, but that's for another PR.

PR #23653, opened 2026-07-06
Default vacuuming in the presence of indexes

artjomPlaunov:
Make vacuuming the default for tables with only ART indexes. 

At a slightly higher level (before talking about the actual remapping), this continues on this PR https://github.com/duckdb/duckdb/pull/22822 that implemented rowid gap support, which had to consider the combinations between the rebuild indexes setting + legacy storage. This builds on that, although to try and simplify it somewhat I added a ```VacuumIndexStrategy``` that once set, determines the vacuuming strategy for indexes:

```cpp
//! How checkpoint vacuum handles table indexes when rowids may change.
enum class VacuumIndexStrategy : uint8_t {
	//! Indexes require stable rowids.
	KEEP_ROW_IDS,
	//! No index maintenance is needed.
	NO_INDEXES,
	//! Rebuild all indexes after vacuum.
	REBUILD,
	//! Incrementally remap affected ART rowids.
	REMAP
};
```

Assuming the REMAP strategy is possible, we do remapping on a per vacuum task basis. Shifted rows are buffered during a vacuum task, storing all the indexed key columns as well as the old + new rowids, and the remap is applied to each ART after the vacuum task. Note that it would have been possible to just store the old rowid and re-derive the new rowid on the fly, but storing both allows us to check for equality and avoid doing a remap in scenarios where the shifted rowid happens to keep its rowid; this just simplifies having to check for this condition. 

Currently takes a very coarsely guarded vacuum lock on index scans, similar to the experimental rebuild setting. I was going to try to remove the lock in this PR, but I suspect the cleanest approach here may be to wait for @sebastiaan-dev's Index lifetime + shadow-art PR to land, and maybe work in vacuuming with the shadow ART so we can publish a consistent (ART, RowGroup) pairing on checkpoint while letting a concurrent scan to read the pre-checkpoint state (just an initial idea, not sure if this will work out).


src/storage/table/row_group_collection.cpp
```diff
@@ -1292,12 +1341,153 @@ class CheckpointTask : public BaseCheckpointTask {
...
+	}
+
+private:
+	void BufferShiftedRows(DataChunk &scan_chunk, idx_t new_rowid_start) {
+		const auto count = scan_chunk.size();
+		const auto rowid_column_index = table_types.size();
+
+		UnifiedVectorFormat old_rowid_data;
+		scan_chunk.data[rowid_column_index].ToUnifiedFormat(old_rowid_data);
+		auto old_rowids = UnifiedVectorFormat::GetData<row_t>(old_rowid_data);
```
Mytherin: nit: can we use the .Values<row_t> iterator here?

src/storage/table/row_group_collection.cpp
```diff
@@ -1292,12 +1341,153 @@ class CheckpointTask : public BaseCheckpointTask {
...
+			for (idx_t col_idx = 0; col_idx < mapped_column_ids.size(); col_idx++) {
+				const auto col_id = mapped_column_ids[col_idx].GetPrimaryIndex();
+				table_chunk.data[col_id].Reference(scan_chunk.data[col_idx]);
+			}
+		};
+
+		for (auto &art_ref : indexes) {
+			auto &art = art_ref.get();
+			IndexLock index_lock;
+			art.InitializeLock(index_lock);
```
Mytherin: Does this even need to be ART specific? Don't indexes generically support delete / insert?

artjomPlaunov:
Looking into TSAN + CI failures, related to this PR

Mytherin (commented):
Thanks for the PR! Looks good - some minor comments

artjomPlaunov:
@Mytherin applied changes, still waiting to see if TSAN pops up, currently trying to see if I can recreate it on existing main (perhaps a new test uncovers something, in which case I will open a separate issue for that)

src/main/attached_database.cpp
```diff
@@ -258,10 +258,16 @@ void AttachedDatabase::InvokeCloseIfLastReference(shared_ptr<AttachedDatabase> &
...
-	if (attached_db.use_count() == 1) {
-		attached_db->Close(close_action);
+	{
+		lock_guard<mutex> guard(*close_lock);
+		if (attached_db.use_count() != 1) {
+			attached_db.reset();
+			return;
+		}
 	}
+	// we hold the last reference: no other thread can obtain a reference to the database anymore,
```
artjomPlaunov: @Mytherin tried out this refactor to get rid of a TSAN warning in a new test I introduced, does this make sense, I think it should be OK here to not have to hold the close lock once we already checked the use count was 1 under the lock?

artjomPlaunov:
applied review changes, opened one review comment about refactoring the close lock in InvokeCloseIfLastReference to get rid of a tsan warning from a new test.

artjomPlaunov:
@Mytherin @taniabogatsch (mentioning for a review on the InvokeCloseIfLastReference refactor) this is ready again; notes for the InvokeCloseIfLastReference refactor:

I was getting a TSAN due to a test having both a wal_lock -> close_lock and close_lock -> wal_lock (although not concurrently). I could look more into the deadlock to see if it is fine to add a suppression, but the refactor reduces the scope to checking the count and decrementing it if we are not the last owner; it would be better to not add a suppression to this function if possible, to catch any potential problems with new code that is introduced.

If we are *are* the last owner, i.e. use_count == 1 which was observed under the close_lock, it is safe to do the close and reference count decrement outside of the close_lock. Once we read use_count == 1 under the close_lock, the count cannot change until the same thread calls reset(). 

For the reference count to rise we would need: 

A) a copy from the sole owner (but since we're the only owner, that doesn't happen)
B) shared_from_this, but we don't do this from our sole reference (and DetachInternal removed it from the set of attached databases; but even if it was in there the count would be >= 2). 

Or C from a weak_ptr, which I am still unsure about, looking into the weak_ptr in ClientContext right now: 

```cpp
	//! Routing target for SQL execution while CONNECT-ed (CONNECT/DISCONNECT). When is_connected is
	//! true and connected_to_database can be locked, the chokepoint dispatches non-control SQL via
	//! `Catalog::RemoteExecute(string)` and wraps the returned TableRef into a SelectStatement.
	weak_ptr<AttachedDatabase> connected_to_database;
```

Mytherin:
Sounds good - agreed it's better not to mess with threadsan suppressions unless we have an extremely good reason. Code changes look good to me so good to go from my side once CI passes.

Mytherin:
For `connected_to_database` I think we can just add an `InvokeCloseIfLastReference` to `ClientContext::PendingStatementOrPreparedStatement` which is the only place where this is referenced.

artjomPlaunov:
ok, will re-introduce the lock refactor. Will add the other fix for the weak_ptr in another PR

Mytherin:
Thanks!

PR #23658, opened 2026-07-07
Correlated non-`INNER` joins

lnkuiper:
This PR makes subqueries in non-`INNER` join predicates go through the existing subquery/dependent-join planning machinery instead of being rejected.

This now covers side-local subqueries and predicates that depend on both join inputs for `LEFT`, `RIGHT`, and `FULL OUTER` joins, as well as `SEMI` and `ANTI` joins. Pair-dependent predicates may also reference an enclosing query.

For example, this previously failed during binding:

```sql
SELECT *
FROM l
LEFT JOIN r ON EXISTS (
	SELECT 1
	FROM s
	WHERE s.x = l.x AND s.y = r.y
);
```

## Implementation

The implementation handles three cases:

1. Side-local subqueries are planned into the join input they depend on
2. Pair-dependent non-`FULL` predicates
3. Pair-dependent `FULL OUTER` predicates

`FlattenDependentJoins` now carries an explicit `UnnestingState` containing the active domain bindings and binding replacements. This lets the existing decorrelator propagate the state through joins, projections, aggregates, windows, set operations, CTEs, and nested dependent joins.

While testing this, I also fixed two issues exposed by the new plans: filter pushdown could leak `MARK`-join binding state between sibling subtrees, and the hash join empty-build path could reuse a probe chunk with the wrong schema when residual predicates required extra probe columns.

Also added many, many generated tests that were verified against PostgreSQL.


Tishj:
Related to https://github.com/duckdb/duckdb/pull/23599 ?

lnkuiper:
Yes, related but both are useful :)

EDIT: I'm waiting until #23599 is merged until I fix up and merge this one

src/execution/operator/join/physical_hash_join.cpp
```diff
@@ -1928,6 +1930,9 @@ unique_ptr<OperatorState> PhysicalHashJoin::GetOperatorState(ExecutionContext &c
 	if (!lhs_probe_columns.col_types.empty()) {
 		state->lhs_probe_data.Initialize(allocator, lhs_probe_columns.col_types);
 	}
+	if (!lhs_output_columns.col_types.empty()) {
+		state->lhs_output_data.Initialize(allocator, lhs_output_columns.col_types);
```
lnkuiper: Nit: this should probably use `InitializeEmpty` (and we should do a check for other `DataChunk` in this file (and `JoinHashTable`) that only ever reference columns, never need to allocate them.

kryonix (changes requested):
### Pair-dependent `FULL OUTER JOIN` leaves stale bindings in enclosing dependent joins

The pair-dependent full-join rewrite replaces the join's public bindings in src/planner/subquery/join_condition_subquery_rewriter.cpp:397. Those replacements are applied to the parent by `src/planner/binder/query_node/plan_subquery.cpp:552`, but only through VisitOperatorBindings.

For a `LogicalDependentJoin`, that updates correlated_columns but does not traverse the right-hand dependent subtree, as shown in `src/optimizer/column_binding_replacer.cpp:79`. Expressions in that subtree therefore continue referencing the full join's old output bindings.

Minimal reproducer:

```sql
SELECT l.a, r.b, (SELECT l.a + r.b) AS z
FROM (VALUES (1), (2)) l(a)
FULL OUTER JOIN (VALUES (10), (20)) r(b)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND r.b = 10)
ORDER BY ALL;
```

Expected:

```
1     10    11
2     NULL  NULL
NULL  20    NULL
```

Actual:
- Default optimizer: `Invalid Error: unordered_map::at: key not found`
- Optimizer disabled: `internal Failed to bind column reference "a" [2.0]`

The defect also reproduces with:

- An enclosing correlated WHERE EXISTS
- LEFT JOIN LATERAL
- A subsequent pair-dependent LEFT, SEMI, or ANTI join
- Chained joins, where it can instead fail in filter pushdown with a projection table-index mismatch

The replacement application at `src/planner/binder/query_node/plan_subquery.cpp:604` needs to know which child changed. When child 0 of a dependent join changes, replacements must also traverse the parent's right subtree while stopping at the changed child.

### Predicate-only correlation is unsupported for SEMI and ANTI

PushDownJoin determines correlation solely from the two child subtrees at `src/planner/subquery/flatten_dependent_join.cpp:1018`. When the outer reference appears only in the join condition, both `left_has_correlation` and `right_has_correlation` are false.

The `SEMI/ANTI` handling at `src/planner/subquery/flatten_dependent_join.cpp:1037` has branches for left, right, and both children, but not neither. Execution consequently reaches the unsupported-join exception at line 1092.

Reproducer:

```sql
SELECT o.x, EXISTS (
  SELECT *
  FROM (VALUES (1)) l(a)
  SEMI JOIN (VALUES (2)) r(b)
    ON l.a = 1 AND o.x = 10
) AS ex
FROM (VALUES (10), (20)) o(x)
ORDER BY o.x;
```

Expected:
```
10  true
20  false
```

Actual under default, optimizer-disabled, and CTE-delim modes:

```
Not implemented Error:
Unsupported join type for flattening correlated subquery
```

The equivalent `ANTI` query fails identically. Its expected booleans are false, true.

The active delim state should be attached to the output/left side for this fourth case, included in the left projection map, and used to rewrite the predicate while the right subtree remains independently decorrelated.

### Missing upstream NULL-semantics fix

This PR does not contain upstream commit 79f50ef263, "Preserve NULL semantics in CTE deliminator". The current implementation at `src/planner/subquery/delim_join_cte_rewriter.cpp:981` can fold a regular `<>` predicate into `IS DISTINCT FROM` for a MARK join, turning SQL UNKNOWN into true for nullable probe values. Currently, an incorrect `true` is returned for a nullable `wh <> s2.wh` correlated `EXISTS` query with `delim_join_as_cte=true`.

This is pre-existing relative to the PR's merge base, not introduced by the PR. However, this PR substantially changes the same rewriter, so the upstream nullability analysis must be reconciled during rebase and its regression test retained.

```
# name: test/sql/join/non_inner_decorrelation_regressions.test
# description: Regression tests for non-inner join decorrelation
# group: [join]

# Pair-dependent FULL OUTER JOIN followed by a correlated scalar subquery.
query III
SELECT l.a, r.b, (SELECT l.a + r.b) AS z
FROM (VALUES (1), (2)) l(a)
FULL OUTER JOIN (VALUES (10), (20)) r(b)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND r.b = 10)
ORDER BY ALL;
----
1     10      11
2     NULL    NULL
NULL  20      NULL

# Pair-dependent FULL OUTER JOIN followed by a correlated EXISTS filter.
query II
SELECT l.a, r.b
FROM (VALUES (1), (2)) l(a)
FULL OUTER JOIN (VALUES (10), (20)) r(b)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND r.b = 10)
WHERE EXISTS (SELECT 1 WHERE l.a + r.b = 11)
ORDER BY ALL;
----
1     10

# Pair-dependent FULL OUTER JOIN followed by a lateral subquery.
query III
SELECT l.a, r.b, x.z
FROM (VALUES (1), (2)) l(a)
FULL OUTER JOIN (VALUES (10), (20)) r(b)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND r.b = 10)
LEFT JOIN LATERAL (SELECT l.a + r.b AS z) x ON true
ORDER BY ALL;
----
1     10      11
2     NULL    NULL
NULL  20      NULL

# Pair-dependent FULL OUTER JOIN followed by another dependent LEFT JOIN.
query III
SELECT l.a, r.b, s.c
FROM (VALUES (1), (2)) l(a)
FULL OUTER JOIN (VALUES (10), (20)) r(b)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND r.b = 10)
LEFT JOIN (VALUES (100), (200)) s(c)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND s.c = 100)
ORDER BY ALL;
----
1     10      100
2     NULL    NULL
NULL  20      NULL

# Pair-dependent FULL OUTER JOIN followed by a dependent SEMI JOIN.
query II
SELECT l.a, r.b
FROM (VALUES (1), (2)) l(a)
FULL OUTER JOIN (VALUES (10), (20)) r(b)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND r.b = 10)
SEMI JOIN (VALUES (100), (200)) s(c)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND s.c = 100)
ORDER BY ALL;
----
1     10

# Pair-dependent FULL OUTER JOIN followed by a dependent ANTI JOIN.
query II
SELECT l.a, r.b
FROM (VALUES (1), (2)) l(a)
FULL OUTER JOIN (VALUES (10), (20)) r(b)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND r.b = 10)
ANTI JOIN (VALUES (100), (200)) s(c)
  ON EXISTS (SELECT 1 WHERE l.a = 1 AND s.c = 100)
ORDER BY ALL;
----
2     NULL
NULL  20

# SEMI JOIN whose correlation exists only in the join condition.
query IT
SELECT o.x, EXISTS (
  SELECT *
  FROM (VALUES (1)) l(a)
  SEMI JOIN (VALUES (2)) r(b)
    ON l.a = 1 AND o.x = 10
) AS ex
FROM (VALUES (10), (20)) o(x)
ORDER BY o.x;
----
10    1
20    0

# ANTI JOIN whose correlation exists only in the join condition.
query IT
SELECT o.x, EXISTS (
  SELECT *
  FROM (VALUES (1)) l(a)
  ANTI JOIN (VALUES (2)) r(b)
    ON l.a = 1 AND o.x = 10
) AS ex
FROM (VALUES (10), (20)) o(x)
ORDER BY o.x;
----
10    0
20    1
```

kryonix (changes requested):
This looks already much better now. But there are a few things remaining:

> **[P1] Keep preserved-side predicates inside ANTI matching**
>
> Attaching the delim state only to the left input can produce an `ANTI` join with an `ON` conjunct that references only that state. Join ordering separates this conjunct from the `ANTI` edge: it evaluates `ANTI(l, r, l.a = r.b)` first and then filters on `o.x = 10`. This transformation is valid for `SEMI`, but not for `ANTI`: when `o.x <> 10`, the complete `ON` condition is false and the left row must survive.
>
> ```sql
> SELECT o.x, EXISTS (
>   SELECT *
>   FROM (VALUES (1)) l(a)
>   ANTI JOIN (VALUES (1)) r(b)
>     ON l.a = r.b AND o.x = 10
> ) ex
> FROM (VALUES (10), (20)) o(x)
> ORDER BY o.x;
> ```
>
> Expected: `(10, false), (20, true)`. With the normal optimizer and with `delim_join_as_cte=true`, the result is `(10, false), (20, false)`. Disabling `join_order` restores the correct result. Please preserve the correlated predicate as part of the `ANTI` match dependency and cover matching, non-matching, duplicate, and null-safe variants.

This interacts with a pre-existing join-order bug: the direct uncorrelated equivalent also fails on `origin/main`. However, `origin/main` rejects the correlated query, while this branch newly supports it and generates the vulnerable plan, so it still needs addressing before merge.

---

**`src/planner/subquery/flatten_dependent_join.cpp:1057`**

> **[P1] Reclassify join conditions after rewriting correlated bindings**
>
> At initial binding time, the depth-qualified `o.x` is not considered a child binding, so `l.a = r.b + o.x` is stored as a comparison condition with `r.b + o.x` on the RHS. This rewrite changes `o.x` into a delim binding emitted by the left child, leaving the RHS expression referencing both children. A comparison join requires each side to be resolvable against one child, so column binding verification fails.
>
> ```sql
> SET debug_verify_column_bindings=true;
>
> SELECT o.x, EXISTS (
>   SELECT *
>   FROM (VALUES (1)) l(a)
>   SEMI JOIN (VALUES (1)) r(b)
>     ON l.a = r.b + o.x
> ) ex
> FROM (VALUES (0), (1)) o(x)
> ORDER BY o.x;
> ```
>
> This raises `INTERNAL Error: Failed to bind column reference "x"`. The same failure occurs with the optimizer disabled or just `join_order` disabled, for both `SEMI` and `ANTI`. The normal optimized `SEMI` query only succeeds because join ordering rewrites the invalid condition; optimized `ANTI` then returns the `SEMI` result. After rewriting correlated references, please reclassify conditions so each comparison side belongs to one child, or represent cross-child expressions as arbitrary join predicates.

## Drop-in regression tests

```sql
# Predicate-only ANTI correlation must retain rows when a preserved-side
# conjunct makes the complete ON condition false.
query IT
SELECT o.x, EXISTS (
	SELECT *
	FROM (VALUES (1)) l(a)
	ANTI JOIN (VALUES (1)) r(b)
		ON l.a = r.b AND o.x = 10
) AS ex
FROM (VALUES (10), (20)) o(x)
ORDER BY o.x;
----
10	0
20	1

# The CTE-delim path reaches the same join-ordering pipeline.
statement ok
SET delim_join_as_cte=true;

query IT
SELECT o.x, EXISTS (
	SELECT *
	FROM (VALUES (1)) l(a)
	ANTI JOIN (VALUES (1)) r(b)
		ON l.a = r.b AND o.x = 10
) AS ex
FROM (VALUES (10), (20)) o(x)
ORDER BY o.x;
----
10	0
20	1

statement ok
SET delim_join_as_cte=false;

# Rewriting a correlated reference must not leave a comparison side
# referencing both join children.
statement ok
SET debug_verify_column_bindings=true;

query IT
SELECT o.x, EXISTS (
	SELECT *
	FROM (VALUES (1)) l(a)
	SEMI JOIN (VALUES (1)) r(b)
		ON l.a = r.b + o.x
) AS ex
FROM (VALUES (0), (1)) o(x)
ORDER BY o.x;
----
0	1
1	0

statement ok
SET debug_verify_column_bindings=false;

# The resulting plan must also work without join-order rewriting.
statement ok
PRAGMA disable_optimizer;

query IT
SELECT o.x, EXISTS (
	SELECT *
	FROM (VALUES (1)) l(a)
	SEMI JOIN (VALUES (1)) r(b)
		ON l.a = r.b + o.x
) AS ex
FROM (VALUES (0), (1)) o(x)
ORDER BY o.x;
----
0	1
1	0
```

kryonix (dismissed):
I found one new **P1 wrong-result blocker** on commit `7d0ac7d961`. The previously reported blockers are fixed.

### P1: Simplified conflict rules allow an INNER join to cross a LEFT join

[join_order_operator.cpp:313](/Users/denis/Documents/git/duckdb-fix-delim/src/optimizer/join_order/join_order_operator.cpp:313):

> **[P1] Do not split INNER occurrences after absorbing their conflict rules**
>
> An empty post-simplification rule set plus unchanged TES size does not prove that this INNER occurrence is independently movable. Here CD-C creates `b -> a` because the INNER join is above a LEFT join, but simplification removes that rule because `a` is already in the occurrence’s SES. This return then classifies the occurrence as non-exact, allowing `CreateHyperGraphEdges` to emit `b IS NOT DISTINCT FROM d` independently from `(d.v IS NULL OR a.v=0)`.
>
> The enumerator can consequently move `d` into the nullable side of the LEFT join. NULL-extension makes the residual true and returns a bogus row. The query below should return no rows, but current HEAD returns `2` under both exact and greedy join-order thresholds. Please preserve enough conflict provenance to keep the full semantic condition scoped to the operator occurrence when non-inner conflicts have been absorbed.

Reproducer:

```sql
WITH a(k,v) AS (VALUES (2,1)),
     b(k) AS (VALUES (3)),
     c(k) AS (VALUES (2)),
     d(k,v) AS (VALUES (NULL::INTEGER,7))
SELECT a.k
FROM a
LEFT JOIN b ON b.k=a.k
INNER JOIN c ON c.k=a.k
INNER JOIN d
  ON d.k IS NOT DISTINCT FROM b.k
 AND (d.v IS NULL OR a.v=0);
```

Results:

- Current default: `2`
- Threshold `0`: `2`
- Threshold `30`: `2`
- Join-order optimizer disabled: empty, correct
- `debug_force_no_cross_product=true`: incorrectly reports that the query requires a cross product

This also conflicts with the CD-C model, where applicability belongs to the complete binary-operator occurrence, not independently emitted fragments of its predicate. See [Moerkotte, Fender, and Eich’s CD-C paper](https://15721.courses.cs.cmu.edu/spring2019/papers/23-optimizer2/p493-moerkotte.pdf).

### NULL rewrite issue

The earlier NULL-rewrite issue still reproduces, and I did not find equivalent dedicated coverage. However, it also reproduces on the pre-PR baseline, so I would not block this PR on it:

```sql
statement ok
CREATE TABLE nn(k INTEGER, wh INTEGER NOT NULL);

statement ok
INSERT INTO nn VALUES (1, 5);

statement ok
CREATE TABLE ids(id INTEGER);

statement ok
INSERT INTO ids VALUES (1), (2);

statement ok
CREATE TABLE ev(id INTEGER, wh INTEGER NOT NULL);

statement ok
INSERT INTO ev VALUES (1, 6), (2, 7);

query IIB
SELECT ids.id, nn.wh,
       EXISTS (
           SELECT 1
           FROM ev
           WHERE ev.id = ids.id
             AND nn.wh <> ev.wh
       )
FROM ids
LEFT JOIN nn ON ids.id = nn.k
ORDER BY ids.id;
----
1	5	true
2	NULL	false
```

Current behavior incorrectly returns `true` for the second row.

Tmonster (approved):
LGMT, thanks!

PR #23659, opened 2026-07-07
Implement `CONNECT <name> EXECUTE <sql>`

carlopi:
This adds some surface, and that's the weak point of the PR, unsure how to work around, I could split a separate PR around adding parser capabilities if that makes it simpler.

The pieces, as reported by 🤖:
```
  1. PEG matcher primitive (matcher.cpp/.hpp, ~70 lines)
  A new InStatementTokenMatcher that consumes any single non-; token, so a grammar can grab opaque text up to a statement boundary. Plus standalone-grammar compilation (compile a small grammar independently of  
  the main one). (This is where the compile bug lived — the override was named MatchParseResult instead of MatchParseResultInternal; now fixed in PR A's commit 1.)

  2. Layer-1 connect-mode parser (connect_mode/, ~406 lines — the heart, 54% of PR A)
  A tiny separate PEG grammar that classifies each statement into CONTROL (CONNECT/DISCONNECT), EXECUTE (CONNECT <name> EXECUTE <payload>), FORBIDDEN (malformed connect forms), or RAW. For EXECUTE it slices out 
  the target name and the payload text verbatim from the source string. The payload never touches the DuckDB parser.

  3. PassthroughStatement (passthrough_statement.* + enum, ~80 lines)
  A new SQLStatement subtype holding two strings: target + payload. Pure data holder (Copy/ToString). New PASSTHROUGH_STATEMENT enum value threaded through the dispatch switches.

  4. Passthrough binder (bind_passthrough.cpp, ~42 lines)
  Binds a PassthroughStatement by resolving the target catalog at bind time, checking it Supports(RemoteCapability::CONNECT), calling Catalog::RemoteExecute(payload) to get a TableRef, then wrapping it as SELECT
  * FROM <remote-ref> and binding through the normal SELECT path. The RemoteExecute base just throws — the real backend is out-of-tree (the quack/postgres fork).

  5. Parse-pipeline wiring (parse_iterator.cpp + client_context.cpp, ~70 lines)
  ParseIterator walks the token stream; when a statement starts with CONNECT/DISCONNECT it runs the Layer-1 classifier, and on the EXECUTE shape emits a PassthroughStatement instead of parsing. client_context   
  marks PASSTHROUGH as connect-control so the sticky chokepoint won't re-route it.
```

The biggest part here is 2, with the rest plumbing it though. Positive note, this code is triggered only in `CONNECT <name> EXECUTE <sql>` cases, and doesn't expand to other code-paths.
Testing this without `quack` or `postgres` makes little sense, but once postgres is build (and loaded), this allows for example
```sql
ATTACH '....' AS pg;
CONNECT pg EXECUTE CREATE USER ...;
```


carlopi:
Fixed up alternative at https://github.com/duckdb/duckdb/pull/23661

PR #23660, opened 2026-07-07
Add `lttb` aggregate function

Maxxen:
This PR adds a new `lttb(x, y, n ORDER BY x)` aggregate function that downsamples a series of (x, y) points to at most `n` points using the "Largest-Triangle-Three-Buckets" (LTTB) algorithm. This is commonly used to reduce large time series for plotting while preserving their visual shape. 

The function returns a `LIST(STRUCT(x, y))`, which can be `UNNEST`ed back into rows, and always keeps the first and last point. The x and y input each accept `FLOAT`, `DOUBLE`, or any `TIMESTAMP` type, but the actual triangle computations used to select the points are performed using `double` arithmetic internally.

Some things to keep in mind:
- The input _needs_ to be sorted on the x-axis, or the output will produce garbage. The `ORDER BY X` in lttb(x, y, n ORDER BY x) is basically mandatory.
- Points with `NULL`s are ignored.
- Empty groups return `NULL`, like `list()`
- `n` needs to be a constant and `n >= 2`.
  
The implementation buffers the points by reusing the list() aggregate's linked-list machinery, which automatically gives us support for e.g. clustered update and `EXPORT_STATE`.


extension/core_functions/aggregate/holistic/lttb.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+			// Average point in the next bucket (never empty: v > n implies width > 1)
+			D_ASSERT(next_end > next_beg);
+			const auto next_len = next_end - next_beg;
+
+			double avg_x = 0;
+			double avg_y = 0;
+			for (idx_t j = next_beg; j < next_end; j++) {
+				avg_x += static_cast<double>(vx[j]);
+				avg_y += static_cast<double>(vy[j]);
```
lnkuiper: Casting absolute TIMESTAMP_NS values directly to double loses nanosecond-scale differences for normal epoch values. This changes LTTB selection: the same series using relative DOUBLE x-values picks different points than TIMESTAMP_NS values around 2020. The timestamp axes should be normalized before double math.

extension/core_functions/aggregate/holistic/lttb.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+//! Adds the lttb overloads for one x type: y can be FLOAT, DOUBLE or any of the TIMESTAMP types
+template <class XTYPE>
+void AddLTTBFunctions(AggregateFunctionSet &set, const LogicalType &x_type) {
+	set.AddFunction(GetLTTBFunction<XTYPE, float>(x_type, LogicalType::FLOAT));
+	set.AddFunction(GetLTTBFunction<XTYPE, double>(x_type, LogicalType::DOUBLE));
+	set.AddFunction(GetLTTBFunction<XTYPE, int64_t>(x_type, LogicalType::TIMESTAMP));
+	set.AddFunction(GetLTTBFunction<XTYPE, int64_t>(x_type, LogicalType::TIMESTAMP_S));
+	set.AddFunction(GetLTTBFunction<XTYPE, int64_t>(x_type, LogicalType::TIMESTAMP_MS));
+	set.AddFunction(GetLTTBFunction<XTYPE, int64_t>(x_type, LogicalType::TIMESTAMP_NS));
```
lnkuiper: This misses `TIMESTAMP_TZ` and `TIMESTAMP_TZ_NS`, is that on purpose?

extension/core_functions/aggregate/holistic/lttb.cpp
```diff
@@ -0,0 +1,359 @@
...
+		const auto v = state.linked_list.total_capacity;
+
+		if (v == 0) {
+			// Empty list (or all points filtered out as NULL)
+			FlatVector::SetNull(result, rid, true);
+			continue;
+		}
+
+		// Materialize the buffered (x, y) points (kept in insertion order, i.e. ordered by x)
+		Vector points(struct_type, v);
```
lnkuiper: This is a fresh vector per finalize (and a selection vector above) can we somehow cache/reuse to avoid repeated allocation?

lnkuiper (commented):
Thanks for the PR! Looks good. I've left some comments, and here's a repro for the comments:
```sql
# TIMESTAMPTZ is a timestamp type and should be accepted if lttb supports timestamp axes.
statement ok
SELECT lttb(ts, y, 3 ORDER BY ts)
FROM (
	VALUES
		(TIMESTAMPTZ '2020-01-01 00:00:00+00', 0::DOUBLE),
		(TIMESTAMPTZ '2020-01-01 01:00:00+00', 1::DOUBLE),
		(TIMESTAMPTZ '2020-01-01 02:00:00+00', 2::DOUBLE),
		(TIMESTAMPTZ '2020-01-01 03:00:00+00', 3::DOUBLE),
		(TIMESTAMPTZ '2020-01-01 04:00:00+00', 4::DOUBLE)
) t(ts, y);

# LTTB point selection should be invariant under translating the x-axis.
query B
WITH data AS (
	SELECT i,
	       (CASE WHEN i = 37 THEN 1000 ELSE i % 7 END)::DOUBLE AS y,
	       make_timestamp_ns(1577836800000000000 + i) AS ts
	FROM range(100) t(i)
),
num AS (
	SELECT UNNEST(lttb(i::DOUBLE, y, 10 ORDER BY i)) AS p
	FROM data
),
ns AS (
	SELECT UNNEST(lttb(ts, y, 10 ORDER BY ts)) AS p
	FROM data
)
SELECT (SELECT list(p.x ORDER BY p.x) FROM num)
       = (SELECT list((epoch_ns(p.x) - 1577836800000000000)::DOUBLE ORDER BY epoch_ns(p.x)) FROM ns)
----
true
```

extension/core_functions/aggregate/holistic/lttb.cpp
Maxxen: Hmm, maybe I misunderstand the timestamp types - but I thought that they were already stored normalized in their int64_t form? But ok, I'll review.

extension/core_functions/aggregate/holistic/lttb.cpp
Maxxen: No, just forgot :) I guess we can add DATE too

extension/core_functions/aggregate/holistic/lttb.cpp
Maxxen: I did consider this but wasn't sure how to make a vector with an arbitrary Allocator now, and setting up a data chunk also doesn't really work since the resizing logic has all changed now that vectors have their own count. But maybe we can at least re-use the vector even if its not arena allocated - I'll have a look.

lnkuiper: If it's too much hassle this is OK for now, not a problem if there aren't many groups. You could potentially allocate the vectour outside of the loop, that would already help

extension/core_functions/aggregate/holistic/lttb.cpp
lnkuiper: I'm also not 100% sure, the time lord probably knows best. This is just a repro that Codex came up with

breckcs:
@reformovo has added a community extension for LTTB: https://duckdb.org/community_extensions/extensions/lttb.

Maxxen:
@lnkuiper Thanks again for the feedback. I've now 
- added normalization for the timestamp types
- added overloads for timestamp_tz* 
- removed a bunch of y-axis overloads (I think just float/double is enough for now)
- made the materialized points vector reusable between groups

Maxxen:
@lnkuiper got bumped out of the merge queue

PR #23661, opened 2026-07-07
Implement `CONNECT <name> EXECUTE <sql>`

carlopi:
This adds some surface, and that's the weak point of the PR, unsure how to work around, I could split a separate PR around adding parser capabilities if that makes it simpler.

The pieces, as reported by 🤖:
```
  1. PEG matcher primitive (matcher.cpp/.hpp, ~70 lines)
  A new InStatementTokenMatcher that consumes any single non-; token, so a grammar can grab opaque text up to a statement boundary. Plus standalone-grammar compilation (compile a small grammar independently of  
  the main one). (This is where the compile bug lived — the override was named MatchParseResult instead of MatchParseResultInternal; now fixed in PR A's commit 1.)

  2. Layer-1 connect-mode parser (connect_mode/, ~406 lines — the heart, 54% of PR A)
  A tiny separate PEG grammar that classifies each statement into CONTROL (CONNECT/DISCONNECT), EXECUTE (CONNECT <name> EXECUTE <payload>), FORBIDDEN (malformed connect forms), or RAW. For EXECUTE it slices out 
  the target name and the payload text verbatim from the source string. The payload never touches the DuckDB parser.

  3. PassthroughStatement (passthrough_statement.* + enum, ~80 lines)
  A new SQLStatement subtype holding two strings: target + payload. Pure data holder (Copy/ToString). New PASSTHROUGH_STATEMENT enum value threaded through the dispatch switches.

  4. Passthrough binder (bind_passthrough.cpp, ~42 lines)
  Binds a PassthroughStatement by resolving the target catalog at bind time, checking it Supports(RemoteCapability::CONNECT), calling Catalog::RemoteExecute(payload) to get a TableRef, then wrapping it as SELECT
  * FROM <remote-ref> and binding through the normal SELECT path. The RemoteExecute base just throws — the real backend is out-of-tree (the quack/postgres fork).

  5. Parse-pipeline wiring (parse_iterator.cpp + client_context.cpp, ~70 lines)
  ParseIterator walks the token stream; when a statement starts with CONNECT/DISCONNECT it runs the Layer-1 classifier, and on the EXECUTE shape emits a PassthroughStatement instead of parsing. client_context   
  marks PASSTHROUGH as connect-control so the sticky chokepoint won't re-route it.
```

The biggest part here is 2, with the rest plumbing it though. Positive note, this code is triggered only in `CONNECT <name> EXECUTE <sql>` cases, and doesn't expand to other code-paths.
Testing this without `quack` or `postgres` makes little sense, but once postgres is build (and loaded), this allows for example
```sql
ATTACH '....' AS pg;
CONNECT pg EXECUTE CREATE USER ...;
```


src/include/duckdb/parser/connect_mode/connect_mode_parser.hpp
```diff
@@ -0,0 +1,103 @@
...
+//!    RemoteExecute. When not CONNECT-ed, fed to the main SQL parser as usual.
+struct ConnectModeChunk {
+	enum class Type : uint8_t { CONTROL, EXECUTE, FORBIDDEN, RAW };
+
+	//! For FORBIDDEN chunks, which malformed shape did we recognize? Lets the dispatcher format a
+	//! diagnostic that names the specific problem ("extra tokens after CONNECT target" vs
+	//! "non-identifier after CONNECT") instead of a generic message.
+	enum class ForbiddenReason : uint8_t {
+		NONE,
+		//! `CONNECT <ident|LOCAL|'string'> <more tokens>` — well-formed prefix plus extra junk.
```
Dtenwolde, line 43: Out of curiosity, do we need all these different forbidden reasons? Wouldn't we better let this a) pass the parser, catch there's extra input of some kind and throw a descriptive error in the transformer or b) throw a generic parser error (e.g. "Expected end of input, instead found xyz")

src/include/duckdb/parser/statement/passthrough_statement.hpp
```diff
@@ -0,0 +1,42 @@
...
+//! catalogs attached between Extract and Prepare/Bind are picked up.
+class PassthroughStatement : public SQLStatement {
+public:
+	static constexpr const StatementType TYPE = StatementType::PASSTHROUGH_STATEMENT;
+
+public:
+	PassthroughStatement(string target_p, string payload_p);
+
+	//! Catalog name to dispatch against (e.g. "pg", "qk").
+	string target;
```
Dtenwolde, line 30: Should this be `Identifier` to make it case-insensitive?

src/main/parse_iterator.cpp
```diff
@@ -1,18 +1,72 @@
...
+		return false;
+	}
+	vector<ConnectModeChunk> chunks;
+	try {
+		ConnectModeParser layer1(query);
+		chunks = layer1.AllRemaining();
+	} catch (...) {
+		// Layer-1 couldn't classify the input — let the normal SQL parser produce its (better) error.
+		return false;
+	}
```
Dtenwolde, line 40: I don't think we want these generic `catch (...)` handlers. But could also be my personal preference

src/include/duckdb/parser/connect_mode/connect_mode_parser.hpp
carlopi: Currently, this is not exercized, so yes, would be a great candidate for streamlining the code.

My idea is though that once this ships, it's hard to ADD more syntax, like, all syntax that one might possibly want to catch has to be known in advance, otherwise over time it changes what's parsed locally vs remotely.

( this is different than expanding regular parsing, where tradeoff is accepting more stuff that would otherwise throw )

But maybe this is a nit / irrelevant.

src/main/parse_iterator.cpp
```diff
@@ -1,18 +1,72 @@
...
+			has_execute = true;
+			break;
+		}
+	}
+	if (!has_execute) {
+		// No EXECUTE form present — let the normal parser handle the whole query (it owns the error
+		// messages for plain/malformed CONNECT and DISCONNECT).
+		return false;
+	}
+	for (auto &chunk : chunks) {
```
Dtenwolde, line 53: This and L42 also seem like we're doing double work? First check if we have an `EXECUTE`, then put the `EXECUTE` in `PassthroughStatement`. Otherwise let the standard parser handle them. Why not combine these two?

src/parser/connect_mode/connect_mode.gram
```diff
@@ -0,0 +1,50 @@
...
+# Catch malformed control statements and surface a clear error rather than a confusing parse failure
+# or silent raw forwarding.
+ForbiddenConnect    <- 'CONNECT' Raw
+ForbiddenDisconnect <- 'DISCONNECT' Raw
+
+# RawChunk MUST NOT start with CONNECT/DISCONNECT — those go to the Forbidden* paths above.
+# (Note: the `!` negative-lookahead is silently ignored by DuckDB's PEG runtime today, so this
+# guard is documentation-only. ConnectChunk/DisconnectChunk being tried first via ordered choice
+# is what actually prevents the misclassification at runtime.)
+RawChunk         <- !ControlKeyword Raw
```
Dtenwolde, line 38: Seems like we should add support for `!` in the matcher

Dtenwolde: Not in this PR necessarily

src/parser/connect_mode/connect_mode.gram
```diff
@@ -0,0 +1,50 @@
...
+ControlKeyword   <- 'CONNECT' / 'DISCONNECT'
+
+# `InStatementToken` is a built-in primitive: any non-TERMINATOR token. Lets us capture opaque
+# text up to a `;` without needing a `!` negative-lookahead operator.
+Raw              <- InStatementToken+
+Semi             <- ';'
+
+# Primitives (copy-pasted; lift to a shared file when a second Layer-1 grammar shows up)
+Identifier        <- QuotedIdentifier / PlainIdentifier
```
Dtenwolde, line 47: Do we need these? Identifier is overwritten in the `Matcher`, not sure how it is handled here

src/parser/connect_mode/connect_mode_parser.cpp
```diff
@@ -0,0 +1,249 @@
...
+	}
+	program = matched->Cast<ListParseResult>();
+}
+
+ConnectModeParser::~ConnectModeParser() = default;
+
+//! Walk a parse subtree and record the min and max char-offset of any leaf node with a set offset.
+//! Leaf nodes (IDENTIFIER, KEYWORD, STRING, NUMBER, OPERATOR) carry the char offset of their token;
+//! container nodes (LIST/REPEAT/OPTIONAL/CHOICE) recurse into their children.
+static void CollectLeafOffsets(ParseResult &node, optional_idx &min_off, optional_idx &max_off) {
```
Dtenwolde, line 66: I'm not sure I understand why we need these offsets? To figure out how far we can possibly get in the query tokens?

src/parser/connect_mode/connect_mode_parser.cpp
```diff
@@ -0,0 +1,249 @@
...
+	if (!max_off.IsValid() || off > max_off.GetIndex()) {
+		max_off = optional_idx(off);
+	}
+}
+
+//! Slice the original SQL string for the source range covered by `node`. Returns empty string if
+//! the subtree contains no offset-bearing leaves. The end of the range is computed by looking up
+//! the last leaf's token in `tokens` and adding its length.
+//!
+//! When `include_trailing_semicolon` is true, the range is extended past any intermediate whitespace
```
Dtenwolde, line 108: Why would we not always include the semicolon in the query we send to the remote? Is it also intended to be parsed by remote engines that don't terminate on `;`? If so, I feel like we can deal with such a problem when we get there.

src/parser/connect_mode/connect_mode_parser.cpp
```diff
@@ -0,0 +1,249 @@
...
+	// The CHOICE wrapper exposes its resolved alternative via GetResult().
+	auto &resolved = inner.Cast<ChoiceParseResult>().GetResult();
+	const auto &name = resolved.name;
+	if (name == "ConnectChunk" || name == "DisconnectChunk") {
+		out.type = ConnectModeChunk::Type::CONTROL;
+	} else if (name == "ExecuteChunk") {
+		out.type = ConnectModeChunk::Type::EXECUTE;
+		// ExecuteChunk LIST shape: [KEYWORD CONNECT, IDENTIFIER target, KEYWORD EXECUTE, LIST(Raw)]
+		auto &exec_list = resolved.Cast<ListParseResult>();
+		auto &target_node = exec_list.GetChild(1);
```
Dtenwolde, line 161: I would much rather see this go through the already existing Transform structure than this manual extraction process. I'd have to think about how the grammar should look int his case, but this feels very manual and error prone. If we change a rule name or add something this breaks. That's precisely the thing we've been trying to avoid with the current transformer generation

src/parser/connect_mode/connect_mode_parser.cpp
```diff
@@ -0,0 +1,249 @@
...
+		auto &payload_node = exec_list.GetChild(3);
+		out.target = target_node.Cast<IdentifierParseResult>().identifier.GetIdentifierName();
+		out.payload = SliceSubtreeText(sql, tokens, payload_node);
+	} else if (name == "ForbiddenConnect1") {
+		out.type = ConnectModeChunk::Type::FORBIDDEN;
+		out.forbidden_reason = ConnectModeChunk::ForbiddenReason::EXTRA_TOKENS_AFTER_CONNECT_TARGET;
+	} else if (name == "ForbiddenConnect2") {
+		out.type = ConnectModeChunk::Type::FORBIDDEN;
+		out.forbidden_reason = ConnectModeChunk::ForbiddenReason::INVALID_CONNECT_TARGET;
+	} else if (name == "ForbiddenConnect3") {
```
Dtenwolde, line 171: Point in case here with these "ForbiddenConnect2", "3", etc.

src/parser/connect_mode/connect_mode_parser.cpp
Dtenwolde: I think it'd be better if we could somehow return a `PartialParseResult` that parses up to a certain break point (We wouldn't be the first parser to do such a thing), and from there determine what needs to be done with the remainder of the query.

Dtenwolde (commented):
I think we need to consider allowing the return of `PartialParseResult` that can parse until a given break point, after which the remainder of the query can be parsed in a different way. The current approach feels brittle and error-prone

carlopi:
Thanks for the review and extensive comments. I'll have a proper look, thanks for the feedback.

carlopi:
@Dtenwolde, but maybe we can also discuss on a side, there are some bits that I think looks more solid (like first 2 commits), and some that indeed require more work.

Dtenwolde:
@carlopi What's the status with this PR?

PR #23662, opened 2026-07-07
[Async I/O] Read Ahead Interface + Parquet

pdet:
This PR introduces a read-ahead interface for the multifile reader. Work is split into jobs, which is the unit one scan thread would process. For example, for parquet, we are talking about whole row group. Jobs are then added into an ordered queue ahead of decoding. 

When a job enters the queue its I/O is immediately scheduled on the async I/O pool. Scanning threads act as both producers and consumers of the queue. Each acquire also tops up the queue until the queue is full, then claims the oldest job (batch-index order, preserving scan order) and blocks until that job's I/O completes.

Behavior is controlled by the `read_ahead_depth`.
1. 0 disables read-ahead
2. A positive number caps the jobs scheduled ahead
3. -1 (the default) instead bounds the bytes of I/O scheduled with a memory budget

The memory budget is currently `memory_limit/4`, and each job is floored with 16MB to avoid tiny jobs not being accounted for. Currently, this budget is hardcoded, but may become configurable later.

 In all modes, when buffer-manager usage exceeds half the memory limit, read-ahead backs off to a single job ahead so prefetching cannot push operators into out-of-memory. 

Read-ahead is only functional for readers that support it (parquet, currently) and when async threads are available. If that's not the case  the existing per-thread scan path is unchanged.

One important caveat related to the parquet filter prefetch strategy is that the filter prefetching is read ahead, but the remaining columns are scheduled on demand during decode, after the filters have been evaluated, using the regular async I/O path.

## Benchmark
To assess the read-ahead capacity we got a AWS EC2 `r7i.16xlarge` machine, and used a partitioned TPC-DS SF 100, snappy parquet files on S3. In total we have about 12,000 files, totaling 40GB of data.

To ensure we are properly benchmarking the ASYNC I/O, we use the following setup for the connections:

```sql
SET httpfs_connection_caching = true;    -- pool/reuse S3 connections, avoid a fresh TCP+TLS handshake
SET enable_external_file_cache = false;  -- every query re-reads from S3 
SET parquet_metadata_cache = false; -- every query re-reads from S3 
```
We do a full run of all 99 queries from DuckDB's `tpcds_queries()` in numeric order, timed with DuckDB's
`.timer` "real".

| Version    | Time    | Improvement vs. Baseline | Extra Configuration |
|------------|---------|--------------------------|---------------------|
| 1.5.4      | 596.9 s | 1.0x                     | -                   |
| 1.5.4      | 379.6 s | 1.8x                     | `SET THREADS = 128` |
| read_ahead | 349.9 s | 3.0x                     | `SET READ_AHEAD_DEPTH = 0`  |
| read_ahead | 196.3 s | 3.0x                     | -                   |


src/common/multi_file/multi_file_read_ahead.cpp
```diff
@@ -0,0 +1,246 @@
...
+}
+
+unique_ptr<MultiFileReadAhead> MultiFileReadAhead::Create(ClientContext &context) {
+	const auto configured_depth = Settings::Get<ReadAheadDepthSetting>(context);
+	if (configured_depth == 0) {
+		return nullptr;
+	}
+	if (configured_depth == -1) {
+		// automatic mode, unlimited depth, the I/O byte budget is the sole gate
+		const auto io_byte_budget = BufferManager::GetBufferManager(context).GetMaxMemory() / 4;
```
lnkuiper: Why is automatic read-ahead capped at memory_limit / 4 here? This is competing with other temp-memory consumers, so I think this should use the TemporaryMemoryManager/ManagedAsyncMemoryGovernor pattern used by AsyncWriteQueue instead

src/common/multi_file/multi_file_read_ahead.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+unique_ptr<MultiFileScanJob> MultiFileReadAhead::ClaimJob() {
+	lock_guard<mutex> guard(lock);
+	if (ready_queue.empty()) {
+		return nullptr;
+	}
+	auto job = std::move(ready_queue.front());
+	ready_queue.pop_front();
+	ReleaseSlot();
+	pending_io_bytes -= job->io_bytes;
```
lnkuiper: This releases the byte budget when the job is claimed, but the job’s scheduled I/O may still be running while the scan task is parked. That means claimed-but-not-completed jobs no longer count against the cap, so the actual in-flight read-ahead can exceed the budget by up to the active scan tasks. The accounting should probably be released after WaitForJob()/I/O completion instead.

src/parallel/executor.cpp
```diff
@@ -431,12 +431,15 @@ void Executor::InitializeInternal(PhysicalOperator &plan) {
 
 void Executor::CancelTasks() {
 	task.reset();
+	unordered_map<Task *, shared_ptr<Task>> to_destroy;
```
lnkuiper: Nit: reference_map_t<Task, shared_ptr<Task>> instead of raw ptr

lnkuiper (commented):
Looks great! Really exciting results. I have a few comments:

pdet:
Thanks for the review @lnkuiper, fixed your points (and some CI shenanigans), can you have another look?

edit: Actually, I'll need a bit more time, I've did some further benchmarking and it seems we are getting tons of connection errors(e.g., "Could not resolve hostname",  "Problem with the SSL CA cert") from stretching our connections.

lnkuiper:
Changes look great! Ready to merge when you feel ready

lnkuiper:
I think the httpfs patch is great because clients are checked out from and returned to the shared httpfs curl connection pool per request, instead of being hoarded inside each `HTTPFileHandle` until that handle is destroyed. This is actually something I wanted to do in the future.

However, for non-curl, i.e., httplib (which can't cache connections), or when `httpfs_connection_caching=false` this patch also stops using the old per-handle client reuse path, so those paths will create/destroy clients per request rather than keeping them on the file handle. This means we'll have an additional roundtrip for each request. I think that's OK for `httpfs_connection_caching=false` but not OK for httplib.

The clean implementation is to make the shared connection cache live at the `HTTPFSUtil` layer instead of being curl-only, then let both curl and httplib `InitializeClient`/`CloseClient` use it.

Maybe @carlopi can chime in?

carlopi (commented):
I was thinking, would not be possible to runtime select whether handle specific or generic?

Something like:
```c++
 unique_ptr<HTTPClient> HTTPFileHandle::GetClient() {
	// Try to fetch a cached client
        if (per_handle_cache) {
	    auto cached_client = client_cache.GetClient();
	    if (cached_client) {
		return cached_client;
	    }
	}
 	return CreateClient();
 }

 void HTTPFileHandle::StoreClient(unique_ptr<HTTPClient> client) {
	if (per_handle_cache) {client_cache.StoreClient(std::move(client)); }
	else {http_params.http_util.CloseClient(std::move(client)); }
 }
```
+ having a `per_handle_cache` on `HTTPFileHandle`, and have that intialize to true ONLY when curl AND connection caching is on?

I think the patch is solid when connection caching is enabled on `curl` backend, but this would allow to keep old behaving on `httplib` or connection caching disabled.

For the rest, I tried to look for potential deadlocks / lock serialization issues / other possible issues, but I think this is the only issue.

carlopi (approved):

carlopi:
Looks solid!

pdet:
Hmm, so doing further benchmarking here, we will still get the DNS/fd storm from the read head if we set `httpfs_connection_caching=false`, even with `read_ahead_depth = 0`, which results in queries failing.

pdet:
@carlopi should we flip the default of `httpfs_connection_caching`? I don't have the context of why it defaults to false to begin with. Maybe we can also share DNS + TLS sessions across curl clients, and try to do an early handle close in the read ahead. I'm open to suggestions


| conn caching | read-ahead      | completed | failed (storm) | real wrong | total    |
|--------------|-----------------|-----------|----------------|------------|----------|
| on           | auto (−1)       | 98/98     | 0              | 0          | 203 s |
| on           | off (depth=0)   | 98/98     | 0              | 0          | 240 s    |
| off          | auto (−1)       | 29/98     | 70             | 0          | collapsed |
| off          | off (depth=0)   | 74/98     | 25             | 0          | collapsed |

carlopi:
Yes, it should be flipped. On `main` for sure, possibly to be considered for `v1.5.5`.

Context is: this came very late, I think shortly before v1.5.0, and I was not really sure if it would be working, so shipped in v1.5.0 the old default (slow), and left as opt-in. There is even https://github.com/duckdb/duckdb-httpfs/pull/343 up on httpfs side (was failing due to some Windows crap, now ready to go in I think).

carlopi:
Actually: at this moment we expect users to just use `SET httpfs_connection_caching = true` for best performance, and it has been (and it will be even more if this becomes the default) properly tested. I think it's reasonable to just switch defaults (see also https://github.com/duckdb/duckdb-httpfs/pull/343) or remove a setting that is a footgun, let's simplify!

pdet:
Alright, I've added a patch to flip the default, we should remove it in a follow-up PR as this PR is already sizeable.

Thanks @carlopi and @lnkuiper for ze feedback!

PR #23665, opened 2026-07-07
Improve TPC-H dbgen progress reporting, ETA stability, and generation performance

kryonix:
This PR hooks `CALL dbgen(...)` into the progress bar and ETA infrastructure and makes execution substantially faster, especially for larger scale factors and persistent databases.

Previously, TPC-H generation could run for a long time without meaningful progress feedback. DuckDB already has CLI progress bar support, but the TPC-H dbgen path did not use it. This PR wires dbgen into the progress system and tracks both generated work and flushed work, so the progress bar keeps moving through the storage phase instead of jumping from partial progress to done.

----

While working on progress reporting, the ETA behavior also needed attention. The previous terminal ETA used the Kalman estimator, which works better for relatively smooth progress signals. TPC-H generation is burstier: parallel workers finish chunks at uneven times, persistent storage introduces flush phases, and progress can arrive in larger steps. That made the ETA overreact to fresh samples and jump around.

This PR replaces the terminal ETA logic with a simpler estimator:
- use elapsed progress as the stable baseline,
- use recent observed progress rate as a correction,
- clamp the correction so short bursts cannot dominate the estimate,
- smooth the predicted completion timestamp,
- let later estimates move outward gradually instead of jumping upward.

The result is less clever, but more predictable: the ETA still adapts when the workload speed changes, but it no longer swings wildly on each progress update. However, I'm not attached to this change, we can also just revert to what we had before.

----

The larger part of the PR rewrites the hot TPC-H dbgen append path. The old implementation generated DBGEN row structs first and then copied those values into DuckDB vectors. That added extra row materialization, extra string handling, and avoidable date formatting/parsing work.

The new path generates directly into DuckDB chunks. In particular, it:

- writes TPC-H date offsets directly as DuckDB dates,
- avoids intermediate `order_t`, `line_t`, `part_t`, etc. append structs,
- avoids repeated distribution string length calculation,
- references stable DBGEN distribution/text-pool strings directly where safe,
- keeps writable vector-owned string storage only for values that DBGEN may mutate,
- tunes parallel generation granularity,
- flushes persistent append state earlier to keep memory use bounded.

The end result is a large runtime improvement:

| Mode       |          Query |   Before  |  After  | Speedup |
|------------|----------------|-----------|---------|---------|
|  In-memory |  dbgen(sf=1)   |  1.249s   |  0.259s |    4.8x |
|  In-memory | dbgen(sf=10)   | 12.818s   |  1.274s |   10.1x |
| Persistent | dbgen(sf=10)   | 46.752s   |  4.159s |   11.2x |

SF 1000 with a memory limit of 20GB now runs in about 470 seconds. I was not willing to wait for the old implementation to finish, but it would have been much longer.

The main practical win of the lower-level rewrite: less intermediate work, fewer copies, less string/date conversion overhead, and much less pressure on the storage append path.

### DIFF
--- a/extension/tpch/dbgen/dbgen.cpp
+++ b/extension/tpch/dbgen/dbgen.cpp
@@ -16,6 +16,7 @@
 #include "duckdb/planner/binder.hpp"
 #include "duckdb/storage/data_table.hpp"
 #include "duckdb/storage/optimistic_data_writer.hpp"
+#include "duckdb/storage/storage_info.hpp"
 #include "duckdb/storage/table/append_state.hpp"
 #include "duckdb/transaction/duck_transaction.hpp"
 
@@ -666,7 +667,7 @@ struct TPCHDBgenParameters {
 
 static constexpr int DBGEN_TABLES[] = {SUPP, CUST, ORDER_LINE, PART_PSUPP, NATION, REGION};
 static constexpr idx_t DBGEN_ROW_BATCH_SIZE = 100000;
-static constexpr idx_t DBGEN_TARGET_CHUNK_ROWS = 122880;
+static constexpr idx_t DBGEN_TARGET_CHUNK_ROWS = 2 * DEFAULT_ROW_GROUP_SIZE;
 
 struct DBGenTableRange {
 	idx_t offset;
--- a/extension/tpch/dbgen/dbgen.cpp
+++ b/extension/tpch/dbgen/dbgen.cpp
@@ -42,14 +42,21 @@ enum class TPCHAppendMode : uint8_t { APPENDER, OPTIMISTIC };
 struct tpch_append_information {
 	duckdb::unique_ptr<InternalAppender> appender;
 	unique_ptr<OptimisticDataWriter> optimistic_writer;
-	unique_ptr<OptimisticWriteCollection> optimistic_collection;
+	optional_ptr<OptimisticWriteCollection> optimistic_collection;
 	optional_ptr<DuckTableEntry> table_entry;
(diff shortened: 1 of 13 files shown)


extension/tpch/dbgen/dbgen.cpp
```diff
@@ -8,287 +8,627 @@
...
+
+	void PrepareOptimisticWriteToDisk() {
+		D_ASSERT(table_entry);
+		D_ASSERT(optimistic_writer);
+		D_ASSERT(optimistic_collection);
+		FinalizeOptimisticAppend();
+		auto &storage = table_entry->GetStorage();
+		auto &row_collection = *optimistic_collection->collection;
+		if (row_collection.GetTotalRows() >= storage.GetRowGroupSize()) {
+			optimistic_writer->WriteUnflushedRowGroups(*optimistic_collection);
```
lnkuiper, line 178: This can write persistent row-group blocks while the optimistic collection is still only owned by the worker appender. If another worker errors or the query is interrupted before this appender is moved into `finished_appenders` and later merged through `FlushOptimistic()`, unwinding destroys the appender without registering the collection in `LocalStorage`, so the normal rollback path cannot drop those blocks. Can we either transfer/register ownership before prewriting, or add RAII cleanup/rollback for unmerged optimistic collections/writers?

lnkuiper (commented):
Looks great! So much improvement. CI seems happy which is the most important part, but Codex has found one issue (see below) feel free to mark "Ready To Merge" once it's addressed

PR #23666, opened 2026-07-07
Add `make parser-grammar` and check for diff in CI

Dtenwolde:
This PR adds `parser-grammar` as a make command and hooks it into the generated-file path. 

`make generate-files` now also calls `make parser-grammar` and regenerates the inlined grammar files, heap-based transformer wrapper and for now the stack based transformer wrapper. 

Also if there are changes to the grammar files but `make parser-grammar` was not used, it shows a diff error in the CI, similar to `make generate-files` or `make format-fix`


Dtenwolde:
How ironic that CI fails on the prepare step 😂

PR #23667, opened 2026-07-07
Fix ROW_NUMBER() OVER () over read_duckdb returning file-local row numbers

### LINKED ISSUE #23586: ROW_NUMBER over read_duckdb can return 0-based file-local row numbers
### What happens?

`ROW_NUMBER() OVER ()` over `read_duckdb(...)` can return the scan virtual `row_number` column instead of SQL window row numbers. The repro creates a DuckDB file containing table `t` with two rows, then reads that file through `read_duckdb`. The window row numbers should be `1, 2`, but DuckDB returns `0, 1`.

This reproduces on current main (ac62d7c180), but not on v1.5.4.

### To Reproduce

To reproduce:

```sql
ATTACH '/private/tmp/sqlcaster_read_duckdb_rn_selfcontained.duckdb' AS rn_single;
CREATE OR REPLACE TABLE rn_single.t(x INTEGER);
INSERT INTO rn_single.t VALUES (10), (20);
DETACH rn_single;

SELECT ROW_NUMBER() OVER () AS rn
FROM read_duckdb('/private/tmp/sqlcaster_read_duckdb_rn_selfcontained.duckdb', table_name='t')
ORDER BY rn;
```

Run:

```sh
```

Actual:

On current main (`ac62d7c180`), running the minimized repro gives:

stdout:

| rn |
| ---: |
| 0|
| 1 |

Expected:

| rn |
| ---: |
| 1 |
| 2 |

Notes From Minimization:

- `SET disabled_optimizers='window_rewriter'` makes the query return `1, 2`.
- The table values are not important; they only provide two input rows for `ROW_NUMBER()`.

---
Possible Cause (generated by coding agent):
Source pointers: `src/optimizer/row_number_rewriter.cpp`, `src/function/table/read_duckdb.cpp`, `src/common/constants.cpp`, and `src/storage/table/row_number_column_data.cpp`.

The `RowNumberRewriter` removes a `LogicalWindow` for bare `ROW_NUMBER() OVER ()` when the child is a `LogicalGet` exposing `COLUMN_IDENTIFIER_ROW_NUMBER`. That is only safe when the scan's virtual row number has SQL window semantics: dense, visible-row-based, and 1-indexed for the current query. `read_duckdb` also exposes a `row_number` virtual column, but this repro shows it can behave like a scan/file-local position instead of a computed SQL window row number. A likely fix is to make table functions opt in with a stronger capability flag for SQL-compatible row numbers, or to explicitly exclude `read_duckdb` from this rewrite until its virtual column semantics match `ROW_NUMBER() OVER ()`.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (ac62d7c180)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

vismaytiwari:
Fixes #23586.

The `RowNumberRewriter` rewrites a bare `ROW_NUMBER() OVER ()` by mapping it onto the child `LogicalGet`'s virtual `row_number` column. That column only has SQL window semantics — dense and 1-based over the visible rows of the scan — for **base-table scans** (backed by `RowNumberColumnData`).

Table functions such as `read_duckdb` also expose a `row_number` virtual column, but with file-local semantics. So `ROW_NUMBER() OVER ()` over `read_duckdb(...)` returned the scan's 0-based file-local numbers (`0, 1`) instead of `1, 2`. (`SET disabled_optimizers='window_rewriter'` avoided it, confirming the rewriter as the cause.)

This restricts the rewrite to base-table scans via `LogicalGet::GetTable()` (non-null only for a real catalog table); table-function scans keep the window operator and compute correct SQL row numbers.

Verified the base-table rewrite still fires (existing `test/optimizer/row_number_virtual_column.test` cases pass) and added a `read_duckdb` case asserting the correct `1, 2` result and that the window is retained. The `read_duckdb` and full `optimizer` test groups pass with no regressions.

The separate case of a `read_duckdb` scan that projects no real columns (e.g. bare `SELECT ROW_NUMBER() OVER ()`), which trips an assertion in assert-enabled builds, is tracked in #23670.

### DIFF
--- a/src/optimizer/row_number_rewriter.cpp
+++ b/src/optimizer/row_number_rewriter.cpp
@@ -37,6 +37,15 @@ bool RowNumberRewriter::CanOptimize(LogicalOperator &op) {
 		return false;
 	}
 
+	// only rewrite for a base-table scan: its virtual row_number column has SQL
+	// window semantics (dense, over the visible rows of this scan). Table
+	// functions such as read_duckdb also expose a row_number virtual column, but
+	// with different (e.g. file-local) semantics, so mapping ROW_NUMBER() OVER ()
+	// onto it would return the wrong values (issue #23586).
+	if (!get.GetTable()) {
+		return false;
+	}
+
 	// cannot rewrite if there are table filters pushed into the scan
 	// the virtual row_number column counts absolute row positions in storage,
 	// which does not match row_number() over() when rows are filtered out
--- a/test/optimizer/row_number_virtual_column.test
+++ b/test/optimizer/row_number_virtual_column.test
@@ -218,3 +218,28 @@ query II
 EXPLAIN SELECT row_number + 1 FROM (SELECT row_number() OVER () AS row_number FROM tbl);
 ----
 physical_plan	<!REGEX>:.*WINDOW.*
+
+# read_duckdb also exposes a row_number virtual column, but with file-local
+# semantics rather than SQL window semantics, so the rewrite must not fire for
+# it: ROW_NUMBER() OVER () must still be computed and return 1-based values, not
+# the scan's 0-based file-local row number. See issue #23586.
+statement ok
+ATTACH '{TEST_DIR}/row_number_read_duckdb.db' AS rd
+
+statement ok
+CREATE TABLE rd.t AS SELECT * FROM (VALUES (10), (20)) v(x)
+
+statement ok
+DETACH rd
+
+query II
+SELECT x, ROW_NUMBER() OVER () AS rn FROM read_duckdb('{TEST_DIR}/row_number_read_duckdb.db', table_name='t') ORDER BY x
+----
+10	1
+20	2
+
+# the window must be kept (not eliminated) for a table function scan
+query II
+EXPLAIN SELECT x, ROW_NUMBER() OVER () FROM read_duckdb('{TEST_DIR}/row_number_read_duckdb.db', table_name='t')
+----
+physical_plan	<REGEX>:.*WINDOW.*


vismaytiwari:
One note on scope. This fixes the wrong result — `SELECT ROW_NUMBER() OVER () FROM read_duckdb(...)` now returns `1, 2` instead of the scan's 0-based file-local `row_number`.

The regression test deliberately selects a real column (`SELECT x, ROW_NUMBER() OVER ()`). The bare `SELECT ROW_NUMBER() OVER () FROM read_duckdb(...)` (no other column) surfaces a separate, pre-existing limitation: with the window now correctly retained, the scan projects zero real columns, and `read_duckdb` — unlike `read_parquet` — exposes no empty/constant virtual column, so it hits `!column_ids.empty()` in assert builds. On `main` this was masked because the buggy rewrite added the `row_number` column.

I tried giving `read_duckdb` the `COLUMN_IDENTIFIER_EMPTY` support the base `MultiFileReader` already provides, but its scan loop doesn't advance the row count when only a constant column is projected, so it hangs — that looks like it needs a change in `read_duckdb`'s scan path best made by someone who knows it. Happy to open a separate issue for the zero-column `read_duckdb` case, or fold it in here if you can point me the right way. This PR stays scoped to the optimizer fix.

hawkfish (approved):
LGTM!  And yes, please file a separate issue.

vismaytiwari:
Thanks! Filed #23670 for the zero-column `read_duckdb` scan case.

vismaytiwari:
The `Windows (64 Bit)` failure looks like an infra flake, not the change — the Build step died on an Azure `BlobNotFound` (cache/blob fetch), and the log has no compile or test errors. Everything else is green, including `Linux Relassert Tests`. A re-run of that one job should clear it.

PR #23668, opened 2026-07-07
Support partial MIN/MAX precompute when some partitions lack exact stats

The-Alchemist:
## Summary
Extend partial MIN/MAX precompute to bare aggregates when some partitions (row groups) lack exact min/max stats. Partitions with usable stats are folded from `get_partition_stats`; the rest are scanned via `set_partitions_to_scan` and merged with `COALESCE`/`least`/`greatest`.

Previously this path only triggered with table filters. Native tables hit it when a column is all-null in some row groups; table functions with per-file partition stats (e.g. Iceberg) are another consumer ([PR for `duckdb-iceberg` ](https://github.com/The-Alchemist/duckdb-iceberg/pull/5)will be ready once this is merged).

## Tests
- Extended `test/optimizer/partial_aggregate_precomputation.test` — bare aggregate partition split (`Row Groups Scanned: 2 / 4`, no `WHERE`)

## Test plan
- [x] `test/optimizer/partial_aggregate_precomputation.test`

## Feedback welcome!

Do we need more tests?

🙇‍♂️

### DIFF
--- a/src/optimizer/statistics/operator/propagate_aggregate.cpp
+++ b/src/optimizer/statistics/operator/propagate_aggregate.cpp
@@ -222,7 +222,8 @@ void StatisticsPropagator::TryExecuteAggregates(LogicalAggregate &aggr, unique_p
 	vector<unique_ptr<Expression>> agg_results;
 	bool need_to_scan = false;
 	vector<idx_t> scan_partition_indices;
-	// we can keep execute eager aggregate if all partitions could be either filtered entirely or remained entirely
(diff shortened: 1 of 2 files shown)


Damon07:
It seems like the same with #23611.

The-Alchemist:
> It seems like the same with https://github.com/duckdb/duckdb/pull/23611.

Oh, ok, i'll close this one out yet.  Looking forward to using in duckdb-iceberg!

PR #23669, opened 2026-07-07
Honor the comment option when quoting CSV output

simi:
The CSV writer unconditionally quoted fields containing '#'. Per my understanding, this was meant to keep '#' values from being read back as comments (the sniffer treats '#' as a default comment candidate), but it was applied regardless of the actual comment option: '#' was quoted even for `comment ''`, `comment '!'` quoted '#' while leaving '!' unquoted, and the output diverged from RFC 4180 minimal quoting.

Thread the comment character into CSVWriterOptions and quote it only when set, defaulting to '#' when the user specifies no comment for backwards compatibility. Default output is unchanged; `comment ''` now produces RFC 4180 minimal quoting and an explicit comment character is the one that gets quoted.

### DIFF
--- a/src/common/csv_writer.cpp
+++ b/src/common/csv_writer.cpp
@@ -30,17 +30,28 @@ CSVWriterState::~CSVWriterState() {
 	}
 }
 
-CSVWriterOptions::CSVWriterOptions(const string &delim, const char &quote, const string &write_newline) {
+CSVWriterOptions::CSVWriterOptions(const string &delim, const char &quote, const string &write_newline,
+                                   char comment_char) {
 	requires_quotes = vector<bool>(256, false);
 	requires_quotes['\n'] = true;
 	requires_quotes['\r'] = true;
-	requires_quotes['#'] = true;
+	// quote values containing the comment char so they are not read back as comments (see #17744)
+	if (comment_char != '\0') {
+		requires_quotes[NumericCast<idx_t>(comment_char)] = true;
+	}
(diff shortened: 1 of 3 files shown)


src/common/csv_writer.cpp
```diff
@@ -63,9 +74,9 @@ CSVWriter::CSVWriter(WriteStream &stream, vector<string> name_list, bool shared)
 
 CSVWriter::CSVWriter(CSVReaderOptions &options_p, FileSystem &fs, const string &file_path,
                      FileCompressionType compression, QueryContext context, bool shared)
-    : options(options_p),
-      writer_options(options.dialect_options.state_machine_options.delimiter.GetValue(),
-                     options.dialect_options.state_machine_options.quote.GetValue(), options.write_newline),
+    : options(options_p), writer_options(options.dialect_options.state_machine_options.delimiter.GetValue(),
```
simi, line 77: 🤔 seems this could be also shorted to, I did the explicit version to show the diff

```
: options(options_p), writer_options(options),
```

simi:
Anything I can help with to make review simpler/easier/smooter @pdet?

pdet:
Hi @simi, sorry for the belated review, LGTM! Ill just start CI.

simi:
@pdet CI looks ok :pray:

simi:
@pdet friendly ping :pray:

pdet:
@simi thanks!

PR #23671, opened 2026-07-07
Check staged extensions in CI

smvv:
### Problem with current alpha version

The issue is that staged alpha CLIs have a version tag like `v2.0.0-alpha34176`, but the matching extensions were deployed under the source SHA path `dfb621a3f8`.

In CI, the command `./build/release/duckdb -c "PRAGMA version;"` [shows](https://github.com/duckdb/duckdb/actions/runs/28859776317/job/85595245370):
```
┌───────────────────┬────────────┬─────────────────────┐
│  library_version  │ source_id  │      codename       │
│      varchar      │  varchar   │       varchar       │
├───────────────────┼────────────┼─────────────────────┤
│ v2.0.0-alpha34176 │ dfb621a3f8 │ Development Version │
└───────────────────┴────────────┴─────────────────────┘
```

That creates two failures:

1. `INSTALL ducklake;` looks for:
   `http://extensions.duckdb.org/v2.0.0-alpha34176/osx_arm64/ducklake.duckdb_extension.gz`

   But the deployed extension exists at:
   `http://extensions.duckdb.org/dfb621a3f8/osx_arm64/ducklake.duckdb_extension.gz`

2. Manually installing the SHA-built extension fails because DuckDB sees a version mismatch:
   extension built for `dfb621a3f8`, CLI reports `v2.0.0-alpha34176`.

The likely fix is to make DuckDB treat alpha builds consistently with dev builds for extension compatibility and repository lookup.

I refactored `DuckDB::ReleaseCodename()` and `ExtensionHelper::IsRelease(const string &version_tag)` so `-alpha` versions behave like `-dev`: use the source ID path for extensions, and validate extensions against the source ID rather than the alpha version string.

The goal is: an alpha CLI should successfully install and load extensions from the staged/source-ID extension repository.

### Changes

- Add `IsReleaseVersion()` to check if it's a release version based on duckdb version suffix.
- Verify staged linux extensions in CI by installing and loading the extensions.

### DIFF
--- a/.github/workflows/Main.yml
+++ b/.github/workflows/Main.yml
@@ -743,7 +743,8 @@ jobs:
       run: |
         python3 scripts/ci/check_staged_extensions.py \
           --git-sha "${{ needs.prepare.outputs.git_sha }}" \
-          --version "${{ needs.prepare.outputs.effective_git_describe }}"
+          --version "${{ needs.prepare.outputs.effective_git_describe }}" \
+          --extensions "httpfs,json,parquet,ducklake,iceberg,icu"
 
  linux-musl-release-cli:
     name: Linux CLI (${{ matrix.config.arch }}-musl)
--- a/.github/workflows/Main.yml
+++ b/.github/workflows/Main.yml
@@ -724,6 +724,27 @@ jobs:
         build/release/benchmark/benchmark_runner benchmark/micro/update/update_with_join.benchmark
         build/release/duckdb -c "COPY (SELECT 42) TO '/dev/stdout' (FORMAT PARQUET)" | cat
 
+ staged-extension-install:
+    name: Staged Extension Install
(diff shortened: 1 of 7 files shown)


.github/workflows/Main.yml
```diff
@@ -724,6 +724,28 @@ jobs:
...
+      with:
+        ref: ${{ needs.prepare.outputs.git_sha }}
+
+    - name: Check staged extension install/load
+      timeout-minutes: 10
+      run: |
+        python3 scripts/ci/check_staged_extensions.py \
+          --git-sha "${{ needs.prepare.outputs.git_sha }}" \
+          --version "${{ needs.prepare.outputs.effective_git_describe }}" \
+          --extensions "httpfs,json,parquet,ducklake,iceberg,icu"
```
smvv, line 747: note: use fixed subset.

Originally, I used `SELECT * FROM duckdb_extensions()` and that lists hardcoded extension names that not necessarily exist in the staging bucket at that time.

PR #23673, opened 2026-07-07
fix hanging test runner

benfleis:
When stderr fills up with deprecation warnings (e.g. ${VAR} should be {VAR}), the test runner fails to read stderr, so the unittest process waits while to write to the (full pipe) stderr. In parallel the test runner ignores stderr, and eventually gets to wait()ing on the unittest process itself to end... which will never happen because it's blocked on the pipe write. Deadlock.

Fix by writing stdout/stderr to temp files instead.

Discovered in Windows on Delta CI runs that were mysteriously timing out.


smvv (approved):
nice find, thank you!

PR #23690, opened 2026-07-08
Task executor use notification to block wait completion

dentiny:
Hi team, when I was debugging a httpfs extension issue, I surprisingly found accessing remote files leads to 100% CPU utilization. The root cause of which is we [busy loop](https://github.com/duckdb/duckdb/blob/a1aacc73df5da3c856da77163da6073dbf24a14c/src/parallel/task_executor.cpp#L50-L59) to wait for tasks completion....

This PR updates to leverage notification-based; in detail, the waiter will be notified under any of the following situation to check whether the desired condition has been satisfied (namely, all enqueued tasks are completed):
- An already enqueued task completes
- A task somehow fails to get scheduled
- A task gets re-enqueued


lnkuiper:
Thanks, great idea!

PR #20222, opened 2025-12-16
return_stats for boolean and numeric (128 bits) during copy

aykut-bozkurt:
This PR lets `COPY .. WITH (return_stats)` output statistics for types boolean and numeric(128 bits).

PR #23693, opened 2026-07-08
Fix swapped min/max for multi-row-group 128-bit DECIMAL in RETURN_STATS

sfc-gh-okalaci:
## Summary

`COPY ... (FORMAT PARQUET, RETURN_STATS)` can return an **incorrect (swapped) min/max** for a `DECIMAL` with precision > 18 (128-bit, `hugeint_t`) when the column spans more than one row group.

Example — the data ranges from 19308 to 126619, but the returned stats swap them:

```sql
CREATE TABLE t AS
  SELECT CASE WHEN i < 2048 THEN 126619 ELSE 19308 END::DECIMAL(38,0) AS c
  FROM range(4096) t(i);

COPY t TO 't.parquet' (FORMAT PARQUET, ROW_GROUP_SIZE 2048, RETURN_STATS);
-- column_statistics reports  min=126619, max=19308   (wrong; should be min=19308, max=126619)
```

The Parquet footer itself is written correctly; only the aggregated stats returned by `RETURN_STATS` are wrong.

## Root cause

`DecimalStatsUnifier<T>` (in `extension/parquet/parquet_writer.cpp`) unifies per-row-group min/max into file-level min/max. For 128-bit decimals it inherited `BaseNumericStatsUnifier::UnifyMinMax`, which compares the raw statistic bytes with a native `Load<hugeint_t>` (little-endian).

But a `DECIMAL` with precision > 18 is stored as a Parquet `FIXED_LEN_BYTE_ARRAY`, and its min/max statistics are the unscaled value encoded as **big-endian two's complement**. Loading those bytes little-endian compares byte-swapped values, so the cross-row-group comparison can pick the wrong extreme and even swap the file-level min and max (a high byte with the sign bit set flips positive values to negative).

Decimals with precision <= 18 use an `INT32`/`INT64` physical type whose statistics are genuinely little-endian, so they are unaffected.

## Fix

Override `UnifyMinMax` (and align `StatsToString`) in `DecimalStatsUnifier` to decode the big-endian bytes via `ParquetDecimalUtils::ReadDecimalValue<hugeint_t>` before comparing. The override is scoped to 128-bit with `if constexpr (sizeof(T) == sizeof(hugeint_t))`, so `INT32`/`INT64` decimals keep the little-endian base path (and the dead branch is discarded at compile time).

As a side benefit, decoding via `ReadDecimalValue` sign-extends short buffers, so minimal-width FLBA stats emitted by other writers (e.g. a `DECIMAL(20,0)` min/max is 9 bytes, not 16) no longer trip the base unifier's `InternalException("Incorrect size for stats in UnifyMinMax")` size check.

## Tests

`test/sql/copy/return_stats.test` gains a `decimal128_multi_rg` case: a two-row-group `DECIMAL(38,0)` where the row-group maxima differ, asserting `RETURN_STATS` reports `min=19308, max=126619`. It fails before the change and passes after.

## Context

This is a follow-up bugfix on top of #20222 (duckdb/duckdb@174ec83, "return_stats for boolean and numeric types (128 bits)"), which added 128-bit (`hugeint_t`) support to `DecimalStatsUnifier`. That change decoded the big-endian bytes in `StatsToString`, but left `UnifyMinMax` on the inherited little-endian comparison, so the cross-row-group min/max could still be swapped. This PR completes that work.

cc @lnkuiper @pdet @aykut-bozkurt (from #20222) — flagging you since this fixes a bug in that change.


lnkuiper:
Thanks! I've rebased to `v1.5-variegata`, because the issue exists there too. Ready to go once CI passes.

sfc-gh-okalaci:
Thanks for adapting this to v1.5, @lnkuiper. Since the same bug is on `main` too — where `RETURN_STATS` still emits the extra `row_group_count` column — should I open a separate `main`-targeted PR (with the 7-column test variant), or will it be covered by a merge-forward from v1.5? Happy to do the former if that's easier for you.

lnkuiper:
This branch will be merged into `main` eventually, that's why I retargeted v1.5

PR #23696, opened 2026-07-08
Improve TPC-DS dsdgen progress reporting and generation performance

kryonix:
This PR gives the TPC-DS data generator the same treatment as the TPC-H generator: progress reporting, stable async execution behavior, direct vectorized appends, parallel generation, and row-group-aware writes for persistent databases.

Previously, `CALL dsdgen(...)` was mostly a black box: the CLI did not get progress updates, and generation was effectively bottlenecked by a single serial append path. That became especially painful for larger persistent TPC-DS databases, where the generator spent a lot of wall-clock time feeding rows into storage.

The new path keeps the existing TPC-DS data semantics, but changes how generated rows enter DuckDB. TPC-DS rows are now written directly into DuckDB DataChunks instead of going through a slower row-oriented appender path. For persistent databases, worker-local optimistic append collections can write row groups directly and then merge them into the target tables. This keeps the storage path much closer to bulk loading, while still preserving deterministic output order. The generator is also now parallelized. Larger tables are split into deterministic work items, generated by worker-local appenders, and merged in table/row-order afterward. The dsdgen state that used to be global has been made thread-local where needed, and the tests compare sequential and parallel output to make sure the result stays equivalent.

Progress reporting now works for TPC-DS as well. The progress estimate is weighted by generated table work rather than just table count, so larger fact tables dominate the reported progress instead of causing large jumps. On my machine, for persistent database generation with a 20GB memory limit, this results in significant speedup[1]:

|  Scale factor |    Before |    After |  Speedup |
|---------------|-----------|----------|----------|
|  SF 1         |    8.648s |   1.549s |     5.6x |
|  SF 10        |   74.156s |   7.629s |     9.7x |
|  SF 100       |  628.860s |  50.666s |    12.4x |

The larger scale factors benefit most because they have enough work to keep the parallel generation and row-group write path busy.

[1]: These numbers are only measured once per scale factor, so they are not statistically significant. They are meant to illustrate the performance difference between the two implementations.


lnkuiper:
Looks great, thanks!

PR #23705, opened 2026-07-08
Packrat parsing stack overflow fix

Dtenwolde:
During testing for the heap-based transformer I found a pretty significant regression for the `Matcher` after https://github.com/duckdb/duckdb/pull/23646. In this case we added `ParserPackratCache::Match` into the call-chain:
```
  MatchParseResult
    -> ParserPackratCache::Match
      -> MatchParseResultInternal
```
This happened, even for rules which would not be memoized, which are most of them. Due to the deep recursion in the grammar, this could quite easily trigger a stack overflow (Heap-based parser when? :))

Now, only rules that can be memoized go through this call-chain, while all other rules go through the old chain: 
```
 MatchParseResult
    -> MatchParseResultInternal
```

cc @Tmonster


Dtenwolde:
I think this PR should be seen as an improvement, but it seems like the new parser still has trouble with deeply nested recursive expressions. Perhaps making it heap-based to avoid stack overflow problems is worthwhile considering

Mytherin:
Thanks! Indeed this would be better to fix more structurally but for now let's get this in.

PR #23706, opened 2026-07-08
Fix/lambda interrupt check

krleonid:
## Problem

`list_transform`, `list_filter`, and `list_reduce` did not respect query interruption (from `max_execution_time` or a cancelled client context) while their inner loops were running.

- **`list_transform` / `list_filter`** share `ExecuteLambda` in `lambda_functions.cpp`. This function processes list elements in batches of `STANDARD_VECTOR_SIZE` (2048). The batch loop had no `InterruptCheck()` calls, so a pending interrupt was never observed until the entire loop finished.
- **`list_reduce`** has a separate fold-left `while (!end)` loop in `list_reduce.cpp` that runs once per list element. It had the same problem.

**Symptoms:** on a 1 M-element list, a 50 ms `max_execution_time` was consistently ignored and the query ran for ~700 ms before returning.

## Fix

Call `state.GetContext().InterruptCheck()` at the top of each processing cycle in both loops. `InterruptCheck()` throws `InterruptException` as soon as `interrupt_state == INTERRUPTED` (set by `duckdb_interrupt()` or when the `query_deadline` is exceeded), allowing the query to abort promptly.

```cpp
// lambda_functions.cpp — ExecuteLambda batch loop
if (elem_cnt == STANDARD_VECTOR_SIZE) {
+   state.GetContext().InterruptCheck();
    execute_info.lambda_chunk.Reset();
    ...
}

// list_reduce.cpp — fold-left loop
while (!end) {
+   state.GetContext().InterruptCheck();
    auto &result_chunk = ...;
    ...
}
```

## Test

`test/sql/function/list/lambdas/interrupt.test_slow` — runs all three functions with `SET max_execution_time=50` against a 1 M-element list and expects `Query exceeded maximum execution time`.

## Checklist

- [x] All three lambda functions interrupted promptly with `max_execution_time`
- [x] SQL regression test added
- [x] No functional change when no interrupt is pending

### DIFF
--- a/extension/core_functions/lambda_functions.cpp
+++ b/extension/core_functions/lambda_functions.cpp
@@ -314,6 +314,7 @@ static void ExecuteLambda(DataChunk &args, ExpressionState &state, Vector &resul
 		for (idx_t child_idx = 0; child_idx < list_entry.length; child_idx++) {
 			// reached STANDARD_VECTOR_SIZE elements
 			if (elem_cnt == STANDARD_VECTOR_SIZE) {
+				state.GetContext().InterruptCheck();
 				execute_info.lambda_chunk.Reset();
 				ExecuteExpression(elem_cnt, child_info, info.column_infos, index_vector, execute_info);
 				auto &lambda_vector = execute_info.lambda_chunk.data[0];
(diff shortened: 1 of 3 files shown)


taniabogatsch (approved):

taniabogatsch:
@krleonid - looks good to me, I think we can undraft.

lnkuiper:
Failure is unrelated (and should be fixed already on main), but I can only merge once everything is green

lnkuiper:
Thanks!

PR #23708, opened 2026-07-08
Implement TRY_CAST for time-related types

### LINKED ISSUE #23707: Time-related type throws exception on TRY_CAST
### What happens?

### To Reproduce

```sql
memory D SELECT TRY_CAST(TIMESTAMP '2262-04-11 23:47:16.854776' AS TIMESTAMP_NS);
Conversion Error:
Could not convert Timestamp to higher precision.
```

### OS:

macos

### DuckDB Version:

b3b5144b551743b36021ddf66d28f8f3f8214f61

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23707

I think currently TRY_CAST operation is not fully implemented for time-related types; this PR is intended to add all missing conversions.


test/sql/cast/timestamp_casts.test
```diff
@@ -30,3 +30,34 @@ query I
 select t::DATE from tss;
 ----
 2026-04-09
+
+query I
+SELECT TRY_CAST(TIMESTAMP '2262-04-11 23:47:16.854776' AS TIMESTAMP_NS);
```
pdet, line 35: Can we check we are not regressing error messages?

e.g.,

1. CAST(TIMESTAMP_MS 'infinity' AS TIME)   
2. CAST('infinity'::TIMESTAMP_NS AS TIME) 
3. CAST(TIMESTAMP_S 'infinity' AS TIME)
4. CAST('infinity'::TIMESTAMP_NS AS TIME_NS)

pdet: It also seems the truncation of the following changes with the PR:
`'1970-01-01 23:59:59.999999999'::DATE`
Maybe @hawkfish has a particular opinion

pdet (commented):
Thanks for the PR, leavign a few comments

test/sql/cast/timestamp_casts.test
dentiny: > Can we check we are not regressing error messages?

That makes sense!

The error message updates here
https://github.com/dentiny/duckdb/blob/1db2e7f8e8ea4e2a43e1ca6b5a2c179a19ad3f0d/src/function/cast/time_casts.cpp#L10-L22

PR #23713, opened 2026-07-08
Add HUGEINT test cases for gcd and lcm

hacoobamatata:
Adds test coverage for the HUGEINT overloads of `gcd` and `lcm`, including values outside the BIGINT range.

Tested by running `test/sql/function/numeric/test_gcd_lcm.test`.


Mytherin:
Please avoid opening PRs that only add tests

PR #23714, opened 2026-07-09
Fix DROP COLUMN corrupting per-column metadata block bookkeeping

ywelsch:
RowGroup::RemoveColumn copied per_column_metadata_blocks and erased the dropped column's entry, but left the entries of all subsequent columns keyed by their old (pre-drop) indices while the columns themselves shift down a position. The next partial-reuse checkpoint (force_column_metadata_reuse) then looked up reused columns by their new indices and attributed extra metadata blocks to the wrong columns. The wrong block list was persisted, still-referenced metadata blocks were freed and reused, and a later scan deserialized garbage, surfacing as "field id mismatch".

Fix: shift the entry indices of all columns after the dropped one down by one. PerColumnMetadataBlocks::RemoveColumn now does exactly that; the old shift-less behavior is renamed to ClearColumn and kept for the ALTER TYPE path, where the column is rewritten in place and positions don't shift.

The regression test fails without the fix at the first checkpoint after the drop with "per_column_metadata_blocks mismatch" (debug_verify_blocks tripwire); the small BLOCK_SIZE and varchar payloads make each column's metadata spill into extra blocks, which is a precondition for the bug.

### DIFF
--- a/src/include/duckdb/storage/table/per_column_metadata_blocks.hpp
+++ b/src/include/duckdb/storage/table/per_column_metadata_blocks.hpp
@@ -32,7 +32,11 @@ public:
 
 	//! Add a column entry with its block IDs
 	void AddColumn(idx_t col_idx, const vector<idx_t> &blocks);
-	//! Remove a column entry and all its block IDs (linear scan)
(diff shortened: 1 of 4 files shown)


Mytherin:
Thanks!

PR #22060, opened 2026-04-14
 Row group append

evertlammerts:
Related to:
- https://github.com/duckdblabs/appian/issues/84
- https://github.com/duckdblabs/appian/issues/81

This introduces an enum to control the ros group append mode, as [suggested](https://github.com/duckdblabs/appian/issues/81#issuecomment-4045743803) by @Mytherin.

cc @taniabogatsch @Dtenwolde

PR #23719, opened 2026-07-09
Keep non-updatable row-groups non-updatable after adding a new segment

### LINKED ISSUE #828: [1.5.x regression] ST_Union_Agg + INSERT corrupts transaction ("Skipping beyond end of binary data") and fails rollback
### What happens?

ST_Union_Agg followed by an INSERT … SELECT into a table with a GEOMETRY column (then CHECKPOINT) intermittently corrupts the
  transaction on a persistent on‑disk database, throwing:

  Invalid Input Error: Skipping beyond end of binary data at position 16

  The error then recurs during the automatic rollback, which escalates to a fatal, connection‑killing state:

  FATAL Error: Failed to rollback transaction. Cannot continue operation.

  It is non‑deterministic: running the exact same script on the exact same data fails at a different GROUP BY partition each time. There is ample free RAM and no disk spilling when it happens. This is a regression in 1.5.x — the identical script on the identical database file completes with no error on 1.4.4, but fails on 1.5.3 (same OS, hardware, data, client, and glibc malloc). Smaller per‑row geometries (e.g. writing the ~32M un‑aggregated input polygons) write fine; only the large aggregated multipolygons produced by
  ST_Union_Agg trigger it.

### To Reproduce

The failure occurs on a ~32 M‑row spatial dataset and is data‑dependent and intermittent, so I have not yet been able to distill a data‑independent minimal example (a synthetic generator producing ~6,900‑vertex unions did not trigger it). The exact, faithful reproducer:

```
  import duckdb

  # postcodes.duckdb (~7 GB) contains one table:
  #   voronoi_raw(postcode VARCHAR, region_code VARCHAR(2), voronoi_geom GEOMETRY)
  #   ~32.2M rows: a Voronoi tessellation of UK address points
  #   ~1.75M distinct postcodes spread across 11 region_code values
  con = duckdb.connect("postcodes.duckdb")
  con.execute("LOAD spatial;")
  con.execute("SET threads=4;")            # also fails at threads=23
  con.execute("SET memory_limit='31GB';")  # no disk spill occurs; RAM is not exhausted

  con.execute("""
      CREATE TABLE voronoi_unioned (
          postcode VARCHAR PRIMARY KEY,
          region_code VARCHAR(2),
          unioned_geom GEOMETRY,
          cell_count INTEGER,
          area_m2 DOUBLE
      )
  """)

  # Union the cells of each postcode into one geometry, one region per INSERT.
  for (region,) in con.execute(
      "SELECT DISTINCT region_code FROM voronoi_raw ORDER BY 1"
  ).fetchall():
      con.execute(f"""
          INSERT INTO voronoi_unioned
          SELECT postcode, region_code, unioned_geom, cell_count, ST_Area(unioned_geom)
          FROM (
              SELECT postcode, region_code,
                     ST_Union_Agg(voronoi_geom) AS unioned_geom,
                     COUNT(*) AS cell_count
              FROM voronoi_raw
              WHERE region_code = '{region}'
              GROUP BY postcode, region_code
          )
          WHERE unioned_geom IS NOT NULL
      """)
      con.execute("CHECKPOINT;")
      print(region, "ok")
```

  Output (the partition at which it dies varies run‑to‑run, e.g. NE one run, WA the next):

 ```
 EE ok
  EM ok
  LN ok
  duckdb.duckdb.FatalException: FATAL Error: Failed to rollback transaction. Cannot continue operation.
  Original Error: Invalid Input Error: Skipping beyond end of binary data at position 16
  Rollback Error: Invalid Input Error: Skipping beyond end of binary data at position 16
```

  The same script against the same file on DuckDB 1.4.4 completes all 11 partitions with no error:
```
  EE ok
  EM ok
  LN ok
  NE ok
  NW ok
  SC ok
  SE ok
  SW ok
  WA ok
  WM ok
  YH ok
```
  The failing operation is plain SQL (ST_Union_Agg + INSERT + CHECKPOINT), so it should reproduce from the DuckDB CLI as well; I have only reproduced it via the Python client
  against this dataset and have not isolated a data‑independent CLI script.

### OS:

Ubuntu 24.04.4 LTS (Linux 6.17.0‑35‑generic, x86_64)

### DuckDB Version:

1.5.3 (regression); 1.4.4 works correctly

### DuckDB Client:

Python client (duckdb PyPI wheel), Python 3.12.3; spatial extension b68b309

### Hardware:

AMD Ryzen 9 3900X (12C/24T), 62 GiB RAM (~49 GiB free during the run), NVMe SSD

### Full Name:

Matt Duggan

### Affiliation:

skutter.ai

- [x] Yes, I have

- [x] Yes, I have

Yes

Maxxen:
`GEOMETRY` (and `VARIANT`?) column row groups need to keep their  `RowGroupAppendMode::REQUIRE_NEW` even after a new child row group is created. Otherwise, when we go through the index append path (`LocalTableStorage::AppendToTable`) we will initialize a regular in-place append and try to append regular chunks into a potentially shredded column data.

Follow up from https://github.com/duckdb/duckdb/pull/22060
Should fix https://github.com/duckdb/duckdb-spatial/issues/828

### DIFF
--- a/src/storage/table/row_group_collection.cpp
+++ b/src/storage/table/row_group_collection.cpp
@@ -184,7 +184,11 @@ void RowGroupCollection::AppendRowGroup(SegmentLock &l, idx_t start_row) {
 	auto new_row_group = make_uniq<RowGroup>(*this, 0U);
 	new_row_group->InitializeEmpty(types, GetColumnDataType(start_row));
 	owned_row_groups->AppendSegment(l, std::move(new_row_group), start_row);
-	row_group_append_mode = RowGroupAppendMode::APPEND_TO_EXISTING;
+
+	// Do not downgrade from REQUIRE_NEW
+	if (row_group_append_mode <= RowGroupAppendMode::SUGGEST_NEW) {
+		row_group_append_mode = RowGroupAppendMode::APPEND_TO_EXISTING;
+	}
 }
 
 optional_ptr<RowGroup> RowGroupCollection::GetRowGroup(int64_t index) {
--- a/test/sql/types/geo/geometry_shred_pk_append.test
+++ b/test/sql/types/geo/geometry_shred_pk_append.test
@@ -1,0 +1,35 @@
+# name: test/sql/types/geo/geometry_shred_pk_append.test
+# description: Append into a re-shredded rowgroup through the index append path
+# group: [geo]
+
+# Ensure that when appending to a rowgroup, the presence and state of an index does not impact the decision wether
+# a new row segment is created or not. If the row group has a non-updateable type (like GEOMETRY), new segments should
+# be created regardless, and the rowgroup should not be marked as updateable.
+
+load __TEST_DIR__/geometry_shred_pk_append.db readwrite v1.5.0
+
+statement ok
+SET geometry_minimum_shredding_size = 0;
+
+statement ok
+CREATE TABLE t(id UUID PRIMARY KEY, g GEOMETRY);
+
+statement ok
+INSERT INTO t VALUES (uuidv7(), 'POINT (0 1)'::GEOMETRY);
+
+statement ok
+CHECKPOINT
+
+statement ok
+INSERT INTO t VALUES (uuidv7(), 'POINT (0 1)'::GEOMETRY);
+
+statement ok
+CHECKPOINT
+
+statement ok
+INSERT INTO t VALUES (uuidv7(), 'POINT (0 1)'::GEOMETRY);
+
+query I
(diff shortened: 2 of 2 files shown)


Tishj:
From the description, that definitely sounds like it could also affect VARIANT, nice find 👍

Maxxen:
@Tishj do you know if the failing variant tests are a regressions or just expected changed behavior now?

Maxxen:
This should probably be closed in favor of https://github.com/duckdb/duckdb/pull/23737

Mytherin:
Fixed in https://github.com/duckdb/duckdb/pull/23763

PR #23728, opened 2026-07-09
Improve pruning via string stats

### LINKED ISSUE #23727: Row-group not pruned for string stats
### What happens?

Hi team, I found string stats are not always leveraged for all types of operations. See repro SQL for example, should be straightforward.

### To Reproduce

```sql
memory D ATTACH '/private/tmp/str_prune.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE s AS
     SELECT CASE WHEN range < 2048 THEN 'a' ELSE 'b' END AS v
     FROM range(4096);

-- Row group not pruned. 
db D EXPLAIN ANALYZE SELECT count(*) FROM s WHERE v > 'a';
╭─ Summary ───────────╮
│ Total Time: 0.0019s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ─────╮
│ Aggregates: count_star()  │
│ 1 row                 0µs │
╰─────────────┒┎────────────╯
╭─ Table Scan ┚┖────────────╮
│ Table: db.main.s          │
│ Type: Sequential Scan     │
│ Filters: v > 'a'          │
│ Row Groups Scanned: 2 / 2 │
│ 2,048 rows            0µs │
╰───────────────────────────╯

-- Prune row group correctly.
db D EXPLAIN ANALYZE SELECT count(*) FROM s WHERE v >= 'b';
╭─ Summary ───────────╮
│ Total Time: 0.0006s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ─────╮
│ Aggregates: count_star()  │
│ 1 row                 0µs │
╰─────────────┒┎────────────╯
╭─ Table Scan ┚┖────────────╮
│ Table: db.main.s          │
│ Type: Sequential Scan     │
│ Filters: v >= 'b'         │
│ Row Groups Scanned: 1 / 2 │
│ 2,048 rows            0µs │
╰───────────────────────────╯
```

### OS:

macos

### DuckDB Version:

985b2f2515adf44ea65dd4239c7bfb30981b94dc

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23727

In the current implementation, we're handling `>=` / `>` && `<=` / `<` the same way, which miss certain optimization opportunities.


dentiny:
Hi @lnkuiper , I'm wondering if you're willing to take a look when you have some time? Hopefully it's not too complicated. :) Thanks!

lnkuiper:
Thanks for the PR! Great idea. Some findings below.

This returns incorrect results for:
```sql
# name: test/temp.test
# description: Strict string zonemap pruning must respect merged truncated stats
# group: [temp]

statement ok
CREATE TABLE strings AS
SELECT CASE WHEN range < 2048 THEN repeat('a', 12) ELSE repeat('a', 13) END AS value
FROM range(4096);

query I
SELECT count(*) FROM strings WHERE value > repeat('a', 12);
----
2048

query I
SELECT count(*) FROM strings WHERE value <> repeat('a', 12);
----
2048
```

`MergeStats` should downgrade the result to truncated whenever the relative ordering cannot be established beyond the stored prefix. The tests should cover values longer than `CURRENT_MAX_STRING_MINMAX_SIZE` across append batches/row groups.

dentiny:
> MergeStats should downgrade the result to truncated whenever the relative ordering cannot be established beyond the stored prefix

You're right -- it seems it be an existing issue with `MergeStats`, we're safe for now only because we're currently read more row groups than necessary; update the impl to consider truncated stats when merge, also add thorough test cases.

One left issue I plan to followup in the next PR (which doesn't affect correctness): only one row group is required.
```sql
query II
EXPLAIN ANALYZE SELECT sum(payload) FROM long_strings WHERE v = repeat('b', 12);
----
analyzed_plan	<REGEX>:.*Row Groups Scanned: 2 / 2.*

query I
SELECT sum(payload) FROM long_strings WHERE v = repeat('b', 12);
----
2048
```

lnkuiper:
Great fix! Thanks :)

PR #23730, opened 2026-07-09
Add Benchmark for GROUP BY on dictionary compressed column 

gropaul:
Edit: https://github.com/duckdb/duckdb/pull/23738 fixes the issue structurally, so this PR is only the benchmark that showed the performance optimization.

---

Hi dear DuckDB team,

Getting the `groups.size()` showed to be a bottleneck in my flame chart plots, so I fixed it to have one `const idx_t groups_size = groups.size();`. 

I added a benchmark running `SELECT DISTINCT l_shipinstruct FROM lineitem ORDER BY l_shipinstruct;`. For this on SF1, the changes query runtime from 15ms to 7ms. 

For larger scale factors, the speedup was similar. 

Do you know why these `.size()` calls are so expensive?

### DIFF
--- a/benchmark/tpch/aggregate/lineitem_distinct_on_dictionary_column.benchmark
+++ b/benchmark/tpch/aggregate/lineitem_distinct_on_dictionary_column.benchmark
@@ -1,0 +1,18 @@
+# name: benchmark/tpch/aggregate/lineitem_distinct_on_dictionary_column.benchmark
+# description: DISTINCT ON over Lineitem
+# group: [aggregate]
+
+include benchmark/tpch/tpch_load.benchmark.in
+
+name Lineitem Distinct On (Dictionary Compressed Column)
+group aggregate
+subgroup tpch
+
+run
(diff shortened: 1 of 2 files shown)


carlopi:
> Do you know why these `.size()` calls are so expensive?

No, I didn't, and this was a overly specific enough question that triggered me. Would commit at https://github.com/carlopi/duckdb/commit/d08bd41b911232a9fb8ce558043e0f4b9bfd0bfd would make any sense? Would that avoid the regression / fix the problem (and possibly some other as well) ?

gropaul:
Yes, it does (see https://github.com/duckdb/duckdb/pull/23738#issuecomment-4926975621), thanks!

carlopi:
@gropaul: unsure if adding the benchmark from https://github.com/duckdb/duckdb/pull/23730/changes/d204f5e9f8d06123fd20f51e36ee755f57c87102 would still be of value, I didn't add it to my PR

gropaul:
@carlopi I reopened it with just the benchmark 👍

PR #23731, opened 2026-07-09
Add external resource type registry + SQL level (and C++ level) `create_external_resource`

carlopi:
An instance-scoped, append-only registry of 'external resource types' — named kinds of external resource (a cloud stack, a container, ...) described purely by lifecycle callbacks referenced by name (create/status/destroy/resolve), resolved lazily against the catalog. Types are data, not C++ objects.

Table functions:
 - register_external_resource_type(name, kind := , create_function := , ...)
 - duckdb_external_resource_types() lists them
 - create_external_resource(type [, params]) provisions: runs create, polls status
      to 'ready', enforces the catalog baseline (uri + attached_db_type), returns the
      endpoint + a deleter binding
 - destroy_external_resource(fn, payload) invokes a deleter binding

This is the standalone provisioning primitive; the WITH EXTERNAL RESOURCE SQL surface builds on it on top.

The patches to `duckdb-aws` are:
* 0002, currently sent as PR at https://github.com/duckdb/duckdb-aws/pull/161, this is not structural, only improving rough edges found while calling into `cloudformation_create_stack`
* 0003, an user of the C++ API that can provision resources directly from C++ extension interface. This is meant as workable example, but will likely improve / be iterated on `duckdb-aws` side.

Thanks to various colleagues at DuckLabs for early feedback provided.


Maxxen:
Cool! Lots of thoughts on this...
- Instead of having an external resource registry - why can't we store these in the catalog instead?
- passing the function names to define an interface is cute - but I worry that this will cause all kinds of issues if referencing functions/macros in different schemas/databases, particularly if the external resource type does not live in a schema of its own (e.g. is not part of the catalog).
- Couldn't we use the new work on TRIGGERs to something similar without needing extra c++ code? E.g. `aws` could create a table in an internal/ephemeral `system.aws` schema (inside the system database, not persistent) with triggers to fire when this table is modified (which could be protected by e.g. extension provided table/pragma functions).

carlopi:
Thanks for taking a look! This is ground work for https://github.com/duckdb/duckdb/pull/23736 (and a not yet ready follow up around `CREATE EXTERNAL RESOURCE` and co).

I need to process the input, one first question I got: wouldn't this be just an implementation details of sort?

The way I see it is that it's nice there is a SQL level interface, since that allows easy prototyping / fixing up / customizations of the various callbacks, and it's nice to allow this C++ side (from extensions in particular). The actual mechanism it's not set in stone, I would need to think better what would be the advantages of having this in the catalog. I guess one thing that pushed towards this implementation is that extensions are at the moment load once, so an additive model sounded simpler (for sure to implement, possibly also to reason about), but I am not sure I see the various tradeoffs.

rustyconover:
Could this be more generic, in particular a "waiter" pattern for any resource, or really why not just make it part of the ATTACH call rather than WITH EXTERNAL RESOURCE?  Couldn't it just be an attach option.

carlopi:
> Could this be more generic, in particular a "waiter" pattern for any resource, or really why not just make it part of the ATTACH call rather than WITH EXTERNAL RESOURCE? Couldn't it just be an attach option.

There is a third PR incoming that would add surface for generic:
```
CREATE EXTERNAL RESOURCE 'extension:my-recipe' AS my_resource (OPT1, OPT2 {'a': 'b'});
----
DESTROY EXTERNAL RESOURCE my_resource;
```

Goal of the PR is doing the groundwork for the specific case of `Catalogs`, since those have the nice simplification that they can have their lifetime tied to `AttachedDatabase`.

Would the above work in your case?

rustyconover:
I mean sure!  I could see that sort of working in ways beyond cloudformation stacks.

Think about it being used for predictive model training/creation.

What I'd press on is `CREATE EXTERNAL RESOURCE` should have a mode where its non blocking, and you can poll for success.

You should also be able to select messages/updates/progress reports on the creation of that resource.

Can resources be scoped to catalogs?  What about transactions? Are they tied into transactions?

carlopi:
> What I'd press on is CREATE EXTERNAL RESOURCE should have a mode where its non blocking, and you can poll for success.

Yes, fully agree, allowing `async` creation would allow more powerful semantics. I was thinking default could be `sync`, and the `async` version to be via explicit opt-in, there is a design for that, but will land as a PR on top.

carlopi:
One idea is that implementation of this feature is somehow still an implementation detail, and the actual mechanic could also be improved on keeping the SQL surface stable (introduced in the follow-up). Users of the low level primitives (like `aws` extension registering the functions) are then expected to change, but that's OK.

That should (in my mind) allow some extra flexibility in allowing a given implementation now, but swapping to more elaborate one (see @Maxxen comments) at a later stage.

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+			throw IOException("create_external_resource: timed out awaiting readiness for \"%s\" (last state '%s')",
+			                  bind_data.type_name, status_state);
+		}
+		std::this_thread::sleep_for(std::chrono::seconds(5));
+	}
+
+	// Baseline enforcement: an catalog resource must expose 'uri' + 'attached_db_type' in its result.
+	string uri, attached_db_type;
+	bool has_uri = MapTryGet(status_result, "uri", uri);
+	bool has_type = MapTryGet(status_result, "attached_db_type", attached_db_type);
```
pdet, line 269: What if result is not a map?

```
CREATE MACRO s(h) AS TABLE SELECT 'ready' AS state, 42 AS result;
CALL register_external_resource_type('boom', kind:='compute', create_function:='c', status_function:='s');
SELECT * FROM create_external_resource('boom');

File: src/function/table/system/create_external_resource.cpp [hunk 2]
```diff
@@ -0,0 +1,250 @@
...
+		throw InvalidInputException(
+		    "create_external_resource: resource type \"%s\" has no status function (required to await readiness)",
+		    bind_data.type_name);
+	}
+
+	// Blocking: poll status(handle) until `state` is terminal ('ready' / 'failed'). A synchronous type's
+	// status simply returns 'ready' on the first poll.
+	Value status_result;
+	auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(300);
+	while (true) {
```
pdet, line 228: What happens for canceled queries? (ctrl c)

pdet: Maybe a better example

```sql

SET max_execution_time = 2000; 
CREATE SEQUENCE seq;
CREATE MACRO c(p) AS TABLE SELECT MAP {'id':'r'} AS handle;
CREATE MACRO s(h) AS TABLE
SELECT CASE WHEN nextval('seq') >= 3 THEN 'ready' ELSE 'pending' END AS state,
MAP {'uri':'u','attached_db_type':'quack'} AS result;
CALL register_external_resource_type('t', kind:='catalog', create_function:='c', status_function:='s');
SELECT uri FROM create_external_resource('t');
```

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+	if (result->HasError()) {
+		throw IOException("create_external_resource: create function \"%s\" failed: %s", type->create_function,
+		                  result->GetError());
+	}
+	if (result->RowCount() == 0) {
+		throw InvalidInputException("create_external_resource: create function \"%s\" returned no rows",
+		                            type->create_function);
+	}
+	// Contract: `create` returns a single 'handle' column.
+	auto handle = result->GetValue(0, 0);
```
pdet: What happens if we throw before  calling `destroy_function`?

pdet (commented):
Thanks for the PR, a few comments.

carlopi:
Thanks a lot for the review, I'll get back on the comments, and ping you back.

pdet:
btw @carlopi this PR is targeting `main` is that correct?

src/main/external_resource_type_registry.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+vector<ExternalResourceType> ExternalResourceTypeRegistry::List() const {
+	lock_guard<mutex> guard(lock);
+	return types;
+}
+
+unique_ptr<ExternalResourceType> ExternalResourceTypeRegistry::Lookup(const string &name) const {
+	lock_guard<mutex> guard(lock);
+	// Latest-wins: scan newest-first so a later registration shadows an earlier one.
```
pdet: Isn't this a bit too problematic? Maybe we should reject duplicates unless explicitly requested to do replace it?

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+	if (!type) {
+		throw InvalidInputException("create_external_resource: unknown resource type \"%s\"", bind_data.type_name);
+	}
+	if (type->create_function.empty()) {
+		throw InvalidInputException("create_external_resource: resource type \"%s\" has no create function",
+		                            bind_data.type_name);
+	}
+
+	// Separate internal connection: the current connection's context lock is held here.
+	Connection con(DatabaseInstance::GetDatabase(context));
```
pdet, line 169: Can this be problematic?
e.g.,
```sql
CREATE SCHEMA app;
CREATE MACRO app.c(p) AS TABLE SELECT MAP{'id':'r'} AS handle;
CREATE MACRO app.s(h) AS TABLE SELECT 'ready' AS state, MAP{'uri':'u','attached_db_type':'q'} AS result;
SET search_path = 'app';
CALL register_external_resource_type('x', kind:='catalog', create_function:='c', status_function:='s');
SELECT uri FROM create_external_resource('x');
```

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+	return_types.emplace_back(map_vv);
+	return std::move(result);
+}
+
+static unique_ptr<GlobalTableFunctionState> CreateExternalResourceInit(ClientContext &context,
+                                                                       TableFunctionInitInput &input) {
+	return make_uniq<CreateExternalResourceState>();
+}
+
+static void CreateExternalResourceFunction(ClientContext &context, TableFunctionInput &data_p, DataChunk &output) {
```
pdet, line 152: How does this work for twhen the query has a LIMIT 0, or it's stacked? can we add tests for that?
e.g., 
  `SELECT * FROM create_external_resource('my_db') LIMIT 0` , `... create_external_resource('my_db') a, ...('my_db') b `

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+	if (type->status_function.empty()) {
+		throw InvalidInputException(
+		    "create_external_resource: resource type \"%s\" has no status function (required to await readiness)",
+		    bind_data.type_name);
+	}
+
+	// Blocking: poll status(handle) until `state` is terminal ('ready' / 'failed'). A synchronous type's
+	// status simply returns 'ready' on the first poll.
+	Value status_result;
+	auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(300);
```
pdet: nit: magic number

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+			break;
+		}
+		if (status_state == "failed") {
+			throw IOException("create_external_resource: resource \"%s\" reported state 'failed'", bind_data.type_name);
+		}
+		if (std::chrono::steady_clock::now() >= deadline) {
+			throw IOException("create_external_resource: timed out awaiting readiness for \"%s\" (last state '%s')",
+			                  bind_data.type_name, status_state);
+		}
+		std::this_thread::sleep_for(std::chrono::seconds(5));
```
pdet: nit: magic number

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+		throw InvalidInputException("create_external_resource: unknown resource type \"%s\"", bind_data.type_name);
+	}
+	if (type->create_function.empty()) {
+		throw InvalidInputException("create_external_resource: resource type \"%s\" has no create function",
+		                            bind_data.type_name);
+	}
+
+	// Separate internal connection: the current connection's context lock is held here.
+	Connection con(DatabaseInstance::GetDatabase(context));
+	auto sql = "SELECT * FROM " + type->create_function + "(" + bind_data.params.ToSQLString() + ")";
```
pdet: KeywordHelper::WriteOptionallyQuoted

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,250 @@
...
+static void DestroyExternalResourceFunction(ClientContext &context, TableFunctionInput &data_p, DataChunk &output) {
+	auto &state = data_p.global_state->Cast<DestroyExternalResourceState>();
+	if (state.done) {
+		return;
+	}
+	auto &bind_data = data_p.bind_data->Cast<DestroyExternalResourceBindData>();
+
+	// CALL <deleter_function>(<deleter_payload>) on a separate internal connection.
+	Connection con(DatabaseInstance::GetDatabase(context));
+	auto sql = "SELECT * FROM " + bind_data.deleter_function + "(" + bind_data.deleter_payload.ToSQLString() + ")";
```
pdet: KeywordHelper::WriteOptionallyQuoted

src/include/duckdb/main/external_resource_type_registry.hpp
```diff
@@ -0,0 +1,65 @@
...
+class DatabaseInstance;
+class ClientContext;
+
+//! An external resource type: a named kind of external resource DuckDB can provision and reap (a
+//! container, a cloud stack, ...), described purely by its lifecycle callbacks. Every field is a
+//! simple, SQL-expressible value — resource types are data, not C++ objects. The callbacks are
+//! referenced by name and resolved against the catalog lazily (when invoked), so they may be defined
+//! independently. `WITH EXTERNAL RESOURCE '<type>' ... ATTACH|CONNECT` is the first consumer, but a type
+//! says nothing about attaching — the backend/endpoint is what the `status` callback returns at
+//! provision time.
```
pdet: nit: this seems a bit excessive comment, can we try to keep it down to 1-2 lines max.

test/sql/table_function/create_external_resource.test
```diff
@@ -0,0 +1,80 @@
+# name: test/sql/table_function/create_external_resource.test
```
pdet, line 1: I think we also need to be a bit more aggressive in our testing, the whole poll/lifecycle is basically untested, also  create/destroy contract and error paths, registry semantics

pdet (commented):
A couple more comments.

carlopi:
Thanks a ton @pdet for the time and the comments, I like it way more once hardened on the various point you raised.
In particular, solved the cross schema / schema changing over lifetime by storing the reference schema name, made this interruptible, and various other improvements. Thanks!

Left not implemented:
* LIMIT case, this is indeed potentially confusing, but while this is exposed to SQL API, this is mostly for compose in a more complex feature, the SQL feature might also (at a later point, be marked internal)
* more testing, that I would propose to be moved to a follow up PR

Failure looks weird / random to me, I think should be irrelevant for this specific PR, and this is ready for another round.

src/function/table/system/external_resource_types.cpp
```diff
@@ -0,0 +1,168 @@
...
+	if (input.inputs[0].IsNull() || StringValue::Get(input.inputs[0]).empty()) {
+		throw InvalidInputException("register_external_resource_type: name must not be NULL or empty");
+	}
+	type.name = StringValue::Get(input.inputs[0]);
+	type.origin = "user";
+	// Capture the registration-time search path so the (by-name, lazily-resolved) callbacks keep resolving
+	// from create_external_resource's internal connection — the same reason views/macros record theirs.
+	type.search_path = CatalogSearchEntry::ListToString(ClientData::Get(context).catalog_search_path->GetSetPaths());
+
+	for (auto &np : input.named_parameters) {
```
pdet, line 39: Should we validate malformed function names?

I believe the following should fail here:

`  CALL register_external_resource_type('q@x', kind:='catalog', create_function:='""', status_function:='st');`

src/function/table/system/create_external_resource.cpp
```diff
@@ -0,0 +1,365 @@
...
+		throw InvalidInputException(
+		    "create_external_resource: resource type \"%s\" has no status function (required to await readiness)",
+		    bind_data.type_name);
+	}
+
+	// Blocking: poll status(handle) until `state` is terminal ('ready' / 'failed'). A synchronous type's
+	// status simply returns 'ready' on the first poll.
+	Value status_result;
+	bool has_deadline = bind_data.timeout_seconds > 0;
+	auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(bind_data.timeout_seconds);
```
pdet, line 227: nit: Should we check for overflow  on timeout_seconds / poll_interval_seconds ?

pdet (commented):
Thanks @carlopi  for all the work, is looking great, had another pass, just found a couple of minor things, then it's good on my side.

I haven't really checked the AWS patch as I'm not very familiar with it, maybe @Tmonster  can have a look there?

Tmonster:
Aws patch seems to only be in cloud formation stuff, which Carlo has been working on, so looks ok to me 👍

.github/patches/extensions/aws/0003-register-quack-on-ec2-resource.patch
```diff
@@ -0,0 +1,248 @@
...
++	auto sql = "SELECT handle FROM cloudformation_create_stack(" + Value(template_url).ToSQLString() +
++	           ", region := " + Value(region).ToSQLString();
++	if (!tp_keys.empty()) {
++		auto template_params = Value::MAP(LogicalType::VARCHAR, LogicalType::VARCHAR, tp_keys, tp_values);
++		sql += ", template_parameters := " + template_params.ToSQLString();
++	}
++	sql += ")";
++
++	Connection con(DatabaseInstance::GetDatabase(context));
++	auto res = con.Query(sql);
```
Tmonster: Isn't the point of this PR to add `create_external_resource` to C++? 
Is there a reason for creating this as SQL then?

Tmonster (commented):
Is there a reason 0003-register-quack-on-ec2-resource.patch needs to be added in this PR?

src/include/duckdb/main/external_resource_type_registry.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+namespace duckdb {
+class DatabaseInstance;
+class ClientContext;
+
+//! A named kind of external resource DuckDB can provision and reap, described purely by lifecycle callbacks
+//! referenced by name. Every field is a simple string — resource types are data, resolved lazily when invoked.
+struct ExternalResourceType {
+	//! The type name, e.g. 'quack@aws:ec2'.
+	string name;
```
Tmonster, line 23: have we thought about what might happen if two extensions both try to register external resource types with the same name? 

do we differentiate using the `origin` string? Or should we add an internally generated `UUID` or something?

src/include/duckdb/main/external_resource_type_registry.hpp
```diff
@@ -0,0 +1,65 @@
...
+	string origin;
+	//! The caller's catalog search path captured at registration (serialized). The callbacks are resolved
+	//! against it when invoked — like views/macros capture their creation-time search path — so unqualified
+	//! callback names keep resolving from create_external_resource's internal connection, regardless of the
+	//! search path in effect when create runs. Empty means "use the internal connection's default".
+	string search_path;
+};
+
+//! In-memory, instance-scoped registry of external resource types (shared across connections). A name is
+//! claimed by its first registration (first-wins): re-registering an existing name errors, and there is no
```
Tmonster, line 46: ah, here is the answer. I feel like we can get around the name conflicts with minimal effort no?

Tmonster (commented):
Is there any thought towards registering persistent external resource types? Or will that be a follow up?

src/include/duckdb/main/external_resource_type_registry.hpp
carlopi: Right, and this is even more of an issue when this happens mid extension load (say via C++ api) since the I guess exception bubbles up, and leaves extension half initialized. Thanks for spotting this.

.github/patches/extensions/aws/0003-register-quack-on-ec2-resource.patch
carlopi: Right, thank at ton. Let's remove the patch, that was mostly here for illustrative purposes / allow to demo this. I'll come back `duckdb/duckdb-aws` side.

src/function/table/system/external_resource_types.cpp
```diff
@@ -0,0 +1,168 @@
...
+                                                                 vector<string> &names) {
+	auto result = make_uniq<RegisterExternalResourceTypeBindData>();
+	auto &type = result->type;
+
+	// Only `name` is positional; kind/create (required) and status/destroy/resolve (optional) are named.
+	if (input.inputs[0].IsNull() || StringValue::Get(input.inputs[0]).empty()) {
+		throw InvalidInputException("register_external_resource_type: name must not be NULL or empty");
+	}
+	type.name = StringValue::Get(input.inputs[0]);
+	type.origin = "user";
```
carlopi, line 34: To be reviewed, can this be either SQL or extension name, or maybe empty or extension name? possibly there is some method in ExtensionLoader that can help lift this somehow.

src/function/table/system/create_external_resource.cpp
carlopi: Added a test showing the limitation.

src/function/table/system/create_external_resource.cpp
carlopi: Handled!

src/include/duckdb/main/external_resource_type_registry.hpp
carlopi: I think I have a way to improve this, but I would like to avoid generalizing with no providers.

carlopi:
Aws patches have been absorbed via bump, or removed, the actual implementation of the callbacks I'll open via a duckdb-aws PR (to be opened after this lands). I think the main unresolved comment is increased testing, and I just realized that a mock test resource provided by a test extension would be the best path forward, but that's not there as of now.

pdet:
Thanks for the all the work in the PR!

PR #23732, opened 2026-07-09
Support SHOW <setting>

JelteF:
In Postgres SHOW can be used to show the current value of a setting:

```sql
> SHOW timezone;
 TimeZone
──────────
 CET
(1 row)

Time: 6.415 ms
```

In DuckDB SHOW used as an alias for DESCRIBE, which means it has a few special ones (like `show tables`), and other than that it describes tables and queries. That means it errors out when trying to use it with the name of a setting:

```sql
memory D show timezone;
Catalog Error:
Table with name timezone does not exist!
Did you mean "pg_indexes"?

LINE 1: show timezone;
             ^
```

To get a setting value you need to use the longer `SELECT
current_setting('timezone')`.

This changes SHOW to work for settings and prefer settings over tables with the same name. It also adds deprecation warnings saying that people should use DESCRIBE instead of SHOW to describe tables and queries.


Dtenwolde:
Thanks for the PR, I think this is a nice addition! Instead of overloading `SHOW` to return either a setting or a table description, could we make the grammar a bit more explicit with `SHOW SETTING <setting_name>`?

That would also avoid having to define a precedence between settings and tables when names collide, which I’m not sure we want here.

I also don’t think bare `SHOW <setting_name>` is necessarily more intuitive than the explicit form, though maybe that’s just me.

JelteF:
The main reason for reusing bare `SHOW` here instead of introducing something new (like `SHOW SETTING`) is that it allows SQL scripts written for Postgres to work "as is" in DuckDB. As well as people that familiar with postgres to use their normal commands. i.e. it has taken me a long time until I got used to getting the value of a setting using `SELECT current_setting('<setting>')` both because it's longer to type and it doesn't autocomplete, but also because it's just second nature to type `show timezone` for me at this point.

> That would also avoid having to define a precedence between settings and tables when names collide, which I’m not sure we want here.

To be clear, the idea is that that precedence situation would only be a temporary thing to allow for a deprecation period. After N releases we'd remove `SHOW <tablename>` and `SHOW <query>` support, and require users to use `DESCRIBE <tablename>` and `DESCRIBE <query>`. I'd also be fine with dropping support for `SHOW <tablename>` and `SHOW <query>` immediately. These forms of `SHOW` are not used by any other database than DuckDB.

Mytherin:
Thanks - but I think we need to approach this carefully given that this is a breaking SQL-level change. 

How about we add a setting for `show_behavior` which defaults to `auto` and can be set to `setting` or `table`. `auto` would then prefer showing tables over settings (i.e. inverse of the behavior implemented here) to preserve backwards compatibility by default, and print the deprecation warning if a table is shown.

Dtenwolde:
@JelteF Could you undraft it so CI can run? :)

src/parser/peg/matcher.cpp
```diff
@@ -1501,6 +1501,11 @@ Matcher &MatcherFactory::CreateMatcher(const char *grammar, const char *root_rul
 
 	// suppress suggestions for catch-all rules that would pollute statement-level autocomplete
 	SuppressSuggestions("ExpressionStatement");
+	// Using SHOW to describe a table/query is deprecated - parse these forms but do not autocomplete them after SHOW
+	// (only setting names and the special SHOW forms are offered). DESCRIBE/SUMMARIZE still complete tables and
+	// queries.
+	SuppressSuggestions("ShowDeprecatedQualifiedTableName");
+	SuppressSuggestions("ShowDeprecatedSelect");
```
Dtenwolde: Should we still suppress auto complete suggestions? Probably fine as we are deprecating them, but just want to ask since it's now a setting

test/sql/function/autocomplete/show.test
```diff
@@ -12,23 +12,102 @@ DESCRIBE 	0
...
+SELECT count(*) FROM sql_auto_complete('SHOW tab') WHERE suggestion = 'tables ';
+----
+1
+
+# SHOW does NOT complete arbitrary table names (using SHOW to describe a table is deprecated)
+statement ok
+CREATE TABLE thermostat(id INTEGER);
+
+query I
+SELECT count(*) FROM sql_auto_complete('SHOW therm') WHERE suggestion = 'thermostat';
```
Dtenwolde, line 71: Shouldn't this now have returned `thermostat` with the setting set to `auto`?

Dtenwolde: I guess this doesn't happen because we avoid autocomplete suggestions for `ShowDeprecatedQualifiedTableName`

tools/shell/shell.cpp
```diff
@@ -941,6 +941,13 @@ ShellState &ShellState::Get() {
 	return *GetReference();
 }
 
+static bool ResultIsDescribeShaped(const duckdb::QueryResult &result) {
```
Dtenwolde, line 939: I don't know enough about the shell to know if this is the way to go, but it seems error prone to check if there are 6 columns and the first one is named `column_name` and the second `column_type`. @Mytherin perhaps you know if this is okay?

Dtenwolde (commented):
Some minor comments around the shell and the autocomplete functionality. Otherwise looks good to me. To me, in general the `SHOW/DESCRIBE` syntax and functionality is not really intuitive and somewhat confusing, but this seems like a nice addition :)

src/parser/peg/matcher.cpp
JelteF: I think it doesn't really make sense for `show_behavior=table` (given that the settings don't work). But given that:
1. the `show_behavior` setting is only really meant to be used for existing scripts
2. tab-complete is only for interactive usecases 

I don't think it's worth gating the suppression on the value of `show_behavior` setting.

tools/shell/shell.cpp
JelteF: Yeah, I didn't like this either but couldn't think of any other way. I believe in practice it should be fine. The `SHOW <setting>` form only returns a single column. The "special" SHOW form have a fixed and different result shape than this. So the only case that matches is the deprecated describe-form of SHOW. 

Another option would be to simply not do this at all and show the describe-form result in the CLI without using the describe rendermode

JelteF:
I fixed the merge conflicts in this PR. The failing test seems to be a fluke that a rerun should fix.

Is anything needed from me on this PR at this point?

taniabogatsch:
> Is anything needed from me on this PR at this point?

If you could just merge with `main` one more time to be sure? Then, @Dtenwolde gave the OK! :)

JelteF:
PR failed to compile after the merge, but it's all green now.

Mytherin:
Thanks - looks good to me now, can you just fix the merge conflicts?

JelteF:
Merge commit is fixed now

Mytherin:
Thanks!

JelteF:
:facepalm: apparently these grammar changes pushed the grammar over a compiler limit in string literal size.... 

```
error: string literal of length 65910 exceeds maximum length 65536 that C++ compilers are required to
```

Mytherin:
I think we need to turn the grammar into a byte array similar to what we do for other big inlined data sets

Tishj:
Isn't this something we want to leave to a dialect extension, especially since it's a breaking change otherwise?

JelteF:
@Tishj my understanding is that DuckDB tries to match Postgres behavior unless it has a reason not to. In this case the only reason would be backwards compatibility, and since `SHOW <table>` is primarily used interactively it's not too big of an issue break compatibility there. And also, with the current change we won't break backwards compatibility (yet), i.e. the `auto` behaviour is a pure superset of the previous behaviour..

JelteF:
The PR is green now. @Mytherin can you add it to the merge queue?

Mytherin:
@Tishj I think this is alright as a breaking change, `SHOW` and `DESCRIBE` being synonyms doesn't make much sense imo so this is cleaning up our SQL dialect. `DESCRIBE` is the better term and also compatible with the original source of this (MySQL). This is also a good time to push for any breaking changes and we have a clean migration strategy here. This just needs to be well documented in the v2.0 blog post.

Mytherin:
Thanks!

PR #23733, opened 2026-07-09
Enable DICT_FSST by default

### LINKED ISSUE #23741: DICT_FSST encoded column gets incorrect result
### What happens?

### To Reproduce

```sql
# name: test/sql/storage/compression/dict_fsst/dict_fsst_expression_filter.test
# description: Test expression filters over DICT_FSST strings
# group: [dict_fsst]

require vector_size 2048

load {TEST_DIR}/dict_fsst_expression_filter.db readwrite v2.0.0

statement ok
CREATE TABLE partial_segment AS
SELECT concat('HumanEval/', i::VARCHAR) AS task_id
FROM range(0, 164) tbl(i);

statement ok
CREATE TABLE full_vector AS
SELECT concat('HumanEval/', i::VARCHAR) AS task_id
FROM range(0, 2048) tbl(i);

statement ok
CHECKPOINT

restart

query I
SELECT count(*) FROM pragma_storage_info('partial_segment')
WHERE column_name = 'task_id'
  AND segment_type = 'VARCHAR'
  AND compression = 'DICT_FSST'
----
1

query I
SELECT count(*) FROM pragma_storage_info('full_vector')
WHERE column_name = 'task_id'
  AND segment_type = 'VARCHAR'
  AND compression = 'DICT_FSST'
----
1

query I
SELECT count(*) FROM partial_segment
WHERE split_part(task_id, '/', 2)::INTEGER >= 10
----
154

statement ok
DELETE FROM partial_segment
WHERE split_part(task_id, '/', 2)::INTEGER >= 10

query I
SELECT count(*) FROM partial_segment
----
10

query I
SELECT count(*) FROM full_vector
WHERE split_part(task_id, '/', 2)::INTEGER >= 10
----
2038

statement ok
DELETE FROM full_vector
WHERE split_part(task_id, '/', 2)::INTEGER >= 10

query I
SELECT count(*) FROM full_vector
----
10
```
get error message
```sh
1. test/sql/storage/compression/dict_fsst/dict_fsst_expression_filter.test:40
Query unexpectedly failed (test/sql/storage/compression/dict_fsst/dict_fsst_expression_filter.test:40)
 (test/sql/storage/compression/dict_fsst/dict_fsst_expression_filter.test:40)!
SELECT count(*) FROM partial_segment
WHERE split_part(task_id, '/', 2)::INTEGER >= 10;
Actual result:
Conversion Error: Could not convert string '' to INT32

LINE 2: WHERE split_part(task_id, '/', 2)::INTEGER >= 10
                                         ^
```

### OS:

macos

### DuckDB Version:

e16f01941e047798044fb0b5832118f88c21dfb1

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Hi team, when I was doing [data ingestion benchmark](https://github.com/dentiny/duckdb/pull/125), I found best compression detection is obvious a bottleneck from on-chip profiling.
image

- DICT and FSST are candidates from v0.10.2 to v1.2.0, with DICT_FSST enabled through v1.3.0 to latest (one main reason is to save algo section latency)
- In the current implementation, IIRC, we're using storage version v0.10.2 (which could be confirmed ); this PR tries to bumpup the version to v2.0.0 to "release" DICT_FSST so users don't need to explicitly set the storage version when they create a table

Happy to discuss, let me know if you have any concerns!

Closes https://github.com/duckdb/duckdb/issues/23741


src/storage/compression/dict_fsst.cpp
```diff
@@ -179,49 +179,9 @@ void DictFSSTSelect(ColumnSegment &segment, ColumnScanState &state, idx_t vector
...
-		}
-		sel_count = approved_tuple_count;
-
-		result.Dictionary(scan_state.dictionary, dict_sel, vector_count);
-		return;
-	}
-	// fallback: scan + filter
-	DictFSSTCompressionStorage::StringScan(segment, state, vector_count, result);
+	// The reusable dictionary contains a reserved NULL entry, so filter actual rows instead.
+	DictFSSTCompressionStorage::StringScanPartial<false>(segment, state, vector_count, result, 0);
```
Mytherin: Why are we removing the above optimization?

Mytherin (commented):
Thanks! I think we should indeed be moving to v2.0 storage by default. One comment

src/storage/compression/dict_fsst.cpp
dentiny: Hey! I found current implementation has a bug, I reported here https://github.com/duckdb/duckdb/issues/23741

Mytherin: Should we then try to fix that bug instead of removing the optimization entirely?

Mytherin: I think we should at least split this into two separate PRs (setting storage version to 2.0 by default, and addressing this bug).

dentiny: Sounds good, updated!

Mytherin:
Thanks!

PR #23734, opened 2026-07-09
Move type pushdown after filter pushdown, don't push it filter is present

### LINKED ISSUE #8310: Epic: Duckdb projection expression pushdown

myrrc:
There is currently a correctness bug in type pushdown optimizer: in
SELECT col::T where condition(), if condition is not pushed, it is reodered over GET. However, if condition() doesn't contain "col", col::T may be pushed to GET so that chain is FILTER -> GET [-> CAST inside]. if condition() filtered out rows on which CAST errors i.e.

SELECT col::UTINYINT where this_filters_col_eq_257(), pushdown turns a working query into an erroring one.

Thus we disallow cast pushdown if after filter pushdown any filter survived.

### DIFF
--- a/src/optimizer/optimizer.cpp
+++ b/src/optimizer/optimizer.cpp
@@ -214,16 +214,6 @@ void Optimizer::RunBuiltInOptimizers() {
 		plan = filter_pullup.Rewrite(std::move(plan));
 	});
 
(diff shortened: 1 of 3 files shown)


lnkuiper:
Thanks!

PR #23735, opened 2026-07-09
bump iceberg

lnkuiper:
For the credential refresh


lnkuiper:
This failure is unrelated to the Iceberg bump, there is some rust stuff going on when building Lance

PR #23736, opened 2026-07-09
EXTERNAL RESOURCE: allow first class externally managed resources (SQL syntax on top of table functions)

carlopi:
Based on https://github.com/duckdb/duckdb/pull/23731 infrastructure, this PR adds SQL level interface for allowing interfacing with external resources as first class citizens of DuckDB SQL.

## 1. Syntax the parser now accepts

Durable — the resource outlives any one attachment, and only `DESTROY` will tear it down:
```sql
CREATE   EXTERNAL RESOURCE '<resource-type>' AS <name> [(RESOURCE_OPTS...)];
REGISTER EXTERNAL RESOURCE '<resource-type>' AS <name> FROM <handle>;
DESTROY  EXTERNAL RESOURCE <name>;
SHOW [ALL] EXTERNAL RESOURCES;
```

Scoped - the resource lives and dies with the attachment/connection:
```sql
ATTACH  TO NEW TEMPORARY EXTERNAL RESOURCE '<resource-type>' [(RESOURCE_OPTS...)] AS <db-name> [(ATTACH_OPTS...)];
CONNECT TO NEW TEMPORARY EXTERNAL RESOURCE '<resource-type>' [(RESOURCE_OPTS...)];

ATTACH  TO EXTERNAL RESOURCE <name> AS <db-name> [(ATTACH_OPTS...)];
CONNECT TO EXTERNAL RESOURCE <name> [(CONNECT_OPTS...)];
```

`NEW TEMPORARY` is the ownership marker: it provisions a fresh resource the statement OWNS and reaps on `DETACH`/`DISCONNECT`. Without it, `<name>` refers to an already-registered resource the statement only bottow — resources are managed explicitly and require a `DESTROY`.

### 2a. CONNECT TO NEW EXTERNAL RESOURCE — the resource lives and dies with the connection

```sql
LOAD aws;
CALL enable_logging('ExternalResource', storage='stdout');
CONNECT TO NEW EXTERNAL RESOURCE 'aws:cloudformation:quack-on-ec2' (REGION 'us-east-1');
DISCONNECT;
```
(logging to in-memory table / file might be preferable, consider https://duckdb.org/docs/lts/operations_manual/logging/overview)
Note this requires about 2 minutes to spin up an EC2 machine from a given public recipe that references a publicly available AMI, setup quack, obtain a SSL certificate, and some polling wait time. This can be significantly reduced by improving setup, but that's independent.

The log narrates the whole lifecycle — `CONNECT` provisions and polls until resource is ready, `DISCONNECT` reaps:
```
memory D CONNECT TO NEW EXTERNAL RESOURCE 'aws:cloudformation:quack-on-ec2' (REGION 'us-east-1');
2026-07-16 12:32:45.070663+00   ExternalResource        INFO    "{'resource_type': 'aws:cloudformation:quack-on-ec2', 'resource_name': NULL, 'operation': create, 'outcome': ok, 'error': NULL, 'extra_info': {}}"
2026-07-16 12:32:45.520402+00   ExternalResource        INFO    "{'resource_type': 'aws:cloudformation:quack-on-ec2', 'resource_name': NULL, 'operation': status, 'outcome': ok, 'error': NULL, 'extra_info': {state=pending}}"
2026-07-16 12:32:50.92983+00    ExternalResource        INFO    "{'resource_type': 'aws:cloudformation:quack-on-ec2', 'resource_name': NULL, 'operation': status, 'outcome': ok, 'error': NULL, 'extra_info': {state=pending}}"

2026-07-16 12:34:13.650125+00   ExternalResource        INFO    "{'resource_type': 'aws:cloudformation:quack-on-ec2', 'resource_name': NULL, 'operation': status, 'outcome': ok, 'error': NULL, 'extra_info': {state=pending}}"
2026-07-16 12:34:19.367423+00   ExternalResource        INFO    "{'resource_type': 'aws:cloudformation:quack-on-ec2', 'resource_name': NULL, 'operation': status, 'outcome': ok, 'error': NULL, 'extra_info': {state=pending}}"
2026-07-16 12:34:25.133674+00   ExternalResource        INFO    "{'resource_type': 'aws:cloudformation:quack-on-ec2', 'resource_name': NULL, 'operation': status, 'outcome': ok, 'error': NULL, 'extra_info': {state=pending}}"
2026-07-16 12:34:31.804265+00   ExternalResource        INFO    "{'resource_type': 'aws:cloudformation:quack-on-ec2', 'resource_name': NULL, 'operation': status, 'outcome': ok, 'error': NULL, 'extra_info': {state=ready}}"
quack:34.205.225.31.nip.io:443 memory D DISCONNECT;
2026-07-16 12:34:57.754806+00   ExternalResource        INFO    "{'resource_type': 'aws:cloudformation:quack-on-ec2', 'resource_name': NULL, 'operation': destroy, 'outcome': ok, 'error': NULL, 'extra_info': {}}"
memory D
```
`create` then `status` (polled to ready) happen under the `CONNECT`; `destroy` fires on `DISCONNECT` with no explicit teardown from the user.

### 2b. SHOW ALL EXTERNAL RESOURCES — reconciling with what exists out there

`SHOW EXTERNAL RESOURCES` lists what this instance manages. `SHOW ALL` additionally calls each resource type's `list` callback and unions in whatever exists externally but is not managed here - the view that makes orphans visible.

```sql
LOAD aws;
CREATE EXTERNAL RESOURCE 'aws:cloudformation:quack-on-ec2' AS louie (REGION 'us-east-1');
SHOW ALL EXTERNAL RESOURCES;
```
```
┌─────────┬─────────────────────────────────┬─────────┬─────────┬───────────┬─────────────────────────────────┬──────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│  name   │              type               │ managed │  state  │ reference │               uri               │ attached_db_type │                                                       handle                                                        │
│ varchar │             varchar             │ boolean │ varchar │  varchar  │             varchar             │     varchar      │                                                map(varchar, varchar)                                                │
├─────────┼─────────────────────────────────┼─────────┼─────────┼───────────┼─────────────────────────────────┼──────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ louie   │ aws:cloudformation:quack-on-ec2 │ true    │ NULL    │ NULL      │ quack:32.199.170.192.nip.io:443 │ quack            │ {stack_name=duckdb-quack-ec2-temp-336b58d5e1d3, stack_id='arn:aws:cloudformation:us-east-1:<IAM>:stack/duckd │
│         │                                 │         │         │           │                                 │                  │ b-quack-ec2-temp-336b58d5e1d3/db767010-8113-11f1-b4f0-0e13ccc4340b', region=us-east-1}                              │
└─────────┴─────────────────────────────────┴─────────┴─────────┴───────────┴─────────────────────────────────┴──────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

`louie` is managed here. Matching is by handle, so a listed resource we already manage is shown once, under its
local name, rather than twice.

Now close the terminal, and once again:
```sql
LOAD aws;
SHOW ALL EXTERNAL RESOURCES;
```
should show:
```
┌─────────┬─────────────────────────────────┬─────────┬─────────┬────────────────────────────────────────────────────────────────────────────┬─────────┬──────────────────┬────────────────────────────────────────────────────────────────────────────┐
│  name   │              type               │ managed │  state  │                                 reference                                  │   uri   │ attached_db_type │                                   handle                                   │
│ varchar │             varchar             │ boolean │ varchar │                                  varchar                                   │ varchar │     varchar      │                           map(varchar, varchar)                            │
├─────────┼─────────────────────────────────┼─────────┼─────────┼────────────────────────────────────────────────────────────────────────────┼─────────┼──────────────────┼────────────────────────────────────────────────────────────────────────────┤
│ NULL    │ aws:cloudformation:quack-on-ec2 │ false   │ ready   │ arn:aws:cloudformation:us-east-1:<IAM>:stack/duckdb-quack-ec2-temp- │ NULL    │ NULL             │ {stack_name=duckdb-quack-ec2-temp-336b58d5e1d3, stack_id='arn:aws:cloudfor │
│         │                                 │         │         │ 336b58d5e1d3/db767010-8113-11f1-b4f0-0e13ccc4340b                          │         │                  │ mation:us-east-1:<IAM>:stack/duckdb-quack-ec2-temp-336b58d5e1d3/db7 │
│         │                                 │         │         │                                                                            │         │                  │ 67010-8113-11f1-b4f0-0e13ccc4340b', region=us-east-1}                      │
└─────────┴─────────────────────────────────┴─────────┴─────────┴────────────────────────────────────────────────────────────────────────────┴─────────┴──────────────────┴────────────────────────────────────────────────────────────────────────────┘
```
Note the two tables are slightly different (no name in the second one) but carry the same information to be able to adopt, attach to or destroy the resource.

### 2c. Resources explicit lifetime: creation / adoption / borrow / destroy

An orphan is recoverable: `ADOPT` takes over its handle under a local name. Borrowing an existing
resource attaches to it without taking ownership.

```sql
SHOW ALL EXTERNAL RESOURCES;
SET VARIABLE handle_map = (SELECT handle FROM duckdb_external_resources(discover:=all) LIMIT 1);
--- syntax here to be improved, but `handle_map` is the representation of the resource

ADOPT EXTERNAL RESOURCE 'aws:cloudformation:quack-on-ec2' AS resource FROM getvariable('handle_map');
--- name a resource for local usage

ATTACH TO EXTERNAL RESOURCE resource AS quack_server;
-- borrows: no deleter bound, quack_server will be of the type dictated by the resource, in this case `quack`
CREATE TABLE quack_server.T AS FROM range(42);

CONNECT quack_server;
.tables 
--- table T should show
DISCONNECT;

DETACH quack_server;
--- ATTACH was not tied to lifetime of resource, so no tear down 
SHOW ALL EXTERNAL RESOURCES;
--- all resources, as above or possibly more

DESTROY EXTERNAL RESOURCE resource;
--- explicit DESTROY will tear down the resource
SHOW ALL EXTERNAL RESOURCES;
```

## 3. Ownership and the registry

The lifetime story is one invariant: the statement that provisions a resource is the one that reaps it. `NEW` binds a deleter onto the `AttachedDatabase`, and `DETACH`/`DISCONNECT` runs it. A borrowed resource binds no deleter, so detaching leaves it registered for a later `DESTROY`.

`CREATE`/`ADOPT` register an instance in an instance-scoped manager; `DESTROY` tears down and deregisters; `SHOW ALL EXTERNAL RESOURCES` / `duckdb_external_resources()` lists them unioning in resources discovered via each type's `list` callback - the reconciliation view that feeds `ADOPT`, so a resource that outlived its DuckDB process can be recovered rather than orphaned.

> Delete (that might trigger remote operations) will NOT be guaranteed to be always run. Recomendation is any/all of the following:
* have some form of idle or timeout based automatic close of the resource provider side
* have a mean of listing such resources (it's an opt-in capability of `external_resource_types`)
* native observability (e.g. `aws'`s cloudformation API)

Not load bearing, but pleasant for debugging, testing and explaining the mechanism, is a new logging type `ExternalResources`, that would allow after (for example) `CALL enable_logging('ExternalResources', storage=stdout)` to demonstrate progress.

Currently this works and has been designed for using `aws:cloudformation` tagged stacks to store state and making it work with the default provided template, but cloudformation templates are pluggable, allowing reuse and adaptation.
There is further work on improving the template AND generalizing to other templates, but this is oversized enough to be worth as is.

Thanks to many discussions that refined this idea, and to @samansmink in particular for the clean idea of the `WITH EXTERNAL RESOURCE`.


carlopi:
Rebased on the new top of the https://github.com/duckdb/duckdb/pull/23731 branch, still 2 extra commits adding the SQL layer for `WITH EXTERNAL RESOURCE`

This does miss the aws registration, so as per this PR it will work only via manual registration via SQL.

carlopi:
Also add relevant logging on each operation (create | status | destroy), that can double as status update if redirected to stdout.

src/parser/peg/grammar/statements/with_external_resource.gram
```diff
@@ -0,0 +1,3 @@
+WithExternalResourceStatement <- WithExternalResourceAttach / WithExternalResourceConnect
```
Dtenwolde: I think these rules have quite some overlap and can be simplified, so its easier to spot the difference between the two. Also there's two `AttachOptions` which seem confusing, so I'd rename one to `ExternalResourceOptions`, same with `ExternalResourceAlias`. Overall the grammar becomes a bit more verbose, but easier to understand in my opinion of what piece does what. 
```
WithExternalResourceStatement <- 'WITH' 'EXTERNAL' 'RESOURCE'
    Expression ExternalResourceAlias? ExternalResourceOptions?
    WithExternalResourceAction AttachOptions?

WithExternalResourceAction <-
    WithExternalResourceAttach / WithExternalResourceConnect

WithExternalResourceAttach <- 'ATTACH' AttachAlias?
WithExternalResourceConnect <- 'CONNECT'

ExternalResourceAlias <- AttachAlias
ExternalResourceOptions <- AttachOptions
```

Dtenwolde (commented):
Comment on the grammar :)

src/execution/operator/helper/launch_external_resource.cpp
```diff
@@ -0,0 +1,75 @@
...
+		throw IOException("WITH EXTERNAL RESOURCE: provisioning '%s' failed: %s", provider, res->GetError());
+	}
+	if (res->RowCount() == 0) {
+		throw IOException("WITH EXTERNAL RESOURCE: create_external_resource returned no rows for '%s'", provider);
+	}
+	LaunchedResource out;
+	auto uri_val = res->GetValue(0, 0);
+	if (uri_val.IsNull()) {
+		throw IOException("WITH EXTERNAL RESOURCE: provisioning '%s' returned a NULL uri", provider);
+	}
```
pdet, line 66: If the uri is missing, does the destructor of the resource gets called correctly? Maybe we can add a test with the logger?

e.g.,
```sql
CREATE MACRO qk_create(p)  AS TABLE SELECT MAP {'id':'q1'} AS handle;
CREATE MACRO qk_status(h)  AS TABLE SELECT 'ready' AS state, MAP {'foo':'bar'} AS result; 
CREATE MACRO qk_destroy(h) AS TABLE SELECT 'destroyed' AS status;
CALL register_external_resource_type('nc@local', kind := 'compute',
       create_function := 'qk_create', status_function := 'qk_status', destroy_function := 'qk_destroy');
WITH EXTERNAL RESOURCE 'nc@local' ATTACH AS z; 

```

src/planner/binder/statement/bind_attach.cpp
```diff
@@ -1,12 +1,32 @@
...
 #include "duckdb/planner/expression_binder/table_function_binder.hpp"
 #include "duckdb/execution/expression_executor.hpp"
 
 namespace duckdb {
 
+void Binder::BindExternalResource(ExternalResourceOptions &external_resource) {
+	TableFunctionBinder binder(*this, context, "External resource", "External resource parameter");
+	// Resolve the resource type expression (the '<type>' / provider).
+	auto bound_type = binder.Bind(external_resource.parsed_type);
+	auto type_val = ExpressionExecutor::EvaluateScalar(context, *bound_type);
```
pdet: What if is a volatile expression? e.g.,
```
CREATE SEQUENCE s;                                                                                                                                                                                                                                                                                                                                                  
WITH EXTERNAL RESOURCE nextval('s') ATTACH AS z;          
```

src/main/database_manager.cpp
```diff
@@ -249,9 +249,15 @@ optional_ptr<AttachedDatabase> DatabaseManager::FinalizeAttach(ClientContext &co
...
 		}
+		auto deleter = detached_db->ExtractDeleter();
 		meta_transaction.DetachDatabase(*detached_db);
 		detached_db->OnDetach(context);
 		detached_db.reset();
+		// Best-effort teardown: the replacement is already in effect, so a deleter failure must not
+		// abort the statement mid-swap; TryDelete logs it as a warning instead.
+		if (deleter) {
+			deleter->TryDelete();
+		}
```
pdet, line 256: Is this transactionally correct?
what happens if

```sql
WITH EXTERNAL RESOURCE 't@local' ATTACH AS db;
BEGIN;
ATTACH OR REPLACE '/tmp/other.db' AS db;
ROLLBACK; 
```

src/execution/operator/helper/physical_connect.cpp
```diff
@@ -19,21 +21,74 @@ SourceResultType PhysicalConnect::GetDataInternal(ExecutionContext &context, Dat
...
+	auto ensure_not_connected = [&]() {
+		if (client.IsConnected()) {
+			auto current = client.TryGetConnectedCatalog();
+			throw InvalidInputException("Already connected to \"%s\"; DISCONNECT first before issuing another CONNECT",
+			                            current ? current->GetName().GetIdentifierName() : "<detached>");
+		}
+	};
+
+	// `WITH EXTERNAL RESOURCE ... CONNECT`: provision the resource, then ephemeral-attach its endpoint
+	// under a hidden alias and connect to it. The deleter is bound so the DISCONNECT reap tears it down.
```
pdet, line 40: What happens if we close a session without DISCONNECT?

src/function/table/system/create_external_resource.cpp
```diff
@@ -180,6 +204,8 @@ static void CreateExternalResourceFunction(ClientContext &context, TableFunction
 	auto sql = "SELECT * FROM " + QualifiedName::Parse(type->create_function).ToString() + "(" +
 	           bind_data.params.ToSQLString() + ")";
 	auto result = con.Query(sql);
+	LogExternalResourceOperation(context, bind_data.type_name, bind_data.resource_name, "create", sql,
```
pdet: Can this leak secrets to the logger?
e,g.,

```sql
CALL enable_logging('ExternalResource', storage='memory');
WITH EXTERNAL RESOURCE 't@local' (api_token 'SECRET') ATTACH AS s;
SELECT sql FROM duckdb_logs_parsed('ExternalResource') WHERE operation='create';
```

src/execution/operator/schema/physical_attach.cpp
pdet: For external resource attach without a database name I think we would fail here, maybe we can fail at bind with a nice error message?

src/main/database_manager.cpp
pdet: Also what happens with a failed-commit rollback? can we add tests for that?

src/execution/operator/helper/launch_external_resource.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+void ApplyLaunchedResource(const LaunchedResource &launched, AttachInfo &info) {
+	info.path = launched.uri;
+	// A NULL/empty db type falls back to the extension prefix extracted from the uri.
+	if (!launched.attached_db_type.empty()) {
+		info.options["type"] = Value(launched.attached_db_type);
+	}
+	// The remaining connect options (e.g. token) flow through as attach options.
+	if (!launched.result.IsNull()) {
```
pdet, line 94: Can this accidentally leak options to the attach? It seems that every non-uri/attached_db_type key becomes an attach option?

e.g.,

```sql
CREATE MACRO qk_status(h) AS TABLE SELECT 'ready' AS state,
 MAP {'uri':':memory:','attached_db_type':'duckdb','type':'sqlite'} AS result;
  WITH EXTERNAL RESOURCE 'o@local' ATTACH AS o1;  
```

src/parser/parsed_data/attach_info.cpp
```diff
@@ -16,11 +16,38 @@ unique_ptr<AttachInfo> AttachInfo::Copy() const {
...
 string AttachInfo::ToString() const {
 	string result = "";
+	// `WITH EXTERNAL RESOURCE '<type>' [AS r] [(params)] ATTACH [AS name] [(opts)]`
+	if (external_resource) {
+		result += "WITH " + external_resource->ToString() + " ATTACH";
+		if (on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
+			result += " IF NOT EXISTS";
+		} else if (on_conflict == OnCreateConflict::REPLACE_ON_CONFLICT) {
+			result += " OR REPLACE";
+		}
```
pdet: Maybe a dumb question but us `ATTACH OR REPLACE` valid (as in accepted in the grammar layer) to begin with for the external resource?

pdet (commented):
Thanks for the PR, a few comments.

src/function/table/system/create_external_resource.cpp
carlopi: Thanks! removed the 'sql' column from logging, it would need to be redacted before this is safe adding back.

src/execution/operator/helper/physical_connect.cpp
carlopi: DISCONNECT is automatic, and trigger DETACH and then resource deletion in all cases but for crashes.

And this needs in any case to have a safety net where one would remove instances past a given [idle] timeout.

src/planner/binder/statement/bind_attach.cpp
carlopi: Great catch, moved to constant string only for now.

src/execution/operator/helper/launch_external_resource.cpp
carlopi: +1, added call to deleter + test using logger.

src/parser/peg/grammar/statements/with_external_resource.gram
carlopi: Thanks! Reworked

src/main/database_manager.cpp
carlopi: I am adding tests, this is currently a sharp edge also in duckdb, that:
```
ATTACH 'file.db' AS db;
BEGIN;
ATTACH OR REPLACE 'file2.db' AS db;
ROLLBACK; 
FROM duckdb_databases();
```
shows no `db`. Here it's sort of behaving the same: resources is created at line 1, destroyed at line 3, and will never come back.

src/execution/operator/helper/launch_external_resource.cpp
carlopi: Thanks! Added a test, via logger indeed, that no URI + provision path will destroy resource right there, and in ADOPT path it would leave it there.

src/parser/peg/grammar/statements/with_external_resource.gram
Dtenwolde: Is it a potential follow up to allow `CONNECT EXECUTE` here?

src/execution/operator/helper/launch_external_resource.cpp
carlopi: This is on purpose, anything MUST become an attach option, I am checking whether a test is there to certify the expected behaviour.

src/parser/parsed_data/attach_info.cpp
carlopi: I think so, but let me double check via tests.

src/execution/operator/helper/launch_external_resource.cpp
carlopi: Checked, this will then (in the ephemeral case) also destroy the resource.

carlopi:
For reviewers, I would expect the following to work in allowing to reproduce the above examples:
```
git clone https://github.com/duckdb/duckdb
cd duckdb
gh pr checkout 23736 --repo duckdb/duckdb
make setup-vcpkg
USE_MERGED_VCPKG_MANIFEST=1 VCPKG_TOOLCHAIN_PATH=$(PWD)/vcpkg/scripts/buildsystems/vcpkg.cmake CORE_EXTENSIONS='quack;httpfs;autocomplete;json;parquet;icu;aws' make
```
and then launching via:
```
./build/release/duckdb
```

carlopi:
Adapting to out of band comments, some changes:

1. `REGISTER` to be the verb to register an internal name for a given resource
2. `CONNECT TO NEW TEMPORARY EXTERNAL RESOURCE` for making explicit an ephemeral resource is created
3. `FROM duckdb_external_resources(discover:=true)` to be the table function, so to be able to filter and all

There is not yet a single statement way to go from an handle to using (or deleting an handle). Currently flow would be:
```sql
--- get the handle
SET VARIABLE machine_handle = (SELECT handle FROM duckdb_external_resources(discover := true) LIMIT 1);

---register to a name
REGISTER EXTERNAL RESOURCE '<type>' AS resource_name FROM $machine_handle;

--- use, for example, delete:
DESTROY EXTERNAL RESOURCE resource_name;
```
given registration an use are in most cases tied 1 to 1, it would be great to fuse them into a single:
```
DESTROY EXTERNAL RESOURCE $machine_handle;
--- or
DESTROY EXTERNAL RESOURCE FROM $machine_handle;
```
This is not part of the PR.

Thanks!!

src/execution/operator/schema/physical_attach.cpp
```diff
@@ -15,14 +19,51 @@ namespace duckdb {
 //===--------------------------------------------------------------------===//
 SourceResultType PhysicalAttach::GetDataInternal(ExecutionContext &context, DataChunk &chunk,
                                                  OperatorSourceInput &input) const {
-	// parse the options
 	auto &config = DBConfig::GetConfig(context.client);
+
+	// Attach mutates the info (name resolution, path prefix stripping, injected provision options), so
+	// work on a copy: the plan-owned info must stay pristine for a re-executed prepared statement.
+	auto attach_info = info->Copy();
```
carlopi: This is an independent bug-fix of sort, since re-executed plans where the Attach would mutate AttachInfo (something an extension could reasonably do) would be invalid / leak information.

This is visible only if modifying AttachInfo (done in this case to carry lifetime of external_resources) couple with debug settings.

src/execution/operator/helper/launch_external_resource.cpp
```diff
@@ -84,6 +99,7 @@ void ApplyLaunchedResource(const LaunchedResource &launched, AttachInfo &info) {
 			if (key == "uri" || key == "attached_db_type") {
 				continue;
 			}
+			RejectOptionCollision(info, key);
 			info.options[key] = kv[1];
```
pdet, line 109: Is Null rejected here? Should we?

src/execution/operator/schema/physical_attach.cpp
```diff
@@ -15,14 +19,66 @@ namespace duckdb {
...
+				                launched.deleter_payload, resource_type, resource_name)
+				    .TryDelete();
+			}
+			throw;
+		}
+	}
+
 	// construct the options
-	AttachOptions options(info->options, config.options.access_mode);
+	AttachOptions options(attach_info->options, config.options.access_mode);
```
pdet: Attach options can throw, is that a problem if we throw here? would teardown work fine?

src/transaction/rollback_state.cpp
```diff
@@ -48,7 +49,18 @@ void RollbackState::RollbackEntry(UndoFlags type, data_ptr_t data) {
...
 		auto db = Load<AttachedDatabase *>(data);
 		auto &db_manager = DatabaseManager::Get(db->GetDatabase());
-		db_manager.DetachInternal(db->name);
+		auto attached_db = db_manager.DetachInternal(db->name);
+		if (attached_db) {
+			// The attachment may own an external resource (ATTACH/CONNECT TO EXTERNAL RESOURCE). Its teardown runs
+			// SQL, which is impossible here: rollback executes under the transaction lock, and the
+			// teardown query would need to start a transaction on the same manager (self-deadlock) —
+			// nor may rollback fail. So extract the deleter (a plain field move, safe under the lock)
+			// and queue it; the DatabaseManager drains the queue best-effort after rollback completes.
```
pdet: nit: can we minimize comments to only absolutely necessary and max 1-2 lines?

test/sql/attach/attach_to_external_resource.test
```diff
@@ -0,0 +1,276 @@
+# name: test/sql/attach/attach_to_external_resource.test
```
pdet, line 1: All the tests are in the provision form? Should we add tests for the borrow branch?

src/parser/parsed_data/attach_info.cpp
```diff
@@ -18,11 +18,37 @@ unique_ptr<AttachInfo> AttachInfo::Copy() const {
...
 	string result = "";
+	// `ATTACH TO [NEW TEMPORARY] EXTERNAL RESOURCE <resource> [(create opts)] AS name [(attach opts)]`
+	if (external_resource) {
+		// No IF NOT EXISTS / OR REPLACE: AttachToExternalResource has no slot for either, so rendering
+		// one would produce SQL that cannot be parsed back.
+		result += "ATTACH TO " + external_resource->ToString();
+		if (!name.empty()) {
+			result += " AS " + SQLIdentifier(name);
+		}
+		if (!parsed_options.empty() || !options.empty()) {
```
pdet: Is this the same of lines 67-78? If so can we reuse?

src/execution/operator/helper/physical_connect.cpp
```diff
@@ -19,21 +22,103 @@ SourceResultType PhysicalConnect::GetDataInternal(ExecutionContext &context, Dat
...
+
+	// `CONNECT TO [NEW] EXTERNAL RESOURCE`: provision the resource, then ephemeral-attach its endpoint
+	// under a hidden alias and connect to it. The deleter is bound so the DISCONNECT reap tears it down.
+	if (info->external_resource) {
+		ensure_not_connected();
+		auto &external_resource = *info->external_resource;
+		// A string type PROVISIONS a resource this connection OWNS (deleter bound, DISCONNECT reaps it);
+		// a bare identifier REFERENCES a registered resource it only BORROWS (no deleter).
+		LaunchedResource launched;
+		string resource_type, resource_name, borrowed_resource_name;
```
pdet: Is resource_name ever assigned?

pdet (commented):
Thanks, is looking great, a few more comments.

src/transaction/rollback_state.cpp
pdet: Maybe also do a pass (or ask the clanker) to see if the comments are still valid, I've noticed some inconsistencies.

test/sql/attach/attach_to_external_resource.test
carlopi: I think tests for the provision form are easier to actually run than borrow+`CONNECT` branch that will need at least quack instance.

Borrow + `ATTACH` should be already tested.

src/parser/parsed_data/attach_info.cpp
carlopi: Great, thanks!

src/include/duckdb/parser/parsed_data/attach_info.hpp
```diff
@@ -37,6 +38,8 @@ struct AttachInfo : public ParseInfo {
 	unordered_map<string, Value> options;
 	//! What to do on create conflict
 	OnCreateConflict on_conflict = OnCreateConflict::ERROR_ON_CONFLICT;
+	//! Set iff parsed as `ATTACH TO [NEW TEMPORARY] EXTERNAL RESOURCE ...`: the resource to attach.
```
Dtenwolde, line 41: Should we introduce a `ExternalResourceAttachInfo` struct. Similar to how `CreateInfo` has several overloads? I'm not sure if we are going to add more `Attach` variations, but then it might be cleaner

src/include/duckdb/parser/parsed_data/connect_info.hpp
```diff
@@ -41,6 +42,8 @@ struct ConnectInfo : public ParseInfo {
 	case_insensitive_map_t<unique_ptr<ParsedExpression>> parsed_options;
 	//! Set of bound (key, value) options forwarded to the implicit ATTACH.
 	unordered_map<string, Value> options;
+	//! Set iff parsed as `CONNECT TO [NEW TEMPORARY] EXTERNAL RESOURCE ...`: the resource to connect to.
+	unique_ptr<ExternalResourceOptions> external_resource;
```
Dtenwolde, line 46: Same here I suppose

src/include/duckdb/parser/parsed_data/external_resource_options.hpp
```diff
@@ -0,0 +1,43 @@
...
+//! The `[NEW TEMPORARY] EXTERNAL RESOURCE '<type>' [(params)]` clause, owned by AttachInfo/ConnectInfo.
+//! provider is its own field rather than a magic key in params: the TYPE is the registry key, not a param.
+struct ExternalResourceOptions {
+	//! Create params (key -> expression); transient — consumed at bind, then `params` holds them.
+	case_insensitive_map_t<unique_ptr<ParsedExpression>> parsed_params;
+	//! The resource type (provider) — the registered external-resource-type name to provision. A string
+	//! literal, so it is fixed at parse time.
+	string provider;
+	//! Bound create params forwarded to the type's create function as a MAP(VARCHAR, VARCHAR).
+	unordered_map<string, Value> params;
```
Dtenwolde, line 30: Should these strings be case insensitive/`Identifier`?

Dtenwolde: I see some more `string` related variables in this struct that could be potential `Identifier`

src/include/duckdb/parser/peg/inlined_grammar.hpp
```diff
@@ -1433,6 +1435,16 @@ const char INLINED_PEG_GRAMMAR[] = {
...
 	"ExternalResourceCreationOptions <- GenericCopyOptionList\n"
+	"# The attach alias is mandatory: it separates the recipe's create params, which precede it, from the\n"
+	"# attachment's own options, which follow. Without it a single option list would be ambiguous.\n"
+	"# CONNECT has no such separator. Borrowing a registered resource has no create params to be confused\n"
+	"# with, so a list there is the connection's; provisioning does, and the transformer rejects the second\n"
+	"# list rather than let position decide.\n"
+	"AttachToExternalResource <- 'ATTACH' 'TO' ExternalResourceSource AttachAlias AttachOptions?\n"
+	"ConnectToExternalResource <- 'CONNECT' 'TO' ExternalResourceSource AttachOptions?\n"
+	"ExternalResourceSource <- ExternalResourceCreateClause / ExternalResourceReferenceClause\n"
+	"ExternalResourceCreateClause <- 'NEW' 'TEMPORARY' 'EXTERNAL' 'RESOURCE' StringLiteral ExternalResourceCreationOptions?\n"
```
Dtenwolde, line 1446: Can we make a rule just for `'EXTERNAL' 'RESOURCE'` so we don't keep repeating ourselves? :)

Dtenwolde (commented):
Couple comments, will continue tomorrow :)

src/execution/operator/schema/physical_attach.cpp
```diff
@@ -15,27 +19,85 @@ namespace duckdb {
...
+		}
+		if (name.empty()) {
+			auto &fs = FileSystem::GetFileSystem(context.client);
+			name = Identifier(AttachedDatabase::ExtractDatabaseName(path, fs));
+		}
+
+		// check ATTACH IF NOT EXISTS
+		auto &db_manager = DatabaseManager::Get(context.client);
+		db_manager.AttachDatabase(context.client, *attach_info, options);
+	} catch (...) {
```
Dtenwolde, line 95: I think we should avoid these generic `catch` blocks

src/execution/operator/schema/physical_attach.cpp
```diff
@@ -15,27 +19,85 @@ namespace duckdb {
...
+		// Compensating teardown (best-effort): a resource we just PROVISIONED has nothing owning it yet.
+		// A borrowed (referenced) resource is never torn down here.
+		reap_if_owned();
```
Dtenwolde, line 98: Why is this a lambda? I think it's only used here. I think this can be cleaner :)

Dtenwolde: Oh it's also in `phyical_connect.cpp` I see, but there it's slightly different. In `physical_connect` it's called with an empty string for resource name

src/execution/operator/schema/physical_attach.cpp
```diff
@@ -15,27 +19,85 @@ namespace duckdb {
...
+		auto &external_resource = *attach_info->external_resource;
+		if (!external_resource.reference_name.empty()) {
+			auto instance = ExternalResourcesManager::Get(context.client).Lookup(external_resource.reference_name);
+			if (!instance) {
+				throw InvalidInputException("no external resource named \"%s\" is registered",
+				                            external_resource.reference_name);
+			}
+			// Resolve the current endpoint from the registered handle (status); bind no deleter. Record the
+			// borrow so DESTROY refuses while this attachment is still pointing at the resource.
+			launched = ProvisionExternalResource(context.client, instance->type, {}, instance->name, instance->handle);
```
Dtenwolde, line 43: Why does this branch not pass the `external_resource.params` or `instance.params`? Does it not have that?

src/parser/peg/grammar/statements/common.gram
```diff
@@ -4,6 +4,8 @@ TopLevelStatement <- Statement? (';'+ / EndOfInput)
 
 Statement <-
 	ExternalResourceStatement /
+	AttachToExternalResource /
```
Dtenwolde, line 7: I think it's also good to start grouping these not under `Statement` but under either 
```
ExternalResource <- ExternalResourceStatement / AttachExternalResource / ConnectExternalResource
```

Then the top level statement doesn't get flooded with options.

src/parser/peg/transformer/transform_generic_copy_option.cpp
```diff
@@ -179,4 +179,28 @@ void PEGTransformerFactory::SplitGenericOptions(const vector<GenericCopyOption>
 	}
 }
 
+void PEGTransformerFactory::CollectGenericOptions(const vector<GenericCopyOption> &options_in,
```
Dtenwolde, line 182: This contains some duplicate code with `PEGTransformerFactory::SplitGenericOptions`. Could be reworked to share some of the logic

Dtenwolde:
Codex also found the following issue: 
The existing test only checks the exact read_only spelling returned by
ro_status. AttachOptions also interprets readonly, readwrite, and read_write as
the same underlying access_mode setting, but the collision check compares
textual keys.

Could we add regression coverage for both the spelling alias and the wider
semantic collision? The first new case currently succeeds unexpectedly,
confirming the bypass.

```
diff --git a/test/sql/attach/external_resource_option_collision.test b/test/sql/attach/external_resource_option_collision.test
index 725b08b01e..27e478a4d6 100644
--- a/test/sql/attach/external_resource_option_collision.test
+++ b/test/sql/attach/external_resource_option_collision.test
@@ -47,6 +47,18 @@ ATTACH TO EXTERNAL RESOURCE pinned_ro AS b1 (READ_ONLY 1);
 ----
 <REGEX>:.*the resource supplies "read_only".*

+# ro_status supplies read_only=true; READONLY is an alternate spelling of that
+same option.
+statement error
+ATTACH TO EXTERNAL RESOURCE pinned_ro AS b1 (READONLY 0);
+----
+<REGEX>:.*the resource supplies "read_only".*
+
+# ro_status supplies read_only=true; READWRITE also controls access_mode, so
+the statement cannot supply it too.
+statement error
+ATTACH TO EXTERNAL RESOURCE pinned_ro AS b1 (READWRITE 1);
+----
+<REGEX>:.*the resource supplies "read_only".*
+
 # ... and the same for the db type the resource dictates.
 statement error
 ATTACH TO EXTERNAL RESOURCE pinned_ro AS b1 (TYPE 'duckdb');
```

Dtenwolde (commented):
Some comments :)

carlopi:
Note, the diff is gigantic, but mostly at:
```
git diff origin/v2.0-cyanoptera  --stat src/include/duckdb/parser/peg/inlined_grammar.hpp
     1 file changed, 966 insertions(+), 905 deletions(-)

git diff origin/v2.0-cyanoptera  --stat test/
     14 files changed, 1269 insertions(+), 3 deletions(-)

git diff origin/v2.0-cyanoptera  --stat src/parser
     13 files changed, 425 insertions(+), 51 deletions(-)
```

I addressed most @Dtenwolde's comments, and I think this is now ready.

Dtenwolde (approved):

PR #23737, opened 2026-07-09
Fix crash when appending to indexed table with shredded columns

Maxxen:
This is an alternative to https://github.com/duckdb/duckdb/pull/23719

I still think the other PR is a more elegant solution, but it does not work with `VARIANT` columns, and the expectations they have around the state of a row group. To recap:

- Shredded `GEOMETRY` segments can not be appended to. Appended data lands unshredded in a new row group segment which only gets shredded back on checkpoint. In this sense, geometry shredding is "all or nothing".
- `VARIANT` columns do not generally have this limitation. They store both shredded and "unshredded" data simultaneously in the same row group. They only disallow appends when _fully shredded_. (and therefore in this case also runs into the same issue as geometry columns). But they still _need_ to be able to support appends in the mixed case, we can't just always create new segments or we mess up e.g. the stats when there are multiple mixed variant row groups.
- The `RowGroupCollection` sets `RowGroupAppendMode::REQUIRE_NEW` for tables containing geometry or variant, but `AppendRowGroup` unconditionally resets the mode to `APPEND_TO_EXISTING` the after a new rowgroup is created. That downgrade is correct for the original purpose which was to signal that "the tail rowgroup was just flushed to disk mid-transaction, don't touch it once" to the `OptimisticDataWriter`, or every append splits into separate row groups, for all types. (this is what messes up the variant stats in #23719, vacuum/merging these splintered groups).
- Checkpoint/load/merge storage resets this flag to an advisory `SUGGEST_NEW`, but indexed tables always ignore this suggestion. Because we can't yet vacuum and reorder indexed row groups (row ids need to be stable), we really want to avoid fragmenting if possible, so indexed tables always append to the last row group (unless its full).

So the full failure path is:

1. `INSERT` into an empty table, `LocalStorage::Flush` takes the `MergeStorage` path instead of `AppendRowGroup` and the transaction local rowgroup (1) is moved entirely into the collection - the `RowGroupAppendMode::REQUIRE_NEW` set in the constructor stays armed (as we never got into AppendRowGroup)
2. `CHECKPOINT` shreds the row group, the append mode is set to `RowGroupAppendMode::SUGGEST_NEW`, but we dont downgrade, so it stays `REQUIRE_NEW`.
3. `INSERT` again, this time the table is not empty so we go through `AppendToTable`. `REQUIRE_NEW` forces a fresh row group (2) but after calling `AppendRowGroup` the append mode is changed to `APPEND_TO_EXISTING`.
4. `CHECKPOINT` again, vacuum-merge row groups (1)+(2), re-shred, and set the append mode to `SUGGEST_NEW`.
5. `INSERT` for the third time. Now `SUGGEST_NEW` would have saved us, but because the table has an index (primary key), this suggestion is ignored and we append to the tail row group, which is shredded, but our input is not -> crash!

Essentially, the `RowGroupAppendMode` is used transiently to "debounce" row-group creation, but can _not_ be relied on to uphold the geometry/variant column data persistence invariants.

This PR works around this by letting the `ColumnData` itself (who has a much better idea of the current state of the segment and the invariants that need to hold) to tell the row group wether it can be appended to or not. In the case of `GEOMETRY` and `VARIANT`, this is of course dependent on if it's shredded or not.

### DIFF
--- a/src/include/duckdb/storage/table/array_column_data.hpp
+++ b/src/include/duckdb/storage/table/array_column_data.hpp
@@ -56,6 +56,7 @@ public:
 	                                             const BaseStatistics &old_stats) override;
 
 	bool IsPersistent() override;
+	bool IsAppendable() const override;
 	bool HasAnyChanges() const override;
 	PersistentColumnData Serialize() override;
 	void InitializeColumn(PersistentColumnData &column_data, BaseStatistics &target_stats) override;
--- a/src/include/duckdb/storage/table/column_data.hpp
+++ b/src/include/duckdb/storage/table/column_data.hpp
@@ -187,6 +187,8 @@ public:
 	virtual void CheckpointScan(ColumnSegment &segment, ColumnScanState &state, idx_t count, Vector &scan_vector) const;
 
 	virtual bool IsPersistent();
+	//! Whether new rows can be appended to this column, e.g. shredded GEOMETRY columns no longer accept appends
+	virtual bool IsAppendable() const;
 	vector<DataPointer> GetDataPointers();
 
 	virtual PersistentColumnData Serialize();
--- a/src/include/duckdb/storage/table/geo_column_data.hpp
+++ b/src/include/duckdb/storage/table/geo_column_data.hpp
@@ -60,6 +60,7 @@ public:
 	                                             const BaseStatistics &old_stats) override;
 
 	bool IsPersistent() override;
+	bool IsAppendable() const override;
 	bool HasAnyChanges() const override;
 	PersistentColumnData Serialize() override;
 	void InitializeColumn(PersistentColumnData &column_data, BaseStatistics &target_stats) override;
--- a/src/include/duckdb/storage/table/list_column_data.hpp
+++ b/src/include/duckdb/storage/table/list_column_data.hpp
@@ -54,6 +54,7 @@ public:
 	                                             const BaseStatistics &old_stats) override;
 
 	bool IsPersistent() override;
+	bool IsAppendable() const override;
 	bool HasAnyChanges() const override;
 	PersistentColumnData Serialize() override;
 	void InitializeColumn(PersistentColumnData &column_data, BaseStatistics &target_stats) override;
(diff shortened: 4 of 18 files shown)


Maxxen:
> But they still need to be able to support appends in the mixed case, we can't just always create new segments or we mess up e.g. the stats when there are multiple mixed variant row groups.

Note Im not exactly sure why just always creating new row groups don't work for variants. Even if it's inefficient, it should still be correct, and so there might be another unrelated issue here that may need investigation.

Maxxen:
Did some more digging, the variant code does seem correct as is, it actually worked before too with just creating small row groups, it just was a slight behavior change, but the result was essentially still correct.

Mytherin:
> INSERT for the third time. Now SUGGEST_NEW would have saved us, but because the table has an index (primary key), this suggestion is ignored and we append to the tail row group, which is shredded, but our input is not -> crash!

Isn't the issue then that we're setting `SUGGEST_NEW` when we should be setting `REQUIRE_NEW`? We shouldn't be setting `SUGGEST_NEW` if it's not a suggestion, no?

Maxxen:
@Mytherin Thats what I did in https://github.com/duckdb/duckdb/pull/23719, but that made variants always push new row groups when appending (which is fine, although a slight performance hit I guess), but as the failing tests in the CI of that PR shows it also made a slight change to the resulting shredding/stats (and also hit an internal exception in a "should not be reached" code path in a separate variant test).

I don't feel like I have the context, authority or knowledge to judge wether the impact #23719 has on `VARIANT` columns is acceptable (or how to fix the internal exception, if its even a problem). Given that I will be OoO the coming week right before v1.5.5 I felt like this PR, while slightly larger, may be the safer option as it keeps the same behavior for `VARIANT` while also fixing the crash/corruption that both `GEOMETRY` and `VARIANT` are currently susceptible to. If there's a better way to solve this then somebody else has to pick it up without me.

Mytherin:
I can pick this up - I think https://github.com/duckdb/duckdb/pull/23763 is the right fix

Mytherin:
Fixed in https://github.com/duckdb/duckdb/pull/23763

PR #23739, opened 2026-07-09
Add boundary and decimal test cases for factorial

hacoobamatata:
Add test coverage for boundary values (33 and 34) and decimal inputs (1.5 and 3.0) for factorial.


Mytherin:
Please refrain from opening PRs that only add new tests

PR #23750, opened 2026-07-10
Add streaming fanout execution for materialized CTEs

kryonix:
This PR adds streaming fanout execution for materialized CTEs.

Previously, a materialized CTE was always a full pipeline breaker. DuckDB first wrote the complete CTE result into a `ColumnDataCollection`; only after that collection was finalized could any CTE consumer start. This PR preserves the logical semantics of a materialized CTE, but allows its physical producer to feed eligible consumer pipelines while the CTE body is still running.

The core addition is a general pipeline-level broadcast exchange. For its first use, a physical materialized CTE can now serve consumers in three ways:

1. Push producer chunks directly into externally-fed consumer pipelines.
2. Let consumers scan chunks retained by the exchange.
3. Materialize the result into the existing `ColumnDataCollection` and scan it after completion.

A single CTE can combine these strategies. The CTE body is still executed once, every consumer observes the same CTE result, and CTEs with side effects still run to completion. The change is entirely in physical planning and execution; it adds no SQL syntax and does not change the meaning of `MATERIALIZED` or `NOT MATERIALIZED`.

## Motivation

Full materialization introduces two costs even when all consumers could process the rows immediately:

- The producer and consumers cannot overlap.
- Every row must be written to and read from intermediate storage.

This is especially relevant for rewrites that deliberately introduce a shared CTE. A grouped distinct aggregate rewrite, for example, can scan an input once and fan it out into multiple independent hash aggregations. Without streaming fanout, the rewrite first materializes the entire input and then scans it once per aggregate. With direct fanout, the input scan drives all aggregate pipelines in parallel without exchange chunk storage.

Streaming can also avoid work. If all relevant consumers stop early, such as through a streaming limit, a side-effect-free CTE producer can stop before exhausting its source.

## Physical CTE Modes

The final strategy is pipeline-dependent. Consumer pipelines are classified while pipelines are built, after logical optimization and physical planning. This is intentional: eligibility depends on the complete physical consumer pipeline, not on properties attached to the CTE when it was parsed or created.

This also means the feature applies to materialized CTEs introduced or retained by optimization. A CTE does not need to appear explicitly in the original SQL. Conversely, a CTE that was inlined before physical planning has no physical CTE operator and therefore does not use the exchange.

### `PIPELINE_DEPENDENT`

A regular `EXPLAIN` can display `Execution Mode: PIPELINE_DEPENDENT`. At that point the physical operator supports exchange execution, but pipeline construction has not yet classified every consumer. The final mode cannot be reported truthfully yet. `EXPLAIN ANALYZE` builds and executes the pipelines, so its profile reports one of the resolved modes below.

### `MATERIALIZED`

No consumer uses the exchange. The producer writes the complete result into the CTE `ColumnDataCollection`, and consumers scan it after producer completion. This is the fallback when exchange execution is unavailable. It is also selected deliberately when materialization is cheaper than streaming, notably for multi-consumer fanout where every consumer needs the complete result and no consumer pipeline is eligible for direct input.

### `STREAMING_FANOUT`

Every CTE consumer is served by the exchange. Consumers can process rows while the CTE body is still producing them. The fanout can be direct, buffered, or a mixture of both. The materialized working table is not populated in this mode.

### `HYBRID_FANOUT`

At least one consumer uses the exchange and at least one consumer scans the materialized working table. The producer both feeds exchange consumers and appends its output to the `ColumnDataCollection` required by materialized consumers. This allows streaming where it is useful without making it a requirement for every consumer. For example, a direct aggregate consumer and an early-stopping consumer can run through the exchange while a consumer with an unsupported pipeline shape uses the normal materialized scan.

## Consumer Modes

Each CTE scan is registered with the exchange and resolved to exactly one consumer mode during pipeline construction. The exchange is the single source of truth for these assignments and their counts, including across prepared statement pipeline rebuilds.

### `DIRECT`

The consumer's source pipeline is converted into an externally-fed pipeline. The CTE producer creates a `PipelineExecutor` for that pipeline and calls `PushExternal` for every producer chunk. Finalization is driven through `FinishExternal` after producer input ends.

Direct fanout avoids copying and intermediate chunk storage. It is only selected when the complete consumer pipeline supports external input:

- The pipeline has a parallel, order-independent sink.
- The sink explicitly supports external input.
- The sink does not require partitioned input.
- Every intermediate operator explicitly supports external input.
- Every intermediate operator is parallel and does not require fixed order.

These checks are properties of the physical pipeline. There are no query-name, CTE-name, or optimizer-rewrite-specific exceptions.

### `BUFFERED`

The CTE scan remains a source, but reads from the exchange instead of waiting for the materialized table. The producer copies chunks into exchange-owned storage and wakes blocked readers. Chunks are retired after every active buffered consumer has advanced past them.

A single exchange consumer uses buffered execution when direct input is unavailable. With multiple consumers, buffered execution is kept only for a pipeline that may stop before source exhaustion. Multi-consumer pipelines that require all rows use materialization unless they qualify for direct input. This avoids paying broadcast-buffering overhead when materialized scans are the simpler and faster execution strategy.

### `MATERIALIZED`

The consumer does not participate in exchange flow control. It scans the existing CTE `ColumnDataCollection` after the producer has completed. In hybrid mode, these consumers coexist with direct and buffered exchange consumers.

## Fanout Fields In `EXPLAIN ANALYZE`

For resolved streaming modes, the CTE operator reports:

- `Fanout Mode: DIRECT`: every exchange consumer is direct.
- `Fanout Mode: BUFFERED`: every exchange consumer reads exchange storage.
- `Fanout Mode: MIXED`: both direct and buffered exchange consumers exist.
- `Consumers`: the number of direct plus buffered consumers.
- `Direct Consumers`: externally-fed pipelines.
- `Buffered Consumers`: exchange scan sources.
- `Materialized Consumers`: consumers scanning the CTE working table.
- `Chunk Storage: NONE`: no buffered consumer exists, so producer chunks are not copied by the exchange.
- `Chunk Storage: POOLED/COLUMN_DATA`: buffered consumers start with pooled in-memory chunks and can transition to shared `ColumnDataCollection` spool storage.
- `Run To Completion: true`: producer execution cannot stop when exchange consumers detach or finish early.

Example:

```text
Execution Mode: HYBRID_FANOUT
Fanout Mode: DIRECT
Consumers: 1
Direct Consumers: 1
Buffered Consumers: 0
Chunk Storage: NONE
Materialized Consumers: 1
```

Individual CTE scan profiles report their resolved `CTE Mode` and consumer index.

## Broadcast Exchange

`PipelineBroadcastExchange` coordinates registration, producer lifecycle, consumer lifecycle, backpressure, cancellation, progress, and transition logging. Direct pipeline execution is owned by the per-thread exchange local state. Buffered storage and its positional bookkeeping are isolated in `BufferState`. The exchange remains the owner of the existing annotated mutex, so the separation does not introduce another lock or a new lock ordering.

### Producer Path

For each input chunk, a producer local state can perform the following work:

1. Resume or push the chunk through every direct consumer executor.
2. Reserve an append for buffered consumers.
3. Copy the chunk outside the exchange lock.
4. Append it to pooled memory or the shared spool.
5. Publish the completed append and wake readers.
6. Append it to the materialized working table when materialized consumers exist.

Only one buffered append reservation can be published at a time. Chunk copying and spool I/O do not hold the exchange mutex. Admission is checked again after copying so cancellation, consumer retirement, and backpressure changes are observed before publication.

### In-Memory Buffering And Backpressure

Buffered positions are monotonically increasing chunk indexes. Each buffered consumer tracks its next position. A pooled chunk can be retired once no active consumer or in-progress reader can still reference it. For a single buffered consumer, the exchange uses chunk-count watermarks:

- At 32 buffered chunks, the producer blocks.
- Once retirement reaches 16 buffered chunks, blocked producers are woken.

Chunk counts are intentionally used instead of byte estimates. This keeps admission and retirement deterministic and avoids maintaining a second approximate memory-accounting model beside DuckDB's buffer management.

### Shared Spool

When broadcast fanout would otherwise keep many pooled chunks alive, the exchange promotes retained chunks into one shared `ColumnDataCollection`-backed spool. This applies when there are multiple exchange consumers, or direct and buffered consumers coexist, and the pooled buffer reaches its high watermark.

The spool provides a shared positional stream. Every buffered consumer has its own reader state, while stored data is shared. Spool segments are retired according to the slowest active or in-progress reader. The spool uses DuckDB's existing buffer management and can spill under memory pressure.

### Blocking And Wakeups

The exchange has distinct wait lists for readers, watermark-blocked writers, and producers waiting for the single append reservation. Callbacks are always collected while holding the exchange lock and invoked after releasing it.

Important transitions are logged, rather than every chunk operation: spool creation, high-watermark blocking, low-watermark wakeup, and consumer unregistration.

### Early Completion And Cancellation

When an exchange consumer finishes or stops reading, it unregisters. Retained chunks are retired against the remaining consumers. If no exchange consumer remains, a side-effect-free producer can return `FINISHED` early and release buffered storage.

CTE bodies with side effects use `RUN_TO_COMPLETION`. Their exchange consumers may finish early, but the producer continues so all DML side effects occur. Cancellation changes the producer state, deactivates consumers, releases storage when no in-progress read or append still owns it, and wakes every class of blocked task.

## Pipeline Dependencies

The exchange introduces dataflow dependencies rather than completion dependencies for streaming consumers:

- Buffered readers can run while the producer is active and block when they catch up.
- Direct consumers are driven by the producer and have an external-finish dependency so their events complete after external input is finalized.
- Materialized consumers retain the normal dependency on CTE producer completion.

Prepared plans can rebuild their pipelines. Before rebuilding, the exchange ends the previous buffer execution, clears direct pipeline registrations and blocked task state, and resets every consumer to `UNRESOLVED`. Pipeline construction then assigns all consumers again. Runtime reset preserves the resolved assignments while resetting positions, lifecycle, progress, and storage.

## Correctness And Safety

The implementation preserves the following invariants:

- A materialized CTE body is executed once.
- Every active buffered consumer observes every published chunk exactly once.
- Direct consumers receive each producer chunk exactly once, including across blocked/resumed pushes.
- Materialized consumers observe the complete working table.
- DML CTE bodies complete even when consumers stop early.
- Pooled chunks and spool segments are not retired while an active or in-progress reader can reference them.
- Append publication is serialized, while copying and spool I/O occur outside the exchange lock.
- Blocked tasks are woken on data publication, retirement, producer completion, cancellation, and pipeline reset.
- Consumer selection has one owner and is rebuilt rather than inferred from stale physical state.

The synchronization paths use `annotated_mutex` and explicit state enums for producer state, consumer lifecycle, read state, append reservation state, watermarks, and external-input event state. This makes illegal or unresolved transitions assertable and keeps TSAN-visible ownership explicit.

## Supporting Changes

Validation of streaming CTE execution exposed runtime-filter cast behavior that was previously hidden by full materialization and different pipeline timing. This PR therefore also makes join runtime-filter reconstruction preserve cast failure semantics. Reconstructed narrowing cast chains use `TRY_CAST` only for the optional runtime filter, while the original join expression remains responsible for query-visible conversion errors. Runtime-filter pushdown remains general and is not keyed to CTEs or query names.

The PR also includes focused fixes and coverage for external joins with many duplicate keys, externally-fed pipeline event lifetime, prepared CTE pipeline rebuilds, correlated CTE output ordering, and extension-created or optimizer-created CTE plans.


lnkuiper:
I saw this in CI:
```
benchmark/tpch/sf10/q07.benchmark
Old timing: 0.0443185
New timing: 0.10119249999999999
```
But I built this branch locally and could not reproduce

kryonix:
The q07 regression was real, but is resolved now.

lnkuiper:
Still some differences (improvements and regressions) but overall geomean is only 1% slower so that's OK from me.
```
benchmark/tpch/sf10/q03.benchmark
Old timing: 0.098633
New timing: 0.055284

benchmark/tpch/sf10/q07.benchmark
Old timing: 0.042372
New timing: 0.0524735
```

kryonix:
That's a much bigger diff. than I measured locally. I will rebase and check q07 again.

kryonix:
Removing the CTE statistics again (for now?) was the right call:

TPC-DS:
```
Old timing geometric mean: 0.01782409098689881
New timing geometric mean: 0.016357362451187598, roughly 8% faster
```

This did not change TPC-H Q07, there's something else going on.

src/execution/operator/set/physical_cte.cpp
```diff
@@ -1,11 +1,118 @@
...
+	case CTEExecutionMode::MATERIALIZED:
+		return "MATERIALIZED";
+	case CTEExecutionMode::STREAMING_FANOUT:
+		return "STREAMING_FANOUT";
+	case CTEExecutionMode::HYBRID_FANOUT:
+		return "HYBRID_FANOUT";
+	default:
+		throw InternalException("Unknown CTE execution mode");
+	}
+}
```
lnkuiper: Shouldn't these be in `EnumUtil`? Here's also OK, just seems like something `EnumUtil` could do for us.

src/include/duckdb/execution/operator/set/physical_cte.hpp
```diff
@@ -27,10 +66,13 @@ class PhysicalCTE : public PhysicalOperator {
 	vector<const_reference<PhysicalOperator>> cte_scans;
 
 	shared_ptr<ColumnDataCollection> working_table;
+	shared_ptr<PipelineBroadcastExchange> exchange;
 
 	TableIndex table_index;
 	Identifier ctename;
 	bool cte_body_is_dml = false;
+	idx_t materialized_consumer_count = 0;
```
lnkuiper: Can we make the exchange the single source of truth for consumer modes/counts?

`ConsumerMode::MATERIALIZED` is stored in `PipelineBroadcastExchange`, while this counter is reset/incremented separately and `PhysicalCTEConsumerSource::consumer_mode` is also updated independently for explain output. `GetExecutionMode()` consequently combines state from multiple owners.

A small exchange summary API, e.g. counts by `ConsumerMode`, would let `PhysicalCTE` derive `MATERIALIZED`/`HYBRID_FANOUT` directly and remove the drift risk across pipeline rebuilds.

src/execution/operator/join/physical_hash_join.cpp
```diff
@@ -1359,19 +1356,97 @@ static unique_ptr<ExpressionFilter> CreateSelectivityOptionalExpressionFilter(un
...
+		return true;
+	}
+	if (source_type.IsSigned() == target_type.IsSigned()) {
+		return false;
+	}
+	if (source_type.IsSigned()) {
+		return true;
+	}
+	return source_size >= target_size;
+}
```
lnkuiper, line 1391: This would be so much cleaner if casts were scalar functions that have `SetFallible(true)`. That's a separate issue, just another reason for the rework!

src/include/duckdb/parallel/pipeline_broadcast_exchange.hpp
```diff
@@ -0,0 +1,215 @@
...
+	enum class WatermarkState : uint8_t { BELOW_HIGH_WATERMARK, ABOVE_HIGH_WATERMARK };
+	enum class WriterWakeMode : uint8_t { LOW_WATERMARK, FORCE };
+	enum class AppendAdmission : uint8_t { READY, BLOCKED, UNCONSUMED, CANCELLED };
+	enum class BufferedPushState : uint8_t { NOT_REQUIRED, APPENDED, BLOCKED, UNCONSUMED, CANCELLED };
+	enum class ExchangeLogEvent : uint8_t {
+		SPOOL_CREATED,
+		HIGH_WATERMARK_BLOCKED,
+		LOW_WATERMARK_WAKE,
+		CONSUMER_UNREGISTERED
+	};
```
lnkuiper, line 129: The state machine cleanup really nice with all the enums now. I do worry that this class is a bit of a "god mode" class that handles everything (external execution, in-memory buffering, spilled buffering, backpressure, cancellation, registration, progress, and logging). Maybe it would be better to separate concerns into classes so it's easier to reason about (`chunks`, `shared_spool`, positions, retirement). What do you think?

lnkuiper (commented):
Some minor comments below, looking nice now

src/include/duckdb/execution/operator/set/physical_cte.hpp
kryonix: Good point. I will change this.

src/include/duckdb/parallel/pipeline_broadcast_exchange.hpp
kryonix: Yeah, it's doing a lot. I will try a split.

Mytherin:
Thanks!

PR #23753, opened 2026-07-10
Preserve NULL semantics in CTE deliminator

kryonix:
TPC-DS Q16 returned incorrect results for generated datasets, initially observed at SF100. The issue was caused a pre-existing bug in the CTE Deliminator.

When decorrelating the query’s EXISTS subquery, the CTE Deliminator removed an evidence join and folded its regular `<>` predicate into a `MARK` join as `IS DISTINCT FROM`. This changes `NULL` semantics: a nullable probe key that should produce `UNKNOWN` for `NULL <> value` was instead treated as a successful inequality match. As a result, Q16 counted an additional order and included its net profit in the aggregate.

The rewrite is now only applied when the probe expression can be proven non-null. The check follows the expression through projections, filters, joins, scans, and CTE references. If nullability cannot be ruled out, the evidence join is retained and regular SQL `NULL` semantics are preserved. Non-null cases, including TPC-H Q21, continue to use the optimized plan.

### DIFF
--- a/extension/tpcds/dsdgen/answers/sf3/16.csv
+++ b/extension/tpcds/dsdgen/answers/sf3/16.csv
@@ -1,2 +1,2 @@
 order count|total shipping cost|total net profit
-691|3264282.42|-407678.69
+691|3264157.47|-407644.04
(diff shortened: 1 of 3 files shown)


kryonix:
There was a hidden bug in the system when we added SF3 for TPC-DS to the regression CI, resulting in a wrong expected result. I fixed the initial bug and the expected result. Therefore, this PR can not pass the TPC-DS SF3 CI.

Mytherin:
Thanks!

PR #23756, opened 2026-07-10
Split up `LogicalSimple` into separate classes

Mytherin:
`LogicalSimple` was originally created mostly to make it easier to write add simple logical operators, but there's not really any point to having it. This PR splits it up into separate classes which is better for readability and future extensibility.


src/planner/binder/statement/bind_alter.cpp
```diff
@@ -112,7 +113,7 @@ BoundStatement Binder::Bind(AlterStatement &stmt) {
 		auto &properties = GetStatementProperties();
 		properties.return_type = StatementReturnType::NOTHING;
 		properties.RegisterDBModify(Catalog::GetSystemCatalog(context), context, DatabaseModificationType::ALTER_TABLE);
-		result.plan = make_uniq<LogicalSimple>(LogicalOperatorType::LOGICAL_ALTER, std::move(stmt.info));
+		result.plan = make_uniq<LogicalAlter>(std::move(stmt.info));
```
Tishj, line 114: This file is still called `bind_simple.cpp`

PR #23757, opened 2026-07-10
Fix segfault in external hash aggregate when radix bits grow after going external

carlopi:
A thread's `abandoned_data` (the external-spill target) is sized once at the global radix bits of when it first went external and then frozen, but the global radix bits can keep growing afterwards along with the thread's own HT data. Merging the larger HT data into the smaller, stale `abandoned_data` makes `RepartitionFinalizeStates` compute `NumberOfPartitions(new_radix_bits - old_radix_bits)` with unsigned arithmetic: new < old underflows to a huge multiplier, indexing the partition and pin-state vectors out of bounds leading to segfault.

Fix: `GrowAbandonedDataToRadixBits` repartitions `abandoned_data` up to the current global radix bits (the supported new >= old direction) before merging into it, called in both the sink-phase external repartition and before the combine-phase Combine. As defense-in-depth, PartitionedTupleData::Repartition now throws instead of underflowing when asked to go to fewer partitions.

### DIFF
--- a/.github/config/extensions/lance.cmake
+++ b/.github/config/extensions/lance.cmake
@@ -3,6 +3,7 @@ if (NOT MINGW AND NOT ${WASM_ENABLED})
             GIT_URL https://github.com/lance-format/lance-duckdb
             GIT_TAG cd592bc76dfeac28bc7d6344a5bf85c4d4c31405
             SUBMODULES extension-ci-tools
+            APPLY_PATCHES
             LOAD_TESTS
             DONT_LINK
     )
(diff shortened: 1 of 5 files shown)


carlopi:
I think I have an explanation of a segfault that was reported.

I managed to reproduce (with some weird configuration / tweaking the code to trigger in a controlled way) what I think is this chains of events:
```
SIGABRT (134): old_radix_bits=5 new_radix_bits=3 → NumberOfPartitions(3-5) underflows → partitions[] OOB → segfault in FinalizePinState
```

This PR is based on the observations, and avoids the above issue, but would need to be properly reviewed and validated, this is very much not my area of the codebase.

PR #23759, opened 2026-07-10
Preserve verbatim ATTACH path via AttachedDatabase::GetOriginalPath()

carlopi:
Capture the user-supplied ATTACH path before extension-prefix stripping, so storage extensions (and introspection) can access it. Stored as an optional<string> on AttachOptions and AttachedDatabase and exposed via `GetOriginalPath()`; unset for databases not attached from an ATTACH statement.


Mytherin:
Thanks!

PR #23764, opened 2026-07-11
Add binom function

hacoobamatata:
## Summary

This PR adds a binom(n, k) scalar function for computing binomial coefficients.

The function returns 0 when k > n, throws an error for negative inputs, and returns a HUGEINT.

The implementation uses GCD-based factor cancellation before multiplication to reduce intermediate overflow.

## Testing

Added sqllogictests covering basic cases, edge cases, NULL inputs, negative inputs, large valid results, and overflow behavior.


lnkuiper:
Thanks!

PR #23765, opened 2026-07-11
Fix CLI JSON rendering for nested VARIANT

### LINKED ISSUE #23745: CLI JSON rendering internal error for nested VARIANT
### What happens?

### To Reproduce

```sql
memory D .mode json
memory D SELECT [{'a': 1, 'b': [1]}::VARIANT] AS v;
INTERNAL Error:
Unsupported type arrived at JSON create function
```

### OS:

macos

### DuckDB Version:

3ea25ffc3ca8e3d3d5f0c2eb177f3d5038ad04dc

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

souvikDevloper:
## Summary

- serialize `VARIANT` values encountered during recursive JSON creation through DuckDB's canonical `VARIANT` to `JSON` cast
- add SQL and CLI regression coverage for nested `VARIANT` values

## Root cause

The CLI JSON renderer casts nested result vectors directly to `JSON`. The JSON creation visitor recursed through lists and structs but treated a nested `VARIANT` as unsupported, even though normal scalar-function binding inserts a `VARIANT` to `JSON` cast for top-level values.

## Impact

Nested `VARIANT` values now render correctly in the CLI's `json` and `jsonlines` modes instead of raising an internal error.

## Validation

- built the DuckDB library and shell with the JSON extension using MSVC
- ran `python -m pytest tools/shell/tests/test_shell_rendering.py --shell-binary <duckdb.exe> -q` (`14 passed`)
- verified `SELECT [{'a': 1, 'b': [1]}::VARIANT] AS v;` in both `json` and `jsonlines` modes
- ran DuckDB's formatter and typo checker

Fixes #23745

### DIFF
--- a/extension/json/json_functions/json_create.cpp
+++ b/extension/json/json_functions/json_create.cpp
@@ -933,6 +933,12 @@ static void CreateValues(const StructNames &names, yyjson_mut_doc *doc, yyjson_m
 		CreateRawValues(doc, vals, string_vector, count);
 		break;
 	}
+	case LogicalTypeId::VARIANT: {
+		Vector json_vector(LogicalType::JSON(), count);
+		VectorOperations::DefaultCast(value_v, json_vector, count);
+		TemplatedCreateValues<string_t, string_t>(doc, vals, json_vector, count);
+		break;
+	}
 	case LogicalTypeId::DECIMAL: {
 		if (DecimalType::GetWidth(type) > 15) {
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #23767, opened 2026-07-11
Convert TIMETZ to UTC on parquet write

### LINKED ISSUE #23766: TIMETZ type not work for roundtrip
### What happens?

Hi team, I found when we write TIMETZ to parquet, we cannot read out the same value; see repro SQL for details.

### To Reproduce

```sql
memory D COPY (
             SELECT TIMETZ '12:34:56+02' AS x
         ) TO '/private/tmp/duckdb_timetz.parquet' (
             FORMAT parquet
         );
memory D SELECT x::VARCHAR, typeof(x)
         FROM read_parquet('/private/tmp/duckdb_timetz.parquet');
┌────────────────────┬─────────────────────┐
│ CAST(x AS VARCHAR) │      typeof(x)      │
│      varchar       │       varchar       │
├────────────────────┼─────────────────────┤
│ 12:34:56+00        │ TIME WITH TIME ZONE │
└────────────────────┴─────────────────────┘
```

### OS:

macos

### DuckDB Version:

2009475939c7b5ada2e55cc74843339bd9be79e3

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23766

The root cause is, on parquet write, we mark it as [converted to UTC timezone](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/extension/parquet/parquet_writer.cpp#L270), but on write it's getting the [microseconds for `dtime_tz_t`](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/extension/parquet/include/writer/parquet_write_operators.hpp#L319-L321) at [local timezone](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/src/include/duckdb/common/types/datetime.hpp#L149-L151) without considering offsets. In one word, we loss the offset information on write.

On read, we're constructing with [UTC timezone](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/extension/parquet/parquet_timestamp.cpp#L149-L152); so the fix in this PR is to store in UTC timezone directly.

### DIFF
--- a/extension/parquet/include/writer/parquet_write_operators.hpp
+++ b/extension/parquet/include/writer/parquet_write_operators.hpp
@@ -11,6 +11,7 @@
 #include "writer/parquet_write_stats.hpp"
 #include "parquet_timestamp.hpp"
 #include "zstd/common/xxhash.hpp"
+#include "duckdb/common/types/time.hpp"
 #include "duckdb/common/types/uhugeint.hpp"
 #include "duckdb/common/types/uuid.hpp"
 
@@ -317,7 +318,7 @@ struct ParquetUUIDOperator : public BaseParquetOperator {
 struct ParquetTimeTZOperator : public BaseParquetOperator {
 	template <class SRC, class TGT>
 	static TGT Operation(SRC input) {
-		return input.time().value;
+		return Time::NormalizeTimeTZ(input).value;
 	}
 
 	template <class SRC, class TGT>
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #23769, opened 2026-07-11
Patch postgres_scanner: report NOTHING for CONNECT-routed commands

carlopi:
When postgres_query prepares a statement that returns no result columns (DDL, or DML without RETURNING) it takes the command_only path. Set the table function's call_return_type to NOTHING there so that, combined with the core `SELECT * FROM func()` propagation, a command routed through CONNECT to a postgres database displays like a native command instead of a "Success" table.

Carried as an extension patch against the pinned postgres_scanner; the permanent change belongs in duckdb/duckdb-postgres.

This commit was somehow lost in history, like merged and then the patch deleted but not applied. Let's bring this back.


carlopi:
Sent out also to postgres directly at https://github.com/duckdb/duckdb-postgres/pull/516

carlopi:
Outdated, see https://github.com/duckdb/duckdb/pull/23777

PR #23770, opened 2026-07-12
Fix lance build on Rust 1.97: patch its Cargo.lock to ethnum 1.5.3

carlopi:
lance-duckdb's committed Cargo.lock pins ethnum 1.5.2, whose error.rs fabricates a `core::num::TryFromIntError` via `mem::transmute(())`. That assumed the type was zero-sized; it is now 8 bits, so on Rust >= 1.97 the build fails with E0512. Rust 1.97 went stable on 2026-07-07, which is what turned this into a CI failure here.

ethnum 1.5.3 replaces the transmutes with safe constructors. jsonb (the crate that pulls ethnum in, via lance-arrow) requires `ethnum ^1.5.2`, so 1.5.3 already satisfies it: the fix is purely a lockfile refresh, a two-line diff with no manifest change and no other dependency moving.

Carry that refresh as an extension patch so the build is unblocked without waiting on upstream. The patch context is generated against the pinned GIT_TAG cd592bc, so the tag and the patch must be bumped together.

The durable fix belongs upstream in lance-format/lance-duckdb, which has not refreshed its lockfile; the sibling lance-format/lance repo already has.

### DIFF
--- a/.github/config/extensions/lance.cmake
+++ b/.github/config/extensions/lance.cmake
@@ -3,6 +3,7 @@ if (NOT MINGW AND NOT ${WASM_ENABLED})
             GIT_URL https://github.com/lance-format/lance-duckdb
             GIT_TAG cd592bc76dfeac28bc7d6344a5bf85c4d4c31405
             SUBMODULES extension-ci-tools
+            APPLY_PATCHES
             LOAD_TESTS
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #14956, opened 2024-11-24
Implement Logical Compaction in Hash Join Operator

YimingQiao:
## Description  
One year ago, I identified an opportunity to optimize the data chunk compaction mechanism in DuckDB. In vectorized execution, processing one chunk at a time is a key to high performance. However, chunk-reducing operators, such as filters and hash joins, can produce smaller chunks, leading to performance bottlenecks. To address this issue, I proposed two solutions: Dynamic Compaction and Logical Compaction. The research on this topic resulted in the paper "*[Data Chunk Compaction in Vectorized Execution](https://yimingqiao.github.io/files/data-chunk-compaction-sigmod25.pdf)*," which has been accepted for SIGMOD 2025.

This pull request implements **Logical Compaction**, which aims to reduce the number of resulting chunks from the hash join operator. 

---

## Design Overview  
In Logical Compaction, the key design changes are as follows:  
- **For the LHS vectors:** The selection vectors of result chunks are concatenated into a single selection vector.  
- **For the RHS vectors:** All payloads are gathered sequentially into the same vector, instead of being spread across multiple vectors.  

This implementation ensures fewer chunks are produced, improving query performance by reducing overhead in subsequent operations. I have a [sample code repository](https://github.com/YimingQiao/Chunk-Compaction-in-Vectorized-Execution), describing the idea of how to implement Logical Compaction.

---

## Implementation Notes  
The current implementation of Logical Compaction differs slightly from the method described in the SIGMOD paper due to a limitation in DuckDB's `gather` function. Currently, the `gather` function cannot be used multiple times on the same vector, as subsequent gathers overwrite the results of earlier ones. Modifying this behavior in DuckDB would require significant changes. (I have attempted to modify the gather function. It currently works for List and common data types but fails for Array).

As a workaround:  
1. Payload pointers are buffered first.  
2. The `gather` function is called once per result chunk to populate the final vector.  

Although this approach is suboptimal compared to the method in the paper, it simplifies integration into DuckDB and achieves considerable performance gains.

---

## Performance Evaluation  
The following query was used to evaluate the impact of Logical Compaction:

```sql
-- DuckDB Settings
SET threads TO 1;

-- Generate Tables
CREATE OR REPLACE TABLE student AS
SELECT
    CAST(stu_id AS INT) AS stu_id,
    CAST((RANDOM() * 5e6) AS INT) AS major_id,
    CAST((RANDOM() * 100) AS TINYINT) AS age
[...]

PR #23771, opened 2026-07-12
Accumulate probe match count thread-locally

gropaul:
https://github.com/duckdb/duckdb/pull/19683 introduced logging for the number of probe matches, which can cause a performance regression in specific cases:

With the very selective n:m joins of the JOB benchmark, it can happen (often) that in a join only a few tuples of an incoming chunk find a match, and that these tuples produce many outputs, as there are many duplicates in the RHS, and therefore `AdvancePointer` / `ResolvePredicates` is called many, many times with only a few tuples. 

In these scenarios, `ht.total_probe_matches.fetch_add` in `ResolvePredicates` became a bottleneck when running JOB with many threads. I have now added a local aggregate for this scenario that is then flushed to the global state once all chains have been exhausted.

A more structural but complicated way could be to do something like https://github.com/duckdb/duckdb/pull/14956 to compact the vectors by asking for more input rows for selective probes before calling `ScanStructure::Next`. This way, `ResolvePredicates` would never be hit with fewer than N tuples. 

I did run experiments on my MacBook Pro M4 with 12 threads. For this setting, JOB 16b is now down from 192ms to 95ms.

Here is a profiler comparison of the whole JOB/IMDB benchmark. Before the PR, Next Inner Join accounted for 18% of the total runtime; now it is at 3.7%.
flamegraph_join_atomic

### DIFF
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -1446,6 +1446,10 @@ void ScanStructure::Next(DataChunk &keys, DataChunk &probe_data, DataChunk &resu
 	default:
 		throw InternalException("Unhandled join type in JoinHashTable");
 	}
+
+	if (PointersExhausted()) {
+		FlushProbeMatches();
+	}
 }
 
 bool ScanStructure::PointersExhausted() const {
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks!

PR #23772, opened 2026-07-12
Fix narrow cast invertibility

### LINKED ISSUE #23687: Optimizer makes narrow cast invertible
### What happens?

### To Reproduce

```sql
memory D INSERT INTO t VALUES (51), (127), (128), (200), (300), (-200);

-- Two errors: the numbers returned are not in the table, and filter not applied.
memory D SELECT list(a ORDER BY a) FROM t WHERE TRY_CAST(a AS TINYINT) > 50;
┌──────────────────────────┐
│    list(a ORDER BY a)    │
│         int32[]          │
├──────────────────────────┤
│ [51, 127, 128, 200, 300] │
└──────────────────────────┘

-- Disabling optimizer leads to the correct result.
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D SELECT list(a ORDER BY a) FROM t WHERE TRY_CAST(a AS TINYINT) > 50;
┌────────────────────┐
│ list(a ORDER BY a) │
│      int32[]       │
├────────────────────┤
│ [51, 127]          │
└────────────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

sakshichitnis27:
Fixes #23687.

### Summary
`BoundCastExpression::CastIsInvertible` incorrectly treated narrowing integral casts as invertible. This allowed the comparison simplification optimizer to remove casts such as `TRY_CAST(INTEGER AS TINYINT)`, resulting in incorrect behaviour for values outside the target type's range.
This change ensures that an integral cast is considered invertible only when the target type can represent all values of the source type.

Regression tests have also been added to cover:
- Correct results for `TRY_CAST`
- Error preservation for regular `CAST`

### Tests
- `make reldebug`
- `build/reldebug/test/unittest test/sql/optimizer/expression/test_comparison_simplification.test`
- Related IN-clause and filter-pushdown optimizer tests

### DIFF
--- a/src/optimizer/rule/comparison_simplification.cpp
+++ b/src/optimizer/rule/comparison_simplification.cpp
@@ -117,6 +117,11 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op, 
 		//! invertible in practice.
 		auto &cast_expression = column_ref_expr.Cast<BoundCastExpression>();
 		auto target_type = cast_expression.source_type();
+		if (cast_expression.IsTryCast() && target_type.IsIntegral() && cast_expression.GetReturnType().IsIntegral() &&
+		    LogicalType::DefaultForceMaxLogicalType(target_type, cast_expression.GetReturnType()) !=
+		        cast_expression.GetReturnType()) {
+			return nullptr;
+		}
 		if (!BoundCastExpression::CastIsInvertible(target_type, cast_expression.GetReturnType())) {
 			return nullptr;
 		}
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks but I think this is too broad - isn't this only a problem when `try_cast` is used (i.e. it is still true for regular casts)?

sakshichitnis27:
@Mytherin 
Thanks, you’re right. I’ve updated this in a0974e0 so the optimization is blocked only for narrowing integral TRY_CAST comparisons. The global cast-invertibility behavior and regular CAST optimization are unchanged. The regression test is now limited to the reported TRY_CAST case

PR #23773, opened 2026-07-12
fix: VARCHAR to DECIMAL scientific-notation rounding truncation

### LINKED ISSUE #23768: Casting scientific-notation VARCHAR to DECIMAL rounds up instead of truncating to zero for negative exponents with mantissa >= 5
### What happens?

Direct VARCHAR -> DECIMAL cast of scientific-notation strings with large
negative exponents returns one unit in the last place instead of 0 when
the mantissa's leading digit is >= 5. The same value cast through DOUBLE,
or written as a plain decimal string, is correct.

This looks like the VARCHAR -> DECIMAL sibling of #20098, which fixed the
same negative-exponent rounding logic for the VARCHAR -> INTEGER cast
("incorrectly return 1 instead of 0 for decimals >= 5"). The DECIMAL path
still exhibits the behavior.

Reproduced on 1.5.4 (PyPI release) and 1.6.0.dev280 (latest dev build),
Python client and CLI, Linux x64.

### To Reproduce

```
SELECT CAST('5.8207660913467407e-11' AS DECIMAL(18,4));
-- returns 0.0001, expected 0.0000

SELECT CAST('5.5e-11' AS DECIMAL(18,4));
-- returns 0.0001, expected 0.0000

SELECT CAST('4.9e-11' AS DECIMAL(18,4));
-- returns 0.0000, correct (leading mantissa digit < 5)

SELECT CAST(CAST('5.8207660913467407e-11' AS DOUBLE) AS DECIMAL(18,4));
-- returns 0.0000, correct (bridged through DOUBLE)

SELECT CAST('0.000000000058207660913467407' AS DECIMAL(18,4));
-- returns 0.0000, correct (plain decimal string, same value)
```

### OS:

Linux x64

### DuckDB Version:

DuckDB 1.5.4 (release) and 1.6.0.dev280 (latest dev build, still reproduces)

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Nikola Grujicic

### Affiliation:

EPS

- [x] Yes, I have

- [x] Yes, I have

Functionhx:
Fixes #23768. Same root cause as #20098 — negative-exponent loop missed the mod==0 guard in the DECIMAL path.

### DIFF
--- a/src/include/duckdb/common/operator/decimal_cast_operators.hpp
+++ b/src/include/duckdb/common/operator/decimal_cast_operators.hpp
@@ -569,7 +569,7 @@ struct DecimalCastOperation {
 				auto mod = state.result % 10;
 				round_up = NEGATIVE ? mod <= -5 : mod >= 5;
 				state.result /= 10;
-				if (state.result == 0) {
+				if (state.result == 0 && mod == 0) {
 					break;
 				}
 			}
--- a/test/sql/types/decimal/decimal_exponent.test
+++ b/test/sql/types/decimal/decimal_exponent.test
@@ -117,3 +117,114 @@ statement error
 SELECT '1f3'::DECIMAL
 ----
 <REGEX>:Conversion Error.*Could not convert string.*
+
+# large negative exponent with mantissa >= 5 should truncate to 0, not round up (issue #23768)
+query I
+SELECT '5.8207660913467407e-11'::DECIMAL(18,4)
+----
+0.0000
+
+query I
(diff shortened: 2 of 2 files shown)


Mytherin:
Thanks - could we add a test?

Functionhx:
Tests added — 21 cases covering the scientific notation rounding fix from #23768, including the original reported bug, positive/negative mantissas, boundary values, and multiple precision scales. All pass locally (51 assertions).

Mytherin:
Thanks!

PR #23778, opened 2026-07-12
bound alp/alprd exception positions to vector size

jmestwa-coder:
**Out-of-bounds write from unchecked ALP/ALPRD exception positions**

Each exception position is read from the compressed segment and used to index the decoded output, but only the exception count and payload sizes were validated, so a crafted file with a position past the vector size writes a controlled value out of bounds when the column is scanned. Bounded the positions against vector_size in both scan paths.


lnkuiper:
Thanks! I've ran the microbenchmarks and this doesn't seem to regress alp/alprd performance.

PR #23779, opened 2026-07-12
Fix external file cache crash on insertion

dentiny:
Closes https://github.com/duckdblabs/duckdb-internal/issues/9916

The issue is
- External file cache store file keys to track what's the cache entries for cache inspection and O(1) cache clear; [keys are managed in object cache](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/src/storage/external_file_cache/external_file_cache.cpp#L17-L58) in LRU style, so they'll get evicted under memory pressure and don't get OOM
  + Impl wise, external file cache itself keeps an [`unordered_set`](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/src/include/duckdb/storage/external_file_cache/external_file_cache.hpp#L114-L116), with eviction part delegates to object cache
- Current implementation is built on an incorrect assumption that, for all the cache keys, these two data structures are kept in-sync, but that's not the case
  + Currently they're "kept in-sync" by callback in object cache entry [constructor](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/src/storage/external_file_cache/external_file_cache.cpp#L19-L22) and [destructor](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/src/storage/external_file_cache/external_file_cache.cpp#L24-L26) 
  + But it's possible that, when a cache key gets evicted from object cache, there're still valid reference (since it's returned in shared pointer to guarantee access always valid even evicted) => when user fails to lookup in object cache and recreate, our [invariant check](https://github.com/duckdb/duckdb/blob/2009475939c7b5ada2e55cc74843339bd9be79e3/src/storage/external_file_cache/external_file_cache.cpp#L367) fails

I thought about a few possible solutions to fix the problem
- We don't store cache keys in external file cache, instead when they're needed, we re-compute it from object cache
  + pro: no threading issue, no lifecycle issue, easy to maintain and reason
  + con: performance is bad to disable cache and getting cache info; for better perf, we need to maintain both an unordered map and an ordered map for cache keys
- To keep the consistency between object cache and external file cache key sets, callbacks should be places at object cache entry insertion and eviction, instead of construction and destruction
  + I did a draft change here: https://github.com/dentiny/duckdb/pull/137
  + con: bigger code change, same complexity as the current implementation
- (selected) Admit one cache key could exist for multiple times in the external file cache key set, instead of exactly one, so we maintain a map of <cache key, key count>, instead of an unordered set
  + pro: little code change, same mental model
  + con: it could be a little difficult to understand the counter


Mytherin:
Thanks - haven't reviewed the code yet but should this target `v1.5-variegata`?

dentiny:
> Thanks - haven't reviewed the code yet but should this target `v1.5-variegata`?

V1.5 and main has big code difference 😅
I plan to create a PR later this week to backport all fixes

lnkuiper:
Looks good! Have a check whether this should be backported, not sure since a lot has changed.

PR #23781, opened 2026-07-13
Fix INTERNAL error casting pre-epoch TIMESTAMP_NS to TIME_NS

### LINKED ISSUE #23722: Timestamp type casting leads to INTERNAL exception
### What happens?

### To Reproduce

```sql
memory D SELECT (TIMESTAMP_NS '1969-12-31 23:59:59.999999999'::TIME_NS)::VARCHAR;
INTERNAL Error:
Information loss on integer cast: value -952 outside of target range [-128, 127]
```

### OS:

macos

### DuckDB Version:

985b2f2515adf44ea65dd4239c7bfb30981b94dc

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

apoorva-01:
Casting a `TIMESTAMP_NS` from the last microsecond before the epoch to `TIME_NS` throws an INTERNAL error:

```
SELECT (TIMESTAMP_NS '1969-12-31 23:59:59.999999999'::TIME_NS)::VARCHAR;
-- INTERNAL Error: Information loss on integer cast: value -952 outside of target range [-128, 127]
```

`GetTimeNs` gets the date by truncating the nanoseconds to microseconds first, so -1 ns lands on the epoch instead of 1969-12-31 and the time-of-day goes negative, which is what overflows the formatter. Switching to `GetDateNS` (already nanosecond-accurate) fixes it, and only pre-epoch values right on midnight change.

`::DATE` and `::TIME` are off on the same input too, but through a different path, so I left those alone. Added tests in `test/sql/types/time/test_time_ns.test` for the crash and a few midnight boundary cases.

Closes #23722

### DIFF
--- a/src/common/types/timestamp.cpp
+++ b/src/common/types/timestamp.cpp
@@ -380,7 +380,7 @@ dtime_ns_t Timestamp::GetTimeNs(timestamp_ns_t input) {
 	if (!input.IsFinite()) {
 		throw ConversionException("Can't get TIME_NS of infinite TIMESTAMP");
 	}
-	date_t date = Timestamp::GetDate(Timestamp::FromEpochNanoSeconds(input.value));
+	date_t date = Timestamp::GetDateNS(input);
 	int64_t nanos;
 	if (!TryMultiplyOperator::Operation<int64_t, int64_t, int64_t>(date.days, Interval::NANOS_PER_DAY, nanos)) {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #23782, opened 2026-07-13
Fix percentile_cont/disc(0) with descending order

### LINKED ISSUE #23692: percentile_cont yields incorrect result
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE p AS SELECT x::DOUBLE AS x FROM range(1, 6) t(x);

-- The first value should be 5.0.
memory D SELECT percentile_cont(0.0) WITHIN GROUP (ORDER BY x DESC) AS p0_desc,
         percentile_cont(0.25) WITHIN GROUP (ORDER BY x DESC) AS p25_desc,
         percentile_cont(1.0) WITHIN GROUP (ORDER BY x DESC) AS p1_desc,
         percentile_cont(0.0) WITHIN GROUP (ORDER BY x ASC) AS p0_asc
         FROM p;
┌─────────┬──────────┬─────────┬────────┐
│ p0_desc │ p25_desc │ p1_desc │ p0_asc │
│ double  │  double  │ double  │ double │
├─────────┼──────────┼─────────┼────────┤
│     1.0 │      4.0 │     1.0 │    1.0 │
└─────────┴──────────┴─────────┴────────┘
```

### OS:

macos

### DuckDB Version:

9ac18da6205b193046261596559a2d92737b201b

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

apoorva-01:
`percentile_cont(0.0) WITHIN GROUP (ORDER BY x DESC)` gives the min instead of the max:

```sql
SELECT percentile_cont(0.0) WITHIN GROUP (ORDER BY x DESC) FROM range(1,6) t(x);
-- returns 1.0, expected 5.0
```

DESC gets passed to the aggregate as a negative fraction, but `0.0` negates to `-0.0` which isn't `< 0`, so a zero fraction loses the descending flag and reads as ascending. The fix keeps a real `-0.0` through the binder and checks the sign with `signbit` (kept it in the existing sign convention rather than rewiring how the order reaches the aggregate). `percentile_disc(0.0)` and the list form had the same thing.

Tests added in both quantile files, fail on main and pass with the fix.

Closes #23692

### DIFF
--- a/extension/core_functions/aggregate/holistic/quantile.cpp
+++ b/extension/core_functions/aggregate/holistic/quantile.cpp
@@ -14,7 +14,15 @@
 #include "duckdb/function/aggregate/list_aggregate.hpp"
 #include "duckdb/function/create_sort_key.hpp"
 
+#include <cmath>
+
 namespace duckdb {
+
+// Descending order is encoded as a negative quantile parameter. A zero fraction
+// is stored as negative zero, so use the sign bit to detect that case as well.
+static bool QuantileDescending(const Value &q) {
+	return std::signbit(q.GetValue<double>());
+}
 
 template <class INPUT_TYPE>
 struct IndirectLess {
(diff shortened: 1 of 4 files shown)


hawkfish (approved):
Thanks for the patch! I kind of wish this was explicit and not encoded like this, but that is for another day.

lnkuiper:
Thanks! Could you have a look at the failing CI?

hawkfish:
> Thanks! Could you have a look at the failing CI?

The failure looks unrelated to me. @apoorva-01 I think you can just merge `main` to fix this.

PR #23784, opened 2026-07-13
Propagate architecture flags to C dependencies when compiling with NATIVE_ARCH=1

krenzland:
Set CMAKE_C_FLAGS alongside CMAKE_CXX_FLAGS when NATIVE_ARCH is enabled, ensuring C dependencies such as jemalloc are compiled with the native CPU target.
Without this patch, jemalloc does not receive the mcpu/march=native flags.


Mytherin:
Thanks!

PR #23785, opened 2026-07-13
validate metadata_offset and leading zero count in chimp scan

jmestwa-coder:
validate metadata_offset and leading_zero_block_count in the chimp scan state before use, so a corrupt segment can no longer read group metadata past the block or overrun the leading_zeros buffer (patas already guards both).


jmestwa-coder:
pushed a fix for the release build config failures. the test switched back to the memory catalog explicitly, which the force_storage configs don't have, so it now qualifies the attached catalog instead.

PR #21783, opened 2026-04-02
Add json_set() to set value at path in JSON document

mustafahasankhan:
## Summary

Adds `json_set(json, path, value) -> JSON` which sets a value at a given path in a JSON document. Creates intermediate parents if the path does not exist, overwrites if it does. This is the most commonly requested missing JSON mutation function, available in SQLite (`json_set`), MySQL (`JSON_SET`), and PostgreSQL (`jsonb_set`).

Related (closed): https://github.com/duckdb/duckdb/pull/12386

```sql
-- Add a new key
SELECT json_set('{"a":1}', '$.b', '2');
-- {"a":1,"b":2}

-- Overwrite existing
SELECT json_set('{"a":1}', '$.a', '99');
-- {"a":99}

-- Create intermediate parents
SELECT json_set('{}', '$.a.b.c', '1');
-- {"a":{"b":{"c":1}}}

-- Array index
SELECT json_set('[1,2,3]', '$[1]', '99');
-- [1,99,3]

-- Append to array
SELECT json_set('[1,2]', '/-', '3');
-- [1,2,3]

-- Set nested value
SELECT json_set('[{"a":1}]', '$[0].b', '2');
-- [{"a":1,"b":2}]
```

Supports both JSONPath (`$.key[0]`) and JSON Pointer (`/key/0`) syntax, matching DuckDB's existing `json_extract` path format. Bare key names are also accepted.

### Implementation

Converts JSONPath to JSON Pointer internally (with proper RFC 6901 escaping for `~` and `/`), then delegates to yyjson's native `yyjson_mut_doc_ptr_set` with fallback to `ptr_add` for array append operations. Uses `TernaryExecutor::ExecuteWithNulls` for the vectorized loop.

## Test plan

- [x] `test_json_set.test` - 39 assertions covering: add/overwrite keys, JSON Pointer and JSONPath syntax, nested paths, intermediate parent creation, array index overwrite, array append (`/-` and `$[#]`), objects in arrays, all value types, type changes, deeply nested, NULL handling, bare key names, explicit JSON cast, constant vectors, table-based queries, real-world entity mutation, composability with json_normalize
- [x] All existing JSON tests pass (80 test cases, 6724 assertions)
- [x] `clang-format 11.0.1` applied

PR #23786, opened 2026-07-13
Add json_set(), json_insert(), json_replace() and json_remove() scalar functions

mustafahasankhan:
This PR adds four scalar functions that modify JSON documents:

| Function | Behavior |
|---|---|
| `json_set(doc, path, value)` | Sets `value` at `path`. Creates the path if missing, overwrites if it exists. |
| `json_insert(doc, path, value)` | Inserts `value` at `path`. Does nothing if a value already exists there. |
| `json_replace(doc, path, value)` | Replaces the value at `path`. Does nothing if the path does not exist. |
| `json_remove(doc, path)` | Removes the value at `path`. Does nothing if the path does not exist. |

`json_set()` started as #21783 and is folded into this PR per review, so the family can be reviewed as a whole.

The names and semantics follow SQLite and MySQL, which ship these functions with the same no-overwrite / no-create / no-op behavior.

```sql
SELECT json_set('{}', '$.a[0].b', '1');
-- {"a":[{"b":1}]}   (missing containers are created like SQLite)
SELECT json_insert('{"a":1}', '$.a', '99');
-- {"a":1}   (existing values are never overwritten)
SELECT json_replace('{"a":1}', '$.b', '2');
-- {"a":1}   (missing paths are never created)
SELECT json_remove('[1,2,3]', '$[#-1]');
-- [1,2]
```

Path parsing is shared with the rest of the extension: a `JSONPathIterator` in `JSONCommon` yields typed path elements (key, index, reverse index, append, wildcard) on top of the existing `ReadKey`/`ReadArrayIndex` primitives, and `ValidatePath`/`GetPath` are rewritten on it. The modification functions consume the same iterator and walk the mutable document directly.

Path syntax:
- JSONPath style: `$.key[0]`, quoted keys with escapes (`$."a\"b"`), `$[#]` to append, `$[#-N]` or `$[-N]` for the N-th element from the end
- JSON Pointer style: `/key/0`, `/-` to append
- A bare top-level key name

Semantics:
- Malformed paths raise the same `JSON path error` as `json_extract`; wildcards raise an error since they cannot address a single location
- Missing containers are created like SQLite: only index 0 is valid in a freshly created array, scalars are never replaced, and the root's type is never converted. Creation is transactional, a partially resolvable path leaves the document untouched
- Well-formed paths that do not resolve (index beyond the array length, reverse index out of range) leave the document unchanged
- `json_remove(doc, '$')` returns `NULL`, same as SQLite
- `NULL` in any argument returns `NULL`

One difference from SQLite/MySQL: they accept variadic path/value pairs in a single call. This implementation takes one path per call, which keeps the signature aligned with the rest of DuckDB's JSON functions.

The four functions are registered in `extension_entries.hpp`. Tests cover objects, arrays, nested paths, appends, reverse indexes, container creation, no-op cases, NULL handling, quoted keys containing dots, slashes, tildes, quotes and backslashes, unicode keys, root operations, malformed paths, wildcard rejection, and constant/table execution.


lnkuiper:
Thanks for the PR! Almost ready to go, just a few comments:
1. Could you add https://github.com/duckdb/duckdb/pull/21783 to this PR? Then they can be reviewed as a whole.
2. Can you add these functions to `extension_entries.hpp`?
3. I think there are a few issues with the JSONPointer to JSONPath conversion:
```sql
# name: test/temp.test
# description: Reproducers for JSON mutation path conversion

require json

# #21783: quoted JSONPath keys must handle escaped quotes
query T
SELECT json_set('{"a\"b":1,"a\\":2}', '$."a\"b"', '9')
----
{"a\"b":9,"a\\":2}

# #21783: [#-1] refers to the last array element
query T
SELECT json_set('[1,2]', '$[#-1]', '9')
----
[1,9]

# #23786: the same escaped-key bug affects removal
query T
SELECT json_remove('{"a\"b":1,"a\\":2}', '$."a\"b"')
----
{"a\\":2}

# #23786: reverse indexes also apply to replace and insert
query T
SELECT json_replace('[1,2]', '$[#-1]', '9')
----
[1,9]

query T
SELECT json_insert('[1,2]', '$[#-1]', '9')
----
[1,2]
```

mustafahasankhan:
Thanks for the review, all three points addressed:

1. `json_set` is folded into this PR and #21783 is closed.
2. `json_insert`, `json_remove`, `json_replace` and `json_set` are registered in `extension_entries.hpp`.
3. Both were real bugs, thanks for the reproducers. Quoted keys now handle `\"` and `\\` escapes, and `#` / `#-N` resolve against the actual array length, so `$[#-1]` targets the last element. The escaped-quote case was nastier than it looks: `$."a\"b"` used to silently modify the sibling key `a\`. Your five cases are in the tests now, plus more around them: multi digit and out of range `#-N`, `#-0`, reverse indexes on empty arrays and non-arrays, chained reverse indexes like `$[#-1][#-1]`, escaped backslash keys, and empty and unicode keys.

mustafahasankhan:
One thing I noticed while testing: when `json_set` or `json_insert` create missing parents, an array leg like `$.a[0]` on `{}` produces `{"a":{"0":1}}`, because yyjson's `create_parent` only creates objects. SQLite creates `{"a":[1]}` here. Should numeric legs create arrays instead? Happy to implement that, or leave it as is and document it. I've kept this case out of the tests for now so we don't lock in either behavior.

lnkuiper:
Thanks for implementing the feedback! I think yes, we should match SQLite behaviour. Could you add that to this PR? There's also a small issue remaining with the json path conversion. Repro's for both:
```sql
# name: test/temp.test

require json

# Malformed JSONPath must not fall through to the empty JSON Pointer (the root)
query T
SELECT json_set('{"a":1}', '$not_a_path', '2')
----
{"a":1}

query T
SELECT json_set('{}', '$.a[0]', '1')
----
{"a":[1]}

query T
SELECT json_insert('{}', '$.a[0]', '1')
----
{"a":[1]}

query T
SELECT json_set('{}', '$.a[0].b', '1')
----
{"a":[{"b":1}]}
```

I am starting to think that these issues stem from the json path -> json pointer implementation. This essentially reimplements json path parsing, which we've already implemented in `JSONCommon`. I understand that in its current form it's not usable for json modification, but I think we can change it. We could extract the existing parser into a shared representation/iterator—e.g. path elements for object key, array index, reverse index, and append. Then we have a single source of truth for json path parsing that can be reused.

mustafahasankhan:
Done, all three:

1. Array legs now create arrays like SQLite. Your three creation repros are in the tests, along with the rest of the matrix I probed directly against SQLite 3.43: only index 0 is valid in a freshly created array, scalars are never replaced, the root's type is never converted, out of range stays a no-op. Creation is also transactional: `json_set('{}', '$.a[2]', '1')` returns `{}` without leaving an empty `{"a":[]}` behind.
2. The fall-through bug is gone. One deliberate difference from your repro: `$not_a_path` now raises the same `JSON path error` as `json_extract` instead of silently returning the document, since a silent no-op on a malformed path tends to hide typos. Happy to make it a no-op instead if you prefer.
3. Path parsing is now shared. I extracted a `JSONPathIterator` into `JSONCommon` that yields typed elements (key, index, reverse index, append, wildcard) on top of the existing `ReadKey`/`ReadArrayIndex` primitives, and rewrote `ValidatePath` and `GetPath` on it. The modification functions consume the same iterator and walk the mutable document directly, so the JSONPath to JSON Pointer conversion is gone entirely. `GetWildcardPathInternal` still uses the primitives directly because of its recursion; that can move to the iterator in a follow-up.

Side effects of sharing the parser, all locked in tests: `$[-1]` works as a reverse index like in `json_extract`, wildcards raise an error in the modification functions, and quoted key escapes now come from the existing reader. One SQLite divergence I kept on purpose: `json_set('{}', '$.a[#]', '1')` gives `{"a":[1]}` here (append behaves as index-equals-length everywhere), where SQLite returns `{}`. Easy to change if you want strict parity.

lnkuiper:
Looks great, thanks!

ericsun2:
https://github.com/duckdb/duckdb/discussions/13103

OBJECT_DELETE() can be expressed as JSON_REMOVE() now. Thanks @mustafahasankhan and @lnkuiper

PR #23790, opened 2026-07-13
TryLookupEntry now uses default schema as fallback

okruitho:
This PR fixes a bug where a reference to an default schema's macro in one catalog would not resolve in a *different* catalog.

This was because the default schema was missing in the Catalog search path.

Now, as an fallback, the default schema will be added to the path if the type could not be found.

For tables this was already the case. Now this is possible for other types as well.

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -462,13 +462,23 @@ vector<SimilarCatalogEntry> Catalog::SimilarEntriesInSchemas(ClientContext &cont
 }
 
 vector<CatalogSearchEntry> GetCatalogEntries(CatalogEntryRetriever &retriever, const string &catalog,
-                                             const string &schema) {
+                                             const string &schema, CatalogType lookup_type = CatalogType::INVALID) {
 	auto &context = retriever.GetContext();
 	vector<CatalogSearchEntry> entries;
 	auto &search_path = retriever.GetSearchPath();
 	if (IsInvalidCatalog(catalog) && IsInvalidSchema(schema)) {
-		// no catalog or schema provided - scan the entire search path
-		entries = search_path.Get();
+		// no catalog or schema provided - scan the entire search path. For non-table lookups, implicit catalogs that
+		// requested precedence (e.g. a view's own catalog while binding its body) are resolved to their default
+		// schema in place so they keep their priority; tables use the separate default-table mechanism instead.
+		//
+		// lookup_type is optional (CatalogType::INVALID means the caller did not specify one, e.g. callers that
+		// only ever pass a concrete catalog/schema); when unknown we treat it like a table lookup and skip the
+		// precedence resolution.
+		if (lookup_type == CatalogType::INVALID || lookup_type == CatalogType::TABLE_ENTRY) {
+			entries = search_path.Get();
+		} else {
+			entries = search_path.GetWithPrecedenceSchemas(context);
+		}
 	} else if (IsInvalidCatalog(catalog)) {
 		auto catalogs = search_path.GetCatalogsForSchema(schema);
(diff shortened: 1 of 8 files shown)


lnkuiper:
Looks good! I think this is already an improvement, so we can probably just merge this as-is, but do you think it would also be possible to fix this?
```sql
# name: test/temp.test
# group: [attach]

# A view's owning catalog should take precedence over the caller's search path
# for unqualified non-table entries in the catalog's default schema.
statement ok
ATTACH ':memory:' AS view_path_c1;

statement ok
CREATE SCHEMA view_path_c1.s1;

statement ok
CREATE MACRO view_path_c1.main.fn() AS 1;

statement ok
CREATE VIEW view_path_c1.s1.v AS SELECT fn() AS result;

statement ok
ATTACH ':memory:' AS view_path_c2;

statement ok
CREATE MACRO view_path_c2.main.fn() AS 2;

statement ok
SET search_path = 'view_path_c2.main';

query I
FROM view_path_c1.s1.v;
----
1

statement ok
RESET search_path;

statement ok
DETACH view_path_c2;

statement ok
DETACH view_path_c1;
```
This is related and should be a relatively small fix (I think?).

test/sql/attach/searchpath_fallback.test
```diff
@@ -0,0 +1,113 @@
...
+statement ok
+CREATE VIEW other_schema.some_view AS SELECT fn(1) AS t;
+
+statement ok
+use test2;
+
+query I
+from test.other_schema.some_view;
+----
+2
```
lnkuiper: Can we also test `RESET seach_path;`? Codex came up with this test:
```sql
# A view's owning catalog should take precedence over the caller's search path
# for unqualified non-table entries in the catalog's default schema.
statement ok
ATTACH ':memory:' AS view_path_c1;

statement ok
CREATE SCHEMA view_path_c1.s1;

statement ok
CREATE MACRO view_path_c1.main.fn() AS 1;

statement ok
ATTACH ':memory:' AS view_path_c2;

statement ok
CREATE MACRO view_path_c2.main.fn() AS 'caller';

statement ok
SET search_path = 'view_path_c2.main';

statement ok
CREATE VIEW view_path_c1.s1.v AS SELECT fn() AS result;

# Verify CREATE VIEW used the owning catalog's INTEGER-returning macro.
query T
SELECT data_type
FROM duckdb_columns()
WHERE database_name = 'view_path_c1'
  AND schema_name = 's1'
  AND table_name = 'v'
  AND column_name = 'result';
----
INTEGER

# Verify rebinding the view also uses the owning catalog.
query I
FROM view_path_c1.s1.v;
----
1

statement ok
RESET search_path;

statement ok
DETACH view_path_c2;

statement ok
DETACH view_path_c1;

```

src/catalog/catalog.cpp
```diff
@@ -918,6 +928,18 @@ CatalogEntryLookup Catalog::TryLookupEntry(CatalogEntryRetriever &retriever, con
 		}
 	}
 
+	// Special case for non-table entries (functions, macros, types): mirroring the default-table fallback above, if
+	// the entry could not be resolved through the regular search path we fall back to the default schema of any
+	// catalog flagged as an implicit search catalog (added to the path with an INVALID_SCHEMA entry), but only
+	// if the catalog did not request precedence. Catalogs that requested precedence were already searched at their
+	// position in the search path.
```
lnkuiper: The comments around this PR are very verbose, can we tone this down?

lnkuiper (changes requested):
Thanks! Looks good, just a few small comments then this is ready to go:

lnkuiper (approved):

PR #23792, opened 2026-07-14
Fix "Resource deadlock avoided" when SET threads runs on a scheduler worker

souvikDevloper:
## What this fixes

A query can crash during cleanup with `Resource deadlock avoided` (and sometimes a follow-up `Invalid argument` that aborts the process) when the global `threads` setting is changed from inside a task that is itself running on a scheduler worker.

## Why it happens

After a query runs, `ClientContext::CleanupInternal` calls `TaskScheduler::RelaunchThreads()`. `threads` is a database-global setting, so if a scheduler task opens an internal connection and runs `SET threads=...`, that cleanup ends up executing on a worker owned by the same scheduler.

`RelaunchThreads()` stops and joins every worker in the pool — including, in this case, the very thread that's calling it. A thread joining itself gives `Resource deadlock avoided`. Worse, if some workers were already joined before the self-join threw, a later retry tries to join an already-joined thread and throws `Invalid argument`. When that exception escapes during teardown, the process can abort.

The self-detection has to happen *before* acquiring `thread_lock`, because another thread may already hold that lock while waiting to join the current worker.

## The fix

- Each scheduler-owned worker records its owning scheduler in thread-local state for the lifetime of `ExecuteForever` (via a new `ExecuteForeverOnInternalThread`).
- `RelaunchThreads()` returns early if it's being called by one of its own workers. The requested thread count is still recorded, so the resize is applied later when the outer query cleans up on a normal (non-worker) thread.
- The pool join loop now checks `joinable()` before joining, so a relaunch can recover cleanly from a worker vector that a previous attempt left partially joined.

Deferred-resize semantics: a worker-originated global `threads` change no longer takes effect immediately — it lands on the next cleanup that runs off the pool. I'd appreciate a maintainer sanity-check that this is acceptable for worker-originated global settings.

## Test plan

- Confirmed the new regression test fails on the unpatched scheduler at the captured worker-side query error.
- Confirmed the changed scheduler and regression-test sources compile without diagnostics with MSVC 19.44.
- Linux RelWithDebInfo focused regression: **passed, 5 assertions in 1 test case**.
- Focused regression repetition: **passed 25/25 runs**.
- Existing `NumberOfThreads`/`RelaunchThreads` deadlock regression: **passed**.
- Three other enabled thread-setting API tests: **passed, 21 assertions**.
- `clang-format` 11.0.1 and `git diff --check`: **passed**.
- Linux under WSL (Ubuntu), RelWithDebInfo, GCC 13, Ninja: full patched `test/unittest` target built successfully.
- Windows/Visual Studio 2022, MSVC 19.44: the patched scheduler unity object/library and regression-test object/library compiled without diagnostics. The full Windows unit-test executable was not linked or run.
- New regression test on unpatched main: **failed as expected** — 1 failed test case; 2 assertions, 1 passed and 1 failed. The failure was at `REQUIRE(state->error.empty())` after worker-side `SET threads=1`.
- New regression test with the fix: **passed** — 5 assertions in 1 test case (0.086 seconds).
- Focused regression repetition with the fix: **passed 25/25 runs**.
- Existing `Test deadlock issue between NumberOfThreads and RelaunchThreads`: **passed** — 1 test case.
- Other enabled thread-setting API tests (`maximum_threads`, external threads, and async threads): **passed** — 21 assertions in 3 test cases.
- `clang-format` 11.0.1: **applied successfully** to all four modified files.
- `git diff --check`: **passed**.
- Exported patch clean-apply check against base `71f0c7a0bf10f60087b7bd42f524ae556c5a1967`: **passed**.

## Extension note

The reproducer comes from `ducklake-cdc` running `SET threads=1` on an internal connection. Since `threads` is database-global (not connection-local), that command mutates the shared scheduler regardless of this fix — the extension should drop it independently. This change only hardens core against the unsafe relaunch; it does not make `threads` connection-local.

Addresses duckdb/ducklake#1448.

### DIFF
--- a/src/include/duckdb/parallel/task_scheduler.hpp
+++ b/src/include/duckdb/parallel/task_scheduler.hpp
@@ -10,6 +10,7 @@
 
 #include "duckdb/common/atomic.hpp"
 #include "duckdb/common/common.hpp"
+#include "duckdb/common/optional_ptr.hpp"
 #include "duckdb/common/vector.hpp"
 #include "duckdb/parallel/task.hpp"
 #include "duckdb/common/array.hpp"
@@ -61,6 +62,10 @@ public:
 	//! Run tasks forever until "marker" is set to false, "marker" must remain valid until the thread is joined
 	void ExecuteForever(atomic<bool> *marker);
 	void ExecuteForever(atomic<bool> *marker, TaskSchedulerType pool_type);
+#ifndef DUCKDB_NO_THREADS
+	//! Run tasks forever on a worker owned by this scheduler
+	void ExecuteForeverOnInternalThread(atomic<bool> *marker, TaskSchedulerType pool_type);
+#endif
 	//! Run tasks until `marker` is set to false, `max_tasks` have been completed, or until there are no more tasks
 	//! available. Returns the number of tasks that were completed.
 	idx_t ExecuteTasks(atomic<bool> *marker, idx_t max_tasks);
@@ -99,6 +104,10 @@ private:
 
 private:
 	DatabaseInstance &db;
+#ifndef DUCKDB_NO_THREADS
+	//! The scheduler that owns the current internal worker thread, if any
+	static thread_local optional_ptr<TaskScheduler> current_worker_scheduler;
+#endif
 	//! Lock for modifying the thread count
 	mutex thread_lock;
 	//! The thread pools
--- a/src/parallel/task_scheduler.cpp
+++ b/src/parallel/task_scheduler.cpp
@@ -27,6 +27,10 @@
 
 namespace duckdb {
 
+#ifndef DUCKDB_NO_THREADS
+thread_local optional_ptr<TaskScheduler> TaskScheduler::current_worker_scheduler;
+#endif
+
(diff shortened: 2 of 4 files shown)


Mytherin:
Thanks but this is adding a lot of complexity to fix a very obscure issue, can we try to fix this in a less invasive manner?

souvikDevloper:
Agreed. I reduced the trigger further: `ducklake-cdc` runs `SET threads = 1` on an internal connection before the DuckLake catalog work, but `threads` is database-global, so that call resizes the scheduler shared with the outer query rather than only serializing the internal connection.

The least-invasive fix is therefore to remove that setting from the extension instead of adding scheduler-worker lifecycle tracking in core. I’ll withdraw the broad core approach and follow up extension-side. If a core defensive guard is still desirable after that, I can return with a smaller, separately motivated change and reproducer. Thanks for the direction.

souvikDevloper:
@Mytherin should i proceed?

souvikDevloper:
@Mytherin need feedback before proceeding!

PR #23793, opened 2026-07-14
Add HTTPUtil::ShouldRetry(request, response) retry-classification hook

carlopi:
HTTPResponse::ShouldRetry() classifies on the status code alone, and is not virtual. Protocols that carry the retry discriminator in the response body cannot express themselves through it: S3 answers a stalled socket with HTTP 400 and <Code>RequestTimeout</Code> in the body, which is transient, while other 400s are fatal.

Add a virtual HTTPUtil::ShouldRetry(request, response) that RunRequestWithRetry dispatches through. The default delegates to HTTPResponse::ShouldRetry(), so behavior is unchanged for every existing caller.

The request is passed so an implementation can restrict retries to idempotent methods -- replaying a POST is not generally safe. HTTPUtil::IsIdempotent() exposes that test over core's RequestType.

Both are marked DUCKDB_API: they exist for out-of-tree implementations and have no in-tree callers, so without an export annotation they would not be reachable from an extension on Windows. This is the first use of DUCKDB_API in http_util.hpp, hence the added winapi.hpp include.

Goal of this PR is allowing simplifying https://github.com/duckdb/duckdb-httpfs/pull/365 by just override the `ShouldRetry` condition for 400 when timeout.

### DIFF
--- a/src/include/duckdb/common/http_util.hpp
+++ b/src/include/duckdb/common/http_util.hpp
@@ -12,6 +12,7 @@
 #include "duckdb/common/case_insensitive_map.hpp"
 #include "duckdb/common/enums/http_status_code.hpp"
 #include "duckdb/common/types/timestamp.hpp"
+#include "duckdb/common/winapi.hpp"
 #include <functional>
 
 namespace duckdb {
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks! We should rebase https://github.com/duckdb/duckdb-httpfs/pull/365 around this

PR #23794, opened 2026-07-14
Bump extensions in preparation for v1.5.5

carlopi:


carlopi:
This will possibly have issue with function registration, in case I'll fix it later.

Mytherin:
Thanks!

PR #23796, opened 2026-07-14
Add bounds check for roaring container_count

Pyolar:
RoaringScanState computes `container_count` from `segment.count` (the on-disk `tuple_count`) without checking it against the available metadata space in the segment. The existing check at `scan.cpp:206` only validates `metadata_offset < segment_size` but doesn't verify that the metadata region `metadata_offset + GetMetadataSize(container_count)` fits within the segment.

When `tuple_count` is inflated in a crafted database file, `container_count` grows proportionally, and `Deserialize(metadata_ptr, container_count)` reads past the segment buffer — a heap over-read (CWE-125).

This adds a sanity check: if `container_count` exceeds the available metadata space (`segment_size - metadata_offset`), throw an `InternalException` before calling `Deserialize`. Each container requires at least one byte of metadata, so this is a conservative lower bound that catches inflated values.


Mytherin:
Thanks - could we add a checked in corrupt database and a test that triggers this?

Pyolar:
@Mytherin Added. The corrupt file is at `data/storage/roaring_corrupt.db.gz`.

I created a 2049-row BOOLEAN table, checkpointed it, then patched two bytes in the binary: changed compression_type to ROARING and moved metadata_offset near the end of segment_size. That makes container_count 2 while available space is only 1, so it hits the check directly.

Pyolar:
@Mytherin Could you approve the CI workflows? Pushed a fix for the test failure.

Pyolar:
@Mytherin Pushed two fixes: changed InternalException to IOException so the test doesn't crash the runner, and added `require vector_size 2048` to skip the test in non-standard vector size builds. Could you approve the CI workflows again?

Pyolar:
@Mytherin The extension build failures look like Docker timeouts (DeadlineExceeded) rather than code issues. Mind re-running those two jobs?

Pyolar:
Hi @Mytherin — bumping this since it's been a few weeks. The corrupt database + regression test you requested are in place, with the `vector_size 2048` guard and `IOException` matching the other corrupt-file checks (ALP/Patas/RLE). The only remaining CI failures were Docker timeouts, not code issues.

Also worth noting: #23891 was closed in favor of this PR, so this is the only open fix for the Roaring `container_count` bounds check.

Could you take another look when you have a moment?

Mytherin:
Thanks!

src/storage/compression/roaring/scan.cpp
```diff
@@ -213,6 +213,10 @@ RoaringScanState::RoaringScanState(ColumnSegment &segment) : segment(segment) {
 	if (segment_count % ROARING_CONTAINER_SIZE != 0) {
 		container_count++;
 	}
+	auto available_metadata_space = segment_size - metadata_offset;
+	if (container_count > available_metadata_space) {
```
dentiny, line 217: It'd be nice to use [DataCorruptionException](https://github.com/duckdb/duckdb/blob/14b0c94b3a64d4655ff92dbe957959a098bb836f/src/include/duckdb/common/exception.hpp#L253) here

Pyolar: Thanks for the suggestion! This is a good idea. Since the PR has already been merged, I'll update this to use `DataCorruptionException` in a follow-up PR.

PR #23798, opened 2026-07-14
Fix quoting rules in QualifiedName::ParseComponents

JelteF:
`QualifiedName::ParseComponents` returned different results than the normal parser when parsing a qualified names: doubled double quotes would disapear instead of being replaced by a single double quote. It also allowed quotes in the middle of a name, which the normal parser rejects. This fixes those issues and adds some tests for these cases.

### DIFF
--- a/src/parser/qualified_name.cpp
+++ b/src/parser/qualified_name.cpp
@@ -35,6 +35,10 @@ string QualifiedName::ToString(QualifiedNameToStringMode mode) const {
 	return result;
 }
 
+//! This parses a superset of the strings that the actual SQL parser accepts: it allows whitespace, most special
+//! characters like ()'- and keywords without requiring double quotes. It only requires double quotes around .
+//! characters and doubled double quotes (which collapse into a single double quote). It's only possible to fully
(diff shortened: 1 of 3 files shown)


src/parser/qualified_name.cpp
```diff
@@ -61,8 +66,18 @@ vector<Identifier> QualifiedName::ParseComponents(const string &input) {
...
+				entry += '"';
+				idx++;
+				continue;
+			}
+			//! unquote; a closing quote must end the component, e.g. "abc"xyz is not a valid identifier
 			idx++;
+			if (idx < input.size() && input[idx] != '.') {
+				throw ParserException("Unexpected character after a quoted identifier in a qualified name! (input: %s)",
+				                      input);
+			}
```
lnkuiper, line 88: Can we also reject an empty quoted component here? ""."t" currently reaches the separator with an empty entry, so it is treated as an absent schema and pragma_table_info('""."t"') resolves the unqualified table t. The SQL parser rejects "" as a zero-length delimited identifier.

PR #23799, opened 2026-07-14
Bump azure, delta, unity_catalog extensions to main/HEAD

benfleis:
- duckdb-azure: abea9d0 -> 4bc0903
- duckdb-delta: 755898c -> e74cf88
- unity_catalog: 81e41ae -> b593157

Both azure patches are already reflected in duckdb-azure main, so remove patches and APPLY_PATCHES
- 0001-identifier-explicit-casts
- 0002-pruned-transitive-logger-include


Mytherin:
Thanks!

PR #23802, opened 2026-07-14
Fix IsComplete for Decimal

evertlammerts:
The completeness check for Decimal had to be negated. IsComplete returns a negated `return !TypeVisitor::Contains`, so a complete type in the TypeVisitor must return `false`.

### DIFF
--- a/src/common/types.cpp
+++ b/src/common/types.cpp
@@ -719,9 +719,13 @@ bool LogicalType::IsComplete() const {
 				return false;
 			}
 			return type.AuxInfo()->Cast<StructTypeInfo>().child_types.empty();
-		case ExtraTypeInfoType::DECIMAL_TYPE_INFO:
-			return DecimalType::GetWidth(type) >= 1 && DecimalType::GetWidth(type) <= Decimal::MAX_WIDTH_DECIMAL &&
-			       DecimalType::GetScale(type) <= DecimalType::GetWidth(type);
+		case ExtraTypeInfoType::DECIMAL_TYPE_INFO: {
+			// A well-formed decimal is not incomplete
+			auto is_well_formed = DecimalType::GetWidth(type) >= 1 &&
+			                      DecimalType::GetWidth(type) <= Decimal::MAX_WIDTH_DECIMAL &&
+			                      DecimalType::GetScale(type) <= DecimalType::GetWidth(type);
+			return !is_well_formed;
+		}
 		default:
 			return false; // Nested types are checked by TypeVisitor recursion
 		}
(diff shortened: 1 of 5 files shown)


Tishj:
```
/Users/thijs/DuckDBLabs/duckdb/test/common/test_logical_type_is_complete.cpp:27:35: error: argument type 'optional_idx' is incomplete
   27 |                 REQUIRE(LogicalType::ARRAY(dec, 4).IsComplete());
      |                                                 ^
/Users/thijs/DuckDBLabs/duckdb/src/include/duckdb/common/exception.hpp:27:7: note: forward declaration of 'duckdb::optional_idx'
   27 | class optional_idx; // NOLINT: matching std style
      |       ^
/Users/thijs/DuckDBLabs/duckdb/test/common/test_logical_type_is_complete.cpp:27:35: error: argument type 'optional_idx' is incomplete
   27 |                 REQUIRE(LogicalType::ARRAY(dec, 4).IsComplete());
      |                                                 ^
/Users/thijs/DuckDBLabs/duckdb/src/include/duckdb/common/exception.hpp:27:7: note: forward declaration of 'duckdb::optional_idx'
   27 | class optional_idx; // NOLINT: matching std style
      |       ^
/Users/thijs/DuckDBLabs/duckdb/test/common/test_logical_type_is_complete.cpp:27:35: error: argument type 'optional_idx' is incomplete
   27 |                 REQUIRE(LogicalType::ARRAY(dec, 4).IsComplete());
      |                                                 ^
/Users/thijs/DuckDBLabs/duckdb/src/include/duckdb/common/exception.hpp:27:7: note: forward declaration of 'duckdb::optional_idx'
   27 | class optional_idx; // NOLINT: matching std style
```

I think something slipped through the cracks?
This was from a `DISABLE_UNITY=1 make debug` build, which shows that these tests are missing includes

lnkuiper:
I think CI is no longer doing proper non-unity builds. Windows builds used to fail immediately when we did something that would violate a non-unity build in the past, but seems like Windows happily builds now (both src and test) even if we're missing an include somewhere.

evertlammerts:
Hm, that does seem to be the problem. I had a look through the workflows and Makefile but it's not that easy to trace. CC @smvv 

I'll put up a PR to add any missing includes.

smvv:
hi @lnkuiper @evertlammerts, the non-unity build has been temporarily disabled because clang**d**-tidy ran out of memory and hanged in CI. That job used `DISABLE_UNITY=1` and would have caught these missing includes. (This [issue](https://github.com/duckdblabs/duckdb-internal/issues/9770) tracks re-enabling clangd-tidy.)

PR #23803, opened 2026-07-14
Fix for malformed JSON when rendering via duckbox

sebastiaan-dev:
Resolves https://github.com/duckdblabs/duckdb-internal/issues/9652. 

Instead of throwing an error when encountering malformed JSON we do not apply any highlighting to the output.


lnkuiper:
Could you have a look at the failing tidy check?

PR #23806, opened 2026-07-14
Ensure NULL children when parent is NULL on Copy paths

evertlammerts:
This PR:
- Broadcasts NULL rows into children in `ColumnDataCopyStruct` the same way that `ColumnDataCopyArray` does.
- Also ensures that `VectorBuffer::Copy` won't read undefined children of `NULL` ARRAY/STRUCT rows. This already works for other vector types (string and list vectors).
- Fixes the `FIXME: verify NULL-ness in child` in `VectorArrayBuffer::VerifyInternal`

### DIFF
--- a/src/common/types/column/column_data_collection.cpp
+++ b/src/common/types/column/column_data_collection.cpp
@@ -886,14 +886,16 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
 	// copy the NULL values for the main struct vector
 	TemplatedColumnDataCopy<StructValueCopy>(meta_data, source_data, source, offset, copy_count);
 
-	// check if the copied range contains any NULL rows
+	// check if the copied range contains any NULL rows, children must not claim valid entries under them
+	auto &parent_validity = source_data.validity;
+	const bool parent_flat = !source_data.sel->IsSet();
 	bool parent_has_nulls = false;
-	if (source_data.validity.IsMaskSet()) {
-		for (idx_t i = 0; i < copy_count; i++) {
-			auto source_idx = source_data.sel->get_index(offset + i);
-			if (!source_data.validity.RowIsValidUnsafe(source_idx)) {
-				parent_has_nulls = true;
-				break;
+	if (parent_validity.IsMaskSet()) {
+		if (parent_flat) {
+			parent_has_nulls = !parent_validity.CheckAllValid(offset + copy_count, offset);
+		} else {
+			for (idx_t i = 0; i < copy_count && !parent_has_nulls; i++) {
+				parent_has_nulls = !parent_validity.RowIsValidUnsafe(source_data.sel->get_index(offset + i));
 			}
 		}
 	}
@@ -912,14 +914,28 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
 
 		if (parent_has_nulls) {
 			D_ASSERT(!child_data.sel->IsSet());
-			// Broadcast and sync the validity of the struct vector to the child vector
-			// This requires creating a copy of the validity mask: we cannot modify the input validity
-			child_data.validity = ValidityMask(child_data.validity, child_data.validity.Capacity());
-			child_data.validity.EnsureWritable();
-			for (idx_t i = 0; i < copy_count; i++) {
-				auto source_idx = source_data.sel->get_index(offset + i);
-				if (!source_data.validity.RowIsValidUnsafe(source_idx)) {
-					child_data.validity.SetInvalidUnsafe(offset + i);
+			// only rows violating the child-NULL invariant need a patched validity mask
+			bool children_correctly_masked = true;
+			if (parent_flat) {
+				// parent and child validity are at identical positions
+				children_correctly_masked =
+				    !AnyChildValidUnderNullParent(parent_validity, child_data.validity, offset, copy_count);
+			} else {
+				for (idx_t i = 0; i < copy_count && children_correctly_masked; i++) {
+					auto source_idx = source_data.sel->get_index(offset + i);
+					children_correctly_masked =
+					    parent_validity.RowIsValidUnsafe(source_idx) && child_data.validity.RowIsValid(offset + i);
+				}
+			}
+			if (!children_correctly_masked) {
+				// requires a copy of the validity mask: we cannot modify the input validity
+				child_data.validity = ValidityMask(child_data.validity, child_data.validity.Capacity());
+				child_data.validity.EnsureWritable();
+				for (idx_t i = 0; i < copy_count; i++) {
+					auto source_idx = source_data.sel->get_index(offset + i);
+					if (!parent_validity.RowIsValidUnsafe(source_idx)) {
+						child_data.validity.SetInvalidUnsafe(offset + i);
+					}
 				}
 			}
 		}
--- a/src/common/types/column/column_data_collection.cpp
+++ b/src/common/types/column/column_data_collection.cpp
@@ -879,6 +879,19 @@ void ColumnDataCopy<list_entry_t>(ColumnDataMetaData &meta_data, const UnifiedVe
 	}
 }
 
+// Independent flat copy carrying patched_validity, so SetValidity does not mutate child's possibly-shared buffer.
+static unique_ptr<Vector> PatchedListChildCopy(const Vector &child, const ValidityMask &patched_validity) {
+	auto count = child.size();
+	SelectionVector identity(count);
+	for (idx_t i = 0; i < count; i++) {
+		identity.set_index(i, i);
+	}
+	auto copy = make_uniq<Vector>(child, identity, count);
+	copy->Flatten();
+	FlatVector::SetValidity(*copy, patched_validity);
+	return copy;
+}
+
 void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorFormat &source_data, Vector &source,
                           idx_t offset, idx_t copy_count) {
 	auto &segment = meta_data.segment;
@@ -912,8 +925,7 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
 		UnifiedVectorFormat child_data;
 		child_vectors[child_idx].ToUnifiedFormat(child_data);
 
-		// child_source is handed to the copier and replaced by a patched copy on the violation path.
-		// note: patched_child must outlive the child_function call below, hence the outer scope.
+		// patched_child (if built below) must outlive the child_function call, hence the outer scope
 		reference<Vector> child_source = child_vectors[child_idx];
 		unique_ptr<Vector> patched_child;
 
@@ -942,13 +954,10 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
 						child_data.validity.SetInvalidUnsafe(offset + i);
 					}
 				}
-				// Only a LIST child's element copy reads validity from the Vector (via
-				// GetConsecutiveChildListInfo), so it must carry the patched mask too; other child types read
-				// only source_data. SetValidity needs a flat buffer, so constant/dict children are left as-is.
+				// only a flat LIST child reads validity off the Vector (GetConsecutiveChildListInfo); patch a copy
 				if (child_vectors[child_idx].GetType().InternalType() == PhysicalType::LIST &&
 				    child_vectors[child_idx].GetVectorType() == VectorType::FLAT_VECTOR) {
-					patched_child = make_uniq<Vector>(Vector::Ref(child_vectors[child_idx]));
-					FlatVector::SetValidity(*patched_child, child_data.validity);
+					patched_child = PatchedListChildCopy(child_vectors[child_idx], child_data.validity);
 					child_source = *patched_child;
 				}
 			}
@@ -987,8 +996,7 @@ void ColumnDataCopyArray(ColumnDataMetaData &meta_data, const UnifiedVectorForma
 	ColumnDataMetaData child_meta_data(child_function, meta_data, child_index);
 	child_vector.ToUnifiedFormat(child_vector_data);
 
-	// child_source is handed to the copier and replaced by a patched copy on the violation path.
-	// note: patched_child must outlive the child_function call below, hence the outer scope.
+	// patched_child (if built below) must outlive the child_function call, hence the outer scope
 	reference<Vector> array_child_source = child_vector;
 	unique_ptr<Vector> patched_child;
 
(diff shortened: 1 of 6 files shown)


src/common/types/column/column_data_collection.cpp
```diff
@@ -886,6 +886,18 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
 	// copy the NULL values for the main struct vector
 	TemplatedColumnDataCopy<StructValueCopy>(meta_data, source_data, source, offset, copy_count);
 
+	// check if the copied range contains any NULL rows
+	bool parent_has_nulls = false;
+	if (source_data.validity.IsMaskSet()) {
+		for (idx_t i = 0; i < copy_count; i++) {
+			auto source_idx = source_data.sel->get_index(offset + i);
+			if (!source_data.validity.RowIsValid(source_idx)) {
```
lnkuiper: Nit: mask is guaranteed to be set so this can be `RowIsValidUnsafe`

src/common/types/column/column_data_collection.cpp
```diff
@@ -898,6 +910,19 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
...
 		child_vectors[child_idx].ToUnifiedFormat(child_data);
 
+		if (parent_has_nulls) {
+			D_ASSERT(!child_data.sel->IsSet());
+			// Broadcast and sync the validity of the struct vector to the child vector
+			// This requires creating a copy of the validity mask: we cannot modify the input validity
+			child_data.validity = ValidityMask(child_data.validity, child_data.validity.Capacity());
+			for (idx_t i = 0; i < copy_count; i++) {
+				auto source_idx = source_data.sel->get_index(offset + i);
+				if (!source_data.validity.RowIsValid(source_idx)) {
```
lnkuiper: Nit: `RowIsValidUnsafe`

src/common/types/column/column_data_collection.cpp
```diff
@@ -898,6 +910,19 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
 		UnifiedVectorFormat child_data;
 		child_vectors[child_idx].ToUnifiedFormat(child_data);
 
+		if (parent_has_nulls) {
+			D_ASSERT(!child_data.sel->IsSet());
+			// Broadcast and sync the validity of the struct vector to the child vector
+			// This requires creating a copy of the validity mask: we cannot modify the input validity
+			child_data.validity = ValidityMask(child_data.validity, child_data.validity.Capacity());
```
lnkuiper: If you do `child_data.validity.EnsureWritable();` you can do `SetInvalidUnsafe` on this mask below.

src/common/vector/struct_vector.cpp
```diff
@@ -254,9 +254,29 @@ void VectorStructBuffer::CopyInternal(const Vector &source, const SelectionVecto
                                       idx_t source_offset, idx_t target_offset, idx_t copy_count) {
 	auto &source_children = StructVector::GetEntries(source);
 	D_ASSERT(source_children.size() == children.size());
-	for (idx_t i = 0; i < source_children.size(); i++) {
-		children[i].Copy(source_children[i], source_sel, source_count, source_offset, target_offset, copy_count);
+	// Copy only non-NULL rows to avoid copying garbage if the child entry for some reason wasn't NULLed
```
lnkuiper: We can keep the original `Copy` fast path if the target does not have NULL values yet

lnkuiper (commented):
Thanks! Looks good, definitely needed. I think we might be able to slightly reduce the impact this has on performance.

evertlammerts:
Regression is due to this PR, looking into it

evertlammerts:
The performance regression happened because we couldn't take the fast path if the parent contains NULLs. So in columns with lots of nulls we copied small / single row slices in a loop.

Basically, I've tried to make the check more detailed by not just checking the parent for NULLs, but also checking whether the children have NULLs in the same places. `AnyChildValidUnderNullParent(parent, child, offset, count)` (added to validity_mask.hpp) returns true if a NULL row in a parent is not marked NULL in the given child for `BITS_PER_VALUE` at a time (more or less `parent_null & child_valid`).

The STRUCT vector check in `CopyInternal` became a bit hard to grasp so I wrapped it in a small struct type, `ChildNullReflection`. The flow for struct vectors is more or less:
- If the parent has no NULL rows we take the fast copy path.
- Otherwise, we scan the child validity vectors once through `ChildNullReflection::Check(...)` if we can (basically only for flat and constant vectors).
- Then if for every NULL row of a parent, all its children also have a NULL correctly set (`ChildNullReflection::AllReflected()`) we also take the fast path
- Otherwise we need to check per NULL row: `ChildNullReflection::RowReflectsNull(source_row)`, copy the non-NULL slices and `SetNull` the children of the violating rows.

`ColumnDataCopyStruct` also regressed, so that uses `AnyChildValidUnderNullParent` as well.

For ARRAY vectors we have to do the per-row check, since the masks don't align.

src/common/types/column/column_data_collection.cpp
```diff
@@ -898,6 +912,34 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
...
+				for (idx_t i = 0; i < copy_count; i++) {
+					auto source_idx = source_data.sel->get_index(offset + i);
+					if (!parent_validity.RowIsValidUnsafe(source_idx)) {
+						child_data.validity.SetInvalidUnsafe(offset + i);
+					}
+				}
+			}
+		}
+
 		child_function.function(child_meta_data, child_data, child_vectors[child_idx], offset, copy_count);
```
lnkuiper: this patches `child_data.validity`, but still passes the original `child_vectors[child_idx]` to the child copier.

src/common/vector/array_vector.cpp
```diff
@@ -67,6 +67,23 @@ idx_t VectorArrayBuffer::GetAllocationSize() const {
 
 void VectorArrayBuffer::VerifyInternal(const LogicalType &type, const SelectionVector &sel, idx_t count) const {
 	D_ASSERT(type.InternalType() == PhysicalType::ARRAY);
+	// a child slot under a NULL array row must be NULL as well, we check before reading any child payload
+	if (child->GetVectorType() == VectorType::FLAT_VECTOR || child->GetVectorType() == VectorType::CONSTANT_VECTOR) {
```
lnkuiper: This leaves the `ARRAY` child-`NULL` verification incomplete for dictionary children.

src/common/vector/array_vector.cpp
```diff
@@ -174,16 +190,59 @@ void VectorArrayBuffer::CopyInternal(const Vector &source, const SelectionVector
...
+	// no NULL rows in the copied range - one wholesale copy
+	if (validity.CheckAllValid(target_offset + copy_count, target_offset)) {
+		copy_valid_slice(0, copy_count);
+		return;
+	}
+	// NULL rows whose source slots are all NULL are handled by validity propagation in the wholesale copy
+	const auto child_vector_type = source_child.GetVectorType();
+	const bool child_checkable =
+	    child_vector_type == VectorType::FLAT_VECTOR || child_vector_type == VectorType::CONSTANT_VECTOR;
+	auto source_child_validity = source_child.Validity();
```
lnkuiper: `source_child.Validity()` is constructed even when `child_checkable` is false. For some vector types, `ToUnifiedFormat()` flattens the vector in place. I think this could cause problems.

Could the validity iterator be constructed only inside the checkable path?

lnkuiper (changes requested):
Thanks for the changes! Looks better now. I've left some more comments below.

This exposes a validity issue: 
```diff
--- a/test/common/test_nested_copy_null_child.cpp
+++ b/test/common/test_nested_copy_null_child.cpp
@@ -385,6 +385,19 @@ TEST_CASE("Verify rejects non-NULL children under a NULL ARRAY row", "[copy]") {
                REQUIRE_THROWS(v.Verify(sel, row_count));
        }
 
+       SECTION("dictionary child") {
+               auto v = make_array(LogicalType::VARCHAR);
+               auto &child = ArrayVector::GetChildMutable(v);
+               SelectionVector dict_sel(child.size());
+               for (idx_t i = 0; i < child.size(); i++) {
+                       dict_sel.set_index(i, 0);
+               }
+               child.Slice(dict_sel, child.size());
+               FlatVector::ValidityMutable(v).SetInvalid(null_row);
+               VerifyVectorsGuard guard;
+               REQUIRE_THROWS(v.Verify());
+       }
+
        SECTION("grandchild under a correctly-NULL struct is caught by recursion (full path)") {
                auto v = make_array(StructOfVarchar());
                auto &struct_child = ArrayVector::GetChildMutable(v);
@@ -518,6 +531,31 @@ TEST_CASE("ColumnDataCollection append must not read child payloads under NULL r
                }
        }
 
+       SECTION("STRUCT(LIST(VARCHAR)): a child under a NULL parent does not shift later lists") {
+               auto type = MakeType(Shape::STRUCT_LIST_VARCHAR);
+               const idx_t row_count = 3;
+               const idx_t null_row = 1;
+               DataChunk chunk;
+               chunk.Initialize(allocator, {type});
+               chunk.SetChildCardinality(row_count);
+               duckdb::vector<Value> expected(row_count);
+               for (idx_t r = 0; r < row_count; r++) {
+                       expected[r] = ValidValue(Shape::STRUCT_LIST_VARCHAR, r);
+                       chunk.data[0].SetValue(r, expected[r]);
+               }
+               // Leave the LIST child valid while marking only its STRUCT parent NULL.
+               FlatVector::ValidityMutable(chunk.data[0]).SetInvalid(null_row);
+
+               ColumnDataCollection collection(allocator, {type});
+               collection.Append(chunk);
+
+               DataChunk result;
+               ScanSingleChunk(collection, result);
+               REQUIRE(result.data[0].GetValue(0) == expected[0]);
+               REQUIRE(result.data[0].GetValue(null_row).IsNull());
+               REQUIRE(result.data[0].GetValue(2) == expected[2]);
+       }
+
        SECTION("dictionary child sharing a slot with a NULL row is not corrupted") {
                auto type = StructOfVarchar();
                DataChunk chunk;
```

src/common/types/column/column_data_collection.cpp
evertlammerts: I think it's correct this way. We pass in `child_data` (which is `child_vectors[child_idx].ToUnifiedFormat(child_data)`) as patched `source_data` and `child_vectors[child_idx]` as `source`. Only `source_data` is used for the copy, `source` is only used to traverse the nested children en get their types.

src/common/vector/array_vector.cpp
evertlammerts: I thought getting the UnifiedFormat for a dictionary vector would flatten it, but that's not true. I added it to the two VerifyInternal functions and to ARRAY's copy.

src/common/types/column/column_data_collection.cpp
evertlammerts: Nope, that was wrong :-) This was exactly the source of the bug. Should be fixed now.

src/common/types/column/column_data_collection.cpp
```diff
@@ -898,7 +912,49 @@ void ColumnDataCopyStruct(ColumnDataMetaData &meta_data, const UnifiedVectorForm
...
+					if (!parent_validity.RowIsValidUnsafe(source_idx)) {
+						child_data.validity.SetInvalidUnsafe(offset + i);
+					}
+				}
+				// Only a LIST child's element copy reads validity from the Vector (via
+				// GetConsecutiveChildListInfo), so it must carry the patched mask too; other child types read
+				// only source_data. SetValidity needs a flat buffer, so constant/dict children are left as-is.
+				if (child_vectors[child_idx].GetType().InternalType() == PhysicalType::LIST &&
+				    child_vectors[child_idx].GetVectorType() == VectorType::FLAT_VECTOR) {
+					patched_child = make_uniq<Vector>(Vector::Ref(child_vectors[child_idx]));
```
lnkuiper: the patched `LIST` validity mutates the original shared buffer

src/common/vector/array_vector.cpp
lnkuiper: sliced `ARRAY` verification does not recursively check descendants beneath `NULL` rows, not blocking, but might be nice to add

lnkuiper (commented):
Thanks for the changes! Almost ready now. I've left two comments below.

Repros for both comments:
```diff
diff --git a/test/common/test_nested_copy_null_child.cpp b/test/common/test_nested_copy_null_child.cpp
index 8c07ea303b..16dc48775c 100644
--- a/test/common/test_nested_copy_null_child.cpp
+++ b/test/common/test_nested_copy_null_child.cpp
@@ -437,6 +437,31 @@ TEST_CASE("Verify rejects non-NULL children under a NULL ARRAY row", "[copy]") {
 		// the struct's own full-size NULL scan catches it even though the array sel excludes NULL rows
 		REQUIRE(message.find("Struct NULL mismatch") != string::npos);
 	}
+
+	SECTION("grandchild under a correctly-NULL array is caught by recursion (sliced path)") {
+		auto inner_type = LogicalType::ARRAY(LogicalType::VARCHAR, ARRAY_SIZE);
+		Vector v(LogicalType::ARRAY(inner_type, ARRAY_SIZE));
+		for (idx_t r = 0; r < row_count; r++) {
+			duckdb::vector<Value> inner_arrays;
+			for (idx_t e = 0; e < ARRAY_SIZE; e++) {
+				duckdb::vector<Value> leaves(ARRAY_SIZE, Value("ok"));
+				inner_arrays.push_back(Value::ARRAY(LogicalType::VARCHAR, leaves));
+			}
+			v.SetValue(r, Value::ARRAY(inner_type, inner_arrays));
+		}
+		FlatVector::SetSize(v, row_count);
+		auto &inner_child = ArrayVector::GetChildMutable(v);
+		FlatVector::ValidityMutable(v).SetInvalid(null_row);
+		for (idx_t e = 0; e < ARRAY_SIZE; e++) {
+			FlatVector::ValidityMutable(inner_child).SetInvalid(null_row * ARRAY_SIZE + e);
+		}
+		SelectionVector sel(row_count);
+		for (idx_t i = 0; i < row_count; i++) {
+			sel.set_index(i, i);
+		}
+		VerifyVectorsGuard guard;
+		REQUIRE_THROWS(v.Verify(sel, row_count));
+	}
 }
 
 namespace {
@@ -606,4 +631,28 @@ TEST_CASE("ColumnDataCollection append must not read child payloads under NULL r
 		REQUIRE(result.data[0].GetValue(1).IsNull());
 		REQUIRE(!FlatVector::Validity(StructVector::GetEntries(result.data[0])[0]).RowIsValid(1));
 	}
+
+	SECTION("patching a STRUCT(LIST) child does not change an aliased column") {
+		auto list_type = LogicalType::LIST(LogicalType::VARCHAR);
+		auto struct_type = LogicalType::STRUCT({{"l", list_type}});
+		const idx_t row_count = 2;
+		DataChunk chunk;
+		chunk.Initialize(allocator, {struct_type, list_type});
+		chunk.SetChildCardinality(row_count);
+		auto first = LeafList(0, 1);
+		auto second = LeafList(1, 1);
+		chunk.data[1].SetValue(0, first);
+		chunk.data[1].SetValue(1, second);
+		StructVector::GetEntries(chunk.data[0])[0].Reference(chunk.data[1]);
+		FlatVector::ValidityMutable(chunk.data[0]).SetInvalid(1);
+
+		ColumnDataCollection collection(allocator, {struct_type, list_type});
+		collection.Append(chunk);
+
+		DataChunk result;
+		ScanSingleChunk(collection, result);
+		REQUIRE(result.data[0].GetValue(1).IsNull());
+		REQUIRE(result.data[1].GetValue(0) == first);
+		REQUIRE(result.data[1].GetValue(1) == second);
+	}
 }
```

src/common/types/column/column_data_collection.cpp
evertlammerts: Another nice find, should be fixed by copying the validity in PatchedListChildCopy.

src/common/vector/array_vector.cpp
evertlammerts: Ack, `VectorArrayBuffer::VerifyInternal` now recurses into children under NULL rows as well

lnkuiper:
Still one more issue:
```diff
diff --git a/test/common/test_nested_copy_null_child.cpp b/test/common/test_nested_copy_null_child.cpp
index 34e7bd5238..b087ce8d54 100644
--- a/test/common/test_nested_copy_null_child.cpp
+++ b/test/common/test_nested_copy_null_child.cpp
@@ -331,6 +331,32 @@ struct VerifyVectorsGuard {
 
 } // namespace
 
+TEST_CASE("copy STRUCT with a sequence child, a NULL row, and a selection", "[copy]") {
+       const idx_t row_count = 5;
+       const idx_t null_row = 1;
+       auto type = LogicalType::STRUCT({{"i", LogicalType::BIGINT}});
+       Vector source(type);
+       StructVector::GetEntries(source)[0].Sequence(10, 1, row_count);
+       FlatVector::SetSize(source, row_count);
+       FlatVector::ValidityMutable(source).SetInvalid(null_row);
+       SelectionVector reverse(row_count);
+       for (idx_t r = 0; r < row_count; r++) {
+               reverse.set_index(r, row_count - r - 1);
+       }
+
+       Vector target(type);
+       target.Copy(source, reverse, row_count, 0, 0, row_count);
+       FlatVector::SetSize(target, row_count);
+       for (idx_t r = 0; r < row_count; r++) {
+               auto source_row = row_count - r - 1;
+               if (source_row == null_row) {
+                       REQUIRE(target.GetValue(r).IsNull());
+               } else {
+                       REQUIRE(target.GetValue(r) == source.GetValue(source_row));
+               }
+       }
+}
+
 TEST_CASE("Verify rejects non-NULL children under a NULL ARRAY row", "[copy]") {
 #ifdef DUCKDB_CRASH_ON_ASSERT
        // the provoked InternalException aborts instead of throwing in crash-on-assert builds
```

evertlammerts:
> Still one more issue:
> 

This should now be addressed as well...

PR #23808, opened 2026-07-14
[DependencyManager] Fix ALTER dependency preservation

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9943

We had a `dependencies` in `CreateInfo` and also in `BoundCreateTableInfo`.
The `BoundCreateTableInfo::dependencies` was used by the `DependencyManager`, but the `CreateInfo::dependencies` is what ended up in the `TableCatalogEntry` and got serialized to disk.

When we replay from the WAL (write-ahead-log) we used this dependencies set to recreate the dependencies in the `DependencyManager`, which caused us to look for entries that were already removed (in this case the sequence entry).

I've removed the `BoundCreateTableInfo::dependencies`, the `CreateInfo::dependencies` is now the single point of truth.
ALTER statements now imprint the `AlterInfo::new_dependencies` with the full set of dependencies the table has after the alter is applied.


Mytherin:
Thanks!

PR #23811, opened 2026-07-14
Fix: reset empty_range in the TIMESTAMP range() table function

wmTJc9IK0Q:
The TIMESTAMP/INTERVAL overload of the range() table function silently drops rows when used in a correlated LATERAL and at least one input row produces an empty range (start >= stop).

RangeDateTimeLocalState is reused across input rows, but GenerateRangeDateTimeParameters only ever SETS empty_range to true and never resets it to false. Once one row yields an empty range, the flag sticks and every subsequent row is treated as empty and emits nothing.

The integer overload (GenerateRangeParameters) already resets the flag, which is why only the timestamp overload was affected -- an identical LATERAL over range(BIGINT, BIGINT, BIGINT) returns the correct rows.

Reproduction (10 rows, only i=5 yields an empty range, so 9 rows are expected):

    CREATE TABLE t AS
      SELECT i, TIMESTAMP '2024-01-01' + INTERVAL (i * 60) SECOND AS ts
      FROM generate_series(0, 9) s(i);

    SELECT COUNT(*) FROM t, LATERAL range(
        ts + (CASE WHEN i = 5 THEN INTERVAL '7 minutes' ELSE INTERVAL '0 minutes' END),
        ts + INTERVAL '5 minutes',
        INTERVAL '10 minutes');
    -- returns 1 (and the surviving row is i=3), expected 9

The corruption is not a truncation: which rows survive depends on where the empty range falls, so the result is wrong in a way that is hard to notice. The scalar range() (returning a LIST) is correct when unnested, so two spellings of the same query disagree.

Adds test/sql/table_function/range_timestamp_lateral_empty.test, which fails before this change and passes after.

### DIFF
--- a/src/function/table/range.cpp
+++ b/src/function/table/range.cpp
@@ -283,6 +283,14 @@ template <bool GENERATE_SERIES>
 static void GenerateRangeDateTimeParameters(DataChunk &input, idx_t row_id, RangeDateTimeLocalState &result) {
 	input.Flatten();
 
+	// The local state is reused for every input row, so empty_range must be reset
+	// here. Without this it is only ever set to true and never back to false, so
+	// once ONE row produces an empty range every subsequent row is treated as
+	// empty and silently emits nothing. The integer overload
+	// (GenerateRangeParameters) already resets it, which is why only the
+	// timestamp overload was affected.
+	result.empty_range = false;
+
 	for (idx_t c = 0; c < input.ColumnCount(); c++) {
 		if (FlatVector::IsNull(input.data[c], row_id)) {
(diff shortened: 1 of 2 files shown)


wmTJc9IK0Q:
@lnkuiper thanks for the merge. Do I need to open a separate PR to get this to land for the 1.5.5 release?

lnkuiper:
Yes. This seems OK for the v1.5.5 release, so if you'd want that you'd need to open a separate PR. I can't guarantee it will go in because we're getting close to building the release binaries

PR #23825, opened 2026-07-15
Expression filter handle NOT IN NULL

### LINKED ISSUE #23824: x NOT IN NULL performs sequential scan
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE t AS SELECT i FROM range(10000000) r(i);

-- i NOT IN NULL is always NULL, so we don't need to do any scan for the query.
memory D EXPLAIN SELECT count(*) FROM t WHERE i NOT IN (NULL);
│ ~1 row                     │
╰──────────────┬─────────────╯
╭─ Seq Scan ───┴─────────────╮
│ Table: memory.main.t       │
│ Filters: NOT (i IN (NULL)) │
│ ~2,000,000 rows            │
```
As a comparison, `X IN NULL` is handled properly and eliminate the whole scan.
```sql
memory D EXPLAIN SELECT count(*) FROM t WHERE i IN (NULL);
╭─ Ungrouped Aggregate ────╮
│ Aggregates: count_star() │
│ ~1 row                   │
╰─────────────┬────────────╯
╭─ Empty Result ───────────╮
╰──────────────────────────╯
```

### OS:

macos

### DuckDB Version:

da3d58bbb1

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23824

### DIFF
--- a/src/planner/filter/expression_filter.cpp
+++ b/src/planner/filter/expression_filter.cpp
@@ -571,6 +571,9 @@ static FilterPropagateResult CheckConjunctionStatistics(optional_ptr<ClientConte
 			if (prune_result == FilterPropagateResult::FILTER_ALWAYS_FALSE) {
 				return FilterPropagateResult::FILTER_ALWAYS_FALSE;
 			}
+			if (prune_result == FilterPropagateResult::FILTER_FALSE_OR_NULL) {
+				return FilterPropagateResult::FILTER_FALSE_OR_NULL;
+			}
 			if (prune_result != result) {
 				result = FilterPropagateResult::NO_PRUNING_POSSIBLE;
 			}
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks, looks correct/good! This is still missing some cases where we could do an `EMPTY_RESULT`:
```sql
# name: test/temp.test
# description: Reproduce NOT IN NULL pruning under conjunction
# group: [optimizer]

statement ok
CREATE TABLE integers(i INTEGER);

statement ok
INSERT INTO integers VALUES (1), (2), (NULL);

query II
EXPLAIN SELECT count(*) FROM integers WHERE i NOT IN (NULL) AND i > 0;
----
physical_plan	<REGEX>:.*Empty Result.*
```

Could we add that to this PR so it's more complete?

PR #23826, opened 2026-07-15
Parquet: validate bloom filter payload length

wufengwind:
This validates Parquet bloom filter payload lengths before using them for row group pruning.

A corrupted Parquet file can declare a `BloomFilterHeader.numBytes` value that is inconsistent with the bloom filter metadata range. The current scan path can then read unrelated bytes as bloom filter data and incorrectly conclude that a row group cannot match.

The fix treats invalid bloom filter metadata as unusable for pruning, so the scan still reads the row group normally. It also makes the fallback prefetch path fall back to a regular read when the fallback buffer does not cover the requested range.

Tests add a corrupted Parquet sample whose first bloom filter header has an inflated payload length, and verify that filtering still returns the matching rows instead of incorrectly returning zero.

Tests:
- `cmake --build build/release --target shell unittest -j 8`
- `build/release/test/unittest test/sql/copy/parquet/parquet_bloom_filter_corruption.test`
- `build/release/test/unittest test/sql/copy/parquet/bloom_filters.test`


lnkuiper:
Thanks! Looks pretty good already, but there's still cases of invalid pruning. Could we fix that in this PR too?
```sql
# name: test/temp.test

require parquet

# PR 23826: an out-of-file bloom_filter_length should disable bloom pruning rather than fail before validation.
query I
SELECT count(*)
FROM read_parquet('/tmp/bloom_filter_bad_length_range.parquet')
WHERE r = 0;
----
1000

# PR 23826: a shorter block-aligned numBytes is still inconsistent with bloom_filter_length and must not prune.
query I
SELECT count(*)
FROM read_parquet('/tmp/bloom_short_numbytes.parquet')
WHERE r = 200;
----
100
```
Files:
[bloom_filter_bad_length_range.parquet.zip](https://github.com/user-attachments/files/30041908/bloom_filter_bad_length_range.parquet.zip)
[bloom_short_numbytes.parquet.zip](https://github.com/user-attachments/files/30041909/bloom_short_numbytes.parquet.zip)

extension/parquet/parquet_statistics.cpp
lnkuiper: This can also prefetch from an invalid location, right?

wufengwind:
Thanks! Both cases are now handled.

The declared bloom filter range is validated against the file size before calling Prefetch, so an out-of-file bloom_filter_length disables bloom pruning instead of attempting an invalid read.

After parsing the header, the code now also requires header_size + numBytes to exactly match bloom_filter_length. A shorter block-aligned payload is therefore rejected for pruning as well.

I added both provided files to the corruption regression test. They now return 1000 and 100 respectively. The focused bloom filter tests and the full test/sql/copy/parquet suite pass locally.

extension/parquet/parquet_statistics.cpp
wufengwind: Yes, good catch. The range is now validated against the file size before SetLocation and Prefetch are called.

PR #23827, opened 2026-07-15
Storage: validate dictionary bitpacking width

wufengwind:
Dictionary string scans trust the bitpacking width stored in dictionary-compressed string segment headers. The width is derived from the dictionary index count when the segment is written, but the scan path used the stored value without validating that relationship.

A checksum-valid corrupted segment can change the stored width and make scans decode the selection buffer with the wrong layout. In release builds this can produce incorrect strings instead of reporting storage corruption.

This PR validates the stored width against the expected minimum bit width for the dictionary index count during scan initialization, and reports invalid metadata as an `IOException`.

A sqllogictest is added with a checksum-valid corrupted dictionary-compressed database file covering an invalid bitpacking width.


src/storage/compression/dictionary/decompression.cpp
```diff
@@ -85,6 +86,12 @@ void CompressedStringScanState::Initialize(ColumnSegment &segment, bool initiali
 		throw IOException(
 		    "Failed to scan dictionary string - dictionary was out of range. Database file appears to be corrupted.");
 	}
+	auto expected_width = BitpackingPrimitives::MinimumBitWidth(index_buffer_count - 1);
```
lnkuiper: This validates the stored width using another "untrusted" header field, so while this catches some issues, it isn't complete, everything should be validated.

lnkuiper (commented):
Thanks for the PR! I've left a comment below. You've been sending multiple PRs like this recently, and while each codec is different, I was thinking that maybe we want to have a shared "bounded metadata reader" to catch these issues in a unified way? We would still need some one-off checks for certain codecs, but that might be worthwhile

src/storage/compression/dictionary/decompression.cpp
wufengwind: Agreed — I’ve extended the validation so the two persisted fields no longer validate only each other. The index count and bitpacking width must agree, and the selection buffer size derived from the segment count and validated width must exactly match the stored index buffer offset. The existing checks then validate the index and dictionary regions.

I’ve also added a checksum-valid fixture where the index count and width agree but the resulting selection buffer layout is invalid. A shared bounded metadata reader still sounds worthwhile as a broader follow-up, but I’ve kept this PR scoped to completing the dictionary validation.

src/storage/compression/dictionary/decompression.cpp
```diff
@@ -80,7 +97,7 @@ void CompressedStringScanState::Initialize(ColumnSegment &segment, bool initiali
 
 	dict = DictionaryCompression::GetDictionary(segment, *handle);
 	auto index_buffer_end = index_buffer_offset + sizeof(uint32_t) * index_buffer_count;
-	if (dict.size > block_size || dict.end > block_size || dict.size > dict.end || index_buffer_count == 0 ||
+	if (dict.size > block_size || dict.end > block_size || dict.size > dict.end ||
```
lnkuiper: `dict.end` is relative to the segment, but this compares it against the full block size without accounting for `segment.GetBlockOffset()`. For a dictionary segment stored at a nonzero block offset, an end beyond the remaining block capacity can pass validation, after which `baseptr + dict.end` points outside the pinned block. Could we validate against `block_size - segment.GetBlockOffset()` using overflow-safe arithmetic and add a nonzero-offset regression case?

lnkuiper (commented):
Thanks for the changes! Just one more safety check I think we should add to this PR:

src/storage/compression/dictionary/decompression.cpp
wufengwind: Thanks, good catch. I now validate that the block offset and dictionary header fit before reading the header, then derive the remaining segment capacity using subtraction. The index region is checked with subtraction and division, and `dict.end` is validated against that segment-relative capacity.

I also added a checksum-valid regression fixture with a dictionary segment at block offset 992 and `dict.end` one byte beyond the remaining block capacity.

lnkuiper:
Looks good now, thanks!

PR #23828, opened 2026-07-15
Storage: validate FSST dictionary offsets

wufengwind:
FSST string scans trust dictionary metadata and delta-decoded dictionary offsets read from persisted string segments. A checksum-valid corrupted FSST segment can shrink `dict_end` or use an invalid dictionary offset and make scans decode strings from the wrong position instead of reporting storage corruption.

This PR validates the FSST dictionary range before constructing string pointers, and validates the shared string dictionary fetch helper before using persisted dictionary offsets or lengths. Invalid metadata is reported as an `IOException`.

A sqllogictest is added with a checksum-valid corrupted FSST database file covering an invalid dictionary end.

Fixes #21939.


wufengwind:
Closing this because #24016 now rejects the corrupted `dict_end` fixture during FSST segment initialization. The remaining per-offset validation would need a separate focused regression test, so keeping this PR open no longer reflects its current scope.

PR #23829, opened 2026-07-15
Fix Parquet bloom filter pruning for blobs

### LINKED ISSUE #23053: Incorrect results: BLOB `=` filter pushdown returns wrong rows for `DELTA_LENGTH_BYTE_ARRAY`-encoded long strings written by parquet-rs
### What happens?

When a `BLOB`/`VARCHAR` column is:

1. longer than 12 bytes (non-inlined `string_t`), and
2. encoded with `DELTA_LENGTH_BYTE_ARRAY`, and
3. written by **parquet-rs** (Apache `arrow-rs` / `parquet` crate),

then an **equality** filter that gets **pushed down** into the parquet scan returns the **wrong** result: a row whose value is exactly equal to the constant is **not** returned.

The bytes are read correctly (`to_hex()` returns the exact value), and the same equality works fine in a scalar projection (`SELECT col = const`). Only the **pushed-down equality filter** (`WHERE`, `JOIN`, `GROUP BY`, `count(DISTINCT)`) is affected.

Range comparisons (`>=`, `<=`, `BETWEEN`) on the same column/value work correctly. Only `=` (and `!=`, which shares the same path) is wrong.

Crucially:

- `SET disabled_optimizers='filter_pushdown'` → result becomes **correct**.
- `SET disabled_optimizers='statistics_propagation'` → still **wrong** (so this is **not** zone-map/statistics pruning).
- Materializing the column first (CTE with `OFFSET 0`, temp table, or `COPY` round-trip) → **correct**.
- Comparing via `to_hex(col)` → **correct**.

A 1-row table makes the bug unmistakable: both `WHERE col = X` and `WHERE col != X` return 0 rows, which is logically impossible unless the row is being dropped by the filter regardless of the actual comparison result.

### To Reproduce

The trigger requires a file written by **parquet-rs** using `DELTA_LENGTH_BYTE_ARRAY`. Files written by `parquet-cpp`/pyarrow or by DuckDB itself (PLAIN encoding) do **not** reproduce it, even with identical bytes — so the encoding + writer combination matters.

#### Step 1 — generate a minimal file with parquet-rs (Rust)

`Cargo.toml`:
```toml
[dependencies]
parquet = "58"
arrow = "58"
```

`src/main.rs`:
```rust
use std::sync::Arc;
use std::fs::File;
use arrow::array::BinaryArray;
use arrow::datatypes::{DataType, Field, Schema};
use arrow::record_batch::RecordBatch;
use parquet::arrow::ArrowWriter;
use parquet::basic::{Encoding, Compression};
use parquet::file::properties::{WriterProperties, WriterVersion};

fn main() {
    // A single 106-byte value with embedded NUL bytes and some high bytes,
    // mimicking a hash/blob key. Any long (>12 byte) binary works.
    let mut v = vec![0u8; 106];
    for i in 0..106 { v[i] = ((i * 37 + 13) & 0xFF) as u8; }
    v[6] = 0; v[7] = 0; v[20] = 0; // embedded NULs

    let arr = BinaryArray::from(vec![Some(v.as_slice())]);
    let schema = Arc::new(Schema::new(vec![Field::new("k", DataType::Binary, false)]));
    let batch = RecordBatch::try_new(schema.clone(), vec![Arc::new(arr)]).unwrap();

    let props = WriterProperties::builder()
        .set_writer_version(WriterVersion::PARQUET_2_0)
        .set_dictionary_enabled(false)                       // avoid dictionary
        .set_encoding(Encoding::DELTA_LENGTH_BYTE_ARRAY)     // force DLBA
        .set_compression(Compression::ZSTD(Default::default()))
        .build();

    let file = File::create("repro.parquet").unwrap();
    let mut w = ArrowWriter::try_new(file, schema, Some(props)).unwrap();
    w.write(&batch).unwrap();
    w.close().unwrap();

    // print the value so we can query for it
    println!("value hex = {}", v.iter().map(|b| format!("{:02x}", b)).collect::<String>());
}
```

Run it; note the printed `value hex`.

#### Step 2 — query with DuckDB

```sql
-- replace <HEX> with the value hex printed by the Rust program
WITH lit AS (SELECT from_hex('<HEX>')::BLOB AS v)

-- 1) scalar projection: CORRECT
SELECT (k = (SELECT v FROM lit)) AS eq      -- => true
FROM 'repro.parquet';

-- 2) pushed-down equality: WRONG (returns 0, should be 1)
SELECT count(*) FROM 'repro.parquet', lit WHERE k = lit.v;

-- 3) range equivalent of equality: CORRECT (returns 1)
SELECT count(*) FROM 'repro.parquet', lit WHERE k >= lit.v AND k <= lit.v;

-- 4) != on the same value: ALSO WRONG (returns 0; with (2) both are 0 on a 1-row table)
SELECT count(*) FROM 'repro.parquet', lit WHERE k != lit.v;

-- 5) disabling pushdown fixes it (returns 1)
SET disabled_optimizers='filter_pushdown';
SELECT count(*) FROM 'repro.parquet', lit WHERE k = lit.v;
SET disabled_optimizers='';

-- 6) statistics_propagation is NOT the cause (still 0)
SET disabled_optimizers='statistics_propagation';
SELECT count(*) FROM 'repro.parquet', lit WHERE k = lit.v;
SET disabled_optimizers='';

-- 7) to_hex workaround: CORRECT (returns 1)
SELECT count(*) FROM 'repro.parquet', lit WHERE to_hex(k) = to_hex(lit.v);
```

Observed on `v1.5.1`:

| query | result | expected |
|-------|--------|----------|
| (1) scalar `k = v` | `true` | `true` ✅ |
| (2) `WHERE k = v` | `0` | `1` ❌ |
| (3) `WHERE k >= v AND k <= v` | `1` | `1` ✅ |
| (4) `WHERE k != v` | `0` | `0` (but see note) ❌ |
| (5) pushdown disabled | `1` | `1` ✅ |
| (6) stats disabled | `0` | `1` ❌ |
| (7) `to_hex` | `1` | `1` ✅ |

(4): on a single-row table, `=` and `!=` returning **both** 0 demonstrates the row is dropped by the filter independently of the comparison outcome.

### Likely cause (analysis)

The `DELTA_LENGTH_BYTE_ARRAY` decoder constructs zero-copy `string_t` values pointing directly into the decompressed page block, introduced in commit `9e3bc26980` ("DeltaLengthByteArray: directly refer to strings from the block without copying"):

```cpp
// extension/parquet/decoder/delta_length_byte_array_decoder.cpp
result_data[result_idx] = string_t(char_ptr_cast(block.ptr), str_len);
block.unsafe_inc(str_len);
```

The previous code did `StringVector::EmptyString` + `memcpy` + `Finalize()`. The new path does **not** call `Finalize()`.

In the pushed-down filter path (`ColumnReader::ApplyFilter` → `ColumnSegment::FilterSelection` → `TemplatedFilterSelection<string_t, Equals>`), equality uses `StringComparisonOperators::Equals`, whose fast path compares the **first 8 bytes of the `string_t` struct (length + 4-byte prefix)** and returns `false` immediately if they differ — there is **no full-data `memcmp` fallback when the prefix differs**:

```cpp
uint64_t a_bulk_comp = Load<uint64_t>(&a);     // length(4) + prefix(4)
uint64_t b_bulk_comp = Load<uint64_t>(&b);
if (a_bulk_comp != b_bulk_comp) return false;  // no fallback if prefix bytes differ
```

By contrast, `GreaterThan`/`LessThan` fall through to `memcmp(left.GetData(), right.GetData(), min_length)` (following the pointer to the real data), which is why range comparisons stay correct on the very same zero-copy `string_t`. This matches the observed behavior (only `=`/`!=` are wrong; `>=`/`<=`/`BETWEEN` are fine), and explains why materialization (which copies + finalizes) and disabling pushdown both fix it.

So the zero-copy DLBA `string_t` ends up with a `prefix`/struct representation that the equality fast-path treats as unequal to an otherwise-identical, properly-finalized literal — even though the underlying bytes are identical.

### OS:

Linux x86_64

### DuckDB Version:

v1.5.1

### DuckDB Client:

Python / CLI

### Hardware:

_No response_

### Full Name:

-

### Affiliation:

-

- [x] Yes, I have

- [x] Yes, I have

Yes

kryonix:
Parquet bloom-filter pruning hashed `BLOB` constants using their escaped display representation (for example, `\xFF`) instead of their raw bytes. This could incorrectly exclude row groups containing matching values, affecting equality filters and join-filter pushdown. The Parquet encoding itself was not the cause. The fix hashes the raw `BLOB` payload and restricts bloom-filter pruning to direct column-to-constant equality predicates. Computed expressions such as `to_hex(blob) = ...` no longer probe the raw column's bloom filter.

fix: https://github.com/duckdb/duckdb/issues/23053
fix: https://github.com/duckdblabs/duckdb-internal/issues/9747

### DIFF
--- a/extension/parquet/parquet_statistics.cpp
+++ b/extension/parquet/parquet_statistics.cpp
@@ -22,6 +22,7 @@
 #include "duckdb/planner/expression/bound_comparison_expression.hpp"
 #include "duckdb/planner/expression/bound_conjunction_expression.hpp"
 #include "duckdb/planner/expression/bound_constant_expression.hpp"
+#include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "reader/uuid_column_reader.hpp"
 #include "duckdb/common/type_visitor.hpp"
 #include "column_reader.hpp"
@@ -670,18 +671,39 @@ unique_ptr<BaseStatistics> ParquetStatisticsUtils::TransformColumnStatistics(con
 	return row_group_stats;
 }
 
+static optional_ptr<const BoundConstantExpression> GetBloomFilterConstant(const Expression &expr) {
+	if (!BoundComparisonExpression::IsComparison(expr)) {
+		return nullptr;
+	}
+	auto &comp = expr.Cast<BoundFunctionExpression>();
+	if (comp.GetExpressionType() != ExpressionType::COMPARE_EQUAL) {
+		return nullptr;
+	}
+	auto &left = BoundComparisonExpression::Left(comp);
+	auto &right = BoundComparisonExpression::Right(comp);
+	optional_ptr<const Expression> column;
+	optional_ptr<const BoundConstantExpression> constant;
+	if (left.GetExpressionClass() == ExpressionClass::BOUND_REF &&
+	    right.GetExpressionClass() == ExpressionClass::BOUND_CONSTANT) {
+		column = left;
+		constant = right.Cast<BoundConstantExpression>();
+	} else if (right.GetExpressionClass() == ExpressionClass::BOUND_REF &&
+	           left.GetExpressionClass() == ExpressionClass::BOUND_CONSTANT) {
+		column = right;
+		constant = left.Cast<BoundConstantExpression>();
+	} else {
+		return nullptr;
+	}
+	if (column->Cast<BoundReferenceExpression>().Index() != 0 || constant->GetValue().IsNull() ||
+	    column->GetReturnType() != constant->GetValue().type()) {
+		return nullptr;
+	}
+	return constant;
+}
+
 static bool HasFilterConstants(const Expression &expr) {
 	if (BoundComparisonExpression::IsComparison(expr)) {
-		auto &comp = expr.Cast<BoundFunctionExpression>();
-		if (comp.GetExpressionType() != ExpressionType::COMPARE_EQUAL) {
-			return false;
-		}
-		auto &right = BoundComparisonExpression::Right(comp);
-		if (right.GetExpressionType() != ExpressionType::VALUE_CONSTANT) {
-			return false;
-		}
-		auto &constant = right.Cast<BoundConstantExpression>();
-		return !constant.GetValue().IsNull();
+		return GetBloomFilterConstant(expr) != nullptr;
 	}
 	if (expr.GetExpressionClass() != ExpressionClass::BOUND_CONJUNCTION) {
 		return false;
@@ -732,8 +754,8 @@ static uint64_t ValueXXH64(const Value &constant) {
 	case PhysicalType::DOUBLE:
 		return ValueXH64FixedWidth<double>(constant);
 	case PhysicalType::VARCHAR: {
-		auto val = constant.GetValue<string>();
-		return duckdb_zstd::XXH64(val.c_str(), val.length(), 0);
+		auto &val = StringValue::Get(constant);
+		return duckdb_zstd::XXH64(val.data(), val.size(), 0);
 	}
 	default:
 		return 0;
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks! Looks good. While I had Codex looking at this, it found an issue in float/double row group pruning:
```sql
# name: test/temp.test
# group: [parquet]

require parquet

statement ok
COPY (SELECT CASE WHEN range = 0 THEN -0.0::FLOAT ELSE 1.0::FLOAT END AS f FROM range(10))
TO '{TEST_DIR}/negative_zero_bloom.parquet'
(FORMAT PARQUET, WRITE_BLOOM_FILTER TRUE, DICTIONARY_SIZE_LIMIT 1000);

query I
SELECT count(*) FROM '{TEST_DIR}/negative_zero_bloom.parquet' WHERE f = 0.0::FLOAT;
----
1
```
Damn -0.0 vs +0.0 again. Pre-existing and unrelated to this PR, just wanted to document this here.

PR #23837, opened 2026-07-15
More missing includes breaking non-unity builds

evertlammerts:
Two more missing includes that prevent non-unity builds from compiling


lnkuiper:
Thanks! Any idea why CI isn't picking these up? Did we remove our non-unity build run?

PR #23839, opened 2026-07-15
Fix canonical bindings for nested common subplans

kryonix:
Nested subplan extraction creates new CTE/projection bindings after canonical table maps are built. Parent materialization later exposed these generated bindings and failed lookup.

fix: https://github.com/duckdblabs/duckdb-internal/issues/9953


lnkuiper:
Thanks!

PR #23840, opened 2026-07-15
Support OLD transition table for AFTER UPDATE triggers

krleonid:
## Summary

Enables `REFERENCING OLD TABLE AS` on `AFTER UPDATE` triggers (previously rejected with a `NotImplementedException`). The base `UPDATE` optionally captures the pre-update (OLD) row image alongside the NEW image, so trigger bodies can join/compare OLD and NEW transition tables.

### How it works
- `LogicalUpdate`/`PhysicalUpdate` carry `capture_old_rows` + `old_row_offset`. When set, the operator's source output is the NEW image followed by the OLD image, both in table order.
- The binder captures each physical column's scanned pre-update value (adding unscanned SET-to-constant columns to the scan) **before** rowid, so rowid stays the final input column, and records the OLD block offset.
- The OLD columns are exposed on the trigger base CTE under reserved internal names; trigger expansion builds the OLD transition alias by renaming them back and recomputing generated columns over the OLD image. Capture is enabled only when an AFTER trigger actually references OLD.

### Collision safety
The reserved OLD-capture column names are derived from the table (the prefix is extended until it is not a case-insensitive prefix of any real column), so they cannot collide with a user column. Without this, a user column named like the internal prefix would make the OLD alias silently read the NEW value — this is now impossible and covered by a test.

## Tests
- `test/sql/trigger/trigger_after_update_old_new.test` — OLD-only, OLD+NEW (both orders), zero-row updates, `SET x=x`, `UPDATE ... FROM`, volatile single-evaluation, forced del+insert path, LIST columns, generated columns, coexistence with RETURNING, `SELECT *` shape, rollback on failing body, and the name-collision case.
- `test/sql/trigger/trigger_after_update_old_new_wal.test` — WAL variant.
- Updated `trigger_referencing_validation.test` to reflect that OLD (and OLD+NEW) are now supported.

Full `test/sql/trigger/*` suite passes (22 cases, 1600 assertions) on `make reldebug`.

### DIFF
--- a/src/include/duckdb/parser/query_node/update_query_node.hpp
+++ b/src/include/duckdb/parser/query_node/update_query_node.hpp
@@ -36,9 +36,6 @@ public:
 	unique_ptr<UpdateSetInfo> set_info;
 	//! bind the same way as ALTER TABLE expressions (table catalog preferred)
 	bool prioritize_table_when_binding = false;
-	//! internal marker (never set by SQL parsing): trigger expansion sets this on its generated base UPDATE
-	//! CTE when an AFTER UPDATE trigger references OLD, so the update also captures the pre-update row image
-	bool capture_old_rows = false;
 
 public:
 	string ToString() const override;
--- a/src/include/duckdb/parser/query_node/update_query_node.hpp
+++ b/src/include/duckdb/parser/query_node/update_query_node.hpp
@@ -36,6 +36,9 @@ public:
 	unique_ptr<UpdateSetInfo> set_info;
 	//! bind the same way as ALTER TABLE expressions (table catalog preferred)
 	bool prioritize_table_when_binding = false;
+	//! internal marker (never set by SQL parsing): trigger expansion sets this on its generated base UPDATE
+	//! CTE when an AFTER UPDATE trigger references OLD, so the update also captures the pre-update row image
+	bool capture_old_rows = false;
 
 public:
 	string ToString() const override;
--- a/src/include/duckdb/planner/binder.hpp
+++ b/src/include/duckdb/planner/binder.hpp
@@ -193,10 +193,13 @@ struct GlobalBinderState {
 	optional_ptr<BoundParameterMap> parameters;
 	//! Tables whose triggers have already been expanded in this query (recursion detection)
 	reference_set_t<TableCatalogEntry> trigger_expanded_tables;
-	//! Generated base-UPDATE nodes that must also capture the pre-update (OLD) row image, mapped to the reserved
-	//! column names under which the captured OLD columns are exposed (physical table order). Internal
-	//! trigger-expansion state, keyed by node identity; never set by SQL parsing and never serialized.
-	reference_map_t<QueryNode, vector<Identifier>> trigger_old_capture_columns;
+	//! Generated base-UPDATE nodes that must capture the pre-update (OLD) row image (the capture request),
+	//! keyed by node identity. Internal trigger-expansion state; never set by SQL parsing and never serialized.
+	reference_set_t<QueryNode> trigger_old_capture;
+	//! Optional presentation detail for the statement-trigger CTE consumer: the reserved column names, in physical
+	//! table order, under which the captured OLD image is exposed via SQL aliases. Only populated for capture
+	//! requests presented through a CTE; a consumer that reads the OLD image directly leaves it absent.
+	reference_map_t<QueryNode, vector<Identifier>> trigger_old_capture_cte_names;
 	//! Set during CREATE TRIGGER body validation to detect self-recursive writes
 	optional_ptr<TableCatalogEntry> trigger_creation_table;
 	//! Name of the trigger being created (for error messages)
--- a/src/include/duckdb/planner/binder.hpp
+++ b/src/include/duckdb/planner/binder.hpp
@@ -193,9 +193,10 @@ struct GlobalBinderState {
 	optional_ptr<BoundParameterMap> parameters;
 	//! Tables whose triggers have already been expanded in this query (recursion detection)
 	reference_set_t<TableCatalogEntry> trigger_expanded_tables;
-	//! Generated base-UPDATE nodes that must also capture the pre-update (OLD) row image. Internal
+	//! Generated base-UPDATE nodes that must also capture the pre-update (OLD) row image, mapped to the reserved
+	//! column names under which the captured OLD columns are exposed (physical table order). Internal
 	//! trigger-expansion state, keyed by node identity; never set by SQL parsing and never serialized.
-	reference_set_t<QueryNode> trigger_old_capture_nodes;
+	reference_map_t<QueryNode, vector<Identifier>> trigger_old_capture_columns;
 	//! Set during CREATE TRIGGER body validation to detect self-recursive writes
 	optional_ptr<TableCatalogEntry> trigger_creation_table;
 	//! Name of the trigger being created (for error messages)
@@ -236,10 +237,6 @@ public:
 public:
 	DUCKDB_API BoundStatement Bind(SQLStatement &statement);
 	DUCKDB_API BoundStatement Bind(QueryNode &node);
-
-	//! Reserved internal column name for a captured OLD physical column in a trigger base UPDATE CTE. The name
-	//! is derived from the table so it cannot collide with any real column name (see the .cpp for details).
-	static string TriggerOldCaptureColumnName(const TableCatalogEntry &table, idx_t physical_index);
 
 	unique_ptr<BoundCreateTableInfo> BindCreateTableInfo(unique_ptr<CreateInfo> info);
 	unique_ptr<BoundCreateTableInfo> BindCreateTableInfo(unique_ptr<CreateInfo> info, SchemaCatalogEntry &schema,
--- a/src/include/duckdb/planner/binder.hpp
+++ b/src/include/duckdb/planner/binder.hpp
@@ -52,6 +52,7 @@ class TableMacroCatalogEntry;
 class UpdateSetInfo;
 class LogicalProjection;
 class LogicalGet;
+class LogicalUpdate;
 class LogicalVacuum;
 
 class ColumnList;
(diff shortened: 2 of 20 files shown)


src/planner/binder/query_node/bind_trigger_expansion.cpp
```diff
@@ -122,6 +124,66 @@ static unique_ptr<CommonTableExpressionInfo> MakeTransitionTableAliasCTE(const I
...
+			auto expr = column.GeneratedExpression().Copy();
+			expr->SetAlias(column.Name());
+			outer->select_list.push_back(std::move(expr));
+		} else {
+			outer->select_list.push_back(make_uniq<ColumnRefExpression>(column.Name()));
+		}
+	}
+	outer->from_table = std::move(subquery);
+	alias_cte->query_node = std::move(outer);
+	return alias_cte;
```
lnkuiper, line 223: This causes issues with column ordering for generated columns that reference each other. Normal table binding already handles this through generated-column expansion. Repro:
```sql
# name: test/temp.test
# description: OLD transition tables preserve generated-column dependency order
# group: [trigger]

statement ok
CREATE TABLE target (
    x INTEGER,
    g2 INTEGER GENERATED ALWAYS AS (g1 + 1),
    g1 INTEGER GENERATED ALWAYS AS (x + 1)
);

statement ok
CREATE TABLE audit (x INTEGER, g2 INTEGER, g1 INTEGER);

statement ok
INSERT INTO target (x) VALUES (1);

statement ok
CREATE TRIGGER trg AFTER UPDATE ON target
REFERENCING OLD TABLE AS old_rows
FOR EACH STATEMENT
    INSERT INTO audit SELECT * FROM old_rows;

statement ok
UPDATE target SET x = 10;

query III
SELECT * FROM audit;
----
1	3	2
```

src/include/duckdb/parser/query_node/update_query_node.hpp
```diff
@@ -36,6 +36,9 @@ class UpdateQueryNode : public QueryNode {
 	unique_ptr<UpdateSetInfo> set_info;
 	//! bind the same way as ALTER TABLE expressions (table catalog preferred)
 	bool prioritize_table_when_binding = false;
+	//! internal marker (never set by SQL parsing): trigger expansion sets this on its generated base UPDATE
+	//! CTE when an AFTER UPDATE trigger references OLD, so the update also captures the pre-update row image
+	bool capture_old_rows = false;
```
lnkuiper: `capture_old_rows` is transient binder control state: it is never produced by SQL parsing and is only set while expanding a generated trigger CTE. Keeping it on `UpdateQueryNode` means `Copy`, `Equals`, and query-node serialization all need to understand internal trigger-expansion state. Could this instead live in scoped binder state or generated-CTE metadata `CommonTableExpressionInfo::is_trigger_generated` is an existing precedent for internal metadata that is copied but not serialized. If the flag moves out of the parsed AST, the added query-node serialization field can go away as well.

src/include/duckdb/planner/operator/logical_update.hpp
```diff
@@ -28,8 +28,12 @@ class LogicalUpdate : public LogicalOperator {
 	TableCatalogEntry &table;
 	//! projection index
 	TableIndex table_index;
-	//! if returning option is used, return the update chunk
+	//! if returning option is used, or transition tables are captured, return the update chunk
 	bool return_chunk;
+	//! if set, the operator also emits the pre-update (OLD) row image after the NEW image
+	bool capture_old_rows = false;
+	//! input-chunk offset at which the captured OLD physical columns begin (only set when capture_old_rows)
+	idx_t old_row_offset = 0;
```
lnkuiper: `old_row_offset` is derived from the child projection layout but is stored and serialized as independent state on the parent operator. `PhysicalUpdate` then relies on the OLD columns being contiguous and in physical table order. Could we carry an explicit capture/return-column mapping, similar to `LogicalDelete::return_columns`, instead? That would remove the duplicate source of truth and make the contract safer for plan serialization and future changes to the UPDATE projection.

src/planner/binder/statement/bind_update.cpp
```diff
@@ -213,10 +266,29 @@ BoundStatement Binder::BindNode(UpdateQueryNode &node) {
...
+		                                      update_table_index, std::move(update_as_logicaloperator));
+		if (capture_old) {
+			// Expose the captured OLD physical columns as extra output columns of the base CTE, under reserved
+			// internal names, so the trigger's OLD transition alias can rename them. Public RETURNING (bound
+			// above) references only the NEW image, so its semantics are unchanged.
+			auto &proj = returning_result.plan->Cast<LogicalProjection>();
+			auto new_column_count = table.GetTypes().size();
+			idx_t physical_index = 0;
+			for (auto &column : table.GetColumns().Physical()) {
+				auto name = TriggerOldCaptureColumnName(table, physical_index);
```
lnkuiper: This converts positional OLD values into synthetic names in the generic RETURNING projection, after which trigger expansion independently recomputes those names and binds them again. Several loops now need to agree on physical-column order, and `TriggerOldCaptureColumnName` has to be exposed publicly on Binder solely to maintain that contract.

Could `BindReturning` be generalized to bind a logical table at a specified output offset? NEW at offset 0 and OLD at the OLD-image offset? That would reuse the existing star/name/generated-column expansion and avoid routing internal positional data through created column names. If names remain necessary (I hope they aren't), at least compute the vector once and pass it through the generated CTE rather than deriving it independently in each layer.

lnkuiper (commented):
Thanks! I've left some comments below.

Related: we have undefined behaviour in `test/sql/update/test_update_from.test:103` which we might want to look into changing before the v2.0 release to make triggers more predictable. Should be done in a separate PR.

src/planner/binder/query_node/bind_trigger_expansion.cpp
krleonid: Good catch — fixed in 16492dcf91. The OLD alias now recursively inlines generated-column references down to physical columns before recomputing, so the recomputation no longer depends on generated-column declaration order. Added your repro as a regression test (g2 declared before g1); it now returns 1|3|2 as expected.

src/include/duckdb/parser/query_node/update_query_node.hpp
krleonid: Done in a9d4e94b68. `capture_old_rows` is no longer on `UpdateQueryNode` — it's tracked in transient binder state (`GlobalBinderState::trigger_old_capture_nodes`), keyed by node identity and active only while binding the generated trigger chain, mirroring the existing `trigger_expanded_tables` recursion guard. The base UPDATE is bound by reference, so node identity targets it precisely (and avoids capturing on an unrelated same-table UPDATE in a runtime trigger cycle). Removed the field from Copy/Equals and dropped the query-node serialization field, since it's never produced by SQL parsing.

src/include/duckdb/planner/operator/logical_update.hpp
krleonid: Done in 27799adc8a. Replaced `old_row_offset` with an explicit `vector<idx_t> old_row_columns` (mirroring `LogicalDelete::return_columns`): each entry is the input-chunk index of one physical column's captured OLD value, so PhysicalUpdate no longer assumes the OLD columns are contiguous.

One check on intent: this keeps an explicit mapping stored on the operator (safer contract, no contiguity assumption), but it's still derived from the child projection at bind time. If by "remove the duplicate source of truth" you meant eliminating stored positional state entirely (e.g. referencing OLD columns via ColumnBindings the optimizer rewrites), that's a larger change closer to the binding rework in your other comment — happy to go that route if that's what you had in mind.

src/planner/binder/statement/bind_update.cpp
krleonid: Partially addressed in c586255229. I removed the public `Binder::TriggerOldCaptureColumnName` helper and the per-layer duplication: the reserved names are now computed once per expansion (file-local `OldCaptureColumnNames`) and threaded to all consumers via scoped binder state (node -> names), so no layer re-derives the contract.

On eliminating the synthetic names entirely / binding a logical table at an offset: I don't think it's achievable without reworking how transition tables are provided. They're SQL CTEs that reference the base CTE by column name, so the base CTE can't expose both NEW and OLD under the real column names (duplicate names) — one set has to be renamed. Binding NEW at offset 0 / OLD at offset N would require the OLD transition table to be a logical base-table binding over the operator output rather than a SQL CTE, which is a larger departure from the current CTE-based expansion (and inconsistent with how NEW and INSERT/DELETE transition tables work). Happy to explore that if you think it's worth it, but wanted to flag the trade-off first.

src/include/duckdb/planner/binder.hpp
```diff
@@ -192,6 +193,10 @@ struct GlobalBinderState {
 	optional_ptr<BoundParameterMap> parameters;
 	//! Tables whose triggers have already been expanded in this query (recursion detection)
 	reference_set_t<TableCatalogEntry> trigger_expanded_tables;
+	//! Generated base-UPDATE nodes that must also capture the pre-update (OLD) row image, mapped to the reserved
+	//! column names under which the captured OLD columns are exposed (physical table order). Internal
+	//! trigger-expansion state, keyed by node identity; never set by SQL parsing and never serialized.
+	reference_map_t<QueryNode, vector<Identifier>> trigger_old_capture_columns;
```
lnkuiper: This map appears to couple two separate concerns: its presence requests physical OLD-row capture, while its value carries synthesized column names used to present that image through the statement-trigger CTE. Future first-class OLD/NEW scopes will need row-image capture without depending on those SQL aliases. Could we introduce a typed capture context, or otherwise separate the capture request from the optional CTE presentation names, so this does not become the extension point for both responsibilities?

lnkuiper (changes requested):
Thanks! I just have one more concern (which is regarding supporting both OLD and NEW at the same time in a future PR):

lnkuiper (approved):
Thanks! LGTM now

PR #23841, opened 2026-07-15
[v1.5] Backport two more fixes

dentiny:
Hi team, I know we're preparing v1.5.5 release, I'm wondering if you think ok to include two more fix PRs? I think they reproduces on v1.5 branch.


test/sql/storage/compression/dict_fsst/dict_fsst_expression_filter.test
```diff
@@ -0,0 +1,128 @@
+# name: test/sql/storage/compression/dict_fsst/dict_fsst_expression_filter.test
+# description: Test expression filters over DICT_FSST strings
+# group: [dict_fsst]
+
+require vector_size 2048
+
+load {TEST_DIR}/dict_fsst_expression_filter.db readwrite v2.0.0
```
lnkuiper: I think this needs to be v1.4.0 or v1.5.0

lnkuiper (commented):
Thanks! I think we can fix this in v1.5 as well. Just one issue in a test:

test/sql/storage/compression/dict_fsst/dict_fsst_expression_filter.test
dentiny: good catch! updated to v1.5.0

PR #23843, opened 2026-07-15
Support `ANALYSE` where we support `ANALYZE`

Dtenwolde:

Let's keep the Brits happy.

`ANALYSE` was already supported by the old parser and is now supported by the new parser as well. Wherever `ANALYZE` works, `ANALYSE` now does too.


lnkuiper:
Love it! ❤️ 

More candidates:
 * `MATERIALIZED` (CTE/View)
 * `SUMMARIZE`

Not sure if PG supports these but we could.

djouallah:
it will be nice if we can have  optimize table/schema/database  x  , it does not need to do anything for now just the verb :)

PR #23848, opened 2026-07-15
Avoid corrupting buffered indexes during shutdown checkpoint

### LINKED ISSUE #23788: Secondary ART index serialized corrupt when a second in-process engine checkpoints a pending WAL (index scan returns 0 rows) — regression since 1.4.0
## What happens?

A **secondary ART index** (`CREATE INDEX`) is serialized **corrupt** when a DuckDB engine at
version **≥ 1.4.0** checkpoints a pending WAL — while a *second, independent* DuckDB engine holds
the same database file open in the **same process**. After that checkpoint, an equality lookup on
the indexed column returns **0 rows via the index scan**, even though the row exists:

- `SELECT count(*) FROM t WHERE a = 1` → **0** (wrong; should be 1) — served by the index scan
- `SELECT count(*) FROM t` → 2 (correct, unfiltered)
- `PRAGMA disable_optimizer; SELECT count(*) FROM t WHERE a = 1` → 1 (correct — bypasses the index)
- `EXPLAIN ANALYZE` shows `TABLE_SCAN … Type: Index Scan … Filters: a=1 … 0 rows`

This is a **silent wrong-results / data-correctness bug**, not a crash. The corrupt bytes are
**persisted** to the database file, so any later stock DuckDB (CLI, any client) that opens the file
read-only reads the wrong answer — no concurrency is needed to *observe* it once the file is written.

It is a **regression**: the *checkpointing* engine's version is what matters.

| checkpointing engine | result |
|---|---|
| 1.1.3 | correct |
| 1.3.2 | correct |
| 1.4.1 | **corrupt** |
| 1.5.1 | **corrupt** |
| 1.5.2 (CLI, manifestation) | reads corrupt file wrong |
| 1.5.4 | **corrupt** (latest tested) |

The **writer's** version is irrelevant — writer 1.5.4 + checkpointer 1.5.4 reproduces (same
version), writer 1.3.2 + checkpointer 1.3.2 does not. So it is **not** a cross-version WAL-format
issue; it is a ≥ 1.4.0 concurrency regression in the checkpoint-of-pending-WAL path for secondary
ART indexes.

**Necessary conditions** (each isolated by turning it off — remove any one and the file is clean):

1. A **secondary** `CREATE INDEX` (ART) on the filtered column. A **PRIMARY KEY**'s implicit ART
   alone does **not** trigger it. Not type-specific (INTEGER and TEXT both reproduce).
2. A **pending WAL** at the moment of the second open (inserts committed but not yet checkpointed;
   the writer connection is still open).
3. A **second, independent** DuckDB engine co-opening the file **in the same process** while the
   writer is still open. DuckDB's database-file lock is **per-PID**, so two engines loaded from two
   separate libduckdb images can co-open one file in one process; two separate *processes* cannot
   (cross-process lock), and a **sequential** handoff (writer process exits, then a fresh engine
   checkpoints) does **not** reproduce — the concurrency is essential.
4. The second engine opens **read-write and checkpoints** (open-time auto-checkpoint suffices, or an
   explicit `CHECKPOINT`). A **read-only** second engine cannot checkpoint → no corruption.
5. The checkpointing engine is DuckDB **≥ 1.4.0**.

Note: if the original writer later closes *cleanly* it re-checkpoints from its correct in-memory ART
and **heals** the on-disk file, so a post-hoc inspection can look clean. The file stays persistently
corrupt when that healing close never happens (e.g. the writer process dies while its connection is
open). The reproducer below deliberately skips the writer's clean close so the corrupt file is left
on disk.

### Part A — create the corruption (minimal, fully self-contained)

Two public DuckDB releases are required, both **the same version** to show it is not cross-version:

```
pip install duckdb==1.5.4
# download the matching C library and unzip it:
#   https://github.com/duckdb/duckdb/releases/download/v1.5.4/libduckdb-linux-amd64.zip
```

Script (writer = Python `duckdb`; checkpointer = the official `libduckdb` C library via `ctypes`):

```python
import ctypes as C
import os
import sys
import duckdb  # the WRITER

LIB = sys.argv[1]  # path to an official libduckdb.so >= 1.4.0
DB = os.path.abspath("duckdb_minimal_repro.duckdb")
for f in (DB, DB + ".wal"):
    try:
        os.remove(f)
    except FileNotFoundError:
        pass

# --- the READER/checkpointer: an independent libduckdb loaded via ctypes ---
lib = C.CDLL(LIB)
for name, argt, rest in [
    ("duckdb_create_config", [C.POINTER(C.c_void_p)], C.c_int),
    ("duckdb_set_config", [C.c_void_p, C.c_char_p, C.c_char_p], C.c_int),
    ("duckdb_open_ext", [C.c_char_p, C.POINTER(C.c_void_p), C.c_void_p, C.POINTER(C.c_char_p)], C.c_int),
    ("duckdb_connect", [C.c_void_p, C.POINTER(C.c_void_p)], C.c_int),
    ("duckdb_query", [C.c_void_p, C.c_char_p, C.c_void_p], C.c_int),
    ("duckdb_value_int64", [C.c_void_p, C.c_uint64, C.c_uint64], C.c_int64),
    ("duckdb_destroy_result", [C.c_void_p], None),
    ("duckdb_disconnect", [C.POINTER(C.c_void_p)], None),
    ("duckdb_close", [C.POINTER(C.c_void_p)], None),
    ("duckdb_destroy_config", [C.POINTER(C.c_void_p)], None),
]:
    fn = getattr(lib, name); fn.argtypes = argt; fn.restype = rest


def reader_scalar(sql, mode="READ_WRITE"):
    """Open DB with the ctypes libduckdb in `mode`, run `sql`, return int64 at (0,0), close.
    Opening READ_WRITE with a pending WAL performs the corrupting checkpoint."""
    cfg = C.c_void_p(); lib.duckdb_create_config(C.byref(cfg))
    lib.duckdb_set_config(cfg, b"access_mode", mode.encode())
    dbh = C.c_void_p(); err = C.c_char_p()
    if lib.duckdb_open_ext(DB.encode(), C.byref(dbh), cfg, C.byref(err)) != 0:
        lib.duckdb_destroy_config(C.byref(cfg))
        raise RuntimeError("open: " + (err.value or b"?").decode())
    lib.duckdb_destroy_config(C.byref(cfg))
    con = C.c_void_p(); lib.duckdb_connect(dbh, C.byref(con))
    res = (C.c_byte * 256)()
    if lib.duckdb_query(con, sql.encode(), C.byref(res)) != 0:
        lib.duckdb_disconnect(C.byref(con)); lib.duckdb_close(C.byref(dbh)); raise RuntimeError("query")
    v = lib.duckdb_value_int64(C.byref(res), 0, 0)
    lib.duckdb_destroy_result(C.byref(res))
    lib.duckdb_disconnect(C.byref(con)); lib.duckdb_close(C.byref(dbh))
    return v


# --- the WRITER: table + SECONDARY ART index + one autocommit insert; leave connection OPEN ---
w = duckdb.connect(DB)
w.execute("CREATE TABLE t(a INTEGER)")
w.execute("CREATE INDEX i ON t(a)")          # secondary ART index (a PRIMARY KEY would NOT trigger it)
w.execute("INSERT INTO t VALUES (1), (2)")   # autocommit -> committed to a PENDING WAL
assert os.path.exists(DB + ".wal"), "expected a pending WAL"
print(f"writer: duckdb {duckdb.__version__}; 2 rows; pending WAL; connection left open")

reader_ver = os.path.basename(os.path.dirname(LIB)) or "?"
reader_scalar("SELECT 1")                                     # first open (READ_WRITE) -> checkpoints the WAL
got = reader_scalar("SELECT count(*) FROM t WHERE a = 1")     # fresh open -> reads the checkpointed ART
bug = got != 1
print(f"reader: libduckdb {reader_ver}; after its checkpoint:  SELECT count(*) FROM t WHERE a=1  ->  {got}"
      f"   (correct = 1){'   <-- BUG' if bug else ''}")

if bug:
    print(f"persistently-corrupt file left at: {DB}")
    sys.exit(0)
sys.exit(1)
```

Run it:

```
python duckdb_minimal_repro.py /path/to/libduckdb-1.5.4/libduckdb.so
```

Output (writer `duckdb` 1.5.4, checkpointer `libduckdb` 1.5.4 — same version):

```
writer: duckdb 1.5.4; 2 rows; pending WAL; connection left open
reader: libduckdb libduckdb-1.5.4; after its checkpoint:  SELECT count(*) FROM t WHERE a=1  ->  0   (correct = 1)   <-- BUG
persistently-corrupt file left at: <cwd>/duckdb_minimal_repro.duckdb
```

### Part B — confirm the persisted corruption from the CLI (plain SQL, read-only)

The creation in Part A needs two independent engines, which a single CLI process cannot provide.
But the corruption is **persisted**, so it is fully reproducible with plain SQL from the stock CLI
against the file Part A leaves behind (here: DuckDB CLI **v1.5.2**):

```
duckdb -readonly duckdb_minimal_repro.duckdb "SELECT count(*) FROM t WHERE a = 1;"
duckdb -readonly duckdb_minimal_repro.duckdb "SELECT count(*) FROM t;"
duckdb -readonly duckdb_minimal_repro.duckdb "PRAGMA disable_optimizer; SELECT count(*) FROM t WHERE a = 1;"
duckdb -readonly duckdb_minimal_repro.duckdb "EXPLAIN ANALYZE SELECT count(*) FROM t WHERE a = 1;"
```

Output:

```
-- SELECT count(*) FROM t WHERE a = 1;   (should be 1)
┌──────────────┐
│ count_star() │
├──────────────┤
│       0      │   <- WRONG
└──────────────┘

-- SELECT count(*) FROM t;
┌──────────────┐
│ count_star() │
├──────────────┤
│       2      │
└──────────────┘

-- PRAGMA disable_optimizer; SELECT count(*) FROM t WHERE a = 1;
┌──────────────┐
│ count_star() │
├──────────────┤
│       1      │   <- correct once the index scan is bypassed
└──────────────┘

-- EXPLAIN ANALYZE ...
┌───────────────────────────┐
│         TABLE_SCAN        │
│      Type: Index Scan     │
│        Filters: a=1       │
│           0 rows          │
└───────────────────────────┘
```

I can attach a pre-corrupted `.duckdb` file on request so the manifestation can be observed with
zero setup, but Part A regenerates it deterministically from scratch.

## Environment

**OS:** Linux, x86_64 (verified on Fedora Linux host and in a Rocky Linux 8.10 container)

**DuckDB Version:** 1.5.4 (latest tested). Regression present since **1.4.0**: 1.4.1, 1.5.1, 1.5.2
(CLI), and 1.5.4 are all affected; 1.1.3 and 1.3.2 are unaffected.

**DuckDB Client:** Reproduced with the **C API** (`libduckdb` via `ctypes`) as the checkpointing
engine and the **Python** client as the writer; the persisted corruption is confirmed with the
**CLI**. The bug is in core storage/checkpoint and is client-independent.

**Hardware:** N/A (not performance-related).

## Related

Not a duplicate of #17004 (that is a *hang* from an unflushed WAL with two cursors on a single
connection, on 1.2.1). This report is *silent wrong results* from a corrupt secondary ART, requires
two independent engines checkpointing concurrently, and is a regression that first appears in 1.4.0.

## Identity Disclosure

**Full Name:** Rares Vernica

**Affiliation:** Paradigm4

## Before Submitting

- **Did you include all relevant configuration to reproduce the issue?** Yes
- **Did you include all code required to reproduce the issue?** Yes (the full creation script is
  inlined above)
- **Did you include all relevant data sets for reproducing the issue?** Not applicable — the
  reproduction does not require a data set (the script generates the table, index, and rows).

Minghan2005:
## Summary

- refuse to serialize an unbound index when it still contains buffered WAL replays and the checkpoint has no active client context
- preserve the WAL instead of persisting stale index storage during a shutdown checkpoint
- add a regression test covering WAL replay followed by a context-free shutdown checkpoint and restart

## Root cause

WAL recovery intentionally buffers index operations while an index remains lazily unbound. A normal checkpoint has a client context and binds the index before serialization, which applies those buffered operations. The shutdown checkpoint path has no client context, so it previously reused the unbound index's pre-WAL storage information and persisted an ART that did not contain the recovered rows. Index scans could then silently return wrong results.

The checkpoint now fails before serializing stale index metadata. On the shutdown path that leaves the WAL available for the next recovery. Once a query binds the index, a later checkpoint can persist it normally.

Fixes #23788.

## Validation

- reproduced the original two-engine failure with DuckDB 1.4.1: indexed count was `0`
- reran the reproducer against this branch: indexed count is `1`; a later read-only open also returns `1`
- `build/reldebug2/test/unittest 'test/sql/index/art/storage/test_art_wal_shutdown_checkpoint.test'`
- `build/reldebug2/test/unittest 'test/sql/index/art/storage/test_art_wal_checkpoint_minimal.test'`
- `scripts/format.py --fix -y HEAD`

## AI disclosure

Implemented and tested with assistance from OpenAI Codex. I reviewed the issue, reproduction, code changes, and test results.

### DIFF
--- a/src/storage/checkpoint_manager.cpp
+++ b/src/storage/checkpoint_manager.cpp
@@ -689,14 +689,22 @@ void SingleFileCheckpointWriter::WriteTable(TableCatalogEntry &table, Serializer
 	// Write the table metadata
 	serializer.WriteProperty(100, "table", &table);
 
+	auto &info = table.GetStorage().GetDataTableInfo();
 	// If there is a context available, bind indexes before serialization.
 	// This is necessary so that buffered index operations are replayed before we checkpoint, otherwise
 	// we would lose them if there was a restart after this.
 	if (context && context->transaction.HasActiveTransaction()) {
-		auto &info = table.GetStorage().GetDataTableInfo();
 		info->BindIndexes(*context);
+	} else {
+		for (auto &index : info->GetIndexes().Indexes()) {
+			if (!index.IsBound() && index.Cast<UnboundIndex>().HasBufferedReplays()) {
+				throw IOException(
+				    "Cannot checkpoint table \"%s\" without a client context because an index has buffered "
+				    "WAL operations",
+				    table.name);
+			}
+		}
 	}
-	// FIXME: If we do not have a context, however, the unbound indexes have to be serialized to disk.
 
 	// Write the table data
 	auto table_lock = table.GetStorage().GetCheckpointLock();
--- a/test/sql/index/art/storage/test_art_wal_shutdown_checkpoint.test
+++ b/test/sql/index/art/storage/test_art_wal_shutdown_checkpoint.test
@@ -1,0 +1,43 @@
+# name: test/sql/index/art/storage/test_art_wal_shutdown_checkpoint.test
+# description: avoid checkpointing buffered index replays without a client context
+# group: [storage]
+
+load {TEST_DIR}/test_art_wal_shutdown_checkpoint.db
+
+statement ok
+SET index_scan_max_count = 1;
+
+statement ok
+PRAGMA wal_autocheckpoint='1TB';
+
+statement ok
+PRAGMA disable_checkpoint_on_shutdown;
+
+statement ok
+CREATE TABLE tbl(i INTEGER);
+
+statement ok
+CREATE INDEX idx ON tbl(i);
+
+statement ok
+INSERT INTO tbl VALUES (42), (43);
+
+# Replay the WAL, leaving the index unbound with buffered inserts.
+restart
+
+# A shutdown checkpoint has no client context in which to bind the index. It must
+# preserve the WAL instead of serializing the stale index storage information.
+restart
+
+query I
+SELECT count(*) FROM tbl WHERE i = 42;
+----
+1
+
+# The query above binds the index. A subsequent shutdown checkpoint can persist it.
+restart
+
+query I
+SELECT count(*) FROM tbl WHERE i = 42;
+----
+1


Minghan2005:
Hi — before updating this PR, I did a fresh audit against current `main` and found an important gap in the existing validation.

The added regression test sets `PRAGMA disable_checkpoint_on_shutdown` before the `restart` steps. That means it does not exercise the no-client-context shutdown-checkpoint path, and therefore does not prove that this PR's guard is reached. I should not treat the current test or validation as sufficient.

The underlying invariant from #23788 still holds on current `main`: an `UnboundIndex` with buffered WAL replays must not be serialized using only its pre-replay `IndexStorageInfo`, because that can persist an index missing recovered rows and lead to silent wrong results. Current `main` still has the corresponding no-context serialization FIXME.

There is also now relevant recovery support in #24260 for a failed checkpoint marker. That makes the design choice here more explicit:

* **A. Preflight/skip:** before `WALStartCheckpoint`, detect an unbound index with buffered replays when there is no client context, and skip this shutdown checkpoint without changing the WAL. A later context-bearing checkpoint can bind the index and persist it normally.
* **C. Fail closed after checkpoint start:** keep the current exception-based approach. This prevents corruption, but the checkpoint marker has already been written; the contract must explicitly cover invalidation/error handling during shutdown and recovery of the main/checkpoint WAL state on the next open.

I would appreciate maintainer guidance on which contract is preferred. Once that direction is clear, I will rebuild the change on current `main` and add an exact regression: two independent DuckDB engines in the same process, a pending WAL with a secondary ART index, the no-context checkpoint path, and a forced index-scan correctness check after reopening. If sqllogictest cannot model the two-engine setup, I will use the smallest appropriate C++ persistence/API test instead.

hellower:
I found a smaller, sequential three-process reproducer for the same stale secondary-index invariant. The manifestation is a deterministic `BoundIndex::Delete` fatal rather than a wrong-result index scan.

Environment: stock DuckDB CLI `v1.5.5 (Variegata) d8cdaa33fd`, Darwin arm64.

### Process 1: create a pending WAL and exit without a shutdown checkpoint

```sh
DB=/tmp/art_shutdown_checkpoint.db
rm -f "$DB" "$DB.wal" "$DB.wal.checkpoint" "$DB.wal.recovery"

duckdb "$DB" -c "
SET wal_autocheckpoint='1TB';
CREATE TABLE t(
    serverid INTEGER PRIMARY KEY,
    host VARCHAR NOT NULL,
    port INTEGER NOT NULL
);
CREATE UNIQUE INDEX t_u01 ON t(host, port);
INSERT INTO t VALUES (1, 'host-a', 6432);
"

test -f "$DB.wal"
```

### Process 2: replay the WAL, do not touch the index, then exit

```sh
duckdb "$DB" -c "SELECT 1;"
test ! -e "$DB.wal"
```

The second process uses a fresh configuration, so `disable_checkpoint_on_shutdown` is no longer set. Opening replays the WAL into the lazy unbound secondary index; closing then runs the context-free shutdown checkpoint.

### Process 3: delete through the healthy primary key

```sh
duckdb "$DB" -c "DELETE FROM t WHERE serverid=1;"
```

Result:

```text
FATAL Error: Invalid Input Error: Failed to delete all rows from index. Only deleted 0 out of 1 rows.
Chunk: Chunk - [3 Columns]
- FLAT INTEGER: 1 = [ 1]
- FLAT VARCHAR: 1 = [ host-a]
- FLAT INTEGER: 1 = [ 6432]
```

The primary-key lookup still finds the table row, but removal from `t_u01` finds no corresponding entry. This is the same persisted table/index divergence viewed through the delete consistency check.

### Positive control

If Process 2 executes an explicit, context-bearing checkpoint instead:

```sh
duckdb "$DB" -c "CHECKPOINT;"
```

Process 3 succeeds and `SELECT count(*) FROM t` returns `0`. I reran the failing and control cases side by side against fresh files: failing shutdown path exited 1 with the fatal above; explicit-checkpoint control exited 0.

This supports option A from the latest discussion: preflight before `WALStartCheckpoint` and skip the context-free shutdown checkpoint while any unbound index has buffered replays. It also gives an exact oracle for the revised regression test. The test must reset the `disable_checkpoint_on_shutdown` setting between Process 1 and Process 2; an in-process `restart` that retains the setting will not reach this path.

hellower:
Follow-up validation for option A: I tested a downstream implementation against the pinned DuckDB v1.5.5 tree. The downstream repositories are private, so there is no externally inspectable patch link; the results below are provided as validation evidence rather than as a contribution to cherry-pick.

The tested implementation performs the check in `CreateCheckpoint`, **before** `WALStartCheckpoint`: when there is no active client transaction, it scans non-internal DuckDB tables and skips the shutdown checkpoint if any unbound index has buffered replays. This leaves the WAL and its index operations intact for a later context-bearing recovery/checkpoint.

The regression explicitly resets the shutdown setting between the first and second restart:

1. disable shutdown checkpointing, create a table with a primary key plus secondary UNIQUE ART, insert one row, and restart to leave/replay the WAL;
2. enable shutdown checkpointing, then restart again without touching the indexes, exercising the context-free shutdown checkpoint;
3. reopen and delete the recovered row through the primary key.

Validation completed on that downstream build:

- the stock v1.5.5 three-process scenario fails deterministically, while the preflight build passes;
- the exact regression passes;
- all 1,666 WAL tests pass;
- 5,984 debug tests pass, with 7 expected skips;
- Darwin arm64 and Linux amd64 static builds both pass their archive and extension verification.

Together with the public reproducer in the preceding comment, this provides behavioral evidence for rebuilding #23848 on current `main`. The existing PR author or maintainers will still need to implement and review the upstream change independently; no duplicate upstream PR is being opened.

aggress:
A third trigger for the same stale-secondary-index persistence, with the writer's exit being the realistic one: a writer process that dies **without closing its connection** (crash / SIGKILL / `process::exit` — no `disable_checkpoint_on_shutdown` needed), followed by any stock CLI open-and-exit.

**Environment:** DuckDB `v1.5.5 (Variegata) d8cdaa33fd` on both sides — the stock CLI and the `duckdb` Rust crate 1.10505.0 (bundled) — Linux 7.0.0 x86_64.

**Sequence** (single-threaded, one process at a time, never two engines on the file):

1. Writer (Rust crate): `INSERT` 30 rows into a table that has two secondary `CREATE INDEX`es (one DATE column, one BIGINT) inside `BEGIN … COMMIT`, then `std::process::exit(0)` without dropping the connection. The `.wal` is left pending. In-process before exit, index-served lookups on the new rows are correct.
2. `duckdb <file> </dev/null` — the CLI opens read-write and exits. No statement at all.
3. Reopen (either engine, read-only or read-write):
   - `SELECT count(*) FROM t WHERE date_col = DATE '…'` → **0** (30 rows exist)
   - `SELECT count(*) FROM t WHERE id_col = <max id>` → **0** (1 row exists)
   - `SELECT count(*) FILTER (WHERE date_col = DATE '…') FROM t` → 30 (scan-side, correct)
   - `EXPLAIN ANALYZE` shows the equality served by an index scan; `DROP INDEX` → 30.
   A second table written in the same step-1 transaction, with one secondary index on a VARCHAR column, lost its 30 entries in two of three runs.

**Controls, same pending WAL, each restores correctness:**
- step 2 done by the Rust crate instead (open, close cleanly) — fine;
- `duckdb -readonly <file>` — fine (cannot checkpoint; WAL stays);
- `duckdb <file> -c "CHECKPOINT"` — fine.

**One observation that matches #23848's root cause:** in a run where step 2 was `duckdb <file> -c "SELECT count(*) FROM t2 WHERE varchar_col = '…'"`, the index that query bound (t2's) was persisted correctly and the two unqueried indexes on t were not. With no statement, all three lost their entries. So it is per-index: whatever the CLI happened to bind before its context-free shutdown checkpoint survives; the rest is written from pre-WAL storage.

**Impact:** this reached a production file and stayed silent for five days — the writer service had been exiting without closing at every restart, and an unrelated `duckdb` CLI session on the file (opened read-write, exited normally) persisted the loss. Equality lookups on three tables returned wrong answers to application guards until an index-vs-scan consistency probe caught it. `DROP INDEX` + `CREATE INDEX` repaired it. The sequential form needs no concurrency and no special pragma, so it is reachable from any crashed writer plus a later CLI open.

dentiny:
Hi, I didn't notice the PR and issue but met the seemingly similar bug; so I open another PR here: https://github.com/duckdb/duckdb/pull/25771
I think one thing better for my solution is, checkpoint doesn't fail -- I don't think throwing `IOException` on normal database usage is a happy user experience; would appreciate if you're willing to take a look and provide some feedbacks, thanks!

PR #23849, opened 2026-07-15
Fix LEFT JOIN null-aware filter reordering

### LINKED ISSUE #23688: Join-related optimizer leads to incorrect result
### What happens?

### To Reproduce

```sql
memory D CREATE TABLE L(id INT, n INT);
memory D CREATE TABLE R(id INT, n INT);
memory D INSERT INTO L VALUES (1, NULL), (2, 2), (3, 3);
memory D INSERT INTO R VALUES (2, 2), (3, 99), (4, NULL);

-- Should be 2
memory D SELECT count(*) FROM L LEFT JOIN R ON L.id = R.id WHERE L.n IS NOT DISTINCT FROM R.n;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            3 │
└──────────────┘

-- Disable optimizer gets the correct result.
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D SELECT count(*) FROM L LEFT JOIN R ON L.id = R.id WHERE L.n IS NOT DISTINCT FROM R.n;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            2 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Minghan2005:
## Summary

- keep filters that originate above a `LEFT JOIN` out of the join graph when they reference the nullable side
- preserve those predicates as post-join filters during join-order reconstruction
- add a regression test for `IS NOT DISTINCT FROM` across both sides of a `LEFT JOIN`

## Root cause

The left-join pinning logic expanded a filter's relation set so it could not be pushed below the outer join. However, endpoint rebinding could subsequently classify the same comparison as a join predicate. For a null-aware comparison referencing both sides, that changed it from a post-join filter into an additional `LEFT JOIN` condition and incorrectly preserved a non-matching row.

## Validation

- `cmake --build build/reldebug2 --target unittest -j4`
- `build/reldebug2/test/unittest 'test/optimizer/joins/left_join_reordering/*'` (454 assertions, 3 test cases)
- verified the reported query returns `2` with the optimizer enabled and its plan keeps the comparison in a `Filter` above the `LEFT` hash join

Fixes #23688.

AI-assisted: I used OpenAI Codex while investigating and implementing this change, and reviewed and tested the resulting patch.

### DIFF
--- a/src/include/duckdb/optimizer/join_order/filter_info.hpp
+++ b/src/include/duckdb/optimizer/join_order/filter_info.hpp
@@ -39,6 +39,9 @@ public:
 	ColumnBinding left_binding;
 	ColumnBinding right_binding;
 	bool from_residual_predicate = false;
+	//! This filter originated above an outer join and references its nullable side.
+	//! It must remain a filter after that join instead of becoming another join condition.
+	bool must_remain_post_join = false;
 	//! Index of the equivalence group for INNER equality/IS NOT DISTINCT FROM join filters.
 	//! All filters transitively connected by equality (a=b, b=c -> a=c all share the same index).
 	//! Used by cardinality estimation to skip redundant transitive conditions.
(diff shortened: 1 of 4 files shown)


lnkuiper:
Thanks for the PR! Nice catch.

This has been superseded by https://github.com/duckdb/duckdb/pull/23895 which stumbled into the same issue and added more tests. Closing.

PR #23850, opened 2026-07-16
Defer setting cardinality set for appender

dentiny:
Disclaimer: the performance metrics baseline is after stats writer optimization at https://github.com/duckdb/duckdb/pull/23846

After apply the stats write optimization, I reran the [data ingestion benchmark](https://github.com/dentiny/duckdb/blob/3ddab96bb164ead583611abae3ad29a74646fbd6/benchmark/micro/appender_lineitem.cpp#L105-L119) and found `SetChildCardinality` takes 11% of CPU utilization.
image

I think the CPU is burning unnecessarily because all children vectors's cardinality is updated after one row append; in the ingestion benchmark table it has 16 columns and 1M rows, children vector cardinality is updated for at least 16M times.

In this PR, I defer the vector cardinality update
- Only update data chunk's logical cardinality on row append
- Update all children vector's cardinality when we flush

Performance improvement
| Metric | Before Cardinality Optimization | After Cardinality Optimization | Latency Reduction |
|---|---:|---:|---:|
| Median | 0.483638 s | 0.431463 s | **10.79%** |
| Mean | 0.485812 s | 0.433207 s | **10.83%** |
| P95 | 0.499798 s | 0.448550 s | **10.25%** |

### DIFF
--- a/src/include/duckdb/common/types/data_chunk.hpp
+++ b/src/include/duckdb/common/types/data_chunk.hpp
@@ -64,13 +64,17 @@ public:
 	void CheckCardinality(idx_t count_p);
 	//! Sets the cardinality of all child vectors of this chunk
 	void SetChildCardinality(idx_t count_p);
+	//! Sets only the logical cardinality; child vector sizes must be synchronized separately.
+	void SetCardinalityUnsafe(idx_t count_p) {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #23852, opened 2026-07-16
Inline Value::IsNan

dentiny:
Disclaimer: the perf baseline is built upon https://github.com/duckdb/duckdb/pull/23851

After applying the Vector::size inline optimization, I reran the [data ingestion benchmark](https://github.com/dentiny/duckdb/blob/3ddab96bb164ead583611abae3ad29a74646fbd6/benchmark/micro/appender_lineitem.cpp#L105-L119) and found `Value::isnan` seems to be the next bottleneck

Observed from flamegraph, `Value::isnan` seems to take 5.27% CPU usage
image
image

This PR brings the function definition into header file and inline it, performance change after the code change
| Metric | Before Inline | After Inline | Latency Reduction |
|---|---:|---:|---:|
| Median | 0.411289 s | 0.401222 s | **2.45%** |
| Mean | 0.412350 s | 0.402643 s | **2.35%** |
| P95 | 0.421596 s | 0.415949 s | **1.34%** |

### DIFF
--- a/src/common/types/value.cpp
+++ b/src/common/types/value.cpp
@@ -556,16 +556,6 @@ bool Value::DoubleIsFinite(double value) {
 }
 
 template <>
-bool Value::IsNan(float input) {
-	return std::isnan(input);
-}
-
-template <>
-bool Value::IsNan(double input) {
-	return std::isnan(input);
-}
-
-template <>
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #23861, opened 2026-07-16
Fix concurrent ALTER and INSERT crash

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/9914

The problem is that the storage got committed (flushed), then the locks are released and an add-column alter runs and bases a new state on top of this committed state, then the commit fails and we have to revert.

Then we have two problems:
- The revert expects the table to be `IsMainTable`, so that isn't true anymore because of the ALTER setting the parent to ALTERED.
- The ALTER transaction is poisoned because the state it based on shouldn't have existed, but as far as the ALTER knows the state was valid.
That parent DataTable could be removed resulting in a crash.

~~The solution is holding the lock so the flushed rows aren't seen until the state is fully committed or reverted, so unfinished state isn't leaked to other transactions.~~

The solution is reverting the table version back to MAIN_TABLE in the INSERT transaction's undo_buffer commit, and detecting that the base table was altered in the COMMIT of the ALTER transaction, failing it if that's the case.

Made in collaboration with @yan-alex
@Mytherin


test/persistence/test_concurrent_alter_insert.cpp
```diff
@@ -0,0 +1,136 @@
...
+		Connection alter_connection(db);
+		Connection insert_connection(db);
+		idx_t successful_inserts = 0;
+		for (idx_t round = 0; round < 25; round++) {
+			atomic<bool> unexpected_error(false);
+			bool insert_succeeded = false;
+			thread alter_thread([&]() {
+				RunConcurrentAlterInsertQuery(
+				    alter_connection, "ALTER TABLE t ADD COLUMN IF NOT EXISTS int_col__tmp INTEGER", unexpected_error);
+				RunConcurrentAlterInsertQuery(alter_connection, "UPDATE t SET int_col__tmp = int_col",
```
Mytherin: Can this be a sqllogic test with `concurrentloop`?

Mytherin (commented):
Thanks for the PR!

In my experience anything to do with "keeping locks around" is pretty gnarly because you run into lock order inversions or other potential hard to fix issues pretty quickly. This also moves us more from optimistic concurrency control to pessimistic concurrency control. 

How about this:

* In `RevertAppend` we set the table back to the main table (if it is not already the main table).
* In the commit of an `ALTER`, we check if the *child* data tables are main tables. If they are, we abort the alter with a concurrency conflict exception.

In the scenario you described the `ALTER` will then be reverted with a conflict exception (during the commit).

test/persistence/test_concurrent_alter_insert.cpp
Tishj: I had that, but it didn't seem to cover the same path, as this would still fail but the sqllogic test passed
see the attached issue for that test

Tishj:
Thanks for the review, that logic makes sense to me
Making the changes now, I can include the sqllogictest as well, but it might need some changes to get the same coverage as the cpp test

Tishj:
Since we rely on the order to be:
1. insert commits V1 to storage
2. alter-add-column changes V1 to ALTERED, creates V2
3. insert commits to WAL, reverts V1 to MAIN_TABLE
4. alter commits V2, realizes V1 is not ALTERED and fails the commit

Isn't there a potential problem if 4 happens before 3 ?

src/transaction/commit_state.cpp
```diff
@@ -272,6 +272,15 @@ void CommitState::CommitEntry(UndoFlags type, data_ptr_t data, CommitInfo &info)
 		D_ASSERT(catalog.IsDuckCatalog());
 
 		auto &new_entry = old_entry.Parent();
+		if (old_entry.type == CatalogType::TABLE_ENTRY && new_entry.type == CatalogType::TABLE_ENTRY) {
+			auto &old_storage = old_entry.Cast<DuckTableEntry>().GetStorage();
+			auto &new_storage = new_entry.Cast<DuckTableEntry>().GetStorage();
+			if (!RefersToSameObject(old_storage, new_storage) && old_storage.IsMainTable()) {
+				throw TransactionException("Failed to alter table \"%s\" because the underlying table state was "
```
ywelsch, line 279: I'm trying to understand the user-visible change here: As far as I know (please correct me if I'm wrong) DuckDB has resolved write-write conflicts between DDL and DML operations always in favor of DDL. The PR here seems to change that, which worries me as it might make it impossible to do DDL operations anymore in a system with a continuous stream of DML operations?

ywelsch: An alternative path here might be to go a route similar to  https://github.com/duckdb/duckdb/pull/23631 which made some of these pre-existing races more explicit due to timing changes (and where I started to fix them part of that PR).

The solution was to use a publish gate (i.e. a new table-level lock) that helps coordinate publication between DML / DDL while still preserving the existing semantics.

I've extracted the relevant change in https://github.com/duckdb/duckdb/compare/v1.5-variegata...ywelsch:duckdb:dat-2253-publish-gate

ywelsch: Or a lock-free solution I explored here: https://github.com/duckdb/duckdb/compare/v1.5-variegata...ywelsch:duckdb:dat-2253-alt-committed-count-clone

In that case the ALTER TABLE would clone the row groups capped at the committed row count, and prevent any concurrent DML operations from committing/publishing (which would then fail).

Mytherin: I think the current solution here is a low impact solution that is a good fit for a bug-fix release. We can consider a different solution for `main` and going forward. Right now if this is triggered this would result in internal errors so this is a direct improvement over the status quo at least.

Mytherin:
Thanks!

PR #23863, opened 2026-07-16
Throw descriptive error for chained operators

Dtenwolde:
Part of: https://github.com/duckdblabs/duckdb-internal/issues/9093#comp-01

The SQL standard does not allow the unparenthesized chaining of comparison operators, such as:
```sql
SELECT 1 = 1 = 1; 
SELECT 1 < 2 < 3; 
SELECT a IS DISTINCT FROM b IS DISTINCT FROM c;
```
The old parser gave a generic syntax error: 
```sql
memory D select 1 < 2 < 3;
Parser Error:
syntax error at or near "<"

LINE 1: select 1 < 2 < 3;
                     ^

memory D SELECT a IS DISTINCT FROM b IS DISTINCT FROM c;
Parser Error:
syntax error at or near "IS"

LINE 1: SELECT a IS DISTINCT FROM b IS DISTINCT FROM c;
                                    ^
```

The new parser allowed it up to now which was incorrect. So I think we should just stick to the SQL standard in this case, but provide a more descriptive error message:
```sql
memory D SELECT 1 < 2 < 3;
Parser Error:
Chained comparisons are not supported, use AND to combine comparisons
```

The valid way to write this expression is: 
```sql
memory D SELECT 1 < 2 AND 2 < 3;
```


Dtenwolde:
Little update, I changed a test that had: `ON t1 = b == t2.d` to use parenthesis: `ON t1 = (b == t2.d)`. 

The old parser did not allow `SELECT 1 = 1 = 1` but did allow an equivalent: `SELECT 1 == 1 = 1`, because at parse time `==` is seen as a multicharacter `Op`, which can be chained. And then during transform this was set to `COMPARE_EQUAL`. `=` was it's own special case for `a_expr` and did correctly not allow chaining. 

For the new parser `=` and `==` are equivalent already at parse time.

Mytherin:
Thanks!

PR #23864, opened 2026-07-16
Optimize GROUP BY on top of GROUP BY by pruning and removing inner aggregates

JelteF:
Previously, a query like

```sql
SELECT a FROM (SELECT a, b FROM t GROUP BY ALL) GROUP BY a
```

planned two full aggregations, with the inner one grouping on both `a` and `b` (forcing the scan to read `b`). But the query is logically equivalent to:

```sql
SELECT a FROM t GROUP BY a
```

Such a query only has a single aggregation and reads fewer columns. This implements an optimization to remove unnecessary group by columns from a group by node, as well as removing unnecessary group by nodes completely. This is done by recognizing when the presence of duplicates is irrelevant to a group by operator and transferring that knowledge downward to eliminate group by operators that don't have any effect.

With this change the first query above now generates this plan:

```sql
memory D create or replace table t as SELECT i%1000 as a , i%10*100 as b from generate_series(1, 100000) i;
Run Time (s): real 0.003 user 0.002809 sys 0.001001
memory D explain SELECT a FROM (SELECT a, b FROM t GROUP BY ALL) GROUP BY a;
╭─ Projection ──────────────────────╮
│ Projections: CAST(#0 AS BIGINT)   │
│ ~1,253 rows                       │
╭─ Perfect Hash Group By ───────────╮
│ Groups: #0                        │
│ ~1,253 rows                       │
╭─ Projection ────┴─────────────────╮
│ Projections: a                    │
│ ~100,000 rows                     │
╭─ Projection ────┴─────────────────╮
│ Projections: CAST(#0 AS SMALLINT) │
│ ~100,000 rows                     │
╭─ Seq Scan ──────┴─────────────────╮
│ Projections: a                    │
│ ~100,000 rows                     │
```


src/optimizer/remove_unused_columns.cpp
```diff
@@ -130,15 +131,305 @@ void RemoveUnusedColumns::ClearUnusedExpressions(vector<T> &list, TableIndex tab
...
+		case LogicalTypeId::MAP:
+		case LogicalTypeId::UNION:
+		case LogicalTypeId::VARIANT:
+			return false;
+		// normalizing comparator (EqualsFloat): -0.0 and 0.0 merge, as do all NaN bit patterns;
+		// a VARCHAR cast or signbit() tells both pairs apart ('0.0' vs '-0.0', 'nan' vs '-nan'),
+		// and 1/x additionally tells the zeros apart (inf vs -inf)
+		case LogicalTypeId::FLOAT:
+		case LogicalTypeId::DOUBLE:
+			return true;
```
JelteF: This is the primary reason to have this whole function. I'm wondering whether we should maybe instead change this behavior and not group `0.0` and `-0.0` together in the first place.

JelteF: To be extra clear, it solves this difference in behaviour:

```
memory D select x from (values (-0.0::float), (0.0)) x(x) group by all;
┌───────┐
│   x   │
│ float │
├───────┤
│  -0.0 │
└───────┘
Run Time (s): real 0.001 user 0.002575 sys 0.000341
memory D select x::text from (values (-0.0::float), (0.0)) x(x) group by all;
┌────────────────────┐
│ CAST(x AS VARCHAR) │
│      varchar       │
├────────────────────┤
│ -0.0               │
│ 0.0                │
└────────────────────┘
```

i.e. grouping by a float/double column has the "side-effect" of merging `-0.0` and `0.0` values in that column. Is that intentional behavior?

lnkuiper:
1. This shouldn't be implemented in `RemoveUnusedColumns`, it should be a separate optimiser
2. The `TypeVisitor` check to check where we can do this optimisation is not safe, it misses `VARIANT`, causing incorrect results:
```sql
SELECT ((-0.0::DOUBLE)::VARIANT) = ((0.0::DOUBLE)::VARIANT);
-- true

SELECT v::VARCHAR
FROM (VALUES ((-0.0::DOUBLE)::VARIANT),
             ((0.0::DOUBLE)::VARIANT)) t(v);
-- -0.0
--  0.0
```
3. The discrepancy between the behaviour you described [here](https://github.com/duckdb/duckdb/pull/23864#discussion_r3718736025) is caused by the `VARCHAR` cast being evaluated before grouping, which should be solved in the planner/binder, not in the optimiser.

src/optimizer/remove_unused_columns.cpp
JelteF: > The discrepancy between the behaviour you described https://github.com/duckdb/duckdb/pull/23864#discussion_r3718736025 is caused by the VARCHAR cast being evaluated before grouping, which should be solved in the planner/binder, not in the optimiser.

I guess I was not clear in what I meant to ask here. So I'll try again: Do you think the current behaviour of combining 0.0 and -0.0 for float/double makes sense (same for interval of 30 days and 1 month)? If you think that behaviour makes sense, then we need something like the TypeVisitor to preserve it. If not, then we could change it in a precursor PR to group them as separate groups and then this PR could get rid of this TypeVisitor logic.

lnkuiper: Th SQL standard defines `0.0` as equal to `-0.0`, but we do need to show the `-` when casting to `VARCHAR`, in which case they are no longer equal. See PostgreSQL:
```sql
postgres=# select g from (values (-0.0::float, 0.0::float)) tbl(g);
 g  
----
 -0
(1 row)

postgres=# select g from (values (0.0::float, -0.0::float)) tbl(g);
 g 
---
 0
(1 row)
```

src/optimizer/remove_unused_columns.cpp
```diff
@@ -272,7 +292,6 @@ void RemoveUnusedColumns::VisitOperator(unique_ptr<LogicalOperator> &op_ref) {
 			return;
 		}
 		if (!everything_referenced) {
-			// FIXME: groups that are not referenced need to stay -> but they don't need to be scanned and output!
```
JelteF, line 372: I moved this FIXME (in an updated form) to the new optimizer

JelteF:
@lnkuiper I think I addressed all your feedback now.

lnkuiper:
Thanks for the changes! I have a few more remarks.

The move into a separate optimizer addresses the ownership point, but the type handling still needs restructuring.

`GroupingMergesDistinguishableValues` should not define core equality semantics inside this optimizer. `BaseTypeAllowsCollapse` in `RemoveDuplicateGroups` models the same property independently, and the two already differ. Could we expose one shared predicate next to the comparison/hash/type logic and use it from both optimizers?

I don't think the large whitelist is needed. Equality is representation-preserving by default, so the shared helper only needs to conservatively identify the exceptional cases (`FLOAT`, `DOUBLE`, `INTERVAL`, collated `VARCHAR`, and `VARIANT`), while `TypeVisitor` handles nesting. That removes the exhaustive switch and keeps future comparator changes in one place. The current `VARIANT` test also does not exercise the `VARIANT` branch because `__collated_group` blocks the optimization first.

Could we also significantly reduce the comments in the new optimizer and test? Many of them restate the control flow, explain individual switch cases, or narrate the following assertions. Please keep short comments only for non-obvious semantic invariants and let the code and test names explain the rest.

PR #23866, opened 2026-07-16
In TableDescription, resolve catalog, schema, and table names through the catalog

evertlammerts:
`ClientContext::TableInfo` resolved a table through the catalog search path but then filled the returned TableDescription's `readonly` and `qualified_name` fields from the caller's input rather than from the entry it actually found. So whenever the input was unqualified and the search path landed somewhere other than the default database, the two fields described the wrong catalog. As far as I can tell this only affects the appender.


kyrylchenko:
could this fix also be added to the older versions? (e.g. 1.4v)?

PR #23868, opened 2026-07-16
[ASYNC IO] CSV Reader Refactor

pdet:
This PR refactors our CSV reader to split the scheduling and processing concerns for Async IO

Async IO can only be enabled for uncompressed, unpiped, unencoded CSV files, as these can be read in a non-sequential fashion.

Hence, this PR refactors the CSV buffer management and scanner so we can schedule the bytes that need to be scanned separatly than actually reading them, without any behavior change.

Buffer positions, sizes and counts are now computed without IO. We introduce a new buffer manager for random-access CSVs that can materialize any buffer independently of scan position. These buffers also don't need the existing pin/unpin mechanism, as they can always be thrown away under memory pressure and reread positionally.

We also separate the scanner construction so that the scheduling and materialization can be done in different steps. Finally, we split MoveToNextBuffer and FinalizeChunkProcess so we can easily implement the blocking there.


src/include/duckdb/execution/operator/csv_scanner/csv_random_buffer_manager.hpp
```diff
@@ -0,0 +1,33 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/execution/operator/csv_scanner/csv_buffer_manager.hpp"
+
+namespace duckdb {
+
+//! Buffer manager for files with known buffer byte ranges: any buffer can be materialized independently
```
Tishj: unknown* ?
Otherwise I don't really understand why it's called `CSVRandomBufferManager`
Oh, does it mean "random access" ?

src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
```diff
@@ -1629,68 +1629,78 @@ void StringValueScanner::ProcessOverBufferValue() {
...
+StringValueScanner::MoveBufferResult StringValueScanner::TryMoveToNextBuffer() {
+	if (iterator.pos.buffer_pos < cur_buffer_handle->actual_size) {
+		return MoveBufferResult::NOT_MOVED;
+	}
+	// a value spans at most two buffers, so the buffer before the previous one is never needed again:
+	// release it before fetching, keeping the peak at two pinned buffers per scanner
+	previous_buffer_handle.reset();
+	const idx_t next_buffer_idx = iterator.pos.buffer_idx + 1;
+	auto next_buffer = buffer_manager->GetBuffer(next_buffer_idx);
+	if (!next_buffer) {
```
Tishj: It feels like this could benefit from early-return ?

src/execution/operator/csv_scanner/buffer_manager/csv_sequential_buffer_manager.cpp
```diff
@@ -0,0 +1,121 @@
...
+		idx_t cur_buffer = buffer_idx + 1;
+		while (reset_when_possible.find(cur_buffer) != reset_when_possible.end()) {
+			cached_buffers[cur_buffer].reset();
+			reset_when_possible.erase(cur_buffer);
+			cur_buffer++;
+		}
+		return;
+	}
+	// We only reset if previous one was also already reset
+	if (buffer_idx > 0 && !cached_buffers[buffer_idx - 1]) {
```
Tishj: nit: `if (buffer_idx == 0 || cached_buffers[buffer_idx - 1]) { reset_when_possible.insert(buffer_idx); return; }` ?

PR #23870, opened 2026-07-16
bound bitpacking group offset against block size

jmestwa-coder:
unchecked per-group data offset in the bitpacking scan:
- LoadNextGroup reads the 24-bit group offset from the segment and builds the group pointer in GetPtr with no range check
- a corrupt bitpacked column with an out-of-range offset makes the frame-of-reference/width/constant reads land outside the pinned block
- only the first metadata pointer was validated, the per-group offset had a release-only D_ASSERT

bounded the offset against the block size, matching the ALP scan check.


jmestwa-coder:
gentle ping

PR #23871, opened 2026-07-16
Add missing row group prune via pruned string min

dentiny:
Followup PR for https://github.com/duckdb/duckdb/pull/23728

### DIFF
--- a/src/storage/statistics/string_stats.cpp
+++ b/src/storage/statistics/string_stats.cpp
@@ -709,7 +709,7 @@ FilterPropagateResult StringStats::CheckZonemap(const BaseStatistics &stats, Exp
 		if (HasMinMax(stats)) {
 			// Special handle cases where constant is equal to the truncated min bound, but we're able to prune a few
 			// row groups. For example, a truncated "bbbbbbbbbbbb" with minimum length 13 has the lower bound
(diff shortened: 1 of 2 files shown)


lnkuiper:
This PR and #23873 don't prune this:
```sql
# name: test/temp.test
# description: Nullable byte-length statistics should still prune false-or-NULL row groups
# group: [temp]

statement ok
CREATE TABLE nullable_string_lengths AS
SELECT CASE
       WHEN i % 2 = 0 THEN NULL
       WHEN i < 122880 THEN 'short'
       ELSE repeat('x', 50)
       END AS v
FROM range(245760) tbl(i);

query II
EXPLAIN ANALYZE
SELECT sum(strlen(v)) FROM nullable_string_lengths WHERE strlen(v) > 30;
----
analyzed_plan	<REGEX>:.*Row Groups Scanned: 1 / 2.*
```
Could we add a fix for that to either of these PRs?

Other than that, both look great!

dentiny:
Hi @lnkuiper , I took a look and found it could be a more general issue on NULL handling, not necessarily related to string stats prune.
I file an issue here https://github.com/duckdb/duckdb/issues/23899 and fix here https://github.com/duckdb/duckdb/pull/23902

PR #23882, opened 2026-07-16
[Secret] Add `TRANSACTION` persist-type for SECRET, added to the `MetaTransaction`

Tishj:
This PR is made as part of duckdb-iceberg work

"vended" credentials are short-lived credentials issued by the catalog server to provide access to the data of a table, it seemed to me that it was the cleanest to model this in core as credentials attached to the transaction.

### DIFF
--- a/src/main/secret/secret_manager.cpp
+++ b/src/main/secret/secret_manager.cpp
@@ -362,8 +362,9 @@ SecretMatch SecretManager::LookupSecret(CatalogTransaction transaction, const st
 }
 
 unique_ptr<SecretEntry> SecretManager::GetSecretByName(CatalogTransaction transaction, const string &name,
-                                                       const string &storage) {
+                                                       const string &storage_p) {
 	InitializeSecrets(transaction);
+	auto storage = Identifier(storage_p);
 	if (storage.empty() || storage == TRANSACTION_STORAGE_NAME) {
 		if (auto transaction_storage = GetTransactionSecretStorage(transaction)) {
 			auto result = transaction_storage->GetSecretByName(name, &transaction);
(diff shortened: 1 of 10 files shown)


carlopi:
I had a similar idea at https://github.com/duckdb/duckdb/pull/23527 (but on connection). I guess one question: it's this supported at SQL level? Like, there are no added sqllogic tests, maybe one would be handy (as self-documentation), with ROLLBACK / etc.

Tishj:
It isn't supposed to be supported at the sql level
but I'll add some sql tests to verify whether it can be reached and if so, extend with behavior tests or add explicit guardrails to disable it for now

I haven't been able to verify whether this meets the needs of iceberg yet, as I'm waiting for iceberg's v1.5's refresh work to upstream into main

Tishj:
> I had a similar idea at #23527 (but on connection). I guess one question: it's this supported at SQL level? Like, there are no added sqllogic tests, maybe one would be handy (as self-documentation), with ROLLBACK / etc.

That surprises me, since the non-persistent credentials are called TEMPORARY, I was expecting them to be bound to the connection already. Just as TEMPORARY tables are.
If they are scoped to the session, I think we should have been using GLOBAL TEMPORARY, following in Spark's footsteps for the terminology of session-scoped temporaries

src/main/secret/secret_manager.cpp
```diff
@@ -342,6 +364,16 @@ SecretMatch SecretManager::LookupSecret(CatalogTransaction transaction, const st
 unique_ptr<SecretEntry> SecretManager::GetSecretByName(CatalogTransaction transaction, const string &name,
                                                        const string &storage) {
 	InitializeSecrets(transaction);
+	if (storage.empty() || storage == TRANSACTION_STORAGE_NAME) {
```
lnkuiper, line 368: The `storage` comparison here is case-sensitive, while all other storage lookups go through `GetSecretStorage(Identifier(storage))`, which are case-insensitive

lnkuiper (commented):
Thanks! Looks good, just one comment:

src/main/secret/secret_manager.cpp
Tishj: They go trough `GetSecretStorage` because their storage is owned by the `SecretManager`, whereas `TRANSACTION` storage is owned by the `MetaTransaction`, so I don't think we can change that.

But I can make it a case-insensitive compare, through Identifier

PR #23888, opened 2026-07-17
StringStats: ignore reversed min/max bounds

wufengwind:
String min/max statistics are considered available whenever both bounds
have a statistics type. Unlike numeric statistics, their ordering was not
validated.

A Parquet file with exact VARCHAR or BLOB statistics where min > max could
therefore cause statistics propagation to replace a filtered scan with an
Empty Result, even when matching rows were present.

Require stored string bounds to be ordered before HasMinMax returns true.
Reversed bounds are treated as unavailable, causing readers to scan and
filter the data instead of incorrectly pruning it.

The regression test includes a small Parquet file with reversed exact
VARCHAR and BLOB statistics and covers equality and range filters.

Tests:
- build/reldebug/test/unittest test/sql/copy/parquet/parquet_reversed_string_stats.test
- build/reldebug/test/unittest "[parquet]"
- build/reldebug/test/unittest


dentiny:
> A Parquet file with exact VARCHAR or BLOB statistics where min > max 

curious in which case will we have min > max?

wufengwind:
The regression test file included in this PR is the simplest case: it has exact VARCHAR and BLOB statistics where both `min` and `max` are present, but `min > max`.

A conforming Parquet writer should never emit this, but Parquet statistics are external file metadata that DuckDB does not control. If we trust the ordering invariant without validation, a predicate that should match rows can be optimized away to an empty result. The fix defensively checks the bounds before treating the stats as usable for pruning.

dentiny:
> A conforming Parquet writer should never emit this, but Parquet statistics are external file metadata that DuckDB does not control

Curious why do you think we should ignore min/max silently, instead of an INVALID INPUT exception since it's a broken parquet file?

wufengwind:
I chose to ignore the reversed bounds because Parquet statistics are optional pruning metadata, not the column payload itself. If the data pages are still readable, rejecting the whole file because an auxiliary min/max range is malformed seems stricter than necessary.

There is already similar defensive behavior for string statistics: exact string stats that fail validation are treated as bogus and ignored rather than making the file unreadable. This PR applies the same idea to another invalid string-statistics case: if both bounds are present but ordered incorrectly, we should not trust them for pruning.

That said, if DuckDB’s preferred policy is to reject malformed Parquet statistics with an InvalidInputException, I can switch the PR in that direction.

PR #23890, opened 2026-07-17
CURL CLI uses 5 minutes to detect an unresponsible server, let us adapt

carlopi:
Goal would be allowing more slow / unresponsive server time to come back and actually send back responses. For S3 reads this is unheard of, but with layers of caching / compression / server side logic, this might be more friendly to users, and assumes that a query pointing to talk to a slow server should be honored.

It's always possible to configure, in a stricter or more relaxed direction, via:
```sql
SET http_timeout = X;
```

### DIFF
--- a/src/include/duckdb/common/http_util.hpp
+++ b/src/include/duckdb/common/http_util.hpp
@@ -29,7 +29,7 @@ struct HTTPParams {
 	}
 	virtual ~HTTPParams();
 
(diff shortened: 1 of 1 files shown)


carlopi:
Actually, I discussed this with @Mytherin, let's leave it to 30 for `v1.5-variegata`, to be reworked on `main`.

PR #23891, opened 2026-07-17
Storage: validate Roaring metadata bounds

wufengwind:
## Summary

Building on #23796, which validates the Roaring container count against the available metadata space, this adds bounds validation for `metadata_offset` against the segment's remaining block capacity and checks the actual metadata bytes consumed during deserialization.

Roaring stores `metadata_offset` relative to the data pointer after the segment header. For persistent column segments, scans may start at a non-zero `block_offset`, while `SegmentSize()` is the usable block size. A corrupted offset could therefore pass validation while making `metadata_ptr` point outside the block data.

The fix checks the offset against the remaining block space, passes the remaining metadata byte count into metadata deserialization, and validates the type/run/array metadata sections before reading them. It also verifies that decoded container payload offsets do not overlap or pass the metadata start.

A regression test creates a valid Roaring validity segment, corrupts the metadata offset while recomputing the block checksum, and verifies that scanning reports a corrupted Roaring segment instead of returning incorrect counts.

## Tests

- Build: `make reldebug`
- C++ regression: `build/reldebug/test/unittest "Roaring metadata offset cannot point outside block bounds"`
- Roaring storage sqllogictest: `build/reldebug/test/unittest test/sql/attach/attach_new_compression.test`
- Storage test group: `build/reldebug/test/unittest "[storage]"`

### DIFF
--- a/src/storage/compression/roaring/metadata.cpp
+++ b/src/storage/compression/roaring/metadata.cpp
@@ -1,5 +1,6 @@
 #include "duckdb/storage/compression/roaring/roaring.hpp"
 
+#include "duckdb/common/exception.hpp"
 #include "duckdb/common/numeric_utils.hpp"
 #include "duckdb/common/bitpacking.hpp"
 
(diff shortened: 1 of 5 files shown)


src/storage/compression/roaring/metadata.cpp
```diff
@@ -153,14 +154,22 @@ idx_t ContainerMetadataCollection::Serialize(data_ptr_t dest) const {
...
+}
+
+void ContainerMetadataCollection::Deserialize(data_ptr_t src, idx_t container_count, idx_t available_size) {
 	container_type.resize(AlignValue<idx_t, BitpackingPrimitives::BITPACKING_ALGORITHM_GROUP_SIZE>(container_count));
 	count_in_segment = container_count;
 
 	// Load the types of the containers
 	idx_t types_size = BitpackingPrimitives::GetRequiredSize(container_type.size(), 2);
+	idx_t read_offset = 0;
+	VerifyMetadataRead(read_offset, types_size, available_size, "container type");
```
yan-alex, line 170: Isn't it better to make this an assertion?
`D_ASSERT(VerifyMetadataRead(...))` and then return a boolean from VerifyMetadataRead()
that way these checks only run in debug mode, and don't slow down release builds

wufengwind: I considered that, but I think this should stay as release-time validation rather than a `D_ASSERT`.

These values are read from the persisted segment data, so a corrupted database file can hit this path in a release build. If the checks are compiled out, the subsequent bitpacking reads/memcpy can read past the valid metadata area before any debug assertion would help. The overhead is limited to metadata deserialization, and the benefit is that corrupted storage fails with an `IOException` instead of relying on debug-only checks.

dentiny: I added a ALWAYS_ASSERT, if you're seeking for runtime validation regardless of build mode

wufengwind: Thanks, I considered `ALWAYS_ASSERT` as well. I used an explicit `IOException` here because the values come from persisted storage data and can be triggered by a corrupted database file, so this is not just an internal invariant violation. Reporting it as a corrupted Roaring segment seemed more useful than surfacing an assertion/InternalException.

wufengwind:
Closing this in favor of #23796, which predates this PR and already has maintainer involvement. This PR explored additional metadata-section and payload bounds, but those are better revisited as a focused follow-up after #23796 rather than competing with the existing fix.

PR #23894, opened 2026-07-17
Fix potential memory leak after realloc failure in /third_party/hyperloglog/sds.cpp

guschin058:
In the sdssplitargs() function, there is a potential memory leak on line 1045 after calling realloc()
https://github.com/duckdb/duckdb/blob/main/third_party/hyperloglog/sds.cpp#L1045

If an error occurs, realloc() returns NULL, and the original memory block is not freed. Since the result of the function is directly assigned to the vector variable, the former address of the allocated block is lost, which makes it impossible to release it later. 

```
sds *sdssplitargs(const char *line, int *argc) {
    const char *p = line;
    char *current = NULL;
    char **vector = NULL;
    ...
    ...
    }
```
```
....
....
 /* add the token to the vector */
            vector = (char**) realloc(vector,((*argc)+1)*sizeof(char*));
            vector[*argc] = current;
            (*argc)++;
            current = NULL;
...
...
```

To prevent this situation, it is recommended to first save the result of realloc() to a temporary pointer (tmp), check it for NULL, and assign it to the vector variable only upon successful reallocation. In case of an error, proceed to the handling of an exceptional situation (goto err), while maintaining access to the original memory block.

Found using the cppcheck static analyzer
Reporter: Guschin Egor (guschin058@yandex.ru)

### DIFF
--- a/third_party/hyperloglog/sds.cpp
+++ b/third_party/hyperloglog/sds.cpp
@@ -1042,11 +1042,11 @@ sds *sdssplitargs(const char *line, int *argc) {
                 if (*p) p++;
             }
             /* add the token to the vector */
-            char **new_vector =(char **)realloc(vector, ((*argc) + 1) * sizeof(char *));
-            if (new_vector == NULL) {
+            char **tmp = (char **)realloc(vector, ((*argc) + 1) * sizeof(char *));
+            if (tmp == NULL) {
                 goto err;
             }
-            vector = new_vector;
+            vector = tmp;
             vector[*argc] = current;
             (*argc)++;
             current = NULL;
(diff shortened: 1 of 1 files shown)


third_party/hyperloglog/sds.cpp
ProjectMutilation: missed whitespace after `=`

ProjectMutilation: `new_vector` is better rename to `tmp`

ProjectMutilation (changes requested):

third_party/hyperloglog/sds.cpp
guschin058: Thanks, I have fixed it

guschin058:
@Mytherin

ProjectMutilation:
failing of tidy job isn't related with pull request, rebase will fix it probably

Mytherin:
Thanks but this is fixing third party code that is not used anywhere in the codebase. There's no point to fixing it as DuckDB doesn't use it and this introduces a divergence with upstream code. Maybe you want to upstream the fix instead.

PR #23908, opened 2026-07-18
Optimize CASE statement when all branches yield identical results

### LINKED ISSUE #23838: CASE with identical result branches is retained
### What happens?

Hi team, currently we don't simplify a searched CASE whose `THEN` and `ELSE` expressions are identical, even when its condition is a safe non-volatile column expression.

### To Reproduce

```sql
memory D CREATE TABLE t(i INTEGER, b BOOLEAN);
memory D INSERT INTO t VALUES (1, true), (2, false), (NULL, NULL);

-- b is unnecessary scanned sequentially
memory D EXPLAIN SELECT CASE WHEN b THEN i ELSE i END FROM t;
╭─ Projection ───────────────────────╮
│ Projections:                       │
│ CASE  WHEN (b) THEN (i) ELSE i END │
│ ~3 rows                            │
╰──────────────────┬─────────────────╯
╭─ Seq Scan ───────┴─────────────────╮
│ Table: memory.main.t               │
│ Type: Sequential Scan              │
│ Projections: b, i                  │
│ ~3 rows                            │
```

### OS:

macos

### DuckDB Version:

642f86f6993d7cc072219eb065e5beeafbd255b1

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

tempewda:
Fixes #23838

When a `CASE` expression has all `THEN` branches identical to the `ELSE` branch, the entire `CASE` is replaced with the `ELSE` expression. This avoids unnecessary column reads and condition evaluation.

Example:

```sql
-- Before: evaluates b, reads both b and i
SELECT CASE WHEN b THEN i ELSE i END FROM t;

-- After optimization: only reads i
SELECT i FROM t;
```
Changes:

- `src/optimizer/rule/case_simplification.cpp`: Added identical-branch detection after existing constant-folding logic
- `test/optimizer/case_simplification.test`: Added tests for single-branch, multi-branch, and non-matching cases

### DIFF
--- a/src/optimizer/rule/case_simplification.cpp
+++ b/src/optimizer/rule/case_simplification.cpp
@@ -42,6 +42,17 @@ unique_ptr<Expression> CaseSimplificationRule::Apply(LogicalOperator &op, vector
 		// no case checks left: return the ELSE expression
 		return std::move(root.ElseMutable());
 	}
+	// if all THEN branches are identical to the ELSE branch, we can just return the ELSE expression
+	bool all_identical = true;
+	for (auto &case_check : root.CaseChecks()) {
+		if (!Expression::Equals(case_check.then_expr, root.ElseMutable())) {
+			all_identical = false;
+			break;
+		}
+	}
+	if (all_identical) {
+		return std::move(root.ElseMutable());
+	}
 	return nullptr;
 }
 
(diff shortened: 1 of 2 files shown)


tempewda:
The `Main / windows / MinGW (64 Bit)` failure is a Chocolatey infrastructure issue - the CI couldn't install the compiler. Unrelated to this change.

tempewda:
updated the PR to explicitly check `IsVolatile()` so we don't accidentally optimize away `WHEN` expressions that have side effects

tempewda:
@dentiny Hey! let me know what you think of this PR when you have a moment to review it!

PR #23910, opened 2026-07-18
EXTERNAL RESOURCES: external resources via table functions

carlopi:
Part 1/4 of https://github.com/duckdb/duckdb/pull/23736

Provision/register/destroy external resources via table functions, improving on existing `src/function/table/system/create_external_resource.cpp`, no new SQL syntax.

### Two registries

**Types** are recipes: a `kind` plus the *names* of lifecycle callbacks (`create`, `status`, `destroy`), resolved lazily when invoked, so an extension can register one without a compiled-in dependency on them. Registration captures the caller's search path, as views and macros do, since the callbacks run later on a different connection.
 - `register_external_resource_type(name, kind := , create_function := , ...)`
 - `duckdb_external_resource_types()`

**Instances** are what this DuckDB instance currently tracks: an in-memory, instance-scoped view, keyed by a local name.
 - `register_external_resource(type, name, handle, uri := , deleter_function := , ...)`
 - `deregister_external_resource(name)` — forgets the record, tears nothing down
 - `duckdb_external_resources()`

### Two operations
 - `create_external_resource(type)` invokes `create`, blocks polling `status` until `ready` (cancellable, `timeout_seconds`-bounded), enforces the catalog baseline (`uri` + `attached_db_type`), and returns a deleter binding. `handle :=` adopts an existing resource instead of provisioning. On failure a guard tears down what we provisioned — but not what we adopted.
 - `destroy_external_resource(deleter_function, payload)` invokes a deleter binding.

Both act on the *world*, not on the registries: creating a resource does not register it, and destroying one does not deregister it. Recording it is a separate, explicit step — which is what lets you register something provisioned elsewhere, or that outlived the process.

Implementation side: callbacks run on a separate internal connection, since the caller's context lock is held — hence non-TEMP callbacks, and fully-qualified callback names.

### DIFF
--- a/src/function/table/system/create_external_resource.cpp
+++ b/src/function/table/system/create_external_resource.cpp
@@ -1,8 +1,4 @@
 #include "duckdb/function/table/system_functions.hpp"
-#include "duckdb/catalog/catalog.hpp"
-#include "duckdb/catalog/catalog_entry/schema_catalog_entry.hpp"
-#include "duckdb/catalog/entry_lookup_info.hpp"
-#include "duckdb/common/enums/on_entry_not_found.hpp"
 #include "duckdb/common/exception.hpp"
 #include "duckdb/common/string_util.hpp"
 #include "duckdb/common/types/value.hpp"
@@ -10,6 +6,7 @@
 #include "duckdb/main/connection.hpp"
 #include "duckdb/main/database.hpp"
 #include "duckdb/main/external_resource_type_registry.hpp"
+#include "duckdb/main/external_resources_manager.hpp"
 #include "duckdb/logging/log_type.hpp"
 #include "duckdb/logging/logger.hpp"
 #include "duckdb/parser/qualified_name.hpp"
@@ -73,22 +70,6 @@ static Value RequireResourceMap(const Value &value, const string &function_name,
 		    column, value.type().ToString());
 	}
 	return value.DefaultCastAs(LogicalType::MAP(LogicalType::VARCHAR, LogicalType::VARCHAR));
-}
-
-//! Resolve a table-producing callback (a table function or table macro) to its fully-qualified
-//! `catalog.schema.name` in `context`'s search path, so the returned deleter stays resolvable from any other
-//! connection later (e.g. when DETACH fires it). Returns empty if the callback can't be resolved.
-static string QualifyTableCallback(ClientContext &context, const string &name) {
-	EntryLookupInfo table_fn(CatalogType::TABLE_FUNCTION_ENTRY, QualifiedName::Parse(name));
-	auto entry = Catalog::GetEntry(context, table_fn, OnEntryNotFound::RETURN_NULL);
-	if (!entry) {
-		EntryLookupInfo table_macro(CatalogType::TABLE_MACRO_ENTRY, QualifiedName::Parse(name));
-		entry = Catalog::GetEntry(context, table_macro, OnEntryNotFound::RETURN_NULL);
-	}
-	if (!entry) {
-		return string();
-	}
-	return QualifiedName(entry->ParentCatalog().GetName(), entry->ParentSchema().name, entry->name).ToString();
 }
 
 //===--------------------------------------------------------------------===//
--- a/src/function/table/system/create_external_resource.cpp
+++ b/src/function/table/system/create_external_resource.cpp
@@ -10,12 +10,32 @@
 #include "duckdb/main/connection.hpp"
 #include "duckdb/main/database.hpp"
 #include "duckdb/main/external_resource_type_registry.hpp"
+#include "duckdb/logging/log_type.hpp"
+#include "duckdb/logging/logger.hpp"
 #include "duckdb/parser/qualified_name.hpp"
 
 #include <chrono>
 #include <thread>
 
 namespace duckdb {
+
+//! Empty `extra_info` map for an ExternalResource log entry.
+static Value NoExtraInfo() {
+	return Value::MAP(LogicalType::VARCHAR, LogicalType::VARCHAR, vector<Value>(), vector<Value>());
+}
+
+//! Single-entry `extra_info` map for an ExternalResource log entry.
+static Value ExtraInfo(const string &key, const string &value) {
+	return Value::MAP(LogicalType::VARCHAR, LogicalType::VARCHAR, {Value(key)}, {Value(value)});
+}
+
+//! Log one recipe callback invocation (on response). `resource_type` may be empty (rendered NULL) when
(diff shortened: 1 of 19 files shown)


src/function/table/system/external_resources.cpp
```diff
@@ -0,0 +1,215 @@
...
+struct RegisterExternalResourceState : public GlobalTableFunctionState {
+	bool done = false;
+};
+
+static unique_ptr<FunctionData> RegisterExternalResourceBind(ClientContext &context, TableFunctionBindInput &input,
+                                                             vector<LogicalType> &return_types, vector<string> &names) {
+	auto result = make_uniq<RegisterExternalResourceBindData>();
+	auto &resource = result->resource;
+	resource.type = StringValue::Get(input.inputs[0]);
+	resource.name = StringValue::Get(input.inputs[1]);
```
pdet, line 115: Should we null guard these?

pdet: Shoukld we also validate the type from the type registry?

pdet: Should we also validate the deleter function?

src/function/table/system/create_external_resource.cpp
```diff
@@ -177,26 +199,45 @@ static void CreateExternalResourceFunction(ClientContext &context, TableFunction
...
+	} else {
+		if (type->create_function.empty()) {
+			throw InvalidInputException("create_external_resource: resource type \"%s\" has no create function",
+			                            bind_data.type_name);
+		}
+		auto sql = "SELECT * FROM " + QualifiedName::Parse(type->create_function).ToString() + "(" +
+		           bind_data.params.ToSQLString() + ")";
+		auto result = con.Query(sql);
+		LogExternalResourceOperation(context, bind_data.type_name, bind_data.resource_name, "create",
+		                             result->HasError() ? result->GetError() : string(), NoExtraInfo());
```
pdet: Is the log written before failure check?

src/include/duckdb/main/external_resources_manager.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "duckdb/common/mutex.hpp"
+#include "duckdb/common/types/value.hpp"
+#include "duckdb/common/vector.hpp"
+
+namespace duckdb {
+class DatabaseInstance;
+class ClientContext;
+
+//! A provisioned external resource this instance owns, registered under a local name (from CREATE/REGISTER
+//! EXTERNAL RESOURCE). The durable identity is `(type, handle)`: everything else is the binding derived
```
pdet: The code in `create_external_resource` seems to dedup on name, should it be name or type,handle?

src/include/duckdb/main/external_resources_manager.hpp
```diff
@@ -0,0 +1,60 @@
...
+//! from it by the type's status/destroy callbacks and cached here for display + teardown. Parallels
+//! ExternalResourceType (the recipe) — this is one provisioned instance of it.
+struct ExternalResource {
+	//! The local registration name (e.g. `yoooo`).
```
pdet: nit: maybe a more realistic example than yoooo?

src/include/duckdb/main/external_resources_manager.hpp
pdet: Is this comment correct at all?

test/sql/table_function/register_external_resource.test
```diff
@@ -0,0 +1,72 @@
+# name: test/sql/table_function/register_external_resource.test
```
pdet, line 1: Can we add more testing?
e.g.,
  - Reject NULL type / NULL name
  - Reject NULL and empty MAP{} handle
  - Reject unregistered type and unresolvable deleter_function
  - Reject duplicate (type, handle) under different names
  - Register inside BEGIN…ROLLBACK survives

pdet (commented):
Thanks for the PR and for breaking it down, I had a few comments.

In addition, can we also add tests for the added logger? And add a slightly more detailed description of what this PR does, in case it is later read by others with less context about it.

src/include/duckdb/main/external_resources_manager.hpp
carlopi: Thanks, nice catch

carlopi: `name` is the only one that actually makes sense, rewrote the comment.

src/function/table/system/create_external_resource.cpp
carlopi: Thanks, now restructured, logged always once, before throwing (or success)

carlopi:
Thanks @pdet for pointing to various improvements, I think I addressed the feedback, and expanded on the PR comment. Ready for next round?

CI is at the moment picky, I think that's not on me.

src/function/table/system/create_external_resource.cpp
```diff
@@ -177,26 +183,53 @@ static void CreateExternalResourceFunction(ClientContext &context, TableFunction
...
+		if (type->create_function.empty()) {
+			throw InvalidInputException("create_external_resource: resource type \"%s\" has no create function",
+			                            bind_data.type_name);
+		}
+		auto sql = "SELECT * FROM " + QualifiedName::Parse(type->create_function).ToString() + "(" +
+		           bind_data.params.ToSQLString() + ")";
+		// Log the operation's real outcome: the callback query returning without error is not yet success, the
+		// result still has to satisfy the contract below. Logging from the catch also covers throws added later.
+		try {
+			auto result = con.Query(sql);
```
pdet, line 203: What happens if we have a recursive create function that calls a `create_external_resource` of its own type?

pdet (commented):
Thanks, just one more comment/curiosity

src/function/table/system/create_external_resource.cpp
carlopi: I would assume that might just work if there is some sort of terminator (like, recursion is bounded, like resource A creates resource B and returns).
If the function is just calling itself, that I would expect to be bad, but then the resource would be invalid.

In both cases, that's not covered by tests, I can expand on that, I am not completely sure which direction, since detecting this is not super solid (since it might loop back going different paths).

Maybe it's possible to just bound the number of recursive Connections that are created, threading a counter through, but that requires more infrastructure.

I would argue: this case can't really be detected nicely with the current infrastructure, but it's a sort of degenerate/broken case.

Can that be tackled on a side? Or it's a blocker?

pdet: It's fine it's as is, I think that's a bit nasty to get there, but we should consider covering it in the next PR.

PR #23911, opened 2026-07-18
Add logging type to cover all load or install events

carlopi:
Adds a new logging type: `ExtensionLoadInstall`

Example:
```sql
CALL enable_logging('ExtensionLoadInstall', storage='stdout');
SET autoload_known_extensions = true;
SET autoinstall_known_extensions = true;
SELECT broadcast('192.168.1.5/24'::inet);
```
```
0	DATABASE					2026-07-18 09:59:44.827796+00	ExtensionLoadInstall	INFO	"{'event': install, 'extension': inet, 'version': NULL, 'install_mode': NULL, 'source': NULL, 'reason': 'autoload for Scalar Function \'broadcast\'', 'error': 'Failed to download extension ""inet"" at URL ""http://extensions.duckdb.org/84d91709cd/osx_arm64/inet.duckdb_extension.gz"" (HTTP 404)
Extension ""inet"" is an existing extension.

For more info, visit https://duckdb.org/docs/current/extensions/troubleshooting?version=84d91709cd&platform=osx_arm64&extension=inet'}"
```

And similarly, on successful INSTALL or LOAD they will also be logged with a `reason`, that can be either of `explicit SQL LOAD statement`, indication of an attempted autoloading, or if done via C++ interface it can be supplied by relevant code bases (for example, extensions). `error` field will be `NULL` on success, or otherwise be the same as the error that would be raised.

This can be in particular handy to explain autoloading / auto-install behaviour I think. Similarly, it can be handy in using the logger to test autoloading and auto-install behaviour (not done as part of this PR).

### DIFF
--- a/src/execution/operator/helper/physical_load.cpp
+++ b/src/execution/operator/helper/physical_load.cpp
@@ -24,6 +24,7 @@ static void InstallFromRepository(ClientContext &context, const LoadInfo &info) 
 	options.throw_on_origin_mismatch = true;
 	options.version = info.version;
 	options.repository = repository;
+	options.reason = "explicit SQL INSTALL statement";
 
 	ExtensionHelper::InstallExtension(context, info.filename, options);
 }
@@ -36,6 +37,7 @@ SourceResultType PhysicalLoad::GetDataInternal(ExecutionContext &context, DataCh
 			options.force_install = info->load_type == LoadType::FORCE_INSTALL;
 			options.throw_on_origin_mismatch = true;
 			options.version = info->version;
+			options.reason = "explicit SQL INSTALL statement";
 			ExtensionHelper::InstallExtension(context.client, info->filename, options);
 		} else {
 			InstallFromRepository(context.client, *info);
@@ -45,6 +47,7 @@ SourceResultType PhysicalLoad::GetDataInternal(ExecutionContext &context, DataCh
 		ExtensionLoadOptions options;
 		options.extension_name = info->filename;
 		options.alias = info->alias;
+		options.reason = "explicit SQL LOAD statement";
 		ExtensionHelper::LoadExternalExtension(context.client, options);
 		// adds an explicitly set extension schema to the search path
 		ExtensionLoader::RefreshSearchPath(context.client);
--- a/src/logging/log_types.cpp
+++ b/src/logging/log_types.cpp
@@ -387,9 +387,8 @@ ExtensionLoadInstallLogType::ExtensionLoadInstallLogType() : LogType(NAME, LEVEL
 
 LogicalType ExtensionLoadInstallLogType::GetLogType() {
 	child_list_t<LogicalType> child_list = {
-	    {"event", LogicalType::VARCHAR},        {"extension", LogicalType::VARCHAR},
-	    {"version", LogicalType::VARCHAR},      {"install_mode", LogicalType::VARCHAR},
-	    {"source", LogicalType::VARCHAR},       {"reason", LogicalType::VARCHAR},
+	    {"event", LogicalType::VARCHAR},        {"extension", LogicalType::VARCHAR}, {"version", LogicalType::VARCHAR},
+	    {"install_mode", LogicalType::VARCHAR}, {"source", LogicalType::VARCHAR},    {"reason", LogicalType::VARCHAR},
 	    {"error", LogicalType::VARCHAR},
 	};
 	return LogicalType::STRUCT(child_list);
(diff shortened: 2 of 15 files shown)


src/main/extension/extension_install.cpp
```diff
@@ -21,6 +24,25 @@ namespace duckdb {
 //===--------------------------------------------------------------------===//
 // Install Extension
 //===--------------------------------------------------------------------===//
+
+//! Emit a structured log event for a completed install (no-op when nothing was installed)
+static void LogInstall(DatabaseInstance &db, const string &extension, optional_ptr<ExtensionInstallInfo> info,
+                       const string &reason) {
+	if (!info) {
```
lnkuiper, line 31: Could we distinguish a completed install from an ETag cache hit here? `InstallFromHttpUrl` returns the existing `install_info` for HTTP 304, so info is non-null even though no extension was installed. Codex reproduced this issue locally

src/main/extension/extension_install.cpp
```diff
@@ -196,7 +218,15 @@ unique_ptr<ExtensionInstallInfo> ExtensionHelper::InstallExtension(DatabaseInsta
 	return nullptr;
 #endif
 	string local_path = ExtensionDirectory(db, fs);
-	return InstallExtensionInternal(db, fs, local_path, extension, options);
+	auto extension_name = ExtensionHelper::GetExtensionName(extension);
+	try {
```
lnkuiper, line 222: Could the `try` cover the setup immediately above it as well? Both overloads call `ExtensionDirectory` before entering the failure-logging path. If directory creation fails, the INSTALL statement errors without emitting an event:
```sql
CALL enable_logging('ExtensionLoadInstall', storage='memory');
COPY (SELECT 1) TO '{TEST_DIR}/not_a_directory';
SET extension_directory='{TEST_DIR}/not_a_directory/subdirectory';
INSTALL no_such_extension;

SELECT count(*)
FROM duckdb_logs_parsed('ExtensionLoadInstall')
WHERE event = 'install' AND error IS NOT NULL;
```

src/execution/operator/helper/physical_load.cpp
```diff
@@ -45,6 +47,7 @@ SourceResultType PhysicalLoad::GetDataInternal(ExecutionContext &context, DataCh
 		ExtensionLoadOptions options;
 		options.extension_name = info->filename;
 		options.alias = info->alias;
+		options.reason = "explicit SQL LOAD statement";
```
lnkuiper, line 50: This reason is lost when the extension is statically linked:
```sql
CALL enable_logging('ExtensionLoadInstall', storage='memory');
LOAD tpch;

SELECT reason
FROM duckdb_logs_parsed('ExtensionLoadInstall')
WHERE extension = 'tpch';
-- NULL, expected "explicit SQL LOAD statement"
```

src/main/extension_manager.cpp
```diff
@@ -38,6 +43,9 @@ void ExtensionActiveLoad::LoadFail(const ErrorData &error) {
 		ExtensionManager::Get(db).RemoveExtensionInfo(alias.GetIdentifierName());
 	}
 	DUCKDB_LOG_INFO(db, "Failed to load extension '%s': %s", extension_name.GetIdentifierName(), error.Message());
+
+	DUCKDB_LOG(db, ExtensionLoadInstallLogType, "load", extension_name.GetIdentifierName(), "", "", "", reason,
```
lnkuiper, line 47: This only records failures after `BeginLoad` has returned an `ExtensionActiveLoad`:
```sql
LOAD '{BUILD_DIRECTORY}/test/extension/loadable_extension_demo.duckdb_extension' AS demo;
CALL truncate_duckdb_logs();

LOAD '{BUILD_DIRECTORY}/test/extension/loadable_extension_demo.duckdb_extension' AS demo;

SELECT count(*)
FROM duckdb_logs_parsed('ExtensionLoadInstall')
WHERE event = 'load' AND error IS NOT NULL;
-- 0, expected 1
```

src/logging/log_types.cpp
```diff
@@ -360,4 +361,35 @@ string AsyncTaskScheduleLogType::ConstructLogMessage(const string &pool, idx_t t
...
+                                                        const string &source, const string &reason,
+                                                        const string &error) {
+	auto or_null = [](const string &value) {
+		return value.empty() ? Value() : Value(value);
+	};
+	child_list_t<Value> child_list = {
+	    {"event", Value(event)},       {"extension", or_null(extension_name)},
+	    {"version", or_null(version)}, {"install_mode", or_null(install_mode)},
+	    {"source", or_null(source)},   {"reason", or_null(reason)},
+	    {"error", or_null(error)},
```
lnkuiper, line 425: Maybe worth adding the `alias` extension field here?

lnkuiper (changes requested):
Thanks! I've left some comments:

src/execution/operator/helper/physical_load.cpp
carlopi: This is expected, LOAD tpch is a nop at that point, like the actual load happened way earlier.

carlopi:
Resolved the conflicts, but still not answered all comments.

PR #23912, opened 2026-07-18
Improve pretty printing for logging to CLI

carlopi:
More compact / less verbose and arguably less ugly on the eye logging mode only for human consumption.

Example:
image

3 components:
- LEVEL:type left-aligned to 20 cols, colored via the configurable LOG_* palette
- elapsed since CLI launch, right-aligned to 8 cols, 2 decimals, scale-adaptive
  (s -> m's -> h'm''), rounded before decomposing so no impossible 60.00 boundary
- message sanitized to a single line (multi-line SQL / control chars collapsed)

Idea is this IS now bolted on the default CLI log, with the following rules:
* ERROR and WARNING, as before multiline
* WARNING, as before, de-duped
* ERROR, changed from before, NOT deduped anymore, all ERROR will be shown\
* INFO or less: compact single line representation

### DIFF
--- a/tools/shell/include/shell_renderer.hpp
+++ b/tools/shell/include/shell_renderer.hpp
@@ -162,7 +162,7 @@ public:
 
 class ShellLogStorage : public duckdb::LogStorage {
 public:
-	explicit ShellLogStorage(ShellState &state) : shell_highlight(state) {};
+	explicit ShellLogStorage(ShellState &state);
 
 	~ShellLogStorage() override = default;
 
@@ -183,10 +183,13 @@ protected:
 private:
 	ShellHighlight shell_highlight;
 
-	// Logs that have already been printed to avoid spamming the shell
+	// Warnings/errors that have already been printed, to avoid spamming the shell (loud logs only)
 	duckdb::unordered_set<uint64_t> printed_logs;
 
-	// lock to ensure thread safety of the printed_logs set
+	// Timestamp (micros) at storage construction (CLI launch), the zero point for elapsed time
+	int64_t start_micros = 0;
+
+	// lock to ensure thread safety of the printed_logs set and the elapsed-time state
 	mutable duckdb::mutex lock;
 };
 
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks - I wonder if this should live in the CLI, though, since I suppose that's the main target for this mode?

carlopi:
I am not sure, like, would this have no value in say Python? I am not sure. Happy to move obv, I don't have a strong opinion.

carlopi:
@Mytherin: thanks for the push, I wrapped this into regular CLI logger, so that now by default it can be as simple as:
```sql
CALL enable_logging('QueryLog');
CREATE TABLE T AS SELECT 42 answer;
```
would before log:
```
memory D CALL enable_logging('QueryLog');
WARNING:
The logging settings have been changed so you may lose warnings printed in the CLI.
To continue printing warnings to the console, set storage='shell_log_storage'.
For more info see https://duckdb.org/docs/stable/operations_manual/logging/overview.

┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D CREATE TABLE T AS SELECT 42 answer;
INFO:
CREATE TABLE T AS SELECT 42 answer;

```
and after the PR:
```
memory D CALL enable_logging('QueryLog');
WARNING:
The logging settings have been changed so you may lose warnings printed in the CLI.
To continue printing warnings to the console, set storage='shell_log_storage'.
For more info see https://duckdb.org/docs/current/operations_manual/logging/overview.

┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D CREATE TABLE T AS SELECT 42 answer;
INFO:QueryLog          16.36  CREATE TABLE T AS SELECT 42 answer;
```

There is still the WARNING on change that can be reworked, but it went form 3 lines to 1 while adding disambiguation of log type (if multiple are enabled) and elapsed time since start of CLI process.

Mytherin:
Thanks!

PR #23917, opened 2026-07-18
fix: correct tpcd typo to tpch in ExtraTests.yml

arijitroy003:
ExtraTests.yml sets CORE_EXTENSIONS to `tpcd;tpcds;httpfs` but `tpcd` should be `tpch`. This means the TPC-H extension is not built for the regression test job, potentially causing TPC-H dependent tests to be skipped.


arijitroy003:
Friendly ping — is there anything needed from my side to move this forward? Happy to make any adjustments.

PR #23919, opened 2026-07-18
Fix bloom filter pruning for DECIMAL types

### LINKED ISSUE #23916: Parquet query doesn't leverage bloom filter on decimal types
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D COPY (
             SELECT (2 * (i % 2048))::DECIMAL(18,2) AS dec, i
             FROM range(8192) t(i)
         ) TO '/tmp/bloom_decimal.parquet' (
             FORMAT parquet,
             ROW_GROUP_SIZE 2048,
             DICTIONARY_SIZE_LIMIT 4096,
             BLOOM_FILTER_FALSE_POSITIVE_RATIO 0.0001
         );

-- the const doesn't exist in bloom filter, which should be pruned if bloom filter is used correctly
memory D SELECT bool_and(bloom_filter_excludes)
         FROM parquet_bloom_probe(
             '/tmp/bloom_decimal.parquet', 'dec', 1001.00::DECIMAL(18,2)
         );
┌─────────────────────────────────┐
│ bool_and(bloom_filter_excludes) │
│             boolean             │
├─────────────────────────────────┤
│ true                            │
└─────────────────────────────────┘

-- but we access all row groups
memory D EXPLAIN ANALYZE
         SELECT count(*)
         FROM '/tmp/bloom_decimal.parquet'
         WHERE dec = 1001.00;
╭─ Summary ───────────╮
│ Total Time: 0.0057s │
│ Data Read:  67.0 kB │
╰─────────────────────╯
╭─ Ungrouped Aggregate ───────────────────╮
│ 1 row                               0µs │
╰────────────────────┬────────────────────╯
╭─ Table Scan ───────┴────────────────────╮
│ Function: PARQUET_SCAN                  │
│ Filters: dec = 1001.00                  │
│ Total Files Read: 1                     │
│ Filename(s): /tmp/bloom_decimal.parquet │
│ Row Groups Scanned: 4 / 4               │
│ 0 rows                           10.0ms │
╰─────────────────────────────────────────╯
```

### OS:

macos

### DuckDB Version:

37b7f2c180b2e25cd89af3d679d69882385ac340

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

sdk011:
**Problem**: 
Issue #23916 reports that Parquet bloom filter pruning fails for `DECIMAL(18,2)` columns. This happens because `BloomFilterSupported` returned false for `DECIMAL`, causing the reader to silently skip the filter.

**Solution**: 
- Added `DECIMAL` to `BloomFilterSupported`.
- Threaded `parquet_physical_type` down to `ValueXXH64` so it can properly hash `INT32` and `INT64` backed decimals using `ValueXH64FixedWidth`.
- Explicitly disabled bloom filter hashing for `FIXED_LEN_BYTE_ARRAY`, `BYTE_ARRAY`, and `INT128` types by returning 0 in `ValueXXH64`. This is safe because DuckDB's `FixedDecimalColumnWriter` does not currently write bloom filters anyway. Returning 0 skips the check for files written by external tools, preventing false-pruning due to big-endian/little-endian hash mismatches.

**Testing**: 
Added `test/sql/copy/parquet/bloom_filter_decimal.test` which validates `parquet_bloom_probe` and scanner pruning for both INT32 and INT64 backed decimals.

Fixes #23916


sdk011:
@dentiny  - can you please have a look at this fix ?

extension/parquet/parquet_statistics.cpp
```diff
@@ -731,7 +731,13 @@ static uint64_t ValueXH64FixedWidth(const Value &constant) {
 // TODO we can only this if the parquet representation of the type exactly matches the duckdb rep!
 // TODO TEST THIS!
 // TODO perhaps we can re-use some writer infra here
-static uint64_t ValueXXH64(const Value &constant) {
+static uint64_t ValueXXH64(const Value &constant, duckdb_parquet::Type::type parquet_physical_type) {
+	if (constant.type().id() == LogicalTypeId::DECIMAL) {
+		if (parquet_physical_type == duckdb_parquet::Type::FIXED_LEN_BYTE_ARRAY ||
```
dentiny: Just curious, in which case will we have this case?
Don't think it possible, let's remove; or throw exception because they shouldn't happen.

extension/parquet/parquet_statistics.cpp
```diff
@@ -749,6 +755,8 @@ static uint64_t ValueXXH64(const Value &constant) {
 		return ValueXH64FixedWidth<uint64_t>(constant);
 	case PhysicalType::INT64:
 		return ValueXH64FixedWidth<int64_t>(constant);
+	case PhysicalType::INT128:
```
dentiny, line 860: Add a comment why we return 0?
Or even better, we should throw InvalidInput exception since it's not valid parquet

dentiny (commented):
Incorrect result
```sql
memory D COPY (
             SELECT
                 CASE WHEN i % 2 = 0
                     THEN 2.00::DECIMAL(18,2)
                     ELSE 4.00::DECIMAL(18,2)
                 END AS d
             FROM range(8192) t(i)
         ) TO '/tmp/decimal_bloom_repro.parquet' (
             FORMAT parquet,
             ROW_GROUP_SIZE 2048,
             DICTIONARY_SIZE_LIMIT 4096,
             BLOOM_FILTER_FALSE_POSITIVE_RATIO 0.000001
         );
memory D SELECT count(*)
         FROM '/tmp/decimal_bloom_repro.parquet'
         WHERE d = 2.00::DECIMAL(18,2);
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            0 │
└──────────────┘
```

extension/parquet/parquet_statistics.cpp
sdk011: We will hit this case when querying Parquet files generated by other engines (like Apache Spark or PyArrow).

extension/parquet/parquet_statistics.cpp
sdk011: added a comment.

sdk011:
> Incorrect result
> 
> ```sql
> memory D COPY (
>              SELECT
>                  CASE WHEN i % 2 = 0
>                      THEN 2.00::DECIMAL(18,2)
>                      ELSE 4.00::DECIMAL(18,2)
>                  END AS d
>              FROM range(8192) t(i)
>          ) TO '/tmp/decimal_bloom_repro.parquet' (
>              FORMAT parquet,
>              ROW_GROUP_SIZE 2048,
>              DICTIONARY_SIZE_LIMIT 4096,
>              BLOOM_FILTER_FALSE_POSITIVE_RATIO 0.000001
>          );
> memory D SELECT count(*)
>          FROM '/tmp/decimal_bloom_repro.parquet'
>          WHERE d = 2.00::DECIMAL(18,2);
> ┌──────────────┐
> │ count_star() │
> │    int64     │
> ├──────────────┤
> │            0 │
> └──────────────┘
> ```



 i have fixed this in the new commit

sdk011:
@dentiny  - thanks for reviewing. have addressed your comments

extension/parquet/parquet_statistics.cpp
dentiny: Yeah for sure, but I still don't understand in which case will it happen? Curious why do you prefer silently return instead of throwing exception?

sdk011: Spark and PyArrow (via arrow-cpp) write large decimals (precision > 18) by mapping the DECIMAL logical type strictly to the FIXED_LEN_BYTE_ARRAY physical type.

In Spark: ParquetWriteSupport.scala automatically writes large DecimalType values as FLBA.
In Arrow: schema.cc maps Decimal128/Decimal256 directly to Parquet's FLBA primitive type.
Because they generate these DECIMAL/FLBA columns with Bloom filters, when DuckDB users query these external files using a filter (e.g., WHERE dec = 123.45), DuckDB parses the constant as LogicalTypeId::DECIMAL and passes the metadata's FIXED_LEN_BYTE_ARRAY physical type into ValueXXH64. Returning 0 here is the way to go here  - if we don't, we will throw a fatal InternalException on the read path for perfectly valid files generated by the rest of the ecosystem.

sdk011:
@dentiny - please review and enable merge if all look good from your end. thanks

sdk011:
@lnkuiper - can you help review this ?

extension/parquet/parquet_statistics.cpp
```diff
@@ -813,6 +819,15 @@ static optional<uint64_t> ValueXXH64(const Value &constant, const ParquetColumnS
 	case LogicalTypeId::TIMESTAMP:
 	case LogicalTypeId::TIMESTAMP_TZ:
 		return TryHashTimestamp(constant, schema);
+	case LogicalTypeId::DECIMAL:
```
lnkuiper, line 829: This uses DuckDB’s storage width instead of Parquet’s physical width, yielding incorrect results:
```sql
require parquet

# Local fixture schema: required INT64 d (DECIMAL(4,2)); values alternate between 2.00 and 4.00.
query I
SELECT count(*)
FROM '/private/tmp/pr23919_decimal4_int64.parquet'
WHERE d = 2.00::DECIMAL(4,2);
----
4096
```

extension/parquet/parquet_statistics.cpp
```diff
@@ -899,6 +918,7 @@ bool ParquetStatisticsUtils::BloomFilterSupported(const ParquetColumnSchema &sch
 	case LogicalTypeId::VARCHAR:
 	case LogicalTypeId::BLOB:
 	case LogicalTypeId::DATE:
+	case LogicalTypeId::DECIMAL:
```
lnkuiper: I think we shouldn't return `true` by default here, we only support the bloom filter if it's int32/64 (not byte array like you have above) maybe create a helper function

lnkuiper (changes requested):
Thanks for the PR and changes! I have some more comments:

extension/parquet/parquet_statistics.cpp
sdk011: addressed this

extension/parquet/parquet_statistics.cpp
sdk011: addressed this

lnkuiper (approved):
Thanks for the changes! LGTM

PR #23920, opened 2026-07-18
Optimize GLOB using string stats for row group pruning

### LINKED ISSUE #23906: Glob operation doesn't leverage string stats to prune row group
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D ATTACH '/tmp/glob_prefix_prune.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE t AS
         SELECT CASE
                    WHEN range < 2048 THEN 'aaa' || range::VARCHAR
                    WHEN range < 4096 THEN 'mmm' || range::VARCHAR
                    ELSE 'zzz' || range::VARCHAR
                END AS s
         FROM range(6144);

-- we have min/max for all row groups, should scan 1 row groups
db D EXPLAIN ANALYZE
     SELECT count(*) FROM t WHERE s GLOB 'mmm*';
╭─ Summary ───────────╮
│ Total Time: 0.0036s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ─────╮
│ Aggregates: count_star()  │
│ 1 row                 0µs │
╰─────────────┒┎────────────╯
╭─ Table Scan ┚┖────────────╮
│ Table: db.main.t          │
│ Type: Sequential Scan     │
│ Filters: s ~~~ 'mmm*'     │
│ Row Groups Scanned: 3 / 3 │
│ 2,048 rows            0µs │
╰───────────────────────────╯

-- we have min/max for all row groups, should scan 0 row groups
db D EXPLAIN ANALYZE
     SELECT count(*) FROM t WHERE s GLOB 'qqq';
╭─ Summary ───────────╮
│ Total Time: 0.0005s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ─────╮
│ Aggregates: count_star()  │
│ 1 row                 0µs │
╰─────────────┬─────────────╯
╭─ Table Scan ┴─────────────╮
│ Table: db.main.t          │
│ Type: Sequential Scan     │
│ Filters: s ~~~ 'qqq'      │
│ Row Groups Scanned: 3 / 3 │
│ 0 rows                0µs │
╰───────────────────────────╯
```

### OS:

macos

### DuckDB Version:

37b7f2c180b2e25cd89af3d679d69882385ac340

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

sdk011:
Fixes #23906

This PR extends the `LikeOptimizationRule` to also handle `GLOB` (`~~~`) and `NOT GLOB` (`!~~~`) expressions. 

When a GLOB expression contains a constant prefix, suffix, or contains-pattern (e.g., `GLOB 'mmm*'`), it is now rewritten to use the optimized `prefix()`, `suffix()`, or `contains()` scalar functions. This allows the execution engine to leverage zone maps (string min/max statistics) and prune row groups rather than performing full sequential scans.

**Changes:**
- Included `~~~` and `!~~~` in the `LikeOptimizationRule` matcher.
- Parameterized wildcard boundary checks to support GLOB wildcards (`*`, `?`, `[`, `]`, `\`) versus LIKE wildcards (`%`, `_`).
- Added a test case (`test_glob_row_group_pruning.test`) to explicitly verify row group pruning on `GLOB` queries.

### DIFF
--- a/src/include/duckdb/optimizer/rule/like_optimizations.hpp
+++ b/src/include/duckdb/optimizer/rule/like_optimizations.hpp
@@ -22,7 +22,7 @@ public:
 	                             bool is_root) override;
 
 	unique_ptr<Expression> ApplyRule(BoundFunctionExpression &expr, const ScalarFunction &function, string pattern,
-	                                 bool is_not_like) const;
+	                                 bool is_not_like, bool is_glob) const;
 };
 
 } // namespace duckdb
--- a/src/include/duckdb/optimizer/rule/like_optimizations.hpp
+++ b/src/include/duckdb/optimizer/rule/like_optimizations.hpp
@@ -13,6 +13,8 @@
 
 namespace duckdb {
 
+enum class PatternMatchType : uint8_t { LIKE, GLOB };
+
 // The Like Optimization rule rewrites LIKE to optimized scalar functions (e.g.: prefix, suffix, and contains)
 class LikeOptimizationRule : public Rule {
 public:
@@ -22,7 +24,7 @@ public:
 	                             bool is_root) override;
 
 	unique_ptr<Expression> ApplyRule(BoundFunctionExpression &expr, const ScalarFunction &function, string pattern,
-	                                 bool is_not_like, bool is_glob) const;
+	                                 bool is_not_like, PatternMatchType match_type) const;
 };
 
 } // namespace duckdb
--- a/test/sql/optimizer/test_glob_row_group_pruning.test
+++ b/test/sql/optimizer/test_glob_row_group_pruning.test
@@ -1,0 +1,54 @@
+# name: test/sql/optimizer/test_glob_row_group_pruning.test
+# description: Row-group pruning on string columns using GLOB
+# group: [optimizer]
+
+statement ok
+ATTACH '{TEST_DIR}/glob_prune.duckdb' AS db (ROW_GROUP_SIZE 2048);
+
+statement ok
+USE db;
+
+statement ok
+CREATE TABLE t AS
+         SELECT CASE
+                    WHEN range < 2048 THEN 'aaa' || range::VARCHAR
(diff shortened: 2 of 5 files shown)


sdk011:
@dentiny  - can you please have a look at this fix ?

test/sql/optimizer/test_glob_row_group_pruning.test
```diff
@@ -0,0 +1,54 @@
...
+----
+analyzed_plan	<REGEX>:.*Row Groups Scanned: 3 / 3.*
+
+query I
+SELECT count(*) FROM t WHERE s NOT GLOB 'qqq';
+----
+6144
+
+query I
+SELECT count(*) FROM t WHERE s NOT GLOB 'qqq';
```
dentiny: Why duplicate tests?

sdk011: removed them

src/include/duckdb/optimizer/rule/like_optimizations.hpp
```diff
@@ -22,7 +22,7 @@ class LikeOptimizationRule : public Rule {
 	                             bool is_root) override;
 
 	unique_ptr<Expression> ApplyRule(BoundFunctionExpression &expr, const ScalarFunction &function, string pattern,
-	                                 bool is_not_like) const;
+	                                 bool is_not_like, bool is_glob) const;
```
lnkuiper: We try not to add many `bool` arguments like this, usually it's a lot easier to read and less error-prone to use one or more `enum`s to do this.

sdk011: addressed this in the new commit

PR #23926, opened 2026-07-19
Add monotonicity properties to scalar math functions

hacoobamatata:
## Summary

Adds `StrictlyIncreasing()` argument properties to `atan`, `sinh`, `asinh`, and `tanh`, enabling statistics propagation and row-group pruning for these functions.

## Tests

Added regression tests to `monotone_function_zonemap_pruning.test`.

### DIFF
--- a/extension/core_functions/scalar/math/numeric.cpp
+++ b/extension/core_functions/scalar/math/numeric.cpp
@@ -1518,7 +1518,7 @@ struct ATanOperator {
 ScalarFunction AtanFun::GetFunction() {
 	ScalarFunction function({LogicalType::DOUBLE}, LogicalType::DOUBLE,
 	                        ScalarFunction::UnaryFunction<double, double, ATanOperator>);
-	function.SetUnaryArgProperties(ArgProperties().StrictlyIncreasing());
+	function.SetUnaryArgProperties(ArgProperties().NonDecreasing());
 	return function;
 }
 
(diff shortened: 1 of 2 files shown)


Mytherin (commented):
Thanks for the PR! Looks fine to me but I think this should be `NonDecreasing()` due to floating point semantics, e.g.:

```sql
SELECT 2.0::DOUBLE AS input, nextafter(2.0::DOUBLE, 'inf') AS next_input,
                atan(2.0::DOUBLE) atan_input, atan(nextafter(2.0::DOUBLE, 'inf')) AS atan_next_input;
┌────────┬────────────────────┬────────────────────┬────────────────────┐
│ input  │     next_input     │     atan_input     │  atan_next_input   │
│ double │       double       │       double       │       double       │
├────────┼────────────────────┼────────────────────┼────────────────────┤
│    2.0 │ 2.0000000000000004 │ 1.1071487177940906 │ 1.1071487177940906 │
└────────┴────────────────────┴────────────────────┴────────────────────┘
```

hacoobamatata:
Good catch, thanks! I've updated the annotations to `NonDecreasing()` and rerun the tests, which all passed.

hacoobamatata:
Hi! While working on the `NonDecreasing()` change, I checked the existing functions currently marked `StrictlyIncreasing()` using adjacent `DOUBLE` values via `nextafter`. I found counterexamples for all of `sqrt`, `cbrt`, `exp`, `ln`, `log2`, and `log10`, where distinct adjacent inputs produce equal outputs. Would a separate cleanup PR changing these annotations to `NonDecreasing()` be welcome?

Mytherin:
Yes, that would be great - thanks

PR #23943, opened 2026-07-19
Fix TopN late materialization with asymmetric join projections

### LINKED ISSUE #23925: `TopNWindowElimination` optimizer crashes on join + geometry predicate + `qualify`
### What happens?

A query that joins two tables, filters with a `spatial` extension predicate involving `GEOMETRY` (`ST_Within(ST_Point(...), geom)`), and picks the top row per group via `QUALIFY row_number() OVER (PARTITION BY ... ORDER BY ...) = 1` throws an internal assertion error from the `TopNWindowElimination` optimizer pass, rather than either returning results or a normal binder/planner error.

### To Reproduce

Minimal reproduction:

```py
import duckdb

con = duckdb.connect()
con.execute("INSTALL spatial")
con.execute("LOAD spatial")

con.execute("CREATE TABLE cand(id INT, k INT, x DOUBLE, y DOUBLE)")
con.execute("INSERT INTO cand VALUES (1,0,5,5),(1,1,1,1),(1,2,2,2),(2,0,9,9)")
con.execute("CREATE TABLE poly(id INT, geom GEOMETRY)")
con.execute("INSERT INTO poly SELECT 1, ST_GeomFromText('POLYGON((0 0,3 0,3 3,0 3,0 0))')")
con.execute("INSERT INTO poly SELECT 2, ST_GeomFromText('POLYGON((0 0,3 0,3 3,0 3,0 0))')")

con.execute("""
    SELECT c.id, c.x, c.y
    FROM cand c
    JOIN poly p USING (id)
    WHERE ST_Within(ST_Point(c.x, c.y), p.geom)
    QUALIFY row_number() OVER (PARTITION BY c.id ORDER BY c.k) = 1
""").fetchall()
```

Actual results:

```sh
duckdb.duckdb.InternalException: INTERNAL Error: Attempted to access index 2 within vector of size 2

Stack Trace:
...
_ZNK6duckdb21TopNWindowElimination23CreateAggregateOperatorERNS_13LogicalWindowE...
_ZN6duckdb21TopNWindowElimination16OptimizeInternalE...
_ZN6duckdb21TopNWindowElimination8OptimizeE...
...
For more information, see https://duckdb.org/docs/stable/dev/internal_errors
```

Expected results: `[(1, 1.0, 1.0)]`

Isolation notes:

- Removing the `spatial`-extension predicate (`ST_Within(ST_Point(...), geom)`) and replacing it with an equivalent plain-SQL predicate (e.g. `BETWEEN`) on the same join+QUALIFY shape does **not** crash — so this isn't just "join + QUALIFY," it specifically involves a `GEOMETRY`-typed filter predicate feeding into the window-elimination rewrite.
- Simply `LOAD`ing the `spatial` extension without using any spatial function in the query does not crash — confirms it's the `GEOMETRY`/`ST_*` usage in the filter, not extension loading itself.
- Replacing the `QUALIFY row_number() OVER (...) = 1` with an equivalent `GROUP BY ... arg_min(x, k), arg_min(y, k)` aggregation avoids the crash and returns correct results — this is a viable workaround, but confirms the bug is specific to this window-function code path.

Workaround:

Use `arg_min(col, order_col)` aggregates instead of `QUALIFY row_number() OVER (...) = 1` when picking a top-1-per-group row downstream of a `spatial` `GEOMETRY` predicate.

### OS:

macOS 26.5.1, arm64

### DuckDB Version:

 v1.5.4 (08e34c447b, Variegata)

### DuckDB Client:

Python (duckdb==1.5.4), Python 3.12.12

### Hardware:

_No response_

### Full Name:

Robert Martin

### Affiliation:

n/a

- [x] Yes, I have

- [x] Yes, I have

souvikDevloper:
## Summary

- skip Top-N window late materialization for the unsafe asymmetric join projection-map case
- add a core-only regression test for a post-join filter followed by `QUALIFY row_number()`

## Root cause

An empty projection map means "project all columns". When only the other join side had an explicit projection map, late materialization appended a row ID to the selected side's empty map. That changed its meaning to "project only the row ID", dropped payload columns, and caused an out-of-range binding access.

The fix conservatively falls back to the existing `struct_pack` path only for that shape. The broader Top-N window rewrite remains enabled.

Fixes #23925.

## Validation

- reproduced the original spatial failure on DuckDB v1.5.4
- reproduced the same failure with the new core-only regression on the unpatched build
- `git diff --check`
- `build/pr/test/unittest.exe test/sql/optimizer/topn_window_elimination_late_materialization.test` (98 assertions)

### DIFF
--- a/src/optimizer/topn_window_elimination.cpp
+++ b/src/optimizer/topn_window_elimination.cpp
@@ -1018,6 +1018,13 @@ bool TopNWindowElimination::CanUseLateMaterialization(const LogicalWindow &windo
 				// We cannot use late materialization by scanning a single table.
 				return false;
 			}
+			if (join.HasProjectionMap()) {
+				const auto &projection_map =
+				    all_right_replaceable ? join.left_projection_map : join.right_projection_map;
+				if (projection_map.empty()) {
+					return false;
+				}
+			}
 
 			TableIndex replace_table_idx = all_right_replaceable ? right_idx : left_idx;
 			for (idx_t i = 0; i < projections.size(); i++) {
--- a/test/sql/optimizer/topn_window_elimination_late_materialization.test
+++ b/test/sql/optimizer/topn_window_elimination_late_materialization.test
@@ -151,3 +151,31 @@ WITH cte AS (SELECT a, b, c::TIMESTAMPTZ AS c FROM casted)
 SELECT * FROM cte QUALIFY row_number() OVER (PARTITION BY a ORDER BY c DESC) = 1
 ----
 x	false	2024-01-01 00:00:00+00
+
+# A join can have an explicit projection map on only the other side.
+statement ok
+CREATE TABLE filter_left (id INTEGER, sort_key INTEGER, x INTEGER, y INTEGER)
+
+statement ok
+INSERT INTO filter_left VALUES (1, 0, 5, 5), (1, 1, 1, 1)
+
+statement ok
+CREATE TABLE filter_right (id INTEGER, z INTEGER)
+
+statement ok
+INSERT INTO filter_right VALUES (1, 1)
+
+statement ok
+SET disabled_optimizers = 'filter_pushdown,join_order'
+
+query III
+SELECT l.id, l.x, l.y
+FROM filter_left l
+JOIN filter_right r USING (id)
+WHERE l.x + r.z < 5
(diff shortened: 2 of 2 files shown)


souvikDevloper:
@Mytherin take a look over it. also want to join the duck db org as remote collaborator i mean member to help it make world's best db. love to learn from you.

PR #23944, opened 2026-07-19
Skip Parquet bloom pruning for NaN

### LINKED ISSUE #23921: DuckDB treats NaN differently for DuckDB table and parquet file
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE nan_payload AS
         SELECT '-NaN'::DOUBLE AS x FROM range(5000);
memory D COPY nan_payload TO '/tmp/bf_nan.parquet' (
           FORMAT PARQUET,
           ROW_GROUP_SIZE 2048,
           WRITE_BLOOM_FILTER TRUE
         );

-- DuckDB table returns 5000
memory D SELECT count(*) FROM nan_payload WHERE x = 'NaN'::DOUBLE;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│         5000 │
└──────────────┘

-- parquet returns 0
memory D SELECT count(*)
         FROM '/tmp/bf_nan.parquet'
         WHERE x = 'NaN'::DOUBLE;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            0 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

souvikDevloper:
## Summary

- skip Parquet bloom-filter pruning for `FLOAT` and `DOUBLE` NaN equality constants
- prevent raw NaN payload differences from excluding matching row groups
- add a regression test covering the reported multi-row-group scan

## Root cause

DuckDB considers NaN values equal regardless of their IEEE payload, while Parquet bloom filters hash their raw bytes. Probing a bloom filter with one NaN encoding therefore cannot safely exclude a row group that may contain another encoding.

## Validation

- `unittest test/sql/copy/parquet/bloom_filters.test` (97 assertions passed)
- manually reproduced the original `0` result before the change and verified `5000` rows after it

Fixes #23921

### DIFF
--- a/extension/parquet/parquet_statistics.cpp
+++ b/extension/parquet/parquet_statistics.cpp
@@ -877,19 +877,30 @@ static optional<uint64_t> ValueXXH64(const Value &constant, const ParquetColumnS
 static bool BloomFilterExcludes(const Value &constant, ParquetBloomFilter &bloom_filter,
                                 const ParquetColumnSchema &schema) {
 	// Floating-point equality treats positive and negative zero as equal, but Parquet hashes their bit patterns.
+	// Likewise, all NaN payloads compare equal, so no single hash can rule out a row group containing one.
 	switch (constant.type().InternalType()) {
-	case PhysicalType::FLOAT:
-		if (constant.GetValue<float>() == 0.0f) {
+	case PhysicalType::FLOAT: {
+		auto float_value = constant.GetValue<float>();
+		if (Value::IsNan(float_value)) {
+			return false;
+		}
+		if (float_value == 0.0f) {
 			return !bloom_filter.FilterCheck(ValueXH64FixedWidth(0.0f)) &&
 			       !bloom_filter.FilterCheck(ValueXH64FixedWidth(-0.0f));
 		}
 		break;
-	case PhysicalType::DOUBLE:
-		if (constant.GetValue<double>() == 0.0) {
+	}
+	case PhysicalType::DOUBLE: {
+		auto double_value = constant.GetValue<double>();
+		if (Value::IsNan(double_value)) {
+			return false;
+		}
+		if (double_value == 0.0) {
 			return !bloom_filter.FilterCheck(ValueXH64FixedWidth(0.0)) &&
 			       !bloom_filter.FilterCheck(ValueXH64FixedWidth(-0.0));
 		}
 		break;
+	}
 	default:
 		break;
 	}
--- a/test/sql/copy/parquet/bloom_filters.test
+++ b/test/sql/copy/parquet/bloom_filters.test
@@ -651,3 +651,33 @@ FROM read_parquet('{TEST_DIR}/bloom_ndf_signed_zero.parquet')
 WHERE x IS NOT DISTINCT FROM -0.0;
 ----
 2048
+
+statement ok
+COPY (
+	SELECT CASE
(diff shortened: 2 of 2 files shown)


souvikDevloper:
@Mytherin take a look over it. also want to join the duck db org as remote collaborator i mean member to help it make world's best db. love to learn from you.

Mytherin (commented):
Thanks - if this is the case should we then not just skip the bloom filter check for NaN entirely instead of only having the hash function return zero?

`NaN` can have different byte representations, and if a file is written by a different (non-DuckDB) writer we don't know which representation made it into the bloom filter.

souvikDevloper:
@Mytherin thanks for pointing out let me revisit the approach!

souvikDevloper:
Thanks, updated as suggested. NaN is now detected in ApplyBloomFilter, and bloom-filter pruning is skipped entirely before hashing. ValueXXH64 is back to hashing only, so this also safely handles arbitrary NaN representations from external Parquet writers. The regression test still passes. @Mytherin also love to join the fantastic org what your thoughts about it?

Mytherin:
You cannot join the organization as it doesn't have members. You can apply to work for our company DuckLabs - see https://ducklabs.com.

souvikDevloper:
@Mytherin merge  now

Mytherin:
Can you fix the merge conflicts?

souvikDevloper:
Sure

souvikDevloper:
@Mytherin conflicts are resolved — the branch is rebased onto current `main` (`8806a9f863`) and now merges cleanly.

The conflict came from `BloomFilterExcludes()`, which landed on `main` with the signed-zero fix in exactly the spot this PR was editing. Rather than replay the old commits, I rebased and folded the NaN check into that helper, next to the ±0.0 case — it's the same underlying issue (floating-point equality semantics not matching bit-pattern hashing), so it belongs in one place:

```cpp
case PhysicalType::DOUBLE: {
	auto double_value = constant.GetValue<double>();
	if (Value::IsNan(double_value)) {
		return false;
	}
	if (double_value == 0.0) {
		return !bloom_filter.FilterCheck(ValueXH64FixedWidth(0.0)) &&
		       !bloom_filter.FilterCheck(ValueXH64FixedWidth(-0.0));
	}
	break;
}
```

This keeps the behaviour you asked for in review: pruning is skipped entirely for NaN constants rather than forcing the hash to zero, so files written by non-DuckDB writers using a different NaN representation are handled correctly. `ValueXXH64()` is untouched and remains hashing-only.

The test moved to sit alongside the signed-zero block and mirrors its structure — NaN payloads in row group 0, ±1.0 in row group 1 — covering both `= 'NaN'` and `IS NOT DISTINCT FROM 'NaN'`.

Verification, local release build (MSVC 2022, `CORE_EXTENSIONS=parquet`):

- `bloom_filters.test` — 229 assertions passed
- `test/sql/copy/parquet/*` — 18985 assertions in 206 test cases passed
- Reverted just the `parquet_statistics.cpp` change and re-ran: the new test fails with `0` instead of `2048`, confirming it is a real regression guard rather than a test that passes either way
- `clang-format` 11.0.1 reports no diff

History is squashed to one commit. Ready for another look when you have a moment.

PR #23945, opened 2026-07-19
Capi add allocation handlers

ProjectMutilation:
Several C-API functions allocated their output via duckdb_malloc/malloc and then dereferenced the result unconditionally — memcpy into the buffer and/or a NUL-terminator write. If allocation failed, this was
  undefined behavior (NULL dereference). A couple of these sites guarded the pointer only with assert(), which is a no-op in release builds.

  This PR adds a NULL check after each such allocation and returns the C-API error sentinel on failure, matching how the C API already represents an "unavailable value" — so the affected call degrades gracefully instead of crashing.


Mytherin:
Thanks!

ProjectMutilation:
@Mytherin hi, [this job](https://github.com/duckdb/duckdb/actions/runs/29742121125/job/88351086227) has failed, there is a connectivity problem.

PR #23946, opened 2026-07-19
Match secret SCOPE only on URI component boundaries

### LINKED ISSUE #22660: HTTP/S3 secret SCOPE uses raw string-prefix matching and can apply credentials to sibling hosts/buckets
### What happens?

DuckDB secret scope matching appears to use raw string-prefix matching.

For HTTP secrets, this can apply a secret scoped to one parsed host to a different parsed host whose URL only shares the same string prefix.

Example:

- Secret scope: `http://127.0.0.1`
- Requested URL: `http://127.0.0.10:<port>/probe.txt`

`which_secret()` selects the scoped secret, and `httpfs` then sends the secret's `Authorization` header to `Host: 127.0.0.10:<port>`.

This also affects S3-style names such as:

- `s3://trusted`
- `s3://trusted-evil/...`

Expected behavior: URI-like scopes should respect scheme/host/port or bucket boundaries before applying credentials.

### To Reproduce

poc.py
```
#!/usr/bin/env python3
import argparse
import json
import socket
import threading
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

import duckdb


TOKEN = "DDB006_HTTP_TOKEN"
EVENTS = []


class Handler(BaseHTTPRequestHandler):
    def do_HEAD(self):
        EVENTS.append({
            "method": self.command,
            "host": self.headers.get("Host", ""),
            "authorization": self.headers.get("Authorization", ""),
        })
        body = b"hello\n"
        self.send_response(200)
        self.send_header("Content-Type", "text/plain")
        self.send_header("Content-Length", str(len(body)))
        self.send_header("Accept-Ranges", "bytes")
        self.end_headers()

    def log_message(self, *args):
        pass


def free_port():
    with socket.socket() as s:
        s.bind(("127.0.0.1", 0))
        return s.getsockname()[1]


def one(con, sql):
    return con.execute(sql).fetchone()[0]


def main():
    ap = argparse.ArgumentParser()
    ap.add_argument("--db", default="repro.duckdb")
    args = ap.parse_args()

    port = free_port()
    server = ThreadingHTTPServer(("0.0.0.0", port), Handler)
    thread = threading.Thread(target=server.serve_forever, daemon=True)
    thread.start()

    try:
        con = duckdb.connect(args.db)
        con.execute("INSTALL httpfs")
        con.execute("LOAD httpfs")
        con.execute(f"""
            CREATE OR REPLACE SECRET h (
              TYPE HTTP,
              SCOPE 'http://127.0.0.1',
              BEARER_TOKEN '{TOKEN}'
            )
        """)

        control = one(con, f"SELECT name FROM which_secret('http://127.0.0.1:{port}/x', 'http')")
        bypass = one(con, f"SELECT name FROM which_secret('http://127.0.0.10:{port}/x', 'http')")
        con.execute(f"SELECT filename, size FROM read_blob('http://127.0.0.10:{port}/x')").fetchall()

        con.execute("""
            CREATE OR REPLACE SECRET s3_bucket (
              TYPE S3,
              KEY_ID 'AKIA_DDB006',
              SECRET 'DDB006_S3_SECRET',
              SCOPE 's3://trusted'
            )
        """)
        s3_control = one(con, "SELECT name FROM which_secret('s3://trusted/data.parquet', 's3')")
        s3_bypass = one(con, "SELECT name FROM which_secret('s3://trusted-evil/data.parquet', 's3')")
        con.close()

        leaked = any(e["host"].startswith(f"127.0.0.10:{port}") and e["authorization"] == f"Bearer {TOKEN}"
                     for e in EVENTS)
        print(json.dumps({
            "http_scope": "http://127.0.0.1",
            "requested_host": f"127.0.0.10:{port}",
            "which_secret_control": control,
            "which_secret_bypass": bypass,
            "captured_requests": EVENTS,
            "s3_scope": "s3://trusted",
            "s3_control": s3_control,
            "s3_bypass": s3_bypass,
            "result": "VULNERABLE" if leaked else "NOT_REPRODUCED",
        }, indent=2))
    finally:
        server.shutdown()


if __name__ == "__main__":
    main()
```

Run:
`python poc.py --db repro.duckdb`

### OS:

Windows 11 x64

### DuckDB Version:

v1.5.2

### DuckDB Client:

DuckDB CLI

### Hardware:

Not performance-related

### Full Name:

oxsignal

### Affiliation:

Independent security researcher

- [x] Yes, I have

- [x] Yes, I have

Yes

gaurav0107:
### Summary

`BaseSecret::MatchScore` selected a secret whenever the requested path had the secret's `SCOPE` as a raw string prefix (`StringUtil::StartsWith`), so a secret could be sent to a host it was never scoped to:

- `SCOPE 'http://127.0.0.1'` was applied to `http://127.0.0.10:PORT/...`
- `SCOPE 's3://trusted'` was applied to `s3://trusted-evil/...`
- `SCOPE 'https://duckdb.org'` was applied to `https://duckdb.org:quack@evil.com`, where the scope stops inside the *userinfo* and the request is actually sent to `evil.com` (raised by @carlopi in review)
- `SCOPE 'http'` was applied to every `http://` URL

### Fix

Scope and path are now split into **scheme, authority and path** and compared component by component, rather than as one flat string:

- **scheme** must be equal, and must be a valid RFC 3986 scheme — so a local path that merely contains `://` is not mistaken for a URI
- **authority** must designate the host the request is actually sent to. A scope may leave off the port (`http://localhost` still covers `http://localhost:8080`), but it may not stop inside a longer authority unless only a port follows — which is what closes the userinfo bypass
- **path** segments are compared whole, so `/v2` no longer covers `/v20/...`

Behaviour that is deliberately preserved:

- `s3://bucket` -> `s3://bucket/key`, and a trailing-separator scope like `http://host/data/`
- a scheme-wide scope (`s3://`) still covers every host of that scheme
- an empty scope still matches everything, and the longest matching scope still wins
- non-URI scopes (local paths) keep plain prefix semantics

### Test plan

`test/sql/secrets/secret_scope_boundary.test` (sqllogictest, no extension or network needed) drives the matcher through `which_secret()` and covers every case above — both the bypasses and the matches that must keep working. It is picked up by the standard `test/unittest` run.

Verified locally on macOS/arm64:

- `build/release/test/unittest test/sql/secrets/secret_scope_boundary.test` — 35 assertions pass
- full fast suite `build/release/test/unittest` — **all tests passed (981140 assertions in 4646 test cases)**, no regressions
- `scripts/format.py --check` clean on both files with the pinned clang-format 11.0.1

`MatchScore` has a single caller (`secret_storage.cpp`), so the blast radius is limited to secret selection.

Fixes #22660

### DIFF
--- a/src/main/secret/secret.cpp
+++ b/src/main/secret/secret.cpp
@@ -11,6 +11,25 @@
 
 namespace duckdb {
 
+static bool IsScopeSeparator(char c) {
+	// a scope prefix may only match a path at a URI component boundary
+	return c == '/' || c == ':' || c == '?' || c == '#';
+}
+
+static bool ScopeMatches(const string &path, const string &prefix) {
+	if (prefix.empty()) {
+		return true;
+	}
+	if (!StringUtil::StartsWith(path, prefix)) {
+		return false;
+	}
+	if (path.size() == prefix.size()) {
+		return true;
+	}
+	// a strict prefix must end on a component boundary (avoids 's3://trusted' matching 's3://trusted-evil')
+	return IsScopeSeparator(prefix.back()) || IsScopeSeparator(path[prefix.size()]);
+}
+
 int64_t BaseSecret::MatchScore(const string &path) const {
 	int64_t longest_match = NumericLimits<int64_t>::Minimum();
 
@@ -24,7 +43,7 @@ int64_t BaseSecret::MatchScore(const string &path) const {
 			longest_match = 0;
 			continue;
 		}
-		if (StringUtil::StartsWith(path, prefix)) {
+		if (ScopeMatches(path, prefix)) {
 			longest_match = MaxValue<int64_t>(NumericCast<int64_t>(prefix.length()), longest_match);
 		}
 	}
--- a/src/main/secret/secret.cpp
+++ b/src/main/secret/secret.cpp
@@ -11,23 +11,92 @@
 
 namespace duckdb {
 
-static bool IsScopeSeparator(char c) {
-	// a scope prefix may only match a path at a URI component boundary
-	return c == '/' || c == ':' || c == '?' || c == '#';
+//! Terminates the authority component and separates path segments
+static bool IsPathSeparator(char c) {
+	return c == '/' || c == '?' || c == '#';
 }
 
-static bool ScopeMatches(const string &path, const string &prefix) {
+//! Locates the authority of a URI: everything between "scheme://" and the first '/', '?' or '#'
+//! Returns false for anything that is not a URI, such as a plain path or a bare bucket name
+static bool TryGetAuthority(const string &uri, idx_t &authority_begin, idx_t &authority_end) {
+	auto scheme_end = uri.find("://");
+	if (scheme_end == string::npos || scheme_end == 0) {
+		return false;
+	}
+	// RFC 3986: scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
+	if (!StringUtil::CharacterIsAlpha(uri[0])) {
+		return false;
+	}
+	for (idx_t i = 1; i < scheme_end; i++) {
+		auto c = uri[i];
+		if (!StringUtil::CharacterIsAlphaNumeric(c) && c != '+' && c != '-' && c != '.') {
+			return false;
+		}
+	}
+	authority_begin = scheme_end + 3;
+	authority_end = uri.size();
+	for (idx_t i = authority_begin; i < uri.size(); i++) {
+		if (IsPathSeparator(uri[i])) {
+			authority_end = i;
+			break;
+		}
+	}
+	return true;
+}
+
+//! A scope path must cover the requested path entirely or stop on a segment boundary within it
+static bool PathPrefixMatches(const string &path, const string &prefix) {
 	if (prefix.empty()) {
 		return true;
 	}
 	if (!StringUtil::StartsWith(path, prefix)) {
 		return false;
 	}
-	if (path.size() == prefix.size()) {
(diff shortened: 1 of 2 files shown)


carlopi (commented):
I had also a look at what was the underlying problem.

I think this is an improvement, but there are still open issues with userinfo@evil.com, like:

Example: secret is on scope `https://duckdb.org`, URL is `https://duckdb.org:quack@evil.com`.
I think this would currently match, but then send the request to evil.com.
I think to solve those string matching is not enough, and one need to introduce some path utilities to normalize and compare paths.

gaurav0107:
Thanks — you're right, and that case is worse than the one in the issue: the request genuinely goes to `evil.com`, so the boundary check was still just string matching with extra steps.

Rewrote it to compare components instead. Scope and path are split into scheme / authority / path, and each is compared separately. The authority rule is that a scope may leave off the port, but may not stop inside a longer authority unless *only* a port follows — so `https://duckdb.org` no longer matches `https://duckdb.org:quack@evil.com`, and neither does `https://duckdb.org:`, nor `https://example.org:443` against `https://example.org:443@evil.com`.

Two more holes fell out of the same rewrite:

- `SCOPE 'http'` matched every `http://` URL, because the scope ended inside the scheme. It no longer does; `s3://` still covers every s3 host as before.
- path segments are now compared whole, so `/v2` no longer covers `/v20/...`.

Scheme validation follows RFC 3986, which also keeps a local path that happens to contain `://` from being read as a URI — non-URI scopes still get plain prefix semantics.

All of this is in `test/sql/secrets/secret_scope_boundary.test`. Locally the full fast suite is green (981140 assertions in 4646 test cases) and `format.py --check` is clean under the pinned clang-format 11.0.1, so I've taken it out of draft.

PR #23947, opened 2026-07-20
Optimizer: restrict equal-or-null rewrite to predicates

### LINKED ISSUE #23685: Incorrect result for NULL / boolean operation
### What happens?

### To Reproduce

```sql
memory D CREATE TABLE t(a INT, b INT);
memory D INSERT INTO t VALUES (NULL, 1), (1, NULL), (NULL, NULL), (1, 1), (1, 2);
memory D SELECT a, b, (a = b) OR (a IS NULL AND b IS NULL) AS r
         FROM t
         ORDER BY a NULLS FIRST, b NULLS FIRST;
┌───────┬───────┬─────────┐
│   a   │   b   │    r    │
│ int32 │ int32 │ boolean │
├───────┼───────┼─────────┤
│  NULL │  NULL │ true    │
│  NULL │     1 │ false   │
│     1 │  NULL │ false   │
│     1 │     1 │ true    │
│     1 │     2 │ false   │
└───────┴───────┴─────────┘
```
For these two rows (1, NULL) and (NULL, 1), the third column operation should be NULL, because
```sql
memory D NULL = NULL;
┌───────────────┐
│ (NULL = NULL) │
│    boolean    │
├───────────────┤
│ NULL          │
└───────────────┘
memory D NULL OR false;
┌─────────────────┐
│ (NULL OR false) │
│     boolean     │
├─────────────────┤
│ NULL            │
└─────────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

wufengwind:
## Summary

EqualOrNullSimplification rewrites `a = b OR (a IS NULL AND b IS NULL)` to `a IS NOT DISTINCT FROM b`. That is only semantics-preserving when the expression is used as a predicate, where `false` and `NULL` both reject rows.

In projections, the rewrite changes SQL three-valued logic by turning `NULL` into `false`.

This restricts the rewrite to root filter/join predicates and keeps projection expressions unchanged.

Fixes #23685.

## Tests

- `build/reldebug/test/unittest test/sql/optimizer/expression/test_equal_or_null_optimization.test`

### DIFF
--- a/src/optimizer/rule/equal_or_null_simplification.cpp
+++ b/src/optimizer/rule/equal_or_null_simplification.cpp
@@ -7,6 +7,22 @@
 
 namespace duckdb {
 
+static bool IsPredicateRoot(LogicalOperator &op, bool is_root) {
+	if (!is_root) {
+		return false;
+	}
+	switch (op.type) {
+	case LogicalOperatorType::LOGICAL_FILTER:
+	case LogicalOperatorType::LOGICAL_COMPARISON_JOIN:
+	case LogicalOperatorType::LOGICAL_ANY_JOIN:
+	case LogicalOperatorType::LOGICAL_DELIM_JOIN:
+	case LogicalOperatorType::LOGICAL_ASOF_JOIN:
+		return true;
+	default:
+		return false;
+	}
+}
+
 EqualOrNullSimplification::EqualOrNullSimplification(ExpressionRewriter &rewriter) : Rule(rewriter) {
 	// match on OR conjunction
 	auto op = make_uniq<ConjunctionExpressionMatcher>();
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks for the PR! I think we can only enable this for `LOGICAL_FILTER` and `LOGICAL_ANY_JOIN`, this produces incorrect results with your changes now:
```sql
# name: test/temp_equal_or_null_review.test
# description: Equal-or-null rewrite must preserve NULL inside join comparison operands
# group: [optimizer]

query I
SELECT count(*)
FROM (VALUES (0::INTEGER, NULL::INTEGER)) l(a, b)
JOIN (VALUES (false)) r(c)
  ON (l.a = l.b OR (l.a IS NULL AND l.b IS NULL)) = r.c;
----
0
```

What do you think?

wufengwind:
Thanks, you are right. `is_root` only identifies the root of the expression being rewritten; for comparison joins that expression can still be an operand of a larger predicate. I restricted the rewrite to `LOGICAL_FILTER` and `LOGICAL_ANY_JOIN`, and added your case as a regression test.

PR #23950, opened 2026-07-20
storage: validate metadata block indexes

wufengwind:
## Summary
- validate metadata sub-block indexes before constructing metadata reader pointers
- reject out-of-range legacy block pointer offsets when converting to `MetaBlockPointer`
- add a regression test for a checksum-valid database header with an invalid metadata block index

## Tests
- `build/reldebug/test/unittest "Test opening a database with invalid metadata block index"`
- `build/reldebug/test/unittest "[api]"`


Mytherin:
Thanks!

PR #23952, opened 2026-07-20
Fix SEMI/ANTI join mishandling duplicate equality keys with residual predicate

### LINKED ISSUE #23932: SEMI/ANTI JOIN mishandles duplicate equality keys with a residual predicate
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE l(k INTEGER, v INTEGER);
memory D CREATE TABLE r(k INTEGER, v INTEGER);
memory D INSERT INTO l VALUES
             (1, 1),
             (-1, 5),
             (2, 1);
memory D INSERT INTO r VALUES
             (1, 5),
             (-1, 0),
             (1, 1);

-- should be (2, 1)
memory D SELECT *
         FROM l ANTI JOIN r
           ON l.k = r.k
          AND l.v + r.v > 2
         ORDER BY k;
┌───────┬───────┐
│   k   │   v   │
│ int32 │ int32 │
├───────┼───────┤
│     1 │     1 │
│     2 │     1 │
└───────┴───────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

EslamAhmed171:
Fixes: #23932

### DIFF
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -1534,7 +1534,7 @@ idx_t ScanStructure::ApplyResidualPredicate(DataChunk &probe_data, SelectionVect
 	idx_t new_match_count = residual_executor->SelectExpression(residual_state->eval_chunk, selected_sel, remaining_sel,
 	                                                            nullptr, match_count);
 
-	std::vector<idx_t> original_indices(match_count);
+vector<idx_t> original_indices(match_count);
 	for (idx_t i = 0; i < match_count; i++) {
 		original_indices[i] = match_sel.get_index(i);
 	}
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -1534,16 +1534,21 @@ idx_t ScanStructure::ApplyResidualPredicate(DataChunk &probe_data, SelectionVect
 	idx_t new_match_count = residual_executor->SelectExpression(residual_state->eval_chunk, selected_sel, remaining_sel,
 	                                                            nullptr, match_count);
 
+	std::vector<idx_t> original_indices(match_count);
+	for (idx_t i = 0; i < match_count; i++) {
+		original_indices[i] = match_sel.get_index(i);
+	}
+
 	for (idx_t i = 0; i < new_match_count; i++) {
 		idx_t dense_idx = selected_sel.get_index(i);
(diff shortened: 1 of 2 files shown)


src/execution/join_hashtable.cpp
```diff
@@ -1534,16 +1534,21 @@ idx_t ScanStructure::ApplyResidualPredicate(DataChunk &probe_data, SelectionVect
 	idx_t new_match_count = residual_executor->SelectExpression(residual_state->eval_chunk, selected_sel, remaining_sel,
 	                                                            nullptr, match_count);
 
+	std::vector<idx_t> original_indices(match_count);
```
dentiny:
```suggestion
vector<idx_t> original_indices(match_count);
```
which provides out-of-bound access protection

EslamAhmed171: Nice catch!

src/execution/join_hashtable.cpp
```diff
@@ -1534,16 +1534,21 @@ idx_t ScanStructure::ApplyResidualPredicate(DataChunk &probe_data, SelectionVect
 	idx_t new_match_count = residual_executor->SelectExpression(residual_state->eval_chunk, selected_sel, remaining_sel,
 	                                                            nullptr, match_count);
 
+	vector<idx_t> original_indices(match_count);
+	for (idx_t i = 0; i < match_count; i++) {
+		original_indices[i] = match_sel.get_index(i);
+	}
```
lnkuiper: Could we avoid materializing `original_indices` here? This adds a heap allocation and full copy to every residual-predicate probe, including calls where no_match_sel is null. All current callers use distinct match/no-match selection vectors, so we could populate `no_match_sel` before compacting `match_sel` in place instead? If a copy is necessary, a reusable `SelectionVector` in `ResidualPredicateProbeState` would be better

lnkuiper (changes requested):
Thanks for the PR! I've left a comment below:

src/execution/join_hashtable.cpp
EslamAhmed171: Yes we could populate `no_match_sel` first!

lnkuiper (approved):
Thanks!

PR #23964, opened 2026-07-20
Shrink the in-memory footprint of deleted rows

ywelsch:
## Problem

Deleting larger amounts of data can quickly hit the memory limits in DuckDB. Further, for observability use cases where DuckDB is continuously ingesting data, and most checkpoints are concurrent (i.e. don't allow vacuum), there can be some time before the memory of deletes can be reclaimed.

Deleting rows is disproportionately memory-hungry relative to the information a delete actually carries. As soon as a single row in a vector is deleted, `ChunkVectorInfo` materializes a full `STANDARD_VECTOR_SIZE × sizeof(transaction_t)` array of per-row delete commit ids — **16 KiB per vector** at default settings, i.e. ~8 bytes for every deleted row. That array is held until the row group is rewritten at a full checkpoint, and **indefinitely for in-memory databases**.

For a bulk delete (or deletion of a lot of "old data") this is almost entirely waste. After `DELETE FROM t` on a 100M-row table, ~815 MB of version-info memory is retained even though every row shares the same delete id. Under a tight memory limit the delete can fail outright:

```sql
SET memory_limit='300MB';
CREATE TABLE t AS SELECT range AS i FROM range(100_000_000);
DELETE FROM t;
-- Out of Memory Error: could not allocate block of size 256.0 KiB (287.2 MiB/286.1 MiB used)
```

## Approach

The insert side already had the pattern: a `constant_insert_id` collapses the per-row array to a single value when all rows share it. This PR mirrors that on the delete side and reclaims the freed memory, in a few independent steps:

1. **`constant_delete_id` (commit-time collapse).** When a delete covers an entire vector — the `DELETE FROM t` shape — the id is stored as a single constant and the per-row array is never allocated (`Delete()`) or freed on commit (`CommitDelete()`). The array is lazily re-materialized from the constant only when a *partial* modification needs per-row ids, which preserves write-write conflict detection and snapshot visibility exactly. Rollback / revert-commit route through the same paths.

2. **Checkpoint-time compression (piecemeal deletes).** Vectors deleted fully across *multiple* transactions carry distinct per-row commit ids, so they can't collapse on commit. Once every id is below the lowest active transaction start, the ids are indistinguishable from a single constant for all current and future snapshots. The checkpoint is the lifetime-safe place to do this collapse (it operates on the live `RowVersionManager` owned by the row group, unlike transaction cleanup which would dangle the pointer stored in `DeleteInfo`), and it runs for `VACUUM_ONLY` checkpoints too, so in-memory databases reclaim on `CHECKPOINT`. Fully-visible non-constant *insert* ids are collapsed in the same pass.

3. **Release empty buffers.** `FixedSizeAllocator::Free` intentionally keeps one empty buffer alive to avoid churn (an ART heuristic). That doesn't apply to the one-shot compression pass, so a fully-compressed row group would otherwise retain one 256 KiB buffer. `CompressVersionIds` now calls `RemoveEmptyBuffers()` at the end.

4. **Cleanup: remove `ChunkConstantInfo` and flatten the hierarchy.** With the above, `ChunkConstantInfo` was just a `ChunkVectorInfo` with both ids constant — a state now represented natively. Removing it also removes the constant→vector conversion in `RowVersionManager::GetVectorInfo` (where delete ids could previously be silently dropped). `ChunkVectorInfo` then became the only `ChunkInfo` subclass, so the base class is merged in: the class is non-virtual (per-vector-scan `GetSelVector` is no longer a virtual call), the in-memory `type` field and `Cast<>` machinery are gone.

**The on-disk format is unchanged.** `ChunkInfoType` (`EMPTY_INFO` / `CONSTANT_INFO` / `VECTOR_INFO`) survives purely as a serialization tag; `ChunkVectorInfo::Write` already emitted `CONSTANT_INFO` for fully-deleted vectors and `Read` reconstructs from the same tags.

## Measurements

100M-row table, `DELETE FROM t`, version-info memory (`BASE_TABLE`) retained after the delete:

| | before | after |
|---|---|---|
| memory retained | **815 MB** | **1.6 MB** (≈1.3 MB pre-existing data bookkeeping + one 256 KiB tail-vector array) |
| delete wall time | 2.31 s | 2.19 s |
| under `memory_limit='300MB'`, no spilling | **OOM mid-delete** | completes, 0 rows |

The peak is eliminated, not just the steady state, because the whole-vector path never allocates the array in the first place. Piecemeal multi-transaction deletes materialize arrays transiently and drop to **0 bytes** after the next full checkpoint (including in-memory `CHECKPOINT`).

### DIFF
--- a/src/storage/table/chunk_info.cpp
+++ b/src/storage/table/chunk_info.cpp
@@ -273,6 +273,8 @@ idx_t ChunkVectorInfo::Delete(transaction_t transaction_id, row_t rows[], idx_t 
 		constant_delete_id = transaction_id;
 		return count;
 	}
+	// we are materializing / modifying per-row delete ids - re-arm the compression check
+	recheck_compression = true;
 	auto segment = allocator.GetHandle(GetInitializedDeletedPointer());
 	auto deleted = segment.GetPtr<transaction_t>();
 
@@ -310,6 +312,8 @@ void ChunkVectorInfo::CommitDelete(transaction_t commit_id, const DeleteInfo &in
 		// all rows already share this exact deleted id - nothing to do
 		return;
 	}
+	// we are materializing / modifying per-row delete ids - re-arm the compression check
+	recheck_compression = true;
 	bool all_equal = true;
 	{
 		auto segment = allocator.GetHandle(GetInitializedDeletedPointer());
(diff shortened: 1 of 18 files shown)


src/storage/table/row_group_collection.cpp
```diff
@@ -1733,6 +1736,8 @@ void RowGroupCollection::Checkpoint(TableDataWriter &writer, TableStatistics &gl
 			if (!RefersToSameObject(row_group.GetCollection(), *this)) {
 				throw InternalException("RowGroup Vacuum - row group collection of row group changed");
 			}
+			// the row group is kept as-is: try to compress its version information
+			row_group.CompressVersionInfo(lowest_active_start);
```
Mytherin, line 1740: I wonder if doing this in the `Checkpoint` is the right place for it - given we will now be doing this relatively frequently and `CompressVersionIds` is not that cheap if there are deletes.

Maybe this can be done in transaction clean-up or commit instead? Instead of running this for all chunk info's all the time we can then run this more targeted for deletes that have just been marked as deleted. 

Alternatively we can maybe keep a `bool check_compression` or so that verifies if there's any point to checking (i.e. any changes have been made since the last time we checked).

src/storage/table/chunk_info.cpp
```diff
@@ -554,20 +591,36 @@ void ChunkVectorInfo::Write(WriteStream &writer, transaction_t checkpoint_id) co
...
+		auto start = reader.Read<idx_t>();
+		auto result = make_uniq<ChunkVectorInfo>(allocator, start);
+		ValidityMask mask;
+		mask.Read(reader, STANDARD_VECTOR_SIZE);
+
+		auto segment = allocator.GetHandle(result->GetInitializedDeletedPointer());
+		auto deleted = segment.GetPtr<transaction_t>();
+		for (idx_t i = 0; i < STANDARD_VECTOR_SIZE; i++) {
+			if (mask.RowIsValid(i)) {
+				deleted[i] = 0;
```
Mytherin, line 679: This is more of a potential future work: we're still using 8 bytes per row if we're not deleting all rows in a vector. This is actually potentially even worse because if there's only one deleted row per vector we will still allocate the 16KB of data for that vector. We could improve this by only using 1 bit per row (256 bytes per vector) in the case where the transaction ids are not necessary by keeping the ValidityMask. This of course then introduces some additional complexity w.r.t. memory management.

Mytherin (commented):
Thanks for the PR! Looks good - nice idea, and I've been meaning to get rid of the constant info / vector info split. Some comments but otherwise I think this is good to go:

src/storage/table/row_group_collection.cpp
ywelsch: Good point, I went with the check_compression suggestion in https://github.com/duckdb/duckdb/pull/23964/commits/341ae92fa97e005cbbe9cb30eed11ec6a337841e

src/storage/table/chunk_info.cpp
ywelsch: Agreed, my focus in this PR here was on bulk deletes (e.g. old data in a table where the deleted data is colocated in the same rowgroup) while receiving a continuous stream of new data (i.e. full checkpoints were rare), i.e. a typical observability use case.

I've left sparse deletes, or a more compact representation for historical deletes, out of scope for this PR.

Mytherin:
Thanks!

baolinhuang:
Hi @ywelsch & @Mytherin 

We (the Alibaba Cloud AliSQL team) hit the same problem in production. In our case the deletes are
discrete — individual rows deleted scattered across many vectors — rather than whole vectors being
fully deleted. As a result, each partially-deleted vector still allocates a full 16 KiB per-row
delete-id array even though only a small fraction of its rows are actually deleted, and this memory
 piles up across row groups.

Building on this patch, I implemented the further optimization discussed above: once a partially-deleted vector's delete ids are all visible to every transaction, its per-row delete[] array is folded into a compact delete 
mask (a ValidityMask, ~256 bytes/vector, 1 bit per row) instead of retaining the full transaction_t
 array.
The PR is: https://github.com/duckdb/duckdb/pull/24336

PR #23968, opened 2026-07-20
Bump Postgres extension on main

staticlibs:
This PR updates Postgres extension to make it easier to consume Redshift-related changes added in:

 - duckdb/duckdb-postgres#520

 Ref: duckdblabs/duckdb-internal#9970


Mytherin:
Thanks!

PR #23972, opened 2026-07-20
Use relative includes in unity build sources for cache reuse

NiclasHaderer:
The unity build generator writes each source into ub_*.cpp as an absolute include (`#include <${CMAKE_CURRENT_SOURCE_DIR}/...>`), baking the checkout path into the compiled translation unit. Compiler caches (ccache/sccache) key on source content, so the same code built from different locations (git worktrees, a moved clone, CI vs local paths) never shares objects.

Emit the include relative to the unity file's directory instead, so the generated content is identical regardless of checkout location.


Mytherin:
Thanks!

PR #23974, opened 2026-07-20
Reduce test time (2+ sec) for some fast tests

smvv:
I used `build/release/test/run --track-runtime=2` locally to find fast tests that take more than 2 sec to run:
- `test_sleep.test`: slow because intentional sleeps add up to about 2s, including `pg_sleep(1.0)`.
- `test_now_prepared.test`: slow because it has two `sleep 1 second` commands.
- `rollback_drop_column.test`: slow because it inserts/checkpoints `50k` rows with `~6.4KB` strings before the actual regression.
- `create_external_resource.test`: slow because it has wall-clock timeout checks: `max_execution_time = 500` plus `timeout_seconds := 1`.
- `test_stress_update_issue_19688.test`: heavy because it runs `8 * 1000 * 3` concurrent DML statements; keep coverage but reduce fast-suite scale.

These fast tests run in most CI jobs (like 40x during `Tests a release build with different configurations`) and adding slow running tests slows down the overall CI pipelines.

### DIFF
--- a/test/sql/function/generic/test_sleep.test
+++ b/test/sql/function/generic/test_sleep.test
@@ -4,7 +4,7 @@
 
 # Test basic sleep function - should return NULL
 query T
-SELECT sleep_ms(10)
+SELECT sleep_ms(1)
 ----
 NULL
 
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks!

PR #23975, opened 2026-07-20
Push down projections into shredded Parquet VARIANT columns

redox:
This avoids reading unrelated Parquet column chunks when a query accesses specific object fields from a shredded `VARIANT`.

Previously, the Parquet reader loaded the entire shredded `VARIANT`, including unrelated sibling fields. This change translates the logical `VARIANT` field paths recorded by the optimizer into physical projections of the Parquet `typed_value` tree.

For each selected object field, the reader retains:

- The field's `value` child, which contains partially shredded or type-mismatched values.
- The required portion of its `typed_value` subtree.

The root `metadata` column is always read, while the root `value` column and unrelated shredded siblings can be omitted.

Multiple paths with shared prefixes are supported.

If a path cannot be resolved safely (including array paths or fields outside the shredded schema) the reader conservatively falls back to reading the complete VARIANT.

On some of our use-cases we're speaking 10-100x less bytes read :)

### DIFF
--- a/extension/parquet/parquet_reader.cpp
+++ b/extension/parquet/parquet_reader.cpp
@@ -592,6 +592,92 @@ ParquetColumnSchema ParquetReader::ParseColumnSchema(const SchemaElement &s_ele,
 	                                              parquet_options);
 }
 
+namespace {
+
+optional_idx FindSchemaChild(const ParquetColumnSchema &schema, const string &name) {
+	for (idx_t child_idx = 0; child_idx < schema.children.size(); child_idx++) {
+		if (schema.children[child_idx].name == name) {
+			return child_idx;
+		}
+	}
+	return optional_idx();
+}
+
+bool TryResolveVariantProjection(const ParquetColumnSchema &typed_value_schema, const ColumnIndex &logical_index,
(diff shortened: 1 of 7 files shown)


redox:
Curious what the VARIANT experts @Mytherin and @Tishj think about this; on our IO(httpfs)-heavy workload it's night and day 🚀

Mytherin:
Thanks for the PR! We already have a WIP PR here - https://github.com/duckdb/duckdb/pull/22478 - that implements this as well (likely in a more battle tested manner). Perhaps you can verify that that PR also has the same effect on your use case.

redox:
> Thanks for the PR! We already have a WIP PR here - #22478 - that implements this as well (likely in a more battle tested manner). Perhaps you can verify that that PR also has the same effect on your use case.

Ah nice, I'll compare implementation and report back.

redox:
Closing as https://github.com/duckdb/duckdb/pull/22478 goes way further 🚀; not easy to test on our side as we're on the 1.5 branch; but will subscribe to the PR and test when I can.

Great job @Tishj ; thanks for the swift reply @Mytherin

PR #23990, opened 2026-07-21
Optimize row group pruning for substr

### LINKED ISSUE #23989: `substr` function doesn't prune row group
### What happens?

See repro SQL.

### To Reproduce

```sql
ATTACH '/tmp/rg_substr.duckdb' AS d (ROW_GROUP_SIZE 2048);
CREATE TABLE d.t AS
SELECT i::INTEGER i, ('p' || lpad(i::VARCHAR, 6, '0')) s
FROM range(8192) r(i);
CHECKPOINT d;

EXPLAIN ANALYZE SELECT sum(i) FROM d.t WHERE s = 'p005000';
-- Row Groups Scanned: 1 / 4

EXPLAIN ANALYZE SELECT sum(i) FROM d.t WHERE substr(s, 2) = '005000';
-- Row Groups Scanned: 4 / 4
```
The two queries are essentially the same because they're of the same prefix "p".

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/23989

The idea is, if (1) min and max share the same prefix, AND (2) the common prefix is not disjoint with requested substring, we could prune row group.

Copy the issue repro SQL here for easier illustration:
```sql
CREATE TABLE d.t AS
SELECT i::INTEGER i, ('p' || lpad(i::VARCHAR, 6, '0')) s
FROM range(8192) r(i);
```
Because both min and max in the row group share the same prefix "p", which is of length 1, it has nothing with a prefix scan, so substr(x, 2) could leverage the zonemap to prune.


src/function/scalar/string/substring.cpp
```diff
@@ -302,6 +302,73 @@ void SubstringFunctionASCII(DataChunk &args, ExpressionState &state, Vector &res
...
 
+bool GetPrefixByteLength(const string &value, idx_t character_count, idx_t &byte_count) {
+	byte_count = 0;
+	for (idx_t character_idx = 0; character_idx < character_count; character_idx++) {
+		if (byte_count >= value.size()) {
+			return false;
+		}
+		auto first_byte = static_cast<uint8_t>(value[byte_count]);
+		idx_t codepoint_size = 0;
+		if (first_byte <= 0x7F) {
```
Mytherin: Can we use existing UTF8 functions here?

Mytherin (commented):
Thanks - looks good to me.

Could we also add propagation of min/max when `offset = 0`? We're currently bailing on min/max propagation there entirely, but I think we can always directly propagate the min/max values in that case.

i.e. if we have:

```
min: ABC
max: ZYX
```

And we get `str[:2]`, the new min/max could be determined to be:

```
min: AB
max: ZY
```

If the min/max strings are shorter than the prefix length then the min/max can just be preserved I think, for example if we have:

```
min: A
max: Z
```

And we run `str[:2]` - then the resulting min/max stats will be the exact same as the input.

src/function/scalar/string/substring.cpp
```diff
@@ -302,6 +302,73 @@ void SubstringFunctionASCII(DataChunk &args, ExpressionState &state, Vector &res
...
+	auto &expr = input.expr;
+	auto &children = expr.GetChildren();
+	if (children.size() != 2 || children[1]->GetExpressionClass() != ExpressionClass::BOUND_CONSTANT) {
+		return nullptr;
+	}
+	auto &offset_value = children[1]->Cast<BoundConstantExpression>().GetValue();
+	if (offset_value.IsNull()) {
+		return nullptr;
+	}
+	auto offset = offset_value.GetValue<int64_t>();
```
dentiny: This prevents offset = 0 optimization

dentiny:
> Could we also add propagation of min/max when offset = 0? 

Add offset=0 handle

src/function/scalar/string/substring.cpp
```diff
@@ -302,6 +302,118 @@ void SubstringFunctionASCII(DataChunk &args, ExpressionState &state, Vector &res
...
+	    children[1]->GetExpressionClass() != ExpressionClass::BOUND_CONSTANT) {
+		return nullptr;
+	}
+	auto &offset_value = children[1]->Cast<BoundConstantExpression>().GetValue();
+	if (offset_value.IsNull()) {
+		return nullptr;
+	}
+	auto offset = offset_value.GetValue<int64_t>();
+	if (offset == 0 || offset == 1) {
+		return SubstringStatsFromPrefix(input, offset);
```
Mytherin: Could you expand a bit on the offset logic here? Shouldn't this be `offset == 1` instead of `offset == 0 || offset == 1`? What does an offset of 0 mean here, or is this generated to mean that the two are identical?

src/function/scalar/string/substring.cpp
```diff
@@ -302,6 +302,118 @@ void SubstringFunctionASCII(DataChunk &args, ExpressionState &state, Vector &res
...
+		return nullptr;
+	}
+	auto length = length_value.GetValue<int64_t>();
+	if (length <= 0 || length > NumericLimits<uint32_t>::Maximum()) {
+		return nullptr;
+	}
+	if (offset == 0) {
+		length--;
+	} else if (offset != 1) {
+		return nullptr;
```
Mytherin: It feels that this is kind of double checking the offset again

src/function/scalar/string/substring.cpp
Mytherin, line 390: Could you maybe investigate why this seems to work for `substr` but not for slicing, which should follow the same operator?

```sql
memory D select stats(substr(s, 2)) from strings;
┌───────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                        stats(substr(s, 2))                                        │
│                                              variant                                              │
├───────────────────────────────────────────────────────────────────────────────────────────────────┤
│ {'has_no_null': true, 'has_null': false, 'has_unicode': true, 'max': bcdefgh99, 'min': bcdefgh00} │
│ {'has_no_null': true, 'has_null': false, 'has_unicode': true, 'max': bcdefgh99, 'min': bcdefgh00} │
└───────────────────────────────────────────────────────────────────────────────────────────────────┘
memory D select stats(s[2:]) from strings;
┌──────────────────────────────────────────────────────────────┐
│                         stats(s[2:])                         │
│                           variant                            │
├──────────────────────────────────────────────────────────────┤
│ {'has_no_null': true, 'has_null': true, 'has_unicode': true} │
│ {'has_no_null': true, 'has_null': true, 'has_unicode': true} │
└──────────────────────────────────────────────────────────────┘

```

src/function/scalar/string/substring.cpp
```diff
@@ -302,6 +302,118 @@ void SubstringFunctionASCII(DataChunk &args, ExpressionState &state, Vector &res
...
+	}
+	auto &offset_value = children[1]->Cast<BoundConstantExpression>().GetValue();
+	if (offset_value.IsNull()) {
+		return nullptr;
+	}
+	auto offset = offset_value.GetValue<int64_t>();
+	if (offset == 0 || offset == 1) {
+		return SubstringStatsFromPrefix(input, offset);
+	}
+	if (offset < 0 || children.size() != 2) {
```
Mytherin: Instead of bailing when a length is provided maybe we could also handle the case where the length is a constant? e.g. if we have the following stats + slice with `substr(str, 2, 4)`


```
min: abcdefgh00
max: abcdefgh99
```

We could detect the common prefix `bcdefgh` and slice that with `offset = 2, length = 4` to get to a `min/max` of `bcde`.

Mytherin (commented):
Thanks for the fixes! Looks good - some more comments

src/function/scalar/string/substring.cpp
dentiny: In the latest commit, I was trying to special handle offset=0 and offset=1 case, because they could be directly processed to prefix scan.

src/function/scalar/string/substring.cpp
dentiny: I think this PR is focusing on `substring` function, while suscriptor `[]` corresponds to `array_slice` function I believe.

Mytherin: Maybe those could be unified in some manner? Or at least the stats propagation aspects could be?

src/function/scalar/string/substring.cpp
dentiny: In the latest commit, I normalize the offset/length so we could handle them in one single function; hopefully it's easier to read.

src/function/scalar/string/substring.cpp
dentiny: Yeah, I think it should be addressed :)

src/function/scalar/string/substring.cpp
dentiny: Yeah I extract out a util function to get substr stats.

Mytherin:
Thanks!

PR #24002, opened 2026-07-21
Parse constant JSON paths once at bind time

mustafahasankhan:
Follows up on the proposal in #23959.

Constant JSON paths are currently re-tokenized for every row. The bind data already detects constant paths, validates them, and stores the normalized string, but execution re-parses that string per row, allocating a scratch buffer and a `string` per key segment each time. Bare keys and arrow paths are normalized into the quoted form (`$."key"`), which takes the allocating parser path, so the most common extract patterns pay this on every row.

Profiling `SELECT count(j->'user')` over 5 million small documents put 31% of the execution time inside path handling, most of it in `operator new` and `free` called from `ReadKey`, while the actual key comparison was around 1%. After this change the path share drops to 6.5% with no allocator frames left; document parsing is the dominant cost, as it should be.

Changes:
- `JSONCommon` gains `ParsePathElements` and `GetPathElements`: parse a validated path into `JSONPathElement`s once, walk them with no tokenization and no allocations
- `JSONReadFunctionData` and `JSONReadManyFunctionData` derive the parsed elements from the stored path string in their constructors, so `Copy()` and rebinds stay self contained
- The modification functions (`json_set`, `json_insert`, `json_replace`, `json_remove`) get bind data (`JSONModifyFunctionData`) with the same treatment, covering JSONPath, bare key, and empty paths
- Wildcard paths, JSON Pointer paths, and non constant paths are unchanged
- New micro benchmarks under `benchmark/micro/json/`, prepared statement tests, and a consistency test comparing VARCHAR and JSON typed inputs across adversarial documents

Benchmarks (single threaded to keep variance under 4%, medians of 5 on an 11 core Apple machine; multithreaded speedups are similar):

| benchmark | main | this PR | speedup |
|---|---|---|---|
| `json_extract_bare_key` | 451.6 ms | 342.0 ms | 1.32x |
| `json_extract_deep_path` | 443.4 ms | 350.5 ms | 1.27x |
| `json_extract_string_path` | 393.7 ms | 309.4 ms | 1.27x |
| `json_extract_many_paths` | 683.4 ms | 553.5 ms | 1.23x |
| `json_set_path` | 1154.1 ms | 868.5 ms | 1.33x |
| `json_extract_large_doc` | 437.2 ms | 410.5 ms | 1.06x |
| `json_extract_wildcard` | 590.9 ms | 589.5 ms | 1.00x |
| `json_transform_struct` | 486.5 ms | 471.0 ms | 1.03x |

Peak memory is unchanged (max RSS 521.7 MB vs 523.6 MB on the benchmark workload plus a 200 expression query): the bind data adds a few hundred bytes per expression, and execution allocates strictly less than before.

Two approaches I tried and rejected, in case they come up in review:
- Locating constant path targets by scanning the raw JSON text and parsing only the target span. Correct, and 1.2x on string valued targets, but character level skipping is not enough cheaper than yyjson parsing: container valued targets regressed up to 32%. Behavior that depends on data shape did not seem worth it.
- Copying the input into the allocator arena and parsing with `YYJSON_READ_INSITU`: the extra copy cancels the gain.

### DIFF
--- a/benchmark/micro/json/json_extract_bare_key.benchmark
+++ b/benchmark/micro/json/json_extract_bare_key.benchmark
@@ -1,0 +1,19 @@
+# name: benchmark/micro/json/json_extract_bare_key.benchmark
+# description: Extract a top-level key from small JSON documents with the arrow operator
+# group: [json]
+
+name JSON extract bare key
+group json
+
+require json
+
+load
+CREATE TABLE docs AS
+SELECT json_object('id', i, 'user', json_object('name', 'user_' || i, 'tags', json_array(i % 10, i % 7)), 'ts', i * 1000) AS j
+FROM range(0, 5000000) t(i);
+
+run
+SELECT count(j->'user') FROM docs
+
+result I
+5000000
--- a/benchmark/micro/json/json_extract_deep_path.benchmark
+++ b/benchmark/micro/json/json_extract_deep_path.benchmark
@@ -1,0 +1,19 @@
+# name: benchmark/micro/json/json_extract_deep_path.benchmark
+# description: Extract a nested path from small JSON documents
+# group: [json]
+
+name JSON extract deep path
+group json
+
+require json
+
+load
+CREATE TABLE docs AS
+SELECT json_object('id', i, 'user', json_object('name', 'user_' || i, 'tags', json_array(i % 10, i % 7)), 'ts', i * 1000) AS j
+FROM range(0, 5000000) t(i);
+
+run
+SELECT count(j->'$.user.tags[0]') FROM docs
+
+result I
+5000000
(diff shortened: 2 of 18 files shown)


lnkuiper:
Thanks for the PR! Good idea. One comment then this is ready to go:
```sql
require json

# Constant modification path errors remain catchable by TRY.
query T
SELECT TRY(json_set('{}', '$not_a_path', '1'));
----
NULL

# Unreachable modification calls must not fail before execution either.
query T
SELECT CASE WHEN false THEN json_set('{}', '$not_a_path', '1') ELSE '{}'::JSON END;
----
{}
```
I think we can preserve execution-time validation by catching the exception in the binder.

mustafahasankhan:
Good catch, fixed. The bind data now catches the parse error and falls back to per-row parsing, so invalid constant paths error only when a row is actually evaluated: `TRY` returns NULL and unreachable branches stay silent. Both your repros are in the tests, plus `TRY` over an invalid reverse index and over a wildcard path. The prepared statement test still confirms the error does surface on real execution.

lnkuiper:
Thanks! LGTM now

PR #24009, opened 2026-07-21
StorageLock: wait on a condition variable instead of busy-spinning

ywelsch:
GetExclusiveLock currently takes the mutex and then spins in a tight loop until all shared holders have drained, burning a core for the entire wait - possibly a long time while waiting for a query to complete.

This became visible while performance-testing https://github.com/duckdb/duckdb/pull/23631

As it's (hopefully) an uncontroversial change of its own, I factored this into a separate small PR here.

It replaces the current implementation with a phase-fair read-write lock built on a condition variable:

- A registering writer drains the readers ahead of it: no new readers are admitted past it, so a steady stream of readers cannot starve an exclusive acquirer (e.g. a checkpoint). This matches the current behavior, where a waiting writer implicitly holds off new readers by sitting on the mutex they need.
- A reader only waits for writers registered before it (writer tickets), not for writers that arrive while it is waiting, so a steady stream of writers cannot starve readers either. This also matches the current behavior, where waiting readers and later writers queue on the same mutex.
- TryGetExclusiveLock and TryUpgradeLock keep their non-blocking semantics and success conditions; lock upgrade still grants the exclusive lock alongside the held shared lock (read_count stays at 1).

The only intended behavioral difference is that waiting no longer consumes CPU.

### DIFF
--- a/src/storage/storage_lock.cpp
+++ b/src/storage/storage_lock.cpp
@@ -5,42 +5,52 @@
 #include "duckdb/common/assert.hpp"
 #include "duckdb/common/thread_annotation/thread_annotation.hpp"
 
+#include <condition_variable>
+
 namespace duckdb {
 
+// A phase-fair read-write lock: waiters sleep on a condition variable instead of busy-spinning.
+// A registering writer drains the readers ahead of it (no new readers are admitted past it), but a reader only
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks! LGTM

PR #24014, opened 2026-07-21
Perf regression: ValidateDictionaryIndex in hot loop and shared_ptr access

gropaul:
Hi dear duckdb team,

This PR avoids `shared_ptr` access when shifting the dict offsets by adding Selection, which led to a 20% performance improvement for the scan part.

While preparing the PR, I came across https://github.com/duckdb/duckdb/pull/23630, which introduced checking each dictionary index for range. The problem was that this was happening in a hot loop over a shared_ptr. This led to a 4x performance regression for the dictionary scan part, which made checking the indices 80% of the full scan time. 

The PR fixes this by hoisting the throw out of the loop.

https://github.com/duckdb/duckdb/pull/23630 also introduces length checks for the dictionary compressed values in a similar fashion. This could lead to regressions for dictionaries with many strings, but I did not do detailed investigations here. 

Also, with this change, checking the dictionary offset takes 25% of the dictionary scan time, so it still comes with a cost. 

image

https://github.com/duckdb/duckdb/pull/24006 could introduce the same problem potentially. 

With this PR, e2e time from ` SELECT DISTINCT l_shipinstruct FROM lineitem ORDER BY l_shipinstruct;` goes from 160ms to 70ms for TPC-H sf10 on a M4 Macbook Pro. 90% of the perf improvement comes from fixing the throw.


src/storage/compression/dictionary/decompression.cpp
```diff
@@ -171,14 +185,8 @@ void CompressedStringScanState::ScanToDictionaryVector(ColumnSegment &segment, V
...
-	if (start_offset != 0) {
-		for (idx_t i = 0; i < scan_count; i++) {
-			sel_vec->set_index(i, sel_vec->get_index(i + start_offset));
-		}
-	}
-	for (idx_t i = 0; i < scan_count; i++) {
-		ValidateDictionaryIndex(sel_vec->get_index(i));
-	}
+	sel_vec->ShiftLeft(start_offset, scan_count);
+	ValidateDictionary(*sel_vec, scan_count);
```
Mytherin, line 204: I wonder if we can merge the validate with the `ShiftLeft` code, since we're already doing a loop / loading the dictionary indexes that might reduce the perf impact more? e.g. `ShiftLeft` can take the max index, and then keep track of the error count, and throw after the loop is over with a corruption exception if the error count is > 0.

Mytherin: Actually we're almost never shifting - so that doesn't really help I think.

Another idea is that the bitpacking width (`current_width`) dictates a clear maximum of what the value can be, e.g. if our bitpacking width is `8` then the maximum value in the dictionary is `2^8` (maybe `2^8-1`?).  If that maximum value is `<= index_buffer_count` we know immediately that all indexes are in range without having to loop over them again.

I wonder if that might help avoid the per-value check in many cases.

Mytherin: Another thought: instead of allocating `index_buffer_count` values in the dictionary, we can also allocate `NextPowerOfTwo(index_buffer_count)` if `index_buffer_count` is small enough. For example, if `index_buffer_count = 5`, we can just allocate a dictionary vector of size 8. We can then set the trailing values to `NULL` in that vector. That together with the bitpacking width should then mean we rarely need to check entries while still keeping the desired safety (i.e. if the indexes are out of range they will just be read as `NULL` values instead of reading out of bounds of the vector).

gropaul: Hi @Mytherin, thanks for the ideas. This would imply that if there is data corruption in the dictionary segment instead of a crash, we would return wrong data and fail silently. This is also the behavior now, as in the current state, the corruption would create indices that are still within the size of the dictionary buffer; it will also output wrong results. But with https://github.com/duckdb/duckdb/pull/23630 it does not crash anymore.

I am not sure about DuckDB's general idea on corrupted segments. Is this something you want to protect against, and how much? I am a bit unsure about whether the current process of adding them on an encoding level is the right way forward, as this does not catch all the corruptions, is an overhead in decoding, and adds a lot of complexity per encoding. There is already the `uint64_t` checksumming, which seems to be a strong protection. 

If you want this protection, I think your approach is a good idea, and I can implement it.

Mytherin: My thoughts on this are that it should not be possible to trigger undefined behavior (e.g. out-of-bounds reads) through reading a (potentially maliciously crafted) DuckDB database file. Ideally these checks would have as little effect as possible on run-time, and returning wrong results instead of throwing an explicit corruption exception is fine, as long as the undefined behavior is not triggered.

gropaul: Cool thanks. I will update this PR then 👍

gropaul:
Hi @Mytherin, I altered `ValidateDictionary` so that it is only 2% of the total runtime now for `SELECT DISTINCT l_shipinstruct FROM lineitem ORDER BY l_shipinstruct` so I think overallocating the dictionary is not necessary anymore. For both larger and smaller dictionaries, the regressions should be smaller now; all my benchmarks showed them at below 10% of the dictionary scan time.

Mytherin:
Sounds good - could you just fix the CI issues (seems to be caused by recent exception type changes in corrupted data)?

gropaul:
Hi @Mytherin, I fixed the error message; CI looks (mostly?) green now. I double-checked the IMDB regressions from the benchmark runner; they don't materialize on my M4, so I think this is github runner jitter.

PR #24016, opened 2026-07-21
Storage: validate deprecated FSST metadata

wufengwind:
Deprecated FSST segments store a bitpacking width that determines where the encoded string length buffer ends and the FSST symbol table begins. The reader trusted this value directly, so a checksum-valid corrupted file could make DuckDB decode strings using the wrong layout instead of reporting corruption.

This patch validates that the persisted bitpacking width matches the stored segment layout before importing the FSST symbol table. It also rejects out-of-range FSST metadata as IO errors.

This complements #23828, which validates dictionary offsets; this PR validates the deprecated FSST header layout/bitpacking width before decoding.

A regression test is added with a checksum-valid FSST database where the persisted bitpacking width has been corrupted.

Tests:
- `make reldebug -j4`
- `build/reldebug/test/unittest test/sql/storage/compression/fsst/invalid_fsst_bitpacking_width.test`
- `build/reldebug/test/unittest test/sql/storage/compression/fsst/fsst_read.test_slow`


Mytherin:
Thanks!

PR #24019, opened 2026-07-21
Add assert_true macro

carlopi:
There is precedence in [Spark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.assert_true.html) and [Databricks](https://docs.databricks.com/aws/en/sql/language-manual/functions/assert_true)

### DIFF
--- a/test/sql/function/generic/assert_true.test
+++ b/test/sql/function/generic/assert_true.test
@@ -1,0 +1,55 @@
+# name: test/sql/function/generic/assert_true.test
+# description: Test assert_true function
+# group: [generic]
+
+statement ok
+PRAGMA enable_verification
+
+# like Spark's assert_true, returns NULL when the assertion holds
+query I
+SELECT assert_true(true)
+----
+NULL
+
+query I
+SELECT assert_true(1 = 1, 'never shown')
+----
+NULL
+
+statement error
+SELECT assert_true(false)
+----
+Assertion failed
+
+statement error
+SELECT assert_true(NULL)
+----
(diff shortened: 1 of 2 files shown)


carlopi:
Follow up of https://github.com/duckdb/duckdb/pull/23214

test/sql/function/generic/assert_true.test
```diff
@@ -0,0 +1,55 @@
+# name: test/sql/function/generic/assert_true.test
+# description: Test assert_true function
+# group: [generic]
+
+statement ok
+PRAGMA enable_verification
```
pdet: nit: deprecated

test/sql/function/generic/assert_true.test
```diff
@@ -0,0 +1,55 @@
+# name: test/sql/function/generic/assert_true.test
```
pdet, line 1: Can we add `SELECT assert_true(true, ['a','b']);` as a test?

pdet (commented):
Thanks for the PR, two comments

test/sql/function/generic/assert_true.test
carlopi: Added to the test, raises a binder error like "Cannot concatenate types VARCHAR and VARCHAR[]"

pdet: Is that the desired behavior?

Mytherin:
Thanks!

PR #24024, opened 2026-07-21
EXTERNAL RESOURCES: Improved interface for listing arbitrary ones

carlopi:
Follow up of https://github.com/duckdb/duckdb/pull/23910, part 2/4 of https://github.com/duckdb/duckdb/pull/23736

This adds an optional listing callback to `register_external_resource_type` that can then be used via:
```sql
FROM duckdb_external_resources(discover := true);
```

Structure of the code is very similar to https://github.com/duckdb/duckdb/pull/23910, with tests on mock resources.

To be followed by SQL level interface.


src/function/table/system/external_resources.cpp
```diff
@@ -31,25 +46,117 @@ struct ExternalResourcesGlobalState : public GlobalTableFunctionState {
...
 	return_types.emplace_back(LogicalType::VARCHAR);
 	names.emplace_back("handle");
 	return_types.emplace_back(LogicalType::MAP(LogicalType::VARCHAR, LogicalType::VARCHAR));
-	return nullptr;
+	return std::move(result);
+}
+
+//! A stable key for a handle, so a discovered resource can be matched against a locally registered one.
+static string HandleKey(const Value &handle) {
+	return handle.IsNull() ? string() : handle.ToString();
```
pdet, line 80: Shouldn't a key be `(type, handle)`?

pdet: If, so, can we add tests for this?

pdet (commented):
Thanks for the PR, just a comment.

src/function/table/system/external_resources.cpp
carlopi: Fixed, thanks a ton!

PR #24025, opened 2026-07-21
Add accessor to get constant function arguments

Maxxen:
This PR adds a set of helper functions to the `Bind(Scalar/Aggregate/Window)FunctionInput` to (try)get "constant" argument values during binding.

The main motivation behind this PR is to 
1. make constant handling less bug-prone (lots of call sites forget to check for unresolved parameters),
2. make error messages/handling when trying to fold constants better and more consistent
3. eventually make `TableFunction`/`SimpleNamedParameterFunction`s handle constants/"named parameters" the same way so that they also can be based on the `FunctionSignature` and share the same binding logic with other functions.

Previously lots of bind callbacks ended up doing something like this:

```c++
static unique_ptr<FunctionData> MyBindCallback(BindScalarFunctionInput &input) {
    // ...
    
    auto &context = input.GetClientContext();
    auto &arguments = input.GetArguments();
       
    if (arguments[1]->HasParameter()) {
        throw ParameterNotResolvedException();
    }
    if (!arguments[1]->IsFoldable()) {
        throw BinderException("Argument to function must be a constant!");
    }
    auto my_constant = ExpressionExecutor::EvaluateScalar(context, *arguments[1]);

    // do something with the constant
    // ...
}
```

Now you can just do

```c++

static unique_ptr<FunctionData> MyBindCallback(BindScalarFunctionInput &input) {
    // ...
    Value my_constant = input.GetConstant(1);
    
    // do something with the constant
    // ...
}
```

This takes care of checking for sql-parameters (e.g. `(?)`/`LogicalTypeId::UNKNOWN`, folding the expression using the expression executor, but also throws nice _consistent_ error messages when a requested argument is not constant. 

E.g.

```sql
select current_setting(i::VARCHAR) from range(0, 10) r(i);
----
Binder Error:
The 'setting_name' argument in function 'current_setting' must be a constant expression                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                            
LINE 1: select current_setting(i::VARCHAR) from range(0, 10) r(i);                                                                                                                                                                                                                                          
                                ^                                         
```

There is also a `optional<Value> TryGetConstant(<arg>)` which will return `std::nullopt` instead of throwing if the argument could not be folded.

Additionally there is a `GetNonNullConstant` variant too, which will also guarantee (or throw a nice error) that the folded constant is non-null.

```sql
select current_setting(null);
----
Binder Error:
The 'setting_name' argument in function 'current_setting' must not be NULL                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                            
LINE 1: select current_setting(null);                                                                                                                                                                                                                                                                       
                               ^            
```

You can call both `input.TryGetConstant(<arg>)` and `input.GetConstant(<arg>)` to get a constant both by parameter name and by argument index, so it also works for variadic functions.

The error messages/name lookup uses the parameter names as defined in the `FunctionSignature`, so this PR also adds explicit (not auto-generated) parameter names to the functions that now use `GetConstant` so that the error messages you get when passing non-constant arguments to these function don't regress in descriptiveness.

--------
There is also another slight behavior change/bugfix in this PR. `IntegerNegateBind` used to not check if the argument was foldable, only if it was a `BOUND_CONSTANT`. This now uses `TryGetConstant` so `-(<foldable expr>)` at a types minimum value promotes correctly. E.g. `-(CAST(-128 AS TINYINT)) -> 128::SMALLINT)` used to overflow, but it now works).


Mytherin:
Thanks for the PR! Looks great - can you just fix up the merge conflicts?

Maxxen:
@Mytherin done

Mytherin:
Thanks!

PR #24028, opened 2026-07-21
Pass bind data to aggregate state size/init callbacks and batch state init

evertlammerts:
This PR changes the size and init callback signatures of aggregate functions to pass in a `AggregateStateInput`, which contains both the `BoundAggregateFunction` (which it already got) and an `optional_ptr<FunctionData> bind_data` (which is the addition).

One other change is that the init callback now gets the `state`s for all groups at once (i.e. `data_ptr_t *states, idx_t count`) so that it gets called for the batch of groups instead of once per group. Note that `update`, `combine`, `finalize`, and `destroy` already take an array of states.

This is prep work for the new C API. Adding the bind data to size and init allows us to generalize across callbacks, and batching init is to avoid ABI crossings.

I've patched the v1 C API bridge to make sure the batched init works correctly without any consumer changes.


Maxxen:
Amazing, thanks for picking this up!

Mytherin:
Thanks!

Maxxen (approved):

PR #24029, opened 2026-07-21
Internal #9937: AsOf Cardinality Estimates

hawkfish:
* Unless it is a right join, AsOf will only prouce at most the cardinality of the left side.

### DIFF
--- a/src/optimizer/join_order/relation_statistics_helper.cpp
+++ b/src/optimizer/join_order/relation_statistics_helper.cpp
@@ -411,6 +411,21 @@ RelationStats RelationStatisticsHelper::CombineStatsOfNonReorderableOperator(Log
 		ret.cardinality = child_cardinalities[0];
 		break;
 	}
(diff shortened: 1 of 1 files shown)


Tmonster (approved):
lgtm, 
Thanks!

Tmonster:
Does this produce bad join orders with poor performance currently? 
Maybe it is handy to also add a micro benchmark, although those are not currently run, so it's not super necessary.

hawkfish:
> Does this produce bad join orders with poor performance currently? Maybe it is handy to also add a micro benchmark, although those are not currently run, so it's not super necessary.

I just noticed this when looking at plans from the Kx NYSE benchmark. I don't think it impacts join orders because AsOf is not movable, but downstream operators might benefit from knowing the lookup table is small. This seems to be a common use case for AsOf: looking up small set of values in huge tables.

Tmonster:
Makes sense, thanks for fixing 🙏

PR #24030, opened 2026-07-21
Fix statistics propagation for window aggregate functions (#23663)

### LINKED ISSUE #23663: BITSTRING_AGG without explicit min/max fails as a window aggregate
### What happens?

`BITSTRING_AGG(id)` can infer the required min/max statistics when used as a regular aggregate over a table column, but the same one-argument aggregate fails when used as a window aggregate over the same table column. The query below should return the one-row bitstring result. Instead, DuckDB raises a binder error saying that required statistics could not be retrieved.

This appears to be a documented aggregate-window shape:

- DuckDB documents `bitstring_agg(arg)` as a general aggregate function, with example `bitstring_agg(A)`: <https://duckdb.org/docs/current/sql/functions/aggregates.html>.
- DuckDB's window function documentation says that all aggregate functions can be used in a windowing context: <https://duckdb.org/docs/current/sql/functions/window_functions.html>.

This happens on current main (23717c6d33) and also on v1.5.4.

### To Reproduce

To reproduce:
```sql
CREATE TABLE bt AS
SELECT *
FROM (VALUES (1)) AS v(id);

-- Works: regular aggregate infers the required min/max range.
SELECT BITSTRING_AGG(id)::VARCHAR AS regular_aggregate
FROM bt;

-- Fails: same one-argument aggregate used as a window aggregate.
SELECT
  id,
  BITSTRING_AGG(id) OVER (
    ORDER BY id
  )::VARCHAR AS window_aggregate
FROM bt
ORDER BY id;
```

Run:
```sh
```

Actual for the one-argument window aggregate:
stderr:
```text
Binder Error: Could not retrieve required statistics. Alternatively, try by providing the statistics explicitly: BITSTRING_AGG(col, min, max)
```

Expected for the failing window query:

| id | window_aggregate |
| --- | --- |
| 1 | 1 |

A workaround:

```sql
-- Workaround: explicit constant min/max avoids the failing stats-derived path.
SELECT
  id,
  BITSTRING_AGG(id, 1, 1) OVER (
    ORDER BY id
  )::VARCHAR AS explicit_bounds_window_aggregate
FROM bt
ORDER BY id;
```
The workaround returns:

| id | explicit_bounds_window_aggregate |
| --- | --- |
| 1 | 1 |

Notes From Minimization:

- Supplying explicit bounds, e.g. `BITSTRING_AGG(id, 1, 1)`, avoids the binder error, so the failing reduced shape is the one-argument statistics-derived form.
- One table row is enough; adding more rows is not required to reproduce the binder failure.

---

Cause Analysis (generated by coding agents):

The failure is in `extension/core_functions/aggregate/distributive/bitstring_agg.cpp`: the one-argument `bitstring_agg` path stores `min`/`max` in `BitstringAggBindData` through the aggregate statistics callback. The regular aggregate path can propagate input column statistics and therefore initializes that bind data. The window aggregate path reaches execution with null `min`/`max` bind data for the same table column and throws:

```text
Could not retrieve required statistics. Alternatively, try by providing the statistics explicitly: BITSTRING_AGG(col, min, max)
```

There are two plausible fixes: either make the window aggregate path propagate the same usable input statistics for `BITSTRING_AGG(arg)`, or explicitly document/reject the one-argument form in window contexts and require constant `BITSTRING_AGG(col, min, max)`. The current behavior is inconsistent with the documented `bitstring_agg(arg)` aggregate plus the documented rule that aggregate functions can be used as window functions.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (23717c6d33)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

marcodelpin:
Fixes #23663.

@hawkfish reviewed the change on the issue and asked me to open a PR ([comment](https://github.com/duckdb/duckdb/issues/23663#issuecomment-5034466055)).

### Problem

Stats-dependent aggregates such as `BITSTRING_AGG` can infer their required min/max statistics when used as a regular aggregate, but the same one-argument form failed as a *window* aggregate with `Could not retrieve required statistics`. `StatisticsPropagator::PropagateStatistics(LogicalWindow)` never propagated statistics to the window function's argument children, nor invoked the aggregate statistics callback, so a stats-dependent window aggregate never received its `SetStatisticsCallback`.

### Fix

- Decouple the aggregate statistics callback from `BoundAggregateExpression`: it now receives the mutable `BoundAggregateFunction`, the distinct flag and the argument children, so both the regular-aggregate and the window-aggregate paths can invoke it.
- `propagate_window.cpp` now propagates statistics to the window arguments and calls the callback for aggregate window functions.

### Test

Adds `test/sql/window/test_window_agg_stats_propagation.test` covering `BITSTRING_AGG` (and related stats-dependent aggregates) used as window functions without explicit min/max.

---
*Disclosure: this change was prepared with AI assistance (Claude) and reviewed by me before submission.*


hawkfish (approved):
LGTM!

hawkfish:
Hi @marcodelpin - The code looks fine to me but there is a CI failure. Looks like the test fails when the optimiser is disabled?

hawkfish (approved):

marcodelpin:
Thanks @hawkfish — sorry for the slow reply. That failure was addressed in 835abe1b2: the new test asserts that `BITSTRING_AGG(col)` works *without* explicit min/max precisely because the optimizer now propagates window-aggregate statistics — so under `debug_disable_optimizer=true` no statistics exist at all and the binder error is the expected property, not a defect. The test is now listed in `test/configs/stats_dependent_tests.json` ("Statistics are not set without optimizers", which `disable_optimizer.json` pulls in via `inherit_skip_tests`) and in `verify_aggregate_state_export.json` (pre-existing FIXME: `bitstring_agg` has no state-export callback), following the existing pattern of both lists.

The current red on the branch is unrelated to this PR — it arrived with the Aug 9 merge of `main` to restore mergeability: `duckdb_eviction_queues` and `concurrent_checkpoint_insert` retry-flakes, an `fsst_greedy_long_strings` wrong-result under `verify_fetch_row` (nothing in this PR touches compression), and an AlmaLinux mirror timeout in the extensions job. I have re-run the failed jobs.

hawkfish:
@marcodelpin Can you mark this as not being draft?

marcodelpin:
@hawkfish Done — it is out of draft. Worth noting for context: `duckdb-draftbot` converts this PR back to draft automatically whenever the branch CI goes red (it did so on Aug 9 and again on Aug 11 at 23:25), so the draft flag has been tracking CI rather than my intent.

Since that CI red is what keeps tripping the bot, I ran the control that separates "our diff" from "our runners", and the result is worth having before you spend review time on it.

**The MinGW failure is not from this PR.** I pushed upstream `main` at `2828abd8a` — the exact merge-base of this branch, verbatim, with none of my commits — to a branch on my fork and ran the same workflow on the same runners. It fails on the identical assertion:

```
FAIL test/sql/parallelism/interquery/concurrent_checkpoint_insert.test
Wrong result in query! (test/sql/parallelism/interquery/concurrent_checkpoint_insert.test:17)
```

Upstream CI on that same commit passes that job, so this is a property of my fork's runners rather than of the code. It is a positive reproduction on unmodified upstream, not an inference from absence.

**The Linux failure looks like runner starvation, though I am less certain of that one.** In the `verify_fetch_row` config the branch timed out on `duckdb_eviction_queues` in one run, and on `duckdb_eviction_queues` **and** `substring_zonemap_pruning` in the re-run — a different test set each time, always a 600 s timeout rather than a wrong result, in a config that took 1982 s while every sibling config in the same job finished in roughly 200 s. The control passed that job, but only once, so I would not claim this half is settled the way the MinGW half is.

One thing I would rather flag than leave for you to find: `substring_zonemap_pruning` is a statistics test, and this PR touches statistics propagation. That adjacency is real. It timed out rather than producing a wrong result, and it passed in the first run, so nothing yet distinguishes it from its slow neighbours — but if it ever fails with a **wrong result** instead of a timeout, that would be a genuine signal and I would want to look again.

For the earlier `debug_disable_optimizer` question, my Aug 11 reply still stands: the new test asserts that `BITSTRING_AGG(col)` works without explicit min/max precisely because the optimizer now propagates window-aggregate statistics, so with optimizers disabled the binder error is the expected property. It is listed in `stats_dependent_tests.json` and `verify_aggregate_state_export.json` following the existing pattern for both.

Happy to rebase onto current `main` if that helps — one of the commits since the merge-base touches `verify_aggregate_state_export.json`, which this PR also edits, so that seam will need resolving either way.

hawkfish:
Yeah sorry there are some failures on `main`. One of mine is blocked too. They are being fixed.

(BTW, [you don't have to rebase](https://www.dailymotion.com/video/x4blt6l) - just merge main.)

hawkfish:
Please check the CI failures.

marcodelpin:
@hawkfish Checked, here is what I find, split by where the red comes from.

**On this PR, the checks that completed are green.** Three `Main` runs are attached to `0a56672a1e`, and GitHub cancelled the first one as superseded:

| run | jobs |
|---|---|
| [33076227339](https://github.com/duckdb/duckdb/actions/runs/33076227339) | 14 cancelled, 1 failure (`Summary`), 9 skipped, **superseded** |
| [33076243356](https://github.com/duckdb/duckdb/actions/runs/33076243356) | 1 success, 23 skipped |
| [33076340964](https://github.com/duckdb/duckdb/actions/runs/33076340964) | **50 success, 21 skipped, 0 failure** |

Every non-green entry in the rollup belongs to the superseded run, and the one `FAILURE` is the aggregator job `Summary`, which reports "one or more required jobs failed or were cancelled". The API currently returns `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` for the PR.

**On my fork's CI there is one genuinely red job**, `Linux Config (Execution)`: 4557 passed, 1 failed, 72 skipped in 1987s. The content is two tests hitting the 600s cap after four retries each, under `test/configs/verify_fetch_row.json`:

- `test/sql/table_function/duckdb_eviction_queues.test`
- `test/sql/optimizer/substring_zonemap_pruning.test`

Both are timeouts, not wrong results, which is what I would expect from a change that only propagates statistics.

This is the same pair I already flagged in [my Aug 12 note](https://github.com/duckdb/duckdb/pull/24030#issuecomment-5272605281), and I want to be careful not to over-claim it. In that note the merge-base-verbatim control settled the **MinGW** half positively (upstream `main` at `2828abd8a`, none of my commits, failing on the identical `concurrent_checkpoint_insert` assertion). For this **Linux** half I said the control had passed the job only once and that I would not claim it settled, and that is still true today.

What today adds is the test I set for myself there. I wrote that `substring_zonemap_pruning` is a statistics test, that the adjacency to this PR is real, and that a **wrong result instead of a timeout** would be the genuine signal. It is a 600s timeout again, after four retries, in the same `verify_fetch_row` config that took 1987s while its sibling configs finish in roughly 200s. So by my own criterion this is still not a signal, but it is still not settled either.

**A useful baseline**: on the previous head `5d2ad642`, `Main` had six red jobs, `Linux Config (Storage Engine)`, `Linux Config (Execution)`, three regression benches and `Summary`. On the current head it is two. The picture improved after the merge, and `Linux Config (Execution)` is the one that persists across both.

**One thing I have not yet ruled out, so I would rather say it than leave it out.** My fork's nightly lane is broadly red (ICU calendar tests, huge nested payloads, concurrent index writes, file-cache spill, plus a few `make` failures), which reads as a lane-wide environment problem rather than anything this diff does. But three of those failures do sit in this PR's own territory:

- `test/optimizer/statistics/statistics_count_not_null_pruning.test:99`
- `test/sql/optimizer/secure_view_statistics.test:18`
- `test/sql/window/test_custom_spooling.test_slow`

There is no baseline for them, that lane ran for the first time on this head. I am re-running the same merge-base-verbatim control against them now and will post the result here, so we can tell my environment apart from this diff rather than assume it.

Happy to rebase, re-run anything, or split the change if that helps review.

marcodelpin:
@hawkfish First, on the CI failures.

On the head you asked about (`0a56672a1e`), `Main` had two red jobs: `Linux Config (Execution)` and `Summary`. To see whether either came from this PR, I pushed a control branch that is upstream `main` **verbatim** at this PR's merge base (`73ee3f2d46`, none of this PR's commits) and ran the same two workflows on it:

| workflow | control, no PR commits | this PR at `0a56672a1e` |
|---|---|---|
| `Main` | 3 failed: `Linux Config (Execution)`, `Summary`, `extensions / Main (core) / windows_amd64_mingw` ([run](https://github.com/marcodelpin/duckdb/actions/runs/33134655701)) | 2 failed: `Linux Config (Execution)`, `Summary` ([run](https://github.com/marcodelpin/duckdb/actions/runs/33076233771)) |
| `NightlyTests` | 15 failed jobs ([run](https://github.com/marcodelpin/duckdb/actions/runs/33134630063)) | 15 failed jobs ([run](https://github.com/marcodelpin/duckdb/actions/runs/33076232961)) |

I found no PR-specific failure on that head: the 15 failing `NightlyTests` jobs matched the control, and both failing `Main` jobs also failed in the control. `Linux Config (Execution)` is the same two tests either way, `duckdb_eviction_queues.test` and `substring_zonemap_pruning.test`, timing out under `test/configs/verify_fetch_row.json`. `OSX` was green on that head ([run](https://github.com/marcodelpin/duckdb/actions/runs/33076232955)).

## A compile break I introduced afterwards, and fixed

Separately: on 2026-08-28 I merged current `main` into the branch (`5afaf43fdb`, 102 commits). My widened callback became incompatible with 12 `AggregateFunction::PropagateInputValueStats` registrations, across five files, added since the branch point. This PR takes `aggregate_statistics_t` from three parameters to five.

The callback receives `BoundAggregateFunction`, but the guard needs the export mode stored on `BoundAggregateExpression`. Rather than drop that guard I carried the export mode on `AggregateStatisticsInput`, which the caller already constructs (`1017c032fe`, 4 files, +18/-7):

- `aggregate_state.hpp`: new `AggregateStateExportMode state_export_mode` member; the added constructor parameter defaults to `NONE`, so `propagate_window.cpp` is unchanged
- `propagate_aggregate.cpp`: passes `aggr.StateExportMode()`, where the expression is still in scope
- `aggregate_function.cpp`: reads `input.state_export_mode`, and takes the return type from `function` rather than the expression

That preserves the existing state-export guard. I compiled each file registering an aggregate statistics callback against the new signature, and checked the compile could actually fail by reverting the declaration, which breaks `minmax.cpp` with two errors.

`f413f7e3ec` is the follow-up commit and is clang-format only, on the two `aggregate_function` files.

## What I would like to confirm

CI on the current head `f413f7e3ec` has finished: [`OSX`](https://github.com/marcodelpin/duckdb/actions/runs/33169190800) green, and [`Main`](https://github.com/marcodelpin/duckdb/actions/runs/33169191286) 83 of 93 jobs green. The two failures are `Linux Config (Execution)` and `Summary` - the same two this PR had at `0a56672a1e`, and a subset of the control's three. `Linux Config (Execution)` is again `verify_fetch_row.json` failing on `duckdb_eviction_queues.test` and `substring_zonemap_pruning.test`, the same pair as the control.

Is preserving the five-parameter callback the right call, or would you rather I reverted the arity change and found another route to the state export mode? Happy to rework it either way. I would rather ask than guess on a signature this widely registered.

hawkfish:
Hi @marcodelpin - I think the correct way to fix this is to just make a synthetic `BoundAggregateExpression` and call the original API with that object. 

I do like the fact that the now have somewhere to compute statistics for non-aggregate window functions, and I should probably add that to the 2.0 APIs.

As for CI failures, I can ask for restarts if they recur.

marcodelpin:
Thanks - that is much better, and the diff is a lot smaller for it. Pushed as `2d05b7ffa3`.

`aggregate_statistics_t`, `AggregateStatisticsInput`, `aggregate_function.cpp`, `propagate_aggregate.cpp`, `count.cpp`, `sum.cpp` and `bitstring_agg.cpp` are all back to their unmodified upstream form, so **`propagate_window.cpp` is now the only source file this PR touches**.

Two things I ran into building it, both worth flagging since they are properties of the callback contract rather than of the window path:

**Callbacks rewrite the function, not just the children.** `CountPropagateStats` calls `ReplaceImplementation(CountStarFun::GetFunction())`, `SetName("count_star")` and `GetChildrenMutable().clear()`, and `SumPropagateStats` calls `ReplaceImplementation` too. Against a synthetic expression those mutations land on its *copy* of the function, so they have to be written back or they are silently dropped - a windowed `count(x)` over a NOT NULL column would keep the original unary function with no children. So the call site moves the children in, and afterwards writes back both the children and the function.

**`removed_aggregate_children` needs maintaining here too.** The aggregate path sets it and `Optimizer::RunOptimizer` reruns unused-column elimination on it; without it a window `count(x)` rewritten to `count_star` leaves the now-unreferenced column in the scan. That was missing from my earlier version as well, not just this one.

Added regression coverage for the rewrite: `count(id) OVER ()` and a framed `ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW` over a NOT NULL column, plus a nullable column where the rewrite must not fire.

On the earlier CI question - thanks, and nothing further needed from me there. The remaining `Linux Config (Execution)` red is the `verify_fetch_row.json` timeout pair, which fails identically on a merge-base control with none of this PR's commits.

One caveat I would rather state than have you discover: I have no local duckdb build, so the new test expectations were checked against a released duckdb rather than against this tree, and the rewrite path itself is only exercised by CI.

marcodelpin:
Quick status roll-up, and one ask.

Since the last review round the branch was reworked per the earlier feedback (head `2d05b7ffa3`). On our fork's runners:

- OSX: green.
- Main: 83 jobs succeeded; two failures remain ({Linux Config (Execution), Summary}, a `verify_fetch_row.json` timeout pair; the rest of the matrix is skips). The same two tests also fail on an upstream-only control branch we keep on the fork (`ci-control/upstream-2828abd8a`, no fork changes), which suggests they are independent of this diff. The suites that execute the new regression file are among the green ones.

Fork runs have also encountered unrelated installation timeouts on earlier rounds, so the ask: would it be possible to run the workflows upstream on this PR for a clean read? Happy to rebase or adjust anything first if that helps.

marcodelpin:
Thanks for enabling auto-merge! One thing I noticed while checking why it has not fired yet: the attempt-3 re-run of `Main` stopped on `Linux Config (Query Verification)` at the `actions/download-artifact` step (`Failed to ListArtifacts: ... 403 Forbidden`), before any test ran. My reading is that a re-run of only the failed jobs could no longer see the build artifacts of the original Aug 28 run; every other job is green and the head (2d05b7ffa3) is unchanged since the last review round.

I have deliberately not pushed anything, since a contributor push would move the PR back to draft and drop the auto-merge. Would a re-run of the whole `Main` workflow (or a direct merge) be the easiest way to unblock it on your side? Happy to do whatever is more convenient for you.

lnkuiper:
I've restarted the failed runs, so it should enter the queue once those are green.

lnkuiper:
Actually, I'll disable auto-merge for now, we're about to create the v2.0 branch, and this should go into that.

PR #22211, opened 2026-04-22
Optimize Recursive CTE Performance

kryonix:
I'm very excited finally being able to share this PR, which I wanted to implement for literally years now, but never had the time to actually do it. Efficient execution of recursive CTEs is not something many people think or care about, so there is little to no research done on it, and (good) open source implementations are basically non-existent. However, I have spent more time than I care to admit on this topic, and have compiled the best insights and ideas into this PR.

Recursive CTEs are expensive to execute, because the current implementation is simple, but lacks optimization. The basic principle of how execution works is surprisingly simple, and is based on the semi-naive evaluation strategy. That however, is where the simplicity ends, and performance, parallelization, and optimization begin. DuckDB's main bottleneck is that for the recursive part of the CTE, all pipelines are created, scheduled, executed, and destroyed for each iteration of the recursive CTE. This is extremely expensive, and the main reason why recursive CTEs can be slow, especially when the recursive part operates on small amounts of data. In that case, the overhead of creating and destroying pipelines dominates the execution time and does not amortize well. The main bottlenecks are:

- The creation and destruction of pipelines for each iteration of the recursive CTE.
- Static and overly parallel execution of the recursive part of the CTE, which does not allow for efficient execution when the recursive part operates on small amounts of data.
- No reuse of any execution state across iterations of the recursive CTE, which leads to redundant work and overhead.

This PR addresses these bottlenecks by introducing a new execution strategy for recursive CTEs, which allows for more efficient execution. The main changes are:

1. Thread counts are now determined dynamically based on the size of the data being processed, allowing for more efficient execution when the recursive part operates on varying amounts of data across iterations. As we know perfect cardinality estimates for the recursive working table, we can use that to determine the optimal number of threads to use for each iteration of the recursive CTE, which allows us to avoid the overhead of parallel execution when it is not beneficial, and to use more threads when it is.
2. Execution state is now reused across iterations of the recursive CTE, which reduces redundant work and overhead. This is especially efficient if the HashJoin operator is used in the recursive part of the CTE, as it can reuse the hash table across iterations.
3. The Planner now tries to put the recursive part of the CTE on the PROBE side of a HashJoin whenever possible, to make use of the hash table reuse optimization.
4. Physical operators now have Reset() methods, which allow for resetting the state of the operator without destroying and recreating it. This is used to reset the state of the operators in the recursive part of the CTE across iterations.
[...]

PR #24031, opened 2026-07-21
Harden recursive CTE execution and optimize `USING KEY` joins

### LINKED ISSUE #20838: Performance Regression Recursive CTE
### What happens?

The query below finishes in a few milliseconds on v1.4, but takes about 1 second on 1.5 and main (release build). The query plans are basically identical. Only difference: With v1.5 and main, the recursive CTE creates a full data chunk with 2048 rows before termination. With v1.4, far less values were produced before early termination by the `LIMIT`.

### To Reproduce

```sql
WITH RECURSIVE
    "@loop" ("$label", "$output", "0") AS (
      (SELECT CAST('$' AS VARCHAR) AS "$label", CAST(NULL AS BIGINT) AS "$output", CAST(NULL AS BIGINT) AS "0")
        UNION  ALL
      (WITH
        "0" AS (SELECT NULL FROM "@loop" WHERE "@loop"."$label" IS NOT DISTINCT FROM '$'),
        "1" (i) AS (SELECT 0 AS i FROM "0" AS "0"),
        "2" (i) AS (SELECT "@loop"."0" AS i FROM "@loop" WHERE "@loop"."$label" IS NOT DISTINCT FROM '$loop0'),
        "3" (i) AS ((SELECT "1".i AS i FROM "1" AS "1"(i)) UNION ALL (SELECT "2".i AS i FROM "2" AS "2"(i))),
        "4" (i) AS (SELECT (SELECT (("3".i + 1) % 1000)) AS i FROM "3" AS "3"(i)),
        "5" ("$output") AS (SELECT "4".i AS "$output" FROM "4" AS "4"(i)),
        "6" ("$label") AS (SELECT '$loop0' AS "$label", i FROM "4" AS "4"(i))
        (SELECT CAST(NULL AS VARCHAR) AS "$label", CAST("5"."$output" AS BIGINT) AS "$output", CAST(NULL AS BIGINT) AS "0" FROM "5" AS "5"("$output"))
          UNION ALL
        (SELECT CAST("6"."$label" AS VARCHAR) AS "$label", CAST(NULL AS BIGINT) AS "$output", CAST("6".i AS BIGINT) AS "0" FROM "6" AS "6"("$label"))))
  SELECT "@loop"."$output" FROM "@loop" AS "@loop"("$label", "$output", "0") WHERE ("@loop"."$label" IS NULL) LIMIT 10
  ;
```

### OS:

not relevant

### DuckDB Version:

main, v1.5, v1.4

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Denis Hirn

### Affiliation:

University of Tübingen

- [x] Yes, I have

- [x] Yes, I have

Yes

### LINKED ISSUE #24109: Recursive CTE does not prune unused columns
### What happens?

See repro for details.

### To Reproduce

```sql
-- unused columns for both tables are not pruned
memory D EXPLAIN
         WITH RECURSIVE r(a, b, c) AS (
             SELECT
                 1,
                 2,
                 repeat('x', 1000)
             UNION ALL
             SELECT
                 a + 1,
                 b + 1,
                 c
             FROM r
             WHERE a < 3
         )
         SELECT a
         FROM r;
╭─ Projection ─────────────────────────────────────╮
│ Projections: #0                                  │
│ ~1,001 rows                                      │
╭─ Rec CTE ───────────────┴────────────────────────╮
│ CTE Name: r                                      │
│ Table Index: 1                                   │
│ ~1,001 rows                                      │
                          ├──────────────────────────────────────╮
╭─ Projection ────────────┴────────────────────────╮  ╭─ Projection ────────╮
│ Projections:                                     │  │ Projections: a + 1, │
│ 1, 2,                                            │  │              b + 1, │
│ 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx… │  │              c      │
│ ~1 row                                           │  │ ~1 row              │
╰─────────────────────────┬────────────────────────╯  ╰──────────┬──────────╯
╭─ Dummy Scan ────────────┴────────────────────────╮  ╭─ Filter ─┴──────────╮
╰──────────────────────────────────────────────────╯  │ Expression: a < 3   │
                                                      │ ~1 row              │
                                                      ╰──────────┬──────────╯
                                                      ╭─ Rec CTE Scan ──────╮
                                                      │ CTE Index: 1        │
                                                      │ ~1 row              │
                                                      ╰─────────────────────╯
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

kryonix:
After #22211 was merged, I went back through recursive CTE execution to audit the assumptions behind retained state, epoch scheduling, recursive sinks, and `USING KEY`. Retaining and resetting the physical recursive plan remains the right foundation, but I found several places where state lifetime was too optimistic or where each epoch still did substantially more work than necessary.

In short, this PR:

- Makes retained invariant builds safe for volatile and side-effecting operators.
- Reuses recursion-independent, physically materialized CTE producers across epochs.
- Preserves early termination through streaming `LIMIT`.
- Reuses cached recursive projections of the generic pipeline schedule and selects workers from the current epoch's work.
- Parallelizes broad `UNION ALL` and `UNION DISTINCT` sinks.
- Adds fixed-point column pruning for `UNION ALL` recursive CTEs.
- Freezes `USING KEY` state at epoch boundaries and reads it without rematerializing recurring state.
- Adds `PhysicalRecursiveCTEKeyJoin` for complete keys and secondary indexes for partial composite keys.
- Makes concurrent fallback scans use task-local hash-table probes and synchronized aggregate finalization.
- Adds conservative row-equality normalization, recurring-state join ordering, structured runtime metrics, and deterministic benchmarks.
- Adds generated Flummi programs covering control-flow-heavy graph, cellular-automaton, clustering, and ray-tracing recursion.
- Adds 16 adapted AoC 2022 workloads plus keyed AoC 2024 shortest-path and flood variants.
- Runs the recursive research, Flummi, AoC22, and AoC24 suites in CI's performance-regression workflow.

## Recursive execution

The most important correctness issue was invariant-build classification. A build is safe to retain only when it is independent of recursive state and repeatable across epochs. The previous check established independence but not repeatability, so a volatile build containing `nextval()` could produce different results when caching was enabled. A second unsafe assumption treated every unfamiliar logical operator as repeatable; this could freeze one unseeded Bernoulli sample across all recursive epochs.

There is now one conservative repeatability classifier shared by physical planning and recursive column pruning. It explicitly classifies the current logical operators, checks scalar, aggregate, and window function stability, recognizes both ordinary and pushed-down sampling, and defaults unknown operators and opaque table functions to rebuilding. Deterministic table functions opt in through a bind-data-aware capability. Seeded samples remain eligible, while unseeded samples, DML, side effects, dynamic filters, and unknown extensions do not.

The same dependency analysis now retains recursion-independent `PhysicalCTE` producers when their selected execution mode is fully materialized. The producer collection survives after the first epoch while each CTE scan is reset normally, so multiple consumers can reread it without rebuilding its source. Streaming and hybrid fan-out modes remain ineligible. This is entirely physical and adds no logical or serialization field.

Caching operators could also defeat partial demand: below a streaming `LIMIT`, a selective pipeline could withhold output until it had a complete vector, preventing early termination. Caching now respects the existing physical `CanStopSourceEarly()` property. This fixes #20838 without a recursive-specific setting or coupling to the materialized-CTE exchange from #23750; the reduced case improved from 0.36 to 0.01 seconds.

Inline and parallel recursion now consume projections of the generic pipeline schedule, while executors and lookup structures are reused on the warm path. These immutable projections are built once during physical pipeline construction, including the variant that omits retained invariant builds, rather than once per recursive state. This matters for nested queries such as AoC 2024 day 15, which creates 1,405 short-lived recursive states. Worker selection uses the current frontier, physical work units, recursive-reference count, configured threads, and the number of execute pipelines it would schedule. Materialized CTE scans count as fan-out work rather than fixed overhead; ordinary `UNION ALL` recursion stays inline unless each worker can amortize the remaining pipelines.

Broad `UNION ALL` epochs concatenate worker-local output at the boundary; small epochs retain locked append because private collections cost more there. `UNION DISTINCT` promotes broad parallel epochs to four persistent hash partitions, reducing the representative 1M-row workload from 0.88 to 0.68 seconds.

The CI regression on AoC day 25 exposed a related fixed-cost problem in narrow epochs. The runtime now avoids parallel-only work-unit discovery, scheduler lookups, worker-local resets, metrics branches, and settings lookups when the frontier already proves that execution is serial. While reducing that overhead, I found a more serious allocator-lifetime bug: after a worker-local collection was combined into the recursive output, its retained segment still owned the worker allocator, but `ResetForReuse()` reset only the destination collection's allocator. Repeated narrow epochs therefore accumulated blocks, spilled, and could fail with spilling disabled. `ResetForReuse()` now resets the retained segment's allocator as well. A 500,000-epoch reproducer that previously exhausted a 64 MB limit now completes without spilling.

Recursive `UNION ALL` column liveness is computed to a fixed point without copying or serializing the recursive subplan during analysis. This removes unused recursive expressions from both the anchor and recursive member, including the wide payload in the public reproducer. `UNION DISTINCT` retains every identity column, and `USING KEY` remains unpruned until key/update liveness can be proven.

Fixes #24109.

The AoC protection cases confirmed that the final scheduler does not materially regress ordinary recursive CTEs. In the final two-thread regression run, all 25 workloads passed and the geometric mean was effectively flat at 0.064 seconds on both builds (`+0.3%`). Day 15 improved from 0.251 to 0.244 seconds after caching the schedule projections, and day 25 remained effectively flat at 0.074 versus 0.073 seconds rather than reproducing the 20% regression from an earlier CI run.

## Frozen `USING KEY` state and indexed probes

The largest remaining `USING KEY` cost was access to `recurring.T`. The sink previously updated keyed state during an epoch and copied the complete aggregate hash table into a recurring collection for later reads. It now buffers candidates and commits them during prepare-finalize, after all execute stages have completed. The recursive term reads that frozen epoch snapshot directly from the aggregate hash table, and the source explicitly drains the final keyed state once recursion finishes.

For complete-key equality joins, I added read-only, missing-key-safe aggregate lookup APIs plus `PhysicalRecursiveCTEKeyJoin`. Planning selects it only for an inner join against a direct recurring-state scan with direct, exactly typed scalar key equalities. It handles either orientation, reordered composite keys, projections, duplicates, NULLs, and missing groups without mutating state.

Proper subsets of composite keys use an indexed lookup rather than a filtered scan. The state builds one epoch-stable secondary hash index per normalized subset, storing aggregate row addresses and collision links and checking full typed equality during probing. The row collection uses `KEEP_EVERYTHING_PINNED`, outlives the indexes, and is frozen during probes, so hash-directory growth cannot invalidate addresses. Only new complete-key groups extend an index; payload updates require no maintenance. The sparse partial-key workload improved from 0.788 to 0.059 seconds median, or 13.3x.

Eligibility is deliberately strict. Residual predicates, wrapped scans, nested or mismatched keys, and non-inner joins use the ordinary hash-join/full-scan path. Direct LEFT, SEMI, and ANTI prototypes were slower on dense access, so I did not add a workload-specific threshold. The operator cannot replace ordinary recursive references: those read the current frontier, whereas accumulated state contains older rows and does not exist for `UNION ALL`. Substitution would change results and potentially termination.

The final hardening binds lookup scratch to its originating hash table and replaces assertion-only executor-cache and schedule failures with checked internal errors.

Fallback scans needed a separate concurrency fix. Each scan had its own cursor lock, but all of them called the aggregate hash table's mutable scan/fetch path and finalized aggregate payloads from the same state. Two recurring references with an `approx_quantile` payload could therefore crash at eight threads. Fallback scans now scan only group columns, resolve row addresses through task-local read-only probe state, and share the CTE-wide aggregate-finalization lock with direct key joins. Append and lookup paths also compose the same owner-checked probe scratch layout.

## Runtime evidence and benchmarks

Runtime metrics use DuckDB's existing `PhysicalOperator` structured log type. Query state checks once whether logging is enabled; the disabled path avoids timers, byte scans, and counter updates. The metrics showed, for example, that the sparse complete-key case replaces roughly 20M full-state scan rows with 20,000 direct probes.

The metrics state is now encapsulated and also reports retained materialized-CTE production and reuse. The existing benchmark definitions are used directly by a dedicated CI regression job. It runs the 57-query recursive suite, covering the research, Flummi, asql-ws2526, and AoC 2022 workloads, plus all 25 AoC 2024 queries against the merge-base and candidate runners with two threads, seven timed runs, answer validation, and the standard confirmed-regression logic with a 10 ms absolute allowance.

I also added a 50,000-epoch, one-row-frontier `UNION ALL` benchmark to protect fixed recursive runtime and allocator-reuse costs. In the final regression run it was flat at 0.110 seconds on `main` and 0.108 seconds on the branch. Across all 57 recursive benchmarks, the geometric mean improved from 0.093 to 0.064 seconds, or 31.3%; every individual workload passed the regression gate.

I added a deterministic benchmark suite based on the queries accompanying [How DuckDB is `USING KEY` to Unlock Recursive Query Performance](https://db.cs.uni-tuebingen.de/publications/2025/using-key/). It covers eight algorithms plus isolated key probes. The Game of Life variant uses an equivalent equality-join formulation rather than another specialized operator.

Two workloads in the research suite exposed general optimizer opportunities. Connected components exposed row equality that hid scalar key predicates. It is now decomposed only at a filter root for non-empty, matching scalar column references, preserving NULL semantics elsewhere. BFS exposed a separate join-order problem: the full edge relation could reach recurring state before frontier restriction. LEFT, SEMI, and ANTI joins whose right subtree contains recurring state are now kept above candidate generation.

I measured the merge base against the PR. Both were Release builds with the same linked extensions on the same macOS arm64 machine. I ran every workload by its individual benchmark name with two threads, one runner warmup, 15 timed runs, and answer validation. This is the same protocol used for the Flummi table below and matches the recursive CI regression job. p95 uses the inclusive nearest-rank percentile.

| Workload | `main` median / p95 | PR median / p95 | Median speedup |
| --- | ---: | ---: | ---: |
| DVR | 0.6512 / 0.6944 s | 0.5302 / 0.5661 s | 1.23x |
| PageRank | 0.6544 / 0.6864 s | 0.3332 / 0.3628 s | 1.96x |
| k-means | 0.0944 / 0.0953 s | 0.0805 / 0.0820 s | 1.17x |
| Sparse complete-key probe | 0.3185 / 0.3696 s | 0.0477 / 0.0778 s | 6.67x |
| Dense complete-key probe | 0.2588 / 0.2958 s | 0.1659 / 0.1703 s | 1.56x |
| Sparse partial-key probe | 0.7877 / 0.9389 s | 0.0591 / 0.0609 s | 13.33x |
| BFS | 1.1782 / 1.2359 s | 0.0327 / 0.0616 s | 36.04x |
| Connected components | 4.5046 / 4.5654 s | 1.1273 / 1.1931 s | 4.00x |

The algorithms do not naturally probe a proper subset of a composite `USING KEY`; the partial-key cases are intentionally isolated operator benchmarks. The metrics also explain why I stopped adding specializations: DVR performs far more probes than state scans, PageRank already benefits from retained invariant joins, and k-means consumes complete state for grouping.

I also added seven answer-validated workloads generated by [Flummi](https://github.com/DBatUTuebingen/flummi). They exercise a very different shape: an imperative program counter, control-flow label, emitted values, and live variables all travel through one recursive CTE. The 384 by 240 ray trace retains shadows and reflections but is small enough for the Release regression job.

Flummi's Generations program exposed two successive worker-classifier mistakes. Its 52 small epochs initially launched too many tasks for its pipeline fan-out, while the post-#23750 classifier then counted materialized CTE scan pipelines both as useful fan-out work and as fixed scheduling overhead. A matched profile showed that this made the branch stay inline at two threads even though two workers were beneficial. CTE scan pipelines now contribute work without also inflating fixed overhead. In the 30-run confirmation, two-thread execution moved from 0.1494 to 0.1528 seconds (`+2.3%`, `+3.4 ms`), while the four-thread protection still avoids `main`'s over-parallelization and moves from 0.2974 to 0.1692 seconds.

Community detection exposed a separate planning regression in the first column-pruning implementation. Every fixed-point step copied the complete generated recursive subplan, and `LogicalOperator::Copy()` performs that copy through serialization. I replaced it with a non-mutating dependency analysis over the original plan. In the confirming profile, recursive dependency analysis fell from 796 to 7 samples; the one remaining `LogicalOperator::Copy()` sample came from unrelated CTE inlining.

The Flummi comparison uses the same matched builds and two-thread protocol.

| Flummi workload | `main` median / p95 | PR median / p95 | Median speedup |
| --- | ---: | ---: | ---: |
| Community detection | 0.0342 / 0.0362 s | 0.0330 / 0.0361 s | 1.03x |
| Connected components | 0.0369 / 0.0374 s | 0.0363 / 0.0584 s | 1.02x |
| Generations | 0.1494 / 0.1548 s | 0.1528 / 0.1600 s | 0.98x |
| k-core `USING KEY` | 0.4864 / 0.4901 s | 0.0016 / 0.0020 s | 305.91x |
| k-means | 0.1666 / 0.1690 s | 0.1670 / 0.1703 s | 1.00x |
| Game of Life | 0.0751 / 0.0772 s | 0.0746 / 0.0764 s | 1.01x |
| Ray tracing | 0.8507 / 0.8803 s | 0.8738 / 0.9176 s | 0.97x |

At two threads the seven-workload geometric mean improves by approximately 55.9%. This aggregate is dominated by keyed k-core. The Generations row uses the focused 30-run confirmation after the classifier fix; its remaining 2.3% delta and ray tracing's 2.7% delta both remain below the CI regression gate.

I also adapted 16 recursive SQL solutions from [DBatUTuebingen/Advent_of_Code](https://github.com/DBatUTuebingen/Advent_of_Code) at `10ead5d33412cb6484bf8b9bc4ed91f454c1fd88`. I did not copy the Advent of Code inputs; the load scripts generate deterministic data that retains the recursive plan shapes. The second group adds nested-list stack and sand state, hierarchical ascent, bit-packed falling rocks, expression-tree self-joins, alternating map branches, moving hazards, and balanced-base conversion. Four queries have answer-equivalent `USING KEY` alternatives. The reverse BFS and directed-graph traversal benefit because their keys remove redundant historical states. The permutation and moving-frontier variants overwrite almost every row in every iteration, so they deliberately remain negative controls.

I measured current local main against this branch with the same matched Release builds, two threads, one runner warmup, 15 timed runs, and answer validation used for the Flummi table.

| AoC 2022 workload | `main` median / p95 | PR median / p95 | Median speedup |
| --- | ---: | ---: | ---: |
| Day 9: chained rope recursion | 0.3207 / 0.3279 s | 0.3178 / 0.3217 s | 1.01x |
| Day 11: state machine | 0.6572 / 0.6679 s | 0.6518 / 0.6631 s | 1.01x |
| Day 12: reverse BFS | 0.2961 / 0.2993 s | 0.2936 / 0.3021 s | 1.01x |
| Day 12: `USING KEY` | 0.0283 / 0.0880 s | 0.0270 / 0.1093 s | 1.05x |
| Day 16: directed traversal and search | 0.6568 / 0.6842 s | 0.6296 / 0.6606 s | 1.04x |
| Day 16: `USING KEY` | 0.0241 / 0.0247 s | 0.0247 / 0.0252 s | 0.98x |
| Day 18: `UNION` flood fill | 1.3416 / 1.3719 s | 1.3666 / 1.3833 s | 0.98x |
| Day 19: materialized branching search | 0.3763 / 0.4070 s | 0.3776 / 0.4069 s | 1.00x |
| Day 20: recursive permutation | 0.1477 / 0.1512 s | 0.1482 / 0.1530 s | 1.00x |
| Day 20: `USING KEY` | 0.1493 / 0.1518 s | 0.1503 / 0.1541 s | 0.99x |
| Day 23: correlated frontier | 0.3335 / 0.3351 s | 0.3316 / 0.3334 s | 1.01x |
| Day 23: `USING KEY` | 0.3365 / 0.3374 s | 0.3367 / 0.3402 s | 1.00x |

The more useful keyed comparison is within the PR: Day 12 moves from 0.2936 to 0.0270 seconds (10.86x), and Day 16 from 0.6296 to 0.0247 seconds (25.54x). Days 20 and 23 are 1.4% and 1.5% slower, respectively. Day 12 has a visibly multimodal two-thread distribution in both builds, with occasional roughly 30 ms scheduler stalls, so its median and p95 are both reported.

Across the original 12 variants, the branch geometric-mean runtime improves by 0.8% with one thread, 0.6% with two threads, and 9.7% with four threads. The four-thread aggregate is strongly influenced by Day 12's scheduler stalls, so I do not use it as a general throughput claim. Day 19 originally exposed a real 56% two-thread regression in the first worker classifier: it treated all six scans of a materialized recursive state as scheduler overhead and ran every epoch inline. Counting those scans as work restored parallelism. Keeping one private recursive-output collection per worker and per scan then added another fan-out-specific combine cost; using the synchronized global output for this shape made the final one-, two-, and four-thread comparisons flat.

I also reformulated two recursive AoC 2024 searches with semantic keys. Day 16 retains the cheapest `(x, y, direction)` state, while Day 20 retains the first distance for each `(x, y)`. In the current two-thread regression run, Day 16 is approximately 2.9x faster with the key and the synthetic Day 20 race-track pair approximately 33x faster. Both produce the same answers and distance aggregates. The keyed terms use anti/improvement joins against recurring state, so they do not use the inner-join-only `PhysicalRecursiveCTEKeyJoin`. All new AoC 2022 and AoC 2024 entries pass the matched-main regression gate.

The recursive state and key-join machinery is physical-only. The optimizer changes reuse existing logical operators.


lnkuiper (commented):

1. Correctness: `PhysicalRecursiveCTEStateScan` only locks its own `RecursiveCTEStateScanGlobalState`, but every scan reads and finalizes aggregate states from the same `recursive_state.ht`. Two different `recurring` references therefore use different locks while `GroupedAggregateHashTable::Scan` shares the hash table's lookup scratch state, and aggregate finalizers are not necessarily read-only. The specialized key join already protects `FinalizeStates` with `recursive_state.ht_finalize_lock`, but the fallback scan does not. I can reproduce a deterministic segfault with two fallback scans and an `approx_quantile` payload at `threads=8`; the same query succeeds at `threads=1`. I left the minimal reproducer in `test/temp.test`. Could we make scans use task-local lookup state and protect aggregate finalization with the CTE-wide lock (or otherwise provide a genuinely concurrent read API)?

2. Correctness: I don't think `LogicalOperatorIsRepeatable` is safe as a blacklist. Any new operator is currently assumed to be repeatable unless it has volatile expressions or is one of the few explicitly rejected operator types. For example, `LogicalSample` falls through as repeatable, so a non-repeatable Bernoulli sample on an invariant recursive hash-build is retained after its first execution. I confirmed that such a query reports `retained_build_executions = 1`, which freezes one random sample across recursive epochs. Could we make this conservative, e.g., an allow-list or an explicit operator capability? `RecursiveSubtreeIsRepeatable` should use the same shared classifier; having a second, slightly different copy in `remove_unused_columns.cpp` makes this easier to get wrong again.

3. `RecursiveCTEState` has grown into a large bag of public state covering USING KEY storage, iteration buffers, executor caching, scheduling, and runtime metrics. This makes it hard to see which invariants are owned where and is also what leads to large functions reaching into many unrelated fields. Could we split this into focused state objects (at least a metrics object and scheduler/executor state), keep their fields private, and put the corresponding transitions on those objects? The source-side `switch (key_source_phase)`, `ExecuteRecursiveInlinePlan`, and the reset/schedule/metrics parts of `ExecuteRecursivePipelines` would then become small calls with clearer preconditions.

4. I have the same concern for `PhysicalRecursiveCTEKeyJoin` and `RecursiveCTEKeyJoinState`. The operator exposes many parallel vectors, while `ExecuteInternal` contains complete-key probing, partial-key cursoring, aggregate finalization, and result projection in one function. Could we introduce typed helpers for the key/projection layout (similar to `JoinProjectionColumns`), wrap `partial_key_index_specs` in a small type rather than using `vector<vector<idx_t>>`, and move the complete/partial probe and emit steps onto private state methods? That would remove a lot of public mutable state and make mismatched ordinals much harder to construct.

5. The USING KEY planning metadata is manually coordinated across `using_key_recursive_ctes`, `using_key_distinct_indices`, `using_key_payload_indices`, and `recursive_state_scans`, all keyed by the same `TableIndex`. Could this be one `unordered_map<TableIndex, RecursiveCTEPlanningInfo>` instead? It would give each CTE one source of truth and avoid partially initialized combinations of these containers.

6. `AggregateHTLookupState` duplicates most of the hash/probe scratch in `AggregateHTAppendState` (`hashes`, offsets, salts, addresses, comparison selections, and the group chunk), but has a separate initialization/owner contract. Could the common probing state live in a shared base/helper that append and lookup compose? In particular, keeping `owner` and `initialized` together with the scratch implementation would make it harder for callers to reuse a state with the wrong hash table.

7. The new recursive-CTE branch in `RemoveUnusedColumns::VisitOperator` now performs the fixed-point dependency analysis, projection rewrite, type rewrite, CTE-ref rewrite, and child pruning inline. Could we extract this into something like `TryPruneRecursiveCTE` with smaller helpers for computing the required columns and applying the projection? This is the long new function here; `BaseColumnPruner::ReplaceBinding` itself still looks reasonably small.

Repro:
```sql
# Regression: concurrent fallback scans of the frozen USING KEY state must synchronize

statement ok
SET threads=8;

query II
WITH RECURSIVE t(k, level, v)
USING KEY (k, max(level) AS level, approx_quantile(v, 0.5) AS v) AS (
	SELECT i, 0, i::DOUBLE
	FROM range(512) r(i)
	UNION ALL
	(
		SELECT n.k, 1, r.v
		FROM t n
		JOIN (SELECT k, v FROM recurring.t WHERE v >= 0) r USING (k)
		WHERE n.level = 0
		UNION ALL
		SELECT n.k, 1, r.v
		FROM t n
		JOIN (SELECT k, v FROM recurring.t WHERE v >= 0) r USING (k)
		WHERE n.level = 0
	)
)
SELECT count(*), sum(v) FROM t;
----
512	130816.0
```

lnkuiper:
I had Codex take another look, and it found another breaking test:
```sql
# Reused direct-probe state must rebind to each nested recursive CTE hash table

statement ok
SET threads=1;

query II rowsort
WITH RECURSIVE outer_cte(a, b) AS (
	SELECT 1, 1
	UNION
	(
		WITH RECURSIVE inner_cte(k, v) USING KEY (k) AS (
			SELECT a, 0 FROM outer_cte WHERE b < 4
			UNION ALL
			SELECT n.k, r.v + 1
			FROM inner_cte n
			JOIN recurring.inner_cte r USING (k)
			WHERE n.v < 1
		)
		SELECT k, b + 1
		FROM inner_cte, outer_cte
		WHERE v = 1 AND b < 4
	)
)
SELECT * FROM outer_cte;
----
1	1
1	2
1	3
1	4
```
Hope it helps

kryonix:
The CI failure can not be resolved. Current `main` is simply too slow to finish the `BFS` regression test.

lnkuiper:
Thanks for the changes!

I still have a few structural comments:

1. `RecursiveCTEState` is still a fairly large bag of public mutable state. Splitting out `RecursiveCTEMetrics` and `RecursiveCTESchedulerState` helps, but the hash table/index state, iteration buffers, source scan state, phase, cached chunks, locks, and flags are still all exposed together. `GetUsingKeyData` and `GetUnionData` consequently still reach into these fields and perform state transitions themselves. Could we move the source/iteration transitions and final-state draining onto `RecursiveCTEState` and make the corresponding fields private? That would make the valid phase transitions and locking requirements much easier to follow.

2. The neutral recursive schedule now keeps the inline and event execution paths in sync, which is a good improvement. However, `BuildRecursiveMetaPipelinePlan`/`BuildRecursivePipelineSchedulePlan` still reconstruct the normal `MetaPipeline` scheduling rules: finish groups, finish events, pipeline dependencies, meta-pipeline dependencies, sibling join-build ordering, and initialize-on-schedule sources. This means future changes to the generic scheduler contract also need to be mirrored here. Could we make this a single source of truth?

3. The new non-mutating recursive column-liveness analysis avoids copying and serializing the logical plan, but `RemoveUnusedColumns::AnalyzeOperator` is now a second operator-specific traversal alongside `VisitOperator`. The handling for aggregates, set operations, projections, DML, scans, materialized/recursive CTEs, and CTE references needs to remain consistent between both switches. Can we also make a single source of truth here?

kryonix:
Thanks, that's great feedback. I will work on that a bit more.

kryonix:
The CI failure seems unrelated to this PR.

lnkuiper:
Thanks for all the changes! Structure looks great now. There's one tiny correctness thing before this is ready to merge:

Combining the analysis and mutation paths into one operator switch addresses the duplicate-source-of-truth concern, but `ANALYZE` is not actually read-only. In `src/optimizer/remove_unused_columns.cpp:504`, `LOGICAL_CHUNK_GET` unconditionally calls `RemoveColumnsFromLogicalColumnDataGet`, so the first recursive liveness iteration can remove columns from a SHOW-backed chunk scan before the fixed point discovers that another column is required. The query below succeeds on base `76361ce4` but fails on head `ba398f7310` with `Invalid Error: unordered_map::at: key not found`. Disabling only `unused_columns` also makes it succeed on the PR head.

Could we guard this mutation in `ANALYZE` mode? I think it would also be worth making the mutation helpers assert that they only run in `APPLY` mode, so a new operator case cannot accidentally make the analysis destructive again.

Repro:

```sql
# name: test/temp.test
# description: Recursive column-liveness analysis must not mutate chunk scans
# group: [cte]

query I
WITH RECURSIVE t(i, payload) AS (
	SELECT 1, 'x'
	UNION ALL
	SELECT i + length(payload), columns.column_type
	FROM t
	CROSS JOIN (SHOW SELECT 42::INTEGER AS value) columns
	WHERE i < 3
)
SELECT i FROM t ORDER BY i;
----
1
2
9
```

lnkuiper:

PR #24033, opened 2026-07-21
Spill wide string payloads of aggregate states

mustafahasankhan:
`DISTINCT ON (k)` over wide VARCHAR columns compiles to a hash aggregate with `first(col)` per payload column. The string states copy each retained value into the hash table's arena allocator and keep a raw pointer. Those arenas are tracked against the memory limit but live outside the spillable page layout: they cannot be offloaded, cannot be relocated, and `Combine` chains every thread local arena's lifetime to the end of the query. With many groups and wide payloads, tracked but unreclaimable memory grows until allocation fails, while the temporary directory sits unused.

The asymmetry is easy to reproduce. The same 3.2 GB of strings (4M groups x 800 bytes) under a 1 GB memory limit:

| the same strings, in a different role | result |
|---|---|
| as group keys (`GROUP BY k, wide`) | succeeds by spilling |
| as fixed size payloads (`first(hash)`) | succeeds |
| as `first(wide)` state payloads (`DISTINCT ON`) | `Out of Memory Error` at 952.3/953.6 MiB |

This PR adds a payload heap for state owned strings: buffer managed blocks shared per hash aggregate through a registry, where only the block currently being written stays pinned. Full blocks are unpinned, so the buffer manager can offload them to temporary storage under pressure, exactly like the rest of the aggregation's intermediates. States store a `(block, offset, length)` handle instead of a pointer, and materialize under a short lived pin at finalize.

With this, the failing query above completes at the same 1 GB limit, with 2.29 GB observed moving through the temporary directory at peak.

It is also faster in memory. `Combine` copies a 16 byte handle instead of re-copying the payload through a fresh arena allocation for every group, and appends are bump allocations into 256 KiB blocks instead of per string tracked allocations with power of two rounding. On the included `first_wide_payload` benchmark, measured with alternating runs and rebuilds between states to cancel machine drift:

| benchmark | main | this PR |
|---|---|---|
| `first_wide_payload` (in memory) | 3.13 s | 0.59 s (5.3x) |
| same query at a 1 GB limit | fails | ~14 s, no cliff |

Decisions that came out of measuring rather than guessing:
- Reads cache the pin of the last read block; states finalize in insertion order, so nearly all reads hit the cache (this alone was worth 1.6x on the finalize path)
- Payloads under 32 bytes stay on the arena: they pin little memory per group and the indirection costs more than it saves (`any_value` over UUIDs regressed 33% without this and is at parity with it)
- `last()` stays on the arena: it overwrites its value for every row, which would keep appending garbage to the heap; its pathological case (few groups, many rows) is exactly the case where the arena is memory safe
- Aggregates bound with `EXPORT_STATE` stay on the arena, since exported states are read as raw bytes (`AggregateObject` now carries the export mode so the hash aggregate can tell)

Verification:
- Full fast test suite: 4828 test cases, 986230 assertions, all pass
- Aggregate suite under a `FORCE_ASSERT` build: all pass
- All 19 existing aggregate micro benchmarks A/B against main: flat within noise
- New regression test `first_wide_payload_spill.test_slow` locks the spill behavior (1M groups x 800 bytes under a 256 MB limit, verifying counts, lengths, and exact values through the spill)

Scope and follow ups: this covers `first`, `any_value`, and the sort key state used for arbitrary types. `min`, `max`, `arg_min`, `arg_max`, and `string_agg` hold state payloads the same way and can adopt the heap with the same pattern; heap block compaction would let `last()` join too. Happy to do those as follow ups.

### DIFF
--- a/benchmark/micro/aggregate/first_wide_payload.benchmark
+++ b/benchmark/micro/aggregate/first_wide_payload.benchmark
@@ -1,0 +1,17 @@
+# name: benchmark/micro/aggregate/first_wide_payload.benchmark
+# description: DISTINCT ON retaining a wide VARCHAR payload for every group
+# group: [aggregate]
+
+name First wide payload
+group aggregate
+
+load
+CREATE TABLE docs AS
+SELECT i AS k, repeat(md5(i::VARCHAR), 25) AS wide
+FROM range(0, 4000000) t(i);
+
+run
+SELECT count(wide) FROM (SELECT DISTINCT ON (k) k, wide FROM docs) s
+
+result I
+4000000
--- a/benchmark/micro/aggregate/first_wide_payload.benchmark
+++ b/benchmark/micro/aggregate/first_wide_payload.benchmark
@@ -1,17 +1,24 @@
 # name: benchmark/micro/aggregate/first_wide_payload.benchmark
-# description: DISTINCT ON retaining a wide VARCHAR payload for every group
+# description: DISTINCT ON retaining a wide VARCHAR payload for every group, spilling under a memory limit
 # group: [aggregate]
 
 name First wide payload
 group aggregate
 
+cache first_wide_payload.duckdb
+
 load
 CREATE TABLE docs AS
 SELECT i AS k, repeat(md5(i::VARCHAR), 25) AS wide
-FROM range(0, 4000000) t(i);
+FROM range(0, 8000000) t(i);
+
+init
+set temp_directory='${BENCHMARK_DIR}/first_wide_payload.duckdb.tmp';
+set memory_limit='4GB';
+set preserve_insertion_order=false;
 
 run
 SELECT count(wide) FROM (SELECT DISTINCT ON (k) k, wide FROM docs) s
 
 result I
-4000000
+8000000
--- a/src/common/row_operations/row_aggregate.cpp
+++ b/src/common/row_operations/row_aggregate.cpp
@@ -51,6 +51,10 @@ void RowOperations::UpdateStates(RowOperationsState &state, AggregateObject &agg
                                  DataChunk &payload, idx_t arg_idx, optional_ptr<const ClusteredAggr> clustered) {
 	auto count = addresses.size();
 	AggregateInputData aggr_input_data(aggr, state.allocator);
+	if (aggr.state_export_mode == AggregateStateExportMode::NONE) {
+		// Exported states are read as raw bytes, so their payloads cannot live on a payload heap
+		aggr_input_data.payload_heap = state.payload_heap;
+	}
 	auto cluster_update = aggr.function.GetStateClusterUpdateCallback();
 	aggr_input_data.clustered = cluster_update ? clustered : nullptr;
 	auto inputs = aggr.child_count ? payload.data.data() + arg_idx : nullptr;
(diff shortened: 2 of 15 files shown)


mustafahasankhan:
Pushed an update to the benchmark: it now runs with a 4GB memory limit and a temporary directory, over 8M groups of 800 byte payloads (about 6.4GB of retained state), so it measures the spilling path rather than the in-memory one. On current main this workload runs out of memory after ~10s (`failed to allocate data of size 16.0 MiB (3.7 GiB/3.7 GiB used)`), with this PR it completes in 9-13s.

mustafahasankhan:
Measured runs of `benchmark/micro/aggregate/first_wide_payload.benchmark` for reference (Apple M series, 18GB memory, default thread count).

This branch:

```
name	run	timing
benchmark/micro/aggregate/first_wide_payload.benchmark	1	13.457949
benchmark/micro/aggregate/first_wide_payload.benchmark	2	10.202831
benchmark/micro/aggregate/first_wide_payload.benchmark	3	8.813563
benchmark/micro/aggregate/first_wide_payload.benchmark	4	11.077440
benchmark/micro/aggregate/first_wide_payload.benchmark	5	9.140979
```

The same workload on current main:

```
Out of Memory Error: failed to allocate data of size 16.0 MiB (3.7 GiB/3.7 GiB used)
```

lnkuiper:
I think we should address arena-backed aggregate spilling as a whole instead of building this functionality just for `first` (also `list`/`string_agg`). This should be possible using the generic aggregate export. This is not an easy PR, and will likely require significant guidance which we may not have the capacity for.

Codex also found this issue:
```sql
# name: test/temp.test
# description: PR 24033 heap-backed combine must preserve exported first states
# group: [aggregate]

statement ok
SET threads=4;

statement ok
PRAGMA verify_parallelism;

statement ok
CREATE TABLE input AS
SELECT i % 10000 AS g, repeat(md5((i % 10000)::VARCHAR), 10) AS s
FROM range(1000000) t(i);

query I
SELECT count(*)
FROM (
	SELECT g, finalize(first(s) EXPORT_STATE) AS f
	FROM input
	GROUP BY g
) grouped
WHERE f <> repeat(md5(g::VARCHAR), 10);
----
0
```

mustafahasankhan:
Thanks for taking a look! The export issue is fixed: the update path already kept EXPORT_STATE aggregates on the arena, but parallel combines still passed the heap through. Combine and finalize now apply the same gate, and the test covers first, last and any_value under verify_parallelism.

On tackling this as a whole: agreed, and most of it is already built on top of this branch as stacked, individually tested and benchmarked conversions: arg_min/arg_max (argument payloads on the heap, by-values stay on the arena since they are compared per row), min/max (a 12 byte inline prefix decides most comparisons without touching the heap), and string_agg (backward linked chunk chains, since that value grows per row). Each fixes an out of memory failure on its workload and is flat or faster in memory. I kept them separate to keep the review pieces small, but I can fold them into this PR if you would rather review the whole problem at once.

For the remaining non-POD states, list/array_agg, quantile/median/mad and the sorted aggregate input buffering all go through the same ListSegment linked lists on the arena, so one spillable segment mechanism would cover all of them, that was my planned next step. mode and histogram allocate plain std containers per state (untracked entirely), and something export-based may indeed be the right fallback there: I considered the generic aggregate export as the primary mechanism, my hesitation is that serializing states on every combine is much heavier than moving a 12 byte handle, and the string payload family is the hot path. A hybrid (handles for string payloads, export-based serialization for the long tail) might get both.

Happy to restructure whichever way works best for your review capacity.

lnkuiper:
I appreciate the effort you put into this but the better long-term direction is a generic spill-aware aggregate-state interface, where native states are used during normal aggregation and export/serialization is invoked only at spill boundaries (which should keep import/export cost low by only incurring it once), instead of special-casing this for string aggregate functions.

mustafahasankhan:
Superseded by #24499, which addresses arena-backed aggregate state spilling as a whole through the aggregate state export, as discussed above. Thanks for the guidance!

PR #24034, opened 2026-07-21
bound fsst dictionary offset and string length in scan

jmestwa-coder:
FSST string decompression trusts the dictionary end and per-string offset/length read from the compressed segment:
- DecompressString forms baseptr + dict.end - dict_offset and reads str_len bytes with no bounds check
- the fsst-vector scan and the point-fetch path repeat the same read through FetchStringFromDict
- a corrupt dict.end/offset/length reads outside the pinned block (ASan heap-buffer-overflow in duckdb_fsst_decompress)
Bound dict.end against the block size and the offset/length against dict.end before decoding.


jmestwa-coder:
any update?

Issue #23927, opened 2026-07-19
Unexpected result when comparing `STRUCT` values in a `JOIN`


suyZhong:
### What happens?

Consider the following test case. The predicate `t0.c2 < t0.c2` evaluates to `false` for five of the six joined rows and to `NULL` for one (second query). So `WHERE NOT (t0.c2 < t0.c2)` should keep the five `false` rows and `WHERE (t0.c2 < t0.c2) IS NULL` should keep the one `NULL` row. Instead the first returns three rows and the second returns none — rows disappear from the filter that the projection says belong there.

This is a regression introduced by [`f90c3ee30e`](https://github.com/duckdb/duckdb/commit/f90c3ee30e9dddc73964842ef9a87f33e2f505c0) ("Nested Comparison Semantics", #23539): the result is correct at its parent commit. I am aware that PR changed nested comparisons to SQL NULL semantics on purpose; the problem is that the filter and the projection now disagree on the same predicate.

*Found by an automated fuzzing & triage agent; the repro is verified but the analysis may be wrong.*

### To Reproduce

```sql
CREATE TABLE t0(c2 STRUCT(a INT));
INSERT INTO t0 VALUES ({'a': 1}), ({'a': 2}), (NULL), (NULL), ({'a': NULL});

SELECT t0.c2 FROM t0 INNER JOIN t0 AS sub0 ON (t0.c2 <= sub0.c2); -- 6 rows

-- the predicate over those 6 rows: five false, one NULL
SELECT (t0.c2 < t0.c2) FROM t0 INNER JOIN t0 AS sub0 ON (t0.c2 <= sub0.c2);
-- false, false, false, false, false, NULL

SELECT t0.c2 FROM t0 INNER JOIN t0 AS sub0 ON (t0.c2 <= sub0.c2) WHERE NOT (t0.c2 < t0.c2);
-- Expected: 5 rows
-- Actual:   3 rows

SELECT t0.c2 FROM t0 INNER JOIN t0 AS sub0 ON (t0.c2 <= sub0.c2) WHERE (t0.c2 < t0.c2) IS NULL;
-- Expected: 1 row
-- Actual:   empty
```

`PRAGMA disabled_optimizers='join_order';` or `='filter_pushdown';` each restore the correct result.

### OS:

Ubuntu 22.04

### DuckDB Version:

117e1a4

### DuckDB Client:

CLI

### Full Name:

Suyang Zhong

### Affiliation:

NUS

[...]

PR #24037, opened 2026-07-21
Issue #23927: Nested Inequality NULLS

hawkfish:
* Only return NULLs for nested op NULL inequalities
* Update comparison test expecteds.


Mytherin:
Thanks!

lnkuiper:
Already an improvement, can be merged as-is. I think this might need a follow-up to fix this pre-existing issue found by Codex:
```sql
statement ok
SET nested_loop_join_threshold=0;

statement ok
CREATE TABLE nested_values(id INTEGER, value STRUCT(child STRUCT(i INTEGER)));

statement ok
INSERT INTO nested_values VALUES
	(1, {'child': {'i': 1}}),
	(2, {'child': {'i': 2}}),
	(3, {'child': NULL});

query IT
SELECT id, value <= value FROM nested_values ORDER BY id;
----
1	true
2	true
3	NULL

query I
SELECT lhs.id * 10 + rhs.id
FROM nested_values lhs
JOIN nested_values rhs ON lhs.value <= rhs.value
ORDER BY ALL;
----
11
12
22
```

I also think we should consider passing an `enum` to the `VectorOperations` instead of `true`/`false` for readability.

hawkfish:
> I also think we should consider passing an `enum` to the `VectorOperations` instead of `true`/`false` for readability.

The next PR will use ExpressionType instead, so we can fine-tune the comparison algorithm as needed. 

The codex failure may be a join algorithm discrepancy, which would need to be handled separately.

hawkfish:
PG says 3 <= 3:

```sql
hawkfish=# SELECT id, value <= value FROM nested_values ORDER BY id;
 id | ?column? 
----+----------
  1 | t
  2 | t
  3 | t
```

PR #24038, opened 2026-07-21
unbound index buffering refactor

artjomPlaunov:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9990. 

When buffering for unbound indexes was introduced, callers constructed narrow index chunks and passed a mapping from each buffered column to its physical table column. UnboundIndex::BufferChunk fixed the mapping supplied by the first caller, assuming that every subsequent caller would use the same projected layout.

This assumption actually doesn't hold for various WAL paths, i.e. it didn't consider replay of optimistic row group appends supplying all physical table columns, whereas tuple and delete paths constructed projections based on the set of all indexed columns, which can also change when an index is created during WAL replay. 

This PR changes the buffering to instead provide a chunk in physical table column layout; the chunk may be sparse, with empty or dummy vectors for non indexed columns, but column "j" consistently represents physical table column j. 

This change then derives and stores a canonical, per-index column mapping during UnboundIndex construction, using column IDs inherited from the index. BufferChunk then selects that index's columns from the table-layout input and buffers them according to this stable mapping. 

Visual description of the changes: 

Before the changes, here is how buffering worked:

```
Row-group input:
    columns [a, b, c, d]
    mapping [0, 1, 2, 3]

Resulting buffers:
    [a, b, c, d, rowid]  ->  i_b
    [a, b, c, d, rowid]  ->  i_d

Tuple/delete input:
    columns [b, d]
    mapping [1, 3]

Resulting buffers:
    [b, d, rowid]  ->  i_b
    [b, d, rowid]  ->  i_d
```
UnboundIndex retained whichever mapping it received first, although later operations could supply a different layout. The intended layout was the tuple/delete layout, but even that table-wide mapping could change when indexes were created during WAL replay.

With this PR, callers pass a chunk in physical table column layout. Every existing BufferChunk call site already has such a chunk available: in some cases it is sparse (with non indexes columns empty) or filled, but it doesn't matter.

```
Row-group input: [a, b, c, d]
Sparse input:    [., b, ., d]    "." means empty/dummy

Each unbound index selects and buffers only its own columns internally, and this mapping is guaranteed to be stable. 

i_b: mapping [1]  ->  [b, rowid]
i_d: mapping [3]  ->  [d, rowid]
```

Previously, each unbound index buffered the entire caller-provided projection, normally all indexed columns, and for row-group replay all physical columns. This duplicated unrelated column data across indexes. With this change, each index buffers only the columns it uses.

A design that deduplicates shared buffered columns globally across indexes could use less space in some cases, but coordinating index creation, deletion, and replay ordering would be substantially more complex. The per-index layout keeps ownership and replay straightforward.


artjomPlaunov:
going to wait for https://github.com/duckdb/duckdb/pull/23653 to get in first for this, as there are some related code paths that will be refactored; I would also like to do some more testing for this code path before getting this in.

Mytherin:
Thanks!

PR #24041, opened 2026-07-21
Deduplicate JSON object keys in VARIANT casts

### LINKED ISSUE #23819: JSON/VARIANT conversion yields duplicate keys
### What happens?

### To Reproduce

```sql
memory D SELECT ('{"a":1,"a":2}'::JSON::VARIANT).a;
┌──────────────────────────────────────────────────────┐
│ (CAST(CAST('{"a":1,"a":2}' AS "JSON") AS VARIANT)).a │
│                       variant                        │
├──────────────────────────────────────────────────────┤
│ 1                                                    │
└──────────────────────────────────────────────────────┘
memory D SELECT '{"a":1,"a":2}'::JSON::VARIANT::JSON;
┌──────────────────────────────────────────────────────────────────┐
│ CAST(CAST(CAST('{"a":1,"a":2}' AS "JSON") AS VARIANT) AS "JSON") │
│                               json                               │
├──────────────────────────────────────────────────────────────────┤
│ {"a":2}                                                          │
└──────────────────────────────────────────────────────────────────┘
memory D SELECT variant_keys('{"a":1,"a":2}'::JSON::VARIANT);
┌────────────────────────────────────────────────────────────────┐
│ variant_keys(CAST(CAST('{"a":1,"a":2}' AS "JSON") AS VARIANT)) │
│                           varchar[]                            │
├────────────────────────────────────────────────────────────────┤
│ [a, a]                                                         │
└────────────────────────────────────────────────────────────────┘
```
There're two issues I could spot
- For the first queries, the results are not consistent
- For the third query (variant_keys), it yields duplicate keys

### OS:

macos

### DuckDB Version:

ccf5b60ba140051e57bc0fb334cc9b6a154c4ef8

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Andrewtangtang:
## Summary

Resolves #23819 

JSON objects with duplicate keys could produce different results for field extraction, key enumeration, and JSON round-tripping after conversion to `VARIANT`. This PR uses the last value for each duplicate key during conversion so these operations have the same behavior.

## Results

```sql
WITH data AS (
    SELECT '{"a":1,"a":2}'::JSON::VARIANT AS v
)
SELECT
    (v).a AS extracted,
    variant_keys(v) AS keys,
    v::JSON AS round_trip
FROM data;
```

### Before

```text
┌───────────┬───────────┬────────────┐
│ extracted │   keys    │ round_trip │
│  variant  │ varchar[] │    json    │
├───────────┼───────────┼────────────┤
│ 1         │ [a, a]    │ {"a":2}    │
└───────────┴───────────┴────────────┘
```

### After

```text
┌───────────┬───────────┬────────────┐
│ extracted │   keys    │ round_trip │
│  variant  │ varchar[] │    json    │
├───────────┼───────────┼────────────┤
│ 2         │ [a]       │ {"a":2}    │
└───────────┴───────────┴────────────┘
```

### DIFF
--- a/src/include/duckdb/function/cast/variant/json_to_variant.hpp
+++ b/src/include/duckdb/function/cast/variant/json_to_variant.hpp
@@ -37,6 +37,10 @@ public:
 	yyjson_doc *doc = nullptr;
 };
 
+static inline string_t GetString(yyjson_val *val) {
+	return string_t(unsafe_yyjson_get_str(val), NumericCast<uint32_t>(unsafe_yyjson_get_len(val)));
+}
+
 } // namespace
 
 template <bool WRITE_DATA, bool IGNORE_NULLS>
@@ -98,14 +102,13 @@ static bool ConvertJSONObject(yyjson_val *obj, ToVariantGlobalResultData &result
 	entries.reserve(iter.max);
 
 	while (auto key = yyjson_obj_iter_next(&iter)) {
-		auto key_string = string_t(yyjson_get_str(key), NumericCast<uint32_t>(unsafe_yyjson_get_len(key)));
-		auto existing_entry = key_to_entry.find(key_string);
+		auto key_string = GetString(key);
+		auto inserted_entry = key_to_entry.emplace(key_string, entries.size());
 		auto val = yyjson_obj_iter_get_val(key);
-		if (existing_entry == key_to_entry.end()) {
-			key_to_entry.emplace(key_string, entries.size());
+		if (inserted_entry.second) {
 			entries.push_back({key, val});
 		} else {
-			entries[existing_entry->second].value = val;
+			entries[inserted_entry.first->second].value = val;
 		}
 	}
 
@@ -132,14 +135,11 @@ static bool ConvertJSONObject(yyjson_val *obj, ToVariantGlobalResultData &result
 
 	//! Iterate over all the children in the Object
 	for (const auto &entry : entries) {
-		auto key = entry.key;
-		auto key_string = yyjson_get_str(key);
-		uint32_t key_string_len = NumericCast<uint32_t>(unsafe_yyjson_get_len(key));
+		auto key_string = GetString(entry.key);
 
 		if (WRITE_DATA) {
-			auto str = string_t(key_string, key_string_len);
 			auto keys_index = start_key_index++;
-			auto dictionary_index = result.GetOrCreateIndex(str);
+			auto dictionary_index = result.GetOrCreateIndex(key_string);
 
 			//! Set the keys_index
 			variant.keys_index_data[start_child_index] = keys_index;
@@ -155,14 +155,6 @@ static bool ConvertJSONObject(yyjson_val *obj, ToVariantGlobalResultData &result
 	}
 	return true;
 }
-
-namespace {
-
-static inline string_t GetString(yyjson_val *val) {
-	return string_t(unsafe_yyjson_get_str(val), NumericCast<uint32_t>(unsafe_yyjson_get_len(val)));
-}
-
-} // namespace
 
 template <bool WRITE_DATA>
 static bool ConvertJSONPrimitive(yyjson_val *val, ToVariantGlobalResultData &result, idx_t result_index, bool is_root) {
--- a/src/include/duckdb/function/cast/variant/json_to_variant.hpp
+++ b/src/include/duckdb/function/cast/variant/json_to_variant.hpp
@@ -86,6 +86,29 @@ static bool ConvertJSONObject(yyjson_val *obj, ToVariantGlobalResultData &result
 	yyjson_obj_iter iter;
 	yyjson_obj_iter_init(obj, &iter);
 
+	struct JSONObjectEntry {
+		yyjson_val *key;
+		yyjson_val *value;
+	};
+
+	// Maps object keys to their entry index.
+	string_map_t<idx_t> key_to_entry;
+	vector<JSONObjectEntry> entries;
+	key_to_entry.reserve(iter.max);
+	entries.reserve(iter.max);
+
+	while (auto key = yyjson_obj_iter_next(&iter)) {
+		auto key_string = string_t(yyjson_get_str(key), NumericCast<uint32_t>(unsafe_yyjson_get_len(key)));
+		auto existing_entry = key_to_entry.find(key_string);
+		auto val = yyjson_obj_iter_get_val(key);
+		if (existing_entry == key_to_entry.end()) {
+			key_to_entry.emplace(key_string, entries.size());
+			entries.push_back({key, val});
+		} else {
+			entries[existing_entry->second].value = val;
+		}
+	}
+
 	auto keys_offset_data = OffsetData::GetKeys(result.offsets);
 	auto children_offset_data = OffsetData::GetChildren(result.offsets);
 	auto values_offset_data = OffsetData::GetValues(result.offsets);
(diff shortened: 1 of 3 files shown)


src/include/duckdb/function/cast/variant/json_to_variant.hpp
```diff
@@ -86,6 +86,29 @@ static bool ConvertJSONObject(yyjson_val *obj, ToVariantGlobalResultData &result
...
+	};
+
+	// Maps object keys to their entry index.
+	string_map_t<idx_t> key_to_entry;
+	vector<JSONObjectEntry> entries;
+	key_to_entry.reserve(iter.max);
+	entries.reserve(iter.max);
+
+	while (auto key = yyjson_obj_iter_next(&iter)) {
+		auto key_string = string_t(yyjson_get_str(key), NumericCast<uint32_t>(unsafe_yyjson_get_len(key)));
```
sebastiaan-dev: This can be replaced with a call to`string_t GetString(yyjson_val *val)`, following the JSON spec keys are always strings.

src/include/duckdb/function/cast/variant/json_to_variant.hpp
```diff
@@ -108,8 +131,8 @@ static bool ConvertJSONObject(yyjson_val *obj, ToVariantGlobalResultData &result
 	keys_offset_data[result_index] += count;
 
 	//! Iterate over all the children in the Object
-	yyjson_val *key, *val;
-	while ((key = yyjson_obj_iter_next(&iter))) {
+	for (const auto &entry : entries) {
+		auto key = entry.key;
 		auto key_string = yyjson_get_str(key);
 		uint32_t key_string_len = NumericCast<uint32_t>(unsafe_yyjson_get_len(key));
```
sebastiaan-dev: These calls can be replaced with a call to `GetString` as well.

src/include/duckdb/function/cast/variant/json_to_variant.hpp
```diff
@@ -86,6 +86,29 @@ static bool ConvertJSONObject(yyjson_val *obj, ToVariantGlobalResultData &result
...
+	vector<JSONObjectEntry> entries;
+	key_to_entry.reserve(iter.max);
+	entries.reserve(iter.max);
+
+	while (auto key = yyjson_obj_iter_next(&iter)) {
+		auto key_string = string_t(yyjson_get_str(key), NumericCast<uint32_t>(unsafe_yyjson_get_len(key)));
+		auto existing_entry = key_to_entry.find(key_string);
+		auto val = yyjson_obj_iter_get_val(key);
+		if (existing_entry == key_to_entry.end()) {
+			key_to_entry.emplace(key_string, entries.size());
```
sebastiaan-dev: Instead of doing both a `find` and `emplace` in the common case (no duplicate key), you can `emplace` always, which returns `pair<iterator, bool>`. The `bool` will then tell you if a new entry was inserted or it already existed. (Saving on a hash operation)

test/sql/types/variant/json_cast.test
sebastiaan-dev, line 1: Previously we could fail a checkpoint when trying to shred a `VARIANT` vector which contains duplicated keys. This PR resolves this issue, but a test covering this case would be great. Something like:

```sql
require json

load {TEST_DIR}/variant_duplicate_keys_shredding.db readwrite v1.5.0

statement ok
SET variant_minimum_shredding_size = 0;

statement ok
CREATE TABLE variants AS
SELECT '{"a":1,"a":2}'::JSON::VARIANT AS v;

restart

query I
SELECT DISTINCT vector_type(v) FROM variants;
----
SHREDDED_VECTOR

query III
SELECT (v).a, variant_keys(v), v::JSON FROM variants;
----
# output
```

test/sql/types/variant/json_cast.test
```diff
@@ -30,6 +31,29 @@ FROM src;
...
+# Duplicate object keys use the last value
+query III
+SELECT (v).a, variant_keys(v), v::JSON
+FROM (SELECT '{"a":1,"a":2}'::JSON::VARIANT AS v);
+----
+2	[a]	{"a":2}
+
+query III
+SELECT (v.outer).a, variant_keys(v, 'outer'), v::JSON
+FROM (SELECT '{"outer":{"a":1,"b":2,"a":3}}'::JSON::VARIANT AS v);
```
sebastiaan-dev, line 43: Can we add a test for different value types as well? For example, `null` and an array.

sebastiaan-dev (changes requested):
Thanks for the PR! Left some comments.

Reading the original issue I realized (with help of @Tishj ) that DuckDB currently supports duplicate keys in the `VARIANT` type, which mirrors `JSON`. However, this breaks down when we introduce shredding and will cause an error on checkpointing. When we look at the [Parquet VARIANT spec](https://parquet.apache.org/docs/file-format/types/variantencoding/) we also find "An object may not contain duplicate keys.". Given this I think we actually do not want duplicate keys in a `VARIANT` at all, and this PR moves in the right direction.

The above does imply that `variant_keys` and `json_keys` will return different results (`[a]` and `[a, a]` respectively for `{"a": 1, "a": 2}`).

Another consideration is that this PR makes the last occurring value authoritative, whereas for example `json_extract` makes the first occurrence authoritative. This means that `('{"a":1,"a":2}'::JSON::VARIANT).a` != `('{"a":1,"a":2}'::JSON).a` after this PR. I think it's best we have an explicit choice for which policy we use @Mytherin.

Andrewtangtang:
Thanks for the review! I’ve addressed the implementation and test comments above. Since the last value or first value is still undecided, would it be okay to leave the current last-value behavior unchanged for now?

src/include/duckdb/function/cast/variant/json_to_variant.hpp
```diff
@@ -132,14 +135,11 @@ static bool ConvertJSONObject(yyjson_val *obj, ToVariantGlobalResultData &result
 
 	//! Iterate over all the children in the Object
 	for (const auto &entry : entries) {
-		auto key = entry.key;
-		auto key_string = yyjson_get_str(key);
-		uint32_t key_string_len = NumericCast<uint32_t>(unsafe_yyjson_get_len(key));
+		auto key_string = GetString(entry.key);
```
sebastiaan-dev: I think `GetString` can be moved into `if (WRITE_DATA)` scope.

sebastiaan-dev (commented):
Thanks for the changes. I have one more comment otherwise it looks good to me.

As for first vs last occurrence semantics, MySQL encountered a similar situation: [bug report](https://bugs.mysql.com/bug.php?id=86866). Given their switch from first to last occurrence, last occurrence seems the way to go.

The only downside with this is that for other `JSON` processing functions to behave similarly we will have to scan the entire object for every key lookup. So what is currently the worst case for first-occurrence lookup would become the standard case for last-occurrence lookup.

src/include/duckdb/function/cast/variant/json_to_variant.hpp
lnkuiper: Once this comment is addressed this PR can be merged

Andrewtangtang: Thanks! I’ve moved GetString into the WRITE_DATA scope.

lnkuiper:
Thanks!

PR #24044, opened 2026-07-22
Prune row group for no-op replace

### LINKED ISSUE #24043: `replace` function doesn't prune row group if no-op
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE string_expr_prune AS
SELECT ('k' || lpad(i::VARCHAR, 8, '0')) AS s
FROM range(0, 300000) tbl(i)
ORDER BY i;

memory D EXPLAIN ANALYZE SELECT sum(length(s)) FROM string_expr_prune WHERE replace(s, 'z', 'z') = 'k00200000';
╭─ Summary ───────────╮
│ Total Time: 0.0052s │
╰─────────────────────╯
╭──────────────────────────────────────╮
│ Ungrouped Aggregate      1 row · 0µs │
│ Projection               1 row · 0µs │
│ Table: memory.main.string_expr_prune │
│ Projections: s                       │
│ Filters:                             │
│ "replace"(s, 'z', 'z') = 'k00200000' │
│ Row Groups Scanned: 3 / 3            │
│ 1 row                         10.0ms │
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24043


lnkuiper (changes requested):
Thanks for the PR! There is a correctness issue:
```sql
# name: test/temp_24044.test
# description: No-op replace removal must bail when its arguments cannot be folded
# group: [optimizer]

statement ok
CREATE TABLE empty_strings(s VARCHAR)

query I
SELECT count(*)
FROM empty_strings
WHERE replace(
    s,
    CAST(CAST('bad' AS INTEGER) AS VARCHAR),
    CAST(CAST('bad' AS INTEGER) AS VARCHAR)
) = 'x'
----
0
```

 Also, maybe instead of adding the rewrite rule, we could handle this fully within `replace`, i.e., when binding the function we check if those two arguments are the same, and then just return the first argument?

I'm realising now that we could remove several `Rule`s if the functions themselves return a more efficient alternative when they are bound. This would reduce the amount of rewrite rules and keep the logic centralised to the functions that it applies to. You've sent a few PRs in the past few weeks that have introduced new `Rule`s, but I think we could clean things up significantly if we move the logic to the binder.

lnkuiper:
Hey, I've thought about it and I'm fine with this being a rule. Just needs the correctness issue fixed then it's good to go

dentiny:
> Hey, I've thought about it and I'm fine with this being a rule. Just needs the correctness issue fixed then it's good to go

Thanks for the ping! Sorry I completely forgot about this PR... Updated

lnkuiper (approved):
Thanks!

PR #24046, opened 2026-07-22
Fix monotonicity annotations for floating-point numeric functions

hacoobamatata:
## Summary

This follows up on the discussion in https://github.com/duckdb/duckdb/pull/23926#pullrequestreview-4734732382, where it was suggested to update the remaining floating-point numeric functions from `StrictlyIncreasing()` to `NonDecreasing()`.

Due to floating-point rounding, adjacent `DOUBLE` values can produce equal outputs (e.g. using `nextafter`), so these functions are monotone but not strictly increasing over representable floating-point values.


Mytherin:
Thanks!

hacoobamatata:
Hi! CI has failed twice due to what appear to be infrastructure issues rather than code changes. Could someone rerun the failed jobs or requeue the PR? Thanks!

Mytherin:
Thanks!

PR #24055, opened 2026-07-22
Rewrite DISTINCT aggregates as CTEs

kryonix:
`DISTINCT` aggregates currently build and drive separate hash tables from inside the physical ungrouped and hash aggregate operators. That makes a substantial part of aggregation invisible at the logical plan level: the optimizer cannot reason about it, the profiler cannot report it as ordinary operators, and extensions that replace or inspect aggregate plans need a second path for work hidden inside the physical aggregate.

These aggregates are now rewritten into explicit logical aggregate plans instead. A distinct aggregate first groups by the original group keys and its distinct arguments, then evaluates the original aggregate without `DISTINCT`. Aggregates with the same distinct arguments and filter share that deduplication branch. Queries with multiple distinct sets, or a mix of distinct and regular aggregates, read from a shared CTE and combine the branch results with a cross product for ungrouped queries or null-safe joins on the group keys for grouped queries.

There are several semantic details behind that basic shape. Aggregate `FILTER` and `ORDER BY` expressions are preserved in the rewritten branch. Volatile group, argument, order, and filter expressions are staged once before CTE fan-out so the rewrite does not multiply their evaluation. Result bindings are remapped through a final projection, including references from result modifiers.

The distinct pass runs after the grouping-sets optimizer. I widened that optimizer so aggregates that cannot use the existing state-combining cascade are expanded into one explicit branch per grouping set over a shared CTE. The distinct pass can then rewrite each resulting single-set aggregate. This avoids depending on the physical grouping-set implementation as a semantic fallback for distinct aggregates and also keeps `GROUPING()` values explicit in the plan.

With streaming CTE fan-out, the shared input is delivered directly to all aggregate branches without materializing it first. The regression coverage checks `EXPLAIN ANALYZE` for `STREAMING_FANOUT`, two direct consumers, and zero buffered consumers. The broader tests cover grouped and ungrouped aggregates, shared and independent distinct sets, mixed regular aggregates, empty and NULL inputs, filters, aggregate ordering, volatile expressions, grouping sets and `GROUPING()`, and exported aggregate states.

I also measured the multi-distinct case that originally regressed:

```sql
SELECT COUNT(DISTINCT l_linenumber), COUNT(DISTINCT l_orderkey)
FROM lineitem;
```

On TPC-H SF100: Across two opposing-order `hyperfine` batches with two warmups and five fresh-process measurements per implementation in each batch, the rewrite had a 1.028 s mean and 1.024 s median (0.989-1.069 s), while disabling `distinct_aggregate_rewrite` had a 0.975 s mean and 0.975 s median (0.965-0.986 s). That is a 5.4% mean and 5.0% median regression for this workload. Three fresh-process RSS samples averaged 5.47 GiB for the rewrite and 5.10 GiB for the legacy path, about 372 MiB or 7.1% more. Sampling still shows aggregate hash-table lookup as the dominant work; direct CTE dispatch is not an obvious hotspot, so I have not added a branch-specific heuristic here.

I am deliberately retaining the legacy physical distinct infrastructure for optimizer-disabled execution and comparison in this PR. Removing it is a follow-up once the logical rewrite has had enough coverage across platforms and workloads; the performance and memory result above is the main remaining trade-off to investigate before making that removal.

Fix: https://github.com/duckdblabs/duckdb-internal/issues/9677

### DIFF
--- a/src/common/enum_util.cpp
+++ b/src/common/enum_util.cpp
@@ -3848,19 +3848,20 @@ const StringUtil::EnumStringLiteral *GetOptimizerTypeValues() {
 		{ static_cast<uint32_t>(OptimizerType::REMOTE_PUSHDOWN), "REMOTE_PUSHDOWN" },
 		{ static_cast<uint32_t>(OptimizerType::GROUPING_SETS), "GROUPING_SETS" },
 		{ static_cast<uint32_t>(OptimizerType::TYPE_PUSHDOWN), "TYPE_PUSHDOWN" },
-		{ static_cast<uint32_t>(OptimizerType::SCALAR_FN_PUSHDOWN), "SCALAR_FN_PUSHDOWN" }
+		{ static_cast<uint32_t>(OptimizerType::SCALAR_FN_PUSHDOWN), "SCALAR_FN_PUSHDOWN" },
+		{ static_cast<uint32_t>(OptimizerType::DISTINCT_AGGREGATE_REWRITE), "DISTINCT_AGGREGATE_REWRITE" }
 	};
 	return values;
 }
 
 template<>
 const char* EnumUtil::ToChars<OptimizerType>(OptimizerType value) {
-	return StringUtil::EnumToString(GetOptimizerTypeValues(), 43, "OptimizerType", static_cast<uint32_t>(value));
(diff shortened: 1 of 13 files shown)


Mytherin (commented):
Amazing, great that you got this to work!

Long-term this shouldn't be done at the optimizer level but we can move that code around later on when we work on removing the old distinct execution code. For now this is great.

PR #24056, opened 2026-07-22
EXTERNAL RESOURCES: SQL surface for CREATE / DESTROY / SHOW / REGISTER

carlopi:
Part 3/4 of https://github.com/duckdb/duckdb/pull/23736, adding SQL surface.

Independent code-wise from https://github.com/duckdb/duckdb/pull/24024, but only merging both will enable `SHOW ALL EXTERNAL RESOURCES` to work end-to-end.


Mytherin:
Thanks!

PR #24058, opened 2026-07-22
Define UPDATE ... FROM and MERGE behavior for multi-matched target rows

krleonid:
## Summary

When a join matches the same target row more than once, DuckDB's two UPDATE Sink paths disagreed: the in-place path applied **every** match (undefined, last-write-wins) while the del+insert path silently deduplicated. This made results and trigger transition tables unpredictable. This PR defines the behavior, matching PostgreSQL/SQLite:

- **`UPDATE … FROM` → deduplicate to one arbitrary match.** Each target row is updated at most once (first match in arrival order wins; which one is unspecified, as in PG/SQLite). A new `update_from` flag on `Logical`/`PhysicalUpdate` gates row-id dedup in the in-place path (shared `updated_rows` set, factored into `DeduplicateUpdatedRowIds`). Plain `UPDATE … WHERE` keeps its lock-free path.
- **`MERGE` → cardinality error on the second action.** A `WHEN MATCHED` UPDATE/DELETE that would affect the same target row twice raises *"cannot affect the same target row more than once"*. Detection runs on the condition-selected rows, so a `WHEN MATCHED AND …` that filters two matches down to one action does **not** error (matches PostgreSQL). `WHEN NOT MATCHED` / `WHEN NOT MATCHED BY SOURCE` are unaffected.
- **Row count fix.** The reported UPDATE count now reflects rows actually updated (distinct), not the number of join matches — also correcting a pre-existing over-count on the del+insert path.

## Compatibility
Verified SQLite locally (one arbitrary match, no error; no MERGE); PostgreSQL behavior from its docs (same for `UPDATE … FROM`; MERGE raises the cardinality violation). The split here mirrors both.

## Tests
- `test/sql/update/update_from_multi_match.test` — basic, cross-chunk (in-place & forced del+insert), RETURNING, self-join, zero-match, row-count assertions, LIST/del+insert.
- `test/sql/merge/merge_multi_match_error.test` — UPDATE/DELETE cardinality errors, conditional `WHEN MATCHED AND`, insert-only, not-matched-by-source, large unique-match.
- `test/sql/trigger/trigger_update_from_multi_match.test` — predictable OLD/NEW transition tables under multi-match, for both the in-place and **del+insert + OLD-capture** paths.
- Updated `test_update_from.test` and `returning_update.test` from the old undefined behavior to the new defined behavior.

### DIFF
--- a/src/common/enum_util.cpp
+++ b/src/common/enum_util.cpp
@@ -67,6 +67,7 @@
 #include "duckdb/common/enums/regex_match_operator_semantics.hpp"
 #include "duckdb/common/enums/relation_type.hpp"
 #include "duckdb/common/enums/row_group_append_mode.hpp"
+#include "duckdb/common/enums/row_id_handling.hpp"
 #include "duckdb/common/enums/set_operation_type.hpp"
 #include "duckdb/common/enums/set_scope.hpp"
 #include "duckdb/common/enums/set_type.hpp"
@@ -4830,6 +4831,25 @@ const char* EnumUtil::ToChars<RowGroupAppendMode>(RowGroupAppendMode value) {
 template<>
 RowGroupAppendMode EnumUtil::FromString<RowGroupAppendMode>(const char *value) {
 	return static_cast<RowGroupAppendMode>(StringUtil::StringToEnum(GetRowGroupAppendModeValues(), 3, "RowGroupAppendMode", value));
+}
+
+const StringUtil::EnumStringLiteral *GetRowIdHandlingValues() {
+	static constexpr StringUtil::EnumStringLiteral values[] {
+		{ static_cast<uint32_t>(RowIdHandling::ASSUME_UNIQUE), "ASSUME_UNIQUE" },
+		{ static_cast<uint32_t>(RowIdHandling::KEEP_FIRST), "KEEP_FIRST" },
+		{ static_cast<uint32_t>(RowIdHandling::ERROR), "ERROR" }
+	};
+	return values;
+}
+
+template<>
+const char* EnumUtil::ToChars<RowIdHandling>(RowIdHandling value) {
+	return StringUtil::EnumToString(GetRowIdHandlingValues(), 3, "RowIdHandling", static_cast<uint32_t>(value));
+}
+
+template<>
+RowIdHandling EnumUtil::FromString<RowIdHandling>(const char *value) {
+	return static_cast<RowIdHandling>(StringUtil::StringToEnum(GetRowIdHandlingValues(), 3, "RowIdHandling", value));
 }
 
 const StringUtil::EnumStringLiteral *GetSampleMethodValues() {
--- a/src/execution/operator/persistent/physical_insert.cpp
+++ b/src/execution/operator/persistent/physical_insert.cpp
@@ -5,6 +5,7 @@
 #include "duckdb/common/types/conflict_manager.hpp"
 #include "duckdb/common/vector_operations/vector_operations.hpp"
 #include "duckdb/execution/expression_executor.hpp"
+#include "duckdb/execution/row_id_deduplicator.hpp"
 #include "duckdb/execution/index/art/art.hpp"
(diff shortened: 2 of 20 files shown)


lnkuiper:
Thanks for the PR! There is one correctness issue, and two more comments:

1. The delete + insert update path can use the wrong row IDs after deduplication. For an input such as
   `[A, A, B, B]`, `DeduplicateUpdatedRowIds` selects `[A, B]` and we slice `update_chunk` and `del_row_ids`
   accordingly. However, `table.Fetch` still receives the original `row_ids` with a count of two, so it fetches
   `[A, A]` while `Delete` and `LocalAppend` operate on `[A, B]`. This can make a valid indexed `UPDATE ... FROM`
   fail with a duplicate-key error, as shown by the repro below.

   Could we use the selected row-ID vector consistently for `Fetch`, `Delete`, and append? The selected vector
   needs to be flattened before `Fetch`, since slicing produces a dictionary vector. This issue predates the PR,
   but since this PR defines duplicate-match UPDATE behavior and refactors this path, I think it should be fixed
   here.

2. Could we use the `.Values<row_t>()` iterator and extract the common row-ID registration into a small shared
   helper? `DeduplicateUpdatedRowIds` is safe as written because `Sink` currently flattens the chunk first, so this
   is not the cause of the issue above. However, UPDATE, MERGE, and `ON CONFLICT DO UPDATE` now all iterate row IDs,
   insert them into a set, and handle duplicates. A helper could do that iteration and
   optionally populate a selection vector, while the operators retain their own locking and keep-first/error
   behavior. DELETE has more specialized two-stage locking, so I would only share the low-level primitive there.

3. I think `update_from` exposes the SQL syntax where the logical and physical operators really need a duplicate
   row-ID policy. Some explicit mode is useful so ordinary UPDATEs keep the lock-free path, but could we make this
   a semantic enum such as `ASSUME_UNIQUE`, `KEEP_FIRST`, and `ERROR` instead? `UPDATE ... FROM` would use
   `KEEP_FIRST`, while MERGE should continue applying `ERROR` at the outer MERGE operator so duplicates spanning
   UPDATE and DELETE actions are detected. If an enum is too broad here, `deduplicate_row_ids` would at least be a
   clearer boolean than `update_from`.

Repro for (1.):

```sql
# name: test/temp.test
# description: UPDATE FROM RETURNING with an indexed target and duplicate matches
# group: [update]

statement ok
CREATE TABLE target (id INTEGER PRIMARY KEY, value INTEGER);

statement ok
INSERT INTO target VALUES (1, 0), (2, 0);

statement ok
CREATE TABLE source (id INTEGER, value INTEGER);

statement ok
INSERT INTO source VALUES (1, 10), (1, 11), (2, 20), (2, 21);

query I rowsort
UPDATE target
SET value = source.value
FROM source
WHERE target.id = source.id
RETURNING target.id;
----
1
2
```

krleonid:
@lnkuiper Thank you for the review. I've fixed the first issue and modified the code based on your comments.

lnkuiper:
Looks good now, thanks!

Tishj:
There's a chance this will need more work, because this is currently preventing iceberg from bumping the duckdb submodule.

This code is expecting there to be a materialized row_id column to infer row uniqueness from, this isn't necessarily true for iceberg.
Starting with V3 there is a `_row_id` column, but before that, there isn't.

Other places use `TableCatalogEntry::GetRowIdColumns` for this

Tishj:
In an attempt to fix the problem on duckdb-iceberg side, a clanker wrote up a "solution" that might be helpful as a thought experiment.

It injected another PhysicalOperator, that takes the columns that make up the "row identifiers" (filename + file_row_number for Iceberg V2 and under) and creates a `Value::TUPLE` from it.

Using a `value_map_t<row_t>` in the global state of that operator, it assigns and keeps track of row ids.

There must be a more efficient way we can do that.
Perhaps merge into needs to inject a hash expression grouping the "row identifier" columns to cleanly transform them into a unique single value.

PR #24060, opened 2026-07-22
Initial implementation of `variant_extract_string` processing function

sebastiaan-dev:
This PR resolves https://github.com/duckdblabs/duckdb-internal/issues/9986, introducing an initial implementation for `variant_extract_string`. This function takes a `VARIANT`-type object as input and a path. It returns the stringified representation of the supplied `variant`. If `path` is specified, return the stringified representation of the subtree at the given `path`. If `path` is a list, the result will be a list of stringified representations. The return type is formatted as `VARCHAR`, object keys are lexicographically ordered.

This implementation differs from `json_extract_string` in that a `VARIANT`-type has higher type resolution, which we propagate in the result.

### Examples

```sql
variant_extract_string({b: 'hello', a: 12}::VARIANT, '');
> {"a":12,"b":"hello"}

variant_extract_string({b: 'hello', a: 12}::VARIANT, 'b');
> hello

variant_extract_string('POINT (30 10)'::GEOMETRY::VARIANT, '');
>  POINT (30 10)
```

### Interface

`variant_extract_string(variant, path)`

#### Parameters

- `path` is of type `VARCHAR` or `VARCHAR[]`

### Considerations

We have expanded the `VARIANT` visitor to support visiting through `VariantNode` next to used the unified view, this introduces some duplication, but I considered it out of scope for this PR to refactor the other visitor call sites to support the `VariantNode` interface. This is thus left as future work, while supporting both approaches in the meantime.

### Follow-up work


src/function/scalar/variant/variant_extract_string.cpp
```diff
@@ -0,0 +1,367 @@
...
+};
+
+class VariantStringSerializer {
+public:
+	explicit VariantStringSerializer(VariantStringExtractLocalState &state) : tmp(state.tmp), heap(state.heap) {
+	}
+
+public:
+	//! Row result writer, modeled as part of `VariantStringSerializer` so we have easy access to function scoped state.
+	void operator()(const optional<VariantNode> &node, VectorWriter<string_t> &string_writer) const;
```
Tishj: Let's give this method a name, just so it's easier to find/debug
EDIT: Hmm, I see, this class is essentially a named `std::function`, carrying these two references as *payload* (not sure what the right terminology is)

Tishj: It's fine then, not sure how you'd do this cleaner while still being able to fit within the existing templated functor parameter

sebastiaan-dev: Yes that's what I ended up with using as I saw churn with `StringHeap` and `string` while profiling. But the interface of `VariantPathFunction::Execute` was a bit too limited to have shared state across function calls. Another option would be to inline a lambda function definition which captures this state?

sebastiaan-dev: Then you get this:

```c++
VariantPathFunction::Execute<string_t>(input, state, result, []() {
		// something something
	});
```

Tishj:
I wonder if we're not better off using yyjson to construct the json data

sebastiaan-dev:
I've changed the implementation where we now use `yyjson` in case of complex objects, and DuckDB native serialization for root primitive types. Could you take a look again @Tishj ?

src/function/scalar/variant/variant_extract_string.cpp
```diff
@@ -0,0 +1,355 @@
...
+	case VariantLogicalType::GEOMETRY: {
+		const auto value = node.GetString();
+		return AddString(Value::GEOMETRY(const_data_ptr_cast(value.GetData()), value.GetSize()).ToString());
+	}
+	case VariantLogicalType::BOOL_TRUE:
+		return heap.AddString("true", 4);
+	case VariantLogicalType::BOOL_FALSE:
+		return heap.AddString("false", 5);
+	case VariantLogicalType::INT8:
+		return StringCast::Operation(node.GetData<int8_t>(), heap);
```
Mytherin: Do we need this function to contain all of this string conversion logic? Can't we have this function emit a `VARIANT` and then cast to `VARCHAR` (or to `JSON`) to directly use the existing logic?

Mytherin: I kind of wonder to what extend this function should be different from "just" being `variant_extract(..., <path>)::VARCHAR`

sebastiaan-dev: There are a couple differences between the output of `variant_extract()::VARCHAR` and what the output of a `VARIANT` equivalent `json_extract_string` produces:

1. `variant_extract` with `path=''` will return `NULL`, whereas the full object would be returned in the case of `extract_string`
2. We have to distinguish between primitive values, and values which are part of an array/object. For the latter we have to add quotations to stringified values
3. `variant_extract` produces output with different formatting: `[1, 2, NULL]` instead of `[1,2,null]`
4. `variant_extract` does not support a list of paths
5. `variant_extract` formats the `double` exponent differently `-1.7976931348623157e+308` vs `yyjson` formatting of `-1.7976931348623157e308`

When doing `variant_extract()::JSON::VARCHAR` we fix 2,3,5 but introduce another issue where primitive strings will now be quoted (so `"hello"` would be returned where we should get `hello`).

sebastiaan-dev: Right now there are essentially 2 formatting paths in this PR, one using DuckDB utility functions and one using `yyjson`. I did this because for primitive types it is faster to use the utility functions instead of constructing a `yyjson` object. But it's not a requirement to make the function work correctly (at least not all of them).

Mytherin: I just feel that `variant_extract_string` is not the place to implement this kind of string formatting logic again - given the fact that the variant -> JSON -> string cast seems very close perhaps we can utilize that here instead?

sebastiaan-dev: I've now implemented it using `VariantCasts::ConvertVariantToJSON`, if you can take a look again?

Mytherin:
Thanks!

PR #24061, opened 2026-07-22
Add SetVal

maiadegraaf:
This PR adds a new sequence function `SETVAL`, which manipulates a sequence by directly setting its next value.  The implementation follows that of [postgres](https://www.postgresql.org/docs/current/functions-sequence.html) meaning it sets the sequence's `last_value` field to the specified value and sets its `is_called` field to `true`

It takes:
- `sequence_name`
- `value`: The new value you want to set
- `is_called`: An optional argument that can be set to either `true` or `false`.  When set to `false`, `nextval` will return the set value.

Eg:
```sql
# Next nextval will return 43
SELECT setval('myseq', 42);

# Same as above           
SELECT setval('myseq', 42, true);    

# Next nextval will return 42
SELECT setval('myseq', 42, false);
```


test/sql/catalog/sequence/setval.test
```diff
@@ -0,0 +1,391 @@
+# name: test/sql/catalog/sequence/setval.test
+# description: Test setval() function
+# group: [sequence]
+
+require skip_reload
+
+statement ok
+CREATE SEQUENCE myseq;
```
Tishj, line 8: Can we add tests with `INCREMENT BY` ?
I wonder how this should behave if the SETVAL is a value that doesn't fit the increment, and also what should happen for the next value in that case.

Also `MAXVALUE` interactions with SETVAL are interesting to look at

maiadegraaf: Good question :)
`SETVALUE` only validates `MIN/MAX`, but not `INCREMENT BY` so it just shifts the anchor. This is also what postgres does:

```sql
postgres=# CREATE SEQUENCE myseq INCREMENT BY 2;
CREATE SEQUENCE
postgres=# SELECT setval('myseq', 3, false);
 setval 
--------
      3
(1 row)

postgres=# SELECT nextval('myseq');
 nextval 
---------
       3
(1 row)

postgres=# 
postgres=# SELECT nextval('myseq');
 nextval 
---------
       5
(1 row)
``` 

To be safe I've added some tests for `INCREMENT BY`, and some more with `MIN/MAX` constraints and `CYCLE`.

Mytherin:
Thanks!

PR #24062, opened 2026-07-22
Optionally spill evicted external file cache blocks to the temporary directory

mustafahasankhan:
### Problem

The external file cache keeps its blocks in destroy-on-eviction buffers. Under memory pressure they are dropped, and the next read of the same range goes back to the source. For remote files that means re-downloading data that was already fetched, every time the working set exceeds the memory limit.

With a 121MB remote Parquet file and `memory_limit='100MB'`, 2 threads: the first read takes ~14-22s (network), and every further read takes ~13s, because the evicted ~36MB is re-fetched from the source each time.

### Change

With `SET external_file_cache_spill=true`, cache blocks are written to the temporary directory when they are evicted, like other data that no longer fits the memory limit, and later reads load them back from there. Nothing is written ahead of time: blocks only move to disk when the buffer pool evicts them, and they move back (and the temporary file is freed) when they are read again.

- new buffer mode `DestroyBufferUpon::SPILL_FAILURE`: prefer writing the buffer to a temporary file on eviction, destroy it only if that is not possible. Since cached data can always be re-fetched, a block whose write fails (no temporary directory, or `max_temp_directory_size` reached) is dropped rather than failing the query
- blocks below the block allocation size keep the old behavior, spilling them would create a temporary file per block (local cache blocks default to 16KB)
- the block re-index that runs when the cache block size changes now skips offloaded blocks; pinning all of them at once to copy them could exceed the memory limit

The setting defaults to false, so this is a no-op unless enabled. Spilled cache blocks share the temporary directory with operator spilling but are not freed between queries, so cached data could crowd out sorts and joins that must spill. Turning it on by default would need cache blocks to be dropped from the temporary directory when other consumers need the space; happy to look into that as a follow-up if there is interest.

### Results

Same workload as above, identical checksums on every read:

| | main | this PR (setting on) |
|---|---|---|
| cold read | 18.5s | 13.9-21.8s (network variance) |
| re-read after eviction | 13.0-13.4s (re-fetches from source) | 0.51-0.64s (reads back from the temporary directory) |
| `duckdb_memory()` temporary storage for `EXTERNAL_FILE_CACHE` | 0 | ~37MB |

The added test covers: default off is a no-op, spilling kicks in when enabled, reads reassemble byte for byte through the spill, a full temporary directory falls back to dropping, and blocks below the block allocation size are never spilled.

A note from testing: `duckdb_temporary_files()` does not list the variable-size `.block` files (only the managed `.tmp` containers); `duckdb_memory().temporary_storage_bytes` does account for them.


src/include/duckdb/common/enums/destroy_buffer_upon.hpp
```diff
@@ -13,9 +13,10 @@
 namespace duckdb {
 
 enum class DestroyBufferUpon : uint8_t {
-	BLOCK = 0,    //! Destroy the data buffer upon destroying the associated BlockHandle (block can be evicted)
-	EVICTION = 1, //! Destroy the data buffer upon eviction to storage (destroy instead of evict)
-	UNPIN = 2     //! Destroy the data buffer upon unpin (destroyed immediately, not added to eviction queue)
+	BLOCK = 0,        //! Destroy the data buffer upon destroying the associated BlockHandle (block can be evicted)
+	EVICTION = 1,     //! Destroy the data buffer upon eviction to storage (destroy instead of evict)
+	UNPIN = 2,        //! Destroy the data buffer upon unpin (destroyed immediately, not added to eviction queue)
+	SPILL_FAILURE = 3 //! Evict to storage like BLOCK, but destroy the data buffer if it cannot be written there
```
dentiny: I'm not sure it's a good naming, especially "FAILURE".

test/sql/storage/external_file_cache/external_file_cache_spill.test_slow
```diff
@@ -0,0 +1,82 @@
...
+FROM read_parquet('__TEST_DIR__/efc_spill.parquet');
+----
+0	0
+
+# blocks below the block allocation size are never spilled: with tiny cache blocks
+# the cache drops them under pressure, and reads still work
+statement ok
+SET external_file_cache_local_block_size=16384;
+
+query I
```
dentiny, line 66: could you please add a few more test cases on other cache situations?
- list external file cache, I'm wondering if we should display cache status as well (in-memory vs spilled)
- disable external file cache, spilled cache block should be deleted
- cache block size change, we should be able to read the correct content after size change

src/common/enum_util.cpp
```diff
@@ -1847,19 +1847,20 @@ const StringUtil::EnumStringLiteral *GetDestroyBufferUponValues() {
 	static constexpr StringUtil::EnumStringLiteral values[] {
 		{ static_cast<uint32_t>(DestroyBufferUpon::BLOCK), "BLOCK" },
 		{ static_cast<uint32_t>(DestroyBufferUpon::EVICTION), "EVICTION" },
-		{ static_cast<uint32_t>(DestroyBufferUpon::UNPIN), "UNPIN" }
+		{ static_cast<uint32_t>(DestroyBufferUpon::UNPIN), "UNPIN" },
```
dentiny: I'm thinking whether `DestroyBufferUpon` is a good naming, from its naming, the buffer is destructs and should disappear. I'd prefer rename to sth else like `BufferEvictionPolicy`.

dentiny (commented):
I like the idea

src/include/duckdb/common/enums/destroy_buffer_upon.hpp
mustafahasankhan: Fair point, renamed to `EVICTION_UNLESS_SPILLED`, which reads as a sentence in the existing scheme: destroy the buffer upon eviction, unless it was spilled.

test/sql/storage/external_file_cache/external_file_cache_spill.test_slow
mustafahasankhan: Added all three, plus a `spilled` column to `duckdb_external_file_cache()` so the listing distinguishes in-memory, spilled and dropped blocks directly. The new cases cover the listing under pressure (`NOT loaded AND spilled`, and never `loaded AND spilled`), that `SET enable_external_file_cache=false` deletes the spilled blocks (listing empties and the temporary storage for the tag returns to zero), and that a block size change re-indexes correctly: the spilled blocks are dropped, reads reassemble the right content afterwards, and their temporary storage is freed.

src/common/enum_util.cpp
mustafahasankhan: I agree the enum name itself is debatable, but it predates this PR and renaming it touches call sites all over the buffer manager, so I kept this change to the new value only. Happy to do a `DestroyBufferUpon` to `BufferEvictionPolicy` rename as a small separate PR if the maintainers like it.

dentiny (commented):
External file cache is made for local file as well, the purpose of which is a more controlled version of page cache (I have some plans for direct IO), so IMO local file shouldn't be spilled.

dentiny:
CC @lnkuiper 

Some thoughts on the relationship with persistent cache idea, which explicitly dumps on-disk cache, meanwhile use in-memory cache blocks to act as page cache.
Both could be used to address the main memory constraint, but persistent caching could be more useful in general, spilling to temporary directory is a much simpler solution though.
- Memory is still used as a on-disk cache, so no performance hurt
- Persistent cache directories could and should be shared among multiple DuckDB instances, which benefits multi-process use case, like pg_mooncake/pg_duckdb/etc

dentiny (commented):
Hi sorry for the delay, another request is `DestroyBufferUpon::BLOCK` already represents “do not destroy this buffer on eviction; write it to temporary storage so it can be loaded again later”. So external file cache spilling does not need a new enum value. Spill-enabled cache blocks can simply be allocated with `can_destroy = false`, which makes the existing buffer manager path use `DestroyBufferUpon::BLOCK`.

I made a temporary example commit here: https://github.com/dentiny/duckdb/pull/156/changes/92531cc035a9ce81f498fbcf6e77c7fd5dcf8a2c

mustafahasankhan:
Done, thanks for the example commit, I adopted its shape: spill-enabled cache blocks are now plain `can_destroy = false` allocations through the existing `DestroyBufferUpon::BLOCK` path, and whether a buffer may be dropped when the temporary write fails is derived from its memory tag, since cached external file data can always be re-read from its source. No new enum value. Behavior is unchanged: no temporary directory or a full one still drops cache blocks instead of failing the query, and the spill test passes unchanged.

dentiny (commented):
A few comments:
- I guess we could further simplify the code? example change: https://github.com/dentiny/duckdb/pull/158/changes/b55d462c121052efbd9df022291ceac5cc361a4a
- Could you please handle local filesystem case? https://github.com/duckdb/duckdb/pull/24062#pullrequestreview-4759938107
- Could you please run formatter?

mustafahasankhan:
Done, all three:
- Adopted your example, the buffer manager diff is now zero. I kept one extra condition in `AllocateCacheBuffer`: no spilling when there is no temporary directory, otherwise enabling the setting on such an instance would make cache blocks unevictable. Happy to drop it if you'd rather not have it.
- Local files no longer spill, only remote ones (`FileSystem::IsRemoteFile` in the allocation decision). The test now asserts that a fully opted-in local file still just drops its blocks. Re-verified the remote path against a real URL: evicted blocks show up as `spilled` in `duckdb_external_file_cache()`, re-reads come back in ~0.5s instead of ~25s cold, same checksums.
- Formatter run, and the test uses `{TEST_DIR}` now.

dentiny:
hmm seems there's formatting issue
```diff
 --- a/src/main/config.cpp
  +++ b/src/main/config.cpp
  @@ -251,12 +251,12 @@ static const ConfigurationOption internal_options[] = {
   
   static const ConfigurationAlias setting_aliases[] = {DUCKDB_SETTING_ALIAS("configure_metrics", 30),
                                                        DUCKDB_SETTING_ALIAS("custom_profiling_settings", 30),
  -                                                     DUCKDB_SETTING_ALIAS("memory_limit", 129),
  +                                                     DUCKDB_SETTING_ALIAS("memory_limit", 130),
                                                        DUCKDB_SETTING_ALIAS("null_order", 61),
  -                                                     DUCKDB_SETTING_ALIAS("profile_output", 152),
  -                                                     DUCKDB_SETTING_ALIAS("user", 171),
  +                                                     DUCKDB_SETTING_ALIAS("profile_output", 153),
  +                                                     DUCKDB_SETTING_ALIAS("user", 172),
                                                        DUCKDB_SETTING_ALIAS("wal_autocheckpoint", 29),
  -                                                     DUCKDB_SETTING_ALIAS("worker_threads", 169),
  +                                                     DUCKDB_SETTING_ALIAS("worker_threads", 170),
                                                        FINAL_ALIAS};
```

src/storage/external_file_cache/external_file_cache.cpp
```diff
@@ -330,6 +338,13 @@ BufferManager &ExternalFileCache::GetBufferManager() const {
 	return buffer_manager;
 }
 
+BufferHandle ExternalFileCache::AllocateCacheBuffer(BufferManager &buffer_manager, const string &path, idx_t nr_bytes) {
+	const bool can_destroy = !FileSystem::IsRemoteFile(path) || nr_bytes < buffer_manager.GetBlockAllocSize() ||
```
dentiny: NOT operator adds additional mind overhead, can we try to avoid? For example,
```cpp
const bool spill = Settings::Get<ExternalFileCacheSpillSetting>(db) &&
                   FileSystem::IsRemoteFile(path) &&
                   buffer_manager.HasTemporaryDirectory() &&
                   nr_bytes >= buffer_manager.GetBlockAllocSize();
return buffer_manager.Allocate(MemoryTag::EXTERNAL_FILE_CACHE, nr_bytes, /*can_destroy=*/!spill);
```

test/sql/storage/external_file_cache/external_file_cache_spill.test_slow
```diff
@@ -0,0 +1,81 @@
...
+statement ok
+SET temp_directory='{TEST_DIR}/external_file_cache_spill';
+
+statement ok
+SET preserve_insertion_order=false;
+
+# roughly 290MB of incompressible data, well above the memory limit
+statement ok
+COPY (SELECT hash(i) a, hash(i * 2) b, hash(i * 3) c FROM range(12000000) t(i))
+TO '{TEST_DIR}/efc_spill.parquet' (FORMAT parquet);
```
dentiny, line 30: Now we disable spilling for local files, why do we use local parquet for testing? I think we should be using remote ones to check spill?

dentiny: I found we do have unit test relying on httpfs
https://github.com/duckdb/duckdb/blob/514669fced0201e759543eb933421da3837c14d0/test/sql/json/table/read_json.test#L402-L417

mustafahasankhan: - added a remote section over httpfs using city_temperature.parquet from duckdb-data, same host the alp/alprd compression tests already fetch from
- kept the local half above the require httpfs so it still runs on builds without the extension
- one thing found: the spilled column only flips after eviction, and eviction under scan pressure is not deterministic for a small remote block (the pool prefers dropping can_destroy buffers first), so the test forces eviction by shrinking memory_limit to 1MB, checks spilled and temp usage, restores the limit, and re-reads the same content back from temp

dentiny:
Hi @mustafahasankhan could you please undraft the PR to trigger CI, thanks!

dentiny (approved):
Thanks! CC @lnkuiper

PR #24065, opened 2026-07-23
Auto-rollback failed implicitly-wrapped multi-statements on all paths

jraymakers:
The StatementPreprocessor wraps multi-statement bodies (e.g. a PIVOT inside a CTAS) in an implicit BEGIN/COMMIT with auto_rollback set, so a failure rolls the whole thing back and leaves the connection usable. That was only honored in ClientContext::Query(); statements executed via the pending API (as client drivers such as the Python and Node bindings do) skipped it, leaving the connection stuck in an aborted transaction until a manual ROLLBACK or reconnect.

Honor auto_rollback in EndQueryInternal, the shared cleanup path all execution routes funnel through, so Query, prepared Execute, and a directly-driven PendingQuery all recover.

### DIFF
--- a/src/main/client_context.cpp
+++ b/src/main/client_context.cpp
@@ -276,7 +276,11 @@ ErrorData ClientContext::EndQueryInternal(ClientContextLock &lock, bool success,
 				}
 			} else if (invalidate_transaction) {
 				D_ASSERT(!success);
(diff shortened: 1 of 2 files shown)


jraymakers:
Refreshing PR.

Mytherin:
Thanks!

PR #24075, opened 2026-07-23
Optimize `regexp_full_match` with literal patterns into equality checks

### LINKED ISSUE #24042: Full string match regex function doesn't prune row group
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE OR REPLACE TABLE r3 AS
         SELECT
             i::INTEGER AS id,
             ('k' || lpad(i::VARCHAR, 8, '0')) AS s
         FROM range(0, 1000000) t(i)
         ORDER BY id;

-- should scan only one row group
memory D EXPLAIN ANALYZE SELECT count(*) FROM r3 WHERE replace(s, 'z', 'z') = 'k00900000';
╭─ Summary ───────────╮
│ Total Time: 0.0107s │
╰─────────────────────╯
│ Table: memory.main.r3                │
│ Filters:                             │
│ "replace"(s, 'z', 'z') = 'k00900000' │
│ Row Groups Scanned: 9 / 9            │
│ 1 row                         50.0ms │

-- should scan only one row group
memory D EXPLAIN ANALYZE SELECT count(*) FROM r3 WHERE regexp_full_match(s, 'k00900000');
╭─ Summary ───────────╮
│ Total Time: 0.0101s │
╰─────────────────────╯
│ Table: memory.main.r3             │
│ regexp_full_match(s, 'k00900000') │
│ Row Groups Scanned: 9 / 9         │
│ 1 row                      40.0ms │
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

tempewda:
Fixes: #24042

regexp_full_match(col, pattern) doesn't get row group pruning when the pattern is a literal string with no regex metacharacters. Since a full match against a literal is equivalent to col = 'pattern', we can rewrite it to an equality check and get pruning through the existing statistics path.

This extends the existing regexp_matches optimizer rule in regex_optimizations.cpp to also handle regexp_full_match. If RE2's parsed AST indicates the pattern is a literal (kRegexpLiteralString, kRegexpLiteral, or kRegexpEmptyMatch), we rewrite to COMPARE_EQUAL. Non-literal patterns are left unchanged.

Tests added:
- `test/sql/function/string/regex_search.test`: literal patterns (plain, with `%`/`_`, empty string, escaped metacharacter, case insensitive flag) alongside non-literal patterns that should be left alone.
- `test/sql/optimizer/string_row_group_pruning.test`: confirms row group pruning actually kicks in for a literal `regexp_full_match`.


src/optimizer/rule/regex_optimizations.cpp
```diff
@@ -176,6 +177,29 @@ unique_ptr<Expression> RegexOptimizationRule::Apply(LogicalOperator &op, vector<
...
+					return nullptr;
+				}
+				exact_string = escaped_like_string.like_string;
+			}
+			auto parameter = make_uniq<BoundConstantExpression>(Value(std::move(exact_string)));
+			return BoundComparisonExpression::Create(ExpressionType::COMPARE_EQUAL,
+			                                         std::move(root.GetChildrenMutable()[0]), std::move(parameter));
+		}
+		// if it's a full match but not a literal, we can't optimize it into a contains/like
+		return nullptr;
```
Mytherin: Can this be integrated with the code below that creates a `LikeString` from a regexp, instead of adding this new branch?

Mytherin (commented):
Thanks for the PR - looks good, one comment

tempewda:
@Mytherin Thanks for the feedback!

I've refactored the rule to reuse the original string extraction logic entirely, and simply fork at the very end to build either an `EQUALITY` or `CONTAiNS` AST node depending on the function name.

Mytherin:
Looks good but there's some merge conflicts - could you resolve them?

tempewda:
@Mytherin done :)

CI failures seem unrelated to this PR, likely stemming from #24146

Issue #23676, opened 2026-07-08
top_n_window_elimination can rank NaN before finite values


Yibo-Dong:
### What happens?

`ROW_NUMBER() OVER (ORDER BY x ASC NULLS LAST) <= 2` should pick the two smallest finite values from `[-1.0, 0.0, NaN]`. With `top_n_window_elimination` enabled, DuckDB instead includes the `NaN` row and drops `0.0`.

DuckDB's numeric type documentation explicitly states that `NaN` compares greater than any other floating-point number: <https://duckdb.org/docs/current/sql/data_types/numeric.html#floating-point-types>. Therefore, for `ORDER BY x ASC`, finite values should come before `NaN`.

This happens on current main (a1aacc73df) and also on v1.5.4.

### To Reproduce

To reproduce:

```sql
WITH src(id, x) AS (
  VALUES
    (1, 0.0::DOUBLE),
    (2, 'NaN'::DOUBLE),
    (3, -1.0::DOUBLE)
)
SELECT id, x
FROM src
QUALIFY ROW_NUMBER() OVER (ORDER BY x ASC NULLS LAST) <= 2
ORDER BY id;
```

Run:

```sh
./build/release/duckdb -f bug.sql
```

Actual:

stdout:
```text
| id | x |
|---:|---:|
| 2 | nan |
| 3 | -1.0 |

```

Expected:

```
| id | x |
|---:|---:|
| 1 | 0.0 |
| 3 | -1.0 |
```

Notes From Minimization:

- `SET disabled_optimizers='top_n_window_elimination'` returns the window-sort result, so the reduced trigger is the top-n rewrite.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (a1aacc73df)

### DuckDB Client:

CLI

### Full Name:

Yibo Dong

### Affiliation:

National University of Singapore

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

DinosL:
Should be resolved in https://github.com/duckdb/duckdb/pull/24080

Yibo-Dong:
Thanks for checking this!

Issue #23675, opened 2026-07-08
top_n_window_elimination can produce a NULL row for ROW_NUMBER on empty input


Yibo-Dong:
### What happens?

top_n_window_elimination can produce a NULL row for `ROW_NUMBER()` on empty input. 

This happens on current main (a1aacc73df) and also on v1.5.4 .

### To Reproduce

To reproduce:

```sql
CREATE TABLE e(s INTEGER, p INTEGER);

SELECT p
FROM e
QUALIFY row_number() OVER (ORDER BY s) <= 1;
```

Run:

```sh
./build/release/duckdb -f bug.sql
```

Actual:

On current main (a1aacc73df), running the minimized repro gives:

stdout:
| p |
| --- |
| NULL |

Expected:

_No rows._

Notes From Minimization:

- `SET disabled_optimizers='top_n_window_elimination'` removes the wrong NULL row, so the reduced trigger is the top-n rewrite.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (a1aacc73df)

### DuckDB Client:

CLI

### Full Name:

Yibo Dong

### Affiliation:

National University of Singapore

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

DinosL:
Should be resolved in https://github.com/duckdb/duckdb/pull/24080

Yibo-Dong:
Thanks for checking this!

PR #24080, opened 2026-07-23
TopNWindowElimination Fixes

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/10038

Fix: Track and rewrite column bindings by their actual column binding identities instead of assuming that projected column positions match the underlying scan.

fixes https://github.com/duckdblabs/duckdb-internal/issues/9902
fixes https://github.com/duckdb/duckdb/issues/23675

Fix: Prevent `NULL` result from `arg_[min/max]` aggregate on empty input by introducing a constant group by. This impacts performance a bit but is way simpler than adding a `count_star` aggregate and a filter on that. I resorted to the simpler solution here as column bindings are hard enough in this optimizer.

fixes https://github.com/duckdblabs/duckdb-internal/issues/9903
fixes https://github.com/duckdb/duckdb/issues/23676

Fix: Use `GreaterThan::Operation` in null-preserving `arg_[min/max]` rather than `>` comparator.

### DIFF
--- a/src/include/duckdb/function/aggregate/minmax_n_helpers.hpp
+++ b/src/include/duckdb/function/aggregate/minmax_n_helpers.hpp
@@ -3,6 +3,7 @@
 #include "duckdb/common/common.hpp"
 #include "duckdb/storage/arena_allocator.hpp"
 #include "duckdb/common/algorithm.hpp"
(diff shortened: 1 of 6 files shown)


lnkuiper:
Thanks!

PR #24083, opened 2026-07-23
Fix warnings and support alpha release version in extensions

smvv:
DuckDB core extensions are [built](https://github.com/duckdb/duckdb/actions/runs/29970384532/job/89093957512#step:28:48) with:
```
-- Found Git: /usr/local/bin/git (found version "2.55.0")
-- git hash c97bd8b96e, version v1.6.0-dev11186, extension folder c97bd8b96e
```
so the extensions use `c97bd8b96e` instead of `v2.0.0-alpha35494`.

This also means the extension is uploaded to an s3 path with the commit hash instead of the alpha version.

So, the CI step `Check staged extensions install/load` [fails](https://github.com/duckdb/duckdb/actions/runs/29970384532/job/89118133238#step:4:12) with:
```
IO Error: Failed to download extension "httpfs" at URL "http://extensions.duckdb.org/v2.0.0-alpha35494/linux_amd64/httpfs.duckdb_extension.gz"
```
because the extension is not built/uploaded correctly.

CMake can use the "fake" git tag in the extension CI job if it matches `VERSIONING_TAG_MATCH` (`v*.*.0`), but that cannot match the alpha version suffix.

So, this PR adds support for release version suffices in `CMakeLists.txt` and `scripts/package_build.py`

Note: I think we should merge/simplify this versioning logic because it seems like a very complex way of handling version strings when the version string is already set in the first CI job.


.github/actions/ccache-action/action.yml
```diff
@@ -14,7 +14,7 @@ runs:
   steps:
      - name: Setup Ccache
        if: ${{ github.workflow != 'NightlyTests' }}
-       uses: hendrikmuhs/ccache-action@main
+       uses: hendrikmuhs/ccache-action@v1.2.22
```
smvv, line 17: note: fixed two nodejs 20 warnings in CI.

PR #24084, opened 2026-07-23
Add batch index support to PhysicalColumnDataScan

kryonix:
This PR adds batch-index support to `PhysicalColumnDataScan` and propagates those indexes through materialized CTE execution. The motivating case is scanning a retained `ColumnDataCollection` while preserving insertion order. Without batch indexes, the scan has to run sequentially:

```sql
CREATE TABLE T AS FROM range(1000000000);

SELECT random() FROM T; -- 0.623s
SELECT random() FROM _; -- 5.905s

SET preserve_insertion_order=false;
SELECT random() FROM _; -- 0.643s
```

Assigning one batch index per emitted vector would recover parallelism, but would also create 2048-row batches. Batch-oriented consumers such as table and Parquet writes would then store and merge those small batches, introducing another full data-copy pass. Instead, I group existing collection chunks into scan tasks using the downstream operator's preferred batch size. The default is the storage row-group size. This keeps the scan parallel while producing batches that can be consumed directly by table and Parquet writers.

## Implementation

`OperatorPartitionInfo` can now carry a preferred batch size. A `PhysicalColumnDataScan` that is asked for batch indexes claims coarse scan tasks, assigns each task a monotonically increasing index, and keeps all chunks from that task on the same local scan state. Scans that do not request batch indexes retain the existing vector-sized task granularity. The scan also supports projected collection columns so column pruning does not invalidate the physical scan.

Materialized CTE planning now records whether the CTE body preserves order and whether its sources can provide batch indexes. Ordered materialization uses `BatchedDataCollection`, while the streaming CTE exchange carries batch metadata through both its in-memory chunk pool and shared spool. Producer-global indexes are normalized into a dense exchange-local namespace, including across ordered multi-pipeline producers.

For a compatible consumer pipeline, the exchange can propagate batch indexes directly into an externally fed `PipelineExecutor`. This avoids buffering and copying the CTE result before batch collectors, table inserts, and Parquet writes. The executor initializes, advances, and finalizes the downstream batch state using the same global batch-index namespace as regular source-driven pipelines.

Streaming result tasks are identified explicitly when they are fed through a direct CTE exchange. This replaces the previous assumption that a blocked result collector always belongs to the last incomplete pipeline, which is not true when the CTE producer and consumer pipelines execute concurrently.

`LogicalColumnDataGet` serializes projected column IDs as an optional property. Older serialized plans omit this property and continue to deserialize as a scan of all collection columns.

## Fallbacks and limitations

The new paths are deliberately conservative:

- A batch-indexed direct consumer is used only when exactly one pipeline produces the CTE. Ordered multi-producer inputs, such as `UNION ALL`, use the buffered exchange because their local batch namespaces cannot yet be forwarded directly without collisions.
- Ordered buffered exchange scans remain single-threaded. Parallel chunk-level scans could split one logical batch across downstream local sink states, causing duplicate batch ownership and additional merging. Parallel batch-preserving execution uses the direct exchange path.
- Direct exchange remains limited to sinks and intermediate operators that explicitly support external input. Order-dependent sinks, fixed-order or non-parallel operators, and partition-column metadata fall back to buffered or materialized execution.
- Recursive CTE execution continues to use the existing non-exchange fallback.
- If consumers request conflicting preferred batch sizes, the CTE uses the default row-group-sized batches.

These cases affect execution strategy only; they do not change query results. Unordered and non-batch-indexed `ColumnDataCollection` scans retain their previous behavior.

## Performance

On a matched release build with 100 million rows and eight threads:

| Query | Main | This branch |
| --- | ---: | ---: |
| Ordered retained collection scan | 0.550s | 0.074s |
| Direct table scan | - | 0.075s |
| CTAS from retained collection | 0.470s | 0.072s |
| Unordered retained collection scan | - | 0.079s |

For a 20-million-row reldebug CTAS, the direct CTE path had a 67ms median versus 62ms for a plain table CTAS.

Fix: https://github.com/duckdblabs/duckdb-internal/issues/9019

### DIFF
--- a/src/common/enum_util.cpp
+++ b/src/common/enum_util.cpp
@@ -4412,6 +4412,25 @@ PipelineBroadcastExchangeConsumerMode EnumUtil::FromString<PipelineBroadcastExch
 	return static_cast<PipelineBroadcastExchangeConsumerMode>(StringUtil::StringToEnum(GetPipelineBroadcastExchangeConsumerModeValues(), 4, "PipelineBroadcastExchangeConsumerMode", value));
 }
 
+const StringUtil::EnumStringLiteral *GetPipelineBroadcastExchangeOrderModeValues() {
+	static constexpr StringUtil::EnumStringLiteral values[] {
+		{ static_cast<uint32_t>(PipelineBroadcastExchangeOrderMode::UNORDERED), "UNORDERED" },
+		{ static_cast<uint32_t>(PipelineBroadcastExchangeOrderMode::SEQUENTIAL), "SEQUENTIAL" },
+		{ static_cast<uint32_t>(PipelineBroadcastExchangeOrderMode::BATCH_INDEX), "BATCH_INDEX" }
+	};
+	return values;
+}
+
+template<>
+const char* EnumUtil::ToChars<PipelineBroadcastExchangeOrderMode>(PipelineBroadcastExchangeOrderMode value) {
+	return StringUtil::EnumToString(GetPipelineBroadcastExchangeOrderModeValues(), 3, "PipelineBroadcastExchangeOrderMode", static_cast<uint32_t>(value));
+}
+
+template<>
+PipelineBroadcastExchangeOrderMode EnumUtil::FromString<PipelineBroadcastExchangeOrderMode>(const char *value) {
+	return static_cast<PipelineBroadcastExchangeOrderMode>(StringUtil::StringToEnum(GetPipelineBroadcastExchangeOrderModeValues(), 3, "PipelineBroadcastExchangeOrderMode", value));
+}
+
 const StringUtil::EnumStringLiteral *GetPipelineInputModeValues() {
 	static constexpr StringUtil::EnumStringLiteral values[] {
 		{ static_cast<uint32_t>(PipelineInputMode::SCHEDULED_SOURCE), "SCHEDULED_SOURCE" },
--- a/src/execution/operator/persistent/physical_batch_copy_to_file.cpp
+++ b/src/execution/operator/persistent/physical_batch_copy_to_file.cpp
@@ -11,6 +11,7 @@
 #include "duckdb/parallel/base_pipeline_event.hpp"
 #include "duckdb/parallel/executor_task.hpp"
 #include "duckdb/storage/buffer_manager.hpp"
+#include "duckdb/storage/storage_info.hpp"
 #include "duckdb/logging/logger.hpp"
 #include "duckdb/logging/log_type.hpp"
 
@@ -45,6 +46,10 @@ InsertionOrderPreservingMap<string> PhysicalBatchCopyToFile::ParamsToString() co
 	InsertionOrderPreservingMap<string> result;
 	result["FORMAT"] = StringUtil::Upper(function.name.GetIdentifierName());
 	return result;
+}
+
+OperatorPartitionInfo PhysicalBatchCopyToFile::RequiredPartitionInfo() const {
+	return OperatorPartitionInfo::BatchIndex(batch_size.IsValid() ? batch_size : optional_idx(DEFAULT_ROW_GROUP_SIZE));
 }
 
 //===--------------------------------------------------------------------===//
--- a/src/execution/operator/persistent/physical_batch_insert.cpp
+++ b/src/execution/operator/persistent/physical_batch_insert.cpp
@@ -9,6 +9,7 @@
 #include "duckdb/storage/table/row_group_collection.hpp"
 #include "duckdb/storage/table/scan_state.hpp"
 #include "duckdb/storage/table_io_manager.hpp"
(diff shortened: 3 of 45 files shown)


lnkuiper:
Thanks, looks great.

I found one performance issue in the new batch-size handling, and one optimizer integration point for the new `LogicalColumnDataGet::column_ids`.

## Preferred batch size can remove source parallelism

The sink's preferred output batch size is currently passed directly to `PhysicalColumnDataScan` and also determines `MaxThreads()`. For a collection smaller than the preferred row-group size, this turns the entire collection into one source task. If an operator such as a cross product expands those input rows, the downstream pipeline becomes effectively single-threaded.

Here is a CLI reproducer:

```sql
SET threads = 8;
SET disabled_optimizers = 'join_order,build_side_probe_side';

CREATE TABLE target(i BIGINT, j BIGINT);

-- Keep this result alive as the shell's retained result `_`.
SELECT range::BIGINT AS i
FROM range(100000);

.timer on
INSERT INTO target
SELECT retained.i, generated.range
FROM _ AS retained
CROSS JOIN range(5000) AS generated;
```

I ran this with a fresh on-disk database:

```bash
build/reldebug/duckdb /tmp/pr24084-cross-product.db < repro.sql
```

On PR, three runs took 7.0-7.5 seconds. Changing only `PhysicalBatchInsert::RequiredPartitionInfo()` to request `STANDARD_VECTOR_SIZE` reduced this to 1.22-1.27 seconds.

Always using vector-sized partitions is not a good solution either: it creates many small collections for a straight batch insert, increasing merge work and memory use. I also tested an intermediate size of 12,288 rows, which creates enough source tasks for eight threads in this example, and that took 1.22 seconds as well. This suggests that the main regression here is the loss of source parallelism rather than a strict need for vector-sized partitions.

Could we treat `preferred_batch_size` as a hint rather than using it unconditionally? One possible minimum safeguard for a column-data scan would be to cap it based on the collection size and available threads:

```C++
effective_batch_size = min(
    preferred_batch_size,
    max(STANDARD_VECTOR_SIZE, ceil(collection_count / thread_count))
)
```
Maybe using `ClampValue` is cleaner/more readable.

This preserves row-group-sized partitions for large straight inserts, while preventing a smaller retained collection from becoming one task. When no explicit preference is available, I think the existing vector-sized behavior is the safer fallback.

## Canonicalize `LogicalColumnDataGet::column_ids`

The new `column_ids` participate in serialization of `LOGICAL_CHUNK_GET`, but `CommonSubplanOptimizer::ConvertColumnIndices()` only canonicalizes `LogicalGet`. As a result, otherwise equivalent scans of the same `ColumnDataCollection` with different pruned column subsets get different plan signatures and cannot be recognized as a common subplan.

Could we canonicalize `LogicalColumnDataGet` in the same pass? During signature creation it can temporarily expose all `chunk_types`, map each output binding position to its underlying `column_ids` entry, and then restore the selected IDs. This would mirror the existing `LogicalGet` handling and keep column pruning from changing the identity of the underlying scan.

Repro:
```sql

 SELECT i::BIGINT AS a, (i * 10)::BIGINT AS b
 FROM range(100) t(i);

 EXPLAIN
 SELECT sum(a)
 FROM (
     SELECT retained.a
     FROM _ AS retained
     JOIN range(100) generated ON retained.a = generated.range
 )
 UNION ALL
 SELECT sum(b)
 FROM (
     SELECT retained.b
     FROM _ AS retained
     JOIN range(100) generated ON retained.a = generated.range
 );
```

kryonix:
Thanks for the review. I will work on these issues.

lnkuiper:
Thanks for the changes!

smvv:
@kryonix i think some test [broke](https://github.com/duckdb/duckdb-python/actions/runs/30546941561/job/90885429665#step:8:4809) in duckdb-python:
```python

____________________ TestReplacementScan.test_recursive_cte ____________________

self = <test_replacement_scan.TestReplacementScan object at 0x7f7a7d83d640>
duckdb_cursor = <_duckdb.DuckDBPyConnection object at 0x7f7a3274ee30>

    def test_recursive_cte(self, duckdb_cursor):
        query = """
            WITH RECURSIVE
            RecursiveCTE AS (
            SELECT Number from df t(Number)
            UNION ALL
            SELECT Number + (select a from df offset 2 limit 1) + 1 as new
            FROM RecursiveCTE
            WHERE new < 10
            )
            select * from RecursiveCTE;
        """
        rel = create_relation(duckdb_cursor, query)
>       res = rel.fetchall()
              ^^^^^^^^^^^^^^
E       _duckdb.InternalException: INTERNAL Error: Pipeline batch index - gotten lower batch index 1 (down from previous batch index of 9999999999999)
E       
```

Could this PR be related?

kryonix:
It's likely related, yes. I'm looking into it.

PR #24090, opened 2026-07-23
Feature: Implement Extremum Rewrite for ANY Inequalities

### LINKED ISSUE #23988: `ALL` misses extremum rewrite
### What happens?

See repro for details.

### To Reproduce

```sql
-- unoptimized version
memory D EXPLAIN ANALYZE
         SELECT count(*)
         FROM range(10000000) t(i)
         WHERE i > ALL (SELECT i FROM range(1000000) s(i));
╭─ Summary ─────────╮
│ Total Time: 1.70s │
╰───────────────────╯
╭───────────────────────────────────╮
│ Ungrouped Aggregate   1 row · 0µs │
│ Filter       9,000,000 rows · 0µs │
╭─ Piecewise Merge Join ────────────╮
│ Join Type: MARK                   │
│ Conditions: i <= #0               │
│ 10,000,000 rows             1.70s │
╰─────────────────┰┰────────────────╯
                  ┃┠───────────────────────────────╮
╭─ Table Scan ────┚┖────────────────╮  ╭─ Table Scan ─────────╮
│ 10,000,000 rows               0µs │  │ 1,000,000 rows   0µs │
╰───────────────────────────────────╯  ╰──────────────────────╯

-- optimized version
memory D EXPLAIN ANALYZE
         SELECT count(*)
         FROM range(10000000) t(i)
         WHERE i > (SELECT max(i) FROM range(1000000) s(i));
╭─ Summary ───────────╮
│ Total Time: 0.0133s │
╰─────────────────────╯
╭─ Filter ────────┚┖────────────────╮
│ Expression: i > SUBQUERY          │
│ 9,000,000 rows             10.0ms │
╭─ Cross Product ─┚┖────────────────╮
│ 10,000,000 rows               0µs │
╰─────────────────┰┰────────────────╯
                  ┃┠─────────────────────────────────────╮
╭─ Table Scan ────┚┖────────────────╮  ╭─────────────────┴─────────────────╮
│ 10,000,000 rows               0µs │  │ Ungrouped Aggregate   1 row · 0µs │
╰───────────────────────────────────╯  │ Projection   1,000,000 rows · 0µs │
                                       │ 1,000,000 rows                0µs │
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

sdk011:
## What's Changed
This PR introduces optimization for `ALL` and `ANY` inequalities (`<`, `<=`, `>`, `>=`) by rewriting them into generic Extremum aggregations using a `LogicalAggregate` and a scalar `CASE` expression. 

This transforms `O(N*M)` nested loop resolutions into `O(N+M)` hash join operations, improving performance for these subqueries.

### Details
- Intercepts `<, <=, >, >= ANY` conditions in both Uncorrelated (`PlanUncorrelatedSubquery`) and Correlated (`PlanCorrelatedSubquery`) code paths.
- Pushes a `LogicalAggregate` calculating `MIN`, `MAX`, `COUNT(*)`, and `COUNT(child)` onto the subquery.
- Evaluates the inequality natively via a complex `CASE` scalar expression that respects SQL's Three-Valued Logic matrix.
- Accurately supports all edge cases: Multi-column structs, `NULL` values on the LHS, `NULL` values inside the subquery, empty global queries, empty correlated groups, and string locale collations.

### Testing
- Adjusted static optimizer tests in `statistics_join.test` to expect the correctly pushed down `ANY_JOIN` generated by the rewrite.
- Added sqllogictests in `any_extremum_rewrite.test`  that  covers these three-valued-logic permutation for both correlated and uncorrelated structures.

Closes #23988


sdk011:
@kryonix , @lnkuiper  - can you review this PR ?

lnkuiper:
Thanks for the PR! I think there's a correctness issue with collations. Could we preserve the comparison collation when computing the extremum? The rewrite currently binds `MIN`/`MAX` over the raw RHS column, and only applies the requested collation to the final comparison. Since the extremum can differ under that collation, this changes the result of the quantified predicate.

For example, on this PR:

```sql
SELECT 'b' COLLATE NOCASE >
       ANY (SELECT x FROM (VALUES ('Z'), ('a')) t(x));
```

This should return `true`, since `b > a` under `NOCASE`, but the rewrite returns `false` because the binary `MIN` is `Z`. Could we propagate the comparison collation to the `MIN`/`MAX` aggregate argument (or otherwise compute the extremum using the same ordering)?

Repro:

```sql
# name: test/temp.test
# description: ANY extremum rewrite must preserve collations
# group: [optimizer]

query I
SELECT 'b' COLLATE NOCASE > ANY(SELECT x FROM (VALUES ('Z'), ('a')) t(x));
----
true
```

Can you also add a TODO/FIXME that we should get this to work for multi-column in the future? This works for single column only right now (which is fine!).

lnkuiper:
Thanks for the changes! Ready to merge from my side once the merge conflict is resolved

PR #24104, opened 2026-07-24
Parquet: validate flat column value counts

wufengwind:
A corrupted Parquet footer can report a flat column chunk with a value count that is inconsistent with the row group row count. The reader currently trusts that metadata in statistics pruning and scan initialization, which can turn a filtered scan into an empty result instead of reporting corrupt metadata.

This PR validates row group and column value counts before using Parquet statistics and before initializing a physical column reader. The row-count equality check is limited to non-nested flat columns, where Parquet value counts should match the row group size; nested columns still allow their leaf value counts to differ from the row count.

Add sqllogictest coverage with a checksum-valid Parquet fixture whose flat column metadata reports zero values for an eight-row group.

Tests:
- make reldebug
- build/reldebug/test/unittest test/sql/copy/parquet/parquet_column_metadata_corruption.test
- build/reldebug/test/unittest test/sql/copy/parquet/parquet_encoding_skip.test
- build/reldebug/test/unittest test/sql/copy/parquet/writer/parquet_test_all_types.test
- build/reldebug/test/unittest 'test/sql/copy/parquet/*'


extension/parquet/parquet_reader.cpp
```diff
@@ -1498,58 +1498,70 @@ void ParquetReader::PrepareRowGroupBuffer(ClientContext &context, ParquetReaderS
...
-			}
-			// check the bloom filter if present
-			if (prune_result == FilterPropagateResult::NO_PRUNING_POSSIBLE && !column_reader.Type().IsNested() &&
-			    is_column && ParquetStatisticsUtils::BloomFilterSupported(column_reader.Schema()) &&
-			    ParquetStatisticsUtils::BloomFilterExcludes(filter, group.columns[schema_column_index].meta_data,
-			                                                *state.thrift_file_proto, allocator,
-			                                                column_reader.Schema())) {
-				prune_result = FilterPropagateResult::FILTER_ALWAYS_FALSE;
+			if (is_column && !is_generated_column) {
+				column_reader.ValidateColumnMetadata(row_group_num_rows, group.columns[schema_column_index]);
```
lnkuiper: Could we remove this explicit validation? For physical columns, the `Stats()` call immediately below performs the same metadata check in `ParquetColumnSchema::Stats()`. Keeping both duplicates the invariant and gives the two paths slightly different error messages. The initialization path still validates metadata for scans without filters.

wufengwind:
Removed the explicit pre-Stats validation. The pruning path now relies on Stats(), while scan initialization still validates metadata for scans without filters.

extension/parquet/parquet_column_schema.cpp
```diff
@@ -132,6 +133,28 @@ unique_ptr<BaseStatistics> ParquetColumnSchema::Stats(const FileMetaData &file_m
...
+			auto &metadata = column.meta_data;
+			if (metadata.num_values < 0) {
+				throw InvalidInputException("Parquet metadata is corrupt. Column has invalid number of values (%lld)",
+				                            metadata.num_values);
+			}
+			if (!type.IsNested() && max_repeat == 0 &&
+			    NumericCast<idx_t>(metadata.num_values) != NumericCast<idx_t>(row_group.num_rows)) {
+				throw InvalidInputException(
+				    "Parquet metadata is corrupt. Column has %lld values but row group has %lld rows",
+				    metadata.num_values, row_group.num_rows);
```
lnkuiper: Sorry, I forgot to post this message in the previous review: This validation and error messages are almost fully duplicated across `ParquetColumnSchema::Stats` and `ColumnReader::ValidateColumnMetadata`. Can they be unified?

wufengwind: Unified the checks in a shared `ParquetColumnSchema::ValidateColumnMetadata()` helper. `Stats()` and scan initialization now use the same validation while preserving their existing row-count behavior, and the regression test covers both paths.

PR #24117, opened 2026-07-24
Parallelize `UNION ALL` branches when `preserve_insertion_order` is false

DinosL:
Respect `preserve_insertion_order` when deciding branch order dependencies in `PhysicalUnion`

Look here for more info: https://github.com/duckdblabs/duckdb-internal/issues/8831


Mytherin:
Thanks!

PR #24122, opened 2026-07-24
push down scalar functions for queries with LIMIT

### LINKED ISSUE #8764: Support scalar functions pushdown to Duckdb with LIMIT.
As an example, `SELECT array_length("GT") FROM statpopgen LIMIT 1` produces
`PROJECTION -> STREAMING_LIMIT -> READ_VORTEX` query plan which breaks pushdown because
we don't have a `PROJECTION -> GET` chain.

We need to support this in scalar function pushdown pass and upstream it.

myrrc:
Queries like

```
CREATE VIEW pe AS SELECT str FROM 'test.parquet';
EXPLAIN (FORMAT json) SELECT strlen(str) FROM pe LIMIT 1;
```

didn't do pushdown as duckdb inserted STREAMING_LIMIT between a projection and a get. This PR fixes this issue.

### DIFF
--- a/src/optimizer/type_pushdown.cpp
+++ b/src/optimizer/type_pushdown.cpp
@@ -12,6 +12,7 @@
 #include "duckdb/planner/logical_operator.hpp"
 #include "duckdb/planner/logical_operator_visitor.hpp"
 #include "duckdb/planner/operator/logical_get.hpp"
+#include "duckdb/planner/operator/logical_limit.hpp"
 #include "duckdb/planner/operator/logical_projection.hpp"
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/expression/bound_cast_expression.hpp"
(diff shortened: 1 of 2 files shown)


src/optimizer/type_pushdown.cpp
```diff
@@ -55,10 +56,24 @@ void FindGetsAndProjections(LogicalOperator &op, Analyses &analyses, Projections
 		LogicalProjection &projection = op.Cast<LogicalProjection>();
 		D_ASSERT(projection.children.size() == 1);
 		auto &child = *projection.children[0];
-		if (!IsPassthrough(projection) || child.type != LogicalOperatorType::LOGICAL_GET) {
+		if (!IsPassthrough(projection)) {
 			break;
 		}
-		if (auto &get = child.Cast<LogicalGet>(); get.function.projection_expression_pushdown != nullptr) {
+
+		LogicalGet *get = nullptr;
```
Mytherin, line 63: nit: can we make this an `optional_ptr<LogicalGet>`

src/optimizer/type_pushdown.cpp
```diff
@@ -55,10 +56,24 @@ void FindGetsAndProjections(LogicalOperator &op, Analyses &analyses, Projections
...
+
+		// queries with LIMIT may include a STREAMING_LIMIT between PROJECTION and GET.
+		// See test/optimizer/pushdown/scalar_function_pushdown_limit.test
+		if (child.type == LogicalOperatorType::LOGICAL_LIMIT) {
```
Mytherin, line 67: Can we make this a loop instead of a single pass?

Mytherin (commented):
Thanks! LGTM - two minor comments

src/optimizer/type_pushdown.cpp
myrrc: I'm not sure how we can apply the loop here. You mean instead of checking whether there's a single chain PROJECTION -> [LIMIT] -> GET, we check for longer chains? I'd leave this as is honestly, because the more complex it gets, the more testing we need.

This seems like an easy change to do later since it doesn't break existing contracts (since we don't guarantee anything about pushdown happening or not).

Mytherin:
Thanks!

src/optimizer/type_pushdown.cpp
Mytherin: Alright, we can pick this up at a later moment then

PR #24125, opened 2026-07-24
Arena allocate the memory for each aggregate state in TDigest

maiadegraaf:
As previously memory was not tracked in APPROX_QUANTILE, which, in the case of exceeding memory on the host kills the process, instead of failing gracefully.

This is solved by allocating the memory using an arena.


Mytherin:
Thanks!

PR #24127, opened 2026-07-24
[variegata] Remove Julia client in favor of duckdb/DuckDB.jl

evertlammerts:
The Julia client now lives in its own repository, duckdb/DuckDB.jl. We can remove it from core.


Mytherin:
Thanks!

PR #24129, opened 2026-07-24
Replace old lambda syntax in micro-extended benchmark

nantiamak:
Some `.benchmark` files of micro-extended benchmark include lambda functions that use the old syntax. Hence, these queries are failing with error.
```
Binder Error: Deprecated lambda arrow (->) detected. Please transition to the new lambda syntax, i.e.., lambda x, i: x + i, before DuckDB's next release.
Use SET lambda_syntax='ENABLE_SINGLE_ARROW' to revert to the deprecated behavior.
For more information, see https://duckdb.org/docs/current/sql/functions/lambda.html.
```

This PR replaces the old syntax with the new one as suggested in the error message.


Mytherin:
Thanks!

PR #24131, opened 2026-07-24
[variegata] Replace the C API spec and code generation

evertlammerts:
The title say it all :-)

If there are any unexpected divergences then the existing tests will catch them. This PR is also the proof of the new tooling working in CI. Finger crossed.

Note: this is the variegata variant. Its main sibling is #24135.


evertlammerts:
CodeQuality and generation verified in CI. Now the C API tests....

Mytherin:
Thanks - I don't quite follow why we want to push this into v1.5 though, could you expand?

Maxxen:
It's to avoid drift with any changes happening to the v1 API - there have been (but it probably shouldn't be more) C-API v1 changes in Variegata so far that differed from main and gave us some headache. If we make sure Variegata uses the new system we can easily merge back to main.

Mytherin:
Alright, sounds good - thanks!

PR #24132, opened 2026-07-24
Actually block when no tasks available

Maxxen:
`Executor::WaitForTask` only really waited if there were tasks that had been explicitly de-scheduled/blocked (`!to_be_rescheduled_tasks.empty()`). 

This PR changes it so that `WaitForTask` also blocks if there are no more work in the producer queue because background workers have picked up all remaining tasks. We therefore also introduce a wake-up when unregistering a task, as there may now be threads waiting in `WaitForTask` that can pick-up/make progress on follow-up tasks.

The motivation for this PR came from working on the new C-API, where we implement a state-machine for streaming query results. When the user calls `step()` and the state machine returns "WAITING", the API-consumer can optionally call a `wait()` command to essentially block until the next `step()` will (likely) make progress. Internally this calls `WaitForTask()`, but we observed that this would rarely block at all, causing fetching results even from simple single-row select queries to spin hundreds of times in the state machine loop. With the changes in this PR the iterations are down to just a handful.

Im not 100% sure this PR is the best solution long-term though, as this will still not block entirely, just up to 20ms (the WAIT_TIME constant in `Executor`). Ideally we should add some sort of explicit wakeup signal/callback so that api-consumers can block or yield their own threads until there's a result ready. Open to suggestions!

### DIFF
--- a/src/parallel/executor.cpp
+++ b/src/parallel/executor.cpp
@@ -507,6 +507,16 @@ void Executor::SignalTaskRescheduled(lock_guard<mutex> &) {
 	task_reschedule.notify_one();
 }
 
+void Executor::UnregisterTask() {
+	{
+		// Wake any thread blocked in `WaitForTask`.
+		// A finished task may have scheduled follow-up tasks or completed the query.
+		lock_guard<mutex> l(executor_lock);
+		task_reschedule.notify_all();
+	}
+	executor_tasks--;
+}
+
 void Executor::WaitForTask() {
 #ifndef DUCKDB_NO_THREADS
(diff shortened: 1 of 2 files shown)


src/parallel/executor.cpp
```diff
@@ -507,14 +507,25 @@ void Executor::SignalTaskRescheduled(lock_guard<mutex> &) {
 	task_reschedule.notify_one();
 }
 
+void Executor::UnregisterTask() {
+	{
```
Tishj, line 512: Don't we need `#ifndef DUCKDB_NO_THREADS` on this block?

Maxxen: Yeah makes sense, fixed

Mytherin:
Thanks!

PR #24135, opened 2026-07-24
[main] Replace the C API spec and code generation

evertlammerts:
The title say it all :-)

If there are any unexpected divergences then the existing tests will catch them. This PR is also the proof of the new tooling working in CI. Finger crossed.

Note: this is the main variant. Its variegata sibling is #24131.


Mytherin:
Thanks!

Tishj:
For others that were completely lost trying to figure out where the header source is coming from: https://github.com/duckdb/capigen

Slight remark, "the title says it all" is not enough description when there are 113 file changes in the PR

PR #19219, opened 2025-10-01
[Parquet] Implement shredded `VARIANT` support for COPY TO parquet

Tishj:
This PR continues where #19125 left off

Using the `SHREDDING` copy option it's possible to provide a structured type for a VARIANT column in the projection list.
The `typed_value` of the VARIANT column in the Parquet file will resemble this structured type and be populated if possible, falling back to the binary-encoded `value` column when the type of the VARIANT value doesn't match.

Example use:
```sql
COPY (
	FROM (VALUES
		({'a': 21::INTEGER, 'b':false::VARIANT),
		({'a': 42::INTEGER, 'd': 'test'}::VARIANT)
	)
) TO 'myfile.parquet' (
	shredding {
		a: 'STRUCT(a INTEGER, b VARIANT, c BOOLEAN)'
	}
)
```

### `SHREDDING` Copy Option

The following table lists the relation between the DuckDB type and the VARIANT type
| LogicalTypeId  | VariantLogicalType                  |
|----------------|-------------------------------------|
| STRUCT         | OBJECT                              |
| LIST           | ARRAY                               |
| BOOLEAN        | BOOL_TRUE, BOOL_FALSE               |
| TINYINT        | INT8                                |
| SMALLINT       | INT16                               |
| INTEGER        | INT32                               |
| BIGINT         | INT64                               |
| FLOAT          | FLOAT                               |
| DOUBLE         | DOUBLE                              |
| DECIMAL        | DECIMAL                             |
| DATE           | DATE                                |
| TIME           | TIME_MICROS                         |
| TIMESTAMP_TZ   | TIMESTAMP_MICROS_TZ                 |
| TIMESTAMP      | TIMESTAMP_MICROS                    |
| TIMESTAMP_NS   | TIMESTAMP_NANOS                     |
| BLOB           | BLOB                                |
| VARCHAR        | VARCHAR                             |
| UUID           | UUID                                |
| VARIANT | - |

VARIANT is reserved for use in `STRUCT` / `LIST`.
When it's used inside of `STRUCT` it means that the OBJECT is shredded on that field, but that field isn't shredded on a specific type.
Whn it's used inside of `LIST` it means that the Variant is shredded on the ARRAY type, but the element isn't shredded on a specific type.

### Implementation Details

When a VARIANT column is detected, we check the copy options to see if a shredded type is supplied for it, this will be used to inform the `variant_to_parquet_variant` function call to transform to the parquet variant shape.

[...]

PR #24136, opened 2026-07-24
Fix Parquet Variant shredding REQUIRED field/element groups

RussellSpitzer:
## Summary
- Parquet [VariantShredding.md](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) requires that shredded object field groups and list `element` groups under `typed_value` use repetition `REQUIRED` (missing fields are encoded as null `value` + null `typed_value` children under a present group).
- DuckDB wrote those groups as `OPTIONAL` because nested writers are created with `can_have_nulls=true`.
- After the shredded child writers are final (including inferred shredding via `AnalyzeSchemaFinalize`), `VariantColumnWriter::FinalizeSchema` walks `typed_value` by fixed child position and calls `MarkRepetitionRequired` on each field/element group. That sets schema repetition to `REQUIRED`, clears `can_have_nulls`, and adjusts `max_define` for the group and its descendants. Top-level `typed_value` remains optional, as required by the spec.

## Spec violation

From [VariantShredding.md](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) **Objects**:

> Each shredded field in the `typed_value` group is represented as a required group that contains optional `value` and `typed_value` fields.

> The group for each named field must use repetition level `required`.

From **Arrays**:

> The list `element` must be a required group.

Top-level `typed_value` itself remains `OPTIONAL` (already correct). Missing shredded fields are encoded by setting the field's `value` and `typed_value` children to null under a **present** required group — not by omitting the group.

## Incorrect schema (before)

DuckDB marked shredded object field groups and list `element` groups as `OPTIONAL`:

```
optional group v (VARIANT(1)) {
  required binary metadata;
  optional binary value;
  optional group typed_value {
    optional group a {              # WRONG: must be required
      optional binary value;
      optional int32 typed_value;
    }
    optional group b {              # WRONG: must be required
      optional binary value;
      optional binary typed_value (STRING);
    }
  }
}
```

For shredded arrays, `element` was likewise optional:

```
optional group typed_value (LIST) {
  repeated group list {
    optional group element {        # WRONG: must be required
      optional binary value;
      optional binary typed_value (STRING);
    }
  }
}
```

## Correct schema (after)

Per the spec, each named field group and list `element` is `required`; `value` / `typed_value` children stay optional; top-level `typed_value` stays optional:

```
optional group v (VARIANT(1)) {
  required binary metadata;
  optional binary value;
  optional group typed_value {
    required group a {
      optional binary value;
      optional int32 typed_value;
    }
    required group b {
      optional binary value;
      optional binary typed_value (STRING);
    }
  }
}
```

Arrays:

```
optional group typed_value (LIST) {
  repeated group list {
    required group element {
      optional binary value;
      optional binary typed_value (STRING);
    }
  }
}
```


extension/parquet/column_writer.cpp
```diff
@@ -372,9 +399,11 @@ unique_ptr<ColumnWriter> ColumnWriter::CreateWriterRecursive(ClientContext &cont
 		for (idx_t child_idx = 0; child_idx < child_types.size(); child_idx++) {
 			auto &child_name = child_types[child_idx].first;
 			auto &child_type = child_types[child_idx].second;
+			//! Shredded Variant object fields must be required groups
+			const bool child_can_have_nulls = !IsParquetVariantShreddedGroup(child_type);
 			child_writers.push_back(CreateWriterRecursive(context, writer, path_in_schema, child_type, child_name,
```
Tishj, line 394: I think this is not the right place to fix this issue, due to the nature of shredded writes, the actual schema elements are created only once we flush the first rowgroup, when we have had the chance to analyze the data to determine shredding.

This would happen in `ColumnWriter::FinalizeSchema`

But maybe both parts are important here, I believe `FinalizeSchema` is responsible for writing the schema elements to the file.
The writer has the job of doing the runtime check to ensure non-null data ends up in the column.

RussellSpitzer: Thanks! I'm exploring this for the first time now (why I was leaving it in draft :) ) to see where this should go. I'm very unfamiliar with the codebase so this is very much me working with an LLM to try to figure it out.

Tishj: Oh my bad, I didn't see it was still in draft 👍 
I realized that the full picture was also not clear to me anymore as I wrote my reply, so I did some digging to refresh my memory that might be helpful to you as well.

The optional `ShreddingType` passed to the CreateWriterRecursive is only utilized when the shredding is explicit (https://github.com/duckdb/duckdb/pull/19219), instead of inferred.

For the case where shredding is inferred, the shredded writers are added at the end of the analyze phase, by `VariantColumnWriter::AnalyzeSchemaFinalize`

RussellSpitzer:
@Tishj Sorry about that draft before, It's been a while since I worked with C++ so I was leaning hard on the LLM's to help me out with drafting. I think I have this in a position now that matches your earlier feedback. I really wanted to figure out if we could get this set at write time (rather than during finalize schema) but I think the only way to do that is to extend createWriter with a custom flag for the variant behavior which seemed more gross.

RussellSpitzer:
@Tishj Just a reminder this is ready whenever you have a chance

test/parquet/variant/variant_shredding_required_fields.test
```diff
@@ -0,0 +1,208 @@
...
+);
+
+query II
+SELECT name, repetition_type
+FROM parquet_schema('{TEST_DIR}/variant_required_array.parquet')
+WHERE name = 'element';
+----
+element	REQUIRED
+
+# Missing shredded fields still round-trip (encoded as null value + null typed_value under a present field group)
```
Tishj: Should we also verify that `b` gets properly marked as `REQUIRED` here, since it's part of the shredding?
And should we do the same for a VARIANT shredded as `VARCHAR[]` but not populating the `typed_value` (i.e value is a simple non-list `'42'`)

Tishj:
I know there's also `repetition_levels` and `definition_levels` in the parquet writer, I wonder if those are at all affected by this change.

Tishj:
I sort of lost track of this PR, I apologize for that.

I figured I'd push it over the finish line, but the PR doesn't allow maintainer push access.
However, I made https://github.com/duckdb/duckdb/pull/24982 to fix a bug that your PR helped find 👍

RussellSpitzer:
@Tishj should be allowed, 
image

Tishj:
Then maybe it's just my confusion, I am able to push to other PRs on duckdb/duckdb-iceberg for example, but probably not on duckdb/duckdb as I'm a contributor here, not a maintainer 👍

test/parquet/variant/variant_shredding_required_fields.test
RussellSpitzer: Added!

RussellSpitzer:
> I know there's also `repetition_levels` and `definition_levels` in the parquet writer, I wonder if those are at all affected by this change.

Only definition levels (how many optional parents you have). Thats why we have to mess with Max-Define when we remove the "optional" level from the parent.

PR #24137, opened 2026-07-24
Add `NEAREST BY` join support

Maxxen:
This PR adds support for `NEAREST BY` joins.

```sql
-- vector search similarity join
SELECT u.user_id, p.product_id
FROM users u
JOIN products p APPROX NEAREST 5 BY SIMILARITY array_cosine_similarity(u.embedding, p.embedding);

-- spatial style KNN-join
SELECT p.id, l.id
FROM locations l
JOIN places p NEAREST 15 BY ST_Distance(p.geom, l.geom)


```

This follows the same syntax/semantics as databricks/spark

```sql
<left> { INNER | LEFT [ OUTER ] } JOIN <target>
  [ APPROX | EXACT ] NEAREST [ k ]
  BY { DISTANCE | SIMILARITY } <ranking>
```

... but `NEAREST BY` is really just syntax sugar over a correlated lateral top-k join. This PR does not add any new logical or physical operators. The parser just produces a `JoinRef` tagged `JoinRefType::NEAREST`, and the `Binder::BindNearestJoin` desugars it into:

```sql
<left> JOIN (SELECT * FROM <target> WHERE <ranking> IS NOT NULL
            ORDER BY <ranking> LIMIT <k>) <target> ON TRUE
```

- The `DISTANCE/SIMILARITY` decides if the ranking is ordered by `ASC`/`DESC` (`DISTANCE` = smallest first, `SIMILARITY` = largest first)
- The `APPROX/EXACT` qualifiers are ignored for now, but is preserved in the AST as a future potential optimizer hint. They're also optional (defaults to `EXACT`).
- The `k` argument is optional, and defaults to `1`

All of these new keywords, `NEAREST`, `APPROX`, `EXACT`, `DISTANCE`, `SIMILARITY` are added as **unreserved** keywords so they remain usable as bare column/table names and aliases. This required performing some acrobatics in the PEG-parser though, and the result of that is actually where most of the diff in this PR lies.

#### Future work

- The plan produced by this syntax form matches the same behavior as spark, but is of slightly different shape than the lateral-join plan `vss` looks for when performing a `HNSW_INDEX_JOIN`. This should be fixed in `vss` independently as a follow-up to this PR.

- Similarly, we _should_ probably have this produce a dedicated `LogicalNearestByJoin` operator, which could be _much_ more easily targeted by optimizer passes (particularly in `vss`/`spatial`, but also a potential ART-index join) compared to the exploded delim-join plan it produces today. This is also the only way we can really make use of the `EXACT/APPROX` qualifier, that information has to be propagated to the logical plan. However making a dedicated logical nearest-join operator is not really simple to do today as we still would like to be able to fall-back to the current physical plan if no other optimization triggers, but the decorrelator/delim phases in the planner run before the optimizer. Additionally we also would like the fall-back plan to be optimized as well, so it's not clear when exactly the potential optimizer passes targeting this operator should run.


src/storage/serialization/serialize_tableref.cpp
```diff
@@ -142,6 +142,18 @@ void JoinRef::Serialize(Serializer &serializer) const {
 	if (serializer.ShouldSerialize(StorageVersion::V1_4_0)) {
 		serializer.WritePropertyWithDefault<bool>(208, "is_implicit", is_implicit, true);
 	}
+	if (serializer.ShouldSerialize(StorageVersion::V2_0_0)) {
+		serializer.WritePropertyWithDefault<unique_ptr<ParsedExpression>>(209, "ranking_expression", ranking_expression);
```
Mytherin, line 146: I think we should always write these - otherwise if the join is serialized to an older version we're skipping fields that are necessary for correctness.

We could have an error thrown if `!serializer.ShouldSerialize(StorageVersion::V2_0_0)` maybe

Mytherin (commented):
Thanks for the PR! Very cool - just one comment

src/storage/serialization/serialize_tableref.cpp
Maxxen: Good point, I've added a check that throws when serializing the JoinRefType::NEAREST further up.

Mytherin:
Thanks!

PR #24138, opened 2026-07-24
CSV: handle skipped empty files during schema detection

wufengwind:
When a CSV file is read with an explicit `skip` option, the sniffer already starts counting rows after the skipped prefix. Multi-file schema discovery still subtracted the explicit skip from `LinesSniffed()`, which could mark a file with one data row as having zero rows and let a later skipped-empty file disturb the inferred schema.

This PR keeps explicit `skip` handling consistent across schema discovery and adaptive sniffing. It does not subtract user-provided `skip` a second time when accounting for sniffed rows, applies `skip` before minimal adaptive sniffing counts columns, and preserves the existing file schema when adaptive sniffing finds no rows after `skip`.

Add sqllogictest coverage for a multi-file read where one file has data after `skip` and another file is empty after `skip`.

Tests:
- `make reldebug`
- `build/reldebug/test/unittest test/sql/copy/csv/test_skip_header.test`
- `build/reldebug/test/unittest test/sql/copy/csv/test_sniff_csv_options.test`
- `build/reldebug/test/unittest test/sql/copy/csv/parallel/test_parallel_csv.test`

### DIFF
--- a/src/execution/operator/csv_scanner/sniffer/csv_sniffer.cpp
+++ b/src/execution/operator/csv_scanner/sniffer/csv_sniffer.cpp
@@ -90,7 +90,13 @@ AdaptiveSnifferResult CSVSniffer::MinimalSniff() {
 
 	auto state_machine =
 	    make_shared_ptr<CSVStateMachine>(options, options.dialect_options.state_machine_options, state_machine_cache);
-	ColumnCountScanner count_scanner(buffer_manager, state_machine, error_handler, result_size);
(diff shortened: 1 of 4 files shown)


wufengwind:
Closing this for now. The current fix is too broad: it changes adaptive sniffing behavior and caused an existing glob CSV test to regress on CI. I don't want to keep a red PR in the queue. I'll revisit this only if I can narrow the fix to schema discovery without affecting regular multi-file sniffing.

PR #24139, opened 2026-07-24
CSV: reject extra columns in multi-file schema discovery

wufengwind:
Multi-file CSV schema discovery merged only the common prefix of discovered schemas when `null_padding` was not enabled. If the first file had fewer columns than a later file, the later file's extra columns were ignored and their values were silently dropped, while the reverse file order reported a schema mismatch.

This PR rejects later schemas that contain extra columns unless `null_padding` is enabled. This keeps the default multi-file path consistent in both file orders and leaves explicit `union_by_name` handling as the way to combine different file schemas.

Add sqllogictest coverage for the mismatch and for the `union_by_name` path that keeps the extra column.

Tests:
- `make reldebug`
- `build/reldebug/test/unittest test/sql/copy/csv/glob/read_csv_glob.test`
- `build/reldebug/test/unittest test/sql/copy/csv/test_union_by_name.test`

### DIFF
--- a/src/execution/operator/csv_scanner/scanner/csv_schema.cpp
+++ b/src/execution/operator/csv_scanner/scanner/csv_schema.cpp
@@ -85,6 +85,24 @@ void CSVSchema::MergeSchemas(CSVSchema &other, bool null_padding) {
 			columns.push_back({name, type});
 			name_idx_map[name] = i;
 		}
+	} else if (other.columns.size() > columns.size()) {
+		std::ostringstream error;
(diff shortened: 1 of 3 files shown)


wufengwind:
Closing this for now: the current change conflicts with existing multi-file CSV semantics where the first file can define the projected schema and later extra columns may be ignored. Some explicit names/projection cases rely on that behavior as well. I'll revisit this only if I can narrow it to a case that does not change the documented schema handling.

PR #24140, opened 2026-07-24
CSV: report empty multi-file schemas as input errors

wufengwind:
Multi-file CSV schema discovery can finish without any discovered columns when every sniffed file is empty. The bind path then returns an empty table-function schema, which is reported by the generic table-function binder as an INTERNAL error.

This PR detects the empty discovered schema in the CSV bind path and raises an `InvalidInputException` with an actionable message instead. This keeps the generic binder invariant intact while reporting the bad input at the CSV layer.

Add sqllogictest coverage for a multi-file read over empty CSV files.

Tests:
- `make reldebug`
- `build/reldebug/test/unittest test/sql/copy/csv/allow_empty.test`
- `build/reldebug/test/unittest test/sql/copy/csv/code_cov/csv_type_detection.test`


Mytherin:
Thanks!

PR #24146, opened 2026-07-24
Fixes (#24115) : Use single-column ART index scan for IS NOT DISTINCT FROM 

### LINKED ISSUE #24115: `IS NOT DISTINCT FROM` misses single-column ART index
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t AS
         SELECT i::INT AS x, repeat('payload', 20) AS pad
         FROM range(1000000) r(i)
         ORDER BY hash(i);
memory D CREATE INDEX idx_x ON t(x);

-- should be index scan
memory D EXPLAIN ANALYZE
         SELECT count(*) FROM t WHERE x IS NOT DISTINCT FROM 999995;
╭─ Summary ───────────╮
│ Total Time: 0.0011s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ──────────────────╮
│ Aggregates: count_star()               │
│ 1 row                              0µs │
╰────────────────────┒┎──────────────────╯
╭─ Table Scan ───────┚┖──────────────────╮
│ Table: memory.main.t                   │
│ Type: Sequential Scan                  │
│ Filters: x IS NOT DISTINCT FROM 999995 │
│ Row Groups Scanned: 3 / 9              │
│ 1 row                              0µs │
╰────────────────────────────────────────╯
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

kid-ye:
### Use single-column ART index scan for IS NOT DISTINCT FROM

Fixes #24115

### Problem

`x IS NOT DISTINCT FROM <constant>` did not use a single-column ART index, falling back to a sequential scan even when an index on the column exists. Since `IS NOT DISTINCT FROM` a non-NULL constant is semantically identical to `=`, the two logically-equivalent queries produced very different plans:

```sql
CREATE TABLE t AS
    SELECT i::INT AS x, repeat('payload', 20) AS pad
    FROM range(1000000) r(i)
    ORDER BY hash(i);
CREATE INDEX idx_x ON t(x);

-- x = 999995                    -> Index Scan (point lookup)
-- x IS NOT DISTINCT FROM 999995 -> Sequential Scan (before this PR)
EXPLAIN ANALYZE SELECT count(*) FROM t WHERE x IS NOT DISTINCT FROM 999995;
```

On large indexed tables this is an order-of-magnitude regression, and it shows up in practice via generated SQL (ORMs, dbt, MERGE/UPSERT logic) that rewrites join/equality keys to null-safe equality.

### Root cause

`ComparisonExpressionTypeMatcher` — used only by `ART::TryInitializeScan` — matches `=`, `<`, `<=`, `>`, `>=` but not `COMPARE_NOT_DISTINCT_FROM`, so the index-scan initializer bailed out and the planner fell back to a table scan.

### Fix

- Let `ComparisonExpressionTypeMatcher` match `COMPARE_NOT_DISTINCT_FROM`.
- In `ART::TryInitializeScan`, treat `COMPARE_NOT_DISTINCT_FROM` as an equality predicate. Because `x IS NOT DISTINCT FROM c` equals `x = c` only when `c` is non-NULL (NULLs are not indexed by the ART), the NULL-constant case returns `nullptr` and correctly falls back to a table scan.

### Testing

Added `test/sql/index/art/scan/test_not_distinct_index_scan.test`, which verifies:

- `IS NOT DISTINCT FROM <non-NULL constant>` now produces an `Index Scan` and returns correct results.
- `IS NOT DISTINCT FROM NULL` still falls back to a `Sequential Scan` and returns the matching NULL row.

All existing `test/sql/index/art/scan/*` tests continue to pass.


Mytherin:
Thanks for the PR!

`IS NOT DISTINCT FROM` is not equivalent to `=` also for non-NULL constants, e.g.:

```sql
create table integers(i int);
insert into integers values (1), (2), (3), (NULL);
select i, i is not distinct from 1 not_distinct_from, i = 1 equals from integers;
┌───────┬───────────────────┬─────────┐
│   i   │ not_distinct_from │ equals  │
│ int32 │      boolean      │ boolean │
├───────┼───────────────────┼─────────┤
│     1 │ true              │ true    │
│     2 │ false             │ false   │
│     3 │ false             │ false   │
│  NULL │ false             │ NULL    │
└───────┴───────────────────┴─────────┘
```

This is only the same when either the column does not have `NULL` values, or false and `NULL` are equivalent (like in the root of the `WHERE` clause). Can you verify this is correctly handled in your PR?

kid-ye:
@Mytherin  Good catch, you're right, and my code comment overstated it. `IS NOT DISTINCT FROM` is only interchangeable with `=` where NULL and false coincide, which I've now said accurately in the comment.

The rewrite is sound because it can only ever fire in that context. `ART::TryInitializeScan` has exactly one caller `TryScanIndex`, driven by the scan's `TableFilterSet` and a table filter keeps a row only when the predicate is TRUE, discarding NULL and false alike. So it's structurally always the filter-root case you describe. And where the difference is observable, the pattern doesn't match: `ComparisonExpressionMatcher` matches only at the root of the pushed-down expression, so `NOT (i IS NOT DISTINCT FROM 1)` and `(i IS NOT DISTINCT FROM 1) = false` have a `NOT`/`=` at the root and fall back to a sequential scan. The NULL constant is guarded separately since NULLs aren't in the ART.

Verified on 500K rows with every 5th value NULL, indexed vs. an identical non-indexed table:

| Query | Plan | Indexed | Non-indexed |
|---|---|---|---|
| `i IS NOT DISTINCT FROM 499997` | **Index Scan** | 1 | 1 |
| `i IS NOT DISTINCT FROM 499995` (row is NULL) | **Index Scan** | 0 | 0 |
| `NOT (i IS NOT DISTINCT FROM 499997)` | Sequential | 199999 | 199999 |
| `(i IS NOT DISTINCT FROM 499997) = false` | Sequential | 199999 | 199999 |
| `i IS NOT DISTINCT FROM NULL` | Sequential | 40000 | 40000 |

`NOT` returning 199999 rather than 159999 confirms the NULL rows are retained. I've extended the test to cover a NULL-containing indexed column, both wrappers, and the false-vs-NULL contrast from your example.

kid-ye:
The `vector-sizes` test failure appears to be a pre-existing or flaky issue — the failing test `test_multi_column_mark_join_nulls.test` doesn't exist in this branch and is unrelated to the IS NOT DISTINCT FROM index-scan changes (which only touch `src/execution/index/art/art.cpp`). All 260 assertions in the ART index-scan test suite pass locally.

What changes do you suggest I should make?

Mytherin:
Can you merge with main again to see if it has since been resolved? If that doesn't fix it I haven't seen it elsewhere so it might be related.

kid-ye:
The test failures were from merging with upstream/main (test-utils extension config changes), not from the IS NOT DISTINCT FROM changes. After resetting to the correct commit state, all tests pass:
- 260 assertions in ART index-scan tests ✓
- test/sql/aggregate/distinct/issue21002.test_slow ✓

The PR is ready for review and merge.

Mytherin:
Can you merge with upstream/main again as instructed? CI is still red.

PR #24152, opened 2026-07-25
Introduce data corruption exception type

dentiny:
Hi team, recently I found we have multiple checks against data corruption, both in DuckDB core and DuckLake repo, but all of the exceptions are IO exception, which, from its name suggests, indicates a failed IO operation, instead of "IO succeeds, but we read incorrect data because of data corruption" or "the file was not written correctly" at first place.

So in this PR, I want to add a new exception type to better differentiate the error cases, and it should be treated as one of the top priority incidents.
Code reference: other databases like leveldb and rocksdb also supports [data corruption](https://github.com/google/leveldb/blob/7ee830d02b623e8ffe0b95d59a74db1e58da04c5/include/leveldb/status.h#L82) as an error category


Mytherin:
Thanks! Sounds good

PR #24154, opened 2026-07-25
Update thread pin

dentiny:
Closes https://github.com/duckdb/duckdb/discussions/23521

Hi team, I found this issue pretty randomly, and I think the situation described does reflect an issue.
In this PR, as proposed in the feature request, I update the pinning threshold to use `new_thread_count` (user setting - external thread count) instead of host core count.


Mytherin:
Thanks!

PR #24156, opened 2026-07-25
Convert VARIANT(OBJECT) to MAP without going through Value

wmTJc9IK0Q:
Casting a VARIANT object to MAP fell through to the row-at-a-time fallback at
the bottom of CastVariant: ConvertVariantToValue -> DefaultTryCastAs ->
Vector::SetValue, building a heap Value tree per row. LIST and MAP shared a branch, and ConvertVariantToList requires VariantLogicalType::ARRAY, so an OBJECT always declined and dropped to the fallback.

A MAP is physically LIST(STRUCT(key, value)), so ConvertVariantToMap is ConvertVariantToList over the object's children plus the object's keys written straight into the key child.

```
  SELECT sum(cardinality(v::MAP(VARCHAR,VARCHAR))) FROM t;   -- 200k x 6 keys
  before: 1.726s   after: 0.241s   (7.2x)
```

Only OBJECT-shaped variants take the new path. A VARIANT built from a MAP is stored as an ARRAY of key/value structs (variant_typeof(MAP{}::VARIANT) is 'ARRAY(0)'), so those keep using ConvertVariantToList exactly as before; a non-VARCHAR key type still uses the Value fallback, which can cast the keys.

Output is byte-identical to the previous implementation across objects, nested values, NULL rows, empty objects, NULL values, non-VARCHAR value types, duplicate keys, MAP-shaped variants and non-object variants.

(cherry picked from commit 606b0a79903e44569f3c84d7b1e267c131036421, adapted to main: array_ptr/make_unsafe_uniq_array_uninitialized for the nested-data buffer, FlatVector::ValidityMutable, FlatVector::GetDataMutable, and the split-out map_vector.hpp header.)

### DIFF
--- a/src/function/cast/variant/from_variant.cpp
+++ b/src/function/cast/variant/from_variant.cpp
@@ -2,6 +2,7 @@
 #include "duckdb/common/vector/constant_vector.hpp"
 #include "duckdb/common/vector/flat_vector.hpp"
 #include "duckdb/common/vector/list_vector.hpp"
+#include "duckdb/common/vector/map_vector.hpp"
 #include "duckdb/common/vector/shredded_vector.hpp"
 #include "duckdb/common/vector/string_vector.hpp"
 #include "duckdb/common/vector/struct_vector.hpp"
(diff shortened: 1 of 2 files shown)


wmTJc9IK0Q:
This change is still valid.

Issue #24091, opened 2026-07-23
Missing arrow type extension for VARIANT <-> arrow.parquet.variant


cmettler:
### What happens?

Both import and export of an arrow parquet variant is not supported.
Export fails with not implemented. 
Import of an canoncial arrow parquet variant is not converted to a Duckdb VARIANT

### To Reproduce

```python
import duckdb
import pyarrow as pa

# 1) EXPORT: raises NotImplementedException
duckdb.sql("SELECT 42::VARIANT AS v").to_arrow_table()

# 2) IMPORT: silently degrades to the bare storage struct
metadata = pa.array([b"\x01\x00\x00"], pa.binary())   # spec-encoded: v1 empty dictionary
value = pa.array([b"\x0c\x2a"], pa.binary())          # int8 42
storage = pa.StructArray.from_arrays(
    [metadata, value],
    fields=[pa.field("metadata", pa.binary(), nullable=False),
            pa.field("value", pa.binary(), nullable=False)])
schema = pa.schema([
    pa.field("v", storage.type, nullable=True,
             metadata={"ARROW:extension:name": "arrow.parquet.variant"}),
])
batch = pa.RecordBatch.from_arrays([storage], schema=schema)
print(duckdb.sql("SELECT v, typeof(v) FROM batch").fetchall())
# [({'metadata': b'\x01\x00\x00', 'value': b'\x0c*'}, 'STRUCT(metadata BLOB, "value" BLOB)')]
#   ^ expected: a VARIANT value 42
```

### OS:

Windows

### DuckDB Version:

1.5.5

### DuckDB Client:

Python

### Full Name:

Christoph Mettler

### Affiliation:

Ethenea

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

No - Other reason (please specify in the issue body)

PR #24157, opened 2026-07-25
Support VARIANT over the Arrow interface (arrow.parquet.variant)

cmettler:
closes (https://github.com/duckdb/duckdb/issues/24091)
- Fix the Arrow appender to finalize children by the extension type's storage type — it walked the logical type's children and crashed on any registered extension type with nested storage.
- Fix Arrow import to resolve registered extension types to their DuckDB type instead of silently degrading them to the bare storage type.
- Register the canonical arrow.parquet.variant extension type over struct<metadata, value>, so VARIANT columns round-trip in bot directions. Values convert through the parquet extension's variant functions


src/common/arrow/arrow_type_extension.cpp
```diff
@@ -509,6 +514,177 @@ struct ArrowGeometry {
...
+			throw InvalidInputException(
+			    "arrow.parquet.variant column must have a struct<metadata, value> storage type, got %lld children",
+			    schema.n_children);
+		}
+		vector<shared_ptr<ArrowType>> children;
+		for (idx_t i = 0; i < 2; i++) {
+			children.push_back(ArrowType::GetTypeFromSchema(context, *schema.children[i]));
+		}
+		return make_uniq<ArrowType>(LogicalType::VARIANT(), make_uniq<ArrowStructInfo>(std::move(children)));
+	}
```
zeroshade, line 637: this rejects valid shredded variant schemas, and the canonical spec also allows a *fully* shredded variant to be of the form `struct<metadata, typed_value>` where `typed_value` is not a binary type. If you're going to reject shredded variants, it needs to be done safely instead of simply assuming that 2 children will always be `struct<metadata, value>`.

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -509,6 +514,177 @@ struct ArrowGeometry {
...
+		}
+	}
+
+	//! Arrow -> DuckDB: concatenate each row's metadata and value bytes (the self-delimiting Variant
+	//! binary form) and decode through `variant_bytes_to_variant`.
+	static void ArrowToDuck(ClientContext &context, Vector &source, Vector &result, idx_t count) {
+		source.Flatten();
+		auto &entries = StructVector::GetEntries(source);
+		Vector &metadata = entries[0];
+		Vector &value = entries[1];
```
zeroshade: Canonically by the spec, the fields may appear in any order. As such, they should be resolved by *name* rather than *position*. The indexes should get resolved during schema binding and then normalize the storage before converting.

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -509,6 +514,177 @@ struct ArrowGeometry {
...
+	                                     const ArrowSchemaMetadata &schema_metadata) {
+		if (schema.n_children != 2) {
+			throw InvalidInputException(
+			    "arrow.parquet.variant column must have a struct<metadata, value> storage type, got %lld children",
+			    schema.n_children);
+		}
+		vector<shared_ptr<ArrowType>> children;
+		for (idx_t i = 0; i < 2; i++) {
+			children.push_back(ArrowType::GetTypeFromSchema(context, *schema.children[i]));
+		}
```
zeroshade, line 632: This will fail on dictionary-encoded metadata. `GetTypeFromSchema` does not check the `schema.dictionary`, this happens only in the `GetArrowLogicalType` path which is dictionary aware. As a result, this would treat a `dictionary<int8, binary>` metadata field as `TINYINT` while the destination remains `BLOB` causing an incompatible conversion.

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -509,6 +514,177 @@ struct ArrowGeometry {
...
+		schema.children = root_holder.nested_children_ptr.back().data();
+
+		const char *child_names[] = {"metadata", "value"};
+		for (idx_t i = 0; i < 2; i++) {
+			auto &child = *schema.children[i];
+			child.format = "z";
+			child.name = child_names[i];
+			child.flags = ARROW_FLAG_NULLABLE;
+			child.release = release_child;
+		}
```
zeroshade, line 584: the metadata field *must* be non-nullable according to the spec. This will mark both fields as nullable, strict consumers could reject the DuckDB exported variant schemas

zeroshade (commented):
Aside from the comments below, the tests are only doing round-trips of duckdb producing AND consuming. Could we get hand-built tests that construct explicit Arrow C Data structures and attempt to import them into DuckDB to cover the following scenarios?

* `metadata` + `typed_value` (if shredded variants will be supported, otherwise there should probably be explicit comments and TODO's for adding support for shredding later on)
* `metadata` + `value` + `typed_value` (same caveat for shredded variants as above)
* reversed field order of `value`, `metadata`
* dictionary-encoded `metadata`
* run-end-encoded `metadata`
* verifying that `metadata` fields exported by duckdb are marked non-nullable

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -509,6 +514,177 @@ struct ArrowGeometry {
...
+		root_holder.nested_children.back().resize(2);
+		root_holder.nested_children_ptr.emplace_back();
+		root_holder.nested_children_ptr.back().push_back(&root_holder.nested_children.back()[0]);
+		root_holder.nested_children_ptr.back().push_back(&root_holder.nested_children.back()[1]);
+		schema.children = root_holder.nested_children_ptr.back().data();
+
+		const char *child_names[] = {"metadata", "value"};
+		for (idx_t i = 0; i < 2; i++) {
+			auto &child = *schema.children[i];
+			child.format = "z";
```
evertlammerts: the format of `metadata` and `value` might be Binary, LargeBinary or BinaryView (`z`, `Z`, `vz`). The appender will pick one based on the `arrow_large_buffer_size` and `arrow_output_version` settings, so the output might be encoded differently than what the schema says. Also see `SetArrowFormat` in src/common/arrow/arrow_converter.cpp

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -509,6 +514,177 @@ struct ArrowGeometry {
...
+
+	//! Binds one of the parquet extension's variant conversion functions by name (triggering an extension
+	//! autoload where enabled) and runs it over `input` into `result`.
+	static void ExecuteConversion(ClientContext &context, const char *function_name, Vector &input, idx_t count,
+	                              Vector &result) {
+		vector<unique_ptr<Expression>> args;
+		args.push_back(make_uniq<BoundReferenceExpression>(input.GetType(), 0ULL));
+		FunctionBinder binder(context);
+		ErrorData error;
+		auto bound = binder.BindScalarFunction(DEFAULT_SCHEMA, function_name, std::move(args), error);
```
evertlammerts: This won't bind when the conversion happens outside of a valid context. There are entry points (you can use `duckdb_result_arrow_array` for a test / repro I think) where the context is _invalid_. This is a known issue with the arrow extension API, and on our list to solve.

Instead of calling `variant_to_parquet_variant` by name through the binder it's probably better to expose `ToParquetVariant` (in extension/parquet/writer/variant/convert_variant.cpp:900) and call that. If you make sure to always pass in a non-shredded vector (which `StorageType()` does) then it should convert as the schema specifies.

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -509,6 +514,177 @@ struct ArrowGeometry {
...
+	//! Maps a tagged schema back: the VARIANT logical type, with type info describing the storage struct
+	//! for the reader's buffer walk (children parsed from the actual schema, so binary/large binary/view
+	//! layouts from foreign writers are all accepted).
+	static unique_ptr<ArrowType> GetType(ClientContext &context, const ArrowSchema &schema,
+	                                     const ArrowSchemaMetadata &schema_metadata) {
+		if (schema.n_children != 2) {
+			throw InvalidInputException(
+			    "arrow.parquet.variant column must have a struct<metadata, value> storage type, got %lld children",
+			    schema.n_children);
+		}
```
evertlammerts, line 624: You should also check the parent's format here (must be `+s`). Right now `+us:0,1` also makes it through.

evertlammerts (commented):
Thanks @cmettler, and @zeroshade for the review. I've left a few other comments.

+1 on adding generated Arrow C Data structures with tests for import.

evertlammerts (changes requested):

src/common/arrow/arrow_type_extension.cpp
cmettler: @zeroshade @evertlammerts Thanks for the review, I'll take a look.

src/common/arrow/arrow_type_extension.cpp
cmettler: Do you prefer moving arrow.parquet.variant registration out of ArrowTypeExtensionSet::Initialize into the parquet extension's load or moving ToParquetVariant and ParquetVariantConversion::ConvertBinary to core ?

src/function/table/arrow_conversion.cpp
```diff
@@ -843,6 +843,11 @@ void ArrowToDuckDBConversion::ColumnArrowToDuckDB(Vector &vector, ArrowArray &ar
 		if (arrow_type.extension_data->arrow_to_duckdb) {
 			// Convert the storage and then call the cast function
 			Vector input_data(arrow_type.extension_data->GetInternalType());
```
evertlammerts, line 845: Child vector can be of arbitrary size, so this vector should have capacity set I think, just as in `ArrowAppendData::AppendChild`:
```suggestion
			Vector input_data(arrow_type.extension_data->GetInternalType(), MaxValue<idx_t>(size, STANDARD_VECTOR_SIZE));
```

It's also worth to test with large nested types, e.g. `select [x::VARIANT FOR x IN range(5000)]`

test/arrow/arrow_roundtrip.cpp
```diff
@@ -1,6 +1,19 @@
...
+#include "duckdb/common/arrow/arrow_type_extension.hpp"
+#include "duckdb/common/arrow/schema_metadata.hpp"
+#include "duckdb/function/table/arrow/arrow_duck_schema.hpp"
+#include "duckdb/function/table/arrow/arrow_type_info.hpp"
+#include "duckdb/common/identifier.hpp"
+
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/parser/parsed_data/create_scalar_function_info.hpp"
+
+#include <atomic>
```
evertlammerts, line 16:
```suggestion
#include "duckdb/common/atomic.hpp"
```

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -523,6 +526,239 @@ struct ArrowGeometry {
...
+			    "arrow.parquet.variant column must have a 'metadata' and a 'value' field, got %lld children",
+			    schema.n_children);
+		}
+		vector<shared_ptr<ArrowType>> children;
+		child_list_t<LogicalType> storage_children;
+		for (idx_t i = 0; i < NumericCast<idx_t>(schema.n_children); i++) {
+			// GetArrowLogicalType, not GetTypeFromSchema: a dictionary-encoded child carries its value
+			// type in schema.dictionary, which only this entry point resolves.
+			children.push_back(ArrowType::GetArrowLogicalType(context, *schema.children[i]));
+			storage_children.emplace_back(schema.children[i]->name, LogicalType::BLOB);
```
evertlammerts, line 631: Maybe verify that the child type is binary here, something like:
```suggestion
			auto child_type = children.back()->GetDuckType(true);
			if (child_type.id() != LogicalTypeId::BLOB) {
				throw InvalidInputException("arrow.parquet.variant field '%s' must be a binary type, got %s", schema.children[i]->name, child_type.ToString());
			}
			storage_children.emplace_back(schema.children[i]->name, LogicalType::BLOB);
```

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -523,6 +526,239 @@ struct ArrowGeometry {
...
+		auto &entries = StructVector::GetEntries(source);
+		Vector &metadata = entries[metadata_idx];
+		Vector &value = entries[value_idx];
+		metadata.Flatten();
+		value.Flatten();
+
+		// The binary decoder does not consult validity, so NULL rows are substituted with the minimal
+		// valid encoding (metadata v1 with an empty dictionary + a Variant null value) and the SQL NULL
+		// is re-applied on the result afterwards.
+		static constexpr const char MINIMAL_NULL_VARIANT[] = "\x01\x00\x00\x00";
```
evertlammerts, line 724: `ParquetBinaryVariantSource::Emit()` checks:
```cpp
		if (!blob_format.validity.RowIsValid(blob_format.sel->get_index(row))) {
			return true;
		}
```

and `BuildVariant` does:
```cpp
		bool is_null = source.Emit(row, builder);
		blob_writer.WriteValue(string_t(builder.blob.data(), NumericCast<uint32_t>(builder.blob.size())));
		if (is_null) {
			//! SPEC: If a Variant is missing in a context where a value is required, readers must return a Variant null
			FlatVector::SetNull(result, row, true);
		}
```

So I think `MINIMAL_NULL_VARIANT[]` is not needed, and I think you can use `FlatVector::SetNull(blob, i, true);` instead of doing it by hand on lines 734-736 below (and remove `is_null` and `has_nulls`, and the whole `if (has_nulls) { result.Flatten(); ... SetNull(result, i, true) ... }` block on 751-758).

evertlammerts (commented):
Thanks for addressing the feedback @cmettler! I left a couple new comments, but overal it looks good.

About the moving of the conversion functions into core: I think it's reasonable like this, but maybe we can slice out just the conversion functions, and keep the rest in the extension (or at least see if we can put `VariantBytesToVariantFunction` and `GetBytesToVariantFunction` in e.g. convert_variant.cpp, since these are functions registered by the parquet extension).

evertlammerts (changes requested):
Thanks for addressing the feedback @cmettler! I left a couple new comments, but overal it looks good.

About the moving of the conversion functions into core: I think it's reasonable like this, but maybe we can slice out just the conversion functions, and keep the rest in the extension (or at least see if we can put `VariantBytesToVariantFunction` and `GetBytesToVariantFunction` in e.g. convert_variant.cpp, since these are functions registered by the parquet extension).

src/common/arrow/arrow_type_extension.cpp
```diff
@@ -523,6 +526,239 @@ struct ArrowGeometry {
...
+		for (idx_t i = 0; i < source_children.size(); i++) {
+			if (source_children[i].first == "metadata") {
+				metadata_idx = i;
+			} else if (source_children[i].first == "value") {
+				value_idx = i;
+			}
+		}
+		if (metadata_idx == DConstants::INVALID_INDEX || value_idx == DConstants::INVALID_INDEX) {
+			throw InternalException("arrow.parquet.variant storage struct is missing its metadata/value fields");
+		}
```
zeroshade, line 714: `GetType` already validates the indexes but this is going to re-resolve the indexes on every chunk. Can we cache the metadata/value (and later typed_value) indexes on the `ArrowTypeExtensionData` object and then just reuse that instead of resolving the indexes on every chunk?

At minimum, the exception here is unreachable given that `GetType` already does the validation and would have thrown an exception above this before getting here.

zeroshade (approved):
Thanks for addressing my comments @cmettler, I just one nitpick that is by no means blocking at all. This is looking great to me!

cmettler:
Rebased onto `v2.0-cyanoptera`; review comments addressed.

- Sized the storage vector for nested children — this one was a real out-of-bounds write: revert just that and `SELECT [x::VARIANT FOR x IN range(5000)]` segfaults. Now covered by a test.
- Storage fields must be binary; dropped the null sentinel (the decode already returns a Variant null for an invalid blob); `duckdb/common/atomic.hpp`.
- `variant_bytes_to_variant` moved back into the parquet extension — core keeps only the conversions.
- @zeroshade metadata/value resolution is O(1) now and the unreachable throw is gone. Caching them on `ArrowTypeExtensionData` isn't reachable though: `cast_arrow_duck_t` is a plain function pointer, so the callback can't carry state. Happy to add a slot for it if you'd prefer.

FWIW `Linux Release Tests` and the cloud extension build also fail on an unmodified `v2.0-cyanoptera` (I ran the base through the same CI), so they look pre-existing rather than from this PR.

evertlammerts (approved):
LGTM! Thanks @cmettler !

test/arrow/arrow_roundtrip.cpp
```diff
@@ -487,3 +428,710 @@ TEST_CASE("Test Parquet Files round-trip", "[arrow][.]") {
...
+}
+
+//! Scans the stream and expects the BIND to refuse it with `expected_error` (a refusal may surface as
+//! an error result or as an exception escaping the relation, depending on where the bind runs).
+void ExpectVariantScanError(Connection &con, ArrowArrayStream &stream, const string &expected_error) {
+	auto params = ArrowTestHelper::ConstructArrowScan(stream);
+	bool refused = false;
+	string error_message;
+	try {
+		auto result = con.TableFunction("arrow_scan", params)->Execute();
```
evertlammerts, line 735:
```suggestion
		auto result = con.TableFunction("arrow_scan", {}, {}, params)->Execute();
```

This should make the test pass.

evertlammerts (changes requested):
@cmettler CI is still failing. I think the below change will fix it.

PR #24158, opened 2026-07-25
test harness: LOCAL_{TEMP,DATA}_DIR, add --env-passthrough, remote temp-dir fixes

benfleis:
test harness: remote temp-dir root, LOCAL_* dirs, --env-passthrough

Makes the --temp-dir-* family usable against a remote root (az://, s3://) and folds its knobs
into the standard test-config option table, so each also takes a `DUCKDB_TEST_<NAME>` env var
and a config-file key.

Local variants

New: `LOCAL_TEMP_DIR` is a guaranteed-local temp dir within tests. The underlying machinery
allows HOME to function correctly in local FS even when working with remote temp dirs.2 common cases:
- If `--temp-dir-root` is local -> both local, and TEMP_DIR = LOCAL_TEMP_DIR
- If `--temp-dir-root` is remote -> LOCAL_ initiated in local fs (def. duckdb_unittest_tempdir)
- all roots can be specified; rare to specify `local` variants, but possible
- DATA_DIR has same LOCAL_ alternate (def. data), useful for certain file comparison checks

Lifecycle

- A run reclaims only what it created, at every level including the run root: a root merely
  adopted -- a reused `--run-id`, or `$ROOT` under `--temp-dir-run-id off` -- is retained.
  Removal is recursive, so otherwise a successful run deletes a caller-supplied root outright,
  or a co-located batch's live tree.
- Co-location is therefore safe only when the driver creates the shared run root itself; a
  batch that created it still reaps it.
- Dropped `--temp-dir-create`. Nothing consumed it, and `never` never cohered: it could not
  create the [TEST_ID] leaf, could not be satisfied under a generated run-id, and still deleted
  the root it refused to create. Creation is unconditional mkdir -p; the created-levels
  bookkeeping is the property that actually mattered.
- A remote root is passthrough only -- unittest holds no credentials, so it never creates or
  reaps one.

`--env-passthrough NAME` pre-registers a caller's env var for `{NAME}` substitution in every
test, the treatment `TEMP_DIR`/`DATA_DIR` already get. A named-but-absent var fails the whole
invocation at startup, unlike `require-env`'s per-test skip. Names the runner resolves for
itself are reserved, including the `{KEYWORD}` substitutions: env substitution runs first, so
an unreserved `{TEST_DIR}` would be silently shadowed run-wide.

Test events `begin` and `end` carry the temp dir, so a consumer identifies the emitting
invocation directly instead of inferring it from the subprocess it launched. It is snapshotted
at `begin`, so a body that rewrites `TEMP_DIR` cannot make the two disagree.

Fixes

- `TEMP_DIR_ABSOLUTE` was wrong against a remote root, joining a scheme'd URI onto the cwd.
- HOME derived from `$ROOT` and hit the same join; it resolves against the local tree now,
  rather than crashing or leaking to the ambient `$HOME`.
- `PrepareTempDir` reports a failed mkdir/rmdir through its error out-param instead of letting
  the exception terminate the process.
- Roots are normalized when parsed (`.`/`..` fold, `TEMP_DIR_ROOT` reports the resolved form);
  an empty root is rejected.

Notes

- `TEMP_DIR_ROOT` and the `LOCAL_*` roots can come from the ambient environment, e.g. to
  default to `/tmp/duckdb/` instead of `./duckdb_unittest_tempdir`.
- test/py/test_temp_dir_contract.py covers what no .test file can observe: what survives on
  disk after a run, and what a run may create on an error path. Standalone for now.

### DIFF
--- a/scripts/test_storage_compatibility.py
+++ b/scripts/test_storage_compatibility.py
@@ -204,14 +204,14 @@ def execute_test(i, test):
         with open(worker_test_config, 'w') as f:
             json.dump(config_contents, f)
         # The durable db (initial_db) lives outside test_temp_dir, so destroy is indifferent;
-        # pin test_temp_dir as the exact base with both levels off (replaces --test-temp-dir).
+        # pin test_temp_dir as the exact root with both levels off (replaces --test-temp-dir).
         # The enclosing TemporaryDirectory cleans everything up, so destroy=never just retains
         # until then.
         unittest_cmd = [
             unittest_program,
             '--test-config',
             worker_test_config,
-            '--temp-dir-base',
+            '--temp-dir-root',
             test_temp_dir,
             '--temp-dir-run-id',
             'off',
--- a/scripts/test_zero_initialize.py
+++ b/scripts/test_zero_initialize.py
@@ -141,12 +141,12 @@ for test in test_list:
     print(f"Running test {test}")
     clear_directories(test_dirs)
     # TEMP_DIR must BE the given dir exactly and survive for the post-run byte diff below,
-    # so pin the base with both levels off and destroy=never (the dirs are pre-rmtree'd
+    # so pin the root with both levels off and destroy=never (the dirs are pre-rmtree'd
     # above, so the default create=on-absent is fine). --database-destroy off keeps the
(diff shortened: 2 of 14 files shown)


lnkuiper:
Superseded by https://github.com/duckdb/duckdb/pull/24355

PR #24159, opened 2026-07-25
Track full source span of parsed nodes, not just offset

Maxxen:
This PR introduces a ~~`Span`~~ `QueryLocation` class which replaces a lot of the `optional_idx query_location` fields that various AST nodes previously carried around. This enables us to print slightly more precise error messages right now, but will also allow us to improve error messages/diagnostics much more in the future.

A `QueryLocation` is basically just a `{offset uint32_t, length uint32_t}` tuple that represent some span of source code, with some conveniences methods to merge or detect invalid spans. 

Using `QueryLocation`s instead of a single `query_location` offset makes the location of internal nodes much clearer to express. E.g. in the case of composite expressions, the location of `a + b` is simply the span containing both the span of `a` and the span of `b` (although we currently don't do this for binary operators, but I'll revisit that in a future PR).

As a first step, this PR makes the formatting of error messages take advantage of this change by printing out multi-character underlines when the error query location has a valid length, e.g:

```sql
select struct_extract((1, 2), add(x,2)) from range(0, 10) r(x);

-- We used to print this:
Binder Error:
The '"index"' argument in function 'struct_extract' must be a constant expression                                                                           
                                                                                                                                                            
LINE 1: select struct_extract((1, 2), add(x,2)) from range(0, 10) r(x);  
                                         ^

-- We now print this: (notice the underlined "add(x,2)")
Binder Error:
The '"index"' argument in function 'struct_extract' must be a constant expression

LINE 1: select struct_extract((1, 2), add(x,2)) from range(0, 10) r(x);
                                      ^^^^^^^^
```

### DIFF
--- a/src/common/error_data.cpp
+++ b/src/common/error_data.cpp
@@ -1,6 +1,7 @@
 #include "duckdb/common/error_data.hpp"
 
 #include "duckdb/common/exception.hpp"
+#include "duckdb/common/span.hpp"
 #include "duckdb/common/string_util.hpp"
 #include "duckdb/common/to_string.hpp"
 #include "duckdb/common/types.hpp"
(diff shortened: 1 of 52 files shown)


Mytherin:
Thanks for the PR! Looks great - I just think `Span` is kind of an overloaded term to use for such a specific use case. Maybe we can call this `QueryLocation` or so instead?

Maxxen:
@Mytherin done!

Mytherin:
Thanks!

PR #5184, opened 2022-11-03
Update OpenSSL to 1.1.1s

sjaenick:
Update OpenSSL to 1.1.1s release for GitHub Python workflow.

(Unable to test this locally, please verify)

PR #24162, opened 2026-07-26
Internal #5184: Nested Equality

hawkfish:
* Convert Boolean to enum.
* Defer equality decisions until all components have been checked.
* Update test for more distinct values.
* Update test for nested <> semantics.


Mytherin:
Thanks!

PR #21628, opened 2026-03-25
Issue #21592: Window Self-Join Framing

hawkfish:
* Don't apply the optimisation for ROWS framing

PR #24164, opened 2026-07-26
Add value-level tests for window self-join ROWS framing (close #21592)

### LINKED ISSUE #21592: WindowSelfJoinOptimizer produces wrong results for ROWS frames
### What happens?

`WindowSelfJoinOptimizer` (new in v1.5.0) incorrectly replaces window aggregates with `GROUP BY + INNER JOIN` when the window has an explicit `ROWS` frame. The optimization is only valid for constant (full-partition) frames, but `CanOptimize()` does not check `w_expr.start` / `w_expr.end`, so it applies to any window without `ORDER BY` — including `ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`, which should produce cumulative results.

This causes **silently wrong results**: every row in a partition gets the same full-partition aggregate instead of the correct per-row cumulative value.

### To Reproduce

```sql
SELECT seq,
       sum(seq) OVER (PARTITION BY grp ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS cumsum
FROM (VALUES (1,'A'),(2,'A'),(3,'A'),(1,'B'),(2,'B')) AS t(seq, grp)
ORDER BY grp, seq;
```

Actual output (v1.5.0 and v1.5.1):
```
┌───────┬────────┐
│  seq  │ cumsum │
│ int32 │ int128 │
├───────┼────────┤
│     1 │      6 │
│     2 │      6 │
│     3 │      6 │
│     1 │      3 │
│     2 │      3 │
└───────┴────────┘
```

Expected output (correct in v1.4.4):
```
┌───────┬────────┐
│  seq  │ cumsum │
│ int32 │ int128 │
├───────┼────────┤
│     1 │      1 │
│     2 │      3 │
│     3 │      6 │
│     1 │      1 │
│     2 │      3 │
└───────┴────────┘
```

`EXPLAIN` confirms the optimizer replaces `WINDOW` with `HASH_GROUP_BY + HASH_JOIN`:
```sql
EXPLAIN SELECT seq,
       sum(seq) OVER (PARTITION BY grp ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)
FROM (VALUES (1,'A'),(2,'A'),(3,'A'),(1,'B'),(2,'B')) AS t(seq, grp)
ORDER BY grp, seq;
```
Shows `HASH_JOIN` + `HASH_GROUP_BY` instead of `WINDOW` operator.

### Root cause

`WindowSelfJoinOptimizer::CanOptimize` (`src/optimizer/window_self_join.cpp:153`) checks `w_expr.orders.empty()` but does not check `w_expr.start` or `w_expr.end`. It assumes "no ORDER BY = frame covers entire partition", which is true for `RANGE` frames but false for `ROWS` frames (`CURRENT_ROW_ROWS` means literally the current row, regardless of peers).

For comparison, `WindowConstantAggregator::CanAggregate` (`src/function/window/window_constant_aggregator.cpp:125`) correctly validates frame boundaries — it only allows `UNBOUNDED_PRECEDING`/`UNBOUNDED_FOLLOWING` and `CURRENT_ROW_RANGE` (without ORDER BY).

### OS:

macOS 26.3.1, arm64

### DuckDB Version:

1.5.0, 1.5.1 (works correctly on 1.4.4)

### DuckDB Client:

CLI

### Full Name:

Ilia Ablamonov

### Affiliation:

Rick.ai

Vamsi-klu:
## Problem

`WindowSelfJoinOptimizer` used to rewrite `ROWS … CURRENT ROW` running aggregates into a full-partition self-join, producing constant totals instead of cumulative values. That was fixed in #21628; the existing test only checked the plan shape.

## Solution

Add value-level assertions for the issue repro next to the existing ROWS plan check so a regression would fail on wrong numbers, not only on EXPLAIN text.

Related to #21592 (fix already in #21628).

### Test

```bash
build/reldebug/test/unittest test/sql/optimizer/test_window_self_join.test
```


hawkfish (approved):
Good catch - LGTM.

Vamsi-klu:
Merged current main in `3de4e5b`. The red Summary on the previous head came only from cancelled jobs rather than real failures, so this should come back green. Test-only change, approved since 30 July, ready whenever a committer has a moment.

hawkfish:
@Vamsi-klu Can you change it from "Draft" to "Ready to review"? It's kind of stuck in the queue right now. (And yes its been reviewed, but that was me getting ahead of the system...)

Vamsi-klu:
@hawkfish , thanks for the comment. i converted for "ready for the review"

PR #12240, opened 2024-05-25
Fix #12225: revert OVERWRITE_OR_IGNORE to previous behavior, move new behavior to OVERWRITE flag

Mytherin:
Fixes #12225

This PR reverts `OVERWRITE_OR_IGNORE` to the previous behavior where files are overwritten if they exist, or ignored (left alone) if they do not. The new behavior introduced in https://github.com/duckdb/duckdb/pull/11787 is instead moved to a new flag (`OVERWRITE`). The new flag is more explicit in what it does, and is also explicitly disabled for remote file systems (e.g. S3) where we don't support removing files yet.

Issue #12225, opened 2024-05-24
Hive partitioned append/write 0.10.3 


daniarleagk:
### What happens?

Before version 0.10.3, you could enforce partitioned append to a Hive partitioned table by specifying the FILENAME_PATTERN 
property and setting OVERWRITE_OR_IGNORE to true. 
However, since the recent changes (as documented in this GitHub pull request https://github.com/duckdb/duckdb/pull/11787),
 the behavior has been modified. Now I am experiencing breaking changes in my applications.  Is there any alternative approach for partitioned append? The code below  illustrates the problem.

### To Reproduce

```sql
create or replace temp view t_val_v  as 
select *
from ( 
    values 
    ('p_1', 'key_1', 'val_1'),
    ('p_2', 'key_1', 'val_1'),
    ('p_3', 'key_1', 'val_1')
) as t_val(p_key, key, val);
copy (
    select *
    from  t_val_v
) to '/home/da/output_tmp/t_val' (
format parquet,
PARTITION_BY (p_key),
FILENAME_PATTERN "pv1_{i}",
OVERWRITE_OR_IGNORE false
);
copy (
    select *
    from  t_val_v
) to '/home/da/output_tmp/t_val' (
format parquet,
PARTITION_BY (p_key),
FILENAME_PATTERN "pv2_{i}",
OVERWRITE_OR_IGNORE false
);
```

output before. changes (version 0.10.2):

```bash
before 0.10.3 
~/output_tmp/t_val/p_key=p_1$ ls -la
total 16
drwxr-xr-x 2 da da 4096 May 24 10:46 .
drwxr-xr-x 5 da da 4096 May 24 10:44 ..
-rw-r--r-- 1 da da  367 May 24 11:49 pv1_0.parquet
-rw-r--r-- 1 da da  367 May 24 11:49 pv2_0.parquet
```
output with version 0.10.3 (removes all old files)

```bash
~/output_tmp/t_val/p_key=p_1$ ls -la
total 12
drwxr-xr-x 2 da da 4096 May 24 11:51 .
drwxr-xr-x 5 da da 4096 May 24 10:44 ..
-rw-r--r-- 1 da da  367 May 24 11:51 pv2_0.parquet
```

### OS:

windows, ubuntu

### DuckDB Version:

0.10.3

### DuckDB Client:

python

### Full Name:

Daniar Achakeev

### Affiliation:

HMS Analytical Software GmbH

[...]

PR #24165, opened 2026-07-26
Regression test OVERWRITE vs OVERWRITE_OR_IGNORE for partitioned COPY

### LINKED ISSUE #10282: OVERWRITE_OR_IGNORE overwrites files instead of directories during partitioned writes
### What happens?

The duckdb docs state the following.

> [Overwriting](https://duckdb.org/docs/data/partitioning/partitioned_writes.html#overwriting)
By default the partitioned write will not allow overwriting existing directories. Use the OVERWRITE_OR_IGNORE option to allow overwriting an existing directory.

However, this is incorrect. Duckdb only overwrites files in a directory. This results in a level of non-determinism when writing to the an existing partition.

### To Reproduce

The easiest way to reproduce this is just performing a partitioned write with a sufficiently large dataset that threads has an effect. Set your number of threads to something greater than 1 (ie: 4). Write the dataset to a hive partitioned parquet directory. Then wait a minute or two. Then set threads to 1 and rewrite the dataset. The do something like `ls -l data-lake/**` and observe the timestamps on the writes. The `data_0.parquet` file (ie: first thread) will have a newer timestamp and be larger than the other 3 files.

Sorry I can't provide a slightly better example. Uploading a dataset becomes the issue. Hopefully below illustrates an idea.

```python
    duckdb.sql(
        """
        set threads = 4;
        copy (
            select
                date,
                account,
                count(*)
            from
                read_parquet(
                    's3://data-lake/warehouse/billing_platform.db/nodeboostbilling/**/*.parquet',
                    hive_partitioning = 1
                )
            where
                date::date >= '2023-11-07'
                and date::date < '2023-11-09'
            group by date, account
        ) to
        'data-lake/warehouse/billing_platform.db/nodeboostbilling_daily/'
        (
            format PARQUET,
            partition_by ('date'),
            overwrite_or_ignore
        )
        """
    )
)

time.sleep(60)

    duckdb.sql(
        """
        set threads = 1;
        copy (
            select
                date,
                account,
                count(*)
            from
                read_parquet(
                    's3://data-lake/warehouse/billing_platform.db/nodeboostbilling/**/*.parquet',
                    hive_partitioning = 1
                )
            where
                date::date >= '2023-11-07'
                and date::date < '2023-11-09'
            group by date, account
        ) to
        'data-lake/warehouse/billing_platform.db/nodeboostbilling_daily/'
        (
            format PARQUET,
            partition_by ('date'),
            overwrite_or_ignore
        )
        """
    )
)
```

```console
❯ ls -l notebooks/data-lake/warehouse/billing_platform.db/nodeboostbilling_daily/**
total 24
-rw-r--r--  1 taylorbeever  staff  613 Jan 19 17:51 data_0.parquet
-rw-r--r--  1 taylorbeever  staff  613 Jan 19 17:51 data_1.parquet
-rw-r--r--  1 taylorbeever  staff  559 Jan 19 17:51 data_2.parquet
❯ ls -l notebooks/data-lake/warehouse/billing_platform.db/nodeboostbilling_daily/**
total 24
-rw-r--r--  1 taylorbeever  staff  746 Jan 19 17:52 data_0.parquet
-rw-r--r--  1 taylorbeever  staff  613 Jan 19 17:51 data_1.parquet
-rw-r--r--  1 taylorbeever  staff  559 Jan 19 17:51 data_2.parquet
```

### OS:

MacOs

### DuckDB Version:

duckdb-0.9.3.dev2934

### DuckDB Client:

Python

### Full Name:

Taylor Beever

### Affiliation:

QuickNode

### Have you tried this on the latest `main` branch?

I have tested with a main build

- [X] Yes, I have

Vamsi-klu:
## What the issue is

Partitioned `COPY` with fewer threads than a previous write can leave **orphan `data_N.parquet` files** (and untouched partition directories) when the wrong overwrite flag is used. Readers that glob `**/*.parquet` then mix generations — classic lake “I overwrote the table but old rows came back” failure.

Historical reports (including on 1.4.x) used `OVERWRITE_OR_IGNORE`. After [#12240](https://github.com/duckdb/duckdb/pull/12240) the product split is intentional:

| Mode | Behavior |
|------|----------|
| default | Error if target directory non-empty |
| **`OVERWRITE`** | Recursive clear + rewrite — correct full replace |
| **`OVERWRITE_OR_IGNORE`** | **No directory clear**; clobber same filenames only; leave orphans |

So the bug is often **mis-selected mode** plus missing regression coverage that multi-thread → fewer-thread rewrite with `OVERWRITE` actually drops extras. This PR locks the correct `OVERWRITE` behavior with a multi-thread test and documents (in-test) that `OVERWRITE_OR_IGNORE` deliberately leaves untouched partitions.

Fixes #10282

## Why I solved it that way

- **Do not change `OVERWRITE_OR_IGNORE` to delete again.** That was a deliberate split for append-style workflows (`#12225` / pattern files). Silently flipping it would break those users.
- **Do not attempt remote S3 directory purge in this PR** — large infra surface; local FS semantics are what most CI and the original repro exercise.
- **Ship a hard regression test** for the multi-thread stale-file scenario under `OVERWRITE`, plus an explicit `OVERWRITE_OR_IGNORE` case that *expects* leftover partitions so the intentional behavior cannot rot into “half clear.”

Note: this branch is **test-only** on tip. Runtime `CheckDirectory` / `OVERWRITE` recursive remove already does the right thing on main; the gap was coverage and a clear statement of which flag to use for idempotent rebuilds. If docs still say OR_IGNORE “overwrites directories,” a docs-only follow-up can fix wording — the behavioral contract is asserted here in SQL.

## How I did it

Extended `test/sql/copy/partitioned/hive_partitioning_overwrite.test`:

1. **Multi-thread write** (`SET threads=4`) of 100k rows partitioned by `p` → assert ≥2 parquet files and 50k rows per partition.
2. **Single-thread OVERWRITE** of 100 rows → assert exactly 2 files left, 50 rows per partition (no orphan `data_N` from the 4-thread write).
3. **Partition directory drop:** write 3 partitions, OVERWRITE with only `p=0` → one file, only partition 0.
4. **OR_IGNORE contrast:** write 3 partitions, `OVERWRITE_OR_IGNORE` with only `p=0` → still 3 distinct partitions (orphans remain by design).

Comments in the test spell out the #10282 / #12240 product rule for future readers.

## What the impact is

- **Correctness:** Guards the lake-replace path users actually need (`OVERWRITE`).
- **Compatibility:** No runtime change; OR_IGNORE semantics untouched.
- **DX:** Issue reporters can be pointed at `OVERWRITE` + this test as the supported rebuild recipe.
- **Residual risk:** Remote object stores may still need separate “purge prefix” work — out of scope; local/POSIX behavior is what this locks.

## Testing plan

```bash
GEN=ninja make reldebug
build/reldebug/test/run test/sql/copy/partitioned/hive_partitioning_overwrite.test
```

Manual sketch:

```sql
SET threads=4;
COPY (SELECT range%2 AS p, range AS v FROM range(100000))
TO 'mt' (FORMAT PARQUET, PARTITION_BY (p));

SET threads=1;
COPY (SELECT range%2 AS p, range AS v FROM range(100))
TO 'mt' (FORMAT PARQUET, PARTITION_BY (p), OVERWRITE);

SELECT count(*) FROM glob('mt/**/*.parquet');  -- 2
SELECT p, count(*) FROM read_parquet('mt/**/*.parquet', hive_partitioning=true)
GROUP BY p ORDER BY p;  -- 0|50, 1|50
```

## Everything else

- Related: #12240 (mode split), #12225 (append / ignore behavior).
- Out of scope: making OR_IGNORE recursive-delete; S3/GCS full prefix clear; changing default COPY conflict mode.
- Recommended issue comment: “For idempotent partition rebuilds use `OVERWRITE`. `OVERWRITE_OR_IGNORE` only clobbers colliding filenames and will leave stale `data_N` files when thread counts shrink.”

Suggested reviewers: @Mytherin @pdet

### DIFF
--- a/test/sql/copy/partitioned/hive_partitioning_overwrite.test
+++ b/test/sql/copy/partitioned/hive_partitioning_overwrite.test
@@ -45,3 +45,131 @@ statement error
 COPY (SELECT 84 AS part_col) TO '{TEST_DIR}/overwrite_test' (FORMAT PARQUET, PARTITION_BY (part_col), OVERWRITE 1, OVERWRITE_OR_IGNORE 1);
 ----
 OVERWRITE
+
+# === Issue #10282: multi-thread write then fewer-thread rewrite ===
+# OVERWRITE must recursively clear so orphan data_N.parquet files / partitions do not remain.
+# OVERWRITE_OR_IGNORE intentionally does NOT clear; it only clobbers same filenames.
+
+# Multi-threaded write creates multiple data_*.parquet files per partition
+statement ok
+SET threads = 4;
+
+statement ok
+COPY (
+	SELECT range % 2 AS p, range AS v
+	FROM range(100_000)
+) TO '{TEST_DIR}/mt_overwrite' (FORMAT PARQUET, PARTITION_BY (p));
+
+# Several files from parallel writers
+query I
+SELECT count(*) >= 2 FROM glob('{TEST_DIR}/mt_overwrite/**/*.parquet');
+----
+true
+
+query II
+SELECT p, count(*)
+FROM read_parquet('{TEST_DIR}/mt_overwrite/**/*.parquet', hive_partitioning = true)
+GROUP BY p
+ORDER BY p;
+----
+0	50000
+1	50000
+
+# Single-threaded rewrite of a smaller dataset with OVERWRITE: orphans must be gone
+statement ok
+SET threads = 1;
+
+statement ok
+COPY (
+	SELECT range % 2 AS p, range AS v
+	FROM range(100)
+) TO '{TEST_DIR}/mt_overwrite' (FORMAT PARQUET, PARTITION_BY (p), OVERWRITE);
+
+# One file per partition (single thread)
+query I
+SELECT count(*) FROM glob('{TEST_DIR}/mt_overwrite/**/*.parquet');
+----
+2
+
+query II
+SELECT p, count(*)
+FROM read_parquet('{TEST_DIR}/mt_overwrite/**/*.parquet', hive_partitioning = true)
+GROUP BY p
+ORDER BY p;
+----
+0	50
+1	50
+
+# OVERWRITE also drops partition directories that no longer appear in the data
+statement ok
+SET threads = 4;
+
+statement ok
+COPY (
+	SELECT range % 3 AS p, range AS v
+	FROM range(30_000)
+) TO '{TEST_DIR}/mt_overwrite_parts' (FORMAT PARQUET, PARTITION_BY (p), OVERWRITE);
+
+query I
+SELECT count(DISTINCT p)
+FROM read_parquet('{TEST_DIR}/mt_overwrite_parts/**/*.parquet', hive_partitioning = true);
+----
+3
+
+statement ok
+SET threads = 1;
+
+statement ok
+COPY (
+	SELECT 0 AS p, range AS v
+	FROM range(10)
+) TO '{TEST_DIR}/mt_overwrite_parts' (FORMAT PARQUET, PARTITION_BY (p), OVERWRITE);
+
+query I
+SELECT count(*) FROM glob('{TEST_DIR}/mt_overwrite_parts/**/*.parquet');
+----
+1
+
+query II
+SELECT p, count(*)
+FROM read_parquet('{TEST_DIR}/mt_overwrite_parts/**/*.parquet', hive_partitioning = true)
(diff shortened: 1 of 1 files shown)


Mytherin:
This PR is only adding tests which is not useful. Also please stop `@`-ing maintainers directly in the PRs you open.

PR #24167, opened 2026-07-26
Do not push TRY_CAST into CSV reader column types

### LINKED ISSUE #23818: TRY_CAST throws exception on CSV read
### What happens?

TRY_CAST should return NULL if casting fails, instead of throwing exceptions.

### To Reproduce

```sql
memory D CREATE OR REPLACE TEMP TABLE bad_csv AS
         SELECT *
         FROM (VALUES
             ('1'),
             ('nope'),
             ('2')
         ) v(content);
memory D COPY bad_csv TO '/tmp/bad.csv' (FORMAT CSV, HEADER false);
memory D SELECT TRY_CAST(x AS INTEGER)
         FROM read_csv('/tmp/bad.csv', columns={'x':'VARCHAR'}, header=false);
Conversion Error:
CSV Error on Line: 2
Original Line: nope
Error when converting column "x". Could not convert string "nope" to 'INTEGER'
```

### OS:

macos

### DuckDB Version:

ccf5b60ba140051e57bc0fb334cc9b6a154c4ef8

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Vamsi-klu:
## Problem

`TRY_CAST(col AS T)` over `read_csv` can still throw a conversion error on dirty cells. Type pushdown treats `TRY_CAST` like a hard cast, retypes the CSV column, and the scanner fails the cell instead of returning NULL.

## Solution

Refuse projection pushdown for `TRY_CAST` in the CSV reader so the column stays VARCHAR and scalar `TRY_CAST` produces NULL. Plain `CAST` pushdown is unchanged.

Fixes #23818

### Test

```bash
build/reldebug/test/unittest test/sql/copy/csv/test_try_cast_csv_pushdown.test
build/reldebug/test/unittest test/optimizer/pushdown/csv_type_pushdown.test
```

### DIFF
--- a/src/function/table/read_csv.cpp
+++ b/src/function/table/read_csv.cpp
@@ -147,6 +147,10 @@ static bool PushdownProjectionExpression(ClientContext &context, const TableFunc
 		return false;
 	}
 	const auto &cast = input.expr.Cast<BoundFunctionExpression>();
+	// CSV scanner only hard-casts; leave TRY_CAST in the projection (issue #23818)
+	if (BoundCastExpression::IsTryCast(cast)) {
+		return false;
+	}
 	const auto &target_type = cast.GetReturnType();
 	auto &bind_data = input.get.bind_data->Cast<MultiFileBindData>();
 	const idx_t idx = input.get.GetColumnIds()[input.column_index].GetPrimaryIndex();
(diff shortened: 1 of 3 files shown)


Vamsi-klu:
@codex review

Please run a full code review on this PR.

src/function/table/read_csv.cpp
```diff
@@ -147,6 +147,10 @@ static bool PushdownProjectionExpression(ClientContext &context, const TableFunc
 		return false;
 	}
 	const auto &cast = input.expr.Cast<BoundFunctionExpression>();
+	// CSV scanner only hard-casts; leave TRY_CAST in the projection (issue #23818)
```
dentiny, line 150: Why do we remove the optimization?

PR #24168, opened 2026-07-26
Throw on HUGEINT SUM/AVG overflow in aggregate Combine

### LINKED ISSUE #24081: SUM/AVG over HUGEINT silently overflows instead of erroring (on Combine step)
### What happens?

SUM(v) and AVG(v) over a HUGEINT column can silently overflow and return a wrong result instead of raising an Out of Range error. The problem occurs in the Combine step.

Patch (by Claude Code):
[overflow.patch](https://github.com/user-attachments/files/30300286/overflow.patch)

### To Reproduce

```
SELECT sum(v) FROM (
  SELECT CAST('100000000000000000000000000000000000000' AS HUGEINT) AS v  -- 1e38
  UNION ALL
  SELECT CAST('100000000000000000000000000000000000000' AS HUGEINT)) t;
-- Result: -140282366920938463463374607431768211456 (= 2e38 − 2^128, a two's-complement wrap)
-- Expected: an Out of Range overflow error (true sum 2e38 > int128 max 1.7014e38)
```

### OS:

Ubuntu 26.04 LTS

### DuckDB Version:

34f2a32f08e9c454205f1ce45a4e9fa38adf4f53

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Viktor Leis

### Affiliation:

TUM

- [x] Yes, I have

- [x] Yes, I have

Yes

Vamsi-klu:
## Problem

`SUM` / `AVG` over `HUGEINT` check overflow on per-row updates, but the multi-partial **Combine** path used unchecked `+=`, so large values can silently wrap to a wrong negative (e.g. under `UNION ALL` or multiple threads).

## Solution

Combine hugeint partials with `Hugeint::Add` (same overflow behavior as the row path) in SUM and AVG.

Fixes #24081

### Test

```bash
build/reldebug/test/unittest test/sql/types/hugeint/hugeint_sum_overflow.test
```


Mytherin:
Thanks!

PR #24171, opened 2026-07-26
Auto-detect ISO-8601 timestamps with offsets as TIMESTAMPTZ in JSON

### LINKED ISSUE #14919: JSON/CSV auto detection of ISO 8601 Timestamp strings with time zone offset
## Overview

Hi duckDB team,

In the documentation, it appears that the `read_json_auto` function should handle properly parsing `ISO 8601` formatted timestamp strings: https://duckdb.org/docs/data/json/loading_json#json-read-functions

> dateformat	Specifies the date format to use when parsing dates. See [Date Format]	VARCHAR	'iso'

Unfortunately, it doesn't look like this parsing works when the timestamps have a timezone offset attached (see to reproduce below).

I would've expected the `json_read_auto` to correctly infer this column's type as `TIMESTAMPTZ`. 

Normally, it wouldn't be a big deal to cast the column, but we use DuckDB to load arbitrary JSON data from our customers, so the type inference is especially helpful to make sure we are able to process our customer's data as intended.

Thanks for the help 🙏 

## To Reproduce

JSON file `test.json`:
```json
{
  "timestamp": "2024-11-12T20:47:01.917818Z"
}
```

duckdb SQL:
```sql
CREATE
OR REPLACE TABLE test_data AS
SELECT *
FROM read_json_auto('test.json');

DESCRIBE test_data;
```

output:

| column_name | column_type | null | key | default | extra |
| ----- | ----- | ---- | ---- | ---- | ---- |
| timestamp | VARCHAR | YES | | |

## Some investigation

Digging into the [code](https://github.com/duckdb/duckdb/blob/c29c67bb971362cd1e9143305acffebb1bc9bd63/extension/json/json_scan.cpp#L101-L101) a bit, it looks like the JSON scanner doesn't have the ability to auto-infer `TIMESTAMPTZ` column types. Additionally, it looks like the scanner infers timestamp strings with a `Z` offset as plain `TIMESTAMP` types without a time zone using the format `"%Y-%m-%dT%H:%M:%SZ"`. I think this could be dangerous as other values in the data may have offsets in different timezones, so I would expect this to be treated as a `TIMESTAMPTZ` type value. 

Another strange finding, it looks like the [CSV scanner](https://github.com/duckdb/duckdb/blob/c29c67bb971362cd1e9143305acffebb1bc9bd63/src/include/duckdb/execution/operator/csv_scanner/sniffer/csv_sniffer.hpp#L204-L204) also does not support inferring `TIMESTAMPTZ` type columns. Strangely, the formats accepted for the CSV scanner differ from the formats for the JSON scanner. Namely, the format mentioned above, `"%Y-%m-%dT%H:%M:%SZ"`, is not present in the list of supported formats for CSV. 

## Related work

If found this [Issue](https://github.com/duckdb/duckdb/issues/1627) and [MR](https://github.com/duckdb/duckdb/pull/1632) from some time ago that seemed to add offset parsing support to the `TIMEZONE` data type, but it looks like it has been removed.

There is an [issue](https://github.com/duckdb/duckdb/issues/12469) currently open related to multiple timestamp string formats in a column, which seems related to the auto detected format strings.

### OS:

macOS 13.6.9 - x86-64 arch

### DuckDB Version:

v1.1.1

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Ian McGraw

### Affiliation:

Arthur AI

Yes

- [X] Yes, I have

- [X] Yes, I have

Vamsi-klu:
## Problem

JSON type detection never considered `TIMESTAMP WITH TIME ZONE`, so ISO-8601 strings with `Z` or numeric offsets became plain `TIMESTAMP` and dropped offset information.

## Solution

Add `TIMESTAMP_TZ` to the JSON auto-detect candidates, accepted only when every non-null sampled value has an offset or `Z`. Plain timestamps without offset stay `TIMESTAMP`.

### Design decisions

1. **Mixed offset + plain (same field) → `VARCHAR`**
   Neither `TIMESTAMPTZ` (would invent a session TZ for naive values) nor `TIMESTAMP` (would drop offsets) is safe. Documented in code comments and pinned by tests (same-vector mixed and across a vector boundary).

2. **Explicit `user_specified_timestamp_format` flag**
   Set only when the user passed `timestamp_format=` / `timestampformat=`. Threaded into `InitializeCandidateTypes` (omit `TIMESTAMP_TZ` from candidates) and `EliminateCandidateTypes`. Do not infer user intent from the date-format map (auto-detect can also converge on one format).

3. **Monotonic / sticky elimination across vectors**
   Offset vs plain observation is sticky on `JSONStructureDescription` across refinement vectors and parallel merge. Once both have been seen, candidates are cleared (`VARCHAR`). Null-only vectors do not disqualify a prior `TIMESTAMPTZ` candidate.

### Edge cases covered

| Case | Result |
|------|--------|
| All values have offset/`Z` | `TIMESTAMPTZ` |
| All values plain (no offset) | `TIMESTAMP` |
| Mixed offset + plain (+ null) | `VARCHAR` |
| All-NULL column | SQLNULL (`"NULL"`), not `TIMESTAMPTZ` |
| Offset spellings `Z`, `+00:00`, `+05:30`, `+0530`, `-08:00` | `TIMESTAMPTZ` |
| Lowercase `z` | `VARCHAR` (same as `CAST` / `TryConvertTimestampTZ`) |
| User `timestamp_format=` | no TZ auto-detect; non-matching stay `VARCHAR` |
| Vector boundary (2048+ rows) offset then plain | `VARCHAR` |
| Vector boundary offset then null | stays `TIMESTAMPTZ` |

Fixes #14919

### Test justifications (`read_json_dates.test`)

- Existing ISO-8601 cases updated from `TIMESTAMP` → `TIMESTAMPTZ` where offsets/`Z` are present
- `issue_22103`: separate columns — offset cols → TZ, plain col → TIMESTAMP
- New mixed same-field (offset, null, plain) → VARCHAR
- New all-null → `"NULL"` type
- New offset spelling battery (`+05:30`, `+0530`, `-08:00`)
- New plain-only → TIMESTAMP
- New lowercase `z` → VARCHAR (parser limitation)
- New vector-boundary mixed / all-TZ / TZ-then-null
- Explicit `timestamp_format=` still skips TZ auto-detect

### Test

```bash
build/release/test/unittest test/sql/json/table/read_json_dates.test
```


test/sql/json/table/read_json_dates.test
```diff
@@ -172,17 +172,18 @@ copy (select * from (values ('{"ts":"2023-02-07T19:12:28.789-07:00"}'))) to '{TE
 query II
 select typeof(ts), ts from read_json_auto('{TEMP_DIR}/iso8601_neg_offset.json')
 ----
-TIMESTAMP	2023-02-08 02:12:28.789
+TIMESTAMP WITH TIME ZONE	2023-02-08 02:12:28.789+00
 
 # Multiple timestamp formats in single JSON (issue 22103)
+# ts1-ts3 have offsets -> TIMESTAMPTZ; ts4 has no offset -> TIMESTAMP
 statement ok
 copy (select * from (values ('{"ts1":"2026-04-01T06:01:00.010000+00:00", "ts2":"2026-04-02T06:03:00+00:01", "ts3":"2026-04-03T06:03:00Z", "ts4":"2026-04-04T06:04:00"}'))) to '{TEMP_DIR}/issue_22103.json' (format csv, quote '', header 0)
```
lnkuiper, line 215: Could we add a test where the same ts field contains an offset timestamp, a null, and a plain timestamp across separate rows? The new invariant requires every non-null sampled value to carry an offset, but the current test puts offset and plain timestamps in separate columns, so it does not exercise candidate elimination across values. This would also document whether the intended mixed-value fallback is VARCHAR, which is the current behavior.

extension/json/json_functions/json_structure.cpp
```diff
@@ -269,7 +296,24 @@ void JSONStructureNode::EliminateCandidateTypes(const idx_t vec_count, Vector &s
 		}
 		const auto type = candidate_types.back();
 		Vector result_vector(type, vec_count);
-		if (date_format_map.HasFormats(type)) {
+		if (type == LogicalTypeId::TIMESTAMP_TZ) {
+			// If the user supplied a single timestamp_format, do not override it with unrestricted
+			// TIMESTAMPTZ detection (non-matching TZ strings should stay VARCHAR).
+			if (date_format_map.HasFormats(LogicalTypeId::TIMESTAMP) &&
+			    date_format_map.NumberOfFormats(LogicalTypeId::TIMESTAMP) == 1) {
```
lnkuiper: Inferring this from the date format map seems hacky: can we make this explicit by passing the options or some other parameter?

lnkuiper (changes requested):
Thanks for the PR! Some comments:

test/sql/json/table/read_json_dates.test
Vamsi-klu: Good call — added that case (offset, null, plain in the same `ts` field).

**Mixed-value decision: VARCHAR.** If a field has both offset and plain timestamps we refuse to guess: TIMESTAMPTZ would invent a session TZ for the naive values, and TIMESTAMP would drop offsets. Documented in the code and pinned with tests (same-vector mixed, and across a 2048-row vector boundary).

Also fixed sticky elimination across vectors: once any non-null value lacks an offset, TIMESTAMPTZ is out for good; null-only vectors no longer wipe a prior TIMESTAMPTZ candidate. Offset-then-null across the boundary stays TIMESTAMPTZ.

extension/json/json_functions/json_structure.cpp
Vamsi-klu: Agreed — that was wrong when auto-detect converges on a single format.

Now `JSONScanData::user_specified_timestamp_format` is set only when the user actually passed `timestamp_format=` / `timestampformat=`. It is threaded into `InitializeCandidateTypes` (TIMESTAMP_TZ is omitted from the candidate list) and `EliminateCandidateTypes` (defense in depth). No more inference from the date-format map.

extension/json/include/json_structure.hpp
```diff
@@ -83,6 +87,11 @@ struct JSONStructureDescription {
 
 	//! Whether any UBIGINT value exceeds the BIGINT range (i.e., >= 2^63)
 	bool has_large_ubigint = false;
+
+	//! Sticky ISO-8601 offset observation across refinement vectors (issue #14919).
+	//! If both are true the field is mixed → candidates cleared (VARCHAR).
+	bool has_timestamp_with_offset = false;
+	bool has_timestamp_without_offset = false;
```
lnkuiper: Can we use an `enum` here, e.g.:
```C++
enum class TimestampOffsetState : uint8_t {
	UNKNOWN,
	WITH_OFFSET,
	WITHOUT_OFFSET,
	MIXED
};
```

extension/json/include/json_scan.hpp
```diff
@@ -37,6 +37,8 @@ struct JSONScanData : public TableFunctionData {
 
 	//! The date format map
 	unique_ptr<DateFormatMap> date_format_map;
+	//! True only when the user passed timestampformat=/timestamp_format= (not auto-detect templates)
+	bool user_specified_timestamp_format = false;
```
lnkuiper: This should live in `JSONReaderOptions` instead, or maybe we should use `timestamp_format` and check for emptiness, or make it an `optional<string>`

lnkuiper (commented):
Thanks! There's a formatting failure, and I have two more comments:

extension/json/include/json_structure.hpp
Vamsi-klu: Done in `fca12a89`, using the enum exactly as you wrote it. There is one transition helper shared by the per-vector refinement and `MergeNodes`, so the parallel merge stays order independent: `WITH_OFFSET` merged with `WITHOUT_OFFSET` gives `MIXED` either way round, `UNKNOWN` is the identity, and `MIXED` absorbs everything. `MIXED` is sticky and clears the timestamp candidates, so the field falls back to `VARCHAR`.

extension/json/include/json_scan.hpp
Vamsi-klu: Went with your middle option in `fca12a89` and dropped the extra flag entirely, checking `options.timestamp_format` for emptiness instead. It only ever gets set from the user-supplied `timestamp_format` or `timestampformat` parameter in `json_multi_file_info.cpp`, and auto-detect never writes the converged format back into options, so emptiness is exactly the signal we need without carrying new state. The formatting failure is fixed in the same commit.

lnkuiper:
Thanks for the changes! We're almost there.

`infinity` is treated as offset evidence here. `TryConvertTimestampTZ` passes `has_offset` to `Date::TryConvertDate`, where that output means “special value”, so ±infinity sets it to true. Combining an infinite value with a naive timestamp then produces MIXED and incorrectly falls back to `VARCHAR`.

Repro:
```sql
COPY (
    SELECT json_object('ts', ts)
    FROM (VALUES
        ('2024-01-01T00:00:00'),
        ('infinity'),
        ('-infinity')
    ) t(ts)
) TO '/tmp/timestamp_specials.json'
  (FORMAT CSV, QUOTE '', HEADER FALSE);
SELECT typeof(ts)
FROM read_json_auto('/tmp/timestamp_specials.json')
LIMIT 1;
```

Vamsi-klu:
Thanks, that is a real bug. `TryConvertTimestampTZ` reuses `has_offset` as the `special` out-param of `Date::TryConvertDate`, so `infinity` and `-infinity` look like offset evidence. Mixed with a naive timestamp that becomes MIXED and falls back to VARCHAR.
I will treat specials as non-evidence in `UpdateTimestampOffsetState` (same as nulls). They are valid for both TIMESTAMP and TIMESTAMPTZ, so they should not vote WITH_OFFSET or WITHOUT_OFFSET. The repro (plain + plus/minus infinity) should stay TIMESTAMP. Offset/Z plus infinity should stay TIMESTAMPTZ. I will also cover `epoch`, which hits the same `special` flag.
Adding tests for those cases in `read_json_dates.test`.

lnkuiper (approved):
Thanks for the changes!

PR #24172, opened 2026-07-26
Expand multi-file UNION columns for Parquet nested remap

### LINKED ISSUE #20037: Unable to select union types containing list types via read_parquet
### What happens?

When attempting to read parquet data via `read_parquet`, if a column is a `UNION` type containing a `List` subtype, the following error is raised

> Struct remap can only remap named structs

This seems to be new behavior as of version `1.3.0`.

I am not sure if there are other failure cases beyond `List` subtypes, that is the only one I've ran into at this time.

### To Reproduce

```sql
CREATE OR REPLACE TABLE origin (
  ufield UNION(val VARCHAR[], omitted BOOLEAN)
);

COPY (SELECT * FROM origin)
TO 'out.parquet'
WITH (FORMAT 'parquet', COMPRESSION 'zstd');

-- Works in <= 1.2.2, fails in >= 1.3.0
-- Will also work in all versions if `val` is a non-list type
SELECT
CAST(ufield AS UNION(val VARCHAR[], omitted BOOLEAN)) AS ufield
FROM read_parquet('out.parquet');

```

### OS:

MacOS

### DuckDB Version:

1.3.0

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Stephen Williams

### Affiliation:

AppOmni

- [x] Yes, I have

- [x] Yes, I have

Yes

Vamsi-klu:
## Problem

Reading Parquet (and other multi-file sources) fails to remap nested `UNION` columns that contain `LIST` (and similar nesting). See #20037.

## Solution

Narrowed after review:

- `CreateFromNameAndType` expands **UNION** children only (named-struct layout with tag via `NamedChildren`). **LIST/MAP are not expanded** here — their child names differ by reader (JSON vs Parquet), and synthesizing them in this shared helper caused regressions.
- `MapColumn` routes `UNION` through `MapColumnStruct`.
- `ConstructMapExpression` only uses `remap_struct` when global and local nested **type kinds match**, so Parquet `STRUCT`→`UNION` falls back to cast (needed for `COPY` into typed UNION tables).

Fixes #20037

### Test

```bash
build/release/test/unittest test/sql/copy/parquet/writer/parquet_union_list.test
```


lnkuiper:
Thanks for the PR! This introduces a few regressions:
```sql
# name: test/temp.test
# description: Regressions from expanding generic multi-file nested columns
# group: [temp]

require json

require parquet

statement ok
COPY (SELECT map(['a'], [1]) AS m) TO '{TEMP_DIR}/map.json' (FORMAT JSON);

query I
SELECT m['a'] FROM read_json('{TEMP_DIR}/map.json', columns = {m: 'MAP(VARCHAR, INTEGER)'});
----
1

statement ok
CREATE TABLE source(u UNION(val INTEGER[], omitted BOOLEAN));

statement ok
INSERT INTO source VALUES (union_value(val := [1, 2])), (union_value(omitted := true));

statement ok
COPY source TO '{TEMP_DIR}/union.parquet' (FORMAT PARQUET);

statement ok
CREATE TABLE target(u UNION(val INTEGER[], omitted BOOLEAN));

statement ok
COPY target FROM '{TEMP_DIR}/union.parquet' (FORMAT PARQUET);

query I
SELECT u FROM target;
----
[1, 2]
true
```

Vamsi-klu:
Thanks for catching those , both were real.

Root cause was expanding LIST/MAP children inside `CreateFromNameAndType`, which is shared across multi-file readers. JSON MAP local defs don’t match the synthetic `key`/`value` children, and the UNION `COPY` path was going through `remap_struct` for STRUCT→UNION (not allowed).

Changes:
- Dropped LIST/MAP expansion from `CreateFromNameAndType`; kept UNION only (same named-struct layout as STRUCT).
- `ConstructMapExpression` now requires matching nested type kinds before using `remap_struct`, so Parquet STRUCT→UNION uses cast instead.
- Pinned both of your reproductions (JSON MAP + COPY UNION) in `parquet_union_list.test`.

lnkuiper:
There's still an issue:

```sql
require json

statement ok
COPY (SELECT union_value(element1 := 42)::UNION(element1 INTEGER, other BOOLEAN) AS data)
TO '{TEMP_DIR}/union_element1.json' (FORMAT JSON);

query I
SELECT data FROM read_json('{TEMP_DIR}/union_element1.json',
	columns = {data: 'UNION(element1 INTEGER, other BOOLEAN)'});
----
42
```

Gives:

```text
Binder Error: Can't remap type UNION(element1 INTEGER, other BOOLEAN)
```

I also could not reproduce the reported #20037 failure on the PR's exact base. Maybe this got fixed in the meantime already? Or is the test not formulated properly? Do we still need this PR?

There's also a formatting issue in CI

Vamsi-klu:
You were right to ask whether this is still needed. I checked #20037 against this PR's own unpatched base (`c47eae1c`) and the original repro passes there, as does this PR's whole test file, so the bug was fixed upstream at some point along the way.

So I have dropped all of the `src/` changes. That also removes the `read_json` UNION failure you hit, since `ConstructMapExpression` no longer routes UNION into `remap_struct` at all. What is left in `e252850` is the test file as regression coverage, with all three of your repros pinned: the JSON MAP case with `columns=`, the `COPY FROM` parquet into a typed UNION table case, and the `read_json` UNION case. Formatting is fixed too.

Happy to keep this as a tests-only PR, or to close it if you would rather not carry the extra coverage. Your call.

lnkuiper:
Thanks! I'm OK with test-only :)

PR #24173, opened 2026-07-26
Add opt-in HUGEINT_ENCODING for Parquet COPY (double/decimal/varchar)

### LINKED ISSUE #21180: Loss of Precision when Writing HUGEINT Columns to Parquet
### What happens?

When a table containing a HUGEINT column is written to a Parquet file, the HUGEINT value is incorrectly converted to a floating-point DOUBLE, which results in a loss of precision. For example, inserting the large value 999999999999999999999999999999 and then reading it back yields 1e+30.

I believe the issue stems from the following:

The HUGEINT logical type is [mapped to the INT128 physical type](https://github.com/duckdb/duckdb/blob/main/src/common/types.cpp#L95), similar to DECIMAL. However, unlike DECIMAL, when writing INT128 for a HUGEINT to Parquet, it appears to be [converted into a floating-point DOUBLE](https://github.com/duckdb/duckdb/blob/main/extension/parquet/parquet_writer.cpp#L90) instead of being stored as a fixed-length byte array.

### To Reproduce

```
duckdb -c "
CREATE TABLE test (huge_integer HUGEINT);
INSERT INTO test VALUES ('999999999999999999999999999999'::HUGEINT);
COPY test TO 'hugeint.parquet' (FORMAT 'parquet');
SELECT * FROM 'hugeint.parquet';
"
```

Actual Result
```
┌──────────────┐
│ huge_integer │
│    double    │
├──────────────┤
│        1e+30 │
└──────────────┘
```

Inspection of the Parquet file's schema using parquet-tools confirms that the underlying physical type for the HUGEINT column is DOUBLE:
```
$ parquet-tools inspect hugeint.parquet
########## file meta data ##########
created_by: DuckDB version v1.1.3 (build 19864453f7)
num_columns: 1
num_rows: 1
num_row_groups: 1
format_version: 1.0
serialized_size: 140

########## Columns ##########
huge_integer

########## Column(huge_integer) ##########
name: huge_integer
path: huge_integer
max_definition_level: 1
max_repetition_level: 0
physical_type: DOUBLE
logical_type: None
converted_type (legacy): NONE
compression: SNAPPY (space_saved: -6%)
```

### OS:

Linux 5.15.0-171-generic #181-Ubuntu SMP Fri Feb 6 22:44:50 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

### DuckDB Version:

latest (1.4.4)

### DuckDB Client:

NA - can be reproduced with duckdb CLI tool

### Hardware:

_No response_

### Full Name:

Zdenko Pulitika

### Affiliation:

Thought Machine

- [x] Yes, I have

- [x] Yes, I have

Vamsi-klu:
## Problem

Writing `HUGEINT` / `UHUGEINT` to Parquet as `DOUBLE` silently loses integer precision for large values (#21180).

An earlier version of this PR changed the **default** to `DECIMAL(38,0)`. That premise was wrong: the conversion is **not** lossless.

| Type | Max digits |
|------|------------|
| HUGEINT | 39 |
| UHUGEINT | 39 |
| DECIMAL(38,0) / Parquet FLBA(16) | 38 |

So `DECIMAL(38,0)` cannot represent the full hugeint range. Silently switching the default would also break partitioned writes for out-of-range partition keys and would change `EXPORT DATABASE` fidelity if the VARCHAR special-case were removed.

## Solution (redesign)

**Do not change the default.** Add an explicit Parquet COPY option:

```sql
COPY t TO 'f.parquet' (FORMAT parquet, HUGEINT_ENCODING 'decimal');
-- 'double'  : current default, lossy DOUBLE, unchanged
-- 'decimal' : DECIMAL(38,0) FLBA(16); values outside range raise Conversion Error
-- 'varchar' : exact string form; always safe, less numeric interop
```

### Behavior

| Path | Encoding |
|------|----------|
| `COPY ...` default / `HUGEINT_ENCODING 'double'` | DOUBLE (same as main) |
| `COPY ... HUGEINT_ENCODING 'decimal'` | cast to `DECIMAL(38,0)` before write; out-of-range → error |
| `COPY ... HUGEINT_ENCODING 'varchar'` | cast to `VARCHAR` before write |
| `EXPORT DATABASE` (FORMAT PARQUET) | always VARCHAR for hugeint (unchanged) so IMPORT restores `HUGEINT` exactly |

Implementation is only in `ParquetWriteSelect` (+ option bind/list/serialize). The writer type map is untouched for the default path, so stats, dictionaries, and partitioned writes behave as before unless the user opts in.

### Fixes called out in review

1. **Lossless premise** — dropped. Title/body no longer claim lossless default conversion.
2. **EXPORT DATABASE** — VARCHAR path kept exactly as on main (not replaced by DECIMAL).
3. **Stats unifier / UHUGEINT ≥ 2^127** — not relevant on the default path (still DOUBLE). Under `'decimal'`, values ≥ 10^38 fail the cast before write (no signed FLBA reinterpretation).
4. **Partitioned writes** — default path works with full hugeint partition keys again. Decimal mode errors on out-of-range partition keys instead of corrupting.

### Read-back type (E9)

| Encoding | `typeof` after `read_parquet` |
|----------|-------------------------------|
| double | `DOUBLE` |
| decimal | `DECIMAL(38,0)` |
| varchar | `VARCHAR` |

DuckDB type is not preserved in the Parquet file under any of these encodings; `EXPORT DATABASE` recovers `HUGEINT` via `schema.sql` + VARCHAR data.

Fixes #21180

### Test plan

```bash
build/release/test/unittest test/sql/copy/parquet/writer/parquet_write_hugeint.test
build/release/test/unittest test/sql/copy/parquet/writer/parquet_write_uhugeint.test
build/release/test/unittest test/sql/export/parquet/export_parquet_hugeint.test
# also green (unchanged expectations):
build/release/test/unittest test/sql/copy/parquet/writer/parquet_test_all_types.test
build/release/test/unittest test/sql/copy/parquet/alltypes-dictionaries.test
build/release/test/unittest test/optimizer/joins/parquet_minmax_cardinality_estimation.test
```

### Test expectation changes vs main

Only the three tests above were extended. No unjustified expectation churn on unrelated tests.

| File | Change | Why |
|------|--------|-----|
| `parquet_write_hugeint.test` | Keep default DOUBLE asserts; add decimal/varchar/partition/range-error cases | Cover opt-in modes + reviewer's PARTITION_BY repro |
| `parquet_write_uhugeint.test` | Same pattern for UHUGEINT | Cover E3 (UHUGEINT ≥ 2^127 errors under decimal) |
| `export_parquet_hugeint.test` | Keep full `test_all_types` hugeint extremes; assert typeof is HUGEINT after IMPORT | Prove EXPORT VARCHAR path still exact (no clamping) |

### Remaining / follow-ups

- Docs page for `HUGEINT_ENCODING` (copy Parquet options reference) not updated in this PR.
- Column name is not always present in the cast error text (`Could not cast value … to DECIMAL(38,0)`); improving that would be a separate cast-diagnostics change.


test/sql/export/parquet/export_parquet_hugeint.test
```diff
@@ -1,20 +1,26 @@
...
 
+# Use in-range values only — full hugeint extremes exceed DECIMAL(38,0)
 statement ok
-create table tbl as select "hugeint" from test_all_types(), range(3);
+create table tbl as
+select * from (values
+((-1180591620717411303424)::HUGEINT),
+((1180591620717411303424)::HUGEINT),
+(NULL::HUGEINT)
+) t(hugeint);
```
lnkuiper: This clearly introduces a regression which is now hidden because the test has changed. Clearly the conversion is not lossless. What's the point of not using DOUBLE then? Neither are lossless. There's also an issue with partitioned writes:
```sql
# name: test/temp.test
# description: PR #24173 partition-key reproducer
# group: [parquet]

require parquet

statement ok
COPY (
	SELECT 170141183460469231731687303715884105727::HUGEINT AS p, 42 AS x
) TO '{TEST_DIR}/partitioned' (
	FORMAT PARQUET,
	PARTITION_BY ('p')
);

query I
SELECT x FROM read_parquet('{TEST_DIR}/partitioned/*/*.parquet')
----
42
```

lnkuiper (changes requested):
Thanks for the PR! I don't think we want to move in this direction. Comment below:

test/sql/export/parquet/export_parquet_hugeint.test
Vamsi-klu: You're right — thanks for catching this.

DECIMAL(38,0) is not lossless for HUGEINT/UHUGEINT (39 digits vs 38), so changing the default was the wrong direction. The test edits that clamped values were also hiding a real regression, including on EXPORT DATABASE.

I've redesigned this as an **opt-in** only:

```sql
COPY t TO 'f.parquet' (FORMAT parquet, HUGEINT_ENCODING 'decimal');
-- 'double'  (default): unchanged lossy DOUBLE path
-- 'decimal': DECIMAL(38,0) FLBA; out-of-range → Conversion Error
-- 'varchar': exact string form
```

Specifically:

- Default behavior matches main again (DOUBLE).
- EXPORT DATABASE still uses the VARCHAR path so IMPORT restores HUGEINT exactly, including full extremes from `test_all_types`.
- Your PARTITION_BY repro works again on the default path; under `'decimal'` an out-of-range partition key errors instead of writing bad data.
- No more signed DecimalStatsUnifier on raw UHUGEINT — decimal mode casts first, so values ≥ 10^38 never reach the writer.

PR title/body updated accordingly. Leaving as draft until you're happy with the direction.

Vamsi-klu:
Redesigned after review — default is no longer changed.

**You're right that DECIMAL(38,0) is not lossless** (HUGEINT/UHUGEINT need 39 digits; Parquet decimal FLBA max precision is 38). The previous approach was wrong on that premise, and the test clamps hid the EXPORT / range regression.

New shape: opt-in `HUGEINT_ENCODING` (`double` | `decimal` | `varchar`), default `double` unchanged. EXPORT DATABASE keeps the VARCHAR path. Partitioned writes work again on the default path.

Details in the updated PR body. Still draft for your take on the direction.

test/sql/export/parquet/export_parquet_hugeint.test
Vamsi-klu: You were right and I was wrong about lossless: `HUGEINT` needs 39 digits and `DECIMAL(38,0)` holds 38, so the premise the first version rested on does not hold. The default `DOUBLE` path is untouched now and this test is back to asserting the original behavior. The `DECIMAL` path is opt-in only, and out-of-range values raise a Conversion Error rather than silently wrapping.

Vamsi-klu:
Gentle ping on the redesigned version, since the changes-requested review predates it by a few days.

The default `DOUBLE` path is completely unchanged now, so your `PARTITION_BY` repro passes again, `EXPORT DATABASE` keeps the same `VARCHAR` path, and `DECIMAL(38,0)` and `VARCHAR` are strictly opt-in through `HUGEINT_ENCODING`. Out-of-range values raise a Conversion Error instead of writing bad data. The latest push (`fbd2078`) adds option-validation tests for case insensitivity and wrong arity.

If you think no option is worth having here I am fine closing it. Otherwise I will mark it ready so the full matrix runs.

lnkuiper:
Hi, thanks for the ping. I think no option is worth having here until we have something that properly losslessly converts. Closing this now, in hope of a better option in the future.

Issue #17114, opened 2025-04-14
Support `BYTE_STREAM_SPLIT` on non-float columns


markasoftware-aqtc:
### What happens?

As of about a year ago, this has been officially allowed: https://github.com/apache/parquet-format/issues/296

arrow-rs has supported this for a while: https://github.com/apache/arrow-rs/issues/6048

However, duckdb will error: https://github.com/duckdb/duckdb/blob/cae7a680bc78d82d0a3f721291c3a36ce99c6561/extension/parquet/decoder/byte_stream_split_decoder.cpp#L34

### To Reproduce

```python
import duckdb
import pyarrow as pa
import pyarrow.parquet as pq
import numpy as np

# Create sample data with floating point values
data = {
    'id': np.arange(1000),
    'float_values': np.random.random(1000),
    'other_data': np.random.random(1000)
}

# Create a PyArrow table
table = pa.Table.from_pydict(data)

# Write the parquet file with the specified encoding
pq.write_table(
    table,
    'example_with_encoding.parquet',
    use_dictionary=False,
    use_byte_stream_split=True,
)

# Verify the encoding
metadata = pq.read_metadata('example_with_encoding.parquet')
for i in range(metadata.num_row_groups):
    row_group = metadata.row_group(i)
    for j in range(row_group.num_columns):
        col = row_group.column(j)
        print(f"Column: {col.path_in_schema}, Type: {col.physical_type}, Encoding: {col.encodings}")


print(duckdb.sql("select * from 'example_with_encoding.parquet'"))
```

Yields error:

```
duckdb.duckdb.Error: BYTE_STREAM_SPLIT encoding is only supported for FLOAT or DOUBLE data
```

### OS:

ubuntu 22

### DuckDB Version:

1.2.2

### DuckDB Client:

1.2.2

### Full Name:

Mark Polyakov

### Affiliation:

Aquatic Capital Management

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #9240, opened 2023-10-06
[Parquet] Implement reading byte stream split encoded data

adamreeve:
This adds support for reading byte stream split encoded Parquet data, fixing https://github.com/duckdb/duckdb/discussions/9104 (although that also mentions encode support, which this doesn't address).

PR #24175, opened 2026-07-26
[Parquet] Add BYTE_STREAM_SPLIT support for INT32 and INT64

hertelukas:
I've extended #9240 to support 32 and 64 bit integers. The decoding is type agnostic anyway, and supporting integers works basically out of the box.

This was also requested in #17114, however, I have not implemented parquet's `FIXED_LEN_BYTE_ARRAY` (which was introduced in parquet at the same time, see apache/parquet-format#229), as this would require a significantly more complex PR. If this should be implemented anyway to get this merged, I'll happily look into it.


data/parquet-testing/silly-names.parquet
Mytherin, line 1: Is this file edited on purpose?

Mytherin (commented):
Thanks! LGTM - one minor question

data/parquet-testing/silly-names.parquet
hertelukas: Yes, I ran `pyarrow-generate-parquet.py` to get the new test case, which also runs `generate_silly_names()`. I assume that due to a newer `pyarrow` version, the binary encoding has changed slightly. Do you want me to revert the change for this file?

Mytherin: That's alright - thanks

Mytherin:
Thanks!

hertelukas:
Thanks for the quick review and merge!

PR #24176, opened 2026-07-26
Fix quoting around extension semantic version pattern

jokasimr:
CMake recognizes double quotes as CMake syntax, but not single quotes. Therefore the single quotes become literal characters in the argument sent to Git.

The intended argument list is:

```
[git, describe, --tags, --always, --match, v*.*.*]
```

The actual argument list is effectively:

```
[git, describe, --tags, --always, --match, 'v*.*.*']
```

Since no tag matches the `'v.*.*.*'` pattern that makes the `EXT_VERSION_*` macro be the commit hash instead of the semantic version tag.


Mytherin:
Thanks!

PR #24177, opened 2026-07-26
Add shared memory management across DuckDB instances

dentiny:
Hi team, DuckDB could be used as an embedded database, which means one single process could have multiple db instances running. From my perspective, I think it'd be good if we have multiple levels of memory management, for example, operator-level, connection-level, instance-level, process-level.

This PR makes memory management shared among multiple database instances, the core idea is to leverage one single memory management class across multiple instances (to get responsible for de/allocation).

Key design decisions
- Introduce a new class `DatabaseMemoryManager` to manage all memory management related instances, including memory allocator and memory holder (i.e., buffer manager and object cache)
- But putting all entries (i.e., object cache entries) in a single instance exposes access issue, for example, one db instance could access another instance's object cache entries, so a memory context id is added. It only includes a database id for now, but could be extended to include connection id if needed. Also see https://github.com/duckdb/duckdb-quack/issues/133

Key code change
- DatabaseMemoryManager class is made to manage all memory related class; when creating a database instance, users could pass their shared memory manager (shared mode), or nothing (exclusive mode)
- On database instance destruction, only self memory items should be cleared without affecting other instances; object cache being an exception, non-evictable entries are deleted right upon instance dtor, while evictable entries rely on LRU cache to evict in a lazy manner.


Mytherin (commented):
Thanks for the PR!

I would prefer if we keep this opaque from an API perspective as much as possible, we could do this in a similar manner to how the `OpenerFileSystem` works, e.g. we can have:

```cpp
ObjectCache::Get(context)
```

Return a wrapper object cache that forwards the memory context info into the appropriate API methods. That prevents having to change all of the downstream code and makes all of this forwarding automatic from a consumer perspective.

dentiny:
> Return a wrapper object cache that forwards the memory context info into the appropriate API methods. That prevents having to change all of the downstream code and makes all of this forwarding automatic from a consumer perspective.

Yeah that makes sense to me -- updated a proxy class for ObjectCache, also hide all putter / getters so users don't misuse.

.github/patches/extensions/ducklake/0001-object-cache.patch
```diff
@@ -0,0 +1,63 @@
...
+diff --git a/src/storage/ducklake_catalog.cpp b/src/storage/ducklake_catalog.cpp
+index 1cbce0f8..9b794205 100644
+--- a/src/storage/ducklake_catalog.cpp
++++ b/src/storage/ducklake_catalog.cpp
+@@ -355,7 +355,7 @@ idx_t DuckLakeCatalog::GetBeginSnapshotForSchemaVersion(TableIndex table_id, idx
+ 
+ shared_ptr<DuckLakeSchemaCacheEntry> DuckLakeCatalog::GetSchemaCacheEntry(DuckLakeTransaction &transaction,
+                                                                           DuckLakeSnapshot snapshot) {
+-	auto &cache = GetObjectCacheInstance();
++	auto cache = ObjectCache::Get(GetDatabase());
```
Mytherin: We can keep a reference here if we construct it in the `ClientContext` or `DatabaseInstance` similar to what we do for file systems.

src/main/database.cpp
```diff
@@ -49,6 +51,11 @@
 
 namespace duckdb {
 
+static MemoryContextId NextMemoryContextId() {
+	static atomic<idx_t> next_database_id {0};
+	return MemoryContextId(next_database_id.fetch_add(1));
```
Mytherin: Can we use a uuid for this instead of a static global?

src/storage/buffer/buffer_pool.cpp
```diff
@@ -12,6 +14,17 @@
 
 namespace duckdb {
 
+static QueryContext GetBlockOwnerContext(QueryContext context, const BlockMemory &memory) {
```
Mytherin: Why do we need this?

Mytherin (commented):
Thanks for the PR! Some more comments from my side. Also let the clanker have a look and he had the following suggestions which seem reasonable to me:

### 1. Letting evictable entries outlive their `DatabaseInstance` is both pointless and unsafe

`NextMemoryContextId()` is monotonic and never reuses ids, so once db1 is gone, its entries are unreachable — no live database can produce a matching key — yet they still hold buffer-pool reservations until LRU pressure happens to evict them. The new test asserts this as intended (`test/api/test_object_cache.cpp`: after `first.reset()`, `first-evictable` still resolves and `GetCurrentMemory() == obj_size`).

Worse, it's a use-after-free surface:

```
ExternalFileCacheObjectCacheEntry -> CachedFile -> shared_ptr<BlockHandle> -> BlockMemory::buffer_manager
```

and `buffer_manager` is `reset()` on the line right after `DropNonEvictableEntries()` in 
`~DatabaseInstance`. The PR papers over exactly that one instance by adding `~ExternalFileCache` →
`SetEnabled(false)`. Nothing stops the next entry type (in-tree or extension) from dangling the same way.

**Recommendation:** drop *all* entries for the context on teardown (`DropEntries(context_id)` over both maps), delete the `~ExternalFileCache` workaround, and remove the "rely on internal LRU cache to evict them" note from the header. That eliminates the whole class of lifetime bug with no whack-a-mole.

If the lazy behavior is wanted for a reason I'm not seeing, please say why — and then `ObjectCacheEntry` needs a documented contract that entries must not capture per-instance state, plus a test that pressure-evicts an entry whose owner is already destroyed.

### 2. Context scoping via string prefix

`MakeCacheKey` builds `"<id>-<key>"` and `DropNonEvictableEntries` recovers ownership with `StartsWith("<id>-")` over the entire map. It is collision-free today only because ids never contain `-` — an implicit invariant on an unrelated format string.

Keying on `pair<idx_t, string>` or a per-context submap makes the drop O(entries for that context), removes the invariant, and removes a `StringUtil::Format` (two, for the typed variants) from every cache lookup.

### 3. Hot-path cost

- `Pin`/`Unpin` now branch and `throw InternalException` per call. A block is always pinned through the   buffer manager that created it, so this is a programmer-error check → `D_ASSERT`, not a throw on the   hottest path in the buffer manager.
- `GetBlockOwnerContext` is called per candidate block inside `IterateUnloadableBlocks`, but the caller's context id is loop-invariant. Hoist it out of the lambda and compare `idx_t`s instead of doing `GetClientContext()` + `GetDatabase()` per block.

### 4. `SetAllocator` / `ShareMemoryWith` are asymmetric

`SetAllocator` after `ShareMemoryWith` throws; `ShareMemoryWith` after `SetAllocator` silently
`memory_manager_options.reset()`s the user's allocator. Both orders should throw.

### 5. Two-headed ObjectCache API

`ObjectCache::GetObjectCache(ClientContext&)` and `DatabaseInstance::GetObjectCache()` remain public/`DUCKDB_API` but now expose only stats. The tests show the awkwardness — several hold a `cache`
and an `object_cache` side by side. Consider renaming the raw accessor to signal its role, or folding stats onto `BoundObjectCache`. `friend class DatabaseInstance` is needed only for `DropNonEvictableEntries`, which reads more naturally as a public admin method.

## Minor

- `memory_context.hpp` pulls `string_util.hpp` into `block_handle.hpp`, `object_cache.hpp` and  `database.hpp` purely for `ToString()`'s `StringUtil::Format`. Move `ToString` out of line to keep  those very widely included headers light.

PR #24185, opened 2026-07-27
Automatically promote `DECIMAL` past the `BIGINT` boundary

lnkuiper:
Instead of throwing an error suggesting to add a `CAST`.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/10051

EDIT: this causes some pretty severe regressions. I'll have to remove it for multiplication (because it promotes there way too often), but also adding statistics propagation for INT128 should help


Mytherin:
Thanks for the PR!

I wonder if we could fix some of these regressions by having separate optimizers that look at hugeint operations and convert them into non-hugeint operations by looking at the min/max stats.

lnkuiper:
I think that's a really good idea. I'll try and see what I can do in this PR (don't want to scope creep by implementing an optimiser if it's not yet necessary), and then we can think about a follow-up that re-enables the multiplication auto-promote and avoids the regression with an optimiser (maybe inside of `StatisticsPropagation`).

Mytherin:
In a way this would be kind of an extension of the compressed materialization optimizer - where we can pull that through arithmetic operators as well.

lnkuiper:
We could do compressed materialization for arithmetic operators too! I think this would have performance implications though, but those could be fixed: I've been wanting to create an optimiser pass after compressed materialization that removes redundant decompress + compress chains. This would let us have the smallest types throughout the entire plan instead of going back and forth.

Mytherin:
Also in the (far) future potentially pushing these casts into the scans would be interesting, as if we have e.g. bitpacked data we support scanning that into various different integer types, so there's no need to first scan into a hugeint and then extract the data from the hugeint, we could immediately scan into smaller integer types.

PR #24187, opened 2026-07-27
Parquet reader: count each column chunk once in the prefetch pre-check

### LINKED ISSUE #24186: Native-Iceberg UPDATE/MERGE fails with "incorrectly set page offsets" — MERGE target scan receives duplicated column_ids, parquet prefetch pre-check misfires
### What happens?

> [!NOTE]
> Filing this in core rather than `duckdb-iceberg`: the misfiring check is in the in-tree parquet
> extension and the duplicated bindings come from `MERGE INTO` planning — the Iceberg extension is
> only the reproduction vehicle (`MERGE INTO` needs a catalog table, so an ATTACHed Iceberg catalog
> is the easiest way to point it at parquet-over-httpfs).

An `UPDATE` (or `MERGE`) on an Iceberg table whose target scan keeps an **unpruned (full) projection**
crashes DuckDB's parquet **prefetch** reader:

```
IO Error: The parquet file '<data-file>' seems to have incorrectly set page offsets.
This interferes with DuckDB's prefetching optimization. DuckDB may still be able to scan this file
by manually disabling the prefetching mechanism using: 'SET disable_parquet_prefetching=true'.
```

The offending data file was written by **DuckDB's own native Iceberg writer**. A plain `SELECT *` of the
same table reads fine (locally, over S3, prefetch on or off). Only the `UPDATE`/`MERGE` **modify-scan**
over the prefetch path fails. All data below is synthetic.

**Smallest natural trigger found:** a `MERGE` whose source contains a **materialized CTE over the target**
(explicit `MATERIALIZED`, or simply a CTE referenced twice — which DuckDB auto-materializes). The
materialization defeats projection pushdown into the MERGE target scan, and the resulting full-width
prefetch modify-scan fails — **down to a 1-row, 2-column table**, with default optimizers.

**Environment**:

- **DuckDB 1.5.5** (latest release) — also reproduces on **1.5.4**.
- Extensions: `iceberg`, `httpfs`, `aws`.
- Iceberg **REST catalog** (Lakekeeper) + **S3** (MinIO). The bug is on the httpfs/S3 prefetch path.

#### Minimal SQL (knob-free, default optimizers)

```sql
-- against an ATTACHed Iceberg REST catalog over S3, extensions iceberg+httpfs+aws:
CREATE TABLE iceberg_catalog.demo.sample AS
  SELECT (i)::VARCHAR AS record_key, ('value_'||i)::VARCHAR AS attr_a
  FROM range(2000) t(i);                                          -- native Iceberg write

SELECT * FROM iceberg_catalog.demo.sample;                        -- OK (full-width read, prefetch on)
UPDATE iceberg_catalog.demo.sample SET record_key=record_key;     -- OK (optimizer prunes target scan)

-- A MERGE whose source contains a MATERIALIZED CTE over the target crashes with DEFAULT optimizers:
MERGE INTO iceberg_catalog.demo.sample t
USING (WITH cur AS MATERIALIZED (SELECT * FROM iceberg_catalog.demo.sample) SELECT * FROM cur) s
ON t.record_key = s.record_key
WHEN MATCHED THEN UPDATE SET attr_a = COALESCE(s.attr_a, t.attr_a)
WHEN NOT MATCHED THEN INSERT (record_key, attr_a) VALUES (s.record_key, s.attr_a);
-- ❌ IO Error: The parquet file '...' seems to have incorrectly set page offsets.
```

Characterization (each verified on a fresh catalog):

- Reproduces down to a **1-row, 2-column** table — so the file is clearly not actually corrupt.
- The explicit `MATERIALIZED` keyword is only for minimality: a CTE over the target **referenced twice**
  is auto-materialized and crashes identically (that is how our production query hits it — no hints,
  no knobs). `NOT MATERIALIZED` + double reference still crashes.
- The same CTE in a plain `SELECT` passes; `UPDATE ... WHERE k IN (WITH cur AS MATERIALIZED ...)`
  passes — the crash is specific to the **MERGE source** path.
- Equivalent knob form (same reader path, deterministic): `SET disabled_optimizers='unused_columns';
  UPDATE iceberg_catalog.demo.sample SET record_key=record_key;` — the knob is *not* the bug, it just
  forces the same unpruned target scan that CTE materialization causes naturally.

#### Self-contained setup (~1 min)

Create the three files below in an empty directory, then:

```bash
docker compose -p duckdbrepro up -d
# wait until the lakekeeper-init container has exited 0:
until [ "$(docker inspect -f '{{.State.Status}}' duckdbrepro-lakekeeper-init-1)" = exited ]; do sleep 2; done

docker run --rm --network duckdbrepro_default \
  -e S3_ENDPOINT_URL=minio:9000 -e AWS_ACCESS_KEY_ID=admin -e AWS_SECRET_ACCESS_KEY=password \
  -e LOCAL_ICEBERG_CATALOG_ENDPOINT=http://lakekeeper:8181/catalog \
  -v "$PWD":/repro:ro python:3.12-slim \
  bash -lc "pip install -q duckdb==1.5.5 && python /repro/REPRO_duckdb_iceberg_pageoffset.py"

docker compose -p duckdbrepro down -v   # teardown
```

Expected: the plain read and the *pruned* UPDATE pass; the **knob-free MERGE with the materialized-CTE
source** fails with `... incorrectly set page offsets` (and the knob-forced UPDATE fails identically).

<details>
<summary><code>docker-compose.yml</code></summary>

```yaml
services:
  minio:
    image: minio/minio:RELEASE.2024-01-31T20-20-33Z
    command: server /data --console-address ":9001"
    environment:
      MINIO_ROOT_USER: admin
      MINIO_ROOT_PASSWORD: password

  minio-init:
    image: minio/mc:latest
    depends_on: [minio]
    entrypoint: >
      /bin/sh -c "
      until (mc alias set loc http://minio:9000 admin password) do echo waiting for minio; sleep 1; done;
      mc mb -p loc/root || true;
      echo minio bucket ready"

  pg:
    image: postgres:16-alpine
    environment:
      POSTGRES_USER: lakekeeper
      POSTGRES_PASSWORD: lakekeeper
      POSTGRES_DB: lakekeeper
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U lakekeeper"]
      interval: 5s
      timeout: 3s
      retries: 20

  lakekeeper-migrate:
    image: quay.io/lakekeeper/catalog:latest
    depends_on:
      pg: {condition: service_healthy}
    environment:
      LAKEKEEPER__PG_DATABASE_URL_READ: "postgres://lakekeeper:lakekeeper@pg:5432/lakekeeper"
      LAKEKEEPER__PG_DATABASE_URL_WRITE: "postgres://lakekeeper:lakekeeper@pg:5432/lakekeeper"
      LAKEKEEPER__PG_ENCRYPTION_KEY: "ThisIsASecretEncryptionKey12345"
    command: ["migrate"]

  lakekeeper:
    image: quay.io/lakekeeper/catalog:latest
    depends_on:
      pg: {condition: service_healthy}
      lakekeeper-migrate: {condition: service_completed_successfully}
    environment:
      LAKEKEEPER__PG_DATABASE_URL_READ: "postgres://lakekeeper:lakekeeper@pg:5432/lakekeeper"
      LAKEKEEPER__PG_DATABASE_URL_WRITE: "postgres://lakekeeper:lakekeeper@pg:5432/lakekeeper"
      LAKEKEEPER__PG_ENCRYPTION_KEY: "ThisIsASecretEncryptionKey12345"
      LAKEKEEPER__BASE_URI: "http://lakekeeper:8181"
      RUST_LOG: info
    command: ["serve"]
    healthcheck:
      test: ["CMD", "/home/nonroot/lakekeeper", "healthcheck"]
      interval: 10s
      timeout: 5s
      retries: 10
      start_period: 10s

  lakekeeper-init:
    image: curlimages/curl:latest
    depends_on:
      lakekeeper: {condition: service_healthy}
      minio-init: {condition: service_completed_successfully}
    volumes:
      - ./init_lakekeeper.sh:/init_lakekeeper.sh:ro
    entrypoint: ["/bin/sh", "/init_lakekeeper.sh"]
```
</details>

<details>
<summary><code>init_lakekeeper.sh</code></summary>

```bash
#!/bin/sh
# Bootstrap Lakekeeper and create the 'default' warehouse backed by the MinIO 'root' bucket.
set -e
echo "Waiting for Lakekeeper..."
sleep 3

echo "Bootstrapping..."
curl -sS -X POST "http://lakekeeper:8181/management/v1/bootstrap" \
  -H "Content-Type: application/json" \
  -d '{"accept-terms-of-use": true}' || echo "(bootstrap may already be done)"

echo ""
echo "Creating warehouse 'default'..."
curl -sS -X POST "http://lakekeeper:8181/management/v1/warehouse" \
  -H "Content-Type: application/json" \
  -d '{
    "warehouse-name": "default",
    "project-id": "00000000-0000-0000-0000-000000000000",
    "storage-profile": {
      "type": "s3",
      "bucket": "root",
      "key-prefix": "",
      "region": "us-east-1",
      "flavor": "minio",
      "sts-enabled": false,
      "endpoint": "http://minio:9000",
      "path-style-access": true,
      "allow-alternative-protocols": true
    },
    "storage-credential": {
      "type": "s3",
      "credential-type": "access-key",
      "aws-access-key-id": "admin",
      "aws-secret-access-key": "password"
    }
  }' || echo "(warehouse may already exist)"
echo ""
echo "Lakekeeper init complete"
```
</details>

<details>
<summary><code>REPRO_duckdb_iceberg_pageoffset.py</code></summary>

```python
"""Minimal reproduction — DuckDB native-Iceberg UPDATE/MERGE:
"parquet file ... seems to have incorrectly set page offsets".

All data here is synthetic (fake table/column names and generated values).

TL;DR
-----
An UPDATE (or MERGE) on an Iceberg table whose target scan keeps an UNPRUNED (full) projection crashes
DuckDB's parquet *prefetch* reader:

    IO Error: The parquet file '<data file>' seems to have incorrectly set page offsets.
    This interferes with DuckDB's prefetching optimization. DuckDB may still be able to scan this file
    by manually disabling the prefetching mechanism using: 'SET disable_parquet_prefetching=true'.

The data file was written by DuckDB's own native Iceberg writer. A plain `SELECT *` of the same table
reads fine (local or over S3, prefetch on or off). Only the UPDATE/MERGE *modify scan* over the prefetch
path fails. The `unused_columns` optimizer normally prunes the target scan and hides the bug;
`SET disabled_optimizers='unused_columns'` reproduces the SAME unpruned target scan that a real query
whose plan legitimately keeps a full target projection produces on its own.

Reproduced on DuckDB 1.5.5 (latest release, 2026-07) against an Iceberg REST catalog (Lakekeeper) + S3
(MinIO); extensions iceberg, httpfs, aws. Also reproduces on 1.5.4. Minimal footprint: fails at 2
columns and 2,000 rows (and every larger size tried).

Env required: S3_ENDPOINT_URL, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, LOCAL_ICEBERG_CATALOG_ENDPOINT
(the REST catalog `.../catalog` endpoint of a warehouse named 'default').
"""
import os
import sys

import duckdb

TABLE = "iceberg_catalog.demo.sample"
N = int(os.environ.get("REPRO_N", "2000"))


def connect():
    conn = duckdb.connect()
    for ext in ("iceberg", "httpfs", "aws"):
        try:
            conn.load_extension(ext)
        except duckdb.IOException:
            conn.install_extension(ext, force_install=True)
            conn.load_extension(ext)
    conn.sql(f"SET s3_endpoint='{os.environ['S3_ENDPOINT_URL']}'; SET s3_url_style='path'; SET s3_use_ssl=false;")
    conn.sql(
        f"SET s3_access_key_id='{os.environ['AWS_ACCESS_KEY_ID']}'; "
        f"SET s3_secret_access_key='{os.environ['AWS_SECRET_ACCESS_KEY']}';"
    )
    conn.sql(
        f"ATTACH IF NOT EXISTS 'default' AS iceberg_catalog "
        f"(TYPE ICEBERG, ENDPOINT '{os.environ['LOCAL_ICEBERG_CATALOG_ENDPOINT']}', AUTHORIZATION_TYPE 'none')"
    )
    conn.sql("CREATE SCHEMA IF NOT EXISTS iceberg_catalog.demo")
    conn.sql("SET enable_external_file_cache=false")  # force real reads, not a warm cache
    return conn


def seed(conn):
    conn.sql(f"DROP TABLE IF EXISTS {TABLE}")
    # Native Iceberg write -> DuckDB-authored parquet data file(s). Fully synthetic values.
    conn.sql(
        f"CREATE TABLE {TABLE} AS "
        f"SELECT (i)::VARCHAR AS record_key, ('value_' || i)::VARCHAR AS attr_a FROM range({N}) t(i)"
    )


def show(label, fn):
    try:
        fn()
        print(f"[ OK ] {label}")
        return True
    except Exception as exc:  # noqa: BLE001
        po = "incorrectly set page offsets" in str(exc)
        tag = "PAGE-OFFSET BUG" if po else "other error"
        print(f"[FAIL] {label}  <-- {tag}\n       {type(exc).__name__}: {str(exc).splitlines()[0]}")
        return False


def main():
    print(f"duckdb {duckdb.__version__}, N={N}")
    conn = connect()
    seed(conn)

    # 1) Reading the file is fine — full projection, prefetch on.
    show("SELECT * FROM t                            (plain full-width read)",
         lambda: conn.sql(f"SELECT * FROM {TABLE}").fetchall())

    # 2) A natural UPDATE is fine — the optimizer prunes the target scan to just `record_key`.
    show("UPDATE t SET record_key=record_key         (optimizer prunes target scan)",
         lambda: conn.sql(f"UPDATE {TABLE} SET record_key=record_key"))

    # 3) KNOB-FREE: a MERGE whose source is a MATERIALIZED CTE over the target -> CRASH with default
    #    optimizers. (CTE materialization defeats projection pushdown into the target scan; reproduces
    #    even on a 1-row table. A CTE referenced twice is auto-materialized and crashes the same way.)
    ok_free = show(
        "MERGE with MATERIALIZED-CTE source           (knob-free crash, default optimizers)",
        lambda: conn.sql(
            f"MERGE INTO {TABLE} t "
            f"USING (WITH cur AS MATERIALIZED (SELECT * FROM {TABLE}) SELECT * FROM cur) s "
            f"ON t.record_key = s.record_key "
            f"WHEN MATCHED THEN UPDATE SET attr_a = COALESCE(s.attr_a, t.attr_a) "
            f"WHEN NOT MATCHED THEN INSERT (record_key, attr_a) VALUES (s.record_key, s.attr_a)"
        ),
    )

    # 4) The same unpruned target scan, forced explicitly via the optimizer knob -> same crash.
    conn.sql("SET disabled_optimizers='unused_columns'")
    ok_knob = show("UPDATE t SET record_key=record_key  [unpruned scan]   (same crash via the knob)",
                   lambda: conn.sql(f"UPDATE {TABLE} SET record_key=record_key"))
    conn.sql("SET disabled_optimizers=''")

    reproduced = not ok_free or not ok_knob
    print("\nRESULT:", "REPRODUCED the page-offset bug" if reproduced else "did NOT reproduce")
    return 0 if reproduced else 1  # exit 0 when the bug reproduces (this is a bug-repro script)


if __name__ == "__main__":
    sys.exit(main())
```
</details>

### Where the error is raised & suspected root cause (code analysis)

The exception comes from the prefetch pre-check in `extension/parquet/parquet_reader.cpp`
(`ParquetReader::Scan`, v1.5.5 ~L1418–1452):

```cpp
uint64_t to_scan_compressed_bytes = 0;
for (idx_t i = 0; i < column_ids.size(); i++) {
    auto col_idx = MultiFileLocalIndex(i);
    PrepareRowGroupBuffer(state, col_idx);
    auto file_col_idx = column_ids[col_idx];
    auto &root_reader = state.root_reader->Cast<StructColumnReader>();
    to_scan_compressed_bytes += root_reader.GetChildReader(file_col_idx).TotalCompressedSize();
}
...
uint64_t total_row_group_span = GetGroupSpan(state);
if (to_scan_compressed_bytes > total_row_group_span) {
    throw IOException("The parquet file '%s' seems to have incorrectly set page offsets. ...");
}
```

Key facts:

- `ColumnReader::TotalCompressedSize()` (`column_reader.cpp` ~L140) returns
  `chunk->meta_data.total_compressed_size` — the file's own metadata, no estimation.
- `GetGroupSpan()` (~L1095) is the physical byte span of the row group
  (max chunk end − min chunk start).
- Distinct column chunks occupy **disjoint** byte ranges inside that span, so for any
  duplicate-free `column_ids` the sum can never exceed the span.

For the failing file from the minimal repro the numbers make this concrete (2,000 rows, 1 row group):

| chunk | first page offset | total_compressed_size |
|---|---|---|
| `record_key` | 4 | 8,049 |
| `attr_a` | 8,053 | 8,045 |

`GetGroupSpan` = 16,094 and Σ(all distinct chunks) = **16,094 — exactly equal** (chunks are
back-to-back; DuckDB wrote this file itself and a plain `SELECT *` reads it fine). Therefore
`to_scan_compressed_bytes > total_row_group_span` can only happen if **`column_ids` contains the same
physical column more than once** — the loop then adds some chunk's size twice.

So the parquet file is healthy and the `IOException` misattributes the problem: the defect appears to be
in the **column binding of the MERGE target scan when the MERGE source contains a materialized CTE over
the target** (projection pushdown absent) — that plan hands the parquet scan a `column_ids` list with a
duplicated column. Two corroborating observations:

- With `SET disable_parquet_prefetching=true` (the message's suggested workaround) we have observed
  **silent byte misreads** (`Invalid string encoding found in Parquet file ... not valid UTF8`) in the
  knob-forced variant of this scan shape — so recommending that setting as a workaround is risky.
- A 1-column table does **not** reproduce; ≥2 columns always do (regardless of which side the
  `WHEN MATCHED SET` expression references) — consistent with a duplicate arising from the
  key/data column mapping rather than from any specific SET expression.

### Confirmed with a patched local build

We built `v1.5.5` from source with two changes to `ParquetReader::Scan` and re-ran the minimal repro:

1. **Diagnostic** — print the `column_ids` handed to each scan. Output for the failing statement:

   ```
   [PATCH] ParquetReader::Scan file=.../data/....parquet column_ids=[0,1]              <- CTE source scan (fine)
   [PATCH] ParquetReader::Scan file=.../data/....parquet column_ids=[0,1,0,2] <-- DUPLICATES   <- MERGE target scan
   ```

   The MERGE target scan indeed receives a **duplicated physical column**: `[0,1,0,2]` — file column
   `0` (the MERGE ON-key) appears twice (index `2` is the row-number child; the file has only two
   physical columns). Every other scan in the same statement gets a clean list. With
   `TotalCompressedSize(0)=8,049`, `(1)=8,045` and span `16,094`, the duplicated `0` makes the sum
   `> span` — precisely the misfired check.

2. **Fix** — count each *distinct* column once in the pre-check:

   ```cpp
   std::unordered_set<idx_t> counted_columns;
   for (idx_t i = 0; i < column_ids.size(); i++) {
       auto col_idx = MultiFileLocalIndex(i);
       PrepareRowGroupBuffer(state, col_idx);
       auto file_col_idx = column_ids[col_idx];
       auto &root_reader = state.root_reader->Cast<StructColumnReader>();
       if (counted_columns.insert(idx_t(file_col_idx)).second) {
           to_scan_compressed_bytes += root_reader.GetChildReader(file_col_idx).TotalCompressedSize();
       }
   }
   ```

   With this change the minimal repro **completes, with verified-correct results**: the merge sets
   `v = COALESCE(s.v, t.v)` with source == target, so the table must come out unchanged — and it does
   (`count(*)` and a full-table hash checksum are identical before/after; per-row spot-check clean).
   So in this shape the duplicated entries are benign for the actual read path — only the prefetch
   accounting breaks — though de-duplicating the binding itself still seems like the cleaner fix.

This also explains the 1-column boundary: with a single data column no duplicate arises
(`column_ids=[0,1]`, the `1` being the row-number child), so the check passes.

### Suggested fix

1. **Root cause**: de-duplicate / fix the `column_ids` built for the MERGE target scan on this plan
   shape (MERGE whose source contains a materialized CTE scanning the target), so target columns
   referenced by the join key / SET expressions reuse the existing binding instead of being appended
   again. (Likely in the MERGE INTO planning or the multi-file column-mapping layer rather than in the
   parquet extension itself.)
2. **Defensive**: compute the prefetch pre-check over *unique* `file_col_idx` (and `D_ASSERT`
   uniqueness of `column_ids` at scan init) so a plan-side duplication cannot masquerade as file
   corruption. **Verified in a locally patched v1.5.5 build** (see "Confirmed with a patched local
   build" above): with this one change the repro completes and the results are correct.
3. **Error message**: stop recommending `SET disable_parquet_prefetching=true` here — in this state it
   converts a loud error into silently wrong results.

> [!IMPORTANT]
> **Fix PR**: #24187 implements the defensive de-duplication (2) against `main`, ported to
> `ParquetReader::RegisterRowGroupReads`. The root-cause option (1) is offered there as a follow-up.

### Still present on current `main` (by inspection)

The pre-check has since moved to `ParquetReader::RegisterRowGroupReads` (main, ~L1881), but the
accounting is unchanged — it sums `TotalCompressedSize()` once per `column_ids` **entry**, and the
per-entry `column_readers` vector means duplicated entries still double-count the same physical chunk:

```cpp
uint64_t to_scan_compressed_bytes = 0;
for (idx_t i = 0; i < column_ids.size(); i++) {
	auto col_idx = MultiFileLocalIndex(i);
	PrepareRowGroupBuffer(context, state, col_idx);
	to_scan_compressed_bytes += state.GetColumnReader(i).TotalCompressedSize();
}
```

(On `main`, local files now fall back to synchronous reads when the check fires; **remote files still
throw** — and the httpfs/S3 path is exactly where this bug bites.)

We could not *run* the repro on a recent nightly in our environment for an unrelated reason: the
nightly Iceberg writer (tested: `1.6.0.dev291`) emits avro with `codec: deflate`, which Lakekeeper
rejects at `CREATE TABLE` — so the repro cannot even seed the table there.

### Production-shaped reproduction (fully synthetic data)

The 6-line minimal repro above was distilled from this **production-shaped** statement, which also
reproduces with default optimizers on entirely synthetic tables (fake column names `c00..c61`,
generated values). Its source is a multi-CTE query whose CTEs over the target/source tables are
**referenced multiple times — hence auto-materialized**, which is exactly the minimal trigger. The
optimizer leaves the MERGE target scan **unpruned** (its `ICEBERG_SCAN` carries no `Projections:`
list), and the merge fails with the page-offset error while re-reading the target's own **freshly
written** data file. Kept here to show a realistic query shape that hits this in practice.

Verified on the fresh stack from the compose file above (`python:3.12-slim`, `duckdb==1.5.5`):

```bash
# with the four files below in $PWD, after `docker compose -p duckdbrepro up -d`:
docker run --rm --network duckdbrepro_default \
  -e S3_ENDPOINT_URL=minio:9000 -e AWS_ACCESS_KEY_ID=admin -e AWS_SECRET_ACCESS_KEY=password \
  -e LOCAL_ICEBERG_CATALOG_ENDPOINT=http://lakekeeper:8181/catalog \
  -v "$PWD":/repro:ro python:3.12-slim \
  bash -lc "pip install -q duckdb==1.5.5 && python /repro/run_synthetic_repro.py"
# -> RESULT: REPRODUCED the page-offset bug
```

Notes:

- **File/metadata provenance is irrelevant**: the crash occurs on a data file the repro itself wrote
  moments earlier (plain `CREATE TABLE ... AS SELECT ... FROM range(...)`).
- A plain `SELECT *` of the same target passes; `UPDATE target SET c00=c00` (pruned scan) passes.
- The reliable crash predictor is the plan: whenever the MERGE target's `ICEBERG_SCAN` has **no
  `Projections:`** list, the statement fails; when a projection list is pushed, it succeeds.

<details>
<summary><code>run_synthetic_repro.py</code> — the runner</summary>

```python
"""Runner for the fully-synthetic knob-free repro of DuckDB's native-Iceberg MERGE
"incorrectly set page offsets" crash. All data/columns synthetic; DEFAULT optimizers (no knobs).

Env: S3_ENDPOINT_URL, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, LOCAL_ICEBERG_CATALOG_ENDPOINT.
Files (same dir): create_syn_target.sql, create_syn_agg.sql, syn_merge.sql
"""
import os
import sys

import duckdb

HERE = os.path.dirname(os.path.abspath(__file__))


def main():
    print(f"duckdb {duckdb.__version__}")
    conn = duckdb.connect()
    for ext in ("iceberg", "httpfs", "aws"):
        try:
            conn.load_extension(ext)
        except duckdb.IOException:
            conn.install_extension(ext, force_install=True)
            conn.load_extension(ext)
    conn.sql(f"SET s3_endpoint='{os.environ['S3_ENDPOINT_URL']}'; SET s3_url_style='path'; SET s3_use_ssl=false;")
    conn.sql(
        f"SET s3_access_key_id='{os.environ['AWS_ACCESS_KEY_ID']}'; "
        f"SET s3_secret_access_key='{os.environ['AWS_SECRET_ACCESS_KEY']}';"
    )
    conn.sql(
        f"ATTACH 'default' AS iceberg_catalog "
        f"(TYPE ICEBERG, ENDPOINT '{os.environ['LOCAL_ICEBERG_CATALOG_ENDPOINT']}', AUTHORIZATION_TYPE 'none')"
    )
    conn.sql("CREATE SCHEMA IF NOT EXISTS iceberg_catalog.demo")
    conn.sql("SET enable_external_file_cache=false")
    conn.sql("SET disabled_optimizers=''")  # default optimizers — NO knob

    conn.sql("DROP TABLE IF EXISTS iceberg_catalog.demo.target_table")
    conn.sql("DROP TABLE IF EXISTS iceberg_catalog.demo.source_batch")
    conn.sql(open(os.path.join(HERE, "create_syn_target.sql")).read())
    conn.sql(open(os.path.join(HERE, "create_syn_agg.sql")).read())
    print("target rows:", conn.sql("SELECT count(*) FROM iceberg_catalog.demo.target_table").fetchone()[0])
    print("source rows:", conn.sql("SELECT count(*) FROM iceberg_catalog.demo.source_batch").fetchone()[0])

    merge_sql = open(os.path.join(HERE, "syn_merge.sql")).read()
    try:
        conn.sql(merge_sql)
        print("RESULT: PASS (did NOT reproduce)")
        return 1
    except Exception as exc:  # noqa: BLE001
        msg = f"{type(exc).__name__}: {exc}"
        if "incorrectly set page offsets" in msg:
            print("RESULT: REPRODUCED the page-offset bug")
            print(msg.splitlines()[0][:300])
            return 0
        print(f"RESULT: OTHER FAILURE\n{msg[:400]}")
        return 2


if __name__ == "__main__":
    sys.exit(main())
```
</details>

<details>
<summary><code>create_syn_target.sql</code> — 36-column synthetic MERGE target</summary>

```sql
CREATE TABLE iceberg_catalog.demo.target_table AS SELECT
  ('k' || i)::VARCHAR AS c00,
  (TIMESTAMPTZ '2026-01-01 00:00:00+00' + (i * INTERVAL 1 SECOND)) AS c01,
  (TIMESTAMPTZ '2026-01-01 00:00:00+00' + (i * INTERVAL 1 SECOND)) AS c02,
  (TIMESTAMPTZ '2026-01-01 00:00:00+00' + (i * INTERVAL 1 SECOND)) AS c03,
  ['k' || i]::VARCHAR[] AS c04,
  ['c05_a' || i]::VARCHAR[] AS c05,
  (TIMESTAMPTZ '2026-01-01 00:00:00+00' + (i * INTERVAL 1 SECOND)) AS c06,
  ['c07_a' || i]::VARCHAR[] AS c07,
  ['c08_a' || i]::VARCHAR[] AS c08,
  ('c09_v' || i)::VARCHAR AS c09,
  ('c10_v' || i)::VARCHAR AS c10,
  ((i) % 256)::INTEGER AS c11,
  ['k' || i]::VARCHAR[] AS c12,
  ['c13_a' || i]::VARCHAR[] AS c13,
  ['c14_a' || i]::VARCHAR[] AS c14,
  ['c15_a' || i]::VARCHAR[] AS c15,
  ['c16_a' || i]::VARCHAR[] AS c16,
  CAST(NULL AS INTEGER[]) AS c17,
  ('k' || i)::VARCHAR AS c18,
  ['c19_a' || i]::VARCHAR[] AS c19,
  ['c20_a' || i]::VARCHAR[] AS c20,
  ('c21_v' || i)::VARCHAR AS c21,
  ((i) % 2 = 0) AS c22,
  ['c23_a' || i]::VARCHAR[] AS c23,
  ['c24_a' || i]::VARCHAR[] AS c24,
  ['c25_a' || i]::VARCHAR[] AS c25,
  ['c26_a' || i]::VARCHAR[] AS c26,
  ['c27_a' || i]::VARCHAR[] AS c27,
  ['c28_a' || i]::VARCHAR[] AS c28,
  ['c29_a' || i]::VARCHAR[] AS c29,
  ('c30_v' || i)::VARCHAR AS c30,
  ['c31_a' || i]::VARCHAR[] AS c31,
  ['c32_a' || i]::VARCHAR[] AS c32,
  ('c33_v' || i)::VARCHAR AS c33,
  ('c34_v' || i)::VARCHAR AS c34,
  ['k' || i]::VARCHAR[] AS c35
FROM range(2465) t(i);
```
</details>

<details>
<summary><code>create_syn_agg.sql</code> — 40-column synthetic source table</summary>

```sql
CREATE TABLE iceberg_catalog.demo.source_batch AS SELECT
  ('c07_v' || i)::VARCHAR AS c07,
  ('c05_v' || i)::VARCHAR AS c05,
  ('k' || i)::VARCHAR AS c12,
  ('c36_v' || i)::VARCHAR AS c36,
  ('c37_v' || i)::VARCHAR AS c37,
  ((i) % 256)::INTEGER AS c38,
  ('k' || i)::VARCHAR AS c04,
  ('c34_v' || i)::VARCHAR AS c34,
  ('c39_v' || i)::VARCHAR AS c39,
  (TIMESTAMPTZ '2026-01-01 00:00:00+00' + (i * INTERVAL 1 SECOND)) AS c40,
  (TIMESTAMPTZ '2026-01-01 00:00:00+00' + (i * INTERVAL 1 SECOND)) AS c41,
  (TIMESTAMPTZ '2026-01-01 00:00:00+00' + (i * INTERVAL 1 SECOND)) AS c42,
  ('c43_v' || i)::VARCHAR AS c43,
  ('c09_v' || i)::VARCHAR AS c09,
  ('c44_v' || i)::VARCHAR AS c44,
  ('c13_v' || i)::VARCHAR AS c13,
  ('c14_v' || i)::VARCHAR AS c14,
  ('c15_v' || i)::VARCHAR AS c15,
  ('c16_v' || i)::VARCHAR AS c16,
  CAST(NULL AS INTEGER[]) AS c17,
  ('k' || i)::VARCHAR AS c18,
  ('c19_v' || i)::VARCHAR AS c19,
  ('c20_v' || i)::VARCHAR AS c20,
  ('c21_v' || i)::VARCHAR AS c21,
  ((i) % 2 = 0) AS c22,
  (TIMESTAMPTZ '2026-01-01 00:00:00+00' + (i * INTERVAL 1 SECOND)) AS c45,
  ('c08_v' || i)::VARCHAR AS c08,
  ('c23_v' || i)::VARCHAR AS c23,
  ('c24_v' || i)::VARCHAR AS c24,
  ('c25_v' || i)::VARCHAR AS c25,
  ('c26_v' || i)::VARCHAR AS c26,
  ('c27_v' || i)::VARCHAR AS c27,
  ('c28_v' || i)::VARCHAR AS c28,
  ('c29_v' || i)::VARCHAR AS c29,
  ('c30_v' || i)::VARCHAR AS c30,
  ('c31_v' || i)::VARCHAR AS c31,
  ('c32_v' || i)::VARCHAR AS c32,
  ('c33_v' || i)::VARCHAR AS c33,
  ('k' || i)::VARCHAR AS c35
FROM range(3853) t(i);
```
</details>

<details>
<summary><code>syn_merge.sql</code> — the MERGE statement (anonymized production shape)</summary>

```sql
MERGE INTO iceberg_catalog.demo.target_table t
USING (WITH cte_batch AS (SELECT * FROM  iceberg_catalog.demo.source_batch),
cte_current AS (SELECT * FROM  iceberg_catalog.demo.target_table),

  cte_r1 AS (
  SELECT
    c04,
    c05,
    c30,
    ROW_NUMBER() OVER (
      PARTITION BY c04
      ORDER BY CASE
              
              WHEN c05 = 'vendor_a' THEN 0
              WHEN c05 = 'vendor_b' THEN 1
              
          END
    ) AS rank
    FROM (
      SELECT
        c04,
        c05,
        COALESCE(
          MAX(
            CASE
              WHEN c30 = 'STATE_X'
              THEN 'STATE_X' 
              WHEN c30 = 'State_x'
              THEN 'STATE_X' 
            END
          ) ,
          MAX(c30) 
        ) AS c30
      FROM (
        SELECT
          c04,
          c05,
          c12,
          c30,
          ROW_NUMBER() OVER (PARTITION BY c04, c05, c12 ORDER BY c40 DESC) AS rn
        FROM cte_batch
        
        WHERE
          (
            c40 > '2026-07-21 13:48:48.029486'
          )
          AND NOT c30 IS NULL
      )
      WHERE
        rn = 1
      GROUP BY
        c04,
        c05
    )
  ), cte_pick1 AS (
    SELECT
      c04,
      c05,
      c30
    FROM cte_r1
    WHERE
      rank = 1
  ), cte_r2 AS (
    
    SELECT
      c04,
      c05,
      c34,
      ROW_NUMBER() OVER (
        PARTITION BY c04, c05
        ORDER BY c42 DESC
      ) AS rank
    FROM cte_batch
    WHERE
      (
        c40 > '2026-07-24 13:39:05.921633'
      )
      AND NULLIF(c34, '') IS NOT NULL
  ), cte_pick2 AS (
    SELECT
      c04,
      c05,
      c34
    FROM cte_r2
    WHERE
      rank = 1
  ), cte_existing_keys AS (
    SELECT
      UNNEST(c04) AS c04,
      c18,
      c10
    FROM cte_current
  ), cte_prep AS (
    SELECT
      SHA256(CONCAT('|', ap.c04, ap.c05)) AS c00,
      ap.c04,
      ap.c05,
      ARRAY_AGG(DISTINCT ap.c07) FILTER(WHERE   ap.c07 IS NOT NULL) AS c07,
      ARRAY_AGG(DISTINCT ap.c09) FILTER(WHERE   ap.c09 IS NOT NULL) AS c09,
      COALESCE(
        ANY_VALUE(CASE WHEN c08 LIKE '% %' THEN c08 END),
        ANY_VALUE(CASE WHEN c08 LIKE '%@%' THEN c08 END),
        ANY_VALUE(c08)
      ) AS c08,
      ANY_VALUE(c44) AS c44,
      MIN(LOWER(ap.c18)) AS email_prep,
      ANY_VALUE(c16) AS c16,
      ARRAY_AGG(DISTINCT c13) FILTER(WHERE   c13 IS NOT NULL) AS c13,
      ARRAY_AGG(DISTINCT c14) FILTER(WHERE   c14 IS NOT NULL) AS c14,
      ARRAY_AGG(DISTINCT c15) FILTER(WHERE   c15 IS NOT NULL) AS c15,
      FLATTEN(ARRAY_AGG(DISTINCT c17) FILTER(WHERE   c17 IS NOT NULL)) AS c17,
      ARRAY_AGG(DISTINCT c19) FILTER(WHERE   c19 IS NOT NULL) AS c19,
      ARRAY_AGG(DISTINCT c20) FILTER(WHERE   c20 IS NOT NULL) AS c20,
      MAX(c21) AS c21,
      BOOL_OR(c22) AS c22,
      COALESCE(
        MAX(
          CASE
            WHEN cis.c30 = 'STATE_X'
            THEN 'STATE_X' 
            WHEN cis.c30 = 'State_x'
            THEN 'STATE_X' 
          END
        ) ,
        MAX(cis.c30) 
      ) AS idp_status_prep,
      ARRAY_AGG(DISTINCT c23) FILTER(WHERE   c23 IS NOT NULL) AS c23,
      ARRAY_AGG(DISTINCT c24) FILTER(WHERE   c24 IS NOT NULL) AS c24,
      ARRAY_AGG(DISTINCT c25) FILTER(WHERE   c25 IS NOT NULL) AS c25,
      ARRAY_AGG(DISTINCT c26) FILTER(WHERE   c26 IS NOT NULL) AS c26,
      ARRAY_AGG(DISTINCT c27) FILTER(WHERE   c27 IS NOT NULL) AS c27,
      ARRAY_AGG(DISTINCT c28) FILTER(WHERE   c28 IS NOT NULL) AS c28,
      ARRAY_AGG(DISTINCT c29) FILTER(WHERE   c29 IS NOT NULL) AS c29,
      ARRAY_AGG(DISTINCT c31) FILTER(WHERE   c31 IS NOT NULL) AS c31,
      ARRAY_AGG(DISTINCT c32) FILTER(WHERE   c32 IS NOT NULL) AS c32,
      ARRAY_AGG(DISTINCT c35) FILTER(WHERE   c35 IS NOT NULL) AS c35,
      ANY_VALUE(ap.c33) AS c33,
      ANY_VALUE(caf.c34) AS c34,
      ARRAY_AGG(DISTINCT c12) FILTER(WHERE   c12 IS NOT NULL) AS c12,
      MIN(c42) AS c02,
      MAX(c42) AS c03
    FROM cte_batch AS ap
    LEFT JOIN cte_pick1 AS cis
      ON ap.c04 = cis.c04 AND ap.c05 = cis.c05
    LEFT JOIN cte_pick2 AS caf
      ON ap.c04 = caf.c04 AND ap.c05 = caf.c05
    WHERE
      (
        c40 > '2026-07-24 13:39:05.921633'
      )
    GROUP BY
      ap.c04,
      ap.c05
  ), prep AS(
    SELECT ANY_VALUE(c00) AS c00,
           prep.c04,
           prep.c05,
           ANY_VALUE(c07) AS c07,
           ARRAY_DISTINCT(FLATTEN(ARRAY_AGG(c09))) AS c09,
           COALESCE(
                    ANY_VALUE(CASE WHEN c08 LIKE '% %' THEN c08 END),
                    ANY_VALUE(CASE WHEN e.c10 LIKE '% %' THEN e.c10 END),
                    ANY_VALUE(CASE WHEN c08 LIKE '%@%' THEN c08 END),
                    ANY_VALUE(CASE WHEN e.c10 LIKE '%@%' THEN e.c10 END),
                    ANY_VALUE(c08),
                    ANY_VALUE(e.c10)
            ) AS c08,
           ANY_VALUE(c44) AS c44,
           ANY_VALUE(email_prep) AS email_prep,
           ANY_VALUE(c16) AS c16,
           ANY_VALUE(c13) AS c13,
           ANY_VALUE(c14) AS c14,
           ANY_VALUE(c15) AS c15,
           ANY_VALUE(c17) AS c17,
           ANY_VALUE(c19) AS c19,
           ANY_VALUE(c20) AS c20,
           ANY_VALUE(c21) AS c21,
           ANY_VALUE(c22) AS c22,
           ANY_VALUE(idp_status_prep) AS idp_status_prep,
           ANY_VALUE(c23) AS c23,
           ANY_VALUE(c24) AS c24,
           ANY_VALUE(c25) AS c25,
           ANY_VALUE(c26) AS c26,
           ANY_VALUE(c27) AS c27,
           ANY_VALUE(c28) AS c28,
           ANY_VALUE(c29) AS c29,
           ANY_VALUE(c31) AS c31,
           ANY_VALUE(c32) AS c32,
           ANY_VALUE(c35) AS c35,
           ANY_VALUE(c33) AS c33,
           ANY_VALUE(c34) AS c34,
           ANY_VALUE(c12) AS c12,
           ANY_VALUE(c02) AS c02,
           ANY_VALUE(c03) AS c03,
           COALESCE(MIN(email_prep), MIN(c18)) AS c18
    FROM cte_prep AS prep
    LEFT JOIN cte_existing_keys AS e
      ON prep.c04 = e.c04
      GROUP BY
      prep.c04,
        prep.c05
  ),
  cte_key AS (
    SELECT
      COALESCE(CASE WHEN c18 = '' THEN NULL ELSE LOWER(c18) END, TRY_CAST(c00 AS TEXT)) AS merge_key,
      *
    FROM prep
  ),
  cte_agg AS (
    SELECT
      merge_key,
      ARRAY_AGG(DISTINCT c04) FILTER(WHERE   c04 IS NOT NULL) AS c04,
      ARRAY_AGG(DISTINCT c05) FILTER(WHERE   c05 IS NOT NULL) AS c05,
      ARRAY_DISTINCT(FLATTEN(ARRAY_AGG(DISTINCT c09))) AS entity_type_distinct,
      FILTER(
        FLATTEN(
          ARRAY_AGG(DISTINCT c09) FILTER(WHERE   c09 IS NOT NULL)
        ),
        x -> x <> 'person_or_service_account'
      ) AS entity_type_non_person,
      FLATTEN(
        ARRAY_AGG(DISTINCT c07) FILTER(WHERE   c07 IS NOT NULL)
      ) AS c07,
      FILTER(
        ARRAY_AGG(DISTINCT c08) FILTER(WHERE   c08 IS NOT NULL),
        x -> x <> ''
      ) AS c08,
      LOWER(MIN(c18)) AS c18,
      ARRAY_AGG(DISTINCT c16) FILTER(WHERE   c16 IS NOT NULL) AS c16,
      FLATTEN(ARRAY_AGG(DISTINCT c13) FILTER(WHERE   c13 IS NOT NULL)) AS c13,
      FLATTEN(ARRAY_AGG(DISTINCT c14) FILTER(WHERE   c14 IS NOT NULL)) AS c14,
      FLATTEN(ARRAY_AGG(DISTINCT c15) FILTER(WHERE   c15 IS NOT NULL)) AS c15,
      FLATTEN(ARRAY_AGG(DISTINCT c17) FILTER(WHERE   c17 IS NOT NULL)) AS c17,
      FLATTEN(ARRAY_AGG(DISTINCT c19) FILTER(WHERE   c19 IS NOT NULL)) AS c19,
      FLATTEN(ARRAY_AGG(DISTINCT c20) FILTER(WHERE   c20 IS NOT NULL)) AS c20,
      MAX(c21) AS c21,
      COALESCE(
        MAX(
          CASE
            WHEN idp_status_prep = 'STATE_X'
            THEN 'STATE_X' 
            WHEN idp_status_prep = 'State_x'
            THEN 'STATE_X' 
          END
        ) ,
        MAX(idp_status_prep) 
      ) AS c30,
      CASE
        WHEN c30 = 'STATE_X'
        THEN TRUE
        WHEN BOOL_OR(c22) IS NULL
        THEN TRUE
        ELSE BOOL_OR(c22)
      END AS c22,
      FLATTEN(ARRAY_AGG(DISTINCT c23) FILTER(WHERE   c23 IS NOT NULL)) AS c23,
      MIN(c02) AS c02,
      MAX(c03) AS c03,
      ANY_VALUE(
          CASE
          
          WHEN c05='vendor_a' AND idp_status_prep = 'STATE_X' AND c08 IS NOT NULL AND c08 <> '' AND c08 LIKE '% %' THEN c08
          WHEN c05='vendor_b' AND idp_status_prep = 'STATE_X' AND c08 IS NOT NULL AND c08 <> '' AND c08 LIKE '% %' THEN c08
        END
      ) AS calculated_display_name_idp,
      ANY_VALUE(CASE WHEN c08 LIKE '% %' THEN c08 END) AS calculated_display_name_with_space,
      ANY_VALUE(CASE WHEN c08 LIKE '%@%' THEN c08 END) AS calculated_display_name_with_at,
      FILTER(
        ARRAY_AGG(c08) FILTER(WHERE c08 IS NOT NULL),
        x -> NOT x IS NULL AND x <> ''
      ) AS calculated_display_name_candidates,
      FILTER(
        ARRAY_AGG(c16) FILTER(WHERE c16 IS NOT NULL),
        x -> NOT x IS NULL AND x <> ''
      ) AS calculated_username_candidates,
      FILTER(
        ARRAY_AGG(LOWER(c18)) FILTER(WHERE LOWER(c18) IS NOT NULL),
        x -> NOT x IS NULL AND x <> ''
      ) AS calculated_email_candidates,
      FLATTEN(ARRAY_AGG(DISTINCT c24) FILTER(WHERE   c24 IS NOT NULL)) AS c24,
      FLATTEN(ARRAY_AGG(DISTINCT c25) FILTER(WHERE   c25 IS NOT NULL)) AS c25,
      FLATTEN(ARRAY_AGG(DISTINCT c26) FILTER(WHERE   c26 IS NOT NULL)) AS c26,
      FLATTEN(ARRAY_AGG(DISTINCT c27) FILTER(WHERE   c27 IS NOT NULL)) AS c27,
      FLATTEN(ARRAY_AGG(DISTINCT c28) FILTER(WHERE   c28 IS NOT NULL)) AS c28,
      FLATTEN(ARRAY_AGG(DISTINCT c29) FILTER(WHERE   c29 IS NOT NULL)) AS c29,
      FILTER(
        FLATTEN(ARRAY_AGG(DISTINCT c31) FILTER(WHERE   c31 IS NOT NULL)),
        c -> c <> ''
      ) AS c31,
      FILTER(
        FLATTEN(ARRAY_AGG(DISTINCT c32) FILTER(WHERE   c32 IS NOT NULL)),
        r -> r <> ''
      ) AS c32,
      FILTER(
        FLATTEN(ARRAY_AGG(DISTINCT c35) FILTER(WHERE   c35 IS NOT NULL)),
        a -> a <> ''
      ) AS c35,
      ANY_VALUE(c33) AS c33,
      ANY_VALUE(c34) AS c34,
      ARRAY_DISTINCT(FLATTEN(ARRAY_AGG(c12) FILTER(WHERE c12 IS NOT NULL))) AS c12
    FROM cte_key
    GROUP BY
      merge_key
  )
  SELECT
    SHA256(merge_key) AS c00,
    TRY_CAST(TRY_CAST(SUBSTRING(HEX(SHA256(merge_key)), 1, 16) AS BIGINT) % 256 AS INT) AS c11,
    c04,
    c05,
    CASE
      WHEN entity_type_distinct = ['person_or_service_account'] THEN 'person_or_service_account'
      WHEN array_length(entity_type_non_person) >= 1 THEN entity_type_non_person[1]
      ELSE NULL
    END AS c09,
    c07,
    c08,
    c18,
    c16,
    c13,
    c14,
    c15,
    c17,
    c19,
    c20,
    c21,
    c30,
    c22,
    c23,
    c02,
    c03,
    CURRENT_TIMESTAMP AS c06,
    CURRENT_TIMESTAMP AS c01,
    COALESCE(
      calculated_display_name_idp,
      calculated_display_name_with_space,
      calculated_display_name_with_at,
      CASE WHEN array_length(calculated_display_name_candidates) >= 1 THEN calculated_display_name_candidates[1] ELSE NULL END,
      CASE WHEN array_length(calculated_username_candidates) >= 1 THEN calculated_username_candidates[1] ELSE NULL END,
      CASE WHEN array_length(calculated_email_candidates) >= 1 THEN calculated_email_candidates[1] ELSE NULL END
    ) AS c10,
    c24,
    c25,
    c26,
    c27,
    c28,
    c29,
    c31,
    c32,
    c33,
    c34,
    c35,
    c12
  FROM cte_agg) s
ON t.c00 = s.c00
WHEN MATCHED THEN UPDATE SET c01 = COALESCE(s.c01, t.c01), c03 = COALESCE(s.c03, t.c03), c04 = ARRAY_DISTINCT(CONCAT(s.c04, t.c04)), c05 = ARRAY_DISTINCT(CONCAT(s.c05, t.c05)), c06 = COALESCE(s.c06, t.c06), c07 = ARRAY_DISTINCT(CONCAT(s.c07, t.c07)), c08 = ARRAY_DISTINCT(CONCAT(s.c08, t.c08)), c09 = CASE
            WHEN t.c09 IS NULL THEN s.c09
            WHEN s.c09 IS NULL THEN t.c09
            WHEN s.c09 = 'person_or_service_account' THEN t.c09
            ELSE s.c09
            END
            , c10 = COALESCE(s.c10, t.c10), c11 = COALESCE(s.c11, t.c11), c12 = ARRAY_DISTINCT(CONCAT(s.c12, t.c12)), c13 = ARRAY_DISTINCT(CONCAT(s.c13, t.c13)), c14 = ARRAY_DISTINCT(CONCAT(s.c14, t.c14)), c15 = ARRAY_DISTINCT(CONCAT(s.c15, t.c15)), c16 = ARRAY_DISTINCT(CONCAT(s.c16, t.c16)), c17 = ARRAY_DISTINCT(CONCAT(s.c17, t.c17)), c18 = COALESCE(s.c18, t.c18), c19 = ARRAY_DISTINCT(CONCAT(s.c19, t.c19)), c20 = ARRAY_DISTINCT(CONCAT(s.c20, t.c20)), c21 = COALESCE(s.c21, t.c21), c23 = ARRAY_DISTINCT(CONCAT(s.c23, t.c23)), c24 = ARRAY_DISTINCT(CONCAT(s.c24, t.c24)), c25 = ARRAY_DISTINCT(CONCAT(s.c25, t.c25)), c26 = ARRAY_DISTINCT(CONCAT(s.c26, t.c26)), c27 = ARRAY_DISTINCT(CONCAT(s.c27, t.c27)), c28 = ARRAY_DISTINCT(CONCAT(s.c28, t.c28)), c29 = ARRAY_DISTINCT(CONCAT(s.c29, t.c29)), c31 = ARRAY_DISTINCT(CONCAT(s.c31, t.c31)), c32 = ARRAY_DISTINCT(CONCAT(s.c32, t.c32)), c33 = COALESCE(s.c33, t.c33), c34 = COALESCE(s.c34, t.c34), c35 = ARRAY_DISTINCT(CONCAT(s.c35, t.c35)), c30 = CASE WHEN s.c03 >= t.c03 THEN s.c30 ELSE t.c30 END, c22 = CASE WHEN s.c03 >= t.c03 THEN s.c22 ELSE t.c22 END
WHEN NOT MATCHED THEN INSERT (c00, c01, c02, c03, c04, c05, c06, c07, c08, c09, c10, c11, c12, c13, c14, c15, c16, c17, c18, c19, c20, c21, c22, c23, c24, c25, c26, c27, c28, c29, c30, c31, c32, c33, c34, c35) VALUES (s.c00, s.c01, s.c02, s.c03, s.c04, s.c05, s.c06, s.c07, s.c08, s.c09, s.c10, s.c11, s.c12, s.c13, s.c14, s.c15, s.c16, s.c17, s.c18, s.c19, s.c20, s.c21, s.c22, s.c23, s.c24, s.c25, s.c26, s.c27, s.c28, s.c29, s.c30, s.c31, s.c32, s.c33, s.c34, s.c35)
```
</details>

### This also occurs in production with **no knob**

We first hit this in a real ETL job — a native `MERGE INTO` over an attached Iceberg catalog on
**DuckDB 1.5.5**, with default optimizers (no `disabled_optimizers`, no other pragmas). The statement
had the shape:

```sql
MERGE INTO <target> t
USING ( /* a multi-CTE SELECT with window functions + GROUP BY aggregation that also re-scans <target> */ ) s
  ON t.<key> = s.<key>
WHEN MATCHED THEN UPDATE SET <every column> = <expr>
WHEN NOT MATCHED THEN INSERT (...) VALUES (...)
```

It failed with the same error, raised from the native-write path (`MERGE INTO` execution against the
ATTACHed catalog):

```
IO Error: The parquet file
  's3://<bucket>/<warehouse>/<schema>/<target>-<uuid>/data/<uuid>.parquet'
  seems to have incorrectly set page offsets. This interferes with DuckDB's prefetching optimization.
  DuckDB may still be able to scan this file by manually disabling the prefetching mechanism using:
  'SET disable_parquet_prefetching=true'.
```

The distinguishing factor versus the pruned/passing cases: in this failing query the **target
`ICEBERG_SCAN` kept a full (unpruned) projection** — the `unused_columns` optimizer did *not* prune it,
even though it pruned the sibling source scans in the same plan. A plain `SELECT *` of the same target
table succeeded. That unpruned target scan is exactly what `SET disabled_optimizers='unused_columns'`
forces in the minimal repro above — i.e. the minimal repro exercises the same reader code path that
fails here with no knob.

Captured `EXPLAIN` of the failing statement (default optimizers). Column names are redacted to
`col_NN` and timestamps to `<ts>`; the plan **structure is verbatim**. The MERGE target `ICEBERG_SCAN`
carries **no `Projections:`** (reads every column) while the source-side `ICEBERG_SCAN` in the same plan
*does* carry `Projections:` (pruned):

```
MERGE_INTO
└─ PROJECTION            WHEN MATCHED SET exprs, e.g. array_distinct(concat(col_19, col_19)),
│                        COALESCE(col_27, col_27), CASE WHEN (#68) THEN col_29 ELSE col_29 END  (36 cols)
   └─ PROJECTION         (36 cols)
      └─ HASH_JOIN   Join Type: RIGHT   Conditions: id = id                     ~2,465 rows
         ├─ ICEBERG_SCAN                                                        ~2,465 rows   ← MERGE TARGET
         │     Function: ICEBERG_SCAN
         │     «no Projections: line»            ⇒ target scan reads ALL columns (UNPRUNED)
         │
         └─ CTE  col_31   (Table Index: 1)                                      ~1,820 rows   ← source
               └─ FILTER  ((col_32 > '<ts>') OR (col_32 > '<ts>') AND NOT (col_29 IS NULL)) ...
                  └─ ICEBERG_SCAN                                               ~1,820 rows
                        Function: ICEBERG_SCAN
                        Projections: col_11, col_09, col_16, col_43, col_44, … ⇐ source scan IS pruned
```

In the full plan only 2 of the 6 `ICEBERG_SCAN` nodes carry `Projections:`, and the MERGE target is not
one of them. The `RIGHT` hash join places the target scan on the build side, which appears to defeat
projection push-down onto it — leaving the all-columns prefetch modify-scan that the reader miscounts.

### Additional findings

- **Not a read bug**: `SELECT *` of the exact data file succeeds (local & S3, prefetch on/off).
- **Not size/width**: reproduces at **2 columns, 2,000 rows, 1 row group**, and every larger size tried.
- **Modify-scan specific**: only `UPDATE`/`MERGE` (which scans the target for the write) trips it.
- **The suggested workaround is unsafe**: `SET disable_parquet_prefetching=true` stops the error but then
  *silently misreads bytes* (observed `Invalid string encoding ... not valid UTF8` on a forced read).
  It is data-corrupting, not a real workaround.

### OS:

Linux aarch64 (Debian bookworm `python:3.12-slim` / Ubuntu 24.04 containers; also hit on aarch64 Linux servers)

### DuckDB Version:

1.5.5 (also reproduces on 1.5.4; on current `main` the same accounting is present by inspection — see above)

### DuckDB Client:

Python (`duckdb==1.5.5`); the repro is plain SQL

### Hardware:

_No response_

### Full Name:

Michal Dumicz

### Affiliation:

Clutch Security

- [X] Yes, I have

- [X] Yes, I have

Yes

michal-clutch:
Fixes #24186 (the user-visible spurious `IOException`; the plan-side duplication it defends against is discussed below).

## Problem

`ParquetReader::RegisterRowGroupReads` sums `TotalCompressedSize()` once per `column_ids` **entry** and compares the total against the physical byte span of the row group; a total exceeding the span is treated as "incorrectly set page offsets" (remote files throw, local files fall back to synchronous reads).

Distinct column chunks occupy disjoint byte ranges inside the row-group span, so a duplicate-free `column_ids` list can never exceed it. But the planner can hand the scan a `column_ids` list that references the same physical column more than once: a `MERGE INTO` an ATTACHed Iceberg table whose source contains a materialized CTE over the target (explicit `MATERIALIZED`, or any CTE referenced twice → auto-materialized) produces `column_ids=[0,1,0,2]` — the ON-key file column appears twice (runtime-verified with a diagnostic build, details in #24186). The duplicated chunk is counted twice, the sum exceeds the span, and the scan fails with the misleading page-offsets error on a healthy file that DuckDB's own Iceberg writer produced moments earlier. Reproduces down to a 1-row, 2-column table with default optimizers (self-contained repro in the issue).

## Fix

Count each distinct column id once in the pre-check accounting. This also keeps `scan_percentage` (used to choose whole-group vs column-wise prefetch) meaningful for such plans.

The duplicated entries are benign for the actual read path — each `column_readers` entry reads its column independently — so only the accounting needs the guard. Verified at runtime on v1.5.5, where the same accounting lived in `ParquetReader::Scan`: a source build with the equivalent dedupe completes the repro `MERGE` with provably correct results (the repro merges the target into itself via `COALESCE`, so the table must come out unchanged — full-table hash checksums are identical before/after). On this branch, the patched translation unit compiles against current `main` headers and the file is byte-identical under the pinned `clang-format` 11.0.1. I could not re-run the Iceberg repro against a `main`/nightly build in our environment for an unrelated reason: the nightly Iceberg writer emits avro `codec: deflate`, which the REST catalog used in the repro (Lakekeeper) rejects at `CREATE TABLE`.

## Testing

I'd appreciate guidance here. The only reproduction we found requires the out-of-tree `iceberg` extension plus a remote (httpfs) file — `MERGE INTO` only targets catalog tables, so an in-tree sqllogictest cannot exercise this path directly. Options considered:

- land this defensive dedupe with #24186 as the evidence (this PR), and/or
- fix the duplication at its source — wherever `MERGE INTO` planning emits the duplicated ON-key binding for the target scan — which should be observable in-tree via `EXPLAIN`. Happy to attempt that as a follow-up if you can point me at the right place in the planner.

### DIFF
--- a/extension/parquet/parquet_reader.cpp
+++ b/extension/parquet/parquet_reader.cpp
@@ -2,6 +2,7 @@
 
 #include <string.h>
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 #include "duckdb/common/optional_ptr.hpp"
@@ -1879,10 +1880,16 @@ ParquetPrefetchStrategy ParquetReader::RegisterRowGroupReads(ClientContext &cont
 	state.group_offset = GetRowGroupOffset(*this, state.group_index);
 
 	uint64_t to_scan_compressed_bytes = 0;
+	unordered_set<idx_t> counted_column_ids;
 	for (idx_t i = 0; i < column_ids.size(); i++) {
 		auto col_idx = MultiFileLocalIndex(i);
 		PrepareRowGroupBuffer(context, state, col_idx);
-		to_scan_compressed_bytes += state.GetColumnReader(i).TotalCompressedSize();
+		// column_ids can reference the same physical column more than once (e.g. the target scan of a MERGE
+		// whose source materializes a CTE over the target) - count each column chunk only once so the sum
+		// stays comparable to the physical byte span of the row group below
+		if (counted_column_ids.insert(column_ids[col_idx].GetId()).second) {
+			to_scan_compressed_bytes += state.GetColumnReader(i).TotalCompressedSize();
+		}
 	}
 
 	auto &group = GetGroup(state);


michal-clutch:
@dentiny just checking if you have a chance to take a look?

PR #24193, opened 2026-07-27
Flush and fsync optimistically written blocks before the WAL lock at commit

ywelsch:
## Problem

A committing bulk append currently does all of its I/O while holding the WAL lock: `LocalStorage::Commit` flushes the remaining optimistically written blocks, the block manager is `FileSync`'ed (the `ROW_GROUP_DATA` entries in the WAL reference those blocks, so they must be durable before the WAL entry), and only then the actual WAL append happens.

On filesystems where fsync is expensive, e.g. network/remote storage, the WAL lock is then held for a substantial amount of time, although the WAL append itself takes milliseconds. Every other commit convoys behind it, including small commits to unrelated tables that would need the lock for microseconds.

## Approach

None of that I/O needs WAL ordering: the blocks are transaction-private until `MergeStorage`. `CommitTransaction` now calls `DuckTransaction::PreFlushOptimisticBlocks` **before** acquiring the transaction/WAL locks:

- It flushes the blocks of every table whose append takes the bulk-append path — at least one full row group and no deletes (`LocalTableStorage::IsBulkAppend`, which `LocalStorage::Flush` shares, so the decision cannot diverge). The decision depends only on transaction-local state, and the blocks are guaranteed to be used as written.
- It `FileSync`s the block manager if the commit references flushed row groups — whether flushed by the pre-flush itself or already during the statement (e.g. by batch inserts, which previously also paid their fsync under the WAL lock).
- Row-start assignment (`AppendLock`/`MergeStorage`) and the WAL append remain under the WAL lock, preserving per-table WAL order = row order.

For the bulk-append decision to be lock-free, `LocalStorage::Flush` no longer merges appends to an empty table directly for on-disk tables: appends below one row group now take the regular row-append path like any other small append. The direct merge for empty tables is kept only where the optimistic writer cannot write to disk (temporary / in-memory / read-only, e.g. WAL replay of a read-only attach), where there are no optimistically written blocks to manage.

With that, every commit that references optimistically written blocks from the WAL has already flushed and synced them before taking any lock, and the `FileSync` under the WAL lock is gone entirely. Two assertions pin the invariant: `LocalStorage::Flush` asserts that flushed blocks are only ever reverted by transactions that deleted rows after appending, and `WriteToWAL` asserts that WAL block references imply the pre-flush synced.

A commit that ends up skipping the WAL write in favor of an inline checkpoint pays one fsync it previously didn't; pure bulk appends rarely take that path (their WAL estimate is the KB-scale block-pointer size, not the data size).

**A pre-flush failure fails the commit**, exactly as a failure of the same flush under the WAL lock would. Retrying is not an option: the flush machinery cannot safely be re-run after an error (already-flushed partial blocks stay registered; a mid-`WriteToDisk` throw leaves dangling segment references), and a failed fsync must not be retried either — the retry can succeed without the data actually being durable.

The concurrency window is unchanged: optimistic writes already run mid-INSERT, long before commit, and `Truncate` only reclaims free-listed blocks.

## Measurements

8 concurrent single-statement commits of ~50MB appends to separate tables, with fsync latency simulated via an injected delay in `LocalFileSystem::FileSync` (db file / WAL file):

| fsync latency (db/wal) | before | after | |
|---|---|---|---|
| none injected (macOS fsync, ~0.1ms) | 1.24s | 0.82s | 1.5x |
| 10ms / 10ms   | 1.65s | 1.15s | 1.4x |
| 100ms / 100ms | 5.10s | 3.86s | 1.3x |
| 100ms / 10ms  | 2.68s | 1.55s | 1.7x |

The first row is a plain laptop with no simulated latency: even with sub-millisecond fsyncs the pre-flush wins 1.5x, because the commit-time flush of the trailing (incomplete) row group - its compression and block writes - now runs in parallel across the committers instead of serialized under the WAL lock. As fsync latency grows, the overlapped db-file fsyncs add to the gap; the whole gap scales with commit concurrency.

The convoy effect shows most clearly in a mixed workload with 8 connections each committing one ~50MB bulk append concurrently with 8 connections each running 50 small 1000-row commits:

| fsync latency (db/wal) | before | after | |
|---|---|---|---|
| 100ms / 0ms  | 2.50s | 1.26s | 2.0x |
| 100ms / 10ms | 8.00s | 6.49s | 1.2x |

In the first row the small commits need the WAL lock only for their (tiny) WAL appends, but spend most of their time queued behind the bulk commits' under-lock fsyncs; the pre-flush removes that wait entirely. The second row shows the remaining serialized cost: with ~850 commits each paying a 10ms WAL sync under the lock, the WAL sync itself dominates both sides — batching WAL syncs across commits (group commit) would be the natural follow-up for slow-fsync storage.

## Future work

- **Group commit**: the per-commit WAL sync stays serialized under the WAL lock (see above).
- The **checkpoint-instead-of-WAL path** still holds the WAL lock for the full inline checkpoint; unchanged here.
- Per-transaction **write-sequence tracking in the block manager** (`my last write seq > synced-up-to seq`) would let commits skip fsyncs already covered by someone else's sync. Wider interface change, left out of this PR.

### DIFF
--- a/src/storage/local_storage.cpp
+++ b/src/storage/local_storage.cpp
@@ -5,6 +5,7 @@
 #include "duckdb/planner/table_filter.hpp"
 #include "duckdb/storage/data_table.hpp"
 #include "duckdb/storage/partial_block_manager.hpp"
+#include "duckdb/storage/storage_manager.hpp"
 #include "duckdb/storage/table/append_state.hpp"
 #include "duckdb/storage/table/data_table_info.hpp"
 #include "duckdb/storage/table/row_group.hpp"
@@ -155,6 +156,22 @@ void LocalTableStorage::FlushBlocks() {
 		optimistic_writer.WriteUnflushedRowGroups(*row_groups);
 	}
 	optimistic_writer.FinalFlush();
+}
+
+bool LocalTableStorage::WritesToDisk() const {
+	return optimistic_writer.CanWriteToDisk();
+}
+
+bool LocalTableStorage::HasFlushedRowGroups() const {
+	return !row_groups->flushed_row_groups.empty();
+}
+
+bool LocalTableStorage::IsBulkAppend() const {
+	if (is_dropped || deleted_rows != 0) {
+		return false;
+	}
+	auto &collection = *row_groups->collection;
+	return collection.GetTotalRows() >= collection.GetRowGroupSize();
 }
 
 ErrorData LocalTableStorage::AppendToIndexes(DuckTransaction &transaction, RowGroupCollection &source,
@@ -325,6 +342,16 @@ shared_ptr<LocalTableStorage> LocalTableManager::MoveEntry(DataTable &table) {
 reference_map_t<DataTable, shared_ptr<LocalTableStorage>> LocalTableManager::MoveEntries() {
 	lock_guard<mutex> l(table_storage_lock);
 	return std::move(table_storage);
+}
+
+vector<shared_ptr<LocalTableStorage>> LocalTableManager::GetEntries() const {
+	lock_guard<mutex> l(table_storage_lock);
+	vector<shared_ptr<LocalTableStorage>> result;
+	result.reserve(table_storage.size());
+	for (auto &entry : table_storage) {
+		result.push_back(entry.second);
+	}
+	return result;
 }
 
 idx_t LocalTableManager::EstimatedSize() const {
@@ -577,13 +604,16 @@ void LocalStorage::Flush(DataTable &table, LocalTableStorage &storage, optional_
 	}
 
 	auto append_count = storage.GetCollection().GetTotalRows() - storage.deleted_rows;
-	const auto row_group_size = storage.GetCollection().GetRowGroupSize();
 
 	TableAppendState append_state;
 	table.AppendLock(transaction, append_state);
-	if ((append_state.row_start == 0 || storage.GetCollection().GetTotalRows() >= row_group_size) &&
-	    storage.deleted_rows == 0) {
-		// table is currently empty OR we are bulk appending: move over the storage directly
+	if (storage.IsBulkAppend() ||
+	    (append_state.row_start == 0 && storage.deleted_rows == 0 && !storage.WritesToDisk())) {
+		// bulk append (at least one full row group, no deletes): move over the storage directly.
+		// Appends to an empty table are also merged directly if the table cannot be written to
+		// disk (temporary / in-memory / read-only, e.g. WAL replay of a read-only attach) -
+		// there are no optimistically written blocks to manage, and merging avoids re-appending
+		// row by row.
 		// first flush any outstanding blocks
 		storage.FlushBlocks();
 		// Append to the indexes.
@@ -594,6 +624,9 @@ void LocalStorage::Flush(DataTable &table, LocalTableStorage &storage, optional_
 		// check if we have written data
 		// if we have, we cannot merge to disk after all
 		// so we need to revert the data we have already written
+		// this only happens for transactions that deleted rows after bulk-appending: a pure bulk
+		// append always takes the merge path above, using its pre-flushed blocks as written
+		D_ASSERT(!storage.HasFlushedRowGroups() || storage.deleted_rows > 0);
 		storage.Rollback();
 		// append to the indexes
 		storage.AppendToIndexes(transaction, append_state);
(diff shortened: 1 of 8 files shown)


ywelsch:
Test failure here is unrelated to this PR, and being fixed in https://github.com/duckdb/duckdb/pull/24190

src/transaction/duck_transaction.cpp
```diff
@@ -206,6 +206,38 @@ bool DuckTransaction::ShouldWriteToWAL(AttachedDatabase &db) {
...
+		auto result = storage->PreFlushBlocks();
+		if (!result.wrote_blocks) {
+			return error;
+		}
+		// the flushed blocks are referenced from the WAL and need a FileSync - but only sync here
+		// if this is provably the commit's final sync (so we never sync more often than before):
+		// otherwise defer to the under-lock FileSync (more blocks may be written under the commit
+		// locks) or to the checkpoint (the commit is predicted to skip the WAL write; a wrong
+		// prediction falls back to the under-lock FileSync as before)
+		auto undo_properties = GetUndoProperties();
```
Mytherin: This does a full scan of the UndoBuffer - we shouldn't do that in this function just to discard it again

src/storage/local_storage.cpp
```diff
@@ -147,14 +147,22 @@ void LocalTableStorage::WriteNewRowGroup(idx_t flushed_row_group_idx) {
...
+	// FinalFlush reports whether a partial block manager existed, i.e. whether blocks may have been written
+	return optimistic_writer.FinalFlush();
+}
+
+bool LocalTableStorage::IsUnconditionalBulkAppend() const {
+	if (is_dropped || deleted_rows != 0) {
+		return false;
+	}
+	auto &collection = *row_groups->collection;
+	return collection.GetTotalRows() >= collection.GetRowGroupSize();
```
Mytherin, line 175: We always call `storage.FlushBlocks()` in the `LocalStorage::Flush()`. If we want to correctly check here if we're done writing we need to check the row groups / optimistic writer if everything has been fully flushed. Just checking the row count is likely not always correct (even though it is often correct).

Mytherin (commented):
Thanks for the PR!

In general this sounds like a good idea but this approach is introducing a large amount of complexity, and only occasionally resolves the stated issue (if there's an insert that *might* trigger additional flushing this optimization does not trigger).

It would be nicer if we could always do the FileSync / flushing to disk prior to obtaining the transaction lock instead of the approximate approach used here. I'm not sure if the logic in `LocalStorage::Flush` makes that much sense either, e.g. this:

```cpp
(append_state.row_start == 0 || storage.GetCollection().GetTotalRows() >= row_group_size)
```

This means we're always doing optimistic writes if it's the first append to a table - but that doesn't really make sense to me. It makes sense for an *append to an in-memory table* to directly merge the row groups onto the table which is where I think this logic comes from, but it doesn't really make sense for optimistic writes to an on-disk table. So it might make sense to revisit that in general.

There's also a separate issue that's not directly related to this optimization, but will interfere with the code changes here when implemented - which is that currently inserting is separate from the regular undo buffer flow which causes a number of issues with indexes / foreign keys / etc. This will need to be refactored at some point in the (near) future as well and will likely interfere with this optimization / require a refactor of this optimization.

src/transaction/duck_transaction.cpp
ywelsch: This problem should be gone now as I've made the pre-flush unconditional. The tradeoff is that a commit that ends up skipping the WAL write in favor of an inline checkpoint pays one early fsync it previously didn't. Pure bulk appends rarely take that path, since their WAL size estimate is the KB-scale block-pointer estimate rather than the data size.

src/storage/local_storage.cpp
ywelsch: Agreed (and reworked). The row count no longer decides whether writing is finished. It only decides whether Flush() will take the bulk-append path. Whether a sync is needed now comes from the writer state itself: after FlushBlocks(), the pre-flush syncs iff the storage holds flushed row groups (HasFlushedRowGroups), which also covers row groups already flushed during the statement (e.g. batch inserts). Flush() still unconditionally calls FlushBlocks(); at commit time that's now a cheap no-op. WriteUnflushedRowGroups early-outs when all row groups are flushed and no partial block managers remain.

ywelsch:
Thanks for the review. I've reworked the PR along the lines you suggested, where the flush/FileSync now always happens before the locks. The complexity should be reduced with these changes too.

Reducing WAL lock contention is a critical thing for us (observed for concurrent write workloads). Next to this PR here and the group commit work the only other thing currently on my radar are inline checkpoints. I'm thinking of disabling these on our service.

Mytherin (approved):
Thanks for the updates - looks a lot better now. We can merge this one once CI passes.

ywelsch:
The [failure](https://github.com/duckdb/duckdb/actions/runs/30445892963/job/90558674907?pr=24193) (tpch benchmark regression on q07) is deterministic and real (reproduced locally), but it's a statistics-accuracy side effect, not a defect of the code changes here. The key change that trigged it was the "small append to an empty on-disk table" case the rework moved from the direct-merge path to the row-append path (The supplier table has 10k rows < 122880). The row-append path rebuilds statistics while re-appending the full data into the target table, which for a unique column clamps to the exact count (supplier.s_suppkey approx-distinct becoming exact (10000 vs 8565)). 

The more accurate distinct count makes DuckDB's join-order optimizer pick a q07 plan whose measured build+probe cardinalities are ~16% higher, changing the join order.

This is not something that the PR here can address, I think.

ywelsch:
@Mytherin can you advise on the above?

Mytherin:
We can ignore the regression failure - thanks

PR #24196, opened 2026-07-27
fix(adbc): return a dense union from ConnectionGetInfo

eitsupi:
The ADBC spec requires `ConnectionGetInfo` to return a dense union, but the old SQL-based implementation could only produce a sparse union.
This builds the array manually instead, reusing the helpers from #23230.

Since arrow_scan cannot consume dense unions either, the C++ tests inspect the result buffers directly, and I additionally verified the built driver end-to-end with the Python ADBC driver manager + PyArrow.

It also stops rejecting a null `error`, which is an optional out parameter.

cc @lidavidm


pdet:
Thanks, looks great!

PR #24198, opened 2026-07-27
Fix NightlyTests

smvv:
- Fix two test [failures](https://github.com/duckdb/duckdb/actions/runs/30182087655/job/89740331661) under `FORCE_ASYNC_SINK_SOURCE=1`.
- Fix Hash Zero test [failures](https://github.com/duckdb/duckdb/actions/runs/30182087655/job/89741976111) by skipping slow (timing out after 120s) tests.

### DIFF
--- a/test/api/test_read_ahead_abandon.cpp
+++ b/test/api/test_read_ahead_abandon.cpp
@@ -1,10 +1,16 @@
 #include "catch.hpp"
 #include "test_helpers.hpp"
 
+#include <cstdlib>
+
 using namespace duckdb;
 using namespace std;
 
 TEST_CASE("Test abandoning a streaming result mid-scan with read-ahead", "[api]") {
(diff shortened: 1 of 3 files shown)


src/execution/operator/schema/physical_create_index.cpp
taniabogatsch, line 1: IIRC fixing this correctly here is more involved, there was some previous work in https://github.com/duckdb/duckdb/pull/23259, and we're aiming to [fix the linked issue for `v2.0`](https://github.com/duckdblabs/duckdb-internal/issues/9642). 

Maybe it would make more sense to skip the test for now?

smvv: @taniabogatsch ok i'll remove the code changes, thanks!
(I think it's fine to wait a few weeks to resolve that test failure.)

PR #24199, opened 2026-07-27
Fix NULL handling in ASOF join prefix comparisons

### LINKED ISSUE #23933: ASOF JOIN ignores a valid NULL-safe equality match
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE trades(id INTEGER, symbol VARCHAR, ts INTEGER);
memory D CREATE TABLE prices(id INTEGER, symbol VARCHAR, ts INTEGER);
memory D INSERT INTO trades VALUES (1, NULL, 5);
memory D INSERT INTO prices VALUES (2, NULL, 3);

-- should be 1   NULL     5   2  NULL     3
memory D SELECT *
         FROM trades ASOF LEFT JOIN prices
           ON trades.symbol IS NOT DISTINCT FROM prices.symbol
          AND trades.ts >= prices.ts;
┌───────┬─────────┬───────┬───────┬─────────┬───────┐
│  id   │ symbol  │  ts   │  id   │ symbol  │  ts   │
│ int32 │ varchar │ int32 │ int32 │ varchar │ int32 │
├───────┼─────────┼───────┼───────┼─────────┼───────┤
│     1 │ NULL    │     5 │  NULL │ NULL    │  NULL │
└───────┴─────────┴───────┴───────┴─────────┴───────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

williampsouza:
This issue results from a combination of two problems.

First, while converting the ASOF conditions, they are placed into separate collections: `lhs_orders` and `lhs_partitions`. However, `lhs_orders.size()` is then used to determine the condition's position in the sort key, causing NULL sensitivity to be assigned to the wrong column.

Then, when processing a NULL sort key during prefix comparison, the code assumes the wrong width and compares bytes belonging to the next column. The solution is to detect NULL values and limit the comparison to the appropriate width.

Fixes #23933


test/sql/join/asof/test_asof_join.test
```diff
@@ -88,6 +88,75 @@ ORDER BY ALL
 
 endloop
 
+# Force the physical ASOF operator for NULL-sensitive condition tests
+statement ok
+PRAGMA asof_loop_join_threshold = 0;
```
hawkfish: These tests should probably be in a different file in case someone adds tests later and doesn't notice this.

src/execution/operator/join/physical_asof_join.cpp
```diff
@@ -751,14 +752,25 @@ struct SortKeyPrefixComparison {
 			auto lhs_width = col.size;
 			auto rhs_width = col.size;
 			int cmp = 1;
+			const auto has_null =
```
hawkfish, line 734: Why do we need to keep modifying the `null_type` field here? The partitions are always sorted `NULLS FIRST`.

hawkfish (changes requested):
Thanks for the PR! the main fix looks good - just some minor nits.

src/execution/operator/join/physical_asof_join.cpp
williampsouza: I added the helper to avoid duplicating the internal sort-key marker values outside `create_sort_key` and added the logic for `null_type` comparison so the helper wouldn't be misused. Would you think that for this case we could remove the `null_order` parameter and leave the comparison with `SortKeyVectorData::NULL_FIRST_BYTE`?

hawkfish: Sorry I was really jet lagged. Now that I can actually see my screen, this looks fine.

hawkfish (approved):

src/execution/operator/join/physical_asof_join.cpp
williampsouza: No worries! I really appreciate you taking the time to review this.

williampsouza:
@Mytherin could you rerun the failed ingestion perf CI job? The regression appears unrelated to this PR

PR #24200, opened 2026-07-27
Add `capitalize_identifier` setting for uppercase identifier folding (v1.5)

dataders:
Original patch and authorship: @MartinNeupauer.

This adds a `capitalize_identifier` setting (meaningful only when `preserve_identifier_case=false`) that folds unquoted identifiers to **uppercase** instead of lowercase — matching the default identifier-folding behavior of Snowflake, Oracle, and DB2.

Closes: https://github.com/duckdb/duckdb/discussions/24197

There's two ways to approach solving this:
1. add a new param `capitalize_identifier` that only works when `preserve_identifier_case=false` (what this PR does)
2. modify  `preserve_identifier_case` to have three states: `on`, `lowercase`, `uppercase` (potential backcompat issues; could theoretically land in 2.0)

Targeting `v1.5-variegata` rather than `main`, since this patch was originally written against the pre-PEG-parser `third_party/libpg_query`-based identifier scanner. `main` has since moved to the in-house PEG parser (`src/parser/peg/matcher.cpp`), so the equivalent change there would need to be reimplemented against that parser rather than cherry-picked — that port isn't done yet.

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -194,6 +194,13 @@
         ]
     },
     {
+        "name": "capitalize_identifier",
+        "description": "When preserve_identifier_case is false, uppercase non-quoted identifiers instead of lowercasing them (no effect when preserve_identifier_case is true)",
+        "type": "BOOLEAN",
+        "default_scope": "local",
+        "default_value": "false"
+    },
+    {
         "name": "catalog_error_max_schemas",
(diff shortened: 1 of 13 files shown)


dataders:
closing in favor of #24375 because this is a "new" feature that is not backwards compatible. there is no `1.6.0` release, only `2.0.0`

dataders:
closing in favor of #24375

PR #24202, opened 2026-07-27
Connect + detach race and add missing InvokeCloseIfLastReference's 

artjomPlaunov:
Fix a race between connect and detach by adding an is_closing bool to attached_db, set only under the close_lock in InvokeCloseIfLastReference after observing the last reference. 

There is also the possibility in ClientContext::PendingStatementOrPreparedStatement of having the last reference to the attached_db, and thus invoking a destructor rather than explicit close, so we add an invoke there. 

Note for reviewer: I made this refactor using the existing TryGetConnectedCatalog(), which made me realize there are more places we should probably add InvokeCloseIfLastReference?


Mytherin:
Thanks!

PR #24203, opened 2026-07-27
Fix indexed DELETE column resolution

### LINKED ISSUE #24102: Delete row with index crashes
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t(id INT PRIMARY KEY, payload VARCHAR);
memory D INSERT INTO t SELECT range, range::VARCHAR FROM range(10);
memory D DELETE FROM t USING range(100) d(i) WHERE d.i = 0;
INTERNAL Error:
Vector::Reference used on vector of different type (source INTEGER referenced BIGINT)
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Andrewtangtang:
## Summary

Resolves #24102

This PR fixes a crash in indexed `DELETE ... USING` queries when optimization changes the child output layout. The
change tracks target columns through optimization and resolves their final positions before execution.

## Root Cause

`LogicalDelete::return_columns` stored pre-optimization `DataChunk` positions for indexed and `RETURNING` columns.
Join reordering could change the final output order, causing `PhysicalDelete` to reference a vector from the `USING` side with an incompatible type.

## Solution

For internal DuckDB tables, keep the required columns as bound expressions through optimization and resolve their final chunk positions during physical planning.

### DIFF
--- a/src/execution/physical_plan/plan_delete.cpp
+++ b/src/execution/physical_plan/plan_delete.cpp
@@ -1,4 +1,5 @@
 #include "duckdb/catalog/catalog_entry/duck_table_entry.hpp"
+#include "duckdb/common/constants.hpp"
 #include "duckdb/execution/operator/persistent/physical_delete.hpp"
 #include "duckdb/execution/physical_plan_generator.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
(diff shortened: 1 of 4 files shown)


Andrewtangtang:
The CI failure appears unrelated to this PR.

Andrewtangtang:
Hi @Mytherin, could you help look at this PR when you have time? Thanks!

Andrewtangtang:
Hello @lnkuiper, could also you help look at this PR when you have time? Thanks!

lnkuiper:
Sorry for the delay, we have less reviewing capacity during the summer months. Looks good to me, thanks!

Andrewtangtang:
No problem at all! Thanks so much for reviewing!

PR #24218, opened 2026-07-28
check page size mismatch for uncompressed data page v2

jmestwa-coder:
PreparePageV2 sizes the page block from uncompressed_page_size but reads compressed_page_size bytes into it:
- a v2 page counts as uncompressed when the chunk codec is UNCOMPRESSED or when the page header sets is_compressed=false, and the size equality check only guarded the first case
- a snappy column with is_compressed=false, uncompressed_page_size 64 and compressed_page_size 803 then writes 803 file-controlled bytes into a 65 byte block
- ASan reports a heap-buffer-overflow WRITE of size 803 on the attached file before the change, and a clean Invalid Input Error after

Valid v2 pages that set is_compressed=false are unaffected, their two sizes agree.


jmestwa-coder:
any update?

PR #24219, opened 2026-07-28
CSV: remove unreachable type check

wufengwind:
AddPossiblyEscapedValue checks whether the target parse vector is VARCHAR before choosing between cast-error handling and escape removal. The VARCHAR branch repeats the same non-VARCHAR condition, so its cast-error block can never execute.

Remove the unreachable duplicate check while leaving the existing non-VARCHAR error path and VARCHAR escape-removal path unchanged.

Tests:
- DUCKDB_EXTENSIONS=json make reldebug -j4
- build/reldebug/test/unittest test/sql/copy/csv/afl/test_fuzz_4793.test --use-colour no
- build/reldebug/test/unittest test/sql/copy/csv/unquoted_escape/basic.test --use-colour no
- build/reldebug/test/unittest test/sql/copy/csv/test_escape_long_value.test --use-colour no
- build/reldebug/test/unittest test/sql/copy/csv/test_quoted_later_escaped.test --use-colour no

### DIFF
--- a/src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
+++ b/src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
@@ -626,22 +626,6 @@ void StringValueResult::AddPossiblyEscapedValue(StringValueResult &result, const
 				result.cur_col_id++;
 				result.chunk_col_id++;
 			} else {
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks!

PR #24221, opened 2026-07-28
Fix nested schema lookup

### LINKED ISSUE #24220: Failed to create view with nested schema
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE SCHEMA s1;
memory D CREATE SCHEMA s1.s2;
memory D CREATE VIEW s1.s2.v AS SELECT 42 AS x;
Catalog Error:
Schema with name s2 does not exist!
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24220

Current implementation for `Schema()` only [returns the leaf schema](https://github.com/duckdb/duckdb/blob/e0692f2bf6fe5c6f68e376d8c3e244380a4152ec/src/include/duckdb/parser/qualified_name.hpp#L60-L61), so all prev nested schemas are ignored.
In this PR, I explicitly extract all schemas out.

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -162,11 +162,21 @@ optional_ptr<CatalogEntry> Catalog::CreateTable(CatalogTransaction transaction, 
 	return CreateTable(transaction, *schema, info);
 }
 
+static SchemaCatalogEntry &GetCreateSchema(Catalog &catalog, CatalogTransaction transaction, CreateInfo &info) {
+	auto &qname = info.GetQualifiedName();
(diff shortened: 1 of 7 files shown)


dentiny:
Ah I think Mark's actively working on nested schema support, don't think it needed.

PR #24223, opened 2026-07-28
Make sure dictionary vectors don't contain invalid values

hannes:
We had this case where we read a dictionary page from a parquet file with a filter on a different column. the selection vector lead to particular values not being filled, which lead to crashes/address sanitizer issues when e.g. copying this vector around. 

This PR fixes this by filling those slots with 0 or the empty string, and adds a verification call to ensure this does not happen elsewhere.

### DIFF
--- a/extension/parquet/include/column_reader.hpp
+++ b/extension/parquet/include/column_reader.hpp
@@ -306,6 +306,13 @@ private:
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


hannes:
This apparently also occurred in variants and in the CSV reader ^^

PR #24224, opened 2026-07-28
Fix bench runs

smvv:
I made a typo when [splitting](https://github.com/duckdb/duckdb/pull/24184) the benchmarks:
```
          - name: Micro
            benchmarks: .github/regression/tpch_parquet.csv
          - name: TPCH-PARQUET
            benchmarks: .github/regression/tpch_parquet.csv
```
so the `Micro` benchmark was actually running `tpch_parquet.csv` again.

### Changes 
- Move TPC-H parquet to namespace.
- Re-enable micro bench in CI.
- Add timeout to install command and CI step.


Mytherin:
Thanks!

PR #24226, opened 2026-07-28
[MultiFileReader] Add column indexes to `DynamicFilterPushdown` parameters

Tishj:
Add `MultiFileDynamicPushdownInfo` to make the method easier to extend in the future.
Added `column_indexes`, but kept `column_ids` for backwards compatibility.

Reason for the change is that IcebergMultiFileList expected a 1-to-1 mapping from column_id -> filter, which is violated when we pushdown variant extracts.

This makes it possible to upgrade that map from `column_t` to `ColumnIndex`, which restores the constraint again.


Mytherin:
Thanks!

PR #24228, opened 2026-07-28
fix: incorrect join order of joins without ON clause

### LINKED ISSUE #24045: Unexpected multiple join involving subquery containing cast
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE a(x INT);
memory D CREATE TABLE b(x INT);
memory D CREATE TABLE c(x INT);
memory D CREATE TABLE d(y INT);
memory D 
memory D INSERT INTO a VALUES (1);
memory D INSERT INTO c VALUES (1);
memory D INSERT INTO d VALUES (9);
memory D 
-- should be 1,9
memory D SELECT c.x, d.y
         FROM a NATURAL LEFT JOIN b
         NATURAL JOIN c
         RIGHT JOIN d ON TRUE;
┌───────┬───────┐
│   x   │   y   │
│ int32 │ int32 │
├───────┼───────┤
│  NULL │     9 │
└───────┴───────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

jokasimr:
Fixes #24045

The problem was that the grammar rule for joins without ON clauses created a right-associated join chain. That was then partially undone by custom transformation logic, but in some cases this transformation did not work correctly.

The fix restricts joins without ON clauses to their immediate `InnerTableRef`. Any following joins remain in the chain to be parsed, creating the expected left-associated join chain.

### DIFF
--- a/src/include/duckdb/parser/peg/inlined_grammar.gram
+++ b/src/include/duckdb/parser/peg/inlined_grammar.gram
@@ -1733,7 +1733,7 @@ JoinClause <- JoinByClause / RegularJoinClause / JoinWithoutOnClause
 RegularJoinClause <- Asof? JoinType? 'JOIN' TableRef JoinQualifier
 JoinByClause <- 'JOIN' 'BY' Parens('TYPE' ColLabel) TableRef JoinQualifier
 Asof <- 'ASOF'
-JoinWithoutOnClause <- JoinPrefix 'JOIN' TableRef
+JoinWithoutOnClause <- JoinPrefix 'JOIN' InnerTableRef
 JoinQualifier <- OnClause / UsingClause
 OnClause <- 'ON' Expression
 UsingClause <- 'USING' Parens(List(ColumnName))
--- a/src/include/duckdb/parser/peg/inlined_grammar.hpp
+++ b/src/include/duckdb/parser/peg/inlined_grammar.hpp
@@ -1578,7 +1578,7 @@ const char INLINED_PEG_GRAMMAR[] = {
 	"RegularJoinClause <- Asof? JoinType? 'JOIN' TableRef JoinQualifier\n"
 	"JoinByClause <- 'JOIN' 'BY' Parens('TYPE' ColLabel) TableRef JoinQualifier\n"
 	"Asof <- 'ASOF'\n"
-	"JoinWithoutOnClause <- JoinPrefix 'JOIN' TableRef\n"
+	"JoinWithoutOnClause <- JoinPrefix 'JOIN' InnerTableRef\n"
 	"JoinQualifier <- OnClause / UsingClause\n"
 	"OnClause <- 'ON' Expression\n"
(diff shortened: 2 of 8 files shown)


lnkuiper:
Thanks for finding and fixing this! Looks like the fix is in the correct direction, just one edge case that needs covering:
```sql
# name: test/temp.test
# description: Conditionless joins preserve postfix PIVOT scope
# group: [join]

query III rowsort
SELECT *
FROM (VALUES (1), (1)) a(i)
CROSS JOIN (VALUES ('x', 10), ('y', 20)) b(k, v)
PIVOT(sum(v) FOR k IN ('x', 'y'))
----
1	10	20
1	10	20
```

jokasimr:
Thanks for taking a look!

> Thanks for finding and fixing this! Looks like the fix is in the correct direction, just one edge case that needs covering:
> 
> ```sql
> # name: test/temp.test
> # description: Conditionless joins preserve postfix PIVOT scope
> # group: [join]
> 
> query III rowsort
> SELECT *
> FROM (VALUES (1), (1)) a(i)
> CROSS JOIN (VALUES ('x', 10), ('y', 20)) b(k, v)
> PIVOT(sum(v) FOR k IN ('x', 'y'))
> ----
> 1	10	20
> 1	10	20
> ```

Shouldn't the result of that query be
```
i    x    y
----
1    20    40
```
?

That is the result produced by duckdb 1.5.5.

lnkuiper:
1.5.5 indeed produces that result, which is incorrect. The base branch that you're PRing against produces the correct result. Your changes regress to the incorrect v1.5.5 behaviour

Dtenwolde (approved):

jokasimr:
> 1.5.5 indeed produces that result, which is incorrect. The base branch that you're PRing against produces the correct result. Your changes regress to the incorrect v1.5.5 behaviour

Sorry I don't fully understand what the intended behavior is.

Does the PIVOT apply to the last referenced table in the join statement, instead of to the entire FROM ... JOIN ... statement?

For example, what should this query return?
```sql
SELECT *
FROM (VALUES (1), (1)) a(i)
JOIN (VALUES ('x', 10), ('y', 20)) b(k, v) ON true
PIVOT(sum(v) FOR k IN ('x', 'y'));
```
On the main branch it returns
```
1    20    40
```
but I would expect it to return the same as the CROSS JOIN version above.

I did a bit of investigation into what other engines do with the PIVOT syntax in this case:
Oracle returns a single row (`1 20 40`) for the CROSS JOIN query and the JOIN ON TRUE version.
BigQuery returns two rows (`1 20 20, 1 20 20`) for the CROSS JOIN query, but it fails to parse the JOIN ON TRUE version.

lnkuiper:
You're right, I'm not sure what happened on the base branch in the meantime. The results are correct. Sorry for the confusion! LGTM

PR #24234, opened 2026-07-28
[MultiFileReader] Add optional `extra_columns` to `MultiFileReaderData`

Tishj:
Currently we have `extra_columns` inside `MultiFileReaderGlobalState`, which decides a set of extra columns that can appear inside the `input_chunk` for all files.

With this PR, we can add extra columns on a more granular basis.
To enable this, the `MultiFileReaderGlobalState` needs to be created and `supports_local_extra_columns` needs to be set.

### DuckLake

My fix exposed a shoddy "extra columns" alternative in DuckLake, I've pushed a patch that fixes it to use `extra_columns` in the right way.
I've also pushed this patch as a PR to DuckLake.

EDIT: That PR to DuckLake is now merged, so the patch doesn't need to be reviewed in as much detail anymore


src/include/duckdb/common/multi_file/multi_file_function.hpp
```diff
@@ -677,15 +712,10 @@ class MultiFileFunction : public TableFunction {
 					}
 					result->scanned_types.emplace_back(entry->second.type);
 				} else {
-					result->scanned_types.push_back(table_types[column_id]);
+					result->scanned_types.push_back(col_idx.HasType() ? col_idx.GetScanType() : table_types[column_id]);
 				}
 			}
 		}
-		if (require_extra_columns) {
-			for (const auto &column_type : result->multi_file_reader_state->extra_columns) {
```
Tishj, line 685: This might be too heavily iceberg-influenced, since we add to the set of "global columns" for our extra columns, to leverage `CreateMapping` for creating expressions+(local)column_indexes for our extra columns.

EDIT: Actually, that's also what DuckLake does

PR #24236, opened 2026-07-28
Make `AtClause::Unit()` return an `Identifier`

JelteF:
The AT clause unit (`VERSION`/`TIMESTAMP`) is case-insensitive, so every
consumer had to spell that out by hand. `Identifier` encodes exactly that
invariant, so consumers can now compare with `==` instead:

https://github.com/duckdb/ducklake/blob/d318a545571d7d46eb751fa2aa5f6f4389285d3c/src/storage/ducklake_metadata_manager.cpp#L4109

`Identifier` serializes as its raw string, so the serialization format is
unchanged.


JelteF:
This one is now actually ready for review

Mytherin:
Thanks!

Issue #24207, opened 2026-07-28
Filter Pushdown Volatile Group/HAVING Wrong Result


dentiny:
### What happens?

When we disable and enable optimizers, we get different results.

### To Reproduce

```sql
memory D CREATE SEQUENCE s START 1;
memory D SELECT nextval('s') AS g, count(*) AS c
         FROM range(3)
         GROUP BY 1
         HAVING g > 1
         ORDER BY g;
┌───────┬───────┐
│   g   │   c   │
│ int64 │ int64 │
├───────┼───────┤
│     4 │     1 │
│     5 │     1 │
└───────┴───────┘
memory D SELECT currval('s');
┌──────────────┐
│ currval('s') │
│    int64     │
├──────────────┤
│            5 │
└──────────────┘
```
after we disable optimizers, we get
```sql
Enter ".help" for usage hints.
memory D PRAGMA disabled_optimizers='filter_pushdown';
memory D CREATE SEQUENCE s START 1;
memory D SELECT nextval('s') AS g, count(*) AS c
         FROM range(3)
         GROUP BY 1
         HAVING g > 1
         ORDER BY g;
┌───────┬───────┐
│   g   │   c   │
│ int64 │ int64 │
├───────┼───────┤
│     2 │     1 │
│     3 │     1 │
└───────┴───────┘
memory D SELECT currval('s');
┌──────────────┐
│ currval('s') │
│    int64     │
├──────────────┤
│            3 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

531911608719cd03b723051b8cb3b3e81cf3e03b

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

prince3235:
Hy @dentiny,
I'd like to investigate this issue.

I'll first reproduce the behavior, compare the logical/physical plans with and without `filter_pushdown`, and trace where the optimizer changes the semantics for the volatile expression in the `HAVING` clause.

If I identify the root cause, I'd be happy to work on a fix and submit a PR.

PR #24239, opened 2026-07-28
Don't push filters on volatile groups through aggregates

DinosL:
Fixes: https://github.com/duckdb/duckdb/issues/24207

Filters on volatile group expressions like `nextval()` were pushed below the aggregate, causing the expression to be re-evaluated and producing wrong results. We now skip pushdown for those.


Mytherin:
Thanks!

Issue #24206, opened 2026-07-28
LIMIT Pushdown Volatile Wrong Result


dentiny:
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE SEQUENCE s START 1;

-- both queries should return 2
memory D SELECT nextval('s') AS v FROM range(3) LIMIT 1 OFFSET 1;
┌───────┐
│   v   │
│ int64 │
├───────┤
│     1 │
└───────┘
memory D SELECT currval('s');
┌──────────────┐
│ currval('s') │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
```
Check postgres result
```sql
CREATE SEQUENCE s START 1;

SELECT nextval('s') AS v
FROM generate_series(1, 3)
LIMIT 1 OFFSET 1;
v 
---
 2
(1 row)

SELECT currval('s');
 currval 
---------
       2
(1 row)
```

### OS:

macos

### DuckDB Version:

531911608719cd03b723051b8cb3b3e81cf3e03b

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

PR #24240, opened 2026-07-28
Fix LIMIT pushdown through a volatile projection with an OFFSET

DinosL:
Fixes: https://github.com/duckdb/duckdb/issues/24206

`LimitPushdown` moves a constant `LIMIT` below its child `PROJECTION`. When the limit has an offset, this discards the offset rows before the projection runs, so a volatile expression produces different values for the rows we keep.


Mytherin:
Thanks!

dentiny:
Weird the issue was not automatically closed hmmm

PR #24248, opened 2026-07-28
Only define EXT_VERSION_<NAME> for the sources that read it

JelteF:
DuckDB defines `EXT_VERSION_<NAME>` for an extension's entire directory. For
in-tree extensions that value defaults to DuckDB's own commit hash. So every
commit changes the compiler command line of every source file of every
extension, and of the third party code they vendor. The only thing reading the
macro is `<name>_extension.cpp`, to implement `Version()`.

An extension can now declare which of its sources read it:

    set_extension_version_sources(parquet parquet_extension.cpp)

That drops the directory-wide definition inherited from DuckDB and applies it to
the listed sources instead. It has to be called before the extension declares
any targets or adds any subdirectories, since those inherit the directory-wide
definition at the point they are created.

For this build:

    DUCKDB_LINKER=lld make release BUILD_EXTENSIONS="json;icu;tpch;tpcds;httpfs"

a new commit now invalidates 12 objects instead of 428. With ccache warm that is
a 10 second rebuild instead of 15. Those 416 translation units were ccache hits
rather than real compiles, but still had to be preprocessed. It is also much
clearer which files actually need recompiling.

Out-of-tree extensions are unaffected unless they opt in. Their version comes
from their own repository, so it does not change when DuckDB does.

### DIFF
--- a/extension/README.md
+++ b/extension/README.md
@@ -148,6 +148,36 @@ results in:
 ...
 ```
 
+# Reporting an extension version
+
+An extension can report its version through the `EXT_VERSION_<NAME_UPPERCASE>` macro, which DuckDB defines to
+the extension version configured at build time:
+```cpp
+std::string ParquetExtension::Version() const {
+#ifdef EXT_VERSION_PARQUET
+	return EXT_VERSION_PARQUET;
+#else
+	return "";
+#endif
+}
+```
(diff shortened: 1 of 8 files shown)


Mytherin:
Thanks!

PR #24249, opened 2026-07-29
Fix missing catalog dependency registration for index_key

### LINKED ISSUE #24191: Bug3: Use-After-Free in index_key() via Stale ART Reference in Prepared Statement Cache
### What happens?

`IndexKeyBind()` in `extension/core_functions/scalar/debug/index_key.cpp` resolves an index name to its `BoundIndex`/`ART` object and stores a raw C++ reference inside `IndexKeyBindData`, which is cached for the lifetime of a `PREPARE`d statement. Unlike every other bind time catalog dependency in DuckDB, it never calls `RegisterDBRead`/`RegisterDBModify`, so `PreparedStatementData::RequireRebind()` never detects that the referenced index was dropped. `DROP INDEX` frees the underlying `ART` object at commit time, and a later `EXECUTE` of the same prepared statement dereferences the now dangling `ART&`, a genuine heap use after free with attacker timed reuse potential. Wrapping the key argument in a non-foldable call and omitting a `FROM` clause avoids both the constant-folding optimizer and the generic table-based rebind check.

### To Reproduce

SQL script (`vuln_001.sql`):

```sql
CREATE TABLE t(a INTEGER);
CREATE INDEX idx ON t USING ART(a);

-- random()*0 keeps the argument non-foldable so the bind data is not
-- discarded early, and the lack of a FROM clause means no table read
-- dependency is registered for the catalog-version rebind check.
PREPARE stmt AS SELECT index_key({'table': 't'}, 'idx', CAST(random() * 0 AS INTEGER) + 42);

EXECUTE stmt;    -- binds once, caches ART& in the prepared statement's bind data
DROP INDEX idx;  -- frees the underlying ART object
EXECUTE stmt;    -- reuses the freed ART reference -> UAF
```

Trigger output (ASan, `build/debug/duckdb`):

```
SUMMARY: AddressSanitizer: heap-use-after-free extension/core_functions/scalar/debug/index_key.cpp:214 in IndexKeyFunction
Shadow bytes around the buggy address:
  ...
=>0x5140000a0000: fa fa fa fa fa fa fa fa[fd]fd fd fd fd fd fd fd
  ...
==2883556==ABORTING
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

gulugulubing:
Fixes: #24191 

#### What this PR does
`IndexKeyBind` resolves the targeted table's `CatalogEntry` via `Catalog::GetEntry`, but did not register a dependency on the owning catalog with the `Binder`. As a result, prepared statements that use `index_key` were not invalidated when the referenced index (or table) was dropped, and executing them afterwards could use stale index metadata.

This PR registers the catalog dependency via `RegisterDBRead` in `IndexKeyBind`, so cached prepared statements are invalidated and rebind after catalog changes. The failure mode from #24191 now surfaces as a clean "index not found" binder error instead of a use-after-free.

#### Scope note
This PR fixes the missing catalog dependency registration, not the underlying index lifetime issue. `IndexKeyBindData` still holds a raw `ART &` reference to the index; the inherent UAF (no concrete reproducer found) should be addressed by the index lifetime refactor in #23570.

### Checklist
- [x] Code formatted via `clang-format`
- [x] Regression test added
- [x] Local `unittest` passed
- [x] Fork CI ran (Windows failed on unrelated `test/sql/parallelism/interquery/concurrent_checkpoint_insert.test:17`)


artjomPlaunov:
@gulugulubing thanks for the PR! 

I think we should wait until https://github.com/duckdb/duckdb/pull/23570 is in, as that may either implicitly fix it, or if not, would make more sense to fix this bug on top of the bigger index lifetime refactor that the PR will introduce.

gulugulubing:
> I think we should wait until #23570 is in, as that may either implicitly fix it, or if not, would make more sense to fix this bug on top of the bigger index lifetime refactor that the PR will introduce.

Thanks for the feedback — that makes sense. I'll close this PR for now and re-check once #23570 lands.

artjomPlaunov:
@gulugulubing I am reopening this, I think this doesn't fix the UAF problem, but it does fix a catalog dependency problem that happens to share the same reproducer as the UAF. Will review this soon.

extension/core_functions/scalar/debug/index_key.cpp
```diff
@@ -132,6 +133,14 @@ static unique_ptr<FunctionData> IndexKeyBind(BindScalarFunctionInput &input) {
 	if (!table_entry.IsDuckTable()) {
 		throw BinderException("index_key: table '%s' is not a DuckDB table", qualified_table);
 	}
+
+	// Register a catalog dependency so that prepared statements are invalidated when the table/index is dropped.
```
artjomPlaunov: rewrite the comment here

// index_key resolves index metadata during binding.
// Register the owning catalog so cached prepared statements
// rebind after catalog changes.

extension/core_functions/scalar/debug/index_key.cpp
```diff
@@ -132,6 +133,14 @@ static unique_ptr<FunctionData> IndexKeyBind(BindScalarFunctionInput &input) {
...
+	// The binder may not be available when this function is called through expression
```
artjomPlaunov: we can also drop these two lines

artjomPlaunov (changes requested):
Looks good to me, just one change for the comment. Could you also merge the latest changes from main into this branch and rerun CI?

Also, can you update the PR title/description, to make it clear that this doesn't fix the inherent UAF, but rather we are fixing a missing catalog dependency registration for index_key, so that prepared statements rebind after the referenced index is dropped. 

Even though we don't have a concrete reproducer, I think the UAF is still lurking, but that should be handled separately with the index lifetime PR.

artjomPlaunov (approved):
Thank you!

gulugulubing:
@artjomPlaunov All 7 failed checks are infrastructure issues (vcpkg downloads 503 / duckdb-data download timeouts). None related to the code change. Could you re-run the failed jobs?

artjomPlaunov:
@gulugulubing let's just rerun the whole CI and hopefully GH wont be acting up again

gulugulubing:
@artjomPlaunov GH service is still flaky. I'll wait a few hours and re-trigger CI.

artjomPlaunov:
CI failure is unrelated

PR #24258, opened 2026-07-29
Fix table filter canonicalization in common subplan optimizer

### LINKED ISSUE #23679: UNION ALL leads to incorrect result
### What happens?

Hi team, I think there's a regression for union all; the following SQL fails on the latest main branch, but works fine in v1.5.4

### To Reproduce

```sql
memory D  CREATE TABLE t(a INT, b INT);
memory D INSERT INTO t VALUES (1, 10), (2, 20), (3, 30);

-- Should be 2,3 on the second column
memory D SELECT 'a' AS q, count(*) AS c FROM t WHERE a > 1
         UNION ALL 
         SELECT 'b', count(*) FROM t WHERE b > 1;
┌─────────┬───────┐
│    q    │   c   │
│ varchar │ int64 │
├─────────┼───────┤
│ a       │     2 │
│ b       │     2 │
└─────────┴───────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

jingdaws:
Fixes #23679.

  Table filter keys in `LogicalGet` refer to positions in `column_ids`. When the common subplan optimizer builds a canonical plan signature, it temporarily replaces `column_ids` with all table columns, but previously left the table filter keys
  unchanged.

  This could give filters on different base columns the same signature. For example, `a > 1` and `b > 1` could both use projection index `0`, causing one branch of a `UNION ALL` to reuse the wrong subplan.

  Remap table filter keys to their base column indexes while serializing the signature, then restore their original indexes afterward. If a unique reverse mapping cannot be constructed, common subplan materialization is skipped for that operator.

  The regression tests cover different filter columns, multiple filters, `IN` filters, and verify that identical filtered subplans can still be materialized.

### DIFF
--- a/src/optimizer/common_subplan_optimizer.cpp
+++ b/src/optimizer/common_subplan_optimizer.cpp
@@ -64,7 +64,7 @@ class PlanSignatureTableIndexMap {
 public:
 	explicit PlanSignatureTableIndexMap(ArenaAllocator &allocator_p)
 	    : allocator(allocator_p), table_index_map(allocator), to_canonical_table_index(allocator),
-	      restore_original_table_index(allocator) {
+	      restore_original_table_index(allocator), restore_original_table_filter_index(allocator) {
 	}
 
 public:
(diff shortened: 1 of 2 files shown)


lnkuiper:
Good find and great fix. Thanks!

PR #24260, opened 2026-07-29
Fix failed checkpoint marker recovery

p1p1bear:
**Problem**
The checkpoint marker is written to the main WAL first, while the `.wal.checkpoint` file is created lazily when concurrent transactions are present. 

If the checkpoint does not complete and the `.wal.checkpoint` file has not yet been created, recovery replays the main WAL, but the original implementation leaves the failed marker in place.

Once the main WAL becomes writable again after recovery, a subsequent checkpoint may write a second marker, corrupting the recovery boundary.

**Fix**
Remove the failed marker!
- Track the start and end offsets of each complete checkpoint marker.
- After validating the WAL generation, use `.wal.recovery` to remove the failed marker while preserving transactions before and after it.
- Continue treating incomplete markers as a torn WAL.
- If multiple markers are found, fail immediately before modifying any files.

**Other && TODO**
Creating `.wal.checkpoint` earlier could reduce the likelihood of this state, but it would affect persistence ordering, empty-file cleanup, and `Initialized()` semantics.

It would also not handle WAL files left behind by older versions. For this reason, this change preserves lazy creation and fixes the recovery logic instead.

### DIFF
--- a/src/storage/wal_replay.cpp
+++ b/src/storage/wal_replay.cpp
@@ -1359,7 +1359,7 @@ void WriteAheadLogDeserializer::ReplayUpdate() {
 
 void WriteAheadLogDeserializer::ReplayCheckpoint() {
 	if (state.checkpoint_position.IsValid()) {
-		throw InvalidInputException("WAL cannot contain more than one checkpoint marker");
+		throw DataCorruptionException("WAL cannot contain more than one checkpoint marker");
 	}
 	auto entry = WALCheckpoint::Deserialize(deserializer);
 	state.checkpoint_id = entry.meta_block;
--- a/src/storage/wal_replay.cpp
+++ b/src/storage/wal_replay.cpp
@@ -56,6 +56,7 @@ public:
 	idx_t wal_version = 1;
 	optional_idx current_position;
 	optional_idx checkpoint_position;
+	optional_idx checkpoint_end_position;
 	optional_idx expected_checkpoint_id;
 	WALReplayState replay_state;
 
@@ -293,6 +294,8 @@ private:
 	                                    WALReplayState replay_state = WALReplayState::MAIN_WAL);
 	void CopyOverWAL(BufferedFileReader &reader, FileHandle &target, data_ptr_t buffer, idx_t buffer_size,
 	                 idx_t copy_end);
+	void RemoveCheckpointFromWAL(const ReplayState &checkpoint_state, BufferedFileReader &main_wal_reader,
+	                             const string &recovery_path);
 	void MergeIntoRecoveryWAL(Connection &con, const ReplayState &checkpoint_state, BufferedFileReader &main_wal_reader,
 	                          const string &recovery_path, unique_ptr<FileHandle> checkpoint_handle);
 
@@ -346,6 +349,30 @@ void WriteAheadLogReplayer::CopyOverWAL(BufferedFileReader &reader, FileHandle &
 
 		target.Write(buffer, read_count);
 	}
+}
+
+void WriteAheadLogReplayer::RemoveCheckpointFromWAL(const ReplayState &checkpoint_state,
+                                                    BufferedFileReader &main_wal_reader, const string &recovery_path) {
+	D_ASSERT(checkpoint_state.checkpoint_position.IsValid());
+	D_ASSERT(checkpoint_state.checkpoint_end_position.IsValid());
+
+	auto recovery_handle =
+	    fs.OpenFile(recovery_path, FileFlags::FILE_FLAGS_WRITE | FileFlags::FILE_FLAGS_FILE_CREATE_NEW);
+
+	static constexpr idx_t BATCH_SIZE = Storage::DEFAULT_BLOCK_SIZE;
+	auto buffer = make_uniq_array<data_t>(BATCH_SIZE);
+
+	// Preserve every WAL entry except the failed checkpoint marker itself.
+	main_wal_reader.Reset();
+	CopyOverWAL(main_wal_reader, *recovery_handle, buffer.get(), BATCH_SIZE,
+	            checkpoint_state.checkpoint_position.GetIndex());
+	main_wal_reader.Seek(checkpoint_state.checkpoint_end_position.GetIndex());
+	CopyOverWAL(main_wal_reader, *recovery_handle, buffer.get(), BATCH_SIZE, main_wal_reader.FileSize());
+
+	recovery_handle->Sync();
+	recovery_handle.reset();
+	main_wal_reader.handle.reset();
+	fs.MoveFile(recovery_path, main_wal_path);
 }
 
 void WriteAheadLogReplayer::MergeIntoRecoveryWAL(Connection &con, const ReplayState &checkpoint_state,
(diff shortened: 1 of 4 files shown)


src/storage/wal_replay.cpp
```diff
@@ -348,6 +351,30 @@ void WriteAheadLogReplayer::CopyOverWAL(BufferedFileReader &reader, FileHandle &
...
+
+	auto recovery_handle =
+	    fs.OpenFile(recovery_path, FileFlags::FILE_FLAGS_WRITE | FileFlags::FILE_FLAGS_FILE_CREATE_NEW);
+
+	static constexpr idx_t BATCH_SIZE = Storage::DEFAULT_BLOCK_SIZE;
+	auto buffer = make_uniq_array<data_t>(BATCH_SIZE);
+
+	// Preserve every WAL entry except the failed checkpoint marker itself.
+	main_wal_reader.Reset();
+	CopyOverWAL(main_wal_reader, *recovery_handle, buffer.get(), BATCH_SIZE,
```
Mytherin: Shouldn't a checkpoint marker always be at the end of the WAL, hence we can just truncate instead of doing this more complex copy / move operation?

src/storage/wal_replay.cpp
```diff
@@ -1307,6 +1358,9 @@ void WriteAheadLogDeserializer::ReplayUpdate() {
 }
 
 void WriteAheadLogDeserializer::ReplayCheckpoint() {
+	if (state.checkpoint_position.IsValid()) {
+		throw InvalidInputException("WAL cannot contain more than one checkpoint marker");
```
Mytherin: nit: `DataCorruptionException`

Mytherin (commented):
Thanks for the PR! Looks good - I wonder if we can simplify this though

artjomPlaunov:
@p1p1bear can we also add this SQL test:

```sql
# name: test/sql/storage/concurrent_wal/double_interrupted_checkpoint.test_slow
# description: Test recovery after two interrupted checkpoints where the second has a concurrent commit
# group: [concurrent_wal]

statement ok
ATTACH '{TEST_DIR}/double_interrupted_checkpoint.db' AS db

statement ok
SET wal_autocheckpoint = '1TB'

statement ok
PRAGMA disable_checkpoint_on_shutdown

statement ok
CREATE TABLE db.t(i BIGINT)

statement ok
INSERT INTO db.t SELECT * FROM range(10)

statement ok
CHECKPOINT db

restart

statement ok
ATTACH '{TEST_DIR}/double_interrupted_checkpoint.db' AS db

statement ok
SET wal_autocheckpoint = '1TB'

statement ok
PRAGMA disable_checkpoint_on_shutdown

statement ok
INSERT INTO db.t VALUES (10)

statement ok
PRAGMA debug_checkpoint_abort = 'before_header'

statement error
CHECKPOINT db
----
Checkpoint aborted before header write

restart

query I
SELECT count(*) FROM glob('{TEST_DIR}/double_interrupted_checkpoint.db.wal*')
----
1

statement ok
ATTACH '{TEST_DIR}/double_interrupted_checkpoint.db' AS db

statement ok
SET wal_autocheckpoint = '1TB'

statement ok
PRAGMA disable_checkpoint_on_shutdown

statement ok
SET debug_checkpoint_sleep_ms = 500

statement ok
PRAGMA debug_checkpoint_abort = 'before_header'

concurrentloop threadid 0 2

onlyif threadid=0
statement ok
SELECT sleep_ms(200)

onlyif threadid=1
statement ok
SELECT sleep_ms(100)

onlyif threadid=1
statement ok
INSERT INTO db.t
SELECT 11
WHERE sleep_ms(300) IS NULL

onlyif threadid=0
statement error
CHECKPOINT db
----
Checkpoint aborted before header write

endloop

restart

query I
SELECT count(*) FROM glob('{TEST_DIR}/double_interrupted_checkpoint.db.wal*')
----
2

statement ok
ATTACH '{TEST_DIR}/double_interrupted_checkpoint.db' AS db

query I
SELECT count(*) FROM db.t
----
12
```

src/storage/wal_replay.cpp
p1p1bear: Fixed. :)

p1p1bear:
> @p1p1bear can we also add this SQL test:
> 
> ```sql
> # name: test/sql/storage/concurrent_wal/double_interrupted_checkpoint.test_slow
> # description: Test recovery after two interrupted checkpoints where the second has a concurrent commit
> # group: [concurrent_wal]
> 
> statement ok
> ATTACH '{TEST_DIR}/double_interrupted_checkpoint.db' AS db
> 
> statement ok
> SET wal_autocheckpoint = '1TB'
> 
> statement ok
> PRAGMA disable_checkpoint_on_shutdown
> 
> statement ok
> CREATE TABLE db.t(i BIGINT)
> 
> statement ok
> INSERT INTO db.t SELECT * FROM range(10)
> 
> statement ok
> CHECKPOINT db
> 
> restart
> 
> statement ok
> ATTACH '{TEST_DIR}/double_interrupted_checkpoint.db' AS db
> 
> statement ok
> SET wal_autocheckpoint = '1TB'
> 
> statement ok
> PRAGMA disable_checkpoint_on_shutdown
> 
> statement ok
> INSERT INTO db.t VALUES (10)
> 
> statement ok
> PRAGMA debug_checkpoint_abort = 'before_header'
> 
> statement error
> CHECKPOINT db
> ----
> Checkpoint aborted before header write
> 
> restart
> 
> query I
> SELECT count(*) FROM glob('{TEST_DIR}/double_interrupted_checkpoint.db.wal*')
> ----
> 1
> 
> statement ok
> ATTACH '{TEST_DIR}/double_interrupted_checkpoint.db' AS db
> 
> statement ok
> SET wal_autocheckpoint = '1TB'
> 
> statement ok
> PRAGMA disable_checkpoint_on_shutdown
> 
> statement ok
> SET debug_checkpoint_sleep_ms = 500
> 
> statement ok
> PRAGMA debug_checkpoint_abort = 'before_header'
> 
> concurrentloop threadid 0 2
> 
> onlyif threadid=0
> statement ok
> SELECT sleep_ms(200)
> 
> onlyif threadid=1
> statement ok
> SELECT sleep_ms(100)
> 
> onlyif threadid=1
> statement ok
> INSERT INTO db.t
> SELECT 11
> WHERE sleep_ms(300) IS NULL
> 
> onlyif threadid=0
> statement error
> CHECKPOINT db
> ----
> Checkpoint aborted before header write
> 
> endloop
> 
> restart
> 
> query I
> SELECT count(*) FROM glob('{TEST_DIR}/double_interrupted_checkpoint.db.wal*')
> ----
> 2
> 
> statement ok
> ATTACH '{TEST_DIR}/double_interrupted_checkpoint.db' AS db
> 
> query I
> SELECT count(*) FROM db.t
> ----
> 12
> ```

Thanks! I added this case in the new commit.

src/storage/wal_replay.cpp
p1p1bear: @Mytherin  Thanks for the review!

You are right that, under normal operation, the checkpoint marker should be at the logical end of the main WAL.

The copy/move path is mainly for backward compatibility. An affected older version could replay a failed checkpoint without removing its marker, make the main WAL writable again, and then append later committed transactions after that stale marker. Truncating at the marker would discard those transactions.

Repeated failures could theoretically leave multiple markers, although that should be rare. This PR does not attempt to repair that case: it rejects the WAL before modifying any files. The copy/move logic only handles a single stale marker while preserving the committed suffix.

If backward recovery for WALs produced by affected versions is not required, and entries after the marker can be treated as corruption, I am happy to simplify this to truncation.

Mytherin: I would say let's do a truncate and throw an error if the WAL marker isn't at the end of the file (similar to the multiple checkpoint markers).

p1p1bear: Fixed now :)

Mytherin:
Thanks!

PR #24261, opened 2026-07-29
Add `Upper`, `Lower` and `StartsWith` helpers to `Identifier`

JelteF:
Code dealing with identifiers regularly needs the raw name in a normalized
casing, or wants to check a case-insensitive prefix. Doing that today requires
reaching for `StringUtil::Lower(id.GetIdentifierName())` and friends, which is
verbose and forces callers to unwrap the identifier just to do something that
is inherently identifier-shaped.

`Identifier::Upper()` and `Identifier::Lower()` return the raw name converted
to upper/lower case, and `Identifier::StartsWith(prefix)` performs a
case-insensitive prefix check, matching the case-insensitive semantics of the
rest of the type. All existing call sites are converted to use them, including
the `DummyBinding::DUMMY_NAME` prefix checks that were previously
case-sensitive.

### DIFF
--- a/extension/core_functions/scalar/debug/index_key.cpp
+++ b/extension/core_functions/scalar/debug/index_key.cpp
@@ -29,7 +29,7 @@ static TableDescription ExtractTableDescription(const child_list_t<LogicalType> 
 	fields["table"] = "";
 
 	for (idx_t i = 0; i < field_types.size(); i++) {
-		auto field_name = StringUtil::Lower(field_types[i].first.GetIdentifierName());
+		auto field_name = field_types[i].first.Lower();
 
 		if (fields.find(field_name) == fields.end()) {
 			throw BinderException("index_key: unknown field '%s' in path", field_types[i].first.GetIdentifierName());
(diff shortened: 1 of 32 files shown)


Mytherin:
Thanks - but I think the lower / upper casing seems to almost always be an issue in subsequent code by looking at the usage (i.e. incorrectly using an `unordered_map<string, ...` instead of an identifier map, incorrectly passing in a `string` where we should be passing in an `Identifier`, etc.).  This makes this incorrect pattern less recognizable and kind of encourages it in code - so I think this is not a good idea. I would rather fix the callsites to consistenly use identifiers instead of falling back to the old way of lowercasing strings.

JelteF:
I think that's fair. 

I'll create a new PR for the StartsWith change only then, because I think that still makes sense.

I'm already working on one to make `ParseOption` and `ParseCopyOption` take `Identifier`. That should remove most of the cases where the Upper/Lower functions are needed.

After that I think it might still make sense to have an Upper/Lower function, but have them return Identifier instead. Primarily for making error messages consistent.

PR #24263, opened 2026-07-29
Add detect_string_types option to read_json

guizmaii:
Implements the feature requested in discussion https://github.com/duckdb/duckdb/discussions/24259.

## Motivation

`read_json` schema auto-detection converts JSON strings that look like dates/times/timestamps/UUIDs into those types. This is often desirable, but when JSON documents are passed through SQL steps and must round-trip unchanged, the original text becomes unrecoverable:

```sql
-- {"formatted_date": "25-11-2025"} is read as DATE and re-serializes as "2025-11-25"
-- {"id": "BBD05AE7-..."} is read as UUID and re-serializes lowercased
SELECT * FROM read_json('docs.json');
```

There is currently no way to opt out of this detection short of specifying the full schema manually via `columns` (which requires knowing the schema up front, defeating the purpose of auto-detection).

## Change

Adds a `detect_string_types` BOOLEAN parameter (default `true`) to `read_json`, `read_ndjson`, `read_json_auto`, and `read_ndjson_auto`. When set to `false`, auto-detection no longer infers `DATE`/`TIME`/`TIMESTAMP`/`UUID` from strings — they stay `VARCHAR`, preserving the original text exactly:

```sql
SELECT * FROM read_json('docs.json', detect_string_types=false);
```

- `convert_strings_to_integers` remains independent: with `detect_string_types=false, convert_strings_to_integers=true`, integer-like strings are still detected as `BIGINT` while temporals/UUIDs stay `VARCHAR`.
- Explicitly requested types (via `columns`) and `dateformat`/`timestampformat` are unaffected.
- Default behavior is unchanged.

The implementation gates the string candidate types (`UUID`/`TIMESTAMP`/`DATE`/`TIME`) in `JSONStructureNode::InitializeCandidateTypes` on the new option, so no candidate refinement work is done for these types when disabled.

## Testing

Added `test/sql/json/table/read_json_detect_string_types.test` covering default behavior, the new flag (top-level and nested structs/lists), both combinations with `convert_strings_to_integers`, `records=false` (bare values), multi-file schema merging with `union_by_name`, inferred MAP value types, interaction with `columns`/`dateformat`, and NULL argument handling.

### DIFF
--- a/extension/json/include/json_reader_options.hpp
+++ b/extension/json/include/json_reader_options.hpp
@@ -116,6 +116,8 @@ struct JSONReaderOptions {
 	idx_t maximum_sample_files = 32;
 	//! Whether we auto-detect and convert JSON strings to integers
 	bool convert_strings_to_integers = false;
+	//! Whether we auto-detect DATE/TIME/TIMESTAMP/UUID types from JSON strings
+	bool detect_string_types = true;
 	//! If a struct contains more fields than this threshold with at least 80% similar types,
 	//! we infer it as MAP type
 	idx_t map_inference_threshold = 200;
--- a/extension/json/include/json_structure.hpp
+++ b/extension/json/include/json_structure.hpp
@@ -33,7 +33,8 @@ public:
 	JSONStructureDescription &GetOrCreateDescription(LogicalTypeId type);
 
 	bool ContainsVarchar() const;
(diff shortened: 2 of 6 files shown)


guizmaii:
Waiting for review

PR #24267, opened 2026-07-29
Add a `StartsWith` helper to `Identifier`

JelteF:
Checking whether an identifier starts with a certain prefix required unwrapping
the identifier with `StringUtil::StartsWith(id.GetIdentifierName(), prefix)`,
which is both verbose and silently case-sensitive - even though every other
operation on `Identifier` is case-insensitive.

`Identifier::StartsWith(prefix)` does the prefix check using
`StringUtil::CIStartsWith`, matching the case-insensitive semantics of the rest
of the type. The `DummyBinding::DUMMY_NAME` prefix checks in the macro/pivot
binders are converted to it, which incidentally makes them case-insensitive as
they should have been.

Related to #24261

### DIFF
--- a/src/common/identifier.cpp
+++ b/src/common/identifier.cpp
@@ -6,6 +6,10 @@
 
 namespace duckdb {
 
+bool Identifier::StartsWith(const string &prefix) const {
+	return StringUtil::CIStartsWith(value, prefix);
+}
+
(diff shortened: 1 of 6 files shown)


lnkuiper:
Thanks!

PR #24269, opened 2026-07-29
Use `Identifier` for COPY and scan option names

JelteF:
Option names are SQL identifiers, but they were carried as strings.
This changes them all to Identifier types which removes a bunch of case
explicit case insensitive comparisons.

Would benefit from #24267


Mytherin:
Thanks!

PR #24271, opened 2026-07-29
Re-enable MERGE for composite row ids

lnkuiper:
Fixes the issue discussed here: https://github.com/duckdb/duckdb/pull/24058#issuecomment-5107569952

The implementation assumed a single-column row identifier, which isn't true for Iceberg. This PR allows composite row ids using a `GroupedAggregateHashTable` and cleans up the implementation a bit.

### DIFF
--- a/src/execution/operator/persistent/physical_insert.cpp
+++ b/src/execution/operator/persistent/physical_insert.cpp
@@ -82,11 +82,14 @@ InsertGlobalState::InsertGlobalState(ClientContext &context, const vector<Logica
 }
 
 InsertLocalState::InsertLocalState(ClientContext &context, const vector<LogicalType> &types,
-                                   const vector<unique_ptr<BoundConstraint>> &bound_constraints)
+                                   const vector<unique_ptr<BoundConstraint>> &bound_constraints, bool deduplicate_rows)
     : collection_index(DConstants::INVALID_INDEX), bound_constraints(bound_constraints) {
 	auto &allocator = Allocator::Get(context);
 	update_chunk.Initialize(allocator, types);
 	append_chunk.Initialize(allocator, types);
+	if (deduplicate_rows) {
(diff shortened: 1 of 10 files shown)


src/execution/operator/persistent/physical_insert.cpp
```diff
@@ -291,7 +295,8 @@ static idx_t PerformOnConflictAction(InsertLocalState &lstate, InsertGlobalState
 
 static void RegisterUpdatedRows(InsertLocalState &lstate, const Vector &row_ids, idx_t count) {
 	// Register all row-ids; a distinct count below `count` means a row-id repeated, which we reject.
-	auto distinct = RegisterRowIds(lstate.updated_rows, row_ids, count);
+	D_ASSERT(lstate.updated_rows);
+	auto distinct = lstate.updated_rows->Register(row_ids, count);
```
Tishj, line 299: I wonder if this needs to check if `ASSUME_UNIQUE` isn't true, since we assert `lstate.updated_rows != nullptr` here?

Tishj: It looks guarded on:
```c++
	if (&tuples == &lstate.update_chunk) {
		// Allow updating duplicate rows for the 'update_chunk'
		RegisterUpdatedRows(lstate, row_ids, conflict_count);
	}
```
But I don't know if that translates to "assume unique isn't true"

Tishj: Actually.. maybe I'm mixing up the call-sites, one is INSERT-only, and the other is MERGE_INTO
This is part of the INSERT path, and `updated_rows` is set when `action_type` is UPDATE, so this is probably fine

Tishj: Github won't let me mark this as resolved, "Failed to resolve thread".
So just read this comment as it being resolved 👍

lnkuiper: It's fine like it is indeed, but it is confusing. Thanks for the review! I'll clarify the behaviour.

PR #24274, opened 2026-07-29
Allow passing an Identifier to SQLQuotedIdentifier directly

JelteF:
`SQLIdentifier` could already be constructed from an `Identifier`, but
`SQLQuotedIdentifier` could not, so callers holding an `Identifier` had to
reach for `GetIdentifierName()` to get at the underlying string first. Giving
both helpers the same set of constructors removes that asymmetry, and lets
`CatalogSearchEntry::WriteOptionallyQuoted` drop its hand-rolled
`"\"" + StringUtil::Replace(name, "\"", "\"\"") + "\""` in favour of
`SQLQuotedIdentifier::ToString`, which does precisely the same escaping.

The added `const char *` overload is needed because `Identifier` is implicitly
constructible from a string literal, which would otherwise make
`SQLQuotedIdentifier("foo")` ambiguous between the `string` and `Identifier`
constructors.

### DIFF
--- a/src/catalog/catalog_search_path.cpp
+++ b/src/catalog/catalog_search_path.cpp
@@ -4,6 +4,7 @@
 #include "duckdb/catalog/catalog.hpp"
 #include "duckdb/common/constants.hpp"
 #include "duckdb/common/exception.hpp"
+#include "duckdb/common/sql_identifier.hpp"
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #20585, opened 2026-01-19
Add max_execution_time setting for per-connection query time limit

dufferzafar:
## What

Adds a new session-scoped setting `max_execution_time` that controls the wall-clock time budget for queries (in milliseconds).

This is akin to PostgreSQL's [`statement_timeout`](https://postgresqlco.nf/doc/en/param/statement_timeout/) and ClickHouse's similarly named [`max_execution_time`](https://clickhouse.com/docs/operations/settings/settings#max_execution_time)

## Why

Ours is an embedded duckdb usecase where we run an in memory DB using the C++ API. 

It's a multi-tenant system where multiple users can submit arbitrary analytical queries. 

Without execution time limits, a single runaway query (e.g., an accidental cartesian product or unbounded recursive CTE) can monopolize resources and impact other tenants. 

We need a mechanism to enforce per-query time budgets at the connection level.

## Usage

```sql
SET max_execution_time=1000;  -- 1 second timeout

-- This should now fail with an Interrupt error!
SELECT COUNT(*) FROM range(100000000) t1, range(1000) t2;
```

## Details

At query start, a deadline is computed using steady_clock. Timeout checks occur at strategic points in the execution pipeline (operators, sort, aggregation, table scans, etc.). When the deadline is exceeded, an InterruptException is thrown.

The overhead should be minimal - an atomic load and compare at existing interrupt check points, but I'll be interested to see if there is a better way of doing this.

## Questions

My goal with the PR is to assess whether this is the right direction. Specifically:

1. Should timeout throw a distinct exception type (e.g., TimeoutException) rather than reusing InterruptException?
2. Are there other execution paths (which I missed) that need timeout checks?

PR #24276, opened 2026-07-29
Shell query timeout

hannes:
This PR adds a CLI query timeout that can be set either with `.timeout 1m` or via a command line flag. When the timeout hits, an error is thrown that shows some progress info.


Mytherin:
Thanks for the PR! There's [`max_execution_time` implemented here](https://github.com/duckdb/duckdb/pull/20585), maybe that can be used here?

PR #24277, opened 2026-07-29
Prune row groups for Parquet INTERVAL columns with bloom filters

### LINKED ISSUE #24020: Parquet reader doesn't prune row group for INTERVAL type
### What happens?

See repro for details.
Notice, this might not be a good first issue.

### To Reproduce

```sql
memory D COPY (
             SELECT
                 ((i % 100) * 2) * INTERVAL '1 day' AS iv,
                 i::BIGINT AS payload
             FROM range(20480) AS t(i)
         )
         TO '/tmp/pruning_interval.parquet'
         (
             FORMAT PARQUET,
             ROW_GROUP_SIZE 2048,
             DICTIONARY_SIZE_LIMIT 1000,
             WRITE_BLOOM_FILTER TRUE,
             OVERWRITE TRUE
         );

-- shouldn't access any row group
memory D EXPLAIN ANALYZE
         SELECT sum(payload)
         FROM read_parquet('/tmp/pruning_interval.parquet')
         WHERE iv = INTERVAL '101 days';
╭─ Summary ───────────╮
│ Total Time: 0.0025s │
│ Data Read: 107.2 kB │
╰─────────────────────╯
│ Aggregates: sum(#0)                              │
╭─ Projection ────────────┴────────────────────────╮
│ Projections: payload                             │
│ Function: READ_PARQUET                           │
│ Projections: payload                             │
│ normalized_interval(iv) = '3 months 11           │
│ days'::INTERVAL                                  │
│ Total Files Read: 1                              │
│ Filename(s):                                     │
│ /tmp/pruning_interval.parquet                    │
│ Row Groups Scanned: 10 / 10                      │
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

u70b3:
Closes #24020.

## Problem

Parquet `INTERVAL` has no defined sort order, so min/max statistics cannot be used for row-group pruning. Although Bloom filters are available, DuckDB normalizes intervals before comparison: equal values such as `INTERVAL '101 days'` and `INTERVAL '3 months 11 days'` can have different 12-byte Parquet encodings and therefore different XXH64 hashes.

The predicate is also bound as `normalized_interval(column) = constant`, which the existing Bloom filter extraction did not recognize. Probing a filter containing only raw-encoding hashes with a normalized hash would risk false negatives, so this requires both writer and reader changes.

## Approach

- The writer inserts both the standard Parquet plain-encoding hash and a normalized-encoding hash for `INTERVAL` dictionary values. Bloom filters are sized for two entries per value.
- Files containing Bloom-filtered `INTERVAL` columns are marked with `duckdb.interval_bloom_filter=normalized-v1`.
- The reader uses the normalized hash only when this marker is present and the column is `INTERVAL` over `FIXED_LEN_BYTE_ARRAY(12)`. Equality, `IN`, `parquet_bloom_probe`, and nested `LIST`/`STRUCT` paths use the same capability and schema checks.

## Compatibility

Files without the marker retain the previous conservative behavior. Standard readers remain compatible because the raw hashes are still present; the additional normalized hashes can only increase false positives, not introduce false negatives.

## Testing

The new `bloom_filter_interval.test` covers absent equality and `IN` values pruning `0 / 10` row groups, marker and `parquet_bloom_probe` behavior, nested `LIST`/`STRUCT` columns, and equivalent intervals with different month/day and millisecond/day encodings.

The existing Parquet Bloom filter tests and the full `build/reldebug/test/unittest` suite pass.

### DIFF
--- a/extension/parquet/parquet_statistics.cpp
+++ b/extension/parquet/parquet_statistics.cpp
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "parquet_decimal_utils.hpp"
+#include "parquet_interval.hpp"
 #include "parquet_timestamp.hpp"
 #include "parquet_float16.hpp"
 #include "reader/string_column_reader.hpp"
@@ -880,12 +881,7 @@ static optional<uint64_t> TryHashTimestamp(const Value &constant, const ParquetC
 }
 
 static uint64_t HashNormalizedInterval(const Value &constant) {
-	auto interval = constant.GetValue<interval_t>().Normalize();
-	data_t bytes[12];
-	Store<uint32_t>(static_cast<uint32_t>(interval.months), bytes);
-	Store<uint32_t>(static_cast<uint32_t>(interval.days), bytes + sizeof(uint32_t));
-	Store<uint32_t>(static_cast<uint32_t>(interval.micros / Interval::MICROS_PER_MSEC), bytes + sizeof(uint32_t) * 2);
-	return duckdb_zstd::XXH64(bytes, sizeof(bytes), 0);
+	return ParquetIntervalUtils::HashNormalized(constant.GetValue<interval_t>());
 }
 
 // TODO we can only this if the parquet representation of the type exactly matches the duckdb rep!
--- a/extension/parquet/parquet_statistics.cpp
+++ b/extension/parquet/parquet_statistics.cpp
@@ -694,21 +694,39 @@ static optional_ptr<const Expression> GetOptionalFilterChild(const Expression &e
 	return nullptr;
 }
 
+static bool UsesNormalizedIntervalHash(ParquetBloomFilterHashStrategy hash_strategy) {
+	return hash_strategy == ParquetBloomFilterHashStrategy::NORMALIZED_INTERVAL_V1;
+}
+
+static bool IsBloomFilterColumn(const Expression &expr, ParquetBloomFilterHashStrategy hash_strategy) {
+	if (expr.GetExpressionClass() == ExpressionClass::BOUND_REF) {
+		return expr.Cast<BoundReferenceExpression>().Index() == 0;
+	}
+	if (!UsesNormalizedIntervalHash(hash_strategy) || expr.GetExpressionClass() != ExpressionClass::BOUND_FUNCTION) {
+		return false;
+	}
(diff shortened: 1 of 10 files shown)


extension/parquet/parquet_statistics.cpp
```diff
@@ -862,10 +879,20 @@ static optional<uint64_t> TryHashTimestamp(const Value &constant, const ParquetC
...
 }
 
+static uint64_t HashNormalizedInterval(const Value &constant) {
+	auto interval = constant.GetValue<interval_t>().Normalize();
+	data_t bytes[12];
+	Store<uint32_t>(static_cast<uint32_t>(interval.months), bytes);
+	Store<uint32_t>(static_cast<uint32_t>(interval.days), bytes + sizeof(uint32_t));
+	Store<uint32_t>(static_cast<uint32_t>(interval.micros / Interval::MICROS_PER_MSEC), bytes + sizeof(uint32_t) * 2);
+	return duckdb_zstd::XXH64(bytes, sizeof(bytes), 0);
+}
```
lnkuiper, line 885: In `parquet_write_operators.hpp` this is implemented as:
```C++
	template <class SRC, class TGT>
	static optional<uint64_t> GetExtraBloomFilterHash(const SRC &source_value, const TGT &) {
		// Preserve the Parquet plain-encoding hash and add the hash used by DuckDB comparisons.
		auto normalized_value = Operation<SRC, TGT>(source_value.Normalize());
		return XXHash64<SRC, TGT>(normalized_value);
	}
```
Any reason this is different?

lnkuiper (commented):
Thanks for the PR! Looks good, just one comment:

extension/parquet/parquet_statistics.cpp
u70b3: Thanks! The two implementations were equivalent, but there was no reason to keep them separate. I refactored the interval encoding and normalized hash into a shared helper so the reader and writer cannot drift, which should make this easier to maintain.

lnkuiper:
Thanks! LGTM now

PR #24278, opened 2026-07-29
Take `vector<Identifier>` in the table function bind callbacks

JelteF:
The names a table function returns become the column names of its binding,
which resolve case-insensitively, so `table_function_bind_t` and
`table_function_bind_operator_t` now hand back `vector<Identifier>` instead of
`vector<string>` - matching `TableFunctionBindInput::input_table_names` and
`BoundStatement::names`, which were already identifiers. This removes the
`StringsToIdentifiers` conversions the binder and the multi-file reader were
doing on every bind, and the COPY binds plus the arrow and parquet
serialization paths convert at their own boundaries since those keep operating
on plain strings.


Mytherin:
Thanks! LGTM - will just be a bit annoying to get in due to all the patches required :)

JelteF:
It definitely was a bit annoying to get all the patches working. But the build is green now. :shipit:

Mytherin:
Thanks!

PR #24305, opened 2026-07-30
No pushdown for functions with side effect

### LINKED ISSUE #24304: Filter pushdown skips side-effecting projection rows
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE SEQUENCE s START 1;
memory D SELECT i, v, currval('s') AS after_seq
         FROM (SELECT i, nextval('s') AS v FROM range(10) t(i))
         WHERE i = 0;
┌───────┬───────┬───────────┐
│   i   │   v   │ after_seq │
│ int64 │ int64 │   int64   │
├───────┼───────┼───────────┤
│     0 │     1 │         1 │
└───────┴───────┴───────────┘
memory D SELECT currval('s') AS final_seq;
┌───────────┐
│ final_seq │
│   int64   │
├───────────┤
│         1 │
└───────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24304

In this PR, I checked expression side effect before pushing down projection, and skip if it does.

### DIFF
--- a/src/include/duckdb/planner/expression.hpp
+++ b/src/include/duckdb/planner/expression.hpp
@@ -50,6 +50,7 @@ public:
 
 	virtual bool IsVolatile() const;
 	virtual bool IsConsistent() const;
+	virtual bool HasSideEffects() const;
 	virtual bool PropagatesNullValues() const;
 	virtual bool IsFoldable() const;
 	virtual bool CanThrow() const;
--- a/src/include/duckdb/planner/expression/bound_function_expression.hpp
+++ b/src/include/duckdb/planner/expression/bound_function_expression.hpp
@@ -51,6 +51,7 @@ public:
 
 	bool IsVolatile() const override;
 	bool IsConsistent() const override;
+	bool HasSideEffects() const override;
 	bool IsFoldable() const override;
 	bool CanThrow() const override;
 	string ToString() const override;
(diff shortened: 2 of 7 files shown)


Mytherin:
Thanks - but isn't `Volatile()` enough here? Why do we need a separate flag?

dentiny:
> Thanks - but isn't `Volatile()` enough here? Why do we need a separate flag?

Thanks for the quick reply! Yes, for example, function `random` is volatile, but doesn't have side-effect, so it could be safely skipped.

Mytherin:
Actually not sure I agree with this - this is what Postgres has to say:

```sql
CREATE SEQUENCE pushdown_projection_side_effect_seq START 1;
SELECT i, v, currval('pushdown_projection_side_effect_seq') AS after_seq
FROM (
    SELECT i, nextval('pushdown_projection_side_effect_seq') AS v
    FROM generate_series(0,9) t(i)
)
WHERE i = 0;

 i | v | after_seq 
---+---+-----------
 0 | 1 |         1


SELECT currval('pushdown_projection_side_effect_seq');
 currval 
---------
       1
```

So we currently match Postgres' behavior. This PR makes us deviate from Postgres' behavior. I don't think we need to ensure that `nextval` is always executed / I don't think this is desirable.

dentiny:
That's a good point!

PR #24312, opened 2026-07-30
Fix encrytion util access

### LINKED ISSUE #24311: Invalid memory access when read parquet
### What happens?

See repro for details.

### To Reproduce

```sql
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D COPY (SELECT 42 AS x) TO '/private/tmp/duckdb_meta_cache_second_read.parquet'
memory D SELECT sum(x) FROM read_parquet('/private/tmp/duckdb_meta_cache_second_read.parquet', encryption_config={footer_key:'key256'});
┌────────┐
│ sum(x) │
│ int128 │
├────────┤
│     42 │
└────────┘
memory D SELECT sum(x) FROM read_parquet('/private/tmp/duckdb_meta_cache_second_read.parquet', encryption_config={footer_key:'key256'});
INTERNAL Error:
Attempted to dereference shared_ptr that is NULL!
```

### OS:

macos

### DuckDB Version:

a706d2c9e3

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24311

The issue is
- When parquet metadata enabled, first access suffer cache miss, so we initialize the encryption util
- On the second access, we have a cache hit, so encryption util initialization is skipped, all later accesses suffer invalid memory access

There're two ways I could think of
- Ensure encryption util is initialized everywhere before access (basically all read functions)
- or, eager initialize the encryption util. Since it doesn't involve any IO operations so pick this way to reduce mind overhead

### DIFF
--- a/extension/parquet/parquet_reader.cpp
+++ b/extension/parquet/parquet_reader.cpp
@@ -1282,9 +1282,9 @@ ParquetReader::ParquetReader(ClientContext &context_p, OpenFileInfo file_p, Parq
 		}
 	} else {
 		metadata = std::move(metadata_p);
-		if (parquet_options.encryption_config) {
-			encryption_util = context_p.db->GetEncryptionUtil(true);
-		}
+	}
+	if (parquet_options.encryption_config && !encryption_util) {
+		encryption_util = context_p.db->GetEncryptionUtil(true);
 	}
 	InitializeSchema(context_p);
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24334, opened 2026-07-30
Fix compilation on Windows

staticlibs:
`RowIdHandling` enum includes a member named `ERROR` that clashes with a macro defined in Windows headers. `windows_undefs.hpp` that is supposed to undef this macro, at least in some build configurations, is not included when `RowIdHandling` is being defined.

This PR adds include of the `windows_undefs.hpp` just before defining the problematic enum.


Mytherin:
Thanks!

PR #24336, opened 2026-07-31
Compress fully-visible partial deletes into a bitmask

baolinhuang:
## Problem

A partially-deleted vector keeps its delete ids in a per-row array of transaction_t: 2048 rows x 8 bytes = 16 KiB per vector, allocated from the row group's FixedSizeAllocator (tagged BASE_TABLE). Deleting even a single row in a vector still pays the full 16 KiB, so a table with many partially-deleted vectors holds on to memory far out of proportion to the number of deleted rows.

Building on the checkpoint-time watermark compression from #23964: once a vector's delete ids are all below the lowest active start watermark they are visible to every current and future transaction, so the exact per-row transaction ids no longer carry information - only "deleted or not" matters.

## Approach

Model the delete side as an explicit state machine DeleteIdState { CONSTANT, MASKED, ARRAY }:

- ARRAY    - the existing per-row transaction_t array, required while any delete
             is uncommitted or still visible to an older snapshot.
- MASKED   - a ~256 byte ValidityMask (1 bit/row) once every delete id is below
             the watermark. valid bit == deleted.
- CONSTANT - the whole vector is deleted (unchanged).

At checkpoint a fully-visible partially-deleted vector folds its 16 KiB array into a mask; a later delete on a masked vector lazily rebuilds the array (MASKED -> ARRAY). The on-disk format is unchanged - Write already serializes the same ValidityMask - so ChunkVectorInfo::Read now loads straight into MASKED instead of inflating back to a per-row array, and the saving survives a restart.

## Measurements

Simple half-delete on 10M rows:

    CREATE TABLE t AS SELECT range::BIGINT AS id FROM range(10000000);
    CHECKPOINT;
    DELETE FROM t WHERE id % 2 = 0;
    -- per-row delete-id arrays still materialized
    SELECT SUM(memory_usage_bytes) FROM duckdb_memory() WHERE tag='BASE_TABLE';
    --> 85983232   (~82 MiB)
    CHECKPOINT;
    -- folded into ~256 byte/vector bitmasks
    SELECT SUM(memory_usage_bytes) FROM duckdb_memory() WHERE tag='BASE_TABLE';
    --> 786432     (~0.75 MiB, -99.1%)

The saving survives a restart: reopening the database and scanning the table loads the on-disk mask straight into MASKED, so BASE_TABLE stays at ~0.75 MiB instead of rebounding to the per-row arrays.

### DIFF
--- a/src/include/duckdb/storage/table/chunk_info.hpp
+++ b/src/include/duckdb/storage/table/chunk_info.hpp
@@ -45,7 +45,8 @@ enum class VersionCompressionResult : uint8_t {
 
 //! The storage state of the delete side of a ChunkVectorInfo. Exhaustive and mutually exclusive:
 //! CONSTANT - all rows share constant_delete_id (NOT_DELETED_ID for none-deleted, or a single delete id)
-//! MASKED   - partially deleted, every delete visible to all transactions: stored in deleted_mask, deleted_data freed
+//! MASKED   - partially deleted, all deleted rows share one committed id (mask_delete_id): which rows are
+//!            deleted is stored in deleted_mask, deleted_data freed
 //! ARRAY    - per-row delete ids materialized in deleted_data
 enum class DeleteIdState : uint8_t { CONSTANT, MASKED, ARRAY };
 
@@ -113,8 +114,9 @@ private:
 	IndexPointer GetInitializedDeletedPointer();
 	//! Frees the per-row delete ids (if any)
 	void FreeDeleteData();
-	//! ARRAY -> MASKED: record alive rows as invalid bits, free the per-row delete array
-	void CompressDeleteToMask();
+	//! ARRAY -> MASKED: record alive rows as invalid bits, free the per-row delete array. mask_id is the
+	//! shared committed id of the deleted rows (0 when they are already visible to all transactions)
+	void CompressDeleteToMask(transaction_t mask_id);
 	//! MASKED -> ARRAY: re-materialize the per-row delete array from the bitmask
 	void DecompressDeleteMask();
 
@@ -129,10 +131,14 @@ private:
 	IndexPointer deleted_data;
 	//! The constant delete id (if there is only one, e.g. because the entire vector was deleted in one transaction)
 	transaction_t constant_delete_id;
-	//! Bitmask used in the MASKED state: valid bit == the row is deleted (committed, visible to all
-	//! transactions), invalid bit == the row is alive. Matches the on-disk VECTOR_INFO orientation.
-	//! Only meaningful when delete_state == DeleteIdState::MASKED.
+	//! Bitmask used in the MASKED state: valid bit == the row is deleted, invalid bit == the row is alive.
+	//! Matches the on-disk VECTOR_INFO orientation. Only meaningful when delete_state == DeleteIdState::MASKED.
 	ValidityMask deleted_mask;
+	//! The single committed id shared by every deleted row in the MASKED state. 0 means the deletes are
+	//! visible to all transactions (the value used when read from disk); a non-zero committed id means the
+	//! mask was folded from one committed transaction whose delete is not yet visible to every snapshot.
+	//! Only meaningful when delete_state == DeleteIdState::MASKED.
+	transaction_t mask_delete_id = 0;
 	//! The current delete-side storage state - the single source of truth for the delete side
 	DeleteIdState delete_state = DeleteIdState::CONSTANT;
 	//! Whether a compression pass could achieve anything for this vector: armed by any id modification,
--- a/src/include/duckdb/storage/table/chunk_info.hpp
+++ b/src/include/duckdb/storage/table/chunk_info.hpp
@@ -10,6 +10,7 @@
 
 #include "duckdb/execution/index/index_pointer.hpp"
 #include "duckdb/common/enums/scan_options.hpp"
+#include "duckdb/common/types/validity_mask.hpp"
 
 namespace duckdb {
 class RowGroup;
@@ -41,6 +42,12 @@ enum class VersionCompressionResult : uint8_t {
 	//! (some rows are not deleted)
 	SETTLED
 };
+
+//! The storage state of the delete side of a ChunkVectorInfo. Exhaustive and mutually exclusive:
+//! CONSTANT - all rows share constant_delete_id (NOT_DELETED_ID for none-deleted, or a single delete id)
+//! MASKED   - partially deleted, every delete visible to all transactions: stored in deleted_mask, deleted_data freed
+//! ARRAY    - per-row delete ids materialized in deleted_data
+enum class DeleteIdState : uint8_t { CONSTANT, MASKED, ARRAY };
 
 class ChunkVectorInfo {
 public:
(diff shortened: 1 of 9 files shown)


src/include/duckdb/storage/table/chunk_info.hpp
```diff
@@ -118,6 +129,12 @@ class ChunkVectorInfo {
 	IndexPointer deleted_data;
 	//! The constant delete id (if there is only one, e.g. because the entire vector was deleted in one transaction)
 	transaction_t constant_delete_id;
+	//! Bitmask used in the MASKED state: valid bit == the row is deleted (committed, visible to all
+	//! transactions), invalid bit == the row is alive. Matches the on-disk VECTOR_INFO orientation.
+	//! Only meaningful when delete_state == DeleteIdState::MASKED.
+	ValidityMask deleted_mask;
```
Mytherin, line 136: Can we use the `IndexPointer` to store the byte mask instead, so that this is memory that is tracked by the buffer manager and this memory cost is not always incurred also when not in use? We would need to have a separate `FixedSizeAllocator` for this size (256 bytes) but I think that's worthwhile.

src/include/duckdb/storage/table/chunk_info.hpp
```diff
@@ -42,6 +43,12 @@ enum class VersionCompressionResult : uint8_t {
 	SETTLED
 };
 
+//! The storage state of the delete side of a ChunkVectorInfo. Exhaustive and mutually exclusive:
+//! CONSTANT - all rows share constant_delete_id (NOT_DELETED_ID for none-deleted, or a single delete id)
+//! MASKED   - partially deleted, every delete visible to all transactions: stored in deleted_mask, deleted_data freed
+//! ARRAY    - per-row delete ids materialized in deleted_data
+enum class DeleteIdState : uint8_t { CONSTANT, MASKED, ARRAY };
```
Mytherin, line 51: Instead of having `MASKED` mean "visible to all transactions" maybe we can add support for a single transaction id to `MASKED`? i.e. similar to how `CONSTANT` works but only for the specified rows. I think that improves memory usage for a very realistic use case (all rows that are deleted in a vector are deleted by the same transaction) for what should be no additional cost. When reading from disk that transaction id can then be `0`.

src/storage/table/chunk_info.cpp
```diff
@@ -235,16 +286,53 @@ IndexPointer ChunkVectorInfo::GetInitializedDeletedPointer() {
 		for (idx_t i = 0; i < STANDARD_VECTOR_SIZE; i++) {
 			deleted[i] = constant_id;
 		}
+		delete_state = DeleteIdState::ARRAY;
```
Mytherin, line 320: Should this function handle `::MASKED`?

src/storage/table/chunk_info.cpp
```diff
@@ -665,26 +803,20 @@ unique_ptr<ChunkVectorInfo> ChunkVectorInfo::Read(FixedSizeAllocator &allocator,
...
-		}
-		// the reconstructed ids are all visible to all transactions, so the vector can only be
-		// compressible if every row is deleted (in which case Write emits CONSTANT_INFO instead,
-		// so with the current format the ids are always settled here)
-		result->recheck_compression = !rows_alive;
+		result->deleted_mask.Read(reader, STANDARD_VECTOR_SIZE);
+		// Write only emits VECTOR_INFO for a partial delete: an all-deleted vector becomes
+		// CONSTANT_INFO and an undeleted one becomes EMPTY_INFO, so the mask always has at least one
+		// deleted (valid) and one alive (invalid) bit - never all-valid, never all-invalid.
+		D_ASSERT(!result->deleted_mask.CheckAllValid(STANDARD_VECTOR_SIZE));
```
Mytherin: This shouldn't be an assert - we're deserializing data from disk here, so we can't make any assumptions about what that data holds. Either we should remove these asserts or we should turn these into run-time checks that throw a `DataCorruptionException` if invalidated.

src/storage/table/chunk_info.cpp
```diff
@@ -98,16 +99,34 @@ idx_t ChunkVectorInfo::TemplatedGetSelVector(transaction_t start_time, transacti
...
+	case DeleteIdState::MASKED: {
+		// per-row delete ids come from the bitmask: deleted row == committed id 0, alive row == NOT_DELETED_ID
+		if (HasConstantInsertionId()) {
+			if (!INSERT_OP::UseInsertedVersion(start_time, transaction_id, ConstantInsertId())) {
+				return 0;
+			}
+			idx_t count = 0;
+			for (idx_t i = 0; i < max_count; i++) {
+				transaction_t deleted_id = deleted_mask.RowIsValid(i) ? transaction_t(0) : NOT_DELETED_ID;
+				if (DELETE_OP::IsDeleted(start_time, transaction_id, deleted_id)) {
```
Mytherin: We don't need to check the transaction id for every row here since in the masked case this is the same for all rows

src/storage/table/chunk_info.cpp
```diff
@@ -98,16 +99,34 @@ idx_t ChunkVectorInfo::TemplatedGetSelVector(transaction_t start_time, transacti
...
-			return 0;
+	case DeleteIdState::MASKED: {
+		// per-row delete ids come from the bitmask: deleted row == committed id 0, alive row == NOT_DELETED_ID
+		if (HasConstantInsertionId()) {
+			if (!INSERT_OP::UseInsertedVersion(start_time, transaction_id, ConstantInsertId())) {
+				return 0;
+			}
+			idx_t count = 0;
+			for (idx_t i = 0; i < max_count; i++) {
+				transaction_t deleted_id = deleted_mask.RowIsValid(i) ? transaction_t(0) : NOT_DELETED_ID;
```
Mytherin: An optimization here would be to loop over the validity mask in `uint64_t` chunks at a time similar to what we do in e.g. the UnaryExecutor and check if the `uint64_t` has all rows valid or invalid so we don't need to do bit extractions in this case.

Mytherin (commented):
Thanks for the PR! Looks good and good idea - some comments from my side

src/include/duckdb/storage/table/chunk_info.hpp
baolinhuang: I tried using FixedSizeAllocator to manage the mask's memory. However

1. FixedSizeAllocator's buffer size is fixed at 256 KB. A 256-byte mask segment therefore lives inside a 256 KB
block that holds ~1024 masks.
2. RowVersionManager is per-RowGroup. A row group only spans 60 vectors (122880 / 2048), so the moment a single vector enters the MASKED。

So row group grabs a whole 256 KB block to hold at most 60 × 256 B = 15 KB of masks — ~6% utilization.

The only way to fix the waste and keep it tracked would be a table shared allocator. But FixedSizeAllocator is not thread-safe, so need to add mutex to protect it.

src/storage/table/chunk_info.cpp
baolinhuang: Modified.
Have moved `DecompressDeleteMask()` inside this function. The previous method was call it on upper side.

src/storage/table/chunk_info.cpp
baolinhuang: Modified, throw exception

src/storage/table/chunk_info.cpp
baolinhuang: modified

src/storage/table/chunk_info.cpp
baolinhuang: Modified

Mytherin:
Thanks!

PR #24343, opened 2026-07-31
Initial implementation of `variant_contains` processing function

sebastiaan-dev:
This PR resolves https://github.com/duckdblabs/duckdb-internal/issues/10159, introducing an initial implementation for `variant_contains`. 

The function takes two `VARIANT` values: a haystack and a needle. It returns whether the needle is contained in the haystack. Containment is recursive, so the needle may match the haystack itself or any value nested inside an array or object.

The implementation follows the structural containment semantics of `json_contains`:
- An object needle matches an object containing all of the needle's key-value pairs. The haystack may contain additional keys.
- An array needle matches an array containing every needle element. Element order does not matter. One occurrence of a value in the haystack can satisfy multiple occurrences in the needle.
- A primitive needle matches an equivalent (rank-based) primitive anywhere in the haystack.

Unlike `json_contains`, equality is based on the comparison ranks introduced by the `VARIANT` comparator. Values do not need to have the same underlying `VARIANT` logical type to compare equal. For example, integer widths, `DECIMAL`, and `BIGNUM` values share the `NUMBER` rank and compare by their canonical numeric value. The same applies to types grouped into the `REAL`, `TIMESTAMP`, `TIMESTAMP_TZ`, and `TIME` ranks.

### Examples

```sql
variant_contains({'a': {'b': 42}}::VARIANT, {'b': 42}::VARIANT);
> true

variant_contains([1, 2, 3]::VARIANT, [3, 1]::VARIANT);
> true

variant_contains(1::INTEGER::VARIANT, 1.00::DECIMAL(3, 2)::VARIANT);
>  true

variant_contains({'a': {'b': 42}}::VARIANT, {'a': 42}::VARIANT)
> false
```

### Interface

`variant_contains(variant_haystack, variant_needle)`

#### Parameters

- `variant_haystack` is of type `VARIANT`
- `variant_needle` is of type `VARIANT`

#### Returns

Returns `true` if `variant_needle` is contained in `variant_haystack`, and `false` otherwise. The return type is `BOOLEAN`.

If either argument is SQL `NULL`, the function returns `NULL`.

### Follow-up work


src/function/scalar/variant/variant_contains.cpp
```diff
@@ -0,0 +1,146 @@
...
+	case VariantComparisonType::GEOMETRY:
+	case VariantComparisonType::VARCHAR:
+		return haystack.GetString() == needle.GetString();
+	}
+
+	return false;
+}
+
+static bool RecursiveHaystackWalk(const VariantNode &haystack, const VariantNode &needle) {
+	if (IsEqual(haystack, needle)) {
```
Mytherin, line 94: Is this recursive walk actually necessary, or could we run `variant_comparator` over the haystack / needle and do a search within the blobs? e.g.:

```sql
create macro variant_contains(haystack, needle) as contains(variant_comparator(haystack::VARIANT)::VARCHAR, variant_comparator(needle::VARIANT)::VARCHAR);
select variant_contains({'a': {'b': 42}}, {'b': 42}) as contains;
┌──────────┐
│ contains │
│ boolean  │
├──────────┤
│ true     │
└──────────┘
select variant_contains({'a': {'b': 42}}, {'bc': 42}) as contains;
┌──────────┐
│ contains │
│ boolean  │
├──────────┤
│ false    │
└──────────┘

```

Mytherin (approved):
Thanks for the PR! Looks good to me - one comment

src/function/scalar/variant/variant_contains.cpp
Mytherin: Ah nevermind, this will not work with this requirement:

> An array needle matches an array containing every needle element. Element order does not matter. One occurrence of a value in the haystack can satisfy multiple occurrences in the needle.

Mytherin:
Thanks!

PR #24346, opened 2026-07-31
Fix TopN window elimination with ROW_NUMBER argument ordering

### LINKED ISSUE #23678: top_n_window_elimination ignores ROW_NUMBER argument ORDER BY
### What happens?

top_n_window_elimination ignores ROW_NUMBER's argument ORDER BY.
According to [doc](https://duckdb.org/docs/lts/sql/functions/window_functions#row_numberorder-by-ordering):
> If an ORDER BY clause is specified, the row number is computed within the frame using the provided ordering instead of the frame ordering.

### To Reproduce

To reproduce:

```sql
SELECT id, g, p
FROM (VALUES
  (1, 1, 1),
  (2, 1, 2),
  (3, 1, 3),
  (4, 2, 1),
  (5, 2, 2)
) AS t(id, g, p)
QUALIFY ROW_NUMBER(ORDER BY id DESC) OVER (
  PARTITION BY g
  ORDER BY p ASC
) <= 1
ORDER BY g, id;
```

Run:

```sh
```

Actual:

stdout:
| id | g | p |
| --- | --- | --- |
| 1 | 1 | 1 |
| 4 | 2 | 1 |

Expected:

| id | g | p |
| --- | --- | --- |
| 3 | 1 | 3 |
| 5 | 2 | 2 |

The expected rows are the highest `id` values in each `g` partition because the
`ROW_NUMBER` argument ordering is `ORDER BY id DESC`.

Notes From Minimization:

- Keeping argument-local `ORDER BY id DESC` on `ROW_NUMBER` is required; removing it makes the window order and rewritten order agree.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (a1aacc73df)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

jingdaws:
## Description

  This PR fixes #23678.

  `TopNWindowElimination` rewrites partitioned `ROW_NUMBER` filters into `arg_min` or `arg_max` aggregates. The rewrite uses the window `OrderBy()` but does not preserve `ROW_NUMBER`'s `ArgOrders()`, which can produce
  incorrect results when both orderings are present.

  The rule now skips the rewrite when `ArgOrders()` is non-empty, falling back to the regular window executor. Partitioned Top-N queries without argument ordering remain optimized.

  ## Tests

  Added regression tests for explicit full-partition frames and default frames that grow up to the current row.

### DIFF
--- a/src/optimizer/topn_window_elimination.cpp
+++ b/src/optimizer/topn_window_elimination.cpp
@@ -627,6 +627,9 @@ bool TopNWindowElimination::CanOptimize(LogicalOperator &op) {
 	}
 	auto &window_expr = window.expressions[0]->Cast<BoundWindowExpression>();
 
+	if (!window_expr.ArgOrders().empty()) {
+		return false;
+	}
 	if (window_expr.OrderBy().size() != 1) {
 		return false;
 	}
--- a/test/sql/optimizer/topn_window_elimination_argument_order.test
+++ b/test/sql/optimizer/topn_window_elimination_argument_order.test
@@ -1,0 +1,36 @@
+# name: test/sql/optimizer/topn_window_elimination_argument_order.test
+# description: Top-N window elimination must preserve ROW_NUMBER argument ordering
+# group: [optimizer]
+
+statement ok
+CREATE TABLE row_number_order (id INTEGER, g INTEGER, p INTEGER);
+
+statement ok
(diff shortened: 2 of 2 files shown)


d-justen (approved):
I think skipping this optimization in the presence of ArgOrders is the reasonable thing to do here. Thanks! LGTM

PR #24350, opened 2026-07-31
Optimizer: propagate statistics for grouped count(*)

### LINKED ISSUE #24116: Grouped HAVING count(*) = 0 scans everything
### What happens?

See repro for details.

### To Reproduce

```sql
-- should emit EMPTY_RESULT directly, since count(*) = 0 is never true here
memory D EXPLAIN ANALYZE
         SELECT i % 1000 AS k
         FROM range(10000000) t(i)
         GROUP BY k
         HAVING count(*) = 0;
╭─ Summary ───────────╮
│ Total Time: 0.0752s │
╰─────────────────────╯
╭───────────────────────────╮
│ Projection   0 rows · 0µs │
│ Filter       0 rows · 0µs │
╰─────────────┬─────────────╯
╭─ Hash Group By ───────────╮
│ Groups: #0                │
│ Aggregates: count_star()  │
│ 1,000 rows         60.0ms │
╰─────────────┒┎────────────╯
╭─ Projection ┚┖────────────╮
│ Projections: k            │
│ 10,000,000 rows    20.0ms │
╰─────────────┒┎────────────╯
╭─ Table Scan ┚┖────────────╮
│ 10,000,000 rows       0µs │
╰───────────────────────────╯
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

alexprabhat99:
Fixes: #24116 

This PR propagates statistics for grouped count() aggregates, allowing the optimizer to infer that `count(*) >= 1` for ordinary groups. As a result, impossible predicates such as HAVING `count(*) = 0`, HAVING `count(*) < 1`, and HAVING `count(*) <= 0` are optimized to an Empty Result, avoiding unnecessary aggregation and table scans. 

The optimization is applied only to plain grouped count() aggregates and excludes ungrouped aggregates, filtered `count(*)`, and empty grouping sets to preserve correctness.


Mytherin:
Thanks for the PR! However, function specific stats propagation should be part of the function as part of a propagation callback, not hard-coded in the statistics propagator itself. Could you move the code to the correct location?

alexprabhat99:
> Thanks for the PR! However, function specific stats propagation should be part of the function as part of a propagation callback, not hard-coded in the statistics propagator itself. Could you move the code to the correct location?

Hi @Mytherin 
Thanks for the feedback! I moved the `count(*)`-specific statistics propagation into the aggregate function's statistics callback and removed the special handling from the generic statistics propagator. The propagator now only passes the generic grouping context needed by the callback.

Does this approach look good to you, or would you prefer a different way of passing the grouping context?

src/include/duckdb/function/aggregate_state.hpp
```diff
@@ -154,13 +154,15 @@ struct AggregateFinalizeData {
...
-	    : bind_data(bind_data_p), child_stats(child_stats_p), node_stats(node_stats_p) {
+	                         optional_ptr<NodeStatistics> node_stats_p, bool groups_are_non_empty_p = false)
+	    : bind_data(bind_data_p), child_stats(child_stats_p), node_stats(node_stats_p),
+	      groups_are_non_empty(groups_are_non_empty_p) {
 	}
 
 	optional_ptr<FunctionData> bind_data;
 	vector<BaseStatistics> &child_stats;
 	optional_ptr<NodeStatistics> node_stats;
+	bool groups_are_non_empty;
```
Mytherin: Instead of passing in this very specific flag maybe we can pass in the `LogicalAggregate &` itself?

PR #24352, opened 2026-07-31
[Bug] Fix issue with worker thread yielding in Async I/O

pdet:
This PR fixes an issue where we could enter a deadlock in a somewhat specific situation. If our worker thread can't be blocked, it would yield and wait for an async thread to do the reads. However, if we don't have async threads available, it could then enter a deadlock of forever waiting. 

The fix is straightforward: if that read has pending io tasks that are not being read by anyone else, that thread will do them instead of yielding.


Mytherin:
Thanks!

PR #24354, opened 2026-07-31
Add a warning when using OFFSET on result set without defined ordering

hannes:
```SQL
CREATE TABLE a AS FROM range(10000);
SET preserve_insertion_order = false;
FROM a OFFSET 10;
```
now throws a warning:

> OFFSET is applied to a query whose row order is not guaranteed - the result may differ between runs. This happens when the query contains order-breaking operators (e.g. hash joins or aggregations), reads from a source that does not preserve order, or when preserve_insertion_order=false. Add an ORDER BY to make the result deterministic.


Mytherin (approved):
Nice! LGTM

PR #24356, opened 2026-07-31
Remove multi-benchmark support and split CTE jobs

smvv:
The log group (needed for multiple benchmarks in one CI job) made it more difficult to find the regression because the log group hid the failing benchmark sometimes.

Most benchmarks were split into different CI jobs, so I also split the recently added CTE benchmarks (`recursive CTE` and `AoC24`) into their own job.


Mytherin:
Thanks!

PR #24358, opened 2026-07-31
Reset batch indexes between recursive CTE iterations

kryonix:
A duckdb-python query combining a replacement scan, a recursive CTE, and a scalar `LIMIT`/`OFFSET` subquery could fail with:

```text
INTERNAL Error: Pipeline batch index - gotten lower batch index 1
(down from previous batch index of 9999999999999)
```

The replacement scan was exposing a lifecycle problem in recursive pipeline reuse. Batch-aware pipeline executors register their current indexes in the pipeline-level batch coordinator. During finalization, they advance to the terminal batch index. The recursive execution introduced by #24031 reuses pipelines and cached or pooled `PipelineExecutor` instances across iterations, but `ResetForReschedule` reset the source and operator state without resetting this shared coordinator. The next iteration therefore restarted its source at batch 1 while the pipeline still contained terminal indexes from the previous iteration.

I now clear the batch coordinator when rescheduling a pipeline whose sink requires partition information. This happens before cached, pooled, or newly constructed executors prepare their local state, allowing the active executors to register again from the pipeline's base batch index. Clearing every old placeholder is important when a broad recursive iteration is followed by a narrower one using fewer executors. Pipelines that do not require partition information avoid the additional lock entirely.

I also moved the existing `ClearSource` cleanup through the same lock-protected helper. Its previous direct access to `batch_indexes` did not use the corresponding mutex.

The regression test includes a synthetic core equivalent of the Python replacement-scan query, a reduced `UNION ALL` case, execution with operator caching disabled, `UNION DISTINCT`, and a broad-to-narrow transition that changes the number of active executors.

### DIFF
--- a/src/include/duckdb/parallel/pipeline.hpp
+++ b/src/include/duckdb/parallel/pipeline.hpp
@@ -221,6 +221,7 @@ private:
 	void ScheduleSequentialTask(shared_ptr<Event> &event);
 	bool LaunchScanTasks(shared_ptr<Event> &event, idx_t max_threads);
 	void ResetSinkAndOperators();
+	void ResetBatchIndexes();
 	shared_ptr<GlobalSourceState> GetSourceState();
 	void SetSourceState(shared_ptr<GlobalSourceState> state);
 	void FinishSourceAndPreventBlocking(ClientContext &context);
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #24359, opened 2026-07-31
Rewrite mode and entropy as multi-stage aggregates

### LINKED ISSUE #16531: MODE() function tries to loads entire data in memory, causing session crash
### What happens?

The MODE() function in DuckDB tries to load the entire dataset into memory, leading to excessive memory usage and session crashes when applied to large datasets.

### To Reproduce

Creating sample data to reproduce the error
```
ATTACH 'duckdb/db';
use db;
SET temp_directory = 'duckdb/temp_table';

CREATE TABLE db.sample_data AS 
WITH numbers AS (FROM range(1, 100_000_000) t(id))
SELECT 
    id,
    (RANDOM() % 3) AS download,   -- Generates 0, 1, or 2 for download usage
    (RANDOM() % 3) AS upload,     -- Generates 0, 1, or 2 for upload usage
    'Application_' || (RANDOM() % 1000) AS app,  -- Generates 1000 unique apps
    LEFT(md5(random()::text), 44) AS key  -- 44-byte unique string with cardinality 2000
FROM numbers;
— It creates 3.9Gb data in persistent disk 
```

now let's create duckdb cli session again with below settings 
```
.timer on
SET memory_limit = '1GB';
SET max_memory = '1GB';
SET enable_progress_bar = true;
SET enable_fsst_vectors=true;
SET preserve_insertion_order = false;

ATTACH 'duckdb/db';
use db;
SET temp_directory = 'duckdb/temp_table';

select app, key, mode(download), mode(upload) from db.sample_data group by app, key;
 
```
It will be killed right away 

I've found WORKAROUND for this behavior 
```
WITH download_counts AS (
    SELECT 
        app, 
        key, 
        SUM(CASE WHEN download = 0 THEN 1 ELSE 0 END) AS low_count,
        SUM(CASE WHEN download = 1 THEN 1 ELSE 0 END) AS average_count,
        SUM(CASE WHEN download = 2 THEN 1 ELSE 0 END) AS high_count
    FROM db.sample_data
    GROUP BY app, key
),
upload_counts AS (
    SELECT 
        app, 
        key, 
        SUM(CASE WHEN upload = 0 THEN 1 ELSE 0 END) AS low_count,
        SUM(CASE WHEN upload = 1 THEN 1 ELSE 0 END) AS average_count,
        SUM(CASE WHEN upload = 2 THEN 1 ELSE 0 END) AS high_count
    FROM db.sample_data
    GROUP BY app, key
)

SELECT 
    d.app, 
    d.key,
    CASE 
        WHEN d.low_count = GREATEST(d.low_count, d.average_count, d.high_count) THEN 0
        WHEN d.average_count = GREATEST(d.low_count, d.average_count, d.high_count) THEN 1
        ELSE 2 
    END AS mode_download,
    CASE 
        WHEN u.low_count = GREATEST(u.low_count, u.average_count, u.high_count) THEN 0
        WHEN u.average_count = GREATEST(u.low_count, u.average_count, u.high_count) THEN 1
        ELSE 2 
    END AS mode_upload
FROM download_counts d
JOIN upload_counts u 
ON d.app = u.app AND d.key = u.key;

```

But it worked because i had categorical column and i knew all possible distinct values. 
Can you fix this issue in general.
 

### OS:

linux

### DuckDB Version:

v1.2.0

### DuckDB Client:

Cli

### Hardware:

I have 24cpu and 48 GB ram and used it on 55GB dataset but it failed

### Full Name:

Suraj Pattewar

### Affiliation:

Student

- [x] Yes, I have

- [x] Yes, I have

kryonix:
`mode` has almost the same execution shape as the logical rewrite for `DISTINCT` aggregates. It currently keeps an `unordered_map` in every aggregate state and updates it from inside the physical aggregate operator. Those maps are not managed by the aggregate hash table or the buffer manager, so they cannot spill and their memory is not controlled by the query's memory limit. With many groups, this is both expensive and capable of consuming enough untracked memory to kill the process, as reported in #16531.

The operation that `mode` performs internally is a frequency aggregation. For example, this query:

```sql
SELECT g, mode(v)
FROM values
GROUP BY g;
```

can be evaluated as two ordinary aggregate stages:

```sql
SELECT g, arg_max(v, frequency)
FROM (
    SELECT g, v, count(*) AS frequency
    FROM values
    GROUP BY g, v
)
GROUP BY g;
```

I implement that rewrite at the logical plan level. This is the root-cause fix for the memory problem: the frequencies are now stored in ordinary aggregate hash tables, which means the existing parallel and out-of-core aggregate implementation can manage them. The optimizer and profiler can also see both stages instead of treating the inner aggregation as opaque aggregate state.

I renamed the `DistinctAggregateRewriter` to `MultiStageAggregateRewriter` and added typed rewrite callbacks to `AggregateFunction`. Direct expression rewrites and logical-plan recipes are separate contracts. A plan recipe is an aggregate DAG: each stage names its input stages, shared outputs use streaming-capable CTE fan-out, and branches can join on the original group keys. This is sufficient for recipes such as MAD, which need to retain frequency rows, compute a median on one branch, and join the median back to those rows. I deliberately made this an aggregate-function contract rather than adding a special `mode` check to the optimizer. The existing ordered `first`, `last`, `arbitrary`, and `any_value` rewrite uses the direct form, while frequency-based aggregates use plan recipes.

`mode` opts into the frequency rewrite and uses `arg_max(value, frequency)` as its final stage. `entropy` used the same per-state frequency-map pattern, so I opted it into the rewrite as well. It now uses two ordinary `sum` aggregates and computes `log2(sum(c)) - sum(c * log2(c)) / sum(c)` in a projection; the internal entropy finalizer is no longer needed. Exact unary `histogram` uses the same count stage, collects `(value, frequency)` rows with `list`, sorts them, and reconstructs the map with `map_from_entries`. Aggregates over the same argument and filter share one count stage. Different arguments, or a mix with regular and `DISTINCT` aggregates, consume a shared default CTE and combine their results afterward. With streaming CTE fan-out, those branches consume the input directly without materializing it.

There are several semantic details behind the basic two-stage plan. I retain aggregate filters in the first-stage grouping keys so groups with no qualifying values still survive and argument, order, and filter expressions keep their original evaluation behavior. The final aggregate combines that filter with the aggregate's NULL policy. `mode(DISTINCT value)` assigns one vote to every distinct value instead of using `count(*)`.

Frequency buckets need to use the same equality as SQL grouping. Several typed maps relied on the standard C++ equality and ordering operators, which treat floating-point NaNs differently from DuckDB. I added shared hash, equality, and ordering policies based on DuckDB's scalar comparison operators and use them in `mode`, exact `histogram`, and the histogram states behind `list_distinct`, `list_unique`, and `list_histogram`. This keeps optimized aggregates, optimizer-disabled execution, windows, and list aggregation consistent for `FLOAT` and `DOUBLE` values containing NaNs.

Ordered `mode` needs separate handling because its order determines the winner when frequencies tie. I create the aggregate sort key before the first stage and retain the minimum sort key for each value. The final stage builds one composite key from descending frequency and ascending retained order, then uses ordinary `arg_min`. This avoids creating a sorted-list state in the final aggregate, which could otherwise exhaust the memory limit for high-cardinality inputs. Unordered ties remain unspecified, as they already are under parallel execution. One timestamp test previously expected a particular winner from three equally frequent values, so I made that query explicitly order its input rather than preserving an accidental implementation detail.

The existing volatile-input staging is shared with the `DISTINCT` rewrite, which prevents CTE fan-out from evaluating volatile group, argument, order, or filter expressions more than once. The pass remaps aggregate bindings through its final projection and uses null-safe joins for grouped branches.

Recipes now declare whether they are mandatory, unconditional, or cost-based. Mandatory lowering runs after binding even when optional optimization is disabled; grouping sets containing a mandatory recipe are expanded first. Optional recipes are gated solely by the optimizer strategy registered with that recipe, and cost-based recipes run after statistics propagation. Eligibility, costing, and plan construction are separate, so a selected callback is invoked once and discarded callback execution cannot allocate table indexes.

`mode`, `entropy`, and exact unary `histogram` use mandatory lowering for ordinary grouped and ungrouped aggregates. The old frequency-map implementations remain for windows, list aggregation, and aggregate state export, but disabling the optimizer no longer restores those untracked states for ordinary aggregation. Multiple independent frequency sets can still require separate hash aggregates and a join, although streaming fan-out avoids rescanning or buffering their common input.

To verify that the rewrite addresses the reported memory behavior, I used a synthetic workload that exercises the same mechanism as #16531 without depending on its data. It creates one million groups, ten rows per group, and two low-cardinality `mode` aggregates with strict majorities:

```sql
SET threads = 4;
SET memory_limit = '128MB';
SET preserve_insertion_order = false;
SET debug_force_external = true;

SELECT count(*), sum(m1), sum(m2)
FROM (
    SELECT i % 1000000 AS g,
           mode(CASE
                    WHEN floor(i / 1000000) < 6 THEN 1
                    ELSE floor(i / 1000000) % 3
                END) AS m1,
           mode(CASE
                    WHEN floor(i / 1000000) < 5 THEN 2
                    ELSE floor(i / 1000000) % 5
                END) AS m2
    FROM range(10000000) t(i)
    GROUP BY g
);
```

I measured a clean reldebug build at `a6b30a6319` on macOS 26.5 arm64, with `core_functions` and `parquet` linked, the settings above, and `/usr/bin/time -l`. After one warmup per path, I ran three interleaved fresh-process samples. Every run returned `(1000000, 1000000, 2000000)`.

| path | real time samples | mean | maximum RSS samples | mean |
|---|---:|---:|---:|---:|
| rewritten | 1.68 s, 1.73 s, 1.75 s | 1.72 s | 190.9 MiB, 190.4 MiB, 192.3 MiB | 191.2 MiB |
| legacy frequency-map implementation | 4.10 s, 4.71 s, 4.02 s | 4.28 s | 2329.5 MiB, 2350.9 MiB, 2329.4 MiB | 2336.6 MiB |

The legacy row was measured before making the lowering mandatory, while the branch still had a temporary strategy switch for comparison. That switch is intentionally no longer part of the final implementation. This is a targeted diagnostic rather than a general performance claim, but it demonstrates the important difference: the rewritten path stays close to the configured memory budget and can use external aggregation, while the legacy per-group maps consume more than 2 GiB outside that budget. `EXPLAIN ANALYZE` for the two-mode query reports `STREAMING_FANOUT`, two direct consumers, zero buffered consumers, and explicit `count_star` and `arg_max` aggregate stages.

Fixes #16531


kryonix:
The CI failure is unrelated.

lnkuiper:

1. `mode(v ORDER BY ...)` still ends in a non-spillable sorted aggregate. The first frequency stage spills, but `CreateFinalAggregate` moves the original `ORDER BY` into `arg_max`, which makes physical planning wrap the final aggregate in `SortedAggregateState`/`ListAggState`. For example, on this branch:

   ```sql
   SET threads=4;
   SET memory_limit='32MB';
   SET preserve_insertion_order=false;
   SET debug_force_external=true;

   SELECT mode(i) FROM range(1000000) t(i);
   -- succeeds

   SELECT mode(i ORDER BY i) FROM range(1000000) t(i);
   -- Out of Memory
   ```

   I think the final choice should be represented entirely in the `arg_min`/`arg_max` key instead. We already retain the minimum encoded order key for every value, so the final aggregate can compare a composite key such as `(frequency DESC, retained_order_key ASC)`, for example with `create_sort_key`. That keeps a fixed-size final aggregate state and preserves the tie-breaking semantics without an aggregate `ORDER BY`.

   The treatment of `DISTINCT` otherwise makes sense to me. `mode(DISTINCT v)` gives every distinct non-NULL value frequency one, so unordered mode may choose any value and an explicit ordering selects the first value in that ordering. Mode is order-dependent only for frequency ties, so retaining one best order key per distinct value is the right model.

2. Entropy is not order-dependent, but both the existing implementation and this rewrite currently retain an aggregate `ORDER BY`:

   ```sql
   SELECT entropy(i) FROM range(1000000) t(i);
   -- succeeds with the settings above

   SELECT entropy(i ORDER BY i) FROM range(1000000) t(i);
   -- Out of Memory
   ```

   Please mark entropy as `AggregateOrderDependent::NOT_ORDER_DEPENDENT`.

3. I do not think we should land `AggregateRewriteInfo` in its current state. It is presented as a generic aggregate-rewrite API, but the optimizer owns a hard-coded frequency plan and it can only customize `NULL` handling and return one final aggregate. I fear it is too specific to build future rewrites on.

   Let's make the contract more generic in this PR. In the future we might want our rewrites to describe:

   - direct aggregate replacements as well as one or more aggregation stages;
   - the grouping keys and intermediate aggregates produced by each stage;
   - multiple final aggregates plus an arbitrary scalar projection over their results;
   - which compatible stages can be shared by several rewritten aggregates;
   - the handling of `FILTER`, `DISTINCT`, ordering, NULLs, and volatile inputs without every function reimplementing it.

   We don't have to add everything in this PR, since this PR tackles the frequency rewrites, but we should think ahead.

   This should also have a deliberate relationship with DuckDB's existing aggregate rewrites. We currently have `AggregateFunctionRewriter` for rewrites such as AVG into multiple aggregates plus a projection, while `OrderedAggregateOptimizer` has a hard-coded name switch for `first`, `last`, `arbitrary`, and `any_value`. Adding a third, narrowly shaped public mechanism will make this harder to generalize later.

4. The existing direct rewrite of `FIRST(v ORDER BY v)` to `arg_min_null(v, create_sort_key(v, ...))` is great and avoids the `LIST`-backed ordered aggregate. It does, however, belong conceptually in a generic aggregate-rewrite facility. The new contract (from the previous point) should be able to express this as a direct, single-stage replacement, so that the function can opt in without another hard-coded function-name list. In other words, `FIRST` should use generic rewrite infrastructure, but not the multi-stage CTE shape.

5. With support for multiple final aggregates and a projection, `__internal_entropy_from_counts` is unnecessary. Entropy can be finalized as:

   ```sql
   COALESCE(
       log2(sum(c)) - sum(c * log2(c)) / sum(c),
       0
   )
   ```

   The two sums can be ordinary final aggregates and the `COALESCE`/arithmetic can live in a projection above them, including the empty/all-filtered case. That is another concrete reason the generic result needs a final projection.

   I would avoid registering a new SQL-callable `__internal_entropy_from_counts` function for this.

6. Please include exact `histogram(v)` in this PR as the second user of the generic frequency rewrite. It has the same frequency decomposition, and adding it now is a useful check that the API is generic rather than mode/entropy-shaped. The result map itself still has to be materialized, so this does not make arbitrarily large histogram results free, but the implementation should use the same relational counting stage.

   The SQL hashing/equality fix should also be generic. Please add something like `sql_value_map_t`, analogous to `column_binding_map_t`, backed by DuckDB's SQL hash and equality operations, and use that from `mode` rather than keeping local `ModeHash`/`ModeEquality` types. The ordered histogram map needs the corresponding SQL ordering comparator as well; changing the global `unordered_map` defaults would be too broad.

   There are already visible NaN bugs in other users of the C++ defaults:

   ```sql
   WITH d(v) AS (
       VALUES ('nan'::DOUBLE), ('nan'::DOUBLE), (1.0), (1.0)
   )
   SELECT histogram(v)::VARCHAR FROM d;
   -- currently: {nan=4}

   SELECT
       list_distinct(['nan'::DOUBLE, 'nan'::DOUBLE, 1.0, 1.0])::VARCHAR,
       list_unique(['nan'::DOUBLE, 'nan'::DOUBLE, 1.0, 1.0]);
   -- currently: [1.0, nan, nan], 3
   ```

   These should have two SQL-distinct values. Please move histogram, `list_distinct`, and `list_unique` to the generic SQL-value map/comparator and add regression tests as part of this change.

7. I think `quantile_disc`, `quantile_cont`, and `median` can also consume `(value, frequency)` pairs, and while `mad` is more difficult, it could also use a rewrite. We should think ahead for rewriting these functions when designing the API.

8. The `static EntropyAggregateRewriteInfo info` is only needed because `aggregate_rewrite_t` returns a `const AggregateRewriteInfo &`; constructing it on the fly would currently return a dangling reference. I would change the API rather than encode singleton lifetime into every aggregate.

kryonix:
The CI failure is unrelated.

lnkuiper:
Thanks, looks really good now!

PR #24360, opened 2026-07-31
[ASYNC I/O] Refactor in preparation for ASYNC in native files

pdet:
This PR is only refactoring, it mainly dedups and cleans up a bit of the existing ASYNC code and splits up ` RowGroup::Scan` and `StandardBufferManager::Prefetch` so we can separate I/O scheduling from decoding in the native storage scan path.

In follow up PR's we will do the on-demand ASYNC-IO and read-ahead.


Mytherin:
Thanks!

PR #24361, opened 2026-07-31
Parallelize ordered buffered CTE scans

kryonix:
The pipeline exchange can preserve order across parallel producers through batch indexes, but an ordered buffered CTE consumer was still forced to a single source task. The buffered scan exposed one flat chunk cursor, so letting several tasks pull from it would reorder chunks before they reached the downstream sink. This was particularly visible when an ordered CTE fed a batch-aware sink such as `CREATE TABLE AS`, `COPY`, or the buffered result collector: the producer could run in parallel, but the exchange-to-sink pipeline became serial.

I changed the buffered exchange so that it can assign complete producer batches to independent scan tasks. A task still reads the chunks within its batch in order, while different tasks can process different batches concurrently. The downstream sink receives the original batch indexes and uses its existing batch-ordering machinery to publish the result in order.

This is selected from the physical pipeline requirements, rather than being attached to a CTE when it is bound or optimized. That matters because CTEs can be introduced during optimization, and because the useful execution strategy depends on the eventual consumer pipeline. During `PhysicalCTE::BuildPipelines`, a buffered consumer uses a batch scan only when all of the following hold:

* the exchange has batch indexes;
* the downstream sink requires batch indexes; and
* the sink does not require partition columns.

All other buffered consumers keep the existing chunk scan. This includes unordered exchanges, sequential exchanges without batch indexes, result paths that do not require batch indexes, and partition-column sinks. Direct and fully materialized CTE consumers are unchanged.

The distinction is visible in `EXPLAIN ANALYZE` for buffered CTE scans:

* `CTE Mode: BUFFERED`, `Scan Mode: BATCH` means that source tasks claim complete exchange batch ranges. The scan can use the exchange's normal parallelism because the sink can reconstruct the required order from batch indexes.
* `CTE Mode: BUFFERED`, `Scan Mode: CHUNK` means that the consumer reads the existing shared chunk stream. If the exchange preserves order, this path remains single-threaded.

I deliberately kept the existing in-memory/spooled chunk stream as the authoritative storage. The batch scan adds a lightweight directory over that stream; it does not copy chunks into a second collection. Each directory entry records the producer and exchange batch index and the begin/end positions in the flat chunk stream. A range is closed when the next batch arrives, when the producer advances its minimum batch index, or when production finishes.

Each local scan can own one range at a time. Consumer-global state records the next unclaimed range, the next unread position of every claimed range, and any spool reads currently outside the exchange lock. Retirement is bounded by all three, so neither an in-memory chunk nor a spooled chunk can be retired while a task can still read it. Idle tasks also publish a batch floor after all currently visible ranges have been claimed; otherwise an idle task could leave a batch-aware sink's global minimum pinned to an old batch indefinitely.

There is one less obvious execution transition here. A dynamic source can finish one batch before data for the next batch is available. Returning an empty data chunk is not sufficient, because the sink still has to close the old local batch and then learn the new global minimum. I added `SourceResultType::BATCH_ADVANCED` for this control-only source result.

The pipeline executor handles that result in two ordered steps. It first runs the existing `NextBatch` path, while the old minimum is still visible. It then updates the local minimum and calls `UpdateMinBatchIndex` on the sink. This ordering is important for batch insert and batch copy: the old batch must be finalized before its last owner disappears from the minimum. Both operations can block and resume independently. The minimum update is also propagated through direct and nested CTE consumers, so a batch boundary is not lost when exchanges are chained.

`PhysicalOperator::UpdateMinBatchIndex` is a no-op by default. The buffered batch collector, batch insert, batch copy, and `PhysicalCTE` override it because they maintain state that depends on the global minimum. This keeps the generic executor transition explicit without requiring unrelated sinks to implement it.

The broadest cost of this change is the new source control result and sink callback in the pipeline executor. Inside the exchange, batch consumers add range and claim metadata plus more retirement bookkeeping. That metadata is only maintained after a batch-scan consumer has been registered, so the existing CHUNK path does not build a parallel batch directory. Spool I/O and callbacks remain outside the exchange mutex, and cancellation, early source termination, reset, and consumer unregister clear outstanding claims before waking blocked readers and writers.

I added coverage for the cases where this protocol is easiest to get wrong: repeated parallel ordered scans, fewer produced batches than scan tasks, early `LIMIT`, mixed direct/buffered consumers with a shared spool, sparse and empty batches, nested buffered exchanges, a single executor, pending-result execution and abandonment, and propagation through a direct nested CTE into ordered Parquet output. The tests verify both query results and the selected scan mode.

For a local performance check, I compared exact main `bf753f906d` with this commit `9b22d512c9` on Apple Silicon/macOS. Both were Release builds with the same linked extensions, both source trees had no tracked modifications, and the query used eight threads. The input consisted of two 10-million-row tables:

```sql
EXPLAIN ANALYZE
CREATE OR REPLACE TABLE out AS
WITH input AS MATERIALIZED (
    SELECT * FROM l
    UNION ALL
    SELECT * FROM r
)
SELECT i, hash(i) AS h
FROM input;
```

With two warmups followed by nine alternating samples, the median total time changed from `0.717s` on main to `0.279s` with the batch scan, a `2.57x` speedup. The main samples ranged from `0.651s` to `0.766s`; the branch samples ranged from `0.265s` to `0.361s`. Both outputs contained exactly 20 million rows, with `min(i) = 0`, `max(i) = 19999999`, and 20 million distinct hashes.

As a protection measurement, I also ran a 100-million-row materialized CTE into a non-batch-aware result path. Its plan remained `Scan Mode: CHUNK`. With three warmups and 31 alternating samples, both main and this branch had a `0.360s` median. The ranges overlapped (`0.241s`-`0.455s` and `0.241s`-`0.460s`), as did the trimmed means (`0.363s` and `0.345s`). I do not read a performance change into those numbers, but they show that the fallback did not inherit the batch directory's metadata or scheduling cost in this workload.


src/parallel/pipeline_executor.cpp
```diff
@@ -352,6 +372,15 @@ PipelineExecuteResult PipelineExecutor::Execute(idx_t max_chunks) {
 				if (source_result == SourceResultType::BLOCKED) {
 					return PipelineExecuteResult::INTERRUPTED;
 				}
+				if (source_result == SourceResultType::BATCH_ADVANCED) {
```
Mytherin: Let's try to move `BATCH_ADVANCED` out of the `SourceResultType` enum because this is not really related to the result processing, instead we can maybe communicate this through some secondary side channel (e.g. `TableFunctionInput`)

src/parallel/pipeline_executor.cpp
```diff
@@ -813,6 +868,25 @@ SinkResultType PipelineExecutor::Sink(DataChunk &chunk, OperatorSinkInput &input
 	return pipeline.sink->Sink(context, chunk, input);
 }
 
+SinkNextBatchType PipelineExecutor::UpdateMinBatchIndex() {
```
Mytherin, line 849: It feels like these should be unified with some of the existing code maybe? Can we try to do that

lnkuiper:
Looks great, thanks!

Mytherin (commented):
Thanks for the PR! Looks good - let's try to clean this up a bit otherwise looks good.

src/parallel/pipeline_executor.cpp
kryonix: I unified the shared mechanics, but kept `UpdateMinBatchIndex` separate because it is a distinct, independently blocking phase: `NextBatch` must finalize the old local batch before the new global minimum is published. Reusing `NextBatch` would finalize sink state twice, so fully unifying these callbacks requires a broader batch-transition API redesign that is better suited to a follow-up PR.

src/parallel/pipeline_executor.cpp
```diff
@@ -792,17 +852,28 @@ SinkResultType PipelineExecutor::Sink(DataChunk &chunk, OperatorSinkInput &input
...
+#ifdef DUCKDB_DEBUG_ASYNC_SINK_SOURCE
+	if (TryDebugBlock(debug_blocked_min_batch_count, interrupt_state)) {
+		return SinkNextBatchType::BLOCKED;
+	}
+#endif
+	OperatorSinkNextBatchInput input {*pipeline.sink->sink_state, *local_sink_state, interrupt_state};
+	return pipeline.sink->UpdateMinBatchIndex(context, input);
+}
+
+SourceResultType PipelineExecutor::FetchFromSource(DataChunk &result, bool &batch_index_advanced) {
```
Mytherin: Maybe this method can return a `struct` instead of taking the `bool &`?

src/parallel/pipeline_executor.cpp
```diff
@@ -792,17 +852,28 @@ SinkResultType PipelineExecutor::Sink(DataChunk &chunk, OperatorSinkInput &input
...
 	D_ASSERT(!pipeline.IsExternalInput());
 	D_ASSERT(global_source_state);
 	D_ASSERT(local_source_state);
 	StartOperator(*pipeline.source);
 
 	OperatorSourceInput source_input = {*global_source_state, *local_source_state, interrupt_state};
 	auto res = GetData(result, source_input);
+	batch_index_advanced = source_input.batch_index_advanced;
```
Mytherin: Can we also make `batch_index_advanced` an enum?

src/parallel/pipeline_executor.cpp
```diff
@@ -354,7 +387,19 @@ PipelineExecuteResult PipelineExecutor::Execute(idx_t max_chunks) {
...
 				// "Regular" path: fetch a chunk from the source and push it through the pipeline
 				source_chunk.Reset();
-				source_result = FetchFromSource(source_chunk);
+				bool batch_index_advanced;
+				source_result = FetchFromSource(source_chunk, batch_index_advanced);
+				if (batch_index_advanced) {
+					D_ASSERT(source_result == SourceResultType::BLOCKED);
+					D_ASSERT(required_partition_info.RequiresBatchIndex());
+					D_ASSERT(!required_partition_info.RequiresPartitionColumns());
+					auto source_data = pipeline.source->GetPartitionData(context, source_chunk, *global_source_state,
```
Mytherin: We're adding this call to `GetPartitionData`, but also have another call down below (in the `if (required_partition_info.AnyRequired()` branch). Could we unify these two code paths (ideally only calling `GetPartitionData` if `FetchFromSource` returns `batch_index_advanced`)?

Mytherin (commented):
Thanks for the updates - looks good, some more comments

Mytherin:
Thanks!

PR #24362, opened 2026-07-31
Unify C-API symbol versioning for clients/extensions, stabilize all v1 API's

Maxxen:
A PR also pushing this to `main` will follow soon.

This PR does three things:

### 1. Unify the C-API symbol versioning/targeting across `duckdb.h` and `duckdb_extension.h`

`duckdb.h` had the concept of "deprecated symbols", but not "unstable" symbols, while `duckdb_extension.h` had the opposite - a way to opt in to "unstable" part of the API, but not to detect "deprecated" symbols.
These now both work the same way in regards to hiding/opting in to "unstable"/"deprecated" symbols , and can now select a view of the API _as of_ a specific version by defining `DUCKDB_API_VERSION_MAJOR` / `_MINOR` / `_PATCH` before including `duckdb.h`.

### 2. Backfill the version history for every C-API symbol, and use it in comments/version `#ifdef` gates

This was necessary to be able to implement 1. Basically each API symbol now has their complete "lifecycle" documented in the spec, which is emitted as part of the documentation string in the header - but is also used to hide/show the symbol based on if the version macros are set or not, e.g.

```c
#if DUCKDB_API_VERSION_AT_LEAST(0, 2, 9) && (DUCKDB_API_VERSION_BELOW(1, 0, 0) || DUCKDB_API_ALLOW_DEPRECATED)
/*!
 *
 * Returns the number of rows present in the result object.
 *
 * history:
 * - stable: v0.2.9
 * - deprecated: v1.0.0
 *
 * @param result The result object.
 * @return idx_t
 */
DUCKDB_C_API idx_t duckdb_row_count(duckdb_result *result);
#endif
```

This turned out to also surface some bugs/accidental breaks in the API, where e.g. we renamed `varint`->`bignum`. But the cool thing with having the consumers of the headers be able to control which "version" they want to see is that we can now "fix" those by including macros/typedefs from the old name to the new when targeting a version before the rename occurred.

You can read more about how this works in the new `VERSIONING.md`

### 3. Stabilize all currently unstable v1 C-API symbols into a yet-to-be-released "v1.5.6" version.

This was the actual goal that triggered all this work. We want to soon release the "V2" C-API, with a different "entrypoint" function. A `unstable` extension therefore has to imply "V2" as that is the _next_ api version. But that would break every  client/c-api extension that mark themselves as targeting `unstable` - but (rightfully) expect the `V1` ABI entrypoint because they want to use all the unstable symbols in the V1 API. 

We could of course just remove the unstable V1 part - it is unstable after all - but that will break basically every client library using the C-API, because as mentioned in 1. there is no concept of "unstable" in `duckdb.h` 🙃. So these symbols have essentially been considered stable (from the perspective of client libraries/embedders of duckdb) regardless.

Additionally, we have not released another "stable" C-API extension version since `v1.2.0`, and almost half the API is currently marked unstable.

Since we are going to deprecate the V1 API as soon as DuckDB v2.0.0 is released anyway - it seems like the better option is to just stabilize everything in V1 as the "final" version of V1, and then immediately deprecate it. This will allow existing C-API extensions to keep using the unstable (now stable, deprecated) api in DuckDB v2.0.0, giving them plenty of time to migrate before it is eventually removed in v3.0.0. And "unstable" will then imply an extension based on the V2 API.

# What does this mean for me as an extension/client maintainer using the C-API?

## You consume `duckdb.h` (client, binding, embedder)

Nothing changes and nothing is required. Visible symbols are identical in every existing configuration:
```
┌────────────────────────────┬────────┬───────┐
│         Your build         │ Before │ After │
├────────────────────────────┼────────┼───────┤
│ default                    │ 543    │ 543   │
├────────────────────────────┼────────┼───────┤
│ -DDUCKDB_API_NO_DEPRECATED │ 495    │ 495   │
└────────────────────────────┴────────┴───────┘
```
`DUCKDB_API_NO_DEPRECATED` keeps working, but it now sets the new `DUCKDB_API_ALLOW_DEPRECATED` switch.

What you gain, opt-in:

- Target a specific version: define `DUCKDB_API_VERSION_{MAJOR,MINOR,PATCH}` (all three or none) to compile against an older surface. This is kinda like the way `_POSIX_C_SOURCE` works if you're familiar with that.
- Deprecation is now __relative to your target__. Previously `NO_DEPRECATED` removed everything ever deprecated; now pinning e..g `v1.5.4` keeps symbols that were deprecated in `v1.5.6`.
- `DUCKDB_API_VERSION_AT_LEAST(x,y,z)` / `_BELOW(x,y,z)` for conditional code.
- Every symbol documents its history in the generated comment, so you can see when something appeared (or got deprecated) without digging through git.

## You maintain an extension on the unstable API

This is the change aimed at you. If you use `build_loadable_extension_capi_unstable`, (the default for most C extensions), you are currently locked 1:1 to a single DuckDB build and must rebuild and republish for every release.

The 189 functions that forced you there are now stable. Once v1.5.6 ships you can switch to:

```cmake
build_loadable_extension_capi(my_ext 1 5 6 ${SOURCES})
```

This will give you access to all previously unstable functions and allow your extension to be loaded by v1.5.6 and every later release. The only reason to stay on the unstable mode is if you use something still unstable, which as of today is nothing (but will include imply V2 API once that gets pushed to main).

## You maintain an extension on the versioned API (v1.2.0)

Your existing build keeps working: the 357 v-table layout you compiled against is an exact prefix of the new one.

What's new is that the pin now does something. Pin the lowest version whose surface you need. Today that means `v1.2.0` (404 slots) or `v1.5.6` (546); everything in between yields the same 404 slots while needlessly excluding older engines, because nothing was stabilized between those two releases.

If you pinned an old version and used e.g. `duckdb_create_varint` / `duckdb_get_varint` / `duckdb_varint`, those spellings now work again below `v1.4.0`.

## You've hacked around with your own CMake config

No changes should be required. Both CMake helpers work as-is, except for one break: passing `-DDUCKDB_API_VERSION_*` directly while including `duckdb_extension.h` now errors, because for an extension the compiled surface and the negotiated ABI must agree. Use `DUCKDB_EXTENSION_API_VERSION_*` instead, like the error tells you to :)

## You're going to contribute C API functions

Please don't add anything more to the V1 C-API :)

Also, two new rules to remember:

- A functions v-table position freezes on stabilization, __not__ introduction. While a function is unstable it is only seen by builds locked to one engine, so it can still be reordered or resignatured. Once its version ships, its position is permanent -so get it right before the release!
- Prefer deprecation to removal. Deprecation is version-relative and costs nothing at runtime: the vtable stays filled, old binaries keep working, new builds get a compile error. Removal may null the function pointer, which will crash extensions that access it and were built before the removal.


Maxxen:
Okay, two things to think some more about before merging this:
- ~~Including `duckdb.h` should maybe have the same semantics as extensions when it comes to "unstable" symbols, i.e. you can only access unstable symbols if you target the latest version.~~ This is now done.
- ~~Similarly deprecated symbols should maybe not be version relative...~~ They aren't, only their visibility is version-gated, but the warnings (are not emitted now) but can be toggled on and are always relative to the _current_ version (i.e. absolute).

Mytherin:
Thanks!

PR #24364, opened 2026-07-31
Restructure `DUCKDB_SMALLER_BINARY` into a preprocessor macro that allows granular attribution

carlopi:
This adds some contraints while adding DUCKDB_SMALLER_BINARY features, but allows to test with an arbitrary subset of features ON or OFF, and allows to track which parts of the code are connected and are meant to logically be together.

User side this can be tweaked in a few different ways:
```
make
SMALLER_BINARY=1 make
SMALLER_BINARY='unary_executor_flat,binary_executor_flat' make
SMALLER_BINARY='vector_specialization' make
SMALLER_BINARY=1 SMALLER_BINARY_EXCLUDE='binary_executor_select_flags' make
```

That respectivelly do:
* regular build
* SMALLER_BINARY build (all features disabled)
* only selected smaller_binary features
* only selected group of smaller_binary features
* enable all SMALLER_BINARY optimization, but for the named one

Combining SMALLER_BINARY and SMALLER_BINARY_EXCLUDE allows to select an arbitrary set of build time optimization that improve binary size (at the cost of performances). This allows both to test features in isolation, allows to tie different logically connected areas of the code, and allows to customize builds.

Bumped into this while looking to add a couple of more SMALLER_BINARY codepaths for reducing duckdb-wasm sizes, and realized currently it's all or nothing, and before adding more I think allowing easier tracking of what part of the codebase are logically tie together would make sense.


Mytherin:
Thanks!

PR #24365, opened 2026-07-31
Fix false corruption error when reading FSST string segments for Clickbench Q11

gropaul:
https://github.com/duckdb/duckdb/pull/24016 added that before reading the symbol table, we check whether there is still `sizeof(duckdb_fsst_decoder_t)` (2304 bytes) of room in the segment. This was added to avoid out-of-bounds reads if the segment has been corrupted. 

However, the on-disk symbol table is a variable-length serialization that is usually smaller. Valid segments with a small symbol table (for example, a mostly-empty column) were wrongly rejected with `Data Corruption Error`. 

This, for example, led to a `DataCorruptionException` when running Clickbench Q11 on a non-corrupted, v1.0 database.

The proposed fix is to modify `duckdb_fsst_import` to include information about the buffer size, so that the out-of-bounds checks can happen internally.  

I also had to update the return of the import before. For segments with strings of length 0, DuckDB allows a  `duckdb_fsst_import` even if there is no symbol table, which is discovered by a version missmatch. Therefore, I the errors `DUCKDB_FSST_IMPORT_VERSION_MISMATCH` and `DUCKDB_FSST_IMPORT_OUT_OF_BOUNDS` to only throw on the 2nd one.

`fsst_small_symbol_table.test` reads back an FSST segment with a serialized symbol table that is smaller than `sizeof(duckdb_fsst_decoder_t)`, which allows reproducing the crash described above.

### DIFF
--- a/src/storage/compression/dict_fsst/compression.cpp
+++ b/src/storage/compression/dict_fsst/compression.cpp
@@ -584,7 +584,8 @@ DictFSSTCompressResult DictFSSTCompressionState::CompressInternal(UnifiedVectorF
 		vector<unsigned char> decompress_buffer;
 		if (verify_compression) {
 			temp_decoder = &temp_decoder_storage;
-			duckdb_fsst_import(&temp_decoder_storage, fsst_serialized_symbol_table.get());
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks!

PR #24367, opened 2026-07-31
Fix wrong results and an out-of-bounds read in the LIST->ARRAY cast

carlopi:
The fast path cast the first array_size * count entries of the source list child, which has nothing to do with the entries the lists point at. A filter leaves the child holding every entry it had before, so a surviving row's entries can live past that window, and the cast then copied never-written payload memory into the result:

```sql
CREATE TABLE t AS
    SELECT i, [(i*3+1)::VARCHAR, (i*3+2)::VARCHAR, (i*3+3)::VARCHAR] AS l
    FROM range(100) t(i);
SELECT l::INT[3] FROM t WHERE i = 99;   -- [0, 0, 0], want [298, 299, 300]
```

In the other direction, a child smaller than that window was read past its end. At STANDARD_VECTOR_SIZE=8 that threw "Vector::SetSize out of range" from the Arrow scan of a fixed-size ARRAY(dict int32) with a nonzero array offset.

Cast the entries the selection references instead. Casting the child at its full size fixes both, but then raises conversion errors from entries no row references.

Found while making a sweep looking for errors at STANDARD_VECTOR_SIZE=8.

### DIFF
--- a/src/function/cast/list_casts.cpp
+++ b/src/function/cast/list_casts.cpp
@@ -225,16 +225,19 @@ static bool ListToArrayCast(Vector &source, Vector &result, idx_t count, CastPar
 		// We can just cast the child vector directly
 		// Note: Its worth doing a CheckAllValid here, the slow path is significantly more expensive
 		if (FlatVector::ValidityMutable(result).CheckAllValid(count)) {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24368, opened 2026-07-31
Fix two wrong-results bugs in the expression rewriter

carlopi:
Both found by generating ~1000 <function> <op> <constant> predicates over a corpus containing NULL, NaN, -0.0 and the infinities, evaluating each as a PROJECTION so the third truth value is visible - inside a WHERE, NULL and FALSE are indistinguishable - and diffing against the same run with each of the 43 optimizers disabled in turn.

1. MoveUnaryMinusRule dropped rows on NaN. It rewrites [-x OP c] to [x FLIPPED_OP -c] and guarded that with AND NOT isnan(x) for every ordered comparison. -NaN is still NaN and DuckDB orders NaN above everything, so [-x > c] and [-x >= c] are TRUE for a NaN input while the flipped [x < -c] / [x <= -c] are FALSE. Those two directions need OR isnan(x).

```sql
-- {nan, 1.0, -5.0}
SELECT count(*) FROM t WHERE (-x) > 0;   -- 1, should be 2
```

   The guard polarity is chosen from the comparison as seen by the negation, so [c > -x] is read as [-x < c] first; getting that orientation wrong is what test/optimizer/move_constants.test catches.

2. DateTruncSimplificationRule folded NULL to FALSE. An unsatisfiable date_trunc comparison folded to a bare TRUE/FALSE constant. That is correct for the NULL-safe IS [NOT] DISTINCT FROM forms, but plain = and <> must keep propagating NULL, so a NULL input silently became FALSE (resp. TRUE). Now uses ExpressionRewriter::ConstantOrNull for the null-propagating variants.

rewrite_nan_null.test pins both across the relevant optimizers and in both the projection and filter forms. A/B verified: reverting the two hunks makes it fail.

### DIFF
--- a/src/optimizer/rule/move_constants.cpp
+++ b/src/optimizer/rule/move_constants.cpp
@@ -201,7 +201,7 @@ static bool CanDuplicateForNaNGuard(const Expression &expr) {
 	       expr.GetExpressionClass() == ExpressionClass::BOUND_REF;
 }
 
-static unique_ptr<Expression> CreateNotIsNanGuard(ClientContext &context, const Expression &expr) {
+static unique_ptr<Expression> CreateIsNanCall(ClientContext &context, const Expression &expr) {
 	vector<unique_ptr<Expression>> children;
 	children.push_back(expr.Copy());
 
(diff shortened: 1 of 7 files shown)


src/optimizer/rule/move_constants.cpp
```diff
@@ -326,8 +347,9 @@ unique_ptr<Expression> MoveUnaryMinusRule::Apply(LogicalOperator &op, vector<ref
 	}
 	BoundComparisonExpression::FlipType(comparison);
 	if (nan_guard) {
-		auto result = make_uniq<BoundConjunctionExpression>(ExpressionType::CONJUNCTION_AND, comparison.Copy(),
-		                                                    std::move(nan_guard));
+		auto conjunction_type =
+		    nan_guard_disjunctive ? ExpressionType::CONJUNCTION_OR : ExpressionType::CONJUNCTION_AND;
+		auto result = make_uniq<BoundConjunctionExpression>(conjunction_type, comparison.Copy(), std::move(nan_guard));
```
Mytherin, line 352: I kind of wonder if this is now not turning into a performance degradation since we're adding complexity in this case. We should consider either (1) just not doing this optimization for floating points, or (2) have `isnan` implement stats propagation so that we can optimize it away if the floating point values we're operating on don't have nan values, so that this can still turn into `x > -c` in the common case where floats don't have nan values.

Mytherin (commented):

src/optimizer/rule/move_constants.cpp
carlopi: I am not sure, this code was recently touched by @dentiny and @lnkuiper, maybe I'll leave to them to make a call.

I think short term the options are: either this PR as is, or disabling whole optimization for floating points (your option 1). Removing is simpler, probably safer path before proper review of this code. I think stats propagation can be done on top, but seems more complex / I would not commit to have it in the short term.

carlopi:
@lnkuiper: Asked for a review in particular if you can take a look at @Mytherin comment and make a call. I am happy to adapt either way, or you are welcome to fixup the PR yourself.

Note that the other bug should be more ready to be merged, I can also split them in two separate PRs.

lnkuiper:
Thanks for finding this, I've added statistics propagation to `isnan`, should be ready to go now

PR #24369, opened 2026-07-31
Fix INTERNAL error from EXPLAIN (FORMAT) and EXPLAIN (FORMAT default)

carlopi:
The EXPLAIN transformer read option.children[0] unconditionally for the FORMAT option, but the grammar (ExplainOption <- ExplainOptionName Expression?) leaves children empty whenever FORMAT has no argument at all, or an argument the transformer stored as an expression rather than a constant. Both cases raised 'INTERNAL Error: Attempted to access index 0 within vector of size 0':

```sql
EXPLAIN (FORMAT) SELECT 42;
EXPLAIN (FORMAT default) SELECT 42;
```

The second is the more awkward one: 'default' is a real format name, and it is the first entry in the list the "is not a valid FORMAT argument" message prints, so copying it from the error message crashed. A bare DEFAULT parses as a DefaultExpression rather than an identifier, which is why it never reached children.

FORMAT with no usable argument is now an InvalidInputException, and a bare DEFAULT selects the default renderer.

Adds coverage/explain_formats.test, which drives all ten format names plus EXPLAIN ANALYZE through a multi-level plan. The published fast-suite coverage report has the non-text renderers at 44-68% (graphviz 50.0%, yaml 44.8%, mermaid 56.4%, json 48.8%) because almost nothing asks for them.

### DIFF
--- a/src/parser/peg/transformer/transform_explain.cpp
+++ b/src/parser/peg/transformer/transform_explain.cpp
@@ -26,7 +26,17 @@ unique_ptr<SQLStatement> PEGTransformerFactory::TransformExplainStatement(
 				if (format_is_set) {
 					throw InvalidInputException("FORMAT can not be provided more than once");
 				}
-				format = ParseProfilerPrintFormat(option.children[0]);
+				if (option.children.empty()) {
+					// no constant/identifier argument: either FORMAT was given nothing at all, or its argument is an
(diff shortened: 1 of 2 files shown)


carlopi:
This is very minor but I think ready.

Value is: adding explicit coverage and removing a somewhat remote case of triggering `INTERNAL Error: Attempted to access index 0 within vector of size 0`

Mytherin:
Thanks!

PR #24370, opened 2026-07-31
Bugfix connected to round-tripping identifiers

carlopi:
This started while polishing `EXTERNAL RESOURCES`, then found a bunch of similar cases elsewhere.

### DIFF
--- a/src/function/macro_function.cpp
+++ b/src/function/macro_function.cpp
@@ -1,5 +1,7 @@
 #include "duckdb/function/macro_function.hpp"
 
+#include "duckdb/common/sql_identifier.hpp"
+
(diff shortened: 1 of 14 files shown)


lnkuiper:

1. Quoted `COPY` format names still do not round-trip because `format` is emitted raw:

   ```sql
   SET debug_verify_statement='REPARSE_STATEMENT';
   COPY (SELECT 1) TO 'out.csv' (FORMAT "odd format");
   ```

   This currently fails while reparsing.

2. The enum escaping fix already landed in #24382. Could you rebase and drop this?

Mytherin:
Thanks!

PR #24371, opened 2026-07-31
EXTERNAL RESOURCES: Unskip discovery tests now that the dependency landed

carlopi:
Both files carry `mode skip "discover not yet landed"`, added when #24056 (the SQL surface, part 3/4) merged ahead of the discovery part. That part landed in #24024 on 2026-07-28, and the skip comment says outright "Skip until that lands (then remove this mode skip / mode unskip)".


Mytherin:
Thanks!

PR #24372, opened 2026-07-31
Un-skip 75 stale entries across 15 test configs

carlopi:
The 54 test configs carry 1853 skip_tests entries. 231 of them have a reason that admits ignorance - TODO, Unknown failure, FIXME: crash., FIXME: binding bug., FIXME: Wrong result in query. Re-running the 191 non-slow ones with skip_tests stripped shows 75 now pass, 116 still fail and 20 name a test file that no longer exists.

Here I iterated on the ones that looks to be now supporting, and confirmed passing with the config's skip list stripped, then re-confirmed against the edited config itself.
Let's clean this list, that has a tendency to rot, and let's see what CI thinks (CI passing is the main oracle here).

### DIFF
--- a/test/configs/compressed_in_memory.json
+++ b/test/configs/compressed_in_memory.json
@@ -9,7 +9,6 @@
       "reason": "Unknown, to be investigated",
       "paths": [
         "test/fuzzer/sqlsmith/current_schemas_null.test",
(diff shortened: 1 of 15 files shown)


Mytherin:
Thanks!

PR #24375, opened 2026-08-01
Make `preserve_identifier_case` a three-state setting

### LINKED ISSUE #24733: Table with single quotes inside a double-quoted name can be created but never referenced
### What happens?

In `2.0`'s new PEG parser that will replace `1.0`'s vendored Postgres parser, a table whose double-quoted name contains single quotes can be created, but can never be referenced again — `INSERT`, `SELECT` and every other access path fail. The catalog error even suggests the exact syntax that does not work.

This is a regression against `v1.5.x`: it works correctly there, but fails on `main`.

This was incidentally patched in #24375 and a regression test was added there for posterity.

### To Reproduce

```sql
CREATE TABLE "'FOO'"(i INT);
SELECT table_name FROM duckdb_tables;
-- 'FOO'      <- created fine

INSERT INTO "'FOO'" VALUES (7);
-- Catalog Error: Table with name FOO does not exist!
-- Did you mean "'FOO'"?
```

Every access form fails the same way:

```sql
SELECT * FROM "'FOO'";   -- Table with name FOO does not exist!
SELECT * FROM 'FOO';     -- Table with name FOO does not exist!
SELECT * FROM FOO;       -- Table with name FOO does not exist!
```

Lower case behaves identically (`"'foo'"` → looks for `foo`), so this is not related to identifier case folding.

### Expected

The table should be reachable, as it is on v1.5.4:

```
$ duckdb -c "CREATE TABLE \"'FOO'\"(i INT);" -c "INSERT INTO \"'FOO'\" VALUES (7);" -c "SELECT * FROM \"'FOO'\";"
7
```

### Cause

In `IdentifierMatcher::MatchParseResultInternal` (`src/parser/peg/matcher.cpp`), the single-quote unwrapping runs *after* the double-quote branch has already unwrapped the token, so a token like `"'FOO'"` is stripped twice:

```cpp
if (IsQuoted(result_text)) {
    result_text = result_text.substr(1, result_text.size() - 2);   // "'FOO'"  ->  'FOO'
    ...
}
if (IsSingleQuoted(result_text) && SupportsStringLiteral()) {
    result_text = result_text.substr(1, result_text.size() - 2);   // 'FOO'    ->  FOO
    ...
}
```

`SupportsStringLiteral()` is only true for `SUGGEST_TABLE_NAME` and `SUGGEST_FILE_NAME`, which is why `CREATE TABLE` records the name correctly while `FROM` / `INSERT INTO` look up the twice-stripped one.

Making the two unwrappings mutually exclusive fixes it.

### Environment

- `main` @ `0a05451f85` — reproduces
- v1.5.4 (released, vendored Postgres parser) — works
- macOS arm64

dataders:
Adds `uppercase` identifier folding by making `preserve_identifier_case` three-state, rather than pairing the existing boolean with a second one.

| `preserve_identifier_case` | unquoted `MySchema.Integers` becomes |
|---|---|
| `preserve_case` (default) | `MySchema.Integers` |
| `lowercase` | `myschema.integers` |
| `uppercase` | `MYSCHEMA.INTEGERS` |

Quoted identifiers are preserved exactly in every mode. The setting still accepts the old BOOLEAN spellings (`true`/`1`/`t`/`y`/`yes` and `false`/`0`/`f`/`n`/`no`), mapped onto `preserve_case`/`lowercase`, so no existing configuration breaks.

### Commits

1. **Make `preserve_identifier_case` a three-state setting** — introduces `IdentifierCaseMode` (on/lowercase/uppercase) in place of the BOOLEAN. Based on @MartinNeupauer's `capitalize_identifier` prototype, credited as co-author here.
2. **Only case-fold tokens that are unfolded identifiers** — fixes two pre-existing bugs that `uppercase` makes visible: a single-quoted path literal in a table position was folded like an identifier (`FROM 'Mixed/Case/File.csv'` looked for the wrong file), and `debug_verify_statement='reparse_statement'` re-folded identifiers that were already quoted. Also fixes #24733 (`CREATE TABLE "'FOO'"` is reachable again).
3. **Rename the preserved-case enum value and simplify lowercase folding** — renames `IdentifierCaseMode::ON` to `PRESERVE_CASE`; folds lowercase via `StringUtil::Lower` instead of a hand-written loop.
4. **Fold to upper case without depending on the locale** — `StringUtil::Upper` now folds through `CharacterToUpper`, matching `Lower`, instead of the locale-dependent `std::toupper` (which can mangle ASCII or produce invalid UTF-8 depending on the host process's locale).
5. **Accept only the three mode names** — drops the transitional alias spellings so the setting takes exactly `preserve_case`, `lowercase`, or `uppercase`.
6. **Keep accepting legacy BOOLEAN spellings for `preserve_identifier_case`** — per review feedback, restores backward compatibility on top of (5): the input is cast through the BOOLEAN cast and mapped onto `preserve_case`/`lowercase`.

### Breaking

`duckdb_settings()` now reports the default as `preserve_case` with `input_type` `VARCHAR`, rather than `true`/BOOLEAN. `Settings::Get<PreserveIdentifierCaseSetting>()` returns `IdentifierCaseMode` instead of `bool`, so out-of-tree extensions reading it need a recompile. `SET preserve_identifier_case = true/false` (and the other boolean spellings) keep working.

Resolves #24733.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/include/duckdb/common/enums/identifier_case_mode.hpp
+++ b/src/include/duckdb/common/enums/identifier_case_mode.hpp
@@ -1,0 +1,18 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// duckdb/common/enums/identifier_case_mode.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/constants.hpp"
+
+namespace duckdb {
+
+//! How non-quoted identifiers are folded by the parser
+enum class IdentifierCaseMode : uint8_t { ON, LOWERCASE, UPPERCASE };
+
+} // namespace duckdb
--- a/src/include/duckdb/common/enums/identifier_case_mode.hpp
+++ b/src/include/duckdb/common/enums/identifier_case_mode.hpp
@@ -13,6 +13,6 @@
 namespace duckdb {
 
 //! How non-quoted identifiers are folded by the parser
-enum class IdentifierCaseMode : uint8_t { ON, LOWERCASE, UPPERCASE };
+enum class IdentifierCaseMode : uint8_t { PRESERVE_CASE, LOWERCASE, UPPERCASE };
 
 } // namespace duckdb
--- a/src/include/duckdb/parser/peg/matcher.hpp
+++ b/src/include/duckdb/parser/peg/matcher.hpp
@@ -155,15 +155,21 @@ struct MatchState {
 		return max_token_index;
 	}
 
-	//! Fold a non-quoted identifier according to the configured case mode
-	string FoldIdentifier(const string &text) const {
+	//! Fold a non-quoted identifier in-place according to the configured case mode
+	void FoldIdentifier(string &text) const {
 		switch (identifier_case_mode) {
 		case IdentifierCaseMode::LOWERCASE:
-			return StringUtil::Lower(text);
+			for (auto &c : text) {
+				c = StringUtil::CharacterToLower(c);
+			}
+			break;
 		case IdentifierCaseMode::UPPERCASE:
-			return StringUtil::Upper(text);
+			for (auto &c : text) {
+				c = StringUtil::CharacterToUpper(c);
+			}
+			break;
 		default:
-			return text;
+			break;
 		}
 	}
 
--- a/src/include/duckdb/parser/peg/matcher.hpp
+++ b/src/include/duckdb/parser/peg/matcher.hpp
@@ -150,11 +150,7 @@ struct MatchState {
 			text = StringUtil::Lower(text);
 			break;
 		case IdentifierCaseMode::UPPERCASE:
-			// not StringUtil::Upper - that one folds through std::toupper, which is locale dependent,
-			// whereas StringUtil::Lower and CharacterToUpper are not
-			for (auto &c : text) {
-				c = StringUtil::CharacterToUpper(c);
-			}
+			text = StringUtil::Upper(text);
 			break;
 		default:
 			break;
--- a/src/include/duckdb/parser/peg/matcher.hpp
+++ b/src/include/duckdb/parser/peg/matcher.hpp
@@ -12,6 +12,7 @@
 #include "duckdb/common/identifier.hpp"
 #include "duckdb/common/vector.hpp"
 #include "duckdb/common/reference_map.hpp"
+#include "duckdb/common/enums/identifier_case_mode.hpp"
 #include "duckdb/parser/parser_extension.hpp"
 #include "duckdb/parser/peg/keyword_helper.hpp"
 #include "duckdb/parser/token_iterator.hpp"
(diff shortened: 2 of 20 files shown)


src/include/duckdb/parser/peg/matcher.hpp
```diff
@@ -154,6 +155,18 @@ struct MatchState {
 		return max_token_index;
 	}
 
+	//! Fold a non-quoted identifier according to the configured case mode
+	string FoldIdentifier(const string &text) const {
+		switch (identifier_case_mode) {
```
Mytherin, line 148: This is always doing string copies even if `IdentifierCaseMode` is not set - can we make this take a `string &text` instead and do the modification in place when case mode is `LOWERCASE` or `UPPERCASE`?

Mytherin (commented):
Thanks for the PR! Looks good to me. One comment

src/include/duckdb/parser/peg/matcher.hpp
dataders: done in d5a05d3ff9b0f503a2ede383c30d5f6478f68cb7!

dataders:
thanks for the review @Mytherin!

we originally opened #24200 before we realized the postgres parser was dropped.

would y'all be amenable to merging #24200 with the temporary `capitalize_identifier` param and releasing it as a 1.5 patch, that then gets dropped for `2.0` when it's released later this year?

we've also got duckdb/duckdb-iceberg/pull/1222 which would complement this work.

cheers!

Mytherin:
Thanks for the updates - since this is a feature this will only be able to make it in the next feature release, so we can only push it into `main`.

Mytherin:
CI seems to be failing - could you have a look?

dataders:
> CI seems to be failing - could you have a look?

yeah looks like there was a latent bug that introducing folding identifiers to upper exposed.

The second commit of this rebased PR addresses this that should fix the failing tests. LMK how it looks to you.

src/include/duckdb/common/enums/identifier_case_mode.hpp
```diff
@@ -0,0 +1,18 @@
...
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/constants.hpp"
+
+namespace duckdb {
+
+//! How non-quoted identifiers are folded by the parser
+enum class IdentifierCaseMode : uint8_t { ON, LOWERCASE, UPPERCASE };
```
Mytherin: `ON` is a bit of an odd name here - can we call this `PRESERVE_CASE`

src/include/duckdb/parser/peg/matcher.hpp
```diff
@@ -154,6 +155,24 @@ struct MatchState {
 		return max_token_index;
 	}
 
+	//! Fold a non-quoted identifier in-place according to the configured case mode
+	void FoldIdentifier(string &text) const {
+		switch (identifier_case_mode) {
+		case IdentifierCaseMode::LOWERCASE:
+			for (auto &c : text) {
+				c = StringUtil::CharacterToLower(c);
```
Mytherin: Can this not be `text = StringUtil::Lower(text)`? Do we need to do the loop here?

src/include/duckdb/parser/peg/matcher.hpp
```diff
@@ -154,6 +155,24 @@ struct MatchState {
...
+			}
+			break;
+		case IdentifierCaseMode::UPPERCASE:
+			for (auto &c : text) {
```
Mytherin: Same here

dataders: I think I've addressed this!

src/include/duckdb/parser/peg/matcher.hpp
dataders: done!

dataders:
hey @Mytherin thanks for all the reviews so far (and those still to come 😺 ).

the changes since your last review center around these things:
1. removing any aliasing of `1.x`'s boolean values (`true `, `false`) for `preserve_identifier_case` for backcompat purposes
2. feedback from @MartinNeupauer's review of some of the resulting code
3. discovering #24733, a regression in the PEG parser, that refactoring for legibility incidentally solves 🤷🏻 

Please let me know how this looks to you! I'm happy to:
1. introduce any/all backcompat that you see fit
2. split this into smaller complementary PRs for easier review.

cheers

lnkuiper:
Looks good to me, just one question for @Mytherin do you think we should have backwards compatible:
```
true → preserve_case, false → lowercase
```
Or is that not necessary?

Mytherin:
Yes I think we should stay backwards compatible there.

dataders:
hey @lnkuiper @Mytherin, I told Claude to bring back the full backwards compatibility which not only maps `true → preserve_case` and `false → lowercase` but also the other boolean combos that are currently valid `1/0`, and `yes/no`.

Hopefully this is what you meant! LMK. Cheers

lnkuiper:
Yes, that's exactly it! Good to go now, just needs to fix some failing CI :)

PR #24384, opened 2026-08-01
Propagate parquet min/max len if possible

### LINKED ISSUE #24383: Parquet string min/max is not leveraged to prune row group
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t_strlen AS
         SELECT repeat('x', CASE WHEN i < 2048 THEN 4 WHEN i < 4096 THEN 20 ELSE 40 END) AS s, i::INTEGER AS i
         FROM range(6144) t(i);
memory D COPY t_strlen TO '/tmp/codex_strlen_prune.parquet' (FORMAT PARQUET, ROW_GROUP_SIZE 2048);

memory D EXPLAIN ANALYZE
         SELECT count(*)
         FROM read_parquet('/tmp/codex_strlen_prune.parquet')
         WHERE strlen(s) > 30;
╭─ Summary ───────────╮
│ Total Time: 0.0010s │
│ Data Read:  16.5 kB │
╰─────────────────────╯
│ Function: READ_PARQUET          │
│ Filters: strlen(s) > 30         │
│ Total Files Read: 1             │
│ Filename(s):                    │
│ /tmp/codex_strlen_prune.parquet │
│ 2,048 rows                  0µs │
```
Parquet, unlike DuckDB native storage, doesn't emit stats on min/max string length, but for the above case where min == max, we're able to deduce min/max length and prune row group.

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24383

Parquet, unlike DuckDB native storage, doesn't emit stats on min/max string length, but for the above case where min == max, we're able to deduce min/max length and prune row group.


extension/parquet/parquet_statistics.cpp
```diff
@@ -512,17 +512,37 @@ ParquetStatisticsUtils::TransformParquetStatistics(const LogicalType &type, cons
...
 		} else if (parquet_stats.__isset.max && StringStatsAreValid(parquet_stats.max, is_varchar, max_stats_type)) {
 			StringStats::SetMax(string_stats, parquet_stats.max, max_stats_type);
+			max_stats = parquet_stats.max;
+			has_max_stats = true;
+		}
+		if (has_min_stats && has_max_stats && min_stats_type == StringStatsType::EXACT_STATS &&
+		    max_stats_type == StringStatsType::EXACT_STATS && min_stats == max_stats) {
+			auto constant_stats = StringStats::CreateEmpty(type);
+			StringStats::FromConstant(constant_stats, string_t(min_stats));
+			StringStats::SetMin(constant_stats, string_t(min_stats), StringStatsType::EXACT_STATS);
```
Mytherin: We're now calling `StringStats::SetMin` / `StringStats::SetMax` multiple times in this code path. Maybe we can clean up this function a bit?

Mytherin (commented):
Thanks for the PR! LGTM - one comment

extension/parquet/parquet_statistics.cpp
dentiny: for sure!

Mytherin:
Thanks!

PR #18518, opened 2025-08-06
Push Hive filtering into Glob()

xevix:
# Summary

Addresses part of #7620 for local filesystem. Part 1 of the work split off from the original PR #18430. The next part will handle fallback to eager loading in case of Hive issues.

Push down Hive filters into Glob(), which reduces time taken for listing files. Change Glob() to combine tree traversal for directories and files instead of doing it separately, using mutual recursion between functions that process flat directories, `**` crawls, and `*` globbing. This allows Hive filtering at the earliest possible discovery of a path matching the given filters, avoiding unnecessary further traversal.

Example Query with [NOAA GHCN](https://registry.opendata.aws/noaa-ghcn/) data set downloaded locally:

```sql
-- Given Hive-partitioned data such as /by_year/YEAR=2025/ELEMENT=TMAX/file.parquet
SELECT 
  DISTINCT ID 
FROM 'by_year/**/*.parquet'
WHERE year = 2024 AND element = 'TMAX'
ORDER BY ALL;
```

The `year` and `element` filters are used to prune the file list during traversal, instead of afterward.

Query times
Before: ~300-400ms
After:  ~10-30ms

The behavior can be enabled via a flag (defaults to false for now).

```sql
SELECT 
  DISTINCT ID 
FROM read_parquet('by_year/**/*.parquet', hive_lazy_listing=1)
WHERE year = 2024 AND element = 'TMAX'
ORDER BY ALL;
```

Notes
* MultiFileList in LocalFileSystem (not httpfs/S3 yet)
* Incompatible with `union_by_name` since it requires full information (likely other flags too like `explicit_cardinality`, need to check)

# Issues and Questions

* Hive autodetect samples just 1 file (instead of all files eagerly like now), which can yield false positives, but handling this will be dealt with in a follow-up PR

In the future, might change this to list a configurable number of files, similar to the CSV sniffer.

* Statistics on files scanned will be missing, since we don't know the number of files we skipped by pruning early

I think this is an acceptable tradeoff for performance, and to know the total file statistics, this optimization can be turned off.

* Disabled by default for now, will be enabled by default in a follow-up PR which handles failure modes

# Query Plan

Before
```
D EXPLAIN ANALYZE SELECT DISTINCT ID FROM read_parquet('/Users/xevix/Downloads/data/noaa/by_year/**/*.parquet') WHERE YEAR = 2024 AND ELEMENT = 'TMAX' ORDER BY ALL;
┌─────────────────────────────────────┐
[...]

PR #24385, opened 2026-08-01
Prune hive partition directories during glob

xevix:
Supersedes https://github.com/duckdb/duckdb/pull/18518. Addresses part of #7620 for local filesystem.

Many details have been greatly simplified compared to earlier attempts thanks to additions like the Pagination API. Thank you again for taking the time to review this. Please note that Claude was used to write a significant amount of the code here, but I've attempted to simplify it as much as I could.

- [ ] Unrelated failure in Iceberg extension currently on main as of 2026/08/01 44200a30882062bb547a11672d35c282a715def5

# Overview

Prunes hive partition directories early, reducing I/O from listing files that would have been filtered out anyway, greatly reducing overall query time for Hive datasets with a large number of partitions and/or a significant number of Hive keys in the path.

Adds a new option `hive_directory_pruning` which defaults to false to keep existing behavior. When true, Hive filters are pushed into local globbing to prune directories to avoid listing paths exhaustively before filtering. 

Instead of reading all files to ensure Hive partition consistency etc., only a subset (defaulting to 100 files) are sampled. This could technically lead to allowing querying a path whose subpaths do not all have consistent types for their columns (that would normally require `union_by_name=true`), but I believe this to be an acceptable tradeoff.

Hive filters are pushed down as a `HivePathFilter` into `MultiFileList`, which then filters lists as `LocalGlobResult::ExpandNextPath()` is called to glob paths. It might make sense to abstract this to a generic path filter, but given `MultiFilePushdownInfo` and such are already Hive-aware I figured it wasn't a terrible move to keep things simple. Please let me know if you want this to be abstracted out though to support other types of path filtering.

`EXPLAIN` with this option enabled will show the number of pruned directories as a statistic to give the user a sense of how many directories were skipped over, since we can't give the number of files due to skipping them entirely.

# Out of scope

- S3 Support (would be nice as a follow-up)
- JOIN support (pruning on JOINs)
```
WITH filters AS (
    SELECT 2024 AS YEAR, 'TMAX' AS ELEMENT
)
SELECT DISTINCT ID
FROM '/Users/xevix/Downloads/data/noaa/by_year/**/*.parquet'
JOIN filters USING (year, element)
ORDER BY ALL;
```

# Results

Using [NOAA GHCN](https://registry.opendata.aws/noaa-ghcn/) data set.

```
memory D EXPLAIN ANALYZE SELECT DISTINCT ID
         FROM '/Users/xevix/Downloads/data/noaa/by_year/**/*.parquet'
         WHERE YEAR = 2024 AND ELEMENT = 'TMAX'
         ORDER BY ALL;
╭─ Summary ──────────╮
│ Total Time: 0.468s │
│ Data Read:  7.0 MB │
╰────────────────────╯
╭─ Order By ───────────────────────────────────────╮
│ 12,920 rows                                  0µs │
╭─ Hash Group By ─────────┴────────────────────────╮
│ Groups: #0                                       │
│ 12,920 rows                               10.0ms │
╭─ Projection ────────────┚┖───────────────────────╮
│ 3,633,379 rows                               0µs │
│ Function: PARQUET_SCAN                           │
│ Projections: ID                                  │
│ File Filters:                                    │
│ (YEAR = 2024)(ELEMENT = 'TMAX')                  │
│ Scanning Files: 8/13593                          │
│ Total Files Read: 8                              │
│ Filename(s):                                     │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR= … │
│ Row Groups Scanned: 8 / 8                        │
│ 3,633,379 rows                            10.0ms │

type .last to show the full tree, .web to open it in a browser
```

```
memory D SET hive_directory_pruning = true;
memory D EXPLAIN ANALYZE SELECT DISTINCT ID
         FROM '/Users/xevix/Downloads/data/noaa/by_year/**/*.parquet'
         WHERE YEAR = 2024 AND ELEMENT = 'TMAX'
         ORDER BY ALL;
╭─ Summary ───────────╮
│ Total Time: 0.0198s │
│ Data Read:   7.0 MB │
╰─────────────────────╯
╭─ Order By ───────────────────────────────────────╮
│ 12,920 rows                                  0µs │
╭─ Hash Group By ─────────┴────────────────────────╮
│ Groups: #0                                       │
│ 12,920 rows                               10.0ms │
╭─ Projection ────────────┚┖───────────────────────╮
│ 3,633,379 rows                               0µs │
│ Function: PARQUET_SCAN                           │
│ Projections: ID                                  │
│ File Filters:                                    │
│ (YEAR = 2024)(ELEMENT = 'TMAX')                  │
│ Scanning Files: 8/108                            │
│ Pruned Directories: 291                          │
│ Total Files Read: 8                              │
│ Filename(s):                                     │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR= … │
│ Row Groups Scanned: 8 / 8                        │
│ 3,633,379 rows                            20.0ms │
```

# Unrelated regression between 1.5.5 and main branch

It seems that the main branch at the time of writing performs slower than 1.5.5. I'm not sure why this is, but wanted to note that the added latency is confirmed to be between these two versions, and not this branch.

1.5.5
```
$ duckdb 
-- Loading resources from /Users/xevix/.duckdbrc
DuckDB v1.5.5 (Variegata)
memory D EXPLAIN ANALYZE SELECT DISTINCT ID
         FROM '/Users/xevix/Downloads/data/noaa/by_year/**/*.parquet'
         WHERE YEAR = 2024 AND ELEMENT = 'TMAX'
         ORDER BY ALL;
┌─────────────────────────────────────┐
│┌───────────────────────────────────┐│
││    Query Profiling Information    ││
│└───────────────────────────────────┘│
└─────────────────────────────────────┘
EXPLAIN ANALYZE SELECT DISTINCT ID FROM '/Users/xevix/Downloads/data/noaa/by_year/**/*.parquet' WHERE YEAR = 2024 AND ELEMENT = 'TMAX' ORDER BY ALL;
│┌──────────────────────────────────────────────┐│
││              Total Time: 0.336s              ││
│└──────────────────────────────────────────────┘│
┌───────────────────────────┐
│           QUERY           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│      EXPLAIN_ANALYZE      │
│    ────────────────────   │
│                           │
│           0 rows          │
│           0.00s           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│          ORDER_BY         │
│    ────────────────────   │
│           ID ASC          │
│                           │
│                           │
│                           │
│        12,920 rows        │
│           0.00s           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│       HASH_GROUP_BY       │
│    ────────────────────   │
│         Groups: #0        │
│                           │
│                           │
│                           │
│        12,920 rows        │
│           0.01s           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         PROJECTION        │
│    ────────────────────   │
│             ID            │
│                           │
│                           │
│                           │
│       3,633,379 rows      │
│           0.00s           │
└─────────────┬─────────────┘
┌─────────────┴─────────────┐
│         TABLE_SCAN        │
│    ────────────────────   │
│         Function:         │
│        PARQUET_SCAN       │
│                           │
│      Projections: ID      │
│                           │
│       File Filters:       │
│  (YEAR = 2024)(ELEMENT =  │
│          'TMAX')          │
│                           │
│      Scanning Files:      │
│          8/13593          │
│                           │
│    Total Files Read: 8    │
│                           │
│        Filename(s):       │
│   /Users/xevix/Downloads  │
│  /data/noaa/by_year/YEAR  │
│     =2024/ELEMENT=TMAX    │
│/0a916734b79044f889cda53b00│
│ 0d215d_0.snappy.parquet,  │
│   /Users/xevix/Downloads  │
│  /data/noaa/by_year/YEAR  │
│     =2024/ELEMENT=TMAX    │
│/0a916734b79044f889cda53b00│
│ 0d215d_1.snappy.parquet,  │
│   /Users/xevix/Downloads  │
│  /data/noaa/by_year/YEAR  │
│     =2024/ELEMENT=TMAX    │
│/0a916734b79044f889cda53b00│
│ 0d215d_2.snappy.parquet,  │
│   /Users/xevix/Downloads  │
│  /data/noaa/by_year/YEAR  │
│     =2024/ELEMENT=TMAX    │
│/0a916734b79044f889cda53b00│
│ 0d215d_3.snappy.parquet,  │
│   /Users/xevix/Downloads  │
│  /data/noaa/by_year/YEAR  │
│     =2024/ELEMENT=TMAX    │
│/0a916734b79044f889cda53b00│
│ 0d215d_4.snappy.parquet, .│
│             ..            │
│                           │
│                           │
│                           │
│       3,633,379 rows      │
│           0.01s           │
└───────────────────────────┘
Run Time (s): real 0.337 user 0.137312 sys 0.219440
```

main branch (44200a30882062bb547a11672d35c282a715def5)

```
$ ./build/release/duckdb
-- Loading resources from /Users/xevix/.duckdbrc
DuckDB v1.6.0-dev12056 (Development Version, 44200a3088)
memory D EXPLAIN ANALYZE SELECT DISTINCT ID
         FROM '/Users/xevix/Downloads/data/noaa/by_year/**/*.parquet'
         WHERE YEAR = 2024 AND ELEMENT = 'TMAX'
         ORDER BY ALL;
╭─ Summary ──────────╮
│ Total Time: 0.497s │
│ Data Read:  7.0 MB │
╰────────────────────╯
╭─ Order By ───────────────────────────────────────╮
│ 12,920 rows                                  0µs │
╭─ Hash Group By ─────────┴────────────────────────╮
│ Groups: #0                                       │
│ 12,920 rows                               10.0ms │
╭─ Projection ────────────┚┖───────────────────────╮
│ 3,633,379 rows                               0µs │
│ Function: PARQUET_SCAN                           │
│ Projections: ID                                  │
│ File Filters:                                    │
│ (YEAR = 2024)(ELEMENT = 'TMAX')                  │
│ Scanning Files: 8/13593                          │
│ Total Files Read: 8                              │
│ Filename(s):                                     │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR=2… │
│ /Users/xevix/Downloads/data/noaa/by_year/YEAR= … │
│ Row Groups Scanned: 8 / 8                        │
│ 3,633,379 rows                            20.0ms │
```

### DIFF
--- a/src/common/local_file_system.cpp
+++ b/src/common/local_file_system.cpp
@@ -12,6 +12,7 @@
 #include "duckdb/main/database.hpp"
 #include "duckdb/logging/file_system_logger.hpp"
 #include "duckdb/logging/log_manager.hpp"
+#include "duckdb/logging/logger.hpp"
 #include "duckdb/common/multi_file/multi_file_list.hpp"
 
 #include <cstdint>
@@ -2180,6 +2181,11 @@ bool LocalGlobResult::ExpandNextPath() const {
 	auto &current_path = next_dir.path;
 	expand_directories.pop();
 
+	if (!is_empty && PathIsPruned(current_path)) {
+		// no file below this directory can match - skip it entirely instead of listing its contents
+		return true;
+	}
+
 	auto &next_split = splits[split_index];
 	bool is_last_component = split_index + 1 == splits.size();
 	auto &next_component = next_split.path;
@@ -2207,7 +2213,10 @@ bool LocalGlobResult::ExpandNextPath() const {
 			}
 		}
 	} else {
-		// glob - need to resolve the glob
+		// glob - need to resolve the glob by listing the contents of the current path
+		if (context) {
+			DUCKDB_LOG_FILE_SYSTEM_LIST(*context, fs, current_path);
+		}
 		if (IsCrawl(next_component)) {
 			if (is_last_component) {
 				// the crawl is the last component - we are looking for files in this directory
--- a/src/common/multi_file/multi_file_list.cpp
+++ b/src/common/multi_file/multi_file_list.cpp
@@ -32,19 +32,12 @@ MultiFilePushdownInfo::MultiFilePushdownInfo(TableIndex table_index, const vecto
 
 // Helper method to do Filter Pushdown into a MultiFileList
 bool PushdownInternal(ClientContext &context, const MultiFileOptions &options, MultiFilePushdownInfo &info,
-                      vector<unique_ptr<Expression>> &filters, vector<OpenFileInfo> &expanded_files) {
-	HivePartitioningFilterInfo filter_info;
-	for (idx_t i = 0; i < info.column_ids.size(); i++) {
-		if (IsVirtualColumn(info.column_ids[i])) {
-			continue;
-		}
-		filter_info.column_map.insert({info.column_names[info.column_ids[i]].GetIdentifierName(), i});
-	}
-	filter_info.hive_enabled = options.hive_partitioning;
-	filter_info.filename_enabled = options.filename;
+                      vector<unique_ptr<Expression>> &filters, vector<OpenFileInfo> &expanded_files,
+                      optional_ptr<HivePathFilter> path_filter) {
+	auto filter_info = HivePartitioning::GetFilterInfo(options, info);
 
 	auto start_files = expanded_files.size();
-	HivePartitioning::ApplyFiltersToFileList(context, expanded_files, filters, filter_info, info);
+	HivePartitioning::ApplyFiltersToFileList(context, expanded_files, filters, filter_info, info, path_filter);
 
 	if (expanded_files.size() != start_files) {
 		return true;
@@ -53,9 +46,30 @@ bool PushdownInternal(ClientContext &context, const MultiFileOptions &options, M
 	return false;
 }
 
-bool PushdownInternal(ClientContext &context, const MultiFileOptions &options, const vector<Identifier> &names,
-                      const vector<LogicalType> &types, const vector<ColumnIndex> &column_indexes,
-                      const TableFilterSet &filters, vector<OpenFileInfo> &expanded_files) {
+// Installs a filter that prunes paths using the hive partition keys they contain, so that directories that can never
+// contain a matching file are never listed while the file list is expanded
+shared_ptr<HivePathFilter> InstallPathFilter(const MultiFileList &file_list, ClientContext &context,
+                                             const MultiFileOptions &options, const MultiFilePushdownInfo &info,
+                                             const vector<unique_ptr<Expression>> &filters) {
+	if (!options.hive_partitioning || !options.hive_directory_pruning) {
+		return nullptr;
+	}
+	if (file_list.GetFileCount().type == FileExpansionType::ALL_FILES_EXPANDED) {
+		// there is nothing left to expand - a path filter can no longer skip anything
+		return nullptr;
+	}
+	auto path_filter = make_shared_ptr<HivePathFilter>(context, filters, HivePartitioning::GetFilterInfo(options, info),
+	                                                   info.table_index);
+	file_list.SetPathFilter(path_filter);
+	return path_filter;
+}
+
+// Helper method to do dynamic Filter Pushdown into a MultiFileList - this expands the list itself, so that the path
+// filter is installed before any directory is listed
+bool PushdownInternal(const MultiFileList &file_list, ClientContext &context, const MultiFileOptions &options,
+                      const vector<Identifier> &names, const vector<LogicalType> &types,
+                      const vector<ColumnIndex> &column_indexes, const TableFilterSet &filters,
+                      vector<OpenFileInfo> &expanded_files) {
 	TableIndex table_index(0);
 	ExtraOperatorInfo extra_info;
 
@@ -77,29 +91,39 @@ bool PushdownInternal(ClientContext &context, const MultiFileOptions &options, c
 		auto filter_expr = expr_filter.ToExpression(*column_ref);
 		filter_expressions.push_back(std::move(filter_expr));
 	}
+	if (filter_expressions.empty()) {
+		// nothing to push down - expanding here would only prevent a later pushdown from pruning paths
+		return false;
+	}
+	// install the path filter before expanding - this way directories that can never match are never listed
+	auto path_filter = InstallPathFilter(file_list, context, options, info, filter_expressions);
 
-	// call the original PushdownInternal method
-	return PushdownInternal(context, options, info, filter_expressions, expanded_files);
+	// FIXME: don't copy list until first file is filtered
+	expanded_files = file_list.GetAllFiles();
+	return PushdownInternal(context, options, info, filter_expressions, expanded_files, path_filter);
 }
 
 //===--------------------------------------------------------------------===//
 // MultiFileListIterator
 //===--------------------------------------------------------------------===//
-MultiFileListIterationHelper MultiFileList::Files() const {
-	return MultiFileListIterationHelper(*this);
+MultiFileListIterationHelper MultiFileList::Files(MultiFileListScanType scan_type) const {
+	return MultiFileListIterationHelper(*this, scan_type);
 }
 
-MultiFileListIterationHelper::MultiFileListIterationHelper(const MultiFileList &file_list_p) : file_list(file_list_p) {
+MultiFileListIterationHelper::MultiFileListIterationHelper(const MultiFileList &file_list_p,
+                                                           MultiFileListScanType scan_type)
+    : file_list(file_list_p), scan_type(scan_type) {
 }
 
 MultiFileListIterationHelper::MultiFileListIterator::MultiFileListIterator(
-    optional_ptr<const MultiFileList> file_list_p)
+    optional_ptr<const MultiFileList> file_list_p, MultiFileListScanType scan_type)
     : file_list(file_list_p) {
 	if (!file_list) {
 		return;
 	}
 
 	file_list->InitializeScan(file_scan_data);
+	file_scan_data.scan_type = scan_type;
 	if (!file_list->Scan(file_scan_data, current_file)) {
 		// There is no first file: move iterator to nop state
 		file_list = nullptr;
@@ -121,10 +145,10 @@ void MultiFileListIterationHelper::MultiFileListIterator::Next() {
 
 MultiFileListIterationHelper::MultiFileListIterator MultiFileListIterationHelper::begin() { // NOLINT: match stl API
 	return MultiFileListIterationHelper::MultiFileListIterator(
-	    file_list.GetExpandResult() == FileExpandResult::NO_FILES ? nullptr : &file_list);
+	    file_list.GetExpandResult() == FileExpandResult::NO_FILES ? nullptr : &file_list, scan_type);
 }
 MultiFileListIterationHelper::MultiFileListIterator MultiFileListIterationHelper::end() { // NOLINT: match stl API
-	return MultiFileListIterationHelper::MultiFileListIterator(nullptr);
+	return MultiFileListIterationHelper::MultiFileListIterator(nullptr, scan_type);
 }
 
 MultiFileListIterationHelper::MultiFileListIterator &MultiFileListIterationHelper::MultiFileListIterator::operator++() {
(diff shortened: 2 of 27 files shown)


src/include/duckdb/common/multi_file/multi_file_list.hpp
```diff
@@ -123,11 +131,23 @@ class MultiFileList {
...
 	virtual bool FileIsAvailable(idx_t i) const;
 	//! Get the i-th expanded file
 	virtual OpenFileInfo GetFile(idx_t i) const = 0;
+	//! Whether or not the given path can be skipped according to the path filter
+	bool PathIsPruned(const string &path) const;
+
+protected:
+	//! The path filter - only affects which paths are expanded, hence it can be set on a const file list
+	//! Held weakly so that it expires together with the pushdown that installed it
+	mutable weak_ptr<HivePathFilter> path_filter;
```
Mytherin, line 150: This being a weak_ptr is a bit odd from an ownership perspective imo - also should this really be mutable?

src/include/duckdb/common/hive_partitioning.hpp
```diff
@@ -23,17 +25,62 @@ struct HivePartitioningFilterInfo {
...
+	//! The filters (by index) that have been used to prune paths so far
+	unordered_set<idx_t> GetPruningFilters() {
+		lock_guard<mutex> guard(lock);
+		return pruning_filters;
+	}
+
+	//! The number of paths that have been pruned so far
+	idx_t GetPrunedPathCount() {
+		lock_guard<mutex> guard(lock);
+		return pruned_path_count;
```
Mytherin, line 49: This is a bit odd to me - isn't the entire premise of pruning during a glob that we don't really know what we're pruning? i.e. if we do `year=2000 and month=4` and we now glob `year=2000/month=4/**.parquet` instead of `**.parquet` we should have no idea how many paths we pruned since we're not even looking at the other paths - no?

src/common/multi_file/multi_file_list.cpp
```diff
@@ -196,10 +230,16 @@ unique_ptr<MultiFileList> MultiFileList::ComplexFilterPushdown(ClientContext &co
 	if (!options.hive_partitioning && !options.filename) {
 		return nullptr;
 	}
+	if (filters.empty()) {
+		// nothing to push down - expanding here would only prevent a later pushdown from pruning paths
+		return nullptr;
+	}
+	// install the path filter before expanding - this way directories that can never match are never listed
+	auto path_filter = InstallPathFilter(*this, context, options, info, filters);
```
Mytherin, line 238: I think we should be creating a new file list here instead of editing this one in place, like we are doing below

src/common/local_file_system.cpp
```diff
@@ -2180,6 +2181,11 @@ bool LocalGlobResult::ExpandNextPath() const {
 	auto &current_path = next_dir.path;
 	expand_directories.pop();
 
+	if (!is_empty && PathIsPruned(current_path)) {
```
Mytherin, line 2184: There seems to be two separate components to this PR:

* Pushing filtering to be performed later on (lazily) in the globbing process instead of up-front in `MultiFileList::ComplexFilterPushdown` like now
* Push the hive partitioning into the glob, so that if we have `year=*/**.parquet` and we query `year=2000` we don't glob everything but only glob `year=2000/**.parquet`

I think this changeset can be simplified by breaking it up along those lines. Could you split this into these two separate PRs?

test/configs/verify_serializer.json
```diff
@@ -9,6 +9,18 @@
...
+    {
+      "reason": "Filtering out every file leaves an empty file list that cannot be re-bound after serialization",
+      "paths": [
+        "test/sql/copy/partitioned/hive_early_directory_pruning.test"
+      ]
+    },
+    {
+      "reason": "The pushdown is not re-applied after serialization, so no directory is pruned",
+      "paths": [
+        "test/sql/copy/partitioned/hive_early_directory_pruning_listings.test"
```
Mytherin, line 22: I wonder if these skips are required because we're editing the `MultiFileList` in-place instead of generating a new one - could you check if these are still required after the changes above are made?

test/sql/copy/partitioned/hive_directory_pruning_option.test
```diff
@@ -0,0 +1,128 @@
...
+statement ok
+RESET GLOBAL hive_directory_pruning
+
+# with the pruning disabled binding keeps inspecting every file, so a plain file that sits past the bind sample
+# still keeps hive partitioning from being auto-detected
+statement ok
+COPY (SELECT 1 AS k, 0 AS val) TO '{TEST_DIR}/hive_pruning_option_mixed' (FORMAT PARQUET, PARTITION_BY (k));
+
+# fill the partition directory past the bind sample, so the plain file below is only reached by a full expansion
+loop i 0 150
```
Mytherin, line 115: These loops are generating a lot of files and running a lot of SQL statements - maybe this needs to be a `.test_slow` or maybe we can generate these SQL statements in bulk in a different manner instead?

Mytherin (commented):
Thanks for the PR! Looks good - left some comments. My main question is if we can break this up into two PRs as suggested below

src/include/duckdb/common/multi_file/multi_file_list.hpp
xevix: This was a result of trying to put this in the existing MultiFileList, but looking at it again I think this should be entirely separate, which will remove the need for this.

src/include/duckdb/common/hive_partitioning.hpp
xevix: When looking at `**/*.parquet` in a Hive path and on a local filesystem, there are paths listed that may be filtered out. For example if doing `WHERE year = 2000 AND month = 4` when doing the recursive glob with `**/*.parquet` we may list `year=2001/` along the way, but we will not traverse it, so that directory is counted as pruned in that sense, as implemented.

As currently implemented with the local filesystem's pagination API, I believe one "page" is one call to `readdir()` which necessarily must read the full directory at least at that level. So in this case, this would be all the `YEAR=...` directories level.

A different implementation of this could be to translate the Hive filters from the WHERE clause into multiple paths, which might be what you're recommending, but please confirm. 

This would translate from for example this: `FROM '**/*.parquet' WHERE year IN (2000, 2001) AND month = 4` to something like `FROM read_parquet(['YEAR=2000/month=4/*.parquet', 'YEAR=2001/month=4/*.parquet'])`. I'm not sure how hard it would be to do that, but that's definitely more performant since we completely avoid listing paths we can logically eliminate ahead of time. I guess we could detect if we have Hive partitions first for the path passed in, then check the filters to see if they match Hive, and if a query rewrite would make sense. I'm not sure how hard this would be to do in the query pipeline, but if this is what you mean I could explore it.

src/common/local_file_system.cpp
xevix: Question about this split: I changed to sampling some files for things like Hive detection (instead of exhaustively checking all partitions), would this fall under the first component/PR you listed? In order not to break existing behavior such as failing if any directory doesn't match the Hive schema, should we still gate this behind a flag defaulting to false for this partial work?

test/configs/verify_serializer.json
xevix: One of these skips was actually reproduced on main I think, I'll see if I can isolate it into a separate PR.

src/common/local_file_system.cpp
xevix: @Mytherin One point of clarification that would help me greatly, where later on in the globbing process would filtering be done, if not in the glob itself like in the second component you mention? Sorry if this is a redundant question, but I wasn't able to come up with a good answer here. This may be related to the question I asked on `hive_partitioning.hpp` as well.

xevix:
Closing in favor of smaller piecemeal PRs based on feedback.

PR #24388, opened 2026-08-01
Use unimplemented exception over internal one

### LINKED ISSUE #24377: ALTER TABLE ... ADD CHECK/FOREIGN KEY throws INTERNAL Error
### What happens?

See repro for details.

### To Reproduce

```sql
db D CREATE TABLE t(a INT);
db D ALTER TABLE t ADD CHECK (a > 0);
INTERNAL Error:
unsupported constraint type in ALTER TABLE statement
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24377

Hi team, I think we should use `Unimplemented` exception instead of `Internal` exception -- we should never throw Internal exception.

### DIFF
--- a/src/catalog/catalog_entry/duck_table_entry.cpp
+++ b/src/catalog/catalog_entry/duck_table_entry.cpp
@@ -1377,7 +1377,8 @@ unique_ptr<CatalogEntry> DuckTableEntry::AddConstraint(ClientContext &context, A
 		table_info.constraints.push_back(info.constraint->Copy());
 
 	} else {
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks! LGTM

PR #24392, opened 2026-08-01
Re-enable five stale mode-skips, including two fuzzer regression tests

carlopi:
`mode skip` is the test-side FIXME: it records that something was broken, and rots the same way when the condition stops holding. 101 directives across 85 files; probing the ten whose comment names a specific defect found five that no longer apply.

Three pass unmodified (5/5 runs each), and their FIXMEs had become false claims:
 - optimizer/statistics/statistics_is_null.test - "we don't yet correctly track when columns don't contain valid values"; IS NULL/IS NOT NULL on an all-NULL column is now statically determined. 52 assertions restored.
 - fuzzer/pedro/parser_roundtrip.test - "explicit cross joins not rendered correctly (yet)"; they now round-trip.
 - fuzzer/sqlsmith/subquery_binding_error.test - bare FIXME.

Two more had their defect fixed while the test's expectation stayed stale, so they were failing for the wrong reason. Both are fuzzer regression tests that have been dark:
 - fuzzer/afl/issue_7551.test - "Segmentation fault on SELECT". No crash and no internal exception now: the VARCHAR-to-DATE comparison is rejected at bind time. Changed to assert that, so the test still guards the crash.
 - fuzzer/sqlsmith/statistics_assertion_trigger.test - the statistics assertion is gone. The query still fails, but for a legitimate data-dependent reason: it uses a VARCHAR column as a join condition and test_all_types()'s varchar value does not convert to BOOL. Changed to assert that.

Green under force_storage, latest_storage and verify_statement_to_string.

Still genuinely skipped, defect confirmed live: transitive_filters (wrong results from the filter combiner), parquet_or_pushdown (wrong results), and the three enum tests (dependencies between enums and tables).

### DIFF
--- a/test/configs/force_storage.json
+++ b/test/configs/force_storage.json
@@ -147,6 +147,7 @@
         "test/fuzzer/sqlsmith/interval_multiply_overflow.test",
         "test/fuzzer/sqlsmith/mad_overflow.test",
         "test/fuzzer/sqlsmith/makedate_overflow.test",
+        "test/fuzzer/sqlsmith/statistics_assertion_trigger.test",
         "test/fuzzer/sqlsmith/timestamp_overflow.test",
         "test/fuzzer/sqlsmith/timestamptz_double_cast.test",
         "test/fuzzer/sqlsmith/window-list-value.test",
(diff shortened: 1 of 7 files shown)


Mytherin:
Thanks!

carlopi:
2 CI failures (for now), but I am not really sure if related, they seems in different files (although, we might be propagating state somehow obviously)

PR #24393, opened 2026-08-01
Range check float => int casts on the rounded value

carlopi:
TryCastWithOverflowCheckFloat checked the range before applying the statistical rounding, so a value below the destination maximum could still round above it. 127.6::TINYINT passed the check, rounded to 128.0, and reached a static_cast that is undefined behaviour for an out-of-range float - in practice it returned -128. The mirror case was rejected outright: -128.4 rounds to -128, which fits, but errored because -128.4 itself is below the minimum.

Check the rounded value instead, which is what postgres's `dtoi2`/`dtoi4` do:
```
127.6::TINYINT   throws (was UB, in regular cases -128)
127.5::TINYINT   throws (was UB, in regular cases -128)
-128.4::TINYINT  is -128 (was a Conversion Error)
126.5::TINYINT   is 126, Banker's rounding unchanged
```

The whole suite passes unchanged, which is also why the hole survived: no test covered either boundary.

I think this sort of makes sense, or more sense than current behaviour.

### DIFF
--- a/src/include/duckdb/common/operator/numeric_cast.hpp
+++ b/src/include/duckdb/common/operator/numeric_cast.hpp
@@ -76,11 +76,14 @@ bool TryCastWithOverflowCheckFloat(SRC value, T &result, SRC min, SRC max) {
 	if (!Value::IsFinite<SRC>(value)) {
 		return false;
 	}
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24394, opened 2026-08-01
Trim sort key layout specializations under `SMALLER_BINARY`

carlopi:
The whole sorting stack - the run sort, the merger, the sort key scatter/gather and the three range joins - is templated on SortKeyType, and every dispatch site enumerated all 9 layouts by hand. That is 9 instances of vergesort, ska_sort and pdqsort in the binary, ~623 KB, and 17 near-identical switch statements to keep in sync.

Replace those switches with a DUCKDB_FOR_EACH_SORT_KEY_TYPE X-macro (plus the payload-only and reorderable subsets) so the enabled set is declared once, and register a sort_key_layouts feature that drops the three layouts replaceable by a wider one already present: a key without payload rounds up to NO_PAYLOAD_FIXED_32, and a payload key over 24 bytes falls back to PAYLOAD_VARIABLE_32.

The other six cannot go. NO_PAYLOAD_FIXED_8 and PAYLOAD_FIXED_16 are the BIGINT sort key and VariableSortKey::Construct(int64_t) throws; PAYLOAD_FIXED_24 is what an index build over an 8-byte column uses, and widening it fails test_art_mem_limit.test, which builds one under a 10MB memory_limit.

This buys code size with time. Order By over 20M rows, trimmed vs not, against a 1.19x noise band measured by running one binary in both slots:

```
  ORDER BY 1x INTEGER  NO_PAYLOAD_FIXED_8  kept          1.06x
  ORDER BY 1x BIGINT   NO_PAYLOAD_FIXED_16 -> FIXED_32   1.56x
  ORDER BY 2x BIGINT   NO_PAYLOAD_FIXED_24 -> FIXED_32   1.26x
  ORDER BY 3x BIGINT   NO_PAYLOAD_FIXED_32 kept          1.03x
```

Collapsing everything onto the variable-size key saves ~90 KB more but costs 1.6-2.4x, because a variable-size key compares through the heap.

----

Size measurements, duckdb CLI, macOS arm64:
```
                                    make   SMALLER_BINARY=1   +EXCEPT=sort_key_layouts
  before                      56,880,936         46,937,432                        n/a
  after                       56,880,936         46,640,152                 46,937,432
```

Both n/a-free columns are byte-identical before and after, so the feature is the only difference: -297,280 on the CLI, -297,296 on libduckdb.dylib, and the same whether it is trimmed alone or inside a full SMALLER_BINARY=1 build.

### DIFF
--- a/src/common/sort/sorted_run.cpp
+++ b/src/common/sort/sorted_run.cpp
@@ -27,24 +27,11 @@ SortedRunScanState::SortedRunScanState(ClientContext &context, const Sort &sort_
 void SortedRunScanState::Scan(const SortedRun &sorted_run, const Vector &sort_key_pointers, DataChunk &chunk) {
 	const auto sort_key_type = sort.key_layout->GetSortKeyType();
 	switch (sort_key_type) {
-	case SortKeyType::NO_PAYLOAD_FIXED_8:
-		return TemplatedScan<SortKeyType::NO_PAYLOAD_FIXED_8>(sorted_run, sort_key_pointers, chunk);
-	case SortKeyType::NO_PAYLOAD_FIXED_16:
-		return TemplatedScan<SortKeyType::NO_PAYLOAD_FIXED_16>(sorted_run, sort_key_pointers, chunk);
-	case SortKeyType::NO_PAYLOAD_FIXED_24:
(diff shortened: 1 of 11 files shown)


Mytherin:
Thanks!

PR #24396, opened 2026-08-01
Proposal to shrink PEG parser code: -1% on regular build, -3% on `SMALLER_BINARY` mode

carlopi:
First some results, before / after x regular / SMALLER_BINARY / SMALLER_BINARY but for this optimization
```
  ┌───────┬───────────────────┬───────────────────────┬──────────────────────────────────────────────────────────────────────────────────────┐
  │       │       make        │ SMALLER_BINARY=1 make │        SMALLER_BINARY=1 SMALLER_BINARY_EXCEPT=peg_trampoline_transformer make        │
  ├───────┼───────────────────┼───────────────────────┼──────────────────────────────────────────────────────────────────────────────────────┤
  │ pre   │        56,880,936 │            46,937,432 │ N/A — cmake rejects: Unknown SMALLER_BINARY_EXCEPT name 'peg_trampoline_transformer' │
  ├───────┼───────────────────┼───────────────────────┼──────────────────────────────────────────────────────────────────────────────────────┤
  │ now   │        56,301,192 │            45,278,904 │                                                                           46,374,232 │
  ├───────┼───────────────────┼───────────────────────┼──────────────────────────────────────────────────────────────────────────────────────┤
  │ delta │ −579,744 (−1.02%) │   −1,658,528 (−3.53%) │                                                                                    — │
  └───────┴───────────────────┴───────────────────────┴──────────────────────────────────────────────────────────────────────────────────────┘
```
Basically this look at parser code gives visible code size advantages using 2 main tecniques:
* restructuring code via helpers, this buys about 1% in every mode, at no cost but code churn (that is a cost on its own)
* avoiding (in `SMALLER_BINARY` mode) to ship 2 different parser implementations (in this case, we skip the trampoline based one given it's not the default). This restrict SQL surface (`SET debug_transformer_trampoline_style=true` would now throw when this compile mode is enabled) for an additional -2%.

This is more a proposal / showing a possible direction, unclear if it can land as-is, but it would be great if some of this would be incorporated giving the size reductions are visible, and opting in to the would be great for `duckdb-wasm` in particular.


Mytherin:
Thanks for the PR!

> avoiding (in SMALLER_BINARY mode) to ship 2 different parser implementations (in this case, we skip the trampoline based one given it's not the default). This restrict SQL surface (SET debug_transformer_trampoline_style=true would now throw when this compile mode is enabled) for an additional -2%.

The idea is in the final release to only have the trampoline style parser - so this is not really useful imo. `debug_transformer_trampoline_style` is a temporary setting (and this seems to be now defaulting to the wrong one long term).

carlopi:
Sure, this was more pointing that on the mostly generated parser code making a pass with the goal of reducing code-size goes a long way, and the fact that's mostly generated offers more occasions to structure it.

And the other part is realising that whatever solution we end up, a single parser might provide a viable surface, so choosing one (possibly behind a `#ifdef`) weight about 2% of current binaries.

Path forward is unclear, likely just close this PR, move this as an improvement idea (internal issue), and let @Dtenwolde have a proper look (to be scheduled whenever works best).

carlopi:
I'll come back to this, there are still code size improvements in the transformer, but let's first settle other work there.

PR #24397, opened 2026-08-01
Regenerate iceberg 0003 patch against the bumped pin

carlopi:
The `bump iceberg` pin move to 26583af7 removed the ApplyEqualityDeletes declaration that sat directly above ParseOption, so the hunk context in 0003-copy-option-identifier-map.patch no longer matched and apply_extension_patches.py failed the Main (cloud) extension build.


carlopi:
Unblocks https://github.com/duckdb/duckdb/pull/24396 + https://github.com/duckdb/duckdb/pull/24394 and `main` CI in general. Not clear to me why this was not caught by merge queue.

carlopi:
Actually, this improve stuff (patches can be applied) but compile fails due to another hidden merge. I think this can be merged (although, it will fail) or someone can have a look at fixing this.

PR #24400, opened 2026-08-01
Fix #24378: Prevent false-positive generated column constraint checks when dropping columns with CASCADE

### LINKED ISSUE #24378: DROP COLUMN ... CASCADE incorrectly fails with generated dependent column plus single-column CHECK
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t(
           a INT,
           b INT GENERATED ALWAYS AS (a + 1),
           c INT,
           CHECK(a > 0)
         );
memory D ALTER TABLE t DROP COLUMN a CASCADE;
Binder Error:
Table does not contain column a referenced in check constraint!
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

sdk011:
### What changes were proposed in this pull request?
Fixes #24378.

When dropping a regular column via `ALTER TABLE ... DROP COLUMN ... CASCADE`, if the dropped column has dependent `GENERATED ALWAYS AS (...)` columns and the table has a `CHECK` constraint, `DuckTableEntry::RemoveColumn` previously failed with:

Binder Error: Column "..." referenced in CHECK constraint not found!

#### Root Cause
In `DuckTableEntry::RemoveColumn`, `dropped_column_is_generated` was calculated by iterating over all columns being removed (`removed_columns`). If **any** column in `removed_columns` was generated (including dependent generated columns dropped via `CASCADE`), `dropped_column_is_generated` was set to `true`.

When `UpdateConstraintsOnColumnDrop(..., bool is_generated)` was subsequently called with `is_generated = true`, it assumed the primary dropped column was generated and that existing constraints could not reference it. Therefore, it re-added all table constraints unchanged—causing `CHECK` constraints referencing the primary dropped column to remain on the altered table and fail binder verification.

#### Fix
Changed `dropped_column_is_generated` in `DuckTableEntry::RemoveColumn` to check only whether the primary dropped column (`removed_index`) is generated:

```cpp
bool dropped_column_is_generated = columns.GetColumn(LogicalIndex(removed_index)).Generated();
```

####  Why are the changes needed?
This enables standard schema migrations (ALTER TABLE ... DROP COLUMN ... CASCADE) on tables that contain regular columns, dependent generated columns, and table check constraints.

#### How was this patch tested?
* Added an sqllogictest regression test in test/sql/generated_columns/virtual/check.test verifying that dropping a regular column with CASCADE correctly drops both the dependent generated column and any check constraints referencing the dropped column.
* Ran full unit test suite (build/reldebug/test/unittest).

### DIFF
--- a/src/catalog/catalog_entry/duck_table_entry.cpp
+++ b/src/catalog/catalog_entry/duck_table_entry.cpp
@@ -823,12 +823,9 @@ unique_ptr<CatalogEntry> DuckTableEntry::RemoveColumn(ClientContext &context, Re
 	if (!removed_columns.empty() && !info.cascade) {
 		throw CatalogException("Cannot drop column: column is a dependency of 1 or more generated column(s)");
 	}
-	bool dropped_column_is_generated = false;
+	bool dropped_column_is_generated = columns.GetColumn(LogicalIndex(removed_index)).Generated();
 	for (auto &col : columns.Logical()) {
 		if (col.Logical() == removed_index || removed_columns.count(col.Logical())) {
-			if (col.Generated()) {
-				dropped_column_is_generated = true;
-			}
 			continue;
 		}
 		create_info->columns.AddColumn(col.Copy());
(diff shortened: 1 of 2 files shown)


sdk011:
@d-justen - Can you review this , if you have context into this ?

PR #24402, opened 2026-08-02
fix: disable streaming for bounded first_value frames

belowzeroff:
Summary:
- Disable streaming execution for bounded first_value window frames.
- Add regression coverage for bounded first_value behavior in streaming window tests.

Tests:
- build/reldebug/test/unittest test/sql/window/test_streaming_window.test


lnkuiper:
Thanks! Looks like someone had the same idea. Superseded by https://github.com/duckdb/duckdb/pull/24497

PR #24404, opened 2026-08-02
Checks inline - throws out of line

carlopi:
This is a code generation improvement, that should be both make code somewhat faster AND leaner.
Code size is easy to put a number on, it should be −2.44% of current CLI binary size, and the nice thing is that it can be enabled by default given the produced code is better.

Resulting behaviour (as demonstrated by no test changes) should be not affected.

Idea is: before exceptions would still need to be constructed and allocated on the function performing the check, and that would cause a cost in all code-paths (both regular, and the one where the check triggers). After the allocation and construction of exception is moved to the slow path. Given exceptions are meant to be exceptional, it should be sound that sinking code into codepath rarely taken is a win both on performances AND code-side.

I have not run proper benchmarks, let's see what CI thinks...

CI think this is mostly positive, with some randomness but on average across all benchmark suites it's about 8 to 9% faster.

Discovered while on a side-quest to figure out how to ship a lean(er) duckdb-wasm base binary.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -58,6 +58,7 @@ add_library_unity(
   fsst.cpp
   gzip_file_system.cpp
   identifier.cpp
+  memory_safety.cpp
   hive_partitioning.cpp
   json_document.cpp
   pipe_file_system.cpp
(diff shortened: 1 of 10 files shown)


Mytherin:
Great find - thanks!

PR #24405, opened 2026-08-02
Add TypeIdToChars and stop building error strings on the success path

carlopi:
TypeIdToString returns a string built from a short literal, so every one of its 1305 out-of-line call sites constructs (and usually immediately destroys) a std::string, on paths that are almost always error paths that never fire. It is also only non-throwing by accident: every PhysicalType name happens to fit libc++'s small-string buffer, so a longer name added later would silently turn an allocation failure into a new failure mode.

Split it the way EnumUtil already does - a const char * core with a string wrapper on top. TypeIdToChars returns the literal and is noexcept; TypeIdToString stays as string(TypeIdToChars(type)) so the public API is unchanged. In-tree callers move to TypeIdToChars, except where the result heads a string concatenation.

CastTimestampOperation took its error message by const string &, so both callers paid for it on every cast: CastTimestampTargetOperation ran StringUtil::Format per value, and CastTimestampPrecisionOperation allocated because its literal is 44 characters and does not fit the small-string buffer. Take a const char * instead and format at the throw site, so a successful cast builds nothing.

This should be neutral on native code size (−0.14%), minor positive on wasm code size, and possibly code-generations improves by moving code to slow path (although I have not run benchmarks, let's see what CI thinks).

### DIFF
--- a/src/common/exception.cpp
+++ b/src/common/exception.cpp
@@ -226,7 +226,7 @@ void Exception::SetQueryLocation(QueryLocation error_location, unordered_map<str
 }
 
 InvalidTypeException::InvalidTypeException(PhysicalType type, const string &msg)
-    : Exception(ExceptionType::INVALID_TYPE, "Invalid Type [" + TypeIdToString(type) + "]: " + msg) {
+    : Exception(ExceptionType::INVALID_TYPE, string("Invalid Type [") + TypeIdToChars(type) + "]: " + msg) {
 }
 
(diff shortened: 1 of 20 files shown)


Tishj:
If the goal is `CastTimestampOperation`, getting the error message could just be a callback, that only gets invoked in the error path, instead of constructing it upfront

carlopi:
I can't get a reliable justification for the code churn in CI numbers, let's pass for now.

PR #24406, opened 2026-08-02
Fix encrypted Parquet DataPageV2 reads

### LINKED ISSUE #24785: DuckDB cannot read encrypted Parquet DataPageV2 pages
## What happens?

DuckDB cannot read Parquet Modular Encryption (PME) files whose data pages use `DataPageV2`. Equivalent `DataPageV1` files succeed.

For uncompressed `DataPageV2`, DuckDB reports an `uncompressed page size mismatch` because the reader compares the framed encrypted size against the plaintext size. For compressed `DataPageV2`, it reports `Too many bytes requested from crypto buffer` because the reader tries to decrypt the level prefix and encoded value suffix separately even though the page body is one authenticated module.

The proposed fix and minimal parquet-rs fixtures are in #24406.

## To reproduce

Download the two public fixtures from #24406:

```bash
curl -LO https://raw.githubusercontent.com/faysou/duckdb/14d43fe45860870e328df89c59ce4a837feaff13/data/parquet-testing/encryption/arrow_rust_pme_data_page_v2_uncompressed_required.parquet
curl -LO https://raw.githubusercontent.com/faysou/duckdb/14d43fe45860870e328df89c59ce4a837feaff13/data/parquet-testing/encryption/arrow_rust_pme_data_page_v2_snappy_required.parquet
```

Run these queries with an unpatched DuckDB CLI:

```sql
LOAD parquet;
PRAGMA add_parquet_key('arrow_key_generated_files', '0123456789abcdef');

SELECT value
FROM read_parquet(
    'arrow_rust_pme_data_page_v2_uncompressed_required.parquet',
    encryption_config = {footer_key: 'arrow_key_generated_files'}
);

SELECT value
FROM read_parquet(
    'arrow_rust_pme_data_page_v2_snappy_required.parquet',
    encryption_config = {footer_key: 'arrow_key_generated_files'}
);
```

The first query fails with an uncompressed page size mismatch. The second fails with `Too many bytes requested from crypto buffer`.

Both should return:

```text
7
9
11
13
```

## Environment

- OS: macOS 26.5.2, arm64.
- DuckDB version: `main` at `562c3e2ebddbef85a9ad788edcc11bf4a15ec379`.
- DuckDB client: CLI.
- Full name: Faysal Aberkane.
- Affiliation: Independent.

## Before submitting

- [x] I included the relevant configuration.
- [x] I included all code required to reproduce the issue.
- [x] I included the relevant data sets through public, commit-pinned links.

faysou:
### Summary

Apply Arrow‑compatible PME framing to DataPageV2 before validating or reading its body. Decrypt the
page body as one authenticated module, retain the uncompressed repetition and definition level
prefix, and decompress only the encoded value suffix when `is_compressed` is true. Reuse the same
framing calculation for V1, V2, and dictionary pages.

### Read flow

```mermaid
flowchart TB
    Header["Read encrypted page header"] --> Size["Subtract length, nonce, and tag framing"]
    Size --> Kind{"DataPageV2?"}
    Kind -->|no| Legacy["Use existing V1 or dictionary decode"]
    Kind -->|yes| Compressed{"Value suffix compressed?"}
    Compressed -->|no| Plain["Decrypt the full page body once"]
    Compressed -->|yes| Module["Decrypt the full page body once"]
    Module --> Levels["Copy repetition and definition level prefix"]
    Levels --> Values["Decompress the encoded value suffix"]
    Plain --> Project["Project decoded column values"]
    Values --> Project
    Legacy --> Project
```

### Root cause

Parquet modular encryption wraps one data page body as one encrypted module. DuckDB already removed
the 32‑byte `length + nonce + tag` overhead for V1 pages, but its V2 path used the framed size and
called the decryptor separately for the level prefix and encoded value suffix. Uncompressed V2
therefore failed its size check by 32 bytes, while compressed V2 opened a second decrypt transport
in the middle of the same encrypted module and over‑requested plaintext.

### Contract

- Remove Arrow PME framing before any V2 page‑size or level‑length comparison.
- Treat `is_compressed = false` as an uncompressed page body even when the column codec is set.
- Validate uncompressed V2 bodies against the adjusted plaintext size.
- Decrypt each encrypted V2 page body exactly once with its page AAD.
- Keep repetition and definition levels uncompressed and decompress only the value suffix.
- Preserve V1 and dictionary page behavior through the shared page‑data size calculation.
- Reject framed page sizes smaller than the encryption overhead instead of underflowing.

### Design decisions

- Keep the fix in `ColumnReader`; no encryption transport or Parquet metadata contract changes.
- Reuse `GetPageDataSize` for all page versions so framing arithmetic remains expressed once.
- Materialize the adjusted compressed V2 body before splitting it. PME authentication covers the
  complete module, so a split decrypt is not valid.
- Commit parquet‑rs 58.3.0 fixtures because they reproduce the interoperable writer behavior that
  DuckDB must accept, independent of DuckLake catalog logic.

### Issue

Fixes #24785.

### Verification

- Baseline DuckDB `main` matrix: all V1 cases passed; uncompressed V2 failed with a 32‑byte size
  mismatch; compressed V2 failed with `Too many bytes requested from crypto buffer`.
- `CMAKE_BUILD_PARALLEL_LEVEL=8 make reldebug`: passed with the core and Parquet extensions.
- Direct patched‑CLI matrix: all V1/V2, uncompressed/Snappy, and required/nullable cases projected
  exact values and null positions.
- `build/reldebug/test/unittest test/sql/copy/parquet/encryption/arrow_compatibility.test`: passed,
  82 assertions.
- `build/reldebug/test/unittest test/parquet/invalid_parquet.test`: passed, 6 assertions.
- `build/reldebug/test/unittest test/parquet/uncompressed_page_size_corruption.test`: passed,
  1 assertion.
- Group 26 DuckLake bare‑file writer oracle against the patched CLI: passed, 1 test.
- Focused DuckDB format and typo check for the changed source and test files: passed.
- `git diff --check`: passed.

### Specification proof

| Parquet rule | Reader invariant | Discriminating proof |
| ------------ | ---------------- | -------------------- |
| [Compression completes before page encryption](https://github.com/apache/parquet-format/blob/master/Encryption.md#3-technical-approach) | DuckDB decrypts the complete V2 page module before it interprets the level prefix or decompresses values. | Both Snappy V2 fixtures project their exact required and nullable values; the old split decrypt fails both fixtures. |
| [AES_GCM_V1 modules use length, nonce, ciphertext, and tag framing](https://github.com/apache/parquet-format/blob/master/Encryption.md#51-encrypted-module-serialization) | The reader subtracts the 4‑byte length, 12‑byte nonce, and 16‑byte tag before page‑body validation. | Both uncompressed V2 fixtures pass exact size equality; baseline reports actual sizes exactly 32 bytes too large. |
| [A data page is an indivisible page unit](https://github.com/apache/parquet-format/blob/master/README.md#glossary) | One authenticated decrypt operation consumes one encrypted DataPageV2 body. | Both compressed V2 fixtures pass after the two decrypt calls become one; baseline fails with a crypto buffer over‑request. |
| [Data pages store repetition levels, definition levels, then encoded values](https://github.com/apache/parquet-format/blob/master/README.md#data-pages) | The decrypted V2 body keeps the level bytes as a prefix and decompresses only the remaining value bytes. | Nullable fixtures exercise definition levels and return `7,NULL,11,13`; required fixtures omit them and return `7,9,11,13`. |
| [`DataPageHeaderV2` defines level byte lengths and `is_compressed`](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift) | Level lengths use the adjusted plaintext body size, and `is_compressed = false` selects the direct read path. | Uncompressed and Snappy fixtures pass for both required and nullable schemas, while malformed V2 size tests still fail. |
| [Page size covers all level and value pieces](https://github.com/apache/parquet-format/blob/master/README.md#data-pages) | Adjusted compressed and uncompressed sizes guard the full V2 body without including encryption framing. | Existing truncated and inconsistent page fixtures retain all expected failures in the invalid‑Parquet tests. |

### Upstream readiness

The DuckDB fix is reviewable as one commit with a baseline reproducer, a complete interoperable PME
matrix, focused corruption coverage, and a DuckLake writer oracle.


justin0mcateer:
We are moving to add modular encryption to our parquet files generated from TypeScript and we have also run into the datapage V2 issue fixed here. In our case, we use LIST and STRUCT heavily so Datapage V2 would potentially be quite beneficial for us.

faysou:
Someone needs to approve the workflows to run, they will likely run successfully after my PR update done a few days ago.

On my side I encountered the issue by trying to implement the officially specified ducklake encryption feature in datafuson-ducklake and there was a bug when using the default Parquet 2.0 format with duckdb as backend.

PR #24408, opened 2026-08-02
remove flaky test

artjomPlaunov:
https://github.com/duckdblabs/duckdb-internal/issues/10173

I tried to add this SQL test as part of https://github.com/duckdb/duckdb/pull/24260, but it is flaky; fine to remove it since there were some C++ tests as part of that PR (I just thought it would have been nice to add the SQL reproducer I had for the issue)


Mytherin:
Thanks!

PR #24409, opened 2026-08-02
Fix#24303 date_trunc comparisons with infinite timestamps

### LINKED ISSUE #24303: Disable optimizer leads to different results
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE dt_inf(d TIMESTAMP);
memory D INSERT INTO dt_inf VALUES
           (TIMESTAMP '-infinity'),
           (TIMESTAMP '2025-01-01 12:00:00'),
           (TIMESTAMP 'infinity');
memory D 
memory D SELECT COUNT(*) FROM dt_inf WHERE date_trunc('day', d) = TIMESTAMP 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            0 │
└──────────────┘
memory D SELECT COUNT(*) FROM dt_inf WHERE date_trunc('day', d) <= TIMESTAMP 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            2 │
└──────────────┘
memory D SELECT COUNT(*) FROM dt_inf WHERE date_trunc('day', d) != TIMESTAMP 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            3 │
└──────────────┘
memory D SELECT COUNT(*) FROM dt_inf WHERE date_trunc('day', d) = TIMESTAMP '-infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            0 │
└──────────────┘

┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 

memory D SELECT COUNT(*) FROM dt_inf WHERE date_trunc('day', d) = TIMESTAMP 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
memory D SELECT COUNT(*) FROM dt_inf WHERE date_trunc('day', d) <= TIMESTAMP 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            3 │
└──────────────┘
memory D SELECT COUNT(*) FROM dt_inf WHERE date_trunc('day', d) != TIMESTAMP 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            2 │
└──────────────┘
memory D SELECT COUNT(*) FROM dt_inf WHERE date_trunc('day', d) = TIMESTAMP '-infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

idle-lab:
Closes #24303
This PR fixes issue #24303 which `DateTruncSimplificationRule` produces incorrect results when comparing against `±infinity`.

`DateTruncSimplificationRule` rewrites predicates of the form `date_trunc(constant_part, column) OP constant` into a range predicates directly on the underlying column.

For finite constants, the rule can rewrite a `date_trunc` comparison into a range comparison safely. This does not work for infinity because adding an interval to infinity still produces infinity. For example: `date_trunc('day', d) = TIMESTAMP 'infinity'` was rewritten to the equivalent of: `d >= TIMESTAMP 'infinity' AND d < TIMESTAMP 'infinity'`.

Since date_trunc preserves temporal infinity values, comparisons against infinity can instead be simplified directly: `date_trunc(part, column) OP infinity` to `column OP infinity`.

And I add some tests covered:

- Positive and negative infinity
- All six comparison operators
- Infinity on either side of the comparison
- TIMESTAMP and DATE columns
- NULL and IS [NOT] DISTINCT FROM semantics
- The TIMESTAMPTZ path provided by ICU

### DIFF
--- a/src/optimizer/rule/date_trunc_simplification.cpp
+++ b/src/optimizer/rule/date_trunc_simplification.cpp
@@ -63,19 +63,12 @@ unique_ptr<Expression> DateTruncSimplificationRule::Apply(LogicalOperator &op, v
 		rhs_comparison_type = FlipComparisonExpression(comparison_type);
 	}
 
+	auto &column_side = col_is_lhs ? left : right;
+	auto &constant_side = col_is_lhs ? right : left;
+
 	// date_trunc preserves temporal infinities, so only remove the function call.
 	if (IsInfinity(rhs.GetValue())) {
-		if (col_is_lhs) {
-			left = column_part.Copy();
-			if (rhs.GetReturnType().id() != left->GetReturnType().id()) {
-				right = CastAndEvaluate(std::move(right), left->GetReturnType());
-			}
-		} else {
-			right = column_part.Copy();
-			if (rhs.GetReturnType().id() != right->GetReturnType().id()) {
-				left = CastAndEvaluate(std::move(left), right->GetReturnType());
-			}
-		}
+		ReplaceDateTruncWithColumn(column_side, constant_side, column_part);
 		changes_made = true;
 		return nullptr;
 	}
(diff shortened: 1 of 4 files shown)


src/optimizer/rule/date_trunc_simplification.cpp
```diff
@@ -63,6 +63,23 @@ unique_ptr<Expression> DateTruncSimplificationRule::Apply(LogicalOperator &op, v
 		rhs_comparison_type = FlipComparisonExpression(comparison_type);
 	}
 
+	// date_trunc preserves temporal infinities, so only remove the function call.
+	if (IsInfinity(rhs.GetValue())) {
+		if (col_is_lhs) {
```
Mytherin: This seems like duplicated code from another part of the rule - can we unify these?

Mytherin (commented):

src/optimizer/rule/date_trunc_simplification.cpp
idle-lab: Thanks for your review. I extracted the shared replacement and casting logic into `ReplaceDateTruncWithColumn`, and also reused it in the other parts of this rule.

PR #24411, opened 2026-08-02
Rewrite IS DISTINCT FROM NULL

### LINKED ISSUE #24410: `NOT NULL` constraint not leveraged to prune row group
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t_isnotdistinct(i INTEGER NOT NULL);
memory D INSERT INTO t_isnotdistinct SELECT range FROM range(300000);

-- IS NOT DISTINCT FROM doesn't prune row groups
memory D EXPLAIN ANALYZE
         SELECT count(*) FROM t_isnotdistinct
         WHERE i IS NOT DISTINCT FROM NULL;
╭─ Summary ───────────╮
│ Total Time: 0.0017s │
╰─────────────────────╯
│ Aggregates: count_star()             │
╰───────────────────┬──────────────────╯
╭─ Table Scan ──────┴──────────────────╮
│ Table: memory.main.t_isnotdistinct   │
│ Filters: i IS NOT DISTINCT FROM NULL │
│ Row Groups Scanned: 3 / 3            │
│ 0 rows                           0µs │

-- IS NOT NULL prunes row group correctly
memory D EXPLAIN ANALYZE
         SELECT count(*) FROM t_isnotdistinct
         WHERE i IS NOT NULL;
╭─ Summary ─────────╮
│ Total Time: 3.00s │
╰───────────────────╯
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24410

I found `IS DISTINCT FROM NULL` doesn't leverage constraint to prune row group, but `IS NULL` prunes well; so in this PR I rewrite the former to later.


src/optimizer/rule/comparison_simplification.cpp
```diff
@@ -150,6 +159,16 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op,
 	if (!ExpressionExecutor::TryEvaluateScalar(GetContext(), constant_expr, constant_value)) {
 		return nullptr;
 	}
+	if (constant_value.IsNull() && column_ref_expr.GetExpressionClass() == ExpressionClass::BOUND_COLUMN_REF) {
```
Mytherin: Why add the `BOUND_COLUMN_REF` restriction here? Can't we always do this rewrite?

Mytherin (commented):

src/optimizer/rule/comparison_simplification.cpp
dentiny: Hi Mark, I had a try to avoid using bound column ref, but it failed with
```sql
test/fuzzer/pedro/strptime_null_argument.test:5
Error: Query unexpectedly failed! (test/fuzzer/pedro/strptime_null_argument.test:5)!
SELECT strptime(NULL,'p') IS DISTINCT FROM DATE '-543572-1-1';
Conversion Error: Type INT32 with value 543573-01-01 (BC) can't be cast to the destination type INT64
LINE 1: SELECT strptime(NULL,'p') IS DISTINCT FROM DATE '-543572-1-1';
                                                   ^^^^^^^^^^^^^^^^^^
```

Checking more on the failed query, I found
```sql
memory D SELECT NULL::TIMESTAMP IS DISTINCT FROM CAST(DATE '-543572-1-1' AS TIMESTAMP);
┌───────────────────────────────────────────────────────────────────────────────────────────┐
│ (CAST(NULL AS TIMESTAMP) IS DISTINCT FROM CAST(CAST('-543572-1-1' AS DATE) AS TIMESTAMP)) │
│                                          boolean                                          │
├───────────────────────────────────────────────────────────────────────────────────────────┤
│ true                                                                                      │
└───────────────────────────────────────────────────────────────────────────────────────────┘

memory D SELECT CAST(DATE '-543572-1-1' AS TIMESTAMP) IS NOT NULL;
Conversion Error:
Type INT32 with value 543573-01-01 (BC) can't be cast to the destination type INT64

LINE 1: SELECT CAST(DATE '-543572-1-1' AS TIMESTAMP) IS NOT NULL;
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
So the thing is,
- For `IS DISTINCT FROM`, we don't seem to execute the unnecessary child, so the improper cast is skipped completely
- In this PR, if we allow all types of expression (not only bound column ref), the query becomes `CAST(DATE '-543572-1-1' AS TIMESTAMP) IS NOT NULL` -- child query will be executed first and leads to cast failure

It reads like a regression to me, so in the latest commit, I went a step back to take the most conservative approach to allow only bound column ref.

Mytherin:
Makes sense - thanks for investigating!

PR #24423, opened 2026-08-03
Fix monotonic clock usage

dentiny:
Followup PR for https://github.com/duckdb/duckdb/pull/23293

I found a few more use cases which don't use system clock and monotonic clock correctly.


Mytherin:
Thanks!

PR #24425, opened 2026-08-03
Reject invalid imported approx_quantile states

### LINKED ISSUE #24279: Bug4: Out-of-Bounds Vector Access in TDigest::process() via Crafted approx_quantile State Import
### What happens?

`ApproxQuantileImportState()` in `extension/core_functions/aggregate/holistic/approximate_quantile.cpp:250-288` builds a `TDigest` from an attacker-supplied AGGREGATE_STATE struct passed to `finalize()`/`combine()`. The import path sets `state.pos` from the struct's `count` field and populates `state.h`'s centroids from the struct's `centroids` list, with only a NULL check and no cross-validation that `count` actually matches the number of centroids supplied. Every other code path in the TDigest implementation assumes `pos != 0` implies the digest has at least one added value, and `ApproxQuantileScalarOperation::Finalize()` (lines 154-167) only checks `state.pos == 0` before unconditionally calling `state.h->compress()`, which calls `TDigest::process()` (`third_party/tdigest/t_digest.hpp:550-583`). `process()` unconditionally executes `unprocessed_[0]` after merging any processed centroids into the unprocessed buffer. With `count=1` (non-zero) and an empty `centroids` list, both buffers end up empty, so `std::vector::operator[]` dereferences an out-of-bounds/null internal storage pointer. A single `SELECT finalize(to_aggregate_state(...))` query using only the built-in `to_aggregate_state`/`finalize` scalar functions reaches this path with no file, ATTACH, or special privilege required.

### To Reproduce

```sql
-- vuln_001.sql
-- VULN #1: Out-of-bounds vector access in TDigest::process()
-- via crafted approx_quantile aggregate state import.
--
-- Trigger: a hand-crafted AGGREGATE_STATE struct with count=1 (non-zero)
-- but centroids=[] (empty list). When finalize() is called, TDigest::process()
-- assumes count > 0 implies at least one unprocessed centroid exists and
-- accesses unprocessed_[0] without checking, causing UB / OOB access.
SELECT finalize(
  to_aggregate_state(
    {'count': 1::UBIGINT, 'min': 0.0::DOUBLE, 'max': 0.0::DOUBLE,
     'centroids': []::STRUCT(mean DOUBLE, weight DOUBLE)[]},
    'approx_quantile', ['DOUBLE', 'FLOAT'], [NULL, 0.5::FLOAT]
  )
);
```

Trigger output:

```
/usr/include/c++/13/bits/stl_vector.h:1129:34: runtime error: reference binding to null pointer of type 'struct value_type'
    #0 0x5b6cb997a985 in std::vector<duckdb_tdigest::Centroid, ...>::operator[](unsigned long) /usr/include/c++/13/bits/stl_vector.h:1129
    #1 0x5b6cb996b168 in duckdb_tdigest::TDigest::process() .../third_party/tdigest/t_digest.hpp:561
    #2 0x5b6cb99674c9 in duckdb_tdigest::TDigest::compress() .../third_party/tdigest/t_digest.hpp:406
    #3 0x5b6cb96cc27f in Finalize<double, ...ApproxQuantileState> .../extension/core_functions/aggregate/holistic/approximate_quantile.cpp:162
    #4 0x5b6cb963a14a in Finalize<...ApproxQuantileState, double, ...ApproxQuantileScalarOperation> .../src/include/duckdb/common/vector_operations/aggregate_executor.hpp:969
    #5 0x5b6cb95e642f in StateFinalize<...> .../src/include/duckdb/function/aggregate_function.hpp:745
    #6 0x5b6cb56fa076 in AggregateStateFinalize .../src/function/scalar/system/aggregate_export.cpp:428
=== EXIT CODE: 1 ===
CRASH DETECTED
```

The crash is a real UBSan/ASan-reported null-pointer/OOB access reached through `Finalize` -> `TDigest::compress` -> `TDigest::process`, confirming the invalid vector access is real and not merely caught by an assertion.

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

u70b3:
This PR validates imported `approx_quantile` aggregate states before constructing the TDigest.

  A state with a non-zero count and an empty centroid list currently reaches TDigest finalization and triggers an INTERNAL error when the digest accesses
  its first centroid. This change rejects that invalid combination at the import boundary with an `InvalidInputException`.

  The regression test constructs the invalid state through `to_aggregate_state` and verifies that it returns a user-facing input error instead of an
  INTERNAL error.

  Fixes #24279.

  Tests:
  - `make reldebug`
  - `build/reldebug/test/unittest test/sql/aggregate/aggregate_state_export/approx_quantile_export_state.test`
  - `build/reldebug/test/unittest 'test/sql/aggregate/aggregate_state_export/*'`

### DIFF
--- a/extension/core_functions/aggregate/holistic/approximate_quantile.cpp
+++ b/extension/core_functions/aggregate/holistic/approximate_quantile.cpp
@@ -269,6 +269,10 @@ void ApproxQuantileImportState(AggregateImportInputData &input) {
 		if (!count_entry.IsValid() || !min_entry.IsValid() || !max_entry.IsValid() || !centroid_list.IsValid()) {
 			throw InvalidInputException("Invalid approx_quantile state - the state fields cannot be NULL");
 		}
+		if (count_entry.GetValue() != 0 && centroid_list.GetListLength() == 0) {
+			throw InvalidInputException(
+			    "Invalid approx_quantile state - non-zero count requires at least one centroid");
+		}
 		arena_vector<duckdb_tdigest::Centroid> centroids(input.allocator);
 		centroids.reserve(centroid_list.GetListLength());
 		for (const auto centroid_entry : centroid_list.GetChildValues()) {
--- a/test/sql/aggregate/aggregate_state_export/approx_quantile_export_state.test
+++ b/test/sql/aggregate/aggregate_state_export/approx_quantile_export_state.test
@@ -85,3 +85,13 @@ query I
 SELECT finalize(combine(approx_quantile(i, 0.5) EXPORT_STATE, (SELECT approx_quantile(j, 0.5) EXPORT_STATE FROM range(500, 1001) s(j)))) FROM range(1, 500) t(i);
 ----
 500
+
+# a non-empty state must contain at least one centroid
+statement error
+SELECT finalize(to_aggregate_state(
+    {'count': 1::UBIGINT, 'min': 0.0::DOUBLE, 'max': 0.0::DOUBLE,
+     'centroids': []::STRUCT(mean DOUBLE, weight DOUBLE)[]},
+    'approx_quantile', ['DOUBLE', 'FLOAT'], [NULL, 0.5::FLOAT]
+));
+----
+Invalid Input Error: Invalid approx_quantile state - non-zero count requires at least one centroid


Mytherin:
Thanks!

PR #22167, opened 2026-04-20
Fix VARIANT + ART-index pushdown-extract: uninitialized memory read in FetchRow

thmsmlr:
## Summary

`VariantColumnData::FetchRow` (unshredded branch) forwarded the caller's pushdown-extract `StorageIndex` straight to the inner `StructColumnData::FetchRow`. That `StorageIndex`'s child is a VARIANT-extract path (e.g. field-based, `field=\"duration_ms\"`, `has_index=false`, `index` uninitialized). `StructColumnData::FetchRow`'s pushdown branch then called `GetPrimaryIndex()` on that child — which only has `D_ASSERT(has_index)` and returns the uninitialized `index` in release builds — and indexed into its `sub_columns` vector. Since the unshredded struct has 4 sub-columns (keys / children / values / data), the resulting out-of-bounds dereference surfaces as:

```
INTERNAL Error: Attempted to access index <raw-pointer-address> within vector of size 4
```

## Reproduction

```sql
CREATE TABLE events(ts TIMESTAMP, name VARCHAR, attrs VARIANT);
CREATE INDEX idx_events_name ON events(name);
INSERT INTO events VALUES
  (now()::TIMESTAMP, 'http.request', CAST({'duration_ms': 100} AS VARIANT)),
  (now()::TIMESTAMP, 'http.request', CAST({'duration_ms': 200} AS VARIANT));
SELECT attrs['duration_ms']::DOUBLE FROM events WHERE name = 'http.request';
```

Triggered only when the unshredded branch is hit (freshly-inserted rows, no `CHECKPOINT`) and the scan reaches `FetchRow` via the ART-index pushdown path. Existing VARIANT tests checkpoint (triggering shredding), so they miss this path.

## Fix

Pass a fresh full-read `StorageIndex(0)` to the inner `FetchRow`. The outer code already runs `VariantUtils::VariantExtract` against the fully-materialized VARIANT afterward, which is the correct path for the unshredded representation. One-line semantic change.

## Related hardening (not in this PR)

[...]

PR #24428, opened 2026-08-03
Fix VARIANT field extraction in index fetch

### LINKED ISSUE #24064: index scan over VARIANT column crashes with internal error
Hello team, I recently encountered a bug using DuckDB. I switched all JSON columns to VARIANT and the DB started crashing for some of my tables. It appears the issue is a combination of: having a PK, using VARIANT, and doing an exact WHERE lookup.

AI generated report, reviewed by myself first:

### What happens?

Selecting a VARIANT field from rows located via an ART index scan (an explicit `CREATE INDEX`, or a `PRIMARY KEY` point lookup, which is how I originally hit this) is broken in both storage representations:

1. **Unshredded (before `CHECKPOINT`):** the query aborts with `INTERNAL Error: Attempted to access index <huge number> within vector of size 4`. The huge number varies per run (uninitialized memory). The error also invalidates the database: every subsequent statement fails with `FATAL Error: database has been invalidated because of a previous fatal error`.
2. **Shredded (after `CHECKPOINT`, fresh connection):** the query succeeds but returns `NULL` for every row instead of the actual values, i.e. silent wrong results.

The same query without the index (full scan) returns correct results. Reproduced on v1.5.4, v1.5.5, and 1.6.0.dev291.

The root cause of the crash flavor was diagnosed in PR #22167 (April 2026): `VariantColumnData::FetchRow` forwards the caller's pushdown-extract `StorageIndex` (field-path child, `has_index=false`, `index` uninitialized) into `StructColumnData::FetchRow`, which calls `GetPrimaryIndex()` on it (only a `D_ASSERT` in release builds) and indexes out of bounds into the 4 sub-columns of the unshredded struct (keys/children/values/data). That PR was closed by its author and then by the stale bot, although a maintainer reproduced the failing test in the thread; as far as I can tell no open issue tracks the bug, hence this report. The shredded/NULL failure mode suggests the one-line fix proposed there may only cure the crash, not the wrong results after checkpointing.

### To Reproduce

Crash (in-memory, CLI or any client):

```sql
CREATE TABLE events(ts TIMESTAMP, ev VARCHAR, attrs VARIANT);
CREATE INDEX idx ON events(ev);
INSERT INTO events VALUES
  (now()::TIMESTAMP, 'http.request', CAST({'duration_ms': 100} AS VARIANT)),
  (now()::TIMESTAMP, 'http.request', CAST({'duration_ms': 200} AS VARIANT));
SELECT attrs['duration_ms']::DOUBLE FROM events WHERE ev = 'http.request';
-- INTERNAL Error: Attempted to access index 34589214840 within vector of size 4
-- expected: 100.0, 200.0 (dropping the CREATE INDEX gives the correct result)
```

Silent NULLs (file-backed, checkpoint, reopen; Python):

```python
import duckdb
cfg = {"storage_compatibility_version": "v1.5.0"}
con = duckdb.connect("repro.duckdb", config=cfg)
con.execute("CREATE TABLE events(ts TIMESTAMP, ev VARCHAR, attrs VARIANT)")
con.execute("CREATE INDEX idx ON events(ev)")
con.execute("""INSERT INTO events VALUES
  (now()::TIMESTAMP, 'http.request', CAST({'duration_ms': 100} AS VARIANT)),
  (now()::TIMESTAMP, 'http.request', CAST({'duration_ms': 200} AS VARIANT))""")
con.execute("CHECKPOINT")
con.close()
con = duckdb.connect("repro.duckdb", config=cfg)
print(con.execute("SELECT attrs['duration_ms']::DOUBLE FROM events WHERE ev = 'http.request'").fetchall())
# [(None,), (None,)] but expected [(100.0,), (200.0,)]
```

### OS:

macOS 26.5.2 (arm64)

### DuckDB Version:

v1.5.5 (also reproduced on v1.5.4 and 1.6.0.dev291)

### DuckDB Client:

Python

### Hardware:

Apple M5 Pro, 64 GB

### Full Name:

Alexandru Rosianu

### Affiliation:

Molin AI Ltd

- [x] Yes, I have

- [x] Yes, I have

u70b3:
This PR fixes VARIANT field extraction when rows are fetched through an ART index.

  For unshredded VARIANT data, `VariantColumnData::FetchRows` forwarded the caller's field-based `StorageIndex` to the internal struct containing the
  VARIANT representation. `StructColumnData::FetchRows` expects positional child indexes for that struct, which could result in an INTERNAL error or
  incorrect NULL results.

  The unshredded branch now performs a full read of the internal VARIANT value and leaves field extraction and casting to the existing `VariantExtract`
  path.

  The checkpointed reproduction in the issue remains unshredded because the row group is below the default shredding threshold. Therefore, both reported
  behaviors share the same root cause.

  Regression tests cover field extraction through a primary-key lookup for both transient and checkpointed persisted-unshredded data while retaining the
  existing full VARIANT projection checks.

  This adapts the original diagnosis from #22167 to the current `FetchRows` implementation.

  Fixes #24064.

  Tests:
  - `make reldebug`
  - `build/reldebug/test/unittest test/sql/storage/types/variant/index_fetch.test`
  - `build/reldebug/test/unittest 'test/sql/storage/types/variant/*'`
  - `build/reldebug/test/unittest 'test/sql/function/variant/*'`
  - `build/reldebug/test/unittest 'test/sql/types/variant/*'`

### DIFF
--- a/src/storage/table/variant_column_data.cpp
+++ b/src/storage/table/variant_column_data.cpp
@@ -467,8 +467,9 @@ void VariantColumnData::FetchRows(TransactionData transaction, ColumnFetchState 
 			VariantUtils::UnshredVariantData(intermediate, unshredded, 1);
 			variant_vec.SetValue(0, unshredded.GetValue(0));
 		} else {
-			sub_columns[0]->FetchRows(transaction, state, storage_index, &offset,
-			                          *FlatVector::IncrementalSelectionVector(), /*count=*/1, variant_vec, 0);
+			StorageIndex full_read(0);
+			sub_columns[0]->FetchRows(transaction, state, full_read, &offset, *FlatVector::IncrementalSelectionVector(),
+			                          /*count=*/1, variant_vec, 0);
 		}
 
 		if (!storage_index.IsPushdownExtract()) {
--- a/test/sql/storage/types/variant/index_fetch.test
+++ b/test/sql/storage/types/variant/index_fetch.test
@@ -12,7 +12,12 @@ SELECT v FROM tbl WHERE i=42;
 ----
 {'a': 42, 'b': 2}
 
-# Now test the same with a persistent db, checkpointed (shredded VARIANT)
+query I
+SELECT v['a']::INTEGER FROM tbl WHERE i=42;
+----
+42
+
+# Now test the same with a persistent DB. This row group stays unshredded because it is below the default threshold.
 
 statement ok
 ATTACH '{TEST_DIR}/variant_index_fetch.db' as db2 (STORAGE_VERSION 'v1.5.0');
@@ -33,6 +38,11 @@ query I
 SELECT v FROM tbl WHERE i=42;
 ----
 {'a': 42, 'b': 2}
+
+query I
+SELECT v['a']::INTEGER FROM tbl WHERE i=42;
+----
+42
 
 statement ok
 SET debug_force_fetch_row=true;


lnkuiper:
Thanks!

PR #24429, opened 2026-08-03
Fix checkpoint Analyze handling for explicit compression method

Liuzw1998:
**Problem**

Columns declared with USING COMPRESSION skipped the checkpoint Analyze scan because codec selection was fixed. Analyze also prepares codec-specific state: validity changes must be scanned, ZSTD needs the exact tuple count for its layout, and Dictionary must inspect values before falling back to Uncompressed.

**Fix**

Skip only the base-column scan for supported non-nested columns. Analyze validity separately, initialize ZSTD layout from the tuple count, and retain the full scan for input-dependent codecs.

**Testing**

Cover RLE validity updates, ZSTD layout, Dictionary fallback, and persistence after restart.

### DIFF
--- a/src/storage/table/column_data_checkpointer.cpp
+++ b/src/storage/table/column_data_checkpointer.cpp
@@ -190,32 +190,25 @@ vector<CheckpointAnalyzeResult> ColumnDataCheckpointer::DetectBestCompressionMet
 	}
 
 	InitAnalyze();
+	// scan over all the segments and run the analyze step
+	ScanSegments([&](Vector &scan_vector) {
+		for (idx_t i = 0; i < checkpoint_states.size(); i++) {
+			auto &functions = compression_functions[i];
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - but instead of adding extra code to the compression codecs maybe we can just run the analyze after all (but only for the selected compression methods)? Or maybe we can add a flag to the compression codec like `ALWAYS_REQUIRES_ANALYZE` or `DOES_NOT_REQUIRE_ANALYZE` that controls this behavior?

Liuzw1998:
Hi Mytherin, thanks for the suggestion. I go with the first approach.

My old implementation also includes a minor optimization that avoids scanning the values if ZSTD is specified. I leave that optimization out of this PR to keep the fix focused.

Mytherin:
Thanks!

PR #24433, opened 2026-08-03
[MultiFileReader] Fail if two incompatible nested types are detected in the `MultiFileColumnMapper`

### LINKED ISSUE #1270: INTERNAL Error "Vector::Reference used on vector of different type (source VARCHAR[] referenced STRUCT(list VARCHAR))" in IcebergMultiFileReader::FinalizeChunk — list-encoding drift across files
## Versions / environment
- DuckDB 1.5.5 (Python client), macOS arm64
- iceberg extension `45163a28` · httpfs `827222f` · aws `efa54a9`
- Table: AWS Security Lake native source `CLOUD_TRAIL_MGMT` v2.0 — Glue-managed Iceberg on S3,
  data files written continuously by Security Lake over months (writer-version drift across files),
  read via `iceberg_scan('s3://…/metadata/<uuid>.metadata.json')` (catalog-free mode)

## Symptom
Projecting the whole table (`SELECT *`, `to_json(t)`) or the `metadata` struct column crashes:

```
INTERNAL Error: Vector::Reference used on vector of different type
(source VARCHAR[] referenced STRUCT(list VARCHAR))
```

Stack (abridged):
```
Vector::Reference
(anon)::RemapChildVectors
(anon)::RemapNested
(anon)::RemapStructFunction
ExpressionExecutor::Execute
MultiFileReader::FinalizeChunk
IcebergMultiFileReader::FinalizeChunk
MultiFileFunction<ParquetMultiFileInfo>::MultiFileScan
```

The struct's current logical type (via `DESCRIBE`):

```
STRUCT(product STRUCT("version" VARCHAR, "name" VARCHAR, vendor_name VARCHAR,
       feature STRUCT("name" VARCHAR)),
       event_code VARCHAR, uid VARCHAR, profiles VARCHAR[], "version" VARCHAR)
```

The error's type pair (`VARCHAR[]` vs `STRUCT("list" VARCHAR)`) points at `metadata.profiles`:
older Security-Lake-written files evidently encode the list such that the reader surfaces it as a
single-field struct named `list` (parquet 3-level list wrapper surfaced literally), while newer
files/the table schema read as a proper `VARCHAR[]`. The multi-file remap path `Reference`s the
child vector rather than converting between the two shapes.

## Aggravator
The `InternalException` **invalidates the whole DuckDB instance** — every subsequent query on any
connection of that instance fails with `FATAL Error: … database has been invalidated because of a
previous fatal error`. For long-lived embedded readers this turns one bad file into a full restart.

## What still works (workarounds we use)
- Explicit column lists excluding the affected struct — every other column reads fine, including
  other nested structs and nested-field *filters* (for what it's worth, we could not reproduce
  #760's nested-filter-pushdown crash on this build).
- **Child projection works**: `SELECT metadata.uid, metadata.event_code …` is fine — only
  materializing the whole parent struct crashes.

## Repro
We cannot share the table (security telemetry), but any Security Lake `cloud_trail_mgmt` 2.0 table
whose files span several months of SL writer versions should reproduce with `SELECT * … LIMIT 1`.
Synthetic shape: an Iceberg table with a `list<string>` field inside a struct, where data files
disagree on the parquet list element encoding/naming across commits; project the parent struct.
Happy to run diagnostics against our table on request.

Tishj:
This PR fixes https://github.com/duckdb/duckdb-iceberg/issues/1270

We blindly trusted that the type of the local column matched that of the global column, resulting in a crash later on during execution.


pdet:
Thanks!

PR #24434, opened 2026-08-03
Skip unnecessary payloads during WAL checkpoint prescan

p1p1bear:
This PR avoids deserializing payloads that are not required during the first WAL pass, which only scans for checkpoint state.

For framed WAL v2/v3 entries, the entry has already been read and integrity-checked. The prescan now only deserializes payloads that update its state, such as WAL version, checkpoint, and row-group data entries. Other entries jump directly to the root terminator. The actual replay pass is unchanged, and WAL v1 entries are still fully deserialized because they do not have entry framing.

This only skips payload deserialization; WAL reading, checksum verification, and decryption still happen as before.

### DIFF
--- a/src/storage/wal_replay.cpp
+++ b/src/storage/wal_replay.cpp
@@ -217,7 +217,13 @@ public:
 			deserializer.End();
 			return true;
 		}
-		ReplayEntry(wal_type);
+		if (CanSkipPayload(wal_type)) {
+			// Framed entries have already been read and integrity-checked, so jump directly to the root terminator.
+			D_ASSERT(stream.GetCapacity() >= sizeof(field_id_t));
+			stream.SetPosition(stream.GetCapacity() - sizeof(field_id_t));
+		} else {
+			ReplayEntry(wal_type);
+		}
 		deserializer.End();
 		return false;
 	}
@@ -230,6 +236,20 @@ public:
 
 protected:
 	void ReplayEntry(WALType wal_type);
+	bool CanSkipPayload(WALType wal_type) const {
+		if (!deserialize_only || !data) {
+			return false;
+		}
+		D_ASSERT(state.wal_version == 2 || state.wal_version == 3);
+		switch (wal_type) {
+		case WALType::INSERT_TUPLE:
+		case WALType::DELETE_TUPLE:
+		case WALType::UPDATE_TUPLE:
+			return true;
+		default:
+			return false;
+		}
+	}
 
 	void ReplayVersion();
 
--- a/src/storage/wal_replay.cpp
+++ b/src/storage/wal_replay.cpp
@@ -237,14 +237,36 @@ public:
 protected:
 	void ReplayEntry(WALType wal_type);
 	bool CanSkipPayload(WALType wal_type) const {
+		// Unframed WAL v1 entries must still be deserialized; Replay* guards suppress their side effects.
 		if (!deserialize_only || !data) {
 			return false;
 		}
 		D_ASSERT(state.wal_version == 2 || state.wal_version == 3);
 		switch (wal_type) {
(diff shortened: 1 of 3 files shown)


Mytherin (commented):
Thanks for the PR!

I wonder if we could just do this for all payloads except for the necessary ones (checkpoint payloads?) instead of only doing this for DML payloads? Then we can also remove a bunch of `DeserializeOnly()` code in the replay entries which will allow us to simplify code there.

p1p1bear:
> Thanks for the PR!
> 
> I wonder if we could just do this for all payloads except for the necessary ones (checkpoint payloads?) instead of only doing this for DML payloads? Then we can also remove a bunch of `DeserializeOnly()` code in the replay entries which will allow us to simplify code there.

@Mytherin   Thanks for the suggestion! I’ve extended the optimization to all unnecessary framed payloads. The skippable types use an explicit allowlist, so newly added WAL types are fully deserialized by default and cannot accidentally skip required prescan work.

I also tried centralizing the remaining DeserializeOnly() handling, but WAL v1 entries are unframed and must still be deserialized to locate the next entry. Removing the checks required splitting deserialization from replay across many handlers, including special cases such as indexes and ROW_GROUP_DATA, which made the change significantly larger. I therefore kept the existing checks as the legacy v1 path. Suggestions for a simpler way to unify this are welcome.

p1p1bear:
I ran a simple test for this PR.


The benchmark compares the original WAL prescan with the final payload-skipping implementation. Both binaries were built in Release mode with `-O3 -DNDEBUG` and used the same database and WAL files.

## Datasets

All DML tables use this schema:

```sql
CREATE TABLE t(id BIGINT, payload VARCHAR);
```

| Scenario | Dataset and operation | WAL size | Entries |
|---|---|---:|---:|
| INSERT | Insert 500,000 rows with 512-byte strings | 249.88 MiB | 250 |
| UPDATE | Update 500,000 512-byte strings | 249.88 MiB | 250 |
| DELETE | Delete 16,000,000 rows from a table containing 512-byte strings | 122.44 MiB | 7,818 |
| Mixed | Update 200,000, delete 100,000, and insert 200,000 rows | 200.67 MiB | 255 |
| ROW_GROUP_DATA | Optimistically insert 2,048,000 rows with 512-byte strings | 0.24 MiB | 6 |
| CREATE_INDEX | Create an ART index over 500,000 unique 512-byte strings | 403.34 MiB | 5 |

The mixed WAL contains 98 INSERT, 50 DELETE, and 98 UPDATE entries. INSERT and UPDATE each account for about 49.81% of the payload bytes, while DELETE accounts for 0.38% because DELETE entries only store row IDs.

## Results

| Scenario | Original | Final | Reduction | Speedup | Instruction reduction |
|---|---:|---:|---:|---:|---:|
| INSERT | 203.49 ms | 101.82 ms | 49.96% | 2.00x | 40.2% |
| UPDATE | 198.57 ms | 98.41 ms | 50.44% | 2.02x | 40.3% |
| DELETE | 56.71 ms | 50.29 ms | 11.31% | 1.13x | 13.4% |
| Mixed | 159.49 ms | 79.35 ms | 50.25% | 2.01x | 37.7% |
| ROW_GROUP_DATA | 2.103 ms | 2.078 ms | 1.18% | 1.01x | Approximately 0% |
| CREATE_INDEX | 370.95 ms | 268.28 ms | 27.68% | 1.38x | 24.5% |

Large INSERT, UPDATE, and mixed DML WALs show an approximately 2x faster prescan. DELETE improves less because its payload is inexpensive to deserialize. `ROW_GROUP_DATA` remains unchanged because the prescan must deserialize its block IDs and mark them as used.

Mytherin:
Thanks for the updates! I think this looks good as-is.

PR #24437, opened 2026-08-03
Improve the performance of timezone function

NiclasHaderer:
The ICU date functions that take a per-row time zone argument resolved the zone again for every row: a `string_t::GetString()` allocation, a lookup over the zone table, and then a fresh `TimeZone` plus a calendar to apply it to. This PR adds a thread-local cache keyed on the time zone name, so each distinct zone is resolved once per thread, plus a one-entry fast path for the common case of consecutive rows sharing a zone.

The map is currently unbounded and different casings of the same timezone could in theory cause it to grow pretty big, but I feel like it is very unlikely that people have so many different cases of the same timezone in the same table.

<details>
<summary>Some benchmarks</summary>

`make reldebug`, 14 cores, best of 3 runs of `SELECT bit_xor(hash(timezone(park_timezone, ts))) FROM bookings`.

2M rows, 5 distinct zones interleaved:

| threads | before | after | speedup |
|--------:|-------:|------:|--------:|
| 1  | 0.419s | 0.378s | 1.1x |
| 2  | 0.188s | 0.107s | 1.8x |
| 4  | 0.106s | 0.077s | 1.4x |
| 8  | 0.064s | 0.042s | 1.5x |
| 16 | 0.057s | 0.034s | 1.7x |

10M rows cycling through all 638 zones (zone changes every row):

| threads | before | after | speedup |
|--------:|-------:|------:|--------:|
| 1  | 2.508s | 1.464s | 1.7x |
| 2  | 1.257s | 0.710s | 1.8x |
| 8  | 0.373s | 0.205s | 1.8x |
| 16 | 0.297s | 0.122s | 2.4x |

100k rows, 638 zones:

| threads | before | after | speedup |
|--------:|-------:|------:|--------:|
| 1 | 0.019s | 0.010s | 1.9x |

</details>

### DIFF
--- a/extension/icu/icu-datefunc.cpp
+++ b/extension/icu/icu-datefunc.cpp
@@ -57,22 +57,67 @@ unique_ptr<FunctionData> ICUDateFunc::Bind(BindScalarFunctionInput &input) {
 	return make_uniq<BindData>(input.GetClientContext());
 }
 
-bool ICUDateFunc::TrySetTimeZone(Calendar *calendar, const string_t &tz_id) {
-	string tz_str = tz_id.GetString();
-	auto tz = ICUHelpers::TryGetTimeZone(tz_str);
-	if (!tz) {
+const ICUDateFunc::CalendarCacheState::CacheEntry &ICUDateFunc::CalendarCacheState::GetEntry(const string_t &tz_id) {
+	if (last && last->first.size() == tz_id.GetSize() &&
+	    memcmp(last->first.data(), tz_id.GetData(), tz_id.GetSize()) == 0) {
+		return last->second;
+	}
+	auto tz_str = tz_id.GetString();
+	auto entry = cache.find(tz_str);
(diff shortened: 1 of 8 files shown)


Mytherin:
Thanks for the PR!

I'm actually in the process of [removing ICU entirely](https://github.com/duckdb/duckdb/pull/24463) so this will have some conflicts there :) but after that PR is merged maybe some of this PR can be salvaged and applied there to still get some of the performance gains.

NiclasHaderer:
Oh nice, cool to see. Do you want to merge this in 1.5 regardless, or should I close it instead?

Mytherin:
I wouldn't consider this a bug-fix so this should target `main` anyway - if you think it's worthwhile you can pick it up post merge of the other PR.

NiclasHaderer:
I switched to main and the performance difference is still very measurable

PR #24459, opened 2026-08-03
unittest: allow require-env for passed-through variables

HendrikLambert:
`--env-passthrough` registers environment variables before the SQLLogic test is parsed. If the test also contains `require-env` for the same variable, the runner currently fails with:
```
Environment variable '<name>' has already been defined
```
This PR lets require-env accept variables already registered by --env-passthrough, so test frameworks can use passthrough without modifying existing SQLLogic tests.

Missing variables behave as before: passthrough fails at startup, while `require-env` skips the test.


lnkuiper:
Thanks!

PR #24469, opened 2026-08-03
Fix integer underflow in VariantMetadata

### LINKED ISSUE #24461: Vuln161: Integer underflow in VariantMetadata dictionary offset subtraction bypasses the bounds check and triggers an uncaught std::length_error
### What happens?

In VariantMetadata::VariantMetadata (extension/parquet/reader/variant/variant_binary_decoder.cpp), the dictionary string length is computed as `next_offset - last_offset` using unsigned idx_t arithmetic, with no check that next_offset >= last_offset. When a crafted metadata blob has next_offset < last_offset, the subtraction wraps around to a huge value, and because the subsequent bounds check `data_start + last_offset + string_size > metadata_buffer_capacity` mathematically cancels the underflow (modular arithmetic makes it equivalent to checking data_start + next_offset), the corrupted length passes the check undetected.

### PoC

```sql
SELECT variant_bytes_to_variant('\x01\x02\xC8\x05\x0A\x00\x00\x00\x00\x00'::BLOB);
```

### Result

Real captured output from the independent verification run:

```
Invalid Error: basic_string::_M_create
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

DinosL:
The variant metadata binary decoder could underflow when subtracting consecutive dictionary offsets, and because the bounds check that follows wraps around by the same amount, the corrupted length slipped through leading to an error.

fixes: https://github.com/duckdb/duckdb/issues/24461

### DIFF
--- a/extension/parquet/reader/variant/variant_binary_decoder.cpp
+++ b/extension/parquet/reader/variant/variant_binary_decoder.cpp
@@ -80,6 +80,10 @@ VariantMetadata::VariantMetadata(const string_t &metadata) : metadata(metadata) 
 	for (idx_t i = 0; i < dictionary_size; i++) {
 		auto next_offset = ReadVariableLengthLittleEndian(header.offset_size, metadata_data, metadata_offset,
 		                                                  metadata_buffer_capacity);
+		if (next_offset < last_offset) {
+			throw IOException(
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #20157, opened 2025-12-12
Apply LRU cache to parquet metadata

dentiny:
Closes https://github.com/duckdb/duckdb-rs/issues/476

Currently parquet metadata cache entries are stored in a per-instance global [object cache](https://github.com/duckdb/duckdb/blob/main/src/include/duckdb/storage/object_cache.hpp), along with several other types of caches (i.e., encryption keys). 
But unfortunately we don't apply any limitation on cache entries, in terms of entry counts or memory consumption, which lead to "memleak" under heavy parquet usage.
This PR replaces the plain unordered map inside of the object cache with LRU cache (capped by overall estimated memory consumption), which is a non-regression compared with its current status.

PR #24472, opened 2026-08-03
Optimize LRU cache

dentiny:
While working on https://github.com/duckdb/duckdb/pull/24177, I found the LRU cache key is copied for multiple times.

In this PR, I save one copy for all keys: the LRU cache impl consists of two STLs, one list and one map, because the STL list always allocates a new and separate memory, which doesn't affect when we add a new entry or remove one, we could use the reference as the map hash key.

The impl was first introduced in my initial LRU cache implementation [here](https://github.com/duckdb/duckdb/pull/20157#discussion_r2618483414), but I reverted some of the optimization when I was debugging on non-reproducible segfault on windows.

I have ran CI on my own fork here: https://github.com/dentiny/duckdb/pull/157

### DIFF
--- a/src/include/duckdb/common/lru_cache.hpp
+++ b/src/include/duckdb/common/lru_cache.hpp
@@ -65,14 +65,14 @@ public:
 		}
 
 		// Add new entry
-		lru_list.emplace_front(key);
+		lru_list.emplace_front(std::move(key));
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24483, opened 2026-08-03
Scan the full child span for Arrow ListView arrays

fallintoplace:
## What problem does this solve?

Arrow ListView arrays store an offset and a size for each row. Unlike a regular ListArray, those ranges may be out of order, disjoint, or overlapping.

DuckDB currently adds all row sizes together and uses that sum as the length of the child scan. That only works when the referenced ranges form one tightly packed block. For example, offsets [4, 0] with sizes [3, 2] reference child[4:7] and child[0:2]. The required child span is child[0:7], but the current code scans only five values while the first list entry still points at offset four.

## What changed?

Compute the child span from the minimum referenced offset to the maximum offset plus size, then rebase each entry against the start of that span. Empty entries get an offset of zero. The existing 32-bit and 64-bit ListView paths use the same corrected calculation.

## Tests

The new C Data Interface tests cover:

- disjoint, out-of-order ranges with 32-bit offsets
- overlapping ranges with 64-bit offsets

Locally verified with:

- build/reldebug/test/unittest "Arrow scan of ListView*"
- build/reldebug/test/unittest "[arrow]"

No API changes.

### DIFF
--- a/src/function/table/arrow_conversion.cpp
+++ b/src/function/table/arrow_conversion.cpp
@@ -156,24 +156,31 @@ static ArrowListOffsetData ConvertArrowListViewOffsetsTemplated(Vector &vector, 
 	// for that reason we need to keep track of the lowest offset, so we can skip all the data that comes before it
 	// when we scan the child data
 
-	auto lowest_offset = size ? offsets[0] : 0;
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #24484, opened 2026-08-04
Fix stale projection ids in LogicalGet after pushdown extract

### LINKED ISSUE #24252: TopN + late materialization crashes on STRUCT field filter with multi-column projection
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t(id INTEGER, s STRUCT(a INTEGER, b VARCHAR), x INTEGER);
memory D INSERT INTO t VALUES
           (1, {'a': 1, 'b': 'x'}, 10),
           (2, {'a': 2, 'b': 'y'}, 20);
memory D 
memory D SELECT s, x
         FROM t
         WHERE s.a = 1
         ORDER BY id
         LIMIT 1;
INTERNAL Error:
Attempted to access index 3 within vector of size 3
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

gulugulubing:
fixes: #24252 

The RHS scan of the late materialization rewrites the struct filter into a pushdown-extract column and prunes column x, but it early returns before rebuilds the projection ids.

The fix is straightforward: return after rebuilds the projection ids.

### DIFF
--- a/src/optimizer/remove_unused_columns.cpp
+++ b/src/optimizer/remove_unused_columns.cpp
@@ -1102,30 +1102,29 @@ void RemoveUnusedColumns::RemoveColumnsFromLogicalGet(LogicalGet &get, unique_pt
 		get.table_filters = std::move(remapped_filters);
 	}
 
-	if (!get.function.filter_prune) {
-		return;
-	}
-	// Now set the projection cols by matching the "selection vector" that excludes filter columns
-	// with the "selection vector" that includes filter columns
-	idx_t col_idx = 0;
-	get.projection_ids.clear();
-	vector<ProjectionIndex> filtered_original_ids;
-	//! Find matching indices between the proj_sel and the col_sel
-	for (auto to_keep : proj_sel) {
-		for (; col_idx < col_sel.size(); col_idx++) {
-			if (to_keep == col_sel[col_idx]) {
-				filtered_original_ids.push_back(to_keep);
-				break;
+	if (get.function.filter_prune) {
+		// Now set the projection cols by matching the "selection vector" that excludes filter columns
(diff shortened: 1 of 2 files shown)


lnkuiper:
This throws an internal exception:
```sql
CREATE TABLE t(s STRUCT(a INTEGER, b INTEGER), x INTEGER);
INSERT INTO t VALUES (row(1, 11), 10), (row(2, 22), 20);
SELECT x FROM t WHERE s.a + s.b = 12;
```

Could we guard only the `projection_ids` rebuild with `filter_prune`, while keeping the filter re-push outside that?

gulugulubing:
> This throws an internal exception:
> 
> ```sql
> CREATE TABLE t(s STRUCT(a INTEGER, b INTEGER), x INTEGER);
> INSERT INTO t VALUES (row(1, 11), 10), (row(2, 22), 20);
> SELECT x FROM t WHERE s.a + s.b = 12;
> ```
Ah! The exception is caused by: the filter referencing extracted struct fields cannot be pushed back into the scan — it references two extracted fields, so it stays as a LogicalFilter on top of the scan, whose output only contains the projected column(s), and the filter fails to bind its column references. Fixed it by falling back to outputting all columns in scan order by clearing projection_ids. Thanks

> Could we guard only the `projection_ids` rebuild with `filter_prune`, while keeping the filter re-push outside that?

Sure — the projection_ids rebuild is now guarded by filter_prune, and the filter re-push stays outside the guard, as you suggested.

lnkuiper:
Thanks!

PR #24488, opened 2026-08-04
Fix reservoir_quantile for DECIMAL types (preserve precision and supp…

### LINKED ISSUE #24189: Bug2: reservoir_quantile() Silently Discards DECIMAL Type and Precision
### What happens?

`GetReservoirQuantileDecimalFunction` registers both a scalar and a list-quantile overload for `reservoir_quantile()` on DECIMAL columns, but both share the same bind callback `BindReservoirQuantileDecimal`, which unconditionally installs the scalar implementation via `ReplaceImplementation`, overwriting the bound `return_type` with a bare integer type and losing the DECIMAL width and scale. `CastToFunctionArguments` then casts the original DECIMAL argument down to that bare integer type before aggregation even starts, rounding away the fractional part. For the list-quantile overload the same scalar implementation is used at execution time, so only the first requested quantile is honored and the declared `LIST(DECIMAL)` return type is silently replaced by a scalar value.

### To Reproduce

SQL script (`vuln_001.sql`):

```sql
CREATE TABLE t(d DECIMAL(10,2));
INSERT INTO t VALUES (1.55);

-- Expected: result_type = DECIMAL(10,2), result ~= 1.55
-- Actual (bug): result_type is an integer type, result = 2 (rounded)
SELECT reservoir_quantile(d, 0.5) AS result, typeof(reservoir_quantile(d, 0.5)) AS result_type FROM t;

-- Expected: LIST(DECIMAL) result [~1.55, ~1.55, ~1.55]
-- Actual (bug): assertion failure in debug/assert builds, silent
-- truncation to a scalar in release builds
SELECT reservoir_quantile(d, [0.1, 0.5, 0.9]) FROM t;
```

Trigger output:

```
┌────────┬─────────────┐
│ result │ result_type │
│ int64  │   varchar   │
├────────┼─────────────┤
│      2 │ BIGINT      │
└────────┴─────────────┘
INTERNAL Error: Assertion triggered in file "extension/core_functions/aggregate/holistic/reservoir_quantile.cpp" on line 160: bind_data.quantiles.size() == 1
This error signals an assertion failure within DuckDB.
```

`DECIMAL(10,2)` 1.55 became `BIGINT` 2 with no error, confirming the precision loss, and the list overload hits the internal assertion confirming the list quantiles are never actually honored by the executed operation.

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

ptquocle:
…ort lists) 
reservoir_quantile was failing on DECIMAL columns because the aggregate function resolution missing physical integer types (INT16, INT32, INT64, INT128) that DuckDB uses to back decimals.
Added the physical type mappings to scalar/list quantile lookups and added a regression test covering decimal quantiles.
fixes #24189


lnkuiper:
Thanks for the PR! This breaks:
```sql
SET debug_verify_serializer=true;
SELECT reservoir_quantile(d, 0.5)
FROM (VALUES (1.55::DECIMAL(10,2))) t(d);
```
Could you have a look?

ptquocle:
Thanks for reviewing the PR. I'll take a look tomorrow and push a fix then

ptquocle:
hi @lnkuiper I have squashed the updates into a single commit

ptquocle:
Hi @lnkuiper I added the test path to verify_aggregate_state_export.json because
Not implemented Error:
Aggregate function "reservoir_quantile" does not have a state type callback defined - cannot export state

ptquocle:
Hi @lnkuiper, looks like the CI run hit an unrelated matrix flake on main. Could you help rerun the failed CI jobs? Thanks!

PR #24489, opened 2026-08-04
Check time bucker operation overflow

### LINKED ISSUE #24284: Bug10: Unchecked Signed Integer Overflow in TIME_BUCKET Origin Arithmetic
### What happens?

`TimeBucket::WidthConvertibleToMicrosCommon()` overflow-checks every intermediate subtraction via `SubtractOperatorOverflowCheck::Operation()` except the final step at line 109, `result_micros += origin_micros;`, which is a plain unchecked `int64_t` addition. With `bucket_width` an INTERVAL whose micro-equivalent is close to `INT64_MAX` (achievable with `days` near 106,751,991) and both `ts` and `origin` before the Unix epoch with `ts < origin`, the intermediate `result_micros` lands near `INT64_MIN`. Adding a sufficiently negative `origin_micros` then overflows below `INT64_MIN`, undefined behavior in C++ that wraps to a large positive value on two's-complement platforms. The wrapped value flows unchecked into `Timestamp::FromEpochMicroSeconds()`, producing a `timestamp_t` result unrelated to the mathematically correct bucket. Any client able to call `TIME_BUCKET` with an attacker-chosen wide INTERVAL and pre-epoch timestamps reaches this silently, with no exception raised.

### To Reproduce

```sql
-- vuln_001.sql
SELECT time_bucket(INTERVAL '106751991' DAY,
                    TIMESTAMP '1800-01-01 00:00:00',
                    TIMESTAMP '1900-01-01 00:00:00');
```

Trigger output:

```
[!] UBSAN runtime error (signed integer overflow) detected in stdout
/data/ylwang/Projects/ExtensionDB/Target/duckdb/extension/core_functions/scalar/date/time_bucket.cpp:109:17: runtime error: signed integer overflow: -9223372022400000000 + -2208988800000000 cannot be represented in type 'long int'
    #0 0x58562ce7b5f9 in WidthConvertibleToMicrosCommon .../extension/core_functions/scalar/date/time_bucket.cpp:109
    #1 0x58562cef196a in Operation<...> .../extension/core_functions/scalar/date/time_bucket.cpp:228
    #6 0x58562cecf649 in TimeBucketOriginFunction<...> .../extension/core_functions/scalar/date/time_bucket.cpp:360
```

UBSan reports a genuine signed integer overflow at the exact addition described (`-9223372022400000000 + -2208988800000000`), confirming the arithmetic wraps below `INT64_MIN` and the resulting bucket boundary is meaningless.

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24284

### DIFF
--- a/extension/core_functions/scalar/date/time_bucket.cpp
+++ b/extension/core_functions/scalar/date/time_bucket.cpp
@@ -1,4 +1,5 @@
 #include "duckdb/common/exception.hpp"
+#include "duckdb/common/operator/add.hpp"
 #include "duckdb/common/operator/cast_operators.hpp"
 #include "duckdb/common/operator/subtract.hpp"
 #include "duckdb/common/types/date.hpp"
@@ -106,7 +107,7 @@ struct TimeBucket {
 			result_micros =
 			    SubtractOperatorOverflowCheck::Operation<int64_t, int64_t, int64_t>(result_micros, bucket_width_micros);
 		}
-		result_micros += origin_micros;
+		result_micros = AddOperatorOverflowCheck::Operation<int64_t, int64_t, int64_t>(result_micros, origin_micros);
 
 		return Timestamp::FromEpochMicroSeconds(result_micros);
 	}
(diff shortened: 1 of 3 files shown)


lnkuiper:
Thanks!

PR #24491, opened 2026-08-04
add option to override connect display name

Tmonster:
Currently, if you run just `CONNECT` to connect to a postgres/rds/redshift instance, the display name on the CLI will default to the URI of the database.

The problem with that is that for certain services, this uri can be very ugly. This option gives extensions the ability to override the display name with something a little more user friendly.

Before
```
memory D connect 'redshift-attach-connect-test' (type redshift);
postgres://'redshift-attach-connect-test.ccg... D ... D 
```

After
```
memory D connect 'redshift-attach-connect-test' (type redshift);
redshift-attach-connect-test D  
```

### DIFF
--- a/src/main/attached_database.cpp
+++ b/src/main/attached_database.cpp
@@ -4,7 +4,9 @@
 #include "duckdb/catalog/duck_catalog.hpp"
 #include "duckdb/common/constants.hpp"
 #include "duckdb/common/enums/checkpoint_on_detach.hpp"
+#include "duckdb/common/exception/binder_exception.hpp"
 #include "duckdb/common/file_system.hpp"
+#include "duckdb/common/string_util.hpp"
 #include "duckdb/main/database.hpp"
 #include "duckdb/main/database_manager.hpp"
 #include "duckdb/main/settings.hpp"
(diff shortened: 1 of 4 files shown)


src/main/attached_database.cpp
```diff
@@ -176,6 +178,15 @@ AttachedDatabase::AttachedDatabase(DatabaseInstance &db, Catalog &catalog_p, Sto
...
 	catalog = storage_extension->attach(storage_info, context, *this, name.GetIdentifierName(), info, options);
+	// The attach function may rename the database: extensions that resolve the path themselves know a
+	// better name for it than the caller did. Adopt it before anything reads the name back.
+	if (info.name != name) {
+		if (NameIsReserved(info.name)) {
+			throw BinderException("Attached database name \"%s\" cannot be used because it is a reserved name",
+			                      info.name.GetIdentifierName());
+		}
+		SetName(info.name);
+	}
```
lnkuiper, line 189: Could we keep the display override separate from the `AttachedDatabase` name? `FinalizeAttach` uses `GetName()` as the database-map key, so using `info.name` here makes the label participate in catalog uniqueness.

Tmonster:
I actually think this might not be necessary here. This should be an option in the catalogs that are created, and they can override the `GetConnectDisplay()` function.

PR #24492, opened 2026-08-04
Only enable mold --icf=safe when the linker supports it

h4nsmuller:
### Problem
`CMakeLists.txt` unconditionally adds `-Wl,--icf=safe` for Release builds linking with mold (added in the "identical code folding" change). But `--icf=safe` was only introduced in **mold 1.1** — older mold, e.g. the **1.0.3** shipped in Ubuntu 22.04 and used by some extensions' GCC CI jobs, aborts *every* link with:

```
mold: unknown --icf argument: safe
collect2: error: ld returned 1 exit status
```

This fails the build at the very first target (`duckdb_platform_binary`), before anything else compiles. It surfaced when an extension bumped its bundled duckdb to a commit containing the ICF change while its CI still installs mold 1.0.3.

### Fix
Probe `mold --version` and only add `-Wl,--icf=safe` when mold is new enough (>= 1.1); otherwise skip ICF (and note it in the linker status message). No behavior change for CI/toolchains with a recent mold.


hannes:
CC @smvv

smvv (approved):
thank you!

PR #24494, opened 2026-08-04
Track and expose Pipeline's intra-MetaPipeline dependencies

jeewonhh:
Adds a helper to track Pipeline's depedencies to other pipelines within a MetaPipeline. Pipeline currently only exposes `dependencies` which are cross-MetaPipeline depedencies. The rest are in `MetaPipeline::pipeline_depedencies`. This PR mirrors `pipeline_depedencies` to `Pipeline::intra_dependencies`, so that the dependencies can outlive MetaPipeline, which goes out of scope after plan construction. This is needed by extensions that inspect pipelines at execution time.

### DIFF
--- a/src/include/duckdb/parallel/meta_pipeline.hpp
+++ b/src/include/duckdb/parallel/meta_pipeline.hpp
@@ -92,6 +92,10 @@ public:
 	                                      MetaPipelineType type = MetaPipelineType::REGULAR);
 
 private:
+	//! Register that 'dependant' depends on 'dependency', both within this MetaPipeline
+	void AddPipelineDependency(Pipeline &dependant, Pipeline &dependency);
+
+private:
 	//! The executor for all MetaPipelines in the query plan
 	Executor &executor;
 	//! The PipelineBuildState for all MetaPipelines in the query plan
--- a/src/include/duckdb/parallel/meta_pipeline.hpp
+++ b/src/include/duckdb/parallel/meta_pipeline.hpp
@@ -90,7 +90,8 @@ public:
 	                                      MetaPipelineType type = MetaPipelineType::REGULAR);
 
 private:
-	//! Register that 'dependant' depends on 'dependency', both within this MetaPipeline
+	//! Register that 'dependency' must have run before 'dependant' can start.
+	//! Both pipelines belong to this MetaPipeline, or to one of its descendants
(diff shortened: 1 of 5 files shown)


lnkuiper:
Thanks for the PR! One comment:

Could we avoid storing the intra-MetaPipeline dependencies twice? The scheduler reads `MetaPipeline::pipeline_dependencies`, while `GetAllDependencies()` reads the mirrored `Pipeline::intra_dependencies`. These are in sync now, but paths such as `CreateUnionPipeline` have to update both manually, so it is easy for them to diverge again.

I think the simplest shape is to make `Pipeline::intra_dependencies` the single source of truth. `MetaPipeline` would only add to it:

```cpp
void Pipeline::AddIntraDependency(Pipeline &dependency) {
	intra_dependencies.emplace_back(dependency.shared_from_this());
}

void MetaPipeline::AddPipelineDependency(Pipeline &dependant, Pipeline &dependency) {
	dependant.AddIntraDependency(dependency);
}
```

`Executor::ScheduleEventsInternal` can then consume the same vector instead of going through `MetaPipeline::GetDependencies()`:

```cpp
for (auto &entry : event_map) {
	auto &pipeline = entry.first.get();
	for (auto &weak_dependency : pipeline.intra_dependencies) {
		auto dependency = weak_dependency.lock();
		D_ASSERT(dependency);
		auto dependency_entry = event_map.find(*dependency);
		D_ASSERT(dependency_entry != event_map.end());
		entry.second.pipeline_event.AddDependency(dependency_entry->second.pipeline_event);
	}
}
```

This lets us remove `MetaPipeline::pipeline_dependencies`. The edges can remain `weak_ptr<Pipeline>` because the executor already keeps all pipelines alive for the query; we do not need a shared dependency container. For unions, maybe add a small `InheritDependencies` helper that copies both the intra- and cross-MetaPipeline dependency vectors.

jeewonhh:
@lnkuiper address the comment:)

lnkuiper:
Thanks for the changes! Any reason for targeting v1.5? I would prefer rebasing on main

jeewonhh:
> Thanks for the changes! Any reason for targeting v1.5? I would prefer rebasing on main

Would be nice if motherduck could cherrypick this easily to v1.5.x versions. I can make a separate PR targeting main as well.

lnkuiper:
Ok, looks good to me, thanks

PR #24496, opened 2026-08-04
Compress FSST input directly instead of copying it in 511-byte chunks

gropaul:
Hi dear DuckDB Team, 

I would propose a change to how DuckDB compresses data when using FSST, as substring search on FSST-compressed data such as https://db.in.tum.de/~pop/fsstlike.pdf relies on FSST's greedy longest-match property.

`compressBulk` broke that by copying each string into a 511-byte buffer and terminating every chunk, which makes the encoding offset-dependent. This offset is also hard to find later when having only compressed data.

Compress straight from the input, stopping 8 bytes short of the end, then compress only the unconsumed tail (at most 8 bytes) plus the terminator. Compression consumes up to 8 bytes per position, so it stops in `[len-8, len-1]`, not reading out of bounds of the string. Then we copy the tail into a buffer that has 8 bytes of extra space and compress it as well.

**Example** 
Symbols `{ab, cd, abcd, efghij}`, input = 508 filler bytes + `abcdefghij`, so `ab` is cut of.

```
xxxxabcdefghij
      | 510 byte offset --> So would have been compressed in two blocks, xxxxab and cdefghij


old (chunked):  ... ab  cd efghij  4 codes 
greedy / new:   ... abcd efghij    2 codes
```

The output now differs from the old encoder for strings >511 bytes, but decompression is unaffected. 

I have a prototype for compressed LIKE for DuckDB, but would like to get this merged early before the storage change to 2.0

### DIFF
--- a/benchmark/micro/compression/fsst/fsst_late_decompression.benchmark
+++ b/benchmark/micro/compression/fsst/fsst_late_decompression.benchmark
@@ -4,7 +4,7 @@
 
 name fsst late decompression benefit
 group fsst
-storage persistent
+storage persistent v1.2.0
 
 load
 DROP TABLE IF EXISTS test;
--- a/benchmark/micro/compression/fsst/fsst_read.benchmark
+++ b/benchmark/micro/compression/fsst/fsst_read.benchmark
@@ -4,7 +4,7 @@
 
 name fsst Compression Scan
 group fsst
-storage persistent
+storage persistent v1.2.0
 
 load
 DROP TABLE IF EXISTS test;
--- a/benchmark/micro/compression/fsst/fsst_read_worst_case.benchmark
+++ b/benchmark/micro/compression/fsst/fsst_read_worst_case.benchmark
@@ -4,7 +4,7 @@
 
 name fsst Compression Scan
 group aggregate
-storage persistent
+storage persistent v1.2.0
 
 load
 DROP TABLE IF EXISTS test;
--- a/benchmark/micro/compression/fsst/fsst_store.benchmark
+++ b/benchmark/micro/compression/fsst/fsst_store.benchmark
@@ -4,7 +4,7 @@
 
 name fsst Compression Write
 group aggregate
-storage persistent
+storage persistent v1.2.0
 require_reinit
 
 load
--- a/benchmark/micro/compression/fsst/fsst_store_worst_case.benchmark
+++ b/benchmark/micro/compression/fsst/fsst_store_worst_case.benchmark
@@ -4,7 +4,7 @@
 
 name name fsst Compression Write
 group fsst
-storage persistent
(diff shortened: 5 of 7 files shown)


lnkuiper:
I don't think this has any implications on 2.0 storage, so that's fine. Can we add a test?
```sql
statement ok
SET force_compression = 'fsst';

statement ok
ATTACH '{TEST_DIR}/fsst_greedy.db' AS db
    (BLOCK_SIZE 262144, STORAGE_VERSION 'v1.2.0');

statement ok
CREATE TABLE db.t AS
SELECT repeat('x', 510) || 'abcdefghij' || lpad(i::VARCHAR, 8, '0') AS s
FROM range(3500) t(i);

statement ok
CHECKPOINT db;

query I
SELECT count(*)
FROM pragma_storage_info('db.t')
WHERE column_name = 's'
  AND compression = 'FSST';
----
1
```
This tests that there is only 1 FSST segment.

gropaul:
Hi @lnkuiper, I can add the test, but I am not sure that it would touch my change. The number of segments will be the same before and after.

My plan with this PR is that if you have a single string that is longer than 511 characters, it would be compressed with FSST into two chunks. First [0, 511] and then [511, len]. Basically, from an FSST compression view, it compresses the two chunks separately, so there can't be a symbol that goes across this range. 

Now, for the `compressed-like` to be efficient, we need the same "take the longest symbol from the symbol table that fits" algorithm at every position in the string. If we have these chunks, this is no longer the case, as e.g. if we are at pos 510 we can only have a one-char symbol. 

This affects both `FSST` and `FSST_DICT` compression, because both use `compressBulk`. But the change will not affect decoding, so it is backward- and forward-compatible.

lnkuiper:
The test fails on the base branch by returning 2, and succeeds on this PR by returning 1. So clearly something has changed there. Does it make sense to test so we won't regress on this? Especially since you're planning to rely on this in future work?

gropaul:
Oh yes, you are right, sorry! I did not expect this to have such an impact on the compression rate. But with the change we can fit 3512 strings into the segment; before that, only 3332. 

I added the test! 
```
SELECT segment_id, count
FROM pragma_storage_info('db.t')
WHERE column_name = 's'
AND compression = 'FSST';
-- before change
┌────────────┬───────┐
│ segment_id │ count │
│   int64    │ int64 │
├────────────┼───────┤
│          0 │  3332 │
│          1 │   168 │
└────────────┴───────┘
-- after change
┌────────────┬───────┐
│ segment_id │ count │
│   int64    │ int64 │
├────────────┼───────┤
│          0 │  3500 │
└────────────┴───────┘
```

lnkuiper:
Thanks for adding the test! LGTM now

gropaul:
Thanks for the review! :)

PR #24497, opened 2026-08-04
Fix FIRST_VALUE streaming eligibility for framed windows

### LINKED ISSUE #24458: FIRST_VALUE without window ORDER BY ignores a singleton ROWS frame
### What happens?

The window query returns a result that differs from the expected/reference result.

#### Code

```sql
WITH t(id, v) AS (
  VALUES
    (1, 'a'),
    (2, 'b')
)
SELECT
  id,
  FIRST_VALUE(v) OVER (
  ) AS actual,
  v AS expected
FROM t
ORDER BY id;
```

Run:

```sh
```

#### Expected

| id | actual | expected |
| ---: | --- | --- |
| 1 | `a` | `a` |
| 2 | `b` | `b` |

#### Actual

| id | actual | expected |
| ---: | --- | --- |
| 1 | `a` | `a` |
| 2 | `a` | `b` |
#### Notes From Minimization

- `SET enable_optimizer=false` changes the output to the expected result.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

u70b3:
## Summary

  Restrict `FIRST_VALUE` streaming execution to window frames supported by the current streaming implementation.

  This preserves streaming for:

  - the default `OVER ()` frame
  - `ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`

  Other row-relative frames now fall back to the regular window operator.

  ## Root Cause

  The streaming implementation of `FIRST_VALUE` retains the first input value and reuses it for subsequent rows.

  However, its streaming eligibility callback previously accepted every `RESPECT NULLS` frame without checking whether the frame start moved. As a result, frames
  such as:

  ```sql
```
  incorrectly returned the first value of the partition for every row.

  ## Tests

  Added regression coverage for:

  - singleton frames
  - preceding frames
  - following frames
  - suffix frames
  - preservation of supported streaming fast paths

  Validated with:

  make reldebug
  build/reldebug/test/unittest test/sql/window/test_streaming_window.test

  The targeted test passes with 396 assertions.

  Fixes #24458

### DIFF
--- a/src/function/window/window_value_function.cpp
+++ b/src/function/window/window_value_function.cpp
@@ -784,7 +784,12 @@ struct WindowFirstValueExecutor : public WindowValueExecutor {
 			return wexpr.WindowStart() == WindowBoundary::UNBOUNDED_PRECEDING &&
 			       wexpr.WindowEnd() == WindowBoundary::CURRENT_ROW_ROWS;
 		}
-		return true;
+		// We can stream first values if the frame start is fixed at the partition start
+		return wexpr.WindowStart() == WindowBoundary::UNBOUNDED_PRECEDING &&
+		       (wexpr.WindowEnd() == WindowBoundary::CURRENT_ROW_ROWS ||
+		        wexpr.WindowEnd() == WindowBoundary::CURRENT_ROW_RANGE ||
+		        wexpr.WindowEnd() == WindowBoundary::CURRENT_ROW_GROUPS ||
+		        wexpr.WindowEnd() == WindowBoundary::UNBOUNDED_FOLLOWING);
 	}
 	static void StreamData(ExecutionContext &context, DataChunk &input, DataChunk &delayed, idx_t delayed_capacity,
 	                       Vector &result, LocalSourceState &state);
--- a/test/sql/window/test_streaming_window.test
+++ b/test/sql/window/test_streaming_window.test
@@ -285,6 +285,13 @@ explain select i, j, first_value(i) over (), first_value(j) over () from integer
 ----
 physical_plan	<REGEX>:.*STREAMING_WINDOW.*
 
+query TT
+EXPLAIN
+SELECT first_value(i) OVER (ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)
+FROM integers;
+----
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/function/window/window_value_function.cpp [hunk 1]
```diff
@@ -784,7 +784,10 @@ struct WindowFirstValueExecutor : public WindowValueExecutor {
 			return wexpr.WindowStart() == WindowBoundary::UNBOUNDED_PRECEDING &&
 			       wexpr.WindowEnd() == WindowBoundary::CURRENT_ROW_ROWS;
 		}
-		return true;
+		// We can stream first values if they are "running totals" or have the default frame
```
hawkfish: This also applies to `CURRENT_ROW_GROUPS` - can we add that too?

Also, now that I think about it, the unbounded following end frames should also be allowed.

test/sql/window/test_streaming_window.test
```diff
@@ -301,6 +308,27 @@ select row_number() over (), first_value(i) over (), first_value(j) over () from
 3	2	2
 4	2	2
 
+# FIRST_VALUE can only stream when the first row of the frame does not move
```
hawkfish, line 311: Can we add some coverage tests for `RANGE`? And include the plans to be sure?

hawkfish (changes requested):

src/function/window/window_value_function.cpp
u70b3: Done - `CURRENT_ROW_GROUPS` and `UNBOUNDED_FOLLOWING` end frames are now allowed, with result and plan tests added for both (verified against the blocking operator).

test/sql/window/test_streaming_window.test
u70b3: Done - added RANGE coverage: plan checks for streaming (`UNBOUNDED PRECEDING` start) and non-streaming (moving start / `ORDER BY`) RANGE frames, plus result tests.

hawkfish (approved):
Thanks - LGTM!

PR #24498, opened 2026-08-04
Return Identifier from GetDefaultSchema/Table/TableSchema

JelteF:
I guess these were missed by accident before.


Mytherin:
Thanks!

PR #24499, opened 2026-08-04
Spill aggregate states when the aggregation goes external

mustafahasankhan:
### Problem

Aggregate states keep their variable size payloads (strings, buffered inputs) on arena allocators. The buffer manager tracks that memory but cannot evict it, so wide payloads retained per group fail under a memory limit even though everything else about the aggregation can go out-of-core:

```sql
SET memory_limit = '256MB';
SELECT k, first(wide) FROM ... GROUP BY k;  -- 1M groups x 800B of state
-- Out of Memory Error: failed to allocate data of size 16.0 MiB (230.5 MiB/244.1 MiB used)
```

### Approach

Follows the direction from #24033's review: native states are used during normal aggregation, and export/serialization is invoked only at spill boundaries, through the existing aggregate state export layouts, with no new callbacks and no per-function changes.

- When a thread-local hash table is abandoned while the sink is external, its abandoned data is rewritten into a fully spillable collection: the group columns and hash as-is, plus one typed column per aggregate state (`AggregateStateSerialization` over the state export layouts). The arena is then freed, so sink memory is bounded by one flush cycle.
- The exported data flows through the same radix partitioning as the rest, and each partition imports its states again right before the existing finalize combine runs.
- Each finalized partition owns its arenas and releases them once it has been scanned, so finalize memory is bounded by the partitions in flight rather than the whole aggregation.
- In-memory queries never reach any of this: the export only happens when the sink has gone external.

One addition to the import path: the state deserialization stores primitive string values that point into the input vector, which is fine for the state export functions (they combine immediately) but not for states that outlive the input, so importing materializes those strings onto the partition arena, walking the same field descriptors.

Aggregates that cannot describe their state, or whose states own resources beyond the arena (`histogram`, `mode`), keep the current behavior, detected per layout.

### Results

The added test retains roughly 800MB of state under a 256MB limit and verifies the payloads byte for byte: `first`, `last`, `min` (compared per row), `string_agg` (grows per row), parallel combines under `verify_parallelism` at 64MB, and `EXPORT_STATE` alongside the spilling.

The added benchmark (8M groups x 800B, `memory_limit='4GB'`): out of memory on current main, completes in 4.8-9.7s with this PR.

Full fast suite passes (5,256 of 5,257; the one failure is a live-network Parquet test that fails identically on current main). The aggregate and export suites also pass under AddressSanitizer with assertions forced on.

### Follow-ups

- Sorted (ORDER BY) aggregates flow through the same mechanism and complete when the aggregation is deeply external, but their finalize working memory near the memory limit boundary needs its own sizing work, so they are not part of the test yet.
- `histogram`/`mode` states own std containers (currently untracked entirely); an export-based path for them needs destructor handling on export.

Supersedes #24033.


lnkuiper:
Thanks! I've resolved the last few issues, looks good now

PR #24503, opened 2026-08-04
Iterative benchmark runs and make report easier to understand

smvv:
<details>
<summary>Original approach</summary>

### Adaptive sampling

Initially, collect 10 samples (2 times a batch of 5 samples) for every query per binary.

For regression candidates, choose an independent confirmation sample size from the observed runtime:

```python
confirmation_runs =
    clamp(5 × ceil((3 seconds / faster_side_median) / 5), 30, 100)
```

Approximate confirmation sizes:

| Query runtime       | Confirmation runs |
|---------------------|------------------:|
| 20 ms               | 100 (capped) |
| 50 ms               | 60 |
| 100 ms              | 30 |
| 300 ms              | 30 |
| 500 ms – a few sec  | 30 (minimum) |

This combines [Hyperfine](https://github.com/sharkdp/hyperfine#basic-benchmarks)’s 10-run/3-second acquisition model with [Criterion](https://bheisler.github.io/criterion.rs/book/user_guide/command_line_output.html)’s 95% confidence and ±2% noise principles.

### Simplify bench report

No regressions detected:

Screenshot 2026-08-06 at 08 55 03

A few slower and faster (but not exceeding regressions):

Screenshot 2026-08-06 at 08 55 20
</details>

### Easier to understand benchmark report
- Median/min/max measurements and compact gray ranges (like [hyperfine](https://github.com/sharkdp/hyperfine)).
- Five-column report with median delta and run counts.

No regressions found:

Screenshot 2026-08-06 at 13 31 01

Some faster/slower, but no regressions:

Screenshot 2026-08-06 at 13 31 45

Since benchmarks use fewer samples (CI time for TPC-DS went from 8min to 4 min), we can increase TPC-DS sf3 to sf10, like TPC-H. With lower scale factor, some query are very fast (`q32` takes `3ms`?!) and those make the bench report too noisy (see previous screenshot).

### Samples
- Alternating `base → PR`, then `PR → base` batches.
- Added `--samples` for local usage (rounded up to five samples).
- Added median-based `--early-stop`: reduce CI runtime if median is in noise range (±2%). 
  - Enabled `--early-stop` in Regression workflow jobs.

### What is a failing regression?
- Individual query regressions at `≥10%` emit CI warnings but do not fail CI.
- CI fails when geomean increases by `≥10%` or `≥50ms`. Same as what CI currently checks.
- `--nofail` suppresses only the geomean gate.

### DIFF
--- a/.github/workflows/Regression.yml
+++ b/.github/workflows/Regression.yml
@@ -141,10 +141,8 @@ jobs:
               python3 scripts/ci/retry.py -- wget -q https://blobs.duckdb.org/data/realnest/realnest.duckdb --output-document=duckdb_benchmark_data/real_nest.duckdb
           - name: Recursive CTE
             benchmark: benchmark/recursive_cte/recursive_cte.csv
-            regression_threshold_seconds: 0.01
           - name: AoC24
             benchmark: benchmark/aoc24/aoc24.csv
-            regression_threshold_seconds: 0.01
 
     steps:
       - uses: actions/checkout@v6
@@ -183,13 +181,9 @@ jobs:
 
       - name: Run regression benchmark
         run: |
-          extra_args=(--clear-benchmark-cache)
+          extra_args=(--benchmark-cache=clear)
           if [[ "${{ matrix.benchmark }}" == ".github/regression/realnest.csv" ]]; then
             extra_args=()
-          fi
-
-          if [[ -n "${{ matrix.regression_threshold_seconds }}" ]]; then
-            extra_args+=(--regression-threshold-seconds "${{ matrix.regression_threshold_seconds }}")
           fi
 
           python scripts/regression/test_runner.py \
@@ -260,9 +254,8 @@ jobs:
             --new build/current/release/benchmark/benchmark_runner \
             --benchmarks ${{ matrix.benchmarks }} \
             --verbose \
-            --clear-benchmark-cache \
+            --benchmark-cache=clear \
             --disable-timeout \
-            --max-timeout 3600 \
             --threads ${{ matrix.threads }}
 
   regression-bench-fivetran:
@@ -314,7 +307,7 @@ jobs:
 
       - name: Run regression benchmark
         run: |
-          python scripts/regression/test_runner.py --old build/base/release/benchmark/benchmark_runner --new build/current/release/benchmark/benchmark_runner --benchmarks benchmark/fivetran/benchmark_list.csv --verbose --threads 2 --clear-benchmark-cache
+          python scripts/regression/test_runner.py --old build/base/release/benchmark/benchmark_runner --new build/current/release/benchmark/benchmark_runner --benchmarks benchmark/fivetran/benchmark_list.csv --verbose --threads 2 --benchmark-cache=clear
 
   regression-test-storage:
     name: Storage Size Regression Test
--- a/.github/workflows/Regression.yml
+++ b/.github/workflows/Regression.yml
@@ -141,11 +141,9 @@ jobs:
               python3 scripts/ci/retry.py -- wget -q https://blobs.duckdb.org/data/realnest/realnest.duckdb --output-document=duckdb_benchmark_data/real_nest.duckdb
           - name: Recursive CTE
             benchmark: benchmark/recursive_cte/recursive_cte.csv
-            timed_runs: 7
             regression_threshold_seconds: 0.01
           - name: AoC24
             benchmark: benchmark/aoc24/aoc24.csv
-            timed_runs: 7
             regression_threshold_seconds: 0.01
 
     steps:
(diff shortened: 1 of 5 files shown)


kryonix (changes requested):
I like the direction here: an inexpensive initial screen followed by independent confirmation is a substantial improvement over repeatedly carrying forward candidates. I also agree that regression testing on GitHub runners needs a pragmatic time budget.

I do not think the current implementation can yet describe its result as a 95% confidence decision, though. The interval itself is correct for independent paired observations, but the sampling procedure does not produce those observations.

### Sampling unit and execution order

Each subprocess invocation produces five timings, and the runner executes five base timings followed by five PR timings. The implementation then treats the individual ratios as independent pairs.

Those timings share initialization, warmup, and the same short period of machine conditions, so the independent unit is closer to the process invocation than the individual timing. Base also always runs first, meaning thermal drift, cache state, frequency changes, or background load can become indistinguishable from a version effect.

This is visible in the Micro job: `choose_correct_filter_function` reported 100 pairs, but those came from 20 base/PR invocation pairs. For that example, the estimated intrabatch correlation of the log-ratios was approximately 0.37, giving a design effect of about 2.48. The run-level CI was `-4.2%…+1.4%`; computing over the invocation-level summaries produced approximately `-6.0%…+2.0%`.

I suggest keeping the five timed runs per process, because amortizing startup is reasonable, but using their median as one observation. Execution order should be counterbalanced:

```text
base → PR
PR → base
base → PR
PR → base
```

The reported measurement would then be, for example, `10 blocks × 5 timed runs`, with the CI calculated over the ten block ratios. This does not require additional executions; it produces a wider but more honest interval.

### Initial screening

The ten-run initial screen is useful and should remain. The PR's own results demonstrate this: initial changes of `-8.6%` and `+5.8%` disappeared during confirmation.

However, a benchmark whose initial median lies within ±2% is currently classified as unchanged without considering the width of its interval. Ten ratios can have a median of `1.00` and a valid median interval of `[0.90, 1.30]`. That is enough to skip expensive confirmation, but it is not evidence that performance is within ±2%.

For GitHub CI, I suggest reporting these as:

```text
not selected for confirmation: initial median within ±2%
```

or simply:

```text
no regression detected in initial screen
```

Only independently confirmed regressions should fail the job. A local `--thorough` mode could confirm every benchmark when someone wants stronger evidence.

### Early stopping

The confirmation loop repeatedly examines an ordinary 95% interval and stops when it enters the ±2% band. A fixed-sample confidence interval does not retain 95% coverage when inspected repeatedly.

The simplest fix is to consume the predetermined confirmation budget. If early stopping saves enough runner time to justify keeping it, the interim checks should use an adjusted confidence level or another sequentially valid rule. Otherwise the result should say only that confirmation stopped early without detecting a regression, rather than claiming confident equivalence within ±2%.

### Runtime-based sampling

Choosing 30–100 timings from the observed runtime is a reasonable way to bound GitHub runner cost. It is not a statistical sample-size calculation: precision depends on variability and distance from the 10% boundary, not runtime.

I would keep the current runtime heuristic but describe it as a confirmation budget. If that budget is exhausted while the interval still overlaps the regression threshold, the result should remain inconclusive.

For local use, I would preserve controls for the initial sample count, confirmation budget, thresholds, and timeouts. Presets such as `--quick`, `--ci`, and `--thorough` could make the intended trade-offs explicit.

### Regression policy

This also changes the regression policy, not only its measurement. The old condition was approximately:

```text
PR > 1.10 × (base + 50 ms)
```

The new condition is:

```text
lower confidence bound of median ratio > 1.10
```

For a 20 ms benchmark, the old gate required more than 77 ms, while the new gate can fail slightly above 22 ms. Removing that absolute floor may be desirable, especially because the old rule was nearly blind to short-query regressions, but it should be documented as an intentional policy change and calibrated rather than presented as a consequence of the statistical method.

I would validate the defaults using A/A comparisons of the same binary plus synthetic slowdowns around 5%, 10%, 12%, and 20%, covering sub-millisecond, medium, long, and noisy benchmarks. The useful outputs are the total CI cost, false-alarm rate, missed 12–20% regressions, and inconclusive rate.

Before merging, my minimum set of changes would be:

1. Counterbalance base/PR execution order.
2. Analyze invocation-level blocks instead of treating all five timings as independent.
3. Label screened-out benchmarks as unconfirmed rather than confidently unchanged.
4. Make early stopping compatible with the stated confidence level.
5. Describe runtime as selecting a cost budget.
6. Document the removal of the absolute regression floor and retain local sampling controls.

This would keep the adaptive approach practical for GitHub runners while making the statistical claims match what was actually measured.

lnkuiper (approved):

Tishj:
What is the replacement for `--root-dir` ?

smvv:
we can omit `--root-dir` because the current work dir should work well.
if not, let me know and I'll add `--root-dir` back!

PR #10053, opened 2023-12-21
Allow extension optimizers to trigger a rerun of the optimizer + bump spatial

Maxxen:
This PR makes the `OptimizerExtensions` `optimize_function` handler return a boolean signaling whether or not to rerun all the optimizer steps after the extension optimizer has been applied. This is useful in scenarios where the optimizer extension significantly rewrites the logical plan so that new opportunities for e.g. column removal or predicate pushdown arise.

Also patches and bumps the spatial extension to the newest version.

PR #24504, opened 2026-08-04
Internal #10053: Postgres Nested Semantics

hawkfish:
* Split UNION comparisons from STRUCT
* Convert nested comparisons to PG semantics
* Convert nested matching to PG semantics
* Update some more tests for nested equality NULL semantics
* Update another test for nested equality NULL semantics


lnkuiper:

One small `UNION` issue again:

`UNION` equality now considers a `NULL` payload equal to every payload with the same tag. The equality path retains an unknown child comparison as “still equal”, without restricting or cleaning up the active `UNION` member. This makes equality non-transitive and inconsistent with both `IS NOT DISTINCT FROM` and the
same value wrapped in a `STRUCT`.
```sql
WITH vals AS (
    SELECT
        union_value(str := 'a')::UNION(num INT, str VARCHAR) AS a,
        union_value(str := NULL)::UNION(num INT, str VARCHAR) AS n,
        union_value(str := 'b')::UNION(num INT, str VARCHAR) AS b
)
SELECT
    a = n,
    n = b,
    a = b,
    a IS NOT DISTINCT FROM n,
    {'u': a} = {'u': n}
FROM vals;
```
This returns `true, true, false, false, false`. The first two comparisons cannot both be `true` while the third is `false`.

hawkfish:
@lnkuiper The CI failure appears unrelated (just strings).

hawkfish:
Hey @kryonix - Can you have a look at the test failure? It's something you added last week. Not sure why changing the nested comparison semantics should change the plan there.

hawkfish:
> Hey @kryonix - Can you have a look at the test failure? It's something you added last week. Not sure why changing the nested comparison semantics should change the plan there.

Actually the failure appears unrelated - I just reproduced it on `main`.

kryonix:
Thanks for pinging me, I will look into it. And yes, I agree, this is unrelated to your changes.

Mytherin:
I wonder if it makes sense to pull unions into a collation similar to how we handle variant comparisons now? It seems relatively straightforward to generate a comparator for unions (`<tag_id><value>` as a sort key, with `<tag_id> = 0` for `NULL`).

hawkfish:
> I wonder if it makes sense to pull unions into a collation similar to how we handle variant comparisons now? It seems relatively straightforward to generate a comparator for unions (`<tag_id><value>` as a sort key, with `<tag_id> = 0` for `NULL`).

Could we file a follow on issue for this?

PR #24508, opened 2026-08-05
Fix projected types for missing columns

### LINKED ISSUE #24253: `read_parquet` throws an INTERNAL exception when projecting
### What happens?

See repro for details.

### To Reproduce

```sql
memory D COPY (
           SELECT {'a': 1, 'b': 2}::STRUCT(a INT, b INT) AS s, 1 AS id
         ) TO '/tmp/f1.parquet' (FORMAT PARQUET);
memory D COPY (
           SELECT 2 AS id
         ) TO '/tmp/f2.parquet' (FORMAT PARQUET);
memory D SELECT id, s.a AS a_only
         FROM read_parquet(['/tmp/f1.parquet', '/tmp/f2.parquet'], union_by_name=true)
         ORDER BY id;
INTERNAL Error:
ExpressionExecutor::Execute called with a result vector of type INTEGER that does not match expression type STRUCT(a INTEGER, b INTEGER)
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Andrewtangtang:
## Summary

Resolves #24253

When `union_by_name` fills a column missing from an input file, it creates a NULL value for that column. Nested projection pushdown can make the scan output a child type instead of the full global column type, which previously caused a type mismatch and an INTERNAL error. This change uses the projected scan type for the generated NULL when available, ensuring it matches the scan output.

### DIFF
--- a/src/common/multi_file/multi_file_reader.cpp
+++ b/src/common/multi_file/multi_file_reader.cpp
@@ -382,7 +382,8 @@ void MultiFileReader::FinalizeBind(MultiFileReaderData &reader_data, const Multi
 			if (not_present_in_file) {
 				// we need to project a column with name \"global_name\" - but it does not exist in the current file
 				// push a NULL value of the specified type
-				reader_data.constant_map.Add(global_idx, Value(type));
+				auto &constant_type = col_id.HasType() ? col_id.GetScanType() : type;
+				reader_data.constant_map.Add(global_idx, Value(constant_type));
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks!

PR #24511, opened 2026-08-05
Fix quantile hugeint precision

### LINKED ISSUE #24314: quantile_cont/median miscomputation
### What happens?

See repro for details

### To Reproduce

```sql
WITH t(v) AS (
           VALUES (9999999999999999999999999999::DECIMAL(38,0)),
                  (9999999999999999999999999997::DECIMAL(38,0))
         )
         SELECT quantile_cont(v, 0.5), median(v) FROM t;
┌──────────────────────────────┬──────────────────────────────┐
│    quantile_cont(v, 0.5)     │          median(v)           │
│        decimal(38,0)         │        decimal(38,0)         │
├──────────────────────────────┼──────────────────────────────┤
│ 9999999999999999583119736832 │ 9999999999999999583119736832 │
└──────────────────────────────┴──────────────────────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Vihaan275:
Fixes #24314

quantile_cont and median return incorrect results when interpolating between two large DECIMAL(38,0) values (stored internally as hugeint_t). For example:

SQL:
```sql
WITH t(v) AS (
    VALUES (9999999999999999999999999999::DECIMAL(38,0)),
           (9999999999999999999999999997::DECIMAL(38,0))
)
SELECT quantile_cont(v, 0.5), median(v) FROM t;
```

returns 9999999999999999583119736832 instead of 9999999999999999999999999998. 

Root cause: the hugeint_t specialization of InterpolateOperator::Operation casts lo and hi directly to double before interpolating:

"return Hugeint::Convert(Operation(Hugeint::Cast<double>(lo), d, Hugeint::Cast<double>(hi)));"

For values this large, double's ~15-17 significant digits of precision aren't enough as both lo and hi round to the same double, so the interpolation runs on an already-corrupted value. This differs from the generic (non-hugeint) InterpolateOperator template, which computes delta = hi - lo first and only casts the delta to double.

Fix: mirror the generic template's approach for the hugeint_t specialization by computing the delta first using Hugeint::TrySubtractInPlance, and then cast only the delta (typically a smaller value) as a double. Because lo and hi can theoretically be far enough apart that even their difference overflows hugeint_t's range, the delta subtraction is checked and if it overflows, we go back to the original implementation which is lossy for the extreme cases but avoid undefined behavior. 

Testing: added two regression tests to test_quantile_cont.test covering the original repro (d=0.5, exact interpolation) and a non-0.5 fraction (d=0.3) to exercise the truncation path in the delta calculation. All existing tests in the file continue to pass, and the previously-failing mad_overflow.test fuzzer test now passes as well.


Mytherin:
Thanks!

PR #24514, opened 2026-08-05
Preserve empty-set results of zero-returning aggregates in correlated subqueries

### LINKED ISSUE #24049: Correlated aggregates return the wrong empty-set value
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D WITH
         t1(id, k) AS (VALUES (11,1), (22,2), (33,3), (44,4)),
         t2(id, k) AS (VALUES (101,1), (301,3), (302,3), (303,3))
         SELECT id,
                (SELECT regr_count(1.0, 1.0) FROM t2 WHERE t2.k=t1.k),
                (SELECT approx_count_distinct(t2.id) FROM t2 WHERE t2.k=t1.k)
         FROM t1 ORDER BY id;
┌───────┬─────────────────────────────────────────────────┬─────────────────────────────────────────────────┐
│  id   │ (SELECT regr_count(1.0, 1.0) FROM t2 WHERE (t2. │ (SELECT approx_count_distinct(t2.id) FROM t2 WH │
│       │                   k = t1.k))                    │               ERE (t2.k = t1.k))                │
│ int32 │                     uint32                      │                      int64                      │
├───────┼─────────────────────────────────────────────────┼─────────────────────────────────────────────────┤
│    11 │                                               1 │                                               1 │
├───────┼─────────────────────────────────────────────────┼─────────────────────────────────────────────────┤
│    22 │                                            NULL │                                            NULL │
├───────┼─────────────────────────────────────────────────┼─────────────────────────────────────────────────┤
│    33 │                                               3 │                                               3 │
├───────┼─────────────────────────────────────────────────┼─────────────────────────────────────────────────┤
│    44 │                                            NULL │                                            NULL │
└───────┴─────────────────────────────────────────────────┴─────────────────────────────────────────────────┘
```
The expected result is
```sql
┌───────┬──────────────────────────────┬──────────────────────────────────┐
│  id   │ regr_count(1.0, 1.0)        │ approx_count_distinct(t2.id)     │
├───────┼──────────────────────────────┼──────────────────────────────────┤
│    11 │                            1 │                                1 │
│    22 │                            0 │                                0 │
│    33 │                            3 │                                3 │
│    44 │                            0 │                                0 │
└───────┴──────────────────────────────┴──────────────────────────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

matthiasgoergens:
An uncorrelated `regr_count()` or `approx_count_distinct()` over no rows returns zero, but the equivalent correlated scalar subquery returns `NULL` for an outer key with no matching inner rows; `entropy()` shares the latent defect. Decorrelating an ungrouped aggregate adds the correlated columns as grouping keys and joins those groups back to the outer domain, so a missing group becomes a row of `NULL` values, and existing code special-cased only `count()` and `count_star()` when replacing that missing result with zero.

The fix adds explicit `returns_zero_on_empty` aggregate metadata, set on `count`, `count_star`, `regr_count`, `approx_count_distinct` and every bound `entropy` implementation. The decorrelator uses the metadata both to choose a LEFT join and to identify results requiring the empty-set rewrite, which appends an internal `count_star()` as a group-presence marker and tests that marker rather than the aggregate result or correlated key.

A new regression in `test/issues/general/test_24049.test` covers the reported aggregates, `entropy`, NULL-on-empty behaviour, DISTINCT and FILTER, and present and missing correlation keys. Fixes #24049.


test/issues/general/test_24049.test
```diff
@@ -0,0 +1,100 @@
...
+        FROM inner_t WHERE inner_t.k=outer_t.k),
+       (SELECT approx_count_distinct(v) FILTER (WHERE false)
+        FROM inner_t WHERE inner_t.k=outer_t.k)
+FROM outer_t
+ORDER BY k;
+----
+1	0	0
+2	0	0
+
+# GROUP BY and HAVING can suppress the aggregate row; that remains NULL rather than an empty-input result.
```
kryonix, line 113: The implementation looks correct, but the tests do not cover `perform_delim=false`. A nested correlation key such as LIST forces that path, where correlated values are carried through internal first aggregates before the new presence marker. Could you add the following case? It also verifies that the replacement is applied in `HAVING` and that explicit `GROUP BY ()` retains ungrouped-aggregate semantics.

```sql
query III
WITH outer_t(k) AS (VALUES ([1]), ([2])),
     inner_t(k, v) AS (VALUES ([1], 10))
SELECT k,
       (SELECT regr_count(v, v)
        FROM inner_t
        WHERE inner_t.k = outer_t.k
        HAVING regr_count(v, v) = 0),
       (SELECT regr_count(v, v)
        FROM inner_t
        WHERE inner_t.k = outer_t.k
        GROUP BY ())
FROM outer_t
ORDER BY k;
----
[1]	NULL	1
[2]	0	0
```

kryonix (commented):
Thanks for the PR! The explicit `returns_zero_on_empty` property seems to be a well scoped contract, and using an unfiltered count_star() as a presence marker correctly distinguishes a missing decorrelated group from a present group whose aggregate result happens to be `NULL` or zero. I tested the original reproducer plus `NULL` correlation keys, duplicate outer rows, `FILTER` and `DISTINCT`, `HAVING`, grouping sets, nested and lateral subqueries, volatile subqueries, optimizer-disabled execution, multiple correlation keys, and the non-delim LIST path. They produced the expected results.

matthiasgoergens:
Added a LIST correlation key test that exercises the `perform_delim=false` path — the nested key type forces correlated values through the first aggregation rather than the deliminator. The empty-set result matches the existing cases.

test/issues/general/test_24049.test
matthiasgoergens: Added the case as-is — it passes on this branch.

lnkuiper:
Thanks for the PR! I think we should avoid adding `returns_zero_on_empty` here.

This currently makes a correlated `count(*) EXPORT_STATE` abort with `AGGREGATE_STATE != BIGINT`. `EXPORT_STATE` changes the bound aggregate’s return type and empty result, but the new flag remains set, so the missing-group rewrite still injects a numeric zero.

Could we instead obtain the actual empty result at execution time? `DEFAULT_NULL_HANDLING` already identifies aggregates whose empty result is guaranteed to be `NULL`. For special-handling aggregates, we could evaluate a copy of the bound aggregate over a `LogicalEmptyResult` and use that result when the presence marker is missing. This would use the normal aggregate lifecycle and also cover exported states, arbitrary empty results, and extension aggregates without adding per-function metadata.

PR #24515, opened 2026-08-05
Avoid array with brackets creating an extra demension

Dtenwolde:

Array with `[]` accidentally created an extra dimension with the new parser: 
```sql
SELECT typeof(NULL::INTEGER ARRAY[3]);
-- Old parser: INTEGER[3]
-- New parser: INTEGER[][3]
```

This should now be fixed


Mytherin:
Thanks!

PR #24519, opened 2026-08-05
Fix escape strings being incorrectly tokenized

Dtenwolde:

The query: `SELECT E'it\'s a test';` was improperly tokenized and would result in untermined string literal. This was a regression compared to `v1.5.5.`. 

With E'...', backslash escapes should be respected while finding the end of the string token. In particular, \' should be treated as part of the literal, not as the closing quote. This is now something we track in the tokenizer. Regular string literals are not changed. 

```sql
./build/release/duckdb -c "SELECT E'it\'s a test';"
Parser Error:
unterminated string literal

LINE 1: SELECT E'it\'s a test';
                             ^
```

This is now corrected: 
```sql
./build/release/duckdb -c "SELECT E'it\'s a test';"
┌────────────────┐
│ 'it''s a test' │
│    varchar     │
├────────────────┤
│ it's a test    │
└────────────────┘
```


Mytherin:
Thanks!

PR #24521, opened 2026-08-05
Prune row groups when an IN-list covers the whole zonemap rangend add

### LINKED ISSUE #24414: `IN` constraint not leveraged to prune row group
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE t(i INTEGER CHECK (i IN (1, 2, 3)));
memory D INSERT INTO t
         SELECT (range % 3 + 1)::INTEGER
         FROM range(6144);

-- predicate not possible to achieve based on the constraint, should prune all row groups
memory D EXPLAIN ANALYZE
         SELECT count(*)
         FROM t
         WHERE i NOT IN (1, 2, 3);
╭─ Summary ───────────╮
│ Total Time: 0.0004s │
╰─────────────────────╯
╰───────────────┬───────────────╯
╭─ Table Scan ──┴───────────────╮
│ Table: memory.main.t          │
│ Filters: NOT (i IN (1, 2, 3)) │
│ Row Groups Scanned: 1 / 1     │
│ 0 rows                    0µs │
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

vvvdwbvvv:
Resolves #24414

### Problem

A NOT IN predicate whose list covers a row group's entire zonemap range still scans that row group:
```CREATE TABLE t(i INTEGER CHECK (i IN (1, 2, 3)));
INSERT INTO t SELECT (range % 3 + 1)::INTEGER FROM range(6144);

EXPLAIN ANALYZE SELECT count(*) FROM t WHERE i NOT IN (1, 2, 3);
```

The zonemap is [1, 3] and i is an integer, so every value the row group can hold is listed. i IN (1, 2, 3) is therefore always true and its negation always false — no row can match, and nothing needs to be read.

### Before

```text
╰───────────────┬───────────────╯
╭─ Table Scan ──┴───────────────╮
│ Table: memory.main.t          │
│ Filters: NOT (i IN (1, 2, 3)) │
│ Row Groups Scanned: 1 / 1     │
│ 0 rows                    0µs │
```
### After
```text
╭─ Ungrouped Aggregate ────╮
│ Aggregates: count_star() │
│ 1 row                0µs │
╰─────────────┬────────────╯
╭─ Empty Result ───────────╮
│ 0 rows               0µs │
╰──────────────────────────╯
```

### Tests covered

New file `test/sql/optimizer/in_list_row_group_pruning.test`. It fails on main at the first assertion. Three cases:

- The fix. Three row groups, each with `zonemap [1, 3]`. `WHERE i NOT IN (1, 2, 3)` becomes Empty Result.
- Don't prune too much. Only the first row group is `[1, 3]`; the other two are `[4, 5]` and must be read. Row Groups Scanned: 2 / 3, answer 4096.
- Don't prune the wrong type. `A DOUBLE `column with` zonemap [1, 3]` holding 1.5. That row is not in the list, so `count(*)` is 1.

### DIFF
--- a/src/storage/statistics/numeric_stats.cpp
+++ b/src/storage/statistics/numeric_stats.cpp
@@ -1,6 +1,7 @@
 #include "duckdb/storage/statistics/numeric_stats.hpp"
 
 #include "duckdb/common/algorithm.hpp"
+#include "duckdb/common/numeric_utils.hpp"
 #include "duckdb/common/operator/comparison_operators.hpp"
 #include "duckdb/common/serializer/deserializer.hpp"
 #include "duckdb/common/serializer/serializer.hpp"
@@ -285,14 +286,16 @@ FilterPropagateResult NumericStats::CheckZonemap(const BaseStatistics &stats, Ex
 
 bool NumericStats::ConstantsCoverRange(const BaseStatistics &stats, array_ptr<const Value> constants) {
 	auto &type = stats.GetType();
-	auto physical_type = type.InternalType();
-	// hugeint_t holds every integral type exactly, except the 128-bit ones
-	if (!type.IsIntegral() || physical_type == PhysicalType::INT128 || physical_type == PhysicalType::UINT128 ||
-	    !NumericStats::HasMinMax(stats)) {
+	// values are compared as hugeint_t, which cannot hold large UINT128 values
+	if (!type.IsIntegral() || type.InternalType() == PhysicalType::UINT128 || !NumericStats::HasMinMax(stats)) {
 		return false;
 	}
 	auto min_value = NumericStats::Min(stats).GetValue<hugeint_t>();
 	auto max_value = NumericStats::Max(stats).GetValue<hugeint_t>();
+	// covering [min, max] requires at least max - min + 1 constants
+	if (max_value - min_value >= NumericCast<int64_t>(constants.size())) {
+		return false;
+	}
 	vector<hugeint_t> values;
 	for (auto &constant_value : constants) {
 		if (constant_value.type() != type) {
--- a/src/storage/statistics/numeric_stats.cpp
+++ b/src/storage/statistics/numeric_stats.cpp
@@ -3,6 +3,7 @@
 #include "duckdb/common/algorithm.hpp"
 #include "duckdb/common/numeric_utils.hpp"
 #include "duckdb/common/operator/comparison_operators.hpp"
+#include "duckdb/common/operator/subtract.hpp"
 #include "duckdb/common/serializer/deserializer.hpp"
 #include "duckdb/common/serializer/serializer.hpp"
 #include "duckdb/common/types/vector.hpp"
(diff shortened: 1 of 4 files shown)


src/storage/statistics/numeric_stats.cpp
```diff
@@ -282,6 +283,38 @@ FilterPropagateResult NumericStats::CheckZonemap(const BaseStatistics &stats, Ex
 	}
 }
 
+bool NumericStats::ConstantsCoverRange(const BaseStatistics &stats, array_ptr<const Value> constants) {
+	auto &type = stats.GetType();
+	auto physical_type = type.InternalType();
+	// hugeint_t holds every integral type exactly, except the 128-bit ones
+	if (!type.IsIntegral() || physical_type == PhysicalType::INT128 || physical_type == PhysicalType::UINT128 ||
```
Mytherin: We only need to exclude `PhysicalType::UINT128` here.

src/storage/statistics/numeric_stats.cpp
```diff
@@ -282,6 +283,38 @@ FilterPropagateResult NumericStats::CheckZonemap(const BaseStatistics &stats, Ex
...
+	    !NumericStats::HasMinMax(stats)) {
+		return false;
+	}
+	auto min_value = NumericStats::Min(stats).GetValue<hugeint_t>();
+	auto max_value = NumericStats::Max(stats).GetValue<hugeint_t>();
+	vector<hugeint_t> values;
```
Mytherin, line 302: Maybe add an early-out if we know the amount of values in the `IN` list cannot possibly cover the entire range, e.g.:

```cpp
if (constants.size() < max_value - min_value + 1) {
      return false;  // not enough constants to cover the range
  }
```

src/storage/statistics/numeric_stats.cpp
```diff
@@ -282,6 +283,38 @@ FilterPropagateResult NumericStats::CheckZonemap(const BaseStatistics &stats, Ex
...
+	for (auto &constant_value : constants) {
+		if (constant_value.type() != type) {
+			return false;
+		}
+		auto constant = constant_value.GetValue<hugeint_t>();
```
Mytherin, line 307: Should we check `NULL`-ness here?

src/storage/statistics/numeric_stats.cpp
```diff
@@ -282,6 +283,38 @@ FilterPropagateResult NumericStats::CheckZonemap(const BaseStatistics &stats, Ex
...
+	// hugeint_t holds every integral type exactly, except the 128-bit ones
+	if (!type.IsIntegral() || physical_type == PhysicalType::INT128 || physical_type == PhysicalType::UINT128 ||
+	    !NumericStats::HasMinMax(stats)) {
+		return false;
+	}
+	auto min_value = NumericStats::Min(stats).GetValue<hugeint_t>();
+	auto max_value = NumericStats::Max(stats).GetValue<hugeint_t>();
+	vector<hugeint_t> values;
+	for (auto &constant_value : constants) {
+		if (constant_value.type() != type) {
```
Mytherin, line 304: Why do we need this type check? Doesn't the binding already take care of ensuring types are consistent? This should maybe be turned into an assertion.

Mytherin (commented):

vvvdwbvvv:
Thanks for suggestion!  I'll thinking for a while

src/storage/statistics/numeric_stats.cpp
vvvdwbvvv: Binding guarantees the constants share a type, but stats here is derived by `TryGetExpressionStats`, not the column's own statistics — its type isn't binder-controlled. On a mismatch `GetValue<hugeint_t>()` can throw during planning, so I'd rather bail than assert. Can hoist it out of the loop and check `constants[0].type()` once if you prefer.

vvvdwbvvv:
@Mytherin Except the type check should be confirmed, all fix in e7cf055

vvvdwbvvv:
Fixed the CI failed cause in 6da75d8

vvvdwbvvv:
why the draft bot kept sending it back🤣

PR #24522, opened 2026-08-05
Improved tokenization for nested comments

Dtenwolde:

The old lexer handled multiple nested comments, but the new tokenizer only kept track of a single multi-line comment. So if you have multiple nested comments this would go wrong. We now keep track of the number of nested comments that we encounter during tokenization. 

Example query: 
```sql
SELECT 1 /* I /* love */ comments */ + 2;
```
(Side note, Github's markdown highlighting doesn't seem to handle nested comments as well as DuckDB now)

The highlighting in the CLI works as well: 
afbeelding


Mytherin:
Thanks!

PR #24525, opened 2026-08-05
`COPY` with `ORDER_BY` without `PARTITION_BY`

lnkuiper:
I didn't get around to doing this, but it was surprisingly a very small change.

CC @Tmonster


Tishj:
Ah cool, guess I can remove https://github.com/duckdb/duckdb-iceberg/commit/617c8f638588d198cd3a67c6317167aa5939d165

lnkuiper:
Yes! Just to clarify, this will write multiple files to a directory, each of which is locally sorted.

Tishj:
Yea that's the intent behind the sort-order in Iceberg.
For lack of a file-local sort I settled for a global sort, that should just be a superset of the file-local sort, in terms of per-file sorted-ness, afaik.

But the file-local sort will surely be more performant!

Issue #24288, opened 2026-07-29
INTERNAL Error: Vector::Reference used on vector of different type (BIGINT/VARCHAR) -- COALESCE over a self-joined SEMI/ANTI JOIN relation (v2.0 regression, clean on 1.5.5))


wanteatfruit:
### What happens?

The following SQL raises an `INTERNAL ERROR`. It might be a regression, as on 1.5.5, the same SQL returns an empty table.

### To Reproduce

```SQL
CREATE TABLE b (id BIGINT, name VARCHAR, created_at VARCHAR);
INSERT INTO b VALUES (-3, 'a', 'a'), (NULL, 'b', 'b');
CREATE TABLE empt AS SELECT * FROM b WHERE 1 = 0;
CREATE VIEW t AS SELECT id, name, created_at FROM b ANTI JOIN empt ON TRUE;

SELECT t5.created_at
FROM (SELECT 'YEAR' AS c0
      FROM t AS t1 INNER JOIN t AS t2 ON t1.created_at <= t2.created_at
      WHERE COALESCE(t1.id, t2.id) IS NOT NULL) AS sq4
INNER JOIN t AS t5 ON sq4.c0 = t5.created_at;
```

Full stack trace:
```
INTERNAL Error:
Vector::Reference used on vector of different type (source BIGINT referenced VARCHAR)

Stack Trace:

0        duckdb::Exception::ToJSON(duckdb::ExceptionType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 48
1        duckdb::InternalException::InternalException(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 32
2        duckdb::InternalException::InternalException<duckdb::LogicalType const&, duckdb::LogicalType const&>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, duckdb::LogicalType const&, duckdb::LogicalType const&) + 136
3        duckdb::Vector::Reference(duckdb::Vector const&) + 228
4        duckdb::Vector::Slice(duckdb::Vector const&, duckdb::SelectionVector const&, unsigned long long) + 32
5        duckdb::ExpressionExecutor::Execute(duckdb::Expression const&, duckdb::ExpressionState*, duckdb::SelectionVector const*, unsigned long long, duckdb::Vector&) + 312
[...]

PR #24526, opened 2026-08-05
Issue #24288: Join Predicate Binding 

hawkfish:
* Restore the bindings member after binding child ANY ANTI JOINs.


lnkuiper:
Thanks!

PR #24533, opened 2026-08-05
Do not cache encrypted Parquet metadata in the plaintext cache

### LINKED ISSUE #24317: Read parquet without encryption key passes through
### What happens?

Hi team, I found when parquet metadata cache enabled, meanwhile with parquet file encrypted, querying on the parquet could pass through via cached metadata without validation. See repro for details.

### To Reproduce

```sql
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D COPY (SELECT i::INT AS i FROM range(10) t(i))
         TO '/private/tmp/enc_cache_no_config.parquet'
memory D 
memory D SELECT count(*), min(i), max(i)
         FROM read_parquet('/private/tmp/enc_cache_no_config.parquet',
                           encryption_config={footer_key:'key256'});
┌──────────────┬────────┬────────┐
│ count_star() │ min(i) │ max(i) │
│    int64     │ int32  │ int32  │
├──────────────┼────────┼────────┤
│           10 │      0 │      9 │
└──────────────┴────────┴────────┘

-- able to query parquet file with no key provided
memory D SELECT count(*), min(i), max(i)
         FROM read_parquet('/private/tmp/enc_cache_no_config.parquet');
┌──────────────┬────────┬────────┐
│ count_star() │ min(i) │ max(i) │
│    int64     │ int32  │ int32  │
├──────────────┼────────┼────────┤
│           10 │      0 │      9 │
└──────────────┴────────┴────────┘
```

### OS:

macos

### DuckDB Version:

fab741bf5cbf5c0bc073f68f2d66a304e7dab724

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

matthiasgoergens:
Closes #24317.

DuckDB's database-global Parquet metadata cache keyed entries only by file
path. After one read with the correct encryption configuration cached a
decrypted encrypted footer, later metadata-only queries on the same file —
without the encryption configuration, or with a wrong key — could reuse the
cached plaintext schema, cardinality, and statistics without the footer being
authenticated again.

Exclude encrypted footers from the plaintext metadata cache. Caching of
unencrypted files is unchanged.

The regression test fails against an unchanged build and passes 11 assertions
on the patched Debug ASan/UBSan build. Four ordinary metadata-cache suites
pass another 205 assertions. Rebased onto current `main` (`a82e24d90b`)
without conflicts.

Verified: `test/issues/general/test_24317.test` fails on main `aedb56e5c1`
(unauthenticated metadata read of the encrypted file returns `10 0 9` from
the plaintext cache instead of the expected encryption error), passes with
this branch (11 assertions).

### DIFF
--- a/extension/parquet/include/parquet_file_metadata_cache.hpp
+++ b/extension/parquet/include/parquet_file_metadata_cache.hpp
@@ -55,6 +55,7 @@ public:
 	bool IsValid(CachingFileHandle &new_handle) const;
 	//! Return if a cache entry is valid.
 	ParquetCacheValidity IsValid(const OpenFileInfo &info, ClientContext &context) const;
+	bool IsEncrypted() const;
 
 private:
 	timestamp_t last_modified;
--- a/extension/parquet/parquet_file_metadata_cache.cpp
+++ b/extension/parquet/parquet_file_metadata_cache.cpp
@@ -93,4 +93,9 @@ ParquetCacheValidity ParquetFileMetadataCache::IsValid(const OpenFileInfo &info,
 	return ParquetCacheValidity::INVALID;
 }
 
+bool ParquetFileMetadataCache::IsEncrypted() const {
+	return crypto_metadata && (crypto_metadata->encryption_algorithm.__isset.AES_GCM_V1 ||
+	                           crypto_metadata->encryption_algorithm.__isset.AES_GCM_CTR_V1);
+}
+
 } // namespace duckdb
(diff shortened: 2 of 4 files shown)


dentiny (commented):
I suspect encryption is needed for a wide range of enterprise usage (my previous company encryption all data stored on disk, whether its object storage or database), disabling caching completely would lead to perf hurt I'm afraid. I'm actually working on another approach -- validate against key before cache access. Wdyt?

matthiasgoergens:
Agreed — I've seen #24721, and validating the key before cache access is strictly better than dropping encrypted footers from the cache. Since #24721 covers #24317 and is green, I'll close this one in its favour. Happy to help review #24721 if useful.

matthiasgoergens:
Superseded by #24721 (key-validated caching, same fix with better performance).

PR #24535, opened 2026-08-05
Preserve GROUPING() bindings in the ordered-list aggregate rewrite

### LINKED ISSUE #23940: `GROUPING` combined with ordered `list` raises an INTERNAL error
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D SELECT grouping(g), list(v ORDER BY v)
         FROM (VALUES (1, 3)) t(g,v)
         GROUP BY g;
INTERNAL Error:
Failed to bind column reference ""GROUPING(t.g)"" [11.0] (bindings: {#[14.0]})
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

matthiasgoergens:
An ordered `list()` whose argument is also its ordering key makes the aggregate-function rewriter insert a projection that cannot resolve the query's `GROUPING()` column, failing with `INTERNAL Error: Failed to bind column reference "GROUPING(t.g)"`.

`AggregateFunctionRewriter` rewrites `list(x ORDER BY x)` into `list_sort(list(x), ...)`: its generic rewrite path adds private helper aggregates and inserts a projection exposing the original groups and aggregates, but `LogicalAggregate` exposes `GROUPING()` results through a third binding family, `groupings_index`, which that projection neither forwarded nor remapped — so the parent retained a binding its child no longer exposed. The fix appends each real grouping-function result to the projection after the original aggregates and maps its old `groupings_index` binding to the new position; the offset uses the aggregate count captured before rewriting, so helper aggregates remain private.

The new issue test covers ordinary grouping, `ROLLUP` with a non-zero grouping result, two simultaneous `GROUPING()` outputs, two rewritten ordered lists with separate private helpers, and a `GROUPING()` binding referenced only by `HAVING` or `ORDER BY` above the rewrite projection. Fixes #23940.

### DIFF
--- a/src/optimizer/aggregate_function_rewriter.cpp
+++ b/src/optimizer/aggregate_function_rewriter.cpp
@@ -392,6 +392,16 @@ private:
 			projection_expressions.push_back(std::move(final_result));
 		}
 
+		// Pass grouping-function bindings through after the original aggregates. Any helper aggregates added by the
+		// rewrite remain private to this projection.
+		for (idx_t grouping_idx = 0; grouping_idx < aggr.grouping_functions.size(); grouping_idx++) {
+			ColumnBinding grouping_binding(aggr.groupings_index, ProjectionIndex(grouping_idx));
+			aggregate_map[grouping_binding] =
+			    ColumnBinding(proj_index, ProjectionIndex(group_count + aggr_count + grouping_idx));
+			projection_expressions.push_back(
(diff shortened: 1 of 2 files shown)


DinosL:
Hi @matthiasgoergens, if this isn’t actually a draft, could you mark it as ready for review so that CI can run?

PR #24538, opened 2026-08-06
Fix infinite loop in equi_width_bins with more bins than boundaries

### LINKED ISSUE #24282: Bug8: Infinite Loop in equi_width_bins() via Zero Step Size
### What happens?

### To Reproduce

```sql
-- vuln_001.sql
```

Trigger output:

```
---- ATTACK RUN ----
Attack run exit code: 124

Control run exit code: 0
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

hanafy-dev:
When the requested bin count exceeds the number of boundaries in the range, the integer step size truncates to zero and the boundary loop in EquiWidthBinsInteger::Operation never terminates, pinning a CPU core. This affected both the nice_rounding and plain paths.

Clamp the integer step to the smallest possible step when it truncates to zero.
Make the hugeint RoundToNumber return its input when the rounding number is zero instead of relying on hugeint division by zero returning zero, mirroring the double implementation.
Return a single bucket from EquiWidthBinsDouble::Operation when the span is too small to compute a step size, instead of throwing an InternalException that was reachable from user input (e.g. equi_width_bins(0.0, 5e-324, 1000000, false)).

Fixes duckdb/duckdb#24282

### DIFF
--- a/extension/core_functions/scalar/generic/binning.cpp
+++ b/extension/core_functions/scalar/generic/binning.cpp
@@ -23,6 +23,9 @@ hugeint_t GetPreviousPowerOfTen(hugeint_t input) {
 enum class NiceRounding { CEILING, ROUND };
 
 hugeint_t RoundToNumber(hugeint_t input, hugeint_t num, NiceRounding rounding) {
+	if (num == 0) {
+		return input;
+	}
 	if (rounding == NiceRounding::ROUND) {
 		return (input + (num / 2)) / num * num;
 	} else {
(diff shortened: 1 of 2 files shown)


lnkuiper:
This only catches arithmetic underflow to zero, but a positive step can still be too small to change `bin_boundary`. For example, `SELECT equi_width_bins(1.0, 1.0000000000000002, 3, false);` hangs because `bin_boundary - step == bin_boundary; true` hangs as well. Also, `equi_width_bins(0.0, 5e-324, 1, true)` returns `[nan]` because `step_power_of_ten` becomes zero. These are pre-existing on the base, but they are the same tiny-span path this hunk is making user-safe. Could we guard the actual progress and finite-rounding invariants and fall back to the max boundary?

DinosL:
Since this PR has gotten a bit stale, I opened https://github.com/duckdb/duckdb/pull/25258 to bring it back to life and address the feedback. Thanks @hanafy-dev for your work!

PR #24543, opened 2026-08-06
Add Aggregate Reuse Optimizer to reduce redundant aggregation across joins

kryonix:
I started this work while evaluating whether DuckDB should have a physical GroupJoin operator. The plans that kept looking promising were doing the same expensive work twice: they scanned and grouped a fact-side input to establish a set of qualifying keys, then scanned the same input again to compute an aggregate consumed above a join. A GroupJoin can fuse those operations, but the useful plans do not require a new physical operator. DuckDB's existing joins, aggregates, materialized CTEs, and aggregate-state functions are enough.

The first rewrite handles an exact aggregate that is already computed below a filtering SEMI join. Instead of reading the grouped scalar result and scanning the fact input again, I keep the exported aggregate state, finalize it for the existing HAVING/projection consumers, turn the SEMI join into an INNER join against the unique grouped keys, and combine the repeated states in the original upper grouping. Retaining that upper combine is the important constraint-free part: duplicate owner rows repeat the state exactly as they repeated the original fact rows, so this does not require a primary or unique key on the owner relation. The matcher still requires an exact function, bind data, argument origin, filter/order/distinct mode, row domain, and equality key match.

The second rewrite handles correlated aggregates whose equivalent input is already present in a materialized CTE. It compares the source-relation graph, equality edges, pushed filters, and correlation keys on both sides. The producer may add one relation joined on the complete grouping key. That relation can duplicate rows, so I only reuse aggregates whose function declares that duplicate elimination cannot change the result, such as MIN and MAX; multiplicity-sensitive aggregates such as SUM and COUNT retain the established plan. This is the property needed by the rewrite, rather than declared uniqueness. I split this proof and mutation path into a separate implementation file from exact SEMI-state reuse because their multiplicity rules are different.

The third rewrite preaggregates COUNT below a preserved-side LEFT or RIGHT join. It requires exact or HLL distinct-count statistics for the aggregate-side join key, at least a fourfold estimated collapse, and sufficient join retention. The lower aggregate is estimated from that key NDV, not from the owner cardinality. The original grouping remains above the join: `count(expr)` becomes `sum(coalesce(lower_count, 0))`, while `count(*)` becomes `sum(coalesce(lower_count_star, 1))`. This preserves duplicate owners, NULL owner keys, and the synthetic unmatched row from outer-join semantics. If reliable key NDV is unavailable, or the aggregate would create too many groups for the retained owner domain, the optimizer declines the rewrite.

The same optimizer also has a one-sided eager group-by path for aggregate states below ordinary inner and preserved-side outer joins. The useful cases reduce both join cardinality and the complexity of the first hash aggregate: for example, they group a fact input by two keys before joining dimension attributes that appear in an eight-column upper grouping. I now require the lower partial aggregate to have strictly fewer grouping expressions than the original aggregate. Without that condition, an Appian query grouped by one key below and one key above: the rewrite saved a join but replaced one hash aggregate with a state aggregate plus a second `combine_aggr` pass, producing a repeatable 7% regression. The stricter gate removes every Appian plan change while preserving the TPC-DS wins below.

I removed the catalog-key analysis that previously gated these transformations. It was not a semantic requirement once the compensating upper aggregate was retained, and it prevented the plans from firing on ordinary DuckDB-generated and real-world schemas without declared PK/UNIQUE metadata. The implementation does not infer uniqueness from observed statistics. Statistics are only a profitability gate for the outer-count rewrite.

There are two smaller protections in the same area. The existing double-eager aggregate pushdown now avoids marginal reductions, and the non-nullable `NOT LIKE` fallback uses the complement of the existing LIKE estimate without rounding a nonempty one-row input down to zero. Nullable inputs retain the conservative estimate because SQL NULL semantics prevent taking the complement.

I kept physical execution unchanged. There is no GroupJoin operator, factorized representation, new hash table, spill coordination, setting, serializer, or metric in this PR. The design was informed by Moerkotte and Neumann's [GroupJoin paper](https://www.vldb.org/pvldb/vol4/p843-moerkotte.pdf) and Fent and Neumann's [practical GroupJoin follow-up](https://www.vldb.org/pvldb/vol14/p2383-fent.pdf), but this PR extracts only transformations that fit DuckDB's existing architecture.

I measured the current RelWithDebInfo worktree based on `e9f877bdab` on an Apple M-series machine with one thread. DuckDB's benchmark runner performed one untimed warmup followed by 11 timed executions in one initialized benchmark state. These are same-revision optimizer ablations on ordinary keyless `dbgen` databases; Q2 uses `delim_join_as_cte=true`, matching the plan this optimizer targets.

| Workload | Scale | Disabled median | Enabled median | Change |
| TPC-H Q2 | SF100 | 0.462832 s | 0.257752 s | -44.3% |
| TPC-H Q13 | SF10 | 1.649836 s | 1.379385 s | -16.4% |
| TPC-H Q18 | SF10 | 1.259973 s | 1.210712 s | -3.9% |

For each query I materialized the enabled and optimizer-disabled outputs and compared both directions with `EXCEPT ALL`; all three produced zero differences. The focused tests additionally cover duplicate and nullable owners, LEFT and RIGHT joins, `count(expr)` and `count(*)`, NULL aggregate results, non-key upper groupings, repeated materialized rows, multiple MIN/MAX aggregates, unsafe SUM rejection, changed row domains, volatile expressions, state serialization, and the high-NDV outer-join case from review. In that protection case the rewrite is declined instead of constructing roughly three million lower groups to retain one owner key.

I also ran a complete TPC-DS plan census on the exact PR revision `9241954a3d` using a local SF100 database. `aggregate_reuse` changes none of the 99 plans; `partial_aggregate_pushdown` changes 13 after the stricter profitability gate. Four unchanged queries clear the performance gate with four threads, two warmups, and 15 paired samples in randomized enabled/disabled order. These are medians from the same binary with only `partial_aggregate_pushdown` disabled; the intervals are paired bootstrap 95% intervals for the enabled-plan speedup.

| Query | Enabled median | Disabled median | Speedup | Bootstrap 95% interval |
| TPC-DS Q4 SF100 | 2.567 s | 5.159 s | 101.0% | [100.5%, 101.3%] |
| TPC-DS Q11 SF100 | 1.494 s | 3.326 s | 122.6% | [122.1%, 123.6%] |
| TPC-DS Q36 SF100 | 0.401 s | 0.460 s | 14.7% | [14.4%, 16.0%] |
| TPC-DS Q74 SF100 | 1.401 s | 2.480 s | 77.0% | [44.9%, 97.3%] |

The same plans remain faster with one thread. Exact SF100 outputs for all four queries match with `EXCEPT ALL` in both directions, and both the exact base revision and the final candidate pass all 99 official TPC-DS SF3 answers. Profiles show that the rewrite removes join-expanded work: Q4 reduces join intermediate rows from about 392 million to 208 million, Q11 from 277 million to 146 million, and Q74 from 277 million to 173 million. Planning medians remain between one and four milliseconds, with no more than one millisecond absolute difference at the CLI timer's resolution.


src/include/duckdb/optimizer/partial_aggregate_pushdown.hpp
```diff
@@ -27,6 +27,8 @@ class PartialAggregatePushdown : public LogicalOperatorVisitor {
 	//! Double-eager (Yan & Larson "eager group-by-count"): pre-aggregate BOTH join inputs by the join key and
 	//! reconstruct aggregates above by repeating each side's state by the other side's row count.
 	bool TryDoubleEagerPushdown(unique_ptr<LogicalOperator> &op);
+	//! Pre-aggregate count below a key-preserving owner-outer join.
```
lnkuiper: This functions now intentionally no longer requires uniqueness, maybe we should update the comment

lnkuiper (commented):
Thanks! LGTM now, just need to rebase and re-trigger CI. One tiny comment nitpick:

src/include/duckdb/optimizer/partial_aggregate_pushdown.hpp
kryonix: Makes sense, I've changed the comment.

PR #24544, opened 2026-08-06
Avoid busyloop on task cancellation

dentiny:
In the current implementation, cancellation busy-waits when tasks are still in flight but no tasks are available in the scheduler queue, unnecessarily consuming a CPU core. This PR replaces that polling with notification-based waiting.

Why I notice this issue, CPU cores are burning when my IO tasks failed for the following SQL
```sql
memory D COPY (
             SELECT *
             FROM read_parquet(
                 'hf://datasets/mlfoundations/MINT-1T-HTML/data_v1_1/*.parquet'
             )
         )
         TO '/tmp/MINT-1T-HTML-v1.1.lance'
         (FORMAT lance, MODE 'overwrite');
```


dentiny:
Hi @lnkuiper I'm wondering if you're willing to take a look when you have some time? Thank you!

lnkuiper:

1. This can miss `task_reschedule` wakeups. `WaitForTask` checks completion and queue state while holding `executor_lock`, but `UnregisterTask` now notifies while holding only `producer_lock`. If a task finishes between those checks and `wait_for`, the notification is lost and the caller waits for the full 20ms timeout. Could we preserve the synchronization with `executor_lock` while acquiring the locks in a consistent `executor_lock` -> `producer_lock` order?

2. I think cancellation should use the same dequeue-or-wait structure as `TaskExecutor::DrainTasks`. The current implementation makes `Executor` coordinate `executor_tasks`, the scheduler's approximate per-producer queue size, and `producer_lock` manually, and exposes `GetTaskCountForProducerLocked` solely for this predicate. Dequeueing directly while holding `producer_lock`, executing outside the lock, and otherwise waiting on `producer_cv` would make the dequeue/wait decision atomic and remove the approximate-size dependency and the extra scheduler API.

dentiny:
> Thanks for the PR! A few comments:

Thanks for the review!

`TaskExecutor::DrainTask`'s code structure make sense to me; one difference to mention is, I discard the task directly instead of executing inline on the cancellation thread.

lnkuiper:
Thanks for the changes!

PR #24546, opened 2026-08-06
Test PostgreSQL nested comparison semantics in the deliminator

### LINKED ISSUE #23686: Incorrect result for query with EXISTS for STRUCT
### What happens?

See repro SQL for example, should be straightforward.

### To Reproduce

```sql
memory D CREATE TABLE a(s STRUCT(x INT, y INT));
memory D INSERT INTO a VALUES
         ({'x': NULL, 'y': 1}),
         ({'x': 1, 'y': 2}),
         ({'x': 9, 'y': 9});

memory D CREATE TABLE b(s STRUCT(x INT, y INT));
memory D INSERT INTO b VALUES
         ({'x': NULL, 'y': 1}),
         ({'x': 1, 'y': 2});

-- Correct
memory D SELECT COUNT(*) FROM a WHERE EXISTS (SELECT 1 FROM b WHERE b.s = a.s);
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘

-- Incorrect, should have 2 rows
memory D SELECT COUNT(*) FROM a WHERE NOT EXISTS (SELECT 1 FROM b WHERE b.s = a.s);
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

matthiasgoergens:
Nested `STRUCT` and `LIST` equality follows PostgreSQL composite/array semantics: matching `NULL` children compare equal, while regular equality against a top-level `NULL` still yields `NULL`.

This adds regression coverage for correlated `EXISTS`/`NOT EXISTS` and `LATERAL` plans across both `delim_join_as_cte` modes. The cases exercise delimiter join elimination and generated-domain cleanup, verifying that optimizer rewrites preserve nested comparison semantics without disabling those optimizations.

The PostgreSQL-semantics change supersedes the original result expectation in #23686.

Also fixes a `group_run` typo in the clustered aggregate implementation comment.


src/include/duckdb/planner/expression_nullability.hpp
```diff
@@ -18,6 +18,10 @@ class ClientContext;
 class Expression;
 class LogicalCTE;
 class LogicalOperator;
+class LogicalType;
+
+//! Eliminating a join on a nested comparison changes child-NULL semantics; top-level nullability is not sufficient.
+bool IsNestedComparison(const LogicalType &type);
```
kryonix: `IsNestedComparison` has no comparison-specific behavior: its implementation is exactly `return type.IsNested();`, and these are its only three callers. Putting it in `expression_nullability.hpp` broadens that API and adds a declaration, definition, forward declaration, and include without encoding another invariant. Please use `GetReturnType().IsNested()` directly at the three guards and remove the helper.

kryonix (commented):
Thanks for the PR. I think we can simplify this a bit tough. The `IsNestedComparison` helper is kind of unnecessary and can be removed.

matthiasgoergens:
Good catch — `IsNestedComparison` was just `type.IsNested()` with no comparison-specific logic, so I've inlined it at the three call sites and removed it from `expression_nullability.hpp`.

PR #24548, opened 2026-08-06
Allow fitting integer literals as typed macro defaults

### LINKED ISSUE #24401: some data types to MACRO parameters unexpectedly require explicit cast of the default value
### What happens?

The CREATE MACRO statement can fail with:
```
Default value '<value>' for parameter '<param>' cannot be implicitly cast to '<type>'. Please add an explicit type cast.
```
where <value> is a literal that would in other context be accepted without explicit cast, <param> is the name of the parameter, and <type> is the declared type

### To Reproduce

```sql
create or replace macro test(param SMALLINT := 1) as 1;

Binder Error:
Default value '1' for parameter 'param' cannot be implicitly cast to 'SMALLINT'. Please add an explicit type cast.
```

similar: 
```sql
create or replace macro test(param TINYINT := 1) as 1;
create or replace macro test(param UBIGINT := 1) as 1;
create or replace macro test(param UHUGEINT := 1) as 1;
create or replace macro test(param UINTEGER := 1) as 1;
create or replace macro test(param USMALLINT := 1) as 1;
create or replace macro test(param UTINYINT := 1) as 1;
```

### OS:

windows 11

### DuckDB Version:

1.5.5

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Roland Bouman

### Affiliation:

EPAM Systems BV

- [x] Yes, I have

- [x] Yes, I have

matthiasgoergens:
Fixes #24401.

Typed macro defaults were checked as ordinary folded values, so a literal `1`
could not default a `TINYINT`, `SMALLINT`, or unsigned parameter even though
the same literal is accepted in an equivalent explicit cast.

The patch retains literal provenance for the default expression and applies
DuckDB's existing value-aware `INTEGER_LITERAL` implicit-cast rule. Computed
and explicitly typed constants are intentionally excluded — only genuine
integer literals that fit the target type are accepted.

### Tests

Focused, broader macro, boundary, table-macro, negative-control, and
persistence checks pass; the regression fails against the unchanged build.

Verified: `test/sql/catalog/function/test_macro_type_overloads.test` fails on
main `033322be14` (`Binder Error: Default value '1' for parameter 'i' cannot
be implicitly cast to 'TINYINT'`), passes with this branch (131 assertions).

### DIFF
--- a/src/planner/binder/statement/bind_create.cpp
+++ b/src/planner/binder/statement/bind_create.cpp
@@ -444,11 +444,16 @@ SchemaCatalogEntry &Binder::BindCreateFunctionInfo(CreateInfo &info) {
 		}
 
 		// Constant-fold all default parameter expressions
+		identifier_set_t integer_literal_defaults;
 		for (auto &it : function->default_parameters) {
 			auto &param_name = it.first;
 			auto &param_expr = it.second;
 
 			if (param_expr->GetExpressionType() == ExpressionType::VALUE_CONSTANT) {
+				auto &value = param_expr->Cast<ConstantExpression>().GetValue();
+				if (value.type().IsIntegral() && !value.IsNull()) {
+					integer_literal_defaults.insert(param_name);
+				}
 				continue;
 			}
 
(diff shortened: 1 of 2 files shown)


lnkuiper:
Could you have a look at the failing CI?

matthiasgoergens:
> Could you have a look at the failing CI?

Fixed now.  Thanks for checking!

PR #24551, opened 2026-08-06
Fix Top-N window elimination for nullable ordering expressions

### LINKED ISSUE #23677: top_n_window_elimination can drop rows when ROW_NUMBER orders by a nullable expression
### What happens?

top_n_window_elimination can drop rows when ROW_NUMBER orders by a nullable expression of a non-null key.

### To Reproduce

To reproduce:

```sql
CREATE TABLE t(id INTEGER PRIMARY KEY);

SELECT
  id,
  ROW_NUMBER() OVER (
    ORDER BY NULLIF(id % 3, 0) DESC NULLS LAST
  ) AS rn
FROM t
QUALIFY rn <= 3
ORDER BY rn;
```

Run:

```sh
```

Actual:

stdout:
| id | rn |
| --- | --- |
| 2 | 1 |
| 1 | 2 |

Expected:

| id | rn |
| --- | --- |
| 2 | 1 |
| 1 | 2 |
| 3 | 3 |

Notes From Minimization:

- Keeping a nullable order expression is required; using a proven non-NULL order key avoids the null-handling mismatch.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (a1aacc73df)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

matthiasgoergens:
`TopNWindowElimination::ExtractOptimizerParameters` inferred the nullability of a `ROW_NUMBER()` ordering expression from the statistics of its referenced columns, but a non-null column does not imply a non-null expression — `NULLIF`, `TRY_CAST` and similar constructs can introduce NULL values. The rewrite's `min_n`/`arg_max` aggregate path ignores NULL keys, so rows whose ordering expression evaluates to NULL were silently discarded, returning fewer rows than the query should.

The fix runs `NotNullExpressionAnalyzer` over the complete ordering and payload expressions and selects the existing null-preserving aggregate path when it cannot prove both non-null. The analyser is deliberately conservative, so uncertainty costs only an optimisation opportunity, never rows.

Extended `test/optimizer/topn_window_elimination.test_slow` covers payload and payload-free rewrites, an entirely NULL ordering expression, a `TRY_CAST` expression that introduces NULLs, partitions, and the multi-payload struct-packing path. Fixes #23677.

### DIFF
--- a/src/include/duckdb/planner/expression_nullability.hpp
+++ b/src/include/duckdb/planner/expression_nullability.hpp
@@ -1,0 +1,37 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// duckdb/planner/expression_nullability.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/optional_ptr.hpp"
+#include "duckdb/common/vector.hpp"
+
+namespace duckdb {
+
+class ClientContext;
+class Expression;
+class LogicalCTE;
+class LogicalOperator;
+
+//! Conservatively proves that an expression cannot be NULL at a logical operator's output.
+class NotNullExpressionAnalyzer {
+public:
+	explicit NotNullExpressionAnalyzer(ClientContext &context, optional_ptr<LogicalOperator> plan_root = nullptr);
+
+	bool IsNotNull(LogicalOperator &op, const Expression &expr);
+
+private:
+	bool IsNotNull(LogicalOperator &op, const Expression &expr, vector<idx_t> &seen_ctes);
+	optional_ptr<LogicalCTE> FindCTE(idx_t cte_index);
+
+private:
+	ClientContext &context;
+	optional_ptr<LogicalOperator> plan_root;
+};
+
+} // namespace duckdb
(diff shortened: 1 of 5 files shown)


d-justen (changes requested):
Hi @matthiasgoergens, this is looking pretty good already. Could you please re-target to `v1.5-variegata`?

matthiasgoergens:
Retargeted to `v1.5-variegata`.

lnkuiper:
This pulled in all changes from main to v1.5. Could you apply just your changes to v1.5?

matthiasgoergens:
Done — the branch is now based directly on `v1.5-variegata` with only the fix and test commits. The port needed `NotNullExpressionAnalyzer` backported from main (it doesn't exist on this branch), adapted to v1.5's statistics API; unknown and empty statistics conservatively take the null-preserving aggregate path. The full `topn_window_elimination.test_slow` passes locally (289 assertions), including the new NULLIF/TRY_CAST cases.

lnkuiper:
Thanks!

PR #24495, opened 2026-08-04
Fix variant shredding dropping case-distinct keys

DinosL:
If the shredding schema spells a field differently from the data (e.g. A vs a), the value was silently dropped. This PR makes field name matching against the shredding schema case-sensitive, both when writing shredded data and when reading it back.

Fixes: https://github.com/duckdb/duckdb/issues/24297
Fixes: https://github.com/duckdb/duckdb/issues/24379

PR #24553, opened 2026-08-06
Avoid shredding colliding VARIANT keys

### LINKED ISSUE #24298: Automatic VARIANT shredding with a/A makes stats(v) fail
### What happens?

`See repro for details.

### To Reproduce

```sql
memory D SET variant_minimum_shredding_size = 0;
memory D RESET force_variant_shredding;
memory D 
memory D CREATE TABLE t(v VARIANT);
memory D INSERT INTO t VALUES ({'a': 1}::VARIANT), ({'A': 2}::VARIANT);
memory D CHECKPOINT;
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D 
memory D SELECT stats(v) FROM t LIMIT 1;
Not implemented Error: Error while casting - duplicate name "a" in struct
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

matthiasgoergens:
Fixes #24298.

VARIANT object keys are case-sensitive, but STRUCT field identifiers are
case-insensitive. Automatic shredding inferred separate fields for keys such
as `a` and `A`, then attempted to represent both in a STRUCT; reading the
column's statistics consequently failed with a duplicate-field error.

Detect case-insensitive key collisions while inferring an object's shredded
type and decline to shred that object. For a nested collision, that whole
nested object — including its non-colliding sibling fields — remains in the
overlay, while unrelated parent fields outside it can still be shredded.
Neither key spelling is merged or discarded. `force_variant_shredding`
bypasses inference, so its validator now rejects shredding types whose
STRUCT fields collide case-insensitively at any nesting depth.

The regression covers statistics, vector shape, case-sensitive extraction
and keys after a checkpoint, nested fallback with an unrelated parent field
that stays shredded, and rejection of colliding forced types.

### DIFF
--- a/src/main/settings/custom_settings.cpp
+++ b/src/main/settings/custom_settings.cpp
@@ -39,6 +39,7 @@
 #include "duckdb/logging/logger.hpp"
 #include "duckdb/logging/log_manager.hpp"
 #include "duckdb/common/type_visitor.hpp"
+#include "duckdb/common/case_insensitive_map.hpp"
 #include "duckdb/function/variant/variant_shredding.hpp"
 #include "duckdb/storage/block_allocator.hpp"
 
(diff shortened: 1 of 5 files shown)


Tishj:
I wonder how this fix relates to https://github.com/duckdb/duckdb/pull/24495
Because they feel like they're fixing the same problem, no?

matthiasgoergens:
They address different cases. #24495 makes matching against the shredding schema case-sensitive on both write and read, so if the schema contains `a`, a data key named `A` remains in the unshredded remainder instead of being dropped. This PR handles schemas that would themselves contain both `a` and `A`: inference declines to shred that object, and `force_variant_shredding` rejects such a schema because STRUCT field names are case-insensitive. So this PR does not subsume #24495's case-mismatch fix.

matthiasgoergens:
The Windows 64-bit failure is the same cache-restore infrastructure error (sh.exe exit code 52) that hit the 32-bit job on the previous run — the 32-bit job passed when re-rolled, and the same branch's Windows 64-bit job passes on my fork (matthiasgoergens/duckdb actions run 31675631215). No test failures anywhere else.

PR #24557, opened 2026-08-06
Resolve partition columns for non-pushdown table functions

### LINKED ISSUE #24327: A table function that sets `get_partition_info` but leaves `projection_pushdown = false` can crash the physical planner.
### What happens?

This error is raised:

INTERNAL Error: Attempted to access index 1 within vector of size 1: 

### What happens

Stack (release build):

```
duckdb::vector<duckdb::ColumnIndex, true>::operator[](unsigned long) const
duckdb::PhysicalPlanGenerator::CreatePlan(duckdb::LogicalAggregate&)
duckdb::PhysicalPlanGenerator::CreatePlan(duckdb::LogicalProjection&)
```

It fires at plan time — `EXPLAIN` is enough, the scan never runs. A milder form of the same defect silently mis-resolves the partition column and quietly disables the `PARTITIONED_AGGREGATE` optimization.

### Root cause

`PhysicalPlanGenerator::HasSingleValuePartitions` (`src/execution/physical_plan/plan_aggregate.cpp:97-107` on main) walks down through `PROJECTION`/`FILTER` nodes to the `PhysicalTableScan`, then maps the resulting indices to base columns:

```cpp
if (!table_scan.projection_ids.empty()) {
    for (auto &partition_col : partition_columns) {
        partition_col = table_scan.projection_ids[partition_col];
    }
}
vector<column_t> base_columns;
for (const auto &partition_idx : partition_columns) {
    auto col_idx = partition_idx;
    col_idx = table_scan.column_ids[col_idx].GetPrimaryIndex();   // <-- unchecked, and wrong space
    base_columns.push_back(col_idx);
}
```

This assumes `partition_idx` indexes the scan's **output**. That holds only for projection-pushdown scans. For a non-pushdown function, `plan_get.cpp:149-153` builds a scan whose output is **all** `returned_types` in base order, with `projection_ids` empty and `column_ids` holding the narrowed/reordered request, then stacks a compensating projection on top whose `BoundReferenceExpression`s carry **base column indices**:

```cpp
auto col_id = column_id.GetPrimaryIndex();
expressions.push_back(make_uniq<BoundReferenceExpression>(type, col_id));
```

So by the time the walk reaches the scan, `partition_columns` is *already* in base-column space. Indexing `column_ids` with it maps a second time, through a shorter, differently-ordered vector — out of range when the narrowed list is short enough, wrong-but-in-range otherwise.

With output `(region=0, year=1, value=2)` and both partition columns declared:

| query | `column_ids` | walk yields | `column_ids[…]` | outcome |
|---|---|---|---|---|
| `GROUP BY region, year`, `SUM(value)` | `[0,1,2]` | `[0,1]` | `[0,1]` | correct (identity) |
| `GROUP BY region` | `[0,2]` | `0` | `0` | correct by coincidence |
| `GROUP BY year`, `SUM(value)` | `[1,2]` | `1` | `2` → `value` | wrong column, silent `HASH_GROUP_BY` |
| `GROUP BY year`, `COUNT(*)` | `[1]` | `1` | — | **crash** |

The crash precedes the `get_partition_info` call, so the callback's behavior is irrelevant to reproducing it.

### Reproduction

This is **not reachable from pure SQL** on a stock build, so the repro needs a table function from outside core. `MultiFileFunction` is the only core source that installs `get_partition_info`, and it sets `projection_pushdown = true` two lines earlier (`src/include/duckdb/common/multi_file/multi_file_function.hpp:78,80`), so every SQL-reachable source takes the correct branch. I confirmed hive-partitioned parquet plans cleanly for `GROUP BY <2nd hive key>` with `COUNT(*)`/`SUM`, with and without a filter on the first key, with `filename=true`, through a view, and under `SET disabled_optimizers='unused_columns'` (that setting doesn't flip the branch — it keys on the static `function.projection_pushdown`, not on whether narrowing happened — it only widens `column_ids` to all-in-order, which makes the second mapping an identity).

The minimum trigger is small:

```cpp
// >= 2 output columns; the grouped one must not be column 0
names = {"region", "year", "value"};
types = {LogicalType::VARCHAR, LogicalType::BIGINT, LogicalType::DOUBLE};

TableFunction f("single_value_parts", {}, Scan, Bind, InitGlobal, InitLocal);
f.projection_pushdown = false;                 // the trigger
f.get_partition_info = [](ClientContext &, TableFunctionPartitionInput &) {
    return TablePartitionInfo::NOT_PARTITIONED;  // never reached — crash is upstream of it
};
```

```sql
EXPLAIN SELECT year, COUNT(*) FROM single_value_parts() GROUP BY year;
-- INTERNAL Error: Attempted to access index 1 within vector of size 1
```

Setting `f.projection_pushdown = true` makes both the crash and the wrong-column fallback disappear (the planner then resolves `year` → base column 1 and picks `PARTITIONED_AGGREGATE`) — which is the workaround, but it forces functions to implement pushdown just to avoid a planner crash.

### Suggested fix

Skip the second mapping when the scan is the non-pushdown shape:

```cpp
vector<column_t> base_columns;
if (!table_scan.function.projection_pushdown) {
    // plan_get.cpp emits returned_types in base order and adds a compensating
    // projection carrying base indices, so partition_columns is already in base space
    base_columns = partition_columns;
} else {
    if (!table_scan.projection_ids.empty()) { /* existing remap */ }
    for (const auto &partition_idx : partition_columns) {
        base_columns.push_back(table_scan.column_ids[partition_idx].GetPrimaryIndex());
    }
}
```

Identity is correct for both non-pushdown sub-cases: with the compensating projection the walk already translated to base indices, and without it (all columns, in order) output slot == base index. A bounds check before the `column_ids` lookup would be worth keeping regardless — turning any residual mismatch into "decline the optimization" rather than an internal error. That matters here because the failure mode for an extension author is an unexplained `INTERNAL Error` in the planner with no pointer back to their function, and `get_partition_info` / `projection_pushdown` are otherwise orthogonal features with nothing documenting the coupling.

### Blast radius

No stock-DuckDB user can hit this today; the exposure is entirely on extension authors who implement `get_partition_info` without projection pushdown.

On `main`, `HasSingleValuePartitions` was hoisted out of `CanUsePartitionedAggregate` and is now also called from `plan_window.cpp:44` for `OVER (PARTITION BY …)`. By inspection the same crash should be reachable from a window query on such a source; I have not tested that (my build predates the refactor).

### Version

Reproduced on a 1.5.4-based build.

### To Reproduce

see above

### OS:

MacOS

### DuckDB Version:

1.5.4

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Rusty Conover

### Affiliation:

Query Farm - https://query.farm

- [x] Yes, I have

- [x] Yes, I have

Yes

matthiasgoergens:
`PhysicalPlanGenerator::HasSingleValuePartitions` walks projections and filters to map aggregate and window partition expressions to a table scan, mapping scan output positions through `projection_ids` and `column_ids` for functions with projection pushdown. A function with `projection_pushdown = false` has a different physical shape: the scan outputs every base column in schema order, and `plan_get.cpp` adds a projection whose bound references already contain base-column indices. The helper nevertheless indexed the narrowed `column_ids` with those base indices a second time, silently querying partition information for the wrong column or throwing an internal out-of-range exception during planning.

The fix uses the already-resolved indices directly for non-pushdown scans, retains the existing mapping for pushdown scans, and adds defensive bounds checks before indexing the projection vectors.

A new C++ table function deliberately disables projection pushdown and makes its partition callback reject any column mapping other than the expected base indices; its regression covers a narrowed `GROUP BY`, an intervening filter and projection, reordered two-column grouping, and the window-planner caller. Fixes #24327.

### DIFF
--- a/src/execution/physical_plan/plan_aggregate.cpp
+++ b/src/execution/physical_plan/plan_aggregate.cpp
@@ -96,16 +96,26 @@ bool PhysicalPlanGenerator::HasSingleValuePartitions(ClientContext &context,
 		return false;
 	}
 	// get the base columns by projecting over the projection_ids/column_ids
-	if (!table_scan.projection_ids.empty()) {
-		for (auto &partition_col : partition_columns) {
-			partition_col = table_scan.projection_ids[partition_col];
-		}
-	}
 	vector<column_t> base_columns;
-	for (const auto &partition_idx : partition_columns) {
-		auto col_idx = partition_idx;
-		col_idx = table_scan.column_ids[col_idx].GetPrimaryIndex();
-		base_columns.push_back(col_idx);
+	if (!table_scan.function.projection_pushdown) {
+		// Non-pushdown scans output every base column in order. Any projection above the scan already maps references
+		// into that base-column space.
+		base_columns = partition_columns;
+	} else {
+		if (!table_scan.projection_ids.empty()) {
+			for (auto &partition_col : partition_columns) {
+				if (partition_col >= table_scan.projection_ids.size()) {
+					return false;
+				}
+				partition_col = table_scan.projection_ids[partition_col];
+			}
+		}
+		for (const auto &partition_idx : partition_columns) {
+			if (partition_idx >= table_scan.column_ids.size()) {
+				return false;
+			}
+			base_columns.push_back(table_scan.column_ids[partition_idx].GetPrimaryIndex());
+		}
 	}
 	// check if the source operator is partitioned by the grouping columns
 	TableFunctionPartitionInput input(table_scan.bind_data.get(), base_columns);
--- a/test/sql/function/table/CMakeLists.txt
+++ b/test/sql/function/table/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_library_unity(test_table_function OBJECT table_in_out.cpp
-                  table_bind_replace.cpp)
+                  table_bind_replace.cpp table_partition_info.cpp)
 set(UNITTEST_OBJECT_FILES
     ${UNITTEST_OBJECT_FILES} $<TARGET_OBJECTS:test_table_function>
     PARENT_SCOPE)
--- a/test/sql/function/table/table_partition_info.cpp
+++ b/test/sql/function/table/table_partition_info.cpp
@@ -1,0 +1,76 @@
+#include "catch.hpp"
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/function/partition_stats.hpp"
+#include "duckdb/parser/parsed_data/create_table_function_info.hpp"
+#include "test_helpers.hpp"
+
+using namespace duckdb;
+
+namespace {
+
+struct NonPushdownPartitionFunction {
+	struct BindData : public TableFunctionData {
+		vector<column_t> expected_partition_ids;
+	};
+
+	static unique_ptr<FunctionData> Bind(ClientContext &, TableFunctionBindInput &input,
+	                                     vector<LogicalType> &return_types, vector<string> &names) {
+		return_types = {LogicalType::VARCHAR, LogicalType::BIGINT, LogicalType::DOUBLE};
+		names = {"region", "year", "value"};
+		auto result = make_uniq<BindData>();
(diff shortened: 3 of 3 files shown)


Mytherin:
Thanks!

PR #24561, opened 2026-08-06
Cheaper Duplicate Eliminated Domain Derivation

lnkuiper:
This PR improves decorrelated subquery plans by deriving duplicate-eliminated domains from cheaper parts of the outer plan instead of always materializing the complete outer payload.

The existing decorrelation machinery can generate an expensive domain. For example, TPC-H Q17 only needs the filtered `part` keys to restrict the correlated `lineitem` aggregate. Constructing that domain from the full outer `lineitem JOIN part` payload does unnecessary work, while removing the restriction altogether aggregates all of `lineitem` before applying the selective `part` filter.

## Duplicate-Eliminated Domains

This adds an optional strategy to the existing DelimJoin CTE lowering boundary. `FlattenDependentJoins` remains responsible for correctness; the new optimizer only chooses a cheaper implementation of the duplicate-eliminated domain.

The optimizer traces correlated keys through an allowlist of projections, filters, aggregates, inner joins, cross products, semi joins, and reusable CTE references. It classifies candidates as exact domains or safe supersets and uses relation statistics to compare their input and distinct-key cardinalities against the complete outer payload.

For an eligible candidate it can:

- inline a duplicate-free copy of the candidate into the generated RHS restriction; or
- factor the candidate into a CTE shared by the original payload path and the duplicate-eliminated domain.

We bail out for volatile, stateful, and fallible expressions, as these can cause a plan to, e.g., throw in the optimised plan when they otherwise wouldn't. In these cases, the existing full-payload domain path is retained.

For TPC-H Q17 this removes the generated domain CTE while retaining the filtered `part` domain below the correlated aggregate. TPC-DS Q1 similarly derives the store-key domain from the filtered `store` input rather than from the complete outer payload.

## Performance

Measured with two threads (like CI does). Median of 5:

| Query | Scale | Base | This PR | Change |
|---|---:|---:|---:|---:|
| TPC-H Q17 | SF1 | 17.0 ms | 14.5 ms | 14.7% faster |
| TPC-H Q17 | SF10 | 183.5 ms | 160.5 ms | 12.5% faster |
| TPC-H Q17 | SF100 | 1.815 s | 1.606 s | 11.5% faster |
| TPC-DS Q1 | SF1 | 6.5 ms | 6.5 ms | unchanged |
| TPC-DS Q1 | SF10 | 32.0 ms | 31.0 ms | 3.1% faster |
| TPC-DS Q1 | SF100 | 330.0 ms | 321.0 ms | 2.7% faster |

Added microbenchmarks:

| Microbenchmark | Base | This PR | Change |
|---|---:|---:|---:|
| Correlated scalar join | 2.907 ms | 2.281 ms | 21.5% faster |
| Pair-dependent `IN` | 519.3 ms | 513.3 ms | 1.2% faster |
| Factored outer join | 1.338 s | 1.300 s | 2.8% faster |
| Reused `EXISTS`/`UNNEST` CTE | 3.675 ms | 3.711 ms | unchanged |


kryonix (commented):
## Re-invented infrastructure

There are three areas where infrastructure is now duplicated:

- [`DuplicateEliminatedDomainCTERegistry`](https://github.com/duckdb/duckdb/blob/855d47476012e63117e06607ae42c052fceaaac2/src/include/duckdb/optimizer/duplicate_eliminated_domain/duplicate_eliminated_domain_cte_registry.hpp#L21-L47) walks the logical tree and records CTE definitions. [`DuplicateFreeBindingAnalyzer`](https://github.com/duckdb/duckdb/blob/855d47476012e63117e06607ae42c052fceaaac2/src/planner/subquery/duplicate_eliminated_domain_properties.cpp#L110-L131) independently performs essentially the same traversal. The latter should consume the registry, or the registry should become a general logical CTE registry.

- `DuplicateFreeBindingAnalyzer` adds another logical uniqueness model while [`JoinElimination`](https://github.com/duckdb/duckdb/blob/855d47476012e63117e06607ae42c052fceaaac2/src/optimizer/join_elimination.cpp#L39-L79) already propagates distinct groups through `DISTINCT`, aggregates and projections. They cannot be substituted directly, but two optimizers now encode overlapping uniqueness semantics. The immediate drift is visible already: the new analyzer handles aggregates but not `LOGICAL_DISTINCT`. A shared unique-key property analysis would be preferable.

- The rewriter has a private linear [`FindBindingIndex`](https://github.com/duckdb/duckdb/blob/855d47476012e63117e06607ae42c052fceaaac2/src/planner/subquery/delim_join_cte_rewriter.cpp#L87) used repeatedly even though `ColumnBindingLayout` already owns a binding-to-position map. Adding a non-throwing `TryGetPosition` method to `ColumnBindingLayout` would remove this duplication.

## Possible simplifications in the future

The largest simplification opportunity is phase placement. The strategy currently runs during planning, before normal optimizer passes such as global filter pushdown, CTE inlining and join elimination. Consequently, it needs its own:

- filter-pushdown invocation;
- CTE materialization reasoning;
- uniqueness analysis;
- expression-safety traversal;
- relation statistics;
- copy/remap infrastructure.

A longer-term design could retain a logical duplicate-domain candidate marker through planning and make the cost/safety decision after ordinary optimizer properties are available. Lowering to a materialized CTE could happen afterward. That is a larger pipeline change, but it would remove several local mirrors of optimizer infrastructure.

Other worthwhile reductions:

- `DelimJoinCTEOptimization`, its decision base class and its three-state result are only used by this strategy. `TryOptimize` then downcasts the base decision with `static_cast`. A typed concrete decision/result would be smaller and safer until a second implementation actually needs the abstraction.

- The safety evaluator manually mirrors expression-throwing and evaluation semantics, partly overlapping `CanThrow`, volatility and repeatability infrastructure. Its conservative fallback protects correctness, but future expression classes will silently become missed optimizations. “Safe to evaluate over an expanded domain” should ideally become a reusable expression property.

- The rewriter now carries several loosely related booleans, including `under_grouping_operator`, `has_grouped_domain_guard`, `can_evaluate_additional_groups` and `preserve_selected_evidence`. Tracking guarded key bindings and evidence provenance explicitly would be easier to reason about than propagating boolean state through selected operator shapes.

lnkuiper:
Thanks for the review! Moving this to a draft for now, I think I might need to split into two PRs.

PR #24562, opened 2026-08-06
Fix infinite loop in `equi_width_bins` via zero step size

### LINKED ISSUE #24282: Bug8: Infinite Loop in equi_width_bins() via Zero Step Size

tempewda:
Fixes #24282

This PR fixes multiple vulnerabilities in `equi_width_bins` caused by a calculated step size of zero, which previously led to an infinite loop (CPU exhaustion) or division-by-zero crashes.

1. Integer Variant: Added a `step == 0` guard in `EquiWidthBinsInteger`. I chose to throw an `InvalidInputException` instead of artificially clamping the step to `1`.

2. Double Variant: Moved the existing `step == 0` guard in `EquiWidthBinsDouble` to the top of the function (before `GetPreviousPowerOfTen` and `MakeNumberNice`) to catch subnormal floating-point underflows earlier. I also reclassified it from `InternalException` to `InvalidInputException`, as this is caused by mathematically impossible user input, not an internal engine logic failure.

3. Nice Rounding Crash: By placing the integer guard before the `nice_rounding` block, this PR also fixes a secondary bug where `MakeNumberNice(0, 0)` would trigger a hard SIGFPE division-by-zero crash if `nice_rounding` was set to true.

After the fix:
```
-- Integer Bug: Massive bin count safely rejected before hanging the engine
memory D SELECT equi_width_bins(1::BIGINT, 2::BIGINT, 1000000, false);
Invalid Input Error:
Invalid input for equi_width_bins - the bin count is too large for the given range, resulting in a step size of 0

LINE 1: SELECT equi_width_bins(1::BIGINT, 2::BIGINT, 1000000, false);
               ^
-- Double Bug: Subnormal float underflow safely rejected instead of causing internal crash
memory D SELECT equi_width_bins(0.0::DOUBLE, 5e-324::DOUBLE, 2, false);
Invalid Input Error:
Invalid input for equi_width_bins - the bin count is too large for the given range, resulting in a step size of 0

LINE 1: SELECT equi_width_bins(0.0::DOUBLE, 5e-324::DOUBLE, 2, false);
               ^
```

### DIFF
--- a/extension/core_functions/scalar/generic/binning.cpp
+++ b/extension/core_functions/scalar/generic/binning.cpp
@@ -130,6 +130,11 @@ struct EquiWidthBinsInteger {
 
 		const hugeint_t span = max - min;
 		hugeint_t step = span / Hugeint::Convert(bin_count);
+		if (step == 0) {
+			throw InvalidInputException(expr,
+			                            "Invalid input for equi_width_bins - the bin count is too large for the given "
+			                            "range, resulting in a step size of 0");
+		}
 		if (nice_rounding) {
 			// when doing nice rounding we try to make the max/step values nicer
 			hugeint_t new_step = MakeNumberNice(step, step, NiceRounding::ROUND);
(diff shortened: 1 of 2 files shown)


tempewda:
Superseded by #24538.

PR #24391, opened 2026-08-01
Try -march x86-64-v3 for CLI and v2 for extensions

smvv:
### Context

Enable the CPU instructions of "modern" (last 10 years) hardware.

`x86-64-v3` is supported by Haswell Xeon servers (appeared in 2014) and all AMD EPYC systems.

This PR is part of a larger project to [enable](https://github.com/duckdb/duckdb/pull/24493) Clang + Link-time optimizations + CPU profile v3 because all combined they give **12-17% improvement on the geomean** of all PR-ran benchmarks:
```
 Bench job                   Geomean     Change
━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━  ━━━━━━━━━
 Micro               0.279s → 0.231s     −17.0%
 Recursive CTE       0.159s → 0.134s     −15.7%
 TPCH-PARQUET SF1    0.046s → 0.039s     −15.5%
 H2OAI               0.356s → 0.303s     −14.9%
 TPCDS SF3           0.025s → 0.021s     −13.7%
 IMDB                0.133s → 0.115s     −13.7%
 AoC24               0.156s → 0.135s     −13.6%
 Ingestion Perf      0.865s → 0.747s     −13.6%
 CSV                 0.298s → 0.261s     −12.4%
 TPCH SF10           0.109s → 0.096s     −12.3%
 RealNest            0.131s → 0.119s      −9.0%
```

Without enabling these CPU profile flags, Clang is conservative in selecting CPU instructions and switching to Clang would result in some minor improvements and some regressions.

Note: this is **not** related to `manylinux_2_28` because that specifies a libc/ABI compatibility; ABI compatibility does not guarantee what CPU instructions the machine supports.

### Use `clang-cl` for windows CI jobs

`clang-cl.exe` is an alias to clang.exe that automatically configure it in MSVC compatible mode. It supports MSVC flag syntax, library search path, ABI, etc.

Over time, this allows us to use clang for most platforms and CI jobs, but this is future work.

### Benchmark results

This PR shows modest benchmark geomean improvements (~3-4%), and makes it possible to adopt Clang + LTO for the full CI pipeline.

#### H20AI

```
FASTER
benchmark  base median                   PR median                     delta median       runs
---------  ----------------------------  ----------------------------  -----------------  ----
group/q09  166.6 ms [164.2 ms…173.1 ms]  161.1 ms [158.6 ms…185.8 ms]    -5.4 ms (-3.3%)    40
[...]

PR #24564, opened 2026-08-06
Unify and optimize scalar executor infrastructure

kryonix:
I ran into a split in DuckDB's scalar executor infrastructure: unary and binary execution had carefully specialized flat and constant paths, while ternary and higher-arity execution went through the generic unified-vector loop. The implementations solved the same representation, validity, and selection problems independently. That made the hot paths difficult to improve consistently, and it meant a cheap ternary operation still paid selection-vector and per-input validity overhead when every input was flat or constant.

The violated invariant was in the vector executor layer. Arity should determine how arguments are expanded and how an operation is called; it should not determine which representation and validity optimizations are available. I moved that common work into an internal `ScalarExecutor` and kept `UnaryExecutor`, `BinaryExecutor`, `TernaryExecutor`, and `VariadicExecutor` as compatibility facades. Their existing entry points and operation conventions remain intact.

The shared engine classifies every input as constant, flat, or generic. It evaluates all-constant inputs once, specializes flat/constant masks through arity three, combines validity a machine word at a time, and falls back to `UnifiedVectorFormat` for other representations. Small adapters preserve lambda, optional-return, unary generic/state, unary string, binary single-type, binary fully typed, and variadic static-operation conventions. The unary facade retains dictionary-domain evaluation and its `FunctionErrors::CANNOT_ERROR` guard, while the binary facade retains comparison-complement folding.

There are two deliberate exceptions to a completely uniform facade. First, the initial shared binary true-only selection loop regressed a measured hot path, so I retained a narrow binary-specific flat loop for that output mode. Second, unary execution preserves a pre-seeded result validity mask when all explicit inputs are valid. Arrow extension conversion relies on this existing contract: it places the Arrow NULL mask on the result before converting an all-valid temporary storage vector. Operations that can add NULLs copy that mask before mutating it. Ternary and variadic execution do not preserve an old result mask, so reused results still clear stale NULLs there. The all-constant path also explicitly clears valid/NULL/valid reuse state.

I bounded specialization because instantiating every mask and every selection-output combination quickly outweighed the runtime benefit. Execution specializes arithmetic inputs through arity three. Variadic selection specializes only the all-flat input mask; adding all ternary input masks increased text by more than 1 MiB without a corresponding measured gain. It does retain compile-time true-only, false-only, and dual-output loops. I initially dispatched output presence at runtime, but profiling the Recursive CTE `game_of_life` regression showed that the resulting pointer checks made the shared BETWEEN loop substantially slower. `DUCKDB_SMALLER_BINARY` can remove these two dimensions independently through `variadic_executor_select_flat` and `variadic_executor_select_flags` in the existing `vector_specialization` group.

The follow-up review also surfaced the small binary-executor change in [duckdb/duckdb#24391](https://github.com/duckdb/duckdb/pull/24391). I evaluated the same four input-selection masks at the shared-engine boundary. The generated code looked cleaner, but direct all-valid measurements on this arm64 host were within noise and exact TPC-H Q1 was reproducibly about 4% slower. I therefore did not retain the all-valid specialization here. The ordinary all-valid path stays compact, while the facade uses the selection masks only for nullable generic binary execution, where they have a measured benefit. This is controlled by `binary_executor_generic_nullable` in the `vector_specialization` smaller-binary group.

I measured the change against `daa81697e31a3dc97a93f11220037cd2213af6cd` on arm64 macOS 26.5 with Apple Clang 21, release optimization, and isolated baseline/candidate builds with compiler launchers disabled. The direct executor benchmark uses a standard-size vector and one million in-process iterations for nonconstant profiles. Paired process-level measurements kept unary flat (923.3 ms versus 919.0 ms), unary dictionary (3.116 s versus 3.128 s), binary flat/flat (1.248 s versus 1.262 s), and binary flat/constant and constant/flat at parity or slightly faster. The retained binary true-only selection path measured about 1% faster than the pinned implementation across eight in-process samples.

I added direct nullable dictionary/flat, flat/dictionary, and dictionary/dictionary binary cases to make that distinction measurable. Dispatching the selection profile once per vector improved them from 10.523 s to 7.477 s, 10.101 s to 8.297 s, and 10.447 s to 8.747 s respectively, or 29%, 18%, and 16%. The row loop tests each validity mask at the selected source index and avoids materializing a mapped validity mask before doing any useful work.

The advisory end-to-end jobs later showed a smaller caveat that those direct benchmarks missed. Against the exact CI base, TPC-H SF10 moved by +1.7% and TPC-DS SF3 by +2.0%; neither crossed the benchmark runner's regression threshold. The largest TPC-H movement, Q15 at +26.8%, came from a host slowdown phase landing on opposite sides of the two ordered medians rather than from the branch. Repeating Q15 and the largest TPC-DS movements locally with exact base/candidate Release builds and both process orders reduced the selected-query geometric-mean difference to roughly +0.5%. TPC-DS Q85 nevertheless retained a small signal: four 100-run processes put the candidate 2.28% behind using the mean of the per-process medians.

Sampling Q85 pointed to binary generic-constant selection. The pinned implementation loaded only the dictionary operand's selection index. The first shared implementation loaded an index for both operands, including the constant operand's zero selection, and generic comparison-complement folding materialized both unified formats once to inspect nullability and then again for execution. In comparable ten-second profiles, the old specialized loop had 1,020 self samples while the candidate's two shared generic-loop variants had 1,389 combined self samples, plus a separate `InputsCanHaveNull` pass.

I fixed that at the shared-engine boundary rather than putting representation loops back into `BinaryExecutor`. For a generic/constant pair, `ScalarExecutor` prepares only the generic operand, passes the constant by value, keeps the generic data pointer restricted, and loads one selection index per row. Separate nullable and no-NULL instantiations preserve NULL semantics without sending nullable comparisons through the full two-input generic loop. A small adapter hook still performs comparison-complement folding after the engine knows the inputs cannot contain NULLs, which removes the duplicate facade preflight.

The direct profiles are back at parity. Across six final alternating process pairs, dictionary/constant selection moved by +0.40%, constant/dictionary by +0.27%, and folded `>=` dictionary/constant by +0.16% using the median paired delta. A final Q85 gate with 16 alternating process pairs and 30 in-process samples per binary had a -0.02% median paired delta, a +0.07% two-pair-trimmed mean, and pooled medians of 14.274 ms versus 14.311 ms (+0.26%). Repeating the five-query advisory TPC subset in both binary orders produced +0.2% in base-to-fixed order and -0.0% in fixed-to-base order, with no regressions. The final Q85 sample puts the two no-NULL generic-constant loops at 840 combined self samples, down from 1,389 in the unfixed candidate, and the duplicate nullability-preflight frame is gone.

A later CI run exposed another cost in nullable generic execution. The first shared implementation mapped every input validity mask into result-row space with `ValidityMask::CopySel` before entering the operation loop. In a ten-second TPC-DS Q15 sample, that prepass took 217 self samples; the old binary executor's complete generic string loop took only 181 inclusive samples. The invariant belongs at the row being produced, so I removed the eager materialization. The shared loop now checks each input validity mask at its selected source index, after unified formats have been captured and result validity has been safely detached. A focused non-identity dictionary test covers result/input validity aliasing, stale-mask clearing, and reuse.

I compared the final source as an exact merge over CI base `76dd1e7d6f` against that same base, using Apple Clang 21, arm64 macOS 26.5, Release, the `tpch`, `tpcds`, `core_functions`, and `parquet` extensions, and two execution threads. TPC-H Q1 is 228 ms versus 229 ms (+0.6%, 50 samples), after removing the all-valid mask specialization reduced its earlier roughly 4% regression. Full TPC-H SF10 is +0.9% over 15 samples with no query regressions; the independent Linux CI result is +0.6%, with Q14 and Q15 slightly faster. Full TPC-DS SF3 is 27.5 ms versus 27.5 ms (+0.0%, 10 samples), also with no query regressions.

I also checked the Recursive CTE job because its CI distribution reported three 12-15% warnings. Native sampling found one separate refactor regression: variadic selection had lost compile-time output-presence dispatch, making the shared BETWEEN loop in `game_of_life` take 560 self samples versus 374 in the old variadic loop. Restoring that dispatch in `ScalarExecutor` brought `game_of_life` and `aoc2022_day11` back within 2%. The final 57-case suite is +0.4% by geometric mean with no query regressions. A consolidated 50-sample run of `direct_probe_sparse`, `asql_cyk_using_key`, and `aoc2022_day12_using_key` puts all three within 2% (+0.5% geometric mean); the bimodal Day 12 case alone is -0.5% over 100 samples.

Cheap ternary execution is where the consolidation pays off. The six mixed flat/constant profiles improved by 21% to 69%, while flat/flat/flat improved by about 2%. All-flat ternary selection improved from 7.060 s to 4.894 s, or 1.44x. Word-at-a-time validity handling also materially improved the NULL-heavy ternary cases: the 50% NULL profile went from 1.957 s to 0.076 s, the 99% NULL profile from 1.833 s to 0.076 s, and validity-word-sized clusters from 2.838 s to 0.355 s. Sampling and flamegraphs identified an initially non-inlined unary validity helper; after fixing that, the 50% NULL unary case was at parity (3.152 s versus 3.132 s).

The one visible microbenchmark trade-off is the all-constant binary case. It is roughly 3.3% slower because the new path clears stale constant validity before producing a valid result, which the pinned implementation omitted. This is one operation per vector rather than a row loop, and I chose the deterministic valid/NULL/valid behavior over preserving that small constant-dispatch difference.

The final specializations make an intentional binary-size trade-off. In the standard release configuration, local-symbol-stripped `libduckdb` grows from 45,252,216 to 46,132,680 bytes (+1.95%), and `__text` grows from 34,349,184 to 35,131,196 bytes (+2.28%). Most of that is the four nullable binary selection profiles instantiated for each operation: trimming only `binary_executor_generic_nullable` removes about 381 KiB from the stripped library, but direct nullable dictionary execution falls back from the 16-29% improvement to baseline performance. I kept the specialization because the runtime gain is substantial and the feature remains independently removable through `DUCKDB_SMALLER_BINARY`.

The fully smaller build strips all executor specializations and grows from 37,010,320 to 37,359,024 bytes (+0.94%), with `__text` growing from 26,560,980 to 26,867,080 bytes (+1.15%). That remaining fixed cost is the generic shared engine and the row-local validity handling. I did not duplicate the old executors behind preprocessor branches merely to hide it. Fully smaller builds retain the runtime-output fallback for variadic selection.

### DIFF
--- a/benchmark/micro/CMakeLists.txt
+++ b/benchmark/micro/CMakeLists.txt
@@ -5,6 +5,7 @@ add_library(
   append_mix.cpp
   bulkupdate.cpp
   cast.cpp
+  executor.cpp
   in.cpp
   storage.cpp
   string_serialize.cpp)
--- a/benchmark/micro/executor.cpp
+++ b/benchmark/micro/executor.cpp
@@ -18,6 +18,7 @@ enum class ExecutorBenchmarkType : uint8_t {
 	TERNARY_OPTIONAL_EXECUTE,
 	VARIADIC_EXECUTE,
 	BINARY_SELECT,
+	BINARY_SELECT_GREATER_THAN_EQUALS,
 	TERNARY_SELECT
 };
 
@@ -200,6 +201,15 @@ public:
 			}
 			break;
 		}
+		case ExecutorBenchmarkType::BINARY_SELECT_GREATER_THAN_EQUALS: {
+			auto true_sel = selection_output == ExecutorSelectionOutput::FALSE_ONLY ? nullptr : &state.true_sel;
+			auto false_sel = selection_output == ExecutorSelectionOutput::TRUE_ONLY ? nullptr : &state.false_sel;
+			for (idx_t i = 0; i < iterations; i++) {
+				state.selected_count = BinaryExecutor::Select<int64_t, int64_t, GreaterThanEquals>(
+				    state.a, state.b, nullptr, EXECUTOR_BENCHMARK_COUNT, true_sel, false_sel);
+			}
+			break;
+		}
 		case ExecutorBenchmarkType::TERNARY_SELECT:
 			for (idx_t i = 0; i < iterations; i++) {
 				state.selected_count = TernaryExecutor::Select<int64_t, int64_t, int64_t, ExecutorBetweenOperator>(
@@ -207,7 +217,9 @@ public:
 			}
 			break;
 		}
-		if (type == ExecutorBenchmarkType::BINARY_SELECT || type == ExecutorBenchmarkType::TERNARY_SELECT) {
+		if (type == ExecutorBenchmarkType::BINARY_SELECT ||
+		    type == ExecutorBenchmarkType::BINARY_SELECT_GREATER_THAN_EQUALS ||
+		    type == ExecutorBenchmarkType::TERNARY_SELECT) {
 			state.checksum = NumericCast<int64_t>(state.selected_count + 1);
 		} else {
 			auto value = state.result.GetValue(EXECUTOR_BENCHMARK_COUNT - 1);
@@ -235,6 +247,7 @@ private:
 			return constant_mask == 0x1 ? EXECUTOR_CONSTANT_BENCHMARK_ITERATIONS : EXECUTOR_BENCHMARK_ITERATIONS;
 		case ExecutorBenchmarkType::BINARY_EXECUTE:
 		case ExecutorBenchmarkType::BINARY_SELECT:
+		case ExecutorBenchmarkType::BINARY_SELECT_GREATER_THAN_EQUALS:
 		case ExecutorBenchmarkType::TERNARY_SELECT:
 			return constant_mask == 0x3 ? EXECUTOR_CONSTANT_BENCHMARK_ITERATIONS : EXECUTOR_BENCHMARK_ITERATIONS;
 		case ExecutorBenchmarkType::TERNARY_EXECUTE:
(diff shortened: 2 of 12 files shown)


Mytherin:
Thanks!

PR #24565, opened 2026-08-06
Recursive runtime invocation metrics

kryonix:
After #24031 was merged, I thought there could still be better evidence for the next recursive CTE optimization pass. End-to-end benchmark timings tell me whether a query improved, but not whether an epoch is dominated by scheduling, duplicate elimination, keyed-state maintenance, recurring-state access, or an invariant build. The existing recursive `PhysicalOperator` log had useful aggregate counts, but it could not answer those questions reliably for nested invocations or separate the major execution phases.

In short, this PR:

- gives every recursive CTE state a stable invocation identity in the existing `PhysicalOperator` log;
- reports bounded per-epoch distributions and frontier capacity metrics;
- attributes direct probes, keyed-state maintenance, fallback scans, final draining, duplicate elimination, active recursive pipelines, retained builds, and retained CTE materialization;
- distinguishes recursive scan work from direct key-probe work when selecting workers; and
- keeps the no-logging hot path free of metric clocks and atomic updates.

## Invocation and epoch evidence

Nested recursive CTEs can create and destroy several runtime states for the same physical operator. Some cached states also outlive the operator data that originally described them. I therefore snapshot the physical operator type and parameters when a logging-enabled state is created and assign an invocation ID that increases monotonically for each physical recursive operator. `RuntimeMetrics`, `EpochSummary`, and `DistinctPromoted` records can be joined by operator identity (for example, table index) plus invocation ID without retaining a reference to shorter-lived operator metadata.

`EpochSummary` uses fixed power-of-two histograms for frontier rows, workers, tasks, and elapsed time. It reports a p50 bucket upper bound and the exact maximum without retaining one value per epoch. The frontier section also reports storage and allocation byte-epochs plus peak capacity, which makes narrow-but-overallocated recursion visible. The terminal histogram bucket is covered directly for the full `idx_t` range; this caught and fixed an earlier decimal-digits-versus-binary-digits mistake.

The phase metrics are deliberately explicit about what they measure:

| Area | Evidence |
| --- | --- |
| Direct `USING KEY` probes | lookup, key-gather, and payload-finalization worker time; probe rows, matches, and partial-index chain visits |
| Keyed state | hash-table commit, partial-index maintenance, fallback recurring scans, and final-state drain worker time |
| `UNION DISTINCT` | grouping worker time plus candidates, insertions, and duplicate rejections; promotion migration is excluded from candidate totals |
| Recursive pipelines | active recursive, retained join/build, and retained materialized-CTE `PipelineExecutor::Execute` worker time |

Fields ending in `_work_ns` are cumulative worker time, not critical-path wall time, so parallel values can exceed the epoch's elapsed time. Operator phases such as direct probes and distinct grouping occur inside pipeline execution; those fields are nested attribution lenses, not values that should be added to the pipeline total. The pipeline metrics cover executor work and intentionally exclude prepare/finalize stages. Existing `RuntimeMetrics` totals remain useful for epoch counts, scheduler inputs, sink-lock wait/work, state rows, retained-build counts, and retained CTE reuse.

I also corrected recursive work discovery for `PhysicalRecursiveCTEKeyJoin`. A direct probe no longer masquerades as a full recurring-state scan. Recursive probe inputs are already represented by their visible recursive scans; independent probe inputs contribute estimated chunks to worker selection. This avoids both double-counting the frozen state and forcing an otherwise broad direct probe through the serial path.


Mytherin:
Thanks!

PR #24566, opened 2026-08-06
Add CIEndsWith and Identifier::EndsWith

JelteF:
To align with StartsWith. We have some uses for this at MotherDuck.


Mytherin:
Thanks!

PR #24567, opened 2026-08-06
Replace benchmark TPC-DS with SF=10, fix mold version check

smvv:
Replace benchmark TPC-DS SF=3 with SF=10 because some queries are very fast (some take only ~3-8 ms) in a benchmark.

Measuring these short queries accurately is difficult in CI runners and thus gives more noise in the benchmark report:
image

We're [iteratively](https://github.com/duckdb/duckdb/pull/24503) sampling more benchmark runs (unless outside the ±2% noise range), and that gives a lower CI job runtime for most CI runs. So, we can increase the scale factor for this benchmark and reduce its measurement noise.


benchmark/tpcds/sf10/q01.benchmark
```diff
@@ -0,0 +1,7 @@
+# name: benchmark/tpcds/sf10/q01.benchmark
```
Mytherin: `sf` is actually a parameter now in the `tpcds sf1` set of benchmarks - can we maybe use that parameter (set to 10) instead of including all of these new files? Having to create separate benchmark files to run separate scale factors is more of a relic of back when we didn't have these parameters.

Mytherin (commented):

Mytherin:
Thanks!

PR #24569, opened 2026-08-06
Cast MAP to a VARIANT OBJECT instead of an ARRAY

### LINKED ISSUE #24421: MAP → VARIANT → JSON does not match MAP → JSON
### What happens?

See repro for details.

### To Reproduce

```sql
memory D SELECT
           MAP {'a': 1, 'b': 2}::JSON AS map_json,
           MAP {'a': 1, 'b': 2}::VARIANT::JSON AS map_variant_json,
           variant_exists(MAP {'a': 1, 'b': 2}::VARIANT, 'a') AS exists_a,
           (MAP {'a': 1, 'b': 2}::VARIANT).a AS dot_a;
┌───────────────┬───────────────────────────────────────────────┬──────────┬─────────┐
│   map_json    │               map_variant_json                │ exists_a │  dot_a  │
│     json      │                     json                      │ boolean  │ variant │
├───────────────┼───────────────────────────────────────────────┼──────────┼─────────┤
│ {"a":1,"b":2} │ [{"key":"a","value":1},{"key":"b","value":2}] │ false    │ NULL    │
└───────────────┴───────────────────────────────────────────────┴──────────┴─────────┘
```
expected result
```sql
┌─────────────────┬──────────────────┬──────────┬────────┐
│    map_json     │ map_variant_json │ exists_a │ dot_a  │
│      json       │       json       │ boolean  │ variant│
├─────────────────┼──────────────────┼──────────┼────────┤
│ {"a":1,"b":2}   │ {"a":1,"b":2}    │ true     │ 1      │
└─────────────────┴──────────────────┴──────────┴────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

vvvdwbvvv:
## Summary

  Resolves #24421

  ### What happened

  A `MAP` cast to `VARIANT` does not round-trip to the same JSON as a `MAP` cast directly to
  `JSON`, and its keys are invisible to every VARIANT path-lookup function:

  ```sql
  SELECT
    MAP {'a': 1, 'b': 2}::JSON                          AS map_json,
    MAP {'a': 1, 'b': 2}::VARIANT::JSON                 AS map_variant_json,
    variant_exists(MAP {'a': 1, 'b': 2}::VARIANT, 'a')  AS exists_a,
    (MAP {'a': 1, 'b': 2}::VARIANT).a                   AS dot_a;
  ```

  A `MAP` is a key/value container, so it should reach VARIANT as an `OBJECT` — the same shape
  `MAP -> JSON` already produces.

  ### Before

  ```sql
┌───────────────┬───────────────────────────────────────────────┬──────────┬─────────┐
│   map_json    │               map_variant_json                │ exists_a │  dot_a  │
│     json      │                     json                      │ boolean  │ variant │
├───────────────┼───────────────────────────────────────────────┼──────────┼─────────┤
│ {"a":1,"b":2} │ [{"key":"a","value":1},{"key":"b","value":2}] │ false    │ NULL    │
└───────────────┴───────────────────────────────────────────────┴──────────┴─────────┘
  ```

  ### After

  ```sql
memory D SELECT
                    MAP {'a': 1, 'b': 2}::JSON AS map_json,
                    MAP {'a': 1, 'b': 2}::VARIANT::JSON AS map_variant_json,
                    variant_exists(MAP {'a': 1, 'b': 2}::VARIANT, 'a') AS exists_a,
                    (MAP {'a': 1, 'b': 2}::VARIANT).a AS dot_a;
┌───────────────┬──────────────────┬──────────┬─────────┐
│   map_json    │ map_variant_json │ exists_a │  dot_a  │
│     json      │       json       │ boolean  │ variant │
├───────────────┼──────────────────┼──────────┼─────────┤
│ {"a":1,"b":2} │ {"a":1,"b":2}    │ true     │ 1       │
└───────────────┴──────────────────┴──────────┴─────────┘
  ```

  ### Cause

`ConvertToVariant` in `to_variant.hpp` sends `MAP` and `LIST` to the same encoder. A `MAP` is
physically a `LIST(STRUCT(key, value))` so this works, but it writes `VariantLogicalType::ARRAY`
and never puts the keys in the VARIANT key dictionary. That's why `variant_exists`, `variant_type`,
`variant_keys`, `variant_array_length` and field access all find nothing.

The four `FIXME: map type is currently casted into a shape inconsistent with respect to JSON`
comments in the tests describe this, and two of them state the expected values.

### Fix

Give `MAP` its own `ConvertMapToVariant` that writes `OBJECT` and adds the keys to the dictionary.
Keys are cast to `VARCHAR` with `VectorOperations::DefaultCast`, the same way `MAP -> JSON` builds
its object keys. `ConvertListToVariant` is unchanged.

### Compatibility — please check this

I'm not sure this part is my call to make, so flagging it:

- This changes how `MAP` is written into VARIANT. A value written by this version would read back
  as an array-of-structs on an older DuckDB.
- Non-`VARCHAR` keys get stringified. `MAP {1: 'x'}::VARIANT::MAP(INT, VARCHAR)` still gives
  `{1=x}`, but only because the key gets re-parsed on the way out.
- `VARIANT -> MAP` now falls back to the per-value path in `from_variant.cpp` instead of the
  vectorized one. Still correct, but slower. I can add `MAP` to the `OBJECT` branch if you'd like
  that fixed here.

If any of this makes the approach wrong, happy to redo it.

### Tests covered

Added the field-access case from the issue to `variant_extract.test` — it was the only column of
the repro with no existing test.

The other six files change because the encoder change moves their output, so they have to be in
the same commit or CI is red in between:

- `variant_exists` / `variant_type` / `variant_keys` / `variant_array_length` — the four `FIXME`s
  removed, expectations set to what those comments asked for
- `test_all_types` — the `map` column
- `variant_map_to_variant_filter` — including the row past the 2048 vector boundary

`build/reldebug/test/unittest "[variant]"` is green. Also checked empty maps, `NULL` maps and
values, non-`VARCHAR` keys, and nested maps by hand.


Tishj:
We've thought about this, both of these shapes are not great solutions:
- `OBJECT(key ARRAY(<key_type>), value ARRAY(<value_type>))`
- `ARRAY(OBJECT(key <key_type>, value <value_type>))`

Our current internal idea that was floated is adding some `extension` type, of which `MAP` would be one. That can preserve the type information, so we can render the value as a MAP when requested.

That wouldn't change the internal representation, only serve to preserve the type information that would otherwise be lost.

vvvdwbvvv:
@Tishj Thanks,  that makes sense. Encoding MAP as OBJECT would lose the key type, while the current ARRAY representation loses the MAP semantic. An extension type/tag sounds like the right layer.
I won’t push the current approach further. I’d be happy to adapt the PR if the extension-type direction is ready to explore. Is there an existing design or issue I should follow?

PR #24571, opened 2026-08-06
fix: avoid int64 overflow in generic InterpolateOperator for fill()

### LINKED ISSUE #23986: `fill()` returns `BIGINT` minimum for an exact zero midpoint
### What happens?

`fill()` returns `-9223372036854775808` for the exact midpoint between the symmetric values `9223372036854775807` and `-9223372036854775807`. The mathematically correct midpoint is `0`.

### To Reproduce

```sql
-- the second row should be NULL, 0
memory D SELECT x, y, fill(y ORDER BY x) OVER () AS actual
         FROM (VALUES
             (0::BIGINT, 9223372036854775807::BIGINT),
             (1::BIGINT, NULL),
             (2::BIGINT, -9223372036854775807::BIGINT)
         ) t(x, y)
         ORDER BY x;
┌───────┬──────────────────────┬──────────────────────┐
│   x   │          y           │        actual        │
│ int64 │        int64         │        int64         │
├───────┼──────────────────────┼──────────────────────┤
│     0 │  9223372036854775807 │  9223372036854775807 │
│     1 │                 NULL │ -9223372036854775808 │
│     2 │ -9223372036854775807 │ -9223372036854775807 │
└───────┴──────────────────────┴──────────────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

ybapat:
This is made to fix #23986

When lo and hi are large int64 values with opposite signs, computing hi - lo in integer space overflows before the cast to double, producing a garbage delta and a wrong interpolated result (INT64_MIN instead of 0).

Fixed it by casting both operands to double before subtracting, matching the approach already used in the hugeint_t specialization. Also fixes the same overflow for int32 and smaller signed integer types, and for DECIMAL columns backed by int64 physical storage in quantile_cont. 

Also added a test.

### DIFF
--- a/src/include/duckdb/common/operator/interpolate.hpp
+++ b/src/include/duckdb/common/operator/interpolate.hpp
@@ -18,8 +18,8 @@ namespace duckdb {
 struct InterpolateOperator {
 	template <typename TARGET_TYPE>
 	static inline TARGET_TYPE Operation(const TARGET_TYPE &lo, const double d, const TARGET_TYPE &hi) {
-		const auto delta = static_cast<double>(hi - lo);
-		return LossyNumericCast<TARGET_TYPE>(lo + static_cast<TARGET_TYPE>(delta * d));
+		const auto delta = static_cast<double>(hi) - static_cast<double>(lo);
+		return LossyNumericCast<TARGET_TYPE>(static_cast<double>(lo) + delta * d);
 	}
 };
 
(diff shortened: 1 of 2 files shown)


hawkfish (approved):
Thanks for the PR LGTM!

Issue #24450, opened 2026-08-03
Vuln149: date_part with a non-foldable (column) part-name argument reaches InternalException when the runtime value names a DOUBLE-only part (epoch/julian)


ylwango613:
### What happens?

BindBinaryDatePart (extension/icu/icu-datepart.cpp) only substitutes the correct extractor when the part-name argument is foldable at bind time. When the argument is a column reference, binding falls through to BinaryTimestampFunction, which calls PartCodeBigintFactory on the runtime-resolved specifier without checking IsBigintDatepart. If that value is "epoch" or "julian", PartCodeBigintFactory hits its default branch and throws InternalException.

### PoC

```sql
LOAD icu;
CREATE TABLE parts(p VARCHAR);
INSERT INTO parts VALUES ('epoch');
SELECT date_part(p, TIMESTAMPTZ '2023-05-15 10:00:00+00') FROM parts;
```

### Result

Real captured output from the independent verification run:

```
INTERNAL Error: Unsupported ICU BIGINT extractor

Stack Trace:

build/codecov/duckdb(+0xa11b87) [0x5e7e48622b87]
build/codecov/duckdb(+0xa11ad4) [0x5e7e48622ad4]
build/codecov/duckdb(+0xa16fb8) [0x5e7e48627fb8]
... (37 more frames of internal binary addresses) ...
build/codecov/duckdb(+0x180455) [0x5e7e47d91455]

This error signals an assertion failure within DuckDB. This usually occurs due to unexpected conditions or errors in the program's logic.
For more information, see https://duckdb.org/docs/current/dev/internal_errors
Process exit code: 1
```

### OS:

Ubuntu 24.04.2 LTS, x86_64 (Linux 6.8.0-55-generic)

### DuckDB Version:

v1.6.0-dev10612, commit `2009475939c7b5ada2e55cc74843339bd9be79e3`

### DuckDB Client:

CLI (Command Line Interface), built from source with `make debug` (ASan + UBSan instrumented)

### Full Name:

Yuelin

### Affiliation:

Individual

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

hawkfish:
[...]

PR #24573, opened 2026-08-06
Issue #24450: DatePart Binary Doubles

hawkfish:
* Convert binary date_part signature to return DOUBLE.
* Always replace constant part binary functions with corresponding unary functions.
* Add reverse mapping from part codes to unary function names.
* Add tests for mixed return type part codes
* Update error messages.
* Make time part extraction for dates consistent.
* Fix case of bound function names.


extension/core_functions/scalar/date/date_part.cpp
```diff
@@ -1770,96 +1815,227 @@ int64_t ExtractElement(DatePartSpecifier type, T element) {
...
-			break;
-		default:
-			throw BinderException("%s can only take DATE or TIMESTAMP arguments", bound_function.GetName());
-		}
-		break;
-	case DatePartSpecifier::EPOCH:
-		arguments.erase(arguments.begin());
-		bound_function.GetArguments().erase(bound_function.GetArguments().begin());
-		bound_function.SetName("epoch");
+	arguments.erase(arguments.begin());
```
Maxxen, line 2026: So... we're trying to avoid arguments being erased during binding... but since we're basically replacing the function altogether here I guess it's fine.

Maybe we want to use the scalar functions "SetBindExpressionCallback/function_bind_expression_t" to replace/rebind the whole expression if we can?

Maxxen:
LGTM, I suppose since we only change the return type, and double is castable to bigint, this won't break any old queries.

hawkfish:
> LGTM, I suppose since we only change the return type, and double is castable to bigint, this won't break any old queries.

We only return `DOUBLE` now when the part code is variable (which is the bug). The rest of the time we will still return `BIGINT`. I just had to pry all of the constant part types out of the binary implementation so it could handle variable part codes.

PR #10187, opened 2024-01-10
Handle 0-list maps (erroring out) and add test-case

carlopi:
Fixes https://github.com/duckdb/duckdb/issues/10035

Together with @taniabogatsch

PR #24577, opened 2026-08-07
Internal #10187: Volatile Window Inputs

hawkfish:
* Don't apply the self-join optimisation when there are volatile inputs.

### DIFF
--- a/src/optimizer/window_self_join.cpp
+++ b/src/optimizer/window_self_join.cpp
@@ -13,6 +13,25 @@
 
 namespace duckdb {
 
+class VolatileExpressionCounter : public LogicalOperatorVisitor {
+public:
+	static idx_t HasVolatiles(LogicalOperator &op) {
+		VolatileExpressionCounter counter;
+		counter.VisitOperator(op);
+		return counter.volatiles;
+	}
+
+	VolatileExpressionCounter() {
+	}
+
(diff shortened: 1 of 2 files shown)


src/optimizer/window_self_join.cpp
```diff
@@ -180,6 +199,11 @@ unique_ptr<LogicalOperator> WindowSelfJoinOptimizer::OptimizeInternal(unique_ptr
 		// Check recursively
 		window.children[0] = OptimizeInternal(std::move(window.children[0]), replacer);
 
+		//	We cannot perform self-join when there are volatile functions below us.
+		if (VolatileExpressionCounter::HasVolatiles(window)) {
```
Mytherin, line 203: I wonder if the correct long-term pattern here is to rewrite the code to instead become a CTE (instead of copying the plan) which we can then refer to twice? The CTE de-correlation/optimization code should then in theory allow for this to be as efficient as the current optimizer.

Mytherin (commented):
Thanks for the PR! I'm fine with merging this as is - but one comment (cc @kryonix )

src/optimizer/window_self_join.cpp
kryonix: Yes, I think that's viable. I would not expect a performance issue, because of the pipeline exchange etc. And a CTE is the right tool to avoid volatile duplication.

PR #24578, opened 2026-08-07
Propagate column renames to trigger UPDATE OF lists

### LINKED ISSUE #24209: UPDATE OF Trigger Stale After Column Rename
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t(i INT, j INT);
memory D CREATE TABLE audit(x INT);
memory D CREATE TRIGGER tr AFTER UPDATE OF j ON t FOR EACH STATEMENT
           INSERT INTO audit VALUES (1);
memory D 
memory D INSERT INTO t VALUES (0,0);
memory D UPDATE t SET j=1;
memory D SELECT count(*) FROM audit;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
memory D 
memory D ALTER TABLE t RENAME COLUMN j TO k;
memory D UPDATE t SET k=2;

-- should be 2
memory D SELECT count(*) FROM audit;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

krleonid:
Fixes [duckdb/duckdb#24209](https://github.com/duckdb/duckdb/issues/24209).

## Problem

A trigger declared with `UPDATE OF <col>` kept the *old* column name after `ALTER TABLE ... RENAME COLUMN`. Because trigger dispatch drops triggers whose `OF` list is disjoint from the `SET` list, the trigger silently stopped firing — no error, just a dead trigger:

```sql
CREATE TABLE t(i INT, j INT);
CREATE TABLE audit(x INT);
CREATE TRIGGER tr AFTER UPDATE OF j ON t FOR EACH STATEMENT INSERT INTO audit VALUES (1);

INSERT INTO t VALUES (0,0);
UPDATE t SET j=1;
SELECT count(*) FROM audit;   -- 1

ALTER TABLE t RENAME COLUMN j TO k;
UPDATE t SET k=2;
SELECT count(*) FROM audit;   -- 1, should be 2
```

## Fix

`DuckTableEntry::RenameColumn` scans the table's trigger `CatalogSet` and alters any trigger whose `OF` list references the renamed column. Affected names are collected in a first pass, because the set cannot be mutated while it is being scanned.

`TriggerCatalogEntry` gains an `AlterEntry` override that rebuilds the entry with the new name substituted, returning `nullptr` when nothing matches so `CatalogSet::AlterEntry` treats it as a no-op.

Two consequences worth calling out:

- **Rollback** works through the existing undo buffer, since `triggers->AlterEntry` pushes its own catalog entry. Verified explicitly.
- **Trigger alters now reach `WALWriteState`**, which previously assumed triggers were only ever created. They are logged as a `CREATE` that replays as a replace, which is correct; the stale comment there is updated.

## Testing

`[trigger]` suite: 1682 assertions across 24 cases, all passing. Built and run with `FORCE_DEBUG=1 FORCE_ASSERT=1 make reldebug`, since the change makes a previously unreachable `D_ASSERT` in `wal_write_state.cpp` reachable (it holds — triggers are never renamed, so the entry type is never `RENAMED_ENTRY`).

Both new tests were confirmed to **fail without the fix** and pass with it.

`trigger_alter_table.test` covers, in-memory: the rename propagating, a non-`OF` column not firing, `ToSQL()` round-trip reflecting the new name, multi-column `OF` lists, several triggers on one table, and renaming an unrelated column leaving the trigger untouched.

`create_trigger_wal.test` covers persistence via the file's existing ATTACH → DETACH → re-ATTACH pattern with checkpointing disabled, so the reload genuinely replays the WAL. It asserts the renamed `OF` column survives replay and that the trigger still *fires* afterwards, plus the negative case so it cannot pass by making every trigger fire.

Also verified by hand: `CHECKPOINT` persistence, case-insensitive matching (`Identifier` compares via `StringUtil::CIEquals`), and `ROLLBACK` reverting the trigger.


test/sql/trigger/create_trigger_wal.test
```diff
@@ -60,6 +60,67 @@ SELECT trigger_name, for_each, columns FROM duckdb_triggers() WHERE trigger_name
 ----
 upd_trigger	STATEMENT	[id]
 
+# ALTER TABLE RENAME COLUMN must propagate to the UPDATE OF list, and survive WAL replay
+statement ok
+USE triggerdb;
+
+statement ok
+CREATE TABLE t2 (i INTEGER, j INTEGER);
+
```
lnkuiper, line 124: Can we add more multi-column trigger tests here?

src/transaction/wal_write_state.cpp
```diff
@@ -49,7 +49,8 @@ void WALWriteState::WriteCatalogEntry(CatalogEntry &entry, data_ptr_t dataptr) {
 
 	switch (parent.type) {
 	case CatalogType::TRIGGER_ENTRY:
-		// Triggers do not support ALTER — always a CREATE
+		// Triggers are never renamed, so this is either a CREATE or a column-rename propagated from
+		// the base table. Both are logged as a CREATE, which replays as a replace.
```
lnkuiper: I don't think this propagated alteration should be written as a `CREATE_TRIGGER`. The owning `ALTER TABLE` already writes an `ALTER_INFO` record, and replaying that record calls `RenameColumn` and reconstructs the trigger change. The trigger catalog version is needed for MVCC and rollback, but it is not an independent durable operation. Could we skip these trigger-to-trigger alterations in the WAL writer while continuing to log genuine trigger creation?

src/catalog/catalog_entry/duck_table_entry.cpp
```diff
@@ -484,6 +484,23 @@ unique_ptr<CatalogEntry> DuckTableEntry::RenameColumn(ClientContext &context, Re
 	auto binder = Binder::CreateBinder(context);
 	auto bound_create_info = binder->BindCreateTableInfo(std::move(create_info), schema, info.bind_mode);
 	SetAlterDependencies(*bound_create_info, info);
+
+	// Update any UPDATE OF triggers whose column list references the renamed column
+	auto txn = catalog.GetCatalogTransaction(context);
+	vector<Identifier> triggers_to_update;
+	triggers->Scan(txn, [&](CatalogEntry &raw_entry) {
```
lnkuiper: This snapshot scan can miss a trigger created by a concurrent transaction. The trigger and table live in independently versioned catalog sets, so the following sequence allows both transactions to commit:

```sql
statement ok con1
CREATE TABLE t(i INTEGER, j INTEGER);

statement ok con1
CREATE TABLE audit(x INTEGER);

# This transaction still sees t(i, j).
statement ok con2
BEGIN;

statement ok con2
CREATE TRIGGER tr AFTER UPDATE OF j ON t
INSERT INTO audit VALUES (1);

# The trigger is uncommitted, so this scan cannot see it.
statement ok con1
ALTER TABLE t RENAME COLUMN j TO k;

statement ok con2
COMMIT;

query T con1
SELECT columns::VARCHAR
FROM duckdb_triggers()
WHERE trigger_name = 'tr';
----
[k]

statement ok con1
UPDATE t SET k = 42;
query I con1
SELECT count(*) FROM audit;
----
1
```

lnkuiper (changes requested):
Thanks! I've left a few comments:

test/sql/trigger/create_trigger_wal.test
krleonid: Done — added a `t3(a, b, c)` table with a `AFTER UPDATE OF b, c` trigger and three sequential renames:
- `b → b2` → columns become `[b2, c]`
- `c → c2` → columns become `[b2, c2]`
- `a → a2` (not in trigger) → columns unchanged `[b2, c2]`

Followed by a detach/re-attach cycle to verify the multi-column list also survives WAL replay.

src/transaction/wal_write_state.cpp
krleonid: Good point — fixed. When `entry.type == parent.type` (both `TRIGGER_ENTRY`) the entry is a column-rename propagation, not a genuine `CREATE TRIGGER`. We now `return` early in that branch instead of calling `WriteCreateTrigger`. The owning table's `ALTER_INFO` record already reproduces the trigger update on replay via `RenameColumn → triggers->AlterEntry`.

src/catalog/catalog_entry/duck_table_entry.cpp
krleonid: Fixed by adding `CatalogSet::ScanWithConflictDetection`, a variant of `Scan` that also invokes a conflict callback for every head entry where `HasConflict(txn, entry.timestamp)` is true (i.e. created by another active transaction, or committed after our start time).

`RenameColumn` now uses this instead of plain `Scan`. If the conflict callback finds a `TRIGGER_ENTRY` that isn't deleted and whose column list contains the renamed column, it throws a `TransactionException` (write-write conflict). The rename then fails, and the caller can retry once the concurrent transaction has committed — at which point the retry's snapshot scan will see the trigger and correctly update its column list.

src/catalog/catalog_entry/duck_table_entry.cpp
```diff
@@ -484,6 +484,41 @@ unique_ptr<CatalogEntry> DuckTableEntry::RenameColumn(ClientContext &context, Re
...
 	auto bound_create_info = binder->BindCreateTableInfo(std::move(create_info), schema, info.bind_mode);
 	SetAlterDependencies(*bound_create_info, info);
+
+	// Update any UPDATE OF triggers whose column list references the renamed column.
+	// Also detect concurrent uncommitted (or recently-committed) triggers that reference the same
+	// column: the snapshot scan cannot see them, so we raise a write-write conflict so the caller
+	// retries after the concurrent transaction completes.
+	auto txn = catalog.GetCatalogTransaction(context);
+	vector<Identifier> triggers_to_update;
+	triggers->ScanWithConflictDetection(
```
lnkuiper, line 494: There's still a concurrency issue here:
```sql
statement ok con1
CREATE TABLE t(i INTEGER, j INTEGER);

statement ok con1
CREATE TABLE audit(x INTEGER);

statement ok con1
BEGIN;

# The trigger scan has completed.
statement ok con1
ALTER TABLE t RENAME COLUMN j TO k;

# This connection still sees t(i, j) and inserts after the scan.
statement ok con2
CREATE TRIGGER tr AFTER UPDATE OF j ON t
INSERT INTO audit VALUES (1);

statement ok con1
COMMIT;

query T con1
SELECT columns::VARCHAR
FROM duckdb_triggers()
WHERE trigger_name = 'tr';
----
[j]

statement ok con1
UPDATE t SET k = 42;

query I con1
SELECT count(*) FROM audit;
----
0
```

Could `CREATE TRIGGER` revalidate at commit that the owning table version it bound against is still current, and raise a catalog conflict if the table was altered concurrently? The rename-side scan already handles the opposite ordering.

lnkuiper (commented):
Thanks! One last issue remains:

src/catalog/catalog_entry/duck_table_entry.cpp
krleonid: Done — `CREATE TRIGGER` now revalidates its UPDATE OF columns at commit time against the current committed table schema (Case A: rename committed before trigger commits) and also checks the HEAD entry for an uncommitted pending rename (Case B: trigger commits first). Both orderings raise a `TransactionException` with a write-write conflict message. Tests covering both scenarios are in `test/sql/trigger/trigger_concurrent_rename.test`.

src/transaction/commit_state.cpp
```diff
@@ -298,6 +300,51 @@ void CommitState::CommitEntry(UndoFlags type, data_ptr_t data, CommitInfo &info)
...
+				// Build a transaction view that sees all committed changes up to this commit
+				CatalogTransaction commit_txn(duck_catalog.GetDatabase(), MAX_TRANSACTION_ID, commit_id + 1);
+				auto &table_set = trig.schema.Cast<DuckSchemaEntry>().GetCatalogSet(CatalogType::TABLE_ENTRY);
+
+				// Case (A): check the committed table schema
+				auto table_entry = trig.schema.GetEntry(commit_txn, CatalogType::TABLE_ENTRY, trig.base_table->Table());
+				if (table_entry) {
+					auto &table = table_entry->Cast<TableCatalogEntry>();
+					for (const auto &col : trig.columns) {
+						if (!table.ColumnExists(col)) {
```
lnkuiper: Checking that a column exists in both versions is not enough to validate the table version:
```sql
  require noforcestorage

  statement ok
  CREATE TABLE t(i INTEGER, j INTEGER);

  statement ok
  CREATE TABLE audit(x INTEGER);

  statement ok con1
  BEGIN;

  statement ok con1
  ALTER TABLE t RENAME COLUMN j TO k;

  statement ok con1
  ALTER TABLE t ADD COLUMN j INTEGER;

  # con2 still binds against the original j, but the pending table also has
  # a different column named j. This currently succeeds.
  statement error con2
  CREATE TRIGGER tr AFTER UPDATE OF j ON t
  INSERT INTO audit VALUES (1);
  ----
  <REGEX>:.*write-write conflict.*

  statement ok con1
  ROLLBACK;
```
Could we conflict whenever the owning table version differs from the one used to bind the trigger? Comparing the table OID is insufficient because `ALTER` preserves it.

lnkuiper (commented):
Thanks for the changes! I have another comment:

src/transaction/commit_state.cpp
krleonid: Fixed — switched from `ColumnExists` to a table-entry pointer comparison. Any structural change to the table (rename, add, drop) produces a new `CatalogEntry` object, so if the pointer at bind time (`transaction.start_time`) differs from the pointer at commit time, we throw a write-write conflict. Case B (uncommitted alter in flight) now also conflicts unconditionally without inspecting column names. Added Scenario C to `trigger_concurrent_rename.test` covering the rename+re-add case from your example.

src/transaction/commit_state.cpp
```diff
@@ -298,6 +300,48 @@ void CommitState::CommitEntry(UndoFlags type, data_ptr_t data, CommitInfo &info)
...
+		// Instead we compare table entry pointers: any structural change to the table produces a
+		// new CatalogEntry object.  Two orderings need to be handled:
+		//   (A) rename commits before this trigger commits  → bound and current entries differ
+		//   (B) this trigger commits before rename commits  → head entry has a write-write conflict
+		// old_entry.type is TRIGGER_ENTRY only for column-rename propagations (ALTER), which we skip.
+		if (new_entry.type == CatalogType::TRIGGER_ENTRY && old_entry.type != CatalogType::TRIGGER_ENTRY) {
+			auto &trig = new_entry.Cast<TriggerCatalogEntry>();
+			if (!trig.columns.empty()) {
+				auto &table_set = trig.schema.Cast<DuckSchemaEntry>().GetCatalogSet(CatalogType::TABLE_ENTRY);
+				// Transaction view at bind time (what the trigger saw when it was created)
```
lnkuiper: This does not necessarily recover the table version the trigger was bound against. This fails:

```sql
  require noforcestorage

  statement ok
  CREATE TABLE t(i INTEGER, j INTEGER);

  statement ok
  INSERT INTO t VALUES (1, 2);

  statement ok
  CREATE TABLE audit(x INTEGER);

  statement ok
  BEGIN;

  statement ok
  ALTER TABLE t RENAME COLUMN j TO k;

  statement ok
  CREATE TRIGGER tr AFTER UPDATE OF k ON t
  INSERT INTO audit VALUES (1);

  statement ok
  COMMIT;

  statement ok
  UPDATE t SET k = 3;

  query I
  SELECT * FROM audit;
  ----
  1
```

I think we can fix it with:
```diff
- // Transaction view at bind time (what the trigger saw when it was created)
- CatalogTransaction bind_txn(duck_catalog.GetDatabase(), MAX_TRANSACTION_ID, transaction.start_time);
+ // Earlier changes by this transaction already carry commit_id.
+ CatalogTransaction bind_txn(duck_catalog.GetDatabase(), commit_id, transaction.start_time);
```

lnkuiper (commented):
Thanks! Just one more comment then it's ready to go:

src/transaction/commit_state.cpp
krleonid: Fixed with the one-liner you suggested — changed `MAX_TRANSACTION_ID` to `commit_id` in `bind_txn`. Since earlier catalog changes in the same transaction are already stamped with `commit_id` by the time we reach the trigger entry, using `commit_id` as `transaction_id` makes them visible via the `timestamp == transaction_id` branch of `UseTimestamp`. Added Scenario D to `trigger_concurrent_rename.test` covering the same-transaction RENAME + CREATE TRIGGER case.

lnkuiper:
An issue popped up after merging current `main`.

`SetAlterDependencies` stores the table's replacement dependencies in `info.new_dependencies`, but this passes the same `RenameColumnInfo` into the trigger alteration. `DependencyManager::AlterObject` therefore cleans the trigger's dependency edges and recreates them from the table's dependency list. After the rename, `DROP TABLE log RESTRICT` succeeds even though the trigger body still references `log`.

Could we pass a copy without `new_dependencies` when altering the triggers, so their existing dependencies are retained? For example:

```cpp
auto trigger_alter_info = info.Copy();
for (const auto &trigger_name : triggers_to_update) {
	triggers->AlterEntry(txn, trigger_name, *trigger_alter_info);
}
```

Repro:

```sql
require noforcestorage

statement ok
CREATE TABLE t(i INTEGER, j INTEGER);

statement ok
CREATE TABLE log(x INTEGER);

statement ok
CREATE TRIGGER tr AFTER UPDATE OF j ON t INSERT INTO log VALUES (1);

statement ok
ALTER TABLE t RENAME COLUMN j TO k;

statement error
DROP TABLE log RESTRICT;
----
<REGEX>:.*trigger "tr" on table "t" depends on table "log".*
```

krleonid:
@lnkuiper good catch. I've fixed it

lnkuiper (approved):
Thanks! LGTM now

lnkuiper:
The failing CI seems related - might need a skip. Could you have a look?

PR #24160, opened 2026-07-25
Remove redundant LIMIT using maximum cardinality

naseweisssss:
Commit to trigger action, added check to remove limit if cardinality is less than needed. Test added.

PR #24579, opened 2026-08-07
Optimize redundant limits using maximum cardinality

### LINKED ISSUE #24072: Nested redundant LIMIT not removed
### What happens?

See repro for details.

### To Reproduce

```sql
-- limit 20 should be removed
memory D EXPLAIN ANALYZE
         SELECT *
         FROM (SELECT * FROM range(10) LIMIT 20) q
         LIMIT 30;
╭─ Summary ───────────╮
│ Total Time: 0.0003s │
╰─────────────────────╯
╭─ Streaming Limit ─╮
│ Limit: 30         │
│ 10 rows       0µs │
╰─────────┒┎────────╯
╭─ Streaming Limit ─╮
│ Limit: 20         │
│ 10 rows       0µs │
╰─────────┒┎────────╯
╭─ Table Scan ──────╮
│ Function: RANGE   │
│ 10 rows       0µs │
╰───────────────────╯
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

yohaku999:
Resolve #24072.

When a `LogicalLimit` has a constant limit value, no effective offset, and a child whose maximum cardinality is less than or equal to that value, the operator cannot alter the result. DuckDB nevertheless retains the redundant operator, which is subsequently converted into a physical limit operator.

**Fix**

- Handle `LogicalLimit` during statistics propagation and use the child’s maximum cardinality to identify redundant limit operators.
- Remove a `LogicalLimit` when its limit value is constant, its offset is unset or zero, and the child’s maximum cardinality is less than or equal to the limit value.
- When a `LogicalLimit` remains, use the constant limit as its maximum cardinality, capped by the child’s maximum cardinality when available.

**Testing**

Added regression tests covering nested limits over `range`, regular table scans, propagation through filters, and nested limit elimination. The new optimization also affected existing `LimitPushdown` and `TopN` tests, so their inputs and limit values were adjusted to preserve the tests’ original intent.


yohaku999:
#24160 has since been merged and appears to resolve #24072, so I’m closing this PR as superseded.

PR #24582, opened 2026-08-07
Fix heap overflow in `string_agg`

### LINKED ISSUE #24188: Bug1: Integer Truncation in string_agg Leads to Heap Buffer Overflow
### What happens?

In `extension/core_functions/aggregate/distributive/string_agg.cpp`, `StringAggState::alloc_size` is declared as `uint32_t` while `PerformOperation()` computes the new accumulated size as a 64 bit `idx_t` and only rejects it once it exceeds `NumericLimits<uint32_t>::Maximum()`. For any accumulated size in the range `(2^31, 2^32-1]`, `NextPowerOfTwo(new_size)` returns exactly `2^32`, which silently truncates to `0` when stored into `alloc_size`. `allocator.Allocate(0)` then hands back an effectively zero sized region, and the following `memcpy` writes the full previously accumulated string, up to a few gigabytes, into that region, corrupting adjacent arena memory. A single `SELECT string_agg(...)` query over attacker sized string data reaches this path with no file, ATTACH, or special privilege required. The same bug was independently rediscovered while auditing a second file that includes this header, confirming it from two audit passes.

### To Reproduce

SQL script (`vuln_001.sql`):

```sql
-- vuln_001: force a single string_agg group to accumulate into the
-- (2^31, 2^32-1] byte window so that NextPowerOfTwo(new_size) == 2^32,
-- which truncates StringAggState::alloc_size (uint32_t) to 0.
SET threads=1;
SELECT length(string_agg(s, '')) AS result
FROM (SELECT repeat('A', 100000000) AS s FROM range(30)) t;
```

Trigger output (ASan, `build/debug/duckdb`):

```
AddressSanitizer:DEADLYSIGNAL
...
SUMMARY: AddressSanitizer: SEGV ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:837 in __memmove_avx512_unaligned_erms
==2566060==ABORTING
```

The crash is a real SIGSEGV reached through `PendingQueryResult::Execute` -> `ClientContext::Query` -> `StringAggFunction::PerformOperation`, confirming the heap corruption is real and not merely caught by an assertion.

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

sebastiaan-dev:
Resolves https://github.com/duckdb/duckdb/issues/24188 by clamping the maximum allocation size.

### DIFF
--- a/extension/core_functions/aggregate/distributive/string_agg.cpp
+++ b/extension/core_functions/aggregate/distributive/string_agg.cpp
@@ -64,7 +64,7 @@ struct StringAggFunction {
 			if (new_size > NumericLimits<uint32_t>::Maximum()) {
 				throw InvalidInputException("string_agg string size exceeds maximum string size");
 			}
-			state.alloc_size = NextPowerOfTwo(new_size);
+			state.alloc_size = MinValue<idx_t>(NextPowerOfTwo(new_size), string_t::MAX_STRING_SIZE);
 			target_data = char_ptr_cast(allocator.Allocate(state.alloc_size));
 			if (current_size > 0) {
 				// copy over the current data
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24584, opened 2026-08-07
Parallelize merge of small collections in PhysicalInsert

NiclasHaderer:
In the parallel insert, rows that are smaller than one row group are re applied to local storage in the `Combine` while holding the global lock. If you have more threads this slows down substantially.

This PR borrows the concept from the `PhysicalBatchInsert` by deferring the leftover collections to Finalize where they are packed into row-group-sized sets and compacted in parallel. This reduces the time spent in the PhysicalInsert from ∼24-25 to ∼11-13 seconds in a customer query while still being a very generic fix which should improve the query speed in general.


src/execution/operator/persistent/physical_insert.cpp
```diff
@@ -721,8 +717,90 @@ SinkCombineResultType PhysicalInsert::Combine(ExecutionContext &context, Operato
...
+			current_rows = 0;
+		}
+	}
+	if (current_merger) {
+		mergers.push_back(std::move(current_merger));
+	}
+	gstate.unmerged_collections.clear();
+
+	// compact the merge sets in parallel
+	TaskExecutor executor(context);
```
Mytherin: We should never be using a `TaskExecutor` inside a `Finalize` method, we should be launching a new event similar to how other finalize-parallelization tasks work.

Mytherin (commented):
Thanks - can we retarget this to `main`? This is too risky to go into `v1.5-variegata`. Also left a comment

src/execution/operator/persistent/physical_insert.cpp
NiclasHaderer: Thanks makes sense.

NiclasHaderer:
I am going to retarget to main once you think it is OK, since our patch is going to to out for 1.5.5 if that works for you

Mytherin (approved):
Thanks! Looks good to me

Mytherin:
Could you just have a look at fixing up the CI?

NiclasHaderer:
I'm currently porting the PR main like you asked and will fix CI as well

src/execution/operator/persistent/physical_insert.cpp
```diff
@@ -709,8 +706,121 @@ SinkCombineResultType PhysicalInsert::Combine(ExecutionContext &context, Operato
...
                                           OperatorSinkFinalizeInput &input) const {
+	auto &gstate = input.global_state.Cast<InsertGlobalState>();
+	if (gstate.unmerged_collections.empty()) {
+		return SinkFinalizeType::READY;
+	}
+	auto &table = gstate.table;
+	auto &data_table = table.GetStorage();
+	const idx_t row_group_size = data_table.GetRowGroupSize();
+
+	if (gstate.insert_count < row_group_size) {
```
NiclasHaderer, line 784: For this heuristic, would it make more sense to do a `(sum(unmerged_collectionsitems.size()) < row_group_size)`

src/execution/operator/persistent/physical_insert.cpp
```diff
@@ -709,8 +706,121 @@ SinkCombineResultType PhysicalInsert::Combine(ExecutionContext &context, Operato
...
+		data_table.FinalizeLocalAppend(append_state);
+		gstate.unmerged_collections.clear();
+		return SinkFinalizeType::READY;
+	}
+
+	// group the leftover thread-local collections into row-group-sized merge sets
+	vector<unique_ptr<CollectionMerger>> mergers;
+	unique_ptr<CollectionMerger> current_merger;
+	idx_t current_rows = 0;
+	for (auto &collection_index : gstate.unmerged_collections) {
```
NiclasHaderer, line 805: Should we create a merger per thread that we can use instead, or is the current approach better?

NiclasHaderer:
I did not change `enum_util.cpp`, so I think something with CI is wrong

Mytherin:
Thanks!

PR #24585, opened 2026-08-07
Fix external file cache seek when uncache read

dentiny:
The bug is
- When serving uncached read, currently internal file handle directly reads from the current position, whatever the requested position is
- It doesn't reproduces usually because if we're (1) always doing cached read or uncached read; and (2) cached content serves requests properly well

In this PR, I seek to the requested position for internal file handle; don't think it matters for non-seekable file handle (i.e., pipe-based).

### DIFF
--- a/src/storage/external_file_cache/caching_file_system.cpp
+++ b/src/storage/external_file_cache/caching_file_system.cpp
@@ -364,10 +364,14 @@ FileBufferHandleGroup CachingFileHandle::Read(idx_t &nr_bytes) {
 	if (!external_file_cache.IsEnabled() || !CanSeek() || no_validation_metadata) {
 		auto buf = AllocateUncachedReadBuffer(external_file_cache.GetBufferManager(), nr_bytes);
 		auto file_handle = GetFileHandle();
+		// Do positional read if handle can seek, otherwise have to fallback to sequential read.
 		if (file_handle->CanSeek()) {
-			file_handle->Seek(position);
+			const auto file_size = file_handle->GetFileSize();
+			nr_bytes = position >= file_size ? 0 : MinValue(nr_bytes, file_size - position);
+			file_handle->Read(context, buf.GetDataMutable(), nr_bytes, position);
+		} else {
+			nr_bytes = NumericCast<idx_t>(file_handle->Read(context, buf.GetDataMutable(), nr_bytes));
 		}
-		nr_bytes = NumericCast<idx_t>(file_handle->Read(context, buf.GetDataMutable(), nr_bytes));
 		vector<FileBufferHandleGroup::MemoryHandle> mem_handles;
 		mem_handles.push_back({std::move(buf), 0, nr_bytes});
 		position += nr_bytes;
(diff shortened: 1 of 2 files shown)


src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -364,6 +364,9 @@ FileBufferHandleGroup CachingFileHandle::Read(idx_t &nr_bytes) {
 	if (!external_file_cache.IsEnabled() || !CanSeek() || no_validation_metadata) {
 		auto buf = AllocateUncachedReadBuffer(external_file_cache.GetBufferManager(), nr_bytes);
 		auto file_handle = GetFileHandle();
+		if (file_handle->CanSeek()) {
+			file_handle->Seek(position);
+		}
 		nr_bytes = NumericCast<idx_t>(file_handle->Read(context, buf.GetDataMutable(), nr_bytes));
```
Mytherin: Can't we just do a positional read here instead of adding a seek? Seek + read is less efficient and a potential race condition

Mytherin (commented):
Thanks for the PR! One comment

src/storage/external_file_cache/caching_file_system.cpp
dentiny: I considered that, but non-seekable file handle usually doesn't implement `pread` / positional read semantics. One concrete example is the well-known shellfs: https://github.com/Query-farm/shellfs/blob/main/src/shell_file_system.hpp

dentiny: Yeah I reconsidered, updated logic to; attempt position read if seekable, otherwise fallback to sequential read.

src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -364,7 +364,14 @@ FileBufferHandleGroup CachingFileHandle::Read(idx_t &nr_bytes) {
 	if (!external_file_cache.IsEnabled() || !CanSeek() || no_validation_metadata) {
 		auto buf = AllocateUncachedReadBuffer(external_file_cache.GetBufferManager(), nr_bytes);
 		auto file_handle = GetFileHandle();
-		nr_bytes = NumericCast<idx_t>(file_handle->Read(context, buf.GetDataMutable(), nr_bytes));
+		// Do positional read if handle can seek, otherwise have to fallback to sequential read.
+		if (file_handle->CanSeek()) {
+			const auto file_size = file_handle->GetFileSize();
```
Mytherin, line 369: `Read` returns the `nr_bytes` to read - why do this with the file size here?

Mytherin (commented):
Thanks - one more comment

src/storage/external_file_cache/caching_file_system.cpp
Mytherin: Ah nevermind, pread throws if we can't read all of the bytes

Mytherin:
Thanks!

Issue #23588, opened 2026-07-03
top_n_window_elimination can collapse duplicate payload rows


Yibo-Dong:
### What happens?

top_n_window_elimination can collapse duplicate payload rows produced by a join when the row_number column and ordering key are not projected.

This reproduces on current main (ac62d7c180), and also on v1.5.4.

### To Reproduce

To reproduce:

```sql
CREATE OR REPLACE TABLE l_topn_dup(id INTEGER, k INTEGER, p VARCHAR);

CREATE OR REPLACE TABLE r_topn_dup(k INTEGER, s INTEGER);

INSERT INTO l_topn_dup VALUES (1, 1, 'x');

INSERT INTO r_topn_dup VALUES (1, 20), (1, 10);

SELECT id, p
FROM (
  SELECT l_topn_dup.id, l_topn_dup.p, r_topn_dup.s
  FROM l_topn_dup
  JOIN r_topn_dup ON l_topn_dup.k = r_topn_dup.k
) AS src
QUALIFY ROW_NUMBER() OVER (ORDER BY s DESC NULLS LAST) <= 2
ORDER BY ALL;
```

Run:

```sh
./build/release/duckdb -json -f bug.sql
```

Actual:

On current main (`ac62d7c180`), running the minimized repro gives:

stdout:

| id | p |
| ---: | --- |
| 1 | x |

Expected:

| id | p |
| ---: | --- |
| 1 | x |
| 1 | x |

The joined source has two rows with different ordering keys:

| id | p | s |
| ---: | --- | ---: |
| 1 | x | 20 |
| 1 | x | 10 |

Both rows satisfy `ROW_NUMBER() <= 2`. The final projection drops `s`, but SQL
bag semantics should still preserve both identical `(id, p)` rows.

Notes From Minimization:
* Turning off `top_n_window_elimination` makes the query return the expected two rows:

---

Possible Cause (generated by coding agent):
Source pointer: `src/optimizer/topn_window_elimination.cpp`.

This is in `TopNWindowElimination`'s late-materialization path, not in the regular window executor and not in the `arg_max(..., n)` aggregate itself.

On current main, the optimized plan for this query uses:

```text
arg_max(rowid, s, 2)
...
SEMI JOIN ON rowid = rowid
```

[...]

PR #24586, opened 2026-08-07
Fix inner join deduplication in TopNWindowElimination

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/9867
fixes https://github.com/duckdb/duckdb/issues/23588
supersedes PR https://github.com/duckdb/duckdb/pull/23702

This PR fixes a bug that would collapse duplicate row ids in the late materialization step. This issue is resolved by reconstructing the duplicates with an inner join during tuple reconstruction. Moreover, we now conservatively only remove unused columns if the child operators do not contain any projection maps as that could drop actually used bindings here.

### DIFF
--- a/src/optimizer/topn_window_elimination.cpp
+++ b/src/optimizer/topn_window_elimination.cpp
@@ -1214,28 +1214,26 @@ unique_ptr<LogicalOperator> TopNWindowElimination::ConstructJoin(unique_ptr<Logi
 
 	const auto rhs_bindings = rhs->GetColumnBindings();
 	const bool use_inner_reconstruction = params.row_ids_may_have_duplicates || params.include_row_number;
-	idx_t rhs_table_idx = DConstants::INVALID_INDEX;
-	if (use_inner_reconstruction) {
-		// The group columns on the aggregate side are only needed to compute the TopN result. They are not required for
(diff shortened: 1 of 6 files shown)


src/optimizer/topn_window_elimination.cpp
```diff
@@ -952,6 +964,11 @@ bool TopNWindowElimination::CanUseLateMaterialization(const LogicalWindow &windo
 			    join.join_type != JoinType::ANTI) {
 				return false;
 			}
+			if (join.join_type == JoinType::INNER) {
```
hawkfish, line 961: Can't a `LEFT` join do this too?

hawkfish (commented):
Looks good - just one question.

src/optimizer/topn_window_elimination.cpp
d-justen: In `CanUseLateMaterialization`, we only accept sub-plans with inner, anti, and semi join.

hawkfish: Ah sorry - I looked right at that and didn't take it in for some reason...

hawkfish (approved):
LGTM!

lnkuiper:
Could you resolve the merge conflicts? Then this should be good to go

PR #24587, opened 2026-08-07
Implement HTTP cache control

### LINKED ISSUE #24474: Regression in DuckDB 1.5.5: HTTPFS cache no longer reused for repeated read_parquet() requests
### What happens?

It looks like there is a regression between DuckDB 1.5.4 and 1.5.5 regarding the HTTPFS cache.

With DuckDB 1.5.4, reading the same remote Parquet file twice correctly reuses the cache. With DuckDB 1.5.5, the second identical query downloads the data again instead of using the cached content.

## Reproduction

Using the following query:

```sql
EXPLAIN ANALYZE
FROM read_parquet('https://api.insee.fr/melodi/file/DS_PRENOM_2025/PARQUET')
WHERE first_name = 'ERIC_1';
```

### DuckDB 1.5.4

First execution:

- `HTTPFS HTTP Stats IN: 6.9 MB`

Second execution (same query):

- `HTTPFS HTTP Stats IN: 0 MB`

This indicates that the HTTPFS cache is working as expected.

### DuckDB 1.5.5

First execution:

- `HTTPFS HTTP Stats IN: 6.9 MB`

Second execution (same query):

- `HTTPFS HTTP Stats IN: 6.9 MB`

The remote data is downloaded again, suggesting that the HTTPFS cache is no longer being reused.

## Expected behavior

The second execution should reuse the cached data, resulting in `HTTPFS HTTP Stats IN: 0 MB`, as in DuckDB 1.5.4.

## Actual behavior

The second execution downloads the same 6.9 MB again, indicating that the HTTPFS cache is not being used.

## Environment

- DuckDB 1.5.4: cache works as expected.
- DuckDB 1.5.5: cache no longer works (appears to be a regression).

### OS:

Win 11

### DuckDB Version:

1.5.5

### DuckDB Client:

Java

### Hardware:

_No response_

### Full Name:

eric mauviere

### Affiliation:

icem7

- [x] Yes, I have

- [x] Yes, I have

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24474

This PR implements HTTP cache control policy.
As mentioned in the issue, previously we disabled the cache when it's impossible to tell version difference because HTTP response doesn't return ETAG and last modification timestamp; but HTTP does have well-documented RFC to illustrate cache control policy.

I asked my agent to summarize the cache usability rule, TLDR
- If the cache expiration timestamp has reached, discard the cache entries and move forward
- last modification timestamp and ETAG is still checked to decide whether cache entry could be reused

Implementation change
- Response headers could contain duplicate keys (i.e., cache-control), so change from unordered map to map<str, vec<str>>
- Cache expiration timestamp is calculated at storage backends and propagated at file metadata

```mermaid
flowchart TD
    OPEN[Open file and snapshot Stats metadata] --> STALE{Cached or current deadline = −∞?}
    STALE -- Yes --> INVALID[Clear cached blocks]
    STALE -- No --> VALIDATE{Validation enabled?}

    VALIDATE -- No --> KEEP[Keep existing blocks]
    VALIDATE -- Yes --> ETAG{Cached or current ETag exists?}
    ETAG -- Yes --> ETAG_EQ{ETags equal?}
    ETAG_EQ -- No --> INVALID
    ETAG_EQ -- Yes --> REVALIDATED[Keep blocks and refresh deadline]

    ETAG -- No --> LM{Usable Last-Modified exists?}
    LM -- Yes --> LM_EQ{Timestamps equal and safely old?}
    LM_EQ -- No --> INVALID
    LM_EQ -- Yes --> REVALIDATED

    LM -- No --> SIZE{File sizes equal?}
    SIZE -- No --> INVALID
    SIZE -- Yes --> ORIGINAL_FRESH{Original deadline exists<br/>and now ≤ deadline?}
    ORIGINAL_FRESH -- No --> INVALID
    ORIGINAL_FRESH -- Yes --> PRESERVE[Keep blocks and original deadline]

    INVALID --> CURRENT[Read current file data]
    KEEP --> SERVE{May cached blocks be served?}
    REVALIDATED --> SERVE
    PRESERVE --> SERVE

    SERVE --> MODE{Validation enabled?}
    MODE -- No --> NO_VALIDATE_FRESH{Deadline absent or fresh?}
    NO_VALIDATE_FRESH -- Yes --> CACHE[Use cached blocks]
    NO_VALIDATE_FRESH -- No --> CURRENT

    MODE -- Yes --> HAS_DEADLINE{Cached deadline present?}
    HAS_DEADLINE -- Yes --> DEADLINE_FRESH{Now ≤ deadline?}
    DEADLINE_FRESH -- Yes --> CACHE
    DEADLINE_FRESH -- No --> CURRENT
    HAS_DEADLINE -- No --> HAS_VALIDATOR{Current ETag or usable<br/>Last-Modified?}
    HAS_VALIDATOR -- Yes --> CACHE
    HAS_VALIDATOR -- No --> CURRENT
```

```mermaid
flowchart TD
    HEADERS[Collect every Cache-Control field line<br/>and split directives outside quoted strings] --> NO_CACHE{no-cache present?}
    NO_CACHE -- Yes --> NEG[−∞: explicitly stale]
    NO_CACHE -- No --> MAXAGE{max-age present?}

    MAXAGE -- Duplicate or invalid --> NEG
    MAXAGE -- Valid --> LIFETIME[Freshness lifetime]
    MAXAGE -- No --> EXPIRES{Expires present?}
    EXPIRES -- Invalid --> NEG
    EXPIRES -- No --> UNKNOWN[nullopt: no explicit freshness]
    EXPIRES -- Valid --> REFERENCE[Expires − Date<br/>or receipt time]
    REFERENCE --> LIFETIME

    LIFETIME --> CURRENT_AGE[Current age = max Age,<br/>apparent age from Date]
    CURRENT_AGE --> EXPIRED{Lifetime ≤ current age<br/>or arithmetic overflow?}
    EXPIRED -- Yes --> NEG
    EXPIRED -- No --> DEADLINE[Receipt time + remaining lifetime]

    NEG --> POLICY[cache_valid_until]
    UNKNOWN --> POLICY
    DEADLINE --> POLICY
    NO_STORE[no-store present] --> NEG
    NO_STORE --> DONT_STORE[Do not insert HTTP metadata]
    POLICY --> META{HTTP metadata lookup}
    META -- Finite deadline expired --> EVICT[Erase metadata and refetch]
    META -- Otherwise --> REUSE[Reuse metadata]
```

ETag and Last-Modified determine whether cached content still represents the file. The tri-state deadline then determines whether that content may be served.

Priority is: **ETag → Last-Modified → file size plus freshness deadline**. The deadline is only used when neither validator exists.

Reference:
- https://datatracker.ietf.org/doc/html/rfc9111
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control


rustyconover:
This is good to see, feel free to look at how I fixed conditional response validation here:

https://github.com/Query-farm-haybarn/haybarn-httpfs/commit/4f100bba05fea83dc8192aa15a14423879002401

.github/patches/extensions/httpfs/0001-cache-metadata.patch
```diff
@@ -0,0 +1,309 @@
+commit 82427d21c7fbbe8dac740c64524515d2f43fef70
```
lnkuiper: This patch no longer applies cleanly

.github/patches/extensions/httpfs/0001-cache-metadata.patch
```diff
@@ -0,0 +1,309 @@
...
++		}
++	}
++	if (!has_freshness_lifetime) {
++		if (has_max_age || headers.HasHeader("Expires")) {
++			return timestamp_t::ninfinity();
++		}
++		return nullopt;
++	}
++
++	// Account for both Age and the apparent age derived from Date (RFC 9111).
```
lnkuiper: This omits `response_delay` from RFC 9111’s age calculation, so slow requests can extend freshness past `max-age`. Could we include `Age + (response_time - request_time)` here?

src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -241,22 +261,30 @@ shared_ptr<FileHandle> CachingFileHandle::GetFileHandle() {
...
+			    ExternalFileCache::IsValid(Validate(), cached_file->validation_info, validation_info);
+			if (!cache_is_valid) {
 				annotated_lock_guard<annotated_mutex> map_guard(cached_file->map_lock);
 				cached_file->blocks.clear();
 				cached_file->cached_block_size.SetInvalid();
 			}
-			cached_file->file_size = file_handle->GetFileSize();
-			cached_file->last_modified = last_modified;
-			cached_file->version_tag = version_tag;
+			// A successful validator check refreshes freshness. Without validators, preserve the original deadline.
```
lnkuiper, line 324: This drops the first finite deadline under `NO_VALIDATION`, leaving `nullopt` to mean unlimited freshness. I reproduced a `max-age=1` response being reused after two seconds; could we retain the current deadline and add a test?

src/storage/external_file_cache/external_file_cache.cpp
```diff
@@ -258,6 +258,31 @@ bool ExternalFileCache::IsValid(bool validate, const string &cached_version_tag,
...
+	if (cached.IsExplicitlyStale() || current.IsExplicitlyStale()) {
+		return false;
+	}
+	if (!validate) {
+		return true; // Assume valid
+	}
+	if (HasValidationMetadata(cached) || HasValidationMetadata(current)) {
+		return IsValid(validate, cached.version_tag, cached.last_modified, current.version_tag, current.last_modified);
+	}
+	// No validators at all: cached data may be served within the freshness deadline the storage backend granted
```
lnkuiper, line 284: This ignores `Vary`, so different request-header variants can reuse the same cached blocks. I reproduced Alice’s response being returned for Bob; could we include the varied fields in the key or disable this reuse when `Vary` is present?

lnkuiper (changes requested):
I had Codex try to break this and it found a few things. Could you have a look?

.github/patches/extensions/httpfs/0001-cache-metadata.patch
dentiny: Yeah `httpfs` extension evolves pretty fast recently; rebased and re-patched

.github/patches/extensions/httpfs/0001-cache-metadata.patch
dentiny: good catch, fixed

src/storage/external_file_cache/caching_file_system.cpp
dentiny: Ah it's because cache file's validation info hasn't been assigned; handled first access separately

src/storage/external_file_cache/external_file_cache.cpp
dentiny: Make sense -- I took the easy path for now, simply ignore caching for `vary` header response

.github/patches/extensions/httpfs/0001-cache-metadata.patch
```diff
@@ -0,0 +1,399 @@
...
++static bool TryParseNonNegativeSeconds(const string &input, int64_t &result) {
++	return TryCast::Operation<string_t, int64_t>(string_t(input), result) && result >= 0;
++}
++
++static vector<string> GetCacheControlDirectives(const HTTPHeaders &headers) {
++	vector<string> result;
++	if (!headers.HasHeader("Cache-Control")) {
++		return result;
++	}
++	for (const auto &header_value : headers.GetHeaderValues("Cache-Control")) {
```
lnkuiper, line 106: The actual `httpfs` response adapters still use `Insert`, so repeated headers are collapsed before this sees them. Could we use `Append` in both the curl and httplib response paths too?

.github/patches/extensions/httpfs/0002-cache-metadata.patch
```diff
@@ -0,0 +1,399 @@
...
+ 	auto content_size = HTTPFileInfoParser::TryParseContentRange(response.headers);
+ 	if (!content_size.IsValid()) {
+@@ -750,6 +924,13 @@ void HTTPFileHandle::ApplyFileInfo(const HTTPResponse &response) {
+ 	if (response.headers.HasHeader("ETag")) {
+ 		etag = response.headers.GetHeaderValue("ETag");
+ 	}
++	cache_valid_until = HTTPFileSystem::ComputeCacheValidUntil(response.headers, request_time, response_time);
++	// TODO(hjiang): `Vary` header indicates that the response should be served separately based on the request headers,
++	// currently it's not cached.
++	cache_reuse_disabled = HasCacheControlDirective(response.headers, "no-store") || response.headers.HasHeader("Vary");
```
lnkuiper: This only sees `Vary` on the metadata response. HEAD may omit fields determined while generating the content. Could we also apply the cache policy from the response whose bytes we store?

src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -242,22 +262,31 @@ shared_ptr<FileHandle> CachingFileHandle::GetFileHandle() {
...
+	auto stats = caching_file_system.file_system.Stats(*file_handle);
+	validation_info.file_size = NumericCast<idx_t>(stats.file_size);
+	validation_info.last_modified = stats.last_modification_time;
+	validation_info.cache_valid_until = stats.cache_valid_until;
+	validation_info.version_tag = std::move(stats.version_tag);
 
 	{
 		annotated_lock_guard<annotated_mutex> meta_guard(cached_file->meta_lock);
-		const bool first_access = (cached_file->file_size == 0);
+		const bool first_access = (cached_file->validation_info.file_size == 0);
```
lnkuiper, line 313: This still leaves stale metadata under `NO_VALIDATION` after the deadline expires. Could we refresh the cached metadata when expiry forces this handle open?

lnkuiper (commented):
Thanks for the changes! I had Codex try to break the updated version and it found a few more things. I've posted more comments below.

There is also this issue, which we can fix in a follow-up PR:
> Cached files are shared across connections and keyed only by path, even when the response depends on `Authorization`. We should avoid sharing `private` or authenticated responses between request identities.

Not sure what this should look like yet.

.github/patches/extensions/httpfs/0002-cache-metadata.patch
dentiny: TIL, didn't realized that; I added an `ApplyCachePolicy` in httpfs to apply cache control for all GET and HEAD HTTP requests.

dentiny:
> Cached files are shared across connections and keyed only by path, even when the response depends on Authorization. We should avoid sharing private or authenticated responses between request identities.

Sure -- I've left TODO items on cache entry sharability (i.e., private, vary)

src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -331,39 +386,15 @@ FileBufferHandleGroup CachingFileHandle::Read(const idx_t nr_bytes, const idx_t
...
-		const annotated_lock_guard<annotated_mutex> meta_guard(current_cached_file->meta_lock);
-		if (!ExternalFileCache::IsValid(true, current_cached_file->version_tag, current_cached_file->last_modified,
-		                                current_version_tag, current_last_modified)) {
-			// Do not reset blocks in place: another reader may already have pinned the same blocks and still need their
-			// byte counts. Removing only the matching map entries preserves those readers while forcing future reads
-			// to fetch fresh blocks.
-			external_file_cache.RetireBlocks(*current_cached_file, first_block, blocks);
-		}
-	}
+	ReconcileCacheAfterRead(*current_cached_file, first_block, blocks);
```
lnkuiper, line 397: This is still too late when `Vary` only appears on GET. Concurrent readers can share the in-flight block before either handle publishes the GET policy, and retiring it afterward does not fix the already-pinned result.

lnkuiper (commented):
Thanks for the changes! I have one more comment. Do you think we should block on that or no?

src/storage/external_file_cache/caching_file_system.cpp
dentiny: Good catch! I think it's a real issue, currently we marked the fetched block as LOADED anyway, so it's shared regardless of its cache control; updated to EMPTY state to trigger re-fetch.

dentiny:
Followup items:
- Update httpfs extension with patch and DuckDB core commit, add unit test for cache-control related unit tests
- Properly implement `vary` and `private` (and maybe other?) sharability related metadata

lnkuiper:
Not sure if related, but there seems to be a build failure.

src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -81,8 +97,16 @@ class FetchBlockTask : public BaseExecutorTask {
 					auto buf =
 					    ExternalFileCache::AllocateCacheBuffer(buffer_manager, caching_file_handle.GetPath(), to_read);
 					caching_file_handle.ReadAndRecord(context, buf.GetDataMutable(), to_read, offset);
+					const bool share_block = !caching_file_handle.IsCacheReuseProhibited();
 
 					lk.lock();
+					if (!share_block) {
```
lnkuiper, line 103: Could we add a regression test with a second reader waiting on the same `LOADING` block? Just to be sure we don't have a race condition here

dentiny:
> Not sure if related, but there seems to be a build failure.

Yeah I think we have conflicting updates to httpfs extension haha

dentiny:
Hey @lnkuiper , I'd appreciate if you could take another look and let me know if it's ready, I see you making frequent updates in httpfs extension so we have to follow and rebase patches on the latest version.. Thanks!

lnkuiper (approved):

lnkuiper:
LGTM now. I'll wait with more httpfs changes until this is merged!

.github/patches/extensions/httpfs/0001-cache-metadata.patch
```diff
@@ -0,0 +1,575 @@
...
++	auto response_valid_until =
++	    HTTPFileSystem::ComputeCacheValidUntil(response.headers, request_time, response_time);
++	// TODO(hjiang): Prevent shared reuse of private responses and authenticated requests without explicit permission.
++	if (HasCacheControlDirective(response.headers, "no-store")) {
++		// no-store applies to this response, so only blocks populated by this read are retired.
++		response_valid_until = timestamp_t::ninfinity();
++	}
++	if (response.headers.HasHeader("Vary")) {
++		// TODO(hjiang): Include Vary-selected request headers in the cache key instead of disabling reuse.
++		response_valid_until = timestamp_t::ninfinity();
```
Tishj, line 252: This broke the caching that iceberg relies on (https://github.com/duckdb/duckdb-iceberg/pull/1371)

For example:
```
(const duckdb::HTTPResponse) $5 = {
  status = OK_200
  url = "http://127.0.0.1:9000/warehouse/default/table_unpartitioned/metadata/snap-8636654708131365959-1-82ffaeb2-8a75-434f-aebb-1dd760a1c152.avro"
  body = ""
  request_error = ""
  reason = "OK"
  headers = {
    headers = size=15 {
      [0] = (first = "Content-Length", second = "4455")
      [1] = (first = "X-Content-Type-Options", second = "nosniff")
      [2] = (first = "Strict-Transport-Security", second = "max-age=31536000; includeSubDomains")
      [3] = (first = "Date", second = "Mon, 07 Sep 2026 09:09:09 GMT")
      [4] = (first = "Content-Type", second = "application/octet-stream")
      [5] = (first = "Accept-Ranges", second = "bytes")
      [6] = (first = "Server", second = "MinIO")
      [7] = (first = "X-Amz-Id-2", second = "dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8")
      [8] = (first = "X-Ratelimit-Remaining", second = "3332")
      [9] = (first = "ETag", second = "\"9e1e4d307319172a071bfb196e175088\"")
      [10] = (first = "X-Xss-Protection", second = "1; mode=block")
      [11] = (first = "Vary", second = "Origin")
      [12] = (first = "X-Amz-Request-Id", second = "18D2FECADA2BC85F")
      [13] = (first = "Last-Modified", second = "Mon, 07 Sep 2026 08:49:20 GMT")
      [14] = (first = "X-Ratelimit-Limit", second = "3332")
    }
    repeated_headers = size=1 {
      [0] = {
        first = "Vary"
        second = {
          std::__1::vector<std::__1::string, std::__1::allocator<std::__1::string> > = size=2 {
            [0] = "Origin"
            [1] = "Accept-Encoding"
          }
        }
      }
    }
  }
  success = true
}
```

PR #24588, opened 2026-08-07
Optimize boolean identity `CASE` at filter root

### LINKED ISSUE #24112: Simple CASE Boolean Identity Not Simplified
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE t AS
         SELECT i::INTEGER AS a
         FROM range(10000) tbl(i);

-- should be simplified to a > 0
memory D EXPLAIN ANALYZE
         SELECT *
         FROM t
         WHERE CASE WHEN a > 0 THEN TRUE ELSE FALSE END;
╭─ Summary ───────────╮
│ Total Time: 0.0004s │
╰─────────────────────╯
╭─ Table Scan ────────────────────────────────────╮
│ Table: memory.main.t                            │
│ Type: Sequential Scan                           │
│ Projections: a                                  │
│ Filters:                                        │
│ CASE  WHEN ((a > 0)) THEN (true) ELSE false END │
│ Row Groups Scanned: 1 / 1                       │
│ 9,999 rows                                  0µs │
╰─────────────────────────────────────────────────╯
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

tempewda:
Fixes #24112

This PR adds an optimization rule to `CaseSimplificationRule` to simplify `CASE WHEN p THEN TRUE ELSE FALSE END` down to `p` when the expression is the root of a `LogicalFilter`. Because `LogicalFilter` discards both `FALSE` and `NULL`, it prevents semantic changes in contexts where `FALSE` and `NULL` behave differently (e.g., projections, `CHECK` constraints, or inside a `NOT(...)`).

Also added regression tests verifying the query plan optimization, and a control test ensuring `NULL` semantics are perfectly preserved in a projection context.

### DIFF
--- a/src/optimizer/rule/case_simplification.cpp
+++ b/src/optimizer/rule/case_simplification.cpp
@@ -2,8 +2,19 @@
 
 #include "duckdb/execution/expression_executor.hpp"
 #include "duckdb/planner/expression/bound_case_expression.hpp"
+#include "duckdb/planner/expression/bound_constant_expression.hpp"
 
 namespace duckdb {
+
+static bool IsBooleanConstant(const Expression &expr, bool expected_value) {
+	if (expr.GetExpressionClass() != ExpressionClass::BOUND_CONSTANT) {
+		return false;
+	}
+
+	const auto &constant = expr.Cast<BoundConstantExpression>().GetValue();
+	return !constant.IsNull() && constant.type() == LogicalType::BOOLEAN &&
+	       BooleanValue::Get(constant) == expected_value;
+}
 
 CaseSimplificationRule::CaseSimplificationRule(ExpressionRewriter &rewriter) : Rule(rewriter) {
 	// match on a CaseExpression that has a ConstantExpression as a check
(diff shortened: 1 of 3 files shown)


src/optimizer/rule/case_simplification.cpp
```diff
@@ -42,7 +53,23 @@ unique_ptr<Expression> CaseSimplificationRule::Apply(LogicalOperator &op, vector
 		// no case checks left: return the ELSE expression
 		return std::move(root.ElseMutable());
 	}
-	return nullptr;
+
+	// This rewrite is only valid at a filter root, where FALSE and NULL both reject the row.
+	if (!is_root || op.type != LogicalOperatorType::LOGICAL_FILTER) {
```
Mytherin, line 58: Instead of limiting this to the logical filter can we generate a `constant_or_null` expression here? That is then further optimized for cases like this.

Mytherin (commented):
Thanks for the PR! One comment

src/optimizer/rule/case_simplification.cpp
tempewda: Makes sense, the direct `WHEN` expression rewrite only preserves filtering semantics, not general expression semantics because it changes NULL to FALSE.

I was thinking of rewriting the CASE generically to:

```sql
p IS NOT DISTINCT FROM TRUE
```
which preserves the original `TRUE/FALSE/FALSE` results for `p = TRUE/FALSE/NULL`. `ComparisonSimplificationRule` already simplifies `p IS NOT DISTINCT FROM TRUE` to `p` at a root `LogicalFilter`.

Is that the intended `NULL`-preserving rewrite, or did you have a particular `constant_or_null` form in mind?

A standalone `constant_or_null(FALSE, p)` would not preserve the TRUE case, so I want to make sure I understand the intended shape.

PR #24589, opened 2026-08-07
Make configs and secrets use Identifiers

JelteF:
Config names and secret names/types/providers are now changed throughout the code to work with `Identifier`.


Mytherin:
Thanks!

PR #24591, opened 2026-08-07
Fix trigger dependency tracking so DROP TABLE respects trigger bodies

### LINKED ISSUE #24208: Trigger Body Dependency Not Protected
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t(i INT);
memory D CREATE TABLE log(i INT);
memory D CREATE TRIGGER tr AFTER INSERT ON t FOR EACH STATEMENT
           INSERT INTO log VALUES (1);
memory D 
memory D DROP TABLE log RESTRICT;
memory D SELECT count(*) FROM duckdb_triggers();
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
memory D INSERT INTO t VALUES (2);
Catalog Error:
Table with name log does not exist!
Did you mean "pg_collation"?
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

EtgarDev:
Fixes #24208.

If you create a trigger whose body reads from or writes to another table, DuckDB currently lets you drop that other table anyway, even though the trigger still needs it. For example:

```sql
CREATE TRIGGER tr AFTER INSERT ON t FOR EACH STATEMENT INSERT INTO log VALUES (1);
DROP TABLE log RESTRICT;   -- succeeds today, but shouldn't
INSERT INTO t VALUES (2);  -- now fails: Catalog Error: Table with name log does not exist!
```

This should be blocked the same way dropping a table used by a view is already blocked, but it wasn't, because of three separate gaps:

1. When a trigger was created, its list of dependencies was computed correctly but then silently thrown away and never actually saved on the trigger itself.
2. Binding the trigger's body didn't track which tables, views, or functions it referenced.
3. Resolving the target table of an INSERT statement used a shortcut that bypassed the mechanism that records dependencies in the first place.

This PR fixes all three, so a table referenced by a trigger body is now correctly protected from being dropped, or altered in ways that would break it, until the trigger is dropped or the dependency is otherwise removed.

Two extra details worth calling out:

- Trigger names are only unique per table, not per schema, so two different triggers can share a name as long as they're on different tables, so I had to add the `table` field to the catalog entry
- Fixing the bugs above also means a trigger now genuinely depends on the table it's defined on, not just on the other tables its body touches. That dependency is explicitly exempted from blocking, so dropping or altering the table a trigger is defined on still works without needing CASCADE, the same way it already works for tables with indexes on them.


src/catalog/dependency_manager.cpp
```diff
@@ -350,6 +368,12 @@ CatalogEntryInfo DependencyManager::GetLookupProperties(const CatalogEntry &entr
 		auto &dependency_entry = entry.Cast<DependencyEntry>();
 		return dependency_entry.EntryInfo();
 	}
+	if (entry.type == CatalogType::TRIGGER_ENTRY) {
+		// triggers live in the catalog set of the table they are defined on, and are only unique within that table
+		auto &trigger = entry.Cast<TriggerCatalogEntry>();
+		return CatalogEntryInfo {entry.type, GetSchemaPath(entry), entry.name,
+		                         trigger.base_table->GetQualifiedName().Name()};
```
lnkuiper, line 364: The dependency protection disappears after renaming the trigger’s owning table. `GetLookupProperties` derives the table qualifier from `trigger.base_table`, but table rename does not update that field. The dependency therefore still identifies `tr` as belonging to `t`, and looking it up through a dependency on log returns nothing.

```sql
CREATE TABLE t(i INT);
CREATE TABLE log(i INT);
CREATE TRIGGER tr AFTER INSERT ON t
  FOR EACH STATEMENT INSERT INTO log VALUES (1);

ALTER TABLE t RENAME TO t2;
DROP TABLE log RESTRICT;  -- succeeds
INSERT INTO t2 VALUES (1); -- log no longer exists
```

`duckdb_triggers()` still reports `table_name = 't'` after the rename. Could we update and re-key the triggers during table rename, or maybe it's easier to keep this alter blocked until their owning-table identity can be migrated.

src/planner/binder/statement/bind_create.cpp
```diff
@@ -718,6 +718,14 @@ SchemaCatalogEntry &Binder::BindCreateTriggerInfo(CreateTriggerInfo &create_trig
 	validation_binder->global_binder_state->trigger_expanded_tables.insert(table);
 	validation_binder->global_binder_state->trigger_creation_table = &table;
 	validation_binder->global_binder_state->trigger_creation_name = create_trigger_info.GetTriggerName();
+	// Track every catalog entry resolved while binding the trigger body as a dependency,
+	// so DROP of a table/view/function referenced by the body is blocked.
+	validation_binder->SetCatalogLookupCallback([&create_trigger_info, &catalog](CatalogEntry &entry) {
+		if (&catalog != &entry.ParentCatalog()) {
```
lnkuiper, line 737: Dependencies in another catalog are silently ignored, which leaves the same broken-trigger state for temporary or attached tables. Since cross-catalog dependency edges are unsupported, I think trigger creation should reject these bodies instead of accepting an untracked dependency.

```sql
CREATE TABLE t(i INT);
CREATE TEMP TABLE log(i INT);
CREATE TRIGGER tr AFTER INSERT ON t
  FOR EACH STATEMENT INSERT INTO temp.main.log VALUES (1);

DROP TABLE temp.main.log RESTRICT; -- succeeds
INSERT INTO t VALUES (1);          -- log no longer exists
```

src/include/duckdb/catalog/dependency.hpp
```diff
@@ -143,6 +143,8 @@ struct CatalogEntryInfo {
 	//! is the path of the *containing* schemas - for a schema entry it is the parent chain (not including itself).
 	vector<Identifier> schema_path;
 	Identifier name;
+	//! The table that owns this entry, for entries that are not unique within their schema (triggers)
+	Identifier table;
```
lnkuiper, line 175: I think we might want to reconsider `schema_path` and `table` here. We only recently added `schema_path` so we can probably still change.

Maybe we can do:
```C++
struct CatalogEntryPathComponent {
	CatalogType type;
	Identifier name;
};

vector<CatalogEntryPathComponent> path;
```
Currently `schema_path` is assumed to be `SCHEMA_ENTRY` only, but I'd argue that the table is on the path to the trigger, so we could add it directly to the vector. @Mytherin probably needs to have a look as well.

src/catalog/dependency_manager.cpp
```diff
@@ -297,6 +302,14 @@ void DependencyManager::CreateDependency(CatalogTransaction transaction, Depende
 	CreateSubject(transaction, info);
 }
 
+//! Whether the subject is the table that the dependent trigger is defined on
+static bool IsOwningTableOfTrigger(const CatalogEntryInfo &dependent, const CatalogEntryInfo &subject) {
+	if (dependent.type != CatalogType::TRIGGER_ENTRY || subject.type != CatalogType::TABLE_ENTRY) {
+		return false;
+	}
+	return subject.schema_path == dependent.schema_path && subject.name == dependent.table;
```
lnkuiper: Could we avoid reconstructing this relationship from the entry names? The binder knows this is the trigger’s containing table when it adds the dependency, and `DependencyDependentFlags` already represents an automatic dependency. If the logical dependency carried and serialized that flag, `CreateDependencies` could consume it directly and `AlterObject` could inspect the stored flag instead of both paths having trigger-specific checks.

lnkuiper (changes requested):
Thanks! I've left some comments

src/planner/binder/statement/bind_create.cpp
EtgarDev: thank you! fixed

src/catalog/dependency_manager.cpp
EtgarDev: good call - fixed.
We're now serialising the dependency flags and use them in both places.

A thought that crossed my mind is if we should maybe consider to have two separate type of dependency flags - one for alteration and the other for drop (we could possibly have the same field but as a bitmask flag), so we won't have to treat the edge-case of "referencing triggers" dependencies.
We may do it when we get more than 1 edge-case like this.

src/catalog/dependency_manager.cpp
EtgarDev: I could now notice that renaming referenced tables used to break the triggers also before this fix, so after the fix in this branch we are left with:

1. Referenced table renaming breaks triggers -> resolved naturally
2. Base table renaming breaks triggers -> still there, we can simply restrict it for now

I think restricting renaming now would be a good decision until we design a solution.
wdyt? @lnkuiper

src/planner/binder/statement/bind_create.cpp
```diff
@@ -633,6 +633,10 @@ SchemaCatalogEntry &Binder::BindCreateTriggerInfo(CreateTriggerInfo &create_trig
 		throw BinderException("CREATE TRIGGER requires a base table");
 	}
 	auto &table = *table_ptr;
+	// Dropping/altering the trigger's own table must not require CASCADE, so this dependency is non-blocking. If the
+	// trigger body also references this same table, the body-reference dependency below is added as blocking. since a
+	// duplicate entry is a no-op, adding the non-blocking one first is what makes it win.
+	create_trigger_info.dependencies.AddDependency(table, DependencyDependentFlags());
```
lnkuiper, line 647: Letting this dependency win discards a blocking body dependency when the trigger body reads from its own table. `AlterObject` consequently permits
column alters that invalidate the body:

```sql
CREATE TABLE t(i INTEGER, j INTEGER);
CREATE TABLE audit(x INTEGER);

CREATE TRIGGER tr AFTER INSERT ON t FOR EACH STATEMENT
INSERT INTO audit SELECT j FROM t;

ALTER TABLE t RENAME COLUMN j TO k; -- succeeds
INSERT INTO t VALUES (1, 2);
-- Binder Error: Referenced column "j" not found
```

The owning relationship needs to remain automatic for `DROP`, while the body reference must still block alters that can invalidate it. I think this requires the separate drop/alter semantics mentioned in the `AlterObject` TODO.

lnkuiper (commented):
Thanks for the changes! There's still an issue (see comment below). We can also discuss the earlier comment I had in `dependency.hpp` once Mark is available again.

src/planner/binder/statement/bind_create.cpp
EtgarDev: I thought I can leave this TODO I wrote for an upcoming PR, but as the decision is now clearer as it would help us solve this problem as well, very easily, I chose to do it know.
Done :)

src/catalog/dependency_manager.cpp
```diff
@@ -318,8 +319,12 @@ void DependencyManager::CreateDependencies(CatalogTransaction transaction, const
 
 	// add the object to the dependents_map of each object that it depends on
 	for (auto &dependency : dependencies.Set()) {
+		auto flags = legacy_default;
+		if (!dependency.flags.IsBlocking()) {
```
lnkuiper: This discards `ALTER_BLOCKING` whenever the dependency also blocks `DROP`. Trigger-body references to external objects have both flags, so `IsBlocking()` is `true` and the combined value is replaced with `legacy_default`, which only contains `BLOCKING`. `AlterObject` then permits column alters that leave the trigger body stale.

```sql
CREATE TABLE t(i INTEGER);
CREATE TABLE src(j INTEGER);
CREATE TABLE audit(x INTEGER);

CREATE TRIGGER tr AFTER INSERT ON t FOR EACH STATEMENT
INSERT INTO audit SELECT j FROM src;

ALTER TABLE src RENAME COLUMN j TO k; -- succeeds
INSERT INTO t VALUES (1);
-- Binder Error: Referenced column "j" not found
```

Could we preserve the complete explicit flag value here and handle legacy-index compatibility without using `IsBlocking()` as a presence marker?

lnkuiper (commented):
Thanks for the changes! Another comment:

src/catalog/dependency_manager.cpp
EtgarDev: @lnkuiper 
I can maybe just check if the entry is index, and whether the flags we have is actually the same as the overall "legacy default" (which means the flags are probably from an older compatibility version, as indexes should be written without the `IS_BLOCKING` bit) - then we can just take the non-drop-blocking flag but also the non-alter-blocking flag (0x00 flag)).

```cpp
	// backward compatibility for indexes: they differed from the default and were actually never blocking, so correct that
	// legacy placeholder value for them specifically,
	// for storage files that were written before we started serializing flags
	static const DependencyDependentFlags legacy_marker = DependencyDependentFlags().SetBlocking();
	for (auto &dependency : dependencies.Set()) {
		auto flags = dependency.flags;
		if (object.type == CatalogType::INDEX_ENTRY && flags == legacy_marker) {
			// the legacy flags used to be for INDEX_ENTRY before we started serializing flags into the storage
			flags = DependencyDependentFlags();
		}
		DependencyInfo info {
		    /*dependent = */ DependencyDependent {object_info, flags},
		    /*subject = */ DependencySubject {dependency.entry, DependencySubjectFlags(), optional_idx()}};
		CreateDependency(transaction, info);
	}
```

This is also behind the scenes relies on the `IS_BLOCKING` bit, so there is a concern of someone in the future may write a "special" index entry that for some reason *would* block `DROP` and then our way of "recognising" the flags as "out-of-default" would no longer be something we can
 rely on.

I can also add a custom code to the deserialization/serialization functions that would mark another bit with "out-of-default", but this would make the mechanism a bit more dirtier.

WDYT?

EtgarDev: @lnkuiper 
In addition, I found another problem in this new mechanism - basically if someone used triggers that are referring to a table `t` in a prior version to this PR, then the `ALTER` will continue not being blocked.
However, it is also not blocked today at all, and triggers are not officially released yet so it makes sense to "preserve" this bug in triggers that were written on an experimental version IMO.

EtgarDev: I anyway pushed this code for now https://github.com/duckdb/duckdb/pull/24591/commits/8f0f4b963c81e1b7f3aa25dff403fb6251be3852

lnkuiper: Good fix!

lnkuiper:
Looks good to me now, thanks for all the changes! Just need the conflict to be resolved, then this can be merged

lnkuiper (approved):

PR #24592, opened 2026-08-07
Fix join projection map bug in TopNWindowElimination

d-justen:
fixes https://github.com/duckdblabs/duckdb-internal/issues/10059
fixes https://github.com/duckdb/duckdb/issues/23925
supersedes https://github.com/duckdb/duckdb/pull/23943

This PR fixes a bug that would pull the rowid through a join by adding it to its projection map, even if that map is empty.  As an empty map projects all columns, we would then only project the rowid and lose everything else.


hawkfish (approved):
LGTM!

PR #24594, opened 2026-08-07
Erase `Function::EraseArgument`

Maxxen:
Having functions modify the expression tree when bound by e.g. removing expressions causes all kinds of headaches for optimizer passes and de/serialization. This PR removes the ability to do that (easily). We should investigate ways to reduce the cost of executing "constant" arguments separately, instead of outright removing them.

Also removes the `original_arguments` vector as it is no longer needed.

I've included BWC plan-serialization tests for the affected aggregates/window functions, but we should probably triple-check this to be sure (and see what CI says).


Mytherin:
Thanks for the PR! Cool stuff - can you just have a look at the CI failure (maybe this just needs a merge with main?)

Maxxen:
CI failure 🎉 unrelated 🎉

PR #24595, opened 2026-08-07
Inject markers, network, memory, and http requests into a Samply profile

smvv:
### Context

I added network throughput, memory usage (RSS), and individual HTTP requests to samply's profiler output so we can interactively explore how duckdb uses system resources.

Screenshot 2026-08-07 at 16 13 42

I used this sql:
```sql
SET samply_tracks = 'all';

SELECT count(*)
FROM 'https://blobs.duckdb.org/data/taxi_2019_04.parquet'
WHERE pickup_at BETWEEN '2019-04-15' AND '2019-04-20';
```
for that screenshot, and the profile can be viewed online: https://share.firefox.dev/4waiAr9

the data is injected into the profile after samply is done. There is no "dependency" on samply to make this work in duckdb core.

i've also added "markers" for the queries, their plan/execute phases and pipelines (based on the existing duckdb profiler data),  and the network requests.

image

### Usage

```shell
build/release/profile -f taxi.sql
```

### TODO
- [ ] Rebase and resolve conflicts.
- [ ] Verify network and memory track values.
- [ ] Verify http requests timeline.

### DIFF
--- a/scripts/samply_profile.py
+++ b/scripts/samply_profile.py
@@ -529,7 +529,14 @@ def duckdb_arguments(arguments: list[str], launcher_path: str | Path) -> list[st
     else:
         wrapper_arguments = []
         command_arguments = arguments
-    return [*wrapper_arguments, "--", str(duckdb), *command_arguments]
+    return [
+        *wrapper_arguments,
+        "--",
+        str(duckdb),
+        "-cmd",
+        "SET samply_tracks='all';",
+        *command_arguments,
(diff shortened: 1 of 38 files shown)


Mytherin:
Thanks for the PR!

I'm not a big fan of adding specialised code for profiling tools like this if we can avoid it - could we instead lift on the (existing) metrics / logging infrastructure to enable this functionality?

PR #24598, opened 2026-08-07
Change extensions and catalog/profiler/dialect types to use Identifier

JelteF:
Extension names and catalog/profiler/dialect types both make sense as Identifiers. In various places in the code an "extension" variable didn't mean "extension name" but instead "extension name or path", those variable names have been made more descriptive and don't use `Identifier`.


JelteF:
This one is finally passing CI now

JelteF:
This one is finally passing CI now, could someone have a look?

PR #24599, opened 2026-08-07
Clean up orphaned temp spill files on startup after a crash

### LINKED ISSUE #24576: Orphaned temporary files are not cleaned up
### What happens?

When DuckDB is terminated unexpectedly, for example by an OOM kill, temporary files created by a query may remain in `temporary_directory`, files such as `duckdb_temp_*.tmp` and `duckdb_temp_*.block`. After DuckDB restarts, these files are no longer needed but are never cleaned up.

### To Reproduce

1. Set `temp_directory` to an empty directory and set a low `memory_limit`.
2. Run a disk-spilling query, for example:
   `SELECT * FROM range(100000000) ORDER BY random();`
3. Terminate DuckDB with `kill -9` while the query is running.
4. Restart DuckDB and observe that `duckdb_temp_*` files remain in `temp_directory`.

### OS:

any

### DuckDB Version:

1.5.2

### DuckDB Client:

C API

### Hardware:

_No response_

### Full Name:

Xiaocong Ding

### Affiliation:

Aliyun Cloud

- [ ] Yes, I have

- [x] Yes, I have

mattyy-k:
Fixes #24576 

## Problem

DuckDB spills data to disk under memory pressure, creating files like
`duckdb_temp_storage_DEFAULT-0.tmp` in `temp_directory`. On a clean shutdown,
these are deleted in `TemporaryDirectoryHandle`'s destructor. On an unclean
shutdown (`kill -9`, OOM kill, power loss), no destructors run, so the files
are orphaned — the process that created them no longer exists, and DuckDB
never revisits them. The temp directory accumulates garbage indefinitely.

## Fix

`TemporaryDirectoryHandle`'s constructor already checks whether
`temp_directory` exists. If it doesn't, it creates it (fresh start, nothing
to clean). This adds an `else` branch: if the directory *does* already
exist, it may contain orphans from a previous crash, so we scan for files
matching the `duckdb_temp_` prefix and remove them before this process
writes anything of its own. This mirrors the scan/delete logic the
destructor already uses on clean shutdown — just moved to also run on the
"directory already existed" startup path.

## Caveats (by design)

- **Cleanup is lazy.** `TemporaryDirectoryHandle` is only constructed the
  first time this process needs to spill (`RequireTemporaryDirectory`). A
  process that opens `temp_directory` but never hits memory pressure won't
  trigger cleanup — orphans persist until some future process actually
  spills into that directory.
- **Assumes `temp_directory` is not shared by concurrent live DuckDB
  processes.** This isn't a new risk from this change: temp file names
  (`duckdb_temp_storage_<size>-<index>.tmp`) contain no per-process
  identifier, so two processes sharing a directory would already collide
  on filenames during normal operation, independent of cleanup. The
  existing destructor cleanup makes the same assumption.

## Testing

Added a regression test that:
- creates orphaned `duckdb_temp_*` files alongside an unrelated file
- opens a new database using the same `temp_directory`
- forces creation of `TemporaryDirectoryHandle` via a spilling query
- verifies orphaned files are removed while the unrelated file remains

Also manually reproduced the original issue: forced a spill, `kill -9`'d
the process, restarted, confirmed orphaned files are now cleaned up on the
next spill (previously they persisted indefinitely).

### DIFF
--- a/src/storage/temporary_file_manager.cpp
+++ b/src/storage/temporary_file_manager.cpp
@@ -744,6 +744,25 @@ TemporaryDirectoryHandle::TemporaryDirectoryHandle(DatabaseInstance &db, string 
 	if (!fs.DirectoryExists(temp_directory)) {
 		fs.CreateDirectory(temp_directory);
 		created_directory = true;
+	} else {
+		// the directory already existed. This can happen if a previous DuckDB process using this
+		// temp directory crashed (e.g. kill -9, OOM kill, power loss) without running its destructors.
+		// any duckdb_temp_* files left behind are orphaned: no live process holds them open, and the
+		// block ids they reference are meaningless to this new process. Remove them before we start
+		// writing new files of our own, so garbage doesn't accumulate indefinitely across crashes.
+		vector<string> orphaned_files;
+		fs.ListFiles(temp_directory, [&](const string &path, bool isdir) {
+			if (isdir) {
+				return;
+			}
+			if (!StringUtil::StartsWith(path, "duckdb_temp_")) {
+				return;
+			}
+			orphaned_files.push_back(fs.JoinPath(temp_directory, path));
+		});
+		if (!orphaned_files.empty()) {
+			fs.RemoveFiles(orphaned_files);
+		}
 	}
 	temp_file->SetMaxSwapSpace(max_swap_space);
 }
--- a/src/storage/temporary_file_manager.cpp
+++ b/src/storage/temporary_file_manager.cpp
@@ -10,7 +10,18 @@
 #include "duckdb/common/encryption_functions.hpp"
 #include "zstd.h"
 
+#include <algorithm>
+
 namespace duckdb {
+
+static mutex temporary_file_ownership_lock;
+static vector<string> live_temporary_file_identifiers;
+
+static bool IsLiveTemporaryFileIdentifier(const string &identifier) {
+	lock_guard<mutex> guard(temporary_file_ownership_lock);
+	return std::find(live_temporary_file_identifiers.begin(), live_temporary_file_identifiers.end(), identifier) !=
+	       live_temporary_file_identifiers.end();
+}
 
 //===--------------------------------------------------------------------===//
 // TemporaryBufferSize
@@ -492,8 +503,9 @@ void TemporaryFileCompressionAdaptivity::Update(const TemporaryCompressionLevel 
 // TemporaryFileManager
 //===--------------------------------------------------------------------===//
 TemporaryFileManager::TemporaryFileManager(DatabaseInstance &db, const string &temp_directory_p,
-                                           atomic<idx_t> &size_on_disk_p)
-    : db(db), temp_directory(temp_directory_p), files(*this), size_on_disk(size_on_disk_p), max_swap_space(0) {
+                                           const string &temporary_file_identifier_p, atomic<idx_t> &size_on_disk_p)
+    : db(db), temp_directory(temp_directory_p), temporary_file_identifier(temporary_file_identifier_p), files(*this),
+      size_on_disk(size_on_disk_p), max_swap_space(0) {
 }
 
 TemporaryFileManager::~TemporaryFileManager() {
@@ -711,8 +723,13 @@ void TemporaryFileManager::EraseUsedBlock(TemporaryFileManagerLock &lock, block_
 
 string TemporaryFileManager::CreateTemporaryFileName(const TemporaryFileIdentifier &identifier) const {
 	return FileSystem::GetFileSystem(db).JoinPath(
-	    temp_directory, StringUtil::Format("duckdb_temp_storage_%s-%llu.tmp", EnumUtil::ToString(identifier.size),
-	                                       identifier.file_index.GetIndex()));
+	    temp_directory, StringUtil::Format("duckdb_temp_storage_%s_%s-%llu.tmp", temporary_file_identifier,
+	                                       EnumUtil::ToString(identifier.size), identifier.file_index.GetIndex()));
+}
+
+string TemporaryFileManager::CreateTemporaryBlockFileName(block_id_t id) const {
+	return FileSystem::GetFileSystem(db).JoinPath(
+	    temp_directory, StringUtil::Format("duckdb_temp_block_%s-%lld.block", temporary_file_identifier, id));
 }
 
 optional_ptr<TemporaryFileHandle> TemporaryFileManager::GetFileHandle(TemporaryFileManagerLock &,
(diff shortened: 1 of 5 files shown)


src/storage/temporary_file_manager.cpp
```diff
@@ -738,6 +738,25 @@ TemporaryDirectoryHandle::TemporaryDirectoryHandle(DatabaseInstance &db, string
...
+		// temp directory crashed (e.g. kill -9, OOM kill, power loss) without running its destructors.
+		// any duckdb_temp_* files left behind are orphaned: no live process holds them open, and the
+		// block ids they reference are meaningless to this new process. Remove them before we start
+		// writing new files of our own, so garbage doesn't accumulate indefinitely across crashes.
+		vector<string> orphaned_files;
+		fs.ListFiles(temp_directory, [&](const string &path, bool isdir) {
+			if (isdir) {
+				return;
+			}
+			if (!StringUtil::StartsWith(path, "duckdb_temp_")) {
```
dentiny, line 752: how do you make sure the temp file is not used by another database instance?

mattyy-k: That's a good point. As implemented, I don't have a way to distinguish orphaned files from files owned by another live DuckDB instance.

My reasoning was that temporary spill files are currently named only by their size class and file index, and TemporaryDirectoryHandle's existing destructor already removes duckdb_temp_* files without any cross-process ownership check. Based on that, I assumed temp_directory was intended to be owned by a single active DuckDB instance rather than shared.

If concurrent sharing of the same temp_directory is expected to be supported, then I agree this approach isn't sufficient. In that case we'd need some notion of ownership (e.g. per-process directories, ownership metadata, or another coordination mechanism) rather than deleting files purely by name.

Is sharing a temp_directory between multiple live DuckDB instances a supported use case?

dentiny: Yes, it's totally valid today.

Session-1:
```sql
DuckDB v0.0.1 (Unknown Version)
memory D SET temp_directory = '/tmp/duckdb-shared-spill';
memory D SET memory_limit = '64MB';
memory D SET threads = 1;
memory D CREATE TEMP TABLE spill_instance_1 AS
         SELECT
             i,
             repeat(md5(i::VARCHAR), 16) AS payload
         FROM range(1000000) t(i);
memory D CALL duckdb_temporary_files();
┌────────────────────────────────────────────────────────────┬───────────┐
│                            path                            │   size    │
│                          varchar                           │   int64   │
├────────────────────────────────────────────────────────────┼───────────┤
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_S32K-0.tmp    │  34111488 │
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_S64K-0.tmp    │   1245184 │
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_S96K-0.tmp    │     98304 │
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_DEFAULT-0.tmp │ 192675840 │
└────────────────────────────────────────────────────────────┴───────────┘
```
Keep the session alive and create another new session
```sql
DuckDB v0.0.1 (Unknown Version)
memory D SET temp_directory = '/tmp/duckdb-shared-spill';
memory D SET memory_limit = '64MB';
memory D SET threads = 1;
memory D CREATE TEMP TABLE spill_instance_2 AS
         SELECT
             i,
             repeat(md5(('instance-2-' || i)::VARCHAR), 16) AS payload
         FROM range(1000000) t(i);
memory D CALL duckdb_temporary_files();
┌────────────────────────────────────────────────────────────┬───────────┐
│                            path                            │   size    │
│                          varchar                           │   int64   │
├────────────────────────────────────────────────────────────┼───────────┤
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_S32K-0.tmp    │  32768000 │
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_S64K-0.tmp    │   1114112 │
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_S96K-0.tmp    │    196608 │
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_S128K-0.tmp   │    262144 │
│ /tmp/duckdb-shared-spill/duckdb_temp_storage_DEFAULT-0.tmp │ 203161600 │
└────────────────────────────────────────────────────────────┴───────────┘
```

dentiny: So IMO simply deleting all temp files is not a valid solution, you need to propose a more thorough design.

mattyy-k: Thanks for the detailed repro — that's convincing. I had assumed temp_directory was effectively single-owner, but your example clearly shows concurrent sharing is a supported use case today, so my current approach isn't safe.

One direction I was considering is using file locking instead of filename matching alone. DuckDB already has `FileLockType` (e.g. `READ_LOCK`/`WRITE_LOCK`) in the filesystem layer. Would it make sense to attempt acquiring an exclusive lock on each `duckdb_temp_*` file before deleting it, only removing files where the lock succeeds and skipping files that are still in use by another process?

Another idea that occurred to me is giving each active DuckDB instance its own subdirectory (or another form of per-instance ownership) under `temp_directory`, so cleanup becomes scoped to that instance instead of scanning a shared namespace. That feels simpler architecturally, but it would obviously be a larger behavioral change.

Or is there a different ownership/recovery mechanism you'd prefer for this? I'd rather align with the intended design than head off in the wrong direction.

dentiny: > One direction I was considering is using file locking instead of filename matching alone. 

I don't get it -- the core of the problem is not concurrent update / deletion on a single file, but how to tell whether a temp file belongs a dead database and could be deleted.

dentiny: > Another idea that occurred to me is giving each active DuckDB instance its own subdirectory (or another form of per-instance ownership) under temp_directory

No. As you can tell, users could set their own directory and it's doesn't make sense to prevent them from using the same directory.

mattyy-k: Thanks. I dug into the implementation a bit more after your comments, and I think I understand the ownership problem better now.

My original locking idea was incomplete because I was implicitly assuming lock state already reflected ownership, which it doesn't today. Temp files are currently opened without any `FileLockType`, so simply attempting to lock them during cleanup wouldn't distinguish a live file from an orphaned one.

Looking at `TemporaryFileHandle`, though, I noticed that each spill file already has a persistent `FileHandle`: `CreateFileIfNotExists()` opens it lazily on first write, and the handle is kept alive until `DeleteIfEmpty()`, where it's closed and the file is removed. There isn't a close/reopen cycle while the file exists.

That made me wonder whether ownership could instead be represented by holding a `WRITE_LOCK` on that existing long-lived file handle for the lifetime of the spill file. After an abnormal termination the OS would release the advisory lock automatically, so a new instance could distinguish live files from orphaned ones by attempting to acquire the lock before deleting.

Before I prototype that, though, I wanted to ask: is there a reason spill files were intentionally left unlocked (for example portability or filesystem semantics), or does this seem like a direction worth exploring?

mattyy-k: @dentiny  Just following up in case this got buried. No rush—I just wanted to check whether you had any thoughts on the ownership discussion whenever you get a chance. Happy to iterate on whichever direction makes the most sense.

PR #24600, opened 2026-08-07
Propagate stats in materialized CTEs and array distance/similarity functions

Maxxen:
When looking at the plans produced by the new `NEAREST BY` join (introduced in #24137) I realized that since the `<ranking expression> IS NOT NULL` is correlated, decorrelation lifts it into a ANY join condition, causing to be evaluated twice, both as part of a `BLOCKWISE_NL_JOIN`, and as part of the actual `TOP-K` further up in the plan. 

Since the ranking expressions in both vss and spatial workloads can be quite expensive to compute, this seems not ideal, especially considering that there are no nulls in either table. 

Fixing this required two changes:

1. Propagating the validity statistics in the array distance functions. This works by just combining the validity of the input arguments, as these functions don't return `NULL` on their own, but rather throw on invalid input (high-five past max for making that decision ✋). In theory we could also derive min/max bounds for some of the variants, but I don't think that would be very useful in practice and I know float-stats propagation can be prickly due to non-commutativity of floating-point ops so I left it out of this PR.

I had hoped this would let the optimizer just remove the filter entirely. However, I also had to add support to...

2. Propagate stats across materialized CTE's (which is what the delim-join now turns into). Im not super confident working with CTE's yet, but this seems like a pretty straightforward and safe thing to do, as long as the CTE is materialized and non-recursive? Would love some feedback of `src/optimizer/statistics/operator/propagate_cte.cpp` from @kryonix! As a bonus this also triggers other stats-based optimizations when scanning materialized CTE's, e.g. compressed-materialization.

Adding 2. also exposed a pre-existing issue with the `TopNWindowElimination`, which read the final `statistics_map` to set `group_stats` on the aggregate it substitutes for a top-N window. Those stats are narrowed by operators above the window, but the aggregate sits below them, so the perfect hash aggregate gets sized too small. This reproducible on main even when no CTEs are involved:

```sql
CREATE TABLE w(g INT, v INT);
INSERT INTO w SELECT (range*7)%40, range FROM range(400);
CREATE TABLE s(k INT); INSERT INTO s VALUES (0),(1),(2),(3);

SELECT count(*) FROM (SELECT g, v, row_number() OVER (PARTITION BY g ORDER BY v) rn FROM w) t
  JOIN s ON t.g = s.k WHERE t.rn = 1;
----
Invalid Input Error: Perfect hash aggregate: aggregate group 8 exceeded total groups 8
```

The join narrows g to [0,3] while the aggregate below still sees all 40 partitions. The fix is to use the `window_expr.PartitionsStats()` instead, which is contain the stats for the aggregates actual input.

There are a lot of other open PRs making various adjustments to the `TopNWindowElimination` code, e.g. https://github.com/duckdb/duckdb/pull/24551, https://github.com/duckdb/duckdb/pull/24346, https://github.com/duckdb/duckdb/pull/23943, https://github.com/duckdb/duckdb/pull/23702. I'm not sure if the fix in this PR overlaps or can be generalized, but it does seem like a sane fix in isolation.

---------------------

With these changes, the join condition now folds into a `BlockwiseNLJoin: true` - basically a cross product - eliminating e.g. a duplicate `array_cosine_similarity` evaluation per row pair. Here's a quick benchmark result:

```
128 users × 16,384 products (2.1M pairs), APPROX NEAREST 5 BY SIMILARITY, threads=4.

┌───────────────┬─────────┬─────────┬────────────────┐
│ Embedding dim │ Before  │  After  │    Speedup     │
├───────────────┼─────────┼─────────┼────────────────┤
│ 256           │ 0.851 s │ 0.529 s │ 1.61x (−37.8%) │
├───────────────┼─────────┼─────────┼────────────────┤
│ 16            │ 0.046 s │ 0.035 s │   1.31x (−24%) │
└───────────────┴─────────┴─────────┴────────────────┘
```

I want to investigate folding `BLOCKWISE_NL_JOIN`s on `true` into cross-products next, to hopefully shave of some minor execution overhead, and make it even easier for optimizers to target this plan shape. We should also be able to do the opposite and turn e.g. inner `BLOCKWISE_NL_JOIN` on `false` into `EMPTY_RESULT`.

### DIFF
--- a/extension/core_functions/scalar/array/array_functions.cpp
+++ b/extension/core_functions/scalar/array/array_functions.cpp
@@ -2,6 +2,7 @@
 #include "core_functions/scalar/array_functions.hpp"
 #include "core_functions/array_kernels.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/storage/statistics/array_stats.hpp"
 
 namespace duckdb {
 
@@ -208,7 +209,23 @@ static auto ArrayGenericFoldStats(ClientContext &context, FunctionStatisticsInpu
 	const auto &lhs_stats = input.child_stats[0];
 	const auto &rhs_stats = input.child_stats[1];
 	auto new_stats = NumericStats::CreateUnknown(input.expr.GetReturnType());
+
+	auto &lhs_child_stats = ArrayStats::GetChildStats(lhs_stats);
+	auto &rhs_child_stats = ArrayStats::GetChildStats(rhs_stats);
+
+	const auto has_any_nulls = lhs_child_stats.CanHaveNull() || rhs_child_stats.CanHaveNull();
+
+	if (has_any_nulls) {
+		// We will throw an error if the child arrays have nulls, so don't propagate any stats
+		return new_stats.ToUnique();
+	}
+
+	// If the child has no nulls, we won't throw.
+	input.expr.FunctionMutable().GetProperties().SetErrorMode(FunctionErrors::CANNOT_ERROR);
+
+	// Forward the validity
 	new_stats.CombineValidity(lhs_stats, rhs_stats);
+
 	return new_stats.ToUnique();
 }
 
--- a/extension/core_functions/scalar/array/array_functions.cpp
+++ b/extension/core_functions/scalar/array/array_functions.cpp
@@ -202,6 +202,16 @@ static void ArrayGenericFold(DataChunk &args, ExpressionState &state, Vector &re
 	}
 }
 
+static auto ArrayGenericFoldStats(ClientContext &context, FunctionStatisticsInput &input)
+    -> unique_ptr<BaseStatistics> {
+	// Propagate validity
+	const auto &lhs_stats = input.child_stats[0];
+	const auto &rhs_stats = input.child_stats[1];
+	auto new_stats = NumericStats::CreateUnknown(input.expr.GetReturnType());
+	new_stats.CombineValidity(lhs_stats, rhs_stats);
+	return new_stats.ToUnique();
+}
+
 //------------------------------------------------------------------------------
 // Function Registration
 //------------------------------------------------------------------------------
@@ -213,11 +223,13 @@ template <class OP>
 static void AddArrayFoldFunction(ScalarFunctionSet &set, const LogicalType &type) {
 	const auto array = LogicalType::ARRAY(type, optional_idx());
 	if (type.id() == LogicalTypeId::FLOAT) {
-		ScalarFunction function({array, array}, type, ArrayGenericFold<float, OP>, ArrayGenericBinaryBind);
+		ScalarFunction function({array, array}, type, ArrayGenericFold<float, OP>, ArrayGenericBinaryBind,
+		                        ArrayGenericFoldStats);
 		function.SetFallible();
 		set.AddFunction(function);
 	} else if (type.id() == LogicalTypeId::DOUBLE) {
-		ScalarFunction function({array, array}, type, ArrayGenericFold<double, OP>, ArrayGenericBinaryBind);
+		ScalarFunction function({array, array}, type, ArrayGenericFold<double, OP>, ArrayGenericBinaryBind,
+		                        ArrayGenericFoldStats);
 		function.SetFallible();
 		set.AddFunction(function);
 	} else {
--- a/src/include/duckdb/optimizer/statistics_propagator.hpp
+++ b/src/include/duckdb/optimizer/statistics_propagator.hpp
@@ -135,8 +135,15 @@ private:
 	optional_ptr<LogicalOperator> root;
 	//! The map of ColumnBinding -> statistics for the various nodes
 	column_binding_map_t<unique_ptr<BaseStatistics>> statistics_map;
-	//! The map of CTE index -> statistics of the columns emitted by the CTE definition
-	unordered_map<TableIndex, vector<unique_ptr<BaseStatistics>>> cte_stats_map;
+	//! The statistics of a materialized CTE definition, which hold for every reference to it
+	struct CTEStatistics {
+		//! Statistics of the columns emitted by the definition, by position
+		vector<unique_ptr<BaseStatistics>> column_stats;
+		//! Cardinality of the definition
+		unique_ptr<NodeStatistics> node_stats;
+	};
+	//! The map of CTE index -> statistics of its definition
+	unordered_map<TableIndex, CTEStatistics> cte_stats_map;
 	//! Node stats for the current node
 	unique_ptr<NodeStatistics> node_stats;
 };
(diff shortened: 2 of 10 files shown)


kryonix:
So, I tried to add statistics for CTEs, too, while working on the pipeline exchange (https://github.com/duckdb/duckdb/pull/23750). The revert commit is here: https://github.com/duckdb/duckdb/pull/23750/changes/f34651364881d0a042274aa9aae840a99d92ba6c. Maybe you can compare what you added, and what I have added. It looks similar. However, the problem I ran into was that compressed materialization introduced a bunch of regressions for directly fed pipelines through the exchange. This was because that adds a bunch of unnecessary overhead. So, that is something to look into. CI does not complain, but TPC-H and TPC-DS show a small regression. But if we can either verify that this is noise, and show the positive effect of the change, I'm all for it. I think long-term we absolutely want to have it.

Maxxen:
@kryonix Nice! It looks like the main difference is that your PR also added the NodeStatistics (i.e. cardinality) for the whole CTE ref while I just propagate the individual columns. I've adapted this PR to include your changes too, but indeed that does seem to cause a slight 1-2% regression on TPC-H SF=1, where almost a third seems to be spent in the optimizer (the regression basically vanishes on SF=10). 

It does seem like a pretty small price to pay all things considered, especially cause I would assume the downstream optimizations (like CM) are more efficient on larger inputs. But I'm investigating whether conditionally applying CM based on the expected cardinality (i.e. don't do it on known low cardinality inputs) could be a band-aid to avoid regressing in this particular case.

Mytherin:
Thanks for the PR!

Changes look good to me - but separately I wonder if the expression getting duplicated is a separate issue and this is mostly working around that duplication by optimizing out `<ranking expression> IS NOT NULL` for columns that don't have NULL values? i.e. won't this still cause duplicate execution for columns that have `NULL` values for which `IS NOT NULL` cannot be optimized out? I also wonder if this expression duplication is correct when using volatile expressions (e.g. `nextval('seq')` etc).

@kryonix could you do a final pass? Then we can merge this.

extension/core_functions/scalar/array/array_functions.cpp
```diff
@@ -202,6 +202,16 @@ static void ArrayGenericFold(DataChunk &args, ExpressionState &state, Vector &re
 	}
 }
 
+static auto ArrayGenericFoldStats(ClientContext &context, FunctionStatisticsInput &input)
+    -> unique_ptr<BaseStatistics> {
+	// Propagate validity
+	const auto &lhs_stats = input.child_stats[0];
+	const auto &rhs_stats = input.child_stats[1];
+	auto new_stats = NumericStats::CreateUnknown(input.expr.GetReturnType());
+	new_stats.CombineValidity(lhs_stats, rhs_stats);
```
kryonix, line 227: `CombineValidity` only considers the validity of the outer arrays, but every fold registered with this callback is fallible and throws when an array contains a NULL element. With `v FLOAT[2] NOT NULL` containing `[1, NULL]`, the merge-base raises for `WHERE array_distance(v, [0,0]::FLOAT[2]) IS NOT NULL`, while this branch folds the predicate to `true` and returns the row without evaluating the function. This affects all five array fold functions. I think we should only advertise the no-NULL result when required evaluation is preserved or inner element statistics prove the arrays safe, and add coverage for both the error case and the valid-array optimization.

```sql
statement ok
CREATE TABLE valid_stats_arrays(v FLOAT[2] NOT NULL);

statement ok
INSERT INTO valid_stats_arrays VALUES ([1, 2]);

statement ok
PRAGMA explain_output = OPTIMIZED_ONLY

# Preserve the intended optimization for fully valid arrays
query II
EXPLAIN SELECT count(*) FROM valid_stats_arrays
WHERE array_distance(v, [0, 0]::FLOAT[2]) IS NOT NULL
----
logical_opt	<!REGEX>:.*IS NOT NULL.*

statement ok
CREATE TABLE invalid_stats_arrays(v FLOAT[2] NOT NULL);

statement ok
INSERT INTO invalid_stats_arrays VALUES ([1, NULL]);

# Do not optimize away the fallible evaluation
statement error
SELECT count(*) FROM invalid_stats_arrays
WHERE array_distance(v, [0, 0]::FLOAT[2]) IS NOT NULL
----
left argument can not contain NULL values
```

kryonix (commented):
`ArrayGenericFoldStats` derives result validity only from the outer arrays. However, all five functions using this callback are fallible and explicitly throw when a non-null array contains a NULL element.

When statistics prove the outer array non-null, the optimizer replaces `array_distance(...) IS NOT NULL` with `true`, removing the only function evaluation and suppressing the documented runtime error.

Minimal reproducer:

```sql
CREATE TABLE a(v FLOAT[2] NOT NULL);
INSERT INTO a VALUES ([1, NULL]);

SELECT count(*)
FROM a
WHERE array_distance(v, [0, 0]::FLOAT[2]) IS NOT NULL;
```

- `main`: `Invalid Input Error: ... left argument can not contain NULL values`
- PR head: returns `1`
- `PRAGMA disable_optimizer` on the PR: raises the expected error
- Directly selecting `array_distance(...)`: raises the expected error

The same silent-success behavior occurs for all functions registered through `AddArrayFoldFunction`:

- `array_distance`
- `array_inner_product`
- `array_negative_inner_product`
- `array_cosine_similarity`
- `array_cosine_distance`

The violated invariant is that statistics must not allow the optimizer to remove the sole evaluation of a fallible expression unless its complete input domain is known safe. It first becomes false in the scalar statistics callback, where outer validity is treated as sufficient despite inner element validity controlling whether evaluation throws.

Maxxen:
Ok, but the fallibility is a separate issue - like mark said, the expression should maybe not be duplicated at all when performing the CTE-based decorrelation?

Maxxen:
Like, I can add a check for `IsFallible()` in the statistics propagator to to check before removing a `IS NOT NULL`, I guess that does fix the soundness issue in general. 

But then we're back to my original problem: the ranking expression is evaluated twice.

As a side-note, it kinda feels like "can throw" should be part of the base statistics so that it can be propagated dynamically.

kryonix:
You are right, the expression duplication is pre-existing and requires a structural fix: compute the ranking once, project it as a hidden value, then reuse it for both NULL filtering and ordering. That is broader and riskier than this PR's statistics work. We should do that as a separate PR instead.

extension/core_functions/scalar/array/array_functions.cpp
Maxxen: Fixed! We now inspect the child stats too before we propagate our own validity stats

Maxxen:
So I've fixed the unsoundness issue raised by @kryonix, and talked to @lnkuiper about how to mitigate the compressed-materialization regression by introducing unchecked casts. However, I would propose that we merge this PR as is now, and do the CM-optimizations as a follow-up as I imagine that will be a bit more substantial change, and we still want to have support for CTE statistics for other optimizations anyway (and this PR also fixes a top-n-elimination bug).

pdet:
@kryonix is this good to go?

kryonix:
Looks good to me!

PR #24601, opened 2026-08-07
Fix grouped printf formats for signed integer arguments

AbhinavPandey01:
## Summary

Grouped float-family `printf` formats reinterpret negative signed integers as unsigned values before DuckDB applies its integer-preserving formatting path. For example:

```sql
SELECT printf('%,.0f', -5::BIGINT);
-- before: 18,446,744,073,709,551,611
-- after:  -5
```

Preserve the source integer's signedness for `a`, `e`, `f`, and `g` conversions. Unsigned values retain the existing unsigned path, including exact formatting above the double-precision limit.

## Tests

Added SQLLogicTest coverage for:

- signed `TINYINT`, `SMALLINT`, `INTEGER`, `BIGINT`, and `HUGEINT`
- all grouped float-family conversion types
- `BIGINT` and `UBIGINT` values above `2^53`
- multiple converted arguments in one call

Built DuckDB from current `main` and ran:

```text
build/upstream-review/test/unittest test/sql/function/string/test_format_extensions.test
All tests passed (43 assertions in 1 test case)
```

### DIFF
--- a/test/sql/function/string/test_format_extensions.test
+++ b/test/sql/function/string/test_format_extensions.test
@@ -97,6 +97,30 @@ select printf('%,.3f', 123456.789::DOUBLE)
 ----
 123,456.789
 
+# preserve the sign when grouped float-family formats receive signed integers
+query IIIII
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24606, opened 2026-08-08
Fix `DECIMAL` syntax silently accepting non-integral parameters

### LINKED ISSUE #24570: DECIMAL syntax accepts non-integer parameters
### What happens?

DuckDB appears to parse and accept `DECIMAL` syntax that specifies the precision and scale type parameters as a non-integer number.

The resulting type is as if the type parameters are rounded to the closest integer which is probably correct if accepting a non-integer would be considered ok. 

That said, I really think any non-integer constant should be rejected for the type parameter. 
I mean, I realize this is user error, but I just defined a table with a decimal column as

`DECIMAL(3.2)`

Stupid, I know, but on my keyboard the `,` and the `.` are next to each other and it's not very hard to make that exact typo. 

Anyway, it took me, well, longer than I'd like to figure out why my decimals were always being truncated. 

### To Reproduce

```sql
CREATE TABLE t(d DECIMAL(3.2));

INSERT INTO t VALUES (1.2), (1.6);

SELECT * FROM t;

┌──────────────┐
│      d       │
│ decimal(3,0) │
├──────────────┤
│            1 │
│            2 │
└──────────────┘
```

### OS:

windows 11

### DuckDB Version:

1.5.5

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Roland Bouman

### Affiliation:

EPAM Systems B.V.

- [x] Yes, I have

- [x] Yes, I have

Yes

tempewda:
Fixes #24570.

Previously, `BindDecimalType` only validated metadata parameters using a permissive `DefaultTryCastAs(LogicalTypeId::UTINYINT)` check. This meant DuckDB would silently accept and cast non-integral types like `DECIMAL(3.2)`, effectively rounding the float down to `3` instead of rejecting the user's typo.

Taking direct inspiration from how `ARRAY` enforces its size modifiers in the exact same file (`BindArrayType`), this PR adds a strict `!value.type().IsIntegral()` check before the cast attempt to ensure only pure integers are allowed through.

So currently only actual DuckDB integer types (e.g., `DECIMAL(3, 2)`) are accepted.
Floats (`3.2`), whole floats (`3.0`), string literals (`'3'`), and booleans (`TRUE`) are now immediately caught and rejected with a `Binder Error: DECIMAL type width/scale must be an integral type`.

### DIFF
--- a/src/catalog/default/default_types.cpp
+++ b/src/catalog/default/default_types.cpp
@@ -31,6 +31,9 @@ LogicalType BindDecimalType(BindLogicalTypeInput &input) {
 		if (width_value.IsNull()) {
 			throw BinderException("DECIMAL type width cannot be NULL");
 		}
+		if (!width_value.type().IsIntegral()) {
+			throw BinderException("DECIMAL type width must be an integral type");
+		}
 		if (width_value.DefaultTryCastAs(LogicalTypeId::UTINYINT)) {
 			width = width_value.GetValueUnsafe<uint8_t>();
 			scale = 0; // reset scale to 0 if only width is provided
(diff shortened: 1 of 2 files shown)


Maxxen (approved):
Thanks, LGTM!

PR #24607, opened 2026-08-08
Fix COUNT(x) still reading the counted column after the COUNT(*) rewrite

### LINKED ISSUE #24593: COUNT(x) rewritten to COUNT(*) doesn't prune x from the scan
### What happens?

Observed when launching a SQL query with `EXPLAIN` in DuckLake on a large dataset (~20B rows): when statistics prove a column cannot be `NULL`, DuckDB rewrites `COUNT(x)` into `count_star()`. That rewrite happens during statistics propagation, which runs after the unused column pruning pass, so nothing revisits the scan afterward.

Of course, the performance cost depends on how wide the discarded `x` column is.
In this case, the `uuid` was `x`, and was the largest column in the table.

On a DuckLake table of 5M rows built from the same shape, `parquet_metadata` reports the counted column as `99.9%` of the stored bytes:
| column | compressed bytes | pct |
| --- | ---: | ---: |
| `uuid` | 161,022,873 | 99.9 |
| `device_type` | 121,982 | 0.1 |
| `app` | 93,426 | 0.1 |

The plan scans a column that no expression in the plan references.

In a table containing `app`, `device_type`, and `uuid` columns, `COUNT(uuid)` still pays to read `uuid`.
Image

It's not the case when we `COUNT(*)`:
Image

The consequence is that the scan reads a column the aggregate no longer references. On a persistent table of 5M rows, comparing the two queries on the same binary:

| query | user CPU, five runs |
| --- | --- |
| `count(uuid)` | 0.10s  0.09s  0.10s  0.10s  0.10s |
| `count(*)` | 0.03s  0.03s  0.03s  0.03s  0.03s |

Both compute `count_star()` over the same rows with the same filter and grouping, so the entire difference is reading a column that is not referenced. And the effect grows with a large DuckLake table.

### To Reproduce

From DuckDB CLI, for displaying the explain from the screenshots:

```sql
CREATE TABLE events(app INTEGER, device_type VARCHAR, uuid VARCHAR NOT NULL);
INSERT INTO events SELECT (i%5), 'd'||(i%8), md5(i::VARCHAR) FROM range(100000) t(i);

EXPLAIN SELECT device_type, count(uuid) FROM events WHERE app < 4 GROUP BY device_type;
```
The aggregate is rewritten to `count_star()`, but the scan still projects `uuid`.

Then, the equivalent `COUNT(*)` query produces the same aggregate and reads `device_type` only:
```sql
EXPLAIN SELECT device_type, count(*) FROM events WHERE app < 4 GROUP BY device_type;
```

I expected to see `uuid` not appear in the scan projection, since no expression in the plan references it.

### OS:

macOS 26.6.1 arm64

### DuckDB Version:

v1.5.5

### DuckDB Client:

CLI

### Hardware:

Apple M4 Pro

### Full Name:

Florian Valeye

### Affiliation:

altertable-ai

- [x] Yes, I have

- [x] Yes, I have

fvaleye:
Fixes #24593.

When statistics prove a column cannot be `NULL`, DuckDB rewrites `count(x)` into `count_star()`. The aggregate no longer needs the column, but the scan keeps reading it. 
The same query written as `count(*)` reads one column fewer. On a wide string column and with large tables, that difference could be most of the query.

It comes down to ordering: column pruning runs before statistics propagation, so it decides what to read while the plan still says `count(x)`.

Reproduce with:
```sql
CREATE TABLE events(app INTEGER, device_type VARCHAR, uuid VARCHAR NOT NULL);
INSERT INTO events SELECT (i%5), 'd'||(i%8), md5(i::VARCHAR) FROM range(5000000) t(i);
EXPLAIN SELECT device_type, count(uuid) FROM events WHERE app < 4 GROUP BY device_type;
```

## The fix
Statistics propagation now records when a callback drops an aggregate's arguments, and column pruning runs again when it did. Recording it at the expression level covers any aggregate that drops arguments, rather than matching `count` by name, thereby removing the separate `count` matcher in `PropagateStatistics`. Queries where nothing was rewritten are untouched.

The rerun clears projection maps first. Joins, filters, and orders address their input the pruner renumbers those positions, and the existing remapping only notices a column disappearing, not moving. `COLUMN_LIFETIME` rebuilds them.

Common subplan materialization runs before statistics propagation, so the rerun is the first time the pruner meets those CTEs, and it reached two pre-existing bugs there. A materialized CTE keeps one "everything is referenced" flag shared by all readers, but it was overwritten rather than accumulated, so a reader contributing no references could clear it and drop columns another reader still needed. It is now only ever set. A CTE whose own output exposes its readers' bindings is no longer pruned either, since the rewrite only covers the RHS and operators above the CTE would keep stale references.

## Tests
* `statistics_count_not_null_pruning.test`: the rewrite and the pruning, `FILTER` over the counted column and over another, `ORDER BY` inside the aggregate, `count(DISTINCT)` which must not be rewritten, a materialized CTE read twice, and non-nullness that comes from a pushed-down filter rather than the column definition. Every result is re-asserted with `statistics_propagation`, `unused_columns` and `column_lifetime` disabled in turn. It also covers the CTE flag: identical union branches that carry a scalar subquery `count` become a single common subplan CTE, and without the fix, that query fails to bind.
* `statistics_count_not_null_join_pruning.test`: the same pruning across inner and left joins, a semi join, and a cross product, under `debug_verify_column_bindings`. One boundary is pinned rather than left implicit: partial aggregate pushdown runs before statistics propagation, so when it splits the aggregate, the partial `count` below the join keeps the column. 
* `common_subplan_nested_cte.test`: its existing nested common subplan query, now with and without `column_lifetime`, under `debug_verify_column_bindings`. Disabling that pass is important because it rewrites the stale bindings afterward and masks the defect; without the fix, the query fails to bind to `interested_id`.
* `compressed_materialization.test_slow` counted a column it also expects to see in the scan projection, so it was relying on the behavior being fixed here. Those two aggregates become `sum`, which genuinely reads its argument. 
* `test_cte_hash_join_exchange.test`: its `STREAMING_FANOUT` expectation relied on the pruning bug keeping the unread `payload` column wide enough to exceed the materialization budget: the `EXPLAIN ANALYZE` now asserts the executed sibling query, which reads `payload`, since pruning unused materialized CTE columns intentionally narrows the width the fanout cost model sees.

## Benchmark
It's starting to be observable around 5M, but the use case is much worse with a table around 20B.
`benchmark/micro/optimizer/count_not_null_column_prune.benchmark`, registered in `.github/regression/micro.csv`: 5M rows, a 32-character `uuid`, persistent storage. 

Running exactly that on the base commit and on this branch:
```
base         Projections: device_type, uuid     0.12 / 0.11 / 0.12 s
this branch  Projections: device_type           0.02 / 0.02 / 0.03 s
```

### DIFF
--- a/.github/regression/micro.csv
+++ b/.github/regression/micro.csv
@@ -34,5 +34,6 @@ benchmark/micro/logger/storage/file/log_message_size/huge_string.benchmark
 benchmark/micro/logger/storage/file/log_message_size/small_string.benchmark
 benchmark/micro/filter/choose_correct_filter_function.benchmark
 benchmark/micro/optimizer/topn_window_elimination.benchmark
+benchmark/micro/optimizer/count_not_null_column_prune.benchmark
 benchmark/micro/aggregate/group_two_string_dictionaries.benchmark
 benchmark/micro/functions/nested_macros.benchmark
--- a/benchmark/micro/optimizer/count_not_null_column_prune.benchmark
+++ b/benchmark/micro/optimizer/count_not_null_column_prune.benchmark
@@ -1,0 +1,27 @@
+# name: benchmark/micro/optimizer/count_not_null_column_prune.benchmark
+# description: Benchmark of column pruning after the COUNT(x) to COUNT(*) rewrite
+# group: [optimizer]
+
+name Count Not Null Column Prune
+group micro
+subgroup optimizer
+storage persistent
+
+load
+CREATE TABLE events AS
+    SELECT 'd' || (i % 8) AS device_type,
+           md5(i::VARCHAR) AS uuid
+    FROM range(5000000) t(i);
+
+run
+SELECT device_type, count(uuid) FROM events GROUP BY device_type ORDER BY device_type;
+
+result II
+d0	625000
+d1	625000
+d2	625000
+d3	625000
+d4	625000
+d5	625000
+d6	625000
+d7	625000
--- a/src/include/duckdb/optimizer/statistics_propagator.hpp
+++ b/src/include/duckdb/optimizer/statistics_propagator.hpp
@@ -42,6 +42,9 @@ public:
 	bool FilterBindingsChanged() const {
 		return filter_bindings_changed;
 	}
+	bool HasRemovedAggregateChildren() const {
+		return removed_aggregate_children;
+	}
 
 	//! Derive output statistics of a monotone function by evaluating it at the corners of its
 	//! argument ranges (see ArgProperties). Returns nullptr when the bounds cannot be derived.
@@ -155,6 +158,8 @@ private:
 	};
 	//! The map of CTE index -> statistics of its definition
 	unordered_map<TableIndex, CTEStatistics> cte_stats_map;
+	//! Whether a statistics callback removed aggregate children, leaving columns that may now be unused
+	bool removed_aggregate_children = false;
 	//! Node stats for the current node
 	unique_ptr<NodeStatistics> node_stats;
 	//! Whether statistics changed which relations a filter depends on
--- a/src/optimizer/optimizer.cpp
+++ b/src/optimizer/optimizer.cpp
@@ -151,6 +151,17 @@ void Optimizer::Verify(LogicalOperator &op) {
 	ColumnBindingResolver::Verify(context, op);
 }
 
+// RemoveUnusedColumns renumbers bindings, so RemapProjectionMap can match a stale binding to a
+// different column rather than dropping it
+static void ClearProjectionMaps(LogicalOperator &op) {
+	for (idx_t child_index = 0; child_index < op.children.size(); child_index++) {
+		if (auto projection_map = LogicalOperatorVisitor::GetProjectionMap(op, child_index)) {
+			projection_map->clear();
+		}
+		ClearProjectionMaps(*op.children[child_index]);
+	}
+}
+
 static bool ContainsDML(const LogicalOperator &op) {
 	switch (op.type) {
 	case LogicalOperatorType::LOGICAL_INSERT:
@@ -434,12 +445,14 @@ void Optimizer::RunBuiltInOptimizers() {
 	// DML CTEs can invalidate the table statistics captured during planning.
 	column_binding_map_t<unique_ptr<BaseStatistics>> statistics_map;
 	bool propagated_statistics = false;
+	bool removed_aggregate_children = false;
 	if (!CTEContainsDML(*plan)) {
 		RunOptimizer(OptimizerType::STATISTICS_PROPAGATION, [&]() {
 			StatisticsPropagator propagator(*this, *plan);
 			propagator.PropagateStatistics(plan);
 			statistics_map = propagator.GetStatisticsMap();
 			propagated_statistics = true;
+			removed_aggregate_children = propagator.HasRemovedAggregateChildren();
 		});
 	}
 	if (propagated_statistics) {
@@ -449,6 +462,7 @@ void Optimizer::RunBuiltInOptimizers() {
 			StatisticsPropagator propagator(*this, *plan);
 			propagator.PropagateStatistics(plan);
 			statistics_map = propagator.GetStatisticsMap();
+			removed_aggregate_children |= propagator.HasRemovedAggregateChildren();
 		}
 	}
 
@@ -463,6 +477,15 @@ void Optimizer::RunBuiltInOptimizers() {
 		CommonAggregateOptimizer common_aggregate;
 		common_aggregate.VisitOperator(*plan);
 	});
+
+	// COUNT(x) becomes COUNT(*) during statistics propagation, leaving unreferenced columns in the scan
+	if (removed_aggregate_children) {
+		RunOptimizer(OptimizerType::UNUSED_COLUMNS, [&]() {
+			ClearProjectionMaps(*plan);
+			RemoveUnusedColumns unused(*this);
+			unused.VisitOperator(plan);
+		});
+	}
 
 	// creates projection maps so unused columns are projected out early
 	RunOptimizer(OptimizerType::COLUMN_LIFETIME, [&]() {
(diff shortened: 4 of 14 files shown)


kryonix (changes requested):
Thanks for the PR! The core `COUNT(x) → COUNT(*)` fix seems sound, but I found one performance regression in the accompanying CTE change. It recreates exactly the class of unnecessary wide-column reads that [PR #24607](https://github.com/duckdb/duckdb/pull/24607) is intended to eliminate.

### [P2] Reader occurrence counts can permanently disable CTE column pruning

Location: [`src/optimizer/remove_unused_columns.cpp:676`](https://github.com/duckdb/duckdb/blob/cde97b1d75b7a2614f0cdfddc098e3e32cafb60c/src/optimizer/remove_unused_columns.cpp#L676)

`referenced_columns` is keyed by each CTE reader's distinct `table_index`, so its size counts reference occurrences across readers, not distinct CTE output columns. For a two-column CTE read three times with every reader selecting only column 0, the second reader makes `referenced_columns.size() == 2` and this change permanently sets `everything_referenced`; the PR head then scans both columns, whereas the base prunes column 1. Please keep the all-columns-reader state separate and determine explicit full-schema coverage from distinct `column_index` values after gathering all readers.

For a two-column CTE read three times, with every reader selecting only `a`:

1. The first reader adds `(reader1, a)`: size 1.
2. The second adds `(reader2, a)`: size 2, which equals the CTE width and permanently sets `everything_referenced`.
3. The third adds `(reader3, a)`, but the new monotonic behavior prevents the false positive from being corrected.

Minimal reproducer:

```sql
CREATE TABLE cte_fanout(source_key INTEGER, unused_payload VARCHAR);
INSERT INTO cte_fanout VALUES (1, 'wide'), (2, 'column');

EXPLAIN
WITH c AS MATERIALIZED (SELECT * FROM cte_fanout)
SELECT sum(source_key)
FROM (
    SELECT source_key FROM c
    UNION ALL
    SELECT source_key FROM c
    UNION ALL
    SELECT source_key FROM c
);
```

The root fix should track an all-columns reader independently from explicit references. In particular, `everything_referenced` should become monotonic based on the current reader's `everything_referenced` barrier, while full explicit coverage should be determined from the distinct `column_index` values after all readers have been visited. Comparing schema width with a map keyed by reader bindings is not valid.

Suggested regression test for `test/optimizer/unused_columns/cte_unused_columns.test`:

```sql
statement ok
CREATE TABLE cte_fanout(source_key INTEGER, unused_payload VARCHAR);

statement ok
INSERT INTO cte_fanout VALUES (1, 'wide'), (2, 'column');

query II
EXPLAIN
WITH c AS MATERIALIZED (SELECT * FROM cte_fanout)
SELECT sum(source_key)
FROM (
    SELECT source_key FROM c
    UNION ALL
    SELECT source_key FROM c
    UNION ALL
    SELECT source_key FROM c
);
----
logical_opt	<!REGEX>:.*unused_payload.*

query I
WITH c AS MATERIALIZED (SELECT * FROM cte_fanout)
SELECT sum(source_key)
FROM (
    SELECT source_key FROM c
    UNION ALL
    SELECT source_key FROM c
    UNION ALL
    SELECT source_key FROM c
);
----
9
```

fvaleye:
> Thanks for the PR! The core `COUNT(x) → COUNT(*)` fix is sound, but I found one performance regression in the accompanying CTE change.

Thanks for the review @kryonix 🙏

> ### [P2] Reader occurrence counts can permanently disable CTE column pruning
>
> `referenced_columns` is keyed by each CTE reader's distinct `table_index`, so its size counts reference occurrences across readers, not distinct CTE output columns. [...] Please keep the all-columns-reader state separate and determine explicit full-schema coverage from distinct `column_index` values after gathering all readers.

Reproduced on my side. The comparison is `chunk_types.size()` against `column_references.size()`, but that map is keyed by (`table_index`, `column_index`) and each reader has its own `table_index`, so its size counts Reader × Column pairs, not columns. The three-reader reproducer demonstrated this.

> The root fix should track an all-columns reader independently from explicit references. In particular, `everything_referenced` should become monotonic based on the current reader's `everything_referenced` barrier, while full explicit coverage should be determined from the distinct `column_index` values after all readers have been visited.

I've added your CTE query as a regression test, with an additional two-reader variant. 
The fix keeps the two questions separate without adding per-reader state:
- `LOGICAL_CTE_REF` accumulates the shared monotonic all-columns barrier
- `LOGICAL_MATERIALIZED_CTE` decides full-schema coverage from `referenced_columns_in_rhs`, which already holds distinct `column_index` values

Fix pushed! 👍

_Worth sharing that the same check is already wrong on main. 
It escapes at three readers only because the assignment lets the last one overwrite the false positive. At two readers it does not: `SELECT source_key FROM c UNION ALL SELECT source_key FROM c` over the same two-column CTE still scans `unused_payload`. 
That case is covered by a test too._

kryonix (approved):
Thanks! LGTM now.

lnkuiper:
Looks like there will be merge conflicts in the merge queue, could you please look into resolving these once the queue has progressed?

fvaleye:
> Looks like there will be merge conflicts in the merge queue, could you please look into resolving these once the queue has progressed?

Sure!

fvaleye:
> Looks like there will be merge conflicts in the merge queue, could you please look into resolving these once the queue has progressed?

FYI, I rebased onto the latest main @lnkuiper. If a conflict does appear again as the queue progresses, I will resolve it right away!

fvaleye:
FYI @lnkuiper, I updated a failing test in the merge queue: `test_cte_hash_join_exchange.test`. 
It relied on the pruning bug this PR fixes.

fvaleye:
All good on my side @lnkuiper

lnkuiper:
Thanks! Queued up.

fvaleye:
> Thanks! Queued up.

Thanks @lnkuiper

PR #13123, opened 2024-07-23
Create file with O_EXCL flag set.

mkaruza:
When O_EXCL is used WITH O_CREAT open will fail if file exists. If flag `FILE_FLAGS_NULL_IF_EXISTS` is used opening will return nullptr instead of throwing error.

PR #24612, opened 2026-08-08
Honour FILE_FLAGS_EXCLUSIVE_CREATE on Windows

### LINKED ISSUE #24610: FILE_FLAGS_EXCLUSIVE_CREATE is silently ignored on Windows
### What happens?

On Windows, LocalFileSystem::OpenFile never consults FileOpenFlags::ExclusiveCreate(). 

ExclusiveCreate() is read in exactly one place in the tree. The POSIX branch of the same function, where it becomes O_EXCL. CREATE_NEW, the Win32 disposition that fails with ERROR_FILE_EXISTS, appears nowhere in src/common/local_file_system.cpp.

FILE_FLAGS_NULL_IF_EXISTS is dead on Windows for the same reason. It may only be set together with EXCLUSIVE_CREATE (FileOpenFlags::Verify), and since the create never fails it can never return null.

### To Reproduce

Add to test/common/test_file_system.cpp, then run unittest "[file_system]":
```


auto fs = FileSystem::CreateLocal();
auto fname = TestCreatePath("excl_repro");
auto flags = FileFlags::FILE_FLAGS_WRITE | FileFlags::FILE_FLAGS_FILE_CREATE |
             FileFlags::FILE_FLAGS_EXCLUSIVE_CREATE;

fs->OpenFile(fname, flags).reset();                    // creates the file
REQUIRE_THROWS(fs->OpenFile(fname, flags));            // FAILS on Windows: the open succeeds
REQUIRE(fs->OpenFile(fname, flags | FileFlags::FILE_FLAGS_NULL_IF_EXISTS) == nullptr);  // FAILS on Windows
```

Passes on Linux and macOS, fails on Windows.

### OS:

Windows

### DuckDB Version:

1.5.5

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Christoph Mettler

### Affiliation:

Ethenea

- [x] Yes, I have

- [x] Yes, I have

cmettler:
closes https://github.com/duckdb/duckdb/issues/24610
FileOpenFlags::ExclusiveCreate() is read in exactly one place in the tree: the POSIX branch of LocalFileSystem::OpenFile, where it maps to O_EXCL. The Windows branch never consults it, so the flag is silently dropped and the open falls through to OPEN_ALWAYS ("open, creating if absent"), which happily returns a handle to the already-existing file. CREATE_NEW -- the Win32 disposition that fails with ERROR_FILE_EXISTS, i.e. exactly this primitive -- appears nowhere in the file.

Consequence: a caller using the flag to create a file only if nobody else got there first silently races on Windows.

FILE_FLAGS_NULL_IF_EXISTS is dead on Windows for the same reason: it may only be set together with EXCLUSIVE_CREATE (see FileOpenFlags::Verify), and since the create never failed it could never return null. The error branch now mirrors the POSIX one, which already handles EEXIST.

Both flags arrived in #13123, which added them, added their combination rules to FileOpenFlags::Verify, and implemented them in the POSIX branch only; the Windows disposition block in the same function dates from 2021 and was not revisited. Neither flag has an in-tree caller, so no existing test could reach the Windows path -- they are used by extensions and over the C API, where DUCKDB_FILE_FLAG_CREATE_NEW has exposed the flag publicly since d4f7b546.

The new test is platform-agnostic: it passes on POSIX before and after this change, and fails on Windows without it.

### DIFF
--- a/src/common/local_file_system.cpp
+++ b/src/common/local_file_system.cpp
@@ -1280,7 +1280,13 @@ unique_ptr<FileHandle> LocalFileSystem::OpenFile(const string &path_p, FileOpenF
 	share_mode |= FILE_SHARE_DELETE;
 
 	if (open_write) {
-		if (flags.CreateFileIfNotExists()) {
+		// EXCLUSIVE_CREATE is tested FIRST: FileOpenFlags::Verify requires it to be combined with
+		// FILE_CREATE/FILE_CREATE_NEW, so every legal caller also sets one of those and testing them first
+		// would swallow it. CREATE_NEW fails with ERROR_FILE_EXISTS instead of opening the existing file,
+		// which is what O_EXCL gives the POSIX branch above.
+		if (flags.ExclusiveCreate()) {
+			creation_disposition = CREATE_NEW;
+		} else if (flags.CreateFileIfNotExists()) {
 			creation_disposition = OPEN_ALWAYS;
 		} else if (flags.OverwriteExistingFile()) {
 			creation_disposition = CREATE_ALWAYS;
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - outside of a unittest is there a scenario in the code that triggers this behavior? Can we add such a test instead?

cmettler:
> Thanks - outside of a unittest is there a scenario in the code that triggers this behavior? Can we add such a test instead?

EXCLUSIVE_CREATE genuinely has no in-tree caller that's why no existing test can reach the Windows path. I found it building an experimental Delta extension that tested put-if-absent on the windows local filesystem for log commits.

The closest consumer I can find is duckdb/unity_catalog: CopyStagedCommitToDeltaLog opens the destination with EXCLUSIVE_CREATE | NULL_IF_EXISTS . I haven't checked whether that path ever resolves to LocalFileSystem on Windows in practice so I'd rather not claim it as a live bug.

PR #24614, opened 2026-08-08
Fix CREATE SEQUENCE throwing INTERNAL error on NULL option values

### LINKED ISSUE #24602: CREATE SEQUENCE with a NULL option value throws INTERNAL Error: Calling GetValue on a value that is NULL
### What happens?

`TransformCreateSequence` in `src/parser/peg/transformer/transform_create_sequence.cpp` reads every sequence option by calling `Value::GetValue<T>()` on the option's parsed constant without first checking whether that constant is NULL:

- line 48 — `info->increment = seq_val_option->value.GetValue<int64_t>();`
- line 71 — `info->min_value = seq_val_option->value.GetValue<int64_t>();`
- line 78 — `info->max_value = seq_val_option->value.GetValue<int64_t>();`
- line 82 — `info->start_value = seq_val_option->value.GetValue<int64_t>();`

`Value::GetValue` rejects a NULL value by throwing an `InternalException` (`src/common/types/value.cpp:1310`, `"Calling GetValue on a value that is NULL"`). The grammar accepts `NULL` wherever a sequence option takes a numeric literal, so a plain `CREATE SEQUENCE` statement reaches that throw and surfaces an internal error to the user.

This is a missing input validation rather than a defect in sequence semantics: NULL is a legal literal at that grammar position, so it should be rejected with a binder/parser error naming the offending option, the same way a zero increment already is (`INCREMENT must not be zero`, line 49-50).

The same statement is handled gracefully everywhere else NULL can reach a sequence — `SELECT nextval(NULL)` returns NULL rather than raising.

### To Reproduce

Every sequence option that takes a numeric literal reproduces it, in a fresh in-memory database, with no prior state:

```sql
CREATE SEQUENCE s START NULL;
```

```sql
CREATE SEQUENCE s MINVALUE NULL;
CREATE SEQUENCE s MAXVALUE NULL;
CREATE SEQUENCE s INCREMENT BY NULL;
CREATE SEQUENCE s START WITH NULL;
```

`CREATE OR REPLACE SEQUENCE`, `CREATE TEMP SEQUENCE`, and combinations such as `CREATE SEQUENCE s CYCLE START NULL` behave identically.

### Result

```
INTERNAL Error: Calling GetValue on a value that is NULL

```

Expected: a regular error naming the option, for example `Binder Error: SEQUENCE start value must not be NULL`.

For contrast, NULL is handled correctly on the read path:

```sql
CREATE SEQUENCE s START 1;
SELECT nextval(NULL);   -- returns NULL, no error
```

### OS:

### DuckDB Version:

commit `7255424293982d7ec53b0d72b73a4b56c1c4ce0c`

### DuckDB Client:

CLI (Command Line Interface), built from source (release build)

### Hardware:

_No response_

### Full Name:

_No response_

### Affiliation:

_No response_

Yes

- [x] Yes, I have

- [x] Yes, I have

---

alwaysprince05:
## Description
CREATE SEQUENCE with a NULL option value (START / MINVALUE / MAXVALUE / INCREMENT BY)
threw an INTERNAL error. This rejects NULL with a clean Parser Error naming the option.

## What was happening
`CREATE SEQUENCE s START NULL;` (and the other numeric options) surfaced
`INTERNAL Error: Calling GetValue on a value that is NULL`.

## What should happen
A user-facing Parser Error, e.g. `Parser Error: START value must not be NULL`.

## Root cause
`TransformCreateSequenceStmt` called `Value::GetValue<int64_t>()` on each option's
parsed constant without checking `IsNull()`. NULL is a legal literal at that grammar
position, so it reached the assert in `Value::GetValue`.

## Changes
- Add `IsNull()` guards for INCREMENT, MINVALUE, MAXVALUE, START in
  `transform_create_sequence.cpp`, throwing `ParserException` naming the option.
- Add regression tests covering all four options, the `START WITH` alias, and a
  combined `CYCLE START NULL` case.

## Testing
- Focused: `test/sql/catalog/sequence/test_sequence.test` (fails before, passes after)
- Related: `test/sql/catalog/sequence/*`, `test/issues/fuzz/sequence_overflow.test`
- Broader: `test/sql/catalog/*`, `test/sql/parser/*`, `test/sql/create/*`
All pass.

## Related Issue
Fixes #24602

### DIFF
--- a/src/parser/peg/transformer/transform_create_sequence.cpp
+++ b/src/parser/peg/transformer/transform_create_sequence.cpp
@@ -45,6 +45,9 @@ unique_ptr<CreateStatement> PEGTransformerFactory::TransformCreateSequenceStmt(
 	for (auto &option : sequence_options) {
 		if (option.first == "increment") {
 			auto seq_val_option = unique_ptr_cast<SequenceOption, ValueSequenceOption>(std::move(option.second));
+			if (seq_val_option->value.IsNull()) {
+				throw ParserException("INCREMENT must not be NULL");
+			}
 			info->increment = seq_val_option->value.GetValue<int64_t>();
 			if (info->increment == 0) {
 				throw ParserException("Increment must not be zero");
@@ -68,6 +71,9 @@ unique_ptr<CreateStatement> PEGTransformerFactory::TransformCreateSequenceStmt(
 				continue;
 			}
 			auto seq_val_option = unique_ptr_cast<SequenceOption, ValueSequenceOption>(std::move(option.second));
+			if (seq_val_option->value.IsNull()) {
+				throw ParserException("MINVALUE must not be NULL");
+			}
 			info->min_value = seq_val_option->value.GetValue<int64_t>();
 			min_value_set = true;
 		} else if (option.first == "maxvalue") {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24620, opened 2026-08-09
Prevent unsafe narrowing-cast rewrite in comparison simplification

### LINKED ISSUE #23687: Optimizer makes narrow cast invertible

shch4747:
Fixes #23687. ComparisonSimplificationRule was stripping casts like TRY_CAST(a AS TINYINT) when rewriting comparisons, using CastIsInvertible as the sole guard. That function has no awareness of integer-narrowing casts (e.g. INTEGER -> TINYINT), so it incorrectly allowed the rewrite even when source values fall outside the target type's range, silently changing query results.

Added an explicit range check: for integer-to-integer casts, only allow the rewrite if the source type's full range fits inside the target type's range. Includes a regression test.

### DIFF
--- a/src/optimizer/rule/comparison_simplification.cpp
+++ b/src/optimizer/rule/comparison_simplification.cpp
@@ -172,6 +172,14 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op, 
 		if (!BoundCastExpression::CastIsInvertible(target_type, cast_expression.GetReturnType())) {
 			return nullptr;
 		}
+		// Narrowing between integer types loses information even when CastIsInvertible allows it
+		// (e.g. INTEGER -> TINYINT: not every INTEGER value fits in a TINYINT).
+		if (target_type.IsIntegral() && cast_expression.GetReturnType().IsIntegral()) {
+			if (Value::MinimumValue(target_type) < Value::MinimumValue(cast_expression.GetReturnType()) ||
+				Value::MaximumValue(target_type) > Value::MaximumValue(cast_expression.GetReturnType())) {
+        		return nullptr;
+			}
+		}
 
 		// Can we cast the constant at all?
 		string error_message;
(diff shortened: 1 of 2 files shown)


src/optimizer/rule/comparison_simplification.cpp
```diff
@@ -191,6 +172,14 @@ unique_ptr<Expression> ComparisonSimplificationRule::Apply(LogicalOperator &op,
 		if (!BoundCastExpression::CastIsInvertible(target_type, cast_expression.GetReturnType())) {
 			return nullptr;
 		}
+		// Narrowing between integer types loses information even when CastIsInvertible allows it
+		// (e.g. INTEGER -> TINYINT: not every INTEGER value fits in a TINYINT).
+		if (target_type.IsIntegral() && cast_expression.GetReturnType().IsIntegral()) {
```
Mytherin: Maybe we can use `BoundCastExpression::CastCanThrow` here?

shch4747:
CastCanThrow takes a try_cast argument, and if that's true, the function immediately returns false. So if you passed the expression's real "is this a TRY_CAST" flag, the check would short-circuit to "safe" every time for exactly the case you're fixing, and the bug would return. Instead I think we should ask "whether this pair of types lose information on a narrowing cast," which is true or false based on the types alone, independent of whether you're looking at CAST or TRY_CAST. So I hardcoded false for that argument regardless of the expression's actual cast kind, which lets the real type-comparison logic run instead of hitting the early exit.

PR #24635, opened 2026-08-09
Fix Windows static-extension builds with MSVC-targeting Clang

term-est:
We should give some leeway to windows builds using clang.
Clang builds the project under windows just fine, and we don't really need to make big changes to support it

The tiny problem is:
```cmake
if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR "${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
```

This is checking whether we use clang, and erroneously pass the `-Wl,--gc-sections -Wl,--exclude-libs,ALL` to the compiler, even when we are building under windows with clang

It is an easy fix, we just need to reorder the checks, so that we check whether we are in a native-ish windows build first and deal with it appropriately. These options are appropriate for the GNU/MinGW-style linker path, but not for a native Windows Clang build.

And again,
```cmake
if(MSVC OR MINGW)
```

can be just simplified to 

```
if(WIN32)
```

We are gonna be linking `ws2_32` and `rstrtmgr` under windows regardless of the compiler anyways.


Mytherin:
Thanks!

PR #24637, opened 2026-08-09
Avoid WindowSegmentTree aggregation for dynamic min/max N windows

### LINKED ISSUE #24449: Top-N MIN with row-dependent n fails in a singleton window frame
### What happens?

Top-N aggregate forms such as `MIN(value, n)` can fail as window aggregates when `n` is row-dependent, even if the frame is `ROWS BETWEEN CURRENT ROW AND CURRENT ROW`.

The frame for every output row contains only the current row, so each row should be allowed to use its own `n`. Instead, DuckDB raises `Invalid Input Error: Mismatched n values in min/max/arg_min/arg_max`.

### To Reproduce

```sql
SELECT
  MIN(i, CASE WHEN i < 16 THEN 1 ELSE 2 END) OVER (
  ) AS window_result
FROM range(17) AS t(i);
```

Run:

```sh
```

#### Expected

| expected |
| --- |
| [{"window_result":[0]},<br> ...<br>{"window_result":[16]}] |

#### Actual

| actual |
| --- |
| Invalid Input Error: Mismatched n values in min/max/arg_min/arg_max |

#### Notes From Minimization

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

jaiswalvikas9991:
Fixes #24449                                                                                                                                                                                 
                                                                                                                                                                                               
Window segment trees combine aggregate states while building higher levels. The min/max and arg_min/arg_max N variants store N as heap capacity, so dynamic N values can create incompatible states during combine.

This falls back to the naive window aggregator when the N argument is not foldable, and adds regression coverage for the affected N-variant aggregates.

Tested with:
```bash
./build/debug/test/unittest test/sql/window/test_minmax_n_dynamic_window.test
```

### DIFF
--- a/src/function/window/window_segment_tree.cpp
+++ b/src/function/window/window_segment_tree.cpp
@@ -10,6 +10,30 @@ namespace duckdb {
 //===--------------------------------------------------------------------===//
 // WindowSegmentTree
 //===--------------------------------------------------------------------===//
+
+// min/max style functions with an N argument store N as the aggregate heap
+// capacity. The segment tree builds range states from leaves and recursively
+// combines them into higher levels, so varying N values inside those ranges can
+// create incompatible non-leaf states. Only allow the segment tree when N is
+// constant/foldable.
+static bool HasFoldableMinMaxNArgument(const BoundWindowExpression &wexpr) {
+	const auto &aggr = *wexpr.AggregateFunction();
+	const auto &children = wexpr.GetChildren();
+	const auto &name = aggr.GetName();
+
+	if ((name == "min" || name == "max") && children.size() == 2) {
+		return !children[1]->HasParameter() && children[1]->IsFoldable();
+	}
+
+	if ((name == "arg_min" || name == "arg_max" || name == "argmin" || name == "argmax" || name == "min_by" ||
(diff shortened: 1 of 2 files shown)


jaiswalvikas9991:
This is my first open source contribution, so please let me know if I have misunderstood anything or if the fix/test coverage should be adjusted. Thanks.

hawkfish:
Hi @jaiswalvikas9991 - Thanks for the PR! 

Unfortunately, this is not the right approach. There are other builtin functions that have this "partition-constant" behaviour that are not in your list, and there could be used-defined aggregates in an extension with the same property.

I have a proper fix coming shortly that simply avoids combining segment tree nodes that cross partition boundaries. This works because these nodes are never used by combine.

jaiswalvikas9991:
@hawkfish 

Thanks for reviewing my PR. I really appreciate the effort.

hawkfish:
@jaiswalvikas9991 My fix [has been merged](https://github.com/duckdb/duckdb/pull/24986) if you want to have a look.

PR #24638, opened 2026-08-09
Fix histogram with row-dependent window bins

### LINKED ISSUE #24451: HISTOGRAM_EXACT with row-dependent bins fails in a singleton window frame
### What happens?

`HISTOGRAM_EXACT(x, bins)` can fail as a window aggregate when `bins` is a row-dependent list, even if the frame is `ROWS BETWEEN CURRENT ROW AND CURRENT ROW`. The equivalent ordinary aggregate over the singleton frame succeeds, and `debug_window_mode='separate'` also succeeds.

### To Reproduce

```sql
SELECT
  i AS id,
  HISTOGRAM_EXACT(0, CASE WHEN i < 16 THEN [0] ELSE [1] END) OVER (
    ORDER BY i
  )::VARCHAR AS window_result
FROM range(17) AS t(i)
ORDER BY id;
```

Run:

```sh
```

#### Expected

| id | expected |
| ---: | --- |
| 0 | `{0=1}` |
| 15 | `{0=1}` |
| 16 | `{1=0, 2147483647=1}` |

#### Actual

| actual |
| --- |
| Not implemented Error: Histogram - cannot combine histograms with different bin boundaries. Bin boundaries must be the same for all histograms within the same group |

#### Notes From Minimization

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

sakshichitnis27:
Fixes #24451

`histogram_exact` with row-dependent bin expressions failed for singleton window frames because the segment tree combined intermediate states with different bin boundaries, even though those rows were never part of the same frame.

This change marks binned histograms with row-dependent bins as fallible and avoids the segment-tree path for these multi-argument aggregates. Constant-bin histograms continue using the optimized path.

A regression test covering the reported query has been added.

Tests:
- `histogram_exact.test`
- `test_binned_histogram.test`
- `test_naive_aggregation.test`
- `test_kurtosis.test`

### DIFF
--- a/src/function/window/window_segment_tree.cpp
+++ b/src/function/window/window_segment_tree.cpp
@@ -24,6 +24,13 @@ bool WindowSegmentTree::CanAggregate(const BoundWindowExpression &wexpr) {
 		return false;
 	}
 
+	// Fallible aggregates with multiple arguments may have row-dependent parameters. The segment tree
+	// combines states outside individual frames, which can trigger errors for otherwise valid frames.
+	if (wexpr.AggregateFunction()->GetErrorMode() == FunctionErrors::CAN_THROW_RUNTIME_ERROR &&
+	    wexpr.GetChildren().size() > 1) {
+		return false;
+	}
+
 	//	We can't handle DISTINCT, ORDER BY args or () args (COUNT(*))
 	return !wexpr.Distinct() && wexpr.ArgOrders().empty() && !wexpr.GetChildren().empty();
 }
--- a/extension/core_functions/aggregate/nested/binned_histogram.cpp
+++ b/extension/core_functions/aggregate/nested/binned_histogram.cpp
@@ -391,6 +391,9 @@ unique_ptr<FunctionData> HistogramBinBindFunction(BindAggregateFunctionInput &in
 	}
 
 	function.ReplaceImplementation(GetHistogramBinFunction<HIST>(arguments[0]->GetReturnType()));
+	if (!arguments[1]->IsFoldable()) {
+		function.SetFallible();
+	}
 	return nullptr;
 }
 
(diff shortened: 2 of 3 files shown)


src/function/window/window_segment_tree.cpp
```diff
@@ -19,6 +19,11 @@ bool WindowSegmentTree::CanAggregate(const BoundWindowExpression &wexpr) {
 		return false;
 	}
 
+	if (wexpr.AggregateFunction()->GetErrorMode() == FunctionErrors::CAN_THROW_RUNTIME_ERROR &&
+	    wexpr.GetChildren().size() > 1) {
```
Mytherin, line 30: Could you expand on why we need the `GetChildren().size() > 1` condition here?

sakshichitnis27: The `GetChildren().size() > 1` condition keeps the fallback limited to fallible aggregates with additional row-dependent arguments. In this case, the second argument contains the histogram bin boundaries, so segment-tree construction can combine states initialized with different boundaries and throw even though those rows are never part of the same frame. Single-argument fallible aggregates do not have this separate per-row configuration argument, so they can retain the optimized path. I added an inline comment explaining this and rebased the PR onto the latest `main`.

sakshichitnis27:
Hello @Mytherin ,
Rebased the PR onto the latest main and resolved the conflicts while keeping the change limited to the original three files. I also added an inline comment explaining the multi-argument guard. The focused regression tests and the full local unit suite pass. The upstream Main workflow is currently awaiting maintainer approval.

PR #24640, opened 2026-08-09
Fix UB for ALP

### LINKED ISSUE #24634: ALP compression NumberToInt64 casts DOUBLE to int64 without a range check on CHECKPOINT — undefined behavior (numeric_utils.hpp:124)
### What happens?

Checkpointing a `DOUBLE` column that contains a value at/beyond the `int64` range triggers undefined behavior in the ALP floating-point compression analysis. When DuckDB picks a compression method for the column (`ColumnDataCheckpointer::DetectBestCompressionMethod`), ALP tries to encode each sampled value as an integer via `NumberToInt64(double)`, which casts the double to `int64` with no range check:

- `alp::AlpCompression<double, ...>::NumberToInt64` (`extension/.../alp/include/.../alp.hpp:124`) calls
- `LossyNumericCast<int64_t, double>` (`src/include/duckdb/common/numeric_utils.hpp:124`): `return static_cast<TO>(val);` — the helper's own comment notes *"no checks performed (at the moment, to be improved adding range checks)"*.

The double `9223372036854775807.0` is `2^63` (one past `INT64_MAX`), so `static_cast<int64_t>` of it is out of range — undefined behavior, flagged by UBSan:

```
src/include/duckdb/common/numeric_utils.hpp:124:25: runtime error: 9.22337e+18 is outside the range of representable values of type 'long'
```

Stack:
```
LossyNumericCast<long, double>                          numeric_utils.hpp:124
alp::AlpCompression<...>::NumberToInt64(double)          alp.hpp:124
alp::AlpCompression<...>::EncodeValue / FindTopKCombinations
AlpAnalyzeState<double>::FlushSegment / AlpFinalAnalyze  alp_analyze.hpp:157
ColumnDataCheckpointer::DetectBestCompressionMethod
CHECKPOINT
```

On a non-sanitized build the out-of-range cast silently yields an implementation-defined value that ALP then uses to score/encode the column, rather than erroring. The cast should be range-checked (or ALP should skip values outside `int64` range).

### To Reproduce

On a build compiled with `-fsanitize=undefined`. The table must be on-disk (so the checkpoint runs compression analysis) and hold enough rows to reach ALP analysis:

```sql
ATTACH 'alp.db' AS d;
CREATE TABLE d.t(v DOUBLE);
INSERT INTO d.t SELECT 9223372036854775807.0 FROM range(3000);
CHECKPOINT d;
```

An in-memory `CHECKPOINT` does not reach the on-disk compression path and does not reproduce it.

### Result

```
src/include/duckdb/common/numeric_utils.hpp:124:25: runtime error: 9.22337e+18 is outside the range of representable values of type 'long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/include/duckdb/common/numeric_utils.hpp:124:25
```

Expected: the double-to-int64 conversion in ALP is range-checked (values outside `int64` range are not fed to `NumberToInt64`), so checkpointing the column does not invoke undefined behavior.

### OS:

### DuckDB Version:

commit `e500d77864fe565f90e68f06d729c25b11e775c5` (main branch)

### DuckDB Client:

CLI, built from source with core extensions statically linked; build with `-fsanitize=address,undefined` (UndefinedBehaviorSanitizer surfaces the UB; on a non-sanitized build the cast is silently implementation-defined).

### Hardware:

_No response_

### Full Name:

_No response_

### Affiliation:

_No response_

- [x] Yes, I have

- [x] Yes, I have

---

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24634

The issue magic-number rounding moves a valid double outside the int64_t range, so we need to echeck the rounded value before casting. The UB error message clearly indicates the code location.

### DIFF
--- a/src/include/duckdb/storage/compression/alp/algorithm/alp.hpp
+++ b/src/include/duckdb/storage/compression/alp/algorithm/alp.hpp
@@ -121,6 +121,9 @@ struct AlpCompression {
 			return ExactNumericCast<int64_t>(AlpConstants::ENCODING_UPPER_LIMIT);
 		}
 		n = n + AlpTypedConstants<T>::MAGIC_NUMBER - AlpTypedConstants<T>::MAGIC_NUMBER;
+		if (IsImpossibleToEncode(n)) {
+			return ExactNumericCast<int64_t>(AlpConstants::ENCODING_UPPER_LIMIT);
+		}
 		return LossyNumericCast<int64_t>(n);
 	}
 
--- a/test/sql/storage/compression/alp/alp_out_of_range.test
+++ b/test/sql/storage/compression/alp/alp_out_of_range.test
@@ -1,0 +1,19 @@
+# name: test/sql/storage/compression/alp/alp_out_of_range.test
+# description: Test ALP compression of values outside the int64 range
+# group: [alp]
+
+load {TEST_DIR}/test_alp_out_of_range.db
+
+statement ok
+CREATE TABLE integers(v DOUBLE);
+
+statement ok
+INSERT INTO integers SELECT 9223372036854775807.0 FROM range(3000);
+
+statement ok
+CHECKPOINT;
+
+query II
+SELECT count(*), count(*) FILTER (WHERE v != 9223372036854775807.0) FROM integers;
+----
+3000	0


Mytherin:
Thanks!

PR #24645, opened 2026-08-10
Add data_page_size_limit option to Parquet COPY

### LINKED ISSUE #24507: Parquet writer: data pages split only at a hardcoded 100 MB uncompressed (no option) — 100x larger pages than other writers, slow downstream reads
### What happens?

The Parquet writer splits data pages only when a page reaches `MAX_UNCOMPRESSED_PAGE_SIZE` = **100 MB uncompressed**, a compile-time constant in [`primitive_column_writer.hpp`](https://github.com/duckdb/duckdb/blob/main/extension/parquet/include/writer/primitive_column_writer.hpp). With large row groups this produces a **single data page per column chunk** — e.g. one page holding 16,000,000 values (~18 MB compressed). There is no `COPY` option or setting to control data page size; the only page knob is `STRING_DICTIONARY_PAGE_SIZE_LIMIT`, which applies to dictionary pages only.

Other mainstream writers default to ~1 MB pages (parquet-mr `parquet.page.size` = 1 MB; arrow-rs `data_page_size_limit` = 1 MB plus `data_page_row_count_limit` = 20k; pyarrow `data_page_size` = 1 MB), so downstream readers are widely optimized for page-granular streaming, page skipping, and intra-chunk parallelism. DuckDB's effective page size is a ~100x outlier, and it is not configurable.

**Measured impact:** on a 142M-row table (9 files x 1 row group x 16M rows, snappy, fully dictionary-encoded), a downstream columnar engine (Microsoft Fabric Direct Lake) cold-loads DuckDB-written columns **5-26x slower** than files written by arrow-rs that are identical in every footer-visible dimension — same format version (1), same encodings (dictionary), same dictionary page sizes, same NDV, same row-group geometry. The only structural difference is the data page count: 1 giant page vs ~800 x 20k-row pages per 16M-row chunk (verified by parsing the page headers). Overall cold-load: 211 s vs 39 s for the same data.

### To Reproduce

```sql
COPY (SELECT i, 'val_' || (i % 400) AS s FROM range(16_000_000) t(i))
TO 'big_pages.parquet'
(FORMAT parquet, ROW_GROUP_SIZE 16_000_000, COMPRESSION snappy);
```

Inspect pages (e.g. `parquet pages big_pages.parquet` with parquet-cli, or any page-header walker): each column chunk contains a single data page with 16M values. `parquet_metadata()` cannot show this — footer metadata is unaffected; the difference is only visible at the page-header level.

### Ask

Expose data page size as a `COPY ... (FORMAT parquet, ...)` option (bytes and/or max rows per page, mirroring `data_page_size_limit` / `data_page_row_count_limit` in arrow-rs), or lower the default split threshold toward the ~1 MB ecosystem norm.

### OS:

Windows 11 / Ubuntu (ubuntu-latest CI) — x86_64

### DuckDB Version:

1.5.5

### DuckDB Client:

Python

### Hardware:

No response

### Full Name:

Mimoune Djouallah

### Affiliation:

Personal

Yes

- [x] Yes, I have

- [x] Yes, I have

---

*Note: this report was investigated and written with AI assistance (Claude); the page-header analysis, source-constant reference, and benchmark numbers were verified on real files.*

djouallah:
Fixes #24507

Adds a `data_page_size_limit` COPY option so data pages can be split below the hardcoded 100 MB threshold, which currently yields one huge page per column chunk. Today I write through DuckDB + delta-rs just to get normal-sized pages, which adds overhead; I'd like to write with DuckDB alone, but the huge pages are very problematic for downstream readers (Power BI in my case).

```sql
COPY tbl TO 'f.parquet' (FORMAT parquet, DATA_PAGE_SIZE_LIMIT 1048576);
```

Mirrors `string_dictionary_page_size_limit` end-to-end; default unchanged (100 MB, still the upper bound); limit is best-effort at vector granularity. Includes a sqllogictest.

Written with Claude Code; reviewed and submitted by me, To be clear: I understand the problem well (it's my daily workflow), but the C++ was AI-written

### DIFF
--- a/extension/parquet/include/parquet_writer.hpp
+++ b/extension/parquet/include/parquet_writer.hpp
@@ -167,6 +167,8 @@ struct ParquetWriterOptions {
 	optional_idx dictionary_size_limit;
 	//! The maximum bytes of string-data to put into the dictionary, per column/field
 	idx_t string_dictionary_page_size_limit;
+	//! The maximum uncompressed size of a data page
+	idx_t data_page_size_limit;
 	//! Whether to use bloom filters or not
 	bool enable_bloom_filters;
 	//! Maximum ratio of false-positives to allow in the written bloom filter
@@ -230,6 +232,9 @@ public:
 	}
 	idx_t StringDictionaryPageSizeLimit() const {
 		return options.string_dictionary_page_size_limit;
+	}
+	idx_t DataPageSizeLimit() const {
+		return options.data_page_size_limit;
 	}
 	bool EnableBloomFilters() const {
 		return options.enable_bloom_filters;
--- a/extension/parquet/parquet_extension.cpp
+++ b/extension/parquet/parquet_extension.cpp
@@ -105,6 +105,9 @@ struct ParquetWriteBindData : public TableFunctionData {
 	//! This is huge but we grow it starting from 1 MB
 	idx_t string_dictionary_page_size_limit = PrimitiveColumnWriter::MAX_UNCOMPRESSED_DICT_PAGE_SIZE;
 
+	//! The maximum uncompressed size of a data page
+	idx_t data_page_size_limit = PrimitiveColumnWriter::MAX_UNCOMPRESSED_PAGE_SIZE;
+
 	bool enable_bloom_filters = true;
 	//! What false positive rate are we willing to accept for bloom filters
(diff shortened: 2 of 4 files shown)


Mytherin:
Thanks for the PR! LGTM

PR #24646, opened 2026-08-10
Add support for GeoJSON in the `json` extension

Maxxen:
This PR adds a pair of scalar functions, a new `geojson := <BOOLEAN>` parameter to `read_json()` and a new `(FORMAT GEOJSON)` copy function variant to the `json` extension, to support handling GeoJSON.

# How it works

### Casting/Conversion

Prior to this PR, a `GEOMETRY->JSON` and `JSON->GEOMETRY` conversion just used the underlying string conversion, i.e. converted to/from WKT, and then quouted/unqouted it.

```sql
select 'POINT(0 1)'::GEOMETRY::JSON;
│ CAST(CAST('POINT(0 1)' AS GEOMETRY) AS "JSON") │                                                                                                                                                                                                                          
│                      json                      │                                                                                                                                                                                                                          
├────────────────────────────────────────────────┤                                                                                                                                                                                                                          
│ "POINT (0 1)"                                  │                                                                                                                                                                                                                          

select '"POINT(0 1)"'::JSON::GEOMETRY;
┌──────────────────────────────────────────────────┐
│ CAST(CAST('"POINT(0 1)"' AS "JSON") AS GEOMETRY) │                                                                                                                                                                                                                        
│                     geometry                     │                                                                                                                                                                                                                        
├──────────────────────────────────────────────────┤                                                                                                                                                                                                                        
│ POINT (0 1)                                      │                                                                                                                                                                                                                        
└──────────────────────────────────────────────────┘       

```
<br>
In addition to `WKT`, we now also allow casting a GeoJSON "fragment" object as well.


```sql
select '{"type":"Point","coordinates":[0.0,1.0]}'::JSON::GEOMETRY;
┌──────────────────────────────────────────────────────────────────────────────┐
│ CAST(CAST('{"type":"Point","coordinates":[0.0,1.0]}' AS "JSON") AS GEOMETRY) │                                                                                                                                                                                            
│                                   geometry                                   │                                                                                                                                                                                            
├──────────────────────────────────────────────────────────────────────────────┤                                                                                                                                                                                            
│ POINT (0 1)                                                                  │                                                                                                                                                                                            
└──────────────────────────────────────────────────────────────────────────────┘ 
```
<br>

This means that a A GeoJSON "fragment" object is now accepted anywhere a GEOMETRY is expected when converting from json (e.g. explicit `columns=`, `COPY FROM`, `::JSON::GEOMETRY`).

This PR also moves over two scalar functions from `spatial` into `json`, to perform this conversion through an explicit function instead
- `ST_AsGeoJSON(GEOMETRY) -> JSON` Convert from a geometry to GeoJSON fragment 
- `ST_GeomFromGeoJSON(JSON/VARCHAR) -> GEOMETRY` Convert from a GeoJSON fragment to a `GEOMETRY`


## Reading 

We do not generally infer stray geojson fragments as `GEOMETRY` values when reading regular json objects using e.g. `read_json` automatically. You can however opt-in to "geojson mode" by passing a new `geojson := true` optional argument.

With `geojson := true`, we convert geojson fragments into `GEOMETRY` values at any nesting depth, just as if they were a WKT string,

```sql
{
  "my_field": 10,
  "a_geometry": {"type":"Point","coordinates":[0.0,0.0]},
  "my_object": {
    "my_nested_field": 20,
    "my_nested_geometry": {"type":"Point","coordinates":[1.0,1.0]}
  }
}

----

┌──────────┬─────────────┬──────────────────────────────────────────────────────────────┐
│ my_field │ a_geometry  │                          my_object                           │                                                                                                                                                                                   
│  int64   │  geometry   │ struct(my_nested_field bigint, my_nested_geometry geometry)  │                                                                                                                                                                                   
├──────────┼─────────────┼──────────────────────────────────────────────────────────────┤                                                                                                                                                                                   
│       10 │ POINT (0 0) │ {'my_nested_field': 20, 'my_nested_geometry': 'POINT (1 1)'} │                                                                                                                                                                                   
└──────────┴─────────────┴──────────────────────────────────────────────────────────────┘ 

```
<br>

but we also detect GeoJSON "Features" as top-level line-delimited objects, as well as single top-level "FeatureCollection" objects. When reading "Features"/"FeatureCollections, the "property" field of each feature is "unnested" into top-level columns.

```sql
{"type":"FeatureCollection","features":[
	{"type":"Feature","geometry":{"type":"Point","coordinates":[0.0,1.0]},"properties":{"x":10}}
]}
----
┌─────────────┬───────┐
│  geometry   │   x   │                                                                                                                                                                                                                                                     
│  geometry   │ int64 │                                                                                                                                                                                                                                                     
├─────────────┼───────┤                                                                                                                                                                                                                                                     
│ POINT (0 1) │    10 │                                                                                                                                                                                                                                                     
└─────────────┴───────┘ 


{"type":"Feature","geometry":{"type":"Point","coordinates":[0.0,1.0]},"properties":{"x":10}}
----
┌─────────────┬───────┐
│  geometry   │   x   │                                                                                                                                                                                                                                                     
│  geometry   │ int64 │                                                                                                                                                                                                                                                     
├─────────────┼───────┤                                                                                                                                                                                                                                                     
│ POINT (0 1) │    10 │                                                                                                                                                                                                                                                     
└─────────────┴───────┘ 

```
<br>

We also support reading the special "feature-level" fields, such as `id` and `bbox` into separate columns. If there are properties with the same name, they get suffixed automatically.

```sql
{"type":"Feature","id":1,"geometry":{"type":"Point","coordinates":[0.0,1.0]},"bbox":[0.0,1.0,0.0,1.0],"properties":{"my_field":10}}
----
┌─────────────┬───────┬──────────────────────┬──────────┐
│  geometry   │  id   │         bbox         │ my_field │
│  geometry   │ int64 │       double[]       │  int64   │
├─────────────┼───────┼──────────────────────┼──────────┤
│ POINT (0 1) │     1 │ [0.0, 1.0, 0.0, 1.0] │       10 │
└─────────────┴───────┴──────────────────────┴──────────┘

```

`geojson := true` is enabled by default when reading a file ending in `.geojson` or `.geojsonl`.

## Writing

When exporting json through `COPY ... (FORMAT JSON)`, geometry values are still written as `WKT`, unless the `json_geometry_format` setting is set to `geojson`, in which case geometries are emitted as geojson "fragments".

```sql
copy (select 'POINT(0 1)'::GEOMETRY as geom, 10 as my_field) to 'test.json';
----
{"geom":"POINT (0 1)","my_field":10}

-- with geojson format
set json_geometry_format='geojson';
copy (select 'POINT(0 1)'::GEOMETRY as geom, 10 as my_field) to 'test.json';
----
{"geom":{"type":"Point","coordinates":[0.0,1.0]},"my_field":10}
```

To emit "Features" and "FeatureCollections, here's now a separate `COPY ... (FORMAT GeoJSON)` copy-format. By default, it emits a single "FeatureCollection", but if you also pass `ARRAY TRUE` (or use `FORMAT GeoJSONL`), you get newline-delimited "Features" instead. You do not need to set `json_geometry_format='geojson'` when using these copy-formats.

```sql
copy (select 'POINT(0 1)'::GEOMETRY as geom, 10 as my_field) to 'test.json' (FORMAT GEOJSON);
----
{"type":"FeatureCollection","features":[
	{"type":"Feature","geometry":{"type":"Point","coordinates":[0.0,1.0]},"properties":{"my_field":10}}
]}

copy (select 'POINT(0 1)'::GEOMETRY as geom, 10 as my_field) to 'test.json' (FORMAT GEOJSONL);
{"type":"Feature","geometry":{"type":"Point","coordinates":[0.0,1.0]},"properties":{"my_field":10}}
```

A column named "id" will automatically be exported as the "feature id" field. But this can be controlled by passing the "ID_COLUMN <COLUMN_NAME>" parameter. In case you want a "id" field to be emitted as a property instead, just set `ID_COLUMN` to `NULL` or `""`

```sql
copy (select 'POINT(0 1)'::GEOMETRY as geom, 1 as id, 10 as my_field) to 'test.json' (FORMAT GEOJSONL);
----
{"type":"Feature","id":1,"geometry":{"type":"Point","coordinates":[0.0,1.0]},"properties":{"my_field":10}}


-- With alternative "id"
copy (select 'POINT(0 1)'::GEOMETRY as geom, 1 as id, 10 as my_field) to 'test.json' (FORMAT GEOJSONL, ID_COLUMN my_field);
----
{"type":"Feature","id":10,"geometry":{"type":"Point","coordinates":[0.0,1.0]},"properties":{"id":1}}

-- With no "id"
copy (select 'POINT(0 1)'::GEOMETRY as geom, 1 as id, 10 as my_field) to 'test.json' (FORMAT GEOJSONL, ID_COLUMN '');
----
{"type":"Feature","geometry":{"type":"Point","coordinates":[0.0,1.0]},"properties":{"id":1,"my_field":10}}

```

You can also pass `BBOX` to automatically generate a bounding-box as a feature-level field for each feature. It is however not possible to pass your own existing bbox column.

```sql
copy (select 'POINT(0 1)'::GEOMETRY as geom, 1 as id, 10 as my_field) to 'test.json' (FORMAT GEOJSONL);
----
{"type":"Feature","id":1,"geometry":{"type":"Point","coordinates":[0.0,1.0]},"bbox":[0.0,1.0,0.0,1.0],"properties":{"my_field":10}}
```

By default we use the first `GEOMETRY` column as the "geometry" field. If you for some reason also want to pass a geometry as part of the "property" field, you can specify which geometry is the "main" geometry by passing the column name to the `GEOMETRY_COLUMN <COLUMN>` parameter:

```sql
copy (select 'POINT(0 1)'::GEOMETRY as a_geom, 'POINT(42 42)'::GEOMETRY as b_geom) to 'test.json' (FORMAT GEOJSONL, GEOMETRY_COLUMN 'b_geom');
----
{"type":"Feature","geometry":{"type":"Point","coordinates":[42.0,42.0]},"properties":{"a_geom":{"type":"Point","coordinates":[0.0,1.0]}}}
```

# Motivation

`spatial` has had support for GeoJSON as an input/output format through the GDAL based `ST_Read` for a long time already. Unfortunately `ST_Read` is not as efficient, and does not support as many features as our own `read_json` function in the json reader (multi-file reads, schema sniffing, streaming reads, partitioning). I've thought about creating our own geojson reader/writer multiple times, but quickly realized that it would just end up duplicating a ton of code that we have already in the `json` extension. It's also very likely that a separate implementation would have lot of slight differences in behavior, e.g. schema inference, causing a pretty bad UX.

Since a GeoJSON reader needs to also perform "general purpose" schema inference to convert feature properties into columns, and since we now have `GEOMETRY` in core (since DuckDB v1.5), it makes more sense to me to just implement this functionality in the `json` extension which is already amazing at handling unstructured json. 

The two scalar functions (`ST_AsGeoJSON`/`ST_GeomFromGeoJSON`) are still part of the `spatial` extension, but can now be removed. Moving them into `json` allow us to re-use the same conversion logic for the table/copy function, as well as potentially remove the `yyjson` dependency from `spatial` altogether.

So in short, this gives us faster, more powerful, more integrated into DuckDB - GeoJSON handling, and also enables us to slim down `spatial`.


Maxxen:
The 👏 CI 👏 failure 👏 is 👏 unrelated 👏

PR #24647, opened 2026-08-10
 Give USING KEY ... UNION changed-key delta semantics

kryonix:
`USING KEY` currently treats `UNION` like `UNION ALL`: every candidate row becomes input to the next iteration, even when applying that candidate does not change the keyed recurring table. I ran into the cost of that behavior while studying an LDBC SF100 pathfinding query. The query generated 20.9 million aggregate candidates, but only 3.7 million distinct keys were observable after their epoch had been applied. Carrying all candidates into the next iteration repeated a large amount of work and made termination depend on candidate production rather than convergence of the keyed state.

I think plain `UNION` has a useful and natural meaning here that remains distinct from the documented candidate-frontier behavior of `UNION ALL`: apply the complete epoch to the declared keyed aggregates, then recurse only from finalized keys whose observable value changed.

In short, this PR:

- gives `USING KEY ... UNION` changed-key delta semantics;
- leaves `USING KEY ... UNION ALL` candidate-frontier semantics unchanged;
- preaggregates duplicate-heavy epochs for eligible aggregates, while keeping tiny and mostly-unique epochs on the direct update path;
- compares finalized payloads with SQL `IS DISTINCT FROM` semantics, including collations;
- gives recurring keys the same collation identity as `GROUP BY`, including specialized complete and partial probes;
- adds detailed delta, preaggregation, and classifier metrics; and
- removes the temporary compatibility setting for the old interpretation of plain `UNION`.

If `C_i` is the candidate relation produced by iteration `i`, `S_i` is the keyed recurring state before applying it, and `W_i` is the next working table, the two forms now differ as follows:

```text
USING KEY ... UNION ALL
S_(i+1) = update(S_i, C_i)
W_(i+1) = C_i

USING KEY ... UNION
S_(i+1) = update(S_i, C_i)
W_(i+1) = { S_(i+1)[k] | S_(i+1)[k] IS DISTINCT FROM S_i[k] }
```

The `UNION` delta contains one finalized row per new or observably changed key. Duplicate candidates, inferior `min`/`max` candidates, and candidates that reproduce the current value still participate in the aggregate update, but they do not become recursive work by themselves. Recursion terminates when no keyed state changes.

## Applying and observing an epoch

The executor buffers an epoch's candidates, snapshots the finalized payload the first time each key is touched, applies every candidate to the recurring aggregate state, and finalizes each touched key once. It emits new keys and existing keys whose finalized payload is `IS DISTINCT FROM` the snapshot. The snapshot remains stable across recurring hash-table growth, and partial-key indexes are updated only after new group addresses are stable.

Raw vector comparison remains the fast path for types where it already implements SQL equality. For payloads requiring collation normalization, physical planning builds an `IS DISTINCT FROM` expression with the same collation transformation used by the binder. This covers explicit and default collations, prepared statements, nested collated values, NULLs, NaNs, signed zero, and multi-column payloads without adding a logical-plan field.

There is one related binder edge case: a syntactically recursive `USING KEY` CTE is allowed even when neither input actually references the CTE. The old non-recursive lowering discarded the key and aggregate declarations and used an ordinary set operation. I now lower this case to `UNION ALL`, followed by the declared keyed aggregate and a projection back to the original column order. This preserves duplicate multiplicity and aggregate output types.

Key identity now follows the same collation normalization that the binder applies to `GROUP BY`. The recurring hash table, changed-key snapshots, epoch preaggregation, and partial-key indexes all use the normalized representation. A hidden physical `first` aggregate retains an original representative for visible key output. Supported complete and partial `RECURSIVE_KEY_JOIN` probes normalize their probe keys before looking directly into the recurring table or secondary index; collated scalar keys therefore keep the specialized path instead of rebuilding an ordinary join hash table.

This exposed an independent vector-hashing contract bug for first-column `LIST` and `ARRAY` keys: their hash loop populated values without setting the result vector cardinality. Setting that cardinality at the generic hash boundary fixes natural `UNION` lookup for nested keys as well as the recursive case.

The logical recursive operator already serializes `union_all`, so none of this requires a new logical operator, serialized field, or compatibility flag. Key and payload normalization expressions, representative aggregates, and partial-key bookkeeping are physical-only state.

## Duplicate-heavy epochs

Applying millions of duplicate keys directly to the recurring hash table leaves a substantial amount of avoidable aggregate and first-touch work. For plain `UNION`, I therefore preaggregate an epoch in a temporary keyed hash table when every payload aggregate has a combine callback and is explicitly not order-dependent. The temporary states are destructively combined into the recurring table, after which the normal finalized changed-key delta is produced. Order-dependent aggregates such as the default `last`, mixed aggregate sets, and extension aggregates without a combine callback stay on the original per-candidate path.

Preaggregation is selected per epoch rather than being enabled for the complete query. Epochs smaller than one standard vector bypass classification, as do non-expanding epochs whose candidate count does not exceed the previous frontier. Larger eligible epochs build a small HyperLogLog sketch over the key columns and preaggregate only when the sketch's conservative upper cardinality bound is below one quarter of the candidate count. The scan exits early once it has enough distinct-key evidence to reject preaggregation. This makes the decision independent of candidate order and avoids the large regression that unconditional preaggregation caused for tiny frontiers, as well as repeated hashing of wide one-to-one keys. The choice only affects the execution strategy; both paths implement the same aggregate and delta semantics.

The existing structured `PhysicalOperator` log now reports candidate, touched, new, changed, and unchanged key counts; delta work; preaggregation candidates and groups; aggregate and combine work; and classifier work. These fields made it possible to validate both the result and the strategy on the larger workload.

## Performance

I measured exact rebased main `e500d77864fe565f90e68f06d729c25b11e775c5` and final branch HEAD `60b6904d4d1e642d38f3a1236311f462c630b18b` as process-isolated Release builds on macOS arm64. Both builds use `tpch;tpcds;core_functions;parquet`, and the LDBC query uses eight threads. The public LDBC SF100 database contains 32 pathfinding pairs; every measured version returns `(32, 29, 108, 3, 5)`.

| Comparison | Baseline | Branch | Result |
| --- | ---: | ---: | ---: |
| Query time, 3/5 timed runs, median (mean ± stdev) | 19.319 s (19.297 ± 0.064 s) | 2.948 s (2.947 ± 0.020 s) | 6.55x median speedup |
| Peak RSS in those runs | 3.918 GB | 2.663 GB | 32% lower |

I also compared all 63 recursive regression benchmarks on exact final HEAD at two threads with adaptive paired sampling. Sixty are within ±2%. The suite geometric mean improves from 48.3 ms to 45.6 ms, or 5.5%.

| Benchmark | Baseline median | Branch median | Change |
| --- | ---: | ---: | ---: |
| Duplicate fan-in | 11.5 ms | 1.9 ms | 83.1% faster |
| Duplicate-heavy preaggregation | 14.8 ms | 0.930 ms | 93.7% faster |
| Wide unique-key frontier | 15.3 ms | 16.2 ms | 6.0% / 0.913 ms slower |

The wide case is the one remaining stable loss. It uses plain `UNION`, so the baseline forwards candidates without proving that the finalized keyed value changed, while this branch must perform that proof before reusing the candidate frontier. The final implementation avoids collection materialization and allocation when the proof succeeds, but the observable-value comparison still costs about 0.9 ms over 102,400 existing-key updates. Ordinary recursion, `UNION ALL`, complete and partial probes, Flummi, AoC, and the other research workloads remain within measurement noise. Removing the residual generically would require aggregate-level observable-change reporting; I am leaving that as a follow-up rather than adding `min`/`max` name checks to the recursive executor.

On the LDBC diagnostic run, the adaptive path preaggregates 17,016,444 of 20,964,191 candidates. The earlier explicit candidate-`DISTINCT` rewrite from the performance study measured roughly 2.29 seconds, so worker-local preaggregation before global candidate collection remains a separate follow-up rather than expanding this PR into a parallel sink redesign.


kryonix:
I want to write down the semantic choice behind this PR more explicitly, because a discussion with colleagues exposed a real and non-obvious trade-off.

The current PR gives `USING KEY ... UNION` a finalized state-delta interpretation. If `Cᵢ` is the candidate relation produced by an epoch and `Sᵢ` is the keyed recurring state before applying it, the semantics are:

```text
Sᵢ₊₁ = update(Sᵢ, Cᵢ)

Wᵢ₊₁ = {
    Sᵢ₊₁[k]
    | Sᵢ₊₁[k] IS DISTINCT FROM Sᵢ[k]
}
```

The next working table therefore contains one finalized row per new or observably changed key. All candidates still participate in the declared aggregates, but duplicate candidates, inferior `min`/`max` candidates, and candidates whose combined effect cancels out do not independently become recursive work.

For example, if an existing key has `sum(v) = 0` and an epoch produces `+1` and `-1`, the final value remains `0`. The key is absent from the next working table. If the key did not exist before, creating it with a final value of `0` is still a change because row existence changed.

Inside the recursive term, this means:

```text
T             = finalized rows for keys changed by the previous epoch
recurring.T   = complete keyed state, frozen for the current epoch
```

After recursion terminates, `T` returns the complete final keyed state.

`USING KEY ... UNION ALL` retains the existing candidate-frontier semantics:

```text
Sᵢ₊₁ = update(Sᵢ, Cᵢ)
Wᵢ₊₁ = Cᵢ
```

Every candidate remains recursive work under `UNION ALL`, including duplicates and candidates that do not affect the final aggregate value.

A useful shorthand for the distinction is therefore:

```text
UNION     = finalized state-change frontier
UNION ALL = candidate/event frontier
```

The unintuitive consequence is that the two working tables do not necessarily have the same types. Under `UNION ALL`, `T` contains aggregate inputs. Under `UNION`, `T` contains finalized aggregate results. For example, integer candidates supplied to `sum()` produce a finalized `HUGEINT`, while candidates supplied to `list(INTEGER)` produce a finalized `INTEGER[]`.

That difference is larger than the usual distinction between SQL `UNION` and `UNION ALL`, and I think it is important to treat it as an explicit language-design trade-off rather than hide it as an implementation detail.

### Candidate-preserving alternative

The alternative we originally discussed was to keep `T` as a candidate relation under both variants and let `UNION` discard candidates that do not result in a keyed-state change.

The robust version of that alternative would have to be defined per key:

```text
Sᵢ₊₁ = update(Sᵢ, Cᵢ)

KΔ = {
    k
    | Sᵢ₊₁[k] IS DISTINCT FROM Sᵢ[k]
}

Wᵢ₊₁ = {
    c ∈ Cᵢ
    | key(c) ∈ KΔ
}
```

This gives `UNION` and `UNION ALL` working tables the same candidate schema:

```text
UNION     = candidates belonging to effectively changed keys
UNION ALL = all candidates
```

However, "retain the candidates that caused a change" cannot generally be defined more precisely than that.

Consider:

```text
old min = 10
candidates = 12, 8, 5
new min = 5
```

It is unclear whether the working table should contain only `5`, both `8` and `5`, or all three candidates. A sequential definition would depend on candidate order: applying `5` first makes `8` ineffective, while applying `8` first makes it a temporary improvement. That would make semantics depend on pipeline scheduling, parallelism, and preaggregation.

Comparing every candidate independently against the old frozen state avoids the ordering problem, but does not represent final state convergence. With `sum = 0` and candidates `+1` and `-1`, both candidates independently change the old state, even though their combined final effect is zero.

Trying to identify only the candidates that contributed to the final result is also not generic. It may be possible for `min()` or `max()`, but there is no corresponding single winning candidate for `sum()`, `avg()`, `list()`, multiple aggregate payloads, or arbitrary extension aggregates.

The only generic candidate-preserving definition is therefore to retain all candidates for a key when that key changed after the complete epoch. This has some surprising consequences of its own:

| Old state | Candidates | Final state | Current PR | Candidate-preserving alternative |
|---|---|---|---|---|
| `sum = 0` | `+1, -1` | `0` | no row | no rows |
| `sum = 0` | `+1, +1` | `2` | one finalized row containing `2` | both `+1` candidates |
| `min = 10` | `12, 8, 5` | `5` | one finalized row containing `5` | all three candidates |
| `list = [0]` | `1, 1` | `[0,1,1]` | one finalized row containing `[0,1,1]` | both `1` candidates |

This alternative provides the more stable surface contract:

```text
T           = candidate rows selected for further work
recurring.T = complete aggregated keyed state
```

The recursive term would obtain the finalized aggregate state from `recurring.T`, while `T` would only identify candidate work.

### Implementation and performance consequences

The current state-delta semantics permits the executor to preaggregate an epoch and discard the raw candidate rows. Once the aggregate states have been combined, it only needs to emit one finalized row per changed key.

A candidate-preserving implementation must retain the raw candidate rows, their multiplicities, and their values until the final changed-key set is known. It could preaggregate a separate copy for state maintenance, but it could not replace the candidate collection with aggregate states. That increases memory traffic and weakens the main optimization in this PR.

This distinction matters for the motivating LDBC workload. It produces roughly 20.9 million aggregate candidates but only about 3.7 million distinct keyed results across the relevant epochs. The current semantics can collapse that fan-in before the next recursive step. The candidate-preserving alternative may still forward many duplicate or inferior candidates whenever their key changed for some other reason. I would therefore need to remeasure the reported performance results if the PR changed to that interpretation.

### Recommendation

My current preference is to retain the finalized state-delta semantics in this PR.

It is deterministic, independent of candidate order, generic across aggregate implementations, and ties termination directly to convergence of the observable keyed state. It also enables the duplicate-collapse and epoch-preaggregation benefits that motivated the work.

The criticism about the working-table types remains valid. The proposed behavior should not be presented as merely the ordinary duplicate-eliminating counterpart of `UNION ALL`. It is deliberately a different recursive mode, and replacing `UNION ALL` with `UNION` can change both the meaning and type of `T` inside the recursive member.

lnkuiper:
LGTM! Just needs the merge conflict to be resolved

PR #24652, opened 2026-08-10
Keep the DISTINCT modifier on last()

### LINKED ISSUE #24625: `last(DISTINCT x)` can return the non-distinct last value
### What happens?

`last(DISTINCT x)` returns the same value as `last(x)` after the expression rewriter runs.

In the minimized input below, the distinct values appear as `1, 2`, so the non-rewritten `DISTINCT` aggregate path returns `2`. The default optimized plan returns `1`, which is the final input row and ignores the `DISTINCT` modifier.

#### Code

```sql
CREATE TABLE t(id INTEGER, x INTEGER);
INSERT INTO t VALUES (1, 1), (2, 2), (3, 1);

SELECT last(DISTINCT x) AS actual
FROM t;

SET disabled_optimizers='expression_rewriter';

SELECT last(DISTINCT x) AS expected
FROM t;
```

Run:

```sh
./build/release/duckdb -json -f bug.sql
```

#### Expected

| expected |
| ---: |
| 2 |

#### Actual

| actual |
| ---: |
| 1 |

#### Notes From Minimization

- `SET disabled_optimizers='expression_rewriter'` changes the output from `1` to `2`.
- Replacing `last(DISTINCT x)` with `last(x)` keeps the default output at `1`.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (`e500d77864fe`)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

saime428:
Fixes #24625.

`DistinctAggregateOptimizer` removes the `DISTINCT` modifier from aggregates marked `NOT_DISTINCT_DEPENDENT`. `first()`, `last()` and `any_value()` share the same two bind functions (`BindFirst`, `BindDecimalFirst`) and were all marked that way unconditionally.

The rewrite is sound for `first()` and `any_value()`. It is not for `last()`: the last distinct value is not the last row's value, so whether the optimizer ran changes the result.

On current main:

```sql
CREATE TABLE t(id INTEGER, x INTEGER);
INSERT INTO t VALUES (1, 1), (2, 2), (3, 1);

SELECT last(DISTINCT x) FROM t;                    -- 1
SET disabled_optimizers='expression_rewriter';
SELECT last(DISTINCT x) FROM t;                    -- 2
```

## Changes

- `first_last_any.cpp`: mark `last()` as `DISTINCT_DEPENDENT`; `first()` and `any_value()` keep the existing behaviour.
- Adds a regression case to the existing `distinct_aggregate_rewrite.test`, placed before `PRAGMA disable_verification` so it runs with verification enabled.

## Validation

- [x] `test/sql/aggregate/distinct/rewrite/distinct_aggregate_rewrite.test` — fails on unpatched main (actual 1, expected 2), passes with this change
- [x] `test/sql/aggregate/*` — 180/181. The one failure is `test_string_agg_overflow.test_slow`, which needs a 4 GiB allocation and runs out of memory on the 7.8 GiB machine I used; unrelated to this change
- [x] `test/sql/window/*` — 77 cases, 96228 assertions, all passing (`DistinctWindowedOptimizer` reads the same flag)
- [x] `clang-format --dry-run -Werror` on the changed source file
- [ ] `make allunit` / `test_configs` / `test_vector` — not run, too slow on the machine available to me

## Notes for reviewers

`first()` is deliberately left alone. On main the deduplication happens to iterate in first-seen order, so `first(DISTINCT x)` agrees with `first(x)` and I could not produce a failing case. On the v1.5.5 release binary the iteration order differs and `first()` is the one that disagrees, so the same class of problem applies to it. I did not change it because there is no reproducible failure on main — happy to cover both if you would rather have it handled in one go.

This does not give DISTINCT aggregates an ordering guarantee. Both functions are documented as "affected by ordering", and with a `VALUES` source `last(DISTINCT x)` still varies between runs after this change. The fix only removes the discrepancy between the optimized and unoptimized plans for the reproducible case.

Heads-up: #24647 touches the immediate neighbourhood of the same two lines (it adds `SetSingleValueIdentity` directly below them). Whichever lands first, the other is a trivial rebase.

### DIFF
--- a/src/function/aggregate/distributive/first_last_any.cpp
+++ b/src/function/aggregate/distributive/first_last_any.cpp
@@ -434,7 +434,9 @@ unique_ptr<FunctionData> BindDecimalFirst(BindAggregateFunctionInput &input) {
 	auto name = function.GetName();
 	function.ReplaceImplementation(GetFirstFunction<LAST, SKIP_NULLS>(decimal_type));
 	function.SetName(std::move(name));
-	function.SetDistinctDependent(AggregateDistinctDependent::NOT_DISTINCT_DEPENDENT);
+	// the last distinct value is not the last row's value, so last() is distinct-dependent
+	function.SetDistinctDependent(LAST ? AggregateDistinctDependent::DISTINCT_DEPENDENT
+	                                   : AggregateDistinctDependent::NOT_DISTINCT_DEPENDENT);
 	function.SetReturnType(decimal_type);
 	return nullptr;
 }
@@ -456,7 +458,9 @@ unique_ptr<FunctionData> BindFirst(BindAggregateFunctionInput &input) {
 	auto name = function.GetName();
 	function.ReplaceImplementation(GetFirstOperator<LAST, SKIP_NULLS>(input_type));
 	function.SetName(std::move(name));
-	function.SetDistinctDependent(AggregateDistinctDependent::NOT_DISTINCT_DEPENDENT);
+	// the last distinct value is not the last row's value, so last() is distinct-dependent
+	function.SetDistinctDependent(LAST ? AggregateDistinctDependent::DISTINCT_DEPENDENT
+	                                   : AggregateDistinctDependent::NOT_DISTINCT_DEPENDENT);
 	return nullptr;
 }
 
--- a/test/sql/aggregate/distinct/rewrite/distinct_aggregate_rewrite.test
+++ b/test/sql/aggregate/distinct/rewrite/distinct_aggregate_rewrite.test
@@ -349,6 +349,24 @@ NULL	P	3	1	0
 NULL	NULL	6	1	1
 
 statement ok
+CREATE TABLE distinct_rewrite_last(id INTEGER, x INTEGER);
+
+statement ok
+INSERT INTO distinct_rewrite_last VALUES (1, 1), (2, 2), (3, 1);
+
+# the last distinct value differs from the last row's value, so the DISTINCT modifier is kept
+query I
+SELECT last(DISTINCT x) FROM distinct_rewrite_last;
+----
+2
+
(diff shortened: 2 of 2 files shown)


Mytherin:
Thanks for the PR - but I'm not sure if this makes much sense. `last(DISTINCT ...)` introduces a hash table which does not have any ordering semantics attached to it, so `last(DISTINCT ...)` would get loose / unspecified semantics. I would say this is not desirable.

saime428:
That makes sense, and I agree with closing it. The hash table has no ordering to offer, so pinning the rewrite in place would only lock in the unspecified answer.

One data point in case it is useful for the issue. I rebuilt current main without the patch and checked how far the discrepancy actually reaches. Across 18 query shapes (plain aggregate, GROUP BY, GROUPING SETS, FILTER, HAVING, mixed DISTINCT aggregates, CTE, correlated subquery, LATERAL, UNION ALL, CTAS, recursive CTE, view, prepared statement, PIVOT, ORDER BY on the aggregate, unnest subquery) `last(DISTINCT x)` returns `last(x)` under default settings every time. The only way I could get `2` was `SET disabled_optimizers='expression_rewriter'`.

That matches the code: the flag is read in exactly two places, `DistinctAggregateOptimizer` and `SimplifyWindowedAggregate`, and the latter sits inside an `EnableOptimizerSetting` check, so both are behind the same switch.

So #24625 only reproduces through a debug pragma. It may be worth closing on the grounds you gave here. Worth noting as well that the report lists the unoptimized `2` as the expected value, which is the opposite of the semantics you described.

Thanks for the quick look.

PR #24654, opened 2026-08-10
[ASYNC-IO] On Demand ASYNC IO for native DuckDB Tables

pdet:
This PR implements on-demand async io for native DuckDB files, with read-ahead to come in a follow-up PR.

With a  TPC-H SF100  S3 DuckDB file, and a `r7i.16xlarge` EC2 machine in the same region, Q6 execution time dropped from 8.8s to 4.5s


src/include/duckdb/storage/table/scan_state.hpp
```diff
@@ -231,6 +232,8 @@ struct PreparedScanVector {
 
 	//! Whether a vector is currently prepared for processing
 	bool prepared = false;
+	//! Whether I/O for the prepared vector has been registered
+	bool io_registered = false;
```
lnkuiper: Can we convert `prepared` and `io_registered` into one enum, e.g.:
```sql
enum class VectorPrepareState : uint8_t {
    NONE,
    PREPARED,
    IO_REGISTERED
};
```

test/sql/storage/prefetch/async_native_scan_positional.test
```diff
@@ -0,0 +1,27 @@
+# name: test/sql/storage/prefetch/async_native_scan_positional.test
+# description: Positional scans cannot receive resume callbacks, async scan I/O must fall back to synchronous execution
+# group: [prefetch]
+
+load __TEST_DIR__/async_native_scan_positional.db
```
lnkuiper: Can we use `{TEST_DIR}` in these tests instead? We're deprecating `__TEST_DIR__`

lnkuiper (commented):
Thanks! Looks great, just two minor comments:

PR #24655, opened 2026-08-10
Add UUID to temporary file names

carlopi:
Currently temporary files (used by duckdb to spill to disk) might collide between different DatabaseInstances, both same process and different co-existing processes. I think it would be cleaner to mark them as different via an UUID each `DatabaseInstance` is stamped with.

This do not properly solves all issues (like the first instance to close down will delete the other instance folder), but this at least turns silent failures into either working situation or a clear throw (that a file has been invalidated).

Possible follow ups to be considered:
1. fold `GenerateUniqueUUID` into a SystemUtils style utility, that could also allow re-initializing this multiple times.
2. add a lock file of sort, and on non-crashing code-paths an instance can avoid clean-up of folders that still have active in-use lock files
3. shortening of the names, currently they might be like:
```
.tmp/
 -> duckdb_temp_3f2a1c8e-7b4d-4a19-9c05-6e8f0d2b7a31_storage_DEFAULT-0.tmp   (70 chars)
 -> duckdb_temp_3f2a1c8e-7b4d-4a19-9c05-6e8f0d2b7a31_block-12345.block
```
vs previous:
```
.tmp/
 -> duckdb_temp_storage_DEFAULT-0.tmp   (34 chars)
 -> duckdb_temp_block-12345.block
```
that on file-systems that limit file length might be an issue we could now run into.

Bumped into this while running multiple unittester side-by-side, current workaround is explicit temporary folder, but we should have a solution that works by default for these cases.

### DIFF
--- a/src/common/types/uuid.cpp
+++ b/src/common/types/uuid.cpp
@@ -1,4 +1,5 @@
 #include "duckdb/common/types/uuid.hpp"
+#include "duckdb/common/atomic.hpp"
 #include "duckdb/common/chrono.hpp"
 #include "duckdb/common/random_engine.hpp"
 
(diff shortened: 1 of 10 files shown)


hannes:
Maybe we could use a shorter random string (e.g. airline booking reference style), this does not have to be globally-unique

carlopi:
> Maybe we could use a shorter random string (e.g. airline booking reference style), this does not have to be globally-unique

Actually implemented also point 2, and had naming be as short as possible, so in the common case is just 2 extra chars.

PR at https://github.com/duckdb/duckdb/pull/24811

PR #24660, opened 2026-08-10
Do not push down an extract on an already extracted column

### LINKED ISSUE #24657: CAST on a nested struct leads to incorrect result in top-N queries
### What happens?

Reading a field out of a nested struct that is reached through a `CAST` returns the value of a different field. 
This is coming from asserting a type on a nested object in semi-structured data, for example `(payload.user::STRUCT(id VARCHAR, name VARCHAR)).name`, which returns the whole `user` struct.

It only goes wrong in top-N queries, and there are two identified ways:
* `ORDER BY ... LIMIT` is planned as a late materialization
* `row_number() OVER (...) = 1` goes through top-N window elimination

The same queries return the correct value with the corresponding optimizer disabled.

```sql
CREATE TABLE windowed AS SELECT
    k % 2 AS grp, {'id': k + 1000, 'ms': {'a': k + 200, 'b': 50 - k}} AS v
FROM range(6) t(k);

SELECT grp, v.id FROM windowed
QUALIFY row_number() OVER (PARTITION BY grp ORDER BY (v.ms::STRUCT(b VARCHAR, a VARCHAR)).b) = 1;
-- 1000 and 1001, expected 1004 and 1005, with no ORDER BY or LIMIT anywhere
```

Extracting one level deeper, it raises an internal error instead:

```
INTERNAL Error: PushdownExtract not supported for StatisticsType::NUMERIC_STATS
```

### To Reproduce

```sql
CREATE TABLE nested AS SELECT
    {'id': k + 1000, 'ms': {'a': k + 200, 'b': k + 30, 'c': {'d': k + 7}}} AS v
FROM range(3) t(k);

-- 1. drop a field from the nested object and reorder, expected 30
SELECT (v.ms::STRUCT(b VARCHAR, a VARCHAR)).b FROM nested ORDER BY v.id LIMIT 1;

-- 2. the other field of the same cast, expected 200
SELECT (v.ms::STRUCT(b VARCHAR, a VARCHAR)).a FROM nested ORDER BY v.id LIMIT 1;

-- 3. drop a field but keep the original order, expected 30
SELECT (v.ms::STRUCT(a VARCHAR, b VARCHAR)).b FROM nested ORDER BY v.id LIMIT 1;

-- 4. a numeric cast target, expected 30
SELECT (v.ms::STRUCT(b BIGINT, a BIGINT)).b FROM nested ORDER BY v.id LIMIT 1;

-- 5. one level deeper, expected 7
SELECT (v.ms::STRUCT(c STRUCT(d VARCHAR), b VARCHAR, a VARCHAR)).c.d
FROM nested ORDER BY v.id LIMIT 1;

-- 6. the same shape over VARIANT, expected 30
CREATE TABLE nested_variant AS SELECT
    {'id': k + 1000, 'ms': {'a': k + 200, 'b': k + 30}}::VARIANT AS v
FROM range(3) t(k);

SELECT (v.ms::STRUCT(b VARCHAR, a VARCHAR)).b
FROM nested_variant ORDER BY v.id::BIGINT LIMIT 1;

-- 7. the same query as 1 without late materialization, expected 30
SET disabled_optimizers = 'LATE_MATERIALIZATION';
SELECT (v.ms::STRUCT(b VARCHAR, a VARCHAR)).b FROM nested ORDER BY v.id LIMIT 1;
SET disabled_optimizers = '';
```

Output on v1.5.5:

```
1  returned  1000                                           expected  30    <- read v.id
2  returned  {'a': 200, 'b': 30, 'c': {'d': 7}}    expected  200   <- read v.ms
3  returned  {'a': 200, 'b': 30, 'c': {'d': 7}}    expected  30    <- read v.ms
4  returned  1000                                           expected  30    <- read v.id
5  returned  INTERNAL Error: PushdownExtract not supported for StatisticsType::NUMERIC_STATS
                                                                         expected  7
6  returned  NULL                                          expected  30
7  returned  30                                               expected  30    <- correct, optimizer disabled
```

The field is looked up by its ordinal in the cast type, but that ordinal is applied to the root column `v` instead of to `v.ms`:
| asked for | ordinal in the cast type | that ordinal of `v` | returned |
|---|---|---|---|
| `(v.ms::STRUCT(b VARCHAR, a VARCHAR)).b` | 0 | `v.id` | `1000` |
| `(v.ms::STRUCT(b VARCHAR, a VARCHAR)).a` | 1 | `v.ms` | the whole `ms` struct |

### OS:

macOS 26.6.1

### DuckDB Version:

1.5.5

### DuckDB Client:

CLI

### Hardware:

Mac 

### Full Name:

Florian Valeye

### Affiliation:

altertable-ai

- [x] Yes, I have

- [x] Yes, I have

fvaleye:
Fixes #24657.

Reading a field out of a nested struct through a `CAST` returns a different field:
```sql
CREATE TABLE nested AS SELECT
    {'id': k + 1000, 'ms': {'a': k + 200, 'b': k + 30}} AS v
FROM range(3) t(k);

SELECT (v.ms::STRUCT(b VARCHAR, a VARCHAR)).b FROM nested ORDER BY v.id LIMIT 1;
-- 1000, expected 30
```

It comes down to the pruner running twice:`LateMaterialization` and `TopNWindowElimination` run `RemoveUnusedColumns` again on a plan it has already rewritten. 
The extract walk stops at the cast, so the first run pushes the `v.ms` prefix into the scan and leaves a `struct_extract` above it. The second run treats that leftover as fresh, and `CheckPushdownExtract` reads `column_id.GetPrimaryIndex()`, which is still the root column. 
`b` is ordinal 0 of the cast type, and ordinal 0 of `v` is `id`, so the scan reads `id`.

`EXPLAIN ... PHYSICAL_ONLY` shows it without reasoning about values:
```
│ Expressions: v, v         │
│ Projections: v.id, v.id   │
```

After, the scan stays narrowed to `v.ms` and the leftover extract stays above it:
```
│ Expressions: struct_extract(v, 'b'), v │
│ Projections: v.ms, v.id                │
```

## The fix
Pushdown extract is disabled for a column that is already a `PushdownExtract`. The narrowed column stays in place, and the projection above does the remaining extraction, which is what already happens for every other shape that cannot be pushed.

This gives up no pushdown: running the pruner once, with `SET disabled_optimizers = 'LATE_MATERIALIZATION'`, already produces that same plan.

## Tests
- `test/sql/optimizer/pushdown_extract_reapplied.test`: every field of the fixture holds a distinct value so that reading the wrong one cannot produce the expected result, which is how the plain `1000` in the reproducer names the field it misread.

### DIFF
--- a/src/optimizer/remove_unused_columns.cpp
+++ b/src/optimizer/remove_unused_columns.cpp
@@ -903,6 +903,11 @@ void RemoveUnusedColumns::CheckPushdownExtract(LogicalOperator &op) {
 				continue;
 			}
 			auto &column_id = get.GetColumnIndex(column_binding);
+			if (column_id.IsPushdownExtract()) {
+				//! We already pushed down an extract for this column, so a new path is relative to that
+				col.supports_pushdown_extract = PushdownExtractSupport::DISABLED;
+				continue;
+			}
 			auto logical_column_index = LogicalIndex(column_id.GetPrimaryIndex());
 			if (!get.function.supports_pushdown_extract || get.function.statistics) {
 				//! Either 'statistics_extended' needs to be set or 'statistics' needs to be NULL
@@ -1042,7 +1047,8 @@ void RemoveUnusedColumns::RemoveColumnsFromLogicalGet(LogicalGet &get, unique_pt
 			if (logical_column_id.IsPushdownExtract()) {
 				//! RemoveUnusedColumns is also used by other optimizers,
 				//! so we have to deal with this case and preserve the PushdownExtract we created earlier
-				D_ASSERT(entry->second.child_columns.empty());
+				D_ASSERT(entry->second.child_columns.empty() ||
+				         entry->second.supports_pushdown_extract == PushdownExtractSupport::DISABLED);
 				new_column_ids.emplace_back(logical_column_id);
 			} else {
 				ColumnIndex new_index(logical_column_id.GetPrimaryIndex(), entry->second.child_columns);
--- a/test/configs/storage_compatibility.json
+++ b/test/configs/storage_compatibility.json
@@ -148,6 +148,7 @@
         "test/sql/storage/types/variant/variant_stats_merging.test_slow",
         "test/parquet/variant/variant_legacy_encoding.test",
         "test/sql/types/variant/tpch_test.test",
+        "test/sql/optimizer/pushdown_extract_reapplied.test",
         "test/sql/storage/compression/default_storage_dict_fsst.test",
         "test/sql/storage/compression/roaring/roaring_bool_array_simple.test",
         "test/sql/storage/compression/roaring/roaring_bool_array_simple_w_null.test",
--- a/test/sql/optimizer/pushdown_extract_reapplied.test
+++ b/test/sql/optimizer/pushdown_extract_reapplied.test
@@ -1,0 +1,110 @@
+# name: test/sql/optimizer/pushdown_extract_reapplied.test
+# description: A pushdown-extract column must not be narrowed a second time by a later pruner run
+# group: [optimizer]
+
+statement ok
+SET late_materialization_max_rows = 100;
+
+# Every field holds a distinct value, so reading the wrong one cannot produce the expected result
+statement ok
+CREATE TABLE nested_extract AS SELECT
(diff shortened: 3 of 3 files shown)


Mytherin:
Thanks!

Issue #18448, opened 2025-07-30
Fuzzer `parse_fuzz_test` crashes with `Calling GetValueInternal on a value that is NULL`


gal1ium:
### What happens?

Hi! We found a crashing PoC by the fuzzing driver `parse_fuzz_test` with the output:
```
ABORT THROWN BY INTERNAL EXCEPTION: Calling GetValueInternal on a value that is NULL

Aborted (core dumped)
```

### To Reproduce

```
PRAGMA enable_progress_bar=null
```

### OS:

Linux

### DuckDB Version:

041b68304fee39b0eb770f745d1b282b7f9fe004

### DuckDB Client:

fuzzer

### Full Name:

gal1ium

### Affiliation:

HKU

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a source build

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [x] Yes, I have

PR #24661, opened 2026-08-10
Fix PRAGMA tpch/tpcds throwing INTERNAL error on NULL argument

### LINKED ISSUE #24603: PRAGMA tpch(NULL) / tpcds(NULL) throws INTERNAL Error: Calling GetValueInternal on a value that is NULL
### What happens?

`PragmaTpchQuery` in `extension/tpch/tpch_extension.cpp` reads the query-number argument by calling `Value::GetValue<int32_t>()` on the pragma's parsed constant without first checking whether that constant is NULL:

- `extension/tpch/tpch_extension.cpp:241` — `auto index = parameters.values[0].GetValue<int32_t>();`

`tpcds` has the identical defect:

- `extension/tpcds/tpcds_extension.cpp:242` — `auto index = parameters.values[0].GetValue<int32_t>();`

`Value::GetValue` rejects a NULL value by throwing an `InternalException` (`Calling GetValueInternal on a value that is NULL`). The pragma grammar accepts `NULL` wherever the call form takes a literal argument, so `PRAGMA tpch(NULL)` reaches that throw and surfaces an internal error to the user.

This is a missing input validation rather than a defect in the pragma's semantics: NULL is a legal literal at that grammar position, so it should be rejected with a binder error naming the argument. The correct behavior already exists on the neighbouring named-parameter path in the same extension — `CALL dbgen(sf=NULL)` returns a plain `Binder Error: Cannot use NULL as function argument`, and the value-checking guard for that path is right there in the same file (`tpch_extension.cpp:60`, `if (kv.second.IsNull())`). Only the positional pragma-call path skips the check.

### To Reproduce

In a fresh in-memory database, with no prior state and no data generated:

```sql
PRAGMA tpch(NULL);
```

```sql
PRAGMA tpcds(NULL);
```

Both reproduce deterministically. `PRAGMA tpch(1)` on an empty database instead returns the ordinary `Catalog Error: Table with name lineitem does not exist!`, confirming the NULL is what reaches the unchecked `GetValue`, not the absence of TPC-H data.

### Result

```
INTERNAL Error: Calling GetValueInternal on a value that is NULL

```

Expected: a regular binder error naming the argument, for example `Binder Error: TPC-H query number must not be NULL`, matching how `CALL dbgen(sf=NULL)` already reports `Binder Error: Cannot use NULL as function argument`.

For contrast, other positional `PRAGMA` calls already reject a bad argument at the binder without asserting:

```sql
PRAGMA database_size(NULL);   -- Binder Error: No function matches ...
PRAGMA show_tables(NULL);     -- Binder Error: No function matches ...
```

### Relationship to #18448

#18448 reports the same `Calling GetValueInternal on a value that is NULL` assertion via a different `PRAGMA` form — the assignment form `PRAGMA enable_progress_bar=NULL`. This report is the positional call form `PRAGMA f(NULL)` reaching an unchecked `GetValue` inside the tpch/tpcds extensions specifically, which is a distinct code path (`PragmaTpchQuery` / `PragmaTpcdsQuery`) from the setting-assignment path in #18448. They may share a fix if the guard is added at the pragma-binding layer; if fixed per-site, this names two more sites.

### OS:

Ubuntu 22.04, x86_64 (Linux)

### DuckDB Version:

commit `7255424293982d7ec53b0d72b73a4b56c1c4ce0c`

### DuckDB Client:

CLI (Command Line Interface), built from source (release build), tpch and tpcds extensions statically linked

### Hardware:

_No response_

### Full Name:

_No response_

### Affiliation:

_No response_

Yes

- [x] Yes, I have

- [x] Yes, I have

---

alwaysprince05:
Fixes #24603

`PRAGMA tpch(NULL)` and `PRAGMA tpcds(NULL)` reached `Value::GetValue<int32_t>()` on a NULL constant, asserting with `INTERNAL Error: Calling GetValueInternal on a value that is NULL`.

This adds `IsNull()` guards in `PragmaTpchQuery` and `PragmaTpcdsQuery`, throwing a `BinderException` naming the argument — matching the existing `CALL dbgen(sf=NULL)` behavior in the same files (`Binder Error: Cannot use NULL as function argument`).

I kept the guard per-site rather than at the pragma-binding layer (which would also cover the assignment-form issue #18448) to keep the change minimal; happy to move it if a general fix is preferred.

Regression tests: `test/sql/tpch/tpch_null_arg.test` and `test/sql/tpcds/tpcds_null_arg.test`. Full tpch (25 cases) and tpcds (13 cases) suites pass.

### DIFF
--- a/extension/tpcds/tpcds_extension.cpp
+++ b/extension/tpcds/tpcds_extension.cpp
@@ -239,6 +239,9 @@ static void TPCDSQueryAnswerFunction(ClientContext &context, TableFunctionInput 
 }
 
 static string PragmaTpcdsQuery(ClientContext &context, const FunctionParameters &parameters) {
+	if (parameters.values[0].IsNull()) {
+		throw BinderException("TPC-DS query number must not be NULL");
+	}
 	auto index = parameters.values[0].GetValue<int32_t>();
 	return tpcds::DSDGenWrapper::GetQuery(index);
 }
--- a/extension/tpch/tpch_extension.cpp
+++ b/extension/tpch/tpch_extension.cpp
@@ -238,6 +238,9 @@ static void TPCHQueryAnswerFunction(ClientContext &context, TableFunctionInput &
 }
 
 static string PragmaTpchQuery(ClientContext &context, const FunctionParameters &parameters) {
+	if (parameters.values[0].IsNull()) {
+		throw BinderException("TPC-H query number must not be NULL");
+	}
 	auto index = parameters.values[0].GetValue<int32_t>();
 	return tpch::DBGenWrapper::GetQuery(index);
 }
--- a/test/sql/tpcds/tpcds_null_arg.test
+++ b/test/sql/tpcds/tpcds_null_arg.test
@@ -1,0 +1,10 @@
+# name: test/sql/tpcds/tpcds_null_arg.test
+# description: PRAGMA tpcds rejects a NULL query number instead of throwing an INTERNAL error
+# group: [tpcds]
+
+require tpcds
+
+statement error
+PRAGMA tpcds(NULL);
+----
+<REGEX>:.*Binder Error.*TPC-DS query number must not be NULL.*
--- a/test/sql/tpch/tpch_null_arg.test
+++ b/test/sql/tpch/tpch_null_arg.test
@@ -1,0 +1,10 @@
+# name: test/sql/tpch/tpch_null_arg.test
+# description: PRAGMA tpch rejects a NULL query number instead of throwing an INTERNAL error
+# group: [tpch]
+
+require tpch
+
+statement error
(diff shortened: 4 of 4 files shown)


lnkuiper:
Thanks, nice find! Can you resolve the merge conflicts?

alwaysprince05:
Thanks @lnkuiper!

While resolving the merge conflicts, I noticed upstream already merged a NULL guard for these PRAGMAs in b4b5bd5924 (`InvalidInputException` with regression tests in `tpch_sf0.test` / `tpcds_sf0.test`). The core fix for #24603 is already covered on `main`, so I'm closing this PR as superseded.

Happy to contribute a follow-up if you'd prefer different error wording or additional test coverage.

PR #24662, opened 2026-08-10
Add one-way settings gating the EXTERNAL RESOURCE feature

carlopi:
Two global settings, both of which can only ever be made more restrictive:

```sql
SET external_resources_mode = 'available' | 'on'   -- the default, nothing gated
SET external_resources_mode = 'listing'            -- no create/register/destroy
SET external_resources_mode = 'off' | 'none'       -- no listing either
SET external_resources_type_registration = false   -- no new resource types
```

Raising either of them errors, and RESET is refused once lowered, so an operator can hand a session to less-trusted code without it being able to undo the restriction. Both are non-generic settings for that reason: a setting with a `default_value` lives in DBConfig::user_settings, where RESET is a bare ClearSetting with no callback and would restore the default.

The modes are an ENUM<ExternalResourcesMode> ordered by increasing restriction, so "degrade only" is simply "the new value must not compare less than the current one".

Callers name a capability rather than a setting: ExternalResources::RequireCapability maps LISTING / MANAGEMENT / TYPE_REGISTRATION onto the settings that govern them, so a future knob is added there instead of at the call sites. TYPE_REGISTRATION requires the full mode, not just LISTING: a type's `list` callback is invoked by discovery, which LISTING still allows, so registering a type would otherwise be a way to get an arbitrary function called under a read-only mode.

Every SQL entry point is gated twice - at bind time, so the failure is early and clear, and again where the work happens, because a statement prepared before the setting was lowered would otherwise still run. The in-process C++ APIs (the manager, the type registry) stay ungated so internal teardown keeps working.

### DIFF
--- a/src/include/duckdb/main/config.hpp
+++ b/src/include/duckdb/main/config.hpp
@@ -15,6 +15,7 @@
 #include "duckdb/common/common.hpp"
 #include "duckdb/common/enums/access_mode.hpp"
 #include "duckdb/common/enums/cache_validation_mode.hpp"
+#include "duckdb/common/enums/external_resources_mode.hpp"
 #include "duckdb/common/enums/thread_pin_mode.hpp"
 #include "duckdb/common/enums/compression_type.hpp"
 #include "duckdb/common/enums/optimizer_type.hpp"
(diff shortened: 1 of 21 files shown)


src/include/duckdb/main/config.hpp
```diff
@@ -144,6 +145,10 @@ struct DBConfigOptions {
 	set<string> allowed_directories;
 	//! Additional configuration options that are allowed to be changed even when the configuration is locked
 	identifier_set_t allowed_configs;
+	//! Which external resource operations are permitted - degrades one way only
+	ExternalResourcesMode external_resources_mode = ExternalResourcesMode::AVAILABLE;
+	//! Whether new external resource types can be registered - degrades one way only
+	bool external_resources_type_registration = true;
```
samansmink, line 151: I feel like this should be an allowlist of desired types instead?

carlopi:
Unsure on semantic, to be re-discussed.

PR #24668, opened 2026-08-10
Fix INTERNAL error when the invoke() lambda throws

### LINKED ISSUE #24644: invoke() scalar function not marked fallible → lambda runtime error surfaces as INTERNAL Error (invoke.cpp:132)
> Same defect class as #24616 (`list_reduce`), #24617 (`__cast` TIMESTAMPTZ→TIMESTAMP_NS), and #24622 (`regexp_full_match`): a function that can raise a runtime error is not registered as fallible, so the error surfaces as an `INTERNAL Error`. This is a **distinct function** (`invoke`) with its own one-line fix, so filing separately. The sibling lambda functions `list_transform`/`list_filter` avoid it by calling `SetFallible()` in their registrations; `invoke` (like `list_reduce`) does not.
>
> Confirmed still present on the current `main` HEAD: `InvokeFun::GetFunction()` (`src/function/scalar/generic/invoke.cpp:132`) has no `SetFallible()` call before `return fun;`. `invoke` was added recently by merged PR #23185.

### What happens?

The `invoke` scalar function executes a user-supplied lambda but is **not registered as fallible**, so any runtime error thrown while evaluating the lambda body is reported as an `INTERNAL Error` (a self-described assertion failure) instead of the ordinary runtime error.

`invoke(lambda ..., args...)` evaluates the lambda body through an `ExpressionExecutor` in `LambdaInvokeFunction` (`src/function/scalar/generic/invoke.cpp:86`, `lstate.executor->ExecuteExpression(...)` at `:95`). The function's registration `InvokeFun::GetFunction()` (`src/function/scalar/generic/invoke.cpp:132`) installs bind / lambda / init-state / serialize callbacks but **never calls `fun.SetFallible()`**. When the lambda body raises a normal runtime error (a failed cast, an explicit `error(...)`, etc.), DuckDB's execution layer detects that a function not marked fallible threw and rewrites the error into:

```
INTERNAL Error: Scalar function "invoke" threw an execution error, but the function is not marked as fallible - the function must call SetFallible(). Error: <original message>
```

For comparison, the sibling lambda functions call `SetFallible()` in their registrations (`extension/core_functions/scalar/list/list_transform.cpp`, `list_filter.cpp`), so the same lambda error propagates cleanly through them; `invoke` (like `list_reduce`, #24616) does not.

### To Reproduce

Fresh in-memory database, no prior state. Any lambda body that raises a runtime error:

```sql
SELECT invoke(lambda x: x::INTEGER, 'abc');   -- failed cast
SELECT invoke(lambda x: error('boom'), 1);    -- explicit error()
```

A lambda that does not throw behaves normally: `SELECT invoke(lambda x: x + 1, 5);` returns `6`.

### Result

```
INTERNAL Error: Scalar function "invoke" threw an execution error, but the function is not marked as fallible - the function must call SetFallible(). Error: Could not convert string 'abc' to INT32 when casting from source column x
```

Reproduces on both a release build and an assertion-enabled build. Expected: the lambda's runtime error should propagate as an ordinary error (e.g. `Conversion Error: Could not convert string 'abc' to INT32`), exactly as that error propagates outside `invoke`. The fix is a one-liner — `fun.SetFallible();` in `InvokeFun::GetFunction()` — matching `list_transform`/`list_filter`.

### OS:

### DuckDB Version:

commit `e500d77864fe565f90e68f06d729c25b11e775c5` (main branch)

### DuckDB Client:

CLI, built from source with core extensions statically linked; reproduces on both `CMAKE_BUILD_TYPE=Release` and an assertion-enabled build, `clang++`.

---

saime428:
Fixes #24644.

`invoke()` evaluates a user-supplied lambda through an `ExpressionExecutor` (`LambdaInvokeFunction`, `src/function/scalar/generic/invoke.cpp:86`), but `InvokeFun::GetFunction()` never calls `SetFallible()`. A runtime error raised by the lambda body is therefore rewritten into an internal assertion failure instead of propagating as the ordinary error:

```
INTERNAL Error: Scalar function "invoke" threw an execution error, but the function is not marked as fallible - the function must call SetFallible(). Error: Could not convert string 'abc' to INT32 when casting from source column x
```

The sibling lambda functions already set the flag (`list_transform.cpp:42`, `list_filter.cpp:50`), so the same error propagates cleanly through them. This looks like a case #24389 missed.

## Changes

- `invoke.cpp`: call `fun.SetFallible()` in the registration, matching `list_transform` / `list_filter`.
- `invoke.test`: three regression cases, one per distinct error source.

After the change:

| trigger | error |
| --- | --- |
| failed cast | `Conversion Error: Could not convert string 'abc' to INT32 ...` |
| explicit `error()` | `Invalid Input Error: boom` |
| arithmetic overflow | `Out of Range Error: Overflow in multiplication of INT64 ...` |

A lambda that does not throw is unaffected: `invoke(lambda x: x + 1, 5)` still returns `6`.

## Validation

- [x] `test/sql/function/generic/invoke.test` — 33 assertions, passing. The three new cases fail on unpatched main, where the INTERNAL Error replaces the expected message.
- [x] `test/sql/function/list/lambdas/*` — 34 test cases, 1552 assertions, all passing.
- [x] Fast unit tests (`build/reldebug/test/unittest`) — 4886 test cases, 1039105 assertions, all passing.
- [x] `clang-format --dry-run -Werror` on the changed source file.
- [ ] `make allunit` — not run locally.


saime428:
@Mytherin could you take a look at this one when you get a chance? It has had no review since it was opened on 2026-08-10.

Different bug from #24652, which you closed — I agreed with that reasoning and this is not a retry of it. Here, a lambda passed to `invoke()` that throws surfaces as an INTERNAL error instead of the thrown error itself. The fix is 19 lines across 2 files, with a test.

Re-checked today: still `MERGEABLE` against `main`, checks green.

Mytherin:
Thanks!

taniabogatsch:
Hi @saime428 - looks like there is an issue with the CI and the runner, could you merge with the latest changes on `main` to re-trigger CI please?

saime428:
Hi, of course @taniabogatsch

taniabogatsch (approved):

PR #24681, opened 2026-08-11
Use __restrict actually in MacOS

flashmouse:
This branch initially tried to improve bitpack decompress performance. After a long trek, I found that ``__restrict`` doesn't actually take effect on macOS, causing DuckDB to lose some compiler optimizations.

On macOS, ``cdefs.h`` has the following macro definition:
```
#if __STDC_VERSION__ < 199901
#define __restrict
#else
#define __restrict      restrict
#endif
```

On Red Hat-based servers, ``cdefs.h`` has the following macro definition:
```
/* __restrict is known in EGCS 1.2 and above. */
#if !__GNUC_PREREQ (2,92)
# if defined __STDC_VERSION__ && __STDC_VERSION__ >= 199901L
#  define __restrict    restrict
# else
#  define __restrict    /* Ignore */
# endif
#endif
```

This means that on Red Hat, GCC >= 2.92 will keep ``__restrict`` as-is, while on macOS it just gets dropped. This PR makes ``__restrict`` an alias for ``__restrict__`` on macOS globally, which I believe is safe.
As a result, some benchmarks show improvement.

This branch:
```
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    1       0.019214
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    2       0.018318
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    3       0.017913
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    4       0.018854
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    5       0.017880
```

main branch:
```
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    1       0.027011
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    2       0.026990
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    3       0.026648
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    4       0.024557
benchmark/micro/compression/bitpacking/bitpacking_read_for.benchmark    5       0.026475
```


lnkuiper:
Great find, and nice performance gains! We prefer not to have the verbose comments explaining what's going on. Also, why is there an undef/define? Can't it just be defined in the same if/else? Also, why is it defined twice?

flashmouse:
> Great find, and nice performance gains! We prefer not to have the verbose comments explaining what's going on. Also, why is there an undef/define? Can't it just be defined in the same if/else? Also, why is it defined twice?

if remove ``undef`` compile would throw warning:
```
/Users/lixy/code/duckdb_work/third_party/fastpforlib/bitpacking.h:18:9: warning: '__restrict' macro redefined [-Wmacro-redefined]
   18 | #define __restrict __restrict__
      |         ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h:264:9: note: previous definition is here
  264 | #define __restrict
      |         ^
1 warning generated.
[384/384] repository
```

I write twice because fastpforlib is in third-party and it doesn't include duckdb's head file. or duckdb has any file could define global macro I haven't find so I can remove redundant definition?

lnkuiper:
I see. I think I'd prefer to have this centralised in a `duckdb/common/restrict.hpp` and then have that included in some other commonly included `hpp` file. That would resolve this issue over the entire code base. I can see that, for example, `fsst` (also in `third_party`) defines this thing too. I'll merge this PR now, it's just a minor code style thing that can be resolved later.

Thanks!

PR #24670, opened 2026-08-10
Bump aws, absorb patches

carlopi:
Also adds listing for `quack-on-ec2`

PR #24682, opened 2026-08-11
bump aws and delete patches

Tishj:


Tishj:
Same as #24670

carlopi:
And I guess they will also fail in the same way due to iceberg being loaded... unsure what's the path forward, skip tests for aws? or patch to remove that test?

PR #24685, opened 2026-08-11
Compare window partitions as sets

### LINKED ISSUE #24629: Duplicate PARTITION BY expressions can make different window partitions share results
### What happens?

Two window aggregates with different `PARTITION BY` clauses can share the wrong physical partitioning when one clause contains a duplicate expression. In this repro, `PARTITION BY a, a` should group both rows together, while `PARTITION BY a, b` should keep them separate. DuckDB returns the wider `a` partition result for both window functions.

This happens on current main (`e500d77864fe`) and also on v1.5.4.

#### Code

```sql
WITH src(a, b, v) AS (
  VALUES (1, 1, 10), (1, 2, 20)
)
SELECT
  SUM(v) OVER (PARTITION BY a, a) AS w_aa,
  SUM(v) OVER (PARTITION BY a, b) AS w_ab
FROM src
ORDER BY b;
```

Run:

```sh
```

#### Expected

| w_aa | w_ab |
| --- | --- |
| 30 | 10 |
| 30 | 20 |

#### Actual

| w_aa | w_ab |
| --- | --- |
| 30 | 30 |
| 30 | 30 |

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (`e500d77864fe`)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

m0g3r:
## Description

Fixes #24629.

Two window functions with different `PARTITION BY` clauses could share one physical
partitioning when one of the clauses contained a duplicate expression:

```sql
WITH src(a, b, v) AS (VALUES (1, 1, 10), (1, 2, 20))
SELECT SUM(v) OVER (PARTITION BY a, a) AS w_aa,
       SUM(v) OVER (PARTITION BY a, b) AS w_ab
FROM src ORDER BY b;
```

`w_ab` should be `10` and `20`, but both columns returned `30`.

`BoundWindowExpression::PartitionsAreEquivalent` compared the two partition lists by
vector length while testing membership through an `expression_set_t`, and it only
checked one direction. For `PARTITION BY a, a` against `PARTITION BY a, b` the lengths
both equal two, so the early return does not fire; the duplicate then collapses inside
the set, so every element of `[a, a]` is found in `{a, b}` and the function reports the
partitionings as equivalent. `PhysicalPlanGenerator::CreatePlan(LogicalWindow)` uses that
result to decide which window expressions may share a partitioning, and the first
expression defines it, so `PARTITION BY a, b` was evaluated over the wider `a` partition.

Partitioning is insensitive to both ordering and duplication, so the fix deduplicates
each side and compares the resulting sets, which also resolves the `TODO` that sat
directly above the comparison.

## Additional context and related issues

`PARTITION BY a, a` remains equivalent to `PARTITION BY a`, since both describe the same
grouping and should continue to share a partitioning. The added test covers that case
alongside the reported one, the reported query with the clause order reversed, and a
three-expression variant, because the operator that defines the shared partitioning is
whichever window expression comes first.

The new test fails on an unpatched build and passes with the change. `test/sql/window/*`
passes, as does the full `test/unittest` suite. On one earlier run of the full suite
`test/sql/storage/compression/fsst/fsst_greedy_long_strings.test` failed; it passes on
its own and on a repeat of the full suite, and it is not reachable from this change.

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
(x) Release notes are required, with the following suggested text:

```markdown
## Bug fixes
* Fix window functions with a duplicated `PARTITION BY` expression sharing a partitioning with a different `OVER` clause. ({issue}`24629`)
```


hawkfish (approved):
Thanks for the PR - LGTM!

PR #24688, opened 2026-08-11
Disable flaky FSST test

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/10297 unblocking CI

@gropaul


gropaul:
Thanks for the fix, sorry for the trouble!

The test is probably flaky because the FSST symbol table creation is non-deterministic; I will check on it!

PR #24691, opened 2026-08-11
EXTERNAL RESOURCE: require WITH before recipe create params

carlopi:
```sql
CREATE EXTERNAL RESOURCE '<type>' AS y (SPOT, REGION 'eu-west-1')
---becomes
CREATE EXTERNAL RESOURCE '<type>' AS y WITH (SPOT, REGION 'eu-west-1')
```

The clause stays optional; WITH is only required once there are params.

This allows to disambiguate once introducing:
```sql
ATTACH TO TEMPORARY EXTERNAL RESOURCE '<type>' [WITH (...create options)] [AS <name>] [(...attach options)];
```
examples, before:
```
ATTACH TO TEMPORARY EXTERNAL RESOURCE '<type>' (...options);
```
would be unclear (either external resource option OR attach option), while after this it's clear (it's there a `WITH`?).
Also cleans up using `AttachOptions` as the name of the bag of options, making code more clear.

The rule is added here rather than alongside the statement that needs it, so that it is defined next to the other EXTERNAL RESOURCE grammar.


src/include/duckdb/parser/peg/inlined_grammar.gram
```diff
@@ -1528,12 +1528,15 @@ ExtractDatePart <-
...
 ExternalResourceStatement <- CreateExternalResourceStmt / RegisterExternalResourceStmt / DestroyExternalResourceStmt / ShowExternalResourcesStmt
 
-CreateExternalResourceStmt <- 'CREATE' 'EXTERNAL' 'RESOURCE' StringLiteral AttachAlias? AttachOptions?
+CreateExternalResourceStmt <- 'CREATE' 'EXTERNAL' 'RESOURCE' StringLiteral AttachAlias? ExternalResourceOptions?
 RegisterExternalResourceStmt <- 'REGISTER' 'EXTERNAL' 'RESOURCE' StringLiteral AttachAlias? 'FROM' Expression
 DestroyExternalResourceStmt <- 'DESTROY' 'EXTERNAL' 'RESOURCE' ColId
 ShowExternalResourcesStmt <- 'SHOW' ShowAllModifier? 'EXTERNAL' 'RESOURCES'
 ShowAllModifier <- 'ALL'
 
+# WITH marks recipe create params, as opposed to the bare option list of the DuckDB-side statement.
```
Dtenwolde, line 1537: Not sure I understand this comment, but essentially this is something future-proofing?

Dtenwolde (commented):
Minor comment to clarify

Dtenwolde:
Regarding: `ATTACH TO TEMPORARY EXTERNAL RESOURCE '<type>' [WITH (...create options)] [AS <name>] [(...attach options)];`

I would keep the alias positioning consistent, so before the `WITH (options)`. But if we have two sets of options (create options and attach options) it would perhaps also be cleaner to make this more explicit. To avoid `ATTACH ... AS foo WITH (<set of create options>) (<set of attach options>)`

carlopi:
Thanks @Dtenwolde, we chatted live, and indeed making attach name compulsory is the way here. Closing

PR #24692, opened 2026-08-11
Reuse unchanged persistent segments during checkpoint

Liuzw1998:
Hi DuckDB team,

Motivated by a workload we encounter in AliSQL DuckDB, we noticed that checkpointing can introduce significant latency on instances that replicate data from InnoDB into DuckDB. We therefore looked into reducing work during checkpoint. It would be nice to discuss it with you for further improvements.

## Motivation

Updating any value in a persistent base column currently makes checkpoint decode, analyze, recompress, and write the entire column of that RowGroup. Sparse updates therefore repeat work for persistent Segments whose payload has not changed. A validity-only update also rewrites the unchanged base payload.

This patch lets checkpoint reference eligible unchanged persistent Segments and only rewrite contiguous ranges that contain changes. It preserves the existing whole-column path whenever the layout is unsupported or the expected benefit is too small.

## Design

Adjacent dirty Segments are combined into maximal contiguous ranges. Each dirty range uses one range-local
Analyze/compression pipeline, while each eligible clean Segment is referenced without decoding or rewriting its
payload. 

Multiple ranges add pipeline setup, finalization, and codec-specific Analyze work. Recompression can also produce
more or smaller Segments than a whole-column rewrite, making the persistent layout less compact over time. The local path is therefore selected only when all of these conditions hold:

- at least 12.5% of tuples can be reused;
- there are no more than four dirty ranges;
- total reuse satisfies both one eighth of the column and five Vectors per additional Analyze/compression pipeline;
- the estimated Segment count is at most 25% above an estimated compact layout.

The admission checks run before Analyze, Block marking, or result-tree writes. Failure therefore returns to the
existing whole-column rewrite without local-path side effects. Nested columns, row-remapping checkpoints, unavailable or globally forced compression layouts, and unsupported validity layouts also retain the original path.

The local path also requires an independently persisted validity layout. Some codecs encode NULL information in the
base Segment, while `COMPRESSION_EMPTY` validity layouts have no independent persistent NULL bitmap. A rewritten range may select a codec that requires separate validity, so both layouts conservatively use whole-column rewriting.

## Evaluation

All values are medians. The matrix contains 10 cases and 120 timed checkpoints; all 60 Candidate samples selected
the local path.

| Codec | Actual reuse | Baseline wall (ms) | Candidate wall (ms) | Wall speedup |
|---|---:|---:|---:|---:|
| FSST | 25.4% | 157.5 | 116.5 | 1.352x |
| FSST | 49.3% | 161.0 | 81.5 | 1.975x |
| FSST | 74.7% | 163.5 | 42.0 | 3.893x |
| FSST | 90.6% | 160.5 | 19.0 | 8.447x |
| FSST | 98.6% | 162.0 | 7.0 | 23.143x |
| Uncompressed | 25.4% | 51.5 | 36.0 | 1.431x |
| Uncompressed | 49.7% | 55.0 | 24.0 | 2.292x |
| Uncompressed | 75.1% | 59.0 | 13.0 | 4.538x |
| Uncompressed | 89.7% | 61.0 | 6.0 | 10.167x |
| Uncompressed | 99.4% | 63.0 | 1.0 | 63.000x |
| **Overall** | **25.4%-99.4%** | | | **5.354x geometric mean** |

| Codec | Actual reuse | CPU speedup | Write reduction | Baseline peak RSS (MiB) | Candidate peak RSS (MiB) |
|---|---:|---:|---:|---:|---:|
| FSST | 25.4% | 1.358x | 25.0% | 95.5 | 83.4 |
| FSST | 49.3% | 1.982x | 48.4% | 98.6 | 69.9 |
| FSST | 74.7% | 3.907x | 73.4% | 98.3 | 52.8 |
| FSST | 90.6% | 8.528x | 89.0% | 93.8 | 43.6 |
| FSST | 98.6% | 23.743x | 95.3% | 91.5 | 37.2 |
| Uncompressed | 25.4% | 1.447x | 25.0% | 76.8 | 65.9 |
| Uncompressed | 49.7% | 2.280x | 48.8% | 76.5 | 55.7 |
| Uncompressed | 75.1% | 4.685x | 73.8% | 76.3 | 44.8 |
| Uncompressed | 89.7% | 9.943x | 88.1% | 76.2 | 38.6 |
| Uncompressed | 99.4% | 36.558x | 97.6% | 74.8 | 33.6 |
| **Overall** | **25.4%-99.4%** | | **25.0%-97.6%** |  | **Lower in 10/10 cases** |

## Limitations

- Projected Segment count is a pre-compression estimate, not a strict bound on codec output.
- Segment count does not directly measure stranded space inside shared Blocks.

### DIFF
--- a/src/include/duckdb/storage/table/column_checkpoint_state.hpp
+++ b/src/include/duckdb/storage/table/column_checkpoint_state.hpp
@@ -45,6 +45,8 @@ public:
 	virtual shared_ptr<ColumnData> GetFinalResult();
 
 	virtual unique_ptr<BaseStatistics> GetStatistics();
+	//! Incorporate a persistent Segment and its metadata into the checkpoint result
+	void AppendReferencedSegment(shared_ptr<ColumnSegment> segment, idx_t row_start);
 
 	virtual void FlushSegmentInternal(unique_ptr<ColumnSegment> segment, idx_t segment_size);
 	virtual void FlushSegment(unique_ptr<ColumnSegment> segment, BufferHandle handle, idx_t segment_size);
--- a/src/include/duckdb/storage/table/column_data_checkpointer.hpp
+++ b/src/include/duckdb/storage/table/column_data_checkpointer.hpp
@@ -68,14 +68,29 @@ public:
 	void FinalizeCheckpoint();
 
 private:
-	void ScanSegments(const std::function<void(Vector &)> &callback);
-	vector<CheckpointAnalyzeResult> DetectBestCompressionMethod();
+	struct SegmentReusePlan;
+
+	void ScanSegments(const ColumnData &column_data, const std::function<void(Vector &)> &callback);
+	void ScanSegment(const ColumnData &column_data, SegmentNode<ColumnSegment> &segment,
+	                 const std::function<void(Vector &)> &callback);
+	vector<CompressionType> PrepareCompressionMethods();
+	vector<unique_ptr<AnalyzeState>> InitAnalyze(idx_t checkpoint_state_idx);
+	void AnalyzeVector(idx_t checkpoint_state_idx, vector<unique_ptr<AnalyzeState>> &states, Vector &scan_vector);
+	CheckpointAnalyzeResult FinalizeAnalyze(idx_t checkpoint_state_idx, vector<unique_ptr<AnalyzeState>> states,
+	                                        CompressionType forced_method);
+	unique_ptr<SegmentReusePlan> TryBuildSegmentReusePlan(const vector<CompressionType> &forced_methods);
+	vector<CheckpointAnalyzeResult> AnalyzeFullColumn(const vector<CompressionType> &forced_methods);
+	void AnalyzeRewriteRanges(SegmentReusePlan &plan, const vector<CompressionType> &forced_methods);
+	bool TryWriteReusedSegments(const vector<CompressionType> &forced_methods);
 	void WriteToDisk();
+	void WriteFullColumn(vector<CheckpointAnalyzeResult> analyze_result);
+	void WriteRewriteRanges(SegmentReusePlan &plan);
+	void WriteValidity(CheckpointAnalyzeResult &analyze_result);
 	void WritePersistentSegments(ColumnCheckpointState &state);
-	bool HasChanges(ColumnData &col_data);
-	void InitAnalyze();
-	void DropSegments();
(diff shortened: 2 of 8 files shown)


Liuzw1998:
Hi @Mytherin, could you take a quick look at whether this PR is useful and worth maintaining upstream?

The main idea is to reuse unchanged persistent segments during checkpoint to reduce CPU and I/O for sparse updates. If this direction makes sense, I’m happy to keep improving and maintaining it.

Mytherin:
Thanks for the PR! I think the idea is certainly interesting. One worry I have here is that this can lead to sub-optimal compression, for example if we have segments like this:

```
Block 1 - 256KB
Block 2 - 256KB
Block 3 - 256KB
Block 4 - 40KB
```

Then we update a string in block 2 and checkpoint it, but the string is slightly bigger than our previous block, we will now end up with this situation:

```
Block 1 - 256KB  (re-used)
Block 2 - 256KB  (NEW)
Block 3 - 1 byte (NEW) 
Block 4 - 256KB  (re-used)
Block 5 - 40KB   (re-used)
```

I think a better approach might be to instead store the update deltas also during checkpoint, so that we can checkpoint updates without having to re-write segments in their entirety, similar to how deletes are stored. Perhaps that approach could be considered as an alternative?

Liuzw1998:
@Mytherin Thanks! Your alternative method makes sense to me. 

I also observed your concerns about sub-optimal compression. My current implementation uses heuristics (which is imperfect) to **detect fragmentation**. Whenever too much fragmentation is detected, it falls back to a full rewrite. Could this way of "conditioned full rewrite" be further polished or you see persisting update deltas as the preferred long-term direction?

Mytherin:
I think update deltas are more broadly applicable, the segment re-use only works for larger types (e.g. for numeric values generally a whole row group will fit onto a single segment, so we're always rewriting those still).

Liuzw1998:
> I think update deltas are more broadly applicable, the segment re-use only works for larger types (e.g. for numeric values generally a whole row group will fit onto a single segment, so we're always rewriting those still).

Thanks! This direction is interesting and valuable to me. Is DuckDB team already working on persisting update deltas, or would it be useful if I open a draft/prototype for that direction? I would be happy to work on that next month :)    I guess this PR can be closed, right?

Mytherin:
We're not working on it so feel free to pick it up. I would say ideally they would be stored using existing infrastructure so they can also be compressed.

PR #24694, opened 2026-08-11
Exempt the buffer manager's temporary files from disabled_filesystems when the temp directory is locked

rubenfiszel:
## The problem

An embedded engine that runs **untrusted SQL in-process** has no way to say "you may spill to
disk, but you may not touch the filesystem for anything else".

`disabled_filesystems='LocalFileSystem'` is the natural fence, and it works well for user file
access. But the buffer manager's temporary files reach the disk through the same virtual file
system, so the fence takes out-of-core execution with it: any query whose working set outgrows
`memory_limit` fails with a permission error instead of spilling. The sandbox and the ability
to run large queries become mutually exclusive.

Nothing in the existing configuration bridges that gap:

- `allowed_directories` / `allowed_paths` enforce nothing while `enable_external_access` is
  true — `DBConfig::CanAccessFile` returns early.
- Turning `enable_external_access` off *does* preserve spilling (startup adds the temp
  directory to `allowed_directories`), but it makes the postgres and mysql extensions refuse
  to `ATTACH` unconditionally, with no allowlist entry able to re-permit it.
- `FileSystem::GetLocal` is not an escape hatch either: `LocalDatabaseFileSystem::GetFileSystem`
  re-applies the `disabled_filesystems` check by hand.
- A remote `temp_directory` is not usable — DuckDB cannot open a remote file for both reading
  and writing.

### Where this comes from

[Windmill](https://github.com/windmill-labs/windmill) runs DuckDB **in-process** in its job
workers, through an FFI cdylib. Every other language runtime it supports is confined by
nsjail/unshare; DuckDB cannot be, because there is no child process to jail. So the worker
mirrors that policy at the connection level instead, splicing `disabled_filesystems` (plus an
INSTALL/LOAD allowlist) ahead of the user's SQL.

That fence is what makes it safe to run untrusted DuckDB scripts at all — but with it in place,
every DuckDB feature that depends on going out-of-core is crippled: large sorts, aggregations,
joins and `CREATE TABLE AS` all fail as soon as they exceed `memory_limit`, on precisely the
worker class that exists because the tenants are not trusted. Spilling is the one filesystem
capability such a deployment needs to keep, and it is the one it cannot express today.

## The change

`lock_temp_directory` (BOOLEAN, global, default `false`, irreversible). Setting it:

1. fixes `temp_directory` in place — it can no longer be modified, and the setting itself
   cannot be unset or reset;
2. routes the buffer manager's temporary files through a local file system that
   `disabled_filesystems` does not apply to (`FileSystem::GetTemporaryFileSystem`).

The two halves are only safe as a pair, and that is the crux of the design: the exemption is
sound **because** the directory can no longer be redirected. Without (1), a user could point
`temp_directory` at an arbitrary path and turn the exemption into a blind write primitive.
With it, the exempted location is fixed by the embedder before any untrusted SQL runs.

`allowed_directories` still applies either way, since the new file system is still an
`OpenerFileSystem` — so an embedder that also restricts paths keeps that restriction.

The temp directory was already treated as a protected resource, which is the precedent this
follows: `TempDirectorySetting` refuses to modify it when `enable_external_access` is false,
and the startup path adds it to `allowed_directories` so that spilling survives that lockdown.

## Notes for review

**No struct member changes offset.** Both new fields are placed in space the structs already
had — `lock_temp_directory` at the end of `DBConfigOptions`' run of bools, where it occupies
an existing padding byte, and `temporary_db_file_system` last in `DatabaseInstance`. Verified
with `clang++ -Xclang -fdump-record-layouts` against the unpatched tree: the only difference is
added lines. This matters for extensions built against the current layout — during development
an earlier revision declared the `DatabaseInstance` member next to `local_db_file_system`,
which pushed `create_api_v1` 8 bytes along, and every prebuilt extension `LOAD` then hung
rather than failing.

**`test_reset.cpp`.** `lock_temp_directory` joins the excluded-options list next to
`disabled_filesystems`, `lock_configuration` and `enable_external_access`: that test requires
every setting to survive a `RESET`, which an irreversible one cannot.

**Generated files.** `src/common/settings.json` regenerated with
`scripts/generate_settings.py`; formatted with clang-format 11.0.1.

## Tests

Two sqllogictests:

- `test/sql/settings/lock_temp_directory.test` — the lock takes effect, `temp_directory` can no
  longer be set or reset, the setting cannot be unset, a local read is still refused, and a
  query that outgrows `memory_limit` still spills and completes.
- `test/sql/settings/lock_temp_directory_unset.test` — without the lock, the fence stops the
  spill. This is what keeps the first test honest: it fails on a build where the query never
  needed to spill in the first place.

Full fast suite on this branch: **974 376 assertions across 4900 test cases, all passing**
(420 skipped for missing optional extensions).

### CI on my fork

Run per CONTRIBUTING before opening. Everything passed except three tests, none of which this
change touches — I chased each one down rather than assume:

- `test/sql/index/art/scan/test_not_distinct_index_scan.test` — fails under the alternate
  storage configs (`compressed_in_memory`, `force_storage`, `latest_storage` and others).
  **Reproduces on an unmodified `main` build**:
  `./build/release/test/run --test-config=test/configs/compressed_in_memory.json test/sql/index/art/scan/test_not_distinct_index_scan.test`
  It passes on the default config, which is presumably why it is not more visible.
- `test/sql/storage/compression/fsst/fsst_greedy_long_strings.test` — flaky on `main`. It
  asserts an exact storage-segment count and returns 2 instead of 1 roughly 8% of the time:
  **16 failures in 200 consecutive runs of that test alone, on an unmodified `main` build**.
- `test/sql/parallelism/interquery/concurrent_checkpoint_insert.test` on Windows — CI's own
  retry logic attempted it twice before failing.

Happy to open any of those separately if useful; I did not want to bundle unrelated fixes here.

Two rounds of that CI also caught genuine problems in *my* tests, both since fixed: they
created a persistent table (which cannot commit once the test has disabled the file system, so
it is now a `TEMP` table), and they lacked `require vector_size 2048`, which the
`STANDARD_VECTOR_SIZE=2` build needs.

---

*Disclosure: this patch was written with AI assistance. I have reviewed it, run the test suite
and the layout verification described above, and I am submitting it under my own name. I am
aware of the note in CONTRIBUTING.md and am flagging it up front rather than leaving you to
infer it — close it on those grounds if you prefer, and I will not take it personally.*

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -980,6 +980,13 @@
         "default_value": "false"
     },
     {
+        "name": "lock_temp_directory",
+        "description": "Fix the temp directory in place: 'temp_directory' can no longer be modified, and the temporary files written by the buffer manager bypass 'disabled_filesystems'",
+        "type": "BOOLEAN",
+        "scope": "global",
+        "custom_implementation": true
+    },
+    {
         "name": "log_query_path",
         "description": "Specifies the path to which queries should be logged (default: NULL, queries are not logged)",
         "type": "VARCHAR",
--- a/src/include/duckdb/common/file_system.hpp
+++ b/src/include/duckdb/common/file_system.hpp
@@ -164,6 +164,10 @@ public:
 	DUCKDB_API static FileSystem &GetFileSystem(ClientContext &context);
 	DUCKDB_API static FileSystem &GetFileSystem(DatabaseInstance &db);
 	DUCKDB_API static FileSystem &GetLocal(DatabaseInstance &db);
+	//! File system through which the buffer manager reaches its temporary files. This is the regular file system
+	//! unless the temp directory has been locked, in which case it is the local one - see
+	//! DBConfigOptions::lock_temp_directory.
+	DUCKDB_API static FileSystem &GetTemporaryFileSystem(DatabaseInstance &db);
 	DUCKDB_API static FileSystem &Get(AttachedDatabase &db);
 
 	DUCKDB_API virtual unique_ptr<FileHandle> OpenFile(const string &path, FileOpenFlags flags,
--- a/src/include/duckdb/main/config.hpp
+++ b/src/include/duckdb/main/config.hpp
@@ -103,6 +103,12 @@ struct DBConfigOptions {
 	bool force_checkpoint = false;
 	//! Run a checkpoint on successful shutdown and delete the WAL, to leave only a single database file behind
 	bool checkpoint_on_shutdown = true;
+	//! Whether the temporary directory is fixed in place. Once set this cannot be unset: 'temporary_directory' can no
+	//! longer be modified, and in exchange the buffer manager reaches its temporary files through the local file
+	//! system rather than the virtual one, so 'disabled_filesystems' no longer applies to them. Declared at the end of
+	//! the bool run so it takes a padding byte the struct already had: extensions are compiled against this layout,
+	//! and every member above keeps the offset it has in a release without this one.
+	bool lock_temp_directory = false;
 	//! Serialize the metadata on checkpoint with compatibility for a given DuckDB version.
 	StorageCompatibility storage_compatibility = StorageCompatibility::Default();
 	//! Initialize the database with the standard set of DuckDB functions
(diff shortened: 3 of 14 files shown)


rubenfiszel:
Filed #24695 to describe the underlying problem separately from this implementation. If a different design would suit the codebase better, that issue is the better place to land it and I would happily close this in favour of one.

PR #24697, opened 2026-08-11
Push down ungrouped aggregates into file readers

### LINKED ISSUE #8310: Epic: Duckdb projection expression pushdown

myrrc:
- Add AggregateFnPushdown optimizer which pushes ungrouped single-argument aggregates (sum(x), min(x), etc.) and count(*) into file readers.
- Support count(*) pushdown in Parquet. This lowers down CPU usage significantly (3-4x) on files with large row group count.
- Add FinalizeScan and TryPushdownAggregates to MultiFileReader interface so implementors like Vortex can make use of it.
- Add `aggregate_pushdown` callback to TableFunction. This currently pushes down only ungrouped aggregates of specific form, but in the future we may push more with the same signature, so comment doesn't specify what exactly we will push. It will be caller's responsibility to check the input.

### DIFF
--- a/extension/parquet/include/parquet_multi_file_info.hpp
+++ b/extension/parquet/include/parquet_multi_file_info.hpp
@@ -72,9 +72,7 @@ struct ParquetMultiFileInfo : MultiFileReaderInterface {
 	                                                           MultiFileGlobalState &global_state) override;
 	unique_ptr<LocalTableFunctionState> InitializeLocalState(ClientContext &, GlobalTableFunctionState &) override;
 
-	bool SupportsReadAhead(const MultiFileBindData &bind_data) const override {
-		return true;
-	}
+	bool SupportsReadAhead(const MultiFileBindData &bind_data) const override;
 	shared_ptr<BaseFileReader> CreateReader(ClientContext &context, GlobalTableFunctionState &gstate,
 	                                        BaseUnionData &union_data, const MultiFileBindData &bind_data_p) override;
 	shared_ptr<BaseFileReader> CreateReader(ClientContext &context, GlobalTableFunctionState &gstate,
@@ -88,6 +86,9 @@ struct ParquetMultiFileInfo : MultiFileReaderInterface {
 	void GetVirtualColumns(ClientContext &context, MultiFileBindData &bind_data, virtual_column_map_t &result) override;
 	unique_ptr<MultiFileReaderInterface> Copy() override;
 	FileGlobInput GetGlobInput() override;
+	bool TryPushdownAggregates(ClientContext &context, MultiFileBindData &bind_data,
+	                           const TableFunctionAggregateInput &input) override;
+	bool FinalizeScan(ClientContext &context, GlobalTableFunctionState &global_state, DataChunk &output) override;
 };
 
 class ParquetScanFunction {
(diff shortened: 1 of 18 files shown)


Mytherin (commented):
Thanks - we already "kind of" do aggregate pushdown into scans using the partition pushdown code (where we push `COUNT(*)`, `MIN`, `MAX`, etc and evaluate them using row group stats). I don't think there's much benefit to adding `COUNT(*)` pushdown to the parquet reader separately from that code path as we're already evaluating this in an optimal manner through the other pushdown path. Maybe these paths could be unified somehow instead of living side-by-side?

myrrc:
> where we push COUNT(*), MIN, MAX, etc and evaluate them using row group stats

This unfortunately isn't a generalisable approach. Vortex doesn't have row groups in the Parquet sense, and zone map granularity (8192 rows) is much smaller, so scheduling costs if we set partition = zone map overweight the benefits greatly. One example of this is you need to report partition index while pushing data to DataChunk. In result you'd return from scan function every 8k rows which gives very bad performance.

Reason number two is this doesn't work if aggregates are filtered. This also makes sense because you generally don't know what to do if a partition is partially filtered out. In Vortex we can read and accumulate the post-filter chunk ourselves.

Reason number three is that our stats may be imprecise (in which case we need to load the chunk and evaluate it).

I suggest adding the new callback for readers which don't look like Parquet or have different granularity (like Vortex) and keep the current implementation for formats where partition stats are enough.

Mytherin:
Thanks for the comments - but I'm still not really seeing a good reason to introduce the extra callbacks here. At least in the current suggested implementation it seems that the main motivation behind this callback is to use metadata to answer aggregates, which is already enabled by the partition stats callback optimizer. Perhaps that callback can instead be modified / extended to make it work for Vortex?

> Reason number two is this doesn't work if aggregates are filtered. This also makes sense because you generally don't know what to do if a partition is partially filtered out. In Vortex we can read and accumulate the post-filter chunk ourselves.

> Reason number three is that our stats may be imprecise (in which case we need to load the chunk and evaluate it).

For `COUNT(*)`, it should be possible to emit chunks that only have a count for the filtered scenario as a `EMPTY` column would be the only column required in the projection list. This should have a very minimal overhead, i.e. it shouldn't be necessary to emit any actual values.

myrrc:
> main motivation behind this callback is to use metadata to answer 
aggregates

Yes, and also support this in queries with filters.
Current distinction is that aggregates propagation happens in optimizer phase (row group is large, reading it and getting partition stats is cheap), but our implementation executes in run phase, because we determine whether the row group is to be skipped/populated from metadata only during query execution (since a row group ~ zone map is 8192 rows, reading it separately from file contents is expensive).

I can in theory tweak the callback optimizer to run during scan and support filtering to avoid scanning partitions which can be supplied from metadata (although this would be a much bigger change), but I can't change granularity of row group partitioning in Vortex.

Returning from Scan() every 8192 rows is impractical from performance perspective.

myrrc:
I therefore believe this callback serves a different optimization which can't be reasonably applied in plan/optimize/bind phase.

myrrc:
Count(*) optimization in Parquet, although it is faster that the current version, is here purely to test changes in one of actual readers since you don't vendor Vortex and can't test our aggreation pushdown.

Mytherin:
Given this then seems to be rather specific to Vortex maybe a better approach here is to move the optimizer to the vortex extension and integrate that there? I'm hesitant to add all of this infrastructure / callbacks for something that doesn't offer benefits to other scans.

myrrc:
This change is Vortex-agnostic and other readers like Parquet can make use of it. I didn't include the changes for brevity. Parquet reader specifically can aggregate min/max as well in runtime if we have a filter, but a specific row group is fully inside the filter range.

I can add the changes should this impact the merge decision.

PR #24702, opened 2026-08-11
C-API-V2: Part 1

Maxxen:
This PR adds an initial draft of the first half of the V2 C-API + the stable "C++ API" built on top of it.

### WARNING: This is a **DRAFT** 

Nothing here should be considered decided or stable - we're still interested in receiving feedback, but don't start building against this yet or panic if certain features are missing. Also lots of the documentation is still clanker-generated vomit left from the initial prototyping phase. Things will evolve significantly in the coming weeks.

This first half covers the following sections:

-  Error handling (new)
- `Environment` (new)
- `Database` and options
- `Connection`
- `Context` (new)
- `Extension`, and new entry-point (new)
- `SQLStatements` (new)
- `QueryResult` (streaming, new)
- `LogicalType`
- `Value`
- `DataChunk`
- `Vector` (new non-flat vector-types)
- `VectorView` (new)

This is focused on just adding enough surface to support "client" use-cases, e.g. open a database, perform some queries, fetch some results - and not so much "extending DuckDB with new features". We've prototyped that other half already in a separate fork (scalar/table/cast/copy functions, filesystem, custom types, arrow interop, logging, etc, etc) which will follow soon.

This PR is big, but Is nonetheless smaller than it might look at first glance due to the large number of lines added by spec files and tests. I think the primarily interesting files to look at is `duckdb_v2.h`, `duckdb_cpp.hpp`. Additionally, almost 60% of the entire API surface is currently dedicated to different `duckdb::Value` constructors/accessors. More on that further below.

I'll try to go through section-by-section and highlight some of the motivations, design decisions, ideas, future work, and notable differences from the V1 API.

# Error Handling

Probably the biggest difference from the V1 API is that we now have a somewhat coherent error handling and reporting story.

The convention we use is that essentially _all_ API functions now return a `DUCKDB_V2_ERROR`-code, and always receive an optional `duckdb_v2_error_info_handle` pointer as their _last_ argument. The actual return-value(s) is returned by an out-pointer. 

There are two exceptions to this rule: the error "inspection" functions, as returning an error when trying to get the error message from another error quickly spirals into sillyness; and the "destructor" functions for handle types, which only return a error code.

Here's an example of how this can look in practice:

```c

// Example function to get chunk size or exit with error
idx_t try_get_chunk_size(duckdb_v2_data_chunk_handle chunk) { 
	// Setup error handle
	duckdb_v2_error_info_handle err = NULL;
	// Setup return value
	idx_t out_size = 0;
	
	// Perform API call, and check status code
	if(duckdb_v2_data_chunk_get_size(chunk, &out_size, &err) != DUCKDB_V2_ERROR_NONE) { 
		
		// Retrieve message
		duckdb_v2_str msg;
		duckdb_v2_error_info_get_text(err, &message_view);
		
		// Print and exit
		fprintf(stderr, "Error!: %.*s\n", (int)msg.len, msg.ptr);
		
		// Free error info
		duckdb_v2_error_info_destroy(&err);
		
		duckdb_v2_error_i
		exit(-1);
	}
	
	// All ok! return
	return out_size;
}                                       
```

Note that the last `duckdb_v2_error_info_handle` argument is always optional, and will only be set if an error actually occurred. So you only have to check the returned error code if you don't care about additional error details.

This means that almost every operation in the V2 API is _fallible_ - because they are. DuckDB itself uses C++ exceptions liberally internally, making it really hard to even know _what_ can fail. There is also a myriad of reasons why an operation might fail unexpectedly even outside of our own control (e.g. std::bad_alloc).  So making sure every API call can report an error at the call site (and not stashed in some other context object where you may forget to look) seems like the most robust way to handle this.

Since a lot of the API operates on opaque "handles", we originally thought of storing errors in the handles themselves, which could be retrieved at any point after an operation, but decided against it. It would require a lot of extra functions to inspect each and every handle type, and still require each regular function to return an error code anyway (or you have to call another function to check if an error occurred after each operation). It would also require us to change the internal data structures backing these handles to now have space to hold errors. Its also not always clear which handle should receive or "own" the error after an API call that operate on multiple handles.

We also considered using global thread-local-storage to stash the "last reported error", like `errno`. We decided against this too as TLS doesn't always work well with dynamic linking, and keeping the errors thread-local requires the consumer to juggle errors around across callbacks (as DuckDB may sometimes execute certain callbacks on different threads) and in their own program. This becomes particularly annoying in multi-tasking environments based on "green threads" (e.g. Go, or even async-rust) which don't always make it clear when a piece of code will yield and continue on a separate system thread.

Having explicit error info handles instead also make error propagation across callbacks straightforward. The idea is that any callback function invoked by the DuckDB will also provide a `duckdb_v2_error_info_handle` that user-code can pass into other API calls. Unfortunately, this PR does not provide any APIs that take callbacks, but heres an example from the prototype of our scalar-function interface that hopefully illustrates the pattern:

```c

typedef struct { bool some_condition } my_struct;

void my_bind_callback(duckdb_v2_scalar_function_bind_info_handle info, duckdb_v2_context_handle context,  
                            duckdb_v2_error_info_handle *err) { 
    
    
    // do some logic
    // ...
        
    // Try to get the "user data"
    void* out_data;
    
    // Forward the error info handle from the callback.
    if(duckdb_v2_scalar_function_bind_get_user_data(info, &out_data, err) != DUCKDB_V2_ERROR_NONE) { 
	    // If the operation failed, the error will be set in the provided error info,
	    // so we can just return and DuckDB will handle the error itself once it regains control!
	    return;
    }
    
    // some more code
    // ...
    
    if(!((my_struct*)out_data)->some_condition) { 
	    // We can also set an error outselves!
	    
	    // A non-zero error code is all thats required for DuckDB to raise an exception internally
	    duckdb_v2_error_info_set_code(*err, DUCKDB_V2_ERROR_INPUT_INVALID);
	    
	    // But we can also provide more details if we want to
	    duckdb_v2_str msg = { .ptr = "some error!", .len = 11};
	    duckdb_v2_error_info_set_text(*err, msg);
	    
	    return;
    }
    
    // ...
}

```

Our hope is also that this convention makes it easy to wrap API calls in a way that propagates errors to the host systems/environments. For example, the C++ API can trivially convert back error info into exceptions again (and vice-versa), and we've also experimented with Rust bindings, where this model fit nicely into the Rust `Result<T, Err>` convention.

# The `Environment`

The C-API-V2 has a new "root" object for clients wanting to interface with DuckDB, the `Environment`. 

The `Environment` is essentially the top-level container/cache for open databases. By requiring an environment to open a database, we can e.g. detect and throw an error when a user tries to open a database that has already been opened in the same process, which currently can cause unexpected problems in the V1-C-API.

Since DuckDB can also operate on multiple databases (through ATTACH), it seems like a good idea to provide some handle in the C-API to interact with DuckDB before any databases have been opened, or outside the context of a single database. So far the `Environment` can only really tell you how many databases have been opened through it, but we may add additional functions operating on it in the future to manage other types of cross-database state too.

The initial steps for interacting with DuckDB now looks like:

```c
int main() { 

    duckdb_v2_error_info_handle err = NULL;
    
    duckdb_v2_environment_handle env = NULL;
    if(duckdb_v2_create_environment(&env, &err) != DUCKDB_V2_ERROR_NONE) { 
        // ... handle error
    }
    
    duckdb_v2_database_handle db = NULL;
    if(duckdb_v2_open(env, duckdb_v2_str {NULL, 0}, NULL, 0, &db, &err) != DUCKDB_V2_ERROR_NONE) {
        // ... handle error
    };
    
    
    duckdb_v2_connection_handle conn = NULL;
    if(duckdb_v2_connect(db, &conn, &err) != DUCKDB_V2_ERROR_NONE) {
        // ... handle error
    }

     // We now have a connection we can use!
}	
```

# `duckdb_v2_str`

Instead of passing `const char*` across the API boundary, the V2-C-API introduces a dedicated `duckdb_v2_str` "string view"/"length delimited" struct. 

```c
struct duckdb_v2_str {
	const char *ptr;
	idx_t len;
};
```

This makes it much more efficient (and safe!) to pass strings across the API both ways, as most other languages (except C) also represent strings as `{ptr, len}` pairs internally (e.g. Rust, Go, C++). You no longer need to make a separate copy of every string you want to pass just to null-terminate it.

Another nice consequence is that a `duckdb_v2_str` is _never owned_, making it much easier to reason about ownership when interfacing in the API in general. You can always pass a `duckdb_v2_str` and DuckDB will make a copy of it internally - and you know that if you receive a `duckdb_v2_str` you need to make a copy of it yourself if you want to extend its lifetime.

# `Database` and `DatabaseOptions`

`TODO`

# `Connection` vs `Context`

One of the major things we want to address with the new V2-C-API is to get better at tracking resource usage and to improve interoperability with extensions. To do this, we try to avoid "context-less constructors", i.e. functions that create handles to internal objects "out of thin air". 

For example, creating a `LogicalType` now requires a `Connection`, because the available logical types to create might differ depending on what extensions are loaded or what databases setting are currently in scope. Similarly, if your code is being invoked from inside of DuckDB - for example during a function callback - there may be other (perhaps transaction-local) state that may affect the behavior of certain API calls.

The API is therefore essentially "split" between functions that you can use from:
- The __outside__ of DuckDB: which requires a `duckdb_v2_connection_handle`
  Most client drivers, or applications that embed DuckDB
- The __inside__ of DuckDB, which take a `duckdb_v2_context_handle`.
  E.g. extension code, or user code supplied as callbacks (such as UDFs)

Unlike a connection, you can't create a `duckdb_v2_context`, but it will almost always be supplied as the first argument to every callback function in the API, as well as the extension entrypoint function. It essentially captures the fact that your code is being invoked by DuckDB as part of a currently running query/transaction - and you can use it to access stuff like connection local state, settings, the current catalog, currently loaded filesystems, and potentially other scoped "services" like memory allocators etc.

It also prevents misuse of the API. You can't for example run a query using a `duckdb_v2_context`, because running a query within another query doesn't make sense (its not supported, and if you open a new connection you lose transactionality).

# V2 `Extension` entrypoint

This PR does not provide any interfaces to do things you would normally do in an extension (such as e.g. defining new functions), but it still implements the basic infrastructure to load V2-C-API extensions. 

One immediate difference between the V1 and V2 is that the "entry point" function for V2-C-API extensions is different. 
Instead of receiving a `DatabaseInstance/Connection`, the signature of the entrypoint function is now:

```c
void <NAME>_init_c_api_v2_internal(duckdb_v2_extension_handle extension, duckdb_v2_context_handle context, duckdb_v2_error_info_handle *err)
```

- `duckdb_v2_extension_handle` is basically the C-equivalent of the `ExtensionLoader`  used by C++ extensions. This is what you will use in the future to register new extension functions/types etc. It is only passed to the entry point during the initial extension load, and should not be stored our outlive the entry point callback.
- `duckdb_v2_context_handle`, as explained above, is the context handle of the currently executing statement, e.g. `LOAD <extension>`, or if this extension is statically linked, a ephemeral dummy connection. This allow you to e.g. access connection local state/settings set by the connection that executed the `LOAD` statement that is currently loading the extension. Unfortunately extension loading is not transactional, but this is one of the ways we could eventually try to make it more so in the future.
- `duckdb_v2_error_info_handle`, as also explained above in the "error handling" section is how you can report failure during the extension load, either by manually reporting an error or by propagating errors from other API calls you make during loading (e.g. trying to register an already existing function).

One of the key insights is that the extension entry point function is essentially just like any other callback invoked by the C-API.

We may change this signature more before 2.0 is released to e.g. also pass some sort of global memory allocator handle that the extension can save and use throughout its entire lifetime.

# SQL Statement Expansion

The V2-C-API provides more fine-grained control over how to execute SQL. Instead of just providing the equivalent of a `query(const char* sql)` function that both parses and executes a single sql statement, the V2 API splits parsing and execution statement-by-statement. This is done using two new concepts, the `duckdb_v2_statement_iterator_handle` and the `duckdb_v2_sql_statement_handle`. The flow looks like this:

```c
// Parse a SQL string (producing potentially multiple SQL statements!), and return an iterator
// (error handling ommited for brevity)
duckdb_v2_statement_iterator_handle iter = NULL;
duckdb_v2_parse_sql(conn, "SELECT 1 + 2; SELECT 4", &iter, NULL)

// Loop over all statements
duckdb_v2_sql_statement_handle stmt = NULL;

while(true) {

	// Extract the next statement from the iterator (the statement is parsed lazily)
	duckdb_v2_statement_iterator_next(iter, &stmt, NULL);
	
	if(!stmt) { 
		break; // No more statements left!
	}
	
	duckdb_v2_result_handle res = NULL;  
	duckdb_v2_statement_execute(conn, stmt, nullptr, nullptr, 0, &res, NULL);
	
	// ... do something with the result `res`
	
	// cleanup
	duckdb_v2_result_destroy(&res);  
	duckdb_v2_sql_statement_destroy(&stmt);
}

duckdb_v2_statement_iterator_destroy(&iter);
```

This solves two issues:
- Some DuckDB statements actually expand into multiple statements internally (e.g. PIVOT), and so these were previously a bit problematic to execute through the C-API.
- Statements are parsed _lazily_, meaning that you can e.g. load a parser extension, and then use the new syntax immediately in the same SQL string

So a SQL string is no longer required to be a single statement, you can parse and execute a whole SQL program if you'd like to, without having to do any statement splitting yourself.

Splitting SQL parsing from binding and execution also enables us to eventually support stuff like "parse and convert this SQL statement into some sort of parse tree/JSON, without binding it" by adding more functions that operate on the `duckdb_v2_sql_statement_handle`.

# Streaming Query Results

Another change in the V2-C-API is that all query results are now _streaming_, and fetched lazily, instead of being completely buffered and materialized.

The `duckdb_v2_result_handle` is basically a small state machine, and using it looks like this:

```c

// ... get a statement to execute from somewhere

duckdb_v2_result_handle res = NULL;  
duckdb_v2_statement_execute(conn, stmt, nullptr, nullptr, 0, &res, NULL);


bool is_done = false;
while (!is_done) {

	duckdb_v2_data_chunk_handle chunk = NULL;
	DUCKDB_V2_RESULT_STEP_STATUS status = DUCKDB_V2_RESULT_STEP_STATUS_WAITING;

	// "Step" the query result once
	duckdb_v2_result_step(res, &chunk, &status, NULL)

	switch (status) {  
	case DUCKDB_V2_RESULT_STEP_STATUS_CHUNK:
	    // We got a DataChunk!
	    // Do something with it!
	    // ...
	    do_something_with_the_data_chunk(chunk);
	    
	    // Cleanup
	    duckdB_v2_data_chunk_destroy(&chunk);
	    break;
	case DUCKDB_V2_RESULT_STEP_STATUS_FINISHED:  
		// Result is finished, we are done!
		is_done = true;
	    break;  
	case DUCKDB_V2_RESULT_STEP_STATUS_CANCELLED:
		// The query was interruped!
		is_done = true;
		break;
	case DUCKDB_V2_RESULT_STEP_STATUS_WAITING:
		// DuckDB is not ready to produce another chunk yet. We could either:
		// - yield and come back later
		// - ask DuckDB to "wait" a bit until way may be able to make progress
		
		// Lets wait...
		// This will block for a bounded amount of time or until more chunks are available.
	    duckdb_v2_result_wait(res, NULL);  
	    break;  
	default:  
	    break;  
	}
}
```

This has two benefits:
- Returning extremely large query results can now be done incrementally without requiring the whole result to fit in memory.
- Having DuckDB return `DUCKDB_V2_RESULT_STEP_STATUS_WAITING` when waiting for more data enables integrating the result fetching API with various "async" runtimes when embedding DuckDB in another language, as it effectively signals that now is a good time for the current co-routine/task to yield.

While not part of this PR, we will support creating "`ColumnDataCollection`"s through the V2-C-API, which will enable you to buffer returned `DataChunks` (with support for larger-than-memory handling) to create your own "materialized results", if you would like to defer handling the result until the query has completed. In that sense, streaming-by-default results are a strictly more powerful interface than materialize-by-default (as it was provided in the V1-C-API).

# Context-scoped Types and Values

This is basically follows the same principle explain in "`Connection` vs `Context`" - we want to be able to track all objects created through the C-API, and also support extension-defined semantics better.

The functions associated with creating `duckdb_v2_value`s and `duckdb_v2_logical_type`s exhibit the consequence of this very prominently. Because these are objects that we need to be able to create and manage both outside and inside of DuckDB, there are now two basically identical sets of these functions, with the only difference being whether they take a `duckdb_v2_connection` or a `duckdb_v2_context`. (indeed, almost 60% of this PR is made up of value constructor functions).

This may feel cumbersome, but the benefit is that it allows us to e.g. cache/deduplicate/centrally allocate all types/values created from a connection/context. It also enables the creation of custom types and the use of custom cast rules defined or altered by users/extensions/settings.

However, if I were to provide a suggestion to client maintainers, it would be to _not_ wrap handles to values and types in native objects directly. Instead define your own type/value objects in the host language, and only convert to and from the DuckDB equivalent when you need to cross the API boundary.

# Non-flat vectors and `VectorView`

The V1-C-API only supported "flat" vectors, meaning that no C-API based extension could leverage DuckDBs ["compressed execution"](https://duckdb.org/docs/current/internals/vector#vector-format) as all vectors passed to the C-API would be flattened internally. This made it much easier to handle vectors for API consumers, but also made invoking C-API defined functions less performant.

This is no longer the case in the V2-C-API, we now no longer flatten vectors by default. You can do it yourself though by calling `duckdb_v2_vector_flatten` - but if you don't you can make use of the new `duckdb_v2_vector_get_view` to get a unified "vector view" to handle both `FLAT`, `CONSTANT` and `DICTIONARY` vectors easily. This is how it works:

```c
// A vector view is defined as such:
struct duckdb_v2_vector_view {  
    const void *data;               // Pointer to the data buffer
    const uint64_t *validity;       // Validity mask marking which rows are null (basically a bitset)
    const duckdb_v2_sel_t *sel;     // "selection" vector for dictionary-compressed vectors
    idx_t count;                    // Number of elements in the vector
};

//...

// Get some vector from somewhere
duckdb_v2_vector_handle vec = get_vector_from_some_function(...);

// Setup a vector view
duckdb_v2_vector_view view;  
duckdb_v2_vector_get_view(vec, &view, NULL);

// Cast data pointer
int32_t* data_ptr = (int32_t*)(view.data)

printf("About to print vector of %llu elements:", view.count);

// Loop over all "rows" in the vector
for(idx_t raw_idx = 0; raw_idx < view.count; raw_idx++) { 
	// Use the selection vector to get the physical index for this row, if provided
	idx_t row_idx = view.sel ? view.sel[raw_idx] : raw_idx;
	
	if(!view.validity || (view.validity[row_idx / 64] & (uint64_t(1) << (row_idx % 64))) != 0)) { 
		// Either:
		// - there is no validity mask => all rows valid
		// - the corresponding bit is set => this row is valid
		
		int32_t value = data_ptr[row_idx];
		
		// Do something with the value!
		printf("Value at row %d: %d", raw_idx, value)
		
	} else { 
		// Value is null! Do something with that
		printf("Value at row %d: is null!", raw_idx)
		
	}
}

printf("Done!");
```


src/include/duckdb_v2.h
```diff
@@ -0,0 +1,5489 @@
+//===----------------------------------------------------------------------===//
+//
+//                         DuckDB
+//
+//
+//===----------------------------------------------------------------------===//
+//
+// !!!!!!!
+// WARNING: this file is autogenerated, manual changes will be overwritten
+// !!!!!!!
```
Maxxen, line 10: The generated `duckdb_v2.h` API header

tools/cpp/duckdb_cpp.hpp
```diff
@@ -0,0 +1,2171 @@
...
+//----------------------------------------------------------------------------------------------------------------------
+// 8888888b.                    888      8888888b.  888888b.               d8888 8888888b. 8888888
+// 888  "Y88b                   888      888  "Y88b 888  "88b             d88888 888   Y88b  888
+// 888    888                   888      888    888 888  .88P            d88P888 888    888  888
+// 888    888 888  888  .d8888b 888  888 888    888 8888888K.           d88P 888 888   d88P  888
+// 888    888 888  888 d88P"    888 .88P 888    888 888  "Y88b         d88P  888 8888888P"   888
+// 888    888 888  888 888      888888K  888    888 888    888        d88P   888 888         888
+// 888  .d88P Y88b 888 Y88b.    888 "88b 888  .d88P 888   d88P       d8888888888 888         888
+// 8888888P"   "Y88888  "Y8888P 888  888 8888888P"  8888888P"       d88P     888 888       8888888
+//----------------------------------------------------------------------------------------------------------------------
```
Maxxen, line 12: The "stable" C++ API header. You only need to include this and `duckdb_cpp.cpp` to use C++ on-top of the V2-C-API from a client. For an extension you also need `duckdb_cpp_extension.hpp` - but thats it.

One of the really cool things Im happy about is that this header by itself _does not_ include `duckdb_v2.h` itself (only the implementation file does), so it does not pollute your source files with a bunch of c-symbols.

test/extension/cpp_api_demo.cpp
```diff
@@ -0,0 +1,18 @@
...
+using namespace duckdb::cxx;
+
+DUCKDB_CPP_EXTENSION_ENTRYPOINT(Extension &extension, Context &context) {
+	(void)extension;
+
+	// Binding a type proves the context is live and has a transaction: this reaches into the catalog.
+	const auto type = context.ParseType("STRUCT(a INTEGER, b VARCHAR)");
+
+	context.Log(LogLevel::INFO, "cpp_api_demo loaded, parsed " + type.ToText(), LOG_TYPE);
+}
```
Maxxen, line 18: Super small example of how the entrypoint for a C++ extension that builds on top of the stable`duckdb_cpp.hpp` header looks.

There are no functions to e.g. register scalar functions and do other extension-like stuff etc yet, but in the future they will be exposed as methods on the `Extension` object. (similar to how the `ExtensionLoader` works today)

api_spec/v2/arena/arena.yaml
```diff
@@ -0,0 +1,48 @@
+module: arena
+#
+# Bump allocation of caller-written bytes whose lifetime belongs to another
```
taniabogatsch: This sentence is hard to follow. Bump allocation to what?

api_spec/v2/arena/arena.yaml
```diff
@@ -0,0 +1,48 @@
...
+# object. An arena has no value semantics of its own: it hands out raw
+# uninitialized bytes, and the object that lends it defines what those bytes
+# may back and how long they live.
+#
+# The only source today is vector_get_arena, whose bytes back the out-of-line
```
taniabogatsch: This line is also confusing because it talks about "source" and then "write"? Maybe the whole description here needs a touch up 😅

api_spec/v2/arena/arena.yaml
```diff
@@ -0,0 +1,48 @@
...
+# once (that borrow is the single string-ness check), reserve bytes with
+# arena_allocate, write them, assemble a duckdb_v2_bytes, and place it via
+# vector_get_data_mutable. Values of at most DUCKDB_V2_BYTES_INLINE_LENGTH
+# bytes are stored inline and need no allocation. The C++ API wraps the whole
+# sequence.
+functions:
+  # TODO: This should probably take an "alignment" parameter too
+  arena_allocate:
+    lifecycle:
+      - [ "stable", "v2.0.0", "2026-08-04" ]
```
taniabogatsch, line 14: Should we align the dates with releases (week dates or so) somehow? I feel like the date like this has very little meaning?

api_spec/v2/arena/arena.yaml
```diff
@@ -0,0 +1,48 @@
...
+    description: |
+      Reserves byte_len bytes from the arena.
+
+      Raw bump allocation: returns a writable pointer to byte_len uninitialized
```
taniabogatsch: "Bump" used again?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+  database:
+    description: An opaque, owned handle to a DuckDB database instance.
+  connection:
+    description: An opaque, owned handle to a DuckDB connection.
+  option:
+    description: |
+      An opaque, owned handle to a single config option: a name and a setting,
+      plus optional metadata (description, default setting, target scope,
+      aliases). option_create builds one from a name and a setting; a
+      database/connection get returns one with the metadata filled in too. The
```
taniabogatsch: Getting the database/connection returns an option? Maybe we can clarify.

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+  #   - `err` is optional: callers may pass nullptr to opt out of the detail.
+  #   - If the call succeeds, `*err` is left as nullptr.
+  #   - If the call fails and `err != nullptr`, DuckDB allocates an info handle
+  #     and stores its pointer in `*err`. The caller owns it and must destroy
+  #     it with error_info_destroy.
+  #   - If `*err` is already non-null on entry, DuckDB destroys the previous
+  #     info before writing. To keep an info across calls, detach it first:
+  #     `saved = *err; *err = nullptr;`.
+  error_info:
+    description: Opaque detail for a failed call; destroy with error_info_destroy.
```
taniabogatsch, line 38: Should we generate these "destroy with" hints? I feel like since we already specify them via "cleanup_with" we shouldn't type this out by hand (for all handles)?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+      array / struct / map / union child types, alias. A type is read-only once
+      constructed; instances arrive from a query result schema, or from one of
+      the create_type_* constructors. Always destroy via logical_type_destroy.
+      Borrowed strings returned by the getters are valid only until then.
+    cleanup_with: logical_type_destroy
+  value:
+    description: |
+      An opaque, owned handle to a single SQL value: a logical type plus a
+      typed payload. Always destroy via value_destroy. The string returned by
+      value_get_varchar / value_get_blob is borrowed and valid only until then.
```
taniabogatsch, line 53: I wonder if we should even put these details here? Same for the logical type above. Because if we do, then we need to extend these notes with every type that'll have similar behavior. Can't these notes just be example lists, or even live only in the function descriptions themselves? Also to make the comments less verbose.

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+    description: |
+      An opaque, owned handle to a data chunk: a set of vectors of equal
+      logical length plus a cardinality (row count). A streaming result
+      produces chunks via result_step / result_fetch_chunk, and the caller
+      destroys them.
+
+      A chunk's lifetime is fully independent of the result, connection, and
+      database that produced it. The chunk owns its data, so it — and the
+      vectors and view pointers borrowed from it — stay valid until
+      data_chunk_destroy, even if the producers are destroyed first.
```
taniabogatsch: Should we expand here that data chunks are a versatile building block used in many different places of the API, not just when consuming results?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+    cleanup_with: data_chunk_destroy
+  vector:
+    description: |
+      A borrowed handle to a vector within a data chunk, carrying one column's
+      worth of values. Valid for as long as the owning chunk is; never destroy
+      it. Read it with vector_get_view, which reports data, validity and
```
taniabogatsch: What if I want to write to it?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+      selection in one unified shape. The row count comes from the caller's
+      context: the chunk for a top-level vector, the parent for a nested child
+      vector.
+  arena:
+    description: |
+      A borrowed handle to a bump allocator owned by another object. Bytes
```
taniabogatsch: Bump again? I don't get it 😅👀

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+      An opaque, owned handle to a caller-defined resource. Bundles the pointer
+      with optional callbacks that destroy and compare the resource.
+    fields:
+      - {name: ptr, type: void, pointer: 1}
+      - {name: destroy, type: opaque_destroy}
+      - {name: equals, type: opaque_equals}
+callbacks:
+  text_sink:
+    description: |
+      Receives text produced by DuckDB, without DuckDB allocating on the
```
taniabogatsch: I find "callers behalf" here a bit confusing, because as far as I understand DuckDB is the caller of the callback?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+
+      The sink runs inside DuckDB's call frame: it must not throw across the
+      boundary, and it must not re-enter the API on the handle being operated
+      on.
+    return_type: void
+    parameters:
+      text:
+        type: str
+        description: Borrowed view of the complete text. Valid only until the sink returns.
+      user_data:
```
taniabogatsch, line 171: To clarify, is this something that the user can set in the connection? Like, a user context?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+
+      err is a live error slot, never NULL. Populate it with
+      error_info_set_code / error_info_set_text to fail the producing call,
+      which then returns that code instead of its own result; that is the
+      sink's only influence on the outcome. Do not destroy it — DuckDB owns it.
+
+      The sink runs inside DuckDB's call frame: it must not throw across the
+      boundary, and it must not re-enter the API on the handle being operated
+      on.
+    return_type: void
```
taniabogatsch, line 166: We don't we return error codes in callbacks, too?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+      user_data:
+        type: void
+        indirection: 1
+        description: The opaque pointer handed to the producing call, passed through untouched.
+      err:
+        type: error_info
+        indirection: 1
+        kind: OUT
+        description: Live error slot, never NULL. Populate it to fail the producing call; do not destroy it.
+  opaque_equals:
```
taniabogatsch, line 180: Missing short description?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+    parameters:
+      a:
+        type: void
+        indirection: 1
+        description: Opaque pointer to the first resource to compare.
+      b:
+        type: void
+        indirection: 1
+        description: Opaque pointer to the second resource to compare.
+  opaque_destroy:
```
taniabogatsch, line 201: Same here, missing short description?

api_spec/v2/common/common.yaml
```diff
@@ -0,0 +1,243 @@
...
+    return_type: void
+    parameters:
+      data:
+        type: void
+        indirection: 1
+        description: Opaque pointer to the resource to destroy. The callback is responsible for freeing it.
+aliases:
+  sel:
+    underlying: u32
+    description: Selection-vector entry. Mirrors duckdb::sel_t.
```
taniabogatsch: Can we expand a little bit here on what a selection vector is? I feel like it's a crucial part of DuckDB, but probably not very generally known. Like, it's basically a dictionary.

taniabogatsch (commented):
Did a first pass over the 'arena' and 'common' yaml files. Left some comments, but love how this is coming together. 🦆

api_spec/v2/arena/arena.yaml
Maxxen: Yeah the date is not really for end-users, but just to make sure functions have a fixed offset in the generated vtable for extensions, so that if you add a new function or stabilize an existing function later (in the same release cycle), it doesn't modify the position of previous functions. Right now they are sorted deterministically based on date and then name.

taniabogatsch: Okay I see, makes more sense then 👍

api_spec/v2/common/common.yaml
Maxxen: No, its something the user passes at the call site (along with the callback)

api_spec/v2/common/common.yaml
Maxxen: No, we just set the error info.

api_spec/v2/common/common.yaml
Maxxen: Good idea, we should make that part of the generator output in the future.

Tishj:
Exciting stuff!

I started reading the description, and I think it could use some polish
The initial example for error handling is incorrect.
- message_view -> msg
- duckdb_v2_error_i -> ???

Also, bit of a nitpick since it is just an example, but in your example the destroying of the error is unnecessary since it exits the process anyway. The destroy would be only be required if the process were meant to continue.

> Note that the last duckdb_v2_error_info_handle argument is always optional, and will only be set if an error actually occurred. So you only have to check the returned error code if you don't care about additional error details.

This is slightly misleading I think? You need to check it if you set the `duckdb_v2_error_info_handle` reference (pass a non-null pointer), because that indicates intent, and causes an allocation of the handle, right?

#### `duckdb_v2_str`

"extend its lifetime"
What exactly is the lifetime guarantee of a duckdb-owned str?
Is this annotated on the returning function or is this implicit only?
I think this needs more consideration if its only implicit.
My initial thought was adding a `const char* lifetime;` to the struct, to annotate with predefined constants like "environment", "database", "connection".
^ that could also be a `void *owning_handle;`, set to the handle who's lifetime its connected to.
Ignore this comment if there's already a good convention in place, then I'd just like to see it in the description :+1:

EDIT:
I went through the PR afterwards, looks like it's annotated on the returning function.
I'm still a little worried about future methods that might have a dynamic owner of returned string data (i.e multiple handles are involved, memory could belong to either one).

#### Statement expansion

Have we not considered separating binding from execution?
That could bite us later, no?

Also:
> Statements are parsed lazily, meaning that you can e.g. load a parser extension, and then use the new syntax immediately in the same SQL string.

I'd hold off on that, because currently the plan for both parser extension types is to require manually enabling them.
And with dialect extensions there's another caviat because the entire query gets tokenized in its entirety before parsing further. We might still fix that before 2.0, but maybe not.

src/include/duckdb_extension_v2.h
```diff
@@ -0,0 +1,874 @@
...
+
+//===--------------------------------------------------------------------===//
+// Function pointer struct
+//===--------------------------------------------------------------------===//
+typedef struct {
+	// capigen:begin appended
+	DUCKDB_V2_ERROR(*duckdb_v2_arena_allocate)
+	(duckdb_v2_arena_handle arena, idx_t byte_len, uint8_t **out_ptr, duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR(*duckdb_v2_bignum_decode)
+	(const uint8_t *in_data, idx_t in_length, uint8_t *out_data, idx_t out_capacity, idx_t *out_length,
```
Tishj, line 78: nit: `source`, `dest` as opposed to `in`, `out`? Since we already reserve `out_` for convention.
Actually.. it's fine, just `out_capacity` as a non-out parameter is what's bothering me a little

src/include/duckdb_extension_v2.h
```diff
@@ -0,0 +1,874 @@
...
+	DUCKDB_V2_ERROR(*duckdb_v2_database_option_get)
+	(duckdb_v2_database_handle db, duckdb_v2_identifier_t name, duckdb_v2_option_handle *out_option,
+	 duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR(*duckdb_v2_database_option_get_by_index)
+	(duckdb_v2_database_handle db, idx_t index, duckdb_v2_option_handle *out_option, duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR(*duckdb_v2_database_option_get_count)
+	(duckdb_v2_database_handle db, idx_t *out_count, duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR(*duckdb_v2_database_option_set)
+	(duckdb_v2_database_handle db, duckdb_v2_option_handle option, duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR (*duckdb_v2_destroy_environment)(duckdb_v2_environment_handle *env);
```
Tishj, line 151: nit: `duckdb_v2_environment_destroy` ?
Aren't there naming conventions that enforce this?

src/include/duckdb_extension_v2.h
```diff
@@ -0,0 +1,874 @@
...
+	(duckdb_v2_value_handle value, char *out_string, idx_t out_capacity, idx_t *out_length,
+	 duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR(*duckdb_v2_vector_constant_set_valid)
+	(duckdb_v2_vector_handle vector, bool validity, duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR(*duckdb_v2_vector_flat_get_validity_mutable)
+	(duckdb_v2_vector_handle vector, uint64_t **out_validity, duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR (*duckdb_v2_vector_flatten)(duckdb_v2_vector_handle vector, duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR(*duckdb_v2_vector_get_arena)
+	(duckdb_v2_vector_handle vector, duckdb_v2_arena_handle *out_arena, duckdb_v2_error_info_handle *err);
+	DUCKDB_V2_ERROR(*duckdb_v2_vector_get_child)
```
Tishj, line 556: Aren't we missing type-specific vector methods?
For example `duckdb_v2_vector_list_set_capacity` ? Or is that coming with the second half, since its part of extension-surface?

src/include/duckdb_v2.h
```diff
@@ -0,0 +1,5507 @@
...
+struct duckdb_v2_bytes {
+	union {
+		struct {
+			uint32_t length;
+			char prefix[4];
+			char *ptr;
+		} pointer;
+		struct {
+			uint32_t length;
+			char inlined[12];
```
Tishj, line 417: 12 -> `DUCKDB_V2_BYTES_INLINE_LENGTH` ?

src/include/duckdb_extension_v2.h
Maxxen: Since all vectors have size now this is no longer needed! You can resize a list child just like a normal vector. Capacity is an implementation detail in that sense. (IMO we should probably have capacity on base vectors too)

src/include/duckdb_extension_v2.h
Maxxen: Yep this should be fixed - some of the "early" handles were created before we settled on strict with the naming convention, so should be revisited (same as .e.g database_open/database_close)

Maxxen:
> Exciting stuff!
> 
> I started reading the description, and I think it could use some polish The initial example for error handling is incorrect.
> 
> * message_view -> msg
> * duckdb_v2_error_i -> ???
>
> Also, bit of a nitpick since it is just an example, but in your example the destroying of the error is unnecessary since it exits the process anyway. The destroy would be only be required if the process were meant to continue.

Good points, will adjust!

> > Note that the last duckdb_v2_error_info_handle argument is always optional, and will only be set if an error actually occurred. So you only have to check the returned error code if you don't care about additional error details.
> 
> This is slightly misleading I think? You need to check it if you set the `duckdb_v2_error_info_handle` reference (pass a non-null pointer), because that indicates intent, and causes an allocation of the handle, right?

You can't create your own duckdb_v2_error_info_hande, so you can't really pass a non-null pointer (or well a pointer to a non-null pointer) unless you already had a previous error (which you should deal with it).
The whole point is that we _dont_ allocate in the success-path, so there is no error info to free. But you can also just call duckdb_v2_error_info_destroy unconditionally as passing a pointer-to-null to destructors is safe.

> #### `duckdb_v2_str`
> "extend its lifetime" What exactly is the lifetime guarantee of a duckdb-owned str? Is this annotated on the returning function or is this implicit only? I think this needs more consideration if its only implicit. My initial thought was adding a `const char* lifetime;` to the struct, to annotate with predefined constants like "environment", "database", "connection". ^ that could also be a `void *owning_handle;`, set to the handle who's lifetime its connected to. Ignore this comment if there's already a good convention in place, then I'd just like to see it in the description 👍
> 
> EDIT: I went through the PR afterwards, looks like it's annotated on the returning function. I'm still a little worried about future methods that might have a dynamic owner of returned string data (i.e multiple handles are involved, memory could belong to either one).

This is more of a theoretical concern than a practical one, I could make the same argument agains passing a `const char*`. The convention we use in this API: that `duckdb_v2_str` is always borrowed is strictly better as it's at least consistent, compared to the V1-API where you sometimes owned and had to free return char* pointers, and sometimes not. 

As of now we don't return dynamic strings as duckdb_v2_str's at all, and I don't think we really want to anyway as you would have to cache it somewhere on a handle (and indeed start thinking about the lifetime/validity of the pointer now). The few places we want to be able to pass dynamically constructed strings we instead use a callback-based approach to make it clear that the string you receive is alive within a transient scope (the duration of the callback). I think callbacks are superior anyway as we could then also look into essentially "streaming" large strings by just repeatedly invoking the user supplied callback with the next part of the string too.

> #### Statement expansion
> Have we not considered separating binding from execution? That could bite us later, no?

We have, its not part of this PR, but there is a `duckdb_v2_prepared_statement` in the fork.

> > Statements are parsed lazily, meaning that you can e.g. load a parser extension, and then use the new syntax immediately in the same SQL string.
> 
> I'd hold off on that, because currently the plan for both parser extension types is to require manually enabling them. And with dialect extensions there's another caviat because the entire query gets tokenized in its entirety before parsing further. We might still fix that before 2.0, but maybe not.

I think you've misunderstood, the lazy iteration is exactly what allows you to "manually enable" e.g. syntax extensions. You can pass `"LOAD my_ext; SET enable_special_syntax; SELECT foo WITH MY SYNTAX"` as one query, which will get expanded into three separate statements, which you can execute sequentially.

I don't know enough about dialect extensions to comment on that. Either way this design was not chosen specifically for parser extensions, but rather by the fact that having users manually split statements by looking for ";" by themselves, or not being able to distinguish if a query result came from a separate statement within the same query string, is very error prone.

Tishj:
"I'd hold off on that" was meant as: don't put this expectation around parser extensions in this description. It's still very much in flux ;)

Tishj:
> You can't create your own duckdb_v2_error_info_hande, so you can't really pass a non-null pointer (or well a pointer to a non-null pointer) unless you already had a previous error (which you should deal with it).
The whole point is that we dont allocate in the success-path, so there is no error info to free. But you can also just call duckdb_v2_error_info_destroy unconditionally as passing a pointer-to-null to destructors is safe.

These functions take a pointer, I imagine it's not strictly required to pass `&error_handle`, you can pass `NULL` to say you're not interested in the additional info right?

api_spec/v2/configuration/configuration.yaml
```diff
@@ -0,0 +1,288 @@
...
+
+  option_get_target_scope:
+    lifecycle:
+      - [ "stable", "v2.0.0", "2026-08-04" ]
+    description: |
+      Returns the option's target scope.
+
+      OPTION_TARGET_SCOPE_UNKNOWN for an option created via option_create until
+      it has been resolved through a database/connection get, and for one whose
+      declaration carries no explicit scope target.
```
staticlibs, line 217: > and for one whose declaration carries no explicit scope target.

Possibly this part needs to be removed. Unless I am missing something, new options cannot be declared through C API and in C++ API `AddExtensionOption` defaults the scope to `SESSION`.

api_spec/v2/metadata.yaml
```diff
@@ -0,0 +1,71 @@
...
+prefix: "duckdb_v2_"
+lifecycle_states:
+  stable:
+    visibility: always
+  deprecated:
+    visibility: opt_out
+    guard: DUCKDB_API_NO_DEPRECATED
+  unstable:
+    visibility: opt_in
+    guard: DUCKDB_EXTENSION_API_VERSION_UNSTABLE
```
okruitho, line 13: Update to the modern variant?

 same for `DUCKDB_API_NO_DEPRECATED`:
> For backwards compatability, the older `DUCKDB_API_NO_DEPRECATED` and `DUCKDB_EXTENSION_API_VERSION_UNSTABLE` macros
still work, and just set the new `DUCKDB_API_ALLOW_DEPRECATED`/`DUCKDB_API_ALLOW_UNSTABLE` macros accordingly.

api_spec/v2/error/error.yaml
```diff
@@ -0,0 +1,134 @@
...
+    return_type: ERROR
+
+  error_info_get_text:
+    lifecycle:
+      - [ "stable", "v2.0.0", "2026-08-04" ]
+    description: |
+      Retrieves the error message associated with an error info handle.
+
+      Returns a borrowed pointer to the info's null-terminated error message.
+      The pointer is owned by DuckDB and is valid until the info is destroyed;
```
okruitho, line 34: `The pointer is owned by DuckDB and is valid until the info is destroyed;` -> `..valid until the info is destroyed or a new text is set.`

api_spec/v2/error/error.yaml
```diff
@@ -0,0 +1,134 @@
...
+        type: error_info
+        description: "The error info handle to query."
+      out_code:
+        type: ERROR
+        indirection: 1
+        kind: OUT
+        description: "The error code associated with the info, if the call succeeds."
+    return_type: ERROR
+
+  error_info_get_text:
```
okruitho, line 27: It seems that `_text` and `*message` are used interchangeably. Might be a good idea to rectify that to `message` for example?

src/include/duckdb/main/capi_v2/capi_v2_internal.hpp
```diff
@@ -0,0 +1,445 @@
...
+// passed a non-null err out-parameter).
+struct CV2ErrorInfo {
+	DUCKDB_V2_ERROR code = DUCKDB_V2_ERROR_NONE;
+
+	// message: full "<Type> Error: <raw>" (ErrorData::Message()). raw_message: the
+	// body with that prefix stripped (ErrorData::RawMessage()), in the engine's
+	// rendered form (caret block, or JSON under errors_as_json); empty for a
+	// directly-set message. Both written on the error path (WithErrorHandler).
+	string message;
+	string raw_message;
```
okruitho, line 268: Nit: currently an empty string is considered no value. Maybe change that to use `std::optional` instead, to specify this case at a type level.

api_spec/v2/error/error.yaml
```diff
@@ -0,0 +1,134 @@
...
+        type: error_info
+        description: "The error info handle to query."
+      out_raw_message:
+        type: str
+        indirection: 1
+        kind: OUT_BORROW
+        description: "Receives a borrowed view of the raw message. Owned by DuckDB; valid until the info handle is destroyed."
+    return_type: ERROR
+
+  error_info_set_code:
```
okruitho, line 77: No dedicated test to this api call, same for error_info_get_code

api_spec/v2/database/database.yaml
```diff
@@ -0,0 +1,202 @@
...
+  open:
+    lifecycle:
+      - [ "stable", "v2.0.0", "2026-08-04" ]
+    description: |
+      Opens a database under the environment.
+
+      `path` semantics:
+        - `nullptr`, ":memory:", or any ":memory:..." path opens a fresh
+          in-memory database. In-memory databases are never deduplicated
+          across calls.
```
staticlibs, line 12: > In-memory databases are never deduplicated across calls.

I think this part is incorrect for `:memory:tag1` paths, tagged memory databases are supposed to be cached/deduplicated.

api_spec/v2/error/error.yaml
okruitho: In `SetErrorInfo` it could have been used, but the values are set directly instead of via the c-api.

api_spec/v2/error/error.yaml
okruitho: > description: |
      Retrieves the error message associated with an error info handle.

api_spec/v2/database/database.yaml
```diff
@@ -0,0 +1,202 @@
...
+      `options` is an optional array of pre-created config option handles
+      applied at construction time. Only the name and setting of each handle
+      are read; the remaining fields are ignored. Options may include pre-open
+      database settings (e.g. access_mode, memory_limit) and ATTACH-style
+      storage options (e.g. BLOCK_SIZE, ENCRYPTION_KEY). Unknown names are
+      accepted and kept for a later extension to consume. Pass
+      `options=nullptr` and `option_count=0` to open with defaults.
+
+      LOCAL_ONLY options are rejected at this scope, as they are by
+      database_option_set.
```
staticlibs, line 26: > as they are by database_option_set.

Suggest to add word "rejected": "as they are rejected by database_option_set"

api_spec/v2/connection/connection.yaml
```diff
@@ -0,0 +1,316 @@
...
+          `SET name = setting`.
+        - GLOBAL writes through to the database, visible to all connections,
+          like `SET GLOBAL`.
+        - LOCAL writes to this connection's session only, like `SET LOCAL` /
+          `SET SESSION`.
+      A disallowed combination returns ERROR_INPUT_INVALID: GLOBAL against a
+      LOCAL_ONLY option, LOCAL against a GLOBAL_ONLY one, and the legacy
+      analogues. Unknown option names are rejected; to set an option whose
+      defining extension has not loaded yet, use database scope or the startup
+      options of open.
```
staticlibs, line 72: > Unknown option names are rejected; to set an option whose defining extension has not loaded yet, use database scope or the startup options of open.

This description is ambiguous, need clarifications:

 - are `GLOBAL` unknowns rejected or not?
 - "database scope" is `GLOBAL` scope to this call or `database_option_set` call

api_spec/v2/connection/connection.yaml
```diff
@@ -0,0 +1,316 @@
...
+        description: "Optional. On failure, receives an opaque info handle the caller must destroy via error_info_destroy."
+    return_type: ERROR
+  # Query processing
+  connection_interrupt:
+    lifecycle:
+      - [ "stable", "v2.0.0", "2026-08-04" ]
+    description: |
+      Interrupts the query currently executing on the connection.
+
+      The cross-thread cancellation entry point for streaming results: safe to
```
staticlibs, line 185: > The cross-thread cancellation entry point for streaming results

This formulation is confusing, I believe that interruption is supported for running queries which results are going to be materialized on query completion.

api_spec/v2/connection/connection.yaml
```diff
@@ -0,0 +1,316 @@
...
+        indirection: 1
+        kind: OUT
+        description: "Receives the option count."
+      err:
+        type: error_info
+        indirection: 1
+        kind: OUT
+        description: "Optional. On failure, receives an opaque info handle the caller must destroy via error_info_destroy."
+    return_type: ERROR
+  connection_option_get_by_index:
```
staticlibs, line 148: This function is not covered by tests.

api_spec/v2/connection/connection.yaml
```diff
@@ -0,0 +1,316 @@
...
+        indirection: 1
+        kind: OUT
+        description: "Receives the populated option handle."
+      err:
+        type: error_info
+        indirection: 1
+        kind: OUT
+        description: "Optional. On failure, receives an opaque info handle the caller must destroy via error_info_destroy."
+    return_type: ERROR
+  connection_option_get_count:
```
staticlibs, line 123: This function is not covered by tests.

api_spec/v2/sql_statement/sql_statement.yaml
```diff
@@ -0,0 +1,252 @@
...
+      rewrites happen inside statement_execute. The SQL string is copied, so
+      the caller may free it once this call returns. An input with no
+      statements — empty, whitespace, or separators only — yields an iterator
+      that is immediately exhausted.
+    role: constructor
+    belongs_to: statement_iterator
+    parameters:
+      conn:
+        type: connection
+        description: The connection supplying the parser state.
```
staticlibs, line 57: > parser state

This state is described later, but a bit confusing at this point, suggest to change to "parser configuration"

api_spec/v2/sql_statement/sql_statement.yaml
```diff
@@ -0,0 +1,252 @@
...
+      DEFAULT) cannot be bound and is rejected with ERROR_INPUT_INVALID;
+      execute it instead.
+
+      *out_schema and *out_parameters are set to NULL on failure.
+    role: method
+    belongs_to: sql_statement
+    parameters:
+      conn:
+        type: connection
+        description: The connection supplying the catalog, transaction, and parser state.
```
staticlibs, line 138: Also suggest to change to "parser configuration".

api_spec/v2/query_result/query_result.yaml
```diff
@@ -0,0 +1,571 @@
...
+    belongs_to: result
+    parameters:
+      result:
+        type: result
+        description: The result to step.
+      out_chunk:
+        type: data_chunk
+        indirection: 1
+        kind: OUT
+        description: Receives an owned chunk iff *out_status is CHUNK; set to nullptr otherwise.
```
staticlibs, line 308: "if" typo

staticlibs (commented):
Added a few comments, mostly unclear documentation, references:

 - duckdblabs/duckdb-internal#10502
 - duckdblabs/duckdb-internal#10505

api_spec/v2/vector/vector.yaml
```diff
@@ -0,0 +1,861 @@
...
+#       /* read via the appropriately-typed cast of view.data */
+#   }
+#
+# view.data semantics by kind; the caller dispatches on the logical type. The
+# storage units for the temporal kinds — DATE in days since 1970-01-01, TIME /
+# TIMESTAMP / TIMESTAMP_MS / TIMESTAMP_NS and their variants as signed offsets
+# from the epoch in their named unit — are the ones DuckDB uses throughout, and
+# the logical type module carries the full table.
+#
+#   - primitives (BOOL, INT*, UINT*, FLOAT, DOUBLE, DATE, TIME*, TIMESTAMP*)
```
maiadegraaf, line 30: Nit: The wildcard notation is confusing mixed with the pointer right underneath.
```suggestion
#   - primitives (BOOL, INT(8/16/32/64), UINT(8/16/32/64), FLOAT, DOUBLE, DATE, TIME, TIME_TZ, TIMESTAMP(_S/_MS/_NS/_TZ))
```

abramk:
It was worth reading through the PR description just to encounter the phrase "clanker-generated vomit." 😆 I'm going to use that!

test/api/capi/v2/test_capi_v2_logical_type.cpp
```diff
@@ -0,0 +1,1629 @@
...
+	auto mood = V2TypeFromText(f.conn, "mood");
+	RequireParamRoundTrip(f.conn, mood);
+	duckdb_v2_logical_type_destroy(&mood);
+}
+
+// ===========================================================================
+// Volume measurement (hidden): run explicitly via "[capi_v2_bench]"
+// ===========================================================================
+
+TEST_CASE("V2 bench: 100k-entry enum inspection cost", "[.][capi_v2_bench]") {
```
taniabogatsch, line 1579: Should add benchmarks in `benchmark/micro/`. For example: https://github.com/duckdb/duckdb/blob/main/benchmark/micro/bulkupdate.cpp

api_spec/v2/vector/vector.yaml
```diff
@@ -0,0 +1,861 @@
...
+        description: Optional. On failure, receives an opaque info handle the caller must destroy via error_info_destroy.
+    return_type: ERROR
+  vector_set_size:
+    lifecycle:
+      - [ "stable", "v2.0.0", "2026-08-04" ]
+    description: |
+      Sets the number of elements in the vector.
+
+      The counterpart of vector_get_size: it declares how many logical elements
+      the vector now holds, and does not allocate or initialize anything.
```
maiadegraaf, line 276: `vector_set_size` does reserve extra space, wouldn't that fall under allocating?

```cpp
DUCKDB_V2_ERROR duckdb_v2_vector_set_size(duckdb_v2_vector_handle vector, idx_t size,
                                          duckdb_v2_error_info_handle *err) {
...
		// Grow the underlying buffer first so the new logical size fits; constant vectors
		// carry a single physical element and must not be reserved against.
		if (vec->GetVectorType() == duckdb::VectorType::FLAT_VECTOR) {
			vec->Reserve(size);
		}
...
}
```

api_spec/v2/vector/vector.yaml
```diff
@@ -0,0 +1,861 @@
...
+    description: |
+      Repoints a vector at another vector's data and type, without copying.
+
+      `vector` takes on the storage and logical type of `source`: no data
+      moves, and the two alias the same buffers until one of them is reset or
+      re-referenced. Works for any type, nested included. The source's data
+      must outlive every read of `vector`. Use it to hand an
+      already-materialized vector — a chunk column produced by
+      arrow_array_to_data_chunk, say — straight to an output vector without a
+      per-row copy.
```
maiadegraaf, line 435: `arrow_array_to_data_chunk` doesn't exist.


```suggestion
      must outlive every read of `vector`. Use it to pass an
      already-materialized vector to an output vector without a
      per-row copy.
```

api_spec/v2/vector/vector.yaml
```diff
@@ -0,0 +1,861 @@
...
+  # The two codecs take plain (data, length) byte ranges rather than a handle,
+  # which makes them the single BIGNUM codec for both values and vectors, in
+  # both directions. To read: resolve a vector's bignum payload the same way
+  # as VARCHAR or BLOB, dispatching inline vs pointer on the transparent bytes
+  # type, or read a BIGNUM value's storage with value_get_blob, then decode.
+  # To write: encode the magnitude, then write the BIGNUM vector exactly like
+  # a VARCHAR or BLOB one — size the buffer, allocate it out of the vector's
+  # arena, encode into it, and assemble the bytes payload.
+  # -----------------------------------------------------------------------
+  bignum_decode:
```
maiadegraaf, line 736: `bignum_decode` and `bignum_encode` are both missing `role` and `belongs_to`.

api_spec/v2/data_chunk/data_chunk.yaml
```diff
@@ -0,0 +1,149 @@
...
+      the vector is populated. The chunk is caller-owned and must be destroyed
+      via data_chunk_destroy.
+    role: constructor
+    belongs_to: data_chunk
+    parameters:
+      types:
+        type: logical_type
+        indirection: 1
+        const: true
+        description: Pointer to an array of logical_type handles, one per column.
```
maiadegraaf, line 39: Maybe worth specifing that the types cannot contain `LogicalTypeId::ANY`.

api_spec/v2/logging/logging.yaml
```diff
@@ -0,0 +1,66 @@
...
+        value: 40
+        description: "A recoverable problem."
+      LOG_LEVEL_ERROR:
+        value: 50
+        description: "An operation failed."
+      LOG_LEVEL_FATAL:
+        value: 60
+        description: "An unrecoverable failure."
+functions:
+  context_log:
```
maiadegraaf, line 34: There are no tests for this function.

maiadegraaf (commented):
Added some comments on the `data_chunk`, `logging`, and `vector` files.

src/include/duckdb_extension_v2.h
taniabogatsch: I think it's fine to leave as-is, at least I don't know what else to name it? I don't want to rename some of them to `src`/`dest`, when they come in/out, or just rename it for capacity? `Bytes available in out_data. Ignored when out_data is NULL.` -> would you improve the description?

src/include/duckdb_extension_v2.h
taniabogatsch: Done

src/include/duckdb_v2.h
taniabogatsch: Tracked in an issue/generator change.

api_spec/v2/error/error.yaml
taniabogatsch: Done

api_spec/v2/error/error.yaml
taniabogatsch: Done

api_spec/v2/metadata.yaml
taniabogatsch: I've replaced `DUCKDB_EXTENSION_API_VERSION_UNSTABLE`, but `DUCKDB_API_NO_DEPRECATED` is a too involved generator change for now, will leave it as-it and just make a note in the generator-changes issue.

src/include/duckdb/main/capi_v2/capi_v2_internal.hpp
taniabogatsch: Done

api_spec/v2/error/error.yaml
taniabogatsch: Done

api_spec/v2/configuration/configuration.yaml
taniabogatsch: Done

api_spec/v2/database/database.yaml
taniabogatsch: Has that been implemented yet? I had assumed that that was planned, but had not yet landed. 🤔

api_spec/v2/database/database.yaml
taniabogatsch: Done

api_spec/v2/database/database.yaml
taniabogatsch: Indeed seems to be missing/an oversight and working in C API V1. Will create an issue to fix that, thanks for catching it!

taniabogatsch: https://github.com/duckdblabs/duckdb-internal/issues/10799

api_spec/v2/connection/connection.yaml
taniabogatsch: Done

api_spec/v2/connection/connection.yaml
taniabogatsch: Done

api_spec/v2/connection/connection.yaml
taniabogatsch: Done

api_spec/v2/sql_statement/sql_statement.yaml
taniabogatsch: Done

api_spec/v2/sql_statement/sql_statement.yaml
taniabogatsch: Done

api_spec/v2/query_result/query_result.yaml
taniabogatsch: Done

test/api/capi/v2/test_capi_v2_logical_type.cpp
taniabogatsch: noted in separate collection issue for remaining chores

api_spec/v2/connection/connection.yaml
taniabogatsch: noted in separate collection issue for remaining chores

api_spec/v2/vector/vector.yaml
taniabogatsch: Done

api_spec/v2/vector/vector.yaml
taniabogatsch: Done

api_spec/v2/vector/vector.yaml
taniabogatsch: Done

api_spec/v2/vector/vector.yaml
taniabogatsch: Tracking in separate issue.

api_spec/v2/data_chunk/data_chunk.yaml
taniabogatsch: Hmmm I think we want to stay a bit more agnostic for now 🤔

api_spec/v2/logging/logging.yaml
taniabogatsch: Tracked in separate issue.

api_spec/v2/database/database.yaml
evertlammerts: Replied on the ticket

api_spec/v2/connection/connection.yaml
evertlammerts: The docstring is wrong, and not just here but also in database_option_set. As far as I can see, it is just not possible on any path to pass in unknown config options.

The database `open` path allows passing in extension options IFF they are present in `EXTENSION_SETTINGS` (in src/include/duckdb/main/extension_entries.hpp) AND autoload is switched on. But other than that, there's no such thing as "unknown option names".

Issue #24439, opened 2026-08-03
HISTOGRAM window combine double-counts NaN map keys


Yibo-Dong:
### What happens?

`HISTOGRAM(x)` over a full frame should match the scalar aggregate over the same visible rows. With three rows `(0.0, NaN, 1.0)`, DuckDB's scalar histogram and the separate window path both return `{0.0=2, 1.0=1}` for every row. The default window path instead returns `{0.0=3}` for the row whose input value is `NaN`.

### To Reproduce

```sql
WITH t(id, x) AS (
  VALUES
    (1, 0.0::DOUBLE),
    (2, 'NaN'::DOUBLE),
    (3, 1.0::DOUBLE)
)
SELECT
  id,
  HISTOGRAM(x) OVER (
    ORDER BY id
    ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
    EXCLUDE TIES
  ) AS actual
FROM t
ORDER BY id;
```

Run:

```sh
./build/release/duckdb -f bug.sql
```

Expected window result:

| id | expected |
| ---: | --- |
| 1 | `{0.0=2, 1.0=1}` |
| 2 | `{0.0=2, 1.0=1}` |
| 3 | `{0.0=2, 1.0=1}` |

Actual window execution:

| id | actual |
| ---: | --- |
| 1 | `{0.0=2, 1.0=1}` |
| 2 | `{0.0=3}` |
| 3 | `{0.0=2, 1.0=1}` |

### Notes From Minimization

- `SET debug_window_mode='separate'` eliminates the mismatch and returns the
  expected histogram for every row.
- `SET debug_window_mode='combine'` does not eliminate the mismatch; it retains
  `{0.0=3}` for `id = 2`.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (a1aacc73df)

### DuckDB Client:

CLI

### Full Name:

Yibo Dong

### Affiliation:

National University of Singapore

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

hawkfish:
FWIW this can happen with regular aggregation too.

hawkfish:
[...]

PR #24703, opened 2026-08-11
Issue #24439: HISTOGRAM NaN Support

hawkfish:
* Use a specialised Less parameter in the map for floating point types.


hawkfish:
Looks like this got fixed in parallel.

Issue #24333, opened 2026-07-30
Windows builds fail on concurrent_checkpoint_insert.test


xevix:
### What happens?

Windows builds seem to fail on and off on this test. I've seen it on 32-bit and 64-bit, seems to have something to do with parallelism. Attempted to reproduce on a local Windows 11 VM but no success, so might be something to do with the GitHub Actions runners setup.

https://github.com/xevix/duckdb/actions/runs/30502306429/job/90756331713 (32 bit)
https://github.com/xevix/duckdb/actions/runs/30478885588/job/90680356308 (64 bit)
https://github.com/duckdb/duckdb/pull/24249 (mentioned)

```
================================================================
error: FAIL test/sql/parallelism/interquery/concurrent_checkpoint_insert.test

  > 17  query I
    18  SELECT COUNT(*) FROM db{j}.tbl where key = {i}
    19  ----
    20  1

1. test/sql/parallelism/interquery/concurrent_checkpoint_insert.test:17
================================================================================

Error: Wrong result in query! (test/sql/parallelism/interquery/concurrent_checkpoint_insert.test:17)!
================================================================================
SELECT COUNT(*) FROM db32.tbl where key = 6;
================================================================================
Mismatch on row 1, column count_star()(index 1)
2 <> 1
================================================================================
Expected result:
================================================================================
1

================================================================================
Actual result:
================================================================================
2
```

### To Reproduce

Unable to consistently reproduce with specific code, but has happened on at least one of my PRs as well as someone else's.

https://github.com/xevix/duckdb/actions/runs/30502306429/job/90756331713 (32 bit)
https://github.com/xevix/duckdb/actions/runs/30478885588/job/90680356308 (64 bit)
https://github.com/duckdb/duckdb/pull/24249 (mentioned)

### OS:

Windows

### DuckDB Version:

HEAD

### DuckDB Client:

N/A

### Full Name:

Alejandro Wainzinger

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

PR #24704, opened 2026-08-11
Fix interval comparison when carrying left leaves a negative field

### LINKED ISSUE #24631: RANGE PRECEDING over an INTERVAL order key can exclude the lower endpoint
### What happens?

`RANGE PRECEDING` over an `INTERVAL` order key can exclude the inclusive lower endpoint when the current value crosses the 30-day interval normalization boundary.

This happens on current main (`e500d77864fe`) and also on v1.5.4.

#### Code

```sql
WITH src(id, k, x) AS (
  VALUES
    (1, INTERVAL 26 DAY, 10),
    (2, INTERVAL 30 DAY, 1)
)
SELECT
  id,
  k,
  SUM(x) OVER (
    ORDER BY k
    RANGE BETWEEN INTERVAL 4 DAY PRECEDING AND CURRENT ROW
  ) AS range_sum
FROM src
ORDER BY id;
```

Run:

```sh
```

#### Expected

| id | k | range_sum |
| ---: | --- | ---: |
| 1 | 26 days | 10 |
| 2 | 30 days | 11 |
| eq | between_ok |
| --- | --- |
| true | true |

#### Actual

| id | k | range_sum |
| --- | --- | --- |
| 1 | 26 days | 10 |
| 2 | 30 days | 1 |

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (`e500d77864fe`)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

kid-ye:
Fixes #24631.

### The bug

`RANGE PRECEDING` over an `INTERVAL` order key excludes the inclusive lower endpoint once the current row's value reaches 30 days:

```sql
WITH src(id, k, x) AS (VALUES (1, INTERVAL 26 DAY, 10), (2, INTERVAL 30 DAY, 1))
SELECT id, SUM(x) OVER (ORDER BY k RANGE BETWEEN INTERVAL 4 DAY PRECEDING AND CURRENT ROW)
FROM src ORDER BY id;
-- id 2 returns 1, expected 11
```

The window is only where it surfaces. The defect is in interval comparison and reproduces with no window at all:

```sql
SELECT INTERVAL 1 MONTH = INTERVAL 30 DAY;                      -- true
SELECT (INTERVAL 1 MONTH - INTERVAL 4 DAY) = INTERVAL 26 DAY;   -- false, expected true
SELECT (INTERVAL 1 DAY  - INTERVAL 1 HOUR) = INTERVAL 23 HOUR;  -- false, expected true
SELECT (INTERVAL 2 MONTH - INTERVAL 40 DAY) = INTERVAL 20 DAY;  -- false, expected true
```

If `1 month == 30 days`, then `1 month - 4 days` has to equal `26 days`.

### Root cause

`interval_t::Normalize` carries micros into days and days into months with truncating division, so a negative remainder is never borrowed. `1 month -4 days` stays unnormalized, and the month-major comparison then ranks it above `26 days`.

The binder rewrites a RANGE boundary as `normalized_interval(k) - offset`. For `k = 30 days` that boundary becomes `1 month -4 days` instead of `26 days`, so the frame start lands past the row that should be included.

### The fix

Borrow after each carry so the remainders are never negative. `Normalize` now yields a genuine canonical form — `months`, `days ∈ [0,30)`, `micros ∈ [0, 1 day)` — which is what the lexicographic comparison already assumed. Equality, ordering, hashing and VARIANT comparison encoding all route through it, so they agree with each other.

`Borrow()` multiplied the out-of-range amount by the field scale with no overflow check. That product only just fitted for the extreme interval, and the extra borrow above pushed it past `INT64_MIN`; UBSan caught it. It now saturates.

Display is unaffected — it prints the raw fields, so `1 month -4 days` still renders as written. It just no longer compares greater than `26 days`.

### Testing

Added to `test_interval_comparison.test`: borrow cases in both signs, plus the saturated extremes covering the `Borrow` path. Added the reported query to `test_window_interval.test`.

Also property-tested locally — 540,225 pairs built from explicit `(months, days, micros)` components across negative and mixed signs, checking `<`, `=`, `>`, `<=` against the true total computed in `HUGEINT`. Zero mismatches. Confirmed hashing agrees with equality: differently-written equal intervals collapse to a single group under `GROUP BY` and `COUNT(DISTINCT)`.

### CI

86 jobs pass on my fork, including Relassert/UBSan, Thread Sanitizer and Vector Sizes. Three failures look unrelated:

- `Linux Config (Execution)` — `duckdb_eviction_queues.test` times out under `verify_fetch_row`; the test uses no intervals and reproduces as a timeout on a clean checkout
- `windows / MinGW (64 Bit)` — the known flaky `concurrent_checkpoint_insert.test` (#24333); the MSVC 64-bit job passes
- `extensions / Main (search) / osx_amd64` — vcpkg failed to fetch zlib (`curl: (60) SSL certificate problem`)

### DIFF
--- a/src/include/duckdb/common/types/interval.hpp
+++ b/src/include/duckdb/common/types/interval.hpp
@@ -181,15 +181,24 @@ public:
 void interval_t::Normalize(int64_t &months, int64_t &days, int64_t &micros) const {
 	auto &input = *this;
 
-	//  Carry left
+	//  Carry left, borrowing so that the remainders are never negative.
+	//  Truncating division would leave e.g. 1 month -4 days unnormalized, which then compares above 26 days.
 	micros = input.micros;
 	int64_t carry_days = micros / Interval::MICROS_PER_DAY;
 	micros -= carry_days * Interval::MICROS_PER_DAY;
+	if (micros < 0) {
+		--carry_days;
+		micros += Interval::MICROS_PER_DAY;
+	}
 
 	days = input.days;
 	days += carry_days;
 	int64_t carry_months = days / Interval::DAYS_PER_MONTH;
 	days -= carry_months * Interval::DAYS_PER_MONTH;
+	if (days < 0) {
+		--carry_months;
+		days += Interval::DAYS_PER_MONTH;
+	}
 
 	months = input.months;
 	months += carry_months;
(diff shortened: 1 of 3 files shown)


hawkfish (approved):
Thanks - LGTM!

PR #24706, opened 2026-08-11
Simplify filters when entire enum domain is present in IN clause

### LINKED ISSUE #24417: ENUM exhaustive domain not pruned
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TYPE mood AS ENUM ('sad', 'ok', 'happy');
memory D CREATE TABLE t(m mood NOT NULL);
memory D INSERT INTO t
         SELECT CASE range % 3
           WHEN 0 THEN 'sad'
           WHEN 1 THEN 'ok'
           ELSE 'happy'
         END::mood
         FROM range(6144);

-- the enum IN predicate is semantically impossible, but still reads row group
memory D EXPLAIN ANALYZE
         SELECT count(*) FROM t
         WHERE m NOT IN ('sad', 'ok', 'happy');
╭─ Summary ───────────╮
│ Total Time: 0.0004s │
╰─────────────────────╯
│ NOT (m IN ('sad', 'ok', 'happy')) │
│ Row Groups Scanned: 1 / 1         │
│ 0 rows                        0µs │

```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

jaiswalvikas9991:
Fixes <https://github.com/duckdb/duckdb/issues/24417>

Enum IN simplification can map VARCHAR literals back to ENUM constants. When the valid literals cover the full ENUM domain, rewrite the predicate to constant_or_null so filters can be simplified to true or false when the input is known non-null.

Add changes to detect enum domain exhaustion in InEnumSimplificationRule::Apply. Add ConstantOrNullSimplification to eliminate the entire filter in cases where the condition boils down to "true" or "false". This elimination is only done when the query has no side effects.

Add regression coverage for full-domain ENUM IN and NOT IN filters, nullable inputs, and DML CTE null handling.

### DIFF
--- a/src/include/duckdb/optimizer/constant_or_null_simplification.hpp
+++ b/src/include/duckdb/optimizer/constant_or_null_simplification.hpp
@@ -1,0 +1,32 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// duckdb/optimizer/constant_or_null_simplification.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/planner/logical_operator.hpp"
+
+namespace duckdb {
+class ClientContext;
+class NotNullExpressionAnalyzer;
+
+class ConstantOrNullSimplification {
+public:
+	explicit ConstantOrNullSimplification(ClientContext &context);
+
+	unique_ptr<LogicalOperator> Optimize(unique_ptr<LogicalOperator> op);
+
+private:
+	unique_ptr<Expression> SimplifyExpression(LogicalOperator &input, unique_ptr<Expression> expr,
+	                                          NotNullExpressionAnalyzer &analyzer);
+	unique_ptr<LogicalOperator> OptimizeFilter(unique_ptr<LogicalOperator> op);
+
+private:
+	ClientContext &context;
+};
+
+} // namespace duckdb
--- a/src/include/duckdb/optimizer/constant_or_null_simplification.hpp
+++ b/src/include/duckdb/optimizer/constant_or_null_simplification.hpp
@@ -21,9 +21,10 @@ public:
 	unique_ptr<LogicalOperator> Optimize(unique_ptr<LogicalOperator> op);
 
 private:
+	unique_ptr<LogicalOperator> OptimizeInternal(unique_ptr<LogicalOperator> op, bool plan_has_side_effects);
 	unique_ptr<Expression> SimplifyExpression(LogicalOperator &input, unique_ptr<Expression> expr,
-	                                          NotNullExpressionAnalyzer &analyzer);
-	unique_ptr<LogicalOperator> OptimizeFilter(unique_ptr<LogicalOperator> op);
+	                                          NotNullExpressionAnalyzer &analyzer, bool allow_folding);
+	unique_ptr<LogicalOperator> OptimizeFilter(unique_ptr<LogicalOperator> op, bool plan_has_side_effects);
 
 private:
 	ClientContext &context;
--- a/src/optimizer/CMakeLists.txt
+++ b/src/optimizer/CMakeLists.txt
@@ -19,6 +19,7 @@ add_library_unity(
   builtin_function_lookup.cpp
   column_binding_replacer.cpp
   column_lifetime_analyzer.cpp
+  constant_or_null_simplification.cpp
   empty_result_pullup.cpp
   common_aggregate_optimizer.cpp
   common_subplan_optimizer.cpp
--- a/src/optimizer/constant_or_null_simplification.cpp
+++ b/src/optimizer/constant_or_null_simplification.cpp
@@ -1,0 +1,167 @@
+#include "duckdb/optimizer/constant_or_null_simplification.hpp"
+
+#include "duckdb/function/scalar/generic_common.hpp"
+#include "duckdb/optimizer/expression_rewriter.hpp"
+#include "duckdb/planner/expression/bound_constant_expression.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/planner/expression/bound_operator_expression.hpp"
+#include "duckdb/planner/expression_iterator.hpp"
+#include "duckdb/planner/expression_nullability.hpp"
+#include "duckdb/planner/operator/logical_empty_result.hpp"
+#include "duckdb/planner/operator/logical_filter.hpp"
+
+namespace duckdb {
+
+ConstantOrNullSimplification::ConstantOrNullSimplification(ClientContext &context_p) : context(context_p) {
+}
+
+static optional<bool> GetBooleanConstant(const Expression &expr) {
+	if (expr.GetExpressionClass() != ExpressionClass::BOUND_CONSTANT) {
+		return optional<bool>();
+	}
+
+	auto &constant = expr.Cast<BoundConstantExpression>().GetValue();
+	if (constant.IsNull() || constant.type().id() != LogicalTypeId::BOOLEAN) {
+		return optional<bool>();
+	}
+
+	return BooleanValue::Get(constant);
+}
+
+static optional<bool> GetConstantOrNullBoolean(Expression &expr) {
+	if (expr.GetExpressionClass() != ExpressionClass::BOUND_FUNCTION ||
+	    expr.GetReturnType().id() != LogicalTypeId::BOOLEAN) {
+		return optional<bool>();
+	}
+
+	auto &func = expr.Cast<BoundFunctionExpression>();
+	if (ConstantOrNull::IsConstantOrNull(func, Value::BOOLEAN(true))) {
+		return true;
+	}
+
+	if (ConstantOrNull::IsConstantOrNull(func, Value::BOOLEAN(false))) {
+		return false;
+	}
+
+	return optional<bool>();
+}
+
+static bool ConstantOrNullInputsAreNotNull(LogicalOperator &input, BoundFunctionExpression &func,
+                                           NotNullExpressionAnalyzer &analyzer) {
+	auto &children = func.GetChildren();
+	D_ASSERT(children.size() >= 2);
+
+	for (idx_t child_idx = 1; child_idx < children.size(); ++child_idx) {
+		if (children[child_idx]->GetExpressionClass() == ExpressionClass::BOUND_CONSTANT) {
+			auto &constant = children[child_idx]->Cast<BoundConstantExpression>().GetValue();
+			if (!constant.IsNull()) {
+				continue;
+			}
+		}
+
+		if (!analyzer.IsNotNull(input, *children[child_idx])) {
+			return false;
+		}
+	}
+
+	return true;
+}
+
+unique_ptr<Expression> ConstantOrNullSimplification::SimplifyExpression(LogicalOperator &input,
+                                                                        unique_ptr<Expression> expr,
+                                                                        NotNullExpressionAnalyzer &analyzer) {
+	ExpressionIterator::EnumerateChildren(*expr, [&](unique_ptr<Expression> &child) {
+		child = SimplifyExpression(input, std::move(child), analyzer);
+	});
+
+	if (expr->GetExpressionClass() == ExpressionClass::BOUND_FUNCTION) {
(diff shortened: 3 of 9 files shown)


jaiswalvikas9991:
I added the constant_or_null simplification under the existing OptimizerType::EXPRESSION_REWRITER rather than introducing a new OptimizerType for it. I am not sure if adding a new optimizer type is appropriate for this bug fix, since that would add more surface area than the actual fix needs. If you prefer this pass to have its own OptimizerType, I can split it out.

JaySon-Huang:
Hi @jaiswalvikas9991!

I also came across #24417 and noticed your PR here. Beyond the expression rewrite, I think this can also be approached from the statistics side. #25133 teaches `NumericStats::ConstantsCoverRange` about ENUM's closed domain and classifies the NOT of a full-domain `IN` as false-or-null, so row groups are pruned directly.

The two routes complement each other more than they conflict

- Your rewrite fully removes the filter for the positive `IN` form and works without column statistics, which my change does not
- Mine additionally collapses `NOT IN` over the full domain for nullable columns and inside DML plans, where `HasSideEffects()` gates your simplification pass

I verified both behaviors against your branch.

If you find the time, I'd be happy for us to review each other's PRs and exchange thoughts on the two approaches.

jaiswalvikas9991:
@JaySon-Huang Thanks for taking a look at my PR. I looked through your PR locally as well, and it looks good to me.

I also think the two changes are complementary, and I agree with both of your points.


  > Mine additionally collapses NOT IN over the full domain for nullable columns and inside DML plans, where HasSideEffects() gates your simplification pass

You are right that my PR was not collapsing the `NOT IN` case. Your comment made me realize I was missing a case in `src/optimizer/constant_or_null_simplification.cpp`: I was not simplifying `NOT constant_or_null(true, ...)` to `constant_or_null(false, ...)`, so it was not being simplified in later stages. I have fixed that now.

  My PR still does not handle the DML case, since it works in the expression rewriter phase and is guarded by `HasSideEffects()`. Your PR handles that in the filter/statistics pruning path,  so I think both changes can be useful independently.

There may be overlap in the final plan for full-domain ENUM `NOT IN`, and if one PR lands first the other may need some `EXPLAIN` expectation updates. But I think the underlying fixes are at different layers.

If you think there is a good way to combine the PRs, I am open to that. Otherwise, I think they can be merged independently if the maintainers agree.

JaySon-Huang:
@jaiswalvikas9991 Everything else looks good to me. I will leave the rest to the maintainers. 👍

JaySon-Huang:
Hi @jaiswalvikas9991, @lnkuiper — I went back to this PR after receiving @lnkuiper's comment on #25133:

> Could we extend the rewrite approach instead? It should cover ENUM as well as VARCHAR, and handle the safe false-or-NULL predicate-root case in DML without a plan-wide HasSideEffects() check.

Re-examining this PR against that question surfaced one correctness issue and one way to narrow the DML side, both with fixes and tests on this PR: https://github.com/jaiswalvikas9991/duckdb/pull/1

**1. Correctness: folding ignores uncommitted appends of the same transaction**

`ConstantOrNullSimplification` folds `constant_or_null(c, x)` into the constant `c` when `NotNullExpressionAnalyzer` proves `x` non-NULL. That proof comes from `table->GetStatistics()`, which only covers **committed** row groups. Rows appended by an *earlier statement of the same transaction* live in `LocalStorage`: the scan sees them, the statistics do not — and since the query itself contains no DML, the `!plan->HasSideEffects()` gate does not catch it. The fold then drops the filter and the freshly appended NULL rows slip through:

```sql
CREATE TABLE t(i INTEGER);
BEGIN;
INSERT INTO t VALUES (NULL);
-- i * 2 <> 5 rewrites to constant_or_null(TRUE, i) via move_constants
SELECT count(*) FROM t WHERE i * 2 <> 5;  -- returns 4, expected 3
```

The same applies to the full-domain ENUM `IN` producer added here (`m IN (full domain)` → `constant_or_null(TRUE, m)`). The fix is on the analyzer side: before trusting statistics in the `LOGICAL_GET` branch, check whether the current transaction holds uncommitted appends for that table, and refuse the NOT-NULL proof if so. That closes the hole for all three consumers at once (`ConstantOrNullSimplification`, `Deliminator`, `PushdownMarkJoin`), not just the fold in this PR.

**2. DML: the safe false-or-NULL root does not need a plan-wide gate — but the fold does**

Rather than removing the plan-wide `HasSideEffects()` check (which would reintroduce issue 1 for e.g. DML CTEs), I split the pass into two capabilities:

- *Shape-only rewrites* — `NOT(constant_or_null(true, x)) → constant_or_null(false, x)` and NOT-over-constant folding — keep every input evaluated per row and never consult statistics, so they are safe on all plans, DML included. On top of that, `FilterStatisticsOptimizer` already prunes `constant_or_null(false, x)` as `FILTER_FALSE_OR_NULL` *unconditionally* (no row can pass), so full-domain `NOT IN` collapses to an empty result inside `DELETE`/`UPDATE` plans without any stats dependency.
- *Folding to a constant* drops the per-row evaluation of the inputs, so it stays gated on the plan having no side effects.

So the direction @lnkuiper suggested is achievable for the false-or-NULL root across both producers (ENUM here, and the non-ENUM `move_constants` producer `i * 2 <> 5`), and the remaining plan-wide gate only protects the fold-to-TRUE direction, which genuinely relies on the statistics proof.

Tests covering both points (the stale-statistics scenario as a red known-issue first, then the fix, plus the DML guard behavior) live on my branch ([verify branch link]), with the analyzer fix being the key change.

JaySon-Huang:
This correctness fix also covers queries like the one reported in #25266. I've opened a separate PR for it: https://github.com/duckdb/duckdb/pull/25277

lnkuiper:
Could you have a look at failing CI? You probably need to use `.Cast<...>()` instead of `dynamic_cast`.

Queries like this still perform a sequential scan (only works for `VARCHAR` currently):
```sql
m NOT IN ('sad'::mood, 'ok'::mood, 'happy'::mood)
```

Maybe we should wait until https://github.com/duckdb/duckdb/pull/25277 lands ?

jaiswalvikas9991:
@lnkuiper  Thanks, I fixed the failing no-RTTI CI issue by replacing the `dynamic_cast<DuckTableEntry *>` with DuckDB's `IsDuckTable()` / `.Cast<DuckTableEntry>()`.

I also added support for the typed ENUM RHS case:

```sql
m NOT IN ('sad'::mood, 'ok'::mood, 'happy'::mood)
```

This now takes the same path as the VARCHAR-literal form.


> Maybe we should wait until https://github.com/duckdb/duckdb/pull/25277 lands ?

This PR currently keeps the conservative guard `TableHasUncommittedAppends` against using table statistics for the `constant_or_null(true, ...) -> true` simplification when there are uncommitted appends, so the uncommitted NULL case should remain correct independently.

That said, #25277 looks like the broader fix for making `GetStatistics()`. If you think it would be safer to wait for that PR, I’m happy to do that.

lnkuiper:
Yes, let's wait for that. It's in the merge queue right now, so it should be merged some time today. Thanks for the patience.

EDIT: I will be offline, please ping someone else if this isn't merged and I don't respond.

src/planner/expression_nullability.cpp
JaySon-Huang, line 1: Hi @jaiswalvikas9991 
https://github.com/duckdb/duckdb/pull/25277 has been merged. Would you like to drop the changes of this files? (and you can use squash to cleanup the commit)

lnkuiper:
Now that [#25277](https://github.com/duckdb/duckdb/pull/25277) has merged, could you rebase and remove `TableHasUncommittedAppends` and its check in `src/planner/expression_nullability.cpp`? `DuckTableEntry::GetStatistics()` now includes transaction-local statistics, so this guard is redundant and prevents valid non-NULL proofs even when the appended rows contain no NULLs.

Once that's fixed, this can be merged. The CI failure is spurious

src/planner/expression_nullability.cpp
jaiswalvikas9991: Done. Thanks.

jaiswalvikas9991:
@lnkuiper Done. I rebased on top of #25277 and removed `TableHasUncommittedAppends` and the corresponding check from `src/planner/expression_nullability.cpp`.

lnkuiper:
Thanks, looks good to me now. However, there are a few extremely verbose comments (6+ lines), could you reduce these? It's not necessary to explain everything to that level. Then it's ready to go

jaiswalvikas9991:
@lnkuiper Done, I shortened the comments.

jaiswalvikas9991:
@lnkuiper I checked the failing `Interrupted QueryAppender flow: interrupt -> clear -> close finishes` test locally with `build/relassert/test/unittest`. It appears flaky: the same test sometimes completes successfully and sometimes hangs in my local relassert build.

```bash
./build/relassert/test/unittest 'Interrupted QueryAppender flow: interrupt -> clear -> close finishes'
Filters: Interrupted QueryAppender flow: interrupt -> clear -> close finishes
[1/1] (100%): Interrupted QueryAppender flow: interrupt -> clear -> close finishes took 0.232s
===============================================================================
All tests passed (3 assertions in 1 test case)
```

jaiswalvikas9991:
@Mytherin It seems like @lnkuiper is not available. Could you please have a look at my last comment about the CI failure? I think the failing test is unrelated/flaky, but I would appreciate your confirmation so this can be merged if everything else looks fine.

dentiny:
Yeah the failed CI does seem flaky, maybe you could merge with base branch to trigger (or admins could re-trigger only the failed pipelines)

jaiswalvikas9991:
@dentiny  Yes. If I merge/rebase again, all tests will run again, and if we are unlucky this test or another flaky test might fail again. It would be helpful if a maintainer can rerun this particular failed test/job.

jaiswalvikas9991:
@lnkuiper

The build failure happened because commit `cda46f5a5b7e15c340e166cbba732b9da02edbb3` added `ClientContext &context` as the first argument to the `ConstantOrNull` methods after the last time this PR was updated. I rebased to the latest changes and incorporated this new argument into the changes made in this PR.


**The change I did**
```bash
diff --git a/src/optimizer/constant_or_null_simplification.cpp b/src/optimizer/constant_or_null_simplification.cpp
index 9f1f1b85dc..7fd87aed7a 100644
--- a/src/optimizer/constant_or_null_simplification.cpp
+++ b/src/optimizer/constant_or_null_simplification.cpp
@@ -133,7 +133,7 @@ unique_ptr<Expression> ConstantOrNullSimplification::SimplifyExpression(LogicalO
                children.push_back(std::move(func_children[child_idx]));
        }

-       return ExpressionRewriter::ConstantOrNull(std::move(children), Value::BOOLEAN(!value.value()));
+       return ExpressionRewriter::ConstantOrNull(this->context, std::move(children), Value::BOOLEAN(!value.value()));
 }

 unique_ptr<LogicalOperator> ConstantOrNullSimplification::OptimizeFilter(unique_ptr<LogicalOperator> op,
diff --git a/src/optimizer/rule/in_clause_simplification_rule.cpp b/src/optimizer/rule/in_clause_simplification_rule.cpp
index 1e737f4976..d1a3c6e3b9 100644
--- a/src/optimizer/rule/in_clause_simplification_rule.cpp
+++ b/src/optimizer/rule/in_clause_simplification_rule.cpp
@@ -179,7 +179,7 @@ unique_ptr<Expression> InEnumSimplificationRule::Apply(LogicalOperator &op, vect
        //      If SOME of them are in the domain, swap out the children for the valid ENUM values.
        if (matched_enum_values.size() == EnumType::GetSize(enum_type)) {
                const bool in_clause = (expr.GetExpressionType() == ExpressionType::COMPARE_IN);
-               return rewriter.ConstantOrNull(in_children[0]->Copy(), Value::BOOLEAN(in_clause));
+               return rewriter.ConstantOrNull(GetContext(), in_children[0]->Copy(), Value::BOOLEAN(in_clause));
        } else if (in_children.size() > 1) {
                children.swap(in_children);
        } else {
```

PR #24707, opened 2026-08-11
Prune out of range fixed size array extracts via statistics callback

### LINKED ISSUE #24376: Fixed-size array out-of-range extract is not pruned
### What happens?

See repro for details.

### To Reproduce

```sql
memory D ATTACH '/tmp/array_oob_prune.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE OR REPLACE TABLE arrays AS
     SELECT [i::INTEGER, (i + 1)::INTEGER, (i + 2)::INTEGER]::INTEGER[3] AS a
     FROM range(6144) tbl(i);

-- row group not pruned
db D EXPLAIN ANALYZE
     SELECT count(*) FROM arrays WHERE a[4] IS NOT NULL;
╭─ Summary ───────────╮
│ Total Time: 0.0024s │
╰─────────────────────╯
│ Table: db.main.arrays                            │
│ array_extract(CAST(a AS INTEGER[]), 4) IS NOT    │
│ NULL                                             │
│ Row Groups Scanned: 3 / 3                        │
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dchung21:
Fixes: https://github.com/duckdb/duckdb/issues/24376

Previously, fixed sized array extracts that were out of range evaluate to null but the optimizer still performs a full table scan despite this.

- Prune out of range fixed sized array extracts via statistics callback `ListExtractStats`
- Added `ListExtractBindData` to contain the array size and index
- In `ListExtractBind` produce the data with the necessary information

### DIFF
--- a/src/function/scalar/list/list_extract.cpp
+++ b/src/function/scalar/list/list_extract.cpp
@@ -123,13 +123,36 @@ static void ListExtractFunction(DataChunk &args, ExpressionState &state, Vector 
 	}
 }
 
+struct ListExtractBindData : public FunctionData {
+	ListExtractBindData(optional_idx array_size, optional<Value> index)
+	    : array_size(array_size), index(std::move(index)) {
+	}
+	optional_idx array_size;
+	optional<Value> index;
+
+	unique_ptr<FunctionData> Copy() const override {
+		return make_uniq<ListExtractBindData>(array_size, index);
+	}
+	bool Equals(const FunctionData &other_p) const override {
+		auto &other = other_p.Cast<ListExtractBindData>();
+		return array_size == other.array_size && index == other.index;
+	}
+};
+
 static unique_ptr<FunctionData> ListExtractBind(BindScalarFunctionInput &input) {
 	auto &context = input.GetClientContext();
 	auto &bound_function = input.GetBoundFunction();
 	auto &arguments = input.GetArguments();
 	D_ASSERT(bound_function.GetArguments().size() == 2);
+
+	optional_idx array_size;
+	optional<Value> index;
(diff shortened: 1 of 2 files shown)


Maxxen:
Hello!

I think this would be better served through a `statistics` callback on the array-extract function - then you dont have to implement a new optimizer pass.

I recently did something similar here https://github.com/duckdb/duckdb/pull/24600/changes#diff-87e96ac3afa36123e9d6cd8ec3a3859eaf2cd15e7a3f1ad0e836eef2ee2b084fR205-R214

Although in this case we would probably inspect the bind data to get the index, and set the validity stats to `has_no_null: false` `has_null`: true` to signal "all are null" if the index is out of range

dchung21:
> Hello!
> 
> I think this would be better served through a `statistics` callback on the array-extract function - then you dont have to implement a new optimizer pass.
> 
> I recently did something similar here https://github.com/duckdb/duckdb/pull/24600/changes#diff-87e96ac3afa36123e9d6cd8ec3a3859eaf2cd15e7a3f1ad0e836eef2ee2b084fR205-R214
> 
> Although in this case we would probably inspect the bind data to get the index, and set the validity stats to `has_no_null: false` `has_null`: true` to signal "all are null" if the index is out of range

Thanks for the feedback @Maxxen, I've updated the PR to follow the suggested approach and updated the title and description as well

PR #24709, opened 2026-08-11
Unpack row comparisons into per-field IS NOT DISTINCT FROM in any context

### LINKED ISSUE #23594: Tuple doesn't leverage pruning
### What happens?

See repro example for details, should be straightforward.

### To Reproduce

```sql
memory D CREATE TABLE facts AS
SELECT r::BIGINT id, (r % 10)::INT b,
(DATE '2020-01-01' + INTERVAL (r // 30000) DAY)::DATE d,
'payload-'||(r%100000) s
FROM range(50_000_000) tbl(r) ORDER BY r;

-- Full table sequential scan when query with tuple
memory D EXPLAIN ANALYZE SELECT count(*) FROM facts WHERE (id, b) = (42, 2);
╭─ Summary ───────────╮
│ Total Time: 0.0327s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ──────────────╮
│ 1 row 0µs │
╰──────────────────┬─────────────────╯
╭─ Filter ─────────┴─────────────────╮
│ Expression: "row"(id, b) = (42, 2) │
│ 1 row 410.0ms │
╰──────────────────┒┎────────────────╯
╭─ Table Scan ─────┚┖────────────────╮
│ Table: memory.main.facts │
│ Type: Sequential Scan │
│ Projections: id, b │
│ Row Groups Scanned: 407 / 407 │
│ 50,000,000 rows 60.0ms │

-- When query with each column, row groups are pruned correctly
memory D SELECT count() FROM facts WHERE id = 42 AND b = 2;
┌──────────────┐
│ count_star() │
│ int64 │
├──────────────┤
│ 1 │
└──────────────┘
memory D EXPLAIN ANALYZE SELECT count() FROM facts WHERE id = 42 AND b = 2;
╭─ Summary ───────────╮
│ Total Time: 0.0012s │
╰─────────────────────╯
│ Aggregates: count_star() │
│ 1 row 0µs │
╰──────────────┒┎─────────────╯
╭─ Table Scan ─┚┖─────────────╮
│ Table: memory.main.facts │
│ Type: Sequential Scan │
│ Filters: id = 42, b = 2 │
│ Row Groups Scanned: 1 / 407 │
│ 1 row 0µs │
```

### OS:

macos

### DuckDB Version:

a81547b43c56c671a7ad1a636f83886713efdd8a

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

gulugulubing:
Fixes #23594

`RowComparisonSimplificationRule` now unpacks row comparisons uniformly: either side may be a row constructor (possibly wrapped in casts) or a folded TUPLE/STRUCT constant, the children may be any expression, and the unpacking happens in any context, not just the top of a filter. `(id, b) = (42, 2)` becomes `id IS NOT DISTINCT FROM 42 AND b IS NOT DISTINCT FROM 2`, so the existing pushdown and zone map pruning applies to it.

- row comparisons compare their fields with IS NOT DISTINCT FROM, so the result is the same everywhere (a NULL field does not turn the comparison into NULL: `(1, NULL) = (1, NULL)` is true) - projections and join conditions included
- fields are compared in the aligned field type, so `(id, b) = (42, 1.5)` unpacks to `id IS NOT DISTINCT FROM 42 AND b IS NOT DISTINCT FROM 2`
- `ROW() = ROW()` unpacks to TRUE, and a NULL constant is left alone: NULL is not a row of NULL fields
- a comparison with a volatile or can-throw child is left alone as well (a field that fails skips the evaluation of the rest)
- tests: `test/sql/filter/test_tuple_comparison_pushdown.test` (pushdown, row group pruning, bind time type alignment, nested rows, a user-written IS NOT DISTINCT FROM row comparison; skipped in the disable_optimizer config since pruning requires the optimizer) and `test/sql/optimizer/expression/test_row_comparison_simplification.test` (flat and nested NULL semantics, NULL constant, empty rows, volatile children)

Note: the filter combiner does not accept IS NOT DISTINCT FROM yet, so the per-field comparisons are pushed as single-column expression filters; #24605 handles folding them to equality, which is also what join conditions need.

### DIFF
--- a/src/include/duckdb/optimizer/rule/comparison_simplification.hpp
+++ b/src/include/duckdb/optimizer/rule/comparison_simplification.hpp
@@ -21,7 +21,7 @@ public:
 	                             bool is_root) override;
 };
 
-//! Rewrites top-level filter equality between row constructors into scalar equalities.
+//! Rewrites top-level filter equality between row constructors (or folded row constants) into scalar equalities.
 class RowComparisonSimplificationRule : public Rule {
 public:
 	explicit RowComparisonSimplificationRule(ExpressionRewriter &rewriter);
--- a/src/optimizer/rule/comparison_simplification.cpp
+++ b/src/optimizer/rule/comparison_simplification.cpp
@@ -2,12 +2,16 @@
 #include "duckdb/optimizer/rule/comparison_simplification.hpp"
 
 #include "duckdb/common/helper.hpp"
+#include "duckdb/common/types.hpp"
 #include "duckdb/common/types/timestamp.hpp"
+#include "duckdb/common/types/value.hpp"
 #include "duckdb/execution/expression_executor.hpp"
+#include "duckdb/planner/expression/bound_cast_expression.hpp"
 #include "duckdb/planner/expression/bound_constant_expression.hpp"
 #include "duckdb/optimizer/expression_rewriter.hpp"
 #include "duckdb/planner/expression/bound_comparison_expression.hpp"
 #include "duckdb/planner/expression/bound_conjunction_expression.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
 #include "duckdb/planner/expression/bound_operator_expression.hpp"
 
 namespace duckdb {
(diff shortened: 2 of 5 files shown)


Mytherin:
Thanks - but this seems to be targeting a far too specific rewrite. Can we instead just "unpack" the comparison, i.e.:

```sql
(a, b) = (c, d)
```

Is equivalent to:

```
a = c AND b = d
```

 This is both simpler, and more composable / flexible.

gulugulubing:
Thanks for the review! I've reworked the PR:

RowComparisonSimplificationRule now handles both sides uniformly — either side can be a row constructor (possibly wrapped in casts) or a folded TUPLE/STRUCT constant — and the comparison is unpacked into per-field scalar comparisons:

- Row-vs-row unpacks with IS NOT DISTINCT FROM, matching the row equality semantics
- Row-vs-constant with a NULL-free constant unpacks with =, which the filter combiner absorbs and pushes down as zone map filters
- Constants containing NULLs unpack with IS NOT DISTINCT FROM (e.g. (a, b) = (NULL, 2) matches rows with a IS NULL), which = cannot express

The CI failures appear to be network-related flakes.

gulugulubing:
Updated the description to match the rework, and merged the base branch so the conflict is gone.

Mytherin:
Thanks, this is closer, but it's still quite special-cased. Could we unpack `row = row / row = constant` into `IS NOT DISTINCT FROM` per field in any context and for any child expressions? Then add a separate rule that rewrites `x IS NOT DISTINCT FROM <non-NULL constant>` to `x = <constant>` at the root of a filter, so pushdown results from that?

gulugulubing:
Thanks - I reworked it along those lines, and updated the description with the details.

One thing I ran into: I first tried the separate rule that rewrites `x IS NOT DISTINCT FROM <non-NULL constant>`
to `x = <constant>` at the root of a filter. For `AND(x IS NOT DISTINCT FROM 1, y IS NOT DISTINCT FROM 2)` the
conjuncts are not really "at the root" - the filter splits the AND, puts the first conjunct back where the AND
used to be and never looks at it again, so a root-only rule only ends up rewriting some of them. Matching the AND
as well would work, but it felt less clean.

Then I remembered I already have #24605 open, which teaches the filter combiner to accept IS NOT DISTINCT FROM
(a non-NULL constant folds to equality, NDF pairs are tracked so NULL semantics do not change). That covers the
second half, so this PR can stay a plain unpacking rule - and on its own it still fixes the issue.

I think this split might be a bit cleaner.

PR #24716, opened 2026-08-12
Preserve case-sensitive VARIANT fields during shredding

### LINKED ISSUE #24297: Forced VARIANT shredding drops case-distinct keys
### What happens?

See repro for details.

### To Reproduce

```sql
memory D SET variant_minimum_shredding_size = 0;
memory D SET force_variant_shredding = 'STRUCT(a INTEGER)';
memory D 
memory D CREATE TABLE t(v VARIANT);
memory D INSERT INTO t VALUES ({'A': 2}::VARIANT), ({'a': 1}::VARIANT);
memory D 
memory D SELECT v::VARCHAR, variant_keys(v), v.a, v.A FROM t ORDER BY v::VARCHAR;
┌────────────────────┬─────────────────┬─────────┬─────────┐
│ CAST(v AS VARCHAR) │ variant_keys(v) │    a    │    A    │
│      varchar       │    varchar[]    │ variant │ variant │
├────────────────────┼─────────────────┼─────────┼─────────┤
│ {'A': 2}           │ [A]             │ NULL    │ 2       │
│ {'a': 1}           │ [a]             │ 1       │ NULL    │
└────────────────────┴─────────────────┴─────────┴─────────┘
memory D CHECKPOINT;
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D SELECT v::VARCHAR, vector_type(v), variant_keys(v), v.a, v.A FROM t ORDER BY v::VARCHAR;
┌────────────────────┬─────────────────┬─────────────────┬─────────┬─────────┐
│ CAST(v AS VARCHAR) │ vector_type(v)  │ variant_keys(v) │    a    │    A    │
│      varchar       │     varchar     │    varchar[]    │ variant │ variant │
├────────────────────┼─────────────────┼─────────────────┼─────────┼─────────┤
│ {'a': 1}           │ SHREDDED_VECTOR │ [a]             │ 1       │ NULL    │
│ {}                 │ SHREDDED_VECTOR │ []              │ NULL    │ NULL    │
└────────────────────┴─────────────────┴─────────────────┴─────────┴─────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

### LINKED ISSUE #24379: variant_extract leads to incorrect result
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE src AS SELECT '{"foo": 2}'::JSON::VARIANT AS v;
memory D COPY src TO '/private/tmp/explicit_case.parquet'
           (SHREDDING {'v': 'STRUCT("Foo" INTEGER)'});
memory D 
memory D SELECT v::VARCHAR, variant_extract(v, 'foo')::VARCHAR
         FROM '/private/tmp/explicit_case.parquet';
┌────────────────────┬────────────────────────────────────────────┐
│ CAST(v AS VARCHAR) │ CAST(variant_extract(v, 'foo') AS VARCHAR) │
│      varchar       │                  varchar                   │
├────────────────────┼────────────────────────────────────────────┤
│ {}                 │ NULL                                       │
└────────────────────┴────────────────────────────────────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

matthiasgoergens:
Fixes #24379. Fixes #24297.

VARIANT object keys are case-sensitive, but DuckDB STRUCT field identifiers are case-insensitive. When a shredding schema contains a key spelled `Foo` and the actual data uses `foo`, the untyped-overlay omission set was case-insensitive and silently discarded the value. The separate issue of schemas that themselves contain both `a` and `A` is handled by #24553; this PR fixes the remaining case where the schema and data differ in spelling alone.

The Parquet and internal-storage writers now use exact matching for the omission set, so a field spelled differently from the shredded schema is retained in the overlay rather than dropped. The read-side lookup in `FindShreddedColumnInternal` also switches from the STRUCT field-name equality operator (case-insensitive) to `GetIdentifierName()` for exact comparison.

The regression covers top-level and nested mismatches in both Parquet and internal storage, forced shredding, and an isolated pushdown through the read path.

### DIFF
--- a/extension/parquet/writer/variant/convert_variant.cpp
+++ b/extension/parquet/writer/variant/convert_variant.cpp
@@ -230,14 +230,13 @@ vector<idx_t> GetChildIndices(const UnifiedVariantVectorData &variant, idx_t row
 		}
 		return child_indices;
 	}
-	//! FIXME: The variant spec says that field names should be case-sensitive, not insensitive
-	case_insensitive_string_set_t shredded_fields = shredding_state->ObjectFields();
+	auto shredded_fields = shredding_state->ObjectFields();
 
 	for (idx_t i = 0; i < nested_data.child_count; i++) {
 		auto keys_index = variant.GetKeysIndex(row, i + nested_data.children_idx);
 		auto &key = variant.GetKey(row, keys_index);
 
-		if (shredded_fields.count(key)) {
+		if (shredded_fields.count(key.GetString())) {
 			//! This field is shredded on, omit it from the value
 			continue;
 		}
--- a/src/function/variant/variant_shredding.cpp
+++ b/src/function/variant/variant_shredding.cpp
@@ -356,13 +356,12 @@ void VariantShreddingState::SetShredded(uint32_t row, uint32_t values_index, uin
 	count++;
 }
 
-case_insensitive_string_set_t VariantShreddingState::ObjectFields() {
+unordered_set<string> VariantShreddingState::ObjectFields() const {
 	D_ASSERT(type.id() == LogicalTypeId::STRUCT);
-	case_insensitive_string_set_t res;
+	unordered_set<string> res;
 	auto &child_types = StructType::GetChildTypes(type);
 	for (auto &entry : child_types) {
-		auto &type = entry.first;
-		res.emplace(type.c_str(), static_cast<uint32_t>(type.size()));
+		res.emplace(entry.first.GetIdentifierName());
 	}
 	return res;
 }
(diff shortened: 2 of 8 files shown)


Mytherin:
Thanks!

PR #24669, opened 2026-08-10
Handle infinite dates in timestamp comparison optimization

jaiswalvikas9991:
Fixes #24306

The timestamp-to-date equality rewrite converts date constants into timestamp bounds. For DATE infinity and -infinity, the finite date-to-timestamp construction using Timestamp::FromDatetime can overflow, which is correctly detected and results in an exception.

This change rewrites infinite DATE comparisons to direct TIMESTAMP equality, and adds regression coverage for both operand orders.

Tested with:

```bash
  ./build/debug/test/unittest test/optimizer/pushdown/timestamp_to_date_infinity_pushdown.test
```

**Run after the fix**

```bash
memory D CREATE TABLE ts_inf(t TIMESTAMP);
memory D INSERT INTO ts_inf VALUES (TIMESTAMP '-infinity'), (TIMESTAMP '2025-01-01'), (TIMESTAMP 'infinity');
memory D SELECT count(*) FROM ts_inf WHERE CAST(t AS DATE) = DATE 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
memory D PRAGMA disable_optimizer;
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D SELECT count(*) FROM ts_inf WHERE CAST(t AS DATE) = DATE 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘

```

PR #24719, opened 2026-08-12
Fix: Don't rewrite timestamp-to-date comparisons for infinite dates

### LINKED ISSUE #24306: Disable optimizers leads to different results
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE ts_inf(t TIMESTAMP);
memory D INSERT INTO ts_inf VALUES (TIMESTAMP '-infinity'), (TIMESTAMP '2025-01-01'), (TIMESTAMP 'infinity');
memory D SELECT count(*) FROM ts_inf WHERE CAST(t AS DATE) = DATE 'infinity';
Conversion Error:
Date and time not in timestamp range

┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D SELECT count(*) FROM ts_inf WHERE CAST(t AS DATE) = DATE 'infinity';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

DinosL:
Bail out and let the comparison be evaluated as-is instead.

Fixes: https://github.com/duckdb/duckdb/issues/24306

### DIFF
--- a/src/optimizer/rule/timestamp_comparison.cpp
+++ b/src/optimizer/rule/timestamp_comparison.cpp
@@ -81,6 +81,9 @@ unique_ptr<Expression> TimeStampComparison::Apply(LogicalOperator &op, vector<re
 		auto original_val = result.GetValue<duckdb::date_t>();
 		auto no_seconds = dtime_t(0);
 
+		if (!original_val.IsFinite()) {
+			return nullptr;
+		}
 		// original date as timestamp with no seconds
 		auto original_val_ts = Value::TIMESTAMP(original_val, no_seconds);
 		auto original_val_for_comparison = make_uniq<BoundConstantExpression>(original_val_ts);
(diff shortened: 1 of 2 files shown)


jaiswalvikas9991:
@DinosL I had also been looking into this issue and opened a PR with a slightly different approach: https://github.com/duckdb/duckdb/pull/24669

From testing, this PR avoids the overflow by bailing out for non-finite DATE values, so the plan keeps the original CAST(t AS DATE) predicate. My approach handles infinite DATE values explicitly and rewrites equality comparisons to direct TIMESTAMP infinity comparisons.




```sql
  EXPLAIN SELECT count(*)
  FROM ts_inf
  WHERE CAST(t AS DATE) = DATE 'infinity';
```

With this PR, I see a plan(for above query) that keeps the cast predicate: optional: (t >= ...::TIMESTAMP) AND (CAST(t AS DATE) = 'infinity'::DATE)


With my branch, this becomes:
Filters: t = 'infinity'::TIMESTAMP

So I think both approaches avoid the crash, but my version preserves the optimization for infinite dates. I’m happy to adapt my PR/tests based on your preferred direction. Thanks.

PR #24721, opened 2026-08-12
Validate encryption key before parquet metadata cache leveraged

### LINKED ISSUE #24317: Read parquet without encryption key passes through

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24317

In the current implementation, when users access parquet metadata, when it's cached it will be directly returned to user, even if the user doesn't provide encryption key or encryption key is not correct.

In this PR, I tried to store the key hash in the metadata cache, and validate before using metadata cache.


Mytherin:
Thanks!

PR #24722, opened 2026-08-12
Skip segments for update stats

dentiny:
Hi team, I found with rows updated in the row group, we prune based on row group instead of segment stats, which leads to unnecessary IO and compute for segments.

For example,
```sql
memory D CREATE TABLE events AS SELECT i, TIMESTAMP '2020-01-01' + i * INTERVAL 1 SECOND AS ts FROM range(100000) t(i);
memory D UPDATE events SET ts = TIMESTAMP '2100-01-01' WHERE i = 1;
memory D PRAGMA enable_profiling='json';
memory D SET tracked_metrics=['query.total_rows_scanned'];
memory D SELECT count(*) FROM events WHERE ts >= TIMESTAMP '2099-01-01';
{
    "query": {
        "total_rows_scanned": 100000
    }
}
```
which clearly shows all rows in the table gets scanned, but in reality only the last segment needs to be scanned.

So in the PR, in the segment scan function `ColumnData::CheckZonemap`, I reduce the update stats to segment level, instead of row group level. Updated result
```sql
memory D CREATE TABLE events AS SELECT i, TIMESTAMP '2020-01-01' + i * INTERVAL 1 SECOND AS ts FROM range(100000) t(i);
memory D UPDATE events SET ts = TIMESTAMP '2100-01-01' WHERE i = 1;
memory D PRAGMA enable_profiling='json';
memory D SET tracked_metrics=['query.total_rows_scanned'];
memory D SELECT count(*) FROM events WHERE ts >= TIMESTAMP '2099-01-01';
{
    "query": {
        "total_rows_scanned": 16032
    }
}
```


repro_art_delta.dSYM/Contents/Info.plist
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Mytherin, line 1: These files seem to have been accidentally checked in

Mytherin:
Thanks but I think this is not very effective as an optimization in practice. Column segments almost always span the entire row group aside from the first row group, and `UpdateSegment::HasUpdates` now incurs an extra cost for every zone map check because it has to traverse update segments to figure out if the stats apply. I don't think this is worth the extra cost / complexity to add to the system.

PR #24727, opened 2026-08-12
Fix NULL dynamic ROWS window frame boundaries

### LINKED ISSUE #24632: Dynamic ROWS NULL PRECEDING bound can reuse a previous frame start
### What happens?

A dynamic `ROWS NULL PRECEDING` bound can intermittently reuse the previous row's frame start. The second row should count only itself, but the default window path sometimes counts both rows.

#### Code

```sql
WITH src AS MATERIALIZED (
  SELECT * FROM (VALUES (1, 0), (2, NULL)) AS v(id, off)
)
SELECT
  id,
  off,
  COUNT(*) OVER (
    ORDER BY id
    ROWS BETWEEN off PRECEDING AND CURRENT ROW
  ) AS actual,
  (
    SELECT COUNT(*)
    FROM src AS i
    WHERE i.id BETWEEN src.id - COALESCE(src.off, 0) AND src.id
  ) AS expected
FROM src
ORDER BY id;
```

Run:

```sh
```

#### Expected

| id | off | expected |
| ---: | ---: | ---: |
| 1 | 0 | 1 |
| 2 | NULL | 1 |

#### Actual

Three fresh-process runs on the latest main produced:

| run | id | off | actual | expected |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 1 | 0 | 1 | 1 |
| 1 | 2 | NULL | 1 | 1 |
| 2 | 1 | 0 | 1 | 1 |
| 2 | 2 | NULL | 2 | 1 |
| 3 | 1 | 0 | 1 | 1 |
| 3 | 2 | NULL | 1 | 1 |

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (`e500d77864fe`)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

idle-lab:
Fixes: #24632

Dynamic `ROWS` frame boundaries previously read the underlying boundary value without first checking its validity. For a `NULL` boundary, the underlying value is unspecified and could produce an incorrect frame boundary.

Treat `NULL` dynamic `ROWS` boundaries as a zero offset in all four affected paths:

- frame start with `PRECEDING`
- frame start with `FOLLOWING`
- frame end with `PRECEDING`
- frame end with `FOLLOWING`

Regression tests cover all four cases and execute the query repeatedly to catch the previously intermittent behavior.

### DIFF
--- a/src/function/window/window_boundaries_state.cpp
+++ b/src/function/window/window_boundaries_state.cpp
@@ -713,9 +713,11 @@ void WindowBoundariesState::FrameBegin(idx_t row_idx, const idx_t count, WindowI
 	case WindowBoundary::EXPR_PRECEDING_ROWS:
 		for (idx_t chunk_idx = 0; chunk_idx < count; ++chunk_idx, ++row_idx) {
 			int64_t computed_start;
-			int64_t boundary_value =
-			    boundary_begin.CellIsNull(chunk_idx) ? 0 : boundary_begin.GetCell<int64_t>(chunk_idx);
-			if (!TrySubtractOperator::Operation(static_cast<int64_t>(row_idx), boundary_value, computed_start)) {
+			if (boundary_begin.CellIsNull(chunk_idx)) {
+				throw InvalidInputException("Window ROWS PRECEDING expression cannot be NULL");
+			}
+			if (!TrySubtractOperator::Operation(static_cast<int64_t>(row_idx),
+			                                    boundary_begin.GetCell<int64_t>(chunk_idx), computed_start)) {
 				window_start = partition_begin_data[chunk_idx];
 			} else {
 				window_start = UnsafeNumericCast<idx_t>(MaxValue<int64_t>(computed_start, 0));
@@ -726,9 +728,11 @@ void WindowBoundariesState::FrameBegin(idx_t row_idx, const idx_t count, WindowI
 	case WindowBoundary::EXPR_FOLLOWING_ROWS:
 		for (idx_t chunk_idx = 0; chunk_idx < count; ++chunk_idx, ++row_idx) {
 			int64_t computed_start;
-			int64_t boundary_value =
-			    boundary_begin.CellIsNull(chunk_idx) ? 0 : boundary_begin.GetCell<int64_t>(chunk_idx);
-			if (!TryAddOperator::Operation(static_cast<int64_t>(row_idx), boundary_value, computed_start)) {
+			if (boundary_begin.CellIsNull(chunk_idx)) {
+				throw InvalidInputException("Window ROWS FOLLOWING expression cannot be NULL");
+			}
+			if (!TryAddOperator::Operation(static_cast<int64_t>(row_idx),
+			                               boundary_begin.GetCell<int64_t>(chunk_idx), computed_start)) {
 				window_start = partition_begin_data[chunk_idx];
 			} else {
 				window_start = UnsafeNumericCast<idx_t>(MaxValue<int64_t>(computed_start, 0));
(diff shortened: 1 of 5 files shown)


hawkfish:
Hi @idle-lab - Thanks for the PR! Unfortunately, this is not desired behaviour. `NULL` means "unknown" not zero.

The SQL standard doesn't even allow variable bounds and running this query on Postgres produces an error:

```sql
ERROR:  argument of ROWS must not contain variables
LINE 6:     ROWS BETWEEN off PRECEDING AND CURRENT ROW
```

We have extended the standard to allow variables, but `NULL` should not be allowed. PG gives an error here:

```sql
ERROR:  frame ending offset must not be null
```

Further details in the review.

src/function/window/window_boundaries_state.cpp
```diff
@@ -713,8 +713,9 @@ void WindowBoundariesState::FrameBegin(idx_t row_idx, const idx_t count, WindowI
 	case WindowBoundary::EXPR_PRECEDING_ROWS:
 		for (idx_t chunk_idx = 0; chunk_idx < count; ++chunk_idx, ++row_idx) {
 			int64_t computed_start;
-			if (!TrySubtractOperator::Operation(static_cast<int64_t>(row_idx),
-			                                    boundary_begin.GetCell<int64_t>(chunk_idx), computed_start)) {
+			int64_t boundary_value =
```
hawkfish: Throw on NULL here and below. If you want to be more informative, you can use `EnumUtils.hpp` to convert the `WindowBoundary` to a readable string for the message.

test/sql/window/test_boundary_expr.test
```diff
@@ -143,3 +143,39 @@ ORDER BY id;
 ----
 1	2	2	2
 2	2	2	2
+
+# Issue #24632: NULL dynamic ROWS bounds are equivalent to zero
+loop i 0 50
```
hawkfish: This can be removed.

test/sql/window/test_boundary_expr.test
```diff
@@ -143,3 +143,39 @@ ORDER BY id;
...
+
+query IIIIII
+WITH src AS MATERIALIZED (
+  SELECT * FROM (VALUES (1, 1), (2, NULL), (3, 1), (4, NULL)) AS v(id, off)
+)
+SELECT
+  id,
+  off,
+  COUNT(*) OVER (
```
hawkfish: Break these up into separate queries so we can check each case independently.

hawkfish (changes requested):

hawkfish (approved):
Thanks - LGTM!

idle-lab:
@hawkfish Thanks for your review!

idle-lab:
@hawkfish Existing tests show that nullable dynamic ROWS offsets previously happened to behave like zero, although this was not guaranteed because validity was not checked. Rejecting `NULL` therefore changes observable behavior and may affect compatibility. Is this acceptable?

hawkfish:
> @hawkfish Existing tests show that nullable dynamic ROWS offsets previously happened to behave like zero, although this was not guaranteed because validity was not checked. Rejecting `NULL` therefore changes observable behavior and may affect compatibility. Is this acceptable?

It looks like the tests that are failing are fuzzer tests that use the `all_types` table, which is mostly `NULL`s? But the original observation that the value is random noise is still valid, so I'd say the tests need to be fixed. The frame bounds for those should probably be wrapped with `COALESCE(col, 0)` now.

PR #24728, opened 2026-08-12
Extensible keywords

Dtenwolde:
First step towards runtime extensible parsers is making the set of keywords be extensible. I now separated out `KeywordExtension` that handles the keywords and which is part of the `ExtensionCallbackManager`. I've kept it separate from `ParserExtensions` for now, though perhaps this is not the right shape, happy to hear feedback. 

The keywords are all validated based on a few criteria: 
- More than one character
- No digits
- Unquoted
- May belong to multiple categories that are compatible. Trying to register in both `reserved` and `unreserved` is not compatible

Reregistering the same keyword in the same category is idempotent.


test/extension/loadable_extension_demo.cpp
```diff
@@ -1259,6 +1259,14 @@ DUCKDB_CPP_EXTENSION_ENTRY(loadable_extension_demo, loader) {
 	}
 
 	// add a parser extension
+	loader.RegisterKeyword("quack_reserved_keyword", ExtensionKeywordCategory::RESERVED);
```
Tishj, line 1262: Should we change these to `RegisterKeywords` and `UnregisterKeywords` (or `RemoveKeywords`), taking a `set<string>, ExtensionKeywordCategory` ?

Or do you want to leave "unregistering" for a follow-up, since that is probably not trivial (could break existing rules / make them unreachable)

Tishj:
Since you mention `ParserExtension`, that got me thinking. Perhaps it's better to make it explicit that these are extensions to the **PEG** parser, these won't have any effect on the old postgres parser.

Dtenwolde:
I think we wanted to emphasize less on the *PEG*ness of the parser, not using it in new class names etc. Just that it's the current parser

test/extension/loadable_extension_demo.cpp
Dtenwolde: I was going to create an internal issue, but after consideration I think unregistering keywords (or allowing their categories to be changed) is too risky and can have unintended consequences. Perhaps it would be useful for `DialectExtensions`, but those would anyway have to register their own set of keywords to work. So that would probably also be a slightly different interface anyway.

Tishj:
I just noticed this in the existing code btw:
```c++
	// keyword overrides
	AddKeywordOverride("TABLE", 1, ' ');
	AddKeywordOverride(".", 0, '\0');
	AddKeywordOverride("(", 0, '\0');
```

Isn't this already a form of keyword extension?

Tishj:
With https://github.com/duckdb/duckdb/pull/24867 this should be easier to do.
If we expose a way to modify the `ParsedGrammar` then the keyword helper follows for free.

Dtenwolde:
Superseded by: https://github.com/duckdb/duckdb/pull/24867
Closing this one

PR #24735, opened 2026-08-12
Fix duckbox rendering of case-sensitive ENUM types

### LINKED ISSUE #24575: DuckDB Cli: duckbox lowercase column type rendering obscures actual type of ENUM-typed columns
### What happens?

The cli output mode duckbox prints the column data type below the column name in lowercase.

For most datatypes, that's ok in the sense that the type as it is printed still clearly identifies the actual type (because DuckDB is typically case insensitive). But, the ENUM type is an exception - `ENUM('a','A') and `ENUM('A','a')` are different types. In these cases, the duckbox output obscures the actual type, which is a (minor) inconvenience.

(note that this issue would also affect any composite type that depends on an ENUM type.)

### To Reproduce

```sql
select 'A'::ENUM('A','a') upper_a
        ,      'A'::ENUM('a','A') lower_a
;
```
output:
```
┌────────────────┬────────────────┐
│    upper_a     │    lower_a     │
│ enum('a', 'a') │ enum('a', 'a') │
├────────────────┼────────────────┤
│ A              │ A              │
└────────────────┴────────────────┘
```

### OS:

win 11

### DuckDB Version:

1.5.5

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Roland Bouman

### Affiliation:

EPAM Systems B.V.

- [x] Yes, I have

- [x] Yes, I have

Yes

mattyy-k:
Fixes #24575

## Problem

The DuckDB CLI's `duckbox` renderer lowercases the complete string returned by `LogicalType::ToString()` when rendering column types.

This loses information for case-sensitive ENUM labels. For example:

```sql
SELECT
    'A'::ENUM('A', 'a') AS upper_a,
    'A'::ENUM('a', 'A') AS lower_a;
```

The two columns have distinct ENUM types, but `duckbox` currently renders both as `enum('a', 'a')`.

`LogicalType::ToString()` already preserves the correct ENUM definition. The information is lost in `BoxRendererImplementation::RenderType()`, where the fallback path applies `StringUtil::Lower()` to the entire type string.

## Fix

Make the type-string lowercasing quote-aware. Type syntax outside single-quoted SQL string literals continues to be lowercased as before, while the contents of quoted literals retain their original case:
```
ENUM('A', 'a') -> enum('A', 'a')
ENUM('a', 'A') -> enum('a', 'A')
```
Escaped single quotes are handled as part of the quoted literal.

This keeps the fix in the CLI presentation layer rather than changing `LogicalType` or global type serialization. The same rendering path also means composite types containing ENUMs benefit naturally from the fix.

## Testing

Added a regression test covering the original issue:

```sql
SELECT
    'A'::ENUM('A', 'a') AS upper_a,
    'A'::ENUM('a', 'A') AS lower_a
LIMIT 0;
```

The test verifies that:

- `enum('A', 'a')` is preserved
- `enum('a', 'A')` is preserved
- the incorrect `enum('a', 'a')` representation is absent

Existing `duckbox` tests also pass locally. A fresh debug build is currently being completed; the newly added regression test will be rerun against the rebuilt binary once available.


mattyy-k:
One CI note: the current Linux Release failure appears unrelated to this change. The job fails during compilation in the external `bwc-test-utils` extension due to `TestUtilsExtension::Load(DuckDB&)` no longer matching the `Extension` API. The PR's formatting/tidy/relassert and other platform/sanitizer checks are passing. I haven't modified anything to work around this unrelated failure.

Mytherin:
Thanks!

PR #24743, opened 2026-08-13
Add utf8_validation (strict, ignore, replace) to Parquet reader

xproblematique:
## Add `utf8_validation='strict'|'replace'|'ignore'` support to Parquet reader

Closes https://github.com/duckdblabs/duckdb-internal/issues/9034

### Summary

Adds a `utf8_validation` named parameter to `read_parquet` / `parquet_scan`, mirroring the `on_error` parameter already supported by `decode()` (`extension/core_functions/scalar/blob/encode.cpp`). Controls how invalid UTF-8 in string columns is handled:

- `'strict'` (default) — throw on invalid UTF-8. This is today's only behavior, unchanged.
- `'replace'` — invalid bytes are replaced with `?`.
- `'ignore'` — invalid bytes are dropped from the string.

### Scope

This PR makes `utf8_validation` configurable for the four encodings: `PLAIN`, `DELTA_LENGTH_BYTE_ARRAY`, `DELTA_BYTE_ARRAY` and dictionary-encoded strings. 

### Included fix: `Utf8Proc::RemoveInvalid`

`third_party/utf8proc/utf8proc_wrapper.cpp`'s `Utf8Proc::RemoveInvalid` never checked the return value of `UTF8ExtraByteLoop`, so a multi-byte sequence truncated at the end of the input was kept in the output instead of dropped, producing invalid UTF-8 despite the function's contract. This also affects `decode(blob, 'ignore')`, independent of Parquet. Fixed here since `'ignore'` mode depends on it directly.

### Implementation notes

- `ParquetOptions` now has a `utf8_validation_option` field (default `STRICT_UTF8`), parsed via `ParquetMultiFileInfo::ParseOption`, following the same pattern as `prefetch_strategy`.
- `StringColumnReader::VerifyString` now returns a `string_t` instead of being throw-only, so callers can substitute a replaced/shortened string instead of the original bytes. For `replace` / `ignore`, it allocates new backing storage via `StringVector::EmptyString` on the vector currently being populated (via the`current_plain_result` pointer, since the`PlainRead` template interface doesn't expose the output `Vector`).
- `DeltaLengthByteArrayDecoder` has a fast path that validates an entire batch of concatenated strings in one pass (`VALIDATE_INDIVIDUAL_STRINGS = false`), previously used whenever `strict` mode only needs a yes/no answer. This doesn't work for `replace` / `ignore`, which need to fix individual strings, so `DeltaLengthByteArrayDecoder::Read()` now also forces the per-string path whenever `utf8_validation` isn't `strict` (previously this path was only forced for JSON columns).

### Known behavior note (pre-existing, unchanged by this PR)

In `strict` mode, `DELTA_LENGTH_BYTE_ARRAY`'s batch-validation error message reports the entire concatenated batch of strings rather than the single offending value (e.g. `"TREL\xC3hello world"` rather than `"TREL\xC3"`), since the batch check doesn't track individual string boundaries. This is existing behavior, not introduced here.

### Testing

SQLLogicTest coverage added (`test/parquet/utf8_validation_plain_dba_dlba_dict.test`).


extension/parquet/reader/string_column_reader.cpp
```diff
@@ -44,28 +44,63 @@ bool StringColumnReader::IsValid(const char *str_data, uint32_t str_len, const b
...
+		return VerifyString(str_data, str_len, true);
 	case StringColumnType::JSON: {
 		const auto error = StringUtil::ValidateJSON(str_data, str_len);
 		if (!error.empty()) {
 			throw InvalidInputException("Invalid JSON found in Parquet file: %s", error);
 		}
-		break;
+		return string_t(str_data, str_len);
 	}
 	default:
```
kryonix, line 104: `StringColumnReader` also reads `LogicalTypeId::BLOB`. Changing from a no-op to an `InternalException` breaks existing binary Parquet scans:
```sql
SELECT * FROM parquet_scan(
    'data/parquet-testing/binary_string.parquet',
    binary_as_string=false
);
```
Expected: `foo`, `bar`, `baz`. Actual: `INTERNAL Error: Unimplemented StringColumnType`. The existing `test/sql/copy/parquet/parquet_blob_string.test` fails at line 12. The branch must continue returning the unvalidated raw `string_t`.

extension/parquet/reader/string_column_reader.cpp
```diff
@@ -44,28 +44,63 @@ bool StringColumnReader::IsValid(const char *str_data, uint32_t str_len, const b
...
+		Utf8Proc::MakeValid(output, str_len);
+		target.Finalize();
+		return target;
+	}
+
+	case StringColumnReader::Utf8ValidationOption::IGNORE: {
+		if (!current_plain_result) {
+			throw InternalException("VerifyString: IGNORE mode requires current_plain_result to be set");
+		}
+		auto new_str = Utf8Proc::RemoveInvalid(str_data, str_len);
```
kryonix, line 78: `Utf8Proc::RemoveInvalid` retains truncated multibyte sequences, so the new `ignore` mode does not produce valid UTF-8. Both supplied fixtures reproduce this:
```sql
SELECT * FROM read_parquet(
    'data/parquet-testing/invalid.parquet',
    utf8_validation='ignore'
);
```
Expected: `TREL`. Actual: `Invalid Input Error: truncated UTF-8 sequence detected in Utf8Proc::UTF8ToCodepoint`.
`invalid_utf8_dlba.parquet` fails identically. This helper must be fixed before using it here, or the Parquet path must independently guarantee that truncated 2/3/4-byte sequences are removed.

extension/parquet/include/parquet_reader.hpp
```diff
@@ -270,7 +271,7 @@ struct ParquetOptions {
 	idx_t explicit_cardinality = 0;
 	bool can_have_nan = false; // if floats or doubles can contain NaN values
 	ParquetPrefetchStrategyOption prefetch_strategy = ParquetPrefetchStrategyOption::AUTO;
-};
+	StringColumnReader::Utf8ValidationOption utf8_validation_option = StringColumnReader::Utf8ValidationOption::STRICT;};
```
kryonix: `ParquetOptionsSerialization::Serialize` and `Deserialize` omit `utf8_validation_option`. A serialized scan therefore silently returns to `STRICT`:
```sql
SET debug_verify_serializer=true;
SELECT * FROM read_parquet(
    'data/parquet-testing/invalid.parquet',
    utf8_validation='replace'
);
```
Expected: `TREL?`. Actual: the strict-mode invalid UTF-8 error. Please add a versioned serialization property with `STRICT` as its backward-compatible default and cover the logical-plan round trip.

kryonix (changes requested):
Thanks for the PR! I think the overall direction makes sense. Exposing the existing UTF-8 handling modes through the Parquet reader is useful, and the distinction between PLAIN and DLBA decoding is handled thoughtfully.

I found a few issues that need addressing before this is ready:

- The changed fallback in `VerifyString` breaks existing BLOB reads with an internal error. BLOB values should continue to pass through without UTF-8 validation.
- `ignore` currently fails for truncated multibyte sequences, including the invalid UTF-8 fixtures in this PR. This comes from the existing `Utf8Proc::RemoveInvalid` bug you identified. We need to fix that first or otherwise ensure this path always returns valid UTF-8.
- The new option needs to be included in `ParquetOptionsSerialization`; otherwise serialized plans silently revert to `strict`.

It would also be good to add the planned SQLLogicTest coverage for PLAIN and DLBA across all three modes, including plan serialization and the existing BLOB test as focused validation.

These are fixable issues rather than concerns with the overall approach. Once they are addressed, I’d be happy to take another look.

The existing `parquet_blob_string.test` already protects finding 1. A focused UTF-8 test should include:

```sql
query T
SELECT * FROM read_parquet(
    'data/parquet-testing/invalid.parquet',
    utf8_validation='ignore'
)
----
TREL

query T rowsort
SELECT * FROM read_parquet(
    'data/parquet-testing/invalid_utf8_dlba.parquet',
    utf8_validation='ignore'
)
----
TREL
hello world

statement ok
SET debug_verify_serializer=true

query T
SELECT * FROM read_parquet(
    'data/parquet-testing/invalid.parquet',
    utf8_validation='replace'
)
----
TREL?

statement ok
SET debug_verify_serializer=false
```

third_party/utf8proc/utf8proc_wrapper.cpp
```diff
@@ -174,23 +174,30 @@ std::string Utf8Proc::RemoveInvalid(const char *s, size_t len) {
...
 			/* 4 byte sequence */
 			int utf8char = c & 0x07;
-			UTF8ExtraByteLoop<3, 0x1F0000>(first_pos_seq, utf8char, i, s, len, nullptr, nullptr);
+			type = UTF8ExtraByteLoop<3, 0x1F0000>(first_pos_seq, utf8char, i, s, len, nullptr, nullptr);
 		} else {
 			// invalid, do not write to output
 			continue;
 		}
 
+		if (type == UnicodeType::INVALID) {
```
kryonix, line 198: `UTF8ExtraByteLoop` advances `i` before discovering an invalid continuation byte. When it returns `INVALID`, the `continue` here causes the outer loop to advance again, so the byte that terminated the malformed sequence is also discarded, even when it is valid ASCII. For example:

```sql
SELECT hex(decode(from_hex('C328'), 'ignore'));
```

This should remove the invalid `C3` and return `28` (`(`), but currently returns an empty string. The same issue occurs with malformed three- and four-byte sequences and also affects Parquet's new `ignore` mode. Please ensure that the terminating byte is processed again, and add these regression cases:

```sql
query T
SELECT hex(decode(from_hex('C328'), 'ignore'));
----
28
query T
SELECT hex(decode(from_hex('E228A1'), 'ignore'));
----
28
query T
SELECT hex(decode(from_hex('E28241'), 'ignore'));
----
41
query T
SELECT hex(decode(from_hex('F0288CBC'), 'ignore'));
----
28
```

kryonix (changes requested):
Thanks for the update! You addressed all of the points from my first review: BLOB reads work again, truncated sequences are handled correctly, the option survives plan serialization, and the new focused tests cover the main PLAIN and DLBA behavior. 

I found one remaining edge case in `Utf8Proc::RemoveInvalid`. When a malformed multibyte sequence is terminated by valid ASCII, `ignore` currently removes that valid byte as well. This affects both `decode(..., 'ignore')` and the new Parquet path. I left a line comment with a reproducer and suggested regression cases.

There are also two cleanup items before this is ready: the changed C++ files still need formatting (`make format-fix`), and I think the description's statement about dictionary and `DELTA_BYTE_ARRAY` encodings should be rechecked because those decoders appear to delegate to `StringColumnReader::Plain`.

This is very close now, and the overall approach looks good. Once the remaining invalid-sequence case is fixed and the cleanup is done, I'd be happy to take one final look. Thanks again for the follow-up!

test/parquet/utf8_validation_plain_dba_dlba_dict.test
```diff
@@ -95,3 +95,63 @@ SELECT * FROM read_parquet(
 )
 ----
 TREL
+
+query T
```
kryonix, line 153: Nice, these cases now cover malformed continuation bytes across 2-, 3-, and 4-byte sequences. Since they exercise the shared `decode()` / `Utf8Proc::RemoveInvalid` behavior rather than Parquet specifically, could we move them to `test/fuzzer/pedro/utf8_invalid_unicode.test`?

There is also one Parquet path worth covering here: dictionary decoding already reaches `StringColumnReader::Plain()`, so `utf8_validation` applies to dictionary-encoded strings as well. A small dictionary-encoded `ignore`/`replace` case would protect that shared path and bring the tests in line with the actual implementation.

kryonix (commented):
Thanks for the update — this is in much better shape now. I re-reviewed the latest revision, and all of the earlier functional issues have been addressed:

- BLOB values continue to pass through without UTF-8 validation.
- `ignore` now handles truncated sequences and malformed continuation bytes correctly, without dropping valid bytes that follow them.
- The new option survives plan serialization.
- The tests cover PLAIN and DELTA_LENGTH_BYTE_ARRAY in all three modes, serializer verification, case-insensitive option parsing, and malformed 2-, 3-, and 4-byte sequences.

I ran the three focused regression files, and they all pass. I also verified the original malformed PLAIN case end to end: `C328` becomes `28` in `ignore` mode and `3F3F` in `replace` mode.

I have one remaining request around scope and test coverage. Dictionary and DELTA_BYTE_ARRAY decoding already flow through `StringColumnReader::Plain()`, so this option applies there too, despite the PR description saying they are deferred. I confirmed the dictionary path directly with malformed data in both `ignore` and `replace` modes. Please update the description to reflect the actual behavior and add focused dictionary coverage; the shared `decode()` regression cases would also fit better in the existing core UTF-8 test.

With that cleanup done, and once the draft is lifted so the currently skipped CI jobs run, this looks ready to approve. Nice work working through the edge cases here.

kryonix (approved):
Thanks for working through all of the feedback here. I completed a final review of the latest revision, and all of the earlier issues have been addressed: BLOB behavior is preserved, malformed and truncated UTF-8 sequences are handled correctly, the option survives serialization, and the tests now cover PLAIN, DELTA_LENGTH_BYTE_ARRAY, DELTA_BYTE_ARRAY, and dictionary encoding across the supported modes. I don't have any further code concerns.

The remaining CSV benchmark failure is unrelated infrastructure setup—the job timed out installing Clang tooling before reaching the build or benchmark, so that check only needs a rerun.

Nice work iterating on this. Approved!

pdet:
Thanks!

pdet:
Maybe as a nice follow-up, we could have a similar option for our other readers?

PR #24746, opened 2026-08-13
Make `Value` immutable

Maxxen:
This PR is stacked upon https://github.com/duckdb/duckdb/pull/24738.

The idea is the same: `duckdb::Value` is commonly part of shared pointer structures, and should therefore be immutable.

The main change is removing `Value::GetTypeMutable()` and making the mutable versions of `Value::TryCastAs` and `Value::Reinterpret` now return a new `optional<Value>` instead of modifying the value in place. 

Also `TryCastAs` used to throw or return false depending on whether you passed an error string, but it now never throw.


hannes (commented):
Nit: "casted" does not exist in English, should be "cast".  However, perhaps it was intentionally used to make it less ambiguous?

Maxxen:
> Nit: "casted" does not exist in English

TIL! Adjusted accordingly.

PR #24749, opened 2026-08-13
Preserve HUGEINT precision in quantile interpolation

matthiasgoergens:
HUGEINT interpolation converts its overflow-safe delta to double, which loses low digits for DECIMAL(38,*) endpoints above 2^53: the median of 0 and 2^54+2 is 9007199254740992 instead of 9007199254740993, and quantile_cont at weight 1/4096 over 0 and 10^38−1 returns 24414062499999999450393023304695808 instead of 24414062499999999999999999999999999.

This PR interpolates same-sign endpoints exactly: the delta is multiplied by the exact 53-bit mantissa of the weight through a three-limb 192-bit product, then shifted right, so neither the endpoints nor the delta are rounded to double. The shift handles limb-aligned weights such as 1/4096 explicitly, opposite-sign and too-wide deltas keep the overflow-safe endpoint fallback, and truncation towards zero is preserved in both sort directions.

The regression extends the #24314 test with the above-2^53 midpoint, a limb-aligned 1/4096 weight, negative quantile weights, the shared window fill() path, and an asymmetric overflow-fallback span.

### DIFF
--- a/src/function/scalar/operator/arithmetic.cpp
+++ b/src/function/scalar/operator/arithmetic.cpp
@@ -1362,9 +1362,8 @@ static hugeint_t ShiftRight192(uint64_t lower, uint64_t middle, uint64_t upper, 
 }
 
 // Computes floor(delta * weight) exactly for a non-negative HUGEINT delta and a finite double weight in [0, 1].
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks but this has already been solved by https://github.com/duckdb/duckdb/pull/24511

PR #24753, opened 2026-08-13
[Dev][Parser] Add `TokenIterator` abstraction

Tishj:
Previously we were passing around `vector<MatcherToken> &tokens, idx_t &token_cursor` in many functions.
This PR groups those variables together and encapsulates the modification of the state.

One instance of the iteration logic already grouped the two variables together, `PEGTransformerState`, this class is removed as part of this PR and usage of it has been replaced by `TokenIterator`.

### Misc
One interesting thing I noticed is what's now encapsulated as `SetPreviousTokenType`, which mutates the tokens vector, which could influence fallback parser extensions or fallback options.


Tishj:
Clanker analysis of the problem mentioned in `Misc`

Within PEG choices, semantic token types currently do not affect matching decisions. Matchers inspect token text and only inspect token type for the two end-of-input sentinels. So an abandoned branch changing IDENTIFIER to TABLE_NAME will not presently change which later alternative matches.

However, there is a genuine backtracking problem:
- Speculative MatchState copies have independent iterator positions.
- They all reference the same mutable token vector.
- Position changes are committed only on success.
- Token annotations are written immediately and are never rolled back.

Therefore, a failed list, optional, repeat iteration, or choice alternative can leave semantic annotations behind. That can affect sql_tokenize output and any later consumer of that vector. It is also fragile: a future matcher that considers semantic token types could accidentally make grammar behavior dependent on failed alternatives.

I would remove annotation from TokenIterator entirely. A cleaner design would be:
- Keep tokens and TokenIterator read-only.
- Store semantic annotations separately in MatchState, indexed by token position.
- Copy or delta-track annotations for speculative branches.
- Commit annotations together with the child position.
- For incomplete autocomplete input, explicitly retain annotations from the furthest viable branch instead of relying on incidental mutations.
- Have sql_tokenize combine lexical tokens with those committed semantic annotations.

That would make extension fallback isolation structural rather than merely an accidental property of MatchParseResult, while retaining the semantic classifications that motivated this behavior.

TL;DR
It might be a problem and we might want to separate these "semantic annotations" from the tokens, only merging them later on.

src/parser/token_iterator.cpp
```diff
@@ -0,0 +1,126 @@
...
+TokenIterator::TokenIterator(TokenIterator &other) : tokens(other.tokens), position(other.position) {
+}
+
+TokenIterator::TokenIterator(TokenIterator &&other) noexcept
+    : owned_tokens(std::move(other.owned_tokens)), tokens(other.tokens), position(other.position) {
+}
+
+bool TokenIterator::AtEnd() const {
+	auto current = Current();
+	return !current || current->type == TokenType::END_OF_INPUT;
```
Dtenwolde, line 27: Does this also need to include `TokenType::END_OF_INPUT_AUTOCOMPLETE`?

src/parser/token_iterator.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+bool TokenIterator::HasMoreStatements() const {
+	for (idx_t index = position; index < tokens.size(); index++) {
+		auto type = tokens[index].type;
+		if (type == TokenType::END_OF_INPUT) {
```
Dtenwolde, line 33: Same here

Dtenwolde (commented):
Minor thing

src/parser/token_iterator.cpp
Tishj: Hmm, that's a valid concern, but the method is created from the existing code:
```c++
	auto at_end_of_real_tokens = [&]() {
		return token_cursor >= tokens->size() || (*tokens)[token_cursor].type == TokenType::END_OF_INPUT;
	};
```

Mytherin:
Thanks!

Tishj:
Merge conflict detected!
*look inside*

old: `auto &token_text = state.tokens[state.token_index].text;`
new: `auto &token_text = token->text;`

PR #24754, opened 2026-08-13
Make External Pipeline Input Dependency-Aware

kryonix:
While looking into the CTE fan-out needed by #22797, it became obvious that marking `PhysicalHashJoin` as supporting external pipeline input is not sufficient. A normal hash-join probe pipeline is scheduled only after its build pipeline has finalized, whereas an externally fed pipeline is driven by the CTE producer instead of its own source tasks. The existing direct CTE path did not transfer the consumer's build-completion dependency to that producer.

As a result, the CTE producer could start pushing probe chunks before the hash table was finalized. Depending on the join orientation chosen by `BuildProbeSide`, this could construct probe state against a null or incomplete hash table. It showed up as wrong results in anti/outer-join rewrites, but the underlying issue was more general: external input was represented as a local property of the consumer, even though deciding whether it is safe requires the complete pipeline dependency graph.

This PR makes external pipeline input dependency-aware and enables it for hash joins. The important invariant is now explicit: every external producer must wait until the consumer has been initialized and all of its blocking inputs are complete, while consumer execution and finalization must remain open until every external producer has finished pushing.

## Resolving external input from the complete pipeline graph

I changed direct CTE selection from an eager decision made while building an individual CTE scan into a resolution step that runs after the complete set of pipelines and dependencies has been constructed, but before the pipelines are readied and scheduled.

A `Pipeline` now records the pipelines that produce its external input. During pipeline construction, eligible CTE scans are retained as unresolved candidates with their materialized source and their external source. `PipelineBuildState::ResolveExternalInputs` then selects the input mode and rewrites the relevant dependency edges in one batch.

The resulting schedule derives the external-input ordering directly from the graph:

- External producers wait for consumer initialization.
- They also inherit the consumer's blocking and dataflow readiness requirements.
- Consumer execution and `PrepareFinalize` wait for every external producer.
- Independent scheduling constraints remain attached to the consumer.

This replaces the previous special-case external-finish dependency. It also means that the decision does not depend on how or where a CTE was created. Optimizer-generated CTEs go through the same final graph resolution as CTEs that were present in the original query.

## Cycles and materialized fallback

Once build dependencies are transferred to external producers, mixed build/probe orientations can form real scheduling cycles. Disabling `BuildProbeSide` happens to avoid some of these cycles, but that would only hide the problem by forcing a particular orientation.

I added iterative cycle detection to the pipeline schedule. Each external scheduling edge retains sparse provenance identifying the consumer that introduced that exact dependency. This is necessary because several consumers can share the same producer and can even introduce the same pair of schedule stages; looking only at pipelines present in a cycle can materialize an unrelated consumer.

When resolution finds a cycle, it first removes ordering dependencies that exist only as parallelism heuristics. Meta-pipeline dependencies are now classified as `REQUIRED` or `OPTIONAL`: semantic ordering remains required, while dependencies used to avoid breadth-first execution or oversubscription may be removed when they conflict with external dataflow.

If a required cycle remains, the resolver falls back to the materialized CTE source for an external consumer whose own edge closes that cycle. It then rebuilds the schedule and repeats until the graph is acyclic. This allows uniformly oriented hash-join branches to stream, while mixed orientations materialize only the consumers needed to make the schedule valid.

## Avoiding serialized hash fan-out regressions

Direct exchange fan-out currently pushes a chunk through each direct consumer's `PipelineExecutor` serially. This is useful when it avoids a large materialization, but it can be slower than materializing a small input once and letting several hash-join pipelines scan it independently.

Operators can therefore report an external-input cost. Hash joins report `SERIALIZED_FANOUT`; ordinary pipelined consumers retain the default `PIPELINED` cost. For a CTE without a preselected consumer mode, cost selection only changes the normal direct-first behavior when all unresolved consumers have serialized fan-out cost and there are at least two of them.

For those CTEs, I estimate the materialized size from the producer cardinality and physical column widths. The resolver keeps the smallest estimated materializations within half of the operator memory limit and streams the remaining inputs. The arithmetic saturates on overflow, and packing uses guarded subtraction so one saturated estimate cannot hide another input that exceeds the budget.

Candidate modes are selected independently. Pipelined candidates retain the normal direct-first behavior, while serialized hash candidates participate in the materialization budget. This lets a small mixed CTE remain hybrid instead of forcing serialized direct pushes into every hash consumer. Preselected direct or buffered consumers retain their mode and count toward fan-out; if another consumer already requires materialization, unresolved hash consumers reuse it. Large mixed fan-outs still stream when their estimated materialization would exceed the budget.

This estimate is deliberately coarse. In particular, physical widths do not fully account for variable-length and nested payloads. A poor estimate can choose a slower mode, but it does not affect correctness or provide a hard memory reservation; materialized CTE storage remains buffer-managed and spillable.

## Hash-join execution details

`PhysicalHashJoin` now advertises external-input support. The dependency resolution ensures that a directly fed probe pipeline cannot run before its build has finalized, and that build-side fan-out can complete before dependent probe-side fan-out starts.

I also gave every direct consumer its own `DataChunk` wrapper. Operators are allowed to slice or otherwise change their input chunk wrapper while processing it. Reusing one wrapper across direct consumers meant that a spilling hash join could change the chunk seen by a later consumer. The vectors still reference the same underlying data, but wrapper state is isolated per executor and retained when a blocked push resumes.

The focused coverage includes `INNER`, `LEFT`, `RIGHT`, `FULL`, `SEMI`, `ANTI`, `MARK`, and `SINGLE` joins; NULLs and duplicates; empty and side-effecting producers; multiple producer pipelines; nested CTEs; prepared execution; spilling; mixed direct/materialized fan-out; exact cycle fallback; and saturated cardinality estimates. It verifies both results and the selected CTE modes.

For example, a single hash-join consumer can now be fed directly:

```sql
WITH c AS MATERIALIZED (SELECT i FROM range(10000) t(i))
SELECT count(*)
FROM range(10000) l(i)
JOIN c ON l.i = c.i;
```

The analyzed plan uses `STREAMING_FANOUT` with one `DIRECT` consumer. A small self-join over the same CTE remains `MATERIALIZED`, because serializing both hash-join consumers would be counterproductive. Under materialization pressure, the same shape can become `HYBRID_FANOUT`, with only the consumer needed to break the dependency cycle reading the materialized collection.


Tishj:
Ah yes, graphs

PR #24755, opened 2026-08-13
Fix checkpoint delta revert

artjomPlaunov:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9893, and possibly may help fix https://github.com/duckdblabs/duckdb-internal/issues/10209, or at least help narrow down ultimately find the bug there. 

If there was a checkpoint active, and a transaction appended to a delta index, upon rollback it would try to remove from the main index instead of the delta. This is largely solved by the index lifetime refactor that @sebastiaan-dev did, since we can just call the IndexEntry RevertAppend method instead, which simply checks for the existence of a delta index (and this is a local path for index append on transaction commit -> rollback, so it is not possible to have e.g. an append to a main index and a delta suddenly showing up on revert).


Mytherin:
Thanks!

PR #24758, opened 2026-08-13
Isolate temporary files per DuckDB instance

XiaocongDing:
**Problem**
DuckDB allows multiple instances to share one temp_directory, but temporary files are not isolated per
 instance. This causes two problems:

-  A running query can fail when another instance sharing the same temp_directory is destroyed. TemporaryDirectoryHandle currently scans the directory on destruction and deletes every file with the
duckdb_temp_ prefix, without knowing which instance owns it. As a result, instance B can delete spill
files still being used by instance A.

- Crashed instances bypass destructor cleanup and leave orphaned temporary files behind (#24576).

**Fix**
The fix isolates temporary files per instance, cleans an instance’s own files on shutdown, and reclaims
 files belonging to confirmed dead owners on startup.

1. Isolate temporary files by UUID

Each instance generates a UUID and includes it in every temporary filename, using names such as
 duckdb_temp_v1__....
The same UUID is recorded in an owner file named duckdb_temp_owner_v1_.lock. The instance keeps
  this file under an exclusive write lock for its lifetime. On POSIX systems this uses fcntl; an in-
  process registry handles multiple instances in the same process, where fcntl locks alone cannot
  distinguish instances.
When an instance is destroyed, it only removes temporary files containing its own UUID.

2. Clean up dead owners on startup

On startup, an instance scans the owner files and attempts to acquire an exclusive write lock on each
 one.
If the lock is acquired and the owner file is valid, no active process owns that UUID. Its temporary
 files are therefore considered orphaned and are removed together with the owner file.
If the owner is active, or its state cannot be safely determined, cleanup skips it. Cleanup failures do
 not prevent temporary files from being used.

### DIFF
--- a/src/include/duckdb/storage/temporary_file_manager.hpp
+++ b/src/include/duckdb/storage/temporary_file_manager.hpp
@@ -263,7 +263,8 @@ class TemporaryFileManager {
 	friend class TemporaryFileHandle;
 
 public:
-	TemporaryFileManager(DatabaseInstance &db, const string &temp_directory_p, atomic<idx_t> &size_on_disk);
+	TemporaryFileManager(DatabaseInstance &db, const string &temp_directory_p, string instance_id_p,
+	                     atomic<idx_t> &size_on_disk);
 	~TemporaryFileManager();
 
 private:
@@ -326,6 +327,8 @@ private:
 	DatabaseInstance &db;
 	//! The temporary directory
 	string temp_directory;
+	//! The identifier of the DuckDB instance that owns these temporary files
+	string instance_id;
 	//! Lock for parallel access
 	mutex manager_lock;
 	//! The set of active temporary file handles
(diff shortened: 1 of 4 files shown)


carlopi:
Hi, looked at PR title, this looks similar to https://github.com/duckdb/duckdb/pull/24655, could you have a look if that solves the issue as well?

XiaocongDing:
> Hi, looked at PR title, this looks similar to #24655, could you have a look if that solves the issue as well?

@carlopi carlopi  Hi, thanks for reply. I had a look, and there is indeed some overlap: #24655 adds a per-instance UUID to temporary file names, preventing different DatabaseInstances from colliding. 

The additional part of my patch is the owner-lock file associated with each UUID. It allows an instance to distinguish active UUIDs from inactive ones. This prevents instance B from deleting instance A’s live temporary files when B is deinitialized, as you mentioned. It also allows temporary files belonging to abnormally terminated instances to be identified and cleaned up as orphans.

If this direction sounds reasonable, I would be happy to continue on top of #24655 and port the owner-lock file implementation there.

PR #24760, opened 2026-08-13
Fix ASOF join over EXCEPT ALL / INTERSECT ALL children

### LINKED ISSUE #24672: INTERNAL Error when ASOF JOIN's left child is EXCEPT ALL / INTERSECT ALL
### What happens?

Please see the repro

### To Reproduce

I got `INTERNAL ERROR` when running the script, which might be unexpected

```sql
CREATE TABLE a(d INTEGER);
INSERT INTO a VALUES (1), (2);
CREATE TABLE b(d INTEGER);
CREATE TABLE r AS SELECT * FROM a;

SELECT * FROM (SELECT * FROM a EXCEPT ALL SELECT * FROM b) t0
ASOF LEFT JOIN r t1 ON t0.d >= t1.d;
```
```
INTERNAL Error: ExpressionExecutor::Execute called with a result vector of type INTEGER
that does not match expression type BIGINT
```
Same assertion with `INTERSECT ALL`:
```sql
SELECT * FROM (SELECT * FROM a INTERSECT ALL SELECT * FROM a) t0
ASOF LEFT JOIN r t1 ON t0.d >= t1.d;
```

However, the following scripts are clean:
```
-- EXCEPT (no ALL) is clean:
-- SELECT * FROM (SELECT * FROM a EXCEPT SELECT * FROM b) t0
-- ASOF LEFT JOIN r t1 ON t0.d >= t1.d;

-- Materializing the set-op before the ASOF is clean:
-- CREATE TABLE t0 AS SELECT * FROM a EXCEPT ALL SELECT * FROM b;
-- SELECT * FROM t0 ASOF LEFT JOIN r t1 ON t0.d >= t1.d;
```

### OS:

macOS

### DuckDB Version:

duckdb v2.0.0-alpha37247 (Cyanoptera) e500d77864

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Junwen An

### Affiliation:

NUS

- [x] Yes, I have

- [x] Yes, I have

Yes

RagamuffinCat:
Fixes #24672.

### Problem

`ASOF JOIN` fails with an internal error when one of its children is `EXCEPT ALL` / `INTERSECT ALL`:

```sql
CREATE TABLE a(d INTEGER); INSERT INTO a VALUES (1), (2);
CREATE TABLE b(d INTEGER);
CREATE TABLE r AS SELECT * FROM a;

SELECT * FROM (SELECT * FROM a EXCEPT ALL SELECT * FROM b) t0
ASOF LEFT JOIN r t1 ON t0.d >= t1.d;
-- INTERNAL Error: ExpressionExecutor::Execute called with a result vector of
--                 type INTEGER that does not match expression type BIGINT
-- (assertion build: expressions.size() == result.ColumnCount())
```

It also reproduces with the set operation on the build side, with a different symptom:

```sql
SELECT * FROM r t1
ASOF LEFT JOIN (SELECT * FROM a EXCEPT ALL SELECT * FROM b) t0 ON t1.d >= t0.d;
-- INTERNAL Error: Vector::Reference used on vector of different type
--                 (source BIGINT referenced INTEGER)
```

### Root cause

`EXCEPT ALL` / `INTERSECT ALL` get bag semantics from a `ROW_NUMBER` column added during physical planning. To make the ANTI/SEMI `PhysicalHashJoin` emit that column, `CreatePlan(LogicalSetOperation &)` appends a `BIGINT` to `op.types` and never restores it:

https://github.com/duckdb/duckdb/blob/c6f7f3e2502c0a5141f339cb208ba21f770d6661/src/execution/physical_plan/plan_set_operation.cpp#L92

The physical plan still ends in a projection that strips the column, so from that point on the `LogicalSetOperation`'s types no longer match the physical operator it was planned into.

`PlanAsOfLoopJoin()` then reads the **logical** children's types *after* they have been planned:

```cpp
auto &left  = CreatePlan(*op.children[0]);   // <- mutates children[0]->types
auto &right = CreatePlan(*op.children[1]);
...
join_op.types = op.children[1]->types;
const auto &probe_types = op.children[0]->types;   // one column too wide
```

Everything downstream shifts by one column: `join_op.types` declares 4 columns for a 3-column NLJ, `right_projection_map` gains a phantom entry, an extra `first(<row_number>)` aggregate is bound, and the final `PhysicalProjection` ends up with 3 expressions writing into a 2-column output chunk.

This is confined to the NLJ optimisation path: `SET asof_loop_join_threshold = 0` avoids it, and `EXCEPT` without `ALL` is fine because it never appends the column. Both match the workarounds reported in the issue.

### Fix

- `plan_asof_join.cpp`: `PlanAsOfLoopJoin()` builds operators on top of the planned children, so use `probe.GetTypes()` / `build.GetTypes()` instead of `op.children[i]->types`. The `debug_asof_iejoin` path had the same pattern (it held a reference to `op.children[1]->types` and appended to it in place), it now copies.
- `plan_set_operation.cpp`: restore `op.types` after the hash join has copied them, so the ROW_NUMBER column stays an implementation detail of the physical plan.

Either change alone fixes the issue. I kept both. The first makes ASOF planning robust against logical/physical type mismatches, the second removes the mismatch at its source.

### Tests

New `test/sql/join/asof/test_asof_join_setops.test`, run at `asof_loop_join_threshold` 0 and 32 so both the `PhysicalAsOfJoin` and the NLJ paths are covered: set operation on the probe side, on the build side, on both sides, `ASOF JOIN` as well as `ASOF LEFT JOIN`, with a non-comparison predicate, and a case where `EXCEPT ALL` actually removes rows so results are checked rather than just the absence of a crash.

### DIFF
--- a/src/execution/physical_plan/plan_asof_join.cpp
+++ b/src/execution/physical_plan/plan_asof_join.cpp
@@ -59,8 +59,12 @@ PhysicalPlanGenerator::PlanAsOfLoopJoin(LogicalComparisonJoin &op, PhysicalOpera
 
 	LogicalComparisonJoin join_op(InverseJoinType(op.join_type));
 
-	join_op.types = op.children[1]->types;
-	const auto &probe_types = op.children[0]->types;
+	// Use the types of the planned children: planning a child can change the types of the logical operator
+	// (EXCEPT ALL/INTERSECT ALL append a ROW_NUMBER column to it)
+	const auto &probe_types = probe.GetTypes();
+	const auto &build_types = build.GetTypes();
+
+	join_op.types = build_types;
 	join_op.types.insert(join_op.types.end(), probe_types.begin(), probe_types.end());
 
 	// Project pk
@@ -72,14 +76,14 @@ PhysicalPlanGenerator::PlanAsOfLoopJoin(LogicalComparisonJoin &op, PhysicalOpera
 	//	we have to track this carefully...
 	join_op.left_projection_map = op.right_projection_map;
 	if (join_op.left_projection_map.empty()) {
-		for (idx_t i = 0; i < op.children[1]->types.size(); ++i) {
+		for (idx_t i = 0; i < build_types.size(); ++i) {
 			join_op.left_projection_map.emplace_back(i);
 		}
 	}
 
 	join_op.right_projection_map = op.left_projection_map;
 	if (join_op.right_projection_map.empty()) {
-		for (idx_t i = 0; i < op.children[0]->types.size(); ++i) {
+		for (idx_t i = 0; i < probe_types.size(); ++i) {
 			join_op.right_projection_map.emplace_back(i);
 		}
 	}
@@ -87,8 +91,8 @@ PhysicalPlanGenerator::PlanAsOfLoopJoin(LogicalComparisonJoin &op, PhysicalOpera
 	// Remap predicate column references.
 	auto predicate = CreatePredicateFromConditions(op.conditions);
 	if (predicate) {
-		const auto lhs_width = op.children[0]->types.size();
-		const auto rhs_width = op.children[1]->types.size();
+		const auto lhs_width = probe_types.size();
+		const auto rhs_width = build_types.size();
 
 		ExpressionIterator::EnumerateExpression(predicate, [&](Expression &child) {
 			if (child.GetExpressionClass() == ExpressionClass::BOUND_REF) {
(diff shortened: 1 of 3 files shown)


hawkfish:
Hi @IrisphoraChen - Can you mark this as no longer a draft?

hawkfish:
Hi @IrisphoraChen - Thanks for the PR! Can you mark it as not a draft so we can get some test coverage before I look too deeply? Also, you can check all the `ASOF` tests locally by sticking them in a file and using `unittest -f <file>`.

RagamuffinCat:
Done, marked as ready for review. Thanks for the -f tip!

hawkfish (approved):
Thanks - LGTM!

Mytherin:
Thanks!

PR #24764, opened 2026-08-14
Add DuckTransactionManager::GetLastWriteCommit(), a commit id that only advances on writes

maxdemarzi:
### Summary

`GetLastCommit()` reads like a "has anything changed in this database" signal, and it is easy to mistake for one, but it cannot serve as that. DuckDB commits a transaction for read-only statements too, and `CommitTransaction` assigns `last_commit = info.commit_id` unconditionally, so the counter advances under a bare `SELECT`. Two consecutive `SELECT count(*)` statements were measured moving it 9 → 11 → 13.

This adds `last_write_commit` beside it, assigned in the same place but only when `DuckTransaction::ChangesMade()` held, and exposes it as `GetLastWriteCommit()`. It also corrects the doc comment on `GetLastCommit()` to say that it advances on reads.

The change is additive and does not alter any existing behaviour: no existing field, method or code path changes meaning, and the cost is one atomic store on the commit path of *writing* transactions only, with no work at all added to the read path.

### Why this is worth having in core

Any extension that caches something derived from table contents — join-elimination verdicts, graph indexes, materialised summaries — wants a cheap "is what I computed still valid?" check. The obvious one is "record the counter, compare it later", and against `GetLastCommit()` it silently never succeeds, because the extension's own validation query advances the counter it is about to re-read.

That failure mode is unusually quiet: the cache stays *correct*, it is merely never useful, so nothing fails and nothing is logged. In the case that motivated this patch the affected cache had a 0% hit rate for every key from the day it was written, and the cost was 7.2 seconds of an 8.2 second query.

The alternatives available to an extension today are all unsound rather than merely awkward:

- **Row counts / `DataTable::GetTotalRows()`** — DuckDB updates in place through `UpdateSegment`, so `UPDATE t SET col = <new value>` moves no rows and changes no count. This is the case that rules out every cheap signal.
- **Catalog version** — unmoved by DML.
- **`StorageManager::GetWALSize()` / `GetWALEntriesCount()`** — reset at checkpoint, so they can go backwards.

`ChangesMade()` is already the predicate DuckDB itself uses to decide whether a transaction needs a WAL write, an automatic checkpoint, or to be kept around for cleanup. It covers appends (transaction-local storage), deletes and in-place updates (undo buffer, `UndoFlags::DELETE_TUPLE` / `UndoFlags::UPDATE_TUPLE`), and catalog changes. Exposing a commit id gated on the predicate you already compute seemed like the smallest thing that makes this class of extension-side caching possible at all.

### One implementation detail worth a reviewer's eye

`ChangesMade()` is sampled **before** `transaction.Commit()` rather than at the existing `last_commit` assignment site below it. `LocalStorage::Commit` moves the transaction-local storage out of the transaction via `LocalTableManager::MoveEntries`, so `storage->ChangesMade()` becomes false once the commit has completed; sampling at the assignment site would miss append-only transactions.

(`LocalStorage::Flush` does also push an `UndoFlags::INSERT_TUPLE` entry, so in practice the undo buffer would usually still report changes — but that is incidental, and it does not hold on the `is_dropped` / fully-deleted early-return paths. Sampling before the commit does not depend on it.)

Related, and deliberately *not* touched here to keep the diff minimal: the existing `last_commit` has no initialiser, and DuckDB builds as C++11 where `std::atomic`'s default constructor does not value-initialise, so it starts indeterminate. The new member initialises explicitly to `0`; commit ids come from `current_start_timestamp++` starting at 2, so `0` is never a real commit id and unambiguously means "nothing has been modified since this database was attached". Happy to fix `last_commit` in the same PR if you would like.

### Tests

`test/api/test_last_write_commit.cpp` covers three things:

- read-only statements do not advance it — and asserts in the same test that `GetLastCommit()` *did* move across those same reads, so "the write counter did not move" is a real observation rather than an artefact of nothing having happened;
- appends, in-place updates, deletes and DDL each do advance it;
- a rolled-back transaction does not.

`CONTRIBUTING.md` strongly prefers sqllogictest, and I'd have used it if I could: the counters are C++ API surface with no SQL-visible spelling, so there is nothing for a `.test` file to observe. This is the "absolutely need to" case rather than a preference.

### Disclosure

The motivating consumer is an out-of-tree extension of mine, so this is internal C++ API being grown for a caller you cannot see, and I understand if that is not a trade you want to make. I am not asking for urgency — nothing is broken in DuckDB itself, and the extension is correct today, just unable to amortise a check.

If you would rather not grow this surface, I would still suggest taking the one-line documentation correction on `GetLastCommit()` on its own. The name invites exactly the misuse described above, and the comment currently says nothing either way.

A per-database counter is also admittedly coarse — it advances for a write to *any* table in the attached database, so consumers over-invalidate. That is the safe direction and it is a large improvement on "never valid", but if you would prefer per-table granularity before merging anything, that is a much larger change and I am happy to be told so.


Mytherin:
Thanks - can't we instead just not increment the last commit on read-only commits? That wouldn't require adding any additional infrastructure and would simplify instead.

maxdemarzi:
Absolutely!  I just assumed there was a very good reason why it was being incremented on read-only commits.

src/transaction/duck_transaction_manager.cpp
```diff
@@ -379,6 +379,12 @@ ErrorData DuckTransactionManager::CommitTransaction(ClientContext &context, Tran
 	CommitInfo info;
 	info.commit_id = GetCommitTimestamp();
 
+	// Whether this transaction modified any data or catalog entries.
+	// This must be read *before* DuckTransaction::Commit below: committing moves the
+	// transaction-local storage out of the transaction (LocalStorage::Commit ->
+	// LocalTableManager::MoveEntries), so ChangesMade() can flip to false afterwards.
```
Mytherin: Can you remove these excessive comments?

test/api/test_last_commit_writes_only.cpp
```diff
@@ -0,0 +1,98 @@
...
+transaction_t ReadLastCommit(Connection &con, const string &table_name) {
+	con.BeginTransaction();
+	// QualifiedName overload, not the (catalog, schema, name) one: that form is
+	// deprecated on main in favour of folding the qualification into a
+	// QualifiedName. The deprecated form also compiles here, but a new test
+	// should not add a deprecation warning to the build.
+	auto &table_entry = Catalog::GetEntry<TableCatalogEntry>(
+	    *con.context, QualifiedName(INVALID_CATALOG, DEFAULT_SCHEMA, table_name));
+	auto &transaction_manager = DuckTransactionManager::Get(table_entry.GetStorage().GetAttached());
+	auto last_commit = transaction_manager.GetLastCommit();
```
Mytherin: Can we add a scalar function that exposes this for a given catalog - e.g. `last_commit(<attached_db>)`? Then we can rewrite these tests to be SQLLogic tests.

Mytherin (commented):
Thanks for the fixes - looks good, some more minor comments

src/transaction/duck_transaction_manager.cpp
maxdemarzi: Trimmed both comments down to one line each.

test/api/test_last_commit_writes_only.cpp
maxdemarzi: Added `last_commit(database)` (src/function/scalar/system/last_commit.cpp, via functions.json + generate_functions.py, following current_transaction_id()'s pattern) and rewrote the tests as test/sql/transactions/last_commit.test. Snapshots use SET VARIABLE/getvariable() rather than a table, since inserting into a table to record a snapshot is itself a write that would advance the very counter being measured.

Ran the full test/sql/transactions/* suite (1396 assertions), the [api] tag (417951 assertions), and 43 checkpoint/WAL tests (11903 assertions) -- covering every current caller of GetLastCommit() -- no regressions.

PR #24768, opened 2026-08-14
Prune segment for nested types on validity check

dentiny:
Hi team, I found for `IS NULL` predicate, not all the types prune segment well. For example
```sql
memory D PRAGMA enable_profiling = 'json';
memory D SET tracked_metrics = ['query.total_rows_scanned'];

-- scalar types prune as expected
memory D CREATE TABLE scalar_null AS
         SELECT CASE WHEN i >= 99999 THEN NULL ELSE i END AS v
         FROM range(100000) t(i);
memory D SELECT count(*) FROM scalar_null WHERE v IS NULL;
{
    "query": {
        "total_rows_scanned": 54944
    }
}

-- struct doesn't prune segment
memory D CREATE TABLE struct_null AS
         SELECT CASE WHEN i >= 99999 THEN NULL ELSE {'v': i} END AS s
         FROM range(100000) t(i);
memory D SELECT count(*) FROM struct_null WHERE s IS NULL;
{
    "query": {
        "total_rows_scanned": 100000
    }
}

-- list doesn't prune segment
memory D CREATE TABLE list_null AS
         SELECT CASE WHEN i >= 99999 THEN NULL ELSE [i] END AS l
         FROM range(100000) t(i);
memory D SELECT count(*) FROM list_null WHERE l IS NULL;
{
    "query": {
        "total_rows_scanned": 100000
    }
}
```
This PR consults validity zonemap for these nested types, after change segment prunes correctly.
```sql
memory D CREATE TABLE struct_null AS
                  SELECT CASE WHEN i >= 99999 THEN NULL ELSE {'v': i} END AS s
                  FROM range(100000) t(i);
memory D SELECT count(*) FROM struct_null WHERE s IS NULL;
{
    "query": {
        "total_rows_scanned": 54944
    }
}

memory D CREATE TABLE list_null AS
                  SELECT CASE WHEN i >= 99999 THEN NULL ELSE [i] END AS l
                  FROM range(100000) t(i);
memory D SELECT count(*) FROM list_null WHERE l IS NULL;
{
    "query": {
        "total_rows_scanned": 54944
    }
}
```

### DIFF
--- a/src/include/duckdb/storage/table/column_data.hpp
+++ b/src/include/duckdb/storage/table/column_data.hpp
@@ -258,6 +258,11 @@ protected:
 
 	idx_t GetVectorCount(idx_t vector_index) const;
 
+	static bool IsDirectNullCheckFilter(const TableFilter &filter);
+	FilterPropagateResult CheckValidityZonemap(ColumnScanState &state, TableFilter &filter,
+	                                           optional_ptr<SegmentNode<ColumnSegment>> &checked_segment,
+	                                           ColumnData &validity_column);
+
 private:
 	void UpdateCompressionFunction(SegmentLock &l, const CompressionFunction &function);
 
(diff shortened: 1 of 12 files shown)


Mytherin:
Thanks!

PR #24769, opened 2026-08-14
Use group matrix for build/test job split of extensions

smvv:
Currently, the extension jobs are sharded, and therefore building and testing a subset of the extension inside one CI job. Some extensions have tests that depend on other extensions (like httpfs) and if that extension in not available in the same CI job, the test is skipped.

This PR splits the extension jobs into build and test jobs. The test job can download the artifacts from the build job (eventually using an older distro image) and run tests on all extensions inside the same machine.

This PR uses new `build.yml` from extension-ci-tools. See related [PR](https://github.com/duckdb/extension-ci-tools/pull/398).


.github/patches/extensions/odbc_scanner/0001-move-strip-flag-to-linker.patch
```diff
@@ -0,0 +1,14 @@
...
+index 7a19484..68b5353 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -149,8 +149,7 @@ set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+ set(CMAKE_C_VISIBILITY_PRESET hidden)
+ set(VISIBILITY_INLINES_HIDDEN ON)
+ if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+-    set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -s")
+-    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -s")
++    set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} -s")
```
smvv, line 11: note: `-s` is a linker flag, not a compile flag.
(Clang gives a warning that the CLI argument is unused while compiling.)

.github/workflows/Extensions.yml
```diff
@@ -49,166 +52,56 @@ env:
...
-      upload_all_extensions: true
-      extension_name: ${{ matrix.artifact_prefix }}
-      # NOTE: on PRs we exclude some archs to speed things up
-      reduced_ci_mode: ${{ needs.load-extension-configs.outputs.reduced_ci_mode }}
-      set_caller_as_duckdb: true
       duckdb_version: ${{ inputs.duckdb_commit }}
-      override_ci_tools_repository: duckdb/extension-ci-tools
+      image_version: ${{ inputs.image_version }}
+      groups: |
+        core:
```
smvv, line 65: note: this new workflow allows us to have 1 matrix job and N build + M test jobs.

Previously, we had M matrix jobs with N build jobs.

PR #24770, opened 2026-08-14
Fix `PARTITION BY` syntax for `COPY TO` on multiple columns

Dtenwolde:
Fix https://github.com/duckdblabs/duckdb-internal/issues/9279

We now also support `PARTITION BY` and `ORDER BY` in the `COPY TO` settings as well as their legacy counterparts `PARTITION_BY`/`PARTITIONED_BY` and `ORDER_BY`.

This was sparked by a previous issue with `PARTITION BY a, HEADER` being read as: Partition this by both the columns `a` and `header`. This was due to the parser greedily consuming possible columns for `partition_by`. This is now fixed as `PARTITION_BY` & `PARTITION BY` will either: 1) only consume a single column `PARTITION BY a` or 2) require parenthesis if you want to partition by multiple columns: `PARTITION BY (a, c)`

@lnkuiper question is whether we should have the same support for `ORDER BY` (Currently this new spelling is only allowed inside the parenthesized option list: 
`COPY (...) TO 'output' (ORDER BY (a, b DESC));`

But we could also add this syntax:  
````
COPY (...) TO 'output' ORDER BY a, HEADER;
COPY (...) TO 'output' ORDER BY (a, b DESC), HEADER;
```
I can update that in this PR, but it might be better as a follow-up


Mytherin:
Thanks!

PR #24771, opened 2026-08-14
EXTERNAL RESOURCE: give the create params their own grammar rule

carlopi:
`CreateExternalResourceStmt` reused `AttachOptions` for its `(k v, ...)` create params. They are a different bag of options that happens to have the same shape, and a future `ATTACH TO ... EXTERNAL RESOURCE` can carry both at once, so give them their own rule.

The accepted syntax is unchanged.

Alternative to https://github.com/duckdb/duckdb/pull/24691, according to @Dtenwolde's comment.


Dtenwolde (approved):
Looks good, indeed as discussed :)

PR #24774, opened 2026-08-14
Add typed_value column to duckdb_settings()

lghavami:
Added an extra **typed_value** column to output of **duckdb_settings()** with column type variant (issue [#6949](https://github.com/duckdblabs/duckdb-internal/issues/6949)) + test.

### DIFF
--- a/src/function/table/system/duckdb_settings.cpp
+++ b/src/function/table/system/duckdb_settings.cpp
@@ -34,9 +34,6 @@ static unique_ptr<FunctionData> DuckDBSettingsBind(ClientContext &context, Table
 	names.emplace_back("value");
 	return_types.emplace_back(LogicalType::VARCHAR);
 
-	names.emplace_back("typed_value");
-	return_types.emplace_back(LogicalType::VARIANT());
-
 	names.emplace_back("description");
 	return_types.emplace_back(LogicalType::VARCHAR);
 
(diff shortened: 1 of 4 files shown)


src/function/table/system/duckdb_settings.cpp
```diff
@@ -34,6 +34,9 @@ static unique_ptr<FunctionData> DuckDBSettingsBind(ClientContext &context, Table
 	names.emplace_back("value");
 	return_types.emplace_back(LogicalType::VARCHAR);
 
+	names.emplace_back("typed_value");
```
artjomPlaunov: append typed_value at the end to preserve the existing column indexes

test/sql/table_function/duckdb_settings.test
```diff
@@ -20,3 +20,8 @@ query II
 SELECT name, value FROM duckdb_settings() WHERE name='default_null_order';
 ----
 default_null_order	NULLS_LAST
+
+query I
```
artjomPlaunov, line 24: can we one or two more test queries here, to showcase inspecting the structured types, e.g. something like:  

```sql
query III
SELECT
    typeof(value),
    typeof(typed_value),
    variant_typeof(typed_value)
FROM duckdb_settings()
WHERE name = 'standard_vector_size';
----
VARCHAR	VARIANT	UINT64
```

artjomPlaunov (changes requested):
Thanks, left two comments. Looks good! I think the CI failure you just need to run the formatter again

src/function/table/system/duckdb_settings.cpp
lghavami: done!

test/sql/table_function/duckdb_settings.test
lghavami: added!

artjomPlaunov (approved):
Thanks!

PR #24772, opened 2026-08-14
[Dev] Refactor `BaseTokenizer`, and rename to `Tokenizer`

Tishj:
This PR is part of https://github.com/duckdblabs/duckdb-internal/issues/10346

Pull out behavior-changing derived classes and rework them into a dependency-injected interface instead.

PR #24775, opened 2026-08-14
[Dev] `ClientContext` plumbing for `Tokenizer`

Tishj:
Built on top of #24772

To make the Tokenizer extensible, we need to plumb the ClientContext through all the existing paths, so the callback can be taken from there and used to create the Tokenizer implementation of an extension.

(The Tokenizer extension will be global-scoped (db-scoped), not client-context-scoped, but to get the dbconfig, the easiest way is through the client context, so this ensures that)


Tishj:
Perhaps it makes sense to first move the `Tokenizer` into the `ParserCache` and revisit this after

PR #24776, opened 2026-08-14
Fix case restoration for mismatched case column references

### LINKED ISSUE #24759: Regression (main): output column name no longer restores the catalog case for case-mismatched column references
### What happens?

Since 0.3.x, a non-aliased column reference written in a different case than the catalog name is renamed back to the catalog case in the result schema ([source](https://github.com/duckdb/duckdb/blob/60279d26a7090e17a06a134b8f650c573033552a/src/planner/bind_context.cpp#L246-L251)). For example, given a table `t` with a column that has been registered as  `MixedCase` in the catalog, `SELECT mixedcase FROM t` returns a column named `MixedCase`. This also applies to quoted column references.

On the current main, the written case wins instead, so the same query as above now returns `mixedcase`, i.e., CTAS statements will materialize the changed name.

The latest release (1.5.5) is unaffected, but the current nightly (1.6.0.dev343) is affected. This behavior change is a side-effect of the recent `Identifier` class refactor (#23161) and seems unintended, hence the ticket.

cc @Mytherin

### To Reproduce

The Python driver is used only as a convenience for installing/using the two versions easily (nightly vs latest). The result names come from the engine's bound statement, and the CTAS repro persists the changed name in the catalog with no client involvement.

```python
# mre.py
import duckdb
print("duckdb", duckdb.__version__)
con = duckdb.connect()
con.execute("CREATE TABLE t (MixedCase INTEGER)")
print(con.execute("SELECT mixedcase FROM t").description[0][0])
con.execute("CREATE TABLE o AS SELECT mixedcase FROM t")
print(con.execute("SELECT column_name FROM duckdb_columns() WHERE table_name = 'o'").fetchall())
```

The script's output before vs after:

```
$ uv run --no-project --with 'duckdb==1.5.5' mre.py
Installed 1 package in 2ms
duckdb 1.5.5
MixedCase
[('MixedCase',)]
$ uv run --no-project --with 'duckdb==1.6.0.dev343' --prerelease allow mre.py
Installed 1 package in 3ms
duckdb 1.6.0.dev343
mixedcase
[('mixedcase',)]
```

----

I believe the root cause is that `CreateColumnReference` now compares `Identifier` values instead of strings ([source](https://github.com/duckdb/duckdb/blob/5e66dad275e117531786fd9dddf4b75f4763873a/src/planner/bind_context.cpp#L246-L251)), which has CI semantics ([1](https://github.com/duckdb/duckdb/blob/5e66dad275e117531786fd9dddf4b75f4763873a/src/include/duckdb/common/identifier.hpp#L100-L102), [2](https://github.com/duckdb/duckdb/blob/5e66dad275e117531786fd9dddf4b75f4763873a/src/common/identifier.cpp#L21-L23)). So, `registered_name != column_name` is essentially dead code now: the two names can only ever differ in case and the CI comparison is blind to that difference.

### OS:

MacOS Tahoe 26.5.2

### DuckDB Version:

Latest nightly build (1.6.0.dev343)

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

George Sittas

### Affiliation:

Fivetran Inc.

- [x] Yes, I have

- [x] Yes, I have

tempewda:
Fixes #24759.

Since the `Identifier` class was recently refactored to use case-insensitive equality, `registered_name != column_name` no longer detects identifiers that differ only by case. This skipped the alias-restoration logic for non-aliased column references, causing output schemas to use the user's typed case instead of the catalog spelling.

This PR uses a strict string comparison, `registered_name.GetIdentifierName() != column_name.GetIdentifierName()`, for this  specific output-name check while preserving case-insensitive identifier lookup behavior.

### DIFF
--- a/src/planner/bind_context.cpp
+++ b/src/planner/bind_context.cpp
@@ -244,7 +244,7 @@ unique_ptr<ParsedExpression> BindContext::CreateColumnReference(const BindingAli
 		return ExpandGeneratedColumn(binding->Cast<TableBinding>(), column_name);
 	}
 	auto &registered_name = binding->GetRegisteredColumnName(column_name);
-	if (registered_name != column_name) {
+	if (registered_name.GetIdentifierName() != column_name.GetIdentifierName()) {
 		// because of case insensitivity in the binder we rename the column to the original name
 		// as it appears in the binding itself
 		result->SetAlias(registered_name);
@@ -286,7 +286,7 @@ unique_ptr<ParsedExpression> BindContext::CreateColumnReference(const Identifier
 		return ExpandGeneratedColumn(binding->Cast<TableBinding>(), column_name);
 	}
 	auto &registered_name = binding->GetRegisteredColumnName(column_name);
-	if (registered_name != column_name) {
+	if (registered_name.GetIdentifierName() != column_name.GetIdentifierName()) {
 		// because of case insensitivity in the binder we rename the column to the original name
 		// as it appears in the binding itself
 		result->SetAlias(registered_name);
--- a/test/sql/binder/test_case_insensitive_binding.test
+++ b/test/sql/binder/test_case_insensitive_binding.test
@@ -138,3 +138,30 @@ query I
 SELECT hello FROM (SELECT 42) tbl("HeLlO")
 ----
 42
+
+statement ok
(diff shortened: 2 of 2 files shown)


lnkuiper:
LGTM. Thanks!

PR #24777, opened 2026-08-14
External Extension Repositories

hannes:
Currently DuckDB can only install extensions from built-in extension repositories (`core`, `core_nightly`, `community`, ...). This PR lets users register their own trusted extension repositories, so an organization can host and sign its own extensions and have them install and load just like the built-in ones:

```sql
-- custom repositories are opted into at startup (see "Security model" below);
-- once enabled, they are used entirely through SQL:
CREATE EXTENSION REPOSITORY my_repo FROM 'https://extensions.example.com';
INSTALL AND LOAD my_ext FROM my_repo;
```

**SQL interface**

A repository has a locally unique name, a URL prefix to download extensions from, and one or more RSA-2048 public keys that are trusted to sign the extensions served from it. The prefix can point at any filesystem DuckDB can read — a local path, http(s), s3, etc. A repository is removed again with:

```sql
DROP EXTENSION REPOSITORY my_repo;
```

**Naming the repository on `LOAD`**

`INSTALL` already takes an optional `FROM` clause naming the repository to install from. `LOAD` now takes the same optional `FROM` clause, which makes the origin explicit and reads naturally:

```sql
INSTALL my_ext FROM my_repo;     -- unchanged
LOAD my_ext FROM my_repo;
LOAD h3 FROM community;          -- the built-in repositories can be named too
```

For `LOAD`, the `FROM` clause is **optional for core and community** but effectively **required for user-repository extensions**. A bare `LOAD my_ext` only resolves core and community extensions (the backwards-compatible path); user-repository extensions live in a per-repository subfolder (see below) and cannot be found by a bare name, so a bare `LOAD` can never accidentally pick up a user-repository extension. The `FROM` clause is also a trust assertion: `LOAD foo FROM core` when `foo` was actually installed from `community` (or from a user repository) is an error, which prevents loading an extension under a repository whose signing keys you did not intend to trust. An unknown repository name is rejected at bind time (`'x' is not a known repository name`).

**`INSTALL AND LOAD`**

`INSTALL AND LOAD x [FROM y]` installs an extension and loads it in a single statement, saving the usual `INSTALL x; LOAD x;` pair:

```sql
INSTALL AND LOAD my_ext FROM my_repo;   -- install from my_repo, then load
INSTALL AND LOAD httpfs;                -- install from core, then load
FORCE INSTALL AND LOAD my_ext FROM my_repo;  -- composes with FORCE
```

It composes with everything `INSTALL` already accepts (the `FROM` clause, the `FORCE` prefix, `VERSION`). The load step reuses the install's repository, so the extension is loaded from exactly where it was just installed.

**On-disk layout**

To de-conflict extension names across repositories, extensions from a user repository are installed under a per-repository subfolder:

```
~/.duckdb/extensions/<version>/<platform>/
    httpfs.duckdb_extension                     # core / community (unchanged, flat)
    repositories/my_repo/my_ext.duckdb_extension  # user repositories
    repositories/other/my_ext.duckdb_extension    # same name, different repo — no collision
```

Core and community extensions keep the existing flat top-level layout, so nothing about their on-disk paths changes.

**Where the keys come from**

At `CREATE` time, DuckDB fetches the repository's public keys from a metadata file the repository serves at `<prefix>/.well-known/duckdb-extension-repo.json` and pins them into the repository definition. The file is a json object with a `signature_keys` array of PEM-encoded keys:

```json
{
  "signature_keys": [
    "-----BEGIN PUBLIC KEY-----\n...\n-----END PUBLIC KEY-----",
    "-----BEGIN PUBLIC KEY-----\n...\n-----END PUBLIC KEY-----"
  ]
}
```

All keys in the array are parsed and pinned, and all are trusted for extensions from this repository, so a repository can publish a new key alongside the old one and rotate without breaking existing installs.

Because the key arrives over the same channel as the extensions themselves, `CREATE` returns the name, the prefix and the sha256 fingerprint of each pinned key:

```
┌─────────┬────────────────────────────────┬───────────────────┐
│  name   │             prefix             │ key_fingerprints  │
├─────────┼────────────────────────────────┼───────────────────┤
│ my_repo │ https://extensions.example.com │ [sha256:a3f1c2…]  │
└─────────┴────────────────────────────────┴───────────────────┘
```

The intended workflow is to compare that fingerprint against one published out of band (docs, an internal wiki, a signed announcement). The key can also be given directly in the `CREATE` statement instead, as full PEM or as the compact base64 body, which skips the fetch entirely:

```sql
CREATE EXTENSION REPOSITORY my_repo FROM 'https://extensions.example.com'
    USING PUBLIC KEY '-----BEGIN PUBLIC KEY----- ...';

-- multiple keys, for rotation
CREATE EXTENSION REPOSITORY my_repo FROM 's3://my-bucket/extensions'
    USING PUBLIC KEYS 'key1', 'key2';
```

`duckdb_extension_repositories()` lists both built-in and user repositories, exposing `key_fingerprints` and `public_keys` as `LIST(VARCHAR)`, so a pinned repository can be audited at any time — including after the prefix it was created from is no longer reachable.

**Which key signed an installed extension**

When an extension is installed, the fingerprint of the key that actually verified its signature is recorded in the install metadata and surfaced as a new `signature_key_fingerprint` column of `duckdb_extensions()`. It uses the same `sha256:…` format as `CREATE EXTENSION REPOSITORY`, so an installed extension can be tied back to the exact pinned key it was signed with. It is `NULL` for unsigned and statically-linked extensions.

**Settings**

- `extension_repository_directory` — where repository definitions are stored (one JSON file per repository, including the pinned keys, so pinning survives restarts). Defaults to `~/.duckdb/extension_repositories`.
- `allow_extension_repositories` — ternary opt-in:
  - `undecided` (default): adding new repositories is blocked until the user opts in. Repositories that already exist stay trusted, so their already-installed extensions can still be listed, loaded and dropped.
  - `allowed`: repositories can be added, and extensions can be installed from them.
  - `forbidden`: additionally distrusts existing repositories, so their keys are no longer accepted and their extensions fail to verify on install and load. Listing and dropping still work.
- `lock_configuration` — creating or dropping a repository changes the set of trusted signing keys, so both statements are gated by the configuration lock, freezing the trust store entirely.

**Security model**

Registering a repository adds a trusted signing key, which is at least as consequential as re-enabling unsigned or community extensions. So `allow_extension_repositories` follows the same rule as `allow_unsigned_extensions` / `allow_community_extensions`: it can only be **enabled at startup** (through the client's configuration), and while the database is running it can only be made *more* restrictive, never less. `forbidden` is therefore a one-way ratchet within a session. This keeps `allow_unsigned_extensions=false` a real guarantee — no runtime SQL connection can escalate to `allowed` and register an attacker-controlled key.

Two further guards close the directory-based bypass:

- Installing an extension from a user-provided repository requires `allow_extension_repositories='allowed'`, so the opt-in cannot be sidestepped by pointing `extension_repository_directory` at a directory of planted repository files and installing from it.
- While signature checking is enabled (`allow_unsigned_extensions=false`), `extension_repository_directory` — the trust anchor that determines which keys are trusted — can only be set at startup, so a runtime connection cannot repoint it at attacker-controlled keys. When unsigned extensions are already allowed the signature-trust model is off, so it stays freely settable.

Together with `lock_configuration` as an additional layer, this means that under the defaults (`allow_unsigned_extensions=false`, `allow_extension_repositories=undecided`) no runtime SQL connection can expand the set of trusted signing keys.


rustyconover:
@hannes, nice to see this. A few points:

## 1. Extension name shadowing

A custom repository can publish extensions that shadow core names — `json`, `core_functions`, etc. `local_extension_path` is `<ext_dir>/<name>.duckdb_extension` (`extension_install.cpp:568`) with no repository component, so `FORCE INSTALL json FROM evil_repo` lands on the same path core `json` would use, and the `.info` records `USER_PROVIDED`/`evil_repo`.

Every later session loads it:

- Autoinstall from `core` is a silent no-op because the file exists `ExtensionInstallOptions::throw_on_origin_mismatch` defaults to `false` (`extension_helper.hpp:93`), so `ThrowErrorOnMismatchingExtensionOrigin` never fires on the autoload path. A later plain `INSTALL json` does not correct it.
- `TryInitialLoad` reads the `.info`, resolves `evil_repo.json`, and the signature verifies under the attacker's pinned key.
- `allow_extension_repositories='undecided'` the current default — explicitly keeps trusting existing repositories, so this survives into sessions that never opted in. Autoload is on by default in the CLI and Python builds.

And `FORCE INSTALL` isn't really the barrier: today, dropping a binary into the extension directory gets you nothing because it won't verify under the core keys. Once a user trusts any repository I control, I can write a validly signed binary and its `.info` directly.

**Suggestion: namespace extensions by repository.**

Install to a repository-prefixed directory so the artifact is `evil_repo/json`, loaded as `LOAD 'evil_repo/json'`. Only new user-repository installs need to change — core and community installs stay flat, so it's backwards compatible. 
The upside is that this gets DuckDB out of being the extension namespace police.

## 2. No way to tell which key signed an on-disk artifact

Multi-key repositories are the right call for rotation, but there's currently no way to answer "which extensions in `~/.duckdb` were signed by key A?" `duckdb_extension_repositories()` shows the key set; `duckdb_extensions()` shows `installed_from` (a URL) and nothing about keys. So when a repository announces a key compromise, the only way to find your blast radius is to remove the key and see what stops loading — i.e. discover it by breaking things.

`CheckKnownSignatures` (`extension_load.cpp:350`) already determines the answer and discards it by returning `bool`. Two ways to surface it, both cheap:

- Record the matching key's fingerprint in the `.info` at install and expose it as a `duckdb_extensions()` column. `ExtensionRepositoryManager::GetPublicKeyFingerprint` already exists in this PR, so it'd be the same string `CREATE EXTENSION REPOSITORY` prints for out-of-band comparison.
- A scalar function taking a filename, for auditing artifacts you didn't install yourself.

The first gives attribution for what's already on disk without re-reading every binary; the second works on arbitrary files. To be clear this is operational auditing, not a security tool.

## 3. Two config gaps

`allow_extension_repositories` can be moved `undecided`→`allowed` at runtime. `AllowCommunityExtensionsSetting::OnSet` and `AllowUnsignedExtensionsSetting::OnSet` both refuse any loosening once `info.db` is set; this one only blocks moving away from `forbidden`. Registering a new signing key seems at least as consequential as re-enabling community extensions.

Separately, `CheckCanCreate` only guards `CreateRepository`. `TryGetRepository` — which `bind_load.cpp` and `InstallFromRepository` use — has no access check, and `extension_repository_directory` is a plain runtime-settable global. So `SET extension_repository_directory` at a directory you wrote (the test in this PR uses `COPY ... TO` to write `public_key.pem`) plus `INSTALL x FROM my_repo` gets the same trust without ever setting `allowed`.

## 4. Expanding the repository model

This PR establishes `public_key.pem` as the first well-known file at a repository prefix. It is worth standardizing more of them: a table of contents of available extensions, and a manifest of available versions per extension. That would give `INSTALL ... VERSION` something real to resolve against instead of always taking latest, and would let DuckDB ship table functions that union over all known repositories for discovery.

One constraint: any TOC or version manifest has to be signed by the repository's key, or it's an unauthenticated redirect — fetching it over the same channel as the extensions just relocates the trust problem.

## 5. NPM/Pypi support

Enterprises (finance, healthcare) really just want to deal with extension repositories they already have mirrors and proxies for (primarily this is Maven, Pypi and NPM). While I cannot name the enterprises due to NDAs, many of them have to go through hoops to onboard duckdb extensions into their regulated environments. Adding support for installing from npm/pypi would be easier for them. Using npm/pypi you'd get Trusted Publishing for free.  Versioning of extensions is also taken care of.  You'd just need a bit more code to handle it.

@VGSML any thoughts?

VGSML:
Thank you for this PR — in my opinion this is a very needed and timely feature. The number of external trusted repositories in the community keeps growing, and DuckDB is increasingly used in enterprise settings, where being able to add your own trusted extension source (a middle ground between DuckDB-only keys and the global `allow_unsigned_extensions`) is genuinely valuable. The direction looks right, and partitioning verification keys by origin is a real hardening — a leak of one key set no longer verifies another origin. A few suggestions below; some overlap with @rustyconover's comment.

**1. Make enabling the trust gate one-way.**
Right now `allow_extension_repositories` can be raised `undecided → allowed` at runtime from any SQL connection (`OnSet` only blocks `forbidden → *`). That's weaker than `allow_unsigned_extensions` /`allow_community_extensions`, which cannot be relaxed at runtime. In an unlocked database this means `allow_unsigned_extensions=false` no longer guarantees the absence of untrusted native code: a connection can `SET allow_extension_repositories='allowed'` → `CREATE EXTENSION REPOSITORY evil ... USING PUBLIC KEY '<attacker>'` → `INSTALL/LOAD` → `dlopen`.
Suggestion: allow the transition to `allowed` only at startup (throw in `OnSet` while the database is running), mirroring `allow_unsigned_extensions`. `lock_configuration` already closes this today, but it would be better as a second layer than as the only line of defense.

**2. Per-repository on-disk layout + qualified LOAD.**
The install path is currently `<dir>/<duckdb_version>/<platform>/<name>.duckdb_extension` (`PathComponents() = {duckdb_version, platform}`) — with no repository component, so `json` from a different repo/community/core share one path and overwrite each other (the shadowing @rustyconover raised). Suggestion: add a repository component (the origin is already written into `.info`): `<dir>/<duckdb_version>/<platform>/<repo>/<name>.duckdb_extension`; keep `LOAD ext` as today, and add `LOAD 'repo/ext'` for explicit selection on name collisions. Multiple versions need not be kept on disk (overwrite is fine) — recording the installed version in `.info` is enough.
Separately: `INSTALL ext FROM repo VERSION '1.2.3'` is already supported by core — the grammar (`VersionNumber`) and `ExtensionUrlTemplate` build `<prefix>/${NAME}/<version>/${REVISION}/${PLATFORM}/...` — so the versioning mechanism is essentially there already; what's missing is discovery (see 3).

**3. Discovery — two levels.**
It would be useful to (a) list which extensions a repo offers, and (b) which versions exist for a given extension. A single monolithic manifest doesn't scale: worst case ~100 extensions × ~10 DuckDB versions × ~20 extension versions × ~6 platforms. So two levels:
- a **lightweight repo-level TOC** (like @rustyconover suggested, kept narrow): per extension — name, list of platforms, current version. Stays small (hundreds of rows).
- **per-extension discovery** as a separate file, only if the repo serves multiple versions: the list of versions for that extension; and/or an info file next to the binary, mirroring its URL:
  - `https://example.repo.com/v1.2.0/osx_arm64/spatial.duckdb_extension.info.json`
  - `https://example.repo.com/spatial/1.2.3/v1.2.0/osx_arm64/spatial.duckdb_extension.info.json`

It would be natural to put this under `.well-known/…`, so a repo can also declare its public key(s) (several, for rotation) and its layout, instead of a hard-coded `public_key.pem` + fixed path. The info file is also a good place for the signing-key fingerprint (matching @rustyconover's idea of exposing the fingerprint in `duckdb_extensions()`).

**4. WASM: verify the same bytes that get loaded.**
On the WASM path the signature is checked on one download (`fs.OpenFile`), while `dlopen` loads a separately re-fetched XHR copy — verify and load can diverge. Better to verify the signature on the exact buffer that gets instantiated (`CheckExtensionBufferSignature` of that buffer) and to check the result of `WebAssembly.validate()`.

All of this is additive and compatible with the off-by-default posture. Overall this PR fully covers what I proposed in #23387 (discussion #23388) — very glad to see it landing in core.

hannes:
@rustyconover @VGSML thanks for the feedback. I'll add the name spacing and sharpen the config. Regarding the info file, so far we have

- list of extensions and their versions
- list of signing keys

to be included in the `.../.well-known/duckdb-extension-repo.json` file. Or maybe a YML?

Anything else you would want to see there? Of course this can always be extended.

I'm considering all other comments but have no immediate questions yet.

VGSML:
I think the json will be enough, yml can be added in the future.

rustyconover:
@hannes, thanks for reading our feedback!

Rather than a .yaml file you could just use a .duckdb file in the extension repository root.

You have `scan_duckdb()`.  That way you can have different tables easily enough, and you'll always have the reading infrastructure file and it easier for repo maintenance. 

# C vs C++ native extension interfaces..

With the release of the C extension API, the versioning of extensions by duckdb build version likely should change.  C API extensions don't need rebuilds for each release, native C++ extensions do.  But C API extensions aren't versioned based on their API version, they're still versioned based on build version of DuckDB.

The difficulty here is knowing if an extension uses the C interface or the C++ native extension, this is something that could be expressed in the extension repository metadata file. 

# Extension updates

It would be nice if there could be automated extension updates, much like how package managers do it.  The problem with this is, there there currently isn't a way for a user to set a version pin when they install an extension.  You may want to think about writing this in the installed info as well, but to do this we'd be getting into solving version pins and version ranges.

Maxxen:
Extensions encode in their metadata if they use the C or C++ ABI, and also specify if they are based on a stable API version (i.e. a published DuckDB version) or if they are unstable (pinned to a specific DuckDB commit hash). So I don't think this is a problem.

Maxxen:
We are looking into changing extension versioning separately though - so that extensions don't just specify what version of DuckDB they target, but also provide _their own_ version number.

VGSML:
Agree this would be useful. One nuance on top of @Maxxen's ABI point: ABI stability isn't the same as SQL-API stability - an extension can rename/remove a function or change behavior between its own versions even with a stable C ABI, so a silent update can break queries that still load fine. So I'd suggest:
1) make auto-update opt-in behind a setting like `autoinstall_known_extensions`/`autoload_known_extensions`, lockable via `lock_configuration` for central on/off;
2) treat an explicit `INSTALL repo/ext VERSION '1.2.3'` as frozen and exclude it from auto-update - the version is already in `.info`, which also partly covers the "no pin at install" blocker @rustyconover mentioned. 

Given enterprise wants it both ways (security fixes vs change control) plus version ranges, this feels like a separate PR alongside the config-lock work rather than part of this one.

amluto:
RSA-2048?  Current advice seems to be that one should use quantum-resistant cryptography for everything important, and this certainly counts.  While RSA-2048 will not fall as quickly as elliptic curve schemes (simply because it’s larger), I wouldn’t count on it.

SLH-DSA would be my go-to choice: it’s based on ordinary cryptographic hashes, which show no signs of being broken now or in the future.  And it’s basically a drop-in replacement.  But really, almost any modern choice would probably be better than RSA-2048.

hannes:
I would say auto update is out of scope for this one, we don't have it for core extensions either. Later.

hannes:
@amluto - we're not changing crypto algorithms at this stage. This could happen later for all of DuckDB. I am also not worried about quantum computers coming after our keys in the immediate future.

hannes:
Slightly reworked syntax, lets keep `INSTALL x FROM y`, but for added consistency then do `LOAD x FROM y` for external repos.

hannes:
and for now I'm not adding the discovery, just the keys for the repo json. this can always be added and I'm wary of feature creep for this pr.

VGSML:
> @amluto - we're not changing crypto algorithms at this stage. This could happen later for all of DuckDB. I am also not worried about quantum computers coming after our keys in the immediate future.

+1 to leaving crypto as-is at this stage. And since this is a signature rather than encryption, there's no "harvest now, forge later" on existing signatures — the only quantum exposure would be future key recovery, so a DuckDB-wide change later (as you say) feels like the right level for this, not this PR. 

One small forward note for whenever that happens: reserving an algorithm tag in the extension footer and key format would let a future scheme be added next to RSA-2048 without a flag-day. Not asking for it here — just a cheap hook that would make that later migration smoother.

hannes:
Also pushed a change that makes the `FROM y` optional in `LOAD x FROM y` if its unambiguous

src/main/settings/custom_settings.cpp
```diff
@@ -145,6 +146,56 @@ void AllowCommunityExtensionsSetting::OnSet(SettingCallbackInfo &info, Value &in
...
+	if (!info.db) {
+		// the value is set before the database is running (startup) - any value is allowed
+		return;
+	}
+	// while the database is running the setting can only be tightened, never loosened. Trusting a new signing key is at
+	// least as consequential as re-enabling unsigned or community extensions, so - mirroring allow_unsigned_extensions
+	// and allow_community_extensions - extension repositories can only be enabled at startup. This also makes the
+	// 'forbidden' state a one-way ratchet within a session
+	auto current_access = ExtensionRepositoryManager::GetAccess(*info.db);
+	if (ExtensionRepositoryPermissiveness(new_access) > ExtensionRepositoryPermissiveness(current_access)) {
```
carlopi:
```suggestion
	if (ExtensionRepositoryPermissiveness(new_access) > ExtensionRepositoryPermissiveness(current_access) &&
           current_access != ExtensionRepositoryAccess::UNDECIDED) {
```
I think this is needed to allow UNDECIDED to move to allowed.
Or maybe it needs to be done even more complex, I think the logic should be:
```
current == UNDECIDED -> all directions are good
current == ALLOWED -> only DISABLED or ALLOWED
current == DISABLED -> only keep to DISABLED
```
so even reordering few lines up can work.

rustyconover:
Hi @hannes,

One more thing I thought about on a trip this weekend was right now you can launch duckdb with an extension named as part of the database name. i.e. 

```
duckdb "airport:grpc+tls://hello-airport.query.farm"
```

Can we add the extension repo name to that CLI launch path?

Rusty

src/main/config.cpp
```diff
@@ -250,14 +252,14 @@ static const ConfigurationOption internal_options[] = {
...
-                                                     DUCKDB_SETTING_ALIAS("wal_autocheckpoint", 29),
-                                                     DUCKDB_SETTING_ALIAS("worker_threads", 171),
+static const ConfigurationAlias setting_aliases[] = {DUCKDB_SETTING_ALIAS("configure_metrics", 31),
+                                                     DUCKDB_SETTING_ALIAS("custom_profiling_settings", 31),
+                                                     DUCKDB_SETTING_ALIAS("memory_limit", 132),
+                                                     DUCKDB_SETTING_ALIAS("null_order", 63),
+                                                     DUCKDB_SETTING_ALIAS("profile_output", 155),
+                                                     DUCKDB_SETTING_ALIAS("user", 175),
+                                                     DUCKDB_SETTING_ALIAS("wal_autocheckpoint", 30),
+                                                     DUCKDB_SETTING_ALIAS("worker_threads", 173),
```
carlopi:
```suggestion
static const ConfigurationAlias setting_aliases[] = {DUCKDB_SETTING_ALIAS("configure_metrics", 31),
                                                     DUCKDB_SETTING_ALIAS("custom_profiling_settings", 31),
                                                     DUCKDB_SETTING_ALIAS("memory_limit", 132),
                                                     DUCKDB_SETTING_ALIAS("null_order", 63),
                                                     DUCKDB_SETTING_ALIAS("profile_output", 156),
                                                     DUCKDB_SETTING_ALIAS("user", 176),
                                                     DUCKDB_SETTING_ALIAS("wal_autocheckpoint", 30),
                                                     DUCKDB_SETTING_ALIAS("worker_threads", 174),
                                                     FINAL_ALIAS};
```

carlopi (approved):
LGTM, thanks

Mytherin:
Thanks!

PR #24778, opened 2026-08-14
[ASYNC I/O] Read Ahead Refactor

pdet:
This PR has no behavior changes, it's mainly a refactor to get the multi-file read-ahead code to be able to be reused for DuckDB native table scans. We basically create a generic `ScanReadAhead<JOB, STATE>` with the read-ahead code.


lnkuiper:
I can't give a review right now because GH is down and I can't see the diff. Just wanted to ask: What's the reason for choosing templating instead of inheritance? The `MultiFileReader` used to be templated, but now uses inheritance, and it's much nicer now. Not saying templating is wrong, just want to hear the reasoning.

pdet:
I think you are right, inheritance is a much cleaner option.

src/parallel/scan_read_ahead.cpp
```diff
@@ -0,0 +1,274 @@
...
+			task->Execute(TaskExecutionMode::PROCESS_ALL);
+			task.reset();
+			continue;
+		}
+		// the remaining I/O is running on other threads, wait for it to finish
+		TaskScheduler::YieldThread();
+	}
+}
+
+ScanReadAheadJob::~ScanReadAheadJob() {
```
lnkuiper: The wait here runs in the base destructor, after the derived members have already been destroyed. `MultiFileScanJob` handles this by repeating the wait in its own destructor, but every future job type would need to know that rule. Could we have `ScanReadAhead` own a wrapper that waits before destroying the job payload instead?

src/parallel/scan_read_ahead.cpp
```diff
@@ -0,0 +1,274 @@
...
+	for (auto &task : io_tasks) {
+		executor->ScheduleTask(make_uniq<ReadAheadIOTask>(*executor, std::move(task), completion));
+	}
+	lock_guard<mutex> guard(lock);
+	if (memory_governor) {
+		memory_governor->UpdateReservation(pending_io_bytes.load());
+		backlog_budget = memory_governor->BackpressureBudget();
+	}
+	// producers push concurrently, so admit jobs to the queue in batch-index order
+	pending_jobs.emplace(job->batch_index, std::move(job));
```
lnkuiper: `batch_index` has a hidden contract here: it must be unique, gapless, and start at 0. A duplicate is silently discarded by `emplace`, while a gap leaves later jobs pending forever. The native table scan currently starts its batch index at 1, is that a problem? Could we make this explicit?

src/include/duckdb/parallel/scan_read_ahead.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+//! Outcome of ScanReadAhead::AcquireJob
+enum class ScanReadAheadAcquire : uint8_t {
+	ACQUIRED,  //! a job with settled I/O is now held
+	EXHAUSTED, //! every job has been produced and claimed, the scan is done
+	PARKED     //! the scan task parked until the claimed job's I/O completes, the caller must yield
+};
+
+//! Throw when the query was interrupted, otherwise yield the thread
+void ScanReadAheadYield(ClientContext &context);
```
lnkuiper: Could we keep these helpers out of the public header? `ScanReadAheadYield` is only used by `AcquireJob`, so I would make it file-local or inline it. `TryGetReadAheadDepth` is only used by `Create`, so it can be folded into that function or made private. `ReadAheadIOTask` above also looks like a `.cpp` implementation detail.

src/include/duckdb/parallel/scan_read_ahead.hpp
```diff
@@ -0,0 +1,210 @@
...
+	//! -1 means automatic mode, unlimited depth with the backlog bounded by a temp-memory reservation.
+	static unique_ptr<ScanReadAhead> Create(ClientContext &context);
+
+public:
+	//! Claims the next job and schedules its I/O, filling io_tasks when the I/O was detached to the pool.
+	//! Returns null when there are no more jobs to produce.
+	using ProduceJobCallback = std::function<unique_ptr<ScanReadAheadJob>(vector<unique_ptr<AsyncTask>> &io_tasks)>;
+
+	//! Try to produce one job into the queue.
+	bool TryProduceJob(const ProduceJobCallback &claim_and_schedule);
```
lnkuiper: Do we need to expose these lower-level state-machine methods? `TryProduceJob`, `IsDone`, `HasActiveProducers`, and `ClaimJob` are only called internally. We could make them private.

src/include/duckdb/parallel/scan_read_ahead.hpp
```diff
@@ -0,0 +1,210 @@
...
+	//! Check if scan is done, i.e., no more jobs to do
+	bool IsDone() const;
+	//! Whether any thread holds a reserved slot it has not pushed a job for yet
+	bool HasActiveProducers() const;
+	//! Pop the oldest queued job
+	unique_ptr<ScanReadAheadJob> ClaimJob();
+	//! Push a finished job's scan state, so learned scan state carries over to jobs created later
+	void PushState(unique_ptr<LocalTableFunctionState> state);
```
lnkuiper: Could we keep this pool in `MultiFileGlobalState` and let future consumers own typed state pools?

lnkuiper (commented):
Looks better now with inheritance. Some comments:

src/include/duckdb/common/multi_file/multi_file_states.hpp
```diff
@@ -218,36 +235,43 @@ enum class MultiFileDecodeResult : uint8_t {
...
-struct MultiFileScanJob {
+struct MultiFileScanJob : public ScanReadAheadJob {
+	MultiFileScanJob() = default;
+	MultiFileScanJob(MultiFileScanJob &&) noexcept = default;
+	MultiFileScanJob &operator=(MultiFileScanJob &&other) noexcept {
+		if (this == &other) {
+			return *this;
+		}
+		// settle the previous job first, its I/O and scan state must go before its reader is released
+		SettleIO();
```
lnkuiper: Maybe we can implement wrappers so derived classes don't have to call `SettleIO()` exactly at the right moment? It looks correct currently but looking ahead I think it would be cleaner if this is built-in.

PR #24782, opened 2026-08-14
Fix #24781: do not merge common subplans of scans that cannot serialize their identity

### LINKED ISSUE #24781: common_subplan merges scans of different attached-SQLite tables with identical column layouts, returning one table's rows for both
### What happens?

When one statement reads **two attached-SQLite tables whose column layouts are identical**, the `common_subplan` optimizer merges the two scans into one materialized subplan and the query silently returns one table's rows for both. Neither the table name nor even the **source database file** participates in the comparison — only the column layout does.

`EXPLAIN` of the failing query shows a single `SQLITE_SCAN Table: flows` materialized as a CTE, with both scalar subqueries reading `CTE_SCAN` of that same CTE index.

Observed boundaries, all on v1.5.5:

- Scalar-subquery pairs, CTE pairs, and `UNION ALL` arms are all affected; it is not counts-only — `string_agg` over one table returns the other table's row contents.
- Tables with the same column types but **different column names** return correct results; different arity never collides. Two same-shape tables **in two different attached SQLite files** DO collide (`a.flows` vs `b.flows` returns `a`'s rows for both).
- A bare unfiltered `count(*)` pair is correct (answered without a scan); adding any filter or reading any column exposes it.
- Native DuckDB tables with identical layouts are unaffected — the same query over `CREATE TABLE ... AS SELECT` copies is correct. The scan identity appears to be the missing equality dimension, so other extension scanners may share the hazard.
- `READ_ONLY` vs read-write attach makes no difference; `SET threads TO 1` makes no difference (deterministic, not a race).
- `SET disabled_optimizers='common_subplan'` restores correct results for every shape above.

Possibly related to #24624 (another `common_subplan` wrong-result, but over native tables with filter-differing branches — this one merges scans of *different tables*).

### To Reproduce

Self-contained (DuckDB creates the SQLite file itself):

```sql
INSTALL sqlite; LOAD sqlite;
ATTACH '/tmp/repro_subplan.db' AS s (TYPE sqlite);
CREATE TABLE s.flows (id TEXT, source TEXT);
CREATE TABLE s.verifiers (id TEXT, source TEXT);
INSERT INTO s.flows VALUES ('f1','file'),('f2','file');
INSERT INTO s.verifiers VALUES ('v1','file'),('v2','file'),('v3','file');

SELECT (SELECT count(*) FROM s.flows     WHERE source='file') AS f,
       (SELECT count(*) FROM s.verifiers WHERE source='file') AS v;
-- expected: f=2, v=3      actual: f=2, v=2

SELECT (SELECT string_agg(id, ',' ORDER BY id) FROM s.flows)     AS f,
       (SELECT string_agg(id, ',' ORDER BY id) FROM s.verifiers) AS v;
-- expected: f=f1,f2  v=v1,v2,v3      actual: f=f1,f2  v=f1,f2

-- both queries now return the expected results
```

### OS:

Linux x86_64 (WSL2, kernel 6.18)

### DuckDB Version:

v1.5.5 (sqlite extension f79b1db); also reproduced identically on v1.5.0 (fb7cf51)

### DuckDB Client:

Node.js, `@duckdb/node-api` 1.5.5-r.4

### Hardware:

x86_64

### Full Name:

Ariel R

### Affiliation:

n/a

I have tested with a stable release (v1.5.5)

Yes

- [x] Yes, I have

- [x] Yes, I have

arieller:
Fixes #24781 (silent wrong results: scans of different attached-SQLite tables with identical column layouts merged into one subplan).

## Root cause

`CommonSubplanOptimizer` compares subplans by serialized bytes (`PlanSignature`). For a table function **without serialization callbacks**, `LogicalGet::Serialize` falls back to writing the call `parameters` / `named_parameters` "for rebinding purposes", and `FunctionSerializer::Serialize` writes only the function name and argument types. A **parameter-less** scan — e.g. one created through an attached catalog via `TableCatalogEntry::GetScanFunction` — therefore serializes nothing that identifies *what* it scans: the file and table name live solely in the bind data. Two such scans over same-shape tables produce identical bytes and are merged, and the query silently returns one table's rows for both.

The gate that was supposed to catch this, `LogicalGet::SupportSerialization()`, answers a different question: its no-callback arm (`function.bind && bind_data->SupportStatementCache()`) is a *rebinding* contract, and `FunctionData::SupportStatementCache()` defaults to `true` — so "serializable enough to rebind" was being used as "serialized bytes capture identity", which does not hold when there are no parameters to rebind from.

This also explains the observed boundaries in #24781: column *names* discriminate (they are serialized), the source *file* does not (bind-data-only), direct `sqlite_scan('file','table')` calls are unaffected (parameters serialize), and native tables are unaffected (serialization callbacks present).

## Fix

`OperatorIsSupported` refuses merge-eligibility for a `LogicalGet` that carries bind data, has no serialization callbacks, and has no parameters — the exact case where equal bytes cannot prove equal scans. Strictly conservative: it only declines an optimization, and in-tree functions (which carry serialization callbacks) keep merging — verified that e.g. two `duckdb_tables()` subqueries still merge, while two attached-SQLite scans no longer do.

## Testing

- Built against `duckdb-sqlite` (the extension statically linked): the issue's repro flips from `f=2, v=2` / both subqueries returning the flows' ids to the correct `f=2, v=3` / per-table ids, and `EXPLAIN` shows two distinct `SQLITE_SCAN`s instead of one scan behind a shared materialized CTE.
- Full fast suite on the patched build: **all tests passed, 242,686 assertions in 4,074 test cases** (skips only for absent extensions/env).
- `test/optimizer/pullup_filters.test`, `test/optimizer/projection_pullup.test`, `test/sql/optimizer/test_common_subplan_cte_binding_order.test`, `test/sql/optimizer/test_common_subplan_projection_maps.test` pass on this branch's build.
- The sqlite extension's `test/sql/storage/*` (906 assertions) and `test/sql/scanner/*` (1,908 assertions) suites pass against the patched core.
- A sqllogictest covering the collision shapes (scalar-subquery pair, `string_agg` contents, `UNION ALL` arms) is proposed to `duckdb-sqlite`, since the trigger needs an extension catalog: duckdb/duckdb-sqlite — companion PR referenced there.

An alternative, less conservative fix would be to serialize the catalog identity (`GetTable()`) in the no-callback fallback of `LogicalGet::Serialize`, which would preserve merging for genuinely-identical catalog scans — I avoided it here because it changes the serialized plan format. Happy to rework in that direction if preferred.

### DIFF
--- a/src/optimizer/common_subplan_optimizer.cpp
+++ b/src/optimizer/common_subplan_optimizer.cpp
@@ -568,7 +568,6 @@ private:
 		case LogicalOperatorType::LOGICAL_TOP_N:
 		case LogicalOperatorType::LOGICAL_DISTINCT:
 		case LogicalOperatorType::LOGICAL_PIVOT:
-		case LogicalOperatorType::LOGICAL_GET:
 		case LogicalOperatorType::LOGICAL_EXPRESSION_GET:
 		case LogicalOperatorType::LOGICAL_DUMMY_SCAN:
 		case LogicalOperatorType::LOGICAL_COMPARISON_JOIN:
@@ -580,6 +579,18 @@ private:
 		case LogicalOperatorType::LOGICAL_EXCEPT:
 		case LogicalOperatorType::LOGICAL_INTERSECT:
 			return true;
+		case LogicalOperatorType::LOGICAL_GET: {
+			auto &get = op.Cast<LogicalGet>();
+			if (get.bind_data && !get.function.HasSerializationCallbacks() && get.parameters.empty() &&
+			    get.named_parameters.empty()) {
+				// Without serialization callbacks, the serialized form carries only the call parameters
+				// (see LogicalGet::Serialize). A parameter-less scan - e.g., one created through an
+				// attached catalog - keeps its identity solely in the bind data, so equal serialized
+				// bytes cannot prove that two scans read the same table.
+				return false;
+			}
+			return true;
+		}
 		case LogicalOperatorType::LOGICAL_CHUNK_GET:
 			// Avoid serializing massive amounts of data (this is here because of the "Test TPCH arrow roundtrip" test)
 			return op.Cast<LogicalColumnDataGet>().collection->Count() < 1000;


arieller:
Closing — I noticed CONTRIBUTING.md's policy on AI-assisted PRs after opening this (the patch was developed with LLM assistance). Moved the proposed fix and test results to #24781 so it can be picked up from there without the review burden.

Issue #24615, opened 2026-08-08
`INTERNAL Error: Operator occurrence N was reconstructed more than once` on `INNER JOIN` with a single-sided `ON` predicate


wanteatfruit:
### What happens?

I got the following error message from the repro script, where the same script works fine on 1.5.5.
```
INTERNAL Error: Operator occurrence 2 was reconstructed more than once
```
Not sure if its duplicated with issue #24604, but got a cleaner repro

### To Reproduce

```sql
CREATE TABLE a AS SELECT i AS x FROM range(8)  s(i);
CREATE TABLE b AS SELECT i AS x FROM range(8)  s(i);
CREATE TABLE c AS SELECT i AS x FROM range(16) s(i);
CREATE TABLE d AS SELECT i AS x FROM range(8)  s(i);

SELECT 1 FROM a, b, c INNER JOIN d ON (c.x > 0);
```

### OS:

macOS

### DuckDB Version:

v2.0.0-alpha37185 (Cyanoptera) e500d77864

### DuckDB Client:

CLI

### Full Name:

Junwen An

### Affiliation:

NUS

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

Issue #24604, opened 2026-08-08
Join-order plan reconstruction (GenerateJoins) fires INTERNAL errors: "Cost predicate consumed before operator occurrence" and "Operator occurrence reconstructed more than once"


xmzyshypnc:
### What happens?

`QueryGraphManager::GenerateJoins` in `src/optimizer/join_order/query_graph_manager.cpp` rebuilds the plan chosen by the join-order dynamic program back into a logical operator tree. During that walk it enforces two invariants on the plan it was handed, and both can be violated, each surfacing as an `InternalException`:

- `src/optimizer/join_order/query_graph_manager.cpp:624` — `reconstructed_operators.insert(descriptor->index)` fails, so the same operator occurrence is reconstructed twice:
  ```
  INTERNAL Error: Operator occurrence %llu was reconstructed more than once
  ```
- `src/optimizer/join_order/query_graph_manager.cpp:630` — an operator's `costing_predicate_indices` points at a filter whose `filter` is already null, i.e. a cost predicate was consumed by an earlier operator occurrence:
  ```
  INTERNAL Error: Cost predicate %llu was consumed before operator occurrence %llu
  ```

Both are consistency assertions on the reconstruction, so reaching either means the dynamic program produced a plan in which an operator occurrence (or its costing predicate) is assigned inconsistently — the same occurrence reachable from two positions in the rebuilt join tree, or a predicate owned by two occurrences. They are not user errors; the query is well-formed and the plan should reconstruct cleanly. The neighbouring throws in the same function (lines 191, 472, 481, 620, 646) guard the same reconstruction, which is why a single ill-formed plan can trip whichever check it reaches first.

Both are reached by queries that join windowed sub-queries under a mix of join types. The estimate drives the plan, so the failure depends on table statistics: stripping columns or rows changes the cardinality estimate, the plan changes, and the assertion no longer fires. The full standard schema is therefore kept in the reproduction below.

### To Reproduce

**A — `Cost predicate ... was consumed before operator occurrence` (self-contained, one in-memory session, fires 3/3):**

```sql
CREATE TABLE metrics (category VARCHAR, value DOUBLE, ts BIGINT, label VARCHAR, flag BOOLEAN);
INSERT INTO metrics VALUES ('a',1.0,1,'one',TRUE),('b',2.5,2,'two',FALSE);
CREATE TABLE documents (id VARCHAR PRIMARY KEY, tenant VARCHAR NOT NULL, secret VARCHAR NOT NULL);
[...]

PR #24784, opened 2026-08-15
Fix unused residual collection consuming must-apply cost predicates

### LINKED ISSUE #24690: INTERNAL Error: Cost predicate 2 was consumed before operator occurrence 1 on LEFT JOIN + RIGHT JOIN + cross product
### What happens?

On `main`, a four-relation query that combines

* a `LEFT JOIN`,
* a `RIGHT JOIN` whose `ON` clause is an equality **plus** a single-sided filter on its preserved side,
* a cross product whose only connection to the rest of the query is a `WHERE` predicate,

fails during join-order optimization with an internal error:

```
INTERNAL Error: Cost predicate 2 was consumed before operator occurrence 1
```

The failure happens at plan time, not at execution time: `EXPLAIN` on the same query fails
identically, and the tables can be completely empty — no data is needed to trigger it.

This is the `costing_predicate_indices` single-consumption check in
`QueryGraphManager::GenerateJoins`
(`src/optimizer/join_order/query_graph_manager.cpp:630` at `e500d77864`), a sibling of the
duplicate-reconstruction check six lines above it at `:624`.

### To Reproduce

```sql
CREATE TABLE e1(x INTEGER);
CREATE TABLE e2(x INTEGER);
CREATE TABLE e3(x INTEGER);
CREATE TABLE e4(x INTEGER);

SELECT count(*)
FROM e1 LEFT JOIN e2 ON (e1.x = e2.x)
        RIGHT JOIN e3 ON (e1.x = e3.x AND e3.x > 0), e4
WHERE e4.x = e1.x;
```

```
INTERNAL Error: Cost predicate 2 was consumed before operator occurrence 1
```

`EXPLAIN SELECT ...` on the same statement fails with the same error.

## Expected behavior

The query should be planned and return `0` for the empty tables above.

With populated tables the expected answer is `1`:

```sql
CREATE TABLE t1 AS SELECT i AS x FROM range(4) s(i);
CREATE TABLE t2 AS SELECT i AS x FROM range(9) s(i);
CREATE TABLE t3 AS SELECT i AS x FROM range(7) s(i);
CREATE TABLE t4 AS SELECT i AS x FROM range(3) s(i);

SELECT count(*)
FROM t1 LEFT JOIN t2 ON (t1.x = t2.x)
        RIGHT JOIN t3 ON (t1.x = t3.x AND t3.x > 1), t4
WHERE t4.x = t1.x;
-- expected 1, actual: INTERNAL Error
```

(`1` is what the query returns with the workaround below, and it matches a manual evaluation:
the `RIGHT JOIN` preserves all 7 rows of `t3`, of which 2 have a non-NULL `t1.x` — namely
`t1.x` ∈ {2, 3}; after `, t4` and `t4.x = t1.x` only `t1.x = 2` survives.)

## Workaround

Disabling the join-order optimizer avoids the error:

```sql
SET disabled_optimizers = 'join_order';
```

## What narrows it down

Each of these single relaxations makes the query plan successfully, so all of them appear to be
required to trigger the error:

| Variant | Result |
| --- | --- |
| repro as above | **INTERNAL Error** |
| drop `, e4` (no cross product) | ok |
| drop the `WHERE` clause | ok |
| `WHERE e4.x > 0` (cross-product relation not correlated with `e1`) | ok |
| `ON (e1.x = e3.x)` (drop the single-sided `AND e3.x > 0`) | ok |
| `ON (e3.x > 0)` (drop the equality) | ok |
| first join `LEFT JOIN` → `INNER JOIN` | ok |
| second join `RIGHT JOIN` → `LEFT JOIN` | ok |
| single-sided term on the null-producing side instead: `AND e1.x > 0` | ok |

These variants still fail, i.e. they are not part of the trigger:

| Variant | Result |
| --- | --- |
| `, e4` written as `CROSS JOIN e4` | INTERNAL Error |
| `WHERE e4.x < e1.x` instead of `=` | INTERNAL Error |
| `e4` listed first: `FROM e4, e1 LEFT JOIN e2 ...` | INTERNAL Error |
| single-sided term is `AND e3.x IS NOT NULL` instead of a comparison | INTERNAL Error |
| two-column tables, first join equating the other column (`ON (e1.y = e2.y)`) | INTERNAL Error |
| `SELECT 1 ... LIMIT 1` instead of `count(*)` | INTERNAL Error |
| empty tables, 2-row tables, and 4/9/7/3-row tables | INTERNAL Error (cardinalities are irrelevant) |

All variants in both tables were checked against a clean build of `e500d77864`.

So the shape needed is: four relations; a `LEFT JOIN` below a `RIGHT JOIN`; the `RIGHT JOIN`'s `ON`
carrying both an equality and a single-sided predicate on its **preserved** side; and a
cross-product relation joined to the `RIGHT JOIN`'s **null-producing** side only through `WHERE`.
Removing any one relation (three relations total) makes it pass.

## Relationship to #24615 (and possibly #24604)

This looks like a distinct failure from #24615, in the same bookkeeping:

* On the same commit, #24615's repro (`SELECT 1 FROM a, b, c INNER JOIN d ON (c.x > 0);`) raises
  `Operator occurrence 2 was reconstructed more than once` (the check at `:624`), while the query
  above raises `Cost predicate 2 was consumed before operator occurrence 1` (the check at `:630`).
* Locally relaxing or suppressing the `:624` check — three different variants, all of which make
  #24615's repro pass — leaves the query above failing at `:630` unchanged. Restricting the
  conflict detector in `join_order_operator.cpp` instead does not help either.

So a fix for #24615 does not appear to cover this case.

### OS:

Linux x86_64 (kernel 5.10)

### DuckDB Version:

built from source at `e500d77864fe565f90e68f06d729c25b11e775c5` (`main`);   `version()` = `v1.6.0-dev12367`, `source_id` = `e500d77864`

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

kenryhuang

### Affiliation:

Cloud

- [x] Yes, I have

- [x] Yes, I have

Yes

gulugulubing:
fixes: #24690

The unused-residual collection in `GenerateJoins` also collected residual predicates that belong
to a must-apply operator occurrence (its cost predicates), so the occurrence could no longer
consume them during reconstruction, failing with `INTERNAL Error: Cost predicate 2 was consumed
before operator occurrence 1`.

The collection now skips predicates registered in `operator_costing_predicates`.

Test: new `test/sql/join/test_must_apply_residual_predicate.test` with the issue's repro, the
variants from the issue that trigger the same error (non-comparison single-sided term, correlated
relation listed first, non-equality correlation), and the populated-data case (expected 1).

### DIFF
--- a/src/optimizer/join_order/query_graph_manager.cpp
+++ b/src/optimizer/join_order/query_graph_manager.cpp
@@ -716,11 +716,14 @@ GenerateJoinRelation QueryGraphManager::GenerateJoins(vector<unique_ptr<LogicalO
 	result_operator->has_estimated_cardinality = true;
 
 	// Collect unused residual predicates that belong to this join. Semantic non-inner joins own their complete
-	// condition directly, so only ordinary (notably INNER) residual predicates reach this path.
+	// condition directly, so only ordinary (notably INNER) residual predicates reach this path. Predicates that
+	// belong to a must-apply operator occurrence are consumed by that occurrence during reconstruction, instead
+	// of being added to unused_residual_predicates.
 	vector<unique_ptr<Expression>> unused_residual_predicates;
 	for (auto &filter_info : filters_and_bindings) {
 		if (filter_info->from_residual_predicate && filters_and_bindings[filter_info->filter_index]->filter &&
-		    filter_info->set.get().count > 0 && JoinRelationSet::IsSubset(*result_relation, filter_info->set)) {
+		    !operator_costing_predicates.count(filter_info->filter_index) && filter_info->set.get().count > 0 &&
+		    JoinRelationSet::IsSubset(*result_relation, filter_info->set)) {
 			unused_residual_predicates.push_back(std::move(filters_and_bindings[filter_info->filter_index]->filter));
 		}
 	}
--- a/test/optimizer/joins/test_must_apply_residual_predicate.test
+++ b/test/optimizer/joins/test_must_apply_residual_predicate.test
@@ -1,0 +1,67 @@
+# name: test/optimizer/joins/test_must_apply_residual_predicate.test
+# description: Residual predicate belonging to a must-apply join operator must not be consumed as an unused residual (issue 24690)
+# group: [joins]
+
+statement ok
+CREATE TABLE e1(x INTEGER)
+
+statement ok
+CREATE TABLE e2(x INTEGER)
+
+statement ok
+CREATE TABLE e3(x INTEGER)
+
+statement ok
+CREATE TABLE e4(x INTEGER)
+
+query I
+SELECT count(*)
+FROM e1 LEFT JOIN e2 ON (e1.x = e2.x)
+        RIGHT JOIN e3 ON (e1.x = e3.x AND e3.x > 0), e4
+WHERE e4.x = e1.x
+----
+0
+
+query I
+SELECT count(*)
+FROM e1 LEFT JOIN e2 ON (e1.x = e2.x)
+        RIGHT JOIN e3 ON (e1.x = e3.x AND e3.x IS NOT NULL), e4
+WHERE e4.x = e1.x
+----
+0
+
+query I
+SELECT count(*)
+FROM e4, e1 LEFT JOIN e2 ON (e1.x = e2.x)
+        RIGHT JOIN e3 ON (e1.x = e3.x AND e3.x > 0)
+WHERE e4.x = e1.x
+----
+0
+
+query I
+SELECT count(*)
+FROM e1 LEFT JOIN e2 ON (e1.x = e2.x)
+        RIGHT JOIN e3 ON (e1.x = e3.x AND e3.x > 0), e4
+WHERE e4.x < e1.x
+----
+0
+
+statement ok
+CREATE TABLE t1 AS SELECT i AS x FROM range(4) s(i)
+
+statement ok
+CREATE TABLE t2 AS SELECT i AS x FROM range(9) s(i)
+
+statement ok
+CREATE TABLE t3 AS SELECT i AS x FROM range(7) s(i)
+
+statement ok
+CREATE TABLE t4 AS SELECT i AS x FROM range(3) s(i)
+
+query I
+SELECT count(*)
+FROM t1 LEFT JOIN t2 ON (t1.x = t2.x)
+        RIGHT JOIN t3 ON (t1.x = t3.x AND t3.x > 1), t4
+WHERE t4.x = t1.x
+----
+1


gulugulubing:
Hi @IrisphoraChen, just a heads-up that I've opened a PR for this issue (you mentioned you had a fix in progress). The patch is tiny, but the join-order internals took me a while to wrap my head around — your review would be very welcome.

RagamuffinCat (commented):
I verified this on top of clean `main` (`f8e1c96a`) and it does fix #24690 — and also Case A of #24604, which reports the same error from a much larger query. The guard matches the one the pushdown loop right below already has, and the one in `Reconstruct()`.

Two notes.

- Every other join-order optimizer test lives in `test/optimizer/joins/` with `# group: [joins]` (e.g. `required_cross_product_activation.test`, `no_cross_product_reordering.test`). `test/sql/join/` is for join *execution* behaviour. Might be worth moving so this ends up in the same `[joins]` group as the rest of the reordering tests.

- There is a second, independent hole in the same function, which this PR does not cover. Applying your patch alone to clean `main`, these still fail:

```sql
CREATE TABLE a AS SELECT i AS x FROM range(8)  s(i);
CREATE TABLE b AS SELECT i AS x FROM range(8)  s(i);
CREATE TABLE c AS SELECT i AS x FROM range(16) s(i);
CREATE TABLE d AS SELECT i AS x FROM range(8)  s(i);

SELECT count(*) FROM a, b, c INNER JOIN d ON (c.x > 0);
-- INTERNAL Error: Operator occurrence 2 was reconstructed more than once
```

```sql
CREATE TABLE t0 AS SELECT i AS x, i%3 AS y FROM range(8)  s(i);
CREATE TABLE t1 AS SELECT i AS x, i%4 AS y FROM range(8)  s(i);
CREATE TABLE t2 AS SELECT i AS x, i%5 AS y FROM range(16) s(i);
CREATE TABLE t3 AS SELECT i AS x, i%2 AS y FROM range(8)  s(i);
CREATE TABLE t4 AS SELECT i AS x, i%7 AS y FROM range(4)  s(i);

SELECT count(*) FROM t0, t1 CROSS JOIN t3, t4 CROSS JOIN t2;
-- INTERNAL Error: Operator occurrence 3 was reconstructed more than once
```

That is #24615, and Case B of #24604. The cause is unrelated to residual predicates: for ordinary occurrences "completed by a relation set" is set containment, and since the two children of a DP node are disjoint the first-completion node is unique — but `IsCompletedBy` takes a different branch for `CROSS_PRODUCT` that uses intersection with both inputs, and two disjoint sets can each intersect both sides. So one cross-product occurrence legitimately gets realized in two sibling subtrees, and the unconditional uniqueness assertion in `GenerateJoins` rejects a plan that is actually correct.

gulugulubing:
> * Every other join-order optimizer test lives in `test/optimizer/joins/` with `# group: [joins]` (e.g. `required_cross_product_activation.test`, `no_cross_product_reordering.test`). `test/sql/join/` is for join _execution_ behaviour. Might be worth moving so this ends up in the same `[joins]` group as the rest of the reordering tests.

Thanks for the thorough review. I've moved the test to test/optimizer/joins/ with the [joins] group.

> That is #24615, and Case B of #24604. The cause is unrelated to residual predicates: for ordinary occurrences "completed by a relation set" is set containment, and since the two children of a DP node are disjoint the first-completion node is unique — but `IsCompletedBy` takes a different branch for `CROSS_PRODUCT` that uses intersection with both inputs, and two disjoint sets can each intersect both sides. So one cross-product occurrence legitimately gets realized in two sibling subtrees, and the unconditional uniqueness assertion in `GenerateJoins` rejects a plan that is actually correct.

Your analysis of the cross-product `IsCompletedBy` issue makes sense — the uniqueness assertion in GenerateJoins does look like it rejects a valid plan there. It's a distinct root cause, so I've kept this PR focused on the residual-predicate fix.

Mytherin:
Thanks!

PR #10413, opened 2024-01-31
Fix results for anti joins on empty tables

Tmonster:
Fixes https://github.com/duckdb/duckdb/issues/10406

In the statistics propagator, anti joins (and sometimes semi joins) get converted to cross products where the right child has `Limit = 1` if the filter is always false/true. 

The problem arises when the right table is empty, because then the cross product produces an empty result. For semi joins this is fine, but for anti joins we should just be propagating the left side.

Only happens when a table is anti joined on an empty table and the type of the column is numeric.

PR #24788, opened 2026-08-15
Fold trivial any-joins back into cross products

Maxxen:
Follow-up to https://github.com/duckdb/duckdb/pull/24600

This PR breaks out the statistic propagation "always true/always false" logic from the comparison join case, and re-uses it for the any-join case too. This makes e.g. the `BLOCKWISE_NL_JOIN: true` produced from the optimization in #24600 fold into a `CROSS_PRODUCT`. We also get the other "always false" -> empty result optimizations for free.

Simplifying into a cross product turns out to actually be a worthwhile optimization:

```
  128 users × 16,384 products, APPROX NEAREST 5 BY SIMILARITY, threads=4, 3 rounds avg:
  ┌─────────────────────────────────────┬─────────┬─────────┐
  │                                     │ 256-dim │ 16-dim  │
  ├─────────────────────────────────────┼─────────┼─────────┤
  │ Baseline: evalute IS NOT NULL       │ 0.851 s │ 0.045 s │
  ├─────────────────────────────────────┼─────────┼─────────┤
  │ PR #24600: fold to Condition: true  │ 0.529 s │ 0.034 s │
  ├─────────────────────────────────────┼─────────┼─────────┤
  │ This PR: fold to (cross product)    │ 0.338 s │ 0.021 s │
  └─────────────────────────────────────┴─────────┴─────────┘
  ```
    
 This PR also removes the `ANTI`/`RIGHT_ANTI` always-true rule. As far as I understand an anti-join against an empty right side must return every left row, but the rule mapped it to an empty result. So this is a pre-existing wrong-results bug in the comparison-join path that was never exercised. I've added tests for it in test_join_always_true_conditions.test, but would love a sanity-check from @lnkuiper

### DIFF
--- a/src/include/duckdb/optimizer/statistics_propagator.hpp
+++ b/src/include/duckdb/optimizer/statistics_propagator.hpp
@@ -85,8 +85,14 @@ private:
 	void UpdateFilterStatistics(const Expression &condition);
 	//! Set the statistics of a specific column binding to not contain null values
 	void SetStatisticsNotNull(ColumnBinding binding);
+	//! Propagate a filter condition and determine whether it can be pruned; does not update any statistics
+	FilterPropagateResult ClassifyFilter(unique_ptr<Expression> &condition);
 	//! Propagate a filter condition
 	FilterPropagateResult HandleFilter(unique_ptr<Expression> &condition);
+	//! Rewrite a join whose condition can never match; returns true if the operator was replaced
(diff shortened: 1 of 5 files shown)


Maxxen:
Had the clanker dig into it some more, and seems like the ANTI-join case was removed in https://github.com/duckdb/duckdb/pull/9654, but then added back again here https://github.com/duckdb/duckdb/pull/10413 😅, I guess by accident?

Mytherin:
Thanks - I think removing that optimization case makes sense to me.

PR #24790, opened 2026-08-16
Fix Parser::Tokenize by emitting trailing token past input end

### LINKED ISSUE #24742: Parser::Tokenize emits a token starting one past the end of the input
### What happens?

`Parser::Tokenize` returns a trailing `SimplifiedToken` whose `start` is one past the last character of the input. There is no character at that offset, so the token cannot describe any text.

On an empty string it returns one token at offset 0 for a zero-length input.

This is new in 2.0.0; 1.5.x did not emit it.

### To Reproduce

Linking straight against libduckdb, no extensions:

```cpp
#include "duckdb.hpp"
#include "duckdb/parser/parser.hpp"
#include <cstdio>

int main() {
  for (std::string sql : {"select 1", "select 1;", "create table zebra (bar varchar);", "", "  "}) {
    auto tokens = duckdb::Parser::Tokenize(sql);
    printf("len=%-3zu tokens=%zu  start/type=[", sql.size(), tokens.size());
    for (size_t i = 0; i < tokens.size(); i++) {
      printf("%s%llu/%d", i ? "," : "", (unsigned long long)tokens[i].start, (int)tokens[i].type);
    }
    printf("]\n");
  }
}
```

Output on `v2.0.0-alpha37626` (`7e14bd24e0`):

```
input "select 1"                           len=8   tokens=3  start/type=[0/4,7/1,8/0]
input "select 1;"                          len=9   tokens=4  start/type=[0/4,7/1,8/3,9/0]
input "create table zebra (bar varchar);"  len=33  tokens=9  start/type=[0/4,7/4,13/0,19/3,20/0,24/4,31/3,32/3,33/0]
input ""                                   len=0   tokens=1  start/type=[0/0]
input "  "                                 len=2   tokens=1  start/type=[2/0]
```

In every case the last token starts at `sql.size()`, and its type is `SIMPLIFIED_TOKEN_IDENTIFIER`.

The same inputs through the `ui` extension's `/ddb/tokenize`, which is a thin wrapper over `Parser::Tokenize`, on 1.5.5 versus 2.0.0:

| input | len | 1.5.5 offsets | 2.0.0 offsets |
| --- | --- | --- | --- |
| `select 1` | 8 | `[0,7]` | `[0,7,8]` |
| `select 1;` | 9 | `[0,7,8]` | `[0,7,8,9]` |
| `create table zebra (bar varchar);` | 33 | 8 tokens, last at 32 | 9 tokens, last at 33 |

### Why it matters

Consumers use these offsets to slice the input — for syntax highlighting, and for splitting a script into statements at `;` tokens. A token that begins past the end reopens a region after the last real one has closed.

Concretely, in the DuckDB UI this makes a notebook cell execute **twice**. `create table foo (bar varchar);` creates the table and then reports `Catalog Error: Table with name "foo" already exists!`, because the phantom causes the app to send both the split statement and the whole cell:

```
/ddb/run  sql=create table zebra (bar varchar)
/ddb/run  sql=create table zebra (bar varchar);
```

Deterministic on 2.0.0, never on 1.5.5, with the same browser driver and a fresh profile and database each run.

### Expected behaviour

Either no token at `start == input.size()`, or, if a terminator token is intentional, something that consumers can identify as not spanning any characters — the current one is indistinguishable from a real identifier.

### OS

macOS 15 (arm64), also reproduced on linux_amd64

### DuckDB Version

v2.0.0-alpha37626 (7e14bd24e0)

### DuckDB Client

C++ / libduckdb

### Full Name

Steve Shreeve

### Affiliation

Independent

AdityaM06:
Fixes #24742

### The problem
`Parser::Tokenize` converts tokens produced by `HighlightTokenizer` without skipping the internal sentinel tokens:
- `TokenType::END_OF_INPUT`
- `TokenType::END_OF_INPUT_AUTOCOMPLETE`

These tokens are appended by `BaseTokenizer::TokenizeInput()`. 

Since these sentinel tokens aren't explicitly handled by the switch statement, they fall to the default case, and are emitted as `SIMPLIFIED_TOKEN_IDENTIFIER` tokens with `start == query.size()`.

### The solution
Just skip these sentinel tokens at the beginning of the for loop in `Parser::Tokenize`

### Adding a unit test
Added a unit test in `test/common/test_parse_expression.cpp` verifying no token start positions are greater than or equal to the input length. This is tested across regular queries, trailing semicolons, empty strings, and whitespace.

### DIFF
--- a/test/common/test_parse_expression.cpp
+++ b/test/common/test_parse_expression.cpp
@@ -99,6 +99,16 @@ TEST_CASE("Parser::Tokenize does not emit trailing token beyond input", "[parse_
 	REQUIRE(tokens[1].start == 7);
 	REQUIRE(tokens[1].type == SimplifiedTokenType::SIMPLIFIED_TOKEN_NUMERIC_CONSTANT);
 
+	// "select 1;" has 9 chars -> 3 tokens: "select" (start index 0), "1" (start index 7), ";" (start index 8)
+	tokens = Parser::Tokenize("select 1;");
+	REQUIRE(tokens.size() == 3);
+	REQUIRE(tokens[0].start == 0);
+	REQUIRE(tokens[0].type == SimplifiedTokenType::SIMPLIFIED_TOKEN_KEYWORD);
+	REQUIRE(tokens[1].start == 7);
+	REQUIRE(tokens[1].type == SimplifiedTokenType::SIMPLIFIED_TOKEN_NUMERIC_CONSTANT);
+	REQUIRE(tokens[2].start == 8);
+	REQUIRE(tokens[2].type == SimplifiedTokenType::SIMPLIFIED_TOKEN_OPERATOR);
+
 	// empty input produces 0 tokens
 	tokens = Parser::Tokenize("");
 	REQUIRE(tokens.empty());
--- a/test/common/test_parse_expression.cpp
+++ b/test/common/test_parse_expression.cpp
@@ -82,3 +82,28 @@ TEST_CASE("Parse Expression List rejects invalid clauses", "[parse_expression]")
 	REQUIRE_THROWS_AS(Parser::ParseExpressionList("first(x) AS x ORDER BY x"), ParserException);
 	REQUIRE_THROWS_AS(Parser::ParseExpressionList("x LIMIT 1"), ParserException);
 }
+
+TEST_CASE("Parser::Tokenize does not emit trailing token beyond input", "[parse_expression]") {
+	std::vector<string> test_queries = {"select 1", "select 1;", "create table zebra (bar varchar);", "", "  "};
+	for (const auto &sql : test_queries) {
+		auto tokens = Parser::Tokenize(sql);
+		for (const auto &token : tokens) {
+			REQUIRE(token.start < sql.size());
+		}
+	}
+	// "select 1" has 8 chars -> 2 tokens: "select" (start index 0) and "1" (start index 7)
+	auto tokens = Parser::Tokenize("select 1");
+	REQUIRE(tokens.size() == 2);
+	REQUIRE(tokens[0].start == 0);
+	REQUIRE(tokens[0].type == SimplifiedTokenType::SIMPLIFIED_TOKEN_KEYWORD);
+	REQUIRE(tokens[1].start == 7);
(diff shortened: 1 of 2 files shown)


test/common/test_parse_expression.cpp
```diff
@@ -82,3 +82,28 @@ TEST_CASE("Parse Expression List rejects invalid clauses", "[parse_expression]")
...
+		}
+	}
+	// "select 1" has 8 chars -> 2 tokens: "select" (start index 0) and "1" (start index 7)
+	auto tokens = Parser::Tokenize("select 1");
+	REQUIRE(tokens.size() == 2);
+	REQUIRE(tokens[0].start == 0);
+	REQUIRE(tokens[0].type == SimplifiedTokenType::SIMPLIFIED_TOKEN_KEYWORD);
+	REQUIRE(tokens[1].start == 7);
+	REQUIRE(tokens[1].type == SimplifiedTokenType::SIMPLIFIED_TOKEN_NUMERIC_CONSTANT);
+
```
Dtenwolde, line 101: Just to be sure, could you add an extra check here that has the `;`?
```
  tokens = Parser::Tokenize("select 1;");
  REQUIRE(tokens.size() == 3);
  REQUIRE(tokens[2].start == 8);
  REQUIRE(tokens[2].type == SimplifiedTokenType::SIMPLIFIED_TOKEN_OPERATOR);
```
I suspect it will already pass, but just in case :)

Dtenwolde (commented):
Looks good, thanks!

Only comment is to add a single test to double-check

test/common/test_parse_expression.cpp
AdityaM06: Added the test case for a query ending with a semicolon in my latest commit (and rebased onto main). Thanks for the review!

Dtenwolde (approved):

Tishj:
@carlopi this relates to https://github.com/duckdblabs/duckdb-internal/issues/10486, I believe the direction we ended on was explicitly not doing this, no?
Because omitting (title is wrong, it says *emitting*) these tokens would limit the ability for an extension to determine whether there are following tokens, which is a regression on what it currently can do

Although.. now that I read this, it seems that it *couldn't* determine that these tokens were the end-of-input sentinels, so it's not a regression - only if existing extensions assume that the last token received *is* the end-of-input sentinel.

PR #22079, opened 2026-04-15
Add shell extension hooks for registering custom dot-commands

rustyconover:
## Motivation

I spent some time adding a .ask mode to DuckDB that integrates with an LLM agent demo:

https://github.com/user-attachments/assets/e17a8dca-1dc4-4a66-b5c3-3d22fae2098d

To make that work well, I needed changes to the shell's APIs that go beyond what a SQL function integration can provide — as @Mytherin noted in [#21197](https://github.com/duckdb/duckdb/pull/21197#issuecomment-4242225953).

This PR is a prototype for shell extension infrastructure: a way for loadable extensions to register their own .dot commands with full shell integration (help, tab completion, error suggestions, REPL sub-loops). The `.ask` use case drove the design, but the same primitives would unlock things like a reporting extension backed by Typst — or anything else that wants interactive shell UI without requiring core changes.

I'm not proposing this merges as-is, but I wanted to surface the API surface I think is necessary for a good UX before the design hardens.Related prior work with @carlopi: [#21197](https://github.com/duckdb/duckdb/pull/21197) [#21201](https://github.com/duckdb/duckdb/pull/21201) [#21041](https://github.com/duckdb/duckdb/pull/21041)

## Changes

- Adds infrastructure allowing DuckDB extensions to register their own dot-commands in the CLI shell, following the existing `ExtensionCallbackManager` pattern used by `ParserExtension`, `OptimizerExtension`, etc.
- Defines `ShellContext` (abstract shell interface), `ShellExtensionCommand` (command descriptor), and `ShellExtensionRegistration` (registered via callback manager) in a new core header
- Shell dispatch (`DoMetaCommand`, `FindMetadataCommand`, `PrintHelp`, `GetMetadataCompletions`) queries the callback registry so extension commands appear in help, tab-completion, and error suggestions
- Includes linenoise additions (`History::Clear`, `History::LoadRaw`, `Terminal::ForceSubmitOnEnter`, callback getters) needed for extensions that run custom REPL sub-loops

## State of things right now. 

There's currently no way for a DuckDB extension to add a dot-command to the CLI shell. The metadata command table is a hardcoded static array, and extensions can't interact with the shell's UI layer. This blocks use cases like AI assistants, data visualization tools, or interactive workflows that want to ship as loadable extensions rather than requiring core shell changes.

## Registration

An extension registers commands during `Load()`:

```cpp
void MyExtension::Load(ExtensionLoader &loader) {
[...]

PR #24793, opened 2026-08-16
Dispatch CLI dot-commands through shell_dot_command_ catalog functions

daviddallakyan2005:
## Summary

Restart of the first slice from #22079, as a new PR rather than a revival of that one.

Credit: @rustyconover opened #22079 with a C++ `ShellContext` / callback-manager approach for plugin `.` commands (including a video of the interaction model). @Mytherin reviewed that and asked not to take the C++ hook path, and instead sketched an incremental one: catalog functions named `shell_dot_command_<name>`, starting there because it does not conflict with the PEG-grammar work. This PR follows that catalog path only. It is not a port of the closed C++ hooks.

What this PR adds:

- Unknown `.cmd` looks up catalog table functions/macros named `shell_dot_command_<cmd>` and `CALL`s them with `(user_input, extra_info)`.
- Result protocol is `command` / `input` columns; only `print` is implemented (stdout + newline).
- Exact match first, then unique prefix among catalog suffixes (so `.ver` still hits `.version`). Ambiguous prefixes error instead of first-match.
- `.help`, tab-completion, and "Did you mean" include those catalog names (skipping names already in the static table). Description comes from `FunctionDescription`.
- `.version` is migrated off the static metadata table onto `shell_dot_command_version` in the shell extension. `-version` / `ShowVersionAndExit` are unchanged.
- Catalog listing runs in a transaction so `.` commands do not hit `ActiveTransaction` with no txn, and listing/dispatch failures do not unwind the CLI to `Exited due to error` after an aborted transaction.

Not in this PR: `~/.duckdb/shell_history.db`, PEG highlighting/autocomplete, streaming `clear_line` / MAP+VARIANT, or folding the rest of the static dot-commands into the grammar.

## Test plan

```
GEN=ninja make reldebug
python3 -m pytest tools/shell/tests/test_shell_dot_command.py --shell-binary build/reldebug/duckdb
python3 -m pytest tools/shell/tests --shell-binary build/reldebug/duckdb
```

Confirm `.version`, `.ver`, and `-version` still behave as before. The new file also covers a plugin `CREATE MACRO shell_dot_command_hello(...)`, unknown `.` commands, `.help` listing, and `.version` after an aborted transaction.

### DIFF
--- a/tools/shell/shell_metadata_command.cpp
+++ b/tools/shell/shell_metadata_command.cpp
@@ -1053,10 +1053,15 @@ static vector<CatalogDotCommandInfo> ListCatalogDotCommands(const ShellState &st
 			});
 		}
 	};
-	context.RunFunctionInTransaction([&]() {
-		collect(duckdb::CatalogType::TABLE_FUNCTION_ENTRY);
-		collect(duckdb::CatalogType::TABLE_MACRO_ENTRY);
-	});
+	try {
+		context.RunFunctionInTransaction([&]() {
+			collect(duckdb::CatalogType::TABLE_FUNCTION_ENTRY);
+			collect(duckdb::CatalogType::TABLE_MACRO_ENTRY);
+		});
+	} catch (std::exception &) {
+		// Listing must not abort the CLI (e.g. an invalidated transaction).
+		return {};
+	}
 	return result;
 }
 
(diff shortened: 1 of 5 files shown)


tools/shell/shell_metadata_command.cpp
```diff
@@ -1017,6 +1005,191 @@ static const MetadataCommand metadata_commands[] = {
...
+}
+
+static vector<CatalogDotCommandInfo> ListCatalogDotCommands(const ShellState &state) {
+	vector<CatalogDotCommandInfo> result;
+	if (!state.conn || !state.conn->context) {
+		return result;
+	}
+	auto &context = *state.conn->context;
+	duckdb::case_insensitive_set_t seen;
+	auto collect = [&](duckdb::CatalogType type) {
```
Mytherin, line 1032: Instead of trying to find the dot command maybe we can just call it and handle the error if it doesn't exist?

Mytherin (commented):
Thanks for the PR - looks good, I think this PR can just be simplified:

PR #24800, opened 2026-08-16
Don't dlopen an extension that is already linked in

### LINKED ISSUE #24799: unittest loads a second copy of a statically linked extension; the CLI does not
### What happens?

`build/test/unittest` loads a second copy of an extension that is already statically linked into the `libduckdb.so` it runs against. The CLI does not, and the test passes silently.

The two copies then have separate constructors and destructors, but the dynamic linker merges the symbol. For an extension with a mutable namespace-scope static, that is a use-after-free at teardown.

### To Reproduce

Build with one out-of-tree extension linked in:

```bash
DUCKDB_EXTENSIONS='sqlite_scanner' make
```

A test file that installs and loads it:

```
# name: test/sql/gate/SQ2.test
# group: [gate]

require-env LOCAL_EXTENSION_REPO

statement ok
INSTALL sqlite_scanner FROM '{LOCAL_EXTENSION_REPO}';

statement ok
LOAD sqlite_scanner;
```

```console
$ LOCAL_EXTENSION_REPO=build/release/repository LD_DEBUG=libs \
    build/release/test/unittest --test-dir . 'test/sql/gate/SQ2.test' 2>&1 \
    | grep -c 'calling init: .*\.duckdb_extension'
1                       # the same statements in build/release/duckdb print 0
```

One query shows the divergence. Same build tree, two binaries:

```console
$ build/release/duckdb -unsigned -c \
    "SELECT extension_name, install_mode FROM duckdb_extensions()
     WHERE extension_name IN ('sqlite_scanner','parquet');"
parquet          STATICALLY_LINKED
sqlite_scanner   STATICALLY_LINKED

# the same query from a .test file under build/release/test/unittest:
parquet          STATICALLY_LINKED
sqlite_scanner   NOT_INSTALLED
```

### Cause

1. `test/sqlite/sqllogic_test_runner.cpp:41` sets `config->options.load_extensions = false`, so the test binary never calls `LoadAllExtensions` (`src/main/database.cpp:370`) and no linked-in extension is registered, while its code sits mapped in the process.
2. `ExtensionHelper::LoadExternalExtension` (`src/main/extension/extension_load.cpp:647-666`) does prefer a linked-in copy, but gates on `internal_extensions[].statically_loaded`, which is set by the `DUCKDB_EXTENSION_<NAME>_LINKED` macros. `sqlite_scanner`'s entry is a hardcoded `false` (`src/main/extension/extension_helper.cpp:109`) and out-of-tree extensions are not in that table at all, so the guard cannot fire for them and the loader falls through to `dlopen`.
3. `LinkedExtensions()`, the CMake-generated list that does know what is linked in, has no callers in `src/`.

Autoload reaches the same code (`TryAutoLoadExtension`, `extension_helper.cpp:206-223`), so an `ATTACH` or a `SET` that triggers autoloading is enough to hit this.

### Impact

In a ducklake build this is a heap-use-after-free at teardown. Under a preloaded libasan it reports on 15/15 runs; the untooled binary aborts with `double free or corruption (fasttop)` on 7 of the same 15, always after every assertion has passed. The undefined behaviour happens on every run where both copies load, and the glibc abort is only the subset whose heap layout trips the fastbin check. Related: `duckdb/ducklake#1050`, and #5612 here, which diagnosed this in 2022 ("not counting statically loaded extensions as already loaded") before going stale. PR #22498 added the guard in step 2, scoped to in-tree extensions, which doesn't cover the out-of-tree case.

### Possible fix

Ask the generated loader instead of the macro table:

```cpp
if (!ExtensionHelper::IsFullPath(options.extension_name)) {
    DuckDB db_wrapper(db);
    if (ExtensionHelper::LoadExtension(db_wrapper, logical_name) == ExtensionLoadResult::LOADED_EXTENSION) {
        return;
    }
}
```

### OS:

Debian 12, x86_64, kernel 6.1.0-44, glibc 2.36

### DuckDB Version:

`main` @ `f8e1c96a53` (`v1.6.0-dev12739`), also reproduces at `ac5c6d11c3`

### DuckDB Client:

`build/test/unittest` and the CLI, Release, clang 19

### Full Name:

Gavin Brand

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

GavinPizza:
Fixes #24799

`build/test/unittest` loads a second copy of an extension that is already statically linked into the `libduckdb.so` it runs against. Both copies then construct and destroy the same statics, which is an ODR violation, and for an extension with a mutable namespace-scope static it is a use-after-free at teardown.

`LoadExternalExtension` already prefers a linked-in copy, but it decides by scanning the hardcoded default-extension table, whose `statically_loaded` flag comes from the `DUCKDB_EXTENSION_<NAME>_LINKED` macros. Out-of-tree extensions are either absent from that table or hardcoded to `false`, so the check can never fire for them. This asks the CMake-generated loader instead, which is the list that actually knows what is linked into the binary.

One consequence worth flagging: `LOAD <ext>` for a linked-in out-of-tree extension now succeeds instead of failing with "Install it first". That already happens for in-tree extensions such as `parquet`, so this removes a difference rather than adding a behaviour.

One test needed a companion change. duckdb-sqlite had no `Version()` override, so once the linked-in copy answers, `extension_version` came back empty where the `.so` footer used to supply it, and `test/sql/extensions/version_is_valid_sqlite.test` failed. That is fixed by duckdb/duckdb-sqlite#213, and duckdb/duckdb-sqlite#214 there restores a `std::move` that C++17 needs, without which duckdb cannot compile the extension at all. This branch bumps the `GIT_TAG` pin to that merge and drops patch 0007, whose hunks are now upstream.


Mytherin:
Thanks!

PR #23251, opened 2026-06-12
Add setting: SET dialect_compatibility_mode='spark' 

yan-alex:
Can be extended to support other dialects as well

PR #24801, opened 2026-08-16
Initial boilerplate for `DialectExtension`

Tishj:
This PR is part of https://github.com/duckdblabs/duckdb-internal/issues/10346

It adds:
- `FROM duckdb_dialects()`
- `SELECT current_setting('current_dialect');`

For now the setting is an empty shell, to be populated by follow up PRs.
The only behavior that's tied to changing `current_dialect` for now is calling `Invalidate` on the parser cache.

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -257,6 +257,18 @@
         ]
     },
     {
+        "name": "current_dialect",
+        "description": "The SQL dialect used by the parser",
+        "type": "VARCHAR",
+        "scope": "global",
+        "default_value": "duckdb",
+        "on_callbacks": [
+            "set",
+            "reset"
+        ],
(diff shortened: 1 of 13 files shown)


Mytherin:
Thanks - should this be integrated with the setting introduced in https://github.com/duckdb/duckdb/pull/23251?

Tishj:
> Thanks - should this be integrated with the setting introduced in #23251?

I think eventually yes, but I'm focusing on the parser side for now.
We would also lose coverage seeing as there's no "spark" dialect added yet, so replacing `test/sql/settings/dialect_compatibility_mode.test` wouldn't be possible yet.

Mytherin:
Alright - sounds good. Thanks!

PR #24720, opened 2026-08-12
Dispatch Go nightly build

mlafeldt:
Go PR: https://github.com/duckdb/duckdb-go-bindings/pull/104

Same as https://github.com/duckdb/duckdb/pull/23352 but for Go.

refs https://github.com/duckdblabs/duckdb-internal/issues/7516

PR #24802, opened 2026-08-16
Fix Go nightly workflow dispatch

mlafeldt:
PR #24720 added the Go nightly dispatch with the wrong workflow extension. 🤦🏻‍♂️ 

Point it at `nightly.yml` so scheduled builds can start.

(I've considered renaming the extension in duckdb-go-bindings, but .yml is way more common in DuckDB land.)

refs https://github.com/duckdblabs/duckdb-internal/issues/7516


Mytherin:
Thanks!

PR #24803, opened 2026-08-16
Bump duckdb-httpfs to main and shrink its patch

carlopi:
Moves the httpfs pin from a7b3862 (2026-08-05) to 5dfa24c (2026-08-13), 38 commits, mostly the S3 request/upload reimplementation.

The setting-name-identifier patch no longer applied. Its source hunks have been absorbed upstream -- httpfs now takes Identifier in TryGetSetting, S3KeyValueReader and SetExtensionOptionValue -- so only the four autoloading test hunks remain, which still expect the quoted form of the "Setting with name %s is not in the catalog" error.


Mytherin:
Thanks!

PR #24804, opened 2026-08-16
Refresh secret infrastructure, and use for `aws`'s credential_chain, that will now by default not persist credentials to disk but re-derivate

carlopi:
Currently (same in `main` and `v1.5-variegata` branches) refreshed secrets still carry the current set of viable credentials as part of the secret, and on refresh just override them.

Proposal is for adding a mode where credentials are NOT derived, basically the secret might become a recipe, and the actual credentials are stored only in-memory (instead of persisted to the actual secret, that potentially is to disk). This allows more complex refresh logic, for example eager refresh when close to the validity window end.

This is stacked on top of https://github.com/duckdb/duckdb/pull/24803 (bumping httpfs), and also add then an extra patch to both `httpfs` and `aws` to demo this behaviour.


src/include/duckdb/main/secret/secret.hpp
```diff
@@ -173,31 +176,122 @@ class BaseSecret {
...
+	void AbortRefresh() {
+		auto expected = SecretRefreshState::IN_REFRESH;
+		refresh_state.compare_exchange_strong(expected, SecretRefreshState::MUST_REFRESH);
+	}
+	SecretRefreshState GetRefreshState() const {
+		return refresh_state;
+	}
+
+public:
+	//! Held while re-deriving, so that concurrent lookups derive exactly once
```
lnkuiper, line 251: Can we use `annotated_mutex` and `DUCKDB_GUARDED_BY`. Also, I feel like these mutexes and atomics should not be public

src/include/duckdb/main/secret/secret.hpp
```diff
@@ -173,31 +176,122 @@ class BaseSecret {
...
+	mutex refresh_lock;
+	//! Whether the derived values can still be handed out
+	atomic<SecretRefreshState> refresh_state;
+	//! The thread that is re-deriving, meaningful only while the state is IN_REFRESH
+	atomic<thread_id> refreshing_thread;
+
+private:
+	//! Guards `current`: C++17 has no atomic shared_ptr
+	mutable mutex snapshot_lock;
+	shared_ptr<const SecretDerivedValues> current;
```
lnkuiper, line 261: We can do `atomic_load` and `atomic_store`, I don't think this lock is needed

src/include/duckdb/main/secret/secret.hpp
```diff
@@ -173,31 +176,122 @@ class BaseSecret {
...
+		return refresh_state;
+	}
+
+public:
+	//! Held while re-deriving, so that concurrent lookups derive exactly once
+	mutex refresh_lock;
+	//! Whether the derived values can still be handed out
+	atomic<SecretRefreshState> refresh_state;
+	//! The thread that is re-deriving, meaningful only while the state is IN_REFRESH
+	atomic<thread_id> refreshing_thread;
```
lnkuiper, line 256: This isn't initialised in the constructor, could that be a problem?

src/include/duckdb/main/secret/secret.hpp
```diff
@@ -243,6 +345,62 @@ class KeyValueSecret : public BaseSecret {
 		return true;
 	}
 
+	//! Install a set of derived values, visible to every copy of this secret
+	void SetDerivedValues(identifier_tree_t<Value> values) const {
```
lnkuiper, line 349: Is this function used? Could this function and potentially more become `private`?

src/include/duckdb/main/secret/secret_manager.hpp
```diff
@@ -128,6 +128,12 @@ class SecretManager {
 	DUCKDB_API BoundStatement BindCreateSecret(CatalogTransaction transaction, CreateSecretInput &info);
 	//! Lookup the best matching secret by matching the secret scopes to the path
 	DUCKDB_API SecretMatch LookupSecret(CatalogTransaction transaction, const string &path, const string &type);
+	//! Re-derive the transient values of a secret in place, by re-running the recipe stored in its
+	//! 'refresh_info'. The catalog entry is left untouched: the re-derived values are installed in the
+	//! state shared by every copy of the secret, so no secret storage is written and no manager lock is
+	//! needed. Concurrent callers for the same secret derive exactly once. Returns false when the secret
+	//! carries no recipe, or when re-running it produced no derived values.
+	DUCKDB_API bool TryRefreshSecret(ClientContext &context, const KeyValueSecret &secret);
```
lnkuiper, line 136: Can we do less verbose comments all over the PR

lnkuiper (changes requested):
Thanks for the PR!

1. I don't understand why CI is green, the httpfs patch calls `TryRefreshSecret` but never calls `MarkMustRefresh`. If I run the httpfs refresh tests myself, they fail
2. `duckdb_secrets()` can now expose sensitive options through `refresh_info`, only top-level fields are redacted, not nested fields

Codex found this:
>  * Refresh capability is inferred from ambiguous runtime state.
>
>     src/include/duckdb/main/secret/secret.hpp:359 treats any literal refresh_info field as a core refresh recipe, while a missing snapshot means derivation >is required.
>
>     This admits two broken states:
>      - A provider with refresh_info but no transient_keys is re-run on every lookup. The “nothing derived” path at src/main/secret/secret_manager.cpp:361 >leaves the snapshot absent, so NeedsDerivation() immediately becomes true again. A focused probe produced derivation counts 1 → 2 → 3 across creation >and two
>        lookups.
>
>      - A provider declaring transient values without refresh_info has them removed before persistence at src/main/secret/secret_manager.cpp:273, but they can >never be recovered. A probe showed the value present before restart, absent afterward, and still absent after lookup.
>
>     Refresh support should be explicit provider metadata/API, with transient-without-recipe rejected and “successfully derived an empty snapshot” distinct >from “never derived.”

Some more comments below:

PR #24805, opened 2026-08-16
Prune row groups with cross-column comparisons

### LINKED ISSUE #23884: Multiple-column predicates doesn't leverage stats properly
### What happens?

See repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE xy AS
         SELECT range::BIGINT AS i, (range + 200000)::BIGINT AS j
         FROM range(400000);
memory D CHECKPOINT;
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 

-- Within every group, max(i) < min(j), so shouldn't read any row group
memory D EXPLAIN ANALYZE SELECT * FROM xy WHERE i > j;
╭─ Summary ───────────╮
│ Total Time: 0.0005s │
╰─────────────────────╯
╭─ Filter ──────────────────╮
│ Expression: i > j         │
│ 0 rows                0µs │
╰─────────────┒┎────────────╯
╭─ Table Scan ┚┖────────────╮
│ Table: memory.main.xy     │
│ Type: Sequential Scan     │
│ Projections: i, j         │
│ Row Groups Scanned: 4 / 4 │
│ 400,000 rows          0µs │
╰───────────────────────────╯
```

### OS:

macos

### DuckDB Version:

0fe29b9a2d33459e4426d7971bcf4bb494a9b69a

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Andrewtangtang:
## Summary

Resolves #23884

Currently, cross-column predicates are evaluated only after every row group has been scanned. This change introduces a new `RowGroupExpressionFilter` to store cross-column expressions. It enables direct numeric column comparisons to be propagated to row-group pruning, so row groups that cannot match can be skipped safely.

## Results

```sql
CREATE TABLE xy AS
SELECT range::BIGINT AS i, (range + 200000)::BIGINT AS j
FROM range(400000);

CHECKPOINT;

EXPLAIN ANALYZE SELECT count(*) FROM xy WHERE i > j;
```

### Before

```sql
╭─ Summary ───────────╮
│ Total Time: 0.0046s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ─────╮
│ Aggregates: count_star()  │
│ 1 row                 0µs │
╰─────────────┬─────────────╯
╭─ Filter ────┴─────────────╮
│ Expression: i > j         │
│ 0 rows                0µs │
╰─────────────┒┎────────────╯
╭─ Table Scan ┚┖────────────╮
│ Table: memory.main.xy     │
│ Type: Sequential Scan     │
│ Projections: i, j         │
│ Row Groups Scanned: 4 / 4 │
│ 400,000 rows          0µs │
╰───────────────────────────╯
```

### After

```sql
╭─ Summary ───────────╮
│ Total Time: 0.0018s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ─────╮
│ Aggregates: count_star()  │
│ 1 row                 0µs │
╰─────────────┬─────────────╯
╭─ Filter ────┴─────────────╮
│ Expression: i > j         │
│ 0 rows                0µs │
╰─────────────┬─────────────╯
╭─ Table Scan ┴─────────────╮
│ Table: memory.main.xy     │
│ Type: Sequential Scan     │
│ Projections: i, j         │
│ Row Groups Scanned: 0 / 4 │
│ 0 rows                0µs │
╰───────────────────────────╯
```

### Note

Once this PR is merged, I plan to open follow-up PRs that reuse the current `RowGroupExpressionFilter` structure to support more general cross-column expressions, including:

- Disjunctions such as `i > j OR k < l`
- Arithmetic expressions such as `i + 1 > j` or `i * 2 <= j`
- Monotonic casts or functions such as `CAST(i AS HUGEINT) > j`

### DIFF
--- a/src/optimizer/filter_combiner.cpp
+++ b/src/optimizer/filter_combiner.cpp
@@ -402,6 +402,41 @@ FilterPushdownResult FilterCombiner::TryPushdownGenericExpression(LogicalGet &ge
 	if (bindings.empty()) {
 		return FilterPushdownResult::NO_PUSHDOWN;
 	}
+	if (bindings.size() == 2 && bindings[0] != bindings[1] &&
+	    get.function.GetName().GetIdentifierName() == "seq_scan" && BoundComparisonExpression::IsComparison(expr)) {
+		const auto &comparison = expr.Cast<BoundFunctionExpression>();
+		const auto &left = BoundComparisonExpression::Left(comparison);
+		const auto &right = BoundComparisonExpression::Right(comparison);
+		const auto comparison_type = comparison.GetExpressionType();
+		const bool supported_comparison = comparison_type == ExpressionType::COMPARE_EQUAL ||
+		                                  comparison_type == ExpressionType::COMPARE_GREATERTHAN ||
+		                                  comparison_type == ExpressionType::COMPARE_GREATERTHANOREQUALTO ||
+		                                  comparison_type == ExpressionType::COMPARE_LESSTHAN ||
+		                                  comparison_type == ExpressionType::COMPARE_LESSTHANOREQUALTO;
+		if (!supported_comparison || left.GetExpressionClass() != ExpressionClass::BOUND_COLUMN_REF ||
+		    right.GetExpressionClass() != ExpressionClass::BOUND_COLUMN_REF || !left.GetReturnType().IsNumeric() ||
+		    !right.GetReturnType().IsNumeric()) {
+			return FilterPushdownResult::NO_PUSHDOWN;
+		}
+		const auto &left_ref = left.Cast<BoundColumnRefExpression>();
+		const auto &right_ref = right.Cast<BoundColumnRefExpression>();
+		if (left_ref.Binding().table_index != get.table_index || right_ref.Binding().table_index != get.table_index ||
+		    left_ref.Binding().column_index >= get.GetColumnIds().size() ||
+		    right_ref.Binding().column_index >= get.GetColumnIds().size() ||
+		    get.GetColumnIds()[left_ref.Binding().column_index].IsVirtualColumn() ||
+		    get.GetColumnIds()[right_ref.Binding().column_index].IsVirtualColumn()) {
+			return FilterPushdownResult::NO_PUSHDOWN;
+		}
+
+		auto left_bound_ref = make_uniq<BoundReferenceExpression>(left_ref.GetAlias(), left_ref.GetReturnType(), 0);
+		auto right_bound_ref = make_uniq<BoundReferenceExpression>(right_ref.GetAlias(), right_ref.GetReturnType(), 1);
+		auto filter_expr =
+		    BoundComparisonExpression::Create(comparison_type, std::move(left_bound_ref), std::move(right_bound_ref));
+		vector<ProjectionIndex> column_indexes {left_ref.Binding().column_index, right_ref.Binding().column_index};
+		get.table_filters.PushRowGroupFilter(
+		    RowGroupExpressionFilter(std::move(column_indexes), std::move(filter_expr)));
+		return FilterPushdownResult::PUSHED_DOWN_PARTIALLY;
+	}
 	// we can only pushdown expressions that refer to exactly one column
 	for (idx_t i = 1; i < bindings.size(); i++) {
 		if (bindings[i] != bindings[0]) {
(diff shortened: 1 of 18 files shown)


lnkuiper:
Thanks for the PR! I think the row-group pruning itself is useful, but we should avoid adding a separate `RowGroupExpressionFilter` representation.

We are planning to generalize table filters so they can reference multiple columns and be evaluated as tuple filters. This structure duplicates most of the machinery that work will need: ownership, serialization, equality, copying, column remapping, and expression/statistics evaluation. The proposed follow-ups would deepen that duplication further.

I would prefer representing this as a generic multi-column table filter: an expression whose BoundReferences refer to a vector of projection indexes (but maybe @kryonix has a more informed opinion of that). For now the native scan can use it only for row-group pruning and keep the original logical filter by returning `PUSHED_DOWN_PARTIALLY`; tuple filtering can later execute the same filter representation.

`ExpressionFilter::TryGetExpressionStatistics` already accepts statistics for multiple inputs, so I think extending the common expression-filter statistics path is also preferable to hard-coding another comparison evaluator in `RowGroup::CheckZonemap`.

Andrewtangtang:
Thanks for the review. I agree that adding a separate filter type would duplicate the existing machinery. I updated the PR to extend `ExpressionFilter` with multi-column projection indexes and removed `RowGroupExpressionFilter`. Row-group pruning now collects the mapped column statistics and uses the common ExpressionFilter statistics path.

Andrewtangtang:
The failed regression jobs appear to be caused by timeouts while installing CI dependencies through the runner's Ubuntu APT mirrors, before any benchmarks are run.

src/optimizer/filter_combiner.cpp
```diff
@@ -402,6 +402,41 @@ FilterPushdownResult FilterCombiner::TryPushdownGenericExpression(LogicalGet &ge
 	if (bindings.empty()) {
 		return FilterPushdownResult::NO_PUSHDOWN;
 	}
+	if (bindings.size() == 2 && bindings[0] != bindings[1] &&
+	    get.function.GetName().GetIdentifierName() == "seq_scan" && BoundComparisonExpression::IsComparison(expr)) {
```
kryonix: Could we avoid identifying the native table scan by the literal function name? This path relies specifically on DuckDB row-group statistics, while a function name is mutable metadata and can be reused. `get.GetTable()` together with `IsDuckTable()` appears to express the required invariant directly. It also avoids silently disabling pruning if seq_scan is renamed or admitting an unrelated function with that name.

```cpp
  auto table = get.GetTable();
  if (bindings.size() == 2 && bindings[0] != bindings[1] && table && table->IsDuckTable() &&
      BoundComparisonExpression::IsComparison(expr)) {
```

kryonix (commented):
Looks good overall. The design reuses `ExpressionFilter` and `TableFilterSet` rather than introducing another filter type, while correctly keeping cross-column filters as pruning-only hints and retaining the original predicate for row-level filtering. The metadata is carried through remapping, serialization, physical planning, and row-group scanning. I found no correctness issues, and the SF100 runs showed no regression.

One non-blocking simplification: I would avoid identifying the native scan through `GetIdentifierName() == "seq_scan"`. `get.GetTable()` with `IsDuckTable()`, or an explicit capability, would express the row-group pruning requirement without relying on a function name.

src/optimizer/filter_combiner.cpp
lnkuiper: Maybe we can do arbitrary bindings? i.e., `bindings.size() > 1`. Also, is this infrastructure suitable for multi-column tuple-level filtering? Just wanted to ask you since you've looked into that before

kryonix: Yes, I think this PR is a good basis for both. The `ExpressionFilter` plus `column_indexes` representation already supports an arbitrary number of columns and is propagated through the required planning and scan stages. For arbitrary bindings, we should count _distinct_ bindings rather than use `bindings.size() > 1`, since an expression may reference the same column multiple times. Each distinct binding can then be mapped to a dense `BoundReference`.

Tuple-level filtering would still require an execution path that loads the referenced columns and evaluates the expression over a `DataChunk`, but it should be possible without redesigning this infrastructure. I did experiment with multi-column tuple-level filtering when I did the tablefilter stuff. However, I could not get that to work without (big) regressions at the time.

PR #24808, opened 2026-08-16
Replace trailing variadic functions with `*args` and `**kwargs`

Maxxen:
**This PR is a work-in-progress**, the functionality is there, but there's a lot of stuff to clean up before this is ready to be merged.

This PR (attempts) to remove support for "trailing" variadic arguments from function using the `FunctionSignature` (e.g. scalar/window/aggregates) and instead introduces support for positional and named _parameter pack_ parameters. In other words, Python-style `*args` and `**kwargs`. 

If a function signature contains a positional parameter pack (`*args`), any extraneous arguments will automatically be packed into an unnamed struct (`LogicalType::TUPLE`) and passed to the function at the same "position" as the positional pack is defined in the signature during execution.

Similarly, the binder will automatically pack any extraneous _named_ arguments into a `STRUCT` where the fields are the passed argument names - if the function signature has a named parameter pack (`**kwargs`).

This is implemented by introducing a new `ARGUMENT_PACK` `OperatorExpression`, which essentially just copies the implementation of the `struct_pack`/`row` functions. Unfortunately, we need a new operator type and can't just manufacture a `struct_pack`/`row` function expression, as they themselves are also variadic functions, and so would bundle their arguments into another pack, which would require another `struct_pack`/`row` call, etc. etc. - indefinite recursion.

This has the following benefits:
- Functions always know exactly what arguments will be given at which position.
- Functions can now distinguish between positional/keyword-varargs, and can _always_ access the name of a given argument by just checking the argument types, no need to know anything about expressions or access `expr->GetAlias()` (which may make it easier to remove expression aliases in the future)
- Functions can now define _keyword-only_ arguments (simply define a named argument _after_ a positional parameter pack in the signature, e.g. `foo(x, *args, this_arg_can_only_be_passed_by_name, **kwargs)`
- We can now support variadic used-defined macros (if we switch them over to also use the `FunctionSignature`
- We can now also move `TableFunction`/`PragmaFunction`s to use the `FunctionSignature` too as their "named parameter map" is essentially just a constant-folded `**kwargs`.
- We can now define `FunctionSignature`s for "type constructors", making the future binding/creation of both built-in and custom types have much nicer error messages and automatically generated documentation.
- We could support __both__ struct and array "spreading"/"unpacking" (i.e.`*`, `**`, or `...` syntax) in function calls, allowing for very expressive column/expression manipulation in SQL, narrowing the gap further between our SQL and various data frame APIs.
- Maps nicely to Python UDFs (which we want to support post v2.0 anyway), but also UDFs in other languages where trailing arguments are packed into lists/tuples (e.g. java/C#).

This PR is quite big cause it moves over all existing variadic scalar/window/aggregate functions to use parameter packs instead of trailing varargs.

Backwards compatibility is preserved by "unrolling" any bound parameter packs back into trailing arguments when serializing to an older storage version. We also unroll parameter packs in the V1-C-API so that c-api functions that support trailing varargs don't need any changes.

However, there is one function that still use trailing varargs, which is why this PR does not remove that functionality yet. That function is `struct_pack`

# We need to talk about `struct_pack`

`struct_pack` is a very different function. It takes an arbitrary list of _named arguments_, e.g. `struct_pack( x := 42, y := 'foo')` and returns a new `STRUCT` column containing the arguments as fields, with their passed names as field names, i.e. `{'x': 42, 'y': 'foo'}`. In theory we could just define its signature as `struct_pack(**kwargs)` - except that it also has the really odd behavior of _automatically naming unnamed arguments based if the argument expression alias_. Aliased expressions are in themselves kind of odd in that they are created very inconsistently, but it is basically what allows you to currently do:

```sql
select struct_pack(i) from range(0,1) as r(i);
----
┌──────────────────┐ -- wait, I never passed ` i := i`?
│  struct_pack(i)  │
│ struct(i bigint) │
├──────────────────┤
│ {'i': 0}         │
└──────────────────┘
```

I don't even know if this is intentional, but it is very annoying cause no other function really work this way (very few functions supported passing arguments by name to begin with), and it also e.g. allows you to pass a positional argument _after_ a named argument _if that argument comes from an aliased expression_. Again, no other language with named arguments support this (it should be a parser error IMO and not even reach the binder).

I actually discovered this back in https://github.com/duckdb/duckdb/pull/22941, and ended up writing a very hacky special binding mode for "implicit argument capture" - which only `struct_extract` uses. I no longer think this is the right way to go, we should just remove this capability from `struct_extract`, make it work like any other function, and reject positional-after-named arguments in the parser way before it reaches the binder (which will simplify a lot of binding code).

Now, I can see that being able to automatically wrap up a bunch of expression in a struct and have the field names be automatically generated could be useful. So my proposal would be that we preserve this behavior when creating structs through the _literal syntax_, i.e. `{'x' := 42, i} from range(0, 1) as r(i)`, by having the parser emit a new parsed `PackExpression`, which the binder turns into a `PACK_OPERATOR` `OperatorExpression`, without going through the function binder. The braced `{}` syntax would then essentially mirror the behavior of `SELECT`-lists, which I think is kinda neat. A table/relation is more or less just an unpacked struct after all?

This is however a minor user facing SQL change - but I would argue that it should be acceptable because it is a very niche, and essentially unintended behavior. We would then be able to unify function call semantics across all of our functions and function types.

### DIFF
--- a/extension/core_functions/scalar/array/array_value.cpp
+++ b/extension/core_functions/scalar/array/array_value.cpp
@@ -1,7 +1,10 @@
 #include "duckdb/common/vector/array_vector.hpp"
 #include "core_functions/scalar/array_functions.hpp"
 #include "duckdb/function/scalar/nested_functions.hpp"
+#include "duckdb/common/vector/struct_vector.hpp"
+#include "duckdb/planner/expression/bound_argument_pack.hpp"
 #include "duckdb/storage/statistics/array_stats.hpp"
+#include "duckdb/storage/statistics/struct_stats.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
 
 namespace duckdb {
@@ -11,20 +14,22 @@ namespace {
 void ArrayValueFunction(DataChunk &args, ExpressionState &state, Vector &result) {
 	auto array_type = result.GetType();
 
+	auto &values = StructVector::GetEntries(args.data[0]);
+
 	D_ASSERT(array_type.id() == LogicalTypeId::ARRAY);
-	D_ASSERT(args.ColumnCount() == ArrayType::GetSize(array_type));
+	D_ASSERT(values.size() == ArrayType::GetSize(array_type));
 
 	auto &child_type = ArrayType::GetChildType(array_type);
 
 	result.SetVectorType(VectorType::CONSTANT_VECTOR);
-	for (idx_t i = 0; i < args.ColumnCount(); i++) {
-		if (args.data[i].GetVectorType() != VectorType::CONSTANT_VECTOR) {
+	for (auto &value : values) {
+		if (value.GetVectorType() != VectorType::CONSTANT_VECTOR) {
 			result.SetVectorType(VectorType::FLAT_VECTOR);
 		}
 	}
 
 	auto num_rows = args.size();
-	auto num_columns = args.ColumnCount();
+	auto num_columns = values.size();
 
 	auto &child = ArrayVector::GetChildMutable(result);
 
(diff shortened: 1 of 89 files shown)


Maxxen:
Superseded by https://github.com/duckdb/duckdb/pull/25991#event-31618303492

PR #24811, opened 2026-08-17
Temp file: avoid races via naming convention with slots

carlopi:
More radical alternative to https://github.com/duckdb/duckdb/pull/24655 and https://github.com/duckdb/duckdb/pull/24758.

Currently temporary files (used by duckdb to spill to disk) might collide between different DatabaseInstances, both same process and different co-existing processes. This PR names every temporary file after the instance that owns it — the process it runs in, plus which DatabaseInstance within that process — so an instance can tell its own files from everyone else's, and can ask the OS whether those others are still alive.

Snapshot of the files in a temporary directory would look like:

before:
```
  .tmp/
   -> duckdb_temp_block-12345.block
   -> duckdb_temp_block-querty.block
   -> duckdb_temp_storage_DEFAULT-0.tmp
   -> duckdb_temp_storage_DEFAULT-37.tmp
```
  after:
```
  .tmp/
   -> duckdb_temp_51234_0_owner
   -> duckdb_temp_51234_0_block-querty.block
   -> duckdb_temp_51234_0_storage_DEFAULT-37.tmp
   -> duckdb_temp_51234_1_owner
   -> duckdb_temp_51234_1_block-12345.block
   -> duckdb_temp_51234_1_storage_DEFAULT-0.tmp
```
Both owners above are in the same process (pid 51234), instances 0 and 1 — the case a pid alone cannot separate. A second process differs in the first number instead, e.g. duckdb_temp_9871_0_storage_DEFAULT-0.tmp. Say process 9871 is gone, then the next process to spill there cleans its files up.

One rule decides what may be removed: only files whose process is no longer running. The rest follows from it:
- our own pid is alive by definition, so nothing is ever removed under it — not even by us;
- a pid we cannot answer for counts as alive, because leaking disk is recoverable and deleting a live instance's spill files is not;
- teardown removes only what this instance owns, which needs no judgement about anyone else.

`duckdb_temp_<pid>_<id>_owner` is an empty file created with FILE_CREATE_NEW, and claiming an id is creating it. It is never locked and no instance ever opens another's — existing is its entire job. Two things make it necessary rather than decorative:
- an instance's data files are not evidence that it is alive. Every block it spilled can be freed again, the files deleted, and the instance goes on living with nothing on disk until it spills once more;
- O_EXCL makes the claim atomic against an instance that shares our pid but not our counter — two copies of duckdb linked into one process each count from zero.

It is removed last at teardown, so no id can be taken while files under that name are still being deleted, and a crashed owner's marker is swept with the rest of its files once its process is gone.

The sweep runs once per process per directory, and only from the first spill — TemporaryDirectoryHandle is built lazily on first eviction, so a database that never spills never reclaims anything. Repeating it per instance would only multiply the chances of misjudging a live process.

This improves in 2 areas:
- 2 processes or 2 instances, same folder, first to close would cancel all files, before would make other instance throw, now it will leave them in place
- 2 processes or 2 instances, same folder, they will not be ever overriden, given overriding would possibly reinterpret bytes

Also fixes a std::terminate reachable on main today: `~TemporaryDirectoryHandle` reaches RemoveFile via TryRemoveFile, which is "try" only in the sense of skipping an absent file and throws on any real failure.

Adds FileSystem::RemoveDirectoryIfEmpty, because RemoveDirectory recurses and the obvious composition — list it, remove it if it looked empty — races whoever writes in between. It is what stops teardown from taking a live peer's files with the directory.

Known limitations, both addressable as follow-ups:
- a pid identifies a process only within one PID namespace on one host. A temp_directory shared across containers, or over NFS between hosts, can hold files whose owner is alive elsewhere but absent locally. A potential follow-up is a policy file in the directory — an unknown policy means "only clean up after yourself" — which then lets a host/namespace key be introduced as a new policy value without older versions destroying files they cannot reason about; That would also avoid checking for PID, allowing an escape hatch if that code were to be off
- files left by a crash are now reclaimed when something next spills into that directory, where before they were wiped by the next instance's teardown. Since that wipe is exactly the behaviour being removed, the trade is deliberate, and a CALL cleanup_temporary_directory() is the natural follow-up for reclaiming on demand


Mytherin:
Maybe we can prefix files with a PID, and then check if the PID is still alive prior to doing a clean-up? That avoids having to do locking

carlopi:
I think if we don't want file locks then I would go for reopening: https://github.com/duckdb/duckdb/pull/24655

Mytherin:
The problem with uuids is that we can never clean up - i.e. this will just accumulate files forever if there's ungraceful shutdowns

carlopi:
I think I see what PID buys you, that is trivial to check if NOT in use (it can fail the other path, but that's ok).

I think both might be needed, so PID + short ID, but could be workable.

So idea is that on shutdown (or also startup) you check on other files, and clean up:
* your PID + your identifiers (those you know)
* different PID that is NOT in use (those you also know)
while leaving alone:
* your PID + different identifier (this can be a same-process doing its things)
* different PID + in use (this you are not sure whether is actually still the same process, but act safely)

And if the folder is empty, you can actually clean it up.

I'll iterate on that.

Mytherin:
I don't think a separate identifier is required. For your own shutdown you should limit to files you are explicitly tracking (which should account for all files you yourself have generated).

carlopi:
> I don't think a separate identifier is required. For your own shutdown you should limit to files you are explicitly tracking (which should account for all files you yourself have generated).

Shutdown, right, but PID + ID are needed on the other side, to avoid different (unconnected) instances on the same PID stamping on each other. I think there the idea of adding a `duckdb_temp_<PID>_<SOME_ID>.lock` file (even not actually locked) is required to say "I am around, please leave the relevant implicit folder to me".

Like, same python process having 2 unconnected duckdb doing parallel work.

Mytherin:
Fair enough. I think it would make sense to give each database a UUID that could then be used for this purpose (and potentially other purposes)

carlopi:
Updated according to latest iteration. Also rewritten the PR description, in particular left as follow ups the following:

Known limitations, both addressable as follow-ups:
- a pid identifies a process only within one PID namespace on one host. A temp_directory shared across containers, or over NFS between hosts, can hold files whose owner is alive elsewhere but absent locally. A potential follow-up is a policy file in the directory — an unknown policy means "only clean up after yourself" — which then lets a host/namespace key be introduced as a new policy value without older versions destroying files they cannot reason about; That would also avoid checking for PID, allowing an escape hatch if that code were to be off
- files left by a crash are now reclaimed when something next spills into that directory, where before they were wiped by the next instance's teardown. Since that wipe is exactly the behaviour being removed, the trade is deliberate, and a CALL cleanup_temporary_directory() is the natural follow-up for reclaiming on demand


I have NOT added UUID generation as part of this PR, was in https://github.com/duckdb/duckdb/pull/24655, I think that is orthogonal and can revive independently.

src/common/local_file_system.cpp
```diff
@@ -1496,6 +1513,25 @@ void LocalFileSystem::RemoveDirectory(const string &directory, optional_ptr<File
 	DeleteDirectoryRecursive(*this, directory, opener);
 }
 
+bool LocalFileSystem::RemoveDirectoryIfEmpty(const string &directory, optional_ptr<FileOpener> opener) {
```
Mytherin: Instead of adding this specialized method maybe we can add a `RemoveDirectoryExtended` that takes a `RemoveDirectoryOptions` field, then we don't need to plumb a bunch of specialized methods through the various layers when we inevitably want to extend it (e.g. Laurens is already adding almost the same method [here](https://github.com/duckdb/duckdb/pull/24840/changes#diff-de3c4b309f422c1d0823306c309bb134a6d369551c4b252d9f29e548382b9b34R610)).

src/storage/temporary_file_manager.cpp
```diff
@@ -703,9 +712,107 @@ void TemporaryFileManager::EraseUsedBlock(TemporaryFileManagerLock &lock, block_
...
+		return false;
+	}
+	owner.pid = NumericCast<int64_t>(pid);
+	owner.instance = NumericCast<idx_t>(instance);
+	return true;
+}
+
+static int64_t CurrentProcessId() {
+#ifdef _WIN32
+	return static_cast<int64_t>(GetCurrentProcessId());
```
Mytherin: Can we move this to a separate class (e.g. `SystemUtil`) to try and limit where we have OS specific information?

carlopi:
Blocked on https://github.com/duckdb/duckdb/pull/24840, that is indeed adding `RemoveDirectoryExtended`, to be then extended with `EMPTY` mode

carlopi:
Now stacked on top of https://github.com/duckdb/duckdb/pull/25265

Mytherin:
Thanks!

PR #24812, opened 2026-08-17
[Dev][Parser] `Identifier` match logic clean up

Tishj:
This PR aims to simplify/clean up the rather complex decision tree that currently makes up `IdentifierMatcher::MatchIdentifier`

I distilled the old logic to pseudo code:
```
if suggestion_type == type_name:
    return false if keyword type of [reserved, type_func, col_name]
else:
    type banned;
    if suggestion_type in [scalar_function, table_function]:
         banned = col_name
    else:
         banned = type_func
    allowed = `type_func` if banned == col_name else `col_name`

    is_banned = keyword type of [reserved, banned]
    is_allowed = keyword type of [unreserved, allowed]

    return false if is_banned and not is_allowed
```

The clean up moves this into a `IsAllowedKeyword` of which the pseudo code is:
```
return true if keyword_type == unreserved
if suggestion_type == type_name and keyword_type == type_name:
    return true
elif suggestion_type in [scalar_function, table_function] and keyword_type == type_func:
    return true
else:
    return true if keyword_type == col_name
return false
```

Which I believe is equivalent to the original


Mytherin:
Thanks!

PR #24814, opened 2026-08-17
Fix precision loss in FLOAT to DECIMAL casts

xuke-hat:
`12345678` can be represented exactly by both `FLOAT` and `DECIMAL(18,2)`, but currently `SELECT 12345678::FLOAT::DECIMAL(18,2)` returns `12345678.08`.

When casting to `DECIMAL(18,2)`, the value is promoted to `DOUBLE` and multiplied by `100.0`, producing the exact value `1234567800`. But it is then unnecessarily narrowed back to `FLOAT`, which rounds it to `1234567808`.

This PR converts the scaled and rounded `DOUBLE` value directly to the DECIMAL storage type, avoiding the unnecessary narrowing conversion.


Mytherin:
Thanks!

PR #24816, opened 2026-08-17
Fix partitioned COPY losing rows on __HIVE_DEFAULT_PARTITION__ collision

### LINKED ISSUE #24308: Partitioned COPY can lose data
### What happens?

See repro for details.

### To Reproduce

```sql
CREATE TABLE t AS
         SELECT * FROM (VALUES ('__HIVE_DEFAULT_PARTITION__', 1), (NULL, 2)) v(p, id);
COPY t TO '/private/tmp/duckdb_partition_hive_default_collision_min'
         (FORMAT PARQUET, PARTITION_BY (p), OVERWRITE);
SELECT count(*) AS rows_read, min(id), max(id)
         FROM read_parquet('/private/tmp/duckdb_partition_hive_default_collision_min/**/*.parquet',
                           hive_partitioning=true);
┌───────────┬─────────┬─────────┐
│ rows_read │ min(id) │ max(id) │
│   int64   │  int32  │  int32  │
├───────────┼─────────┼─────────┤
│         1 │       2 │       2 │
└───────────┴─────────┴─────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

carlopi:
Fixes #24308.

A partitioned COPY writes NULL partition values to a directory named __HIVE_DEFAULT_PARTITION__, while non-NULL values are URL-encoded. URL encoding leaves '_' unreserved, so the literal string '__HIVE_DEFAULT_PARTITION__' encoded to itself and shared a directory with the NULL partition. Since the output file index is tracked per partition value, both partitions wrote data_0.parquet into that directory and the second write silently clobbered the first, losing rows.

Escape a non-NULL partition value that would collide with the NULL sentinel by percent-encoding its first character, which still unescapes back to the original value on read.

### DIFF
--- a/src/common/hive_partitioning.cpp
+++ b/src/common/hive_partitioning.cpp
@@ -88,11 +88,13 @@ string HivePartitioning::Escape(const string &input) {
 
 string HivePartitioning::EscapeValue(const string &input) {
 	auto result = Escape(input);
-	if (result != DEFAULT_PARTITION_NAME) {
+	// the comparison is case-insensitive because on a case-insensitive file system a value that differs only in
+	// case still lands in the directory that is reserved for NULL values
+	if (!StringUtil::CIEquals(result, DEFAULT_PARTITION_NAME)) {
 		return result;
 	}
-	// this value escapes to the directory name that is reserved for NULL values - percent-encode the first
-	// character so the value gets its own directory while still unescaping back to the original value
+	// percent-encode the first character so the value gets its own directory while still unescaping back to the
+	// original value
 	static constexpr const char *HEX_DIGIT = "0123456789ABCDEF";
 	const auto first = static_cast<unsigned char>(result[0]);
 	string escaped = "%";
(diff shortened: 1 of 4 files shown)


lnkuiper (commented):
Thanks! Looks good, just realised we have an issue with case-insensitive file systems (like macos):
```sql
COPY (
    SELECT * FROM (
        VALUES ('__hive_default_partition__', 1), (NULL, 2)
    ) v(p, id)
) TO '/tmp/hive_case_collision'
(FORMAT PARQUET, PARTITION_BY (p), OVERWRITE);
SELECT count(*)
FROM read_parquet('/tmp/hive_case_collision/**/*.parquet',
                  hive_partitioning=true);
```
Should return 2, not 1.

Can we make the check case-insensitive and add this test?

carlopi:
Great angle the case-insensitivity

Mytherin:
Thanks!

Issue #24626, opened 2026-08-09
`partial_aggregate_pushdown` returns wrong holistic aggregate results over joins


Yibo-Dong:
### What happens?

`partial_aggregate_pushdown` rewrites an aggregate over a join into lower aggregate states plus an upper `combine_aggr`. For `quantile_cont`, `quantile_disc`, `median`, and `mad`, the default optimized plan returns values that do not match the ordinary join result.

In the minimized input below, the joined multiset of `a` values is:

```text
0, 0, 1, 1, 100, 100, 100, 100, 101, 101, 101, 101
```

The expected 25th-percentile continuous and discrete quantiles are both `1`, the median is `100`, and the median absolute deviation is `1`. The default plan returns `0.75`, `0`, `50.5`, and `50.0`.

#### Code

```sql
CREATE TABLE f(k INTEGER, a INTEGER, b INTEGER);
INSERT INTO f VALUES
  (0, 0, 1),
  (0, 1, 1),
  (1, 100, 1),
  (1, 101, 1);

CREATE TABLE d(k INTEGER, g INTEGER);
INSERT INTO d VALUES
  (0, 0),
  (0, 0),
  (1, 0),
  (1, 0),
  (1, 0),
  (1, 0);

SELECT
  b,
  g,
  quantile_cont(a, 0.25) AS q_cont,
  quantile_disc(a, 0.25) AS q_disc,
  median(a) AS med,
  mad(a) AS mad
FROM f
JOIN d USING (k)
GROUP BY b, g
ORDER BY b, g;

SET disabled_optimizers='partial_aggregate_pushdown';

SELECT
  b,
  g,
  quantile_cont(a, 0.25) AS q_cont,
  quantile_disc(a, 0.25) AS q_disc,
  median(a) AS med,
  mad(a) AS mad
FROM f
JOIN d USING (k)
GROUP BY b, g
ORDER BY b, g;
```

Run:

```sh
./build/release/duckdb -json -f bug.sql
```

#### Expected

| b | g | q_cont | q_disc | med | mad |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 0 | 1.0 | 1 | 100.0 | 1.0 |

#### Actual

| b | g | q_cont | q_disc | med | mad |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 0 | 0.75 | 0 | 50.5 | 50.0 |

#### Notes From Minimization

[...]

PR #24817, opened 2026-08-17
Add multiplicity-aware combine support and fix LIST state combination

### LINKED ISSUE #24806: Wrong `MEDIAN` / `QUANTILE_CONT` / `LIST` over an inner join
### What happens?

`PartialAggregatePushdown::TryDoubleEagerPushdown` rewrites `MEDIAN(fact.x) … JOIN … GROUP BY dim.g` into per-join-key `median … EXPORT_STATE` on one side, `count_star` on the other, a join of those partials, and `combine_aggr(state, opposite_count)`. That is correct for `SUM` / `COUNT` / `AVG` (the tests cover those). It is **not** correct for `MEDIAN`, `QUANTILE_CONT`, `MAD`, or `LIST`: the dimension-side multiplicity never makes it into the quantile/list state, so the result is the median of the *unweighted* per-key bags. On th table, group `g=1` is six `1`s and four `42`s (median **1.0**); the rewrite answers **21.5** (`median({1,1,42,42})`).

### To Reproduce

```sql
CREATE TABLE t(i INT, g INT, d INT, txt INT, chr INT);
INSERT INTO t VALUES
  (1, 1, 1, 1, 0),
  (1, 1, 1, 1, 0),
  (42, 1, 2, 3, 9),
  (42, 1, 1, 2, 2),
  (NULL, -7, 1, 0, 0),
  (42, NULL, 3, NULL, 3);

-- Expected: g=1 median 1.0. Actual: 21.5.
SELECT t1.g, MEDIAN(t3.i) AS m
FROM t t1
INNER JOIN t t2 ON t1.d = t2.d
INNER JOIN t t3 ON t2.txt = t3.chr
GROUP BY t1.g
ORDER BY 1 NULLS FIRST;
-- => -7, 1.0
-- =>  1, 21.5     WRONG
```
With the optimizer disabled, the result is correct
```sql
SET disabled_optimizers='partial_aggregate_pushdown';
SELECT t1.g, MEDIAN(t3.i) AS m
FROM t t1
INNER JOIN t t2 ON t1.d = t2.d
INNER JOIN t t3 ON t2.txt = t3.chr
GROUP BY t1.g
ORDER BY 1 NULLS FIRST;
-- => -7, 1.0
-- =>  1, 1.0      expected
```

### OS:

macOS

### DuckDB Version:

v2.0.0-alpha37826 (Cyanoptera)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Junwen An

### Affiliation:

NUS

- [x] Yes, I have

- [x] Yes, I have

Yes

Alignyx:
Fix #24806 

### Problem

`PartialAggregatePushdown::TryDoubleEagerPushdown` rewrites aggregates over joins into partial aggregate states. It exports a state from one side of the join, counts matching rows on the other side, and then calls `combine_aggr(state, multiplicity)`.

This works for aggregates whose combine implementation honors the multiplicity. However, list-backed aggregates such as `LIST`, `MEDIAN`, `QUANTILE_CONT`, and `MAD` used a custom `ListCombineFunction` that appended each state only once and ignored the multiplicity.

As a result, join duplicates were lost. For example, a joined multiset containing six 1s and four 42s was effectively reduced to [1, 1, 42, 42], producing a median of `21.5` instead of `1.0`.

### Fix

The commit introduces an explicit multiplicity-aware state-combine capability:

- Aggregates must explicitly provide a multiplicity-aware combine callback before double-eager partial aggregate pushdown can use them.
- `combine_aggr(state, multiplicity)` dispatches to this callback and rejects aggregates that do not support multiplicities.
- `ListCombineFunction` now correctly handles multiplicities:
      - `NULL` and zero are no-ops.
      - Negative values raise an error.
      - A multiplicity of one can use the existing destructive merge.
      - Larger multiplicities copy the list state the required number of times.

- Existing executor-backed and list-backed aggregates are registered with the new capability.

### DIFF
--- a/src/function/scalar/system/aggregate_export.cpp
+++ b/src/function/scalar/system/aggregate_export.cpp
@@ -435,6 +435,7 @@ void CombineAggrStateCombine(Vector &source, Vector &target, AggregateInputData 
 	auto &bind_data = aggr_input_data.bind_data->Cast<ExportAggregateBindData>();
 	AggregateInputData combine_input(bind_data.aggr, bind_data.bind_data.get(), aggr_input_data.allocator,
 	                                 aggr_input_data.combine_type);
+	combine_input.combine_multiplicities = aggr_input_data.combine_multiplicities;
 	bind_data.aggr.GetStateCombineCallback()(source, target, combine_input, count);
 }
 
--- a/src/include/duckdb/function/aggregate/list_aggregate.hpp
+++ b/src/include/duckdb/function/aggregate/list_aggregate.hpp
@@ -99,6 +99,22 @@ inline void ListClusterUpdate(Vector inputs[], AggregateInputData &aggr_input_da
 	}
 }
 
+inline void ListAbsorbState(ListAggState &source, ListAggState &target) {
+	if (source.linked_list.total_capacity == 0) {
+		// NULL, no need to append. This can happen with a filtered aggregate.
+		return;
+	}
+	if (target.linked_list.total_capacity == 0) {
+		target.linked_list = source.linked_list;
+		return;
+	}
+
+	// Append the linked list.
+	target.linked_list.last_segment->next = source.linked_list.first_segment;
+	target.linked_list.last_segment = source.linked_list.last_segment;
+	target.linked_list.total_capacity += source.linked_list.total_capacity;
+}
+
 inline void ListAbsorbFunction(Vector &states_vector, Vector &combined, AggregateInputData &aggr_input_data,
                                idx_t count) {
 	D_ASSERT(aggr_input_data.combine_type == AggregateCombineType::ALLOW_DESTRUCTIVE);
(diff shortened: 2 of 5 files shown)


Mytherin:
Thanks but I think we shouldn't expose another callback just for this - can you instead try to fix the callback for aggregates that don't implement it correctly?

Alignyx:
Hi @Mytherin, thank you for your guidance. I’ve fixed the callback for the affected aggregates.

The problem occurred during partial aggregate pushdown. DuckDB first computed an intermediate aggregate state on one side of the join and then called combine_aggr(state, multiplicity) to reproduce the effect of the joined rows. The multiplicity tells the aggregate how many times each intermediate state should contribute to the final result.

The standard aggregate combine callback already supports this through `AggregateInputData::combine_multiplicities`. However, list-backed aggregates such as `LIST`, `MEDIAN`, `QUANTILE`, and `MAD` use a custom `ListCombineFunction`. That function ignored the multiplicity and appended each intermediate state only once. As a result, duplicate rows introduced by the join were lost, producing incorrect aggregate results.

I fixed `ListCombineFunction` so that it now reads and applies the multiplicity passed to the existing combine callback:

- A NULL or zero multiplicity contributes nothing.
- A multiplicity of one keeps the existing efficient destructive-merge path.
- A multiplicity greater than one appends the state’s values the required number of times.
- A negative multiplicity raises an error.

I also fixed the `combine_aggr` wrapper so that it forwards `combine_multiplicities` when it calls the underlying aggregate’s existing combine callback.

Mytherin:
Thanks!

Alignyx:
Hi @Mytherin, could you please review the pr again? The unit tests were updated because the original `EXPLAIN` assertion was configuration-dependent: aggregate-state export intentionally disables partial aggregate pushdown, while a vector size of 2 changed the cost estimate enough to select a different valid plan. The plan assertion was moved to a dedicated optimizer test with stable, high-collapse input, while the issue test continues to verify result correctness across configurations.

Alignyx:
It seems this PR can also fix https://github.com/duckdb/duckdb/issues/24626

Alignyx:
Hi @Mytherin, so sorry that I didn't notice that you marked this as auto-merge, so squash the three commits into a single one. The current one is identical to the previous commits.

Alignyx:
Hi @Mytherin it seems that the CI failure is unrelated to the PR. How can I handle this?

PR #24821, opened 2026-08-17
Fix Parquet interval millisecond overflow

### LINKED ISSUE #24810: Parquet writer corrupts INTERVAL values: milliseconds field overflow due to incorrect days/milliseconds split
### What happens?

When writing an `INTERVAL` value to a Parquet file, DuckDB incorrectly stores the total interval duration (in milliseconds) directly into the Parquet `milliseconds` field (a `uint32`), without properly splitting the portion that exceeds one day into the `days` field. This causes an unsigned 32-bit integer overflow, resulting in data corruption when the Parquet file is read back.

For the interval `580063 hours 44 minutes 54 seconds`, the original value is:

```text
580063:44:54
```

But after a round-trip through Parquet, the value becomes:

```text
243:09:48.144
```

This is not a limitation of the Parquet format; the format can represent such intervals correctly by splitting them into `days` and `milliseconds`. It is a bug in DuckDB’s Parquet writer.

### To Reproduce

Run the following SQL in DuckDB v1.5.4 (or later). It creates a table with an `INTERVAL` column, writes it to a Parquet file, and then reads it back.

```sql
DROP TABLE IF EXISTS repro_duckdb_db1_interval_r;
DROP TABLE IF EXISTS repro_duckdb_db1_interval_l;
DROP TABLE IF EXISTS repro_duckdb_db1_interval_source;

CREATE TABLE repro_duckdb_db1_interval_source (
    vp_rowid BIGINT,
    c0 INTERVAL
);

INSERT INTO repro_duckdb_db1_interval_source VALUES
    (0, INTERVAL '580063 hours 44 minutes 54 seconds');

CREATE TABLE repro_duckdb_db1_interval_l (
    vp_rowid BIGINT
);

INSERT INTO repro_duckdb_db1_interval_l
SELECT vp_rowid
FROM repro_duckdb_db1_interval_source;

CREATE TABLE repro_duckdb_db1_interval_r (
    vp_rowid BIGINT,
    c0 INTERVAL
);

INSERT INTO repro_duckdb_db1_interval_r
SELECT vp_rowid, c0
FROM repro_duckdb_db1_interval_source;

COPY repro_duckdb_db1_interval_r
TO 'repro_duckdb_db1_interval_r.parquet'
(FORMAT PARQUET, COMPRESSION ZSTD);

-- Original source query
SELECT vp_rowid, c0
FROM repro_duckdb_db1_interval_source
ORDER BY vp_rowid;

-- Parquet round-trip query
SELECT l.vp_rowid, p.c0
FROM repro_duckdb_db1_interval_l l
JOIN read_parquet('repro_duckdb_db1_interval_r.parquet') p
  USING (vp_rowid)
ORDER BY l.vp_rowid;
```

Actual results:

```text
Source:  0 | 580063:44:54
VP:      0 | 243:09:48.144
```

Expected result: the Parquet round-trip should return the original value:

```text
Source:  0 | 580063:44:54
VP:      0 | 580063:44:54
```

### Root cause analysis

Parquet’s `INTERVAL` logical type is physically stored as a `fixed_len_byte_array(12)` containing three independent unsigned 32-bit integers:

```text
months       uint32
days         uint32
milliseconds uint32
```

The `milliseconds` field has a maximum value of `2^32 - 1 = 4,294,967,295` milliseconds (about 49.7 days). The `days` field can hold up to `2^32 - 1` days (about 11.7 million years).

For the interval `580063:44:54`:

```text
Total milliseconds = 580063 * 3600 * 1000 + 44 * 60 * 1000 + 54 * 1000
                   = 2,088,229,494,000 ms
```

This value far exceeds the `uint32` maximum. DuckDB’s Parquet writer appears to write the total milliseconds directly into the `milliseconds` field without first splitting full days into the `days` field. The resulting unsigned 32-bit overflow produces:

```text
2,088,229,494,000 mod 4,294,967,296 = 875,388,144 ms
```

And:

```text
875,388,144 ms = 875,388.144 s = 243:09:48.144
```

This exactly matches the observed corrupted value.

The correct Parquet representation for this interval should be:

```text
580063:44:54 = 24169 days + 07:44:54
```

That is:

```text
days         = 24169
milliseconds = 27,894,000   (= 7*3600*1000 + 44*60*1000 + 54*1000)
```

Both fields are well within their valid `uint32` ranges.

Therefore, the bug lies in DuckDB’s Parquet writer: it fails to properly decompose an `INTERVAL` into Parquet’s `months`, `days`, and `milliseconds` fields, and instead places the entire duration into `milliseconds`, causing overflow and silent data corruption.

### OS:

Ubuntu 20.04.6 LTS, x86_64

### DuckDB Version:

v1.5.4

### DuckDB Client:

Java JDBC (also reproducible via DuckDB CLI)

### Hardware:

Intel Xeon Gold 5218R, 80 cores, 1486G RAM

### Full Name:

Annie

### Affiliation:

ECNU

- [x] Yes, I have

- [x] Yes, I have

Yes

idle-lab:
Fixes: #24810

Fix Parquet `INTERVAL` corruption by carrying complete days from the microseconds component before storing the remaining milliseconds as `uint32_t`.

### DIFF
--- a/extension/parquet/include/parquet_interval.hpp
+++ b/extension/parquet/include/parquet_interval.hpp
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "duckdb/common/helper.hpp"
+#include "duckdb/common/numeric_utils.hpp"
 #include "duckdb/common/types/interval.hpp"
 #include "zstd/common/xxhash.hpp"
 
@@ -21,8 +22,8 @@ struct ParquetIntervalUtils {
 		const auto days = int64_t(interval.days) + interval.micros / Interval::MICROS_PER_DAY;
 		const auto micros = interval.micros % Interval::MICROS_PER_DAY;
 		Store<uint32_t>(static_cast<uint32_t>(interval.months), target);
-		Store<uint32_t>(static_cast<uint32_t>(days), target + sizeof(uint32_t));
-		Store<uint32_t>(static_cast<uint32_t>(micros / Interval::MICROS_PER_MSEC), target + sizeof(uint32_t) * 2);
+		Store<uint32_t>(NumericCast<uint32_t>(days), target + sizeof(uint32_t));
+		Store<uint32_t>(NumericCast<uint32_t>(micros / Interval::MICROS_PER_MSEC), target + sizeof(uint32_t) * 2);
 	}
 
 	static uint64_t HashNormalized(const interval_t &interval) {
--- a/extension/parquet/include/parquet_interval.hpp
+++ b/extension/parquet/include/parquet_interval.hpp
@@ -18,10 +18,11 @@ struct ParquetIntervalUtils {
 	static constexpr const idx_t PARQUET_INTERVAL_SIZE = 12;
 
 	static void Encode(const interval_t &interval, data_ptr_t target) {
+		const auto days = int64_t(interval.days) + interval.micros / Interval::MICROS_PER_DAY;
+		const auto micros = interval.micros % Interval::MICROS_PER_DAY;
 		Store<uint32_t>(static_cast<uint32_t>(interval.months), target);
-		Store<uint32_t>(static_cast<uint32_t>(interval.days), target + sizeof(uint32_t));
-		Store<uint32_t>(static_cast<uint32_t>(interval.micros / Interval::MICROS_PER_MSEC),
-		                target + sizeof(uint32_t) * 2);
+		Store<uint32_t>(static_cast<uint32_t>(days), target + sizeof(uint32_t));
+		Store<uint32_t>(static_cast<uint32_t>(micros / Interval::MICROS_PER_MSEC), target + sizeof(uint32_t) * 2);
 	}
 
 	static uint64_t HashNormalized(const interval_t &interval) {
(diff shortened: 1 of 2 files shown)


extension/parquet/include/parquet_interval.hpp
```diff
@@ -18,10 +18,11 @@ struct ParquetIntervalUtils {
 	static constexpr const idx_t PARQUET_INTERVAL_SIZE = 12;
 
 	static void Encode(const interval_t &interval, data_ptr_t target) {
+		const auto days = int64_t(interval.days) + interval.micros / Interval::MICROS_PER_DAY;
+		const auto micros = interval.micros % Interval::MICROS_PER_DAY;
 		Store<uint32_t>(static_cast<uint32_t>(interval.months), target);
-		Store<uint32_t>(static_cast<uint32_t>(interval.days), target + sizeof(uint32_t));
-		Store<uint32_t>(static_cast<uint32_t>(interval.micros / Interval::MICROS_PER_MSEC),
-		                target + sizeof(uint32_t) * 2);
+		Store<uint32_t>(static_cast<uint32_t>(days), target + sizeof(uint32_t));
```
Mytherin: Can we add a check that checks that `days` fits in the `uint32_t` and throw an error if it doesn't?

Mytherin (commented):

extension/parquet/include/parquet_interval.hpp
idle-lab: Thanks for the review!
I've replaced the `static_cast` with `NumericCast`. 

I had originally used `static_cast` because I thought days was bounded (max ≈ 2.25e9 < uint32 max) and negative intervals are already rejected before `Encode` is called.

Mytherin:
Thanks!

PR #24822, opened 2026-08-17
Reset default database after detachment

### LINKED ISSUE #24783: DETACH of the first-attached database leaves default_database dangling; new connections fail with `Catalog "" does not exist!`
### What happens?

`DatabaseManager::default_database` is set to the first database attached (`FinalizeAttach`, `database_manager.cpp:227`). It is never cleared or updated when that database is detached, since `DetachInternal` does not touch it. It keeps holding a name that no longer resolves.

Any connection that then resolves the default catalog fails:

```
Binder Error: Catalog "" does not exist!
```

Some details:

- The connection that ran the `DETACH` keeps working, because its own search path names a database that is still attached. New connections start with an unset search path, fall back to the stale `default_database`, and fail. In an application that uses more than one connection, this looks like the session breaking only partway.
- Plain expressions like `SELECT 1` still work. Anything that needs a catalog or function lookup fails, including `duckdb_databases()`.
- To recover, you have to attach a database under the original name. Attaching one under a different name does not work, and having other databases attached does not help either.
- The error message points at the wrong thing. `Catalog::GetCatalogEntry` substitutes the default catalog when the requested name is empty (`catalog.cpp:92`), but `Catalog::GetCatalog` throws using the requested name instead of the resolved one (`catalog.cpp:107`). The catalog that is actually missing here is `memory`. Reporting the resolved name would make the error much easier to diagnose.
- `Catalog::LookupEntry` already handles this case. Both `catalog.cpp:846` and `catalog.cpp:965` carry the comment "Check if the default database is actually attached", so a dangling default is expected there. `GetCatalog` has no equivalent check.

A possible fix: clear or update `default_database` in `DetachInternal` when it names the database being detached, the same way renames are handled at `database_manager.cpp:321`.

### To Reproduce

CLI, no extensions:

```sql
ATTACH ':memory:' AS other;
USE other;          -- required: DETACH refuses to drop the calling context's default
DETACH memory;      -- `memory` was the first database attached, so it is the manager default
RESET search_path;  -- puts this connection into the state a NEW connection starts in
SELECT database_name FROM duckdb_databases();
-- Binder Error: Catalog "" does not exist!
```

`RESET search_path` is not part of the scenario a user would hit. It is only there to reach, inside a single CLI session, the state that every new connection starts in. The original report came from an application whose background thread opens a connection on each poll.

Both of the following succeed, which shows the problem needs both steps and is not caused by either one alone:

```sql
-- RESET without the DETACH
ATTACH ':memory:' AS other; USE other; RESET search_path;
SELECT database_name FROM duckdb_databases();

-- DETACH without the RESET (search path still names `other`)
ATTACH ':memory:' AS other; USE other; DETACH memory;
SELECT database_name FROM duckdb_databases();
```

The problem is a dangling name, not an empty database list. Attaching a database under a different name does not fix it:

```sql
ATTACH ':memory:' AS other; USE other; DETACH memory;
ATTACH ':memory:' AS zzz;
RESET search_path;
SELECT database_name FROM duckdb_databases();
-- Binder Error: Catalog "" does not exist!
```

Replace `zzz` with `memory` and it succeeds.

First reported on v1.5.2 (Windows) through the UI extension as duckdb/duckdb-ui#234, where a background connection created on each poll produced repeating `Catalog "" does not exist!` errors and the UI could not list attached databases. `DetachInternal` is unchanged on `main`, so 2.0.0 is affected too.

### OS:

macOS 15 (arm64)

### DuckDB Version:

v1.5.5 (d8cdaa33fd)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Jeff Raymakers

### Affiliation:

MotherDuck

- [x] Yes, I have

- [x] Yes, I have

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24783

In the current impl, default database becomes dangling reference after its detachment; this PR resets default if there're still valid databases.

### DIFF
--- a/src/main/database_manager.cpp
+++ b/src/main/database_manager.cpp
@@ -345,6 +345,9 @@ shared_ptr<AttachedDatabase> DatabaseManager::DetachInternal(const Identifier &n
 		}
 		attached_db = std::move(entry->second);
 		databases.erase(entry);
+		if (name == default_database) {
+			default_database = databases.empty() ? Identifier() : databases.begin()->first;
+		}
 	}
 	if (attached_db && attached_db->GetCatalog().Supports(RemoteCapability::IS_REMOTE)) {
 		--remote_catalog_count;
--- a/test/sql/attach/detach_manager_default_database.test
+++ b/test/sql/attach/detach_manager_default_database.test
@@ -1,0 +1,22 @@
+# name: test/sql/attach/detach_manager_default_database.test
+# description: Test detaching the database manager's default database
+# group: [attach]
+
+require skip_reload
+
+require noforcestorage
+
+statement ok con1
+ATTACH ':memory:' AS other;
+
+statement ok con1
+USE other;
+
+statement ok con1
+DETACH memory;
+
+# Connections without an explicit search path should use a remaining database
+query T con2
+SELECT database_name FROM duckdb_databases() WHERE database_name = 'other';
+----
+other
--- a/test/sql/attach/detach_manager_default_database.test
+++ b/test/sql/attach/detach_manager_default_database.test
@@ -2,21 +2,22 @@
 # description: Test detaching the database manager's default database
 # group: [attach]
 
-require skip_reload
-
 require noforcestorage
 
-statement ok con1
+statement ok
 ATTACH ':memory:' AS other;
 
-statement ok con1
+statement ok
 USE other;
 
-statement ok con1
+statement ok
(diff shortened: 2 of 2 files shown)


Mytherin:
Thanks - I think we should probably get rid of `default_database` entirely, though. I'm pretty sure this is a relic and not actually useful anymore.

PR #24823, opened 2026-08-17
Publish linked extensions on the config, so an extension can load them too

carlopi:
Basically: currently LoadExtension would encode state of compile units.

Currently, let's say you build core + parquet, json and core_functions linked in + icu as loadable extension (but linking in static duckdb)

core, parquet, json and core_functions would agree in the fact 3 functions are already loaded.
icu would be aware only of himself.

This is problematic since depending on code-path that call to `LoadExtensions` takes different results will ensure. This is not great, and worked around in a couple of places.

Solution: once extensions are linked in, they will respond with the actual set of statically loaded extensions from DatabaseInstance, not the set they have backed-in at compile time.

---

ExtensionHelper::LoadExtension was generated code: CMake expanded an if-chain naming every statically linked extension, and a build linking none got a stub returning `NOT_LOADED`. That makes the capability set a property of one binary's link line, invisible to anyone holding a different copy of DuckDB - so any extension calling LoadExtension silently got `NOT_LOADED` for extensions the host had built in.

Publish the set as data instead. DBConfig carries a vector of {name, loader}, the generated code registers into it rather than defining a lookup, and LoadExtension becomes ordinary core code doing the lookup - one implementation shared by a binary that links extensions and one that links none. The dummy loader no longer stubs LoadExtension out; it only reports that it registers nothing, so a binary linking it can still load whatever it was handed.

DatabaseInstance::Configure carries the set over from a caller-supplied config, which is what lets a database created by an extension inherit the extensions of the binary that created it.

A vector rather than a map: extensions load in registration order, which is the order they are listed in the build config, and that has to stay stable across runs and platforms.

### DIFF
--- a/extension/CMakeLists.txt
+++ b/extension/CMakeLists.txt
@@ -44,7 +44,7 @@ configure_file(
 
 # generated_extension_loader.hpp
 set(EXT_LOADER_NAME_LIST "")
-set(EXT_LOADER_BODY "")
+set(EXT_REGISTER_BODY "")
 set(EXT_CAPI_DECLARATIONS "")
 foreach(EXT_NAME IN LISTS STATICALLY_LINKED_EXTENSIONS)
   string(TOUPPER ${EXT_NAME} EXT_NAME_UPPERCASE)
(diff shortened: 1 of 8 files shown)


carlopi:
This goes one step further on fixing https://github.com/duckdb/duckdb/issues/24799

Mytherin:
Thanks!

PR #24824, opened 2026-08-17
A fix to common_subplan_optimizer

### LINKED ISSUE #24781: common_subplan merges scans of different attached-SQLite tables with identical column layouts, returning one table's rows for both

staticlibs:
This is a continution to the PR #24782, see detailed description in the linked issue.

The problem can be reproduced with `duckdb-sqlite` - duckdb/duckdb-sqlite#215

Fixes: #24781

### DIFF
--- a/src/optimizer/common_subplan_optimizer.cpp
+++ b/src/optimizer/common_subplan_optimizer.cpp
@@ -568,7 +568,6 @@ private:
 		case LogicalOperatorType::LOGICAL_TOP_N:
 		case LogicalOperatorType::LOGICAL_DISTINCT:
 		case LogicalOperatorType::LOGICAL_PIVOT:
-		case LogicalOperatorType::LOGICAL_GET:
 		case LogicalOperatorType::LOGICAL_EXPRESSION_GET:
 		case LogicalOperatorType::LOGICAL_DUMMY_SCAN:
 		case LogicalOperatorType::LOGICAL_COMPARISON_JOIN:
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks!

PR #24836, opened 2026-08-18
Prune map extraction row group

dentiny:
Hi team, I found for map operation, stats are not fully leveraged to prune row groups, for both DuckDB native format and parquet.
```sql
-- should access only one row group
memory D CREATE TABLE mymaps AS
         SELECT MAP {'k': i} AS m
         FROM range(122880 * 3) tbl(i);
memory D EXPLAIN ANALYZE SELECT count(*) FROM mymaps WHERE m['k'] BETWEEN 2900 AND 3100;
╭─ Summary ───────────╮
│ Total Time: 0.0047s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ───────────────────────────╮
│ 1 row                                       0µs │
╰────────────────────────┒┎───────────────────────╯
╭─ Table Scan ───────────┚┖───────────────────────╮
│ Table: memory.main.mymaps                       │
│ Type: Sequential Scan                           │
│ Filters:                                        │
│ map_extract_value(m, 'k') BETWEEN 2900 AND 3100 │
│ Row Groups Scanned: 3 / 3                       │
│ 201 rows                                 10.0ms │
╰─────────────────────────────────────────────────╯

-- should access only one row group
memory D COPY (SELECT MAP {'k': i} AS m FROM range(6144) t(i))
           TO '/tmp/pd_map.parquet' (FORMAT PARQUET, ROW_GROUP_SIZE 2048);
memory D EXPLAIN ANALYZE SELECT count(*) FROM '/tmp/pd_map.parquet'
         WHERE m['k'] BETWEEN 2900 AND 3100;
╭─ Summary ───────────╮
│ Total Time: 0.0052s │
│ Data Read:  41.2 kB │
╰─────────────────────╯
╭─ Ungrouped Aggregate ───────────────────────────╮
│ Aggregates: count_star()                        │
│ 1 row                                       0µs │
╰────────────────────────┒┎───────────────────────╯
╭─ Table Scan ───────────┚┖───────────────────────╮
│ Function: PARQUET_SCAN                          │
│ Filters:                                        │
│ map_extract_value(m, 'k') BETWEEN 2900 AND 3100 │
│ Total Files Read: 1                             │
│ Filename(s): /tmp/pd_map.parquet                │
│ Row Groups Scanned: 3 / 3                       │
│ 201 rows                                    0µs │
╰─────────────────────────────────────────────────╯
```
In this PR, I made two changes:
- Propagate parquet stats -- DuckDB treats map list<struct<key, value>> -- so we handle it as the same way as list type; DuckDB native format already has the stats 
- Implement the stats callback for map value extraction function, so it expression filter could leverage the value min/max to prune

### DIFF
--- a/extension/core_functions/scalar/map/map_extract.cpp
+++ b/extension/core_functions/scalar/map/map_extract.cpp
@@ -4,6 +4,8 @@
 #include "duckdb/function/scalar/list/contains_or_position.hpp"
 #include "duckdb/function/scalar/nested_functions.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/storage/statistics/list_stats.hpp"
+#include "duckdb/storage/statistics/struct_stats.hpp"
 
 namespace duckdb {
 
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks!

PR #24837, opened 2026-08-18
Remove default database

dentiny:
Followup PR for https://github.com/duckdb/duckdb/pull/24822#issuecomment-5324718792, which is a better solution for the initial dangling reference bug.

In one word, `default_database` is a redundant field, which could be replaced completely from attached databases; removing which saves a bunch of maintenance overhead (i.e., detach a DB should update to another attached one).

Regression test has been added in the above-mentioned PR.


Mytherin:
Thanks!

Mytherin:
Ah seems there's some issues around extensions - could you have a look

PR #24838, opened 2026-08-18
[Dev][Parser] Move `Matcher` implementations + `MatcherFactory` to separate files

Tishj:
needs #24753 

Clean up the `matcher.cpp` file, making the `Matcher` implementations public so they can be derived from.

Also moves the `MatcherFactory` to its own header+source for the same purpose, to be eventually derived from.


Mytherin:
Thanks!

PR #24839, opened 2026-08-18
Use Identifier for CSV/JSON column names in the COPY FROM bind path

JelteF:
`copy_from_bind_t`'s `expected_names` and the `name_list` it feeds are now `vector<Identifier>` instead of `vector<string>`, removing the conversions that surrounded every copy bind now that `table_function_bind_t` and `MultiFileBindInternal` already take identifiers.

Note that `sniff_csv`'s `names_csv` and `return_types_csv` are dropped rather than converted: nothing has assigned them since the options started flowing through `FromNamedParameters()` into `name_list` directly, so both reads only ever copied an empty vector over an empty one.


Mytherin:
Thanks!

PR #24840, opened 2026-08-18
Clean up files created by `COPY` if query aborts

lnkuiper:
We currently leave these files behind. While files that were finalized are readable, it's not very transactional to leave them there. Files that were not finalized are not readable at all, and should definitely be removed. I've built some infrastructure for this that is shared between both file copying operators.


src/common/file_system.cpp
```diff
@@ -599,6 +607,10 @@ bool FileSystem::TryRemoveFile(const string &filename, optional_ptr<FileOpener>
 	return false;
 }
 
+bool FileSystem::TryRemoveEmptyDirectory(const string &directory, optional_ptr<FileOpener> opener) {
```
Mytherin: Maybe this can be flags, otherwise we will end up with `TryRemoveNonEmptyDirectory` soon :)

src/common/file_system.cpp
```diff
@@ -500,6 +500,14 @@ void FileSystem::CreateDirectory(const string &directory, optional_ptr<FileOpene
 	throw NotImplementedException("%s: CreateDirectory is not implemented!", GetName());
 }
 
+bool FileSystem::CreateDirectoryIfNotExists(const string &directory, optional_ptr<FileOpener> opener) {
+	if (DirectoryExists(directory, opener)) {
+		return false;
+	}
+	CreateDirectory(directory, opener);
+	return false;
```
Mytherin: Should this be `return true`?

src/common/file_system.cpp
```diff
@@ -500,6 +500,14 @@ void FileSystem::CreateDirectory(const string &directory, optional_ptr<FileOpene
 	throw NotImplementedException("%s: CreateDirectory is not implemented!", GetName());
 }
 
+bool FileSystem::CreateDirectoryIfNotExists(const string &directory, optional_ptr<FileOpener> opener) {
```
Mytherin: Maybe this can also be a flag on e.g. `CreateDirectoryExtended` (maybe also unifying with `CreateDirectoriesRecursive`)? I've done some work on adding new overloads for functions before with `ListFiles` and `OpenFile` that this can be modeled off of.

src/common/local_file_system.cpp
```diff
@@ -714,19 +714,23 @@ bool LocalFileSystem::DirectoryExists(const string &directory, optional_ptr<File
 }
 
 void LocalFileSystem::CreateDirectory(const string &directory, optional_ptr<FileOpener> opener) {
-	struct stat st;
+	CreateDirectoryIfNotExists(directory, opener);
```
Mytherin: Is this exposing the same behavior as `CreateDirectoryIfNotExists`? That seems a bit odd.

src/execution/operator/persistent/copy_output_lifecycle.cpp
```diff
@@ -0,0 +1,149 @@
...
+	vector<string> committed_files;
+	vector<string> directories;
+	{
+		lock_guard<mutex> guard(lock);
+		if (successful) {
+			return;
+		}
+		for (auto &file : files) {
+			if (file.committed) {
+				committed_files.push_back(std::move(file.path));
```
Mytherin: Could you expand on why we're only trying to clean-up committed files (presumably writers clean-up non finalized files themselves?)

src/execution/operator/persistent/physical_copy_to_file.cpp
```diff
@@ -3385,8 +3355,22 @@ void CopyToFileGlobalState::ScheduleFileStateOpen(PendingFileStateOpen pending_f
...
 		    open_job, CopyFileLifecycleWaitMode::INTERRUPTIBLE,
 		    [this, open_job, pending_file_state = std::move(pending_file_state_open.pending_file_state)]() mutable {
+			    if (aborting) {
+				    open_job->CompleteAborted();
+				    return;
+			    }
 			    EnsureOutputDirectoryReady();
-			    open_job->Complete(InitializeFileState(std::move(pending_file_state)));
+			    if (aborting) {
+				    open_job->CompleteAborted();
```
Mytherin: This seems to be added in many different places - might be a less error prone way of doing this?

Mytherin (commented):
Thanks for the PR!

Good to tackle this. I left some comments below.

I think in general this seems to be a lot harder to do / require much more code to do because the writers themselves own the async writers.

Maybe a cleaner solution is to lift the async writer management outside of the individual copy methods and into the copy operator? i.e. not have `ParquetWriter`, `CSVWriter`, etc own the `AsyncFileWriter`, but have the `AsyncFileWriter` and other file-writing lifetimes live in the `PhysicalCopyToFile` and have that be passed into the callback for writing specific files?

Moving all of that to a centralized place might make other code that manages the async writes easier to write as well, since co-ordination between different writers can all happen in a central location instead of requiring various callbacks to be copy pasted for every single format we introduce.

src/execution/operator/persistent/copy_output_lifecycle.cpp
lnkuiper: Exactly, the writers themselves need to clean up non-committed files otherwise we have to go through the process of committing them first before we can clean them up

lnkuiper:
I've thought about this a bit and was able to get the behaviour we want in a much less invasive way

carlopi:
Hey @lnkuiper, I took the part relevant to `RemoveDirectoryExtended` and sent a PR at https://github.com/duckdb/duckdb/pull/25012 (same as your code, so merge should be clean, BUT for renaming `SINGLE` to `IF_EMPTY`). That's needed since there is the possibility that needs to be back-ported.

PR #24844, opened 2026-08-18
[Dev][Parser] Put the `PEGKeywordHelper` into the `CompiledGrammar`

Tishj:
needs #24838 

The `CompiledGrammar` (renamed from `PEGMatcher`) should hold all the items required for parsing.

To facilitate making keywords extendable in follow up PR(s), the `PEGKeywordHelper` is now part of the `CompiledGrammar`.

`PEGKeywordHelper` becomes an interface, with currently only 1 implementation: `DuckDBKeywordHelper`, which is a singleton.

Some operations don't have a `ClientContext` or equivalent context, and will need to keep using the `DuckDBKeywordHelper` directly.

### Planned Follow Up

Giving the `Tokenizer` the same treatment, letting it be owned by the `CompiledGrammar`.


src/include/duckdb/parser/peg/compiled_grammar.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include "duckdb/parser/peg/transformer/peg_transformer.hpp"
+
+namespace duckdb {
+
+struct CompiledGrammar {
+public:
+	CompiledGrammar();
+
+public:
+	Matcher &ProgramMatcher() {
```
Mytherin: Given this is shared should these return `const Matcher &`?

src/parser/peg/compiled_grammar.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "duckdb/main/database.hpp"
+
+namespace duckdb {
+
+CompiledGrammar::CompiledGrammar() : keyword_helper(DuckDBKeywordHelper::Instance()) {
+}
+
+shared_ptr<CompiledGrammar> CompiledGrammar::Get(ClientContext &context) {
+	auto &db = DatabaseInstance::GetDatabase(context);
+	return CompiledGrammar::Get(db);
```
Mytherin: It's probably not going to matter that much (at least for now) but this requires atomic operations on the reference count of the compiled grammar - also when the grammar is unchanged.

Maybe a nicer approach would be to store the `shared_ptr<CompiledGrammar>` in the `ClientConfig`, and then doing an (atomic) version check to see if that is still the relevant version? Then this can return a `CompiledGrammar &` instead of a `shared_ptr<CompiledGrammar>`

src/parser/peg/compiled_grammar.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+shared_ptr<CompiledGrammar> CompiledGrammar::Get(DatabaseInstance &db) {
```
Mytherin, line 32: I wonder if we want this at all, or if we want to require a `ClientContext` to get the compiled grammar?

src/parser/peg/keyword_helper/duckdb_keyword_helper.cpp
```diff
@@ -0,0 +1,66 @@
+#include "duckdb/parser/peg/keyword_helper/duckdb_keyword_helper.hpp"
+
+namespace duckdb {
+
+DuckDBKeywordHelper::DuckDBKeywordHelper() : initialized(false) {
+	InitializeKeywordMaps();
+}
+
+const DuckDBKeywordHelper &DuckDBKeywordHelper::Instance() {
+	static DuckDBKeywordHelper instance;
```
Mytherin, line 10: This method should probably be removed going forward, but maybe better in a follow-up PR

Mytherin (commented):

src/parser/peg/compiled_grammar.cpp
Mytherin: e.g. that could look like this:

```cpp
CompiledGrammar &CompiledGrammar::Get(ClientContext &context) {
	auto &client_config = ClientConfig::GetConfig(context);
	auto &cache = db.GetParserCache();
	if (!client_config.cached_grammar || client_config.cached_grammar->Version() != cache.LatestParserVersion()) {
		client_config.cached_grammar = cache.GetCompiledGrammar();
	}
	return *client_config.cached_grammar;
}
```

Tishj: I mentioned in the internal issue that we likely want to find a better way to guard access to them, I think this is a good suggestion.

src/parser/peg/compiled_grammar.cpp
Tishj: As mentioned here: https://github.com/duckdblabs/duckdb-internal/issues/10346#issuecomment-5325423278

Some paths deliberately don't have a client context, but still need to be able to work with the defaults.

src/parser/peg/keyword_helper/duckdb_keyword_helper.cpp
Tishj: I would expect that comment on `InitializeKeywordMaps`, but not on the `Instance` :eyes:

The default DuckDB grammar is unchanged throughout execution, so the places that don't respond to changes (no client context or db instance) can cheaply use this without recompiling the keyword helper.

Mytherin: I think we should never be using a parser without a database instance / client connection going forward

Issue #24256, opened 2026-07-29
&& stats pruning returns NULL geometry rows


dentiny:
### What happens?

See repro for details.

### To Reproduce

```sql
memory D COPY (
           SELECT
             CASE
               WHEN i % 10 = 0 THEN NULL
               ELSE printf('POINT(%d %d)', i % 100, i % 100)::GEOMETRY
             END AS g,
             i AS id
           FROM range(1000) tbl(i)
         ) TO '/private/tmp/duckdb_geo_nullable_codex_019fac4c.parquet'
           (FORMAT PARQUET, GEOPARQUET_VERSION 'V2', ROW_GROUP_SIZE 1000);
memory D SELECT count(*)
         FROM '/private/tmp/duckdb_geo_nullable_codex_019fac4c.parquet'
         WHERE g && 'POLYGON((-1 -1, 200 -1, 200 200, -1 200, -1 -1))'::GEOMETRY;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│         1000 │
└──────────────┘

-- with and without optimizer leads to different results
memory D SET disabled_optimizers='filter_pushdown';
memory D SELECT count(*)
         FROM '/private/tmp/duckdb_geo_nullable_codex_019fac4c.parquet'
         WHERE g && 'POLYGON((-1 -1, 200 -1, 200 200, -1 200, -1 -1))'::GEOMETRY;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│          900 │
└──────────────┘
```

### OS:

macos

### DuckDB Version:

f91908958d7c7d0ac6161c02e7b593900ae6eaa6

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

dentiny:
Max fixed it in the v1.5 branch, think it will be ported to main branch before v2.0 release, close the issue.

PR #24845, opened 2026-08-18
fix geometry row group pruning with nulls/empty (on v1.5)

Maxxen:
Closes https://github.com/duckdb/duckdb/issues/24256


Mytherin:
Thanks!

Issue #22096, opened 2026-04-15
TopN Optimizer Unequal types (TIMESTAMP WITH TIME ZONE != VARCHAR)


hawkfish:
> Hi @hawkfish,
> I am still getting a similar error on a0a08d511d with a different query. Is this a related issue?
> 
> ```sql
> CREATE TABLE t1 (a VARCHAR, b BOOLEAN, c TIMESTAMPTZ);
> CREATE TABLE t2 (a VARCHAR, b BOOLEAN, c VARCHAR);
> INSERT INTO t2 VALUES ('x', false, '2024-01-01');
> 
> INSERT INTO t1
> WITH cte AS (
>     SELECT a, b, c::TIMESTAMPTZ AS c
>     FROM t2
> )
> SELECT * FROM cte QUALIFY ROW_NUMBER() OVER (PARTITION BY a ORDER BY c DESC) = 1;
> ```
> 
> ```
> INTERNAL Error:
> Failed to bind column reference "" [16.2]: inequal types (TIMESTAMP WITH TIME ZONE != VARCHAR)
> 
> Stack Trace:
> 
> 0        duckdb::Exception::Exception(duckdb::ExceptionType, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 268
> 1        duckdb::InternalException::InternalException(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&) + 36
> 2        duckdb::InternalException::InternalException<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, unsigned long long&, unsigned long long&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, unsigned long long&, unsigned long long&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>&&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>&&) + 268
[...]

PR #24846, opened 2026-08-18
Fix top_n_window_elimination late materialization corrupting try-cast and nested-cast payloads

### LINKED ISSUE #24609: TopNWindowElimination late-materialization drops TRY_CAST NULL-semantics and collapses nested casts (regression on main vs fixed v1.5.4/v1.5.5)
## Summary
The `top_n_window_elimination` optimizer, when it takes the rowid late-materialization (SEMI-JOIN) path, reconstructs a
`TRY_CAST` payload as a **strict, throwing** `CAST`, and collapses nested cast chains. This produces two wrong behaviours:

1. **`TRY_CAST` → `Conversion Error`**: a query that should return `NULL` for un-castable winning rows instead aborts.
2. **Nested casts collapsed → silent wrong values**: `CAST(CAST(x AS INTEGER) AS DOUBLE)` loses the inner truncation and
   returns the raw value, with **no error**.

This was **fixed on the v1.5.4 / v1.5.5 release branches** (a stricter variant of #22892 requiring a direct column ref), but
**`main` regresses it** — its variant of #22892 re-admits casts as "recreatable", so the next feature release (2.0) would
re-ship both defects.

## Versions affected (empirical, differential vs `SET disabled_optimizers='top_n_window_elimination'`)
| Shape | v1.5.3 | v1.5.5 | main nightly (v2.0.0-alpha35572, 2026-07-24) |
|---|---|---|---|
| `TRY_CAST` below window, ≥2 payload cols | **Conversion Error** | OK | **Conversion Error** |
| Nested strict casts below window | **silent wrong value** | OK | **silent wrong value** |
| Inline `TRY_CAST` in select list (single payload) | OK | OK | OK |

## Reproduction (plain `duckdb` CLI, self-contained)
```sql
-- Repro 1: TRY_CAST NULL semantics -> Conversion Error
CREATE TABLE t (key VARCHAR, ts INTEGER, val VARCHAR, note VARCHAR);
INSERT INTO t VALUES ('a',1,'100','n1'), ('a',2,'oops','n2'), ('b',1,'7','n3'), ('b',2,'8','n4');
-- Expected (returned with the optimizer disabled): a|NULL|n2, b|8|n4
-- Actual on v1.5.3 / main nightly: Conversion Error: Could not convert string 'oops' to INT32
SELECT key, v, note FROM (
  SELECT key, ts, TRY_CAST(val AS INTEGER) AS v, note FROM t
)
QUALIFY row_number() OVER (PARTITION BY key ORDER BY ts DESC) = 1
ORDER BY key;

-- Repro 2: nested casts -> silent wrong values
CREATE TABLE d (key VARCHAR, ts INTEGER, x DOUBLE, note VARCHAR);
INSERT INTO d VALUES ('a',1,1.1,'n1'), ('a',2,3.7,'n2'), ('b',1,2.2,'n3'), ('b',2,8.9,'n4');
-- Expected: a|4.0|n2, b|9.0|n4.  Actual on v1.5.3 / main nightly: a|3.7|n2, b|8.9|n4 (no error!)
SELECT key, r, note FROM (
  SELECT key, ts, CAST(CAST(x AS INTEGER) AS DOUBLE) AS r, note FROM d
)
QUALIFY row_number() OVER (PARTITION BY key ORDER BY ts DESC) = 1
ORDER BY key;

-- Control: both correct after
--   SET disabled_optimizers = 'top_n_window_elimination';
```
**Trigger requirements:** ≥2 payload columns (the `struct_pack` late-materialization path — single-payload shapes never
fire it), the cast computed **below** the window (view / subquery / non-materialized CTE), base-table scan supporting
rowid late materialization.

## Root cause (source read)
`src/optimizer/topn_window_elimination.cpp`:
- `CanUseLateMaterialization` gates payload eligibility at `IsRecreatableByLateMaterialization` (main), which **strips all
  casts including `TRY_CAST`** and accepts any cast-chain over a column ref.
- `UpdateTopmostBindings` then rebuilds the payload as `BoundCastExpression::AddDefaultCastToType(expr, target)` — a
  **default (throwing) cast** (`try_cast=false`), applied only when the raw column type differs from the output type. So the
  `TRY_` is lost (→ throw), and an equal-endpoint nested chain becomes no cast at all (→ silent wrong value).

The v1.5.x release branches use the stricter `ExtractSingleBinding(..., require_direct_column_ref=true)`, which blocks
non-column payloads from late materialization and carries the computed value through the `arg_min/arg_max` aggregate
instead — empirically correct in every shape tested. #22892's **main** commit (`f53b1cb245`) deliberately kept casts
eligible; that is the regression.

## Related (distinct)
- #22849 / PR #22892 — same family (`parse_filename` payload dropped); the fix **excludes plain CAST/TRY_CAST**.
- #23588 — late-materialization duplicate-row collapse (same pass, different symptom).
- #23925 — TopNWindowElimination assertion crash (open).
- PR #24080 — TopNWindowElimination fixes (does **not** touch CAST/TRY_CAST semantics).

## Workaround
`SET disabled_optimizers = 'top_n_window_elimination';` (per-connection; removes both the correctness defect and the
degenerate rewritten plan).

## Fix
A candidate ~6-line patch to `IsRecreatableByLateMaterialization` (reject TRY_CAST + multi-cast chains, keep the single
plain cast) restores the release-branch behaviour on main. Happy to open a PR if useful.

alwaysprince05:
Fixes #24609

### Summary
When `top_n_window_elimination` takes the rowid late-materialization path, payload eligibility was gated by `IsRecreatableByLateMaterialization`, which stripped *any* cast chain and accepted everything over a column reference. The payload is then rebuilt by the rowid semi-join as the raw base-table column, and `UpdateTopmostBindings` re-applies at most one **default (strict)** cast. Consequently:

1. **TRY_CAST → Conversion Error**: `TRY_CAST(val AS INTEGER)` was rebuilt as a strict `CAST`, so an un-castable winning row aborted instead of returning NULL.
2. **Nested casts collapsed → silent wrong values**: `CAST(CAST(x AS INTEGER) AS DOUBLE)` was rebuilt without the intermediate truncation, returning the raw value with no error (3.7 instead of 4.0).

The v1.5.4/v1.5.5 release branches used a stricter gate (direct column ref only) and were unaffected; main's variant of #22892 (`f53b1cb245`) re-admitted cast chains, regressing this.

### Changes Made
- `src/optimizer/topn_window_elimination.cpp`: restrict `IsRecreatableByLateMaterialization` to a plain column ref optionally wrapped in **at most one non-try cast** — the only shapes the topmost projection reconstructs exactly. Try-cast and multi-cast payloads now fall back to carrying the computed value through the aggregate.
- `test/sql/optimizer/topn_window_elimination_late_materialization.test`: regression cases for both shapes — EXPLAIN checks (window still eliminated, but no rowid semi-join), result checks, and differential checks with the optimizer disabled.

### Notes
- Single plain casts over a base column remain late-materialization eligible (#22096 behaviour preserved by the pre-existing test at the bottom of the test file).
- Verified locally: `test/sql/optimizer/*` (74 cases), `test/sql/topn/*` (21), `test/sql/window/*` (82) all pass.

### DIFF
--- a/src/optimizer/topn_window_elimination.cpp
+++ b/src/optimizer/topn_window_elimination.cpp
@@ -187,8 +187,7 @@ bool IsRecreatableByLateMaterialization(const Expression &expr) {
 	if (expr.GetExpressionClass() == ExpressionClass::BOUND_COLUMN_REF) {
 		return true;
 	}
-	if (BoundCastExpression::IsCast(expr) &&
-	    !BoundCastExpression::IsTryCast(expr.Cast<BoundFunctionExpression>())) {
+	if (BoundCastExpression::IsCast(expr) && !BoundCastExpression::IsTryCast(expr.Cast<BoundFunctionExpression>())) {
 		return BoundCastExpression::Child(expr.Cast<BoundFunctionExpression>()).GetExpressionClass() ==
 		       ExpressionClass::BOUND_COLUMN_REF;
 	}
--- a/src/optimizer/topn_window_elimination.cpp
+++ b/src/optimizer/topn_window_elimination.cpp
@@ -177,16 +177,22 @@ LHSColumnInfo GetLHSColumnInfo(const unique_ptr<LogicalOperator> &op, idx_t colu
 	return result;
 }
 
-//! Late materialization recreates a payload by re-scanning the base table column and re-applying a type cast at the
-//! topmost projection. That only reproduces the original value when the projection expression is a plain column
-//! reference, optionally wrapped in casts. A value-transforming expression (e.g. parse_filename(col)) cannot be
-//! rebuilt that way, so it must not be eligible for late materialization.
+//! Late materialization recreates a payload by re-scanning the base table column and re-applying a single default
+//! (strict) cast at the topmost projection. That only reproduces the original value when the projection expression
+//! is a plain column reference, optionally wrapped in a single non-try cast. A value-transforming expression
+//! (e.g. parse_filename(col)) cannot be rebuilt that way. A try cast would be reconstructed as a strict cast
+//! (losing its NULL semantics), and a deeper cast chain would lose its intermediate conversions, so none of those
+//! shapes are eligible for late materialization.
 bool IsRecreatableByLateMaterialization(const Expression &expr) {
-	reference<const Expression> current = expr;
-	while (BoundCastExpression::IsCast(current.get())) {
-		current = BoundCastExpression::Child(current.get().Cast<BoundFunctionExpression>());
+	if (expr.GetExpressionClass() == ExpressionClass::BOUND_COLUMN_REF) {
+		return true;
 	}
-	return current.get().GetExpressionClass() == ExpressionClass::BOUND_COLUMN_REF;
+	if (BoundCastExpression::IsCast(expr) &&
+	    !BoundCastExpression::IsTryCast(expr.Cast<BoundFunctionExpression>())) {
+		return BoundCastExpression::Child(expr.Cast<BoundFunctionExpression>()).GetExpressionClass() ==
+		       ExpressionClass::BOUND_COLUMN_REF;
+	}
+	return false;
 }
 
 } // namespace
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24847, opened 2026-08-18
Fix incorrect filtered MAD window results by correctly counting rows across excluded subframes

### LINKED ISSUE #24442: MAD FILTER over an excluded ROWS frame can include filtered/excluded rows
### What happens?

`MAD(x) FILTER (...)` used as a window aggregate can return a different result from the equivalent ordinary aggregate when the window frame uses `EXCLUDE CURRENT ROW`.

This happens on current main (5bb2025d2e) and also on v1.5.4.

### To Reproduce

```sql
WITH src(id, x, keep) AS (
  VALUES
    (1, 0, true),
    (2, 100, true),
    (3, 10, true),
    (4, 20, false),
    (5, 30, true)
),
win AS (
  SELECT
    id,
    MAD(x) FILTER (WHERE keep) OVER (
      ORDER BY id
      ROWS BETWEEN 3 PRECEDING AND 1 FOLLOWING
      EXCLUDE CURRENT ROW
    ) AS actual
  FROM src
),
expected AS (
  SELECT
    o.id,
    (
      SELECT MAD(i.x)
      FROM src AS i
      WHERE i.id BETWEEN o.id - 3 AND o.id + 1
        AND i.id <> o.id
        AND i.keep
    ) AS expected
  FROM src AS o
)
SELECT
  id,
  actual,
  expected
FROM win
JOIN expected USING (id)
WHERE actual IS DISTINCT FROM expected
ORDER BY id;
```

Run:

```sh
```

#### Expected

| expected |
| --- |

The comparison query should return no rows. For id = 4, the frame covers ids 1 through 5, excludes the current row id = 4, and keeps only rows with`keep = true`, so MAD is computed over x values 0, 100, 10, and 30.

#### Actual

| id | actual | expected |
| --- | --- | --- |
| 4 | 20.0 | 15.0 |

#### Notes From Minimization

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main(3958a013ed)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

Alignyx:
Fix #24442 

### Problem 

The issue affected `MAD` window calculations when a `FILTER` or nullable input was combined with an `EXCLUDE` clause. Exclusion can split a logical window frame into multiple disjoint subframes, but the optimized `MAD` implementation incorrectly counted the entire span between the first and last subframe, including the excluded gap.

For example, the reported row should calculate `MAD` over `[0, 100, 10, 30]`. Its median is `20`, and the absolute deviations are `[20, 80, 10, 10]`, so the correct `MAD` is `15`; however, DuckDB returned `20`.

The root cause was in the optimized incremental `MAD` path used by `debug_window_mode = 'window'`. It reuses row indexes between adjacent frames, but because the frame size was overestimated, it also processed stale entries left in the reusable buffer. A stale valid index could then be treated as part of the current frame, duplicating one row or displacing another and producing an incorrect result. The combine and separate modes were correct because they use different aggregation paths and do not rely on this faulty bookkeeping.

### Fix

The fix changes the frame count from the width of the enclosing span:

`frames.back().end - frames.front().start`

to the exact number of rows contained in all subframes:

`FrameSet(frames).Size()`

This prevents excluded gaps and stale indexes from being counted.

### DIFF
--- a/extension/core_functions/aggregate/holistic/mad.cpp
+++ b/extension/core_functions/aggregate/holistic/mad.cpp
@@ -238,7 +238,7 @@ struct MedianAbsoluteDeviationOperation : QuantileOperation {
 			}
 
 			//  Lazily initialise frame state
-			window_state.SetCount(frames.back().end - frames.front().start);
+			window_state.SetCount(FrameSet(frames).Size());
 			auto index2 = window_state.m.data();
 			D_ASSERT(index2);
 
--- a/test/issues/general/test_24442.test
+++ b/test/issues/general/test_24442.test
@@ -1,0 +1,89 @@
+# name: test/issues/general/test_24442.test
+# description: MAD window filtering respects frame exclusion
+# group: [general]
+
+statement ok
+CREATE TABLE src(id INTEGER, x INTEGER, keep BOOLEAN)
+
+statement ok
+INSERT INTO src VALUES (1, 0, true), (2, 100, true), (3, 10, true), (4, 20, false), (5, 30, true)
+
+statement ok
+CREATE TABLE peers(rid INTEGER, k INTEGER, x INTEGER, keep BOOLEAN)
+
+statement ok
+INSERT INTO peers VALUES
+	(1, 1, 0, true),
+	(2, 2, 100, true),
+	(3, 2, 10, false),
+	(4, 2, 20, true),
+	(5, 3, 30, true)
+
+foreach windowmode "window" "combine" "separate"
+
+statement ok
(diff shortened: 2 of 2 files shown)


hawkfish (approved):
Thanks - LGTM!

(Please convert this from Draft so we can get it into the queue.)

Alignyx:
Hi @hawkfish, thanks for your work! I have converted this from Draft.

PR #24849, opened 2026-08-18
fix geometry row group pruning with nulls/empty (on main)

Maxxen:
Same as https://github.com/duckdb/duckdb/pull/24845, but on main


Mytherin:
Thanks!

Maxxen:
The regression failures are unrelated - network timeout

PR #24853, opened 2026-08-18
Async streaming results

evertlammerts:
This introduces asynchronous streaming query result consumption. It doesn't change the existing API, just adds to it, and existing streaming results (which are not async) keep working as normal. They do profit somewhat from these changes in terms of performance (see the bechmarks).

This work would enable asynchronous concurrency in clients - think Node and Python.

## Usage

- Enable async mode with the setting `SET streaming_execution_mode = 'async'`.
- For blocking fetches on a streaming result in async mode, use `stream.Fetch()`.
- For non-blocking fetches on a streaming result in async mode, use `stream.TryFetchChunk()`.
- For callback notifications of state changes (`CHUNK_READY`, `EXECUTION_FINISHED`, `EXECUTION_CANCELLED` or 
  `EXECUTION_ERROR`), set `notify_callback` on a connection's ClientConfig (`ClientConfig::GetConfig(*con.context)`).

**Important notes**:
- `streaming_execution_mode` is a connection-local config setting. Different connections can use sync or async mode,
  and are allowed to switch modes at any time.
- Async is only supported:
   - On streaming results. Make sure to set `QueryResultOutputType::ALLOW_STREAMING` in the query parameters. Also 
     make sure to check `result->GetResultType() == QueryResultType::STREAM_RESULT` after creating a result since 
     non-SELECT queries will always be materialized, i.e. not support async.
   - For multithreaded setups. Enabling async in single threaded mode will error, and `SET threads` / `SET 
   external_threads` when async is enabled on one or more connections will error.

## Example: blocking fetch in async mode

```cpp
Connection con(db);
con.Query("SET streaming_execution_mode = 'async'");

auto result = con.PendingQuery(query, QueryParameters(QueryResultOutputType::ALLOW_STREAMING))->Execute();
auto &stream = result->Cast<StreamQueryResult>();
// Fetch() blocks until a chunk is ready but never executes engine work on this thread.
while (auto chunk = stream.Fetch()) {
    Process(*chunk);
}
// Errors will surface here, no need to check the state
if (stream.HasError()) {
    HandleError(stream.GetError());
}
```

## Example: async execution with a callback

```cpp
Connection con(db);
con.Query("SET streaming_execution_mode = 'async'");

// Set the callback.
std::mutex m;
std::condition_variable cv; 
bool wake = false;
ClientConfig::GetConfig(*con.context).notify_callback = [&]() {
    std::lock_guard<std::mutex> guard(m);
    wake = true;
    cv.notify_one();
};

auto result = con.PendingQuery(query, QueryParameters(QueryResultOutputType::ALLOW_STREAMING))->Execute();
auto &stream = result->Cast<StreamQueryResult>();

unique_ptr<DataChunk> chunk;
while (true) {
    auto state = stream.TryFetchChunk(chunk);
    if (state == StreamExecutionResult::CHUNK_READY) {
            Process(*chunk);
            continue;
    }
    if (state == StreamExecutionResult::EXECUTION_FINISHED) {
            break;
    }
    if (state == StreamExecutionResult::EXECUTION_ERROR || state == StreamExecutionResult::EXECUTION_CANCELLED) {
            HandleError(stream.GetError());
            break;
    }
    // No chunk yet. Sleep until the engine notifies.
    std::unique_lock<std::mutex> guard(m);
    cv.wait(guard, [&]() { return wake; });
    wake = false;
}
```

**Notes**:
- The callback runs on an engine thread and so it _must_ be tiny and _must not_ call back into DuckDB.
- Callbacks might overlap and the state it triggered on might have changed by the time the consumer tries to fetch, so 
  the consumer must check the stream's state every time.
- `std::condition_variable` is not suitable for use in signal handlers on POSIX. E.g. the DuckDB CLI handles ctrl+c by
  calling Interrupt(), which will call the notify callback. If your code does something similar then use a more 
  suitable idiom, like a pipe.

# Changes

First some terminology: I make a distinction here between a **consumer thread** and an **engine thread**:
- A consumer thread is a thread from which a query is started, i.e. a thread that does something like `result = con->Query(...)`. In my mental model, this thread is "external".
- An engine thread is a thread that executes DuckDB pipeline tasks, and is part of the engine's threadpool. In my mental model, this thread is "internal".

In sync mode (i.e. how things currently work) when a consumer thread does this:
```cpp
Connection con(db);
auto result = con.PendingQuery(query, QueryParameters(QueryResultOutputType::ALLOW_STREAMING))->Execute();
auto &stream = result->Cast<StreamQueryResult>();
while (auto chunk = stream.Fetch()) {
    Process(*chunk);
}
```
... the consumer thread does multiple things when it calls `Fetch()`:
1. It participates in task execution alongside the engine threads.
1. It marks blocked tasks as unblocked.
1. It returns a data chunk to the caller, if one is available.

This PR introduces an async mode, where the consumer thread **never participates in task execution**. It adds:
- **Async query execution mode**: `streaming_execution_mode = [sync | async]`, a connection-scoped setting that controls the execution mode of queries running on the connection.
- **Polling**: `StreamExecutionResult StreamQueryResult::TryFetchChunk(unique_ptr<DataChunk> &out_chunk)`, which is unblocking in the sense that it will not make the caller wait until a chunk is available. This can be used to poll a query.
- **Async callback**: `ClientConfig::notify_callback`, an optional (connection scoped) callback that can be set to notify a consumer about a relevant change in state of the execution.

# Benchmarks

The below compares main (always sync) with this branch's sync and async modes. Two notes:
- Async will noticeably perform better than sync on buffer sizes that are tuned to the workload, with a sweet spot across different workloads between 16mb and 8mb. 
- More importantly: async enables evented (client) concurrency that significantly outperforms multihreaded concurrency, both in time and memory.

Running it:
```sh
BUILD_BENCHMARK=1 make release
./build/release/benchmark/benchmark_runner "benchmark/micro/result_collection/stream_(sync|async).*" --threads=5 --timed-runs 7 --buffer_size 8mb
```

Results
```
┌─────────┬─────────────┬───────────────┬────────────────┬──────────────────┬───────────────────┐
│ buffer  │ main_sync_s │ branch_sync_s │ branch_async_s │ sync_vs_main_pct │ async_vs_main_pct │
│ varchar │   double    │    double     │     double     │      double      │      double       │
├─────────┼─────────────┼───────────────┼────────────────┼──────────────────┼───────────────────┤
│ 100kb   │       0.558 │         0.493 │           0.55 │            -11.7 │              -1.5 │
│ 1mb     │       0.486 │         0.481 │          0.528 │             -0.9 │               8.6 │
│ 8mb     │       0.469 │         0.418 │          0.379 │            -10.9 │             -19.2 │
│ 64mb    │       0.143 │         0.111 │          0.174 │            -22.5 │              21.6 │
│ 100gb   │       0.121 │          0.11 │          0.137 │             -9.2 │              13.1 │
└─────────┴─────────────┴───────────────┴────────────────┴──────────────────┴───────────────────┘
```

# Related

duckdb/duckdb-quack#207
duckdb/duckdb-quack#208
duckdb/duckdb-quack#229
duckdb/duckdb-quack#252

# (Possible) follow ups

* Support async for `CREATE TABLE AS SELECT` type queries. Should be a small follow-up.
* Add support for custom collectors. Arrow & Quack use cases.
* Detach calling the notify callback from task execution (see [comment](https://github.com/duckdb/duckdb/pull/24853#issuecomment-5440346718)).
* Introduce something like a `ActiveQueryState` that is shared between the executor and the client context to wrap the state-data that needs to be shared, i.e. `interrupt_state`, `active_result_notifier`, `streaming_buffered_data`, `execution_result`, etc. I could also hold the `QueryProfiler`, `QueryProgress`, and maybe other things. Right now these are all scattered, and sometimes intertwined with execution (e.g. `CheckPulse()` does a dry run `ExecuteTask` to make sure blocked tasks are unblocked _and_ to get the current state; and we keep both the notifier callback and a pointer to the result buffer on the `Executor`).

### DIFF
--- a/src/include/duckdb/main/query_result_notifier.hpp
+++ b/src/include/duckdb/main/query_result_notifier.hpp
@@ -1,0 +1,60 @@
+//===----------------------------------------------------------------------===//
+//                         DuckDB
+//
+// duckdb/main/query_result_notifier.hpp
+//
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "duckdb/common/mutex.hpp"
+
+#include <functional>
+
+namespace duckdb {
+
+//! Notifies a waiting stream consumer that the result's observable state may have
+//! changed. Notifications might overlap and the state that triggered it may have
+//! changed by the time the consumer tries to fetch. The callback runs on an engine
+//! thread and so it _must_ be tiny and _must not_ call back into DuckDB. A callback
+//! that re-enters DuckDB can deadlock the connection.
+class QueryResultNotifier {
+public:
+	using notify_callback_t = std::function<void()>;
+
+	void Set(notify_callback_t callback_p) {
+		lock_guard<mutex> guard(lock);
+		callback = std::move(callback_p);
+	}
+	//! After Clear returns the callback is never called again
+	void Clear() {
+		lock_guard<mutex> guard(lock);
+		callback = nullptr;
+	}
+	//! Callers must not hold any engine lock. The callback runs under the notifier's lock.
+	void Notify() {
+		lock_guard<mutex> guard(lock);
+		if (callback) {
+			callback();
+		}
+	}
+	//! Non-blocking notify for signal handlers (ClientContext::Interrupt). On contention
+	//! the notification is dropped. That is safe. A contending Notify already wakes the
+	//! consumer, and a contending Clear means the result is going away.
+	void TryNotify() {
+		if (!lock.try_lock()) {
+			return;
+		}
+		if (callback) {
+			callback();
+		}
+		lock.unlock();
+	}
+
+private:
+	mutex lock;
+	notify_callback_t callback;
+};
+
+} // namespace duckdb
--- a/src/main/buffered_data/batched_buffered_data.cpp
+++ b/src/main/buffered_data/batched_buffered_data.cpp
@@ -1,71 +1,188 @@
 #include "duckdb/main/buffered_data/batched_buffered_data.hpp"
 #include "duckdb/common/printer.hpp"
+#include "duckdb/execution/executor.hpp"
 #include "duckdb/main/client_context.hpp"
 #include "duckdb/main/stream_query_result.hpp"
 #include "duckdb/common/helper.hpp"
 #include "duckdb/execution/operator/helper/physical_buffered_batch_collector.hpp"
 #include "duckdb/common/stack.hpp"
 
+#include <chrono>
+
 namespace duckdb {
 
-void BatchedBufferedData::BlockSink(const InterruptState &blocked_sink, idx_t batch) {
-	lock_guard<mutex> lock(glock);
-	D_ASSERT(!blocked_sinks.count(batch));
-	blocked_sinks.emplace(batch, blocked_sink);
-}
+bool BatchedBufferedData::AppendOrBlock(const DataChunk &to_append, idx_t batch, const InterruptState &blocked_sink) {
+	// Copy first. The block decision and the byte accounting share one critical
+	// section, so racing producers cannot exceed the cap. A block keeps the copy
+	// for wake time
+	auto chunk = make_uniq<DataChunk>();
+	chunk->Initialize(Allocator::DefaultAllocator(), to_append.GetTypes());
+	to_append.Copy(*chunk, 0);
+	const idx_t allocation_size = chunk->GetDataSize();
 
-BatchedBufferedData::BatchedBufferedData(ClientContext &context)
-    : BufferedData(BufferedData::Type::BATCHED, context), buffer_byte_count(0), read_queue_byte_count(0), min_batch(0) {
-	read_queue_capacity = (idx_t)(static_cast<double>(total_buffer_size) * 0.6);
-	buffer_capacity = (idx_t)(static_cast<double>(total_buffer_size) * 0.4);
-}
-
-bool BatchedBufferedData::ShouldBlockBatch(idx_t batch) {
-	lock_guard<mutex> lock(glock);
-	bool is_minimum = IsMinimumBatchIndex(lock, batch);
-	if (is_minimum) {
-		// If there is room in the read queue, we want to process the minimum batch
-		return read_queue_byte_count >= ReadQueueCapacity();
+	shared_ptr<QueryResultNotifier> notifier;
+	bool signal_chunk_ready = false;
+	{
+		annotated_lock_guard<annotated_mutex> lock(glock);
+		D_ASSERT(batch >= min_batch);
+		max_seen_chunk_bytes = MaxValue<idx_t>(max_seen_chunk_bytes, allocation_size);
+		if (ShouldBlockBatch(lock, batch, allocation_size)) {
+			// Park holding the finished copy. Restart selection deposits it at wake time
+			auto entry = blocked_sinks.emplace(batch, BlockedSink {blocked_sink, allocation_size, std::move(chunk)});
+			(void)entry;
+			D_ASSERT(entry.second);
+			return true;
+		}
+		auto is_minimum = IsMinimumBatchIndex(lock, batch);
+		if (is_minimum) {
+			for (auto &it : buffer) {
+				auto batch_index = it.first;
+				if (batch_index >= min_batch) {
+					break;
+				}
+				// There should not be any batches in the buffer that are lower or equal to the minimum batch index
+				throw InternalException("Batches remaining in buffer");
+			}
+			signal_chunk_ready = read_queue.empty();
+			read_queue.push_back(std::move(chunk));
+			read_queue_byte_count += allocation_size;
+			if (signal_chunk_ready) {
+				// The chunk is visible before the wake below, so a woken consumer always finds it
+				notifier = result_notifier;
+			}
+		} else {
+			auto &in_progress_batch = buffer[batch];
+			auto &chunks = in_progress_batch.chunks;
+			in_progress_batch.completed = false;
+			buffer_byte_count += allocation_size;
+			chunks.push_back(std::move(chunk));
+		}
+		peak_buffered_bytes = MaxValue<idx_t>(peak_buffered_bytes, read_queue_byte_count + buffer_byte_count);
 	}
-	return buffer_byte_count >= BufferCapacity();
+	if (signal_chunk_ready) {
+		// This collector's Sink holds no outer lock, so the buffer signals directly
+		SignalChunkAvailable(notifier);
+	}
+	return false;
+}
+
+BatchedBufferedData::BatchedBufferedData(ClientContext &context, bool async)
+    : BufferedData(BufferedData::Type::BATCHED, context, async), buffer_byte_count(0), read_queue_byte_count(0),
+      min_batch(0) {
+	// Both capacities must be able to admit a chunk. The split rounds tiny sizes to zero
+	read_queue_capacity = MaxValue<idx_t>((idx_t)(static_cast<double>(total_buffer_size) * 0.6), 1);
+	buffer_capacity = MaxValue<idx_t>((idx_t)(static_cast<double>(total_buffer_size) * 0.4), 1);
+}
+
+bool BatchedBufferedData::ShouldBlockBatch(annotated_lock_guard<annotated_mutex> &lock, idx_t batch,
+                                           idx_t incoming_bytes) {
+	if (IsMinimumBatchIndex(lock, batch)) {
+		// The read queue admits a chunk that fits, and always one chunk when empty
+		return read_queue_byte_count > 0 && read_queue_byte_count + incoming_bytes > ReadQueueCapacity();
+	}
+	// An oversized non-minimum chunk blocks until its batch becomes the minimum.
+	// The advance then moves its sink to the read queue rule above, so it is not lost
+	return buffer_byte_count > 0 && buffer_byte_count + incoming_bytes > BufferCapacity();
+}
+
+bool BatchedBufferedData::PopMayFreeSinks(annotated_lock_guard<annotated_mutex> &lock) {
+	// Pops only shrink the read queue, so only the minimum batch sink can become eligible
+	return read_queue_byte_count < ReadQueueCapacity();
+}
+
+bool BatchedBufferedDataAsync::ShouldBlockBatch(annotated_lock_guard<annotated_mutex> &lock, idx_t batch,
+                                                idx_t incoming_bytes) {
+	const idx_t total = read_queue_byte_count + buffer_byte_count;
+	if (IsMinimumBatchIndex(lock, batch)) {
+		// Only block the minimum batch producer while the consumer has chunks to pop.
+		// Every pop lowers the total, so consumption always wakes it again
+		return total + incoming_bytes > total_buffer_size && !read_queue.empty();
+	}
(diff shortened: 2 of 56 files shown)


carlopi:
I had a very fun (and fruitful?) conversation with @evertlammerts, work looks great, I guess my main question is whether some locking logic can be simplified / removed, and that would unlock to remove the single thread requirement.

What we might have realized is that locks are needed to do deduplication, but if client/driver is responsible of that, then that might allows to simplify some code like `TryNotify`. I am not sure whether that's viable / desirable, but author will dig and figure it out.

Great stuff.

evertlammerts:
Re locks, neither `QueryResultNotifier::Notify()` nor `QueryResultNotifier::TryNotify()` tries to acquire any engine locks itself.

In terms of callsites, no paths hold a lock when calling `QueryResultNotifier::Notify()`. It's true that `QueryResultNotifier::TryNotify()`, cannot be guaranteed to not run under a lock. But that's a very narrow case: it is used only for `ClientContext::Interrupt()`, which may run in a signalhandler. On POSIX systems the handler will run on any existing program thread, at any given point. So it might or might not hold arbitrary engine locks. But the `QueryResultNotifier::TryNotify()` itself does not require any system lock, so apart from its own mutex it can't race (assuming the callback doesn't do strange things, in which case all bets are off anyway).

That said, I do think (as a follow up) we could detach the callback from task execution. A simple example:
- Every task that transitions the pipeline into some state that the consumer needs to know about sets an atomic bool to true.
- At a limited amount of points that are guaranteed to be lock free (task lifecycle + buffer mutating operations) we check the bool and call the callback.

This way we can't shoot ourselves in the foot later on, and accidentally call `Notify()` (through some other chain) while holding a lock.

src/main/buffered_data/simple_buffered_data.cpp
```diff
@@ -1,26 +1,76 @@
...
 	buffer_size = total_buffer_size;
 }
 
 SimpleBufferedData::~SimpleBufferedData() {
 }
 
-void SimpleBufferedData::BlockSink(const InterruptState &blocked_sink) {
-	lock_guard<mutex> lock(glock);
-	blocked_sinks.push(blocked_sink);
+bool SimpleBufferedData::BufferSaturated() {
```
guillesd, line 22: You mentioned in the PR description that the sync path is unchanged, but this seems to affect it, correct? This is the logic that gatekeeps `CHUNK_READY`

src/main/buffered_data/simple_buffered_data.cpp
```diff
@@ -77,49 +151,86 @@ StreamExecutionResult SimpleBufferedData::ExecuteTaskInternal(StreamQueryResult
...
 
-	unique_lock<mutex> lock(glock);
-	buffered_count += allocation_size;
-	buffered_chunks.push(std::move(chunk));
+	shared_ptr<QueryResultNotifier> notifier;
+	bool was_empty;
+	{
+		annotated_unique_lock<annotated_mutex> lock(glock);
+		// The buffer admits a chunk that fits, and always one chunk when empty
+		if (buffered_count > 0 && buffered_count + allocation_size > BufferSize()) {
```
guillesd: This also changes the sync mode? Also why `allocation_size` and not `chunk_size` (feels more intuitive to read)

src/main/buffered_data/simple_buffered_data.cpp
```diff
@@ -77,49 +151,86 @@ StreamExecutionResult SimpleBufferedData::ExecuteTaskInternal(StreamQueryResult
...
 
-	if (chunk) {
-		auto allocation_size = chunk->GetDataSize();
-		buffered_count -= allocation_size;
+		if (chunk) {
+			auto allocation_size = chunk->GetDataSize();
+			buffered_count -= allocation_size;
+		}
+		// The pop restarts blocked producers below the low-water mark
+		if (buffered_count < LowWaterMark(BufferSize())) {
```
guillesd, line 185: This also changes the restarting of the pipeline execution in sync mode, correct? The trend is basically: the sync execution is changed so I don't think we can make the claim in the PR body that it didn't. Not against it particularly, just flagging

src/main/buffered_data/simple_buffered_data.cpp
```diff
@@ -77,49 +151,86 @@ StreamExecutionResult SimpleBufferedData::ExecuteTaskInternal(StreamQueryResult
...
 	return chunk;
 }
 
-void SimpleBufferedData::Append(const DataChunk &to_append) {
+bool SimpleBufferedData::AppendOrBlock(const DataChunk &to_append, const InterruptState &blocked_sink) {
+	// Copy first. The cap check must use the flat copy's size. The source can be
+	// dictionary-compressed and much smaller. A block keeps the copy for wake time.
 	auto chunk = make_uniq<DataChunk>();
 	chunk->Initialize(Allocator::DefaultAllocator(), to_append.GetTypes());
 	to_append.Copy(*chunk, 0);
```
guillesd, line 198: Everytime the sink is blocked this copy is wasted. I guess for specific queries where an operator may be slow and sinks may block a lot this can be quite inefficient? Also this cost affects the sync path as well. You could park the copy in the local sink state and re-use it rather than throwing it away.

src/main/buffered_data/batched_buffered_data.cpp
```diff
@@ -1,71 +1,188 @@
...
+	if (IsMinimumBatchIndex(lock, batch)) {
+		// Only block the minimum batch producer while the consumer has chunks to pop.
+		// Every pop lowers the total, so consumption always wakes it again
+		return total + incoming_bytes > total_buffer_size && !read_queue.empty();
+	}
+	// Read-ahead batches may not use up the share reserved for the minimum batch.
+	// The reserve covers at least one chunk, so the empty-queue append above stays
+	// under the cap. A chunk larger than every chunk seen before can exceed it once,
+	// by at most its own size. It then blocks until its batch becomes the minimum.
+	const idx_t reserve = MaxValue<idx_t>(MaxValue<idx_t>(total_buffer_size / 8, max_seen_chunk_bytes), 1);
```
guillesd, line 105: the max_seen_chunk_bytes only grows. Could there be a case were an unusually large chunk makes the reserve too big and then read ahead is less effective?

src/include/duckdb/main/query_result_notifier.hpp
```diff
@@ -0,0 +1,60 @@
...
+	//! After Clear returns the callback is never called again
+	void Clear() {
+		lock_guard<mutex> guard(lock);
+		callback = nullptr;
+	}
+	//! Callers must not hold any engine lock. The callback runs under the notifier's lock.
+	void Notify() {
+		lock_guard<mutex> guard(lock);
+		if (callback) {
+			callback();
```
guillesd, line 39: I guess there is a risk of a deadlock here if the callback closes/destroys results (this will run `Clear()` under the same lock). Maybe worth a comment although it may be common sense enough.

src/main/client_context.cpp
```diff
@@ -653,16 +670,32 @@ ClientContext::PendingPreparedStatementInternal(ClientContextLock &lock,
...
+	// submission, once per query. The setting and the task scheduler maintain the
+	// invariant that async connections have a DuckDB-managed worker thread.
+	const auto async_result = stream_result && client_config.streaming_execution_mode == StreamingExecutionMode::ASYNC;
 	if (!stream_result && client_config.get_result_collector) {
 		get_collector = client_config.get_result_collector;
 	}
 	statement_data.output_type =
 	    stream_result ? QueryResultOutputType::ALLOW_STREAMING : QueryResultOutputType::FORCE_MATERIALIZED;
 	statement_data.memory_type = parameters.query_parameters.memory_type;
+	statement_data.execution_mode = async_result ? StreamingExecutionMode::ASYNC : StreamingExecutionMode::SYNC;
```
guillesd, line 684: either here or in the if condition bellow, i'd add `async_result || collector->Cast<PhysicalResultCollector>().IsStreaming())`
The reason for this is that extensions like Quack, that have their own result collectors (which may be streaming even though `stream_result` may be false because this only really depends on the query result type being streaming), will want to benefit from the notification machinery. So this is a very important one for me :)

guillesd (commented):
Overall I like the idea a lot. Two takes:
1) For Quack the notifying machinery is very interesting. Quack does have its own result collector (and in the future I think it will have its own query result type, maybe), so I'm more interested in the notifier machinery than in the most superficial API from the Quack perspective.
2) From the client perspective this is pretty great I think. I still wonder if this has to be a connection setting or it could also be potentially a per query parameter so that connections can be pooled without having to default to a mode? Maybe in `QueryParameters`?

That being said, the PR is huge and I think it requires a review from someone that is more familiar with some of this components than I am.

src/main/buffered_data/simple_buffered_data.cpp
evertlammerts: You're right, the description should say that there are no changes to the "regular" API, which is sync. Internally, there are some changes that affect sync mode, making it faster. Editing now.

src/main/buffered_data/simple_buffered_data.cpp
evertlammerts: Ack. Description changed. See response below.

src/main/buffered_data/simple_buffered_data.cpp
evertlammerts: Good catch :) It's just naming though, `allocation_size` is actually the chunk size, not the allocation size:

```
auto allocation_size = chunk->GetDataSize();
```

I've renamed it.

src/main/buffered_data/simple_buffered_data.cpp
evertlammerts: I think this was true in yesterday's version but not anymore. We now don't disregard the copy but just keep it and add it when there's space.

src/main/buffered_data/batched_buffered_data.cpp
evertlammerts: Yes, that is possible. I don't know how often extreme skew might occur, tbh, but if it does then it's possible that read ahead is disabled. I could introduce a different metric, a quantile for example. Or I could change it to decrease with some factor on every append, e.g. `MaxValue(incoming, max_seen - max_seen / 64)`. That's simpler maybe. But it would need tests as well to motivate whatever we choose, and I'd prefer to experiment with this in a potential follow up.

src/include/duckdb/main/query_result_notifier.hpp
evertlammerts: Yeah, the callback should not be re-entrant at all. As soon as it calls back into duckdb all bets are off. I have a follow up listed that would deal with making this safer by guaranteeing a callback is never run under a lock at all.

src/main/client_context.cpp
evertlammerts: Good idea, but I think it needs dedicated work tbh.

Right now the result type is bound to the physical collector and the buffer (`StreamQueryResult` -> `BufferedData`), so right now you wouldn't be able to use a `StreamQueryResult` with another collector in async mode unless you use the buffers as well. That's not really an issue for Quack I guess, since it won't use `StreamQueryResult` right?

I would propose we do this in a follow up and enable it for the arrow collectors. That way we can formalize the design (or at least document the expectations). If we add support for a custom collector now then it would be untested here. And it's on the collector to guarantee liveness, i.e. decide when to park and restart tasks, and when to notify.

PR #24855, opened 2026-08-18
Arena allocate per-group buffers in bitstring_agg, bit_and/bit_or/bit_xor, and mode

maiadegraaf:
`bitstring_agg`, the `bit_and`/`bit_or`/`bit_xor` bitwise aggregates over `BIT`, and `mode's` non-inlined string/window state all allocated with raw `new char[]`, so memory_limit couldn't be enforced for these aggregates.  In some circumstances this could kill the process instead of failing gracefully. This mirrors #24125 for TDigest/`APPROX_QUANTILE`: the memory now goes through an arena so it's tracked.                                                                                        

`mode` needed a bit more care since its per-state buffer can shrink and grow again.  It now tracks the arena allocation's capacity separately from the current key's length so it can reuse the buffer in place when the new key fits, instead of churning a new allocation on every update.


Mytherin:
Thanks!

PR #24856, opened 2026-08-18
Keep the list_aggregate bind data unambiguous

carlopi:
list_aggregate has three bind outcomes: a ListAggregatesBindData, a VariableReturnBindData when the argument is SQLNULL, and no bind data at all when the argument is a prepared statement parameter. Serialization then had to tell them apart with a dynamic_cast.

Return no bind data from the SQLNULL path too, matching the parameter path, so the bind data of this function is either a ListAggregatesBindData or absent.

This also removes the only way a foreign bind data type could reach ListAggregatesFunction, which casts it with the unchecked Cast<>.

The added test covers a list of only NULLs, which binds successfully but has a SQLNULL return type, so it is not distinguishable from a bind failure by the return type alone. It runs under test/configs/verify_serializer.json.


Mytherin:
Thanks!

PR #4706, opened 2022-09-14
Fix #4677. Correctly set_not_null when table contains generated column

zippond:
Symptom
===============
Fix Issue #4677. heap-buffer-overflow when set not null against table with generated column presented.

	
Root Cause
===============
Related to generated columns. TableCatalogEntry::columns contains all columns include the generated columns, while DataTable::column_definitions contains the real columns only. 
The index of Constraint and BoundConstraint doesn't match after bounding. And the NotNullConstraint.index might overflow the DataTable::column_definitions' size() which cause a segfault.

Fix
===============
Use BoundNotNullConstraint to check new added constraint. And the BoundNotNullConstraint::index can be get from ColumnDefinition::StorageOid()

Test 
===============
Tests has been added to add/drop not_null with generated column being presented.

Other Info
===============
With generated column presented, I see similiar issues with "alter type", "update table", "remove column", etc. 
I'm not sure if there could be a fix in generated_column logic to fix all of them. Anyway, I'll open seperated issues for now.

PR #24857, opened 2026-08-18
Add overload: regexp_extract('string', 'pattern', 'options')

lghavami:
A fix for issue [#4706](https://github.com/duckdblabs/duckdb-internal/issues/4706)

Adds a new overload to regex_extract which takes a string, pattern, and regex option(s) as an argument. This overload makes it so you don't need to specify a group or name_list to access options.


src/function/scalar/string/regexp.cpp
```diff
@@ -369,28 +369,31 @@ static unique_ptr<FunctionData> RegexExtractBind(BindScalarFunctionInput &input)
...
+
 	if (arguments.size() >= 4) {
 		ParseRegexOptions(input.GetConstant(3), options, nullptr, &no_match_returns_input);
 	}
 
 	int8_t group_index = 0;
 	if (arguments.size() >= 3) {
-		Value group = input.GetConstant(2);
-		if (group.IsNull()) {
+		Value arg2 = input.GetConstant(2);
```
artjomPlaunov: Here we are checking for NULL first, then the argument type. This means if we have NULL for the options, we enter the null group branch even if we have the options VARCHAR overload, and never call ParseRegexOptions() (see related test comment for the expected behavior here).

test/sql/function/string/regex_extract.test
```diff
@@ -106,7 +106,63 @@ NULL
...
+----
+platy
+
+query T
+SELECT regexp_extract('how much good would a good duck duck if a good duck could duck good????', 'g..', '')
+----
+goo
+
+query T
+SELECT regexp_extract('q u a c k', 'q..', CAST(NULL AS VARCHAR))
```
artjomPlaunov: (see other review comment). 

test in regexp_replace_constant.test for what the behavior should be (if NULL options hits the ParseRegexOptions):

```sql
statement error
SELECT regexp_extract('foobarbaz', 'B..', 0, NULL);
----
must not be NULL
```

src/function/scalar/string/regexp.cpp
artjomPlaunov: nit: let's rename arg2 to group_or_options. We can also reuse it in the ParseRegexOptions() call below instead of calling input.GetConstant(2) again.

test/sql/function/string/regex_extract.test
```diff
@@ -106,7 +106,63 @@ NULL
...
+
+statement error
+SELECT regexp_extract(s, p, o) FROM test_opts
+----
+<REGEX>:.*Binder Error.*must be a constant.*
+
+query T
+SELECT regexp_extract('all hail Anas platyrhynchos', 'zzzz', 'k')
+----
+all hail Anas platyrhynchos
```
artjomPlaunov, line 145: 🦆

artjomPlaunov (changes requested):
Left some comments!

test/sql/function/string/regex_extract.test
lghavami: quack quack

src/function/scalar/string/regexp.cpp
lghavami: fixed!

test/sql/function/string/regex_extract.test
lghavami: fixed !

src/function/scalar/string/regexp.cpp
lghavami: also fixed !

test/sql/function/string/regex_extract.test
artjomPlaunov: @lghavami sorry I should have been more clear, I was just showing that as an existing test behavior with the 4 argument overload (that test already exists in regexp_replace_constant.test, and the three argument overload (the new overload) test you added previously should have similar behavior. 

can you add the new overload test back and make sure it has the correct "must not be NULL" behavior instead of (empty)?

test/sql/function/string/regex_extract.test
```diff
@@ -106,7 +106,63 @@ NULL
...
+----
+platy
+
+query T
+SELECT regexp_extract('how much good would a good duck duck if a good duck could duck good????', 'g..', '')
+----
+goo
+
+statement error
+SELECT regexp_extract('foobarbaz', 'B..', 0, NULL);
```
artjomPlaunov: @lghavami sorry I should have been more clear, I was just showing this as an existing test behavior with the 4 argument overload (that test already exists in regexp_replace_constant.test, and the three argument overload (the new overload) test you added previously should have similar behavior.

add the new overload test back and make sure it has the correct "must not be NULL" behavior instead of (empty)

lghavami: modified the test & changed behavior to match

artjomPlaunov (approved):
Looks good to go!

PR #24858, opened 2026-08-18
No rtti build

carlopi:
On top of https://github.com/duckdb/duckdb/pull/24856, more changes for allowing to compile duckdb with `-fno-rtti`. Due to the limitations while compiling / linking arbitrary libraries, this is currently not recommended as default, but I think making sure we can compile with `-fno-rtti`, either locally or in CI, would allow to limit regressions on this side, and should be a stronger proof that cross module / extensions boundaries duckdb still has the same behaviour.

Change in https://github.com/duckdb/duckdb/pull/24856 was the most complex, landed on its own,
rest are more distributed but somewhat mechanical changes around the code base allowing to remove all `dynamic_casts`. Given this is not the default, main test is whether this passes the CI check that has yet to run.


Mytherin:
Thanks!

.github/workflows/Main.yml
```diff
@@ -1211,6 +1211,50 @@ jobs:
...
+    runs-on: ${{ needs.prepare.outputs.runner_linux_x64 }}
+    needs:
+      - prepare
+      - linux-relassert
+    env:
+      # Job-level on purpose: the Test step's `make unittest_release` depends on `release`, so a
+      # step-scoped flag would silently reconfigure and rebuild WITH rtti and test the wrong binary.
+      DISABLE_RTTI: 1
+      EXTENSION_CONFIGS: .github/config/in_tree_extensions.cmake
+      DUCKDB_TEST_DESCRIPTION: 'Compiled with DISABLE_RTTI=1 (-fno-rtti).'
```
carlopi, line 1226: Comment from @smvv: could / should this job be folded in another job?

carlopi (commented):
Should CI job folded elsewhere?

PR #24860, opened 2026-08-18
Assorted bugfixes tied to conversion errors

carlopi:
Six different independent commits, fixing corner cases around conversion errors.


Mytherin:
Thanks!

PR #24861, opened 2026-08-18
Avoid overwriting streaming NTH_VALUE result on NULL rows

### LINKED ISSUE #24827: Streaming `NTH_VALUE IGNORE NULLS` loses a found value
### What happens?

`NTH_VALUE(x, 2 IGNORE NULLS)` over a legal running `ROWS` frame returns `NULL` after the second non-NULL value has entered the frame. Disabling the optimizer restores the expected result.

### To Reproduce

```sql
WITH src(id, x) AS (VALUES (1, 10), (2, 20), (3, NULL))
SELECT id, NTH_VALUE(x, 2 IGNORE NULLS) OVER (
) AS actual
FROM src ORDER BY id;
```

### Expected

| id | expected |
| ---: | ---: |
| 1 | NULL |
| 2 | 20 |
| 3 | 20 |

### Actual

| id | actual |
| ---: | ---: |
| 1 | NULL |
| 2 | NULL |
| 3 | NULL |

### Notes From Minimization

- `SET enable_optimizer=false` changes the output to the expected `NULL, 20, 20` sequence.
- `SET debug_window_mode='separate'` changes the output to the expected sequence.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (`11dc00c898`)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

jaiswalvikas9991:
Fixes #24827

Streaming NTH_VALUE with IGNORE NULLS keeps a running count of non-NULL values. After the requested value has been found, later NULL rows should not overwrite the cached result.

This change only captures the NTH_VALUE result on rows that advance the running count, and adds regression coverage for trailing NULL rows.

Tested with:

  ```bash
./build/debug/test/unittest test/sql/window/test_streaming_nthvalue.test
```

### DIFF
--- a/src/function/window/window_value_function.cpp
+++ b/src/function/window/window_value_function.cpp
@@ -1127,13 +1127,13 @@ void WindowNthValueStreamingState::StreamData(ExecutionContext &context, DataChu
 	for (idx_t i = 0; i < count; ++i) {
 		if (!any_value || validity.RowIsValidUnsafe(unified.sel->get_index(i))) {
 			++nth_count;
-		}
-		// One-based comparison.
-		if (nth_count == nth_index) {
-			auto v = arg.GetValue(i);
-			vec.Reference(v, count_t(count));
-			s = 1;
-			split.SetValue(s, v);
+			// One-based comparison.
+			if (nth_count == nth_index) {
+				auto v = arg.GetValue(i);
+				vec.Reference(v, count_t(count));
+				s = 1;
+				split.SetValue(s, v);
+			}
 		}
 		sel.set_index(i, s);
 	}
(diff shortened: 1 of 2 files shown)


jaiswalvikas9991:
The failing checks look unrelated to this PR from what I can tell. The changed files are limited to `src/function/window/window_value_function.cpp` and `test/sql/window/test_streaming_nthvalue.test`, and the relevant test jobs passed. The remaining failures seem to be regression/benchmark timeouts with exit code 124.  

I do not have permission to rerun the failed jobs, but I am happy to investigate further if needed.

hawkfish (approved):
Thanks - LGTM!

jaiswalvikas9991:
@lnkuiper This has been reviewed and marked good to go by @hawkfish. Could you take a look when you have a chance? Thanks!

PR #24865, opened 2026-08-18
[Dev][Parser] Add `optional_ptr<ClientContext>` to `ParserCache::GetMatcher`

Tishj:
needs #24864 

Very small PR, pushing the `ClientContext` into the `ParserCache::GetMatcher` call (where possible).

Paving the way for acting on registered extensions within this method, when creating the `CompiledGrammar`.

PR #24866, opened 2026-08-18
[Dev][Parser] Add `ParsedGrammar` intermediate to `CompiledGrammar` creation

Tishj:
needs #24865 

Before this change, the grammar file got directly turned into `CompiledGrammar`.
Now it turns into a `ParsedGrammar` first, which gets bundled with the transformer functions.

This gives us a surface on which extensions can make changes, before creating a `Matcher` from the grammar and outputting a finished `CompiledGrammar`.

A map of `CompiledGrammarRule` gets put into `CompiledGrammar`.
We push references to the individual rules through the `Matcher` and the `ParseResult`.
The `PEGTransformerFactory` constructor only exists to populate these rules into the `ParsedGrammar` now.


src/include/duckdb/parser/peg/transformer/parse_result.hpp
```diff
@@ -134,10 +135,24 @@ class ParseResult {
...
 	//! Source length: for leaf tokens the token length; for composite results the enclosing extent of children
 	optional_idx length;
 
+	void SetRule(const CompiledGrammarRule &rule_p) {
+		rule = rule_p;
+	}
+	bool HasRule() const {
+		return rule != nullptr;
+	}
+	const CompiledGrammarRule &GetRule() const {
```
Tishj: Perhaps we just want to use `optional_ptr<const CompiledGrammarRule>`, absorbing the `HasRule` method in the process.

src/include/duckdb/parser/peg/compiled_grammar.hpp
```diff
@@ -27,6 +28,7 @@ struct CompiledGrammar {
 	const Tokenizer &GetTokenizer() const {
 		return tokenizer;
 	}
+	const CompiledGrammarRule &GetRule(const string &rule_name) const;
```
Tishj: Same comment

src/include/duckdb/parser/peg/matcher_factory.hpp
```diff
@@ -14,7 +14,7 @@ class MatcherFactory {
 	struct MatcherList {
 	public:
 		struct Entry {
-			explicit Entry(Matcher &matcher) : matcher(matcher), function_name(0U) {
+			explicit Entry(Matcher &matcher) : matcher(matcher), function_name(string_t("")) {
```
Tishj: This is ugly, but it used to use a `string.empty()` to check for the root rule. Perhaps it's cleaner to use `optional<string_t>` instead

src/include/duckdb/parser/peg/parsed_grammar.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace duckdb {
+
+class ParseResult;
+class PEGTransformer;
+struct TransformFrameOps;
+
+using grammar_transform_function_t = std::function<unique_ptr<TransformResultValue>(PEGTransformer &, ParseResult &)>;
+
+struct ParsedGrammarRule {
```
Tishj, line 24: Note: we'll probably want to guard modifications to this when we make it modifiable by extensions, and track which modifications are made, to provide better checks of compatibility between extensions.

src/include/duckdb/parser/peg/parsed_grammar.hpp
```diff
@@ -0,0 +1,90 @@
...
+	ParsedGrammarRule(string name_p, PEGRule recipe_p) : name(std::move(name_p)), recipe(std::move(recipe_p)) {
+	}
+
+	string name;
+	PEGRule recipe;
+	grammar_transform_function_t transform;
```
Tishj: Maybe these variables should be lifted into a separate struct, since they're shared between `ParsedGrammarRule` and `CompiledGrammarRule`

src/include/duckdb/parser/peg/parsed_grammar.hpp
```diff
@@ -0,0 +1,90 @@
...
+	DUCKDB_API ParsedGrammar(ParsedGrammar &&other) noexcept;
+	DUCKDB_API ParsedGrammar &operator=(ParsedGrammar &&other) noexcept;
+	ParsedGrammar(const ParsedGrammar &) = delete;
+	ParsedGrammar &operator=(const ParsedGrammar &) = delete;
+
+	DUCKDB_API static ParsedGrammar Parse(const string &grammar);
+	DUCKDB_API static ParsedGrammar CreateDefault();
+
+	DUCKDB_API bool HasRule(const string &rule_name) const;
+	DUCKDB_API const ParsedGrammarRule &GetRule(const string &rule_name) const;
```
Tishj: Same comment about use of `optional_ptr`

src/include/duckdb/parser/peg/parsed_grammar.hpp
```diff
@@ -0,0 +1,90 @@
...
+	DUCKDB_API void AddRule(const string &rule_definition, grammar_transform_function_t transform = nullptr,
+	                        grammar_transform_function_t trampoline_transform = nullptr);
+	DUCKDB_API void ReplaceRule(const string &rule_definition, grammar_transform_function_t transform = nullptr,
+	                            grammar_transform_function_t trampoline_transform = nullptr);
+	DUCKDB_API void RemoveRule(const string &rule_name);
```
Tishj: Debatable whether we want to keep this, there's no need to remove a rule, it can just simply stop being referenced by other rules.

src/parser/peg/transformer/peg_transformer_factory.cpp
```diff
@@ -209,12 +208,13 @@ PEGTransformerFactory::PEGTransformerFactory() {
 	RegisterKeywordsAndIdentifiers();
 }
 
-const case_insensitive_map_t<PEGTransformer::AnyTransformFunction> &
-PEGTransformerFactory::GetTransformFunctions(ParserOptions &options) const {
-	if (options.debug_transformer_trampoline_style) {
-		return trampoline_transform_functions;
+void PEGTransformerFactory::RegisterDefaultTransforms(ParsedGrammar &grammar) {
+	PEGTransformerFactory factory(grammar);
```
Tishj, line 204: This sets the transformer functions into the `grammar`, not a great API, but it was the path of least resistance seeing as all the "RegisterX" methods are generated.

src/parser/peg/matcher_factory.cpp
```diff
@@ -90,21 +90,21 @@ Matcher &MatcherFactory::CreateMatcher(PEGParser &parser, string_t rule_name, ve
...
-		matchers.insert(make_pair(rule_name, reference<Matcher>(matcher)));
+		matchers.insert(make_pair(string_t(entry->second->name), reference<Matcher>(matcher)));
 	}
 
-	MatcherList list(parser, *this);
+	MatcherList list(*this);
 	list.AddRootMatcher(matcher);
 	// fill the matcher from the given set of rules
-	auto &rule = entry->second;
+	auto &rule = entry->second->recipe;
```
Tishj, line 107: nit: since this is a PEG (Parsing Expression Grammar), maybe `expression` makes more sense. I initially went with `recipe` because that's how the children of a rule work in my mental model

Mytherin:
Thanks!

PR #24867, opened 2026-08-18
[Dev][Parser] Generate `PEGKeywordHelper` from `ParsedGrammar`

Tishj:
needs #24866 

Previously, the only implementation of `PEGKeywordHelper` was `DuckDBKeywordHelper`, which is a singleton.

This PR adds a `ParsedGrammarKeywordHelper` implementation of the interface.
This allows the keyword helper to respond to changes made to the grammar.

Using the mapping of rule name to keyword map we populate this at runtime.
```c++
	unordered_map<string, reference<case_insensitive_set_t>> keyword_maps {
	    {"ReservedKeyword", reserved_keyword_map},  {"UnreservedKeyword", unreserved_keyword_map},
	    {"ColumnNameKeyword", colname_keyword_map}, {"TypeFuncKeyword", typefunc_keyword_map},
	    {"TypeNameKeyword", typename_keyword_map},
	};
```

The `TypeFuncKeyword` rule is added to the grammar, and some small tweaks were made to the generator script to support this. (generated trampoline ops checks had to be adjusted)

This does mean that parser extensions are still limited to our predefined set of keyword categories.
In the future we might be able to extend this, if we make our keyword categories strings and add the ability for an extension to provide their own `PEGKeywordHelper` implementation.


src/include/duckdb/parser/peg/keyword_helper/parsed_grammar_keyword_helper.hpp
```diff
@@ -0,0 +1,26 @@
...
+	bool KeywordCategoryType(const string &text, PEGKeywordCategory type) const override;
+	bool IsKeyword(const string &text) const override;
+	vector<ParserKeyword> KeywordList() const override;
+
+private:
+	case_insensitive_set_t reserved_keyword_map;
+	case_insensitive_set_t unreserved_keyword_map;
+	case_insensitive_set_t colname_keyword_map;
+	case_insensitive_set_t typefunc_keyword_map;
+	case_insensitive_set_t typename_keyword_map;
```
pdet: This seems fairly similar to `duckdb_keyword_helper.hpp` maybe we could reuse?

src/parser/peg/keyword_helper/parsed_grammar_keyword_helper.cpp
```diff
@@ -0,0 +1,101 @@
...
+		return colname_keyword_map.count(text) != 0;
+	case PEGKeywordCategory::KEYWORD_TYPE_NAME:
+		return typename_keyword_map.count(text) != 0;
+	default:
+		return false;
+	}
+}
+
+bool ParsedGrammarKeywordHelper::IsKeyword(const string &text) const {
+	return reserved_keyword_map.count(text) != 0 || unreserved_keyword_map.count(text) != 0 ||
```
pdet: Should we add `|| typename_keyword_map.count(text) != 0`?

src/parser/peg/keyword_helper/parsed_grammar_keyword_helper.cpp
```diff
@@ -0,0 +1,101 @@
...
+		result.push_back({kw, KeywordCategory::KEYWORD_UNRESERVED});
+	}
+	for (auto &kw : typefunc_keyword_map) {
+		result.push_back({kw, KeywordCategory::KEYWORD_TYPE_FUNC});
+	}
+	for (auto &kw : colname_keyword_map) {
+		result.push_back({kw, KeywordCategory::KEYWORD_COL_NAME});
+	}
+	return result;
+}
```
pdet: These functions seem very similar to `duckdb_keyword_helper.cpp`, maybe some reuse is possible?

pdet (commented):
Thanks, some minor comments.

Dtenwolde (approved):

PR #24876, opened 2026-08-19
Prune row group for `list_contains` function

dentiny:
Hi team, I found for `list_contains` operation, row groups are not correctly pruned, for example
```sql
memory D COPY (SELECT [i] AS l FROM range(6144) t(i))
           TO '/tmp/pd_list_contains.parquet' (FORMAT PARQUET, ROW_GROUP_SIZE 2048);
memory D EXPLAIN ANALYZE SELECT count(*) FROM '/tmp/pd_list_contains.parquet'
         WHERE list_contains(l, 3000);
╭─ Summary ───────────╮
│ Total Time: 0.0013s │
│ Data Read:  41.8 kB │
╰─────────────────────╯
│ Function: PARQUET_SCAN          │
│ Filters: list_contains(l, 3000) │
│ Total Files Read: 1             │
│ Filename(s):                    │
│ /tmp/pd_list_contains.parquet   │
```
The above query should be reading only one row group, but actually all row groups are accessed.

In this PR, I added the filter-prune callback, which consulted list child min/max when exists to prune impossible row groups.


dentiny:
image
CI failure doesn't seem to related to my change

src/function/scalar/list/contains_or_position.cpp
```diff
@@ -21,10 +25,44 @@ static void ListSearchFunction(DataChunk &input, ExpressionState &state, Vector
...
+		needle_expr = &BoundCastExpression::Child(needle_expr->Cast<BoundFunctionExpression>());
+	}
+	if (needle_expr->GetExpressionType() != ExpressionType::VALUE_CONSTANT) {
+		return FilterPropagateResult::NO_PRUNING_POSSIBLE;
+	}
+	auto needle = needle_expr->Cast<BoundConstantExpression>().GetValue();
+
+	auto &child = ListStats::GetChildStats(*list_stats);
+	if (!list_stats->CanHaveNoNull() || !child.CanHaveNoNull() || needle.IsNull()) {
+		return FilterPropagateResult::FILTER_ALWAYS_FALSE;
```
Mytherin: Should this not be `FILTER_FALSE_OR_NULL`?

src/function/scalar/list/contains_or_position.cpp
```diff
@@ -21,10 +25,44 @@ static void ListSearchFunction(DataChunk &input, ExpressionState &state, Vector
...
 
+static FilterPropagateResult ListContainsFilterPrune(const FunctionStatisticsPruneInput &input) {
+	auto list_stats = input.ChildStats(0);
+	if (!list_stats || list_stats->GetStatsType() != StatisticsType::LIST_STATS) {
+		return FilterPropagateResult::NO_PRUNING_POSSIBLE;
+	}
+
+	const Expression *needle_expr = input.function.GetChildren()[1].get();
+	while (BoundCastExpression::IsCast(*needle_expr)) {
+		needle_expr = &BoundCastExpression::Child(needle_expr->Cast<BoundFunctionExpression>());
```
Mytherin: I'm not sure if skipping casts blindly is always safe here.

Can't we look at the stats of the needle instead of getting the constant value?

Mytherin: I also wonder if this is required at all given the current setup - if needle is a constant wouldn't the constant folding remove the casts anyway?

src/function/scalar/list/contains_or_position.cpp
```diff
@@ -21,10 +23,71 @@ static void ListSearchFunction(DataChunk &input, ExpressionState &state, Vector
...
+	if (child.GetStatsType() == StatisticsType::NUMERIC_STATS &&
+	    needle_stats->GetStatsType() == StatisticsType::NUMERIC_STATS) {
+		auto zonemap = StatisticsPropagator::PropagateComparison(child, *needle_stats, ExpressionType::COMPARE_EQUAL);
+		if (zonemap == FilterPropagateResult::FILTER_ALWAYS_FALSE ||
+		    zonemap == FilterPropagateResult::FILTER_FALSE_OR_NULL) {
+			return filter_false;
+		}
+	} else if (child.GetStatsType() == StatisticsType::STRING_STATS &&
+	           needle_stats->GetStatsType() == StatisticsType::STRING_STATS) {
+		if (StringStats::HasMin(*needle_stats) &&
```
Mytherin: Instead of having this extra code here - could we move this into `StatisticsPropagator::PropagateComparison` and just use that in this function? That would more generically solve the problem.

src/function/scalar/list/contains_or_position.cpp
```diff
@@ -21,10 +23,71 @@ static void ListSearchFunction(DataChunk &input, ExpressionState &state, Vector
...
+			auto needle_value =
+			    child.GetType().id() == LogicalTypeId::BLOB ? Value::BLOB_RAW(needle_min) : Value(needle_min);
+			auto zonemap = StringStats::CheckZonemap(child, ExpressionType::COMPARE_GREATERTHANOREQUALTO,
+			                                         array_ptr<const Value>(&needle_value, 1));
+			if (zonemap == FilterPropagateResult::FILTER_ALWAYS_FALSE) {
+				return filter_false;
+			}
+		}
+		if (StringStats::HasMax(*needle_stats) &&
+		    StringStats::GetMaxType(*needle_stats) == StringStatsType::EXACT_STATS) {
```
Mytherin: We should use `StringStats::TryGetValidMax` instead of checking for `EXACT_STATS` to still allow this optimization to work for truncated min/max.

Mytherin (commented):
Thanks for the updates - looks good, some more minor comments then this is ready

src/function/scalar/list/contains_or_position.cpp
dentiny: Sounds good! Add string stats support for comparison propagation

src/optimizer/statistics/expression/propagate_comparison.cpp
```diff
@@ -3,13 +3,107 @@
...
 #include "duckdb/optimizer/expression_rewriter.hpp"
+#include "duckdb/storage/statistics/string_stats.hpp"
 
 namespace duckdb {
 
+static Value TryGetStringStatsMin(const BaseStatistics &stats) {
+	if (stats.GetType().id() == LogicalTypeId::VARCHAR) {
+		return StringStats::TryGetValidMin(stats);
+	}
+	if (stats.GetType().id() == LogicalTypeId::BLOB && StringStats::GetMinType(stats) == StringStatsType::EXACT_STATS) {
```
Mytherin: We don't need to filter on `EXACT_STATS` here. `EXACT_STATS` means that the minimum and maximum value occur within the dataset. That's different from `TRUNCATED_STATS` where the stats are potentially wider than the values that occur in the dataset.

For example, if our max is `aaaaaaaaaaaaaaaaaaaaaaaaaaaa`, we might choose to have `aaaaaaaaaaaaaaaaab` as our truncated max. That value is still larger than all values in the domain, it's just not present in the values.

Whether or not the min/max stats are exact or not is not relevant for this optimization here, so this check can be left out.

src/optimizer/statistics/expression/propagate_comparison.cpp
```diff
@@ -3,13 +3,107 @@
 #include "duckdb/planner/expression/bound_comparison_expression.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
 #include "duckdb/optimizer/expression_rewriter.hpp"
+#include "duckdb/storage/statistics/string_stats.hpp"
 
 namespace duckdb {
 
+static Value TryGetStringStatsMin(const BaseStatistics &stats) {
+	if (stats.GetType().id() == LogicalTypeId::VARCHAR) {
+		return StringStats::TryGetValidMin(stats);
```
Mytherin: `StringStats::TryGetValidMin` exists to get a valid UTF8 value in case we need to move that value into e.g. a query plan as a predicate to compare to varchar values. I think we can just always use `BLOB_RAW` here to create these values - we don't need to create a valid UTF8 value to do the min/max value filtering here for varchars since we're only using them in a single comparison function with other blobs.

Mytherin (commented):
Thanks for the fixes - looks great, some more minor comments then this is good to go.

Mytherin:
Thanks!

PR #24878, opened 2026-08-19
Always quote identifiers in error messages

JelteF:
`Identifier` now formats in exceptions through `SQLQuotedIdentifier` instead of `SQLIdentifier`. That way an identifier interpolated into an error message is always quoted rather than only when quoting is required in SQL. This makes it clear that a word is a name of an object and not a normal part of the sentence.

Most format strings already wrapped their `%s` in literal quotes, which was only correct as long as `Identifier` did not need quotes in SQL. Any name that needed quoting, for example one containing a space or special character, already produced doubled quotes like `""my db""`. Those literal quotes are now removed wherever the argument is an `Identifier`, leaving the quoting to the formatter, which also escapes an embedded `"` correctly.

Where the identifier is only one component of a larger construct - a dotted `catalog.schema.table` path, or the `col INTEGER AS (...)` snippet in a generated-column error - quoting every component would be noise, so those keep `SQLIdentifier` (quote only when required) and the surrounding construct is wrapped in single quotes instead.

Call sites that hand-quoted `X.GetIdentifierName()` now pass the `Identifier` itself, so they escape correctly instead of emitting a raw `"`.

### DIFF
--- a/.github/config/extensions/aws.cmake
+++ b/.github/config/extensions/aws.cmake
@@ -3,5 +3,6 @@ if (NOT MINGW AND NOT ${WASM_ENABLED})
             LOAD_TESTS
             GIT_URL https://github.com/duckdb/duckdb-aws
             GIT_TAG 6c044194df8f9fa6ebc4365aa11c75095c1d5ef2
+            APPLY_PATCHES
             )
 endif()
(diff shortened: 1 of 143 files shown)


Mytherin:
Thanks! Great to get this consistent everywhere

PR #24879, opened 2026-08-19
Preserve potentially failing DATE-to-TIMESTAMP casts unless column statistics prove the conversion safe

### LINKED ISSUE #24241: Timestamp cast displays different result when optimizer disabled
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t(d DATE);
memory D INSERT INTO t VALUES
           (DATE '294247-01-10'),
           (DATE '294247-01-11'),
           (DATE '5874897-12-31'),
           (DATE 'infinity'),
           (NULL);
memory D 
memory D SELECT d
         FROM t
         WHERE CAST(d AS TIMESTAMP) > TIMESTAMP '294247-01-10 04:00:00'
         ORDER BY d;
┌───────────────┐
│       d       │
│     date      │
├───────────────┤
│ 294247-01-11  │
│ 5874897-12-31 │
│ infinity      │
└───────────────┘

┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
memory D SELECT d
         FROM t
         WHERE CAST(d AS TIMESTAMP) > TIMESTAMP '294247-01-10 04:00:00'
         ORDER BY d;
Conversion Error:
Type INT32 with value 294247-01-11 can't be cast to the destination type INT64 when casting from source column d

LINE 3: WHERE CAST(d AS TIMESTAMP) > TIMESTAMP '294247-01-10 04:00:00'
              ^^^^^^^^^^^^^^^^^^^^
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Alignyx:
Fix #24241 

### Problem                                                                                                                                                                                                                                                  

DuckDB’s comparison simplification optimizer could incorrectly remove a potentially failing `DATE → TIMESTAMP` cast.

A simplified version of the reported query is:

```
  SELECT d
  FROM dates
  WHERE CAST(d AS TIMESTAMP) >
        TIMESTAMP '294247-01-10 04:00:00';
```

DuckDB’s `DATE` type stores a signed 32-bit number of days, while `TIMESTAMP` stores a signed 64-bit number of microseconds. Despite using a wider integer representation, `TIMESTAMP` covers a much smaller range because each day must be multiplied by the number of microseconds per day. Consequently, some valid finite `DATE` values, such as `DATE '294247-01-11'` cannot be represented as a `TIMESTAMP`. An explicit cast must therefore raise a conversion error. `TRY_CAST` must return `NULL` for the same value.

### Incorrect optimizer behavior

With the expression rewriter enabled, DuckDB transformed `CAST(d AS TIMESTAMP) > TIMESTAMP '294247-01-10 04:00:00'` into a direct `DATE` comparison similar to `d > DATE '294247-01-10'`
               
The rewritten comparison no longer executed the original cast. As a result:

  - `CAST` stopped raising an error for out-of-range finite dates.
  - `TRY_CAST`, which should have produced `NULL`, behaved like a normal comparison.
  - The optimized and unoptimized execution paths returned different results.

Disabling only the `expression_rewriter` optimizer reproduced the correct conversion error, which isolated the defect to ComparisonSimplificationRule.

### Root cause

The rule used `CastIsInvertible` to decide whether a cast could be moved from the column side to the constant side. However, invertibility and totality are different properties:

  - An invertible cast does not map two successfully converted source values to the same target value.
  - A total cast succeeds for every possible source value.

`DATE → TIMESTAMP` is invertible for values that convert successfully, but it is not total because sufficiently distant finite dates overflow the TIMESTAMP range. The optimizer checked the first property but implicitly assumed the second one. Infinity also requires special treatment. DuckDB intentionally maps: `DATE 'infinity' → TIMESTAMP 'infinity'`

Therefore, merely checking whether the statistical minimum and maximum can be converted is insufficient. If the maximum is `DATE 'infinity'`, that endpoint converts successfully while an out-of-range finite date immediately below it may still fail.

### Fix

The comparison simplification rule now first asks whether the cast can throw.

For casts that cannot throw, the existing optimization continues unchanged. For a potentially throwing `DATE → TIMESTAMP` cast, the optimizer removes the cast only when it can prove that every value read from the column is representable as a `TIMESTAMP`.

The proof works as follows:

  1. The cast operand must be a direct, non-correlated column reference.
  2. The optimizer locates the corresponding table scan.
  3. It obtains the column’s min/max statistics through the table-scan statistics callback.
  4. Both bounds must be present and finite.
  5. Both bounds must successfully convert to `TIMESTAMP`.

For finite dates, conversion to midnight `TIMESTAMP` is monotonic, and the representable source domain is contiguous. Therefore, if both finite endpoints of the column’s range convert successfully, every date between them also converts successfully.

If statistics are unavailable, either endpoint is infinite, or either conversion fails, the optimizer conservatively keeps the original cast.

This is also safe for transaction-local modifications: DuckDB’s table-scan statistics callback returns no statistics when the table has outstanding local changes, so newly inserted values cannot be missed by the proof.

### Resulting behavior

For ordinary tables containing dates such as `2024` or `2025`, DuckDB can still simplify the predicate to a direct `DATE` comparison and retain the existing optimization.

For a table containing an out-of-range finite date, the plan keeps `CAST(d AS TIMESTAMP)` and execution raises the same conversion error regardless of whether the expression rewriter is enabled.

For `TRY_CAST`, failed conversions remain `NULL` instead of being replaced by a direct `DATE` comparison.

DuckDB may still derive an optional `DATE` filter for zonemap pruning, but that filter is only an optimization hint. The authoritative `CAST` expression remains in the plan and preserves the required error or `NULL` semantics.

### DIFF
--- a/src/optimizer/rule/comparison_simplification.cpp
+++ b/src/optimizer/rule/comparison_simplification.cpp
@@ -8,7 +8,10 @@
 #include "duckdb/optimizer/expression_rewriter.hpp"
 #include "duckdb/planner/expression/bound_comparison_expression.hpp"
 #include "duckdb/planner/expression/bound_conjunction_expression.hpp"
+#include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/expression/bound_operator_expression.hpp"
+#include "duckdb/planner/operator/logical_get.hpp"
+#include "duckdb/storage/statistics/numeric_stats.hpp"
 
 namespace duckdb {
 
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - but it's ok if optimizers remove errors, so the current behavior seems fine to me.

PR #24881, opened 2026-08-19
Return the aggregate's empty-input result for correlated subqueries

SKeyerror:
`count` returns 0 when aggregating zero rows, and so do `regr_count`,
`approx_count_distinct` and `entropy`. In a correlated subquery over an empty
group, however, only `count`/`count_star` returned 0 — the others returned NULL:

```sql
CREATE TABLE t1 AS SELECT * FROM (VALUES (1), (99)) t(k);
CREATE TABLE t2 AS SELECT * FROM (VALUES (1, 10), (1, 20)) t(k, v);
       
SELECT k,
       (SELECT count(*) FROM t2 WHERE t2.k = t1.k),
       (SELECT regr_count(t2.k, t2.v) FROM t2 WHERE t2.k = t1.k),
       (SELECT approx_count_distinct(t2.v) FROM t2 WHERE t2.k = t1.k),
       (SELECT entropy(t2.v) FROM t2 WHERE t2.k = t1.k)
FROM t1 ORDER BY k;
-- 1  | 2 | 2    | 2    | 1.0
-- 99 | 0 | NULL | NULL | NULL   <-- should be 0, 0, 0.0
```

  The planner recognized the two count functions by comparing name plus callbacks
  — which cannot reach the aggregates in core_functions — and hardcoded 0 as the
  replacement value.
  
  This derives it from callbacks every aggregate already has: initializing a state
  and finalizing it without aggregating a row is the result for zero input rows.
  PushDownAggregate uses that Value to 
  
  - force the LEFT join that keeps a row for every delim key (entropy propagates
    NULL values, so it got an INNER join and its row was dropped entirely), and
  - rewrite the reference into CASE WHEN x IS NULL THEN <empty result> ELSE x END.
  
  No aggregate declares anything, no name or callback comparisons remain, and
  extension aggregates are covered automatically. Volatile aggregates are excluded
  — finalizing them can have observable effects, so they may only be evaluated
  during execution. 
  
  Sweeping every single- and two-argument aggregate shows no remaining mismatch
  between the plain empty-input result and the correlated one. The new test covers
  the four affected aggregates, DISTINCT/FILTER modifiers, and a guard that
  sum/list still return NULL.
  
  Note: entropy is registered with DEFAULT_NULL_HANDLING yet returns 0.0 for
  zero rows, contradicting what VerifyNullHandling documents. This PR does not
  depend on that annotation — happy to fix it here or separately.

The planner no longer compares function names or callback pointers here.

### DIFF
--- a/src/include/duckdb/planner/subquery/flatten_dependent_join.hpp
+++ b/src/include/duckdb/planner/subquery/flatten_dependent_join.hpp
@@ -84,7 +84,7 @@ private:
 	void AddReplacementAliases(const BindingReplacementGraph &replacements);
 	Binder &binder;
 	column_binding_map_t<ColumnBinding> correlated_aliases;
-	column_binding_map_t<idx_t> replacement_map;
+	column_binding_map_t<Value> replacement_map;
 	const CorrelatedColumns &correlated_columns;
 	vector<LogicalType> delim_types;
 
--- a/src/include/duckdb/planner/subquery/rewrite_correlated_expressions.hpp
+++ b/src/include/duckdb/planner/subquery/rewrite_correlated_expressions.hpp
@@ -31,15 +31,16 @@ private:
 	column_binding_map_t<ColumnBinding> &correlated_aliases;
 };
 
-//! Helper class that rewrites COUNT aggregates into a CASE expression turning NULL into 0 after a LEFT OUTER JOIN
+//! Helper class that rewrites aggregates returning a non-NULL result for zero rows into a CASE expression turning
+//! the NULL of a LEFT OUTER JOIN back into that result
 class RewriteCountAggregates : public LogicalOperatorVisitor {
 public:
-	static void Rewrite(LogicalOperator &op, column_binding_map_t<idx_t> &replacement_map);
+	static void Rewrite(LogicalOperator &op, column_binding_map_t<Value> &replacement_map);
 
 private:
-	explicit RewriteCountAggregates(column_binding_map_t<idx_t> &replacement_map);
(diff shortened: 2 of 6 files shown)


Mytherin:
Thanks - instead of having a flag this specific, maybe we can add a callback that returns a constant `Value` for empty inputs (which is set to `Value(0)` for the relevant functions here)?

Mytherin:
Or better yet - instead of adding a new callback maybe we can just use the existing callbacks (`init` -> `finalize`) to obtain this information

SKeyerror:
> Or better yet - instead of adding a new callback maybe we can just use the existing callbacks (`init` -> `finalize`) to obtain this information

You mean calculate the result dynamically? If so, it involves more code changes, but it offers a more generic abstraction layer and off-the-shelf aggregate paths.

Mytherin:
I meant to call `init -> finalize` in the flatten dependent joins to figure out if the empty result returns `NULL` or something else (`0`).

SKeyerror:
@Mytherin Hi, I added a new func of `AggregateEmptyResult()`, please take a look again.

Mytherin:
Thanks!

SKeyerror:
@Mytherin Could you approve the CI run again? I've fixed the `verify_aggregate_state_export.json` failures (skipped plan-time finalize for export-state aggregates and added the new test to the config's skip list).

SKeyerror:
@Mytherin Hi, I have fixed the header conflicts in `flatten_dependent_join.cpp`, all tests have passed in my local env, please take a look again, thanks!

Issue #24871, opened 2026-08-19
ARM64 (aarch64) portability issues: UB in interpolation, equi_width_bins hang, and related divergences


u70b3:
# ARM64 (aarch64) portability issues: UB in interpolation, binning hang, and related divergences

Found via a targeted audit of risky code patterns on aarch64 (Linux, GCC 11). On ARM, `char` is unsigned by default, `fcvtzs` saturates out-of-range float→int conversions (x86 `cvttsd2si` yields the integer-indefinite value instead), integer division overflow does not trap, and the memory model is weaker than x86 TSO — so several latent bugs either diverge across platforms or only manifest on one of them.

All scenarios below are reproducible end-to-end via SQL. Verified on aarch64 Linux with v1.6.0-dev12869 (61e00e80cc).

## 1. `equi_width_bins` hangs forever (uninterruptible) when the bin step truncates to 0 — HIGH

```sql
SELECT equi_width_bins(0, 1, 1000000, false);
SELECT equi_width_bins(0, 1, 1000000, true);
SELECT equi_width_bins(TIMESTAMP '2020-01-01', TIMESTAMP '2020-01-01 00:00:00.0005', 1000000, false);
```

In `EquiWidthBinsInteger::Operation` (`extension/core_functions/scalar/generic/binning.cpp`), the internal step is computed as `span / bin_count` in hugeint arithmetic. When `span * 1000 < bin_count` the step truncates to 0, and the loop `for (bin_boundary = max; bin_boundary > min; bin_boundary -= step)` never advances (the duplicate-boundary `continue` path keeps it spinning). The loop contains no `InterruptCheck`, so **`SET max_execution_time` cannot abort the query** — it is a full hang (verified: process had to be killed with SIGKILL after surviving the execution-time limit).

The double path has a `step == 0` guard (throwing an internal error); the integer path does not. Platform-independent, but reported here as part of the same audit.

Expected: an `InvalidInputException` ("bin count is too large for the range"), matching the spirit of the double-path guard.

## 2. Window `FILL` interpolation overflows at integer bounds — results differ between x86 and ARM — HIGH

```sql
SELECT x, fill(y ORDER BY x) OVER () FROM
  (VALUES (0::BIGINT, 9223372036854775807::BIGINT), (1::BIGINT, NULL::BIGINT), (2::BIGINT, 9223372036854775807::BIGINT)) t(x,y)
ORDER BY x;
```

[...]

PR #24883, opened 2026-08-19
ARM64 portability: fix UB with observable behavior in interpolation, lead/lag and casts

u70b3:
Split out from #24872 (behavior-affecting half). See #24871 for the full ARM64 audit context.

These changes fix undefined behavior whose **observable results differ between platforms** (x86 vs ARM, and MSVC). Each fix comes with a black-box sqllogictest locking in the behavior at the type bounds.

### `quantile_cont` / window `FILL` at INT64/TIMESTAMP bounds
The interpolation computed `lo + delta*d` in floating point and cast the possibly out-of-range result back to int64 — UB that is platform-dependent (x86 `cvttsd2si` yields the integer indefinite value, ARM `fcvtzs` saturates), so the same query returned different results on x86 and ARM.

`InterpolateInt64` now computes exactly in integers when `hi - lo` fits in int64 — the result stays within `[lo, hi]`, so it cannot overflow and is platform-independent (this also fixes MSVC, where `long double` is only 64 bits and could not represent near-max int64 endpoints at all). The `long double` path remains as a fallback for overflowing deltas, now with a clamp before the cast.

### `lead`/`lag` with offsets at BIGINT bounds
Two issues: negating `INT64_MIN` in the streaming path is UB, and the non-streaming per-row index addition overflowed or not **depending on chunk boundaries** — at `STANDARD_VECTOR_SIZE=2048` the batch default-fill skips the addition for later rows, at `=2` it throws. The streaming path now declines `INT64_MIN`, and the row index saturates instead of throwing, so every row lands outside the partition and yields the default.

Behavior change: `lag(v, INT64_MIN)` now returns the default instead of an `Out of Range` error — consistent with `lead(v, INT64_MAX)` and independent of vector size.

### Casts
- boolean cast: pass `unsigned char` to `std::tolower` (UB for non-ASCII bytes)

The `equi_width_bins` fixes this PR originally included were dropped after rebasing: the hangs were independently fixed on the base branch (efbaab593d, 9da4ce1ac) with degrade-gracefully semantics. The remaining, behavior-neutral hardening from #24872 (memory orderings, memcpy-ization, typed ALP buffers, ctype guards, ...) is in a separate PR.

### Maintainer decision needed: `.5` tie-break change in temporal interpolation

The exact-integer path in `InterpolateInt64` (used by `quantile_cont`/`median` on `TIME`/`TIMESTAMP`) changes the rounding of exact `.5` ties when the interpolated value is **negative** — in practice, pre-1970 timestamps (`TIME` is always non-negative and unaffected; window `FILL` truncates and is unaffected).

- Old: `llround(lo*(1-d) + hi*d)` rounds the interpolated *total* half **away from zero**
- New: `lo + llround((hi-lo)*d)` rounds the *offset* half away from zero, i.e. the result half **toward the upper endpoint**

Verified locally against the pre-change build:

```sql
SELECT quantile_cont(y, 0.5) FROM (VALUES
  (TIMESTAMP '1969-12-31 23:59:59.999998'),
  (TIMESTAMP '1970-01-01 00:00:00.000001')) t(y);
-- before: 1969-12-31 23:59:59.999999   (midpoint -0.5µs rounded away from zero)
-- after:  1970-01-01 00:00:00          (offset +1.5µs rounded up)
```

Ties with a positive midpoint are unchanged (rounding a positive total away from zero already lands on the upper endpoint), as is every non-tie result — the flip only affects an exact half-tie whose midpoint is negative.

The flip is observable in the existing suite: `quantile_cont(ts, 0.25)` over `{-infinity, epoch, infinity}` interpolates at an exact .5 tie with a negative midpoint, so `test/sql/types/timestamp/test_infinite_time.test` shifts by 1µs (`...612096` → `...612097`). Those expectations are updated here, and the rule above is locked in by new cases in `test/sql/aggregate/aggregates/test_quantile_cont_bounds.test`.

I believe the new rule is the better semantics: it is sign-independent, falls out of the exact integer computation naturally, matches what `INTERVAL` interpolation already does (`MultiplyByDouble` rounds the offset), and is identical on every platform (the old rule could not be kept exactly once the computation moved into integer space — emulating it would reintroduce the platform-dependent float rounding this PR removes). **Maintainers: please confirm this tie-break flip is acceptable**, or let me know if you need bit-for-bit preservation of the old rule and I will look into emulating it.


Mytherin:

> Maintainer decision needed: .5 tie-break change in temporal interpolation

> The exact-integer path in InterpolateInt64 (used by quantile_cont/median on TIME/TIMESTAMP) changes the rounding of exact .5 ties when the interpolated value is negative — in practice, pre-1970 timestamps (TIME is always non-negative and unaffected; window FILL truncates and is unaffected).

CC @hawkfish could you have a look here? My intuition is that it's alright but it is a breaking change of course.

u70b3:
One more data point for the tie-break decision: the new rule is the one DuckDB already uses for `INTERVAL` interpolation. The `interval_t` specialization of `InterpolateOperator` in `src/function/scalar/operator/arithmetic.cpp` computes `lo + MultiplyByDouble(hi - lo, d)`, and `MultiplyByDouble` is `llround(delta * d)` — i.e. it rounds the *offset* half away from zero and adds it to `lo`, exactly what `InterpolateInt64` now does for `TIME`/`TIMESTAMP`. The old timestamp behavior (rounding the interpolated *total*) was the outlier among DuckDB's interpolating types; this PR aligns temporal interpolation on a single, sign-independent rule.

And since this is a (narrow) breaking change: I'm happy to draft the release-notes entry for it (e.g. "`quantile_cont`/`median` on `TIME`/`TIMESTAMP`: `.5` tie-break flips for negative midpoints") — just let me know the preferred format and place.

u70b3:
@hawkfish gentle ping — when you have a moment, could you weigh in on the `.5` tie-break question above? (exact-integer interpolation flips rounding for negative midpoints in `quantile_cont`/`median` on `TIME`/`TIMESTAMP`; `INTERVAL` already uses the new rule.) It's the one open decision on this PR — everything else is green. If you'd rather keep bit-for-bit compatibility with the old rule, I'm happy to look into emulating it instead.

u70b3:
Rebased onto latest `v2.0-cyanoptera` and **dropped the `equi_width_bins` changes** from this PR.

The infinite-loop hangs this PR was addressing have been fixed on the base branch in the meantime, with degrade-gracefully semantics rather than the error this PR threw:

- efbaab593d (fixes #24282): clamps the integer step to the smallest possible step when the bin count truncates it to zero, and returns only the max boundary when the double step is 0 (this also removed the `InternalException` that was reachable from user input)
- 9da4ce1ac: guards the double-path progress invariant (`max - step >= max` → return only the max boundary), which also covers the `0 < step < ulp(boundary)` case

Since the hangs are gone and upstream has chosen graceful degradation, keeping the `InvalidInputException` version here would only introduce a conflicting third behavior — removed the `binning.cpp` changes and `test_equi_width_bins_bounds.test` accordingly.

The rest applies cleanly on the new base: the `lead`/`lag` UB (INT64_MIN negation in the streaming path, chunk-size-dependent overflow) and the interpolation UB (platform-dependent out-of-range float→int64 cast) are both still present upstream, so those fixes are still needed. The PR is now down to three small, independent commits: interpolation, lead/lag, boolean cast.

The one remaining open decision is the `.5` tie-break in temporal interpolation from the PR description (@hawkfish pinged above) — everything else is ready from my side.

PR #24884, opened 2026-08-19
ARM64 portability hardening: mechanical UB fixes without behavior change

u70b3:
Split out from #24872 (behavior-neutral half). See #24871 for the full ARM64 audit context. The behavior-affecting fixes (interpolation, binning hang, lead/lag bounds) are in #24883.

These changes remove undefined or implementation-defined behavior **without changing any observable results**:

- **checksum / uuid**: go through `Load`/`Store` instead of `reinterpret_cast` (alignment/aliasing UB)
- **`StringUtil::CIHash`**: zero-extend bytes >= 0x80 so the hash is identical on signed-char (x86) and unsigned-char (ARM) platforms — comes with a C++ unit test locking in the exact hash values
- **join hashtable / copy-to-file / user settings**: correct atomic memory orderings in publish/read patterns. The `InsertRowToEntry` EXPECT_EMPTY CAS had the two orderings inverted for their roles: success is a publish (the new row's next-pointer is initialized immediately before → release), failure is a load whose result is immediately dereferenced for key comparison (→ acquire) — the else-branch CAS just below already used release-on-success. Probe/insert loads feeding row-pointer dereferences become acquire; `WaitAll`'s flag load pairs with the release in workers' `--pending_tasks`; `user_settings` version check becomes acquire (payload is mutex-protected today; model hygiene). Zero codegen cost on x86-64 (TSO); on ARM64 `ldr`→`ldar` on cache-miss-dominated paths — measured in **Verification** below.
- **ALP/ALPRD**: typed buffers instead of byte-buffer casts — decompression side, plus the compress-side `values_encoded` (was `uint8_t[]` feeding `PackBuffer<uint64_t>`, whose fastpack writes `uint32_t` units; now typed `uint32_t[]`). Byte size and on-disk layout unchanged.
- **`SignedLength`**: unsigned arithmetic instead of `(value ^ sign) - sign`
- **left shift**: guard `0 << width`
- **ctype calls** (`isdigit`/`isalnum`/`isalpha`): pass `unsigned char`
- **SLURM memory limit parsing**: guard the double→`idx_t` conversion against overflow. The clamp boundary uses `>=` because `double(UINT64_MAX)` rounds up to 2^64 — an input landing exactly on the boundary otherwise produced exactly 2^64, an out-of-range conversion with platform-divergent results (x86: 2^63, ARM64: `UINT64_MAX`) — and `nan` is rejected (fast_float parses it, and NaN bypasses every comparison into the same cast). Covered by a new unit test (suffixes, invalid/NaN input, the exact clamp boundary).

### Verification (128-core ARM64, Kunpeng-class, gcc 11.4, reldebug)

This PR (`523094ee9e`) vs its base commit (`326202015f`), interleaved runs on an idle machine:

- Fast unittest suite: zero failures on both sides — 4,959 vs 4,961 cases (the +2 are this PR's new unit tests), identical 434-case skip set.
- Hash join, 40M-row build (one publish CAS/row) + 200M-row probe (one acquire load/row), 5 interleaved runs: median **1.200s (base) vs 1.186s (PR)**, means identical at 2.009s.
- `COPY (30M rows) TO PARQUET` (`WaitAll` path): median **7.453s (base) vs 7.189s (PR)**.

(CI note: the regression-bench failures on the first commit died in the workflow's Install step before any benchmark ran — infra flake, not a regression signal.)

### DIFF
--- a/extension/core_functions/scalar/operators/bitwise.cpp
+++ b/extension/core_functions/scalar/operators/bitwise.cpp
@@ -258,6 +258,9 @@ struct BitwiseShiftLeftOperator {
 		if (shift == 0) {
 			return input;
 		}
+		if (input == 0) {
+			return 0;
+		}
 		TA max_value = UnsafeNumericCast<TA>((TA(1) << (max_shift - shift - 1)));
 		if (input >= max_value) {
 			throw OutOfRangeException("Overflow in left shift (%s << %s)", NumericHelper::ToString(input),
--- a/src/common/checksum.cpp
+++ b/src/common/checksum.cpp
@@ -66,11 +66,10 @@ hash_t ChecksumRemainder(const void *ptr, size_t len) noexcept {
 
 uint64_t Checksum(const uint8_t *buffer, size_t size) {
 	uint64_t result = 5381;
-	auto ptr = reinterpret_cast<const uint64_t *>(buffer);
 	size_t i;
 	// for efficiency, we first checksum uint64_t values
 	for (i = 0; i < size / 8; i++) {
-		result ^= Checksum(ptr[i]);
+		result ^= Checksum(Load<uint64_t>(buffer + i * 8));
 	}
 	if (size > i * 8) {
 		// the remaining 0-7 bytes we hash using a string hash
--- a/src/common/string_util.cpp
+++ b/src/common/string_util.cpp
@@ -445,7 +445,8 @@ uint64_t StringUtil::CIHash(const string &str) {
 uint64_t StringUtil::CIHash(const char *str, idx_t size) {
 	uint32_t hash = 0;
 	for (idx_t i = 0; i < size; i++) {
-		hash += static_cast<uint32_t>(StringUtil::CharacterToLower(static_cast<char>(str[i])));
+		// convert through uint8_t so the hash is identical on platforms with signed and unsigned char
+		hash += static_cast<uint32_t>(static_cast<uint8_t>(StringUtil::CharacterToLower(static_cast<char>(str[i]))));
 		hash += hash << 10;
 		hash ^= hash >> 6;
 	}
--- a/src/common/types/uuid.cpp
+++ b/src/common/types/uuid.cpp
@@ -1,5 +1,6 @@
 #include "duckdb/common/types/uuid.hpp"
 #include "duckdb/common/chrono.hpp"
+#include "duckdb/common/helper.hpp"
 #include "duckdb/common/random_engine.hpp"
 
 namespace duckdb {
@@ -152,7 +153,7 @@ hugeint_t BaseUUID::Convert(const std::array<uint8_t, 16> &bytes) {
 hugeint_t UUIDv4::GenerateRandomUUID(RandomEngine &engine) {
 	std::array<uint8_t, 16> bytes;
 	for (int i = 0; i < 16; i += 4) {
-		*reinterpret_cast<uint32_t *>(bytes.data() + i) = engine.NextRandomInteger();
+		Store<uint32_t>(engine.NextRandomInteger(), bytes.data() + i);
 	}
 	// variant must be 10xxxxxx
 	bytes[8] &= 0xBF;
--- a/src/execution/join_hashtable.cpp
+++ b/src/execution/join_hashtable.cpp
@@ -762,15 +762,15 @@ static inline data_ptr_t InsertRowToEntry(atomic<ht_entry_t> &entry, const data_
 			StorePointer(nullptr, row_ptr_to_insert + pointer_offset);
 
 			ht_entry_t expected_entry;
-			entry.compare_exchange_strong(expected_entry, desired_entry, std::memory_order_acquire,
-			                              std::memory_order_relaxed);
+			entry.compare_exchange_strong(expected_entry, desired_entry, std::memory_order_release,
+			                              std::memory_order_acquire);
 
 			// The expected entry is updated with the encountered entry by the compare exchange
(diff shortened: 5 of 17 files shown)


u70b3:
Follow-up commit `523094ee9e`:

- **`ParseMemoryLimitSlurm` clamp boundary**: `double(UINT64_MAX)` rounds up to 2^64, so with `>` an input landing exactly on `double(idx_max)/multiplier` (e.g. `SLURM_MEM_PER_NODE=18446744073709552K`) produced exactly 2^64 in the multiply branch — an out-of-range double→`idx_t` conversion with platform-divergent results (x86: 2^63, ARM64: saturates to `UINT64_MAX`). Changed to `>=` so the boundary clamps to unlimited everywhere; also reject `nan` (fast_float parses it, and NaN bypasses every comparison into the same cast).
- **ALP compress side**: `values_encoded` was still `uint8_t[]` feeding `PackBuffer<uint64_t>` (fastpack writes `uint32_t` units) — same pattern already fixed here for ALPRD; now typed `uint32_t[]`. Byte size and on-disk layout unchanged.
- Added a `ParseMemoryLimitSlurm` unit test covering suffixes, invalid/NaN input, and the exact clamp boundary.

Verified locally (reldebug): new test, CIHash signedness test, bitwise ops, all fast ALP/ALPRD tests incl. `*_storage_version`, join/inner, decimal, uuid, hugeint suites all pass.

Mytherin:
Thanks for the PR! Most changes look good to me, one question about this one:

> join hashtable / copy-to-file / user settings: strengthen atomic memory orderings (publish CAS acquire→release, matching acquire loads)

Can you expand here? Which issue does this solve / why is this required? The reason I ask is that this might have a performance impact in particular on many core machines.

u70b3:
Happy to expand — all four sites are the same textbook pattern: a publisher writes a payload, then makes it visible through an atomic; a reader observes the atomic and dereferences the payload. On weakly-ordered hardware (ARM64) the existing orderings don't establish the required happens-before edges:

**join_hashtable.cpp** — the EXPECT_EMPTY CAS had the two orderings inverted relative to their roles. Success is a *publish*: `StorePointer(nullptr, row_ptr + pointer_offset)` initializes the new row's next-pointer immediately before, and only a release success ordering guarantees that store is visible to whoever observes the entry — with acquire, a concurrent prober can see the entry but read an uninitialized next-pointer / stale row bytes. Failure is a *read*: on collision `expected_entry` is updated to the encountered entry and the caller immediately dereferences its row for key comparison — a failed CAS is a load, so it needs acquire, not relaxed. (The else-branch CAS 15 lines below already used `memory_order_release` on success — this aligns the EXPECT_EMPTY branch with the idiom the file already relies on.) The two `load(relaxed → acquire)` sites feed paths that dereference the obtained row pointer, and chain-walk visibility is only transitive if every link acquires.

**physical_copy_to_file.cpp `WaitAll`** — flag-and-payload: workers write job/file state, then `--pending_tasks` (seq_cst RMW, i.e. release); the spinner read the flag relaxed and then consumed that state. On ARM the payload reads can be reordered before the flag read; the acquire load pairs with the release in the decrement.

**user_settings.cpp** — the payload copy is mutex-protected today, so this one is memory-model hygiene rather than a live race; it keeps the version check correct if payload ever moves out of the lock. Cost is one acquire load per cache miss.

**On performance** — measured on a 128-core ARM64 (Kunpeng-class) machine, gcc 11.4, reldebug, this PR (`523094ee9e`) vs its base commit (`326202015f`):

*Standard suite (default fast set):* zero failures on both sides — 4,959 vs 4,961 cases (the +2 are this PR's new unit tests), identical 434-case skip set.

*Hash join — 40M-row build (one publish CAS/row) + 200M-row probe (one acquire load/row), unique build keys (chains ≈1) to isolate the per-op ordering cost from chain-walk variance; 5 interleaved runs, wall-clock seconds. The op density is deliberately saturated — an upper bound on what any real workload could pay:*

| run | r1 | r2 | r3 | r4 | r5 | median | mean |
|---|---|---|---|---|---|---|---|
| base | 1.195 | 3.229 | 3.239 | 1.184 | 1.200 | **1.200** | 2.009 |
| PR | 3.229 | 1.186 | 1.165 | 3.307 | 1.157 | **1.186** | 2.009 |

*COPY (30M rows → PARQUET, `WaitAll` path), seconds:*

| run | r1 | r2 | r3 | r4 | r5 | median | min |
|---|---|---|---|---|---|---|---|
| base | 7.453 | 8.462 | 7.219 | 9.327 | 7.217 | **7.453** | 7.217 |
| PR | 7.219 | 7.189 | 7.188 | 7.048 | 7.189 | **7.189** | 7.048 |

The join's bimodality (~1.2s vs ~3.2s, at 63s vs 300s user CPU) is thread-scheduling variance on this 128-core box — the interleaving kept it fair: both modes appear in both binaries, and the means are identical to three decimals. On x86-64 the change is free by construction (TSO — it only constrains compiler reordering, and `lock cmpxchg` is a full barrier either way); on ARM64 it's `ldr`→`ldar` on cache-miss-dominated paths. I'd also argue the many-core concern cuts the other way: these publication races only manifest under weak ordering *and* high concurrency. Happy to split the ordering changes into a separate PR if you'd rather benchmark them in isolation.

(CI note: the 4 regression-bench failures on the first commit died in the workflow's Install step before any benchmark ran — infra flake, not a regression signal. Current head is green. Condensed summary folded into the PR description.)

Mytherin:
Makes sense - thanks for the explanation and benchmark!

PR #24888, opened 2026-08-19
Reserve column_ids.size() in CollectionScanState::Initialize

jeewonhh:
`CollectionScanState::Initialize` calls `column_scans.reserve(column_scans.size())`, but the line above asserts the vector is empty, so the reserve is always `reserve(0)` and does nothing. The intent was `column_ids.size()`.

Without it the vector reallocates its way up while emplacing one `ColumnScanState` per scanned column.


hannes:
thanks!

PR #24889, opened 2026-08-19
Use container in regression CI jobs

smvv:
Use a docker image with pre-installed tools to avoid installing many tools in CI.

Installing tools in CI could slow down or time out the job (making CI unreliable), so we can install those tools once and use the container.

We need to use the manylinux image also for the `linux-release` and `linux-base` because:
- when the regression jobs use manylinux, they require a binary that uses at most glibc 2.28,
- the `linux-base` tooling version have to be the same as `linux-release` to avoid measuring runtime differences caused by different tools.

### Tooling changes

Using the main image means we're bumping GCC 13 to 14 for the `linux-release` and `linux-base` CI jobs.

The main image is also used for running:
- all regression test jobs,
- Linux release and config test jobs.


Mytherin:
Nice! Thanks

scripts/exported_symbols_check.py
```diff
@@ -17,6 +17,7 @@
     '@GLIBC',
     '@GCC',
     '@CXXABI',
+    '__cxa_call_terminate',
```
smvv, line 20: note: gcc 14 exports this libstdc++ symbol
(This was not exported by gcc 13.)

PR #24890, opened 2026-08-19
Issue #24826: Stale MODE Ties

### LINKED ISSUE #24826: `MODE(... ORDER BY ...)` can use a stale tie-break value in a sliding ROWS frame
### What happens?

`MODE(x ORDER BY id)` must use the aggregate-local order to break frequency ties inside the current window frame. In a sliding `ROWS` frame, the optimized window path can keep using a stale tie-break value from rows that have already left the frame.

For row `id = 5` below, the frame is rows `2..5`: `B, A, B, A`. Both values occur twice, so `MODE(x ORDER BY id)` should return `B`, the earliest value in the current frame. DuckDB returns `A`.

#### Code

```sql
WITH t(id, x) AS (
  VALUES
    (1, 'A'),
    (2, 'B'),
    (3, 'A'),
    (4, 'B'),
    (5, 'A')
),
window_result AS (
  SELECT
    id,
    MODE(x ORDER BY id) OVER (
      ORDER BY id
      ROWS BETWEEN 3 PRECEDING AND CURRENT ROW
    ) AS actual
  FROM t
),
reference_result AS (
  SELECT
    o.id,
    (
      SELECT MODE(i.x ORDER BY i.id)
      FROM t AS i
      WHERE i.id BETWEEN o.id - 3 AND o.id
    ) AS expected
  FROM t AS o
)
SELECT
  w.id,
  r.expected,
  w.actual
FROM window_result w
JOIN reference_result r USING (id)
ORDER BY w.id;
```

#### Expected

| id | expected | actual |
| ---: | --- | --- |
| 1 | A | A |
| 2 | A | A |
| 3 | A | A |
| 4 | A | A |
| 5 | B | B |

#### Actual

| id | expected | actual |
| ---: | --- | --- |
| 1 | A | A |
| 2 | A | A |
| 3 | A | A |
| 4 | A | A |
| 5 | B | A |

#### Notes From Minimization

- `PRAGMA debug_window_mode='separate'` changes the output to the expected result.
- `SET enable_optimizer=false` changes the output to the expected result.
- Removing `ORDER BY id` from `MODE(x ORDER BY id)` removes the deterministic tie-break mismatch.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (`11dc00c898`)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

hawkfish:
Invalidate the cached mode when another value ties its count, and break
ties by rescanning the frames in row order instead of trusting first_row,
which ModeRm does not maintain.

fixes: https://github.com/duckdb/duckdb/issues/24826


hawkfish:
(Generated by Claude, reviewed by me.)

PR #24891, opened 2026-08-19
Internal #10417: AsOf Hive Partitioning

hawkfish:
* Generalise the group management to support partitioning Values.
* Add partitioning planning to AsOf joins;
* Break up the planning code for readability;
* Fix AsOf scheduling confusion of group_idx and bin_idx (caused hangs with higher cardinality equalities).
* Try to hack around Win64 compilation issue.
* Various tidy infractions


hawkfish:
The only CI failure is a Mingw download failure. I can keep pushing merges, but I think there is nothing important left.

src/execution/operator/join/physical_asof_join.cpp
```diff
@@ -64,51 +66,161 @@ PhysicalAsOfJoin::PhysicalAsOfJoin(PhysicalPlan &physical_plan, LogicalCompariso
...
-	size_t child = 1;
+	size_t &child;
 	//! The child's partitioning description
-	vector<SortStrategyPtr> sort_strategies;
+	array<SortStrategyPtr, 2> sort_strategies;
 	//! The child's partitioning buffer
-	vector<SortStrategySinkPtr> strategy_sinks;
+	array<value_map_t<SortStrategySinkPtr>, 2> strategy_sinks;
+	//! The number of sunk rows (for progress)
+	atomic<idx_t> count;
```
lnkuiper: Is this field used?

src/execution/operator/join/physical_asof_join.cpp
```diff
@@ -64,51 +66,161 @@ PhysicalAsOfJoin::PhysicalAsOfJoin(PhysicalPlan &physical_plan, LogicalCompariso
...
+	D_ASSERT(lhs_partitions.size() == rhs_partitions.size());
+	if (!lhs_partition_cols.empty()) {
+		partition_infos.emplace_back(std::move(lhs_partition_cols));
+		partition_infos.emplace_back(std::move(rhs_partition_cols));
+	}
+}
+
+PhysicalAsOfJoin::PhysicalAsOfJoin(PhysicalPlan &physical_plan, LogicalComparisonJoin &op, PhysicalOperator &left,
+                                   PhysicalOperator &right)
+    : PhysicalAsOfJoin(physical_plan, op, left, right, {}, {}) {
```
lnkuiper: Is this constructor used anymore?

src/execution/physical_plan/plan_asof_join.cpp
```diff
@@ -350,9 +357,40 @@ PhysicalOperator &PhysicalPlanGenerator::PlanAsOfJoin(LogicalComparisonJoin &op)
 				return *result;
 			}
 		}
-		return Make<PhysicalAsOfJoin>(op, left, right);
+
+		//	Check to see if we can leverage partitioning
+
+		//	TODO: HasSingleValuePartitions takes expressions instead of pointers.
+		//	Which is convenient for the other clients, but not for us...
```
lnkuiper, line 364: Do we have a regression test for `HasSingleValuePartitions`?

src/execution/operator/join/physical_asof_join.cpp
```diff
@@ -64,51 +66,161 @@ PhysicalAsOfJoin::PhysicalAsOfJoin(PhysicalPlan &physical_plan, LogicalCompariso
...
-		sort_strategies.emplace_back(std::move(sort));
+	AsOfGlobalSinkState(ClientContext &client, const PhysicalAsOfJoin &op);
 
-		auto &rhs = op.children[1].get();
-		sort = SortStrategy::Factory(client, op.rhs_partitions, op.rhs_orders, rhs.GetTypes(), partitions_stats,
-		                             rhs.estimated_cardinality, true);
-		strategy_sinks.emplace_back(sort->GetGlobalSinkState(client));
-		sort_strategies.emplace_back(std::move(sort));
+	bool SupportsReuse() const override {
+		return true;
```
lnkuiper: This is only used for recursive CTEs. Is that tested? Or do we not support that?

lnkuiper (commented):
Thanks for the PR! Some comments:

src/execution/operator/join/physical_asof_join.cpp
hawkfish: I don't think it ever was! Probably copy and paste from windowing. It does have a comment about progress, and I've noticed that AsOf does not implement the progress callback, but that is for another PR.

src/execution/operator/join/physical_asof_join.cpp
hawkfish: Apparently not. (Windowing is used internally in some places without these arguments being available, but AsOf is only constructed by explicit planning.)

src/execution/operator/join/physical_asof_join.cpp
hawkfish: Nope!

```cpp
		throw NotImplementedException("AsOf joins are not supported in recursive CTEs yet");
```

(I didn't realize that was what it was for - again I was working from windowing as a model. Another victory for code reviews!)

src/execution/physical_plan/plan_asof_join.cpp
hawkfish: You mean a test with hive partitioning? Yeah I should add one - I just wasn't sure how one generates something like that.  I think I'll see if I can add a benchmark too.

hawkfish: I was unable to generate fake Kx data that had different performance from the old code, but I reverified that the Kx data was an improvement. I think their data is skewed in some way.

lnkuiper:
Thanks!

PR #24894, opened 2026-08-19
Upload latest alpha version file to staging bucket

smvv:
See also:
- https://github.com/duckdb/duckdb-install-scripts/pull/3
- https://github.com/duckdblabs/duckdb-internal/issues/9572

The alpha version file can be uploaded from duckdb main CI (initial approach, this PR) or using the release event system (later).

For example, when the benchmarks succeeded, we can bump the latest alpha version so the alpha versions remain performant (by omitting alpha versions that regressed).


Mytherin:
Thanks!

PR #24895, opened 2026-08-19
Add CompressionSegmentReader for fixed-size uncompressed segments

HendrikLambert:
This PR adds `CompressionSegmentReader`, a bounds-checked reader for compression segment data. The goal is to provide a general abstraction that compression algorithms can use to check whether their reads stay within the segment bounds. This avoids implementing the same logic separately in every algorithm. `FixedSizeUncompressor` is the first to use it, other compression methods will be moved over separately.

## Changes
- Introduce`CompressionSegmentReader` which adds checked sequential and random reads, typed arrays, alignment, and sub-readers.
- Use the reader for fixed-size scans and row fetches
- Enforce `0 <= row_id < count` in `ColumnSegment::FetchRow`


Mytherin:
Thanks for the PR! Looks great - could you just have a look at the CI failures? One easy path there might be to just remove the added C++ tests (these are alright for initial testing & implementation but in my experience don't add much long-term).

PR #24896, opened 2026-08-19
Fix percentile_cont/disc(0) with descending order 

hawkfish:
Descending order reaches the quantile aggregate as a negative fraction, but 0 negates to -0 which isn't < 0, so a zero percentile with DESC lost the order and returned the minimum instead of the maximum. Detect the sign with signbit, and keep the negative zero through the binder.

(Originally submitted by @apoorva-01, just shepherding it)


Mytherin:
Thanks!

Issue #24897, opened 2026-08-19
A single large string in a VARCHAR column can cause an explosion of memory use


mdlavin:
### What happens?

While using DuckDB in production, I noticed that in some rare cases the presence of a single large (5MB) value in a VARCHAR could cause gigs of memory to be used in queries.  It took me a while to narrow it down, but it happens when a large row value is followed by many rows of NULL values.

It turns out that scanning an uncompressed `VARCHAR` column where NULLs follow an out-of-line "overflow" string (any value ≥ the string block limit) allocates roughly `STANDARD_VECTOR_SIZE (2048) × overflow_string_size` of non-spillable `OVERFLOW_STRINGS` memory instead of one copy. A single 2 MB value among NULLs needs ~4 GB to scan; the same column with `''` instead of NULL needs ~0.

### Explanation 

Overflow strings are stored with a negative dictionary offset (that's the "big string" marker). A `NULL` is stored as a verbatim copy of the previous row's offset so the delta-based length calc (abs(current) - abs(previous)) yields 0 — but this copies the sign too. A `NULL` trailing an overflow string therefore inherits its negative offset, and on scan `FetchStringFromDict` routes every negative offset to `ReadOverflowString`, which allocates and pins the full string onto the result vector. Every trailing `NULL` re-reads the same big value .

Order is decisive, because the marker only propagates forward:

| big value position | surrounding NULLs | scan peak |
| ------------------|---------------------|-----------|  
| first row          | all NULLs after   | 4 GB      | 
| middle             | before + after    | 4 GB      |
| last row           | all NULLs before  | ~0        |  

NULLs before copy offset 0 (positive) and hit the cheap dict path. Only NULLs after inherit the negative marker. Results are always correct (the value is masked NULL); this is purely a memory/perf defect, tag OVERFLOW_STRINGS.

### To Reproduce

```sql
PRAGMA force_compression='uncompressed';
CREATE TABLE t AS
  SELECT CASE WHEN i = 0 THEN repeat('x', 2*1024*1024) ELSE NULL END AS v
  FROM range(50000) tbl(i);
CHECKPOINT;

SET memory_limit='512MB';
SELECT sum(length(v)) FROM t;   -- Out of Memory Error
```

Change `ELSE NULL` to `ELSE ''` and peak memory usage is ~0. 

### OS:

any, amd64

### DuckDB Version:

1.5.x

### DuckDB Client:

Node

### Full Name:

Matt Lavin

### Affiliation:

[...]

PR #24898, opened 2026-08-19
Fix pathological scan memory for NULLs trailing an overflow string (#24897)

mdlavin:
Uncompressed VARCHAR overflow strings are stored with a negative dictionary offset, and a NULL copies the previous row's offset verbatim to make its delta-computed length 0. This copies the sign too, so a NULL following an overflow string inherits its negative offset and FetchStringFromDict routes it to ReadOverflowString, allocating a full pinned copy of the big string per trailing NULL (~STANDARD_VECTOR_SIZE copies per vector, tag OVERFLOW_STRINGS).

A NULL that copied a marker has a negative offset but a computed length of 0, whereas a real overflow string always has length BIG_STRING_MARKER_SIZE. Short- circuit the negative branch when the length is 0 and return an empty string. This is a scan-side fix at the single choke point, so it also repairs already- written databases with no storage-format change

### DIFF
--- a/src/include/duckdb/storage/string_uncompressed.hpp
+++ b/src/include/duckdb/storage/string_uncompressed.hpp
@@ -245,6 +245,15 @@ public:
 
 			auto str_ptr = char_ptr_cast(dict_pos);
 			return string_t(str_ptr, string_length);
+		} else if (string_length == 0) {
+			// NULL values are stored as a copy of the previous entry's dictionary offset (see
+			// StringAppendBase). When the previous entry is a big (overflow) string, its offset is
+			// negative, so the NULL inherits that negative offset even though it references no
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks for the fix! Makes sense to me.

mdlavin:
@Mytherin I'm happy to see that the fix was good. Can I backport it to the 1.5 branch too? I'm currently running a packaged version to avoid this problem in production. I'd love to have it in an official version.

Mytherin:
Sure, feel free, although there's currently no immediate plans to publish a new v1.5 version.

mdlavin:
Just in case 1.5.6 shows up before 1.6, here is the backport: https://github.com/duckdb/duckdb/pull/24918

PR #24901, opened 2026-08-19
Don't narrow a pushed-down filter constant to the file's physical type

### LINKED ISSUE #24900: Multi-file filter pushdown narrows the constant to the file's physical type, giving wrong results in both directions
### What happens?

When a multi-file scan presents a column at a wider logical type than the file physically stores, filter pushdown rewrites the predicate into the file's local type and narrows the constant with a rounding cast. The scan then applies a different predicate than the one in the query.

A FLOAT `0.1` read under a DOUBLE schema comes back as `0.10000000149011612`. That value is greater than `0.1` and it is not equal to `0.1`. With pushdown enabled `c > 0.1` does not match it and `c = 0.1` does.

### To Reproduce

```sql
COPY (SELECT 0.1::FLOAT AS c) TO 'a.parquet' (FORMAT parquet);

SELECT c FROM read_parquet('a.parquet',
  schema=map{'c': {'name':'c','type':'DOUBLE','default_value':NULL}});
-- 0.10000000149011612

SELECT count(*) FROM read_parquet('a.parquet',
  schema=map{'c': {'name':'c','type':'DOUBLE','default_value':NULL}}) WHERE c > 0.1;
-- expected 1, actual 0

SELECT count(*) FROM read_parquet('a.parquet',
  schema=map{'c': {'name':'c','type':'DOUBLE','default_value':NULL}}) WHERE c = 0.1;
-- expected 0, actual 1
```

`SET disabled_optimizers='filter_pushdown';` returns the correct answer to both.

This is not specific to FLOAT and DOUBLE. Any lossy numeric narrowing does it. A BIGINT file holding `2` read as DOUBLE gives 0 rows for `c > 1.5` and 1 row for `c = 1.5`, and a DECIMAL(18,2) file holding `1.25` read as DECIMAL(18,3) gives 0 rows for `c > 1.245`.

### Cause

1. `MultiFileColumnMap` sets `filter_conversion = CAST_FILTER` whenever `local_type != global_type` (`src/common/multi_file/multi_file_column_mapper.cpp:49-50`), and `CreateFilters` casts the filter to the file's physical type (`:1134-1136`).
2. For a comparison against a constant, the constant is narrowed to that type at `:1006-1012` through `TryCastConstant` (`:886-896`), whose only guard is `StatisticsPropagator::CanPropagateCast`. That returns true for every numeric-to-numeric pair (`src/optimizer/statistics/expression/propagate_cast.cpp:32-75`), and `DefaultTryCastAs` then succeeds by rounding.
3. `CanPropagateCast` answers whether statistics survive a cast in the direction the data flows, where rounding is harmless because a wider min/max is still valid. The filter rewrite goes the other way and has to be exact, because it replaces the global filter inside the reader instead of being applied on top of it.

`RewriteDynamicFilterExpression` (`:941-952`) narrows the same way, so join-generated dynamic filters are exposed too.

### Impact

Wrong results with no error, in both directions: `>` `<` `<>` lose qualifying rows, while `=` `>=` `IN` return rows that fail the predicate. Which direction you get depends on how the constant rounds in the file's type.

The affected rows are those whose local value falls in the rounding gap between the constant and its narrowed form, which is one FLOAT value. Every row holding it is affected: one million rows of FLOAT `0.1` beside a DOUBLE file return 1 instead of 1000001 for `c > 0.1`, since row group pruning then discards the whole file.

The code is in `src/common/multi_file/`, so any extension that presents an evolved schema through it is affected. I hit this through `duckdb/ducklake#1307`.

### Possible fix

`BoundCastExpression::CastIsInvertible` is the predicate for that question, and `comparison_simplification.cpp:191` and `in_clause_simplification_rule.cpp:34` already gate the equivalent rewrite on it:

```cpp
static bool TryCastConstant(Value &constant, const LogicalType &target_type) {
    if (!BoundCastExpression::CastIsInvertible(constant.type(), target_type)) {
        return false;
    }
```

### OS:

Debian 12, x86_64, kernel 6.1.0-44, glibc 2.36

### DuckDB Version:

`main` @ `ce512b864c` (`v1.6.0-dev13001`), also reproduces at `f8e1c96a53` and `ac5c6d11c3`. Introduced by `1ac3f54d72` (2025-03-27), so present since v1.3.0.

### DuckDB Client:

CLI, Release, gcc 12

### Hardware:

N/A

### Full Name:

Gavin Brand

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

GavinPizza:
Fixes #24900

A multi-file scan that presents a column at a wider type than the file physically stores narrows pushed-down filter constants to the file's type. The rounded constant changes which rows match. A FLOAT `0.1` read under a DOUBLE schema comes back as `0.10000000149011612`, but `c > 0.1` does not match it and `c = 0.1` does.

`TryCastConstant` guarded the narrowing with `StatisticsPropagator::CanPropagateCast`, which accepts every numeric pair because carrying statistics through a cast tolerates rounding. Rewriting a predicate does not. This switches the guard to `BoundCastExpression::CastIsInvertible`, which `comparison_simplification.cpp` and `in_clause_simplification_rule.cpp` already use for the same rewrite.

Declining the rewrite does not give up row group pruning. The existing fallback registers the cast in `reader.expression_map` and prunes through `ExpressionColumnReader::Stats()`, so the cost is materialising and casting the column before filtering rather than scanning more row groups.

### DIFF
--- a/src/common/multi_file/multi_file_column_mapper.cpp
+++ b/src/common/multi_file/multi_file_column_mapper.cpp
@@ -8,7 +8,6 @@
 #include "duckdb/planner/expression/bound_operator_expression.hpp"
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "duckdb/parser/expression/constant_expression.hpp"
-#include "duckdb/optimizer/statistics_propagator.hpp"
 #include "duckdb/planner/filter/list.hpp"
 #include "duckdb/function/scalar/struct_functions.hpp"
 #include "duckdb/function/scalar/struct_utils.hpp"
@@ -884,7 +883,9 @@ static unique_ptr<Expression> CreateStructExtractExpression(unique_ptr<Expressio
 }
 
 static bool TryCastConstant(Value &constant, const LogicalType &target_type) {
-	if (!StatisticsPropagator::CanPropagateCast(constant.type(), target_type)) {
+	// this filter replaces the global one in the reader, so a constant that does not survive the cast
+	// exactly would silently change which rows match
+	if (!BoundCastExpression::CastIsInvertible(constant.type(), target_type)) {
 		return false;
 	}
 	auto cast = constant.DefaultTryCastAs(target_type);
--- a/test/sql/copy/parquet/parquet_filter_narrowing_cast.test
+++ b/test/sql/copy/parquet/parquet_filter_narrowing_cast.test
@@ -1,0 +1,113 @@
+# name: test/sql/copy/parquet/parquet_filter_narrowing_cast.test
+# description: Filter pushdown into Parquet with a narrowing cast
+# group: [parquet]
+
+require parquet
+
+# the file stores FLOAT but is read as DOUBLE, so the scan yields 0.10000000149011612 - pushing the
+# filter must not round the constant down to FLOAT and make that row compare equal to 0.1
+
+statement ok
+COPY (SELECT 0.1::FLOAT AS c) TO '{TEST_DIR}/narrow_float.parquet' (FORMAT parquet)
+
+statement ok
+CREATE VIEW f AS SELECT * FROM read_parquet('{TEST_DIR}/narrow_float.parquet',
+	schema=map{'c': {'name': 'c', 'type': 'DOUBLE', 'default_value': NULL}})
+
+query I
+SELECT c FROM f
+----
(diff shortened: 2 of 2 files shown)


Mytherin:
Thanks!

PR #24907, opened 2026-08-20
Preserve filter rewrites across pushdown and statistics propagation

kryonix:
I found this while comparing the TPC benchmark queries against 1.5.5. TPC-DS Q15 had become roughly 60% slower because three optimizer decisions interacted badly.

Filter pushdown first moves a cross-table predicate into a join residual. That residual contains a large constant `IN` clause, but the IN-clause rewriter only visits filters and projections, so it can no longer turn the list into a MARK join. Statistics propagation then proves another branch of the filter impossible from the `cs_sales_price` statistics. Once that branch disappears, the remaining filter refers to fewer relations and another filter-pushdown pass can place it much earlier, but the optimizer did not revisit filter pushdown at that point.

I initially expressed the reoptimization trigger by recognizing the particular OR/comparison shape in Q15. That was too specific, and it also put lifecycle policy directly in `Optimizer`. I replaced it with a dedicated `FilterStatisticsOptimizer`. It considers residual filters that still span multiple bindings, runs statistics propagation in a filter-simplification mode, and reruns filter pushdown only when simplification changes the filter's binding ownership. Materialized CTE filter propagation follows that pass so the CTE definition sees the newly pushed filters as well.

The distinction between filter simplification and the normal statistics pass is explicit. The early pass applies WHERE-filter truth semantics, where both `FALSE` and `NULL` remove the row, but it does not reuse those rules under `NOT`, in projections, or in join predicates. Compressed materialization remains part of the full statistics pass only.

I also keep residual predicates containing a rewritable `IN` clause in a `LogicalFilter` until the IN-clause rewriter can consume them. The recursive search and rewrite now continue through a non-rewritable parent `IN`, so a dynamic outer list cannot hide an independently rewritable constant list below it.

| workload | main median | branch median | range on main | range on branch |
|---|---:|---:|---:|---:|
| TPC-DS Q15 | 0.058 s | 0.027 s | 0.057–0.060 s | 0.024–0.037 s |

### DIFF
--- a/src/optimizer/optimizer.cpp
+++ b/src/optimizer/optimizer.cpp
@@ -16,6 +16,7 @@
 #include "duckdb/optimizer/multi_stage_aggregate_rewriter.hpp"
 #include "duckdb/optimizer/empty_result_pullup.hpp"
 #include "duckdb/optimizer/expression_heuristics.hpp"
+#include "duckdb/optimizer/filter_statistics.hpp"
 #include "duckdb/optimizer/filter_pullup.hpp"
 #include "duckdb/optimizer/filter_pushdown.hpp"
 #include "duckdb/optimizer/grouping_sets_optimizer.hpp"
@@ -256,6 +257,11 @@ void Optimizer::RunBuiltInOptimizers() {
 		plan = filter_pushdown.Rewrite(std::move(plan));
 	});
 
+	if (!CTEContainsDML(*plan)) {
+		FilterStatisticsOptimizer filter_statistics(*this);
+		filter_statistics.Optimize(plan);
+	}
+
 	// derive and push filters into materialized CTEs
 	RunOptimizer(OptimizerType::CTE_FILTER_PUSHER, [&]() {
 		CTEFilterPusher cte_filter_pusher(*this);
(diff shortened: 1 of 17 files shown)


kryonix:
The CI run looks promising, but flagged two regressions that have to be fixed before we can think about merging this:

```text
FASTER
benchmark  base median                   PR median                     delta median        runs
---------  ----------------------------  ----------------------------  ------------------  ----
q18         90.5 ms [ 83.5 ms…109.9 ms]   88.3 ms [ 84.1 ms…105.0 ms]    -2.3 ms ( -2.5%)    45
q09        364.5 ms [354.7 ms…378.5 ms]  353.2 ms [343.3 ms…363.6 ms]   -11.2 ms ( -3.1%)    40
q32          4.8 ms [  3.8 ms…  5.9 ms]    4.5 ms [  3.3 ms…  7.4 ms]  -321.0 µs ( -6.6%)   110
q92         13.2 ms [ 12.6 ms… 17.9 ms]   12.2 ms [ 11.5 ms… 16.0 ms]  -908.5 µs ( -6.9%)   110
q23          1.3 s  [  1.2 s …  1.5 s ]    1.1 s  [  1.1 s …  1.3 s ]  -148.7 ms (-11.6%)    40
q62         38.4 ms [ 36.0 ms… 44.3 ms]   30.2 ms [ 28.0 ms… 37.5 ms]    -8.2 ms (-21.4%)   110
q10         57.4 ms [ 52.8 ms… 63.9 ms]   42.4 ms [ 38.5 ms… 48.1 ms]   -15.0 ms (-26.1%)    85
q99         50.8 ms [ 50.0 ms… 57.4 ms]   35.8 ms [ 35.0 ms… 40.4 ms]   -15.0 ms (-29.5%)    95
q15         52.3 ms [ 46.9 ms… 62.4 ms]   31.7 ms [ 26.6 ms… 39.6 ms]   -20.6 ms (-39.4%)   105

REGRESSIONS (≥+10%)
benchmark  base median                   PR median                     delta median        runs
---------  ----------------------------  ----------------------------  ------------------  ----
q35        123.7 ms [115.8 ms…137.3 ms]  147.7 ms [140.3 ms…161.1 ms]   +23.9 ms (+19.3%)    40
q69         45.2 ms [ 43.4 ms… 54.3 ms]   58.4 ms [ 56.0 ms… 65.0 ms]   +13.2 ms (+29.3%)    80
```

src/optimizer/optimizer.cpp
```diff
@@ -426,14 +520,29 @@ void Optimizer::RunBuiltInOptimizers() {
...
 		});
 	}
+	if (filters_simplified) {
+		RunOptimizer(OptimizerType::FILTER_PUSHDOWN, [&]() {
+			FilterPushdown filter_pushdown(*this);
+			unordered_set<TableIndex> top_bindings;
+			filter_pushdown.CheckMarkToSemi(*plan, top_bindings);
+			plan = filter_pushdown.Rewrite(std::move(plan));
+		});
+		RunOptimizer(OptimizerType::STATISTICS_PROPAGATION, [&]() {
```
Mytherin: This all seems very specific and likely too fitted on specific queries - I also think this shouldn't live in the `Optimizer` directly

Mytherin (commented):
Thanks for the PR! Great to get this fixed - maybe we can move this into separate PRs, though? Also left a comment

src/optimizer/optimizer.cpp
kryonix: Agreed on both points. The current trigger is too narrowly expressed in terms of a multi-relation OR containing a column/constant comparison. I'll rewrite this PR to contain only the Q15/statistics/IN-clause work, move that lifecycle into a dedicated optimizer component, and use the structural invariant rather than inspecting specific expression shapes. I'll move the projection-pullup and delimiter-domain fixes into separate PRs.

PR #24908, opened 2026-08-20
Reduce keyed recursive UNION update overhead

kryonix:
While profiling the changed-key implementation from #24647 on the public LDBC SF100 pathfinding workload, I found that duplicate-heavy recursive epochs still paid two avoidable costs before reaching the recurring hash table. Parallel workers published candidate chunks through a shared append lock, and redundant updates were only collapsed after all of that output had been materialized globally.

This is a follow-up to #24647. It keeps the keyed `UNION` semantics from that PR and changes only how eligible candidate streams are published and preaggregated.

In short, this PR:

- buffers keyed `UNION` output per worker and publishes it once instead of taking the shared append lock for every chunk;
- adaptively preaggregates duplicate-heavy worker output before it reaches the recurring hash table;
- combines worker-local aggregate states with any raw residual candidates against the same pre-epoch delta; and
- adds structured metrics for classification, worker-local preaggregation, residual rows, and combine work.

Each keyed `UNION` sink starts by buffering its output locally. Once it has enough rows to justify classification, it carries a HyperLogLog sketch across a bounded number of vectors. It selects worker-local preaggregation only when the sketch's conservative distinct-count upper bound shows substantial fan-in. Tiny, non-expanding, and locally unique streams remain as raw candidates, and sampling stops after a bounded budget when the stream stays unique.

Preaggregation is restricted to payload aggregates that have a combine callback and are explicitly not order-dependent. Selected workers build private keyed aggregate hash tables. At the epoch boundary, the recursive operator combines those states and applies any raw residual candidates through the existing generic changed-key path. This preserves aggregate allocator ownership, the first-touch snapshot boundary, and the finalized SQL-visible result. Order-dependent and non-combinable aggregates retain the raw path.

The buffering change also keeps publication semantics deliberately narrow. `USING KEY ... UNION ALL` continues to forward every candidate, including duplicates, and ordinary recursive CTEs retain their existing output path. The output-publication tests cover keyed `UNION`, keyed `UNION ALL`, and ordinary recursive CTEs under parallel execution.

### DIFF
--- a/src/execution/operator/set/physical_recursive_cte.cpp
+++ b/src/execution/operator/set/physical_recursive_cte.cpp
@@ -218,6 +218,10 @@ void RecursiveCTEState::CombineOutput(ColumnDataCollection &output) {
 	    collect_metrics ? std::chrono::steady_clock::now() : std::chrono::steady_clock::time_point();
 	const auto row_count = output.Count();
 	CurrentOutputTable().Combine(output);
+	// Keyed workers can resume shared appends after publishing local output.
+	if (op.using_key) {
+		InitializeSharedOutputAppend();
+	}
 	if (collect_metrics) {
 		const auto after_work = std::chrono::steady_clock::now();
 		RecordSinkMetrics(
--- a/src/execution/operator/set/physical_recursive_cte.cpp
+++ b/src/execution/operator/set/physical_recursive_cte.cpp
@@ -373,6 +373,8 @@ unique_ptr<GlobalSinkState> PhysicalRecursiveCTE::GetGlobalSinkState(ClientConte
 	return make_uniq<RecursiveCTEState>(context, *this);
 }
 
+enum class RecursiveCTELocalPreaggregationDecision : uint8_t { DEFER, PREAGGREGATE, DIRECT };
+
 class RecursiveCTELocalPreaggregationState {
 public:
 	RecursiveCTELocalPreaggregationState(ClientContext &context_p, const PhysicalRecursiveCTE &op_p)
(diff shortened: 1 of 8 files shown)


src/function/aggregate/distributive/minmax.cpp
```diff
@@ -268,6 +336,14 @@ struct VectorMinMaxBase {
 		OP::template Execute<string_t, STATE, OP>(target, source.value, input_data);
 	}
 
+	template <class STATE, class OP>
+	static bool CombineWithChange(const STATE &source, STATE &target, AggregateInputData &input_data) {
+		if (!source.is_set) {
+			return false;
+		}
+		return OP::template ExecuteWithChange<string_t, STATE, OP>(target, source.value, input_data);
```
Mytherin: This is adding a lot of templated code (that will increase binary size, in addition to increasing aggregate function API complexity). I'm not opposed to adding new APIs but it seems the use here is rather narrow - maybe we can come up with a more generic / generically useful API to add, or figure out a way of getting these performance benefits without adding all this extra generated code?

kryonix: Good point. I have added the API, because the aggregate implementation is the only layer that can cheaply and correctly determine whether an update or combine changed the aggregate's finalized SQL value. Comparing internal states is not generally sound, while finalizing and comparing values before and after every update is exactly the overhead this PR avoids.

I used separate callbacks to keep ordinary aggregation branch-free, but I agree that this duplicates templated scatter/combine implementations and may increase binary size. I will investigate expressing change reporting as an optional output of the existing update/combine path, or otherwise sharing the generated driver, and measure both binary size and ordinary aggregate performance.

While I'm adding this here, this API is not intended to be specific to this recursive CTE optimization.

Mytherin: > While I'm adding this here, this API is not intended to be specific to this recursive CTE optimization.

That's fair - but maybe we can give some extra thought on other locations where this might be useful, or if there's some way to redesign this so that this is more broadly applicable.

kryonix: To give some context, this PR is part of a larger recursive-state sequence. The next follow-up adds a `changes.T` relation inside a keyed recursive CTE. For every finalized state transition in an epoch, it exposes one row of the form `(key columns…, old STRUCT, new STRUCT)`. That relation lets recursive queries react to state transitions directly. A shortest-path relaxation can propagate only improved distances, while graph peeling or invalidation can react to keys that disappeared, without joining a candidate relation back against the complete recurring state.

The generic abstraction shared by these changes is at the grouped aggregate hash-table boundary: apply a batch and return the groups whose finalized SQL values changed. The aggregate callbacks are an optional accelerator for that operation, not the operation itself. Unsupported aggregates retain the generic snapshot/finalize/`IS DISTINCT FROM` implementation.

But I agree, blowing up the API surface and binary size is a concern, and not my intention without careful consideration.

Mytherin: Maybe we can leave that change for a separate PR? It seems orthogonal enough to the other changes.

kryonix: Makes sense! I will pull this out of this PR.

src/execution/operator/set/physical_recursive_cte.cpp
```diff
@@ -719,21 +985,103 @@ void RecursiveCTEState::CommitUsingKeyUpdatesInternal() {
...
+			const auto hash_work_ns =
+			    NumericCast<idx_t>(std::chrono::duration_cast<std::chrono::nanoseconds>(hash_end - hash_start).count());
+			D_ASSERT(snapshot_work_ns <= hash_work_ns);
+			GetEpochMetrics().RecordKeyedHashCommit(hash_work_ns - snapshot_work_ns);
+		}
+	}
+
+	AggregateHTScanState epoch_scan_state;
+	epoch_ht->InitializeScan(epoch_scan_state);
+	while (epoch_ht->ScanGroups(epoch_scan_state, distinct_rows)) {
```
lnkuiper: This code seems duplicated from some fo the code in `RecursiveCTEState::CommitPreaggregatedUsingKeyUpdatesInternal()`. Could we move that to a shared helper?

lnkuiper (commented):
Thanks! One nitpick:

src/execution/operator/set/physical_recursive_cte.cpp
kryonix: Thanks! Makes sense.

lnkuiper:
Thanks!

PR #24909, opened 2026-08-20
Preserve projection maps during column pruning

kryonix:
While auditing the projection-map handling, I found an issue in `RemoveUnusedColumns`. The generic logical-operator visitor remaps a projection map by matching the bindings produced before and after a child rewrite. If a selected binding disappears, it clears the map. That is a reasonable containment fallback for an unknown rewrite, but it is the wrong abstraction for a pass whose purpose is to remove and renumber bindings: an empty projection map means "project every surviving column", not "the selected column was pruned".

This showed up in the unused-column pass run by Top-N window elimination while optimizing a materialized common subplan. A parent projection map selected a CTE-derived binding that column pruning then removed. The visitor cleared the complete map, widening the parent layout to every remaining child column. Renumbering has a second subtle case: after removing an earlier column, an old binding can still occur numerically in the new layout while now representing a different original column, so matching bindings without their replacement provenance can silently select the wrong output.

I make projection-map maintenance part of the pruning rewrite. The pass records binding replacements when projection, aggregate, CTE, and related outputs are compacted, including the one-to-many replacements created by pushed-down nested extracts. After visiting a child, it resolves the old selection at that output boundary and rebuilds the map from the bindings that remain reachable. Replacement targets stop at the new boundary, which distinguishes simultaneous renumbering from a transitive rewrite even when their numeric bindings overlap. If every selected binding was removed, the pass retains the minimum valid child layout instead of turning the map into "all columns".

Recursive CTE pruning has an additional lifetime boundary. Its reference pruner rewrites projection maps inside the recursive term before `RemoveUnusedColumns` traverses that term again. I therefore keep those reference replacements local until the follow-up traversal has completed, and only then publish them for ancestors outside the recursive operator. Publishing them earlier would allow an already-renumbered binding to be interpreted as an old binding with the same numeric index and remapped a second time.

The replacement relation and remapping logic remain private to `RemoveUnusedColumns` and its CTE-reference helper. This does not introduce the indexed plan, ancestor traversal, or mutation protocol; the trade-off is a small amount of pass-specific bookkeeping in exchange for giving pruning explicit ownership of its output-layout semantics.

### DIFF
--- a/src/optimizer/remove_unused_columns.cpp
+++ b/src/optimizer/remove_unused_columns.cpp
@@ -31,6 +31,7 @@
 #include "duckdb/optimizer/filter_pushdown.hpp"
 #include "duckdb/optimizer/optimizer.hpp"
 #include "duckdb/planner/logical_operator_repeatability.hpp"
+#include "duckdb/planner/subquery/column_binding_layout.hpp"
 #include <utility>
 
 namespace duckdb {
@@ -162,6 +163,9 @@ void RemoveUnusedColumns::RewriteRecursiveCTEReferences(LogicalRecursiveCTE &rec
 	D_ASSERT(mode == RemoveUnusedColumnsMode::APPLY);
 	CTERefPruner cte_ref_pruner(rec.table_index, required_columns);
 	cte_ref_pruner.VisitOperator(*rec.children[1]);
+	for (auto &replacement : cte_ref_pruner.binding_replacements) {
+		root.projection_map_replacements[replacement.old_binding] = {replacement.new_binding};
+	}
 	ColumnBindingReplacer column_binding_replacer;
 	column_binding_replacer.replacement_bindings = std::move(cte_ref_pruner.binding_replacements);
 	column_binding_replacer.VisitOperator(*rec.children[1]);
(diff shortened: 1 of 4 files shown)


src/optimizer/remove_unused_columns.cpp
```diff
@@ -241,15 +248,102 @@ void RemoveUnusedColumns::ClearUnusedExpressions(vector<T> &list, TableIndex tab
 		}
 		if (should_replace) {
 			// column is used but the ColumnBinding has changed because of removed columns
-			auto created_bindings =
-			    ReplaceBinding(current_binding, ColumnBinding(table_idx, ProjectionIndex(new_col_idx)));
+			auto new_binding = ColumnBinding(table_idx, ProjectionIndex(new_col_idx));
+			auto created_bindings = ReplaceBinding(current_binding, new_binding);
+			auto &map_replacements = root.projection_map_replacements[current_binding];
```
lnkuiper, line 256: Deleted bindings are still missing from the replacement relation. If old binding (t, 1) is selected and removed while (t, 2) shifts into the new (t, 1), `ResolveProjectionBindings` treats that new occupant as the original selection.

lnkuiper (commented):
Thanks for the PR! Nice find. I have one comment (pre-existing issue) that we might be able to resolve in this PR too:

src/optimizer/remove_unused_columns.cpp
kryonix: Good point, I will look into that.

kryonix: This should be fixed now.

PR #24911, opened 2026-08-20
Use CompressionSegmentReader for validity and bitpacking

HendrikLambert:
This PR moves uncompressed validity and bitpacking reads to `CompressionSegmentReader`.

## Changes
- Extend `CompressionSegmentReader` and `array_ptr` with array slices and aligned byte reads.
- Bounds-check validity data during scans and row fetches.
- Bounds-check bitpacking metadata, group headers, offsets and compressed payloads.
- Reject invalid bitpacking modes, widths, group offsets and reads past the segment end.
- Avoid signed overflow when decoding bitpacked values.


src/storage/compression/bitpacking.cpp
```diff
@@ -553,26 +587,28 @@ static void ApplyFrameOfReference(T *dst, T frame_of_reference, idx_t size) {
 }
 
 // Based on https://github.com/lemire/FastPFor (Apache License 2.0)
-template <class T>
+template <class T, class T_U = typename MakeUnsigned<T>::type>
 static T DeltaDecode(T *data, T previous_value, const size_t size) {
```
Tishj: Have we verified that `data` has at least `size` bytes when we call this function?
And also that `size != 0`? Since we assert it, that should be the case, otherwise we should upgrade this to a `DataCorruptionException` throw instead.

src/storage/compression/bitpacking.cpp
```diff
@@ -581,90 +617,122 @@ static T DeltaDecode(T *data, T previous_value, const size_t size) {
...
 		}
 
 		// Read third value
 		if (current_group.mode == BitpackingMode::DELTA_FOR) {
-			current_delta_offset = *reinterpret_cast<T *>(current_group_ptr);
-			current_group_ptr += sizeof(T);
+			current_delta_offset = group_reader.template Read<T>();
 		}
+
+		idx_t expected_payload_size = 0;
```
Tishj, line 838: Should we save this somewhere while we're dealing with this group?
Since we use `ReadBytes` as opposed to creating a sub-reader, we are now responsible for making sure all uses of this pointer don't exceed the `expected_payload_size`.

src/storage/compression/bitpacking.cpp
```diff
@@ -844,65 +922,64 @@ void BitpackingScan(ColumnSegment &segment, ColumnScanState &state, idx_t scan_c
...
-		// overflow check
-		T result;
-		bool multiply = TryMultiplyOperator::Operation(multiplier, scan_state.current_constant, result);
-		bool add = TryAddOperator::Operation(result, scan_state.current_frame_of_reference, result);
-		D_ASSERT(multiply && add);
-#endif
-		*current_result_ptr = (multiplier * scan_state.current_constant) + scan_state.current_frame_of_reference;
+		// Use unsigned arithmetic for defined wrapping
+		idx_t multiplier = scan_state.current_group_offset;
+		*current_result_ptr = static_cast<T>((static_cast<T_U>(scan_state.current_constant) * multiplier) +
```
Tishj: I think this is one of the places that needs verification, since `current_frame_of_reference` is a value read from disk.
Perhaps it makes sense to make it a convention to put a comment above variables that are read from disk.

for example:
```c++
//! Current frame of reference (read from disk)
T current_frame_of_reference;
```
To help create a signal to readers/writers of the code that the value can't be trusted

src/storage/compression/bitpacking.cpp
```diff
@@ -844,65 +922,64 @@ void BitpackingScan(ColumnSegment &segment, ColumnScanState &state, idx_t scan_c
...
 
-	BitpackingPrimitives::UnPackBlock<T>(data_ptr_cast(scan_state.decompression_buffer),
-	                                     decompression_group_start_pointer, scan_state.current_width, skip_sign_extend);
+	idx_t offset_in_compression_group =
+	    scan_state.current_group_offset % BitpackingPrimitives::BITPACKING_ALGORITHM_GROUP_SIZE;
+	auto algorithm_group = scan_state.GetAlgorithmGroup(scan_state.current_group_offset);
 
-	*current_result_ptr = scan_state.decompression_buffer[offset_in_compression_group];
-	*current_result_ptr += scan_state.current_frame_of_reference;
+	// FOR residuals are non-negative.
```
Tishj, line 1118: Do we verify anywhere that they are actually non-negative?

src/storage/compression/bitpacking.cpp
HendrikLambert: The caller guarantees it, DeltaDecode is used in both skip and scan. It's a bit of a trace, but I don't really know how to guard this in a meaningful and nice way without a performance impact. I think I should rename`size` to `count` however, since it represents elements of `T` not bytes, which also doesn't help with making it readable. The paths for scan and skip:

### Scan
destination is the result vector, to_scan elements from result_data + result_offset + scanned:
```
if (scan_count == 0) return; // :829
while (scanned < scan_count) { // scan_count - scanned >= 1
  offset_in_compression_group = current_group_offset % 32; // 0..31
  to_scan = MinValue(scan_count - scanned, 32 - offset_in_compression_group); // 1..32
  current_result_ptr = result_data + result_offset + scanned; // scanned + to_scan <= scan_count
```

### Skip 
destination is a fixed member buffer:
```
T decompression_buffer[BITPACKING_METADATA_GROUP_SIZE]; // T[>=512]
while (skipped < skip_count) { // remaining_to_skip >= 1
  offset_in_compression_group = current_group_offset % 32; // 0..31
  skipping_this_algorithm_group = MinValue(remaining_to_skip, 32 - offset_in_compression_group); // 1..32
  decompression_ptr = decompression_buffer + offset_in_compression_group; // highest index 31
```

size != 0 because both terms of the MinValue are positive: the loop condition keeps the remaining count at 1 or more, and 32 - (x % 32) is in [1..32]

While moving the algorithms over, my main focus was to guard against indexes, counts, etc that come out of the files themselves and leave DuckDB programming errors as `D_ASSERT`, but let me know what you think?

src/storage/compression/bitpacking.cpp
HendrikLambert: I took a second look. As far as I can tell any `T` is a valid frame of reference (for `CONSTANT_DELTA` the writer stores `compression_buffer[0]`, the group's first value) and it's only ever added to the output value, not used as an index or anything.

src/storage/compression/bitpacking.cpp
HendrikLambert: No, we don't check it anywhere. The comment might be a bit unclear, but it is explaining why we skip sign extension. Residuals are stored as offsets from the group minimum, so there's no sign bit. A corrupt one would just get zero-extended and added to the frame of reference, so you'd get a wrong value rather than an unsafe read. The width is checked though, in `ValidateBitpackingWidth`.

src/storage/compression/bitpacking.cpp
```diff
@@ -581,90 +617,122 @@ static T DeltaDecode(T *data, T previous_value, const size_t size) {
...
-		if (bitpacking_metadata_ptr >= handle.GetDataMutable() + current_segment.GetBlockSize()) {
-			throw InternalException("Bitpacking offset is out of range at block \"%llu\" - corrupt database file",
-			                        segment.GetBlockHandle()->BlockId());
-		}
-
-		// load the first group
-		LoadNextGroup();
+	explicit BitpackingScanState(BufferHandle handle_p, ColumnSegment &segment)
+	    : handle(std::move(handle_p)),
+	      reader(CompressionSegmentReader::FromSegment(handle, segment, "bitpacking segment")),
```
Tishj, line 711: Maybe `reader` should be an `optional<...>` so at the end of the constructor we can recreate it to exclude the "metadata_end" we just read, and also exclude the `bitpacking_metadata_encoded_t[]` we read from it.
That would help catch corruption / compression bugs that offset into reserved space.

src/storage/compression/bitpacking.cpp
```diff
@@ -844,65 +922,64 @@ void BitpackingScan(ColumnSegment &segment, ColumnScanState &state, idx_t scan_c
...
+template <class T, class T_U = typename MakeUnsigned<T>::type>
 void BitpackingFetchRow(ColumnSegment &segment, ColumnFetchState &state, row_t row_id, Vector &result,
                         idx_t result_idx) {
-	BitpackingScanState<T> scan_state(state.context, segment);
-	scan_state.Skip(segment, NumericCast<idx_t>(row_id));
+	D_ASSERT(row_id >= 0);
+	auto row_index = NumericCast<idx_t>(row_id);
+	D_ASSERT(row_index < segment.count);
+	auto &buffer_manager = BufferManager::GetBufferManager(segment.GetDatabase());
+	auto handle = buffer_manager.Pin(segment.GetBlockHandle());
```
Tishj: This is missing the `state.context`, `buffer_manager.Pin(state.context, segment.GetBlockHandle())`.
The context is passed here for bytes-read tracking, observability purposes

src/storage/compression/bitpacking.cpp
```diff
@@ -540,205 +583,388 @@ void BitpackingFinalizeCompress(CompressionState &state_p) {
...
-		// Assert we can are in the correct metadata group
-		idx_t remaining_to_skip = skip_count - skipped;
-		D_ASSERT(current_group_offset + remaining_to_skip < BITPACKING_METADATA_GROUP_SIZE);
-
-		if (current_group.mode == BitpackingMode::CONSTANT || current_group.mode == BitpackingMode::CONSTANT_DELTA ||
-		    current_group.mode == BitpackingMode::FOR) {
-			// Skipping within a constant or constant delta is done by increasing the current_group_offset
+			idx_t target_group_index = group_index + meta_groups_to_skip;
+			bool skip_lands_exactly_on_group_boundary =
+			    (skip_count + group_offset) % BITPACKING_METADATA_GROUP_SIZE == 0;
```
Tishj, line 889: Unrelated to your refactor, but something that just crossed my mind: `BITPACKING_METADATA_GROUP_SIZE` is defined based on `STANDARD_VECTOR_SIZE`. It would require a custom build, but compressing and decompressing with different `STANDARD_VECTOR_SIZE` values could make the data unreadable 🤔

Tishj (approved):
Thanks for the PR, looks good now!
Let's try to avoid refactoring the existing code as much in the other methods, to reduce risk 👍

Mytherin:
Thanks!

PR #24912, opened 2026-08-20
Unify `Match` and `MatchParseResult`

Dtenwolde:
Related to: https://github.com/duckdblabs/duckdb-internal/issues/6380

Previously we had two paths that did essentially the same thing: `Match` and `MatchParseResult`. 
`Match` was the initial implementation and was only used for autocomplete, only needing to detect if parsing was success or fail. 
`MatchParseResult` is used for by the parser to return (optionally) a `ParseResult`. 

Since these two have the same logic for matching, it's probably cleaner to make it one unified path that is used by both. However, we do still have `MatchMode`: `BUILD_PARSE_RESULT` & `RECOGNIZE_ONLY`. This allows the autocomplete to still use the fast path and not materialize `ParseResult`s that will not be used anyways. 

Other than that, this PR removed the `Match` path. 

Next up, I want to look into making the `Matcher` heap-based, so we can handle deeply nested queries in the matcher, as well as in the transformer.


Mytherin:
Cool - thanks - could you just have a look at the merge conflict?

Dtenwolde:
> Cool - thanks - could you just have a look at the merge conflict?

Will do!

PR #22251, opened 2026-04-23
Add support for static linking C-API extensions

Maxxen:
This PR adds support for statically linking C-API extensions using the same cmake configuration setup that DuckDB uses for C++ extensions.

Instead of adding an include like we do for C++ extensions in the generated extension loader, we instead forward declare the extern C function entry-point and invoke it during loading.

To track which loader stub to generate, I've added a `DUCKDB_EXTENSION_KIND` target property to each extension build target which marks it as a C-API or C++ extension.

The `duckdb_extension.h` header has also been changed to conditionally setup the vtable/define the indirection macros depending on if `DUCKDB_BUILD_STATIC_EXTENSION` is defined or not.

Im not sure how to make a test for this. I've tested locally by just adding `duckdb_extension_load(demo_capi)` in `extension/extension_config.cmake`, with and without `DONT_LINK` and that seems to work. Maybe we should just always do this for DEBUG builds or is there some other way to detect that we are running in CI/test environment from cmake?

PR #24913, opened 2026-08-20
Restore static linking behaviour for C API

ThijsDreef:
#22251 added support for statically linking C extensions, where #24135 changed the way the C API was generated. In #24135 the change to the code generation from #22251 was missed when migrating to the new code generation. Consequence of this is that any C extension that is statically linked to DuckDB now triggers a segfault as getAPI is a nullptr (due to the missing guards). 

This PR adds the guards from #22251 back into the template added by #24135

This can be reproduced by removing the DONT_LINK option from the capi_demo in main_core.cmake. I did not see a nice way to add a test to Ci for this, let me know if you have a suggestion for this.


Maxxen:
Thanks, LGTM!

Maxxen (approved):

PR #24914, opened 2026-08-20
Keep variable-size projections below joins

kryonix:
I found two regressions in TPC-DS Q62 and Q99 while comparing the current optimizer against 1.5.5. Projection pullup had started moving computed variable-width expressions, including `substring`, above joins. That makes the join carry the original variable-width payload and evaluates the expression after join fanout, turning a relatively small string operation into millions of allocations and copies.

I keep a computed projection below a comparison or ANY join when either its result or one of its referenced input columns is variable-size. Looking only at the return type is insufficient: `length(varchar_column)` returns a fixed-width `BIGINT`, but pulling it up still carries the `VARCHAR` through the join and evaluates `length` after fanout.

This is deliberately narrower than making variable-width columns a general projection barrier. Plain column references can still be removed, and fixed-width expressions over fixed-width inputs remain eligible for pullup and join reordering.

| query | main median | branch median | range on main | range on branch |
|---|---:|---:|---:|---:|
| TPC-DS Q62 | 0.100 s | 0.076 s | 0.099–0.101 s | 0.075–0.077 s |
| TPC-DS Q99 | 0.147 s | 0.0995 s | 0.145–0.149 s | 0.098–0.101 s |


Mytherin:
Thanks for the PR!

Evaluating expressions before or after joins seems like an interesting avenue for optimization - but the decision seems complex to make:

* Expressions can either increase or decrease the size of the payload (e.g. `substring` decreases, `repeat` increases, cast from varchar to int decreases, cast from int to varchar increases)
* Joins can either increase or decrease the amount of tuples to perform the operation on (foreign key / exploding joins vs filtering joins).

I'm a bit worried here that the heuristic used here is very tailored towards a specific benchmark query versus actually considering the design space here. I'm alright with including this as a stop-gap but likely this needs a more complex solution to avoid regressions elsewhere.

kryonix:
Good catch. After digging more into this, your concern is actually very much warranted. I think this has to be cardinality-aware if we want to make it work generally.

Mytherin:
Thanks!

kryonix:
I have fixed the isses that came up in the merge queue. Should be good to go.

Mytherin:
Thanks!

PR #24915, opened 2026-08-20
Preserve selected domains during delim join rewriting

kryonix:
I found a related group of regressions in TPC-DS Q10 and Q16 and TPC-H Q20 while comparing the current optimizer against 1.5.5. They came from the delimiter-join-to-CTE rewrite losing the selected outer domain before all correlated evidence consumers had used it. Once that domain disappears, the remaining MARK, SEMI, or ANTI side scans its full input instead of the much smaller set of correlated keys.

The difficult part is deciding which consumers need the domain. A marker can be consumed by a filter, an aggregate `FILTER`, or a projection; nested delimiter joins can introduce either the same outer key or a genuinely additional correlation key; and LEFT, RIGHT, and SINGLE joins have output-bearing nullable sides that cannot be treated like MARK/SEMI/ANTI evidence sides.

I now collect marker consumers from every logical-operator expression, trace delimiter expressions through projections and aggregate group bindings, and use the same ownership boundary both when detecting selected domains and when classifying nested correlation. Evidence-only sides are skipped, while nullable or scalar-output sides are skipped only when the traced delimiter expressions are proven to come exclusively from the preserved side. Direct projected markers retain a domain only for a single equality or null-safe equality; a marker with multiple correlated equality keys remains in the ordinary MARK path.

The general and nested marker sets remain separate through recursive rewriting and are combined only for late generated-domain cleanup. If lineage cannot be proven through an operator, the rewrite is conservative and can retain an unnecessary delimiter CTE rather than discard a domain that an evidence consumer still needs.

| query | main median | branch median | change |
|---|---:|---:|---:|
| TPC-DS Q10 | 0.118 s | 0.099 s | 16% faster |
| TPC-DS Q16 | 0.0185 s | 0.0155 s | 16% faster |
| TPC-DS Q35 | 0.241 s | 0.243 s | 0.8% slower |
| TPC-DS Q69 | 0.116 s | 0.116 s | unchanged |

### DIFF
--- a/src/include/duckdb/planner/subquery/delim_join_cte_rewriter.hpp
+++ b/src/include/duckdb/planner/subquery/delim_join_cte_rewriter.hpp
@@ -27,12 +27,16 @@ private:
 	                                                bool null_rejecting_filter_above = false,
 	                                                bool preserve_evidence_side = false);
 	BindingReplacementGraph MaterializeDelimJoinAsCTE(unique_ptr<LogicalOperator> &plan, LogicalOperator &rewrite_root,
-	                                                  bool null_rejecting_filter_above, bool preserve_evidence_side);
+	                                                  bool null_rejecting_filter_above, bool preserve_evidence_side,
+	                                                  bool preserve_nested_evidence_side);
 
 private:
 	Binder &binder;
 	bool cte_deliminator_enabled;
 	vector<TableIndex> generated_dedup_cte_indexes;
+	set<TableIndex> marker_indexes;
+	set<TableIndex> preserve_marker_indexes;
+	set<TableIndex> preserve_nested_marker_indexes;
 };
 
 } // namespace duckdb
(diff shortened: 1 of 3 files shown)


lnkuiper:
Thanks for the PR! Codex found two related issues in the marker analysis:

1. `CollectComplexMarkerReferences` classifies all marker references in an operator as complex as soon as any one expression contains a complex marker. For example, a two-key projected marker uses the ordinary MARK path on its own, but putting an unrelated `EXISTS(...) OR EXISTS(...)` next to it causes an extra delim CTE to be created for the two-key marker as well. In my repro this materializes a domain estimated at 40 billion rows. Could we classify the marker references per expression instead?

2. Marker use is not traced through projection aliases. An inline `coalesce(EXISTS(two_key_correlation), false)` retains the selected domain, while projecting the `EXISTS` as `m` first and then evaluating `coalesce(m, false)` does not. These are equivalent, so I think the analysis should carry each marker's origin through projection bindings.

I think a projection-aware, per-marker consumer analysis would address both points and make the behavior less dependent on the surrounding select-list shape. I did not find any result correctness issues, and the tests pass, but the first case can introduce a substantial unnecessary plan, so I think it should be fixed before merging.

lnkuiper:
Thanks LGTM now

PR #24916, opened 2026-08-20
Use unchecked casts for proven integer ranges

kryonix:
Numeric casts currently use the checked `TryCastLoop` even when statistics have already proved that every input value is representable in the target type. This is particularly visible for a cast such as `BIGINT` to `TINYINT`: once the optimizer has a complete range of `[0, 127]`, the per-row overflow check cannot change the result, but there was no cast-owned mechanism for turning that proof into a different implementation.

My first version represented this as an unchecked alternative inside `BoundCastInfo`. That made the optimization available by type alone, however, and left it to optimizer callers to activate the alternative correctly. The invariant I actually need is stronger: the cast implementation may only change as part of a successful statistics proof, and the cast function itself must define what that proof means. Custom and extension casts can have different semantics even when they use the same logical source and target types, so a type-based utility cannot safely make that decision for them.

I moved cast statistics propagation into the cast subsystem and added an optional statistics callback to `BoundCastInfo`. The `__cast` scalar function delegates its normal statistics callback to the bound cast, which removes the cast-specific branch from `StatisticsPropagator`. Built-in casts receive the existing default propagation behavior, while a custom cast without a callback produces unknown statistics and keeps its supplied implementation. This also gives extensions an explicit place to define cast-specific statistics in the future.

Native integral casts use a specialized callback. It first restores the regular `TryCastLoop`, then asks the shared cast statistics code to derive target statistics. Only a successful proof replaces the function with the `static_cast`-based vector loop. This covers exact signed and unsigned bounds as well as all-NULL inputs; missing bounds or an overflowing endpoint leave the checked function in place. Resetting the function before each attempt also means that repeated propagation cannot retain an unchecked implementation after a later proof fails.

Compressed materialization now goes through the same callback instead of switching casts directly. Its internal representation change has a stronger requirement than an ordinary SQL cast: narrowing and restoring must compose to the identity. A statistics callback can describe the output range of a custom cast, but it cannot prove that two custom functions are inverses. I therefore use the built-in cast pair whose value-preserving semantics compressed materialization checks, and bind that same pair for preflight statistics, compression, and restoration. Context-registered custom casts remain visible to explicit SQL casts but cannot silently alter an internal materialization. If either built-in direction cannot propagate its statistics, compressed materialization falls back to the existing arithmetic compression functions.

I kept the specialization limited to C++ native integral types and excluded boolean, floating-point, decimal, `HUGEINT`, `UHUGEINT`, nested, fallback, and custom casts. `SMALLER_BINARY=unchecked_numeric_casts` compiles the specialized callbacks and unchecked templates out entirely, so size-sensitive builds retain the checked path without carrying the template instantiations.

```sql
SELECT sum(i::TINYINT) FROM safe_values;
```

The checked build had a median of 0.3257 seconds (range 0.3245–0.3281), while the specialized build had a median of 0.1325 seconds (range 0.1315–0.1364), a 2.46x median speedup. A separate 10 million-row sort benchmark still showed overlapping variability between the checked and specialized versions, so I am not claiming an end-to-end compressed-materialization speedup from that workload.


src/function/cast/numeric_casts.cpp
```diff
@@ -4,30 +4,48 @@
...
+	static DST Operation(SRC input) {
+		return static_cast<DST>(input);
+	}
+};
+
+template <class SRC, class DST>
+static BoundCastInfo GetNumericCast() {
+	auto result = BoundCastInfo(&VectorCastHelpers::TryCastLoop<SRC, DST, duckdb::NumericTryCast>);
+	if constexpr (std::is_integral<SRC>::value && std::is_integral<DST>::value) {
+		result.SetUncheckedFunction(&VectorCastHelpers::TemplatedCastLoop<SRC, DST, UncheckedNumericCast>);
```
Mytherin: Let's not add these casts with `SMALLER_BINARY` enabled as an ifdef - this is just a minor perf optimization that increases binary size so we should not be adding it for e.g. WASM builds

src/optimizer/statistics/expression/propagate_cast.cpp
```diff
@@ -269,6 +269,9 @@ unique_ptr<BaseStatistics> StatisticsPropagator::PropagateCast(BoundFunctionExpr
 	}
 	auto result_stats =
 	    TryPropagateCast(*child_stats, BoundCastExpression::Child(cast).GetReturnType(), cast.GetReturnType());
+	if (result_stats) {
+		BoundCastExpression::GetBoundCastMutable(cast).TrySetUnchecked();
```
Mytherin: Shouldn't we check that an unchecked cast is actually correct? Casts could still overflow, no?

Mytherin: Given that casts are now functions - I also wonder if we want to move stats propagation to the actual function instead (similar to how stats propagation for regular scalar functions works)

src/include/duckdb/function/cast/default_casts.hpp
```diff
@@ -137,14 +137,34 @@ struct BoundCastInfo {
...
+	}
 	void SetFunction(cast_function_t new_function) {
 		function = new_function;
+		unchecked_function = nullptr;
+		is_unchecked = false;
 	}
 
 private:
 	cast_function_t function;
+	cast_function_t unchecked_function = nullptr;
```
Mytherin: Instead of registering an unchecked function maybe having a stats propagation callback is a nicer idea - that could then overwrite the function if desired - see note above

Mytherin (commented):

src/function/cast/cast_function_set.cpp
```diff
@@ -62,6 +65,9 @@ BoundCastInfo CastFunctionSet::GetCastFunction(const LogicalType &source, const
 		input.query_location = get_input.query_location;
 		auto result = bind_function.function(input, source, target);
 		if (result.HasFunction()) {
+			if (i == 1 && !result.HasStatisticsCallback()) {
```
Mytherin: Can you expand on this `i == 1` here?

Mytherin (commented):
Thanks for the fixes! Looks great - one question, and can you also have a look at the merge conflict?

src/function/cast/cast_function_set.cpp
kryonix: `i == 1` identified the built-in fallback because it occupied the first vector entry. However, I found a better solution and removed that implicit indexing by storing the built-in and registered providers separately. This also makes the statistics-callback distinction explicit.

lnkuiper:
Another merge conflict, the merge queue is busy today!

kryonix:
It's resolved. For now 😅

PR #24917, opened 2026-08-20
Use main container in more CI jobs

smvv:
- Use main container in more jobs.
- Switch clang 20 to clang 21.
- Use `ubuntu_24_04_amd64_lint` for `tidy-check` CI job.


.github/workflows/Main.yml
```diff
@@ -1165,17 +1131,20 @@ jobs:
             static-libs-windows-mingw.zip
 
  no-string-inline:
-    name: No String Inline / Destroy Unpinned Blocks
+    name: No String Inline / No RTTI / Destroy Unpinned Blocks
```
smvv: I moved `no-rtti` into this job because adding more compilation jobs increases CI workload and github cache pressure.

.github/workflows/Main.yml
```diff
@@ -1218,54 +1183,12 @@ jobs:
         if: ${{ !fromJson(needs.prepare.outputs.skip_tests) }}
         run: make unittest_relassert
 
- no-rtti:
-    name: No RTTI
```
smvv: moved into CI job `no-string-inline`.

Mytherin:
Thanks!

PR #24918, opened 2026-08-20
Fix pathological scan memory for NULLs trailing an overflow string (#24897)

mdlavin:
This is a a backport of https://github.com/duckdb/duckdb/pull/24898. The testcase needed to avoid the use of `include_segment_info` which does not exist in 1.5.x

Uncompressed VARCHAR overflow strings are stored with a negative dictionary offset, and a NULL copies the previous row's offset verbatim to make its delta-computed length 0. This copies the sign too, so a NULL following an overflow string inherits its negative offset and FetchStringFromDict routes it to ReadOverflowString, allocating a full pinned copy of the big string per trailing NULL (~STANDARD_VECTOR_SIZE copies per vector, tag OVERFLOW_STRINGS).

A NULL that copied a marker has a negative offset but a computed length of 0, whereas a real overflow string always has length BIG_STRING_MARKER_SIZE. Short- circuit the negative branch when the length is 0 and return an empty string. This is a scan-side fix at the single choke point, so it also repairs already- written databases with no storage-format change.

### DIFF
--- a/src/include/duckdb/storage/string_uncompressed.hpp
+++ b/src/include/duckdb/storage/string_uncompressed.hpp
@@ -252,6 +252,15 @@ public:
 
 			auto str_ptr = char_ptr_cast(dict_pos);
 			return string_t(str_ptr, string_length);
+		} else if (string_length == 0) {
+			// NULL values are stored as a copy of the previous entry's dictionary offset (see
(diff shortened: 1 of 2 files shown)


taniabogatsch (approved):

Mytherin:
Thanks!

PR #24919, opened 2026-08-20
[Parser] Add `GrammarExtension` extension API

Tishj:
Added `GrammarExtension` to the `ExtensionCallbackManager`.
A `GrammarExtension` has a name and an optional description.
It has a virtual `vector<GrammarChange> GetChanges()` method to produce changes to be made to the grammar before compilation.

`GrammarExtension`s are opt-in, not enabled immediately upon LOAD of an extension.
Instead they are activated through the (local) `active_grammar_extensions` config setting which is of type `VARCHAR[]`.
The list of available grammar extensions can be viewed through `FROM duckdb_grammar_extensions()`.

The reason they are not all enabled by default is that due to the nature of PEG being order-based (expressions are greedy), some extensions might conflict with each other.

To test the new `GrammarExtension`, we've built a small c++ test adding Google's pipe syntax.

### Extension Surface
The modifications that can be made to the grammar through this are:
- Add a terminal rule (i.e directly provide a `unique_ptr<Matcher>` that referencing the rule expands to)
- Add a regular rule (provide a well-formed PEG expression for the rule + a transform function)
- Replace a regular rule (same as above) (NOTE: not sure if we should keep this)
- Adjust the transform function of an existing rule (subset of "Replace a regular rule")
- Adjust an existing rule: add choice, prepend choice, remove choice (all take "cursor" functions that are optional for all except "remove choice")

#### `ParsedGrammar::AddTerminalRuleOverride`
ParsedGrammar also receives a `AddTerminalRuleOverride` method, which can be used by `GRAMMAR` `ParserChange`s to *add* matcher overrides for rules (no modification/removal allowed).

#### `ParserOptions::compiled_grammar`
Replaced `optional_ptr<ParserCache>` with `shared_ptr<CompiledGrammar> compiled_grammar`. The `ParserCache` is a DatabaseInstance-only struct now, to hold the compiled base grammar only.
Temporary solution most likely, relates to https://github.com/duckdblabs/duckdb-internal/issues/10433

### Future Work
Probably through a prefix added to the name, i.e: `<extension_name>:<grammar_extension_name>`.
Look closely into guarding certain rules from being adjusted / adding special syntax to control grammar extensions, so we can guard that instead, to prevent "soft-lock" possibilities.


src/include/duckdb/parser/parser_change.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+namespace duckdb {
+
+class ParsedGrammar;
+class DatabaseInstance;
+
+enum class ParserChangeType : uint8_t { GRAMMAR };
+
+//! A database-wide change applied while constructing the PEG parser.
+class ParserChange {
```
Tishj: I think maybe we just change this to `GrammarExtension`
I don't think we'll have other types of "parser change" that work through the same mechanism.

For `DialectExtension` it doesn't use `ParserChange`, since its changes are exclusive

src/include/duckdb/parser/parser_options.hpp
```diff
@@ -28,6 +29,7 @@ struct ParserOptions {
 	optional_ptr<const ExtensionCallbackManager> extensions;
 	AllowParserOverride parser_override_setting = AllowParserOverride::DEFAULT_OVERRIDE;
 	optional_ptr<ParserCache> parser_cache;
+	optional_ptr<ClientContext> context;
```
Mytherin: We've generally tried to avoid putting the `ClientContext` in the `ParserOptions` - maybe we can make this `shared_ptr<CompiledGrammar>` instead?

Also let's maybe add a constructor to `ParserOptions` that takes a `ClientContext &` to avoid having to duplicate the logic to move over these settings.

src/parser/peg/compiled_grammar.cpp
```diff
@@ -45,50 +38,165 @@ static void ValidateParsedGrammarRoots(const ParsedGrammar &grammar) {
...
+	}
+	return Create(selected_extensions, parser_version);
+}
+
+shared_ptr<CompiledGrammar> ParserCache::GetMatcher(optional_ptr<ClientContext> context) {
+	if (context) {
+		auto &client_config = ClientConfig::GetConfig(*context);
+		shared_ptr<CompiledGrammar> result = client_config.cached_grammar;
+		if (!result || result->Version() != LatestParserVersion()) {
+			result = CompiledGrammar::Create(*context, client_config.active_grammar_extensions, LatestParserVersion());
```
Mytherin: Does this mean each connection now compiles its own grammar always, versus using the database-instance wide grammar? If there's no deviations from the "standard" grammar (the very common case) maybe we can re-use the database-wide grammar?

Mytherin: Or maybe we could move the cache to the database entirely with `active_grammar_extensions` as a key for those caches. We can assume `active_grammar_extensions` rarely changes / really only has a very limited set of available options I think.

Mytherin (commented):

src/main/extension_callback_manager.cpp
```diff
@@ -63,6 +68,16 @@ void ExtensionCallbackManager::Register(ParserExtension extension) {
 	callback_registry.atomic_store(new_registry);
 }
 
+void ExtensionCallbackManager::Register(shared_ptr<GrammarExtension> extension) {
+	if (!extension) {
+		throw InvalidInputException("Cannot register a null parser extension");
+	}
+	lock_guard<mutex> guard(registry_lock);
+	auto new_registry = make_shared_ptr<ExtensionCallbackRegistry>(*callback_registry);
+	new_registry->grammar_extensions.push_back(std::move(extension));
```
Mytherin: This should maybe be an `unordered_map` or check for double registration?

src/parser/peg/parsed_grammar.cpp
```diff
@@ -104,6 +110,97 @@ void ParsedGrammar::AddRule(const string &rule_definition, grammar_transform_fun
...
+	if (!find_cursor) {
+		throw InvalidInputException("RemoveChoice requires a choice cursor");
+	}
+	auto &rule = GetMutableRule(rule_name);
+	if (rule.recipe.expression.type != PEGExpression::Type::CHOICE) {
+		throw InvalidInputException("Grammar rule '%s' does not contain a choice", rule.name);
+	}
+	auto &children = rule.recipe.expression.children;
+	auto cursor = FindChoice(rule, find_cursor);
+	children.erase_at(cursor);
```
Mytherin, line 202: I'm guessing it shouldn't be possible for `children.size()` to be equal to 1 prior to the removal - but maybe add an internal exception?

src/include/duckdb/parser/parser_options.hpp
Tishj: I have to be honest, I don't really understand what ParserOptions is supposed to represent, in terms of scope/lifetime. It feels like it's a subset of `ClientContext` already?

Mytherin: The original idea was that we could use the parser without a `ClientContext` - maybe that's not necessary anymore and we should just pass in a `ClientContext` at this point. I would say if we go for that then we should remove `ParserOptions` entirely and just have a `ClientContext &` passed in to the `Parser` like with other constructs (e.g. the Binder).

carlopi: If at all possible being to use at least a subset of the parser functionality without `ClientContext` it would be great, in particular for Wasm. I can refactor later into classes later obviously, you see what makes the most sense.

Tishj: Does it really make sense to keep this complexity if the alternative is just always creating an in-memory connection in wasm to fulfill this requirement? That sounds like far less complexity than keeping these context-less paths around.

src/parser/peg/compiled_grammar.cpp
Tishj: I've been struggling with figuring out what we want to do with this cache to be honest.

For the common case we want to keep it cached, but `active_grammar_extensions` is a local setting, so I don't think it makes sense to cache these at all.

And for `DialectExtension` it's going to cache the compiled grammar, but inside the `DialectExtension` itself.

I think we can get rid of the cache altogether and just cache the base grammar at the `DBConfig` level (assuming we don't want a singleton)

Tishj: Implemented this, `ParserCache` only holds the base compiled grammar. `ClientConfig::cached_grammar` gets populated whenever the `active_grammar_extensions` are set, and set back to `null` when its RESET or set to NULL/[].

Mytherin:
Thanks! Looks good

Mytherin:
Could you have a look at the merge conflicts?

Tishj:
Sick, another day of getting stuck on merge conflicts and timed-out CI runners

src/include/duckdb/parser/parser_options.hpp
ted-wq-x: There is a related requirement for extensions that use an external parser: whether a particular extension participates in parsing may depend on the current session, even when parsing itself is context-free.
Consider two parser extensions loaded into the same database. A session explicitly enables one of them. If a shared parser configuration causes both callbacks to run, the inactive extension may claim a statement because their syntaxes overlap. Rejecting it later during binding or rewriting is too late—the intended parser has already been bypassed.
Query text alone cannot resolve that ambiguity. Connection initialization is also insufficient if another extension subsequently changes the shared configuration. What is needed is an independent activation decision for each extension, evaluated before it can claim a statement.
Would you consider a context-aware activation hook at the parser dispatch boundary? It could receive ClientContext to read session configuration and produce an activation decision, while the actual parsing callback remains independent of the context. Catalog access, binding, and execution would stay outside parsing.
A session-derived snapshot of active parser extensions in ParserOptions would also satisfy this requirement and would be consistent with the approach taken for compiled grammar.

PR #18274, opened 2025-07-16
Use set for row ID scanning during index scans

taniabogatsch:
- removes duplicate elimination
- removes explicit sorting step
- adds extra copy of row IDs

I ran the benchmarks and I don't see any performance degradation. 

Close https://github.com/duckdblabs/duckdb-internal/issues/4924.

PR #24920, opened 2026-08-20
Optimize art rowid container

Liuzw1998:
Hi DuckDB team,

I am trying to optimize the ART index scan and would like to see your opinions.

An ART index scan has two main phases:

1. Probe the ART and collect sorted Row IDs. ART index scans currently collect matching Row IDs in a `std::set<row_t>`. This PR replaces that query-only path with an `unsafe_vector`-backed collector that defers ordering and deduplication until needed. This outperforms a `std::set<row_t>`. I noticed that [PR #18274](https://github.com/duckdb/duckdb/pull/18274) changed this path from `std::vector` to `std::set`. Was there a particular reason for this change? Note that #24942 may confict with this PR.

2. Fetch the projected table columns for those Row IDs. I have a separate optimization for this phase and plan to open another PR later.

### Design

`Iterator::Scan` submits each Row ID through `TryAdd()`. When the raw vector reaches capacity, the collector normalizes it before deciding whether the next candidate must trigger the existing table-scan fallback. This preserves the existing set-based distinct-count capacity semantics across multiple probes and ART storage overlays.

While the vector remains sorted, duplicates equal to its tail are skipped directly. The collector uses binary search for
duplicates of earlier Row IDs only within the final 16 capacity slots. This prevents repeated normalization in the
measured near-capacity case without adding membership state. For a specially constructed workload of 20K candidates starting with `capacity - 16` unique Row IDs followed by older duplicates, this collector takes 0.70 ms versus 0.47 ms for `std::set`.

During normalization, the collector selects one of the following paths:

- non-decreasing input: `unique`;
- reverse input: `reverse` plus `unique`;
- dense unordered input: bitmap materialization;
- other unordered input: DuckDB's branchless pdqsort plus `unique`.

The bitmap requires at least 8,192 Row IDs, a non-negative and safely representable span, and bitmap storage no larger than one quarter of the input Row-ID array's size. The minimum and maximum Row IDs and reverse ordering are checked only during normalization.

### Benchmarks

The selected official DuckDB ART benchmarks, run with 32 threads, show no measurable regressions, while the broader scans improve:

| Benchmark | Baseline | This PR | Change |
|---|---:|---:|---:|
| `point_query_with_art` | 0.72 | 0.73 | no measurable regression |
| `point_query_with_art_and_many_keys` | 41.93 | 21.66 | 48.35% faster |
| `wide_range_query_with_art` | 956.44 | 517.30 | 45.91% faster |
| `delete_colocated_fetch` | 1886.85 | 1875.71 | no regression |

A custom 100K Row-ID collector benchmark shows improvements across all five tested distributions:

| Distribution | `std::set` | This PR | Change |
|---|---:|---:|---:|
| ascending | 16.91 | 1.03 | 93.91% faster |
| descending | 17.20 | 1.10 | 93.61% faster |
| dense unordered | 22.02 | 2.96 | 86.57% faster |
| sparse unordered | 22.96 | 4.02 | 82.47% faster |
| duplicates | 2.84 | 1.24 | 56.33% faster |

### Further Questions

ART lookup also remains serial and materializes all Row IDs before table fetching begins. Is there an existing plan to
parallelize ART lookup or stream Row-ID batches into table fetching?

### DIFF
--- a/src/execution/index/art/iterator.cpp
+++ b/src/execution/index/art/iterator.cpp
@@ -1,6 +1,8 @@
 #include "duckdb/execution/index/art/iterator.hpp"
 
+#include "duckdb/common/algorithm.hpp"
 #include "duckdb/common/limits.hpp"
+#include "duckdb/common/vector_size.hpp"
 #include "duckdb/execution/index/art/art.hpp"
 #include "duckdb/execution/index/art/const_prefix_handle.hpp"
 #include "duckdb/execution/index/art/node.hpp"
@@ -8,6 +10,36 @@
 #include "duckdb/execution/index/art/prefix.hpp"
 
 namespace duckdb {
+
+RowIdVectorOutput::RowIdVectorOutput(const idx_t capacity_p) : capacity(capacity_p) {
+	row_ids.reserve(MinValue<idx_t>(capacity, STANDARD_VECTOR_SIZE));
+}
+
+bool RowIdVectorOutput::TryAdd(const row_t row_id) {
+	D_ASSERT(row_ids.size() <= capacity);
+	if (row_ids.size() >= capacity) {
+		Normalize();
+		if (row_ids.size() >= capacity) {
+			return false;
+		}
+	}
+	row_ids.push_back(row_id);
+	return true;
+}
+
+void RowIdVectorOutput::Reset() {
+	row_ids.clear();
+}
+
+void RowIdVectorOutput::Normalize() {
+	std::sort(row_ids.begin(), row_ids.end());
+	row_ids.erase(std::unique(row_ids.begin(), row_ids.end()), row_ids.end());
+}
+
+unsafe_vector<row_t> RowIdVectorOutput::TakeRows() {
+	Normalize();
+	return std::move(row_ids);
+}
 
 //===--------------------------------------------------------------------===//
 // IteratorKey
@@ -64,10 +96,9 @@ ARTScanResult Iterator::Scan(const ARTKey &upper_bound, Output &output, bool equ
 
 		switch (last_leaf.GetType()) {
 		case NType::LEAF_INLINED: {
-			if (output.IsFull()) {
+			if (!output.TryAdd(last_leaf.GetRowId())) {
 				return ARTScanResult::PAUSED;
 			}
-			output.Add(last_leaf.GetRowId());
 			break;
 		}
 		case NType::LEAF: {
@@ -80,11 +111,10 @@ ARTScanResult Iterator::Scan(const ARTKey &upper_bound, Output &output, bool equ
 			}
 			// Try to output the next entry in the deprecated leaf chain.
 			while (resume_state.cached_row_ids_it != resume_state.cached_row_ids.end()) {
-				if (output.IsFull()) {
+				if (!output.TryAdd(*resume_state.cached_row_ids_it)) {
 					// If we pause here, then scanning will resume at cached_row_ids_it.
 					return ARTScanResult::PAUSED;
 				}
-				output.Add(*resume_state.cached_row_ids_it);
 				++resume_state.cached_row_ids_it;
 			}
 			resume_state.has_cached_row_ids = false;
@@ -101,13 +131,12 @@ ARTScanResult Iterator::Scan(const ARTKey &upper_bound, Output &output, bool equ
 			}
 			// Try to output the next inlined leaf.
 			while (last_leaf.GetNextByte(art, resume_state.nested_byte)) {
-				if (output.IsFull()) {
+				row_id[ROW_ID_SIZE - 1] = resume_state.nested_byte;
+				ARTKey rid_key(&row_id[0], ROW_ID_SIZE);
+				if (!output.TryAdd(rid_key.GetRowId())) {
 					// If we pause here, then scanning will resume at nested_byte in the current leaf.
 					return ARTScanResult::PAUSED;
 				}
-				row_id[ROW_ID_SIZE - 1] = resume_state.nested_byte;
-				ARTKey rid_key(&row_id[0], ROW_ID_SIZE);
-				output.Add(rid_key.GetRowId());
 
 				if (resume_state.nested_byte == NumericLimits<uint8_t>::Maximum()) {
 					break;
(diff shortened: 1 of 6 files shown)


artjomPlaunov (changes requested):
Hi @Liuzw1998, thanks for the PR! :-)

I believe we originally moved to using a set because of a previous bug involving the ConflictManager. However, I think that bug has since been fixed, so we may indeed be able to reintroduce unsafe_vector here. I will double-check this during a more thorough review pass, but I have some general comments first:

1. I suspect that most of the end-to-end query improvement comes from replacing the set with a vector, normalizing it when required, and transferring ownership of the resulting vector directly to the scan state.

   A quick comparison with my clanker also suggests that this simpler implementation captures nearly all the improvement in representative SQL workloads. Therefore, I think we should limit this PR to that and remove the 16-entry duplicate-search window, bitmap, reverse-order fast path, and other special-case normalization optimizations.

   Normalization is generally only one part of end-to-end query execution. If we can retain most of the improvement without these additional paths, I would prefer to go ahead with the simpler `unsafe_vector` change first. Subsequent optimizations can be proposed separately if there are demonstrable SQL workloads that benefit from them.

2. Please remove the C++ tests. It is fine (and useful) to use them for local testing, but we generally do not add C++ tests unless there is a special reason to do so. The modified paths should instead be exercised through SQL tests, with SQL benchmarks used for performance evaluation.

3. Please rebase this PR onto `main`. The branch this PR originally targeted has since become the 2.0 branch, and only bug fixes are now being accepted there. Sorry for the trouble.

4. Also for your follow up question, W.r.t. anything in terms of parallelizing the ART lookup, I think bigger changes like that to the core of the ART will have to wait until I finish implementing buffer managed ART indexes, since that changes a lot for node ownership and lifetime model on which parallel traversal would depend. Estimated timeline for this is roughly 2 months since I am working on it about once a week, and need to carefully vet every line since the ART is already in a fairly stable state, and I don't want to introduce any regressions in the core implementation. There are still broader opportunities outside the core ART to improve e.g. execution of ART scans as you're doing in this PR, expanding when ART scans can be used in the first place is also quite limited right now.

Liuzw1998:
@artjomPlaunov thanks for your review :D 
yes, for official DuckDB ART benchmarks, the other techniques do not help much. I changed this PR and kept only the basic unsafe_vector implementation. Please have a look again.

| Workload | Baseline | Simplified vector | Improvement |
|---|---:|---:|---:|
| many-keys | 42.644 ms | 22.864 ms | 46.38% |
| wide-range | 874.152 ms | 440.166 ms | 49.65% |

The 16-entry duplicate-search window was to prevent duplicate rowids from frequently forcing Normalizations. But this will not happen frequently (only during checkpoint), right?

src/include/duckdb/execution/index/art/iterator.hpp
```diff
@@ -70,16 +70,40 @@ struct RowIdSetOutput {
...
 	}
 	void SetKey(const IteratorKey &, const idx_t) {
 		// No-op: we don't need keys for row ID output.
 	}
-	void Add(const row_t rid) {
-		row_ids.insert(rid);
+};
+
+//! Collects query-scan Row IDs contiguously and normalizes them outside the ART locks when possible.
+class ARTRowIdCollection {
```
artjomPlaunov: Maybe we can rename this to RowIdVectorOutput, to be more consistent with the naming scheme of other policies? i.e. we have RowIdSetOutput, KeyRowIdOutput, and now RowIdVectorOutput

src/execution/index/art/iterator.cpp
```diff
@@ -9,6 +11,36 @@
 
 namespace duckdb {
 
+ARTRowIdCollection::ARTRowIdCollection(const idx_t capacity_p) : capacity(capacity_p) {
```
artjomPlaunov: Can we rename ARTRowIdCollection to RowIdVectorOutput, to be consistent with the other output policy names (e.g. RowIdSetOutput, KeyRowIdOutput)

artjomPlaunov (changes requested):
Thanks! Just one more change requested.

src/execution/index/art/iterator.cpp
Liuzw1998: Sure, I have renamed it :)

test/sql/index/art/scan/test_art_rowid_collection_deduplication.test
```diff
@@ -0,0 +1,108 @@
...
+
+onlyif threadid=1
+statement ok
+DELETE FROM checkpoint_duplicates WHERE key < 3
+
+onlyif threadid=1
+statement ok
+SELECT write_log('checkpoint delete committed', scope := 'database', log_type := 'RowIdCollectionTest')
+
+onlyif threadid=2
```
artjomPlaunov: Actually, can we get rid of this test; in my experience when I added tests that depended on ms sleeps to enforce a certain ordering, it might not work as expected on a different architecture then cause the test to be flaky on CI. I think it's fine to just get rid of this test.

src/include/duckdb/execution/index/art/iterator.hpp
```diff
@@ -70,16 +70,40 @@ struct RowIdSetOutput {
...
+		return true;
 	}
 	void SetKey(const IteratorKey &, const idx_t) {
 		// No-op: we don't need keys for row ID output.
 	}
-	void Add(const row_t rid) {
-		row_ids.insert(rid);
+};
+
+//! Collects query-scan Row IDs contiguously and normalizes them outside the ART locks when possible.
```
artjomPlaunov: Can you clarify this, or maybe drop the comment about locking; I think this class doesn't control locking context in itself, maybe we can just be more direct about what it does e.g. storing in a vector and sorting + deduplicating it.

artjomPlaunov (changes requested):
Sorry just two more nits I didn't catch before, then it should be good to go!

src/include/duckdb/execution/index/art/iterator.hpp
Liuzw1998: This comment was meaningless and I forgot to remove it... Fixed it.

test/sql/index/art/scan/test_art_rowid_collection_deduplication.test
Liuzw1998: Sure, I have removed it. Nice to learn about the experience :)

artjomPlaunov (approved):
Thanks, LGTM! :-)

artjomPlaunov:
Note to maintainers: we have different scan output policies now, so the original issue with ConflictManager + requiring set output is unaffected, as we still have set output there.

PR #24925, opened 2026-08-21
feature: add streaming select to ZSTD and DICTIONARY

fanvanzh:
ZSTD and DICTIONARY did not implement the optional select callback of the compression function framework, so a filtered point lookup fell back to ColumnData::Select, which scans the entire vector and then slices out the matching rows. With roughly 1 MB per row, returning a single row decompressed and materialized about 2 GB, an amplification of about 2000x that could push an instance into OOM under concurrency.

Changes made:
- zstd.cpp: add ZSTDStorage::Select and ZSTDScanState::SelectPartial, which decompress only the selected rows and advance the zstd frame over the gaps through the existing fixed-size skip_buffer. Peak memory drops from the size of the whole vector to the size of the selected rows plus about 128 KB. CPU is unchanged because a zstd frame can only be read forwards, so skipped bytes must still be decompressed.
- dictionary/decompression.cpp: add CompressedStringScanState::Select, unpacking only the bitpacking groups that contain a selected row instead of every group the vector spans. The dictionary format is not streaming and hands out zero-copy pointers into the pinned block, so this saves CPU rather than memory. Unpacking goes to a local buffer so that a dictionary vector emitted by an earlier scan keeps owning sel_vec.
- dictionary_compression.cpp: register the new select callback.

Testing performed:
- 2048 rows of about 180 KB in a single ZSTD vector: a one-row point lookup fails with "failed to allocate data of size 512.0 MiB" on the unpatched build and succeeds under memory_limit=100MB with the patch.
- Results compared against the non-pushdown plan for single-row, vector-boundary, sparse, dense and range selections, with and without NULLs, for both compression methods.
- test/sql/storage/compression: 158 test cases, 423170 assertions.

### DIFF
--- a/src/include/duckdb/storage/compression/dictionary/decompression.hpp
+++ b/src/include/duckdb/storage/compression/dictionary/decompression.hpp
@@ -22,6 +22,7 @@ public:
 	void ScanToFlatVector(Vector &result, idx_t result_offset, idx_t start, idx_t scan_count);
 	void ScanToDictionaryVector(ColumnSegment &segment, Vector &result, idx_t result_offset, idx_t start,
 	                            idx_t scan_count);
+	void Select(Vector &result, idx_t start, const SelectionVector &sel, idx_t sel_count);
 
 private:
 	string_t FetchStringFromDict(uint32_t dict_offset, uint16_t string_len);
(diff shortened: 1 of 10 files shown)


fanvanzh (commented):
support ZSTD/DICT streaming decompress,  reduce memory usage.

Tishj:
Thanks, could you add some tests for these?

fanvanzh:
Added three tests. 
zstd_select_memory.test is the regression guard: 2048 rows of 128 KB in one ZSTD vector fail under memory_limit='100MB' without the patch, pass with it. 
zstd_select.test and dictionary_select.test check correctness across single-row, boundary, sparse, dense and full selections, with NULLs.

fanvanzh:
@Mytherin @Tishj  Do you have any other questions about this PR? 

It can significantly reduce memory usage for large fields such as BLOB/VARCHAR/JSON.

PR #24926, opened 2026-08-21
Fall back to naive window aggregation when dynamic Top-N arguments make partial aggregate states unsafe to combine

### LINKED ISSUE #24449: Top-N MIN with row-dependent n fails in a singleton window frame

Alignyx:
Fix #24449 

### Problem

The bug affected DuckDB’s Top-N aggregate variants:`min(value, n)`, `max(value, n)`, `arg_min(..., n)`, and `arg_max(..., n)`, when they were used as window functions with a row-dependent `n`.

For example, this query has a window frame containing exactly one row:

```
  SELECT min(i, CASE WHEN i < 16 THEN 1 ELSE 2 END) OVER (
  )
  FROM range(17) t(i);
```

Every result should therefore contain only the current value: `[0], [1], …, [16]`. Instead, DuckDB raised `Mismatched n values in min/max/arg_min/arg_max`

The problem was caused by optimized window execution. DuckDB’s segment tree eagerly constructed partial aggregate states over groups of rows, even though those rows would never appear together in any requested singleton frame. A Top-N state fixes its heap capacity from the first `n` value it sees. When partial states created with different capacities were later combined, the aggregate correctly detected the mismatch and threw an exception.

The segment tree has a fanout of `16`, which explains why the original `17`-row example triggered the bug: one partial state was initialized with `n = 1`, while the state beginning at row `17` was initialized with `n = 2`.

The same underlying assumption also existed in other optimized window paths. In particular:

  - A `DISTINCT` version of the query failed in the distinct-window accelerator.
  - A full-partition dynamic-N window could fail when parallel constant-window states were combined.

### Fix

The implemented fix introduces a small capability property on the bound aggregate function indicating whether arbitrary partial states are safe to combine during window aggregation. It defaults to true, so existing aggregates are unaffected.

The Top-N binders set this property to false when `n` is row-dependent, non-foldable, or parameterized. The window executor checks the property before selecting any optimized implementation. When combining states is unsafe, it uses the existing naive window executor, which evaluates the actual window frames without constructing invalid cross-frame partial states.

### DIFF
--- a/extension/core_functions/aggregate/distributive/arg_min_max.cpp
+++ b/extension/core_functions/aggregate/distributive/arg_min_max.cpp
@@ -891,6 +891,7 @@ unique_ptr<FunctionData> ArgMinMaxNBind(BindAggregateFunctionInput &input) {
 	const auto val_type = arguments[0]->GetReturnType().InternalType();
 	const auto arg_type = arguments[1]->GetReturnType().InternalType();
 	function.SetReturnType(LogicalType::LIST(arguments[0]->GetReturnType()));
+	function.SetWindowStateCombineSafe(!arguments[2]->HasParameter() && arguments[2]->IsFoldable());
 
 	// Specialize the function based on the input types
 	auto function_data = make_uniq<ArgMinMaxFunctionData>(NULL_HANDLING, NULLS_LAST);
--- a/src/function/aggregate/distributive/minmax.cpp
+++ b/src/function/aggregate/distributive/minmax.cpp
@@ -565,6 +565,7 @@ unique_ptr<FunctionData> MinMaxNBind(BindAggregateFunctionInput &input) {
 	}
 
 	const auto val_type = arguments[0]->GetReturnType().InternalType();
+	function.SetWindowStateCombineSafe(!arguments[1]->HasParameter() && arguments[1]->IsFoldable());
 
 	// Specialize the function based on the input types
 	SpecializeMinMaxNFunction<COMPARATOR>(val_type, function);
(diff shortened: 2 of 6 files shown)


hawkfish:
Hi @Alignyx - Thanks for the PR!

Unfortunately this is not the right approach. There are a number of other built in aggregates with "partition-constant" arguments, not to mention possible user-defined aggregates.

I will be pushing a fix shortly that solves the problem by skipping the combine operation for nodes that cross partition boundaries, as they will never be used anyway. This makes windowed aggregation match the existing behaviour of regular aggregation without any additional APIs and machinery.

PR #24928, opened 2026-08-21
Allow some archs like musl to fail (for alpha releases)

smvv:
If a musl build [fails](https://github.com/duckdb/duckdb/actions/runs/32432696462/job/96629753610) for an extension group, we're not running the downstream checks (benchmarks, clients, etc.) because the dispatch call is checking if all previous CI jobs succeeded.

The extension workflow is one "job" in that sense so any extension job that fails gives:
```
### Check CI status
Run NEEDS_JSON='{
One or more required jobs failed or were cancelled.
extensions: failure
Error: Process completed with exit code 1.
```
and the release dispatch call is skipped (because GitHub actions cannot distinguish if the whole extensions workflow failed or a subset).

`allow_failure_archs` [sets](https://github.com/duckdb/extension-ci-tools/pull/399) `continue-on-error: true` for specific platforms (like musl) so we can continue the CI run and downstream checks for alpha release builds.

Note: we omit `allow_failure_archs` for official release builds, so we avoid releasing a duckdb version without (partial) musl extensions.

### Follow-up tasks

We should [track](https://github.com/duckdblabs/duckdb-internal/issues/10485) CI failures from partially failed CI runs to create github issues for these optional musl job failures.

### DIFF
--- a/.github/workflows/Extensions.yml
+++ b/.github/workflows/Extensions.yml
@@ -27,6 +27,10 @@ on:
         type: string
         required: false
         default: ''
+      allow_failure_archs:
+        type: string
+        required: false
+        default: ''
       runners:
         type: string
         required: true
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #24930, opened 2026-08-21
re-enable art prefix test

artjomPlaunov:
I think this flaky test was fixed, possibly by https://github.com/duckdb/duckdb/pull/23103.

It seemed to be triggering a use after free crash in the row group append path. I could reliably reproduce it on commits around the time it was reported as failing on CI, and not on recent main. 

However, if it fails again please just mark it as skipped and let me know so I can take another look at it.


Mytherin:
Thanks!

PR #24933, opened 2026-08-21
Fix integer underflow in NTILE with argument ordering on future-only frames

### LINKED ISSUE #24831: `NTILE` with argument `ORDER BY` fails on a future-only `ROWS` frame
### What happens?

The legal `NTILE(2 ORDER BY y)` window query fails with an internal error when its frame is `ROWS BETWEEN 1 FOLLOWING AND 1 FOLLOWING`. The query should complete and return valid bucket values; instead, DuckDB aborts while converting an unsigned underflow value to a signed integer.

### Code

```sql
SELECT
  id,
  NTILE(2 ORDER BY y) OVER (
    ORDER BY y
    ROWS BETWEEN 1 FOLLOWING AND 1 FOLLOWING
  ) AS bucket
FROM (VALUES (1, 1), (2, 2)) AS t(id, y)
ORDER BY id;
```

### Expected

| result |
| --- |
| The query succeeds and returns two rows. Every returned `bucket` is a valid value in `[1, 2]`. |

### Actual

| result |
| --- |
| Query execution fails before returning rows. |

The DuckDB error is:

```text
INTERNAL Error: Information loss on integer cast: value 18446744073709551615 outside of target range [-9223372036854775808, 9223372036854775807]
```

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (`11dc00c898`)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

AdityaM06:
Fixes #24831

### The Problem
When evaluating `NTILE` with an argument `ORDER BY` clause over a future-only `ROWS` frame (for example: `ROWS BETWEEN 1 FOLLOWING AND 1 FOLLOWING`), the current `row_idx < begin`. In `WindowNtileExecutor::GetData`, computing `partition_idx = row_idx - begin` resulted in unsigned integer underflow on `idx_t`, which then threw an internal error during `NumericCast<int64_t>(partition_idx)`.

### The Solution
Clamp the relative row positions to the frame bounds (`[begin, end - 1]`) in `WindowNtileExecutor::GetData` and `WindowRowNumberExecutor::GetData` to prevent underflow when rows precede or follow the frame.

### Adding tests
Added regression tests to `test/sql/window/test_rownumber_orderby.test` covering future-only and past-only `ROWS` frames with argument ordering.

### DIFF
--- a/src/function/window/window_rownumber_function.cpp
+++ b/src/function/window/window_rownumber_function.cpp
@@ -288,12 +288,12 @@ void WindowNtileExecutor::GetData(ExecutionContext &context, DataChunk &eval_chu
 			}
 			int64_t n_size = (n_total / n_param);
 			// find the row idx within the group
-			D_ASSERT(row_idx >= begin);
 			idx_t partition_idx = 0;
 			if (grstate.token_tree) {
-				partition_idx = grstate.token_tree->Rank(begin, end, row_idx) - 1;
+				partition_idx = MinValue(grstate.token_tree->Rank(begin, end, row_idx) - 1, idx_t(n_total - 1));
 			} else {
-				partition_idx = row_idx - begin;
+				const auto frame_row_idx = MinValue(MaxValue(begin, row_idx), idx_t(end - 1));
+				partition_idx = frame_row_idx - begin;
 			}
 			auto adjusted_row_idx = NumericCast<int64_t>(partition_idx);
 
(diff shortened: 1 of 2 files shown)


hawkfish (approved):
Thanks for the PR - LGTM!

Mytherin:
Thanks - could you just resolve the merge conflicts?

AdityaM06:
@Mytherin Rebased onto main and resolved the merge conflicts. Thanks!

PR #24934, opened 2026-08-21
Fix Lambda invoke expression property propagation

d-justen:
`invoke` stores its bound lambda body in `LambdaFunctionData`, outside the function expression's ordinary children. As a result, volatility, consistency, and fallibility inside the lambda were invisible to several planner and execution paths.

This could cause volatile lambdas to be common-subexpression eliminated. With all-constant arguments, the scalar executor could also evaluate a volatile lambda only once per vector instead of once per row.

This PR propagates `IsVolatile`, `IsConsistent`, and `CanThrow` from Lambda bind data through `BoundFunctionExpression`.

### DIFF
--- a/src/function/scalar/generic/invoke.cpp
+++ b/src/function/scalar/generic/invoke.cpp
@@ -9,6 +9,15 @@ namespace duckdb {
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #24937, opened 2026-08-21
Evaluate multi-arm simple-CASE as invoke expressions

### LINKED ISSUE #22576: Simple CASE rewrites subject into searched form, losing the distinction and causing wrong results with non-deterministic subjects

d-justen:
supersedes https://github.com/duckdb/duckdb/pull/23024
depends on https://github.com/duckdb/duckdb/pull/24934 (I actually wanted to try creating a stacked PR for this, but seems like I've failed 🤷)

fixes https://github.com/duckdb/duckdb/issues/22576
fixes https://github.com/duckdblabs/duckdb-internal/issues/9235

This is my attempt to fix the semantically wrong simple-CASE lowering to searched-CASE expression. Similar to https://github.com/duckdb/duckdb/pull/23024, this PR transforms (some) simple-CASE expressions to `invoke`, while trying to preserve BWC for previous storage versions.

The aim here is to preserve the current behavior for as many cases as possible as the evaluation via `invoke` comes with a performance penalty. Essentially, we only use `invoke` if the simple-CASE statement has multiple arms and a volatile operand. Otherwise, reformulating this as a searched-CASE statement is safe. For older storage versions, we use lossy serialization, i.e., represent all simple-CASE statemtents as searched-CASE, which may be wrong but is our only option due to the absence of `invoke`. 

Importantly, simple-CASE statements with multiple WHEN expressions, a volatile operand, and subquery or `UNNEST` containing branches now return an error as `invoke` currently does not support either of those. Whenever those are supported, we may use them to evaluate the CASE statements as well. I would argue that this is the only viable option at the moment, since the alternative would be to knowingly produce wrong results.


Maxxen:
Out of curiosity, do you have some sense of how big the performance penalty of invoke is? Id like to eventually do more re-writes with it in the future.

d-justen:
> Out of curiosity, do you have some sense of how big the performance penalty of invoke is? Id like to eventually do more re-writes with it in the future.

The penalty was around 30% overhead for the queries I tested with. But I would expect this to differ depending on the exact CASE expression (e.g., number of WHEN branches, complexity of operand). I think that would definitely be worth a micro benchmark.

src/parser/expression/case_expression.cpp
```diff
@@ -1,13 +1,64 @@
...
 
 CaseExpression::CaseExpression() : ParsedExpression(ExpressionType::CASE_EXPR, ExpressionClass::CASE) {
 }
 
 string CaseExpression::ToString() const {
-	return ToString<CaseExpression, ParsedExpression>(*this);
+	if (!case_operand) {
+		return ToString<CaseExpression, ParsedExpression>(*this);
+	}
+	string case_str = "CASE " + case_operand->ToString();
```
Mytherin, line 15: `case_operand` should probably also be placed within brackets?

src/parser/query_node/select_node.cpp
```diff
@@ -8,6 +11,59 @@ SelectNode::SelectNode()
     : QueryNode(QueryNodeType::SELECT_NODE), aggregate_handling(AggregateHandling::STANDARD_HANDLING) {
 }
 
+static bool ContainsSimpleCase(QueryNode &node);
+
+static bool ContainsSimpleCase(const ParsedExpression &expression) {
+	if (expression.GetExpressionClass() == ExpressionClass::CASE && expression.Cast<CaseExpression>().CaseOperand()) {
+		return true;
+	}
+	if (expression.GetExpressionClass() == ExpressionClass::SUBQUERY) {
```
Mytherin, line 20: This now fails:

```sql
CREATE MACRO m() AS (SELECT 2);
SELECT CASE m() WHEN 1 THEN 'a' WHEN 2 THEN 'b' END;
```

Could you investigate why? We shouldn't be copying the case operand at all anymore now, no?

Mytherin (commented):
Thanks! Looks good - some comments

PR #24942, opened 2026-08-22
Support multi-index scans with full filter coverage

idle-lab:
Hi team,

As mentioned in the [discussion](https://github.com/duckdb/duckdb/discussions/24148), DuckDB currently does not support combining multiple indexes for a table scan. I‘m interested in this feature and want to implement it.

To keep the changes easier to review and verify, I plan to split the implementation into two PRs:

1. Support multi-index scans when all filters pushed down into the table scan can be fully covered by indexes.
2. Add filter-aware row fetching for index scans, allowing uncovered predicates to be evaluated as residual filters after fetching candidate rows.

### What this PR does

This PR enables a table scan to use multiple single‑column ART indexes. 

During index selection, it first checks whether every pushed‑down table filter can be matched to a single‑column ART index. If all filters are covered: 
1. Each selected ART index is scanned independently. 
2. The row IDs produced by the different ART indexes are intersected 
3. The resulting row IDs are fetched using the existing index scan path. 
 
Otherwise, no ART scan is performed and the it falls back to a table scan.

### Performance results

The benchmark uses:
- A 10M-row pseudo-randomly ordered 2D table
- Two single-column ART indexes
- 2 execution threads
- 10 paired samples per binary
- A release build
- A warm-up run before timed samples

<p>The physical order is permuted so zonemaps cannot prune the relevant row groups. Equality selectivity is <code dir="ltr">0.04%</code> for <code dir="ltr">a</code> and <code dir="ltr">0.025%</code> for <code dir="ltr">b</code>, both below the default <code dir="ltr">0.1%</code> index-scan threshold. Data generation and index construction use a persistent benchmark cache and are not included in query timings.

Query | Base | This PR | Speedup
-- | -- | -- | --
Two-index point lookup (a = 42 AND b = 3141) | 7.1 ms | 0.782 ms | 9.1×
Point + one-sided range (a = 42 AND b > 3997) | 7.7 ms | 1.0 ms | 7.7×


src/function/table/table_scan.cpp
```diff
@@ -809,6 +745,138 @@ bool TryScanIndex(ART &art, IndexEntry &entry, const ColumnList &column_list, Ta
...
+		                  max_count, candidate_row_ids)) {
+			continue;
+		}
+		scanned_filters.insert(candidate.filter_index);
+		if (first_index) {
+			row_ids = std::move(candidate_row_ids);
+			first_index = false;
+		} else {
+			IntersectRowIds(row_ids, candidate_row_ids);
+		}
```
vvvdwbvvv, line 896: Could we short-circuit here when row_ids.empty() and add a regression test? After a successful TryScanIndex, an empty result from any conjunct proves that the persistent side is empty. Currently, if another candidate exceeds max_count, scanned_filters remains incomplete and we fall back to a sequential scan. DuckIndexScan still scans local storage with the full filter set, so returning true here remains correct for transaction-local rows.

test/sql/index/art/scan/test_multi_index_scan.test
```diff
@@ -0,0 +1,202 @@
...
+
+query II
+EXPLAIN ANALYZE SELECT payload FROM multi_index WHERE a = 42 AND b = 7;
+----
+analyzed_plan	<REGEX>:.*Type: Index Scan.*
+
+query I
+SELECT payload FROM multi_index WHERE a = 42 AND b = 7;
+----
+4207
```
vvvdwbvvv, line 29: Maybe we can add two focused cases here: a supported one-sided range combined with another indexed predicate, e.g. a >= 40 AND b = 42, and a query intersecting three separate single-column indexes. Both paths currently work, but the tests only cover two-index equality/IN scans. If the empty-result short-circuit is added, please include its regression case here as well.

idle-lab:
@vvvdwbvvv Thanks for your review! I solved the problems you mentioned.

vvvdwbvvv:
All good for me now, can you cc a member for review

idle-lab:
@Mytherin Could you take a look at this when you have a chance? Thanks!

idle-lab:
While self-reviewing this PR, I found a correctness issue #25155 in the main branch and opened https://github.com/duckdb/duckdb/pull/25156 to fix it.

idle-lab:
A Table Scan fallback may be occur when a series of indexs have been scanned. That will cause a performance regression.

Execution path | Median
-- | --
Direct full scan | 7.792 ms
Index attempts followed by full scan | 8.686 ms
Regression | +11.5%

PR #24947, opened 2026-08-22
Abandon() and any_abandoned

carlopi:
Fix two issues with current coupling between `ht.Abandon()` and setting `gstate.any_abandoned = true`: it must happen in all call site for which `any_abandoned` might still be consulted, and correct ordering that prevent races is set first, work later.

Fixes a bug introduced in https://github.com/duckdb/duckdb/pull/22211.
Found by stress testing own private test suite, in this case via `HASH_ZERO=1` configuration where:
```sql
CREATE TABLE a AS SELECT i AS x, (i*7)%997
AS y FROM range (3000) t(i);
CREATE TABLE b AS SELECT i AS x, (i*11)%997 AS y FROM range(3000) t(i);
SELECT sum( (SELECT count (*) FROM b WHERE b.x > a.x AND b.y < a.y)) FROM a;
```
would fail like:
```
Invalid Input Error: More than one row returned by a subquery used as an expression - scalar subqueries can only return a single row
```
instead of returning the correct `2200518`.

To be considered a `gstate.Abandon(ht)` that could package both `any_abandoned = true; ht.Abandon();` in an utility function.

### DIFF
--- a/src/execution/radix_partitioned_hashtable.cpp
+++ b/src/execution/radix_partitioned_hashtable.cpp
@@ -833,8 +833,8 @@ void RadixPartitionedHashTable::Sink(ExecutionContext &context, DataChunk &chunk
 		// 'Reset' the HT without taking its data, we can just keep appending to the same collection
 		// This only works because we never resize the HT
 		// We don't do this when running with 1 or 2 threads, it only makes sense when there's many threads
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks!

PR #24950, opened 2026-08-23
Add hive_sample_size option to sample files in binder

xevix:
Follow-up from #24385, split off to make it easier to review.

Adds a `hive_sample_size` option that changes the Binder's Hive detection and partition type checking to sample a subset of the files instead of the whole dataset. The primary motivation is preparation work for lazy listing and early pruning of Hive partitions. 

`hive_sample_size` defaults to -1, i.e. unset, which preserves existing behavior.

Note that due to sampling, Hive detection and partition type checking will necessarily be incomplete and so could provide false positives, but since this is an opt-in optimization I think that's a fair tradeoff.

```sql
-- Local filesystem expands enough paths to get 100 files thanks to LazyMultiFileList
SELECT * FROM read_parquet('path/to/**/*.parquet', hive_sample_size = 100);

-- Other filesystems like S3 could theoretically similarly benefit from this
SELECT * FROM read_parquet('s3://path/to/**/*.parquet', hive_sample_size = 100);
```

For Hive datasets with a lot of files/folders, reduces query time a bit, so this also stands on its own.

```sql
memory D SELECT COUNT(*) FROM read_parquet('data/noaa/by_year/**/*.parquet', hive_sample_size=100);
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│ 3.13 billion │
└──────────────┘
Run Time (s): real 0.657 user 1.002071 sys 2.924276
memory D SELECT COUNT(*) FROM read_parquet('data/noaa/by_year/**/*.parquet');
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│ 3.13 billion │
└──────────────┘
Run Time (s): real 0.709 user 1.053385 sys 2.951471
```


xevix:
@samansmink The idea for implementation of early Hive path pruning has evolved a fair bit over time. I think this is a small first step that makes sense, but please let me know if it makes more sense to move back to Discussions before continuing, or if there's a better way to move things forward.

samansmink (approved):
I think this is a sane first step!

PR #24953, opened 2026-08-23
Prune row group for prefix functions

dentiny:
Hi team, I found prefix doesn't always prune row group as expected, for example,
```sql
memory D ATTACH '/tmp/prefix_zonemap_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE prefix_pruning AS
     SELECT i::INTEGER AS i, lpad(i::VARCHAR, 6, '0') AS s
     FROM range(6144) r(i);
-- no value starts with '00', so no row group should be scanned.
db D EXPLAIN ANALYZE SELECT count(*) FROM prefix_pruning WHERE NOT prefix(s, '00');
╭─ Summary ───────────╮
│ Total Time: 0.0015s │
╰─────────────────────╯
╰───────────────┬───────────────╯
╭─ Table Scan ──┴───────────────╮
│ Table: db.main.prefix_pruning │
│ Filters: NOT prefix(s, '00')  │
│ Row Groups Scanned: 3 / 3     │
│ 0 rows                    0µs │
```
This PR updates two cased for row group pruning for prefix scan filter:
- Set pruning callback for `starts_with`, previously it's only set to `prefix` function
- When min/max already to starts with prefix, we could return ALWAYS_TRUE directly

### DIFF
--- a/src/function/scalar/string/string_common.cpp
+++ b/src/function/scalar/string/string_common.cpp
@@ -3,6 +3,7 @@
 #include "duckdb/common/types/string_type.hpp"
 #include "duckdb/common/types/value.hpp"
 #include "duckdb/planner/expression/bound_constant_expression.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
 #include "duckdb/storage/statistics/base_statistics.hpp"
 #include "duckdb/storage/statistics/string_stats.hpp"
 
--- a/src/function/scalar/string/string_common.cpp
+++ b/src/function/scalar/string/string_common.cpp
@@ -1,0 +1,98 @@
+#include "duckdb/function/scalar/string_common.hpp"
+
+#include "duckdb/common/types/string_type.hpp"
+#include "duckdb/common/types/value.hpp"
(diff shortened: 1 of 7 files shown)


src/function/scalar/string/string_common.cpp
```diff
@@ -0,0 +1,98 @@
...
+		    StringStats::CompareStringStats(string_t(upper_bound.c_str(), upper_bound.size()),
+		                                    string_t(min.c_str(), min.size()), StringStats::GetMinType(*column_stats));
+		if (min_compare < 0) {
+			return FilterPropagateResult::FILTER_ALWAYS_FALSE;
+		}
+		if (min_compare == 0 && StringStats::GetMinType(*column_stats) == StringStatsType::EXACT_STATS) {
+			return FilterPropagateResult::FILTER_ALWAYS_FALSE;
+		}
+	}
+	// Cases where min and max both start with the prefix.
```
dentiny, line 88: This is the new logic, all above are copied old implementation.

PR #21831, opened 2026-04-04
Partial aggregate precomputation from row group statistics

Damon07:
This PR extends StatisticsPropagator to support partial precomputation of aggregates (min, max, count_star) from row group statistics, with a PROJECTION merge layer — an approach suggested in the review of #21797. The resulting implementation is significantly simpler.

Performance
Data Set:  tpch_sf10
SQL: SELECT min(l_linenumber), max(l_linenumber) FROM lineitem WHERE l_orderkey < 30000000;
Machine: Macbook pro m2 16G
| Benchmark | This PR  | Nightly  | Speedup |
|-----------|--------------|---------------|---------|
| `min, max` | 0.6ms | 14ms | **23x** |

PR #24962, opened 2026-08-23
Avoid plan-time partial precomputation when not complete

carlopi:
https://github.com/duckdb/duckdb/pull/21831 introduced a super cool optimization, but I think that under load (like multiple connections inserting data) then numbering of row group blocks.

Current proposal is avoiding the optimization when it can't be applied for all row groups, that should keep some improvements, but also bear a cost. I am not sure what's the path to stable numbering for row groups, but I can imagine that would allow to re-enable this.

Problematic test is `test/sql/parallelism/interquery/concurrent_checkpoint_insert.test`, in particular with system slowed down (possibly by TSan build), since scheduling here is needed to align for the bad path to occur.


Mytherin:
Thanks! I think this issue might pop up elsewhere so this should be solved properly (e.g. by caching the row group collection pointer).

PR #24964, opened 2026-08-23
Make `FunctionSet` and `Bound*Function` contain `shared_ptr<const Function>`s

Maxxen:
This makes it easier to identify which function a "bound" function was instantiated from, as well as makes it cheaper to "copy" functions and function sets (especially now that they carry `FunctionSignature`s.

Most of the diff is the patches, almost all the changes in the files are just one-line dereference-the-shared-pointer edits.

### DIFF
--- a/src/function/aggregate/distributive/minmax.cpp
+++ b/src/function/aggregate/distributive/minmax.cpp
@@ -402,7 +402,10 @@ unique_ptr<FunctionData> BindMinMax(BindAggregateFunctionInput &input) {
 	auto expr = minmax_func.Bind(context, std::move(arguments));
 	arguments = std::move(expr->GetChildrenMutable());
 
+	auto definition = function.GetDefinition();
 	function = std::move(expr->FunctionMutable());
+	// the specialized implementation is not the function we were bound from
+	function.SetDefinition(std::move(definition));
 	return std::move(expr->BindInfoMutable());
 }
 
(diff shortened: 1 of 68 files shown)


Mytherin:
Thanks!

src/function/aggregate/distributive/minmax.cpp
```diff
@@ -402,7 +402,10 @@ unique_ptr<FunctionData> BindMinMax(BindAggregateFunctionInput &input) {
 	auto expr = minmax_func.Bind(context, std::move(arguments));
 	arguments = std::move(expr->GetChildrenMutable());
 
+	auto definition = function.GetDefinition();
 	function = std::move(expr->FunctionMutable());
+	// the specialized implementation is not the function we were bound from
+	function.SetDefinition(std::move(definition));
```
carlopi, line 408: Shouldn't 405-408 be an helper function?
Like:
```c++
function.AssignWhileKeepingDefinition(std::move(other));
```
it seems to be this `GetDefinition` + `operator=` + `SetDefinition` is bound to be repeated / create issues.

Maxxen: Yeah maybe, there is a .ReplaceImplementation() for aggregates - but I also feel like this is an anti-pattern in general and should not really be incentivized.

PR #20580, opened 2026-01-19
Allow extension to register custom compression filesystem

dentiny:
Hi team, I've seen multiple discussion threads from duckdb users calling for more compression supports, like [1](https://github.com/duckdb/duckdb/discussions/12232) and [2](https://github.com/duckdb/duckdb/discussions/20475).

For now, there're two workarounds I could see
- [shellfs](https://github.com/query-farm/shellfs): which allows user to interact with the shell environment
- [zipfs](https://github.com/isaacbrodsky/duckdb-zipfs): which adds zip (de)compressor support but have to tolerate additional filepath prefix

I would personally prefer better integration with duckdb filesystem if possible.
The biggest obstacle preventing extensions from making their own extension is the enum [`FileCompressionType`](https://github.com/duckdb/duckdb/blob/1c62e11b8201923466431e7d109a6ba8b3710c7b/src/include/duckdb/common/enums/file_compression_type.hpp#L15). 
It makes sense for duckdb core, because it only needs to maintain limited number of compressions, and min required dependencies (good for builds and CI as well).

In this PR, I propose to add support to allow extensions to register their own compression filesystems into duckdb filesystem, by simply adding extra check for externally registered compression filesystems, so it doesn't affect current duckdb compatibility or performance.

PR #24968, opened 2026-08-24
Allow extension to register custom compression filesystem

dentiny:
Revive and rework https://github.com/duckdb/duckdb/pull/20580, I still think it's a helpful feature to get in, copy the PR description here.

Hi team, I've seen multiple discussion threads from duckdb users calling for more compression supports, like [1](https://github.com/duckdb/duckdb/discussions/12232) and [2](https://github.com/duckdb/duckdb/discussions/20475).

For now, there're two workarounds I could see
- [shellfs](https://github.com/query-farm/shellfs): which allows user to interact with the shell environment
- [zipfs](https://github.com/isaacbrodsky/duckdb-zipfs): which adds zip (de)compressor support but have to tolerate additional filepath prefix

I would personally prefer better integration with DuckDB filesystem if possible.
The biggest obstacle preventing extensions from making their own extension is the enum [`FileCompressionType`](https://github.com/duckdb/duckdb/blob/1c62e11b8201923466431e7d109a6ba8b3710c7b/src/include/duckdb/common/enums/file_compression_type.hpp#L15), so in this PR I update it completely to string type.
It makes sense for DuckDB core, because it only needs to maintain limited number of compressions, and min required dependencies (good for builds and CI as well).

In this PR, I propose to add support to allow extensions to register their own compression filesystems into duckdb filesystem, by simply adding extra check for externally registered compression filesystems, so it doesn't affect current DuckDB compatibility or performance.

### DIFF
--- a/.github/patches/extensions/spatial/0007-compression-file-open-options.patch
+++ b/.github/patches/extensions/spatial/0007-compression-file-open-options.patch
@@ -1,0 +1,28 @@
+commit 16f027631e496e8ed902f9ff64722a93aa88d03e
+Author: dentinyhao <dentinyhao@gmail.com>
+Date:   Mon Aug 24 02:47:13 2026 -0700
+
+    Fix compression on file open option
+
+diff --git a/src/spatial/modules/shapefile/shapefile_module.cpp b/src/spatial/modules/shapefile/shapefile_module.cpp
+index 6ef8fb5..fe81e1f 100644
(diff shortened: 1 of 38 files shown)


Mytherin:
Thanks! LGTM.

We very likely will still want to rework how compressed files work in the future as the current APIs are not very performant, but who knows when we will get to that, and this is an improvement for now :)

PR #24970, opened 2026-08-24
Dont remove lambda expressions from functions

Maxxen:
Similar to #24594, this PR changes the lambda-taking functions to not delete the bound lambda expression from their arguments.

Compatibility-wise, we can "recover" the lambda expression from the bind-data when we read old plans - but we don't do anything special for forwards compatibility, as IIRC we don't read new plans on old storage versions anyway(?). We also now serialize the lambda parameter names in the expression too so that we can print the full expression in the explain output.

This means that we also need to instantiate vectors in the input data chunk corresponding to an lambda argument slot, but since lambdas don't really have a value, we just instantiate a dummy constant TINYINT vector. In the future we could maybe instantiate this vector with the lambda return type, so that invoking a lambda has a natural place to put the result.

Additionally we also prevent the CSE optimizer/expression iterator from traversing the bound lambda body as it gets confused around lambda parameter bindings, but this can (and should!) probably be relaxed in the future so that we can optimize lambda bodies too.


Mytherin:
Thanks!

PR #24972, opened 2026-08-24
Fix out of bounds read when a GZIP member is missing its footer

### LINKED ISSUE #24955: Out-of-bounds read in Parquet GZIP decompression when a GZIP member lacks its 8-byte footer
> Filing as a regular bug (not a security advisory) per DuckDB's SECURITY.md — a crafted Parquet file is
> out of the vulnerability boundary, but this is a clear memory-safety bug worth fixing.

### What happens

`MiniZStream::Decompress` (used for Parquet GZIP page decompression) advances past each GZIP member by
`GZIP_FOOTER_SIZE + stream.total_in` **without first checking that the 8 footer bytes are actually present**
in the remaining buffer:

```cpp
// third_party/miniz/miniz_wrapper.hpp  (member loop)
while (compressed_size > 0) {                          // :56
    if (compressed_size < GZIP_HEADER_MINSIZE) { ... }  // :58  (header minimum only)
    auto gzip_hdr = (const unsigned char *) compressed_data;   // :61
    ... header checks ...
    compressed_data += GZIP_HEADER_MINSIZE;             // :66
    compressed_size -= GZIP_HEADER_MINSIZE;             // :67
    ... inflate the deflate stream (MZ_STREAM_END) ...  // :82-85
    compressed_data += GZIP_FOOTER_SIZE + stream.total_in;   // :89
    compressed_size -= GZIP_FOOTER_SIZE + stream.total_in;   // :90  <-- can underflow (unsigned)
}
```

If a member's inner deflate stream ends with **fewer than 8 bytes remaining** (a truncated/absent footer),
the unsigned subtraction at `:90` underflows to a huge value, `while (compressed_size > 0)` stays true, and
the next iteration dereferences `gzip_hdr = compressed_data` **past the end of the page buffer** at `:61`,
reading `gzip_hdr[0..3]` out of bounds.

Reachable when reading a crafted Parquet file with a GZIP-compressed page via `read_parquet` (the GZIP
branch calls `MiniZStream::Decompress` from `extension/parquet/column_reader.cpp:538-546`).

### Reproduction (AddressSanitizer)

A standalone harness that copies the exact member-loop arithmetic + constants from
`third_party/miniz/miniz_wrapper.hpp` and feeds a member whose footer bytes are absent (full source below):

```
$ g++ -g -O0 -fsanitize=address -fno-omit-frame-pointer poc_gzip_footer.cpp -o poc_gzip_footer
$ ./poc_gzip_footer
==...==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000024 ...
READ of size 1 at 0x502000000024 thread T0
SUMMARY: AddressSanitizer: heap-buffer-overflow poc_gzip_footer.cpp:34 in copied_member_loop_model
```

The harness models the arithmetic faithfully; an end-to-end trigger needs a crafted `.parquet` with a GZIP
page whose deflate stream consumes all bytes leaving no footer.

<details><summary>Full reproducer — <code>poc_gzip_footer.cpp</code></summary>

```cpp
#include <cstddef>
#include <cstdint>
#include <cstdlib>
#include <cstring>
#include <iostream>

static constexpr uint8_t GZIP_HEADER_MINSIZE = 10;       // from third_party/miniz/miniz_wrapper.hpp:56-67
static constexpr uint8_t GZIP_FOOTER_SIZE = 8;           // from third_party/miniz/miniz_wrapper.hpp:88-90
static constexpr uint8_t GZIP_COMPRESSION_DEFLATE = 0x08; // from third_party/miniz/miniz_wrapper.hpp:62
static constexpr unsigned char GZIP_FLAG_UNSUPPORTED =
    0x1 | 0x2 | 0x4 | 0x10 | 0x20; // from third_party/miniz/miniz_wrapper.hpp:62-63

struct FakeStream {
	size_t total_in;
	size_t total_out;
};

static void copied_member_loop_model(const char *compressed_data, size_t compressed_size) {
	FakeStream stream {};
	int iterations = 0;
	const auto base = reinterpret_cast<uintptr_t>(compressed_data);

	// Loop over blocks
	while (compressed_size > 0) { // from third_party/miniz/miniz_wrapper.hpp:56
		std::cout << "iteration=" << iterations << " compressed_size=" << compressed_size
		          << " offset=" << (reinterpret_cast<uintptr_t>(compressed_data) - base) << "\n";

		// Read block header
		if (compressed_size < GZIP_HEADER_MINSIZE) { // from third_party/miniz/miniz_wrapper.hpp:58
			std::cout << "header-minimum check rejected\n"; // from third_party/miniz/miniz_wrapper.hpp:58-60
			return;
		}
		auto gzip_hdr = reinterpret_cast<const unsigned char *>(compressed_data); // from third_party/miniz/miniz_wrapper.hpp:61
		if (gzip_hdr[0] != 0x1F || gzip_hdr[1] != 0x8B || gzip_hdr[2] != GZIP_COMPRESSION_DEFLATE ||
		    gzip_hdr[3] & GZIP_FLAG_UNSUPPORTED) { // from third_party/miniz/miniz_wrapper.hpp:62-63
			std::cout << "header check rejected\n"; // from third_party/miniz/miniz_wrapper.hpp:64
			return;
		}
		compressed_data += GZIP_HEADER_MINSIZE; // from third_party/miniz/miniz_wrapper.hpp:66
		compressed_size -= GZIP_HEADER_MINSIZE; // from third_party/miniz/miniz_wrapper.hpp:67

		// Model the MZ_STREAM_END branch with a deflate stream consuming all remaining inner bytes.
		stream.total_in = compressed_size; // from third_party/miniz/miniz_wrapper.hpp:82-85
		stream.total_out = 0;              // from third_party/miniz/miniz_wrapper.hpp:91-92

		// Update indices
		compressed_data += GZIP_FOOTER_SIZE + stream.total_in; // from third_party/miniz/miniz_wrapper.hpp:89
		compressed_size -= GZIP_FOOTER_SIZE + stream.total_in; // from third_party/miniz/miniz_wrapper.hpp:90

		if (++iterations > 2) {
			std::cout << "unexpectedly survived repeated loop\n";
			return;
		}
	}
}

int main() {
	const size_t body_bytes = 2;
	const size_t footer_bytes_present = 0;
	const size_t provided_size = GZIP_HEADER_MINSIZE + body_bytes + footer_bytes_present;
	char *buf = static_cast<char *>(std::malloc(provided_size));
	if (!buf) {
		return 2;
	}
	std::memset(buf, 0, provided_size);
	buf[0] = 0x1F;
	buf[1] = static_cast<char>(0x8B);
	buf[2] = GZIP_COMPRESSION_DEFLATE;
	buf[3] = 0;

	std::cout << "provided_size=" << provided_size << " body_bytes=" << body_bytes
	          << " footer_bytes_present=" << footer_bytes_present << "\n";
	copied_member_loop_model(buf, provided_size);
	std::free(buf);
	return 0;
}
```

</details>

### Impact

Out-of-bounds **read** → crash / DoS on a process that reads a crafted Parquet file. No OOB write or code
execution shown.

### Suggested fix

Verify the footer is present before advancing past it:

```cpp
if (compressed_size < GZIP_FOOTER_SIZE + stream.total_in) {
    // truncated/absent footer: reject instead of underflowing
    throw ...;   // or break with an error
}
compressed_data += GZIP_FOOTER_SIZE + stream.total_in;
compressed_size -= GZIP_FOOTER_SIZE + stream.total_in;
```

### Found with

Argo, an AI-assisted static-analysis tool I develop and maintain; I verified and reproduced this myself.

Robinbinu:
Fixes https://github.com/duckdb/duckdb/issues/24955

`MiniZStream::Decompress` walks a gzip stream member by member, skipping the deflate bytes it consumed plus the 8 byte footer at the end of each one. Nothing checked that those footer bytes were actually in the buffer, so a member whose deflate stream ends with fewer than 8 bytes left underflows `compressed_size`, which is a `size_t`. The loop condition stays true and the next iteration reads a gzip header past the end of the input. The parquet reader takes this path for `GZIP` pages and sizes the buffer from file metadata, so a crafted file gets there through `read_parquet`.

Compiling `third_party/miniz` as it is and handing `Decompress` a member with its last byte removed:

```
AddressSanitizer: heap-buffer-overflow
READ of size 1
  duckdb::MiniZStream::Decompress(char const*, unsigned long, char*, unsigned long) miniz_wrapper.hpp:62
located 1 bytes after 38-byte region
```

Line 62 is the `gzip_hdr[0]` read at the top of the member loop.

The test asserts on the error message rather than only that the call throws. Before this change the out of bounds read usually landed on bytes that failed the magic number check, so the call threw anyway and a plain check for throwing would have passed on the broken code too. Without a sanitizer the unpatched build reports `Input is invalid/unsupported GZIP stream` and the patched one reports `incomplete gzip member footer`, so the test does fail on unpatched code in a normal build.

### DIFF
--- a/test/common/CMakeLists.txt
+++ b/test/common/CMakeLists.txt
@@ -10,6 +10,7 @@ add_library_unity(
   test_column_data_allocator.cpp
   test_file_buffer_handle_group.cpp
   test_file_system.cpp
+  test_gzip_member_oob.cpp
   test_caching_file_system_wrapper.cpp
   test_external_file_cache.cpp
   test_hyperlog.cpp
--- a/test/common/test_gzip_member_oob.cpp
+++ b/test/common/test_gzip_member_oob.cpp
@@ -1,0 +1,92 @@
+#include "catch.hpp"
+#include "duckdb/common/helper.hpp"
+#include "miniz_wrapper.hpp"
+
+#include <cstring>
+#include <string>
+
+using namespace duckdb;
+
+namespace {
+
+// Builds one complete gzip member (header + deflate stream + footer) into an exactly sized buffer.
+unsafe_unique_array<char> CompressToExactBuffer(const std::string &input, size_t &member_size) {
+	auto scratch_size = MiniZStream::MaxCompressedLength(input.size());
+	auto scratch = make_unsafe_uniq_array<char>(scratch_size);
+
+	MiniZStream compressor;
+	member_size = scratch_size;
+	compressor.Compress(input.c_str(), input.size(), scratch.get(), &member_size);
+
+	auto member = make_unsafe_uniq_array<char>(member_size);
+	memcpy(member.get(), scratch.get(), member_size);
+	return member;
+}
+
+// Copies the first `keep` bytes into an allocation of exactly that size, so a read past the end of
+// the member lands outside a real allocation rather than in slack the allocator happened to give us.
+unsafe_unique_array<char> TruncateToExactBuffer(const unsafe_unique_array<char> &member, size_t keep) {
+	auto truncated = make_unsafe_uniq_array<char>(keep);
+	memcpy(truncated.get(), member.get(), keep);
+	return truncated;
+}
+
+} // namespace
+
+TEST_CASE("gzip member missing its footer is rejected", "[miniz]") {
+	const std::string payload(4096, 'q');
+
+	size_t member_size;
+	auto member = CompressToExactBuffer(payload, member_size);
+	REQUIRE(member_size > MiniZStream::GZIP_HEADER_MINSIZE + MiniZStream::GZIP_FOOTER_SIZE);
+
+	// The intact member round trips.
+	{
+		auto out = make_unsafe_uniq_array<char>(payload.size());
+		MiniZStream decompressor;
+		decompressor.Decompress(member.get(), member_size, out.get(), payload.size());
+		REQUIRE(memcmp(out.get(), payload.c_str(), payload.size()) == 0);
+	}
+
+	// Dropping any part of the 8 byte footer used to underflow the remaining-size counter, so the
+	// member loop ran once more and read a gzip header out of bounds. It has to be reported as a
+	// truncated member instead. Checking the message matters: the out of bounds read usually landed
+	// on bytes that failed the magic check, so "it throws" alone held before the fix as well.
+	for (size_t missing = 1; missing <= MiniZStream::GZIP_FOOTER_SIZE; missing++) {
+		auto keep = member_size - missing;
+		auto truncated = TruncateToExactBuffer(member, keep);
+
+		auto out = make_unsafe_uniq_array<char>(payload.size());
+		MiniZStream decompressor;
+		bool threw = false;
+		try {
+			decompressor.Decompress(truncated.get(), keep, out.get(), payload.size());
+		} catch (const std::exception &ex) {
+			threw = true;
+			REQUIRE(std::string(ex.what()).find("incomplete gzip member footer") != std::string::npos);
+		}
+		REQUIRE(threw);
+	}
+}
+
+TEST_CASE("concatenated gzip members still decompress", "[miniz]") {
+	// The bounds check sits inside the member loop, so keep a multi member stream covered.
+	const std::string first(1024, 'a');
+	const std::string second(2048, 'b');
+
+	size_t first_size;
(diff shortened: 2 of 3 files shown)


Tishj (changes requested):
Thanks, the fix makes sense, but I'd prefer a non-cpp test. What we usually do for a data corruption test like this is upload the file to the `data` directory and write a sql test that uses that file.

See for example `test/parquet/parquet_fuzzer_issues.test`

PR #24975, opened 2026-08-24
Fix `Value::ToSQLString` nested type round-tripping

dentiny:
This PR fixes round-trip for STRUCT, BIT, and UNION values.


Mytherin:
Thanks!

PR #24976, opened 2026-08-24
feature: fetch selected rows for overflow string scans

fanvanzh:
A vector that spans column segments cannot use the compression specific Select, so the generic path scans the entire vector and slices it. 
Every string at or above the string block limit gets its own overflow buffer, so a point lookup materializes STANDARD_VECTOR_SIZE buffers to return a single row, which can push the scan out of memory. 
At low selectivity fetch only the selected rows through the batched FetchRows API instead.

### DIFF
--- a/src/include/duckdb/function/compression_function.hpp
+++ b/src/include/duckdb/function/compression_function.hpp
@@ -10,6 +10,7 @@
 
 #include "duckdb/common/common.hpp"
 #include "duckdb/common/enums/compression_type.hpp"
+#include "duckdb/common/enums/scan_vector_type.hpp"
 #include "duckdb/common/map.hpp"
 #include "duckdb/common/insertion_order_preserving_map.hpp"
 #include "duckdb/common/mutex.hpp"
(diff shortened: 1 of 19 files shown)


fanvanzh:
image

Significantly reduces the memory overhead of OVERFLOW_STRINGS.

Mytherin:
Thanks but this is the wrong place to implement this. This is a compression algorithm specific optimization that should not be in the `ColumnData`, but in the specific compression function implementation.

fanvanzh:
Thanks — you're right, moved it into the compression functions. 

This needed two extra arguments on compression_select_t:
- result_offset, since a vector can be assembled from several segments;
- scan_type, because emitting a compressed result (constant in RLE/numeric_constant, dictionary in dict_fsst) is only legal for a whole-vector scan.

StandardColumnData::Select used to bail to the generic scan-and-slice path whenever scan_type != SCAN_ENTIRE_VECTOR. In my case (uncompressed BLOB, 16 KiB blocks) a segment holds fewer rows than a vector, so every vector spans a boundary and the compression-specific select was never reached — hence the offset. That gate now only remains for updates.

It is a breaking change for out-of-tree compression functions; I skipped adding a second optional callback to keep the code simple, happy to switch if you'd prefer.

Also: ValiditySelect was implemented but never assigned to the function, so that path was dead.

fanvanzh:
@Mytherin  Do you have any more comments on this patch?

PR #24979, opened 2026-08-24
GetDefaultSchema returns optional<Identifier>

Tmonster:
Certain remote catalogs/Lakehouse catalogs do not have a default schema, and core seems to want to resolve ambiguous references to a default schema always (i.e Catalog::TryLookupDefaultSchema). `CatalogSearchPath` also does not provide overrides

In certain cases, if a catalog wants to avoid ambiguous lookups, that is not possible, since the default schema will still be used. For remote catalogs like iceberg, we may want to avoid this expensive lookup, and we want to avoid Binder Errors using default schema names that do not exist (in case a user did not provide one).

This cannot be done extension side only, because the empty string `""` is already registered internally as `INVALID_SCHEMA`. If core sees `INVALID_SCHEMA` as the DefaultSchema, it will try all schemas for the catalog when attempting to resolve a table or function. For remote catalogs, this is undesireable.

Overriding `CheckAmbiguousCatalogOrSchema` to return false will not help here, since it is only for catalog or schema ambiguity, but not table resolution. 

Therefore, if we change the `GetDefaultSchema` function to return optional<Identifier> instead, then extensions can avoid having core DuckDB perform lookups on default schemas that are technically not a default schema for that catalog.

The storage extension API changes slightly, and extensions can maintain the same behavior with minor changes. 
This will help with Iceberg since Iceberg does not have a concept of a default_schema, and we do not want to muddy attach options with a "default_schema" parameter that doesn't exist for other iceberg engines.

### DIFF
--- a/.github/config/extensions/fts.cmake
+++ b/.github/config/extensions/fts.cmake
@@ -4,4 +4,5 @@ duckdb_extension_load(fts
         GIT_URL https://github.com/duckdb/duckdb-fts
         GIT_TAG 2266e67d01c737cb65114e77c7c672e58869a635
         TEST_DIR test/sql
+        APPLY_PATCHES
 )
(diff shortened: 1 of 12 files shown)


Mytherin:
Thanks!

PR #24980, opened 2026-08-24
convert recursion to stack-based transversal of variant to parquet

xproblematique:
Closes https://github.com/duckdblabs/duckdb-internal/issues/10306

Fix unbounded recursion when converting deeply nested VARIANT arrays and objects to Parquet.

`EmitIterator` previously used recursion to transverse nested VARIANT values. A deeply nested VARIANT could exhaust the native stack and cause a stack overflow.

### Changes

* Replace recursive traversal in `EmitIterator` with stack-based traversal.
* Track the current container and child index in each frame, allowing deeply nested values to be processed without growing the native call stack.
* Reserve the stack up front using the nesting limit to avoid reallocations during traversal.
* Add an initial VARIANT nesting depth of 8 and grow the stack as needed using `NextPowerOfTwo`.
* Collapse `BeginArray` and `BeginObject` into `BeginContainer`, which returns an optional_idx to make the empty-container case explicit.

### Testing

Added regression coverage for VARIANT arrays around the initial stack capacity.

* Verified 7 and 8 levels are successfully converted without requiring stack growth.
* Verified 9 levels are successfully converted, proving stack growth beyond the initial capacity.
* Reproduced the original deeply nested (~15,000 levels) input and confirmed it now returns a controlled `InvalidInputException` instead of causing a stack overflow/crash.

### DIFF
--- a/src/include/duckdb/common/types/variant/variant_builder.hpp
+++ b/src/include/duckdb/common/types/variant/variant_builder.hpp
@@ -101,6 +101,8 @@ struct VariantBuilder {
 	//! maps a key string to its (unsorted) dictionary index, owned by the result's keys vector
 	OrderedOwningStringMap<uint32_t> &dictionary;
 
+	static constexpr idx_t MAX_RECURSION_DEPTH = 16;
+
 	//! the offsets at which the current row's entries begin
 	idx_t row_values = 0;
 	idx_t row_children = 0;
@@ -154,7 +156,10 @@ struct VariantBuilder {
 
 	//! Emit an ARRAY value with 'n' elements. 'emit_fn(i)' must emit exactly one value for element i.
 	template <class EMIT_FN>
-	void EmitArray(idx_t n, EMIT_FN &&emit_fn) {
+	void EmitArray(idx_t n, EMIT_FN &&emit_fn, idx_t depth = 0) {
+		if (depth >= MAX_RECURSION_DEPTH) {
+			throw InvalidInputException("VARIANT nesting exceeds maximum of %d", MAX_RECURSION_DEPTH);
+		}
 		auto byte_offset = NumericCast<uint32_t>(blob.size());
 		type_ids.push_back(static_cast<uint8_t>(VariantLogicalType::ARRAY));
 		byte_offsets.push_back(byte_offset);
@@ -474,7 +479,10 @@ auto CollectObjectChildren(const NODE &it) {
 
 //! Traverse a VariantNode-like cursor 'it' (any type exposing the node concept) into the builder.
 template <class NODE>
-void EmitIterator(const NODE &it, VariantBuilder &builder) {
+void EmitIterator(const NODE &it, VariantBuilder &builder, idx_t depth = 0) {
+	if (depth >= VariantBuilder::MAX_RECURSION_DEPTH) {
+		throw InvalidInputException("VARIANT nesting exceeds maximum of %d", VariantBuilder::MAX_RECURSION_DEPTH);
+	}
 	if (it.IsNull() || it.IsMissing()) {
 		builder.EmitNull();
 		return;
@@ -486,12 +494,13 @@ void EmitIterator(const NODE &it, VariantBuilder &builder) {
 		auto children = CollectObjectChildren(it);
 		builder.EmitObject(
 		    children.size(), [&](idx_t i) { return children[i].key; },
-		    [&](idx_t i) { EmitIterator(children[i].value, builder); });
+		    [&](idx_t i) { EmitIterator(children[i].value, builder, depth + 1); });
 		break;
 	}
 	case VariantLogicalType::ARRAY: {
 		auto array = it.GetArrayChildren();
-		builder.EmitArray(array.size(), [&](idx_t i) { EmitIterator(array[i], builder); });
+		builder.EmitArray(
+		    array.size(), [&](idx_t i) { EmitIterator(array[i], builder, depth + 1); }, depth + 1);
 		break;
 	}
 	default:
--- a/src/include/duckdb/common/types/variant/variant_builder.hpp
+++ b/src/include/duckdb/common/types/variant/variant_builder.hpp
@@ -29,6 +29,7 @@
 #include "duckdb/common/owning_string_map.hpp"
 #include "duckdb/common/numeric_utils.hpp"
 #include "duckdb/common/limits.hpp"
+#include "reader/variant/parquet_variant_iterator.hpp"
 
 #include <type_traits>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks for the PR! Looks good. I wonder if it would be possible to try and convert the recursive iteration into a stack-based iteration instead, e.g. by keeping a `vector<VariantIterationObject> stack`  and pushing elements there instead of relying on the call stack for iteration.

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -29,6 +29,7 @@
 #include "duckdb/common/owning_string_map.hpp"
 #include "duckdb/common/numeric_utils.hpp"
 #include "duckdb/common/limits.hpp"
+//#include "duckdb/extension/parquet/include/reader/variant/parquet_variant_iterator.hpp"
```
pdet, line 32: nit: accidental?

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -154,7 +157,10 @@ struct VariantBuilder {
 
 	//! Emit an ARRAY value with 'n' elements. 'emit_fn(i)' must emit exactly one value for element i.
 	template <class EMIT_FN>
-	void EmitArray(idx_t n, EMIT_FN &&emit_fn) {
+	void EmitArray(idx_t n, EMIT_FN &&emit_fn, idx_t depth = 0) {
```
pdet, line 160: Is this still necessary?

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -455,6 +461,47 @@ struct VariantBuilder {
 			throw InternalException("EmitPrimitiveNode: unhandled VariantLogicalType (%d)", static_cast<int>(type_id));
 		}
 	}
+
+	idx_t BeginObject(idx_t n) {
```
pdet, line 465: Can we collapse this and `BeginArray`

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -474,29 +521,80 @@ auto CollectObjectChildren(const NODE &it) {
...
+			auto &entry = frame.object_children[i];
+			builder.AssignObjectChild(frame.block, i, entry.key);
+			ProcessNode(entry.value);
+		} else {
+			builder.AssignArrayChild(frame.block, i);
+			ProcessNode((*frame.array_children)[i]);
+		}
+		// frame may be invalidated by ProcessNode's push_back reallocating `stack` —
+		// nothing below this line touches frame, and the loop re-fetches stack.back()
+		// fresh on the next iteration.
```
pdet, line 597: nit: lets avoid multi-line comments unless strictly necessary

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -474,29 +521,80 @@ auto CollectObjectChildren(const NODE &it) {
...
+		}
+		case VariantLogicalType::ARRAY: {
+			auto array_iter = node.GetArrayChildren();
+			auto n = array_iter.size();
+			auto block = builder.BeginArray(n);
+			if (block != DConstants::INVALID_INDEX) {
+				Frame frame;
+				frame.is_object = false;
+				frame.block = block;
+				frame.array_children = unique_ptr<ArrayIterT>(new ArrayIterT(std::move(array_iter)));
```
pdet, line 566: frame.array_children = make_uniq<ArrayIterT>(std::move(array_iter));

pdet: Also, do you see allocations in the hot-path as a potential issue?

pdet (commented):
Thanks for the PR. Is it possible for us to also add tests?

src/include/duckdb/common/types/variant/variant_builder.hpp
xproblematique: Yes :D

src/include/duckdb/common/types/variant/variant_builder.hpp
xproblematique: No - I see two options. Keep a `stack.size()` check or heap-bound. Let me know which way to go for this one.

src/include/duckdb/common/types/variant/variant_builder.hpp
xproblematique: Yes

src/include/duckdb/common/types/variant/variant_builder.hpp
xproblematique: Could be because those allocations did not exist in the recursive version and now we must alloc for `unique_ptr` and `Frame`. I can swap the unique_ptr<ArrayIterT> for optional<ArrayIterT> . This removes a heap allocation per array node

pdet: Yeah, but maybe to clarify, I think what we can do is to allocate the frame on MAX_RECURSION_DEPTH at initialization outside of the loop and then never reallocate again.

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -455,6 +457,32 @@ struct VariantBuilder {
 			throw InternalException("EmitPrimitiveNode: unhandled VariantLogicalType (%d)", static_cast<int>(type_id));
 		}
 	}
+
+	idx_t BeginContainer(VariantLogicalType type_id, idx_t n) {
```
taniabogatsch: Maybe we can return `optional_idx` here, that way we make it more explicit that the function can return an invalid index.

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -474,29 +502,90 @@ auto CollectObjectChildren(const NODE &it) {
...
+	struct Frame {
+		bool is_object = false;
+		idx_t index = 0;
+		idx_t block = 0;
+		vector<EntryT> object_children;
+		optional<ArrayIterT> array_children;
+	};
+
+	vector<Frame> stack;
+	stack.reserve(VariantBuilder::MAX_NESTING_DEPTH);
```
taniabogatsch: Maybe instead of the `MAX_NESTING_DEPTH` being a hard limit, we can use it as the initial reservation size. Then, if we exceed that size, we resize by the power of two. 

So `MAX_NESTING_DEPTH` becomes `INIT_NESTING_DEPTH`. And we could be a bit more conservative with pre-allocating here, say, set it to 8, but that's optional.

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -474,29 +502,90 @@ auto CollectObjectChildren(const NODE &it) {
...
+			return;
+		}
+		auto type_id = node.GetTypeId();
+		switch (type_id) {
+		case VariantLogicalType::OBJECT: {
+			auto children = CollectObjectChildren(node);
+			auto n = children.size();
+			if (stack.size() >= VariantBuilder::MAX_NESTING_DEPTH) {
+				throw InvalidInputException("VARIANT nesting exceeds maximum of %d", VariantBuilder::MAX_NESTING_DEPTH);
+			}
```
taniabogatsch: Here, we would resize instead. We should have a `NextPowerOfTwo` function. And let's also add a test for that resizing happening. It'll be nice to make full use of this being stack-based now, with allowing arbitrary depth.

taniabogatsch (commented):
Thanks for the PR, left three comments. :)

taniabogatsch (approved):

taniabogatsch:
Thanks for the changes, LGTM now!

PR #24982, opened 2026-08-24
[Serialization][Parquet][Variant] Fix bug related to `ShreddingType` equality operator

Tishj:
Found in https://github.com/duckdb/duckdb/pull/24136

It was previously not found because many variant tests skip the "verify serialization" test config, due to them testing stat-related behavior, which isn't consistent through serialization.


Mytherin:
Thanks!

Issue #10305, opened 2024-01-22
Python client casts boolean to Pandas object type (not boolean)


davetapley:
### What happens?

If I `SELECT` a `boolean` and use `.df()` the series has the `object` type, but Pandas has a [`boolean`](https://pandas.pydata.org/docs/user_guide/boolean.html) type.

I can use [`astype`](https://stackoverflow.com/a/67405428/21115) as a workaround, but it would be nicer to have it be `boolean` by default.

### To Reproduce

I tried various queries and they all result in `object` type:

```py
import duckdb

con = duckdb.connect()
con.sql('select true as a;').df().dtypes
# a    bool

con.sql('select cast (NULL as bool) as a;').df().columns
# a    object

con.execute('create table foo (bar bool);')
con.execute('insert into foo values (NULL);')
con.sql('select bar from foo;').df().dtypes
# bar    object

con.execute('delete from foo;')
con.execute('insert into foo values (True);')
con.sql('select bar from foo;').df().dtypes
# bar    bool

con.execute('drop table foo;')
con.execute('create table foo (bar bool not null);')
con.execute('insert into  foo values (True);')
con.sql('select bar from foo;').df().dtypes
# bar    bool
```

### OS:

Ubuntu 20.04.6 LTS x64

### DuckDB Version:

0.9.3.dev2938

### DuckDB Client:

Python

### Full Name:

Dave Tapley

### Affiliation:

JE Fuller

### Have you tried this on the latest `main` branch?

I have tested with a main build

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [X] Yes, I have

Tishj:
You should take this up with pandas I think, because this is essentially what we do internally:
```py
import duckdb
import pandas

con = duckdb.connect()
res = con.sql('select true as a;').fetchnumpy()
print(res['a'].dtype)
# bool
df = pandas.DataFrame.from_dict(res)
print(df.columns)
# Index(['a'], dtype='object')
```

davetapley:
Okay, this was quite the rabbit hole 🤪 

First, FYI @Tishj my initial post was misleading (I will edit it too, documenting here for clarity). 

As I now understand it:
The `dtype` from `df.columns` is the type of the _index_ (`a`), not the type of the column.

So instead of `print(df.columns)`, it should be:
```py
print(df.dtypes)
# a    bool          <- type of 'a'
# dtype: object        <- type of the series of types
```
[...]

PR #24984, opened 2026-08-24
JSON Deep Merge Fix

lghavami:
(Partial) fix for issue [#10305](https://github.com/duckdblabs/duckdb-internal/issues/10305).

The function json_deep_merge() in the json extension allows for unbounded recursion. If input is too deeply nested this results in stack overflow. Refactored the function to be iterative instead of recursive, so it can now take deeply nested input. Put temporary depth counters for other functions but will refactor those as well in subsequent PRs.


lnkuiper:
Thanks for the PR! I think a related test is failing.

We also typically use `StackChecker` for this, but I'm not sure if that's applicable here. Alternatively, the long-term solution is to eliminate recursion entirely for many JSON functions, keeping stack frames in a `vector` instead. If we do this, we should tackle multiple functions in a unified way, e.g., also tackle `json_transform`.

lghavami:
Hi @lnkuiper, I got rid of recursion entirely for the deep_merge() function. I will do the same where possible for the other functions related to the original issue and make separate PRs for them. I think once that is done we can try to have a more generalized solution for the entire extension. Any feedback is appreciated!

lnkuiper:
Thanks! I think this looks good! It's a bit of shame but some of the `yyjson` functions themselves are recursive, which we can't do much about.

One thing needs to be fixed, then this is good to go from my side:
```sql
  SELECT json_deep_merge('{"a":1,"b":2}', '{"a":null}');
  -- base: {"a":1,"b":2}
  -- PR:   {"b":2,"a":1}
```

lghavami:
@lnkuiper fixed the behavior + added a test for it

PR #24986, opened 2026-08-24
Issue #24449: MINMAX Constant N

hawkfish:
* Move WindowConstantAggregatorGlobalState partitioning logic and data to WindowAggregatorGlobalState
* Pass partitioning mask into WindowSegmentTreeGlobalState
* Use the partitions to skip building of tree nodes that cross partition boundaries (they are never used)
* Add test derived from the issue.


hawkfish:
Tidy failure is in `shredded_vector.cpp`. Not sure how it got committed?

Mytherin:
Thanks!

PR #24995, opened 2026-08-25
Propagate stats for timestamp conversion

dentiny:
Hi team, I found timestamp conversion doesn't prune row groups as expected. For example,
```sql
memory D ATTACH '/tmp/ts_unit_cast_prune.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE ts_ms AS SELECT (TIMESTAMP '2000-01-01' + to_seconds(range))::TIMESTAMP_MS AS ts FROM range(10240);
db D EXPLAIN ANALYZE SELECT count(*) FROM ts_ms WHERE ts > TIMESTAMP '2000-01-01 02:50:00';
╭─ Summary ───────────╮
│ Total Time: 0.0010s │
╰─────────────────────╯
│ Table: db.main.ts_ms                             │
│ CAST(ts AS TIMESTAMP) > '2000-01-01              │
│ 02:50:00'::TIMESTAMP                             │
│ Row Groups Scanned: 5 / 5                        │
│ 39 rows                                      0µs │
```
I think the thing is, on main, stats are not propagated through timestamp unit conversions; only some downcasts can prune via relaxed temporal filter pushdown. This PR changes stats propagation through all plain<->plain and tz<->tz timestamp unit conversions.
Also refactor the conversion code -- old implementation lists all possible enums which is easy to miss.


Mytherin:
Thanks!

PR #24996, opened 2026-08-25
Maintain distinct statistics on ALTER TYPE and ADD COLUMN

pdet:
This PR implements the updates for the distinct statistics for, as these were not updated before for alter type and add column.

Fix: https://github.com/duckdblabs/duckdb-internal/issues/10080


Mytherin:
Thanks!

PR #24997, opened 2026-08-25
Propagate stats for aggregate functions

dentiny:
Hi team, I found for aggregate functions, whose result is always one of (or bounded by) their input values, currently don't propagate any statistics: among all aggregates, only count, sum and bitstring_agg had a statistics callback. This PR adds a shared statistics callback and registers it for:
- Result is one of the input values: min, max, first, last, any_value/arbitrary, arg_min, arg_max, mode, quantile_disc
- Result lies within the input range: median, quantile_cont

The output of these aggregates inherits the min/max statistics of their first argument, which allows the optimizer to fold predicates on aggregate results, for example,
- HAVING clause that is impossible given the input range now folds to an EMPTY_RESULT
- An always-true filters are removed

One thing to notice, avg is intentionally not covered: the rewriter decomposes avg(x) into sum(x) / count(x) before statistics propagation, so a callback on avg would never fire in practice.
Another followup is, this PR only handles cases where aggregate where stats sits within input range, which doesn't handle expression cases, for example, `max(expr(col))`.


src/include/duckdb/function/aggregate_state.hpp
```diff
@@ -161,6 +161,8 @@ struct AggregateStatisticsInput {
 	optional_ptr<FunctionData> bind_data;
 	vector<BaseStatistics> &child_stats;
 	optional_ptr<NodeStatistics> node_stats;
+	//! Whether the aggregate can see zero input rows (e.g. ungrouped over an empty table)
+	bool input_may_be_empty = true;
```
Mytherin: It seems this field is always true - do we need to add this?

Mytherin (commented):

dentiny:
```sh
--- clangd-tidy stdout tail (7 lines) ---
src/common/arrow/arrow_merge_event.cpp:92:28: Error: Use of undeclared identifier 'DEFAULT_ROW_GROUP_SIZE' [undeclared_var_use]
   90 |  		idx_t start_index = transformer.GetIndex();
   91 |  		idx_t end_index = start_index;
   92 |  		while (tuples_for_task < DEFAULT_ROW_GROUP_SIZE) {
      |                             ^~~~~~~~~~~~~~~~~~~~~~
   93 |  			idx_t batch_size;
   94 |  			if (!transformer.TryGetNextBatchSize(batch_size)) {
--- end clangd-tidy stdout tail ---
```
doesn't seem related

src/include/duckdb/function/aggregate_state.hpp
dentiny: Yeah -- that was added in the initial version, but refactored later.
One real issue I found is, additive stats should be reset for aggregate functions' stats propagation.

Mytherin:
Thanks!

PR #24998, opened 2026-08-25
Fix wrong results when constant folding MAX_INT

hannes:
This fixes a bug with constant folding a negation when the constant happened to be MAX_INT. The fix is to bail out of the rewrite. 

Will have to see if there are any regressions popping up as a result.

### DIFF
--- a/src/optimizer/rule/move_constants.cpp
+++ b/src/optimizer/rule/move_constants.cpp
@@ -67,14 +67,8 @@ unique_ptr<Expression> MoveConstantsRule::Apply(LogicalOperator &op, vector<refe
 		}
 		auto result_value = Value::HUGEINT(outer_value).DefaultTryCastAs(constant_type);
 		if (!result_value) {
-			if (comparison.GetExpressionType() != ExpressionType::COMPARE_EQUAL) {
-				return nullptr;
-			}
-			// if the cast is not possible then the comparison is not possible
-			// for example, if we have x + 5 = 3, where x is an unsigned number, we will get x = -2
-			// since this is not possible we can remove the entire branch here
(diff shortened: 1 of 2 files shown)


src/optimizer/rule/move_constants.cpp
```diff
@@ -67,14 +67,8 @@ unique_ptr<Expression> MoveConstantsRule::Apply(LogicalOperator &op, vector<refe
...
 		if (!result_value) {
-			if (comparison.GetExpressionType() != ExpressionType::COMPARE_EQUAL) {
-				return nullptr;
-			}
-			// if the cast is not possible then the comparison is not possible
-			// for example, if we have x + 5 = 3, where x is an unsigned number, we will get x = -2
-			// since this is not possible we can remove the entire branch here
-			return ExpressionRewriter::ConstantOrNull(
-			    std::move(arithmetic.GetChildrenMutable()[arithmetic_child_index]), Value::BOOLEAN(false));
+			// the moved constant is out of range - bail so the addition can still raise its overflow
```
Mytherin: I think this is a bit too aggressively removing optimizations - as a general rule it is fine if optimizations optimize away (overflow) errors (i.e. turning an overflow error into a result, as long as that result would be correct if the value had not overflown, is fine).

Mytherin (commented):

src/optimizer/rule/move_constants.cpp
hannes: Thanks, that is a good point.

Mytherin:
Thanks!

hannes:
also CC @viktorleis

PR #24999, opened 2026-08-25
Prune row groups for length()/char_length() filters via code-point lower bound

### LINKED ISSUE #25001: `length()` doesn't leverage string length stats for row groups containing unicode
### What happens?

#23873 taught `strlen(v) < K` / `> K` filters to prune row groups via `min_string_length` / `max_string_length`. The same statistics could prune `length()` predicates, but today they only help when a partition is ASCII-only: any partition that contains unicode gets a minimum character count of 0/1, so `length(v) < K` scans such groups even when the statistics alone prove every row fails the predicate.

Since VARCHAR values are guaranteed valid UTF-8 and every code point takes at most four UTF-8 bytes, a partition whose shortest string has n bytes cannot contain anything shorter than ceil(n / 4) characters. With `min_string_length = 40`, `length(v) < 10` is false for all rows and the whole group could be dropped from statistics alone.

### To Reproduce

```sql
CREATE TABLE t AS
SELECT CASE WHEN i < 122880 THEN 'é' || repeat('a', i % 10 + 4)
            ELSE repeat('é', 20) END AS v
FROM range(245760) tbl(i);

EXPLAIN ANALYZE SELECT sum(length(v)) FROM t WHERE length(v) < 10;
-- Group B strings are exactly 40 bytes = 20 characters, so no row matches,
-- yet both groups are read:
--   Row Groups Scanned: 2 / 2
```

For contrast, the byte-length predicate on the same table already prunes
correctly since #23873:

```sql
EXPLAIN ANALYZE SELECT sum(strlen(v)) FROM t WHERE strlen(v) < 10;
--   Row Groups Scanned: 1 / 2
```

### OS:

macOS

### DuckDB Version:

6a3a26ffa8

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Junshen Huang

### Affiliation:

PingCAP

- [x] Yes, I have

- [x] Yes, I have

JaySon-Huang:
close https://github.com/duckdb/duckdb/issues/25001

Hi team!

This extends #23873 to code-point functions: filters like `WHERE length(v) < K` can now prune row groups containing unicode strings.

Every code point takes at most four UTF-8 bytes, so a string of n bytes has at least ceil(n / 4) code points. The statistics propagation of `length()` now derives the minimum character count from `min_string_length` using this bound instead of the previous fixed 0/1. The max side already used the byte upper bound and is unchanged, and VARCHAR is guaranteed valid UTF-8, so the bound always holds.

`length_grapheme()` keeps the conservative minimum because grapheme clusters can span arbitrarily many code points. The callback is templated on the counting semantics so the split holds by construction; the ASCII fast-path applies to both. Aliases (`char_length`, `character_length`, `len`) resolve to the same function entry and benefit automatically.

On a table with two unicode row groups where group B holds only 40-byte (= 20-char) strings:

```sql
EXPLAIN ANALYZE SELECT sum(length(v)) FROM t WHERE length(v) < 10;
-- main:  Row Groups Scanned: 2 / 2
-- patch: Row Groups Scanned: 1 / 2
```

Happy to also take the aggregate-side MIN/MAX over length() as a separate follow-up, which #24967 leaves open.

### DIFF
--- a/src/function/scalar/string/length.cpp
+++ b/src/function/scalar/string/length.cpp
@@ -61,6 +61,7 @@ struct BitStringLenOperator {
 	}
 };
 
+template <bool COUNTS_CODEPOINTS>
 unique_ptr<BaseStatistics> LengthPropagateStats(ClientContext &context, FunctionStatisticsInput &input) {
 	auto &child_stats = input.child_stats;
 	auto &expr = input.expr;
@@ -82,8 +83,13 @@ unique_ptr<BaseStatistics> LengthPropagateStats(ClientContext &context, Function
 			// ASCII-only: the character count is exactly the byte count
 			min_length = NumericCast<int64_t>(min_string_length.GetIndex());
 		} else if (min_string_length.GetIndex() > 0) {
-			// non-empty strings contain at least one character
-			min_length = 1;
+			if (COUNTS_CODEPOINTS) {
+				// every codepoint takes at most four bytes
+				min_length = NumericCast<int64_t>((min_string_length.GetIndex() + 3) / 4);
+			} else {
+				// a grapheme cluster can span arbitrarily many codepoints - no better bound exists
+				min_length = 1;
+			}
 		}
 	}
 	auto result = NumericStats::CreateEmpty(expr.GetReturnType());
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks! LGTM

PR #25000, opened 2026-08-25
Determine return type when folding a function call with a typed NULL argument

hannes:
Under `DEFAULT_NULL_HANDLING`, a function call with a `NULL` argument returns `NULL`. For functions with `ANY` return type, the result was an untyped `SQLNULL`. This caused issues. We now call the bind function if needed to figure out the correct type of `NULL` to return.

### DIFF
--- a/src/function/function_binder.cpp
+++ b/src/function/function_binder.cpp
@@ -566,6 +566,46 @@ unique_ptr<Expression> FunctionBinder::BindScalarFunction(const ScalarFunctionCa
 	return BindScalarFunction(func, std::move(arguments), error, is_operator, binder);
 }
 
+//! Returns true if any argument is an untyped NULL, i.e. has type SQLNULL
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks for the PR! @Maxxen can you have a look here given your work on function binding  - my intuition here is that we need to add the `resolve_type` callback to nicely solve this as we've discussed before.

Tishj:
Maybe partially related, this abandoned PR to fix round-tripping of typed NULLs through stringification: https://github.com/duckdb/duckdb/pull/22687

Maxxen:
I think this make sense, we should probably add an assert somewhere that you can't return ANY without also defining a bind callback (or indeed a `resolve_type`).

Maxxen:
We probably need to do the same for window/aggregate functions too though?

Mytherin:
Yes I think a generic `resolve_type` callback seems like a good idea for all function binding

hannes:
so I'll draft this for now and wait for a superseding PR?

PR #25002, opened 2026-08-25
Fix typed NULL value stringification bug that loses type information

Tishj:
This PR is a (partial) revival of #22687


Mytherin:
Thanks!

Tishj:
I wonder if maybe allowing a CAST in the "simple default expression" path check is problematic. Since some casts might still rely on settings, like timezone for TIMESTAMP_TZ casts?

PR #25003, opened 2026-08-25
Optimize zero-length string prefix IN predicates

vvvdwbvvv:
## Summary

  Rewrite zero-length string prefixes to `constant_or_null(input, '')`:

  ```sql
  left(s, 0)
  substring(s, 1, 0)
  substr(s, 1, 0)
```
  This allows downstream predicate and aggregate optimizations to reuse the
  same rewrite, including:

  ```sql
WHERE left(s, 0) IN ('')
  WHERE left(s, 0) = ''
```

  The rewrite preserves NULL semantics:

```sql
  non-NULL input -> ''
  NULL input     -> NULL
```

  ## Performance

  On a 100M-row table with 814 row groups:

  - Before: 1.14s, 814 / 814 row groups scanned
  ```sql
build/reldebug/duckdb -c "
>   PRAGMA threads=1;
>   ATTACH '/tmp/zero_prefix_1m.duckdb' AS db;
> 
>   EXPLAIN ANALYZE
>   SELECT count(*)
>   FROM db.prefixes
>   WHERE left(s, 0) IN ('');
>   "

╭─ Summary ───────────╮
│ Total Time: 0.0158s │
╰─────────────────────╯
╰───────────────┒┎──────────────╯
╭─ Filter ──────┚┖──────────────╮
│ Expression: "left"(s, 0) = '' │
│ 1,000,000 rows         10.0ms │
╰───────────────┒┎──────────────╯
╭─ Table Scan ──┚┖──────────────╮
│ Table: db.main.prefixes       │
│ Projections: s                │
│ Row Groups Scanned: 9 / 9     │
│ 1,000,000 rows            0µs │

```
  - After: 0.0000s, COLUMN_DATA_SCAN
  ```sql
build/reldebug/duckdb -c "
>   PRAGMA threads=1;
>   ATTACH '/tmp/zero_prefix_1m.duckdb' AS db;
> 
>   EXPLAIN ANALYZE
>   SELECT count(*)
>   FROM db.prefixes
>   WHERE left(s, 0) IN ('');
>   "

╭─ Summary ───────────╮
│ Total Time: 0.0000s │
╰─────────────────────╯

```

  The optimized query no longer emits a table-scan operator and can use aggregate
  statistics directly.

  ## Tests

  - zero_length_prefix_stats.test: 23 assertions
  - string_prefix.test: 57 assertions
  - Full optimizer suite: run separately before submitting
  - RelWithDebInfo build passed


Mytherin:
Thanks but this optimization is targeting a much too specific expression. The optimization that should be applied is having `left` or `substr` with a constant 0 length / size be replaced with `ConstantOrNull(Value(""))`.

vvvdwbvvv:
@Mytherin Okay I got this! Also I found that one million case seems not proper for demonstration because it is `:memory`. I'll enclose another way for this case.

src/function/scalar/generic/constant_or_null.cpp
```diff
@@ -70,6 +72,28 @@ static void ConstantOrNullFunction(DataChunk &args, ExpressionState &state, Vect
 	}
 }
 
+static unique_ptr<BaseStatistics> ConstantOrNullStatistics(ClientContext &, FunctionStatisticsInput &input) {
+	if (input.expr.GetReturnType().id() != LogicalTypeId::VARCHAR) {
+		return BaseStatistics::CreateUnknown(input.expr.GetReturnType()).ToUnique();
+	}
+	auto result = StringStats::CreateEmpty(input.expr.GetReturnType());
```
Mytherin: This seems incorrect to me `constant_or_null` should propgate the stats of the underlying constant together with setting `CanHaveNull()` to true.

src/optimizer/rule/string_prefix.cpp
```diff
@@ -21,6 +21,29 @@ static unique_ptr<BoundConstantExpression> RewriteAsNull(const BoundFunctionExpr
...
+		    children[1]->Cast<BoundConstantExpression>().GetValue().GetValue<int64_t>() != 0) {
+			return nullptr;
+		}
+	} else if (function_name == "substring" || function_name == "substr") {
+		if (children.size() != 3 || children[1]->GetExpressionClass() != ExpressionClass::BOUND_CONSTANT ||
+		    children[2]->GetExpressionClass() != ExpressionClass::BOUND_CONSTANT ||
+		    children[1]->Cast<BoundConstantExpression>().GetValue().IsNull() ||
+		    children[2]->Cast<BoundConstantExpression>().GetValue().IsNull() ||
+		    children[2]->Cast<BoundConstantExpression>().GetValue().GetValue<int64_t>() != 0) {
+			return nullptr;
```
Mytherin, line 51: This reads pretty confusingly - can we split this up and annotate with comments so we know what's going on here?

vvvdwbvvv: okay, I'll add

vvvdwbvvv:
Refactored with helper in 650ba7d, this should be better readable.

src/optimizer/rule/string_prefix.cpp
```diff
@@ -71,6 +118,9 @@ unique_ptr<Expression> StringPrefixRule::Apply(LogicalOperator &op, vector<refer
 	auto &func = bindings[1].get().Cast<BoundFunctionExpression>();
 	const auto &constant = bindings[4].get().Cast<BoundConstantExpression>();
 	auto &children = func.GetChildrenMutable();
+	if (auto result = RewriteZeroLength(func)) {
```
Mytherin: This seems like it's added by mistake - isn't this why we have `StringPrefixZeroLengthRule` already?

test/optimizer/zero_length_prefix_stats.test
```diff
@@ -0,0 +1,90 @@
...
+ATTACH '{TEST_DIR}/zero_length_prefix_stats.duckdb' AS db (ROW_GROUP_SIZE 2048);
+
+statement ok
+DETACH tmp;
+
+# An empty LEFT prefix is always in a set containing the empty string
+query II
+EXPLAIN SELECT count(*) FROM db.prefixes WHERE left(s, 0) IN ('');
+----
+physical_plan	<REGEX>:.*UNGROUPED_AGGREGATE.*
```
Mytherin: Why are we checking for `UNGROUPED_AGGREGATE` here?

Mytherin (commented):
Thanks - some more minor comments then this looks good

src/optimizer/rule/string_prefix.cpp
vvvdwbvvv: Umm.. Yes you're right, thank you for correct this.

vvvdwbvvv:
Changes to check the column scan behavior, this is more precise.

vvvdwbvvv:
The patch [6264d6b](https://github.com/duckdb/duckdb/pull/25003/commits/6264d6b08ff999ec100806fb1a27af078856cc5d) skips `zero_length_prefix_stats.test` under `verify_aggregate_state_export` because that config rewrites aggregates as `EXPORT_STATE`.
It intentionally disables aggregate precomputation, so the plan contains a normal `SEQ_SCAN` instead of the test’s expected `COLUMN_DATA_SCAN`. 
I think it is quite complete now, thank you for advising.

vvvdwbvvv:
@Mytherin Hi, I think it is more complete right now, thank you for advices.

Mytherin:
Thanks!

vvvdwbvvv:
I'll check the test, lint doesn't seems to belongs this PR edits.
Seems like it's out-of-scope in this PR

vvvdwbvvv:
@Mytherin Can you help me re-run the queue? Thanks! If there's something that I can help me please feel free tag me.

Mytherin:
Thanks!

vvvdwbvvv:
I'll fix it, thanks! 🫡

dentiny:
> I'll fix it, thanks! 🫡

I'm not sure if this is your issue... maybe the team could help re-add to the merge queue? or maybe @vvvdwbvvv could merge with latest v2.0?

vvvdwbvvv:
> > I'll fix it, thanks! 🫡
> 
> 
> 
> I'm not sure if this is your issue... maybe the team could help re-add to the merge queue? or maybe @vvvdwbvvv could merge with latest v2.0?

yes... I think it may be related to other issue.. 

I think I'd better merge with v2.0? What do you suggest?

vvvdwbvvv:
However, I'll check again. Will be a follow-up once finished, I'll ask for help if needed🙏

PR #25004, opened 2026-08-25
Raise an error on integer division by zero instead of returning NULL

hannes:
Division by zero for `//` and `%` (integers, HUGEINT/UHUGEINT, DECIMAL), `INTERVAL / 0`, and floating point `/` and `%` with `ieee_floating_point_ops=false` now throws an Invalid Input Error that names the failing expression and points at TRY(...) and the new `null_on_division_by_zero` setting, which restores the old NULL result.

The arithmetic simplification rule only rewrites `x // 0` to NULL when that setting is enabled.

Error message:

`Division by zero in expression (1 // 0). Use TRY(...) to return NULL for this expression, or SET null_on_division_by_zero=true to return NULL for all divisions by zero.`


Mytherin:
Thanks! LGTM

hannes:
CC @viktorleis

thomas-0816:
To me, "null_on_division_by_zero=true" is unclear from the naming:

"select 0/0" gives "-nan" instead of "null"

```
memory D SET null_on_division_by_zero=true;
memory D select 0/0;
┌─────────┐
│ (0 / 0) │
│ double  │
├─────────┤
│    -nan │
└─────────┘
```

Mytherin:
That's because of ieee floating point ops being active:

```sql
SET null_on_division_by_zero=true;
SET ieee_floating_point_ops=false;
select 0/0;
┌─────────┐
│ (0 / 0) │
│ double  │
├─────────┤
│    NULL │
└─────────┘
```

But agreed maybe the setting should be inverted to `error_on_division_by_zero`. I'll open a PR.

thomas-0816:
maybe also consider error_on_integer_division_by_zero to have a separation between integer and floating points.

PR #25006, opened 2026-08-25
[JSON] Fix `DECIMAL` value roundtrip bug that was caused by lossy `json_read` behavior

### LINKED ISSUE #24989: JSON conversion round-trip precision loss
### What happens?

See repro for details.

### To Reproduce

```sql
memory D SELECT '883406386745030.3'::JSON::DECIMAL(16, 1);
┌─────────────────────────────────────────────────────────────┐
│ CAST(CAST('883406386745030.3' AS "JSON") AS DECIMAL(16, 1)) │
│                        decimal(16,1)                        │
├─────────────────────────────────────────────────────────────┤
│                      883406386745030.2                      │
│                      (883.41 trillion)                      │
└─────────────────────────────────────────────────────────────┘
```

### OS:

macos

### DuckDB Version:

044a04a7cd39e6e8235f756597ae42dde084e5e5

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

Tishj:
This PR fixes #24989 

Some decimals may look like DOUBLE, so they are read as "number", and lose precision because of that, now all numbers are read as "raw" when the target is DECIMAL.

### Misc

I had to disable the `BIGNUM_AS_RAW` flag when setting `NUM_AS_RAW`, it seems they don't play nicely together and `BIGNUM_AS_RAW` supersedes the `NUM_AS_RAW` when present. Which is interesting because it feels like `NUM_AS_RAW` is just a superset of `BIGNUM_AS_RAW`.

We could also limit the impact of this and exclude smaller width decimals, as those shouldn't be at risk of losing double precision (?)

Should we also add `NUM_AS_RAW` to `JSONReadFlags`?
```c++
enum class JSONReadFlags {
	NONE = 0,
	ALLOW_INVALID_UNICODE = 1 << 0,
	ALLOW_INF_AND_NAN = 1 << 1,
	ALLOW_TRAILING_COMMAS = 1 << 2,
	BIGNUM_AS_RAW = 1 << 3
};
```


lnkuiper:
Thanks!

PR #25007, opened 2026-08-25
[Dev][Parser] Fix `MatchParseResult` regression

Tishj:
This PR relates to https://github.com/duckdblabs/duckdb-internal/issues/10511

Carry the rule reference in the MatchState, instead of adding it after the fact, fixing the regression in `MatchParseResult`.

### DIFF
--- a/src/include/duckdb/parser/peg/matcher.hpp
+++ b/src/include/duckdb/parser/peg/matcher.hpp
@@ -155,7 +155,7 @@ struct MatchState {
 	MatchState(MatchState &state)
 	    : token_iterator(state.token_iterator), suggestions(state.suggestions), allocator(state.allocator),
 	      max_token_index(state.max_token_index), identifier_case_mode(state.identifier_case_mode),
-	      packrat_cache(state.packrat_cache), mode(state.mode) {
+	      packrat_cache(state.packrat_cache), mode(state.mode), rule(state.rule) {
 	}
 
 	TokenIterator token_iterator;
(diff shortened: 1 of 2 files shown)


smvv (approved):
Ah nice find! I think that solves it but i'm not too familiar with the parser to decide.

(I'm looking into a new test config or unittest flag to detect large stack usage in worker threads when running SQL tests in CI. That should help with finding these issues in duckdb core CI without breaking only in musl builds.)

Mytherin:
Thanks!

Tishj:
> Ah nice find! I think that solves it but i'm not too familiar with the parser to decide.
> 
> (I'm looking into a new test config or unittest flag to detect large stack usage in worker threads when running SQL tests in CI. That should help with finding these issues in duckdb core CI without breaking only in musl builds.)

Great idea, I couldn't really produce a test for this currently, sounds like that should make it more testable in the future 👍

Tishj:
I was about to push a patch for the failing CI, but I'll wait until it fails again, seeing as you restarted and marked it as auto-merge already
```patch
diff --git a/src/include/duckdb/parser/peg/matcher.hpp b/src/include/duckdb/parser/peg/matcher.hpp
index fdb63209680..bb244b1a4c9 100644
--- a/src/include/duckdb/parser/peg/matcher.hpp
+++ b/src/include/duckdb/parser/peg/matcher.hpp
@@ -27,6 +27,7 @@ class PEGTransformerFactory;
 class ParseResultAllocator;
 class Matcher;
 class MatcherAllocator;
+struct CompiledGrammarRule;
 
 enum class SuggestionState : uint8_t {
 	SUGGEST_KEYWORD,
diff --git a/test/optimizer/remove_unused_columns.cpp b/test/optimizer/remove_unused_columns.cpp
index 3dc74ed090e..8a231d3c9b6 100644
--- a/test/optimizer/remove_unused_columns.cpp
+++ b/test/optimizer/remove_unused_columns.cpp
@@ -2,6 +2,7 @@
 
 #include "duckdb/execution/column_binding_resolver.hpp"
 #include "duckdb/main/connection.hpp"
+#include "duckdb/main/database.hpp"
 #include "duckdb/optimizer/optimizer.hpp"
 #include "duckdb/optimizer/remove_unused_columns.hpp"
 #include "duckdb/parser/parser.hpp"

```

PR #25008, opened 2026-08-25
Fix ALTER TYPE dropping concurrently committed updates

ywelsch:
The column rewrite resolved values at the altering transaction's snapshot, never locked the parent, and marked it altered only after scanning every row group. Each of those lets an acknowledged update vanish, because the new column does not inherit the old column's update segments:

- read at GetLastCommit() + 1, as index rebuild and vacuum already do
- lock parent.append_lock, as the add column, remove column and add constraint constructors already do; this one locked its own, unreachable mutex
- mark the parent altered before reading rather than after the scan, so a transaction committing meanwhile conflicts instead of being silently dropped

The mark has to be restored if the rewrite throws: a failed cast throws before the catalog entry is swapped, so no undo entry restores it, and the table would reject every modification from then on.

A modification committing during a rewrite that is later rolled back now aborts where it previously succeeded.

### DIFF
--- a/src/storage/data_table.cpp
+++ b/src/storage/data_table.cpp
@@ -186,7 +186,7 @@ DataTable::DataTable(ClientContext &context, DataTable &parent, idx_t changed_id
 	auto &transaction = DuckTransaction::Get(context, db);
 	auto &local_storage = LocalStorage::Get(transaction);
 	// prevent any tuples from being added to the parent
-	lock_guard<mutex> lock(append_lock);
+	lock_guard<mutex> parent_lock(parent.append_lock);
 	for (auto &column_def : parent.column_definitions) {
 		column_definitions.emplace_back(column_def.Copy());
 	}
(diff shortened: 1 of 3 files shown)


Mytherin (commented):
Thanks for the PR! LGTM - I wonder if we also want to add a test with `concurrentloop` to ensure there's no race conditions here that could still cause data to be lost / incorrectly set?

JaySon-Huang:
> Thanks for the PR! LGTM - I wonder if we also want to add a test with `concurrentloop` to ensure there's no race conditions here that could still cause data to be lost / incorrectly set?

Hi @Mytherin @ywelsch ,
On the `concurrentloop` idea: this interleaving can be made fully deterministic with the sync-point hooks from #25019, which fits this race better than a stress loop.

I ran that scenario on current main (before this fix): the UPDATE commits successfully and the final value is `0` — the committed update is silently dropped, reproducibly 3/3 runs. With this PR's fix (marking `ALTERED` before reading, reading at `GetLastCommit()+1`) the same scenario flips to green, so it doubles as the regression guard for exactly this bug.

Mytherin:
Thanks for verifying - interesting idea for sure. Of course the flip side is that by making it deterministic we're only testing a specific order (namely the one specified by the sync point) whereas an actual concurrent scenario is not deterministic, but can also sporadically test other orders (that might still have unsolved bugs). In any case - sounds like this PR is the correct fix for this issue.

JaySon-Huang:
@Mytherin 

Thanks — fully agree, and I think that's exactly where the two approaches complement each other.

A sync point pins one specific order, so it can never cover the whole space of interleavings a real concurrent scenario can produce; the stress-style suites (concurrentloop etc.) are the ones that keep exploring unknown orders, and they should stay and keep growing.

Where the sync point earns its place is the second half of that loop: once a stress test surfaces a flaky failure and we identify the specific interleaving behind it, the sync point turns it into a deterministic regression test — it reproduces the bug reliably on the unpatched tree and stays green once the fix lands, so that same race can't quietly come back through a later refactor. In other words, stress tests find the races, sync points bind the known ones.

ywelsch:
FYI I had started with a generic concurrent test (before opening the PR), but (at least on my machine) it would not get a run where it would fail, so I left it out for the PR. I did got it to fail when injecting a sleep at the right place, but I also felt like that would not make for a great test basis (having to explicitly define these points in the implementation).

JaySon-Huang:
> I did got it to fail when injecting a sleep at the right place, but I also felt like that would not make for a great test basis (having to explicitly define these points in the implementation).

That's exactly the problem I was trying to solve — I don't think injecting sleeps at the right place makes a great test basis either, precisely because the sleep lives inside the implementation with no contract around it.

The sync point is essentially that sleep made explicit: a named hook in the code that only acts when a test enables it. The footprint on production code is minimal — one line next to where you'd otherwise add the sleep. You can see the two examples in #25019: the ALTER TYPE rewrite in `data_table.cpp`, and the partial-precompute index capture in `propagate_aggregate.cpp`. In release builds the hook compiles out entirely, so there's zero effect on production performance.

What it adds over the sleep is a deterministic harness: the same test reproduces the race reliably on the unpatched tree, and flips green once the fix is in — so it becomes a permanent regression guard instead of a one-off investigation.

PR #25010, opened 2026-08-25
Test musl thread stack size in CI on all (slow) tests for release

smvv:
There are linux amd64 and arm64 musl test failures in CI:

- [Aug 22](https://github.com/duckdb/duckdb/actions/runs/32540385326/job/96949958060)
- [Aug 21](https://github.com/duckdb/duckdb/actions/runs/32432696462/job/96629753610).

It also crashes on [Aug 24](https://github.com/duckdb/duckdb/actions/runs/32793666975/job/97642563619#step:11:71), but recovered after the second retry.

The segfault is caused by a worker-thread stack overflow in DuckDB’s PEG parser.

Alpine/musl gives new threads only 128 KiB of stack by default, versus several MiB under glibc. The unittest binary’s PT_GNU_STACK requests no larger size. [musl thread stack documentation](https://wiki.musl-libc.org/functional-differences-from-glibc.html#thread-stack-size)

See issue: https://github.com/duckdblabs/duckdb-internal/issues/10511
And parser stack overflow fix: https://github.com/duckdb/duckdb/pull/25007.

### Minimum stack size computation on linux

The linux release binary’s large static thread local storage (TLS) footprint raises glibc’s actual minimum stack:

- Fixed glibc minimum: 16,384 bytes
- Actual binary minimum: 139,264 bytes
- TLS overhead: 122,880 bytes
- Total configured for 128 KiB usable stack: 253,952 bytes

The unittest binary now calibrates this overhead with `pthread` APIs, adds it to the requested usable stack, and verifies the effective size.

This PR builds on top of https://github.com/duckdb/duckdb/pull/25364 so we can see stack traces for segfaulting tests.

### DIFF
--- a/.github/workflows/Main.yml
+++ b/.github/workflows/Main.yml
@@ -724,7 +724,7 @@ jobs:
         tar -xzf build/release-artifact.tar.gz -C build
 
     - name: Run all release unit tests
-      run: make allunit
+      run: make allunit T="--test-flags='--thread-stack-size 131072'"
 
     - name: Release tests
       shell: bash
--- a/scripts/ci/run_tests.py
+++ b/scripts/ci/run_tests.py
@@ -996,6 +996,10 @@ def parse_failure_info(message: str | None, stdout: str, stderr: str, batch, ret
             if detail_lines:
                 detail_lines.append("")
             detail_lines.extend(preferred_assertion.detail_lines)
+    if not snippet_lines and test_name is not None and line_number is not None:
+        snippet_lines = render_test_snippet(test_name, line_number)
+    if not detail_lines and stderr_info.failing_summary_block:
+        detail_lines.extend(stderr_info.failing_summary_block)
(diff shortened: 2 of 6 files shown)


Mytherin:
Thanks!

Mytherin:
Given that this is more of a work-around for CI rather than a structural fix - I wonder if we want to have a CI run where we explicitly run our test suite with a lower stack limit (e.g. 128KB?) to verify that this still works under that configuration.

Mytherin:
> Given that this is more of a work-around for CI rather than a structural fix - I wonder if we want to have a CI run where we explicitly run our test suite with a lower stack limit (e.g. 128KB?) to verify that this still works under that configuration.

Nevermind that's what this PR does :)

smvv:
Superseded by https://github.com/duckdb/duckdb/pull/25613.

PR #25013, opened 2026-08-25
Incremental message serializer

guillesd:
Adds a `class MessageSerializer : public BinarySerializer` that has the ability to backpatch serialized payloads. This is meant to be used for Quack to patch things like chunk counts or batch index orders after the payloads have been serialized (for efficiency).

The PR also adds some new capabilities to the `generate_serialization.py` script, that now needs to support an incremental writing capability rather than doing one-off serialization. So you can now provide this in the config:

```json
{
  "class": "TestMessage",
  "incremental_writer": true,          <- class a writer
  "members": [
    { "id": 3, "name": "total_items", "type": "idx_t", "backpatch": true }   <- slot to fill
  ]
}
```
Which generates a header demo_writer.hpp:

```cpp
#pragma once

#include "duckdb/common/serializer/message_serializer.hpp"

namespace duckdb {

class TestMessageWriter {
public:
      explicit TestMessageWriter(MessageSerializer &serializer) : serializer(serializer) {
      }

public:
      MessageSerializer::ReservedSlot ReserveTotalItems();

private:
      void CheckFieldOrder(const field_id_t field_id);

private:
      MessageSerializer &serializer;
#ifdef DEBUG
      idx_t last_field_id = DConstants::INVALID_INDEX;
#endif
};

} // namespace duckdb
```

and a serialize_demo.cpp (which was pre-existent but only contained the one-shot serialization).
```cpp
//===----------------------------------------------------------------------===//
// This file is automatically generated by scripts/generate_serialization.py
// Do not edit this file manually, your changes will be overwritten
//===----------------------------------------------------------------------===//

#include "duckdb/common/serializer/serializer.hpp"
#include "duckdb/common/serializer/deserializer.hpp"
#include "demo_writer.hpp"

namespace duckdb {

void TestMessage::Serialize(Serializer &serializer) const {
      serializer.WritePropertyWithDefault<idx_t>(3, "total_items", total_items);
}

unique_ptr<TestMessage> TestMessage::Deserialize(Deserializer &deserializer) {
      auto result = duckdb::unique_ptr<TestMessage>(new TestMessage());
      deserializer.ReadPropertyWithDefault<idx_t>(3, "total_items", result->total_items);
      return result;
}

MessageSerializer::ReservedSlot TestMessageWriter::ReserveTotalItems() {
      CheckFieldOrder(3);
      return serializer.ReserveProperty<idx_t>(3, "total_items");
}

void TestMessageWriter::CheckFieldOrder(const field_id_t field_id) {
#ifdef DEBUG
      if (last_field_id != DConstants::INVALID_INDEX && field_id <= last_field_id) {
              throw InternalException("TestMessageWriter wrote field %d after field %llu, but a message must be written in "
                                      "schema order",
                                      field_id, last_field_id);
      }
      last_field_id = field_id;
#else
      (void)field_id;
#endif
}

} // namespace duckdb
```

There is also a test example under `tests/serialize` that contains also

### The PR also includes
- A encoding utilitie to be able to figure out the width of the slot that is reserved `ReserveVarint(EncodingUtil::MaxLEB128Width<T>(), initial_value);` for a specific type.
- Another two encoded utilities for runtime checks: a) the value fits the reserved slot and b) the utility that encodes the padded LEB128.


guillesd:
I'm fixing CI

guillesd:
jeje killing this PR. Good byee

PR #25014, opened 2026-08-25
[Parquet][Read] Add an optional 'children' field to the `SCHEMA` parameter

Tishj:
This PR supersedes https://github.com/duckdb/duckdb/pull/24407

Intention is the same: this should make it easier to test the MultiFileReader's mapping ability, without requiring going through an extension like iceberg, delta, ducklake

### DIFF
--- a/extension/parquet/parquet_reader.cpp
+++ b/extension/parquet/parquet_reader.cpp
@@ -1258,9 +1258,11 @@ static void VerifyParquetSchemaDefinitionType(const LogicalType &definition_type
 		                      definition_type.ToString());
 	}
 	auto &fields = StructType::GetChildTypes(definition_type);
-	if (fields.size() < 3) {
+	if (fields.size() != 3 && fields.size() != 4) {
 		throw InvalidInputException(
-		    "'schema' expects the STRUCT to have 3 children, 'name', 'type' and 'default_value'");
+		    "'schema' expects the STRUCT to have 3 or 4 fields, 'name', 'type', 'default_value' and optionally "
+		    "'children', not %d",
+		    fields.size());
 	}
 	if (fields[0].first != "name") {
 		throw InvalidInputException("'schema' expects the first field of the struct to be called 'name'");
@@ -1279,7 +1281,7 @@ static void VerifyParquetSchemaDefinitionType(const LogicalType &definition_type
 	if (fields[2].first != "default_value") {
 		throw InvalidInputException("'schema' expects the third field of the struct to be called 'default_value'");
 	}
-	if (fields.size() > 3 && fields[3].first != "children") {
+	if (fields.size() == 4 && fields[3].first != "children") {
 		throw InvalidInputException("'schema' expects the fourth field of the struct to be called 'children'");
 	}
 }
--- a/extension/parquet/parquet_reader.cpp
+++ b/extension/parquet/parquet_reader.cpp
@@ -1330,7 +1330,7 @@ static void VerifyParquetSchemaChildren(const ParquetColumnDefinition &column) {
 		break;
 	}
 	case LogicalTypeId::LIST:
-		VerifyParquetSchemaChildType(column, column.children[0], string(), ListType::GetChildType(column.type), false);
+		VerifyParquetSchemaChildType(column, column.children[0], "element", ListType::GetChildType(column.type), true);
 		break;
 	case LogicalTypeId::MAP:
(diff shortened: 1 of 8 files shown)


src/function/cast/variant/from_variant.cpp
```diff
@@ -512,6 +512,24 @@ static bool CastVariantToJSON(FromVariantConversionData &conversion_data, Vector
 	return true;
 }
 
+static bool CastVariantToSQLNull(FromVariantConversionData &conversion_data, Vector &result, const SelectionVector &sel,
+                                 idx_t offset, idx_t count, optional_idx row) {
+	for (idx_t i = 0; i < count; i++) {
+		auto row_index = row.IsValid() ? row.GetIndex() : i;
+		if (conversion_data.variant.RowIsValid(row_index)) {
```
pdet: ```SELECT [{'b': NULL},NULL]::VARIANT::STRUCT(b "NULL")[];```

src/function/cast/variant/from_variant.cpp
```diff
@@ -512,6 +512,24 @@ static bool CastVariantToJSON(FromVariantConversionData &conversion_data, Vector
 	return true;
 }
 
+static bool CastVariantToSQLNull(FromVariantConversionData &conversion_data, Vector &result, const SelectionVector &sel,
```
pdet: Is this function being tested?

extension/parquet/parquet_reader.cpp
```diff
@@ -1249,13 +1249,112 @@ ParquetOptions::ParquetOptions(ClientContext &context) {
...
+		throw InvalidInputException("'schema' expects the third field of the struct to be called 'default_value'");
+	}
+	if (fields.size() == 4 && fields[3].first != "children") {
+		throw InvalidInputException("'schema' expects the fourth field of the struct to be called 'children'");
+	}
+}
+
+static void VerifyParquetSchemaChildType(const ParquetColumnDefinition &column, const ParquetColumnDefinition &child,
+                                         const string &expected_name, const LogicalType &expected_type,
+                                         bool verify_name) {
```
pdet: is `verify_name` ever false?

extension/parquet/parquet_reader.cpp
```diff
@@ -1249,13 +1249,112 @@ ParquetOptions::ParquetOptions(ClientContext &context) {
...
+	if (verify_name && child.name != expected_name) {
+		throw BinderException("Parquet schema column \"%s\" expects child %s to be named \"%s\", not \"%s\"",
+		                      column.name, column.type.ToString(), expected_name, child.name);
```
pdet: are we printing the correct type?

pdet (commented):
Thanks for the PR, just some minor comments.

src/function/cast/variant/from_variant.cpp
Tishj: Thanks, I actually had a test but forgot to push it:
```patch
+# SQLNULL fields should round-trip without relying on the generic nested-type fallback
+query I
+SELECT cast_to_type(MAP {'x': {'default_value': NULL}}::VARIANT, MAP {'x': {'default_value': NULL}})
+----
+{x={'default_value': NULL}}
```

Or actually, I had trouble reproducing it on `main`

Tishj: Same for yours:
```sql
query I
SELECT [{'b': NULL},NULL]::VARIANT::STRUCT(b "NULL")[];
----
[{'b': NULL}, NULL]
```
This passes

Tishj: Hmm it was using a fallback:
```
		// Fallback: try to convert to Value and then cast
		// This may throw bind-time exceptions for incompatible type combinations
		// In those cases, we should fail gracefully
```

Tishj: Aaaah, it comes down to the MAP<->VARIANT problem (https://github.com/duckdblabs/duckdb-internal/issues/9186). That conversion is lossy.
In the fallback, it tries to convert this VARIANT row into a Value and casting that, but a MAP -> VARIANT -> Value:
```
MAP {'missing': ..., 'source_x': ... }
```
becomes:
```
	'children': [
		{
			'key': missing,
			'value': {
				'default_value': 9,
				'name': missing,
				'type': BIGINT
			}
		},
		{
			'key': source_x,
			'value': {
				'name': renamed_x,
				'type': BIGINT,
				'default_value': NULL
			}
		}
	]
```

pdet:
LGTM!

PR #25016, opened 2026-08-25
Internal #10417: AsOf Hive Partitioning

hawkfish:
* Add benchmark based on synthetic version of Kx Q84 medium scale // 10.
* Performance improvement 1.38s => 0.96s


Mytherin:
Thanks!

PR #25019, opened 2026-08-25
Add test-only sync points + a deterministic repro of the partial precompute race

JaySon-Huang:
Hi team!

I thought the plan-time partial aggregate precomputation in #21831 was really cool — folding aggregates from row group statistics before scanning is the kind of trick that makes analytical queries feel fast — so seeing it switched off in #24962 because nobody could write a deterministic test for its race felt like a shame. DuckDB has no infrastructure for forcing engine-internal interleavings in tests: `concurrentloop` gives you threads but no ordering, and `debug_checkpoint_sleep_ms` only widens a window and hopes. I've hit this exact gap before (rendezvous-style test hooks in another columnar engine), so I'd like to bring that here.

This PR adds `SyncPointCtl`, a named two-way rendezvous exposed as `SYNC_POINT(name)`. Business code calls `SYNC_POINT(name)`; when the point is enabled, the calling thread signals its arrival — making the test-side `WaitAndPause` return — and suspends until the test calls `Next`. Disabling a point (or scope-guard destruction) releases every thread parked on it. Everything is compiled in only when assertions are on (`D_ASSERT_IS_ENABLED`) and expands to nothing in NDEBUG builds, so production binaries and their performance are untouched.

As a first user, a regression test parks a query at the partial-precompute index capture and forces the #24962 interleaving: a checkpoint swaps the row group collection between planning and execution, the stale partition indices mis-map, and the precomputed partition is scanned again. It fails deterministically (3/3 runs) with `5120` instead of `3072` on a tree without the #24962 early return, and passes on current main; the api suite shows no changes outside the new tests (173 test cases / 418014 assertions).

What it covers:

- `SyncPointCtl`/`SYNC_POINT` plus RAII scope guard and a timeout variant, unit-tested (handshake ordering, no-op when disabled, disable-while-waiting, queued arrivals).
- One engine hook at the partial-precompute index capture, with the naming convention documented in the header.
- The deterministic regression test in `test/optimizer` next to `partial_aggregate_precomputation.test`.

If this gives #21831's optimization enough confidence to be switched back on, I'd be very happy.


carlopi:
Thanks for the addition to the test suite, someone will review, I think we are planning a more general fix and then reverting https://github.com/duckdb/duckdb/pull/24962, see comment at https://github.com/duckdb/duckdb/pull/24962#issuecomment-5391730379.

I do have a weird setup that repro the issue, and indeed have not found a way to upstream, I can look at whether your infra would help.

PR #10514, opened 2024-02-07
Export stack trace as part of extra exception info if DUCKDB_DEBUG_STACKTRACE is defined (DEBUG_STACKTRACE cmake variable)

Mytherin:
e.g.:

```sql
D set errors_as_json=true;
D select x;
[...]

PR #25020, opened 2026-08-25
Internal #10514: AsOf Progress

hawkfish:
* Add Progress estimates based on task completion instead of cardinality.


Mytherin:
Thanks!

PR #25026, opened 2026-08-26
Fix mixed path separators for extension/loader in amalgamation (v1.5 backport)

mlafeldt:
'extension/loader' is the only `compile_directories` entry not built with `os.path.join`, so on Windows it yields mixed paths that `build_package()` cannot split into parent directories. The fix: route it through `normalize_path()`.

This is one of the Windows bugs I found and hotfixed while working on https://github.com/duckdb/duckdb-rs/pull/831

Backport of https://github.com/duckdb/duckdb/pull/25025 to v1.5.


Mytherin:
Thanks!

PR #25029, opened 2026-08-26
Use -march=haswell for windows_amd64_mingw (GCC 10 breaks)

smvv:
- Allow `windows_amd64_mingw` to fail for alpha releases.
- Use `-march=haswell` for windows amd64 (GCC 10 does not support `-march=x86-64-v3`).

GCC 10 does [not](https://github.com/duckdb/duckdb/actions/runs/32915454700/job/98030463838#step:16:570) have CPU profile "group names", so the flag `-march=haswell` comes closest to v3 compatibility-wise.


Mytherin:
Thanks!

PR #25031, opened 2026-08-26
Allow integral types in variant_extract via cast to UINTEGER

### LINKED ISSUE #24941: calling `variant_extract( val, intliteral )` throws binder error "'variant_extract' expects the second argument to be of type VARCHAR or UINTEGER, not INTEGER"
### What happens?

Calling `variant_extract( variant_value, integer_literal )` throws 

```
Binder Error:
'variant_extract' expects the second argument to be of type VARCHAR or UINTEGER, not INTEGER
```
Logically the unsigned integer type makes perfect sense. But it's counterintuitive that a plain integer literal (which defaults to plain `integer` not `uinteger`) should be explicitly cast to `uinteger`. In addition, it would be expected that any integer-like expression would also be accepted as long as it can be cast to a `uinteger`

### To Reproduce

```sql
select variant_extract( ['A','B']::variant, 1)
```
throws:
```
Binder Error:
'variant_extract' expects the second argument to be of type VARCHAR or UINTEGER, not INTEGER
```
workaround:
```sql
select variant_extract( ['A','B']::variant, 1::uinteger)
```

### OS:

windows 11

### DuckDB Version:

1.5.5

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Roland Bouman

### Affiliation:

EPAM Systems B.V.

- [x] Yes, I have

- [x] Yes, I have

asxvi:
Fixes #24941

Calling `variant_extract( variant_value, integer_literal)` with non `UINTEGER` but valid `INTEGER` second argument previously threw `Binder Error`.

Fix extends support for integral logical types (`LogicalType::IsIntegral()`) by inserting an explicit cast to `UINTEGER`. All range validation is handled by existing logic.

Tests check support for various second argument values:
- Trivial case: 1
- Valid bound UINTMIN: 0 
- Invalid bound UINTMAX+1: 4294967296
- Invalid bound: -1

### DIFF
--- a/src/function/scalar/variant/variant_extract.cpp
+++ b/src/function/scalar/variant/variant_extract.cpp
@@ -60,9 +60,8 @@ static unique_ptr<FunctionData> VariantExtractBind(BindScalarFunctionInput &inpu
 		throw BinderException("'variant_extract' expects two arguments, VARIANT column and VARCHAR path");
 	}
 	const auto &path = *arguments[1];
-	if (path.GetReturnType().id() != LogicalTypeId::VARCHAR && path.GetReturnType().id() != LogicalTypeId::UINTEGER &&
-	    !path.GetReturnType().IsIntegral()) {
-		throw BinderException("'variant_extract' expects the second argument to be of type VARCHAR or UINTEGER, not %s",
+	if (path.GetReturnType().id() != LogicalTypeId::VARCHAR && !path.GetReturnType().IsIntegral()) {
+		throw BinderException("'variant_extract' expects the second argument to be of type VARCHAR or any integer type, not %s",
 		                      path.GetReturnType().ToString());
 	}
 
@@ -73,7 +72,7 @@ static unique_ptr<FunctionData> VariantExtractBind(BindScalarFunctionInput &inpu
 	} else if (constant_arg.type().IsIntegral()) {
 		return make_uniq<VariantExtractBindData>(constant_arg.GetValue<uint32_t>());
 	} else {
-		throw InternalException("Constant-folded argument was not of type UINTEGER or VARCHAR");
+		throw InternalException("Constant-folded argument was not of type VARCHAR or any integer type");
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/function/scalar/variant/variant_extract.cpp
```diff
@@ -61,8 +62,14 @@ static unique_ptr<FunctionData> VariantExtractBind(BindScalarFunctionInput &inpu
 	}
 	const auto &path = *arguments[1];
 	if (path.GetReturnType().id() != LogicalTypeId::VARCHAR && path.GetReturnType().id() != LogicalTypeId::UINTEGER) {
-		throw BinderException("'variant_extract' expects the second argument to be of type VARCHAR or UINTEGER, not %s",
-		                      path.GetReturnType().ToString());
+		// Accept any integer-ish literal by casting it to UINTEGER explicitly
+		if (path.GetReturnType().IsIntegral()) {
+			arguments[1] = BoundCastExpression::AddCastToType(input.GetClientContext(), std::move(arguments[1]), LogicalType::UINTEGER);
```
Mytherin: I don't think we need to add a cast here - we should be able to run `constant_arg.GetValue<uint32_t>` on the constant for any integral value.

src/function/scalar/variant/variant_extract.cpp
```diff
@@ -60,7 +60,8 @@ static unique_ptr<FunctionData> VariantExtractBind(BindScalarFunctionInput &inpu
 		throw BinderException("'variant_extract' expects two arguments, VARIANT column and VARCHAR path");
 	}
 	const auto &path = *arguments[1];
-	if (path.GetReturnType().id() != LogicalTypeId::VARCHAR && path.GetReturnType().id() != LogicalTypeId::UINTEGER) {
+	if (path.GetReturnType().id() != LogicalTypeId::VARCHAR && path.GetReturnType().id() != LogicalTypeId::UINTEGER &&
```
Mytherin: `path.GetReturnType().id() != LogicalTypeId::UINTEGER` is unnecessary now since it's covered by `IsIntegral` - could you just remove this as well?

asxvi: Made this change as well as slightly modifying error messages to reflect allowance of any integer type.

Mytherin:
Thanks!

PR #25032, opened 2026-08-26
Avoid reconstructing CountZeros lookup tables

p1p1bear:
This PR makes the De Bruijn lookup tables in `CountZeros` static. GCC otherwise copies each 512-byte table to the stack on every call.

In my test, the existing 100M HLL test improved from 2.104s to 0.399s(for GCC 11.2.1). 

This is a small, behavior-preserving change.


Mytherin:
Thanks!

PR #25034, opened 2026-08-26
Adjacent string fixes

Dtenwolde:

The new parser wasn't in line with PostgreSQL / SQL standard with regards to strings separated by a newline. These should be concatenated: 
```sql
memory D SELECT 'foo'
         'bar';
┌──────────┐
│ 'foobar' │
│ varchar  │
├──────────┤
│ foobar   │
└──────────┘
memory D
memory D SELECT 'one' -- comment
         'two';
┌──────────┐
│ 'onetwo' │
│ varchar  │
├──────────┤
│ onetwo   │
└──────────┘
```

Somebody thought this was a good idea to include in the SQL standard...

This PR implements that behavior while keeping same-line adjacent literals invalid. It also ensures dollar-quoted strings don’t concatenate and preserve their original token text and source location for `sql_tokenize` and parser errors: 
```sql
memory D SELECT $tag$a
         $tag$'b';
Parser Error:
syntax error at or near "'b'"

LINE 2: $tag$'b';
             ^^^
```
This follows PostgreSQL in behaviour


src/include/duckdb/parser/peg/matcher_token.hpp
```diff
@@ -25,6 +25,8 @@ struct MatcherToken {
 	idx_t offset = 0;
 	idx_t length = 0;
 	bool unterminated = false;
+	bool preceded_by_newline = false;
+	bool preceded_by_block_comment = false;
```
lnkuiper, line 29: These could be folded into an `enum`, e.g., `PrecedingSeparator::{SAME_LINE, NEWLINE, BLOCK_COMMENT}`

lnkuiper:
Thanks! LGTM. I've left a nitpick comment, but it's not a blocker - added to the merge queue

PR #25036, opened 2026-08-26
FinalizeScan() for MultiFileReader

### LINKED ISSUE #8628: Duckdb 2.0
Cyanoptera release will happen in fall 2026.
We need to prepare for release around a month before it happens.
Many APIs have been made deprecated, and we need to recheck
relative performance.

- [ ] recheck filter pushdown, scalar function pushdown, and aggregate pushdown to check whether we still benefit from it.
- [x] Implement async file reader for Vortex as it's done for Parquet

myrrc:
MultiFileReader lacks a method which would allow to populate another DataChunk after all scanning is finished (for example, custom aggregation machinery needs to write final rows after processing threads exit).


myrrc:
cc @Mytherin as discussed

Mytherin:
Thanks!

PR #25037, opened 2026-08-26
Update httplib & mbedtls

okruitho:
This pull requests bumps httplib from 0.27.1-patched to 0.53.1, and mbedtls from 3.6.4 to 3.6.7.

In addition to that, the update script for mbedtls is fixed to now also copy nested source files.


src/main/http/http_util.cpp
```diff
@@ -57,10 +57,10 @@ string HTTPHeaders::GetHeaderValue(const string &key) const {
 }
 
 #ifndef DUCKDB_DISABLE_BUILTIN_HTTPLIB
-unique_ptr<HTTPResponse> TransformResponse(duckdb_httplib::Result &res) {
+unique_ptr<HTTPResponse> TransformResponse(httplib::Result &res) {
```
carlopi, line 60: I think we should keep aiming for `duckdb_httplib` if at all possible.

carlopi (changes requested):
I would suggest splitting this in 2 independent PRs, since they are independent and can be indepedently evaluated, and would consider whether to keep namespacing to `duckdb_httplib` as is.

In particular we did a few `httplib` changes over time, so it would be handy to have them reapplied.

PR #25040, opened 2026-08-27
Simplify NULLIF(expr, expr) to constant NULL in case simplification rule

### LINKED ISSUE #24413: NULLIF(expr, expr) is not simplified to a constant NULL
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE integers AS
         SELECT range::INTEGER AS i
         FROM range(6144);

-- NULLIF(i, i) is always NULL, so row group could be pruned
memory D EXPLAIN ANALYZE
         SELECT count(*)
         FROM integers
         WHERE NULLIF(i, i) IS NOT NULL;
╭─ Summary ───────────╮
│ Total Time: 0.0037s │
╰─────────────────────╯
│ Table: memory.main.integers                      │
│ CASE WHEN ((i = i)) THEN (NULL) ELSE i END IS    │
│ NOT NULL                                         │
│ Row Groups Scanned: 1 / 1                        │
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

kid-ye:
### Simplify `NULLIF(expr, expr)` to constant NULL

Fixes #24413.

### The bug

`NULLIF(expr, expr)` is always NULL but the optimizer did not simplify it, leaving the expanded `CASE WHEN (expr = expr) THEN NULL ELSE expr END` in the plan. This prevented downstream optimizations like row-group pruning:

```sql
CREATE TABLE integers AS SELECT range::INTEGER AS i FROM range(6144);

EXPLAIN ANALYZE SELECT count(*) FROM integers WHERE NULLIF(i, i) IS NOT NULL;
-- Row Groups Scanned: 1 / 1  (scans all rows despite result always being 0)
```

### Root cause

`NULLIF(a, b)` is a macro that expands to `CASE WHEN a=b THEN NULL ELSE a END`. When both arguments are the same non-volatile expression, the result is always NULL: if `a` is non-NULL then `a=a` is true and the CASE returns NULL; if `a` is NULL then `a=a` is NULL (falsy) and the CASE returns `a`, which is NULL. Either way, NULL.

The existing `CaseSimplificationRule` only folded CASE checks with constant-foldable WHEN expressions. It did not detect this structural pattern.

### The fix

Extended `CaseSimplificationRule` to recognise a single-check CASE where the WHEN is an equality comparison with structurally identical, non-volatile operands and the THEN is a constant NULL. When matched, the entire CASE is replaced with a typed NULL constant.

After the fix:

```sql
EXPLAIN ANALYZE SELECT count(*) FROM integers WHERE NULLIF(i, i) IS NOT NULL;
-- Empty Result  (no scan at all)
```

### Testing

Extended `test/sql/function/generic/test_null_if.test` with:
- `NULLIF(col, col)` returns NULL for every row (including NULL inputs)
- `WHERE NULLIF(col, col) IS NOT NULL` returns 0 rows
- Volatile expressions (`random()`) are not simplified

### DIFF
--- a/src/include/duckdb/optimizer/rule/comparison_simplification.hpp
+++ b/src/include/duckdb/optimizer/rule/comparison_simplification.hpp
@@ -30,4 +30,13 @@ public:
 	                             bool is_root) override;
 };
 
+//! Rewrites expr = expr into constant_or_null(true, expr).
+class SelfComparisonSimplificationRule : public Rule {
+public:
+	explicit SelfComparisonSimplificationRule(ExpressionRewriter &rewriter);
+
+	unique_ptr<Expression> Apply(LogicalOperator &op, vector<reference<Expression>> &bindings, bool &changes_made,
+	                             bool is_root) override;
+};
+
 } // namespace duckdb
--- a/src/optimizer/optimizer.cpp
+++ b/src/optimizer/optimizer.cpp
@@ -78,6 +78,7 @@ Optimizer::Optimizer(Binder &binder, ClientContext &context) : context(context),
 	rewriter.rules.push_back(make_uniq<DateTruncSimplificationRule>(rewriter));
 	rewriter.rules.push_back(make_uniq<ComparisonSimplificationRule>(rewriter));
 	rewriter.rules.push_back(make_uniq<RowComparisonSimplificationRule>(rewriter));
+	rewriter.rules.push_back(make_uniq<SelfComparisonSimplificationRule>(rewriter));
 	rewriter.rules.push_back(make_uniq<InClauseSimplificationRule>(rewriter));
 	rewriter.rules.push_back(make_uniq<InEnumSimplificationRule>(rewriter));
 	rewriter.rules.push_back(make_uniq<EnumCompareSimplificationRule>(rewriter));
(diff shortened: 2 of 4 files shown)


Mytherin:
Thanks but I think this is a bit too specific - maybe we can start off by folding `a=a` into `constant_or_null(true, a)` which is more broadly applicable than this very specific rule?

kid-ye:
Thanks for the feedback! Reworked the approach,  instead of the NULLIF-specific CASE rule, this now folds a = a into constant_or_null(true, a) as a new SelfComparisonSimplificationRule in the comparison simplification rules.

The rule matches COMPARE_EQUAL where both operands are structurally identical (left.Equals(right)) and non-volatile, then rewrites to ExpressionRewriter::ConstantOrNull(child, Value::BOOLEAN(true)).

This is more broadly applicable, any a = a pattern benefits, not just NULLIF. The NULLIF case (CASE WHEN a=a THEN NULL ELSE a END) now gets its WHEN rewritten to constant_or_null(true, a), which is semantically correct (always NULL result). Further CASE folding for constant_or_null could be a follow-up.

PR #25042, opened 2026-08-27
[Dev] Fix `PhysicalIEJoin` out-of-bounds access on `completed`

### LINKED ISSUE #21262: SIGILL in the testcase test/sqlite/test_sqllogictest.cpp
### What happens?

```
/usr/ports/databases/duckdb/work/duckdb-1.5.0/test/sqlite/test_sqllogictest.cpp:212: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGILL - Illegal instruction signal
```

Version: 1.5.0
FreeBSD 15 STABLE

### To Reproduce

run tests

### OS:

FreeBSD 15 STABLE

### DuckDB Version:

1.5.0

### DuckDB Client:

C++

### Hardware:

_No response_

### Full Name:

Yuri Vic

### Affiliation:

n/a

- [x] Yes, I have

- [x] Yes, I have

Tishj:
This PR fixes https://github.com/duckdb/duckdb/issues/21262

`completed` was missing a slot for `IEJoinSourceStage::DONE`.
`IEJoinSourceStage::DONE` was mistakenly seen as meaning the "end of the array".

The only relevant change is:
```c++
array<atomic<idx_t>, static_cast<size_t>(IEJoinSourceStage::DONE) + 1> completed;
```
I was just cleaning up when I stumbled into the cause of the bug.

### DIFF
--- a/src/execution/operator/join/physical_iejoin.cpp
+++ b/src/execution/operator/join/physical_iejoin.cpp
@@ -234,19 +234,19 @@ OperatorResultType PhysicalIEJoin::ExecuteInternal(ExecutionContext &context, Da
 // Source
 //===--------------------------------------------------------------------===//
 enum class IEJoinSourceStage : uint8_t {
-	INIT,
-	SINK_L1,
-	FINALIZE_L1,
-	MATERIALIZE_L1,
-	EXTRACT_LI,
-	SINK_L2,
-	FINALIZE_L2,
-	MATERIALIZE_L2,
-	EXTRACT_P,
-	INNER,
-	OUTER,
-	ANTI,
-	DONE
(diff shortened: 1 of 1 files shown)


pdet:
Can we add a test?

Tishj:
Is that worth the effort?
It should have already been hit by existing tests, it's a simple off-by-1 error that was causing the last index to be out of bounds.
Since we use `std::array`, there is no bounds checking logic to prevent this.

If `completed` should never be indexed with `DONE` then it likely needs a more in-depth fix, but in the mean time I imagine not having it cause an out-of-bounds error is still a desirable fix.

PR #25044, opened 2026-08-27
Fix foreign key side effects in `CREATE TABLE IF NOT EXISTS`

### LINKED ISSUE #25015: CREATE TABLE IF NOT EXISTS adding FOREIGN KEYs to an existing populated table silently corrupts parent-side FK metadata
### What happens?

When `CREATE TABLE IF NOT EXISTS` re-declares an **already-existing, populated** table and the new declaration adds `REFERENCES` clauses that the existing table does not have, the statement is correctly a no-op for the child table — `duckdb_constraints()` still reports 0 foreign keys on it, and `EXPORT DATABASE` emits the original FK-free `CREATE TABLE`.

But the **referenced (parent) tables** are silently given a foreign-key/index entry anyway. From that point on:

* `SELECT` and `INSERT` keep working normally, so nothing looks wrong.
* Any `UPDATE` or `DELETE` that touches a parent's primary key fails with
  `INTERNAL Error: Attempting to dereference an optional pointer that is not set`, raised from `TableIndexList::VerifyForeignKey`, and **invalidates the connection** (all subsequent queries return `FATAL Error: database has been invalidated`).
* The state is **persistent** — it survives reopening the database file.
* The state is **unreachable from SQL**: dropping the child table does not clear it. After `DROP TABLE budget_transactions`, an `UPDATE` on the parent fails with

  ```
  Catalog Error: Table with name budget_transactions does not exist!
  ```

  i.e. the parent still holds a reference to a table that no longer exists.

The only recovery we have found is `EXPORT DATABASE` → fresh file → `IMPORT DATABASE`.

This combination — invisible to every SQL-level check, reads unaffected, writes fatal — makes it hard to detect. We hit it three times in production over three months before identifying the cause, because the only symptom was a background job whose queue quietly stopped shrinking while every dashboard rendered perfectly.

### To Reproduce

```sql
CREATE TABLE budgets(budget_id VARCHAR PRIMARY KEY);
CREATE TABLE transactions(tx_id VARCHAR PRIMARY KEY, value VARCHAR);
CREATE TABLE budget_transactions(
    budget_id VARCHAR,
    tx_id VARCHAR,
    PRIMARY KEY (budget_id, tx_id)
);

INSERT INTO budgets VALUES ('b1');
INSERT INTO transactions VALUES ('t1', 'x');
INSERT INTO budget_transactions VALUES ('b1', 't1');

-- No-op for the child table, but the parents are affected.
CREATE TABLE IF NOT EXISTS budget_transactions(
    budget_id VARCHAR NOT NULL REFERENCES budgets(budget_id),
    tx_id VARCHAR NOT NULL REFERENCES transactions(tx_id),
    PRIMARY KEY (budget_id, tx_id)
);

-- Still reports 0 -- the child really was not modified:
SELECT count(*) FROM duckdb_constraints()
WHERE table_name = 'budget_transactions' AND constraint_type = 'FOREIGN KEY';

-- INTERNAL Error: Attempting to dereference an optional pointer that is not set
UPDATE transactions SET tx_id = tx_id;
```

`DELETE FROM transactions WHERE ...` fails the same way. `INSERT INTO transactions ...` succeeds, which is part of why this hides — an `UPDATE` is a delete+insert, so only the delete half runs FK verification.

Reproduces identically with `:memory:` and with an on-disk database, on both versions tested.

**Expected behaviour:** either the `CREATE TABLE IF NOT EXISTS` is a complete no-op (leaving the parents untouched, matching what it does to the child), or it raises an error because the requested definition conflicts with the existing table. Silently modifying only the parent side leaves the catalog internally inconsistent.

### OS

Ubuntu 24.04.4 LTS, x86_64 (Linux 6.8.0-138-generic)

### DuckDB Version

1.5.5 (current latest release) and 1.4.4 — both affected

### DuckDB Client

Python

### Hardware

_No response_

### Full Name

Rahul-htx

### Affiliation

Personal project

Yes

- [X] Yes, I have

- [X] Yes, I have

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25015

The issue is, when we create a table, before we "return early" because the table already exists, we make updates to parent table; the proposed fix in this PR is straightforward, just to check existence before we make any updates.

The function call chain looks below, should be straightforward 
```sh
PhysicalCreateTable::GetDataInternal
└─ Catalog::CreateTable
   └─ DuckSchemaEntry::CreateTable(child)
      ├─ FindForeignKeyInformation(child)
      ├─ DuckSchemaEntry::Alter(parent, AFT_ADD)
      │  └─ CatalogSet::AlterEntry(parent)
      │     └─ DuckTableEntry::AlterEntry
      │        └─ DuckTableEntry::AddForeignKeyConstraint
      │           ├─ Add FK_TYPE_PRIMARY_KEY_TABLE metadata to parent
      │           └─ Create a new DuckTableEntry sharing the existing storage  <-- parent table FK created already
      │     └─ CatalogSet::map.UpdateEntry(parent)
      │
      └─ AddEntryInternal(child, IGNORE_ON_CONFLICT)  <-- return due to already existence table entries
         └─ CatalogSet::GetEntry(child)
            └─ child already exists → return nullptr
```

### DIFF
--- a/src/catalog/catalog_entry/duck_schema_entry.cpp
+++ b/src/catalog/catalog_entry/duck_schema_entry.cpp
@@ -187,27 +187,13 @@ optional_ptr<CatalogEntry> DuckSchemaEntry::AddEntryInternal(CatalogTransaction 
 }
 
 optional_ptr<CatalogEntry> DuckSchemaEntry::CreateTable(CatalogTransaction transaction, BoundCreateTableInfo &info) {
-	// Check table existence before modifying the catalog.
-	auto &base = info.Base();
-	if (base.on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
-		auto &set = GetCatalogSet(CatalogType::TABLE_ENTRY);
-		if (set.GetEntry(transaction, base.GetTableName())) {
-			return nullptr;
-		}
-	}
-
 	auto table = make_uniq<DuckTableEntry>(catalog, *this, info);
-	auto &dependencies = base.dependencies;
+	auto &dependencies = info.Base().dependencies;
 
-	// add a foreign key constraint in main key table if there is a foreign key constraint
 	vector<unique_ptr<AlterForeignKeyInfo>> fk_arrays;
 	FindForeignKeyInformation(*table, AlterForeignKeyType::AFT_ADD, fk_arrays);
 	for (idx_t i = 0; i < fk_arrays.size(); i++) {
-		// alter primary key table
 		auto &fk_info = *fk_arrays[i];
-		Alter(transaction, fk_info);
-
-		// make a dependency between this table and referenced table
 		auto &set = GetCatalogSet(CatalogType::TABLE_ENTRY);
 		dependencies.AddDependency(*set.GetEntry(transaction, fk_info.GetQualifiedName().Name()));
 	}
@@ -215,9 +201,14 @@ optional_ptr<CatalogEntry> DuckSchemaEntry::CreateTable(CatalogTransaction trans
 		table->dependencies.AddDependency(dep);
 	}
 
-	auto entry = AddEntryInternal(transaction, std::move(table), base.on_conflict, dependencies);
+	auto entry = AddEntryInternal(transaction, std::move(table), info.Base().on_conflict, dependencies);
 	if (!entry) {
 		return nullptr;
+	}
+
+	// add a foreign key constraint in main key table if there is a foreign key constraint
+	for (auto &fk_info : fk_arrays) {
+		Alter(transaction, *fk_info);
 	}
 
 	return entry;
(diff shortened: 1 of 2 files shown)


src/catalog/catalog_entry/duck_schema_entry.cpp
```diff
@@ -187,8 +187,17 @@ optional_ptr<CatalogEntry> DuckSchemaEntry::AddEntryInternal(CatalogTransaction
 }
 
 optional_ptr<CatalogEntry> DuckSchemaEntry::CreateTable(CatalogTransaction transaction, BoundCreateTableInfo &info) {
+	// Check table existence before modifying the catalog.
+	auto &base = info.Base();
+	if (base.on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
+		auto &set = GetCatalogSet(CatalogType::TABLE_ENTRY);
+		if (set.GetEntry(transaction, base.GetTableName())) {
```
Mytherin: Maybe we can just move the explicit adding of dependencies below the `AddEntryInternal` instead of adding a secondary check here? Now we are doing the `IGNORE_ON_CONFLICT` check twice (once here and once in `AddEntryInternal`)

Mytherin:
Thanks!

PR #25045, opened 2026-08-27
Add reusable logical plan column binding verification

kryonix:
Column binding verification is currently coupled to `debug_verify_column_bindings`: `ColumnBindingResolver::Verify` has no result value and reports the first defect through an `InternalException`. The extension-operator branch also deliberately clears type information around `ResolveColumnBindings` to preserve compatibility. That is the earliest point where complete binding-and-type verification becomes impossible, because the verifier no longer has an authoritative extension output schema.

I added `ColumnBindingResolver::VerifyAlways` as a setting-independent entry point while preserving the existing setting-aware API and its callers. The setting-aware wrapper delegates to the same verifier only when enabled and retains its legacy first-failure exception behavior. The unconditional entry point instead returns a `LogicalPlanCompilerResult<LogicalPlanVerificationSuccess>`.

The result represents failures as owned, structured issues. Paths identify operators and expressions by structural child ordinals; construct identities distinguish operators, expressions, functions, types, mismatches, and extensions; and facts carry stable scalar details without relying on diagnostic text. Valid failures are normalized, sorted, and deduplicated by their complete message-independent identity. Explanatory messages remain available, but deliberately do not affect equality or ordering. Malformed producer results remain representable so callers can reject them through `IsValid()` rather than encountering an exception during normalization.

I added a default-false `SupportsTypeBindingVerification()` hook for extension operators together with a stable, non-throwing verification identifier. An opted-in operator is traversed using a fresh input scope followed by its normal children and expressions. Its output is published only when binding and type counts match, every binding component is valid, output bindings are unique, and every type is complete. Invalid ownership indexes and duplicate table-index ownership are also reported structurally. When a child cannot publish a valid output schema, the verifier avoids resolving dependent parents while continuing checks that remain structurally safe elsewhere in the plan.

Existing extensions retain the previous path, including their custom `ResolveColumnBindings` behavior and the clearing of type information before and after it. They therefore acquire no new validation unless they opt in. I also kept ordinary, non-verifying column binding resolution unchanged, including its handling of matching unresolved placeholder types. Expected binding and extension-contract defects become structured issues, while unrelated exceptions from operator implementations are not swallowed or relabeled.

The main trade-off is that extensions receive complete type-aware checking only after explicitly opting in and providing a stable identity and unambiguous output schema. `VerifyAlways` also remains a mutating logical-plan pass because it recomputes operator types; this keeps verification aligned with DuckDB’s existing type-resolution machinery instead of maintaining a second type model. The structured result is more explicit than the previous string-only exception path, but that explicitness is what allows callers to make deterministic decisions without depending on settings or diagnostic wording.


src/execution/column_binding_resolver.cpp
```diff
@@ -249,37 +512,135 @@ unique_ptr<Expression> ColumnBindingResolver::VisitReplace(BoundColumnRefExpress
...
+			seen_indexes.emplace(index, verification_state.GetPath(op));
+		}
+	}
+}
+
+LogicalPlanCompilerResult<LogicalPlanVerificationSuccess> ColumnBindingResolver::VerifyAlways(LogicalOperator &op) {
+	return VerifyAlwaysInternal(op, nullptr);
+}
+
+LogicalPlanCompilerResult<LogicalPlanVerificationSuccess>
```
Mytherin: Let's call this the `LogicalPlanVerifier` given that this is meant as a verification pass and not any sort of transformation to the actual plan

src/execution/column_binding_resolver.cpp
```diff
@@ -249,37 +512,135 @@ unique_ptr<Expression> ColumnBindingResolver::VisitReplace(BoundColumnRefExpress
...
+ColumnBindingResolver::VerifyAlwaysInternal(LogicalOperator &op, optional_ptr<string> first_error) {
+	ColumnBindingVerificationState verification_state(op);
+	ResolveOperatorTypes(op, verification_state);
+	VerifyColumnBindings(op, verification_state);
+
+	unordered_map<TableIndex, LogicalPlanCompilerPath> seen_indexes;
+	VerifyTableIndexes(op, verification_state, seen_indexes);
+	if (!verification_state.issues.empty()) {
```
Mytherin: Let's make this run as part of a test config hidden behind a debug setting, then we can verify this runs over all logical plans the system generates.

src/planner/logical_plan_verification_result.cpp
```diff
@@ -0,0 +1,524 @@
...
+		default:
+			return false;
+		}
+	}
+	default:
+		return false;
+	}
+}
+
+static bool IsStableFactType(const LogicalType &type) {
```
Mytherin, line 375: Maybe have a look at this function

src/include/duckdb/planner/logical_plan_compiler_result.hpp
```diff
@@ -0,0 +1,196 @@
...
+	                                                                           LogicalType actual_type);
+	DUCKDB_API static LogicalPlanCompilerConstructIdentity Extension(string identifier);
+	DUCKDB_API static LogicalPlanCompilerConstructIdentity ExportFeature(string identifier);
+
+	DUCKDB_API bool IsValid() const;
+	DUCKDB_API bool operator==(const LogicalPlanCompilerConstructIdentity &other) const;
+	DUCKDB_API bool operator<(const LogicalPlanCompilerConstructIdentity &other) const;
+};
+
+enum class LogicalPlanCompilerIssueCode : uint8_t {
```
Mytherin: Can we remove the storage qualifier (unless we want to add them to serialization)?

Mytherin (commented):
Thanks! Looks great - some comments

Mytherin:
Thanks!

PR #25049, opened 2026-08-27
Initial implementation of `variant_group_object` processing function

sebastiaan-dev:
This PR resolves https://github.com/duckdblabs/duckdb-internal/issues/10400 by introducing `variant_group_object`, an aggregate that combines key/value pairs from multiple rows into a `VARIANT` object.

Unlike the JSON equivalent, values retain the richer logical types supported by `VARIANT`.

### Examples

```sql
select variant_group_object('duck', 'goose'::VARIANT);
> {'duck': goose}

select variant_group_object(k, v)
from (values
    ('number', 42::VARIANT),
    ('nested', {'a': 1, 'b': 2}::VARIANT),
    ('array', [1, 2, 3]::VARIANT)
) t(k, v);
> {'number': 42, 'nested': {'a': 1, 'b': 2}, 'array': [1, 2, 3]}
```

### Interface

`variant_group_object(key, value)`

#### Parameters

- `key` must be a `VARCHAR` value and becomes an object property name.
- `value` may be any type that can be represented as `VARIANT`.

#### Returns

Returns a `VARIANT` object containing all aggregated key/value pairs.

An empty group returns `NULL`.

#### Behaviour

Similar to `json_group_object` `NULL` keys are not accepted, but `variant_group_object` is more strict wrt duplicate keys and throws when it encounters any.

### Performance

The current implementation is between 5 - 13x slower (1M rows, 1-4 threads) compared to `json_group_object`. When combined with shell rendering it can be a factor 200x times slower for 50k rows. 

When looking at the causes we find 2 major hotspots:

1. `VariantBuilder`: Currently we transform `VARIANT` objects in `ListSegments`, then restore these `VARIANT` objects again in the finalize phase, only to reencode them row-by-row into a single `VARIANT` object using the `VariantBuilder`. Without shell rendering the finalize phase of the aggregation takes up roughly 60% of the runtime, this includes reconstruction from the segments, duplicate-key detection and construction of the result `VARIANT`.
2. `ShellRenderer`: Since the dynamically generated `VARIANT` is not shredded it triggers the fallback path. This path converts to a unified format, then constructs a `STRUCT` type with `VARCHAR` values, which results in many casting and allocation operations.

As optimizations for the above would grow outside the scope of implementing `variant_group_object` I suggest to spin off separate issues for each.


Mytherin:
Thanks!

PR #25051, opened 2026-08-27
Update mbedtls to 3.6.7. Update diff and python script

okruitho:
This pull requests bumps mbedtls from 3.6.4 to 3.6.7.

In addition to that, the update script for mbedtls is fixed to now also copy nested source files.


third_party/mbedtls/inline_mbedtls.diff
```diff
@@ -83,13 +83,174 @@ index 424490951d..3e743b0291 100644
  
      /*
       * Store in result iff the signs are the same (i.e., iff different_sign == false). If
+@@ -1785,33 +1785,33 @@
+     if (I != NULL) {
+         MBEDTLS_MPI_CHK(mbedtls_mpi_grow(I, N->n));
+     }
++    {
```
okruitho, line 90: clang complains about the `goto` and skipping initializing variables when using `MBEDTLS_MPI_CHK` inside of a function. Therefore you need brackets to scope the variables.

carlopi (approved):
Thanks!

PR #25052, opened 2026-08-27
Added duckdb_v2_vector_get_logical_type to the C V2 API

okruitho:
This PR adds an function to the C API for fetching the logical type from a vector; this in turn can be used to verify the logical_type to the expected one.


src/main/capi/v2/capi_v2_vector.cpp
```diff
@@ -60,6 +60,20 @@ DUCKDB_V2_ERROR duckdb_v2_vector_get_vector_type(duckdb_v2_vector_handle vector,
...
+DUCKDB_V2_ERROR duckdb_v2_vector_get_logical_type(duckdb_v2_vector_handle vector,
+                                                  duckdb_v2_logical_type_handle *out_type,
+                                                  duckdb_v2_error_info_handle *err) {
+	return WithErrorHandler(err, [&]() {
+		if (!vector || !out_type) {
+			throw duckdb::InvalidInputException("null argument to duckdb_v2_vector_get_logical_type");
+		}
+		*out_type = nullptr;
+		auto *vec = Convert(vector);
+		auto *lt = new duckdb::LogicalType(vec->GetType());
```
Tishj, line 72: This is an owned copy, not a borrowed reference, right?

Maxxen: Yeah I think the comment needs an update - LogicalTypes are always returned owned.

Maxxen (approved):

PR #20170, opened 2025-12-12
Revert linux toolchain downgrade

hannes:
We used to use GCC 12 instead of the manylinux-default GCC 14 for compatibility reasons. This seems no longer required, and the GCC14 Arm bug with UUID has been resolved in https://github.com/duckdb/duckdb/pull/19318

PR #17963, opened 2025-06-18
Revert toolchain downgrade

samansmink:
This PR aims to revert https://github.com/duckdb/duckdb/pull/17776. I think this PR might not be what we want and would like to propose to revert it. 

I made this draft PR to spark a discussion exploring how to move forward here.

## Background
DuckDB uses the [manylinux](https://github.com/pypa/manylinux) images to build the linux binaries. These binaries are a battle-tested way to build binaries that are binary compatible with many linux distro's. Recently we migrated to the `manylinux_2_28` image for both the `linux_arm64` and `linux_amd64` builds, removing the `manylinux2014`-based `linux_amd64_gcc4` builds. 

The `manylinux_2_28` image is an AlmaLinux 8 based image that uses the a GCC14 based toolchain to produce binaries compatible with GLIBC2.28. This is why it's called `manylinux_2_28`. Binaries built with this image *should* be compatible with distro's that have glibc versions of >=2.28 like:

Debian 10+
Ubuntu 18.10+
Fedora 29+
CentOS/RHEL 8+

[source](https://github.com/pypa/manylinux?tab=readme-ov-file#manylinux_2_28-almalinux-8-based)

## PR #17776
This PR states that `Building with GCC 14 creates binaries incompatible with mainstream Linux distributions using glibc 2.28.`. In this PR The proposed and accepted solution here is to override the default toolchain in the image by downgrading to a lower GCC version. 

To start off, I feel l modifying the `manylinux_2_28` image which is designed to be glibc2.28 compatible to provide glibc2.28  compatibility seems like it should not be necessary and makes me doubt the correctness of the fix.

Another problem I have with this is that this change was not propagates through all DuckDB's CI, meaning that we now compile our libraries and binaries with a different toolchain (GCC12) than we do our extensions (GCC14).  While this might actually be fine in most cases, it could cause subtle and hard to find bugs. We ideally produce all our linux binaries using a consistent toolchain.

Finally, another problem I have with this is that it deviates from the standardised toolchain that we strive to leverage by using the `manylinux_2_28` image. I think our goal should be to stay as close to the original image as possible to keep things as simple as possible.

## How to move forward
There are two paths:
- We accept the fact that the toolchain downgrade is required and propagate it throughout all linux builds/images
- We look for another solution for the issues mentioned and revert the toolchain downgrade

Would love to hear your input here @James-Gilbert- @taniabogatsch!

PR #25053, opened 2026-08-27
Use clang for most CI jobs

smvv:
The docker main profile CI container verifies that compilation of the distributed static library works for `amd64/arm64 x manylinux/alpine` in CI.

The binary archive file is tested on:
- `alpine:3.22` (for musl)
- `debian:12` (for glibc 2.28).

Switching to `clang` solves the GCC-12 dependency problem because clang does not add the undefined `__cxa_call_terminate` symbol (added by GCC 13+).

Fixes https://github.com/duckdblabs/duckdb-internal/issues/8952.
The static archive is verified in: https://github.com/duckdb/duckdb-ci/pull/27.

See also:
- https://github.com/duckdb/duckdb/pull/20170
- https://github.com/duckdb/duckdb/pull/17963

### Future work

This means we can drop the GCC-12 static library CI jobs from duckdb core CI. We still need to create a tarball with static library files, but that likely can happen from the `linux-release` matrix jobs.


Mytherin:
Thanks!

PR #25054, opened 2026-08-27
Optimize ART index scans with position-list materialization

Liuzw1998:
## Problem

Persistent ART scans materialize sorted row IDs through point fetches. When many matches fall in the same physical
vector, compressed columns repeatedly initialize or traverse decoder state even though DuckDB's existing `Scan` and `Select` paths can decode that range once.

For 2,000 consecutive index hits, the four-way comparison is:

| Codec | Baseline ART (ms) | Sequential Scan (ms) | after commit 1 (ms) | after commit 2 (ms) |
| DICT_FSST | 57.95 | 2.60 | 2.43 | 2.44 |
| ZSTD | 135.21 | 1.16 | 1.09 | 1.08 |

## Design

Commit 1 groups sorted persistent ART row IDs into physical-vector windows aligned relative to each RowGroup, then
packs them into output batches containing at most `STANDARD_VECTOR_SIZE` candidates. RowGroup MVCC resolves candidate
visibility before materialization. Each read column uses its existing storage path:

- `Scan` for a fully selected physical window, including a fully selected short tail;
- `Select` when a partially selected complete vector meets the codec's threshold;
- `FetchRows` for other partial windows, including partially selected short tails and mixed-compression columns.

Commit 2 combines consecutive windows that can only use `FetchRows` within the same batch and pinned RowGroup.
Eligibility includes filter columns and uses candidate count as an upper bound on visible count with the existing
codec thresholds. Full windows and windows containing truncated candidates stop a run. Each combined run builds one offset list, checks MVCC once, and appends through the existing `RowGroup::FetchRows`. Other windows retain their per-column Scan/Select/FetchRows paths.

Workers reuse column scan state when the RowGroup and vector offset match. Combined Fetch runs leave that state
associated with its last scanned RowGroup, preserving decoder invalidation before a later Scan/Select. The planned
RowGroup tree is pinned across checkpoint, and the ART vacuum lock keeps row IDs stable. A shared revert lock protects each batch while it rechecks both the live collection boundary and the pinned node boundary. Windows removed from the old tree are skipped even if later appends grow the current tree. Transaction-local scanning, projection and row accounting remain shared with the Fetch path.

Minimum visible counts for `Select` on complete 2,048-row vectors are 3 for fixed-size uncompressed, 1 for RLE, and 2  for ALP, ALPRD, ZSTD and DICT_FSST. BitPacking uses 3 below eight scheduler threads and 1 at eight or more; this uses the scheduler pool size, not the number of workers assigned to the scan. Uncompressed VARCHAR/BIT, LIST/INTERVAL and other unsupported layouts retain `FetchRows` for partial windows. These values should be retuned if streaming-selection work such as [#24925](https://github.com/duckdb/duckdb/pull/24925) changes a codec's `Select` or scan decoder path.

### Official workloads

Official workloads at 32 threads. DELETE includes index maintenance and commit, with sequential reads in every variant.

| Official workload | Baseline ART (ms) | Sequential Scan (ms) | after commit 1 (ms) | after commit 2 (ms) |
| Big Range Query (ART) | 446.78 | 4.09 | 122.01 | 120.99 |
| Point Query (ART) With Many Keys | 26.13 | 2.74 | 10.68 | 9.73 |
| Point Query (ART) | 1.90 | 39.91 | 2.09 | 1.86 |
| IN filter with index scan | 12.88 | 13.19 | 12.84 | 12.41 |
| Delete colocated fetch | 2,014.99 | 2,119.28 | 2,060.12 | 2,021.97 |

The single-point timings are variable and do not establish an incremental gain from commit 2.

### Sparse compressed payloads

These cases select 100,000 rows with a row-ID gap of 1,024:

| Codec | Threads | Baseline ART (ms) | Sequential Scan (ms) | after commit 1 (ms) | after commit 2 (ms) |
| --- | --- | ---: | ---: | ---: | ---: |
| ALP | 1 | 269.27 | 493.90 | 295.41 | 294.99 |
| ALP | 32 | 40.42 | 19.97 | 25.64 | 25.77 |
| ALPRD | 1 | 518.89 | 761.01 | 550.41 | 556.24 |
| ALPRD | 32 | 40.08 | 27.12 | 35.00 | 34.96 |
| BitPacking | 1 | 196.74 | 498.83 | 219.24 | 201.12 |
| BitPacking | 32 | 40.83 | 19.80 | 25.80 | 25.78 |
| DICT_FSST | 1 | 8,412.64 | 3,129.89 | 2,986.56 | 2,976.53 |
| DICT_FSST | 32 | 348.46 | 111.12 | 140.67 | 143.89 |
| RLE | 1 | 237.17 | 366.02 | 162.73 | 168.35 |
| RLE | 32 | 40.82 | 14.71 | 19.34 | 19.37 |
| ZSTD | 1 | 6,752.95 | 4,767.50 | 4,521.60 | 4,515.40 |
| ZSTD | 32 | 282.67 | 153.52 | 195.34 | 196.49 |

At one thread, ALP and ALPRD remain **+9.55% / +7.20%** slower than Baseline ART after commit 2.

### Commit 2: Fetch-only windows

Single-threaded, warm UNCOMPRESSED INTEGER payloads, with one hit per 2,048-row vector and 16 vectors per RowGroup:

| Windows | Projected columns | Baseline ART (ms) | Sequential Scan (ms) | after commit 1 (ms) | after commit 2 (ms) |
| --- | --- | ---: | ---: | ---: | ---: |
| 16 | 1 | 0.69 | 0.78 | 0.72 | 0.70 |
| 16 | 8 | 1.18 | 1.30 | 1.25 | 1.19 |
| 256 | 1 | 0.85 | 2.05 | 0.96 | 0.84 |
| 256 | 8 | 1.79 | 3.87 | 2.18 | 1.80 |
| 4,096 | 1 | 3.11 | 21.61 | 5.51 | 3.34 |
| 4,096 | 8 | 11.04 | 45.40 | 17.79 | 11.07 |

CPU time for 4,096 windows:

| Projected columns | Baseline ART (ms) | Sequential Scan (ms) | after commit 1 (ms) | after commit 2 (ms) |
| 1 | 3.13 | 21.49 | 5.50 | 3.36 |
| 8 | 11.05 | 45.39 | 17.81 | 11.09 |

At 4,096 windows, commit 2 changes wall time by **-39.35% / -37.76%** and CPU time by **-38.93% / -37.73%** versus
commit 1, for one / eight columns. Final wall time remains **+7.37% / +0.34%** versus Baseline ART.


Liuzw1998:
Rebase WIP

artjomPlaunov (changes requested):
Hi, thanks for the PR!

Before a more thorough review, I have two higher-level questions:

1. When is this beneficial with normal planning settings? The largest reported gains use index_scan_percentage = 1.0, yet a table scan is still faster in those examples. What representative workload would choose an index scan under the default settings and benefit substantially from this optimization compared with a table scan? 

I think to justify this optimization I would like to first see the queries/workloads where we can't just use a table scan instead. 

3. For ZTSD; The threshold appears to choose Select when just two rows match in a full vector. Since ZSTD does not appear to have a specialized Select callback, it seems this would materialize the entire vector before keeping those two rows, whereas the existing path fetches them individually. For large strings, could that cause a substantial memory increase or an out-of-memory failure? Is there something in this path that prevents itI’d like to understand these points before reviewing in more detail.

PR #25058, opened 2026-08-27
fix parquet double conversion for decimal with precision over 38

jgmills98:
When reading parquet decimals with precision > 38, duckdb returns the column as double. The conversion walks the value's big-endian two's-complement bytes in 8-byte chunks. The trailing chunk can be shorter than 8 bytes, but it was accumulated with the weight of a full chunk. Writers like arrow use the minimal byte length for the decimal, so it's not guaranteed to be a multiple of 8, causing the column to decode incorrectly.

# Examples
```python
import decimal, pyarrow as pa, pyarrow.parquet as pq, duckdb

vals = [decimal.Decimal("1.50"), decimal.Decimal("-2.25"), decimal.Decimal("123456789.10")]
t = pa.table({"x": pa.array(vals, type=pa.decimal256(39, 2))})
pq.write_table(t, "dec39.parquet")

print(duckdb.sql("SELECT x FROM 'dec39.parquet'").fetchall())
# [(1.080863910568919e+17,), (-1.614090106449586e+17,), (8.895999190193886e+24,)]
# expected approximately [(1.5,), (-2.25,), (123456789.1,)]
```

# Impact
With L = the stored value's byte length (the FLBA `type_length`, or each value's own length for BYTE_ARRAY)
1. values inflate by 2^(8·(8 − L mod 8)); only L divisible by 8 decodes correctly
2. negatives are off by a different amount than positives of the same magnitude, because the sign XOR doesn't cover the trailing chunk's padding (`-0.01` reads back correctly while `-0.02` reads back as `-720575940379279.4`)
3. row-group min/max statistics decode through the same function, so zone maps are corrupted too
4. BYTE_ARRAY-encoded decimals carry a per-value length, so each value gets a different inflation factor, which can invert ordering between values

# Fix
Build each chunk right-justified so the sign XOR covers exactly the values bytes, and scale the accumulator by the chunk's actual byte count instead of always 2^64

# Tests
- 17-byte and 21-byte FLBA files (decimal(39,2) and decimal(50,2)), including `-0.01` and `-0.02`
- a two-row-group file exercising the statistics path through filter counts
- a BYTE_ARRAY decimal file covering per-value lengths and the statistics_propagation regression


jgmills98:
Hi @staticlibs. Think you could help me out with this one? Having some [CI timeout issues](https://github.com/jgmills98/duckdb/actions/runs/33115203442/job/99162716407) on the "Linux Config (Execution)" step of my forked run that I'm not seeing locally. Happy to pass the task over to you if that makes things easier.

image

staticlibs:
@jgmills98 

Hi, thanks for the PR! I got a comment internally that this test was observed to fail on forks. I wonder whether you can undraft the PR to make it run on main CI?

I am not qualified to review the Parquet impl, but this PR looks to me as something we may want for 2.0 (so may need to go in now). I can take it over if it will require technical changes, just lets make a full run first.

jgmills98:
@staticlibs 

Thanks for helping out! I just marked it as ready for review. I don think CI will automatically start for me as this is my first pr to duckdb proper.

jgmills98:
Hi @lnkuiper, apologies for not following up on this sooner - things got busy on my end. Is this still something the team would like out for 2.0? This is my first pr and I had troubles with local CI that seem unrelated. What should the next steps be? Happy to pass the task over to you if that makes things easier.

PR #25061, opened 2026-08-28
Prune numeric comparison for NOTEQUAL operation

dentiny:
Hi team, I found certain NOTEQUAL / DISTINCT_FROM operations don't prune row group as expected for numeric and integer types, for example
```sql
memory D CREATE TABLE db.numeric_values AS
         SELECT (CASE range // 2048 WHEN 0 THEN 1 WHEN 1 THEN 1 ELSE 2 END) AS a,
                (CASE range // 2048 WHEN 0 THEN 1 WHEN 1 THEN 2 ELSE 2 END) AS b
         FROM range(6144);
memory D EXPLAIN ANALYZE SELECT sum(a) FROM db.numeric_values WHERE a != b;
╭─ Summary ───────────╮
│ Total Time: 0.0013s │
╰─────────────────────╯
╰────────────────┰────────────────╯
╭─ Projection ───┸────────────────╮
│ Projections: a                  │
│ 2,048 rows                  0µs │
╰────────────────┰────────────────╯
╭─ Projection ───┸────────────────╮
│ Projections: #0                 │
│ 2,048 rows                  0µs │
╰────────────────┰────────────────╯
╭─ Filter ───────┸────────────────╮
│ Expression: a != b              │
│ 2,048 rows                  0µs │
│ Table: db.main.numeric_values   │
│ Projections: a, b               │
│ 6,144 rows                  0µs │
```
In the above example, a and b columns only differ in the third row group, so only one row group is expected to access.

The root cause of the issue is, we didn't implement the push down for these operations :)


src/optimizer/filter_combiner.cpp
```diff
@@ -411,6 +411,7 @@ FilterPushdownResult FilterCombiner::TryPushdownGenericExpression(LogicalGet &ge
 		const auto &right = BoundComparisonExpression::Right(comparison);
 		const auto comparison_type = comparison.GetExpressionType();
 		const bool supported_comparison = comparison_type == ExpressionType::COMPARE_EQUAL ||
+		                                  comparison_type == ExpressionType::COMPARE_NOTEQUAL ||
```
dentiny, line 414: There're two other possible operations, DISTINCT and NOT DISTINCT, which are seem to related to subquery decorrelation, which I haven't figured it out :-(

Mytherin:
Thanks!

PR #25063, opened 2026-08-28
Prune row group on FALSE_OR_NULL

dentiny:
Hi team, I found for `list_contains` function, row group pruning doesn't always work as expected, for example,
```sql
memory D ATTACH '/tmp/list_contains_row_group_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE nullable_lists AS
     SELECT CASE WHEN i % 100 = 0 THEN NULL ELSE [i] END AS l
     FROM range(6144) tbl(i);
db D EXPLAIN ANALYZE SELECT count(*) FROM nullable_lists WHERE list_contains(l, 3001);
╭─ Summary ───────────╮
│ Total Time: 0.0016s │
╰─────────────────────╯
│ Table: db.main.nullable_lists   │
│ Filters: list_contains(l, 3001) │
```
Here we should only access one single row group, but somehow all of them are accessed.

I think the reason is, currently we only prune for [`ALWAYS_FALSE`](https://github.com/duckdb/duckdb/blob/5b1ef771b29aa1bf7d1013028cd622352e5adeda/src/storage/table/row_group.cpp#L763-L765), which doesn't consider NULL cases.

The same prune result is applied [earlier in the same function](https://github.com/duckdb/duckdb/blob/5b1ef771b29aa1bf7d1013028cd622352e5adeda/src/storage/table/row_group.cpp#L742-L745).

### DIFF
--- a/src/storage/table/row_group.cpp
+++ b/src/storage/table/row_group.cpp
@@ -760,7 +760,8 @@ bool RowGroup::CheckZonemap(optional_ptr<ClientContext> context, ScanFilterInfo 
 		} else {
 			prune_result = GetColumn(base_column_index).CheckZonemap(context, base_column_index, filter);
 		}
-		if (prune_result == FilterPropagateResult::FILTER_ALWAYS_FALSE) {
+		if (prune_result == FilterPropagateResult::FILTER_ALWAYS_FALSE ||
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25064, opened 2026-08-28
Prune segment on FALSE_OR_NULL

dentiny:
Hi team, similar to https://github.com/duckdb/duckdb/pull/25063, I think we could prune segments more properly.
For example, 
```sql
memory D CREATE TABLE integers AS
         SELECT CASE WHEN range % 2 = 0 THEN range::INTEGER ELSE NULL END AS x
         FROM range(100000);
memory D PRAGMA enable_profiling = 'json';
memory D SET tracked_metrics = ['query.total_rows_scanned'];
memory D SELECT count(*) FROM integers WHERE x > 100000 - x;
{
    "query": {
        "total_rows_scanned": 100000
    }
}
```
The query scans all rows and segments unexpectedly, but only later segments should be accessed.

The reason is, when pruning segments, either FALSE or NULL should be used to prune, but we only check [ALWAYS_FALSE](https://github.com/duckdb/duckdb/blob/5b1ef771b29aa1bf7d1013028cd622352e5adeda/src/storage/table/row_group.cpp#L793-L795) for now.

### DIFF
--- a/src/storage/table/row_group.cpp
+++ b/src/storage/table/row_group.cpp
@@ -790,7 +790,8 @@ bool RowGroup::CheckZonemapSegments(CollectionScanState &state) {
 
 		optional_ptr<SegmentNode<ColumnSegment>> current_segment;
 		auto prune_result = column_data.CheckZonemap(state.column_scans[column_idx], filter, current_segment);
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks!

PR #25067, opened 2026-08-28
Merge static libs jobs into Linux, macOS and Windows release CI jobs

smvv:
Follow-up of https://github.com/duckdb/duckdb/pull/25053.

We drop the GCC-12 `static-libs-*` CI jobs from duckdb core CI.

We still need to upload the static library files, but that can happen from the `linux-release(-cli)` matrix jobs.

Similarly, we can builds static libs in macOS and Windows CI jobs.


Mytherin:
Thanks!

PR #25068, opened 2026-08-28
Add support for `duckdb -serve` and `duckdb -connect` to the CLI

Mytherin:
This PR adds the ability to quickly spin up and connect to a `quack` server to the CLI using the `-serve` and `-connect` CLI parameters, e.g.:

```
$ duckdb -serve
┌──────────────────────┬───────────────────────┬────────────┐
│      listen_uri      │      listen_url       │ auth_token │
│       varchar        │        varchar        │  varchar   │
├──────────────────────┼───────────────────────┼────────────┤
│ quack:localhost:9494 │ http://localhost:9494 │ quack      │
└──────────────────────┴───────────────────────┴────────────┘
```

```
$ duckdb -connect                    
quack:localhost memory D 
```

This is syntactic sugar around `quack`. `-serve` calls the `.serve_command` on start-up:

```sql
CALL quack_serve({serve_secret|})
```

`-connect` calls the `.connect_command` on start-up:

```sql
CONNECT 'quack:'{connect_secret|}
```

These commands can be customized using [quack secrets](https://github.com/duckdb/duckdb-quack/pull/262). By default the unnamed secret is used. If no secret is specified, a default connection is setup (`localhost:9494` with token `quack`). Other secrets can be specified using the `-secret <named_secret>` parameter, e.g.:

```sql
CREATE PERSISTENT SECRET other_config(TYPE quack, TOKEN 'super_secret');
```

```
duckdb -serve quack:other_config
┌─────────────────┬───────────────────────┬──────────────┐
│   listen_uri    │      listen_url       │  auth_token  │
│     varchar     │        varchar        │   varchar    │
├─────────────────┼───────────────────────┼──────────────┤
│ quack:localhost │ http://localhost:9494 │ super_secret │
└─────────────────┴───────────────────────┴──────────────┘
```

```
duckdb -connect quack:other_config
```


carlopi (commented):
I think this is great, I don't like the coupling that the default secret has to be specified in duckdb/duckdb codebase. I need to think for a second, can also be refactored later.

I would maybe default secret name to be quack provide, and have then: `CONNECT 'quack_secret:';` that when empty becomes the default secret that is quack provided.

Also I think I would look into a way to avoid default token, possibly via convention that that will create a persistent secret with a given name. Let me think some more.

Mytherin:
The reason it's there is that by default `quack` generates a random token, which means `duckdb -serve` and `duckdb -connect` can't work symmetrically unless we specify the token through some other mechanism. If you have a better idea happy to hear about it.

carlopi:
Chatted on a side, then more thoughts:
1. considering adding:
```sh
duckdb -connect       # defaults to quack

duckdb -connect quack     # use default quack secret
duckdb -connect postgres      # use default postgres secret
```
but then why not remove `-secret` entirely? Like:
```sh
duckdb -connect quack:my_secret      #use `my_secret`, and check it has to be quack
```

2. and then same for serve:
```
duckdb -serve            # use default quack secret
duckdb -serve quack     # use default quack secret
duckdb -serve quack:my_secret      #use `my_secret`, and check it has to be quack
```
and make providing a file implicitly forbidden in both cases?

2 needs depends on a flag like `--file <file_name>` to be explicit

PR #25076, opened 2026-08-28
Validate block allocation sizes and validity masks

HendrikLambert:
A corrupt database header could specify a block allocation size that underflows the usable block size or causes an excessive allocation. Corrupt row-version metadata could also contain validity-mask indices that write past the mask.

Validate both before use and report a `DataCorruptionException` for invalid values.


Mytherin:
Thanks!

PR #25077, opened 2026-08-28
Expand on #25006, type recursion is also needed

carlopi:
Expand on https://github.com/duckdb/duckdb/pull/25006, via fixing also cases where decimal might be within a struct / map / list.


Tishj:
Thanks, that's an oversight on my part, can you add a test to `test/sql/json/test_json_cast.test` for this though?

Mytherin:
Thanks!

PR #25079, opened 2026-08-28
Prune row group for `list_position` function

dentiny:
Hi team, I found `list_position` is not properly pruned as `list_contains`, for example,
```sql
memory D ATTACH '/tmp/list_contains_row_group_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE lists AS
     SELECT [i] AS l
     FROM range(6144) tbl(i);
db D EXPLAIN ANALYZE SELECT count(*) FROM lists WHERE list_position(l, 3000) IS NOT NULL;
╭─ Summary ───────────╮
│ Total Time: 0.0035s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ──────────────╮
│ Aggregates: count_star()           │
│ 1 row                          0µs │
╰──────────────────┒┎────────────────╯
╭─ Table Scan ─────┚┖────────────────╮
│ Table: db.main.lists               │
│ Type: Sequential Scan              │
│ Filters:                           │
│ list_position(l, 3000) IS NOT NULL │
│ Row Groups Scanned: 3 / 3          │
│ 1 row                          0µs │
```
The above query should only access one row group, but in reality we access three all.
The reason is `list_position` function doesn't hook with stats callback unfortunately; this PR reuses the stats propagation function with `list_contains`.

### DIFF
--- a/src/function/scalar/list/contains_or_position.cpp
+++ b/src/function/scalar/list/contains_or_position.cpp
@@ -2,7 +2,9 @@
 #include "duckdb/function/scalar/nested_functions.hpp"
 #include "duckdb/function/scalar/list/contains_or_position.hpp"
 #include "duckdb/optimizer/statistics_propagator.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
 #include "duckdb/storage/statistics/list_stats.hpp"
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25080, opened 2026-08-28
tighten rclone ext push, avoid pre-PUT HEAD call

benfleis:
Delta and Azure ext distribution pushes are failing, appearing to be a
failed HEAD call (which is disallowed by the token). Add extra cli arg
to skip the HEAD call altogether.


Mytherin:
Thanks!

PR #25087, opened 2026-08-28
Support more types for multicolumn pushdown

dentiny:
Hi team, I found for multi-column comparison, currently we only support [numeric types](https://github.com/duckdb/duckdb/blob/25a14295a609678ac2e12e91c8bf39d24ca3c260/src/optimizer/filter_combiner.cpp#L420-L421), which leads to missing row group pruning.
```sql
memory D CREATE TABLE db.temporal_values AS
         SELECT ('2000-01-01'::date + (CASE range // 2048 WHEN 0 THEN 1 WHEN 1 THEN 1 ELSE 2 END) * INTERVAL '1 day') AS a,
                ('2000-01-01'::date + (CASE range // 2048 WHEN 0 THEN 1 WHEN 1 THEN 2 ELSE 2 END) * INTERVAL '1 day') AS b
         FROM range(6144);
memory D EXPLAIN ANALYZE SELECT count(*) FROM db.temporal_values WHERE a != b;
╭─ Summary ───────────╮
│ Total Time: 0.0017s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ──────────╮
│ Aggregates: count_star()       │
│ 1 row                      0µs │
╰────────────────┰───────────────╯
╭─ Filter ───────┸───────────────╮
│ Expression: a != b             │
│ 2,048 rows                 0µs │
╰────────────────┒┎──────────────╯
╭─ Table Scan ───┚┖──────────────╮
│ Table: db.main.temporal_values │
│ Type: Sequential Scan          │
│ Projections: a, b              │
│ Row Groups Scanned: 3 / 3      │
│ 6,144 rows                 0µs │
╰────────────────────────────────╯
```
In the above query we should only access one row group but in reality all three are read.

This PR adds more type support to cover temporal types and boolean, but interval type is not covered -- it needs special normalization handle.


src/optimizer/filter_combiner.cpp
```diff
@@ -343,6 +343,10 @@ static bool TypeSupportsConstantFilter(const LogicalType &type) {
 	return false;
 }
 
+static bool TypeSupportsMultiColumnComparison(const LogicalType &type) {
+	return type.IsNumeric() || type.IsTemporal() || type.id() == LogicalTypeId::BOOLEAN;
```
Mytherin: Should we look at the internal type here, instead of at numeric / temporal / boolean?

Mytherin:
Thanks!

dentiny:
Appreciate a re-merge, PR dropped due to merge conflict :)

PR #25089, opened 2026-08-28
FileSystem: Add optional async TryStartRead (disabled by default)

carlopi:
This adds some machinery, currently disabled, that once armed could unlock async IO to be driven by external (e.g. application) owned threads. This would also unlock, together with an independent change to HTTPUtil, to experiment with moving towards `multi_curl` and async I/O to land for single threaded Wasm (delegating to JS event loop `fetch` the eventual completion).

Main changes are the following:
1. Add `FileSystem::TryStartRead`, that default to `return false;`
2. Use that in `TryExecuteAsync`:
```c++
	AsyncTaskExecutionResult TryExecuteAsync(AsyncIOCallback on_complete) final {
		auto request = PrepareRead();
		if (request.handle.TryStartRead(request.nr_bytes, request.location, request.destination,
		                                std::move(on_complete))) {
			return AsyncTaskExecutionResult::PENDING;
		}
		// this file system reads synchronously - do the read here, holding the calling thread
		PerformRead(request);
		FinishRead();
		return AsyncTaskExecutionResult::FINISHED;
	}
```
3. Enabled in the parquet read path
4. Expand the test `debug_fs` to experiment / stress test this

Note that as long as `TryStartRead` defaults to `return false;`, this should be provably just shuffling code around, but enable to restructure the code to have non-duckdb owned threads drive FileSystem work, at least for the parquet path.

### DIFF
--- a/src/include/duckdb/common/file_system.hpp
+++ b/src/include/duckdb/common/file_system.hpp
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include "duckdb/common/async_io_callback.hpp"
 #include "duckdb/common/common.hpp"
 #include "duckdb/common/enums/file_compression_type.hpp"
 #include "duckdb/common/enums/file_glob_options.hpp"
@@ -128,6 +129,8 @@ public:
 	// File offset will not be changed.
 	DUCKDB_API void Read(void *buffer, idx_t nr_bytes, idx_t location);
 	DUCKDB_API void Read(QueryContext context, void *buffer, idx_t nr_bytes, idx_t location);
+	//! Try to start an asynchronous read, see FileSystem::TryStartRead
+	DUCKDB_API bool TryStartRead(void *buffer, idx_t nr_bytes, idx_t location, AsyncIOCallback callback);
 	DUCKDB_API void Write(QueryContext context, void *buffer, idx_t nr_bytes, idx_t location);
 	DUCKDB_API void Seek(idx_t location);
 	DUCKDB_API void Reset();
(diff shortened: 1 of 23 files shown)


src/storage/external_file_cache/caching_file_system.cpp
```diff
@@ -339,17 +350,41 @@ Allocator &CachingFileHandle::GetBufferAllocator() const {
 	return external_file_cache.GetBufferManager().GetBufferAllocator();
 }
 
+bool CachingFileHandle::UsesUncachedReadPath() {
+	return !external_file_cache.IsEnabled() || !external_file_cache.ShouldCacheFile(path.path) || !CanUseCache();
+}
+
+bool CachingFileHandle::TryStartRead(const idx_t nr_bytes, const idx_t location, FileBufferHandleGroup &out_group,
```
lnkuiper, line 357: Should this function take the query context? I think this is now regressing the `bytes_read` metric. Before, it called the `Read` function that took it, and now it calls `TryStartRead` which doesn't.

src/include/duckdb/common/file_system.hpp
```diff
@@ -211,6 +214,12 @@ class FileSystem {
 	//! Read exactly nr_bytes from the specified location in the file. Fails if nr_bytes could not be read. This is
 	//! equivalent to calling SetFilePointer(location) followed by calling Read().
 	DUCKDB_API virtual void Read(FileHandle &handle, void *buffer, int64_t nr_bytes, idx_t location);
+	//! Try to start an asynchronous read of exactly nr_bytes from [location] into [buffer], releasing the calling
+	//! thread while the read is in flight. Returns false when this file system cannot read asynchronously, in which
+	//! case the caller must fall back to Read(). When it returns true [callback] is invoked exactly once, and
+	//! [handle] and [buffer] must stay alive until it fires.
+	DUCKDB_API virtual bool TryStartRead(FileHandle &handle, void *buffer, int64_t nr_bytes, idx_t location,
+	                                     AsyncIOCallback callback);
```
lnkuiper, line 222: I'm worried that the contract of who owns the handle/buffer is not clear enough and this will easily lead to use-after-free. Maybe something like this would be less of a footgun:
```C++
  class FileReadDestination {
  public:
  	virtual ~FileReadDestination() = default;
  	virtual data_ptr_t Data() = 0;
  	virtual idx_t Size() const = 0;
  };

  struct FileReadRequest {
  	shared_ptr<FileHandle> handle;
  	shared_ptr<FileReadDestination> destination;
  	idx_t location;
  };

  class PendingFileRead {
  public:
  	virtual ~PendingFileRead() = default;
  	virtual void Cancel() = 0;
  	virtual void Wait() = 0;
  };

  struct FileReadSubmission {
  	FileReadSubmissionState state; // UNSUPPORTED, COMPLETED, PENDING
  	unique_ptr<PendingFileRead> operation;
  };

  virtual FileReadSubmission
  TryStartRead(shared_ptr<const FileReadRequest> request,
               AsyncIOCallback callback);
```

lnkuiper (commented):
Thanks! I've left two comments. Also let's tone down the comments, it's verbose.

PR #25092, opened 2026-08-28
HTTPUtil: Add support for optional async requests

carlopi:
Refactor `HTTPUtil` to allow override of `HTTPUtil::Wait`, and restructure code around retries.
Add a new function, `Send`, that is currently always implemented as `SendRequest(...); callback(...);`, but allow whoever overrides `HTTPUtil` to provide actual externally driven async network requests.

This would, together with the connected, but independent PR at https://github.com/duckdb/duckdb/pull/25089, to allow embedders such as duckdb-wasm to experiment with externally driven (in this case the JS event loop) proper async network requests to work also in the single threaded case.

Plan would be to add this callbacks, allowing extensions to experiment with this (at their risk!), default path for `httpfs` provided stack will stay syncronous.

### DIFF
--- a/src/include/duckdb/common/http_util.hpp
+++ b/src/include/duckdb/common/http_util.hpp
@@ -316,7 +316,7 @@ enum class HTTPRequestState : uint8_t {
 
 //! Invoked exactly once when a request completes, unless the call that started it threw.
 //! For a GET with a content_handler the body has already been streamed, so [response] carries status and headers.
-using HTTPResponseCallback = std::function<void(unique_ptr<HTTPResponse> response, optional_ptr<ErrorData> error)>;
+using HTTPResponseCallback = std::function<void(unique_ptr<HTTPResponse> response, ErrorData error)>;
 
 class HTTPClient {
 public:
--- a/src/include/duckdb/common/http_util.hpp
+++ b/src/include/duckdb/common/http_util.hpp
@@ -383,12 +383,17 @@ enum class HTTPRetryDecision : uint8_t {
 struct HTTPAttempt {
 	//! The response, null when the attempt threw
 	unique_ptr<HTTPResponse> response;
-	//! The exception the attempt threw, if any
+	//! The exception the attempt threw, if any. Only set when the attempt threw on this thread.
 	std::exception_ptr caught_e = nullptr;
+	//! The error the attempt reported, kept whole so a failure keeps its type when it is rethrown
+	ErrorData error;
 	string exception_error;
 	//! Status and Retry-After recovered from an HTTPException, used for throttle detection
 	string caught_status;
 	string caught_retry_after;
+
+	//! Record [error_p] as this attempt's failure, recovering the status and Retry-After it carries
+	DUCKDB_API void SetError(ErrorData error_p);
 };
 
 //! The retry policy of one HTTP request, carried across its attempts. It does no waiting of its own:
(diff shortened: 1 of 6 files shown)


src/main/http/http_util.cpp
```diff
@@ -163,6 +163,127 @@ bool HTTPUtil::ShouldRetry(const BaseRequest &request, const HTTPResponse &respo
...
+public:
+	AsyncRetryDriver(HTTPUtil &http_util, BaseRequest &request, unique_ptr<HTTPClient> &client,
+	                 HTTPResponseCallback on_complete)
+	    : http_util(http_util), request(request), client(client), on_complete(std::move(on_complete)) {
+	}
+
+	//! Issue the request, reporting whether everything resolved before returning
+	HTTPRequestState Start() {
+		Attempt();
+		return delivered ? HTTPRequestState::COMPLETED : HTTPRequestState::PENDING;
```
lnkuiper, line 190: This is a data race when completion comes from another thread: the callback writes `delivered` while `Start()` reads it here. Could we use the same atomic state handshake as #25089 and test it?

src/main/http/http_util.cpp
```diff
@@ -163,6 +163,127 @@ bool HTTPUtil::ShouldRetry(const BaseRequest &request, const HTTPResponse &respo
...
+	void Attempt() {
+		auto self = shared_from_this();
+		// the returned state is not needed: an attempt that finished inline has already called back
+		client->Send(request, HTTPExecutionMode::DEFERRABLE,
+		             [self](unique_ptr<HTTPResponse> response, optional_ptr<ErrorData> error) {
+			             self->OnAttemptDone(std::move(response), error);
+		             });
+	}
+
+	void OnAttemptDone(unique_ptr<HTTPResponse> response, optional_ptr<ErrorData> error) {
```
lnkuiper, line 203: Can we do `optional<ErrorData>` so it's passed by value instead? I think that's safer

src/main/http/http_util.cpp
```diff
@@ -440,133 +561,160 @@ void HTTPUtil::DecomposeURL(const string &input, string &path_out, string &proto
...
+	// throttle responses get extra, capped, jittered backoff so bursts degrade instead of failing queries
+	const bool throttled = (response && (response->status == HTTPStatusCode::TooManyRequests_429 ||
+	                                     response->status == HTTPStatusCode::ServiceUnavailable_503)) ||
+	                       attempt.caught_status == "429" || attempt.caught_status == "503";
+#ifndef DUCKDB_NO_THREADS
+	static constexpr idx_t THROTTLE_EXTRA_RETRIES = 5;
+#else
+	// without threads we cannot sleep between retries, so do not add zero-delay retries
+	static constexpr idx_t THROTTLE_EXTRA_RETRIES = 0;
+#endif
```
lnkuiper, line 631: Is it correct to check `DUCKDB_NO_THREADS` here? Doesn't that mean WASM can't do extra retries? Shouldn't it depend on the ability to `Wait()`?

src/include/duckdb/common/http_util.hpp
```diff
@@ -326,6 +396,21 @@ class HTTPUtil {
...
 
 	virtual unique_ptr<HTTPResponse> SendRequest(BaseRequest &request, unique_ptr<HTTPClient> &client);
+	//! SendRequest, delivering the result through [on_complete] instead of returning it.
+	//! BLOCKING goes through SendRequest above, so an implementation that overrides that one keeps
+	//! its behaviour exactly. DEFERRABLE retries here instead, driving HTTPClient::Send one attempt
+	//! at a time, so a client that defers a request still gets the retry policy rather than losing
+	//! it - a backend implements HTTPClient::Send, and does not need to override this at all.
+	//! [request] and [client] must stay alive until the completion fires.
+	DUCKDB_API virtual HTTPRequestState Send(BaseRequest &request, unique_ptr<HTTPClient> &client,
+	                                         HTTPExecutionMode mode, HTTPResponseCallback on_complete);
```
lnkuiper, line 406: Is it safe passing these by value for the async request? Then the caller has to make sure they are kept alive. Seems safer if the request owns it.

lnkuiper (commented):
Thanks for the PR! Good direction. I've left some comments. Also the comments are incredibly verbose (up to 6 lines to describe a single line) please tone it down

carlopi:
Addressed the comments, rebased onto `v2.0-cyanoptera`, and iterate on this some more.

PR #25093, opened 2026-08-28
Decouple the sqllogictest runner from Catch behind a TestReporter

carlopi:
The runner called Catch's FAIL/FAIL_LINE/REQUIRE/SKIP_TEST and getCurrentTestName directly, which tied it to the unittest binary's Catch session. Route those through a TestReporter interface instead: the unittest binary installs a Catch-backed implementation, and the default implementation turns a failure into a TestFailureException (deliberately not a std::exception, so the runner's own handlers cannot swallow a verdict).

This leaves the runner usable outside a Catch session, which is what a loadable test-runner extension needs.


Mytherin:
Thanks!

PR #25094, opened 2026-08-28
Bump vcpkg to 2026.06.24

okruitho:
This PR bumps vcpkg to ~~2026.07.29~~ 2026.06.24, so packages like `curl` are now at a more modern version.

Changed behaviour for if an vcpkg repo was already installed - this will now detatch to the correct version.


carlopi:
I think this is one of the places, and this is great, but I would look also into:

```
git grep 84bab45d415d22042bd0b9081aea57f362da3f35
```
```
.github/workflows/Main.yml:        vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/Main.yml:        vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/Main.yml:        vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/Main.yml:        vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/Main.yml:          vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/Main.yml:          vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/Main.yml:        vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/NightlyTests.yml:        vcpkgGitCommitId: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/NightlyTests.yml:        vcpkgGitCommitId: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/Regression.yml:          vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
.github/workflows/Regression.yml:          vcpkg-commit: 84bab45d415d22042bd0b9081aea57f362da3f35
Makefile:#### Setup VCPKG to correct version 2025.12.12 tag is 84bab45d415d22042bd0b9081aea57f362da3f35
scripts/merge_vcpkg_deps.py:    "builtin-baseline": "84bab45d415d22042bd0b9081aea57f362da3f35",
scripts/sync_out_of_tree_extensions.py:VCPKG_BUILTIN_BASELINE = '84bab45d415d22042bd0b9081aea57f362da3f35'
```

smvv:
Besides Carlo's places, we should also update vcpkg in duckdb-ci because the container caches the vcpkg source repo to avoid downloading the repo in each CI run. FYI there is a vcpkg commit marker [check](https://github.com/duckdb/duckdb-ci/blob/b04577afa9c044f98d6b14b0300ec7ed55c65a23/build-duckdb/action.yml#L57).

After merging duckdb-ci, we get a new docker image version. We can use that image version in duckdb core CI and update vcpkg-commit values in the same PR.

I think we should use `reduced_ci_mode: disabled` (in `.github/workflows/Extensions.yml`, use only **temporarily** in this PR to test before merging), so we also see if most platforms/architectures work for the extensions.

okruitho:
When running with reduced_ci_mode to disabled we encountered an issue with the newest vcpkg release: https://github.com/microsoft/vcpkg/issues/53135. The spdx generation scripts requires a much higher version of CMAKE than vcpkg actually minimally supports. Therefore the alpine packages fail to build.

There are a few options to fix this, but unfortunately the best options are to:
- Update the hash to a newer, non release version
- Install CMAKE 4.4.0 in the alpine container for pypi / other channel

@smvv @carlopi what do you think?

carlopi:
Can we try a different release? Like maybe https://github.com/microsoft/vcpkg/releases#release-2026.06.24 has no such a requirement?

That's not really a fix, but N-1 release further is still better than none.

smvv:
I've [added](https://github.com/duckdb/duckdb-ci/blob/8698fb4622ae00a09e92eceadf81d50e78f39733/docker/alpine_3_22/build.sh#L24) cmake 4.3.2 to duckdb-ci images. Is that version sufficient, or does it have to be 4.4.0?

For context: I'm [replacing](https://github.com/duckdb/duckdb/pull/24769) the runtime built containers for extensions with prebuilt containers, so cmake used in alpine extensions CI job is much older in this PR.

okruitho:
I will experiment with an older version first. I think 4.4.0 is an requirement unfortunately.

smvv:
i [added](https://github.com/duckdb/duckdb-ci/pull/55) cmake `4.4.3`, but i'm not sure we should use the _latest_ vcpkg commit.

btw, do we need this vcpkg commit upgrade for duckdb 2.0? i'm asking because we likely need to also change the vcpkg commit in the extension-ci-tools and community extension repos.

Also, having different vcpkg commit versions across the ecosystem could also be a problem (causing a different build in the extension repo's CI compared to duckdb core's CI).

carlopi (approved):
Thanks!!

PR #25095, opened 2026-08-28
[ASYNC-IO] Read-Ahead for DuckDB Tables

pdet:
This PR implements the read-ahead async IO for the native DuckDB tables. We can now claim row-group assignments ahead of decoding and fully overlap storage reads with execution.

We also avoid unnecessary I/O by excluding vectors rejected either by sampling or zonemap checks. Consecutive blocks are also combined into larger reads, up to 32 MB. In cases where a run cannot fit within the memory limit, we branch back to the normal synchronous path to avoid OOMs.

Obs: There is ~2x faster response times and network saturation is visible on this branch vs main


src/include/duckdb/parallel/scan_read_ahead.hpp
```diff
@@ -89,6 +89,35 @@ struct ScanReadAheadJobWrapper final : public ScanReadAheadJob, public JOB_STATE
...
+	}
+	//! Take every pooled state
+	vector<unique_ptr<STATE>> TakeAll() {
+		lock_guard<mutex> guard(lock);
+		return std::move(states);
+	}
+
+private:
+	mutex lock;
+	vector<unique_ptr<STATE>> states;
```
lnkuiper: Nit: we could use `annotated_mutex` and `DUCKDB_GUARDED_BY` here

src/include/duckdb/storage/table/row_group_collection.hpp
```diff
@@ -102,9 +102,11 @@ class RowGroupCollection {
 	                              const vector<StorageIndex> &column_ids, idx_t start_row, idx_t end_row);
 	static bool InitializeScanInRowGroup(ClientContext &context, CollectionScanState &state,
 	                                     RowGroupCollection &collection, SegmentNode<RowGroup> &row_group,
-	                                     idx_t vector_index, idx_t max_row);
+	                                     idx_t vector_index, idx_t max_row, bool initialize_columns = true);
 	void InitializeParallelScan(ParallelCollectionScanState &state);
-	bool NextParallelScan(ClientContext &context, ParallelCollectionScanState &state, CollectionScanState &scan_state);
+	//! Claims the next assignment and returns its rows, zero once none is left, deferring column setup when asked
+	idx_t NextParallelScan(ClientContext &context, ParallelCollectionScanState &state, CollectionScanState &scan_state,
```
lnkuiper: Would an `optional_idx` be more clear here?

lnkuiper (commented):
Thanks! Only a few nitpicks below, and we could maybe make things a bit more clear if we use `enum` instead of `bool` but not required. Feel free to ignore and merge if you think it's ready

src/function/table/table_scan.cpp
```diff
@@ -370,6 +409,77 @@ class DuckTableScanState : public TableScanGlobalState {
...
+					return true;
+				}
+			}
+			auto &job_scan = *l_state.job->scan_state;
+			// the job's I/O was registered when it was produced, so no I/O is scheduled here
+			switch (ScanPersistentStorage(context, data_p, l_state, job_scan, output)) {
+			case PersistentScanResult::YIELD:
+				return true;
+			case PersistentScanResult::NEXT_VECTOR:
+				continue;
```
Mytherin, line 476: Ideally we return control here as well and not have a `while (true)` here - otherwise if we have e.g. a very selective filter then we get stuck "finishing" this loop before returning control.

src/function/table/table_scan.cpp
```diff
@@ -370,6 +409,77 @@ class DuckTableScanState : public TableScanGlobalState {
...
+			if (output.size() > 0) {
+				return true;
+			}
+			// the job is exhausted, fold its scan counters into this thread and recycle its state
+			l_state.job_rows_scanned += job_scan.RowsScanned();
+			job_scan.table_state.rows_scanned = 0;
+			job_scan.local_state.rows_scanned = 0;
+			state_pool.Push(std::move(l_state.job->scan_state));
+			l_state.job.reset();
+			context.InterruptCheck();
```
Mytherin: I would also prefer not to add manual `InterruptCheck` here - in general a manual `InterruptCheck` is imo a code smell. We should be returning control up the stack and handling interrupts centrally. While `InterruptCheck` solves part of the problems of a `while (true)` it doesn't solve e.g. `PendingQueryResult` wanting to do small-size executions for e.g. UI re-rendering.

src/storage/standard_buffer_manager.cpp
```diff
@@ -244,39 +244,42 @@ void StandardBufferManager::BatchRead(QueryContext context, PrefetchRun &run) {
...
+		// the blocks are read, now assign them to the individual block handles
+		for (idx_t block_idx = 0; block_idx < block_count; block_idx++) {
+			auto &handle = run.handles[block_idx];
+			D_ASSERT(handle->BlockId() == run.first_block + NumericCast<block_id_t>(block_idx));
+
+			// reserve memory for the block
+			auto &block_memory = handle->GetMemory();
+			idx_t required_memory = block_memory.GetMemoryUsage();
+			unique_ptr<FileBuffer> reusable_buffer;
+			auto reservation = EvictBlocksOrThrow(context, block_memory.GetMemoryTag(), required_memory,
```
Mytherin: It's a bit odd to use `EvictBlocksOrThrow` to then catch and swallow the exception - can we just use `buffer_pool.EvictBlocks` directly instead?

Mytherin (commented):
Thanks! Looks good - some comments

PR #25096, opened 2026-08-28
Fix incorrect column expansion

sebastiaan-dev:
Resolves https://github.com/duckdblabs/duckdb-internal/issues/10313 by normalizing filter predicates.

Previously the below query would be expanded incorrectly:
```sql
CREATE TABLE v AS VALUES (1,10,100),(2,NULL,200),(3,30,NULL);

SELECT sum(x) FILTER (WHERE COLUMNS(*) IS NOT NULL) FROM  v AS t(x,a,b);

--- rough expanded version
SELECT 
sum(x) FILTER (WHERE x IS NOT NULL), 
sum(x) FILTER (WHERE a IS NOT NULL), 
sum(x) FILTER (WHERE b IS NOT NULL) 
FROM  v AS t(x,a,b);
```

With normalization we now get:
```sql
SELECT sum(x) FILTER ( WHERE 
x IS NOT NULL AND
a IS NOT NULL AND
b IS NOT NULL
) FROM  v AS t(x,a,b);
```

This normalization was previously missing for `FILTER` expressions in the `SELECT` list, `DISTINCT ON` targets, `ORDER BY` expressions, `QUALIFY`, and `HAVING` clauses.

### DIFF
--- a/src/include/duckdb/planner/binder.hpp
+++ b/src/include/duckdb/planner/binder.hpp
@@ -618,7 +618,8 @@ private:
 	                                   vector<unique_ptr<ParsedExpression>> &replacements, StarExpression &star,
 	                                   optional_ptr<duckdb_re2::RE2> regex);
 	void BindWhereStarExpression(unique_ptr<ParsedExpression> &expr);
-	void BindFilterStarExpressions(ParsedExpression &expr);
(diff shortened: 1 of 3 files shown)


lnkuiper:
Thanks!

PR #25097, opened 2026-08-28
[Overflow] Use `TryMultiplyOperator` in `arg_max` and check with `NextPowerOfTwo` before multiplying

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/10248

Added a slow test, but kept it skipped because I don't think it's worth it to add 20sec to every release build CI workflow.


Mytherin:
Thanks!

PR #25098, opened 2026-08-28
[Dev] aggregate_state fuzzing fixes

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/10300

One issue only shows up in `DEBUG` builds, so the test is added but skipped.


Mytherin:
Thanks! Is this the same fix as https://github.com/duckdb/duckdb/pull/25291?

Tishj:
Partially, yes
That's the uniqueness check, the other part is the unbounded allocation
I'll strip this PR to include only the relevant part, as I like their fix for the uniqueness better

Mytherin:
Thanks!

PR #25099, opened 2026-08-28
Change Concat_ws to accept lists

xproblematique:
Closes https://github.com/duckdblabs/duckdb-internal/issues/3620

`concat_ws` previously stringified `LIST` arguments as a whole (e.g. producing something like `[a, b, c]` as a single joined value) instead of treating list elements as individual values to join. This PR makes `concat_ws` flatten `LIST` arguments in place alongside scalar arguments, joining every element with the given separator.

```sql
SELECT concat_ws(',', 'a', ['x', 'y', 'z'], 'b');
-- a,x,y,z,b
```

## Behaviour

- List arguments are flattened in place: elements from a list argument are joined using the same separator as every other argument, in call order, mixed freely with scalar arguments.
- Multiple list arguments in a single call are supported.
- Existing NULL-handling rules extend naturally to lists, with no special-casing:
  - A `NULL` list argument contributes nothing (same as a NULL scalar argument)
  - A `NULL` element *within* a non-NULL list is skipped individually; other elements in that same list are still joined normally
  - An empty list (`[]`) contributes nothing
  - The separator itself being `NULL` still causes the whole result to be `NULL`, unchanged from prior behavior
- Non-`VARCHAR` list element types are implicitly cast to `VARCHAR` (e.g. `concat_ws(',', [1, 2, 3])` → `1,2,3`), matching how non-`VARCHAR` scalar arguments are already handled
- Nested lists (`LIST(LIST(...))`) are rejected at bind time with a clear error rather than attempting to recursively flatten

## Implementation

- `BindConcatWSFunction`: inspects each vararg's resolved type. For `LIST` arguments, casts the *child* type to `VARCHAR` when needed (rather than casting the list as a whole, which was the source of the original stringification behavior) and throws a `BinderException` on nested lists. Non-list, non-`VARCHAR` arguments are cast to `VARCHAR` as before. Records which argument positions are lists in a new `ConcatWSBindData`
- `ConcatWSFunction`: extended the existing loop to branch per argument on whether it's a list. List arguments iterate their elements via `VectorListType<string_t>`, contributing each valid element to the running length/write exactly as a scalar argument would, sharing the same `has_result` flag so flattening reads naturally left-to-right regardless of which arguments are lists vs. scalars

## Tests

Added test cases to `test/sql/function/string/test_concat_ws.test`
Also changed the output of the test case of `test/sql/types/list/list_to_varchar_cast.test` to match the function's new behaviour

### DIFF
--- a/src/function/scalar/string/concat_ws.cpp
+++ b/src/function/scalar/string/concat_ws.cpp
@@ -1,15 +1,40 @@
 #include "duckdb/function/scalar/string_functions.hpp"
+#include "duckdb/planner/expression/bound_cast_expression.hpp"
 
 #include <string.h>
 
 namespace duckdb {
 
+struct ConcatWSBindData : public FunctionData {
+	explicit ConcatWSBindData(vector<bool> is_list_p) : is_list(std::move(is_list_p)) {
+	}
+	vector<bool> is_list; // one entry per non-separator argument, in call order
+
+	unique_ptr<FunctionData> Copy() const override {
+		return make_uniq<ConcatWSBindData>(is_list);
+	}
+	bool Equals(const FunctionData &other_p) const override {
+		return is_list == other_p.Cast<ConcatWSBindData>().is_list;
+	}
+};
+
 static void ConcatWSFunction(DataChunk &args, ExpressionState &state, Vector &result) {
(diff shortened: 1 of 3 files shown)


lnkuiper:
Thanks! LGTM

PR #25101, opened 2026-08-28
Fix crash in Top-N with LIMIT 0 (#24794)

### LINKED ISSUE #24794: Repeated execution of query with LEFT_DELIM_JOIN / MARK JOIN causes native crash (SIGSEGV) in DuckDB JDBC
### What happens?

Executing a complex `SELECT` with `LIMIT 0` repeatedly on the same JDBC connection causes a `SIGSEGV` crash in the DuckDB native layer after about 1000 iterations. The crash occurs in `org.duckdb.DuckDBNative.duckdb_jdbc_execute_pending`. Single executions of both the source and split queries return 0 rows successfully; repeated executions of the source query do not crash.

### To Reproduce

Use DuckDB JDBC (v1.5.4) and execute the following SQL statements.

```sql
DROP TABLE IF EXISTS repro_duckdb_db0_src;
DROP TABLE IF EXISTS repro_duckdb_db0_l;
DROP TABLE IF EXISTS repro_duckdb_db0_r;

CREATE TABLE repro_duckdb_db0_src (
    vp_rowid BIGINT PRIMARY KEY,
    c0 BOOLEAN,
    c1 BOOLEAN,
    c2 BIGINT
);

INSERT INTO repro_duckdb_db0_src VALUES
    (1, TRUE, NULL, NULL),
    (2, TRUE, FALSE, 42);

CREATE TABLE repro_duckdb_db0_l AS
SELECT vp_rowid, c2
FROM repro_duckdb_db0_src;

CREATE TABLE repro_duckdb_db0_r AS
SELECT vp_rowid, c0, c1
FROM repro_duckdb_db0_src;

SET threads = 1;
SET disabled_optimizers = '';

SELECT repro_duckdb_db0_src.c2
FROM repro_duckdb_db0_src repro_duckdb_db0_src
ORDER BY repro_duckdb_db0_src.vp_rowid
LIMIT 0;

SET threads = 4;
SET disabled_optimizers = 'filter_pushdown,join_order';

SELECT repro_duckdb_db0_src.c2
FROM (
    SELECT
        l.vp_rowid,
        r.c0,
        r.c1,
        l.c2
    FROM repro_duckdb_db0_l l
    JOIN repro_duckdb_db0_r r
      ON l.vp_rowid = r.vp_rowid
    WHERE NOT EXISTS (
        SELECT 1
        FROM repro_duckdb_db0_r rx
        WHERE rx.vp_rowid = l.vp_rowid
          AND rx.vp_rowid <> l.vp_rowid
    )
) repro_duckdb_db0_src
ORDER BY repro_duckdb_db0_src.vp_rowid
LIMIT 0;
```

Then execute the last `SELECT` (the “split” query) repeatedly on the same JDBC connection, e.g. using a loop in Java. After about 1000 executions, the JVM crashes.

Output of single executions:

```text
source: 0 rows
split:  0 rows
```

Repeated execution result:

```text
source query repeated ~1000 times: no crash
split query repeated ~1000 times: JVM/native crash
```

### Observed execution plans

The source query plan:

```text
PROJECTION
  ORDER_BY
    HASH_JOIN
      Join Type: SEMI
      SEQ_SCAN repro_duckdb_db0_src
      TOP_N Top: 0
```

The split query plan includes:

```text
TOP_N Top: 0
  FILTER (NOT SUBQUERY)
    LEFT_DELIM_JOIN
      Join Type: MARK
      HASH_JOIN
        Join Type: INNER
      HASH_JOIN
        Join Type: MARK
      HASH_GROUP_BY
      CROSS_PRODUCT
        SEQ_SCAN repro_duckdb_db0_r
        DELIM_SCAN
```

The plans diverge significantly; the split query triggers `LEFT_DELIM_JOIN`, `MARK JOIN`, `HASH_GROUP_BY` and `DELIM_SCAN`.

### OS:

Ubuntu 20.04.6 LTS, x86_64

### DuckDB Version:

v1.5.4

### DuckDB Client:

Java JDBC

### Hardware:

Intel Xeon Gold 5218R, 80 cores, 1486G RAM

### Full Name:

Annie liu

### Affiliation:

ECNU

- [x] Yes, I have

- [x] Yes, I have

Yes

RagamuffinCat:
Fixes: #24794 

`TopNHeap` accessed `heap.front()` when `heap_size == 0`, causing an unchecked read from an empty `unsafe_arena_vector` and potentially dereferencing a stale `string_t` pointer. This is reachable when `filter_pushdown` is disabled, since `LIMIT 0` is normally rewritten to `LogicalEmptyResult`.

The fix handles zero-sized heaps in `EntryShouldBeAdded()` and `Sink()`, guards the global boundary update with `!heap.empty()`, and adds `D_ASSERT`s for the heap invariants.

Tests: new `test/sql/order/top_n_limit_zero.test` covering 500 consecutive executions of the reproducer. The query crashes on `main` under ASAN and passes with the patch. The relevant `[order]`, `[optimizer]`, `[explain]`, and `[subquery]` test groups pass.

### DIFF
--- a/src/execution/operator/order/physical_top_n.cpp
+++ b/src/execution/operator/order/physical_top_n.cpp
@@ -138,8 +138,12 @@ public:
 
 private:
 	inline bool EntryShouldBeAdded(const string_t &sort_key) {
+		if (heap_size == 0) {
+			// the heap has no capacity (LIMIT 0) - no entry can ever be added
+			return false;
+		}
 		if (heap.size() < heap_size) {
-			// heap is full - check the latest entry
+			// heap is not full yet - the entry can be added
 			return true;
 		}
 		if (sort_key < heap.front().sort_key) {
@@ -151,7 +155,9 @@ private:
 	}
 
 	inline void AddEntryToHeap(const TopNEntry &entry) {
+		D_ASSERT(heap_size > 0);
 		if (heap.size() >= heap_size) {
+			D_ASSERT(!heap.empty());
 			std::pop_heap(heap.begin(), heap.end());
 			heap.pop_back();
 		}
@@ -347,6 +353,11 @@ bool TopNHeap::CheckBoundaryValues(DataChunk &sort_chunk, DataChunk &payload, To
 void TopNHeap::Sink(DataChunk &input, optional_ptr<TopNBoundaryValue> global_boundary) {
 	static constexpr idx_t SMALL_HEAP_THRESHOLD = 100;
 
+	if (heap_size == 0) {
+		// LIMIT 0 without OFFSET - the heap can never hold any entry, so there is nothing to do
+		return;
+	}
+
 	// compute the ordering values for the new chunk
 	sort_chunk.Reset();
 	executor.Execute(input, sort_chunk);
@@ -372,7 +383,7 @@ void TopNHeap::Sink(DataChunk &input, optional_ptr<TopNBoundaryValue> global_bou
 
 	// if we modified the heap we might be able to update the global boundary
 	// note that the global boundary only applies to FULL heaps
-	if (heap.size() >= heap_size && global_boundary) {
+	if (!heap.empty() && heap.size() >= heap_size && global_boundary) {
 		global_boundary->UpdateValue(heap.front().sort_key);
 	}
 }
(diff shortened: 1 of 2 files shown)


carlopi:
I think it would be valuable to also backport this to `v1.5-variegata`

RagamuffinCat:
@carlopi Thanks for the advice! I've opened #25103 for `v1.5-variegata`

Mytherin:
Thanks!

PR #10579, opened 2024-02-11
[Tester] Add `--require <name>`

Tishj:
Same idea as https://github.com/duckdb/duckdb/pull/10577 but a different execution, more selective.

By specifying `--require json` we add `json` to the list of extensions that have to be present, if we reach a `require json` line and it is not present that constitutes a test failure, rather than silently skipping the test.

This flag works additively, so we can add `--require <name>` for all the extensions we require to be present.

PR #25102, opened 2026-08-28
Internal #10579: .manual Light Mode

hawkfish:
* Use BLACK not WHITE text in light mode.


Mytherin:
Thanks!

Mytherin:
This should really be using the `HighlightElement` infrastructure instead, but we can pick that up at a later time :)

hawkfish:
> This should really be using the `HighlightElement` infrastructure instead, but we can pick that up at a later time :)

Yeah I thought about that but I really just wanted it for my slides and I'd never seen the code before!

PR #25103, opened 2026-08-28
Backport #25101: Fix crash in Top-N with LIMIT 0 (#24794)

RagamuffinCat:
Backport of #25101 to v1.5-variegata.

Applies cleanly. Verified locally: the regression test crashes without the fix and passes with it.

### DIFF
--- a/src/execution/operator/order/physical_top_n.cpp
+++ b/src/execution/operator/order/physical_top_n.cpp
@@ -137,8 +137,12 @@ public:
 
 private:
 	inline bool EntryShouldBeAdded(const string_t &sort_key) {
+		if (heap_size == 0) {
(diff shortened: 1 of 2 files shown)


RagamuffinCat:
The `Forwards compatibility tests` failure appears unrelated to this PR. It pipes the bash-only install script into `sh` (`dash` on Ubuntu), causing the 1.2.1 CLI download to fail. Reproduced locally: `sh` fails, while `bash` works.

Everything related to this patch is green: Linux Debug, Linux Release, Valgrind, Thread Sanitizer, Vector Sizes, and Amalgamation.

Mytherin:
Thanks!

PR #25106, opened 2026-08-28
Do not reuse blocks freed by a checkpoint until its header is durable

ywelsch:
WriteHeader moves a checkpoint's freed blocks out of modified_blocks and into free_list, releases single_file_block_lock, and only then flushes metadata and writes the new header. Anything allocating in that window (e.g. optimistic writes) gets one of those blocks, and overwrites a block the header currently on disk still points at.

Usually this is invisible, because the new header lands and supersedes the one that referenced the block. If there's a crash in-between though, reopening the database afterwards reads a next-pointer out of a block that now holds row data:

```
IO Error: Could not read enough bytes from file "...": attempted to read 262144 bytes from location 39413575680
```

The fix is to free the blocks only after the header is durable instead of before.

The test reproduces the race with concurrent optimistic writers and uses debug_checkpoint_abort='after_free_list_write' to stop the checkpoint inside the window, so the old header stays newest on disk, then asks whether the database reopens. It failed 5 runs out of 5 before this change and passes after.

### DIFF
--- a/src/storage/single_file_block_manager.cpp
+++ b/src/storage/single_file_block_manager.cpp
@@ -1323,12 +1323,9 @@ void SingleFileBlockManager::WriteHeader(QueryContext context, DatabaseHeader he
 	header.iteration = ++iteration_count;
 
 	set<block_id_t> all_free_blocks = free_list;
-	set<block_id_t> fully_freed_blocks;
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25115, opened 2026-08-28
Issue#25085:  Interval Invert Overflow

### LINKED ISSUE #25085: Vuln48: Signed Integer Overflow in Interval::Invert() via DATE/TIMESTAMP Subtraction
### What happens?

`SubtractOperator::Operation(date_t, interval_t)` and its timestamp counterpart in `src/function/scalar/operator/subtract.cpp` call `Interval::Invert()` to negate the interval before adding it. `Interval::Invert()` in `src/common/types/interval.cpp` negates `months`, `days`, and `micros` unconditionally with no bounds check. When an interval literal's `months` or `days` field equals `INT32_MIN` (`-2147483648`), which the normal numeric-literal parsing path in `Interval::FromCString` accepts without error, the negation `-interval.months` overflows a signed 32-bit integer, which is undefined behavior in C++. A guard for exactly this case already exists in the `... ago` keyword parsing branch of `Interval::FromCString`, but it is missing from `Interval::Invert()`, so the same unguarded negation is reachable directly through plain `DATE - INTERVAL` and `TIMESTAMP - INTERVAL` subtraction.

### To Reproduce

This is reproducible with pure SQL directly from the duckdb CLI, no input file needs to be constructed.

```sql
SELECT DATE '2000-01-01' - INTERVAL '-2147483648 months';
SELECT DATE '2000-01-01' - INTERVAL '-2147483648 days';
SELECT TIMESTAMP '2000-01-01 00:00:00' - INTERVAL '-2147483648 months';
```

```shell
./build/debug/duckdb <<'EOF'
SELECT DATE '2000-01-01' - INTERVAL '-2147483648 months';
SELECT DATE '2000-01-01' - INTERVAL '-2147483648 days';
SELECT TIMESTAMP '2000-01-01 00:00:00' - INTERVAL '-2147483648 months';
EOF
```

```
.../src/common/types/interval.cpp:551:20: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
EXIT=1
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

hawkfish:
* Negating an interval field at its minimum value (INT32_MIN months/days, INT64_MIN micros) is undefined behavior; 
* throw OutOfRangeException instead.

fixes: #25085

### DIFF
--- a/src/common/types/interval.cpp
+++ b/src/common/types/interval.cpp
@@ -546,6 +546,10 @@ interval_t Interval::FromMicro(int64_t delta_us) {
 }
 
 interval_t Interval::Invert(interval_t interval) {
+	if (interval.months == NumericLimits<int32_t>::Minimum() || interval.days == NumericLimits<int32_t>::Minimum() ||
+	    interval.micros == NumericLimits<int64_t>::Minimum()) {
+		throw OutOfRangeException("Interval value is out of range for negation");
+	}
 	interval.days = -interval.days;
 	interval.micros = -interval.micros;
 	interval.months = -interval.months;
--- a/test/sql/types/interval/interval_invert_overflow.test
+++ b/test/sql/types/interval/interval_invert_overflow.test
@@ -1,0 +1,41 @@
+# name: test/sql/types/interval/interval_invert_overflow.test
+# description: Interval negation of extreme values should error instead of overflowing
+# group: [interval]
+
+# Issue #25085: DATE/TIMESTAMP subtraction negates the interval without bounds checking
+
+statement error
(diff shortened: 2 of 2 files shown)


hawkfish:
Developed by Claude, reviewed by me - ready to merge.

Mytherin:
Thanks!

hawkfish:
Looks like someone else's Claude fix already got merged...

PR #25116, opened 2026-08-28
Issue#25071: Interval Colon Overflow

### LINKED ISSUE #25071: Bug9: Unchecked int64 Overflow in Interval Colon-Time Addition
### What happens?

`Interval::FromCString()` in `src/common/types/interval.cpp` parses interval literals term by term, and every numeric term addition goes through the overflow-checked `IntervalTryAddition`/`TryAddOperator` path except the colon-separated time-of-day branch (`interval_parse_time`, around line 172), which does a raw `result.micros += time.value;`. Because `Time::TryConvertInternal` allows up to 9 digits of hours specifically to support intervals, a colon-time term can contribute a value close to `INT64_MAX` on its own. If a prior, individually-checked term has already pushed `result.micros` near `INT64_MAX`, the unchecked addition in the colon-time branch overflows a signed 64-bit integer, which is undefined behavior and can silently wrap into a corrupted interval value on release builds.

### To Reproduce

This is reproducible with a single SQL statement from the duckdb CLI, no input file is needed.

```sql
SELECT '9223372036854775807 microseconds 999999999:00:00'::INTERVAL;
```

```shell
./build/debug/duckdb -c "SELECT '9223372036854775807 microseconds 999999999:00:00'::INTERVAL;"
```

```
build/debug/src/common/types/../../../../../src/common/types/interval.cpp:172:16: runtime error: signed integer overflow: 9223372036854775807 + 3599999996400000000 cannot be represented in type 'long int'
EXIT=1
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

hawkfish:
Colon-separated time terms were added to the accumulated micros without overflow checking, and a negative sign negated the entire accumulation (UB at INT64_MIN). Apply the sign to the time term only and add it via the checked IntervalTryAddition path, matching PostgreSQL semantics for mixed-sign inputs such as '1 hour -00:30:00'.

fixes: #25071

### DIFF
--- a/src/common/types/interval.cpp
+++ b/src/common/types/interval.cpp
@@ -169,11 +169,10 @@ interval_parse_time : {
 	if (!Time::TryConvertInterval(str + start_pos, len - start_pos, pos, time)) {
 		return false;
 	}
-	result.micros += time.value;
-	found_any = true;
-	if (negative) {
-		result.micros = -result.micros;
+	if (!IntervalTryAddition<int64_t>(result.micros, negative ? -time.value : time.value, 1, error_message)) {
+		return false;
 	}
+	found_any = true;
 	goto end_of_string;
 }
 interval_parse_identifier:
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25119, opened 2026-08-29
Reject `CREATE TEMPORARY INDEX` / `TRIGGER`, inherit trigger temporariness

carlopi:
Similar to https://github.com/duckdb/duckdb/pull/25118, this I am less sure, should `TEMPORARY` indexes or trigger be explicitly tagged, or (as done in the PR) they follow the characterisation of the relevant resource?

### DIFF
--- a/src/parser/peg/transformer/transform_create_table.cpp
+++ b/src/parser/peg/transformer/transform_create_table.cpp
@@ -38,6 +38,18 @@ PEGTransformerFactory::TransformCreateStatement(PEGTransformer &transformer, con
 		secret_info.persist_type = temporary ? *temporary : SecretPersistType::DEFAULT;
 	}
 	result->info->temporary = temporary && *temporary == SecretPersistType::TEMPORARY;
+	if (result->info->temporary) {
+		// the grammar allows TEMPORARY in front of every create variation - reject it where the entry does not
+		// carry a user-supplied flag, but derives one from the table it belongs to
+		switch (result->info->type) {
+		case CatalogType::INDEX_ENTRY:
+			throw ParserException("Temporary indexes are not supported");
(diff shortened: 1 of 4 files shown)


Mytherin:
I think disallowing explicit `TEMPORARY` for these types makes sense - indeed they should follow the temporary-ness of the target table. Thanks!

lnkuiper:
I think we need a follow-up. Repros below.

`PERSISTENT` silently becomes temporary:

```sql
CREATE TEMP TABLE t(i INTEGER);
CREATE TEMP TABLE audit(i INTEGER);

CREATE PERSISTENT INDEX idx ON t(i);
CREATE PERSISTENT TRIGGER trg AFTER INSERT ON t
    INSERT INTO audit VALUES (1);

SELECT index_name, database_name
FROM duckdb_indexes()
WHERE index_name = 'idx';
-- idx | temp

SELECT trigger_name, database_name, temporary
FROM duckdb_triggers()
WHERE trigger_name = 'trg';
-- trg | temp | true
```

A catalog copy loses trigger temporariness:

```sql
CREATE TEMP TABLE tmp(i INTEGER, j INTEGER);
CREATE TEMP TABLE tmp_audit(i INTEGER);
CREATE TRIGGER tmp_trg AFTER UPDATE OF j ON tmp
    INSERT INTO tmp_audit VALUES (1);

SELECT temporary, columns
FROM duckdb_triggers()
WHERE trigger_name = 'tmp_trg';
-- true | [j]

ALTER TABLE tmp RENAME COLUMN j TO k;

SELECT temporary, columns
FROM duckdb_triggers()
WHERE trigger_name = 'tmp_trg';
-- false | [k]
```

Not sure how serious.

PR #25121, opened 2026-08-29
HASH_SKEW: compile time flag, similar to HASH_ZERO, remixing hash function via parametrized and deterministic re-mixing

carlopi:
This helped find https://github.com/duckdb/duckdb/pull/24947, unsure if it can be of generic use.

Idea is that HASH_ZERO is maximally bad, but that also mean many tests become patologically slow.
Here we provide a param, that ties to what percentage of buckets gets re-mapped to a the same bucket.

In some way HASH_SKEW=100 has the same effect of HASH_ZERO=1, so fully colliding, but lower percentages allow to tune this and explore spaces where one bucket is particular hot while other buckets are also existing (with unmodified percentages).


Mytherin:
Thanks for the PR!

I think this is still pretty specific - it might be nicer to make the hash function configurable instead, that would then make implementing either hash zero or hash skew trivial (and might also unlock other functionality). This would be more plumbing of course.

PR #25133, opened 2026-08-29
Prune row groups for ENUM predicates whose constants cover the full domain

### LINKED ISSUE #24417: ENUM exhaustive domain not pruned

JaySon-Huang:
Hi team!

This fixes #24417: a `NOT IN` listing every value of an ENUM column is unsatisfiable, but DuckDB still reads every row group and filters row by row. The integer side of this has worked since #24521; ENUM falls one step short of the same machinery.

`NumericStats::ConstantsCoverRange` gates full-range constant coverage on `type.IsIntegral()`, and ENUM is not in that set even though ENUM columns carry numeric min/max zonemaps. An ENUM is a closed domain declared by its type, so coverage is a set test against the declared domain rather than a `[min, max]` interval test and needs no zonemaps at all: this adds `NumericStats::ConstantsCoverDomain` and routes ENUM statistics to it from `ConstantsCoverRange`. For nullable columns the child `IN` can never classify as always-true (it is true-or-NULL), so `CheckNotOperatorStatistics` detects the coverage itself and returns `FILTER_FALSE_OR_NULL`, which the existing empty-result machinery collapses. And this also reaches DELETE/UPDATE plans through the default statistics propagation.

On three row groups whose values span the full domain:

```sql
EXPLAIN ANALYZE SELECT count(*) FROM t WHERE m NOT IN ('sad', 'ok', 'happy');
-- main:  Row Groups Scanned: 3 / 3
-- patch: Empty Result
```

I notice that there is an open expression-rewrite attempt for the same issue in #24706. While the changes in this branch takes the statistics-layer route, and also covers nullable columns and DML plans where a rewrite pass keeps the filter in place.
#24706 can do two things this route cannot. For the positive `IN` over the full domain it removes the filter entirely - here an OR rewrite above the scan still evaluates per row. And its rewrite needs no column statistics, so it also helps stats-free inputs like `VALUES`, where this change does nothing.

```sql
EXPLAIN SELECT count(*) FROM t WHERE m IN ('sad', 'ok', 'happy');
-- #24706: filter gone - the IN rewrites to a constant true on non-NULL columns
-- patch:  scan-level IN filter dropped, but (m = 'sad') OR (m = 'ok') OR (m = 'happy') stays and evaluates per row

SELECT count(*) FROM (VALUES ('sad'::mood), ('ok'::mood), ('happy'::mood)) v(m) WHERE m NOT IN ('sad', 'ok', 'happy');
-- #24706: predicate rewritten to a cheap per-row constant_or_null without any statistics
-- patch:  nothing to classify without statistics, the filter passes through unchanged
```
I verified those two behaviors against both branches as well and add test cases in `test/sql/optimizer/enum_domain_row_group_pruning.test`

### DIFF
--- a/src/include/duckdb/storage/statistics/numeric_stats.hpp
+++ b/src/include/duckdb/storage/statistics/numeric_stats.hpp
@@ -76,6 +76,9 @@ struct NumericStats {
 	//! Whether the constants cover every value in [min, max] - only meaningful for integral types
 	static bool ConstantsCoverRange(const BaseStatistics &stats, array_ptr<const Value> constants);
 
+	//! Whether the constants cover the entire declared domain of an ENUM - no zonemaps needed
+	DUCKDB_API static bool ConstantsCoverDomain(const BaseStatistics &stats, array_ptr<const Value> constants);
+
 	DUCKDB_API static void Merge(BaseStatistics &stats, const BaseStatistics &other_p);
 
 	DUCKDB_API static void Serialize(const BaseStatistics &stats, Serializer &serializer);
--- a/src/planner/filter/expression_filter.cpp
+++ b/src/planner/filter/expression_filter.cpp
@@ -609,6 +609,38 @@ static FilterPropagateResult CheckInOperatorStatistics(optional_ptr<ClientContex
 	return result;
 }
 
+static bool InCoversEnumDomain(optional_ptr<ClientContext> context_p, const Expression &expr,
+                               array_ptr<const BaseStatistics> input_stats) {
+	if (expr.GetExpressionType() != ExpressionType::COMPARE_IN) {
+		return false;
+	}
+	auto &op_expr = expr.Cast<BoundOperatorExpression>();
+	if (op_expr.GetChildren().size() <= 1) {
+		return false;
+	}
+	vector<unique_ptr<BaseStatistics>> owned_stats;
+	auto filter_stats = TryGetFilterStats(context_p, *op_expr.GetChildren()[0], input_stats, owned_stats);
+	if (!filter_stats || filter_stats->GetType().id() != LogicalTypeId::ENUM) {
+		return false;
+	}
+	vector<Value> values;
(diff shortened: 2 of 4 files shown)


lnkuiper:
Thanks for the PR! I think #24706 puts the logic in the right place, rather than adding it to `NumericStats`/`BaseStatistics` as done here.

Full domain coverage is a property of the `ENUM` type and the `IN` constants, not of the stored statistics: `ConstantsCoverDomain` only uses `stats.GetType()`. It is also re-evaluated for every row group, including rebuilding the coverage set each time.

Could we extend the rewrite approach instead? It should cover `ENUM` as well as `VARCHAR`, and handle the safe false-or-NULL predicate-root case in DML without a plan-wide `HasSideEffects()` check. That would retain the nullable `NOT IN` and DML behavior here, while also helping stats-free inputs and avoiding this new statistics API.

PR #25135, opened 2026-08-29
Make the regex 'm' option enable newline-sensitive matching (#17056)

### LINKED ISSUE #17056: Regex `m` option parameter does not enable "Newline sensitive matching"
### What happens?

When using `regexp_extract`, with the `m` option as a final argument, the resulting query does not apply the correct, documented behavior.

```sql
SELECT regexp_extract(
    'Line 1
     Some text
     Line 2
     More text',
    '^Line.*$',
  0,
  'm');
```

Will yield an empty string, `''`

The [DuckDB documentation](https://duckdb.org/docs/stable/sql/functions/regular_expressions#options-for-regular-expression-functions) states that:

```
'm', 'n', 'p' | Newline sensitive matching
```

If I provide `(?m)` as a prefix, then the newline sensitive matching works correctly...

```sql

SELECT regexp_extract(
    'Line 1
     Some text
     Line 2
     More text',
    '(?m)^Line.*$');
```
== `'Line 1'`

From what I've gathered, it might be found here:

https://github.com/duckdb/duckdb/blob/e1dcf13ed533d2f071d94597b4144d83242e78b9/src/function/scalar/string/regexp.cpp#L224-L321

### To Reproduce

Provide this query:

```sql

SELECT regexp_extract(
    'Line 1
     Some text
     Line 2
     More text',
    '^Line.*$',
  0,
  'm');

```

You will get an empty string `''`

### OS:

macOS (arm64)

### DuckDB Version:

1.2.0

### DuckDB Client:

CLI / C++

### Hardware:

_No response_

### Full Name:

Ryan Melehan

### Affiliation:

Coco Alemana

- [x] Yes, I have

- [x] Yes, I have

marcodelpin:
Fixes #17056.

## Problem

The regex `'m'` option (and its `'n'`/`'p'` aliases) does not enable newline-sensitive matching: `regexp_matches(s, '^Line', 'm')` never matches at a line boundary, because the option is parsed but not applied to the compiled pattern.

## The change

- Thread a `multiline` flag through option parsing and apply it as a `(?m)` prefix on the effective pattern - to constant patterns during binding and to dynamic patterns at compile time - covering constant and dynamic patterns across `regexp_matches` / `regexp_replace` / `regexp_extract` / `regexp_extract_all` / `string_split_regex`.
- The three-argument `regexp_extract(input, pattern, 'm')` form parses options before the prefix is applied, so both arities share the same path (group-name LIST validation compiles the effective pattern, capture numbering unchanged).
- The `string_split_regex` slow path (non-constant patterns) now compiles with the bind-data options instead of RE2 defaults. This was dropping every option there, not just `'m'` - a dynamic-pattern `'i'` split behaved differently from its constant twin; both are locked down in the tests.
- `RegexpBaseBindData::Equals` now includes `multiline`, preventing multiline and non-multiline bind data from comparing equal.
- The regex-to-`contains`/`prefix` optimizer bails out on multiline patterns only when the pattern carries a top-level text/line anchor; anchor-free patterns mean the same thing with or without `'m'` and keep the rewrite.

## Semantics notes (deliberate, flagged for review)

- `'sm'` combined: both behaviors are now active (`.` matches newline AND `^`/`$` match at line boundaries). Previously the later flag effectively won. The new tests pin the 2x2 (`s`/`m`/`sm`/none).
- `'p'` follows the same full line-anchor behavior as `'m'`/`'n'`. Postgres treats `'p'` as partial newline-sensitivity; if that distinction matters here, happy to adjust.

## Testing

`test/sql/function/string/regex_multiline_option.test` (40 test directives): constant + dynamic patterns per function, 3-arg and 4-arg extract, group-name LIST form, the `'sm'` discriminator, `'n'`/`'p'`, `$` before an embedded newline, dynamic-`'i'` split lock-down. Each fix was written red-first against the previous head. Suites: `regex*` 762 assertions, `string/*` 4484, `test/optimizer/*` 8,859 assertions across 173 tests - all green. `clang-format` clean on the touched files.

## Disclosure

Developed with AI assistance; all changes reviewed, tested and verified by me.


marcodelpin:
Could someone take a first look at this PR? It addresses #17056 by applying the regex 'm' option as a (?m) prefix for both constant and dynamic patterns.

The upstream [Main workflow run](https://github.com/duckdb/duckdb/actions/runs/35323135812) completed successfully on the current head da3a2c7bb4.

I am happy to adjust the approach or scope based on feedback.

PR #25136, opened 2026-08-29
Bail out of top-N window elimination when ROW_NUMBER has an argument-level ORDER BY (#23678)

### LINKED ISSUE #23678: top_n_window_elimination ignores ROW_NUMBER argument ORDER BY

marcodelpin:
Fixes #23678.

## Problem

`top_n_window_elimination` ignores `ROW_NUMBER`'s argument-level `ORDER BY` (`ROW_NUMBER(ORDER BY x) OVER (...)`): the rewrite ranks by the frame's `ORDER BY` only, so when the two orderings conflict the optimized query returns different rows than the unoptimized plan. On the official nightly at `8616efa9da`, the issue's shape returns `(1,1,1)/(4,2,1)` optimized vs `(3,1,3)/(5,2,2)` with `SET disabled_optimizers = 'top_n_window_elimination'`.

## The change

Bail out of the rewrite in `CanOptimize` when `window_expr.ArgOrders()` is non-empty - after the `WINDOW_ROW_NUMBER` check, before the frame `OrderBy` checks. Teaching the rewrite argument-level ordering semantics (frame-sensitive ranking, shared or partially shared orderings) looked like a larger optimizer feature than this correctness fix warranted; happy to attempt it if that is the preferred direction.

## Testing

- New `test/sql/optimizer/topn_window_elimination_arg_order.test`: deliberately conflicting argument-level and frame orderings, asserted equal to the disabled-optimizer result (13 assertions).
- Mutation check during development: removing the bail-out fails the new test and reproduces the nightly's wrong output byte-identically; the five sibling topn test files (391 assertions) stay green either way, so the guard has zero collateral.
- Full topn set green on the fixed build: arg_order 13, late_materialization 127, set_elimination 8, group_stats 13, cte_mixed_partition_types 20, elimination.test_slow 223, parquet.test_slow 133.

## Disclosure

Developed with AI assistance; all changes reviewed, tested and verified by me.


marcodelpin:
Closing: this was fixed independently upstream before the PR was opened.

`ad2f342827` ("Fix TopN window elimination with argument ordering", 2026-07-31) adds the same `ArgOrders().empty()` bail-out in `topn_window_elimination.cpp`, and ships `test/sql/optimizer/topn_window_elimination_argument_order.test` covering it. My branch predates that landing on my side but not on yours, so what is left here is a duplicate guard plus a comment.

One thing in my test is not covered by the upstream one, noted only so it is on record rather than lost: the second case re-runs the same query under `SET disabled_optimizers = 'top_n_window_elimination'` and asserts the two agree, instead of pinning a hardcoded expectation. Happy to send that as a small separate addition if it is wanted; not opening one unasked.

Thanks for the retarget onto `v2.0-cyanoptera` earlier this week.

PR #25137, opened 2026-08-29
Support ifnull() via function chaining syntax (#11907)

### LINKED ISSUE #11907: ifnull is not chainable
### What happens?

Function chaining works correctly for the nullif function, but returns an error for the ifnull function.

### To Reproduce

nullif:
`SELECT ['abc'].nullif(['def'])` correctly returns ['abc']

ifnull without chaining:
`SELECT ifnull(['abc'], ['def'])` correctly returns ['abc'].

ifnull with function chaining:
`SELECT ['abc'].ifnull(['def'])` returns an error of "Parser Error: Wrong number of arguments to IFNULL."

### OS:

Mac M1

### DuckDB Version:

0.10.1

### DuckDB Client:

Python

### Full Name:

Dan Gottlieb

### Affiliation:

Central Square Solutions

Yes

- [X] Yes, I have

- [X] Yes, I have

marcodelpin:
Fixes #11907.

## Problem

The parser rewrites six names into syntax nodes instead of resolving them as catalog functions: `unpack`, `try`, `construct_array`, `ifnull`, `coalesce`, `date`. That rewrite lived inline in the plain-call path only, so the method-chain path never reached it, and each of those names failed through `x.f(y)` with `Catalog Error: Scalar Function with name <x> does not exist!` while `f(x, y)` worked.

#11907 reports it for `ifnull` (`SELECT ['abc'].ifnull(['def'])`); the test includes that reproduction. It is one instance of a gap that covered every rewritten name.

## The change

The rewrites move into one helper, `TransformParserFunctionRewrite(name, args, call_form)`, called from both paths.

`coalesce` takes the call form into account. It has a grammar rule of its own (`CoalesceExpression`), which is how the plain path normally spells it; reaching the plain path *under that name* means the caller wrote it quoted or schema-qualified, which resolves in the catalog and lets a user macro named `coalesce` own it. The method chain has no grammar form, so there the rewrite is the only route. Hence `ParserRewriteCallForm` rather than one undifferentiated set.

`if` already resolves through a catalog macro upstream (`76d05aef33`) and is unchanged here; the branch's last commit only drops the now-stale `.if()` cases from the test.

### What changes observably

Chained calls now apply these parser rewrites before catalog lookup, including when a same-named catalog function or macro exists - previously the chain reached the catalog entry and the plain form did not. A rewritten chained call also applies the rewrites' own argument checks, and rejects `DISTINCT` and `ORDER BY`, which the method grammar accepts on every call but a syntax node has nowhere to put. Existing plain-call behaviour is preserved.

The two paths still differ in one place, deliberately: the plain path returns its rewrite before it reaches modifier handling, so `ifnull(1, 2 ORDER BY 3)` discards the modifier where the chained form rejects it. Making the plain path reject as well would change behaviour unrelated to this issue.

## Testing

`test/sql/parser/parser_rewrites_method_chain.test` - one block per rewritten name, each asserting the chained form agrees with the equivalent plain call, plus:

- the issue's reproduction, nested chains resolving inner-first, and `nullif` as an unaffected ordinary function;
- `unpack` twice: the chained form reaching the binder with the same complaint as the plain call, and then in a valid STAR/COLUMNS context, so the rewritten child is checked and not only its error;
- `coalesce` variadic (the chain must not acquire an arity limit) and short-circuiting (skipped arguments never evaluated);
- `coalesce` written as a function call, quoted and schema-qualified, still reaching a user macro on the plain path while the chain still rewrites;
- the catalog-precedence change above, and a name with no rewrite still reaching its macro through the chain;
- arity and named-argument validation surviving the chain, and an unknown name still being an ordinary catalog lookup;
- modifier rejection on a rewritten chain, with the boundary case beside it: a chained call that is *not* a rewrite keeps its modifiers.

The full build and test suite have not been re-run against the current head.

## Disclosure

Written with AI assistance. I have reviewed every change and am answerable for it.


src/parser/peg/transformer/transform_expression.cpp
```diff
@@ -103,7 +103,18 @@ PEGTransformerFactory::TransformBaseExpression(PEGTransformer &transformer,
 		} else if (indirection_expr->GetExpressionClass() == ExpressionClass::FUNCTION) {
```
lnkuiper, line 190: Could we avoid making the method-chain assembly know about `ifnull`? This bypasses the normal parser lowering, which leaves `(NULL::INTEGER).coalesce(1)` failing as a catalog lookup. Maybe route the constructed method call through shared function lowering so `COALESCE`, `IFNULL`, and chained calls use the same path.

lnkuiper (changes requested):
This only fixes `.ifnull`; `.coalesce` still fails:

Maxxen:
Can't we just turn both of these into macros instead? That would preserve the short-circuiting and remove them from the parser entirely.

lnkuiper:
Also fine with me if that works generically

marcodelpin:
I tried the macro route before pushing anything, because if it works it is clearly the better shape. It works for `ifnull` and it does not work for `coalesce`, and the reason turned out not to be the one I expected.

**Where the macro route holds.** Short-circuiting really is preserved, and method-chaining reaches a macro with no special-casing at all:

```sql
CREATE MACRO mc(a,b) AS CASE WHEN a IS NOT NULL THEN a ELSE b END;
SELECT mc(1, error('boom'));      -- 1
SELECT (1).mc(error('boom'));     -- 1
```

Type resolution matches too: `typeof(coalesce(NULL::INTEGER, 3.5))` and `typeof(mc(NULL::INTEGER, 3.5))` are both `DECIMAL(11,1)`.

**Where it breaks.** A `CASE`-based macro names its first argument twice, and macro expansion is AST substitution, so both occurrences are evaluated:

```sql
CREATE SEQUENCE s START 1;
SELECT mc(nextval('s'), 99);   -- 2, and currval('s') is 2   -> evaluated twice
SELECT coalesce(nextval('s'), 99);  -- evaluated once
```

For a side-effecting or non-deterministic first argument that is a semantic change, and for an expensive one a performance regression. There is no let-binding in a macro body to avoid it.

The harder blocker is arity. `coalesce` is unbounded today (`coalesce(<50 args>)` is fine), and macros are not variadic:

```sql
CREATE MACRO v(a, ...) AS a;   -- Parser Error: syntax error at or near "..."
```

Comma-separated fixed-arity overloads are the only multi-arity mechanism, and a macro body cannot call itself, so a recursive definition is not available either. That leaves a finite enumeration of arities standing in for an unbounded operator, which is why I do not think it works generically.

One practical note: the two halves are coupled, so there is no incremental way to try this. A macro named `coalesce` is never reached, because the parser rewrites the call before any catalog lookup:

```sql
CREATE MACRO coalesce(a,b) AS 999;
SELECT coalesce(NULL, 5);   -- 5, not 999
```

**What I pushed instead.** The original objection was right and the fix was too narrow. `281e74a6a` extracts the rewrite table into one helper and calls it from both the plain-call path and the method-chain path, so the two cannot drift again. `.coalesce()`, `.if()`, `.try()`, `.unpack()`, `.construct_array()` and `.date()` all work now, with the execution semantics unchanged: these stay `OPERATOR_COALESCE` / `CaseExpression` nodes, so the vectorised per-row short-circuit is exactly what it was.

The test file covers `.coalesce()` at two, three and four arguments plus `(1).coalesce(error('must not be evaluated'))` for the short-circuit. `extract()` is still not reachable by chaining: its grammar form is `EXTRACT(field FROM source)`, so the chained argument order does not map onto it mechanically. That looks like a separate change rather than an omission of this one, but say the word if you would rather see it here.

marcodelpin:
Rebased onto v2.0-cyanoptera (was 804 behind, now 0). No conflicts; the four commits replay unchanged and the net diff is identical.

Taking it out of draft ran the full suite for the first time, and one job is red: Thread Sanitizer reports a data race at `src/storage/arena_allocator.cpp:120` in `ArenaAllocator::AlignNext()`, on `test/sql/show_select/describe_subquery.test` (6289 passed, 1 failed; it retried twice within the run and failed both times).

I don't want to claim it's unrelated, so here is what I can and can't establish:

- Thread Sanitizer passed on the base commit e3946f2327 (run 33934596285 - the run was cancelled overall, but the threadsan job itself concluded success), and failed on this branch's head.
- That is one observation on each side of a nondeterministic detector, so it does not settle much either way.
- This branch touches only `src/parser/peg/transformer/transform_expression.cpp` and a new test file under `test/sql/parser`. The race is in the storage layer and the failing test is neither of mine.
- Thread Sanitizer was skipped on the two earlier runs of this same commit, so the failing run is its first execution here.

I can't re-run the job myself. If someone can re-run just `threadsan` a couple of times, that would tell us whether it reproduces or whether I got unlucky.

On the substance, the coalesce gap and Maxxen's macro suggestion are answered in my comment above from 4 September - short version: a CASE-based macro double-evaluates its first argument, and macros can't be variadic, so coalesce can't move to the catalog without losing unbounded arity. The change here lowers both through the same shared helper, so `.coalesce()` works at any arity.

marcodelpin:
Rebased onto v2.0-cyanoptera (was 804 behind, now 0). No conflicts; the four commits replay unchanged and the net diff is identical.

Taking it out of draft ran the full suite for the first time, and one job is red: Thread Sanitizer reports a data race at `src/storage/arena_allocator.cpp:120` in `ArenaAllocator::AlignNext()`, on `test/sql/show_select/describe_subquery.test` (6289 passed, 1 failed; it retried twice within the run and failed both times).

I don't want to claim it's unrelated, so here is what I can and can't establish:

- Thread Sanitizer passed on the base commit e3946f2327 (run 33934596285 - the run was cancelled overall, but the threadsan job itself concluded success), and failed on this branch's head.
- That is one observation on each side of a nondeterministic detector, so it does not settle much either way.
- This branch touches only `src/parser/peg/transformer/transform_expression.cpp` and a new test file under `test/sql/parser`. The race is in the storage layer and the failing test is neither of mine.
- Thread Sanitizer was skipped on the two earlier runs of this same commit, so the failing run is its first execution here.

I can't re-run the job myself. If someone can re-run just `threadsan` a couple of times, that would tell us whether it reproduces or whether I got unlucky.

On the substance, the coalesce gap and Maxxen's macro suggestion are answered in my comment above from 4 September - short version: a CASE-based macro double-evaluates its first argument, and macros can't be variadic, so coalesce can't move to the catalog without losing unbounded arity. The change here lowers both through the same shared helper, so `.coalesce()` works at any arity.

lnkuiper:
Thanks for the changes! A few more comments:

1. This edge case regresses:

   ```sql
   CREATE SCHEMA s;
   CREATE MACRO "coalesce"(a, b) AS 999;
   CREATE MACRO s.coalesce(a, b) AS 777;

   SELECT "coalesce"(NULL, 5), s.coalesce(NULL, 5);
   ```

   The base returns `999, 777`; this branch returns `5, 5` because both calls become `OPERATOR_COALESCE` expressions.

2. These queries now succeed:

   ```sql
   SELECT (NULL).coalesce(DISTINCT 5);
   SELECT (NULL).coalesce(5 ORDER BY 1);
   ```

   The equivalent plain `COALESCE` forms are parser errors. I think the chained forms should reject these modifiers rather than silently ignore them.

marcodelpin:
Both points from your September 7 comment are addressed in 4000683000:

1. With the macros from your example, `"coalesce"(NULL, 5), s.coalesce(NULL, 5)` returns `999, 777` again. The shared rewrite helper now takes the call form and applies the `coalesce` rewrite only to method chains, restoring the base's rewrite set for plain calls. The macro regression case is in `test/sql/parser/parser_rewrites_method_chain.test`.
2. `(NULL).coalesce(DISTINCT 5)` and `(NULL).coalesce(5 ORDER BY 1)` now produce parser errors. The tests cover both modifiers for chained `coalesce` and `ifnull`. Modifier rejection applies only when a rewrite fires; an ordinary chained call retains its modifiers, covered by a `string_agg` case.

The two later commits touch only tests: d02dd4d9e0 removes the `.if()` cases after upstream made `if` a catalog macro, and 2809bd8691 corrects three test comments.

On the current head, 2809bd8691, Main run 35323137493 completed successfully with 55 jobs and zero failures. The branch has no merge conflicts but is 25 commits behind the v2.0-cyanoptera base tip.

@lnkuiper, could you take another look when you have a moment?

PR #25141, opened 2026-08-29
Move SigV4 signing logic from httpfs into core

staticlibs:
This PR moves the implementation of the [Signature Version 4](https://docs.aws.amazon.com/AmazonS3/latest/developerguide/sig-v4-authenticating-requests.html) from `duckdb-httpfs` into `HTTPUtil` in core.

Implementation of a few hash functions is also moved from `httpfs` to core as a required dependency to SigV4 generation.

There are no functional changes, generating logic and its usage in `httpfs` remains the same as before.

Ref: duckdblabs/duckdb-internal#5731

### DIFF
--- a/.github/patches/extensions/httpfs/0004-hash-functions-core.patch
+++ b/.github/patches/extensions/httpfs/0004-hash-functions-core.patch
@@ -1,0 +1,70 @@
+diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
+index ebc01e0..cd50fb9 100644
+--- a/src/CMakeLists.txt
++++ b/src/CMakeLists.txt
+@@ -1,5 +1,4 @@
+-set(HTTPFS_SOURCES hffs.cpp hash_functions.cpp create_secret_functions.cpp
+-                   httpfs_extension.cpp)
++set(HTTPFS_SOURCES hffs.cpp create_secret_functions.cpp httpfs_extension.cpp)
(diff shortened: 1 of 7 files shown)


staticlibs:
Clang-tidy failure:

```c++
--- clangd-tidy stdout tail (7 lines) ---
src/common/hash_functions.cpp:17:59: Error: The parameter 'message' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
   15 |  }
   16 |  
   17 |  void hmac256(EncryptionUtil &encryption_util, std::string message, hash_bytes secret, hash_bytes &out) {
      |                                                            ^~~~~~~
   18 |  	hmac256(encryption_util, message, secret, sizeof(hash_bytes), out);
   19 |  }
--- end clangd-tidy stdout tail ---
```

I am not familiar with the motivation behind the initial impl of the `hmac256` function, but given that it declares both const and no-const version of it - I would think passing by value here may have been intentional:

```c++
void hmac256(EncryptionUtil &encryption_util, const std::string &message, const_data_ptr_t secret, idx_t secret_len, hash_bytes &out);

void hmac256(EncryptionUtil &encryption_util, std::string message, hash_bytes secret, hash_bytes &out);
```

lnkuiper:
Hey, I'm down to move this into core, but we need the tidy check to succeed before this can be merged. Could you have a look?

PR #25142, opened 2026-08-29
fix: list_sort orders INTERVALs consistently with comparison/ORDER BY

### LINKED ISSUE #25108: list_sort orders INTERVALs inconsistently with comparison operators and ORDER BY
### What happens?

`list_sort` orders INTERVALs inconsistently with comparison operators and `ORDER BY`.

`create_sort_key` encodes intervals by raw `(months, days, micros)` without the normalization (1 month = 30 days, 1 day = 24 h) that comparisons, `ORDER BY`, aggregates and window functions apply. This affects `list_sort`, `list_reverse_sort`, `list_grade_up`, `list_distinct` and any other `create_sort_key` consumer.

### To Reproduce

```sql
SELECT INTERVAL '31 days' > INTERVAL '1 month';
-- true

SELECT i FROM (VALUES (INTERVAL '31 days'), (INTERVAL '1 month')) t(i) ORDER BY i;
-- 1 month, 31 days

SELECT list_sort([INTERVAL '31 days', INTERVAL '1 month']);
-- [31 days, 1 month]   <- disagrees with both

SELECT create_sort_key(INTERVAL '31 days', 'ASC NULLS LAST') < create_sort_key(INTERVAL '1 month', 'ASC NULLS LAST');
-- true (raw field order), although INTERVAL '31 days' > INTERVAL '1 month'
```

### OS:

macOS 15 (arm64)

### DuckDB Version:

v1.6.0-dev13791 (built from source at 8616efa9da (latest main at time of filing))

### DuckDB Client:

CLI / C++ API

### Hardware:

Apple M4

### Full Name:

peterxcli

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

---

Found by a RapidCheck property-based test harness ([peterxcli/duckdb#4](https://github.com/peterxcli/duckdb/pull/4)).

thatssoheil:
## What

Fixes #25108: `list_sort` (and `list_reverse_sort`, `list_grade_up`, `list_distinct`, `create_sort_key`) orders `INTERVAL`s inconsistently with comparison operators and `ORDER BY`.

## Root cause

`create_sort_key` encodes the raw `interval_t` struct `(months, days, micros)` with no normalization. But `interval_t::operator>` (used by comparisons, `ORDER BY`, aggregates, window functions) **normalizes** first — carrying 1 month = 30 days and 1 day = 24h into a canonical form with `days in [0, 30)` and `micros in [0, MICROS_PER_DAY)` — then compares `(months, days, micros)` lexicographically.

So the two orderings disagree. E.g.:

```
SELECT INTERVAL '31 days' > INTERVAL '1 month';   -- true
SELECT list_sort([INTERVAL '31 days', INTERVAL '1 month']);
-- before: [31 days, 1 month]  (disagrees)
```

## Fix

Encode the **normalized** `interval_t` in the sort key via a dedicated `SortKeyIntervalOperator`. After normalization the Radix byte-comparison order of `(months, days, micros)` matches the lexicographic comparison `interval_t::operator>` uses, so all `create_sort_key` consumers agree with comparison/`ORDER BY`. The key length is unchanged (`sizeof(interval_t)`), and `interval_t::operator==` likewise normalizes, so the sort-key decode round-trip asserted in `create_sort_key.cpp` still holds.

## Tests

`test/sql/function/list/list_sort_interval.test` asserts:

- `ORDER BY` -> `[1 month, 31 days]`
- `list_sort` -> `[1 month, 31 days]`
- `list_reverse_sort` -> `[31 days, 1 month]`
- `list_grade_up` -> `[2, 1]`
- `list_distinct` -> `[1 month, 1 month 1 day]`
- `create_sort_key('31 days') > create_sort_key('1 month')` -> `true`

Verified: ran the new test + the existing `list_sort_vector_types`, `list_sort_having`, `list_distinct`, `list_intersect`, `create_sort_key`, and all `test/sql/types/interval/*.test` suites (incl. `test_interval_comparison.test`) — all passing, no regressions.

Fixes #25108.

### DIFF
--- a/extension/core_functions/scalar/list/list_sort.cpp
+++ b/extension/core_functions/scalar/list/list_sort.cpp
@@ -9,6 +9,7 @@
 #include "duckdb/planner/expression/bound_reference_expression.hpp"
 #include "duckdb/main/config.hpp"
 #include "duckdb/common/sorting/sort.hpp"
+#include "duckdb/planner/expression_binder.hpp"
 #include "duckdb/parallel/thread_context.hpp"
 
 namespace duckdb {
@@ -48,6 +49,13 @@ ListSortBindData::ListSortBindData(OrderType order_type_p, OrderByNullType null_
 	// get the BoundOrderByNode
 	auto idx_col_expr = make_uniq_base<Expression, BoundReferenceExpression>(LogicalType::USMALLINT, 0U);
 	auto lists_col_expr = make_uniq_base<Expression, BoundReferenceExpression>(child_type, 1U);
+	// Normalize the sort key without changing the sorted values (#25108): push the
+	// type's collation (for INTERVAL this wraps the expression in
+	// normalized_interval(...), the same normalization comparison operators, ORDER BY
+	// and aggregates apply) onto the key expression BEFORE constructing its
+	// BoundOrderByNode. The key then byte-compares like ORDER BY, while the list
+	// payload keeps the original (non-normalized) values.
+	ExpressionBinder::PushCollation(context, lists_col_expr, child_type);
 	vector<BoundOrderByNode> orders;
 	orders.emplace_back(OrderType::ASCENDING, OrderByNullType::ORDER_DEFAULT, std::move(idx_col_expr));
 	orders.emplace_back(order_type, null_order, std::move(lists_col_expr));
--- a/src/function/scalar/create_sort_key.cpp
+++ b/src/function/scalar/create_sort_key.cpp
@@ -239,6 +239,53 @@ struct SortKeyConstantOperator {
 	}
 };
 
+struct SortKeyIntervalOperator {
+	using TYPE = interval_t;
+
+	static idx_t GetEncodeLength(TYPE input) {
+		return sizeof(interval_t);
+	}
+
+	template <bool FLIP_BYTES>
+	static idx_t Encode(data_ptr_t result, TYPE input) {
+		// Encode the NORMALIZED interval. interval_t::operator> (used by comparisons,
+		// ORDER BY and aggregates) normalizes the interval (1 month = 30 days, 1 day = 24h,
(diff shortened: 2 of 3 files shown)


lnkuiper:
We should push the interval collation onto `lists_col_expr` before constructing its `BoundOrderByNode` (via `ExpressionBinder::PushCollation`). That makes the key normalized but preserves the original list payload (the current state of this PR changes the values when the sort key is encoded/decoded).

thatssoheil:
Addressed in 6e9f77d1ad: `create_sort_key.cpp` is reverted to base (byte-identical to 8616efa9da) and the interval collation is pushed onto `lists_col_expr` via `ExpressionBinder::PushCollation` before its `BoundOrderByNode` is constructed - the same pattern the ORDER BY binder applies. The sort key is normalized (via `normalized_interval`), while the list payload keeps the original values.

Verified:
- `list_sort`/`list_reverse_sort`/`list_grade_up` now agree with `ORDER BY` on the regression cases, including negatives (`-31 days` vs `-1 month`, the case where raw encoding diverges from normalized) and NULLS FIRST/LAST
- `list_distinct` round-trips original values (`[1 month, 31 days]`, previously re-displayed as normalized forms)
- equality with `ORDER BY` checked over 1000 randomized intervals (day/microsecond/mixed/negative), 0 mismatches
- TIMETZ/BIT lists sort consistently with ORDER BY (PushCollation is the same path ORDER BY uses)
- test/sql/function/list/* (111 cases), test/sql/types/interval/* (9), create_sort_key.test (204 assertions), test/sql/order/* all pass

New tests added at the tip: negative-interval ordering (ASC/DESC vs ORDER BY), NULLS FIRST/LAST placement, and the key-normalized/payload-preserved case.

PR #25144, opened 2026-08-29
Prune row group for boolean column expression filter

dentiny:
Hi team, I found row group pruning doesn't always work as expected for expression filters against boolean column; for example,
```sql
memory D ATTACH '/tmp/bool_ref_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE tb AS SELECT
       CASE WHEN range // 2048 = 0 THEN TRUE
            WHEN range // 2048 = 1 THEN FALSE
            ELSE (range % 2 = 0) END AS b,
       range::INTEGER AS x
     FROM range(6144);

-- prune as expected
db D EXPLAIN ANALYZE SELECT sum(x) FROM tb WHERE b = true;
╭─ Summary ───────────╮
│ Total Time: 0.0012s │
╰─────────────────────╯
╭─ Projection ───┚┖───────────────╮
│ Projections: x                  │
│ 3,072 rows                  0µs │
│ Table: db.main.tb               │
│ Projections: x                  │
│ Filters: b = true               │
│ Row Groups Scanned: 2 / 3       │
│ 3,072 rows                  0µs │

-- doesn't prune, access all row groups
db D EXPLAIN ANALYZE SELECT sum(x) FROM tb WHERE b;
╭─ Summary ───────────╮
│ Total Time: 0.0004s │
╰─────────────────────╯
╭─ Projection ───┚┖───────────────╮
│ Projections: x                  │
│ 3,072 rows                  0µs │
│ Table: db.main.tb               │
│ Projections: x                  │
│ Filters: b                      │
│ 3,072 rows                  0µs │
```


src/planner/filter/expression_filter.cpp
```diff
@@ -609,6 +609,28 @@ static FilterPropagateResult CheckInOperatorStatistics(optional_ptr<ClientContex
...
+	const auto &stats = input_stats[index];
+	const auto min_v = NumericStats::Min(stats).GetValue<bool>();
+	const auto max_v = NumericStats::Max(stats).GetValue<bool>();
+	if (min_v != max_v) {
+		return FilterPropagateResult::NO_PRUNING_POSSIBLE;
+	}
+	if ((negated ? !min_v : min_v) == false) {
+		return FilterPropagateResult::FILTER_ALWAYS_FALSE;
+	}
+	return stats.CanHaveNull() ? FilterPropagateResult::NO_PRUNING_POSSIBLE : FilterPropagateResult::FILTER_ALWAYS_TRUE;
```
Mytherin: This should be returning `FILTER_TRUE_OR_NULL` / `FILTER_FALSE_OR_NULL` (and then maybe we need to patch some code to deal with this upstream).

Mytherin (commented):

PR #25145, opened 2026-08-29
[Dev] Use `Event` for streaming result collection readiness signaling

Tishj:
Previously, to indicate that the result is "ready" and the `StreamQueryResult` can be created, we used this logic:
```c++
bool Executor::ResultCollectorIsBlocked() {
	if (!HasStreamingResultCollector()) {
		return false;
	}
	for (auto &kv : to_be_rescheduled_tasks) {
		auto &task = kv.second;
		if (task->TaskBlockedOnResult()) {
			return true;
		}
	}
	return false;
}
```
```c++
bool PipelineTask::TaskBlockedOnResult() const {
	return pipeline.IsStreamingResultPipeline() && pipeline_executor->RemainingSinkChunk();
}
```
```c++
bool Pipeline::IsStreamingResultPipeline() const {
	if (external_streaming_result_producer) {
		return true;
	}
	return sink && sink->type == PhysicalOperatorType::RESULT_COLLECTOR &&
	       sink->Cast<PhysicalResultCollector>().IsStreaming();
}
```
Which basically comes down to: "Are we using a streaming result, and is there at least 1 task that got blocked trying to execute a Sink to the last pipeline (or is externally feeding that last pipeline)"

This has always felt hacky to me, even when I initially introduced it, but I saw no better way of doing it.

This PR changes this, using the `Event` system to signal this instead. (which funnily enough already existed when this was first created).

We introduce a `PipelineResultReadyEvent`, which is placed between the pipeline events that already exist, stealing the dependencies of the last pipeline.

When the dependencies finish, that causes `PipelineResultReadyEvent::Schedule` to get called, this flips the newly introduced `atomic<bool> result_ready` to true (taking the place of the previously existing `ResultCollectorIsBlocked()` call).

This causes the caller to get notified that the "result collector is ready" the next time they call `ExecuteTask`. Upon receiving that, `FetchResult` will get called.
`ClientContext::FetchResultInternal` will then cause the `PipelineResultReadyEvent` to `Finish()`, finally scheduling the last pipeline.

- `TaskBlockedOnResult` is removed, because we no longer rely on this to signal readiness anymore.
- `ResultCollectorIsBlocked` is removed, same reason ^.
- `PipelineTask::GetPipelineExecutor` was already unused, also removed.
- `external_streaming_result_producer` also disappears (this was part of `ResultCollectorIsBlocked` logic).

### Note

As a result of this rework, errors that occur in the last pipeline don't get surfaced until `Fetch` is called, currently those surface earlier on if they occur in the first chunk.
Because instead of letting the final pipeline run until it hits Sink, we don't run it at all now, until `Fetch` is called.

This blocking behavior was dependent on the data (based on the buffer size, a streaming result might never block, if its result size is small enough to fit in the buffer size)
Now that doesn't happen anymore, because the last pipeline doesn't start at all until the query result is created.


src/parallel/executor.cpp
```diff
@@ -78,6 +79,89 @@ static shared_ptr<Event> CreatePipelineScheduleEvent(const PipelineScheduleStage
...
+	vector<bool> visited(schedule.stages.size(), false);
+	vector<idx_t> to_visit = schedule.stages[stage_idx].dependencies;
+	while (!to_visit.empty()) {
+		auto dependency = to_visit.back();
+		to_visit.pop_back();
+		if (visited[dependency]) {
+			continue;
+		}
+		visited[dependency] = true;
+		if (streaming_execute_stages[dependency]) {
```
Tishj: Hmm, this might be problematic, if `dependency` is an index outside of `streaming_execute_stages.size()`
But `schedule.stages` is a flat vector that should contain all stages, and `streaming_execute_stages` is initialized to the same size as that, so that should be impossible.

That being said, might make sense to just use a `ValidityMask`, rather than a `vector<bool>`

Tishj:
It's debatable whether we want to make this change for streaming query results 🤔 
I wonder if Quack might prefer this though? @guillesd

Tishj:
@Mytherin what are your thoughts on this? I've always felt that how this gets signaled is brittle and hacky. Might just be that I'm biased against it / overly critical because I wrote it ?
In any case, this method of stalling the final result collector is (or can be made) much more generic. The only tradeoff is that we do lose the fast-path for smaller result sets with this.

src/main/client_context.cpp
guillesd: There is a problem with this condition. For custom streaming collectors like Quack, this new gate that you introduce will mean Quack (or any other custom streaming collector) won't produce any results. The thing is `create_stream_result` only takes a look at whether the client is requesting a `StreamingQueryResult`, not at whether the result collector is streaming. I would consider changing the gate to:

```cpp
bool create_stream_result = prepared.properties.output_type == QueryResultOutputType::ALLOW_STREAMING && pending.allow_stream_result;
bool keep_result_open = create_stream_result || executor.HasStreamingResultCollector();

unique_ptr<QueryResult> result;
D_ASSERT(executor.HasResultCollector());
result = executor.GetResult();
if (!keep_result_open) {
      CleanupInternal(lock, result.get(), false);
} else {
      active_query->SetOpenResult(*result);
      executor.OpenResultPipeline();
}
```

Tishj: Thanks, that's a good point 👍 
That should probably go in regardless of this PR stranding or not

Tishj:
@evertlammerts as requested, this is the PR we talked about 👍

Tishj:
I like the idea of producing tasks based on the state of the buffer, to better control the pressure.
We'll have to see how impactful that change will be on the current `Event` system. I believe currently the task creation is always synchronous, all tasks are produced and added to the queue in the `Schedule` of the `Event`.

Mytherin:
Thanks for the PR! Sounds great - nice to integrate this into the pipeline system.

I wonder a bit about this:

> As a result of this rework, errors that occur in the last pipeline don't get surfaced until Fetch is called, currently those surface earlier on if they occur in the first chunk.
Because instead of letting the final pipeline run until it hits Sink, we don't run it at all now, until Fetch is called.

I think it might be nicer to not treat results as special in this manner, but to instead integrate results into the regular pipeline infrastructure fully instead. My intuition here is that we would just want a `ResultSink` (or similar) as the root operator, which partially materializes the result (using streaming query result semantics). The user can then call `Fetch()` to get that streaming result and start retrieving rows. If the user wants a different result (e.g. materialized result) on calling `Fetch` we can promote the result sink to fully materialize everything and then return.

Tishj:
Closing in favor of #25257

PR #25146, opened 2026-08-29
Propagate stats for sum aggregation function

dentiny:
Hi team, I found current implementation doesn't propagate stats for `sum` aggregates; for example
```sql
memory D CREATE TABLE t AS SELECT range::INTEGER AS x FROM range(10000);
memory D EXPLAIN SELECT sum(x) FROM t HAVING sum(x) < 0;
╭─ Filter ────────────────────────╮
│ Expression: sum(x) < 0          │
│ ~0 rows                         │
╰────────────────┬────────────────╯
│ ~1 row                          │
╰────────────────┬────────────────╯
╭─ Projection ───┴────────────────╮
│ Projections: x                  │
│ ~10,000 rows                    │
╰────────────────┬────────────────╯
╭─ Seq Scan ─────┴────────────────╮
│ Table: memory.main.t            │
│ Projections: x                  │
│ ~10,000 rows                    │
```
The above query should be able to prune all row groups, but we're accessing every row groups.

The reason is 
- We do have stats propagation callback for `sum` aggregate, but it's only to [replace a fast sum implementation](https://github.com/duckdb/duckdb/blob/8616efa9da9921b9111fe46373af7936a5d96d16/extension/core_functions/aggregate/distributive/sum.cpp#L280-L313) when possible
- BUT the calculates stats is never returned -- all paths are returning NULL

This PR does a few things:
- The fast sum aggregation is kept for INT16/INT32 types when possible
- Propagated the calculated stats so they could be used to prune
- Expand the sum aggregate type from limited ones, to all possible types, except HUGEINT

### DIFF
--- a/extension/core_functions/aggregate/distributive/sum.cpp
+++ b/extension/core_functions/aggregate/distributive/sum.cpp
@@ -279,37 +279,58 @@ AggregateFunction GetSumAggregateNoOverflowDecimal() {
 
 unique_ptr<BaseStatistics> SumPropagateStats(ClientContext &context, BoundAggregateExpression &expr,
                                              AggregateStatisticsInput &input) {
-	if (input.node_stats && input.node_stats->has_max_cardinality) {
-		auto &numeric_stats = input.child_stats[0];
-		if (!NumericStats::HasMinMax(numeric_stats)) {
-			return nullptr;
-		}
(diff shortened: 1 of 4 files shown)


Mytherin:
Thanks!

PR #25147, opened 2026-08-30
Check simple load and autoload of extensions

carlopi:
Iterates on https://github.com/duckdb/duckdb/pull/24777, by making so that:

* Explicit `LOAD <name>` can only ever accept core or community keys (this is on par with 1.5.5, but was relaxed by the PR in non standard setups)
* Autoload can only ever accept core keys (this is stricter than 1.5.5, where also community keys were accepted IF name were to match, and that relies on no collisions being merged)

I think this avoid some potential issues / unclarity of misuse, since issuing a `FROM 's3://bucket/path/to/file.parquet';` will autoload httpfs only if:
1. autoloading is enabled
2. the current local state does have a core signature extension at `httpfs.duckdb_extension`

Similarly, if extension is not available, autoinstall will also require a core signature on the binary.

----

Note `LOAD httpfs FROM my_repo;` would still allow that to override `httpfs` behaviour, but that needs then to be explicitly issued.

Note that the following cases have also been considered (and keeps working as intended):
* re-distributing core-signed extensions via a different repository works
* those restrictions are available only when signature checks are performed, so unsigned local development tests are kept functional

### DIFF
--- a/src/include/duckdb/main/extension_helper.hpp
+++ b/src/include/duckdb/main/extension_helper.hpp
@@ -182,6 +182,13 @@ public:
 	static vector<string> GetTrustedPublicKeys(DatabaseInstance &db, ExtensionRepositoryType repository_type,
 	                                           const string &repository_name);
 
+	//! The origin whose signing keys a load trusts. Without an explicit FROM (a bare load, which is also what
+	//! autoloading performs) only the fixed core and community keys are trusted: a community extension keeps its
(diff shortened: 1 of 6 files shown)


hannes:
This sounds good to me

lnkuiper:
Seems good can we tone down the verbose comments

Mytherin:
Thanks!

PR #25149, opened 2026-08-30
Do not corrupt the transaction lists when an allocation fails

carlopi:
Two failure paths, both reached by failing an allocation during a query:

MetaTransaction::GetTransaction registered a database in all_transactions and then inserted it into transactions. When that insert failed to allocate, the two were left out of sync, and MetaTransaction::Rollback - which walks all_transactions and looks each entry up in transactions - dereferenced the end iterator. That is a segfault in release; the D_ASSERT that would have caught it is compiled out. Reserve first, then insert, then push_back into the reserved space, so a failure happens before either is modified.

DuckTransactionManager::RemoveTransaction is noexcept and allocated: it moves the transaction out of active_transactions before re-homing it in recently_committed_ transactions or in the cleanup info, so an allocation failure in between terminated the process and lost the transaction. Reserving inside the function does not fix this - reserve allocates too, and it is still noexcept. Move the allocation to the callers, which already hold transaction_lock and may throw: CreateCleanupInfo allocates the cleanup info and reserves the space for both lists, leaving RemoveTransaction genuinely allocation-free.

Between them these accounted for eight tests that crashed under allocation failure injection (two segfaulting in MetaTransaction::Rollback, three aborting in RemoveTransaction, and three more whose stacks were too corrupt to attribute); all eight now survive 36 injected runs each.

### DIFF
--- a/src/include/duckdb/transaction/duck_transaction_manager.hpp
+++ b/src/include/duckdb/transaction/duck_transaction_manager.hpp
@@ -96,10 +96,17 @@ protected:
 private:
 	//! Generates a new commit timestamp
 	transaction_t GetCommitTimestamp();
+	//! Allocates the cleanup info, and reserves the space RemoveTransaction needs to re-home a transaction.
+	//! RemoveTransaction is noexcept, so it cannot do this itself: it must not allocate at all. Call this with
+	//! transaction_lock held, immediately before RemoveTransaction, so that a failure to allocate is reported
+	//! while the transaction lists are still untouched.
+	unique_ptr<DuckCleanupInfo> CreateCleanupInfo();
 	//! Remove the given transaction from the list of active transactions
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks - I kind of wonder if we should just remove `noexcept` from these functions though. It's (unfortunately) pretty hard to guarantee a function is actually `noexcept` in C++ so it might be better to just patch the upstream callsites to deal with exceptions.

carlopi:
I am not sure, this was a limited fix in response to a crash I observed (while messing with allocators).

In general, I think `noexcept` is cool for functions called in destructors, possibly as part of RAII, and I feel propagating `noexcept` does keep some information that is otherwise hard to get back.

I would think for now I'll keep as is, but maybe going forward a discussion on what's the proper path forward would be great.

PR #25150, opened 2026-08-30
Remove `BoundExpression`

Maxxen:
This PR attempts to remove the `BoundExpression` class, replacing it with a dedicated `BoundExpressionMap` that lives within the `Binder`.

The `BoundExpression` has been a long standing quirk in the query execution frontend: Its a "fake" `ParsedExpression` that secretly holds a bound `Expression`, whose only purpose was to represent a partially or already bound part of the expression tree. The way this worked was that during binding, the binder mutated the parsed tree in place, replacing each successfully bound `ParsedExpression` node with a `BoundExpression` wrapper so that re-binding passes (notably the correlated-subquery flow, where the same tree is re-bound against outer scopes at increasing depths) could skip already-bound parts and just "unwrap" the result.

This had several problems:

- It corrupted the parsed tree. Binding destroyed the original parse tree as a side effect. Anything that wanted to re-bind, print, copy, or serialize the original expression had to work around the wrappers or make a lot of expression copies. This was  also one of the main blockers preventing us from making the `ParsedExpression` tree fully immutable.
- It punched a hole in the parser/planner layering. A planner-owned `Expression` traveled inside a `ParsedExpression`, so `BoundExpression` needed stub implementations of `Copy`/`Serialize`/`Equals`/`Hash` that mostly just threw, and `ExpressionClass::BOUND_EXPRESSION` polluted the parser-level enum. The `ParsedExpression` and `BoundExpression` does not actually need to have any inheritance relation anymore.
- This quirk also leaked outside the binder. E.g. the CSV writer built parsed trees with BoundExpression nodes smuggled in just to get pre-bound columns through the regular binder, and similarly the window binding passed parsed `OrderByNodes` into function bind callbacks for the same reason.

**The change in this PR:**

There's should be no visible change in behavior, this PR just changes the expression binding by replacing the in-place `BoundExpression` substitution with a dedicated `BoundExpressionMap`, living in the `Binder`:

- The map keys parsed nodes by identity and stores their bound `Expression`. On a successful bind, `ExpressionBinder::Bind` inserts into the map instead of replacing the node; the "already bound, skip" check becomes `IsBound(node)` instead of a class check, and results are retrieved with `Get`/`GetMutable` or extracted with `Consume`. The parsed tree itself is left untouched.
- Entry lifetime is managed by an RAII `BoundExpressionScope` opened per bind cycle: entries not consumed by the end of the cycle (e.g. left behind by failed bind attempts) are erased on scope exit, including when unwinding on a binder error, so no entry can outlive its parsed node. In `DEBUG` builds we also stamp entries with the node's expression class/type to catch stale entries under pointer reuse.
- `BoundExpression` and the `ExpressionClass::BOUND_EXPRESSION` are deleted outright.

Downstream cleanups that fall out of this:

- Expression binders (bind_case, bind_between, bind_operator, bind_window, bind_aggregate, bind_function, etc.) now pull children from the map instead of casting-and-unwrapping BoundExpression nodes.
- FunctionBinder::BindWindowFunction/ResolveFunction take vector<LogicalType> order types instead of parsed OrderByNodes, since bound order expressions no longer masquerade as parsed ones.
- The CSV writer builds bound expressions (BoundReferenceExpression, BoundConstantExpression, FunctionBinder::BindScalarFunction) directly rather than constructing a fake parsed tree with smuggled BoundExpression nodes and a throwaway binder.
- USING/NATURAL join columns and window RANGE frame handling seed/consume the map explicitly, with scopes marking the boundaries.

In conclusion: binding no longer mutates the parsed representation, the parser layer no longer knows about bound expressions, and the "have we bound this yet" bookkeeping lives in a separate structure instead of being encoded in the tree nodes themselves.

In the future, if we make the `Expression` class itself fully immutable, we could change its child pointers from `unique_ptr` to `shared_ptr`, and then simply just memoize/cache _everything_ we bind, to deduplicate expressions as well (although we need to handle `QueryLocation` differently.


Mytherin:
Very cool! Thanks

PR #25151, opened 2026-08-30
Fix out-of-bounds path offset for file:// URIs without a path

### LINKED ISSUE #25083: Vuln50: Heap-buffer-overflow in Path::ParseFileSchemes via off-by-one path_offset for `file://<authority>` URIs
### What happens?

`Path::ParseFileSchemes` in `src/common/path.cpp` computes the consumed prefix length of a `file://<authority>` URI by summing `scheme.size() + authority.size() + anchor.size()` instead of using the length actually returned by `ParseURIScheme`. When the input has no trailing path after the authority, `ParseURIScheme` synthesizes a virtual `/` anchor that was never actually consumed from the input, so this sum overcounts the true offset by one byte. Neither `ParseFilePathTail` nor `Path::NormalizeSegments` nor `SegmentAndNormalizePath` validates that the resulting `path_offset` stays within the string bounds, so `NormalizeSegments` constructs a `begin` iterator past `raw.end()` and `SegmentAndNormalizePath` dereferences it, causing AddressSanitizer to catch a heap-buffer-overflow read one byte past the string allocation.

### To Reproduce

This is reproducible with pure SQL from the duckdb CLI, no input file needs to be constructed.

```sql
SELECT path_join('file://localhost');
```

```shell
./build/debug/duckdb -c "SELECT path_join('file://localhost');"
```

```
=================================================================
==3010034==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x5030004a6e41 at pc 0x6438cb80d00f bp 0x7ffc6f919570 sp 0x7ffc6f919560
READ of size 1 at 0x5030004a6e41 thread T0
    #0 0x6438cb80d00e in SegmentAndNormalizePath src/common/path.cpp:31
    #1 0x6438cb8127fd in duckdb::Path::NormalizeSegments(...) src/common/path.cpp:224
    #2 0x6438cb80d656 in duckdb::Path::FromString(...) src/common/path.cpp:72
    #3 0x6438d4207992 in ProcessRow src/function/scalar/string/path_join.cpp:39
    #4 0x6438d4208328 in PathJoinFunction src/function/scalar/string/path_join.cpp:59

0x5030004a6e41 is located 0 bytes after 17-byte region [0x5030004a6e30,0x5030004a6e41)
allocated by thread T0 here:
    #2 0x6438c97d453a in duckdb::string_t::GetString[abi:cxx11]() const src/include/duckdb/common/types/string_type.hpp:126
    #3 0x6438d42078d8 in ProcessRow src/function/scalar/string/path_join.cpp:37

SUMMARY: AddressSanitizer: heap-buffer-overflow src/common/path.cpp:31 in SegmentAndNormalizePath
==3010034==ABORTING
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

alonfaraj:
Fixes #25083

`Path::ParseFileSchemes` computed the consumed prefix of a `file://<authority>` URI by summing `scheme.size() + authority.size() + anchor.size()`. `ParseURIScheme` synthesizes `anchor = "/"` even when nothing follows the authority, so for `file://localhost` the sum overcounts by one, `path_offset` lands past the end of the string, and `SegmentAndNormalizePath` reads out of bounds - reachable from SQL via `SELECT path_join('file://localhost')`.

`ParseURIScheme` already returns the correct offset; use it instead of the sum, and assert `path_offset <= raw.size()` in `NormalizeSegments`. Only the broken case changes behavior: `file://localhost` now parses as `file://localhost/`.

### DIFF
--- a/src/common/path.cpp
+++ b/src/common/path.cpp
@@ -221,6 +221,7 @@ void Path::NormalizeSegments(const string &raw, size_t path_offset) {
 	}
 
 	vector<string> raw_segs;
+	D_ASSERT(path_offset <= raw.size());
 	SegmentAndNormalizePath(raw.begin() + static_cast<string::difference_type>(path_offset), raw.end(), raw_segs);
 	segments.clear();
 	for (auto &seg : raw_segs) {
@@ -298,12 +299,12 @@ size_t Path::ParseFileSchemes(const string &input, Path &parsed) {
 		parsed.anchor = "/";
 		path_begin = 8;
 	} else if (/* file:// */ input_len >= 7 && input[6] == '/') {
-		ParseURIScheme(input, parsed);
+		// NOTE: the anchor may be synthesized, so the offset must come from ParseURIScheme, not from the sizes
+		path_begin = ParseURIScheme(input, parsed);
 		if (StringUtil::Lower(parsed.authority) != "localhost") {
 			throw InvalidInputException("Path: file:// scheme only supports localhost authority, got: %s",
 			                            parsed.authority);
 		}
-		path_begin = parsed.scheme.size() + parsed.authority.size() + parsed.anchor.size();
 	} else /* file:/ */ {
 		parsed.scheme = "file:";
 		parsed.anchor = "/";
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks!

PR #25153, opened 2026-08-30
Fix use-after-free on EXECUTE of prepared table functions that bind a catalog entry

### LINKED ISSUE #25075: Vuln36: Stale prepared-statement physical plan caches raw CatalogEntry reference across DROP TABLE, causing UAF on EXECUTE
### What happens?

`PragmaTableInfoBind()` in `src/function/table/system/pragma_table_info.cpp` stores a raw, non-owning `CatalogEntry &` for the target table inside `PragmaTableFunctionData`, but unlike `bind_basetableref.cpp` it never calls `RegisterDBRead()`, so `StatementProperties::read_databases` stays empty for a `PREPARE`d statement that reads through `pragma_table_info()`. Because that set is empty, both the rebind-forcing guard in `Optimizer::Optimize()` and `PreparedStatementData::RequireRebind()` treat the statement as safe to skip, so `PhysicalPlanGenerator::CreatePlan(LogicalPrepare&)` caches the physical plan and later `PhysicalPlanGenerator::CreatePlan(LogicalExecute&)` reuses it without any catalog-identity check. If the underlying table is dropped between two `EXECUTE`s, the catalog's undo-buffer cleanup frees the `DuckTableEntry` the cached plan still references, and the next `EXECUTE` dereferences it in `PragmaTableInfoFunction()`, causing a heap use-after-free read.

### To Reproduce

This is reproducible with pure SQL fed directly to the DuckDB CLI, no input file needs to be constructed.

```sql
CREATE TABLE t(a INTEGER);
PREPARE p AS SELECT * FROM pragma_table_info('t');
EXECUTE p;
DROP TABLE t;
EXECUTE p;
```

```bash
./build/debug/duckdb :memory: < poc.sql
```

```
==2194459==ERROR: AddressSanitizer: heap-use-after-free on address 0x51700000a180 at pc 0x64d63569efaf bp 0x7ffe3a0a8690 sp 0x7ffe3a0a8680
READ of size 8 at 0x51700000a180 thread T0
    #0 0x64d63569efae in PragmaTableInfoFunction src/function/table/system/pragma_table_info.cpp:277
    #1 0x64d637c3a65e in duckdb::PhysicalTableScan::GetDataInternal(...) src/execution/operator/scan/physical_table_scan.cpp:180

freed by thread T0 here:
    #0 0x7f53a72ff5e8 in operator delete(void*, unsigned long)
    #1 0x64d62b042946 in duckdb::DuckTableEntry::~DuckTableEntry() src/include/duckdb/catalog/catalog_entry/duck_table_entry.hpp:24
    #9 0x64d62b19f3c8 in duckdb::CatalogEntryMap::DropEntry(duckdb::CatalogEntry&) src/catalog/catalog_set.cpp:76
    #10 0x64d62b1b015a in duckdb::CatalogSet::CleanupEntry(duckdb::CatalogEntry&) src/catalog/catalog_set.cpp:500

previously allocated by thread T0 here:
    #1 0x64d62aff9a9c in duckdb::make_uniq<duckdb::DuckTableEntry, ...>(...) src/include/duckdb/common/helper.hpp:77
    #2 0x64d62af1e744 in duckdb::DuckSchemaEntry::CreateTable(...) src/catalog/catalog_entry/duck_schema_entry.cpp:190

SUMMARY: AddressSanitizer: heap-use-after-free src/function/table/system/pragma_table_info.cpp:277 in PragmaTableInfoFunction
==2194459==ABORTING
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

alonfaraj:
Fixes #25075

`PragmaTableInfoBind` stores a raw `CatalogEntry &` in its bind data but never calls `RegisterDBRead`, so `read_databases` stays empty and both the rebind guard in `Optimizer::Optimize()` and `PreparedStatementData::RequireRebind()` treat the cached physical plan as safe to reuse. Dropping the table between two `EXECUTE`s frees the `DuckTableEntry` the plan still references - a heap-use-after-free.

Three more table functions bind a catalog entry the same way and reproduce the same UAF: `pragma_show`, `pragma_storage_info` and `duckdb_table_sample`. All four now register the read via a small `Binder::RegisterEntryRead` helper, mirroring what `bind_basetableref.cpp` already does for base tables, and return the same `Catalog Error: Table with name t does not exist!` that a plain table scan gives.

### DIFF
--- a/src/function/table/system/pragma_storage_info.cpp
+++ b/src/function/table/system/pragma_storage_info.cpp
@@ -114,6 +114,7 @@ static unique_ptr<FunctionData> PragmaStorageInfoBind(ClientContext &context, Ta
 	CatalogEntryRetriever retriever(context);
 	qname = Binder::BindTableName(retriever, qname);
 	auto &table_entry = Catalog::GetEntry<TableCatalogEntry>(context, qname);
+	Binder::RegisterEntryRead(input.binder, context, table_entry);
 	return make_uniq<PragmaStorageFunctionData>(table_entry, options);
 }
 
--- a/src/function/table/system/pragma_table_info.cpp
+++ b/src/function/table/system/pragma_table_info.cpp
@@ -173,6 +173,7 @@ static unique_ptr<FunctionData> PragmaTableInfoBind(ClientContext &context, Tabl
 	CatalogEntryRetriever retriever(context);
 	qname = Binder::BindTableName(retriever, qname);
 	auto &entry = Catalog::GetEntry(context, CatalogType::TABLE_ENTRY, qname);
+	Binder::RegisterEntryRead(input.binder, context, entry);
 	return make_uniq<PragmaTableFunctionData>(entry, IS_PRAGMA_TABLE_INFO);
 }
 
--- a/src/function/table/system/pragma_table_sample.cpp
+++ b/src/function/table/system/pragma_table_sample.cpp
@@ -52,6 +52,7 @@ static unique_ptr<FunctionData> DuckDBTableSampleBind(ClientContext &context, Ta
 		names.push_back(col.GetName());
 	}
 
+	Binder::RegisterEntryRead(input.binder, context, entry);
 	return make_uniq<DuckDBTableSampleFunctionData>(entry);
 }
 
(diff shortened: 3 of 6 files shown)


Mytherin:
Thanks!

PR #25154, opened 2026-08-30
Propagate numeric stats for unsigned integers

dentiny:
Hi team, I found current numeric columns only propagate stats for signed values, which leads missing row group pruning for unsigned values, for example,
```sql
memory D ATTACH '/tmp/uint_arith_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE ti(x INTEGER NOT NULL);
db D INSERT INTO ti SELECT range::INTEGER FROM range(0, 10240);

-- signed columns prune row groups correctly
db D EXPLAIN SELECT count(*) FROM (SELECT (x + 100) AS y FROM ti) WHERE y > 100000;
╭─ Ungrouped Aggregate ────╮
│ Aggregates: count_star() │
│ ~1 row                   │
╰─────────────┬────────────╯
╭─ Empty Result ───────────╮
╰──────────────────────────╯

db D CREATE TABLE tu(x UINTEGER NOT NULL);
db D INSERT INTO tu SELECT range::UINTEGER FROM range(0, 10240);

-- but unsigned columns don't prune at all
db D EXPLAIN SELECT count(*) FROM (SELECT (x + 100::UINTEGER) AS y FROM tu) WHERE y > 100000::UINTEGER;
│ Aggregates: count_star()    │
│ ~1 row                      │
╰──────────────┬──────────────╯
╭─ Projection ─┴──────────────╮
│ Projections: 42             │
│ ~2,048 rows                 │
╰──────────────┬──────────────╯
╭─ Seq Scan ───┴──────────────╮
│ Table: db.main.tu           │
│ Type: Sequential Scan       │
│ Filters: (x + 100) > 100000 │
│ ~2,048 rows                 │
```

### DIFF
--- a/src/function/scalar/operator/arithmetic.cpp
+++ b/src/function/scalar/operator/arithmetic.cpp
@@ -109,6 +109,16 @@ static Value NumericStatsValue(const LogicalType &type, hugeint_t value) {
 	return Value::Numeric(type, value);
 }
 
+static Value NumericStatsValue(const LogicalType &type, uint64_t value) {
+	D_ASSERT(type.IsNumeric());
+	return Value::Numeric(type, uhugeint_t(value));
+}
+
+static Value NumericStatsValue(const LogicalType &type, uhugeint_t value) {
+	D_ASSERT(type.IsNumeric());
+	return Value::Numeric(type, value);
+}
+
 template <class T>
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25156, opened 2026-08-30
Fix distinct comparisons in index value extraction

### LINKED ISSUE #25155: Incorrect results from ART index scan when IN is combined with IS NOT DISTINCT FROM
### What happens?

DuckDB can return incorrect results when an ART index scan is used for a predicate that combines an IN condition with IS NOT DISTINCT FROM on the same column.

The query should only return rows where a = 1, but the indexed execution path also returns rows where a = 3.

When the index scan is disabled, the query returns the correct result.

### To Reproduce

SQL script:

```sql
SET index_scan_percentage = 1.0;

CREATE TABLE t(a INTEGER, p INTEGER);

INSERT INTO t VALUES
    (1, 10),
    (3, 30);

CREATE INDEX t_a ON t(a);

SELECT p
FROM t
WHERE a IN (1, 3)
  AND a IS NOT DISTINCT FROM 1
ORDER BY p;
```

Actual output:

```sql
┌───────┐
│   p   │
│ int32 │
├───────┤
│    10 │
│    30 │
└───────┘
```

Expected output:

```sql
┌───────┐
│   p   │
│ int32 │
├───────┤
│    10 │
└───────┘
```

### OS:

macOS, arm64

### DuckDB Version:

v1.6.0-dev13791

### DuckDB Client:

DuckDB CLI

### Hardware:

Apple Silicon M5

### Full Name:

Yibo Chen

### Affiliation:

Independent developer

- [x] Yes, I have

- [x] Yes, I have

idle-lab:
Closes #25155 

This PR fixes incorrect ART index scan results when an `IN` predicate is combined with `IS DISTINCT FROM` or `IS NOT DISTINCT FROM` on the same column.

During exact-value extraction for an index scan, `ValueQualifies` did not handle `COMPARE_DISTINCT_FROM` and `COMPARE_NOT_DISTINCT_FROM`. These comparisons therefore reached the default case and incorrectly retained all candidate values extracted from the `IN` predicate.

### DIFF
--- a/src/function/table/table_scan.cpp
+++ b/src/function/table/table_scan.cpp
@@ -658,6 +658,12 @@ static bool ValueQualifies(const Value &value, const vector<ComparisonCondition>
 		case ExpressionType::COMPARE_LESSTHANOREQUALTO:
 			passes = ValueOperations::LessThanEquals(value, comp.constant);
 			break;
+		case ExpressionType::COMPARE_DISTINCT_FROM:
+			passes = ValueOperations::DistinctFrom(value, comp.constant);
+			break;
+		case ExpressionType::COMPARE_NOT_DISTINCT_FROM:
+			passes = ValueOperations::NotDistinctFrom(value, comp.constant);
+			break;
 		default:
 			return true;
 		}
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25159, opened 2026-08-30
Fix wrong results of TopN dynamic filter pushdown through a cast on the sort key

### LINKED ISSUE #25158: TopN dynamic filter pushdown across type-changing casts causes wrong results
### What happens?

`ORDER BY <bare column>` triggers a TopN dynamic filter pushdown (`TopN::PushdownDynamicFilters`) even when the sort key column reaches the base table scan through a type-changing integral cast (e.g. `GROUP BY CAST(k AS BIGINT)`, or a `UNION ALL` child projection with a cast). The pushed filter's boundary value is constructed in the **sort key's type**, while the raw scan column has a **different storage type**; the pushdown target's `mode` / `runtime_filter_casts` / `storage_type` are ignored.

At runtime, the row-group zonemap pruner (`NumericStats::CheckZonemap` → `CheckZonemapTemplated`) compares the storage-type stats against the sort-key-typed boundary:

- **Assert builds** (`FORCE_ASSERT=1 make relassert`): `D_ASSERT(constant_value.type() == stats.GetType())` (and a second `D_ASSERT` inside `Value::GetValueUnsafe<T>`) fails → `INTERNAL Error` crash.
- **Release builds**: `Value::GetValueUnsafe<T>()` reinterprets the boundary value's raw bytes as the storage type without casting (type punning, UB). Widening casts (INTEGER storage → BIGINT sort key) happen to read the low bytes exactly and silently behave; narrowing casts (BIGINT storage → INTEGER sort key) read uninitialized bytes of the `Value` union, producing a garbage boundary that can prune row groups containing rows that belong in the result → **wrong results (missing rows)**.

### To Reproduce

All reproductions are plain SQL run from the DuckDB CLI.

**Repro 1 — widening cast through an aggregate (assert-build crash; filter is pushed on release):**

```sql
CREATE TABLE tint AS SELECT i::INTEGER AS k FROM range(0, 204800) t(i);
EXPLAIN ANALYZE SELECT CAST(k AS BIGINT) AS x, count(*) FROM tint GROUP BY CAST(k AS BIGINT) ORDER BY x LIMIT 10;
SELECT CAST(k AS BIGINT) AS x, count(*) FROM tint GROUP BY CAST(k AS BIGINT) ORDER BY x LIMIT 3;
```

Observed on a release build (`make reldebug`) — the dynamic filter is pushed onto the INTEGER column with a BIGINT-typed boundary:

```
╭─ Table Scan ────────────────┐
│ Table: memory.main.tint     │
│ Projections: k              │
│ Filters: optional: Dynamic Filter (k)
│ Row Groups Scanned: 2 / 2   │
┌───────┬──────────────┐
│   x   │ count_star() │
│ int64 │    int64     │
├───────┼──────────────┤
│     0 │            1 │
│     1 │            1 │
│     2 │            1 │
└───────┴──────────────┘
```

In an assert build this instead fails with an internal assertion.

**Repro 2 — same bug through set-op child projections:**

```sql
CREATE TABLE ta AS SELECT i::INTEGER AS k FROM range(0, 100000) t(i);
CREATE TABLE tb AS SELECT (100000 + i)::INTEGER AS k FROM range(100000, 200000) t(i);
SELECT k FROM (SELECT CAST(k AS BIGINT) AS k FROM ta UNION ALL SELECT CAST(k AS BIGINT) AS k FROM tb) st ORDER BY k LIMIT 5;
```

`EXPLAIN ANALYZE` shows `Filters: optional: Dynamic Filter (k)` on both underlying scans.

**Repro 3 — narrowing cast (release, UB: may silently drop rows that belong in the result):**

```sql
CREATE TABLE tbig AS SELECT i::BIGINT AS k FROM range(0, 3000000) t(i);
SELECT CAST(k AS INTEGER) AS x, count(*) FROM tbig GROUP BY CAST(k AS INTEGER) ORDER BY x DESC LIMIT 3;
```

Here the boundary is an INTEGER value (4 bytes written into the `Value` union) that the zonemap path reads as `int64_t` (8 bytes) via `GetValueUnsafe<int64_t>()`; the upper 4 bytes are uninitialized. Depending on memory contents the garbage boundary makes the zonemap check prune row groups that contain the true top-3 values, so the query returns fewer/wrong rows. (The assert-build crash of Repro 1 is deterministic; whether a given release build manifests wrong results for this direction depends on the uninitialized bytes.)

### OS:

macOS

### DuckDB Version:

8616efa9da

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Jayson Huang

### Affiliation:

PingCAP

- [x] Yes, I have

- [x] Yes, I have

JaySon-Huang:
Hi team!

This fixes #25158: instead of bailing out of the TopN dynamic filter pushdown whenever the sort key and the scan column differ by a type cast, we now push the cast down into the filter's constant side for the cases that can be reconstructed. This removes the type-mismatched filter that #25158 reports, rather than just refusing to push.

## How it works

`TopN::PushdownDynamicFilters` records the cast chain on the pushed column, which the same `runtime_filter_casts` mechanism the join filter pushdown already uses, and replays it through a small `RuntimeFilterCastUtil` shared with the hash join. The dynamic filter constant is then compared in the raw scan column's type, so the boundary comparison is always well-typed. Extracting the util is a near-pure refactor on the join side.

The original unconditional guard stays as a safety net: we still bail out when the sort key cannot be reconstructed on top of the raw scan value in its own type (e.g. a non-integral cast or a VARIANT in between).

Set-op (UNION) children are handled independently — each child's path to its scan is analysed on its own, so a cast chain is reconstructed per child with no cross-child type contamination.

When `nulls_first` or a cast in the chain could fail, the filter is wrapped in `OR IS NULL(...)` so rows whose sort key is NULL or would raise a cast error are not dropped. The legacy serialization path drops a filter evaluated on a reconstructed expression, so older readers stay safe.

### DIFF
--- a/src/execution/operator/join/physical_hash_join.cpp
+++ b/src/execution/operator/join/physical_hash_join.cpp
@@ -1367,86 +1367,18 @@ static unique_ptr<ExpressionFilter> CreateSelectivityOptionalExpressionFilter(un
                                                                               const LogicalType &column_type,
                                                                               SelectivityOptionalFilterType type);
 
-static LogicalType GetRuntimeFilterInputType(const JoinFilterPushdownColumn &column, const LogicalType &runtime_type) {
-	if (column.mode == JoinFilterPushdownMode::RECONSTRUCT_EXPRESSION && !column.runtime_filter_casts.empty()) {
-		return column.runtime_filter_casts.back().target_type;
-	}
-	return runtime_type;
-}
-
 struct RuntimeFilterInput {
 	unique_ptr<Expression> expression;
 	bool preserves_cast_errors;
 };
 
-static bool RuntimeFilterCastCanFail(const LogicalType &source_type, const LogicalType &target_type) {
-	if (source_type == target_type) {
-		return false;
-	}
-	if (source_type.id() == LogicalTypeId::VARIANT || target_type.id() == LogicalTypeId::VARIANT ||
-	    !source_type.IsIntegral() || !target_type.IsIntegral()) {
-		return true;
-	}
-
-	const auto source_size = GetTypeIdSize(source_type.InternalType());
-	const auto target_size = GetTypeIdSize(target_type.InternalType());
-	if (source_size > target_size) {
-		return true;
-	}
-	if (source_type.IsSigned() == target_type.IsSigned()) {
-		return false;
-	}
-	if (source_type.IsSigned()) {
-		return true;
-	}
-	return source_size >= target_size;
-}
-
-static bool RuntimeFilterUsesTryCast(const JoinFilterPushdownColumn &column) {
-	auto source_type = column.storage_type;
-	for (auto &cast : column.runtime_filter_casts) {
-		if (cast.mode == RuntimeFilterCastMode::TRY_CAST || RuntimeFilterCastCanFail(source_type, cast.target_type)) {
-			return true;
-		}
-		source_type = cast.target_type;
-	}
-	return false;
-}
-
-static bool RequiresRuntimeFilterExpressionReconstruction(const JoinFilterPushdownColumn &column,
-                                                          const LogicalType &runtime_type) {
-	if (column.mode != JoinFilterPushdownMode::RECONSTRUCT_EXPRESSION) {
-		return false;
-	}
-	auto source_type = column.storage_type;
-	for (auto &cast : column.runtime_filter_casts) {
-		if (RuntimeFilterCastCanFail(source_type, cast.target_type)) {
-			return true;
-		}
-		source_type = cast.target_type;
-	}
-	D_ASSERT(source_type == GetRuntimeFilterInputType(column, runtime_type));
-	return false;
-}
-
 static RuntimeFilterInput CreateRuntimeFilterInputExpression(ClientContext &context,
                                                              const JoinFilterPushdownColumn &column,
                                                              const LogicalType &runtime_type) {
-	D_ASSERT(column.storage_type.IsValid());
-	unique_ptr<Expression> input = make_uniq<BoundReferenceExpression>(column.storage_type, idx_t(0));
-	auto source_type = column.storage_type;
 	bool preserves_cast_errors = false;
-	for (auto &cast : column.runtime_filter_casts) {
-		const auto cast_can_fail = RuntimeFilterCastCanFail(source_type, cast.target_type);
-		const auto is_try_cast = cast.mode == RuntimeFilterCastMode::TRY_CAST || cast_can_fail;
-		if (source_type != cast.target_type) {
-			input = BoundCastExpression::AddCastToType(context, std::move(input), cast.target_type, is_try_cast);
-		}
-		preserves_cast_errors |= cast.mode == RuntimeFilterCastMode::DEFAULT_CAST && cast_can_fail;
-		source_type = cast.target_type;
-	}
-	D_ASSERT(source_type == GetRuntimeFilterInputType(column, runtime_type));
-	return {std::move(input), preserves_cast_errors};
+	auto expression = RuntimeFilterCastUtil::CreateRuntimeFilterInputExpression(context, column, preserves_cast_errors);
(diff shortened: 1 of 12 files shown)


lnkuiper:
Instead of bailing out, do you think it would be possible to push type casts down there? That's what we do for join filter pushdown. We cast to the correct type before evaluating the filter instead of bailing on the filter. There should be some infrastructure to do that already. Might be difficult when there's, e.g., a UNION going into the top N, but I'd like to know if it's feasible.

JaySon-Huang:
@lnkuiper Thank you for your review. It would be ideal if we could push down even with casts. Following your guidance, I see some functions in `physical_hash_join` that seem to help handle this situation, and I'm currently verifying them.

JaySon-Huang:
Hi @lnkuiper,

The branch is reworked around it now. I've extracted a runtime-filter cast helper out of hash-join path. And the last commit handle the topn semantics.

Instead of bailing when the sort key and the scan column differ by a cast, we replay the cast chain on top of the raw scan column and evaluate the dynamic filter there. The filter boundary is built in the sort key's type, and it's now compared against the reconstructed value in that same type. So the comparison is always well-typed and the original "reinterpreted against a different storage type" bug can't surface.

Two correctness details in that last commit:
- When `NULLS FIRST` is in play, or a cast in the chain could raise an error, the filter is wrapped in `OR IS NULL(...)` so rows that would otherwise be silently dropped (NULL sort keys, or values that error in the original expression) stay in the TopN candidate set.
- A filter built on a reconstructed expression deserializes to an empty filter on older readers, so nothing breaks.

On the UNION-into-TopN case you flagged — that was the tricky bit. The pushdown walks each union arm independently: every branch gets its own column bindings and its own cast chain, so a cast on one arm can't bleed into another. The limitation is that the dynamic-filter boundary is shared across branches, so if any single arm can't be reconstructed, I bail the whole pushdown rather than pushing a partial one. Conservative, but it keeps the semantics exactly right.

lnkuiper:
Thanks for the PR! I think we should simplify this before merging.

TopN and the row-group pruner should only follow cast paths that preserve the raw column's ordering. Replaying arbitrary cast chains means reproducing `TRY_CAST` NULLs and default-cast errors, while `RowGroupPruner` currently ignores the recorded chain. This still produces wrong results:

```sql
SET threads=1;
CREATE TABLE t(k INTEGER);
INSERT INTO t SELECT 0 FROM range(122880);
INSERT INTO t SELECT 1000 FROM range(122880);

SELECT TRY_CAST(k AS TINYINT) AS x
FROM t
ORDER BY x NULLS FIRST
LIMIT 3;
```

This returns three zeroes instead of three NULLs. The issue predates this patch, but it shows that arbitrary cast chains are not a safe shared contract for order-based optimizations.

Could we keep the existing cast-chain reconstruction for join filters, but have TopN and `RowGroupPruner` bail on explicit `TRY_CAST` and any cast that can throw? For the remaining widening integral casts, TopN can cast the raw input directly to the final sort type, and the row-group pruner can safely use the raw statistics.

JaySon-Huang:
Thanks for the review! You're right — arbitrary cast chains are not a safe shared contract for order-based optimizations. Reworked accordingly:

- `TopN::PushdownDynamicFilters` and the row-group pruner now bail on explicit `TRY_CAST` and any cast that can throw (`RuntimeFilterCastUtil::RuntimeFilterUsesTryCast`). Only order-preserving widening integral casts are followed.
- For those, TopN replays the (now monotone, error-free) cast chain on top of the raw scan value and compares the boundary in the sort key's type — equivalent to casting the raw input directly to the final sort type.
- The row-group pruner applies the same guard; for the remaining chains the raw statistics are order- and NULL-preserving, so they can be used safely.
- The `OR IS NULL` wrapper is now only added for `NULLS FIRST` (the error-preserving branch is unreachable by construction).

Your reproduction now returns three NULLs; added as a regression test in `topn_dynamic_filter_cast.test`.

lnkuiper:

PR #25161, opened 2026-08-30
[Dev] Check the result of calling `weak_ptr::lock` in `StreamQueryResult` creation

Tishj:
This PR cherry picks a fix from #25145 

We weren't checking the result, added a test that triggers this.


Mytherin:
Thanks!

PR #25171, opened 2026-08-31
Allow max_memory to be set as a percentage

carlopi:
block_allocator_memory already accepted a percentage while max_memory did not, so "SET max_memory='50%'" was a parser error ("Unknown unit for memory: '%'") even though the default is itself computed as a percentage of system memory.

Factor the percentage handling out of BlockAllocatorMemorySetting into a shared helper and use it for max_memory as well. The base differs: block_allocator_memory resolves against maximum_memory, while max_memory resolves against the available system memory, as resolving it against maximum_memory would be circular.

The base is computed lazily, since options can be configured on a DBConfig before the database starts - at which point config.file_system is not yet set - and an absolute value must not require it. When a percentage is given before startup, fall back to a temporary local filesystem.

### DIFF
--- a/src/main/settings/custom_settings.cpp
+++ b/src/main/settings/custom_settings.cpp
@@ -68,6 +68,30 @@ static DatabaseInstance &GetDB(DatabaseInstance *db) {
 	return *db;
 }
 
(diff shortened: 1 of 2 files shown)


lnkuiper:
Thanks! LGTM

PR #25172, opened 2026-08-31
Add `preset` to tune family of related settings together, informing the system of intended usage

carlopi:
Basic SQL surface is:
```sql
CALL preset('host:dedicated');
CALL preset('host:shared');
CALL preset('host:background');
```
or the equivalent CLI-only
```
duckdb -preset host:dedicated
```
would set a group of logically tied settings together, allowing to feed DuckDB information about the presumed usage mode.

More advanced, allowing to provide your own mapping, such as:
```sql
--- session only
CALL register_preset('my_presets:custom', {'threads': 3, 'preserve_insertion_order': false});
--- persisted to disk in implicit folder
CALL register_preset('my_presets:custom', {'threads': 3, 'preserve_insertion_order': false}, persistent:=true);
--- persisted to disk in arbitrary location
CALL register_preset('path/to/some/file.json', {'threads': 3, 'preserve_insertion_order': false}, persistent:=true);
```
that can pair to:
```sql
CALL preset('my_presets:custom'); ---- for same session OR for persisted ones
CALL preset('path/to/some/file.json'); ---- from a file
```

Listing known presets can be done via:
```sql
FROM duckdb_presets();
```

Motivation / goal of the PR is offering a path to users (or extension maintainers) to bundle in a client-independent way to provide customization of duckdb behaviour, limited only to settings. Examples would be setting a proxy, or tweaking network other network behaviour given own situation, or tweaking threads consumption given information about a given machine or workload to follow.

Note all this is already viable via SQL, by having an init script of sort, but init scripts or .rc files are a CLI only feature, and I think the more limited (that is settings only) subset could be of general use.

Also, this is NOT saying defaults are not good over a wide range of usage, this is providing convenience capabilities (like storing to file) to allow reproducibility AND ease of use for situations where one might want to opt-in to extra capabilities. One example would be logging, you could have a given logging setup, and avoid having to remember which is exactly via `CALL preset('my_presets:log_to_csv');`. This also makes presets shareable cross machines / users.


src/include/duckdb/main/setting_preset.hpp
```diff
@@ -0,0 +1,126 @@
...
+	Value old_value;
+	Value new_value;
+	//! "applied", "reset", or "skipped"
+	string status;
+	//! Why the setting was skipped; empty otherwise
+	string note;
+};
+
+//! Presets registered at runtime, and where persistent ones are kept. Owned by the DBConfig.
+class PresetManager {
```
guillesd, line 76: Should we add a `Remove` API here? how does this work with secrets?

src/main/settings/setting_preset.cpp
```diff
@@ -0,0 +1,444 @@
...
+	return Value::BIGINT(NumericCast<int64_t>(MaxValue<idx_t>(SystemThreads(config) / divisor, 1)));
+}
+
+//! The default async thread count: four per core, capped
+Value DefaultAsyncThreads(DBConfig &config) {
+	return Value::BIGINT(NumericCast<int64_t>(MinValue<idx_t>(4 * SystemThreads(config), 256)));
+}
+
+//===--------------------------------------------------------------------===//
+// host: how much of this machine may DuckDB use?
```
guillesd, line 52: I think we should give more thought to the built-in presets, although that can be a follow up PR. There are so many settings that could be worth a more dedicated pre-set so people can easily use them (like `allocator_background_threads` or `block_allocator_memory`  or `late_materialization_max_rows`).

src/function/table/system/duckdb_presets.cpp
```diff
@@ -0,0 +1,297 @@
...
+		output.SetValue(2, count,
+		                result.new_value.IsNull() ? Value(LogicalType::VARCHAR) : Value(result.new_value.ToString()));
+		output.SetValue(3, count, Value(result.status));
+		output.SetValue(4, count, result.note.empty() ? Value(LogicalType::VARCHAR) : Value(result.note));
+		count++;
+	}
+	output.SetCardinality(count);
+}
+
+void PresetFun::RegisterFunction(BuiltinFunctions &set) {
```
guillesd, line 182: we should add a `reset_preset`, to reset all settings to their initial defaults if desired. I'm also not convinced about the name `preset`, I kind of liked the `profile` thing that you had in your issue better.

carlopi:
Actually, walking back (or forward) on this: why not just a way to execute an arbitrary SQL file like by line?
Then, we don't need to be opinionated, and it's free for all (and the same logic could basically work).

I need to think some more, and look at what postgres and sqlite offer in this direction.

guillesd:
Besides the few comments, I'd like to mention a couple more things:
1. The table function is good for a prototype but if we agree to support this (which unless we surface that 100 settings are useless, we should) then we probably need some nice syntax for this like `CREATE PRESET` or `CREATE PROFILE` whatever we call it.
2. I like the idea also of serializing this into the catalog (similar to Macros), which makes them potentially better to share?

src/main/settings/setting_preset.cpp
```diff
@@ -0,0 +1,444 @@
...
+// autoload_known_extensions comes first so that the httpfs settings below can be resolved; they
+// are optional because httpfs may not be installed.
+
+const PresetMember NETWORK_FLAKY[] = {
+    {"autoload_known_extensions", [](DBConfig &) { return Value::BOOLEAN(true); }, false},
+    {"async_threads", [](DBConfig &c) { return ThreadFraction(c, 1); }, false},
+    {"http_retries", [](DBConfig &) { return Value::BIGINT(10); }, true},
+    {"http_retry_backoff", [](DBConfig &) { return Value::FLOAT(2); }, true},
+    {"http_timeout", [](DBConfig &) { return Value::BIGINT(120000); }, true},
+};
```
taniabogatsch, line 93: Where are these numbers/presets coming from? I'd be a bit more conservative about adding these "magic number presets" on top of our default values. Could we maybe change this PR so that it allows people to have their own presets, but we don't ship any of those ourselves, yet? We could just document these bundles instead?

But that and the size of this PR got me thinking that to my knowledge people already have some JSON files (or other init files) laying around and use them on DB/connection startup. In which case I feel like this PR doesn't really add much, other than adding functionality for some common use cases, which could also just be documented on the website... In the end, with the number of settings we have, people might already have to fine tune to their specific workloads sometimes.

I guess that brings me back to my point that removing settings and looking into whether we can be smarter at adaptively adjusting internals when running our queries would be a more long-term approach to the settings problem. Instead of now adding grouping on "too many settings already"...

taniabogatsch: > Note all this is already viable via SQL, by having an init script of sort, but init scripts or .rc files are a CLI only feature, and I think the more limited (that is settings only) subset could be of general use.

Not necessarily, the duckdb-go client has a `Connector`, and you can pass options to the DSN. We could make sure that we have similar interfaces with the other drivers.

taniabogatsch: > Actually, walking back (or forward) on this: why not just a way to execute an arbitrary SQL file like by line?
Then, we don't need to be opinionated, and it's free for all (and the same logic could basically work).

> I need to think some more, and look at what postgres and sqlite offer in this direction.

I think this goes more into my direction, where we provide the presets (in your suggestion in a SQL file), but don't ship them "hardcoded" in the code. We already have all the logic to run a bunch of SQL configuration statements, we might just need to give that a polish.

guillesd: I agree that the most important thing is looking at the current settings and:
1) Kill every unseful settings
2) For the long tail (settings that nobody uses) -> add a note to the description like WHEN WOULD YOU USE THIS? Example `arrow_output_list_view` (but I have 50 more like this)

Then I still think we need to document what impact settings better so, if we don't provide bundles, people should be able to make their own informed bundle.

carlopi: This is mostly a draft, and the built-in presets where a way to demonstrate some capabilities / play with this. Maybe can we check live at some later point today?

carlopi: And indeed the point is that while arbitrary client can implement the logic, each one will do it's own, and it would be great to move some of that to the SQL layer.

taniabogatsch: Yea I think that I am very much on board with the "moving to SQL layer", but, also, how is that different/more valuable than the SQL layer we already have (you can already, e.g., in a JSON with setting key-value pairs, store all your configs) that you run after you've opened your connection?

carlopi: I think what `preset` or whatever the name would buy is: only safe subsets of SQL, and some level of atomicity that doesn't need to be re-implemented user-by-user.

Like saying "just take this config file, iterate and set them all" is not super user friendly, while `CALL preset('path/to/file.sql');` seems more clear (to me).


One thing I realized: no custom json, but just a SQL file is strictly better, can be changed by hand, and can still be forced a type (like allow only settings-related changes, or allow all).

PR #25176, opened 2026-08-31
[Storage] Add `byte_size` to `DataPointer`

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/10584

This should prevent information leaks due to malicious data/compression bugs.
Previously we only serialized the block size. Because of this, it was impossible to be sure we weren't reading outside of the bounds of our segment, into other segment(s) that reside on the same block.

#### Misc

Found issues with the `scripts/regression/test_runner.py` script as part of this PR;
- Errors weren't being surfaced correctly. `INCORRECT` would just get cast to `float` and fail.
- Benchmark runs between old and new are reusing the same `cache_directory`, resulting in benchmark runs being influenced by side-effects from the other run (in this case visible through storage reuse)


Tishj:
Now it correctly reports:
```
1: benchmark/micro/logger/logging_overhead/duckdb_persistent_q1_with_default_logging.benchmark
Base:
 Base benchmark runner reported INCORRECT for benchmark/micro/logger/logging_overhead/duckdb_persistent_q1_with_default_logging.benchmark on run 1:
INCORRECT RESULT: TransactionContext Error: Failed to commit: Failed to deserialize: expected end of object, but found field id: 106
PR:
 No failure
- 52
```

Tishj:
@smvv looks like there's an isolation issue between the old/new runs of the same benchmark

smvv:
@Tishj fyi some benchmarks require/omit [clearing](https://github.com/duckdb/duckdb/blob/f055e15597e1538d78e58e0d443ceeee8827e919/.github/workflows/Regression.yml#L194) the benchmark cache dir.
Preferably we clear it always (so we isolate bench runs better, as you said), but i'm not sure if/when a benchmark requires clearing?

Tishj:
> @Tishj fyi some benchmarks require/omit [clearing](https://github.com/duckdb/duckdb/blob/f055e15597e1538d78e58e0d443ceeee8827e919/.github/workflows/Regression.yml#L194) the benchmark cache dir. Preferably we clear it always (so we isolate bench runs better, as you said), but i'm not sure if/when a benchmark requires clearing?

Hmm, seeing as the only one where that's the case is `realnest`, my guess is that that's just because of the size of the benchmark data.
I would think we just want to make two separate benchmark directories for the two runners and symlink the required data of a benchmark into the directories ?

src/storage/table/column_data.cpp
```diff
@@ -900,7 +900,7 @@ void ColumnData::InitializeColumn(PersistentColumnData &column_data, BaseStatist
 		auto segment = ColumnSegment::CreatePersistentSegment(
 		    GetDatabase(), block_manager, data_pointer.block_pointer.block_id, data_pointer.block_pointer.offset,
 		    data_pointer.tuple_count, data_pointer.compression_type, std::move(data_pointer.statistics),
-		    std::move(data_pointer.segment_state));
+		    std::move(data_pointer.segment_state), data_pointer.byte_size);
```
Mytherin: Maybe we want to move the data pointer into this constructor?

src/include/duckdb/storage/table/column_segment.hpp
```diff
@@ -174,6 +185,8 @@ class ColumnSegment : public SegmentBase<ColumnSegment> {
 	idx_t offset;
 	//! The allocated segment size, which is bounded by Storage::BLOCK_SIZE
 	idx_t segment_size;
+	//! Number of bytes occupied by this segment within its block, if recorded
+	optional<uint32_t> byte_size;
```
Mytherin: `data_size`?

Mytherin (commented):
Thanks! Looks great - minor comments, otherwise great idea

Mytherin:
Thanks!

PR #25177, opened 2026-08-31
Special handling for block comments after an operator

Dtenwolde:

Having a block comment `/* */` right after an operator caused incorrect tokenization.

```sql
memory D SELECT 1+/**/2;
┌─────────┐
│ (1 + 2) │
│  int32  │
├─────────┤
│       3 │
└─────────┘
memory D SELECT 1*/**/2;
┌─────────┐
│ (1 * 2) │
│  int32  │
├─────────┤
│       2 │
└─────────┘
memory D SELECT 1+/* comment */2;
┌─────────┐
│ (1 + 2) │
│  int32  │
├─────────┤
│       3 │
└─────────┘
```

This was a regression between the old parser and the new parser, so now we have the same behaviour as PostgreSQL here again :)

### DIFF
--- a/src/include/duckdb/parser/peg/tokenizer/tokenizer.hpp
+++ b/src/include/duckdb/parser/peg/tokenizer/tokenizer.hpp
@@ -70,6 +70,7 @@ private:
 	//! `TokenizeInput()` is the one that appends `GetTerminator()` (clean) or `END_OF_INPUT`
 	//! (dirty) based on the return value.
 	bool TokenizeInputInternal(TokenizerBehavior &behavior) const;
+	static void PushOperatorToken(TokenizerBehavior &behavior, idx_t start, idx_t end);
 
 public:
 	bool IsSpecialOperator(const string &sql, idx_t pos, idx_t &op_len) const;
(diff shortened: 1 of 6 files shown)


carlopi:
What should happen with open ended comments? Like: `SELECT 1\*` ?

Dtenwolde:
> What should happen with open ended comments? Like: `SELECT 1\*` ?

That's an unterminated comment in the CLI. While using `-c` strips the comment: 
```bash
build/release/duckdb -c "select 1/*;"
┌───────┐
│   1   │
│ int32 │
├───────┤
│     1 │
└───────┘
```

carlopi:
Old parsed did:
```
/opt/homebrew/bin/duckdb -c "select 1/*;"

Parser Error:
unterminated /* comment at or near "/*;"

LINE 1: select 1/*;
```
that I think I would expect as behaviour

Dtenwolde:
I think that makes most sense, I'll adjust the PR to restore that behavior :)

Dtenwolde:
CI failure is unrelated

Tishj (approved):
Thanks!

PR #25179, opened 2026-08-31
[Dev] Use `PhysicalResultCollector::IsStreaming` rather than just the hint if streaming is allowed

Tishj:
This PR is cherry picked from #25145 

Check the type of the result collector, not just whether "allow stream result" is set, to also support custom result collectors.


Mytherin:
Thanks!

PR #25181, opened 2026-08-31
Fix uint32_t blob offset overflow when casting oversized rows to VARIANT

### LINKED ISSUE #25078: Vuln45: Heap Buffer Overflow from uint32_t Blob-Offset Wraparound When Casting Large Rows to VARIANT
### What happens?

`ConvertPrimitiveTemplated()` in `primitive_to_variant.hpp` accumulates every field's byte length into a shared per-row `uint32_t blob_offset` (`to_variant_fwd.hpp`'s `OffsetData::GetBlob`) without ever checking the running total against `UINT32_MAX`. When a single STRUCT/ROW's combined VARCHAR/BLOB field sizes exceed about 4.29 GB, this accumulator wraps around during the first (sizing) pass, so `StringVector::EmptyString()` in `to_variant.cpp` allocates a row blob buffer that is far too small. The second (writing) pass then writes each field at its true, still non-wrapped-at-that-point byte offset via `memcpy` in `WriteData<T>`, driving the write far past the end of the under-allocated buffer. The same unguarded arithmetic pattern also appears in `ConvertVariantToVariant()` in `variant_to_variant.hpp`, where a developer comment explicitly flags the missing bounds check as unaddressed.

### To Reproduce

This is reproducible with a single, plain SQL statement fed directly to the DuckDB CLI, no input file needs to be constructed beforehand.

```sql
SELECT length(CAST(ROW(repeat('a', 1600000000), repeat('b', 1600000000), repeat('c', 1600000000)) AS VARIANT)) AS result;
```

```bash
# from the duckdb source root, after `make debug`
cat > poc.sql <<'EOF'
SELECT length(CAST(ROW(repeat('a', 1600000000), repeat('b', 1600000000), repeat('c', 1600000000)) AS VARIANT)) AS result;
EOF
./build/debug/duckdb -f poc.sql
```

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2641197==ERROR: AddressSanitizer: SEGV on unknown address 0x7047d1500680 (pc 0x704b4fba196b bp 0x7fff49192ef0 sp 0x7fff49192eb8 T0)
==2641197==The signal is caused by a WRITE memory access.
    #0 0x704b4fba196b in __memmove_avx512_unaligned_erms ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:839
    #1 0x620c4a7ee089 in WriteData<duckdb::string_t> src/include/duckdb/function/cast/variant/primitive_to_variant.hpp:61
    #2 0x620c4a882c4c in ConvertPrimitiveTemplated<...> src/include/duckdb/function/cast/variant/primitive_to_variant.hpp:235
    #3 0x620c4a9078f7 in ConvertPrimitiveToVariant<true, false>(...) src/include/duckdb/function/cast/variant/primitive_to_variant.hpp:362
    #4 0x620c4a8efd45 in ConvertToVariant<true, false>(...) src/include/duckdb/function/cast/variant/to_variant.hpp:62
    #5 0x620c4a902faa in ConvertTupleToVariant<true, false>(...) src/include/duckdb/function/cast/variant/tuple_to_variant.hpp:73
    #7 0x620c4a7f1c35 in WriteVariantResultData src/function/cast/variant/to_variant.cpp:103
    #8 0x620c4a7f36c1 in CastToVARIANT src/function/cast/variant/to_variant.cpp:245
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:839 in __memmove_avx512_unaligned_erms
==2641197==ABORTING
```

### OS:

24.04.2 LTS, x86_64 (Linux 6.8.0-55-generic)

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

asxvi:
## What
`ConvertPrimitiveTemplated()` in `primitive_to_variant.hpp` accumulated a running byte-length total in a `uint32_t` without checking it against `UINT32_MAX`. As reported in #25078, after a single STRUCT/ROW's combined VARCHAR/BLOB field sizes exceeded ~4.29GB, the accumulator would silently wrapped (leading to under-sized buffer allocation followed by out-of-bounds write).

Additionally, as mentioned in #25078, the same accumulation pattern was found independently in 2 additional places in `ConvertVariantToVariant`, where it instead hit an unguarded narrowing cast that threw `InternalException`

While debugging and testing, a third location, `VariantBuilder` (reached during unshredding/flattening a VARIANT value for display, for both literal and table-sourced input) had issue with same unchecked accumulation pattern. This led to the same potential overflows. `VariantBuilderCheckBlobSize()` was added to three blob-growing helpers (`AppendVarint`, `AppendFixed`, `AppendBytes`) immediately after
the buffer grows and before anything is written into the new space.

## Fix
Added `OffsetData::AddBlobOffset()`, a checked accumulator that throws
`InvalidInputException` instead of silently wrapping or hitting an
internal-only exception type. Used in both `ConvertPrimitiveTemplated()` and `ConvertVariantToVariant()`.

## Changes
- `to_variant_fwd.hpp` — adds `OffsetData::AddBlobOffset()`
- `primitive_to_variant.hpp - ConvertPrimitiveTemplated()` — original reported site, now uses `AddBlobOffset()`
- `variant_to_variant.hpp - ConvertVariantToVariant()` — same pattern, previously flagged by an existing `// FIXME` comment, now uses `AddBlobOffset()`
- `variant_builder.hpp - VariantBuilder` — independent bug in the unshredding/flatten path, adds and uses `VariantBuilderCheckBlobSize()`

## Tests
`test/sql/function/variant/blob_offset_overflow.test`
- A small sanity check -> OK
- A value exactly at the UINT32_MAX boundary -> OK
- A value exceeding UINT32_MAX,  -> InvalidInputException.
- Original report case (three 1.6 GB fields) -> InvalidInputException.

## Notes
- CI's batch timeout (120s) kills the largest test case with SIGKILL before it can finish — the throw is reached correctly, but stack unwinding through the multi-GB live allocations in a Relassert build takes longer than the timeout allows. Confirmed locally with a debugger that the throw itself fires as expected; the SIGKILL is a return-to-caller timing issue, not a correctness issue.
- Difficulty running and creating valid test cases due to the large buffer sizes involved (~4.29GB+ per case) — my dev machine is memory-constrained at this scale.

Fixes #25078

### DIFF
--- a/src/include/duckdb/common/types/variant/variant_builder.hpp
+++ b/src/include/duckdb/common/types/variant/variant_builder.hpp
@@ -47,19 +47,10 @@ namespace duckdb {
 //! Sentinel marking an array child (whose 'key_id' is NULL)
 constexpr uint32_t VARIANT_INVALID_KEY = NumericLimits<uint32_t>::Maximum();
 
-inline void VariantBuilderCheckBlobSize(const string &blob) {
-	if (blob.size() > NumericLimits<uint32_t>::Maximum()) {
-		throw InvalidInputException(
-		    "Cannot convert value to VARIANT: encoded row size exceeds the maximum supported %u bytes",
-		    NumericLimits<uint32_t>::Maximum());
-	}
-}
-
 inline void VariantBuilderAppendVarint(string &blob, uint32_t value) {
 	auto size = GetVarintSize(value);
 	auto pos = blob.size();
 	blob.resize(pos + size);
-	VariantBuilderCheckBlobSize(blob);
 	VarintEncode<uint32_t>(value, data_ptr_cast(blob.data()) + pos);
 }
 
@@ -67,13 +58,11 @@ template <class T>
 void VariantBuilderAppendFixed(string &blob, T value) {
 	auto pos = blob.size();
 	blob.resize(pos + sizeof(T));
-	VariantBuilderCheckBlobSize(blob);
 	Store<T>(value, data_ptr_cast(blob.data()) + pos);
 }
 
 inline void VariantBuilderAppendBytes(string &blob, const_data_ptr_t data, idx_t size) {
 	blob.append(const_char_ptr_cast(data), size);
-	VariantBuilderCheckBlobSize(blob);
 }
 
 inline uint32_t VariantBuilderGetOrCreateIndex(OrderedOwningStringMap<uint32_t> &dictionary, const string_t &key) {
@@ -137,14 +126,18 @@ struct VariantBuilder {
 	//! Emit a VARIANT_NULL value
 	void EmitNull() {
 		type_ids.push_back(static_cast<uint8_t>(VariantLogicalType::VARIANT_NULL));
-		byte_offsets.push_back(NumericCast<uint32_t>(blob.size()));
+		//! 'blob' may already exceed uint32_t mid-row; this offset is only read once the row is known to fit
+		//! (checked in BuildVariant once the row is fully emitted), so a narrowing wrap here is harmless.
+		byte_offsets.push_back(static_cast<uint32_t>(blob.size()));
 	}
 
 	//! Emit an OBJECT value with 'n' children (assumed to be in lexicographic key order). 'key_fn(i)'
 	//! returns the (string_t) key of child i; 'emit_fn(i)' must emit exactly one value for child i.
 	template <class KEY_FN, class EMIT_FN>
 	void EmitObject(idx_t n, KEY_FN &&key_fn, EMIT_FN &&emit_fn) {
-		auto byte_offset = NumericCast<uint32_t>(blob.size());
+		//! 'blob' may already exceed uint32_t mid-row; this offset is only read once the row is known to fit
+		//! (checked in BuildVariant once the row is fully emitted), so a narrowing wrap here is harmless.
+		auto byte_offset = static_cast<uint32_t>(blob.size());
 		type_ids.push_back(static_cast<uint8_t>(VariantLogicalType::OBJECT));
 		byte_offsets.push_back(byte_offset);
 		VariantBuilderAppendVarint(blob, NumericCast<uint32_t>(n));
@@ -166,7 +159,9 @@ struct VariantBuilder {
 	//! Emit an ARRAY value with 'n' elements. 'emit_fn(i)' must emit exactly one value for element i.
 	template <class EMIT_FN>
 	void EmitArray(idx_t n, EMIT_FN &&emit_fn) {
-		auto byte_offset = NumericCast<uint32_t>(blob.size());
+		//! 'blob' may already exceed uint32_t mid-row; this offset is only read once the row is known to fit
+		//! (checked in BuildVariant once the row is fully emitted), so a narrowing wrap here is harmless.
+		auto byte_offset = static_cast<uint32_t>(blob.size());
 		type_ids.push_back(static_cast<uint8_t>(VariantLogicalType::ARRAY));
 		byte_offsets.push_back(byte_offset);
 		VariantBuilderAppendVarint(blob, NumericCast<uint32_t>(n));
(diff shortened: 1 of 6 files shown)


src/include/duckdb/function/cast/variant/to_variant_fwd.hpp
```diff
@@ -31,6 +31,15 @@ struct OffsetData {
 	static uint32_t *GetBlob(DataChunk &offsets) {
 		return FlatVector::GetDataMutable<uint32_t>(offsets.data[3]);
 	}
+	//! Adds len to offset, throwing if the running total would exceed what a uint32_t blob offset can represent.
+	static void AddBlobOffset(uint32_t &offset, uint32_t len) {
+		if (offset > NumericLimits<uint32_t>::Maximum() - len) {
```
Mytherin: Can we use `uinteger_t` here instead of re-implementing these overflow checks?

src/include/duckdb/common/types/variant/variant_builder.hpp
```diff
@@ -47,22 +47,33 @@ namespace duckdb {
...
 	blob.resize(pos + size);
+	VariantBuilderCheckBlobSize(blob);
 	VarintEncode<uint32_t>(value, data_ptr_cast(blob.data()) + pos);
 }
 
 template <class T>
 void VariantBuilderAppendFixed(string &blob, T value) {
 	auto pos = blob.size();
 	blob.resize(pos + sizeof(T));
+	VariantBuilderCheckBlobSize(blob);
```
Mytherin: Instead of checking this at every single append can we just check it at the very end when we finalize the blob?

Mytherin (commented):

src/include/duckdb/common/types/variant/variant_builder.hpp
asxvi: Thank you for the review! 

Currently, the reason why we check per append is because this blob passing UINT32_MAX overflow can occur at any append, not  just the last one in a row.

Without the check at each step, once overflow occurs mid-row, the buffer will continue growing (which may not be significant for typical inputs, but larger inputs this means unnecessary expansion before the throw ever happens)

Happy to move to a single end-of-row check if that tradeoff isn't a
concern in practice. Let me know what you think!

Mytherin: That trade-off isn't worth it - let's just move to end-of-row check

Mytherin:
Thanks!

taniabogatsch:
@asxvi, the PR was rejected from the merge queue, would you mind taking a look at the failing tests? Thank you!

asxvi:
hello @taniabogatsch, thanks for flagging. I believe the faliure was in a boundary test case landing exactly on `UINT32_MAX`/`UINT32_MAX+1`. It ran successfully locally, but did not reproduce identially on the CI `release` build.

I widened the boundary test's margin to ~100KB on either side so it's robust to whatever that discrepancy is, rather than depending on exact-byte precision. Pushed the update.

taniabogatsch:
@asxvi, can we move this back to "Ready for Review", so that CI can run?

taniabogatsch:
We'd be happy to merge this contribution :)

asxvi:
Sorry for the delay on my end @taniabogatsch . The current changeset's unit tests now pass. The delay came from a failing test case needing ~4gb of live data to exercise the overflow boundary. We now `mode skip` this, but keep it in the file as documentation from the [example in the original issue](https://github.com/duckdb/duckdb/issues/25078)

PR #25185, opened 2026-08-31
Fix COMMENT ON COLUMN when a remote catalog is attached

pdet:
This pr fixes an unresolved catalog type access for `COMMENT ON COLUMN` in the remote pushdown optimizer. Came across it in DuckLake.


Mytherin:
Thanks!

PR #25187, opened 2026-08-31
State the snapshot visibility rule once

ywelsch:
DuckDB's MVCC visibility test is currently written out by hand wherever it is needed. Two types of comparisons,
- `timestamp < start_time` for "visible to this snapshot" and
- `timestamp < TRANSACTION_ID_START` for "committed"

appear inline across the catalog, the table storage, the version managers and the transaction manager. A bound and a timestamp are both `transaction_t`, so they can be swapped, assigned to each other, or compared inclusively without the compiler noticing. The codebase actually has all three: the `DuckTransactionManager` constructor initialises `lowest_active_id` and `lowest_visibility_bound` to each other's identity, `CheckpointOptions` defaults its bound to `MAX_TRANSACTION_ID` (a value above the split) without saying so, and `TransactionData` takes `(transaction_id, visibility_bound)` while every storage function takes them the other way round.

In preparation for #23631, where a transaction's visibility bound is no longer always its start time, this PR gives these visibility tests a single, explicit form and makes the bound a type of its own.

```cpp
inline bool IsCommitted(transaction_t timestamp) {
	return timestamp <= MAX_COMMIT_ID;
}

// a VisibilityBound is built, never assigned from a timestamp
VisibilityBound::Through(last_commit);   // visible up to and including last_commit
VisibilityBound::Before(start_time);     // visible strictly before start_time

// and these are the only comparisons against one: <= and > do not compile
timestamp < visibility_bound;
timestamp >= visibility_bound;
```

### DIFF
--- a/extension/tpcds/dsdgen/append_info-c.cpp
+++ b/extension/tpcds/dsdgen/append_info-c.cpp
@@ -156,7 +156,7 @@ void tpcds_append_information::FinalizeOptimisticAppend() {
 		return;
 	}
 	FlushChunk();
-	duckdb::TransactionData transaction_data(0, 0);
+	auto transaction_data = duckdb::TransactionData::Unversioned();
 	auto &row_collection = *optimistic_collection->collection;
 	row_collection.FinalizeAppend(transaction_data, append_state);
 	finalized = true;
--- a/extension/tpch/dbgen/dbgen.cpp
+++ b/extension/tpch/dbgen/dbgen.cpp
@@ -161,7 +161,7 @@ struct tpch_append_information {
 			return;
 		}
 		FlushChunk();
-		TransactionData transaction_data(0, 0);
+		auto transaction_data = TransactionData::Unversioned();
 		auto &row_collection = *optimistic_collection->collection;
 		row_collection.FinalizeAppend(transaction_data, append_state);
 		finalized = true;
--- a/src/catalog/catalog_set.cpp
+++ b/src/catalog/catalog_set.cpp
@@ -202,7 +202,7 @@ bool CatalogSet::CreateEntry(CatalogTransaction transaction, const Identifier &n
 	CheckCatalogEntryInvariants(*value, name);
 
 	// Mark this entry as being created by the current active transaction
-	value->timestamp = transaction.transaction_id;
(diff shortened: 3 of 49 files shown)


Mytherin:
Thanks! I think cleaning this up for more clarity makes sense. Instead of only doing renames though I would suggest going a step further and replacing `transaction_t` with a struct (e.g. `SnapshotId`) that has clearly named methods for each of these. This would also help with differentiating between transaction ids (`>= TRANSACTION_ID_START`) and snapshot / commit ids.

For `DuckTransaction` it might also make sense to combine `start_time` and `transaction_id` into a single struct since they should generally be used together, so packing them together avoids unintentionally using them separately / incorrectly.  `TransactionData` already kind of does this so maybe that could be unified there.

ywelsch:
I went with a separate type for the *bound* rather than for `transaction_t` itself. A bound and a timestamp being the same type is what caused the biggest risk for bugs (in particular these being compared with just integer comparison in various places).

The new `VisibilityBound` struct makes it clear what someone means (e.g. created via  `Through()` / `Before()` / `AllCommitted()` / `IncludingUncommitted()`), and only `timestamp < bound` and `timestamp >= bound` compile, so a bound stays always exclusive.

I've added your second suggestion as `SnapshotView`: it combines the bound plus the transaction id, and has a method `Sees()` which answers visibility once. It is used both by `TransactionData` and `CatalogTransaction`. `Transaction`, and in particular `DuckTransaction`, now has a `GetSnapshotView()` method that return its `SnapshotView` object. For now, this is based on `DuckTransaction`'s `start_time`, but once we go to https://github.com/duckdb/duckdb/pull/23631, the SnapshotBound will instead be based on what's been made durable before the TX started, not the `start_time`.

I did prototype the full split of commit ids from transaction ids. That felt overly complex as we sometimes use these interchangeably so would have required a third union-like type. For example, a stored version slot holds a transaction id while its writer is in flight and its commit id after. I also think doing this on top of the SnapshotBound would not help catch more interesting bugs.

Let me know what you think.

ywelsch:
This is ready now

Mytherin:
Looks good, thanks!

PR #25189, opened 2026-08-31
Fix: Do not produce invalid UTF8 from printf's %c conversion

### LINKED ISSUE #25166: INTERNAL Error and SIGSEGV in string functions applied to `printf('%c', 255)`
### What happens?

Consider below test cases:

### To Reproduce

```sql
SELECT hex(printf('%c', 255)) AS printf_c, hex(chr(255)) AS chr;
-- FF|C3BF -- printf emits the raw low byte, chr encodes the code point

SELECT lower(printf('%c', 255));
-- INTERNAL Error: Scalar function ""lower"" threw an execution error, but the function is not marked as fallible - the function must call SetFallible(). Error: invalid lead byte detected in Utf8Proc::UTF8ToCodepoint, likely due to invalid UTF-8

SELECT upper(printf('%c', 255));
-- INTERNAL Error: Scalar function ""upper"" threw an execution error, but the function is not marked as fallible - the function must call SetFallible(). Error: invalid lead byte detected in Utf8Proc::UTF8ToCodepoint, likely due to invalid UTF-8

SELECT rtrim(printf('%c', 255));
-- INTERNAL Error: Scalar function ""rtrim"" threw an execution error, but the function is not marked as fallible - the function must call SetFallible(). Error: Cannot create a string of size: '18446744073709551613', the maximum supported string size is: '4294967295'

SELECT trim(printf('%c', 255));
-- INTERNAL Error: Scalar function ""trim"" threw an execution error, but the function is not marked as fallible - the function must call SetFallible(). Error: Cannot create a string of size: '18446744073709551613', the maximum supported string size is: '4294967295'

SELECT strip_accents(printf('%c', 255));
-- Segmentation fault (core dumped) -- exit code 139; ends the session
```

### OS:

Ubuntu 22.04

### DuckDB Version:

v1.6.0-dev13791 (8616efa9da)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Suyang Zhong

### Affiliation:

NUS

- [x] Yes, I have

- [x] Yes, I have

alonfaraj:
Fixes #25166

`printf('%c', 255)` writes the raw byte `0xFF` into a VARCHAR, which isn't valid UTF-8. 
Everything downstream then falls on a string that was already corrupt before it got there:

```sql
SELECT lower(printf('%c', 255));         -- INTERNAL Error: invalid lead byte
SELECT trim(printf('%c', 255));          -- INTERNAL Error: string of size 18446744073709551613
SELECT strip_accents(printf('%c', 255)); -- Segfault
```

So reject a format result that isn't valid UTF-8 in the path shared by `printf` and `format`:

```
Invalid Input Error: Invalid UTF8 produced by format string "%c" - note that
%c writes a single byte, use chr(...) to write a Unicode code point
```

## Tests

Added to `test/sql/function/string/test_printf.test`

### DIFF
--- a/extension/core_functions/scalar/string/printf.cpp
+++ b/extension/core_functions/scalar/string/printf.cpp
@@ -3,6 +3,7 @@
 #include "duckdb/common/limits.hpp"
 #include "fmt/format.h"
 #include "fmt/printf.h"
+#include "utf8proc_wrapper.hpp"
 
 namespace duckdb {
 
@@ -172,6 +173,11 @@ static void PrintfFunction(DataChunk &args, ExpressionState &state, Vector &resu
 
 		// finally actually perform the format
 		string dynamic_result = FORMAT_FUN::template OP<CTX>(format_string.c_str(), current_args);
+		if (!Utf8Proc::IsValid(dynamic_result.c_str(), dynamic_result.size())) {
+			throw InvalidInputException("Invalid UTF8 produced by format string \"%s\" - note that %%c writes a "
+			                            "single byte, use chr(...) to write a Unicode code point",
+			                            format_string);
+		}
 		result_data.WriteValue(dynamic_result);
 	}
 }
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25190, opened 2026-08-31
Fix dangling CTE scope reference in the PEG transformer

### LINKED ISSUE #25173: Vuln70: Stack-UAF in `stored_cte_map` via PIVOT with CTE source subquery
### What happens?

In `TransformSelectStatementInternalRule` (transform_select.cpp:54-57), a `CommonTableExpressionMap` is allocated on the stack and a non-owning `reference<>` to it is pushed into `PEGTransformer::stored_cte_map`. The function then returns and destroys the local map without removing the reference from `stored_cte_map`. When `TransformPivotStatement` subsequently calls `ExtractCTEsRecursive` (peg_transformer.cpp:314) to enumerate CTEs for the PIVOT column resolution, it dereferences the dangling reference to the already-destroyed stack frame, producing a stack-use-after-return memory safety violation.

### To Reproduce

```sql
PIVOT (WITH x AS (SELECT 1 AS col) SELECT * FROM x) ON col;
```

```sh
echo "PIVOT (WITH x AS (SELECT 1 AS col) SELECT * FROM x) ON col;" | ./build/debug/duckdb
```

```
==1871591==ERROR: AddressSanitizer: stack-use-after-return on address 0x7f309b8190f0 at pc 0x6526da094e77 bp 0x7fff985653c0 sp 0x7fff985653b0
READ of size 8 at 0x7f309b8190f0 thread T0
    #0 ... __gnu_cxx::__normal_iterator ... stl_iterator.h:1077
    #2 ... duckdb::InsertionOrderPreservingMap<...>::begin() insertion_order_preserving_map.hpp:47
    #3 ... duckdb::PEGTransformer::ExtractCTEsRecursive(duckdb::CommonTableExpressionMap&) peg_transformer.cpp:314
    #4 ... duckdb::PEGTransformerFactory::TransformPivotStatement(...) transform_pivot.cpp:121
    #35 ... duckdb::PEGTransformerFactory::TransformSelectStatementInternalRule(...) transform_select.cpp:59
Address 0x7f309b8190f0 is located in stack of thread T0 at offset 240 in frame
    #0 ... duckdb::PEGTransformerFactory::TransformSelectStatementInternalRule(...) transform_select.cpp:52
  This frame has 11 object(s):
    [240, 320) 'cte_map' (line 54) <== Memory access at offset 240 is inside this variable
SUMMARY: AddressSanitizer: stack-use-after-return /usr/include/c++/13/bits/stl_iterator.h:1077 in __gnu_cxx::__normal_iterator<...>::__normal_iterator(...)
==1871591==ABORTING
```

### OS:

### DuckDB Version:

### DuckDB Client:

CLI (Command Line Interface)

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

alonfaraj:
Fixes #25173.

`SelectStatementInternal` pushes a reference to its CTE map onto `stored_cte_map` so an enclosing PIVOT can find those CTEs, but nothing ever popped it. Once the subquery's `WITH` clause went out of scope the stale reference was still there, and
`ExtractCTEsRecursive` read it:

```sql
PIVOT (WITH x AS (SELECT 1 AS col) SELECT * FROM x) ON col;
```

Stack-use-after-return in the recursive transformer, heap-use-after-free in the trampoline one. Fix is to pop the scope when the select statement is done with it (in the trampoline path it spans two `Finalize` calls, so `manual_state = 2` marks that a pop is owed).

Also means CTEs are properly scoped now, an inner subquery's CTEs no longer leak into an enclosing PIVOT.

## Test
Regression test in `test/sql/pivot/pivot_subquery.test`.


test/sql/pivot/pivot_subquery.test
```diff
@@ -35,6 +35,12 @@ ON empid USING SUM(COALESCE("2000_1",0) + COALESCE("2000_2",0) + COALESCE("2000_
 ----
 29400	142200
 
+# PIVOT over a subquery that has its own CTE
```
Dtenwolde, line 38: I noticed an edge case that now causes the trampoline mode and the recursive one to have different behaviour. Could you add a test case showing that the CTE scope remains active while transforming result modifiers? 

```sql
query I
WITH x AS (SELECT 1 AS col)
SELECT 42
ORDER BY (SELECT * FROM (PIVOT x ON col));
----
42
```

It succeeds on v1.5.5 and with the trampoline mode, but not with the recursive implementation.

src/parser/peg/transformer/transform_select.cpp
```diff
@@ -119,13 +122,19 @@ PEGTransformerFactory::FinalizeSelectStatementInternalTrampoline(PEGTransformer
                                                                  TransformStackFrame &frame) {
 	if (frame.manual_state == 0) {
 		auto &cte_map = frame.GetResult<CommonTableExpressionMap>(0);
+		frame.manual_state = 1;
```
Dtenwolde: I understand why this needs to run in stages, but the `manual_state = 0, 1, and 2` are a bit hard to follow. Do we need state 2 for instance? Result slot 0 still contains the CTE map, so couldn't we check whether that map is empty to decide whether the CTE scope needs to be popped?

src/parser/peg/transformer/transform_select.cpp
```diff
@@ -53,12 +53,15 @@ unique_ptr<SelectStatement> PEGTransformerFactory::TransformSelectStatementInter
...
+	bool pushed_cte_scope = !cte_map.map.empty();
+	if (pushed_cte_scope) {
 		transformer.stored_cte_map.push_back(cte_map);
 	}
 	auto select_statement = transformer.Transform<unique_ptr<SelectStatement>>(list_pr.Child<ListParseResult>(1));
 
-	if (!cte_map.map.empty()) {
+	if (pushed_cte_scope) {
+		// the CTEs are only visible while transforming this select statement
+		transformer.stored_cte_map.pop_back();
```
Dtenwolde, line 65: I think this needs to be moved to below the modifier transformation

Dtenwolde: See the test below

Dtenwolde (commented):
Thanks for the PR! 
Couple of comments :)

src/parser/peg/transformer/transform_select.cpp
alonfaraj: Nope, gone!
 The pop is derived from the map in slot 0 and `manual_state` is back to 0/1. I check it with `GetResult` before the `TakeResult` so no dangling reference sits in `stored_cte_map`.

test/sql/pivot/pivot_subquery.test
alonfaraj: Added verbatim. Fails in recursive mode without the modifier fix, passes in both modes now.

src/parser/peg/transformer/transform_select.cpp
alonfaraj: Moved the pop and the `std::move` into the node below the modifier transformation, so the scope survives `ORDER BY` / `LIMIT` subqueries.

Dtenwolde (approved):
Thanks!

lnkuiper:
A merge conflict has popped up, could you have a look?

alonfaraj:
@lnkuiper 
Rebased on top of `v2.0-cyanoptera `. 

Only `transform_select.cpp` conflicted, from the `TransformStackFrame`/`TransformStack` -> `GeneratedTransformProcess` refactor. Same logic, `frame` -> `process`. Pop semantics still hold: `PushChild` is LIFO and `finalize` only runs once the queue drains, so the scope still covers the body and the modifiers.

Built clean, pivot + peg_parser groups green in both parser modes.

PR #25191, opened 2026-08-31
Add a session-scoped refresh overlay for secret storages

### LINKED ISSUE #195: Using STS assume role with persistent secret fails
I have the following persistent secret configured on session initialization. 

```
CREATE OR REPLACE PERSISTENT SECRET persistent_credential_chain (
      TYPE s3,
      PROVIDER credential_chain,
      CHAIN 'sts',
      assume_role_arn 'arn:aws:iam::x:role/my-role',
      external_id 'generic-role',
      region 'us-west-2'
  );
```

This configuration results in the following error. 
```
Exception thrown while trying to refresh secret persistent_credential_chain. To fix this, please recreate or remove the secret and try again. Error: 'Invalid Input Error: Cannot create temporary secrets in a persistent secret storage!'
```

This looks like a bug to me. When using `sts`, credentials are automatically refreshed and it seems that it replace existing credentials with persistent storage.

kinghuang:
Enables the in-place credential refresh that duckdb/duckdb-httpfs#412 asks for, and resolves duckdb/duckdb-httpfs#195. Opening this for direction — happy to adjust the shape.

## Problem

A credential refresh re-resolves material that has gone stale. That material is session-local, but today an extension has only one way to record it: call `CreateSecret` again and let it write back to the storage the secret came from. That is wrong in both directions:

* **read-only storage session** → the write throws, so a read the fresh credentials could have served fails instead (duckdb/duckdb-httpfs#195, for an `sts`-chain persistent secret);
* **writable shared storage** → the refresh silently persists to a store other sessions read. A long-lived session was observed resurrecting rows another session had deleted hours earlier.

Landing the refreshed copy in some *other* existing storage does not work either: two copies of one name in two storages is exactly the state `GetSecretByName` and `DROP SECRET` report as ambiguous.

## Change

A `refresh` storage that holds the refreshed copy for the life of the session and shadows the original.

**Shadowing** is the new `SecretStorage` property: a shadowing storage holds a transient copy of a secret that also lives elsewhere, so by-name resolution reaches it first instead of counting it as a second match, and a `DROP SECRET` takes the shadow along with the secret it shadows.

```cpp
virtual bool Shadowing() const {
	return false;
}
```

Implementation notes:

* it reuses `ConnectionSecretStorage` — that per-`ClientContext` container already gives session lifetime, crash-robust cleanup and transaction rollback. Its `RegisteredStateManager` key is now derived from the storage name so the two instances keep separate secrets;
* its tie-break offset is `1`, below every storage but the transaction one. `LoadSecretStorage` already rejects a colliding offset, so a refreshed secret outranks its origin whatever offset that origin chose;
* `CREATE SECRET … IN refresh` is rejected in `BindCreateSecret` — the overlay is written by the refresh path, not by SQL;
* both copies stay visible in `duckdb_secrets()`, so a user can see that a refresh shadowed a stored secret:

```
┌───────────────┬────────────┬────────────┐
│     name      │ persistent │  storage   │
├───────────────┼────────────┼────────────┤
│ persistent_s3 │ true       │ local_file │
│ persistent_s3 │ false      │ refresh    │
└───────────────┴────────────┴────────────┘
```

## Consumer

duckdb/duckdb-httpfs#416 is the httpfs side and is what exercises this. Its `GenerateRefreshSecretInfo` becomes:

```cpp
if (result.persist_type != SecretPersistType::TRANSACTION) {
	result.persist_type = SecretPersistType::TEMPORARY;
	result.storage_type = Identifier(SecretManager::REFRESH_STORAGE_NAME);
}
```

so a refresh never writes to a secret storage — one rule, for every storage.

That PR also stands against an alternative, duckdb/duckdb-httpfs#413, which fixes the same correctness bug with no core change by routing the refresh to the transaction storage. It works, but the copy dies at each commit. Measured over three reads in three transactions against a mock S3 server: **12 wasted stale-credential round-trips** and three credential-chain resolutions, versus **4 and one** with this overlay — the same cost as a `memory`-backed secret. Maintainers can pick either; this is the one that needs core.

## Tests

Core suites run green: `test/sql/secrets/*` (including `connection_secret_storage.test` and `create_transaction_secret.test`), the hidden `[secret]` C++ tests (custom secret storage, tie-break penalty collisions, storage name collisions), `[api]` (171 cases), and `test/sql/catalog/*` (148 cases).

I have not added core-side tests yet — the behaviour is currently covered from the httpfs side. Glad to add them here once the shape looks right to you.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/include/duckdb/main/secret/secret_manager.hpp
+++ b/src/include/duckdb/main/secret/secret_manager.hpp
@@ -101,6 +101,8 @@ public:
 	static constexpr const char *TRANSACTION_STORAGE_NAME = "transaction";
 	//! Connection-scoped storage: secrets live only for the creating connection and are reaped when it closes.
 	static constexpr const char *CONNECTION_STORAGE_NAME = "connection";
+	//! Session-scoped storage holding refreshed copies of secrets that live in another storage
+	static constexpr const char *REFRESH_STORAGE_NAME = "refresh";
 
 	//! Static Helper Functions
 	DUCKDB_API static SecretManager &Get(ClientContext &context);
--- a/src/include/duckdb/main/secret/secret_storage.hpp
+++ b/src/include/duckdb/main/secret/secret_storage.hpp
@@ -40,6 +40,7 @@ public:
 	//! Default storage backend offsets. Lower offsets win tie-breaks (equal path-match score), so the connection-scoped
 	//! storage is preferred over the global ones when a connection secret and a global secret match equally well.
 	static const int64_t TRANSACTION_STORAGE_OFFSET = 0;
+	static const int64_t REFRESH_STORAGE_OFFSET = 1;
 	static const int64_t CONNECTION_STORAGE_OFFSET = 5;
 	static const int64_t TEMPORARY_STORAGE_OFFSET = 10;
 	static const int64_t LOCAL_FILE_STORAGE_OFFSET = 20;
@@ -74,6 +75,12 @@ public:
 
 	virtual bool Persistent() const {
 		return persistent;
+	}
+
+	//! Whether this storage shadows the others. A shadowing storage holds a transient copy of a secret that also lives
+	//! in another storage, so by-name resolution reaches it first instead of reporting the two copies as ambiguous.
+	virtual bool Shadowing() const {
+		return false;
 	}
 
 protected:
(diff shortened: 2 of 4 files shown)


kinghuang:
Closing this one.

Per [the comment on duckdb/duckdb-httpfs#416](https://github.com/duckdb/duckdb-httpfs/pull/416#issuecomment-5568030872), this is not the shape you want — you are planning a proper structural fix to the refresh mechanism rather than an overlay layered on top of the current one. That makes sense, and there is no point in this sitting in your review queue in the meantime.

Happy to do the httpfs side once your core change lands — I will watch duckdb/duckdb-httpfs#412 and duckdb/duckdb-httpfs#416. The failure modes this was trying to address are written up in duckdb/duckdb-httpfs#412 if any of it is useful input.

PR #25193, opened 2026-08-31
Fix duplicate table filter cardinality estimation

xuke-hat:
The current implementation of `RelationStatisticsHelper::ExtractGetStats` first calls `get.EstimateCardinality` to get the base cardinality. If the `get` has table filters, it then estimates the output cardinality after applying those filters.

However, the result of `LogicalGet::EstimateCardinality` may already include the effect of the table filters. As a result, `ExtractGetStats` can apply the same filters twice and produce an incorrect estimate.

The main issue is that `LogicalGet::EstimateCardinality` does not distinguish between the source cardinality and the output cardinality of `LogicalGet`. This PR adds `LogicalGet::EstimateSourceCardinality` to return the cardinality of the source itself.

In the following example, the table filter selectivity is applied twice. The pre-aggregation optimizer estimates `fact` as having only `100 * 0.2 * 0.2 = 4` rows and incorrectly skips pre-aggregation. After the fix, pre-aggregation is selected successfully.

```sql
CREATE TABLE dim(k INTEGER);
INSERT INTO dim VALUES (1), (2);

CREATE TABLE fact(k INTEGER);
INSERT INTO fact SELECT 1 + i % 2 FROM range(100) t(i);

ANALYZE;

EXPLAIN
SELECT o.k, count(f.k)
FROM dim o
LEFT JOIN fact f ON o.k = f.k AND f.k > 0
GROUP BY o.k;
```

After this fix, an existing test that previously did not trigger pushdown now does, which is the expected behavior.

### DIFF
--- a/src/include/duckdb/planner/operator/logical_get.hpp
+++ b/src/include/duckdb/planner/operator/logical_get.hpp
@@ -87,6 +87,8 @@ public:
 	vector<ColumnIndex> &GetMutableColumnIds();
 	vector<ColumnBinding> GetColumnBindings() override;
 	idx_t EstimateCardinality(ClientContext &context) override;
+	//! Estimate cardinality from the source without using this operator's cached estimate.
+	idx_t EstimateSourceCardinality(ClientContext &context);
 	bool TryGetStorageIndex(const ColumnIndex &column_index, StorageIndex &out_index) const;
 	void SetScanOrder(unique_ptr<RowGroupOrderOptions> options);
 	void SetPartitionsToScan(vector<idx_t> partition_indices);
--- a/src/optimizer/relation_statistics/relation_statistics_scan.cpp
+++ b/src/optimizer/relation_statistics/relation_statistics_scan.cpp
@@ -133,7 +133,7 @@ DistinctCount RelationStatisticsHelper::GetDistinctCount(LogicalGet &get, Client
 
 RelationStats RelationStatisticsHelper::ExtractGetStats(LogicalGet &get, ClientContext &context) {
 	RelationStats result;
-	auto base_table_cardinality = get.EstimateCardinality(context);
+	auto base_table_cardinality = get.EstimateSourceCardinality(context);
 	auto cardinality_after_filters = base_table_cardinality;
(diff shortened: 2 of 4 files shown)


Mytherin:
Thanks - great to fix this but this seems like a work-around rather than a proper fix. `LogicalGet::EstimateCardinality` should return the cardinality estimate of the operator - rather than having there now being two ways / places where this is being computed. Can we instead fix `LogicalGet::EstimateCardinality` (or the calls that lead to this becoming incorrect / filters being applied twice)?

xuke-hat:
@Mytherin  Thanks for the guidance. Before making changes, could you clarify the intended design of `EstimateCardinality` and `RelationStatisticsHelper`? 

Currently, `EstimateCardinality` seems to be a simple estimate. The base implementation just takes the maximum cardinality of the children and caches it, so `LogicalFilter`, which does not override it, does not estimate filter selectivity. Also, the uncached path of `LogicalGet::EstimateCardinality` uses the source cardinality without estimating `table_filters`.

In contrast, `RelationStatisticsHelper::ExtractGetStats` performs a more detailed estimate using column statistics and `table_filters`, and the join-order optimizer subsequently writes that result back into the same `estimated_cardinality` field.

Therefore, an `estimated_cardinality` may come from the cheap operator-level estimate or the more detailed relation-statistics path. For this bug, `ExtractGetStats` treats `get.EstimateCardinality()` as the pre-filter source cardinality, but it may return a post-filter estimate written by an earlier relation-statistics extraction.

Should `EstimateCardinality()` remain a cheap and potentially inaccurate fallback, with the more detailed estimates kept in `RelationStatisticsHelper`? Or should `EstimateCardinality()` become the authoritative estimate and reuse the logic from `ExtractGetStats()`? 

Once I understand the intended semantics of these estimates, I can fix this issue in the appropriate layer.

xuke-hat:
@lnkuiper Sorry for the ping. I think you may be familiar with this part of the optimizer. Could you clarify the intended design of `EstimateCardinality()` and `RelationStatisticsHelper` here?

PR #25195, opened 2026-08-31
Fix cache freshness exclusiveness

dentiny:
I'm checking [HTTP RFC](https://www.rfc-editor.org/rfc/rfc9111.html#section-4.2-8), the freshness definition is exclusive (aka, freshness_lifetime **>** current_age).


Mytherin:
Same as with https://github.com/duckdb/duckdb/pull/25196 - let's PR this to `httpfs` separately? We can rework this PR to be only the changes to the external file cache.

dentiny:
> let's PR this to httpfs separately? 

This PR touches both core and extension unfortunately :(

Mytherin:
Are those fully connected? It seems this is just changing `<` to `<=` in core, and doing some more complex things in `httpfs`. Can they not be submitted separately?

dentiny:
> Same as with #25196 - let's PR this to `httpfs` separately? We can rework this PR to be only the changes to the external file cache.

Done! I have merged required changes in https extension; this PR bumps up the extension sha and only keeps changes in DuckDB core

PR #25196, opened 2026-08-31
Do not share private http response

dentiny:
According to the [HTTP RFC](https://www.rfc-editor.org/rfc/rfc9111.html#name-private), 
> The unqualified private response directive indicates that a shared cache MUST NOT store the response (i.e., the response is intended for a single user). It also indicates that a private cache MAY store the response

In the current implementation, we don't respect `private` header which means it could be shared among users; in this PR I took the conservative but correct approach to not cache response with `private` header.

Alternatives considered and potential followup items: from the spec it's allowed to cache and reuse the response from the same user, there're two ways I could think of
- Implement a requester-aware cache, similar to what `Authorization` requires; for example record bearer token as part of the cache key
- Store the cached response inside of `ClientContext` so it's used similar to what we did for `HTTPMetadataCache` in httpfs extension; the caveat is block cache is much larger than metadata entries, so we need to do better memory management (i.e., with buffer pool or object cache)

### DIFF
--- a/.github/patches/extensions/httpfs/0004-cache-metadata.patch
+++ b/.github/patches/extensions/httpfs/0004-cache-metadata.patch
@@ -1,4 +1,4 @@
-commit a25824e34450b334fe246f4c12958deea416254b
+commit d9d2f8844f67c565fa12955a53031855ef749dfa
 Author: dentinyhao <dentinyhao@gmail.com>
 Date:   Tue Aug 4 02:50:57 2026 -0700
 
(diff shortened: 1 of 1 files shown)


Mytherin:
Thanks - but could you PR this to the httpfs repo directly? Putting changes to repos in patches is likely to get lost - patches are more intended for fixes that are needed to get CI working in this repo.

dentiny:
> but could you PR this to the httpfs repo directly?

Sure! Let me bump up the extension directly

dentiny:
code change merged into https://github.com/duckdb/duckdb-httpfs/pull/428, no need for this PR

PR #25197, opened 2026-08-31
Make round_even exact for DECIMAL

tanner-lai:
round_even's macro has 2 issues that are fixed in this PR

1. sometimes DECIMALs are not rounded properly 
2. it always returns a DOUBLE, unliked round which preserves the input type 

```sql
(x, n) AS CASE ((abs(x) * power(10, n+1)) % 10) WHEN 5 THEN round(x/2, n) * 2 ELSE round(x, n) END
```

Since power returns a DOUBLE and gets multiplied with x, DECIMAL inputs will check `WHEN 5` as doubles, which can sometimes produce incorrect results depending on the double representation:

```sql
SELECT round_even(128.045, 2);   -- 128.05, expected 128.04
SELECT round_even(128.055, 2);   -- 128.06, correct
```
in this example 128.045 multiplied by power(10, 2 + 1) and stored as a double gives 128044.99999999999 while 128.055 gives 128055.0, so only the latter passes `WHEN 5`

```sql
  SELECT (abs(128.045) * power(10, 3)) % 10;   -- 4.999999999985448
  SELECT (abs(128.055) * power(10, 3)) % 10;   -- 5.0
```

This change makes round_even a scalar function like round while factoring out some code in common between them. Added a couple test cases to `round` as well since that was not testing any kind of rounding policy.

It didn't seem reasonable to keep round_even as a macro due to the branching logic for the different types, e.g. DECIMAL needs logic executed in integral types but DOUBLE would need it in floating point.


extension/core_functions/scalar/math/numeric.cpp
```diff
@@ -897,19 +897,63 @@ ScalarFunctionSet TruncFun::GetFunctions() {
...
+	if (remainder > half || (remainder == half && ROUND_POLICY::RoundsAway(quotient))) {
+		quotient = UnsafeNumericCast<T>(input < 0 ? quotient - 1 : quotient + 1);
+	}
+	return quotient;
+}
+
+struct RoundHalfAwayFromZero {
+	static constexpr const char *Name = "ROUND";
+
+	static constexpr double Nearest(double value) {
```
lnkuiper: Windows CI is failing because of the `constexpr` here

extension/core_functions/scalar/math/numeric.cpp
```diff
@@ -897,19 +897,63 @@ ScalarFunctionSet TruncFun::GetFunctions() {
...
+		return true;
+	}
+};
+
+struct RoundHalfToEven {
+	static constexpr const char *Name = "ROUND_EVEN";
+
+	static double Nearest(double value) {
+		// std::nearbyint rounds half to even
+		return std::nearbyint(value);
```
lnkuiper: Codex is saying this function relies on the calling thread's rounding mode, which can apparently be set using `fesetround`, which sounds a bit scary. I think it's better to just implement this one ourselves.

extension/core_functions/scalar/math/functions.json
```diff
@@ -236,6 +236,14 @@
...
         "type": "scalar_function_set"
     },
+    {
+        "name": "round_even",
+        "parameters": "x,precision",
+        "description": "Rounds x to s decimal places, rounding halfway cases to the nearest even digit",
+        "example": "round_even(24.5, 0)",
+        "type": "scalar_function_set",
+        "aliases": ["roundbankers"]
+    },
```
lnkuiper, line 246: The functions also need to be added to `extension_entries.hpp` for autoloading to work

lnkuiper (commented):
Thanks! I've left some comments:

extension/core_functions/scalar/math/numeric.cpp
tanner-lai: Makes sense, replaced with RoundToNearestEven. fyi did notice that there is other usage of nearbyint in the codebase but I didn't touch those.

tanner-lai:
> Thanks! I've left some comments:

Thank you, think I've addressed the comments. There's a CI failure for arm64 linux release compat, it seems unrelated to my changes but I'm not quite sure how to interpret the failure

extension/core_functions/scalar/math/numeric.cpp
```diff
@@ -897,19 +898,62 @@ ScalarFunctionSet TruncFun::GetFunctions() {
...
 //===--------------------------------------------------------------------===//
 namespace {
+
+template <class T, class ROUND_POLICY>
+inline T RoundDivide(T input, T power_of_ten) {
+	// power_of_ten is ten raised to the digits being dropped and is at least ten because the binder replaces round
+	// with ScalarFunction::NopFunction if the round would not drop digits
+	D_ASSERT(power_of_ten >= 10);
+	T quotient = UnsafeNumericCast<T>(input / power_of_ten);
+	T remainder = UnsafeNumericCast<T>(input % power_of_ten);
```
lnkuiper: Can we avoid computing both `/` and `%` here? For hugeint-backed decimals, this invokes `Hugeint::DivMod` twice. Once that's fixed, this is ready to go.

Dtenwolde:
@tanner-lai Make sure to undraft your PR so that CI can run :)

tanner-lai:
> @tanner-lai Make sure to undraft your PR so that CI can run :)

Thanks! looks like CI is green

PR #25198, opened 2026-08-31
[Parser] Replace `PEGToken` with `PEGExpression`

Tishj:
This PR is a cherry picked subset of changes from #24919 

We move `PEGToken` to a private detail of the `PEGParser`, replacing it with `PEGExpression` in public-facing methods.

As a result of this change, the `MatcherFactory` logic is impacted, and replaced to a queue-based iterative creation pattern. (to better deal with highly recursive grammars and malicious crafted grammars)

The `ParsedGrammarKeywordHelper` used to use `PEGToken` as well, this has also been moved to the new `PEGExpression`.


Mytherin:
Thanks!

PR #25200, opened 2026-09-01
Prune row group for boolean type under all NULL cases

dentiny:
Followup PR for https://github.com/duckdb/duckdb/pull/25144, there's one case I didn't handle, for example,
```sql
memory D ATTACH '/tmp/bool_ref_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE nullable_tb AS SELECT
       CASE WHEN range // 2048 = 0 THEN CASE WHEN range % 2 = 0 THEN TRUE ELSE NULL END
            WHEN range // 2048 = 1 THEN CASE WHEN range % 2 = 0 THEN FALSE ELSE NULL END
            WHEN range // 2048 = 2 THEN (range % 2 = 0)
            ELSE NULL END AS b
     FROM range(8192);

db D EXPLAIN ANALYZE SELECT count(*) FROM nullable_tb WHERE b;
╭─ Summary ───────────╮
│ Total Time: 0.0010s │
╰─────────────────────╯
│ 1 row                  0µs │
╰──────────────┒┎────────────╯
╭─ Table Scan ─┚┖────────────╮
│ Table: db.main.nullable_tb │
│ Filters: b                 │
│ Row Groups Scanned: 3 / 4  │
│ 2,048 rows             0µs │
```
In the above example, the stats distribution is
Row group 0: b is TRUE or NULL
Row group 1: b is FALSE or NULL
Row group 2: b alternates TRUE/FALSE
Row group 3: b is NULL
so we should only access two row groups instead of three; the root cause of which is we didn't handle all NULL cases.


Mytherin:
Thanks!

Mytherin:
It might be nice to turn `FilterPropagateResult` into a struct that has a bunch of methods - e.g.:

```cpp
struct FilterPropagateResult {
public:
	bool AlwaysTrueAndValid() const;
	bool AlwaysFalseAndValid() const;
	bool AlwaysFalseOrNull() const;
	bool AlwaysTrueOrNull() const;
private:
      // enum
};
```

I think right now there's a bunch of places where `FilterPropagateResult` is not correctly used (e.g. where `AlwaysFalseAndValid` is checked, when `AlwaysFalseOrNull` is meant). Having a struct here with explicit methods would likely help detect and minimize these kinds of bugs / missed optimizations.

PR #25201, opened 2026-09-01
Cleanup unused else branch

dentiny:
Random find, the else branch shouldn't be accessed anyway (`partial_reuse` already checked several lines above).


Mytherin:
Thanks!

PR #25203, opened 2026-09-01
fix(cmake): Stop CMake configure on FATAL_ERROR messages

darion-yaphet:
## Summary
  - Fix four `message(FATAL_ERROR, "...")` calls in the CMake build.
  - CMake does not treat the comma as an argument separator, so the first argument became `FATAL_ERROR,` and configure printed the text instead of
  aborting.
  - Invalid `OSX_BUILD_UNIVERSAL` / `OSX_BUILD_ARCH` on non-Apple hosts, unsupported extension static builds, and a remote extension without
  `GIT_TAG` now fail configure as intended.
  ## Test plan
  - [ ] `cmake -S . -B build/check -DOSX_BUILD_UNIVERSAL=ON` on Linux fails with a FATAL_ERROR
  - [ ] A normal `make reldebug` / Ninja configure still succeeds


Mytherin:
Thanks!

PR #25204, opened 2026-09-01
Heap-based matcher

Dtenwolde:
Fixes: https://github.com/duckdblabs/duckdb-internal/issues/10518
Fixes: https://github.com/duckdblabs/duckdb-internal/issues/9555
Part of: https://github.com/duckdblabs/duckdb-internal/issues/6380

This PR introduces a heap-based `Matcher` to be used together with the heap-based `Transformer` introduced in: https://github.com/duckdb/duckdb/pull/23578  

The implementation is experimental, behind the following setting, and disabled by default: `SET debug_heap_based_parser=true;`

Because of the recursive nature of PEG, there were cases where we ran out of stack space earlier than we used to with the old parser. Instead of relying on the stack, we therefore need an implementation where we can manage the stack ourselves. This does come with a performance penalty in the first implementation of a roughly 3x parser slowdown. 
Running TPC-H and TPC-DS SF=0(!), 50 runs per suite: 

**This is just measuring the parser timing**
| Suite | Stack | Heap | Heap overhead |
|---|---:|---:|---:|
| TPC-H, 1,100 parses | 684.9 ms total / 623 µs per query | 2,129.6 ms / 1.936 ms | 3.11× (+211%) |
| TPC-DS, 4,950 parses | 6,022.9 ms / 1.217 ms | 17,193.3 ms / 3.473 ms | 2.85× (+185%) |

For SF 1, the total overhead for TPC-H is about 4%, TPC-DS 8%, which is "not great, not terrible" 

The heap-based implementation is behind a setting `SET debug_heap_based_parser=true;`, which now includes both the `Matcher` and the `Transformer`. I renamed the previous `SET debug_transformer_trampoline_style=true;` rather than add an extra setting. The setting is also used with autocomplete (`RECOGNIZE_ONLY` mode), tokenization and other parser-check paths. 

The heap matcher implements the `Optional`, `Choice`, `List` and `Repeat` matchers with heap-allocated frames. Packrat memoization is preserved through `Lookup`.

The maximum expression depth is still guarded by a setting, so excessive nesting that would exceed the guard triggers a parser error.


src/parser/peg/matcher_stack.cpp
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+private:
+	const ChoiceMatcher &choice_matcher;
+	unique_ptr<MatchState> child_state;
+	idx_t child_index = 0;
+	optional_idx start_offset;
+};
+
+class ListMatchStackFrame : public MatchStackFrame {
```
Tishj: Can we please put these in separate files?

Dtenwolde: Oeh right!

Dtenwolde:
Thanks to codex we already discovered a couple of optimizations related to allocations: 1. Terminal frames don't need to be a separate frame, but can instead be inserted directly in the parent and 2. The choice matcher's result could also be inlined. There's a couple more, but I want to save these for follow-up PRs, rather than overload this one

src/include/duckdb/parser/peg/matcher.hpp
```diff
@@ -166,6 +167,7 @@ struct MatchState {
 	IdentifierCaseMode identifier_case_mode = IdentifierCaseMode::PRESERVE_CASE;
 	ParserPackratCache *packrat_cache;
 	MatchMode mode;
+	bool use_heap_based_parser;
```
Mytherin, line 170: Do we need this or could we just switch to it immediately?

Mytherin: I think keeping the setting is fine for now - but we should make sure to remove it before v2.0

Mytherin:
Thanks! Great stuff

src/include/duckdb/parser/peg/matcher.hpp
Dtenwolde: I didn't want to make the switch immediately, given the 3x parser slowdown. But I can make the switch in the next PR :)

PR #25205, opened 2026-09-01
[Dev][Parser] `TypedTransformResult` result fetching/unwrapping changes

Tishj:
Reworks a change made by #24858 in preparation for #24919 

In short, we return the wrapped T directly, rather than wrapping it in another `TypedTransformResult`, this is what:

> removes the remaining cross-DLL dependency on the concrete wrapper type

Which I won't pretend to fully understand, but it makes sense to reduce the amount of classes that are used across linking boundaries.


Mytherin:
Thanks!

Tishj:
Closing because it'll cause merge conflicts with #24919 
These changes are all present in the mentioned PR regardless.

PR #25209, opened 2026-09-01
TryScanIndex fixes

artjomPlaunov:
Supersedes https://github.com/duckdb/duckdb/pull/24532 (commit included, there were more issues in this function), fixes https://github.com/duckdb/duckdb/issues/24246 as well as some new bugs found. 

For review, I would recommend just reading the new changes without thinking about the old bugs too much, but here are the bugs that were present, and how they were fixed:

Bugs in the old code: 

1) 

```cpp
if (input.column_ids[i] == indexed_columns[0])
```
input column_ids refers to logical column positions, whereas indexes_columns has physical column positions (i.e. this is a logical vs. physical comparison). 

With generated columns, this could lead to 1) a different logical column having the same value as the indexed physical column and 2) 

The fix here is a new find_scan_position function that resolves the physical index column to the scan position. 

2) 

```cpp
if (bound_column_ref_expr.binding.column_index ==
    indexed_columns[0]) {
    bound_column_ref_expr.binding.column_index =
        updated_index_column;
}
```

The expression index is an ordinal into the index's indexes_columns list (which stores physical columns). This could incorrectly skip a rewrite, for example, let's say we have an index on physical column 1, and indexed_columns = [1]. The expression leaf has ref(0) as the expression, i.e. pointing to physical column 1 in the indexed_columns list. So we are comparing 0 == 1, then skip the rewrite. 

This didn't lead to any correctness bugs, but it did lead to missing index scans. 

The fix is to convert the expression ordinal to a physical column id, map it to the current scan position (using find_scan_position introduced above), then rebind the copied expression to that. 

3) 

Finally, and this is the bug fix from https://github.com/duckdb/duckdb/pull/24532, we had: 

```cpp
auto &col =
    column_list.GetColumn(
        LogicalIndex(indexed_columns[0])
    );
```

which again is just a physical vs logical mismatch. The fix here was to just use the physicalIndex overload on GetColumn to get the proper indexed column.


Mytherin:
Thanks! LGTM

artjomPlaunov:
CI failure looks unrelated

PR #25210, opened 2026-09-01
Add support for type constructor signatures and overloads

Maxxen:
This PR adds support for adding a set of `FunctionSignature`s to a type catalog entry/"constructor".

A `TypeCatalogEntry` carried a single opaque function pointer, `bind_logical_type_function_t`, which got a `vector<TypeArgument>` when creating "parameterized" types. Having the logic to inspect/validate the parameters imperatively had some downsides:

  - No overloading. `DECIMAL -> DECIMAL(18,3)` vs `DECIMAL(w) -> scale 0` vs `DECIMAL(w,s)` were all hand crafted branches inside a single callback.
  - lots of duplicated binding logic and inconsistent error messages when validating e.g. arity, NULL, type and cast checks.
  - The support parameters/invariants were completely invisible to the rest of the system, e.g. `duckdb_types()` couldn't say whether a type took modifiers, or what they were called.

We now essentially treat type "constructors" as regular functions, by making them associated with a`FunctionSignature`. This enables us to re-use the same function binding logic to perform e.g. argument type validation, implicit casts, and overload-resolution.

This is _mostly_ preparatory work for better custom type support in the future, but some interesting side-effects of this that are somewhat useful even now:

- `duckdb_types()` emits one row per overload with parameters / parameter_types / varargs. This enables us to also print documentation about types in the `.manual`. `pg_catalog.pg_type` is unaffected as the existing type_oid dedup NULLs the extra rows and the view already filters on it. 

```sql
select distinct(type_name), parameters, parameter_types, varargs 
from duckdb_types() 
where parameters != [] or varargs is not null 
order by all;

┌───────────┬────────────────┬──────────────────────┬─────────┐
│ type_name │   parameters   │   parameter_types    │ varargs │
│  varchar  │   varchar[]    │      varchar[]       │ varchar │
├───────────┼────────────────┼──────────────────────┼─────────┤
│ array     │ [child, size]  │ [TYPE, BIGINT]       │ NULL    │
│ bit       │ [length]       │ [BIGINT]             │ NULL    │
│ bitstring │ [length]       │ [BIGINT]             │ NULL    │
│ bpchar    │ [collation]    │ [VARCHAR]            │ NULL    │
│ bpchar    │ [length]       │ [BIGINT]             │ NULL    │
│ char      │ [collation]    │ [VARCHAR]            │ NULL    │
│ char      │ [length]       │ [BIGINT]             │ NULL    │
│ datetime  │ [precision]    │ [UTINYINT]           │ NULL    │
│ dec       │ [width, scale] │ [UTINYINT, UTINYINT] │ NULL    │
│ decimal   │ [width, scale] │ [UTINYINT, UTINYINT] │ NULL    │
│ enum      │ []             │ []                   │ VARCHAR │
│ geometry  │ [crs]          │ [VARCHAR]            │ NULL    │
│ interval  │ [precision]    │ [UTINYINT]           │ NULL    │
│ list      │ [child]        │ [TYPE]               │ NULL    │
│ map       │ [key, value]   │ [TYPE, TYPE]         │ NULL    │
│ numeric   │ [width, scale] │ [UTINYINT, UTINYINT] │ NULL    │
│ nvarchar  │ [collation]    │ [VARCHAR]            │ NULL    │
│ nvarchar  │ [length]       │ [BIGINT]             │ NULL    │
│ row       │ []             │ []                   │ TYPE    │
│ string    │ [collation]    │ [VARCHAR]            │ NULL    │
│ string    │ [length]       │ [BIGINT]             │ NULL    │
│ struct    │ []             │ []                   │ TYPE    │
│ text      │ [collation]    │ [VARCHAR]            │ NULL    │
│ text      │ [length]       │ [BIGINT]             │ NULL    │
│ timestamp │ [precision]    │ [UTINYINT]           │ NULL    │
│ tuple     │ []             │ []                   │ TYPE    │
│ union     │ []             │ []                   │ TYPE    │
│ varchar   │ [collation]    │ [VARCHAR]            │ NULL    │
│ varchar   │ [length]       │ [BIGINT]             │ NULL    │
└───────────┴────────────────┴──────────────────────┴─────────┘
```

- Better error messages when constructing type expressions, including line information:

```sql
SELECT '0.1'::DECIMAL('hello');
----
Binder Error:
Type "decimal" does not accept type parameters (VARCHAR)
        Candidate definitions:
        decimal()
        decimal(width UTINYINT, scale UTINYINT := 0)

LINE 1: SELECT '0.1'::DECIMAL('hello');
                      ^^^^^^^^^^^^^^^^   

```
```sql
SELECT '0.1'::DECIMAL(NULL);
Binder Error:
Type parameter "width" for type "decimal" cannot be NULL

LINE 1: SELECT '0.1'::DECIMAL(NULL);
                              ^^^^    
```


Mytherin:
Good idea, thanks!

PR #25211, opened 2026-09-01
Do not assume NULL columns on right side of LEFT JOIN

### LINKED ISSUE #24540: Wrong result on  LEFT JOIN ... WHERE col IS NULL (v1.5 regression)
### What happens?

Wrong result on current main. Works correctly in v1.5.

### To Reproduce

```
SELECT * FROM (VALUES (1),(2),(NULL)) AS a(x)
LEFT JOIN (SELECT p.y AS y, q.z AS z
           FROM (VALUES (2),(3),(NULL)) AS p(y)
           LEFT JOIN (VALUES (1),(3),(NULL)) AS q(z) ON p.y <> q.z) AS s
ON a.x <> s.y
WHERE s.z IS NULL;
--   correct: one row, NULL|NULL|NULL
--   duckdb:  three rows — 1|NULL|NULL, 2|NULL|NULL, NULL|NULL|NULL
-- a.x = 1 DOES match s (1 <> 2 is true, and s has y = 2), and all its
-- matches carry z = 1 or 3, so nothing of a.x = 1 survives `z IS NULL`.
```

### OS:

Ubuntu

### DuckDB Version:

current main

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Viktor Leis

### Affiliation:

TUM

- [x] Yes, I have

- [x] Yes, I have

Tmonster:
Fixes https://github.com/duckdb/duckdb/issues/24540. 
Fixes https://github.com/duckdblabs/duckdb-internal/issues/10240

OuterJoinSimplification walks the plan top-down, recording columns that a predicate forces to be NULL  in null_required_columns. This allows the optimizer to rewrite LEFT joins into ANTI joins. When an IS NULL predicate sits above a LEFT join, the `IS NULL` is carried into the join's right child, as a way to assume only NULL values will survive, to apply any optimizations knowing that will hold true. This is incorrect, however, since the LEFT JOIN will extend non-matching rows and produce new NULLS. Therefore, holding onto the `null_required_columns` on the right side of a left join is incorrect, this caused the second join to be rewritten into an ANTI join.
 
Before descending into the side a join can null-extend (the right child for LEFT, the left child for RIGHT, both for OUTER), that side's column bindings are now erased from null_required_columns. The same is done for the right child when the join itself has just been converted to ANTI, since its columns produce no output at all.

This may be a candidate to backport to the v2.0.0 branch once it gets cut

### DIFF
--- a/src/include/duckdb/optimizer/outer_join_simplification.hpp
+++ b/src/include/duckdb/optimizer/outer_join_simplification.hpp
@@ -56,6 +56,7 @@ private:
 	bool HasNullRequiredColumns(const vector<ColumnBinding> &bindings);
 	bool HasRequiredColumns(const vector<ColumnBinding> &bindings);
 	void MarkEliminatedNullColumns(const vector<ColumnBinding> &bindings);
+	void EraseNullRequiredColumns(LogicalOperator &child);
 	vector<ColumnBinding> GetRightBindings(LogicalJoin &join);
 	void SimplifyOuterJoinType(LogicalComparisonJoin &join);
 
--- a/src/optimizer/outer_join_simplification.cpp
+++ b/src/optimizer/outer_join_simplification.cpp
@@ -186,6 +186,12 @@ bool OuterJoinSimplification::HasRequiredColumns(const vector<ColumnBinding> &bi
 	return false;
 }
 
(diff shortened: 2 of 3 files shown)


Mytherin:
Thanks!

PR #25212, opened 2026-09-01
Validate LIST offsets before use

HendrikLambert:
A LIST column stores the cumulative end offset of every row and calculates a length by subtracting neighbouring offsets. That subtraction is only safe if the offsets are non-decreasing and stay within the child column. Without them, a crafted file underflows the length into a near-2^64 child scan or reads past the child column.

## Changes

- Add `GetListLength` (rejects a later offset below its predecessor) and `ValidateListOffset` (rejects an offset past `child_column->GetMaxEntry()`)
- Route every offset consumer through them: `FetchListOffset`, `ScanCount`, `Skip`, and `FetchRows`.
- Guard `FetchListOffset` against a row id past the column count.


Mytherin:
Thanks! Looks good

PR #24852, opened 2026-08-18
Backport DuckDB 2.0 API spellings to 1.5

JelteF:
Makes it possible to write code in "DuckDB 2.0 style" against 1.5.6. This is useful for extensions that want to continue developing for 1.5, while already preparing for DuckDB 2.0.

This changes no existing behaviour, it only adds new APIs that can optionally be called by extensions. The only non-additive part is moving `SQLIdentifier`/`SQLString` out of `exception_format_value.hpp` into the new `sql_identifier.hpp`.

Where 1.5 cannot express what 2.0 does, a `NOTE(backport)` comment records what differs and why.

Added:

- `Identifier`, the `SQLIdentifier`/`SQLQuotedIdentifier`/`SQLString` format helpers, and their `Value` and (de)serializer support
- `StringUtil::CIEndsWith` and `KeywordHelper::WriteQuotedAndEscaped`
- accessors on `Function`, `FunctionSet` and `CreateFunctionInfo`: `GetName`/`SetName`, `GetArguments`, `GetOriginalArguments`, `GetVarArgs`
- `DataChunk::SetChildCardinality`, `CheckCardinality` and `SetCardinalityUnsafe` - all equivalent to `SetCardinality` here, since vectors on this branch carry no size of their own to resize or verify
- name accessors on the `CreateXInfo` classes (`GetTableName`/`SetTableName`, `GetViewName`, `GetIndexName`, `GetTypeName`, `GetSequenceName`, `GetFunctionName`) plus `SetCatalog`/`SetSchema` on `CreateInfo`
- `QualifiedName` constructors, `Catalog()`/`Schema()`/`Name()`, `WithName`, `WithCatalog`, `StripCatalog` and `operator==`
- `Get`/`SetQualifiedName` on `CreateInfo`, `AlterInfo` and `DropInfo`
- `Catalog::GetEntry<T>(context, QualifiedName, ...)`

PR #25214, opened 2026-09-01
Backport #24878: Always quote identifiers in error messages

JelteF:
On main, an `Identifier` interpolated into an exception format string is rendered through `SQLQuotedIdentifier`, so it is always quoted rather than only when quoting is required in SQL. On v1.5 it still went through `SQLIdentifier`, so an extension throwing e.g. `BinderException("Table %s does not exist", identifier)` produced differently quoted error messages on v1.5 than on main. This only matters to extensions that use the new APIs on the v1.5.

This is a follow-up to #24852 to backport some new behaviour from main that was introduced after that initial backport was merged.


Mytherin:
Thanks!

PR #23291, opened 2026-06-15
Add helper iterators around SQLStatements (either from the parser or post pre-processing)

carlopi:
Alternative to https://github.com/duckdb/duckdb/pull/23041

This allows a given query string to be unpacked into multiple statements, extracted and executed one at a time, that means that for example the state of the parser might change.

Example:
```
carlo@MacBook-Pro-2 duckdb-stable % ./build/release/duckdb -unsigned
DuckDB v1.6.0-dev8256 (Development Version, 4523194c2f)
Enter ".help" for usage hints.
memory D QUACK quack quack;
Parser Error:
syntax error at or near "quack"

LINE 1: QUACK quack quack;
              ^
memory D LOAD './build/release/test/extension/loadable_extension_demo.duckdb_extension'; QUACK quack quack;
┌─────────┐
│  quack  │
│ varchar │
├─────────┤
│ QUACK   │
│ QUACK   │
│ QUACK   │
└─────────┘
memory D 
memory D 
```

Where `LOAD loadable_extension_demo.duckdb_extension` changes to the parser are directly visible in the same statement.

PR #25215, opened 2026-09-01
Fix the profiler for the parser

Dtenwolde:
After (I think) https://github.com/duckdb/duckdb/pull/23291 the profile timings for the parser were broken, showing parser time took 0.0000000 seconds. Unfortunately that's too good to be true. 

Now the parser timing is handled again in the `StatementIterator` and properly handed off when a statement is executed.


Mytherin:
Thanks!

PR #25217, opened 2026-09-01
Use tracked allocator for JSON parsing in the cast to VARIANT

JelteF:
Casting a JSON value to VARIANT parsed each value with a bare
three-argument `yyjson_read(data, size, flags)`, without passing a
`yyjson_alc`. The parse tree therefore came from raw `malloc` and was
invisible to DuckDB's memory accounting and `memory_limit` is not
adhered too (possibly causing OOM-kills). This fixes that by using an
explicit allocator for these yyjson calls.


Mytherin:
Thanks!

PR #25218, opened 2026-09-01
Use CompressionSegmentReader for ALP and ALPRD

HendrikLambert:
This PR moves ALP and ALPRD decompression reads to `CompressionSegmentReader`.

## Changes
- Extend `CompressionSegmentReader` with checked byte and array copy helpers.
- Bounds-check ALP and ALPRD segment metadata, vector offsets and compressed payloads.
- Reject invalid ALP and ALPRD encoding parameters, dictionaries and exception metadata.
- Validate ALP and ALPRD scan, skip and row fetch ranges.


src/include/duckdb/storage/compression/alp/alp_scan.hpp
```diff
@@ -122,108 +194,58 @@ struct AlpScanState : public SegmentScanState {
...
-			throw DataCorruptionException("Corrupted ALP segment: exceptions_count (%d) exceeds vector_size (%d)",
-			                              vector_state.exceptions_count, vector_size);
+			ThrowAlpExceptionCountOutOfRange(vector_state.exceptions_count, vector_size);
 		}
 		if (vector_state.v_factor > vector_state.v_exponent) {
-			throw DataCorruptionException("Corrupted ALP segment: v_factor (%d) exceeds v_exponent (%d)",
-			                              vector_state.v_factor, vector_state.v_exponent);
+			ThrowAlpFactorOutOfRange(vector_state.v_factor, vector_state.v_exponent);
 		}
-		if (vector_state.bit_width > sizeof(uint64_t) * 8) {
```
Tishj, line 180: I get the feeling this was and is wrong, I don't think the "frame of reference" has anything to do with this, which is what's now used in the calculation of `MAX_BIT_WIDTH`.

I assume that this was supposed to use `FloatingToExact<T>` instead of hardcoding to `sizeof(uint64_t)`

src/include/duckdb/storage/compression/alp/alp_scan.hpp
```diff
@@ -235,6 +257,13 @@ struct AlpScanState : public SegmentScanState {
 public:
 	//! Skip the next 'skip_count' values, we don't store the values
 	void Skip(ColumnSegment &col_segment, idx_t skip_count) {
+		D_ASSERT(total_value_count <= count);
+		D_ASSERT(skip_count <= count - total_value_count);
+		// Reaching the segment end does not require loading the final partial vector.
```
Tishj: Is there a test for this new optimization?
I'm not sure how much we're actually gaining from this, if you look at `ScanVector` this already exists mostly, no `memcpy` is issued at all when skipping

src/include/duckdb/storage/compression/alprd/alprd_scan.hpp
```diff
@@ -52,81 +65,132 @@ struct AlpRDVectorState {
 
 public:
 	idx_t index;
-	uint8_t left_encoded[AlpRDConstants::ALP_VECTOR_SIZE * 8];
-	uint8_t right_encoded[AlpRDConstants::ALP_VECTOR_SIZE * 8];
+	//! Packed left indexes read from the segment through a bounded vector reader.
+	//! BitpackingPrimitives::UnPackBuffer reads the left stream through aligned DICTIONARY_ELEMENT_TYPE pointers.
+	alignas(AlpRDConstants::DICTIONARY_ELEMENT_TYPE) uint8_t left_encoded[AlpRDConstants::ALP_VECTOR_SIZE * 8];
```
Tishj: Great find 🚀 
I wonder if we can even just change the type from `uint8_t` to `DICTIONARY_ELEMENT_TYPE` ?
And adjust the size accordingly as well

src/include/duckdb/storage/compression/alprd/alprd_scan.hpp
```diff
@@ -52,81 +65,132 @@ struct AlpRDVectorState {
...
+	//! Packed right parts read from the segment through a bounded vector reader.
+	//! BitpackingPrimitives::UnPackBuffer reads the right stream through aligned EXACT_TYPE pointers.
+	alignas(EXACT_TYPE) uint8_t right_encoded[AlpRDConstants::ALP_VECTOR_SIZE * 8];
```
Tishj: Same as above, I think this should never have been a `uint8_t[]` in the first place?

src/include/duckdb/storage/compression/alprd/alprd_scan.hpp
```diff
@@ -263,6 +281,13 @@ struct AlpRDScanState : public SegmentScanState {
 public:
 	//! Skip the next 'skip_count' values, we don't store the values
 	void Skip(ColumnSegment &col_segment, idx_t skip_count) {
+		D_ASSERT(total_value_count <= count);
+		D_ASSERT(skip_count <= count - total_value_count);
+		// Reaching the segment end does not require loading the final partial vector.
```
Tishj: Same comment as I left here for ALP

src/include/duckdb/storage/compression/alp/alp_scan.hpp
HendrikLambert: From what I can see, ALP bit-packs the `uint64_t` FOR deltas derived from the `int64_t` encoded values, not the original floating-point representations. This would mean that for FLOAT, using `FloatingToExact<T>` would set the limit to 32 even though a valid delta can be wider (`0.0f` and `1e10f` produce a 34-bit delta). Current storage versions would use ALP’s per-vector uncompressed fallback for this vector, but I think that is a more recent addition?

HendrikLambert: If the above is correct, I do agree that tying the maximum directly to `FOR_SIZE` isn’t very clean. Would this then work for you?

```cpp
using ENCODED_VALUE_TYPE = uint64_t;
using FRAME_OF_REFERENCE_TYPE = ENCODED_VALUE_TYPE;

static constexpr BIT_WIDTH_TYPE MAX_BIT_WIDTH = sizeof(ENCODED_VALUE_TYPE) * 8;
```

Tishj: Ah I see, then disregard my comment 👍 Thanks for looking into it

Tishj (approved):
Thanks, looks great!

PR #25219, opened 2026-09-01
Improve IN operator propagation result

dentiny:
Hi team, I found the `NOT IN` operator doesn't seem to prune row group as expected when NULL involved; for example,
```sql
memory D ATTACH '/tmp/in_list_prune.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE mixed_nullable AS
         SELECT CASE
             WHEN range < 2048 THEN CASE WHEN range % 4 = 0 THEN NULL ELSE (range % 3 + 1)::INTEGER END
             ELSE (range % 2 + 4)::INTEGER
         END AS i
         FROM range(6144);

-- should only need to access two row groups, but actually access three
db D EXPLAIN ANALYZE SELECT count(*) FROM mixed_nullable WHERE i NOT IN (1, 2, 3);
╭─ Summary ───────────╮
│ Total Time: 0.0015s │
╰─────────────────────╯
╰───────────────┒┎──────────────╯
╭─ Table Scan ──┚┖──────────────╮
│ Table: db.main.mixed_nullable │
│ Filters: NOT (i IN (1, 2, 3)) │
│ Row Groups Scanned: 3 / 3     │
│ 4,096 rows                0µs │
```

### DIFF
--- a/src/planner/filter/expression_filter.cpp
+++ b/src/planner/filter/expression_filter.cpp
@@ -623,7 +623,7 @@ static FilterPropagateResult CheckInOperatorStatistics(optional_ptr<ClientContex
 		result = FilterPropagateResult::FILTER_ALWAYS_TRUE;
 	}
 	if (result == FilterPropagateResult::FILTER_ALWAYS_TRUE && filter_stats->CanHaveNull()) {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks! LGTM

PR #25221, opened 2026-09-01
Rework `ExpressionBinder` to use single-pass scope resolution

Maxxen:
This is a follow-up/superseding PR to https://github.com/duckdb/duckdb/pull/25150, which attempts to provide a more elegant solution to the initial problem. Instead of storing already-bound expressions in a map keyed by the parsed expression pointer - this PR avoid the need to "retry' binding entirely by instead binding/qualifying expressions into the correct "scope" immediately. This PR might actually be a negative diff - most of the added code is merging error messages and new tests.

## Motivation

As explained in https://github.com/duckdb/duckdb/pull/25150, DuckDB resolved out-of-scope column references by failing and retrying. `ExpressionBinder::BindCorrelatedColumns` walked the enclosing binders outward, re-qualifying and re-binding the whole expression at depth+1 until something stuck.

Because an expression could be bound repeatedly, bound state had to survive between attempts, so binding wrote bound nodes into the parsed tree (`BoundSubqueryNode` overwrote a parsed `SelectStatement`; `BoundExpressionMap` (and the `BoundExpression before it) was a side table keyed by parsed-node identity). That in turn meant a parsed statement couldn't always be copyable or printable mid-bind, and guards existed throughout the binder to stop half-bound expressions from being mistaken for finished ones.

This PR removes the "retry loop" and then everything that existed to serve it.

## What changes

A name that does not resolve in the current scope is now looked up in the enclosing scopes where it occurs, and bound by the scope that owns it (using `DispatchToScope`) so that scope's own semantics still apply. That enables us to preserves lateral detection, HAVING's correlated-subquery errors, outer-GROUP BY matching and table-function fallthrough without re-deriving any of them.
- An aggregate belongs to the innermost level in which one of its arguments resolves a column.
- GROUPING belongs to the innermost level that groups by all of its arguments.

## Consequences

__The parsed syntax tree is never mutated with bound state__. `BoundSubqueryNode` and `BoundExpressionMap` are now both gone (`QueryNodeType::BOUND_SUBQUERY_NODE` remains for now though to not shift the enum values).

This is a precondition to make the `ParsedExpression`/parsed syntax tree fully immutable. And potentially making expression/binding non-recursive is now a lot easier too.

__Bound children flow by return value__. `Bind` returns `BindResult`, and `BindChild` returns the child. I've marked them with `[[nodiscard]]` to ensure the new semantics are known and people don't accidentally drop the return type thinking the input will be mutated.

1. Correlated windows now work. `SELECT i, (SELECT SUM(i + 1) OVER ()) FROM integers` returns exactly what the nested-subquery workaround returns, across `WHERE`, `ORDER BY` and `PARTITION BY`. This resolves the "`#FIXME - why not?`" in `test_correlated_window.test`.
2. Correlated lambda-expressions now work, e.g. `sum(list_transform([1,2], lambda y: y + t.x)[1])` used to error with `Referenced table "t" not found!`!
3. A window bound by a binder that doesn't handle windows raised `NotImplementedException: Unimplemented expression class: WINDOW`, but now gets a proper error naming the clause that rejected it.
4. A constant-only aggregate reports the clause it appears in, so wrapping a query in a subquery no longer changes the message.` SELECT (WITH t AS (SELECT 3 WHERE count(*)) ...)` now prints "`WHERE clause cannot contain aggregates`", matching the same query unwrapped.
5. The lambda/JSON fallback is basically the only remaining "retry" case, but it no longer inherits a failed attempt's bindings. `->` is ambiguous between lambda and JSON syntax, so the binder tries lambda and falls back, but the fallback was reusing whatever the failed lambda attempt had already bound, including a `BoundLambdaExpression` as a JSON argument. Since it now performs a full rebind, the old state is discarded.


src/planner/binder/expression/bind_case_expression.cpp
```diff
@@ -9,19 +9,21 @@ namespace duckdb {
 BindResult ExpressionBinder::BindExpression(CaseExpression &expr, idx_t depth) {
 	// first try to bind the children of the case expression
 	ErrorData error;
+	vector<pair<unique_ptr<Expression>, unique_ptr<Expression>>> checks;
```
Tishj, line 12: nit:
```c++
namespace {

struct CaseExpressionComponent {
    unique_ptr<Expression> when_expr;
    unique_ptr<Expression> then_expr;
};

} // namespace
```

src/include/duckdb/planner/expression_binder.hpp
```diff
@@ -48,6 +48,18 @@ struct BoundColumnReferenceInfo {
 	QueryLocation query_location;
 };
 
+//! The outcome of resolving a name against the chain of enclosing query scopes
+struct ColumnResolution {
```
Tishj, line 52: Isn't this essentially a `BindResult` with an added `idx_t depth` ?

Tishj:
This is a very exciting PR :eyes:

src/include/duckdb/planner/expression_binder.hpp
Maxxen: Yeah, we should probably rework both BindResult and ColumnResolution into something like a std::variant eventually, and maybe make this more elegant in the future.

src/planner/binder/expression/bind_case_expression.cpp
Maxxen: Sure, I'll remember it for the inevitable followup PR (or if there is other feedback to this PR that requires more changes I can fix it then)

Maxxen:
> This is a very exciting PR 👀

Thanks for having a look! Indeed! there's a bunch of cool stuff and cleanup this unlocks so I'm very excited I managed to figure out something that seems to work without being _too_ invasive of a change.

Mytherin:
Awesome, looks great! Thanks

Tishj:
Who knew the functional programming people were right all along ;)

PR #25222, opened 2026-09-01
5812 fix qualified count star

### LINKED ISSUE #18879: Qualified count star over asymmetric join miscounts
### What happens?

A qualified count_star over an asymmetric join will include an erroneous count where there are no matching rows from the joined table.

### To Reproduce

```
create table a (id int4);
create table b (id int4);
insert into a (id) values (1), (2), (3);
insert into b (id) values (2), (3), (3);

select a.id
	, count(*)
	, count(a.*) as a_star
	, count(b.*) as b_star
	, count(b.id) as b_id
from a
left join b on a.id = b.id
group by a.id
;
```

The last two columns should be equal; the `b_star` column is erroneous.

### OS:

macos

### DuckDB Version:

DuckDB v1.4.0-dev3790 (Development Version) 879df8212b

### DuckDB Client:

duckdb cli

### Hardware:

_No response_

### Full Name:

Tim McEwan

### Affiliation:

Climate Risk

I have tested with a nightly build

Yes

- [x] Yes, I have

- [x] Yes, I have

Tmonster:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/5812
Fixes https://github.com/duckdb/duckdb/issues/18879

COUNT(b.\*) was being treated as plain COUNT(\*). When the parser saw a star as the only argument to a function it dropped the argument entirely, without checking if the star was qualified — so COUNT(b.\*) and COUNT(\*) parsed to exactly the same thing. That is wrong for an outer  join: in a LEFT JOIN b, rows of a with no match still produce a row, so COUNT(\*) counts them, but there is no row of b to count. In the example from  the issue, COUNT(b.\*) returned 1 instead of 0 for the unmatched row, disagreeing with COUNT(b.id) right next to it.

The parser now only drops the argument for an unqualified \*, so COUNT(b.\*) keeps its star and reaches the binder. The binder expands the star into the columns of b and replaces the argument with CASE WHEN b.id IS NOT NULL OR b.uniq IS NOT NULL THEN 1 END. That gives the argument a value of 1 for  any row that has at least one non-NULL column, and NULL for a row where every column is NULL — and since COUNT ignores NULL, only real rows of b are  counted. Rows invented by the outer join are all-NULL, so they drop out, which matches what PostgreSQL returns. The rewrite runs everywhere aggregates are allowed (select list, HAVING, QUALIFY, ORDER BY, window functions), and the original COUNT(b.\*) text is kept as the column name so the rewrite isn't visible in the result.

Edit: This is also a candidate to get merged back into the v2.0.0 branch


test/sql/aggregate/aggregates/test_count_qualified_star.test
```diff
@@ -0,0 +1,402 @@
...
+----
+1
+
+# struct columns are expanded as well
+query I
+SELECT count(st.*) FROM (SELECT {'x': 1, 'y': 2} AS st) t;
+----
+1
+
+query I
```
Dtenwolde, line 377: This is inconsistent with PostgreSQL, which returns 1 in this case. A row with all NULL still counts as a row with `count(*)`

Dtenwolde: Also on v1.5 this is the case:
```sql
SELECT count(st.*) FROM (SELECT {'x': NULL, 'y': NULL} AS st) t;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
```

src/planner/binder/expression/bind_star_expression.cpp
```diff
@@ -254,6 +257,61 @@ optional_ptr<ParsedExpression> Binder::GetResolvedColumnExpression(ParsedExpress
 	return expr;
 }
 
+static bool IsQualifiedStar(const ParsedExpression &expr) {
+	if (!StarExpression::IsStar(expr)) {
+		return false;
+	}
+	auto &star = expr.Cast<StarExpression>();
+	return !star.RelationName().empty() && star.ExcludeList().empty() && star.ReplaceList().empty() &&
```
Dtenwolde: Can we split these up and have some early returns? Reads a little nicer :)

src/planner/binder/expression/bind_star_expression.cpp
```diff
@@ -254,6 +257,61 @@ optional_ptr<ParsedExpression> Binder::GetResolvedColumnExpression(ParsedExpress
...
+	       star.RenameList().empty();
+}
+
+//! COUNT(tbl.*) counts the rows of tbl - a row of tbl is NULL if all of its columns are NULL
```
Dtenwolde: > a row of tbl is NULL if all of its columns are NULL 

I don't think this is correct
A row of tbl is not null if the row exists, even if all of its columns are null

Dtenwolde: Here's postgreSQL's "explanation": https://www.postgresql.org/message-id/1228519.1647638908@sss.pgh.pa.us

src/parser/peg/transformer/transform_expression.cpp
```diff
@@ -16,10 +16,17 @@
...
 #include "duckdb/parser/parsed_expression_iterator.hpp"
+#include "duckdb/parser/expression/star_expression.hpp"
 #include "duckdb/common/string_util.hpp"
 
 namespace duckdb {
 
+//! Only an unqualified star (e.g. COUNT(*)) refers to the rows themselves - a qualified star (e.g. COUNT(tbl.*))
+//! refers to the columns of that relation and is expanded by the binder
+static bool ExpressionIsUnqualifiedStar(const ParsedExpression &expr) {
+	return PEGTransformerFactory::ExpressionIsEmptyStar(expr) && expr.Cast<StarExpression>().RelationName().empty();
```
Dtenwolde, line 29: Not something this PR missed, but I just noticed, can we add the check whether `RenameList` is also empty inside the `PEGTransformerFactory::ExpressionIsEmptyStar()`? We do check it correctly in the binder, but not in the transformer

src/planner/binder/query_node/bind_select_node.cpp
```diff
@@ -449,6 +449,7 @@ void Binder::BindWhereStarExpression(unique_ptr<ParsedExpression> &expr) {
 }
 
 void Binder::NormalizeFilterStarExpression(ParsedExpression &expr) {
+	TransformQualifiedCountStar(expr);
```
Dtenwolde: (AI assisted review comment); I think it's better to handle this in the aggregate/window binding, `BaseSelectBinder::BindAggregate`, rather than in the filter case. This qualified-star handling should happen before the argument-binding loop
There's a case with using the `count(t.*)` that regresses: 
```
CREATE MACRO count_rows() AS count(t.*);
Binder Error:
STAR expression is not supported here

LINE 1: CREATE MACRO count_rows() AS count(t.*);
                                           ^^^
```

The same goes for window functions: 
```
 SELECT
               count(*) OVER () AS all_rows,
               count(t.*) OVER () AS qualified_rows
           FROM (VALUES (NULL::INT, NULL::INT)) AS t(x, y);
┌──────────┬────────────────┐
│ all_rows │ qualified_rows │
│  int64   │     int64      │
├──────────┼────────────────┤
│        1 │              0 │
└──────────┴────────────────┘
```
This should return 1 for both cases (PostgreSQL behavior)

Dtenwolde (commented):
Couple of comments

hannes:
Thanks!

PR #25226, opened 2026-09-01
Optimize self-comparison filter pruning

vvvdwbvvv:
Prune always-false comparisons of the same bound column:
  `>`, `<`, and `<>`.

before:
```sql
memory D ATTACH '/tmp/self_compare.duckdb' AS rg (ROW_GROUP_SIZE 2048);
memory D   USE rg;
rg D 
rg D   CREATE TABLE self_compare AS
       SELECT CASE WHEN range % 2 = 0 THEN NULL ELSE range END::BIGINT AS i
       FROM range(8192);
Catalog Error:
Table with name "self_compare" already exists!
rg D   CHECKPOINT;
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
rg D 
rg D   EXPLAIN ANALYZE
       SELECT count(*)
       FROM self_compare
       WHERE i > i;
╭─ Summary ───────────╮
│ Total Time: 0.0011s │
│ Data Read: 262.1 kB │
╰─────────────────────╯
│ Aggregates: count_star()    │
│ 1 row                   0µs │
╰──────────────┬──────────────╯
╭─ Table Scan ─┴──────────────╮
│ Table: rg.main.self_compare │
│ Type: Sequential Scan       │
│ Filters: i > i              │
│ Row Groups Scanned: 4 / 4   │
│ 0 rows                  0µs │
```

after

```sql
memory D   ATTACH '/tmp/self_compare.duckdb' AS rg (ROW_GROUP_SIZE 2048);
memory D   USE rg;
rg D 
rg D   CREATE TABLE self_compare AS
       SELECT CASE WHEN range % 2 = 0 THEN NULL ELSE range END::BIGINT AS i
       FROM range(8192);
Catalog Error:
Table with name "self_compare" already exists!
rg D   CHECKPOINT;
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
rg D 
rg D   EXPLAIN ANALYZE
       SELECT count(*)
       FROM self_compare
       WHERE i > i;
╭─ Summary ───────────╮
│ Total Time: 0.0026s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ────╮
│ Aggregates: count_star() │
│ 1 row                0µs │
╰─────────────┬────────────╯
╭─ Empty Result ───────────╮
│ 0 rows               0µs │
╰──────────────────────────╯
```

### DIFF
--- a/src/planner/filter/expression_filter.cpp
+++ b/src/planner/filter/expression_filter.cpp
@@ -443,6 +443,10 @@ static FilterPropagateResult CheckComparisonStatistics(optional_ptr<ClientContex
 		case ExpressionType::COMPARE_GREATERTHAN:
 		case ExpressionType::COMPARE_LESSTHAN:
 			return FilterPropagateResult::FILTER_FALSE_OR_NULL;
+		case ExpressionType::COMPARE_DISTINCT_FROM:
+			return FilterPropagateResult::FILTER_ALWAYS_FALSE;
+		case ExpressionType::COMPARE_NOT_DISTINCT_FROM:
+			return FilterPropagateResult::FILTER_ALWAYS_TRUE;
 		default:
 			break;
 		}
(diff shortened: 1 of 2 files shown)


dentiny (commented):
Thanks for the PR! I'm wondering if it's intentional to miss a few other operations? For example, self-reference fast path could also handle `COMPARE_DISTINCT_FROM` as `FILTER_ALWAYS_FALSE` and `COMPARE_NOT_DISTINCT_FROM` as `FILTER_ALWAYS_TRUE`? Unlike regular equality, both results are independent of nullability.

PR #25228, opened 2026-09-01
Restructure MERGE INTO to no longer have internal operators and to instead create pipelines with merge sources that do the various merge actions

Mytherin:
`MERGE INTO` currently creates the child operators (`INSERT`, `DELETE`, etc) internally and calls`Sink`/`Combine`/`Finalize` on them manually. This has a number of drawbacks:

* `MERGE INTO` does not slot into the existing pipeline executor machinery, and doesn't support e.g. blocking sinks / async I/O
* The internal operators are invisible from the users' point of view, i.e. in `EXPLAIN` / `EXPLAIN ANALYZE` there's only a big `MERGE INTO` operator

Extensions need to do a bunch of dancing around these restrictions - e.g. both DuckLake and Iceberg have special "MergeInto[Insert|Update|Delete]" operators instead of using the out-of-the-box operators.

This PR expands the actions into the query plan and integrates them into regular pipeline execution.

### Merge action pipelines

Every action that has an operator gets a `PhysicalMergeActionSource` that is planned as the child of the operators of that action. 

The merge into pushes the rows that belong to an action into the
`MergeActionQueue` of that action, and the source of the action pipeline scans them, so the actions are regular pipelines that run concurrently with the pipelines that feed the merge into. The source of an action blocks while its queue is empty, and the merge into blocks when a queue is full and the
pipeline of an action cannot keep up.

With `RETURNING` the action operators emit the modified rows themselves - every action gets a projection that adds the `merge_action` column and becomes the source of one of the result pipelines.

### Planning actions through the regular INSERT/UPDATE/DELETE plans

`Catalog::PlanMergeInto` now has a default implementation - it plans every action through the regular `PlanInsert`/`PlanUpdate`/`PlanDelete` of the catalog. The merge action source takes the place of the child plan that a regular INSERT/UPDATE/DELETE reads from, so a catalog does not need to know that it is planning a merge action at all: UPDATE projects the rows of the action into `[new values, row ids]`, DELETE references the row id columns of the merge input, and INSERT reads the projection of the action. The values are cast to the column types when planning, because the operators of a regular INSERT/UPDATE read the values of a row in the types of the table.

### DuckLake and Iceberg

`DuckLakeCatalog::PlanMergeInto`, `DuckLakeMergeInsert`, `DuckLakeMergeUpdate`,
`IcebergCatalog::PlanMergeInto`, `IcebergMergeInsert` and `IcebergMergeUpdate` are removed - both extensions use their regular insert, update and delete plans now.


.github/patches/extensions/ducklake/zzzzzzzzz-merge-into-action-pipelines.patch
```diff
@@ -0,0 +1,687 @@
+diff --git a/src/storage/ducklake_delete.cpp b/src/storage/ducklake_delete.cpp
```
Tishj, line 1: I've seen this before, but what's the reason for using `zzzzzzzz` as the prefix? This is just going to silently break the patch if earlier patches are added that conflict, rather than being able to build on top of this patch, no?

The only benefit I can think of is avoiding ordering problems caused by silent merge conflicts (when two PRs make a `0002-<name>.patch` for example)

Mytherin: It's the clanker's idea, I agree it's not very nice. We should tell it to use numbers in the `AGENTS.md`.

Mytherin:
@kryonix could you maybe have a look at the pipeline machinery here?

.github/patches/extensions/ducklake/zzzzzzzzz-merge-into-action-pipelines.patch
Tishj: Agreed, I'll give that a go. I've wanted to expand/distribute the AGENTS.md file before, this seems like a good first use case for it.

kryonix:
I think the _action pipelines_ are the right direction, and the implementation looks correct. But I am not sure if introducing `MergeActionQueue` as an independent second pipeline-exchange implementation is the right direction. It duplicates backpressure, wakeup, cancellation, materialization, spilling, and progress behavior already handled by `PipelineBroadcastExchange`, and these implementations can (and propably will) diverge. I think it should be possible to factor the common transport/lifecycle into a generic exchange and implement broadcast versus routed distribution as policies. MERGE needs routed rather than broadcast semantics, including a materialize-to-completion mode for actions whose consumers are dependency-blocked.

PR #25229, opened 2026-09-01
Add new `AGENTS.md` file for creating extension patches

Tishj:
This PR originates from this [comment](https://github.com/duckdb/duckdb/pull/25228/changes#r3906181120)

Route to it from the root `AGENTS.md` so its used exclusively when the situation calls for it, without polluting/diluting the context of every session.


Mytherin:
Thanks!

PR #25232, opened 2026-09-01
Fix parallel CSV reader duplicate rows with \r\r\n line endings

### LINKED ISSUE #25164: Parallel CSV reader emits duplicate rows on files with \r\r\n line endings
### What happens?

When a CSV file uses `\r\r\n` as its line terminator, the parallel CSV reader emits byte-identical duplicates of rows that appear exactly once in the source file. The duplicated rows are well-formed and valid — they are simply returned twice. Reading the same file with `parallel=false` returns the correct row count with no duplicates.

The number of duplicates scales with the number of buffer boundaries in the file: a smaller `buffer_size` (or a larger file) produces more duplicates. The result is deterministic for a given file and buffer size — it does not vary between runs or with the thread count — which suggests a boundary-handling error rather than a race.

Files with ordinary `\r\n` or `\n` terminators are unaffected.

This is silent data corruption: no error, no warning, and nothing in `reject_errors`. It was found only because row counts were being cross-checked against an independent count of the source file.

### To Reproduce

```python
import duckdb

with open("rr.csv", "w", newline="") as f:
    f.write("id,val\r\r\n")
    for i in range(10000):
        f.write(f"{i},x\r\r\n")

con = duckdb.connect()
con.execute("SET threads=8")

q = """SELECT COUNT(*) AS rows, COUNT(DISTINCT id) AS distinct_ids
       FROM read_csv('rr.csv', auto_detect=false, header=true, delim=',',
            columns={'id':'BIGINT','val':'VARCHAR'},
            strict_mode=false, ignore_errors=true,
            max_line_size=2048, buffer_size=4096, parallel=%s)"""

print("parallel=true :", con.execute(q % "true").fetchone())
print("parallel=false:", con.execute(q % "false").fetchone())
```

Output:

```
parallel=true : (10002, 10000)
parallel=false: (10000, 10000)
```

The file contains 10,000 data rows. The parallel reader returns 10,002 rows containing only 10,000 distinct ids. The duplicated ids in this case are 1943 and 6039.

### Scaling with buffer size

Same generator, `threads=8`, varying row count and `buffer_size`:

| rows    | buffer_size | rows returned | extra |
|---------|-------------|---------------|-------|
| 10,000  | 4,096       | 10,002        | +2    |
| 10,000  | 16,384      | 10,001        | +1    |
| 10,000  | 65,536      | 10,000        |  0    |
| 200,000 | 4,096       | 200,027       | +27   |
| 200,000 | 16,384      | 200,008       | +8    |
| 200,000 | 65,536      | 200,002       | +2    |

Identical results at `threads=4` and `threads=8`.

### Only `\r\r\n` is affected

Same 200,000-row generator, 31 columns, `buffer_size=131072`, `threads=8`:

| line ending | expected | returned | diff |
|-------------|----------|----------|------|
| `\r\r\n`    | 200,000  | 200,001  | +1   |
| `\r\n`      | 200,000  | 200,000  |  0   |
| `\n`        | 200,000  | 200,000  |  0   |

### Real-world case

Found on ~20 GB option-quote CSVs (~83M rows, 31 columns) written with `\r\r\n` endings, read at the **default** buffer size:

```
records in source (grep -c):  82,984,768
parallel=true              :  82,984,769 rows, 1 duplicate group, 41 rejects   (35s)
parallel=false             :  82,984,768 rows, 0 duplicate groups, 40 rejects  (267s)
```

Note that the reject count also differs by one between the two paths.

Across a batch of 15 such files, 9 were affected, with 1–2 duplicated rows each. Because the duplicates are valid, well-formed rows, they are invisible to any schema- or error-based validation.

A second, possibly related symptom on the same files: with `\r\r\n` endings the reader consumes `\r\r` as the terminator and leaves the `\n` attached to the front of the first field, so `symbol` arrives as `"\nSPXW"` rather than `"SPXW"` on nearly every row. That happens under both parallel and sequential reads, so it may be a separate issue.

### OS:

macOS 26.5.1, x86_64

### DuckDB Version:

1.5.2 (CLI), also reproduced on 1.4.4 and 1.5.5

### DuckDB Client:

CLI and Python

### Hardware:

2019 Mac Pro, 16-core Intel Xeon, 256 GB RAM

### Full Name:

Mark Dragovan

### Affiliation:

DRD

- [x] Yes, I have

- [x] Yes, I have

alwaysprince05:
Fixes #25164

Problem

When a CSV file uses \r\r\n line endings and is read in parallel mode, the parallel CSV reader emits duplicate rows. This happens when a buffer boundary splits a row terminator between its two \r characters:

The previous buffer's scanner processes the first \r and counts the row via AddRow().
ProcessOverBufferValue() then consumes the remaining \r\n from the new buffer, but after its CARRY_ON newline-consuming loop it falls through to the "second buffer" loop, which starts parsing the next row's data.
As a result, the next row is counted by both the current scanner (via ProcessExtraRow) and the next scanner, producing duplicate rows.
Fix

In ProcessOverBufferValue(), after the CARRY_ON newline-consuming while loop, return early when:

No cross-buffer value exists (over_buffer_string is empty)
Newline characters were consumed
The previous buffer's scanner already fully counted the row (last_position > actual_size)
The first consumed character is \r (distinguishing \r\r\n from \r\n)
This prevents the function from falling through to the "second buffer" loop and double-counting the next row.

Testing

The regression test (test/sql/copy/csv/parallel/test_rrrn_parallel.test) generates the CSV on the fly via COPY ... (FORMAT BLOB) — no committed data file. Its layout is engineered so the bug actually triggers: fixed 25-byte rows follow a 26-byte header, placing row 39's first \r on the last byte of the 1024-byte buffer (offset 1023), so the buffer boundary splits the terminator as \r | \r\n.

Verified on a relassert build:

Without the fix, the test fails: parallel read returns 61 rows and id 40 is duplicated (sequential returns the correct 60)
With the fix, parallel and sequential both return exactly 60 rows
All parallel CSV tests pass with 0 regressions


dentiny (commented):
The test file is too crazy -- I'm wondering cannot we generate the test CSV on the fly?

Image

alwaysprince05:
Thanks for the review @dentiny! I've addressed your feedback:

**Removed the 10K-row data file** and replaced it with on-the-fly CSV generation using DuckDB SQL:

```sql
COPY (
    SELECT ('id,val' || chr(13) || chr(13) || chr(10) ||
        string_agg(CAST(id AS VARCHAR) || ',x' || CAST(id AS VARCHAR) || chr(13) || chr(13) || chr(10), '')
    )::BLOB
    FROM generate_series(0, 251) t(id)
) TO '{TEST_DIR}/rrrn_test.csv' (FORMAT BLOB);
```

The test now generates a 252-row CSV with `\r\r\n` line endings at runtime — no pre-committed data file needed. Net change went from +10,063 lines to -10,015 lines.

All 16 parallel CSV tests pass with the new approach.

alwaysprince05:
Hi @lnkuiper, the `Linux Relassert Tests` failure is unrelated to this PR's changes.

The one failing test is `Interrupted QueryAppender flow: interrupt → clear → close finishes` — a C++ unit test about the QueryAppender API that **timed out after 300s** (retried 4 times). This is a pre-existing flaky test that times out under CI resource contention (12 parallel workers).

It has nothing to do with CSV scanning or `\r\r\n` line endings.

Could you re-run or re-queue this PR? All 5,650+ other tests pass including the new `test_rrrn_parallel.test`.

alwaysprince05:
@dentiny @pdet — regression test updated per review feedback. The earlier layout passed even without the fix, so the file is now generated with fixed-width rows that place a \r\r\n terminator exactly across the 1024-byte buffer boundary. Verified on a relassert build: the test fails on unfixed code (parallel returns 61 rows, id 40 duplicated) and passes with the fix (60/60). All parallel CSV tests green.

pdet:
Thanks for the PR!

Can we also add a test with a midbuffer line ending?

e.g.,

```sql
  statement ok
  COPY (
      SELECT (repeat('p', 22) || chr(13) || chr(13) || chr(10) ||
          string_agg(lpad(CAST(t.id AS VARCHAR), 2, '0') || ',' || repeat('x', 19) || chr(13) || chr(13) || chr(10), ''))::BLOB
      FROM generate_series(0, 89) t(id)
  ) TO '{TEST_DIR}/rrrn_mid.csv' (FORMAT BLOB);

  query II
  SELECT count(*), count(DISTINCT id) FROM read_csv('{TEST_DIR}/rrrn_mid.csv',
      auto_detect=false, header=true, delim=',', columns={'id':'BIGINT','val':'VARCHAR'},
      strict_mode=false, buffer_size=4096, max_line_size=2048, parallel=true)
  ----
  90    90
```

src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
```diff
@@ -1482,6 +1483,18 @@ void StringValueScanner::ProcessOverBufferValue() {
 				state_machine->Transition(states, buffer_handle_ptr[iterator.pos.buffer_pos]);
 				iterator.pos.buffer_pos++;
 			}
+			// If we consumed newline characters but didn't add a row, and the previous
+			// buffer's scanner already fully counted the row (last_position overshoots the
+			// buffer), return early to avoid double-counting the next row. This handles
+			// \\r\\r\\n line endings where the first \\r triggered AddRow in the previous
+			// buffer and the remaining \\r\\n is here. We check that the first consumed
+			// character is \\r (not \\n) to avoid false-positives for \\r\\n line endings
+			// where \\r was at the end of the previous buffer and \\n is here.
```
pdet: nit: lets avoid multi-line comments

alwaysprince05:
Hi @pdet — both review items addressed in the latest push:

1. Single-line comment (done)
2. Midbuffer line ending test added — 90 rows with 4096-byte buffer (`rrrn_mid.csv`) (done)

Also closed #25338 per @lnkuiper's guidance — no need to port v2.0 fixes to main.

Let me know if anything else is needed!

alwaysprince05:
Hi @pdet — the `Main / prepare` CI failure appears to be a pre-existing branch issue on `v2.0-cyanoptera`, unrelated to this PR's changes. All 20 subsequent checks are skipped because `prepare` gates them.

The CSV fix is minimal (+102/-0, 2 files) and both review items are addressed. Could you re-run CI or proceed with review when convenient?

Thanks!

lnkuiper:
The ci failure is a formatting issue in a file you touched, could you have a look at fixing it?

alwaysprince05:
Hi @lnkuiper — fixed the clang-format issue in string_value_scanner.cpp. CI is now green. Ready for merge!

Mytherin:
Thanks!

PR #25233, opened 2026-09-01
Support string type for multicolumn pushdown

dentiny:
Followup PR for https://github.com/duckdb/duckdb/pull/25087

Currently string type is not supported for multicolumn pushdown, for example,
```sql
memory D CREATE TABLE db.string_values AS
         SELECT CASE range // 2048 WHEN 0 THEN 'a' WHEN 1 THEN 'a' ELSE 'b' END AS a,
                CASE range // 2048 WHEN 0 THEN 'a' WHEN 1 THEN 'b' ELSE 'b' END AS b
         FROM range(6144);
memory D EXPLAIN ANALYZE SELECT count(*) FROM db.string_values WHERE a != b;
╭─ Summary ───────────╮
│ Total Time: 0.0013s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ────────╮
│ Aggregates: count_star()     │
│ 1 row                    0µs │
╰───────────────┰──────────────╯
╭─ Filter ──────┸──────────────╮
│ Expression: a != b           │
│ 2,048 rows               0µs │
╰───────────────┒┎─────────────╯
╭─ Table Scan ──┚┖─────────────╮
│ Table: db.main.string_values │
│ Type: Sequential Scan        │
│ Projections: a, b            │
│ Row Groups Scanned: 3 / 3    │
│ 6,144 rows               0µs │
╰──────────────────────────────╯
```
Different from the above mentioned PR, for string stats we need to make sure both exact and inexact stats are properly handled.

### DIFF
--- a/src/optimizer/filter_combiner.cpp
+++ b/src/optimizer/filter_combiner.cpp
@@ -346,7 +346,7 @@ static bool TypeSupportsConstantFilter(const LogicalType &type) {
 
 static bool TypeSupportsMultiColumnComparison(const LogicalType &type) {
 	auto physical = type.InternalType();
-	return TypeIsNumeric(physical) || physical == PhysicalType::BOOL;
+	return TypeIsNumeric(physical) || physical == PhysicalType::VARCHAR || physical == PhysicalType::BOOL;
 }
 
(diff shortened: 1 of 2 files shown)


dentiny:

PR #25234, opened 2026-09-01
Prune row group for `map_contains_value`

dentiny:
Hi team, `map` type already stores stats for both keys and values, but function `map_contains_value` doesn't seem to leverage to prune row groups. For example,
```sql
memory D ATTACH '/tmp/map_extract_row_group_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE maps AS
     SELECT MAP {'k': i} AS m
     FROM range(6144) tbl(i);

-- should only access one row group
db D EXPLAIN ANALYZE SELECT count(*) FROM maps WHERE map_contains_value(m, 3000);
╭─ Summary ───────────╮
│ Total Time: 0.0063s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ──────────────────╮
│ Aggregates: count_star()               │
│ 1 row                              0µs │
╰────────────────────┒┎──────────────────╯
╭─ Table Scan ───────┚┖──────────────────╮
│ Table: db.main.maps                    │
│ Type: Sequential Scan                  │
│ Filters: contains(map_values(m), 3000) │
│ Row Groups Scanned: 3 / 3              │
│ 1 row                              0µs │
╰────────────────────────────────────────╯
```


dentiny:

lnkuiper:
Thanks! One issue:
```sql
CREATE TABLE maps(m MAP(VARCHAR, INTEGER));
INSERT INTO maps VALUES (MAP {'k': NULL});
SELECT count(*)
FROM maps
WHERE list_position(map_values(m), NULL) = 1;
```
Returns 1 on the base branch, 0 on this PR.

dentiny:
Hey @lnkuiper , actually it looks like a bug fixed by Hannes, but merged to main instead of v2.0 https://github.com/duckdb/duckdb/pull/25365

lnkuiper:
Could you cherry-pick that onto this PR?

lnkuiper:
Thanks! Should be good to go now. Maybe we should add the additional test case though?

dentiny:
> Thanks! Should be good to go now. Maybe we should add the additional test case though?

Already did :)

PR #25235, opened 2026-09-02
Create multi-column filter for rowid column

dentiny:
Hi team, I found for multi-column queries involving rowid column, they don't seem to be pruned correctly; for example,
```sql
memory D CREATE TABLE rowid_column_prune AS
         SELECT CASE
                    WHEN range >= 122880 AND range < 245760 THEN 400000
                    ELSE -1
                END::BIGINT AS boundary
         FROM range(300000);
memory D EXPLAIN ANALYZE SELECT count(*) FROM rowid_column_prune WHERE rowid < boundary;
╭─ Summary ───────────╮
│ Total Time: 0.0013s │
╰─────────────────────╯
╭─ Ungrouped Aggregate ─────────────────╮
│ Aggregates: count_star()              │
│ 1 row                             0µs │
╰───────────────────┰───────────────────╯
╭─ Filter ──────────┸───────────────────╮
│ Expression: rowid < boundary          │
│ 122,880 rows                      0µs │
╰───────────────────┒┎──────────────────╯
╭─ Table Scan ──────┚┖──────────────────╮
│ Table: memory.main.rowid_column_prune │
│ Projections: boundary                 │
│ Row Groups Scanned: 3 / 3             │
│ 300,000 rows                      0µs │
```
In the above example, we should be able to only access one row group.

There're two issues I could tell
- When creating a multi-column filter pushdown, when rowid is contained, it's [skipped directly](https://github.com/duckdb/duckdb/blob/4a0d6fdecbf8363d15a2d8132338a346f9225957/src/optimizer/filter_combiner.cpp#L482-L483)
- Even if filter pushdown has been activated correctly, we're not yet creating stats [for rowid column](https://github.com/duckdb/duckdb/blob/4a0d6fdecbf8363d15a2d8132338a346f9225957/src/storage/table/row_group.cpp#L735-L738)

### DIFF
--- a/src/optimizer/filter_combiner.cpp
+++ b/src/optimizer/filter_combiner.cpp
@@ -478,8 +478,11 @@ static unique_ptr<ExpressionFilter> TryCreateMultiColumnExpressionFilter(Logical
 	vector<ProjectionIndex> column_indexes;
 	column_indexes.reserve(distinct_bindings.size());
 	for (const auto &binding : distinct_bindings) {
-		if (binding.table_index != get.table_index || binding.column_index >= get.GetColumnIds().size() ||
-		    get.GetColumnIds()[binding.column_index].IsVirtualColumn()) {
+		if (binding.table_index != get.table_index || binding.column_index >= get.GetColumnIds().size()) {
+			return nullptr;
+		}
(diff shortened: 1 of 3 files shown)


dentiny:

PR #25237, opened 2026-09-02
Improve filter propagation result for string-result functions

dentiny:
Hi team, I think currently string-related functions return over-conservative filter propagation result, which leads to unnecessary row group access; for example,
```sql
memory D ATTACH '/tmp/prefix_zonemap_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE prefix_nullable_mixed AS
     SELECT CASE
                WHEN range < 2048 THEN CASE WHEN range % 2 = 0 THEN 'abc' ELSE NULL END
                WHEN range < 4096 THEN 'zzz'
                ELSE 'abc'
            END::VARCHAR AS s
     FROM range(6144);

-- we should only access one row group
db D EXPLAIN ANALYZE SELECT count(*) FROM prefix_nullable_mixed WHERE NOT prefix(s, 'a');
╭─ Summary ───────────╮
│ Total Time: 0.0022s │
╰─────────────────────╯
│ Aggregates: count_star()             │
│ Table: db.main.prefix_nullable_mixed │
│ Filters: NOT prefix(s, 'a')          │
│ Row Groups Scanned: 2 / 3            │
│ 2,048 rows                       0µs │
```

### DIFF
--- a/src/function/scalar/string/contains.cpp
+++ b/src/function/scalar/string/contains.cpp
@@ -161,7 +161,7 @@ static FilterPropagateResult ContainsFilterPrune(const FunctionStatisticsPruneIn
 	if (FindStrInStr(string_t(min), string_t(needle)) == DConstants::INVALID_INDEX) {
 		return FilterPropagateResult::FILTER_ALWAYS_FALSE;
 	}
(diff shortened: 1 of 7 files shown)


dentiny:

Mytherin:
Thanks!

PR #25243, opened 2026-09-02
Fix join elimination dropping rows when UNNEST follows DISTINCT

### LINKED ISSUE #25213: Join elimination drops rows when UNNEST follows DISTINCT
### What happens?

`join_elimination` removes a `LEFT JOIN` that can still increase the row count when an `UNNEST` is between `SELECT DISTINCT` and the join. 
The query below returns 8 rows instead of 20. 
Disabling `join_elimination` returns the expected result. Neither table has constraints.

`LogicalUnnest::GetColumnBindings` preserves its child's bindings even though `UNNEST` can emit multiple rows per input row. `JoinElimination::OptimizeChildren` does not invalidate the distinct group at `LOGICAL_UNNEST`, so `TryEliminateJoin` assumes the right side can match at most once and removes the join.

The same issue reproduces with `GROUP BY`, struct and nested `UNNEST`, `unnest(list(v))`, `FULL JOIN`, `USING`, and CTEs.

### To Reproduce

```sql
CREATE TABLE dup_keys(k INTEGER);
INSERT INTO dup_keys VALUES (1),(1),(2),(3);

CREATE TABLE probe_keys(idb INTEGER);
INSERT INTO probe_keys VALUES (1),(1),(2),(3),(4),(NULL),(2),(3);

SELECT count(*) FROM probe_keys LEFT JOIN (SELECT k, unnest([1,2,3]) e FROM (SELECT DISTINCT k FROM dup_keys) s) t
ON t.k = probe_keys.idb;
```

```
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            8 │
└──────────────┘
```

The expected result is 20. 
Disabling the rule returns it:

```sql
SET disabled_optimizers = 'join_elimination';
SELECT count(*) FROM probe_keys LEFT JOIN
  (SELECT k, unnest([1,2,3]) e FROM (SELECT DISTINCT k FROM dup_keys) s) t
  ON t.k = probe_keys.idb;
```

```
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│           20 │
└──────────────┘
```

### OS:

macOS 26.6.2

### DuckDB Version:

v1.5.5

### DuckDB Client:

CLI

### Hardware:

Apple Silicon, arm64

### Full Name:

Florian Valeye

### Affiliation:

altertable.ai

- [x] Yes, I have

- [x] Yes, I have

asubbarao:
Fixes https://github.com/duckdb/duckdb/issues/25213

`LogicalUnnest` keeps the child's column bindings while multiplying rows. Join elimination still treated those bindings as DISTINCT-unique, so a LEFT JOIN that could add rows was deleted (8 rows instead of 20 on v1.5.5).

Skip recording distinct groups under `LOGICAL_UNNEST`. The flag is restored walking back up the pipe so a DISTINCT above UNNEST can still eliminate the join.


asubbarao:
Superseded by #25552 (bd30bfb4b4, "Do not carry a distinct group across UNNEST in join_elimination"), which fixes #25213 the same way with equivalent tests; this PR's test file passes on the current base without its code.

PR #25251, opened 2026-09-02
enable auto-loading of extensions via their log types/messages

Tmonster:
fixes https://github.com/duckdblabs/duckdb-internal/issues/10398

In previous versions of DuckDB, when you enabled logging on core extensions that were not loaded, you would get the following,
```
memory D call enable_logging(['Iceberg', 'HTTP']);
Invalid Input Error:
Unknown log type: 'Iceberg'
memory D load iceberg;
memory D call enable_logging(['Iceberg', 'HTTP']);
```

This PR will auto load the extension if you enable logging for a core extension.
Currently only enabled for Iceberg and DuckLake. 

`call enable_logging('HTTP')` works because `HTTP` is a core log type since logging happens in src/main/http/http_util.cpp


Mytherin:
Thanks!

PR #25253, opened 2026-09-02
Add a planner-time statement rewrite hook for parser extensions

ted-wq-x:
## Summary

This PR adds an optional planner-time statement rewrite hook for parser extensions.

The hook allows an extension-owned statement to be rewritten into a DuckDB
`SQLStatement` after parsing and before binding. The rewritten statement then
continues through the regular planning pipeline using the same `ClientContext`
and transaction.

## Motivation

Parser extensions can introduce statements that require context-dependent
translation before they can be planned. The existing parse callback does not
have access to `ClientContext`, while performing the translation through a
separate query path would bypass the normal planning lifecycle.

A planner-time rewrite hook provides the required context while keeping
planning, binding, transaction handling, and error propagation within the
existing DuckDB execution path.

## Design

- Adds an optional statement rewrite callback to parser extensions.
- Invokes the callback during planning, before binding.
- Passes the active `ClientContext` to the callback.
- Requires the callback to return exactly one top-level `SQLStatement`.
- Sends the rewritten statement through the regular planning pipeline.
- Leaves parser extensions without a rewrite callback unchanged.

## Scope

This PR intentionally supports rewriting one extension statement into one
top-level DuckDB statement. It does not introduce general multi-statement
execution or extension-specific behavior into DuckDB.


Tishj:
Could you verify that this isn't possible with #24919 ?
We generally aren't looking to expand `ParserExtension` behavior, at this point its legacy and we only want to maintain it

ted-wq-x:
Thanks — I checked #24919.

It covers grammar extension and context-free AST construction well. A grammar
transform can produce a regular DuckDB `SQLStatement`, so this PR would not be
needed for extensions whose lowering is independent of connection state.

The remaining case is context-dependent lowering. The grammar transform
currently receives `PEGTransformer &` and `ParseResult &`, but not the active
`ClientContext`. The existing `plan_function` does receive `ClientContext`, but
its result is restricted to a `TableFunction`, so it cannot lower the parsed
representation into a regular DuckDB statement that continues through the
normal binder and planner.

The capability needed here is therefore:

- parse extension syntax through `GrammarExtension`;
- retain an extension-owned parsed representation;
- lower it into exactly one regular `SQLStatement` with the active
  `ClientContext`, before binding.

I agree that this should not expand the legacy `ParserExtension` API. Would you
prefer this to be designed as an optional planner-time lowering callback
associated with `GrammarExtension`, or as a separate non-legacy extension API?

I am happy to rework this PR on top of #24919. @Tishj

PR #25254, opened 2026-09-02
Don't let terminal matchers have a separate frame

Dtenwolde:
A first optimization for the heap based matcher; don't let terminal matchers create new frames, but instead resolve them directly, reducing allocation overhead. 

This improves performance by about 25% compared to the previous PR for TPC-H and TPC-DS queries. The overhead for the heap based matcher compared to stack is 2.65x. More optimizations to come.


Mytherin:
Thanks!

PR #25256, opened 2026-09-02
Backport (#24855) Arena allocate per-group buffers in bit/mode aggregates

ThijsDreef:
A backport of (#24855) for 1.5, removed changes related to 2.0 when directly cherry-picked.

Changes made
- Change from UnaryAggregateDestructor -> UnaryAggregate was required.

Feel free to close if there is no wish to backport this to 1.5


Mytherin:
Thanks but we're only adding critical fixes to v1.5-X at this point

PR #25257, opened 2026-09-02
[async results] Result buffer redesign

evertlammerts:
Stack (only land in order):
- [x] #25287 Adds a drain mode to the benchmark runner to distinguish between materializing and dropping chunks, and two regression tests
- [x] #25288 Enables the two new regression tests
- [ ] This PR
- [ ] API work: merge pending, streaming and materialized results into a single `QueryResult` type that supports non-blocking `Poll()` and `TryFetch()`, setting a notifier callback, and a blocking `Fetch()` that participates in / drives execution. I think we should keep shims for the old result types while the CLI and Sqlogic runner still depends on them.

**Only land after #25288**

Related to #24853. First a a few to introduce async results.

This tries to rework how query results are collected as suggested by @Mytherin in #25145. It introduces a result sink as root operator that can decide between a streaming buffer and full materialization. The sink is parameterized based on:
- Retention: either fully materialized, or streaming / bounded. This is now still chosen when submitting a query.
- Ordering: unordered, ordered by batch index, or ordered by source order. This depends on the query.

It's just a regular sink that parks an `InterruptState` in `PhysicalResultSink::Sink()` when the first chunk is ready and the result's retention is still `ResultLifetime::UNDECIDED`. The result's retention state changes to:
- `ResultLifetime::DRAINING` when the user calls `StreamQueryResult::ExecuteTask()` or `QueryResult::Fetch()`
- `ResultLifetime::RETAINED` when the user calls `StreamQueryResult::Materialize()`. Note that if this is called _after_ retention is decided, the drain just uses the streaming buffer an stores the chunks in a new CDC.

The existing collectors are removed with the exception of `PhysicalBatchCollector`, which is still needed for the Arrow collector.

One more note: I renamed `streaming_buffer_size` to `max_streaming_buffer_size` and changed the default from 1mb to 10mb, which seems to help with throughput.

Expected next PRs:
- API work: merge pending, streaming and materialized results into a single `QueryResult` type that supports non-blocking `Poll()` and `TryFetch()`, setting a notifier callback, and a blocking `Fetch()` that participates in / drives execution. I think we should keep shims for the old result types while the CLI and Sqlogic runner still depends on them.
- Deprecate the "custom collector" idea with an output format API that uses a format agnostic (but managed) buffer underneath. Move the Arrow collector to this API as the first consumer.


src/main/buffered_data/simple_buffered_data.cpp
```diff
@@ -77,49 +128,58 @@ StreamExecutionResult SimpleBufferedData::ExecuteTaskInternal(StreamQueryResult
...
 	}
+	InvokeUnblocks(to_unblock);
 	return chunk;
 }
 
-void SimpleBufferedData::Append(const DataChunk &to_append) {
-	auto chunk = make_uniq<DataChunk>();
-	chunk->Initialize(Allocator::DefaultAllocator(), to_append.GetTypes());
-	to_append.Copy(*chunk, 0);
-	auto allocation_size = chunk->GetDataSize();
```
Tishj, line 118: I have a feeling you can kill `GetDataSize` now, and all the implementations that are connected to it

Actually..
looks like it's used in one other place:
```c++
void OperatorMetrics::GatherMetrics(ClientContext &context, double elapsed_time, optional_ptr<DataChunk> chunk) {
	time += elapsed_time;
	if (chunk) {
		elements_returned += chunk->size();
		intermediate_size_bytes += LossyNumericCast<idx_t>(chunk->GetDataSize());
	}
```

src/main/buffered_data/buffered_data.cpp
```diff
@@ -7,12 +12,95 @@ namespace duckdb {
...
+	}
+	default:
+		break;
+	}
+	return size;
+}
+
+idx_t BufferedData::FlatChunkDataSize(DataChunk &chunk) {
+	// The size the chunk's data occupies once buffered: the flat entries, the validity
+	// masks, and the string payloads. Measured without materializing a flat copy
```
Tishj: "Without materializing a flat copy" - `FlatVectorDataSize` calls `Flatten` for nested vectors :eyes:
At this point I don't know if it makes sense to kill `GetDataSize`, looks like you just want to improve its accuracy ? maybe give it an enum if the accuracy isn't always required

evertlammerts: This is not needed anymore, I removed `FlatChunkDataSize`. `GetDataSize` is good enough for accounting now.

src/main/buffered_data/simple_buffered_data.cpp
evertlammerts: We can't kill it, we still need it for our buffer accounting. It'd be even better if we could just use the allocator for that but we can't, because it holds the chunk even after it is popped from the buffer. I don't really see a way to change that...

src/main/stream_query_result.cpp
Mytherin, line 157: Could we grab the `ColumnDataCollection` directly from the result collector here instead of making another copy?

Mytherin (commented):
Thanks for the PR! Nice - I think this is the right direction. One comment

src/main/stream_query_result.cpp
evertlammerts: This is now `MaterializeByDraining`.

`Materialize()` (below) calls it only if the result is _not_ set to `ResultLifetime::RETAINED`, i.e. if it was already set to streaming by the consumer, on the first fetch. Otherwise the buffer is a CDC and this doesn't get called.

Mytherin (approved):
LGTM - merge conflicts just need to get fixed.

evertlammerts:
@Mytherin conflicts fixed

Mytherin:
Thanks!

PR #25258, opened 2026-09-02
Fix infinite loop in equi_width_bins

### LINKED ISSUE #24282: Bug8: Infinite Loop in equi_width_bins() via Zero Step Size

DinosL:
Supersedes: https://github.com/duckdb/duckdb/pull/24538
Fixes: https://github.com/duckdb/duckdb/issues/24282

Builds on the initial PR, adding the two guards requested in the review.


Mytherin:
Thanks!

PR #25260, opened 2026-09-02
Add structural SQL export for bound expressions

kryonix:
DuckDB can bind parsed SQL into expression trees, but `Expression::ToString()` is intended for display, not reconstructing equivalent SQL. I added `BoundExpressionSQLExporter` as a fallible conversion from a valid bound `Expression` to an owned `ParsedExpression`.

Column references use a caller-provided resolver keyed by `ColumnBinding`. Failures carry structured paths and never expose a partial AST. Coverage includes typed constants, columns, casts, comparisons, `BETWEEN`, conjunctions, `CASE`, selected operators, scalar calls and aggregates with their modifiers.

Function reconstruction uses the retained logical definition and concrete logical signature, independently of physical specialization. An aggregate executing through `sum_no_overflow` can therefore export as `sum(...)`; the receiving DuckDB may choose a different implementation.

Calls require complete catalog/schema/name qualification. An unqualified fallback is unsafe even for valid optimizer-generated expressions: `LIKE 'a%'` can become `prefix(...)` and rebind to an earlier-schema macro, changing true to false. Similarly, AVG's generated `count(...)` can resolve to a conflicting macro. These cases now return `UNSUPPORTED_FUNCTION`. Qualifying optimizer-introduced builtins is a follow-up PR.

The exporter trusts semantically consistent native definitions; it does not authenticate deliberately misleading C++ metadata. Generic binding performs no catalog lookup or extension autoloading. Ordinary binary round trips are not categorically excluded, and the existing binary format remains unchanged.

Bind callbacks and bind data alone do not prevent export: rebinding reconstructs them. Calls remain unsupported when required arguments, captured aliases or other SQL-visible information are unavailable. There are no expression provenance recipes or function-specific export hooks.

Whole-plan scoping, windows, subqueries and broader syntax coverage remain separate follow-ups.


src/function/function_binder.cpp
```diff
@@ -650,8 +638,43 @@ unique_ptr<Expression> FunctionBinder::BindScalarFunction(const ScalarFunctionCa
...
+			    QualifiedName(selected_function->GetCatalogName(), selected_function->GetSchemaName(),
+			                  selected_function->GetName()),
+			    std::move(sql_export_arguments), result_func->GetReturnType(),
+			    selected_function->GetSQLExportCallback(),
+			    selected_function->HasBindCallback() || selected_function->GetCaptureArgumentAliases());
+		}
+		result = std::move(result_func);
+	}
+	return result;
 }
```
Maxxen: So if Im reading this right, this is basically saying: "if we bind the function, and it somehow changes, attach a "recipe" of the original function we resolved". Isn't this just the function definition + the arguments expressions? Why do we need this new "recipe" to identify the function?

src/function/function_binder.cpp
```diff
@@ -26,6 +25,13 @@ FunctionBinder::FunctionBinder(ClientContext &context_p) : binder(nullptr), cont
 FunctionBinder::FunctionBinder(Binder &binder_p) : binder(&binder_p), context(binder_p.context) {
 }
 
+template <class CATALOG_ENTRY>
+static bool IsTransactionVisibleCatalogEntry(ClientContext &context, const CATALOG_ENTRY &entry) {
+	auto transaction = entry.schema.GetCatalogTransaction(context);
+	auto visible_entry = entry.schema.GetEntry(transaction, entry.type, entry.name);
+	return visible_entry && RefersToSameObject(*visible_entry, entry);
```
Maxxen: When is a function not "transaction visible"?

src/include/duckdb/function/aggregate_function.hpp
```diff
@@ -326,6 +327,24 @@ class AggregateFunctionCallbacks {
...
+	aggregate_init_local_state_finalize_t init_local_state_finalize = nullptr;
+	aggregate_cluster_update_t cluster_update = nullptr;
+	aggregate_window_t window = nullptr;
+	aggregate_wininit_t window_init = nullptr;
+	aggregate_window_batch_t window_batch = nullptr;
+	aggregate_destructor_t destructor = nullptr;
+	aggregate_get_state_type_t get_state_type = nullptr;
+	aggregate_finalize_t export_aggregate_state = nullptr;
+	aggregate_import_state_t import_aggregate_state = nullptr;
+};
```
Maxxen: I don't understand why this distinction is necessary either, comparing function pointers for identify is not always safe anyway due to C++ extensions having potentially different addresses.

I think the only real function identify is the qualified name + the function signature

src/function/function_binder.cpp
kryonix: For normal SQL binding the catalog entry is transaction-visible. The distinction matters because these `FunctionBinder` overloads are public and accept a catalog-entry reference directly. That reference does not necessarily come from a lookup in this `ClientContext`. For example, callers can construct a `ScalarFunctionCatalogEntry` or `AggregateFunctionCatalogEntry` against a real catalog/schema without registering it. It can have the same qualified name and can even contain the actual live overload pointer, but it is still not the catalog entry visible through this transaction. An older entry version after replacement/drop, or an entry obtained under another transaction snapshot, has the same issue. The check therefore looks the entry up by type/name in the current transaction and requires object identity with the supplied entry. Without that check, passing a forged or stale entry through the public catalog-entry binder would be enough to attach an authenticated recipe.

src/function/function_binder.cpp
kryonix: There are two separate points here. This condition is not "the function changed, so remember the original." It verifies that the selected overload came from this exact transaction-visible catalog entry and that binding still identifies that overload as its canonical definition. Binding may specialize the current execution implementation, but that specialization is not what grants the recipe.

The definition plus argument expressions are not sufficient because they do not authenticate the expression occurrence. `BoundScalarFunction` and `BoundAggregateFunction` are copyable and retain their definition pointers. A caller can copy an authenticated bound function, change its name or callbacks, and construct a new public bound expression with the same children. For a name-sensitive function, that expression can e.g. execute as `8` while SQL reconstructed from the retained definition binds and executes as `7`. A live definition pointer can likewise be placed in a forged/non-visible catalog entry.

The recipe is the expression-owned provenance marker that distinguishes those cases. Copying the complete expression preserves it. Copying a function or using a public constructor does not. Broad mutation consumes it, while the narrow execution-only specialization APIs preserve it under the optimizer's semantic-equivalence contract. The argument ASTs are not duplicated in the recipe, only the authenticated logical name/signature is kept separately from the current execution implementation.

src/include/duckdb/function/aggregate_function.hpp
kryonix: I agree that function pointers are not a stable function identity. They are not intended to be used as identity here. The qualified name and signature identify the logical SQL function, while the callback pointers only select how that function executes inside the current DuckDB process.

This separate type is needed because `ReplaceImplementation` changes the whole function, including its name, return type, properties, bind callbacks, and serialization callbacks. E.g. for the `sum` optimization, we only want to replace the execution callbacks with those from `sum_no_overflow`, while the expression remains logically `sum`. The pointer comparisons are only used in the test to confirm that the local optimization happened.

kryonix: Still. I will check if I can "dumb down" this PR some more.

src/function/function_binder.cpp
Maxxen: This seems unlikely to actually occur to me - are there any places in the code where we create and bind and ad-hoc `ScalarFunctionCatalogEntry`? I know that we sometimes bind raw functions (e.g. by calling `StructExtractFun::GetFunction().Bind(...)` - but it would seem a lot easier to just eliminate these call sites and instead have them go through the catalog to get a catalog entry -> bind.

AFAIK functions are not really part of transactions anyway, except for macros - but they are not serialized as they are always inlined.

lnkuiper:
A few more comments:

1. **`alias()` can change results after export.**

   ```sql
   SELECT alias(i) AS renamed FROM (VALUES (1), (2)) t(i);
   ```

   the original returns `'renamed'`, while exporting and rebinding it returns `'i'`. `alias(i + 1)` also changes its result because the argument's expression text changes.

2. **Some constants export successfully but produce invalid SQL.** `'hello'::VARIANT` becomes `CAST(VARIANT(hello) AS VARIANT)`, which fails because there is no `variant` function. Likewise, exporting the value from `SELECT sum(i) EXPORT_STATE FROM (VALUES (1), (2)) t(i)` produces `CAST(3 AS AGGREGATE_STATE)`, which fails.

3. **Could we clarify the boundary around physical planning?** Calling `FunctionBinder::BindSortedAggregate` on a normally bound `first(i ORDER BY i DESC)` moves the ordering into bind data.This becomes `first(i)`, changing `2` to `1` over `(VALUES (1), (2)) t(i)`.

Maxxen:
I think `SetRequiresExpressionNames` is going to be problematic. We can't really rely on aliases having semantics meaning cause we produce a bunch of alias expressions that we _dont_ want to interpret as argument names. For example, any `ColumnRefExpression` gets aliased as the column name - which allows this to work, even thought it shouldn't (I'm trying to remove this for 2.0):

```
SELECT struct_pack(i) from range(0, 10) as r(i);
```

I also don't see why we need `LogicalArguments`/`LogicalReturnType`?

Maxxen:
Or you know what, lets just go for it for now - I can probably remove it once I get the `*args`/`**kwargs` work in since it will then be possible to distinguish named arguments properly.

PR #17267, opened 2025-04-27
Fix #17251: Eliminate duplicates after a set operation using set sema…

tianjq16:
Currently, the set operation using set semantics does not consider the impact of collation when eliminating duplicates. This can be fixed by adding a DistinctModifier in the transform process.

Fixes https://github.com/duckdb/duckdb/issues/17251

PR #25261, opened 2026-09-02
Avoid deduplicating UNION twice

ywelsch:
We noticed a regression in 1.5 (introduced by #17267) where a UNION is deduplicating the data twice.

In particular #17267 was putting a collation-aware DISTINCT modifier on top of every non-ALL set operation but left the operation itself with set semantics, so the physical planner still appended its own HASH_GROUP_BY.

### DIFF
--- a/src/planner/binder/query_node/bind_setop_node.cpp
+++ b/src/planner/binder/query_node/bind_setop_node.cpp
@@ -291,7 +291,10 @@ BoundStatement Binder::BindNode(SetOperationNode &statement) {
 
 	if (!statement.setop_all) {
 		statement.modifiers.insert(statement.modifiers.begin(), make_uniq<DistinctModifier>());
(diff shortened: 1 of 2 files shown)


src/planner/binder/query_node/bind_setop_node.cpp
```diff
@@ -291,7 +291,10 @@ BoundStatement Binder::BindNode(SetOperationNode &statement) {
 
 	if (!statement.setop_all) {
 		statement.modifiers.insert(statement.modifiers.begin(), make_uniq<DistinctModifier>());
-		statement.setop_all = false; // Already handled
```
ywelsch, line 294: Not sure what this was meant to do (the field was already `false` when reaching this place).

Mytherin:
Thanks!

PR #25262, opened 2026-09-02
Flip the switch to heap based parser

Dtenwolde:
This PR makes the switch from the stack parser to the heap based one. For now I do this by flipping the default setting for `heap_based_parser` to `true`. The stack based one remains available by using `SET heap_based_parser=false`. I plan on removing the stack based implementation in a follow up PR if that's okay.


Dtenwolde:
I think the regression failure for aoc is related

Dtenwolde:
Oh, rerunning the CI and updating main fixed the regression 🤔

kryonix:
That's strange. My suspicion was, that somewhere in the recursive part of the CTEs, parsing/transformation work was done repeatedly (sorting?), showing the bigger overhead as a regression.

Dtenwolde:
It seems there's still regression, but it's now within the limits I suppose: https://github.com/duckdb/duckdb/actions/runs/33730156450/job/100573967440?pr=25262#step:9:4605

kryonix:
Uh yeah, that's barely within the limit. But the other recursive CTE benchmark also shows regressions. Some are quite significant 🤨

```
flummi_ray                         1.4 s  [  1.4 s …  1.6 s ]    1.7 s  [  1.7 s …  2.0 s ]  +278.1 ms (+19.4%)    40
```

Dtenwolde:
> Some are quite significant 

How are those not flagged though? 20% is quite a lot

Dtenwolde:
I'll investigate :)

Edit: `flummi_ray` is a parser heavy workload, so this regression is expected with the heap-based parser. Doesn't seem related to recursive CTEs

PR #25263, opened 2026-09-02
Use CompressionSegmentReader for RLE

HendrikLambert:
This PR moves RLE decompression reads to `CompressionSegmentReader`.

## Changes
- Pin RLE scan and fetch blocks with the supplied query context.
- Bounds-check the count offset and the paired value and run-count regions.
- Validate run counts against the remaining segment rows as they are reached.
- Use the validated layout for RLE scans, skips, row fetches, selects and filters.
- Validate logical access ranges and handle zero-length runs without unbounded traversal.


src/storage/compression/rle.cpp
```diff
@@ -244,41 +245,136 @@ void RLEFinalizeCompress(CompressionState &state_p) {
...
+
+	static SegmentLayout ReadSegmentLayout(const BufferHandle &handle, ColumnSegment &segment, idx_t segment_count) {
+		auto reader = CompressionSegmentReader::FromSegment(handle, segment, "RLE segment");
+		auto rle_count_offset = reader.Read<uint64_t>();
+		if (rle_count_offset < RLEConstants::RLE_HEADER_SIZE || rle_count_offset > reader.Size()) {
+			ThrowRLECountOffsetInvalid();
+		}
+
+		auto value_capacity = (rle_count_offset - RLEConstants::RLE_HEADER_SIZE) / sizeof(T);
+		auto count_capacity = (reader.Size() - rle_count_offset) / sizeof(rle_count_t);
```
Tishj, line 335: This is not something we can do, `reader.Size()` could be the remainder of the entire block, not just the part that is occupied by our segment

src/storage/compression/rle.cpp
```diff
@@ -244,41 +245,136 @@ void RLEFinalizeCompress(CompressionState &state_p) {
...
-	          reinterpret_cast<const T *>(handle.Ptr() + segment.GetBlockOffset() + RLEConstants::RLE_HEADER_SIZE)),
-	      index_pointer(
-	          reinterpret_cast<const rle_count_t *>(handle.Ptr() + segment.GetBlockOffset() + rle_count_offset)),
-	      max_entry_pos(static_cast<idx_t>(reinterpret_cast<const_data_ptr_t>(handle.Ptr() + segment.GetBlockSize()) -
-	                                       reinterpret_cast<const_data_ptr_t>(index_pointer)) /
-	                    static_cast<idx_t>(sizeof(rle_count_t))) {
-		if (rle_count_offset < RLEConstants::RLE_HEADER_SIZE) {
-			//! This would make the index_pointer point into a region reserved for the header data
-			throw DataCorruptionException("Corrupted RLE segment: rle_count_offset is corrupted");
+	struct ValidatedRun {
```
Tishj, line 263: Right now you do a `GetRun` and use that result, I wonder if it's not safer to add these to a vector after validating (which we can reserve for upfront) and instead return an index into that vector?

src/storage/compression/rle.cpp
```diff
@@ -244,41 +245,136 @@ void RLEFinalizeCompress(CompressionState &state_p) {
...
+		auto run_counts = reader.template GetArray<rle_count_t>(rle_count_offset, count_capacity);
+		return SegmentLayout(values, run_counts, entry_capacity, segment_count);
+	}
+
+	explicit RLEScanState(BufferHandle handle_p, ColumnSegment &segment)
+	    : handle(std::move(handle_p)), segment_count(segment.count.load()),
+	      layout(ReadSegmentLayout(handle, segment, segment_count)), entry_pos(0), position_in_entry(0) {
+	}
+
+	rle_count_t GetCurrentRunCount() {
```
Tishj: My biggest concern is that these sound like non-mutating getters, but they're not.
Continuing on my earlier comment about using indexes, that would more cleanly allow us to separate the validation from the retrieval logic

src/storage/compression/rle.cpp
HendrikLambert: Yes, these are upper bounds and could leak into other segments in the same block, but I don't see how we could protect against this? [25176](https://github.com/duckdb/duckdb/pull/25176) will help for newer files, but older files lack this information altogether

Tishj: Correct, I only referenced it to provide context
I just think we have to be aware of this, and after that PR lands we can verify it for newly written files only

src/storage/compression/rle.cpp
HendrikLambert: In my latest commit, I made validation and retrieval more explicit. However, I still held off on moving every validated run into a vector, as that could double the required space. Please let me know if you think the extra space is worth the additional safety guard.

P.S. reserving the complete vector upfront doesn’t work all that well, as `entry_capacity` is only an upper bound and, as you noted earlier, is tied to `reader.Size()`, which may cover the remainder of the block.

Tishj (approved):
Thanks!

Tishj:
I don't think the size is a huge issue, since these are fixed-size items anyways, the extra overhead is negligible. We could put it on the stack or get it through the Allocator so its tracked.

But I'm also fine with not refactoring the existing code too much, maybe we can do a follow up at some point if we care enough :+1:

PR #25265, opened 2026-09-02
Add ProcessUtil: OS dependent informations about process id & co

carlopi:
Three questions about processes have no home in `common/`: who we are, whether a given process is still alive, and how to name one in an error message. The third is answered today by three static AdditionalProcessInfo copies in local_file_system.cpp; the first two are about to be answered a fourth time, in temporary_file_manager.cpp, which needs to tell whether the instance that left a temporary file behind is still running.

Collect them as ProcessUtil, with one #if per platform in the .cpp:
```
CurrentProcessId()                  getpid / GetCurrentProcessId
ProcessIsRunning(pid)               kill(pid, 0) / OpenProcess + GetExitCodeProcess
GetProcessDescription(fs, pid)      proc_pidinfo / /proc / QueryFullProcessImageNameW
```

This restructures and move code around, addressing comment at https://github.com/duckdb/duckdb/pull/24811#discussion_r3820088191

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -63,6 +63,7 @@ add_library_unity(
   hive_partitioning.cpp
   json_document.cpp
   pipe_file_system.cpp
+  process_util.cpp
   local_file_system.cpp
   memory_mapped_file.cpp
   error_data.cpp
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks!

PR #25277, opened 2026-09-02
planner: merge transaction-local statistics into table->GetStatistics

### LINKED ISSUE #25266: Wrong result: `IS NOT DISTINCT FROM` semi-join dropped NULL rows after an in-transaction INSERT
### What happens?

A correlated `EXISTS` subquery with a null-safe equality produces a SEMI join whose condition is `IS NOT DISTINCT FROM`. When one side of that join key is proven NOT NULL from table statistics, `SimplifyNullSafeSemiJoinConditions` (in `src/optimizer/pushdown/pushdown_mark_join.cpp`, commit `3550efc054` "Improve mark join pushdown null checks") downgrades the condition to a regular `=`. Regular equality drops the NULL==NULL match, so if that "NOT NULL" side in fact received NULL rows within the same transaction (appended by an earlier, still uncommitted statement), the probe-side NULL rows that should match are silently dropped. Result: one row short, no error.

### To Reproduce

Run the whole script in one DuckDB CLI session:

```sql
CREATE TABLE a(x INTEGER);
INSERT INTO a VALUES (1), (NULL);

-- b.y currently has no NULLs: its committed statistics claim NOT NULL
CREATE TABLE b(y INTEGER);
INSERT INTO b VALUES (1);

BEGIN;
-- same transaction, earlier statement: b becomes {1, NULL}, but the
-- optimizer's statistics snapshot (taken now) does not see this row
INSERT INTO b VALUES (NULL);

-- a.x = NULL should match b.y = NULL under IS NOT DISTINCT FROM
SELECT count(*) FROM a
WHERE EXISTS (SELECT 1 FROM b WHERE a.x IS NOT DISTINCT FROM b.y);
ROLLBACK;
```

Output:

```text
1
```

Expected output (null-safe semantics preserved):

```text
2
```

The `a` row with `x = NULL` is dropped: `b.y` is wrongly proven NOT NULL from committed-only statistics, so the SEMI join condition `a.x IS NOT DISTINCT FROM b.y` is rewritten to `a.x = b.y`, and `NULL = NULL` no longer matches the freshly appended `b.y = NULL` row.

The bug is deterministic. It is reproducible only when the appended NULL is still uncommitted (the committed statistics then do not reflect it). A guard that refuses the NOT-NULL proof while the current transaction holds uncommitted appends for the scanned table restores the correct result

### OS:

macOS

### DuckDB Version:

916aeeb64b0bd2e0f172266ac485daa929a4af98

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Jayson Huang

### Affiliation:

PingCAP

- [x] Yes, I have

- [x] Yes, I have

JaySon-Huang:
Fixes #25266

Found this when investigating the improvement of pr #24706. `NotNullExpressionAnalyzer` proves a column non-NULL from `table->GetStatistics()`, which only covers committed row groups. Rows appended by an earlier statement of the same transaction live in LocalStorage: they are visible to scans but invisible to the statistics. `SimplifyNullSafeSemiJoinConditions` then rewrites an `IS NOT DISTINCT FROM` semi-join condition to a regular `=`, dropping the NULL==NULL match and losing the probe rows.

This PR fixes `DuckTableEntry::GetStatistics`. When the transaction holds appends for the scanned table, the transaction-local column statistics are merged into the committed statistics, so the returned statistics describe the transaction-visible data. An uncommitted NULL invalidates a NOT-NULL proof, while an uncommitted append without NULLs preserves it.

Updates and deletes of committed rows do not need special handling: their write paths update the affected column statistics, and deleting rows can only make the statistics conservatively report NULLs that are no longer present.

The same blind spot affected `ConstantOrNullSimplification` and the `Deliminator`, which share the analyzer, and is closed here as well.

### DIFF
--- a/src/catalog/catalog_entry/duck_table_entry.cpp
+++ b/src/catalog/catalog_entry/duck_table_entry.cpp
@@ -30,6 +30,7 @@
 #include "duckdb/storage/table_storage_info.hpp"
 #include "duckdb/storage/table/data_table_info.hpp"
 #include "duckdb/storage/data_table.hpp"
+#include "duckdb/transaction/local_storage.hpp"
 #include "duckdb/main/attached_database.hpp"
 
 namespace duckdb {
@@ -246,7 +247,23 @@ const ColumnList &DuckTableEntry::GetColumns() const {
 }
 
 unique_ptr<BaseStatistics> DuckTableEntry::GetStatistics(ClientContext &context, const StorageIndex &column_id) {
-	return storage->GetStatistics(context, column_id);
+	// Get the committed statistics
+	auto stats = storage->GetStatistics(context, column_id);
+	if (!stats) {
+		return nullptr;
+	}
+	// Merge the transaction-local statistics in so the result describes the transaction-visible data.
+	auto &local_storage = LocalStorage::Get(context, ParentCatalog());
+	auto local_table_storage = local_storage.GetStorage(*storage);
+	if (!local_table_storage) {
+		return stats;
+	}
+	auto local_stats = local_table_storage->GetCollection().CopyStats(column_id);
+	if (!local_stats) {
+		return stats;
+	}
+	stats->Merge(*local_stats);
+	return stats;
 }
 
 unique_ptr<BaseStatistics> DuckTableEntry::GetStatistics(ClientContext &context, column_t column_id) {
@@ -259,7 +276,7 @@ unique_ptr<BaseStatistics> DuckTableEntry::GetStatistics(ClientContext &context,
 		return nullptr;
 	}
 	auto storage_index = GetStorageIndex(ColumnIndex(column_id));
-	return storage->GetStatistics(context, storage_index);
+	return GetStatistics(context, storage_index);
 }
 
 unique_ptr<BlockingSample> DuckTableEntry::GetSample() {
--- a/src/function/table/table_scan.cpp
+++ b/src/function/table/table_scan.cpp
@@ -878,18 +878,12 @@ unique_ptr<GlobalTableFunctionState> TableScanInitGlobal(ClientContext &context,
 
 static unique_ptr<BaseStatistics> TableScanStatistics(ClientContext &context, TableFunctionGetStatisticsInput &input) {
 	auto &column_id = input.column_index;
-	auto &bind_data = input.bind_data->Cast<TableScanBindData>();
-	auto &duck_table = bind_data.table.Cast<DuckTableEntry>();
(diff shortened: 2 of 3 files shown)


Mytherin:
Thanks, but this is the wrong approach. If this issue is happening it should be resolved in `table->GetStatistics` directly.

JaySon-Huang:
@Mytherin  Thanks for the review! The fix has been reworked as you suggested. `DuckTableEntry::GetStatistics` now pulls the transaction-local statistics from `LocalStorage` and merges them into the committed statistics, so the statistics it returns describe the transaction-visible data.

While reworking this I noticed that `TableScanStatistics` in https://github.com/duckdb/duckdb/blob/9c21294d984dfe9a1da5d055b5fce3e8a0d634b2/src/function/table/table_scan.cpp#L905-L908 still returns no statistics when the table has uncommitted data. Now that `DuckTableEntry::GetStatistics` returns merged statistics, would you prefer to remove that guard in this PR and trust the merged statistics, or keep it unchanged for now?

For context: the scan's cardinality estimate already includes transaction-local rows (`LocalStorage::AddedRows` in `TableScanCardinality`), so trusting the merged statistics would make the column statistics consistent with the row count.

Mytherin:
Thanks - maybe the easiest solution is to move that check to `DuckTableEntry::GetStatistics` so we just don't emit any stats for tables that have transaction-local changes?

If we want to emit stats in this case then I think your approach is fine - just please have a look at the CI failures.

JaySon-Huang:
I checked the ordering between transaction commit and the statistics updates, and merging the transaction-local statistics is both correct and strictly better than dropping them. `LocalStorage` is per-transaction, so we only ever merge our own uncommitted writes; and the committed stats are merged at commit time under the same lock readers use, before the transaction is published to new snapshots. So for any snapshot, the stats can only over-approximate the nullability of the rows it sees. That means `has_null` is never spuriously false, which is exactly what keeps the `IS NOT DISTINCT FROM → =` rewrite safe.
I think dropping the statistics instead would be less ideal, since it would give up the stats for the rest of the transaction even when the local writes contain no NULLs.

JaySon-Huang:
Looking at the Thread Sanitizer run ([link](https://github.com/duckdb/duckdb/actions/runs/34194844474/job/101960907679?pr=25277)), I believe the failure is not caused by the changes in this PR. However, the log only includes the last 100 lines of the raw unittest stderr, so there isn't enough information to determine what triggered the race. It would be helpful for the Thread Sanitizer job to output the full raw unittest stderr.

Mytherin:
Thanks!

PR #25283, opened 2026-09-02
Harden temp file reads

carlopi:
Improve some reads from corrupted compressed `.tmp` or `.block` files, where in release we wouldn't perform any checks. Those checks are basically integer comparisons, and better be sure not be dealing with clearly invalid data.


Mytherin:
Thanks!

PR #25284, opened 2026-09-02
Fixes to allow compiling with LLVM 23 cxx library

carlopi:
This is basically future proofing against newer libcxx implementation


Mytherin:
Thanks!

PR #25286, opened 2026-09-02
Fix Substring Clamping Issue

### LINKED ISSUE #24766: `SUBSTR` ASCII fast path clamps a hugely negative start; Unicode path returns empty

maiadegraaf:
This PR fixes a bug where the ASCII path and the unicode path computed the negative-offset window differently. The ASCII path incorrectly clamped the offset to the string bounds before applying the length, losing track of how far out of range the window was. This PR fixes that and makes both paths share the same computation.
          
Fixes: https://github.com/duckdb/duckdb/issues/24766

### DIFF
--- a/src/function/scalar/string/substring.cpp
+++ b/src/function/scalar/string/substring.cpp
@@ -52,16 +52,16 @@ static string_t SubstringSlice(Vector &result, const char *input_data, int64_t o
 }
 
 // compute start and end characters from the given input size and offset/length
-static bool SubstringStartEnd(int64_t input_size, int64_t offset, int64_t length, int64_t &start, int64_t &end) {
+static bool SubstringStartEnd(int64_t offset, int64_t length, int64_t &start, int64_t &end) {
 	if (length == 0) {
 		return false;
 	}
 	if (offset > 0) {
 		// positive offset: scan from start
-		start = MinValue<int64_t>(input_size, offset - 1);
+		start = offset - 1;
 	} else if (offset < 0) {
 		// negative offset: scan from end (i.e. start = end + offset)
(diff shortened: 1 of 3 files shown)


maiadegraaf:
CI failure seems unrelated.

PR #25287, opened 2026-09-02
[async results] Add drain mode param to interpreted_benchmark's resultmode

evertlammerts:
Stack:
- [ ] This PR
- [ ] Enable the two new regression tests
- [ ] Buffer redesign
- [ ] API work: merge pending, streaming and materialized results into a single `QueryResult` type that supports non-blocking `Poll()` and `TryFetch()`, setting a notifier callback, and a blocking `Fetch()` that participates in / drives execution. I think we should keep shims for the old result types while the CLI and Sqlogic runner still depends on them.

This adds a drain mode (either `drop` or `materialize`) to the interpreted benchmark runner, and two benchmarks that are just selects on a large dataset. They're not added to micro.csv yet, because I think that'll crash the Regression job since the baseline runner won't support three params yet. I'll add them in a follow up to this.

This is in preparation for the async result mode.

### DIFF
--- a/benchmark/group_descriptions.list
+++ b/benchmark/group_descriptions.list
@@ -30,6 +30,10 @@ The join micro benchmark set contains benchmarks that look at the speed of vario
 [Order]
 The order micro benchmark set contains benchmarks that look at the speed of sorting data using the ORDER BY clause.
 
+[result_collection]
+[Result Collection]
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks!

PR #25292, opened 2026-09-02
Fix column drop after index drop

### LINKED ISSUE #24086: Cannot drop column inside transaction if index is dropped inside same transaction
### What happens?

```bash
cat <<EOF | duckdb
create table "table1" ("id" integer not null, "name" varchar not null);

begin;
create table "table2" ("id" integer not null, "name" integer not null, "name2" varchar);
insert into "table2" ("id", "name", "name2") select "id", "name", '' as "name2" from "table1";
drop table "table1";
alter table "table2" rename to "table1";
create index "films_id_idx" on "table1" ("name2");
drop index "films_id_idx";
alter table "table1" drop column "name2";
commit;
EOF

// Catalog Error:
// Cannot drop this column: an index depends on it!

```

same without transaction:

```bash
cat <<EOF | duckdb
create table "table1" ("id" integer not null, "name" varchar not null);

create table "table2" ("id" integer not null, "name" integer not null, "name2" varchar);
insert into "table2" ("id", "name", "name2") select "id", "name", '' as "name2" from "table1";
drop table "table1";
alter table "table2" rename to "table1";
create index "films_id_idx" on "table1" ("name2");
drop index "films_id_idx";
alter table "table1" drop column "name2";
EOF

// no error

```

### To Reproduce

```bash
cat <<EOF | duckdb
create table "table1" ("id" integer not null, "name" varchar not null);

begin;
create table "table2" ("id" integer not null, "name" integer not null, "name2" varchar);
insert into "table2" ("id", "name", "name2") select "id", "name", '' as "name2" from "table1";
drop table "table1";
alter table "table2" rename to "table1";
create index "films_id_idx" on "table1" ("name2");
drop index "films_id_idx";
alter table "table1" drop column "name2";
commit;
EOF


```
### OS:

Linux

### DuckDB Version:

1.5.5

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Thomas Bley

### Affiliation:

-

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/24086

The root cause is straightforward: when we drop a column, we only check "whether the column is depended by an index", and we didn't validate "whether the index has been dropped already (in the same transaction)".


lnkuiper:
Hey, I don't have much experience with this part of the code base, but here's a Codex-assisted review:
> Thanks for the PR! Two blocking issues:
>
> 1. Skipping the dropped index during validation leaves the physical index attached to the rewritten table. A subsequent insert in the same transaction still uses its old column ID and aborts, because local storage is flushed before CommitDropState removes the index. The rewritten storage needs an index set
>    consistent with the transaction-visible schema.
>
> CREATE TABLE t(a INTEGER, b INTEGER, c INTEGER);
> CREATE INDEX idx ON t(c);
> BEGIN;
> DROP INDEX idx;
> ALTER TABLE t DROP COLUMN c;
> INSERT INTO t VALUES (1, 2);
> COMMIT;
>
> This aborts with Attempted to access index 2 within vector of size 2. The base safely rejects the ALTER, so the crash is exposed by this change.
>
> 2. Looking up the physical index by name is ambiguous when that name is reused. Here the replacement index is visible, so the dropped index on t1 is no longer recognized and the valid column drop is rejected:
>
> CREATE TABLE t1(a INTEGER, b INTEGER);
> CREATE TABLE t2(x INTEGER);
> CREATE INDEX idx ON t1(b);
> BEGIN;
> DROP INDEX idx;
> CREATE INDEX idx ON t2(x);
> ALTER TABLE t1 DROP COLUMN b;
>
> I think this should track/filter the exact index identity rather than exposing and interpreting the catalog version-chain child through EntryLookup::deleted_entry.

PR #25294, opened 2026-09-02
Generalize multi-column expression pruning

Andrewtangtang:
## Summary

This PR does two things:

- Generalizes multi-column predicate pushdown by replacing specialized checks with a unified capability checker based on expression statistics.
- Makes the checker recursive over supported AND/OR expression trees so nested predicates can also use row-group pruning.


Andrewtangtang:
Hi @kryonix ,could you help look at this PR when you have time? Thanks!

test/sql/optimizer/multi_column_expression_stats_pruning.test
```diff
@@ -0,0 +1,140 @@
+# name: test/sql/optimizer/multi_column_expression_stats_pruning.test
```
dentiny: I'm wondering if we should unify the test file with existing test/sql/optimizer/cross_column_or_stats_pruning.test?

Andrewtangtang:
Hi @lnkuiper , could you help review this PR when you have time. Thanks!

kryonix (approved):
Looks good to me. Replacing the specialized checks with a capability check makes this easier to extend, and recursive AND/OR support fits naturally. Keeping the residual filter and declining pruning when statistics are unavailable preserves the right safety boundaries.

PR #25303, opened 2026-09-03
Fix adding `UNIQUE` constraint on `ALTER TABLE`

### LINKED ISSUE #25304: Alter table to add UNIQUE constraint doesn't guarantee uniqueness
### What happens?

Hi team, `ALTER TABLE` already supports adding uniqueness constraint, but somehow the uniqueness is not enforced.

### To Reproduce

```sql
memory D CREATE TABLE t(c INT);
memory D INSERT INTO t VALUES (1);
memory D ALTER TABLE t ADD UNIQUE (c);
memory D INSERT INTO t VALUES (1);
memory D SELECT * FROM t;
┌───────┐
│   c   │
│ int32 │
├───────┤
│     1 │
│     1 │
└───────┘
```

### OS:

macos

### DuckDB Version:

6d7abb1090

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25304

Primary key and unique constraint are implemented roughly the same way, the only difference being: PK doesn't allow NULL entries and it builds foreign key index extraly.

The root cause is, when alter table, we [only create index on primary key addition](https://github.com/duckdb/duckdb/blob/ee95013f175bea7aa813f56c86fea983db6bc8e5/src/planner/binder/statement/bind_alter.cpp#L163-L168).

This PR does mostly two things:
- Create index on unique constraint creation when we alter a table
- Alongside with the index creation on alter, (1) serialize index to the WAL; (2) replay index from WAL; (3) remove index on rollback.


Mytherin:
Thanks!

dentiny:
Hi @Mytherin sorry this one needs your help as well -- been kicked out again 😅

PR #25307, opened 2026-09-03
Prevent duplicate streaming window final output

### LINKED ISSUE #25306: [Wrong result] Streaming LEAD followed by UNNEST duplicates the final row
### What happens?

A query that applies `UNNEST` downstream of a streaming `LEAD` window returns one extra row and an incorrect aggregate result.

The input to `LEAD` contains one row. The default value makes `x` equal to `7`, and `range(200000)` should expand that row to exactly 200,000 rows. DuckDB instead returns 200,001 rows with a sum of 1,400,007.

This reproduces with default settings on both the current stable release and the latest `main` commit. It also reproduces consistently with one thread. No external data or extensions are involved.

### To Reproduce

Run the following query in DuckDB:

```sql
SELECT count(*), sum(x)
FROM (
    SELECT unnest(range(200000)) AS u, x
    FROM (
        SELECT lead(i, 1, 7) OVER () AS x
        FROM range(1) t(i)
    )
    LIMIT 200001
);
```

Actual output:

```text
count_star(),sum(x)
200001,1400007
```

Expected output:

```text
count_star(),sum(x)
200000,1400000
```

The result was reproduced repeatedly on:

```text
DuckDB v1.5.5 (d8cdaa33fda8df955cc76ef58a280f68f4cd43fa)
DuckDB v2.1.0-dev83379 (780c7c743fec8d377f047422075dce4ecf4f16b7)
```

### OS:

Ubuntu 24.04.3 LTS, Linux 6.17.0-29-generic, x86_64

### DuckDB Version:

v1.5.5 and v2.1.0-dev83379 (`780c7c743fec8d377f047422075dce4ecf4f16b7`)

### DuckDB Client:

CLI and Python

### Hardware:

Intel Xeon E5-2686 v4，64G memory

### Full Name:

Kaka Chen

### Affiliation:

[AstroVela.ai](https://github.com/AstroVela)

- [x] Yes, I have

- [x] Yes, I have

kaka11chen:
Fixes #25306.

  ## Description

  A streaming window can emit its final delayed chunk more than once when a downstream operator returns `HAVE_MORE_OUTPUT` and pipeline execution resumes.

  `PhysicalStreamingWindow::FinalExecute` previously returned the delayed output without recording that the final chunk had already been flushed. A subsequent call could therefore emit the same rows again.

  This change reuses the existing `flushed_delayed` state to:

  - skip delayed output that has already been flushed;
  - mark the final delayed chunk as flushed after it is emitted.

  This keeps the change local to the streaming window operator and follows the existing finalization pattern where operators clear or advance their cached-output state after returning it.

  A regression test covers a streaming `LEAD` followed by `UNNEST` and `LIMIT`. The query previously returned 200,001 rows instead of 200,000.

  ## Tests

  - `GEN=ninja make reldebug`
  - `build/reldebug/test/unittest test/sql/window/test_streaming_lead_lag.test`
  - `build/reldebug/test/unittest 'test/sql/window/*'`
  - `build/reldebug/test/unittest 'Test TryFlushCachingOperators interrupted ExecutePushInternal'`
  - `build/reldebug/test/unittest 'Caching TableInOutFunction'`
  - `make format-fix`

### DIFF
--- a/src/execution/operator/aggregate/physical_streaming_window.cpp
+++ b/src/execution/operator/aggregate/physical_streaming_window.cpp
@@ -449,7 +449,7 @@ OperatorFinalizeResultType PhysicalStreamingWindow::FinalExecute(ExecutionContex
 	auto &gstate = gstate_p.Cast<StreamingWindowGlobalState>();
 	auto &state = gstate.local_state->Cast<StreamingWindowState>();
 
-	if (state.initialized && state.lead_count) {
+	if (state.initialized && state.lead_count && !state.flushed_delayed) {
 		auto &delayed = state.delayed;
 		//	There are no more input rows
 		auto &input = state.shifted;
@@ -462,6 +462,7 @@ OperatorFinalizeResultType PhysicalStreamingWindow::FinalExecute(ExecutionContex
 			return OperatorFinalizeResultType::HAVE_MORE_OUTPUT;
 		}
 		ExecuteDelayed(context, delayed, input, output, gstate_p);
+		state.flushed_delayed = true;
 	}
 
 	return OperatorFinalizeResultType::FINISHED;
--- a/test/sql/window/test_streaming_lead_lag.test
+++ b/test/sql/window/test_streaming_lead_lag.test
@@ -158,6 +158,17 @@ FROM range(10) tbl(i);
 8	9
 9	50
 
+# Test final output with a resuming downstream operator
+query II
+SELECT count(*), sum(x)
+FROM (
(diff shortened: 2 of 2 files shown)


carlopi:
For context, https://github.com/duckdb/duckdb/pull/25331 is also solving another angle of the same issue.

kaka11chen:
@carlopi Thanks. I think #25331 is the better central solution.

hawkfish:
Agreed - but thanks for the PR!

PR #25308, opened 2026-09-03
Save cache for workflow_dispatch events

smvv:
The nightly and lunch CI runs use `workflow_dispatch`. If used, save the cache.

This removes the need for detecting if the current branch is a (new) release branches or main.

I disabled saving the cache for merge queue branches, because github's cache [isolation](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache) prevents using those saved cache files in other (merge queue) CI runs.


Mytherin:
Thanks!

PR #25315, opened 2026-09-03
Add issue tracker workflow

DinosL:


.github/workflows/issue-forward.yml
```diff
@@ -0,0 +1,40 @@
...
+    types: [created]
+
+permissions:
+  contents: read
+
+jobs:
+  forward:
+    # issue_comment fires for PRs too so we skip those.
+    if: ${{ github.event_name == 'issues' || github.event.issue.pull_request == null }}
+    runs-on: ubuntu-latest
```
smvv:
```suggestion
    runs-on: ubuntu-slim
```

if this works with `ubuntu-slim` (`jq` and `gh` are available), i would use that.

smvv (approved):
I'm not too familiar with the issue tracker, but this looks good. Thanks!

.github/workflows/issue-forward.yml
DinosL: Thanks for the suggestion! Since `jq` and `gh` are available it should work! I changed it.

Mytherin:
Thanks!

PR #25322, opened 2026-09-03
Decouple string scan states

HendrikLambert:
Small PR that separates the scan states used by uncompressed strings, FSST and dictionary compression. Doing this now makes moving the individual algorithms over to the `CompressionSegmentReader` touch fewer files. It also ensures query context is used when pinning uncompressed string read buffers.

## Changes
- Make the FSST scan state own its buffer handle instead of inheriting it from `StringScanState`.
- Remove the dictionary scan state's redundant `StringScanState` base.
- Use the supplied query context when pinning uncompressed string scan, fetch and overflow blocks.
- Pass the query context through the dictionary entry helper shared with FSST so overflow block pins use it.


Tishj (approved):
Thanks!

PR #25325, opened 2026-09-03
Fix parallel CSV reader duplicate rows with \r\r\n line endings

### LINKED ISSUE #25164: Parallel CSV reader emits duplicate rows on files with \r\r\n line endings

alwaysprince05:
Fixes #25164. Port of #25232 to main.

Problem: When a CSV file uses \r\r\n line endings and is read in parallel mode, a buffer boundary can fall between the two \r characters of a row terminator. The previous buffer's scanner counts the row via AddRow(), and ProcessOverBufferValue() then consumes the remaining \r\n from the new buffer but falls through to the "second buffer" loop — so the next row is counted by both the current scanner and the next scanner, producing duplicate rows.

Fix: In ProcessOverBufferValue(), return early after the CARRY_ON newline-consuming while loop when no cross-buffer value exists (over_buffer_string is empty), newline characters were consumed, the previous buffer's scanner already fully counted the row (last_position > actual_size), and the first consumed character is \r.

Testing: Regression test test/sql/copy/csv/parallel/test_rrrn_parallel.test generates the CSV on the fly with fixed-width rows that place a terminator's first \r on the last byte of a 1024-byte buffer, forcing the \r | \r\n split. Verified on a relassert build against main: the test fails without the fix (parallel returns 61 rows, id 40 duplicated) and passes with it (60/60, matching sequential). Full parallel CSV suite passes (13 test cases, 13015 assertions).

### DIFF
--- a/src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
+++ b/src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
@@ -1454,6 +1454,7 @@ void StringValueScanner::ProcessOverBufferValue() {
 				iterator.pos.buffer_pos++;
 			}
 		} else {
+			idx_t pre_carry_pos = iterator.pos.buffer_pos;
 			while (iterator.pos.buffer_pos < cur_buffer_handle->actual_size &&
 			       (buffer_handle_ptr[iterator.pos.buffer_pos] == '\n' ||
 			        buffer_handle_ptr[iterator.pos.buffer_pos] == '\r')) {
@@ -1481,6 +1482,18 @@ void StringValueScanner::ProcessOverBufferValue() {
 				}
 				state_machine->Transition(states, buffer_handle_ptr[iterator.pos.buffer_pos]);
 				iterator.pos.buffer_pos++;
+			}
+			// If we consumed newline characters but didn't add a row, and the previous
+			// buffer's scanner already fully counted the row (last_position overshoots the
+			// buffer), return early to avoid double-counting the next row. This handles
+			// \\r\\r\\n line endings where the first \\r triggered AddRow in the previous
+			// buffer and the remaining \\r\\n is here. We check that the first consumed
+			// character is \\r (not \\n) to avoid false-positives for \\r\\n line endings
+			// where \\r was at the end of the previous buffer and \\n is here.
+			if (over_buffer_string.empty() && iterator.pos.buffer_pos > pre_carry_pos &&
+			    result.last_position.buffer_pos > previous_buffer_handle->actual_size &&
+			    buffer_handle_ptr[pre_carry_pos] == '\r') {
+				return;
 			}
 		}
 	}
--- a/test/sql/copy/csv/parallel/test_rrrn_parallel.test
+++ b/test/sql/copy/csv/parallel/test_rrrn_parallel.test
@@ -1,0 +1,81 @@
+# name: test/sql/copy/csv/parallel/test_rrrn_parallel.test
+# description: Test parallel CSV reader handles \r\r\n line endings without duplicate rows (issue #25164)
+# group: [parallel]
+
+require notwindows
+
+# Generate a CSV file with \r\r\n line endings whose layout forces a buffer
+# boundary to split a row terminator between the two \r characters:
+#   - header line: 23 'p' chars + "\r\r\n" terminator  -> 26 bytes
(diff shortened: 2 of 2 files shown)


alwaysprince05:
@pdet @dentiny — this is the same fix ported to main. Identical code and regression test as #25232. Same local verification: test fails without the fix (61 rows, id 40 duplicated) and passes with it (60/60). Linking #25232 for context.

Tishj:
@dentiny can you please push https://github.com/dentiny/duckdb/pull/181 to supersede this?

PR #25327, opened 2026-09-03
Remove Scans for column references in removed expressions

gropaul:
Hi dear DuckDB team,

While playing around with DDB, I found that when statistics propagation removes a redundant filter, the scan keeps reading the column that the filter referenced. `RemoveUnusedColumns` has already run by then, so nothing prunes it. 

The query I spotted it on is 
```sql
SELECT c.c_name, o.o_totalprice FROM customer c LEFT JOIN orders o ON c.c_custkey = o.o_custkey
WHERE o.o_comment IS NOT NULL AND o.o_totalprice > 100000;   
```

Filter pushdown turns the join into an inner join and pushes `o_comment IS NOT NULL` into the scan; statistics propagation proves it is always true and removes it, but the scan kept decompressing `o_comment`. 

The same happens for join conditions and COALESCE arguments that statistics propagation removes. The pruning pass already re-ran after statistics propagation for the `COUNT(x)` to `COUNT(*)` rewrite; this PR generalizes that trigger to every removal. With this, the query drops from 25 ms to 9 ms.

It is a bit hard to spot, as EXPLAIN does not show the columns that are still there, as the column is projected out of the scan right away. So for the test, I used `json_serialize_plan` which requires JSON; let me know if this is fine to do. 

Happy to integrate any wished changes!


Mytherin:
Thanks!

gropaul:
Hi @Mytherin, there was a problem in the merge queue with some adjacent fix; it is resolved now.

PR #25329, opened 2026-09-03
[Transaction] Honor the `current_transaction_invalidation_policy` in the `ErrorResult` constructor function

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/10649

`ErrorResult` is used to construct a `QueryResult` of type `T` that holds an error, but no invalidation was performed based on the error, silently ignoring the `current_transaction_invalidation_policy`


Mytherin:
Thanks!

PR #25330, opened 2026-09-03
add debug and deprecated settings

xproblematique:
Closes https://github.com/duckdblabs/duckdb-internal/issues/10644

Clean up the `duckdb_settings()` output by introducing separate handling for debug and deprecated settings, while also cleaning up obsolete settings and renaming newer debug-only settings. Deprecated settings will show a warning when used.

* Added `debug` and `deprecated` metadata to the settings definition pipeline.

  * Both default to `false`.
  * The metadata is propagated through the generated settings structs and `ConfigurationOption`.
* Extended `duckdb_settings()` with:

  * `debug := false` — controls whether debug settings are included.
  * `deprecated := false` — controls whether deprecated settings are included.
* Marked the following settings as deprecated:

  * `profiling_mode`
  * `produce_arrow_string_view`
  * `extension_directory`
  * `old_implicit_casting`
* To be decided:

  * `username`
  * `password`
  * `user`
* Renamed debug-oriented settings to use a `debug_` prefix, while retaining their previous names as aliases:

  * `force_update_to_del_and_insert` → `debug_force_update_to_del_and_insert`
  * `force_variant_shredding` → `debug_force_variant_shredding`
  * `enable_caching_operators` → `debug_enable_caching_operators`
  * `force_mbedtls_unsafe` → `debug_force_mbedtls_unsafe`
  * `force_bitpacking_mode` → `debug_force_bitpacking_mode`
* Updated the generated settings/configuration code to carry the new debug/deprecated metadata.
* Updated the relevant setting descriptions for deprecated settings.

## Compatibility

Renamed settings continue to work through aliases, so existing users of the old names are not immediately broken.

Deprecation warnings for aliases/deprecated settings are **not** included in this PR. There is currently no generic warning mechanism that `warnings_as_errors` can hook into, so adding that plumbing is left for a follow-up.

## Out of scope

The following settings require separate product decisions and are intentionally not changed here:

* `show_behavior`
* `enable_http_metadata_cache`
* `enable_fsst_vectors`
* `__delta_only_variant_encoding_enabled`

`lambda_syntax` and `debug_heap_based_parser` are also left untouched because their removal is tied to larger changes.

## Testing

* Verified that the settings generation pipeline propagates `IsDebug` and `IsDeprecated` correctly into generated setting structs.
* Added/updated coverage for the `duckdb_settings()` filtering and setting aliases.
* Existing settings generation and configuration tests continue to pass.


taniabogatsch (approved):
Looks good, thanks!

> Deprecation warnings for aliases/deprecated settings are not included in this PR. There is currently no generic warning mechanism that warnings_as_errors can hook into, so adding that plumbing is left for a follow-up.

That's fine by me, we can spin out an issue for this once this PR has landed. 👍 I'm surprised though that there was no straightforward way to add `DUCKDB_LOG_WARNING(context [or database instance], msg);` when setting/using a deprecated/renamed setting?

carlopi:
Are we sure about removing those 3? I got the impression they are there to help postgres compatibility
```sql
username
password
user
```

taniabogatsch:
> Are we sure about removing those 3? I got the impression they are there to help postgres compatibility

@JelteF, maybe you have a take here?

taniabogatsch:
@carlopi, would you prefer we deprecate them instead of removing?

Mytherin:
Thanks!

> Are we sure about removing those 3? I got the impression they are there to help postgres compatibility
> 
> ```sql
> username
> password
> user
> ```

I think we should keep these yes, we have had them since the beginning and they are common settings.

src/catalog/default/default_table_functions.cpp
```diff
@@ -76,7 +76,7 @@ WHERE type ILIKE log_type
 )"},
 	{DEFAULT_SCHEMA, "duckdb_profiling_settings", {}, {}, R"(
 SELECT * EXCLUDE(input_type, scope, aliases, typed_value)
-  FROM duckdb_settings()
+  FROM duckdb_settings(deprecated := true)
```
taniabogatsch, line 79: Should we switch to `false` here? So we don't keep showing deprecated settings to users?

PR #25331, opened 2026-09-03
Fixup TryFlushCachingOperators to avoid doubly emitted chunks

carlopi:
Fixes https://github.com/duckdb/duckdb/issues/25306

Proper CI test is "will everything else stay in place"?

### DIFF
--- a/test/sql/limit/streaming_limit_pipeline_flush2.test
+++ b/test/sql/limit/streaming_limit_pipeline_flush2.test
@@ -1,0 +1,19 @@
+# name: test/sql/limit/streaming_limit_pipeline_flush2.test
+# description: Test that pipeline correctly flushes operators, see https://github.com/duckdb/duckdb/issues/25306
+# group: [limit]
+
+statement ok
+SET threads = 1;
+
+query II
+SELECT count(*), sum(x)
+         FROM (
+             SELECT unnest(range(200000)) AS u, x
+             FROM (
+                 SELECT lead(i, 1, 7) OVER () AS x
+                 FROM range(1) t(i)
+             )
+             LIMIT 200001
+         );
+----
+200000	1400000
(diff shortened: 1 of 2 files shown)


carlopi:
I think this is the proper central solution, then the fact that PhysicalStreamingWindows behaves somewhat weirdly can also be fixed (see https://github.com/duckdb/duckdb/pull/25307/changes), but I think this is the more principled fix, were we avoid calling `FinalExecute` twice (or more).

test/sql/limit/streaming_limit_pipeline_flush2.test
```diff
@@ -0,0 +1,19 @@
...
+         FROM (
+             SELECT unnest(range(200000)) AS u, x
+             FROM (
+                 SELECT lead(i, 1, 7) OVER () AS x
+                 FROM range(1) t(i)
+             )
+             LIMIT 200001
+         );
+----
+200001	1400007
```
kaka11chen: This regression test appears to encode the buggy result rather than the expected behavior. `range(200000)` expands to 200,000 rows and `x` is `7` for every row, so the fixed result should be `200000\t1400000`. Also, because the query returns two INTEGER columns, the declaration should be `query II` rather than `query I`; otherwise sqllogictest reports a column-count mismatch. As written, the test will fail when the fix produces the correct result.

carlopi: Indeed.. sorry.

kaka11chen (approved):
LGTM

test/sql/limit/streaming_limit_pipeline_flush2.test
```diff
@@ -0,0 +1,19 @@
+# name: test/sql/limit/streaming_limit_pipeline_flush2.test
+# description: Test that pipeline correctly flushes operators, see https://github.com/duckdb/duckdb/issues/25306
+# group: [limit]
+
+statement ok
+SET threads = 1;
+
+query I
```
kaka11chen: query II ?

kaka11chen (approved):
LGTM

PR #25335, opened 2026-09-03
Properly sort missing‑parameter error message by parameter index

### LINKED ISSUE #25299: v2: wrong order of missing parameters in prepared statements
### What happens?

```
DuckDB v1.5.5 (Variegata)
memory D CREATE TABLE t (n INTEGER NULL, i INTEGER NULL, b BIGINT NULL, d DECIMAL(10, 2) NULL, v VARCHAR NULL);
memory D prepare foo as INSERT INTO t VALUES ($aa, $bb, $cc, $dd, $ee);
memory D execute foo(aa:=null, bb:=203);
Invalid Input Error:
Values were not provided for the following prepared statement parameters: cc, dd, ee
memory D 

memory D CREATE TABLE t (n INTEGER NULL, i INTEGER NULL, b BIGINT NULL, d DECIMAL(10, 2) NULL, v VARCHAR NULL);
memory D prepare foo as INSERT INTO t VALUES ($aa, $bb, $cc, $dd, $ee);
memory D execute foo(aa:=null, bb:=203);
Invalid Input Error:
Values were not provided for the following parameters: ee, cc, dd
memory D 
```

> Values were not provided for the following parameters: ee, cc, dd

should be cc, dd, ee

### To Reproduce

see above

### OS:

Linux

### DuckDB Version:

v2.0.0-alpha39998

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Thomas Bley

### Affiliation:

-

- [x] Yes, I have

- [x] Yes, I have

Yes

huoguocheng:
Use std::map to preserve and sort missing parameters by their original index, ensuring error messages list parameters in the position‑based order rather than unordered‑set iteration order. Also avoid duplicate entries for the same named parameter in error output.

Add sqllogictest cases for ordered missing‑parameter error output.

Partially fix #25299

### DIFF
--- a/src/include/duckdb/main/prepared_statement.hpp
+++ b/src/include/duckdb/main/prepared_statement.hpp
@@ -146,6 +146,7 @@ public:
 	                                     const identifier_map_t<PAYLOAD> &values, ClientContext *context = nullptr) {
 		// Missing values
 		identifier_set_t missing_set;
+		map<idx_t, Identifier> missing_idx_values;
 		for (auto &pair : parameters) {
 			auto &name = pair.first;
 			if (!values.count(name)) {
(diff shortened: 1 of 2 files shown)


asxvi:
I opened #25336 for the same issue (#25299) shortly before/after this. Looks like we both spotted it around the same time. Same root-cause fix; your test cases (declaration order, reverse order, duplicate-param) are more thorough than mine. 

No conflict from my side, happy to close mine in favor of this one if maintainers prefer, or defer to whichever the team wants to take forward.

Tishj (approved):
Thanks!

huoguocheng:
Thanks @asxvi! Great minds think alike. Let’s wait for maintainers to decide which one to take.

Mytherin:
Thanks - I merged https://github.com/duckdb/duckdb/pull/25336 because I saw that one first. I'll close this one in favor of that one.

PR #25336, opened 2026-09-03
Report missing prepared statement parameters in declaration order (#25299)

### LINKED ISSUE #25299: v2: wrong order of missing parameters in prepared statements

asxvi:
## Summary
- `PreparedStatement::MissingValuesException()` collected the names of missing prepared-statement parameters into an `identifier_set_t` (an `unordered_set`), so the order reported in the error message depended on hash-bucket layout rather than the order the parameters were declared in the statement.
- The parser already assigns each named parameter an incrementing index as it's parsed (`PEGTransformer::SetParam`), stored as the `idx_t` value in the `named_param_map`. That index was available but unused for ordering.
- Fixed by collecting `(index, name)` pairs while scanning for missing parameters and sorting by index before formatting the message, so the reported order now matches declaration order deterministically.

Fixes #25299

### DIFF
--- a/src/include/duckdb/main/prepared_statement.hpp
+++ b/src/include/duckdb/main/prepared_statement.hpp
@@ -8,7 +8,9 @@
 
 #pragma once
 
+#include "duckdb/common/algorithm.hpp"
 #include "duckdb/common/identifier.hpp"
+#include "duckdb/common/pair.hpp"
 #include "duckdb/common/winapi.hpp"
 #include "duckdb/main/materialized_query_result.hpp"
 #include "duckdb/main/pending_query_result.hpp"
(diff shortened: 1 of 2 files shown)


asxvi:
Noticed #25335 addresses the same issue (#25299) with essentially the same fix, opened around the same time. Looks like we hit this independently. Leaving both up in case reviewers prefer one approach's test coverage or want to combine them.

Mytherin:
Thanks!

PR #25340, opened 2026-09-03
C-API-V2: Part 3

Maxxen:
This PR is the third (and final?) part of the new C-API-V2 push that adds basically everything else from the prototype-fork.
- Part 1: https://github.com/duckdb/duckdb/pull/24702
- Part 2: https://github.com/duckdb/duckdb/pull/25184

Again, nothing is set in stone - there will for sure be more follow-up PRs, polishes, and features added that are currently blocked on core-work - but they will most likely not be as massive as the 3 PRs so far have been.

I'll add more detailed descriptions tomorrow, but below you'll find a short "table of content"/summary of the interfaces added/changed here. Note that all of these are of course also exposed in the "stable c++ API", `duckdb_cpp.hpp`

### Table Functions

 User-defined table functions: bind / init global / init local / exec / progress phases, named parameters with defaults via function-signature, bind-time cardinality, projection pushdown, and expression claim-based **filter pushdown**. Filter pushdown has been a long requested feature, and we will continue to refine this API to make it more powerful as we hopefully rework expressions more soon.
 
### Copy Functions

 User-defined `COPY TO` and `COPY FROM` formats. TO side: bind, batch size, init, batch (input ownership transferred via `take_input`), flush, finalize. FROM side: bind, init global / init local, exec, progress. Option lists arrive single-valued (automatically packed into `TUPLE`s).

### Cast Functions
 User-defined casts keyed by (source, target) type, with an implicit-cast cost and a `CAST_MODE` telling `TRY_CAST` apart.

### Replacement Scans

Replacement scans registered on a connection, database or extension: the callback resolves an unknown table name to a function call, a collection or a subquery, optionally aliased.

### Expressions

Read-only inspection of bound expressions handed out by filter pushdown: one `EXPRESSION_TYPE` enum mirroring the engine's values for the node types a filter can contain, children, return type, and type-prefixed accessors for constants, column references, functions (name and owned qualified name) and cast mode.

### File System

 The engine's virtual file system from a connection or context: open options with `FILE_FLAG`s, read / write / read-at / write-at, seek, tell, size, sync.

### Catalog

`connection_describe_table` returning a table description: qualified name, read-only flag, and per-column name, type, default and generated markers.

### Arrow

Arrow C Data Interface bridges: an importer turning Arrow arrays into chunks, an exporter turning chunks into arrays, and `result_to_arrow_stream` to turn streaming query result into arrow streams.

### Prepared Statements

 Create, execute with parameters, destroy, and check/guarantee if a statement will actually be re-used or re-bound with `reuses_plan`.

### Identifiers

`identifier_render_quoted`, the engine's quoting rules for embedding a name in SQL.

### Qualified Names

Parse, create from parts, part access, render, equality, hash. Also adopted by `create_type_from_name`, which now takes a qualified name instead of an identifier.

### Custom Types 

Registering a named alias over a base type on a connection or extension. Will hopefully support custom type "signatures" w/ modifiers/parameters soon.

### Column Data Collections

`column_data_collection_clear` added to the existing module to make it more efficient to re-use CDCs - making it possible to implement e.g. an `Appender` in user-space (like the C++ API wrapper)

### DIFF
--- a/api_spec/v2/arrow/arrow.yaml
+++ b/api_spec/v2/arrow/arrow.yaml
@@ -5,60 +5,78 @@ module: arrow
 #
 # The three standard Arrow structs -- ArrowSchema, ArrowArray and ArrowArrayStream -- are
 # emitted into the generated header under the same ARROW_C_DATA_INTERFACE /
-# ARROW_C_STREAM_INTERFACE guards that pyarrow, nanoarrow and arrow use, so the definitions
+# ARROW_C_STREAM_INTERFACE guards the Arrow specification prescribes, so the definitions
 # coexist with a consumer's own Arrow headers (first definition wins; the layouts are
-# byte-identical by spec). They keep their upstream names and are deliberately NOT wrapped in
-# opaque duckdb_v2_arrow_* handles, because the ownership contract is caller-allocates /
-# library-fills: a binding author must be able to declare an ArrowSchema on the stack with only
-# this header in hand.
+# byte-identical by spec). They keep their upstream names and are not wrapped in opaque
+# duckdb_v2_arrow_* handles: the ownership contract is caller-allocates / library-fills, so a
+# binding must be able to declare an ArrowSchema on the stack with only this header in hand.
 #
-# Ownership follows the Arrow C Data Interface, not this API's usual handle rules, and is
-# restated per function. Exporting fills a caller-allocated struct and sets its `release`, which
-# the caller then calls. Importing transfers the source array's data into the produced data
-# chunk and nulls the source's `release`, so the caller must not release it afterwards.
+# Two handles cover the two directions:
 #
-# The converters take a `context` because they are meant to run inside a callback -- a scalar
-# function, a table function, a cast -- that already holds one with an active transaction.
-# Building an Arrow schema is not a pure function of the logical types: extension
-# populate-schema callbacks and ENUM dictionaries read the catalog. Code outside a callback
-# reaches Arrow through `result_to_arrow_stream`, which supplies the context itself.
+#   arrow_importer   created from an ArrowSchema, converts ArrowArrays into data chunks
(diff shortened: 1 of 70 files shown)


thomas-0816:
I have a general question on v2-api:

Is double calling (to get the length) the expected way of using v2-api?

e.g.

```
/* Pass 1: size the buffer (out_text = NULL) without rendering. */
if (duckdb_v2_logical_type_to_text(v2_type, NULL, 0, &decl_len, &v2_err) == DUCKDB_V2_ERROR_NONE) {
    char *decl_buf = (char *) emalloc(decl_len + 1);
    /* Pass 2: render into the buffer; on failure fall back to the short name. */
    if (duckdb_v2_logical_type_to_text(v2_type, decl_buf, decl_len + 1, &decl_len, &v2_err) == DUCKDB_V2_ERROR_NONE) {
```

Maxxen:
> I have a general question on v2-api:
> 
> Is double calling (to get the length) the expected way of using v2-api?
> 
> e.g.
> 
> ```
> /* Pass 1: size the buffer (out_text = NULL) without rendering. */
> if (duckdb_v2_logical_type_to_text(v2_type, NULL, 0, &decl_len, &v2_err) == DUCKDB_V2_ERROR_NONE) {
>     char *decl_buf = (char *) emalloc(decl_len + 1);
>     /* Pass 2: render into the buffer; on failure fall back to the short name. */
>     if (duckdb_v2_logical_type_to_text(v2_type, decl_buf, decl_len + 1, &decl_len, &v2_err) == DUCKDB_V2_ERROR_NONE) {
> ```

Yes, but Im not happy about it. Im planning to move into using callbacks for all non-referenced strings, like e.g. `duckdb_v2_result_render_box`

Maxxen:
CI failure unrelated

PR #25348, opened 2026-09-04
Fix INTERNAL bind error for UNNEST comma join with CAST filter

### LINKED ISSUE #25167: INTERNAL Error: "Failed to bind column reference" for `unnest()` in a comma cross join
### What happens?

`SELECT t0.c1 FROM t0, unnest(t0.c1) AS x(c0) WHERE CAST('x' AS BOOLEAN)` raises `INTERNAL Error: Failed to bind column reference "" [20.0] (bindings: {#[2.0]})` instead of the `Conversion Error` the same predicate produces everywhere else.

Introduced by `fe871ebf45d777c97950cfe3e07a1a77a45f1503` (PR #23333, "Step 6: Flip the switch and have delim join as CTEs by default", landed 2026-06-17); verified present at the culprit and absent at its parent.

*Found by an automated fuzzing/triage agent; the repro is verified, but the above analysis may be wrong.*

### To Reproduce

```sql
CREATE TABLE t0(c1 INTEGER[]);
INSERT INTO t0 VALUES ([1]);

SELECT t0.c1 FROM t0, unnest(t0.c1) AS x(c0) WHERE CAST('x' AS BOOLEAN);
-- Expected: Conversion Error: Could not convert string 'x' to BOOL
-- Actual:   INTERNAL Error: Failed to bind column reference "" [20.0] (bindings: {#[2.0]})

-- the same query without the base column in the SELECT list:
SELECT 1 FROM t0, unnest(t0.c1) AS x(c0) WHERE CAST('x' AS BOOLEAN);
-- Conversion Error: Could not convert string 'x' to BOOL

-- the same query without the unnest:
SELECT t0.c1 FROM t0 WHERE CAST('x' AS BOOLEAN);
-- Conversion Error: Could not convert string 'x' to BOOL
```

### OS:

Ubuntu 22.04

### DuckDB Version:

v1.6.0-dev13791 (8616efa9da)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Suyang Zhong

### Affiliation:

NUS

- [x] Yes, I have

- [x] Yes, I have

asubbarao:
Fixes https://github.com/duckdb/duckdb/issues/25167

`SELECT t0.c1 FROM t0, unnest(t0.c1) AS x(c0) WHERE CAST('x' AS BOOLEAN)` raised `INTERNAL Error: Failed to bind column reference "" [20.0]` instead of the Conversion Error the same CAST produces without UNNEST or without the base column in the SELECT list.

Introduced when delim joins became CTEs by default (#23333). The unnest rewriter inlined the domain CTE but only remapped CTE refs in the join subtree (`topmost_op`). Projection pullup can leave a projection *above* a pushed constant filter, so that projection still pointed at the removed CTE scan.

Visit `domain_cte.children[1]` (the whole continuation) so those pulled-up projections are rewritten too.

Verified on the v2.0-cyanoptera binary:

- issue SQL → Conversion Error: Could not convert string 'x' to BOOL
- `test/issues/general/test_25167.test` — 9 assertions
- `test/optimizer/unnest*` — 181 assertions
- `test/sql/types/list/unnest*` — 599 assertions

### DIFF
--- a/src/optimizer/unnest_rewriter.cpp
+++ b/src/optimizer/unnest_rewriter.cpp
@@ -454,8 +454,10 @@ bool UnnestRewriter::RewriteInlineCTEDedupCandidate(unique_ptr<LogicalOperator> 
 		return false;
 	}
 
-	// After replacing the CTE refs by the shared domain producer, expressions in the join subtree must read
-	// directly from the producer bindings. The dedup bindings are retargeted to their original delimiter columns.
+	// After replacing the CTE refs by the shared domain producer, expressions in the continuation must
+	// read directly from the producer bindings. Projection pullup can leave projections above a pushed
+	// filter/join; those still reference the domain CTE scan and must be rewritten with the join subtree.
+	// The dedup bindings are retargeted to their original delimiter columns.
 	ColumnBindingReplacer domain_ref_replacer;
 	for (idx_t binding_idx = 0; binding_idx < lhs_bindings.size(); binding_idx++) {
 		domain_ref_replacer.replacement_bindings.emplace_back(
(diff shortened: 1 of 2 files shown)


kryonix (approved):
Thanks for the PR! Looks good. This is the correct fix for the issue.

PR #25350, opened 2026-09-04
[Backport] Add issue tracker workflow

DinosL:
Backporting to `2.0`. I was today years old when I learned that workflows only run on the default branch, which we recently switched from `main` to `2.0`.


Mytherin:
Thanks!

PR #25352, opened 2026-09-04
Use CompressionSegmentReader for uncompressed strings

### LINKED ISSUE #21916: SEGV in ReadStringMarker after corrupting a persisted row offset in an uncompressed VARCHAR segment
### What happens?

This issue is triggered by editing a persisted DuckDB database image and then running:

```sql
SELECT account_id, kind, length(payload)
FROM account_events
ORDER BY account_id;
```

On an ASAN build the process crashes with the following stack:

```text
duckdb::UncompressedStringStorage::ReadStringMarker
<- duckdb::UncompressedStringStorage::StringScanPartial
<- duckdb::ColumnData::ScanVector
<- duckdb::ColumnData::ScanVector
<- duckdb::StandardColumnData::Scan
<- duckdb::RowGroup::Scan
<- duckdb::CollectionScanState::Scan
<- duckdb::DataTable::Scan
<- duckdb::DuckTableScanState::TableScanFunc
<- duckdb::TableScanFunc
<- duckdb::PhysicalTableScan::GetDataInternal
<- duckdb::PipelineExecutor::GetData
<- duckdb::PipelineExecutor::FetchFromSource
<- duckdb::PipelineExecutor::Execute
<- duckdb::PipelineTask::ExecuteTask
<- duckdb::ExecutorTask::Execute
<- duckdb::TaskScheduler::ExecuteForever
```

The semantic mutation is not an arbitrary byte flip. It targets the most significant byte of the first `int32_t` row offset in the offset table of a `duckdb::UncompressedStringStorage` segment. `StringScanPartial()` reads that table as:

```c++
auto base_data = reinterpret_cast<int32_t *>(baseptr + DICTIONARY_HEADER_SIZE);
auto current_offset = base_data[start + i];
```

In this reduced case, `account_events.payload` is stored as an `Uncompressed` `VARCHAR` segment at `block_id = 3` and `block_offset = 56`. After the 8-byte dictionary header, the first row offset is normally `48`. The minimized replay flips only the high byte of that first offset from `0x00` to `0xFF`; the other five changed bytes only repair checksums for this concrete file layout.

That mutation turns a small positive dictionary offset into a large negative `int32_t`. Once that happens, `duckdb::UncompressedStringStorage::FetchStringFromDict()` stops taking the regular-string branch and instead treats the value as an overflow marker:

```c++
ReadStringMarker(base_ptr + dict_end_offset - AbsValue<int32_t>(dict_offset), block_id, offset);
```

With the corrupted offset, that pointer is driven far before the valid dictionary/marker region. `ReadStringMarker()` then blindly reads a `block_id_t` and an `int32_t` with `memcpy`, which becomes an out-of-bounds read and an ASAN `SEGV`.

### To Reproduce

Use an ASAN-enabled DuckDB CLI from a version that still contains this bug. The script builds the binary, applies the minimized replay patch, and reproduces the crash in a single step.
ASAN is optional, but it provides a clearer, more informative stack trace.

```sh
#!/usr/bin/env bash
set -euo pipefail

ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
DUCKDB_BIN="${1:-${DBIMAGE_DUCKDB_CLI_BIN:-$ROOT/target/duckdb_cli_asan}}"
if [[ ! -x "$DUCKDB_BIN" ]]; then
  DUCKDB_BIN="$(command -v "$DUCKDB_BIN" 2>/dev/null || true)"
fi
if [[ -z "$DUCKDB_BIN" || ! -x "$DUCKDB_BIN" ]]; then
  echo "duckdb binary is not executable: ${1:-${DBIMAGE_DUCKDB_CLI_BIN:-$ROOT/target/duckdb_cli_asan}}" >&2
  exit 1
fi

workdir="$(mktemp -d)"
trap 'rm -rf "$workdir"' EXIT
db="$workdir/read-string-marker-asan.duckdb"

# Create the minimal base database with the same ASAN DuckDB CLI that will reopen it.
"$DUCKDB_BIN" "$db" \
  -c "CREATE TABLE accounts(id INTEGER, email VARCHAR, active BOOLEAN, quota BIGINT);" \
  -c "INSERT INTO accounts VALUES (100, 'alice@example.com', true, 4096), (101, 'bob@example.com', false, 8192), (102, 'carol@example.com', true, 16384);" \
  -c "CREATE TABLE account_events(account_id INTEGER, kind VARCHAR, payload VARCHAR);" \
  -c "INSERT INTO account_events VALUES (100, 'login', 'ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-ok-'), (101, 'sync', 'state-state-state-state-state-state-state-state-state-state-state-state-'), (102, 'quota', 'raise-raise-raise-raise-raise-raise-raise-raise-raise-raise-');" \
  -c "CREATE INDEX idx_accounts_email ON accounts(email);" \
  >/dev/null

# Make sure the current on-disk layout still matches this exact replay patch.
base_header="$(od -An -tx1 -j 798723 -N 5 "$db" | tr -d ' \n')"
base_offset="$(od -An -tx1 -j 798795 -N 1 "$db" | tr -d ' \n')"
if [[ "$base_header" != "3fff0b10fb" || "$base_offset" != "00" ]]; then
  echo "The base layout changed for this DuckDB build; this exact replay patch must be regenerated." >&2
  echo "expected header=3fff0b10fb offset=00, got header=$base_header offset=$base_offset" >&2
  exit 1
fi

# The replay uses five checksum repair bytes plus the high byte of the first row offset.
printf '\xE4\xD2\x0C\x68\x8B' | dd of="$db" bs=1 seek=798723 conv=notrunc status=none
printf '\xFF' | dd of="$db" bs=1 seek=798795 conv=notrunc status=none

# Run the trigger query. On an affected ASAN build it aborts in ReadStringMarker.
"$DUCKDB_BIN" "$db" \
  -c "SELECT account_id, kind, length(payload) FROM account_events ORDER BY account_id;"
```

### OS:

Ubuntu 22.04.1 (Linux 6.8.0-106, x86_64)

### DuckDB Version:

1.5.1

### DuckDB Client:

duckdb cli

### Hardware:

_No response_

### Full Name:

Zi-Xuan Fu

### Affiliation:

Beihang University

- [x] Yes, I have

- [x] Yes, I have

Yes

HendrikLambert:
This PR moves uncompressed string reads to `CompressionSegmentReader`

Fixes #21916
Fixes https://github.com/duckdblabs/duckdb-internal/issues/10538

## Changes

- Bounds-check the dictionary header, offset table and dictionary entries.
- Validate dictionary offset bounds, ordering and overflow marker encoding before each entry is used.
- Require overflow block references to belong to the string segment.
- Bounds-check on-disk overflow pages and in-memory overflow buffers, including chained block reads.
- Assert logical access ranges for scans, selects and row fetches.

## Performance

Benchmark of two variants against a baseline. `ValidateDictionaryEntryFast` is a SIMD friendly version of `ValidateDictionaryEntry` that matches the logic with boolean arithmetic.

| Workload | Baseline | Without Fast | With Fast | Overhead with Fast |
|---|---:|---:|---:|---:|
| Full scan, 16-byte strings | 7.18 | 18.18 | 12.42 | +73% |
| Full scan, 8-byte inline strings | 17.05 | 26.52 | 20.65 | +21% |
| Full scan, NULL/empty strings | 10.47 | 21.96 | 16.25 | +55% |
| Select 0.1% | 5.39 | 13.60 | 7.82 | +45% |
| Select 33% | 8.73 | 17.29 | 11.58 | +33% |
| Select 99% | 12.29 | 22.79 | 16.87 | +37% |
| FetchRow, 100,000 rows | 8.64 | 9.03 | 9.00 | +4% |

Measured on Apple M1 against upstream `7e1c37c0e9`. Times are medians of 200 runs in milliseconds.


src/include/duckdb/storage/string_uncompressed.hpp
```diff
@@ -224,18 +268,16 @@ struct UncompressedStringStorage {
...
 	static string_t ReadOverflowString(const QueryContext &context, ColumnSegment &segment, Vector &result,
 	                                   block_id_t block, int32_t offset);
-	static string_t ReadString(data_ptr_t target, int32_t offset, uint32_t string_length);
-	static string_t ReadStringWithLength(data_ptr_t target, int32_t offset);
+	static string_t ReadStringWithLength(CompressionSegmentReader &reader, int32_t offset);
 	static void WriteStringMarker(data_ptr_t target, block_id_t block_id, int32_t offset);
-	static void ReadStringMarker(data_ptr_t target, block_id_t &block_id, int32_t &offset);
 
+	// FIXME: Remove this raw pointer helper once FSST no longer needs it.
 	inline static string_t FetchStringFromDict(const QueryContext &context, ColumnSegment &segment,
```
Tishj: Don't we want to push the `CompressionSegmentReader` into this as well?

src/storage/compression/string_uncompressed.cpp
```diff
@@ -384,9 +500,13 @@ string_t UncompressedStringStorage::ReadOverflowString(const QueryContext &conte
 		auto handle = buffer_manager.Pin(context, block_handle);
 
 		// read header
-		uint32_t length = Load<uint32_t>(handle.GetDataMutable() + offset);
+		auto block_size = segment.GetBlockSize();
+		auto string_space = block_size - sizeof(block_id_t);
+		CompressionSegmentReader block_reader(handle.GetDataMutable(), block_size, "overflow string block");
```
Tishj: Just noticed this says `GetDataMutable`, we should only need a const pointer, no?

src/storage/compression/string_uncompressed.cpp
```diff
@@ -434,20 +553,15 @@ string_t UncompressedStringStorage::ReadOverflowString(const QueryContext &conte
...
+	CompressionSegmentReader reader(final_buffer, string_block.get().offset, "in-memory overflow string block");
+	return ReadStringWithLength(reader, offset);
 }
 
-string_t UncompressedStringStorage::ReadStringWithLength(data_ptr_t target, int32_t offset) {
-	auto ptr = target + offset;
-	auto str_length = Load<uint32_t>(ptr);
-	auto str_ptr = char_ptr_cast(ptr + sizeof(uint32_t));
-	return string_t(str_ptr, str_length);
+string_t UncompressedStringStorage::ReadStringWithLength(CompressionSegmentReader &reader, int32_t offset) {
```
Tishj: Just a note, `ReadStringWithLength` now modifies the offset of the `reader`, whereas before it had no side-effects on the input. Maybe you want to create a copied `CompressionSegmentReader` to modify, just because this side-effect might be unexpected

Tishj (changes requested):
Thanks, I left some comments, but it looks pretty good.
I'm not sure if keeping the `DictionaryCursor` makes sense though, since it's only used in a single location.

src/storage/compression/string_uncompressed.cpp
```diff
@@ -9,6 +9,137 @@
...
+                                                      uint32_t previous_dictionary_offset) const {
+	auto current_dictionary_offset = ValidateAndGetDictionaryOffset(current_offset);
+	if (current_dictionary_offset < previous_dictionary_offset) {
+		ThrowDecreasingStringOffset(current_dictionary_offset, previous_dictionary_offset);
+	}
+
+	// The dictionary offset must advance by a string's size or BIG_STRING_MARKER_SIZE.
+	auto string_length = current_dictionary_offset - previous_dictionary_offset;
+	if (current_offset < 0) {
+		if (string_length == 0) {
```
Tishj: These could maybe use some extra comments

Mytherin:
Thanks for the PR - code looks good but I'm mostly worried about the perf aspect of these changes. Some AI suggestions:

> 1. Validation is per value where it could be per array. ValidateAndGetDictionaryOffset runs the INT32_MIN check, the sign branch, and the > dictionary_size bounds check on every string, and CreateDictionaryEntry adds a monotonicity check plus two more on the negative path. Since the offsets are monotonic by construction, a single pass over the int32_t offsets array per segment (or per vector) validating bounds + ordering would let the inner loop go back to the old branchless abs(cur) - abs(prev) form. The offsets are already being read, and that pass vectorizes.
> 2. Select/StringFetchRow re-validate the previous offset for every row. GetDictionaryEntry calls ValidateAndGetDictionaryOffset(offsets[i-1]) in addition to the current one, doubling the check cost — which is why the Select path (+116%) is worse than the sequential one. I verified this by building a variant that does the same thing in the sequential loop: it went to +288%, worse than the PR. (Corollary: the loop-carried previous_dictionary_offset accumulator is not the problem — removing it costs more than it saves.)
> 3. UncompressedStringSegmentState::GetHandle gained an O(n) std::find over on_disk_blocks plus a TotalBlocks() call that takes the block-manager mutex, on first touch of each overflow block, under block_lock. It's amortized by the handles map, so it didn't show in my overflow benchmark (2000 strings), but it's O(n²) in overflow blocks per segment over a cold scan. Cheap to fix with a set.

Could you have a look and do some perf analysis + try to minimize perf impact?

Tishj (approved):
Thanks!

PR #25353, opened 2026-09-04
Fix bind for summarize for stdin

Dtenwolde:
Fix https://github.com/duckdblabs/duckdb-internal/issues/7725

`SUMMARIZE` was binding the input twice, which breaks for inputs such as `/dev/stdin`. During the first bind, csv schema detection read from the stdin, and then the second bind encountered an exhausted stream. 
Now we reuse the initially bound plan instead.


Mytherin:
Thanks!

PR #25354, opened 2026-09-04
Add helpful error for alter table drop constraint

Dtenwolde:
Part of https://github.com/duckdblabs/duckdb-internal/issues/9093

Even though we don't support `ALTER TABLE x DROP CONSTRAINT`, I think it's useful to throw a more descriptive error like we did in `v1.5` than a generic syntax error: 
```sql
memory D ALTER TABLE t DROP CONSTRAINT pk;
Not implemented Error:
No support for that ALTER TABLE option yet!
```


Mytherin:
Thanks!

PR #25356, opened 2026-09-04
Fix compiler warnings (variable unused or method)

smvv:
Different compilers show many warnings CI jobs:
- https://github.com/duckdb/duckdb/actions/runs/33839747683/job/100958659658 (gcc 13)
- https://github.com/duckdb/duckdb/actions/runs/33839747683/job/100919594728 (clang 20)
- https://github.com/duckdb/duckdb/actions/runs/33839747683/job/100921488698 (gcc 14)
- https://github.com/duckdb/duckdb/actions/runs/33839747683/job/100919594730 (clang 21, LTO)

Summary of the solved warnings:

| Jobs | Warning groups |
|---|---|
| GCC 13 / GCC 14 | `numeric.cpp`: anonymous-namespace bind data captured by a lambda (`-Wsubobject-linkage`); `date_part.cpp`: two non-void switch functions missing a final return/throw (`-Wreturn-type`); `catalog.cpp`: four deprecated internal `GetEntry` calls; `buffer_pool.cpp`: possible null producer-token dereference reported as `-Wstringop-overflow` inside concurrentqueue. |
| Clang 20 / Clang 21 LTO | 30 assertion-related warnings in release builds: 23 unused variables, 2 variables set but unused, 2 unused helper functions, and 3 assertion-only planner helpers with unnecessary internal linkage. |

The Clang warnings occur because `D_ASSERT` compiles out in release mode.

### DIFF
--- a/extension/core_functions/scalar/math/numeric.cpp
+++ b/extension/core_functions/scalar/math/numeric.cpp
@@ -1062,14 +1062,17 @@ struct DecimalRoundNegativePrecisionOperator {
 		T divide_power_of_ten =
 		    UnsafeNumericCast<T>(POWERS_OF_TEN_CLASS::POWERS_OF_TEN[-info.target_scale + source_scale]);
 		T multiply_power_of_ten = UnsafeNumericCast<T>(POWERS_OF_TEN_CLASS::POWERS_OF_TEN[-info.target_scale]);
+		const bool check_overflow = info.check_overflow;
 
 		UnaryExecutor::Execute<T, T>(input.data[0], result, [&](T input) {
 			auto rounded =
 			    UnsafeNumericCast<T>(RoundDivide<T, ROUND_POLICY>(input, divide_power_of_ten) * multiply_power_of_ten);
-			if constexpr (std::is_same_v<T, hugeint_t>) {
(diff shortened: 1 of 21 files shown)


extension/core_functions/scalar/math/numeric.cpp
```diff
@@ -1018,10 +1019,12 @@ struct DecimalRoundNegativePrecisionOperator {
 				input += addition;
 			}
 			auto rounded = UnsafeNumericCast<T>(input / divide_power_of_ten * multiply_power_of_ten);
-			if constexpr (std::is_same_v<T, hugeint_t>) {
-				if (info.check_overflow && (rounded <= -Hugeint::POWERS_OF_TEN[Decimal::MAX_WIDTH_DECIMAL] ||
-				                            rounded >= Hugeint::POWERS_OF_TEN[Decimal::MAX_WIDTH_DECIMAL])) {
-					throw OutOfRangeException("Overflow in ROUND of DECIMAL(38)");
```
smvv: `RoundPrecisionFunctionData` is in an anonymous namespace, but the lambda captures `info` inside an externally linked template:
```
warning: lambda has a field whose type uses the anonymous namespace [-Wsubobject-linkage]
```

GCC warns because the lambda then contains a reference to a translation-unit-local type.

Fix: copy `info.check_overflow` into a `bool` and capture that instead.

Mytherin:
Thanks!

PR #25362, opened 2026-09-04
Make decimal division round to even

Dtenwolde:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/10550

Decimal division currently truncates the result, while decimal average rounds it. This changes `decimal_division` to round to nearest with ties to even, keeping the behavior consistent.

The separate `round_even` issue is handled in: https://github.com/duckdb/duckdb/pull/25197


Tishj:
Isn't this a breaking change, that needs a setting to enable/disable for backwards compatibility?
Similar to https://github.com/duckdb/duckdb/pull/25004

Dtenwolde:
`decimal_division` is not in the `v1.5-variegata` releases, so I don't think so? Happy to add a setting otherwise

Mytherin:
Thanks! I think that's indeed not necessary.

PR #25363, opened 2026-09-04
[Parser] Rework `Matcher` logic to make recursive/iterative share the same flow

Tishj:
Currently there are two separate paths for a Matcher to evaluate with, one is recursive, the other is iterative.

They share very little, which is problematic when it comes to extensions.
This PR creates common structures that can be used irrespective of the stack/heap approach to executing the matching process.

#### `MatchProcess`
Created by a `Matcher` to start the matching process, holds the state of the matching process if the matcher is stateful (for example it has children).

##### `MatchProcess::Resume([child])`
Return either a result or the input to a child matcher to invoke, of which the result should be fed back with the next `Resume` call of the process.

#### `MatchStep`
Holds either a result or the input to start a child matcher, returned by `Resume`.


Mytherin:
Thanks!

PR #25370, opened 2026-09-04
Properly transform qualified columns in rename/exclude

Dtenwolde:
cc @evertlammerts 

The previous implementation for parsing columns in `SELECT * EXCLUDE()` and `RENAME` was a bit too simplistic and lost the context of whether columns were wrapped with `""` or not. This caused the following queries to have incorrect results compared to v1.5.5, but are now fixed: 
```sql
memory D SELECT * RENAME ("a.b" AS d) FROM (SELECT 1 AS "a.b", 2 AS c);
┌───────┬───────┐
│   d   │   c   │
│ int32 │ int32 │
├───────┼───────┤
│     1 │     2 │
└───────┴───────┘
memory D  SELECT * EXCLUDE ("a.b") FROM (SELECT 1 AS "a.b", 2 AS c);
┌───────┐
│   c   │
│ int32 │
├───────┤
│     2 │
└───────┘
```

I only handle up to 4 components and not an arbitrary amount, is that something worth adding as well?

### DIFF
--- a/src/include/duckdb/parser/peg/transformer/peg_transformer.hpp
+++ b/src/include/duckdb/parser/peg/transformer/peg_transformer.hpp
@@ -394,6 +394,7 @@ public:
 	static vector<reference<ParseResult>> ExtractParseResultsFromList(ParseResult &parse_result);
 	static bool ExpressionIsEmptyStar(const ParsedExpression &expr);
 	static QualifiedName StringToQualifiedName(vector<string> input);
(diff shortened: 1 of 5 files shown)


Mytherin:
Thanks - I think we should adjust this to be able to handle an arbitrary amount of identifiers, though. Maybe you can pick that up in a follow-up?

Dtenwolde:
> Maybe you can pick that up in a follow-up?

Sure!

PR #25372, opened 2026-09-04
Update compression tests for extension runs

staticlibs:
A few of recently added compression tests use:

```sql
load {TEST_DIR}/path_to_db_file.db
```

And expect the loaded DB to be available as a default catalog.

This is not the case in DuckLake test runs that use an atached DuckLake catalog as a default one using `--test-config test/configs/attach_ducklake.json` ([link](https://github.com/duckdb/ducklake/blob/0f569a6bba3028637c5020ccd31af24739e3fb42/test/configs/attach_ducklake.json)).

This PR adds `USE <catalog_name>` calls at the beginnings of problematic tests.

CC: @pdet, @HendrikLambert

### DIFF
--- a/test/sql/storage/compression/alp/alp_corrupt_count.test
+++ b/test/sql/storage/compression/alp/alp_corrupt_count.test
@@ -8,6 +8,9 @@ unzip data/storage/corrupt/alp_corrupt_count.db.gz {TEST_DIR}/alp_corrupt_count.
 
 load {TEST_DIR}/alp_corrupt_count.db readonly
 
+statement ok
+USE alp_corrupt_count
+
 query I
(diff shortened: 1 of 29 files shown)


HendrikLambert:
Hi @staticlibs, https://github.com/duckdb/duckdb/pull/25263 also added more tests, so did https://github.com/duckdb/duckdb/pull/24911, I suppose those should also be updated?

HendrikLambert:
Sorry, forgot about [these](https://github.com/duckdb/duckdb/pull/25076)

staticlibs:
Thanks for the pointers! Let me update these ones too.

staticlibs:
> Sorry, forgot about https://github.com/duckdb/duckdb/pull/25076

These three look like they don't rely on the current catalog, should be fine.

PR #25374, opened 2026-09-04
Add shared SWAR word primitives

pdet:
This PR adds duckdb/common/swar.hpp, a header with the word-at-a-time byte primitives (SWAR, "SIMD within a register").


Mytherin:
Thanks!

PR #25375, opened 2026-09-04
Return an input error for invalid regex split patterns

fvaleye:
Invalid dynamic regex patterns currently raise an internal error in the regex split functions.

Mark these functions as fallible so they return an `Invalid Input Error`, matching constant patterns.


Mytherin:
Thanks!

PR #25376, opened 2026-09-04
Fix chained function calls on case-insensitive column matches

### LINKED ISSUE #25285: 1.6.0.dev379: chained nullif expression fails during binding
### What happens?

chaining nullif fails on the latest 2.0.0 alpha. it worked fine with 1.5.

### To Reproduce

```
select strftime(
    ci.purchasedate.nullif('1899-12-30T00:00:00'), '%Y-%m-%d'
)
from (values (timestamp '2026-01-01')) as ci("PurchaseDate");
```

error: `Binder Error: Macro "nullif"() has positional argument following named argument`

### OS:

Ubuntu 22.04

### DuckDB Version:

1.6.0.dev379

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Daniel Gut

### Affiliation:

Aveniq

- [x] Yes, I have

- [x] Yes, I have

DinosL:
Chaining a macro onto a column referenced with a different letter case than it was declared with fails to bind. 

Spoiler: the letter case is not the actual problem.

Identifiers are case-preserving, so when the registered name and the column name differ the binder adds an alias carrying the original name to the qualified reference. That's fine until the function is rewritten and the alias lands in argument position, where it means a named parameter. This PR clears the alias before the reference becomes an argument.

Fixes: https://github.com/duckdb/duckdb/issues/25285

### DIFF
--- a/src/planner/column_qualifier.cpp
+++ b/src/planner/column_qualifier.cpp
@@ -302,6 +302,7 @@ optional_ptr<CatalogEntry> ColumnQualifier::QualifyFunction(FunctionExpression &
 	if (!new_colref) {
 		new_colref = std::move(colref);
 	}
+	new_colref->ClearAlias();
 	// we can! transform this into a function call on the column
 	// i.e. "x.lower()" becomes "lower(x)"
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25378, opened 2026-09-04
Bound the unselected prefix range key lookup by the filter count

J-Meyers:
`PrefixRangeBitmap::LookupKeys` (no-selection overload) ignored its `count`
parameter and iterated `keys.ValidValues<T>()`, which spans the vector's
buffer capacity rather than the rows being filtered. The caller sizes
`result_sel` to `approved_tuple_count` via `PrepareCapacity`, so a vector
holding more valid values than that count overflows the selection buffer.

The write is unconditional per iteration (the cursor only advances on a
match), so it overflows even when nothing matches.

Fixed by bounding the loop by `count`, matching the selection overload
directly below it.

Found via AddressSanitizer debugging a flaky ducklake test:

```
WRITE of size 4 at 0x7767409efc78 thread T73
  #0 SelectionVector::set_index                          selection_vector.hpp:126
  #1 PrefixRangeBitmap<uint64_t>::LookupKeys<int64_t,..> table_filter_prefix_range_function.cpp:148
  #2 NumericPrefixRangeFilter<int64_t>::LookupKeys       :322
  #3 PrefixRangeFilterExecutor::FilterSelection          table_filter_state.cpp:386
0x7767409efc78 is located 0 bytes after 88-byte region
```

88 bytes = 22 `sel_t` entries; the write lands at index 22.

No test: existing prefix-range tests do not trip ASAN, and I could not
construct a deterministic DuckDB-level reproducer. Observed via a DuckLake
concurrency test on a musl build, where the corrupted chunk aborts in
`free()`; glibc tolerates it silently.

Made with AI help

### DIFF
--- a/src/planner/filter/table_filter_prefix_range_function.cpp
+++ b/src/planner/filter/table_filter_prefix_range_function.cpp
@@ -136,16 +136,24 @@ public:
 
 	template <typename T, typename CONVERTER>
 	idx_t LookupKeys(Vector &keys, SelectionVector &result_sel, idx_t count) const {
+		UnifiedVectorFormat key_data;
+		keys.ToUnifiedFormat(key_data);
+
+		const auto keys_data = UnifiedVectorFormat::GetData<T>(key_data);
 		idx_t found_count = 0;
-		for (const auto &entry : keys.template ValidValues<T>()) {
-			const U comparable = CONVERTER::Convert(entry.GetValue());
+		for (idx_t i = 0; i < count; i++) {
+			const auto key_idx = key_data.sel->get_index(i);
+			if (!key_data.validity.RowIsValid(key_idx)) {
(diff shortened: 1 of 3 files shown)


Mytherin:
Thanks - but we shouldn't be using the legacy `ToUnifiedFormat` anymore. Could you rewrite this to use `.Values<T>` instead? Could you also add a test that triggers this issue?

J-Meyers:
@Mytherin

I could not get consistent reproduction with any SQL only tests, added a Catch2 style test instead, I tried looping with the SQL but still it was quite intermittent and difficult to repro

Made the change also to other LookupKeys just to keep the two functions mirroring each other

Mytherin:
Alright - thanks for looking into it. We can keep it like this for now.

PR #25380, opened 2026-09-04
Experiment: Remove manually unrolled memory routines

smvv:
I tried removing fast memory routines (manually unrolled `memcpy`/`memset`/`memcmp`) and I did not see any performance regression in CI.

I think not having regressions is explained by using CPU baseline v3 for duckdb core, since Clang can now optimize those routines. Without those instructions, the routines can become a bottleneck because they are used in most of the codebase.


Mytherin:
Thanks for the PR!

@lnkuiper given you added these before - is there anything else (another benchmark or platform) we should test prior to removing these?

smvv:
Had a DM chat with Laurens;
- he asked to benchmark "sorting on macOS arm64" because that was improved by this code.
- I suggested we can use the benchmark results that run for alpha releases to spot regressions for macOS arm64 over time. We should see a regression in at least some benchmark queries if this code is still active.

PR #25382, opened 2026-09-05
Fix collation handling for DISTINCT aggregates (part 1)

### LINKED ISSUE #24319: Column collation is not propagated to the aggregate `DISTINCT` modifier or to `UNIQUE`/`PRIMARY KEY` constraints
### What happens?

For a column declared `VARCHAR COLLATE NOCASE`, DuckDB uses two different notions of string equality inside the
same table:

* the **comparison / grouping** paths (`=`, `SELECT DISTINCT`, `GROUP BY`, `UNION`, `INTERSECT`) correctly honour
  `NOCASE`;
* the **aggregate `DISTINCT` modifier** (`count(DISTINCT s)`, `list(DISTINCT s)`, `string_agg(DISTINCT s, …)`,
  `array_agg(DISTINCT s)`) and the **`UNIQUE` / `PRIMARY KEY` constraint enforcement** de-duplicate by raw bytes
  and ignore the collation.

The consequences are a silently wrong `count(DISTINCT …)` and a `UNIQUE` column that stores two rows the engine
itself considers equal — so an equality lookup on a `UNIQUE` column returns two rows.

### To Reproduce

```python
# python3.13, duckdb 1.5.5
import duckdb
c = duckdb.connect()
c.execute("CREATE TABLE z (id INTEGER, s VARCHAR COLLATE NOCASE)")
c.execute("INSERT INTO z VALUES (1,'a'),(2,'A'),(3,'a '),(4,'A ')")

# Baseline: the engine's own `=` gives exactly two equivalence classes {1,2} and {3,4}
# ('a' = 'A' is true under NOCASE; 'a' = 'a ' is false, DuckDB does not pad-trim).
print(c.execute("SELECT a.id, b.id FROM z a, z b WHERE a.s = b.s ORDER BY 1,2").fetchall())
# -> [(1,1),(1,2),(2,1),(2,2),(3,3),(3,4),(4,3),(4,4)]        2 classes

# Paths that agree with the baseline (all return 2 -- correct):
print(c.execute("SELECT count(*) FROM (SELECT DISTINCT s FROM z)").fetchall())               # -> [(2,)]
print(c.execute("SELECT count(*) FROM (SELECT s FROM z GROUP BY s)").fetchall())             # -> [(2,)]
print(c.execute("SELECT count(*) FROM (SELECT s FROM z UNION SELECT s FROM z)").fetchall())  # -> [(2,)]
print(c.execute("SELECT count(*) FROM (SELECT s FROM z INTERSECT SELECT s FROM z)").fetchall())# -> [(2,)]

# Defect 1 -- aggregate DISTINCT modifier de-duplicates by bytes:
print(c.execute("SELECT count(DISTINCT s) FROM z").fetchall())            # -> [(4,)]              WRONG, expected 2
print(c.execute("SELECT list(DISTINCT s) FROM z").fetchall())             # -> [(['a','A ','a ','A'],)]   WRONG
print(c.execute("SELECT string_agg(DISTINCT s,'/') FROM z").fetchall())   # -> [('A/a/A /a ',)]    WRONG
print(c.execute("SELECT array_agg(DISTINCT s) FROM z").fetchall())        # -> [(['A','a ','a','A '],)]   WRONG
print(c.execute("SELECT approx_count_distinct(s) FROM z").fetchall())     # -> [(4,)]              WRONG
print(c.execute("SELECT histogram(s) FROM z").fetchall())                 # -> [({'A':1,'A ':1,'a':1,'a ':1},)]  WRONG
print(c.execute("SELECT count(DISTINCT s) FILTER (WHERE id>0) FROM z").fetchall())        # -> [(4,)]  WRONG
print(c.execute("SELECT DISTINCT count(DISTINCT s) OVER () FROM z").fetchall())           # -> [(4,)]  WRONG

# An explicit COLLATE on the aggregate argument does not help either, so this is not merely a
# question of column metadata propagation -- the aggregate de-duplicator itself is not rewritten:
print(c.execute("SELECT count(DISTINCT v COLLATE NOCASE) FROM (VALUES ('a'),('A')) t(v)").fetchall())
# -> [(2,)]     WRONG, expected 1
print(c.execute("SELECT count(*) FROM (SELECT DISTINCT v COLLATE NOCASE FROM (VALUES ('a'),('A')) t(v))").fetchall())
# -> [(1,)]     correct control: the same COLLATE on SELECT DISTINCT works

# Defect 2 -- UNIQUE and PRIMARY KEY enforce byte equality:
c.execute("CREATE TABLE u (id INTEGER, s VARCHAR COLLATE NOCASE UNIQUE)")
c.execute("INSERT INTO u VALUES (1,'a')")
c.execute("INSERT INTO u VALUES (2,'A')")        # ACCEPTED -- WRONG, the engine's own `=` says 'a' = 'A'
print(c.execute("SELECT * FROM u ORDER BY id").fetchall())   # -> [(1,'a'),(2,'A')]
print(c.execute("SELECT id FROM u WHERE s = 'A' ORDER BY id").fetchall())
# -> [(1,),(2,)]   an equality lookup on a UNIQUE column returns TWO rows
try:
    c.execute("INSERT INTO u VALUES (3,'a')")
except Exception as e:
    print(e)  # Constraint Error: Duplicate key "s: a" violates unique constraint.  (byte duplicate IS caught)

c.execute("CREATE TABLE pk (s VARCHAR COLLATE NOCASE PRIMARY KEY)")
c.execute("INSERT INTO pk VALUES ('a')")
c.execute("INSERT INTO pk VALUES ('A')")         # ACCEPTED -- WRONG, same as UNIQUE
print(c.execute("SELECT count(*) FROM pk").fetchall())       # -> [(2,)]
```

### OS:

macOS arm64

### DuckDB Version:

1.5.5

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Jiahao Ding

### Affiliation:

Xiamen University

- [x] Yes, I have

- [x] Yes, I have

yes

Tobias2099:
Fixes #24319 

# Overview

This PR fixes collation handling for grouped and ungrouped `DISTINCT` aggregates by applying collation-aware comparison during the multi-stage aggregate rewrite. So, values such as `Apple` and `apple` are treated as equal under `NOCASE`. By applying collation to the rewrite’s grouping keys, DuckDB can reuse its existing collation-aware `GROUP BY` machinery instead of adding separate collation logic to the physical `DISTINCT` operators.

The rewrite converts a `DISTINCT` aggregate into an ordinary aggregate over a deduplicated input. The deduplication groups by the normalized comparison key, while hidden `first(...)` aggregates preserve an original representative for aggregates that return values such as `LIST(DISTINCT name)`.

# Implementation details

`CreateDistinctBranch` now builds deduplication branches using collation-aware comparison keys. Each `DISTINCT` child is copied and rebound to the branch input, then passed through `PushCollation` when necessary. The resulting normalized expressions become grouping keys.

When a child is normalized, the rewrite also creates a hidden `first(...)` aggregate over the original expression. The final aggregate consumes this representative instead of the normalized key, preserving the original value returned by aggregates like `LIST(DISTINCT name)`.

`DISTINCT` aggregates may share a deduplication branch only when their children, filters, ordering expressions, and collation metadata are compatible. This prevents expressions with different comparison semantics from incorrectly sharing state.

`ORDER BY` expressions are handled separately from `DISTINCT` children when necessary. Collation metadata is included when matching aggregate arguments to ordering keys, ensuring that expressions such as `name COLLATE NOCASE` and `name COLLATE C` are stored and ordered independently.

The new collation helpers compare and hash collation metadata recursively, including nested types such as lists, arrays, maps, structs, and unions. `ClientContext` is passed through the relevant binding paths because collation resolution and hidden aggregate binding require query and database state.

# Out of scope

When the optimizer is disabled, the multi-stage rewrite does not run and DuckDB falls back to the existing physical `DISTINCT` implementation. Collation handling in that fallback path remains a transition limitation and is outside the scope of this PR.

This PR does not cover window DISTINCT aggregates, UNIQUE or PRIMARY KEY constraints, or make the multi-stage rewrite mandatory.


src/execution/operator/aggregate/distinct_aggregate_data.cpp
```diff
@@ -115,6 +152,67 @@ DistinctAggregateData::DistinctAggregateData(const DistinctAggregateCollectionIn
 	}
```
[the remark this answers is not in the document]
Tobias2099: Fixed. Folded the change into the latest commit.

kryonix (changes requested):
Thanks for working on this! The collation issue definitely needs fixing, and preserving an original representative for aggregates such as `list` is an important part of that.

I'd prefer to address this in the multi-stage aggregate rewrite, though. We're moving towards expressing DISTINCT aggregates as ordinary aggregations over a deduplicated input, with CTEs sharing that input where needed. Eventually, I'd like to remove the dedicated DISTINCT machinery from the physical aggregate operators.

This PR takes us in the opposite direction: it adds normalization and representative-value handling to the physical DISTINCT path, then excludes collated aggregates from the rewrite. That makes the old path necessary for correctness and leaves us with functionality we'll need to move again before we can remove it.

Could we instead extend `CreateDistinctBranch` to group by the normalized comparison keys and retain original values through `first(...)` aggregates? The final non-distinct aggregate would then consume those representatives. We already use this approach when binding collated `GROUP BY` expressions, so that should provide a useful reference, and perhaps an opportunity to share some of the construction logic.

There is still a transition to account for: DISTINCT rewriting can currently be disabled, so supporting collation in the rewrite won't immediately let us delete the physical implementation. But I'd like the rewrite to be the primary implementation of these semantics. We can then work towards making that lowering mandatory.

The result tests here would be useful for that implementation too. I'd also add plan checks showing that collated DISTINCT aggregates continue through the multi-stage rewrite, including when mixed with other aggregates.

kryonix (commented):
Thanks for reworking this! This now fits the direction we want: collation handling lives in the multi-stage rewrite, with normalized grouping keys and original representatives passed to the final aggregate.

The remaining optimizer-disabled behavior is the transition limitation we discussed.

Looks good to me. Please update the PR description. It still describes the previous physical implementation and says the multi-stage rewrite is excluded.

Tobias2099:
> Thanks for reworking this! This now fits the direction we want: collation handling lives in the multi-stage rewrite, with normalized grouping keys and original representatives passed to the final aggregate.
> 
> The remaining optimizer-disabled behavior is the transition limitation we discussed.
> 
> Looks good to me. Please update the PR description. It still describes the previous physical implementation and says the multi-stage rewrite is excluded.

Thanks for the feedback! Really appreciate the detailed explanations in the comments. 
I have updated the description to match the multi-stage rewrite approach and have marked this PR for review.

src/optimizer/multi_stage_aggregate_rewriter.cpp
```diff
@@ -62,56 +65,91 @@ static optional_idx FindExpression(const vector<unique_ptr<Expression>> &express
...
 		}
 	}
 }
 
+static unique_ptr<BoundAggregateExpression> BindFirst(ClientContext &context, unique_ptr<Expression> child) {
+	auto first_function = FirstFunctionGetter::GetFunction(child->GetReturnType());
+	vector<unique_ptr<Expression>> children;
+	children.push_back(std::move(child));
+	return first_function.Bind(context, std::move(children));
+}
```
kryonix, line 117: Could you resolve `first` through the catalog here, in line with #25472 and #25602? The direct factory call bypasses catalog lookup, leaving the introduced aggregate without its catalog/schema identity. The helper could become:

```suggestion
static unique_ptr<BoundAggregateExpression> BindFirst(ClientContext &context, unique_ptr<Expression> child) {
	auto first_function =
	    GetBuiltinAggregateFunction(context, FirstFun::Name, {child->GetReturnType()});
	vector<unique_ptr<Expression>> children;
	children.push_back(std::move(child));
	return first_function->Bind(context, std::move(children));
}
```

This also needs `distributive_functions.hpp` for `FirstFun` and the `builtin_function_lookup.hpp` header. The lookup helper is already on `v2.0-cyanoptera`; its header moves from `duckdb/optimizer/` to `duckdb/function/` in #25602, so the include depends on the base used.

kryonix (changes requested):
One small follow-up to align this with #25472 and #25602: could you bind the hidden `first(...)` through `GetBuiltinAggregateFunction(context, FirstFun::Name, {child->GetReturnType()})` instead of `FirstFunctionGetter::GetFunction(...)`?

We're moving introduced builtins through catalog lookup so the bound function retains its catalog/schema identity. The direct factory call here bypasses that lookup. #25602 also preserves that identity when binding specializes the implementation.

The rewrite itself looks good. This is about keeping its function binding consistent with that work. After that change, this PR should be ready to go.

src/optimizer/multi_stage_aggregate_rewriter.cpp
Tobias2099: Fixed! I've applied the suggested catalog lookup change and verified the updated logic with local regression tests.

kryonix (approved):
Thanks! LGTM now. Approved.

kryonix:
Hey @Tobias2099, can you please rebase, and check out the CI failures? The Thread Sanitizer should be fixed after rebasing, don't worry about that one 😉

Tobias2099:
> Hey @Tobias2099, can you please rebase, and check out the CI failures? The Thread Sanitizer should be fixed after rebasing, don't worry about that one 😉

Looks like `collated.test` was failing for the optimizer-disabled runs.

I've rebased and added `collated.test` to the optimizer-disabled skip list. Should be fixed now!

PR #25383, opened 2026-09-05
Add comment and tags for table functions in bind callback.

rustyconover:
Table functions can optionally supply per-column comments and tags through `TableFunctionBindInput` in their existing bind callback. The metadata is carried by `LogicalGet` and exposed as JSON in the existing `extra` column of `DESCRIBE`, `SHOW`, and `pragma_show`, preserving the six-column result schema.

Metadata follows simple column references and aliases; computed expressions have no metadata. Views report their own column comments. Regression tests cover table and view comments, tagged extension columns, table-function metadata without scan execution, aliases, computed expressions, and ordinality.

Resolves the merge conflict with `main` while preserving the expected-schema fields used by multi-file readers.

Related discussion: #25381.

Validation on the branch merged with current `main`:

- `make reldebug` with JSON, ICU, TPC-H, and TPC-DS extensions: passed.
- 16 focused metadata, DESCRIBE, and multi-file regression tests: passed normally and with `debug_verify_serializer=true`.
- `make allunit` against the reldebug build: 7,240 passed, 110 skipped, 0 failures.
- Formatting and whitespace checks against `origin/main`: passed.

### DIFF
--- a/src/function/table/system/pragma_table_info.cpp
+++ b/src/function/table/system/pragma_table_info.cpp
@@ -10,6 +10,7 @@
 #include "duckdb/planner/binder.hpp"
 
 #include "duckdb/common/exception.hpp"
+#include "duckdb/common/json_document.hpp"
 #include "duckdb/common/limits.hpp"
 
 #include <algorithm>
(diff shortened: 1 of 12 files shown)


Mytherin:
Thanks - but we don't need this to be a new callback. We could just put the extra result columns in the `TableFunctionBindInput` that is passed into the function. Old functions will then not fill them while new functions optionally can. 

I also don't think it's desirable to adjust the result schema of `DESCRIBE`. Note that the result schema of `DESCRIBE` matches the one from MySQL on purpose. This also already has the `extra` column that such information could be added to.

rustyconover:
Hi @Mytherin,

I've removed the new bind callback and just added the new columns on the TableFunctionBindInput like you asked.  I've also put the new data in the extra column and removed all of the changes to DESCRIBE.  This made the diff much smaller.

I think it is ready for another review.

Rusty

Mytherin:
Thanks for the updates - I think this is still a bit too involved of a change in a rather central place. There's already a way of getting the underlying `TableCatalogEntry` which has a `ColumnList` (namely through `get_bind_info`) - could that be used here to simplify this PR?

rustyconover:
Hi @Mytherin,

This PR is about returning tags and comments from table functions not tables.  They don't have a `TableCatalogEntry`.

I'm going to split this PR up into smaller pieces that will make things easier to follow and review.

Rusty

PR #25389, opened 2026-09-05
Tidy snapshot handling in the transaction manager

ywelsch:
This is a PR making a couple of cosmetic (but no behavioral) changes on the v2.0-cyanoptera branch in order to allow #23631 (group commit), which is targeting the main branch, to have a smaller diff and be easier to review. As the v2.0-cyanoptera branch is still under active development and will be for a while, it also makes merging fixes back into the main branch easier.

The PR consists of the following three changes:
- It embeds SnapshotView fully into DuckTransaction, leaving a single place where the VisibilityBound is determined by the transaction's start_time. #23631 will change that.
- It factors out two functions UpdateLowestVisibilityBound and SweepCommittedTransactions, which are currently used in a single place, but will be called by multiple places in #23631 (as cleanups can also trigger after fsync). It also creates a new QueueCleanup method that is called in two places already.
- It removes lowest_active_id, which wasn't really of use. Its only reader, the checkpoint scan in RowGroup::Scan, was never consulting it (inserts are all rows, deletes are decided by the bound alone, updates are disallowed), so the scan now passes MAX_TRANSACTION_ID there, as its committed-rows branch already did.

### DIFF
--- a/src/include/duckdb/transaction/duck_transaction.hpp
+++ b/src/include/duckdb/transaction/duck_transaction.hpp
@@ -36,13 +36,13 @@ struct CommitInfo {
 class DuckTransaction : public Transaction {
 public:
 	DuckTransaction(DuckTransactionManager &manager, ClientContext &context, transaction_t start_time,
-	                transaction_t transaction_id, idx_t catalog_version);
+	                SnapshotView view, idx_t catalog_version);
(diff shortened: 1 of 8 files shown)


Mytherin:
Thanks!

PR #22062, opened 2026-04-14
Rewrite GLOB matcher from recursive matching to iterative one

IfkumRfnl:
Fixes #22050

GLOB matcher used naive recursion before, it was rewritten to use iterative approach, also some tests were added for couple corner cases. Also fixes the recursive calls not forwarding the `allow_question_mark`.

Before this query:
```
SELECT repeat('a', 35) || 'c' GLOB '*a*a*a*a*a*a*a*a*a*a*a*a*a*a*b';
```
Was taking 30+seconds to complete.

With the new matcher:
```
Run Time (s): real 0.001 user 0.000677 sys 0.000000
```

PR #19521, opened 2025-10-28
[ART] Iterative Iterator

artjomPlaunov:
Make the ART iterator iterative. It was already tail recursive, so this is just a simple transformation into a loop. @taniabogatsch

PR #25390, opened 2026-09-05
Increase stack size on Linux with musl libc

staticlibs:
I was investigating a musl-specific test failure, that appeared to be a [stack overflow](https://github.com/duckdb/ducklake/pull/1430#issuecomment-5550167044).

Default stack size with musl libc is 128KB, while on glibc it is 8MB.

This PR increases default stack size for musl libc to 8MB to have it closer in behaviour to glibc.

Testing: new musl-specific test is added that tries to allocate 7MB on stack.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,6 +247,7 @@ endif()
 
 if (MUSL_ENABLED)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__MUSL_ENABLED__")
+  add_link_options(-Wl,-z,stack-size=8388608)
 endif()
 
 set(M32_FLAG "")
--- a/test/common/CMakeLists.txt
+++ b/test/common/CMakeLists.txt
@@ -33,7 +33,8 @@ add_library_unity(
   test_utf8_codepoint_oob.cpp
   test_temporary_memory_manager.cpp
   test_versioning_utils.cpp
-  test_value_to_sql_string.cpp)
+  test_value_to_sql_string.cpp
+  test_musl_stack_size.cpp)
 
 set(UNITTEST_OBJECT_FILES
     ${UNITTEST_OBJECT_FILES} $<TARGET_OBJECTS:test_common>
--- a/test/common/test_musl_stack_size.cpp
+++ b/test/common/test_musl_stack_size.cpp
@@ -1,0 +1,21 @@
+#include "catch.hpp"
+
+#include <array>
+#include <thread>
+
+TEST_CASE("Check that 7MB can be allocated on stack with musl libc", "[musl_stack_size]") {
+#ifndef __MUSL_ENABLED__
+	return;
+#endif
+
+	std::thread th([] {
+		std::array<char, 7 * (1 << 20)> arr;
+		volatile char *ptr = arr.data();
+
+		for (size_t i = 0; i < arr.size(); i += (1 << 10)) {
+			ptr[i] = 42;
+		}
+	});
+
+	th.join();
+}


smvv:
@staticlibs i made a [PR](https://github.com/duckdb/duckdb/pull/25010) to detect [these](https://github.com/duckdblabs/duckdb-internal/issues/10522) stack overflows in `linux release tests` under glibc by reducing the worker thread's stack size (similar size as musl).

We can increase the worker thread's stack size with the linker flag, but that would increase duckdb's memory usage for all musl builds, right?

I think it's useful to know why we need a >128KiB stack for a worker thread. For example, the PEG parser used the stack instead of allocating on the heap. I think we can [use](https://nickdesaulniers.github.io/blog/2023/01/20/debugging-wframe-larger-than/) `-Wframe-larger-than=1024` to find functions that allocate a lot on the stack (but it does not help with recursion-heavy code). I'll open a PR to see if that flag is useful.

staticlibs:
Hi Sander, thanks for your comments!

> @staticlibs i made a PR to detect these stack overflows in linux release tests under glibc by reducing the worker thread's stack size (similar size as musl).

I think the linked PR is really only tangentially related to this one (by touching the same options). The intention of this PR is to "have it [musl] closer in behaviour to glibc", thus making the playing field level for both glibc and musl.

> We can increase the worker thread's stack size with the linker flag, but that would increase duckdb's memory usage for all musl builds, right?

No, the Resident Set Size is expected to stay the same.

> I think it's useful to know why we need a >128KiB stack for a worker thread. For example, the PEG parser used the stack instead of allocating on the heap. I think we can use -Wframe-larger-than=1024 to find functions that allocate a lot on the stack (but it does not help with recursion-heavy code).

See recursion details in the stack trace linked in OP. It may or may not be useful to change these functions to use less stack space. Just I think that making them behave the same on both glibc and musl is useful.

hannes (approved):

smvv:
> Resident Set Size is expected to stay the same.

Agree! By default, RSS stays the same because it only increases virtual address space. But, when the stack usage of the thread increases, it uses more memory for the lifetime of the thread.

The linked PR simulates musl's smaller worker thread stack in CI so we catch large stack usage under glibc builds (and we avoid breaking only in musl CI jobs).

For example, the failing test uses recursion:
```
    #41 0x7f2a9467961a in duckdb::JoinOrderOptimizer::Optimize duckdb/src/optimizer/join_order/join_order_optimizer.cpp:74:41
    #42 0x7f2a9469fbb7 in duckdb::RelationManager::ExtractJoinRelations duckdb/src/optimizer/join_order/relation_manager.cpp:576:37
    #43 0x7f2a9467b34d in duckdb::QueryGraphManager::Build duckdb/src/optimizer/join_order/query_graph_manager.cpp:173:38
```
and i'm not sure if we should continue to do recursion like that. Recently, I've noticed we're changing more recursive code with iterative alternatives [[1](https://github.com/duckdb/duckdb/pull/25419), [2](https://github.com/duckdb/duckdb/pull/25230), [3](https://github.com/duckdb/duckdb/pull/25198), [4](https://github.com/duckdb/duckdb/pull/19521), [5](https://github.com/duckdb/duckdb/pull/22062)].

> same on both glibc and musl is useful.

Agree, the linked PR does that by reducing the stack space on glibc.
I'm worried that if we increase the stack space for all builds, we don't find the code that uses excessive recursion/large stack frames. How would we deal with that?

@Mytherin do we want to support a small stack size in worker threads, or increase the stack size for worker threads in all builds?

staticlibs:
Sander,

>> same on both glibc and musl is useful.

> the linked PR does that by reducing the stack space on glibc.

Unless I am mistaken, the linked PR #25010 is only doing this for test runs on CI. Thus the idea, that it does not conflict with this PR (that increases musl stack size for all builds) in any way.

I would think we can make musl builds to use the same large stack size as glibc in release builds (in actual use, in extension test runs) and at the same time continue running the CI tests with smaller stack size?

smvv:
ah i see; let's merge this!

Mytherin:
Thanks!

PR #25394, opened 2026-09-06
Fix BETWEEN to follow SQL three-valued logic for NULL bounds

### LINKED ISSUE #25170: Unexpected results of `UNION ALL` over two identical RIGHT JOINs with an `EXISTS` in the ON clause
### What happens?

A `RIGHT JOIN` whose non-equi `ON` predicate contains an `EXISTS` subquery returns 529 rows on its own, but `q UNION ALL q` over the same query returns 46 instead of 1058 (= 529 + 529). `SET disabled_optimizers='common_subplan'` makes the `UNION ALL` return the expected 1058.

Both `CREATE UNIQUE INDEX` statements and the `PRIMARY KEY` are load-bearing: dropping any one returns the expected result. The insertion pattern is also load-bearing. The 23 rows must be loaded using the 12 autocommitted `INSERT` statements shown below. Using only 22 rows, reducing the number of statements to 11, inserting all 23 rows in one statement, loading them as `1 x 12 + 11 x 1`, or wrapping the shown inserts in a single `BEGIN`/`COMMIT` all return the expected result, the `UNION ALL` matching twice the single query.

Bisected to `7becbbfed86ff04946fb657ba8eaa4303db2b579` (PR #24084, "Add batch index support to PhysicalColumnDataScan", landed 2026-07-30); verified present at the culprit and absent at its first parent.

#24624 is another `common_subplan` wrong-result involving `UNION ALL`, but its repro returns the expected result on this build while this one still fails.

### To Reproduce

```sql
CREATE TABLE t0(c0 BOOLEAN, c1 INT, c2 INT, PRIMARY KEY(c1));

CREATE UNIQUE INDEX i0 ON t0 (c1);
CREATE UNIQUE INDEX i1 ON t0 (c1);

INSERT INTO t0(c1, c0) VALUES (1, false), (2, false);
INSERT INTO t0(c1, c0) VALUES (3, false), (4, false);
INSERT INTO t0(c1, c0) VALUES (5, false), (6, false);
INSERT INTO t0(c1, c0) VALUES (7, false), (8, false);
INSERT INTO t0(c1, c0) VALUES (9, false), (10, false);
INSERT INTO t0(c1, c0) VALUES (11, false), (12, false);
INSERT INTO t0(c1, c0) VALUES (13, false), (14, false);
INSERT INTO t0(c1, c0) VALUES (15, false), (16, false);
INSERT INTO t0(c1, c0) VALUES (17, false), (18, false);
INSERT INTO t0(c1, c0) VALUES (19, false), (20, false);
INSERT INTO t0(c1, c0) VALUES (21, false), (22, false);
INSERT INTO t0(c1, c0, c2) VALUES (23, false, -1);

-- q =
-- SELECT 1 FROM t0 RIGHT JOIN (SELECT 1 AS col0 FROM t0) AS sub0
--   ON ((EXISTS (SELECT 1 FROM t0 WHERE t0.c2 < 0))
--       NOT BETWEEN (t0.c2 = t0.c1) AND t0.c0)

SELECT count(*) FROM (
  SELECT 1 FROM t0 RIGHT JOIN (SELECT 1 AS col0 FROM t0) AS sub0
    ON ((EXISTS (SELECT 1 FROM t0 WHERE t0.c2 < 0))
        NOT BETWEEN (t0.c2 = t0.c1) AND t0.c0)
) x;
-- 529

SELECT count(*) FROM (
  SELECT 1 FROM t0 RIGHT JOIN (SELECT 1 AS col0 FROM t0) AS sub0
    ON ((EXISTS (SELECT 1 FROM t0 WHERE t0.c2 < 0))
        NOT BETWEEN (t0.c2 = t0.c1) AND t0.c0)
  UNION ALL
  SELECT 1 FROM t0 RIGHT JOIN (SELECT 1 AS col0 FROM t0) AS sub0
    ON ((EXISTS (SELECT 1 FROM t0 WHERE t0.c2 < 0))
        NOT BETWEEN (t0.c2 = t0.c1) AND t0.c0)
) x;
-- Expected: 1058 (= 529 + 529)
-- Actual:   46

SELECT count(*) FROM (
  SELECT 1 FROM t0 RIGHT JOIN (SELECT 1 AS col0 FROM t0) AS sub0
    ON ((EXISTS (SELECT 1 FROM t0 WHERE t0.c2 < 0))
        NOT BETWEEN (t0.c2 = t0.c1) AND t0.c0)
  UNION ALL
  SELECT 1 FROM t0 RIGHT JOIN (SELECT 1 AS col0 FROM t0) AS sub0
    ON ((EXISTS (SELECT 1 FROM t0 WHERE t0.c2 < 0))
        NOT BETWEEN (t0.c2 = t0.c1) AND t0.c0)
) x;
-- 1058
```

### OS:

Ubuntu 22.04

### DuckDB Version:

v1.6.0-dev13791 (8616efa9da)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Suyang Zhong

### Affiliation:

NUS

- [x] Yes, I have

- [x] Yes, I have

Byte-Naut:
## Summary
Fixes #25170.

- `__between` was registered with the default NULL handling, so the expression executor short-circuited to NULL whenever any argument reached it as a `CONSTANT` NULL vector. `BETWEEN` is `input >= lower AND input <= upper`, and Kleene AND can still be FALSE when one side is UNKNOWN and the other is FALSE, so the short-circuit made the result depend on vector representation instead of the logical value:
  ```sql
  SELECT ((EXISTS (SELECT 1)) BETWEEN NULL::BOOLEAN AND false);
  -- NULL (wrong: the upper comparison is FALSE, so the AND is FALSE)
  ```
  This is what #25170 hits: `PhysicalCrossProduct` references the smaller side's rows as `CONSTANT` vectors, so a `RIGHT JOIN` condition with a `BETWEEN ... NULL ...` bound only picks up the short-circuit once the build side happens to be the larger one, e.g. once a shared CTE subplan changes which side of a duplicated `UNION ALL` branch is scanned first. `SetNullHandling(FunctionNullHandling::SPECIAL_HANDLING)` lets `__between`'s own three-valued logic decide the result for every representation.
- `StatisticsPropagator::PropagateBetween` had the same bug one layer up: whenever a comparison could statically be proven `FALSE_OR_NULL`, it folded the whole `BETWEEN` into `constant_or_null(false, input, lower, upper)` regardless of what the other comparison could produce, and fed that unrelated bound into the tuple. Both let a NULL mask a FALSE the other comparison already guaranteed, and let an unrelated NULL bound leak into the result. This PR only folds when the other comparison cannot be FALSE, and only guards on the bound that actually produced `FALSE_OR_NULL`.

## Test plan
- Added `test/sql/join/test_between_null_bound_join.test`, covering:
  - the reduced repro (no UNION ALL / no indexes needed): a `RIGHT JOIN` whose build side is referenced as `CONSTANT` vs `FLAT` vectors by the cross product, with a `BETWEEN` NULL bound in the condition
  - the constant-NULL-bound witness from the issue directly
  - the statistics-propagation FALSE_OR_NULL rewrite with one comparison FALSE and the other UNKNOWN
  - the original issue repro (`UNION ALL` over two identical `RIGHT JOIN`s sharing a common subplan)
- Verified the new test fails on `v2.0-cyanoptera` before this change and passes after
- `test/sql/join/*`, `test/sql/optimizer/*`, `test/optimizer/*`, `test/sql/filter/*`, `test/sql/subquery/*`, `test/sql/cte/*`, `test/sql/setops/*`, `test/sql/function/generic/*`, `test/sql/projection/*`, `test/sql/types/null/*` all pass
- Full `unittest` run: only unrelated pre-existing failure in a `[!mayfail]` timing-based C API test

### DIFF
--- a/src/function/scalar/comparison/between.cpp
+++ b/src/function/scalar/comparison/between.cpp
@@ -214,6 +214,7 @@ unique_ptr<FunctionData> BetweenFunctionDeserialize(Deserializer &deserializer, 
 ScalarFunction BetweenFun::GetFunction() {
 	ScalarFunction between_fun("__between", {LogicalType::ANY, LogicalType::ANY, LogicalType::ANY},
 	                           LogicalType::BOOLEAN, BetweenFunction, BindBetweenFun);
+	between_fun.SetNullHandling(FunctionNullHandling::SPECIAL_HANDLING);
 	between_fun.SetToStringCallback(BetweenToString);
 	between_fun.SetGetExpressionTypeCallback(BetweenGetExpressionType);
 	between_fun.SetLegacySerializeCallback(BetweenLegacySerializeCallback);
--- a/src/optimizer/statistics/expression/propagate_between.cpp
+++ b/src/optimizer/statistics/expression/propagate_between.cpp
@@ -39,12 +39,17 @@ unique_ptr<BaseStatistics> StatisticsPropagator::PropagateBetween(BoundFunctionE
 		expr_ptr = make_uniq<BoundConstantExpression>(Value::BOOLEAN(false));
 	} else if (lower_prune == FilterPropagateResult::FILTER_FALSE_OR_NULL ||
 	           upper_prune == FilterPropagateResult::FILTER_FALSE_OR_NULL) {
-		// either one of the filters is false or null: replace with a constant or null (false)
-		vector<unique_ptr<Expression>> children;
-		children.push_back(std::move(input));
-		children.push_back(std::move(lower_bound));
-		children.push_back(std::move(upper_bound));
-		expr_ptr = ExpressionRewriter::ConstantOrNull(std::move(children), Value::BOOLEAN(false));
+		// one comparison is false or null: the result is false or null only if the other comparison cannot be false
+		// the result is null only when the input or the bound of the false-or-null comparison is null
+		auto &false_bound = lower_prune == FilterPropagateResult::FILTER_FALSE_OR_NULL ? lower_bound : upper_bound;
+		auto other_prune = lower_prune == FilterPropagateResult::FILTER_FALSE_OR_NULL ? upper_prune : lower_prune;
+		if (other_prune == FilterPropagateResult::FILTER_ALWAYS_TRUE ||
+		    other_prune == FilterPropagateResult::FILTER_TRUE_OR_NULL) {
+			vector<unique_ptr<Expression>> children;
+			children.push_back(std::move(input));
+			children.push_back(std::move(false_bound));
+			expr_ptr = ExpressionRewriter::ConstantOrNull(std::move(children), Value::BOOLEAN(false));
+		}
 	} else if (lower_prune == FilterPropagateResult::FILTER_TRUE_OR_NULL &&
 	           upper_prune == FilterPropagateResult::FILTER_TRUE_OR_NULL) {
 		// both filters are true or null: replace with a true or null
--- a/test/sql/join/test_between_null_bound_join.test
+++ b/test/sql/join/test_between_null_bound_join.test
@@ -1,0 +1,93 @@
+# name: test/sql/join/test_between_null_bound_join.test
+# description: BETWEEN with a NULL bound must follow SQL three-valued logic regardless of vector representation
+# group: [join]
+
+# https://github.com/duckdb/duckdb/issues/25170
+
+statement ok
+CREATE TABLE t0(c0 BOOLEAN, c1 INT, c2 INT);
+
+statement ok
+INSERT INTO t0 SELECT false, r, NULL FROM range(1, 23) t(r);
+
+statement ok
+INSERT INTO t0 VALUES (false, 23, -1);
+
+# the RHS is larger than the LHS, so the LHS rows are fed to the condition as constant vectors
+query I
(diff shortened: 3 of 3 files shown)


Mytherin:
Thanks!

Byte-Naut:
Hi Mytherin,

The Thread Sanitizer job failed on this run. Looking at the log, the failure is in an httpfs extension test (`_deps/httpfs_extension_fc-src/test/sql/secrets/create_secret_binding.test:91`, a `DROP SECRET` statement) which exhausted all retries, plus two TSan warnings (a MIN/MAX string aggregate signal-unsafe notice and an `ArenaAllocator::AlignNext` data race) that both recovered on retry. None of these touch the BETWEEN/three-valued-logic code this PR changes, and the new regression test does not appear in the failing job at all.

Thread Sanitizer appears to be flaky on the current branch — I checked recent `Main` runs and it has failed on several unrelated PRs (e.g. on the FTS `test_boolean_query.test`, an inet extension test, and an interrupt/appender timeout), always on different tests unrelated to the changes on those PRs.

Could you re-run the failed job? If it passes, no code change should be needed for this PR.

Byte-Naut:
@Mytherin  , could you please re-run the failed TSan job when you have a moment? Thanks!

PR #25397, opened 2026-09-06
Fix primary key update to NOT NULL

### LINKED ISSUE #25396: Primary key allow update to NOT NULL
### What happens?

Hi team, I found a weird situation that:
- We allow users to update DuckDB primary key to NOT NULL, the update statement will succeed
- To make things worse, a successful update statement doesn't really drop the constraint; later if we insert a NULL to primary key column it still fails

### To Reproduce

```sql
memory D CREATE TABLE pk_col(i INTEGER PRIMARY KEY, j INTEGER);
memory D ALTER TABLE pk_col ALTER COLUMN i DROP NOT NULL;
memory D INSERT INTO pk_col VALUES (NULL, 1);
Constraint Error:
NOT NULL constraint failed: pk_col.i
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25396

I checked the postgres behavior: it doesn't allow primary key to be not null.
```sql
CREATE TABLE pk_col (
    i INTEGER PRIMARY KEY,
    j INTEGER
);
ALTER TABLE pk_col ALTER COLUMN i DROP NOT NULL;
```
gets error message
```sh
CREATE TABLE
psql:commands.sql:6: ERROR:  column "i" is in a primary key
```
In this PR, I stick to PG's behavior to disable certain behavior.

### DIFF
--- a/src/catalog/catalog_entry/duck_table_entry.cpp
+++ b/src/catalog/catalog_entry/duck_table_entry.cpp
@@ -1150,6 +1150,14 @@ unique_ptr<CatalogEntry> DuckTableEntry::SetNotNull(ClientContext &context, SetN
 
 unique_ptr<CatalogEntry> DuckTableEntry::DropNotNull(ClientContext &context, DropNotNullInfo &info) {
 	auto not_null_idx = GetColumnIndex(info.column_name);
+	if (const auto pk = GetPrimaryKey()) {
+		auto &unique = pk->Cast<UniqueConstraint>();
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25398, opened 2026-09-06
Speed up the PEG parser

shreeve:
# Speed up the PEG parser

This PR optimizes the hand-written PEG matcher in `src/parser/peg` without changing what it accepts, the parse results it produces, or the error positions and autocomplete suggestions it reports. Both the heap-based matcher (the default on this branch) and the recursive matcher are updated in lockstep.

## Results

Measured with the new `benchmark/micro/parser.cpp` micro-benchmarks against the `v2.0-cyanoptera` tip (release build, Apple Silicon, `heap_based_parser` at its default of `true`, medians of 15 timed runs, before/after binaries interleaved):

| Benchmark | Before | After | Speedup |
|---|---:|---:|---:|
| ParserTPCH (22 queries x 200) | 4.102 s | 0.706 s | 5.8x |
| ParserTPCDS (99 queries x 50) | 11.499 s | 1.958 s | 5.9x |
| ParserExpressions (500 iterations) | 4.282 s | 0.851 s | 5.0x |
| ParserDDL (200 iterations) | 6.535 s | 1.663 s | 3.9x |
| TokenizerTPCDS (99 queries x 200) | 0.258 s | 0.192 s | 1.3x |

Parsing the full TPC-DS query set goes from 2.3 ms to 0.40 ms per query.

Speedup of the same binaries by platform and toolchain (Linux x86_64 is a Core Ultra 9 185H, both builds against libstdc++):

| Benchmark | macOS arm64, Apple clang | Linux, clang 21 | Linux, gcc 15 |
|---|---:|---:|---:|
| ParserTPCH | 5.8x | 3.5x | 3.7x |
| ParserTPCDS | 5.9x | 3.7x | 4.0x |
| ParserExpressions | 5.0x | 3.3x | 3.5x |
| ParserDDL | 3.9x | 2.8x | 3.0x |
| TokenizerTPCDS | 1.3x | 1.3x | 1.3x |

The smaller Linux ratio is explained under "What remains" below.

## Changes

Four commits, 36 files, +884/-192 lines, on top of the `v2.0-cyanoptera` tip:

| Commit | Files | Lines |
|---|---:|---:|
| Add parser micro-benchmarks | 2 | +203 |
| Speed up the PEG matcher | 32 | +598 / -179 |
| Memoize the qualification rules of the PEG grammar | 2 | +10 |
| Reuse frame storage in the heap-based parser | 5 | +74 / -14 |

**Add parser micro-benchmarks** (`benchmark/micro/parser.cpp`). Parse the TPC-H and TPC-DS query sets, a generated expression-heavy query, and a set of DDL statements through `Parser::ParseQuery`, plus a tokenizer-only benchmark. These are what the table above was measured with.

**Speed up the PEG matcher.** Profiling showed the time going to three places: comparing keyword strings at every `KeywordMatcher`, trying alternatives that could never match the current token, and allocating parse results and their child vectors. The commit addresses each:

- *Intern literals.* `MatcherAllocator` owns a `KeywordTable` that assigns every literal in the grammar an id when the matcher is allocated. `TokenIterator` resolves the id of the current token once and caches it, so a keyword match is an integer compare instead of a case-insensitive string compare.
- *First-set pruning.* `MatcherFactory::ComputeFirstSets` computes, for every matcher, the set of literal ids that can start it (a fixpoint over the matcher graph). `ChoiceMatcher`, `OptionalMatcher` and their stack-frame counterparts skip children whose first set does not contain the current token, and `Matcher::MatchParseResult` / `MatchStack::InitializeFrame` check the same thing before doing any work. Matchers that are nullable or can start with a non-literal token are never pruned, and nothing is pruned at the autocomplete cursor, so error positions and suggestions are unchanged.
- *Dense packrat cache.* `ParserPackratCache` is indexed by `(token offset, memoized rule slot)` into a flat vector instead of a hash map keyed by `(rule id, position)`.
- *Arena parse results.* `ParseResultAllocator` places parse results in an `ArenaAllocator` and gives list/repeat matchers a LIFO scratch stack for collecting children, so `ListParseResult`/`RepeatParseResult` no longer own a `vector` each and a discarded speculative match costs only a pointer reset.
- *Keyword categories as a bitmask.* `PEGKeywordHelper::KeywordCategories` returns all categories of a token in one lookup, cached on the token; the identifier matchers mask against it instead of doing one map lookup per category, and the tokenizer's keyword check becomes a single lookup.
- *Small hot-path cleanups.* `TokenIterator` accessors are inline, and `MatchState::added_suggestions` is allocated only when autocomplete actually adds a suggestion.

**Memoize the qualification rules of the PEG grammar.** `CatalogQualification`, `SchemaQualification`, `ReservedSchemaQualification`, `TableQualification` and `ReservedTableQualification` are retried at the same position by every alternative of `ColumnReference`, `FunctionIdentifier` and the qualified table and type names. Adding them to `packrat_memoized_rules` in `grammar_types.yml` removes those retries.

**Reuse frame storage in the heap-based parser.** After the above, profiling the heap-based parser showed about 40% of its time in allocating and freeing a `MatchStackFrame` per non-terminal step plus a `MatchState` per alternative tried by `ChoiceMatchStackFrame`. Frames are pushed and popped strictly LIFO, so `MatchStack` now places them in fixed-size blocks that are reused as frames pop, and the choice frame keeps the state of the alternative it is trying in an `optional` (which is why the `MatchState` and `TokenIterator` copy constructors now take a const reference).

What remains in the profile is mostly the transformer building the statement tree; its `TransformStack` frames could use the same block storage, but that is a separate change. On Linux there is one more matcher cost worth noting: `ParseResult::name` is a `string` copied from the rule for every parse result, and with libstdc++'s 15-byte small-string buffer most rule names heap-allocate (libc++ keeps up to 22 bytes inline, which is why the Apple Silicon speedups are higher). The name duplicates the `rule` pointer the result already carries, but replacing it touches every reader of `name`, so that is also left for a follow-up.

## Testing

- The full unit test suite (`test/run '*'`) passes on the final build on macOS (Apple clang, libc++) and on Linux x86_64 (clang 21, libstdc++), with the autocomplete, tpch, tpcds and json extensions loaded. A handful of memory-limit and out-of-core `test_slow` tests fail under parallel workers on both machines and pass when run alone; `memory_limit_batch_load.test_slow` fails on the Linux box on the baseline build as well.
- `test/sql/peg_parser/*`, `test/sql/parser/*` and `test/sql/function/autocomplete/*` pass with both `heap_based_parser = true` (default) and `heap_based_parser = false`.
- `make format-check` is clean for the changed files, and `make parser-grammar` followed by the formatter leaves no diff, so the generated `matcher_factory.cpp` matches `grammar_types.yml`.
- The full CI matrix ran on my fork for this exact commit (https://github.com/shreeve/duckdb/actions/runs/34007691125). Every build, test and tidy job passed, with three exceptions that I checked individually: the Thread Sanitizer job reports a race in the t-digest arena of `approx_quantile` during `test/sql/show_select/summarize_subquery.test`, which a sanitizer build of the base commit without this change reports at the same rate (10 of 15 runs), so it is pre-existing; the Relassert and No String Inline jobs both hit the 600 s batch timeout in `Interrupted QueryAppender flow`, which passes in the other test jobs of the same run and in repeated local relassert runs; and the two regression benchmark jobs compared against `main` instead of `v2.0-cyanoptera` because the workflow falls back to `main` on a branch push.

## How this was written

I want to be upfront that I used Claude as a pair programmer on this change: it did much of the profiling and drafted most of the code. I did my best to review every line individually and to hold it to the quality this codebase expects: I read the full diff, checked each optimization against the matcher semantics (in particular that first-set pruning cannot change the error position or the autocomplete suggestions), and ran the tests, the format check, the generated-file check and the benchmarks on both the baseline and the patched build, on macOS and Linux. I am aware of the generative AI note in CONTRIBUTING.md; I am submitting this because I have reviewed it to the point where I can answer for every line, and I am happy to rework anything that does not fit.

### DIFF
--- a/benchmark/micro/CMakeLists.txt
+++ b/benchmark/micro/CMakeLists.txt
@@ -7,6 +7,7 @@ add_library(
   cast.cpp
   executor.cpp
   in.cpp
+  parser.cpp
   storage.cpp
   string_serialize.cpp)
 
--- a/benchmark/micro/parser.cpp
+++ b/benchmark/micro/parser.cpp
@@ -1,0 +1,202 @@
+#include "benchmark_runner.hpp"
+#include "duckdb_benchmark_macro.hpp"
+#include "duckdb/common/file_system.hpp"
+#include "duckdb/common/string_util.hpp"
+#include "duckdb/main/client_context.hpp"
+#include "duckdb/parser/parser.hpp"
+
+#include <algorithm>
+
+using namespace duckdb;
+
+namespace {
+
+struct ParserBenchmarkState : public DuckDBBenchmarkState {
+	explicit ParserBenchmarkState(string path) : DuckDBBenchmarkState(std::move(path)) {
+	}
+
+	vector<string> queries;
+};
+
+string ReadFile(FileSystem &fs, const string &path) {
+	auto handle = fs.OpenFile(path, FileFlags::FILE_FLAGS_READ);
+	string result(handle->GetFileSize(), '\0');
+	handle->Read(&result[0], result.size());
+	return result;
+}
+
+// reads every .sql file in the directory, sorted by name so the corpus order is stable across runs
+vector<string> ReadQueryDirectory(const string &directory) {
+	auto fs = FileSystem::CreateLocal();
+	vector<string> files;
+	fs->ListFiles(directory, [&](const string &name, bool is_directory) {
+		if (!is_directory && StringUtil::EndsWith(name, ".sql")) {
+			files.push_back(fs->JoinPath(directory, name));
+		}
+	});
+	if (files.empty()) {
+		throw IOException("Parser benchmark found no .sql files in \"%s\"", directory);
+	}
+	std::sort(files.begin(), files.end());
+	vector<string> queries;
+	for (auto &file : files) {
+		queries.push_back(ReadFile(*fs, file));
+	}
+	return queries;
+}
+
+void ParseCorpus(ParserBenchmarkState &state, idx_t iterations) {
+	// the connection's options carry the compiled grammar, so the grammar is shared across parses
+	auto options = state.conn.context->GetParserOptions();
+	for (idx_t i = 0; i < iterations; i++) {
+		for (auto &query : state.queries) {
+			Parser parser(options);
+			parser.ParseQuery(query);
+		}
+	}
+}
+
+// a single SELECT with a wide expression list, nested parentheses, a long IN list, CASE, chained function calls and
+// window functions
+string BuildExpressionQuery() {
+	string sql = "SELECT ";
+	for (idx_t i = 0; i < 40; i++) {
+		if (i > 0) {
+			sql += ", ";
+		}
+		sql += StringUtil::Format("(a%d + b%d * 3 - c%d / 2) %% 7 AS x%d", i, i, i, i);
+	}
+	sql += ", CASE WHEN a0 > 10 AND b0 < 20 THEN 'low' WHEN a0 BETWEEN 20 AND 30 OR c0 IS NULL THEN 'mid' ELSE "
+	       "'high' END AS bucket";
+	sql += ", ((((((((((a1 + 1) * 2) - 3) / 4) + 5) * 6) - 7) / 8) + 9) * 10) AS nested";
+	sql += ", upper(trim(concat(cast(a2 AS VARCHAR), '-', lower(coalesce(s2, 'none'))))) AS chained";
+	sql += ", row_number() OVER (PARTITION BY a3, b3 ORDER BY c3 DESC ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT "
+	       "ROW) AS rn";
+	sql += ", sum(a4) FILTER (WHERE b4 > 0) OVER (ORDER BY c4 RANGE BETWEEN 10 PRECEDING AND 10 FOLLOWING) AS ws";
+	sql += " FROM t WHERE a0 IN (";
+	for (idx_t i = 0; i < 200; i++) {
+		if (i > 0) {
+			sql += ", ";
+		}
+		sql += std::to_string(i);
+	}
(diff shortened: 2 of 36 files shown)


Tishj:
There is very active work going on in this part of the code, so this is kind of a bad time to be contributing to this part of the project.

See #25363. This same rework is planned for the transformer.

shreeve:
@tishj - Ah! Ok, no worries. :-)

shreeve:
Makes sense. I'll leave this as a draft and rebase it onto #25363 once that lands, then re-measure. FWIW, most of the speedup here doesn't touch how the frames execute (literal interning, first-set pruning, arena-allocated parse results, the dense packrat layout), so it should carry over. If any of it is useful to you before then, feel free to grab it, and the benchmarks in the first commit are self-contained if they help with measuring the rework.

Tishj:
I'm just worried that these optimizations are based on behavior observed from the *default* grammar, and are hardcoded from that.
For example the keywords, the use of MatcherType, etc..

shreeve:
Fair concern, so I checked. Nothing is tied to the default grammar: the keyword table is filled from whatever grammar gets compiled, first sets are computed after the extension overrides are applied, and any matcher type other than keyword/list/choice/optional/repeat is treated as "can start with anything" and never pruned. The grammar-extension tests pass with it. If your rework drops `MatcherType`, that switch just becomes a virtual, which is small. Happy to leave it there until the dust settles.

Dtenwolde:
Hi @shreeve 

Thanks for providing the PR, cool stuff! As @Tishj has mentioned, there's a lot of work going on in this area of the code base, so it's difficult to accept contributions at the moment given the constant merge conflicts. Some of the ideas in this PR were just implemented for instance, and others I have on my short-term todo-list. I hope you won't mind though that I take inspiration from your PR to improve the heap-based parser further :)

shreeve:
Hi @Dtenwolde, please do! That's the best outcome. The benchmarks in the first commit are self-contained if they're useful for measuring as you go, and I'm happy to re-run them against your branches or dig into any of the pieces if a question comes up. I'll leave this PR as a reference for now, but feel free to close it whenever it stops being useful. Thanks to you both for taking the time to look, and for all the work on the parser.

PR #25400, opened 2026-09-06
Fix missing `deleted_index`assignment on local table storage when table schema change

### LINKED ISSUE #25399: Reinsert the deleted unique key in the same transaction fails
### What happens?

Hi team, I found when we delete + re-insert a new unique key inside of a transaction, the re-insertion will fail somehow.

### To Reproduce

```sql
memory D CREATE TABLE t (id INTEGER PRIMARY KEY, val INTEGER);
memory D INSERT INTO t VALUES (1, 10), (2, 20);
memory D BEGIN;
memory D DELETE FROM t WHERE id = 1;
memory D ALTER TABLE t ADD COLUMN extra INTEGER DEFAULT 0;

-- the deleted row should succeed for re-insertion
memory D INSERT INTO t VALUES (1, 30, 0);
Constraint Error:
Duplicate key "id: 1" violates primary key constraint.
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25399

The root cause is
- On schema evolution inside of a transaction, a new `LocalTableStorage` is created to record uncommitted changes in the transaction
- Both appended and deleted table indexes should be inherited from parent, but now we're missing delete index assignment

### DIFF
--- a/src/storage/local_storage.cpp
+++ b/src/storage/local_storage.cpp
@@ -45,6 +45,7 @@ LocalTableStorage::LocalTableStorage(ClientContext &context, DataTable &new_data
 	row_groups->collection = std::move(new_collection);
 
 	append_indexes.Move(parent.append_indexes);
+	delete_indexes.Move(parent.delete_indexes);
 }
 
 LocalTableStorage::LocalTableStorage(DataTable &new_data_table, LocalTableStorage &parent,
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25406, opened 2026-09-06
Merge cyanoptera into main

carlopi:


carlopi:
Unsure if there is anything to do about the failure, but just accepting it / noting that multiple PRs drifted all the same direction past the bar.

carlopi:
@Mytherin: I think CI will keep failing, given there was a regression in some tests. I think this might just need to be merged with failing ci.

Mytherin:
Alright, thanks

PR #25407, opened 2026-09-06
fix: LIST comparisons for lists of different lengths

### LINKED ISSUE #25169: Unexpected results when comparing a LIST column
### What happens?

In a single LIST column, two identical values compared against the same constant return different answers. With rows `[]`, `[-2, -1]`, `[-2, -1]`, the predicate `c > [0]` evaluates to `false`, `false`, `true` — the two `[-2, -1]` rows hold the same value, so whichever answer is right, they cannot differ from each other. The same happens to `list_sort(c) = c`, where two identical `[-1823031693]` rows evaluate to `true` and `false`.

It bisects to `53b8c9c41ed086951f697955979829b70ccfdeaf` (PR #24162, "Internal #5184: Nested Equality", landed 2026-07-27), verified present at the culprit and absent at its parent. The commit is part of the ongoing nested-comparison rework, so this may already be known or already covered by work in flight--please close it if so.

### To Reproduce

```sql
SELECT c, c > [0] AS gt FROM (VALUES ([]::INTEGER[]), ([-2, -1]), ([-2, -1])) v(c);
-- Expected: the two [-2, -1] rows agree
-- Actual:   false, false, true

SELECT c, list_sort(c) = c AS eq FROM (VALUES ([0]), ([0]), ([0]), ([1, NULL, 2]), ([-1823031693]), ([1, NULL, 2]), ([-1823031693])) v(c);
-- Expected: the two [-1823031693] rows agree
-- Actual:   true, true, true, false, true, false, false
```

A filter then splits the identical rows: with a third row holding the same value as the second, one is kept and the other is not, and the complementary filter picks up exactly the one that was dropped.

```sql
CREATE TABLE t0(c0 INTEGER[], c1 INT);
INSERT INTO t0 VALUES ([], 1), ([-2, -1], 2), ([-2, -1], 3);

SELECT c1 FROM t0 WHERE NOT (c0 > [0]) ORDER BY c1;
-- Expected: rows 2 and 3 are both present or both absent
-- Actual:   1, 2

SELECT c1 FROM t0 WHERE (c0 > [0]) ORDER BY c1;
-- Actual:   3
```

### OS:

Ubuntu 22.04

### DuckDB Version:

v1.6.0-dev13791 (8616efa9da)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Suyang Zhong

### Affiliation:

NUS

- [x] Yes, I have

- [x] Yes, I have

NodirBobiev:
Fixes #25169

For Lists, call `NestedScatter `with `active_count `instead of `remaining_count `because slots `active_count`...`remaining_count `hold stale indices and uninitialized bytes

Structs have no such problem because every reamining row is compared for every field.


Mytherin:
Thanks - could you just rewrite the tests to be SQL based instead of using the C++ API?

test/api/test_list_comparison_regressions.cpp
```diff
@@ -0,0 +1,128 @@
+#include "catch.hpp"
```
hawkfish: Please rewrite this as a SQL test.

test/api/CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ set(TEST_API_OBJECTS
     test_windows_unicode_path.cpp
     test_object_cache.cpp
     test_partial_executor.cpp
+    test_list_comparison_regressions.cpp
```
hawkfish: This can be removed.

hawkfish (changes requested):
Thanks for the PR! For future reference, we only add C++ tests for very specific situations like internal API tests. For incorrect query results, just write a SQL test.

test/api/test_list_comparison_regressions.cpp
NodirBobiev: done

test/api/CMakeLists.txt
NodirBobiev: done

NodirBobiev:
Seems like the bug has been fixed. So, closing this PR.

PR #25409, opened 2026-09-06
Track I/O operation counts alongside byte counters

Richard-Durkee:
Adds `io.total_read_operations` / `io.total_write_operations`, tracking the number of read/write operations issued to storage alongside the existing `io.total_bytes_read` / `io.total_bytes_written` counters. An "operation" is one file-system read/write call, so that I/O amplification can be calculated.


Mytherin:
Thanks! Sounds like a good idea to me

PR #25410, opened 2026-09-07
aggregate: widen FirstStringStateBase::alloc_size to idx_t to stop the 2GB integer truncation

### LINKED ISSUE #25269: Bug93: Integer Truncation in `FirstStringStateBase::Assign` Leads to Heap Buffer Overflow with Large Strings
### What happens?

In `src/function/aggregate/distributive/first_last_any.cpp`, `FirstStringStateBase::Assign()` computes the arena allocation size by calling `UnsafeNumericCast<uint32_t>(NextPowerOfTwo(len))`. For any string whose length is in the range `(2^31, 2^32-1]` — which DuckDB's own `string_t::MAX_STRING_SIZE` cap permits — `NextPowerOfTwo` returns `2^32` (a 33-bit value), and the subsequent `uint32_t` cast truncates it to `0`. In a release build this causes `ArenaAllocator::Allocate(0)` to return a zero-length allocation; the immediately following `memcpy(ptr, input.GetData(), len)` then writes multiple gigabytes past the end of the allocation, producing a heap buffer overflow with attacker-controlled length and content. In the debug build the checked-cast path catches the truncation and throws an `INTERNAL Error`, crashing the process. The bug is reachable by any of the `first()`, `last()`, or `any_value()` aggregates on a VARCHAR/BLOB column.

### To Reproduce

The crash is triggered by pure SQL; no external files or configuration are needed.

```sql
SELECT last(x) FROM (SELECT repeat('A', 2147483649) AS x);
```

```sh
echo "SELECT last(x) FROM (SELECT repeat('A', 2147483649) AS x);" | ./build/debug/duckdb
```

```
INTERNAL Error: Information loss on integer cast: value 4294967296 outside of target range [0, 4294967295]

Stack Trace:

/...duckdb(+0x2353ba80) [0x62707a78aa80]
/...duckdb(+0x2353b369) [0x62707a78a369]
/...duckdb(+0x2353acdb) [0x62707a789cdb]
/...duckdb(+0x23545e50) [0x62707a794e50]
/...duckdb(+0x214c7bf5) [0x627078716bf5]
/...duckdb(+0x21551cdf) [0x6270787a0cdf]
/...duckdb(+0x2157b826) [0x6270787ca826]
/...duckdb(+0x21572041) [0x6270787c1041]
/...duckdb(+0x21568e0b) [0x6270787b7e0b]
/...duckdb(+0x29857568) [0x627080aa6568]

EXIT=1
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

AetherAI3:
Fixes #25269.

`FirstStringStateBase::Assign()` stored the arena allocation size in a `uint32_t` after casting `NextPowerOfTwo(len)`. For any input string with `len` in `(2^31, 2^32-1]` — a size DuckDB's own `string_t` cap permits — `NextPowerOfTwo(len)` returns `2^32`, and the outer `UnsafeNumericCast<uint32_t>` truncates that to `0`. Release builds then call `ArenaAllocator::Allocate(0)`, get back a stub, and `memcpy` gigabytes of attacker content past it — an attacker-controlled heap buffer overflow reachable through any `first()`, `last()`, or `any_value()` over a `VARCHAR` or `BLOB` column. Debug builds crash at the checked cast.

Widening `alloc_size` to `idx_t` lets `NextPowerOfTwo`'s `uint64_t` result survive without truncation. The doc comment on the field already notes it's not part of the exported state, so the four extra bytes only touch the aggregate's internal state layout. The `alloc_size >= len` reuse check keeps working because `len` is promoted to `idx_t` at the comparison.

## Verified

`main` at `154c2c8d`, debug build (`make debug` with `-j2` and shell target; the parquet+other loadable-extension link steps OOM at `-j4` on a 15 GB box).

- The reporter's own ASan trace in the issue is the primary evidence for the pre-fix crash; on a debug build the 2 GB scalar string generation dominates wall-time before the aggregate path is even entered, so I didn't independently re-crash it. The truncation itself is arithmetically inevitable for any `len` in that range — `NextPowerOfTwo` returns `2^32`, `uint32_t` cast returns `0`, `memcpy` then writes `len` bytes past the stub.
- Post-fix smoke checks all pass on the same debug build: `last`, `first`, `any_value` over short and 1 MB `VARCHAR` values return the expected lengths; multi-row `first(x ORDER BY i)` and `last(x ORDER BY i)` return the expected boundary rows.

## Notes

Reported and root-caused by @ylwango613; the fix follows directly from the analysis in the issue.

I used an AI coding assistant while working on this. Every hunk was reviewed by a human before commit, the fix and smoke tests were run on real hardware, and the diff is two lines with no logic change beyond the width of `alloc_size`.

### DIFF
--- a/src/function/aggregate/distributive/first_last_any.cpp
+++ b/src/function/aggregate/distributive/first_last_any.cpp
@@ -87,7 +87,7 @@ struct FirstStringStateBase {
 	//! Whether the state has been set (i.e. we have seen a row)
 	bool is_set;
 	//! The size of the arena allocation for a non-inlined string value - not part of the exported state
-	uint32_t alloc_size;
+	idx_t alloc_size;
 
 	void Assign(string_t input, AggregateInputData &input_data) {
 		if (input.IsInlined()) {
@@ -99,7 +99,7 @@ struct FirstStringStateBase {
 			if (alloc_size >= len) {
 				ptr = value.GetDataWriteable();
 			} else {
-				alloc_size = UnsafeNumericCast<uint32_t>(NextPowerOfTwo(len));
+				alloc_size = NextPowerOfTwo(len);
 				ptr = char_ptr_cast(input_data.allocator.Allocate(alloc_size));
 			}
 			memcpy(ptr, input.GetData(), len);


Mytherin:
Thanks - could you rebase on v2.0 so it doesn't pull in the changes from main?

AetherAI3:
@Mytherin Rebased onto v2.0-cyanoptera — new HEAD fa5b686 is a single commit on top of 487155f6ef, just the 2-line widen. Nothing from main pulled in this time.

carlopi (approved):
Thanks!

carlopi:
Could you undraft? Then a maintainers can trigger CI.
I based https://github.com/duckdb/duckdb/pull/25952 on this commit.

PR #25411, opened 2026-09-07
Fix ALTER TABLE ADD COLUMN with NOT NULL constraint

### LINKED ISSUE #25401: Add column with NOT NULL doesn't work
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t(a INTEGER);
memory D INSERT INTO t VALUES (1);
memory D ALTER TABLE t ADD COLUMN b INTEGER NOT NULL;
memory D ALTER TABLE t ADD COLUMN c INTEGER;
memory D ALTER TABLE t ADD COLUMN d INTEGER;
memory D ALTER TABLE t ADD COLUMN e INTEGER;
-- succeeds, which is expected to fail
memory D INSERT INTO t VALUES (2, NULL, 1, 1, -1);
memory D 
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25401

### DIFF
--- a/src/include/duckdb/parser/parsed_data/alter_table_info.hpp
+++ b/src/include/duckdb/parser/parsed_data/alter_table_info.hpp
@@ -181,13 +181,16 @@ private:
 // AddColumnInfo
 //===--------------------------------------------------------------------===//
 struct AddColumnInfo : public AlterTableInfo {
-	AddColumnInfo(const AlterEntryData &data, ColumnDefinition new_column, bool if_column_not_exists);
+	AddColumnInfo(const AlterEntryData &data, ColumnDefinition new_column, bool if_column_not_exists,
+	              bool add_not_null);
 	~AddColumnInfo() override;
 
 	//! New column
 	ColumnDefinition new_column;
 	//! Whether or not an error should be thrown if the column exist
 	bool if_column_not_exists;
+	//! Whether to add a NOT NULL constraint after adding the column
+	bool add_not_null = false;
 
 public:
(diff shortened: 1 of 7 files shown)


lnkuiper:
This succeeds on the base branch but fails in this PR:
```sql
CREATE TABLE t(a INT);
INSERT INTO t VALUES (1);
ALTER TABLE t ADD b INT DEFAULT (1+1) NOT NULL;
```

dentiny:
> This succeeds on the base branch but fails in this PR:

Hi @lnkuiper I think it's the pre-existing issue before my fix, and it's intentional choice (for now)
The reason why it surfaces in this PR, is main branch completely ignores the `NOT NULL` constraint so it's no-op
See https://github.com/duckdb/duckdb/pull/25453#discussion_r3952905318

dentiny:
And you can repro it in a different way in main branch
```sql
memory D INSERT INTO t VALUES (1);
memory D BEGIN;
memory D ALTER TABLE t ADD COLUMN b INT DEFAULT (1 + 1);
memory D ALTER TABLE t ALTER COLUMN b SET NOT NULL;
TransactionContext Error:
Cannot create index with outstanding updates
```

lnkuiper:
Thanks for the explanation, that makes sense.

PR #25412, opened 2026-09-07
Add a constant-pattern fast path for LIKE ... ESCAPE

JoeJRW:
LIKE / NOT LIKE ... ESCAPE has no bind function, so like_escape always runs the generic character-by-character matcher - even when the escape character never occurs in the pattern. For example, LIKE 'a#_b' ESCAPE '#' should take the same  SIMD contains path a plain LIKE does.

This PR binds a LikeMatcher when the pattern and the escape character are both  constant, and resolve escape sequences while building it instead of bailing out on them: an escaped character is matched literally, so it just extends the current constant segment.

Moreover, this PR fixes a bug: the loop consuming trailing '%' wildcards did not check  for the escape character. Under ESCAPE '%', 'abc' LIKE 'abc%%' ESCAPE '%' returned true even though that pattern denotes the constant "abc%", and a pattern ending in a dangling escape was accepted as a match instead of raising the syntax error the main loop raises for it.

### DIFF
--- a/benchmark/micro/string/like_escape_escaped_regular.benchmark
+++ b/benchmark/micro/string/like_escape_escaped_regular.benchmark
@@ -1,0 +1,19 @@
+# name: benchmark/micro/string/like_escape_escaped_regular.benchmark
+# description: Contains word 'regular' in the l_comment (11.5%~), written with an escape sequence in the pattern
+# group: [string]
+
+name Like Escape ('%regula#r%' ESCAPE '#')
+group string
+
+require tpch
+
+cache tpch_sf1.duckdb
+
+load
+CALL dbgen(sf=1);
+
+run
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks - could we maybe split off the bug fix from the performance improvement into a separate PR?

JoeJRW:
@Mytherin Thanks for taking a look — split done, the bug fix is now #25422 and this PR keeps only the fast path.

For now this branch carries #25422's commit at its base so the tests pass. Once that one is merged I will rebase onto main, leaving a single commit here. Happy to restructure this differently if you would rather have it another way.

JoeJRW:
The rebase is complete. One commit is left and ready for review.

PR #25413, opened 2026-09-07
Fix ALTER TABLE ALTER column USING NULL

### LINKED ISSUE #25414: `ALTER COLUMN TYPE ... USING NULL` silently writes NULLs into `NOT NULL` column
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE t(a INTEGER NOT NULL, b INTEGER);
memory D INSERT INTO t VALUES (1, 10), (2, 20);
memory D ALTER TABLE t ALTER a SET DATA TYPE INTEGER USING NULL;
memory D SELECT a, b FROM t ORDER BY a;
┌───────┬───────┐
│   a   │   b   │
│ int32 │ int32 │
├───────┼───────┤
│  NULL │    10 │
│  NULL │    20 │
└───────┴───────┘
```
I tried the query on postgres as well
```sql
CREATE TABLE t (
    a INTEGER NOT NULL,
    b INTEGER
);

INSERT INTO t VALUES (1, 10), (2, 20);

ALTER TABLE t
    ALTER COLUMN a TYPE INTEGER
    USING NULL::INTEGER;
```
It errors out with the following error message
```sh
CREATE TABLE
INSERT 0 2
psql:commands.sql:10: ERROR:  column "a" of relation "t" contains null values
```

### OS:

macos

### DuckDB Version:

6d7abb1090e1079f252b15ba0366310bb87d8582

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25414

The issue is
- When we change column type, `NOT NULL` constraint is [ignored](https://github.com/duckdb/duckdb/blob/v2.0-cyanoptera/src/catalog/catalog_entry/duck_table_entry.cpp#L1237-L1238), no action taken
- When we create a new table, after type change has been applied, we [never check whether constraint is respected or not](https://github.com/duckdb/duckdb/blob/e3946f2327a3cc622e1ec7fe71d51de49f93e61d/src/storage/data_table.cpp#L181-L226)

FAQ:
Q: Why only `NOT NULL` constraint suffer the issue?
A: Because other constraints reject column type change [upfront](https://github.com/duckdb/duckdb/blob/e3946f2327a3cc622e1ec7fe71d51de49f93e61d/src/catalog/catalog_entry/duck_table_entry.cpp#L1229-L1260)


lnkuiper:
This succeeds on the base branch but throws in this PR:
```sql
CREATE TABLE t(a INT NOT NULL);
INSERT INTO t VALUES (1), (2);
BEGIN;
DELETE FROM t WHERE a = 1;
ALTER TABLE t ALTER a TYPE BIGINT
  USING CASE WHEN a = 1 THEN NULL ELSE a END;
COMMIT;
```

dentiny:
> This succeeds on the base branch but throws in this PR:

I believe the failure is not caused by this PR, the linked merged PR has already resolved the issue.
```sql
memory D CREATE TABLE t(a INT NOT NULL);
memory D INSERT INTO t VALUES (1), (2);
memory D BEGIN;
memory D DELETE FROM t WHERE a = 1;
memory D ALTER TABLE t ALTER a TYPE BIGINT
           USING CASE WHEN a = 1 THEN NULL ELSE a END;
memory D COMMIT;
```

PR #25418, opened 2026-09-07
Fix task executor regression

### LINKED ISSUE #25282: v2 C API: compute-heavy streamed query intermittently ~20x slower (wall 27s vs 1.2s), mostly asleep in Executor::WaitForTask
## Summary

Driving `SELECT count(DISTINCT i) FROM range(100000000) t(i)` through the v2
C API (`statement_execute` → `result_fetch_chunk`), roughly 1 in 3 runs takes
~27s instead of ~1.2s. The slow runs are not computing: ~7s user CPU over 27s
wall (~26% of one core), versus ~3.4s user over 1.2s wall (~290%) on fast
runs. Whether a run is fast or slow is decided at query start and sticks for
the whole query.

Build: `main @ 1582849bf9` (v2.0.0-dev83323), macOS arm64 self-built
`GEN=ninja make release`; also observed as intermittent ~45s spikes for a
200M-row variant on linux-amd64 (22-core), same pin.

## Reproducer (pure C, no wrapper code)

`heavy.c` below: env → open(:memory:) → connect → parse → execute →
fetch_chunk loop, three runs on one connection, per-run getrusage. Typical
output (three trials — note the slow run moves around):

    run 1: wall 1.303s  user 3.368s  sys 0.178s
    run 2: wall 1.217s  user 3.614s  sys 0.086s
    run 3: wall 27.961s  user 7.065s  sys 1.149s

    run 1: wall 1.176s  user 3.471s
    run 2: wall 27.531s  user 6.623s
    run 3: wall 1.086s  user 3.280s

Re-run just before filing, same binary, same pin — 2 of 3 slow this time:

    run 1: wall 1.670s  user 3.751s  sys 0.657s (1 chunks)
    run 2: wall 23.654s  user 4.127s  sys 0.225s (1 chunks)
    run 3: wall 23.607s  user 4.073s  sys 0.128s (1 chunks)

Fresh-connection-per-run shows the same rate, so the state is per-run, not
per-connection. The CLI (v1 materialized path) on the same build is stable at
~1.07s across runs.

<details>
<summary>heavy.c</summary>

```c
// heavy.c: does the v2 API's streaming path degrade after the first heavy
// query? Runs count(DISTINCT) over range(100M) three times on one
// connection, timing each, draining via result_fetch_chunk.
#include <stdio.h>
#include <string.h>
#include <sys/resource.h>
#include <time.h>
#include "duckdb_v2.h"

static void die(const char *where, DUCKDB_V2_ERROR code, duckdb_v2_error_info_handle err) {
	duckdb_v2_str text = {0};
	if (err) {
		duckdb_v2_error_info_get_text(err, &text);
	}
	fprintf(stderr, "FAIL %s: code=%d %.*s\n", where, (int)code, (int)text.len, text.ptr ? text.ptr : "");
	duckdb_v2_error_info_destroy(&err);
	__builtin_trap();
}

#define TRY(where, call)                                                                                               \
	do {                                                                                                               \
		duckdb_v2_error_info_handle err = NULL;                                                                        \
		DUCKDB_V2_ERROR rc = (call);                                                                                   \
		if (rc != DUCKDB_V2_ERROR_NONE)                                                                                \
			die(where, rc, err);                                                                                       \
	} while (0)

static double now(void) {
	struct timespec ts;
	clock_gettime(CLOCK_MONOTONIC, &ts);
	return ts.tv_sec + ts.tv_nsec / 1e9;
}

static const char *SQL = "SELECT count(DISTINCT i) FROM range(100000000) t(i)";

int main(int argc, char **argv) {
	duckdb_v2_environment_handle env = NULL;
	TRY("create_environment", duckdb_v2_create_environment(&env, &err));

	duckdb_v2_str memory = {NULL, 0};
	duckdb_v2_database_handle db = NULL;
	TRY("open", duckdb_v2_open(env, memory, NULL, 0, &db, &err));

	// With any argument: fresh connection per run — splits per-connection
	// state from per-database state.
	int reconnect = argc > 1;
	duckdb_v2_connection_handle conn = NULL;
	TRY("connect", duckdb_v2_connect(db, &conn, &err));

	for (int run = 1; run <= 3; run++) {
		if (reconnect && run > 1) {
			duckdb_v2_disconnect(&conn);
			TRY("reconnect", duckdb_v2_connect(db, &conn, &err));
		}
		duckdb_v2_statement_iterator_handle iter = NULL;
		TRY("parse_sql", duckdb_v2_parse_sql(conn, SQL, &iter, &err));
		duckdb_v2_sql_statement_handle stmt = NULL;
		TRY("iterator_next", duckdb_v2_statement_iterator_next(iter, &stmt, &err));

		struct rusage r0, r1;
		getrusage(RUSAGE_SELF, &r0);
		double start = now();
		duckdb_v2_result_handle result = NULL;
		TRY("statement_execute", duckdb_v2_statement_execute(conn, stmt, NULL, NULL, 0, &result, &err));

		long chunks = 0;
		for (;;) {
			duckdb_v2_data_chunk_handle chunk = NULL;
			TRY("fetch_chunk", duckdb_v2_result_fetch_chunk(result, &chunk, &err));
			if (!chunk)
				break;
			chunks++;
			duckdb_v2_data_chunk_destroy(&chunk);
		}
		double wall = now() - start;
		getrusage(RUSAGE_SELF, &r1);
		double user = (r1.ru_utime.tv_sec - r0.ru_utime.tv_sec) + (r1.ru_utime.tv_usec - r0.ru_utime.tv_usec) / 1e6;
		double sys = (r1.ru_stime.tv_sec - r0.ru_stime.tv_sec) + (r1.ru_stime.tv_usec - r0.ru_stime.tv_usec) / 1e6;
		printf("run %d: wall %.3fs  user %.3fs  sys %.3fs (%ld chunks)\n", run, wall, user, sys, chunks);
		duckdb_v2_result_destroy(&result);
		duckdb_v2_sql_statement_destroy(&stmt);
		duckdb_v2_statement_iterator_destroy(&iter);
	}

	duckdb_v2_disconnect(&conn);
	duckdb_v2_close(&db);
	duckdb_v2_destroy_environment(&env);
	printf("PASS\n");
	return 0;
}
```

</details>

## What it looks like from outside

The wall/user ratio of slow runs fits a nap-per-task pattern:
`Executor::WaitForTask` (executor.cpp) waits on `task_reschedule` bounded by
`WAIT_TIME` = 20ms; ~100M rows / 122880-row morsels ≈ 814 tasks × 20ms ≈ 16s
of sleep, plus the doubled user time, lands right at the observed 27s. It
looks like the fetching thread loses the wake race once at the start of a run
and then pays the full 20ms timeout per task for the rest of the query.

Two refinements from further testing:

- The ~26% CPU of a slow run (vs ~290% fast) says the worker pool itself is
  starved, not just the fetching thread napping: if workers were draining
  the pipeline normally the query would finish in ~1s regardless of how the
  caller waits. It looks like tasks sit unclaimed while workers sleep —
  i.e., a lost wakeup on the scheduler side — and the query trickles
  through the caller's bounded 20ms pulls.
- Prior activity on the database worsens the odds sharply. A fresh process
  running this query as its first statement is fast ~8 of 8 tries; the same
  query issued after ~50 short statements and some connection churn (our
  test suite's natural preamble) hits the slow mode roughly every other
  run. Driving the result via result_step for a few ms before falling back
  to result_fetch_chunk helps the fresh case but not the churned one.

## Related observations from the same integration work

1. `WaitForTask`'s wake conditions (task completion / reschedule) do not
   include chunk-became-ready, so a consumer of a *streaming* result that
   outruns the producer eats bounded naps between chunks even in the fast
   mode. A chunk-ready signal would fix both consumers.
2. A result driven purely by `result_step` executes essentially serially
   (~17x slower on the aggregation above than `result_fetch_chunk`'s good
   mode). If step is meant to only ever run one caller-thread work unit,
   that is worth a line in the spec text; if worker threads are meant to
   participate, they do not appear to.
3. Workaround we ship (FWIW): a dedicated fetch thread that steps with a
   ~5ms time budget per chunk before falling back to `result_fetch_chunk`.
   Oddly, this dodges the slow mode entirely (7/7 runs ~0.97s, tighter than
   the API's own fast mode) — consistent with the caller-thread stepping
   keeping the executor primed past whatever race the pure blocking fetch
   loses.

This looks related in symptom to #12105 (duckdb_stream_fetch_chunk slow,
closed not-planned), but with the v2 API, a deterministic reproducer, and a
scheduler-shaped explanation.

Happy to re-run any variant, this reproduces in under a minute here.

Maxxen:
Fixes https://github.com/duckdb/duckdb/issues/25282

Basically, don't just check if there are new tasks, also check if we  _currently_ hold a partially processed task.

Also fix the `blocked_thread_time` to increment by the actual time waited, not just a fixed `WAIT_TIME_NS` even when the thread wakes up earlier.

### DIFF
--- a/src/include/duckdb/execution/executor.hpp
+++ b/src/include/duckdb/execution/executor.hpp
@@ -187,7 +187,7 @@ private:
 	//! Currently alive executor tasks
 	atomic<idx_t> executor_tasks;
 
-	//! Total time blocked while waiting on tasks. In ticks. One tick corresponds to WAIT_TIME.
+	//! Total time blocked while waiting on tasks, in microseconds
 	atomic<idx_t> blocked_thread_time;
 };
 } // namespace duckdb
--- a/src/include/duckdb/execution/executor.hpp
+++ b/src/include/duckdb/execution/executor.hpp
@@ -55,6 +55,11 @@ public:
 	void Initialize(unique_ptr<PhysicalOperator> physical_plan);
 
 	void CancelTasks();
+	//! Whether the thread driving ExecuteTask holds a partially processed task.
+	//! `task` is owned by that thread alone, so only it may call this.
+	bool HasTaskInProgress() const {
+		return task != nullptr;
+	}
 	PendingExecutionResult ExecuteTask(bool dry_run = false);
 	void WaitForTask();
 	void SignalTaskRescheduled(lock_guard<mutex> &);
--- a/src/main/client_context.cpp
+++ b/src/main/client_context.cpp
@@ -686,7 +686,12 @@ ClientContext::PendingPreparedStatementInternal(ClientContextLock &lock,
 }
 
 void ClientContext::WaitForTask(ClientContextLock &lock, BaseQueryResult &result) {
-	active_query->executor->WaitForTask();
+	auto &executor = *active_query->executor;
+	if (executor.HasTaskInProgress()) {
+		// This thread is holding a partially processed task, the next step resumes it without waiting.
+		return;
+	}
+	executor.WaitForTask();
 }
 
 bool ClientContext::ErrorInvalidatesTransaction(ExceptionType type) {
--- a/src/parallel/executor.cpp
+++ b/src/parallel/executor.cpp
@@ -388,8 +388,10 @@ void Executor::WaitForTask() {
 	}
 	// Nothing to run on this thread, all remaining tasks are either running on other threads or descheduled.
 	// Wait (bounded), but wake up on task completion or reschedule.
-	blocked_thread_time += blocked_micros + WAIT_TIME_MS.count();
+	const auto wait_begin = TimePoint::Tick();
 	task_reschedule.wait_for(l, WAIT_TIME_MS);
+	const auto wait_micros = NumericCast<idx_t>(TimePoint::ElapsedMicros(wait_begin, TimePoint::Tick()));
+	blocked_thread_time += blocked_micros + wait_micros;
 #endif
 }
 
--- a/src/parallel/executor.cpp
+++ b/src/parallel/executor.cpp
@@ -376,11 +376,6 @@ void Executor::WaitForTask() {
 		blocked_thread_time += blocked_micros;
 		return;
 	}
-	if (task) {
-		// This thread is holding a partially processed task, the next step will make progress without waiting
-		blocked_thread_time += blocked_micros;
-		return;
-	}
 	if (TaskScheduler::GetScheduler(context).GetTaskCountForProducer(*producer) > 0) {
 		// A new task is available for the calling thread, the next step will make progress without waiting
 		blocked_thread_time += blocked_micros;
--- a/src/parallel/executor.cpp
+++ b/src/parallel/executor.cpp
@@ -376,9 +376,13 @@ void Executor::WaitForTask() {
 		blocked_thread_time += blocked_micros;
 		return;
 	}
-	auto &scheduler = TaskScheduler::GetScheduler(context);
-	if (scheduler.GetTaskCountForProducer(*producer) > 0) {
-		// A task is available for the calling thread, the next step will make progress without waiting
+	if (task) {
+		// This thread is holding a partially processed task, the next step will make progress without waiting
+		blocked_thread_time += blocked_micros;
+		return;
+	}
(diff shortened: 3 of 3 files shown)


Mytherin:
Thanks!

PR #25421, opened 2026-09-07
`ArenaAllocator` for `MatchFrames`

Dtenwolde:
More optimizations for the heap-based `Matcher`. We now store stack frames in reusable `ArenaAllocator` segments instead of allocating each frame separately. Each segment can hold 64 frames. I tested a range of 16, 32, 64 and 128 and found 64 to be marginally faster. We also inline 2 segments, which seems enough for most queries. 

There's followup needed to make this work with `MatcherProcess`, which I'll work on next.


src/include/duckdb/parser/peg/matcher_stack.hpp
```diff
@@ -44,17 +46,36 @@ struct MatchStackFrame {
...
+	void SetActiveFrameSegment(idx_t segment_index);
+	data_ptr_t AllocateFrameSlot();
+	void DestroyTopFrame();
 	void PushFrame(MatchInput input);
 	void InitializeFrame(MatchStackFrame &frame);
 	void ExecuteFrame(MatchStackFrame &frame);
 	MatcherResult FinalizeFrame(MatchStackFrame &frame);
 
 private:
-	vector<unique_ptr<MatchStackFrame>> frames;
```
lnkuiper, line 57: Wouldn't `vector<MatchStackFrame>` also fix the allocation issue? Is there any reason we need the segments?

Dtenwolde: Yes I think after the `MatchProcess` got added this makes more sense and also simplifies it. Going to try it out

Dtenwolde: Performance remained about the same, so going with `vector<MatchStackFrame>`. Thanks for the suggestion :)

PR #25422, opened 2026-09-07
Fix trailing '%' wildcards ignoring the ESCAPE character

JoeJRW:
The loop that consumes trailing '%' wildcards at the end of a LIKE pattern did not check whether a '%' was escaped, so it treated an escaped one as a wildcard. Under ESCAPE '%', 'abc' LIKE 'abc%%' ESCAPE '%' returned true even though that pattern denotes the constant string "abc%".

A pattern ending in a dangling escape character is invalid, and the same loop determined whether that got reported. The main matching loop raises a syntax error for such a pattern, but only reaches the escape while the string still has characters left to match; once the string runs out, control falls through to this loop, which stopped at the escape and reported "no match" instead. The error therefore depended on the data: 'abcd' LIKE 'abc#' ESCAPE '#' raised it, while 'abc' LIKE 'abc#' ESCAPE '#' silently returned false.

The loop now recognises the escape character, so the first pattern no longer matches "abc" and the second raises the error whatever the string is.

### DIFF
--- a/src/function/scalar/string/like.cpp
+++ b/src/function/scalar/string/like.cpp
@@ -220,7 +220,18 @@ bool TemplatedLikeOperator(const char *sdata, idx_t slen, const char *pdata, idx
 			return false;
 		}
 	}
-	while (pidx < plen && pdata[pidx] == PERCENTAGE) {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25423, opened 2026-09-07
Prevent native stack overflow during join-order optimization

XiaocongDing:
Problem:
Correlated subqueries can make join-order optimization recurse deeply and exhaust the native thread stack before max_expression_depth can stop it.

Fix:
Add a Linux NativeStackChecker with native stack-bound lookup. Invoke it from the generic StackChecker constructor to cover recursive parser and expression binding paths, and directly at each join-order optimizer entry. Raise an input error before the native stack safety margin is exhausted.

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -73,6 +73,7 @@ add_library_unity(
   radix_partitioning.cpp
   re2_regex.cpp
   random_engine.cpp
+  stack_checker.cpp
   stacktrace.cpp
   string_util.cpp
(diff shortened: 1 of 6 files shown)


taniabogatsch:
Hi @XiaocongDing! Thanks for the contribution. However, we are worried about this breaking for (some) musl builds and because it can break for builds with specific stack protection flags.

We're instead working on more detection (and fixing) of deep recursive components and code paths, see for example here to detect it: https://github.com/duckdb/duckdb/pull/25010

PR #25424, opened 2026-09-07
Remove `LogicalType` from parsed `CastExpression`

Maxxen:
This PR removes the `LogicalType cast_type` member from the parsed `CastExpression`, replacing it with a `TypeExpression`.

This is mostly to future-proof serialization and a minor step towards removing bound `LogicalTypes` from the parser/parse tree entirely.

Backwards compatibility is preserved by serializing into a UNBOUND type, which itself gets default-bound when serialized.


Mytherin:
Thanks!

PR #25425, opened 2026-09-07
Run CSV Byte skipping over 64 bytes

pdet:
When reading CSV files we skip the bytes the state machine does not care about. The bytes that matter are:
 1. the delimiter
 2. the quote
 3. the newline
 4. the comment
 5. the escape 
 Everything else is data, and won't cause a change of state in the state machine. 

We used to test 8 bytes at a time for whether any special byte was present, then walk byte by byte to find it. This PR builds the patterns from the dialect, runs them over 64-byte blocks and gets a bitmap of the positions that may be special. The state machine then only looks at those positions.

One important caveat is that the bitmap has false positives, a flagged byte might not be a change of state (e.g., a delimiter inside a quoted value), that won't cause a correctness issue as the state machine will check it, but it's still a bit of wasted performance, I'll solve it in a follow-up PR.


Mytherin:
Thanks!

PR #25427, opened 2026-09-07
Apply collations to struct and tuple fields

### LINKED ISSUE #24047: Row comparisons ignore element collations
### What happens?

See repro for details.

### To Reproduce

```sql
-- should be true, true
memory D SELECT
           'a' COLLATE NOCASE < 'B' AS scalar_cmp,
           ('a' COLLATE NOCASE, 0) < ('B', 0) AS row_cmp;
┌────────────┬─────────┐
│ scalar_cmp │ row_cmp │
│  boolean   │ boolean │
├────────────┼─────────┤
│ true       │ false   │
└────────────┴─────────┘
```

### OS:

macos

### DuckDB Version:

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

HendrikLambert:
Tuple and struct comparisons ignored field collations. For example, `('a' COLLATE NOCASE, 0) < ('B', 0)` returned `false` instead of `true`.

### Changes
- This PR applies collations recursively to struct and tuple fields using a lambda. This also fixes ordering of nested `TIMETZ` values. 
- Aliased structs, lists and arrays are left to registered collation callbacks, because the extraction and transformation functions used here do not accept aliased types by default. Without a callback, field collations are not applied automatically.

Fixes https://github.com/duckdb/duckdb/issues/24047
Fixes https://github.com/duckdblabs/duckdb-internal/issues/9701

### DIFF
--- a/src/function/scalar/list/contains_or_position.cpp
+++ b/src/function/scalar/list/contains_or_position.cpp
@@ -1,12 +1,43 @@
 #include "duckdb/function/scalar/list_functions.hpp"
 #include "duckdb/function/scalar/nested_functions.hpp"
 #include "duckdb/function/scalar/list/contains_or_position.hpp"
+#include "duckdb/common/type_visitor.hpp"
 #include "duckdb/optimizer/statistics_propagator.hpp"
+#include "duckdb/planner/expression/bound_cast_expression.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/planner/expression_binder.hpp"
 #include "duckdb/storage/statistics/list_stats.hpp"
 #include "duckdb/storage/statistics/numeric_stats.hpp"
 
 namespace duckdb {
+
+static unique_ptr<FunctionData> ListSearchBind(BindScalarFunctionInput &input) {
+	auto &arguments = input.GetArguments();
+	auto &function = input.GetBoundFunction();
+	auto &context = input.GetClientContext();
+	auto list_type = arguments[0]->GetReturnType();
+	if (list_type.id() != LogicalTypeId::LIST && list_type.id() != LogicalTypeId::ARRAY) {
+		return nullptr;
+	}
+	auto &child_type =
+	    list_type.id() == LogicalTypeId::LIST ? ListType::GetChildType(list_type) : ArrayType::GetChildType(list_type);
+	auto is_struct = [](const LogicalType &type) {
+		return StructType::IsStruct(type);
+	};
+	if (!TypeVisitor::Contains(child_type, is_struct) &&
+	    !TypeVisitor::Contains(arguments[1]->GetReturnType(), is_struct)) {
+		return nullptr;
+	}
+	auto comparison_type = LogicalType::MaxLogicalType(context, child_type, arguments[1]->GetReturnType());
+	// Merge the element types before pushing collations so each field uses the collation from either argument.
+	vector<LogicalType> types {LogicalType::LIST(comparison_type), comparison_type};
+	for (idx_t i = 0; i < arguments.size(); i++) {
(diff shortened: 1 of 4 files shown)


src/planner/collation_binding.cpp
```diff
@@ -106,26 +109,78 @@ void CollationBinding::RegisterCollation(CollationCallback callback) {
...
+	auto lambda_parameter_expr = make_uniq<BoundReferenceExpression>(lambda_parameter, sql_type, idx_t(0));
+	const bool is_tuple = sql_type.id() == LogicalTypeId::TUPLE;
+	vector<unique_ptr<Expression>> fields;
+
+	bool requires_collation = false;
+	auto &child_types = StructType::GetChildTypes(sql_type);
+	for (idx_t i = 0; i < child_types.size(); i++) {
+		vector<unique_ptr<Expression>> arguments;
+		arguments.push_back(lambda_parameter_expr->Copy());
+		arguments.push_back(make_uniq<BoundConstantExpression>(Value::BIGINT(NumericCast<int64_t>(i + 1))));
```
Mytherin, line 151: Does this work with nested structs, e.g. `{'x': {'y': {'z': 'HELLO'}}}`?

src/function/scalar/list/contains_or_position.cpp
```diff
@@ -1,13 +1,44 @@
...
+		return StructType::IsStruct(type);
+	};
+	if (!TypeVisitor::Contains(child_type, is_struct) &&
+	    !TypeVisitor::Contains(arguments[1]->GetReturnType(), is_struct)) {
+		return nullptr;
+	}
+	auto comparison_type = LogicalType::MaxLogicalType(context, child_type, arguments[1]->GetReturnType());
+	// Merge the element types before pushing collations so each field uses the collation from either argument.
+	vector<LogicalType> types {LogicalType::LIST(comparison_type), comparison_type};
+	for (idx_t i = 0; i < arguments.size(); i++) {
```
Mytherin, line 34: Ideally we shouldn't be pushing collations through a binding callback but through `FunctionCollationHandling` as a property in the scalar function

src/planner/collation_binding.cpp
Mytherin: Also how about lists inside structs, and structs inside lists? e.g.:

```
{'x': ['HELLO']}
[{'x': 'HELLO'}]
```

Mytherin (commented):

src/planner/collation_binding.cpp
HendrikLambert: Thanks for the feedback. Nested structs/lists work, but I added some more tests for coverage

Mytherin:
Thanks for adding the tests - looks good.

Had the clanker do another pass and it found the following issue in `min/max`:

```sql
# name: test/sql/collate/collate_struct_minmax.test
# description: min/max must agree with ORDER BY on collated struct fields
# group: [collate]

statement ok
CREATE TABLE tuples AS SELECT * FROM (VALUES (('a' COLLATE NOCASE, 0)), (('B', 0))) v(s)

query I
SELECT s FROM tuples ORDER BY s DESC LIMIT 1
----
(B, 0)

query I
SELECT max(s) FROM tuples
----
(B, 0)

query I
SELECT min(s) FROM tuples
----
(a, 0)

statement ok
CREATE TABLE structs AS SELECT * FROM (VALUES ({'x': 'a' COLLATE NOCASE}), ({'x': 'B'})) v(s)

query I
SELECT s FROM structs ORDER BY s DESC LIMIT 1
----
{'x': B}

query I
SELECT max(s) FROM structs
----
{'x': B}

query I
SELECT min(s) FROM structs
----
{'x': a}

```

PR #25431, opened 2026-09-07
Update projection column by correct index in expression pushdown optimizers

### LINKED ISSUE #8628: Duckdb 2.0

myrrc:
This change fixes a bug in cast/scalar fn pushdown optimizers: whenever column was part of a Projection wrapping a Get, column's index within the projection may not have been the same as column's index within a Get.

This change also fixes a bug in FinalizeScan where a thread without a `scan.job` (exhausted without scanning anything) got to run FinalizeScan which failed in MultiFileGetPartitionData after.

### DIFF
--- a/src/include/duckdb/common/multi_file/multi_file_function.hpp
+++ b/src/include/duckdb/common/multi_file/multi_file_function.hpp
@@ -868,13 +868,6 @@ public:
 
 	static void MultiFileScan(ClientContext &context, TableFunctionInput &data_p, DataChunk &output) {
 		if (!data_p.local_state) {
-			auto &gstate = data_p.global_state->Cast<MultiFileGlobalState>();
-			auto &bind_data = data_p.bind_data->CastNoConst<MultiFileBindData>();
-			if (gstate.global_state && bind_data.interface &&
-			    bind_data.interface->FinalizeScan(context, *gstate.global_state, output)) {
-				data_p.async_result = SourceResultType::HAVE_MORE_OUTPUT;
-				return;
-			}
 			data_p.async_result = SourceResultType::FINISHED;
 			return;
 		}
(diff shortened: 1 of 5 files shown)


lnkuiper:
Thanks for the PR! Two (codex-assisted) comments:

1. Could we let the projection derive its types from its expressions? The index correction fixes the crash, but leaves four places manually updating `projection->types` in `type_pushdown.cpp` and `scalar_fn_pushdown.cpp`.

   The replacement visitors already update the expressions. I would remove those four writes and resolve types once at the end of `PushdownOptimize` in `type_pushdown.hpp`:

   ```cpp
   if (any_pushed) {
   	Replace(analyses, projections).VisitOperator(*op);
   	op->ResolveOperatorTypes();
   }
   ```

   Then `GetBinding` only needs the scan analysis and scan column index. Keep the existing binding lookup and eligibility checks. The `sum(strlen(s))` query in `test/optimizer/pushdown/type_pushdown_projection_index.test` aborts on the parent ("index 2 within vector of size 1") and returns `100` here.

2. The `data.job` guard in `multi_file_function.hpp` can drop the final aggregate row. Vortex emits that row through `FinalizeScan`, including for empty inputs. With no jobs, neither exhaustion path calls it anymore.

   I think finalization needs its own output metadata. Keep local state for exhausted scans when reader global state exists, and allow finalization without a job.

myrrc:
Hi @lnkuiper. We can reason about (1) later but I don't want to introduce this change here to keep PR small and merge it faster. I've updated (2).

pdet:
Thanks!

PR #25432, opened 2026-09-07
Report and allow aborting on a WAL that cannot be fully replayed

hellower:
### The problem

A WAL whose tail cannot be replayed is handled two different ways depending on the exception that stopped the replay (`src/storage/wal_replay.cpp`):

```cpp
if (config.options.abort_on_wal_failure || error.Type() != ExceptionType::SERIALIZATION) {
    error.Throw("Failure while replaying WAL file \"" + wal_path + "\": ");
}
```

* `SERIALIZATION` — a torn WAL — is swallowed. The replay stops at the last entry that succeeded, everything after it is discarded, and startup continues.
* Anything else is rethrown and startup fails.

The tolerant half is **silent**. `WriteAheadLogReplayer::ReplayLog` writes nothing to the logger, so a database that came back with part of its committed data missing is indistinguishable from one that came back whole. From the outside the only symptom is that rows are gone.

### Why it cannot be noticed afterwards

The evidence does not survive. `WriteAheadLog::Initialize` truncates the file to `successful_offset` **lazily, at the first WAL write**. So the discarded tail is on disk only between the open and the first write of the new session; by the time anyone looks, the file already matches what was replayed.

There is no later moment at which this can be observed, which is why it has to be reported when it happens.

### `abort_on_wal_failure` is not reachable

The option already exists to turn the tolerant half into a failure, but it is not registered as a setting — the only places that assign it are DuckDB's own test helpers. No embedder or CLI user can reach it, which also means the default (tolerant) path is the one that ships while the tests exercise the other one.

### The change

* Log a warning when a replay stops early, naming the file, the offset that was replayed successfully, and the error that stopped it.
* Register `abort_on_wal_failure` as a regular global setting. It keeps its `DBConfigOptions` field, so the existing tests that set it directly are unaffected.

The new test tears a WAL in the middle of a commit and opens the same file twice: once with the setting off, where the tail is discarded and startup continues, and once with it on, where the open throws.

### How this was found

In a DuckDB-backed product, a `.wal` had grown to 25 MB while the database file had not changed for days. Restarts came back with data rewound and **not a single log line** explaining it. Reconstructing what happened meant reading `wal_replay.cpp` — there was nothing else to go on. Adding this warning is what would have made it a one-minute diagnosis.

### A related gap, not addressed here

`checkpoint_on_shutdown` has the same shape: it is not in `settings.json`, so it cannot be set with `SET`; only `PRAGMA disable_checkpoint_on_shutdown` exists, and that is one-way. The description of `checkpoint_on_detach` even refers to "the global `checkpoint_on_shutdown` setting", which users cannot actually set. Left out to keep this PR to one subject — happy to follow up if that is wanted.

### Testing

macOS arm64, `make debug` (ASAN):

* `Test abort_on_wal_failure setting` — 10 assertions, passes.
* `Test torn WAL writes` and `Test torn WAL writes followed by successful commits` — 242 assertions, pass. The tolerant path's behaviour is unchanged; only a log line is added.
* `[storage]` — 32/33 test cases pass. The one failure is `test/sql/storage/test_show_tables_persistent.test`, an ASAN stack overflow in `ListMatcher::MatchParseResultInternal`. **It reproduces identically on unmodified `main`** at 3f7a54d0fb — a plain `show tables;` is enough — so it is pre-existing and unrelated to this change. Filed separately as #25433.

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -7,6 +7,12 @@
         "custom_implementation": true
     },
     {
+        "name": "abort_on_wal_failure",
+        "description": "Whether to abort startup when the write-ahead log cannot be fully replayed, instead of discarding the part that could not be replayed",
+        "type": "BOOLEAN",
+        "scope": "global"
+    },
+    {
         "name": "access_mode",
         "description": "Access mode of the database (AUTOMATIC, READ_ONLY or READ_WRITE)",
         "type": "ENUM<AccessMode>",
--- a/src/include/duckdb/main/settings.hpp
+++ b/src/include/duckdb/main/settings.hpp
@@ -117,6 +117,19 @@ struct DeltaOnlyVariantEncodingEnabledSetting {
 	static Value GetSetting(const ClientContext &context);
 };
 
+struct AbortOnWalFailureSetting {
+	using RETURN_TYPE = bool;
+	static constexpr const char *Name = "abort_on_wal_failure";
+	static constexpr const char *Description = "Whether to abort startup when the write-ahead log cannot be fully "
+	                                           "replayed, instead of discarding the part that could not be replayed";
+	static constexpr const char *InputType = "BOOLEAN";
+	static constexpr bool IsDebug = false;
+	static constexpr bool IsDeprecated = false;
+	static void SetGlobal(DatabaseInstance *db, DBConfig &config, const Value &parameter);
+	static void ResetGlobal(DatabaseInstance *db, DBConfig &config);
+	static Value GetSetting(const ClientContext &context);
+};
+
 struct AccessModeSetting {
 	using RETURN_TYPE = AccessMode;
 	static constexpr const char *Name = "access_mode";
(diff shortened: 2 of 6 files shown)


Mytherin:
Thanks! Exposing this and logging a warning makes sense to me.

Mytherin:
Could you retarget this to `v2.0-cyanoptera`?

hellower:
Sure — I'll rebase onto `v2.0-cyanoptera` and retarget.

hellower:
Rebased this PR onto `v2.0-cyanoptera` and retargeted it. The only adjustment was in `src/main/config.cpp`: main references settings in the alias table by numeric index, so inserting a setting requires renumbering the aliases, while the release branch references them by name, which removes that renumbering and leaves only the new setting registration line. The rest of the patch applied unchanged, and the debug build passed all three relevant tests: `Test abort_on_wal_failure setting`, `Test torn WAL writes`, and `Test torn WAL writes followed by successful commits`.

PR #25435, opened 2026-09-07
Fix missing includes for non-unity build

tanishqjh-tron:
Description: Compiling the main branch with DISABLE_UNITY=1 fails on duck_schema_entry.cpp and checkpoint_manager.cpp due to missing includes for CreateWindowFunctionInfo and CreateViewInfo. This patch explicitly includes their respective headers so the dynamic_cast can resolve the complete types.


Mytherin:
Thanks!

PR #25436, opened 2026-09-07
Add support for directly pushing OpenFileInfo into the MultiFileReader list from SQL

Mytherin:
Currently the MultiFileReader only supports getting a list of file names. Internally, however, we support the more complex `OpenFileInfo` struct which can hold a bunch of extra information (e.g. file size, whether or not to check the cache, etc). This PR exposes this at the SQL level, allowing more complex reads to be constructed through SQL as well. 

CC @Tishj

### DIFF
--- a/src/common/multi_file/multi_file_reader.cpp
+++ b/src/common/multi_file/multi_file_reader.cpp
@@ -118,6 +118,69 @@ vector<string> MultiFileReader::ParsePaths(const Value &input) {
 	}
 }
 
+OpenFileInfo MultiFileReader::ParseFileEntry(const Value &input) {
+	if (input.IsNull()) {
+		throw ParserException("%s reader cannot take NULL input as parameter", function_name);
+	}
+	if (input.type().id() == LogicalTypeId::VARCHAR) {
+		return OpenFileInfo(StringValue::Get(input));
+	}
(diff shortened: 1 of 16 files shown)


src/common/multi_file/multi_file_reader.cpp
```diff
@@ -94,28 +94,67 @@ void MultiFileReader::AddParameters(TableFunction &table_function) {
...
-				throw ParserException("%s reader cannot take NULL input as parameter", function_name);
-			}
-			if (val.type().id() != LogicalTypeId::VARCHAR) {
-				throw ParserException("%s reader can only take a list of strings as a parameter", function_name);
+		return OpenFileInfo(StringValue::Get(input));
+	}
+	if (input.type().id() == LogicalTypeId::VARIANT) {
+		// a VARIANT lets every file carry its own set of open options - unpack it to its logical value
+		// a variant never unpacks to another variant, so this recurses at most once
+		return ParseFileEntry(VariantValue::GetValue(input));
```
Tishj, line 107: At most once at the outer level*
But children of the variant value could still be of type variant
Imagine a `[42, true, "test"]` ARRAY of int, bool, varchar respectively, the outer `VARIANT` will unwrap to `VARIANT[]`

PR #25437, opened 2026-09-07
Reuse `MatcherProcess` with `ArenaAllocator`

Dtenwolde:
Follow-up on https://github.com/duckdb/duckdb/pull/25421

The newly introduced `MatcherProcess` is now also backed by an arena allocator, avoiding individual heap allocations for each process. Both the heap-based and recursive matcher (still to be removed) use the same pool, which handles different sizes and alignment. Custom extension matchers can use `allocator.Make<T>()`. 

Improves performance by 9-14% compared to https://github.com/duckdb/duckdb/pull/25421, tested on Flummi, TPC-H, TPC-DS and AoC. The two PRs combined provide a 22-26% improvement against `v2.0-cyanoptera`


src/include/duckdb/parser/peg/match_process_allocator.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+#pragma once
+
+#include "duckdb/parser/peg/matcher.hpp"
+#include "duckdb/storage/arena_allocator.hpp"
+
+namespace duckdb {
+
+//! Reusable process storage; checkpoints must be restored in stack order after destruction.
+class MatchProcessAllocator {
```
Tishj: Don't you just want to use `ArenaAllocator` ?

src/parser/peg/matcher_process.cpp
```diff
@@ -115,8 +152,8 @@ class ListMatchProcess : public MatchProcess {
 	bool awaiting_child = false;
 };
 
-unique_ptr<MatchProcess> ListMatcher::StartMatch(MatchState &state) const {
-	return make_uniq<ListMatchProcess>(*this, state);
+arena_ptr<MatchProcess> ListMatcher::StartMatch(MatchState &state, MatchProcessAllocator &allocator) const {
```
Mytherin: I also wonder about the API here - maybe we can include the allocator in `MatchState` instead, then we can do `state.Make<>`?

src/include/duckdb/parser/peg/match_process_allocator.hpp
Dtenwolde: I think that's possible as well, but that doesn't allow us to reuse memory left by `MatchProcess`es that are finished right? It does simplify the implementation, but from some benchmarking it seems that the peak memory usage is 2-3x higher with the `ArenaAllocator`

src/parser/peg/matcher_process.cpp
Dtenwolde: `Make`s sense! Changed it :)

src/include/duckdb/parser/peg/match_process_allocator.hpp
Mytherin: What's the difference between the two? The `Rewind` method? Couldn't we add a similar method to the `ArenaAllocator`?

Dtenwolde: I made the change to `ArenaAllocator`. If we see that memory usage is a bottleneck then there's room for optimization :)

extension/autocomplete/autocomplete_extension.cpp
```diff
@@ -469,11 +469,14 @@ static unique_ptr<SQLTokenizeFunctionData> GenerateTokens(ClientContext &context
 	idx_t max_token_index = 0;
 	TokenIterator token_iterator(tokens);
 	auto parser_options = context.GetParserOptions();
-	MatchContext match_context(suggestions, parse_allocator, max_token_index, MatchMode::RECOGNIZE_ONLY,
-	                           parser_options.identifier_case_mode, parser_options.heap_based_parser);
+	ArenaAllocator process_allocator(Allocator::DefaultAllocator());
```
Tishj, line 472: nit: I think we want `BufferAllocator::Get(context)` instead, since we have the context here to do so 👍 
That way the allocations go through the buffer manager and thus tracking them properly.

Tishj (approved):
Thanks!

src/parser/peg/matcher_stack.cpp
```diff
@@ -76,27 +92,28 @@ void MatchStack::InitializeFrame(MatchStackFrame &frame) {
 	frame.process = matcher.StartMatch(state);
 }
 
-void MatchStack::ExecuteFrame(MatchStackFrame &frame) {
+bool MatchStack::ExecuteFrame(MatchStackFrame &frame) {
```
Tishj, line 95: nit: shouldn't we just return the `optional<MatcherResult>` ?

PR #25438, opened 2026-09-07
Fix remap_struct leaving rows uninitialized for constant MAP/LIST input

### LINKED ISSUE #25272: Bug96: RemapMap() leaves result[1..N-1] uninitialized when input is CONSTANT_VECTOR
### What happens?

In `src/function/scalar/struct/remap_struct.cpp`, `RemapMap()` (lines 91–134, bug at 102–110) handles a `CONSTANT_VECTOR` input by copying a single `list_entry_t` into `result[0]` via `memcpy`, but never marks the output vector as `CONSTANT_VECTOR` and never broadcasts the value to the remaining rows. When the enclosing struct is a `FLAT_VECTOR` of `N` rows (e.g., 2048), `result[1..N-1]` remain uninitialized; the garbage `{offset, length}` values are subsequently used to size and index into the map's key/value child vectors, causing either out-of-bounds reads or — as observed — a `D_ASSERT(false)` crash in the allocator when the uninitialized `length` exceeds `MAXIMUM_ALLOC_SIZE`.

### To Reproduce

This issue is triggered entirely by SQL — no binary modification or special setup is required.

```sql
SELECT remap_struct(
  struct_pack(id := id, m := MAP {1: 'a', 2: 'b'}),
  NULL::STRUCT(id INTEGER, m2 MAP(INTEGER, VARCHAR)),
  {'id': 'id', 'm2': ROW('m', {'key': 'key', 'value': 'value'})},
  NULL
) FROM range(2048) t(id);
```

```sh
./build/debug/duckdb < poc.sql
```

```
INTERNAL Error: Assertion triggered in file "/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/src/common/allocator/../../../../../src/common/allocator/allocator.cpp" on line 122: false

Stack Trace:

/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24c68a80) [0x63467115da80]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24c68369) [0x63467115d369]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24c67cdb) [0x63467115ccdb]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24c72e50) [0x634671167e50]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24e86385) [0x63467137b385]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24be1be7) [0x6346710d6be7]

EXIT=1
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

alonfaraj:
Fixes #25272.

`RemapMap()`/`RemapList()` copied one `list_entry_t` into `result[0]` for a `CONSTANT_VECTOR` source, leaving rows `1..N-1` stale and the result size unset. The garbage `{offset, length}` pairs then index the child vectors - assertion failure on debug builds, empty maps/lists otherwise.

Both branches are identical, so they refactored into one `CopyListEntries()` helper that broadcast the entry over every row via the writer, which also sets the size.

Added regression test.

### DIFF
--- a/src/function/scalar/struct/remap_struct.cpp
+++ b/src/function/scalar/struct/remap_struct.cpp
@@ -88,6 +88,33 @@ void RemapChildVectors(const Vector &result, const vector<reference<Vector>> &in
 	}
 }
 
+//! Copy the list_entry_t values from the input vector into the result, preserving top-level validity
+//! Returns false if the input is a constant NULL - in that case the result is set to a constant NULL
+bool CopyListEntries(Vector &input, Vector &result, idx_t result_size) {
+	bool is_constant = input.GetVectorType() == VectorType::CONSTANT_VECTOR;
+	if (is_constant && ConstantVector::IsNull(input)) {
+		ConstantVector::SetNull(result, count_t(result_size));
+		return false;
+	}
+	auto writer = FlatVector::Writer<list_entry_t>(result, result_size);
+	if (is_constant) {
+		// broadcast the single list entry over all result rows
+		auto list_entry = *ConstantVector::GetData<list_entry_t>(input);
+		for (idx_t i = 0; i < result_size; i++) {
+			writer.WriteValue(list_entry);
+		}
+		return true;
+	}
+	for (const auto entry : input.Values<list_entry_t>()) {
+		if (entry.IsValid()) {
+			writer.WriteValue(entry.GetValueUnsafe());
+		} else {
+			writer.WriteNull();
+		}
+	}
+	return true;
+}
+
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25440, opened 2026-09-07
Fix out-of-bounds write on empty CSV line with duplicate empty nullstr

### LINKED ISSUE #25268: Bug92: CSV EmptyLine Duplicate nullstr Out-of-Bounds Write
### What happens?

`StringValueResult::EmptyLine()` (`src/execution/operator/csv_scanner/string_value_scanner.cpp`, ~line 994) iterates over every `null_str` entry whose length is zero and, for each match, writes a row into `parse_chunk` and increments `number_of_rows` — all inside the loop — deferring the chunk-capacity guard (`number_of_rows >= result_size`) until after the loop exits. The CSV read path never deduplicates or rejects multiple empty entries in `nullstr` (only the COPY-TO write path does), so a caller can supply N duplicate empty strings; when an empty line is scanned near the end of a chunk, `EmptyLine()` writes 16-byte `string_t` values and validity-mask bits past the end of `parse_chunk`'s fixed-capacity backing arrays — a CWE-787 out-of-bounds heap write whose magnitude is directly controlled by the attacker via the length of the `nullstr` list.

### To Reproduce

The bug is triggered by a single SQL statement that reads a crafted single-column CSV file containing an empty line.

```sql
SET threads=1;
SELECT count(*) FROM read_csv(
    'poc.csv',
    columns={'a': 'VARCHAR'},
    header=false,
    parallel=false,
    nullstr=['', '', '', '', '', '', '', '', '', '',
             '', '', '', '', '', '', '', '', '', '',
             '', '', '', '', '', '', '', '', '', '']
);
```

```sh
# Step 1: generate the CSV (2029 data rows, one empty line, 20 more data rows)
python3 -c "
with open('poc.csv', 'w') as f:
    for _ in range(2029): f.write('0123456789\n')
    f.write('\n')
    for _ in range(20):  f.write('0123456789\n')
"

# Step 2: save the SQL above as poc.sql, then run
./build/debug/duckdb < poc.sql
```

```
INTERNAL Error: ValidityMask::SetInvalid - row_idx 2048 is out-of-range for mask with capacity 2048

LINE 1: SELECT count(*) FROM read_csv(
                             ^^^^^^^^^

```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

alonfaraj:
Fixes #25268.

`StringValueResult::EmptyLine()` added a row for every empty `nullstr` entry instead of one row per empty line, and only checked for a full chunk afterwards. Duplicate empty strings in `nullstr` then duplicated rows and wrote past the end of `parse_chunk`.

The loop now just tests if any null string is empty, so an empty line writes at most one row.

### DIFF
--- a/src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
+++ b/src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
@@ -999,19 +999,21 @@ bool StringValueResult::EmptyLine(StringValueResult &result, const idx_t buffer_
 		result.last_position.buffer_pos++;
 	}
 	if (result.number_of_columns == 1) {
-		for (idx_t i = 0; i < result.null_str_count; i++) {
-			if (result.null_str_size[i] == 0) {
-				bool empty = false;
-				if (!result.state_machine.options.force_not_null.empty()) {
-					empty = result.state_machine.options.force_not_null[0];
-				}
-				if (empty) {
-					static_cast<string_t *>(result.vector_ptr[0])[result.number_of_rows] = string_t();
-				} else {
-					result.validity_mask[0]->SetInvalid(static_cast<idx_t>(result.number_of_rows));
-				}
-				result.number_of_rows++;
+		bool empty_is_null = false;
+		for (idx_t i = 0; i < result.null_str_count && !empty_is_null; i++) {
+			empty_is_null = result.null_str_size[i] == 0;
+		}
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25441, opened 2026-09-07
Test C++ 23 standard in one CI job

smvv:
Fixes https://github.com/duckdblabs/duckdb-internal/issues/10625.

Adding the compile flag to `no-rtti` CI job so we can test C++ 23 compatibility in each PR. An alternative was adding a dedicated CI job in NightlyTests, but that would delay the feedback to the next day (detected in nightly CI run, and an issue forwarded to engineering).

See also: https://github.com/duckdb/duckdb/pull/22895.

### DIFF
--- a/src/include/duckdb/common/vector_operations/aggregate_executor.hpp
+++ b/src/include/duckdb/common/vector_operations/aggregate_executor.hpp
@@ -827,7 +827,7 @@ public:
 		}
 
 		STATE_TYPE &Get() {
-			return *reinterpret_cast<STATE_TYPE *>(&storage);
+			return *reinterpret_cast<STATE_TYPE *>(storage);
 		}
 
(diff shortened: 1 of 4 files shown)


src/include/duckdb/common/vector_operations/aggregate_executor.hpp
```diff
@@ -845,7 +845,7 @@ class AggregateExecutor {
 		}
 
 	private:
-		typename std::aligned_storage<sizeof(STATE_TYPE), alignof(STATE_TYPE)>::type storage;
```
smvv, line 848: `aligned_storage` is [deprecated](https://en.cppreference.com/cpp/types/aligned_storage) since C++23.

Mytherin:
Thanks!

PR #25442, opened 2026-09-07
Create isolated directories for compared versions in test_runner.py

nantiamak:
When reusing the same benchmark directory for both old and new DuckDB versions in `scripts/regression/test_runner.py`, the following error could occur:

```
PR:
```
This PR removes the `benchmark_cache` option and every time creates separate directories for benchmarking each version. 
It also adds a test for the newly introduced functions.


nantiamak:
Closing, as I think something went wrong when I changed the target branch to main and the change set is huge.

Tishj:
I assume you based the branch off of `v2.0-cyanoptera`, you can rebase the branch and force-push and the other changes should disappear

Tishj:
`git rebase --onto main v2.0-cyanoptera <this-branch-name>`

nantiamak:
@Tishj Thanks for the suggestion. Indeed that was the issue. Anyway, I ended up creating a new PR (#25469), because I was hitting conflicts during rebase.

PR #25443, opened 2026-09-07
Fix CSV reader leaking trailing \n of \r\r\n terminators into next value

alwaysprince05:
Fixes the second symptom reported in #25164 (supplements #25232)

## Problem

Issue #25164 also reports a second symptom on files with `\r\r\n` line
endings: the reader consumes `\r\r` as the row terminator and leaves the
`\n` attached to the front of the first field of the next row (e.g.
symbol arriving as `"\nSPXW"` instead of `"SPXW"`). Unlike the
duplicate-rows bug fixed in #25232, this happens on **sequential reads
too**, and is invisible when the first column is cast to an integer type
(the cast silently trims the `\n`).

## Root cause

With `auto_detect=false` (or mixed-newline files) the scanner runs in
`CARRY_ON` newline mode, where five sites advance the value start
position by a hard-coded 2 bytes — written for 2-byte `"\r\n"`
terminators. With a 3-byte `"\r\r\n"` terminator the leftover `"\n"`
lands on the start of the next value, both when the terminator is fully
inside one buffer and when a buffer boundary splits it as `"\r" |
"\r\n"` or `"\r\r" | "\n"`.

## Fix

Make those sites aware of the 3-byte terminator:
- `AddRow` / `UnsetComment`: when the bytes after the terminator are
  `"\r\r\n"`, skip 3 instead of 2
- `ProcessOverBufferValue`: after consuming the remaining `"\r\n"` of a
  split terminator, refresh `last_position` to point past it (both
  early-return paths, including the double-count guard added in #25232)
- `ProcessOverBufferValue`: start the next value after a
  boundary-leading newline and never append leading newline bytes to a
  pending value

## Testing

New regression test `test/sql/copy/csv/test_rrrn_leading_newline.test`
(uses VARCHAR ids so the leak is assertable): three layouts — all rows
in a single buffer, boundary splitting as `"\r" | "\r\n"`, and as
`"\r\r" | "\n"` — asserting in both parallel and sequential modes that
no id/val starts with `"\n"`, row counts are exact and values intact.

Verified: the test **fails without the fix** and passes with it.
Full `test/sql/copy/*` suite: 549 test cases, 186,683 assertions, all
passing. Formatting checked with clang-format.

### DIFF
--- a/src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
+++ b/src/execution/operator/csv_scanner/scanner/string_value_scanner.cpp
@@ -215,6 +215,10 @@ bool StringValueResult::UnsetComment(StringValueResult &result, idx_t buffer_pos
 	result.comment = false;
 	if (result.state_machine.dialect_options.state_machine_options.new_line.GetValue() != NewLineIdentifier::CARRY_ON) {
 		result.last_position.buffer_pos = buffer_pos + 1;
+	} else if (buffer_pos + 2 < result.buffer_size && result.buffer_ptr[buffer_pos + 1] == '\r' &&
+	           result.buffer_ptr[buffer_pos + 2] == '\n') {
+		// The row terminator is \r\r\n (3 bytes), not \r\n (2 bytes), so skip it fully
+		result.last_position.buffer_pos = buffer_pos + 3;
 	} else {
 		result.last_position.buffer_pos = buffer_pos + 2;
 	}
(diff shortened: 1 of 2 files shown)


alwaysprince05:
Friendly ping for review whenever someone has a chance. All checks are passing, there are no conflicts with the base branch, and this is a small, self-contained follow-up to the already-merged #25232 — it fixes the second symptom from #25164 (a leading \n leaking into the first field of the next row with \r\r\n line endings). Happy to rebase or make any changes requested. Thanks!

PR #25444, opened 2026-09-07
Add double star and helix join order solvers

tanejaansh007:
Ports two shape-specific join reordering algorithms from a DataFusion prototype so they can be benchmarked head to head against the DPhyp plan enumerator. Both are off by default.

A double star (bowtie) is two hubs with their single-edge spokes bridged by a central relation, solved analytically by ordering spokes cheapest fanout first and searching the split point on each side. A helix is a chain of diamonds, which contains cycles and is solved by dynamic programming over connected subgraphs.

Only the search algorithms are ported. Cardinalities come from the CardinalityEstimator and edge selectivity is derived from it as card({a,b}) / (card(a) * card(b)), so a shape solver prices the same graph the dynamic programming does. The chosen order is materialized through the existing EmitPair, so the plan is costed by the same CostModel and rebuilt by QueryGraphManager::Reconstruct whichever algorithm chose it.

Settings:
  double_star_join_order          select the bowtie solver
  helix_join_order                select the helix solver
  helix_join_order_max_relations  bound the 3^n helix search

Claude-Session: https://claude.ai/code/session_01Fu4sminrLZNEWmc2J2CXPQ

### DIFF
--- a/src/common/settings.json
+++ b/src/common/settings.json
@@ -588,6 +588,13 @@
         "custom_implementation": true
     },
     {
+        "name": "double_star_join_order",
+        "description": "Reorder double star (bowtie) join graphs with a shape-specific solver instead of dynamic programming",
+        "type": "BOOLEAN",
+        "default_scope": "local",
+        "default_value": "false"
+    },
+    {
         "name": "duckdb_api",
         "description": "DuckDB API surface",
         "type": "VARCHAR",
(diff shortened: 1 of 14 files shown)


lnkuiper:
Thanks for the PR! We're not interested in merging these algorithms. If join ordering becomes prohibitively expensive, I think [Neumann's "Adaptive Optimization of Very Large Join Queries"](https://dl.acm.org/doi/10.1145/3183713.3183733) (and the [follow-up](https://portal.fis.tum.de/en/publications/lindp-generalizing-linearized-dp-to-crossproducts-and-non-inner-j/?utm_source=chatgpt.com))is the way to go.

Issue #24307, opened 2026-07-30
ROWS ... FOLLOWING gives incorrect non-empty frames


dentiny:
### What happens?

See repro for details.

### To Reproduce

```sql
SELECT i,
                count(*) OVER (
                  ORDER BY i
                  ROWS BETWEEN 9223372036854775807 FOLLOWING
                           AND 9223372036854775807 FOLLOWING
                ) AS cnt,
                list(i) OVER (
                  ORDER BY i
                  ROWS BETWEEN 9223372036854775807 FOLLOWING
                           AND 9223372036854775807 FOLLOWING
                ) AS members
         FROM range(3) t(i);
┌───────┬───────┬───────────┐
│   i   │  cnt  │  members  │
│ int64 │ int64 │  int64[]  │
├───────┼───────┼───────────┤
│     0 │     0 │ NULL      │
│     1 │     3 │ [0, 1, 2] │
│     2 │     3 │ [0, 1, 2] │
└───────┴───────┴───────────┘
```

### OS:

macos

### DuckDB Version:

d54e4f6a20c75e4cf1bdd7bbecb24bd08a62a491

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

hawkfish:
Hi @dentiny - I believe the fix has been merged.

PR #25446, opened 2026-09-07
Issue #24307: Negative PRECEDING FOLLOWING

hawkfish:
* Reject negtive preceding and following offsets
* Clamp over/underflows to the correct end of the partition.
* Use GREATEST to clamp some test queries to legal ranges.


Mytherin:
Thanks!

Issue #25048, opened 2026-08-27
Wrong timezone for VARIANT cast


thomas-0816:
### What happens?

```
memory D set timezone = 'Europe/Berlin';
memory D select '1992-09-20 12:30:00.123456789+05:00'::TIMESTAMPTZ::VARCHAR;
┌────────────────────────────────────────────────────────────────────────────────────────────┐
│ CAST(CAST('1992-09-20 12:30:00.123456789+05:00' AS "TIMESTAMP WITH TIME ZONE") AS VARCHAR) │
│                                          varchar                                           │
├────────────────────────────────────────────────────────────────────────────────────────────┤
│ 1992-09-20 09:30:00.123456+02                                                              │
└────────────────────────────────────────────────────────────────────────────────────────────┘
memory D select '1992-09-20 12:30:00.123456789+05:00'::TIMESTAMPTZ::VARIANT::VARCHAR;
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ CAST(CAST(CAST('1992-09-20 12:30:00.123456789+05:00' AS "TIMESTAMP WITH TIME ZONE") AS "VARIANT") AS VARCHAR) │
│                                                    varchar                                                    │
├───────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 1992-09-20 07:30:00.123456+00                                                                                 │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
memory D select '1992-09-20 12:30:00.123456789+05:00'::TIMESTAMPTZ::VARIANT;
┌──────────────────────────────────────────────────────────────────────────────────────────────┐
│ CAST(CAST('1992-09-20 12:30:00.123456789+05:00' AS "TIMESTAMP WITH TIME ZONE") AS "VARIANT") │
│                                           variant                                            │
├──────────────────────────────────────────────────────────────────────────────────────────────┤
│ 1992-09-20 07:30:00.123456+00                                                                │
└──────────────────────────────────────────────────────────────────────────────────────────────┘
```

(similar to https://github.com/duckdb/duckdb/issues/23500)

### To Reproduce

set timezone = 'Europe/Berlin';
select '1992-09-20 12:30:00.123456789+05:00'::TIMESTAMPTZ::VARCHAR;
select '1992-09-20 12:30:00.123456789+05:00'::TIMESTAMPTZ::VARIANT::VARCHAR;
[...]

PR #25450, opened 2026-09-07
Issue #25048: VARIANT Cast Overrides

hawkfish:
* Use the full set of casts when casting to e.g., TIMESTAMP_TZ, not just the defaults.


Mytherin:
Thanks!

PR #25453, opened 2026-09-07
Fix NOT NULL constraint verification after transaction-local deletes

### LINKED ISSUE #25452: NOT NULL constraint verification doesn't consider transaction-local deletes
### What happens?

Hi team, I found when we verify constraints, local deletes are considered, only permanent ones are visible; see repro SQL for details.

### To Reproduce

```sql
memory D CREATE TABLE t(i INTEGER);
memory D INSERT INTO t VALUES (NULL), (42);
memory D BEGIN TRANSACTION;
memory D DELETE FROM t WHERE i IS NULL;
memory D ALTER TABLE t ALTER COLUMN i SET NOT NULL;
Constraint Error:
NOT NULL constraint failed: t.i
```

### OS:

macos

### DuckDB Version:

6d7abb1090e1079f252b15ba0366310bb87d8582

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25452

I found this issue when I was debugging https://github.com/duckdb/duckdb/pull/25413#issuecomment-5568280499, the root cause of which is in the current implementation, constraints are verified only on [committed transactions](https://github.com/duckdb/duckdb/blob/070ce1f695bb17ef68f1fdd5ba20cad4a3ac6646/src/storage/table/row_group_collection.cpp#L2305), namely local deletions are completely ignored.

In this PR, I update the visibility to
- Able to view all the deleted rows
- Throws when there's outstanding updates, which aligns with the current behavior

Key code here
```cpp
TransactionData constraint_visibility(transaction.transaction_id, transaction_manager.GetLastCommit() + 1);
ScanOptions scan_options(constraint_visibility);
scan_options.insert_type = InsertedScanType::ALL_ROWS;
scan_options.update_type = UpdateScanType::DISALLOW_UPDATES;
```


src/storage/table/row_group_collection.cpp
```diff
@@ -2300,17 +2300,23 @@ void RowGroupCollection::VerifyNewConstraint(const QueryContext &context, DataTa
...
 	InitializeCreateIndexScan(state);
 
+	auto client_context = context.GetClientContext();
+	D_ASSERT(client_context);
+	auto &transaction = DuckTransaction::Get(*client_context, parent.db);
+	auto &transaction_manager = DuckTransactionManager::Get(parent.db);
+	TransactionData constraint_visibility(transaction.transaction_id, transaction_manager.GetLastCommit() + 1);
+	ScanOptions scan_options(constraint_visibility);
+	scan_options.insert_type = InsertedScanType::ALL_ROWS;
+	scan_options.update_type = UpdateScanType::DISALLOW_UPDATES;
```
dentiny, line 2314: We don't allow outstanding updates anyway, see repro
```sql
memory D CREATE TABLE t(i INTEGER);
memory D INSERT INTO t VALUES (NULL), (42);
memory D BEGIN TRANSACTION;
memory D UPDATE t SET i = 7 WHERE i IS NULL;
memory D ALTER TABLE t ALTER COLUMN i SET NOT NULL;
TransactionContext Error:
Cannot create index with outstanding updates
```

dentiny:
One of the TSAN failures exposed in the CI is likely flaky without my change, need to take another look: https://github.com/duckdb/duckdb/actions/runs/34167871438/job/101883732296?pr=25453

src/storage/table/scan_state.cpp
```diff
@@ -266,7 +270,11 @@ bool CollectionScanState::Scan(DuckTransaction &transaction, DataChunk &result)
 			return false;
 		}
 		do {
-			row_group = GetNextRowGroup(*row_group).get();
+			if (l) {
+				row_group = GetNextRowGroup(*l, *row_group).get();
+			} else {
+				row_group = GetNextRowGroup(*row_group).get();
+			}
```
lnkuiper, line 277: This if/else is candidate for a share helper, as we have the same logic twice. Just a nitpick though, not blocking

dentiny: Yeah it'd be good to make a util, I was copying the existing code 
https://github.com/duckdb/duckdb/blob/816a3eb2d512ce359efb40d9319f6db59788422a/src/storage/table/scan_state.cpp#L293-L297

Issue #24215, opened 2026-07-28
COPY FROM DATABASE fails on valid foreign-key data


dentiny:
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE pk_integers(i INTEGER PRIMARY KEY);
memory D CREATE TABLE fk_integers(j INTEGER, FOREIGN KEY (j) REFERENCES pk_integers(i));
memory D INSERT INTO pk_integers VALUES (1),(2),(3);
memory D INSERT INTO fk_integers VALUES (1),(2);
memory D ATTACH ':memory:' AS db1;
memory D COPY FROM DATABASE memory TO db1;
Constraint Error:
Violates foreign key constraint because key "i: 1" does not exist in the referenced table
```

### OS:

macos

### DuckDB Version:

531911608719cd03b723051b8cb3b3e81cf3e03b

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

Issue #16785, opened 2025-03-22
Fail to reclaim space using `COPY FROM DATABASE x TO y;`


bubnov:
### What happens?

I tried to reclaim space using approach described here: [Compacting a Database by Copying](https://duckdb.org/docs/stable/operations_manual/footprint_of_duckdb/reclaiming_space.html#compacting-a-database-by-copying).

The copying fails with the error:

```
» ATTACH 'reclaim_space_demo.db' AS old;
» ATTACH 'reclaim_space_demo_copy.db' AS new;
» COPY FROM DATABASE old TO new;
Constraint Error:
Violates foreign key constraint because key "id: 1" does not exist in the referenced table
```

Am I doing something wrong or this is a bug?

The same error occurs when I try to do the same using `duckdb-rs` crate (v1.2.1).

### To Reproduce

1. **Create tables and fill them with some data**:

```sql
CREATE SEQUENCE IF NOT EXISTS id_sequence START 1;

CREATE TABLE IF NOT EXISTS nodes (
   id    VARCHAR PRIMARY KEY,
   name  VARCHAR NOT NULL
);

CREATE TABLE IF NOT EXISTS nodes_watchlist (
   node_id   VARCHAR UNIQUE NOT NULL,
   FOREIGN KEY (node_id) REFERENCES nodes(id)
);

CREATE TABLE IF NOT EXISTS datasources (
   id    INTEGER PRIMARY KEY DEFAULT nextval('id_sequence'),
   name  VARCHAR NOT NULL UNIQUE
);

CREATE TABLE IF NOT EXISTS datasources_nodes (
   datasource_id   INTEGER NOT NULL,
   node_id         VARCHAR NOT NULL,
   last_sync_date  TIMESTAMP_MS,
   FOREIGN KEY (datasource_id) REFERENCES datasources(id),
   FOREIGN KEY (node_id) REFERENCES nodes(id),
   UNIQUE (datasource_id, node_id)
);

CREATE TABLE IF NOT EXISTS data (
   datasource_id   INTEGER NOT NULL,
   created         TIMESTAMP_MS NOT NULL,
   value           DOUBLE NOT NULL,
   FOREIGN KEY (datasource_id) REFERENCES datasources(id),
   UNIQUE (datasource_id, created)
);

CREATE VIEW IF NOT EXISTS v_nodes_info AS
WITH
   cte_last_sync_dates AS (
      SELECT
         n.id,
         dsn.last_sync_date
      FROM nodes n
[...]

Issue #13819, opened 2024-09-09
Transactions not working as expected with DELETE/TRUNCATE


gordonhart:
### What happens?

Unable to delete from multiple tables with foreign key constraints in a transaction.

Expected behavior: changes are committed on `commit()` or can be rolled back as desired

Actual behavior: foreign key constraint failure due

### To Reproduce

```py
import duckdb

conn = duckdb.connect()
conn.sql("CREATE TABLE a (id INTEGER PRIMARY KEY)")
conn.sql("CREATE TABLE b (a_id INTEGER NOT NULL, FOREIGN KEY (a_id) REFERENCES a (id))")
conn.sql("INSERT INTO a (id) VALUES (1), (2), (3)")
conn.sql("INSERT INTO b (a_id) VALUES (1), (2), (3)")
conn.begin()
conn.sql("TRUNCATE b")
conn.sql("TRUNCATE a")  # <-- Violates foreign key constraint because key "a_id: 1" is still referenced
conn.commit()
```

Full traceback:

```
Traceback (most recent call last):
  File ".../duckdb_transaction_bug.py", line 11, in <module>
    conn.sql("TRUNCATE a")
duckdb.duckdb.ConstraintException: Constraint Error: Violates foreign key constraint because key "a_id: 1" is still referenced by a foreign key in a different table
```

Verified that this happens on 0.10.3 as well as 1.0.0.

### OS:

MacOS Sonoma 14.1

### DuckDB Version:

0.10.3

### DuckDB Client:

Python

### Full Name:

Gordon Hart

### Affiliation:

Kolena

### What is the latest build you tested with? If possible, we recommend testing with the latest nightly build.

I have tested with a stable release

### Did you include all relevant data sets for reproducing the issue?

Yes

### Did you include all code required to reproduce the issue?

- [X] Yes, I have

### Did you include all relevant configuration (e.g., CPU architecture, Python version, Linux distribution) to reproduce the issue?

- [X] Yes, I have

gordonhart:
Looks like this might be covered under the [Over-Eager Unique Constraint Checking](https://duckdb.org/docs/sql/indexes#over-eager-unique-constraint-checking) known limitation in the documentation about indexes. Would still appreciate a second look from somebody more familiar with DuckDB.

Love the project btw — way more joy than pain has come from using DuckDB so far.

gordonhart:
[...]

Issue #10574, opened 2024-02-11
:link: FOREIGN KEY issue on self referencing table  while batch loading


adriens:
### What happens?

While loading a table that references itself wi a FK from a `csv` I get the following error: 

```
Violates foreign key constraint 
```

### To Reproduce

Create a test file:

```sh
tee family.csv <<EOF
bart,omer
lisa,omer
maggie,omer
omer,grampa
grampa,
EOF
```

Then in duckdb instance:

```sql
create or replace table simpsons_family(
    person varchar primary key,
    father varchar references simpsons_family(person)
);
```

To reproduce the issue:

```sql
insert into simpsons_family 
    select * from 'family.csv';
```

leads to : 

```
Error: Constraint Error: Violates foreign key constraint because key "person: omer"
does not exist in the referenced table
```

![image](https://github.com/duckdb/duckdb/assets/5235127/92a5de19-b93c-4b6e-a934-7c7e26ef8627)

## Expected behavior

Behind the scene, the csv should be seen as a whole.

```sql
insert into simpsons_family (person) values ('grampa');
insert into simpsons_family (person, father) values ('omer', 'grampa');
insert into simpsons_family (person, father) values ('bart', 'omer');
insert into simpsons_family (person, father) values ('lisa', 'omer');
insert into simpsons_family (person, father) values ('maggie', 'omer');
select * from simpsons_family;
```

![image](https://github.com/duckdb/duckdb/assets/5235127/bfa17480-1855-4c03-95c3-19c3cc010834)

### OS:

Linux Mint 21.3 Virginia

### DuckDB Version:

v0.9.2 3c695d7ba9

### DuckDB Client:

duckdb (cli)

### Full Name:

Adrien Sales

### Affiliation:

dev.to/adriens

### Have you tried this on the latest [nightly build](https://duckdb.org/docs/installation/?version=main)?

I have tested with a nightly build

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [X] Yes, I have

adriens:
Hi guys, any news for this issue ? 🙏

taniabogatsch:
Hi @adriens, at the moment, this is expected behavior; see this issue: https://github.com/duckdb/duckdb/issues/7168. I've picked up work on the indexes again, but I am currently focusing on issues with FKs causing (very) slow performance (https://github.com/duckdb/duckdb/pull/12808). I plan to pick this up, but I cannot give you a timeline yet.

[...]

PR #25457, opened 2026-09-08
Defer foreign key constraint validation until the end of the statement

### LINKED ISSUE #7168: Defer foreign key constraint validation until the end of the statement
### What happens?

When a table contains a self reference, then I would expect to be able to insert a complete tree / graph in a multi row insert statement. However, this doesn't seem to be possible, currently. Children can't reference parents from the same `INSERT`. I have to split the statement into multiple statements.

### To Reproduce

```sql
create table t (
  id int not null primary key,
  parent int references t (id)
);
insert into t values (1, null), (2, 1);
```

This produces:

> SQL Error: Constraint Error: Violates foreign key constraint because key "id: 1" does not exist in the referenced table

I have to run 2 separate `INSERT` statements, instead:

```sql
insert into t values (1, null);
insert into t values (2, 1);
```

### OS:

Microsoft Windows [Version 10.0.22621.1555]

### DuckDB Version:

v0.7.1

### DuckDB Client:

JDBC

### Full Name:

Lukas Eder

### Affiliation:

Data Geekery / jOOQ

### Have you tried this on the latest `master` branch?

- [X] I agree

- [X] I agree

gulugulubing:
Fixes #7168, #10574, #24215, #16785.

Foreign keys are currently checked while a statement is still executing (per chunk in the insert sinks), before the rows of that chunk have been appended. A statement that inserts multiple rows into a self-referencing table therefore fails, even though all referenced rows exist by the end of the statement. The same early checking makes statements that write multiple tables depend on pipeline execution order: COPY FROM DATABASE is a single statement with one INSERT per table, and whether it succeeds depends on the table order and on parallel pipeline execution.

Issue relations: #7168 and #10574 are the same problem on a single table (#10574 is batch loading into a self-referencing table). #24215 reports the same failure as the original #16785 report; the thread of #16785 additionally contains a self-referencing FK variant, which is covered by the same change.

This PR defers the check to the end of the statement:

- The FK case is removed from the eager per-chunk VerifyAppendConstraints (NOT NULL / CHECK / UNIQUE remain eager).
- Transaction-local storage records the row range each statement appended (statement = query number, range = local row positions), in the two functions through which appends enter local storage (LocalStorage::Append and LocalStorage::LocalMerge).
- Once a statement has finished executing (all pipelines completed, before any commit), the client context verifies the foreign keys for the rows that statement appended. Only the foreign key columns are read, in batches of STANDARD_VECTOR_SIZE rows, so memory stays bounded; no pipeline ordering is needed for foreign keys.
- The key columns are read back from transaction-local storage with a range scan over the appended rows (vectorized, only the FK key columns); index construction already read these columns at the end of the insert, so this is usually cache-warm.
- Nothing is buffered or copied - the rows already live in transaction-local storage, and only the appended row range is tracked.
- Verification runs before the commit decision, and a foreign key violation is reported as a statement error through the existing error path.
- Appends made while no statement is running (the internal appender can be used that way) keep their eager checks, since there is no statement end to defer to. When the internal appender is used during a statement, its appends are deferred like any other.
- Catalogs whose transaction objects are not DuckDB transactions (storage extensions) are skipped during verification, so attaching such a catalog does not interfere with statement execution.

Not covered by this PR:

- DELETE/TRUNCATE-side statement-level visibility (#13819).

Performance (release build, Apple M3):

- benchmark/micro/index/insert/insert_pk_fk.benchmark (in-memory, three foreign keys): median 310.4 ms -> 314 ms (+0.7%). An earlier version read the key columns back row by row and showed +17.9% (310.4 ms -> 365.8 ms) on the same benchmark.
- Largest overhead we measured: inserting 6M rows of TPC-H SF1 lineitem into a 16-column table with one foreign key, on a file-backed database where completed row groups are written to disk and compressed while the insert runs: +8-14%.
- insert_pk_fk.benchmark is not part of .github/regression/micro.csv, so CI does not run this workload today.

Tests:

- new: test/sql/constraints/foreignkey/fk_statement_level_validation.test (+ _slow), and test/api/test_appender_api.cpp "Test internal appender foreign key checks"
- new: test/api/test_storage_extension_alias.cpp "Statements executed while a non-DuckDB catalog is attached"
- un-skipped: test/sql/copy_database/copy_database_fk.test (previously "mode skip instable" with an FK ordering FIXME) now passes
- updated expectations: test_fk_self_referencing.test and foreign_key_matching_columns.test. A row may now reference itself, which is consistent with statement-level visibility (SQLite allows this as well); both tests encoded the previous, earlier checking behavior.

### DIFF
--- a/src/include/duckdb/main/client_context.hpp
+++ b/src/include/duckdb/main/client_context.hpp
@@ -321,6 +321,9 @@ private:
 	ErrorData EndQueryInternal(ClientContextLock &lock, bool success, bool invalidate_transaction,
 	                           optional_ptr<ErrorData> previous_error);
 
+	//! Verify the foreign keys of all rows appended by the current statement
+	void VerifyDeferredForeignKeys();
+
 	//! Wait until a task is available to execute
 	void WaitForTask(ClientContextLock &lock, BaseQueryResult &result);
 	PendingExecutionResult ExecuteTaskInternal(ClientContextLock &lock, BaseQueryResult &result, bool dry_run = false);
--- a/src/include/duckdb/storage/data_table.hpp
+++ b/src/include/duckdb/storage/data_table.hpp
@@ -259,6 +259,9 @@ public:
 	//! Verify constraints with a chunk from the Append containing all columns of the table
 	void VerifyAppendConstraints(ConstraintState &constraint_state, ClientContext &context, DataChunk &chunk,
 	                             optional_ptr<LocalTableStorage> local_storage, optional_ptr<ConflictManager> manager);
+	//! Verify the foreign keys of rows appended to the transaction-local storage in [start, end)
+	void VerifyAppendedForeignKeys(ClientContext &context, idx_t start, idx_t end,
+	                               const vector<unique_ptr<BoundConstraint>> &bound_constraints);
 
 	shared_ptr<DataTableInfo> &GetDataTableInfo();
 
--- a/src/include/duckdb/storage/table/append_state.hpp
+++ b/src/include/duckdb/storage/table/append_state.hpp
@@ -16,6 +16,7 @@
 #include "duckdb/transaction/transaction_data.hpp"
 
 namespace duckdb {
+class ClientContext;
 class ColumnSegment;
 class DataTable;
 class LocalTableStorage;
@@ -135,6 +136,8 @@ struct LocalAppendState {
 	TableAppendState append_state;
 	LocalTableStorage *storage;
 	unique_ptr<ConstraintState> constraint_state;
+	//! The client context this append belongs to
+	optional_ptr<ClientContext> context;
 };
 
 } // namespace duckdb
--- a/src/include/duckdb/transaction/local_storage.hpp
+++ b/src/include/duckdb/transaction/local_storage.hpp
@@ -10,6 +10,7 @@
 
 #include "duckdb/storage/table/table_index_list.hpp"
 #include "duckdb/storage/optimistic_data_writer.hpp"
+#include "duckdb/common/constants.hpp"
 #include "duckdb/common/error_data.hpp"
 #include "duckdb/common/reference_map.hpp"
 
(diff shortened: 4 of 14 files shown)


Mytherin:
Thanks but I think this is the wrong way to go about it. We shouldn't be duplicating all data we just inserted only for foreign key verification. I also wonder if we want to allow rows with self-referential foreign keys where the rows reference themselves, which I think this enables?

gulugulubing:
Thanks for taking a look — agreed, buffering full copies of the inserted rows is not the right approach.                                  
                                                                                                                                  
The motivating case is #7168: rows referencing other rows of the same statement (e.g. a tree in a self-referencing table) fail today because validation runs per chunk, before the statement's own rows are visible.                                            
                                                                                                                                  
On rows referencing themselves: that's a side effect of deferring the check rather than the goal — at the end of the statement  such a row exists, so the constraint holds. Postgres seems to allow this as well. If you'd rather keep pure self-references an error, we can keep that with an explicit rule; please let me know which semantics you prefer.                                    
                                                                                                                                  
On avoiding the copy, two options:                                                                                               
                                                                                                                                  
(a) Buffer only the foreign key columns (projected per constraint) and verify against those at the end of the statement — memory proportional to keys, nothing re-read afterwards.                                                                                
                                                                                                                                  
(b) No buffer: track the row range this statement appended to the local storage and scan it at the end of the statement — flat memory, but large file-backed inserts may have spilled row groups to disk by then, requiring a read-back.                        
                                                                                                                                  
I'd prefer (a); happy to rework — or follow a different architecture if you have one in mind.

gulugulubing:
Updated: the deferred verification now buffers only the foreign key columns (projected per constraint) instead of the full rows.

gulugulubing:
Reworked this with a different approach: after looking into the other FK reports (#10574, #24215, #16785) it seemed that a more fundamental fix was needed, so the check now runs at the end of the statement and nothing is buffered (details are in the updated description). The earlier concern about this variant was the cost of reading the keys back, but index construction already reads those columns at the end of the insert, so that read is likely cache-warm - which removed most of that concern.

On self-referential FKs: I went with allowing a row to reference itself. I verified this in SQLite, and I expect PostgreSQL to behave the same, although I have not verified it.

gulugulubing:
Updated: the keys are now read back with a range scan instead of per-row fetches. Performance numbers are in the updated description (insert_pk_fk.benchmark: +0.7%).

PR #25458, opened 2026-09-08
perf(variant): Optimize shredded vector value retrieval

darion-yaphet:
## Summary

  - Avoid intermediate Value and STRUCT materialization in ShreddedVectorBuffer::GetValue
  - Directly slice the requested row from the shredded backing vector before unshredding it
  - Add regression coverage for partially shredded nested VARIANT values, fallback values, and NULLs
  - Add a focused microbenchmark for shredded VARIANT GetValue calls

  ## Performance

  ShreddedVectorGetValue benchmark, 10 timed runs:

| Benchmark | Timed Runs | Before (Median) | After (Median) | Improvement |
|---|---:|---:|---:|---:|
| `ShreddedVectorGetValue` | 10 | 5.087 s | 3.289 s | 35.34% faster (1.55x) |

  This reduces elapsed time by 35.34% (1.55x faster) for repeated row-wise reads from a partially shredded nested VARIANT vector.

  ## Testing

  - make reldebug
  - make benchmark
  - build/reldebug/test/unittest test/sql/storage/types/variant/variant_shredding_small.test
  - build/reldebug/test/unittest test/sql/storage/types/variant/variant_shredded_slice.test
  - clang-format --dry-run --Werror benchmark/micro/shredded_vector.cpp src/common/vector/shredded_vector.cpp

  ## Risk and Rollback

  The change reuses the existing UnshredVariantData implementation and preserves the GetValue API behavior. Roll back by reverting this commit.

### DIFF
--- a/src/common/vector/shredded_vector.cpp
+++ b/src/common/vector/shredded_vector.cpp
@@ -40,23 +40,10 @@ void ShreddedVectorBuffer::SetVectorType(VectorType new_vector_type) {
 }
 
 Value ShreddedVectorBuffer::GetValue(const LogicalType &type, idx_t index) const {
-	// FIXME: this is extremely inefficient
-	auto &shredded = StructVector::GetEntries(*shredded_data)[1];
-	auto &unshredded = StructVector::GetEntries(*shredded_data)[0];
-
-	auto shredded_val = shredded.GetValue(index);
-	auto unshredded_val = unshredded.GetValue(index);
-
-	child_list_t<LogicalType> shredded_subtypes;
-	shredded_subtypes.emplace_back(make_pair("unshredded", unshredded.GetType()));
-	shredded_subtypes.emplace_back(make_pair("shredded", shredded.GetType()));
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks - change looks good to me, could we just remove the microbenchmark? I think that's a bit excessive and I would prefer not to add more C++ microbenchmarks.

darion-yaphet:
> Thanks - change looks good to me, could we just remove the microbenchmark? I think that's a bit excessive and I would prefer not to add more C++ microbenchmarks.

Yes, but actually I just wanted to measure the effect of this optimization.

Mytherin:
Sure, and I can see how that was useful for testing the change, but I would prefer not to upstream that benchmark.

darion-yaphet:
> Sure, and I can see how that was useful for testing the change, but I would prefer not to upstream that benchmark.

Well, perhaps I should provide more detailed explanations regarding the performance in the summary section.

Mytherin:
Thanks!

PR #25459, opened 2026-09-08
[ART] Fix oversized Node256Leaf validity mask allocation

Liuzw1998:
Summary

This fixes an oversized validity mask allocation in Node256Leaf.

The fix itself is small: 
validity_t mask[CAPACITY / sizeof(validity_t)] -> validity_t mask[CAPACITY / ValidityMask::BITS_PER_VALUE];

The main complication is storage compatibility. Existing databases may already contain ART indexes written with the old, larger Node256Leaf allocation size. The patch therefore preserves the persisted allocator stride when loading existing indexes, while new indexes use the corrected compact layout.

Tests cover both the new layout and compatibility with the previous persisted layout.

### DIFF
--- a/src/execution/index/art/art.cpp
+++ b/src/execution/index/art/art.cpp
@@ -47,6 +47,28 @@ struct ARTIndexScanState : public IndexScanState {
 // ART
 //===--------------------------------------------------------------------===//
 
+static idx_t GetNode256LeafSegmentSize(const IndexStorageInfo &info, StorageVersion storage_version) {
+	// Preserve the oversized mask and padding of the persisted legacy layout.
+	struct LegacyNode256Leaf {
+		uint16_t count;
+		validity_t mask[Node256Leaf::CAPACITY / sizeof(validity_t)];
+	};
+	constexpr idx_t compact_size = sizeof(Node256Leaf);
+	constexpr idx_t legacy_size = sizeof(LegacyNode256Leaf);
+	const auto allocator_idx = NodePtr::GetAllocatorIdx(NType::NODE_256_LEAF);
+	if (info.allocator_infos.size() > allocator_idx) {
+		// Preserve the persisted segment size so existing buffers retain their original layout.
+		const auto saved_segment_size = info.allocator_infos[allocator_idx].segment_size;
+		if (saved_segment_size != compact_size && saved_segment_size != legacy_size) {
+			throw SerializationException("Invalid Node256Leaf segment size %llu (expected %llu or %llu)",
+			                             saved_segment_size, compact_size, legacy_size);
+		}
+		return saved_segment_size;
+	}
+	// New indexes must retain the original layout when targeting older storage versions.
+	return storage_version >= StorageVersion::V2_0_0 ? compact_size : legacy_size;
+}
+
 ART::ART(const Identifier &name, const IndexConstraintType index_constraint_type, const vector<column_t> &column_ids,
          TableIOManager &table_io_manager, const vector<unique_ptr<Expression>> &unbound_expressions,
          AttachedDatabase &db,
@@ -83,6 +105,7 @@ ART::ART(const Identifier &name, const IndexConstraintType index_constraint_type
 		owns_data = true;
 		auto prefix_size = NumericCast<idx_t>(prefix_count) + NumericCast<idx_t>(Prefix::METADATA_SIZE);
 		auto &block_manager = table_io_manager.GetIndexBlockManager();
+		const auto leaf256_size = GetNode256LeafSegmentSize(info, db.GetStorageManager().GetStorageVersion());
 
 		array<unsafe_unique_ptr<FixedSizeAllocator>, ALLOCATOR_COUNT> allocator_array = {
 		    make_unsafe_uniq<FixedSizeAllocator>(prefix_size, block_manager),
@@ -93,7 +116,7 @@ ART::ART(const Identifier &name, const IndexConstraintType index_constraint_type
 		    make_unsafe_uniq<FixedSizeAllocator>(sizeof(Node256), block_manager),
 		    make_unsafe_uniq<FixedSizeAllocator>(sizeof(Node7Leaf), block_manager),
 		    make_unsafe_uniq<FixedSizeAllocator>(sizeof(Node15Leaf), block_manager),
-		    make_unsafe_uniq<FixedSizeAllocator>(sizeof(Node256Leaf), block_manager),
+		    make_unsafe_uniq<FixedSizeAllocator>(leaf256_size, block_manager),
 		};
 		allocators =
 		    make_shared_ptr<array<unsafe_unique_ptr<FixedSizeAllocator>, ALLOCATOR_COUNT>>(std::move(allocator_array));
@@ -1317,6 +1340,13 @@ bool ART::MergeIndexes(IndexLock &state, BoundIndex &source_index) {
 	if (other_art.owns_data) {
 		if (prefix_count != other_art.prefix_count) {
 			throw InternalException("Failed to merge ARTs - prefix count does not match");
+		}
+		const auto target_size = NodePtr::GetAllocator(*this, NType::NODE_256_LEAF).GetSegmentSize();
+		const auto source_size = NodePtr::GetAllocator(other_art, NType::NODE_256_LEAF).GetSegmentSize();
+		if (target_size != source_size) {
+			throw InternalException(
+			    "Failed to merge ARTs - Node256Leaf segment sizes do not match (target %llu, source %llu)", target_size,
+			    source_size);
 		}
 		if (tree.HasMetadata()) {
 			// Fully deserialize other_index, and traverse it to increment its buffer IDs.
--- a/test/sql/index/art/storage/test_art_node256_leaf_layout.test
+++ b/test/sql/index/art/storage/test_art_node256_leaf_layout.test
@@ -1,0 +1,220 @@
+# name: test/sql/index/art/storage/test_art_node256_leaf_layout.test
+# description: Node256Leaf bitmap boundaries, reinsertion and persisted allocator layouts
+# group: [storage]
+
+statement ok
+SET index_scan_max_count = 1000;
+
+statement ok
+SET index_scan_percentage = 1.0;
+
+foreach version v1.0.0 v1.5.0 v2.0.0
+
+# Small blocks exercise multiple compact allocator buffers with a modest row count.
+statement ok
+ATTACH '{TEST_DIR}/node256_{version}.db' AS leaf_db (STORAGE_VERSION '{version}', BLOCK_SIZE 16384)
+
+statement ok
+CREATE TABLE leaf_db.t AS SELECT i // 256 AS k, i AS id FROM range(262144) r(i)
+
+statement ok
+CREATE INDEX leaf_idx ON leaf_db.t(k)
+
+statement ok
+CHECKPOINT leaf_db
+
+statement ok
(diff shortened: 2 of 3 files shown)


src/execution/index/art/art.cpp
```diff
@@ -47,6 +47,24 @@ struct ARTIndexScanState : public IndexScanState {
 // ART
 //===--------------------------------------------------------------------===//
 
+static idx_t GetNode256LeafAllocationSize(const IndexStorageInfo &info, StorageVersion storage_version) {
```
artjomPlaunov: We can call this GetNode256LeafSegmentSize()

src/execution/index/art/art.cpp
```diff
@@ -47,6 +47,24 @@ struct ARTIndexScanState : public IndexScanState {
...
+	struct LegacyNode256Leaf {
+		uint16_t count;
+		validity_t mask[Node256Leaf::CAPACITY / sizeof(validity_t)];
+	};
+	const auto allocator_idx = NodePtr::GetAllocatorIdx(NType::NODE_256_LEAF);
+	if (info.allocator_infos.size() > allocator_idx) {
+		// Init restores the stride but does not recalculate the allocator's buffer layout.
+		const auto size = info.allocator_infos[allocator_idx].segment_size;
```
artjomPlaunov: maybe call this saved_segment_size.

src/execution/index/art/art.cpp
```diff
@@ -47,6 +47,24 @@ struct ARTIndexScanState : public IndexScanState {
...
 //===--------------------------------------------------------------------===//
 
+static idx_t GetNode256LeafAllocationSize(const IndexStorageInfo &info, StorageVersion storage_version) {
+	struct LegacyNode256Leaf {
+		uint16_t count;
+		validity_t mask[Node256Leaf::CAPACITY / sizeof(validity_t)];
+	};
+	const auto allocator_idx = NodePtr::GetAllocatorIdx(NType::NODE_256_LEAF);
+	if (info.allocator_infos.size() > allocator_idx) {
+		// Init restores the stride but does not recalculate the allocator's buffer layout.
```
artjomPlaunov: I feel this comment is a bit unclear here; this function is selecting and returning a segment size, and the comment is describing how another part of the code (that is invoking this function) will use it. For this specific part, maybe we can clarify a more local explanation, i.e. why we choose the existing segment size here?

src/execution/index/art/art.cpp
```diff
@@ -47,6 +47,24 @@ struct ARTIndexScanState : public IndexScanState {
...
+		const auto size = info.allocator_infos[allocator_idx].segment_size;
+		if (size != sizeof(Node256Leaf) && size != sizeof(LegacyNode256Leaf)) {
+			throw SerializationException("Invalid Node256Leaf allocation size");
```
artjomPlaunov: perhaps we could give more information here:

```cpp
if (segment_size != sizeof(Node256Leaf) && segment_size != sizeof(LegacyNode256Leaf)) {
	throw SerializationException(
	    "Invalid Node256Leaf segment size %llu: expected %llu or %llu",
	    segment_size, idx_t(sizeof(Node256Leaf)), idx_t(sizeof(LegacyNode256Leaf)));
}
```

src/execution/index/art/art.cpp
```diff
@@ -1160,7 +1179,15 @@ bool ART::SupportsDeltaIndexes() const {
...
-	return make_uniq<ART>(name, constraint_type, GetColumnIds(), table_io_manager, unbound_expressions, db);
+	auto result = make_uniq<ART>(name, constraint_type, GetColumnIds(), table_io_manager, unbound_expressions, db);
+	const auto idx = NodePtr::GetAllocatorIdx(NType::NODE_256_LEAF);
+	auto &source = *(*allocators)[idx];
+	if ((*result->allocators)[idx]->GetSegmentSize() != source.GetSegmentSize()) {
+		// A database upgrade does not rewrite existing index buffers. Keep delta layouts compatible.
+		(*result->allocators)[idx] =
+		    make_unsafe_uniq<FixedSizeAllocator>(source.GetSegmentSize(), source.block_manager);
+	}
+	return std::move(result);
```
artjomPlaunov: Can you clarify why we need this block?

src/execution/index/art/art.cpp
```diff
@@ -1313,6 +1340,10 @@ bool ART::MergeIndexes(IndexLock &state, BoundIndex &source_index) {
 		if (prefix_count != other_art.prefix_count) {
 			throw InternalException("Failed to merge ARTs - prefix count does not match");
 		}
+		const auto leaf256_idx = NodePtr::GetAllocatorIdx(NType::NODE_256_LEAF);
+		if ((*allocators)[leaf256_idx]->GetSegmentSize() != (*other_art.allocators)[leaf256_idx]->GetSegmentSize()) {
+			throw InternalException("Failed to merge ARTs - Node256Leaf allocation sizes do not match");
```
artjomPlaunov: Similar to the other error review comment, it could be useful to have the error be slightly more informative here

test/sql/index/test_art_index.cpp
```diff
@@ -1,12 +1,381 @@
...
+#include "duckdb/storage/table/index_entry.hpp"
+#include "duckdb/storage/table/table_index_list.hpp"
 #include "test_helpers.hpp"
 
 #include <cfloat>
 #include <iostream>
 
 using namespace duckdb;
 
+namespace {
```
artjomPlaunov: Let's remove the new C++ tests, and expand/improve the SQL tests instead (comments added there)

test/sql/index/art/storage/test_art_node256_leaf_layout.test
```diff
@@ -0,0 +1,96 @@
...
+
+statement ok
+DETACH leaf_db
+
+statement ok
+ATTACH '{TEST_DIR}/node256_{version}.db' AS leaf_db
+
+# Includes every bit boundary, and leaves in multiple allocator buffers.
+query III
+SELECT count(*), min(id), max(id) FROM leaf_db.t WHERE k=1023
```
artjomPlaunov, line 40: we should sprinkle in some EXPLAIN ANALYZE statements along with the actual queries checking the values, to confirm that an index scan is actually being used (see other tests that check the regex for indexes), i.e. keep these tests but right next to it you can add an EXPLAIN ANALYZE for the same test, that checks for an index scan.

test/sql/index/art/storage/test_art_node256_leaf_layout.test
```diff
@@ -0,0 +1,96 @@
...
+
+query I
+SELECT count(*) FROM leaf_db.t
+----
+262400
+
+statement ok
+DETACH leaf_db
+
+endloop
```
artjomPlaunov, line 113: can we check an explicit upgrade scenario using an SQL test?

artjomPlaunov (changes requested):
Hi @Liuzw1998, great find! 

Left a lot of mostly small comments, had one question about CreateEmptyCopy() that I wasn't sure of myself.

src/execution/index/art/art.cpp
Liuzw1998: I checked again and find it unnecessary, thanks!

Liuzw1998:
@artjomPlaunov Thanks for the review! I've implemented your suggestions. Please have a look again :D

src/execution/index/art/art.cpp
```diff
@@ -1313,6 +1335,12 @@ bool ART::MergeIndexes(IndexLock &state, BoundIndex &source_index) {
 		if (prefix_count != other_art.prefix_count) {
 			throw InternalException("Failed to merge ARTs - prefix count does not match");
 		}
+		const auto leaf256_idx = NodePtr::GetAllocatorIdx(NType::NODE_256_LEAF);
+		if ((*allocators)[leaf256_idx]->GetSegmentSize() != (*other_art.allocators)[leaf256_idx]->GetSegmentSize()) {
```
artjomPlaunov: nit: perhaps we can clean this up a bit here, i.e.: 

```cpp
const auto target_size = NodePtr::GetAllocator(*this, NType::NODE_256_LEAF).GetSegmentSize();
const auto source_size = NodePtr::GetAllocator(other_art, NType::NODE_256_LEAF).GetSegmentSize();
```

src/execution/index/art/art.cpp
```diff
@@ -47,6 +47,27 @@ struct ARTIndexScanState : public IndexScanState {
...
+static idx_t GetNode256LeafSegmentSize(const IndexStorageInfo &info, StorageVersion storage_version) {
+	struct LegacyNode256Leaf {
+		uint16_t count;
+		validity_t mask[Node256Leaf::CAPACITY / sizeof(validity_t)];
+	};
+	const auto allocator_idx = NodePtr::GetAllocatorIdx(NType::NODE_256_LEAF);
+	if (info.allocator_infos.size() > allocator_idx) {
+		// Preserve the persisted segment size so existing buffers retain their original layout.
+		const auto saved_segment_size = info.allocator_infos[allocator_idx].segment_size;
+		const auto compact_size = sizeof(Node256Leaf);
```
artjomPlaunov: nit: maybe we can move these out before the if, as constexpr, i.e. define as compac_size and legacy_size so we can use them both within the branch, and after the if branch.

src/execution/index/art/art.cpp
```diff
@@ -47,6 +47,27 @@ struct ARTIndexScanState : public IndexScanState {
 // ART
 //===--------------------------------------------------------------------===//
 
+static idx_t GetNode256LeafSegmentSize(const IndexStorageInfo &info, StorageVersion storage_version) {
+	struct LegacyNode256Leaf {
```
artjomPlaunov, line 52: maybe it is worth leaving a comment here that this is intentional above the struct, i.e. 

```
// Preserve the oversized mask and padding of the persisted legacy layout.
```

artjomPlaunov (changes requested):
@Liuzw1998 thanks! Just a few more nits then I think this will be ready (although I will request one more reviewer on this to be safe here, due to the backwards compatibility concerns).

Liuzw1998:
> @Liuzw1998 thanks! Just a few more nits then I think this will be ready (although I will request one more reviewer on this to be safe here, due to the backwards compatibility concerns).

Done! Sure, we should be careful about backwards compatibility...

artjomPlaunov:
Can you look into the CI failure? The stack trace was cutoff, so hard to tell if this is related or not.

artjomPlaunov:
Nevermind, I see this popping up on other CI runs as well

Liuzw1998:
> Nevermind, I see this popping up on other CI runs as well

Shall we retarget to v2.0 instead of main?

dentiny:
> Shall we retarget to v2.0 instead of main?

bugfix PR should target against v2.0

artjomPlaunov:
I'm actually unsure which branch is best here; I was leaning towards main because even though we were over-allocating, the code was still functional because it was just using the first 256 bits of the over-allocated space; which incidentally makes the backwards compatibility concerns here easier to deal with), and we want to keep 2.0 as stable as possible at this point. 

@taniabogatsch I would say main, but do you have any input here? (Also requested a final review pass from you)

taniabogatsch (approved):

taniabogatsch:
I think we can keep this for `v2.0`, I'd say less memory allocated by the ART is quite desirable.

artjomPlaunov:
Alrighty, then @Liuzw1998 this is good to go!

PR #25461, opened 2026-09-08
Identifier in error messages currently always quoted, revert back to quoted only if necessary

carlopi:
Minor, but bumped in the context of bumping extensions, and saw error messages changing formatting, see https://github.com/duckdb/duckdb-mysql/pull/300/changes/b2b23010465fb788879db6aa0963425d37f7495e.

I think this can be fixed duckdb/duckdb side by avoiding unnecessary quotes around Identifiers (introduced by Identifier in error messages rework).

Very much not critical, and might be also intentional as is now the code.


Tishj:
unquoted* ?

carlopi:
There are way more sites, unsure if this is worth changing back, or just go with the flow, since it was intentional in https://github.com/duckdb/duckdb/pull/24878. Let's close this.

PR #25465, opened 2026-09-08
Error on byte mismatch when writing to local file system

lnkuiper:
If there is, for any reason, a byte mismatch here, the write should fail instead of silently succeeding with a wrong byte count.


Mytherin:
Thanks!

PR #25469, opened 2026-09-08
Create isolated directories for compared versions in test_runner.py

nantiamak:
When reusing the same benchmark directory for both old and new DuckDB versions in `scripts/regression/test_runner.py`, the following error could occur:
```
PR:
```
This PR removes the `benchmark_cache` option and every time creates separate directories for benchmarking each version.
It also adds a test for the newly introduced functions.


Tishj:
Why is this to main and not v2.0-cyanoptera?

Tishj (commented):
It looks good as far as I can tell 👍 
I'd maybe add another change to make the symlinked data read-only (if possible), since it's shared data, it should ideally be static and const.
Also, double check if we are cleaning up the `BENCHMARK_DATA_DIRECTORY` contents between regression tests, so we aren't accumulating giant files throughout the lifetime of the benchmark workflow.

nantiamak:
> I'd maybe add another change to make the symlinked data read-only (if possible), since it's shared data, it should ideally be static and const.

@Tishj By this, do you mean chmoding the files in the source?

nantiamak:
> I'd maybe add another change to make the symlinked data read-only (if possible), since it's shared data, it should ideally be static and const.

Actually, if we do so, it looks like [this line](https://github.com/nantiamak/duckdb/blob/10af5617d1191198d59607c59781821c63052632/benchmark/interpreted_benchmark.cpp#L54) will break for `load` commands. We could pass `READ_ONLY` as a value for `access_mode` depending on the benchmark, but maybe that should be done in a separate PR, as it touches the underlying C++ code?

nantiamak:
> Also, double check if we are cleaning up the BENCHMARK_DATA_DIRECTORY contents between regression tests, so we aren't accumulating giant files throughout the lifetime of the benchmark workflow.

We call `cleanup` on the directories [here](https://github.com/nantiamak/duckdb/blob/10af5617d1191198d59607c59781821c63052632/scripts/regression/test_runner.py#L939). Is this what's you mean?

nantiamak:
> Why is this to main and not v2.0-cyanoptera?

Talked with @smvv and agreed to rebase to v2.0-cyanoptera.

smvv (approved):
Looks good, thanks!

Mytherin:
Thanks!

PR #25471, opened 2026-09-08
Add `TableFunctionMultiFileWrapper` which allows a single-file-read table function to become a multi-file-read function, and migrate `read_json` to it

Mytherin:
Currently the `MultiFileReader` requires a separate interface to be implemented - the `MultiFileReaderInterface`. This interface looks a lot like a `TableFunction` but has some minor differences. This causes a number of issues:

* When implementing a multi-file reader function a subtly different API must be implemented
* Implementing a multi-file reader function is not possible outside of the C++ API

This PR takes a step towards solving that problem by introducing the `TableFunctionMultiFileWrapper`. This is essentially a `MultiFileReaderInterface` that wraps a table function, allowing a table function that can read a single file to be automatically wrapped in order to expose the unified multi-file reader interface.

As a showcase of this working we migrate `read_json` to this new interface. We introduce a new function - `read_single_json_file` - which is a mirror of `read_json` but without multi-file reading capabilities. The various `read_json` functions (`read_json`, `read_json_objects`, etc) are then implemented by wrapping `read_single_json_file` using the newly introduced method.

In order to facilitate this, a few changes are made to the `TableFunction` interface.

* `TableFunctionBindInput` now gets three new fields (`expected_names`, `expected_types`, `expected_bind_data`). This allows the binding of a follow-up scan to look at data from a previously executed scan, e.g. when we determine the result schema based on the first file this allows follow-up scans to adhere to that schema.
* Three new callbacks are introduced: `table_function_claim_batch_t` and `table_function_finish_batch_t` for more granular managing of the underlying scan, and `table_function_combine_schema_t` for custom schema combinations

The final goal of this is to move all `MultiFileReader` functions to this model, and to also expose this in the C API (likely through a single callback, e.g. `duckdb_register_multi_file_function(context, "read_single_json_file", "read_json", ".json")`) - which would then enable extensions that implement file formats to be migrated to this trivially (e.g. avro but also various other external extensions).


Mytherin:
CC @Maxxen

Maxxen:
Very cool. Can you elaborate on the "scan unit" concept? Thats new to me, but I suppose it maps to something in the MultiFileReader?

We've been thinking about adding "partition stats" to the C-API too - how does that interplay with a "scan unit" (isn't that just a "batch"?)

Mytherin:
Yes, the idea is to scan only a single batch from a file so that the multi file reader can provide batch indexes. I'll rename the function so that it matches the naming in other parts of the code.

PR #25472, opened 2026-09-08
Bind optimizer-introduced builtins through the catalog

kryonix:
I'm following up on [@Maxxen's suggestion in #25260](https://github.com/duckdb/duckdb/pull/25260#discussion_r3914898469) to resolve functions introduced by optimizer rewrites through the catalog. Currently, `s LIKE 'abc%'` and `prefix(s, 'abc')` can produce the same function call with different definition metadata: the explicitly bound call carries its catalog and schema, while the optimizer-generated call comes from `PrefixFun::GetFunction()` and has neither.

The catalog entry sets this qualification on the function definitions in its overload set, and bound functions inherit it from their definition. Several optimizer rewrites bypass that path by binding functions obtained directly from C++ factories. This affects string rewrites, introduced counts and aggregate-state combinations, struct extraction, and the manually constructed `constant_or_null` expressions.

I added shared helpers to resolve scalar and aggregate builtins in `system.main`, select an overload from the argument types, and bind the resulting definition. The fully qualified lookup keeps optimizer-generated calls independent of macros or user-defined functions on the search path. Partial aggregate pushdown also has a lookup variant that returns `nullptr` when no overload matches, preserving its existing fallback behavior.

`ExpressionRewriter::ConstantOrNull` now takes a `ClientContext` and uses the registered function's bind callback to derive the return type and bind data from the leading constant. This replaces the manual construction in the rewriter. The declared parameter types remain `ANY`; the bound return type still comes from the constant.

I kept the internal compression and decompression functions constructed directly. They are registered for serialization, but their bind callback rejects ordinary binding, so they cannot use this path. `variant_comparator`, which has no such restriction, now goes through the catalog.

There is still a separate qualification problem in `BoundAggregateFunction::ReplaceImplementation`: it copies the catalog and schema from the replacement function, which can itself come from a factory. For example, statistics propagation can replace `count` with an unqualified `count_star` even when the original call was bound through the catalog. This PR addresses the optimizer's initial construction of registered builtins; implementation replacement and factory binding elsewhere in the planner and executor remain follow-up work.

### DIFF
--- a/src/optimizer/builtin_function_lookup.cpp
+++ b/src/optimizer/builtin_function_lookup.cpp
@@ -32,6 +32,20 @@ shared_ptr<const AggregateFunction> GetBuiltinAggregateFunction(ClientContext &c
 	return entry.functions.GetFunctionByArguments(context, arguments);
 }
 
+shared_ptr<const AggregateFunction> TryGetBuiltinAggregateFunction(ClientContext &context, const Identifier &name,
+                                                                   const vector<LogicalType> &arguments) {
+	auto &catalog = Catalog::GetSystemCatalog(context);
+	auto &entry = catalog.GetEntry<AggregateFunctionCatalogEntry>(context, BuiltinName(catalog, name));
+
+	ErrorData error;
+	FunctionBinder function_binder(context);
+	auto index = function_binder.BindFunction(entry.functions.name, entry.functions, arguments, error);
+	if (!index.IsValid()) {
+		return nullptr;
+	}
+	return entry.functions.GetFunctionByOffset(index.GetIndex());
+}
+
 unique_ptr<BoundFunctionExpression> BindBuiltinScalarFunction(ClientContext &context, const Identifier &name,
                                                               vector<unique_ptr<Expression>> children) {
 	vector<LogicalType> arguments;
--- a/src/optimizer/builtin_function_lookup.cpp
+++ b/src/optimizer/builtin_function_lookup.cpp
@@ -1,0 +1,52 @@
+#include "duckdb/optimizer/builtin_function_lookup.hpp"
+
+#include "duckdb/catalog/catalog.hpp"
+#include "duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp"
+#include "duckdb/catalog/catalog_entry/scalar_function_catalog_entry.hpp"
+#include "duckdb/function/function_binder.hpp"
+#include "duckdb/planner/expression/bound_function_expression.hpp"
+
+namespace duckdb {
+
+namespace {
+
+//! Built-ins are always looked up fully qualified in the system catalog, so a macro or user-defined function of the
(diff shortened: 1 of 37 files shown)


src/optimizer/builtin_function_lookup.cpp
```diff
@@ -0,0 +1,66 @@
...
+		arguments.push_back(child->GetReturnType());
+	}
+	auto function = GetBuiltinScalarFunction(context, name, arguments);
+
+	FunctionBinder function_binder(context);
+	auto result = function_binder.BindScalarFunction(std::move(function), std::move(children));
+	if (result->GetExpressionClass() != ExpressionClass::BOUND_FUNCTION) {
+		throw InternalException("Optimizer exception - binding built-in function %s did not return a function", name);
+	}
+	return unique_ptr_cast<Expression, BoundFunctionExpression>(std::move(result));
```
Maxxen: Isn't this the same as `ScalarFunction::Bind()`? There is also AggegateFunction/WindowFunction::Bind

Maxxen:
Great! All functions should be retrieved from the Catalog - no more `XXXFun::GetFunction()`!

src/optimizer/builtin_function_lookup.cpp
kryonix: Absolutely right! This is gone now—Working on the other places.

lnkuiper:
Thanks! A few small comments:

1. Join-filter pushdown still constructs `min` and `max` directly from their factories. Both their executing functions and retained definitions have empty catalog/schema names. Could we move these through the new helper too?

2. The column-pruning test uses an explicit `count(*)`, which already gets its qualified definition during binding. It still passes with `unused_columns` disabled. Could we instead check the `count_star()` introduced when pruning an unused aggregate? For example:

   ```sql
   CREATE TABLE integers AS SELECT i::INTEGER AS i FROM range(100) t(i);
   SELECT 1 FROM (SELECT sum(i) FROM integers WHERE random() > 0.5) t;
   ```

3. Small cleanup: `DEBindCombineAggr` now duplicates `DEBindAggregate`; it could delegate using `CombineAggrFun::Name`. `ConstantOrNull::Bind(Value)` also has no remaining callers, so its declaration and definition can go.

lnkuiper:
Thanks! LGTM now

Mytherin:
Thanks - could you just resolve the merge conflicts?

Maxxen:
Semi-related - im trying to change the bound function de/serialization so that it also re-binds through the catalog from the raw argument expressions instead of trying to piece-together the already bound function from the serialized data - this helps a lot with that!

kryonix:
Very nice! I already have a follow-up for this PR staged, that refactors more places (there are a few in the decorrelation) similarly to what I've done here. With the de/serialization you are working on, this should cleanup a lot of stuff. I like it!

PR #25473, opened 2026-09-08
Replace `Value` with new `Literal` class in `ConstantExpression`

Maxxen:
This PR removes the `Value` from the parsed `ConstantExpression`, and replaces it with a untyped text-based `Literal` instead.

## Motivation

The parser currently decides the type of every literal. E.g. a numeric literal will be turned into a INTEGER, BIGINT, `HUGEINT`,  `UHUGEINT`, `BIGNUM`, `DECIMAL` or `DOUBLE` in the transformer. Similarly the parser also builds `VARCHAR` and `BLOB` values. This puts a `Value`, and therefore a `LogicalType`, into every constant of the parsed tree. 

This PR is another step toward a parse tree with no `LogicalType`, literals are kept as the source text that was parsed, and all typing decisions move into the binder where they are turned into `Value`s once the constant expression is bound into a `BoundConstantExpression`. This centralizes `LogicalType` creation, and enables us to perform literal binding with the presence of a `ClientContext`, potentially enabling custom literals or dialect-defined literal type inference rules in the future. (E.g. PostgreSQL supports custom literals for custom types).

### Internal changes

- The parse tree gains a new `Literal` type (src/include/duckdb/parser/literal.hpp), which holds a a kind (`NULL`, `BOOLEAN`, `INTEGER`, `NUMERIC`, `STRING`, `HEX`, `BIT`), plus the parsed source in textual form. Nested type literals,  (`[..]`, `{..}`, `(..)`, `MAP {..}`) is already desugared into function expressions and is therefore unaffected.
- Negation folds into the number text, which replaces the separate `TryNegateValue` and raw-text fast paths and keeps e.g. `-9223372036854775808` exact by construction.
- Similarly we also delete the `ConstructConstantFromExpression` - its only used for pivots, but pivots fold constants when binding anyway so this is not necessary in the parser (it was actually removed in v1.5 but then added back when the new parser was introduced by mistake)
- We add a `ConstantExpression::FromValue(const Value &value)` which enables "unbinding" a Value back into a literal again. We make some effort to construct function expressions for e.g. nested types, but fall back to calling `.ToString()` on the value and wrapping it in a `CastExpression` to the original type. This is also how we maintain backwards compatibility when reading old parsed constant expressions. There is also a `ConstantExpression(Value)` constructor to not break all extensions, but it should probably be deprecated in the future.
- Multi-file column-defaults now bind with a `ConstantBinder`, instead of trying to inspect the expression type and manually attempt to extract a Value from a ConstantExpression.

### Storage and serialization

This PR should be both forwards and backwards compatible when it comes to storage.

-  Field `200: value` becomes read-only and is written only when targeting a storage version before v2.0.0 (via `literal.ToValue()`, so older readers see what the old parser produced). Field `201: literal` is written from v2.0.0. Since literals basically model a subset of Value, we can always convert a `Literal` into a `Value` in a backwards compatible way.
- Deserialization uses the literal when present and falls back to `FromValue` for old files, so existing databases with views macros, defaults and struct remaps load unchanged.

### Behavior changes

- Parsed-constant equality is now textual: `01` and `1` are no longer equal `ConstantExpressions` (this should only affect `client_verify` dedup). 
- Rendering: SQL text regenerated from the tree keeps literals as written (`1_000`, `1e3`, `X'FF'`, `B'0101'`), and values that had to be wrapped render as casts (`CAST(5 AS BIGINT)`, `CAST('1970-01-01' AS DATE)`).
- `BIT` literals are now supported, e.g. `B'0101'` creates a `BIT` type.


Mytherin:
Thanks! Looks good - great idea

PR #25474, opened 2026-09-08
Add IMMEDIATE and DEFERRED constraint timing syntax

artjomPlaunov:
```sql
statement error
CREATE TABLE deferred_column_unique(i INTEGER UNIQUE DEFERRED);
----
Not implemented Error: Deferred UNIQUE constraints are not implemented yet

statement error
CREATE TABLE deferred_column_primary_key(i INTEGER PRIMARY KEY DEFERRED);
----
Not implemented Error: Deferred PRIMARY KEY constraints are not implemented yet
```

Add IMMEDIATE and DEFERRED to the grammar for UNIQUE and PRIMARY KEY constraints. DEFERRED throws not implemented for now; IMMEDIATE keeps the current behavior. Part 1 of implementing https://github.com/duckdblabs/duckdb-internal/issues/6885, subsequent PR will be the actual implementation for UNIQUE + PRIMARY keys.

Internally we keep three values: DEFAULT, IMMEDIATE, and DEFERRED. DEFAULT means no modifier was specified, and currently behaves the same as IMMEDIATE. We preserve the distinction so that we can change the default in the future without changing constraints that were explicitly declared IMMEDIATE. The declared mode is preserved when copying, rendering back to SQL, and serializing constraints.

For persistent tables, explicit IMMEDIATE requires storage version v2.0.0 or newer, since older formats cannot preserve the declared mode. We reject this up front rather than silently turn IMMEDIATE into DEFAULT when reopening the database. Existing older files can still be opened, and constraints without an explicit modifier continue to work with older storage versions. Temporary and in-memory tables do not have this storage restriction. DEFERRED still throws not implemented regardless of storage version.

Note: also fixes a regression in the parser that was allowing various ADD COLUMN <constraint>'s to parse, even though those constraints were not actually implemented and were silently ignored. Added parser errors for those, as well as a test for each one (each test case can be updated as that constraint is actually implemented for ADD COLUMN). I will backport this specific fix to 2.0 as well.

--------------------------------------

Note in the issue it was proposed to make DEFERRED the default (right now we have IMMEDIATE behavior by default). I think it makes sense to start out implementing DEFERRED as an opt-in at first, and after it is fully supported we can consider making it the default, perhaps in v3.0.

There are a lot of considerations here, but I think taken together they come down to allowing us to implement full DEFERRED support in a piece-meal manner before deciding on making the switch to having it be the default:

1) It changes existing semantics for test cases + code running transactions in a big way, since we would allow an inconsistent state w.r.t. constraints during transaction processing. This can be useful, but other databases such as Postgres also have deferred checking as an "opt-in" feature, not the default mode of processing. So this could potentially be confusing for people used to eager constraint checking semantics as a default.

2) We can implement the full DEFERRED feature set piece-meal, i.e. the subsequent PR will just start out implementing it for UNIQUE + PRIMARY KEYs, then we can do FKs, etc.

3) We keep the grammar simple at first: a constraint can explicitly specify IMMEDIATE or DEFERRED, or leave the modifier out to use the default. There are also some considerations around "toggling" of modes. There is some not so friendly looking SQL in the standard for setting things like DEFERRABLE INITIALLY DEFERRED, SET CONSTRAINTS, etc. This looks like a good opportunity for some more friendly SQL, but supporting toggling and figuring out a nicer grammar can be a future PR. This PR just records the declared mode; it does not add toggling.

4) Keep ON CONFLICT the same for now under IMMEDIATE behavior, and when we implement DEFERRED I think we can initially just reject using DEFERRED constraints as an arbiter.


src/parser/peg/transformer/transform_alter.cpp
```diff
@@ -287,8 +287,16 @@ AddColumnEntry PEGTransformerFactory::TransformAddColumnEntry(
 				}
 				new_column.default_value = std::move(constraint.expression);
 			} else if (constraint.constraint_name == "NotNullConstraint" &&
-			           constraint.constraint_type_info.second == ConstraintType::NOT_NULL) {
+			           constraint.constraint_type_info.type == ConstraintType::NOT_NULL) {
 				new_column.is_not_null = true;
+			} else if (constraint.constraint_name == "UniqueConstraint") {
```
Dtenwolde: These string comparisons ain't pretty. Perhaps we can use an enum instead? Ideally it's just `ConstraintType` that we can compare on, but I am not sure that's enough

Dtenwolde: Constraints are a bit messy in the transformer unfortunately...

Dtenwolde (commented):
Just a nit, but grammar wise looks good :)

artjomPlaunov:
Applied the fix for the nit. For more broader cleanliness/extending the grammar for other constraint types, I think will defer refactoring this until we come back and add timing for more constraints (maybe when adding FK), also just leaving your other suggestion here for reference when I come back to this:

> It looks good as is, just if you want this across all constraints, it's maybe easier to pull the ConstraintTiming up one level 
Then (just a rough sketch), this rule CreateTableConstraint <- TopLevelConstraint could be CreateTableConstraint <- TopLevelConstraint ConstraintTiming? instead

src/parser/peg/transformer/transform_alter.cpp
artjomPlaunov: > These string comparisons ain't pretty. Perhaps we can use an enum instead? Ideally it's just `ConstraintType` that we can compare on, but I am not sure that's enough

done, thanks!

artjomPlaunov:
@Dtenwolde applied the change, but re-running CI :-)

PR #25477, opened 2026-09-08
[async results] Add a unified QueryResult type

evertlammerts:
Stack (only land in order):
- [x] #25287 Adds a drain mode to the benchmark runner to distinguish between materializing and dropping chunks, and two regression tests
- [x] #25288 Enable the two new regression tests
- [x] #25257 Result buffer redesign
- [x] This PR
- [ ] Allow setting a notifier callback when submitting a query. The callback will be called when data is available or when a query errors, finishes or is interrupted.
- [ ] Expose async query execution in the C API v2

This PR reworks the result type, and introduces async query execution. It:
- Adds / adapts blocking and non-blocking query execution methods on `ClientContext`.
- Replaces the existing result types with a single `QueryResult`.
- Adds a `QueryResultStream` that can take ownership of a `QueryResult` and provides streaming semantics.
- Adds polling and non-blocking fetch methods.

### Blocking and Non-blocking Query Execution

`PendingQueryResult`, `MaterializedQueryResult`, and `StreamQueryResult` are replaced with a single, unified `QueryResult` type. You get a `QueryResult` by running a query, either through a family of non-blocking methods (`Submit`) or a family of blocking methods (`Query`):
- `ClientContext::Submit`: (non-blocking) submit a query or statement. The engine runs it iff `threads - external_threads > 0`, but produces no data until the caller either calls a materializing method on the result, or moves the result into a `QueryResultStream` (see below).
- `ClientContext::Query`: (blocking) runs a query or statement to completion, and returns the result containing all data materialized.

### `QueryResult`

A `QueryResult` provides:
- `QueryResultState Poll()`: (non-materializing) Returns the result's current state
- `QueryResultState ExecuteTask()`: (non-materializing) Executes a single task, not exactly like but comparable to `PendingQueryResult::ExecuteTask()`. It drives execution of the pipeline, but it will not settle the result buffer on its own. The caller must at some point either call e.g. `Materialize()` or move the result into a `QueryResultStream` and continue calling `ExecuteTask` and `TryFetch` on that.
- `void Materialize()`: (materializing, non-blocking) Lets the engine materialize the data. Note that if the engine does not have any threads of its own, the caller must still call one of the blocking methods below and participate in task execution.
- `void Complete()`: (materializing, blocking) Like `Materialize()`, but will block (and participate in execution) until execution is finished.
- `ColumnDataCollection &Collection()`: (materializing, blocking) Like `Complete()`, but will return a reference to the resulting CDC.
- `unique_ptr<ColumnDataCollection> TakeCollection()`: (materializing, blocking) Like `Collection()` but will pass ownership of the CDC to the caller.
- `unique_ptr<DataChunk> Fetch()`: (materializing, blocking) Fetches the next chunk. Will participate in execution.
- `unique_ptr<DataChunk> FetchRaw()`: (materializing, blocking) Like `Fetch()`, but will not normalize vector types.

### `QueryResultStream`

A `QueryResultStream` provides streaming semantics. It can be created by giving it ownership of a `QueryResult` that does not have materialized data yet (i.e. whose result sink buffer has not been decided yet).

The reason to introduce it, is that cdc-backed results do behave differently than buffer-backed results, because the latter is read-once while the former is not. For example, you _could_ support Materialize and TakeCollection on a stream, but they would be significantly slower since these operations would go through the streaming buffer. And Materializing a stream after reading one or two chunks is not the same as reading a few chunks from a CDC-backed buffer and then asking for the CDC.

Once created, a stream provides:
- `QueryResultState Poll()`: Returns the result's current state
- `QueryResultState ExecuteTask()`: Like `QueryResult::ExecuteTask()`. Can be used together with `TryFetch` to stream chunks out in a single-threaded application.
- `QueryResultState TryFetch(unique_ptr<DataChunk> &out_chunk)`: (non-blocking) Like `Poll()` but fetches a chunk if it's available.
- `unique_ptr<DataChunk> Fetch()`:  (blocking) Fetches the next chunk. Will participate in execution.


src/planner/binder/query_node/bind_trigger_expansion.cpp
```diff
@@ -720,7 +720,7 @@ BoundStatement Binder::ExpandRowTriggers(QueryNode &node, vector<unique_ptr<Pars
 
 	auto &properties = GetStatementProperties();
 	properties.return_type = StatementReturnType::CHANGED_ROWS;
-	properties.output_type = QueryResultOutputType::FORCE_MATERIALIZED;
+	properties.complete_on_return = true;
```
Mytherin: nit: enums please

src/main/connection.cpp
```diff
@@ -159,15 +142,11 @@ unique_ptr<PreparedStatement> Connection::Prepare(unique_ptr<SQLStatement> state
...
-	parameters.query_parameters.memory_type = QueryResultMemoryType::BUFFER_MANAGED;
-	auto pending = PendingQuery(query, parameters);
-	if (pending->HasError()) {
-		return make_uniq<MaterializedQueryResult>(pending->GetErrorObject());
-	}
-	return pending->Execute();
+	QueryParameters parameters;
+	parameters.statement_args = named_params;
+	parameters.memory_type = QueryResultMemoryType::BUFFER_MANAGED;
+	parameters.eager = true;
```
Mytherin: nit: enums please

src/include/duckdb/execution/executor.hpp
```diff
@@ -120,6 +123,15 @@ class Executor {
 	}
 	void UnregisterTask();
 
+	//! Set the buffer of the result this query produces. Called at submission, before execution starts
+	void SetResultBuffer(shared_ptr<BufferedData> result_buffer_p);
+	shared_ptr<BufferedData> GetResultBuffer();
+	//! Set the notifier of the result this query produces (may be null)
+	void SetResultNotifier(shared_ptr<QueryResultNotifier> result_notifier_p);
```
Mytherin: Can we move the notifier infrastructure to a separate PR stacked on top of this one?

Mytherin (commented):
Thanks for the PR! Looks good - some minor comments

evertlammerts:
@Mytherin thanks! I moved the notifier part out and will put up a follow up after this lands, and I added a `ResultEagerness` enum to replace the bools.

src/parallel/executor.cpp
```diff
@@ -488,26 +471,58 @@ PendingExecutionResult Executor::ExecuteTask(bool dry_run) {
...
+	lock_guard<mutex> l(executor_lock);
+	if (to_be_rescheduled_tasks.empty()) {
+		return QueryResultState::NO_TASKS_AVAILABLE;
+	}
+	// At least one task is blocked
+	if (ResultCollectorIsBlocked()) {
+		return QueryResultState::READY;
+	}
+	return QueryResultState::BLOCKED;
+}
```
carlopi, line 505:
```suggestion
QueryResultState Executor::IdleState() {
        lock_guard<mutex> l(executor_lock);
        if (!to_be_rescheduled_tasks.empty() && ResultCollectorIsBlocked()) {
                return QueryResultState::READY;
        }
        if (HasTaskInProgress()) {
                return QueryResultState::NOT_READY;
        }
        if (to_be_rescheduled_tasks.empty()) {
                return QueryResultState::NO_TASKS_AVAILABLE;
        }
        return QueryResultState::BLOCKED;
}
```
what about this?

With current code, one might potentially just go idle on `NO_TASKS_AVAILABLE` while holding a task in hand.

carlopi: Note, this is not really reachable right now from current callers, but still good to improve I think

carlopi: Or I guess there are 2 different questsion.

1. Should `READY` be checked first or later
2. Should `HasTaskInProgress()` be part of the decision or it's irrelevant

evertlammerts: I agree this is suboptimal, but for this PR I don't think it's blocking. I can address this a little more structurally in a follow up. The idea would be to stop re-deducing the state of the executor on each call, and instead set the state in a field on the executor as it goes through transitions. I think that can be a relatively isolated change. Plus it forces us to make clear which states there are and what they mean (right now, "ready" means multiple things and every state can be a terminal state). Does that sound like a plan @carlopi ?

Mytherin:
Thanks!

evertlammerts:
Closed and re-opened to trigger a new Main CI run

PR #25478, opened 2026-09-08
Fix correlated HAVING empty input handling

### LINKED ISSUE #24563: `HAVING` without `GROUP BY` in a **correlated** subquery is interpreted one way by `EXISTS` and the opposite way by the scalar form — the engine disagrees with itself about how many rows the subquery produced
### What happens?

`HAVING` without `GROUP BY` is a construct on which engines genuinely disagree, and this report
deliberately does **not** ask DuckDB to pick a side. The problem is that DuckDB picks **both sides
at once**, within a single statement: the `EXISTS` path treats the subquery as producing one row,
while the scalar path treats it as producing none.

### To Reproduce

```python
# python3.13, duckdb 1.5.5 and nightly 1.6.0.dev343
con.sql("CREATE TABLE ha(x INT); INSERT INTO ha VALUES (1),(2);")
con.sql("CREATE TABLE hb(y INT); INSERT INTO hb VALUES (1);")

con.sql("""
SELECT ha.x,
       EXISTS (SELECT 1 FROM hb WHERE hb.y = ha.x HAVING 1=1) AS e,
       (SELECT 1 FROM hb WHERE hb.y = ha.x HAVING 1=1)        AS v
FROM ha ORDER BY ha.x
""").fetchall()
# -> [(1, True, 1), (2, True, None)]
#                      ^^^^  ^^^^  contradictory: e says >=1 row, v says 0 rows
```
Either answer is defensible on its own; **the pair is not**:

- Under the standard reading (`HAVING` without `GROUP BY` implies `GROUP BY ()`, one group, one row
  even on empty input — SQL:2016 §7.10 + §7.9), the subquery produces one row whose value is the
  literal `1`, so **`v` must be `1`**.
- Under the MySQL-style reading (`HAVING` degenerates into a row filter, so the subquery produces
  zero rows for `x = 2`), **`e` must be `false`**.

An additional, purely local argument: the select list is the **literal `1`**, which cannot evaluate
to NULL under any reading. So `v = NULL` can only mean "zero rows", which directly contradicts
`e = true`.

### OS:

macOS

### DuckDB Version:

DuckDB 1.5.5(Python Package) && nightly 1.6.0.dev343.

### DuckDB Client:

Python

### Hardware:

_No response_

### Full Name:

Jiahao Ding

### Affiliation:

Xiamen University

- [x] Yes, I have

- [x] Yes, I have

Yes

YimingQiao:
Fixes #24563.

When flattening a correlated subquery with an ungrouped aggregate, correlated columns are added as grouping keys. For an empty input, the grouped RHS may produce no row, so reconnecting it with an `INNER JOIN` can be semantically different from the original ungrouped aggregate.

Using a `LEFT JOIN` is always safe, but may be less efficient. This PR keeps the `INNER JOIN` optimization only when the expressions above the aggregate guarantee that removing the empty-input row is unobservable; otherwise it falls back to `LEFT JOIN`.

Regression tests cover the original issue, join selection, volatile side effects, and throwing expressions.

### DIFF
--- a/src/include/duckdb/planner/expression_nullability.hpp
+++ b/src/include/duckdb/planner/expression_nullability.hpp
@@ -12,12 +12,19 @@
 #include "duckdb/common/table_index.hpp"
 #include "duckdb/common/vector.hpp"
 
+#include <functional>
+
 namespace duckdb {
 
 class ClientContext;
+class BoundColumnRefExpression;
 class Expression;
 class LogicalCTE;
 class LogicalOperator;
+
+//! Returns whether an expression becomes NULL when the provided column reference becomes NULL.
+bool ExpressionBecomesNull(const Expression &expr,
+                           const std::function<bool(const BoundColumnRefExpression &)> &column_becomes_null);
 
 //! Conservatively proves that an expression cannot be NULL at a logical operator's output.
 class NotNullExpressionAnalyzer {
--- a/src/optimizer/join_order/join_order_operator.cpp
+++ b/src/optimizer/join_order/join_order_operator.cpp
@@ -1,13 +1,12 @@
 #include "duckdb/optimizer/join_order/join_order_operator.hpp"
 
 #include "duckdb/common/assert.hpp"
-#include "duckdb/function/function.hpp"
 #include "duckdb/planner/expression/bound_columnref_expression.hpp"
 #include "duckdb/planner/expression/bound_comparison_expression.hpp"
 #include "duckdb/planner/expression/bound_conjunction_expression.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
 #include "duckdb/planner/expression/bound_operator_expression.hpp"
-#include "duckdb/planner/expression_iterator.hpp"
+#include "duckdb/planner/expression_nullability.hpp"
 
 namespace duckdb {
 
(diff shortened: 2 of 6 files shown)


lnkuiper:
Thanks for the PR! Looks correct, but I think we can simplify this a bit for maintainability:

1. Could we share the NULL-propagation check? `src/planner/subquery/delim_join_cte_rewriter.cpp` already has an `ExpressionNullPropagatesForChild`, and `src/optimizer/join_order/join_order_operator.cpp` has similar `ExpressionBecomesNull` logic. A shared helper would be great

2. We could clean up the `has_throwing_child ` recursion if the caller checks `IsVolatile() || CanThrow()` once but that would cause us to choose a few more `LEFT JOIN`s. Not required, just a thought.

3. Could we add regex plan assertions for the INNER/LEFT choice? The current tests check results and side effects, but would also pass with an always-LEFT implementation.

YimingQiao:
> Thanks for the PR! Looks correct, but I think we can simplify this a bit for maintainability:
> 
> 1. Could we share the NULL-propagation check? `src/planner/subquery/delim_join_cte_rewriter.cpp` already has an `ExpressionNullPropagatesForChild`, and `src/optimizer/join_order/join_order_operator.cpp` has similar `ExpressionBecomesNull` logic. A shared helper would be great
> 2. We could clean up the `has_throwing_child ` recursion if the caller checks `IsVolatile() || CanThrow()` once but that would cause us to choose a few more `LEFT JOIN`s. Not required, just a thought.
> 3. Could we add regex plan assertions for the INNER/LEFT choice? The current tests check results and side effects, but would also pass with an always-LEFT implementation.

Thanks for your suggestions. I have updated the code to share NULL-propagation check and add regex plan assertions. For the `has_throwing_child` recursion, We cannot simply check `CanThrow()` on the root expression because it marks expressions such as 0.2 * avg(...) as unsafe even though they cannot actually throw on NULL input, causing TPC-H Q17 to fall back to a LEFT JOIN and fail the existing CI test.

PR #25479, opened 2026-09-08
Reject SET/DROP NOT NULL,SET DEFAULT,ALTER TYPE on nested struct fields

### LINKED ISSUE #25456: Setting `NOT NULL` constraint on struct field doesn't work
### What happens?

See repro for details

### To Reproduce

```sql
memory D CREATE TABLE t(s STRUCT(a INTEGER, b INTEGER));
memory D INSERT INTO t VALUES ({'a': NULL, 'b': 1});
memory D ALTER TABLE t ALTER s.a SET NOT NULL;
memory D <-- it passes through, but should fail
```

### OS:

macos

### DuckDB Version:

070ce1f695bb17ef68f1fdd5ba20cad4a3ac6646

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

asxvi:
## Summary
`ALTER TABLE ... ALTER <col>.<field>` silently applied `SET/DROP NOT NULL`, `SET DEFAULT`, and `ALTER COLUMN TYPE` to the top-level column instead of the nested field, since the transformer only ever read the first path segment of the dotted column name.

## Fix
Throw a clear `NotImplementedException` instead of silently misapplying the operation to the wrong column, for all four affected operations, until real nested-field support exists for them.

The check lives in `DuckTableEntry` (where these operations actually execute), not the parser transformer, which only captures the full dotted path (`column_path`) onto the relevant `AlterTableInfo` classes and leaves the decision of what's supported to the catalog layer.

## Tests
`test/sql/alter/struct/alter_nested_struct_field_not_implemented.test` - covers all four operations erroring on a nested field, confirms the underlying row is left untouched, and confirms simple (non-nested) columns are unaffected by the guard.

Fixes #25456

### DIFF
--- a/src/include/duckdb/parser/parsed_data/alter_table_info.hpp
+++ b/src/include/duckdb/parser/parsed_data/alter_table_info.hpp
@@ -297,8 +297,7 @@ struct ChangeColumnTypeInfo : public AlterTableInfo {
 	LogicalType target_type;
 	//! The expression used for data conversion
 	unique_ptr<ParsedExpression> expression;
-	//! The full dotted column path as written (e.g. ["s", "a"] for "s.a"); empty when not a nested-field target.
-	//! Used only to detect (and reject) a nested-field target - 'column_name' remains the operative field.
+	//! Full dotted column path, e.g. ["s", "a"] for "s.a"
 	vector<Identifier> column_path;
 
 public:
@@ -325,8 +324,7 @@ struct SetDefaultInfo : public AlterTableInfo {
 	Identifier column_name;
 	//! The expression used for data conversion
 	unique_ptr<ParsedExpression> expression;
-	//! The full dotted column path as written (e.g. ["s", "a"] for "s.a"); empty when not a nested-field target.
-	//! Used only to detect (and reject) a nested-field target - 'column_name' remains the operative field.
+	//! Full dotted column path, e.g. ["s", "a"] for "s.a"
 	vector<Identifier> column_path;
 
 public:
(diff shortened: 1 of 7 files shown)


Tishj (changes requested):
Please make this change at the DuckTable level, not at the transformer 👍

Tishj (changes requested):
Thanks!
The changeset is bigger than I expected, but I believe this isn't detectable at a later point without this.

One thing though, you added a `default` which primary purpose is to enable forward-compatibility, so that a new feature doesn't break serializability until its used.
This however *always* adds `column_path` with a non-default value.

I don't know how interesting/impactful that is in practice though? @Mytherin 

To be safer, we could make it empty if size <= 1, so it actually gets omitted for non-nested paths. Potentially renaming it to make this condition clearer

src/include/duckdb/parser/parsed_data/alter_table_info.hpp
```diff
@@ -322,6 +324,8 @@ struct SetDefaultInfo : public AlterTableInfo {
 	Identifier column_name;
 	//! The expression used for data conversion
 	unique_ptr<ParsedExpression> expression;
+	//! Full dotted column path, e.g. ["s", "a"] for "s.a"
+	vector<Identifier> column_path;
```
Mytherin, line 328: Instead of just adding all of these extra fields can we remove `column_name` and always use `column_path`?

PR #25484, opened 2026-09-08
Improved caching behavior in MAD

sebastiaan-dev:
Resolves https://github.com/duckdblabs/duckdb-internal/issues/10484.

I am not sure about what the original change was that introduced the regression, so I went a bit on a wild-goose chase trying to find a way to improve MAD execution, where I found out a couple things.

### Page switching

Whenever a frame was sliding over a `page` (read `DataChunk`) boundary, `QuantileCursor` lookups triggered a `page` load whenever a row was requested which was currently not in the active `page`. This might not seem too bad, as we have a cache and therefore refreshing a `DataChunk` should be relatively cheap. However, because MAD may use row ids out of order we might have access requests that look like: `row i -> page A`, `row i+100 -> page B`, `row i+1 -> page A`, essentially bouncing from `page` to `page`.

To illustrate, you can separate out the existing MAD implementation into (i) first median, (ii) partition, (iii) second median phases. Which using the existing benchmark had a combined total of about **156 million** page misses, of which phase (iii) contributed most.

Given the above there seemed to be two options: either reduce the number of page misses, or make page misses cheaper. As there were quite a lot, I focused on the former. As a first experiment to reduce the number of page misses I used the existing `QuantileCompare` constructor. Instead of passing in one cursor accessor we pass in two, with each cursor having its own active `page`. This reduces bounding between pages on boundaries since the comparator now has its own page for `lhs` and `rhs`, reducing the total page misses to about **52 million**.

### Materializing deviations

The third phase previously used `QuantileCompare` with `std::nth_element`, this resulted in in the following access pattern:

```
compare row index L -> cursor lookup -> load value L -> abs(L - median)
compare row index R -> cursor lookup -> load value R -> abs(R - median)
```
The implementation of `std::nth_element` is not explicitly defined, but it accesses elements multiple times and does so out of order. Thus a significant amount of these lookups and computations happen again and again for the same values. 

So for the MST path, instead of accessing values on demand, we materialize the deviations before passing them to `std::nth_element`. The benefit here being that we now access the rows once, in the same order as their physical layout, preventing `page` bounces.

#### Skip list

The above works fine for MST, and improves the skip list path significantly. But for skip list we can go further as instead of using `std::nth_element`, we can use the fact that the skip-list is already ordered. This means that in the lower half, absolute deviations decrease as values approach the median. And in the upper half absolute deviations increase as values move away from the median. If you then reverse the lower half you get two non-decreasing sorted ranges. On these ranges we can use a binary partition search to find the MAD rank and save on materialization ([thanks leetcode](https://leetcode.com/problems/median-of-two-sorted-arrays/description/)).

### Performance

These changes led to a runtime reduction from **14.458s** to **0.421s** for the existing benchmark, while also having improved scaling w.r.t the frame size. For example, previously using `PRECEDING 500` for the same benchmark would have an execution time of **82.05s**, whereas now it takes about **0.721s**. 

As materialization and using the skip-list are already plenty of an improvement, the two cursor approach has been removed.

### DIFF
--- a/extension/core_functions/aggregate/holistic/mad.cpp
+++ b/extension/core_functions/aggregate/holistic/mad.cpp
@@ -1,97 +1,16 @@
 #include "core_functions/aggregate/holistic_functions.hpp"
 #include "core_functions/aggregate/quantile_state.hpp"
+#include "duckdb/common/helper.hpp"
 #include "duckdb/common/operator/abs.hpp"
 #include "duckdb/common/operator/cast_operators.hpp"
 #include "duckdb/common/operator/subtract.hpp"
 #include "duckdb/common/smaller_binary.hpp"
+#include "duckdb/common/typedefs.hpp"
 #include "duckdb/planner/expression.hpp"
 
 namespace duckdb {
 
 namespace {
-
-struct FrameSet {
-	inline explicit FrameSet(const SubFrames &frames_p) : frames(frames_p) {
-	}
-
-	inline idx_t Size() const {
-		idx_t result = 0;
-		for (const auto &frame : frames) {
-			result += frame.end - frame.start;
-		}
-
-		return result;
-	}
-
-	inline bool Contains(idx_t i) const {
-		for (idx_t f = 0; f < frames.size(); ++f) {
-			const auto &frame = frames[f];
-			if (frame.start <= i && i < frame.end) {
-				return true;
-			}
-		}
-		return false;
-	}
-	const SubFrames &frames;
-};
-
-struct QuantileReuseUpdater {
-	idx_t *index;
-	idx_t j;
-
-	inline QuantileReuseUpdater(idx_t *index, idx_t j) : index(index), j(j) {
-	}
-
-	inline void Neither(idx_t begin, idx_t end) {
-	}
-
-	inline void Left(idx_t begin, idx_t end) {
-	}
-
-	inline void Right(idx_t begin, idx_t end) {
-		for (; begin < end; ++begin) {
-			index[j++] = begin;
-		}
-	}
-
-	inline void Both(idx_t begin, idx_t end) {
-	}
-};
-
-void ReuseIndexes(idx_t *index, const SubFrames &currs, const SubFrames &prevs) {
-	//  Copy overlapping indices by scanning the previous set and copying down into holes.
-	//	We copy instead of leaving gaps in case there are fewer values in the current frame.
-	FrameSet prev_set(prevs);
-	FrameSet curr_set(currs);
-	const auto prev_count = prev_set.Size();
-	idx_t j = 0;
-	for (idx_t p = 0; p < prev_count; ++p) {
-		auto idx = index[p];
-
-		//  Shift down into any hole
-		if (j != p) {
-			index[j] = idx;
-		}
-
-		//  Skip overlapping values
-		if (curr_set.Contains(idx)) {
-			++j;
-		}
-	}
-
-	//  Insert new indices
-	if (j > 0) {
-		QuantileReuseUpdater updater(index, j);
-		AggregateExecutor::IntersectFrames(prevs, currs, updater);
-	} else {
-		//  No overlap: overwrite with new values
-		for (const auto &curr : currs) {
-			for (auto idx = curr.start; idx < curr.end; ++idx) {
-				index[j++] = idx;
-			}
-		}
-	}
-}
 
 //===--------------------------------------------------------------------===//
 // Median Absolute Deviation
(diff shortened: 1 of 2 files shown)


extension/core_functions/aggregate/holistic/mad.cpp
```diff
@@ -237,26 +201,53 @@ struct MedianAbsoluteDeviationOperation : QuantileOperation {
...
-
-			// Compute mad from the second index
-			ID indirect(data);
-
 			MAD mad(med);
 
-			MadIndirect mad_indirect(mad, indirect);
-			rdata[ridx] = interp.template Operation<idx_t, RESULT_TYPE, MadIndirect>(index2, result, mad_indirect);
+			if (gstate && gstate->HasTree()) {
+				deviations.clear();
```
hawkfish, line 208: Do we know that the old frame reuse with the MST is worse than the repeated frame enumeration here? I thought the evaluation of MAD was in our or Adrian's paper but I don't see it there.

hawkfish (commented):
This is really awesome! I'd just like to check the MST perf before we change it.

extension/core_functions/aggregate/holistic/mad.cpp
sebastiaan-dev: This is a good point which I did not realize. I changed MST because it simplified the code a bunch but this was maybe a bit too enthusiastic. I did not immediately find a noteworthy regression on the MST path, but also no speedup. So maybe it is desirable to leave the MST path unchanged?

hawkfish: I suspect that the original perf regression was from the paging work (which is why I didn't see it in my profiling - it was a while ago). Prior to that IIRC all the data was in memory, so paging wasn't a thing. 

I was wondering if allowing a cursor to have multiple paging chunks might be a Good Idea? They are used in other places, often with bi-directional scanning, so making that an implementation detail might lead to wider improvements?

hawkfish: > I did not immediately find a noteworthy regression on the MST path, but also no speedup. So maybe it is desirable to leave the MST path unchanged?

Well if it is no different, then simpler may be better!

lnkuiper:
Merge queue ran into an unrelated tidy issue, but I'm not sure if queueing it up again will help

sebastiaan-dev:
@lnkuiper, should I close this PR and open up another one then? Or wait for the tidy to be resolved?

lnkuiper:
I'm not sure how even the tidy issue shows up in the merge queue since it probably didn't show up in any pr. Let's wait for a bit

extension/core_functions/aggregate/holistic/mad.cpp
sebastiaan-dev: @hawkfish to get back on this in a little more detail, the query I used was:

```sql
ROWS BETWEEN mod(b * 47, 521) PRECEDING
         AND 100 - mod(b * 47, 521) FOLLOWING
```

This resulted in roughly the same performance, but to get a better view I also ran some other queries (generated by codex) which trigger the MST path. I'm not sure how representative they are for real-world workloads however:

### Significant overlap with periodic jump

```sql
CREATE TABLE rank_mst AS
SELECT
    b % 100 AS a,
    b,
    mod(b * 2, 521) AS p
FROM range(10000000) tbl(b);

SELECT sum(m)
FROM (
    SELECT mad(a) OVER (
        ORDER BY b
        ROWS BETWEEN p PRECEDING
                 AND 100 - p FOLLOWING
    ) AS m
    FROM rank_mst
) q;
```

### Long slides with significant overlap separated by jumps

```sql
CREATE TABLE rank_mst_blocks AS
SELECT
    b % 100 AS a,
    b,
    CASE WHEN (b // 1024) % 2 = 0 THEN 0 ELSE 500 END AS p
FROM range(10000000) tbl(b);

SELECT sum(m)
FROM (
    SELECT mad(a) OVER (
        ORDER BY b
        ROWS BETWEEN p PRECEDING
                 AND 100 - p FOLLOWING
    ) AS m
    FROM rank_mst_blocks
) q;
```

### Almost no overlap

```sql
CREATE TABLE rank_mst_alternating AS
SELECT
    b % 100 AS a,
    b,
    CASE WHEN b % 2 = 0 THEN 0 ELSE 100 END AS p
FROM range(10000000) tbl(b);

SELECT sum(m)
FROM (
    SELECT mad(a) OVER (
        ORDER BY b
        ROWS BETWEEN p PRECEDING
                 AND 100 - p FOLLOWING
    ) AS m
    FROM rank_mst_alternating
) q;
```

### Slowly expanding frame

```sql
CREATE TABLE rank_mst_width AS
SELECT
    b % 500 AS a,
    b,
    mod(b, 501) AS p
FROM range(10000000) tbl(b);

SELECT sum(m)
FROM (
    SELECT mad(a) OVER (
        ORDER BY b
        ROWS BETWEEN p PRECEDING AND CURRENT ROW
    ) AS m
    FROM rank_mst_width
) q;
```

### Two subframes

```sql
CREATE TABLE rank_mst_exclude AS
SELECT
    b % 500 AS a,
    b,
    mod(b, 501) AS p
FROM range(10000000) tbl(b);

SELECT sum(m)
FROM (
    SELECT mad(a) OVER (
        ORDER BY b
        ROWS BETWEEN p PRECEDING AND p FOLLOWING
        EXCLUDE CURRENT ROW
    ) AS m
    FROM rank_mst_exclude
) q;
```

## Results

MST frame behavior | Baseline | PR | Improvement
-- | -- | -- | --
Significant overlap with periodic jump | 17.460s | 10.236s | 41.4%
Long slides with significant overlap separated by jumps | 11.551s | 9.981s | 13.6%
Almost no overlap | 27.005s | 10.008s | 62.9%
Slowly expanding frame | 75.527s | 24.116s | 68.1%
Two subframes | 33.141s | 3.944s | 88.1%

PR #25485, opened 2026-09-08
Move more includes from src/include to cpp files

carlopi:
Cuts some 3-4% of include graph (once counting for dependencies), that translates to somewhat faster rebuilds on some changes to the codebase.

If this builds and passes CI, good to go.


lnkuiper:
Thanks! Could you look at the merge conflict? Feel free to mark ready to merge once done

PR #25487, opened 2026-09-08
Buffer managed indexes 4a PrefixHandle::{AppendByte, Append, Split}

artjomPlaunov:
(Part 4 https://github.com/duckdblabs/duckdb-internal/issues/10748 is for porting everything related to insertion in the ART, there will be several PRs, hence 4a).

The big change here is PrefixHandle::Split, it is mostly the same with one difference: we allocate the replacement node (node being inserted at the split point) before calling Split(), and now let Split() install it + handle gate status installation internally. This simplified the code somewhat with PrefixHandles. I walked through all possible permutations of the logic; note that afterwards, the caller still installs the returned value in the replacement node at the split byte. 

### Prefix split cases examples

`[]` = prefix bytes, `R` = replacement Node4, `*` = gate (note that * means the pointer to the object has the gate set, not the node itself), `T` = unchanged subtree. Prefix capacity is 4 for illustration.

```text

CASE                 BEFORE                   AFTER SPLIT                          AFTER CALLER ATTACHES CHILD

Middle byte (b)       [abcd] -> T              Tree:   [a] -> R                     [a] -> R --b--> [cd] -> T
                                              Return: [cd] -> T

First byte (a)        [abc] -> T               Tree:   R                            R --a--> [bc] -> T
                                              Return: [bc] -> T
                                              Original prefix freed.

Last, partial (c)    [abc] -> T               Tree:   [ab] -> R                    [ab] -> R --c--> T
                                              Return: T
                                              No suffix allocated.

Last, full (d)       [abcd] -> T              Tree:   [abc] -> R                   [abc] -> R --d--> T
                                              Return: T
                                              No suffix allocated.

Only byte (a)        [a] -> T                 Tree:   R                            R --a--> T
                                              Return: T
                                              Original freed; no suffix.


GATED ORIGINAL PREFIX — these prefix bytes belong to the row-ID key

First byte (a)       *[abc] -> T              Tree:   *R                           *R --a--> [bc] -> T
                                              Return: [bc] -> T
                                              Original freed; gate transferred.

Middle byte (b)      *[abc] -> T              Tree:   *[a] -> R                    *[a] -> R --b--> [c] -> T
                                              Return: [c] -> T
                                              Gate stays on original pointer.

Last, partial (c)    *[abc] -> T              Tree:   *[ab] -> R                   *[ab] -> R --c--> T
                                              Return: T
                                              Gate stays; no suffix.

Last, full (d)       *[abcd] -> T             Tree:   *[abc] -> R                  *[abc] -> R --d--> T
                                              Return: T
                                              Gate stays; no suffix.

Only byte (a)        *[a] -> T                Tree:   *R                           *R --a--> T
                                              Return: T
                                              Original freed; gate transferred.


SUFFIX CONTINUATION

Ordinary, at b       [abcd] -> [efgh] -> T    Tree:   [a] -> R                     [a] -> R --b--> [cdef] -> [gh] -> T
                                              Return: [cdef] -> [gh] -> T
                                              Old [efgh] allocation freed.

Gated, at b          [abcd] -> *[xy] -> T     Tree:   [a] -> R                     [a] -> R --b--> [cd] -> *[xy] -> T
                                              Return: [cd] -> *[xy] -> T
                                              Gated node preserved.
```


artjomPlaunov:
note @taniabogatsch New and NewInternal are just moved over to PrefixHandle, no changes there, these were already rewritten in the last PR but I forgot to move them over. 

Just AppendByte, Append, and Split need to be reviewed.

src/execution/index/art/art_merger.cpp
taniabogatsch: minor nit: how come this is called `cast_pos`? Shouldn't this be `split_pos` or something?

src/execution/index/art/art_merger.cpp
```diff
@@ -279,16 +280,13 @@ void ARTMerger::MergePrefixes(NodeEntry &entry) {
 		const auto r_byte = Prefix::GetByte(art, entry.right, cast_pos);
 
 		// Split and reduce.
-		reference<NodePtr> left_ref(entry.left);
-		NodePtr l_child;
-		const auto status = Prefix::Split(art, left_ref, l_child, cast_pos);
+		NodePtr replacement;
```
taniabogatsch: Here, I am also not fully happy with the name `replacement`, maybe `branching_node4` or something would be more descriptive?

src/execution/index/art/prefix_handle.cpp
```diff
@@ -1,11 +1,136 @@
...
+	}
+	prefix.Child(art).Clear();
+	return prefix;
+}
+
+PrefixChain PrefixHandle::New(ART &art, const ARTKey &key, const idx_t depth, const idx_t count) {
+	D_ASSERT(count > 0);
+
+	NodePtr root;
+	auto first_count = UnsafeNumericCast<uint8_t>(MinValue<idx_t>(art.PrefixCount(), count));
```
taniabogatsch: nit: since we call it `root`, shouldn't we call this `root_count`? Or we call it `first`, and then `first_count`?

src/execution/index/art/prefix_handle.cpp
```diff
@@ -1,11 +1,136 @@
...
+	auto prefix = NewInternal(art, root, key.data, first_count, depth);
+	auto tail = std::move(prefix).IntoChild(art);
+
+	idx_t offset = first_count;
+	while (offset < count) {
+		auto this_count = UnsafeNumericCast<uint8_t>(MinValue<idx_t>(art.PrefixCount(), count - offset));
+		auto next = NewInternal(art, tail.Get(), key.data, this_count, depth + offset);
+		tail = std::move(next).IntoChild(art);
```
taniabogatsch: I had slight difficulties following the code here without checking out the PR and navigating through it a bit. I was mislead assuming that we would return the entire child node handle, but we just pin the current one and then return the ref to the child node pointer, no?

```cpp
	//! Returns the child NodePtr. The reference is valid while this handle owns the prefix pin.
	NodePtr &Child(const ART &art) {
		return ChildRef(art, handle);
	}

	//! Transfer this prefix's pin to a handle for its child NodePtr storage location.
	NodePtrHandle IntoChild(const ART &art) && {
		return NodePtrHandle(Child(art), std::move(handle));
	}

	//! Get a mutable reference to the child NodePtr of the prefix.
	static NodePtr &ChildRef(const ART &art, NodeHandle &handle) {
		return *reinterpret_cast<NodePtr *>(handle.GetPtr() + art.PrefixCount() + 1);
	}
```

EDIT:

Or I guess what I find scary here is that I only now get that we store the handle to the parent (current) (pin it), but that's because we want to alter the ref? I guess, again, why do we need anything other than `NodeHandle` for that? Ah, so we can make changes to the child ptr while owning it 🤔  

Ahhh, and in the [previous PR](https://github.com/duckdb/duckdb/pull/23619) we called it `SlotHandle`, which I also found confusing 💀 

How are you feeling about `ChildPtrHandle`, then? Would that be more intuitive? And maybe instead of `IntoChild` we call it `GetChildPtrHandle`? Or will I be confused again at the next PR? 😅

src/execution/index/art/prefix_handle.cpp
```diff
@@ -1,11 +1,136 @@
...
+		memcpy(prefix.Data(), data + offset, count);
+	}
+	prefix.Child(art).Clear();
+	return prefix;
+}
+
+PrefixChain PrefixHandle::New(ART &art, const ARTKey &key, const idx_t depth, const idx_t count) {
+	D_ASSERT(count > 0);
+
+	NodePtr root;
```
taniabogatsch: Something I've noticed more generally is that we've now renamed `Node` to `NodePtr`, however, out naming conventions for variables don't really follow that yet. Should we try to name `root` `root_ptr` instead? To keep the semantics? And handles can similarly also have a `_handle` suffix... WDYT? I think the main problem with ART code to me sometimes is readability, and I know what a function is supposed to do already 😅 doesn't necessarily have to happen in this PR, but maybe you also think its a good idea.

src/execution/index/art/prefix_handle.cpp
taniabogatsch: I wonder if we are overdoing it. If we need to construct a long prefix chain, then we could also just throw all the node handles in a vector or so, and throw the vector away when we're done? Then, we keep the nodes in which to change the child pointers pinned, and don't have to do these super fine-grained lifetimes? A bit similar to how we just move the handle lifetime inside an extra `{ }` block in `Append`.

taniabogatsch: But maybe there are other places where we need the current granularity?

src/execution/index/art/prefix_handle.cpp
```diff
@@ -1,11 +1,136 @@
...
+			prefix.Child(art) = next;
+		}
+
+		NodePtr::FreeNode(art, other);
+		other = next;
+	}
+	prefix.Child(art) = other;
+}
+
+NodePtr PrefixHandle::Split(ART &art, NodePtr &node, NodePtr &replacement, const uint8_t pos) {
```
taniabogatsch: Again, maybe we can rename `replacement`. And `node` is the prefix that we're splitting?

src/include/duckdb/execution/index/art/art_operator.hpp
```diff
@@ -404,14 +404,12 @@ class ARTOperator {
 		const auto cast_pos = UnsafeNumericCast<uint8_t>(pos);
 		const auto byte = Prefix::GetByte(art, node_ref, cast_pos);
 
-		NodePtr child;
-		const auto split_status = Prefix::Split(art, node_ref, child, cast_pos);
+		NodePtr replacement;
```
taniabogatsch: Replacement nit again.

src/include/duckdb/execution/index/art/prefix_handle.hpp
```diff
@@ -44,6 +44,14 @@ class PrefixHandle {
 	//! Create a non-empty prefix chain and return its root pointer and pinned final child location.
 	static PrefixChain New(ART &art, const ARTKey &key, const idx_t depth, const idx_t count);
 
+	//! Splits the prefix at pos. replacement must identify an already allocated, ungated node.
```
taniabogatsch: replacement, node.

taniabogatsch (commented):
Hi @artjomPlaunov, thanks for the PR! I mostly left nits on naming on where I think we can improve the readability of a complex code path. The logic itself looks good to me. 

I think the main thing to re-discuss is the `IntoChild` (slots) that were introduced in the previous PR, and how we either make them a bit more intuitive to understand, or revisit them in general.

src/execution/index/art/prefix_handle.cpp
artjomPlaunov: First, yes I agree with all the renames. 

1) ChildPtrHandle - I think this point is a little more difficult to resolve, so we should make sure we get this right :-) 

I think what is unavoidable to encode in whatever abstraction we pick is a <pinned node, pointer location within that pinned node> abstraction (Note this isn't entirely true, I have another idea down below). Previously we never had to encode a pinned node whenever we had a pointer location, we could just refer to any pointer location. The alternative is to basically follow the same model as before and assume we have a pinned node (i.e. a caller could create a NodeHandle) and pass NodePtr & refs around. 

We have a lot of places in the code where we have a NodePtr &, so this abstraction is to just be able to work with that in the same way, the difference being that a caller needs to create a node handle + convert that to a ChildPtrHandle (i.e. really the same as a NodeHandle but encoded with location information for which child pointer location we are referring to. 

The other idea would be to make it not memory specific at all, i.e. we can also just have a <NodePtr, logical child> abstraction that is "unpinned", and needs to be pinned whenever we want to refer to the child pointer. But I don't like this as much, since before we basically were assuming we could just pass in a NodePtr & to be modified in place, which is what I am trying to retain functionally. 

Either way, I think we need something that encodes which child we are actually referring to, since the only alternative would be to just assume the memory location is referring to a pinned node, which I don't think is safe?

2) Prefix chains 

With a vector, I am worried it would be easy to create a malformed scenario where we pin a very long chain of prefix nodes and run out of memory? 

However I think a simpler approach could be to just keep the current tail pinned while attaching the next prefix, and maybe simplify the code a bit here.

artjomPlaunov: hmm nevermind, maybe the scoping + NodePtr & is fine, will think about this some more.

artjomPlaunov: hmm, maybe we may need it in other places, but I think you're right that I was overdoing it at least for the current changes. I think my new approach will be to see how far I can take the scoped approach with NodePtr& within a scope as far as possible, since that is probably much easier to read. If it really becomes warranted for a complicated operation (maybe like merging, I think), we could figure out the proper abstraction just local to the special operations that need more care. 

I think we can easily fix the prefix traversal by just keeping a PrefixHandle node while we iterate, instead of storing a vector. 

Thanks :-)

src/execution/index/art/prefix_handle.cpp
artjomPlaunov: yeah makes sense. I was thinking about this in the rename PR, but thought I would defer it until it became an issue for readability. I think we can do that in the next PR :-)

artjomPlaunov:
@taniabogatsch thanks for the review! 

You were right about overdoing it with the NodePtrHandle stuff; I think the best way to go forward is to just use the normal RAII scoping/NodePtr & refs within those scopes as much as possible. If there are some specific scenarios that warrant something like a NodePtrHandle, then it can be figured out then/contained to some specific operations if it's actually necessary. 

For some reason I got into my mind that we want to be really "safe" when referring to a child pointer location, but I think this is totally fine and makes the code easier to read. The prefix chain construction was handled simply as well by just keeping the current tail pinned. I still kept around the simple "PrefixChain" struct which is a root node and pinned tail at the end of construction; I prefer this to a vector of pinned nodes, but the traversal doesn't have the weird ChildPtrHandle code anymore. 

Also applied all the renames local to this PR. 

So this PR is ready to review again :-)

artjomPlaunov:
@taniabogatsch also, for the variable renames outside the scope of this PR (i.e. changing node -> node_ptr, root -> root_ptr; I will prepare a PR to do that after the current sequence of PRs up to insertion is complete, so please bear with those variable names just for a few small PRs).

src/execution/index/art/prefix_handle.cpp
taniabogatsch: Yes, let's start to roll it in in the next PRs, and later we can ask 🤖 to catch places we don't have it yet.

taniabogatsch:
> do that after the current sequence of PRs up to insertion is complete

yes, sounds good, no rush on that :)

And on the other big topic around `NodePtrHandle`, happy to hear that you think it's feasible to revisit at a later point, if we really think that it's necessary. And good that we explored it 👍

taniabogatsch (approved):

PR #25489, opened 2026-09-08
More Sqllogictest cleanup /2

carlopi:
Stacked on top of https://github.com/duckdb/duckdb/pull/25488, adds primitives to allow unittester to run in parallel in the same binary, by moving away from global into either class members or thread_local globals.

Minor refactor, should be a nop for unittester, but allows to reuse the same code as loadable extension.

### DIFF
--- a/test/helpers/test_helpers.cpp
+++ b/test/helpers/test_helpers.cpp
@@ -569,6 +569,18 @@ bool DeleteTestPath() {
 
 static bool emit_test_events = false;
 
+static std::function<void(DuckDB &)> static_extension_loader;
+
+void SetStaticExtensionLoader(std::function<void(DuckDB &)> loader) {
+	static_extension_loader = std::move(loader);
+}
+
+void LoadStaticExtensions(DuckDB &db) {
+	if (static_extension_loader) {
+		static_extension_loader(db);
+	}
+}
+
 void SetEmitTestEvents(bool emit) {
 	emit_test_events = emit;
 }
--- a/test/helpers/test_helpers.cpp
+++ b/test/helpers/test_helpers.cpp
@@ -48,9 +48,17 @@ static string local_temp_dir_root_override; // --local-temp-dir-root ("" -> the 
 // set, not one per root: a remote root is env-var passthrough only -- unittest holds no credentials, so
 // it can never mkdir or reap one, and the local tree is the only thing with a lifecycle.
 static vector<string> run_created_levels;  // $ROOT..$RUN_ID
-static vector<string> test_created_levels; // $TEST_ID
 static vector<string> home_created_levels; // the HOME sandbox sibling
-static string active_test_leaf;            // currently-materialized $TEST_ID dir ("" when none)
+
+//! The $TEST_ID dir a test materialized, and the levels its path created. Keyed by TEST_ID rather
+//! than held in one global: test files can run concurrently, and every thread working on one test
+//! (a concurrentloop spawns several) has to find that test's entry.
+struct ActiveTestTempDir {
+	string leaf;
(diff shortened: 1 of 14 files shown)


smvv:
> allow unittester to run in parallel in the same binary

what is the end goal of making this binary run tests in parallel?

I think this could conflict with the test runner script:
- incorrect test durations,
- increased peak memory usage, 
- too many concurrent tests freezing the CI instance

carlopi:
> > allow unittester to run in parallel in the same binary
> 
> what is the end goal of making this binary run tests in parallel?
> 
> I think this could conflict with the test runner script:
> 
> * incorrect test durations,
> * increased peak memory usage,
> * too many concurrent tests freezing the CI instance

I didn't articulate clearly, the goal is not changing the unittest itself, that works fine as is, but allow multiple instance of the unittest to work side-by-side for an extension that would allow to do:
```sql
INSTALL AND LOAD sqllogictest;
.timer on
FROM run_tests('test/sql/types/uuid/');  --- run locally available tests
```
```
     ┌──────────────────────────────────────────────────────────────────────────────────────────────────────────┬─────────┬─────────┬───────────────────┬───────────────────┬────────────────────┬────────────┬─────────────┐
     │                                                   name                                                   │ status  │ message │ statements_passed │ statements_failed │ statements_skipped │ assertions │ duration_ms │
     │                                                 varchar                                                  │ varchar │ varchar │      uint64       │      uint64       │       uint64       │   uint64   │   double    │
     ├──────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────┼─────────┼───────────────────┼───────────────────┼────────────────────┼────────────┼─────────────┤
     │ ./test/sql/types/uuid/test_uuid.test_slow          │ SUCCESS │ NULL    │                45 │                 0 │                  0 │        123 │  172.118167 │
     │ ./test/sql/types/uuid/test_uuid_cast.test          │ SUCCESS │ NULL    │                10 │                 0 │                  0 │         10 │    7.279583 │
     │ ./test/sql/types/uuid/test_uuid_uhugeint_cast.test │ SUCCESS │ NULL    │                12 │                 0 │                  0 │         12 │    9.148584 │
     └──────────────────────────────────────────────────────────────────────────────────────────────────────────┴─────────┴─────────┴───────────────────┴───────────────────┴────────────────────┴────────────┴─────────────┘
     Run Time (s): real 0.207 user 0.132500 sys 0.030221
```

This PR in particular would allow to bolt parallelism on top:
```sql
FROM run_tests('test/sql/types/');  --- run locally available tests, 464 of them
--- 18.29s (single sample, there is randomness)
FROM run_tests('test/sql/types/', workers:=8);  --- run locally available tests, 464 of them, up to 8 at a time
--- 4.85 (single sample, there is randomness, but some parallelism kicks in)
```

If you want to give a review in particular to https://github.com/duckdb/duckdb/pull/25489/changes/2954415f861199fd174a99641133b8339b8b19a7, that's the commit that deals with parallelism, very welcome.

Mytherin:
I think we shouldn't have multiple ways of running tests in parallel - but I don't really feel strongly about where the parallelism lives. @carlopi @smvv can you decide on the approach here?

smvv (approved):
we discussed it in the office; let's merge this PR.

I would like to explore [rr chaos mode](https://robert.ocallahan.org/2016/02/introducing-rr-chaos-mode.html) in the (near?) future to discover threading/scheduling issues and make them reproducible by replaying the recorded state.

At some point (`2.1+`), we would also like to look into moving the unittest binary into an extension so we dont have to build/ship the unittest binary to run tests (locally or in CI).

PR #25490, opened 2026-09-08
Close the main WAL handle before renaming over it during WAL recovery

### LINKED ISSUE #24767: Windows: WAL recovery after crash mid-checkpoint always fails with "Could not move file: Access is denied" (main WAL handle still open during recovery rename)
### What happens?

If the host crashes (power loss / hard reset) while a checkpoint is in progress, the database is left with both `<db>.wal` and `<db>.wal.checkpoint` on disk — a state DuckDB's recovery path is designed to handle. On **Windows**, however, that recovery deterministically fails on every subsequent read-write open with:

```
_duckdb.IOException: IO Error: Could not move file: Access is denied.
```

and the database can never be opened read-write again until the files are repaired manually. For a service that auto-restarts, this is a permanent crash loop (we hit it twice in production; the second time it kept a data-ingestion service down for 25 hours).

**Root cause (from reading v1.5.2 and current `main` — the code is identical in both):**

In `src/storage/wal_replay.cpp`, when the main WAL contains a checkpoint marker but the checkpoint did not complete, `WriteAheadLogReplayer::ReplayLog` calls:

```cpp
MergeIntoRecoveryWAL(con, checkpoint_state, reader, recovery_path, std::move(checkpoint_handle));
```

`MergeIntoRecoveryWAL` copies the main WAL + checkpoint WAL into `<db>.wal.recovery`, syncs it, closes the recovery handle, and then:

```cpp
fs.MoveFile(recovery_path, main_wal_path);
```

At this point `reader` — the `BufferedFileReader` over the **main WAL**, passed in by reference from `ReplayLog` — still holds an open handle to `<db>.wal`. `LocalFileSystem::MoveFile` on Windows uses `MoveFileExW(MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)`, which fails with `ERROR_ACCESS_DENIED` when the target file has an open handle. POSIX `rename(2)` allows replacing an open file, which is why this never shows up on Linux/macOS CI.

Because the failure happens before any state is changed, every retry re-runs the same merge and fails the same way — the `.wal.recovery` file is fully written and re-created on each attempt (its mtime refreshes every open), but the final rename can never succeed.

**Supporting evidence from the production incident** (byte sizes line up with the merge):

```
hot.duckdb.wal             16,776,311
hot.duckdb.wal.checkpoint     109,965
hot.duckdb.wal.recovery    16,886,216   (= wal + checkpoint − 60-byte version header)
```

Opening the same database `read_only=True` in this state works fine (recovery merge is skipped for read-only connections), which further isolates the failure to this rename.

**Suggested fix:** close/reset the main-WAL `BufferedFileReader` (and its underlying handle) before `fs.MoveFile(recovery_path, main_wal_path)` in `MergeIntoRecoveryWAL`, the same way `recovery_handle.reset()` is already done a few lines earlier. The reader is only needed again after the move to replay the merged WAL, which is reopened from `main_wal_path` anyway.

**Workaround** (zero data loss, works because all handles are closed once the failed `connect()` returns): finish the rename externally — move `<db>.wal.recovery` over `<db>.wal`, delete `<db>.wal.checkpoint`, and reconnect.

### To Reproduce

The precondition is a hard process/host kill *while a checkpoint with concurrent writes is in progress*, so the trigger is timing-dependent, but the failing state and code path are fully deterministic once the files exist:

1. On Windows, open a database and build up a non-trivial WAL with many small transactions (bulk inserts largely bypass the WAL): e.g. `SET checkpoint_threshold='1TB'`, then thousands of single-row `INSERT`s.
2. Start a `CHECKPOINT` while another connection keeps writing (so the secondary `<db>.wal.checkpoint` WAL is created and written to — see `StorageManager::WALStartCheckpoint`), and hard-kill the process (`os._exit` / power loss) before the checkpoint completes.
3. The directory now contains `<db>.wal` (with a checkpoint marker inside) and a non-empty `<db>.wal.checkpoint`.
4. Reopen the database read-write:

```python
import duckdb
con = duckdb.connect("crash.db")   # IO Error: Could not move file: Access is denied.
```

Every subsequent read-write open fails identically; `<db>.wal.recovery` is (re)created on each attempt. `duckdb.connect("crash.db", read_only=True)` succeeds.

We observed this twice in production (2026-08-11 and 2026-08-13), both times immediately after an unclean host shutdown recorded in the Windows event log, with the file states described above.

### OS:

Windows 11 Home 10.0.26200, x86_64

### DuckDB Version:

1.5.2 (failure observed); the relevant code in `src/storage/wal_replay.cpp` / `src/common/local_file_system.cpp` is unchanged on v1.5.5 and current main

### DuckDB Client:

Python (duckdb 1.5.2, Python 3.12)

### Hardware:

_No response_

### Full Name:

bryanliew-heptix (GitHub handle)

### Affiliation:

Independent

Not applicable - the reproduction does not require a specific data set

- [x] Yes, I have

- [x] Yes, I have

carlopi:
Fixes #24767.

After a crash mid-checkpoint the database is left with both <db>.wal and <db>.wal.checkpoint. Recovery merges them into <db>.wal.recovery and renames that over the main WAL, but WriteAheadLogReplayer::ReplayLog still holds the main WAL open through its BufferedFileReader while MergeIntoRecoveryWAL calls MoveFile(recovery_path, main_wal_path). POSIX rename(2) replaces an open file, so this never surfaces in CI; Windows MoveFileExW refuses, and because nothing has changed when it fails, every retry fails identically - the database can never be opened read-write again without manual repair.

The "checkpoint was successful" branch of ReplayLog has the same defect: it closes the source checkpoint handle before MoveFile(checkpoint_wal, wal_path) but leaves the target, the main WAL, open through the same reader.

Close the reader's handle at both moves. It is only needed again after the rename, and both paths reopen the WAL from its path anyway.

### DIFF
--- a/src/storage/wal_replay.cpp
+++ b/src/storage/wal_replay.cpp
@@ -382,6 +382,8 @@ void WriteAheadLogReplayer::MergeIntoRecoveryWAL(Connection &con, const ReplaySt
 	// move over the recovery WAL over the main WAL
 	recovery_handle->Sync();
 	recovery_handle.reset();
+	// the main WAL is the target of the move - Windows refuses to replace a file that is still open
+	main_wal_reader.handle.reset();
 
 	if (debug_checkpoint_abort == CheckpointAbort::DEBUG_ABORT_BEFORE_MOVING_RECOVERY) {
 		throw FatalException("Checkpoint aborted before moving recovery file because of PRAGMA checkpoint_abort flag");
@@ -496,7 +498,10 @@ unique_ptr<WriteAheadLog> WriteAheadLogReplayer::ReplayLog(unique_ptr<FileHandle
 				}
 				// if this is not a read-only connection we need to finish the checkpoint
 				// overwrite the current WAL with the checkpoint WAL
+				// both files must be closed - Windows refuses to move a file that is still open, and refuses to
+				// replace a target that is still open
 				checkpoint_handle.reset();
+				reader.handle.reset();
 
 				fs.MoveFile(checkpoint_wal, wal_path);
 


dentiny:
Hi @carlopi I'm wondering if you think reasonable to update `MoveFile` windows implementation with `FileRenameInformationEx` and `FILE_RENAME_POSIX_SEMANTICS` -- haven't tried on windows, but the goal is to provide same move file semantics as unix platform?
- It'd be nice to keep semantics consistency across different platforms
- There could be other unsurfaced places suffering the same issue

Reference: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/4217551b-d2c0-42cb-9dc1-69a716cf6d0c

carlopi:
> Hi @carlopi I'm wondering if you think reasonable to update `MoveFile` windows implementation with `FileRenameInformationEx` and `FILE_RENAME_POSIX_SEMANTICS` -- haven't tried on windows, but the goal is to provide same move file semantics as unix platform?
> 
> * It'd be nice to keep semantics consistency across different platforms
> * There could be other unsurfaced places suffering the same issue
> 
> Reference: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/4217551b-d2c0-42cb-9dc1-69a716cf6d0c

I am not really sure, yes, could semplify logic, but unsure what's the support surface, basically, will the flag be respected or this will pop up in some more remote config?

@staticlibs might have answer / preferences

dentiny:
> I am not really sure, yes, could semplify logic, but unsure what's the support surface, basically, will the flag be respected or this will pop up in some more remote config?

I could do some experiments later today :)

staticlibs:
Hi!

> I'm wondering if you think reasonable to update MoveFile windows implementation with FileRenameInformationEx and FILE_RENAME_POSIX_SEMANTICS -- haven't tried on windows, but the goal is to provide same move file semantics as unix platform?

Also never used this flag, I think it is indeed reasonable to try it. It is a relatively recent addition, I would think it was added to cover some cases similar to this one with WAL.

dentiny:
I'm playing with windows rename API in https://github.com/dentiny/duckdb/pull/191, technically seems possible; plan to create a PR after code cleanup.

PR #25492, opened 2026-09-08
Set unary arg property for last_day function

dentiny:
Hi team, I found `last_day` function doesn't have arg property set, so queries don't prune row groups as expected; for example,
```sql
memory D ATTACH '/tmp/last_day_prune.duckdb' AS last_day_db (ROW_GROUP_SIZE 2048);
memory D CREATE TABLE last_day_db.t AS
         SELECT CASE WHEN range < 2048 THEN DATE '2025-01-15'
                     WHEN range < 4096 THEN DATE '2025-02-15' ELSE DATE '2025-03-15' END AS d
         FROM range(6144);

-- should only access one row group, but access all
memory D EXPLAIN ANALYZE SELECT d FROM last_day_db.t WHERE last_day(d) = DATE '2025-02-28';
╭─ Summary ───────────╮
│ Total Time: 0.0021s │
╰─────────────────────╯
╭─ Table Scan ─────────────────────╮
│ Table: last_day_db.main.t        │
│ Type: Sequential Scan            │
│ Projections: d                   │
│ Filters:                         │
│ last_day(d) = '2025-02-28'::DATE │
│ Row Groups Scanned: 3 / 3        │
│ 2,048 rows                   0µs │
╰──────────────────────────────────╯
```


Mytherin:
Thanks!

PR #25494, opened 2026-09-09
List fixes: fix list_slice and comparator to not read uninitialized memory

### LINKED ISSUE #23715: CAST from list_slice result throws exception
### What happens?

### To Reproduce

```sql
memory D WITH t(l) AS (
           VALUES ([['bad'], ['1','2'], ['3','4']]),
                  ([['bad'], ['5','6'], ['7','8']])
         )
         SELECT CAST(list_slice(l, 2, 3) AS INTEGER[][2])
         FROM t;
Conversion Error:
Could not convert string 'bad' to INT32

LINE 5: SELECT CAST(list_slice(l, 2, 3) AS INTEGER[][2])
               ^
```
I expect ["bad"] been removed from list_slice, so I don't expect conversion failure. 

### OS:

macos

### DuckDB Version:

4c4243e09cac76267267adc1fb667a67b311590e

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

carlopi:
The fix at src/common/vector_operations/comparison_operators.cpp could be worth backporting to v1.5-variegata, I have not done it (yet) due to the API changes.

Fixes https://github.com/duckdb/duckdb/issues/23715 and an issue found while looking into that issue.

### DIFF
--- a/src/common/vector_operations/comparison_operators.cpp
+++ b/src/common/vector_operations/comparison_operators.cpp
@@ -695,17 +695,17 @@ static void ListOrArrayComparator(const Vector &left, const Vector &right, int8_
 			DistinctComparatorTypeSwitch(left_child, right_child, child_result.get(), left_child_sel, right_child_sel,
 			                             active_count);
 			// partition active into resolved vs still-remaining
-			remaining_count = NestedScatter(child_result.get(), result_data, remaining_count, remaining_lhs_sel,
+			remaining_count = NestedScatter(child_result.get(), result_data, active_count, remaining_lhs_sel,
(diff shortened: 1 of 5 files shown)


carlopi:
@hawkfish: this partially fix and revert a test change in https://github.com/duckdb/duckdb/pull/24162/changes

I am not completely sure of the changes, they seem self consistent, but I think it would be great if you can have a look.

hawkfish (approved):

PR #25495, opened 2026-09-09
Compare nested children with distinct semantics for = and <>

carlopi:
Single commit on top of https://github.com/duckdb/duckdb/pull/25494, fixing another bug found in the area.

`StandardComparatorExecute::Execute` records a NULL comparison in the validity
mask and leaves `result_data[i]` unwritten. The EQUAL/NOTEQUAL branches of the
LIST/ARRAY and STRUCT comparators passed `nullptr` as the scatter's
`result_validity`, so that mask was discarded — `null` was always false — and the
predicate read the uninitialized byte out of `child_result`. The result then
disagreed with the constant-folded Value path and depended on the other rows in
the chunk:

```sql
SELECT [NULL] = [3];                       -- false
CREATE TABLE q(a INT[], b INT[]);
INSERT INTO q VALUES ([NULL], [3]);
SELECT a = b FROM q;                       -- true

INSERT INTO q VALUES ([2, 1], [2, NULL]);  -- false on its own, true after another row
```

### DIFF
--- a/src/common/vector_operations/comparison_operators.cpp
+++ b/src/common/vector_operations/comparison_operators.cpp
@@ -421,46 +421,20 @@ static void StructComparator(const Vector &left, const Vector &right, int8_t *re
 		case ExpressionType::COMPARE_GREATERTHAN:
 		case ExpressionType::COMPARE_LESSTHANOREQUALTO:
 		case ExpressionType::COMPARE_GREATERTHANOREQUALTO:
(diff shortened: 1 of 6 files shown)


lnkuiper:
My clanker is happy with both this and the prerequisite PR. Maybe @hawkfish should have a quick look at both before we merge

Mytherin:
I think this looks good - thanks!

PR #25503, opened 2026-09-09
Fix hang for database rename

### LINKED ISSUE #25502: Attach database hangs forever after rename
### What happens?

Hi team, I found attaching a database could hang forever, if we rename the attached database beforehand; see repro for details.

### To Reproduce

```sql
memory D ATTACH '/tmp/attach_alias_repro.db' AS orig_name;
memory D CREATE TABLE orig_name.t AS SELECT range AS i FROM range(3);
memory D ALTER DATABASE orig_name SET ALIAS TO other_name;
memory D ATTACH IF NOT EXISTS '/tmp/attach_alias_repro.db' AS orig_name;
-- hangs forever
```

### OS:

macos

### DuckDB Version:

bade309c1c5ded881fec74cbc94ac14048ad0107

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25502

There're two maps managing database-related information
- [`databases`](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/include/duckdb/main/database_manager.hpp#L125-L126) maps from alias to `AttachedDatabase`
- [`DatabasePathManager`](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/include/duckdb/main/database_manager.hpp#L135-L136) maps from database path to database info [here](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/include/duckdb/main/database_file_path_manager.hpp#L48-L55)
- These two maps should be kept consistent in terms of attached database

When we rename a database, we only update the first map [here](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/main/database_manager.cpp#L303-L323), which breaks the invariant

Why it leads to infinite loop and how it's fixed in the PR
- Before the fix, both [`already_exists`](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/main/database_file_path_manager.cpp#L34) and [`already_in_system`](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/main/database_file_path_manager.cpp#L35) are true, so [`ALREADY_EXISTS`](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/main/database_file_path_manager.cpp#L47) is returned and we're trapped [in the deadloop](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/main/database_manager.cpp#L165)
- After the fix, `already_exists` (decided by matching attached database name, which our fix updates on rename) is false, so [exception is directly thrown](https://github.com/duckdb/duckdb/blob/43f897e5f3446bde2b36cef5dc137eea14211fd9/src/main/database_file_path_manager.cpp#L54-L57)

### DIFF
--- a/src/include/duckdb/main/database_file_path_manager.hpp
+++ b/src/include/duckdb/main/database_file_path_manager.hpp
@@ -41,6 +41,9 @@ public:
 	void EraseDatabasePath(const string &path);
 	//! Called when a database is detached, but before it is fully finished being used
 	void DetachDatabase(DatabaseManager &manager, const string &path);
+	//! Rewrite `path`'s tracked name from `old_name` to `new_name`.
+	//! Silent no-op if `path` is not tracked or the tracked name is not `old_name`.
+	void RenameDatabasePath(const string &path, const Identifier &old_name, const Identifier &new_name);
 
 private:
 	//! The lock to add entries to the db_paths map
--- a/src/main/database_file_path_manager.cpp
+++ b/src/main/database_file_path_manager.cpp
@@ -87,4 +87,19 @@ void DatabaseFilePathManager::DetachDatabase(DatabaseManager &manager, const str
 	}
 }
 
+void DatabaseFilePathManager::RenameDatabasePath(const string &path, const Identifier &old_name,
+                                                 const Identifier &new_name) {
+	if (path.empty() || path == IN_MEMORY_PATH) {
+		return;
+	}
+	const lock_guard<mutex> path_lock(db_paths_lock);
+	auto entry = db_paths.find(path);
+	if (entry == db_paths.end()) {
+		return;
+	}
+	if (entry->second.name == old_name.GetIdentifierName()) {
+		entry->second.name = new_name.GetIdentifierName();
+	}
+}
+
 } // namespace duckdb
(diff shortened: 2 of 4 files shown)


lnkuiper:
1. Codex thinks there's a new hang when two database instances attach the same file read-only using the same name, and then rename. Alias tracking needs to distinguish instances.

2. This still times out:
```sql
ATTACH '/tmp/alias_case_repro.db' AS orig_name;
ALTER DATABASE ORIG_NAME SET ALIAS TO other_name;
SET max_execution_time=500;
ATTACH IF NOT EXISTS '/tmp/alias_case_repro.db' AS orig_name;
```

PR #25505, opened 2026-09-09
 Fix validity mask bounds during append rollback

Liuzw1998:
### Problem 

  A primary-key conflict during commit can trigger a fatal rollback error in Debug builds: ValidityRevertAppend clears bits up to the next byte boundary, exceeding the mask’s logical capacity.

```
void ValidityRevertAppend(ColumnSegment &segment, idx_t new_count) {
...
	for (idx_t i = start_bit; i < bit_end; i++) {
			mask.SetValid(i);  // Here i would possibly exceed segment.count
		}
...
```

```
	inline void SetValid(idx_t row_idx) {
#ifdef DEBUG
		if (row_idx >= capacity) {
			throw InternalException("ValidityMask::SetValid - row_idx %d is out-of-range for mask with capacity %llu",
			                        row_idx, capacity);
		}
#endif
```
  Validation: the native Debug reproducer now returns a normal constraint error. Running the new regression test is blocked by a compilation error in test_http_transport_manager.cpp.

### DIFF
--- a/src/storage/compression/validity_uncompressed.cpp
+++ b/src/storage/compression/validity_uncompressed.cpp
@@ -622,7 +622,7 @@ void ValidityRevertAppend(ColumnSegment &segment, idx_t new_count) {
 		idx_t byte_pos = start_bit / 8;
 		idx_t bit_end = (byte_pos + 1) * 8;
 		ValidityMask mask(reinterpret_cast<validity_t *>(buffer_ptr), segment.count);
-		for (idx_t i = start_bit; i < bit_end; i++) {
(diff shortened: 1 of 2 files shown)


dentiny:
I'm wondering if bugfix PR should get into v2.0 instead of main?

lnkuiper:
Ah yes, you're right. I didn't notice. Please rebase onto the v2.0 branch

EDIT: I was finally able to remove it from the queue.

Liuzw1998:
> Ah yes, you're right. I didn't notice. Please rebase onto the v2.0 branch
> 
> EDIT: GH won't let me remove from queue!

Sure, I will rebase later.

PR #25507, opened 2026-09-09
`HTTPRetryState`

lnkuiper:
I noticed some retry attempts were getting multiplied by the retries in core + httpfs. This PR implements `HTTPRetryState`, which has a clear contract, and should prevent multiplication of retries.


src/main/http/http_util.cpp
```diff
@@ -488,135 +482,96 @@ void HTTPUtil::DecomposeURL(const string &input, string &path_out, string &proto
...
-			}
+		    // Request errors and caught exceptions are eligible without a response status.
+		    should_retry = !response || params.http_util.ShouldRetry(request, *response);
+		    if (!should_retry) {
+			    auto response_code = static_cast<uint16_t>(response->status);
+			    response->success = (response_code >= 200 && response_code < 300) ||
+			                        response->status == HTTPStatusCode::NotModified_304;
+			    return HTTPRetryDecision::Finish();
+		    }
+		    if (!HTTPUtil::IsIdempotent(request.type)) {
```
Tishj, line 538: Can we make this extendable somehow? Or introduce support for [this](<https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Idempotency-Key>) ?

lnkuiper: Yes seems reasonable. I need this PR in today so I can do some more httpfs changes today, but feel free to tinker with this.

src/include/duckdb/main/http/http_retry_budget.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+private:
+	//! Core's hook runs after admission/backoff, outside the attempt's transport-error handling.
+	void Run(const std::function<HTTPRetryDecision()> &attempt, const std::function<void()> &before_retry);
+	bool ConsumeAndWait(const HTTPRetryDecision &decision);
+
+private:
+	//! Retry settings captured for the operation.
+	const uint64_t retries;
+	const uint64_t retry_wait_ms;
```
dentiny, line 60:
```suggestion
	const uint64_t retry_backoff_base_ms;
```
since it doesn't mean the overall retry timeout, but a base value.

I'm wondering do you think it's better to cap total retry attempt time duration? 
My take
- Nested retry attempts itself is not evil, having multiplication of retries is not easy to avoid in some cases, for example, 3rd party lib could have its own retry logic (don't think it's the case for DuckDB), sub-request (i.e., write operation splits into initiate + multiple part upload + finalization), etc.
- The goal, I assume, is to avoid indefinite retries across multiple layers. I think as long as all retry attempts follow exponential retry policy, and the operation overall is capped by a timeout, it should be good.

golang's [`context`](https://pkg.go.dev/context) design is used across their IO system, you could tell it adopts the similar philosophy:
- A centralized "token" (or "budged" in this PR), which passes down across all layers
- The `context` itself carries operation timeout information (instead of a retry count)

dentiny: Also think the design is general enough, maybe could be extracted into a more common code place? :thinking_face:

lnkuiper: Thanks for the review. I don't have time to rerun CI today, going to merge as-is. Feel free to improve on this in the coming weeks!

PR #25508, opened 2026-09-09
[Parser] Add `DialectExtension` extension API

Tishj:
This PR is part of https://github.com/duckdblabs/duckdb-internal/issues/10434

And includes https://github.com/duckdb/duckdb/pull/25437

### `DialectExtension`
Interface exposing virtual methods:
- `ApplyGrammarChanges` - direct access to the `ParsedGrammar` to modify.
- `CreateKeywordHelper` - provide a custom `PEGKeywordHelper` implementation.
- `CreateTokenizer` - provide a custom `Tokenizer` implementation.
- `CreateMatcherFactory` - provide a custom `MatcherFactory` implementation.

The `DialectExtension` is named, registered in the `ExtensionCallbackManager`.
The available dialects can be listed by `FROM duckdb_dialects()`.
The `current_dialect` setting can be changed to one of these options to select a dialect.

When a `DialectExtension` is selected, its `CompiledGrammar` will be created and cached in the instance. Any `CompiledGrammar::Get(...)` calls afterwards will return the compiled grammar of the dialect, until the `current_dialect` is changed to another dialect or `RESET` to NULL.
The `DialectExtension` can also define a `DialectCompatibilityMode` to set the `dialect_compatibility_mode` setting when the dialect is selected, removing the need for manually changing the `dialect_compatibility_mode` setting.

Note that any `GrammarExtension` will be ignored by the system as long as a `DialectExtension` is active.


src/common/settings.json
```diff
@@ -270,8 +270,7 @@
         "name": "current_dialect",
         "description": "The SQL dialect used by the parser",
         "type": "VARCHAR",
-        "scope": "global",
-        "default_value": "duckdb",
+        "scope": "local",
```
Tishj, line 273: Made the setting local to a connection, removed the default setting, as we don't provide a `duckdb` dialect extension, instead that's just the base grammar.

src/include/duckdb/parser/peg/transformer/parse_result.hpp
```diff
@@ -431,7 +431,7 @@ class StringLiteralParseResult : public ParseResult {
 		return result;
 	}
 
-	unique_ptr<ParsedExpression> ToExpression() {
+	virtual unique_ptr<ParsedExpression> ToExpression() {
```
Tishj, line 434: This enables a custom `Matcher` implementation to produce custom `ParseResult` implementations for literals, called in `PEGTransformerFactory::TransformLiteralExpression` (only for StringLiteral currently, should probably be less hardcoded)

src/include/duckdb/parser/peg/dialect_extension.hpp
```diff
@@ -0,0 +1,86 @@
...
+class MatcherFactory;
+class Tokenizer;
+class PEGKeywordHelper;
+class MatcherAllocator;
+class ParsedGrammar;
+struct CompiledGrammar;
+struct CompiledGrammarRule;
+class Matcher;
+
+using compiled_rules_map_t = case_insensitive_map_t<unique_ptr<CompiledGrammarRule>>;
```
Tishj, line 29: This should be moved to a central header

src/include/duckdb/parser/peg/dialect_extension.hpp
```diff
@@ -0,0 +1,86 @@
...
+	virtual unique_ptr<MatcherFactory> CreateMatcherFactory(CreateMatcherFactoryInput &input);
+	virtual unique_ptr<Tokenizer> CreateTokenizer(CreateTokenizerInput &input);
+	virtual unique_ptr<PEGKeywordHelper> CreateKeywordHelper(CreateKeywordHelperInput &input);
+
+private:
+	//! The name to reference this dialect with the 'current_dialect' setting.
+	string name;
+	//! If set, the name of the 'dialect_compatibility_mode' setting to use upon selection.
+	optional<DialectCompatibilityMode> compatibility_mode;
+	mutex lock;
```
Tishj, line 81: nit: maybe an additional atomic so we can skip the lock check after its compiled?

src/parser/peg/compiled_grammar.cpp
```diff
@@ -6,17 +6,30 @@
...
+                                 const Matcher &program_matcher, const Matcher &top_level_statement_matcher)
+    : allocator(std::move(allocator_p)), keyword_helper(std::move(keyword_helper_p)), tokenizer(std::move(tokenizer_p)),
+      rules(std::move(rules_p)), program_matcher(program_matcher),
+      top_level_statement_matcher(top_level_statement_matcher) {
 }
 
 shared_ptr<CompiledGrammar> CompiledGrammar::Get(ClientContext &context) {
 	auto &client_config = ClientConfig::GetConfig(context);
+	auto &callback_manager = ExtensionCallbackManager::Get(context);
+	if (client_config.current_dialect) {
```
Tishj, line 25: DialectExtension takes precedence over any active grammar extensions

src/parser/peg/compiled_grammar.cpp
```diff
@@ -134,15 +147,24 @@ CompiledGrammar::Create(const case_insensitive_map_t<reference<GrammarExtension>
...
-	new_matcher->top_level_statement_matcher = factory.GetMatcher("TopLevelStatement");
+
+	MatcherAllocator allocator;
+	auto terminal_rule_overrides = grammar.BuildTerminalRuleOverrides(*keyword_helper);
+	MatcherFactory factory(allocator, grammar, rules, std::move(terminal_rule_overrides));
+
+	auto &program_matcher = factory.CreateRootMatcher("Program");
+	auto &top_level_statement_matcher = factory.GetMatcher("TopLevelStatement");
+
+	auto new_matcher = shared_ptr<CompiledGrammar>(new CompiledGrammar(std::move(allocator), std::move(keyword_helper),
```
Tishj, line 165: The `CompiledGrammar` is now fully composed of its components, rather than already created and passed around while its partially configured

src/parser/peg/dialect_extension.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+	lock_guard<mutex> guard(lock);
+	if (!cache) {
+		cache = result;
+	}
+	return cache;
+}
+
+void DialectExtension::ApplyGrammarChanges(GrammarChangesInput &) {
+	return;
```
Dtenwolde, line 65: This is intended to just return? Feels like I am missing something

Dtenwolde (commented):
Minor thing;

Also I am not sure I understand the use case for the `DialectCompatibilityMode`, nor did I see it in a test.

src/parser/peg/dialect_extension.cpp
Tishj: It's a virtual method, this is the base implementation
I'm also fine with making this pure virtual, so it doesn't have a base implementation

Dtenwolde (approved):

PR #25509, opened 2026-09-09
Implement ALTER TABLE ADD FOREIGN KEY statement.

zippond:
Trying to implement ALTER TABLE ADD FOREIGN KEY statement. Issue #57


zippond:
For the failing case, I cannot reproduce it in my local build. And it seems like what PR 25511 trying to fix. Not sure if it's caused by my change.

dentiny:
Hmmm I think the Wal replay bug I fixed should be deterministic -- previously replay doesn't work for rename + primary key in single transaction. Anyway my PR has been merged, feel free to revise and retrigger CI :)

zippond:
Sorry. The PR I mentioned should be #25505 , not 25511.

zippond:
Gentle bump. Happy to address feedback whenever you have bandwidth

PR #25512, opened 2026-09-09
add optional machine friendly json-formatted output to `unittest`

Maxxen:
Pending a future unit testing rework, this should make it much easier for agents/scripts/machines to inspect the output of `unittest`. 

A full green run with `--output=json` now produces 14kb of stdout output, compared to ~507kb.

```
  ┌───────────────────────────────────┬─────────┬───────┐
  │                                   │  bytes  │ lines │
  ├───────────────────────────────────┼─────────┼───────┤
  │ Before this work (default output) │ 506,860 │ 6,377 │
  ├───────────────────────────────────┼─────────┼───────┤
  │ JSON emitting every test          │ 950,910 │ 6,274 │
  ├───────────────────────────────────┼─────────┼───────┤
  │ JSON now (non-passing only)       │ 14,110  │ 73    │
  └───────────────────────────────────┴─────────┴───────┘
  ```
  
  Example:
  ```bash
  ./build/reldebug/test/unittest --output='json' "test/sql/function/list/*"
{"event":"test","name":"test/sql/function/list/aggregates/avg.test","status":"skip","duration":0.0127215,"assertions":{"passed":18,"failed":0},"skip_reason":"require longdouble"}
{"event":"summary","total":117,"passed":116,"failed":0,"skipped":1,"assertions":{"passed":9667,"failed":0},"failed_tests":[],"skip_reasons":{"require longdouble":1}}
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,6 +33,62 @@ build/reldebug/test/unittest "*"
 
 It is recommended to use `make reldebug` and `build/reldebug/test/unittest` unless a good reason exists to use the debug build - the debug build is much slower than the reldebug build.
 
+Multiple filters go in **one** invocation, comma-separated. Space-separated filters are concatenated
+into a single (usually unmatchable) pattern rather than OR-ed:
+
+```bash
+# correct - runs both
+build/reldebug/test/unittest "test/sql/order/test_limit.test,test/sql/order/limit_union.test"
+
+# wrong - looks for one test named "...test_limit.testtest/sql/order/limit_union.test"
+build/reldebug/test/unittest "test/sql/order/test_limit.test" "test/sql/order/limit_union.test"
+```
+
+Progress lines go to stdout and failure output to stderr, so `2>/dev/null` keeps only progress and
+`1>/dev/null` keeps only failures.
+
+### Machine-Readable Output (`--output=json`)
+
+`--output=json` replaces the human-readable output with JSON Lines on stdout: a `test` object for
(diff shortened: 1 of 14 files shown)


smvv:
idea: can we add a similar flag `-json` to `./build/*/test/run`? That runner script runs tests automatically in parallel (among many other features). I think we should use that `test/run` script as the primary entry point instead of invoking the unittest binary directly.

(At least for the runner script, I think we can replace `--output=...` with `-json`.)

Maxxen:
@smvv Added a `-json` flag to the script, and updated `AGENTS.md` WDYT?

AGENTS.md
```diff
@@ -33,6 +33,75 @@ build/reldebug/test/unittest "*"
...
 
+Multiple filters go in **one** invocation, comma-separated. Space-separated filters are concatenated
+into a single (usually unmatchable) pattern rather than OR-ed:
+
+```bash
+# correct - runs both
+build/reldebug/test/unittest "test/sql/order/test_limit.test,test/sql/order/limit_union.test"
+
+# wrong - looks for one test named "...test_limit.testtest/sql/order/limit_union.test"
+build/reldebug/test/unittest "test/sql/order/test_limit.test" "test/sql/order/limit_union.test"
```
smvv, line 44: would be great if we can allow this in the test runner script and join these multiple arguments with a space.

I doubt this AND-pattern is useful to keep in the test runner script. If really needed, an engineer can always use the unittest binary directly to get that pattern.

Issue #25069, opened 2026-08-28
latest alpha DuckDB 2.0 with Quack server on Windows/Linux got `Authentication failed` since several days ago


imkow:
### What happens?

Installed required quack and httpfs extensions.
started server, and run cli client to access the server with quack.
Got this error:
```
Invalid Input Error:
Authentication failed

LINE 1: FROM quack_query(
```
last known good version is alpha build 38947.
no prolem on macOS for the same test .

### To Reproduce

Install required quack and httpfs extensions.
server (must be Windows/Linux): ```CALL quack_serve('quack:0.0.0.0:9090', allow_other_hostname => true, token => '1234');```
client: ```FROM quack_query(
             'quack:localhost:9090',
             'SELECT * FROM <some table> ORDER BY time DESC LIMIT 10',
             token => '1234', disable_ssl => true);```

### OS:

Windows 11 AMD64

### DuckDB Version:

latest alpha

### DuckDB Client:

CLI

### Hardware:

AMD64

### Full Name:

Jack

### Affiliation:

NONE

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

bnjjj:
Hey I got the same issue on the 2.0 alpha release. It was working with the 1.5.x version. I have error on both amd64 and arm64 on Linux. 

I have an error, here's a minimal reproduction:
```
LOAD quack;

SELECT version() AS duckdb_version;

SELECT extension_name, extension_version, install_mode
FROM duckdb_extensions()
WHERE extension_name = 'quack';

CALL quack_serve(
    'quack:127.0.0.1:9495',
    token = 'repro-token',
    allow_other_hostname = true
);

FROM quack_query(
    'quack:127.0.0.1:9495',
    'SELECT 42 AS answer',
    token = 'repro-token'
);
```
I have this error : `Invalid Input Error: Authentication failed` Any idea what I'm doing wrong ?

scaleoutsean:
> I have this error : `Invalid Input Error: Authentication failed` Any idea what I'm doing wrong ?

The OP already established it's a suspected bug, what's the point of asking that question?

The bug may be observed with `v2.1.0-alpha40775` on x64 Linux.

carlopi:
[...]

PR #25513, opened 2026-09-09
Generate stable transform result type names without RTTI

smvv:
### Summary

CI [fails](https://github.com/duckdb/duckdb/actions/runs/33619641074/job/100219139766#step:12:22) when DuckDB extensions are built with Clang 20+ but the host `unittest` binary is built with GCC 14.

The JSON extension loads successfully, but initialization fails when a transformed `SQLStatement` crosses the extension/host boundary.

DuckDB’s RTTI-free type checking uses `TransformResultTypeName<T>()`, derived from compiler-specific `__PRETTY_FUNCTION__` output. GCC and Clang produce different strings for the same C++ type, so `TryCastTransformResult` incorrectly reports:

```
Transformer for rule 'Statement' returned an unexpected type.
```

We fix this error by generating stable type identifiers at parser generation time from the parser grammar.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/10656.
See also: https://github.com/duckdblabs/duckdb-internal/issues/10709.

### DIFF
--- a/scripts/parser/generate_transformer.py
+++ b/scripts/parser/generate_transformer.py
@@ -1,4 +1,5 @@
 import argparse
+import json
 import re
 import sys
 from dataclasses import dataclass, field
@@ -9,6 +10,7 @@ from typing import List
 sys.path.insert(0, str(Path(__file__).parent))
 from inline_grammar import parse_peg_grammar, PEGTokenType
 from grammar_types import (
+    load_additional_transform_result_types,
     load_grammar_types,
     load_grammar_types_yaml,
     load_matcher_rule_overrides,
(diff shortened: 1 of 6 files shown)


guillesd:
Also fixes https://github.com/duckdb/duckdb/issues/25069 (apparently the clanker ended up finding the same issue, just not propagated correctly in Quack)

Dtenwolde (approved):

Mytherin:
Thanks!

carlopi:
One thing that was discussed / raised yesterday, is what should/will happens with extensibility, so @Tishj / @Dtenwolde please double check when you get further there.

smvv:
Indeed! There is a macro available, but perhaps we need more to improve compile/runtime safety.

PR #25515, opened 2026-09-09
[Dev] Add `annotated_mutex` and friends to more places in the storage

Tishj:
By annotating the mutexes, through clang's analysis a couple (small) problems were found in our use of lock guards. Besides those fixes, annotating these contracts helps everyone that deals with this code.

### DIFF
--- a/src/function/table/table_scan.cpp
+++ b/src/function/table/table_scan.cpp
@@ -484,6 +484,7 @@ unique_ptr<GlobalTableFunctionState> DuckTableScanInitGlobal(ClientContext &cont
 	// Check if row_number column is requested and initialize row_number_base
 	for (idx_t i = 0; i < input.column_ids.size(); i++) {
 		if (input.column_ids[i] == COLUMN_IDENTIFIER_ROW_NUMBER) {
+			annotated_lock_guard lock(g_state->state.scan_state.lock);
 			g_state->state.scan_state.row_number_base = 0;
 			break;
 		}
--- a/src/storage/data_table.cpp
+++ b/src/storage/data_table.cpp
@@ -101,7 +101,7 @@ DataTable::DataTable(ClientContext &context, DataTable &parent, ColumnDefinition
 	default_executor.AddExpression(default_value);
 
 	// prevent any new tuples from being added to the parent
-	lock_guard<mutex> parent_lock(parent.append_lock);
+	annotated_lock_guard parent_lock(parent.append_lock);
 
 	this->row_groups = parent.row_groups->AddColumn(context, new_column, default_executor);
 
@@ -116,7 +116,7 @@ DataTable::DataTable(ClientContext &context, DataTable &parent, idx_t removed_co
     : db(parent.db), info(parent.info), version(DataTableVersion::MAIN_TABLE) {
 	// prevent any new tuples from being added to the parent
 	auto &local_storage = LocalStorage::Get(context, db);
-	lock_guard<mutex> parent_lock(parent.append_lock);
+	annotated_lock_guard parent_lock(parent.append_lock);
 
 	for (auto &column_def : parent.column_definitions) {
 		column_definitions.emplace_back(column_def.Copy());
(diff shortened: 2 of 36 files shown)


src/function/table/table_scan.cpp
```diff
@@ -484,6 +484,7 @@ unique_ptr<GlobalTableFunctionState> DuckTableScanInitGlobal(ClientContext &cont
 	// Check if row_number column is requested and initialize row_number_base
 	for (idx_t i = 0; i < input.column_ids.size(); i++) {
 		if (input.column_ids[i] == COLUMN_IDENTIFIER_ROW_NUMBER) {
+			annotated_lock_guard lock(g_state->state.scan_state.lock);
 			g_state->state.scan_state.row_number_base = 0;
```
Tishj, line 488: This is guarded in other places by this lock, but it was missing here

dentiny:
Thank you!

src/storage/table/row_group_collection.cpp
```diff
@@ -187,27 +187,28 @@ MetadataManager &RowGroupCollection::GetMetadataManager() {
...
 	owned_row_groups = std::move(new_row_groups);
 }
 
 //===--------------------------------------------------------------------===//
 // Initialize
 //===--------------------------------------------------------------------===//
 void RowGroupCollection::Initialize(PersistentTableData &data) {
-	D_ASSERT(owned_row_groups->GetBaseRowId() == 0);
-	auto l = owned_row_groups->Lock();
+	auto row_groups = GetRowGroups();
```
Tishj, line 203: Checking the assertion before grabbing the lock could silently be wrong, if the data changed between the assertion and grabbing the lock

Other methods were also using `owned_row_group` directly, without going through the lock that controls this. Not sure if those were already sane due to some contract that guards `owned_row_group` from being changed that we can't express with this?
But I doubt it, seeing as `SetRowGroups` actually gets called in a couple places.

src/storage/data_table.cpp
```diff
@@ -306,9 +306,12 @@ idx_t DataTable::NextParallelScan(ClientContext &context, ParallelTableScanState
 	if (row_groups->NextParallelScan(context, state.scan_state, scan_state.table_state)) {
 		return scan_state.table_state.row_group->GetCount();
 	}
-	if (state.scan_state.row_number_base.IsValid()) {
-		// start the row number for transaction-local rows from the final row count in the base table
-		scan_state.local_state.row_number_base = state.scan_state.row_number_base.GetIndex();
+	{
+		annotated_lock_guard guard(state.scan_state.lock);
+		if (state.scan_state.row_number_base.IsValid()) {
```
Tishj, line 311: Another case of `row_number_base` being checked without a lock

src/storage/single_file_block_manager.cpp
```diff
@@ -1189,7 +1193,12 @@ void SingleFileBlockManager::ReadBlock(Block &block, bool skip_block_header) con
 
 void SingleFileBlockManager::Read(QueryContext context, Block &block) {
 	D_ASSERT(block.id >= 0);
-	D_ASSERT(std::find(free_list.begin(), free_list.end(), block.id) == free_list.end());
+#ifdef D_ASSERT_IS_ENABLED
```
Tishj, line 1196: `free_list` is protected by the lock, without it the list's references/pointers could be invalidated during `find`

src/storage/wal_replay.cpp
```diff
@@ -1325,6 +1326,10 @@ void WriteAheadLogDeserializer::ReplayRowGroupData() {
 		throw InternalException("Corrupt WAL: insert without table");
 	}
 	auto &storage = state.current_table->GetStorage();
+	auto &transaction = DuckTransaction::Get(context, db);
+	TableAppendState append_state;
```
Tishj, line 1330: This was an actual missing lock I believe? @Mytherin 
Actually, no, we have guaranteed exclusive access at this point, this just quiets the static analysis

PR #25518, opened 2026-09-09
[Dev] annotated `IndexLock`

Tishj:
Gets rid of the two-stage initialization in the process (`InitializeLock`)

cc @artjomPlaunov @Maxxen @taniabogatsch 
If there's a good reason we shouldn't do this (tie IndexLock to BoundIndex) then we can close this


artjomPlaunov (approved):
Looks good! @Maxxen can correct me if I'm wrong here (I had my clanker do a pass over VSS and Spatial extensions), it doesn't look like there's any usage of the old API that would break, albeit there are some comments that can be cleaned up there, but perhaps that can be done separately. 

And for core, this looks good. Thanks!

Mytherin:
Thanks!

PR #25519, opened 2026-09-09
chore: drop delta and unity_catalog patches

benfleis:
remove patches and update gitrefs for delta and unity_catalog extensions


Mytherin:
Thanks!

PR #25520, opened 2026-09-09
Experiment: Flatten MatchStack to reduce pointer chasing

smvv:
Idea from https://www.cs.cornell.edu/~asampson/blog/flattening.html.

### DIFF
--- a/src/include/duckdb/parser/peg/matcher_stack.hpp
+++ b/src/include/duckdb/parser/peg/matcher_stack.hpp
@@ -7,11 +7,33 @@
 
 #pragma once
 
+#include "duckdb/common/array.hpp"
 #include "duckdb/common/optional.hpp"
 #include "duckdb/common/optional_idx.hpp"
 #include "duckdb/parser/peg/matcher.hpp"
+#include "duckdb/storage/arena_allocator.hpp"
 
 namespace duckdb {
+
+using match_frame_index_t = idx_t;
+
+class MatchStack;
+
+//! Locates a MatchState either outside the stack or within a stable frame slot.
+struct MatchStateReference {
+	explicit MatchStateReference(MatchState &state_p) : external_state(state_p) {
+	}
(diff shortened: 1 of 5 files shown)


Tishj:
Essentially we allocate some additional space next to each frame, to try to inline the MatchProcess there, for better locality?

src/include/duckdb/parser/peg/matcher_stack.hpp
```diff
@@ -7,12 +7,34 @@
...
+	}
+	MatchStateReference(match_frame_index_t frame_index_p, idx_t frame_offset_p)
+	    : frame_index(frame_index_p), frame_offset(frame_offset_p) {
+	}
+
+	MatchState &Get(MatchStack &stack);
+
+private:
+	optional_ptr<MatchState> external_state;
+	optional_idx frame_index;
```
Tishj, line 34: The one thing I don't get is this. We push children one at a time (list doesn't push every child all at once), so the parent frame is always the one below it, can't that knowledge be used to simplify this?

Dtenwolde: I think so as well. Also I don't think we'll need an explicit index as the addresses are stable in the arena (they never move as far as I know). So we can also use `MatchState &match_state`

Dtenwolde (commented):
I like the idea, but we can probably simplify the solution. Also we can reuse this for `ParseResult` as well. 
Have you looked at the performance?

src/include/duckdb/parser/peg/matcher_stack.hpp
smvv: i'm not too familiar with these arena-specifics. In general, if we need to expand the arena, we would allocate a different chunk of memory, and we need to update each address. Indices do not have that problem because they are used as an offset to a base address. Let me try to change the implementation to see if it's needed.

PR #25523, opened 2026-09-09
Make float or double depending on precision

Dtenwolde:

This is something we didn't properly port from the old parser
We now respect the `FLOAT(p)` again. If `p` is between 1-24 it results in `FLOAT`, 25-53 returns `DOUBLE`. Outside of this range we throw an error


Maxxen:
This probably wont work once https://github.com/duckdb/duckdb/pull/25473 lands. Can we make this decision during binding instead? I.e. add a constructor/overload for `FLOAT` in `default_types.cpp`?

Maxxen:
Something like this:

```c++
// default_types.cpp
const builtin_type_array BUILTIN_TYPES = {
    // ...
    {"float", LogicalTypeId::FLOAT, nullptr},
    {"real", LogicalTypeId::FLOAT, RegisterFloatConstructors},
    {"float4", LogicalTypeId::FLOAT, nullptr},
    // ...
}                                           

void RegisterFloatConstructors(TypeConstructorSet &set) {
	// Basic constructor without parameters
	set.AddFunction(TypeConstructor::Identity("float"));

	// Constructor with parameter
	auto signature = TypeConstructor::Signature();
	signature.AddParameter("precision", LogicalType::BIGINT);
	set.AddFunction(TypeConstructor(std::move(signature), [](BindLogicalTypeInput &input) -> LogicalType {
		auto precision = input.modifiers[0].GetValue().GetValue<int64_t>();
		if (precision <= 24) {
			return LogicalType::FLOAT;
		} else if (precision <= 53) {
			return LogicalType::DOUBLE;
		} else {
			throw BinderException(input.GetLocation(0), "FLOAT type precision must be between 1 and 53");
		}
	}));

}
```

Dtenwolde:
Sure, thanks for the heads up!
Changed it to be based on: https://github.com/duckdb/duckdb/pull/25473
See https://github.com/duckdb/duckdb/pull/25523/changes#diff-b9644edd59844ad5bbab14ea4e8c6ef8036232fb6478a9295a3d89d0f3a33bf4R21

Maxxen:
Can you merge with upstream again? Then ill add it to the merge queue

Issue #25430, opened 2026-09-07
Expression::HasParameter crashes planning a correlated subquery with a STRUCT-access join predicate


suyZhong:
### What happens?

Planning a query that cross-joins a table with (t0 INNER JOIN a correlated subquery) on a STRUCT field access (`t0.c2['k0']`), with `WHERE (true) IS NOT DISTINCT FROM (t0.c1)`, crashes the process.

### To Reproduce

```sql
CREATE TABLE t0(c1 BOOLEAN, c2 STRUCT(k0 BOOLEAN, k1 VARCHAR));
CREATE TABLE t1(c0 INT, c1 VARCHAR(500));
SELECT t1.* FROM t1, t0 INNER JOIN (SELECT t0.c2 AS col0 FROM t0) AS sub0 ON t0.c2['k0'] WHERE ((true)IS NOT DISTINCT FROM(t0.c1));
```

### OS:

Ubuntu 22.04

### DuckDB Version:

v2.1.0-dev83390 (Development Version, 3f7a54d0fb)

### DuckDB Client:

CLI

### Full Name:

Suyang Zhong

### Affiliation:

NUS

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

huoguocheng:
I would like to work on this issue. Please assign it to me.

PR #25524, opened 2026-09-09
Fix use-after-free when multiple filters are pushed into one scan

### LINKED ISSUE #25428: INTERNAL Error: Attempted to dereference unique_ptr that is NULL
### What happens?

Below test case trigger an assertion error in CLI. Besides, executing the trigger query aborts the JVM when using jdbc.

### To Reproduce

```sql
CREATE TABLE t1(c0 VARCHAR, c1 INT, c2 STRUCT(k0 BOOLEAN, k1 VARCHAR, k2 VARCHAR, k3 INT, k4 BOOLEAN), PRIMARY KEY(c1, c0));
INSERT INTO t1(c0, c1, c2) VALUES ('*', -1537721958, {'k0': false, 'k1': 'O', 'k2': 'w', 'k3': -1093516081, 'k4': false});
SELECT t1.c1 FROM t1 INNER JOIN (SELECT (t1.c0 || t1.c0) AS col0 FROM t1 WHERE t1.c2['k3']) AS sub0 ON t1.c2['k3'] WHERE (t1.c1 > t1.c1);

```

### OS:

Ubuntu 22.04

### DuckDB Version:

DuckDB v2.1.0-dev83390 (Development Version, 3f7a54d0fb)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Suyang Zhong

### Affiliation:

NUS

- [x] Yes, I have

- [x] Yes, I have

alonfaraj:
Fixes #25428

When the pruner turns a scan's pushed-down filters back into expressions, it visited each one right after appending it to the list. Visiting a struct extract records a reference to that slot in the list, so the next append reallocates the vector and leaves the  reference dangling. 
Needs two filters on the same scan with the extract not being the last one pushed.

Fix is to build the list first and visit it afterwards. Did the same for the multi-column filter loop below it, which visited a local that died at the end of each iteration.


Mytherin:
Thanks!

alonfaraj:
@Mytherin is something blocking this auto merge?

Mytherin:
Yes, there are merge conflicts. Could you resolve them?

alonfaraj:
> Yes, there are merge conflicts. Could you resolve them?

Seems like #25430 landed the same fix for the filter_expressions loop, so I dropped my half of that. What's left is just the multi-column filter loop below it, which still builds its column ref in a local that dies at the end of each iteration while  column_references is holding a reference to it.

PR #25525, opened 2026-09-09
Report the temporary files a directory holds, and let a caller reclaim them

### LINKED ISSUE #24576: Orphaned temporary files are not cleaned up

carlopi:
#### Example
```sql
-- claim the temp directory now rather than at the first spill, and reclaim
-- whatever processes that are no longer running left in it
FROM initialize_temporary_directory();

-- or claim it reclaiming nothing, which also holds the once-per-directory
-- claim, so the first spill does not reclaim either
FROM initialize_temporary_directory(cleanup := false);

-- what this instance has out there ...
FROM duckdb_temporary_files();
-- ... and what the whole directory holds, other instances included
FROM duckdb_temporary_files(external := true);

-- reclaim again later, without waiting for a spill of our own
FROM cleanup_temporary_directory();
```
Those are only additive, in the sense that allow explicit control for some behaviours that before were only implicit.
Only improvement is listing also `.block` files, that were before invisible via `FROM duckdb_temporary_files()`.

#### Details
duckdb_temporary_files() describes the grouped .tmp files, from the manager that knows how much of each is in use, plus the one-file-per-buffer .block spills and the file claiming this instance's id, read from disk. The .block half used to resolve against the working directory and so never appeared, though the size the directory is held to was counted correctly throughout. external defaults to false, so summing the function still means what it always meant.

cleanup_temporary_directory() uses the sweep's rule unchanged: remove only files whose owning process is gone. Our own process is running, so a live spill is never a candidate. It is best-effort and never throws, and it deletes a dead owner's claim last, so a process inheriting that pid cannot take the id and begin writing the paths being deleted. It earns its keep on a long-running server, which under the once-per-directory rule may never sweep again after its first spill.

initialize_temporary_directory() throws if asked twice, because the choice is spent and silence would tell the caller they had suppressed a sweep that had already run; silent := true forgives being second when the choice was the same one.

The file claiming an id is duckdb_temp_<pid>_<instance>__claim: the pid and the instance already say who owns it, and the doubled underscore sorts below the first letter of every other suffix, so the claim heads its owner's files.

All three return with the same schema, returning either existing OR deleted files.

Fixes https://github.com/duckdb/duckdb/issues/24576


carlopi:
This also fixes a bug where `.block` files would not be shown by `FROM duckdb_temporary_files()`, and generally adds observability / testability capabilities, that could help in some operation cases.

src/function/table/system/duckdb_temporary_files.cpp
```diff
@@ -22,39 +24,124 @@ static unique_ptr<FunctionData> DuckDBTemporaryFilesBind(ClientContext &context,
...
+};
+
+static unique_ptr<FunctionData> InitializeTemporaryDirectoryBind(ClientContext &context, TableFunctionBindInput &input,
+                                                                 vector<LogicalType> &return_types,
+                                                                 vector<Identifier> &names) {
+	TemporaryFilesBind(context, input, return_types, names);
+	auto result = make_uniq<InitializeTemporaryDirectoryData>();
+	for (auto &entry : input.named_parameters) {
+		if (entry.first == "cleanup") {
+			result->cleanup = BooleanValue::Get(entry.second.DefaultCastAs(LogicalType::BOOLEAN));
```
dentiny, line 115: should we check whether the value is `NULL`?

carlopi: Fixed, thanks!

carlopi:
input: I need to look at whether moving back to cleanup only on close does make more sense.

PR #25526, opened 2026-09-09
patch debug and deprecated settings

xproblematique:
Closes https://github.com/duckdblabs/duckdb-internal/issues/10714

Follow-up fix for `duckdb_settings()` debug/deprecated filtering.

* Add `is_debug` and `is_deprecated` flags to `ExtensionOption` and pass them through `DBConfig::AddExtensionOption`.
* Apply the debug/deprecated filtering to extension settings in `duckdb_settings()`.
* Fix the block in `src/catalog/default/default_table_functions.cpp`.
* Add regression tests verifying debug and deprecated settings are mutually exclusive.

All Extension settings currently default to False.

### DIFF
--- a/test/sql/settings/debug_deprecated_settings.test
+++ b/test/sql/settings/debug_deprecated_settings.test
@@ -22,7 +22,17 @@ SELECT list_count(
 ----
 0
 
-query I
-SELECT count(*) FROM duckdb_settings(debug := true, deprecated := true);
+query T
+SELECT name FROM duckdb_settings(debug := true) WHERE name = 'debug_enable_caching_operators';
 ----
-0
+debug_enable_caching_operators
+
+query T
+SELECT count(*) != 0 FROM duckdb_settings(debug := true);
+----
+true
+
+query T
+SELECT count(*) != 0 FROM duckdb_settings(deprecated := true);
+----
+true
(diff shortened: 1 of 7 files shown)


test/sql/settings/debug_deprecated_settings.test
```diff
@@ -0,0 +1,28 @@
...
+    list_intersect(
+        (SELECT list(name) FROM duckdb_settings(deprecated := true)),
+        (SELECT list(name) FROM duckdb_settings(deprecated := false))
+    )
+);
+----
+0
+
+query I
+SELECT count(*) FROM duckdb_settings(debug := true, deprecated := true);
```
taniabogatsch: Is this test correct? I think we could potentially have a debug setting that we want to deprecate? Maybe we add a `FIXME` that, for now, we have none of these settings.

If we do that, then we can do the same for extension settings (after loading an extension with `require parquet`, for example). That way we at least touch that code path?

test/sql/settings/debug_deprecated_settings.test
taniabogatsch, line 1: Maybe we can add the following additional tests:

- choose one specific setting we know is a debug/deprecated one and check its classified correctly: 
```sql
SELECT name FROM duckdb_settings(debug := true) WHERE name = 'my_setting';
```
- make sure there are debug/deprecated settings:
```sql
SELECT count(*) != 0 FROM duckdb_settings(debug := true);
```

taniabogatsch (commented):
Hi @xproblematique, thanks for the quick fix! I've left two comments, let's add a few more tests!

taniabogatsch (approved):

Mytherin:
Thanks!

PR #25527, opened 2026-09-09
Optimize keyword matching for the PEG parser

Dtenwolde:
This PR adds two optimizations for keyword matching: 
1. Each literal in the grammar gets an integer ID. We store both the keyword category using a `uint8_t` bitmask (a keyword may be in two categories) and a dense ID (`uint16_t`) which we cache on the first lookup. This avoids repeated string comparisons and keyword-category lookups. 
2. For rules consisting only of literals, such as `UnreservedKeyword` we create a special `LiteralChoiceMatcher`. It uses the token's ID to directly select the matching branch, avoid attempts that would fail anyway. We do still go to the respective KeywordMatcher to update token position and return a ParseResult. 

Grammar extensions and custom matchers should not be influenced. 

Benchmarks show a 24-34% improvement for TPC-H, TPC-DS, Flummi, and AoC. Perhaps I should create some parser specific performance tests.


src/include/duckdb/parser/peg/grammar_literal_table.hpp
```diff
@@ -0,0 +1,85 @@
...
+	explicit LiteralInfo(uint32_t literal_id) : value(literal_id) {
+		D_ASSERT(literal_id <= MAX_LITERAL_ID);
+	}
+
+	uint32_t LiteralId() const {
+		return value & MAX_LITERAL_ID;
+	}
+
+	bool IsKeyword() const {
+		return (value >> 24) != 0;
```
Tishj: I think we can derive both this constant and the MAX_LITERAL_ID from the same source?

src/include/duckdb/parser/peg/grammar_literal_table.hpp
```diff
@@ -0,0 +1,85 @@
...
+	bool operator==(const LiteralInfo &other) const {
+		return value == other.value;
+	}
+
+private:
+	static uint32_t CategoryMask(PEGKeywordCategory category) {
+		if (category == PEGKeywordCategory::KEYWORD_NONE || category > PEGKeywordCategory::KEYWORD_TYPE_NAME) {
+			return 0;
+		}
+		return uint32_t(1) << (23 + static_cast<uint8_t>(category));
```
Tishj: As well as this one

src/include/duckdb/parser/peg/keyword_helper/parsed_grammar_keyword_helper.hpp
```diff
@@ -15,9 +16,13 @@ class ParsedGrammarKeywordHelper : public PEGKeywordHelper {
 	bool KeywordCategoryType(const string &text, PEGKeywordCategory type) const override;
 	bool IsKeyword(const string &text) const override;
 	vector<ParserKeyword> KeywordList() const override;
+	optional_ptr<const GrammarLiteralTable> GetLiteralTable() const override {
```
Tishj, line 19: Don't we just want to retire `PEGKeywordCategory` for explicit constants defined by the keywordhelper?
This feels like it's sidestepping the current shoddy system rather than replacing it

The keyword categories are a contract between the grammar, the keywordhelper and the matcher implementations, they don't need to make sense to anyone else (besides having a unique string to show in `duckdb_keywords`)

src/include/duckdb/parser/peg/grammar_literal_table.hpp
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+private:
+	uint32_t value = 0;
+};
+
+//! Immutable after construction, including literals only present in keyword-category rules.
+class GrammarLiteralTable {
+public:
+	DUCKDB_API GrammarLiteralTable(const ParsedGrammar &grammar, const DefaultKeywordMaps &keyword_maps);
```
Tishj, line 63: This is tied to duckdb/postgres keyword categories (through `DefaultKeywordMaps`), making it not extendable

Tishj:
I left some comments, and got to work on making a patch to check the feasibility, and it's feasible, but it's maybe better left as a followup

Tishj (approved):
Thanks!

Dtenwolde:
Merged with upstream to include the parser regression CI runs, lets see what they say!

Dtenwolde:
```
FASTER
benchmark                  base median                   PR median                     delta median        runs
-------------------------  ----------------------------  ----------------------------  ------------------  ----
ParserNestedExpressions      1.9 s  [  1.9 s …  1.9 s ]    1.8 s  [  1.8 s …  1.8 s ]  -129.0 ms ( -6.8%)    40
ParserKeywordIdentifiers     1.3 s  [  1.2 s …  1.3 s ]    1.1 s  [  1.1 s …  1.2 s ]  -126.7 ms (-10.1%)    40
ParserMalformedSelect      901.9 ms [864.5 ms…  1.2 s ]  736.8 ms [726.0 ms…917.0 ms]  -165.1 ms (-18.3%)    40
ParserStatements             2.1 s  [  2.1 s …  2.6 s ]    1.7 s  [  1.7 s …  1.8 s ]  -431.1 ms (-20.4%)    40
ParserWideSelect             4.4 s  [  4.4 s …  4.5 s ]    3.5 s  [  3.4 s …  3.5 s ]  -924.2 ms (-20.8%)    40
ParserTPCH                   2.0 s  [  2.0 s …  2.2 s ]    1.6 s  [  1.5 s …  1.7 s ]  -460.0 ms (-22.9%)    40
ParserTPCDS                  4.3 s  [  4.2 s …  5.7 s ]    3.3 s  [  3.2 s …  3.5 s ]    -1.1 s  (-24.7%)    40
ParserAoC                    3.1 s  [  2.9 s …  3.6 s ]    2.3 s  [  2.1 s …  2.5 s ]  -840.8 ms (-27.0%)    40
ParserFlummi                 3.7 s  [  3.4 s …  4.4 s ]    2.6 s  [  2.3 s …  2.9 s ]    -1.1 s  (-30.4%)    40
SLOWER (+2%…<+10%)
benchmark                  base median                   PR median                     delta median        runs
-------------------------  ----------------------------  ----------------------------  ------------------  ----
ParserGrammarConstruction    1.8 s  [  1.7 s …  2.2 s ]    2.0 s  [  1.9 s …  2.1 s ]  +128.4 ms ( +7.0%)    40
geomean: 2.3 s -> 1.9 s  -404.1 ms (-17.7%)  (initial 10 samples)
```

Tishj:
CI failed because of Sander's PR
I only checked the merge for conflicts, didn't think to check compilation 🙃

Dtenwolde:
Yep, will fix it :)

PR #25530, opened 2026-09-09
fix(parser): Preserve CTE metadata when copying

darion-yaphet:
## Summary

  - Reuse `CommonTableExpressionInfo::Copy()` in both CTE container copy paths instead of maintaining duplicate field-copy logic.
  - Preserve `is_trigger_generated` when copying CTEs and mark `Copy()` as `const`.
  - Add focused tests covering property preservation, deep-copy isolation, and both CTE copy entry points.

  ## Testing

  - `cmake --build build/reldebug --target unittest -j 9`
  - `build/reldebug/test/unittest "[api][parser]"` — 3 test cases and 30 assertions passed
  - `build/reldebug/test/unittest "test/sql/cte/*"` — 85 test cases passed, 3 skipped
  - `build/reldebug/test/unittest "test/sql/trigger/*"` — 28 test cases passed, 1 skipped
  - `build/reldebug/test/unittest test/sql/parser/statement_tostring_roundtrip.test` — 174 assertions passed
  - `clang-format --dry-run --Werror` and `git diff --check`

  ## Risk and Rollback

  Risk is low because the change routes existing copy paths through the canonical deep-copy implementation. The intended behavior change is preserving trigger-generated
  CTE metadata. Reverting commit `6b99ca6526` restores the previous behavior.

### DIFF
--- a/src/include/duckdb/parser/common_table_expression_info.hpp
+++ b/src/include/duckdb/parser/common_table_expression_info.hpp
@@ -40,7 +40,7 @@ struct CommonTableExpressionInfo {
 
 	void Serialize(Serializer &serializer) const;
 	static unique_ptr<CommonTableExpressionInfo> Deserialize(Deserializer &deserializer);
-	unique_ptr<CommonTableExpressionInfo> Copy();
+	unique_ptr<CommonTableExpressionInfo> Copy() const;
 };
 
 } // namespace duckdb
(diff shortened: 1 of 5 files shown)


kryonix (approved):
Thanks! Looks good to me.

carlopi:
@kryonix: should this also be backported to `v2.0-cyanoptera` ?

kryonix:
Yes. I did not realize that this was targetting `main`.

PR #25531, opened 2026-09-09
Prune row group for shared prefix for case functions

dentiny:
Hi team, I found for case functions (`lower`/`upper`), currently row groups are only pruned when min/max are exactly the same, but actually row groups could be also pruned when min/max shared a common prefix. For example
```sql
memory D ATTACH '/tmp/upper_lower_zonemap_pruning.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE case_pruning AS
     SELECT i::INTEGER AS i, CASE WHEN i < 2048 THEN 'alpha' WHEN i < 4096 THEN 'Beta' ELSE 'GAMMA' END AS s
     FROM range(6144) r(i);
db D CREATE TABLE case_prefix AS SELECT s || i::VARCHAR AS s FROM case_pruning;

-- should access only one row group, but access all
db D EXPLAIN ANALYZE SELECT count(*) FROM case_prefix WHERE lower(s) LIKE lower('beta%');
╭─ Summary ───────────╮
│ Total Time: 0.0011s │
╰─────────────────────╯
│ Table: db.main.case_prefix        │
│ Filters: prefix(lower(s), 'beta') │
│ Row Groups Scanned: 3 / 3         │
│ 2,048 rows                    0µs │
```
The root cause is, when min != max, stats are [returned directly](https://github.com/duckdb/duckdb/blob/a0315f712c64627fde0f77879730f28a4e9c8c46/src/function/scalar/string/caseconvert.cpp#L151) without assigning an inexact stats.

### DIFF
--- a/src/function/scalar/string/caseconvert.cpp
+++ b/src/function/scalar/string/caseconvert.cpp
@@ -144,16 +144,13 @@ static unique_ptr<BaseStatistics> CaseConvertPropagateStats(ClientContext &conte
 	if (!StringStats::HasMinMax(child_stats[0])) {
 		return result.ToUnique();
 	}
-	// When min == max, all values share the stored string (exact) or the stored prefix (truncated).
-	// Case conversion is codepoint-local, so the converted string/prefix bounds the result.
+	// All values share the common prefix of min and max; case conversion preserves this property.
 	auto min = StringStats::Min(child_stats[0]);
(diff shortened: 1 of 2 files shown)


pdet:
@kryonix could you a look on this PR?

Mytherin:
Thanks - LGTM

PR #25534, opened 2026-09-09
Fix integer overflow in suffix() for 2GB+ suffixes

### LINKED ISSUE #25270: Bug94: Integer Truncation in `SuffixFunction` Causes INTERNAL Error / False-Positive Suffix Match on 2 GiB+ Strings
### What happens?

In `src/function/scalar/string/suffix.cpp` (lines 14–31), `SuffixFunction()` casts the suffix byte length from `idx_t` to `int32_t` via `UnsafeNumericCast<int32_t>(suffix_size)` at line 23 without any prior range check. When the suffix string is ≥ 2^31 bytes, this cast overflows: in a debug build the checked `NumericCast` path detects the out-of-range value and throws a fatal `INTERNAL Error` (process exits with code 1); in a release build the unchecked `static_cast` silently wraps `suffix_size` to a negative `suf_idx`, the comparison loop `for (; suf_idx >= 0; ...)` never executes, and `suffix()`/`ends_with()` unconditionally returns `true` regardless of the actual byte content.

### To Reproduce

This is triggered by pure SQL with no external data or special configuration.

```sql
SELECT suffix(repeat('a', 2147483648), repeat('b', 2147483648));
```

```sh
echo "SELECT suffix(repeat('a', 2147483648), repeat('b', 2147483648));" | ./build/debug/duckdb
```

```
INTERNAL Error: Information loss on integer cast: value 2147483648 outside of target range [-2147483648, 2147483647]

Stack Trace:

/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24c68a80) [0x5ee1581b7a80]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24c68369) [0x5ee1581b7369]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24c67cdb) [0x5ee1581b6cdb]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x24c72e50) [0x5ee1581c1e50]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x22de9bf2) [0x5ee156338bf2]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x2422c56f) [0x5ee15777b56f]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x2438cb32) [0x5ee1578dbb32]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x242ef96e) [0x5ee15783e96e]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x2427f3be) [0x5ee1577ce3be]
/data/ylwang/Projects/DirectedDB/target/duckdb/build/debug/duckdb(+0x2d6cae94) [0x5ee160c19e94]

EXIT=1
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

alonfaraj:
Fixes #25270

`suffix()` cast the suffix length to `int32_t` to drive its comparison loop. Strings can be up to 4 GiB, so anything >= 2 GiB overflowed .

Swapped the loop for a single `memcmp` over the tail so everything stays in `idx_t`. Nicer and faster for normal strings too.

### Testing
Added `suffix_large_string.test_slow`, but with `mode skip`' because it needs ~9 GB and ~40 s per query, similar to `arg_max_overflow.test_slow`

### DIFF
--- a/src/function/scalar/string/suffix.cpp
+++ b/src/function/scalar/string/suffix.cpp
@@ -18,16 +18,7 @@ static bool SuffixFunction(const string_t &str, const string_t &suffix) {
 		return false;
 	}
 
-	auto suffix_data = suffix.GetData();
-	auto str_data = str.GetData();
-	auto suf_idx = UnsafeNumericCast<int32_t>(suffix_size) - 1;
-	idx_t str_idx = str_size - 1;
-	for (; suf_idx >= 0; --suf_idx, --str_idx) {
-		if (suffix_data[suf_idx] != str_data[str_idx]) {
-			return false;
-		}
-	}
-	return true;
+	return memcmp(str.GetData() + str_size - suffix_size, suffix.GetData(), suffix_size) == 0;
 }
 
 struct SuffixOperator {
--- a/test/sql/function/string/suffix_large_string.test_slow
+++ b/test/sql/function/string/suffix_large_string.test_slow
@@ -1,0 +1,17 @@
+# name: test/sql/function/string/suffix_large_string.test_slow
+# group: [string]
+
+mode skip Test allocates 9GB and takes ~40 seconds on a release build
+
+# suffixes of 2^31 bytes or more used to overflow the comparison index
+query I
+SELECT suffix(repeat('a', 2147483648), repeat('b', 2147483648))
+----
+false
+
+query I
+SELECT suffix(repeat('a', 2147483648), repeat('a', 2147483648))
+----
+true
+
+mode unskip


src/function/scalar/string/suffix.cpp
```diff
@@ -18,16 +18,7 @@ static bool SuffixFunction(const string_t &str, const string_t &suffix) {
 		return false;
 	}
 
-	auto suffix_data = suffix.GetData();
```
Maxxen, line 21: This is technically UB under `DUCKDB_DEBUG_NO_INLINE` when the string is empty and the suffix_size is 0, as it would call `memcmp` with a NULL pointer. Could we add a small guard/check for that edge-case?

Maxxen (dismissed):

src/function/scalar/string/suffix.cpp
Maxxen: Actually, `DUCKDB_DEBUG_NO_INLINE` still considers 0-sizeds strings to be inlined, my bad!

Maxxen (approved):

PR #25538, opened 2026-09-09
Prune row group for contains on inexact stats

dentiny:
Hi team, I found for `contains` function, row groups are not pruned as expected for inexact stats, for example
```sql
memory D ATTACH '/tmp/contains_prune.duckdb' AS db (ROW_GROUP_SIZE 2048);
memory D USE db;
db D CREATE TABLE requests AS
     SELECT CASE WHEN i = 0 THEN NULL WHEN i < 4096 THEN '/api/health/' ELSE '/api/orders/' END || i::VARCHAR AS path
     FROM range(6144) t(i);

-- only third row group contains non-"health" strings, so should only access one rg
db D EXPLAIN ANALYZE SELECT path FROM requests WHERE path NOT LIKE '%health%';
╭─ Summary ───────────╮
│ Total Time: 0.0016s │
╰─────────────────────╯
╭─ Table Scan ──────────────────────────╮
│ Table: db.main.requests               │
│ Projections: path                     │
│ Filters: NOT contains(path, 'health') │
│ Row Groups Scanned: 3 / 3             │
│ 2,048 rows                        0µs │
```
The reason is, in the current implementation, we conclude [unable to prune when not exact stats](https://github.com/duckdb/duckdb/blob/a0315f712c64627fde0f77879730f28a4e9c8c46/src/function/scalar/string/contains.cpp#L150-L153); the missing case is, for inexact stats we're possible to tell the needle exists in the row group.

### DIFF
--- a/src/function/scalar/string/contains.cpp
+++ b/src/function/scalar/string/contains.cpp
@@ -147,8 +147,7 @@ static FilterPropagateResult ContainsFilterPrune(const FunctionStatisticsPruneIn
 		return FilterPropagateResult::FILTER_ALWAYS_FALSE;
 	}
 
-	if (StringStats::GetMinType(*haystack_stats) != StringStatsType::EXACT_STATS ||
-	    StringStats::GetMaxType(*haystack_stats) != StringStatsType::EXACT_STATS) {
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25547, opened 2026-09-10
Skip parse/plan/optimize for Connection::BeginTransaction/Commit/Rollback

dentiny:
Hi team, I'm playing with WAL replay and found for small transactions (i.e., BEGIN; single row append; COMMIT;) replay latency is linear in the number of transactions

Example benchmark
```py
for n in (10000, 50000, 200000):
    with open(f'/tmp/wal_bench/A{n}.sql', 'w') as f:
        f.write("CREATE TABLE t(a INT, b VARCHAR);\n")
        for i in range(n):
            f.write(f"BEGIN; INSERT INTO t VALUES ({i}, 'row_{i}'); COMMIT;\n")
```
Checkpoint on shutdown is disabled, WAL auto checkpoint is set to an unreachably high value to guarantee replay takes all the insertions and transactions.

I did a flamegraph on WAL replay, from the rendered graph, we could easily tell most of the time is spent on `BeginTransaction` and `Commit`, with planner + optimizer + executor work inside, which is not necessary for WAL replay. The actual replay work takes only ~10% of the CPU time.
image

The commit pseudocode looks like (asked opus-4.7 to draw the callstack), begin transaction looks alike
```sh
con.Commit()                                       ← what we call
 └─ Query("COMMIT")                                ← Connection::Commit implementation
     └─ Parser::ParseQuery("COMMIT")
     └─ Binder::Bind
     └─ Planner::CreatePlan
     └─ Optimizer::Optimizer(binder, context)      ← every rewrite rule is constructed!
        (ContainsToInClauseRule, EnumCompareSimplification,
         MoveConstantsRule, StringPrefixRule, ...)
     └─ PhysicalPlanGenerator::CreatePlan
     └─ Executor::ExecuteTask
        └─ PhysicalTransaction::GetDataInternal(COMMIT)
           └─ switch(type): case COMMIT:
                if (IsAutoCommit()) throw ...
                client.transaction.Commit();       ← the one line that does real work
                client.SuppressInterrupts();
```

So in this PR, I propose to simplify the replay to skip all the redundant work and only include necessary work
- Start a transaction with `TransactionContext::SetAutoCommit(false)`
- Commit a transaction with `TransactionContext::Commit`

Performance comparison before and after the change
| Scenario                        | Before   | After     | Speedup    |
|---------------------------------|---------:|----------:|-----------:|
| 10 k single-row transactions    | 0.72 s   | **0.11 s**| **6.5×**   |
| 50 k single-row transactions    | 3.65 s   | **0.59 s**| **6.2×**   |
| 200 k single-row transactions   | 14.16 s  | **2.04 s**| **6.9×**   |
| 2000 CREATE/INSERT/DROP cycles  | 0.48 s   | **0.06 s**| **8.0×**   |
| 5 M-row single INSERT           | 0.02 s   | 0.02 s    | no change  |
| 200 k rows + 10× UPDATE-all     | 0.05 s   | 0.05 s    | no change  |

### DIFF
--- a/src/execution/operator/helper/physical_transaction.cpp
+++ b/src/execution/operator/helper/physical_transaction.cpp
@@ -1,89 +1,12 @@
 #include "duckdb/execution/operator/helper/physical_transaction.hpp"
 
-#include "duckdb/common/exception/transaction_exception.hpp"
 #include "duckdb/main/client_context.hpp"
-#include "duckdb/main/config.hpp"
-#include "duckdb/main/database_manager.hpp"
-#include "duckdb/main/valid_checker.hpp"
-#include "duckdb/transaction/meta_transaction.hpp"
-#include "duckdb/transaction/transaction_manager.hpp"
-#include "duckdb/main/settings.hpp"
 
 namespace duckdb {
 
 SourceResultType PhysicalTransaction::GetDataInternal(ExecutionContext &context, DataChunk &chunk,
                                                       OperatorSourceInput &input) const {
-	auto &client = context.client;
-
-	auto type = info->type;
-	if (type == TransactionType::COMMIT && ValidChecker::IsInvalidated(client.ActiveTransaction())) {
-		// transaction is invalidated - turn COMMIT into ROLLBACK
-		type = TransactionType::ROLLBACK;
-	}
-	switch (type) {
-	case TransactionType::BEGIN_TRANSACTION: {
-		if (client.transaction.IsAutoCommit()) {
-			// start the active transaction
-			// if autocommit is active, we have already called
-			// BeginTransaction by setting autocommit to false we
-			// prevent it from being closed after this query, hence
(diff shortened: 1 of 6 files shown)


Mytherin:
Thanks - very interesting / nice change. I wonder if there's a path to instead optimize `Connection::BeginTransaction` and `Connection::Commit` to no longer issue a SQL statement? That would benefit more code paths rather than just WAL replay.

dentiny:
> I wonder if there's a path to instead optimize Connection::BeginTransaction and Connection::Commit to no longer issue a SQL statement?

Hmmm let me check that out

dentiny:
> I wonder if there's a path to instead optimize Connection::BeginTransaction and Connection::Commit to no longer issue a SQL statement?

Yeah I think that make sense, I moved the optimization into `Connection`

Mytherin:
Thanks!

PR #25551, opened 2026-09-10
Fix external file cache spill

dentiny:
Closes https://github.com/duckdblabs/duckdb-internal/issues/10700

The root cause is, spilled external cache blocks were reloaded as `MANAGED_BUFFER` type, conflicting with the `EXTERNAL_FILE` type retained by `BlockMemory`, which fails the type assertion [here](https://github.com/duckdb/duckdb/blob/668f94740b280bf44c7c081df664ae1eb310c090/src/storage/buffer/block_handle.cpp#L36). In this PR, I construct the destination buffer with the retained type before reading the temporary file into it.


Mytherin:
Thanks!

PR #25555, opened 2026-09-10
Fix row_number gaps in vector-sized parallel scans

### LINKED ISSUE #25553: ROW_NUMBER() OVER () produces gaps with PRAGMA verify_parallelism on a base table
### What happens?

`ROW_NUMBER() OVER ()` returns non-contiguous numbers when scanning an ordinary table with `PRAGMA verify_parallelism` enabled. For 2,050 rows, the maximum row number is 2,052 instead of 2,050. Disabling the optimizer produces the expected result.

This is reproducible in the standalone DuckDB CLI without the GQL extension. The SQL below uses the unmodified, 1-based `ROW_NUMBER()` result (no subtraction).

### To Reproduce

Run in a fresh in-memory database:

```sql
PRAGMA verify_parallelism;
CREATE TABLE t AS SELECT i FROM range(2050) r(i);

SELECT count(*) AS row_count,
       min(rn) AS min_rn,
       max(rn) AS max_rn,
       count(DISTINCT rn) AS distinct_rn
FROM (SELECT row_number() OVER () AS rn FROM t);
```

Actual:

```text
row_count  min_rn  max_rn  distinct_rn
2050       1       2052    2050
```

Expected:

```text
row_count  min_rn  max_rn  distinct_rn
2050       1       2050    2050
```

Control, in the same connection:

```sql
PRAGMA disable_optimizer;
SELECT count(*) AS row_count,
       min(rn) AS min_rn,
       max(rn) AS max_rn,
       count(DISTINCT rn) AS distinct_rn
FROM (SELECT row_number() OVER () AS rn FROM t);
```

This returns the expected `2050, 1, 2050, 2050`. Omitting `PRAGMA verify_parallelism` in a fresh connection also returns the expected result for this small reproducer.

#### Suspected cause

The scan pushdown introduced in #21971 appears relevant. `RowNumberRewriter` replaces the window with the scan's virtual row-number column.

In `RowGroupCollection::NextParallelScan` (`src/storage/table/row_group_collection.cpp`), `verify_parallelism` splits a row group into vector-sized scan assignments, but `state.row_number_base` advances by `current_row_group.GetVisibleRowCount(tx)` for every assignment, i.e. the visible count of the entire row group. For 2,050 rows split into 2,048 + 2, the second assignment therefore starts at SQL row number 2,051 instead of 2,049.

A possible fix is to advance the base by the number of transaction-visible rows in the assigned range, preserving deletion/MVCC semantics rather than simply adding its physical row count. This is a source-based diagnosis; no patch has been applied or validated.

Related #23586 / #23667 concern file-local row numbers from `read_duckdb`. This reproducer instead uses an ordinary base table.

### OS:

macOS 26.3.1, arm64

### DuckDB Version:

`v2.0.0-dev84235`, source ID `516be109ab` (reported by `PRAGMA version`). 

### DuckDB Client:

Standalone CLI, RelWithDebInfo build (`build/reldebug/duckdb`).

### Hardware:

Apple M1 max,32GB

### Full Name:

ted-wq-x

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

ted-wq-x:
Fixes #25553.

With `PRAGMA verify_parallelism`, `ROW_NUMBER() OVER ()` on a 2,050-row table returns a maximum of 2,052 instead of 2,050. Each vector-sized scan assignment advances the row-number base by the visible count of the entire row group.

Count transaction-visible rows within each assignment instead. Preserve the existing whole-row-group counting interfaces and reuse the current visibility checks and locking.

### DIFF
--- a/src/include/duckdb/storage/table/row_group.hpp
+++ b/src/include/duckdb/storage/table/row_group.hpp
@@ -206,6 +206,8 @@ public:
 	idx_t GetCommittedRowCount();
 	//! Returns the number of rows visible to the given transaction
 	idx_t GetVisibleRowCount(TransactionData transaction);
+	//! Count visible rows in a scan assignment starting at a vector boundary
+	idx_t GetVisibleRowCount(TransactionData transaction, idx_t start_vector, idx_t scan_count);
 	bool CanReuseMetadata(RowGroupWriter &writer) const;
 	RowGroupWriteData WriteToDisk(RowGroupWriter &writer);
 	RowGroupPointer Checkpoint(RowGroupWriteData write_data, RowGroupWriter &writer, TableStatistics &global_stats,
--- a/src/include/duckdb/storage/table/row_version_manager.hpp
+++ b/src/include/duckdb/storage/table/row_version_manager.hpp
@@ -26,6 +26,8 @@ public:
 
 	//! Returns the number of non-deleted rows in this segment
 	idx_t GetRowCount(ScanOptions options, idx_t count);
+	//! Count visible rows starting at a vector boundary, including a partial final vector
+	idx_t GetRowCount(ScanOptions options, idx_t start_vector, idx_t count);
 
 	idx_t GetSelVector(ScanOptions options, idx_t vector_idx, SelectionVector &sel_vector, idx_t max_count);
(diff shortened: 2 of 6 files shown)


Mytherin:
Thanks!

PR #25557, opened 2026-09-10
Drain pending tasks in the TaskExecutor destructor

carlopi:
Tasks hold a reference to the TaskExecutor, and in the threaded build destroying the ProducerToken does not remove queued tasks from the scheduler queue - a worker can still dequeue and run them. Unwinding between ScheduleTask and WorkOnTasks therefore left tasks running against a destroyed executor, and against whatever the task references in the caller frame.

Drain in the destructor so that the tasks can never outlive the executor. Tasks that have not started yet bail out immediately.

---

I have not reproed any actual issue, but the shape of the problem would be like (pseudocode)
```c++
TaskExecutor executor(...);
for (int i=0; i<50; i++) {
      executor.ScheduleTask(...);
      if (i == 42) throw std::exception();
}
executor.WorkOnTask();
```
now, executor would be destroyed, while some 40 tasks have been scheduled, their callback will reach for a non-available destructor, and bad things might happen.

There are callsites (like https://github.com/duckdb/duckdb/blob/v2.0-cyanoptera/src/common/serializer/async_task_queue.cpp#L307) that protect from this manually, probably since it hit already, but attempting a best effort `CancelAndDrain()` in the destructor allows via RAII to ensure this in a better way. I have not changed any callsite yet, could be done as follow up.

Bumped into this while working on a task pointed by @lnkuiper, https://github.com/duckdb/duckdb/pull/25560, but not a blocker for that.

### DIFF
--- a/src/parallel/task_executor.cpp
+++ b/src/parallel/task_executor.cpp
@@ -14,6 +14,12 @@ TaskExecutor::TaskExecutor(ClientContext &context_p, TaskSchedulerType type_p)
 }
 
 TaskExecutor::~TaskExecutor() {
+	// tasks can still be queued if we unwound between scheduling them and draining them
+	// they hold a reference to this executor, so they must not outlive it
+	try {
+		CancelAndDrain();
+	} catch (...) { // NOLINT
+	}
 }
 
 void TaskExecutor::PushError(ErrorData error) {


Mytherin:
Thanks!

Mytherin:
I think a better design would be for tasks to not have a reference to the `TaskExecutor` at all but to its internal state (that's a `shared_ptr`, e.g. `shared_ptr<TaskExecutorState>`). Then the task executor being destroyed itself would not matter. I think the current solution is still pretty error-prone given another exception would still cause the same issue.

carlopi:
> I think a better design would be for tasks to not have a reference to the `TaskExecutor` at all but to its internal state (that's a `shared_ptr`, e.g. `shared_ptr<TaskExecutorState>`). Then the task executor being destroyed itself would not matter. I think the current solution is still pretty error-prone given another exception would still cause the same issue.

Actually, I am not sure I see this the same way, I tried to make sure that a Task is always notified of either execution or cancellation, basically making a throw not influencing this invariant, basically `CancelAndDrain` will always cancel all tasks no after https://github.com/duckdb/duckdb/pull/25644.

There is also a different issue as play: there are cases where state is still stack based / local to a function, and allowing Tasks to outlive a given function does increase complexity (that can be fixed, but I assume that might be currently be off / unexpected in various places)
* Having TaskExecutor itself to outlive a function / scope might not necessarily have a semantic in all callsites

I would propose https://github.com/duckdb/duckdb/pull/25644 as solution to the concern, and there is also a discussion at https://github.com/duckdblabs/duckdb-internal/issues/10745

PR #25558, opened 2026-09-10
Fix correlated CTE reference bindings by column identity

### LINKED ISSUE #25554: INTERNAL Error: correlated USING KEY recursion with recurring subquery binds different STRUCT columns incorrectly
### What happens?

A correlated recursive CTE using `USING KEY` and a subquery over `recurring.t` fails with an internal column-binding type mismatch when two outer columns have different STRUCT types.

This reproduces in the CLI on `v2.0.0-dev84244`, source ID `668f94740b` (upstream commit 668f94740b280bf44c7c081df664ae1eb310c090). The same SQL returns the expected result on v1.5.1 (7dbb2e646f).

### To Reproduce

Run in a fresh in-memory CLI session, e.g. `duckdb -init /dev/null :memory:`. No extensions, tables, or external data are required.

```sql
SELECT *
FROM (SELECT {x: 1} AS a, {y: 3} AS b) seed,
LATERAL (
    WITH RECURSIVE t(n) USING KEY (n) AS (
        SELECT a.x
        UNION ALL
        SELECT n + 1
        FROM t
        WHERE n < b.y
          AND NOT EXISTS (
              SELECT 1 FROM recurring.t r WHERE r.n = b.y
          )
    )
    SELECT * FROM t
) q
ORDER BY n;
```

Actual result:

```text
INTERNAL Error: Failed to bind column reference "b" [23.1]: inequal types (STRUCT(y INTEGER) != STRUCT(x INTEGER))
```

Expected result (also verified on v1.5.1):

```text
a         b         n
{'x': 1}  {'y': 3}   1
{'x': 1}  {'y': 3}   2
{'x': 1}  {'y': 3}   3
```

The error also reproduces when running `PRAGMA disable_optimizer;` before the query in a fresh connection.

Additional observations:

- Removing the `NOT EXISTS` subquery over `recurring.t` makes the query succeed.
- Using the same field name for both outer structs (`{x: 1} AS a, {x: 3} AS b`, and replacing `b.y` with `b.x`) makes it succeed.
- The failure reaches `ColumnBindingResolver::VisitReplace(BoundColumnRefExpression &,...)`. I have not established the exact root cause.

### OS:

macOS 26.3.1, arm64.

### DuckDB Version:

- Fails: v2.0.0-dev84244, source ID 668f94740b.
- Works: v1.5.1, source ID 7dbb2e646f, installed CLI. Both default and optimizer-disabled execution return the expected three rows.

### DuckDB Client:

CLI, in-memory database.

### Hardware:

Apple M1 max, 32GB

### Related reports

I found #23315 (correlated LIST / recursive CTE inside LATERAL) and #22691 (BIGINT/INTEGER mismatch with LATERAL and recursive CTEs), both closed. Their reproducers differ from this keyed-recursion / recurring-table case; I have not established whether the underlying cause is related.

### Full Name:

ted-wq-x

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

ted-wq-x:
## Summary

Fixes #25554.

Correlated recursive CTEs can store outer columns in a different order from that expected by a nested dependent join. Previously, CTE references assumed matching positions when enough correlated columns were available, causing errors such as `STRUCT(y INTEGER) != STRUCT(x INTEGER)`.

Match correlated columns by binding identity using the existing alias resolution. Reuse the mapping for both direct CTE reads and domain join conditions, and select the direct path only when every required column is available.

### DIFF
--- a/src/include/duckdb/planner/subquery/flatten_dependent_join.hpp
+++ b/src/include/duckdb/planner/subquery/flatten_dependent_join.hpp
@@ -97,8 +97,9 @@ private:
 	                             bool include_names) const;
 	void AddDelimColumnsToGroup(LogicalAggregate &aggr, const vector<ColumnBinding> &state) const;
 	void AddCorrelatedFirstAggregates(LogicalAggregate &aggr, const vector<ColumnBinding> &state) const;
+	vector<optional_idx> GetCTERefCorrelatedPositions(const LogicalCTERef &cteref) const;
 	void AddCTERefJoinConditions(LogicalComparisonJoin &join, const LogicalCTERef &cteref,
-	                             const vector<ColumnBinding> &state) const;
+	                             const vector<optional_idx> &positions, const vector<ColumnBinding> &state) const;
 	void AddCorrelatedJoinConditions(LogicalJoin &join, const vector<ColumnBinding> &left_state,
 	                                 const vector<ColumnBinding> &right_state) const;
 	vector<ColumnBinding> CreateDelimCrossProduct(unique_ptr<LogicalOperator> &plan,
--- a/src/planner/subquery/flatten_dependent_join.cpp
+++ b/src/planner/subquery/flatten_dependent_join.cpp
@@ -514,11 +514,17 @@ vector<optional_idx> FlattenDependentJoins::GetCTERefCorrelatedPositions(const L
 	}
 	auto rec_cte = binder.recursive_ctes.find(cteref.cte_index);
 	if (rec_cte == binder.recursive_ctes.end()) {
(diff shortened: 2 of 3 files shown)


kryonix (commented):
The design looks sound to me. Matching by binding identity fixes the positional assumption, and sharing the mapping keeps direct reads and domain joins consistent. The internal exceptions are appropriate because correlated readers must already have registered CTE metadata. My only suggestion is to include the CTE index and relevant column counts in those messages to make failures easier to diagnose.

ted-wq-x:
Done

PR #25560, opened 2026-09-10
TaskExecutor: allow single parked task in (opt-in) JOINED mode

carlopi:
This has been discussed with @lnkuiper in the context of improving this pattern:
```c++
	TaskExecutor executor(scheduler, TaskSchedulerType::ASYNC);

	for (idx_t idx = 0; idx < num_blocks; idx++) {
		executor.ScheduleTask(make_uniq<FetchBlockTask>(*this, executor, context,
		                                                external_file_cache.GetBufferManager(), blocks[idx],
		                                                first_block + idx, block_size, pins[idx]));
	}
	executor.WorkOnTasks();
```
Here tasks will be handed out, most likely to IO-threads, while this thread (likely a regular one) will stay idle waiting for the others to come back.
A minor optimization, that is though easy to roll, is: let's park the first produced task, and have WorkOnTask first perform the parked task, and only then wait on the others (possibly picking them up more in the process).

This is in most cases invisible, since in most cases bottleneck is either compute or network, and this matters only in balanced cases, say for example you have 4 compute threads + 40 IO ones. And say that by chance you got a single fetch task. Before, you would block 1 compute thread + 1 IO one, with the first just idle waiting for the second. After the PR, compute thread will do the fetch itself, freeing 1 IO thread.

This is relevant only in minor cases, but it was still a smell and flagged as improvements.

PR here adds an optional `TaskExecutorMode`, by default `TaskExecutorMode::BACKGROUND`, but one can opt in to `TaskExecutorMode::JOINED` where first produced task is parked and done in the same thread doing the scheduling.

### DIFF
--- a/src/parallel/task_executor.cpp
+++ b/src/parallel/task_executor.cpp
@@ -31,12 +31,12 @@ void TaskExecutor::ThrowError() {
 void TaskExecutor::ScheduleTask(unique_ptr<Task> task) {
 	{
 		const annotated_lock_guard<annotated_mutex> lock(token->producer_lock);
+		++total_tasks;
 		if (mode == TaskExecutorMode::JOINED && !parked_task) {
 			// the joining thread has to wait for the other tasks anyway - park this one for it to execute
 			parked_task = std::move(task);
 			return;
 		}
-		++total_tasks;
 	}
 	try {
 		scheduler.ScheduleTask(*token, std::move(task), type);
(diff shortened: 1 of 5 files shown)


src/parallel/task_executor.cpp
```diff
@@ -31,6 +31,11 @@ void TaskExecutor::ThrowError() {
 void TaskExecutor::ScheduleTask(unique_ptr<Task> task) {
 	{
 		const annotated_lock_guard<annotated_mutex> lock(token->producer_lock);
+		if (mode == TaskExecutorMode::JOINED && !parked_task) {
```
dentiny, line 35: I think there's a behavior change.
When there's new tasks being scheduled on `DrainTasks`, old behavior will increment total task count and block wait all tasks to finish; while under the JOIN mode, new task could be assigned to `parked_task` and potentially never checked again.

I would suggest to increment `total_tasks` for tasks when they get enqueued, instead of the parked task to be executed.

carlopi: Thanks for pointing this out. I will review how `total_tasks` is used, how / if needs to be adapted.

One thing for which not bumping right away would make some sense is that `parked` is not really available from a different thread.

I'll review more properly, this might be path dependent, let me check.

carlopi: Thanks, should be ready for another go.

src/parallel/task_executor.cpp
```diff
@@ -32,6 +32,11 @@ void TaskExecutor::ScheduleTask(unique_ptr<Task> task) {
 	{
 		const annotated_lock_guard<annotated_mutex> lock(token->producer_lock);
 		++total_tasks;
+		if (mode == TaskExecutorMode::JOINED && !parked_task) {
+			// the joining thread has to wait for the other tasks anyway - park this one for it to execute
+			parked_task = std::move(task);
```
dentiny, line 37: Since control thread's block waiting for notification at L70, I'm wondering if we should notify?
```suggestion
			parked_task = std::move(task);
			token->producer_cv.notify_one();
```

dentiny (commented):
Would also appreciate if we could add new C++ unit tests to cover new cases:
- Case-1: Control thread enqueue one task + block wait completion
- Case-2: New task enqueued when another task is ongoing

src/parallel/task_executor.cpp
carlopi: I might have realized something is off in a rework I've done, and needs to check more properly, that is: is parked task guaranteed to be handled by calling thread (I am not sure it's guaranteed right now). Let me check this properly, I think my last push might have broken this. And indeed C++ tests are good to have.

carlopi:
As is it can't be merged, I need to think about this some more.

PR #25561, opened 2026-09-10
Fix column pruning of sibling inputs beside materialized CTEs

### LINKED ISSUE #25556: Unused projection survives column pruning beside a correlated MATERIALIZED CTE (regression from 1.5.5)
### What happens?

An unreferenced `struct_pack` expression survives both optimized logical and physical planning in a query with a correlated MATERIALIZED CTE and another join branch. The expression belongs to the other join branch, outside the CTE, and is explicitly omitted by its enclosing SELECT.

The same standalone SQL prunes the expression on v1.5.5 (d8cdaa33fd) and v1.5.1 (7dbb2e646f), but retains it on a locally built v2.0.0-dev84235 (516be109ab65ebb828b38aa851fe36727a22364a). No extensions are loaded. Query results are correct; this report concerns unnecessary projection evaluation. I have not measured a runtime or memory slowdown with this four-row example, and have not bisected the first affected commit.

### To Reproduce

Run in a fresh in-memory CLI session with default settings, e.g. `duckdb -init /dev/null :memory:`:

```sql
CREATE TABLE v(id INTEGER PRIMARY KEY);
INSERT INTO v VALUES (1), (2), (3), (4);

PRAGMA explain_output = 'all';

EXPLAIN
SELECT b.value
FROM (SELECT id AS k, 0 AS domain FROM v) a
JOIN LATERAL (
    WITH t AS MATERIALIZED (SELECT a.k AS n)
    SELECT n FROM t
) q ON true
JOIN (
    SELECT b.k, b.domain, b.value
    FROM (
        SELECT id AS k, 0 AS domain,
               struct_pack(id := id) AS unused_payload,
               id AS value
        FROM v
    ) b
) b ON b.k = q.n
   AND b.domain IS NOT DISTINCT FROM a.domain
WHERE a.k = 1;
```

Inspect the projection over the scan of `v` in the `b` branch. `unused_payload` is present in the unoptimized plan as expected. After optimization:

| Version | Optimized logical projection | Physical projection |
| --- | --- | --- |
| v1.5.5, d8cdaa33fd | k, domain, value | k, domain, value |
| v2.0.0-dev84235, 516be109ab | k, domain, unused_payload, value | k, domain, unused_payload, value |

Expected: prune `unused_payload`, since it is never referenced outside its defining SELECT. Removing EXPLAIN returns the single value `1` on both versions.

The example needs no recursion, external data, or extensions. Related reports found while searching: #20958 concerns columns inside a materialized pass-through CTE; #24109 concerns unused recursive CTE outputs. This example instead retains an unused expression in a separate join branch, and does not reproduce on v1.5.5.

### OS:

macOS 26.3.1 (a), arm64

### DuckDB Version:

Affected: locally built v2.0.0-dev84235, 516be109ab65ebb828b38aa851fe36727a22364a.
Not reproduced: v1.5.5 (d8cdaa33fd), v1.5.1 (7dbb2e646f).

### DuckDB Client:

Standalone CLI, in-memory database, no extensions loaded

### Hardware:

Apple M1 Max, 32 GiB RAM

### Full Name:

ted-wq-x

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes — all four rows are created by the SQL above.

ted-wq-x:
## Summary

A materialized CTE that cannot be pruned sets `everything_referenced` on the shared column-pruning visitor. This can prevent pruning unused expressions in subsequently visited sibling join inputs.

Keep the conservative behavior local to the CTE input through its existing child visitor. Preserve all CTE pruning guards and leave the parent visitor's incoming state unchanged.

Fixes #25556.

### DIFF
--- a/src/optimizer/remove_unused_columns.cpp
+++ b/src/optimizer/remove_unused_columns.cpp
@@ -697,10 +697,7 @@ void RemoveUnusedColumns::VisitOperator(unique_ptr<LogicalOperator> &op_ref) {
 		// Distinct column indexes here, unlike the per-reader map, so comparing sizes is a valid width check.
 		if (cte_map_entry.everything_referenced || readers_visible_in_output ||
 		    referenced_columns_in_rhs.size() == cte.children[0]->GetColumnBindings().size()) {
-			if (!analyze) {
-				everything_referenced = true;
-			}
-			// We may opt out here, but we still need to traverse the left-hand side of the CTE.
+			// Preserve the CTE input without marking unrelated sibling outputs as referenced.
 			RemoveUnusedColumns remove(*this, true);
 			remove.VisitOperator(cte.children[0]);
 			return;
--- a/test/optimizer/materialized_cte_sibling_column_pruning.test
+++ b/test/optimizer/materialized_cte_sibling_column_pruning.test
@@ -1,0 +1,137 @@
+# name: test/optimizer/materialized_cte_sibling_column_pruning.test
+# description: CTE input preservation must not prevent pruning unrelated join inputs
+# group: [optimizer]
+
+statement ok
+SET debug_verify_column_bindings = true;
+
+statement ok
+CREATE TABLE v(id INTEGER PRIMARY KEY);
+
+statement ok
+INSERT INTO v VALUES (1), (2), (3), (4);
+
+# Exercise the original join order as well as the default optimizer order.
+foreach disabled '' 'join_order,build_side_probe_side'
+
+statement ok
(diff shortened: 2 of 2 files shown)


kryonix (approved):
Looks good to me. The child visitor already preserves all CTE input columns, so removing the assignment keeps that protection local without affecting sibling pruning. I confirmed the reproducer is fixed and found no regressions in the CTE tests or additional join and NULL cases.

Mytherin:
Thanks!

PR #25564, opened 2026-09-10
TaskExecutor: allow interruptible drain

carlopi:
Single commit on top of https://github.com/duckdb/duckdb/pull/25560, since both change the same file.

----

Commit diff viewable at https://github.com/duckdb/duckdb/commit/4d2504965101ece9805701ce65a993f40e6105f7

DrainTasks waited on the producer condition variable with no timeout and no interrupt check, so a thread that had handed all of its work to other threads was unreachable: cancelling the query, or hitting max_execution_time, could not get it out of the wait.

Wait for a bounded time instead and check for interruption in between waits, when the executor was given a client context. A drain that cancels never checks, because it must always run to completion - it is what keeps tasks from outliving the executor. An interrupt thrown out of WorkOnTasks unwinds into the destructor, which cancels and drains, so the tasks that have not started bail out immediately.

Pass the query context through in CachingFileHandle::Read, which is the one JOINED user that constructs its executor from the scheduler.

### DIFF
--- a/src/common/enum_util.cpp
+++ b/src/common/enum_util.cpp
@@ -183,6 +183,7 @@
 #include "duckdb/parallel/pipeline_schedule.hpp"
 #include "duckdb/parallel/scan_read_ahead.hpp"
 #include "duckdb/parallel/task.hpp"
+#include "duckdb/parallel/task_executor.hpp"
 #include "duckdb/parser/constraint.hpp"
 #include "duckdb/parser/expression/lambda_expression.hpp"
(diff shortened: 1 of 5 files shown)


carlopi:
Currently opted in in a single location, but this would be beneficial to be reviewed (together with previous PR) also whenever the pattern is used.

carlopi:
I realized it also depends on https://github.com/duckdb/duckdb/pull/25557, let's merge both the preconditions, then figure it out.

PR #25566, opened 2026-09-10
Specify node256 leaf method as `const`

dentiny:
Hi team, I found the inconsistency between different node leaves here
https://github.com/duckdb/duckdb/blob/f2895015f7041a3830cd793d3d212e9549f50624/src/execution/index/art/node.cpp#L301-L306

I think having `const` actually improves performance and IO operations
- When we [get node reference](https://github.com/duckdb/duckdb/blob/f2895015f7041a3830cd793d3d212e9549f50624/src/include/duckdb/execution/index/art/node.hpp#L135-L140), whether the leaf node is const or not decides the `dirty` argument's value
- [Whether or not the node is dirty](https://github.com/duckdb/duckdb/blob/f2895015f7041a3830cd793d3d212e9549f50624/src/include/duckdb/execution/index/fixed_size_allocator.hpp#L59-L77) decides whether the buffer needs to be flushed in WAL and checkpoint, even if there's no actual change


Mytherin:
Thanks!

PR #25570, opened 2026-09-10
Allow Pinning/Unpinning of ColumnReader::block

maiadegraaf:
`ColumnReader::block`, the buffer holding each page's decompressed data, was previously allocated with an `Allocator` and was always kept pinned.

This moves `block` onto the buffer manager (`BlockHandle`/`BufferHandle`), pinning it only around each vector's read and unpinning it in between. `ByteBuffer` was changed to use an offset-based cursor instead of an absolute pointer. Further, every decoder that caches a pointer into `block` (dictionary, RLE, delta-binary-packed, byte-stream-split) now rebases itself against the new address after each re-pin.

One buffer was deliberately left on `Allocator`: the bloom filter built while writing a dictionary page. It's short-lived and never re-pinned, so buffer-manager tracking would only add overhead there with no eviction benefit.

I ran a small benchmark locally, no speed regression, and it now keeps working once memory gets tight enough that main just runs out of memory.

Also includes a small unrelated build fix in `expression_binder.cpp` I hit along the way.

This looks like a lot of changes, but a lot of the changes are because many of the functions were renamed for formatting reasons (ie. `inc()` -> `Inc()`).


extension/parquet/column_reader.cpp
```diff
@@ -742,6 +749,32 @@ bool ColumnReader::PrepareRead(idx_t read_now, data_ptr_t define_out, data_ptr_t
 	return true; // No defines, so everything is valid
 }
 
+void ColumnReader::RebaseDecoders() {
+	// Block's offset has moved on so we need to use the decoders own stored offset to calculate current location.
+	if (repeated_decoder) {
+		repeated_decoder->Rebase(block->GetPtr() + repeated_decoder_offset);
```
Mytherin, line 762: This seems a bit error-prone - if we forget to call this rebase we run into issues. Would it be possible for the decoders to instead refer to the block and get the pointer from there instead of storing one?

Mytherin (commented):

extension/parquet/column_reader.cpp
maiadegraaf: Doing it via block directly would mean every `Read()`/`Skip()` in these decode loops resolves the pointer through an extra indirection instead of a direct dereference. I could add an extra `D_ASSERT` instead to catch a missing `Rebase()`?

Mytherin: Maybe the pointer can instead be resolved on `Pin()` instead of being kept around in these decoders?

I'm just a bit worried about caching these pointers as the current code looks rather fragile - it seems easy to forget calling a `Rebase()` or to forget calling a `Pin()` only to be hit with crashes, and I would prefer if the code would protect against this directly.

extension/parquet/decoder/delta_byte_array_decoder.cpp
```diff
@@ -23,42 +23,49 @@ class Vector;
...
 	idx_t prefix_count, suffix_count;
+
 	auto &prefix_buffer = reader.encoding_buffers[0];
+	prefix_buffer.Pin(buffer_manager);
+	ReadDbpData(buffer_manager, block, prefix_buffer, prefix_count);
+
 	auto &suffix_buffer = reader.encoding_buffers[1];
-	ReadDbpData(allocator, block, prefix_buffer, prefix_count);
-	ReadDbpData(allocator, block, suffix_buffer, suffix_count);
+	suffix_buffer.Pin(buffer_manager);
```
Mytherin: Instead of having to remember to `Pin / Unpin` can we use an object with a destructor similar to our buffer manager (a `BufferHandle` like object).

Mytherin (commented):
Thanks - looks better but I still think the API seems error prone, one more comment

PR #25573, opened 2026-09-10
[1.5] Report the real storage version when opening a v2.0+ database file

DinosL:
Files written by `v2.0.0` and later have `DEPRECATED_VERSION_NUMBER` (999) as their main header version number with the real storage version in the database headers. Older versions read that as an actual version leading to a confusing error. 

This PR improves the error message so that when the main header has the `DEPRECATED_VERSION_NUMBER`, the version is read from the database header instead.


src/storage/single_file_block_manager.cpp
```diff
@@ -654,6 +659,11 @@ void SingleFileBlockManager::LoadExistingDatabase(QueryContext context) {
 	ReadAndChecksum(context, header_buffer, Storage::FILE_HEADER_SIZE * 2ULL);
 	h2 = DeserializeDatabaseHeader(main_header, header_buffer.buffer);
 
+	if (main_header.version_number == MainHeader::DEPRECATED_VERSION_NUMBER) {
+		auto &active_db_header = h1.iteration > h2.iteration ? h1 : h2;
```
DinosL, line 663: The active header check is repeated later. Could potentially reduce it to one check

carlopi (approved):
LGTM, thanks!

Mytherin:
Thanks!

Issue #7162, opened 2023-04-20
CSV reader - add option to convert quoted values to NULL values 


Mytherin:
Currently we explicitly do not compare quoted values to the provided `nullstr` in the CSV reader:

```cpp
if ((!has_quotes || return_types[column].id() != LogicalTypeId::VARCHAR) && !options.force_not_null[column] &&
     Equals::Operation(str_val, string_t(options.null_str))) {
       FlatVector::SetNull(parse_chunk.data[column], row_entry, true);
....
```

The reason for this is that it is impossible to round-trip certain data sets in CSV files without this behavior, as we use the quotes to distinguish between e.g. `NULL` and the string value `'NULL'`.

However, it turns out many tools do quote null values, and other tools like Pandas do not support this behavior. For compatibility reasons, I propose moving this behavior behind a flag (`allow_quoted_nulls`) which defaults to true to match Pandas and other tools' behavior.

##### test.csv
```csv
int,str
42,""
,a
88,
```

##### pandas

```py
import pandas as pd
pd.read_csv('test.csv')
    int  str
0  42.0  NaN
1   NaN    a
2  88.0  NaN
```

##### current behavior
```sql
FROM test.csv;
┌───────┬─────────┐
│  int  │   str   │
│ int64 │ varchar │
├───────┼─────────┤
│    42 │         │
│  NULL │ a       │
│    88 │ NULL    │
└───────┴─────────┘
```

##### desired behavior
```sql
-- this should become the default behavior
FROM test.csv;
┌───────┬─────────┐
│  int  │   str   │
│ int64 │ varchar │
├───────┼─────────┤
│    42 │ NULL    │
│  NULL │ a       │
│    88 │ NULL    │
└───────┴─────────┘

SELECT * FROM read_csv_auto('test.csv', allow_quoted_nulls=False);
┌───────┬─────────┐
│  int  │   str   │
│ int64 │ varchar │
├───────┼─────────┤
│    42 │         │
│  NULL │ a       │
│    88 │ NULL    │
└───────┴─────────┘
SELECT * FROM read_csv_auto('test.csv', allow_quoted_nulls=True);
┌───────┬─────────┐
│  int  │   str   │
│ int64 │ varchar │
├───────┼─────────┤
│    42 │ NULL    │
│  NULL │ a       │
│    88 │ NULL    │
└───────┴─────────┘
```

##### side note - this should also allow us to auto-detect the settings for this file correctly:
```csv
int,str
42,""
"",a
88,
[...]

PR #25578, opened 2026-09-10
Make EXPORT DATABASE round-trip quoted empty strings and null strings

### LINKED ISSUE #25501: EXPORT DATABASE / IMPORT DATABASE round-trip loses the NULL vs empty-string distinction at default settings
### What happens?

`EXPORT DATABASE` writes `""` for an empty string and a bare empty field for `NULL`, deliberately distinguishing the two. The `load.sql` it generates does not set `allow_quoted_nulls false`, so `IMPORT DATABASE` reads both back as `NULL`.

The two commands are documented as a pair — *"The `EXPORT DATABASE` command allows you to export the contents of the database to a specific directory. The `IMPORT DATABASE` command allows you to then read the contents again."* — CSV is the default format, and the page carries no fidelity caveat. So the default round trip silently changes data.

### The contract

Quoting is the escape: a value that would otherwise read as the null marker is quoted, and quoting means "this is data, not the marker". `EXPORT DATABASE` writes exactly that. Under `NULLSTR 'NULL'` the four cases are:

| on disk | should read back as |
|---|---|
| `NULL` | a real null |
| `"NULL"` | the string `NULL` |
| `""` | an empty string |
| *(bare empty field)* | an empty string |

With `allow_quoted_nulls false` all four hold. With the `load.sql` that `EXPORT DATABASE` generates, row 2 collapses to a null.

Under the default export the marker is the empty field instead, so the two cases are `""` → empty string and *(bare)* → null — and there, row 1 of that pair is the one that collapses.

### To Reproduce

```sql
-- session 1
CREATE TABLE t(v VARCHAR);
INSERT INTO t VALUES (NULL), ('');
EXPORT DATABASE '/tmp/out';
```

`/tmp/out/t.csv` — the writer distinguishes them correctly:

```
v

""
```

`/tmp/out/load.sql`:

```sql
COPY t FROM '/tmp/out/t.csv' (FORMAT 'csv', quote '"', delimiter ',', header 1);
```

```sql
-- session 2, fresh database
IMPORT DATABASE '/tmp/out';
SELECT count(*) FILTER (WHERE v IS NULL) AS nulls,
       count(*) FILTER (WHERE v = '')   AS empties FROM t;
```

| nulls | empties |
|------:|--------:|
| 2 | 0 |

Expected `1 | 1`.

The same applies with an explicit `NULLSTR`: `EXPORT DATABASE '/tmp/out' (FORMAT csv, NULLSTR 'NULL')` writes the literal string `'NULL'` as `"NULL"` — again quoted precisely to disambiguate — and `IMPORT DATABASE` returns it as `NULL`.

### Not a duplicate of #7162

#7162 made the reader's behaviour intentional: `allow_quoted_nulls` defaults to `true` to match Pandas, and that issue explicitly notes it *"is impossible to round-trip certain data sets in CSV files without this behavior, as we use the quotes to distinguish between e.g. `NULL` and the string value `'NULL'`."*

That default is reasonable for `read_csv` over foreign files. This report is narrower: `EXPORT`/`IMPORT` is DuckDB reading its own output, where the quoting is known to be meaningful because DuckDB just wrote it. The compatibility default is being inherited by a path that has no compatibility problem to solve.

Nothing new is needed to fix it. #7162's own *desired behavior* section already shows the preserving read, and kept it reachable on purpose:

```sql
SELECT * FROM read_csv_auto('test.csv', allow_quoted_nulls=False);
┌───────┬─────────┐
│  int  │   str   │
├───────┼─────────┤
│    42 │         │   -- the quoted empty string, preserved
└───────┴─────────┘
```

The ask here is only that the `load.sql` written by `EXPORT DATABASE` sets that flag, since it is the one caller that always knows the quoting came from DuckDB.

### Suggested fix

Have `EXPORT DATABASE` emit `allow_quoted_nulls false` in the `COPY` statements it writes into `load.sql`. That is local to the generated file, changes no default, and makes the documented round trip lossless.

### Workaround

Either rewrite the generated `load.sql`:

```
sed -i "s/header 1)/header 1, allow_quoted_nulls false)/" out/load.sql
```

or export with `(FORMAT parquet)`, which round-trips exactly.

Parquet is the right answer when the artifact never leaves the database world. It is the wrong one when it does: a CSV/TSV export can be `grep`ed, diffed against last week's, opened in an editor, reviewed in a pull request, or hand-corrected in an emergency — and none of that is true of a parquet file. That is the whole reason to reach for the text format, and the reason a silent NULL/empty-string collapse is worth fixing rather than routing around.

It is also very close to working already. `EXPORT DATABASE` writes the quoting correctly today; the only gap is that the `load.sql` it generates does not ask the reader to honour it.

### OS

macOS 15 (arm64)

### DuckDB Version

v2.0.0-alpha38195

### DuckDB Client

CLI

saime428:
Fixes #25501.

`EXPORT DATABASE` followed by `IMPORT DATABASE` with the default CSV format reads every empty string back as `NULL`:

```sql
CREATE TABLE t(v VARCHAR);
INSERT INTO t VALUES (NULL), (''), ('NULL');
EXPORT DATABASE 'out';
-- fresh database
IMPORT DATABASE 'out';
SELECT count(*) FILTER (WHERE v IS NULL), count(*) FILTER (WHERE v = ''), count(*) FILTER (WHERE v = 'NULL') FROM t;
-- 2 | 0 | 1, expected 1 | 1 | 1
```

`EXPORT DATABASE 'out' (FORMAT CSV, NULLSTR 'NULL')` on the same table gives `2 | 0 | 0`: the empty string and the string `'NULL'` both come back as `NULL`.

`test/sql/copy/csv/test_export_not_null.test` already shows the first case: it exports four empty strings and expects the nullable column to come back as `NULL`.

### Cause

`CSVWriter::RequiresQuotes` quotes any value equal to the null string, so the writer emits `""` for an empty string (or `"NULL"` under `NULLSTR 'NULL'`) and a bare marker for `NULL`. The `load.sql` generated by `EXPORT DATABASE` does not set `allow_quoted_nulls false`, so the reader's default of `true` (chosen in #7162 for foreign CSV files) turns the quoted value back into `NULL`.

A second gap makes the `FORMAT CSV` case worse. `EXPORT DATABASE` keeps the format string as written, while `COPY` lowercases it (`bind_copy.cpp`). Both places that add CSV-specific read options compare against `"csv"`, so with `FORMAT CSV` the generated `load.sql` carries neither the `header`, `delimiter` and `quote` defaults nor the `force_not_null` list, and the import falls back to sniffing. On main the `NULLSTR` example above produces `COPY t FROM 't.csv' (FORMAT 'CSV', nullstr 'NULL');`.

### Change

- `WriteCopyStatement` adds `allow_quoted_nulls false` to the `COPY ... FROM` statements it writes for CSV exports, next to the `force_not_null` list it already adds. `allow_quoted_nulls` is a read-only CSV option, so `EXPORT DATABASE` rejects it and there is no user value to override. The reader default does not change.
- `Binder::Bind(ExportStatement)` lowercases the format before the CSV checks, matching `COPY`. `FORMAT CSV` now gets the same `load.sql` options as `FORMAT csv`.

With the change, the first example generates `COPY t FROM 't.csv' (FORMAT 'csv', allow_quoted_nulls false, quote '"', delimiter ',', header 1);` and returns `1 | 1 | 1`.

### Tests

- `test/sql/export/export_quoted_nulls.test` (new) round-trips `NULL`, `''` and `'NULL'` with the default options and, on a two-column table, with `(FORMAT CSV, NULLSTR 'NULL')`, and checks that `load.sql` contains the flag.
- `test/sql/copy/csv/test_export_not_null.test`: the last expected row changes from `(empty) (empty) (empty) NULL` to four `(empty)`.

On main, the new test fails at the `load.sql` check; with that check removed, the default round trip returns `2 | 0 | 1`, and reading the two-column CSV with the options main writes into `load.sql` turns row 3 (`'NULL'`) into a real `NULL`.

### Validation

- `make reldebug` (WSL2, gcc), then `build/reldebug/test/unittest "test/sql/export/*"` (25 cases) and `"test/sql/copy/csv/test_export*"`: all pass.
- `build/reldebug/test/unittest` (fast suite): 6286 cases, 6268 passed, 18 failed. 17 of the failures are CSV sniffing/rejects tests that fail on this Windows-mounted checkout because the CSV fixtures had CRLF line endings; after normalizing the fixtures to LF they all pass (`"test/sql/copy/csv/*"`: 311/312, the remaining one is a symlink-based `.test_slow`). The 18th, `test/sql/secrets/create_secret_filesystem_logging.test`, fails on a secret-file permission check that the mounted filesystem cannot satisfy. None of the 18 use `EXPORT`/`IMPORT DATABASE`.
- `clang-format` 11.0.1 on both changed source files: no changes.
- Not run: `make allunit`, `test_configs`, `test_vector`.

### Notes

Under `NULLSTR 'NULL'`, an empty string in a single-column table is still written as an empty line, which the reader skips as a blank line. That is writer/reader behaviour for blank lines and is unchanged here; the new test uses a two-column table for the `NULLSTR` case so the empty string sits in a field.

### DIFF
--- a/src/execution/operator/persistent/physical_export.cpp
+++ b/src/execution/operator/persistent/physical_export.cpp
@@ -71,6 +71,8 @@ static void WriteCopyStatement(FileSystem &fs, stringstream &ss, CopyInfo &info,
 		for (auto &not_null_column : exported_table.not_null_columns) {
 			info.options["force_not_null"].push_back(not_null_column);
 		}
+		// the writer quotes empty strings and the null string so they can be told apart from NULL
+		info.options["allow_quoted_nulls"] = {Value::BOOLEAN(false)};
 	}
 	for (auto &copy_option : info.options) {
 		if (copy_option.second.empty()) {
--- a/src/planner/binder/statement/bind_export.cpp
+++ b/src/planner/binder/statement/bind_export.cpp
@@ -278,6 +278,7 @@ BoundStatement Binder::Bind(ExportStatement &stmt) {
 	                                          stmt.info->GetQualifiedName().Name()));
 	// prepare the options for export
 	auto &format = stmt.info->format;
+	format = StringUtil::Lower(format);
 	auto &options = stmt.info->options;
 	if (format == "csv") {
 		// insert default csv options, if not specified
--- a/test/sql/copy/csv/test_export_not_null.test
+++ b/test/sql/copy/csv/test_export_not_null.test
@@ -56,4 +56,4 @@ IMPORT DATABASE '{TEST_DIR}/multiple';
 query IIII
 select * from tbl_2
 ----
-(empty)	(empty)	(empty)	NULL
+(empty)	(empty)	(empty)	(empty)
--- a/test/sql/export/export_quoted_nulls.test
+++ b/test/sql/export/export_quoted_nulls.test
@@ -1,0 +1,66 @@
+# name: test/sql/export/export_quoted_nulls.test
+# description: Test that EXPORT/IMPORT DATABASE round-trips NULL, empty strings and the null string
+# group: [export]
+
+statement ok
+BEGIN TRANSACTION
+
+statement ok
+CREATE TABLE tbl(v VARCHAR)
+
+statement ok
+INSERT INTO tbl VALUES (NULL), (''), ('NULL')
+
+statement ok
+EXPORT DATABASE '{TEST_DIR}/export_quoted_nulls'
+
+statement ok
+ROLLBACK
+
+# the generated load.sql must not read the quoted values back as NULL
+query I
+SELECT contains(content, 'allow_quoted_nulls false') FROM read_text('{TEST_DIR}/export_quoted_nulls/load.sql')
+----
+true
+
+statement ok
+IMPORT DATABASE '{TEST_DIR}/export_quoted_nulls'
+
+query III
+SELECT count(*) FILTER (WHERE v IS NULL), count(*) FILTER (WHERE v = ''), count(*) FILTER (WHERE v = 'NULL') FROM tbl
+----
+1	1	1
+
+# same with an explicit null string: the string 'NULL' is written quoted and must survive the round trip
+statement ok
+DROP TABLE tbl
+
+statement ok
+BEGIN TRANSACTION
+
+statement ok
+CREATE TABLE tbl2(i INTEGER, v VARCHAR)
+
+statement ok
+INSERT INTO tbl2 VALUES (1, NULL), (2, ''), (3, 'NULL')
+
+statement ok
+EXPORT DATABASE '{TEST_DIR}/export_quoted_nulls_nullstr' (FORMAT CSV, NULLSTR 'NULL')
+
+statement ok
+ROLLBACK
+
+statement ok
+IMPORT DATABASE '{TEST_DIR}/export_quoted_nulls_nullstr'
+
+query III
+SELECT i, v IS NULL, length(v) FROM tbl2 ORDER BY i
+----
+1	true	NULL
+2	false	0
+3	false	4
+
+query III
+SELECT count(*) FILTER (WHERE v IS NULL), count(*) FILTER (WHERE v = ''), count(*) FILTER (WHERE v = 'NULL') FROM tbl2
+----
+1	1	1


saime428:
The `extensions / Deploy extensions` failure is `make check-extension-entries` refusing to load the autocomplete extension over a version-string mismatch. The `workflow_dispatch` run on `main` from earlier today ([34729581746](https://github.com/duckdb/duckdb/actions/runs/34729581746)) fails at the same step with the same error, and `main` has no commits since 2026-09-11, so it does not come from this branch.

PR #25581, opened 2026-09-10
ART Batch Scan

artjomPlaunov:
https://github.com/duckdblabs/duckdb-internal/issues/9231 

Since this API was requested, this PR was trying to find a place where it may be worth introducing a batch scan API into core. This turned out to be scanning multiple equality lookups on an ART. Instead of decomposing these into multiple equality expressions and setting up a scan state + key generation for each predicate, we can do this in bulk: one scan state + key generation step for multiple equality predicates.

The ART scan state now distinguishes between equality, range, and batch equality scans. For a batch equality scan, the caller supplies a DataChunk containing the values to look up. The existing ART key generation code converts those values into encoded keys. We then look up each key and append its matching row IDs to a single row-ID collection for the entire batch.

The table scan code now extracts the equality values directly into that chunk. We still apply the same row-ID limit: if the batch exceeds it, we clear the collected row IDs and fall back to a table scan.

The internal search helpers also return an `ARTSearchResult` instead of a bool, to make the distinction between `COMPLETED` and `CAPACITY_EXCEEDED` clearer. A key with no matching rows is still a completed search; it doesn't stop the rest of the batch. The iterator's result is named `ARTScanProgress` to distinguish its scan progress from the overall search result. The existing public `ART::Scan` interface keeps its bool return value; this should be a follow up PR (didn't do it here, since it would require some patching in extensions most likely). 

### Benchmarks

Measured on an Apple M1 Pro with 4 threads, using 100-million-row tables with unique ART indexes and distributed keys. All queries below used ART index scans in both builds.

The latest rerun reproduces the earlier improvements. Within the BIGINT tests, increasing the batch size from 50 to 2,048 keys increased the measured speedup from 9.6% to 17.1% (as expected for a larger batch amortizing the scan preparation). The other key types also benefited from batching.

| Query | Before (ms) | After (ms) | Improvement |
|---|---:|---:|---:|
| Single BIGINT key | 0.259 | 0.258 | 0.6% |
| BIGINT IN, 50 keys | 0.828 | 0.749 | 9.6% |
| BIGINT IN, 100 keys | 1.256 | 1.106 | 12.0% |
| BIGINT IN, 500 keys | 5.056 | 4.260 | 15.7% |
| BIGINT IN, 2,048 keys | 19.585 | 16.229 | 17.1% |
| INNER JOIN against 50 BIGINT keys | 0.602 | 0.511 | 15.1% |
| BIGINT range (`id < 200`) | 0.447 | 0.466 | −4.0% |
| INTEGER IN, 100 keys | 1.253 | 1.076 | 14.1% |
| VARCHAR IN, 100 keys | 3.077 | 2.903 | 5.7% |
| DOUBLE IN, 100 keys | 1.317 | 1.138 | 13.6% |
| DECIMAL(18,2) IN, 100 keys | 1.681 | 1.491 | 11.3% |


src/execution/index/art/art.cpp
```diff
@@ -32,15 +32,19 @@
 
 namespace duckdb {
 
+enum class ARTScanType : uint8_t { FULL, PREDICATE, BATCH_EQUALITY };
```
taniabogatsch: Should we add comments to the `ENUM` on what these mean? I didn't find them fully intuitive. 🤔

src/execution/index/art/art.cpp
```diff
@@ -278,8 +282,17 @@ unique_ptr<IndexScanState> ART::TryInitializeScan(const Expression &expr, const
 	return InitializeScanSinglePredicate(high_value, high_comparison_type);
 }
 
+unique_ptr<IndexScanState> ART::InitializeBatchScan(unique_ptr<DataChunk> values) const {
```
taniabogatsch: nit: maybe we call this chunk `key_columns` instead of values, as it contains the keys we want to scan?

src/execution/index/art/art.cpp
```diff
@@ -278,8 +282,17 @@ unique_ptr<IndexScanState> ART::TryInitializeScan(const Expression &expr, const
...
+	if (!values || values->GetTypes() != logical_types) {
+		throw InternalException("ART batch scan keys must have the index's logical types");
+	}
+	auto result = make_uniq<ARTIndexScanState>(ARTScanType::BATCH_EQUALITY);
+	result->batch_equality_values = std::move(values);
```
taniabogatsch: Should we just have a separate constructor that takes the type and the chunk?

src/execution/index/art/art.cpp
```diff
@@ -802,13 +815,74 @@ bool ART::SearchCloseRange(const ARTKey &lower_bound, const ARTKey &upper_bound,
 	return it.Scan(upper_bound, row_ids, right_equal) == ARTScanResult::COMPLETED;
 }
 
+bool ART::ScanChunk(DataChunk &input, RowIdVectorOutput &row_ids) const {
```
taniabogatsch: I've always wondered if we want to "enum" this `bool` return value as well... WDYT? I guess we would break some extensions? So maybe in a different PR?

src/execution/index/art/art.cpp
```diff
@@ -802,13 +815,74 @@ bool ART::SearchCloseRange(const ARTKey &lower_bound, const ARTKey &upper_bound,
...
+	if (HasLegacyGeometryKeys()) {
+		DataChunk converted;
+		ConvertKeyInput(input, converted);
+		GenerateKeys<>(arena, converted, keys);
+	} else {
+		GenerateKeys<>(arena, input, keys);
+	}
+	IndexLock guard(*this);
+	for (const auto &key : keys) {
+		if (!key.Empty() && !SearchEqual(key, row_ids)) {
```
taniabogatsch: Is this correct? So we search until we find a key for which there is no row ID? Don't we just want to gather all row IDs matching the batch of keys?

src/execution/index/art/art.cpp
```diff
@@ -802,13 +815,74 @@ bool ART::SearchCloseRange(const ARTKey &lower_bound, const ARTKey &upper_bound,
...
+			return false;
+		}
+	}
+	return true;
+}
+
+bool ART::ScanBatch(DataChunk &values, RowIdVectorOutput &row_ids) const {
```
taniabogatsch: Do we want this to happen here? I wonder if it would make more sense for the caller to loop over its values and send batches of `STANDARD_VECTOR_SIZE` to the ART? Actually, if we just loop over it and return the rows, why do we need to split this by the standard vector size at all?

src/function/table/table_scan.cpp
```diff
@@ -815,36 +815,45 @@ static bool ExtractValuesFromExpression(const Expression &expr, value_set_t &val
...
+	auto &expr_filter = ExpressionFilter::GetExpressionFilter(filter, "TryInitializeIndexScan");
+	ColumnBinding binding(TableIndex(0), storage_index);
+	BoundColumnRefExpression bound_ref(col.Name(), col.Type(), binding);
 
-	// Extract all exact values we can derive from the filter tree.
-	vector<unique_ptr<Expression>> expressions;
 	value_set_t values;
-	if (ExtractValuesFromExpression(*expr_filter.expr, values)) {
-		ExtractExpressionsFromValues(values, *bound_ref, expressions);
+	if (!ExtractValuesFromExpression(*expr_filter.expr, values)) {
```
taniabogatsch: I wonder if we already want to try and extract the values directly into a vector here? That would save us an extra copy on all the values, no?

src/function/table/table_scan.cpp
```diff
@@ -815,36 +815,45 @@ static bool ExtractValuesFromExpression(const Expression &expr, value_set_t &val
...
+	if (!column.Equals(index_expr)) {
+		return nullptr;
+	}
+	for (const auto &value : values) {
+		if (value.type() != column.GetReturnType()) {
+			return nullptr;
+		}
+	}
+	auto value_chunk = make_uniq<DataChunk>();
+	value_chunk->Initialize(Allocator::DefaultAllocator(), {column.GetReturnType()}, values.size());
```
taniabogatsch: I assume that we use a data chunk here to prepare for the case where the index/key can be compounded from multiple columns? Maybe we can add a note?

test/sql/index/art/scan/test_batch_scan.test
```diff
@@ -0,0 +1,52 @@
...
+CREATE UNIQUE INDEX idx ON t(i);
+
+statement ok
+SET dynamic_or_filter_threshold = 100;
+
+statement ok
+SET disabled_optimizers = 'compressed_materialization';
+
+statement ok
+CREATE VIEW in_lookup AS SELECT i FROM t WHERE i IN (
```
taniabogatsch, line 18: Cries in SQL. 😉 

```sql
CREATE VIEW in_lookup AS SELECT i FROM t WHERE i IN (
   SELECT 1 + 10 * j FROM range(100) r(j)
);
```

taniabogatsch (changes requested):
Hi @artjomPlaunov, thanks for the PR! I've left some comments.

src/execution/index/art/art.cpp
artjomPlaunov: done, also, I split up PREDICATE into RANGE and EQUALITY, does that sound good?

src/execution/index/art/art.cpp
artjomPlaunov: what about making it return an enum internally for private methods and expose it in a future PR?

src/execution/index/art/art.cpp
artjomPlaunov: I think the NULL key check was not necessary, here is how it looks now with the enum :-) 

```cpp
	for (const auto &key : keys) {
		D_ASSERT(!key.Empty());
		if (SearchEqual(key, row_ids) == ARTLookupResult::CAPACITY_EXCEEDED) {
			return ARTLookupResult::CAPACITY_EXCEEDED;
		}
	}
	return ARTLookupResult::COMPLETED;
```

src/execution/index/art/art.cpp
artjomPlaunov: hmm I think you're right, for some reason I was thinking that maybe this would benefit if we reach the limit, but generated all the keys in advance. Perhaps we can just have a simpler implementation at first, and consider this if it every actually becomes an issue though.

src/function/table/table_scan.cpp
artjomPlaunov: Ah, I didn't consider this, it was just for the existing GenerateKeys API. But this seems like a natural benefit as well :-)

src/execution/index/art/art.cpp
artjomPlaunov: also, I got rid of FULL since that seemed to be introduced for checkpoints, but I should have removed it when I added InsertMerge and RemovalMerge

taniabogatsch: > I split up PREDICATE into RANGE and EQUALITY, does that sound good?

Yes, sounds great!

src/execution/index/art/art.cpp
taniabogatsch: Love the enums!

src/execution/index/art/art.cpp
```diff
@@ -726,49 +747,41 @@ bool ART::HasLegacyGeometryKeys() const {
...
-		return true;
-	}
-	Iterator it(*this);
-	it.FindMinimum(tree);
-	const auto empty_key = ARTKey();
-	return it.Scan(empty_key, row_ids, false) == ARTScanResult::COMPLETED;
-}
-
-bool ART::SearchEqual(const ARTKey &key, RowIdVectorOutput &row_ids) const {
+ARTLookupResult ART::SearchEqual(const ARTKey &key, RowIdVectorOutput &row_ids) const {
```
taniabogatsch: Minor nit: Should we use `ARTSearchResult`?

src/execution/index/art/art.cpp
```diff
@@ -726,49 +747,41 @@ bool ART::HasLegacyGeometryKeys() const {
...
 	if (!leaf) {
-		return true;
+		return ARTLookupResult::COMPLETED;
 	}
 
 	Iterator it(*this);
 	it.FindMinimum(leaf.Get());
 	const auto empty_key = ARTKey();
-	return it.Scan(empty_key, row_ids, false) == ARTScanResult::COMPLETED;
+	return it.Scan(empty_key, row_ids, false) == ARTScanResult::COMPLETED ? ARTLookupResult::COMPLETED
```
taniabogatsch: And can we rename `ARTScanResult` to `ARTScanProgress` maybe?

taniabogatsch (approved):

PR #25583, opened 2026-09-10
[WIP] Fix duplication index creation

### LINKED ISSUE #25582: INTERNAL error when creating an index conflicting with an internal index name
### What happens?

Hi team, I found when I create an index with the index name same as an internal index (i.e., primary key, unique constraint, foreign key, etc), the index creation seems to go through and crashes sometimes. See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE p(id INTEGER PRIMARY KEY);
memory D CREATE TABLE c(id INTEGER, FOREIGN KEY (id) REFERENCES p(id));
memory D INSERT INTO p VALUES (1);
memory D INSERT INTO c VALUES (1);
memory D CREATE INDEX IF NOT EXISTS FOREIGN_c_0 ON c(id);
memory D DROP INDEX FOREIGN_c_0;
memory D DELETE FROM p WHERE id = 1;
INTERNAL Error:
TableIndexList::VerifyForeignKey failed to find foreign key index
```

### OS:

macos

### DuckDB Version:

6ee2305dc786e2429c85eb971f8854f5a9807149

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25582

Two pre-knowledge for this issue:
- An invariant for DuckDB: each index should have its unique name.
- Primary key, unique constraint, and foreign key all create their own index, and are named as `PRIMARY_<table>_<ordinal>`, `UNIQUE_<table>_<ordinal>`, `FOREIGN_<table>_<ordinal>`

In the current implementation, we could mistakenly create two index with the same indexes with the same name.

### DIFF
--- a/src/catalog/catalog_entry/duck_schema_entry.cpp
+++ b/src/catalog/catalog_entry/duck_schema_entry.cpp
@@ -293,8 +293,11 @@ optional_ptr<CatalogEntry> DuckSchemaEntry::CreateIndex(CatalogTransaction trans
 
 	// currently, we can not alter PK/FK/UNIQUE constraints
 	// concurrency-safe name checks against other INDEX catalog entries happens in the catalog
-	if (info.on_conflict != OnCreateConflict::IGNORE_ON_CONFLICT &&
-	    !table.GetStorage().IndexNameIsUnique(info.GetIndexName().GetIdentifierName())) {
+	// constraint indexes only exist in table storage, so they need a separate conflict check
+	if (!table.GetStorage().IndexNameIsUnique(info.GetIndexName().GetIdentifierName())) {
+		if (info.on_conflict == OnCreateConflict::IGNORE_ON_CONFLICT) {
+			return nullptr;
+		}
 		throw CatalogException("An index with the name " + info.GetIndexName() + " already exists!");
 	}
 
--- a/src/execution/operator/schema/physical_create_index.cpp
+++ b/src/execution/operator/schema/physical_create_index.cpp
@@ -155,8 +155,10 @@ SinkFinalizeType PhysicalCreateIndex::Finalize(Pipeline &pipeline, Event &event,
 			return SinkFinalizeType::READY;
 		}
 
-		auto index_entry = schema.CreateIndex(schema.GetCatalogTransaction(context), *info, table).get();
-		D_ASSERT(index_entry);
+		auto index_entry = schema.CreateIndex(schema.GetCatalogTransaction(context), *info, table);
+		if (!index_entry) {
+			return SinkFinalizeType::READY;
+		}
 		auto &index = index_entry->Cast<DuckIndexEntry>();
 		index.initial_index_size = bound_index->GetInMemorySize();
 
--- a/test/sql/index/art/storage/test_art_if_not_exists_constraint_name.test
+++ b/test/sql/index/art/storage/test_art_if_not_exists_constraint_name.test
@@ -1,0 +1,48 @@
+# name: test/sql/index/art/storage/test_art_if_not_exists_constraint_name.test
+# description: Prevent IF NOT EXISTS from reusing hidden constraint index names
+# group: [storage]
+
+statement ok
+CREATE TABLE parent (id INTEGER PRIMARY KEY)
+
+statement ok
+CREATE TABLE child (id INTEGER REFERENCES parent(id))
+
+statement ok
+INSERT INTO parent VALUES (1)
+
+statement ok
+INSERT INTO child VALUES (1)
+
+statement ok
+CREATE INDEX IF NOT EXISTS FOREIGN_child_0 ON child(id)
+
+# The no-op CREATE INDEX must not make the hidden foreign-key index droppable.
+statement error
+DROP INDEX FOREIGN_child_0
+----
+<REGEX>:Catalog Error.*Index with name FOREIGN_child_0 does not exist.*
+
+statement error
+DELETE FROM parent WHERE id = 1
+----
+<REGEX>:Constraint Error.*still referenced by a foreign key.*
+
+statement ok
+CREATE TABLE constrained (id INTEGER PRIMARY KEY, value INTEGER UNIQUE)
+
+statement ok
+CREATE INDEX IF NOT EXISTS PRIMARY_constrained_0 ON constrained(id)
+
+statement ok
+CREATE INDEX IF NOT EXISTS UNIQUE_constrained_1 ON constrained(value)
+
+# IF NOT EXISTS must continue to ignore ordinary catalog index collisions.
+statement ok
+CREATE TABLE regular (id INTEGER)
+
+statement ok
+CREATE INDEX regular_id ON regular(id)
+
+statement ok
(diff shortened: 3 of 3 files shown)


artjomPlaunov:
Can you double check this test, and see if it's a regression with the current changes: 

```sql
CREATE TABLE t(i INTEGER);
CREATE UNIQUE INDEX idx ON t(i);

BEGIN;
DROP INDEX idx;
CREATE UNIQUE INDEX IF NOT EXISTS idx ON t(i);
COMMIT;

SELECT count(*) FROM duckdb_indexes() WHERE index_name = 'idx';
INSERT INTO t VALUES (1), (1);
```

test/sql/index/art/create_drop/test_art_drop_recreate_index_same_transaction.test
```diff
@@ -0,0 +1,34 @@
+# name: test/sql/index/art/create_drop/test_art_drop_recreate_index_same_transaction.test
+# description: Test recreating a dropped index with IF NOT EXISTS in the same transaction
+# group: [create_drop]
+
+# TODO: WAL serialization can select the dropped unbound index after a forced reload.
+require skip_reload
```
dentiny, line 6: This is a pre-existing issue in v2.0 branch before this PR, I will handle separately.
```sql
haojiang@HaoJiangs-MacBook-Pro$ ./build/reldebug/duckdb /tmp/another_repro.db
DuckDB v2.0.0-dev10468 (Development Version, 02e71e5e97)
another_repro D CREATE TABLE t(i INTEGER);
another_repro D CREATE UNIQUE INDEX idx ON t(i);
another_repro D CHECKPOINT;
another_repro D .quit

haojiang@HaoJiangs-MacBook-Pro$ ./build/reldebug/duckdb /tmp/another_repro.db
DuckDB v2.0.0-dev10468 (Development Version, 02e71e5e97)
another_repro D BEGIN;
another_repro D DROP INDEX idx;
another_repro D CREATE UNIQUE INDEX IF NOT EXISTS idx ON t(i);
another_repro D COMMIT;
libc++abi: Pure virtual function called!

💣 Program crashed: Signal 6: Backtracing from 0x1902fe5e8...Attempt to set cursor outside bounds -- stopping DWARFanalysis of /Users/haojiang/Desktop/duckdb/build/reldebug/src/libduckdb_static.a(ub_duckdb_optimizer_join_order.cpp.o)
```

dentiny:
Hi @artjomPlaunov thanks for the review! I think it does cover a regression introduced by this PR, could you please take another look when you have some time? Thank you!

src/catalog/catalog_entry/duck_schema_entry.cpp
```diff
@@ -294,9 +294,17 @@ optional_ptr<CatalogEntry> DuckSchemaEntry::CreateIndex(CatalogTransaction trans
...
-	    !table.GetStorage().IndexNameIsUnique(info.GetIndexName().GetIdentifierName())) {
-		throw CatalogException("An index with the name " + info.GetIndexName() + " already exists!");
+	// constraint indexes only exist in table storage, so they need a separate conflict check
+	if (!table.GetStorage().IndexNameIsUnique(info.GetIndexName().GetIdentifierName())) {
+		if (info.on_conflict != OnCreateConflict::IGNORE_ON_CONFLICT) {
+			throw CatalogException("An index with the name %s already exists!",
+			                       info.GetIndexName().GetIdentifierName());
+		}
+		auto &indexes = GetCatalogSet(CatalogType::INDEX_ENTRY);
+		auto entry = indexes.GetHeadEntry(info.GetIndexName());
```
artjomPlaunov, line 304: This check leads to another issue: 

```sql
CREATE TABLE t(i INTEGER PRIMARY KEY);
CREATE TABLE other(i INTEGER);

CREATE INDEX PRIMARY_t_0 ON other(i);

BEGIN;
DROP INDEX PRIMARY_t_0;
CREATE INDEX IF NOT EXISTS PRIMARY_t_0 ON t(i);
COMMIT;

SELECT table_name, index_name FROM duckdb_indexes();
-- Actual: t, PRIMARY_t_0
-- Expected under this patch's policy: no rows.

DROP INDEX PRIMARY_t_0;
-- Actual: succeeds.
-- Expected: catalog error, since creation should have been skipped.

INSERT INTO t VALUES (1), (1);
-- Actual: succeeds, violating the primary key.
-- Expected: constraint error.

SELECT * FROM t;
-- Actual:
-- 1
-- 1
```

artjomPlaunov: (the catalog head could be a deletion marker for an index on another table).

src/catalog/catalog_entry/duck_schema_entry.cpp
```diff
@@ -294,9 +294,17 @@ optional_ptr<CatalogEntry> DuckSchemaEntry::CreateIndex(CatalogTransaction trans
 
 	// currently, we can not alter PK/FK/UNIQUE constraints
 	// concurrency-safe name checks against other INDEX catalog entries happens in the catalog
-	if (info.on_conflict != OnCreateConflict::IGNORE_ON_CONFLICT &&
-	    !table.GetStorage().IndexNameIsUnique(info.GetIndexName().GetIdentifierName())) {
-		throw CatalogException("An index with the name " + info.GetIndexName() + " already exists!");
+	// constraint indexes only exist in table storage, so they need a separate conflict check
+	if (!table.GetStorage().IndexNameIsUnique(info.GetIndexName().GetIdentifierName())) {
```
artjomPlaunov, line 298: This is existing, but I'm wondering if the name of "NameIsUnique" is a little unclear or misleading? It reports a collision for any matching unique or foreign key physical index, but ignores matching non-unique indexes. Maybe something like HasUniqueOrForeignIndexNamed with an inverted return, perhaps? 

Calling it a constraint index check could still be misleading because it does not distinguish a unique constraint’s index from a user-created unique index.

artjomPlaunov (changes requested):
One more bug found by codex, I'm going to open up a separate issue internally for discussion and link you to it, since I think we should handle this properly on main, but since 2.0 is coming up I think we just get these more localized bug fixes in on 2.0.

dentiny:
> One more bug found by codex, I'm going to open up a separate issue internally for discussion and link you to it, since I think we should handle this properly on main, but since 2.0 is coming up I think we just get these more localized bug fixes in on 2.0.

Hi @artjomPlaunov the farther I go the more cornercases I found to resolve index; I think constraint and index check will get simplified a lot if we get index and constraint OID checkin first, convert to draft first, should be achievable before v2.0.

artjomPlaunov:
@dentiny Sounds good, with all these edge cases popping up maybe better to just implement the unique identities for constraint indexes which I think would avoid these issues in the first place

dentiny:
I will put a PR for unique constraint OID refactor later today

PR #25584, opened 2026-09-10
Cleanup deprecated loop iterator

dentiny:
Saw test warning as below, as my agents to do a repo-wise cleanup.
```sh
Replacing deprecated loop iterator ${j} in test "test/sql/storage/metadata_reuse_optimistic_write.test_slow" - please use the new loop iterator {j}
Replacing deprecated loop iterator ${j} in test "test/sql/storage/meta
```


Mytherin:
Thanks!

PR #25585, opened 2026-09-10
Prune row group for greatest with all NULL

dentiny:
Hi team, I found in the current implementation, `greatest`/`least` function doesn't prune row groups with all NULL values; for example,
```sql
memory D CREATE TABLE metrics AS SELECT CASE WHEN i < 202752 THEN NULL ELSE i END AS value FROM range(204800) t(i);

-- should only access the second row group
memory D EXPLAIN ANALYZE SELECT sum(value) FROM metrics WHERE greatest(value, 0) > 100;
╭─ Summary ───────────╮
│ Total Time: 0.0017s │
╰─────────────────────╯
│ Aggregates: sum\_no\_overflow(#0)   │
╭─ Projection ────┚┖────────────────╮
│ Projections: value                │
│ 2,048 rows                    0µs │
│ Table: memory.main.metrics        │
│ Projections: value                │
│ Filters: greatest(value, 0) > 100 │
│ Row Groups Scanned: 2 / 2         │
│ 2,048 rows                    0µs │
```

### DIFF
--- a/src/function/scalar/generic/constant_or_null.cpp
+++ b/src/function/scalar/generic/constant_or_null.cpp
@@ -3,6 +3,7 @@
 #include "duckdb/common/operator/comparison_operators.hpp"
 #include "duckdb/execution/expression_executor.hpp"
 #include "duckdb/planner/expression/bound_function_expression.hpp"
+#include "duckdb/storage/statistics/base_statistics.hpp"
 
 namespace duckdb {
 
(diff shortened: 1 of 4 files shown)


src/function/scalar/generic/constant_or_null.cpp
```diff
@@ -70,6 +71,16 @@ static void ConstantOrNullFunction(DataChunk &args, ExpressionState &state, Vect
 	}
 }
 
+static FilterPropagateResult ConstantOrNullFilterPrune(const FunctionStatisticsPruneInput &input) {
+	for (idx_t i = 0; i < input.function.GetChildren().size(); i++) {
+		auto stats = input.ChildStats(i);
+		if (stats && !stats->CanHaveNoNull()) {
+			return FilterPropagateResult::FILTER_FALSE_OR_NULL;
```
Mytherin, line 78: This seems incorrect to me - `constant_or_null` emits either the constant value or `NULL`. We should therefore check what the constant value is to know what the propagate result is, e.g.:

```sql
select constant_or_null(false, x) as const_false, constant_or_null(true, x) as const_true from (values (42), (null)) t(x);
┌─────────────┬────────────┐
│ const_false │ const_true │
│   boolean   │  boolean   │
├─────────────┼────────────┤
│ false       │ true       │
│ NULL        │ NULL       │
└─────────────┴────────────┘
```

Mytherin: I also wonder if we need a separate filter prune callback or if a stats propagate callback is sufficient?

Mytherin (commented):

PR #25587, opened 2026-09-10
Fix rolled-back DELETE poisoning uncheckpointed_delete_commit

dentiny:
TLDR: In the current implementation, whenever we have a rollback-ed deletion operation, row group checkpoint cannot enter into fast path.

Background introduction
- `uncheckpointed_delete_commit` is used to decide whether the current row group has uncommitted delete operations, and could be used to [reuse metadata](https://github.com/duckdb/duckdb/blob/0691eeb9d699640f272ab4e52a7aaf2ec8afe85c/src/storage/table/row_group.cpp#L1615) on checkpoint
- Checkpoint reusing metadata is the fast path and is extremely beneficial to write performance, for example, it contains segment stats (i.e., min/max/count) and compression method, etc, so we don't need to re-calculate

In the current implementation, 
- Deletion rollback is implemented as record deletion commit id as [`NOT_DELETE_ID`](https://github.com/duckdb/duckdb/blob/0691eeb9d699640f272ab4e52a7aaf2ec8afe85c/src/transaction/rollback_state.cpp#L37-L42), and [updates the `uncheckpointed_delete_commit`](https://github.com/duckdb/duckdb/blob/0691eeb9d699640f272ab4e52a7aaf2ec8afe85c/src/storage/table/row_version_manager.cpp#L202-L204) regardless
- Later on, even if we perform a checkpoint operation, the `uncheckpointed_delete_commit` won't be reset because `NOT_DELETE_ID` is the [largest integer](https://github.com/duckdb/duckdb/blob/0691eeb9d699640f272ab4e52a7aaf2ec8afe85c/src/common/constants.cpp#L22) and [visibility bound, as a normal transaction id, won't be able to reset](https://github.com/duckdb/duckdb/blob/0691eeb9d699640f272ab4e52a7aaf2ec8afe85c/src/storage/table/row_version_manager.cpp#L277-L281)
- Conclusion: all later checkpoints won't be able to enter into the fast path

In this PR, I propose a small code change, `NOT_DELETE_ID` is a well-known constant to represent unreachable deletion commit id, we special handle it and skip updating `uncheckpointed_delete_commit`.

### DIFF
--- a/src/storage/table/row_version_manager.cpp
+++ b/src/storage/table/row_version_manager.cpp
@@ -199,8 +199,12 @@ idx_t RowVersionManager::DeleteRows(idx_t vector_idx, transaction_t transaction_
 void RowVersionManager::CommitDelete(idx_t vector_idx, transaction_t commit_id, const DeleteInfo &info) {
 	lock_guard<mutex> lock(version_lock);
 	needs_compression_check = true;
-	if (!uncheckpointed_delete_commit.IsValid() || commit_id > uncheckpointed_delete_commit.GetIndex()) {
-		uncheckpointed_delete_commit = commit_id;
+	// `uncheckpointed_delete_commit` is used to decide whether there're uncommitted deletes, no need to overwrite on
+	// rollback-ed transaction.
+	if (commit_id != NOT_DELETED_ID) {
+		if (!uncheckpointed_delete_commit.IsValid() || commit_id > uncheckpointed_delete_commit.GetIndex()) {
+			uncheckpointed_delete_commit = commit_id;
+		}
 	}
 	GetVectorInfo(vector_idx).CommitDelete(commit_id, info);
 }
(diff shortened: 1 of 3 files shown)


src/storage/table/row_version_manager.cpp
```diff
@@ -199,8 +199,12 @@ idx_t RowVersionManager::DeleteRows(idx_t vector_idx, transaction_t transaction_
 void RowVersionManager::CommitDelete(idx_t vector_idx, transaction_t commit_id, const DeleteInfo &info) {
 	lock_guard<mutex> lock(version_lock);
 	needs_compression_check = true;
-	if (!uncheckpointed_delete_commit.IsValid() || commit_id > uncheckpointed_delete_commit.GetIndex()) {
-		uncheckpointed_delete_commit = commit_id;
+	// `uncheckpointed_delete_commit` is used to decide whether there're uncommitted deletes, no need to overwrite on
+	// rollback-ed transaction.
+	if (commit_id != NOT_DELETED_ID) {
```
Mytherin: Could we change this to `commit_id >= TRANSACTION_ID_START`? Otherwise we will get the same issue when committing transaction-local commits.

dentiny: make sense, updated to util function `IsCommitted`

PR #25589, opened 2026-09-11
Fix foreign key constraint violation error message

### LINKED ISSUE #25588: Incorrect error message for foreign key constraint violation
### What happens?

Hi team, I found the error message for foreign key constraint violation is incorrect, see repro for details.

### To Reproduce

```sql
memory D CREATE TABLE fk_local_p(id INT PRIMARY KEY);
memory D CREATE TABLE fk_local_c(pid INT, FOREIGN KEY (pid) REFERENCES fk_local_p(id));
memory D BEGIN;
memory D INSERT INTO fk_local_p VALUES (10), (20);
memory D INSERT INTO fk_local_c VALUES (10), (20), (30);
Constraint Error:
Violates foreign key constraint because key "id: 10" does not exist in the referenced table
```

### OS:

macos

### DuckDB Version:

ed8592347a1bdf90c7074c810d8dc6e052171a3e

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25588

The root cause is for `GetFirstInvalidIndex`, 
- when `negate=true`(default), it's getting the first matching value (in our case, it's the first key which exists in the parent table)
- when `negate=false`(our fix), it's getting the first value which doesn't exist

### DIFF
--- a/src/storage/data_table.cpp
+++ b/src/storage/data_table.cpp
@@ -743,9 +743,9 @@ void DataTable::VerifyForeignKeyConstraint(optional_ptr<LocalTableStorage> stora
 		if (!global_conflicts && !local_conflicts) {
 			conflict = 0;
 		} else if (!global_conflicts && local_conflicts) {
-			conflict = local_conflict_manager.GetFirstInvalidIndex(count);
+			conflict = local_conflict_manager.GetFirstInvalidIndex(count, /*negate=*/true);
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25590, opened 2026-09-11
Propagate stats for all NULL row group for all type cast

dentiny:
Hi team, I found for type casting, row groups with NULL value are not handled well to propagate stats; for example,
```sql
memory D ATTACH '/tmp/cast_pruning.duckdb' AS rg (ROW_GROUP_SIZE 2048);
memory D USE rg;
rg D CREATE TABLE raw_events AS
     SELECT CASE WHEN i < 202752 THEN NULL ELSE '2025-01-15' END AS raw_date
     FROM range(204800) t(i);

-- should only access limited number of row groups, but all are accessed
rg D EXPLAIN ANALYZE
     SELECT * FROM raw_events
     WHERE TRY_CAST(raw_date AS DATE) >= DATE '2025-01-01';
╭─ Summary ───────────╮
│ Total Time: 0.0019s │
╰─────────────────────╯
╭─ Table Scan ─────────────────────────────────────╮
│ Table: rg.main.raw_events                        │
│ Projections: raw_date                            │
│ TRY_CAST(raw_date AS DATE) >= '2025-01-01'::DATE │
│ Row Groups Scanned: 100 / 100                    │
│ 2,048 rows                                10.0ms │
```
In this PR, I added handle for all NULL row group at the very beginning of `BoundCastExpression::PropagateStatistics`, which serves as the entrypoint for all possible type casts.


dentiny:
I found variant type stats don't seem to be propagated correctly, fixed here https://github.com/duckdb/duckdb/pull/25594

PR #25591, opened 2026-09-11
Propagate expression stats for constant row group

dentiny:
Hi team, I think for row groups of the same value (either NULL or real values), as long as the expression is consistent, we should be able to propagate the expression stats directly. Currently it's not implemented, for example,
```sql
memory D ATTACH '/tmp/constant_pruning.duckdb' AS rg (ROW_GROUP_SIZE 2048);
memory D USE rg;
rg D CREATE TABLE tenants AS SELECT i // 2048 AS tenant_id FROM range(204800) t(i);
rg D EXPLAIN ANALYZE SELECT * FROM tenants WHERE hash(tenant_id) = hash(42::BIGINT);
╭─ Summary ───────────╮
│ Total Time: 0.0006s │
╰─────────────────────╯
╭─ Table Scan ──────────────────────────╮
│ Table: rg.main.tenants                │
│ Projections: tenant_id                │
│ Filters:                              │
│ hash(tenant_id) = 7199933130570745587 │
│ Row Groups Scanned: 100 / 100         │
│ 2,048 rows                        0µs │
```
In this PR, when such constant row group detected, I calculate the stats for the constant expression value and propagate.

### DIFF
--- a/src/optimizer/statistics/expression/propagate_function.cpp
+++ b/src/optimizer/statistics/expression/propagate_function.cpp
@@ -24,9 +24,19 @@ bool TryEvaluateAtConstants(ClientContext &context, const BoundFunctionExpressio
 	return ExpressionExecutor::TryEvaluateScalar(context, *clone, result);
 }
 
-// Equal bounds need not imply identical inputs for certain types, so we skip this optimization for them.
-bool CanInferConstantFromNumericBounds(PhysicalType type) {
-	return type != PhysicalType::FLOAT && type != PhysicalType::DOUBLE && type != PhysicalType::INTERVAL;
+// Equal bounds need not imply identical inputs for certain types, so we skip this optimization for those values.
+// Floating-point bounds only lose information for the sign of zero.
+bool CanInferConstantFromNumericBounds(const Value &value) {
+	switch (value.type().InternalType()) {
+	case PhysicalType::FLOAT:
+		return value.GetValue<float>() != 0.0F;
+	case PhysicalType::DOUBLE:
+		return value.GetValue<double>() != 0.0;
+	case PhysicalType::INTERVAL:
+		return false;
+	default:
+		return true;
+	}
 }
 } // namespace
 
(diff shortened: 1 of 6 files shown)


src/optimizer/statistics/expression/propagate_function.cpp
```diff
@@ -6,19 +6,73 @@
...
+	values.reserve(func.GetChildren().size());
+	for (idx_t idx = 0; idx < func.GetChildren().size(); ++idx) {
+		auto &child = *func.GetChildren()[idx];
+		auto &stats = child_stats[idx];
+		Value value;
+		if (child.IsFoldable()) {
+			if (!ExpressionExecutor::TryEvaluateScalar(context, child, value)) {
+				return nullptr;
+			}
+		} else if (stats.CanHaveNull()) {
```
dentiny: Row group with all NULL values could be supported, but I met an existing issue: https://github.com/duckdb/duckdb/pull/25594, could followup

src/optimizer/statistics/expression/propagate_function.cpp
```diff
@@ -6,19 +6,73 @@
...
-	return ExpressionExecutor::TryEvaluateScalar(context, clone, result);
+	// Functions such as alias() inspect expression metadata during execution.
+	auto clone = func.Copy();
+	clone->Cast<BoundFunctionExpression>().GetChildrenMutable() = std::move(children);
+	return ExpressionExecutor::TryEvaluateScalar(context, *clone, result);
+}
+
+// Equal bounds need not imply identical inputs for certain types, so we skip this optimization for them.
+bool CanInferConstantFromNumericBounds(PhysicalType type) {
+	return type != PhysicalType::FLOAT && type != PhysicalType::DOUBLE && type != PhysicalType::INTERVAL;
```
Mytherin: Why isn't this the case for float / double? Because of `-0.0` and `0.0`? We should be able to figure that out from the stats itself then by checking if the constant is `0.0` - we can then still do this for other bounds.

src/optimizer/statistics/expression/propagate_function.cpp
```diff
@@ -6,19 +6,73 @@
...
+		auto &child = *func.GetChildren()[idx];
+		auto &stats = child_stats[idx];
+		Value value;
+		if (child.IsFoldable()) {
+			if (!ExpressionExecutor::TryEvaluateScalar(context, child, value)) {
+				return nullptr;
+			}
+		} else if (stats.CanHaveNull()) {
+			return nullptr;
+		} else if (stats.GetStatsType() == StatisticsType::NUMERIC_STATS && NumericStats::HasMinMax(stats)) {
```
Mytherin: Can we move this to a separate method that does the checking for all stats? e.g. `bool TryInferConstantBounds(BaseStatistics &stats, Value &constant_bounds)`

Mytherin (commented):
Thanks for the PR! Good idea - some comments

src/optimizer/statistics/expression/propagate_function.cpp
dentiny: Yes, because certain functions do differentiate between positive/negative 0, would like to play safe as the initial step

Mytherin: I'm fine with not doing it for `0`, but we can verify that the actual stats are `0` in that case and then do it if they aren't (i.e. still do it if `min = max = 2.0`, or any other number really)

dentiny: Yeah that sounds good, updated.

Mytherin:
Thanks!

PR #25592, opened 2026-09-11
Fix checkpoint corrupting -0.0 into 0.0 for FLOAT/DOUBLE columns

### LINKED ISSUE #25580: CHECKPOINT turns +0.0 into -0.0 for FLOAT/DOUBLE columns
### What happens?

Persisting a table (via CHECKPOINT, or simply by using a persistent database) with both -0.0 and 0.0 values in a DOUBLE/FLOAT column corrupts the positive zero into a negative zero, observable via `signbit()`.

Reproduces with a plain in-process CHECKPOINT, no restart or compression setting required. Likely related to how column statistics/min-max encode FLOAT/DOUBLE (see also #25417, which was
a similar non-injective -0.0/0.0 encoding bug in ORDER BY's sort key, now fixed separately).

### To Reproduce

```
CREATE TABLE zero_probe(id INTEGER, v DOUBLE, vf FLOAT);
INSERT INTO zero_probe VALUES (1, '-0.0', '-0.0'), (2, '0.0', '0.0');
CHECKPOINT;
SELECT id, signbit(v), signbit(vf) FROM zero_probe ORDER BY id;
```
```
Actual:
1  true   true
2  true   true    --should be false  false

Expected:
1  true   true
2  false  false
```

### OS:

macOS (arm64)

### DuckDB Version:

DuckDB v2.0.0-dev84142 (Development Version) 6acf711387

### DuckDB Client:

cli

### Hardware:

n/a

### Full Name:

asxvi

### Affiliation:

n/a

- [x] Yes, I have

- [x] Yes, I have

asxvi:
## Summary
`NumericStats::IsConstant()` treats `min == max` as proof a segment is constant, but `-0.0 == 0.0` for floats without being bit-identical. Checkpoint then collapses such segments to one value, losing the sign of zero. RLE has the same flaw: it merges adjacent values with `==`, folding a `-0.0` run into a `0.0` run.

## Fix
- `numeric_stats.cpp`: `IsConstant()` returns `false` for `FLOAT`/`DOUBLE` when min is zero.
- `rle.cpp`: added `RLEValueEqual<T>`, comparing bits instead of `==` for `float`/`double`.

## Tests
- `constant_negative_zero.test`: checks `signbit()` survives checkpoint/restart for both the constant-segment and forced-RLE paths.

Fixes #25580

### DIFF
--- a/src/storage/statistics/numeric_stats.cpp
+++ b/src/storage/statistics/numeric_stats.cpp
@@ -322,6 +322,13 @@ bool NumericStats::ConstantsCoverRange(const BaseStatistics &stats, array_ptr<co
 }
 
 bool NumericStats::IsConstant(const BaseStatistics &stats) {
+	auto physical_type = stats.GetType().InternalType();
+	if ((physical_type == PhysicalType::FLOAT || physical_type == PhysicalType::DOUBLE) &&
+	    NumericStats::Min(stats) == 0) {
+		// -0.0 and 0.0 compare equal but are not bit-identical, so min == max == 0 does not
+		// guarantee that all values in the segment have the same sign
+		return false;
+	}
 	return NumericStats::Max(stats) <= NumericStats::Min(stats);
 }
 
--- a/src/storage/compression/rle.cpp
+++ b/src/storage/compression/rle.cpp
@@ -14,6 +14,23 @@ namespace duckdb {
 
 using rle_count_t = uint16_t;
 
+template <class T>
+static inline bool RLEValueEqual(const T &a, const T &b) {
+	return a == b;
+}
+
+template <>
+inline bool RLEValueEqual(const float &a, const float &b) {
+	// avoid folding -0.0 and 0.0 (or differently-signed NaNs) into a single run: they compare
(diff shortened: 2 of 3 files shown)


dentiny (commented):
Thanks for the PR! I think the RLE one has been fixed in v2.0, the constant encoding one hasn't, maybe you could rebase the PR?

src/storage/statistics/numeric_stats.cpp
```diff
@@ -322,6 +322,13 @@ bool NumericStats::ConstantsCoverRange(const BaseStatistics &stats, array_ptr<co
 }
 
 bool NumericStats::IsConstant(const BaseStatistics &stats) {
+	auto physical_type = stats.GetType().InternalType();
+	if ((physical_type == PhysicalType::FLOAT || physical_type == PhysicalType::DOUBLE) &&
+	    NumericStats::Min(stats) == 0) {
```
dentiny, line 327: But if all values are -0.0 or +0.0 we could still leverage constant encoding?

PR #25594, opened 2026-09-11
Fix variant validity

### LINKED ISSUE #25593: Variant validity stats are incorrect
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE v AS
         SELECT i::VARIANT AS non_null, NULL::VARIANT AS all_null,
                CASE WHEN i = 0 THEN i::VARIANT END AS mixed
         FROM range(2) t(i);
memory D SELECT non_null, all_null, mixed FROM v ORDER BY rowid;
┌──────────┬──────────┬─────────┐
│ non_null │ all_null │  mixed  │
│ variant  │ variant  │ variant │
├──────────┼──────────┼─────────┤
│ 0        │ NULL     │ 0       │
│ 1        │ NULL     │ NULL    │
└──────────┴──────────┴─────────┘

-- should be true, false
memory D SELECT s.has_no_null, s.has_null FROM (SELECT stats(non_null) s FROM v LIMIT 1);
┌─────────────┬──────────┐
│ has_no_null │ has_null │
│   variant   │ variant  │
├─────────────┼──────────┤
│ false       │ false    │
└─────────────┴──────────┘

-- should be false, true
memory D SELECT s.has_no_null, s.has_null FROM (SELECT stats(all_null) s FROM v LIMIT 1);
┌─────────────┬──────────┐
│ has_no_null │ has_null │
│   variant   │ variant  │
├─────────────┼──────────┤
│ false       │ false    │
└─────────────┴──────────┘

-- should be true, true
memory D SELECT s.has_no_null, s.has_null FROM (SELECT stats(mixed) s FROM v LIMIT 1);
┌─────────────┬──────────┐
│ has_no_null │ has_null │
│   variant   │ variant  │
├─────────────┼──────────┤
│ false       │ false    │
└─────────────┴──────────┘
```

### OS:

macos

### DuckDB Version:

39ec07729f71f907887f86f479ebe26a85972509

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25593

### DIFF
--- a/src/storage/table/variant_column_data.cpp
+++ b/src/storage/table/variant_column_data.cpp
@@ -392,6 +392,8 @@ void VariantColumnData::Append(ColumnAppendState &state, const Vector &vector, i
 }
 
 void VariantColumnData::FinalizeAppend(ColumnDataFinalizeAppendState &finalize_state, ColumnAppendState &state) {
+	validity->FinalizeAppendLocked(finalize_state, state.child_appends[0]);
+
 	for (idx_t i = 0; i < sub_columns.size(); i++) {
 		ColumnDataFinalizeAppendState child_finalize_state(finalize_state, LogicalTypeId::VARIANT);
 		sub_columns[i]->FinalizeAppend(child_finalize_state, state.child_appends[i + 1]);
--- a/test/configs/stats_dependent_tests.json
+++ b/test/configs/stats_dependent_tests.json
@@ -20,6 +20,7 @@
         "test/sql/function/timetz/test_date_part.test",
         "test/sql/storage/types/variant/variant_extract_stats.test",
         "test/sql/storage/types/variant/variant_comparator_stats.test",
+        "test/sql/storage/types/variant/variant_validity_stats.test",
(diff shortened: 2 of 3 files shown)


Tishj:
Thanks!
Since this shows up in FinalizeAppend, I imagine this only affects non-shredded VARIANTs? Can we also check with forced shredding (set minimum shredding size 1) to verify that doesn't suffer from this issue? (Since that shouldn't be going through Append at all)

dentiny:
> Since this shows up in FinalizeAppend, I imagine this only affects non-shredded VARIANTs? Can we also check with forced shredding (set minimum shredding size 1) to verify that doesn't suffer from this issue? (Since that shouldn't be going through Append at all)

Thanks for the quick review! I covered two more test cases: threshreding threshold -1 & 1 to verify validity stats for both non-shredded and shredded VARIANTs.

Mytherin:
Thanks!

PR #25596, opened 2026-09-11
Fix integer truncation in ArgMinMaxValueAssign::Assign

### LINKED ISSUE #25577: Vuln184: Integer Truncation in `ArgMinMaxValueAssign::Assign` Leads to Heap Buffer Overflow with Large Strings
### What happens?

In `extension/core_functions/aggregate/distributive/arg_min_max.cpp`, `ArgMinMaxValueAssign<string_t>::Assign()` computes the arena allocation size by calling `UnsafeNumericCast<uint32_t>(NextPowerOfTwo(len))` where `len` is the byte length of the argument string. For any argument string whose length is in the range `(2^31, 2^32-1]`, `NextPowerOfTwo` returns `2^32`, and the subsequent `uint32_t` cast truncates it to `0`. In a release build, `ArenaAllocator::Allocate(0)` returns a zero-length region, and the immediately following `memcpy(ptr, new_value.GetData(), len)` writes gigabytes past that region, producing a heap buffer overflow. The bug is reachable by either the `arg_min()` or `arg_max()` aggregate when the first (argument) column is a VARCHAR/BLOB.

### To Reproduce

```sql
SET threads=1;
SELECT arg_min(s, 1)
FROM (
    FROM range(22)
) t;
```

```sh
echo "SET threads=1; SELECT arg_min(s, 1) FROM (SELECT string_agg(repeat('A', 100000000), '') AS s FROM range(22)) t;" | ./build/reldebug/duckdb
```

```
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

asxvi:
## Summary
`ArgMinMaxValueAssign<string_t>::Assign()` computed the arena allocation size via `UnsafeNumericCast<uint32_t>(NextPowerOfTwo(len))`. For a string length in `(2^31, 2^32-1]`, this returns `2^32`, which truncates to `0`, causing a zero-size allocation followed by a heap-overflowing `memcpy`.

## Fix
Widened `alloc_size` from `uint32_t` to `idx_t` in `arg_min_max.cpp`, dropping the truncating cast.

## Tests
- `arg_min_max_string_truncation_overflow.test_slow`: reproduces the truncation under a `memory_limit` too small to satisfy the correct allocation, so it deterministically fails cleanly (fixed) or crashes (unfixed).

Fixes #25577

### DIFF
--- a/extension/core_functions/aggregate/distributive/arg_min_max.cpp
+++ b/extension/core_functions/aggregate/distributive/arg_min_max.cpp
@@ -29,7 +29,7 @@ struct ArgMinMaxValueAssign {
 template <>
 struct ArgMinMaxValueAssign<string_t> {
 	//! The size of the arena allocation for a non-inlined string value
-	uint32_t alloc_size;
+	idx_t alloc_size;
 
 	void Assign(string_t &target, string_t new_value, AggregateInputData &aggregate_input_data) {
 		if (new_value.IsInlined()) {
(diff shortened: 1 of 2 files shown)


DinosL:
Hi @asxvi, I believe this is fixed in https://github.com/duckdb/duckdb/pull/25662

asxvi:
Closing this PR.

@DinosL, is there any way to close the related issue #25577 which is fixed in #25662

PR #25597, opened 2026-09-11
Fix wrong key reported in foreign key constraint error message

### LINKED ISSUE #25588: Incorrect error message for foreign key constraint violation

linsen458-spec:
Fixes #25588

When the referenced parent keys are inserted in the same transaction, the constraint error reports the key of the first row of the insert instead of the key that is actually missing. On current master:

```sql
CREATE TABLE fk_local_p(id INT PRIMARY KEY);
CREATE TABLE fk_local_c(pid INT, FOREIGN KEY (pid) REFERENCES fk_local_p(id));
BEGIN;
INSERT INTO fk_local_p VALUES (10), (20);
INSERT INTO fk_local_c VALUES (10), (20), (30);
-- Violates foreign key constraint because key "id: 10" does not exist ...  <- 30 is the missing one
```

The cause is in the append-FK error handling in `DataTable::VerifyForeignKeyConstraint` (`data_table.cpp`): the validity array of the `ConflictManager` records which rows were *found*, so the "local hits, global misses" branch needs the first row *without* a hit. The call there was `GetFirstInvalidIndex(count)`, which with the default `negate = false` returns the first row that *was* found — exactly backwards for this purpose. `FirstMissingMatch`, used by the no-local-verification path, already passes `negate = true`; this branch just missed it. The same one-word fix applies to the mirrored `global_conflicts && !local_conflicts` branch.

Verified locally by rebuilding the CLI and re-running the reproduction (now reports `"id: 30"`), plus the committed-parents case, the single-row case, the delete-side FK check, and a normal insert (all unchanged). `test/sql/constraints/foreignkey/fk_error_message_key.test` covers all of these.

### DIFF
--- a/src/storage/data_table.cpp
+++ b/src/storage/data_table.cpp
@@ -743,9 +743,12 @@ void DataTable::VerifyForeignKeyConstraint(optional_ptr<LocalTableStorage> stora
 		if (!global_conflicts && !local_conflicts) {
 			conflict = 0;
 		} else if (!global_conflicts && local_conflicts) {
-			conflict = local_conflict_manager.GetFirstInvalidIndex(count);
+			// The validity array records which rows were found, so we need
+			// to negate: the first row without a hit is the first missing
+			// one, which is the row the error message should report.
+			conflict = local_conflict_manager.GetFirstInvalidIndex(count, true);
 		} else if (global_conflicts && !local_conflicts) {
-			conflict = global_conflict_manager.GetFirstInvalidIndex(count);
+			conflict = global_conflict_manager.GetFirstInvalidIndex(count, true);
 		} else {
 			auto &global_validity = global_conflict_manager.GetFirstValidity();
 			auto &local_validity = local_conflict_manager.GetFirstValidity();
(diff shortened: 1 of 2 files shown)


dentiny:
Hi @linsen458-spec I raised the issue and already made a PR

linsen458-spec:
Apologies — I looked for open work on this before starting but missed #25589, which takes the same approach (the validity array records hits, so the lookup needs `negate = true`) and was opened a few hours earlier. Closing this in favor of it. Thanks @dentiny, and sorry for the noise!

PR #25600, opened 2026-09-11
Fix async readahead deadlock in parquet multi file scan

smvv:
### Root cause

An async read-ahead I/O task can fail[[1](https://github.com/duckdb/duckdb/actions/runs/34548620845/job/103153591832#step:7:154), [2](https://github.com/duckdb/duckdb/actions/runs/34548620845/job/103153591832#step:7:303), [3](https://github.com/duckdb/duckdb/actions/runs/34548620845/job/103153591835#step:7:120), [4](https://github.com/duckdb/duckdb/actions/runs/34548620845/job/103153591835#step:7:205), [5](https://github.com/duckdb/duckdb/actions/runs/34548620845/job/103153591816#step:7:17)] with OOM while file-open tasks remain queued on the same `TaskExecutor`.

Once the executor contains an error, `BaseExecutorTask::Execute()` skips queued tasks. A skipped `FileOpenTask` therefore never runs its callback, leaving the corresponding reader permanently in `MultiFileFileState::OPENING`.

`WaitForAsyncOpen()` continued polling that state without checking the executor error. With no runnable async work remaining, scan threads yielded forever.

### Sample trace

A local stress run reproduced the hang. The main thread and a worker were both spinning while claiming the next read-ahead job:

```text
MultiFileFunction<ParquetMultiFileInfo>::ClaimNextJobInternal
  ScanReadAhead::TryRunPendingTask
  TaskScheduler::YieldThread
```

Meanwhile, all async scheduler threads were idle:

```text
TaskScheduler::ExecuteForever
  semaphore_wait_trap
```

This confirms that the scan was waiting for a file-open callback that had already been cancelled.

### Fix

`ScanReadAhead::TryRunPendingTask()` now propagates recorded executor errors:

```cpp
ThrowIfError();
if (!executor->GetTask(task)) {
    return false;
}
task->Execute(TaskExecutionMode::PROCESS_ALL);
ThrowIfError();
```

Checking both before and after inline execution ensures that:

- an existing I/O error aborts the wait before cancelled file-open work is dequeued;
- an error raised by the task being executed is surfaced immediately.

### DIFF
--- a/src/include/duckdb/parallel/scan_read_ahead.hpp
+++ b/src/include/duckdb/parallel/scan_read_ahead.hpp
@@ -157,7 +157,7 @@ public:
 	void ScheduleFileOpen(std::function<void()> open_fn);
 	//! Whether another file-open may be scheduled without exceeding the open-ahead window
 	bool CanScheduleOpen() const;
-	//! Run one queued async task inline, returns false when none is queued
+	//! Run one queued async task inline, throwing any recorded async error; returns false when none is queued
 	bool TryRunPendingTask();
 
 private:
--- a/src/parallel/scan_read_ahead.cpp
+++ b/src/parallel/scan_read_ahead.cpp
@@ -300,11 +300,13 @@ bool ScanReadAhead::CanScheduleOpen() const {
 }
 
 bool ScanReadAhead::TryRunPendingTask() {
+	ThrowIfError();
 	shared_ptr<Task> task;
 	if (!executor->GetTask(task)) {
 		return false;
 	}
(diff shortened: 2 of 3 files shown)


carlopi:
I was looking at something similar at https://github.com/duckdb/duckdb/pull/25557, and I think both should be merged (there are not independent, but merging #25600 increase the need for #25557.

Then looking harder, I think that while #25557 is a solution for the specific code-path, the proper solution is one level up restructuring BaseExecutorTask contract.

Mytherin:
Thanks!

PR #25602, opened 2026-09-11
Bind planner and executor builtins through the catalog

kryonix:
#25472 moved the optimizer rewrites onto the catalog, so a builtin they introduce keeps the catalog and schema name its entry stamped on it. While writing that PR I kept finding the same factory-and-bind shape outside `src/optimizer`, and this picks those up.

Decorrelation is the clearest case. `FlattenDependentJoins` builds the `ROW_NUMBER()` it pushes onto the LHS with `RowNumberFun::GetFunction().Bind(...)`, and so do the full-outer-join pairing path, `WITH ORDINALITY`, set operations and the asof join. `plan_subquery` does the same for the `min`/`max`, `count`, `count_star` and `first` it introduces while flattening subqueries, the delim-join hash table does it for its correlated counts, and `bind_unnest_expression` builds `struct_extract` that way — the `StructExtractFun::GetFunction().Bind(...)` Maxxen pointed at in the review on #25260 that started all of this. Around twenty-five sites in total.

Window functions had no lookup helper yet, but they needed nothing new: `WindowFunctionCatalogEntry` stamps catalog and schema onto its overloads exactly like the scalar and aggregate entries, `WindowFunctionSet::GetFunctionByArguments` already exists, and `WindowFunction::Bind` mirrors `ScalarFunction::Bind`. So `GetBuiltinWindowFunction` is a direct transcription of the aggregate one. I also moved `builtin_function_lookup` from `duckdb/optimizer` to `duckdb/function`, because multi-file column mapping needs it and having `src/common` include an optimizer header inverts the layering.

Converting the call sites turned out not to be enough for the aggregates, which is the more interesting half of this. After the conversion I checked what the bound functions actually carried and they were still unqualified. The reason is `BoundAggregateFunction::ReplaceImplementation`: it copies `name`, `schema_name` and `catalog_name` off the replacement, and every caller passes a function built by a factory rather than one handed out by a catalog entry. `BindFirst` reaches its type-specialized implementation through it, so does the `sum` to `sum_no_overflow` swap in statistics propagation, and so do `quantile`, `mode`, `histogram` and `arg_min`/`arg_max` on decimals. The qualification the entry had just set was being dropped a moment later. That also means this was never only a planner problem: `SELECT sum(v) FROM t` loses it too, however the aggregate was bound.

The replacement is a specialized implementation of the function we were bound from, not a different function, so it should not be the authority on where that function came from. I made `ReplaceImplementation` take the replacement's qualification only when it actually has one. The name stays assignable, deliberately: `CountPropagateStats` renames `count` to `count_star` through this path when the child cannot be null, and the collated min/max path really does hand over a catalog-derived `arg_min`/`arg_max`, which carries its own qualification and should win.

A few things worth flagging. `first` and `last` looked like the risky conversion, since the call sites ask for `FirstFunctionGetter::GetFunction(type)` and the catalog set holds only an `ANY` and a `DECIMAL` entry. They resolve to the same thing: `BindFirst` specializes the implementation for the argument type and applies the same direct-rewrite, single-value-identity and statistics callbacks the factory sets directly, plus `SetDistinctDependent`, which the factory path never set. In `AddCorrelatedFirstAggregates` I replaced a hand-built `BoundAggregateExpression` carrying null bind data with a real bind, so code now runs there that previously did not — the end state is the same, but it is a deliberate behavioural change rather than a mechanical substitution.

Two more sites turned up after I thought the sweep was done, and I found them by re-grepping rather than by reading the code, which is the honest provenance: `PhysicalPlanGenerator::CreatePlan` still built its `first()` from `FirstFunctionGetter` in two places, for the `DISTINCT ON` rewrite in `plan_distinct` and for the key representative in `plan_recursive_cte`. Both feed `FunctionBinder::BindAggregateFunction` like the asof join does, so they convert the same way; `plan_distinct` keeps its null bind info and `AggregateType::NON_DISTINCT`. Both files had been including `distributive_function_utils` only to reach the getter, so that becomes `distributive_functions` for `FirstFun::Name`, which the unity build would otherwise have hidden.

I left `src/planner/filter` alone on purpose, and added a comment saying why. The bloom, prefix-range, dynamic and optional filter functions are registered, but only so that plans containing them can be serialized and deserialized: `TableFilterFunctions::Bind` throws "Table filter functions are for internal use only!", and the call sites build the expression directly with hand-made bind data. Routing them through the catalog would turn every dynamic filter into a binder error. This is the same situation as the compressed materialization helpers, which #25472 left for the same reason. Cast functions remain unqualified too, since they come from the cast function set rather than a catalog entry, and that is a separate piece of work.

The test from #25472 grew a branch for window expressions, which hold either an aggregate or a window function depending on whether you wrote `sum(x) OVER ()` or `row_number()`, so the planner-introduced ones are visible to it at all.


Maxxen:
LGTM, thanks!

PR #25605, opened 2026-09-11
Fix spilled temp block file report

### LINKED ISSUE #25604: Spilled block file doesn't account in `duckdb_temporary_files` function
### What happens?

Hi team, when temporary directory configured, oversized table will be spilled to disk block files. But I found the memory accounting via `duckdb_temporary_files` doesn't seem to report well.

### To Reproduce

```sql
memory D SET threads = 1;
memory D SET temp_directory = '/tmp/duckdb-temp-files-repro';
memory D SET memory_limit = '50MB';
memory D ATTACH '/tmp/duckdb-temp-files-repro.db' AS audit;
memory D CREATE TABLE audit.t (id INTEGER, val VARCHAR);
memory D INSERT INTO audit.t
         SELECT range, hex(repeat(md5(range::VARCHAR)::BLOB, 9000))
         FROM range(20);
memory D INSERT INTO audit.t
         SELECT range + 20, hex(repeat(md5(range::VARCHAR)::BLOB, 9000))
         FROM range(20);
memory D INSERT INTO audit.t
         SELECT range + 40, hex(repeat(md5(range::VARCHAR)::BLOB, 9000))
         FROM range(20);
memory D INSERT INTO audit.t
         SELECT range + 60, hex(repeat(md5(range::VARCHAR)::BLOB, 9000))
         FROM range(20);
memory D INSERT INTO audit.t
         SELECT range + 80, hex(repeat(md5(range::VARCHAR)::BLOB, 9000))
         FROM range(20);

-- we do have multiple spilled block files under temp directory 
memory D SELECT count(*) > 0 AS block_files_exist
         FROM glob('/tmp/duckdb-temp-files-repro/*.block');
┌───────────────────┐
│ block_files_exist │
│      boolean      │
├───────────────────┤
│ true              │
└───────────────────┘

-- but temporary files function doesn't collect them somehow
memory D SELECT count(*) AS reported_block_files
         FROM duckdb_temporary_files()
         WHERE path LIKE '%.block';
┌──────────────────────┐
│ reported_block_files │
│        int64         │
├──────────────────────┤
│                    0 │
└──────────────────────┘
```

### OS:

macos

### DuckDB Version:

39ec07729f71f907887f86f479ebe26a85972509

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25604

The root cause is, `LocalFileSystem::ListFiles` pass filename, instead of filepath, to the callback, so the open file handle [here](https://github.com/duckdb/duckdb/blob/88902aaa3972216c928c8e79a89217a413be414c/src/storage/standard_buffer_manager.cpp#L736-L737) cannot really locate the block file.
The process doesn't crash because non-existent files are skipped silently via `FileFlags::FILE_FLAGS_NULL_IF_NOT_EXISTS`.

### DIFF
--- a/src/storage/standard_buffer_manager.cpp
+++ b/src/storage/standard_buffer_manager.cpp
@@ -734,13 +734,14 @@ vector<TemporaryFileInformation> StandardBufferManager::GetTemporaryFiles() {
 		}
 
 		// Another process or thread can delete the file before we can get its file size.
-		auto handle = fs.OpenFile(name, FileFlags::FILE_FLAGS_READ | FileFlags::FILE_FLAGS_NULL_IF_NOT_EXISTS);
+		auto path = fs.JoinPath(temporary_directory.path, name);
+		auto handle = fs.OpenFile(path, FileFlags::FILE_FLAGS_READ | FileFlags::FILE_FLAGS_NULL_IF_NOT_EXISTS);
 		if (!handle) {
 			return;
 		}
 
 		TemporaryFileInformation info;
-		info.path = name;
+		info.path = std::move(path);
 		info.size = NumericCast<idx_t>(fs.GetFileSize(*handle));
 		handle.reset();
 		result.push_back(info);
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25607, opened 2026-09-11
Propagate expression stats for all NULL constant row group

dentiny:
Followup PR for https://github.com/duckdb/duckdb/pull/25591 to propagate expression stats for row group with all NULL values -- previously unimplemented due to a pre-dating bug for NULL stats propagation.


Mytherin:
Thanks!

PR #25608, opened 2026-09-11
Implement consistent `MoveFile` behavior for all platforms

dentiny:
Followup PR for https://github.com/duckdb/duckdb/pull/25490

In the current `MoveFile` implementation (maybe other filesystem APIs as well), it behaves differently across multiple platforms. For example, windows by default doesn't allow destination to be overwritten by default with `MoveFileExW`, while unix platform doesn't block.

In this PR, I'd like to establish a consistent behavior for all these platforms, so we no longer fixing issues, say, concurrent open + delete / move on windows platform.

The core of the PR is to leverage `REPLACE_IF_EXISTS | POSIX_SEMANTICS` flag to switch to unix behavior.
Also fix a bug for extended UNC paths parsing.


src/common/local_file_system.cpp
```diff
@@ -150,7 +152,9 @@ static std::wstring NormalizePathAndConvertToUnicode(FileSystem &fs, const strin
 	}
 
 	if (abs_path.find(L"\\\\") == 0) {
-		return WINDOWS_UNC_LONG_PATH_PREFIX + abs_path;
+		// Extended UNC paths use "\\?\UNC\server\share", so remove the original leading "\\".
+		// See https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#namespaces
+		return WINDOWS_UNC_LONG_PATH_PREFIX + abs_path.substr(2);
```
staticlibs, line 157: Good catch! We may want to remove the trailing slashes from `WINDOWS_UNC_LONG_PATH_PREFIX` to use it as a "pure prefix". But then it needs to be changed in a few more places where it is used, so the fix in current form is fine!

src/common/local_file_system.cpp
```diff
@@ -1602,13 +1606,43 @@ void LocalFileSystem::FileSync(FileHandle &handle) {
 	}
 }
 
+static bool TryMoveFileWithPosixSemantics(const std::wstring &source, const std::wstring &target) {
+	constexpr DWORD delete_access = 0x00010000L;                                     // DELETE
+	constexpr auto file_rename_info_ex = static_cast<FILE_INFO_BY_HANDLE_CLASS>(22); // FileRenameInfoEx
+	const auto file_name_length = target.size() * sizeof(WCHAR);
+	const auto rename_info_size = offsetof(FILE_RENAME_INFO, FileName) + file_name_length + sizeof(WCHAR);
+	const auto rename_info_size_dw = NumericCast<DWORD>(rename_info_size);
+	auto rename_info_buffer = unique_ptr<data_t[]>(new data_t[rename_info_size]);
```
staticlibs: I would suggest memsetting it to '\0' just in case, current variant is correct though because all fields are filled.

src/common/local_file_system.cpp
```diff
@@ -1602,13 +1606,43 @@ void LocalFileSystem::FileSync(FileHandle &handle) {
...
+	rename_info->RootDirectory = nullptr;
+	rename_info->FileNameLength = NumericCast<DWORD>(file_name_length);
+	std::copy(target.c_str(), target.c_str() + target.size() + 1, rename_info->FileName);
+
+	// FileRenameInfoEx renames the file identified by a handle opened with DELETE access.
+	// See https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-file_rename_info
+	auto raw_source_handle =
+	    CreateFileW(source.c_str(), delete_access, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nullptr,
+	                OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OPEN_REPARSE_POINT, nullptr);
+	if (raw_source_handle == INVALID_HANDLE_VALUE) {
```
staticlibs, line 1634: Suggest to change this to open the handle in `LocalFileSystem::MoveFile` and throw its error immediately if the handle cannot be opened (without going to `MoveFileExW`).

src/common/local_file_system.cpp
```diff
@@ -1602,13 +1606,43 @@ void LocalFileSystem::FileSync(FileHandle &handle) {
...
+
+	return SetFileInformationByHandle(source_handle.get(), file_rename_info_ex, rename_info, rename_info_size_dw);
+}
+
 void LocalFileSystem::MoveFile(const string &source, const string &target, optional_ptr<FileOpener> opener) {
 	auto source_unicode = NormalizePathAndConvertToUnicode(*this, source, opener);
 	auto target_unicode = NormalizePathAndConvertToUnicode(*this, target, opener);
-	DWORD flags = MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH;
 
+	if (TryMoveFileWithPosixSemantics(source_unicode, target_unicode)) {
```
staticlibs: Suggest to change the error handling for this call - perhaps it will be better to follow the [error handling done in Git for Windows](https://github.com/git-for-windows/git/blob/0cba4f34323c254de6a029e992faabbfd8944239/compat/mingw.c#L3065-L3090), so to only fall back to `MoveFileEx` when one of the linked 3 error is returned and throw promptly otherwise.

I would think we do not need the [handling for ERROR_ACCESS_DENIED](https://github.com/git-for-windows/git/blob/0cba4f34323c254de6a029e992faabbfd8944239/compat/mingw.c#L3098-L3116), as we likely to move the file we've written ourselves. Or maybe left this as a possible future improvement.

src/common/local_file_system.cpp
```diff
@@ -1602,13 +1606,43 @@ void LocalFileSystem::FileSync(FileHandle &handle) {
...
+		return;
+	}
+
+	DWORD flags = MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH;
```
staticlibs, line 1653: I suggest to add here `MOVEFILE_COPY_ALLOWED` just in case.

staticlibs (approved):
The change looks good to me! The suggestion in comments is to adjust the fall-back logic slightly, basically to not fall-back in cases where we don't expect the repeat to succeed.

carlopi:
I think "establish a consistent behavior for all these platforms" is a bit overpromising, since I would assume OS support is basically complete, but support for say a file on a NAS box or equivalent virtualized memory would go back to legacy path. That's minor usage, but the path will be still exercized in a bunch of relevant cases I think (and those will get very limited coverage).

src/common/local_file_system.cpp
dentiny: sure, updated

src/common/local_file_system.cpp
dentiny: sounds good! TIL -- updated

src/common/local_file_system.cpp
dentiny: I'm not sure this is the correct behavior, here's my thoughts
- `MOVEFILE_COPY_ALLOWED` flag is used to copy cross volume (see https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexa); current behavior matches unix one, `EXDEV` error will be returned when cross mounted filesystem be requested (see https://man7.org/linux/man-pages/man2/rename.2.html)
- Current implementation in v2.0 seems to [have the same behavior](https://github.com/duckdb/duckdb/blob/714b5904e607eecf40ce5abcc21e308c34ad4142/src/common/local_file_system.cpp#L1608)
- It doesn't provide atomicity (it's basically copy + delete operation involve data replication), which is required and often assumed by WAL write
- If we have certain feature requested in the future, I'm thinking to have another API to provide cross-volume move without atomicity guarantee

staticlibs: Got it, makes perfect sense to me!

staticlibs:
Updated version looks good to me!

Mytherin:
Thanks!

PR #25609, opened 2026-09-11
Add sql tokenizer to C API

evertlammerts:
This exposes the sql tokenizer through the new C API. There's not much interesting happening I think:
- We add a `tokenize_sql(conn, sql, *out_iterator, *err)` function that takes a `duckdb_v2_str sql` string and creates an opaque `token_iterator`. Users can call `token_iterator_next(*out_type, *out_start, *out_length, *err)` to get the token type (from the `TOKEN_TYPE` enum), the start position in the input string, and the length of the lexeme in the input string.
- The C++ API wraps this by adding the `Connection::Tokenize(sql)` method, which returns a vector of `Token`s (a struct that includes the type, start index and length)

The reason that the tokenization happens in the context of a connection is to respect any grammar extensions that might be loaded on that connection. The tokenizer is eager though, so the iterator does not need the connection after its been created.

### DIFF
--- a/src/include/duckdb_v2.h
+++ b/src/include/duckdb_v2.h
@@ -3252,6 +3252,117 @@ DUCKDB_C_API DUCKDB_V2_ERROR duckdb_v2_schema_destroy(duckdb_v2_schema_handle *s
 /* --- Struct definitions for schema --- */
 
 /* ============================================================================
+ * MODULE: tokenizer
+ * ============================================================================ */
+
+/* --- Enums for tokenizer --- */
+
+//! The lexical class of a token.
+typedef enum DUCKDB_V2_TOKEN_TYPE {
+	//! Not an actual token class; out_type is set to this when `duckdb_v2_token_iterator_next()` fails.
+	DUCKDB_V2_TOKEN_TYPE_INVALID = 0,
+
+	//! A keyword of the connection's grammar.
+	DUCKDB_V2_TOKEN_TYPE_KEYWORD = 1,
+
+	//! A bare or double-quoted identifier, quotes included.
+	DUCKDB_V2_TOKEN_TYPE_IDENTIFIER = 2,
+
+	//! A quoted or dollar-quoted string, delimiters included.
+	DUCKDB_V2_TOKEN_TYPE_STRING_LITERAL = 3,
+
+	//! A numeric literal.
+	DUCKDB_V2_TOKEN_TYPE_NUMBER_LITERAL = 4,
+
+	//! An operator or punctuation run other than the statement terminator.
+	DUCKDB_V2_TOKEN_TYPE_OPERATOR = 5,
+
+	//! A line or block comment, delimiters included.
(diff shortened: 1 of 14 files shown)


Maxxen:
Does this need to be a C-API function? I think we added SQL-function to support this specifically in the V1-API.

```
select * from sql_tokenize('select 1 + 2');
┌───────┬────────────────┬─────────┐
│ start │   token_type   │  word   │                                                                                                                                                                                                                                                                                                                           
│ int32 │    varchar     │ varchar │                                                                                                                                                                                                                                                                                                                           
├───────┼────────────────┼─────────┤                                                                                                                                                                                                                                                                                                                           
│     0 │ KEYWORD        │ select  │                                                                                                                                                                                                                                                                                                                           
│     7 │ NUMBER_LITERAL │ 1       │                                                                                                                                                                                                                                                                                                                           
│     9 │ OPERATOR       │ +       │                                                                                                                                                                                                                                                                                                                           
│    11 │ NUMBER_LITERAL │ 2       │                                                                                                                                                                                                                                                                                                                           
└───────┴────────────────┴─────────┘ 
```

Maxxen:
But I suppose its ok since the surface is not that large anyway

Maxxen:
We're not going to add any new token-types are we?

Maxxen (approved):

evertlammerts:
> Does this need to be a C-API function? I think we added SQL-function to support this specifically in the V1-API.

@Maxxen It would be nice not to have to run a query for this, which would invalidate other stuff a user might be doing. Especially in the way notebooks use this it could be an issue.

src/include/duckdb_v2.h
```diff
@@ -3251,6 +3251,117 @@ DUCKDB_C_API DUCKDB_V2_ERROR duckdb_v2_schema_destroy(duckdb_v2_schema_handle *s
...
+	DUCKDB_V2_TOKEN_TYPE_COMMENT = 6,
+
+	//! A statement-terminating semicolon.
+	DUCKDB_V2_TOKEN_TYPE_TERMINATOR = 7,
+
+	/*!
+	 * Not a token; reported by token_iterator_next once the input is exhausted, with start equal to the input length
+	 * and length 0.
+	 */
+	DUCKDB_V2_TOKEN_TYPE_END_OF_INPUT = 8,
```
Tishj, line 3290: This is lacking the autocomplete_end_of_input, no?
I had plans to get rid of `end_of_input` tokens in general, but that also touched on the old `ParserExtension` route which wasn't desired, so I kind of shelved it: https://github.com/duckdblabs/duckdb-internal/issues/10486

But that doesn't take away that we currently need the distinction between an end-of-input and an autocomplete-end-of-input token

evertlammerts: Good catch, I added `token_iterator_ends_unterminated()` for that. If this returns false, that means the last token before END_OF_INPUT ended unterminated.

PR #25614, opened 2026-09-11
Let's not do UB within StringUtil::TryParseFormattedBytes

carlopi:
Follow up of a conversation with @okruitho


Mytherin:
Thanks!

PR #25617, opened 2026-09-11
Add parser microbenchmarks

Dtenwolde:
Given the recent spree of parser optimizations we are doing, I figured it might be a good time to add parser specific microbenchmarks. It only calls `Parser::ParseQuery()`, so no execution is happening. There's a couple of queries included: 
- all of TPC-H
- All of TPC-DS
- AoC
- Flummi (significantly regressed at some point, large query)
- Malformed `SELECT ((((((((((` to test the packrat caching
- Nested expression
- Multi-statement

Timings are reported in seconds per batch, not per query. For example, one TPC-H run parses all 22 queries 50 times, so its reported time covers 1,100 parses. Batch sizes are documented in the README and must remain unchanged when comparing revisions. Query benchmarks reuse the compiled grammar. The grammar creation time is tested separately. 

They are included in the regression CI.

### DIFF
--- a/.github/regression/parser.csv
+++ b/.github/regression/parser.csv
@@ -1,0 +1,9 @@
+ParserSimpleSelect
+ParserKeywordIdentifiers
+ParserWideSelect
+ParserNestedExpressions
+ParserCTEs
+ParserStatements
+ParserTPCH
+ParserTPCDS
+ParserFlummi
--- a/.github/regression/parser.csv
+++ b/.github/regression/parser.csv
@@ -1,9 +1,9 @@
-ParserSimpleSelect
 ParserKeywordIdentifiers
 ParserWideSelect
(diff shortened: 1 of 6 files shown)


Tishj (approved):
Thanks!
Might be good to mention the timing output quirks in the description though
It's in the README but it can't hurt to put it in the description, for easier git-blame based debugging/discovery

Dtenwolde:
>  Might be good to mention the timing output quirks in the description though

I updated the readme. 

I suppose the CI is intended to fail the first time on the regression test? I suppose because the baseline doesn't have the timings? Lets see

Tishj:
Yea that's intended (or at least expected behavior) 👍

Mytherin:
Thanks!

PR #25630, opened 2026-09-11
Improve exception message for cast failure

dentiny:
Hi team, I found currently a bunch of cast failure message reports DuckDB internal physical type, instead of user-specified logical type, which could be confusing to users. For example,
```sql
memory D SELECT 'x'::UUID;
Conversion Error:
Could not convert string 'x' to INT128

LINE 1: SELECT 'x'::UUID;
                  ^^^^^^
```


Mytherin:
Thanks!

PR #25634, opened 2026-09-11
Fix dangling pointer in RemoveUnusedColumns filter collection (#25430)

### LINKED ISSUE #25430: Expression::HasParameter crashes planning a correlated subquery with a STRUCT-access join predicate

huoguocheng:
Summary
Fix a dangling pointer bug in RemoveUnusedColumns::RemoveColumnsFromLogicalGet.
filter_expressions is a vector<unique_ptr>. When push_back(std::move(filter_expr)) triggers vector reallocation, all pointers/references to existing vector elements are invalidated.

The original implementation visited &filter_expressions.back() right after each push_back. If reallocation happened during insertion, previously stored pointers became dangling and caused undefined behavior.

Changes
Collect all filter expressions into filter_expressions first
After the loop finishes collecting all filters, iterate the vector and invoke VisitExpression on every entry
Add regression test with join + struct field predicate to reproduce the issue
Testing
Added struct_projection_pushdown_optimizer_bug.test case.

Fixes https://github.com/duckdb/duckdb/issues/25430

### DIFF
--- a/src/optimizer/remove_unused_columns.cpp
+++ b/src/optimizer/remove_unused_columns.cpp
@@ -1135,8 +1135,10 @@ void RemoveUnusedColumns::RemoveColumnsFromLogicalGet(LogicalGet &get, unique_pt
 			filter_expr = std::move(column_ref);
 		}
 		filter_expressions.push_back(std::move(filter_expr));
+	}
+	for (auto &filter_expression : filter_expressions) {
 		//! Now visit the filter to add to the 'column_references'
-		VisitExpression(&filter_expressions.back());
+		VisitExpression(&filter_expression);
 	}
 	for (const auto &filter : get.table_filters.GetMultiColumnFilters()) {
 		const auto &expression_filter = ExpressionFilter::GetExpressionFilter(*filter, "RemoveUnusedColumns::VisitGet");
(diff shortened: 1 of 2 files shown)


Mytherin:
Thanks!

PR #25637, opened 2026-09-12
Fix fetch row implementation for chimp compression

### LINKED ISSUE #25636: Chimp compression suffer INTERNAL error when fetch row mode
### What happens?

See repro for details.

### To Reproduce

```sql
haojiang@HaoJiangs-MacBook-Pro$ ./build/reldebug/duckdb test/sql/storage/compression/chimp/chimp.db
DuckDB v2.0.0-dev84402 (Development Version, 88902aaa39)
chimp D SET debug_force_fetch_row=true;
chimp D SELECT temperature::VARCHAR FROM temperatures_double LIMIT 3;
INTERNAL Error:
Expected vector of type UINT64, but found vector of type DOUBLE
```

### OS:

macos

### DuckDB Version:

88902aaa39

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25636

I think the issue is, for the failure case mentioned in the issue, chimp decompresses into a uint64_t value, and that uint64's 8 bytes are the double's 8 bytes.
In one word, the bytes are correct and identical at the bit level; only the type names disagree.

I confirmed with the [`patas`](https://github.com/duckdb/duckdb/blob/10de9573794001c649621013bdd93553b54e00c9/src/include/duckdb/storage/compression/patas/patas_fetch.hpp#L25) and [`alp`](https://github.com/duckdb/duckdb/blob/10de9573794001c649621013bdd93553b54e00c9/src/include/duckdb/storage/compression/alp/alp_fetch.hpp#L29) have the same implementation as well.

### DIFF
--- a/src/include/duckdb/storage/compression/chimp/chimp_fetch.hpp
+++ b/src/include/duckdb/storage/compression/chimp/chimp_fetch.hpp
@@ -22,7 +22,7 @@ void ChimpFetchRow(ColumnSegment &segment, ColumnFetchState &state, row_t row_id
 
 	ChimpScanState<T> scan_state(segment);
 	scan_state.Skip(segment, UnsafeNumericCast<idx_t>(row_id));
-	auto result_data = FlatVector::GetDataMutable<INTERNAL_TYPE>(result);
+	auto result_data = FlatVector::GetDataMutableUnsafe<INTERNAL_TYPE>(result);
 
 	if (scan_state.GroupFinished() && scan_state.total_value_count < scan_state.segment_count) {
 		scan_state.LoadGroup(scan_state.group_state.values);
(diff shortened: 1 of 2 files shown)


Tishj (approved):
Thanks, these use `using EXACT_TYPE = typename FloatingToExact<T>::TYPE` which indeed returns `uint64_t` for `double`.
Because of this shared behavior, I wonder if `ALP` and `ALPRD` suffer from the same problem? Those use this same mechanism

dentiny:
> I wonder if ALP and ALPRD suffer from the same problem? Those use this same mechanism

These two are already using `GetMutableUnsafe` -- see my PR description :)

PR #25639, opened 2026-09-12
Bump spatial to take the && join operand check

estebanzimanyi:
spatial eb1e57c's join optimizer plans any two-argument boolean function
named && as a spatial join, deciding on the function name alone. An
extension that registers && over its own types sees those joins planned as
spatial joins over values that are not GEOMETRY: MobilityDuck's index-join
tests fail against the repository build of spatial eb1e57c on v1.5.5.

duckdb/duckdb-spatial PR 853 (c27bc2b00, on v1.5-variegata) restricts the
rewrite to GEOMETRY operands. eb1e57c is 6 commits behind it; c0c9fc9e,
the head of spatial's v1.5-variegata, is 12 ahead. 0001-reset-storage.patch
still applies there through apply_extension_patches.py's patch -p1
--forward. DuckDB v1.5.5 with spatial c0c9fc9e and APPLY_PATCHES is the
pairing MobilityDuck builds on Linux, macOS and MinGW.

The bump also brings the other fixes on that branch: ST_Contains on a
polygon vertex, RTREE WAL replay of non-flat row ids, NaN and zero-length
cases in linear referencing, a truncated WKB raising an input error, and
ST_Subdivide.

### DIFF
--- a/.github/config/extensions/spatial.cmake
+++ b/.github/config/extensions/spatial.cmake
@@ -3,7 +3,7 @@ if (${BUILD_COMPLETE_EXTENSION_SET})
 duckdb_extension_load(spatial
     DONT_LINK LOAD_TESTS
     GIT_URL https://github.com/duckdb/duckdb-spatial
-    GIT_TAG eb1e57c9d92c0f3f76eb03eaa52c315090f328cc
+    GIT_TAG c0c9fc9e1f8c8fa8bacac54ac692cc488b9661ee
     INCLUDE_DIR src/spatial
     TEST_DIR test/sql
     APPLY_PATCHES


estebanzimanyi:
A reproducer on stock DuckDB v1.5.5 (`d8cdaa33fd`) with the repository build of spatial (`eb1e57c`), no other extension loaded:

```sql
LOAD spatial;
CREATE TABLE a AS SELECT * FROM (VALUES (1, [10, 20, 30]::ubigint[]), (2, [40, 50]::ubigint[])) t(id, cells);
CREATE TABLE b AS SELECT * FROM (VALUES (1, [20, 25]::ubigint[]), (2, [99]::ubigint[])) t(id, cells);
SELECT a.id, b.id FROM a JOIN b ON a.cells && b.cells;
```

```
INTERNAL Error: Failed to find function ST_IsEmpty(UBIGINT[])
```

Without `LOAD spatial` the same join answers `1 | 1`. It is the case spatial's own `test/sql/join/spatial_join_non_geometry_list.test` covers at `c0c9fc9e`. An extension that registers `&&` over its own types meets the same error: MobilityDuck's temporal index join answers `INTERNAL Error: Failed to find function ST_IsEmpty(stbox)` once spatial is loaded first.

Maxxen:
This wont work until we apply the patches in https://github.com/duckdb/duckdb/tree/v2.0-cyanoptera/.github/patches/extensions/spatial to spatial too - but don't worry we will get it done before v1.5.6.

PR #25641, opened 2026-09-12
CopyToFile/1: Count a partitioned COPY local into its sink state under the global lock

carlopi:
Sink picked the sink state under the global lock but counted itself in only after releasing it. A flush initiated in between finalized the state without the joining local, which then sank into a finalized state and finalized it again on its own combine. Count in inside the same locked section; the state lock nests inside the global lock as Finalize already does.

The test overlaps joins and flush triggers with a low flush threshold and many threads, and checks count, distinct count and sum of the output.

Test is not clear cut red / green, it's more greenish / green, but I think OK to include.


Mytherin:
Thanks!

PR #25642, opened 2026-09-12
CopyToFile/2: Remove the partitioned COPY straggler guard that could never fire

carlopi:
`ShouldStopFlushing` compared locals against a combined counter that was never incremented, and a thread flushing from Sink has not combined yet, so the guard was always false. No behaviour change.

Not a correctness change, as is it's inert, so can also just stay.


Mytherin:
Thanks!

PR #25643, opened 2026-09-12
CopyToFile/3: Check that a partitioned COPY state is finalized once and never blocks

carlopi:
`FinalizeState` is reached from three places under three conditions; assert the invariant they share, that a state is finalized once. Its sort strategy result was discarded: the flush runs the strategy's tasks itself, so a `BLOCKED` result could not be resumed. Make that an internal error.

Idea is basically: `sort_strategy->Finalize`'s result is thrown away, we could as well assert it shouldn't be `BLOCKED`.

This is not a correctness fix, just clean up / future proofing.


Mytherin:
Thanks!

PR #25644, opened 2026-09-12
TaskExecution API improvements, fixing up some users

carlopi:
This started while looking into https://github.com/duckdb/duckdb/pull/25557 and then the local fix at https://github.com/duckdb/duckdb/pull/25600.

Question was: why local fix? what's the actual shape to solidify usage of `TaskExecutor` ?
Down the rabbit hole, I centralized into capabilities of the `TaskExecutor` class patterns like:
```c++
	try {
		ScheduleTasksInternal();
		executor->WorkOnTasks();
	} catch (...) {
		try {
			executor->WorkOnTasks();
		} catch (...) {
		}
		throw;
	}
	RethrowTaskError();
```
to
```c++
	{
		// join before leaving this scope, whether the scheduling succeeds or throws
		TaskExecutor::JoinGuard join(*executor);
		ScheduleTasksInternal();
	}
```

Adds 3 capabilities to TaskExecutor, centralizing them

### Add Cancel(), and ensure one (and only one) of ExecuteTask() and Cancel() are called

This gives a place where to execute tear down, depending on whether on actual task execution OR on cancellation.

Cancel() is then used to solve both the issue reported and solved at https://github.com/duckdb/duckdb/pull/25600 and a potential issue where a throw at a bad moment would contribute to hang a `COPY ... TO`. Now invariant is simple to reason about, and should solve all those cases properly.

### Add `virtual TaskExecutionResult ExecuteTaskStep()`

Iceberg code hand rolled a stepped ExecuteTask, that required returning a TaskExecutionResult. That has also been moved into core, and the `iceberg` code can be then simplified

### TaskExecutor::JoinGuard

This ensures `WorkOnTasks()` is performed before leaving a given scope, avoiding repeated patterns.

Note this does add code in:
```
src/include/duckdb/parallel/task_executor.hpp
src/parallel/task_executor.cpp
--- plus for solving bug at
src/execution/operator/persistent/physical_copy_to_file.cpp
```
while removing code AND complexity elsewhere.

### DIFF
--- a/.github/patches/extensions/iceberg/0004-manifest-read-task-execute-step.patch
+++ b/.github/patches/extensions/iceberg/0004-manifest-read-task-execute-step.patch
@@ -1,0 +1,53 @@
+diff --git a/src/planning/scan_plan/iceberg_client_scan_plan_provider.cpp b/src/planning/scan_plan/iceberg_client_scan_plan_provider.cpp
+index f07d2acc65..8e6339691e 100644
+--- a/src/planning/scan_plan/iceberg_client_scan_plan_provider.cpp
++++ b/src/planning/scan_plan/iceberg_client_scan_plan_provider.cpp
+@@ -40,11 +40,9 @@ public:
+ 	    : BaseExecutorTask(state.executor), state(state), reader(*state.scan) {
+ 	}
+ 
+-	void ExecuteTask() override {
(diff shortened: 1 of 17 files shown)


Mytherin:
Thanks!

PR #25645, opened 2026-09-12
Format the main CMakeLists.txt through the standard channel

JaySon-Huang:
`make format-check-silent` prints this on every run:

```
      File "scripts/format.py", line 448, in process_file
    AttributeError: 'str' object has no attribute 'filename'
```

The check still passes, so it's just noise - but it hides real failures, and the main CMakeLists.txt is never actually formatted either.

Both go back to #23211, which meant to route the main CMakeLists.txt through the normal formatting path. It never worked:

- `can_format_file` only looks at files under `formatted_directories`, so the root CMakeLists.txt was rejected and the entry did nothing.
- The entry was a plain string instead of a `ToFormatFile`, so `process_file` crashed on it.

Fix: accept the root CMakeLists.txt in `can_format_file`, add it as a proper `ToFormatFile`, and run cmake-format on the file which drifted since #23211.


Mytherin:
Thanks!

Issue #25647, opened 2026-09-12
Table-in/out functions silently drop predicates when filter_pushdown is enabled


rustyconover:
### What happens?

A table-in/out function that sets `TableFunction::filter_pushdown = true` cannot receive the pushed `TableFilterSet`. DuckDB's filter-pushdown optimizer can move an eligible predicate into `LogicalGet::table_filters` and remove the residual filter, but the table-in/out physical-planning path then drops that filter state. The function therefore cannot fulfill the filter-pushdown contract, and the query can return unfiltered rows.

This is a correctness issue rather than only a missed optimization.

The relevant core path at commit `70e7a857595b5872842b6ec242ac79c182e58dde` is:

1. `FilterPushdown::PushdownGet` generates `LogicalGet::table_filters` when `get.function.filter_pushdown` is true:
   https://github.com/duckdb/duckdb/blob/70e7a857595b5872842b6ec242ac79c182e58dde/src/optimizer/pushdown/pushdown_get.cpp#L113-L160
2. `PhysicalPlanGenerator::CreatePlan(LogicalGet &op)` takes the `op.children` table-in/out branch and returns a `PhysicalTableInOutFunction` before the normal `MoveTableFilters` path:
   https://github.com/duckdb/duckdb/blob/70e7a857595b5872842b6ec242ac79c182e58dde/src/execution/physical_plan/plan_get.cpp#L33-L85
3. `PhysicalTableInOutFunction` constructs both global and local `TableFunctionInitInput` with `nullptr` filters:
   https://github.com/duckdb/duckdb/blob/70e7a857595b5872842b6ec242ac79c182e58dde/src/execution/operator/projection/physical_tableinout_function.cpp#L44-L70

This is consistent with the contract clarified in #19818: after setting `filter_pushdown`, the table function is expected to handle the pushed table filters. A table-in/out function currently has no opportunity to do that.

Expected behavior: predicates accepted into `LogicalGet::table_filters` and based solely on the table-in/out function's output should either be provided through `TableFunctionInitInput::filters` or retained as residual filters. They should never be silently discarded.

This issue is intentionally not requesting any change to correlated-filter semantics. Predicates involving columns from the input child should continue to follow their existing behavior.

### To Reproduce

Register a minimal table-in/out function that echoes an integer input column and advertises filter pushdown:

```cpp
TableFunction function("filter_probe", {LogicalType::TABLE}, nullptr, Bind, GlobalInit, LocalInit);
[...]

PR #25650, opened 2026-09-12
Fix filter pushdown for table in-out functions

rustyconover:
This PR fixes issue #25647.

Table in-out functions can advertise filter and projection pushdown, but the physical table in-out operator did not pass the generated `TableFilterSet` or projection IDs to `init_global` and `init_local`. Filters removed from the logical plan could therefore be silently skipped, producing incorrect results.

This PR:

1. forwards pushed filters and projection IDs to table in-out initialization;
2. keeps type-selective pushdown conservative, without changing partial or correlated-filter semantics;
3. retains filters above functions using `WITH ORDINALITY`, so surviving rows preserve their original ordinal positions;
4. remaps the ordinality column correctly after projection pruning.

The tests cover filter delivery to global and local initialization, filter-only columns, `filter_prune = false`, type-limited and correlated filters, lateral projection mapping, and `WITH ORDINALITY` behavior.


Mytherin:
Thanks!

PR #25651, opened 2026-09-12
fix: Integer Truncation in approx_k_values() leading to heap buffer overflow 

### LINKED ISSUE #25575: Vuln182: Integer Truncation in `ApproxTopKValue::CopyValue` Leads to Heap Buffer Overflow with Large Strings
### What happens?

In `extension/core_functions/aggregate/holistic/approx_top_k.cpp`, `CopyValue()` computes the re-allocation size by calling `UnsafeNumericCast<uint32_t>(NextPowerOfTwo(value.size))` where `value.size` is the byte length of the input string. For any string whose length is in the range `(2^31, 2^32-1]`, `NextPowerOfTwo` returns `2^32`, and the subsequent `uint32_t` cast truncates it to `0`. In a release build, `ArenaAllocator::Allocate(0)` hands back a zero-length region, and the immediately following `memcpy(value.dataptr, input.str.GetData(), value.size)` writes gigabytes of data past the end of that region, producing a heap buffer overflow. The bug is reachable by the `approx_top_k()` aggregate on a VARCHAR/BLOB column.

### To Reproduce

```sql
SET threads=1;
SELECT approx_top_k(s, 3)
FROM (
    FROM range(22)
) t;
```

```sh
echo "SET threads=1; SELECT approx_top_k(s, 3) FROM (SELECT string_agg(repeat('A', 100000000), '') AS s FROM range(22)) t;" | ./build/reldebug/duckdb
```

```
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

synix-pasha:
Fixes https://github.com/duckdb/duckdb/issues/25575

## Summary
`CopyValue()` rounded a strings size up to the next power of two, then truncated that result into a `uint32_t` via `UnsafeNumericCast` before allocating. For strings sized in (2^31, 2^32-1], the rounded size is exactly 2^32, which wraps to 0 on truncation. This made Allocate(0) return a zero byte buffer while the following memcpy still copied the full string into it.

## Fix
Widened `ApproxTopKValue::capacity` from `uint32_t` to `idx_t` and removed the truncating `UnsafeNumericCast<uint32_t>` around the power of two calculation.
The allocator now always receives the correctly computed size.
Oversized allocations now cleanly fail through DuckDBs normal `OutOfMemoryException` path instead of corrupting the heap.

## Test
`test_approx_top_k_overflow.test_slow`: produces a string leading to `2^32` rounding case and it will hit a clean Out of Memory error.


Mytherin:
Thanks!

Mytherin:
Could you just run the formatter?

synix-pasha:
Hi @Mytherin  just ran the formatter. Everything looks clean now. Can u please review it?

synix-pasha:
This was my first pull request. Done with formatting Finally. @Mytherin

DinosL:
Just a heads-up: you can run `make format-fix` before submitting the PR. It’ll automatically fix any formatting issues for you.

PR #25652, opened 2026-09-12
Optimize conflict manager rowid insertion

dentiny:
`ConflictManager` is a common data structure used in data insertion when uniqueness required, for example, UNIQUE constraint, primary key, etc.

I did some benchmark on data ingestion with `ConflictManager` involved, and I found there's some room we could improve.

TLDR: 
- Currently there're two data structures storing conflict information, one [hashset](https://github.com/duckdb/duckdb/blob/10de9573794001c649621013bdd93553b54e00c9/src/include/duckdb/common/types/conflict_manager.hpp#L145-L146), another is [`ConflictData`](https://github.com/duckdb/duckdb/blob/10de9573794001c649621013bdd93553b54e00c9/src/include/duckdb/common/types/conflict_manager.hpp#L150-L198)
- The former, from my understanding, covers the functionality for the hashset (namely, superset), so we could remove the hashset

Benchmark script
```sql
# name: benchmark/micro/index/point/repeated_fk_lookup_with_art.benchmark
# description: Insert repeated foreign keys that hit the same hot ART leaf
# group: [point]

name Repeated FK Lookup With ART
group art

load
PRAGMA threads=1;
CREATE TABLE parent (id BIGINT PRIMARY KEY);
INSERT INTO parent VALUES (42);
CREATE TABLE input AS SELECT 42::BIGINT AS id FROM range(10000000);
CREATE TABLE child (id BIGINT REFERENCES parent(id));

run
INSERT INTO child SELECT id FROM input;

cleanup
DELETE FROM child;
```
From the flamegraph, ~17% CPU is spent on the hashset-related operation 
image

Performance change for the benchmark
| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| Mean latency | 1.896 s | 1.510 s | 20.4% |
| Median latency | 1.872 s | 1.493 s | 20.2% |


artjomPlaunov (approved):
Thanks, LGTM!

PR #25653, opened 2026-09-12
Reduce one access syscall on local filesystem

dentiny:
Notice this when I was checking local filesystem, for all existence checks, we first do `access` then `stats`, which, from my perspective, could be reduced to one single `stats` call.

There're two concerns here:
- No change for error status or message propagation -- current implementation doesn't differ on `access` or `stats` failure
- There is one behavior difference but I don't think it's a concern for DuckDB users: `access` validate by real real UID/GID, but `stats` validates by effective UID/GID


Mytherin:
Thanks!

PR #25655, opened 2026-09-13
Avoid delete-and-insert for non-indexed UPDATE RETURNING

### LINKED ISSUE #25654: Unexpected foreign key constraint violation for `RETURNING`
### What happens?

See repro for details.

### To Reproduce

```sql
memory D CREATE TABLE parent (
             id INTEGER PRIMARY KEY,
             value INTEGER
         );
memory D CREATE TABLE child (
             parent_id INTEGER REFERENCES parent(id)
         );
memory D INSERT INTO parent VALUES (1, 10);
memory D INSERT INTO child VALUES (1);

-- id is not updated in this transaction, the FK constraint violation doesn't make sense
memory D UPDATE parent
         SET value = 11
         WHERE id = 1
         RETURNING value;
Constraint Error:
Violates foreign key constraint because key "parent_id: 1" is still referenced by a foreign key in a different table. If this is an unexpected constraintviolation, please refer to our foreign key limitations in the documentation
```

### OS:

macos

### DuckDB Version:

88902aaa3972216c928c8e79a89217a413be414c

### DuckDB Client:

SQL

### Hardware:

_No response_

### Full Name:

dentiny

### Affiliation:

N/A

- [x] Yes, I have

- [x] Yes, I have

Yes

dentiny:
Closes https://github.com/duckdb/duckdb/issues/25654

In the current implementation, [`LogicalUpdate::columns`](https://github.com/dentiny/duckdb/blob/ca5ce12b7dd33555f58d81ade0990442cb6eefca/src/include/duckdb/planner/operator/logical_update.hpp#L38) represents both columns explicitly modified by SET and columns merely materialized for constraint checks or RETURNING.

For `UPDATE ... RETURNING`, [BindUpdateConstraints](https://github.com/duckdb/duckdb/blob/88902aaa3972216c928c8e79a89217a413be414c/src/catalog/catalog_entry/table_catalog_entry.cpp#L297-L302) materialized every physical column through `BindExtraColumns`. Those columns were appended to `update.columns`, which was subsequently used to check constraints.

As a result, RETURNING made an update to a non-key column appear to modify the primary key, incorrectly rewriting it as DELETE + INSERT. Deleting the referenced parent row then produced a false foreign key violation.

In this PR, I split the responsibility into two parts:
- `referenced_columns`, which represents columns materialized for constraints, RETURNING, or reinsertion
- `columns_to_update`: columns explicitly targeted by the SET clause

### DIFF
--- a/src/include/duckdb/planner/operator/logical_merge_into.hpp
+++ b/src/include/duckdb/planner/operator/logical_merge_into.hpp
@@ -26,7 +26,9 @@ public:
 	//! Condition - or NULL if this should always be performed for the given action
 	unique_ptr<Expression> condition;
 	//! The set of referenced physical columns (for UPDATE)
-	vector<PhysicalIndex> columns;
+	vector<PhysicalIndex> referenced_columns;
+	//! Columns targeted by the UPDATE action
+	vector<PhysicalIndex> updated_columns;
 	//! Set of expressions for INSERT or UPDATE
 	vector<unique_ptr<Expression>> expressions;
 	//! Deprecated: Column index map (for INSERT)
--- a/src/include/duckdb/planner/operator/logical_merge_into.hpp
+++ b/src/include/duckdb/planner/operator/logical_merge_into.hpp
@@ -28,7 +28,7 @@ public:
 	//! The set of referenced physical columns (for UPDATE)
 	vector<PhysicalIndex> referenced_columns;
 	//! Columns targeted by the UPDATE action
-	vector<PhysicalIndex> updated_columns;
+	vector<PhysicalIndex> columns_to_update;
 	//! Set of expressions for INSERT or UPDATE
 	vector<unique_ptr<Expression>> expressions;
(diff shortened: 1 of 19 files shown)


src/include/duckdb/planner/operator/logical_merge_into.hpp
```diff
@@ -26,7 +26,9 @@ class BoundMergeIntoAction {
 	//! Condition - or NULL if this should always be performed for the given action
 	unique_ptr<Expression> condition;
 	//! The set of referenced physical columns (for UPDATE)
-	vector<PhysicalIndex> columns;
+	vector<PhysicalIndex> referenced_columns;
+	//! Columns targeted by the UPDATE action
+	vector<PhysicalIndex> columns_to_update;
```
dentiny, line 31: `columns_to_update` is a subset for `referenced_columns`, so we could also use vector index for `columns_to_update`, but think having another vector is clearer.

PR #21382, opened 2026-03-14
Checkpoint transactions

artjomPlaunov:
Have checkpoints operate within their own transaction in SingleFileCheckpointWriter::CreateCheckpoint. This happens when there is a client context available, in which case we create a new connection and transaction for the checkpoint. 
If there is no context (shutdown path), then we use the previous behavior of checkpointing on the last commit ID. 

Profiling should still work as is for now since 1) The StartTimer invocation for checkpoint latency contains in its scope the new connection + transaction and 2) in the CreateCheckpoint function itself, the original client context is still passed to various operations, thus registering e.g. total_bytes_written as an example within the original profiling context_data. 

--------------

More comments:

The motivation for this was that checkpointing has some transaction-like logic but isn't wrapped in its own transaction, and the cleanest fix might be to just wrap it in a transaction. This logic had to do with 1) preventing garbage collection of transactions using the active checkpoint ID to constrain the lowest_start_time in RemoveTransactions and 2) triggering the cleanup of a transaction. The start time of the new checkpoint transaction defines the active checkpoint ID which also deals with routing logic for concurrent operations while a checkpoint is running.

There is an is_checkpoint_transaction in DuckTransaction to prevent auto-checkpointing during a checkpoint transaction. 

The main locking considerations were to grab the WAL lock then the transaction lock in WALStartCheckpoint, to prevent a lock inversion with DuckTransactionManager::CommitTransaction which has a transaction lock release -> wal lock -> transaction lock.

PR #25660, opened 2026-09-13
Use checkpoint's transaction to bind indexes

Flogex:
When checkpointing a database with an unbound index and shortly after triggering a force checkpoint, this can result in a deadlock because both control paths try to acquire the same two locks but in reverse order. This is fixed by instead using the read-only transaction from `ActiveCheckpointWrapper` during index binding.

I also added a test, but note that it is relying on timing of two threads and could therefore silently pass. The sqllogictest runner now allows `sleep` inside `concurrentloop` since it has no shared state.

### Problem explanation

Two threads enter DuckTransactionManager::Checkpoint around the same time:

```
CHECKPOINT (thread 0)
-> Checkpoint(context, force=false)
-> checkpoint_lock.TryGetExclusiveLock()  [ACQUIRED]
-> needs start_transaction_lock           [BLOCKED]
FORCE CHECKPOINT (thread 1)
-> Checkpoint(context, force=true)
-> start_transaction_lock                 [ACQUIRED]
-> while (!lock) TryGetExclusiveLock()    [SPINS FOREVER]
```

The reason why the first checkpoint wants to acquire the `start_transaction_lock` is because there is an unbound index. Reduced stacktrace:
```
DuckTransactionManager::Checkpoint(context)
  checkpoint_lock.TryGetExclusiveLock()
  SingleFileStorageManager::CreateCheckpoint
    SingleFileCheckpointWriter::CreateCheckpoint
      WriteTable
        info->BindIndexes(*context)
          TableIndexList::Bind
            catalog.GetEntry<TableCatalogEntry>(context,…)
              Catalog::GetCatalogTransaction(context)
                CatalogTransaction(catalog, context)
                  Transaction::Get(context, catalog)
                    MetaTransaction::GetTransaction(db)
                      StartTransaction(context)
                        lock_guard(start_transaction_lock) // duck_transaction_manager
```

### Fix

Since #21382, a `SingleFileCheckpointWriter` creates their own transaction. And importantly, the transaction is read-only:
https://github.com/duckdb/duckdb/blob/5e9508026e1bbbd28f25863c33cfb03500063f81/src/storage/checkpoint_manager.cpp#L68

A read-only transaction does not lock the `start_transaction_lock`.
https://github.com/duckdb/duckdb/blob/10de9573794001c649621013bdd93553b54e00c9/src/transaction/duck_transaction_manager.cpp#L79-L82

With this PR, the `ClientContext` from `ActiveCheckpointWrapper` is passed to `DataTableInfo::BindIndexes`, used after `GetCheckpointTransaction` was already called and opened a read-only transaction. When `GetCheckpointTransaction` runs on thread 0, is already holds the `checkpoint_lock` but never tries to acquire the `start_transaction_lock`, therefore avoiding the circular wait.

Using a read-only transaction to bind an index is fine because it writes neither catalog nor data.


src/storage/checkpoint_manager.cpp
```diff
@@ -693,12 +699,15 @@ void SingleFileCheckpointWriter::WriteTable(TableCatalogEntry &table, Serializer
 	// Write the table metadata
 	serializer.WriteProperty(100, "table", &table);
 
-	// If there is a context available, bind indexes before serialization.
-	// This is necessary so that buffered index operations are replayed before we checkpoint, otherwise
-	// we would lose them if there was a restart after this.
-	if (context && context->transaction.HasActiveTransaction()) {
+	// Explicit checkpoints bind indexes before serialization, so that buffered index operations are replayed
+	// and not lost on a restart. During a commit-time checkpoint the caller has no active transaction.
+	if (context && context->transaction.HasActiveTransaction() && checkpoint_context) {
```
Flogex, line 704: I only appended to the original condition and kept `context->transaction.HasActiveTransaction()` around. Currently, only an explicit checkpoint binds indexes and automatic checkpoints on commits do not, and this clause is what distinguishes both. I didn't want to change the behavior (vacuum/remapping only during explicit checkpoints).

test/sql/parallelism/interquery/concurrent_force_checkpoint_unbound_index.test_slow
```diff
@@ -0,0 +1,64 @@
+# name: test/sql/parallelism/interquery/concurrent_force_checkpoint_unbound_index.test_slow
```
Flogex, line 1: Reproduction conditions: To call `BindIndexes` during checkpoint, we need a freshly attached database to have an unbounded index and a non-empty WAL so the checkpoint does work.

Mytherin:
Thanks!

PR #25662, opened 2026-09-13
Fix integer truncation in ArgMinMaxValueAssign leading to heap buffer overflow (#25576)

### LINKED ISSUE #25576: Vuln183: Integer Truncation in `ModeStringType::Update` Leads to Heap Buffer Overflow with Large Strings
### What happens?

In `extension/core_functions/aggregate/holistic/mode.cpp`, `ModeStringType::Update()` computes the arena allocation size by calling `UnsafeNumericCast<uint32_t>(NextPowerOfTwo(len))` where `len` is the byte length of the candidate mode string. For any string whose length is in the range `(2^31, 2^32-1]`, `NextPowerOfTwo` returns `2^32`, and the subsequent `uint32_t` cast truncates it to `0`. In a release build, `ArenaAllocator::Allocate(0)` returns a zero-length region, and the immediately following `memcpy(ptr, key.GetData(), len)` writes gigabytes past that region, producing a heap buffer overflow. The bug is reachable by the `mode()` aggregate on a VARCHAR/BLOB column.

### To Reproduce

```sql
SET threads=1;
SELECT mode(s)
FROM (
    FROM range(22)
) t;
```

```sh
echo "SET threads=1; SELECT mode(s) FROM (SELECT string_agg(repeat('A', 100000000), '') AS s FROM range(22)) t;" | ./build/reldebug/duckdb
```

```
```

### OS:

### DuckDB Version:

### DuckDB Client:

### Hardware:

_No response_

### Full Name:

Yuelin

### Affiliation:

Individual

- [x] Yes, I have

- [x] Yes, I have

Yes

tanishqjh-tron:
Fixes #25576

### Summary
Fixes a `SIGSEGV` crash occurring during `mode`, `arg_min`, and `arg_max` aggregations when processing string values that require an arena buffer allocation $\ge 4\text{ GiB}$.

### Root Cause
In `ArgMinMaxValueAssign<string_t>::Assign` (`extension/core_functions/aggregate/distributive/arg_min_max.cpp`), the allocation size was stored as a 32-bit unsigned integer and truncated via `UnsafeNumericCast<uint32_t>`:

```cpp
alloc_size = UnsafeNumericCast<uint32_t>(NextPowerOfTwo(len));
```

For string payloads where $len > 2^{31} - 1$ (e.g., 2.2 GB), `NextPowerOfTwo(len)` evaluates to $2^{32} = 4{,}294{,}967{,}296$, which exceeds `UINT32_MAX`. Casting this value to `uint32_t` overflows and wraps to `0`. Consequently, `aggregate_input_data.allocator.Allocate(0)` provisions a zero-sized chunk, and the subsequent `memcpy(ptr, new_value.GetData(), len)` writes past the allocated boundary, triggering a heap buffer overflow and `SIGSEGV`.

### Solution

* Promoted `alloc_size` in `ArgMinMaxValueAssign<string_t>` from `uint32_t` to 64-bit `idx_t`.
* Removed `UnsafeNumericCast<uint32_t>` wrappers around the length calculation and `NextPowerOfTwo` call so allocations scale safely with 64-bit address space.

### Reproduction Query

```sql
PRAGMA memory_limit='30GB';
SET threads=1;

SELECT mode(s) FROM (
    FROM range(22)
) t;
```

### Verification

* Before fix: Instant `SIGSEGV` during `memcpy` due to 0-byte buffer allocation.
* After fix: Allocator correctly requests a 4 GiB buffer. The query runs to completion when memory is available, or gracefully raises an `Out of Memory Error` via the Buffer Manager when bounded by `memory_limit`.


tanishqjh-tron:
The core tests and benchmarks have passed. I have the reproduction query ready, but since it requires a 4 GiB allocation to trigger the 32-bit overflow, I held off on adding it directly to the SQLLogicTest suite to avoid OOMing the standard CI runners.

Let me know if you would like me to add it under a specific require no_memory_limit flag or if manual verification is sufficient for this patch!

Mytherin:
Thanks - could you retarget this to `v2.0-cyanoptera`?

tanishqjh-tron:
I have re-targeted to v2.0 cynoptera. Kindly merge after automated tests are cleared.

tanishqjh-tron:
Thanks for the review and merge! Digging into the arena allocator was a great experience. I will be sending an application over to the internships@duckdblabs.com this week. I would love the opportunity to contribute to these core systems full-time during a winter/summer internship or part-time during the semester.

tanishqjh-tron:
Just a quick update, it looks like the internships@duckdblabs.com address I mentioned above is bouncing (likely due to the AWS transition!).

Is there a preferred active email or channel to reach out regarding winter/summer intern headcount and remote possibilities? (I also CC'd hello@ducklabs.com just in case).

I am also currently working on trade_math (a vectorized quant extension for DuckDB) and will be submitting it to the community repo!

PR #25664, opened 2026-09-13
C API V1: mark extension scalar functions as fallible by default

staticlibs:
In 2.0 scalar functions must call `SetFallible()` to be able to report execution errors. This creates a problem for existing C API V1 extension that register scalar functions.

For example [odbc-scanner](https://github.com/duckdb/odbc-scanner) running on `v2.0-cyanoptera` reports the following:

> INTERNAL Error: Scalar function ""odbc_bind_params"" threw an execution error, but the function is not marked as fallible - the function must call SetFallible().

In C API V1 we don't have the `SetFallible()` and even if it is added there, it won't immediately help to existing C API V1 extensions that are binary compatible with all DuckDB 1.2.0+ versions.

Apparent solution is to move the extension to C API V2, but that may be a non-trivial task, and additionally makes the extension to lose both binary and source compatibility with `v1.4` LTS.

Thus it is proposed to mark all scalar functions registered through C API V1 entry point as fallible by default.

Testing: test case added that checks that the error message reported from C API V1 function does not contain "INTERNAL Error".

Ref: duckdblabs/duckdb-internal#7407

CC: @Maxxen


Maxxen:
Awesome, ran into this in the rust bindings too and was about to make a PR, thanks!

taniabogatsch:
@Maxxen, I only see the Rust PR, is this something that also needs to change in core?

Maxxen:
@taniabogatsch this did change it in core?

taniabogatsch:
Nevermind, I was confused 😅

PR #25665, opened 2026-09-13
Support scalar subqueries in table AT clauses

rustyconover:
Table `AT` clauses are bound before normal expression execution, which currently requires callers to discover a snapshot in one query and interpolate it into a second query. This change accepts expressions containing an uncorrelated scalar subquery in a table `AT` clause for top-level `SELECT` statements.

```sql
SELECT *
FROM lake.events AT (
    VERSION => (
        SELECT max(snapshot_id)
        FROM ducklake_snapshots('lake')
        WHERE schema_version = 1
    )
);
```

The PEG transformer records each selector as a statement-owned standalone `SELECT`. Before the main statement is verified or bound, `ClientContext` executes each selector once, retains the typed `Value`, and replaces the matching `AT` expression with a `ConstantExpression`. Catalog and extension time-travel bind paths therefore continue to receive the constant expressions they expect.

The initial scope is deliberately narrow:

- selectors are uncorrelated and do not inherit outer aliases or enclosing CTEs
- parameterized and prepared statements are rejected
- persistent/wrapper contexts such as views, macros, triggers, and `EXPLAIN` are rejected
- dynamic `PIVOT` in selectors is rejected

Statement copies preserve the pre-bind selectors, multiple selectors are evaluated before the main bind, and nested selector dependencies are resolved in parse order.


rustyconover:
Published a write-up of what this draft prototype uncovered, including the bind-time lifecycle, transaction boundary, typed handoff, and the path from uncorrelated selectors to same-schema lateral temporal joins:

[From Time Travel to Lateral Temporal Tables in DuckDB](https://rusty.today/blog/duckdb-at-clause-lateral-temporal-tables/)

The PR remains a draft; the article captures both the working proof of concept and the production design constraints.

PR #25666, opened 2026-09-13
fix: Preserve tags when copying index catalog entries

rustyconover:
## Summary

Restore index tags from `CreateIndexInfo` when constructing an `IndexCatalogEntry`.

`IndexCatalogEntry::GetInfo()` already copies tags into `CreateIndexInfo`, but the constructor did not restore them. As a result, copying an index catalog entry silently dropped its tags.

## Tests

- `build/reldebug/test/unittest`


rustyconover:
Hi @Mytherin,

This is a one line change, but the CI test seem to be a unreliable rather than related to this PR.

Rusty

Mytherin:
I think `main` is broken atm, but I think this should be targeting `v2.0` anyway. Could you retarget?

rustyconover:
Hi @Mytherin,

I've changed the base branch as asked and this should be ready for your review now.

Rusty

rustyconover:
@Mytherin it seems CI is still flaking out here on the v2.0 branch as well?  There isn't a reason why this diff should cause a performance regression failure.

PR #25672, opened 2026-09-14
Fix temporary file spill

dentiny:
Hi team, I met several issues when spilling blocks to disk fails, for example
```sql
haojiang@HaoJiangs-MacBook-Pro$ ./build/reldebug/duckdb /tmp/offload_failure_accounting.db
DuckDB v2.0.0-dev84153 (Development Version, 96f4ca8005)
offload_failure_accounting D SET default_block_size=262144;
offload_failure_accounting D PRAGMA memory_limit='8MB';
offload_failure_accounting D SET max_temp_directory_size='0B';
offload_failure_accounting D CREATE TABLE t AS SELECT repeat('x', 300000)::VARCHAR s, i FROM range(200) tbl(i);
Out of Memory Error:
failed to offload data block of size 296.0 KiB (0 bytes/0 bytes used).
....
offload_failure_accounting D SELECT count(*) FROM duckdb_memory() WHERE temporary_storage_bytes > 0;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            1 │
└──────────────┘
offload_failure_accounting D CALL duckdb_temporary_files();
┌─────────┬───────┐
│  path   │ size  │
│ varchar │ int64 │
└─────────┴───────┘
      0 rows
```
Also filesystem persistence
```sh
haojiang@HaoJiangs-MacBook-Pro$ ls /tmp/offload_failure_accounting.db.tmp
duckdb_temp_60601_0_block-4611686018427388007.block
duckdb_temp_60601_0_owner

haojiang@HaoJiangs-MacBook-Pro$ ls -l /tmp/offload_failure_accounting.db.tmp/duckdb_temp_60601_0_block-4611686018427388007.block
-rw-r--r--@ 1 haojiang  wheel  0 Sep 13 16:39 /tmp/offload_failure_accounting.db.tmp/duckdb_temp_60601_0_block-4611686018427388007.block
```

There're several issues
- On spill failure, there should be no temporary storage bytes, but DuckDB reports one entry
- There's an empty temporary block file created, to make things worse, it's not tracked by DuckDB

In this PR, I did two changes to fix the issue at best effort
- Account for disk storage space only after the spill operation succeds
- Delete the temporary block file if write fails; it's a best effort attempt, which doesn't save in cases where, say, process crashes


Mytherin:
Thanks!

PR #25673, opened 2026-09-14
Do not traverse window and aggregate children twice in filter pushdown

### LINKED ISSUE #25373: Filter pushdown takes exponential time for nested window operators
### What happens?

The filter-pushdown optimizer recursively processes the same child plan twice for each nested `LOGICAL_WINDOW`. Therefore, planning time approximately doubles for every additional window layer, even when the query contains no filter and reads only ten generated rows:

| Window layers | Wall time | `optimizer_filter_pushdown` | Execution CPU |
|---:|---:|---:|---:|
| 12 | 0.0053 s | 0.0018 s | 0.0013 s |
| 14 | 0.0122 s | 0.0071 s | 0.0032 s |
| 16 | 0.0322 s | 0.0276 s | 0.0020 s |
| 18 | 0.1142 s | 0.1093 s | 0.0022 s |
| 20 | 0.4497 s | 0.4444 s | 0.0023 s |
| 22 | 1.7820 s | 1.7762 s | 0.0024 s |

When disabling `filter_pushdown`, the 22-layer query takes 0.0060 seconds.

### To Reproduce

```python
import json
import time

import duckdb


def build_query(layers):
    query = "SELECT i % 2 AS id, i AS value FROM range(10) t(i)"
    for layer in range(layers):
        query = f"""SELECT *, MAX(value) OVER (PARTITION BY id) AS w{layer}
        FROM ({query}) q{layer}"""
    return f"SELECT COUNT(*) FROM ({query}) q"


for layers in (12, 14, 16, 18, 20, 22):
    connection = duckdb.connect()
    connection.execute("PRAGMA profiling_mode = 'detailed'")
    started = time.perf_counter()
    profile_json = connection.execute(
        "EXPLAIN (ANALYZE, FORMAT JSON) " + build_query(layers)
    ).fetchone()[1]
    profile = json.loads(profile_json)
    print(
        layers,
        time.perf_counter() - started,
        profile["optimizer_filter_pushdown"],
        profile["cpu_time"],
    )
```

### OS:

MacOS

### DuckDB Version:

1.5.5

### DuckDB Client:

Python

### Hardware:

Macbook Pro 14 M2 Pro

### Full Name:

Leonard Stimpfle

### Affiliation:

Ghent University

- [x] Yes, I have

- [x] Yes, I have

RyanLin5967:
`PushdownWindow` and `PushdownAggregate` rewrite their child, then call `FinishPushdown`, which rewrites every child again. so each nested window or aggregate walks its subtree twice and planning doubles per level.

`PushFinalFilters` does the finishing work without the second descent, like how the join handlers already do.

30 nested windows over 100 rows: 167s before, 0.09s after. aggregates are similar, 82.9s to 0.02s.

fixes #25373


kryonix (approved):
Thanks! LGTM.

PR #25674, opened 2026-09-14
Fix invalid memory access on optimistic write during scan

dentiny:
Hi team, I met a SIGSEGV or SIGBUS (non-deterministically) on the following scenario.
```sql
haojiang@HaoJiangs-MacBook-Pro$ ./build/reldebug/duckdb /tmp/optimistic_write_cyclic_scan_single_thread.db
DuckDB v2.0.0-dev10995 (Development Version, 88902aaa39)
optimistic_write_cyclic_scan_single_thread D PRAGMA threads=1;
optimistic_write_cyclic_scan_single_thread D SET write_buffer_row_group_count=1;
optimistic_write_cyclic_scan_single_thread D CREATE TABLE t(a INTEGER);
optimistic_write_cyclic_scan_single_thread D BEGIN;
optimistic_write_cyclic_scan_single_thread D INSERT INTO t SELECT range FROM range(100000) tbl(range);
optimistic_write_cyclic_scan_single_thread D INSERT INTO t SELECT * FROM t;
💣 Program crashed: Bus error at 0x00000000005b8bde
Platform: arm64 macOS 26.5.2 (25F84)
Thread 0 crashed:
  0 0x00000001028230ac duckdb::StandardColumnData::GetVectorScanType(duckdb::ColumnScanState&, unsigned long long, duckdb::Vector&) + 104 in duckdb
  1 0x00000001028232e0 duckdb::StandardColumnData::Scan(duckdb::TransactionData, unsigned long long, duckdb::ColumnScanState&, duckdb::Vector&, unsigned longlong) + 80 in duckdb
```
The root cause is
- In a single-threaded self-referencing `INSERT INTO t SELECT * FROM t`, the scan and the append interleave on the same local `RowGroupCollection`
- On the scan side, scan state [holds a reference](https://github.com/duckdb/duckdb/blob/10de9573794001c649621013bdd93553b54e00c9/src/include/duckdb/storage/table/scan_state.hpp#L168-L169) instead of an ownership, so when `OptimisticWriter` optimistically flushes the row group and advances to the next, the reference becomes invalid

Pseudocode looks like (asked GPT to draw the code)
```sh
InitializeScan(state, node_0):
    state.row_group = node_0                            // cursor, non-owning borrow
    for each column i:
        column_scans[i].segment_tree = &old_rg0.column[i].segment_tree   // ← raw pointer INTO old_rg0
        column_scans[i].current      = segment node of old_rg0.column[i] // ← raw pointer
    state.vector_index = 0

// ===== pipeline loop: scan/append interleaved on the same thread =====
while row group 0 not fully scanned:        // note: NO re-initialization happens in here!
    // --- scan side: read 2048 rows ---
    row_group->GetNode().Scan(options, state, chunk)
        PrepareScan:         compute visible rows for this vector via version_info
        ProcessPreparedScan: 
            for each column i:
                col_data = this->GetColumn(i)            // this = whichever row group Scan was called on
                col_data.Scan(state.column_scans[i])     // ← follows the raw pointers in column_scans
        state.vector_index++                             // advance one vector WITHIN the same row group

    // --- append side: write back into the SAME collection ---
    Append(chunk)
        ... row group 0 fills up (122880 rows) ...
        FlushToDisk:  new_rg0 = WriteToDisk(old_rg0)     // brand-new persistent row group
        SetRowGroup(0, new_rg0)
            node_0.SetNode(new_rg0)                      // in-place payload swap, old_rg0 all destructed

// ===== next loop iteration: the scan is STILL inside row group 0 =====
    row_group->GetNode().Scan(options, state, chunk)
    //   node_0 is alive; GetNode() returns new_rg0 — no UAF at this level
    //   but state.column_scans[i].segment_tree / .current still point into old_rg0's FREED ColumnData
    //    → col_data.Scan follows dangling pointers → SIGSEGV (or silently reads garbage rows if the memory was reused)
```

In this PR, I propose to pin in the row group scan is accessing, so all scan access is valid
pseudocode explanation on why do we need two variables
```sh
advance:   row_group = node_k;  pinned = node_k.payload        // converge
flush:     node_k.payload = new_rg_k                           // window opens: row_group->GetNode() = new,
                                                               //   pinned = old (kept alive), column_scans → old
scan:      pinned->Scan(...)                                   // reads the snapshot, safe
advance:   pinned = node_{k+1}.payload                         // window closes: old rg destroyed here
```


Mytherin:
Thanks - having both `pinned_row_group` and `row_group` seems somewhat error-prone though. Could we perhaps replace `row_group` with the `shared_ptr` instead?

dentiny:
> Thanks - having both `pinned_row_group` and `row_group` seems somewhat error-prone though. Could we perhaps replace `row_group` with the `shared_ptr` instead?

Thanks for the quick review! This is sth I considered, the concern is
- `row_group` currently points to the `SegmentNode`, which is needed for row offsets and advancing/reordering 
- while the shared pointer pins its replaceable payload, I think replacing it directly with `shared_ptr<RowGroup>` would lose the traversal cursor?

Mytherin:
Could we then maybe make these two fields private so we can ensure they never get out of sync?

dentiny:
> Could we then maybe make these two fields private so we can ensure they never get out of sync?

sure! makes perfect sense to me, updated

Mytherin:
Thanks - code looks good to me, could you fix the CI?

PR #25675, opened 2026-09-14
Fix DISTINCT column pruning state leaking into sibling inputs

ted-wq-x:
### Problem

The SQL below prunes an unused struct payload on DuckDB 1.5.5, but retains it on the tested 2.0 development builds. Query results remain correct; the regression is in unused-column pruning.

Versions tested:

- v1.5.5 (`d8cdaa33fd`): unused payload is pruned.
- v2.0 development (`4d4b9b0ad8`, `a36e86587f`): unused payload remains.

The DISTINCT state handling predates 2.0. These observations establish a behavioral regression for this query, but do not identify the commit that first exposed it.

### Reproduction

```sql
CREATE TABLE person (
    id INTEGER PRIMARY KEY,
    name VARCHAR,
    age INTEGER
);
INSERT INTO person VALUES
    (1, 'alice', 10),
    (2, 'bob', 20),
    (3, 'carol', 30);

CREATE TABLE knows(src_id INTEGER, dst_id INTEGER);
INSERT INTO knows VALUES (1, 2), (1, 3), (2, 3);

SET explain_output = 'optimized_only';

EXPLAIN
SELECT f.name
FROM person p
LEFT JOIN (
    SELECT k.src_id, n.name
    FROM knows k
    JOIN (
        SELECT *, struct_pack(age := age) AS unused_payload
        FROM person
    ) n ON n.id = k.dst_id
    JOIN (
        SELECT DISTINCT id FROM person
    ) s ON s.id = k.src_id
) f ON p.id = f.src_id;
```

Expected: `unused_payload` is absent from the optimized plan because nothing references it.

Actual on the tested 2.0 builds: the projection retains `unused_payload` and `age`. On 1.5.5, that projection contains only `id` and `name`.

### Fix

Ordinary DISTINCT sets `RemoveUnusedColumns::everything_referenced` to preserve its input columns, but previously left that flag set after traversing its subtree. Subsequent sibling inputs visited by the same pruner could therefore retain unused expressions.

Save and restore the incoming flag around the DISTINCT subtree traversal. This preserves DISTINCT semantics and existing column-reference tracking while allowing sibling inputs to be pruned. DISTINCT ON retains its existing behavior.

Add SQL regression coverage for unused payload pruning, unselected DISTINCT keys affecting result multiplicity, nested DISTINCT, DISTINCT ON, and materialized CTEs.


Mytherin:
Thanks!

PR #25676, opened 2026-09-14
Fix filter pushdown through projections before delim CTE conversion

### LINKED ISSUE #25565: Regression: outer filter not pushed into LATERAL recursive CTE input after delim joins became CTEs
### What happens?

An outer `id = 3` filter is no longer pushed to the left input of a LATERAL join containing a recursive CTE when the join is wrapped in a derived table.

Results remain correct, but recursion processes all 200,000 input rows instead of one: recursive output increases from 8 rows in v1.5.5 to 1,600,000 in the tested development build.

### To Reproduce

Run in a fresh DuckDB CLI session; no extensions or external data are needed:

```sql
SET threads = 1;

CREATE TABLE vertices AS
SELECT i::INTEGER AS id FROM range(1, 200001) AS t(i);

EXPLAIN (ANALYZE, FORMAT JSON)
SELECT count(*)
FROM (
    SELECT t.id, r.depth
    FROM vertices AS t,
    LATERAL (
        WITH RECURSIVE walk(depth) AS (
            SELECT 0 + t.id * 0
            UNION ALL
            SELECT depth + 1 FROM walk WHERE depth < 7
        )
        SELECT depth FROM walk
    ) AS r
) AS q
WHERE id = 3;
```

Without `EXPLAIN`, both versions return `8`. Actual operator cardinalities:

| Version | Base scan output rows | Recursive output rows | Scan filter |
| --- | ---: | ---: | --- |
| v1.5.5 (`d8cdaa33fd`) | 1 | 8 | `id = 3` |
| v2.0.0-dev84248 (`4d4b9b0ad8`) | 200,000 | 1,600,000 | absent |

Removing the derived-table wrapper and applying `WHERE t.id = 3` directly to the lateral join restores the scan filter and 8 recursive rows on the same development build.

The regression was reproduced across #23333: parent `614e8e7ef0` pushes the filter with its default settings; merge `fe871ebf45` does not. At those revisions, setting `delim_join_as_cte=true` reproduces the missing pushdown, while `false` restores it.

### OS:

macOS 26.3.1 (a), arm64

### DuckDB Version:

Affected: v2.0.0-dev84248 (`4d4b9b0ad836c94cb64fcfd0443e592f92ec9d53`), local release build.
Comparison: v1.5.5 (`d8cdaa33fd`), Homebrew CLI.

### DuckDB Client:

CLI, in-memory database

### Hardware:

Apple M1 Max, 32 GiB RAM; `threads = 1`

### Full Name:

ted-wq-x

### Affiliation:

Independent developer

- [x] Yes, I have

- [x] Yes, I have

Yes — generated by the SQL above.

ted-wq-x:
Fixes #25565.

Outer filters on derived tables could miss the filter pushdown performed before delim joins are converted to CTEs. The resulting materialized input remained unfiltered, causing lateral recursive queries to process unnecessary rows.

Extend the existing pre-conversion pushdown to traverse projection chains and remap direct column references. The new path handles left-input predicates on INNER and LEFT delim joins, rejects unsafe expressions, and preserves predicates that cannot be pushed.

For the issue reproducer, scan output drops from 200,000 rows to 1 and recursive output from 1,600,000 rows to 8.


kryonix (approved):
Thanks for working on this! LGTM.

PR #25677, opened 2026-09-14
Fix non standard vector size CI run

carlopi:
Two independent fixes, moving "[Tests non-default vector and block size](https://github.com/duckdb/duckdb/actions/runs/34729584451/job/103652708882#logs)" to pass (or at least fail later).

First, remove assertion that hold on most cases (but not all...) and second adding `require vector_size 2048` (and swapping around so sensible query becomes the last.


Mytherin:
Thanks!

PR #25678, opened 2026-09-14
Fix `USING COMPRESSION` for `ALTER TABLR ADD COLUMN`

dentiny:
Hi team, I found `USING COMPRESSION` doesn't take effect when we add a new column to  the table; for example,
```sql
memory D CREATE TABLE compression_bug(i INTEGER);
memory D INSERT INTO compression_bug SELECT * FROM range(10000);
memory D ALTER TABLE compression_bug
         ADD COLUMN added INTEGER DEFAULT 42
         USING COMPRESSION UNCOMPRESSED;
memory D CHECKPOINT;
-- should be "uncompressed"
memory D SELECT DISTINCT compression FROM pragma_storage_info('compression_bug') WHERE column_name = 'added' AND segment_type = 'INTEGER';
┌─────────────┐
│ compression │
│   varchar   │
├─────────────┤
│ Constant    │
└─────────────┘
```


Tishj (approved):
Thanks!

PR #25680, opened 2026-09-14
Fix cardinality for descending generate_series

p1p1bear:
Fix the row-count estimate for descending generate_series, which could cause the optimizer to incorrectly remove LIMIT and return too many rows.

It is a tiny fix. :)


Mytherin:
Thanks - change looks good to me, could we retarget this to `v2.0-cyanoptera`?

p1p1bear:
OK，I'll open an new PR:)

p1p1bear:
> Thanks - change looks good to me, could we retarget this to `v2.0-cyanoptera`?

@Mytherin  https://github.com/duckdb/duckdb/pull/25691 It is a new PR, thanks

PR #25681, opened 2026-09-14
Added connection_progress_get to the capi v2

okruitho:
This PR adds connection_progress_get to replace the 4 needed calls before in the C API V2.


Maxxen:
hmmmm it should probably be called `connection_get_progress` - but okay, I can adjust it in a follow-up PR, I'm sure we will have to do another pass over naming for the entire API later anyway.

carlopi:
Was this intended to be merged in `main` or `v2.0-cyanoptera` ?

Maxxen:
... probably cyanoptera

PR #25682, opened 2026-09-14
Support the euclidian operator `<->`

Dtenwolde:
Part of https://github.com/duckdblabs/duckdb-internal/issues/9093

Suports the `<->` operator again, which is a built-in operator for measuring the euclidian distance. 

```sql
DuckDB v2.0.0-dev84544 (Development Version, f79c7dd614)
memory D SELECT [1.0, 2.0] <-> [2.0, 4.0];
┌─────────────────────────────────────────────────┐
│ (list_value(1.0, 2.0) <-> list_value(2.0, 4.0)) │
│                     double                      │
├─────────────────────────────────────────────────┤
│                                2.23606797749979 │
└─────────────────────────────────────────────────┘
```


Maxxen:
Q: Why does the parser need all these special cases for operators?

AFAIK in PSQL any combination of ``+ - * / < > = ~ ! @ # % ^ & | ` ?`` is parsed as an operator. 

I think an additional cool thing would be if we also automatically derived the operator _precedence_ based on the actual symbols in the operator itself, so that e.g. `++` binds harder than `+`, but not as hard as `*` (or some other scheme).

Dtenwolde:
I think handling the "generic" operators in the grammar can be removed, but we do need to keep some special handling for recognizing operators with different precedence levels. For instance all the `LIKE` functions `~~`, `!~~`, etc. we could also detect those inside an `OperatorMatcher`. Also we could look into generating some of this precendence from a file and generate it. Though I don't think it's likely we're ever touching/changing the precendence of operators. 

I worked on a patch to remove the operators from the grammar and that seems pretty doable, but I will do that in a follow-up PR.

Mytherin:
Thanks - agreed that we should minimize explicit operators in the parser so would be great if you can clean that up in a follow-up

Dtenwolde:
Opened this issue: https://github.com/duckdblabs/duckdb-internal/issues/10769

PR #25683, opened 2026-09-14
Parallelize invariant recursive builds with small frontiers

ted-wq-x:
## Problem

Recursive CTEs with a small initial frontier can execute a large invariant hash build on a single worker, even when multiple threads are available.

The recursive runtime selects inline execution based on the frontier size. This applies a worker limit of one to every EXECUTE stage, including recursion-independent build pipelines whose scan work can be much larger than the frontier.

For example, a two-hop traversal starting from one vertex can build a hash table over 40 million bidirectional edges through the single-worker inline path.

## Changes

- Mark schedule stages belonging to already-classified invariant builds.
- Allow invariant build stages to use the configured thread limit in the Event-based path. `Pipeline::GetMaxThreads()` still enforces source, operator, and sink parallelism limits.
- In the inline path, prepare the pipeline and check its available parallelism before allocating an Event:
  - Single-worker pipelines continue inline.
  - Parallel builds use the existing `RecursiveCTEPipelineEvent` and wait for completion before advancing to the remaining stages.
- Add `SchedulePrepared()` to schedule an already-prepared pipeline without resetting its state twice.
- Add `EpochSummary.max_pipeline_workers`, an atomically maintained maximum of the worker count scheduled for one EXECUTE stage across the recursive invocation.

Existing `workers_max` and `scheduled_workers` retain their original scheduling-policy semantics. The new metric does not include finalize tasks or represent the number of CPU threads active at the same instant.

Build-side selection, invariant-state reuse, dependency ordering, finalization, and frontier scheduling remain unchanged.

## Reproducer

No external dataset or extension is required.

The following SQL generates 20 million directed edges over 2 million vertices, with ten outgoing edges per vertex. The query expands them bidirectionally and traverses two hops from a single starting vertex.

```sql
SET threads = 16;
SET memory_limit = '16GB';

CREATE TABLE edges AS
SELECT
    (i // 10)::BIGINT AS src,
    (((i // 10) * 13 + i % 10 + 7) % 2000000)::BIGINT AS dst
FROM range(20000000) AS t(i);

WITH RECURSIVE bidirectional_edges AS (
    SELECT src, dst FROM edges
    UNION ALL
    SELECT dst AS src, src AS dst
    FROM edges
    WHERE src <> dst
), walk(node, depth) AS (
    SELECT 65504::BIGINT, 0
    UNION ALL
    SELECT e.dst, w.depth + 1
    FROM walk w
    JOIN bidirectional_edges e ON e.src = w.node
    WHERE w.depth < 2
)
SELECT depth, count(*) AS paths, sum(node) AS checksum
FROM walk
GROUP BY depth
ORDER BY depth;
```

Expected result:

```text
depth  paths  checksum
0      1      65504
1      20     16719860
2      400    310599200
```

Run the query once to warm up. Then repeat the same `WITH RECURSIVE` statement prefixed with:

```sql
EXPLAIN (ANALYZE, FORMAT JSON)
```

Compare `query.total_time`, excluding `CREATE TABLE` time. No optimizer settings need to be disabled.

Both measured plans scan 40 million edge rows and retain the static edge relation on the hash-build side.

## Performance

Environment:

- Intel Xeon E5-2630 v4 @ 2.20 GHz
- 40 available logical CPUs; 16 query threads
- 16 GB query memory limit
- Release build
- DuckDB commit `0cb577bfcf`, before and after applying the patch

Each version ran in a separate session on the same machine, with one warmup followed by five profiled executions.

| Version | Median query time |
|---|---:|
| Before | 1.850 s |
| After | 0.410 s |

The reproducer is approximately **4.51× faster**, with identical results.

The original LDBC SF100 IC6 workload also improved from **4.830 s to 1.908 s** in an earlier comparison of the scheduling change, while retaining one invariant build.

These measurements demonstrate the improvement on this machine and workload; they are not a guarantee for all recursive queries.

## Validation

- Release shell build passed.
- Formatting and whitespace checks passed.
- Reproducer results matched before and after the patch.
- IC6 checks at 1, 16, and 40 threads retained exactly one invariant build.
- The new `max_pipeline_workers` metric reported 1, 16, and 40 respectively.
- Existing worker metrics retained their previous values in those checks.
- A 100,000-edge recursive chain returned identical results before and after the scheduling change.

Unit tests were not run. Cancellation, spill, and nested-recursion boundary coverage remain outstanding.

## Scope

This change only expands parallelism for builds recognized by the existing invariant classifier. Other independent scans and invariant CTE materializations are outside its scope.

The inline path keeps separate pipelines in their existing dependency order while allowing multiple workers within an eligible build pipeline.


kryonix (changes requested):
I would like this PR to add a dedicated benchmark before merging. The existing tiny-frontier benchmark has no invariant build, so it cannot protect this improvement. Specifically, I want the following synthetic two-hop traversal, with edge generation in the benchmark's **load phase**:

```sql
CREATE TABLE edges AS
SELECT
    (i // 10)::BIGINT AS src,
    (((i // 10) * 13 + i % 10 + 7) % 2000000)::BIGINT AS dst
FROM range(20000000) AS t(i);
```

Only this query should be timed:

```sql
WITH RECURSIVE bidirectional_edges AS (
    SELECT src, dst FROM edges
    UNION ALL
    SELECT dst AS src, src AS dst
    FROM edges
    WHERE src <> dst
), walk(node, depth) AS (
    SELECT 65504::BIGINT, 0
    UNION ALL
    SELECT e.dst, w.depth + 1
    FROM walk w
    JOIN bidirectional_edges e ON e.src = w.node
    WHERE w.depth < 2
)
SELECT depth, count(*) AS paths, sum(node) AS checksum
FROM walk
GROUP BY depth
ORDER BY depth;
```

The expected-answer file should contain:

```text
depth|paths|checksum
0|1|65504
1|20|16719860
2|400|310599200
```

Please leave the thread count configurable so we can exercise both the single-thread fallback and the parallel build. I validated this benchmark with one warmup and five measured runs per revision: at eight threads, the median improved from **330 ms to 121 ms** on an M5 Max.

I would also add a smaller regression test checking that the build is retained exactly once, `workers_max` remains 1, and `max_pipeline_workers` follows the configured thread count. That protects the scheduling decision directly, alongside the benchmark protecting its performance.

src/execution/operator/set/physical_recursive_cte_runtime.cpp
```diff
@@ -807,6 +826,17 @@ static void ExecuteRecursiveInlinePlan(RecursiveCTEState &state, Executor &execu
 		switch (stage.type) {
 		case PipelineScheduleStageType::EXECUTE: {
 			pipeline.ResetForReschedule(false);
+			// Invariant builds have independent source work even when the recursive frontier is tiny.
+			if (stage.is_invariant_build) {
+				auto worker_limit = TaskScheduler::GetScheduler(executor.context).NumberOfThreads();
+				if (GetRecursivePipelineMaxThreads(pipeline, worker_limit) > 1) {
+					auto event = make_shared_ptr<RecursiveCTEPipelineEvent>(pipeline.shared_from_this(), state,
+					                                                        worker_limit, stage.metric_type);
+					event->SchedulePrepared();
```
kryonix: The inline path computes `GetRecursivePipelineMaxThreads()` to decide whether to schedule an Event, but `SchedulePrepared()` immediately recomputes it through `Schedule()`. That repeats the operator traversal and locking. Could we pass the computed worker count into `SchedulePrepared()` and share the task-creation logic with `Schedule()`? The normal Event path would still calculate its worker count after preparing the pipeline.

ted-wq-x:
Thank you for your review. I have made some changes based on your feedback, and I would appreciate another review.

kryonix (approved):
Thanks! LGTM now.

PR #25684, opened 2026-09-14
Optimize RLE index fetches with batch-local decoder reuse

ted-wq-x:
## Summary

RLE index lookups previously restarted decoding from the beginning of the segment for every row. This change reuses decoder progress within each batch, avoiding repeated traversal of the same runs.

The implementation preserves input order, handles duplicate and decreasing offsets, and retains the existing single-row and non-RLE paths. Storage format and transaction semantics are unchanged.

## Reproduction

Create the dataset once and reuse the same database file for both binaries:

```sql
CREATE TABLE edges (
    postid BIGINT USING COMPRESSION rle,
    tagid UBIGINT USING COMPRESSION bitpacking
);

INSERT INTO edges
SELECT i // 4, hash(i) % 2000
FROM range(8000000) t(i);

CREATE INDEX tag_idx ON edges(tagid);
CHECKPOINT;
```

This produces 8 million rows with repeated RLE values and scattered index matches. Compression is explicitly specified to reproduce the relevant MPP storage layout.

Reopen the database read-only and run:

```sql
SET threads = 8;
SET index_scan_max_count = 0;
SET index_scan_percentage = 0.001;

SELECT count(*), sum(postid::HUGEINT)
FROM edges
WHERE tagid = 4;
```

## Performance

Local macOS ARM64 release builds; three warmups followed by seven measured runs:

| Indexed query | Before | After |
|---|---:|---:|
| RLE payload | 84 ms | 4 ms |
| BitPacking-only projection | 2 ms | 2 ms |
| BitPacking payload control | 4 ms | 3 ms |

The RLE query improved approximately **21×**. Results are medians excluding process startup and shutdown.


src/storage/compression/rle.cpp
```diff
@@ -660,17 +660,44 @@ void RLEFetchRow(ColumnSegment &segment, ColumnFetchState &state, row_t row_id,
 	result_data[result_idx] = scan_state.ValidateAndGetCurrentRun().value;
 }
 
+template <class T>
+void RLEFetchRows(ColumnSegment &segment, ColumnFetchState &state, const row_t *row_ids, idx_t fetch_count,
```
Tishj: Can we use `const unsafe_array_ptr<row_t> &row_ids` instead?

src/storage/compression/rle.cpp
```diff
@@ -660,17 +660,44 @@ void RLEFetchRow(ColumnSegment &segment, ColumnFetchState &state, row_t row_id,
...
+	auto &buffer_manager = BufferManager::GetBufferManager(segment.GetDatabase());
+	auto result_data = FlatVector::GetDataMutable<T>(result);
+	idx_t idx = 0;
+	// Reuse the decoder within each nondecreasing sequence. A backwards request starts a fresh decoder,
+	// preserving arbitrary input order without retaining state or pinned blocks beyond this batch.
+	while (idx < fetch_count) {
+		auto handle = buffer_manager.Pin(state.context, segment.GetBlockHandle());
+		RLEScanState<T> scan_state(std::move(handle), segment);
+		idx_t position = 0;
+		do {
```
Tishj: I'm not a huge fan of this, can't we make sure rows are read in order? Or at least handle this outside of each `fetch_rows` implementation, creating separate calls for every group of incrementing row ids?

We might need an API for this that's more like:
`unsafe_array_ptr<row_t> row_ids` alongside a `const vector<vector<idx_t>> result_indexes`, to indicate the multi-index relation between read row_id and the result row(s) that are populated from it.

src/storage/compression/rle.cpp
```diff
@@ -660,17 +660,44 @@ void RLEFetchRow(ColumnSegment &segment, ColumnFetchState &state, row_t row_id,
...
+			const auto target = NumericCast<idx_t>(row_ids[idx]);
+			if (target < position) {
```
Tishj: I was also wondering what happens when the same row is fetched multiple times in a row, `if (target < position)` doesn't hit that case, and we will blindly `Skip` 0 rows, but our `scan_state` has likely already forwarded once, for the row we just read.

Mytherin:
Thanks - could we remove the `fetch_row` implementation? I would prefer to have only one implementation that implements `fetch` versus keeping both.

ted-wq-x:
Thank you for your review. I have made some changes based on your feedback, and I would appreciate another review.

Mytherin:
Thanks - `fetch_row` still exists as a callback, though. Could we fully unify the two so we only have one callback?

ted-wq-x:
Done

ted-wq-x:
It looks like the CI failure wasn't caused by this change.

ted-wq-x:
@Tishj Could you review this pr again?

PR #25685, opened 2026-09-14
Preserve identifiers, CTE separators and VARIANT payloads in generated SQL

kryonix:
SQL printed from an AST needs to retain the meaning of the identifiers and values it contains. There are three independent problems here, kept in separate topic commits.

An identifier beginning with an underscore and a digit can be read as part of a numeric literal after a qualification dot. A reference such as `t."_0_a"` therefore needs its quotes when printed. I apply that rule in the SQL printer and keep the identifier validation used by CSV reject-table options unchanged: `_0_errors` remains a valid option name. `SQLIdentifier` uses the same printing rule as other qualified references.

A printed WITH clause also needs a separator before its main query. I add the missing space there and remove the duplicate space in recursive `UNION ALL`. This affects shared query-node printing, including view definitions and other callers that reparse the resulting SQL.

The final change concerns `ConstantExpression::FromValue`. Its VARIANT conversion extracts a typed payload, then recasts its children into separate VARIANT values before constructing their expressions. For a nested object inside an array, that extra conversion can reorder the object keys. The difference becomes visible through `variant_extract_string`, even though the values compare equal as VARIANTs. A payload such as `{'a': [{'b': 1, 'c': 2, 'a': 3}], 'b': 4, 'c': 5}::VARIANT` can consequently acquire a different string representation after reconstruction.

I retain the typed payload returned by the converter and construct its expression directly before the final VARIANT cast. The expression builder already preserves scalar types recursively, so the intermediate recasting is unnecessary. This keeps the object order while retaining narrow numeric types and heterogeneous array elements.


Mytherin:
Thanks!

PR #25687, opened 2026-09-14
Enable bugprone-exception-escape and run clang-tidy in NightlyTests

smvv:
### Context

> Clang tidy rule `bugprone-exception-escape` helps detect cases where `noexcept` functions can throw (which would lead to `std::terminate` being called and the program being exited). It is currently disabled - we should look into enabling it and fixing the issues it flags.

Fixes https://github.com/duckdblabs/duckdb-internal/issues/8942.

### Why in NightlyTests?

We need to run clang-tidy (not clang**d**-tidy) for `bugprone-exception-escape`.

Clang**d**-tidy does not perform the advanced interprocedural AST checks that clang-tidy does. However, clang-tidy is also too slow to run in each PR.

So, we run clang-tidy in `NightlyTests` and any detected failures become github issues.

### Suppressions

I've created a github [issue](https://github.com/duckdblabs/duckdb-internal/issues/10793) to follow up on any fixing `NOLINT` suppressions, if needed.


Mytherin:
Thanks!

PR #25688, opened 2026-09-14
Keep a block in the eviction queue when its spill to the temporary directory fails

carlopi:
`IterateUnloadableBlocks` dequeues a block before unloading it.
If the unload throws, for example because max_temp_directory_size is reached, the block stayed loaded but had no queue entry any more, so it could not be evicted until it was pinned and unpinned again. With enough failed spills, lowering the memory limit kept failing after the temporary directory had room again. Re-enqueue the node before rethrowing.


Mytherin:
Thanks!

Issue #25623, opened 2026-09-11
Documented syntax highlighting dot commands don't exist in cli


fmoor:
### What happens?

If I use any of these dot commands documented [here](https://duckdb.org/docs/current/clients/cli/syntax_highlighting) then the cli exits on startup with an error.

```
.constant
.cont
.cont_sel
```

related to https://github.com/duckdb/duckdb/issues/25622

### To Reproduce

```
# ~/.duckdbrc
.constant brightwhite
```

```
$ duckdb
-- Loading resources from /home/fmoor/.duckdbrc
WARNING: .constant [COLOR] will be removed in a future release, use .render_color constant brightwhite instead
Unknown element 'constant', supported options: error, keyword, numeric_constant, string_constant, line_indicator, database_name, schema_name, table_name, column_name, column_type, numeric_value, string_value, temporal_value, null_value, footer, layout, startup_text, startup_version, continuation, continuation_selected, bracket, comment, suggestion_catalog_name, suggestion_schema_name, suggestion_table_name, suggestion_column_name, suggestion_file_name, suggestion_directory_name, suggestion_function_name, suggestion_setting_name, table_layout, view_layout, primary_key_column, prompt, error_emphasis, error_suggestion, log_trace, log_debug, log_info, log_warning, none
Encountered errors while executing init file "/home/fmoor/.duckdbrc". Exiting.
```

### OS:

ubuntu 24.04

### DuckDB Version:

v1.5.5 (Variegata) d8cdaa33fd

### DuckDB Client:

CLI

### Full Name:

Frederick Moor

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

Issue #25622, opened 2026-09-11
Warning for old syntax highlighting commands suggests nonexistant .render_color command


fmoor:
### What happens?

Using any of the documented syntax highlighting commands in `.duckdbrc` results in a warning that recomends using `.render_color`, but that command doesn't exist. The correct command is `.highlight_colors`.

### To Reproduce

```
# ~/.duckdbrc
.keyword brightred
```

```bash
$ duckdb
-- Loading resources from /home/fmoor/.duckdbrc
WARNING: .keyword [COLOR] will be removed in a future release, use .render_color keyword brightred instead
DuckDB v1.5.5 (Variegata)
Enter ".help" for usage hints.
memory D
```

### OS:

ubuntu 24.04

### DuckDB Version:

v1.5.5 (Variegata) d8cdaa33fd

### DuckDB Client:

cli

### Full Name:

Frederick Moor

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Not applicable - the reproduction does not require a data set

DinosL:
Hi @fmoor, thanks for opening this series of issues regarding the syntax highlighting commands. Both the documentation and the deprecation warnings were outdated. I have PRs open in both DuckDB and the documentation to address them.

PR #25689, opened 2026-09-14
Fix deprecated CLI highlighting dot commands suggesting .render_color

DinosL:
The old syntax highlighting dot commands `.keyword`, `.comment`, `.error`, `.constant`, `.cont` and `.cont_sel` are deprecated in favor of `.highlight_colors`. Their deprecation warning pointed to `.render_color`, which does not exist, so the suggested replacement didn't work. The warning now points to `.highlight_colors`.

In addition, the `.cont` and `.cont_sel` were renamed to `continuation` and `continuation_selected`, but the deprecated commands still passed the old names to `.highlight_colors`. Both commands now translate to the new names and work again.

Similarly, `.constant` was split into `.numeric_constant` and `.string_constant`. Using `.constant` now shows an error that explains the split.

Fixes: https://github.com/duckdb/duckdb/issues/25622
Fixes: https://github.com/duckdb/duckdb/issues/25623


Mytherin:
Thanks!

PR #25690, opened 2026-09-14
Fix MAX on hive partition column after file pruning (#25629)

### LINKED ISSUE #25629: MAX on a hive partition column crashes when prune leaves one file
### What happens?

`SELECT MAX(hive_col) … WHERE hive_col < date on read_parquet(..., hive_partitioning = true)` raises an INTERNAL Error instead of returning the max remaining partition value.

  Equality filters on the same hive column work. A less selective MAX on the same view can also work; a tighter bound on the same connection then crashes. Seems to happen when selecting a max from just one file?

### To Reproduce

setup:
```sql
COPY (
    SELECT 1 AS a, 0.0 AS b, 100.0 AS c, d AS day
    FROM (VALUES (DATE '2026-01-09'), (DATE '2026-01-10'), (DATE '2026-01-13')) s(d)
  ) TO '/tmp/hive_repro2' (FORMAT PARQUET, PARTITION_BY (day));
  CREATE OR REPLACE VIEW t AS
    SELECT *
    FROM read_parquet('/tmp/hive_repro2/**/*.parquet', hive_partitioning = true);
```
table is shown below. note each row is on a separate file.
```
┌───────┬──────────────┬──────────────┬────────────┐
│   a   │      b       │      c       │    day     │
│ int32 │ decimal(2,1) │ decimal(4,1) │    date    │
├───────┼──────────────┼──────────────┼────────────┤
│     1 │          0.0 │        100.0 │ 2026-01-09 │
│     1 │          0.0 │        100.0 │ 2026-01-10 │
│     1 │          0.0 │        100.0 │ 2026-01-13 │
└───────┴──────────────┴──────────────┴────────────┘
```

the first query works, but the second that only reads one file fails.
```sql
  SELECT MAX(day) FROM t WHERE day < DATE '2026-01-13';
  -- ok, 2026-01-10
  SELECT MAX(day) FROM t WHERE day < DATE '2026-01-10';
  -- INTERNAL Error: Attempted to access index 3 within vector of size 3
```

### OS:

rocky 9

### DuckDB Version:

v2.1.0-alpha40775

### DuckDB Client:

cli

### Hardware:

_No response_

### Full Name:

Amir Vakili

### Affiliation:

Voleon Group

- [x] Yes, I have

- [x] Yes, I have

Yes

avakili-voleon:
Fixes #25629

- Fix a crash/assert when MAX/MIN runs on a hive partition column after file pruning leaves a single parquet file
- Hive partition columns can have storage indices outside the parquet file schema; thus we get rid of the assert the code was making and guard GetColumnStatistics and MinMaxIsExact instead of asserting on bounds (same pattern as [parquet_multi_file_info.cpp](https://github.com/duckdb/duckdb/blob/main/extension/parquet/parquet_multi_file_info.cpp#L395-L397))
- Add a small regression test in parquet_hive.test for MAX on hive date columns after pruning
- fork PR with CI (mostly) green: https://github.com/avakili-voleon/duckdb/pull/2


Mytherin:
Thanks!

PR #25691, opened 2026-09-14
Fix cardinality for descending generate_series

p1p1bear:
Fix the row-count estimate for descending generate_series, which could cause the optimizer to incorrectly remove LIMIT and return too many rows.

It is a tiny fix. :)

New PR  from https://github.com/duckdb/duckdb/pull/25680


Mytherin:
Thanks!

PR #25692, opened 2026-09-14
Add warning for losing precision when casting `DECIMAL` for `VALUES` clause

Dtenwolde:
Fix https://github.com/duckdblabs/duckdb-internal/issues/4694

We now print a warning whenever there is a `values` clause where an implicit cast of a `DECIMAL` value will lower its scale. Such as with the following query: 
```sql
VALUES
      (32942348563242.111222333444555666777888::DECIMAL(38,24)),
      (NULL::NUMERIC);
```

The resulting subtype is `DECIMAL(38,23)`. To preserve both width and scale we would require 39 digits, however this exceeds the maximum width of 38. Therefore an implicit cast to 38 happens. This is now reported in the binder: 
```sql
memory D VALUES (32942348563242.111222333444555666777888::numeric(38,24)), (NULL::numeric);
WARNING:
Potential loss of decimal precision while resolving a VALUES column: type DECIMAL(38,24) is being cast to DECIMAL(38,23), reducing the scale from 24 to 23. This may cause values to be rounded. Explicitly cast all values to a compatible DECIMAL type to avoid this warning.

┌────────────────────────────────────────┐
│                  col0                  │
│             decimal(38,23)             │
├────────────────────────────────────────┤
│ 32942348563242.11122233344455566677789 │
│                                   NULL │
└────────────────────────────────────────┘
```

One problem I see it that this is might happen in lots of other places as well. I'm not sure if there's a general place where to add this check such that we consistenly print a warning with this behaviour

`CASE` is just another example: 
```
memory D   SELECT
               typeof(x),
               x
           FROM (
               SELECT CASE
                   WHEN true
                   THEN 32942348563242.111222333444555666777888::DECIMAL(38,24)
                   ELSE NULL::NUMERIC
               END AS x
           );
┌────────────────┬────────────────────────────────────────┐
│   typeof(x)    │                   x                    │
│    varchar     │             decimal(38,23)             │
├────────────────┼────────────────────────────────────────┤
│ DECIMAL(38,23) │ 32942348563242.11122233344455566677789 │
└────────────────┴────────────────────────────────────────┘
```


Mytherin:
Thanks!

PR #25695, opened 2026-09-14
Check `max_expression_depth` at the boundary between transformer and binder

carlopi:
Bumped in a funky segfault with a SQL file like:
```sql
SELECT abs(abs(abs(..........(1)......)))));
```

there with 10000 of them it would throw on reaching the ``MaximumExpressionDepth`, but say at 20000 it would segfault.
Issue is that `Binder::FindStarExpression` run before checking expression depth, and that itself can run deeper than available resources. First solution was to also carry the check explicitly, now, next iteration: let's just always validate actual expression depth vs setting `MaximumExpressionDepth` at the boundary between PEG's tranformer and Binder.

Now it will always fail due to:
```
Parser Error: Max expression depth limit of 1000 exceeded.
Use "SET max_expression_depth TO x" to increase the maximum expression depth.
```

Note we might want to change error message, to signal it's dangerous to do so.


Dtenwolde:
It's good that you raise this, but I don't think checking post-transform is the way to go. I think it'd be better to fix `CheckStack` and make it work with the iterative parser/transformer. Then we can check during transforming at any point whether we pass the max expression depth.

Right now there's gaps in the post-transform walk that allow for edge cases to still escape the expression depth check.

carlopi:
I would still consider having a check at the boundary, since that would also allow verification / iterations on the Transformer internal checks. On the other side, yet another source of error / more code.

Also, it's not a given to me that measuring recursion depth is the same as tree depth, since created tree can be potentially deeper than the maximal recursion depth.

I think at the end both a bound on Transformer AND a check at the start of the Binder (or, as here, at the boundary) are both handy to have, and I don't see how one replaces the other.

Obv, feel free to consider this more of an issue than a PR.

Issue #23723, opened 2026-07-09
Function throws INTERNAL exception on invalid option


dentiny:
### What happens?

See repro SQL for examples, should be straightforward.

### To Reproduce

```sql
memory D SELECT * FROM repeat_row(1, num_rows=NULL);
INTERNAL Error:
Calling GetValue on a value that is NULL

LINE 1: SELECT * FROM repeat_row(1, num_rows=NULL);
                      ^
memory D SELECT * FROM repeat_row(1, num_rows=-1);
INTERNAL Error:
Information loss on integer cast: value -1 outside of target range [0, 18446744073709551615]

LINE 1: SELECT * FROM repeat_row(1, num_rows=-1);
                      ^
```

### OS:

macos

### DuckDB Version:

985b2f2515adf44ea65dd4239c7bfb30981b94dc

### DuckDB Client:

SQL

### Full Name:

dentiny

### Affiliation:

N/A

### Did you include all relevant configuration (e.g., CPU architecture, Linux distribution) to reproduce the issue?

- [x] Yes, I have

### Did you include all code required to reproduce the issue?

- [x] Yes, I have

### Did you include all relevant data sets for reproducing the issue?

Yes

dentiny:
I remember briefly discussed with the team, that we should have a not-NULL type / value for most, if not all, function options. I will handle it myself.

taniabogatsch:
https://github.com/duckdb/duckdb/pull/25696

Issue #10057, opened 2023-12-21
Referencing a column from your SELECT clause, pointing at a non-existent upstream column, causes the wrong error message


jraede:
### What happens?

Error message is "referenced column not found in FROM clause", but should be "column X does not exist".

### To Reproduce

Run the following in any duckdb (I ran it in motherduck)

```sql
with test_data as (
  select 'foo' as a
)

select test_data.foobar as new_column from test_data where new_column is not null;
```

The trick is defining `new_column` as some sort of reference to a non-existent column in the upstream CTE.

The error message DuckDB returns is "referenced column new_column not found in FROM clause", which is not the case (caused me to pull my hair out wondering about mismatched DuckDB versions and stuff). The actual error is because there is no column named `foobar` in the `test_data` CTE.

### OS:

Linux x64

### DuckDB Version:

0.9.2

### DuckDB Client:

Python

### Full Name:

Jason Raede

### Affiliation:

Pliable

### Have you tried this on the latest `main` branch?

I have tested with a main build

### Have you tried the steps to reproduce? Do they include all relevant data and configuration? Does the issue you report still appear there?

- [X] Yes, I have

PR #25696, opened 2026-09-14
Repeat rows bug

lghavami:
A fix for internal issue [#10057](https://github.com/duckdblabs/duckdb-internal/issues/10057) / public issue [#23723](https://github.com/duckdb/duckdb/issues/23723) 

repeat_row function now has proper exceptions thrown when num_rows is NULL or less than 0. More comprehensive tests for repeat_row as well.


Mytherin:
Thanks!

src/function/table/repeat_row.cpp
taniabogatsch, line 69: I know we already merged the PR, but maybe we could change the type here to `LogicalType::UBIGINT`? That way, we don't have to special-case the negative case in the binding code and can instead catch it in our generic "do the input parameters match the function spec" code?

src/function/table/repeat_row.cpp
```diff
@@ -32,7 +32,14 @@ static unique_ptr<FunctionData> RepeatRowBind(ClientContext &context, TableFunct
 	if (inputs.empty()) {
 		throw BinderException("repeat_rows requires at least one column to be specified");
 	}
-	return make_uniq<RepeatRowFunctionData>(inputs, NumericCast<idx_t>(entry->second.GetValue<int64_t>()));
+	if (entry->second.IsNull()) {
+		throw BinderException("num_rows should be an integer value >= 0");
+	}
+	auto repeat_rows = entry->second.GetValue<int64_t>();
+	if (repeat_rows < 0) {
+		throw BinderException("num_rows cannot be less than zero");
```
taniabogatsch, line 40: We call it `repeat_rows`, but then the parameter is called `num_rows`. That's slightly confusing, maybe we can rename? :)

taniabogatsch (commented):
Hey @lghavami! thanks for the fixes! I took another look at this PR and left some thoughts on how we could slightly improve the fixes. Happy to review a follow-up! 

PS: always nice to see a bunch of test cases!

src/function/table/repeat_row.cpp
taniabogatsch: I also see a commit that changed this to `UBIGINT`, but that did not work?

lghavami: changing to ```UBIGINT``` _did_ work. I think I then tried to apply the same thing to the ```repeat()``` function and the behavior was a bit wonky, and at the time I wanted everything to be uniform. I'm happy to update repeat_rows with this (makes sense for an int that should be >= 0 anyways). In that case I would also want to make sure the behavior from ```repeat()``` is similar.

taniabogatsch: What was wonky with `repeat`? Maybe you can open a draft PR and ping me on it, then we can have a look? :)

lghavami: To answer question, wonky in the sense that it once it was changed to UBIGINT, you suddenly MUST cast the second param in repeat() to a UBIGINT or it throws a binder error:

```sql
from repeat (3, 4);
Binder Error:
No function matches the given name and argument types 'repeat(INTEGER, INTEGER)'. You might need to add explicit type casts.                                                         
        Candidate functions:                                                                                                                                                         
        "repeat"(ANY, UBIGINT) 
```

I'll have a new draft pr open in a bit, I'm going to try to investigate a bit first

taniabogatsch: Sounds good, that feels off. I would've expected us to auto-cast there 🤔 cc @Maxxen

lghavami: Apparently named parameters & positional arguments for table functions go through completely different paths in the binder, so when positional arg is passed, the binder _does_ do some implicit casting, but right now the rules defined in ```src/function/cast_rules.cpp``` don't allow implicit casting from an ```INTEGER``` to a ```UBIGINT```

```cpp
static int64_t ImplicitCastInteger(const LogicalType &to) {
	switch (to.id()) {
	case LogicalTypeId::BIGINT:
	case LogicalTypeId::HUGEINT:
	case LogicalTypeId::FLOAT:
	case LogicalTypeId::DOUBLE:
	case LogicalTypeId::DECIMAL:
	case LogicalTypeId::BIGNUM:
		return TargetTypeCost(to);
	default:
		return -1;
	}
}
```

taniabogatsch: Hmmmm okay, then maybe one more question: how come we are assuming it's an `INTEGER` in the first place? 🕵️ so that we would later need the cast?

lghavami: in ```src/parser/literal.cpp``` the method ```NumberToValue``` is called when the parser encounters a literal that should be interpreted as a numeric value. Once it determines that the input can be cast as some kind of integer (i.e. no decimal point, no exponent), it tries to cast as ```Value::Integer``` first, then ```BIGINT```, ```HUGEINT``` and finally ```UHUGEINT```. So in this case if I call ```repeat(2, 3)```, 3 hits this clause & is cast as an ```INTEGER```:

```cpp
if (try_cast_as_integer) {
		int32_t int_value;
		if (TryCast::Operation<string_t, int32_t>(str_val, int_value)) {
			return Value::INTEGER(int_value);
		}
...
}
```

PR #25697, opened 2026-09-14
Json strip nulls vuln

lghavami:
Another fix for internal issue [#10305](https://github.com/duckdblabs/duckdb-internal/issues/10305)

* transformed recursive ```json_strip_nulls``` to an iterative function + added relevant test
* changed previous recursive -> iterative implementations to use the namespace duckdb ```vector``` instead of ```std::vector```

Future work required to fix recursion in json_deserialize_sql() + create a generic/template solution for all function in [#10305](https://github.com/duckdblabs/duckdb-internal/issues/10305) where recursion has been eliminated.


Mytherin:
Thanks!

PR #25706, opened 2026-09-14
Track I/O Read And Write Time Alongside Byte Counters

Richard-Durkee:
Currently, we emit metrics for bytes read and written, as well as total number of read and write operations. These new metrics, `io.total_read_time` and `io.total_write_time`, are intended to help understand the latency cost of I/O, which may differ from I/O volume. They may also give some indication of whether a query is I/O bound.


Mytherin:
Thanks!

PR #25709, opened 2026-09-15
Fix RLE encoding for floating point values

dentiny:
Hi team, I found in the current implementation, we're [using `==` operator to check equality](https://github.com/duckdb/duckdb/blob/3e0f8b61a349f8d2d20fe95a375b0f94b978dcff/src/storage/compression/rle.cpp#L56), which means +0.0 and -0.0 are treated as the same value.

See the repro SQL how it's introducing bug, case without checkpoint query works fine
```sql
memory D PRAGMA force_compression='rle';
memory D CREATE TABLE z(x DOUBLE);
memory D INSERT INTO z SELECT -0.0::DOUBLE FROM range(50);
memory D INSERT INTO z SELECT  0.0::DOUBLE FROM range(50);
memory D INSERT INTO z VALUES (1.0);
memory D SELECT signbit(x), round(x,1), count(*) FROM z GROUP BY 1, 2 ORDER BY 2, 1;
┌────────────┬─────────────┬──────────────┐
│ signbit(x) │ round(x, 1) │ count_star() │
│  boolean   │   double    │    int64     │
├────────────┼─────────────┼──────────────┤
│ false      │         0.0 │           50 │
│ true       │         0.0 │           50 │
│ false      │         1.0 │            1 │
└────────────┴─────────────┴──────────────┘
```
With checkpoint and RLE encoding involved, we get the incorrect query result
```sql
memory D PRAGMA force_compression='rle';
memory D CREATE TABLE z(x DOUBLE);
memory D INSERT INTO z SELECT -0.0::DOUBLE FROM range(50);
memory D INSERT INTO z SELECT  0.0::DOUBLE FROM range(50);
memory D INSERT INTO z VALUES (1.0);
memory D CHECKPOINT;
memory D SELECT signbit(x), round(x,1), count(*) FROM z GROUP BY 1, 2 ORDER BY 2, 1;
┌────────────┬─────────────┬──────────────┐
│ signbit(x) │ round(x, 1) │ count_star() │
│  boolean   │   double    │    int64     │
├────────────┼─────────────┼──────────────┤
│ true       │         0.0 │          100 │
│ false      │         1.0 │            1 │
└────────────┴─────────────┴──────────────┘
```
So in this PR, I propose to do bit-wise comparison.

Why it's not a problem for other compression algorithms?
- uncompressed: values are written unchanged
- ALP: -0.0 is handled [separately](https://github.com/duckdb/duckdb/blob/3e0f8b61a349f8d2d20fe95a375b0f94b978dcff/src/include/duckdb/storage/compression/alp/algorithm/alp.hpp#L114), and decided impossible to encode
- Chimp / Patas: floating points are converted to integer first before encoding, so +0.0 and -0.0 are different before encoding
- ALPRD: similar to chimp, values are loaded into integer before encoding


src/include/duckdb/storage/statistics/numeric_stats.hpp
```diff
@@ -87,6 +90,15 @@ struct NumericStats {
 	static inline void UpdateValue(T new_value, T &min, T &max) {
 		min = LessThan::Operation(new_value, min) ? new_value : min;
 		max = GreaterThan::Operation(new_value, max) ? new_value : max;
+		if constexpr (std::is_floating_point<T>::value) {
+			// IEEE == treats +0 and -0 as equal; keep both extrema so CONSTANT is not selected.
+			if (new_value == T(0) && min == T(0) && std::signbit(new_value)) {
```
Mytherin: I'm a bit worried about this causing performance issues - we have previously not differentiated between `-0.0` and `0.0` in stats, is this required?

dentiny: Yeah I don't like the change here as well, but I do think it's a bug and didn't think out a smaller and cleaner change yesterday, repro see
```sql
haojiang@HaoJiangs-MacBook-Pro$ ./build/reldebug/duckdb /tmp/const_stats.db
DuckDB v2.0.0-dev10878 (Development Version, a2ed1ea619)
const_stats D CREATE TABLE mixed AS SELECT * FROM (VALUES (-0e0::DOUBLE), (0e0::DOUBLE)) tbl(x);
const_stats D CHECKPOINT;
┌─────────┐
│ Success │
│ boolean │
└─────────┘
  0 rows 
const_stats D SELECT signbit(x) FROM mixed ORDER BY signbit(x);
┌────────────┐
│ signbit(x) │
│  boolean   │
├────────────┤
│ true       │
│ true       │
└────────────┘
const_stats D SELECT column_name, segment_type, compression, stats, persistent, block_id
              FROM pragma_storage_info('mixed');
┌─────────────┬──────────────┬─────────────┬───────────────────────────────────────────────────┬────────────┬──────────┐
│ column_name │ segment_type │ compression │                       stats                       │ persistent │ block_id │
│   varchar   │   varchar    │   varchar   │                      variant                      │  boolean   │  int64   │
├─────────────┼──────────────┼─────────────┼───────────────────────────────────────────────────┼────────────┼──────────┤
│ x           │ DOUBLE       │ Constant    │ {                                                 │ true       │       -1 │
│             │              │             │   'has_no_null': true,                            │            │          │
│             │              │             │   'has_null': false,                              │            │          │
│             │              │             │   'max': -0.0,                                    │            │          │
│             │              │             │   'min': -0.0                                     │            │          │
│             │              │             │ }                                                 │            │          │
├─────────────┼──────────────┼─────────────┼───────────────────────────────────────────────────┼────────────┼──────────┤
│ x           │ VALIDITY     │ Constant    │ {'has_no_null': true, 'has_null': false}          │ true       │       -1 │
└─────────────┴──────────────┴─────────────┴───────────────────────────────────────────────────┴────────────┴──────────┘
```

dentiny: This PR actually contains two fixes for two different encodings, one for RLE, one for constant; the RLE one is much cleaner. Updated the PR to only RLE fix only, will pickup constant one later.

Mytherin:
Thanks!

PR #25711, opened 2026-09-15
Fix index block data loss on `VACUUM`

dentiny:
Hi team, I found the following query returning incorrect result after `VACUUM`.
```sql
memory D SET threads = 1;
memory D SET checkpoint_threshold = '100GB';
memory D ATTACH '/tmp/nested_vacuum_another.db' AS leaf_db (BLOCK_SIZE 16384);
memory D CREATE TABLE leaf_db.t AS SELECT i // 256 AS k, i AS id FROM range(338944) r(i);
memory D CREATE INDEX leaf_idx ON leaf_db.t(k);
memory D CHECKPOINT leaf_db;
memory D DETACH leaf_db;

-- reattach
memory D ATTACH '/tmp/nested_vacuum_another.db' AS leaf_db;
memory D SELECT count(*) FROM leaf_db.t WHERE k = 1250;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│          256 │
└──────────────┘
memory D DELETE FROM leaf_db.t WHERE k >= 824 AND k < 1224;
memory D VACUUM leaf_db.t;
memory D CHECKPOINT leaf_db;
memory D DETACH leaf_db;

-- reattach
memory D ATTACH '/tmp/nested_vacuum_another.db' AS leaf_db;

-- should be 256
memory D SELECT count(*) FROM leaf_db.t WHERE k = 816;
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│           28 │
└──────────────┘
```
I think the reason is, when we vacuum index blocks, target block is not marked as dirty, so on checkpoint it's not persisted.


artjomPlaunov (approved):
Hi @dentiny, good catch! LGTM.

PR #25714, opened 2026-09-15
Fix silent truncation of very long integer literals into HUGEINT

DinosL:
Fixes: https://github.com/duckdblabs/duckdb-internal/issues/10790


Mytherin:
Thanks!

PR #25716, opened 2026-09-15
string_t: fast path for inlined strings in GreaterThan

carlopi:
GreaterThan compared the 4-byte prefix and then fell back to memcmp over the shared length, re-reading the prefix and, for inlined strings, going through a pointer to bytes that sit in the string_t itself.

Now one 8-byte load per string_t yields both length and prefix. When the prefixes match and the shorter string is at most 4 bytes, the length decides. Otherwise the second 8 bytes of both are loaded once: for two inlined strings they are compared as one byte-swapped integer (the inline area is zero-padded past the length, so this is exact with the length as tie-breaker); for anything longer memcmp starts after the prefix. The debug no-inline build keeps the previous code.

Measured on 400M comparisons per shape (greatest() over a cross product): inlined strings sharing a prefix -36%, distinct-prefix short strings -7% to -10%, 16-byte pointer strings level to -4%, a 50/50 inlined/pointer mix +2% to +3% from the extra branch. Ordering verified against byte order on random strings with embedded zeros and high bytes.


carlopi:
This one I am not really sure, only proper benchmarking can distinguish between irrelevant and OK here.
Also most likely compiler dependent, seems to work better on MacOS (clang) than CI (gcc?)
Unsure if worth having then, if this requires more eyes to decide.

PR #25720, opened 2026-09-15
Retain SQL invocation information through binding and optimization

kryonix:
The logical plan needs to retain enough information to reconstruct its SQL semantics. Binding and optimization currently consume some of that information: a catalog-qualified call becomes a specialized function implementation, an argument moves into bind data, or an OFFSET is partly satisfied by row-group pruning. Recovering the original invocation later by recognizing the resulting operator shape is fragile.

I retain that information at the point where it is known and carry it through the copies, serialization and rewrites that maintain the plan. This builds on the function catalog qualification work. Function identity includes the catalog and schema needed to resolve the call again, while retained invocation data describes arguments that binding has specialized or absorbed. Examples include the list sort key, `struct_insert` argument names and the name observed by `alias`.

The optimizer annotations have the same purpose. Original window RANGE inputs and the unpruned LIMIT/TopN offset preserve user-visible boundaries after optimization rewrites them. Compression-origin annotations distinguish representation-only expressions created by compressed materialization from ordinary user expressions. File predicates consumed by pruning remain available so SQL reconstruction does not lose a filter merely because it has already restricted the scan's file set.

Secure views need both their source identity and a stable correspondence between visible columns and retained source expressions. Caller predicates pushed through the secure-view boundary are retained over those source positions before child bindings are rewritten. The native pushdown and barrier behavior remains in place. The SQL consumer can then refer to the named view and its caller predicates without exposing the view's private child plan. Missing correspondence remains explicit instead of being reconstructed from display strings.

These fields describe SQL semantics and provenance. They do not serialize a frozen catalog, inferred file list or runtime observation. A reconstructed invocation can consult those inputs again when it is rebound. Added optional serialization properties have defaults, and copies retain the annotations needed by their consumers.


src/include/duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp
```diff
@@ -24,7 +24,13 @@ class AggregateFunctionCatalogEntry : public FunctionEntry {
...
 	AggregateFunctionCatalogEntry(Catalog &catalog, SchemaCatalogEntry &schema, CreateAggregateFunctionInfo &info);
 
+	bool IsRegisteredFunction(const shared_ptr<const AggregateFunction> &function) const;
+
 	//! The aggregate functions
 	AggregateFunctionSet functions;
+
+private:
+	//! The aggregate functions installed when the catalog entry was created
+	vector<shared_ptr<const AggregateFunction>> registered_functions;
```
Maxxen, line 34: Why is this necessary when we have the `AggregateFunctionSet` already?

src/include/duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp
```diff
@@ -24,7 +24,13 @@ class AggregateFunctionCatalogEntry : public FunctionEntry {
 public:
 	AggregateFunctionCatalogEntry(Catalog &catalog, SchemaCatalogEntry &schema, CreateAggregateFunctionInfo &info);
 
+	bool IsRegisteredFunction(const shared_ptr<const AggregateFunction> &function) const;
```
Maxxen, line 27: Is this actually used anywhere?

src/include/duckdb/function/scalar/generic_common.hpp
```diff
@@ -17,6 +17,22 @@
 namespace duckdb {
 class BoundFunctionExpression;
 
+struct AliasBindData final : public FunctionData {
+	explicit AliasBindData(Identifier alias_p) : FunctionData(InternalKind::ALIAS), alias(std::move(alias_p)) {
```
Maxxen, line 21: I don't really see why FunctionData has to have a `InternalKind`?

src/include/duckdb/function/function_serialization.hpp
```diff
@@ -20,6 +20,20 @@ class ClientContext;
...
+	static void SerializeSchemaPath(Serializer &serializer, const TableFunction &function) {
+		auto path = function.GetQualifiedName().Path();
+		if (path.size() > 3) {
+			path.pop_back();
+		} else {
+			path.clear();
+		}
+		serializer.WritePropertyWithDefault(507, "schema_path", path, vector<Identifier>());
+	}
+
```
Maxxen, line 36: We should just change functions to have a QualifiedName and serialize the qualified name for all functions types - e.g. replace the:

```c++
		serializer.WriteProperty(500, "name", function.GetName());
	    ...
		serializer.WritePropertyWithDefault<Identifier>(505, "catalog_name", function.GetCatalogName(), Identifier());
		serializer.WritePropertyWithDefault<Identifier>(506, "schema_name", function.GetSchemaName(), Identifier());
```

With a

```c++
        ...
	    serializer.WriteProperty(507, "qname", function.GetQualifiedName()
```

src/include/duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp
kryonix: Not in this preparatory PR. The follow-up uses the aggregate version to check that PIVOT's internal list aggregate still resolves to an original registered definition before reconstructing it. The scalar equivalent currently has no caller. I could move it to the next PR.

src/include/duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp
kryonix: I use this to distinguish the definitions installed when the catalog entry was created from later replacements. `AggregateFunctionSet::ApplyToFunctions()` replaces each definition with a modified copy, so the live set no longer retains that identity. The snapshot shares the original definitions rather than copying the function objects. This, too, is relevant in the follow up when validating PIVOT's internal list aggregate, where recognizing the name and signature alone is insufficient.

src/include/duckdb/function/function_serialization.hpp
kryonix: Good call! A single `QualifiedName` across function types is cleaner than adding a separate schema-path property for table functions. I will consolidate the representation and serialization, retaining the legacy fields when targeting older versions and reconstructing the qualified name when reading older serialized plans.

src/include/duckdb/function/scalar_function.hpp
```diff
@@ -165,6 +165,9 @@ typedef unique_ptr<Expression> (*function_bind_expression_t)(FunctionBindExpress
 //! Convert a scalar function to string
 typedef string (*function_to_string_t)(FunctionToStringInput &input);
 
+//! Get the SQL argument names of a bound scalar function
+typedef vector<Identifier> (*scalar_function_argument_names_t)(const BoundScalarFunction &function);
```
Maxxen, line 169: I dont think this is necessary either - named arguments are reordered into positional arguments during binding. The only functions where the argument names matter will store it into their bind data anyway.

If we really want to have the argument names saved so that we can recreate the SQL, we should just do that - store them separately next to the argument expressions themselves. This is what we do in the parsed `FunctionExpression`, and we could do something similar here too. (maybe not introduce a `BoundFunctionArgument`, but just have separate:

```c++
class BoundFunctionExpression { 
  BoundScalarFunction function;

  vector<unique_ptr<Expression>> argument_exprs;
  vector<Identifier> argument_names;
}
```

But again, the fundamental problem is that we have functions that alter their input expressions at bind-time. I think that is the real problem we will have to solve before we can go further with this sql-reexport without creating a million special cases and introducing a ton of complexity.

src/include/duckdb/function/scalar/generic_common.hpp
kryonix: Mh yeah, fair point. I introduced this as a defensive check for the SQL exporter, but it adds a parallel type system to `FunctionData`. I will remove it and keep bind-data interpretation in function-owned helpers, relying on the existing binder/deserializer contract.

src/include/duckdb/function/scalar_function.hpp
Maxxen: Ok, keeping the argument vectors in sync across optimizations may be difficult as well, depending on what kind of optimizations we do, but at that point, why not just have a `typedef unique_ptr<ParsedExpression> scalar_function_unbind_t(const BoundScalarFunction &function)` that just gives you the ParsedExpression immediately?, or even a `string scalar_function_get_sql_t(const BoundScalarFunction &function)`

src/include/duckdb/planner/expression/bound_window_expression.hpp
```diff
@@ -189,6 +200,11 @@ class BoundWindowExpression : public Expression {
 	unique_ptr<Expression> start_expr;
 	unique_ptr<Expression> end_expr;
 
+	//! Literal SQL offsets before endpoint arithmetic; these are not execution children.
+	unique_ptr<Expression> sql_range_start;
```
Maxxen, line 204: So these should be ParsedExpressions then?

src/include/duckdb/planner/sql_export_helpers.hpp
```diff
@@ -53,5 +53,24 @@ inline bool IsSQLRepresentableType(const LogicalType &type) {
 	       !TypeVisitor::Contains(type, [](const LogicalType &child) { return !IsSQLExportType(child.id()); });
 }
 
+inline bool SQLTypesMatch(const LogicalType &left, const LogicalType &right) {
```
Maxxen, line 56: Doesn't the `LogicalType::operator==` already handle this?

src/include/duckdb/storage/serialization/expression.json
```diff
@@ -194,7 +194,8 @@
       {
         "id": 200,
         "name": "value",
-        "type": "Value"
+        "type": "Value",
+        "serialize_property": "value.WithType(return_type)"
```
Maxxen, line 198: What triggered to make this necessary?

src/include/duckdb/planner/expression/bound_window_expression.hpp
kryonix: Yes. These are retained SQL offsets, not execution expressions, so ParsedExpression is a better representation.

src/include/duckdb/function/scalar_function.hpp
kryonix: Yes, this fits the implemented follow-up better than maintaining parallel argument vectors. The exporter already reconstructs children using the current column-binding context, so it could pass those parsed children, the bound function and its bind data to an unbind callback. The function could then construct its own `ParsedExpression`, while qualification checks and result-type preservation remain in the exporter. I prefer that over returning SQL text, so composition and quoting continue through the existing AST and printer.

src/optimizer/expression_rewriter.cpp
```diff
@@ -25,13 +25,18 @@ static unique_ptr<Expression> ApplyRule(LogicalOperator &op, const vector<refere
...
 			auto result = rule.get().Apply(op, bindings, rule_made_change, is_root);
 			if (result) {
 				changes_made = true;
 				// the base node changed: the rule applied changes
 				if (!alias.empty()) {
 					result->SetAlias(std::move(alias));
 				}
+				if (result->GetReturnType() == return_type) {
+					// Type equality ignores annotations such as collation.
+					result->SetReturnType(std::move(return_type));
```
Maxxen: This seems like a bigger (separate) problem. Can we open an issue?

src/include/duckdb/planner/sql_export_helpers.hpp
kryonix: Not for collations: `StringTypeInfo::EqualsInternal()` explicitly ignores them. `VARCHAR` and `VARCHAR COLLATE nocase` therefore compare equal, including when nested inside lists or structs. This helper first checks ordinary type equality, then compares the nested string collations. The exporter needs that stronger comparison to avoid treating expressions with different SQL comparison semantics as interchangeable.

src/include/duckdb/storage/serialization/expression.json
kryonix: A constant such as `'A' COLLATE nocase` exposes the problem. The COLLATE binder updates the expression's return type, but its stored `Value` still has the original uncollated type. Previously, serialization wrote only that value, and deserialization constructed the constant's return type from it, losing the collation.

src/optimizer/statistics_propagator.cpp
```diff
@@ -161,7 +161,12 @@ unique_ptr<BaseStatistics> StatisticsPropagator::PropagateExpression(Expression
 }
 
 unique_ptr<BaseStatistics> StatisticsPropagator::PropagateExpression(unique_ptr<Expression> &expr) {
+	auto return_type = expr->GetReturnType();
 	auto stats = PropagateExpression(*expr, expr);
+	if (expr->GetReturnType() == return_type) {
+		// Replacing an expression must preserve annotations such as collation.
+		expr->SetReturnType(std::move(return_type));
+	}
```
Maxxen: Again, this seems like a bandaid for a bigger problem.

src/planner/expression/bound_aggregate_expression.cpp
```diff
@@ -159,6 +159,8 @@ unique_ptr<Expression> BoundAggregateExpression::Deserialize(Deserializer &deser
 		// return type mismatch - push a cast
 		auto &context = deserializer.Get<ClientContext &>();
 		return BoundCastExpression::AddCastToType(context, std::move(result), return_type);
+	} else {
+		result->SetReturnType(std::move(return_type));
```
Maxxen: Same here

src/optimizer/expression_rewriter.cpp
kryonix: Agreed. I will try to extract the annotation-preservation issue into a separate PR with the reduced reproducers and a focused fix. That keeps this PR about retaining reconstruction metadata rather than mixing in broader optimizer correctness changes.

src/planner/expression/bound_function_expression.cpp
```diff
@@ -164,6 +181,8 @@ void BoundFunctionExpression::Serialize(Serializer &serializer) const {
 	serializer.WriteProperty(201, "children", children);
 	FunctionSerializer::Serialize(serializer, function, bind_info.get());
 	serializer.WriteProperty(202, "is_operator", is_operator);
+	serializer.WritePropertyWithDefault(204, "compression_origin", compression_origin,
+	                                    CompressedMaterializationOrigin::NONE);
```
Maxxen: This seems like it should not be something the FunctionSerializer should be concerned about

src/optimizer/statistics_propagator.cpp
kryonix: Same problem, yes.

src/planner/expression/bound_function_expression.cpp
```diff
@@ -150,11 +151,27 @@ void BoundFunctionExpression::Verify() const {
...
 void BoundFunctionExpression::Serialize(Serializer &serializer) const {
 	if (!serializer.ShouldSerialize(StorageVersion::V2_0_0) && function.HasLegacySerializeCallback()) {
 		// serialize legacy expression for backwards compatibility
 		FunctionToStringInput input(function, bind_info.get(), children);
 		auto legacy_expr = function.GetLegacySerializeCallback()(input);
+		legacy_expr->SetReturnType(return_type);
+		legacy_expr->SetAlias(alias);
+		legacy_expr->SetQueryLocation(query_location);
 		legacy_expr->Serialize(serializer);
 		return;
```
Maxxen, line 174: AFAIK, we never serialize plans back to an older version anyway.

src/planner/expression/bound_function_expression.cpp
kryonix: This is expression-level optimizer metadata, rather than function identity or bind data. The property is written by `BoundFunctionExpression::Serialize`, outside `FunctionSerializer`. It needs to survive serialization because `LogicalOperator::Copy()` uses that path. The follow-up uses it to distinguish compression-generated expressions from ordinary calls, especially casts. Would you prefer this to live in a more general expression-annotation mechanism?

src/include/duckdb/planner/expression/bound_function_expression.hpp
```diff
@@ -64,6 +67,8 @@ class BoundFunctionExpression : public Expression {
 
 	void Serialize(Serializer &serializer) const override;
 	static unique_ptr<Expression> Deserialize(Deserializer &deserializer);
+	//! Serialize without bind data for legacy readers that rebind the function.
+	static unique_ptr<Expression> SerializeAsLegacyRebind(FunctionToStringInput &input);
```
Maxxen: This seems backwards to me legacy readers _do not_ always rebind the function, in fact most of serialization headaches come from those that don't. If anything we should change how we serialize so that we _always_ rebind.

src/planner/expression/bound_function_expression.cpp
kryonix: You're right about the normal plan-copy path: it explicitly uses the latest format. I added the fallback after testing an older serialization target, but the callers I found that select older targets are verification paths, not a supported cross-version plan-export workflow. I will remove it.

src/planner/expression/bound_function_expression.cpp
```diff
@@ -150,11 +151,27 @@ void BoundFunctionExpression::Verify() const {
...
+	auto function = input.bound_function;
+	function.SetSerializeCallback(nullptr);
+	function.SetDeserializeCallback(nullptr);
+	function.SetLegacySerializeCallback(nullptr);
+	vector<unique_ptr<Expression>> children;
+	for (auto &child : input.children) {
+		children.push_back(child->Copy());
+	}
+	auto bind_data = input.bind_data ? input.bind_data->Copy() : nullptr;
+	return make_uniq<BoundFunctionExpression>(std::move(function), std::move(children), std::move(bind_data));
```
Maxxen: Doing all these copies of child expressions, which in turn can also be "SerializeAsLegacyRebind`'d will have quadratic complexity, which is problematic for deep expression trees right?

src/planner/expression/bound_function_expression.cpp
```diff
@@ -202,29 +221,34 @@ unique_ptr<Expression> BoundFunctionExpression::Deserialize(Deserializer &deseri
 	    deserializer, CatalogType::SCALAR_FUNCTION_ENTRY, children, return_type);
 
 	auto is_operator = deserializer.ReadProperty<bool>(202, "is_operator");
+	auto compression_origin = deserializer.ReadPropertyWithExplicitDefault<CompressedMaterializationOrigin>(
+	    204, "compression_origin", CompressedMaterializationOrigin::NONE);
```
Maxxen: Again, id like this to leak into the general expression tree given that it's a relatively self-contained and specialized optimizer pass.

src/include/duckdb/planner/expression/bound_function_expression.hpp
kryonix: Yeah, the comment is too broad. This helper forces the no-bind-data path for these particular functions. It does not describe legacy deserialization generally. I will remove it with the older-target fallback.

Always rebinding would be a cleaner direction, but we need to retain the complete binding inputs first. For example, list sorting resolves ordering and collation during binding, so rebinding against changed session defaults can alter the result. That ties into the invocation-preservation work discussed above.

src/planner/expression/bound_window_expression.cpp
```diff
@@ -195,6 +202,9 @@ unique_ptr<Expression> BoundWindowExpression::Copy() const {
 	new_window->exclude_clause = exclude_clause;
 	new_window->start_expr = start_expr ? start_expr->Copy() : nullptr;
 	new_window->end_expr = end_expr ? end_expr->Copy() : nullptr;
+	new_window->sql_range_start = sql_range_start ? sql_range_start->Copy() : nullptr;
+	new_window->sql_range_end = sql_range_end ? sql_range_end->Copy() : nullptr;
```
Maxxen, line 212: Maybe this is a nitpick, but it does seem like we would like to avoid copying here. Start and end expressions are probably not that complex - but maybe Its easier if we just preserve the parsed expressions during binding?

src/planner/expression/bound_function_expression.cpp
kryonix: Good catch. I will revisit this.

src/planner/expression/bound_window_expression.cpp
```diff
@@ -318,6 +332,11 @@ unique_ptr<Expression> BoundWindowExpression::Deserialize(Deserializer &deserial
 	deserializer.ReadProperty(212, "exclude_clause", result->exclude_clause);
 	deserializer.ReadProperty(213, "distinct", result->distinct);
 	deserializer.ReadPropertyWithExplicitDefault(214, "arg_orders", result->arg_orders, vector<BoundOrderByNode>());
+	deserializer.ReadPropertyWithExplicitDefault(215, "sql_range_start", result->sql_range_start,
+	                                             unique_ptr<Expression>());
+	deserializer.ReadPropertyWithExplicitDefault(216, "sql_range_end", result->sql_range_end, unique_ptr<Expression>());
```
Maxxen: Like argument expressions, if we just serialize the "original", we can re-derive the one we use for execution during deserialization (by rebinding) instead of serializing both?

src/planner/operator/logical_get.cpp
```diff
@@ -328,6 +328,7 @@ void LogicalGet::Serialize(Serializer &serializer) const {
 	serializer.WritePropertyWithDefault<unique_ptr<RowGroupOrderOptions>>(214, "row_group_order_options",
 	                                                                      row_group_order_options);
 	serializer.WritePropertyWithDefault(215, "scan_partition_indices", scan_partition_indices, vector<idx_t>());
+	serializer.WritePropertyWithDefault<unique_ptr<TableRef>>(216, "table_function_ref", table_function_ref);
```
Maxxen: I suppose we can store the table ref, but couldn't we also have a callback here instead, e.g. `table_function_to_sql()` that can be opt-in? I presume most table functions know how to represent themselves as SQL just given their bind data.

src/planner/expression/bound_window_expression.cpp
kryonix: That would avoid serializing both representations. We would need to factor the RANGE binding logic so deserialization can rebuild the endpoint arithmetic and ordering coercions from the retained frame inputs. One constraint from the implemented follow-up is that optimizer changes must remain authoritative: serializing the original syntax must not silently undo a changed endpoint. I will check that contract as part of the refactor rather than simply replacing the serialized execution expressions with the original AST.

src/storage/serialization/serialize_logical_operator.cpp
```diff
@@ -673,12 +673,14 @@ void LogicalLimit::Serialize(Serializer &serializer) const {
...
 	serializer.WriteProperty<BoundLimitNode>(200, "limit_val", limit_val);
 	serializer.WriteProperty<BoundLimitNode>(201, "offset_val", offset_val);
+	serializer.WritePropertyWithDefault<optional_idx>(202, "unpruned_offset", unpruned_offset, optional_idx());
 }
 
 unique_ptr<LogicalOperator> LogicalLimit::Deserialize(Deserializer &deserializer) {
 	auto limit_val = deserializer.ReadProperty<BoundLimitNode>(200, "limit_val");
 	auto offset_val = deserializer.ReadProperty<BoundLimitNode>(201, "offset_val");
 	auto result = duckdb::unique_ptr<LogicalLimit>(new LogicalLimit(std::move(limit_val), std::move(offset_val)));
+	deserializer.ReadPropertyWithExplicitDefault<optional_idx>(202, "unpruned_offset", result->unpruned_offset, optional_idx());
```
Maxxen, line 683: In general, do these offsets have to be constants or can they be expressions? Like, what is the desired semantics if I do e.g.

```
SET VARIABLE foobar = 2
SELECT * FROM tbl LIMIT $foobar
```

Do we re-export the SQL as bound (with foobar resolved to 2), or unbound?

src/planner/operator/logical_get.cpp
kryonix: The follow-up already has a `to_sql` callback returning a table-reference AST. The retained `TableRef` supports its generic fallback, rather than being required by every callback. I will revisit whether that fallback should require explicit opt-in and whether each supported function can reconstruct from its bound inputs instead. Bind data alone won't necessarily retain the original argument expressions.

src/storage/serialization/serialize_logical_operator.cpp
kryonix: LIMIT and OFFSET can be expressions. This field only records a constant offset that row-group pruning has partly consumed. Expression-valued limits and offsets remain in `BoundLimitNode`.

For a value already resolved to `2` in the bound plan, the intended export is `2`, rather than recovering the original parameter spelling.

src/include/duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp
Maxxen: I dont think we ever add functions or call `ApplyToFunctions()` at "runtime", its just during registration that the function set actually gets mutated, so I still don't see why this is necessary?
Functions identity should (is) be by value anyway, so even if the functions are replaced, they should still compare properly.

src/include/duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp
Maxxen: Could you elaborate on the distinction between a "original registered" function and one that is not? In my mind any function that is part of a function-set in the catalog is by definition a registered function.

src/include/duckdb/function/function_serialization.hpp
Maxxen: Awesome!

src/include/duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp
kryonix: Right. I was treating replacement of the shared pointer as a loss of function identity, which is stronger than the contract we need. The snapshot was defending against manually modified definitions rather than a normal execution path. I should be able to remove it and use the existing function identity/comparison mechanisms where the exporter needs to validate a function.

src/include/duckdb/catalog/catalog_entry/aggregate_function_catalog_entry.hpp
kryonix: By _original_ I meant the objects present when the catalog entry was constructed. That was an implementation distinction, not a meaningful registration distinction. Functions subsequently present in the catalog set are registered too. `IsRegisteredFunction` is therefore misleading, and I will remove both the method and the snapshot.

src/planner/operator/logical_get.cpp
Maxxen: ah cool, I suppose you only have to actually store it if the `to_sql` is not set then - or always store it, I guess it doesn't really matter that much.

src/storage/serialization/serialize_logical_operator.cpp
Teggy: Sorry to jump in here, but would this be the time to also consider the use of expressions in the `SAMPLE` clause?  Currently its argument is forced to be a constant.  I have had countless cases where I wanted to write

~~~
SET VARIABLE n = 100
FROM t SAMPLE $n;
~~~

but could not.

kryonix: Hi @Teggy! Yes, sounds good to me, I'd like that too 👍🏻 I think it deserves a separate PR, though. SAMPLE currently rejects non-constant arguments during parsing, so we'd need to carry the expression into binding and evaluate and validate it there. It's related to what I'm doing here, but it adds functionality beyond retaining the information needed for SQL reconstruction (this PR really is the basis for a proper `EXPLAIN (SQL)`, but 🤫).

Teggy: Thanks, reasonable. I move that request to its own PR. 👍🏻

P.S. `EXPLAIN (SQL)`: want! 😉

kryonix: _Soon™_ 😉

src/function/scalar/compressed_materialization_utils.cpp
```diff
@@ -1,7 +1,47 @@
...
+	function.SetDefinition(std::move(definition));
+}
+
+CMExpressionType CMUtils::GetExpressionType(const BoundFunctionExpression &expression) {
+	auto &function = expression.Function();
+	if (function.GetBindCallback() != Bind) {
+		return CMExpressionType::NONE;
+	}
+	if (function.GetName() == InternalCompressedMaterializationCastFun::Name &&
+	    function.GetDeserializeCallback() == CastFun::GetFunction().GetDeserializeCallback()) {
```
Maxxen, line 32: This doesnt always work as function pointer equality is not preserved across extensions... But I guess extensions dont call this code anyway so I guess it's fine? Just something to keep in mind for the future.

kryonix: Good point. I could replace the callback-address comparisons with exact qualified-name checks for these internal functions, and explicitly qualify the functions constructed directly by the optimizer. That would avoid the cross-extension address issue without adding another tag or serialization field.

It looks like a small change across the compression helpers. The assumption would then be that these reserved names identify the internal operations.

PR #25721, opened 2026-09-15
Utf8Proc::Analyze: check 32 bytes per iteration while the input is ASCII

carlopi:
The ASCII scan checked 8 bytes per iteration with a branch each. Folding four loads into one mask check per 32 bytes is neutral on short string, bails out at first problem (so neutral on long one), but consume faster when appending long ASCII-only strings.


Mytherin:
Thanks - good idea

PR #25723, opened 2026-09-15
[async results] ResultFormats 1/3: Introduce a ResultUnit

evertlammerts:
Part 1/3 of result format generalization (part of **async results**)

### Background

By generalizing result formats we're trying to provide a single abstraction over different formats a result can be fetched as. We model this by introducing a `ResultUnit`. The canonical unit is a `DataChunk`, and the materialized collection a `ColumnDataCollection`. Other formats (arrow, quack, ...) define their own `ResultUnit`, and a materialized collection of those units (the `ResultUnitCollection`) is always an ordered list.

A custom format implements a `ResultFormat`, declares its name, its unit type and its global state, and implements `InitGlobal`, `InitLocal`, `Append` and `Finish`. The format defines exactly one unit type, and every accessor and stream constructor compares the settled format's name against the one it was asked for before it casts a unit. The unit itself carries no type tag, so `ResultUnit::Cast<T>()` is the usual `DynamicCastCheck`, the same as for bind data and operator states.

<details>
<summary>
<b>Complete proposed API (across all 3 PRs)</b>
</summary>

```cpp
//! A piece of query output: what the buffer holds and what the consumer pops. Its concrete type is
//! fixed by the format that produced it, which the consumer checks by name before casting
class ResultUnit {
	ResultUnit(idx_t row_count, idx_t byte_size);
	virtual ~ResultUnit();
	template <class TARGET> TARGET &Cast();   // DynamicCastCheck, like FunctionData
	const idx_t row_count;
	const idx_t byte_size;                    // what the cap counts; identical at push and pop
};

//! The chunk format's unit
class ChunkUnit : public ResultUnit {
	explicit ChunkUnit(unique_ptr<DataChunk> chunk);
	unique_ptr<DataChunk> chunk;
};

//! Storage of a materialized result in a format other than chunks: its units, in consumption order
class ResultUnitCollection {
	explicit ResultUnitCollection(vector<unique_ptr<ResultUnit>> units);
	idx_t Count() const;        // rows the collection was built with; fetching does not change it
	idx_t UnitCount() const;    // units it was built with, fetched or not
	unique_ptr<ResultUnit> Fetch();                       // returns a copy of the next unit
	const deque<unique_ptr<ResultUnit>> &Units() const;   // the units not yet fetched
};

class ResultFormatGlobalState { virtual ~...; template <class T> T &Cast(); };
class ResultFormatLocalState  { virtual ~...; template <class T> T &Cast(); };

//! Everything a format's global state is built from. Captured on the client thread at submission
struct ResultFormatContext {
	vector<LogicalType> types;
	vector<Identifier> names;
	ClientProperties client_properties;
	ResultOrdering ordering = ResultOrdering::UNORDERED;
};

class ResultFormat {
	// every subclass declares:  using Unit; using GlobalState; static constexpr const char *NAME;
	virtual ~ResultFormat();
	//! Identifies this format across every accessor and stream check
	virtual const char *Name() const = 0;
	//! Once per query, on the thread that settles the format
	virtual unique_ptr<ResultFormatGlobalState> InitGlobal(const vector<LogicalType> &types,
	                                                       const vector<Identifier> &names,
	                                                       const ClientProperties &properties,
	                                                       ResultOrdering ordering) = 0;
	//! Once per worker thread, at its first Append
	virtual unique_ptr<ResultFormatLocalState> InitLocal(ResultFormatGlobalState &gstate) = 0;
	//! Convert one chunk into the unit under construction. Runs on a worker thread
	virtual void Append(ResultFormatGlobalState &gstate, ResultFormatLocalState &lstate, DataChunk &chunk) = 0;
	//! Hand over the next unit that reached the format's size target, with row_count and byte_size set.
	//! With flush_partial the unit under construction is handed over short of the target: the sink
	//! flushes at a batch boundary and at a producer's end of input, so no unit spans two batch
	//! indexes. Null when there is nothing to hand over
	virtual unique_ptr<ResultUnit> Finish(ResultFormatGlobalState &gstate, ResultFormatLocalState &lstate,
	                                      bool flush_partial) = 0;
	bool IsChunk() const;
	static const shared_ptr<ResultFormat> &Chunk();
};

//! The identity format. NAME = "chunk". Its accessors return DataChunks and its retained store is a
//! ColumnDataCollection, in memory or buffer-managed
class ChunkFormat : public ResultFormat {
	explicit ChunkFormat(QueryResultMemoryType memory_type = QueryResultMemoryType::IN_MEMORY);
	static const shared_ptr<ResultFormat> &InMemory();
	static const shared_ptr<ResultFormat> &BufferManaged();
};

// Choosing a format: at submission, or on the handle before its first consuming call
struct QueryParameters { ...; shared_ptr<ResultFormat> format; };   // null means chunks
class QueryResult {
	void SetFormat(shared_ptr<ResultFormat> format);
	const ResultFormat &Format() const;
	// Every typed accessor throws InvalidInputException when FORMAT is not the settled format
	template <class FORMAT = ChunkFormat> unique_ptr<typename FORMAT::Unit> Fetch();
	template <class FORMAT = ChunkFormat> ResultCollectionOf<FORMAT>::type &Collection();   // CDC for chunks
	template <class FORMAT = ChunkFormat> unique_ptr<ResultCollectionOf<FORMAT>::type> TakeCollection();
	template <class FORMAT> const typename FORMAT::GlobalState &FormatState() const;
};

// Streams: the constructor checks the settled format's name once, every pop casts under that check
class QueryResultStream : public ResultStreamBase {                    // chunk format
	QueryResultState TryFetch(unique_ptr<DataChunk> &out_chunk);
	unique_ptr<DataChunk> Fetch();
};
template <class FORMAT> class FormattedResultStream : public ResultStreamBase {
	unique_ptr<typename FORMAT::Unit> Fetch();
	QueryResultState TryFetch(unique_ptr<typename FORMAT::Unit> &out_unit);
	const typename FORMAT::GlobalState &FormatState() const;
};

// Part 3
class ArrowUnit : public ResultUnit { ArrowArrayWrapper array; };
class ArrowFormat : public ResultFormat {     // NAME = "arrow"; Unit = ArrowUnit
	explicit ArrowFormat(idx_t batch_size);   // rows per record batch, except at a batch boundary and at end of input
};
DUCKDB_V2_ERROR duckdb_v2_result_to_arrow_stream(duckdb_v2_result_handle *result, idx_t batch_size, ...);
```

</details>

### This PR

This PR introduces the `ResultUnit` in the sink and the streaming buffers. Nothing changes for users yet: the only unit is `ChunkUnit`, and `QueryResultStream` pops units and unwraps the chunk. The buffers (`SimpleBufferedData`, `BatchedBufferedData`) hold `unique_ptr<ResultUnit>` instead of chunks and do their byte accounting on `byte_size`. That field and `row_count` are const constructor arguments.


src/include/duckdb/main/result_unit.hpp
```diff
@@ -0,0 +1,70 @@
...
+			throw InternalException("Failed to cast result unit to type - result unit type mismatch");
+		}
+		return reinterpret_cast<const TARGET &>(*this);
+	}
+
+public:
+	//! What this unit holds
+	ResultUnitType type;
+	//! The rows in this unit
+	idx_t row_count = 0;
```
Tishj: Don't you want to only make this part of the interface in the shape of `virtual idx_t RowCount() const = 0;` to allow a result unit implementation to potentially be lazily evaluated? Same goes for the other members. And I'd make `batch_index` a constructor argument, so it can stay `const`

src/include/duckdb/main/result_unit.hpp
```diff
@@ -0,0 +1,70 @@
...
+	//! The bytes this unit counts against the buffer's cap
+	idx_t byte_size = 0;
+	//! INVALID_INDEX unless the sink is batch ordered
+	idx_t batch_index = DConstants::INVALID_INDEX;
```
Tishj: Should we use `optional<idx_t>` instead? Using `INVALID_INDEX` is error-prone, which we deprecated for `optional_idx`, but now that we have `optional` I'd say that's preferred over `optional_idx`

src/include/duckdb/main/result_unit.hpp
```diff
@@ -0,0 +1,70 @@
...
+	const TARGET &Cast() const {
+		if (TARGET::TYPE != type) {
+			throw InternalException("Failed to cast result unit to type - result unit type mismatch");
+		}
+		return reinterpret_cast<const TARGET &>(*this);
+	}
+
+public:
+	//! What this unit holds
+	ResultUnitType type;
```
Tishj: If this is meant to be extendable then using a closed `ResultUnitType` enum is problematic.
I'm not sure what would work here though.

Recently Sander dealt with a similar problem for the transform result, which resulted in https://github.com/duckdb/duckdb/pull/25513/changes#diff-97175fb83509172b9cb90cc9fb755414367780cc3bb3e138bd630e89a543b47b

Maybe something worth copying here?

src/main/query_result_stream.cpp
```diff
@@ -94,7 +95,9 @@ QueryResultState QueryResultStream::TryFetch(unique_ptr<DataChunk> &out_chunk) {
 			return state;
 		}
 		if (state == QueryResultState::READY) {
-			out_chunk = buffer.Scan();
+			if (auto unit = buffer.Scan()) {
+				out_chunk = std::move(unit->Cast<ChunkUnit>().chunk);
```
Tishj, line 99: Can we make this assumption about the unit's type? This results in an `InternalException` if the type isn't `CHUNK` (which for now it can't be anything else, but in the future I imagine it could(?))

evertlammerts: In this PR we can, since it's the only type we have. In the [follow up](https://github.com/duckdb/duckdb/pull/25727) we introduce a `ResultStreamBase` that can be implemented by other formats.

src/include/duckdb/main/result_unit.hpp
evertlammerts: I don't think we can allow lazy row counts or byte sizes, the streaming buffers need it to decide whether or not to admit a unit. These should be consts, I moved them to the ctor.

src/include/duckdb/main/result_unit.hpp
evertlammerts: I've moved the batch index away from the unit back onto the sink. The unit itself doesn't have any use for it.

src/include/duckdb/main/result_unit.hpp
evertlammerts: Agreed, the enum is gone. I first copied Sander's idea (a name per unit type, compared by pointer and then by content), but as it turns out, the unit does not need a name of its own. A format's name is good enough, see #25727.

Tishj (approved):
Thanks, looks good to me!

PR #25724, opened 2026-09-15
Fix missing reset for `OptimisticWriteCollection`

dentiny:
Hi team, the following query throw INTERNAL exception for me.
```sql
haojiang@HaoJiangs-MacBook-Pro$ ./build/reldebug/duckdb /tmp/crash_sep_15.db
DuckDB v2.0.0-dev10748 (Development Version, 263ec1ea1d)
crash_sep_15 D CREATE TABLE t(a INTEGER, b VARCHAR, c DOUBLE);
crash_sep_15 D BEGIN TRANSACTION;
crash_sep_15 D INSERT INTO t SELECT range, 'val_'||range, range*1.5 FROM range(700000);
crash_sep_15 D ALTER TABLE t ALTER COLUMN c TYPE BIGINT;
crash_sep_15 D INSERT INTO t SELECT range+700000, 'new_'||(range+700000), range*2 FROM range(700000);
INTERNAL Error:
Row group prev allocated size is larger than currently allocated size
```
Two key concepts for row group collection:
- `allocation_size`: The cumulative transient memory allocated by append operations for the current collection
- `prev_allocated_size`: A snapshot of that collection’s `allocation_size` at the previous `WriteNewRowGroup` call, used to calculate newly allocated memory since then, which, by definition, is no larger than `allocation_size`

The root cause is
- On table schema change, a new `LocalTableStorage` and `RowGroupCollection` is created fresh; within which [`allocation_size` is re-initialized](https://github.com/duckdb/duckdb/blob/38c5eebcb09d8d315fdd2875e778d710ebd5f9d1/src/storage/table/row_group_collection.cpp#L143)
- In the current implementation, we only assign current `allocation_size`, but [not `prev_alloc_size`](https://github.com/duckdb/duckdb/blob/38c5eebcb09d8d315fdd2875e778d710ebd5f9d1/src/storage/local_storage.cpp#L75-L76), which is inherited from parent row group collection
- So the fix for this issue is to reset everything when new collection created

Why the issue is not surfaced normally? because it's hard to trigger due to a few pre-conditions
- One transaction needs to contain INSERT, ALTER, INSERT in order
- The first INSERT has to exceed row group flush threshold, so `prev_alloc_size` is non-zero
- The second INSERT also needs to exceed flush threshold to trigger the allocation size invariant check


Mytherin:
Thanks!

PR #25726, opened 2026-09-15
[Arrow][Dev] For `arrow_scan`, use `TableFunctionInfo` to carry the `ArrowScanFactory`

Tishj:
This PR fixes https://github.com/duckdblabs/duckdb-internal/issues/10780

Rather than feeding it through the table function input as `POINTER` values, this cleans up the interface quite a bit, removing the safety concerns that using `POINTER` through a public API carries with it.

We add `shared_ptr<TableFunctionInfo>` to:
- `TableFunctionRelation`
- `TableFunctionRef`
- `LogicalGet`
Disabling serialization for these objects if the shared_ptr is non-null.

We might also be able to clean up `TableRef::external_dependency` as a result, but that might require additional work, so is left as a potential follow up

### Breaking Change

This will likely break at least Python, but possibly other clients, since the source of `arrow_scan` are implemented per client. Though since we are migrating most extensions to use the C-API, the affected surface might be minimal.


src/function/table/arrow.cpp
```diff
@@ -59,28 +59,13 @@ unique_ptr<FunctionData> ArrowTableFunction::ArrowScanBindDumb(ClientContext &co
...
-	auto stream_factory_ptr = input.inputs[0].GetPointer();
-	auto stream_factory_produce = (stream_factory_produce_t)input.inputs[1].GetPointer();       // NOLINT
-	auto stream_factory_get_schema = (stream_factory_get_schema_t)input.inputs[2].GetPointer(); // NOLINT
-
-	auto res = make_uniq<ArrowScanFunctionData>(stream_factory_produce, stream_factory_ptr, std::move(dependency));
-
-	auto &data = *res;
-	stream_factory_get_schema(reinterpret_cast<ArrowArrayStream *>(stream_factory_ptr), data.schema_root.arrow_schema);
-	PopulateArrowTableSchema(context, res->arrow_table, data.schema_root.arrow_schema);
+	auto factory = shared_ptr_cast<TableFunctionInfo, ArrowScanFactory>(input.ref.bind_info);
```
evertlammerts, line 66: nit: maybe prepend a `DynamicCastCheck<ArrowScanFactory>(input.ref.bind_info.get());` assertion?

evertlammerts (commented):
Nice, and LGTM. 🤖  found this though:
- `TableFunctionRef::SerializableFunction()` should throw `NotImplementedException`. Scanning an arrow object as `v` in `WITH t AS (SELECT * FROM v) SELECT i FROM t UNION ALL SELECT i FROM t LIMIT 3` will fail.
- Same happens with `debug_verify_serializer` enabled.

Maxxen:
Awesome, as a followup we should remove the LogicalTypeId/LiteralKind::POINTER then too as soon as the python API has adjusted

Tishj:
I believe I checked and we use POINTER in other places internally (aggregates), but maybe we can remove at least the parts that are more public-facing

Maxxen:
Yeah keeping internally make sense, but I had to add it to the literal parse expression cause of the python API, but that should indeed be possible to remove now.

PR #25727, opened 2026-09-15
[async results] ResultFormats 2/3: Introduce ResultFormat

evertlammerts:
Stacked on #25723

Part 2/3 of result format generalization (part of **async results**)

### Background

By generalizing result formats we're trying to provide a single abstraction over different formats a result can be fetched as. We model this by introducing a `ResultUnit`. The canonical unit is a `DataChunk`, and the materialized collection a `ColumnDataCollection`. Other formats (arrow, quack, ...) define their own `ResultUnit`, and a materialized collection of those units (the `ResultUnitCollection`) is always an ordered list.

A custom format implements a `ResultFormat`, declares its name, its unit type and its global state, and implements `InitGlobal`, `InitLocal`, `Append` and `Finish`. The format defines exactly one unit type, and every accessor and stream constructor compares the settled format's name against the one it was asked for before it casts a unit. The unit itself carries no type tag, so `ResultUnit::Cast<T>()` is the usual `DynamicCastCheck`, the same as for bind data and operator states.

<details>
<summary>
<b>Complete proposed API (across all 3 PRs)</b>
</summary>

```cpp
//! A piece of query output: what the buffer holds and what the consumer pops. Its concrete type is
//! fixed by the format that produced it, which the consumer checks by name before casting
class ResultUnit {
	ResultUnit(idx_t row_count, idx_t byte_size);
	virtual ~ResultUnit();
	template <class TARGET> TARGET &Cast();   // DynamicCastCheck, like FunctionData
	const idx_t row_count;
	const idx_t byte_size;                    // what the cap counts; identical at push and pop
};

//! The chunk format's unit
class ChunkUnit : public ResultUnit {
	explicit ChunkUnit(unique_ptr<DataChunk> chunk);
	unique_ptr<DataChunk> chunk;
};

//! Storage of a materialized result in a format other than chunks: its units, in consumption order
class ResultUnitCollection {
	explicit ResultUnitCollection(vector<unique_ptr<ResultUnit>> units);
	idx_t Count() const;        // rows the collection was built with; fetching does not change it
	idx_t UnitCount() const;    // units it was built with, fetched or not
	unique_ptr<ResultUnit> Fetch();                       // returns a copy of the next unit
	const deque<unique_ptr<ResultUnit>> &Units() const;   // the units not yet fetched
};

class ResultFormatGlobalState { virtual ~...; template <class T> T &Cast(); };
class ResultFormatLocalState  { virtual ~...; template <class T> T &Cast(); };

//! Everything a format's global state is built from. Captured on the client thread at submission
struct ResultFormatContext {
	vector<LogicalType> types;
	vector<Identifier> names;
	ClientProperties client_properties;
	ResultOrdering ordering = ResultOrdering::UNORDERED;
};

class ResultFormat {
	// every subclass declares:  using Unit; using GlobalState; static constexpr const char *NAME;
	virtual ~ResultFormat();
	//! Identifies this format across every accessor and stream check
	virtual const char *Name() const = 0;
	//! Once per query, on the thread that settles the format
	virtual unique_ptr<ResultFormatGlobalState> InitGlobal(const vector<LogicalType> &types,
	                                                       const vector<Identifier> &names,
	                                                       const ClientProperties &properties,
	                                                       ResultOrdering ordering) = 0;
	//! Once per worker thread, at its first Append
	virtual unique_ptr<ResultFormatLocalState> InitLocal(ResultFormatGlobalState &gstate) = 0;
	//! Convert one chunk into the unit under construction. Runs on a worker thread
	virtual void Append(ResultFormatGlobalState &gstate, ResultFormatLocalState &lstate, DataChunk &chunk) = 0;
	//! Hand over the next unit that reached the format's size target, with row_count and byte_size set.
	//! With flush_partial the unit under construction is handed over short of the target: the sink
	//! flushes at a batch boundary and at a producer's end of input, so no unit spans two batch
	//! indexes. Null when there is nothing to hand over
	virtual unique_ptr<ResultUnit> Finish(ResultFormatGlobalState &gstate, ResultFormatLocalState &lstate,
	                                      bool flush_partial) = 0;
	bool IsChunk() const;
	static const shared_ptr<ResultFormat> &Chunk();
};

//! The identity format. NAME = "chunk". Its accessors return DataChunks and its retained store is a
//! ColumnDataCollection, in memory or buffer-managed
class ChunkFormat : public ResultFormat {
	explicit ChunkFormat(QueryResultMemoryType memory_type = QueryResultMemoryType::IN_MEMORY);
	static const shared_ptr<ResultFormat> &InMemory();
	static const shared_ptr<ResultFormat> &BufferManaged();
};

// Choosing a format: at submission, or on the handle before its first consuming call
struct QueryParameters { ...; shared_ptr<ResultFormat> format; };   // null means chunks
class QueryResult {
	void SetFormat(shared_ptr<ResultFormat> format);
	const ResultFormat &Format() const;
	// Every typed accessor throws InvalidInputException when FORMAT is not the settled format
	template <class FORMAT = ChunkFormat> unique_ptr<typename FORMAT::Unit> Fetch();
	template <class FORMAT = ChunkFormat> ResultCollectionOf<FORMAT>::type &Collection();   // CDC for chunks
	template <class FORMAT = ChunkFormat> unique_ptr<ResultCollectionOf<FORMAT>::type> TakeCollection();
	template <class FORMAT> const typename FORMAT::GlobalState &FormatState() const;
};

// Streams: the constructor checks the settled format's name once, every pop casts under that check
class QueryResultStream : public ResultStreamBase {                    // chunk format
	QueryResultState TryFetch(unique_ptr<DataChunk> &out_chunk);
	unique_ptr<DataChunk> Fetch();
};
template <class FORMAT> class FormattedResultStream : public ResultStreamBase {
	unique_ptr<typename FORMAT::Unit> Fetch();
	QueryResultState TryFetch(unique_ptr<typename FORMAT::Unit> &out_unit);
	const typename FORMAT::GlobalState &FormatState() const;
};

// Part 3
class ArrowUnit : public ResultUnit { ArrowArrayWrapper array; };
class ArrowFormat : public ResultFormat {     // NAME = "arrow"; Unit = ArrowUnit
	explicit ArrowFormat(idx_t batch_size);   // rows per record batch, except at a batch boundary and at end of input
};
DUCKDB_V2_ERROR duckdb_v2_result_to_arrow_stream(duckdb_v2_result_handle *result, idx_t batch_size, ...);
```

</details>

### This PR

This PR introduces the `ResultFormat` abstraction. See the API in the section above.

The `ResultFormat` contract is in result_format.hpp / .cpp. A format provides:
- `const char *Name() const`: the format's name
- `unique_ptr<ResultFormatGlobalState> InitGlobal(...)`: global state with the `ResultFormatContext` (types, names, client properties, ordering).
- `unique_ptr<ResultFormatLocalState> InitLocal(ResultFormatGlobalState &gstate)`: local state with a single worker's buffer.
- `void Append(ResultFormatGlobalState &gstate, ResultFormatLocalState &lstate, DataChunk &chunk)`: converts the given datachunk into a `ResultUnit` according to the format, and appends it to the local buffer.
- `unique_ptr<ResultUnit> Finish(..., bool flush_partial)`: hands off a finished unit to the global buffer. `flush_partial` is needed to allow flushing units that have a rowcount less than the configured capacity. This is needed when a batch is finished but the unit isn't full yet.

A format gets a `ResultUnitCollection` for retained results. For `DataChunk`s, this is always a ColumnDataCollection. For all other units, it's basically just a vector of `ResultUnit`s

After this we still need a few things to get a fully consistent API:
- Stop letting the planner set retention. Right now the planner assumes we do not want to stream anything other than SELECT's. The reason is not materialization though, it is eagerness. The assumption is that anything that doesn't stream is executed eagerly, but in an async query execution world that is not true. An important change is that we should no longer allow a new result to close the old, but rather error if a result already exists.
- Hide expanding statements. A handful of statements expand to a group of statements during preprocessing. An example is `COPY ... RETURNING ...`. We should provide an abstraction over a `StatementGroup` that hides this fact from a user, so that they get predictable semantics when submitting a statement, regardless of the statement. As a first step we introduce the abstraction and make sure a consumer's poll advances through all fragments.
- As a second step to hiding expanding statements, we should let the engine itself advance through the fragments even without the user interacting. That is the last step we need to provide fully consistent behavior across all statement types.


Maxxen:
Q: How is the `ResultFormat` materially different from the `PhysicalResultCollector`? what is the justification for adding this extra abstraction layer/indirection?

evertlammerts:
> Q: How is the `ResultFormat` materially different from the `PhysicalResultCollector`? what is the justification for adding this extra abstraction layer/indirection?

@Maxxen A collector is a PhysicalOperator that handles buffering, ordering, retention, backpressure and the result object. A custom format is a PhysicalOperator subclass that reimplements all of that which is why arrow needs two classes, PhysicalArrowCollector and PhysicalArrowBatchCollector (one per result ordering) and still has no streaming variant.

A ResultFormat doesn't have to deal with pipelines, batch indexes, the buffer, parking, the result, etc. It just converts chunks into units and the new PhysicalResultSink we introduced in #25257 and #25865 does the rest.

I agree the PR is large, but once it lands we can move both arrow (see #25801) and quack (see https://github.com/duckdb/duckdb-quack/pull/273) over, and then get rid of all of the collector stuff and special casing.

Maxxen:
Cool! Im not that familiar with this part of the code so just trying to understand more of the problems we're trying to solve here

Maxxen:
If @Tishj can sign off ill merge

PR #25729, opened 2026-09-15
Parse generic operators using longest matching tokens

Dtenwolde:
Fix https://github.com/duckdblabs/duckdb-internal/issues/10769

This PR improves the parsing of operators (such as `<->`). Previously, the tokenizer was doing a bit too much work and recognized several special cases. This caused an evergrowing list of special cases to be added. Such as for `<->`, which would've previously split into `<` `->` if not for a special case. 

The tokenizer now consumes the longest sequence of operator characters. There's still some special cases for comments (`--` and `/* */`), PostgreSQL's trailing `+/-` rules and tokens such as `::` and `:=`. The last one is a bit annoying. `:` is not part of the set of generic operators, and it's important that `::` and `:=` are not seen as individual tokens. Else `: :` or `:      =` becomes valid syntax. 

This also makes it possible to remove the list of generic operators from the grammar files. These generic operators are now matched by the `OperatorMatcher`. Operators with a special precedence (such as `~~`, `!~~` and arithmetic operators) are passed on and still handled in the grammar files. 

`ALL_OPERATORS` is used for two cases that make an operator deviate from its normal precedence level:
- Composite ANY/ALL expressions such as = ANY (...)
- Qualified operator syntax such as OPERATOR(schema.+)

In a future we could explore making the precendence ladder generated, but I consider that another follow-up.


Mytherin:
Thanks!

PR #25730, opened 2026-09-15
Fix streaming LEAD with UNION ALL

### LINKED ISSUE #25701: `LEAD` over `UNION ALL` duplicates rows at a pipeline boundary
### What happens?

A two-row `UNION ALL` followed by `LEAD` returns three rows, including two copies of the first input row with different window values.

### Code

```sql
SET threads=1;
SET enable_optimizer=true;

WITH src AS (
  SELECT 0 AS i
  UNION ALL
  SELECT 1 AS i
)
SELECT
  i,
  LEAD(i, 1, -1) OVER () AS x
FROM src;
```

### Expected

| i | x |
| ---: | ---: |
| 0 | 1 |
| 1 | -1 |

### Actual

| i | x |
| ---: | ---: |
| 0 | -1 |
| 0 | 1 |
| 1 | -1 |

### Notes From Minimization

- `SET enable_optimizer=false` changes the output to the expected two rows.
- For two collects containing N rows each, the optimized query returns 3N rows from 2N input rows.

### OS:

macOS 15.7.4, arm64

### DuckDB Version:

current main (`8bc48eb3fe36f13b5e57e5e136c3ba85cc7d74d4`, `v2.1.0-dev84598`)

### DuckDB Client:

CLI

### Hardware:

_No response_

### Full Name:

Yibo Dong

### Affiliation:

- [x] Yes, I have

- [x] Yes, I have

Yes

Syedowais312:
Fix incorrect results from streaming `LEAD` when its input is produced by `UNION ALL`.

  Each UNION child runs in a separate pipeline. Previously, a streaming window was finalized after each child, causing `LEAD` to flush buffered rows before the next UNION child was processed.

  Defer streaming-window finalization until the final UNION child completes.

  Added regression tests for `LEAD` over two and three `UNION ALL` inputs.

Fixes #25701


Mytherin:
Thanks - but I think pushing this into the `pipeline` is too invasive. Perhaps this can be solved in a less invasive manner?

Syedowais312:
Hey Mark, I’m thinking of an operator-local approach where PhysicalUnion would mark the non-final downstream streaming-window pipelines during construction, and let PhysicalStreamingWindow only flush on the final one. This avoids changes to Pipeline/PipelineExecutor. Does that sound reasonable?

Mytherin:
I think the issue here is a bit more fundamental - effectively `FinalExecute` operates as *final execute for this local state*, whereas the `PhysicalStreamingWindow` is incorrectly using it as *final execute for this operator*. I think the correct way to fix this would be to add a separate callback to the `PhysicalOperator` that is the *actual* final execute of that operator also if there's union operators. Then `PhysicalStreamingWindow` can build the semantics it requires on top of that callback.

PR #25732, opened 2026-09-15
C API V2 PR1 remaining review feedback

taniabogatsch:
See comments here: https://github.com/duckdb/duckdb/pull/24702


test/api/capi/v2/test_capi_v2_error_info.cpp
```diff
@@ -277,6 +283,7 @@ TEST_CASE("V2 error: error_info_destroy is null-safe", "[capi_v2][error]") {
 	}
 }
 
+// Why are these commented out?
```
taniabogatsch: @Maxxen - should I just remove them?

Maxxen: Sure, I can also do it later.

